PR #4297, opened 2021-01-19
[PATH] First pass at a post processor for Fanuc

shadowbane1000:
probably useful for other Fanuc controllers

I couldn't run the tests.  They always failed with 
FreeCAD: /home/user/dev/external/FreeCAD/src/Mod/Part/Gui/TaskDimension.cpp:1312: static void PartGui::ArcEngine::initClass(): Assertion `SoEngine::getClassTypeId() != SoType::badType()' failed.
Aborted (core dumped)
Though I highly doubt it's related to the new file getting added. 

I started by copying the mach2 post processor.  A few simple programs have been run on a real machine.  I have not yet tried the Rigid tapping on a real machine, but the generated code looks about right.

### DIFF
--- a/src/Mod/Path/PathScripts/post/fanuc_post.py
+++ b/src/Mod/Path/PathScripts/post/fanuc_post.py
@@ -1,0 +1,513 @@
+# ***************************************************************************
+# *   Copyright (c) 2014 sliptonic <shopinthewoods@gmail.com>               *
+# *   Copyright (c) 2021 shadowbane1000 <tyler@colberts.us>                 *
+# *                                                                         *
+# *   This file is part of the FreeCAD CAx development system.              *
+# *                                                                         *
+# *   This program is free software; you can redistribute it and/or modify  *
+# *   it under the terms of the GNU Lesser General Public License (LGPL)    *
+# *   as published by the Free Software Foundation; either version 2 of     *
+# *   the License, or (at your option) any later version.                   *
+# *   for detail see the LICENCE text file.                                 *
+# *                                                                         *
+# *   FreeCAD is distributed in the hope that it will be useful,            *
+# *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+# *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+# *   GNU Lesser General Public License for more details.                   *
+# *                                                                         *
+# *   You should have received a copy of the GNU Library General Public     *
+# *   License along with FreeCAD; if not, write to the Free Software        *
+# *   Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  *
+# *   USA                                                                   *
+# *                                                                         *
+# ***************************************************************************/
+from __future__ import print_function
+import FreeCAD
+from FreeCAD import Units
+import Path
+import argparse
+import datetime
+import shlex
+import os.path
+from PathScripts import PostUtils
+from PathScripts import PathUtils
+
+TOOLTIP = '''
+This is a postprocessor file for the Path workbench. It is used to
+take a pseudo-gcode fragment outputted by a Path object, and output
+real GCode suitable should be suitable for most Fanuc controllers.
+It has only been tested on a 21i-MB controller on a 3 axis mill.
+This postprocessor, once placed in the appropriate PathScripts folder,
+can be used directly from inside FreeCAD, via the GUI importer or via
+python scripts with:
+
+import fanuc_post
+fanuc_post.export(object,"/path/to/file.ncc","")
+'''
+
+now = datetime.datetime.now()
+
+parser = argparse.ArgumentParser(prog='fanuc', add_help=False)
+parser.add_argument('--no-header', action='store_true', help='suppress header output')
+parser.add_argument('--no-comments', action='store_true', help='suppress comment output')
+parser.add_argument('--line-numbers', action='store_true', help='prefix with line numbers')
+parser.add_argument('--no-show-editor', action='store_true', help='don\'t pop up editor before writing output')
+parser.add_argument('--precision', default='3', help='number of digits of precision, default=3')
+parser.add_argument('--preamble', help='set commands to be issued before the first command, default="G17\nG90"')
+parser.add_argument('--postamble', help='set commands to be issued after the last command, default="M05\nG17 G90\nM2"')
+parser.add_argument('--inches', action='store_true', help='Convert output for US imperial mode (G20)')
+parser.add_argument('--no-modal', action='store_true', help='Don\'t output the Same G-command Name USE NonModal Mode')
+parser.add_argument('--no-axis-modal', action='store_true', help='Don\'t output the Same Axis Value Mode')
+parser.add_argument('--no-tlo', action='store_true', help='suppress tool length offset (G43) following tool changes')
+
+TOOLTIP_ARGS = parser.format_help()
+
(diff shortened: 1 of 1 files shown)


Schildkroet:
Please prepend [PATH] to the title

Schildkroet:
Can this pp be used ONLY for 21i-MB or does it also run on Fanuc in general?

shadowbane1000:
I believe right now it is generic enough to work on most Fanuc controllers.  But since I only have access to the 21i-MB, I can't guarantee anything outside of that.  I'll get [PATH] put in the tile asap.

Schildkroet:
Then pls rename the pp to fanuc_post, so it is clear

shadowbane1000:
I have completed the requested changes.

src/Mod/Path/PathScripts/post/fanuc_post.py
```diff
@@ -0,0 +1,509 @@
...
+from PathScripts import PathUtils
+
+TOOLTIP = '''
+This is a postprocessor file for the Path workbench. It is used to
+take a pseudo-gcode fragment outputted by a Path object, and output
+real GCode suitable for a fanuc_21i-MB 3 axis mill. This postprocessor, once placed
+in the appropriate PathScripts folder, can be used directly from inside
+FreeCAD, via the GUI importer or via python scripts with:
+
+import fanuc_21imb
```
plaes, line 41: Don't you need to fix this after rename?

shadowbane1000: Yes.  I think there may be a couple of other leftovers from the original mach2 code as well.  Good catch.

src/Mod/Path/PathScripts/post/fanuc_post.py
```diff
@@ -0,0 +1,512 @@
+# ***************************************************************************
+# *   Copyright (c) 2014 sliptonic <shopinthewoods@gmail.com>               *
```
Schildkroet, line 2: Also update copyright info

src/Mod/Path/PathScripts/post/fanuc_post.py
```diff
@@ -0,0 +1,512 @@
...
+can be used directly from inside FreeCAD, via the GUI importer or via 
+python scripts with:
+
+import fanuc_post
+fanuc_post.export(object,"/path/to/file.ncc","")
+'''
+
+now = datetime.datetime.now()
+
+parser = argparse.ArgumentParser(prog='fanuc_21imb', add_help=False)
```
Schildkroet, line 49: Update prog

src/Mod/Path/PathScripts/post/fanuc_post.py
shadowbane1000: What should I do with the copyright information?  It's copied from Sliptonic's file..  just just update the date, or replace it with my name, or what?

Schildkroet: add a new line with current year and your name

src/Mod/Path/PathScripts/post/fanuc_post.py
shadowbane1000: done

shadowbane1000:
As far as I know, all suggestions have been addressed.

sliptonic:
Hi shadowbane1000.  Thanks for the contribution.  Can you please add to the description some details about how the post differs from linuxcnc or mach2?

I'm in the process of refactoring the postprocessor system to try to eliminate repeated code and applying my changes to this post will be difficult if I don't know what it's doing differently.

shadowbane1000:
Sure.

shadowbane1000:
By 'description' do you mean in the heredoc within the file, or right here, or the description of the pull request?

sliptonic:
I was thinking about the comment block at the top.  After the copyright and before the imports.
On the other hand, if you want to work with me, we could just team up to implement your fanuc post as part of the refactor and kill two birds with one stone.

shadowbane1000:
I'm okay with either way.  Do you have a branch you are working in that I could check out?

Here are the points I remember off hand.  I think there will be more as I work more with the machine and FreeCAD.

1) All text must be in uppercase.  That includes comments, gcodes, etc...

2) Several of the default options were reversed because Fanuc requires the opposite values.  This could have been resolved another way (calling the other post with options, or something like that)

3) header is different.  It needs to be:
'  %
  ;
  Onnnn (visible comment)`

Where nnnn is the program number.  To solve this, I put the filename (without .ngc) in that place.  So if saved as O1000.ngc, that line will have O1000.  The comment on that line is what is visible within the controller GUI, so it ideally would have the job name and freecad file name.  Right now I just have placeholder text in there since I couldn't see how to get that information.

4) It merges canned op sequences, so this:
` G83 X0 Y0
G80
G83 X1 Y1
G80` 
becomes
` G83 X0 Y0
X1 Y1
G80`
For this, I just skipped G80 if the next command was the same canned op.

5) It supports rigid tapping (though I am just getting to where I can test it).  A G81 or G83 is converted to a M29 G84 sequence (with some other settings) if there is a tap as the tool.

I think that is all of it right now.

shadowbane1000:
On second thought, I would like working with you on this.  My queries for the tool type are broken now that I'm on the latest build.  I was working against the latest AppImage as of a couple days ago.  However, I guess the app image generation has been broken for a while, so the version was a little too old.  I'd rather have it work with whatever is latest.

sliptonic:
I've pushed my branch to https://github.com/sliptonic/FreeCAD/tree/feature/post_refactor
The basic idea is to move most of the postprocessor logic into a base class in the file postprocessor.py  (note that this does not have the required <fn>_post.py  filename.
Then the individual posts would create a postprocessor object from that base class and override whatever they want.
So far, I've reimplemented the linuxcnc_post.py  file so you can look at it as an example.

Now I'm working on smoothie_post.py.   I want to be able to override the constructor method as well so default properties can be overrriden with more sensible defaults on a per-post basis. 

Also, I'm trying hard not to break anything in the existing post system so if people are using their own custom posts, they won't break.

shadowbane1000:
Question for you.  If I want to reverse the default state of an option, for example _modal should default to True, rather than False, How would you recommend going about changing the option.  I can override __init__ easily enough, and set it, but the argument is --modal and store_true.  Since the Fanuc controllers can have very limited memory, I would like to default it to saving as much as possible.

shadowbane1000:
Is there a better way to converse than via this pull request?

I could change the getArgs() in postprocessor.py to look at the current value and switch which argument it creates depending on the value that _ _ init _ _ set.  Seems a little messy though to do that with each argument (though it could be factored out to it's own method easily enough)

shadowbane1000:
perhaps the equivalent of:
if self.output_header:
          parser.add_argument('--no-header', action='store_true', help='suppress header output')
else:
        parser.add_argument('--header', action='store_true', help='include header output')

and then later:
            if args.no_header:
                self._output_header = False
            if args.header:
                self._output_header = True


Still feels a little messy though.

sliptonic:
> Is there a better way to converse than via this pull request?
> 
Absolutely.  Most of the Path devs hang out on gitter. 
https://gitter.im/FreeCAD/Path

> I could change the getArgs() in postprocessor.py to look at the current value and switch which argument it creates depending on the value that _ _ init _ _ set. Seems a little messy though to do that with each argument (though it could be factored out to it's own method easily enough)

I've changed that a little already.  I'll push an updated version in the next hour or so

PR #4301, opened 2021-01-22
Schedules bug.

JuanJ67m:
I leave a related post here:

https://forum.freecadweb.org/viewtopic.php?t=53934

a bug with Schedules apparently was not fixed, I mean, the Schedule container does not work.

here a sample:

image

image

[FreeCAD Arch Workbench - Simple Wall.zip](https://github.com/FreeCAD/FreeCAD/files/5853742/FreeCAD.Arch.Workbench.-.Simple.Wall.zip)

anyone can confirm if this bug will be fixed?


luzpaz:
@rebeltaz best to bite the bullet and migrate to v0.19  
@JuanJ67m this PR is against v0.18. Please make PR against `master`

wwmayer:
The PR makes no sense.

PR #4303, opened 2021-01-22
[Arch] Fixed typo in Arch_Schedule

Roy-043:
Typo fix


hyarion:
@Roy-043 Nice catch!
Do you have any example where this change fixes an issue?

Roy-043:
The type->ifctype conversion is there for compatibility reasons. Without it many schedules created with a previous version of the code would have to be edited. But 'type' should not be used for new schedules.

Here is the related forum discussion:
https://forum.freecadweb.org/viewtopic.php?f=23&t=53934

I have also edited the documentation a bit. But that still requires more work:
https://wiki.freecadweb.org/Arch_Schedule

yorikvanhavre:
Thanks!!

PR #4304, opened 2021-01-22
[Sketcher] Partial redundant notification and new solver equal distance constraint

abdullahtahiriyo:
Thread:
https://forum.freecadweb.org/viewtopic.php?f=8&t=53466&start=40#p464168

### DIFF
--- a/src/Mod/Sketcher/App/planegcs/GCS.cpp
+++ b/src/Mod/Sketcher/App/planegcs/GCS.cpp
@@ -1045,10 +1045,12 @@ int System::addConstraintArcDiameter(Arc &a, double *radius, int tagId, bool dri
     return addConstraintProportional(a.rad, radius, 0.5, tagId, driving);
 }
 
-int System::addConstraintEqualLength(Line &l1, Line &l2, double *length, int tagId, bool driving)
+int System::addConstraintEqualLength(Line &l1, Line &l2, int tagId, bool driving)
 {
-    addConstraintP2PDistance(l1.p1, l1.p2, length, tagId, driving);
-    return addConstraintP2PDistance(l2.p1, l2.p2, length, tagId, driving);
+    Constraint *constr = new ConstraintEqualLineLength(l1, l2);
+    constr->setTag(tagId);
+    constr->setDriving(driving);
+    return addConstraint(constr);
 }
 
 int System::addConstraintEqualRadius(Circle &c1, Circle &c2, int tagId, bool driving)
@@ -4051,7 +4053,7 @@ SolverReportingManager::Manager().LogToFile("GCS::System::diagnose()\n");
             // identifyDependentParametersSparseQR(J, jacobianconstraintmap, pdiagnoselist, true)
             //
             // Debug:
-            // auto fut = std::async(std::launch::deferred,&System::identifyDependentParametersSparseQR,this,J,jacobianconstraintmap, pdiagnoselist, false);
+            //auto fut = std::async(std::launch::deferred,&System::identifyDependentParametersSparseQR,this,J,jacobianconstraintmap, pdiagnoselist, false);
             auto fut = std::async(&System::identifyDependentParametersSparseQR,this,J,jacobianconstraintmap, pdiagnoselist, /*silent=*/true);
 
             makeSparseQRDecomposition( J, jacobianconstraintmap, SqrJT, rank, R, /*transposed=*/true, /*silent=*/false);
--- a/src/Mod/Sketcher/App/planegcs/GCS.cpp
+++ b/src/Mod/Sketcher/App/planegcs/GCS.cpp
@@ -226,6 +226,7 @@ public:
     void LogQRSystemInformation(const System &system, int paramsNum = 0, int constrNum = 0, int rank = 0);
 
     void LogGroupOfConstraints(const std::string & str, std::vector< std::vector<Constraint *> > constraintgroups);
+    void LogSetOfConstraints(const std::string & str,  std::set<Constraint *> constraintset);
     void LogGroupOfParameters(const std::string & str, std::vector< std::vector<double *> > parametergroups);
 
     void LogMatrix(const std::string str, Eigen::MatrixXd matrix);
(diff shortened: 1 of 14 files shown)


src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -590,6 +590,58 @@ bool CmdSketcherSelectMalformedConstraints::isActive(void)
...
+    // get the needed lists and objects
+    const std::vector< int > &solverpartiallyredundant = vp->getSketchObject()->getLastPartiallyRedundant();
+    const std::vector< Sketcher::Constraint * > &vals = Obj->Constraints.getValues();
+
+    getSelection().clearSelection();
+
+    // push the constraints
+    int i = 0;
+    for (std::vector< Sketcher::Constraint * >::const_iterator it= vals.begin();it != vals.end(); ++it,++i) {
+        for(std::vector< int >::const_iterator itc= solverpartiallyredundant.begin();itc != solverpartiallyredundant.end(); ++itc) {
```
vosk, line 629:
```suggestion
        for(int partial : solverpartiallyredundant) {
```

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -4560,20 +4584,28 @@ void System::identifyConflictingRedundantConstraints(   Algorithm alg,
                 conflictingTags.begin());
 
     // output of redundant tags
-    SET_I redundantTagsSet;
-    for (std::set<Constraint *>::iterator constr=redundant.begin();
-            constr != redundant.end(); ++constr)
+    SET_I redundantTagsSet, partiallyRedundantTagsSet;
+    for (std::set<Constraint *>::iterator constr=redundant.begin(); constr != redundant.end(); ++constr) {
```
vosk, line 4588:
```suggestion
    for (Constraint *constr : redundant) {
```

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -6395,6 +6415,13 @@ QString ViewProviderSketch::appendRedundantMsg(const std::vector<int> &redundant
                         redundant);
 }
 
+QString ViewProviderSketch::appendPartiallyRedundantMsg(const std::vector<int> &partiallyredundant)
```
vosk, line 6418: I know the code cut is soon, so i will NOT say these statics really are begging to be moved to an independent cpp/h pair.

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -4560,20 +4584,28 @@ void System::identifyConflictingRedundantConstraints(   Algorithm alg,
...
         redundantTagsSet.insert((*constr)->getTag());
+        partiallyRedundantTagsSet.insert((*constr)->getTag());
+    }
+
     // remove tags represented at least in one non-redundant constraint
-    for (std::vector<Constraint *>::iterator constr=clist.begin();
-        constr != clist.end(); ++constr) {
+    for (std::vector<Constraint *>::iterator constr=clist.begin(); constr != clist.end(); ++constr)
```
chennes, line 4594: Following @vosk's suggestions, I'd make the same change here, as long as you are editing the line anyway:
 ```suggestion
    for (auto &constr : clist)
```
(and then you can lose the *s in the next line)

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
abdullahtahiriyo: Thanks for reviewing. I will extend in this first comment beyond what is suggested to give you the rationale about why I tend to maintain style and reuse proven code at first instead of directly refactoring and deferring it.

Iterator based loop is equivalent to "new" (c++11) range based loop. There is no execution impact. Developers are used to it. One could talk about readability (range loop is generally more readable). But these files are full of iterator-based loops and it makes no sense to change the style of new code while leaving the rest (if one should be changed, all should, but iterators are still amply used in FreeCAD). I mostly tend to maintain the style of the existing files for new code. Obviously, the body of the loop is still expecting an iterator, so I reply to the intent rather than the fact that the code will not compile.

However, this file has a lot of analogous commands that are begging for a refactor. There are 3 or so "equal" commands save for if it is redundant, partially redundant or conflicting, such code repetition makes no sense and affects maintainability. In fact, the refactor should not just restrict to this file, but to  all the "chain" from GCS to the Viewer. All these are constraint ids (called Tags in the solver), it is simply the status that is different (redundant, conflicting,...). Therefore, one new class should unify this information and simplify the signaling mechanism. This is also desirable if the interface with the solver is to support future solver's, such as DeepSOIC's ConstraintSolver.

There are reasons why I do not refactor this at this time and prefer to replicate the existing structure. The first has to do with producing incremental code for new functionality that may need changes based on community feedback (it could even be removal of the feature and substitution with an alternative idea). The idea is to perform changes by addition, not by replacement. When the functionality settles down, then I think is a better time to consolidate and refactor. Unlike other projects where feedback comes from the customer after release, we work with very short latency in the feedback-implementation loop. This afternoon we will have the first impressions of this work and it is likely there will be change requests. The second relates to the release coming soon and not wanting to extend testing forever. Incremental code allows incremental testing. So I tend to produce the smallest differences, if possible reusing already tested code from other parts.

All that say, when v0.20 development cycle begins, it would be an appropriate time to refactor this. I have not forgotten about your other proposal either. If you identify parts that would benefit from refactoring, I am for refactoring.

src/Mod/Sketcher/App/planegcs/GCS.cpp
abdullahtahiriyo: idem

src/Mod/Sketcher/App/planegcs/GCS.cpp
abdullahtahiriyo: idem

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
vosk: Thank you for taking the time  to respond :)

PR #4306, opened 2021-01-23
v0.20: Normalize UI files

davidosterberg:
Normalize *.ui XML files throughout the code base. These changes has been made by a script from here:
https://github.com/davidosterberg/Qt-UI-file-sorter

Forum discussion
https://forum.freecadweb.org/viewtopic.php?f=10&t=54503

This first PR only beautifies the XML.
i.e the script guarantees that the C++ generated by uic is unchanged:

```sh
#!/bin/bash

git fetch upstream
git reset --hard upstream/master

i=0

find . -type f -name "*.ui" -print -exec ./add_if_identical {} --no-remove-stdset --no-remove-native --no-sort-qgridlayout \;
git commit -m "Beautify xml step 1: These changes do not affect the uic output (A)"
if [ $? -eq 0 ]; then
    i=$((i+1))
fi
git restore *.ui

find . -type f -name "*.ui" -print -exec ./add_if_identical {} --no-remove-stdset --no-sort-qgridlayout \;
git commit -m "Beautify xml step 1: These changes do not affect the uic output (B)"
if [ $? -eq 0 ]; then
    i=$((i+1))
fi
git restore *.ui

# Squash the commits
git reset --soft HEAD~"$i"
git commit -m 'Beautify xml step 1: These changes do not affect the uic output'
```

where the shell script `add_if_identical` is defined as:
```sh
cpp_before="$1_before.cpp"
cpp_after="$1_after.cpp"

uic "$1" > "$cpp_before"

../Qt-UI-file-sorter/sort_ui $2 $3 $4 "$1"

uic "$1" > "$cpp_after"

diff -u --color=always "$cpp_before" "$cpp_after"

if [ $? -eq 0 ]; then
    git add "$1"
    rm "$cpp_before"
    rm "$cpp_after"
fi
```


hyarion:
Just a thought:
I'm all for cleaner pull requests, but wouldn't this cause merge conflicts for pull requests that's already in the pipeline?
I think this can wait until after 0.19 is released, to avoid further delays.

davidosterberg:
> Just a thought:
> I'm all for cleaner pull requests, but wouldn't this cause merge conflicts for pull requests that's already in the pipeline?
> I think this can wait until after 0.19 is released, to avoid further delays.

You are right. There is no hurry.

wwmayer:
I added v0.20 to the title to make sure to postpone it.

davidosterberg:
@wwmayer , I have rebased this. This one should be save to merge.

wwmayer:
In XML a quotation mark usually is encoded as &quot; but this PR replaces them with the quotation mark symbol. Is this really safe? Or wouldn't it make more sense to extend the utility to preserve the &quot;?

davidosterberg:
It is safe according to https://stackoverflow.com/questions/5186179/when-do-i-need-to-use-the-quot-in-xml . It is also more readable.

luzpaz:
will this commit at all impact translations? I didn't see any strings change...but still wondering?

PR #4307, opened 2021-01-23
Part: fix ViewProviderExt::updateVisual() on incomplete mesh

realthunder:
For some invalid shapes, there may be faces with null mesh. In this case, Part::ViewProviderExt may produce an invalid SoBrepFaceSet::partIndex.

### DIFF
--- a/src/Mod/Part/Gui/ViewProviderExt.cpp
+++ b/src/Mod/Part/Gui/ViewProviderExt.cpp
@@ -1133,7 +1133,10 @@ void ViewProviderPartExt::updateVisual()
             const TopoDS_Face &actFace = TopoDS::Face(faceMap(i));
             // get the mesh of the shape
             Handle (Poly_Triangulation) mesh = BRep_Tool::Triangulation(actFace,aLoc);
-            if (mesh.IsNull()) continue;
+            if (mesh.IsNull()) {
+                parts[ii] = 0;
+                continue;
+            }
 
             // getting the transformation of the shape/face
             gp_Trsf myTransf;


chennes:
Do you have a test case handy?

realthunder:
It's hard to produce a test case for that, as it requires an invalid shape, which is visually broken in the first place. Besides, the way the upstream `SoBrepFaceSet` uses `partIndex` means this type of error is mostly harmless, which is why this bug is undiscovered for more than 10 years, I guess. I found it because I use `partIndex` heavily in my branch, and can actually lead to crash if not careful.

chennes (approved):
Got it. OK, well I can say that it compiles and runs, and doesn't have any obvious negative side effects. If it solves your problem, then it looks good to me.

wwmayer:
The example is a bit constructed but nevertheless demonstrates what can happen.

Create a cylinder solid and out of its faces create a compound

```
s = Part.makeCylinder(3,10)
c = Part.Compound(s.Faces)

doc = App.newDocument()
part = doc.addObject("Part::Feature")
part.Shape = c
```

The cylindrical face has id 1 and the planar faces the id 2 and 3.

Now if you replace the compound with a new compound whose first face is invalid
```
line = Part.makeLine(App.Vector(), App.Vector(10))
line2 = Part.makeLine(App.Vector(10), App.Vector(10,10))
wire = Part.Wire((line, line2))
face = Part.Face(wire) # broken face as it's not closed
c2 = Part.Compound([face, s.Face2, s.Face3])
part.Shape = c2
```
you end up with an object where both planar faces now show id 1, instead of 2 and 3

The PR fixes this issue.

chennes:
Thank you for the detailed explanation, @wwmayer.

PR #4292, opened 2021-01-18
[TD] MDIViewPage: Add support for Qt5 automatic paper size handling …

aapo-aapo:
…with many new paper sizes possible.  See forum request and discussion in [here](https://forum.freecadweb.org/viewtopic.php?f=35&t=54512), apparently "Arch C" and other paper sizes are missing.  This commit should add all the new paper sizes in Qt 5.3 and newer, whereas for Qt 4 the paper size handling is unchanged and only the old paper sizes remain possible.

Thank you for creating a pull request to contribute to FreeCAD! To ease integration, please confirm the following:

- [x] Branch rebased on latest master `git pull --rebase upstream master`
- [ ] Unit tests confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [x] Commit message is [well-written](https://chris.beams.io/posts/git-commit/)
- [ ] Commit message includes `issue #<id>` or `fixes #<id>` where `<id>` is the [associated MantisBT](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) issue id if one exists

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.19 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=34586).

---

PR #4309, opened 2021-01-24
[TD] add Arch paper size templates

donovaly:
thanks to the recently merged PR #4292 one can use custom page sizes such as the templates in https://wiki.freecadweb.org/Arch_templates

This PR add them to be directly provided by FC.

Forum thread: https://forum.freecadweb.org/viewtopic.php?f=35&t=54512&start=10#p470230


chennes (approved):
This looks great. I double-checked the sizes and they all correspond correctly to the table at https://www.agooddaytoprint.com/page/paper-size-chart-faq. All the template size fields have been updated correctly, and everything seems to be rendering the correct size. Thanks!

chennes:
I suggest removing the `xml:space="default"` from the editable fields: Qt's DOM parser doesn't like the xml namespace, and when FC is compiled in debug mode this results in an `abort()` being called in FreeCAD. It's an upstream bug, but Qt DOM is no longer maintained so I doubt it will get fixed.

donovaly:
> I suggest removing the `xml:space="default"` from the editable fields: Qt's DOM parser doesn't like the xml namespace

As I explained in the forum:
https://forum.freecadweb.org/viewtopic.php?f=35&t=54512&start=20#p470480
and following posts, the xml:space is a valid SVG 1.x tag. Thus it is a bug that FreeCAD does not handle it correctly and this needs to be fixed instead.

Not following the ISO standards will sooner or later make other problems so that cannot be our way to go. Since the bug in FC only occurs in debug mode, there must be a solution to ignore the xml:space DOM parser results when the parser cannot handle this.

donovaly:
OK, for the meantime I removed the xml:space tags so that this PR can be merged.

yorikvanhavre:
Thanks guys!

PR #4310, opened 2021-01-24
[TD] Fix equilateral angular tolerance missing degree unit char, fix arbitary under/overtolerances

aapo-aapo:
This PR fixes two bugs:  The new equilateral tolerances are missing a degree sign for angles. Also, arbitrary under/overtolerances have somehow lost the possibility to be set separately.  E.g. "std project tolerance" for under, and "as tight as possible" for upper.  The bugs were reported [here](https://forum.freecadweb.org/viewtopic.php?f=35&t=54740)

---

### DIFF
--- a/src/Mod/TechDraw/App/DrawViewDimension.cpp
+++ b/src/Mod/TechDraw/App/DrawViewDimension.cpp
@@ -110,9 +110,9 @@ DrawViewDimension::DrawViewDimension(void)
     ADD_PROPERTY_TYPE(References3D, (0,0), "", (App::Prop_None), "3D Geometry References");
     References3D.setScope(App::LinkScope::Global);
 
-    ADD_PROPERTY_TYPE(FormatSpec, (getDefaultFormatSpec()), "Format", App::Prop_Output,"Dimension Format");
-    ADD_PROPERTY_TYPE(FormatSpecUnderTolerance, (getDefaultFormatSpec(true)), "Format", App::Prop_Output, "Dimension tolerance format");
-    ADD_PROPERTY_TYPE(FormatSpecOverTolerance, (getDefaultFormatSpec(true)), "Format", App::Prop_Output, "Dimension tolerance format");
+    ADD_PROPERTY_TYPE(FormatSpec, (getDefaultFormatSpec()), "Format", App::Prop_Output,"Dimension format");
+    ADD_PROPERTY_TYPE(FormatSpecOverTolerance, (getDefaultFormatSpec(true)), "Format", App::Prop_Output, "Dimension overtolerance format");
+    ADD_PROPERTY_TYPE(FormatSpecUnderTolerance, (getDefaultFormatSpec(true)), "Format", App::Prop_Output, "Dimension undertolerance format");
     ADD_PROPERTY_TYPE(Arbitrary,(false), "Format", App::Prop_Output, "Value overridden by user");
     ADD_PROPERTY_TYPE(ArbitraryTolerances, (false), "Format", App::Prop_Output, "Tolerance values overridden by user");
 
@@ -143,6 +143,7 @@ DrawViewDimension::DrawViewDimension(void)
 
     // by default EqualTolerance is true, thus make UnderTolerance read-only
     UnderTolerance.setStatus(App::Property::ReadOnly, true);
+    FormatSpecUnderTolerance.setStatus(App::Property::ReadOnly, true);
 
     measurement = new Measure::Measurement();
     //TODO: should have better initial datumLabel position than (0,0) in the DVP?? something closer to the object being measured?
@@ -213,11 +214,16 @@ void DrawViewDimension::onChanged(const App::Property* prop)
                 UnderTolerance.setValue(0.0);
                 OverTolerance.setReadOnly(true);
                 UnderTolerance.setReadOnly(true);
+                FormatSpecOverTolerance.setReadOnly(true);
+                FormatSpecUnderTolerance.setReadOnly(true);
             }
             else {
                 OverTolerance.setReadOnly(false);
-                if (!EqualTolerance.getValue())
+                FormatSpecOverTolerance.setReadOnly(false);
+                if (!EqualTolerance.getValue()) {
                     UnderTolerance.setReadOnly(false);
+                    FormatSpecUnderTolerance.setReadOnly(false);
+                }
             }
             requestPaint();
         }
@@ -233,11 +239,15 @@ void DrawViewDimension::onChanged(const App::Property* prop)
                 UnderTolerance.setValue(-1.0 * OverTolerance.getValue());
                 UnderTolerance.setUnit(OverTolerance.getUnit());
                 UnderTolerance.setReadOnly(true);
+                FormatSpecUnderTolerance.setValue(FormatSpecOverTolerance.getValue());
+                FormatSpecUnderTolerance.setReadOnly(true);
             }
             else {
                 OverTolerance.setConstraints(&ToleranceConstraint);
-                if (!TheoreticalExact.getValue())
+                if (!TheoreticalExact.getValue()) {
                     UnderTolerance.setReadOnly(false);
+                    FormatSpecUnderTolerance.setReadOnly(false);
+                }
             }
             requestPaint();
         } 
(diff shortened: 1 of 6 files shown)


donovaly:
Don't beat me for it but I think this PR would introduce 2 regressions ;-).
I commented in the forum: https://forum.freecadweb.org/viewtopic.php?f=35&t=54740&p=470651#p470651
Let's discuss this in the forum further.

donovaly:
Thanks for this PR. I have the following remarks:

A. the Undertolerance format specifier appears above the overtolerance specifier - it should be the ordered alphabetically
B. when Equal Tolerance is true, then Undertolerance is purposely locked and so should the Undertolerance specifier.
C. changing Arbitrary Tolerances to true or false, does nothing visible in the drawing
D. change Arbitrary Tolerances to true, the write something into the Overtolerance Spec field and nothing happens in the drawing. It should immediately display the typed text as it is the case for the Format Spec field. 
E. in the Dimension dialog set Arbitrary Tolerance Text to True but nothing happens
F. in the Dimension dialog there is no field for the arbitrary undertolerance

aapo-aapo:
> Thanks for this PR. I have the following remarks:
> 
> A. the Undertolerance format specifier appears above the overtolerance specifier - it should be the ordered alphabetically
> B. when Equal Tolerance is true, then Undertolerance is purposely locked and so should the Undertolerance specifier.
> C. changing Arbitrary Tolerances to true or false, does nothing visible in the drawing
> D. change Arbitrary Tolerances to true, the write something into the Overtolerance Spec field and nothing happens in the drawing. It should immediately display the typed text as it is the case for the Format Spec field.
> E. in the Dimension dialog set Arbitrary Tolerance Text to True but nothing happens
> F. in the Dimension dialog there is no field for the arbitrary undertolerance

Okay, I have now amended the PR, and as agreed on the forum, all of the above should be fixed now.  As discussed in the forum, this PR fixes the angular label printing to standard-compliant form; and all the redraw regressions with this and other recent pull requests should also be fixed.  So, this should now be ready for review

donovaly:
Thanks for the commits. I tested again and only B is not fixed for me.

I think this is not an issue, but I am not sure:

G. Text and Tolerance is Arbitrary, then only text is shown, not also tolerance

I have no strong opinion regarding this. Is it sensible to show also arbitrary tolerances when dimension is arbitrary?

donovaly:
With these 2 files (include your changes) also B is fixed for me: 
[TechDraw.zip](https://github.com/FreeCAD/FreeCAD/files/5884367/TechDraw.zip)

aapo-aapo:
> With these 2 files (include your changes) also B is fixed for me:

Thanks, I added these and fixed yet another related problem in Dimension Task

wwmayer:
@donovaly @aapo-aapo Is it ready to be merged now?

aapo-aapo:
> @donovaly @aapo-aapo Is it ready to be merged now?

Although this is ready on my part, I'd wait a little bit more, if there's no hurry.  I believe @chennes was still planning to review this, and he usually has good suggestions.  Also, @donovaly is maybe gonna have a look on this tonight once again.

chennes:
I am trying to be particularly thorough with my testing on this one, since we have had some regressions here recently. But of course if @wwmayer thinks it's ready to go, then it's ready!

chennes:
> I have no strong opinion regarding this. Is it sensible to show also arbitrary tolerances when dimension is arbitrary?

Is it possible to indicate to the user that they cannot set tolerance format specs if the dimension is arbitrary? Disable/gray out those fields or something? Also, the tooltips on the over and under format specs are the same, which was very confusing when my window was too narrow to show the full name of the property. Can you change those tips to include the words "over" and "under"?

aapo-aapo:
> Is it possible to indicate to the user that they cannot set tolerance format specs if the dimension is arbitrary? Disable/gray out those fields or something?

Yes, the fields should now generally be disabled it the tolerance format spec(s) cannot be set for some reason.  In the case of arbitrary dimensions, the fields are not disabled, because based on the comment by @donovaly I realized that with arbitrary dimensions, the user should always be able to use arbitrary tolerances.  So, with latest version of PR, arbitrary dimensions can have arbitrary (or non-arbitrary) tolerances, and thus the format spec fields are editable in that case, too.

> Also, the tooltips on the over and under format specs are the same, which was very confusing when my window was too narrow to show the full name of the property. Can you change those tips to include the words "over" and "under"?

chennes:
A couple bugs/unexpected behaviors (might be intended, might not!):
* If "Equal tolerances" is set to `true` and both `Arbitrary` and `Arbitrary tolerances` are `true`, the tolerances do not display (but you can interact with the format field and menu).
* On an angular measure, if `Arbitrary` is set to `false` and `Arbitrary tolerances` is set to `true`, the degree sign is added to the tolerance value.

aapo-aapo:
> A couple bugs/unexpected behaviors (might be intended, might not!):

I'm not sure how these cases are supposed to work, but because I liked your suggested logic, I implemented it.  It was a bit more complicated than I first thought, and the changes all concerned the new Equal Tolerances, so I added these changes in a new commit.  Thanks for reviewing!

donovaly:
Thanks aapo, I tested again and all previously reported issues are fixed, but there are further issues:

H. set arbitrary text to True, then arbitrary tolerance text must be disabled, this will also fix the issues J
I. this issue only applies for angular dimensions; set arbitrary text to True, set Theoretically Exact to true -> as result you don't get the arbitrary text
I. this issue only applies for angular dimensions; set Theoretically Exact to true, now toggle arbitrary tolerance text to true/false -> the degree sign is added/removed. The fix for this is the one for H.

donovaly:
This diff on top of your changes fix issues H, I and J for me:
[4310-I-H-J.zip](https://github.com/FreeCAD/FreeCAD/files/5891145/4310-I-H-J.zip)

If these changes are in the PR do the job for you as well, then from my point of view @wwmayer can merge.

aapo-aapo:
> If these changes do the job for you as well and you included them to the PR, then from my point of view @wwmayer can merge. Otherwise I will check tomorrow again.

Yes, they concern the TheoreticallyExact dimensions, which I have to admit I don't use, so these details unfortunately were not on my radar.  The changes look fine and worked, so I added them verbatim; with one extra line to turn automatically ArbitraryTolerances to false, if the user selects TheoreticallyExact.

chennes (approved):
Nice work, this is looking good. I tried as many bizarre combinations of values as I could think of, and everything worked in a way that made sense to me. It is clear that you put in a lot of work making sure this is standards-compliant in its default state. I took a brief look at the code and nothing jumped out at me, and the behavior of the UI is as expected.

donovaly:
> so I added them verbatim; with one extra line to turn automatically ArbitraryTolerances to false, if the user selects TheoreticallyExact.

Why? I think just disabling is sufficient so when you accidentally clicked the Theoretically Exact" option, your settings won't be lost on clicking the option subsequently again.
If you insist on this, however, the  you should do this also in TaskDimension.cpp to be consistent.

aapo-aapo:
> Why? I think just disabling is sufficient so when you accidentally clicked the Theoretically Exact" option, your settings won't be lost on clicking the option subsequently again.
> If you insist on this, however, the you should do this also in TaskDimension.cpp to be consistent.

I admit that I'd prefer that the editing fields would not have any effect on others, but this one I actually added to be consistent with your code, which zeroes the "OverTolerance" and "UnderTolerance" fields, if you click the "Theoretically Exact" option.  At first, I didn't like it, but learned to live with it, and wanted to be consistent.

That said, you're absolutely right about being consistent about TaskDimension.cpp, so I added the same there.

donovaly:
> being consistent about TaskDimension.cpp, so I added the same there.

Fine with me, but this change is not yet in. When it is in, everything is ready to be merged from my point of view.

donovaly (approved):
> being consistent about TaskDimension.cpp, so I added the same there.

Fine with me, but this change is not yet in. When it is in, everything is ready to be merged from my point of view..

aapo-aapo:
> Fine with me, but this change is not yet in. When it is in, everything is ready to be merged from my point of view..

Sorry, I forgot to push the change, but now it is pushed.  So, now everything should finally be in order, and this can be merged.  Phew!

donovaly:
> now everything should finally be in order, and this can be merged.

Yes. Many thanks!

wwmayer:
Since everybody is happy now I will merge it...

aapo-aapo:
> Since everybody is happy now I will merge it...

Thanks!

PR #4311, opened 2021-01-24
[v0.20]Update Libpack to 12.4.3

sgrogan:

- [x ] Branch rebased on latest master `git pull --rebase upstream master`
- [ x] Commit message is [well-written](https://chris.beams.io/posts/git-commit/)

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -284,12 +284,12 @@ before_install:
        # clcache stats before compilation
        cmd.exe //C 'C:\Users\travis\build\FreeCAD\FreeCAD\clcache.exe -s'
 
-       curl -L https://github.com/FreeCAD/FreeCAD/releases/download/0.19_pre/FreeCADLibs_12.1.6_x64_VC15.7z --output FreeCADLibs.7z
+       curl -L https://github.com/FreeCAD/FreeCAD/releases/download/0.19_pre/FreeCADLibs_12.4.3_x64_VC17.7z --output FreeCADLibs.7z
        7z x FreeCADLibs.7z -oFreeCADLibs > /dev/null
        rm -f FreeCADLibs.7z
-       export FREECAD_LIBPACK_DIR="$TRAVIS_BUILD_DIR\FreeCADLibs\FreeCADLibs_12.1.6_x64_VC15"
+       export FREECAD_LIBPACK_DIR="$TRAVIS_BUILD_DIR\FreeCADLibs\FreeCADLibs_12.4.3_x64_VC17"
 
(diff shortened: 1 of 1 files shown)


sgrogan:
Forum thread https://forum.freecadweb.org/viewtopic.php?f=10&t=54561#p468923

yorikvanhavre:
The windows build failed with an error likely due to this new libpack (shiboken python module not found I think)

sgrogan:
I'll look, thanks

wwmayer:
It's rather unlikely that shiboken is the culprit. Due to its hook scripts to handle importing modules its name appears in error messages and thus may confuse users to think it's related to it.
The real error is: ImportError: DLL load failed while importing Fem: The specified module could not be found.

It's probably best to test the libpack locally and running it with the debugger of VS. This may tell you which DLL exactly causes the trouble.

sgrogan:
I don't have the problem locally with a clean build.
I will try with a scratch build and make sure I copy the dependencies the same as on Travis.

Good tip though, in debug mode I get some FEM test errors that mention shiboken.

wwmayer:
Since Py3.8 a pyd file does not any longer respect the PATH system environment variable to load DLLs it depends on. Instead os.add_dll_directory() must be used to add all needed DLL search paths.

What may happen here is that not all DLLs are copied over from the Libpack and thus the Fem.pyd file cannot be loaded. One way to fix this is by setting the environment variable FREECAD_LIBPACK_BIN that points to the bin directory of the LibPack.
The other way is to find out which DLL isn't copied over.

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/4363 is a test to check if setting the environment variable FREECAD_LIBPACK_BIN fixes the problem.

wwmayer:
This is the log of my tested PR https://travis-ci.org/github/FreeCAD/FreeCAD/jobs/757531103 and it even fails to load three modules.
Yesterday I downloaded this LibPack and built FreeCAD locally. Everything  was working fine after defining the env. variable FREECAD_LIBPACK_BIN.

wwmayer:
Not sure how to fix this one. So, I will postpone it for v0.20

sgrogan:
OK
It works for me locally without setting the env variable. I will post a test build in the forum and see if my build is somehow not portable.

wwmayer:
> It works for me locally without setting the env variable.

I looked at the script and it moves the content of the LibPack's bin folder to the build folder. That's why it works w/o the env. variable.

wwmayer:
With fcf5a13eab7 LP 12.5 is used which now makes this PR obsolete.

PR #4312, opened 2021-01-24
[WIP] [Spreadsheet] allow valid expressions as alias

donovaly:
currently it is not possible to set e.g. an alias named "Höhe" despite this is a valid expression and you can for example use non-ASCII characters already use in sketch constraint names.

To overcome this issue, this PR checks the alias if it is a valid expression and not if the alias contains only ASCII characters.
See this thread why this PR is marked as work in progress: https://forum.freecadweb.org/viewtopic.php?f=10&t=54754

(- it also adds 2 comments and some whitespace automagically added by MSVC)

### DIFF
--- a/src/Mod/Spreadsheet/App/PropertySheet.cpp
+++ b/src/Mod/Spreadsheet/App/PropertySheet.cpp
@@ -113,7 +113,10 @@ const Cell * PropertySheet::getValueFromAlias(const std::string &alias) const
 
 bool PropertySheet::isValidAlias(const std::string &candidate)
 {
-    static const boost::regex gen("^[A-Za-z][_A-Za-z0-9]*$");
+    // the following characters and spaces are invalid for expressions and thus aliases:
+    // +-*/^_°<>(){}[].,=#!?§$%&:;\|~´`'" 
+    // we will not check this using a regexp but by parsing
+
     boost::cmatch cm;
 
     /* Check if it is used before */
@@ -124,22 +127,34 @@ bool PropertySheet::isValidAlias(const std::string &candidate)
     if (ExpressionParser::isTokenAUnit(candidate))
         return false;
 
-    /* Check to make sure it doesn't match a cell reference */
-    if (boost::regex_match(candidate.c_str(), cm, gen)) {
-        static const boost::regex e("\\${0,1}([A-Z]{1,2})\\${0,1}([0-9]{1,5})");
+    App::Expression* ExpressionResult = 0;
 
-        if (boost::regex_match(candidate.c_str(), cm, e)) {
-            const boost::sub_match<const char *> colstr = cm[1];
-            const boost::sub_match<const char *> rowstr = cm[2];
-
-            // A valid cell address?
-            if (App::validRow(rowstr.str()) >= 0 && App::validColumn(colstr.str()) >= 0)
-                return false;
-        }
-        return true;
+    // check if the alias is a valid expression by parsing it
+    try {
+        ExpressionResult = App::ExpressionParser::parse(owner, candidate.c_str());
     }
-    else
+    catch (Base::Exception& exception) {
+        ExpressionResult = new App::StringExpression(owner, candidate.c_str());
+        //new App::DocumentObjectExecReturn(exception.what());
+        std::ostringstream ss;
+        ss << "Failed to adjust link for " << owner->getFullName() << " in expression "
+            << ExpressionResult->toString() << ": " << exception.what();
+        throw Base::RuntimeError(ss.str());
         return false;
+    }
+
+    // it could be parsed now check it doesn't match a cell reference
+    static const boost::regex regExpression("\\${0,1}([A-Z]{1,2})\\${0,1}([0-9]{1,5})");
+    if (boost::regex_match(candidate.c_str(), cm, regExpression)) {
+        const boost::sub_match<const char*> colstr = cm[1];
+        const boost::sub_match<const char*> rowstr = cm[2];
+
+        // A valid cell address?
+        if (App::validRow(rowstr.str()) >= 0 && App::validColumn(colstr.str()) >= 0)
+            return false;
+    }
+
+    return true;
 }
 
 std::set<CellAddress> PropertySheet::getUsedCells() const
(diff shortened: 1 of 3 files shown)


src/Mod/Spreadsheet/App/PropertySheet.cpp
```diff
@@ -124,22 +127,34 @@ bool PropertySheet::isValidAlias(const std::string &candidate)
...
         return false;
 
-    /* Check to make sure it doesn't match a cell reference */
-    if (boost::regex_match(candidate.c_str(), cm, gen)) {
-        static const boost::regex e("\\${0,1}([A-Z]{1,2})\\${0,1}([0-9]{1,5})");
-
-        if (boost::regex_match(candidate.c_str(), cm, e)) {
-            const boost::sub_match<const char *> colstr = cm[1];
-            const boost::sub_match<const char *> rowstr = cm[2];
+    App::Expression* ExpressionResult = 0;
```
hyarion, line 130:
```suggestion
    App::Expression* ExpressionResult = nullptr;
```

src/Mod/Spreadsheet/App/PropertySheet.cpp
```diff
@@ -124,22 +127,34 @@ bool PropertySheet::isValidAlias(const std::string &candidate)
...
     }
-    else
+    catch (Base::Exception& exception) {
+        ExpressionResult = new App::StringExpression(owner, candidate.c_str());
+        //new App::DocumentObjectExecReturn(exception.what());
+        std::ostringstream ss;
+        ss << "Failed to adjust link for " << owner->getFullName() << " in expression "
+            << ExpressionResult->toString() << ": " << exception.what();
+        throw Base::RuntimeError(ss.str());
         return false;
```
hyarion, line 143: No need to return after throw, everything after return in the scope is dead code
```suggestion
```

src/Mod/Spreadsheet/App/PropertySheet.cpp
```diff
@@ -124,22 +127,34 @@ bool PropertySheet::isValidAlias(const std::string &candidate)
...
+    App::Expression* ExpressionResult = 0;
 
-            // A valid cell address?
-            if (App::validRow(rowstr.str()) >= 0 && App::validColumn(colstr.str()) >= 0)
-                return false;
-        }
-        return true;
+    // check if the alias is a valid expression by parsing it
+    try {
+        ExpressionResult = App::ExpressionParser::parse(owner, candidate.c_str());
```
hyarion, line 134: `ExpressionResult` doesn't need to be used after parsing.
```suggestion
        delete App::ExpressionParser::parse(owner, candidate.c_str());
```

src/Mod/Spreadsheet/App/PropertySheet.cpp
```diff
@@ -124,22 +127,34 @@ bool PropertySheet::isValidAlias(const std::string &candidate)
...
     }
-    else
+    catch (Base::Exception& exception) {
+        ExpressionResult = new App::StringExpression(owner, candidate.c_str());
+        //new App::DocumentObjectExecReturn(exception.what());
+        std::ostringstream ss;
+        ss << "Failed to adjust link for " << owner->getFullName() << " in expression "
+            << ExpressionResult->toString() << ": " << exception.what();
```
hyarion, line 141: You already have the string in `candidate` no need in creating a temporary object just for printing
```suggestion
        std::ostringstream ss;
        ss << "Failed to adjust link for " << owner->getFullName() << " in expression "
            << candidate << ": " << exception.what();
```

hyarion (changes requested):
I would prefer if white space changes where not committed for files and areas that aren't otherwise modified.
This for two reasons:
1. Shorter reviews are easier to read
2. It creates unnecessary merge conflicts with other PRs

If you want to keep the white space changes, I would prefer it to be placed in a separate commit. Same goes for the bonus comments.

donovaly:
I am surprised that you reviewed this. WIP stands for "work in progress" and I only made this PR to have a reference what I am talking about in the forum. Thus I did not care yet for optimal code, messages etc.
Meanwhile it also turned out that the PR is the wrong approach. So I will have to close this PR soon.

> I would prefer if white space changes where not committed for files and areas that aren't otherwise modified.

The point is that MS Visual C add then automagically when you copy code around. I am simply too laze to undo them, especially since they follow also out coding guidelines.

src/Mod/Spreadsheet/App/PropertySheet.cpp
donovaly: Ah, thanks. I was not aware that there is a difference since in other places of the code a '0' is used as if that would be the equivalent of a nullptr.

src/Mod/Spreadsheet/App/PropertySheet.cpp
donovaly: As said above, the code was not intended to go in, thus I just copy/pasted existing code just to get the error output and to debug.
Sorry that I caused work for you.

src/Mod/Spreadsheet/App/PropertySheet.cpp
hyarion: nullptr was introduced in c++11, both works but nullptr is the preferred way nowadays in c++ land

hyarion:
> I am surprised that you reviewed this. WIP stands for "work in progress" and I only made this PR to have a reference what I am talking about in the forum. Thus I did not care yet for optimal code, messages etc.
> Meanwhile it also turned out that the PR is the wrong approach. So I will have to close this PR soon.

Oh, I thought WIP was just that it shouldn't be merged but feedback is welcome.

Anyway, I don't mind giving feedback early on but I'll keep it in mind that it might just be for discussion, sorry :)

> > I would prefer if white space changes where not committed for files and areas that aren't otherwise modified.
> 
> The point is that MS Visual C add then automagically when you copy code around. I am simply too laze to undo them, especially since they follow also out coding guidelines.

I don't use MS Visual-tools but in all git tools I've used I've been able to select which files / chunks / lines I want to commit. I would be suprrised if MS doesn't have something similar.
Keeping commits small is important to make it review friendly. If the committer is lazy the reviewer gets more work. More work for reviewers => longer time for a PR to get merged ;)

donovaly:
> > The point is that MS Visual C add then automagically when you copy code around. I am simply too laze to undo them, especially since they follow also out coding guidelines.
> 
> I don't use MS Visual-tools but in all git tools

I speak about the Visual C IDE. When you copy a portion of code around, it reformats the code according to the set coding style. As i said, I am too laze to revert this and the result follows also out inofficial style guidelines. Moreover, some lines with whitespace don't harm.

donovaly:
Closing as discussed in the form as the wrong approach.

PR #4313, opened 2021-01-25
[TD] show by default center marks

donovaly:
as requested here: https://forum.freecadweb.org/viewtopic.php?f=35&t=54718#p470334 and following


yorikvanhavre:
Makes sense... Thx!

PR #4315, opened 2021-01-25
[Path]: Bugfix/vcarve selection and feature processing

mlampert:
fixes issue https://forum.freecadweb.org/viewtopic.php?f=15&t=54726 by
* processing BaseShapes
* allowing multiple models and base objects
* only fall back to job model if no Base and no BaseShapes are set
* only call the base addBaseGeometry() once

This also disables the base geometry list widget resizing because it ends up way too small to be usable on HPD monitors - and I needed to see what I'm doing in order to fix this issue.


sliptonic:
Built, verified.  Works with both empty base object list and individually selected items.  Vcarve is looking very solid now.

PR #4317, opened 2021-01-25
[Draft] Fix snap to center of faces of solids

Roy-043:
The code would only find a center snap on the face with index=0 of solids.

In V0.19 Part.getShape was introduced (line 399). But not all consequences were not fully implemented.

In the '# we are snapping to an edge' section (line 411) the code could be cleaned up. There is no need to check if the index of the edge is correct for the parent object since we are no longer dealing with a parent object. That portion was effectively dead code.

The '# we are snapping to a face' section (line 429 in the revised code) has been modified accordingly, which fixes the bug.

Forum discussion:
https://forum.freecadweb.org/viewtopic.php?f=23&t=54747

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_snapper.py
+++ b/src/Mod/Draft/draftguitools/gui_snapper.py
@@ -410,14 +410,8 @@ class Snapper:
                 if (not self.maxEdges) or (len(shape.Edges) <= self.maxEdges):
                     if "Edge" in comp:
                         # we are snapping to an edge
-                        edge = None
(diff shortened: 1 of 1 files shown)


chennes (approved):
This runs without any problems and fixes the issue noted in the description. Some light testing did not reveal any regressions. Thanks for making the change, I don't often use snap to center, but never would use it at all if it didn't work!

yorikvanhavre:
Makes sense. Thanks for the fix and review!

PR #4318, opened 2021-01-25
[UI] Replace SVG icons for Surface WB commands

bitacovir:
Icons of Surface WB do not follow Artwork Guidelines and could have a better quality. A design improvement was requested.

This commit replace SVG files with new icons for Surface wb commands. Also, it makes the necessary changes on Command.cpp and Surface.qrc files.

The new SVG icons follow the FreeCAD Artwork Guidelines and were saved as Plain SVG format.

Forum Discussion: https://forum.freecadweb.org/viewtopic.php?f=34&t=50823


chennes:
The changes to `Command.cpp` and `Surface.qrc` compile without issue and all five main icons display. As usual, very nice design work. I gather the remaining icons represent features that are not yet implemented -- if that's not correct and there is a way to make them show in FreeCAD I'd be happy to test them as well.

chennes (approved):

bitacovir:
> The changes to `Command.cpp` and `Surface.qrc` compile without issue and all five main icons display. As usual, very nice design work. I gather the remaining icons represent features that are not yet implemented -- if that's not correct and there is a way to make them show in FreeCAD I'd be happy to test them as well.

Thanks for your review. It looks like only 5 commands are implemented, but I replaced all existing icons in the icon folder. Maybe in the future someone will work expanding the Surface WB and will use the available icons.

luzpaz:
Awesome work @bitacovir and thanks for testing @chennes !

PR #4319, opened 2021-01-26
[Part] - Fix memory leaks

vosk:
A valgrid session revealled a few easy fixes, the most important with respect to non-obviousness  is in `src/Mod/Part/App/PropertyGeometryList.cpp`

- [ ] Commit message is [well-written](https://chris.beams.io/posts/git-commit/)

---

### DIFF
--- a/src/Gui/DlgSettingsColorGradientImp.cpp
+++ b/src/Gui/DlgSettingsColorGradientImp.cpp
@@ -61,6 +61,9 @@ DlgSettingsColorGradientImp::DlgSettingsColorGradientImp( QWidget* parent, Qt::W
 DlgSettingsColorGradientImp::~DlgSettingsColorGradientImp()
 {
     // no need to delete child widgets, Qt does it all for us
+    // Validators are not owned by the widgets
+    delete fMaxVal;
+    delete fMinVal;
 }
 
 void DlgSettingsColorGradientImp::setColorModel( App::ColorGradient::TColorModel tModel)
--- a/src/Gui/View3DInventor.cpp
+++ b/src/Gui/View3DInventor.cpp
@@ -773,8 +773,10 @@ bool View3DInventor::setCamera(const char* pCamera)
 
     SoNode * Cam;
     SoDB::read(&in,Cam);
+    Cam->ref();
 
     if (!Cam){
+        Cam->unref();
         throw Base::RuntimeError("Camera settings failed to read");
     }
 
@@ -802,6 +804,7 @@ bool View3DInventor::setCamera(const char* pCamera)
             CamViewerP->farDistance   = ((SoPerspectiveCamera *)Cam)->farDistance;
             CamViewerP->focalDistance = ((SoPerspectiveCamera *)Cam)->focalDistance;
         } else {
+            Cam->unref();
             throw Base::TypeError("Camera type mismatch");
         }
     } else if (Cam->getTypeId() == SoOrthographicCamera::getClassTypeId()) {
@@ -815,10 +818,11 @@ bool View3DInventor::setCamera(const char* pCamera)
             CamViewerO->aspectRatio      = ((SoOrthographicCamera *)Cam)->aspectRatio ;
             CamViewerO->height           = ((SoOrthographicCamera *)Cam)->height;
         } else {
+            Cam->unref();
             throw Base::TypeError("Camera type mismatch");
         }
     }
-
+    Cam->unref();
     return true;
 }
 
--- a/src/Gui/ViewProviderLine.cpp
+++ b/src/Gui/ViewProviderLine.cpp
@@ -68,13 +68,11 @@ void ViewProviderLine::attach ( App::DocumentObject *obj ) {
     sep->addChild ( pCoords );
 
     SoIndexedLineSet *pLines  = new SoIndexedLineSet ();
-    pLines->ref();
     pLines->coordIndex.setNum(3);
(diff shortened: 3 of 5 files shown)


chennes:
I see this is marked as "Draft" -- are you continuing to track down more leaks, or is it appropriate for testing as it is?

vosk:
I am tracking a stubborn in `Gui::View3DInventor::setCamera` and I will update accordingly!

vosk:
Fixed! Coin3D objects must be first ref() before unref() to delete!

chennes:
May as well just keep adding to this PR. I gave it a run and it didn't give me any problems, but it's easy enough to check out anything new you commit to it up until a dev merges it.

luzpaz:
@vosk would you be interested in in creating a tutorial on how to use valgrind to find issues in FreeCAD?

vosk:
> @vosk would you be interested in in creating a tutorial on how to use valgrind to find issues in FreeCAD?

Sure, what do you have in mind?

luzpaz:
@vosk lets discuss in the forum ? Here's an example of an older post on valgrind usage on FreeCAD https://forum.freecadweb.org/viewtopic.php?style=4&t=14343

src/Gui/DlgSettingsColorGradientImp.cpp
```diff
@@ -60,7 +60,10 @@ DlgSettingsColorGradientImp::DlgSettingsColorGradientImp( QWidget* parent, Qt::W
  */
 DlgSettingsColorGradientImp::~DlgSettingsColorGradientImp()
 {
-    // no need to delete child widgets, Qt does it all for us
+    // no need to delete child widgets, Qt does it all for us
+    // Validators are not owned by the widgets
+    delete fMaxVal;
+    delete fMinVal;
```
wwmayer, line 66: Explicitly deleting the validators here do no harm but is superfluous because contrary to your comment they are owned by the dialog:
```
    fMaxVal = new QDoubleValidator(-1000,1000,ui->spinBoxDecimals->maximum(),this);
    ui->floatLineEditMax->setValidator(fMaxVal);
    fMinVal = new QDoubleValidator(-1000,1000,ui->spinBoxDecimals->maximum(),this);
    ui->floatLineEditMin->setValidator(fMinVal);
```

The third argument in the constructor is the parent. So, when the parent (i.e. this dialog) is deleted it should also delete its children.

vosk: You are absolutely right on that one. I missed the parent That was a good start  for me :D :D

src/Gui/View3DInventor.cpp
```diff
@@ -815,10 +818,11 @@ bool View3DInventor::setCamera(const char* pCamera)
             CamViewerO->aspectRatio      = ((SoOrthographicCamera *)Cam)->aspectRatio ;
             CamViewerO->height           = ((SoOrthographicCamera *)Cam)->height;
         } else {
+            Cam->unref();
             throw Base::TypeError("Camera type mismatch");
         }
     }
-
+    Cam->unref();
```
wwmayer, line 825: It's known to be not good practice to free memory in all possible exit points of a function. A smarter way is to rely on C++'s RAII idiom which has two advantages:
* it keeps the code simple
* it even works if an unexpected exception is thrown

vosk: I know that! But is there a scope lock/ handle/ smart ptr that can wrap an SoNode?

src/Gui/DlgSettingsColorGradientImp.cpp
wwmayer: > It technically DOES hurt, in theory this may lead to releasing something irrelavant.

No, it (usually) doesn't hurt. When a sub-object is explicitly deleted before its parent then it gets a notification be Qt and makes sure to avoid a double destruction.

Anyway, when merging this PR I will omit this change...

src/Gui/View3DInventor.cpp
vosk: For posterity:
CoinPtr works because Coin3D has already defined the intrusive_ptr methods in SoBase.h
```
// support for boost::intrusive_ptr<SoBase>
inline void intrusive_ptr_add_ref(SoBase * obj) { obj->ref(); }
inline void intrusive_ptr_release(SoBase * obj) { obj->unref(); }
```

wwmayer: As said the CoinPtr class may do that. I am testing it right now...

Btw, I just realized two bugs in your code:
You call `Cam->ref();` directly after using the `SoDB::read` function and at the point Cam still could be a null pointer.
Then in the next line there is a check `if (!Cam){` and you call `Cam->unref();` which will cause a crash.

vosk: Would you prefer for me to push again without (hopefully ) the stupidity and THEN delete my github account or maybe just delete my account, and you push directly to master?

wwmayer: See https://github.com/FreeCAD/FreeCAD/commit/b49c523a541ce7ace520463e9e0b695ef6b20e5b

> delete my github account

You don't have to delete your account. Deleting the branch is enough :)

wwmayer:
Merged + fixed some issues

luzpaz:
Relevant commits:  
4bd5c2669d006254cdfc93e20c2e265226dfec5d  
b49c523a541ce7ace520463e9e0b695ef6b20e5b

PR #4323, opened 2021-01-27
[Gui] Make inventor markers symmetric

hyarion:
These changes makes the round markers symmetric horizontally, vertically and on the diagonal.

The changes to the round markers, that are used in sketcher, are shown in the screenshot below.
image

The circle versions has similar changes applied.

This forum post highlighted, unintentionally the issue: https://forum.freecadweb.org/viewtopic.php?p=470763#p470763

---

- ~Commit message includes `issue #<id>` or `fixes #<id>` where `<id>` is the [associated MantisBT](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) issue id if one exists~

---

### DIFF
--- a/src/Gui/Inventor/MarkerBitmaps.cpp
+++ b/src/Gui/Inventor/MarkerBitmaps.cpp
@@ -345,12 +345,12 @@ const char circleLine11_marker[CIRCLE_LINE11_WIDTH * CIRCLE_LINE11_HEIGHT + 1] =
 "           "
 "   xxxxxx  "
 "  xxxxxxxx "
-"  xx    xx "
+" xxx    xxx"
 " xx      xx"
 " xx      xx"
 " xx      xx"
 " xx      xx"
-"  xx    xx "
+" xxx    xxx"
 "  xxxxxxxx "
 "   xxxxxx  "};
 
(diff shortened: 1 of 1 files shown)


chennes (approved):
Looks good (literally). Thanks! I hope we can get these as vectors someday (if only so we don't have to have that specific list of sizes), but this is a worthwhile improvement in the meantime.

luzpaz:
>  I hope we can get these as vectors someday

I wonder if this is on @Vanuan's HiDPi Roadmap ?

Vanuan:
@luzpaz Not particularly. It goes under general clause "Make all pixel graphics vector or available at various pixel density".

We can fix markers in various ways:
- providing more sizes for them so that we have more detail and they would be "crisper" and "rounder" on HiDPI display.
- "blurring" markers, thus providing sort of anti-aliasing
- add alpha channel and various shades of red to xpms
- converting markers to SVG and rendering on demand

So it needs both higher resolution and smooth edges (AA). Anti-aliasing is not included in HiDPI roadmap, I suppose, as it's a different concern.

hyarion:
@Vanuan
I don't think AA is an HiDPI issue. The improvement with AA is much more visible on low DPI devices. (I can't even see any jagged edges on markers on my HiDPI ~2016 macbook).

> * providing more sizes for them so that we have more detail and they would be "crisper" and "rounder" on HiDPI display.

This would be the fastest route

> * "blurring" markers, thus providing sort of anti-aliasing: add alpha channel and various shades of red to xpms

Alpha channel is only supported with indexed markers and index channels doesn't use the color information from materials.
Maybe we could add support for alpha markers (SoAlphaMarkerSet?) in upstream coin3d instead?

> * converting markers to SVG and rendering on demand

I don't think we would need to render vector graphics on demand. Instead we could render just the markers sizes we need to the coin marker system. That way we could use the current pipeline we use today.

PR #4201, opened 2020-12-27
Path: fix unnecessary copy on read for Command.parameters dict

J-Dunn:
Contents of underlying c++ std::map data is copied to a new PyDict on every read. This is contrary to expected python behaviour which would normally just return the pointer to a python object and increment the reference counter for it.  This was leading to massive redundant deletion and copying in all post processors which reference this variable in a nested loop. This PR adds a permanent dict member to the class and keeps track of changes to avoid unnecessary copying.

https://forum.freecadweb.org/viewtopic.php?f=15&t=50583&start=30#p457914

Thank you for creating a pull request to contribute to FreeCAD! To ease integration, please confirm the following:

- [x] Branch rebased on latest master `git pull --rebase upstream master`
- [ ] Unit tests confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [x] Commit message is [well-written](https://chris.beams.io/posts/git-commit/)
- [x] Commit message includes `issue #<id>` or `fixes #<id>` where `<id>` is the [associated MantisBT](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) issue id if one exists

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.19 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=34586).

---

PR #4329, opened 2021-01-28
[TD]+[APP] - memleaks

vosk:
A couple of python leaks plus one in DrawViewPart that shows up on --run-test 
- [ ] Commit message is [well-written](https://chris.beams.io/posts/git-commit/)

---

### DIFF
--- a/src/App/PropertyContainerPyImp.cpp
+++ b/src/App/PropertyContainerPyImp.cpp
@@ -522,11 +522,19 @@ PyObject *PropertyContainerPy::getCustomAttributes(const char* attr) const
         PyObject *dict = PyDict_New();
         if (dict) {
             for ( std::map<std::string,App::Property*>::iterator it = Map.begin(); it != Map.end(); ++it )
+            {
 #if PY_MAJOR_VERSION >= 3
-                PyDict_SetItem(dict, PyUnicode_FromString(it->first.c_str()), PyUnicode_FromString(""));
+                PyObject * key = PyUnicode_FromString(it->first.c_str());
+                PyObject * emptyStr = PyUnicode_FromString("");
+                PyDict_SetItem(dict, key, emptyStr);
 #else
-                PyDict_SetItem(dict, PyString_FromString(it->first.c_str()), PyString_FromString(""));
+                PyObject * key = PyString_FromString(it->first.c_str());
+                PyObject * emptyStr = PyString_FromString("");
+                PyDict_SetItem(dict, key, emptyStr);
 #endif
+                Py_DECREF(key);
+                Py_DECREF(emptyStr);
+            }
             if (PyErr_Occurred()) {
                 Py_DECREF(dict);
                 dict = NULL;
--- a/src/App/PropertyStandard.cpp
+++ b/src/App/PropertyStandard.cpp
@@ -1921,6 +1921,7 @@ PyObject *PropertyMap::getPyObject(void)
             throw Base::UnicodeError("UTF8 conversion failure at PropertyMap::getPyObject()");
         }
         PyDict_SetItemString(dict,it->first.c_str(),item);
+        Py_DECREF(item);
     }
 
     return dict;
--- a/src/Gui/CommandWindow.cpp
+++ b/src/Gui/CommandWindow.cpp
@@ -370,7 +370,7 @@ class FilterStatusBar : public QObject
 //    Q_OBJECT
 
 public:
-    FilterStatusBar(Action * action):QObject() {this->action = action;}
+    FilterStatusBar(Action * action):QObject(action) {this->action = action;}
 //    virtual ~FilterStatusBar() {}
 protected:
     Action * action;
--- a/src/Gui/Quarter/QuarterWidget.cpp
+++ b/src/Gui/Quarter/QuarterWidget.cpp
@@ -289,7 +289,7 @@ QuarterWidget::QuarterWidget(QtGLContext * context, QWidget * parent, const QtGL
 void
 QuarterWidget::constructor(const QtGLFormat & format, const QtGLWidget * sharewidget)
 {
-  QGraphicsScene* scene = new QGraphicsScene;
+  QGraphicsScene* scene = new QGraphicsScene(this);
   setScene(scene);
   setViewport(new CustomGLWidget(format, this, sharewidget)); 
   
--- a/src/Gui/SoFCColorBar.cpp
+++ b/src/Gui/SoFCColorBar.cpp
@@ -42,7 +42,7 @@ SO_NODE_ABSTRACT_SOURCE(SoFCColorBarBase)
 /*!
   Constructor.
 */
-SoFCColorBarBase::SoFCColorBarBase()
+SoFCColorBarBase::SoFCColorBarBase(): _windowSize(0,0)
 {
     SO_NODE_CONSTRUCTOR(SoFCColorBarBase);
 }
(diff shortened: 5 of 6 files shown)


src/App/PropertyContainerPyImp.cpp
```diff
@@ -521,12 +521,20 @@ PyObject *PropertyContainerPy::getCustomAttributes(const char* attr) const
...
-            for ( std::map<std::string,App::Property*>::iterator it = Map.begin(); it != Map.end(); ++it )
-#if PY_MAJOR_VERSION >= 3
-                PyDict_SetItem(dict, PyUnicode_FromString(it->first.c_str()), PyUnicode_FromString(""));
-#else
-                PyDict_SetItem(dict, PyString_FromString(it->first.c_str()), PyString_FromString(""));
-#endif
+            for ( std::map<std::string,App::Property*>::iterator it = Map.begin(); it != Map.end(); ++it )
+            {
+#if PY_MAJOR_VERSION >= 3
+                PyObject * key = PyUnicode_FromString(it->first.c_str());
```
vosk, line 527: PyUnicode_FromString returns a new reference and PyDict_SetItem does not steal reference

src/App/PropertyStandard.cpp
```diff
@@ -1920,7 +1920,8 @@ PyObject *PropertyMap::getPyObject(void)
             Py_DECREF(dict);
             throw Base::UnicodeError("UTF8 conversion failure at PropertyMap::getPyObject()");
         }
-        PyDict_SetItemString(dict,it->first.c_str(),item);
+        PyDict_SetItemString(dict,it->first.c_str(),item);
+        Py_DECREF(item);
```
vosk, line 1924: PyUnicode_DecodeUTF8 returns new reference a couple of lines above

src/Gui/CommandWindow.cpp
```diff
@@ -370,7 +370,7 @@ class FilterStatusBar : public QObject
 //    Q_OBJECT
 
 public:
-    FilterStatusBar(Action * action):QObject() {this->action = action;}
+    FilterStatusBar(Action * action):QObject(action) {this->action = action;}
```
vosk, line 373: FilterStatusBar is leaking, I am not sure about this one, but attaching it to the action attaches to the MainWindow and destroys succesfully

src/Gui/SoFCColorBar.cpp
```diff
@@ -42,7 +42,7 @@ SO_NODE_ABSTRACT_SOURCE(SoFCColorBarBase)
 /*!
   Constructor.
 */
-SoFCColorBarBase::SoFCColorBarBase()
+SoFCColorBarBase::SoFCColorBarBase(): _windowSize(0,0)
```
vosk, line 45: valgrind complains on uninitialized comparisson

src/Mod/TechDraw/App/DrawViewPart.cpp
```diff
@@ -348,6 +348,8 @@ void DrawViewPart::onChanged(const App::Property* prop)
 void DrawViewPart::partExec(TopoDS_Shape shape)
 {
 //    Base::Console().Message("DVP::partExec()\n");
+    if(geometryObject)
```
vosk, line 351: This I have no idea if it is the correct fix, but partExec leaks an object on every run

src/Gui/Quarter/QuarterWidget.cpp
```diff
@@ -289,7 +289,7 @@ QuarterWidget::QuarterWidget(QtGLContext * context, QWidget * parent, const QtGL
 void
 QuarterWidget::constructor(const QtGLFormat & format, const QtGLWidget * sharewidget)
 {
-  QGraphicsScene* scene = new QGraphicsScene;
+  QGraphicsScene* scene = new QGraphicsScene(this);
```
vosk, line 292: https://doc.qt.io/qt-5/qgraphicsview.html#setScene
The view does not take ownership of scene.

src/Mod/TechDraw/App/DrawViewPart.cpp
aapo-aapo: Seems to me it's correct.  Moreover, partExec() is only called from lines 291-308 in the same file, and originally there has been a similar delete-if-not-nullptr functionality.  But, as seen from the code on those lines, after it has been touched, the author has forgotten to add the deletion logic to the other call.  I think it avoids mistakes to have the deletion logic inside the function like you propose.  Note, that lines 301-305 of DrawViewPart.cpp should be removed (they will be superfluous as they have the same functionality as your proposed addition)

aapo-aapo (approved):

src/Gui/Quarter/QuarterWidget.cpp
aapo-aapo: I Agree, it should

src/App/PropertyContainerPyImp.cpp
aapo-aapo: Should it be Py_XDECREF() instead, as PyUnicode_FromString() could in principle return NULL?  Otherwise, it may segfault when calling Py_DECREF() with nullptr.

J-Dunn:
Can I suggest you look at PR which got merged to Path recently and associated forum thread.

https://github.com/FreeCAD/FreeCAD/pull/4201

wwmayer suggested in the thread that it should not be necessary to do INCing and DECing if we remain inside the CCX code wrappers. I have not analysed this PR but maybe you could consider if that is pertinent. 

That PR itself was concerned with very wasteful copy on read when using Pdict. This was causing _heavy speed penalties_ in Path where a read reference to a Pdict ( which was unchanging in the context ) was made inside a loop. 

Python programming would expect such a read reference to simply return the pointer and inc the python refcount. However, Pdict destroys the existing Pdict, creates a new one and copies all the same content back in tuple by tuple. 

To avoid this I made Pdict member in the class and kept track of whether it had been written to. Thus reads can just return a pointer where appropriate in a typical pythonist way. 

This seems to be necessary evil in trying to hack together code in two languages with such different memory allocation strategies.  Pdict gives python access to the underlying C++ data structures but it can be very inefficient due to language differences. 

It may be worth looking at whether this gives a performance hit in this context. 

It probably will not matter on trivial test shapes but in some kind of complex, multi-faceted shapes ( perhaps derived by importing other formats into FreeCAD ) it could cause significant slowing. 

The issue is more general than Pdict but this is one instance where I can point you to an issue and a solution.

luzpaz:
@vosk perhaps prepend [0.19] to this PR ?

vosk:
> @vosk perhaps prepend [0.19] to this PR ?

If it were up to me, sure!

vosk:
I would be happy to rip through everything I touched and use PyCXX instead.

src/App/PropertyContainerPyImp.cpp
vosk: While you are correct in GENERAL, the property map should not have any nulls in it, AFAIK.

src/Mod/TechDraw/App/DrawViewPart.cpp
wwmayer: See 542bf23cb0

src/Gui/Quarter/QuarterWidget.cpp
wwmayer: 17ec9b7c08708

src/Gui/CommandWindow.cpp
wwmayer: c4e383829b0

src/App/PropertyContainerPyImp.cpp
wwmayer: A much better way is to use the PyCXX API here as it:
* automatically handles the ref counting
* raises an exception on errors
* simplifies the code a lot

wwmayer: See 9f72a7b4061 and c331d608aa6

wwmayer:
Merged-

PR #4339, opened 2021-01-31
[0.20] [Path] Add simple test framework for Slot op

Russ4262:

---

### DIFF
--- a/src/Mod/Path/PathTests/TestPathSlot.py
+++ b/src/Mod/Path/PathTests/TestPathSlot.py
@@ -31,13 +31,17 @@ from PathTests.PathTestUtils import PathTestBase
 class TestPathSlot(PathTestBase):
     """Unit tests for the Slot operation."""
 
-    @staticmethod
-    def _setup_test_environment():
-        """_setup_test_environment()...
-        This method is upon instantiation of this test class.  Add code and objects here
-        that are needed for the duration of the test() methods in this class.
-        This method does not have access to the class `self` reference.
+    @classmethod
+    def setUpClass(cls):
+        """setUpClass()...
+        This method is called upon instantiation of this test class.  Add code and objects here
+        that are needed for the duration of the test() methods in this class.  In other words,
+        set up the 'global' test environment here; use the `setUp()` method to set up a 'local'
+        test environment. 
+        This method does not have access to the class `self` reference, but it
+        is able to call static methods within this same class.
         """
+
         # doc = FreeCAD.open(FreeCAD.getHomePath() + 'Mod/Path/PathTests/test_slot_01.fcstd')
         doc_title = "TestSlot"
         doc = FreeCAD.newDocument(doc_title)
@@ -56,22 +60,12 @@ class TestPathSlot(PathTestBase):
         cut = doc.addObject('Part::Cut', 'Cut')
         cut.Base = box0
         cut.Tool = box1
+        cut.Placement = FreeCAD.Placement(FreeCAD.Vector(10.0, 10.0, 0.0), FreeCAD.Rotation(FreeCAD.Vector(0,0,1), 0))
         doc.recompute()
 
         # Create Job object
         PathJob.Create('Job', [doc.Cut], None)
         doc.recompute()
-
-    @classmethod
-    def setUpClass(cls):
-        """setUpClass()...
-        This method is called upon instantiation of this test class.  Add code and objects here
-        that are needed for the duration of the test() methods in this class.
-        This method does not have access to the class `self` reference.  This method
-        is able to call static methods within this same class.
-        """
-        # FreeCAD.Console.PrintMessage("TestPathSlot.setUpClass()\n")
-        cls._setup_test_environment()
 
     @classmethod
     def tearDownClass(cls):
@@ -81,8 +75,9 @@ class TestPathSlot(PathTestBase):
         This method does not have access to the class `self` reference.  This method
         is able to call static methods within this same class.
         """
-        # FreeCAD.Console.PrintMessage("TestPathSlot.tearDownClass()\n")
+        # Comment out to leave test file open and objects and paths intact after all tests finish
         FreeCAD.closeDocument(FreeCAD.ActiveDocument.Name)
+        pass
 
     def setUp(self):
         """setUp()...
@@ -101,22 +96,45 @@ class TestPathSlot(PathTestBase):
 
     def _set_depths_and_heights(self, op):
         # Set Depths
-        # set start and final depth in order to eliminate effects of stock (and its default values)
+        # Set start and final depth in order to eliminate effects of stock (and its default values)
         op.setExpression('StartDepth', None)
         op.StartDepth.Value = 10.0
         op.setExpression('FinalDepth', None)
         op.FinalDepth.Value = 5.0
+
+        # Set step down so as to only produce one layer path
         op.setExpression('StepDown', None)
         op.StepDown.Value = 5.0
 
         # Set Heights
         # default values used
+        pass
 
-    def _format_point(self, cmd):
+    def _format_point(self, cmd, z=0.0):
         """Accepts command dict and returns point string coordinate"""
         x = round(cmd["X"], 2)
         y = round(cmd["Y"], 2)
-        return "({}, {})".format(x, y)
+        if z:
+            z = round(z, 2)
+        return "({}, {}, {})".format(x, y, z)
+
+    def _find_slot_points(self, slot):
+        pnts = list()
+        pnt_params = list()
+        z = -1.0
+        for c in slot.Path.Commands:
+            p = c.Parameters
+            if p.get("Z"):
+                if len(pnt_params) > 1:
+                    break
+                else:
+                    z = p.get("Z")
+            elif p.get("X") and p.get("Y"):
+                pnt_params.append(p)
+        for p in pnt_params:
+            pnts.append(self._format_point(p, z))
+        pnts.sort()
+        return pnts
 
     def test00(self):
         '''Test horizontal rectangular face: ReverseDirection=True; LayerMode=Single-pass'''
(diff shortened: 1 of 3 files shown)


Russ4262:
I am not sure why gcc 7 Travis Slot op test failed and the other two succeeded.  Rebased to rerun the tests.

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,264 @@
...
+import PathScripts.PathJob as PathJob
+import PathScripts.PathSlot as PathSlot
+from PathTests.PathTestUtils import PathTestBase
+
+
+class TestPathSlot(PathTestBase):
+    """Unit tests for the Slot operation."""
+
+    @staticmethod
+    def _setup_test_environment():
```
mlampert, line 34: please don't use `_` names for classes or functions, camel case is the way to go

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,264 @@
...
+        slot.LayerMode = "Single-pass"
+
+        slot.recompute()
+        self.doc.recompute()
+        
+        pnts = list()
+        for c in slot.Path.Commands:
+            p = c.Parameters
+            keys = str(p.keys())[9:]
+            if "'X'" in keys and "'Y'" in keys:
```
mlampert, line 150: why not just use `c.Parameters.get('X') and c.Parameter.get('Y')` ?

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,264 @@
...
+        if len(pnts) != 2:
+            for c in slot.Path.Commands:
+                p = c.Parameters
+                print("Parameters: {}".format(p))
+            self.assertEqual(len(pnts), 2)
+
+        # Verify each line-segment point, excluding arcs
+        verify_points = ['(7.5, 0.0)', '(7.5, 15.0)']
+        for i in range(2):
+            self.assertEqual(pnts[i], verify_points[i])
```
mlampert, line 165: I highly recommend not comparing doubles directly but use PathGeom to make the test more robust.

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,264 @@
...
+    @classmethod
+    def setUpClass(cls):
+        """setUpClass()...
+        This method is called upon instantiation of this test class.  Add code and objects here
+        that are needed for the duration of the test() methods in this class.
+        This method does not have access to the class `self` reference.  This method
+        is able to call static methods within this same class.
+        """
+        # FreeCAD.Console.PrintMessage("TestPathSlot.setUpClass()\n")
+        cls._setup_test_environment()
```
mlampert, line 73: why this layer of indirection?

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,264 @@
...
+        box1.Placement = FreeCAD.Placement(FreeCAD.Vector(5.0, 0.0, 5.0), FreeCAD.Rotation(FreeCAD.Vector(0,0,1), 0))
+        doc.recompute()
+        cut = doc.addObject('Part::Cut', 'Cut')
+        cut.Base = box0
+        cut.Tool = box1
+        doc.recompute()
+
+        # Create Job object
+        PathJob.Create('Job', [doc.Cut], None)
+        doc.recompute()
```
mlampert, line 62: It would be simpler to just add a test file and load that. Would also make the tests simpler since you don't have to search for your feature each time.

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,264 @@
...
+    def _format_point(self, cmd):
+        """Accepts command dict and returns point string coordinate"""
+        x = round(cmd["X"], 2)
+        y = round(cmd["Y"], 2)
+        return "({}, {})".format(x, y)
+
+    def test00(self):
+        '''Test Slot on horizontal rectangular face: ReverseDirection=True; LayerMode=Single-pass'''
+
+        # Identify base feature(s) to be used for operation's Base Geometry
```
mlampert, line 123: the important question is _which_ feature you're looking for ;)

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,264 @@
...
+        # default values used
+
+    def _format_point(self, cmd):
+        """Accepts command dict and returns point string coordinate"""
+        x = round(cmd["X"], 2)
+        y = round(cmd["Y"], 2)
+        return "({}, {})".format(x, y)
+
+    def test00(self):
+        '''Test Slot on horizontal rectangular face: ReverseDirection=True; LayerMode=Single-pass'''
```
mlampert, line 121: It is good practice to include the expected outcome of a test - the description should tell me what the test intends to verify.

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,264 @@
...
+        # Set Operation
+        slot.ReverseDirection = True
+        slot.LayerMode = "Single-pass"
+
+        slot.recompute()
+        self.doc.recompute()
+        
+        pnts = list()
+        for c in slot.Path.Commands:
+            p = c.Parameters
```
mlampert, line 148: what about the `Command` ?
It seems unnecessarily roundabout testing - you know the start point and the end point, so there should be a
* `G0` to safe height
* `G0` to starting point
* `G1` in z down to operation depth
* `G1` to end point
* `G1` (or `G0`) up to safe height
There should be no `G2` or `G3` commands and every other command can be ignored, like comments. As a reader I would immediately see what is being tested.

mlampert (changes requested):

src/Mod/Path/PathTests/TestPathSlot.py
```diff
@@ -0,0 +1,263 @@
...
+        '''
+
+        # Identify base feature to be used for operation's Base Geometry
+        base = self.doc.Cut
+        edge_list = list()
+        for i in range(0, len(base.Shape.Edges)):
+            edge = "Edge%d" % (i+1)
+            e = base.Shape.getElement(edge)
+            eBB = e.BoundBox
+            if (e.Length == 15.0 and round(eBB.ZLength, 6) == 0.0 and round(eBB.ZMin, 6) == 5.0):
```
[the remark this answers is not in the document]
Russ4262: Are you suggesting I convert the current `if...` statement to `PathGeom.isRoughly(e.Length, 15.0) and PathGeom.isRoughly(eBB.ZLength, 0.0) and PathGeom.isRoughly(eBB.ZMin, 5.0)` ?

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: I want to only work with commands that contain both `X` and `Y` parameters.  I am doing some code scenario testing to see if I understand what you are suggesting here.  I will try to respond with my findings.

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: In this instance, the `self.assertEqual()` is comparing strings as I have seen done elsewhere in Path unit tests.  Should I compare strings differently?

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: I will remove this and combine the methods and add additional comments to explain the purpose of the code.  The one-liner seen here is a nice clear indication of what is being done - that was my thinking, as well as educating those who follow after and want to learn from this unit test.

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: I have not found a good answer for this one due to the challenges your suggestion introduces.  Yes, I do have a test file started and would like to use it to simplify things a bit.  The problem I run into is that when I modify the geometry of the test file I am working on, the edge and face numbering are disrupted, which in turn forces me to have to resynchronize many of the edge and face names used in `test##()` methods.  
. 
So, for now, I am preferring to search for each feature until I have a test file that will not need much maintenance or will not re-enumerate the edge and face names with future additions.

For example:
Snip macro screenshot-149768
This test file is coming along nicely, but when I make additions, the face and edge numbers get disrupted, causing the need to go back into the test module and edit edge and face name references.

My goal is to use a test file.  For now, I would like to get a basic unit test created and functional and merged so we have a starting point and some help with testing changes to the operation.

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: Okay.  Thanks.  I will edit the `test##()` method docStrings throughout all the unit tests I am creating to better communicate the purpose of the test and desired outcome.

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: I don't think I understand what change you are suggesting.  
.
My current strategy is to search all the commands for the two points that should exist within the command set.  
.
Having revisited this Slot op now that I found the issue with the Linux gcc-7 build, I know I need to include depth verification in all the unit tests I am creating.  I will have to figure out how to incorporate such verification across the board in all of them.  I think I have an idea how to do it.

src/Mod/Path/PathTests/TestPathSlot.py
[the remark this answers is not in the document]
mlampert: that would be my recommendation, yes.

src/Mod/Path/PathTests/TestPathSlot.py
mlampert: you're converting all parameters into a string and then check for 'X' and 'Y'.. What I meant is that `Parameters` already is a dictionary, so you can checkk if the keys exist directly.

src/Mod/Path/PathTests/TestPathSlot.py
mlampert: you're converting a float to a string - could you end up with something like `'(7.499999999999, 0.00000000001)'` ?

src/Mod/Path/PathTests/TestPathSlot.py
mlampert: wow - this is not a  test file, it's a torture file ;)

maybe use multiple files, it's easier to verify multiple simple shapes than one big one which includes everything. sliptonic has a new internet connection, so we're allowed to use more test files ;)

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: Okay.  Thanks.  I am simplifying all these and similar checks to the style below, which is more in line with what you are suggesting (I think):
`                    p = c.Parameters
                    if p.get("X") and p.get("Y"):
                        # Do Action
`

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: > you're converting a float to a string - could you end up with something like `'(7.499999999999, 0.00000000001)'` ?

I don't think so.  I am rounding values to say three or six decimal places, prior to comparison, to avoid such scenarios.  So long as the `round()` function is working as I am expecting, the subsequent string representation should be only, say, two decimal places max per my rounding argument used.

Does this response address your concern with my methodology, or am I still missing something in the process?

src/Mod/Path/PathTests/TestPathSlot.py
Russ4262: >... maybe use multiple files, it's easier to verify multiple simple shapes than one big one which includes everything. ...

My thinking is to use one file, but it will contain multiple bodies to be used as bases for Jobs in this Slot unit test.  This way I think I could add another body with more targeted slots related to the attached image above without having to alter that particular body. 
.
Does this single-file, multi-body strategy seem okay?

PR #4340, opened 2021-01-31
PartDesign: Select sketching plane by double click in picker dialog

davidosterberg:
Small UI improvement. Allow the user to pick the sketch plane by double clicking instead of click to select followed by click on OK. 

Forum discussion:
https://forum.freecadweb.org/viewtopic.php?f=8&t=54876#p471829

Thanks to @0penBrain for solving the only real difficulty with this PR. 

double_click_plane_pick

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskFeaturePick.cpp
+++ b/src/Mod/PartDesign/Gui/TaskFeaturePick.cpp
@@ -26,6 +26,7 @@
 #ifndef _PreComp_
 # include <QListIterator>
 # include <QTimer>
+# include <QListWidgetItem>
 #endif
 
 #include <Gui/Application.h>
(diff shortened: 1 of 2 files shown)


chennes (approved):
Nice addition: small, but definitely fixes a small annoyance, and small annoyances can add up! I like that bit of code from @OpenBrain to use the `Qt::QueuedConnection`.

0penBrain:
@davidosterberg : maybe the 2 commits could be squashed to have the master history looking good. ;)
Anyway, as @chennes I like the baby step improvements :+1:

abdullahtahiriyo:
@davidosterberg 

Would you squash the two commits into one? 

If I press "squash and merge", I think that you might lose the ownership metainformation, it being assigned to me. I do not think that is right. It is your contribution. 

When squashing you will have the opportunity to edit the commit message. Then, maybe the new commit message could have an indication that it is for PartDesign. That should help those looking at the commit log.

davidosterberg:
> Would you squash the two commits into one?

@abdullahtahiriyo , squashed and rebased.

PR #4343, opened 2021-01-31
[PD] hole dialog UI fixes

donovaly:
This PR is the first in a series of 3 PRs to fix known hole dialog bugs. (PR #4344 and PR #4347 depend on this PR)

This one fixes:

- the .ui file issues
- readonly status issues of some widgets/properties (e.g. when the hole is through all, disable drill point settings)
- the bug that updateHoleCutParams() overwrote the previously correctly determined hole diameter
- the missing check that the counterbore/sink diameter must not be <= than the hole diameter
- just a trifle: change a function name to fit into the naming scheme

Here is the changed dialog as it looks for me:
FreeCAD_lcZwDdKIjP

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHole.cpp
+++ b/src/Mod/PartDesign/App/FeatureHole.cpp
@@ -424,7 +424,9 @@ const double Hole::metricHoleDiameters[36][4] =
         { 6.0,      6.4,    6.6,    7.0},
         { 7.0,      7.4,    7.6,    8.0},
         { 8.0,      8.4,    9.0,    10.0},
+        // 9.0 undefined
         { 10.0,     10.5,   11.0,   12.0},
+        // 11.0 undefined
         { 12.0,     13.0,   13.5,   14.5},
         { 14.0,     15.0,   15.5,   16.5},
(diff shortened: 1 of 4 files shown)


src/Mod/PartDesign/App/FeatureHole.cpp
```diff
@@ -1009,11 +1033,16 @@ void Hole::onChanged(const App::Property *prop)
     }
     else if (prop == &ThreadSize) {
         updateDiameterParam();
-        updateHoleCutParams();
+        // updateHoleCutParams() will automatcially be called because updateDiameterParam() changes &Diameter
```
chennes, line 1036:
```suggestion
        // updateHoleCutParams() will automatically be called because updateDiameterParam() changes &Diameter
```

chennes (approved):
It feels like this has been a long road! I re-tested the items in the PR description and they are all resolved. The code looks good, it compiles, runs, and I did not observe any regressions in my testing. I think it's good to merge.

src/Mod/PartDesign/App/FeatureHole.cpp
donovaly: Thanks, fixed now

PR #4344, opened 2021-01-31
[PD] fix UTS clearance holes

donovaly:
This PR is based on PR #4343. It fixes:

- the bug that we did not use the normed clearance hole diameters for UTS holes
- missing recompute when changing existing hole from Metric to UTS type

Only review the commit c6f5d2f, the other commit are just the changes from PR #4343 .

This PR is meant to be merged after PR #4343 was merged.


chennes (approved):
This compiles, runs, and makes the changes outlined in the PR description. I have confirmed that the values for the UTS clearance holes match the standard as described in the comment.

donovaly:
> This compiles, runs, and makes the changes outlined in the PR description.

Many thanks for the review.

I updated the PR according to PR #4343 so it can directly be merged after #4343 .

wwmayer:
Merged.

See c676c07316

PR #4347, opened 2021-02-01
[PD] hole fix custom cut handling 

donovaly:
This PR is based on PR #4344. It:

handles custom counterbore/sink values correctly
See bug B reported here: https://forum.freecadweb.org/viewtopic.php?f=3&t=54408&sid=91194a40335be7962781e0aa7958a760
We have normed screw head cuts. These values can be overridden but we must store the info about the overriding. Why? - because when you have e.g. made a custom change to a normed countersink and then change to another countersink norm, you would either not get the values defined in the norm or you get these values but loose e.g. your depth settings

Only review the commit cad4d82, the other commits are just the changes from PR #4343 and PR #4344 .

This PR is meant to be merged after PR #4344 was merged.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHole.cpp
+++ b/src/Mod/PartDesign/App/FeatureHole.cpp
@@ -71,7 +71,8 @@ namespace PartDesign {
 
 const char* Hole::DepthTypeEnums[]                   = { "Dimension", "ThroughAll", /*, "UpToFirst", */ NULL };
 const char* Hole::ThreadTypeEnums[]                  = { "None", "ISOMetricProfile", "ISOMetricFineProfile", "UNC", "UNF", "UNEF", NULL};
-const char* Hole::ThreadFitEnums[]                   = { "Standard", "Close", "Wide", NULL};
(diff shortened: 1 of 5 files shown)


chennes:
I think there is still a small issue with the enabling/disabling of the custom values checkbox. Here is the procedure I just tested:

1. Set up a norm-conforming hole using ISO metric regular, and M2 screw
2. For Hole cut choose ISO 10642 - custom values is auto-checked and disabled, custom value fields are enabled (this is correct)
3. Change to an M3 size - Custom values is still checked and disabled, and the custom values are disabled

What should be happening after step 3 is that the custom checkbox should be enabled, and the state of the custom values fields should depend on whether it is checked or not.

donovaly:
> What should be happening after step 3 is that the custom checkbox should be enabled, and the state of the custom values fields should depend on whether it is checked or not.

Many thanks. Good catch.
This is now fixed.

chennes:
I can't find the commit that is making the change, but there is a missing comma at the end of line 7 of iso7046.json.

wwmayer:
Merged.

See 1c58c3b677c and 45dd1a7f608

PR #4342, opened 2021-01-31
Add Clone feature to Part workbench

pavltom:
This pull request adds "Clone" functionality to Part workbench

PR #4350, opened 2021-02-01
Part: Fix source typo in API Base::Placement getplacement() 

luzpaz:
Found originally by @pavltom in https://github.com/FreeCAD/FreeCAD/pull/4342/files#diff-35c6defb9ecd4c77532452e5aec35493d3728057e4e7b205655cd2d4b4e00181

### DIFF
--- a/src/Mod/Part/App/TopoShape.cpp
+++ b/src/Mod/Part/App/TopoShape.cpp
@@ -718,7 +718,7 @@ void TopoShape::setPlacement(const Base::Placement& rclTrf)
     _Shape.Location(loc);
 }
 
-Base::Placement TopoShape::getPlacemet(void) const
+Base::Placement TopoShape::getPlacement(void) const
 {
     TopLoc_Location loc = _Shape.Location();
     gp_Trsf trsf = loc.Transformation();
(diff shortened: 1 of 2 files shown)


yorikvanhavre:
Even passing the CI tests (the windows build fails but seems unrelated), this is a dangerous change as it would affect any other part of FreeCAD using that function... It sounds crazy that this function is not used anywhere... Did you check through the whole source code?

luzpaz:
:100:  certain. As I ran a search and found 0 results. I did find **many** results for `getPlacement()` and also in conjunction with `Base::Placement`.

wwmayer:
You have to blame me: https://github.com/FreeCAD/FreeCAD/commit/be18a46df9927a81fcbca7e19b2b28497d2779b8

I can't remember why I once have added it but probably it's a bit faster than the implementation in the base class ComplexGeoData. Since the two methods getPlacement() and setPlacement() in the base class are non-virtual they shouldn't be re-implemented in TopoShape.

Either we rename these methods or remove them.

wwmayer:
See 3ebd7d119b

luzpaz:
@wmayer great. Quick question. I'm not seeing any feedback to the user via the console mentioning this is an obsolete function. That way anyone using it has time to change their code before the end of v0.20. Which by that time we can remove from the API. Or am i missing something?

wwmayer:
In client code nothing needs to be changed.

The only reason I kept it is to test if the implementation in TopoShape is faster than that of the base class. If yes, then we can make the method of the base class virtual and rename the methods in TopoShape again. If not, then it can be removed.

PR #4352, opened 2021-02-03
[PD] fix typo in iso7046.json

donovaly:
This mistake was introduced by me in commit 4ab3c955

### DIFF
--- a/src/Mod/PartDesign/Resources/Hole/iso7046.json
+++ b/src/Mod/PartDesign/Resources/Hole/iso7046.json
@@ -4,7 +4,7 @@
     "thread_type": "metric",
     "angle": 90,
     "data": [
-    { "thread": "M1.6", "diameter":  3.6 }
+    { "thread": "M1.6", "diameter":  3.6 },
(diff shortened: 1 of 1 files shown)


chennes:
Oh, I'm glad to see this, I thought I'd gone crazy - the comma was missing in my local file, but I couldn't find the commit where it happened! Thanks!

wwmayer:
When running the TestPartDesignApp unit test (without this fix) some error messages of the form appear:
> Failed reading ‘/home/FreeCAD/share/Mod/PartDesign/Resources/Hole/iso7046.json’ with: [json.exception.parse_error.101] parse error at line 8, column 5: syntax error while parsing array - unexpected '{'; expected ']'

However, all the unit tests still succeed. So, IMO we should make the tests to fail if there is a syntax error in a json file.

wwmayer:
The exception is raised inside Hole::readCutDefinitions() and there is a try/catch block to report the syntax error.
Inside the catch block we could re-throw the exception but this has the side-effect that any attempt to create a hole feature will fail because readCutDefinitions() is called inside the constructor.

PR #4354, opened 2021-02-03
[0.19][Surface] Fix segfault in surface task view

chennes:
Forums discussion: https://forum.freecadweb.org/viewtopic.php?f=8&t=55036

If the surface task view is not closed prior to double-clicking on a sketch, the code segfaults due to the active dialog being removed while Sketcher is asking the user if they want to do that. This corrects the bad access by removing the task panel's self-deletion when unsetEdit was called, and replacing it with code that clears the active highlights and selection gate when the task view is hidden. This ensures that Escape-key behavior is still correct.

---

- No tracker ticket

### DIFF
--- a/src/Gui/TaskView/TaskView.cpp
+++ b/src/Gui/TaskView/TaskView.cpp
@@ -775,6 +775,10 @@ void TaskView::removeTaskWatcher(void)
 
 void TaskView::accept()
 {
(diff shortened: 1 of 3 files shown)


wwmayer:
9c9244dfb3

wwmayer:
See
1e7929ba59
9c3c562f7

chennes:
Great, thanks!

PR #4356, opened 2021-02-03
[0.19][ARCH] Arch: Fix LGTM warning of missing param for translate() in ArchPanel.py

luzpaz:
Even though this is ignored per the docs[1], LGTM complains[2] about it. This commit adds the superfluous param. Note: there is precedent in ArchReference.py[3] and ArchWall.py[4].
 
[1] https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Draft/draftutils/translate.py#L73-L78  
[2] https://lgtm.com/projects/g/FreeCAD/FreeCAD/snapshot/17db14c9700ae65d042d3548ef07f9518a115ba4/files/src/Mod/Arch/ArchPanel.py#L32  
[3] https://github.com/FreeCAD/FreeCAD/blob/45dd1a7f608f822aaab4dc52d2308edb977a3f5a/src/Mod/Arch/ArchReference.py#L39  
[4] https://github.com/FreeCAD/FreeCAD/blob/45dd1a7f608f822aaab4dc52d2308edb977a3f5a/src/Mod/Arch/ArchWall.py#L42

### DIFF
--- a/src/Mod/Arch/ArchPanel.py
+++ b/src/Mod/Arch/ArchPanel.py
@@ -29,7 +29,7 @@ if FreeCAD.GuiUp:
     import draftguitools.gui_trackers as DraftTrackers
 else:
     # \cond
-    def translate(ctxt,txt):
+    def translate(ctxt,txt,utf8_decode=False):
         return txt
     def QT_TRANSLATE_NOOP(ctxt,txt):
         return txt


hyarion (approved):
This should indeed fix the issues. :+1: 

Just for the record, the error can be found on line 222 in the file @luzpaz referenced: https://lgtm.com/projects/g/FreeCAD/FreeCAD/snapshot/17db14c9700ae65d042d3548ef07f9518a115ba4/files/src/Mod/Arch/ArchPanel.py#L222

chennes (approved):
FWIW, I think fixing these is a good idea: it's like making sure your compile doesn't throw any warnings. If you just live with it, you train yourself to ignore the warnings, even when they are important. By eliminating all the things that a tool like LGTM can catch (even when they are basically nothing) then when later you get a new one you know you have to look at it. So, thanks for doing this!

PR #4361, opened 2021-02-04
[0.20] [GUI] Add tooltip to f(x) when there is no expression

chennes:
As suggested by @luzpaz, this adds a tooltip to the f(x) icon of all of the widgets that use it: `QuantitySpinBox`, `UIntSpinBox`, `IntSpinBox`, `DoubleSpinBox`, and `ExpLineEdit`. These five classes are divided across three files, and all five classes duplicate the same basic code. In the existing code, no tooltip is shown if there is no expression. If there is an expression, it (alone) is used as the tooltip.

This commit modifies that behavior to have explanatory text as the tooltip when there is no expression, and when there is, to prepend a short string to the beginning.

To reduce further code duplication, this is implemented by adding a new function to `ExpressionLabel` allowing the text of the expression to be set. It checks that text and if it's empty, uses a preset default string as the tooltip. If there is expression text, it instead prepends another bit of text to the front and sets the entire expression+prefix as the tooltip.

TODO: In the future a lot of the preexisting code duplication could be removed by making the `ExpressionLabel` class do more internal management of the expression.

Thanks to @kisolre for the assistance tracking these down, and suggesting the default text used. Forum discussion topic: https://forum.freecadweb.org/viewtopic.php?f=17&t=55146

---

- No tracker ticket

### DIFF
--- a/src/Gui/QuantitySpinBox.cpp
+++ b/src/Gui/QuantitySpinBox.cpp
@@ -307,6 +307,7 @@ QuantitySpinBox::QuantitySpinBox(QWidget *parent)
     iconLabel->setPixmap(pixmap);
     iconLabel->setStyleSheet(QString::fromLatin1("QLabel { border: none; padding: 0px; padding-top: %2px; width: %1px; height: %1px }").arg(iconHeight).arg(frameWidth/2));
     iconLabel->hide();
+    static_cast<ExpressionLabel *>(iconLabel)->setExpressionText(QString());
     lineEdit()->setStyleSheet(QString::fromLatin1("QLineEdit { padding-right: %1px } ").arg(iconHeight+frameWidth));
     // When a style sheet is set the text margins for top/bottom must be set to avoid to squash the widget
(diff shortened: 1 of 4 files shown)


wwmayer:
The calls of dynamic_cast should be replaced with static_cast. The reason is that most source analyzers will complain if a dynamic_cast is used but not checked if it returns a null pointer.

chennes:
> The calls of dynamic_cast should be replaced with static_cast. The reason is that most source analyzers will complain if a dynamic_cast is used but not checked if it returns a null pointer.

Makes sense - change made.

luzpaz:
@wmayer can we merge this before the next crowdin push ? (preferably after #4557 gets merged)

wwmayer:
e151f19d41 avoids most code duplication of UIntSpinBox, IntSpinBox and DoubleSpinBox by implementing the class ExpressionSpinBox

wwmayer:
5d49bf78de reduces code duplication of QuantitySpinBox

PR #4293, opened 2021-01-19
[TD] make pointers to the UI a std::unique_ptr

donovaly:
as noted in https://github.com/FreeCAD/FreeCAD/pull/4271#discussion_r554673632
the pointer to the UI should be a unique pointer.

This PR does this for all TD dialogs that don't already use a unique_ptr.

PR #4271, opened 2021-01-11
[TD] add dialog to modify some dimension settings

donovaly:
The dimension object is the only one in TechDraw not having a dialog. This PR add this.
It can either be called by double-clicking on a dimension in the model tree or by double-clicking directly onto a dimension in the drawing. (The latter is a real boost in productivity since one usually has dozens of dimensions and every time first single-click on one in the drawing, then going to the model tree, the double-click needs is with the time tiring/annoying.)

There is how the dialog looks:
FreeCAD_BoJXxnoXJz

Maybe some more settings should appear there, however, for now it is important to have at least a dialog.

I also changed with the PR some wording.

The PR contains already the fixes I made in PR #4263 to avoid merge conflicts.

PR #4368, opened 2021-02-05
[Sketcher] make more pointers to the UI std::unique_ptr

donovaly:
addendum to PR #4362

(Same as PR #4293, just for Sketcher

as noted in https://github.com/FreeCAD/FreeCAD/pull/4271#discussion_r554673632
the pointer to the UI should be a unique pointer.

This PR does this for all Sketcher dialogs that don't already use a unique_ptr.)


wwmayer:
Merged.

PR #4209, opened 2020-12-28
PartDesign: New features AdditiveHelix and SubtractiveHelix

davidosterberg:
New feature proposal for the PartDesign WB.

This feature allows for simple creation of a helical sweeps from within PartDesign.
Sample application is threads, springs, coils, augers, etc.

Forum discussion:
https://forum.freecadweb.org/viewtopic.php?f=8&t=53714

Icon discussion
https://forum.freecadweb.org/viewtopic.php?f=34&t=53820

Workflow with new tool is similar to the Revolve/Groove tool. Code is based  primarily on on the Pipe feature in PartDesign.

For additive Helix we start with a sketch (the construction line is one way to defines the axis)
sketch

We click the Additive Helix icon in the toolbar
icon_add

Dialog and result:
add_dialog_and_result

For subractive helix we start with the stock
thread_stock

Followed by the sketch:
sketch_and_construction_line

Press the icon in the toolbar:
icon_sub

And set up the helix
dialog_and_result

**Todo list**
- [x] Reverse option
- [x] Fix issue with axis from feature line
- [x] Fix issue where base object is moved
- [x] Fix issue with self intersecting not being detected and handled correctly
- [x] Fix issue with helix starting point when axis from construction line
- [x] Create Update view checkbox, as recomputes sometimes can be slow 
- [x] Fix bug present with OCC version >= 7.5.0 which causes the Helix path to be oriented incorrectly
- [x] Update icons with feedback from the UI/UX forum
- [x] Add different input modes (pitch-turns, heigh-turns) 
- [x] Add some operation validation
- [x] Add "remove outside" option, so that boolean intersection is covered as a possibility. 
- [x] Make sure that command is usable from Python
- [x] Make sure that recorded macros play back correctly

**Wont fix**
- Up to face (usecase not so great, and complicated)
- Silent failure in case of multiple solids. Issue error message in this case. (appears all other PD features have this problem except Pad. Not clear to me how Pad solves this)

PR #4373, opened 2021-02-05
[Do not merge] PartDesign: Fix Helix Icons - Wiki discoverability

davidosterberg:
As reported in #4209 the Wiki's What's This page is automatically generated and requires a specific naming convention of icons. This PR attempts to address this issue.  A drawback of this naming convention is that the names appear to deviate from the naming convention in PartDesign

Before:
image

After:
image


davidosterberg:
@donovaly , do you think we should fix all the icons in PartDesign in the same PR. Basically the changes will be identical for all the features.

donovaly:
> do you think we should fix all the icons in PartDesign in the same PR

Yes, but please not before the feature freeze has been announced. Then we have time to take of this since we will not affect pending PRs and discuss this also in the forum.

PR #4380, opened 2021-02-06
LGTM: Remove catch block for plain C strings

chennes:
LGTM complains about this construct leaking memory. Commit 13927b2465 eliminated the parts of FreeCAD that throw those strings. This commit eliminates the catch block.


luzpaz:
Nice. Thank you! BTW, @chennes  I'll trigger a new LGTM analysis now.

PR #4274, opened 2021-01-11
[0.20] PartDesign: Add true threads to the Hole feature

davidosterberg:
This PR adds possibility to make true threaded holes  with the Hole feature (possible to e.g 3D print). 

Forum discussion:
https://forum.freecadweb.org/viewtopic.php?f=34&t=54240

It should work for all supported thread types
both_direction_demo
demo

Todo

- [x] Give the user possibility to tweak the clearance of the thread (important for 3D printing)
- [x] Turn off auto-recompute when Model Thread is active. It is too slow if there are large hole patterns.
- [x] Rename "Model Actual Thread" to simply "Model Thread" in the Ui
- [x] Default threading depth to 3.5 x P from the hole bottom. 
- [x] Add UI support for specifying thread depth
- [x] Add  backend support for specifying thread depth
- [x] Remove column 2 from hreadClass_ISOmetric_data
- [x] Fix donovaly bug A
- [x]  Fix donovaly bug B
- [x]  Fix donovaly bug C
- [x] Wait for PR #4134

PR #4381, opened 2021-02-06
[0.20] PartDesign: Hole performance improvements

davidosterberg:
With PR #4274, the hole feature can be very slow if thread modeling is enabled for a large hole pattern. 
This PR contains some performance optimization. With debug build the performance improvement is about 30%. 

  | Build type | Debug |   |  
-- | -- | -- | -- | --
  | Test case | Hole pattern 6 x M10x25 |   |  
  |   |   |   |  
Commit | Variant | time (average of 10) |   | Improvement
db9525e7d7 | Original Hole | 3,1325 | s | 0,00 %
6f63b105da | Compound then cut | 2,22 | s | 29,13 %
354ab5528 | remove unused code | 2,19 | s | 30,09 %

A note about changes in behavior.

Previously it was posssible for the holes to overlap each other. Now this is no-longer supported as this would be more computationally demanding. If it is seen as needed to support overlapping holes, support for overlapping holes can be implemented if necessary.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHole.cpp
+++ b/src/Mod/PartDesign/App/FeatureHole.cpp
@@ -1959,9 +1959,9 @@ App::DocumentObjectExecReturn *Hole::execute(void)
         BRep_Builder builder;
         TopoDS_Compound holes;
         builder.MakeCompound(holes);
-
         TopTools_IndexedMapOfShape edgeMap;
         TopExp::MapShapes(profileshape, TopAbs_EDGE, edgeMap);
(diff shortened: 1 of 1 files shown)


donovaly:
Since PR #4274 should go in for 0.19, this PR is really necessary since otherwise it is too slow for normal usage.

tomasix (approved):
Works for me

davidosterberg:
@wwmayer, Any objections to this being merged?

wwmayer:
> Any objections to this being merged?

No, but I just haven't had the time yet to test it. I will have a look today afternoon/evening.

PR #4383, opened 2021-02-07
[Path]: Set operation Visibility to False, and turn it back on while editing.

mlampert:
This is a small change to hide an op by default, make it visible while it is being edited and restore the previous visibility afterwards. Typically the generated Path of the op is still visible through the visibility of the Job object.

The reason for this change is the many confusion and annoyance the double display of an op's path causes for users, especially newer users.


sliptonic:
https://forum.freecadweb.org/viewtopic.php?f=15&t=55299

PR #4390, opened 2021-02-07
FEM: Improve Gmsh log, when Gmsh is not installed

davidosterberg:
**Before this PR:**
Even when Gmsh is not installed, everything looks fine:
everything_looks_fine
but_it_is_not

**After PR:**
after

Also ensure that any unexpected exception is also shown in the Gmsh dialog log

### DIFF
--- a/src/Mod/Fem/femmesh/gmshtools.py
+++ b/src/Mod/Fem/femmesh/gmshtools.py
@@ -40,6 +40,10 @@ from femtools import femutils
 from femtools import geomtools
 
 
+class GmshError(Exception):
+    pass
+
+
 class GmshTools():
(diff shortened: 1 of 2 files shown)


davidosterberg:
ping @berndhahnebach

berndhahnebach:
I am a week off. I'be back on 14'th of February.

davidosterberg:
@wwmayer I think this one is straight forward and should be considered a bug fix.

wwmayer:
So, far the problem only occurs if the option "Search in known binary directories" is set and gmsh not installed. If the above option is not set then it raises an error, too.

I will test the PR locally and if everything works as expected I will merge it.

PR #4393, opened 2021-02-07
[Path]: bugfix/issue 4557

mlampert:
https://tracker.freecadweb.org/view.php?id=4557


luzpaz:
hey @mlampert thanks for the fix. To auto-close mantis tickets make sure you use the correct triggers https://wiki.freecadweb.org/Tracker#GitHub_and_MantisBT

PR #4396, opened 2021-02-08
[0.19] [Mesh] LGTM: Correct minimum distance calculation in SplitFacet

chennes:
The function `SplitFacet()` appears to be unused (it has a "FIXME: Implement" note at the top), but LGTM has identified a reasonably clear error in it, which this commit resolves in the interest of making someone's life easier in the future (and shutting LGTM up). In calculations of a minimum distance, the variable `minDist` is initialized to `FLOAT_MAX`, and then later is compared to see if it is _less than_ some other value, which cannot ever be true. The sense of the comparison has been reversed.

Found via LGTM.

---

- No tracker ticket filed.


hyarion (approved):
`fMinDist` is initialized with `FLOAT_MAX` in all cases so this must indeed be a bug.

luzpaz:
FTR, here's the FIXME line:
https://github.com/FreeCAD/FreeCAD/pull/4396/files#diff-8e238f28cbc35232a18d39139e198e0de34efd17d3774f6071a49fe7ba8a0b5cR1112

vosk:
Why not delete the unused code?

PR #3893, opened 2020-09-23
[PD] add along length option for pad

donovaly:
[PD] add along length option for pad

- as discussed here: https://forum.freecadweb.org/viewtopic.php?f=17&t=50466&start=10#p433327
this PR is the first step for new Pad features.
It implements 3 features requested by users:
1. an option to measure the pad length along its direction if there is a custom direction
2. when there are two lengths, the reverse feature should be active
3. the pad dialog shows now always the used pad direction, no matter if it is a custom one or the sketch's normal

- Furthermore there are some code improvements (avoid rounding and blocking signals)

- Also adapt TestPad.py because its test used the case two lengths + reversed but this was not possible before this PR - the reverse option was not take into account for two lengths in FeatureSketchBased.cpp)
Now it is and therefore the test must be updated:
In the test the second sketch has the normal vector 0, -1, 0. As the reverse option is set the pad direction is 0, 1, 0. So in y direction is the length (1mm) and in -y direction is length2 (2mm). This gives together with the other pad of volume 1, a total volume of 4.

- fix UI issue: either reversed or midplane
As noticed by @chennes, when the pad uses symmetric, reversed is not sensible and vice versa.
This commit fixes the missing Gui side for the case midplane is checked and the App side.

- the PR also fix two typos

PR #4409, opened 2021-02-09
[TD] fix balloon and dimension dialog rejection

donovaly:
the balloon and the Dimension dialog did not restore the previous settings when the user press Cancel

The PR also adds a missing recompute, reported as bug B. in https://forum.freecadweb.org/viewtopic.php?p=474482#p472939

### DIFF
--- a/src/Mod/TechDraw/Gui/TaskBalloon.cpp
+++ b/src/Mod/TechDraw/Gui/TaskBalloon.cpp
@@ -107,6 +107,9 @@ TaskBalloon::TaskBalloon(QGIViewBalloon *parent, ViewProviderBalloon *balloonVP)
     }
     // new balloons have already the preferences BalloonKink length
     ui->qsbKinkLength->setValue(parent->dvBalloon->KinkLength.getValue());
+
+    // store the initial values to restore them on reject()
+    getParameters();
 }
 
 TaskBalloon::~TaskBalloon()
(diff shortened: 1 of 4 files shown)


wwmayer:
This is not the correct way to handle the restoring of properties when rejecting the changes. And it's error-prone as you always need to adjust the code when a new property have been added.

The way to work with transactions is:
For Accept:
```
openTransaction("Change object");
# change the various properties
...
commitTransaction();
```

For Revert:
```
openTransaction("Change object");
# change the various properties
...
abortTransaction();
```

donovaly:
> The way to work with transactions is:...

Thanks for having a look. I'll try to fix this.
I took the method I used here from other TD dialogs so there are more dialogs to be fixed.

(That's why I often wrote we should taking a breath to take care to fix such long-standing issues in a feature freeze. Now I cannot promise I will have the time to do this work in the remaining 9 days. The feature freeze is too short and still not al pending PRs are in to be tested by the crowd.)

By the way, can you please look at the long-standing PR #3893  that one fixes annoying things I see frequently when working with FC.

wwmayer:
> Thanks for having a look. I'll try to fix this.

No need. I am working on it right now. There are a few more issues to fix.

wwmayer:
bb0d75b6cc fixes of the balloon feature

wwmayer:
3288c02eba fixes of the dimension feature

wwmayer:
So this PR has become obsolete.

donovaly:
Thanks. However, with your commits I cannot undo changes and also pressing Cancel in the dialog is not respected, see https://github.com/FreeCAD/FreeCAD/commit/bb0d75b6cc29dcc722b6596b180fd8d205f5073b#commitcomment-47057145 for a recipe how to reproduce.

PR #4414, opened 2021-02-09
LGTM: Remove unsafe time functions

chennes:
This PR is in two commits: the first, 7f1dc84, replaces the global-shared-memory versions of the time functions with those that work with local memory. The second, 32a34f5, eliminates entirely an unused function that would otherwise have needed both a similar fix for the time functions, and a function signature change (from char* to string) to eliminate a local version of the same flaw.

---

- [X]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0` (except the four that were failing before!)
- No tracker ticket.

### DIFF
--- a/src/App/Application.cpp
+++ b/src/App/Application.cpp
@@ -2292,7 +2292,16 @@ void Application::logStatus()
 {
     time_t now;
     time(&now);
-    Console().Log("Time = %s", ctime(&now));
+    struct tm timeStruct; 
+#ifdef FC_OS_WIN32
+    localtime_s(&timeStruct, &now);
+#else
+    localtime_s(&now, &timeStruct);
+#endif
+    char timeString[100];
+    if (std::strftime(timeString, sizeof(timeString), "%c", &timeStruct)) {
+        Console().Log("Time = %s\n", timeString);
+    }
 
     for (std::map<std::string,std::string>::iterator It = mConfig.begin();It!= mConfig.end();++It) {
         Console().Log("%s = %s\n",It->first.c_str(),It->second.c_str());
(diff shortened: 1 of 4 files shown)


wwmayer:
There is a problem with localtime_s. According to https://en.cppreference.com/w/c/chrono/localtime the MS implementation is incompatible to the C standard because the arguments are swapped.

Then in Document.cpp we have to include ctime in the ifndef _PreComp_ section.

chennes:
> There is a problem with localtime_s. According to https://en.cppreference.com/w/c/chrono/localtime the MS implementation is incompatible to the C standard because the arguments are swapped.

It gets worse - it turns out this is one of the places that C++ and C are not synched. While the C11 standard defines the _s version of the function, C++11 does not. So it's a non-portable, implementation-defined feature. Which is to say, IMO we can't use it. C++ `std::chrono` doesn't really support the direct output of this style of time-string, even in C++20. So, if we want to address this, I believe we are left with `boost::chrono`. I'm happy to write it, but this has gotten to be less like a "quick bug fix" and more like a "feature". So what do you think about delaying it until after 0.19 is released?

wwmayer:
> So what do you think about delaying it until after 0.19 is released?

Yes, that's probably best. But nevertheless I could cherry-pick the change to remove ConsoleSingleton::Time() as it's not used anywhere.

wwmayer:
Cherry-picked 4ca6f68 became a0bbb173f47f

chennes:
I'm closing this PR since the part that worked has been cherry picked, and the remainder needs to be rewritten (or we just need to tell LGTM to ignore the issue).

PR #4418, opened 2021-02-09
[Part] Coverity: dtors can't throw

chennes:
Catch and report exceptions in destructors to prevent them from propagating out. Found via Coverity.

---

- Small fix
- No ticket

### DIFF
--- a/src/Mod/Part/Gui/DlgProjectionOnSurface.cpp
+++ b/src/Mod/Part/Gui/DlgProjectionOnSurface.cpp
@@ -178,7 +178,12 @@ DlgProjectionOnSurface::~DlgProjectionOnSurface()
   delete ui;
   for ( auto it : m_projectionSurfaceVec)
   {
-    higlight_object(it.partFeature, it.partName, false, 0);
+    try {
+      higlight_object(it.partFeature, it.partName, false, 0);
+    }
+    catch (Standard_NoSuchObject& e) {
+      Base::Console().Warning("DlgProjectionOnSurface::~DlgProjectionOnSurface: %s", e.GetMessageString());
+    }
     PartGui::ViewProviderPartExt* vp = dynamic_cast<PartGui::ViewProviderPartExt*>(Gui::Application::Instance->getViewProvider(it.partFeature));
     if (vp)
     {
--- a/src/Gui/ViewProviderLink.cpp
+++ b/src/Gui/ViewProviderLink.cpp
@@ -2739,9 +2739,14 @@ void ViewProviderLink::dragMotionCallback(void *data, SoDragger *) {
 }
 
 void ViewProviderLink::updateLinks(ViewProvider *vp) {
-    auto ext = vp->getExtensionByType<ViewProviderLinkObserver>(true);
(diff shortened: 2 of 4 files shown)


src/Mod/Part/Gui/DlgProjectionOnSurface.cpp
```diff
@@ -178,7 +178,12 @@ DlgProjectionOnSurface::~DlgProjectionOnSurface()
   delete ui;
   for ( auto it : m_projectionSurfaceVec)
   {
-    higlight_object(it.partFeature, it.partName, false, 0);
+    try {
+      higlight_object(it.partFeature, it.partName, false, 0);
```
luzpaz, line 182: should this be `highlight_object` ?

chennes: Yes. That said, in the interest of keeping these static analysis PRs focused I am trying not to make any unrelated changes.

wwmayer:
The actual exception is theoretically thrown inside the function ViewProviderLink::updateLinks(). Wouldn't it be better to handle it directly there because then in case of failures it's much easier to get the context where it happened?

chennes:
> The actual exception is theoretically thrown inside the function ViewProviderLink::updateLinks(). Wouldn't it be better to handle it directly there because then in case of failures it's much easier to get the context where it happened?

That's the one I put in CrossSections.cpp, right? I will look into moving the catch to someplace more local (unless you've already done it).

wwmayer:
Yes, there is a defect report about MeshPart/Gui/CrossSections.cpp. From the defects of type "Uncaught exception" the majority is due to the use of removeViewProvider() in a destructor which after several steps uses the function ViewProviderLink::updateLinks().

chennes:
>  due to the use of removeViewProvider() in a destructor which after several steps uses the function ViewProviderLink::updateLinks().

Thanks -- I moved the handling of that exception up to the `updateLinks()` function, which no longer throws (though I didn't mark it `noexcept` -- should I?). The only "handling" I do is report the exception.

wwmayer:
> though I didn't mark it noexcept -- should I?

Well, at the moment we don't use it anywhere but for the future we can consider using more of the new C++ stuff.

> The only "handling" I do is report the exception.

That's OK as it can only raise this exception type in theory but in practice it I can't see when this can happen.

Btw, do you remember what function raised an exception when using clearSelection()? Here we could do the same to avoid to add a try/catch block to all the affected destructors.

chennes:
> Btw, do you remember what function raised an exception when using clearSelection()? Here we could do the same to avoid to add a try/catch block to all the affected destructors.

It's in `void SelectionSingleton::notify(SelectionChanges &&Chng)` -- the call to `isSelected()` includes the document name, which is checked for existence.

wwmayer:
In some older reports checkSelection() was already mentioned where it reported the unhandled Py::Exception. The function at the very end was View3DInventorViewer::resetEditingRoot() and the reason is that a new Py::Exception could be raised in the catch handler. So, I thought 3a8a28ce920e4 should fix that one.

PR #4423, opened 2021-02-10
[PD] rename icons to match Wiki pages

donovaly:
For the "What's this" feature we need to have a uniform naming of the features and its icons. For some PD icons this is not the case and since we are in feature freeze, this is the right time to address this.
(I see the same is already done for Mesh.)


wwmayer:
Merged with 2834075d6a

PR #4426, opened 2021-02-10
[v0.20] Sketcher: Change constraint-conversion MsgBox to DlgCheckableMessageBox

davidosterberg:
Forum discussion
https://forum.freecadweb.org/viewtopic.php?f=34&t=55437

opt-out-notify

The 3 commits outlines several possibilities for how to deal with the existing preference.

**Only commit 1**
In this case the Sketcher settings checkbox becomes kind of broken if the user uses the checkbox in the MessageBox. 

**Commit 1 +  2**
In this case the Sketcher Settings checkbox is still the master. The preference can be changed either from the Settings dialog, or from the messagebox.

**Commit 1 + 3**
In this case the MessageBox preference replaces the old preference. The preference is removed from the Sketcher Settings dialog

**Commit 1 + 2 + 3**
The old preference is still master. But it is controlled from the MessageBox only.  If you want to get your warning back, you have to use the Preferences editor.

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -41,6 +41,7 @@
 #include <Gui/DlgEditFileIncludePropertyExternal.h>
 #include <Gui/Action.h>
 #include <Gui/BitmapFactory.h>
+#include <Gui/DlgCheckableMessageBox.h>
 
 #include <Mod/Part/App/Geometry.h>
 #include <Mod/Sketcher/App/SketchObject.h>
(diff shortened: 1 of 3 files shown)


abdullahtahiriyo:
I have merged the same principle with some modifications:
1. I did not like to need to introduce a new parameter to control the old parameter (this was necessary due to the limitations of the checkable dialog that hardcoded the preference path). 
2.  I wanted to maintain the possibility to restore the modal dialogs directly from the sketcher preferences (maintain the old behaviour).
3. Having parameters from different workbenches in a single prefpath did not promote workbench separation. While it is a weaker coupling than code coupling, it is better to avoid it when appropriate.

Moving away from this required to extend the funcionality of the checkable dialog. This was done in 4bf6b1984971547372b994b99371ca2a4351e138.

So basically I merged the first two commits squashed as d759a8e43d607442250b26efae70abda582a1007, and then modified it to use the new extended framework of the checkable dialog. Because there will be another two constraint substitutions when the trimming PR is merged, I refactored the code into a separate function taking the message, which is case specific as a parameter. This is c08a19626fa637dc411abb30b392d70d817994d7.

I hope this fulfills everyone's wishes.

PR #4429, opened 2021-02-11
Coverity: geoit can be end() when dereferenced

chennes:
Coverity has identified a potential bug that this PR seeks to address: in the Sketcher function `analyseBlockedGeometry()` there is a conditional:
```
auto geoit = std::find(blockedGeoIds.begin(),blockedGeoIds.end(),c->Second);
if(geoit != blockedGeoIds.end() || onlyblockedGeometry[c->Second]) { // internal alignment geometry found, add to list
```
and then later on inside that conditional
```
blockedGeoIds.push_back(*geoit);
```
However, since the condition is an OR, it is apparently possible that `geoit` is `end()`, in which case dereferencing it is illegal. The fix proposed in this PR is to modify the OR to an AND. However, this needs review.

This fix is a joint effort with @hyarion.

### DIFF
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -184,7 +184,7 @@ bool Sketch::analyseBlockedGeometry( const std::vector<Part::Geometry *> &intern
 
                 auto geoit = std::find(blockedGeoIds.begin(),blockedGeoIds.end(),c->Second);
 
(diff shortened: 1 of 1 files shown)


wwmayer:
Replace `blockedGeoIds.push_back(*geoit);` with `blockedGeoIds.push_back(c->Second);`

hyarion:
Just for the record, here's the forum post regarding the proposed fix: https://forum.freecadweb.org/viewtopic.php?p=476886#p476886

chennes:
Closed in favor of the real fix in https://github.com/FreeCAD/FreeCAD/pull/4442

PR #4430, opened 2021-02-11
[0.20] Draft: allow items to be skipped in draft array

davidosterberg:
Implemented for Array, Polar array, and Circular array.  

Motivation for feature:
Often when you make pattern (e.g a bolt pattern), there will by one or two items that should be skipped for some reason.

Forum discussion:
https://forum.freecadweb.org/viewtopic.php?f=23&t=55459

If you consider this interesting, I will add support for the other Array types. 
At a later stage this can also be exposed in the GUI

skip_otho
draft_array_skip

**Advanced indexing:**
Skip expressions are expanded in a python/matlab like way

        Example 1: skip_expression = 1:5
            Will be expanded to [1, 2, 3, 4]

        Example 2: skip_expression = [2:5, 9]
            Will be expanded to [2, 3, 4, 9]

        Example 3: skip_expression = [1, 3:7:2, 9]
            Will be expanded to [1, 3, 5, 9]

advanced_indexing

### DIFF
--- a/src/Mod/Draft/draftobjects/array.py
+++ b/src/Mod/Draft/draftobjects/array.py
@@ -34,6 +34,7 @@ split in separate classes so that they are easier to manage.
 ## \addtogroup draftobjects
 # @{
 import math
+import json
 from PySide.QtCore import QT_TRANSLATE_NOOP
 
 import FreeCAD as App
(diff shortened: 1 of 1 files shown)


davidosterberg:
There is already a way (perhaps not so discoverable) to skip items in draft array. Simply expand the array, the hide the item you don't want. The the array can be un-expanded. Therefore I close this PR.

PR #4431, opened 2021-02-11
Part: Allow helix primitive with negative Angle

davidosterberg:
Forum discussion:
https://forum.freecadweb.org/viewtopic.php?f=8&t=55449

negative_helix_angle

### DIFF
--- a/src/Mod/Part/App/PrimitiveFeature.cpp
+++ b/src/Mod/Part/App/PrimitiveFeature.cpp
@@ -74,7 +74,7 @@
 
 
 namespace Part {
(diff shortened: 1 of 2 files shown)


donovaly:
@wwmayer , I would say it is 0.19 stuff. The helix is a new feature in just since a week. Users tested it and in fact negative angles were simply forgotten. So this PR polishes a new feature to make it fully usable.
Moreover the "fix" of this PR is trivial.

donovaly (approved):
Thanks. Now the helix feature should be fully usable.

wwmayer:
There are really not many changes so it can be accepted for v0.19.

davidosterberg:
Thanks

PR #4432, opened 2021-02-11
[CMAKE] Add missing PrintFinal() version info

chennes:
Modeled after the code for Eigen3, this adds a version check for Coin. If pkg_config (or some other method) finds the version string, just print that. If not, do a REGEX MATCH on the file in Coin's headers that includes the version information. Note that in this case the beta version information is excluded. 

There is no pkg_config for MEDFile, so it just scans the appropriate header for the version info.

It removes an old, no-longer-needed custom search for XercesC and uses the built-in CMake version, which provides version info.

Finally, it prints out the version info for the other libraries where it is known.


cMake/FreeCAD_Helpers/PrintFinalReport.cmake
```diff
@@ -166,7 +166,7 @@ macro(PrintFinalReport)
     message(STATUS "OpenGL:              ${OPENGL_gl_LIBRARY}")
     message(STATUS "OpenGLU:             ${OPENGL_glu_LIBRARY} [${OPENGL_glu_LIBRARY}][${OPENGL_INCLUDE_DIR}]")
```
luzpaz, line 167: It's weird that `${OPENGL_glu_LIBRARY}` is repeated here twice

chennes: My guess is that it was a copy-and-paste job where the author intended to replace the first one with `OPENGL_glu_VERSION` (which doesn't exist, as far as I am aware, the OpenGL stuff doesn't have easy-to-access version info). We should probably remove it.

wwmayer:
Should we add this for v0.19 or wait for v0.20? So far I have tested it on Linux and Windows (debug + release builds) and it worked for me.

luzpaz:
I think we should add it.

chennes:
I don't feel strongly either way, that's more your call than mine :) .

wwmayer:
OK, then let's wait because you never know if this causes unexpected problems anywhere.

eszlari:
Coin provides `coin-config-version.cmake`. Wouldn't it make more sense to just remove `FindCoin3D.cmake` and require are current version of coin?

chennes:
Sure, if `find_package` will work for the versions of Coin we care about now, that's a great solution. This may just be one of those old checks that has been around for a decade and is no longer needed. That's the change I made to XercesC.

wwmayer:
Merged.

luzpaz:
0ed341a84c1ded47b1ea08c3389a84ef08a7e50f
74b0c6efaebd633a5d1ff0409fad1cbd0dddd24a
221a43255c921c90a113194cfb2df266ce2ab935

PR #4434, opened 2021-02-11
[Arch] Fix for ArchPipe: getProfile returned Shape instead of Wire

Roy-043:
Fix for ArchPipe: getProfile returned Shape instead of Wire

Forum topic:
https://forum.freecadweb.org/viewtopic.php?f=23&t=55401


yorikvanhavre:
Makes sense. Thanks!

PR #4435, opened 2021-02-11
[0.19] PartDesign: Fix hardcoded through all distance in ProfileBased

davidosterberg:
Before the throughall distance was 10 m, which is not enough for
many applications. The fix is to use the bounding box of the base shape
together with the sketch profile to calculate dynamically a large enough
length.

**Before:**
before

**After:**
after

### DIFF
--- a/src/Mod/PartDesign/App/FeatureSketchBased.cpp
+++ b/src/Mod/PartDesign/App/FeatureSketchBased.cpp
@@ -543,6 +543,20 @@ void ProfileBased::getUpToFace(TopoDS_Face& upToFace,
     }
 }
 
+double ProfileBased::getThroughAllLength() {
+    TopoDS_Shape profileshape;
+    TopoDS_Shape base;
+    profileshape = getVerifiedFace();
+    base = getBaseShape();
+    Bnd_Box box;
+    BRepBndLib::Add(base, box);
+    BRepBndLib::Add(profileshape, box);
+    box.SetGap(0.0);
+    // The diagonal of the bounding box, plus 1%  extra to eliminate risk of
+    // co-planar issues, gives a length that is guaranteed to go through all.
+    return 1.01*sqrt(box.SquareExtent());
+}
+
 void ProfileBased::generatePrism(TopoDS_Shape& prism,
                                 const TopoDS_Shape& sketchshape,
                                 const std::string& method,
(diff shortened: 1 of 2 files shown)


donovaly:
Why is this for 0.20? It fixes a clear bug - through all is not through all. It would be sad not to merge bugfixes for 0.19.

@wwmayer , what do you think?

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -543,6 +543,18 @@ void ProfileBased::getUpToFace(TopoDS_Face& upToFace,
...
+double ProfileBased::getThroughAllLength() {
+    TopoDS_Shape profileshape;
+    TopoDS_Shape base;
+    profileshape = getVerifiedFace();
+    base = getBaseShape();
+    Bnd_Box box;
+    BRepBndLib::Add(base, box);
+    BRepBndLib::Add(profileshape, box);
+    box.SetGap(0.0);
+    return 1.01*sqrt(box.SquareExtent());
```
donovaly, line 555: I miss documentation on what you do why.
- why the magic number 1.01? I guess to be a bit longer than necessary, so a comment would be helpful or others
- why do you take the root of the square extent and not the extent itself?

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -558,7 +570,7 @@ void ProfileBased::generatePrism(TopoDS_Shape& prism,
         if (method == "ThroughAll")
             // "ThroughAll" is modelled as a very long, but finite prism to avoid problems with pockets
```
donovaly, line 571: also here should be an updated comment what s done and the comment info that 1E6 once created problems is quite meaningless and should go.

donovaly (changes requested):

src/Mod/PartDesign/App/FeatureSketchBased.cpp
davidosterberg: I removed the comment. I think the name of the function explains it well enough.

src/Mod/PartDesign/App/FeatureSketchBased.cpp
davidosterberg: Comment added.

wwmayer:
> what do you think?

+1

donovaly (approved):
Thanks, From my point of view it can go in.

Just a minor formatting issue: The common formatting seems to be to have spaces around math operators, so "1.01 * sqrt(box.SquareExtent());"

wwmayer:
Merged.

PR #4440, opened 2021-02-12
[Path]: bugfix/invalid base geometry robustness

mlampert:
currently if the base geometry of an op becomes invalid the op cannot be edited to fix the base geometry because it throws an exception during task panel initialisation which prevents the event handlers to be registered - rendering the task panel mostly unresponsive (OK and Cancel still works though).

This change checks the base geometry and resets it if it is found invalid, allowing the task panel to be created successfully and let the user fix the base geometry.


ljkenny:
Small typo in the comment `s/stil/still/`.

Other than that, LGTM, thanks for fixing this.

PR #4442, opened 2021-02-12
Sketcher: Coverity fix in Sketch::analyseBlockedGeometry

abdullahtahiriyo:
========================================================

Users chennes and hyarion made me aware of this coverity issue:

Fixes Coverity: geoit can be end() when dereferenced
https://github.com/FreeCAD/FreeCAD/pull/4429/files#

When analysing the block where the dereferrencing appears, it
appears that it is a left-over that no longer makes sense:
- The algorithm classifies block constraints into those that are
not affected by any other driving constraint and those that are
affected by other driving constraints.
- The offending block deals with internal aligned geometry, thus
per definition has a driving internal alignment constraint, for which
the previous block already set the need of post-analysis.
- No matter what, the geometries, the complex one and the internal one
will have at least the driving internal alignment constraint, so they
cannot become "not affected by any other driving constraint".
- If the geometry had a block constraint on it, it was already added for
post-analysis in the previous block. If it did not have one block constraint,
the fact that it is internal aligned geometry is an irrelevant consideration.

Probably there was a point during development when this made sense, but with
the current post-analysis, it does not appear to make sense anymore. So the
block was removed.

This commit adds a unit test for blocked geometry (new block constraint).

### DIFF
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -138,7 +138,9 @@ bool Sketch::analyseBlockedGeometry( const std::vector<Part::Geometry *> &intern
                                      std::vector<bool> &onlyblockedGeometry,
                                      std::vector<int> &blockedGeoIds) const
 {
-    bool isSomethingBlocked = false;
+    // To understand this function read the documentation in Sketch.h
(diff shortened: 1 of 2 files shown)


abdullahtahiriyo:
git commit 5086c5f38dd32fb12d0e3c5b5cca40da6574bde1

PR #4443, opened 2021-02-12
[Path]: Added support for Part.OffsetCurve to flipEdge.

mlampert:
gitter user found error when using deburr - seems a new type of curve is not supported for flipping, which was required in this case.


ljkenny:
LGTM.  Thanks for fixing this.

PR #4446, opened 2021-02-13
[0.20][PATH] Deburr improve

Schildkroet:
Added option to select an angled face (chamfer) for a deburr operation
Added option to deburr an arc segment (open circle)

image


J-Dunn:
This is nice. I've always thought it was a shame not be able have the chamfer on the model. This is important when designing in relation to other features.  Thanks.

Russ4262 (approved):
I recall working on this with Schrildkroet and tracking down a bug.  It works as intended.  Current Deburr test passes, as do the others.

Thanks for the improvement, sir.

Russ4262:
Afternoon Schildkroet.
Does this PR need to be rebased due to changes that have occurred since its submission?

Schildkroet:
I updated the branch. No coflicts. Checks are still running

sliptonic:
This looks ok so I'll merge.  The opExecute method is getting quite long and heavily indented.  That's usually a clue that it's doing too much and should be refactored.  I see a couple parts of it that could be pulled out to helper functions.  Not critical but worth noting.

PR #4451, opened 2021-02-14
[0.20] PartDesign: Allow use of circle edge as axis in ProfileBased

davidosterberg:
Forum post.
https://forum.freecadweb.org/viewtopic.php?f=19&t=55573

pd_circle_selection_mode


davidosterberg:
I do all in #4469

PR #4454, opened 2021-02-14
PathToolBitLibraryGui: Fix LinuxCNC tool table format

lag-linaro:
The currently exported tool table for LinuxCNC results in the following warning:

  "Requested tool 1 not found in the tool table"

The debug output provides:

  "File: default_tool_table.tbl Unrecognized line skipped: \
      T1 P X0 Y0 Z0 A0 B0 C0 U0 V0 W0 D4.00 I0 J0 Q0;      \
      4mm_2F_Compression_Carbide_Endmill"

This is due to a missing/blank P (pocket) entry.

Signed-off-by: Lee Jones <lee.jones@linaro.org>

### DIFF
--- a/src/Mod/Path/PathScripts/PathToolBitLibraryGui.py
+++ b/src/Mod/Path/PathScripts/PathToolBitLibraryGui.py
@@ -736,7 +736,7 @@ class ToolBitLibrary(object):
                 if bit:
                     PathLog.track(bit)
 
-                    pocket = bit.Pocket if hasattr(bit, "Pocket") else ""
+                    pocket = bit.Pocket if hasattr(bit, "Pocket") else "0"
                     xoffset = bit.Xoffset if hasattr(bit, "Xoffset") else "0"
                     yoffset = bit.Yoffset if hasattr(bit, "Yoffset") else "0"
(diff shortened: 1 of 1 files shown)


hyarion (changes requested):
I will start off by saying that I haven't actually tested your PR and that the following review will only be based on the documentation for linuxcnc that I've been able to find on this topic.

The linuxcnc documentation states that [(1) tool pockets should be >= 1](http://wiki.linuxcnc.org/cgi-bin/wiki.pl?ToolTable), [(2) tool pockets should all be unique and (3) the pocket numbers you use must make sense on your machine](http://linuxcnc.org/docs/html/gcode/tool-compensation.html).

With that in mind, I don't think this is a good solution as:
1. the new default behaviour is to use slot number 0 which is out of bounds according to the specification
2. if multiple tools uses the default value they will share the same pocket which means the pocket numbers will not be unique (they will all use P0)
3. I'm not sure if this field can be automatically assigned as it will depend on how your machine is set up. If I understand how linuxcnc works, all tools would use tool 0 from the tool changer.

Maybe it would be better to raise an exception and display a message to the user that a slot needs to be specified? Or what do you think?

PR #4456, opened 2021-02-15
[0.19] [TD] fix dragging of balloons

donovaly:
as reported in https://forum.freecadweb.org/viewtopic.php?f=10&t=55109&start=30#p478147
dragging a balloon pollutes the undo steps destroying working with undo and drawings.

### DIFF
--- a/src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
+++ b/src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
@@ -644,8 +644,8 @@ void QGIViewBalloon::drawBalloon(bool dragged)
         y = Rez::guiX(balloon->Y.getValue() * refObj->getScale());
     }
     else {
-        x = Rez::guiX(balloonLabel->X() / refObj->getScale());
-        y = -Rez::guiX(balloonLabel->Y() / refObj->getScale());
+        x = balloonLabel->X();
+        y = -balloonLabel->Y();
     }
     Base::Vector3d lblCenter(x, -y, 0.0);
 
--- a/src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
+++ b/src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
@@ -532,6 +532,7 @@ void QGIViewBalloon::balloonLabelDragFinished()
     Gui::Command::commitCommand();
 
     m_dragInProgress = false;
+    m_originDragged = false;
 }
 
 //from QGVP::mouseReleaseEvent - pos = eventPos in scene coords?
--- a/src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
+++ b/src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
@@ -153,13 +153,13 @@ void QGIBalloonLabel::mouseReleaseEvent(QGraphicsSceneMouseEvent * event)
 void QGIBalloonLabel::mouseDoubleClickEvent(QGraphicsSceneMouseEvent * event)
 {
     QGIViewBalloon* qgivBalloon = dynamic_cast<QGIViewBalloon*>(parentItem());
-    if (qgivBalloon == nullptr) {
+    if (!qgivBalloon) {
         qWarning() << "QGIBalloonLabel::mouseDoubleClickEvent: No parent item";
         return;
     }
 
     auto ViewProvider = dynamic_cast<ViewProviderBalloon*>(qgivBalloon->getViewProvider(qgivBalloon->getViewObject()));
-    if (ViewProvider == nullptr) {
+    if (!ViewProvider) {
         qWarning() << "QGIBalloonLabel::mouseDoubleClickEvent: No valid view provider";
         return;
     }
@@ -408,13 +408,12 @@ void QGIViewBalloon::updateView(bool update)
 //    Base::Console().Message("QGIVB::updateView()\n");
     Q_UNUSED(update);
     auto balloon( dynamic_cast<TechDraw::DrawViewBalloon*>(getViewObject()) );
-    if( balloon == nullptr )
+    if (!balloon)
         return;
 
     auto vp = static_cast<ViewProviderBalloon*>(getViewProvider(getViewObject()));
-    if ( vp == nullptr ) {
+    if (!vp)
         return;
-    }
 
     if (update) {
         QString labelText = QString::fromUtf8(balloon->Text.getStrValue().data());
@@ -437,15 +436,15 @@ void QGIViewBalloon::updateBalloon(bool obtuse)
 //    Base::Console().Message("QGIVB::updateBalloon()\n");
     (void) obtuse;
     const auto balloon( dynamic_cast<TechDraw::DrawViewBalloon *>(getViewObject()) );
-    if( balloon == nullptr ) {
+    if (!balloon) {
         return;
     }
     auto vp = static_cast<ViewProviderBalloon*>(getViewProvider(getViewObject()));
-    if ( vp == nullptr ) {
+    if (!vp) {
         return;
     }
     const TechDraw::DrawViewPart *refObj = balloon->getViewPart();
-    if (refObj == nullptr) {
+    if (!refObj) {
         return;
     }
 
@@ -476,10 +475,9 @@ void QGIViewBalloon::updateBalloon(bool obtuse)
 
 void QGIViewBalloon::balloonLabelDragged(bool ctrl)
 {
-//    Base::Console().Message("QGIVB::bLabelDragged(%d)\n", ctrl);
     m_ctrl = ctrl;
     auto dvb( dynamic_cast<TechDraw::DrawViewBalloon *>(getViewObject()) );
-    if (dvb == nullptr)
+    if (!dvb)
         return;
 
     if (!m_dragInProgress) {           //first drag movement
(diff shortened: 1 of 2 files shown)


src/Mod/TechDraw/Gui/QGIViewBalloon.h
```diff
@@ -163,7 +163,7 @@ public Q_SLOTS:
     void updateBalloon(bool obtuse = false);
 
 protected:
-    void draw() override;
```
chennes, line 166: I don't know the internal details here, but by not providing the virtual version of `draw()`, you run the risk of breaking anything that was trying to use a base class pointer to access the subclass's `draw()` functions. It might be a good idea to implement the `void draw() override` to have it call your newly-revised function.

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
```diff
@@ -493,29 +492,51 @@ void QGIViewBalloon::balloonLabelDragged(bool ctrl)
...
+        m_originDragged = true;
     }
+}
+
+void QGIViewBalloon::balloonLabelDragFinished()
+{
+    // stores the final drag position for undo
+
+    auto dvb(dynamic_cast<TechDraw::DrawViewBalloon*>(getViewObject()));
+    if (dvb == nullptr)
```
chennes, line 513: Stylistic only: Core Guidelines suggest using `if (dvb)` rather than `if (dvb == nullptr)`. Obviously not critical, but if you end up needing to make any other changes, may as well. (See http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions)

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
```diff
@@ -493,29 +492,51 @@ void QGIViewBalloon::balloonLabelDragged(bool ctrl)
...
+        return;
+
+    double scale = 1.0;
+    DrawView* balloonParent = getSourceView();
+    if (balloonParent != nullptr)
```
chennes, line 518: Ditto previous comment.

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
```diff
@@ -493,29 +492,51 @@ void QGIViewBalloon::balloonLabelDragged(bool ctrl)
     DrawView* balloonParent = getSourceView();
     if (balloonParent != nullptr) {
         scale = balloonParent->getScale();
+        double xPos = Rez::appX(balloonLabel->X() / scale);
```
chennes, line 495: Are these really locals? They look unused.

chennes (commented):
I took a look at the code, but I haven't tested it in action yet. Clearly an important bugfix, so I'll give it a spin next.

src/Mod/TechDraw/Gui/QGIViewBalloon.h
donovaly: I lack knowledge here, so please correct me: There is a base draw() void. Before my patch this one was overriden to use the draw() defined for Balloon. Now I use a different void. I could have given it a completely different name but was lazy. Since I declared it in the private: section, I thought all calls from within QGIViewBalloon.cpp go now to the new draw(bool) void, no?

I can and will now change it as you recommended but would like to understand why this is better.

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
donovaly: Oh, in the past when I just try to fix a bug I took the existing code and changed often its style because the nice Visual Studio IDE informed me about style issues. However, I got feedback that this increased the diff unnecessarily.
Meanwhile I am on your side, also fixes should be used to improve the coding style, so I will change the style now for this PR.

src/Mod/TechDraw/Gui/QGIViewBalloon.h
chennes: My concern was that this class used to provide an overridden `draw()` function, but now no longer does. I didn't look to see if that function did anything, I just saw its removal and was concerned. If you've checked and that function was not needed, then no need to change anything. You are correct that your new function will not be called by a base class call to `draw()`, since your function is non-virtual (and has a different signature besides).

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
chennes: I agree that changes to unrelated lines clutter the diffs, but IMO when you are already changing a particular line, may as well make all of the changes :). Opinions differ, of course.

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
donovaly: Yes. thanks for spotting, is now fixed

src/Mod/TechDraw/Gui/QGIViewBalloon.h
donovaly: thanks for the reply. However, I followed your suggestion and kept now the draw() overriden and use it as redirect.

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
```diff
@@ -608,16 +629,24 @@ void QGIViewBalloon::draw()
...
-    float y = Rez::guiX(balloon->Y.getValue() * refObj->getScale());
+    float x, y;
+    // when not dragging take the X/Y properties otherwise the current label position
+    if (!dragged) {
+        x = Rez::guiX(balloon->X.getValue() * refObj->getScale());
+        y = Rez::guiX(balloon->Y.getValue() * refObj->getScale());
+    }
+    else {
+        x = Rez::guiX(balloonLabel->X() / refObj->getScale());
+        y = -Rez::guiX(balloonLabel->Y() / refObj->getScale());
```
chennes, line 648: The label's `X()` and `Y()` are already GUI-side, so shouldn't this just be:
```suggestion
        x = balloonLabel->X();
        y = -balloonLabel->Y();
```

donovaly: Indeed, thanks. I changed this now.

chennes (approved):
(For the record, I really was only suggesting you make the `nullptr` change to the lines you were already editing, not the whole file -- sorry to make so much work for you!)

All of the questions I had about the code have been resolved. It compiles and runs correctly -- the balloon is positioned correctly both when dragging and when released, and only results in a single "undo" action being created. I tested dragging about a dozen balloons around the screen and was able to undo all of the changes. Thanks for fixing!

src/Mod/TechDraw/Gui/QGIViewBalloon.h
```diff
@@ -183,6 +184,7 @@ public Q_SLOTS:
 
     TechDraw::DrawView* getSourceView() const;
     bool m_dragInProgress;
+    bool m_originDragged = false;
```
aapo-aapo, line 187: This bool m_originDragged is set false only in the class header, i.e. during object construction time.  As it is set true in balloonLabelDragged() and tested later, shouldn't it also be set false after line 534 where m_dragInProgress is set false?

donovaly: > shouldn't it also be set false after line 534

Yes. I'll fix this later.

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
hyarion: I know I've given you feedback before but I think that was when there where changes in otherwise untouched functions. In this case I agree with @chennes as you already touches the line 👍 

If there are a lot of unrelated style changes I want to do in a function I usually try to split it into a separate commit to make it easier on the reviewer, but I guess opinion differ here too :)

src/Mod/TechDraw/Gui/QGIViewBalloon.h
donovaly: fixed

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
```diff
@@ -489,33 +487,50 @@ void QGIViewBalloon::balloonLabelDragged(bool ctrl)
...
+    auto dvb(dynamic_cast<TechDraw::DrawViewBalloon*>(getViewObject()));
+    if (!dvb)
+        return;
+
     double scale = 1.0;
     DrawView* balloonParent = getSourceView();
-    if (balloonParent != nullptr) {
+    if (balloonParent)
         scale = balloonParent->getScale();
-    }
```
hyarion, line 496: this can be rewritten as:
```c++
    if (const auto balloonParent = getSourceView())
        scale = balloonParent->getScale();
```

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
```diff
@@ -525,22 +540,22 @@ void QGIViewBalloon::placeBalloon(QPointF pos)
 //    Base::Console().Message("QGIVB::placeBalloon(%s)\n",
 //                            DrawUtil::formatVector(pos).c_str());
     auto balloon( dynamic_cast<TechDraw::DrawViewBalloon*>(getViewObject()) );
-    if( balloon == nullptr ) {
+    if (!balloon) {
         return;
     }
 
     DrawView* balloonParent = dynamic_cast<DrawView*>(balloon->SourceView.getValue());
```
hyarion, line 548: maybe change this `DrawView*` to `auto`, to make it conform to the style of `balloon`, `featPage `, `vp ` while you're at i

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
```diff
@@ -549,9 +564,9 @@ void QGIViewBalloon::placeBalloon(QPointF pos)
     QPointF viewPos;
     Gui::ViewProvider* objVp = QGIView::getViewProvider(balloonParent);
     auto partVP = dynamic_cast<ViewProviderViewPart*>(objVp);
-    if ( partVP != nullptr ) {
+    if (partVP) {
```
hyarion, line 568:
```c++
    if (auto partVP = dynamic_cast<ViewProviderViewPart*>(objVp)) {
```

hyarion:
My suggestions above are just optional suggestions, feel free to go either way :)

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
donovaly: This is a bugfix PR to be committed for 0.19 to fix an essential Techdraw feature - being able to undo changes.

I only make the code beautification because chennes recommended it. There are some more possible code improvements but not in this bugfix. Please make a new PR for further improvements - for 0.20.

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
donovaly: dito

src/Mod/TechDraw/Gui/QGIViewBalloon.cpp
donovaly: dito.

aapo-aapo (approved):
Now everything compiles & works fine, and I don't see anything wrong in the code.  This is an important fix, so thanks for this!

yorikvanhavre:
Thanks for the fine work guys!

PR #4457, opened 2021-02-15
[0.19] PartDesign: Fix preview bug in Helix

davidosterberg:
There was a bug where the base object became invisible in the preview after the select reference axis option was used. This fixes this.

Current behavior was highlighted as "slightly annoying" in a YouTube tutorial here: https://youtu.be/G6aPsFPGPDQ?t=502

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskHelixParameters.h
+++ b/src/Mod/PartDesign/Gui/TaskHelixParameters.h
@@ -82,7 +82,8 @@ protected:
     void changeEvent(QEvent *e) override;
     bool updateView() const;
     void getReferenceAxis(App::DocumentObject *&obj, std::vector<std::string> &sub) const;
-
+    void startReferenceSelection(App::DocumentObject* profile, App::DocumentObject* base) override;
+    void finishReferenceSelection(App::DocumentObject* profile, App::DocumentObject* base) override;
 
     //mirrors of helixes's properties
     App::PropertyLength*      propPitch;
--- a/src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
@@ -305,7 +305,6 @@ void TaskHelixParameters::updateUI()
 void TaskHelixParameters::onSelectionChanged(const Gui::SelectionChanges& msg)
 {
     if (msg.Type == Gui::SelectionChanges::AddSelection) {
-
         exitSelectionMode();
         std::vector<std::string> axis;
         App::DocumentObject* selObj;
@@ -363,12 +362,15 @@ void TaskHelixParameters::onAxisChanged(int num)
     if (lnk.getValue() == 0) {
         // enter reference selection mode
         TaskSketchBasedParameters::onSelectReference(true, true, false, true);
+        return;
     } else {
         if (!pcHelix->getDocument()->isIn(lnk.getValue())){
             Base::Console().Error("Object was deleted\n");
             return;
         }
         propReferenceAxis->Paste(lnk);
+
+        // in case user is in selection mode, but changed his mind before selecting anything.
         exitSelectionMode();
     }
 
@@ -474,6 +476,32 @@ void TaskHelixParameters::getReferenceAxis(App::DocumentObject*& obj, std::vecto
 
         obj = lnk.getValue();
         sub = lnk.getSubValues();
+    }
+}
+
+void TaskHelixParameters::startReferenceSelection(App::DocumentObject* profile, App::DocumentObject* base)
+{
+    PartDesign::ProfileBased* pcHelix = dynamic_cast<PartDesign::Helix*>(vp->getObject());
+    if (pcHelix->getAddSubType() == PartDesign::FeatureAddSub::Subtractive) {
+        Gui::Document* doc = vp->getDocument();
+        if (doc) {
+            doc->setHide(profile->getNameInDocument());
+        }
+    } else {
+        TaskSketchBasedParameters::startReferenceSelection(profile, base);
+    }
+}
+
+void TaskHelixParameters::finishReferenceSelection(App::DocumentObject* profile, App::DocumentObject* base)
+{
+    PartDesign::ProfileBased* pcHelix = dynamic_cast<PartDesign::Helix*>(vp->getObject());
+    if (pcHelix->getAddSubType() == PartDesign::FeatureAddSub::Subtractive) {
+        Gui::Document* doc = vp->getDocument();
+        if (doc) {
+            doc->setShow(profile->getNameInDocument());
+        }
+    } else {
+        TaskSketchBasedParameters::finishReferenceSelection(profile, base);
     }
 }
 
--- a/src/Mod/PartDesign/Gui/TaskSketchBasedParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskSketchBasedParameters.cpp
@@ -92,6 +92,27 @@ const QString TaskSketchBasedParameters::onAddSelection(const Gui::SelectionChan
     return refStr;
 }
 
+
+void TaskSketchBasedParameters::startReferenceSelection(App::DocumentObject* profile, App::DocumentObject* base)
+{
+    Gui::Document* doc = vp->getDocument();
+    if (doc) {
+        doc->setHide(profile->getNameInDocument());
+        if (base)
+            doc->setShow(base->getNameInDocument());
+    }
+}
+
+void TaskSketchBasedParameters::finishReferenceSelection(App::DocumentObject* profile, App::DocumentObject* base)
+{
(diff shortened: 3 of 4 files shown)


donovaly:
thanks.

chennes:
This compiles, runs, and fixes the bug described in the video. I cannot provide a nicer solution, unfortunately -- I imagine you fought with this one for a bit! While poking at it I did notice that you might want a `return` statement after the call to `onSelectReference()`, since you know at that point you don't have an axis yet. It makes it a little more responsive.

davidosterberg:
> I imagine you fought with this one for a bit! 

Certainly I tried a few different things before I stumbled something that worked! I imagine that the real fix involves actually understanding how the preview mode works, and why it is disturbed by the selection mode.  Since I was too lazy to dig into that, it actually did not take that long :)

davidosterberg:
Ok, found a little bug that I introduced. No biggie. Fixing int now...

wwmayer:
The function TaskSketchBasedParameters::exitSelectionMode() calls onSelectReference() to clear the selection gate and there is a call to hide the base object.

So, far there are only these two sub-classes that use exitSelectionMode(): TaskHelixParameters and TaskRevolutionParameters.
The latter works a bit differently compared to TaskHelixParameters because for a groove it doesn't show the base shape and the transparent red part but it directly shows the final result. So, there it makes sense to hide the base shape as otherwise you won't see the difference.

For the helix it's bad as you only will see the red part.

A more elegant solution is to add two virtual functions: e.g. startReferenceSelection()  and finishReferenceSelection().
As argument both get the pcSketchBased and prevSolid (see onSelectReference) as argument. The default implementation does what TaskSketchBasedParameters and the overridden function in TaskHelixParameters can ignore the prevSolid argument.

davidosterberg:
> The function TaskSketchBasedParameters::exitSelectionMode() calls onSelectReference() to clear the selection gate and there is a call to hide the base object.

This I also found out. What I initially tried was to do the 

```
doc->setShow(prevSolid->getNameInDocument());
```
right after the call to exitSelectionMode. This did not work, because then the red solid disappeared instead :).

wwmayer:
This is what works for me:
```

void TaskSketchBasedParameters::startReferenceSelection(App::DocumentObject* profile, App::DocumentObject* base)
{
    Gui::Document* doc = vp->getDocument();
    if (doc) {
        doc->setHide(profile->getNameInDocument());
        if (base)
            doc->setShow(base->getNameInDocument());
    }
}

void TaskSketchBasedParameters::finishReferenceSelection(App::DocumentObject* profile, App::DocumentObject* base)
{
    Gui::Document* doc = vp->getDocument();
    if (doc) {
        doc->setShow(profile->getNameInDocument());
        if (base)
            doc->setHide(base->getNameInDocument());
    }
}

void TaskSketchBasedParameters::onSelectReference(const bool pressed, const bool edge, const bool face, const bool planar) {
    // Note: Even if there is no solid, App::Plane and Part::Datum can still be selected

    PartDesign::ProfileBased* pcSketchBased = dynamic_cast<PartDesign::ProfileBased*>(vp->getObject());
    if (pcSketchBased) {
        // The solid this feature will be fused to
        App::DocumentObject* prevSolid = pcSketchBased->getBaseObject( /* silent =*/ true );

        if (pressed) {
            startReferenceSelection(pcSketchBased, prevSolid);
            Gui::Selection().clearSelection();
            Gui::Selection().addSelectionGate
                (new ReferenceSelection(prevSolid, edge, face, planar));
        } else {
            Gui::Selection().rmvSelectionGate();
            finishReferenceSelection(pcSketchBased, prevSolid);
        }
    }
}
```

and 
```

void TaskHelixParameters::startReferenceSelection(App::DocumentObject* profile, App::DocumentObject* /*base*/)
{
    Gui::Document* doc = vp->getDocument();
    if (doc) {
        doc->setHide(profile->getNameInDocument());
    }
}

void TaskHelixParameters::finishReferenceSelection(App::DocumentObject* profile, App::DocumentObject* /*base*/)
{
    Gui::Document* doc = vp->getDocument();
    if (doc) {
        doc->setShow(profile->getNameInDocument());
    }
}
```

davidosterberg:
@wwmayer, when you hand it to me like that, then I cannot refuse :). It is longer than my solution. But I agree it is more elegant, because it does not require new member functions in the view provider. I have adapted it, and tested it. I can confirm it works.

src/Mod/PartDesign/Gui/TaskHelixParameters.h
```diff
@@ -82,7 +82,8 @@ private Q_SLOTS:
     void changeEvent(QEvent *e) override;
     bool updateView() const;
     void getReferenceAxis(App::DocumentObject *&obj, std::vector<std::string> &sub) const;
-
+    void startReferenceSelection(App::DocumentObject* profile, App::DocumentObject* /*base*/);
```
wwmayer, line 85: @davidosterberg Can you please add override at the end as otherwise clang will report several warnings?

src/Mod/PartDesign/Gui/TaskHelixParameters.h
```diff
@@ -82,7 +82,8 @@ private Q_SLOTS:
...
+    void finishReferenceSelection(App::DocumentObject* profile, App::DocumentObject* /*base*/);
```
wwmayer, line 86: Same here

src/Mod/PartDesign/Gui/TaskHelixParameters.h
davidosterberg: done

src/Mod/PartDesign/Gui/TaskHelixParameters.h
davidosterberg: done

davidosterberg:
I just realized that @wwmayer solution will not work for additive helix without adding a conditional. I will fix that tomorrow.

davidosterberg:
ok, it became a bit too verbose for my taste. Having to implement both a start and finish method with much the same code. So I simplified by only overriding  onSelectReference instead. I think it is easier to read the code this way, and it is general enough. I hope it is ok with you.

wwmayer:
> I just realized that @wwmayer solution will not work for additive helix without adding a conditional. I will fix that tomorrow.

I don't know what exactly you mean but I realized that in the linked video the problem with the disappearing cylinder when you create an additive helix is not solved by these changes. Since the cylinder immediately disappears without selecting a reference axis before I guess the problem must be somewhere else.

davidosterberg:
> > I just realized that @wwmayer solution will not work for additive helix without adding a conditional. I will fix that tomorrow.
> 
> I don't know what exactly you mean but I realized that in the linked video the problem with the disappearing cylinder when you create an additive helix is not solved by these changes. Since the cylinder immediately disappears without selecting a reference axis before I guess the problem must be somewhere else.

What I mean is that AdditiveHelix is designed to behave like Revolution (i.e no preview). 
Subtractive helix has the preview because it is otherwise easy to loose track of where it is.
 
I don't understand what you mean that the problem is not solved. The problem is solved with these changes as far a I can tell.

wwmayer:
> I don't understand what you mean that the problem is not solved. The problem is solved with these changes as far a I can tell.

There are two problems:
The problem that the preview disappears when selecting a reference axis is solved.

If you watch the video carefully there is an example where a circle and a cylinder is given. As soon as you select the circle and create an additive helix both objects disappear and the helix re-appears as soon as you select the correct reference. This is because in this case  the default vertical axis causes the helix feature to fail.

It only happens for an additive but not subtractive helix and the reason is the following:
When adding a new (solid) object to a body it will hide all previous solids, i.e. the cylinder in this case.
Then the helix will enter edit mode and this is how it's implemented (which is totally correct):
```
bool ViewProviderHelix::setEdit(int ModNum)
{

    if (ModNum == ViewProvider::Default ) {
        auto* prim = static_cast<PartDesign::Helix*>(getObject());
        setPreviewDisplayMode(prim->getAddSubType() == PartDesign::FeatureAddSub::Subtractive);
    }
    return ViewProviderAddSub::setEdit(ModNum);
}
```

The function setPreviewDisplayMode() then makes the base object (i.e. the cylinder) tmp. visible again -- but only for a subtractive helix. The cylinder for an additive helix stays invisible.

A solution to this problem can be that when creating an additive helix we first assign the shape of the base object to the helix because then it has something to show in case its initial recompute fails.

This second issue is independent of this PR but I mentioned t because I saw it in the video when I tried to understand the problematic. In other words the PR can be merged now.

davidosterberg:
@wwmayer , now I get what you mean. This problem is known to me, but at the time I thought that this was just how things worked in PartDesign (only 1 solid allowed).  Now I think I have a way to improve the user experience. I will make a separate PR for that. Thanks.

wwmayer:
Here I have already checked-in an elegant solution: 2661e0ea47

PR #4458, opened 2021-02-15
[0.19] PartDesign: Improve Helix parameter proposals

davidosterberg:
The current initial proposed parameters (pitch and height) can cause self intersection if the selects a different helix axis. In this PR the diagonal of the bounding box is used, guaranteeing that the pitch and height parameters never lead to self intersecting. 

Current behavior was identified in a YouTube tutorial here: https://youtu.be/G6aPsFPGPDQ?t=68

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -504,7 +504,12 @@ double Helix::safePitch()
 void Helix::proposeParameters(bool force)
 {
     if (force || !HasBeenEdited.getValue()) {
-        double pitch = 1.1*safePitch();
+        TopoDS_Shape sketchshape = getVerifiedFace();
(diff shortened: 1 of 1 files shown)


chennes (approved):
For a moment I was confused, I thought "haven't I already looked at this code?" -- because it's basically the same as the through-hole code, right? It's a good use of it here. I wish we could come up with a "safe fillet radius"!

davidosterberg:
> For a moment I was confused, I thought "haven't I already looked at this code?" -- because it's basically the same as the through-hole code, right? It's a good use of it here. I wish we could come up with a "safe fillet radius"!

Hehe, that is so often how it works though. You learn something solving one problem. And that knowledge soon unexpectedlyy becomes useful somewhere else.  I agree about the fillets. The current behavior of the model disappearing when you pick a too large radius does not seem ideal :)

Thanks for reviewing.

PR #4462, opened 2021-02-15
[UI] Add commands to Mesh WB toolbar

bitacovir:
Mesh WB toolbar displays only a few icon commands.

This commit add more commands of the Mesh workbench to the toolbar. Additionally, it arranges these tools in 6 groups (following the current Mesh_Tool_Navi arrangement of the wiki). 

Forum discussion: https://forum.freecadweb.org/viewtopic.php?f=34&t=47494&start=60#p475594


bitacovir:
This is my first time creating toolbars in the workbench.cpp . If a developer can have a look to see if my edition in the code is correct I will appreciated.
Thanks

chennes (approved):
This compiles, runs, and works as described. All icons appear in the toobar correctly, and activate and deactivate as expected. The code looks good to me, I don't see any issues. This is a great improvement to the Mesh WB.

bitacovir:
Yay! Thanks @chennes :)

PR #4464, opened 2021-02-15
3Dconnexion: Clarify ambivalent console output warning [skip ci]

luzpaz:
This commit simply demystifies the warning that most users will see when they boot in to FreeCAD "Couldn't connect to spacenav daemon." by adding 
"Please ignore if you don't have a spacemouse."


luzpaz:
I also added this since most users will have no idea what `spacenav daemon` is. So this commit lets them know it's not important unless they have a spaceball mouse to connect to said daemon.

yorikvanhavre:
Normal users normally don't see the log messages... But this doesn't harm either :)

PR #4469, opened 2021-02-16
[0.20] PartDesign: Allow use of circle edge as axis in PolarPattern and Sketchbased

davidosterberg:
Forum discussion
https://forum.freecadweb.org/viewtopic.php?f=19&t=55573

Demo:
circle_reference

Demo2:
pd_circle_selection_mode

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePolarPattern.cpp
+++ b/src/Mod/PartDesign/App/FeaturePolarPattern.cpp
@@ -26,6 +26,7 @@
 # include <TopoDS.hxx>
 # include <TopoDS_Face.hxx>
 # include <gp_Lin.hxx>
+# include <gp_Circ.hxx>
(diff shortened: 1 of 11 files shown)


wwmayer:
The code looks good and things work as expected.

However, there is one thing that should be cleaned-up sometime:
The function calls ReferenceSelection, onSelectReference and addReferenceSelectionGate now carry up to six boolean as arguments which makes it quite error-prone to set the correct values from the calling instances.

A more elegant solution is to use an enum combined with Qt's QFlags class. This QFlags can replace all the boolean and has two benefits:
* in the calling instances the various enum values must be explicitly written that makes it easier for the reader to understand the purpose
* whenever a new type should be supported only the enum must be extended and the function that handles the new type. The argument list of all other functions doesn't need to be extended any more

PR #4483, opened 2021-02-18
[Path] Test Profile Op fixes and proposed unit test module together -  DO NOT MERGE

Russ4262:
This PR is only for testing purposes.  It includes fixes for the Profile op for 0.19 and a Profile test module also.  I need to test these together because I am unable to reproduce the Linux gcc-7 failures on my machine.

I will submit another PR with only the same Profile operation fixes contained here.

Thank you for creating a pull request to contribute to FreeCAD! To ease integration, we ask you to conform to the following items. Pull requests which don't satisfy all the items below might be rejected. If you are in doubt with any of the items below, don't hesitate to ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10)!

- [ ]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [ ]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [ ]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [ ]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [ ]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [ ]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the [FreeCAD bug tracker issue number](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) in case a particular commit solves or is related to an existing issue on the tracker. Ex: `Draft: fix typos - fixes #0004805`

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.19 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=34586).

---

PR #4338, opened 2021-01-31
[0.20] [Path] Add simple test framework for Profile op

Russ4262:
This PR will fail unit tests until PR #4328 is merged because this PR is a new unit test that includes a test that catches the bug that currently exists with open-edges when very small bit diameters are used.

Thank you for creating a pull request to contribute to FreeCAD! To ease integration, please confirm the following:

- [x] Branch rebased on latest master `git pull --rebase upstream master`
- [ ] Unit tests confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [x] Commit message is [well-written](https://chris.beams.io/posts/git-commit/)
- [ ] Commit message includes `issue #<id>` or `fixes #<id>` where `<id>` is the [associated MantisBT](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) issue id if one exists

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.19 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=34586).

---

PR #4484, opened 2021-02-18
[Path] Fixes for Profile operation [0.19]

Russ4262:
Each commit in this PR addresses an error in the Profile operation.

1. The DraftGeomUtils module is moved to LazyLoader usage
2. The `HandleMultipleFeatures` default value is changed to `Collectively` in order to maintain 0.18 Profile Faces behavior in detecting loops of vertical faces.
3. The fix for profiling open edges using micro bits
4. The fix for the undefined `profileshape` variable when using `HandleMultipleFeatures` is set to `Collectively`
5. A fix for improper filtering/processing of faces and edges when both types are included in the selection

Forum discussion available at [0.19 Profile Op fixes](https://forum.freecadweb.org/viewtopic.php?f=15&t=55740).

There is a testing-purposes-only PR, #4483 that tests these bug fixes in a newly proposed Profile op test module that is independently submitted in PR #4338 .  PR #4483 passed and did so testing the fixes addressed in this PR.  Whereas, PR #4338 is failing because of bugs addressed within this PR here.  Travis log data from the build failures point to the fixes addressed here.  Once this PR here is merged, then I will gladly rebase PR #4338 and allow Travis to have her way with it - it should pass with flying colors (or, so I am quite optimistic since it is the same test module in that PR).

---

### DIFF
--- a/src/Mod/Path/PathScripts/PathProfile.py
+++ b/src/Mod/Path/PathScripts/PathProfile.py
@@ -660,8 +660,9 @@ class ObjectProfile(PathAreaOp.ObjectOp):
                                 else:
                                     PathLog.error(self.inaccessibleMsg)
 
-                            tup = openEdges, False, 'OpenEdge', 0.0, 'X', obj.StartDepth.Value, obj.FinalDepth.Value
-                            shapes.append(tup)
+                            if openEdges:
+                                tup = openEdges, False, 'OpenEdge', 0.0, 'X', obj.StartDepth.Value, obj.FinalDepth.Value
+                                shapes.append(tup)
                         else:
                             PathLog.error(self.inaccessibleMsg)
                     # Eif
--- a/src/Mod/Path/PathScripts/PathProfile.py
+++ b/src/Mod/Path/PathScripts/PathProfile.py
@@ -1337,7 +1337,10 @@ class ObjectProfile(PathAreaOp.ObjectOp):
         # PathLog.debug('_makeStop()')
         rad = self.radius
         ofstRad = self.ofstRadius
-        extra = self.radius / 10
+        extra = self.radius / 5.0
+        lng = 0.05
+        med = lng / 2.0
+        shrt = lng / 5.0
 
         E = FreeCAD.Vector(pB.x, pB.y, 0)  # endpoint
         C = FreeCAD.Vector(pA.x, pA.y, 0)  # checkpoint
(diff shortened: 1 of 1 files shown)


src/Mod/Path/PathScripts/PathProfile.py
```diff
@@ -352,12 +358,11 @@ def areaOpShapes(self, obj):
                 # Efor
             else:
                 stock = PathUtils.findParentJob(obj).Stock
-                for (base, subList) in obj.Base:
+                for (base, subList) in otherBase:
```
mlampert, line 361: This doesn't look right, shouldn't this still be using `obj.Base` ?

src/Mod/Path/PathScripts/PathProfile.py
```diff
@@ -322,12 +324,16 @@ def areaOpShapes(self, obj):
 
         # Pre-process Base Geometry to process edges
         if obj.Base and len(obj.Base) > 0:  # The user has selected subobjects from the base.  Process each.
-            shapes.extend(self._processEdges(obj))
+            edge_shapes, otherBase = self._processEdges(obj)
+            shapes.extend(edge_shapes)
 
-        if obj.Base and len(obj.Base) > 0:  # The user has selected subobjects from the base.  Process each.
+        if otherBase is False:
```
mlampert, line 330: `otherBase` gets initialized with `None`, and now it's compared to `False` ?

it would be better if `otherBase` gets intialized to `[]` and `_processEdges` also returns an empty list - variables that switch types are a recipe for pain.

src/Mod/Path/PathScripts/PathProfile.py
```diff
@@ -322,12 +324,16 @@ def areaOpShapes(self, obj):
...
+            # Edges were already processed, or whole model targeted.
+            PathLog.debug("otherBase is False")
+        elif otherBase and len(otherBase) > 0:  # The user has selected subobjects from the base.  Process each.
             if obj.EnableRotation != 'Off':
-                for p in range(0, len(obj.Base)):
-                    (base, subsList) = obj.Base[p]
+                for p in range(0, len(otherBase)):
+                    (base, subsList) = otherBase[p]
```
mlampert, line 336: I gotta say the old code was more readable, it's pretty clear what `obj.Base` is, it's not clear what `otherBase` is though - makes it hard to understand what's going on.

src/Mod/Path/PathScripts/PathProfile.py
```diff
@@ -671,12 +671,10 @@ def _processEdges(self, obj):
...
-        delPairs.sort(reverse=True)
-        for p in delPairs:
-            # obj.Base.pop(p)
-            pass
-
-        return shapes
+        if newBase:
+            return shapes, newBase
+        else:
+            return shapes, False
```
mlampert, line 677: it's not good practice to return different types - why not return an empty collection `newBase` ?

src/Mod/Path/PathScripts/PathProfile.py
Russ4262: Normally, yes, `obj.Base` would be used, as it was before these changes.  However there are related issues that these changes fix.
.
First, I learned that `obj` attributes should, for the most part, be read-only - so as to not introduce errors into the operation.  
.
Second, because the Profile op is handling edges and faces, it is processing those two feature sets(selections) separately.  Above, near line 325 the `obj.Base` object is pre-processed to extract and process edges.  
I am using the new `otherBase` as a container identical to `obj.Base` to contain features that were not handled in the edge processing section and need to be processed in the next face-processing section.
In the edges processing code, edges are removed from the selection(normally `obj.Base`) and processed into removal shapes, which are returned as part of the tuple from ` self._processEdges(obj)` call.  The other object returned in the tuple is the remaining list of features provided in `obj.Base` that still need processing.  I am storing this remainder in the new `otherBase` object because I don't know of another way to easily flag or mark those edge features within `obj.Base` that are already processed and now need to be ignored.  Creating the `newBase` and corresponding `otherBase` is an easy way to return these unprocessed features without having to filter out already processed edge features.  And, I am not modifying `obj.Base` as stated in my first declaration.
.
Yes, now that you mention it, the `newBase` and `otherBase` should carry the same name so readers see that they are one in the same.  Yes, I am open for a better name, like `remainingBaseGeometry`.  I am open to ideas.
.
The overall fix does work.  From my limited perspective it works fairly well.  Yes, I need to improve readability, and will work to do so.  Ideas to improve?

src/Mod/Path/PathScripts/PathProfile.py
Russ4262: Yes, I will revise and use the same data type throughout uses of `newBase` and related `otherBase`.

Russ4262:
@mlampert 
I have improvements for the `obj.Base`, `newBase`, `otherBase` issue we are discussing.  Should I update the same commit we are discussing and working on, or should I add a new commit to the PR?  
.
I am thinking I should update the current(affected) commit we are discussing, but I don't want to mess up our conversational comments.  Should I add a new commit with the changes, then squash the original with the new when we agree, or just update the current commit we are discussing and working on?
Thanks.

PR #4485, opened 2021-02-18
[0.20] PartDesign: Allow customization of Helix preview modes

davidosterberg:
Introduce two new preferences:

AdditiveHelixPreview (true / false) with default false. Turning this on will create a yellow preview of the helix. Independent of the base object. This is useful as it will give hints to why the helix op fails (e.g the helix does not intersect the base object).
image

SubtractiveHelixPreview (true / false) with default true. Turning this off will make the behavior similar to the groove tool


wwmayer:
Don't repeat yourself (DRY)! I think it's best to move this into a single function instead of implementing 3x.

wwmayer:
Merged as 318972ce70

PR #4490, opened 2021-02-19
[PD] fix dressup feature display when broken

donovaly:
As discussed in https://forum.freecadweb.org/viewtopic.php?p=479647#p479377
the user can end up without either seeing nothing or an inexisting dressup feature.

This PR fixes this. See https://forum.freecadweb.org/viewtopic.php?f=3&t=55735&p=479707#p479707 for details.

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
@@ -102,6 +102,9 @@ TaskChamferParameters::TaskChamferParameters(ViewProviderDressUp *DressUpView, Q
         this, SLOT(setSelection(QListWidgetItem*)));
     connect(ui->listWidgetReferences, SIGNAL(itemDoubleClicked(QListWidgetItem*)),
         this, SLOT(doubleClicked(QListWidgetItem*)));
+
+    // the dialog can be called on a broken chamfer, then hide the chamfer
+    hideOnError();
 }
 
 void TaskChamferParameters::setUpUI(PartDesign::Chamfer* pcChamfer)
@@ -231,6 +234,8 @@ void TaskChamferParameters::onRefDeleted(void)
     pcChamfer->Base.setValue(base, refs);
     // recompute the feature
     pcChamfer->recomputeFeature();
+    // hide the chamfer if there was a computation error
+    hideOnError();
 
     // if there is only one item left, it cannot be deleted
     if (ui->listWidgetReferences->count() == 1) {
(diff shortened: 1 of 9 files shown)


src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
```diff
@@ -389,10 +417,11 @@ TaskDlgChamferParameters::~TaskDlgChamferParameters()
 //}
 bool TaskDlgChamferParameters::accept()
 {
-    parameter->showObject();
     parameter->apply();
 
     return TaskDlgDressUpParameters::accept();
+    // show only if there was no computation error
+    parameter->showObject();
```
chennes, line 424: This code follows an unconditional return - it won't get called.

donovaly: It will get called, but only when there was no computation error and that's the idea.

The use case is the following:
pMx0NcIz63

So when you try to acept an invalid dressup and get an error as return, the broken (actually inexisting dressup) must not be shown.
As you can see this is now the case. And when there is no computation error, the dressup is shown.

chennes: I am only looking at the code: line 422 is:
```
return TaskDlgDressUpParameters::accept();
```
How can any code on following lines possibly execute?

donovaly: > How can any code on following lines possibly execute?

not at all ;-) Stupid mistake by me. Since the feature is already shown when executing the dialog and later also when  a recompute led to no error, showing it is superfluous and in case there is an error we must not show.

I fixed this now.

chennes (approved):
I tested this for a couple of different cases, and it works well: it's a very good improvement over the current behavior. I notice that when first entering the task the feature still disappears for a moment before reappearing, so future work on this might prevent that flash. Purely cosmetic at that point however, this fixes the true usability issue.

src/Mod/PartDesign/Gui/TaskThicknessParameters.cpp
```diff
@@ -285,6 +300,18 @@ int TaskThicknessParameters::getMode(void) const {
     return ui->modeComboBox->currentIndex();
 }
 
+void TaskThicknessParameters::hideOnError() {
+    PartDesign::Thickness* pcThickness = static_cast<PartDesign::Thickness*>(DressUpView->getObject());
+    // in case of an error, e.g. thickness too thin, show the base feature and hide the broken thickness
+    App::DocumentObject* baseFeature = getBase();
+    if (baseFeature) {
+        PartDesignGui::ViewProvider* view = dynamic_cast<PartDesignGui::ViewProvider*>(Gui::Application::Instance->getViewProvider(baseFeature));
```
wwmayer, line 308: This line is superfluous.

wwmayer:
DRY:
There is no need to re-implement hideOnError() for each task dialog separately. This can go to the base class TaskDressUpParameters and you don't need to cast to any of the sub-classes Fillet, Chamfer, ... because you only call the method isError() which is already defined in App::DocumentObject.

wwmayer:
Merged as e0d07257b7

PR #4492, opened 2021-02-19
Draft: Fixes #0004562: Active working plane changes when I close Draft Edit panel

marioalexis84:
To reproduce the reported error, two BuildingParts must be activated consecutively, so that the DraftWorkingPlane.save method is executed and overwrite the DraftWorkingPlane.stored attribute with the information from the previous working plane.
This behavior is not limited to the edit command. The same happens with the rest of the modification commands in Draft but when they are executed twice.
Removing the call to DraftWorkingPlane.restore in draftguitools's finish method solve the problem.
https://tracker.freecadweb.org/view.php?id=4562

---

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_base_original.py
+++ b/src/Mod/Draft/draftguitools/gui_base_original.py
@@ -183,7 +183,6 @@ class DraftTool:
             self.ui.sourceCmd = None
         if self.planetrack:
             self.planetrack.finalize()
-        App.DraftWorkingPlane.restore()
         if hasattr(Gui, "Snapper"):
             Gui.Snapper.off()
         if self.call:
--- a/src/Mod/Draft/draftguitools/gui_edit.py
+++ b/src/Mod/Draft/draftguitools/gui_edit.py
@@ -361,7 +361,6 @@ class Edit(gui_base_original.Modifier):
             self.deformat_objects_after_editing(self.edited_objects)
         
         super(Edit, self).finish()
-        App.DraftWorkingPlane.restore()
         if Gui.Snapper.grid:
             Gui.Snapper.grid.set()
         self.running = False


hyarion:
Would it be possible to add a test case for this bug?
That way it would be easier to test both the issue and fix.

(off topic: I notice an old commit of mine references this PR by misstake. Apperently I made an error which caused it to reference github issue/pr number instead of a mantis reference.. sorry for that)

luzpaz:
+1 for unit test

bitacovir:
Report in the forum with file for test case
https://forum.freecadweb.org/viewtopic.php?f=23&t=54519

marioalexis84:
@hyarion @luzpaz There is no units test for draftguitools (only some related to the import of some modules).

chennes:
IMO, if there is not an existing TestDraftGuiTools target, then adding one is too large to consider for 0.19 (it's quite a lot of CMake changes, and will itself need quite a bit of testing -- I added tests for StartGui a few months ago and am still running across problems with them). If it works, I think we should merge this PR now, and add "Improve DraftGui unit tests" on our 0.20 to-do list.

hyarion:
I'm not too familiar with this part of the code base so I might be completely wrong here.
But to me, the removal of restore calls sounds a bit dangerous, especially when there are a bunch of other `App.DraftWorkingPlane.restore()` calls in the BIM workbench too (which was used to find the bug). Are you sure that you're removing the right ones?

yorikvanhavre:
I think the call to WP.restore() at that time is correct.. I'd prefer it to be called each time, but simply not restore anything if there is nothing to restore. So it would depend on each Draft command to store a WP position or not before beginning.
Maybe what is buggy here is that restore() didn't wipe out saved positions on previous runs... I'll have a look too

marioalexis84:
If another solution is implemented, note that the call to WP.restore() in the finish method of Edit class is unnecessary since it is already called in the finish method of the parent class gui_base_original.Modifier.

carlopav:
Hello guys :) Here I am, sorry for the long absence :)

I checked the code and your commit.
Before I started working of Draft Edit, the working plane did align automatically to the edited object (in proceed() method the lines are still commented).
With multi-object editing this became tricky and I did disabled it, thinking that you can always set the working plane before start editing the object (or also during editing, with [wiki]Draft_SelectPlane[/wiki])
So I think that it's good if we prevent Draft_Edit to do anything with the working plane... 
so ok for me on this PR 

By the way. I also noticed that also [mono]gui_base_original.DraftTool [/mono] restore() the workingplane on finish(). I'm wondering if it's good or not...

yorikvanhavre:
I'll merge this PR now as it works and I still didn't implement anything better...

PR #4494, opened 2021-02-20
[0.19] [FEM] correct sorting order

donovaly:
all other menu items are sorted and we also keep the sorting in the Wiki


wwmayer:
This makes the ordering in the toolbar inconsistent to the ordering in the menu.

donovaly:
Oops, sorry. Now it is correct.

PR #4495, opened 2021-02-20
[0.19] [Part] add missing bindings

donovaly:
For FC 0.19 we added bindings to the PD primitives but we forgot the Part primitives

I think this is trivial enough to go in for 0.19 and it improves the usability.

For the future also the primitive location parameters should be bound (reported as bug https://tracker.freecadweb.org/view.php?id=4573).


chennes (approved):
This compiles, runs, and provides the expected expression binding capability to the shape parameters for the parts I tested. The code is straightforward and unlikely to cause any problem: while this is on a fine line between bug fix and feature, I agree with @donovaly that this could go into 0.19.

PR #4496, opened 2021-02-20
App: Fix highlight bug for children of part container

davidosterberg:
Fixes bug discussed in forum here:
https://forum.freecadweb.org/viewtopic.php?f=3&t=55579

Fixes: issue 4569


donovaly:
already fixed: https://github.com/FreeCAD/FreeCAD/commit/e5e07445dc977e0b80b046c287ac9c542d2969af

davidosterberg:
Well f**k . Would've been nice if that was mentioned in the forum thread before I spent hours tracking this down. Ah well, at least I know how it works now.

PR #4501, opened 2021-02-21
Part: Part.CompoundTools.Explode.explodeCompound access a property of...

luzpaz:
...a sometimes null object with an error

Python code, file `src/Mod/Part/CompoundTools/Explode.py`
The `explodeCompound` function can be called in a context without assigning the `ViewObject` property, for example from a command line script.

The error that is fixed by this patch. 
```
  ...
  File "/opt/freecad/Mod/Part/CompoundTools/Explode.py", line 23, in explodeCompound
    cf.ViewObject.DontUnhideOnDelete = True
AttributeError: 'NoneType' object has no attribute 'DontUnhideOnDelete']
```

Patch submitted by marioamb.  
Fix 0004421  
https://tracker.freecadweb.org/view.php?id=4421

### DIFF
--- a/src/Mod/Part/CompoundTools/Explode.py
+++ b/src/Mod/Part/CompoundTools/Explode.py
@@ -20,7 +20,8 @@ def explodeCompound(compound_obj, b_group = None):
         cf.Base = compound_obj
         cf.FilterType = 'specific items'
         cf.items = str(i)
-        cf.ViewObject.DontUnhideOnDelete = True
+        if cf.ViewObject is not None:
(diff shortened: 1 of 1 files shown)


luzpaz:
Note: have not tested this patch, but it looks correct. Please review.

chennes (approved):
I don't have a failing example to test it on, but the code change itself is basically as safe as safe can be, simply adding a check against `None` before accessing the structure. If it fixes the submitter's bug, it looks good to me.

PR #4502, opened 2021-02-21
[Gui] fix NaviCube edge selection

donovaly:
This PR
- adds missing edge faces -> fixes bug 4495
- implements a preferences option to rotate towards the nearest possible state (like other CAD programs provide)
- to make the edge faces fit, use triangles for the corners
- exposes cube size setting to preferences dialog
- adds new button to view the currently invisible backside:
jO7IDuBgGA
- get rid of transparent cube faces to avoid wrong clicks, fixes issues like this (you want to get the Right side but accidentally click through the cube and hit the backside of a corner):
dcCqWMv1bk
- gets rid of the cube axis label because as reported in https://forum.freecadweb.org/viewtopic.php?f=3&t=52118&start=20#p481781 the x y z labels are not scaled on HiDPI screens
moreover they are often "in they way" when you want to click on a corner and in principle they don't provide more info since we already have the CS with the axis labels

- also fix an arithmetic overflow compiler warning

### DIFF
--- a/src/Gui/DlgSettingsNavigation.cpp
+++ b/src/Gui/DlgSettingsNavigation.cpp
@@ -94,6 +94,7 @@ void DlgSettingsNavigation::saveSettings()
     ui->CheckBox_UseAutoRotation->onSave();
     ui->qspinNewDocScale->onSave();
     ui->prefStepByTurn->onSave();
+    ui->naviCubeToNearest->onSave();
     ui->prefCubeSize->onSave();
 
     bool showNaviCube = ui->groupBoxNaviCube->isChecked();
@@ -119,6 +120,7 @@ void DlgSettingsNavigation::loadSettings()
     ui->CheckBox_UseAutoRotation->onRestore();
     ui->qspinNewDocScale->onRestore();
     ui->prefStepByTurn->onRestore();
+    ui->naviCubeToNearest->onRestore();
     ui->prefCubeSize->onRestore();
 
     ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath
--- a/src/Gui/DlgSettingsNavigation.ui
+++ b/src/Gui/DlgSettingsNavigation.ui
@@ -7,14 +7,14 @@
     <x>0</x>
     <y>0</y>
     <width>500</width>
-    <height>363</height>
+    <height>391</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Navigation</string>
   </property>
-  <layout class="QGridLayout" name="gridLayout_3">
-   <item row="0" column="0">
+  <layout class="QVBoxLayout" name="verticalLayout_2">
+   <item>
     <widget class="QGroupBox" name="groupBoxNaviCube">
      <property name="title">
       <string>Navigation cube</string>
@@ -34,6 +34,9 @@
        <widget class="Gui::PrefSpinBox" name="prefStepByTurn">
         <property name="toolTip">
          <string>Number of steps by turn when using arrows (default = 8 : step angle = 360/8 = 45 deg)</string>
+        </property>
+        <property name="alignment">
+         <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
         </property>
         <property name="minimum">
          <number>4</number>
@@ -108,6 +111,25 @@
         </item>
        </widget>
       </item>
+      <item row="1" column="0" colspan="2">
+       <widget class="Gui::PrefCheckBox" name="naviCubeToNearest">
+        <property name="toolTip">
+         <string>Rotates to nearest possible state when clicking a cube face</string>
+        </property>
+        <property name="text">
+         <string>Rotate to nearest</string>
+        </property>
+        <property name="checked">
+         <bool>true</bool>
+        </property>
+        <property name="prefEntry" stdset="0">
+         <cstring>NaviRotateToNearest</cstring>
+        </property>
+        <property name="prefPath" stdset="0">
+         <cstring>NaviCube</cstring>
+        </property>
+       </widget>
+      </item>
       <item row="2" column="0">
        <widget class="QLabel" name="label_3">
         <property name="text">
@@ -119,6 +141,9 @@
        <widget class="Gui::PrefSpinBox" name="prefCubeSize">
         <property name="toolTip">
          <string>Size of the navigation cube</string>
+        </property>
+        <property name="alignment">
+         <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
         </property>
         <property name="minimum">
          <number>10</number>
@@ -143,7 +168,7 @@
      </layout>
     </widget>
    </item>
-   <item row="1" column="0">
+   <item>
     <widget class="QGroupBox" name="GroupBox12">
      <property name="enabled">
       <bool>true</bool>
@@ -333,7 +358,7 @@ Turntable: the part will be rotated around the z-axis.</string>
           </property>
          </widget>
         </item>
-        <item row="4" column="0">
+       <item row="4" column="0">
          <widget class="QLabel" name="label_2">
           <property name="text">
            <string>New document scale</string>
@@ -341,7 +366,7 @@ Turntable: the part will be rotated around the z-axis.</string>
          </widget>
         </item>
         <item row="4" column="3">
-         <widget class="PrefUnitSpinBox" name="qspinNewDocScale">
+         <widget class="PrefUnitSpinBox" name="qspinNewDocScale" native="true">
           <property name="minimumSize">
            <size>
             <width>120</width>
@@ -527,12 +552,12 @@ Mouse tilting is not disabled by this setting.</string>
           </property>
          </widget>
         </item>
-       </layout>
+        </layout>
       </item>
      </layout>
     </widget>
    </item>
-   <item row="2" column="0">
+   <item>
     <spacer name="verticalSpacer">
      <property name="orientation">
       <enum>Qt::Vertical</enum>
@@ -549,6 +574,11 @@ Mouse tilting is not disabled by this setting.</string>
  </widget>
  <customwidgets>
   <customwidget>
+   <class>Gui::PrefSpinBox</class>
+   <extends>QSpinBox</extends>
+   <header>Gui/PrefWidgets.h</header>
+  </customwidget>
+  <customwidget>
    <class>Gui::PrefCheckBox</class>
    <extends>QCheckBox</extends>
    <header>Gui/PrefWidgets.h</header>
@@ -563,11 +593,6 @@ Mouse tilting is not disabled by this setting.</string>
    <extends>QWidget</extends>
    <header>Gui/PrefWidgets.h</header>
    <container>1</container>
-  </customwidget>
-  <customwidget>
-   <class>Gui::PrefSpinBox</class>
-   <extends>QSpinBox</extends>
-   <header>Gui/PrefWidgets.h</header>
   </customwidget>
  </customwidgets>
  <resources/>
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -945,9 +945,7 @@ void NaviCubeImplementation::handleResize() {
 		m_PrevWidth = view[0];
 		m_PrevHeight = view[1];
 		m_View3DInventorViewer->getSoRenderManager()->scheduleRedraw();
-
 	}
-
 }
 
 void NaviCubeImplementation::drawNaviCube(bool pickMode) {
@@ -1294,6 +1292,12 @@ bool NaviCubeImplementation::mouseReleased(short x, short y) {
 		ParameterGrp::handle hGrpNavi = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/NaviCube");
 		bool toNearest = hGrpNavi->GetBool("NaviRotateToNearest", true);
 
+		// 3D animation (UseAutoRotatio) collides with the rotation to the nearest state
+		// thus when enabled, disable them temprarily for the cube rotation
+		bool UseAutoRotation = hGrp->GetBool("UseAutoRotation", false);
+		if (UseAutoRotation && toNearest)
+			m_View3DInventorViewer->setAnimationEnabled(false);
+
 		switch (pick) {
 		default:
 			return false;
@@ -1624,6 +1628,10 @@ bool NaviCubeImplementation::mouseReleased(short x, short y) {
 			handleMenu();
 			break;
 		}
+
+		// re-anable UseAutoRotation after the cube rotation was done
+		if (UseAutoRotation && toNearest)
+			m_View3DInventorViewer->setAnimationEnabled(true);
 	}
 	return true;
 }
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -196,14 +196,14 @@ private:
 	void setHilite(int);
 
 	void initNaviCube(QtGLWidget*);
-	void addFace(const Vector3f&, const Vector3f&, int, int, int, int,bool flag=false);
+	void addFace(const Vector3f&, const Vector3f&, int, int, int, bool flag=false);
 
-	GLuint createCubeFaceTex(QtGLWidget*, float, float, const char*);
+	GLuint createCubeFaceTex(QtGLWidget* gl, float gap, const char* text, int shape);
 	GLuint createButtonTex(QtGLWidget*, int);
 	GLuint createMenuTex(QtGLWidget*, bool);
 
 	void setView(float ,float );
-	void rotateView(int ,float );
+	void rotateView(int axis, float rotAngle, SbVec3f customAxis = SbVec3f(0, 0, 0));
 
 	QString str(char* str);
 	char* enum2str(int);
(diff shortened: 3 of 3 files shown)


wwmayer:
There are two problems:
* clicking on one of the corners shows a different behaviour now
* if the cube has arbitrary orientation and you click on an edge it doesn't align the cube (as it does when clicking on a face)

donovaly:
>     * clicking on one of the corners shows a different behaviour now

Strange, since I haven't touched their rotation code yet. I'll have a look the next days.

>     * if the cube has arbitrary orientation and you click on an edge it doesn't align the cube (as it does when clicking on a face)

That is what I meant with we need to make a decision on how the edge should rotate. I implemented now the logic used by e.g. CAD Assistant. For more info, see my forum post: https://forum.freecadweb.org/viewtopic.php?f=3&t=52118&start=10#p480628

Despite the cube behaves now as it should (despite the corners), it look pixeled. Maybe one could turn on anti-aliasing.

donovaly:
In my opinion, the PR is now ready for a review (to be merged for 0.20)

chennes:
Nice set of changes, I'll give it a try next week. Is it possible to make the cube obey your antialiasing settings?

donovaly:
>  Is it possible to make the cube obey your antialiasing settings?

What do you mean?
I tried out to draw the cube with antialialising on but there was no visible benefit. However, I could not test if there would be a benefit on HiDPI screens.

chennes:
> I tried out to draw the cube with antialialising on but there was no visible benefit. 

That's what I mean - it's not being antialiased, even if you turn AA on in your settings. I'd guess it's being rendered totally separately and then comped on top, but I haven't looked at the code.

It's not important for HiDPI screens, where antialiasing isn't really needed, it's low-resolution displays that will show the most benefit. I find on some low-resolution displays that the text is difficult to read, and often antialiasing will help with that.

donovaly:
> That's what I mean - it's not being antialiased, even if you turn AA on in your settings.

It's OK to work on that but this PR does not change the AA. There are different issues with the NaviCube and here I only make the first step to improve the situation.

Besides this I tested it on my old Laptop with small screen and no Hi-DPI and the NaviCube as of FC 0.19 looks OK.

luzpaz:
Can https://tracker.freecadweb.org/view.php?id=4041 also be addressed in this PR ?

donovaly:
> Can https://tracker.freecadweb.org/view.php?id=4041 also be addressed in this PR ?

No.
However, there is a pending feature request to change the cube labeling text and I plan to implement this. However, bug 4041 occurs because the user misses a font and the font replacement mechanism fails to find a substitution. There is not much we can do other than to use "osifont" because this font is shipped with FC and thus always available. I will give this a try in another PR.

wwmayer:
What is the "Rotate to nearest" option good for? And why it's on by default?
When clicking on a corner, edge or face of the cube it leads to unexpected positions in many situations.

One example:
If Top is the front face and you click on the top edge it turns the cube upside down. If I click on the bottom edge then it rotates by 45 deg but in the wrong direction.

Kh0rn3:
Is there any change of a drag-the-cube-to-rotate-the-view option like in Blender or commercial CAD packages?

hyarion:
> Is there any change of a drag-the-cube-to-rotate-the-view option like in Blender or commercial CAD packages?

This would be really nice addition (that was the first thing I tried when I opened up FreeCAD for the first time)... However...
I think it's important to limit the scope of Pull Requests. This one is already quite big.

@donovaly I'm sorry for introducing merge conflicts with this PR. Do you want to solve the conflicts yourself or would you like some assistance?

donovaly:
> What is the "Rotate to nearest" option good for? And why it's on by default?

As developed in this forum thread: https://forum.freecadweb.org/viewtopic.php?f=3&t=52118
it is annoying to have the view always rotated. When one clicks on e.g. the Top face, one wants to see this face, but not also rotate it always to the same position.
E.g. CAD Assistant has also a NaviCube and it is super useful to rotate only to the nearest possible state.
(Personally the unwanted rotations are annoying when you don't have time, e.g. have to inspect a STEP during a phone call.)

> If Top is the front face and you click on the top edge it turns the cube upside down. If I click on the bottom edge then it rotates by 45 deg but in the wrong direction.

You mean you click the upper edge (Edge Top-Front) and then this edge is not brought to Front?
I am at the moment busy with the FC 0.19 build and will have a look ASAP. Do you have a screencast for me?

donovaly:
> @donovaly I'm sorry for introducing merge conflicts with this PR. Do you want to solve the conflicts yourself or would you like some assistance?

No problem, now this PR has the attention I wanted ;-)
I can fix the merge conflicts, but I would appreciate testing the PR and feedback in the forum thread: https://forum.freecadweb.org/viewtopic.php?f=3&t=52118

hyarion:
@donovaly I'll take a look after you've rebased it then :)

wwmayer:
> You mean you click the upper edge (Edge Top-Front) and then this edge is not brought to Front?
I am at the moment busy with the FC 0.19 build and will have a look ASAP. Do you have a screencast for me?

When the "Top" face points to the user and you click on the edge above the face (it's the common edge between Top and Rear) it turns it upside down. If you click on the edge below (it's the common edge between Top and Front) then it rotates the cube by 45 degree but in the wrong direction.

donovaly:
> @donovaly I'll take a look after you've rebased it then :)

Done

donovaly:
> When the "Top" face points to the user and you click on the edge above the face (it's the common edge between Top and Rear) it turns it upside down. If you click on the edge below (it's the common edge between Top and Front) then it rotates the cube by 45 degree but in the wrong direction.

I cannot reproduce this. It works correct for me, the clicked face comes correctly to the front.
71QvFL933a

donovaly:
> I cannot reproduce this. 

OK, I see now what you mean, but this is by convention because we had to ind one. In the forum we came to the conclusion that the side view are preferred to Top/Bottom. Without the "to nearest" option, we have only one single rotation end state for every face. Now we have a cube where the Top and Rear face descriptions are rotated by 180 to each other:
FreeCAD_FYQ6CBywl6

So clicking the Rear-Top edge must bring this edge face to front, but in what end state? With our convention Rear is preferred over Top, so the edge face will be rotated that you can read "Rear" unrotated.

However these annoying and often unwanted rotations are omitted by the new default "to nearest" because then it is checked what your start point was. Coming from Top, will rotate so that you can read "Top" unrotated, while coming from Rear you can read "Rear" unrotated:
qivDPd1Qxy

hyarion (changes requested):
I've tested the latest version and I think this is an improvement to how the old cube worked. 👍 
I've also read through the discussion on the linked forum thread and this implementation follows the discussion held there. 👍 

There are some few minor things I would like to be addressed:
1. **Backfaces**: I don't think it makes sense to keep the back faces with this new, gap-less, design. I think they should be removed (holds true even with semi-transparent colors).
2. **Front face default color**: I think the default opaque gray color looks a bit out dated and it also clashes with the colors of the buttons surrounding the cube. I think it would look a lot better if you changed it to something like `QColor(226, 233, 239, 192)` (with the backfaces removed)

This is how it looks with the suggestions above:
image

donovaly:
> There are some few minor things I would like to be addressed:

Thanks for testing. I followed your suggestions and changed the color and removed the backfaces.

hyarion:
> I followed your suggestions and changed the color and removed the backfaces.

I noticed that some more backface related code can be removed. I've created a PR against your branch with some suggestions (I thought it would be easier that way)
The PR can be found here: https://github.com/donovaly/FreeCAD/pull/3

donovaly:
> I noticed that some more backface related code can be removed. I've created a PR against your branch 

Thanks. I merged it into this PR.

I now also merged the commits so this PR is ready to be merged to master.

hyarion (approved):

wwmayer:
> I cannot reproduce this. It works correct for me, the clicked face comes correctly to the front.

I have tried your latest changes on Windows and Linux and for none of them it works as expected. The issue I have described above is not solved for me.

wwmayer:
Another problem is that the cube looks very ugly now especially when clicking on one of the corners so that you can see three of its faces. The cube as it's in master just looks fine.

I have installed CAD Builder now and here the cube and text look good and also the behaviour when clicking on faces, edges or corners is very intuitive.

wwmayer:
BTW the tool CAD Assistant is available for Linux and it has a view cube, too. Visually it looks a bit different to that of the CAD Builder but functionality-wise it behaves exactly the same.

wwmayer:
> I cannot reproduce this. It works correct for me, the clicked face comes correctly to the front.

I have found the reason. It happens when the option "Enable animation" is activated. So, it's clearly a regression compared to the current implementation that doesn't suffer from this problem.

donovaly:
> I have tried your latest changes on Windows and Linux and for none of them it works as expected. The issue I have described above is not solved for me.

Have you read my post?: https://github.com/FreeCAD/FreeCAD/pull/4502#issuecomment-797849815

As explained what you see is correct since geometrically there will always be two cube faces where one face has the label turned in respect to the second face.
Hyarion also had a look and could approve that the cube rotates as it should.

> I have found the reason. It happens when the option "Enable animation" is activated. So, it's clearly a regression compared to the current implementation that doesn't suffer from this problem.

This is what I described and that's the reason why the "to nearest" option is on by default. Because without it and clicking on an edge face you get the nasty rotations for faces whose labels are turned  in respect to each other. So without the "to nearest" of course also animation makes the same additional rotations and without it.

I spend a lot of time in the rotation logic and cannot see how edge face support could be done else. Geometry is the limit here. So when a user purposely turn off the "to nearest" option he knows what he is doing and that he will then get additional rotations.

> Another problem is that the cube looks very ugly now especially when clicking on one of the corners so that you can see three of its faces. The cube as it's in master just looks fine.

Can you please post a screencast because the 2 other persons who tested are fine with it and also for me I don't see an issue. So to be able to change something, I need to know what you see.

donovaly:
> So, it's clearly a regression compared to the current implementation that doesn't suffer from this problem.

Maybe we have a misunderstanding: The current implementation has no edge faces and this PR adds them. So how can you compare edge face and the "to nearest" behavior to the current implementation? I checked that when you torn off "to nearest" and don't click on an edge face, it works as the current implementation. Maybe I missed something, then please tell me a recipe to reproduce.

hyarion:
@donovaly when you look perpendicular onto a face it kinda looks like there are edges. But it's actually the backside of one of the connecting sides:
navicube_old
@wwmayer Is this what you mean?

New behaviour as refence:
navcube_new

The old "edge"-clicking behaviour feels more like a happy coincidence than a feature. This as it only works when you look straight ahead on a face.
I think the new behaviour is better and you don't accidentally click through the cube to a side you didn't want to go to.

@donovaly I've found two issue that I think needs to be fixed:
navicube_bug
and
navicube_bug2
Could it be some kind of gimbal-lock issue maybe?

wwmayer:
FYI: Today I stumbled over the view cube of the official OpenInventor: https://www.openinventor.com/en/features/oil-gas-geosciences/viewing-cube (The icons of the shown menu look very familiar :) )

And here the class declaration: https://developer.openinventor.com/refmans/latest/RefManCpp/class_so_viewing_cube.html

hyarion:
> FYI: Today I stumbled over the view cube of the official OpenInventor: https://www.openinventor.com/en/features/oil-gas-geosciences/viewing-cube (The icons of the shown menu look very familiar :) )

haha lol

> And here the class declaration: https://developer.openinventor.com/refmans/latest/RefManCpp/class_so_viewing_cube.html

Interesting, maybe we could take a similar approach and I like their style options.
@donovaly what do you think?

donovaly:
> when you look perpendicular onto a face it kinda looks like there are edges. But it's actually the backside of one of the connecting sides:

Yes, and exactly this was why I once started the action. I wanted to view the scene ad the edge but ended up like in your example by clicking "through" the cube. And if clicking-through or not depends on a few pixels precision.

> I think the new behaviour is better and you don't accidentally click through the cube to a side you didn't want to go to.

Perfectly described ;-) Thanks.

> @donovaly I've found two issue that I think needs to be fixed:
> navicube_bug

I cannot reproduce this:
wntALNpIC9

> navicube_bug2

Also not this one:
QXYFAnra0M

> Could it be some kind of gimbal-lock issue maybe?

I don't know how and in your examples there is still a rotation.

donovaly:
> And here the class declaration: https://developer.openinventor.com/refmans/latest/RefManCpp/class_so_viewing_cube.html

We should not reinvent the wheel. I had a look and cannot find that this source code is available to be used directly in FreeCAD, or did I miss something?
So if it is not available as OpenSource, we need our own solution, otherwise I would like to use the SoViewingCube class.

Their cube is nicely customizable, however, for now I would be happy to have at least one layout but with the features I need:
- viewing at edges
- no click-through the cube
- no additional rotations but to rotate to nearest

Other things like e.g. the bad label font quality, customizable face colors etc. can be done in a next step.

donovaly:
> I cannot reproduce this:

Now I can. This is the remaining bug when you sue the animation option. I am working now on this.

donovaly:
I fixed the issue when the animation option is on:

SO's auto rotation has its own rotation logic that ends at fixed positions. This collides with the "to nearest" option whose aim is to rotate to the next sensible state taking into account from what position the rotation was started.

Thus the solution is to disable SO's auto rotation while the cube is rotate.

@hyarion please test again with the 2 cases where you saw errors. I think now all issues are fixed.

What remains for a further PR is the label font resolution. Werner pointed me to this and depending on the cube size the font is also on Windows badly visible. Despite this has nothing to do with this PR and exists also with the current cube, it should be fixed. Any idea is welcome how this can be achieved.

hyarion:
> I fixed the issue when the animation option is on:
> 
> SO's auto rotation has its own rotation logic that ends at fixed positions. This collides with the "to nearest" option whose aim is to rotate to the next sensible state taking into account from what position the rotation was started.
> 
> Thus the solution is to disable SO's auto rotation while the cube is rotate.

I don't quite like that it has to be disabled as I find the animation part of it useful.
I think the issue is that the code first sets the orientation then rotates the view in a secondary step and that the secondary step (rotate view) takes override the first one (as the first is delayed and hasn't finished yet)

Could you try creating an `sbRotation` quaternion which combines the operations of both the setView and rotateView calls and then set that using `m_View3DInventorViewer->setCameraOrientation(sbRotation orientation)`?

This file over complicate things by using euler angles and transformation matrices when quaternions could be used directly instead.

> @hyarion please test again with the 2 cases where you saw errors. I think now all issues are fixed.

Disabling the animation removed the issue but as mentioned above, I think it could be solved in a better way.

> 
> What remains for a further PR is the label font resolution. Werner pointed me to this and depending on the cube size the font is also on Windows badly visible. Despite this has nothing to do with this PR and exists also with the current cube, it should be fixed. Any idea is welcome how this can be achieved.

I agree that this is out of scope of this pull request.

> > And here the class declaration: https://developer.openinventor.com/refmans/latest/RefManCpp/class_so_viewing_cube.html
> 
> We should not reinvent the wheel. I had a look and cannot find that this source code is available to be used directly in FreeCAD, or did I miss something?
> So if it is not available as OpenSource, we need our own solution, otherwise I would like to use the SoViewingCube class.

I also looked into this and I think it's only available in the commercial version. The free version doesn't seem to have got much love lately :(

> Their cube is nicely customizable, however, for now I would be happy to have at least one layout but with the features I need:
> 
>     * viewing at edges
>     * no click-through the cube
>     * no additional rotations but to rotate to nearest
> 
> Other things like e.g. the bad label font quality, customizable face colors etc. can be done in a next step.

Indeed, lets not feature-creep this PR.
In the long run I think it would be good to rewrite this viewing cube to use the openinventor/coin3d api so we don't depend on OpenGL directly (people are moving away from it to vulcan and similar apis). If that is done it might be good to model it after their implementation and maybe even send it upstream to coin3d.

wwmayer:
> We should not reinvent the wheel. I had a look and cannot find that this source code is available to be used directly in FreeCAD, or did I miss something?

It's not about re-inventing the wheel. The idea is to make a proper SoNode from the NaviCube because this has some advantages. But this shouldn't be handled by this PR anyway.

> Other things like e.g. the bad label font quality, customizable face colors etc. can be done in a next step.

Sure.

> Despite this has nothing to do with this PR and exists also with the current cube, it should be fixed. Any idea is welcome how this can be achieved.

An idea is to check the anti-aliasing option of QPainter

> I don't quite like that it has to be disabled as I find the animation part of it useful.

The current "fix" is just an ugly hack and we must find a proper solution for it. I agree that the animation is an important feature for the user not to lose the overview when changing the viewing direction.

donovaly:
> > Thus the solution is to disable SO's auto rotation while the cube is rotate.
> 
> I don't quite like that it has to be disabled as I find the animation part of it useful.

But I did not disable animations, see my commit. They still work fine.

The point is that the cube rotation invokes several rotation steps and in each step the auto-rotation (animation) performs an action too.
But while the cube is rotated, there is no need for an additional auto-rotation. So animations are only disabled when the cube is rotated and turned on after the cube rotation is finished. Therefore animations still work as they should. If not, please show me a testcase that will now fail.

donovaly:
> The current "fix" is just an ugly hack and we must find a proper solution for it.

It is no hack (see also my previous comment). We rotate the cube they way we want it to rotate. So we explicitly don't want an auto-rotation but only out rotation. Therefore it is in my opinion the correct solution to turn-off the auto-rotation for the time in which we rotate the cube.

wwmayer:
Merged.

> The point is that the cube rotation invokes several rotation steps and in each step the auto-rotation (animation) performs an action too.

The problem is that getting and setting the camera rotation is done inside the functions and if the animation is enabled the rotation is taken at a moment where the animation is still active and thus has a wrong value.

To make this working the current rotation should be read at the beginning, then the additional rotations concatenated and then set at the end. This is done with f525904c1b

donovaly:
> For me as Todo: the corner "Left-Rear-Bottom" does not work. I'll have a look later today.

Here is the PR: #4692
(It was just a copy/paste issue of commit f525904.)

wwmayer:
Oops.

PR #4504, opened 2021-02-21
[0.20] PartDesign: Performance improvements in FeatureTransformed

davidosterberg:
This PR improves performance significantly for larger patterns.

**Forum thread**
https://forum.freecadweb.org/viewtopic.php?f=19&t=55860

**Benchmark model 1:**
[knurls.zip](https://github.com/FreeCAD/FreeCAD/files/6018216/knurls.zip)
partdesign_knurling

**Before**
commit  20d513cd6674 (Baseline)
Time until UI becomes responsive after recompute: 3 min 45s

**After**
Parallel computation + Fuzzy fuse/cut + compound
Time until UI becomes responsive after recompute: 56 s

### DIFF
--- a/src/Mod/PartDesign/App/FeatureTransformed.cpp
+++ b/src/Mod/PartDesign/App/FeatureTransformed.cpp
@@ -36,6 +36,9 @@
 # include <Bnd_Box.hxx>
 #endif
 
+#ifndef FC_DEBUG
+#include <ctime>
+#endif
 
 #include "FeatureTransformed.h"
 #include "FeatureMultiTransform.h"
@@ -198,7 +201,11 @@ short Transformed::mustExecute() const
 
 App::DocumentObjectExecReturn *Transformed::execute(void)
 {
-    rejected.clear();
+
+#ifndef FC_DEBUG
+    std::clock_t start0;
+    start0 = std::clock();
+#endif
 
     std::vector<App::DocumentObject*> originals = Originals.getValues();
     if (originals.empty()) // typically InsideMultiTransform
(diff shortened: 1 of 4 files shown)


wwmayer:
Test testMirroredOffsetFailureCase fails

davidosterberg:
> Test testMirroredOffsetFailureCase fails

This is to be expected. I needed to change the behavior when transformed shapes are not intersecting the base object. See forum thread. I have removed the test.

src/Mod/PartDesign/App/FeatureTransformed.cpp
```diff
@@ -276,132 +285,97 @@ App::DocumentObjectExecReturn *Transformed::execute(void)
...
+        } else {
+            std::unique_ptr<BRepAlgoAPI_BooleanOperation> mkBool(new BRepAlgoAPI_Cut(current, toolShape));
+            if (!mkBool->IsDone()) {
+                std::stringstream error;
+                error << "Boolean operation failed";
+                return new App::DocumentObjectExecReturn(error.str());
             }
+            current = mkBool->Shape();
         }
+        try {} catch (Standard_Failure& e) {
```
wwmayer, line 353: What is this empty try block good for?

src/Mod/PartDesign/PartDesignTests/TestMirrored.py
```diff
@@ -69,24 +69,6 @@ def testMirroredPrimitiveCase(self):
         self.Doc.recompute()
         self.assertAlmostEqual(self.Mirrored.Shape.Volume, 2.0)
 
-    def testMirroredOffsetFailureCase(self):
```
wwmayer, line 72: We don't have to drop this test function. Instead it can be renamed to testMirroredOffsetCase and "Invalid" must be changed to "Up-to-date".

wwmayer:
Merged.

PR #4505, opened 2021-02-21
[0.19] Draft: Fix regression in PathTwistedArray

davidosterberg:
This fixes a bug reported 
https://forum.freecadweb.org/viewtopic.php?f=3&t=55861

The bug was introduced here: 43feb585ad7


chennes (approved):
This PR compiles, runs, and fixes the reported issue.

PR #4508, opened 2021-02-22
[0.19] [TD] fix balloon dragging regression

donovaly:
Fixes bug that dragging while pressing CTRL does not work anymore, see https://forum.freecadweb.org/viewtopic.php?f=35&t=55863


chennes (approved):
This compiles, runs, and fixes the issue with the Control key for single items. Handling multiple-selection correctly is a feature/bugfix for another PR I think.

PR #4512, opened 2021-02-22
[GUI] Make UrlLabel in About styleable

chennes:
Right now the `UrlLabel` in the About FreeCAD box is not styleable because its rich-text contents overrides any styling applied to it. This commit converts it to a plain text label that can then be styled in a stylesheet using the `Gui--UrlLabel` selector. If no stylesheet is applied, the `UrlLabel` styles itself using the old-school blue text with an underline, but any applied stylesheet completely overrides this default.

This does not affect the "clickability" of a `UrlLabel` widget, which is handled independently of the label's content (including its textual hyperlink, which was and still is ignored).

I have not marked a specific version target for this, considering the late date, but I do think it would be nice if we can get it into 0.19.

Thanks to @luzpaz and @hyarion for their assistance in tracking this down.

### DIFF
--- a/src/Gui/AboutApplication.ui
+++ b/src/Gui/AboutApplication.ui
@@ -115,7 +115,10 @@
            <item>
             <widget class="Gui::UrlLabel" name="labelAuthor">
              <property name="text">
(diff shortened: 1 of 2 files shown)


luzpaz:
Related forum threads:  
https://forum.freecadweb.org/viewtopic.php?f=34&t=55844&p=480924#p480924  
https://forum.freecadweb.org/viewtopic.php?f=34&t=50744#p436008

yorikvanhavre:
I think it's best to keep it for 0.20 as it would change a translation string

chennes:
> I think it's best to keep it for 0.20 as it would change a translation string

0.20 is fine, but I don't think this should change a translation string: in the .ui file it's marked as `notr` since we do a search-and-replace on it.

PR #4517, opened 2021-02-23
[ARCH] Arch_Roof fix for adjust_list_len function

Roy-043:
LGTM code analysis did not like that the adjust_list_len function would mutate default list values. The function now always clones the list.
See: https://forum.freecadweb.org/viewtopic.php?f=23&t=55943


luzpaz:
@wwmayer this should be good to go, right? (Even though the CI fails, which looks like a totally different issue)

chennes:
That's using the older CMake stuff - syncing it with master will probably clear up the Travis issue.

wwmayer:
> @wwmayer this should be good to go, right? (Even though the CI fails, which looks like a totally different issue)

I do not handle Arch or Draft PRs as it's up to Yorik to accept or reject the changes.

yorikvanhavre (approved):
Looks OK.. Just redoing the CI tests

luzpaz:
@yorikvanhavre CI is still broken.  
```
Your build has been stopped.
Libpack NOT found.
```
This should be a straightforward merge. Perhaps we can skip the CI test for now?

wwmayer:
It should work as soon as it's merged into master. But it's up to @yorikvanhavre to accept or reject the changes.

PR #4519, opened 2021-02-23
Crowdin: Gui: move whitespace out of translation string [skip ci]

luzpaz:
ref: https://crowdin.com/translate/freecad/549/en-en?filter=basic&value=0#6577624

### DIFF
--- a/src/Gui/DlgUnitsCalculatorImp.cpp
+++ b/src/Gui/DlgUnitsCalculatorImp.cpp
@@ -142,7 +142,7 @@ void DlgUnitsCalculator::valueChanged(const Base::Quantity& quant)
     // since it expects then a scientific notation number like "1e3"
     if ( (ui->UnitInput->text().mid(0, 2) == QString::fromLatin1("ee")) ||
         Base::Unit(ui->UnitInput->text()).getTypeString().isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/Gui/DlgUnitsCalculatorImp.cpp
```diff
@@ -142,7 +142,7 @@ void DlgUnitsCalculator::valueChanged(const Base::Quantity& quant)
     // since it expects then a scientific notation number like "1e3"
     if ( (ui->UnitInput->text().mid(0, 2) == QString::fromLatin1("ee")) ||
         Base::Unit(ui->UnitInput->text()).getTypeString().isEmpty()) {
-        ui->ValueOutput->setText(tr("unknown unit: ") + ui->UnitInput->text());
+        ui->ValueOutput->setText(tr("unknown unit:") + " " + ui->UnitInput->text());
```
luzpaz, line 145: Does this look kosher?

davidosterberg: Looks OK to me

davidosterberg (approved):

wwmayer:
This won't work because we have disabled the constructor of QString that accepts const char*
So, the alternative way is: 879ab9ed84

PR #4532, opened 2021-02-27
[PD] fix sorting error

donovaly:
as reported here: https://forum.freecadweb.org/viewtopic.php?f=3&t=56065#p482274

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePocket.cpp
+++ b/src/Mod/PartDesign/App/FeaturePocket.cpp
@@ -61,8 +61,6 @@ PROPERTY_SOURCE(PartDesign::Pocket, PartDesign::ProfileBased)
 
 Pocket::Pocket()
 {
-    addSubType = FeatureAddSub::Subtractive;
-
     ADD_PROPERTY_TYPE(Type,((long)0),"Pocket",App::Prop_None,"Pocket type");
     Type.setEnums(TypeEnums);
     ADD_PROPERTY_TYPE(Length,(100.0),"Pocket",App::Prop_None,"Pocket length");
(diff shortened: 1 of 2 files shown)


donovaly:
@davidosterberg , I guess this PR is no longer necessary after commit 16ff933b09176d83fb0785e74f36038311f0eaeb right?

davidosterberg:
I am not sure that this PR fixes the issue discussed in the forum thread

donovaly:
> I am not sure that this PR fixes the issue discussed in the forum thread

My PR fixes the issue since now e.g. the pocket Features are correctly sorted: 
FreeCAD_xPYSqkaI8r
(Base -> PartDesign -> Pocket)

After I made the PR I pulled latest master and now also in master the sorting is correct. Therefore I think your commit fixed this. And indeed, reverting changes from your commit the sorting would be wrong in master.

So when you could confirm that current master does no longer show the problem reported in the forum, I can close my PR.

davidosterberg:
> > I am not sure that this PR fixes the issue discussed in the forum thread 
> My PR fixes the issue since now e.g. the pocket Features are correctly sorted:

Maybe I misunderstood the wish in the forum thread. I thought that Kisolre wanted the properties under the "Part Design" headline. I.e Refine. To appear at the bottom of the properties list. This is clearly not the case in your screenshot. My explanation is that P comes before S, therefore the Part Design properties will be listed before the Sketch Based properties.

wwmayer:
> I guess this PR is no longer necessary after commit 16ff933 right?

Yes. So, it can be closed.

> My PR fixes the issue since now e.g. the pocket Features are correctly sorted:

No, your changes do not affect the sorting at all but it adds a regression by declaring the revolution feature as subtractive which is clearly wrong. A revolution always adds material to a solid and its counterpart that removes material is called groove.

About the sorting David is right that the groups are sorted alphabetically but inside the group the elements are in the same order as the properties are added to a feature.

donovaly:
> your changes do not affect the sorting at all 

I would like to understand this because because I shifted existing code "addSubType = FeatureAddSub::Subtractive;" below the new features in Pocket.
This changed for me the sorting and afterwards I got the section "PartDesign" above "Pocket" and that was the goal.

> but it adds a regression by declaring the revolution feature as subtractive

Concerning the revolution, this was just a paste mistake.

wwmayer:
> This changed for me the sorting and afterwards I got the section "PartDesign" above "Pocket" and that was the goal.

The property editor doesn't know anything about specific things of the extension modules (i.e. the addSubType here). What it does is to retrieve the group name of a property and put that into a std::map. The std::map then sorts the content alphabetically and that's the property editor shows them this way.

donovaly:
thanks

PR #4534, opened 2021-02-28
[OpenSCAD] Add unit test framework to module

chennes:
The OpenSCAD module did not have any unit tests. This commit adds the basic framework (and two very simple tests). The framework is modeled after the unit tests in the FEM module.

### DIFF
--- a/src/Mod/OpenSCAD/CMakeLists.txt
+++ b/src/Mod/OpenSCAD/CMakeLists.txt
@@ -15,8 +15,44 @@ SET(OpenSCAD_SRCS
     colorcodeshapes.py
     expandplacements.py
     replaceobj.py
+    TestOpenSCADApp.py
 )
+IF (BUILD_GUI)
+	LIST(APPEND OpenSCAD_SRCS TestOpenSCADGui.py)
+ENDIF (BUILD_GUI)
 SOURCE_GROUP("" FILES ${OpenSCAD_SRCS})
+
+SET(OpenSCADTests_SRCS
+    OpenSCADTest/__init__.py
+    OpenSCADTest/test_information.md
+)
+
+SET(OpenSCADTestsApp_SRCS
+    OpenSCADTest/app/__init__.py
+    OpenSCADTest/app/test_importCSG.py
+)
+
+SET(OpenSCADTestsGui_SRCS
+    OpenSCADTest/gui/__init__.py
+    OpenSCADTest/gui/test_dummy.py
+)
+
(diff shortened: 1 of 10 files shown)


src/Mod/OpenSCAD/CMakeLists.txt
```diff
@@ -59,3 +101,7 @@ INSTALL(
        "${CMAKE_INSTALL_DATADIR}/Mod/OpenSCAD/Resources/icons"
 )
 
+INSTALL(FILES ${OpenSCADTests_SRCS} DESTINATION Mod/OpenSCAD/OpenSCADtest)
```
wwmayer, line 104: OpenSCADtest must become OpenSCADTest as otherwise a "make installed" FreeCAD will fail to load the test modules.

wwmayer:
All travis builds fail for different reasons:
* On Windows it fails because OpenSCAD is either not installed or not in PATH. The unit test could print an error message but shouldn't make the test fail
* On Linux the unit test fails because TestOpenSCAD cannot be loaded. Here after a _make_ it creates a directory called OpenSCADTest. However, when doing _make install_ the directory is called OpenSCADtest
* On my machine the test test_open_scad (line 55) `self.assertTrue (group is not None)` fails with the reason: AssertionError: False is not true

chennes:
>     * On my machine the test test_open_scad (line 55) `self.assertTrue (group is not None)` fails with the reason: AssertionError: False is not true

Thanks for testing - the first two are easy fixes (I'll push a fix shortly), but this last one I'm a little unsure of. I was surprised by this difference between loading a CSG file and loading an SCAD file (CSG doesn't create a group, but SCAD does), but figured it was being done on purpose. But it seems it's not even always true. I'll have to look closer at the creation of that group.

src/Mod/OpenSCAD/CMakeLists.txt
chennes: I completely forgot to test the install target, sorry about that! Fixed now.

chennes:
It seems to me that testing the existence of "Group" is actually a poor choice on my part, it's not an important component of the file. The test really should only be looking for the three objects, not the specific structure of the import, which is more of an implementation detail. So I've dropped that check.

wwmayer:
I still have two problems:
* at startup an error is raised because OpenSCADUtils isn't imported in InitGui.py
* the App unit tests fail with the error message that group is not defined

Both of them are easy to fix which I will quickly do.

wwmayer:
Merged as c1408f130c

PR #4542, opened 2021-03-01
[PD] fix multi-transform view

donovaly:
- fixes regression bug 4581
  See the first issue reported here:
  https://forum.freecadweb.org/viewtopic.php?f=3&t=56093#p482553

- also fix issue that one could set 1 occurrence for polar patterns despite 2 are required at least

- also fix potential dereferencing null pointer (reported by MSVC)

When this goes in, it should also go in for FC 0.19.1

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskMultiTransformParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskMultiTransformParameters.cpp
@@ -229,7 +229,8 @@ void TaskMultiTransformParameters::closeSubTask()
 
 void TaskMultiTransformParameters::onTransformDelete()
 {
-    if (editHint) return; // Can't delete the hint...
+    if (editHint)
+        return; // Can't delete the hint...
     int row = ui->listTransformFeatures->currentIndex().row();
     PartDesign::MultiTransform* pcMultiTransform = static_cast<PartDesign::MultiTransform*>(TransformedView->getObject());
     std::vector<App::DocumentObject*> transformFeatures = pcMultiTransform->Transformations.getValues();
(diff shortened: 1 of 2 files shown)


src/Mod/PartDesign/Gui/TaskMultiTransformParameters.cpp
```diff
@@ -308,66 +316,85 @@ void TaskMultiTransformParameters::onTransformAddLinearPattern()
...
-    FCMD_OBJ_CMD(Feat,"Length = 100");
-    FCMD_OBJ_CMD(Feat,"Occurrences = 2");
+    FCMD_OBJ_CMD(Feat, "Length = 100");
+    FCMD_OBJ_CMD(Feat, "Occurrences = 2");
 
     finishAdd(newFeatName);
+    // show the new view when no error
+    if (Feat)
+        if (!Feat->isError())
+        TransformedView->getObject()->Visibility.setValue(true);
```
chennes, line 348: Indentation looks off here - did you mean to combine the two if statements?

donovaly: Just a paste mistake. I fixed this now.

src/Mod/PartDesign/Gui/TaskPolarPatternParameters.cpp
```diff
@@ -156,6 +156,7 @@ void TaskPolarPatternParameters::setupUI()
 
     ui->polarAngle->bind(pcPolarPattern->Angle);
     ui->spinOccurrences->setMaximum(INT_MAX);
+    ui->spinOccurrences->setMinimum(2);
```
wwmayer, line 159: Some months ago the restriction of linear patterns has been lifted to also allow only one occurrence. As far as I can see this wasn't done for polar patterns yet. So, if this makes sense for linear patterns then it also makes perfectly sense for polar patterns.

donovaly: I thought also for linear patterns only one occurrence is senseless, but in contrary to polar patterns, it doesn't lead to an error. So I only changed it for polar patterns since there only 1 occurrence always leads to an error.

wwmayer: Allowing one occurrence doesn't harm and in GUI mode you probably won't use it. But the guy who implemented it for the linear pattern had a use case in combination with scripting and expressions.

donovaly:
Can this please also go int the 0.19 branch, since it fixes a regression bug?

PR #4545, opened 2021-03-01
Sketcher: fix performance issue with drawing svg icons

grggnzlz:
Made a fix to improve Sketcher dragging performance, especially when many constraint icons are drawn.  Avoids unnecessary calls to expensive pixmapFromSvg function.  Details in [this thread](https://forum.freecadweb.org/viewtopic.php?f=3&t=56064).

I tried running `./bin/FreeCAD --run-test 0` but get an "Illegal storage access" error after "Try importing femtaskpanels.task_material_common" even without my change.


davidosterberg:
Good job, tracking this down! I will compile and test it later today.

davidosterberg (approved):
This compiles, runs and fixes the issue. When I run the example model in the forum thread. My FPS increases from around 5 fps to above 10 fps. Good job. Regarding the code change, the change is very small. I am not absolutely certain what it mean to only redraw when we have Mode==STATUS_NONE. But in my testing it seems to work. I don't see any adverse effects on the drawing of the sketch.

abdullahtahiriyo:
@davidosterberg 
Thanks for testing and giving the feedback.

If you search for "Mode" or "STATUS_NONE" in ViewProviderSketch.cpp, you will see that it has to do with the kind of operation it is being performed (selecting a point, selecting a curve, dragging,...)

```
enum SketchMode{
        STATUS_NONE,              /**< enum value View provider is in neutral. */
        STATUS_SELECT_Point,      /**< enum value a point was selected. */
        STATUS_SELECT_Edge,       /**< enum value an edge was selected. */
        STATUS_SELECT_Constraint, /**< enum value a constraint was selected. */
        STATUS_SELECT_Cross,      /**< enum value the base coordinate system was selected. */
        STATUS_SKETCH_DragPoint,  /**< enum value while dragging a point. */
        STATUS_SKETCH_DragCurve,  /**< enum value while dragging a curve. */
        STATUS_SKETCH_DragConstraint,  /**< enum value while dragging a compatible constraint. */
        STATUS_SKETCH_UseHandler, /**< enum value a DrawSketchHandler is in control. */
        STATUS_SKETCH_StartRubberBand, /**< enum value for initiating a rubber band selection */
        STATUS_SKETCH_UseRubberBand /**< enum value when making a rubber band selection *//**< enum value a DrawSketchHandler is in control. */
    };
```
From memory (without checking the code for each case), I cannot ensure that all of these, but STATUS_NONE do not require to redraw the icons and update the color. So here, my doubt is rather if "Mode==STATUS_NONE" is the right condition or whether it should be something like if (Mode != STATUS_SKETCH_DragPoint && Mode != STATUS_SKETCH_DragCurve && ...). It may well be that Mode==STATUS_NONE is just fine.

I hope to be able to review this before the end of the week. I am doing overtime at work...

grggnzlz:
Added more fixes described [here](https://forum.freecadweb.org/viewtopic.php?f=10&t=56375l).

abdullahtahiriyo:
I have taken a look:
1. Skipping constraint icon/color update in draw() is ok. I have just a remark that it is needed in mod STATUS_SKETCH_UseHandler as well.
2. Caching the icons is a sound idea too. They are heavily reused.
3. Moving the constraint icon/color update logic from the selection mechanism of ViewProvider to the Constraints' widget is a no go under the Single Responsibility Principle. I have taken a look over how this is done in other places in the code base. As far as I have seen, it is done the same. If we want to improve this (I am not sure it is meaningful after the other improvements are in), then we need a different mechanism.
4. The frame skipping mechanism I am not sure I understood it well. The difference is huge. As I see it, it is a mechanism that skips draws. But this draws may be necessary. Not sure whether it makes sense after the other improvements.

So, I have merged the first commit (011e19060aaa98c7c05f1b3412c78612a3d00a32) and have excised the caching from the second commit and merged it as well (89b8e81d5c3b11572a0288720d0b4145d2ec2a7e).

My proposal now is to first re-evaluate the performance with the current master and see if improvements are necessary. If improvements are necessary, we can use the forum thread to discuss it:
https://forum.freecadweb.org/viewtopic.php?f=10&t=56375

grggnzlz:
Thanks.  Sounds good.

PR #4546, opened 2021-03-01
[DRAFT] Remove unused loop variable

chennes:
LGTM is concerned about the unused loop variable in this algorithm. While there was nothing wrong here, to eliminate the alarm the loop was removed entirely, since Python allows the use of the multiplication operator here to achieve the same effect.

---

- [X]  Minor change
- [X]  No ticket

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_trimex.py
+++ b/src/Mod/Draft/draftguitools/gui_trimex.py
@@ -120,8 +120,7 @@ class Trimex(gui_base_original.Modifier):
             self.extrudeMode = True
             self.ghost = [trackers.ghostTracker([self.obj])]
             self.normal = self.obj.Shape.Faces[0].normalAt(0.5, 0.5)
-            for v in self.obj.Shape.Vertexes:
-                self.ghost.append(trackers.lineTracker())
+            self.ghost += [trackers.lineTracker()] * len(self.obj.Shape.Vertexes)
         elif len(self.obj.Shape.Faces) > 1:
             # face extrude mode, a new object is created
             ss = Gui.Selection.getSelectionEx()[0]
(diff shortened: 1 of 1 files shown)


yorikvanhavre:
Marking for 0.20 as this is a non-essential fix...

hyarion (approved):
Removes unused variable 👍
Makes code more pythonic 👍

src/Mod/Draft/draftguitools/gui_trimex.py
```diff
@@ -120,8 +120,7 @@ def proceed(self):
             self.extrudeMode = True
             self.ghost = [trackers.ghostTracker([self.obj])]
             self.normal = self.obj.Shape.Faces[0].normalAt(0.5, 0.5)
-            for v in self.obj.Shape.Vertexes:
-                self.ghost.append(trackers.lineTracker())
+            self.ghost += [trackers.lineTracker()] * len(self.obj.Shape.Vertexes)
```
marioalexis84, line 123: @chennes 
self.ghost + = [trackers.lineTracker ()] * len (self.obj.Shape.Vertexes)
add references to the same lineTracker object and here we need references to several lineTrackers (one for each vertex).
To recreate the problem, draw a face, extrude it using the trimex command (pressing the z key).
It will be observed that the ghost line is generated for only one vertex.
At the end of the command, several errors are reported due to the inconsistency in the number of trackers explained above.
Is there another way to make the analyzer not complain?

marioalexis84: does the analyzer complain even adding an underscore to the unused loop variables?

chennes: No, the underscore convention will work, as will naming the variable "unused".

marioalexis84: ok, we can add this to the coding conventions file.

PR #4553, opened 2021-03-02
[PD] disallow invalid polar/linear pattern settings

donovaly:
as discussed in https://github.com/FreeCAD/FreeCAD/commit/356db441a6c4bf546828e149d7269b2d61c9f039 we should disallow invalid property settings

at the moment we allow errors like these: 
Zf9cUPnInH

This PR sets constraints to avoid this.

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePolarPattern.cpp
+++ b/src/Mod/PartDesign/App/FeaturePolarPattern.cpp
@@ -48,12 +48,17 @@ namespace PartDesign {
 
 PROPERTY_SOURCE(PartDesign::PolarPattern, PartDesign::Transformed)
 
+const App::PropertyIntegerConstraint::Constraints intOccurrences = { 1, INT_MAX, 1 };
+const App::PropertyAngle::Constraints floatAngle = { Base::toDegrees<double>(Precision::Angular()), 360.0, 1.0 };
+
 PolarPattern::PolarPattern()
 {
-    ADD_PROPERTY_TYPE(Axis,(0),"PolarPattern",(App::PropertyType)(App::Prop_None),"Direction");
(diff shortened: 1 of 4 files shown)


src/Mod/PartDesign/App/FeaturePolarPattern.cpp
```diff
@@ -48,12 +48,17 @@ namespace PartDesign {
 
 PROPERTY_SOURCE(PartDesign::PolarPattern, PartDesign::Transformed)
 
+const App::PropertyIntegerConstraint::Constraints intOccurrences = { 1, INT_MAX, 1 };
+const App::PropertyAngle::Constraints floatAngle = { FLT_MIN, 360, 1.0 };
```
wwmayer, line 52: Defining the minimum of FLT_MIN, i.e. ~1.1e-38 doesn't make much sense because practically you cannot set such a small value in the GUI and in the execute() method there is another check for Precision::Confusion() which is 1e-7.
However, using Precision::Confusion() for angles isn't the best idea either since OCCT provides a special tolerance for angles: Precision::Angular() which is 1e-12.

This value, however, is given in radians and not degrees. So, the minimum can be defined as
`Base::toDegrees<double>(Precision::Angular())` or hard-coded as 6.0e-11

The addi

donovaly: Many thanks! I updated the PR accordingly. i thin it is better not to hard-code in case OCCT changes the tolerance in future.

While making the PR I realized that many PD features miss some constraints. I will try to improve the situation bit by bit the next days.

wwmayer:
Merged as 7f1b47d1185442

PR #4554, opened 2021-03-02
[PD] disallow angles <=0 for holes

donovaly:
neither the taper, drill or countersink angle can be zero or negative for a valid hole, thus set a constraint

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHole.cpp
+++ b/src/Mod/PartDesign/App/FeatureHole.cpp
@@ -645,7 +645,7 @@ const char* Hole::ThreadDirectionEnums[]  = { "Right", "Left", NULL};
 
 PROPERTY_SOURCE(PartDesign::Hole, PartDesign::ProfileBased)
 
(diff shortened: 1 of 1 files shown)


src/Mod/PartDesign/App/FeatureHole.cpp
```diff
@@ -645,6 +645,8 @@ const char* Hole::ThreadDirectionEnums[]  = { "Right", "Left", NULL};
 
 PROPERTY_SOURCE(PartDesign::Hole, PartDesign::ProfileBased)
 
+const App::PropertyAngle::Constraints floatAngle = { FLT_MIN, 360, 1.0 };
```
wwmayer, line 648: The same applies here as in https://github.com/FreeCAD/FreeCAD/pull/4553

donovaly: Thanks, now fixed.

donovaly:
something went wrong with the branch name, I had to create the new PR #4561 that replaces this one.

PR #4560, opened 2021-03-02
[Draft] Eliminate unused imports

chennes:
As is common in FreeCAD, Draft has two files whose sole purpose is to import the workbench -- the files do not, in general, use anything they have imported. Static analyzers such as Flake and LGTM complain about these imports, so this commit adds fake "use" of each of those imports. It also includes a comment to silence the Pylint warning about code that does nothing on those lines.

As an aside, I hate adding hundreds of lines of "code" to appease the analyzers here, so if there is another way to handle this I'd love to hear it.

---

- [X]  Small change
- [X]  No tracker ticket


hyarion:
I don't like how these files looks (not your fault). Maybe we could clean it up a bit and solve this in a cleaner way?

How about creating a helper function to preload modules and attributes and called that with a series of lists with strings instead?

The helper function could look something like this:
```python
from importlib import import_module
def preload(entries):
    for entry in entries:
        module = import_module(entry[0])
        for attr in entry[1]:
            getattr(module, attr)
```

And then call it this way:
```python
preload_modules([
    ['draftutils.utils', [
        'type_check', 'typecheck', 'get_param_type', 'getParamType', 'get_param', 'getParam', 'set_param',
        'setParam', 'precision', 'tolerance', 'epsilon'  # , ...
    ],['draftfunctions.svg', [
        'get_svg', 'getSVG'
    ]]
])
```
The code above hasn't been tested, but I think something like that should work :tm:

chennes:
Since this would represent a fairly large departure from the norm, we should probably move the discussion to the forums:
https://forum.freecadweb.org/viewtopic.php?f=10&t=56311

yorikvanhavre:
In any case, it's probably better to leave this discussion to after the 0.19 release is done and consumed...

yorikvanhavre:
Marking as "on hold" while the discussion is on...

chennes:
Following the forums discussion, I'm going to rework this PR to use an alternate strategy.

PR #4561, opened 2021-03-03
[PD] disallow angles <=0 for holes and revolution

donovaly:
neither the taper, drill or countersink angle can be zero or negative for a valid hole, thus set a constraint

### DIFF
--- a/src/Mod/PartDesign/App/FeatureRevolution.cpp
+++ b/src/Mod/PartDesign/App/FeatureRevolution.cpp
@@ -53,7 +53,7 @@ namespace PartDesign {
 
 PROPERTY_SOURCE(PartDesign::Revolution, PartDesign::ProfileBased)
 
-const App::PropertyAngle::Constraints floatAngle = { Base::toDegrees<double>(Precision::Angular()), 360, 1.0 };
+const App::PropertyAngle::Constraints Revolution::floatAngle = { Base::toDegrees<double>(Precision::Angular()), 360.0, 1.0 };
 
 Revolution::Revolution()
 {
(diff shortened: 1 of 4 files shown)


src/Mod/PartDesign/App/FeatureRevolution.cpp
```diff
@@ -53,13 +53,16 @@ namespace PartDesign {
 
 PROPERTY_SOURCE(PartDesign::Revolution, PartDesign::ProfileBased)
 
+const App::PropertyAngle::Constraints floatAngle = { Base::toDegrees<double>(Precision::Angular()), 360, 1.0 };
```
wwmayer, line 56: You must be careful with redefinitions of floatAngle in the global namespace as this violates ODR (One Definition Rule). 
Either make them a static variable of the corresponding classes or put them into `namespace {...}` because then the compiler makes sure that different namespaces are generated.

donovaly: Many thanks for this hint!
Indeed, when I comment out the floatAngle definition, the MSVC IDE warns me that it has different definitions and it is unclear what to use. Since it showed me floatAngle definitions an chamfer and polar pattern, I made PR #4563 to fix this as well. With the latest commit to this PR and  #4563 the IDE is silenced.

wwmayer:
Merged as c0de42335b

wwmayer:
About the odr issue I was wrong. It didn't violate the odr because the variables are declared as const and according to https://www.c-plusplus.net/forum/topic/251918/global-modulglobal-lokal the const leads to internal linking.

Only if the const keyword was dropped it would violate the odr and additionally causes a linker error.
Sorry, for the extra work.

PR #4562, opened 2021-03-03
[PD] set constraints for helix properties

donovaly:
at the moment it is possible by a click to set e.g. -1 turn which would cause an error

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -71,24 +71,29 @@ const char* Helix::ModeEnums[] = {"pitch-height", "pitch-turns", "height-turns",
 
 PROPERTY_SOURCE(PartDesign::Helix, PartDesign::ProfileBased)
 
+// we purposely use not FLT_MAX because this would not be computable
+const App::PropertyFloatConstraint::Constraints floatTurns = { Precision::Confusion(), INT_MAX, 1.0 };
+const App::PropertyAngle::Constraints floatAngle = { -89.0, 89.0, 1.0 };
+
 Helix::Helix()
 {
     addSubType = FeatureAddSub::Additive;
 
-    ADD_PROPERTY_TYPE(Base,(Base::Vector3d(0.0,0.0,0.0)),"Helix", App::Prop_ReadOnly, "Base");
-    ADD_PROPERTY_TYPE(Axis,(Base::Vector3d(0.0,1.0,0.0)),"Helix", App::Prop_ReadOnly, "Axis");
-    ADD_PROPERTY_TYPE(Pitch,(10.),"Helix", App::Prop_None, "Pitch");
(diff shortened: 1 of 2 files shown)


src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -71,24 +71,29 @@ const char* Helix::ModeEnums[] = {"pitch-height", "pitch-turns", "height-turns",
 
 PROPERTY_SOURCE(PartDesign::Helix, PartDesign::ProfileBased)
 
+// we purposely use not FLT_MAX because this would not be computable
+const App::PropertyFloatConstraint::Constraints floatTurns = { Base::toDegrees<double>(Precision::Confusion()), INT_MAX, 1.0 };
```
wwmayer, line 75: Why Base::toDegrees? floatTurns doesn't seem to restrict an angle.

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -71,24 +71,29 @@ const char* Helix::ModeEnums[] = {"pitch-height", "pitch-turns", "height-turns",
...
+const App::PropertyAngle::Constraints floatAngle = { -89.0, 89.0, 1.0 };
```
wwmayer, line 76: This is a bad restriction as you cannot enter -89.5 or +89.5. I think it's better to set the range [-90.0, +90.0] and then in the execute method raise an error if the angle is too close to the lower or upper limit.

src/Mod/PartDesign/App/FeatureHelix.cpp
donovaly: Damn it. Seems my force push with the correction did not went through, not it is correct.

src/Mod/PartDesign/App/FeatureHelix.cpp
donovaly: I am aware of that. But David restrict it to 89.0 in the dialog too.  I first thought "90 - Precision::Confusion() " might be a better upper limit, but even 89.9 takes ages to compute and I cannot see a real life use case for such a helix beast. Therefore I think david choose 89.0 for good reason and the limits in App and Gui should be equal.

donovaly: @davidosterberg  , what is your opinion?

wwmayer: OK but independent of what value range we go for we should not hard-code the limits in the feature **and** the dialog.

Today I have added some convenience functions to more easily get the limits and these functions should then be used for the dialog. The advantage is that whenever we change the range we have to do it only once.

davidosterberg: I agree that angle > 89deg are not usefu with the current implementation. Would've been nice if 90deg was supported (i.e a flat spiral). Unfortunately makeLongHelix is not general enough.

PR #4567, opened 2021-03-03
Gui: Change the Part icon

davidosterberg:
This PR changes the icon of the Part container.
image

Motivation:
The new icon makes it clear that the Part object can contain multiple objects of different type making it suitable for assemblies etc. 

Disclosure: This icon is actually the current logo for the Assembly 4 workbench. Used for this purpose with the permission of @Zolko-123 

Original forum thread:
https://forum.freecadweb.org/viewtopic.php?f=8&t=56055

Icon specific forum thread
https://forum.freecadweb.org/viewtopic.php?f=34&t=56253

### DIFF
--- a/src/Gui/Icons/Geofeaturegroup.svg
+++ b/src/Gui/Icons/Geofeaturegroup.svg
@@ -1,258 +1,91 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!-- Created with Inkscape (http://www.inkscape.org/) -->
-
-<svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-   xmlns:svg="http://www.w3.org/2000/svg"
-   xmlns="http://www.w3.org/2000/svg"
-   xmlns:xlink="http://www.w3.org/1999/xlink"
-   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
-   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   width="64"
-   height="64"
-   id="svg3559"
-   version="1.1"
-   inkscape:version="0.91+devel+osxmenu r12922"
-   sodipodi:docname="Part_v09b.svg"
-   viewBox="0 0 64 64">
-  <defs
-     id="defs3561">
-    <linearGradient
-       inkscape:collect="always"
-       id="linearGradient73208">
-      <stop
-         style="stop-color:#c4a000;stop-opacity:1"
-         offset="0"
-         id="stop73210" />
-      <stop
-         style="stop-color:#edd400;stop-opacity:1"
-         offset="1"
-         id="stop73212" />
-    </linearGradient>
-    <linearGradient
-       id="linearGradient73194"
-       inkscape:collect="always">
-      <stop
-         id="stop73202"
-         offset="0"
-         style="stop-color:#c4a000;stop-opacity:1" />
-      <stop
-         id="stop73204"
-         offset="1"
-         style="stop-color:#fce94f;stop-opacity:1" />
-    </linearGradient>
-    <linearGradient
-       id="linearGradient4383-3"
-       inkscape:collect="always">
-      <stop
-         id="stop73188"
-         offset="0"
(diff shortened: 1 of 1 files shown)


hyarion:
The outline has two different colors:
 * black(-ish) for the yellow part and a
 * dark ivory (real dark yellow) color for the blue part.

Is this intensional?

Anyway, I think this will be a nice improvement 👍

Zolko-123:
> Is this intensional?

no, it's not. I check

Zolko-123:
you're right, and the 2 colours come from the Part (App::Part) and Body (PartDesign::Body) icons. I changed all the outlines to the darker colour. If you can use this version (can only attach zipped files, but inside it's the SVG)

[Assembly4.zip](https://github.com/FreeCAD/FreeCAD/files/6078926/Assembly4.zip)

sasobadovinac:
I am not in favor for this change, Part represents a single element so the current icon is ok, but we have come more or less to the agreement that Part in FreeCAD should work similar to Part in NX and Fusion, that is that there is only one component type object for both Parr and Assembly. So what should actually be done in FreeCAD is that the Part icon should automatically change to the above proposed new icon, when other Parts get nested in to it, because then the function of the parent Part object changes in to an Assembly. This is also how it works in NX and Fusion.

adrianinsaval:
Part is not a single element, it's capable of housing multiple parts, bodies and anything else. Changing the icon according to the content is unnecessary IMO.

About the outline, I think something like this would be better:
Geofeaturegroup

[Geofeaturegroup.zip](https://github.com/FreeCAD/FreeCAD/files/6079400/Geofeaturegroup.zip)

sasobadovinac:
This Assembly/Part hierarchical structure is used by all parametric cad systems, regardless if they use one or two component types to represent them, it is also used by the step file format when exporting/importing between different cad systems, in general also BOMs are build out of this structure,... Ignoring this and just implementing something that maybe first some current unfinished implementations will IMO push FreeCAD just further away from being a more professional tool.

Zolko-123:
> So what should actually be done in FreeCAD is that the Part icon should automatically change to the above proposed new icon, when other Parts get nested in to it, because then the function of the parent Part object changes in to an Assembly.

This is a good idea. The type of children that should be checked-for should be App::Part, PartDesign::Body and App::Link. If an App::Part has any of these as subObject, and only these, then it's an assembly (or aggregate) .

sasobadovinac:
For Link it would probably be better only in case if it is a Link of Part or Body

Zolko-123:
How about: 

Geofeaturegroup

See [forum thread](https://forum.freecadweb.org/viewtopic.php?f=34&p=484158#p484158) for SVG file

luzpaz:
I'm sorry to rain on the parade here. I'm usually the one pushing for changes especially in the UI. But...i gotta say that there is a process to do it. I see that a large thread is linked in the PR and there was some discussion. BUT it was off-topic of the original thread. If you want to propose a change like this it, in my experience it needs to be presented in a exclusive thread with a clear RFC approach. That way the community can weigh in on it on the forum and not here in the repo issue queue. Again, i'm cringing about the fact of hampering forward momentum but we have a process to prepose UI/UX changes and they happen in the forum and ideally in the UI/UX subforum.

Never thought i'd promote so much bureaucracy until I joined FC. But as these projects grow it makes sense to have structures in place.

davidosterberg:
But there is a specific thread in the UX/UI forum linked above! Please post your objections there.

luzpaz:
> But there is a specific thread in the UX/UI forum linked above! Please post your objections there.

yikes! I missed that one. cancel clear!

davidosterberg:
Ok, I have updated with the yellow-green-blue proposal, that I now think is the public favorite.

Zolko-123:
> Ok, I have updated with the yellow-green-blue proposal, that I now think is the public favorite.

ahem ... could you please replace it with this version here (there *was* a last misalignment)

[part_new.zip](https://github.com/FreeCAD/FreeCAD/files/6084121/part_new.zip)

wwmayer:
Can this be merged now?

Zolko-123:
I just checked, and the Geofeaturegroup.svg icon in David's repo is the correct version with all fixes (outline colours and perspective/parallel lines).

yorikvanhavre:
I would vote for merging it after the 0.19 release as it won't look like in the docs anymore and it'll take time to update the docs. But as you prefer @wwmayer ;)

luzpaz:
Who's gonna update the wiki ? :smirk:

PR #4503, opened 2021-02-21
[Part] Set App::Part's Material property type to App::PropertyLink (instead of App::PropertyMap)

howetuft:
App::Part Material property type is currently App::PropertyMap, but to be consistent with other domains using materials (like Arch, especially), it should rather be an App::PropertyLink, in order to link to a Material card.

Please note this standardization is requested for Render Workbench, in order to be able to use Material rendering features with Part objects, as discussed in [this thread](https://github.com/FreeCAD/FreeCAD-render/issues/123). 

- [X]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [ ]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [X]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [X]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [X]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [ ]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the [FreeCAD bug tracker issue number](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) in case a particular commit solves or is related to an existing issue on the tracker. Ex: `Draft: fix typos - fixes #0004805`

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.19 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=34586).

---

PR #4568, opened 2021-03-03
[PartDesign] Add Material property to Body

howetuft:

---

### DIFF
--- a/src/Mod/Part/App/BodyBase.cpp
+++ b/src/Mod/Part/App/BodyBase.cpp
@@ -44,6 +44,7 @@ BodyBase::BodyBase()
     Tip.setScope(App::LinkScope::Child);
 
     ADD_PROPERTY(BaseFeature , (0) );
+    ADD_PROPERTY_TYPE(Material, (0), 0, App::Prop_None, "The Material for this Part");
 
     App::OriginGroupExtension::initExtension(this);
 }
--- a/src/Mod/Part/App/BodyBase.h
+++ b/src/Mod/Part/App/BodyBase.h
@@ -57,6 +57,14 @@ public:
      */
     App::PropertyLink BaseFeature;
 
+
+    /**
+     * A material card this body is associated with
+     *
+     */
+    App::PropertyLink Material;
+
+
     /// Returns all Group objects prepanded by BaseFeature (if any)
     std::vector<App::DocumentObject *> getFullModel () {
         std::vector<App::DocumentObject *> rv;


wwmayer:
What exactly is this good for? Is this somehow needed for the Arch wb?

howetuft:
Hello,
The need comes from Render workbench: the PR aims at rendering part bodies with assigned materials. We've been asked for such a capability twice in last weeks: 
https://forum.freecadweb.org/viewtopic.php?f=3&t=55297&start=20#p483298
https://github.com/FreeCAD/FreeCAD-render/issues/123#issuecomment-782855995
and I think this will become a recurring demand.
Moreover, it would seem natural to me to extend Material system to Part domain (not limited to Arch or FEM).

wwmayer:
OK, thanks for the link.

realthunder:
I don't think that's enough reason to add a static property with Body itself never uses. Why not just add a dynamic property when needed? You can now easily add any type of property to any type of object. You can take advantage of the simple 'namespace' like functionality to avoid name conflict. If you add a property using its group name as the property name prefix, the property will be displayed in the property editor with the prefix hidden.

The following code adds a property named 'Render_Material' to an object named 'Body', but you'll see in the property view with a property named 'Material' in group 'Render'. Be careful not to repeatedly add the property with the same name. The current upstream will just silently auto rename a duplicated property. I have a pending PR to throw exception in that case.

```python
App.ActiveDocument.getObject('Body').addProperty('App::PropertyLink', 'Render_Material', 'Render')
```

howetuft:
It is not "Render Material", it is simply "Material" (https://wiki.freecadweb.org/Material). Why should Body never use a Material?

wwmayer:
realthunder is actually right that there is no real need to make this a static property.
You add it at runtime with:
`App.ActiveDocument.getObject('Body').addProperty('App::PropertyLink', 'Material', 'Render')`

realthunder:
> It is not "Render Material", it is simply "Material" (https://wiki.freecadweb.org/Material). Why should Body never use a Material?

I mean it is currently used by the Render WB, and not by PartDesign Body itself. Sure, Body most likely will have some Material property in the future, and that's exactly where the problem is. It is going to cause conflict. For example, my branch supports multiple solids inside the Body as a container. So there needs a way to specify multiple materials. In fact, there is a hot topic discussing whether we should rename the Body to a Part container and so on.

howetuft:
Hello guys,

First, thank you for your answers.

However, just to be sure I'm not misunderstood: I'm not asking for a Render-specific feature to be implemented in PartDesign Body. I'm asking for the implementation of the *standard* Material property - ie the one described in documentation (see my link above), discussed in dedicated Material forum and already implemented in some workbenches, for instance in Arch and FEM.

From that point, the objection that it would be used only by Render appears unjustified to me - it has a more general scope. Unjustified as well is the objection that it would conflict to any further development, *unless you plan to make a non-standard Material implementation for Body.*

I would also like to draw your attention to several points:
- users report about the Material management to be deficient and inconsistent across FreeCAD (for instance, see here: https://forum.freecadweb.org/viewtopic.php?f=3&t=55297&start=20#p483298, https://forum.freecadweb.org/viewtopic.php?f=38&t=41403...)
- users have been requiring PartDesign to have a material property for a long time (https://forum.freecadweb.org/viewtopic.php?p=237373#p237010)
- as an add-on (Render) developper, I join myself to these expectations. I've trustfully based Render on FreeCAD Material system, following a discussion with Yorik (https://github.com/FreeCAD/FreeCAD-render/issues/64#issuecomment-642072755 - please read this post, it also clarifies a great deal what Material should be) but I really get hampered by the yet deficient and non-harmonised implementation of Material in FCD.

So, my conclusion is: all right, you are the ones in charge, so I will follow your guidelines, use a dynamic property for the moment and wait for your final development. But you should really consider to address the point quickly and in a standardized manner, IMHO. Thx.

realthunder:
> However, just to be sure I'm not misunderstood: I'm not asking for a Render-specific feature to be implemented in PartDesign Body. I'm asking for the implementation of the _standard_ Material property - ie the one described in documentation (see my link above), discussed in dedicated Material forum and already implemented in some workbenches, for instance in Arch and FEM.

Understood. But my point still stands. A static property is not flexible enough, especially so if what you want is a _standard_ Material property. Like I mentioned above, how shall we handle multiple solids inside a body? And I am sure there are people use FEM without `PartDesign::Body`, shall we add the Material property to every Part object? And what about a Link. Can it have a different material than the linked object?

Add a dynamic property on demand is one way to solve the problems, but of course there are more details need to be worked out for it to be a _standard_. If I understand correctly, FEM's material is currently implemented like some kind of group that references other objects as a way to assign the material. This is actually not a bad solution, either.

howetuft:
If I may not follow your order:

> If I understand correctly, FEM's material is currently implemented like some kind of group that references other objects

Yes, in FEM, the material points to the objects (parts etc.) made of it, whereas, in Arch, the object points to its material. That's an illustration of the lack of harmonization I mentioned in the last post...
Personally, I give my preference to Arch implementation. First, as far as I know, in major CG/CAD applications (Blender, AutoCAD...), objects reference their materials, so in a POLA approach I would not reverse the paradigm. I find FEM implementation less readable, as you have to open every material to know which is the one of your part. And the same in code: you have to loop through all materials to find the material information of a given object. So, I would personally stick to Arch and Part approach: the object points to its material.

> shall we add the Material property to every Part object?

But Part has already a Material property ([here](https://github.com/FreeCAD/FreeCAD/blob/master/src/App/Part.h#L56))! The trouble is that it is rather unusable (and likely most unused) because it is of wrong type, compared to Arch "standard", and on top of that it's hidden. I posted a PR to fix that a few days ago ([4503](https://github.com/FreeCAD/FreeCAD/pull/4503)).

> how shall we handle multiple solids inside a body?
Yes, this is the key question, or more generally: how to handle containers and components. To me: we could have a Material property to whatever object for which it makes sense. It includes Part, Body, Link... And then, we could have a convention: the subcomponent's material, if any, overrides the container's material.
So, if we have the following:

CompoundObj (Material#1)
|-- SubObj1 (No material)
|-- SubObj2 (Material#2)

...we should understand SubObj1 has Material#1 and SubObj2 has Material#2.

PR #4569, opened 2021-03-03
Gui: Update icon for Sketcher Constraint Block

00001000bit:
The existing icon for the sketcher constraint for "constraint block" is easily confused with the icon for the diameter constraint. A small modification to add a 2nd cross bar will visually distinguish it (especially at smaller icon sizes).

icons_proposed


hyarion:
Thank you! I feel so stupid each type I click the wrong one xD

Here's a link to the forum post that 00001000bit created for the topic: https://forum.freecadweb.org/viewtopic.php?f=34&t=56247

PR #4570, opened 2021-03-03
[PD] Add missing translation calls

chennes:
The items in the PartDesign menu that have submenus are not implemented using the standard Command architecture (which implements the translation strings internally). This commit adds explicit `QT_TR_NOOP` calls around those strings.

Fixes #0004583. Suitable for back-porting if desired.
https://tracker.freecadweb.org/view.php?id=4583

---

- [X]  Small change

### DIFF
--- a/src/Mod/PartDesign/Gui/Workbench.cpp
+++ b/src/Mod/PartDesign/Gui/Workbench.cpp
@@ -490,33 +490,33 @@ Gui::MenuItem* Workbench::setupMenuBar() const
 
     // datums
     Gui::MenuItem* datums = new Gui::MenuItem;
-    datums->setCommand("Create a datum");
+    datums->setCommand(QT_TR_NOOP("Create a datum"));
     *datums << "PartDesign_Point" << "PartDesign_Line"
         << "PartDesign_Plane";
 
     // additives
     Gui::MenuItem* additives = new Gui::MenuItem;
-    additives->setCommand("Create an additive feature");
+    additives->setCommand(QT_TR_NOOP("Create an additive feature"));
     *additives << "PartDesign_Pad" << "PartDesign_Revolution"
         << "PartDesign_AdditiveLoft" << "PartDesign_AdditivePipe" << "PartDesign_AdditiveHelix";
 
     // subtractives
     Gui::MenuItem* subtractives = new Gui::MenuItem;
-    subtractives->setCommand("Create a subtractive feature");
+    subtractives->setCommand(QT_TR_NOOP("Create a subtractive feature"));
     *subtractives << "PartDesign_Pocket" << "PartDesign_Hole"
         << "PartDesign_Groove" << "PartDesign_SubtractiveLoft"
         << "PartDesign_SubtractivePipe" << "PartDesign_SubtractiveHelix";
 
     // transformations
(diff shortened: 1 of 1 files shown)


wwmayer:
These extra calls of QT_TR_NOOP are wrong because when running lupdate it adds this to the .ts files:
```
<context>
    <name>PartDesignGui::Workbench</name>
    <message>
        <location filename="../../Workbench.cpp" line="493"/>
        <source>Create a datum</source>
        <translation type="unfinished"></translation>
    </message>
...
</context>
```
As you can see the context is "PartDesignGui::Workbench" but the core system doesn't know about this context name and expects everything to be in "Workbench". So, the strings won't be translated at runtime.

Therefore we write this block at the top of the Workbench.cpp file:
```
#if 0 // needed for Qt's lupdate utility
    qApp->translate("Workbench", "&Part Design");
    qApp->translate("Workbench", "&Sketch");
    qApp->translate("Workbench", "Create a datum");
    qApp->translate("Workbench", "Create an additive feature");
    qApp->translate("Workbench", "Create a subtractive feature");
    qApp->translate("Workbench", "Apply a pattern");
    qApp->translate("Workbench", "Apply a dress-up feature");
    qApp->translate("Gui::TaskView::TaskWatcherCommands", "Face tools");
    qApp->translate("Gui::TaskView::TaskWatcherCommands", "Sketch tools");
    qApp->translate("Gui::TaskView::TaskWatcherCommands", "Create Geometry");
#endif
```

Now what's odd is when looking into the current .ts files then the above string literals are all missing and I have no idea why. Possibly the Python script that runs lupdate is broken because when I open a terminal window, change to src/Mod/PartDesign/Gui and run the commands below it adds 163 new strings:
```
qmake -project
lupdate Gui.pro
```
I have tested this with the v0.19 release branch because in current master we have already changed a few string literals.

AKielhorn:
This seems to be a general problem.

I see untranslated sub menus in "Spreadsheet" (Alignment and Styles). Both are in the  `.ts` file but not translated in the app. (They are both preceded with an `&`, does that interfere with the substitution?)

In "Part" (Create a copy) is not in `.ts` (but all the others are fine).

In "Mesh design" "Cutting" is not in `.ts`, the other sub menus are.

I don't want to report all cases. If this is a bug in the `lupdate` script, they will probably all get fixed with the next run.

wwmayer:
> I see untranslated sub menus in "Spreadsheet" (Alignment and Styles). Both are in the .ts file but not translated in the app. (They are both preceded with an &, does that interfere with the substitution?)

The & shouldn't cause problems. The issue somehow must be that the Python script fails to fetch new strings and add them to the .ts files.

And what's not inside the .ts files cannot be translated, of course.

luzpaz:
> The issue somehow must be that the Python script fails to fetch new strings and add them to the .ts files.

Yes! and i think this isn't the only place this is happening.

chennes:
Closing then, since this is the wrong approach.

chennes:
Just an update, I think that there is a regex problem in the python script: the regex that is removing "src/Mod/Part" is also removing "src/Mod/PartDesign".

chennes:
Replacement PR, fixing the Python script instead: https://github.com/FreeCAD/FreeCAD/pull/4576

PR #4572, opened 2021-03-04
[PD] use same helix angle step as for other features

donovaly:
All other PD features use a step size of 1.0 for angles.

(Personally also for me a step of 5 is not convenient)

@davidosterberg , OK for you as well?

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -73,7 +73,7 @@ PROPERTY_SOURCE(PartDesign::Helix, PartDesign::ProfileBased)
 
 // we purposely use not FLT_MAX because this would not be computable
 const App::PropertyFloatConstraint::Constraints floatTurns = { Precision::Confusion(), INT_MAX, 1.0 };
-const App::PropertyAngle::Constraints floatAngle = { -89.0, 89.0, 5.0 };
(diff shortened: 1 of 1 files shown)


davidosterberg:
Personally I prefer the stepsize of 5 for this application. I like to see a visible change when I use the step buttons. And I think that most often I want cone angles like 10, 15, 45 etc. Values that you can reach in a few seconds with current step. For when you dont want to reach for the keyboard. In my view the step buttons are not there for precision entry. I guess this preference would also hold for the other PD angle features. Maybe I am in the minority view, if so I wont be upset if the step is changed.

wwmayer:
Well, then let's merge this PR and revert the change in the dialog to use the feature's step size. So, in the property editor the step size will be again 1 and in the dialog it's still 5. That's then the behaviour how it was before.

wwmayer:
See 55896f316e

donovaly:
> Well, then let's merge this PR and revert the change in the dialog to use the feature's step size. So, in the property editor the step size will be again 1 and in the dialog it's still 5. That's then the behaviour how it was before.

I don't like this. If 5° is preferred, then OK, it should be 5°. But the Gui should take the setting from App since all other PD features do now.

Personally, I constructed just last weeks helices and needed slight angles (e.g. 3.2° fpr 3D printing) thus I was wondering why helix uses a different step than the other angles in PD.

PR #4574, opened 2021-03-04
[Core] Library requirement update for 0.20

chennes:
As discussed [on the FreeCAD forums](https://forum.freecadweb.org/viewtopic.php?f=10&t=56177), this PR updates our CMake build system to require the following software versions (corresponding to Ubuntu 18.04 LTS):
* GCC 7.3
* Clang 6.0
* CMake 3.10
* Python 3.6
* Qt - 5.9

It also sets the C++ standard to C++17 -- this PR is set as a Draft pending figuring out the problem with that standard version and Boost::Bind on one of the Travis builds.

Note that many of the changes in the diffs are indentation-only, since the Qt5 conditional was removed in several places. I left the BUILD_QT5 variable in place, but it defaults to ON and is ignored by the main build system (some Mods still refer to it -- those will be dealt with in later PRs).


chennes:
Mistakenly included an extra commit...

PR #4592, opened 2021-03-06
CMake: correct non-POSIX friendly 'sed-i' arg in FindPySide2Tools.cmake

luzpaz:
resolves #0004588  
ticket: https://tracker.freecadweb.org/view.php?id=4588

### DIFF
--- a/cMake/FindPySide2Tools.cmake
+++ b/cMake/FindPySide2Tools.cmake
@@ -55,10 +55,10 @@ MACRO(PYSIDE_WRAP_UI outfiles)
     else()
         # Especially on Open Build Service we don't want changing date like
         # pyside2-uic generates in comments at beginning., which is why
-        # we follow the tool command with in-place sed.
+        # we follow the tool command with a POSIX-friendly sed.
         ADD_CUSTOM_COMMAND(OUTPUT ${outfile}
           COMMAND "${PYSIDE2UICBINARY}" ${UICOPTIONS} "${infile}" -o "${outfile}"
-          COMMAND sed -i "/^# /d" "${outfile}"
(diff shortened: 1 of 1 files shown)


luzpaz:
@chennes got a moment to review?

chennes (approved):
The change looks fine, and in principle the two sed commands are the same -- I can't test it in a real application because I don't know where this macro is called from.

I personally didn't know that the -i option was non-POSIX, so you might consider tweaking the comment above this where it says "in-place" to specifically note that you aren't using -i for that reason.

luzpaz:
Thanks @chennes I tweaked the comment per your advice. Good catch. I added a `[skip ci]` the commit  message to avoid an unnecessary compile due to a comment change. 

@wwmayer this should be ready for merge.

luzpaz:
Note: this PR was a patch that was submitted from the FreeBSD folks to make their already complex packaging tasks easier :wink:

wwmayer:
Here some background information:
The sed command has been introduced with https://github.com/FreeCAD/FreeCAD/commit/2b9da837d79b80bbad546e8753ba7f021e5c7315#diff-70c718cc8ed63da585359f36123ba88fb61574e5d4531161d76d954f94434f2b where the openSuse maintainers wanted to avoid changing checksums because of the build date of the generated file.

With a later change the execute_command function was replaced with a add_custom_command that uses the -i option: https://github.com/FreeCAD/FreeCAD/commit/75c698d1807ffd2bd0c6d4e18a2974eedc580dcd#diff-70c718cc8ed63da585359f36123ba88fb61574e5d4531161d76d954f94434f2b

PR #4595, opened 2021-03-07
[Main] set Coin environment variable

donovaly:
as discussed in https://forum.freecadweb.org/viewtopic.php?f=10&t=56386&start=10 we need to set this environment variable for Coin on Windows

### DIFF
--- a/src/Main/MainGui.cpp
+++ b/src/Main/MainGui.cpp
@@ -160,6 +160,10 @@ int main( int argc, char ** argv )
         }
         argv_.push_back(0); // 0-terminated string
     }
+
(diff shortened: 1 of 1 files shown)


wwmayer:
Merged.

luzpaz:
@donovaly do you mind documenting this on the wiki as well? Thanks for the work here, btw!

donovaly:
@wwmayer , since we need this for our FC 0.19 builds, I made the PR for this branch. I see you now only merged it to master. Can you please also put it into the 019 branch so that e.g. @sgrogan can take this branch to make his builds out of the box?

PR #4599, opened 2021-03-08
[Gui] fix too large abut Dialog

donovaly:
as reported here https://forum.freecadweb.org/viewtopic.php?f=10&t=56425
the about dialog is too large for Hi-DPI screens

If this goes in, please also for the 0-19 branch

New layout with this PR:
FreeCAD_Y7xAKQWAaX

### DIFF
--- a/src/Gui/AboutApplication.ui
+++ b/src/Gui/AboutApplication.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>470</width>
-    <height>617</height>
+    <width>472</width>
+    <height>542</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -22,7 +22,16 @@
      <property name="spacing">
       <number>6</number>
      </property>
-     <property name="margin">
+     <property name="leftMargin">
+      <number>0</number>
+     </property>
+     <property name="topMargin">
+      <number>0</number>
+     </property>
+     <property name="rightMargin">
+      <number>0</number>
+     </property>
+     <property name="bottomMargin">
       <number>0</number>
      </property>
      <item>
@@ -50,7 +59,7 @@
    <item row="0" column="0">
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>1</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="tab_about">
       <attribute name="title">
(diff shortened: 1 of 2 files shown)


wwmayer:
Please test b14394c7

chennes:
> Please test [b14394c](https://github.com/FreeCAD/FreeCAD/commit/b14394c7e7a138187fc61e107420901cabc515ae)

When I scale my screen to 300% and let Windows override FreeCAD's scaling (in order to replicate the problem @donovaly reported), b14394c does allow the window to fit, but the 1/4 screen height image is pretty small (smaller than needed to fix the issue). For 0.19 maybe we aren't really worried about this edge case, but the fix I suggested in 0d3b5f126 results in less visual change to the resulting window.

I think going forward in 0.20 the best solution is to set a max window height, and then let the layout manager calculate an optimal size for the image region, then scale the image to that size. I think that will handle both large and small screens as gracefully as possible. I didn't want to back-port that to 0.19 because I've replaced this scaling section with Qt's built-in (as of Qt 5.6 I think) method.

wwmayer:
> b14394c does allow the window to fit, but the 1/4 screen height image is pretty small (smaller than needed to fix the issue).

It only affects a small number of people. Alternatively we can also hide the label with the splashcreen when there is not enough space.

> For 0.19 maybe we aren't really worried about this edge case, but the fix I suggested in 0d3b5f1 results in less visual change to the resulting window.

I don't think this works reliably. On my computer the height of the dialog is 617 px. When I hard-code a limit of e.g. 560 px then the final height is 599 px. So the layout manager doesn't guarantee that a widget really fits into the given height as otherwise it had to truncate widgets which it doesn't do. So, on a real system with a small display the dialog can be higher than the 90% of the screen.

Btw, https://doc.qt.io/qt-5/qwidget.html#adjustSize says that the maximum height is 2/3 of the screen height. Wouldn't this be the better option?

wwmayer:
As a test I have set QT_SCALE_FACTOR=3.0 in order to create a big enough dialog. From the dialog I can only see its windows bar but not the OK button. With this code snippet it can be checked whether the image should be shown or not:
```C++
adjustSize();
if (height() > rect.height()) {
    ui->labelSplashPicture->hide();
}
```

donovaly:
I fully agree with @wwmayer that this PR does not fix the problem, but it improves the situation in general because:
- the dialog is now movable -> this is very practical when e.g. making a screenshot for the forum. So you can then present the issue and the About Dialog in one screenshot
- unnecessary whitespace should in general be avoided.

So it should go into master just with the first commit of the PR.

wwmayer:
Please check with latest master: 300ab23600

luzpaz:
Just an aside, the git hash can be a minimum of 7 characters right? (we don't need the whole hash)

donovaly:
> Just an aside, the git hash can be a minimum of 7 characters right? (we don't need the whole hash)

Good point. I thought one should use 8 digits, but in every case we should shorten the hash to that.

luzpaz:
> > Just an aside, the git hash can be a minimum of 7 characters right? (we don't need the whole hash)
> 
> Good point. I thought one should use 8 digits, but in every case we should shorten the hash to that.

We could also make it a hyperlink to the github commit associated with it

donovaly:
> Try it on your computer where the dialog wasn't fully visible.

I will but what does this have to do with my argument that I want to
- squash the whitespace in general
- shorten the hash digit length

wwmayer:
I don't understand why you insist on merging a PR that is known that it doesn't fix the problem. Vanuan has already shown you a way how to fix it properly so that a scaling of 150% doesn't scale to 200%.

PR #4601, opened 2021-03-08
[0.19][Web] Remove HiDPI fix when using QTWEBVIEW

chennes:
The screen scale adjustment code here should not be necessary when using the Chromium-based QtWebView widget, which correctly handles logical vs. device px units. The scaling causes poor font size choice when used on ultrawide monitors, as reported in https://forum.freecadweb.org/viewtopic.php?p=485431

---

- [X]  Confined strictly to a single module.
- [X]  Discussed in the forum
- [X]  Branch is [rebased](https://git-scm.com/docs/git-rebase) on 0.19 release branch
- [X]  Unit tests pass
- [X]  No tracker ticket

### DIFF
--- a/src/Mod/Web/Gui/BrowserView.cpp
+++ b/src/Mod/Web/Gui/BrowserView.cpp
@@ -239,6 +239,7 @@ Py::Object BrowserViewPy::setHtml(const Py::Tuple& args)
 WebView::WebView(QWidget *parent)
     : QWEBVIEW(parent)
 {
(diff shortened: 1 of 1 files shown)


Vanuan:
Cool. So `setTextSizeMultiplier` method implementation can be removed as well for QT_WEBENGINE.

chennes:
@wwmayer Can this commit get added to the trunk as well?

PR #4602, opened 2021-03-08
[OpenSCAD] Minor fixes from LGTM analysis

chennes:
LGTM fixes that don't affect the run, plus one fix that ensures all resources are closed when using `Popen` to get the version.

---

- [X]  Small fix
- [X]  Branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master
- [X]  All FreeCAD unit tests are confirmed to pass
- [X]  Pull request is well written and has a good description, and its title starts with the module name
- [X]  No tracker ticket

### DIFF
--- a/src/Mod/OpenSCAD/InitGui.py
+++ b/src/Mod/OpenSCAD/InitGui.py
@@ -29,10 +29,6 @@
 #***************************************************************************/
 
 import FreeCAD
-import OpenSCADUtils
-param = FreeCAD.ParamGet("User parameter:BaseApp/Preferences/Mod/OpenSCAD")
-openscadfilename = param.GetString('openscadexecutable')
-
 
 class OpenSCADWorkbench ( Workbench ):
     "OpenSCAD workbench object"
@@ -64,7 +60,6 @@ class OpenSCADWorkbench ( Workbench ):
         param = FreeCAD.ParamGet("User parameter:BaseApp/Preferences/Mod/OpenSCAD")
         openscadfilename = param.GetString('openscadexecutable')
         if not openscadfilename:
-
             import OpenSCADUtils
             openscadfilename = OpenSCADUtils.searchforopenscadexe()
             if openscadfilename: #automatic search was succsessful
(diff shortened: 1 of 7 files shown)


src/Mod/OpenSCAD/InitGui.py
```diff
@@ -70,6 +67,7 @@ def QT_TRANSLATE_NOOP(scope, text):
             if openscadfilename: #automatic search was succsessful
                 FreeCAD.addImportType("OpenSCAD Format (*.scad)","importCSG") 
                 param.SetString('openscadexecutable',openscadfilename) #save the result
+                FreeCAD.__unit_test__ += ["TestOpenSCADGui"]
```
hyarion, line 70: I don't understand why this line was moved here. Could you explain why we the unit test only should be added if `param.GetString('openscadexecutable')` is empty?

Also there's one `import OpenSCADUtils` too much, either the one just above or the one at the top of the file.

src/Mod/OpenSCAD/OpenSCADFeatures.py
```diff
@@ -235,7 +235,6 @@ def updateProperty(self, fp, prop, value):
             setattr(fp, prop, value)
 
     def execute(self, fp):
-        pass
```
hyarion, line 238: Removing this `pass` statement brings back dead code into life. Is this intensional and did you test the undead code? (I don't see any comments in either PR or commit message about it)

chennes: That is not my understanding of how `pass` works -- of course, I've been wrong before! But `pass` is not `return`, it's just used as a placeholder where Python requires there to be some code and you don't plan on putting any (yet). So an empty function that you haven't written yet, etc. In this case the `pass` both literally and figuratively does nothing, and is not necessary.

src/Mod/OpenSCAD/InitGui.py
chennes: Actually, I'd say that whole conditional is probably indented one too many times. I'm not sure why the addImportType exists here at all, to be honest, this is the GUI code, and presumably that import doesn't require the GUI.

src/Mod/OpenSCAD/OpenSCADFeatures.py
hyarion: I'm so sorry  😓 You are of cause correct.
I wonder what the reason the original coder had for having it there then 🤔

hyarion: Reading it in its context made it clear that it must just be a stray `pass`

src/Mod/OpenSCAD/InitGui.py
chennes: Fixed now, thanks. I'll submit a separate PR to deal with the non-storage of the executable in the main module.

hyarion (approved):
Looks good :+1:

PR #4603, opened 2021-03-09
Gui: Add option for changing NaviCube size in Preferences

hyarion:
This pull request adds an option in `Preferences -> Display -> Navigation -> Navigation Cube` to change the navigation cube's size.

In order to do this the layout had to be rebuilt with a grid layout as the options would not otherwise be aligned.

**Before** (sorry for humongous screenshots.. ):
image

**After**:
image

---

- [x] Your pull request is confined strictly to a single module.
- [x] Small change or has a link to a forum post
- [x] Your branch is rebased on latest master git pull --rebase upstream master
- [x] All FreeCAD unit tests are confirmed to pass by running ./bin/FreeCAD --run-test 0
- [x] All commit messages are well-written ex: Fixes typo in Draft Move command text
- [x] Your pull request is well written and has a good description, and its title starts with the module name, ex: Draft: Fixed typos
- [x] Commit message includes ticket if available


src/Gui/NaviCube.cpp
```diff
@@ -172,13 +172,16 @@ class Face {
...
-class NaviCubeImplementation {
+class NaviCubeImplementation : public ParameterGrp::ObserverType {
 public:
 	NaviCubeImplementation(Gui::View3DInventorViewer*);
 	virtual ~ NaviCubeImplementation();
 	void drawNaviCube();
 	void createContextMenu(const std::vector<std::string>& cmd);
 
+	/// Observer message from the ParameterGrp
+    virtual void OnChange(ParameterGrp::SubjectType &rCaller,ParameterGrp::MessageType Reason);
```
chennes, line 183: I hate to suggest this (because I am an all-spaces-all-the-time coder), but... the code on either side uses tabs. It would probably be best to stick to that here.

src/Gui/DlgSettingsNavigation.ui
```diff
@@ -24,89 +24,122 @@
      </property>
      <layout class="QGridLayout" name="gridLayout_2">
       <item row="0" column="0">
-       <widget class="QLabel" name="stepLabel">
```
chennes, line 27: The grid elements got out of order (because Qt Designer doesn't care about it) -- can you run either @davidosterberg's or my sorter on this? 

- David's (command line): https://github.com/davidosterberg/Qt-UI-file-sorter/
- Mine (GUI): https://github.com/chennes/QtUIXmlCleaner

src/Gui/DlgSettingsNavigation.ui
```diff
@@ -24,89 +24,122 @@
...
-        <item>
-         <property name="text">
-          <string>Bottom left</string>
-         </property>
+        <item row="2" column="1">
+         <widget class="Gui::PrefSpinBox" name="prefCubeSize">
+          <property name="toolTip">
+           <string>Size of the navigation cube</string>
+          </property>
+          <property name="minimum">
```
chennes, line 118: Spinning by one seems a bit small to me, it might be better to spin by five or ten each click.

src/Gui/DlgSettingsNavigation.ui
hyarion: This big diff is due to that I changed the layout type for the navicube pane to use grid layout to allow for aligned columns when I added a new row to it.

I've tried your tool but couldn't install it on macos 10.14.6 (std::filesystem::copy and path seems to require 10.15) and @davidosterberg's tool didn't change the order, it just removed a bunch of `stdset="0"` attributes I've pushed in these changes as an extra clean-up commit before I applied my patches.

Could you try running your (or david's) tool and verify that it rearranges the cell order for you?

hyarion: Apperently `stdset="0"` was important, it fails to build after the tool removes those.

I've removed the prepended cleaning commit to restore the branch to a compilable state.
It would still be interesting to see if you have these issues too.

src/Gui/DlgSettingsNavigation.ui
hyarion: I've added ten to single step [L125](https://github.com/FreeCAD/FreeCAD/pull/4603/files#diff-cd7e65479ecf30d24f0da42072b605500ff01b38688564ce32a732de9ac80e07R125)

wwmayer:
Both builds on Linux and Windows fail probably because of a lacking include statement in the generated ui_ header file.

src/Gui/DlgSettingsNavigation.ui
chennes: Mine definitely won't work here, I don't support nested grid layouts (sorry, I didn't notice that's what was going on here), but I would think that David's would since he actually parses the XML.

donovaly:
My pending PR #4502 includes this addition already. Therefore it would be nice if you could review my PR so that this can be merged.

wwmayer:
Then there is also a crash because the destructor of NaviCubeImplemetation class doesn't detach itself from the parameters.

Here is my branch where all issues are fixed: https://github.com/wwmayer/FreeCAD/pull/new/feature-change-navicube-size

src/Gui/DlgSettingsNavigation.ui
davidosterberg: > @davidosterberg's tool didn't change the order, it just removed a bunch of `stdset="0"` attributes I've pushed in these changes as an extra clean-up commit before I applied my patches.

I have pushed a new version. I know the PR is closed now. But hopefully the issue is fixed now.

davidosterberg: Also, removing stdset="0" is slightly preferable when the setter is defined. Because the code would be a tiny bit more efficient. And the xml is shorter. But as you discovered it does not always work. The script allows to keep the stdset by running the command with --no-remove-stdset

hyarion: @davidosterberg nice, I'll try it next time I poke around in UI files. Btw, wwmayer wrote here https://github.com/FreeCAD/FreeCAD/pull/4603#issuecomment-794086886 about why it didn't work without `stdset="0"`.

PR #4614, opened 2021-03-12
[OpenSCAD] Reimplement surface() to match OpenSCAD

chennes:
The original implementation of the OpenSCAD `surface()` function used a simple B-spline representation for the surface, which created degenerate surfaces with several of OpenSCAD's demo input files. This commit modifies the algorithm to generate a discrete surface identical to the one generated within OpenSCAD itself. It also adds several unit tests to identify future regressions.

Note that PNG input is not yet supported for the surface() function.

---

- [X]  Single module
- [X]  Small feature
- [X]  OpenSCADApp unit tests pass
- [X]  Your pull request is well written
- [X]  No tracker ticket


wwmayer:
When running the tests locally then I get a failure with makeSurfaceVolume() because the file path to open Surface.dat is broken.
The filename starts with .. and then the absolute path to Surface.dat is appended so that the result will be:
../home/user/FreeCAD/Mod/OpenSCAD/OpenSCADTest/data/Surface.dat

The function is called from utility_create_scad (line 204) that is used in the test function test_import_surface

chennes:
Odd, I don't know where that ".." is coming from. I create the path to that file using
```
self.test_dir = join(FreeCAD.getHomePath(), "Mod", "OpenSCAD", "OpenSCADTest", "data")
```
Am I doing that wrong?

Does `test_import_import_stl` work? I'm doing basically the same thing for that one.

wwmayer:
> Am I doing that wrong?

That's fine.

> Does test_import_import_stl work? I'm doing basically the same thing for that one.

All other tests succeed. The  path name `testfile = join(self.test_dir, "Surface.dat").replace('\\','/')` is correct and also the generated scad code in test_import_surface is fine but the paths gets corrupted somewhere later.

wwmayer:
At least in utility_create_scad the path generation is wrong because it uses os.pathsep. This however is not the slash or backslash but a colon or semicolon. The correct character is provided with os.path.sep

Unfortunately, fixing this minor issue doesn't fix the test failure.

chennes:
> Unfortunately, fixing this minor issue doesn't fix the test failure.

Is it the same failure, or something else now?

wwmayer:
> Is it the same failure, or something else now?

It's still the same failure.

wwmayer:
OK, I found the culprit now. It's openscad itself.

In OpenSCADUtils.py there is a function callopenscad that invokes `openscad -o outputfilename inputfilename`
Inside inputfilename the filename is still correct but inside outputfilename the .. are added to the beginning of the filename.

I can break it down to the following steps:
1. Create a file /tmp/input.scad with this content: `surface(file = "/home/user/FreeCAD/Mod/OpenSCAD/OpenSCADTest/data/Surface.dat", center = true, convexity = 5);`
2. Run `/usr/bin/openscad -o /tmp/output.csg /tmp/input.scad `
3. Open the file /tmp/output.csg and the content will be `surface(file = "../home/user/FreeCAD/Mod/OpenSCAD/OpenSCADTest/data/Surface.dat", center = true, invert = false, timestamp = 1615648397);`

Can't you confirm this odd behaviour? So, openscad creates a path from the absolute path relative to the output filename but depending on the current-working-directory the file inside the csg cannot be loaded any more.

openscad version is: OpenSCAD version 2019.05

wwmayer:
From the terminal I can run this specific unit test with:
`./FreeCADCmd -t OpenSCADTest.app.test_importCSG.TestImportCSG.test_import_surface` which fails due to the CWD issue.
If I change to /tmp and run the FreeCAD executable from there with
`../home/user/FreeCAD/bin/FreeCADCmd -t OpenSCADTest.app.test_importCSG.TestImportCSG.test_import_surface` then it works because OpenSCAD is able to find the Surface.dat file

wwmayer:
So I got it working with 0581409fb and 9a105e340

wwmayer:
Argh! The latter change causes to fail most of the other OpenSCAD tests. So, I reverted it and applied the trick to change the cwd only for the test_import_surface

See 371b089ea0

Now for me all unit tests pass.

chennes:
> Can't you confirm this odd behaviour?

I am running the unit tests using the Test workbench in Windows -- could that explain the difference? Your new fix works fine for me, too, but the old code was working on my machine. I'll make sure to run these tests on a Linux install before my next PR. I'm also running a newer version of OpenSCAD right now, but I think when I submitted this PR I was actually on a very old version, like 2015 something.

wwmayer:
> I am running the unit tests using the Test workbench in Windows -- could that explain the difference?

No, that doesn't make a difference. I switched to the command line version to run exclusively that test because on my local build I added many print() statements to figure where the problem is coming from.

> I'm also running a newer version of OpenSCAD right now, but I think when I submitted this PR I was actually on a very old version, like 2015 something.

So, then it's possibly a regression with a newer version.

PR #4620, opened 2021-03-14
App: Fix UserAppData dir on macOS

pilhuhn:
`UserAppData` dir  on macOS is under `~/Library/Preferences/FreeCAD`, but should be under `~/Library/Application Support` is reserved for entries created by the macOS preferences system.

To migrate from the old behaviour: 

```
$ cd ~/Library/Preferences/
$ mv FreeCAD ../Application\ Support/

```
Fixes #0004596
Forum thread: https://forum.freecadweb.org/viewtopic.php?f=22&t=7104&start=10

PR #4623, opened 2021-03-15
App: Refactor the setup of UserAppData path

hyarion:
Disclaimer:
This branch has, at this time, only been tested on macos.
I'm waiting on results for windows and linux, which is also why it is marked as a draft and not ready for review.

---

This refactors the `UserAppData` setup code to use `boost::filesystem` for paths and creating directories.
The motivation behind this change was to simplify the code to make it easier to add migration for the `UserAppData` path change proposed in #4620.
To make maintenance easier I also changed the code paths for windows and *nix platforms.

Even though `std::filesystem` is part of c++17, it isn't implemented in macos 10.14 (which is still supported by Apple).
However `boost::filesystem` is mostly compatible and when 10.14 support is dropped we will be able to change from `boost::filesystem` to `std::filesystem` with a simple search and replace. 

---

- [x] Your pull request is confined strictly to a single module.
- [x] Small change or has a link to a forum post
- [x] Your branch is rebased on latest master git pull --rebase upstream master
- [x] All FreeCAD unit tests are confirmed to pass by running ./bin/FreeCAD --run-test 0
- [x] All commit messages are well-written ex: Fixes typo in Draft Move command text
- [x] Your pull request is well written and has a good description, and its title starts with the module name, ex: Draft: Fixed typos
- [x] Commit message includes ticket if available

### DIFF
--- a/src/App/Application.cpp
+++ b/src/App/Application.cpp
@@ -2775,56 +2775,37 @@ void Application::ExtractUserPath()
         mConfig["UserHomePath"] = userHome.toUtf8().data();
     }
 
-    std::string path = pwd->pw_dir;
-    if (!userData.isEmpty()) {
-        path = userData.toUtf8().data();
-    }
+    boost::filesystem::path appData(pwd->pw_dir);
+    if (!userData.isEmpty())
+        appData = userData.toUtf8().data();
 
-    std::string appData(path);
-    Base::FileInfo fi(appData.c_str());
-    if (!fi.exists()) {
+    if (!boost::filesystem::exists(appData)) {
         // This should never ever happen
-        std::stringstream str;
-        str << "Application data directory " << appData << " does not exist!";
-        throw Base::FileSystemError(str.str());
+        throw Base::FileSystemError("Application data directory " + appData.string() + " does not exist!");
     }
 
-    // In order to write into our data path, we must create some directories, first.
     // If 'AppDataSkipVendor' is defined, the value of 'ExeVendor' must not be part of
     // the path.
-    appData += PATHSEP;
-    appData += ".";
     if (mConfig.find("AppDataSkipVendor") == mConfig.end()) {
-        appData += mConfig["ExeVendor"];
-        fi.setFile(appData.c_str());
-        if (!fi.exists() && !Py_IsInitialized()) {
-            if (!fi.createDirectory()) {
-                std::string error = "Cannot create directory ";
-                error += appData;
-                // Want more details on console
-                std::cerr << error << std::endl;
-                throw Base::FileSystemError(error);
-            }
-        }
-        appData += PATHSEP;
-    }
-
-    appData += mConfig["ExeName"];
-    fi.setFile(appData.c_str());
-    if (!fi.exists() && !Py_IsInitialized()) {
-        if (!fi.createDirectory()) {
-            std::string error = "Cannot create directory ";
-            error += appData;
-            // Want more details on console
-            std::cerr << error << std::endl;
-            throw Base::FileSystemError(error);
-        }
+        appData /= "." + mConfig["ExeVendor"];
+        appData /= mConfig["ExeName"];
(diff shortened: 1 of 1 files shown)


chennes:
I tested on Linux Mint 19 and it appears to be working properly:

- Workbench from old version was found in this version
- Deleted ~/.FreeCAD directory is recreated by this version

chennes:
OK, I tested on Windows 10 now and it works for me there, too:

- Workbench from old version was found in this version
- Deleted ~/AppData/Roaming/FreeCAD directory, and it was recreated

wwmayer:
On my machine I have installed clang-6, clang-10 and clang-11 and even when using the latter and forcing C++17 there is no std::filesystem. For all these compiler versions the classes and functions live in the namespace std::experimental::filesystem.

I realized this issue recently when I wanted to build chennes utility to fix Qt UI files.

hyarion:
> On my machine I have installed clang-6, clang-10 and clang-11 and even when using the latter and forcing C++17 there is no std::filesystem. For all these compiler versions the classes and functions live in the namespace std::experimental::filesystem.

Good to know! I'll guess that will delay switching out boost::filesystem to std::filesystem.
In this PR it doesn't matter, as I've used boost's version.

wwmayer:
Is boost::filesystem unicode aware? The PR will work under Linux but I am not completely sure what happens under Windows if the user name has some unicode characters.

hyarion:
> Is boost::filesystem unicode aware? The PR will work under Linux but I am not completely sure what happens under Windows if the user name has some unicode characters.

This is a good question. I just assumed that since we used `WideCharToMultiByte` it would just work, but you are right, we should explicitly test this before merge.

@chennes could you help me test this please?

wwmayer:
> This is a good question. I just assumed that since we used WideCharToMultiByte it would just work, but you are right, we should explicitly test this before merge.

The function returns a wide char string that is encoded as UTF-8 with WideCharToMultiByte. This string was passed to Base::FileInfo that is unicode aware (i.e. it uses all the _w* variants).

When passing a std::string to boost::filesystem then it probably doesn't handle unicode paths correctly. The question is whether boost offers an overloaded version that accepts a std::wstring.

hyarion:
I interpret the documentation as boost::filesystem::path auto-converts the data from and too any type of string or array of any sized char type.

The native format for boost::filesystem::path seems to be wchar based strings so the conversation might not me needed.

I don’t have a windows computer which I can try this on so it is difficult for me to verify the claims I make.

I would appreciate if anyone could test this for me on windows and let me know the results. Or even better, send me a PR if any changes are needed.

chennes:
> @chennes could you help me test this please?

Do I need to make a user with unicode in their username, or is it enough to make a directory with unicode chars in it?

wwmayer:
For testing purposes the line `SUCCEEDED(SHGetFolderPathW(NULL, CSIDL_PERSONAL, NULL, 0, szPath` can be replaced with something like `szPath = L"C:/Temp/path_with_special_chars_维基百科"`
If underneath this directory the FreeCAD directory is created it should be fine.

wwmayer:
Testing this is a bit more tricky because when compiling the hard-coded string its content is different.

So, to test it you have to set the path in the debugging parameters of the FreeCADMain target. And inside ExtractUserPath() this code snippet is needed:
```C++
    QString path = QString::fromUtf8(_argv[1]);
    std::wstring wstr = path.toStdWString();
    const WCHAR* szPath = wstr.c_str();
```

So, when encoding the string to UTF-8 and passing it to boost::filesystem::path the test will fail. Fortunately, boost::filesystem::path also accepts wide string and szPath can be directly passed. So, we don't even need the function calls of WideCharToMultiByte any more.

wwmayer:
Fixed & Merged.

luzpaz:
Heads up: issue on macOS #6179

PR #4627, opened 2021-03-17
[Stylesheet] ProDark: major bugfix for add-ons and hi-contrast outlines for all buttons 

turn211:
Fixed overlapping QPushButton showstopper bug in all Add-ons that use a fixed grid (not responsive design) Themed all toolbar buttons with Hi Contrast border while hovering and focus. Styled Open/Close buutons accordingly the same Hi Contrast border statically. Very easy to use with or without eye disabilities.

---

### DIFF
--- a/src/Gui/Stylesheets/ProDark.qss
+++ b/src/Gui/Stylesheets/ProDark.qss
@@ -74,10 +74,23 @@ QToolBar * {
     padding: 0px;
 }
 
+/*==================================================================================================
+Style Links
+==================================================================================================*/
+QLabel[haslink="true"] {
+    color: #55aaff;
+}
+
+Gui--UrlLabel {
+    color: #55aaff;
+}
 
 /*==================================================================================================
 Main window
 ==================================================================================================*/
+QWidget {
+    background-color: #333333;
+}
 QMainWindow,
 QDialog,
 QDockWidget,
@@ -271,7 +284,7 @@ QToolBar::separator:vertical {
 Group box
 ==================================================================================================*/
 QGroupBox {
-    color: rgba(255,255,255,120);
+    color: #bcbcbc;
     border:1px solid rgba(255,255,255,20); /* lighter than its own border-color */;
     border-radius: 2px;
     margin-top: 10px;
@@ -300,7 +313,7 @@ Tooltip
 ==================================================================================================*/
 QToolTip {
     color: #ffffff;
-    background-color: #1e1e1e;
+    background-color: #2a2a2a;
     /*opacity: 90%; doesn't correctly work */
     padding: 4px;
     border-radius: 2px; /* has no effect */
@@ -319,10 +332,10 @@ QDockWidget {
 
 QDockWidget::title {
     text-align: center;
-    background-color: rgba(0,0,0,40);
-    border: 4px solid #333333; /* fix to simulate margin between this :title and tabs */ /* same as main background color */
-    border-radius: 6px; /* bigger than normal due to previous border fix */
-    padding: 4px 0px; /* also needed because of previous border fix */
+    background-color: #2a2a2a;
+    border-bottom: 4px solid #333333; /* fix to simulate margin between this :title and tabs */ /* same as main background color */
+    margin-left: 7px;
+    margin-right: 7px;
 }
 
 QDockWidget::close-button,
@@ -344,17 +357,18 @@ QDockWidget::float-button {
 
 QDockWidget::close-button:hover,
 QDockWidget::float-button:hover {
-    background-color: rgba(0,0,0,60);
+    background-color: #557bb6;
 }
 
 QDockWidget::close-button:pressed,
 QDockWidget::float-button:pressed {
-    background-color: rgba(0,0,0,120);
+    background-color: #42608d;
+    border: 2px solid #76acfd
 }
 
 /* fix for Python Console (probably there is a smarter way to arrive to it) */
 QDockWidget > QFrame {
-    background-color: #3C3C3C;
+    background-color: #3c3c3c;
 
     border: 6px solid #333333;
 }
@@ -366,6 +380,7 @@ Progress bar
 QProgressBar,
 QProgressBar:horizontal {
     color: white;
+    min-height: 24px;
     background-color: rgba(0,0,0,70);
     text-align: center;
     border: 1px solid rgba(0,0,0,140);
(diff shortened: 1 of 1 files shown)


turn211:
I don't how commit 4d7124d got added. I rebased locally and git said it was gone. I only want commit 
f42be44

src/Gui/Stylesheets/ProDark.qss
```diff
@@ -1470,38 +1514,182 @@ Push button
...
+    border: 1px solid #f5f5f5;
+    background-color: #557BB6;
+}  
+
+QPushButton:pressed#buttonAddLevel,
+QPushButton:pressed#buttonAddProxy,
+QPushButton:pressed#buttonDelete,
+QPushButton:pressed#buttonToggle,
+QPushButton:pressed#buttonIsolate,
+QPushButton:pressed#buttonSaveView,
```
hyarion, line 1640: Would it be possible to make more generic selectors so you don't need to be so specific and name all buttons?
Or is there a reason why you need to do it this way?

Doing it this way will require a style update if a new button is added anywhere in the UI and I don't see how this will scale and how you will be able to keep up with changes people make.

I think a lot of these could be substituted with something like `QWidget#somename QPushButton:pressed {` to match all pressed push buttons inside a specific named widget. Or maybe some other combination of descendant or child selectors. Documentation for these can be found here in case you haven't read it already: https://doc.qt.io/qt-5/stylesheet-syntax.html#selector-types

If it isn't possible, let me know if there is something that needs to be changed in freecad for this to work. I'll be happy to help.

hyarion (changes requested):

hyarion:
> I don't how commit [4d7124d](https://github.com/FreeCAD/FreeCAD/commit/4d7124de312ad8d788c2b6b2527c7562826343fd) got added. I rebased locally and git said it was gone. I only want commit
> [f42be44](https://github.com/FreeCAD/FreeCAD/commit/f42be44e651498b4f56a4a0075032dad3ea9f9d6)

If you're used to git in the terminal you can rebase interactively, see: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

src/Gui/Stylesheets/ProDark.qss
turn211: No you are right. It won't scale. But as I see, there are no guide lines for Add-on or Macros. Wild West. Problem here is everyone uses QPushButtons when QToolbarButtons would be better choice from a UX/UI design point of view. The QPushButtons i separate out for Ui design theming would become QToolbarButtons eliminating hundreds of lines of code and be scalable. Take example Defeaturing Tools. A fixed Grid base app with no text on the buttons. A perfect example to use QToolbarButtons. This is a non scalable, non responsive designed app. The above code you quoted above is from the BIM View Manager. These too could have been QToolbarButtons. If you look closely I've separated these even more to add a left margin. This not scalable but i know by following FreeCAD for years it is not going to be fixed in hard code either. What I'm saying is I am trying to dot all the i's that weren't dotted in hard code with stylesheets to make FreeCAD look polished and professional. All the hard work that was done by Pablo Gil was a great start but all the shipping stylesheets have major issues. This commit should be merge strictly for the showstopper bugfixes for the add-ons that were brought to my attention in the forums. I didn't even know they existed. I will consolidate code as you suggest once i have a polished design that works with the multi cultural UI/UX design approaches of FreeCAD. I am going to do a full rewrite from scratch that is not based off any existing stylesheets but as i own and run a construction business it might take a while. Thanks for your help.

luzpaz: Well it may be time to start hardening up code guidelines to get addons accepted in to the addon manager?

turn211: Yes and no. Of course it would be a Great idea but considering the above Bim Addon comes from a core FreeCAD developer maybe it's not important. I see most YouTubers and makers don't use a Stylesheet and don't care about Ui/Ux looks only features (so you don't want to bite the hand so to speak and limit add-on creators) Linux KDE users don't need a Stylesheet as they can use a Global theme. So without statistics i don't know if it's worth tipping the apple cart. Do you the majority of FreeCAD users even care about UX/UI design as long as it has features they want? Even with my very hand pick isolated code elements if one the addons adds a new feature it could be added faster than typing this comment so i wouldn't worry too much about scalability at this point in time.

hyarion: @turn211: I don't see how patching these issues in stylesheets is a good way forward. Doing that just means that the issues will continue to grow. It might be ok for you to fix this theme (and faster too) but other stylesheets would need to make the same modifications as well. I'm also thinking about what would happen to this style sheet if you can't stay maintaining it. Can others jump in in your place if that would happen or will it degrade and need to be removed?
I think that it is important to think about maintainability and I think it can be improved in this case.

I don't want to be in your way of improving the UI, so I will lift the Change Requested status on this PR.

Regarding the issues you're trying to work around. Could you write them down in a list? ..and maybe including some screenshots of the issues / workbench too?
@luzpaz do you have an idea on how to coordinate this? where should the list live? should we write issues in the external workbenches or keep the list of issues in all workbenches in one single document?
I'm willing to help fixing issues that gets reported, but I hope workbench authors will aid in this work too.

turn211: I'm going to close this PR. I'll create another. I've patched all concerned QSS using scalable coding as best as the underlying hard code allows me to do. The ProDark theme is built on top of Pablo Gils Darker-Blue and is very modified and not compatible with any of the stock themes. The shipping themes are broken hence why this theme was created. (Undock Combo View using any of the shipping dark themes ( no painting ), the BIM View Manager icons are huge, etc) i can't maintain all of the shipping themes only mine. As for the Workbench issues nothing short of UI/UX guidelines and rewrites will fix them. I bring up Defeaturing Tools again. Look at the Close button. A designer has to jump through hoops to make this look good in a stylesheet. All of the shipping stylesheets have overlapping icons on these buttons. The ProDark theme corrects all these issues but steps way out of the shipping stylesheets. If it doesn't work for the Main FreeCAD branch I say remove it. I'll keep on impring it on my Repository. As for workarounds i could point them all out but in the end somebody has to correct them. As I am the one creating the stylesheet it is moot. I've still got unanswered questions still in the Forums.

turn211:
Closing PR for recoding

PR #4629, opened 2021-03-17
[OpenSCAD] Add angle and $fn to rotate_extrude()

chennes:
As pointed out in [Issue #0004353](https://tracker.freecadweb.org/view.php?id=4353), the OpenSCAD Workbench does not correctly implement the `angle` parameter to `rotate_extrude()` (it's a relatively recent addition to OpenSCAD), nor does it attempt to do anything with a specified `$fn` ("number of fragments," akin to an angular resolution for discretized geometry). This PR adds both features. 

To add `$fn` handling, the code from the cylinder extrusion was mimicked, allowing FreeCAD to create perfect, smooth representations when `$fn` exceeds a user specified value in Preferences, but creating OpenSCAD's discrete surface representation when using a lower `$fn`. Note that this determination is made at creation time: if the user later increases the `segments` parameter in the new object, it will remain prismatic regardless of how high the value is made.

The second commit also implements this for `rotate_extrude()` with a `file` parameter, though this does not seem to be available in OpenSCAD itself.

---

- [X]  Discussed on the FreeCAD forum
- [X]  Commit messages include `issue #<id>` or `fixes #<id>`°

### DIFF
--- a/src/Mod/OpenSCAD/OpenSCADFeatures.py
+++ b/src/Mod/OpenSCAD/OpenSCADFeatures.py
@@ -389,7 +389,7 @@ class Frustum:
 class Twist:
     def __init__(self, obj,child=None,h=1.0,angle=0.0,scale=[1.0,1.0]):
         obj.addProperty("App::PropertyLink","Base","Base",
-                        "The base object that must be tranfsformed")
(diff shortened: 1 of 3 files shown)


wwmayer:
Merged.

PR #4632, opened 2021-03-17
[GUI] Refactoring and cleanup of the About window

chennes:
Removes old Qt code, restores the conditional inclusion of the 3Dconnexion license information, shortens the displayed hash, and links the hash to its repo.

---

- [X]  Discussed on the FreeCAD forum
- [X]  Unit tests pass
- [X]  No tracker ticket

### DIFF
--- a/src/Doc/LICENSE.html
+++ b/src/Doc/LICENSE.html
@@ -575,7 +575,9 @@ DAMAGES.
 
 <h4>END OF TERMS AND CONDITIONS</h4>
 
-<h2>3D Mouse Support</h2>
-<p>Development tools and related technology provided under license from 3Dconnexion.(c) 1992 - 2012 3Dconnexion. All rights reserved.</p>
+<hr/>
+
+<!--PLACEHOLDER_FOR_ADDITIONAL_LICENSE_INFORMATION-->
+
 </body>
 </html>
--- a/src/Gui/AboutApplication.ui
+++ b/src/Gui/AboutApplication.ui
@@ -146,7 +146,7 @@
             <number>6</number>
            </property>
            <item row="6" column="1">
-            <widget class="QLabel" name="labelBuildHash">
+            <widget class="Gui::UrlLabel" name="labelBuildHash">
              <property name="text">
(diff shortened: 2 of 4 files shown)


donovaly:
I get now this error:

```
D:\FreeCADGit\src\Gui\Splashscreen.cpp(347,61): error C2429: language feature 'init-statements in if/switch' requires compiler flag '/std:c++17'
5>D:\FreeCADGit\src\Gui\Splashscreen.cpp(350,64): error C2429: language feature 'init-statements in if/switch' requires compiler flag '/std:c++17'
```

Do we really now require C++17? I thought we raise only to C++14. Going to C++17 is fine with me, but then also the CMake entry for FreeCADGui needs to be changed

chennes:
> Do we really now require C++17?

We do, and the latest CMake should be requiring it. You might need to delete your CMakeCache.txt to get it to update, though.

donovaly:
> We do, and the latest CMake should be requiring it. You might need to delete your CMakeCache.txt to get it to update, though.

My mistake was that I came from building the FC 0.19 release binaries and thus my MSVC setup was C++14. When I switch to master, the C++ settings are not updated in MSVC.
Can this maybe be improved, because I see only a switch to the C++ level "default", but this is in MSVC 2019 C++14.

chennes:
> Can this maybe be improved, because I see only a switch to the C++ level "default", but this is in MSVC 2019 C++14.

Do you mean that's what you see in the CMake GUI? Or are you looking at something in Visual Studio?

donovaly:
I am only speaking about MSVC. Here on my second PC where I can compile, everything is fine when I only compile master. When i then switch to the 0.19 branch then MSVC still has C++17 as "Std". So in fact when switching, one must manually re-run CMake to get MSVC's "Std" to C++14.

I'll document this in the Wiki.

chennes:
> I'll document this in the Wiki.

That would be great, thanks. I don't know enough cMake to figure out how to make it handle that automatically -- I've just got two copies of FreeCAD on my system, one the head, and one the 0.19 branch, so I am not switching the same compilation between the two.

wwmayer:
You can use `git worktree add <path> <branch>`
This way you have one FreeCAD repository but the sources for v0.19 are in a separate directory. So, you don't have to switch between master and the release branch any more

donovaly:
I updated now https://wiki.freecadweb.org/Compile_on_Windows

> This way you have one FreeCAD repository but the sources for v0.19 are in a separate directory.

Thanks.

PR #4635, opened 2021-03-18
[OpenSCAD] Improve helical extrusion, fix scaling

chennes:
In `linear_extrude()`, modify the auxiliary spine to be a true helix, resulting in a much smoother extrusion when a twist angle is applied. This also corrects a bug in the scaling during linear extrusion, where non-uniform scaling was not handled correctly.

Several unit tests are modified to use an analytical volume calculation for comparison, and unit tests are added for simple (slice) projections.

Finally, a few minor PEP8 changes were applied.

---

- [X]  Pull request is confined strictly to a single module
- [X]  Small bug fixes
- [X]  Rebased
- [X]  No tracker ticket

### DIFF
--- a/src/Mod/OpenSCAD/OpenSCADFeatures.py
+++ b/src/Mod/OpenSCAD/OpenSCADFeatures.py
@@ -390,7 +390,7 @@ class Twist:
     def __init__(self, obj,child=None,h=1.0,angle=0.0,scale=[1.0,1.0]):
         obj.addProperty("App::PropertyLink","Base","Base",
                         "The base object that must be tranfsformed")
-        obj.addProperty("App::PropertyAngle","Angle","Base","Twist Angle in degrees") #degree or rad
+        obj.addProperty("App::PropertyAngle","Angle","Base","Twist angle") #degree or rad
         obj.addProperty("App::PropertyDistance","Height","Base","Height of the Extrusion")
         obj.addProperty("App::PropertyFloatList","Scale","Base","Scale to apply during the Extrusion")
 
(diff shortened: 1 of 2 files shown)


src/Mod/OpenSCAD/OpenSCADFeatures.py
```diff
@@ -404,45 +404,45 @@ def execute(self, fp):
...
+                for wire1,wire2 in zip(lower_face.Wires,upper_face.Wires):
+                    pipe_shell = Part.BRepOffsetAPI.MakePipeShell(spine)
+                    pipe_shell.setSpineSupport(spine)
+                    pipe_shell.add(wire1)
+                    pipe_shell.add(wire2)
+                    pipe_shell.setAuxiliarySpine(auxiliary_spine,True,0)
+                    print(pipe_shell.getStatus())
+                    assert(pipe_shell.isReady())
+                    pipe_shell.build()
                     faces.extend(pipeshell.shape().Faces)
```
wwmayer, line 446: pipeshell must be renamed to pipe_shell 2490629a4a

wwmayer:
Something is weird with the test function test_import_projection because it goes to full CPU load and doesn't seem to finish the operation (it's inside MultiCommon::execute). I have extracted the relevant code:

```
import tempfile
import importCSG


def utility_create_scad(scadCode, name):
  with tempfile.TemporaryDirectory() as temp_dir:
    filename = temp_dir + os.path.sep + name + ".scad"
    f = open(filename,"w+")
    f.write(scadCode)
    f.close()
    return importCSG.open(filename)

# test_import_projection
#
base_shape = "linear_extrude(height=5,center=true,twist=90,scale=0.5){square([1,1],center=true);}"
hole = "cube([0.25,0.25,6],center=true);"
cut_shape = f"difference() {{ {base_shape} {hole} }}"

doc = utility_create_scad(f"projection(cut=true) {cut_shape}", "projection_slice_square_with_hole")
```

chennes:
What system are you testing on? In the OpenSCAD workbench, does this code work?
```
projection(cut=true) {
    difference() { 
        linear_extrude(height=5,center=true,twist=90,scale=0.5){
            square([1,1],center=true);
        } 
        cube([0.25,0.25,6],center=true);
        }
    }
```

wwmayer:
> What system are you testing on?

The problem could be OCCT 7.3 because it hangs in BRepAlgoAPI_Common.

wwmayer:
I have tried it with an OCCT 7.5 build and there it works just fine.

chennes:
> The problem could be OCCT 7.3 because it hangs in BRepAlgoAPI_Common.

OK, I'll look into it, thanks for investigating.

PR #4645, opened 2021-03-21
[Part] Improve Part.Shape methods python documentation

FlachyJoe:
Format UserDocu to follow the template:
```
 Description
 method(args) -> return
 --
 More infos
```

### DIFF
--- a/src/Mod/Part/App/TopoShapePy.xml
+++ b/src/Mod/Part/App/TopoShapePy.xml
@@ -58,8 +58,10 @@ exportStep(filename)
     </Methode>
     <Methode Name="exportBrep" Const="true">
       <Documentation>
-        <UserDocu>Export the content of this shape to an BREP file. BREP is a CasCade native format.
+        <UserDocu>Export the content of this shape to an BREP file.
 exportBrep(filename)
+--
+BREP is an OpenCasCade native format.
         </UserDocu>
       </Documentation>
     </Methode>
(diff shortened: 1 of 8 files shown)


src/Mod/Part/App/TopoShapePy.xml
```diff
@@ -30,74 +30,98 @@ Sub-elements such as vertices, edges or faces are accessible as:
...
+        </UserDocu>
       </Documentation>
     </Methode>
     <Methode Name="exportBrepToString" Const="true">
       <Documentation>
-        <UserDocu>Export the content of this shape to a string in BREP format. BREP is a CasCade native format.</UserDocu>
+        <UserDocu>Export the content of this shape to a string in BREP format.
+exportBrepToString() -> string
+--
+BREP is a CasCade native format.</UserDocu>
```
luzpaz, line 78: `CasCade` ? Why not `OpenCasCade`

luzpaz:
Will these also get translated ?

src/Mod/Part/App/TopoShapePy.xml
FlachyJoe: Historical reason I suppose.

FlachyJoe:
> Will these also get translated ?

I can't find any translated python embedded doc.
This file generate .h and .cpp at build time and I think we don't want localized build so lets keep it English only.

FlachyJoe:
Please keep this open, I'll format every TopoShape*Py.xml

Does it need [skip ci] ?

PR #4646, opened 2021-03-21
[FEM] add missing material cards to CMakeLists.txt

donovaly:
This fixes issue 1. reported here: https://forum.freecadweb.org/viewtopic.php?f=18&t=56912#p489666

### DIFF
--- a/src/Mod/Material/CMakeLists.txt
+++ b/src/Mod/Material/CMakeLists.txt
@@ -25,16 +25,114 @@ SET (MaterialLib_Files
     StandardMaterial/None.FCMat
     StandardMaterial/ABS-Generic.FCMat
     StandardMaterial/Acrylic-Glass-Generic.FCMat
+    StandardMaterial/AlMg3F24.FCMat
+    StandardMaterial/AlMgSi1F31.FCMat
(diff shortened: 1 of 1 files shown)


berndhahnebach:
[quote=uwestoehr post_id=489666 time=1616358762 user_id=23505]
This thread is intended to sort out and fix the different material issues.

List of issues:

1. in the material dialog one cannot load all existing material cards. For example we have already many cards for steel, but they don't appear in the dropdown list. One can also not load them via the material editor because they are not listed anywhere.
-> Will be fixed with https://github.com/FreeCAD/FreeCAD/pull/4646

[/quote] AFAIR: we added not all cards to the rolling dev version. But at the release all cards should be available without this commit.  There have been discussion on the forum. 

@sgrogan

donovaly:
> AFAIR: we added not all cards to the rolling dev version. But at the release all cards should be available without this commit. 

Yes, the are i the release, but most FEM users use the rolling release, for example to benefit from the latest Gmsh features.

I don't see why a few small textfiles should not appear in the rolling releases. When you simulate the different steel material cards are very valuable.

PR #4648, opened 2021-03-22
[FEM] fix specific heat of air

donovaly:
- also adjust values slightly according to various sources

Since a factor 1000 is a big mistake and air is the most common gas, please commit this also for FC 0.19


donovaly:
Please also commit this fix also to the 0.19 branch

PR #4647, opened 2021-03-22
[Base] add electrical conductivity

donovaly:
- this is necessary to perform electrodynamics simulations. Since the typical values are in the range of 1e6 S/m, we need the new units kS and MS

- since this is for simulations, also add electrical conductivity to the FEM material cards

PR #4649, opened 2021-03-22
[FEM] add material cards for gases

donovaly:
also add electrical conductivity for air and water

To be merged after #4648 and #4647 are in.


luzpaz:
@berndhahnebach the PR states to merge this after #4647 (which is still unmerged). Is that a problem ?

PR #4651, opened 2021-03-22
[Do not merge][Travis] Update CI to use new libpack location

chennes:
The Windows Travis builds were failing because the old libpack is no longer available in the specified location. This PR updates Travis to pull the Windows Libpack from the 0.19.1 release tag -- note that this also updates the libpack version and therefore all included libraries.


hyarion (approved):

luzpaz:
We should merge this ASAP. Thanks chennes!

chennes:
Just an update: this is now building and running on Travis/Windows, but the FEM test app is not loading.

chennes:
I'm marking this as ready to review since it does fix the problem it claims to: but it also leaves us with another problem, since the FEM unit test won't load.

hyarion:
Please add [Do not merge] to the title while testing the windows builds so it doesn’t get merged by accident.

wwmayer:
I will cherry-pick a0620dc2cc97aea1c8e and add it to master because as it is now all Windows builds fail

luzpaz:
> Please add [Do not merge] to the title while testing the windows builds so it doesn’t get merged by accident.

Or just set the PR to a [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/)

chennes:
@wwmayer has integrated the only real contribution from this PR (the rest was just working on testing Travis), so I'm closing it now that he's found the solution to the FEM stuff.

wwmayer:
In I have added the system32 directory in e825148cf454 to the search path which didn't solve the problem. My alternative idea (8277b3f9e9) was to disable Netgen support to avoid to try to load nglib.dll and this did the trick.

Conclusion: apparently on the travis system the files opengl32.dll and glu32.dll are missing. Can this really be true?

sgrogan:
For opengl32.dll  this is a problem  reported for Win virtual machines. Qt provides openglsw.dll which is in the Libpackx. If this is renamed opengl.dll it is known to solve the problem, sometimes.

PR #4654, opened 2021-03-23
[FEM] add mesh algorithms

donovaly:
- Gmsh has since a while 2 new mesh algorithms
- on the other hand the two algorithms "Frontal Delaunay" and "Frontal Hex" are deprecated and not listed anymore in the docs: https://gmsh.info/doc/texinfo/gmsh.html#index-Mesh_002eAlgorithm3D
and indeed all attempts with these 2 algorithms fail

Forum thread: https://forum.freecadweb.org/viewtopic.php?f=18&t=56971

### DIFF
--- a/src/Mod/Fem/femmesh/gmshtools.py
+++ b/src/Mod/Fem/femmesh/gmshtools.py
@@ -101,6 +101,8 @@ class GmshTools():
             self.algorithm2D = "7"
         elif algo2D == "DelQuad":
             self.algorithm2D = "8"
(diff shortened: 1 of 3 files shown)


donovaly:
> If you are loading an old fcstd file containing _algo3D_ == "Frontal Delaunay", ._algorithm3D_ will default silently to "Automatic" without changing dropdown menu of properties view.
> There should some handling of depreciated properties in _def onDocumentRestored(self, obj)_

Thanks, this is now fixed.

PR #4661, opened 2021-03-24
[Draft] Eliminate required GUI for importDXF

chennes:
Bug reported by forums member tarrox -- the importDXF code assumes the existence of the GUI, but does not actually need it (a `FreeCAD.GuiUp` check suffices to protect the one line of code that was using the view provider).

---

- [X]  Single module
- [X]  Discussed in forum: [[BUG] Error using importDXF without GUI](https://forum.freecadweb.org/viewtopic.php?f=23&t=56711)
- [X]  Draft unit tests pass
- [X]  No tracker ticket

### DIFF
--- a/src/Mod/Draft/importDXF.py
+++ b/src/Mod/Draft/importDXF.py
@@ -55,7 +55,7 @@ import six
 import FreeCAD
 import Part, Draft, Mesh
 import DraftVecUtils, DraftGeomUtils, WorkingPlane
-from Draft import _Dimension, _ViewProviderDimension
+from Draft import _Dimension
 from FreeCAD import Vector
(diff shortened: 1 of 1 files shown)


hyarion (approved):
There is a variable `gui = FreeCAD.GuiUp` in this file that could be used instead but that doesn't seem to be consistently used.
I wouldn't mind if it was cleaned up and only one way was used, but either way I think this PR is ok to be merged.

yorikvanhavre:
this `gui = FreeCAD.GuiUp` is probably a leftover from some earlier structure... In any case indeed this should totally work without the GUI. Thanks for the fix!

PR #4662, opened 2021-03-24
Path: Added postprocessor for UC-CNC

luzpaz:
Forum: https://forum.freecadweb.org/viewtopic.php?f=15&t=56987  

edit: I'm helping papaathome submit their PR to the FreeCAD repository. All credit goes to @papaathome

### DIFF
--- a/src/Mod/Path/PathScripts/post/uccnc_post.py
+++ b/src/Mod/Path/PathScripts/post/uccnc_post.py
@@ -1,0 +1,671 @@
+# -*- coding: utf-8 -*-
+
+#***************************************************************************
+#*                                                                         *
+#*   based upon linuxcnc_post.py (c) sliptonic (shopinthewoods@gmail.com)  *
+#*                                                                         *
+#*   changed, but not enough to claim copyrights 2019-2021                 *
+#*   maintainer: A.H.M. Steenveld                                          *
+#*                                                                         *
+#*   This program is free software; you can redistribute it and/or modify  *
+#*   it under the terms of the GNU Lesser General Public License (LGPL)    *
+#*   as published by the Free Software Foundation; either version 2 of     *
+#*   the License, or (at your option) any later version.                   *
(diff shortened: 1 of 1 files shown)


papaathome:
@luzpaz 
I checked the build log to see why it is failing and two things came up (for me).

The first one is simple. Around line 10986 - 11003 in the build log files are listed in the .../Path/PathScripts/port/ folder.
The script uccnc_port.py is missing.
My explanation (to be checked):
I did add the uccnc_port.py file but have not looked at the build process at all. The build process is not doing anything with the new script.

The second one is more confusing (for me) and I have no explanation for it.
At the end of the build log there is a message about a .py file that cannot be parsed.
The file in question is not near the area of ../Path/ and the chances that I by accident did hit the file is very small.
Capture

I do not think it is a problem related to my actions.
It would be interesting to know if the same problem also occurs in other builds but I have not checked it.

recap:
A simple suggestion to create a 'Pull Request' turns out to be a bit more work than adding a file and a request to take it up in the standard distribution. So far I'm enjoying the experience and I will have a look into the build process and will try to get this working.

I also will try again to be able to build FreeCAD on my machine. Last year I tried but it failed because I could not solve all dependencies for a windows based build.

chennes:
That second bug was introduced in commit af3643853cc0d5fbcbdbff1f84d39e8d6f0bc218, just a few hours ago.

chennes:
Also, I should note that there are two issues that are preventing the Travis Windows builds from succeeding -- the first is simply that the dependencies have been moved. But the second is that the dependency library includes an updated version of Python, and this appears to be resulting in Travis being unable to run the unit tests for FEM, which counts as a test failure. So it's not really the fault of a specific commit here.

luzpaz:
Closing in favor of #4664

papaathome:
> 
> 
> Closing in favor of #4664

Thanks :) I was a bit slow on that action.

PR #4665, opened 2021-03-24
[skip ci] [FEM] fix typo

donovaly:
- the property has the name "Element order" thus the dialog should use the same name to avoid confusions

- also uniform naming for the element dimension and remove the redundant "Mesh" (is already in the UI groupBox heading


luzpaz:
You can probably add `[skip ci]` for this type of commit next time.

donovaly:
> You can probably add `[skip ci]` for this type of commit next time.

Thanks. I'll do so.

PR #4666, opened 2021-03-25
[skip ci] [FEM] don't hide Gmsh mesh by default

donovaly:
it is very annoying that the mesh is hided when closing the task dialog because one often has to play with the Gmsh properties and need to see the mesh.
See also https://forum.freecadweb.org/viewtopic.php?f=18&t=56401&start=10#p489734


berndhahnebach:
It is not annoying for me :-) But for new user this might be strange because they make a new mesh but it is not visible. Thus I am ok with this.

PR #4668, opened 2021-03-25
[GUI] Remove explicit triangle styling from tabs

chennes:
On Windows the lower tabs of the PropertyView were being set to triangular, which breaks stylesheets. This removes that code, so they now take on whatever the standard shape is on the system. Fixes [#0004599](https://tracker.freecadweb.org/view.php?id=4599).

---

- [X]  Single module
- [X]  Forum discussion: [0.19 Release breaks all shipping Stylesheets Tab theming](https://forum.freecadweb.org/viewtopic.php?f=34&t=56875)
- [X]  Commit messages include `issue #<id>` or `fixes #<id>`

### DIFF
--- a/src/Gui/PropertyView.cpp
+++ b/src/Gui/PropertyView.cpp
@@ -89,9 +89,6 @@ PropertyView::PropertyView(QWidget *parent)
     tabs = new QTabWidget (this);
     tabs->setObjectName(QString::fromUtf8("propertyTab"));
     tabs->setTabPosition(QTabWidget::South);
-#if defined(Q_OS_WIN32)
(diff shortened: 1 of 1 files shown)


donovaly:
@wwmayer , can you please also merge this also to the 0-19 branch?
Due to a Qt bug we need to update the binaries for FC 0.19 that broke all styles on Windows. To be able to distinguish the binary release with the broken styles and the new one using Qt 5.15.2, we need a 0.19.2 release. And while we are at the styles, this fix should go in for 0.19.2.

Forum thread: https://forum.freecadweb.org/viewtopic.php?f=9&t=56531&start=30

PR #4682, opened 2021-03-28
[Sketcher] Box corrections 

abdullahtahiriyo:
For Werner review (only the second commit. It is just two lines).

Thread:
https://forum.freecadweb.org/viewtopic.php?f=10&t=56550

Reason:
The proposed fix appears to solve the bug, but I do not understand why. I do not know why it might be necessary to switch the render type to image and back to native. Not doing it appears to fix the bug.

### DIFF
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -948,9 +948,6 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                     doBoxSelection(prvCursorPos, cursorPos, viewer);
                     rubberband->setWorking(false);
 
-                    //disable framebuffer drawing in viewer
-                    const_cast<Gui::View3DInventorViewer *>(viewer)->setRenderType(Gui::View3DInventorViewer::Native);
-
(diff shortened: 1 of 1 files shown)


wwmayer:
I can confirm the exact same behaviour under Linux (while under Windows it works). When looking at the git history then the frame buffer handling has been introduced with 72cab0c0022 where the function was called setRenderFramebuffer that later has been renamed and extended.

I don't know why setRenderFramebuffer has been used there but maybe there were issues with the Qt4 OpenGL widget. At least with Qt5 everything still seems to work as expected and fixes the issue.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1372,7 +1369,6 @@ bool ViewProviderSketch::mouseMove(const SbVec2s &cursorPos, Gui::View3DInventor
         case STATUS_SKETCH_StartRubberBand: {
             Mode = STATUS_SKETCH_UseRubberBand;
             rubberband->setWorking(true);
-            viewer->setRenderType(Gui::View3DInventorViewer::Image);
```
wwmayer, line 1375: Removing these two lines look OK

abdullahtahiriyo:
@wwmayer 

Thanks!

PR #4684, opened 2021-03-28
[Gui] fix compilation after commit b55cc8aec

donovaly:
Fix mistake of after commit b55cc8aec to be able to compile again


wwmayer:
Merged.

PR #4686, opened 2021-03-29
[Gui] add recompute icon to context menu

donovaly:
as discussed here: https://forum.freecadweb.org/viewtopic.php?f=34&t=57139&start=10

context menu with this PR:
8iow53XAgj


wwmayer:
Merged (c712b2feb6)

PR #4687, opened 2021-03-29
Gui: Enable compression of tablet motion events

matthijskooijman:
By default (on platforms that support it, X11 and Windows currently) QT
applies compression for high frequency events (mouse move, touch, window
resizes) to keep things smooth even when handling the event takes a
while (e.g. to calculate snapping).

However, tablet pen move events (and mouse move events synthesised from
those, which is what FreeCAD uses) are not compressed by default (to
allow maximum precision when e.g. hand-drawing curves), leading to
unacceptable slowdowns using a tablet pen.

This commit enable compression for tablet events here to solve that and
make use of a tablet just as smooth as a mouse with FreeCAD.

This can (and likely will) lead to some movement events being dropped,
but since there is no freeform curve drawing tool, that should not be
problematic (and if it is ever added, it could still work without
compression if the mouse movement event handler is written to be fast
enough).

- [ ] ~~In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum~~ simple bugfix
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0` (waiting for compilation to complete after rebase, patch was manually tested on few days old master)
- [ ]  ~~Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the [FreeCAD bug tracker issue number](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) in case a particular commit solves or is related to an existing issue on the tracker. Ex: `Draft: fix typos - fixes #0004805`~~ no issue to close

Some additional info:
 - This issue was previously discussed on the forum: https://forum.freecadweb.org/viewtopic.php?f=8&t=45046&p=490962#p490962
 - The QT implementation of compression is [here for X11](https://github.com/qt/qtbase/blob/2801c1988742b18557c581a544a70a02f4faefa1/src/plugins/platforms/xcb/qxcbconnection.cpp#L893) and [here for Windows](https://github.com/qt/qtbase/blob/420e27c9efc1ef1a6400dc287345e897036abc71/src/plugins/platforms/windows/qwindowspointerhandler.cpp#L88).
 - For reference, here's an excerpt from [the relevant QT docs](https://doc.qt.io/qt-5/qt.html#ApplicationAttribute-enum):

Qt::AA_CompressHighFrequencyEvents | Enables compression of certain frequent events. On the X11 windowing system, the default value is true, which means that QEvent::MouseMove, QEvent::TouchUpdate,  and changes in window size and position will be combined whenever they  occur more frequently than the application handles them, so that they  don't accumulate and overwhelm the application later. On Windows 8 and  above the default value is also true, but it only applies to touch  events. Mouse and window events remain unaffected by this flag. On other  platforms, the default is false. (In the future, the compression  feature may be implemented across platforms.) You can test the attribute  to see whether compression is enabled. If your application needs to  handle all events with no compression, you can unset this attribute.  Notice that input events from tablet devices will not be compressed. See  AA_CompressTabletEvents if you want these to be compressed as well.  This value was added in Qt 5.7.
-- | --
Qt::AA_CompressTabletEvents | Enables  compression of input events from tablet devices. Notice that  AA_CompressHighFrequencyEvents must be true for events compression to be  enabled, and that this flag extends the former to tablet events.  Currently supported on the X11 windowing system, Windows 8 and above.  The default value is false. This value was added in Qt 5.10.

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -1833,6 +1833,20 @@ void Application::runApplication(void)
         QApplication::setAttribute(Qt::AA_UseSoftwareOpenGL);
     }
 
+    #if (QT_VERSION >= QT_VERSION_CHECK(5, 10, 0))
+        // By default (on platforms that support it, see docs for
+        // Qt::AA_CompressHighFrequencyEvents) QT applies compression
+        // for high frequency events (mouse move, touch, window resizes)
+        // to keep things smooth even when handling the event takes a
+        // while (e.g. to calculate snapping).
+        // However, tablet pen move events (and mouse move events
+        // synthesised from those) are not compressed by default (to
+        // allow maximum precision when e.g. hand-drawing curves),
+        // leading to unacceptable slowdowns using a tablet pen. Enable
+        // compression for tablet events here to solve that.
+        QCoreApplication::setAttribute(Qt::AA_CompressTabletEvents);
+    #endif
+
     // A new QApplication
     Base::Console().Log("Init: Creating Gui::Application and QApplication\n");
 


matthijskooijman:
Thanks for merging!

I would suggest to consider backporting this change to 0.19. Is there something I should do to request or facilitate that?

I think this fix does not need any particular documentation, but should be mentioned in a changelog. Is there something I can do for that (the PR template links to a release notes forum thread, but that's 0.19 only)?

luzpaz:
//cc @DeepSOIC (you might be interested in this)

wwmayer:
> Is there something I should do to request or facilitate that?

I have cherry-picked and pushed the commit to the v0.19 release branch

DeepSOIC:
Nice!!, thanks for the ping.

matthijskooijman:
Cool, thanks!

I noticed there is still a small issue with using the tablet pen. Unlike using a mouse, you pretty much always move the pen slightly when you click, which I'm now seeing is often just enough to break the snap (i.e. you hold still for a short while to snap, then click and you end up putting a point in a different place than intended. Probably an issue completely unrelated to this PR, though, so maybe I'll open a forum thread to discuss it. Maybe just increasing the snap radius could already be enough to fix this, even.

adrianinsaval:
> I think this fix does not need any particular documentation, but should be mentioned in a changelog. Is there something I can do for that (the PR template links to a release notes forum thread, but that's 0.19 only)?

https://wiki.freecadweb.org/Release_notes_0.20
https://forum.freecadweb.org/viewtopic.php?f=10&t=56135

matthijskooijman:
@adrianinsaval Thanks for the links, I added this change to the wiki changelog directly. I also wanted to add it to the 0.19.x changelog, but I'm not quite sure that exists (also asked here: https://forum.freecadweb.org/viewtopic.php?f=10&t=58053&p=498907#p498907).

PR #4689, opened 2021-03-29
Arch: Fix window rendering with mixed rotating and static components

matthijskooijman:
While rendering each component of a window, the rotdata variable was
not reinitialized for each loop iteration. This means that any component
that does not define any rotation will use the rotation of the previous
component.

This is fixed by first clearing out the rotdata variable for each
component, rather than just once for the entire window.

This probably did not show up for the default windows, since they all
probably define the rotating parts last.

- [ ]  ~~In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum~~
- [ ]  ~~All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`~~ Full test was too invasive (cannot run in background), `TestArch` ran succesfully.
- [ ]  ~~Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the [FreeCAD bug tracker issue number](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) in case a particular commit solves or is related to an existing issue on the tracker. Ex: `Draft: fix typos - fixes #0004805`~~

---

Here is an example file:

[WithAndWithoutRotation.zip](https://github.com/FreeCAD/FreeCAD/files/6223890/WithAndWithoutRotation.zip)

Without this PR applied, this would render like this:

image

### DIFF
--- a/src/Mod/Arch/ArchWindow.py
+++ b/src/Mod/Arch/ArchWindow.py
@@ -1442,6 +1442,8 @@ class _ArchWindowTaskPanel:
         self.field3.setText('')
         self.field4.setText('')
         self.field5.setText('')
+        self.field6.setText(QtGui.QApplication.translate("Arch", "Get selected edge", None))
+        self.field7.setCurrentIndex(0)
         self.addp4.setChecked(False)
         self.addp5.setChecked(False)
         self.newtitle.setVisible(True)
@@ -1489,6 +1491,8 @@ class _ArchWindowTaskPanel:
             if self.obj:
                 if comp in self.obj.WindowParts:
                     ind = self.obj.WindowParts.index(comp)
+                    self.field6.setText(QtGui.QApplication.translate("Arch", "Get selected edge", None))
+                    self.field7.setCurrentIndex(0)
                     for i in range(5):
                         f = getattr(self,"field"+str(i+1))
                         t = self.obj.WindowParts[ind+i]
--- a/src/Mod/Arch/ArchWindow.py
+++ b/src/Mod/Arch/ArchWindow.py
@@ -615,11 +615,11 @@ class _Window(ArchComponent.Component):
         self.sshapes = []
         self.vshapes = []
         shapes = []
-        rotdata = None
         for i in range(int(len(obj.WindowParts)/5)):
             wires = []
             hinge = None
             omode = None
+            rotdata = None
             ssymbols = []
             vsymbols = []
             wstr = obj.WindowParts[(i*5)+2].split(',')


matthijskooijman:
I added a second commit that fixes a highly similar issue when editing window components. This commit does introduce some duplication of a translated string, but I'm not familiar enough with the codebase to tell if this is considered a problem and if so, what the preferred fix would be.

matthijskooijman:
Hm, looking at this again, it seems like the rendering behavior might actually be intentional? Looking at how default windows are defined, it seems only the frame gets a rotation value, but the glass has no rotation defined. This ensures that only one floorplan symbol arc is produced, but to also ensure that the glass actually rotates when opening>0, it must inherit its `rotdata` (but not `hinge` and `omode`) from the previous component. I guess it also makes it easier to define rotating parts if you just need to define rotation parameters once.

It does actually seem that this is [documented](https://wiki.freecadweb.org/Arch_Window) as intentional ("If you defined a hinge in this component or any other earlier in the list, ..."), though maybe it could be made a bit more prominent. Also, this requires a specific ordering (all static parts before moving parts) and there is currently no way to reorder components, so reordering might be something to consider as a feature.

So, I guess my first commit actually problematic, but I guess the second commit is still useful (since without it, I think you cannot really manage to create a rotating frame followed by a non-rotating glass, since the second component will use the same values still in the gui elements).

Anyway, probably have to see what @yorikvanhavre says :-)
This does mean

yorikvanhavre:
Indeed that was intentional, as you found out @matthijskooijman . I have no strong opinion if it needs to stay that way or not, but it seems to work well... Requiring explicit rotation info for each subcomponent might be better at some point, but then it also needs finer control as to display the arcs or not.
The whole window system is of course still a bit clumsy and hard to manipulate by the end user, but IMHO it just needs a better UI on top of it, which  "abstracts" the window components system somehow.
I'll still merge your other commit, though

yorikvanhavre:
@luzpaz I know you really like those tags, but please? :speak_no_evil:

yorikvanhavre:
Committed with cc88c21fae

matthijskooijman:
> Requiring explicit rotation info for each subcomponent might be better at some point, but then it also needs finer control as to display the arcs or not.

Yeah, agreed.

> The whole window system is of course still a bit clumsy and hard to manipulate by the end user, but IMHO it just needs a better UI on top of it, which "abstracts" the window components system somehow.

Yup. The system itself has a nice balance of simplicity and power, but the UI could indeed be improved. I also suspect the storage format could be made a bit more extensible (i.e. the current comma-separated list with all this string matching doesn't look terrible robust to me).

PR #4695, opened 2021-03-30
[GUI] Remove code for Qt < 5.9

chennes:
This PR removes any code from the main GUI module that was only included for Qt < 5.9, the current minimum required version. It only addresses explicit dependencies on the Qt version, not implied dependencies like testing for the existence of particular features (e.g. WebKit, etc.).

---

- [X]  Single module
- [X]  "Small" change
- [X]  GUI unit tests are confirmed to pass
- [X]  No tracker ticket

### DIFF
--- a/src/Gui/BitmapFactory.cpp
+++ b/src/Gui/BitmapFactory.cpp
@@ -37,9 +37,6 @@
 # include <QStyleOption>
 # include <sstream>
 #endif
-#if defined (FC_OS_WIN32) && QT_VERSION < 0x050000
-#define QTWEBKIT
-#endif
 
 #ifdef QTWEBKIT
 #include <QWebView>
@@ -177,9 +174,7 @@ QStringList BitmapFactoryInst::findIconFiles() const
         filters << QString::fromLatin1("*.%1").arg(QString::fromLatin1(*it).toLower());
 
     QStringList paths = QDir::searchPaths(QString::fromLatin1("icons"));
-#if QT_VERSION >= 0x040500
     paths.removeDuplicates();
-#endif
     for (QStringList::ConstIterator pt = paths.begin(); pt != paths.end(); ++pt) {
         QDir d(*pt);
         d.setNameFilters(filters);
@@ -188,9 +183,7 @@ QStringList BitmapFactoryInst::findIconFiles() const
             files << it->absoluteFilePath();
     }
 
-#if QT_VERSION >= 0x040500
     files.removeDuplicates();
-#endif
     return files;
 }
 
@@ -583,13 +576,8 @@ QPixmap BitmapFactoryInst::merge(const QPixmap& p1, const QPixmap& p2, Position 
 {
     // does the similar as the method above except that this method does not resize the resulting pixmap
     int x = 0, y = 0;
-#if QT_VERSION >= 0x050000
     qreal dpr1 = p1.devicePixelRatio();
     qreal dpr2 = p2.devicePixelRatio();
-#else
-    qreal dpr1 = 1;
-    qreal dpr2 = 1;
-#endif
 
     switch (pos)
     {
--- a/src/Gui/Qt4All.h
+++ b/src/Gui/Qt4All.h
@@ -161,12 +161,6 @@
 #include <QUiLoader>
 #include <QtDesigner/QFormBuilder>
 
-// QtWebKit
-// Only needed in Web module
-//#include <QWebFrame>
-//#include <QWebView>
-//#include <QWebSettings>
-
 #include "qmath.h"
 #include <QGraphicsView>
 #include <QGraphicsScene>
--- a/src/Gui/Qt4All.h
+++ b/src/Gui/Qt4All.h
@@ -49,9 +49,7 @@
 #include <qtimer.h>
 #include <qtranslator.h>
 #include <QUrl>
-#if QT_VERSION >= 0x050000
 #include <QUrlQuery>
-#endif
 #include <qvariant.h>
 #include <QWaitCondition>
 // QtGui
@@ -69,9 +67,7 @@
 #include <qcolordialog.h>
 #include <qcombobox.h>
 #include <qcursor.h>
-#if QT_VERSION >= 0x040200
 #include <QDesktopServices>
-#endif
 #include <QDesktopWidget>
 #include <QDialogButtonBox>
 #include <QDockWidget>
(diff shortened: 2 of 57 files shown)


src/Gui/Qt4All.h
```diff
@@ -168,12 +162,10 @@
 #include <QtDesigner/QFormBuilder>
 
 // QtWebKit
-#if QT_VERSION >= 0x040400
 // Only needed in Web module
 //#include <QWebFrame>
 //#include <QWebView>
 //#include <QWebSettings>
-#endif
```
hyarion, line 169: I think you can remove this whole section

src/Gui/Quarter/QuarterWidget.cpp
```diff
@@ -1053,14 +1047,6 @@ QuarterWidget::redraw(void)
...
 
-// #if QT_VERSION >= 0x050500 && QT_VERSION < 0x050600
-#if 1
-  // With Qt 5.5.x there is a major performance problem
-  this->viewport()->update();
-#else
-  this->viewport()->repaint();
-#endif
-
 #endif
```
hyarion, line 1050: Do we really need to save the `#if`-`#else`-`#endif` construct?

src/Gui/TreeView.cpp
```diff
@@ -47,10 +47,7 @@ TreeView::TreeView(QWidget* parent)
     this->setDropIndicatorShown(false);
     this->setRootIsDecorated(false);
     this->setSelectionMode(QAbstractItemView::ExtendedSelection);
-#if QT_VERSION >= 0x040200
-    // causes unexpected drop events (possibly only with Qt4.1.x)
     this->setMouseTracking(true); // needed for itemEntered() to work
-#endif
 }
```
hyarion, line 51: Should we investigate if this still is needed? (the comment suggests it might be a qt 4.1.x only issue)

src/Gui/WidgetFactory.cpp
```diff
@@ -466,24 +450,14 @@ Py::Object PythonWrapper::fromQObject(QObject* object, const char* className)
...
 #endif
-
 #if 0 // Unwrapping using sip/PyQt
     Q_UNUSED(className);
-#if QT_VERSION >= 0x050000
     return qt_wrapInstance<QObject*>(object, "QObject", "sip", "PyQt5.QtCore", "wrapinstance");
-#else
-    return qt_wrapInstance<QObject*>(object, "QObject", "sip", "PyQt4.Qt", "wrapinstance");
-#endif
 #endif
```
hyarion, line 461: I think you can remove this `#if`-`#endif` block

src/Gui/WidgetFactory.cpp
```diff
@@ -505,23 +479,15 @@ Py::Object PythonWrapper::fromQWidget(QWidget* widget, const char* className)
...
 
 #if 0 // Unwrapping using sip/PyQt
     Q_UNUSED(className);
-#if QT_VERSION >= 0x050000
     return qt_wrapInstance<QWidget*>(widget, "QWidget", "sip", "PyQt5.QtWidgets", "wrapinstance");
-#else
-    return qt_wrapInstance<QWidget*>(widget, "QWidget", "sip", "PyQt4.Qt", "wrapinstance");
-#endif
 #endif
```
hyarion, line 491: same here

hyarion (changes requested):

src/Gui/WidgetFactory.cpp
chennes: There are three instances of that code in this file -- I am operating on the assumption that it was commented out but left in place for some reason, so I'm hesitant to remove it.

src/Gui/TreeView.cpp
chennes: My read of that comment is that `this->setMouseTracking(true);` is needed, but had problems with Qt 4.1. It's currently compiled.

src/Gui/Quarter/QuarterWidget.cpp
chennes: I don't think so, I'll remove it.

src/Gui/Qt4All.h
chennes: I think you're right, I don't see any good reason to leave that in.

src/Gui/TreeView.cpp
hyarion: My bad, I misread the `#if`

src/Gui/WidgetFactory.cpp
hyarion: I undestand but afaik, freecad changed from sip+pyqt to shiboken+pyside back in around 0.15.
This makes me believe that it should be safe to remove. Maybe it's best to tag in someone who knows this stuff better?

chennes: I'm sure @wwmayer can tell us 😄

chennes: (I've added a separate commit, 056bbfb, that does this removal, in the expectation that it turns out that it's fine to remove.)

hyarion:
I think the code looks good now, but I'll leave the "request changes" status as we need some guidance with 056bbfb as stated above

src/Gui/WidgetFactory.cpp
wwmayer: It's true that we moved to PySide a few years ago but the above code should stay as a fallback solution because PySide packages can be broken or do not even exist.

src/Gui/BitmapFactory.cpp
```diff
@@ -37,9 +37,6 @@
 # include <QStyleOption>
 # include <sstream>
 #endif
-#if defined (FC_OS_WIN32) && QT_VERSION < 0x050000
-#define QTWEBKIT
-#endif
 
 #ifdef QTWEBKIT
```
wwmayer, line 41: The whole block
```c++
#ifdef QTWEBKIT
...
#endif QTWEBKIT
```
can be removed. This was a workaround with Qt4 because the QtSvg module was very limited and often failed to create a valid icon. So, we used the Webkit engine to create the icons.
With Qt5 QtSvg is much more powerful and creates valid icons for actually all plain SVG files

wwmayer:
The function MainWindow::showMainWindow() and the used timer can be completely removed, too: e7c91d68af

wwmayer:
GESTURE_MESS was set when using Qt4 under Windows 7 or later. So we can completely remove the files WinNativeGestureRecognizers.h and WinNativeGestureRecognizers.cpp and all code using the classes there: 72eceb91c6

wwmayer:
Merged.

chennes:
Thanks for all the additional info and fixes, @wwmayer -- nice to see such a large cleanup in this class!

PR #4697, opened 2021-03-31
[Main] Remove deprecated Qt < 5.9 code

chennes:
This PR removes any code from the main module that was only included for Qt < 5.9, the current minimum required version. 

---

- [X]  Single module
- [X]  Small change
- [X]  Unit tests are confirmed to pass
- [X]  No tracker ticket

### DIFF
--- a/src/Main/MainGui.cpp
+++ b/src/Main/MainGui.cpp
@@ -120,13 +120,6 @@ int main( int argc, char ** argv )
     // Make sure to setup the Qt locale system before setting LANG and LC_ALL to C.
     // which is needed to use the system locale settings.
     (void)QLocale::system();
-#if QT_VERSION < 0x050000
-    // http://www.freecadweb.org/tracker/view.php?id=399
-    // Because of setting LANG=C the Qt automagic to use the correct encoding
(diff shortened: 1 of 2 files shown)


src/Main/MainGui.cpp
```diff
@@ -189,9 +182,8 @@ int main( int argc, char ** argv )
     App::Application::Config()["SplashTextColor" ] = "#ffffff"; // white
     App::Application::Config()["SplashInfoColor" ] = "#c8c8c8"; // light grey
 
-#if (QT_VERSION >= QT_VERSION_CHECK(5, 7, 0))
     QGuiApplication::setDesktopFileName(QStringLiteral("org.freecadweb.FreeCAD.desktop"));
-#endif
+
```
hyarion, line 187: This will be three new lines in a row, maybe it could be reduced to just one?

hyarion (approved):
Looks good, just a nitpick comment :)

src/Main/MainGui.cpp
chennes: It was really only two, but I force-pushed it down to one :smile:.

PR #4701, opened 2021-03-31
[TD] Remove deprecated Qt < 5.9 code

chennes:
This PR removes any code from the TechDraw module that was only included for Qt < 5.9, the current minimum required version. 

NOTE: While working on this PR I ran across a bug in `MDIViewPage::printPdf(std::string file)`, around src/Mod/TechDraw/Gui/MDIViewPage.cpp line 811 in the original code. There is an unused local variable there -- I believe its scope is wrong. A fix for that seemed inappropriate to include in this PR, so I left a note there that should be revisited once this is merged.

---

- [X]  Single module
- [X]  Small change
- [X]  TD unit tests are confirmed to pass
- [X]  No tracker ticket


hyarion (approved):
Looks good! 👍

hyarion (approved):

wwmayer:
The problem has been added with e6032b7d533

wwmayer:
The fix: 033b5397f3

PR #4703, opened 2021-03-31
[Web] Remove deprecated Qt < 5.9 code

chennes:
This PR removes any code from the Web module that was only included for Qt < 5.9, the current minimum required version. For Web this also implies removing support for Qt WebKit, which was deprecated in Qt 5.5 in favor of Qt WebEngine. The result is that this PR is a more extensive change to the underlying module than the others. Nothing was broken in my testing, but since the unit tests are of limited use in this case, actual user tests would be appreciated!

---

- [X]  Single module
- [X]  Small change
- [X]  Unit tests are confirmed to pass
- [X]  No tracker ticket

### DIFF
--- a/src/Mod/Web/Gui/BrowserView.cpp
+++ b/src/Mod/Web/Gui/BrowserView.cpp
@@ -52,7 +52,6 @@
 #endif
 
 
-#if QT_VERSION >= 0x050700 && defined(QTWEBENGINE)
 # include <QWebEnginePage>
 # include <QWebEngineView>
 # include <QWebEngineSettings>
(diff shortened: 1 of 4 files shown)


hyarion:
> actual user tests would be appreciated!

@chennes Is there anything specific you would like to be tested?

chennes:
> @chennes Is there anything specific you would like to be tested?

Mostly just make sure that you can click on things in Start and they still work. Really, as long as it actually compiles and runs Start on various systems we're probably OK.

hyarion (approved):
Code looks good to me and I've verified that Start still works on macos 👍

src/Mod/Web/Gui/BrowserView.cpp
```diff
@@ -237,65 +222,25 @@ Py::Object BrowserViewPy::setHtml(const Py::Tuple& args)
  */
 
 WebView::WebView(QWidget *parent)
-    : QWEBVIEW(parent)
+    : QWebEngineView(parent)
 {
-    // Increase html font size for high DPI displays
```
wwmayer, line 242: Why is this whole block removed? The function setTextSizeMultiplier() is used for the QtWebEngine.

wwmayer:
On my system I have to use the nouveau driver that causes frequent crashes with Qt's webengine while it works without problems with webkit.

In https://forum.freecadweb.org/viewtopic.php?f=10&t=50289&p=457470&hilit=qtwebkit#p457705 we found a workaround. However, I am not sure whether this 100% fixes the crashes or what performance issues may arise by rendering OpenGL in software. Thus, I prefer to stay with webkit.

wwmayer:
With 51b57f9844 I have reworked the PR to leave any webkit code there.

> Why is this whole block removed? The function setTextSizeMultiplier() is used for the QtWebEngine.

Sorry, just found your comment that the function is obsolete with QtwebEngine. efba06737f

wwmayer:
So, I hope I have merged all relevant changes and didn't break anything :)

PR #4707, opened 2021-04-03
[UI] Replace SVG icon of TechDraw RedrawPage

bitacovir:
TechDraw RedrawPage icon is too similar with Std_Refresh icon and it could be difficult to identify. A design improvement was requested.

This commit replaces SVG file with new icon design for this command. Also, it makes the necessary changes on Command.cpp and TechDraw.qrc files, following the FC convention for file names.

The new SVG icon follows the FreeCAD Artwork Guidelines and was saved as Plain SVG format.

Forum Discussion: https://forum.freecadweb.org/viewtopic.php?f=34&t=57139&start=10#p492312


luzpaz:
There is definately more work to do on that .qrc file to get it uniform with the rest of the FC workbenches. Thanks bitacovir!

PR #4685, opened 2021-03-29
[PD] add feature to select pad direction via an edge

donovaly:
This was requested by users in the forum and is indeed very valuable for the usability.

also fix minor UI logic issues

Here is the new feature in action:
Do8LYsCrwV

PR #4708, opened 2021-04-04
[PartDesign] add direction to cylinder primitive

donovaly:
We already have the possibility to create skewed prisms. I recently stumbled upon that I would need this feature for cylinders too.

This PR takes the existing prism extrude direction feature and use it for cylinders too, in Part and PartDesign.

A showcase what is now possible (additive and subtractive skewed cylinders):

FreeCAD_Bn0ObFKHg2

### DIFF
--- a/src/Mod/Part/App/PrimitiveFeature.cpp
+++ b/src/Mod/Part/App/PrimitiveFeature.cpp
@@ -79,6 +79,7 @@ namespace Part {
     const App::PropertyQuantityConstraint::Constraints angleRangeU = {0.0,360.0,1.0};
     const App::PropertyQuantityConstraint::Constraints angleRangeV = {-90.0,90.0,1.0};
     const App::PropertyQuantityConstraint::Constraints quantityRange  = {0.0,FLT_MAX,0.1};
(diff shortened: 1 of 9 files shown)


wwmayer:
It makes more sense to move the code for the shearing to a common base class to avoid code duplication. Then the Prism and Cube features and later Cube or Cone features can inherit from that class, too.

Or we implement a separate Shear feature that links a part feature.

wwmayer:
84c5d22ccb35b implements the prism function as an extension to reduce code duplication.
This PR is split into two commits based on the above change: c0e6debcfa and 7185ee93f

donovaly:
Many thanks also for the much simpler prism creation.

For a skewed box, one can already use a prism. I was not yet aware of that also cones can be skewed ->I'll make a PR the next days.

As discussed privately, to finish the "skewness" work, the pending PR #4685 implements to set the pad direction using any edge (this turned out to be a real time-saver in daily work), when this is in, I will pocket give the same features as pad.

donovaly:
> For a skewed box, one can already use a prism. I was not yet aware of that also cones can be skewed ->I'll make a PR the next days.

I had a look and it seems that OCC cannot create oblique circular cones. So providing "skewed" cones is not possible.

PR #4714, opened 2021-04-05
Base: Extend domain of GetAngle method in Vector3 class to smaller length values

marioalexis84:
Since the value of Precision::Confusion is 1e-7, the GetAngle method should be able to work with a par of vectors with length of this order.

---

### DIFF
--- a/src/Base/Vector3D.cpp
+++ b/src/Base/Vector3D.cpp
@@ -422,7 +422,7 @@ _Precision Vector3<_Precision>::GetAngle (const Vector3 &rcVect) const
 
     divid = Length() * ((Vector3<_Precision>&)rcVect).Length();
 
-    if ((divid < -1e-10f) || (divid > 1e-10f)) {
+    if ((divid < -1e-14f) || (divid > 1e-14f)) {
         fNum = (*this * rcVect) / divid;
         if (fNum < -1)
             return traits_type::pi();


src/Base/Vector3D.cpp
```diff
@@ -422,7 +422,7 @@ _Precision Vector3<_Precision>::GetAngle (const Vector3 &rcVect) const
 
     divid = Length() * ((Vector3<_Precision>&)rcVect).Length();
 
-    if ((divid < -1e-10f) || (divid > 1e-10f)) {
+    if ((divid < -1e-14f) || (divid > 1e-14f)) {
```
chennes, line 425: Is there a reason this is hard-coded instead of using `Precision::Confusion*Precision::Confusion`?

marioalexis84: As I understand it, the code in Base should be independent of (but useful for) libraries like OpenCascade, Coin, etc. Therefore it would not be appropriate to use Precision::Confusion here.
Maybe we can define a suitable macro for this value. 
There is another identical case:
https://github.com/FreeCAD/FreeCAD/blob/373e3a68aa2eae60e3b91c10ecff4bc24e63f3c7/src/Base/Rotation.cpp#L488-L490

wwmayer: If you want consistency with OCCT then using Precision::Confusion() isn't the right way either. Here is the OCCT code to determine the angle of two vectors:
```c++
inline Standard_Real gp_Vec::Angle (const gp_Vec& Other) const
{
  //    Commentaires :
  //    Au dessus de 45 degres l'arccos donne la meilleur precision pour le
  //    calcul de l'angle. Sinon il vaut mieux utiliser l'arcsin.
  //    Les erreurs commises sont loin d'etre negligeables lorsque l'on est
  //    proche de zero ou de 90 degres.
  //    En 3d les valeurs angulaires sont toujours positives et comprises entre
  //    0 et Pi.
  
  gp_VectorWithNullMagnitude_Raise_if
    (coord.Modulus()       <= gp::Resolution() ||
     Other.coord.Modulus() <= gp::Resolution(), " ");
  return (gp_Dir(coord)).Angle(Other);
}
```
where Modulus() is the length and gp::Resolution() is implemented as
```c++
inline Standard_Real gp::Resolution  () 
{ 
  return RealSmall(); 
}
```
RealSmall() is defined in Standard_Real.hxx as
```
inline Standard_Real     RealSmall() 
{ return DBL_MIN; }
```

So, the right way is to use Vector3's epsilon() method.

wwmayer:
See 69aa35578e

wwmayer:
PR is obsolete now.

marioalexis84:
now it looks much better, thanks for the insight!

PR #4716, opened 2021-04-06
[PD] UI file fixes

donovaly:
the rotation angle settings of primitives had a different layout than the other parameters, this things like shown in the image below happened.

- The PR uniforms all layouts to grid and for the PD layout
- it sets for the cylinder page a minimum width
- the PD UI file is now also sorted using David's file sorter
- fix the mess with stdset="0"> I accidentally introduced with the merges from yesterday (I simply forgot to install to Qt designer the FreeCAD widgets and then designer pollutes the file with these tags for all FC-specific widgets since it doesn't know them)

FreeCAD_YL55T1OVXy

### DIFF
--- a/src/Mod/Part/Gui/DlgPrimitives.ui
+++ b/src/Mod/Part/Gui/DlgPrimitives.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>360</width>
-    <height>298</height>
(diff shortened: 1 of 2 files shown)


wwmayer:
Merged.

wwmayer:
Commit in master dff561a1b0

PR #4726, opened 2021-04-09
[OpenSCAD] Fix twist angle to not limit to 360 deg

chennes:
The linear_extrude OpenSCAD function was mistakenly limited to a twist angle of [-360°, 360°] because it was stored in a PropertyAngle. This PR modifies that to use a PropertyQuantity with an Angle unit set instead. See https://forum.freecadweb.org/viewtopic.php?p=495254

---

- [X]  Single module
- [X]  Small change
- [X]  No tracker ticket


yorikvanhavre:
Clever trick... Didn't know about this

PR #4730, opened 2021-04-11
[PD] fix primitive issues

donovaly:
- to avoid invalid results:
  * don't allow to set equal radii for cones
  * don't allow to set equal wedge parameters
  * don't allow zero quantities like e.g. for the box width

- use full internal precision because it must be possible to create an e.g. 23.5 um wide box and using the default 2 displayed digits

- take spinbox limits from App

- add missing tooltips

- remove disturbing commented out line from Workbench.cpp

- some coding style issues fixed automatically by the MSVC IDE

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePrimitive.cpp
+++ b/src/Mod/PartDesign/App/FeaturePrimitive.cpp
@@ -56,10 +56,13 @@ using namespace PartDesign;
 
 namespace PartDesign {
 
-const App::PropertyQuantityConstraint::Constraints torusRangeV = {-180.0,180.0,1.0};
-const App::PropertyQuantityConstraint::Constraints angleRangeU = {0.0,360.0,1.0};
-const App::PropertyQuantityConstraint::Constraints angleRangeV = {-90.0,90.0,1.0};
-const App::PropertyQuantityConstraint::Constraints quantityRange  = {0.0,FLT_MAX,0.1};
+const App::PropertyQuantityConstraint::Constraints torusRangeV = { -180.0, 180.0, 1.0 };
+const App::PropertyQuantityConstraint::Constraints angleRangeU = { 0.0, 360.0, 1.0 };
+const App::PropertyQuantityConstraint::Constraints angleRangeV = { -90.0, 90.0, 1.0 };
+// it turned out that OCC cannot e.g. create a box with a width of Precision::Confusion()
+// with two times Precision::Confusion() all geometric primitives can be created
+const App::PropertyQuantityConstraint::Constraints quantityRange = { 2 * Precision::Confusion(), FLT_MAX, 0.1 };
+const App::PropertyQuantityConstraint::Constraints quantityRangeZero = { 0.0, FLT_MAX, 0.1 };
 
 PROPERTY_SOURCE_WITH_EXTENSIONS(PartDesign::FeaturePrimitive, PartDesign::FeatureAddSub)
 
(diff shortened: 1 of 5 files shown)


src/Mod/PartDesign/App/FeaturePrimitive.cpp
```diff
@@ -56,10 +56,13 @@ using namespace PartDesign;
...
 
-const App::PropertyQuantityConstraint::Constraints torusRangeV = {-180.0,180.0,1.0};
-const App::PropertyQuantityConstraint::Constraints angleRangeU = {0.0,360.0,1.0};
-const App::PropertyQuantityConstraint::Constraints angleRangeV = {-90.0,90.0,1.0};
-const App::PropertyQuantityConstraint::Constraints quantityRange  = {0.0,FLT_MAX,0.1};
+const App::PropertyQuantityConstraint::Constraints torusRangeV = { -180.0, 180.0, 1.0 };
+const App::PropertyQuantityConstraint::Constraints angleRangeU = { 0.0, 360.0, 1.0 };
+const App::PropertyQuantityConstraint::Constraints angleRangeV = { -90.0, 90.0, 1.0 };
+// it turned out that OCC cannot e.g. create a box with a width of Precision::Confusion()
+// with double Precision::Confusion() all geometrixc promitives can be created
```
chennes, line 63: A couple typos, and a suggested word change (since the C++ programmer in me read the word "double" to mean something else entirely :smile: .
```suggestion
// with twice Precision::Confusion() all geometric primitives can be created
```

src/Mod/PartDesign/Gui/TaskPrimitiveParameters.cpp
```diff
@@ -630,75 +633,77 @@ void TaskBoxPrimitives::onWedgeXminChanged(double v) {
...
         QString name(QString::fromLatin1(Gui::Command::getObjectCmd(obj).c_str()));
         QString cmd;
         App::Document* doc = App::GetApplication().getActiveDocument();
         if (!doc) {
-            return;
+            return false;
         }
+        // To allow e.g. a width of 2.2 um while the users has e.g. 2 user decimals
+        // we cannot use Base::UnitsApi::getDecimals().
+        // It is also more sensible to to disallow values OCC allows, thus use OCC's limit
```
chennes, line 695: There is a typo here: a repeated "to". I also don't follow the stated logic: do you mean "disallow values OCC **dis**allows"?
```suggestion
        // It is also more sensible to disallow values OCC (dis?)allows, thus use OCC's limit
```

chennes (commented):
This compiles, runs, and seems to work as expected. I am unable to enter an edge length for a cube that is less than `confusion()`, I get the expected error messages when creating a cone with equal radii, and in my testing I did not encounter any unexpected consequences for the changes made.

My comments relate entirely to comments in the code: mostly just typos.

src/Mod/PartDesign/Gui/TaskPrimitiveParameters.cpp
donovaly: Many thanks for the review.

I fixed now the typos and wording as you suggested.

donovaly:
Many thanks for the review.

I fixed now the typos and wording as you suggested.

src/Mod/PartDesign/App/FeaturePrimitive.cpp
donovaly: thanks, fixed

chennes (approved):
Looks good, thanks!

wwmayer:
> use full internal precision because it must be possible to create an e.g. 23.5 um wide box and using the default 2 displayed digits

Increasing the number of decimals still doesn't help if you want to use a very small number like 0.01 nm (e.g. for Xmin of a wedge) because the number will be rounded to 0.00 mm. In order to allow small values and respecting the user settings for number of decimals is to use 'g' for the format string. With cd9319f1 I have added a convenience function UnitApi::toNumber() to do that.

wwmayer:
Merged as  b9b42a52f43508

PR #4747, opened 2021-04-21
AddonManager cannot download anything if working dir is on ecryptfs (aka ubuntu's encrypted home)

SniperCZE:
This pull request fixes AddonManager to work with encrypted Ubuntu home dir with ecryptfs. Ecryptfs has internal limitation of path lenght to 140 characters only, but AddonManager respects only 260 chars limitation of normal filesystems.

### DIFF
--- a/src/Mod/AddonManager/addonmanager_workers.py
+++ b/src/Mod/AddonManager/addonmanager_workers.py
@@ -600,8 +600,8 @@ class ShowWorker(QtCore.QThread):
                 name = path.split("/")[-1]
                 if name and path.startswith("http"):
                     storename = os.path.join(store, name)
-                    if len(storename) >= fscharlimit:
-                        remainChars = (fscharlimit - 1) - (len(store) + len(wbName) + 1)
+                    if len(storename) >= 260:
+                        remainChars = 259 - (len(store) + len(wbName) + 1)
                         storename = os.path.join(store, wbName+name[-remainChars:])
                     if not os.path.exists(storename):
                         try:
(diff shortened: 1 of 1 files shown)


luzpaz:
Hi @SniperCZE can you explain to the uninformed how this circumvents the char limit ?

SniperCZE:
AddonManager, when trying to download new workbench, generates filename with path to Images. When this generated path is longer than 260 chars, it is trunked to 260 chars. This behavior works perfectly on filesystems with path length limit of 260 or more. If this is used on ecryptfs (aka encrypted home on Ubuntu, with max lenght of 143 chars), open() will throw OSError exception. This patch handles OSError exception in this part of code and assume "if there is OSError, we should use shortes filename/path", so it cuts filepath to 140 chars and try to open that shorter version. It is not a perfect solution (this would be something like FS detection), but good enaugh.

luzpaz:
Cool, follow-up question: so what would a truncated path look like? Would the addon (Workbench/macro) work if the paths have now changed ?

SniperCZE:
Yep, it is fully working. Path for AddonManager includes prefix of working dir (on Ubuntu it is /home/user/.FreeCAD/AddonManager/Images) and some long string looking like workbenchNameabc158ffedd5496add549bb946... Working dir will probably never longer than 140 chars, so only that long string is truncated. I tried it on several workbenches and everything was working without any issue. When I looked into that long-name files, those are some svgs or jpegs.

yorikvanhavre:
There are spaces/tabs mismatches in your code that render it both unreadable here on github and can cause errors when running. In FreeCAD code we use strictly 4-spaces for indentation (no tabs). Could you fix the code? Thanks!!

yorikvanhavre:
Ok should be safe enough indeed. Better fix later would be to pinpoint the exception better than using an OSError (which could be issued for other causes)

PR #4763, opened 2021-04-26
Misc: remove py2 code

luzpaz:


luzpaz:
@wwmayer should i submit fixes for `src/CXX/` files as well ?

wwmayer:
> @wwmayer should i submit fixes for src/CXX/ files as well ?

No, this is not necessary.

PR #4768, opened 2021-04-29
Draft: Fixed 3 gui_*.py files that did not execute Gui.addModule("Draft") if UsePartPrimitives is True

Roy-043:
If UsePartPrimitives is True Draft.autogroup(...) is part of _cmd_list, therefore Draft module must be available. Since the Draft module imports the Part module, there is no need to load the latter separately.


yorikvanhavre:
Good catch, thanks!!

PR #4769, opened 2021-04-29
Draft: Fixed make_ellipse.py: the face argument was not used

Roy-043:
The face argument was not used and therefore "filled mode" did not work in the GUI.

### DIFF
--- a/src/Mod/Draft/draftmake/make_ellipse.py
+++ b/src/Mod/Draft/draftmake/make_ellipse.py
@@ -60,22 +60,28 @@ def make_ellipse(majradius, minradius, placement=None, face=True, support=None):
     support : 
         TODO: Describe.
     """
+
     if not App.ActiveDocument:
         App.Console.PrintError("No active document. Aborting\n")
         return
+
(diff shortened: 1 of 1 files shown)


src/Mod/Draft/draftmake/make_ellipse.py
```diff
@@ -60,22 +60,28 @@ def make_ellipse(majradius, minradius, placement=None, face=True, support=None):
...
     obj = App.ActiveDocument.addObject("Part::Part2DObjectPython", "Ellipse")
     Ellipse(obj)
+
     if minradius > majradius:
         majradius, minradius = minradius, majradius
     obj.MajorRadius = majradius
     obj.MinorRadius = minradius
     obj.Support = support
+
+    if face != None:
```
luzpaz, line 77: Probably should be changed to `if face is not None:`  
ref: https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none

Roy-043: Thanks I'll try to remember this next time.

PR #4772, opened 2021-05-02
Base: add new API Rotation::get/setEulerAngles()

realthunder:
Forum post https://forum.freecadweb.org/viewtopic.php?f=22&t=55229&start=10#p485189

There is a testing script inside the post.

### DIFF
--- a/src/Base/Rotation.cpp
+++ b/src/Base/Rotation.cpp
@@ -27,6 +27,7 @@
 # include <climits>
 #endif
 
+#include <boost/algorithm/string/predicate.hpp>
 #include "Rotation.h"
 #include "Matrix.h"
 #include "Base/Exception.h"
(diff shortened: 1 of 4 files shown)


src/Base/Rotation.cpp
```diff
@@ -712,3 +713,271 @@ bool Rotation::isNull() const
             this->quat[2] == 0.0 &&
             this->quat[3] == 0.0);
 }
+
+//=======================================================================
+// The following code is borrowed from OCCT gp/gp_Quaternion.cxx
```
chennes, line 718: That's not LGPL'ed code -- does their license allow this use?

realthunder: Really? But the license comment at the beginning of gp_Quaternion.cxx says its under LGPL. It also mentions that the implementation is based on other samples and posts in public.

chennes: You are right, I was looking at an [old version](http://git.dev.opencascade.org/gitweb/?p=occt.git;a=blob;f=src/gp/gp_Quaternion..cxx;h=cb9ab504ca14cac70c8c77bb9e0ec35e60f14482;hb=f6f03db9d08051ceb1741ba8a18adfd172d98a91), they did re-license it when it got rewritten a few years ago. Thanks for checking.

realthunder: No problem. BTW, I'll respond to your other code comment later. You know it takes time to switch to the branch to modify and compile and all. I am currently busy with other stuff.

luzpaz:
CC @KeithSloan

PR #4775, opened 2021-05-03
[Arch]: allow Rebar OffsetStart and OffsetEnd in negative also

amrit3701:
### Output with negative offsets

image
image


luzpaz:
@yorikvanhavre  backport to 0.19 ?

yorikvanhavre:
Don't think so... It's not a bugfix, it's a new feature. Unless you think it's necessary @amrit3701 ?

amrit3701:
> Don't think so... It's not a bugfix, it's a new feature. Unless you think it's necessary @amrit3701 ?

ah! yes. We can consider it as a new feature. 😁

PR #4776, opened 2021-05-04
[Path] Fixes for drilling operation

dubstar-04:
Forum Discussion: https://forum.freecadweb.org/viewtopic.php?f=15&t=58306&sid=1eea82eb33515fc35e70affd64fbbe70

Bug: https://tracker.freecadweb.org/view.php?id=4652

Changes:
- Start drilling operations from OpStartDepth + SafeHeightOffset rather than StockTop + SafeheightOffset
- Enable the use of tipoffset when using toolbits. Display an unsupported warning if using legacy tools. Toolbits FTW!
- Set the retract height relative to the operations startDepth + SafeHeightOffset.


dubstar-04:
Works as expected for all the examples I tested. Hold released.

PR #4787, opened 2021-05-12
[GUI] Add Preference Pack support

chennes:
# Summary 

This PR adds a system for storing and applying sets of user preferences called "Preference Packs". These are intended to give users the ability to make bulk modifications to their FreeCAD setup by applying one or more of these packs. The packs are designed to be distributable, and to that end this PR introduces a new type of metadata file, "package.xml". 

# Details

This PR comprises three main additions to FreeCAD:

## App::Metadata
C++ and Python support for reading and writing the new metadata format. That format is based on [REP 149](https://ros.org/reps/rep-0149.html), with minor modifications to accommodate use within FreeCAD. The format is documented at https://wiki.freecadweb.org/Package_Metadata. 

Preference Packs require this file to be present and to describe the pack, and the long-term goal is that this metadata file can be added to other types of Add-Ons, facilitating a more powerful add-on manager, and eliminating the need to compile in external workbench icons, etc.

## Gui::PreferencePack
This class manages a single preference pack, which itself is made of a number of separate files, grouped together into a folder with the same name as the preference pack (as described by its package.xml metadata file). 

A preference pack consists of the following files (all technically optional):
* pre.FCMacro -- run prior to applying the pack. If it raises an exception, application of the pack is aborted.
* Package Name.cfg -- The core of a preference pack, a well-formatted user.cfg file containing only those elements that the preference pack sets. Anything that can be in a user.cfg file is allowed.
* post.FCMacro -- run after applying the pack. If it raises an exception, the application of the pack is rolled back.
* Auxiliary files referenced by the .cfg file -- most commonly a QSS stylesheet, but in the future might include icons, etc.

Because of the sensitivity of user preferences, and the possibility that a pack will make a change the user does not like that is not discovered immediately, when applying a pack two separate backups are taken. One is used to roll back the most recent change in the event that post.FCMacro raises an exception, and the other is a set of backups managed by the PreferencePackManager (described next) and retained for one week. There is currently no user interface for rolling those back, they must be manually copied by the user.

## Gui::PreferencePackManager

This class manages a list of all available preference packs, and also provides the ability to save the user's current settings as a new pack. This facility is provided by including a set of "Preference Pack Template" files. These are needed to support saving of preferences that the user has not explicitly set, and are using their hardcoded default values -- in that case, the user.cfg file does not have any information about the current setting, which is only currently available in the code that accesses that setting. 

To overcome this obstacle, Preference Pack Templates are intended to provide a complete "user.cfg"-formatted set of preferences set to their default values. When a user indicates selects "Save as new" in the GUI, several locations are scanned for these template files (including installed add-ons), and the user is presented with a set of checkboxes to select which items they would like stored. This is mainly intended to provide "Theme" authors with an easy way of creating new preference packs with UI color information in them. To that end, a relatively complete set of templates is provided for UI color information, and only a single "demonstration" template is included in this PR with other preference information in it. In the long term it is expected that a robust set of templates will be developed and included with FreeCAD covering the most-requested Preference Pack settings.

# Future Work

A separate PR will provide an extension to the Add-on Manager to support the new package.xml data, and to add "Preference Pack" as a new type of add-on. For now, packs can be managed as a standard add-on, as long as they provide a package.xml file and follow the appropriate folder structure and naming convention. This includes manual installation into the Mod directory.

# Forums discussions

* Metadata: https://forum.freecadweb.org/viewtopic.php?f=10&t=57448
* Initial theme discussion: https://forum.freecadweb.org/viewtopic.php?f=34&t=57301
* Preference pack UI discussion: https://forum.freecadweb.org/viewtopic.php?f=8&t=58210

---

- [ ]  Your pull request is confined strictly to a single module.
- [X]  Discussed in forums
- [X]  All FreeCAD unit tests are confirmed to pass
- [X]  No tracker ticket


src/App/Metadata.cpp
```diff
@@ -0,0 +1,750 @@
...
+    if (rootTagName != "package")
+        throw Base::XMLBaseException("Malformed package.xml document: Root <package> group not found");
+
+    auto formatVersion = XMLString::parseInt(_dom->getAttribute(XUTF8Str("format").unicodeForm()));
+    switch (formatVersion) {
+    case 1:
+        parseVersion1(_dom);
+        break;
+    default:
+        throw Base::XMLBaseException("pacakge.xml format version is not supported by this version of FreeCAD");
```
0penBrain, line 94: pacakge -> package

src/App/Metadata.cpp
```diff
@@ -0,0 +1,750 @@
...
+Metadata::Metadata(const DOMNode* domNode, int format) : _dom(nullptr)
+{
+    auto element = dynamic_cast<const DOMElement*>(domNode);
+    if (element) {
+        switch (format) {
+        case 1:
+            parseVersion1(element);
+            break;
+        default:
+            throw Base::XMLBaseException("pacakge.xml format version is not supported by this version of FreeCAD");
```
0penBrain, line 111: pacakge -> package

src/App/Metadata.cpp
```diff
@@ -0,0 +1,750 @@
...
+
+Meta::Version App::Metadata::freecadmax() const
+{
+    return _freecadmax;
+}
+
+std::multimap<std::string, Metadata> Metadata::content() const
+{
+    return _content;
+}
```
0penBrain, line 203: Couldn't these functions return const values ?

src/App/Metadata.cpp
```diff
@@ -0,0 +1,750 @@
...
+            // If none of this node's children have children of their own, it is a simple element and we
+            // can handle it as a GenericMetadata object
+            auto children = element->getChildNodes();
+            bool hasGrandchildren = false;
+            for (XMLSize_t i = 0; i < children->getLength() && !hasGrandchildren; ++i)
+                if (children->item(i)->getChildNodes()->getLength() > 0)
+                    hasGrandchildren = true;
+            if (!hasGrandchildren)
+                _genericMetadata.insert(std::make_pair(tagString, Meta::GenericMetadata(element)));
+        }
```
0penBrain, line 589: Maybe a switch..case would be more readable. Also all cases that just 'emplace_back(element)' could be grouped.

src/Gui/DlgCreateNewPreferencePackImp.cpp
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+}
+
+void DlgCreateNewPreferencePackImp::on_lineEdit_textEdited(const QString& text)
+{
+    if (!text.isEmpty())
+        ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(true);
+    else
+        ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(false);
```
0penBrain, line 152: Not sure it's general, but to me is more readable 'ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(!text.isEmpty())'

src/Gui/PreferencePackTemplates/Behavior/Main_window_layout.cfg
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
+<FCParameters>
+  <FCParamGroup Name="Root">
+    <FCParamGroup Name="BaseApp">
+      <FCParamGroup Name="MainWindow">
+        <FCParamGroup Name="DockWindows">
+          <FCBool Name="Std_SelectionView" Value="0"/>
+          <FCBool Name="Std_ComboView" Value="1"/>
+          <FCBool Name="Std_ReportView" Value="0"/>
+          <FCBool Name="Std_PythonView" Value="0"/>
```
0penBrain, line 10: Worth listing 'Tree View', 'Property View' and 'DAG View' ?

src/App/Metadata.cpp
chennes: What would be the advantage?

src/App/Metadata.cpp
chennes: C++ can't switch on non-integral types -- the strings are listed in the order of their documentation, rather than in the order of their implementation strategy. I don't feel strongly about that order, but it does make sense to me.

src/Gui/DlgCreateNewPreferencePackImp.cpp
chennes: Good idea, will change.

src/Gui/PreferencePackTemplates/Behavior/Main_window_layout.cfg
chennes: Yeah, probably. Right now that file is purely a skeleton for testing, I'd love some help fleshing it it, and creating other Behavior templates.

src/App/Metadata.cpp
0penBrain: Correct. And using a constexpr would probably be another mess here. :)

src/App/Metadata.cpp
0penBrain: Hmmm... Maybe no one now we have C++11 as a minimum requirement...

src/Gui/PreferencePackTemplates/Behavior/Main_window_layout.cfg
chennes: I have added those three, but will leave this conversation open in case other additions come up.

luzpaz:
@chennes would mpetrikas's [VerticalUI](https://forum.freecadweb.org/viewtopic.php?f=34&t=58747) be a good usecase for this ?

chennes:
> @chennes would mpetrikas's [VerticalUI](https://forum.freecadweb.org/viewtopic.php?f=34&t=58747) be a good usecase for this ?

Fo some of mpetrika's suggestions, yes: the ones that are just preference-driven, or are part of the Qt Theme system. But other suggestions in that post are not things that are (currently) stored as user preferences, and do require code modifications.

src/Gui/PreferencePackTemplates/Behavior/Main_window_layout.cfg
chennes: At the suggestion of Forum member jnxd I have added Editor font information into a template file.

balrobs:
@chennes here is my test report:
https://forum.freecadweb.org/viewtopic.php?p=526323#p526323
I had some troble with Test E. Maybe I missunderstood something ;-)

balrobs:
@chennes here is my test report:
https://forum.freecadweb.org/viewtopic.php?p=526323#p526323
I had some trouble with test E ... maybe I missunderstood something ;-)

yorikvanhavre:
This looks like a mastery addition... Will test too ASAP!

chennes:
> I had some troble with Test E. Maybe I missunderstood something ;-)

Thanks again for testing: I've pushed a fix to the bug you discovered.

balrobs:
Thank you for the fix. Now everything is fine ... only one minor issue remaining? See https://forum.freecadweb.org/viewtopic.php?f=8&t=58210&start=40#p526935

Symbian9:
> "Preference Packs" ... `package.xml`

Why not just use "Preferences Preset" and store in `preferences.xml`?

chennes:
> Why not just use "Preferences Preset" and store in `preferences.xml`?

No particular reason for "Preference Pack" - we discussed it in the forum and that was the consensus. Basically the same for the xml file, though also note that that file is not specific to Preference Packs, the intention is that in the long term all mods are distributed with that file, so we stuck with the "standard" filename as used in that ROS specification.

berndhahnebach:

berndhahnebach:
FreeCAD CI-repository:
branch:
https://gitlab.com/berndhahnebach/FreeCAD/-/commits/PR_4787

pipeline:
https://gitlab.com/berndhahnebach/FreeCAD/-/pipelines/377336996

status:
success

chennes:
Thanks, @yorikvanhavre . 

1. I decided to put it on the general tab for two reasons: first, because otherwise that tab had a lot of unused space on it, and this can expand to fit, or contract as needed. And second, because when you apply a pack the window has to get closed and reopened, this way it appears to be reopening where you left off! Even though actually it just opens to the first page all the time.
2. Eventually that will be connected to the add-on manager, but I didn't want to do that until I'd actually written the new Add-on manager code to handle the metadata and preference packs. I'll hide that button for now unless the user has saved some preference packs, and add a tooltip to it.
3. Agreed. It made sense at the time, I think, but several months later I agree that it's pretty worthless. I'll look into what else could be displayed instead. Maybe a list of "tags"? The pack description will be too long, I think.

yorikvanhavre:
ok! the description could be in a tooltip maybe..

yorikvanhavre:
So far so good :sweat_smile:

PR #4791, opened 2021-05-14
[Base] Fix lost filename in err msg

heewa:
In some circumstances, FileExceptions are constructed empty, then have a filename assigned to them, but the error message in these scenarios is left as the default "unknown" one, which is missing both the exception message and the filename.

This change fixes that by re-calculating both when an except is updated with a `PyObject`. In addition, I consolidated that logic into a private method so that construction & later-updating stay consistent over time.

To verify, I tried to save a project after deleting its containing dir. Before this change:

image

After change:

image

This doesn't fix [Issue 4098](https://tracker.freecadweb.org/view.php?id=4098), but it helps with the error msg.

- [x]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [x]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [x]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  ~~All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`~~ (relying on CI to verify tests. Is that ok?)
- [x]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [x]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [x]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the [FreeCAD bug tracker issue number](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) in case a particular commit solves or is related to an existing issue on the tracker. Ex: `Draft: fix typos - fixes #0004805`

PR #4792, opened 2021-05-14
[Gui] Fix app quitting despite failing to save project

heewa:
Currently, a failure during a document save can return `true` in some cases, which among other possible problems, can cause the App to exit despite not successfully saving the document, resulting in data loss (eg [Issue 4098](https://tracker.freecadweb.org/view.php?id=4098)). This change causes the method to return `false` to properly communicate the failure, which the app-quit process already handles by aborting. 

Steps to reproduce / verify fix:

1. Create a document in its own dir & save it
2. Without closing the document or the app, delete the containing dir
3. Quit the app

**Before this change**: app asks you to save, which fails, but the app exits anyway
**After this change**: app asks you to save, which fails, but the app aborts exit, giving you the chance to "save as" somewhere else

- [ ]  ~~All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`~~ (relying on CI to verify tests. is that ok?)

### DIFF
--- a/src/Gui/Document.cpp
+++ b/src/Gui/Document.cpp
@@ -1161,6 +1161,7 @@ bool Document::save(void)
         catch (const Base::Exception& e) {
             QMessageBox::critical(getMainWindow(), QObject::tr("Saving document failed"),
                 QString::fromLatin1(e.what()));
+            return false;
         }
         return true;
     }


luzpaz:
If fix is valid, we should consider backporting as well.

chennes:
Does this have any interaction with https://github.com/FreeCAD/FreeCAD/pull/4996?

AjinkyaDahale:
> Does this have any interaction with #4996?

The changes made here are very helpful, but I don't believe necessary for reporting a failed save. AFAICT in 06e35bfa6fa7a37c3164fff1d50275540925a5ef I only use the fact that the exception is a `Base::FileException`. Without this commit only the generic "Unknown exception..." error should show, but I haven't tested it.

That said, I had included this commit in #4996 (with attribution to @heewa), so it is already merged through 820e88f95b5c6d34da09d2152fc3aa3b268e7a95.

heewa:
Closing, since these changes have been merged. Thanks, @AjinkyaDahale!

AjinkyaDahale:
> Closing, since these changes have been merged. Thanks, @AjinkyaDahale!

Sorry I think I confused this PR with #4791. This PR is still totally valid. I was unsure what `bool Document::save(void)` returns so I ignored the `BaseException` catches, but returned `false` when `FileException` happened and nothing was done.

adrianinsaval:
shouldn't this PR be open again then?

heewa:
Ok, reopening. 

As for the interaction with #4996, I don't see a description that would lead me to fully understand that PR. Regardless, this PR makes the modified function behave correctly: if there was a failure to save a document, then the `Document.save()` function should return `false`, otherwise data loss can occur. This is a dangerous bug to leave in the code, regardless of whether there are currently any bugs elsewhere which would cause that incorrect code path to be run.

chennes:
Great, thanks, that makes sense.

PR #4793, opened 2021-05-14
[TD] fix centerline crashes and bugs

donovaly:
the PR fixes the following things:
- change the way centerlines between 2 lines are calculated. The current implementation leads to many bugs and even crashes (e.g. when the 2 selected lines are parallel ones of a square). There are different forum threads from the last 1.5 years.
The new endpoint line sorting is the one described here: https://forum.freecadweb.org/viewtopic.php?f=35&start=20&t=44255&sid=989a668890f954c13ef81e4a04ee6912#p501179

- as consequence the erroneous and misleading line end flipping can go and is removed (only used internally)

- when creating a new centerline, you see it immediately

- when creating a new or editing an existing centerline and press Cancel, the creation/editing is aborted

- fix crash when the 2 selected lines are bot horizontal and one tries to create a vertical centerline

- fix crash when changing the orientation  of an existing centerline and the result would be invalid

- cleanup the code a bit

-----------

Todo for a further PR: after editing an existing centerline, the change cannot be undone. (I am lost here since no committed command will appear in the Undo. But the centerline creation command appears correctly in Undo).

### DIFF
--- a/src/Mod/TechDraw/App/Cosmetic.cpp
+++ b/src/Mod/TechDraw/App/Cosmetic.cpp
@@ -39,6 +39,7 @@
 
 #include <Base/Console.h>
 #include <Base/Exception.h>
+#include <Base/Matrix.h>
 #include <Base/Parameter.h>
 #include <Base/Reader.h>
 #include <Base/Tools.h>
@@ -977,6 +978,20 @@ std::pair<Base::Vector3d, Base::Vector3d> CenterLine::calcEndPoints(DrawViewPart
     return result;
 }
 
+bool CenterLine::Circulation(Base::Vector3d A, Base::Vector3d B, Base::Vector3d C)
+{
+    Base::Matrix4D CircMatrix(
+        A.x, A.y, 1, 0,
+        B.x, B.y, 1, 0,
+        C.x, C.y, 1, 0,
+        0, 0, 0, 1);
+
+    if (CircMatrix.determinant() > 0)
+        return true;
+    else
+        return false;
+}
+
 std::pair<Base::Vector3d, Base::Vector3d> CenterLine::calcEndPoints2Lines(DrawViewPart* partFeat,
                                                       std::vector<std::string> edgeNames, 
                                                       int mode, double ext,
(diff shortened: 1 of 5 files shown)


src/Mod/TechDraw/App/Cosmetic.cpp
```diff
@@ -1018,17 +1033,21 @@ std::pair<Base::Vector3d, Base::Vector3d> CenterLine::calcEndPoints2Lines(DrawVi
     Base::Vector3d l2p1 = edges.back()->getStartPoint();
     Base::Vector3d l2p2 = edges.back()->getEndPoint();
 
-    if (flip) {             //reverse line 2
-        Base::Vector3d temp;
-        temp = l2p1;
-        l2p1 = l2p2;
-        l2p2 = temp;
+    // the centerline is drawn using the midpoints of the two lines that connect l1p1-l2p1 and l1p2-l2p2
+    // However, we don't know which point of the secont line should be l2p1 to get a geometrically
```
luzpaz, line 1037: secont should be **second**

src/Mod/TechDraw/Gui/TaskCenterLine.ui
```diff
@@ -371,20 +371,6 @@
      </item>
     </layout>
    </item>
-   <item>
-    <widget class="QCheckBox" name="cbFlip">
-     <property name="enabled">
-      <bool>false</bool>
-     </property>
-     <property name="toolTip">
-      <string>Flips endpoints of selected lines for centerline creation,
```
luzpaz, line 380: This tooltip text is deleted. Is it substituted somewhere else then?

donovaly: > This tooltip text is deleted.

Yes, since the checkbox that it described was deleted too.

src/Mod/TechDraw/App/Cosmetic.cpp
donovaly: thanks, fixed

PR #4795, opened 2021-05-17
Crowdin: remove superfluous whitespace from translation 

luzpaz:
ref: https://crowdin.com/translate/freecad/548/en-en?filter=basic&value=2#6587168

### DIFF
--- a/src/Mod/Draft/draftmake/make_arc_3points.py
+++ b/src/Mod/Draft/draftmake/make_arc_3points.py
@@ -156,7 +156,7 @@ def make_arc_3points(points, placement=None, face=False,
     try:
         _edge = Part.Arc(p1, p2, p3)
     except Part.OCCError as error:
-        _err(translate("draft","Cannot generate shape: ") + "{}".format(error))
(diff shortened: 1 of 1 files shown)


src/Mod/Draft/draftmake/make_arc_3points.py
```diff
@@ -156,7 +156,7 @@ def make_arc_3points(points, placement=None, face=False,
     try:
         _edge = Part.Arc(p1, p2, p3)
     except Part.OCCError as error:
-        _err(translate("draft","Cannot generate shape: ") + "{}".format(error))
+        _err(translate("draft","Cannot generate shape:") + " " + "{}".format(error))
```
chennes, line 159: An alternative would be to incorporate the space into the second string:
```suggestion
        _err(translate("draft","Cannot generate shape:") + " {}".format(error))
```

luzpaz: @chennes funny you should mention this, i waffled on doing just this but if you look at the rest of the file the former approach is used. So instead of changing them all, I kept the trend.

PR #4797, opened 2021-05-18
[Path] GUI scripting support and other improvements

Russ4262:
This PR makes some useful changes to the existing operations to ease scripting with GUI support for the Path WB.  This improvement is done by relocating the creation of the `PathOpGui.CommandResources` object from being an object inside the `PathOpGui.SetupOperation()` function to each GUI operation module, and passing this independent resource object to the `PathOpGui.SetupOperation()` function in place of the original arguments that were used to create the same resource object.  With this relocation, the ability to disable the initial Task Panel GUI interaction is added.  This means users can script with GUI/view-provider support without having to interact with the Task Panel for new operations or jobs.  The good news is that the normal/existing ability to edit operations and jobs with the Task Panel is still available once the new operation or job is created.

This PR adds the ability to pass the parent job for a new operation as an optional argument in the `Create()` function within each operation module.  This allows the new operation to be pre-assigned to the provided parent job instead of needing to ask the user during GUI sessions.

This PR refactors the PathDressupBoundary module only by isolating the boundary dressup algorithm into an independent class so it is accessible without the requirement of creating a new dressup object.

This PR contains a fix for failed toolbit lookups from fixed-path toolbits used in shared files, that allows Path to locate the equivalent bit on the local machine.  There is bug 4647 in the tracker system that this might affect in a positive way.  I have not investigated whether the fix in this PR has a positive affect on that particular bug.

This PR also adds an improvement to the Job Simulator usage.  It simply adds an auto-loading action of a Job in the simulator if the user has one selected, otherwise the simulator feature is untouched.  This improvement is really just an ease-of-use improvement.  If the user has a job selected prior to running the job simulator, then that job will be auto-loaded in the simulator.
 
 
---

### DIFF
--- a/src/Mod/Path/PathScripts/PathAdaptive.py
+++ b/src/Mod/Path/PathScripts/PathAdaptive.py
@@ -737,6 +737,16 @@ def SetupProperties():
     return setup
 
 
+def SetupProperties():
+    setup = ["Side", "OperationType", "Tolerance", "StepOver",
+             "LiftDistance", "KeepToolDownRatio", "StockToLeave",
+             "ForceInsideOut", "FinishingProfile", "Stopped",
+             "StopProcessing", "UseHelixArcs", "AdaptiveInputState",
+             "AdaptiveOutputState", "HelixAngle", "HelixConeAngle",
+             "HelixDiameterLimit", "UseOutline"]
+    return setup
+
+
 def Create(name, obj=None, parentJob=None):
     '''Create(name) ... Creates and returns a Adaptive operation.'''
     if obj is None:
--- a/src/Mod/Path/PathScripts/PathAdaptive.py
+++ b/src/Mod/Path/PathScripts/PathAdaptive.py
@@ -737,9 +737,9 @@ def SetupProperties():
     return setup
 
 
-def Create(name, obj=None):
+def Create(name, obj=None, parentJob=None):
     '''Create(name) ... Creates and returns a Adaptive operation.'''
     if obj is None:
         obj = FreeCAD.ActiveDocument.addObject("Path::FeaturePython", name)
-    obj.Proxy = PathAdaptive(obj, name)
+    obj.Proxy = PathAdaptive(obj, name, parentJob)
     return obj
--- a/src/Mod/Path/PathScripts/PathCustom.py
+++ b/src/Mod/Path/PathScripts/PathCustom.py
@@ -70,9 +70,9 @@ def SetupProperties():
     return setup
 
 
-def Create(name, obj=None):
+def Create(name, obj=None, parentJob=None):
     '''Create(name) ... Creates and returns a Custom operation.'''
     if obj is None:
         obj = FreeCAD.ActiveDocument.addObject("Path::FeaturePython", name)
-    proxy = ObjectCustom(obj, name)
+    obj.Proxy = ObjectCustom(obj, name, parentJob)
     return obj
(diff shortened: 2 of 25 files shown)


Russ4262:
I added a commit to improve Path Simulator by way of auto-loading the pre-selected Job in the simulator.

sliptonic:
Can the tiling feature proposal be split out as a separate (later) PR?

Russ4262:
> Can the tiling feature proposal be split out as a separate (later) PR?

For clarity, you are asking that this current PR be cleansed of the Tile Job Utility portion, and limited to the scripting and other improvements?  And then create another Tile Job Utility PR independently, that would be dependent on this PR being merged, yes?

If this is the correct interpretation of your question, I believe the answer to be yes.  I believe I can repurpose this current PR into a scripting-only improvements PR, and then create a secondary follow-up PR that contains the Tile Job Utility.
 
Awaiting confirmation and/or correction.
Thanks.

Russ4262:
On a separate note, if @dubstar-04 is able to test the included commit [https://github.com/FreeCAD/FreeCAD/pull/4797/commits/1761bc1f59366b5e240ff085ccfcdcd892ed598b](https://github.com/FreeCAD/FreeCAD/pull/4797/commits/1761bc1f59366b5e240ff085ccfcdcd892ed598b) as it pertains to bug 4647 in the tracker system, I would be appreciative.  
During development of this new tiling feature, I also ran into a problem with the toolbit system's resolution/location of a standard bit used within a shared file from another user.  My machine failed to locate the same bit on my machine because my bit file was not at the location specified in the ToolBit object in the shared file.
 
The fix proposed at the commit in this PR, at the link above, fixed the issue on my machine, but does need review by others - another reason for this PR being a Draft.  I will try to figure out how to reproduce bug 4647 and then test with this commit in discussion here.

sliptonic:
> 
> For clarity, you are asking that this current PR be cleansed of the Tile Job Utility portion, and limited to the scripting and other improvements? And then create another Tile Job Utility PR independently, that would be dependent on this PR being merged, yes?
> 
Yes, exactly.  Big PRs are problematic.  Not just because there's a lot of code to review but also because acceptance is all or nothing.  I haven't played with the feature yet but it might be that we want to move forward with the scripting improvements and give the new feature more time for evaluation.

Russ4262:
@sliptonic 
This PR has now been reduced to scripting improvements and fixes only.  The proposed Tile Job Utility feature has been removed.  
.
I will see if I can reproduce the relative/absolute toolbit path bug posted in the bug tracker and see if the toolbit path related fix committed in this PR has any influence on it.

Russ4262:
There is a small change/improvement I would like to make to this PR regarding the current scripting improvements.  What I have in mind should make these changes a little easier for users to implement in scripts.  I am doing some more research on the changes and testing to see which is better for interaction and simplicity.

Russ4262:
Okay.  Small change made.  Added a simple `CreateOp()` method for simpler scripting with GUI/viewprovider support. 
.
Might need to take out commit 89137c2.  Doing some research on toolbit bug 4697 in tracker.  I don't think the fix here improves that bug.

Russ4262:
I removed the toolbit-related fix commit from this PR.

Russ4262:
@sliptonic 
This PR is ready.  I removed the gui-related scripting improvements and their commits.  I learned how to successfully add a ViewProvider to a non-gui operation.  It would be nice to add some code to make this easier, but that can be later down the road.  I also fixed a bug in the ff3adac commit in this PR.
Thanks

PR #4800, opened 2021-05-18
App: Abort save when unable to finish

heewa:
While applying a `BackupPolicy` during a `Document` save, there are two error cases where the project's file will not contain the latest changes, but only a warning was being logged, while the save was considered successful. Not only is that technically incorrect, but it can lead to data loss, such as on app quit.

This fix addresses that by throwing an exception, which will abort the save, inform the user, and give them a chance to take alternate action.

### Steps to Reproduce (Linux)

1. Create a test directory and save a project in there (empty is fine)
2. Make some change in the project, but don't yet save, and leave it open in FreeCAD
3. Set the sticky bit of the dir, and change owner (eg `chmod +t <dir> && sudo chown root <dir>`)
4. Lock the main project file by changing its ownership, while retaining read permissions (`chmod 644 <project.FCStd> && sudo chown root <project.FCStd>`)
5. Try to quit FreeCAD without saving, and choose to save the project when prompted

**Results before this change:** The save is considered successful and FreeCAD exits without actually updating the main project file or indicating to the user that anything went wrong.

**Results after this change:** Currently, the same! However, with the addition of PR #4792, an error dialog is shown to the user, and the app aborts exiting, giving the user a chance to take alternate action.

This is just one way of triggering this issue, but really [any number of errors](https://man7.org/linux/man-pages/man2/renameat2.2.html#ERRORS) can happen to trigger it.

Forum Post: [How to handle failures during Document Save](https://forum.freecadweb.org/viewtopic.php?f=10&t=58645)

- [ ]  ~~All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`~~ (relying on CI for testing)

### DIFF
--- a/src/App/Document.cpp
+++ b/src/App/Document.cpp
@@ -2398,8 +2398,8 @@ private:
 
         Base::FileInfo tmp(sourcename);
         if (tmp.renameFile(targetname.c_str()) == false) {
-            Base::Console().Warning("Cannot rename file from '%s' to '%s'\n",
-                                    sourcename.c_str(), targetname.c_str());
+            throw Base::FileException(
+                "Cannot rename tmp save file to project file", targetname);
         }
     }
     void applyTimeStamp(const std::string& sourcename, const std::string& targetname) {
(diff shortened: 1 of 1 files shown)


yorikvanhavre:
This looks correct to me. Leaving for @wwmayer to press the button

chennes:
I just tested this after rebasing to the latest master (particularly in light of PR #4996), and as far as I can tell it is still working correctly, though I find it's hard to know for sure I'm triggering a test of this bug now that saving is so much more reliable. Merging this still seems to be to be the safest path forward, if somehow this does get triggered just writing an error to the console isn't really adequate.

yorikvanhavre:
Okay let's merge then. In any case it's better than just printing a warning and going forward..

PR #4801, opened 2021-05-18
Draft: Fixed make_label to accept list of strings for custom_text

Roy-043:
The make_label function should accept a list of strings for custom_text. Compare the make_text function. The new code was mostly taken from there.

### DIFF
--- a/src/Mod/Draft/draftmake/make_label.py
+++ b/src/Mod/Draft/draftmake/make_label.py
@@ -282,12 +282,13 @@ def make_label(target_point=App.Vector(0, 0, 0),
     if not custom_text:
         custom_text = "Label"
     try:
-        utils.type_check([(custom_text, (str, list))])
+        utils.type_check([(custom_text, (str, list))], name=_name)
     except TypeError:
         _err(translate("draft","Wrong input: must be a list of strings or a single string."))
         return None
 
-    if not all(isinstance(element, str) for element in custom_text):
+    if (type(custom_text) is list
+            and not all(isinstance(element, str) for element in custom_text)):
         _err(translate("draft","Wrong input: must be a list of strings or a single string."))
         return None
 
--- a/src/Mod/Draft/draftmake/make_label.py
+++ b/src/Mod/Draft/draftmake/make_label.py
@@ -119,8 +119,10 @@ def make_label(target_point=App.Vector(0, 0, 0),
         - `'Area'` will show the `Area` of the target object's `Shape`,
           or of the indicated `'FaceN'` in `target`.
 
-    custom_text: str, optional
+    custom_text: str, or list of str, optional
         It defaults to `'Label'`.
+        If it is a list, each element in the list represents a new text line.
+
         It is the text that will be displayed by the label when
         `label_type` is `'Custom'`.
 
@@ -280,10 +282,17 @@ def make_label(target_point=App.Vector(0, 0, 0),
     if not custom_text:
         custom_text = "Label"
     try:
-        utils.type_check([(custom_text, str)], name=_name)
+        utils.type_check([(custom_text, (str, list))])
     except TypeError:
-        _err(translate("draft","Wrong input: must be a string."))
+        _err(translate("draft","Wrong input: must be a list of strings or a single string."))
         return None
+
+    if not all(isinstance(element, str) for element in custom_text):
+        _err(translate("draft","Wrong input: must be a list of strings or a single string."))
+        return None
+
+    if isinstance(custom_text, str):
+        custom_text = [custom_text]
 
     _msg("direction: {}".format(direction))
     if not direction:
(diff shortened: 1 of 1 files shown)


src/Mod/Draft/draftmake/make_label.py
```diff
@@ -280,11 +282,18 @@ def make_label(target_point=App.Vector(0, 0, 0),
...
-        _err(translate("draft","Wrong input: must be a string."))
+        _err(translate("draft","Wrong input: must be a list of strings or a single string."))
+        return None
+
+    if not all(isinstance(element, str) for element in custom_text):
+        _err(translate("draft","Wrong input: must be a list of strings or a single string."))
         return None
 
+    if isinstance(custom_text, str):
+        custom_text = [custom_text]
```
0penBrain, line 295: According https://github.com/FreeCAD/FreeCAD/blob/9b2ebec35dca5497b1595d97682768a196ad348b/src/Mod/Draft/draftobjects/label.py#L170 this isn't really needed as CustomText property can also accept a single string. However I didn't perform any run check about that.

Roy-043: You are completely correct. Thanks for pointing this out. Note that these lines were copied from make_text where they can also be omitted.

Roy-043: I am going to try to update this PR...

Roy-043:
First time I have updated a PR. I hope what I have done is OK.

0penBrain:
@Roy-043 at least what you did with the PR works. Optimally, you should squash the commits together or force push.

src/Mod/Draft/draftmake/make_label.py
```diff
@@ -280,9 +282,13 @@ def make_label(target_point=App.Vector(0, 0, 0),
...
         custom_text = "Label"
     try:
-        utils.type_check([(custom_text, str)], name=_name)
+        utils.type_check([(custom_text, (str, list))])
     except TypeError:
-        _err(translate("draft","Wrong input: must be a string."))
+        _err(translate("draft","Wrong input: must be a list of strings or a single string."))
+        return None
+
+    if not all(isinstance(element, str) for element in custom_text):
```
matthijskooijman, line 290: This seems wrong: If `custom_text` is not a list, but a string, I think this will end up looping over each element of the string to check they are strings (since elements of a string are single-character strings themselves, the check works, but this wastes processing).

matthijskooijman: I just realized that your second commit actually introduces this problem, with just the first commit `custom_text` is always a list at this point.

src/Mod/Draft/draftmake/make_label.py
```diff
@@ -280,9 +282,13 @@ def make_label(target_point=App.Vector(0, 0, 0),
     if not custom_text:
         custom_text = "Label"
     try:
-        utils.type_check([(custom_text, str)], name=_name)
+        utils.type_check([(custom_text, (str, list))])
```
matthijskooijman, line 285: Why does this remove the `name` argument? That doesn't seem related to the main change, and it looks like passing a name here is still relevant? If it is somehow not possible with the main change, then IMHO at least the commit message should document why it must be removed. If this is just because the `make_text` function doesn't pass the argument, then it should probably be kept here (and maybe added to `make_text` even).

src/Mod/Draft/draftmake/make_label.py
Roy-043: > with just the first commit `custom_text` is always a list at this point.

This cannot be correct. The 2nd commit removed code on line 294 and 295.

src/Mod/Draft/draftmake/make_label.py
Roy-043: My bad. I will fix that.

src/Mod/Draft/draftmake/make_label.py
matthijskooijman: Yes, so "with just the first commit" (in other words, if you omit the second commit), then those lines still exist, and those lines ensure that `custom_text` is also a list. So removing those lines means that `custom_text` might be a string instead of a list, so the because of that you might end up iterating a string instead of a list.

Right?

Roy-043:
@matthijskooijman I think I have now fixed the issues you have mentioned. Thanks for checking.

src/Mod/Draft/draftmake/make_label.py
Roy-043: No the code that ensured custom_text was a list came after line 290 and had no impact there.

src/Mod/Draft/draftmake/make_label.py
Roy-043: FYI the mentioned name argument is used for messages in the Report view:
`typecheck[make_label]: '1' is not (<class 'str'>, <class 'list'>)`

src/Mod/Draft/draftmake/make_label.py
matthijskooijman: Oh, you're right, I hadn't looked properly. You're right, it was broken in the first commit already :-) Seems fixed now!

yorikvanhavre:
Thanks everybody for the hard work and checking!

PR #4805, opened 2021-05-21
StartWB: Add missing tooltip for 'Show tips' preference dialog

luzpaz:


luzpaz:
Eventually we'll need to run the .ui cleaner that sorts all these option in a top-down tab-friendly sequence. Noted in https://tracker.freecadweb.org/view.php?id=4664#c15671

PR #4808, opened 2021-05-22
Draft: Do not move children of App::Link objects

matthijskooijman:
This caused the original object to be moved when using the Draft move
tool to move a link object, when the original object has MoveWithHost
set (e.g. a Window). There was already a similar exception for clones,
so it makes sense to extend this for links too.

Note that there seem to be more problems with the "MoveWithHost"
mechanism and fixing them might completely refactor this code, but until
then, this is a simple and targeted fix that at least makes moving
links to windows work as expected.

See https://forum.freecadweb.org/viewtopic.php?f=23&t=57223 for
discussion of this bug, its fix and the additional problems mentioned.


matthijskooijman:
If merged, I think this would be a candidate to backport to 0.19.

PR #4811, opened 2021-05-22
Draft: Edit improvements to context menu and code organization

carlopav:
Reference to post by @matthijskooijman: https://forum.freecadweb.org/viewtopic.php?f=23&t=58643&start=10#p505619

---

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_edit.py
+++ b/src/Mod/Draft/draftguitools/gui_edit.py
@@ -843,6 +843,20 @@ class Edit(gui_base_original.Modifier):
                 obj_gui_tools.restore_object_style(obj, self.objs_formats[obj.Name])
 
 
+    def get_specific_object_info(self, obj, pos):
+        """Return info of a specific object at a given position.
+        """
+        selobjs = Gui.ActiveDocument.ActiveView.getObjectsInfo((pos[0],pos[1]))
+        if not selobjs:
+            return
+        for info in selobjs:
+            if not info:
+                continue
+            if obj.Name == info["Object"] and "x" in info:
+                # prefer "real" 3D location over working-plane-driven one if possible
+                pt = App.Vector(info["x"], info["y"], info["z"])
+                return info, pt
+
     def get_selected_obj_at_position(self, pos):
         """Return object at given position.
 
--- a/src/Mod/Draft/draftguitools/gui_edit.py
+++ b/src/Mod/Draft/draftguitools/gui_edit.py
@@ -703,8 +703,12 @@ class Edit(gui_base_original.Modifier):
             return
 
         for (label, callback) in actions:
+            def wrapper(callback=callback):
+                callback()
+                self.resetTrackers(obj)
+
             action = self.tracker_menu.addAction(label)
-            action.setData(callback)
+            action.setData(wrapper)
 
         self.tracker_menu.popup(Gui.getMainWindow().cursor().pos())
 
--- a/src/Mod/Draft/draftguitools/gui_edit.py
+++ b/src/Mod/Draft/draftguitools/gui_edit.py
@@ -457,7 +457,16 @@ class Edit(gui_base_original.Modifier):
             ):#left click
             if not event.wasAltDown():
                 if self.editing is None:
-                    self.startEditing(event)
+
+                    pos = event.getPosition()
+                    node = self.getEditNode(pos)
+                    node_idx = self.getEditNodeIndex(node)
+                    if node_idx is None:
+                        return
+                    doc = App.getDocument(str(node.documentName.getValue()))
+                    obj = doc.getObject(str(node.objectName.getValue()))
+
+                    self.startEditing(obj, node_idx)
                 else:
                     self.endEditing(self.obj, self.editing)
             elif event.wasAltDown():  # left click with ctrl down
@@ -486,32 +495,25 @@ class Edit(gui_base_original.Modifier):
                     self.overNode.setColor(COLORS["default"])
                     self.overNode = None
 
-    def startEditing(self, event):
+    def startEditing(self, obj, node_idx):
         """Start editing selected EditNode."""
-        pos = event.getPosition()
-        node = self.getEditNode(pos)
-        ep = self.getEditNodeIndex(node)
-        if ep is None:
+        self.obj = obj # this is still needed to handle preview
+        if obj is None:
             return
 
-        doc = App.getDocument(str(node.documentName.getValue()))
-        self.obj = doc.getObject(str(node.objectName.getValue()))
-        if self.obj is None:
-            return
-
-        App.Console.PrintMessage(self.obj.Name
+        App.Console.PrintMessage(obj.Name
                                  + ": editing node number "
-                                 + str(ep) + "\n")
+                                 + str(node_idx) + "\n")
 
         self.ui.lineUi()
         self.ui.isRelative.show()
-        self.editing = ep
-        self.trackers[self.obj.Name][self.editing].off()
+        self.editing = node_idx
+        self.trackers[obj.Name][node_idx].off()
 
         self.finalizeGhost()
-        self.initGhost(self.obj)
+        self.initGhost(obj)
 
-        self.node.append(self.trackers[self.obj.Name][self.editing].get())
+        self.node.append(self.trackers[obj.Name][node_idx].get())
         Gui.Snapper.setSelectMode(False)
         self.hideTrackers()
 
--- a/src/Mod/Draft/draftguitools/gui_edit.py
+++ b/src/Mod/Draft/draftguitools/gui_edit.py
@@ -672,109 +672,6 @@ class Edit(gui_base_original.Modifier):
         except Exception:
             return
 
-    # -------------------------------------------------------------------------
-    # EDIT OBJECT TOOLS : Add/Delete Vertexes
-    # -------------------------------------------------------------------------
-
-    def addPoint(self, event):
-        """Add point to obj and reset trackers.
-        """
-        pos = event.getPosition()
-        # self.setSelectState(obj, True)
-        selobjs = Gui.ActiveDocument.ActiveView.getObjectsInfo((pos[0],pos[1]))
-        if not selobjs:
-            return
-        for info in selobjs:
-            if not info:
-                return
-            for o in self.edited_objects:
-                if o.Name != info["Object"]:
-                    continue
-                obj = o
-                break
-            if utils.get_type(obj) == "Wire" and 'Edge' in info["Component"]:
-                pt = App.Vector(info["x"], info["y"], info["z"])
-                self.addPointToWire(obj, pt, int(info["Component"][4:]))
-            elif utils.get_type(obj) in ["BSpline", "BezCurve"]: #to fix double vertex created
-                # pt = self.point
-                if "x" in info:# prefer "real" 3D location over working-plane-driven one if possible
-                    pt = App.Vector(info["x"], info["y"], info["z"])
-                else:
-                    continue
-                self.addPointToCurve(pt, obj, info)
-        obj.recompute()
-        self.resetTrackers(obj)
-        return
-
-    def addPointToWire(self, obj, newPoint, edgeIndex):
-        newPoints = []
-        if hasattr(obj, "ChamferSize") and hasattr(obj, "FilletRadius"):
-            if obj.ChamferSize > 0 and obj.FilletRadius > 0:
-                edgeIndex = (edgeIndex + 3) / 4
-            elif obj.ChamferSize > 0 or obj.FilletRadius > 0:
-                edgeIndex = (edgeIndex + 1) / 2
-
-        for index, point in enumerate(obj.Points):
-            if index == edgeIndex:
-                newPoints.append(self.localize_vector(obj, newPoint))
-            newPoints.append(point)
-        if obj.Closed and edgeIndex == len(obj.Points):
-            # last segment when object is closed
-            newPoints.append(self.localize_vector(obj, newPoint))
-        obj.Points = newPoints
-
-    def addPointToCurve(self, point, obj, info=None):
-        import Part
-        if  utils.get_type(obj) not in ["BSpline", "BezCurve"]:
-            return
-        pts = obj.Points
-        if utils.get_type(obj) == "BezCurve":
-            if not info['Component'].startswith('Edge'):
-                return  # clicked control point
-            edgeindex = int(info['Component'].lstrip('Edge')) - 1
-            wire = obj.Shape.Wires[0]
-            bz = wire.Edges[edgeindex].Curve
-            param = bz.parameter(point)
-            seg1 = wire.Edges[edgeindex].copy().Curve
-            seg2 = wire.Edges[edgeindex].copy().Curve
-            seg1.segment(seg1.FirstParameter, param)
-            seg2.segment(param, seg2.LastParameter)
-            if edgeindex == len(wire.Edges):
-                # we hit the last segment, we need to fix the degree
-                degree=wire.Edges[0].Curve.Degree
-                seg1.increase(degree)
-                seg2.increase(degree)
-            edges = wire.Edges[0:edgeindex] + [Part.Edge(seg1),Part.Edge(seg2)] \
-                + wire.Edges[edgeindex + 1:]
-            pts = edges[0].Curve.getPoles()[0:1]
-            for edge in edges:
-                pts.extend(edge.Curve.getPoles()[1:])
-            if obj.Closed:
-                pts.pop()
-            c = obj.Continuity
-            # assume we have a tangent continuity for an arbitrarily split
-            # segment, unless it's linear
-            cont = 1 if (obj.Degree >= 2) else 0
-            obj.Continuity = c[0:edgeindex] + [cont] + c[edgeindex:]
-        else:
-            if (utils.get_type(obj) in ["BSpline"]):
-                if (obj.Closed == True):
-                    curve = obj.Shape.Edges[0].Curve
-                else:
-                    curve = obj.Shape.Curve
-            uNewPoint = curve.parameter(point)
-            uPoints = []
-            for p in obj.Points:
-                uPoints.append(curve.parameter(p))
-            for i in range(len(uPoints) - 1):
-                if ( uNewPoint > uPoints[i] ) and ( uNewPoint < uPoints[i+1] ):
-                    pts.insert(i + 1, self.localize_vector(obj, point))
-                    break
-            # DNC: fix: add points to last segment if curve is closed
-            if obj.Closed and (uNewPoint > uPoints[-1]):
-                pts.append(self.localize_vector(obj, point))
-        obj.Points = pts
-
     # ------------------------------------------------------------------------
     # DRAFT EDIT Context menu
     # ------------------------------------------------------------------------
@@ -833,7 +730,8 @@ class Edit(gui_base_original.Modifier):
             actions = obj_gui_tools.evaluate_context_menu_action(self, obj, idx, action_label)
 
         elif action_label == "add point":
-            self.addPoint(self.event)
+            if obj and obj_gui_tools:
+                obj_gui_tools.add_point(self.event)
 
         elif action_label == "invert arc":
             pos = self.event.getPosition()
--- a/src/Mod/Draft/draftguitools/gui_edit.py
+++ b/src/Mod/Draft/draftguitools/gui_edit.py
@@ -678,7 +678,6 @@ class Edit(gui_base_original.Modifier):
 
     def display_tracker_menu(self, event):
         self.tracker_menu = QtGui.QMenu()
-        self.event = event
         actions = None
 
         if self.overNode:
@@ -689,22 +688,23 @@ class Edit(gui_base_original.Modifier):
 
             obj_gui_tools = self.get_obj_gui_tools(obj)
             if obj_gui_tools:
-                actions = obj_gui_tools.get_edit_point_context_menu(obj, ep)
+                actions = obj_gui_tools.get_edit_point_context_menu(self, obj, ep)
 
         else:
             # try if user is over an edited object
-            pos = self.event.getPosition()
+            pos = event.getPosition()
             obj = self.get_selected_obj_at_position(pos)
-            if utils.get_type(obj) in ["Line", "Wire", "BSpline", "BezCurve"]:
-                actions = ["add point"]
-            elif utils.get_type(obj) in ["Circle"] and obj.FirstAngle != obj.LastAngle:
-                actions = ["invert arc"]
+
+            obj_gui_tools = self.get_obj_gui_tools(obj)
+            if obj_gui_tools:
+                actions = obj_gui_tools.get_edit_obj_context_menu(self, obj, pos)
 
         if actions is None:
             return
 
-        for a in actions:
-            self.tracker_menu.addAction(a)
+        for (label, callback) in actions:
+            action = self.tracker_menu.addAction(label)
+            action.setData(callback)
 
         self.tracker_menu.popup(Gui.getMainWindow().cursor().pos())
 
@@ -713,32 +713,9 @@ class Edit(gui_base_original.Modifier):
                                self.evaluate_menu_action)
 
 
-    def evaluate_menu_action(self, labelname):
-        action_label = str(labelname.text())
-
-        doc = None
-        obj = None
-        idx = None
-
-        if self.overNode:
-            doc = self.overNode.get_doc_name()
-            obj = App.getDocument(doc).getObject(self.overNode.get_obj_name())
-            idx = self.overNode.get_subelement_index()
-
-        obj_gui_tools = self.get_obj_gui_tools(obj)
-        if obj and obj_gui_tools:
-            actions = obj_gui_tools.evaluate_context_menu_action(self, obj, idx, action_label)
-
-        elif action_label == "add point":
-            if obj and obj_gui_tools:
-                obj_gui_tools.add_point(self.event)
-
-        elif action_label == "invert arc":
-            pos = self.event.getPosition()
-            obj = self.get_selected_obj_at_position(pos)
-            obj_gui_tools.arcInvert(obj)
-
-        del self.event
+    def evaluate_menu_action(self, action):
+        callback = action.data()
+        callback()
 
 
     # -------------------------------------------------------------------------
--- a/src/Mod/Draft/draftguitools/gui_edit_arch_objects.py
+++ b/src/Mod/Draft/draftguitools/gui_edit_arch_objects.py
@@ -73,12 +73,6 @@ class ArchWallGuiTools(GuiTools):
             if vz.Length > 0:
                 obj.Height = vz.Length
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
-
 
 class ArchWindowGuiTools(GuiTools):
 
@@ -112,12 +106,6 @@ class ArchWindowGuiTools(GuiTools):
         for obj in obj.Hosts:
             obj.recompute()
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
-
 
 class ArchStructureGuiTools(GuiTools):
 
@@ -137,12 +125,6 @@ class ArchStructureGuiTools(GuiTools):
         nodes = obj.Nodes
         nodes[node_idx] = v
         obj.Nodes = nodes
-
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
 
     def get_object_style(self, obj):
         return (obj.ViewObject.DisplayMode,
@@ -177,12 +159,6 @@ class ArchSpaceGuiTools(GuiTools):
         if node_idx == 0:
             obj.ViewObject.TextPosition = v
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
-
 
 class ArchPanelCutGuiTools(GuiTools):
 
@@ -202,12 +178,6 @@ class ArchPanelCutGuiTools(GuiTools):
         if node_idx == 0:
             obj.TagPosition = v
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
-
 
 class ArchPanelSheetGuiTools(GuiTools):
 
@@ -226,12 +196,5 @@ class ArchPanelSheetGuiTools(GuiTools):
             obj.TagPosition = v
         else:
             obj.Group[node_idx-1].Placement.Base = v
-
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
-
 
 ## @}
--- a/src/Mod/Draft/draftguitools/gui_edit_base_object.py
+++ b/src/Mod/Draft/draftguitools/gui_edit_base_object.py
@@ -61,14 +61,33 @@ class GuiTools:
         """
         pass
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        """ Return a list of Draft_Edit context menu actions.
+    def get_edit_point_context_menu(self, edit_command, obj, node_idx):
+        """ Get the context menu associated to edit points (user is over an editpoint)
+
+        Return a list of tuples containig menu labels and associated functions:
+            return [
+                ("action label", lambda: self.handle_action_label(edit_command, obj, node_idx)),
+            ]
+
+        Parameters:
+        edit_command: running Draft_Edit command
+        obj: the edited object
+        node_idx: number of the edited node
         """
         pass
-    
 
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        """ Do something when a Draft_Edit context menu action is triggered over a node.
+    def get_edit_obj_context_menu(self, edit_command, obj, position):
+        """ Get the context menu associated to edited object (user is over the object)
+
+        Return a list of tuples containig menu labels and associated functions:
+            return [
+                ("action label", lambda: self.handle_action_label(edit_command, obj, position)),
+            ]
+
+        Parameters:
+        edit_command: running Draft_Edit command
+        obj: the edited object
+        position: position of the cursor on the screen (x, y)
         """
         pass
 
--- a/src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
+++ b/src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
@@ -176,12 +176,6 @@ class DraftRectangleGuiTools(GuiTools):
         elif node_idx == 2:
             obj.Height = DraftVecUtils.project(v, App.Vector(0,1,0)).Length
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
-
 
 class DraftCircleGuiTools(GuiTools):
 
@@ -414,12 +408,6 @@ class DraftEllipseGuiTools(GuiTools):
             else:
                 obj.MinorRadius = obj.MajorRadius
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
-
 
 class DraftPolygonGuiTools(GuiTools):
 
@@ -443,12 +431,6 @@ class DraftPolygonGuiTools(GuiTools):
         elif node_idx == 1:
             obj.Radius = v.Length
         obj.recompute()
-
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
 
 
 class DraftDimensionGuiTools(GuiTools):
@@ -474,12 +456,6 @@ class DraftDimensionGuiTools(GuiTools):
             obj.Dimline = v
         elif node_idx == 3:
             obj.ViewObject.TextPosition = v
-
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
 
 
 class DraftBezCurveGuiTools(GuiTools):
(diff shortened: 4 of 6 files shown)


carlopav:
@matthijskooijman thanks for the code example :)
how does it looks like to you?

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
```diff
@@ -123,7 +132,43 @@ def update_preview_object(self, edit_command, obj, node_idx, v):
             if obj.Closed:
                 pointList.append(pointList[0])
             edit_command.ghost.updateFromPointlist(pointList)
-        
+
+    def add_point(self, edit_command, obj, pos):
+        """Add point to obj.
+        """
+        # self.setSelectState(obj, True)
+        selobjs = Gui.ActiveDocument.ActiveView.getObjectsInfo((pos[0],pos[1]))
```
matthijskooijman, line 145: Should this be done here? It might make more sense to just use `obj` here? I realize that this is just the original code, but given that `obj` is now passed, it makes this "double work" stand out more.

I do think there will be a change in behaviour if you use `obj`, though. From the looks of it, the old intended behaviour is to add the point to *all* edited objects at the current location, not just the one `obj`. However, it seems that that behavior is currently bugged (if you add have intersecting lines, edit one of them, add a point at the intersection, you'll get two points in the edited line, not one. Also, if you edit a wire that has two segments very close together and add a point between them, it seems the result is even more messed up). Also, the current behavior does not really make sense to me: The "add point" command has been added to the menu based on `obj`, not *all* objects under the cursor, so it would make sense to only modify `obj` here (and in fact, now that add point handling is split among different draft object types, it would even be hard to add a point to e.g. both a wire and a spline in the same action).

Looking more closely at your code, you actually already change this behavior by checking `if obj.Name != info["Object"]:`. This means that you could really just leave out the for loop and just act on `obj`, which would result in the same behavior as your current code, AFAICS.

Leaving out the call to `getObjectsInfo` and the loop, also removes most duplicate code between the `add_points` methods that you introduced (that I was going to make a comment about, but that no longer seems necessary).

Finally, I would suggest you make these changes to the original `addPoint` method in a separate commit, by passing the `obj` to that `addPoint` and dropping the loop there already. This cleanly separates the behavior change in a small and easy to review commit, and the bigger generalization commit can then just move around code without changing any behavior.

src/Mod/Draft/draftguitools/gui_edit_arch_objects.py
```diff
@@ -73,12 +73,6 @@ def update_object_from_edit_points(self, obj, node_idx, v, alt_edit_mode=0):
             if vz.Length > 0:
                 obj.Height = vz.Length
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
-        pass
```
matthijskooijman, line 80: I would put the removals of all these empty functions in a separate commit, since I think they are orthogonal to generalization changes (I think they can be done either before or after the main commit, iow, the main commit does not require removing these, and removing these doesn't require the main commit, so best to separate into separate commits).

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
```diff
@@ -714,4 +796,64 @@ def smoothBezPoint(self, obj, point, style='Symmetric'):
...
+            return
+        for info in selobjs:
+            if not info:
+                return
+            for o in edit_command.edited_objects:
+                if o.Name != info["Object"]:
+                    continue
+                obj = o
+                break
+            if utils.get_type(obj) == "BezCurve": #to fix double vertex created
```
matthijskooijman, line 822: Is this check still needed? Can DraftBezCurveGuiTools edit any other types?

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
matthijskooijman: Reading the code a bit more closely, I think that the call to `getObjectsInfo` might actually still be needed, since i *think* it does not only figure out which object to act on, but also figures out which part of the object is nearest and what the exact coordinates are (i.e. `pos` might be right next to the line, while `info[x/y/z]` is always *on* the line.

Given that, I think that it would still be good to reduce the duplicate code by creating something like `getSpecificObjectInfo(obj, (x,y))` that essentially does what your code does now: Call `getObjectsInfo` and pick out only the results that point to `obj` (maybe even just *one* result which is closest, I think you rarely want to add multiple points when line segments are close together). Maybe it could also already create a Vector out of `info[x/y/z]` since that seems to be used everywhere as well. Adding this new function (and the associated behavior change) could still be done first in `addPoint` in the separate commit I mentioned.

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
```diff
@@ -176,10 +260,7 @@ def update_object_from_edit_points(self, obj, node_idx, v, alt_edit_mode=0):
         elif node_idx == 2:
             obj.Height = DraftVecUtils.project(v, App.Vector(0,1,0)).Length
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        pass
-
-    def evaluate_context_menu_action(self, edit_command, obj, node_idx, action):
+    def get_edit_obj_context_menu(self, edit_command, obj, position):
         pass
```
matthijskooijman, line 272: Why leave this empty function?

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
```diff
@@ -149,6 +194,45 @@ def update_preview_object(self, edit_command, obj, node_idx, v):
...
+                return
+            if obj.Name != info["Object"]:
+                return
+            if "x" in info:# prefer "real" 3D location over working-plane-driven one if possible
+                point = App.Vector(info["x"], info["y"], info["z"])
+            else:
+                continue
+
+            pts = obj.Points
+            if (utils.get_type(obj) in ["BSpline"]):
```
matthijskooijman, line 223: Is this check needed? Can this class ever edit anything else?

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
```diff
@@ -123,7 +132,43 @@ def update_preview_object(self, edit_command, obj, node_idx, v):
...
+        # self.setSelectState(obj, True)
+        selobjs = Gui.ActiveDocument.ActiveView.getObjectsInfo((pos[0],pos[1]))
+        if not selobjs:
+            return
+        for info in selobjs:
+            if not info:
+                return
+            if obj.Name != info["Object"]:
+                return
+            if utils.get_type(obj) == "Wire" and 'Edge' in info["Component"]:
```
matthijskooijman, line 153: Is this check still needed? Can instance of this class (not subclasses that override this method) ever edit anything else?

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
```diff
@@ -103,15 +104,23 @@ def update_object_from_edit_points(self, obj, node_idx, v, alt_edit_mode=0):
...
-            self.delete_point(obj, node_idx)
-            edit_command.resetTrackers(obj)
+    def get_edit_obj_context_menu(self, edit_command, obj, position):
+        return {
+            "add point": lambda: self.handle_add_point(edit_command, obj, position),
+        }
+
+    def handle_delete_point(self, edit_command, obj, node_idx):
+        self.delete_point(obj, node_idx)
+        edit_command.resetTrackers(obj)
```
matthijskooijman, line 120: It seems that *all* handle_something methods you defined call `edit_command.resetTrackers(obj)`. If this is really something that should always happen after running a command, maybe it can be called in edit_gui.py before calling the action callback (that probably means you'll have to create a wrapper around the callback, since evaluate_menu_action no longer has access to the object). Then maybe passing `edit_command` to `get_edit_point_context_menu` is also no longer necessary?

Or can you imagine a situation where calling `resetTrackers` after an action is not needed (or even problematic)?

src/Mod/Draft/draftguitools/gui_edit_base_object.py
```diff
@@ -61,17 +61,36 @@ def update_object_from_edit_points(self, obj, node_idx, v, alt_edit_mode=0):
...
         pass
 
-    def get_edit_point_context_menu(self, obj, node_idx):
-        """ Return a list of Draft_Edit context menu actions.
+    def get_edit_point_context_menu(self, edit_command, obj, node_idx):
+        """ Get the context menu associated to edit points (user is over an editpoint)
+
+        Return a dictionary of menu labels and associated functions:
+            return {
+                "add point": lambda: self.handle_add_point(edit_command, obj, node_idx),
```
matthijskooijman, line 69: Maybe it is better to use a list of tuples here, instead of a dictionary after all. Because:
 - A list is ordered and (until CPython 3.6, dictionaries are not)
 - A list is probably more lightweight
 - We only need to iterate over the actions, never look up an action by its label, so there is no need for a dict.
 
 
 This would look like e.g.

```python
return [
    ("add point", lambda: self.handle_add_point(edit_command, obj, node_idx)),
]
```

matthijskooijman (commented):
Thanks for picking this up! I really like where this is going.

Overall, the code looks great as-is. I left a bunch of inline comments on where I think things could be made a bit better.

I also left some comments about splitting this PR into more smaller commits, which would greatly help the reviewability of it. One additional commit that you could maybe split off is a commit where the contents of `addPoints` is spread around the various `GuiTools` base classes, but without all the menu generalization yet. Instead, as an interim approach, you could just have `evaluate_menu_action` call `obj_gui_tools.add_point()` instead of `self.addPoint()`. By itself, that's not very pretty, but as a temporary situations to be improved in a following commit, it would be fine. Combined with the other commit-splitting suggestions I made, I think that the main "generalize actions" commit would then be significantly smaller, making it a lot easier to see what's going on.

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
carlopav: Yes, you got it perfectly. I was thinking that we can add another argument in get_edit_obj_context_menu(self, edit_command, obj, position) and pass the function the real vector of the point in the 3d scene, so this stuff can just be ignored by the object code and handled completely in the Draft_Edit command. I'll give it a try!

src/Mod/Draft/draftguitools/gui_edit_arch_objects.py
carlopav: I agree that separated commits would have been better, but can I keep the suggestion for the next contribution? :-)

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
carlopav: thx, deleted!

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
carlopav: I was not sure about the check 'Edge' in info["Component"], but i tried and it works, so, perfect to remove it!

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
carlopav: I was thinking about that. 
1. No I can't imagine any scenario where `resetTrackers `is not called after editing, but I didn't know even what a wrapper is :-(
2. I think passing the edit_command to `get_edit_point_context_menu `is desirable because it enables you to do that: https://github.com/yorikvanhavre/BIM_Workbench/blob/db0017f00e6e538d421551e36b06f6e9e07f4a62/archobjects/wall.py#L1138 : you can use a menu action to start an alternative editing mode. That is done also with arcs and circles.

src/Mod/Draft/draftguitools/gui_edit_base_object.py
carlopav: Done!

src/Mod/Draft/draftguitools/gui_edit.py
```diff
@@ -362,7 +355,6 @@ def finish(self, closed=False):
             self.deformat_objects_after_editing(self.edited_objects)
         
         super(Edit, self).finish()
-        App.DraftWorkingPlane.restore()
```
carlopav, line 365: I did it again: i had to put this change in a separated commit...

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
matthijskooijman: Right, I was mostly talking about the Wire check, but I guess that a Wire can never produce anything else than edges, right?

carlopav: I think so. maybe it can produce an invalid shape (even if i never experienced that). but if so, the object is not displayed, so no way to be able to retrieve it from the scenegraph....
Uhhh, I got it. If your wire have fillets or chamfers, the editpoints are not coincident with `Draft_Wire` vertexes, so the user can `mouseover` a vertex and display context menu. In this usecase `info` does not contain edges, but vertexes! I'm reverting this change. thank you!

carlopav: Well, the correct answer was: no. A Draft Wire produces Vertexes and Edges (I do not think that in OCC you can have Edges without Vertexes). But in the Draft Edit usecase you can't mouseover a Vertex and call the `get_edit_obj_context_menu()`, because the event will be forwarded to `get_edit_point_context_menu()`. This unless you have the Vertexes not in sync with the edit points like when the wire is filleted or chamfered.

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
matthijskooijman: I'm not sure I understand what other argument you mean? Isn't the `pos` already the real vector? Or is that the 2D vector and you mean to do the `getObjectsInfo` in `display_tracker_menu` and pass the 3D vector result, i.e. from `info[x/y/z]`? If the latter, then I think you'd also need to pass `info`, since IIRC some functions use more from info than just the vector.

src/Mod/Draft/draftguitools/gui_edit_arch_objects.py
matthijskooijman: Well, ultimately, that's not up to me, but I would think that splitting/restructuring the commits to improve reviewability would be helpful to reduce load on maintainers and make it easier to merge this. Especially now that you're adding commits that amend to original thing, it becomes more difficult to review this commit-by-commit, leaving the only option of reviewing the complete PR as one, which I would not be happy about as a maintainer.

If you're intimidated by this restructuring because git can be a challenge to work with (I'm quite proficient with it and love its power, but it's certainly not easy if you're less experienced) but you a're open to investing the extra time needed to improve the commit structure, I'll be happy to share some tips on how I'd approach this using some git commands to reduce the amount of work needed.

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
matthijskooijman: > but I don't know even what a wrapper is :-(

A wrapper function is a simple function that replaces another function, and calls the original function and adds something to it. That probably doesn't really help, so let me say it in code:

```python
        for (label, callback) in actions:
            action = self.tracker_menu.addAction(label)
            def wrapper(callback=callback):
                callback()
                self.resetTrackers(obj)
            action.setData(wrapper)
```

This defines a new `wrapper` function, that calls the original callback and then calls resetTrackers. Note that this creates a new function for every loop iteration, so each of these wrapper functions captures potentially different callback, self and obj values so things work as expected. Note I did something weird with the `callback` value there: Since functions like these capture a variable, not its value at the time of function definition, and loop variables like `label` and `callback` are a single variable shared between all loop iterations, capturing `callback` directly ends up with all wrappers calling the last callback (i.e. they all capture the same variable and by the time the wrappers are called, that variable will contain the last of the callbacks). This is commonly solved by using a default value in the function definition, since such a default value uses the *value* of the (in this case) `callback` variable, not capturing the variable.
Maybe I'm only making things more confusing, though...

One other thing I realized, when you posted the link to the BIM workbench, is that we can't really *change* the behavior completely as you did now, since there are third-party gui tools that would need to be updated separately. So I guess we'll need to keep compatibility with the old way of having a separate `get_edit_point_context_menu` and ` evaluate_context_menu_action`. This is not too hard, though, since you can just create callback functions that just call ` evaluate_context_menu_action` and then the rest of the code can use those. Maybe something like:

```python
          for item in actions:
              # Compatibility with the old way of specifying actoins
              if isinstance(item, str):
                  label = item
  
                  def wrapper(label=label):
                      obj_gui_tools. evaluate_context_menu_action(self, obj, ep, label)
              else:
                  (label, callback) = item
  
                  def wrapper(callback=callback):
                      callback()
                      self.resetTrackers(obj)
  
              action = self.tracker_menu.addAction(label)
              action.setData(wrapper)

```

If you add a compatibility mode, this gives even another option for more commit separation, because then you can modify display_tracker_menu first in one commit, and then modfiy all GuiTools subclasses in a next commit.

> I think passing the edit_command to get_edit_point_context_menu is desirable because it enables you to do that: https://github.com/yorikvanhavre/BIM_Workbench/blob/db0017f00e6e538d421551e36b06f6e9e07f4a62/archobjects/wall.py#L1138 : you can use a menu action to start an alternative editing mode. That is done also with arcs and circles.

Ah, indeed. So passing `edit_command` makes sense, even when it's not needed for the `resetTrackers`.

carlopav: I think i'll be reading it again tomorrow, but it is less or more clear, thanks for the deep explanation.
That code is really elegant and I'm really in favour of keeping the resetTracker call inside main Draft_Edit module.
From the commit history point of view this is for sure good, but I don't think we have to care about backward compatibility in this case: I'm 99% sure that i'm the only one that implemented the support for other objects outside Draft WB. And those BIM Tools are marked as experimental anyway.

src/Mod/Draft/draftguitools/gui_edit_arch_objects.py
carlopav: ok, let's do it (i feel a bit guilty cause you are loosing a lot of time). I'm using GitKraken because the command line interface is quite scaring to me as a Windows user...

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
matthijskooijman: > but I don't think we have to care about backward compatibility in this case: 

Ok, but that does mean that the experimental tools in the BIM module are going to break when this is merged, and then break in older FC versions when BIM is updated. But that might be acceptable, especially if it's only these experimental modules, of course.

carlopav: > BIM module are going to break when this is merged
I'm going to fix it as soon as it's merged, since i can directly merge my commits in BIM wb master :)
It would be a problem for older versions, but we are talking about an experimental feature marked as unstable for production. (also if I can assure it's super usable)

src/Mod/Draft/draftguitools/gui_edit_arch_objects.py
matthijskooijman: > (i feel a bit guilty cause you are loosing a lot of time)

Don't worry, I can guard my own time quite well, and this would be an investment in FreeCAD's future :-p

I'll provide a more in-depth reply in the main conversion thread of this PR, since it's not longer about this particular change.

In the meanwhile, maybe you could resolve the conversations that you have fixed already? Seems only the PR author and people with repo write access can do that :-)

carlopav: Ok! I suggest a dedicated post in a differente section like Open discussion or Developers forum to have a little bit more visibility: Draft/Arch/Bim are not very watched...

matthijskooijman: Oh, I actually meant in the main thread here on github, but maybe it would be interesting to also point to this conversation on the forum, good point.

matthijskooijman:
> > > I agree that separated commits would have been better, but can I keep the suggestion for the next contribution? :-)

> > Well, ultimately, that's not up to me, but I would think that splitting/restructuring the commits to improve reviewability would be helpful to reduce load on maintainers and make it easier to merge this. Especially now that you're adding commits that amend to original thing, it becomes more difficult to review this commit-by-commit, leaving the only option of reviewing the complete PR as one, which I would not be happy about as a maintainer.

> > If you're intimidated by this restructuring because git can be a challenge to work with (I'm quite proficient with it and love its power, but it's certainly not easy if you're less experienced) but you a're open to investing the extra time needed to improve the commit structure, I'll be happy to share some tips on how I'd approach this using some git commands to reduce the amount of work needed.

> ok, let's do it (i feel a bit guilty cause you are loosing a lot of time). I'm using GitKraken because the command line interface is quite scaring to me as a Windows user...

Ok, here goes. I can't really comment on gitkraken (I always use the commandline since there it is, to me, clearest what happens under the hood and you can always use all features, rather than the subset exposed by the GUI), but I'll try to also explain the concept so you can find the same thing in gitkraken.

First off: I think it would be good to first further update the code to look like you want it to look (e.g. generalizing the getObjectInfo stuff), since you'd otherwise need to make those changes twice later.

*Intended commit history*
Then, to plan what we want the commit history to look at eventually, I think something like below. Note that I'm just talking about what the resulting commits should look like, not yet about how to get there from the current history. So the starting point of the below list is the master commit you started from.

 1. Remove all the empty `pass` functions
 2. Remove the event parameter from `startEditing` (you added this commit last, but since `evaluate_context_menu_action` already has an `obj` and `node_idx` parameter, I think you can do this change earlier, preventing the intermediate breakage by removing references to `edit_command.event` before removing the variable itself).
 3. Remove the ` App.DraftWorkingPlane.restore()` line (I do not really understand this change, but you mentioned it should be its own commit?)
 4. Introduce a `getSpecificObjectInfo` function and update the original `addPoint` function to use it (or whatever other approach you choose to extract the object-type-agnostic code from `addPoint` so it does not have to be duplicated).
 5. Split out `addPoint` to multiple `add_point` methods in the `GuiTools` subclasses (and replacing the call to `addPoint` in `evaluate_menu_action` with a call to `obj_gui_tools.add_point`).
 6. Rewriting `evaluate_menu_action` and `display_tracker_menu` to use the new callback structure, and rewrite all GuiTools subclasses to also adapt (you could split this into two commits for clarity, but then the first commit would break things, which is usually not a good idea, so I'd keep this as one commit, unless you want to add the backward compatibility wrapper).
 7. Add the `reverse wire` command

*Basic approach*
Given that most of these changes are pretty independent, the basic workflow I would choose here is to:
 - Reset you branch back to the original master commit you started working from. The default "mixed" flavor of the `git reset` command resets your branch and the index/staging area but not the working copy, so this would put you in a position where you would have been if you had done all the coding in your checkout, but not committed any of it yet. On the commandline, this would be `git reset origin/master` (assuming that you `origin/master` still points to the commit you branched off from, if not you might need to rebase first, or replace `origin/master` with the SHA of the branch-off point).
 - Now, you can use interactive staging to select a subset of all your changes and create a new commit for that. You can do this for each commit to be created in turn. On the commandline, you would use `git add -p` for this, but I suspect that in kraken you can do a similar thing in e.g. a diff view where you select lines of the diff to be staged. On the commandline you can even edit the to-be-staged diff in a text editor (useful in case you need to commit one change in a line but not another, or in adjacent lines), but if Kraken doesn't support this, and you'd need it, you can always just revert one of the changes in the file manually, commit that, ^Z in your editor and commit the other change.

Note that especially the `git add -p` / interactive staging is IMHO a very powerful tool, which allows me to have a much improved workflow where I can (in a lot of cases) just focus on coding and making things work first, and then still be able to split the changes in proper commits at a later stage (sometimes at the end, somethings I already commit things halfway when I'm confident that a particular change is somewhat complete and will be included in the final history).

*Additional work for commit 4 and 5*
This approach will not work for all commits above. In particular, number 4 and 5 need to make changes to code in `gui_edit.py` that is already removed in the current PR (so with the above approach, that code will be gone from your working copy already). For this, I would suggest first making a backup copy of `gui_edit.py` and then use `git checkout` to throw away all uncommitted changes from `gui_edit.py` (or alternatively, only throw away a selection of changes, `git checkout` has an interactive `-p` option too). After this, you should again have the original `addPoint` and `evaluate_menu_action` code, so you can make changes to those and commit that (this is the part where you'll need to do some manual code editing, and will be writing/copying back some code that will be thrown away again in the next commit, but that's ok). After you created these two commits, you can restore your `gui_edit.py` backup again, and start interactively committing the rest of the commits.

*Verifying commits work*
Note that you might want to verify in between that your separated changes actually work. One easy way for that is to first commit them (even when you're not entirely sure that you got everything, you can always amend the commit later) and then use `git stash` to stash away all uncommitted changes, so you can test the code as you committed it. Then unstash and continue with the next commit.

*Interactive rebase and fixup commits*
One additional thing I would want to point out, in case you haven't used it before, is the interactive rebase in git. Though rebase is intended to be used to "rewind" a number of commits you made and "replay" them onto another starting point (e.g. a more recent version of the master branch), you can also rebase on top of an earlier commit in your existing history (instead of some newer master version), so you end up rewinding to that earlier commit, and then simply replaying the same commits on top of the same earlier base commit again (i.e. nothing changes). But if you then use rebase in interactive mode, you can select which commits should be replayed in which order. This allows you to easily reorder your commits or remove some of them.

Even more, you can tell git to squash some commits together in the process, which means it becomes very easy to fix mistakes later: If you already committed something, but spot a mistake in it, you just make a new commit with just the fix, and then use an interactive rebase to move that "fixup" commit backwards in history to be right after the original commit, and then ask git to squash them together into one commit, as if the mistake never happened.

Git even automates a part of this process using `git commit --fixup`, which marks such "fixup" commits in such a a way that the commit message indicates what previous commit they are fixing, so an interactive rebase (with `--autosquash`) can automatically reorder and squash them. I use this workflow often where I put fixes for earlier mistakes into such fixup commits, but do not do the interactive rebase right away, but wait until some later point (so I can rebase/squash a few of them at the same time to save time, or so I can push out the fixup commits to a PR so reviewers that already saw the mistake can see the fixup more clearly before I make both disappear again). Since  git remembers for me which fixup is meant to fix which mistake, I can easily postpone the actual rebasing/squashing since I do not have to remember :-) Of course if multiple commits touch the same code and you reverse their order (i.e. a fixup commit that fixes a mistake in code that was modified again after the mistake commit), such a rebase is likely to cause conflicts, so be prepared to fix some conflicts along the way (but knowing how to fix conflicts is useful in any case, of course).

*Force pushing*
One word of caution, though: All this rebase/fixup (and also the reset + recommit stuff) rewrites history, which is typically not appropriate once commits have been merged to master, or some other branch that is intended to have some stability. If you rewrite history on a branch that you already pushed somewhere, you'll have to force push to overwrite the previous history, which means that pulling from such a branch might cause problems (since as far as git is concerned, the original and rewritten commit histories look like separately developed commits which happen to touch all the same code, leading to conflicts). Force pushing is not acceptable for master usually (though this is a matter of convention, git doesn't care), but for pull request branches, this is usually ok (people pulling from PR branches should pull carefully, though again, project-specific convention might differ).

*Commit messages*
One additional advantage of splitting code into different commits, is that you can use each commit's message to more specifically motivate the changes in that commit. In this particular case, some of the changes may seem meaningless when the commit is viewed in isolation (i.e. commit 4 would split off some code into its own function which is then called only once), but makes sense in the whole. What I usually do for such commits is explicitly note this in the commit message (e.g. something like "This commit prepares for splitting out the addPoint code into multiple versions in a future commit, reducing the amount of code to duplicate when that happens").

*Final words*
I hope this will add some tools to your toolbox, and allow you to restructure this PR in a easier-to-review manner. If things are still unclear or you're stuck somewhere, just shout and I'll try to help a bit more :-)

On another note: given you're still working on this PR, maybe you should mark it as a draft for now. Then once everything is cleaned up and we're happy, it can be marked as done, maybe marking a more clear point for the maintainers to have a look at the PR.

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
carlopav: I meant to pass the 3D vector result, but it's true we would also need to pass `info`, so better to leave the `pos[x/y]` at least for the moment.

carlopav: > Given that, I think that it would still be good to reduce the duplicate code by creating something like getSpecificObjectInfo(obj, (x,y)) that essentially does what your code does now: Call getObjectsInfo and pick out only the results that point to obj

Reading this again perhaps you are right; 
for reference this is a result of `getObjectInfo`:   `{'x': 123.58872985839844, 'y': -6.468395233154297, 'z': 0.0, 'Document': 'Unnamed', 'Object': 'Wire', 'Component': 'Edge3'}`

the `getSpecificObjectInfo(obj, (x,y))` method could return:
- or the entire obj info;
- or the clicked subobject and the 3D vector result (I guess this would cover all the usecases I can imagine, because we already know the object and the callback mechanism is bound to the view, so I guess no chance to be activated on another document).

src/Mod/Draft/draftguitools/gui_edit_arch_objects.py
carlopav: Wow, it was worth the attempt only to know the reset function :-) .
I kept using GitKraken, also if I will definitely try the command line git in the future because I stepped in several workarounds to make it, expecially dragging an dropping files in explorer to edit intermediate commits... 
It would be really nice to have it in the FC wiki IMHO.

yorikvanhavre:
As a general note, GIT mastery is never a formal requirement here... We all try to learn and make things better, PR after PR.. In that regard, @matthijskooijman 's comments are precious (I learned a couple of things too), it's time more than well spent to help us making things better ;) But I don't think we would refuse a PR or anything because of not optimal GIT formatting (unless things are very, very bad of course :D )

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
```diff
@@ -117,15 +117,12 @@ def get_edit_obj_context_menu(self, edit_command, obj, position):
 
     def handle_delete_point(self, edit_command, obj, node_idx):
         self.delete_point(obj, node_idx)
-        edit_command.resetTrackers(obj)
```
matthijskooijman: Note that all these `handle_foo` methods are now just calling a single method, so I would suggest removing them entirely and just calling that method (`delete_point` here) directly from the lambda. e.g.:

```py
        return [
                  ("delete point", lambda: self.delete_point(obj, node_idx)),
        ]
```

matthijskooijman (commented):
@carlopav Nice, restructured commits look good!

I left some small remarks on the code inline.

As for commit structure: I saw you added the getSpecificObjectInfo and resetTracker-wrapper as two additional commits at the end, rather than folding them back into the history. This is of course fine (as Yorik also emphasized, no need for commit structure to be perfect), but let me at least suggest some advantages of reordering those changes:
 - If you introduce getSpecificObjectInfo in `addPoint` before splitting it out, then that commit becomes a lot smaller (only one place to change) and easier to review and understand the behavior changes it makes (only working on one `obj`, rather than all `self.selected_objects` and/or all hovered objects as before). Similarly, the `add_point` split commit becomes smaller too.
 - If you make those changes to `addPoint` first, then the `add_point` split commit really just moves around code, without changing any behavior, which is again easier to review than a commit that moves around a lot of code and *also* makes some subtle changes in behavior.
 - If you introduce the `resetTracker`-wrapper right away in the "restructure" commit, then you can leave out the `resetTracker` calls from the `hande_foo` methods, and even more, leave out the `handle_foo` methods entirely, making that commit again smaller and simpler. Note that I just realized you could maybe even split this further: first move all the `resetTracker` calls into a single call in `evaluate_menu_action`, and then do the "rewriting" commit after that, which can then be a little shorter.

Responding to @yorikvanhavre:
> As a general note, GIT mastery is never a formal requirement here...

Thanks for confirming that. I do have a tendency to sometimes go a bit overboard with git history perfectionism (both as a contributor and a maintainer for other projects) sometimes, but as a contributor, I think it is important to do it properly because it helps making reviews easier, which allows more people to participate in the review process and takes load off the maintainers, allowing more changes to be merged. Looking at this project, I have the feeling that maintainer time and energy is indeed a more limiting factor than contributor time and energy, so IMHO it would be worth it for me, as a contributor, to spend, say, an hour extra on my commit history, if that shaves of 10 minutes of maintainer review time.

From the maintainer point of view, there is another argument to want a clean git history: It makes reading the history easier too. When figuring out how a piece of code works, the git history is often an invaluable tool for me, to see when and why a surprising bit of code is introduced. A small commit with a isolated change, ideally with a clear message that explains why a change is needed and/or links to a bug report comes a long way to understanding the origins and evolution of a bit of code and often helps to figure out which of many ways to fix a bug is the most appropriate.

Having said all that, there is of course a balance between a which for a clean history, and being open and having a low bar for new contributors. Git by itself can be tricky enough to scare away potential contributors, even without the fancy history editing stuff, especially when the contributors are not experience programmers, which seems to be often the case due to the specific domain that Freecad is in. So it's probably good to not be too picky about git structure, but here's my attempt to maybe raise the git-level of contributions a bit without raising the bar too much :-p

src/Mod/Draft/draftguitools/gui_edit.py
```diff
@@ -833,7 +729,8 @@ def evaluate_menu_action(self, labelname):
             actions = obj_gui_tools.evaluate_context_menu_action(self, obj, idx, action_label)
 
         elif action_label == "add point":
-            self.addPoint(self.event)
+            if obj and obj_gui_tools:
+                obj_gui_tools.addPoint(self.event)
```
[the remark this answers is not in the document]
carlopav: HELP! Sorry, I do not get how to fix this: should I `checkout `the `Draft: Edit, Split out addPoint in multiple methods in the GuiTools`, commit the new change and then squash it? but doing so it seems I create another "line" and I lose the possibility to include the last 4 commits with inteactive rebase... 
Also i'm trying to stay with GitKraken: I'm biased by my education towards graphical interfaces...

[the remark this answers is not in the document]
matthijskooijman: If you do a "checkout" of an earlier commit, that changes your working copy to point to that commit, without a clear way to "replay" the later commits that you skipped over going backwards (you could use cherry-pick to get them again, but that's a lot of manual work). Also, if you checkout a specific commit (rather than a branch), you end up in "detached HEAD" state, where your working copy no longer points to a  branch, which is usually not what you want.

In this case, I would suggest using an interactive rebase. I'm not sure how the Git Kraken UI does this, but on the commandline you pass a "base" commit for the rebase, which in this case should be a commit in your current history *before* the one that you want to edit. It's ok if you use an older commit, typically the branching point of your own changes is fine. This is the commit that git will "rewind" to, so all commits *after* the selected commit can be edited.

It seems in kraken you can right click a commit (or maybe only branch, not entirely sure) and select "Interactive Rebase": https://support.gitkraken.com/working-with-repositories/interactive-rebase/

You should then get a list of commits that have been rewound and will be replayed, and you can select actions for each of them. The default is "pick" - just apply the commit and continue (internally, this is AFAIU exactly equivalent to cherry-pick). In this case, you would change the action for the commit you want to change to "Edit". After confirming, kraken should start applying the selected actions. When it comes to "edit", it applies that commit and then stops, so you can make any changes and commit them (with "amend" to modify the commit you want to edit, though you can also create a new commit to *insert* a new commit in the history). Once you are done, you "continue" the rebase, and git will proceed applying the rest of the selected commits.

Does that help?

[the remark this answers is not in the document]
carlopav: Indeed it helps! the point is that i think that gitKraken doesnt' provide the edit feature when interactive rebasing... I'll try with the command line...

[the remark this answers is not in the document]
matthijskooijman: Heh, it seems you're right, it's not listed in the documentation either. I guess it's a bit harder for a GUI to offer this feature (on the comandline, you just drop into a shell and have to call `git rebase --continue` when you're done, for a GUI, it probably needs to keep track of the current state etc. OTOH, Kraken must surely implement this tracking anyway, in case you run into a conflict during a rebase, so it does surprise me a bit that they didn't implement the edit feature in Kraken.

In this case, I think you really need the rebase and edit feature. Often, an alternative is to commit a fixup commit at the end and then reorder and squash it, but in this case, you need to edit code that will be removed in a later commit, so you cannot create such a fixup commit at the end, since the code is already gone.

Note that if you get the rebase with edit going on the commandline, you might run into a conflict when continuing the rebase (since you'll have modified some code which is removed in a later commit, which likely produces a conflict when applying that later comit). Easy enough to fix, of course, just remove the code as intended :-)

src/Mod/Draft/draftguitools/gui_edit.py
```diff
@@ -843,6 +843,20 @@ def deformat_objects_after_editing(self, objs):
                 obj_gui_tools.restore_object_style(obj, self.objs_formats[obj.Name])
 
 
+    def get_specific_object_info(self, obj, pos):
+        """Return info of a specific object at a given position.
+        """
+        selobjs = Gui.ActiveDocument.ActiveView.getObjectsInfo((pos[0],pos[1]))
+        if not selobjs:
+            return
```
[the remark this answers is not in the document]
carlopav: `getObjectsInfo `can return None, even if i'm not sure that could be the case, since we already know the user is over an object at this point. What can be objected is that someone else could want to use that method in the future for other purpouses, so to me better to leave that check... do you agree?

src/Mod/Draft/draftguitools/gui_edit.py
```diff
@@ -843,6 +843,20 @@ def deformat_objects_after_editing(self, objs):
...
+        for info in selobjs:
+            if not info:
+                return
```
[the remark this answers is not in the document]
carlopav: done!

src/Mod/Draft/draftguitools/gui_edit_draft_objects.py
carlopav: done, in the second last commit. It's not so clean to introduce those methods and then remove them, I know, but but... pleeeeease, ignore it!

carlopav:
@matthijskooijman thanks, what I learned from the first command line approach: 
- do not use `MS Windows Command Prompt`, but use `Git Bash `instead
- `git status` and `git log --oneline` are really helpful (to quit form git log press q and enter!).
- `git rebase --interactive` (yes, i meant `--interactive`, not `-interactive`) is just wonderful! mark the commit you want to change with `edit`, let the rebase arrive to that commit, do the changes and save the file, `git status` to check what you have changed, `git add filename` to stage the changed file, `git rebase --continue` to go to the next commit! wow!

src/Mod/Draft/draftguitools/gui_edit.py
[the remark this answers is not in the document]
matthijskooijman: Yeah, maybe better be safe than sorry (though one could argue that getObjectsInfo should just always return a list, possibly an empty one if nothing is found, but if that is not currently the case, maybe indeed leave it like this).

carlopav:
@matthijskooijman can we unmark the PR as Draft? what do you think?

matthijskooijman:
If you have no further changes you want to make, then yes :-)

carlopav:
@yorikvanhavre when you have some time the PR is ready to be reviewed/merged!
cheers!

yorikvanhavre:
So be it ;)

carlopav:
thanks!

PR #4812, opened 2021-05-24
Path: PathArray based on multiple bases, and choice of direction

jimzim111:

---

### DIFF
--- a/src/Mod/Path/PathScripts/PathArray.py
+++ b/src/Mod/Path/PathScripts/PathArray.py
@@ -24,6 +24,7 @@ import FreeCAD
 import FreeCADGui
 import Path
 import PathScripts
+from PathScripts import PathLog
 from PySide import QtCore
 import math
 
@@ -36,8 +37,8 @@ def translate(context, text, disambig=None):
 class ObjectArray:
 
     def __init__(self, obj):
-        obj.addProperty("App::PropertyLink", "Base",
-                        "Path", "The path to array")
+        obj.addProperty("App::PropertyLinkList", "Base",
+                        "Path", "The path(s) to array")
         obj.addProperty("App::PropertyEnumeration", "Type",
                         "Path", QtCore.QT_TRANSLATE_NOOP("App::Property", "Pattern method"))
         obj.addProperty("App::PropertyVectorDistance", "Offset",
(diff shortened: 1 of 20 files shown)


jimzim111:
These changes are intended to help reduce tool travel time.  Discussion of the first one was at https://forum.freecadweb.org/viewtopic.php?f=15&t=58768&p=505517#p505517.

The first change allows PathArrays to be based on multiple Paths instead of just one.  It makes a copy of all of the bases in each arrayed position before moving on to the next position.

The second change allows swapping the direction in which copies are arrayed (X first instead of Y first).

Russ4262 (approved):
I built and tested.  Works as advertised.  Code changes look good.

src/Mod/Path/PathScripts/PathArray.py
```diff
@@ -52,6 +54,12 @@ def __init__(self, obj):
                         "Path", "The number of copies in Linear 1D and Polar pattern")
         obj.addProperty("App::PropertyVector", "Centre",
                         "Path", "The centre of rotation in Polar pattern")
+        obj.addProperty("App::PropertyBool", "SwapDirection",
```
sliptonic, line 57: Please wrap all user-visible property descriptions in the QT_TRANSLATE_NOOP function so they get translated into other languages.   See 'pattern method' above.

src/Mod/Path/PathScripts/PathArray.py
```diff
@@ -52,6 +54,12 @@ def __init__(self, obj):
...
+                        "Path", "Make copies in X direction before Y in Linear 2D pattern")
+        obj.addProperty("App::PropertyInteger", "JitterPercent",
+                        "Path", "Percent of copies to randomly offset")
+        obj.addProperty("App::PropertyVectorDistance", "JitterMagnitude",
+                        "Path", "Maximum random offset of copies")
         obj.addProperty("App::PropertyLink", "ToolController",
                         "Path", QtCore.QT_TRANSLATE_NOOP("App::Property", "The tool controller that will be used to calculate the path"))
```
sliptonic, line 64: Does the array really need its own TC reference?  Under what circumstances would the user change the TC in the array and not in the underlying operations?  If the op really needs a local copy but the user shouldn't change it, please hide the property from the user.

src/Mod/Path/PathScripts/PathArray.py
```diff
@@ -67,6 +75,8 @@ def __setstate__(self, state):
         return None
 
     def setEditorProperties(self, obj):
+        obj.setEditorMode('JitterPercent', 0)
+        obj.setEditorMode('JitterMagnitude', 0)
```
sliptonic, line 79: Interesting. What's the use-case for a jitter?

src/Mod/Path/PathScripts/PathArray.py
```diff
@@ -155,54 +166,92 @@ def rotatePath(self, path, angle, centre):
...
+
 
-            obj.ToolController = obj.Base.ToolController
+    def execute(self, obj):
+        if isinstance(obj.Base, list):
+            base = obj.Base
+        else:
+            base = [obj.Base]
+
+        if len(base)>0:
```
sliptonic, line 183: The consequenc if len(base) == 0 isn't visible until the end.  Looking ahead, nothing happens.  So, instead you should check for that case first and  and return if it's true, raising an error of necessary.  Currently all the subsequent code is indented and dependent on that check which is confusing.

src/Mod/Path/PathScripts/PathArray.py
```diff
@@ -155,54 +166,92 @@ def rotatePath(self, path, angle, centre):
...
+            for b in base:
+                if not b.isDerivedFrom("Path::Feature"):
+                    return
+                if not b.Path:
+                    return
+                if not b.ToolController:
+                    return
+                if b.ToolController != obj.ToolController:
+                    # this may be important if Job output is split by tool controller
+                    PathLog.warning('Arrays of paths having different tool controllers are handled according to the tool controller of the first path.')
```
sliptonic, line 194: Not required for acceptance of this PR but if you're up for it, you could also check if the base feature is a dressup and allow for arrays of those as well.  Currently there's no way to make an array of dressed up operations.

src/Mod/Path/PathScripts/PathArray.py
jimzim111: Will do

src/Mod/Path/PathScripts/PathArray.py
jimzim111: I think that makes sense to control the Array's TC by the underlying operations.  Will hide.

src/Mod/Path/PathScripts/PathArray.py
jimzim111: It's an easy way to create a hand-crafted appearance.  When you have a lot of the same cut repeated over and over with small variations it looks more natural and less mechanical.

sliptonic:
Merging.
In the future, unless other people are basing future work on your specific branch, please rebase your PR on upstream master rather than merging those commits back to your branch.  Rebasing keeps the history cleaner

PR #4816, opened 2021-05-29
Fix misc. typos

luzpaz:


luzpaz:
@yorikvanhavre should be good to go

PR #4824, opened 2021-06-01
Draft: gui_utils wrongly used a hard-coded name for construction group

Roy-043:
gui_utils wrongly used a hard-coded name for the construction group. This could lead to two construction groups in the same file. See: https://forum.freecadweb.org/viewtopic.php?f=3&t=59055

- [*]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [*]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [*]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [*]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`

### DIFF
--- a/src/Mod/Draft/draftutils/gui_utils.py
+++ b/src/Mod/Draft/draftutils/gui_utils.py
@@ -350,7 +350,7 @@ def format_object(target, origin=None):
     if ui:
         doc = App.ActiveDocument
         if ui.isConstructionMode():
-            lcol = fcol = ui.getDefaultColor("constr")
+            lcol = ui.getDefaultColor("constr")
             tcol = lcol
             fcol = lcol
 
--- a/src/Mod/Draft/draftutils/gui_utils.py
+++ b/src/Mod/Draft/draftutils/gui_utils.py
@@ -353,10 +353,13 @@ def format_object(target, origin=None):
             lcol = fcol = ui.getDefaultColor("constr")
             tcol = lcol
             fcol = lcol
-            grp = doc.getObject("Draft_Construction")
+
+            # Get the construction group or create it if it doesn't exist
+            gname = utils.get_param("constructiongroupname", "Construction")
+            grp = doc.getObject(gname)
             if not grp:
-                grp = doc.addObject("App::DocumentObjectGroup", "Draft_Construction")
-                grp.Label = utils.get_param("constructiongroupname", "Construction")
+                grp = doc.addObject("App::DocumentObjectGroup", gname)
+
             grp.addObject(target)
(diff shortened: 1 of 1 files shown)


yorikvanhavre:
Using a hard-coded group name is exactly the intent, so there aren't two in a same file. You can always locate the construction group, independently on if it was created on another machine with other settings... Note that the code checks if a group with that name exists before creating one.
What I understand from the problem stated on the forum is that a new construction group is created inside the active Part, which is probably not what we want. So IMHO the correct way to fix this is to make sure the construction group is created in the document root, not the active Part.

Roy-043:
If that is the case then gui_groups.py needs to be changes accordingly.

Roy-043:
I seem to have closed this somehow... I do not know how that happened.

yorikvanhavre:
> IMO. 'Construction group name' should become 'Construction group label'.

I agree too

Roy-043:
> What I understand from the problem stated on the forum is that a new construction group is created inside the active Part, which is probably not what we want.

That is not the problem. The construction group is not created in the active part. As already mentioned gui_groups.py needs to be updated as it does not create a construction group with the same hard-coded name. Draft_AddConstruction will still create a construction group using the name from the parameters. It is the presence of two construction groups, one with the hard-coded name, the other with the name from the parameters, that is behind the forum post.

Taking this a step further: You would also need to change the name of the parameter from "constructiongroupname" to "constructiongrouplabel" I suppose. And do some cleanup? I notice that there is no code to delete a parameter in draftutils.utils.

```
if utils.get_param("constructiongroupname") == "":
    grp-lbl = utils.get_param("constructiongrouplabel", "Construction")
elif utils.get_param("constructiongrouplabel") != "": # constructiongrouplabel has to be added to draftutils.utils for this to work
    grp-lbl = utils.get_param("constructiongrouplabel")
else:
    grp-lbl = utils.get_param("constructiongroupname")
    utils.set_param("constructiongroupname", "")
    utils.set_param("constructiongrouplabel", grp-lbl)
```

yorikvanhavre:
Good catch @Roy-043 I just fixed gui_groups.py in 54139caff9
See what you think of it? I don't think there is a need to change the parameter, just use it only for group label and not for the name

Roy-043:
That will work, thanks. I'll close this one.

PR #4827, opened 2021-06-02
Draft: Fix gui_stretch

Roy-043:
If a wire is created to replace a rectangle, the MakeFace property of the wire should match that of the rectangle.


yorikvanhavre:
Makes sense. Thanks!

PR #4830, opened 2021-06-03
[TD] comment the new centerline behavior

donovaly:
add missing info to understand the new centerline behavior as introduced with commit b981f905

### DIFF
--- a/src/Mod/TechDraw/App/Cosmetic.cpp
+++ b/src/Mod/TechDraw/App/Cosmetic.cpp
@@ -980,12 +980,16 @@ std::pair<Base::Vector3d, Base::Vector3d> CenterLine::calcEndPoints(DrawViewPart
 
 bool CenterLine::Circulation(Base::Vector3d A, Base::Vector3d B, Base::Vector3d C)
 {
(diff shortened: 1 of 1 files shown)


src/Mod/TechDraw/App/Cosmetic.cpp
```diff
@@ -980,12 +980,16 @@ std::pair<Base::Vector3d, Base::Vector3d> CenterLine::calcEndPoints(DrawViewPart
 
 bool CenterLine::Circulation(Base::Vector3d A, Base::Vector3d B, Base::Vector3d C)
 {
+    // the determinant of this matrix calculates the aread of a triangle, see
```
luzpaz, line 983: Typo: should be `area of a triangle` not `aread`

luzpaz (changes requested):

src/Mod/TechDraw/App/Cosmetic.cpp
0penBrain: @luzpaz : in such a case, you can use suggestion feature so the author can just accept it. ;)
```suggestion
    // the determinant of this matrix calculates the area of a triangle, see
```

PR #4835, opened 2021-06-06
[Sketch] add new sketch object rounded rectangle

donovaly:
while constructing I miss often that one cannot insert directly rounded rectangles. As it is, one needs first to insert a rectangle, then go to fillet mode, then click 8 times, then set equal constraint, then often also replace the vertical/horizontal constraints since a rectangle must not necessarily be parallel to a sketch axis

As solution this PR adds rounded rectangles in one step. This saves a lot of time since this object is e.g. often used for milled rectangular pockets, for 3D-printed pads etc.

Since the slot is a special case of a rounded rectangle and the use cases are similar as well, I grouped the lot and rounded rectangle.

Here is a screencast how I implemented it:

d6Lnbwl4UD

Forum thread: https://forum.freecadweb.org/viewtopic.php?f=17&t=59210

(The code beautification in this PR of existing code was kindly done automatically be MSVC.)

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandAlterGeometry.cpp
+++ b/src/Mod/Sketcher/Gui/CommandAlterGeometry.cpp
@@ -82,7 +82,7 @@ CmdSketcherToggleConstruction::CmdSketcherToggleConstruction()
     // list of toggle construction commands
     Gui::CommandManager &rcCmdMgr = Gui::Application::Instance->commandManager();
     rcCmdMgr.addCommandMode("ToggleConstruction", "Sketcher_CreateLine");
-    rcCmdMgr.addCommandMode("ToggleConstruction", "Sketcher_CreateRectangle");
+    rcCmdMgr.addCommandMode("ToggleConstruction", "Sketcher_CompCreateRectangles");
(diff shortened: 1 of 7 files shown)


adrianinsaval:
very useful feature, are the construction lines necessary? Isn't it enough to use two point on object constraints? Like this:
Screenshot_20210606_185526
Also, minor correction/advise: you don't really need to click 8 times to make 4 fillets, just click once on each vertex instead of each pair of lines

donovaly:
> very useful feature, are the construction lines necessary? Isn't it enough to use two point on object constraints?

How is this done? And  how can you later modify the rectangle and set e.g dimensional constraints?
Let's discuss this here: https://forum.freecadweb.org/viewtopic.php?f=17&t=59210

> Also, minor correction/advise: you don't really need to click 8 times to make 4 fillets, just click once on each vertex instead of each pair of lines

Indeed. Thanks!

donovaly:
>  Isn't it enough to use two point on object constraints?

Thanks. This is sufficient and saves some objects. I updates the PR accordingly - see the updates screencast in the initial comment of this PR:

PR #4836, opened 2021-06-06
[Sketch] update icons - part1

donovaly:
save icons with uniform DPI and as plain SVG


luzpaz:
Is there a forum thread associated with this commit?

matthijskooijman:
And why is this in four parts? Commit messages seem nearly identical, if this is a single logical change, why not just have one commit? If there is some reason to separate this in multiple commits, maybe one pull request with four commits would be easier to manage?

donovaly:
> Is there a forum thread associated with this commit?

Not necessary, I just saved all Sketcher icons as plain SVG

PR #4843, opened 2021-06-08
[sketch] rework the slot command

donovaly:
in practical work it turned out that the slot command is not as convenient as it should because one cannot snap to a proper endpoint and one is restricted to work perpendicular to the sketch axes

This PR fixes this by setting the second snap point to the center of the second arc and removing the perpendicularity.

Benefits:
- when you want to get the slot at a line, edge or sketch axis, set the two center points accordingly. The line does not have to be parallel to one of the sketch axes
- when you want to define the slot using 2 points in the sketch you can do this now

Here is a screencast how it works:
KZooJ8S2lg

Forum thread: https://forum.freecadweb.org/viewtopic.php?f=17&t=59243

(The code cosmetics in this PR was kindly automatically done by MSVC.)


luzpaz:
@donovaly just one point of feedback. Please use `[Sketcher]` and not `[Sketch]` in your PRs for Sketcher workbench in the future. It makes it easier for us to grep the `git log` accurately. Thanks in advance!

PR #4844, opened 2021-06-08
Improve icon installation

pinotree:
Improve the way icons are installed, mostly in the XDG hicolor icon theme:
- drop the old XPM application icon, as installed in the legacy pixmaps location, and generally not needed with the icons in the XDG hicolor icon theme
- rename the application icons in the XDG hicolor icon theme to `org.freecadweb.FreeCAD`, to match what the desktop file specifies already; this way, instead of only one icon (the SVG one), there are also 4 PNG icons available for XDG icon theme users
- drop a duplicate version of the SVG application icon
- drop the manual icon name for the XDG MIME type, as there is already an icon for it installed with the proper name

In the pull request checklist, there is:

It is hard to follow this and split these changes across, as they would lose the "atomicity" -- renaming the application icon and changing all the places at the same time IMHO is a single change, otherwise it will break on bisection (and it will be hard to correlate the different changes together).


luzpaz:
HI @pinotree thanks for the PR, please open forum thread to discuss it with the community. Thanks

yorikvanhavre:
I would vote against this change, as it binds the icon to our current internet web domain, which will hopefully change soon ;)
And in any case, FreeCAD is FreeCAD, not a "child of the freecadweb.org domain"... (it's more the contrary, actually) I don't think we should make it a habit of going that way unless it's mandatory. Just my opinion of course

yorikvanhavre:
Since we are in the middle of a domain name change, I would rather close this PR, and we decide of an official position regarding domain-dependent resource names...

PR #4846, opened 2021-06-08
Draft: Fix wrong tree icon for Draft_PointArray

Roy-043:
The code wrongly used the PointList property to detect a Draft_PointArray. PointObject should be used instead.

For consistency also changed the code to detect a Draft_PathTwistedArray to use the RotationFactor property.


Roy-043:
Forum topic were this was reported:
https://forum.freecadweb.org/viewtopic.php?f=3&t=59236

PR #4860, opened 2021-06-13
[ArchWindow] add SketchArch attachment support

paullee0:
FreeCAD Forum Discussion :
- on Intuitive Automatic Windows/Doors Placement
- https://forum.freecadweb.org/viewtopic.php?f=23&t=50802

---

### DIFF
--- a/src/Mod/Arch/ArchWindow.py
+++ b/src/Mod/Arch/ArchWindow.py
@@ -531,6 +531,14 @@ class _Window(ArchComponent.Component):
         obj.IfcType = "Window"
         obj.MoveWithHost = True
 
+        try:
+            import ArchSketchObject
+            print(" Done :  __init__() import ArchSketchObject ")
+            ArchSketchObject.ArchSketch.setPropertiesLinkCommon(self, obj)  # ,linkFp)
+            print(" Done :  __init__() ArchSketchObject.setPropertiesLinkCommon() ")
+        except:
+            print(" ArchSketchObject add-on module is not installed yet, __init__() ")
+
     def setProperties(self,obj):
 
         lp = obj.PropertiesList
@@ -579,6 +587,14 @@ class _Window(ArchComponent.Component):
 
         ArchComponent.Component.onDocumentRestored(self,obj)
         self.setProperties(obj)
+
+        try:
+            import ArchSketchObject
+            print("  Done :  onDocumentRestored() import ArchSketchObject ")
+            ArchSketchObject.ArchSketch.setPropertiesLinkCommon(self, obj, mode='ODR')
+            print("  Done :  onDocumentRestored() ArchSketchObject.setPropertiesLinkCommon() ")
+        except:
+            print(" ArchSketchObject add-on module is not installed yet, onDocumentRestored() ")
 
     def onBeforeChange(self,obj,prop):
 
@@ -865,6 +881,28 @@ class _Window(ArchComponent.Component):
                 obj.Placement = pl
         if hasattr(obj,"Area"):
             obj.Area = obj.Width.Value * obj.Height.Value
+
+        try:
+            import ArchSketchObject
+            print(" Done : execute() import ArchSketchObject")
+            ArchSketchObject.updateAttachmentOffset(obj)
+            print(" Done : execute() updateAttachmentOffset(obj)")
+        except:
+            print(" ArchSketchObject add-on module is not installed yet, execute()")
+
+    def appLinkExecute(self, fp, linkFp, index, linkElement):
+
+        try:
+            import ArchSketchObject
+            print(" Done :  appLinkExecute() import ArchSketchObject")
+            ArchSketchObject.ArchSketch.setPropertiesLinkCommon(self, fp, linkFp)
+            print(" Done :  appLinkExecute() ArchSketchObject.setPropertiesLinkCommon()")
+        except:
+            print(" ArchSketchObject add-on module is not installed yet, appLinkExecute() ")
+        try:
+            ArchSketchObject.updateAttachmentOffset(fp, linkFp)
+        except:
+            print(" ArchSketchObject add-on module is not installed yet, appLinkExecute() - updateAttachmentOffset() not available")
 
     def getSubVolume(self,obj,plac=None):
 


matthijskooijman:
Is it me, or is the ArchSketchObject.py file missing from your commit?

paullee0:
> Is it me, or is the ArchSketchObject.py file missing from your commit?

Thanks for checking :)

The discussion is in https://forum.freecadweb.org/viewtopic.php?f=23&t=50802&start=140#p509116 , the entire experimental module is suggested to be installed via the add-on manager.  It is only available for user who installed it manually at the moment.

matthijskooijman:
Ah, I see, that explains the try/catch, I could have seen that if I read properly :-)

In that case, I'm not so sure this PR is something to be merged in master. I'm not sure what the policies in FreeCAD are, but from a software development perspective, I would not put references to externally installable plugins in the core code like this, since that creates a particular dependency on that external code that is hard to track and does not scale well (what if a doze more of these addons would be created?).

Instead, in cases like these, an often-used approach is that the core code provides hooks, callbacks or signals of some sort, which can be used by the addon to do the right thing at the right moment. This means that the core code stays generic: It just offers an interface to be used by any addon that wants to (or maybe none), while the initiative, and the burden for version compatibility, stays with the addon.

I'm not sure how this should work in this particular case: The various arch objects could offer specific signals or global callbacks or so, but I guess some FreeCAD-wide solution would make more sense. Maybe the proxy-support is actually the existing mechanism for this, but that would require making a subclass of e.g. ArchWindow and set the object's proxy type to that subclass. Alternatively, I think there is the concept of object extensions, that could probably be used without needing to replace the object's proxy type (but not 100% sure if they can be defined in python too?)

paullee0:
Thanks for the suggestion.  Not trained as programmer, there is limited knowledge for me to see the available options, so I did with similar approach previously :D

Can you help elaborating your idea, or maybe there is available tutorial on similar cases?

yorikvanhavre:
I agree with @matthijskooijman ... With this PR, each and every window created by a person with the addon not installed will print a lot of output. Another problem is that the person reading the code will have no clue of what is wrong.

I have nothing against using optional functionality from addon workbenches, after all it's not different than checking for Python libraries, but it should be made very clear and silent if the user doesn't want to install anything. So I would suggest 1) encapsulating all ArchSketchObject functionality in only one place, not (as much as possible) spread all over the code, 2) add very clear comments in the code about what is going on and where to find the needed modules and 3) do not print anything if the module is not there (or maybe some discrete information somewhere IF it makes sense for the user to install the addon)

matthijskooijman:
> Can you help elaborating your idea, or maybe there is available tutorial on similar cases?

The general idea of my suggestion is like how the signal / event handlers work, where the generic code just exposes a signal or event and the addon registers an event handler for such a signal or event. I'm not sure how this same thing is done within FreeCAD currently (and I wouldn't recommend inventing some event handler structure just for this case, AFAICS FreeCAD need *less* complexity and fragmentation, not *more*). Maybe there are existing events available already, accessible from Python, but I'm not sure.

As for my more specific suggestions of using a Proxy subclass, or object extension, I'm not yet familiar enough with FreeCAD to offer a more specific direction here... IIUC object extensions can be added to an object, and each extension can provide additional functionality for an object. One aspect of this is that when e.g. onChanged is called on the object, it is *also* called on each extension in turn, which can then take additional actions as needed. This kind of thing seems like would be appropriate for ArchSketch.

Hoever, from a quick look over the source, I suspect that an object extension (i.e. like the existing attachment extension, which you would add to an existing object using e.g. `obj.addExtension("Part::AttachExtensionPython")`) can only be defined from C++ (there is an DocumentObjectExtensionPy class, but I think that is only intended to support accessing a C++ extension object from Python, not defining one from Python, but I'm not quite sure.

> I have nothing against using optional functionality from addon workbenches, after all it's not different than checking for Python libraries, 

I'm not sure I agree with this entirely. When including an (optional) library, you subsequently making specific calls to that library, which has a clear and well-defined interface and control over what happens and is supposed to happen remains with the FreeCAD core.

With a fairly free-form call to e.g. `ArchSketchObject.ArchSketch.setPropertiesLinkCommon`, this is not so well-defined. What does that function do exactly? When should it be called? What happens if FreeCAD or the ArchSketch WB is changed later, how is compatibility arranged? Can we ever remove these calls again later? What if ArchSketch has different requirements in the future (i.e. what if in addition to `setPropertiesLinkCommon`, some extra things need to happen when called from `__init__`, but not in `onDocumentChanged`? This now requires changes in FreeCAD core, which would break older ArchSketch versions as well).

So *if* you would accept something like the current form (instead of moving to some extension object or other more generic event handler style method as I suggested vaguely above), in addition to the conditions posed by @yorikvanhavre (which I agree with), I would also suggest to make the interface more clear and well-defined. In particular, this could be done by calling e.g. `ArchSketchObject.ArchSketch.ArchWindowInit`, `ArchSketchObject.ArchSketch.ArchWindowOnDocumentRestored`, etc. In other words, name the methods based on when they should be called, rather than what they *do*, so it is more clear from the FreeCAD core perspective how to treat these calls in the future, and preventing needing changes in FreeCAD core when changes are made to ArchSketch. For more forward compatibility, you could even consider moving this stuff from ArchWindow to ArchComponent (even less chance for changes needed in FreeCAD core in the future if ArchSketch is extended), but that would also make the performance impact of this change more pronounced.

paullee0:
> I have nothing against using optional functionality from addon workbenches, after all it's not different than checking for Python libraries, but it should be made very clear and silent if the user doesn't want to install anything. So I would suggest 1) encapsulating all ArchSketchObject functionality in only one place, not (as much as possible) spread all over the code, 2) add very clear comments in the code about what is going on and where to find the needed modules and 3) do not print anything if the module is not there (or maybe some discrete information somewhere IF it makes sense for the user to install the addon)

Thanks all for mentoring :)

It seems there are 4 additions are required and some of them can be place in ArchComponent.py indeed.  In that case, the 'additional functionality' will be available to all ArchObjects, other than ArchEquipment and ArchWindow which were tested in the quoted discussion thread.

1.  __init__() :-   Seems ok to be in ArchComponent.__Init()
2.  onDocumentRestored() :-  Seems ok to be in ArchComponent.onDocumentRestored()
3.  execute() :-  Seems this needs to be in ArchEquipment, ArchWindow etc.
4.  appLinkExecute() :-  This is support to Realthunder's Links, seems needs to be in every Arch Objects as execute()

Will have a closer look and test if above quick idea works.

Thanks again !

paullee0:
> Hoever, from a quick look over the source, I suspect that an object extension (i.e. like the existing attachment extension, which you would add to an existing object using e.g. `obj.addExtension("Part::AttachExtensionPython")`) can only be defined from C++ (there is an DocumentObjectExtensionPy class, but I think that is only intended to support accessing a C++ extension object from Python, not defining one from Python, but I'm not quite sure.

Without a deep understanding of how underlying FreeCAD core works, but just reading forum / wiki about Attachment Extension support (happens intending to add that to Arch Objects also), it is something in C++, unreachable for me :D 

> > I have nothing against using optional functionality from addon workbenches, after all it's not different than checking for Python libraries,
> 
> I'm not sure I agree with this entirely. When including an (optional) library, you subsequently making specific calls to that library, which has a clear and well-defined interface and control over what happens and is supposed to happen remains with the FreeCAD core.
> 
> With a fairly free-form call to e.g. `ArchSketchObject.ArchSketch.setPropertiesLinkCommon`, this is not so well-defined. What does that function do exactly? When should it be called? What happens if FreeCAD or the ArchSketch WB is changed later, how is compatibility arranged? Can we ever remove these calls again later? What if ArchSketch has different requirements in the future (i.e. what if in addition to `setPropertiesLinkCommon`, some extra things need to happen when called from `__init__`, but not in `onDocumentChanged`? This now requires changes in FreeCAD core, which would break older ArchSketch versions as well).
> ..

I will see if Yorik's indicated framework works and make things clearer as suggested :)  In fact, the appLinkExceute() is a new callback Realthunder provides support for Links.  It is something interesting  -  discussion can be found at [ObjectPython / Part FeaturePython lose its 'Proxy Methods' ?](https://forum.freecadweb.org/viewtopic.php?f=22&t=42184&start=50)

paullee0:
Maybe a check of the ArchObject calling the Component.__init__(), and just add the external module support if it is Equipment / Window, which were tested and discussed in the forum thread.

```
class Component(ArchIFC.IfcProduct):
    ...
    def __init__(self, obj):
        ...

        ## TODO 2021.6.16
        SketchArchSupportedTypes = ['Equipment', 'Windoow']  # Better become Class attribute
        if Draft.getType(obj.getLinkedObject()) in SketchArchSupportedTypes:
        try:
            import ArchSketchObject
            ArchSketchObject.ArchSketch.setPropertiesLinkCommon(self, obj)
        except:
            pass

	# further remark on Draft.getType(obj.getLinkedObject())
	# checking any need to use getLinkedObject() as Links would not has only appLinkExecute() and would not call __init()
```

matthijskooijman:
> Maybe a check of the ArchObject calling the Component.init(), and just add the external module support if it is Equipment / Window, which were tested and discussed in the forum thread.

That would reduce the performance impact slightly, but again puts archsketch-specific logic in the FreeCAD core (in other words, if you then add support for some other ArchComponent in SketchArch, the FreeCAD core must be modified again).


There is a caveat with the `execute()` method: Not all subclasses of ArchComponent call their super method. I consider this a bug, but it seems that this bug actually hides another bug in movewithhost handling, so adding 

> appLinkExecute() :- This is support to Realthunder's Links, seems needs to be in every Arch Objects as execute()

I'm not sure what you're saying here?


Note that there might be one caveat when you put the calls in ArchComponent instead of the subclasses: This only works if the subclasses properly call their superclass methods as well (which they should). I haven't checked the methods you're referring to, but I've previously seen that `ArchWall.onBeforeChanged()` does not call its superclass (which is harder to fix than it seems since fixing this bug exposes others, see https://forum.freecadweb.org/viewtopic.php?f=23&t=57223&p=502966&hilit=super#p502966). I don't think you're using `onBeforeChanged`, but the same situation might be the case for other methods too.

yorikvanhavre:
What I would do here (in both this one and #4858) is basically:
- remove all print() functions that are not errors... No need to flood the user. I think it's better to add  something to the Arch wiki page
- add something like  `# this code is executed if the ArchSketch workbench/module is installed` wherever you use Arch Sketch code
That way there is not much to change, and it will be very clear to anyone looking at the code

paullee0:
> What I would do here (in both this one and #4858) is basically:
> 
>     * remove all print() functions that are not errors... No need to flood the user. I think it's better to add  something to the Arch wiki page
> 
>     * add something like  `# this code is executed if the ArchSketch workbench/module is installed` wherever you use Arch Sketch code
>       That way there is not much to change, and it will be very clear to anyone looking at the code

Hi, thanks for  the comments !

I had in fact 're-paragraph' the added codes and tested in my own pc for quite some while since June.  May have another gap to learn editing wiki, tidying up codes soon :)

berndhahnebach:

src/Mod/Arch/ArchWindow.py
```diff
@@ -531,6 +531,14 @@ def __init__(self,obj):
...
         obj.MoveWithHost = True
 
+        try:
+            import ArchSketchObject
+            print(" Done :  __init__() import ArchSketchObject ")
+            ArchSketchObject.ArchSketch.setPropertiesLinkCommon(self, obj)  # ,linkFp)
+            print(" Done :  __init__() ArchSketchObject.setPropertiesLinkCommon() ")
+        except:
+            print(" ArchSketchObject add-on module is not installed yet, __init__() ")
+
```
yorikvanhavre, line 541: I would do all these blocks like this, simply:

```suggestion
      try:
            # use the ArchSketchObject addon if available
            import ArchSketchObject
            ArchSketchObject.ArchSketch.setPropertiesLinkCommon(self, obj)  # ,linkFp)
        except:
            pass
```

paullee0: Thanks reviewing :)

I had revamped early in Sept, maybe I try ArchEquipment first.

paullee0: #5241 Still something wrong with git ... :(

paullee0: Completely start over cloning everything again and open a new PR #5243, closed the #5241.

Hope do it right this time :)

PR #4889, opened 2021-06-23
Draft: Fix misplaced Draft_Edit nodes for BezCurve

Roy-043:
The poles in the resetTrackersBezier function are already in the global coordinate system. They should therefore not be translated from the LCS to the GCS.

---


carlopav:
Can you test the following: create an App::Part, move the bezcurve inside, move the Part or rotate It. Try to Edit the bezcurve. 
Are editpoints in the right Place?
I have not my laptop with me this evening...

carlopav (commented):
The sentence reflect how every object work except the bezcurve. I'd try to keep It if we can fix the bezcurve behaviour

Roy-043:
> Are editpoints in the right Place?

No they are not. I did not consider that case. It is more complex than I thought.
how about this solution:
```
        for index, pwm in enumerate(pointswithmarkers):
            p, marker = pwm
            if obj.Parents:
                p = obj.Placement.inverse().multVec(p)
                p = obj.getGlobalPlacement().multVec(p)
            self.trackers[obj.Name].append(trackers.editTracker(p, obj.Name,
                index, obj.ViewObject.LineColor, marker=marker))
```
It may also be interesting to look for the 1st geometric parent of the curve. See https://forum.freecadweb.org/viewtopic.php?f=22&p=512227#p512227

Roy-043:
> he knows the overall behaviour

But in the code we find globalize_vectors f.e. The comment: 'No conversion to global coordinate system is needed' then seems incorrect to me.

carlopav:
Uff.. that thing is complicated...
The point is that the BezCurve behaves differently from every other object, why? because it's trackers need to have different shapes based on if they are on the curve points, or are just handles, and if they are on the curve points, they can have also 3 different shape (sharp, tangent, symmetric).
When we generalized Draft Edit, to remove a lot of duplicated code, making easy to extend the command to new objects and to context menu, BezCurve remained unfinished.

We could take 2 different paths:

1. generalize more, so every object can define custom trackers (that's a lot of refactoring again, but it's the best way to me);
2. let it like this and put a small patch on the bug you spotted (only problem is that it leaves a specific code into the generic Edit module);

I'd have no much time at the moment, so better to take the second path as you did :)

`                p = obj.Placement.inverse().multVec(p)
                p = obj.getGlobalPlacement().multVec(p)`

I think you don't need to check for parents, this code should always work (except if you have a nested link in the tree), because the GlobalPlacement already contains the Placement itself, so if they are equal they just "eliminate" each other.
You could also theoretically condensate it in:
`p = obj.getGlobalPlacement().multiply(obj.Placement.inverse().multVec(p))`

can you try it?

carlopav:
> 
> 
> > he knows the overall behaviour
> 
> But in the code we find globalize_vectors f.e. The comment: 'No conversion to global coordinate system is needed' then seems incorrect to me.

yes, true, the sentence was meant to tell the user that the conversion is already done in the Edit command, so the "GuiTools" class (that is the object that the user can derive and add to the `DocumentObject `to enable editing it with Draft Edit) do not need to take it into account, but can work directly with object LCS coordinates.

Roy-043:
I have made a mess here. I'll open a new branch to handle this issue. And close this one.

PR #4913, opened 2021-07-09
Base: Added stiffness units

Helios113:
Adds units of stiffness to src/Base, it sets N/m as the smallest unit in the metric system and lbf/inch in the imperial system. This unit will be used for implementing spring boundary conditions for a new FEM solver.


luzpaz:
Hi @Helios113 , this is for MoFEM solver, correct?

Helios113:
Hi @luzpaz , yes this is for MoFEM, I wrote it on the forum, not sure if I should have put it here as well.
Here is the link: https://forum.freecadweb.org/viewtopic.php?f=17&t=60024
Regards

luzpaz:
hey @berndhahnebach, what is this PR pending on? (wmayer's input?)

src/Base/Quantity.h
```diff
@@ -244,6 +244,8 @@ class BaseExport Quantity
     static Quantity KiloNewton;
     static Quantity MegaNewton;
 
+    static Quantity NewtonPerMeter;
+    
     static Quantity Pascal;
     static Quantity KiloPascal;
```
donovaly, line 250: What is the typical prefix? I mean if the typical use case is e.g. kN/m, then you should also define this here

donovaly (changes requested):
Many thanks for this addition.

However, before this can be applied more changes are necessary. You need to update e.g. also the lexer. To see what needs to be done to add a new unit, you can take this commit as reference:
https://github.com/FreeCAD/FreeCAD/commit/bcc669b930ae655d16c361f5e308ec14a9

For how to update the lexer, see the first comment in the file https://github.com/FreeCAD/FreeCAD/blob/master/src/Base/QuantityParser.l

berndhahnebach:
great.

Would it be possible to squash all commits into one commit?

cheers bernd

Helios113:
I messed this one up, so they are definitely not squashed properly. I will make a new one where it's just one commit.

luzpaz:
See #4963

PR #4914, opened 2021-07-10
[PD] fix property issues

donovaly:
- Helix: the property Turns has the App::PropertyFloatConstraint thus also the pointer to it must have the same property

- LinearPattern and Helix: make constraints static class member (forgot this in commit 3704ad68)

- Draft: make the constraint const

- PolarPattern and LinearPattern assure that the header file is loaded at last to avoid potential troubles

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.h
+++ b/src/Mod/PartDesign/App/FeatureHelix.h
@@ -90,7 +90,7 @@ protected:
 
     // handle changed property
     virtual void handleChangedPropertyType(Base::XMLReader& reader, const char* TypeName, App::Property* prop);
-    static const App::PropertyFloatConstraint::Constraints Helix::floatTurns;
+    static const App::PropertyFloatConstraint::Constraints floatTurns;
     static const App::PropertyAngle::Constraints floatAngle;
 
 private:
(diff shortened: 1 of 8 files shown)


src/Mod/PartDesign/App/FeatureHelix.h
```diff
@@ -90,6 +90,8 @@ class PartDesignExport Helix : public ProfileBased
 
     // handle changed property
     virtual void handleChangedPropertyType(Base::XMLReader& reader, const char* TypeName, App::Property* prop);
+    static const App::PropertyFloatConstraint::Constraints Helix::floatTurns;
```
chennes, line 93: The "Helix" in this context is redundant (and you don't use it on the next line):
```suggestion
    static const App::PropertyFloatConstraint::Constraints floatTurns;
```

chennes (approved):
These changes look good. It compiles and runs without issue (tested on Windows 10, Visual Studio 16.8.6). Just one incredibly minor comment, not a merge blocker.

src/Mod/PartDesign/App/FeatureHelix.h
yorikvanhavre: do you want to apply the change, @donovaly ? apparently it's just a push of a button now :)

yorikvanhavre:
Ok I did it :innocent:

PR #4915, opened 2021-07-11
Draft: fixed Snap behaviour when no document is active.

carlopav:
Draft toolbar and statusbar can be activated also when no document is active, because Snap is not related to active document.

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_snaps.py
+++ b/src/Mod/Draft/draftguitools/gui_snaps.py
@@ -35,6 +35,7 @@ from PySide.QtCore import QT_TRANSLATE_NOOP
 import FreeCADGui as Gui
 import draftguitools.gui_base as gui_base
 
+from draftutils.messages import _msg, _log
 from draftutils.translate import translate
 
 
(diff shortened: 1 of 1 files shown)


src/Mod/Draft/draftguitools/gui_snaps.py
```diff
@@ -102,8 +103,20 @@ def sync_snap_statusbar_button(button, status):
...
+        self.command_name = name
 
-class Draft_Snap_Lock(gui_base.GuiCommandSimplest):
+    def IsActive(self):
+        return True
+
+    def Activated(self):
+        _log("GuiCommand: {}".format(self.command_name))
+        _msg("{}".format(16*"-"))
+        _msg("GuiCommand: {}".format(self.command_name))
```
yorikvanhavre, line 116: You sure about this? :sweat_smile: it looks very verbose...

yorikvanhavre, line 116:
```suggestion
        #_msg("{}".format(16*"-"))
        #_msg("GuiCommand: {}".format(self.command_name))
```

yorikvanhavre: I think I'd do this :innocent:

PR #4924, opened 2021-07-17
Set default keycodes for 3dconnexion SpacePilot Pro

dbdbdb:
These keycodes matches what my device generate using
spacenavd in Fedora 33.

The old xml definition define two operations on the same KeyCode, one
for short press and one for long press. That doesn't work at all on my
system and the SpacePilot Pro device generate separate KeyCodes for
short presses and long presses.

https://forum.freecadweb.org/viewtopic.php?f=8&t=60265

### DIFF
--- a/src/Gui/3Dconnexion/3DConnexion.xml
+++ b/src/Gui/3Dconnexion/3DConnexion.xml
@@ -64,35 +64,36 @@
 <ButtonMap DeviceName="SpacePilot Pro">
    <Mapping>
       <Map Description="Menu"          KeyCode="0"  DownTime="short" Command=""/>
-      <Map Description="Fit"           KeyCode="1"  DownTime="short" Command=""/>
+      <Map Description="Fit"           KeyCode="1"  DownTime="short" Command="Std_ViewFitAll"/>
       <Map Description="Top"           KeyCode="2"  DownTime="short" Command="Std_ViewTop"/>
-      <Map Description="Bottom"        KeyCode="2"  DownTime="long"  Command="Std_ViewBottom"/>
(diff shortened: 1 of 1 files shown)


yorikvanhavre:
I'd like someone else to test, just to make sure it doesn't break any working workflow...

luzpaz:
@dbdbdb know any other 3dconnexion users who can test ?

dbdbdb:
I'm afraid I don't know any other users that have one. I opened the forum thread above to try to find someone with a SpacePilot Pro to help out. Especially windows/mac since I myself can only test in Linux.

The patch above only changes the default values and anyone could get the mouse to work by assigning buttons manually. But it's much easier if it work out of the box like it do for me now with this pull request.

hyarion:
> know any other 3dconnexion users who can test ?

@numo68 I noticed you added the initial support in #4406. Do you still have access to the device and could test this PR?

numo68:
> @numo68 I noticed you added the initial support in #4406. Do you still have access to the device and could test this PR?

I have SpaceMouse Pro, not SpacePilot Pro, so I unfortunately cannot test this.

BSmith-66:
running a Space Pilot Pro on Windows 10  Freecad 0.19
tested this by changing the C:\Program Files\FreeCAD 0.19\data\3Dconnexion3DConnexion.xml 
this now works as expected .. that's 2 people on 2 devices  on 2 operating systems - i would also most call that merge worthy

luzpaz:
@yorikvanhavre maybe merge and if is an issue, we rollback?

yorikvanhavre:
Ok, let's do it then. Thanks everybody!!

PR #4928, opened 2021-07-18
modified ViewProviderPart to show different icon for Assembly-Type

Zolko-123:
This PR modifies the ViewProviderPart, in that  it adds its own explicit getIcon() function, overriding that of ViewProvider. Only the GUI part is changed (ViexProviderPart.h/cpp and icons)

The getIcon() function tests for the (existing) Type property (defined in Part.h) , and if this flag is set to "Assembly" returns another icon than the standard one. Therefore, 2 icons are provided : the standard Geofeaturegroup.svg, that I have reverted to the old (pre-0.19) yellow icon, and a new Geoassembly.svg icon that is the modified icon already used by Assembly4 and is similar to the assembly icon used in other CAD software).

new_icons

The result is the following in the model tree:

Geoassembly

see discussions in the forum:

https://forum.freecadweb.org/viewtopic.php?f=8&t=60308


Zolko-123:
Thanx

luzpaz:
Any documentation we need to update based on this merge?

Zolko-123:
Well, that's the point: anything App::Part (Std_Part) related is still valid as per v0.19, has the same icon. At this point, only Assembly4 takes advantage of this feature, so I'd suggest the answer is : "no". 

Obviously, Assembly4's documentation could need some love, but that's another matter.

luzpaz:
Thanks for the insight. I'll give @realthunder and @kbwbe a heads up about this feature then too

Zolko-123:
the Manipulator WB might be a good candidate too

luzpaz:
@easyw :point_up_2:

PR #4930, opened 2021-07-19
Draft: Some shutil don't have the .which method.

Kawashima-Azumi:
Directly calling subprocess.Popen without checking the presence of the
program is the optimal solution because:
1) Avoids race condition, where the program disappears between the
  check and the Popen call.
2) Popen will raise exception if the program is absent anyway,
  no race condition here.


yorikvanhavre:
Looks good. Thanks!

PR #4931, opened 2021-07-19
Start: Properly hash thumbnail filenames

Kawashima-Azumi:
Image filenames should be encoded as URI before being hashed.

Also assume that filenames are already utf8 because the community has been advertiseing utf8 usage since the beginning of this centry AFAIK.

Calling .encode("utf8") on strings that are already in utf8 simply raises exception:
```
  UnicodeDecodeError: 'ascii' codec can't decode byte ...
for non-ascii (already utf8) strings.
```

It's in fact impossible to precisely determine pathname encoding because different components within the path may have different encodings, e.g. a utf8 directory name followed by an mbcs filename is valid on Linux native filesystems. It's the user's responsibility to keep the iocharset consistent.

### DIFF
--- a/src/Mod/Fem/Gui/Resources/ui/Material.ui
+++ b/src/Mod/Fem/Gui/Resources/ui/Material.ui
@@ -246,7 +246,7 @@
            <double>2.000000000000000</double>
           </property>
           <property name="maximum" stdset="0">
-           <double>2000000000.000000000000000</double>
+           <double>90000000000.000000000000000</double>
           </property>
           <property name="unit" stdset="0">
            <string notr="true">Pa</string>
--- a/src/Mod/Fem/Gui/Resources/ui/Material.ui
+++ b/src/Mod/Fem/Gui/Resources/ui/Material.ui
@@ -246,7 +246,7 @@
            <double>2.000000000000000</double>
           </property>
           <property name="maximum" stdset="0">
-           <double>90000000000.000000000000000</double>
+           <double>2000000000.000000000000000</double>
           </property>
           <property name="unit" stdset="0">
            <string notr="true">Pa</string>
--- a/src/Mod/Start/StartPage/StartPage.py
+++ b/src/Mod/Start/StartPage/StartPage.py
@@ -109,8 +109,8 @@ def getInfo(filename):
             import gnomevfs
         except Exception:
             # alternative method
-            import hashlib
-            fhash = hashlib.md5(("file://"+path).encode("utf8")).hexdigest()
+            import hashlib, urllib
+            fhash = hashlib.md5(urllib.quote("file://"+path,safe=":/")).hexdigest()
             thumb = os.path.join(os.path.expanduser("~"),".thumbnails","normal",fhash+".png")
         else:
             uri = gnomevfs.get_uri_from_local_path(path)


luzpaz:
@Kawashima-Azumi  IMO we shouldn't assume because it's 'common knowledge'. So is there an alternate solution that doesn't raise exceptions that you can propose?

Kawashima-Azumi:
> @Kawashima-Azumi IMO we shouldn't assume because it's 'common knowledge'. So is there an alternate solution that doesn't raise exceptions that you can propose?

No. On Windows possibly yes, but FreeCAD is cross platform, so that's not possible in principle.

luzpaz:
So help me understand, you want to remove a function that makes it possible to load filenames that are utf8 from a program that is used globally by different user demographics ?

Kawashima-Azumi:
> So help me understand, you want to remove a function that makes it possible to load filenames that are utf8 from a program that is used globally by different user demographics ?

On the contrary, I want to remove a function that loads ascii but raises the exception with utf8:
`UnicodeDecodeError: 'ascii' codec can't decode byte 0x?? in position 0: ordinal not in range(128)`
or not only utf8 to be precise, but MBCS including utf8.

The related code does not apply to Windows anyway. So no worry about MS.

On *nix systems, it is often perfectly ok to have different encodings for filenames on even the same filesystem, with a few exceptions (mostly when MS filesystems are mounted). It is up to the user to make it correct by specifying the iocharset or just store filenames in utf8 where iocharset is not applicable.

Kawashima-Azumi:
Well, I've checked the spec and the actual behavior of KDE. The common practice should be:

`hashlib.md5(urllib.quote('file://'+path, safe=':/')).hexdigest()`

I've now updated this pr accordingly.

Kawashima-Azumi:
Well, not quite ...

luzpaz:
Please change the commit summary by prepending 'Start:' and then the most up-to-date revision of what this PR does. Thanks!

luzpaz:
You changed the PR summary but not the commit itself. Here's how to do it from CLI.  
Make sure you're in your branch
```
git commit --amend -m "Start: Properly hash thumbnail filenames"
git push --force
```

Kawashima-Azumi:
Hopefully right this time & Sorry for the delay! Had something else to do and coding is only my hobby.

luzpaz:
"It's done when it's done" is the FreeCAD motto. So all good :wink:

yorikvanhavre:
This looks good to me... I suppose you checked and that it works correctly, @Kawashima-Azumi ? I mean, that the system screenshots are correctly retrieved with that method? (ie. the hash corresponds to the png filename)? Otherwise I can try and test here too

Kawashima-Azumi:
> This looks good to me... I suppose you checked and that it works correctly, @Kawashima-Azumi ? I mean, that the system screenshots are correctly retrieved with that method? (ie. the hash corresponds to the png filename)? Otherwise I can try and test here too

Yes, checked that FreeCAD can load image files' thumbnails produced by KDE accessories and a non-KDE image viewer (geeqie). All their paths match.

Not sure how FreeCAD produce thumbnails itself, so don't know if FreeCAD can load its own thumbnails.

yorikvanhavre:
ok good to go then. thanks!

PR #4950, opened 2021-08-01
App: fix LinkMode enforcement in Link on restore

realthunder:
The patch fixes Link.LinkMode becoming ineffective after restore. It is brought to my attention with thread https://forum.freecadweb.org/viewtopic.php?f=3&t=52772

### DIFF
--- a/src/App/Link.cpp
+++ b/src/App/Link.cpp
@@ -53,7 +53,7 @@ namespace bp = boost::placeholders;
 EXTENSION_PROPERTY_SOURCE(App::LinkBaseExtension, App::DocumentObjectExtension)
 
 LinkBaseExtension::LinkBaseExtension(void)
-    :myOwner(0),enableLabelCache(false),hasOldSubElement(false)
+    :enableLabelCache(false),hasOldSubElement(false)
 {
     initExtensionType(LinkBaseExtension::getExtensionClassTypeId());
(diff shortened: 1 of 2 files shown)


luzpaz:
@realthunder should we consider backporting ?

realthunder:
Yes, it's a minor fix, although not many people are using `LinkGroup`.

chennes (approved):
This compiles, runs, and addresses the issue that @realthunder discusses in the description. To test, I followed Roy_043's steps in the first forums post, but saved the document before removing the items from the link group. Prior to this change, the cylinder object link is retained, and after this change, the cylinder link is removed due to its ownership. This is the intended behavior of the link group in that particular case.

The steps I followed were:

1. Create a Part_Box.
2. Create a Part_Cylinder.
3. Select the box ('Cube') and create a Link to is with Std_LinkMake.
4. Deselect the created Link.
5. Right click in the Tree and select: Link actions > Make link group > Simple group
6. Drag the box Link into the group.
7. Change the Link Mode of the group to 'Auto Unlink'.
8. Drag the cylinder into the group, a Link to the cylinder will be created inside the group.
9. Save the file, close it, and reopen it.
10. Drag the box Link out of the group and onto the document node.
11. Do the same for the cylinder Link.

Result: The box Link is still available, but the Cylinder Link has been erased.

yorikvanhavre:
thanks for the fix @realthunder and thanks for testing @chennes !

PR #4952, opened 2021-08-02
Fix external document loading

realthunder:
Forum thread https://forum.freecadweb.org/viewtopic.php?p=495078#p495078

The problem happens when partial loading is enabled. If document A contains a link to some object in document B, it will load B as partial document with only that object and its necessary dependencies. But if document A contains another link to some object in document C which also has a link to some object in document B, the link in document C may not be restored, because document B is partially loaded without the linked object. This patch will check for this case and reload document B for more objects.

### DIFF
--- a/src/App/Application.cpp
+++ b/src/App/Application.cpp
@@ -55,6 +55,8 @@
 #include <sys/sysctl.h>
 #endif
 
+#include <boost/algorithm/string/predicate.hpp>
+
 #include "Application.h"
 #include "Document.h"
 
@@ -99,6 +101,7 @@
 #include "Document.h"
 #include "DocumentObjectGroup.h"
 #include "DocumentObjectFileIncluded.h"
+#include "DocumentObserver.h"
 #include "InventorObject.h"
 #include "VRMLObject.h"
 #include "Annotation.h"
@@ -487,6 +490,7 @@ bool Application::closeDocument(const char* name)
         setActiveDocument((Document*)0);
     std::unique_ptr<Document> delDoc (pos->second);
     DocMap.erase( pos );
+    DocFileMap.erase(FileInfo(delDoc->FileName.getValue()).filePath());
 
     _objCount = -1;
 
@@ -566,8 +570,10 @@ int Application::addPendingDocument(const char *FileName, const char *objName, b
         return -1;
     assert(FileName && FileName[0]);
     assert(objName && objName[0]);
-    auto ret =  _pendingDocMap.emplace(FileName,std::set<std::string>());
-    ret.first->second.emplace(objName);
+    if(!_docReloadAttempts[FileName].emplace(objName).second)
+        return -1;
+    auto ret =  _pendingDocMap.emplace(FileName,std::vector<std::string>());
+    ret.first->second.push_back(objName);
     if(ret.second) {
         _pendingDocs.push_back(ret.first->first.c_str());
         return 1;
@@ -623,6 +629,41 @@ Document* Application::openDocument(const char * FileName, bool createView) {
     return 0;
 }
 
+Document *Application::getDocumentByPath(const char *path, int checkCanonical) const {
+    if(!path || !path[0])
+        return nullptr;
+    if(DocFileMap.empty()) {
+        for(auto &v : DocMap) {
+            const auto &file = v.second->FileName.getStrValue();
+            if(file.size())
+                DocFileMap[FileInfo(file.c_str()).filePath()] = v.second;
+        }
+    }
+    auto it = DocFileMap.find(FileInfo(path).filePath());
+    if(it != DocFileMap.end())
+        return it->second;
+
+    if (!checkCanonical)
+        return nullptr;
+
+    std::string filepath = FileInfo(path).filePath();
+    QString canonicalPath = QFileInfo(QString::fromUtf8(path)).canonicalFilePath();
+    for (auto &v : DocMap) {
+        QFileInfo fi(QString::fromUtf8(v.second->FileName.getValue()));
+        if (canonicalPath == fi.canonicalFilePath()) {
+            if (checkCanonical == 1)
+                return v.second;
+            bool samePath = (canonicalPath == QString::fromUtf8(filepath.c_str()));
+            FC_WARN("Identical physical path '" << canonicalPath.toUtf8().constData() << "'\n"
+                    << (samePath?"":"  for file '") << (samePath?"":filepath.c_str()) << (samePath?"":"'\n")
+                    << "  with existing document '" << v.second->Label.getValue()
+                    << "' in path: '" << v.second->FileName.getValue() << "'");
+            break;
+        }
+    }
+    return nullptr;
+}
+
 std::vector<Document*> Application::openDocuments(const std::vector<std::string> &filenames,
                                                   const std::vector<std::string> *paths,
                                                   const std::vector<std::string> *labels,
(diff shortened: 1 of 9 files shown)


gbroques:
Thank you @realthunder!!

I'll see if I can find some time to review and verfiy this fix.

src/App/Application.cpp
```diff
@@ -649,118 +691,162 @@ std::vector<Document*> Application::openDocuments(const std::vector<std::string>
...
-        try {
-            _objCount = -1;
-            std::set<std::string> objNames;
-            if (_allowPartial) {
-                auto it = _pendingDocMap.find(name);
-                if (it != _pendingDocMap.end())
-                    objNames.swap(it->second);
-            }
+    int pass = 0;
+    do {
```
chennes, line 701: Why use a `do...` here? Given the exit condition `!_pendingDocs.empty()`, you actively don't want this to guarantee the first iteration given your access to `_pendingDocs.front()`. In addition, the Core Guidelines advise against its use (see http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es75-avoid-do-statements). I suggest switching to a `while...`:

src/App/Application.cpp
```diff
@@ -623,6 +629,41 @@ Document* Application::openDocument(const char * FileName, bool createView) {
...
+    auto it = DocFileMap.find(FileInfo(path).filePath());
+    if(it != DocFileMap.end())
+        return it->second;
+
+    if (checkCanonical == MatchAbsolute)
+        return nullptr;
+
+    std::string filepath = FileInfo(path).filePath();
+    QString canonicalPath = QFileInfo(QString::fromUtf8(path)).canonicalFilePath();
+    for (auto &v : DocMap) {
```
chennes, line 651: It looks like this can be a const (I didn't test it though). This applies to several `auto &` ranged for loops in this PR, so you might want to take a quick look through the rest as well.

src/App/Application.h
```diff
@@ -120,6 +120,33 @@ class AppExport Application
     App::Document* getActiveDocument(void) const;
     /// Retrieve a named document
     App::Document* getDocument(const char *Name) const;
+
+    /// Path matching mode for getDocumentByPath()
+    enum PathMatchMode {
```
chennes, line 125: This can be an enum class (see http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class).

src/App/Document.h
```diff
@@ -74,7 +74,8 @@ class AppExport Document : public App::PropertyContainer
         PartialDoc = 7,
         AllowPartialRecompute = 8, // allow recomputing editing object if SkipRecompute is set
         TempDoc = 9, // Mark as temporary document without prompt for save
-        RestoreError = 10
+        RestoreError = 10,
+        LinkStampChanged = 11, // Indicates during restore time if any linked document's time stamp has changed
```
chennes, line 78: Line ending doesn't match the rest of the file (according to GitHub, anyway).

src/Gui/ViewProviderDocumentObject.cpp
```diff
@@ -196,12 +196,19 @@ void ViewProviderDocumentObject::onChanged(const App::Property* prop)
             // this is undesired behaviour. So, if this change marks the document as
             // modified then it must be be reversed.
             if (!testStatus(Gui::ViewStatus::TouchDocument)) {
-                bool mod = false;
-                if (pcDocument)
-                    mod = pcDocument->isModified();
+                // Note: reverting document modified status like that is not
+                // appropreiate because we can't tell if there is any other
```
chennes, line 200:
```suggestion
                // appropriate because we can't tell if there is any other
```

src/Mod/Part/Gui/ViewProviderExt.cpp
```diff
@@ -472,6 +473,12 @@ void ViewProviderPartExt::onChanged(const App::Property* prop)
         // if the object was invisible and has been changed, recreate the visual
         if (prop == &Visibility && (isUpdateForced() || Visibility.getValue()) && VisualTouched) {
             updateVisual();
+            // updateVisual() may not be triggered by any change (e.g.
+            // triggered by an external object through forceUpdate()). And
+            // since DiffuseColor is not changed here either, do not falsly set
```
chennes, line 478:
```suggestion
            // since DiffuseColor is not changed here either, do not falsely set
```

chennes (changes requested):
I've added a few comments on the code (I have not tested the PR yet, so this is just code inspection stuff).

luzpaz:
@realthunder when you have a free moment, can you respond to the reviewers? thx!

chennes:
No rush on my end, obviously, I'm just slowly working my way through these old PRs in the hopes that we can get them resurrected and merged.

berndhahnebach:

src/App/Application.cpp
realthunder: Well, I think this would be a 'genuine examples where a do-statement is a clear statement of a solution'. `_pendingDocs` may get pushed with new entry at the very end of this very long loop statement. And there is a big block of comments explaining the reason, about 10 lines above the exiting `while()`. Although this is not the only place where `_pendingDocs` is refilled, but I just want emphasize that push which is the solution to the problem that this PR is trying to solve.

chennes: My concern here is that when someone reads this do...while loop, it calls their attention to the fact that you are trying to guarantee at least one execution, regardless of loop end condition. That is contrary to the code itself, however: if _pendingDocs were indeed empty upon entering this loop, you very much do NOT want to execute it once, the statement `_pendingDocs.front()` is almost sure to crash.

I argue that it's much more comfortable for a programmer reading this code to keep the loop end condition in mind as they read, despite the length of the loop. As written the only way to read it is to first scroll to the end, note the end condition, then scroll back to the top. Then you have to think "OK, why do we want to guarantee one execution of this loop, even if the end condition is satisfied?" And then you see line 704... it's very confusing. I don't think you accomplish your stated goal of emphasizing the final push: your large comment block emphasizes that without trouble: using this very non-idiomatic loop construction only muddies the waters.

realthunder: > My concern here is that when someone reads this do...while loop, it calls their attention to the fact that you are trying to guarantee at least one execution, regardless of loop end condition.

I think that's where our difference is. When I see a do {} while loop, I think of it as not worth it to check for loop entry condition because it is already guaranteed. And it is, as you can see with the few lines before the loop. Maybe it would be more obvious if I move the code to populate _pendingDocs closer together?

I see do {} while loop more for the emphasis on exit condition rather than entry, and the second half of this loop (which is newly added by this PR) is dedicated for checking the exit condition, so it would be logical to put the final check at the end.

chennes: We can agree to disagree on this point: every developer is a little different, after all! I don't think this is a deal-breaker for the merge, it's not worth either of us spending more time considering it, IMO.

chennes:
I notice that when I load the original example file from that forums discussion (which now works fine), if I then immediately close the file, it asks if I want to save my changes, even though I haven't knowingly made any. Is that just a consequence of the file structure? Is it expected?

realthunder:
The author probably didn't save the external files in the correct order, that is, save the external documents first before the depending document, which FreeCAD will automatically do when saving the main document. As mentioned in the 'big block of comments' I mentioned earlier, the Link will mark itself as 'touched' if it detects the linked document has changed its `LastModifiedDate` property. If you do a recompute, save the file, close all, and then re-open it, it won't prompt for save again when you close it.

PR #4972, opened 2021-08-17
Revert "Start: Properly hash thumbnail filenames"

chennes:
Reverts FreeCAD/FreeCAD#4931 -- in Python 3 `quote` is now located in the `urllib.parse` module, and not directly in `urllib`. It also takes an optional `encoding` parameter, which I think is necessary in this case since the hash function requires unicode be encoded prior to execution.

### DIFF
--- a/src/Mod/Start/StartPage/StartPage.py
+++ b/src/Mod/Start/StartPage/StartPage.py
@@ -109,8 +109,8 @@ def getInfo(filename):
             import gnomevfs
         except Exception:
             # alternative method
-            import hashlib, urllib
-            fhash = hashlib.md5(urllib.quote("file://"+path,safe=":/")).hexdigest()
+            import hashlib
+            fhash = hashlib.md5(("file://"+path).encode("utf8")).hexdigest()
             thumb = os.path.join(os.path.expanduser("~"),".thumbnails","normal",fhash+".png")
         else:
             uri = gnomevfs.get_uri_from_local_path(path)


yorikvanhavre:
sorry, should have tested better :) it didn't work?

chennes:
No - at least on Windows 10 it has a problem with the Python 2-style use of urllib, and even when I switched to the Python 3 it was having encoding issues. I figured it was best to have the original author look at it again, since I saw some back and forth about doing the encode.

yorikvanhavre:
agree. thanks for checking!

Kawashima-Azumi:
Hi, so should we discuss it here or at #4931 ?

If Python2 must be incompatible, the best is to do it at configure time when running cmake, and not to break it randomly when built and running.

I'd suggest to submit another patch for complete python3-izing instead, if that's the case.

Kawashima-Azumi:
> No - at least on Windows 10 it has a problem with the Python 2-style use of urllib, and even when I switched to the Python 3 it was having encoding issues. I figured it was best to have the original author look at it again, since I saw some back and forth about doing the encode.

Confused about the expression "on Windows 10 it has a problem with the Python 2-style use of urllib". Do you mean Windows 10 = Python3? Or Python2 behaves differently on Windows 10 and Linux?

chennes:
FreeCAD 0.20 does not support Python 2 at all -- removing all of the old Py2 stuff is a work in progress (of course! there was a lot of it!). So there is no need to support Py2, just write the Py3 code, which in this case means using `urllib.parse.quote` rather than `urllib.quote` -- the Py3 `quote` function also takes an optional `encoding` parameter that you can use, but I didn't do any testing on that, I wanted to get HEAD back to a functional state ASAP.

PR #4978, opened 2021-08-18
Gui: Improved tooltips

carlopav:
ref. forum discussion: https://forum.freecadweb.org/viewtopic.php?f=34&t=58747

---

### DIFF
--- a/src/Gui/Command.cpp
+++ b/src/Gui/Command.cpp
@@ -856,8 +856,22 @@ void Command::applyCommandData(const char* context, Action* action)
 {
     action->setText(QCoreApplication::translate(
         context, getMenuText()));
-    action->setToolTip(QCoreApplication::translate(
-        context, getToolTipText()));
+    // build the tooltip
+        QString tooltip;
+        tooltip.append(QString::fromLatin1("<h3>"));
+        tooltip.append(QCoreApplication::translate(
+            context, getMenuText()));
+        tooltip.append(QString::fromLatin1("</h3>"));
+        QRegularExpression re(QString::fromLatin1("([^&])&([^&])"));
+        tooltip.replace(re, QString::fromLatin1("\\1\\2"));
+        tooltip.replace(QString::fromLatin1("&&"), QString::fromLatin1("&"));
+        tooltip.append(QCoreApplication::translate(
+            context, getToolTipText()));
+        tooltip.append(QString::fromLatin1("<br><i>("));
+        tooltip.append(QCoreApplication::translate(
+            context, getWhatsThis()));
+        tooltip.append(QString::fromLatin1(")</i> "));
+    action->setToolTip(tooltip);
     action->setWhatsThis(QCoreApplication::translate(
         context, getWhatsThis()));
     if (sStatusTip)


carlopav:
@chennes here it is :)

src/Gui/Command.cpp
```diff
@@ -856,8 +856,19 @@ void Command::applyCommandData(const char* context, Action* action)
...
+    tooltip.append(QCoreApplication::translate(
+        context, getMenuText()));
+    tooltip.append(QString::fromLatin1("</h3>"));
+    tooltip.append(QCoreApplication::translate(
+        context, getToolTipText())); //TODO: remove hardcoded shortcuts?
+    tooltip.append(QString::fromLatin1("<br><i>("));
+    tooltip.append(QCoreApplication::translate(
+        context, getWhatsThis()));
+    tooltip.append(QString::fromLatin1(")</i>"));
+    tooltip.remove(QString::fromLatin1("&"));
```
yorikvanhavre, line 870: I wonder if this is safe... The text could contain a & character that is intended... Maybe we could remove only the & chars that are not surrounded by spaces?

carlopav: thx for the quick review! Hmm, I guess you are right, also if at the moment no command listed in the wiki contain the "&" characted.
After a bit of tests I noticed that line 373 of Command.cpp also perform the purge of "&" character when command is triggered:
`boost::replace_all(displayText,"&","");`

So I guess that "&" should not be allowed inside the command name string.
What do you think?

chennes: I thought (assumed?) the command name was required to be a valid C++/Python identifier, but maybe that's not true. Under what circumstances is there an ampersand to remove? I assume that line was added to address a specific problem that someone encountered.

carlopav (commented):
`QCoreApplication::translate(context, getMenuText())` almost always return a string with an ampersand character inside... dont know why...

src/Gui/Command.cpp
carlopav: `QCoreApplication::translate(context, getMenuText())` almost always return a string with an ampersand character inside... dont know why...

yorikvanhavre: Command names should not have amperstand at all indeed. I don't think it's a formal requirement (up to now it is just a string), but it seems like a safe requirement to enforce, you never know what could be the future. But menu texts often have one. It's QT's way to define an underlined character available via an ALT keyboard shortcut. And translators might decide to put it on another letter. So in menutext it's very likely to be some, in bothethe original and the translated strings.

chennes: Do you know if Qt provides an escape character to allow a literal ampersand? If not, then I think this code should be safe as is, and if so, then we need a more sophisticated search and replace that does that escape action (maybe Qt provides such a function somewhere?).

carlopav: I have not the skill to assess if this is safe or not, but thanks for the help guys. :)

I also tried this, but it doesn't work... :-S
`    int i = tooltip.indexOf(QString::fromLatin1("&"));
    while (tooltip.contains(QString::fromLatin1("&"))) {
        if ((tooltip[i + 1] != QString::fromLatin1(" ")) && (tooltip[i - 1] != QString::fromLatin1(" "))) {
            tooltip.remove(i);
        }
        i = tooltip.indexOf(QString::fromLatin1("&", i));
        if (i == -1) { break; }
    }`

I guess I have some problems with `tooltip.remove(i);`

chennes: Yeah, I definitely wouldn't start playing around with indexing directly, you're asking for a segfault if the ampersand is the first or last character. If we need to actually support ampersands in the string, I think you'll want a regex that only matches ampersands that are followed by another character. Let me dig around in Qt for a minute here and see if it's really needed.

chennes: OK, found it: it is possible to "escape" an ampersand by using a double ampersand (that is, to display an actual ampersand in the text you actually enter "&&"). So a regex that detects an ampersand followed by any character that's NOT an ampersand should be sufficient.

chennes: So, I guess something like (tested the regex, but not the code):
```
QRegularExpression re("[^&]&[^&]");
tooltip.replace(re, QString());
tooltip.replace("&&","&");
```

carlopav: @chennes almost works, the point is that if I replace the regular expression with an empty string, it also replace the characters before and after the "&"...

chennes: Oh, right -- we need to use capturing groups. Something like (untested):
```
QRegularExpression re(QString::fromLatin1("([^&])&([^&])"));
tooltip.replace(re, QString::fromLatin1("\\1\\2"));
tooltip.replace(QString::fromLatin1("&&"), QString::fromLatin1("&"));
```

carlopav: this appears to work:
```
    QString tooltip;
    tooltip.append(QString::fromLatin1("<h3>"));
    tooltip.append(QCoreApplication::translate(
        context, getMenuText()));
    tooltip.append(QString::fromLatin1("</h3>"));
    tooltip.append(QCoreApplication::translate(
        context, getToolTipText())); //TODO: remove hardcoded shortcuts?
    tooltip.append(QString::fromLatin1("<br><i>("));
    tooltip.append(QCoreApplication::translate(
        context, getWhatsThis()));
    tooltip.append(QString::fromLatin1(")</i>"));
    QRegularExpression re(QString::fromLatin1("[^&]&[^&]"));
    QRegularExpressionMatch match = re.match(tooltip);
    if (match.hasMatch()) {
        tooltip.remove(match.capturedStart()+1, 1);
    }
    tooltip.replace(QString::fromLatin1("&&"), QString::fromLatin1("&"));
```
It removes the damn "&" character and then replaces the "&&". It can only remove one lonely "&" from the QString, but I assume that we shouldn't have more...

chennes: Did the capture groups code I posted give an error? I can do some debugging on it tomorrow.

carlopav: You are too kind Chris. :)
It worked, but It didn't give the expected result: the ([^&])&([^&]) was replaced with "/1/2" or something similari :(

chennes: Strange -- I just tested and it's working on my machine. Did you copy and paste the code from above?

carlopav: @chennes that's embarassing :-S  it works perfectly... dont' know what i did wrong yesterday evening :)

yorikvanhavre:
You guys are the best :)

carlopav:
:) thanks @chennes, I learned more than a couple of things!

src/Gui/Command.cpp
carlopav: @chennes reading One of your posts in the forum, i have the feat I had to declare QString and QRegularExpression at the top of the file... Did I?

chennes: QString is included by many other Qt include files, so I'm sure that QMessageBox (at least) is already including it. QRegularExpression might be more problematic -- and without any CI running right now, I don't know if anyone on Linux tested this. I only tested on Windows.

carlopav: I didn't. Ok, lets include them just in case... Tomorrow i'll submit a fix.

carlopav: https://github.com/FreeCAD/FreeCAD/pull/4988

luzpaz:
@carlopav would you mind making a quick animated gif of the feature and adding it to the [v0.20 Release page on the wiki](https://wiki.freecadweb.org/Release_notes_0.20)?

carlopav:
@luzpaz done!

PR #4981, opened 2021-08-18
[TD] Prevent unused parameter warning

0penBrain:

### DIFF
--- a/src/Mod/TechDraw/App/Cosmetic.cpp
+++ b/src/Mod/TechDraw/App/Cosmetic.cpp
@@ -1004,6 +1004,7 @@ std::pair<Base::Vector3d, Base::Vector3d> CenterLine::calcEndPoints2Lines(DrawVi
                                                       
 {
 //    Base::Console().Message("CL::calc2Lines() - mode: %d flip: %d edgeNames: %d\n", mode, flip, edgeNames.size());
+    (void)flip;
     std::pair<Base::Vector3d, Base::Vector3d> result;
     if (edgeNames.empty()) {
         Base::Console().Warning("CL::calcEndPoints2Lines - no edges!\n");


chennes:
Is there a reason this isn't getting used? There's a whole set of conditionals and operations in this function doing transforms, is flip being skipped for a reason?

0penBrain:
@chennes honestly I don't really know. :) I saw that in this function, determining the direction is fine with a so-called 'circulation test'. So I guess 'flip' is of no use. However I'm not a TechDraw expert and I basically just polished things. ;)

chennes:
@donovaly Can you take a look at this? I'm hesitant to just blindly axe this warning, when it really seems to be indicating a potential error (a missing implementation of flip, which it looks like *might* be getting referenced in TaskCenterline (I'm looking at TaskCenterline.cpp line 420).

yorikvanhavre:
I have no idea either :( In any case, if we're only removing it for cleanliness, I'd rather comment it out and add a note...

0penBrain:
@yorikvanhavre we can't just remove it as it is an argument. If we remove it we will break the function interface.

chennes:
I think @donovaly is our de facto TD expert now that @WandererFan is gone -- he's the last one to touch code related to this, so I think we can just wait until he has time to weigh in. I'm concerned that this is one of those times when the static analyzer may have actually identified a genuine problem and shouldn't just be shut up.

AjinkyaDahale:
This can probably be closed. @wwmayer just made this exact change recently in https://github.com/FreeCAD/FreeCAD/commit/378dec720c808f5b05216ed5e890e6d49bdc999a.

0penBrain:
Indeed. Closing.

PR #4990, opened 2021-08-22
[Sketcher] Refactor split code

chennes:
LGTM complains about using `continue` statements inside a loop whose condition is always false. In addition, the C++ core guidelines recommend against using the `do...while` construct, and in this case it was really serving as a `goto`, just hiding the actual `goto` keyword. This PR replaces the loop and continue structure with simple conditionals.

I should note that the commit looks like a much larger change than it is due to the indentation change when removing the loop... I really only replaced the conditionals with the continue statements.

---

- [X]  Discussed with other developers: https://forum.freecadweb.org/viewtopic.php?f=19&t=61420
- [X]  All Sketcher unit tests are confirmed to pass
- [X]  No tracker ticket

### DIFF
--- a/src/Mod/Sketcher/App/SketchObject.cpp
+++ b/src/Mod/Sketcher/App/SketchObject.cpp
@@ -2982,156 +2982,149 @@ int SketchObject::split(int GeoId, const Base::Vector3d &point)
     std::vector<Part::Geometry *> newGeometries;
     std::vector<int> newIds;
     std::vector<Constraint *> newConstraints;
-    bool ok = false;
 
     Base::Vector3d startPoint, endPoint, splitPoint;
     double radius, startAngle, endAngle, splitAngle;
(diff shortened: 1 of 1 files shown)


chennes:
@pavltom @0penBrain could you double-check me on this? I think there is no functional change here, and this should silence the LGTM warning. Thanks!

0penBrain (approved):
@chennes I didn't compile or run but only do a code review.
It's correct to me, behavior is strictly preserved without any change.
For sake of curiosity, did you try to replace 'continue' with 'break' to see if LGTM still complains ?

abdullahtahiriyo:
Personally I do not like do {} while(false) structures, because they do not show intent (or worse show the intent of a loop when no loop is intended). But I also know that some programmers like them. I think that what is proposed is clearer to understand and better. Thank you. Merged.

PR #4996, opened 2021-08-26
[Core] (Partially?) Fix data loss on dir rename

AjinkyaDahale:

- [ ]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones: Edits in `Base` may be skipped for now and added from PR #4791. Edit in `App` addresses the same issue, but I can make a new PR if needed.

---

### DIFF
--- a/src/Gui/MainWindow.h
+++ b/src/Gui/MainWindow.h
@@ -77,6 +77,13 @@ class GuiExport MainWindow : public QMainWindow
     Q_OBJECT
 
 public:
+    enum ConfirmSaveResult {
+        Cancel = 0,
+        Save,
+        SaveAll,
+        Discard,
+        DiscardAll
+    };
     /**
      * Constructs an empty main window. For default \a parent is 0, as there usually is
      * no toplevel window there.
--- a/src/App/Document.cpp
+++ b/src/App/Document.cpp
@@ -66,6 +66,7 @@ recompute path. Also, it enables more complicated dependencies beyond trees.
 # include <climits>
 # include <bitset>
 # include <random>
+# include <boost/filesystem.hpp>
 #endif
 
 #include <boost/algorithm/string.hpp>
@@ -141,6 +142,8 @@ using namespace zipios;
 #if FC_DEBUG
 #  define FC_LOGFEATUREUPDATE
 #endif
+
+namespace fs = boost::filesystem;
 
 // typedef boost::property<boost::vertex_root_t, DocumentObject* > VertexProperty;
 typedef boost::adjacency_list <
@@ -2610,6 +2613,8 @@ bool Document::saveToFile(const char* filename) const
         fn += uuid;
     }
     Base::FileInfo tmp(fn);
+    // In case some folders in the path do not exist
+    fs::create_directories(fs::path(filename).parent_path());
 
     // open extra scope to close ZipWriter properly
     {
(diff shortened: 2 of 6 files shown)


AjinkyaDahale:
Thanks to @chennes, learned that `boost::filesystem` is available to create the parent directories. Pushed a commit that makes the parent folder while saving.

src/Gui/MainWindow.h
```diff
@@ -77,6 +77,13 @@ class GuiExport MainWindow : public QMainWindow
     Q_OBJECT
 
 public:
+    enum ConfirmSaveResult {
```
chennes, line 80: I think you can make this an `enum class` with no other code changes (since you already specify the full scope when using it).

AjinkyaDahale: I tried doing it but there were some complications because this is also a Qt slot.

chennes:
OK, this is looking good now, I've tested it on Windows and Linux and it does a great job of stopping accidental data loss due to failed saves.

luzpaz:
Awesome! Lets backport this? Pretty please?!

berberic2:
This commit breaks saving of files that are loaded via parameters without path. P.e.

freecad X.FCStd

luzpaz:
Heads up @AjinkyaDahale :point_up:

AjinkyaDahale:
> This commit breaks saving of files that are loaded via parameters without path. P.e.
> 
> freecad X.FCStd
> 
> see: https://forum.freecadweb.org/viewtopic.php?f=13&t=62319&p=534156#p534156

Sounds like we need to make sure whenever files are loaded their filenames are saved with absolute value, or at least the paths resolved when files are saved.

AjinkyaDahale:
So best I could get for now: files when loaded by `freecad xxxx.FCStd` only store the filename by relative path. The loading happens here: https://github.com/FreeCAD/FreeCAD/blob/2fd71242ba4a302aaeb70bb5dce4e9652eee5e44/src/App/Application.cpp#L791. I believe somewhere in that initialization we should make the file path absolute. I'm hesitating in just putting a `boost::filesystem::absolute` or similar here because there is a parameter used in `App::Application::openDocument()` called `paths`, and that could mess with a few things.

donovaly:
Is this in combination with PR #5045 safe to be backported to 0.19.3 in your opinion?

AjinkyaDahale:
> Is this in combination with PR #5045 safe to be backported to 0.19.3 in your opinion?

I don't really dabble with backporting, but I can't see why not. That said, there are other commits that fixed the issue these changes intended to address (can't list them off the top of my head but will update if I dig them out). So these changes mostly are now features we may not want to handle in backporting.

donovaly:
Thanks for the info. I came here since your PRs were the last 2 undecided backport requests according to our list: https://forum.freecadweb.org/viewtopic.php?f=10&t=57659#p495659

So since you say these are mostly new features a backport is not necessary.

AjinkyaDahale:
> Thanks for the info. I came here since your PRs were the last 2 undecided backport requests according to our list: https://forum.freecadweb.org/viewtopic.php?f=10&t=57659#p495659
> 
> So since you say these are mostly new features a backport is not necessary.

Oh yes. @chennes mentioned them in that thread: https://forum.freecadweb.org/viewtopic.php?f=10&t=57659&p=534632#p534725.

PR #5016, opened 2021-09-07
[TechDraw] prevent crash where user has duplicated page without depen…

mwganson:
…dencies and then tries to remove a view from one of the pages by editing the Views property

This PR prevents the crash where the user edits the Views property of a TechDraw page to remove a view that also exists in another page.

Note: this also provides a means for a user who has duplicated a page without also duplicating the views, dimensions, template, etc. to delete the duplicate page without also deleting all of its content (which also deletes it from all duplicate pages).  The way to do this is to edit the Views property of the page to remove the views from that page before deleting that page.  Ditto for the Template property.

Let's say the original document has:

```
Page
    ->Template
    -> View
        -> Dimension
```
A duplicate is made without dependencies, so now you also have:
```
Page001
    ->Template
    -> View
        -> Dimension
```
Prior to this PR if you edit either page's Views property to remove the View or the Dimension or both, you get a crash.  After the PR if a user wishes to remove Dimension from Page or Page001 and still retain it in the other page he can do so by editing the Views property.

Normally, the way you would remove a Dimension is to simply delete it, but if you do this and it also exists in a duplicate page it gets removed from both pages.  This would be consistent with behavior in other workbenches.  For example, if you have a Cut object made from a Cylinder and a Cube and you duplicate the Cut without dependencies, and then you delete the Cube form the Cut001 object you also delete it from the original Cut and indeed the Cube is deleted from the document.  But the difference is if you delete Cut001 made in this manner it does not also delete Cube and Cylinder, which doesn't break original Cut object.  In TechDraw if you delete Page001 in the above example it also deletes Template, View, and Dimension from the document, breaking Page (or at least deleting all of its content, which might be the desired behavior in some cases).  IMO, deleting Page001 should not also delete all of its content if that content also exists in another page, but I haven't worked out yet how to accomplish this.  Right now, I'm satisfied to just prevent a crash.

---

### DIFF
--- a/src/Mod/TechDraw/Gui/MDIViewPage.cpp
+++ b/src/Mod/TechDraw/Gui/MDIViewPage.cpp
@@ -485,7 +485,7 @@ void MDIViewPage::fixOrphans(bool force)
              * if the user duplicated the page without duplicating dependencies
              */
             int numParentPages = qv->getViewObject()->countParentPages();
-            if (thisPage != pp && numParentPages == 1) {
+            if (thisPage != pp && numParentPages == 0) {
                m_view->removeQView(qv);
             }
         }


yorikvanhavre:
I trust you fully here as you are deep looking at the problem for some time ;) Thanks for that!

luzpaz:
@mwganson please mention on the [0.20 Release Page](https://wiki.freecadweb.org/Release_notes_0.20)

mwganson:
Thanks, Yorik.  I guess we can always revert it if there's a problem.  Too bad WandererFan isn't active.  I'm sure he'd know exactly what to do with this.

yorikvanhavre:
Yes, indeed :/ /but Wandererfan himself inherited that WB from its original author ( @mrlukeparry ) and had to learn it from scratch too... We can do it too! ;)

PR #5019, opened 2021-09-09
Gui: Language/Translator.cpp remove extra .ko translation + alphabetize

luzpaz:
ref: https://forum.freecadweb.org/viewtopic.php?p=531658#p531658


luzpaz:
@chennes I haven't tested but in your opinion does this change the order of the languages in the Preferences UI? (which is not my intention)

chennes:
No, this is a map, its "order" is not set by the order the elements are created, and I think that UI element is sorted totally separately anyway.

yorikvanhavre:
In any case, that section of Translator.cpp is now updated automatically by the crowdin script as soon as a new language reaches 25% (at the end of the list), so that order might get changed anyway...
Apparently the list in the UI is sorted not by the english name of the language, but by its translation
For completeness, I wouldn't remove a language entry there without also removing the ts and qm files, then. Better to keep everything in sync. I asked myself too, should we remove languages that have fallen below 25%, but it looks like a punitive action, and I thought maybe best to leave them, after all it doesn't really harm...

FreeCAD-Tools:
> In any case, that section of Translator.cpp is now updated automatically by the crowdin script as soon as a new language reaches 25% (at the end of the list), so that order might get changed anyway...
> Apparently the list in the UI is sorted not by the english name of the language, but by its translation

What script are we talking about? Maybe just add there: 

`lang_dict = dict(sorted(lang_dict.items ()))`

PR #5020, opened 2021-09-09
[Gui] Edit/Placement not active when no object selected

0penBrain:

### DIFF
--- a/src/Gui/CommandDoc.cpp
+++ b/src/Gui/CommandDoc.cpp
@@ -1574,7 +1574,7 @@ void StdCmdPlacement::activated(int iMsg)
 
 bool StdCmdPlacement::isActive(void)
 {
-    return (Gui::Control().activeDialog()==0);
+    return Gui::Selection().countObjectsOfType(App::GeoFeature::getClassTypeId()) == 1;
 }
 
 //===========================================================================


src/Gui/CommandDoc.cpp
```diff
@@ -1574,7 +1574,7 @@ void StdCmdPlacement::activated(int iMsg)
 
 bool StdCmdPlacement::isActive(void)
 {
-    return (Gui::Control().activeDialog()==0);
+    return (Selection().getCompleteSelection().size() > 0) || (Gui::Control().activeDialog() != 0);
```
yorikvanhavre, line 1577: So the command is active if (selection > 0 OR active dialog open)... shouldn't it be (selection > 0 AND no active dialog)?

0penBrain: Indeed I was too fast on this and this is inappropriate. Fortunately you saw it.
Should be better now. I applied same rule as Std_TransformManip.
There could be still improved filtering on this, but it already brings some improvement and consistency.

0penBrain: @yorikvanhavre Actually last time I messed up my git commands and it was not fixed. Now is OK (for sure). :)

yorikvanhavre:
Thanks for the fix! This lets the command active even if another task dialog is running... Did you check if that's OK? (The Edit/Placement command is able to close the former task dialog)?

0penBrain:
@yorikvanhavre yes that works, it closes the actual dialog and opens the new one. Anyway this is a good remark. Is that still what we really want? Or do you think it's better if the command is disabled if a dialog is already open?

yorikvanhavre:
Ok, thanks for checking! I have no strong opinion on this, I was just thinking of staying in line with what was currently implemented, there might have been a reason why the IsActive method was made that way... But it might as well have simply been copy/pasted from another command.

PR #5026, opened 2021-09-12
Crowdin: Expose View -> 'Standard views' -> 'Axonometric' to translation

luzpaz:
Closes https://github.com/FreeCAD/FreeCAD-translations/issues/63

### DIFF
--- a/src/Gui/Workbench.cpp
+++ b/src/Gui/Workbench.cpp
@@ -484,6 +484,7 @@ std::list<std::string> Workbench::listCommandbars() const
     qApp->translate("Workbench", "&File");
     qApp->translate("Workbench", "&Edit");
     qApp->translate("Workbench", "Standard views");
+    qApp->translate("Workbench", "Axonometric");
     qApp->translate("Workbench", "&Stereo");
     qApp->translate("Workbench", "&Zoom");
     qApp->translate("Workbench", "Visibility");


luzpaz:
Note: I haven't tested if this fix in fact works

marioalexis84:
that change will not be compiled because it is inside the block `#if 0 ... #endif.`
https://stackoverflow.com/questions/2852913/what-exactly-does-an-if-0-endif-block-do

chennes:
I don't think it's supposed to be compiled -- it's one of those weird machinations we go through to get Qt's translation facilities to produce the strings we need. This change doesn't affect the compiled executable, just the output of `lupdate` (which it turns out ignores preprocessor directives).

yorikvanhavre:
Using Qt_TRANSLATE_NOOP doesn't do anything at compile time (Normally, we use translate() or tr()). Qt_TRANSLATE_NOOP allows the text to be translatable (picked up by lupdate), so its translation is available to FreeCAD, but the text is not translated immediately, but by another mechanism of FreeCAD somewhere else. This is mostly used in places where you still want to keep the english string (for example stuff that needs to be written to a FCStdc file), and FreeCAD will translate it itself, when needed.

luzpaz:
@yorikvanhavre where can we add these tips for devs/'interested users' to learn about ?

yorikvanhavre:
Maybe somewhere here? https://wiki.freecadweb.org/Localisation#Development_-_How_to_Add_Localisation there might need a new paragraph to explain all the Qt translation structures... I'll try to start something

luzpaz:
> Maybe somewhere here? https://wiki.freecadweb.org/Localisation#Development_-_How_to_Add_Localisation there might need a new paragraph to explain all the Qt translation structures... I'll try to start something

Remember you did add a lot of info about translations to https://wiki.freecadweb.org/Translating_an_external_workbench

PR #5030, opened 2021-09-14
[Spreadsheet] Prevent infinite loop when copy/paste of non-rectangula…

mwganson:
…r selection

https://forum.freecadweb.org/viewtopic.php?f=3&t=62207&start=10

---

### DIFF
--- a/src/Mod/Spreadsheet/App/PropertySheet.cpp
+++ b/src/Mod/Spreadsheet/App/PropertySheet.cpp
@@ -391,14 +391,16 @@ void PropertySheet::pasteCells(XMLReader &reader, const CellAddress &addr) {
                 roffset = addr.row() - from.row();
                 coffset = addr.col() - from.col();
             }else
-                range.next();
+                if (!range.next())
+                    break;
             while(src!=*range) {
                 CellAddress dst(*range);
                 dst.setRow(dst.row()+roffset);
                 dst.setCol(dst.col()+coffset);
                 owner->clear(dst);
                 owner->cellUpdated(dst);
-                range.next();
+                if (!range.next())
+                    break;
             }
             CellAddress dst(src.row()+roffset, src.col()+coffset);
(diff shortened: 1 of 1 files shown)


chennes:
Isn't this equivalent to removing the single calls to `range.next()` and shifting them into the while loop?
```
while(range.next() && src!=*range) {
```

0penBrain:
@mwganson Indeed that would not be the same. Could you add the curly braces for the 'else' statement ? Despite the indentation, I find it hard to read. ;)

chennes:
> Could you add the curly braces for the 'else' statement ? Despite the indentation, I find it hard to read. ;)

Oh good grief, you're right, I was reading it completely wrong! +1 on the curly braces.

mwganson:
I don't think it is good practice to mix this usage in the same document.  I personally prefer using the curly braces in all cases, but if the previous code doesn't use them then I don't.

chennes:
IMO the ultimate criteria is to write the clearest code possible -- "style" should be secondary. That said, I'm happy to merge as-is and then add the curlies in a separate commit so that later readers know who to blame for the style faux pas 😄. No sense holding things up over this.

mwganson:
There was a forum discussion on this, but I don't have a link to it.  It was generally agreed not to mix them on the same page or at least not in the same block.  Consider adding the curly braces everywhere if they are to be added anywhere, or at least at a minimum for code in the same function/method.

chennes:
https://wiki.qt.io/Qt_Coding_Style is often suggested: in this particular case, by my read it recommends (but does not require) adding the curly braces to the "else if", since the "if" has them. Then again, it also supports the `} else if () {` on a single line, which we typically don't do. YMMV...

donovaly:
I think this should be backported to FC 0.19 since this fixes a crash. I've put it now to our list: https://forum.freecadweb.org/viewtopic.php?f=10&t=57659

PR #5037, opened 2021-09-17
[Tools] Refactor updatets to support two-lang mods

chennes:
The script to extract translation strings from source code files was designed to skip the C++ translations of directories with Python. In some cases this was manually worked around, but in others it was not. This PR is an attempt to refactor that code to ensure that both C++ and Python translations are included whenever a Mod has both languages in it.

---

- [X]  "Small" bug
- [X]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) e
- [X]  No ticket


chennes:
@yorikvanhavre and @luzpaz can you have a look at this PR and see if this actually works to get the necessary translation strings?

yorikvanhavre:
Yeah, that script evolved a lot over time... Good cleaning! Maybe we could simply remove the pyfolders list completely and let the script run both C++ and py gathering in all Mod folders? That way nobody has to worry about that anymore...

chennes:
If that doesn't break anything, then yeah, I think that's the best solution. I figured there was some reason we were so carefully avoiding running both lupdate and pylupdate on the same directory, but obviously we have some where we have to now.

chennes:
@yorikvanhavre OK, I just force-pushed another pass at refactoring the code. Because of the variability in the directory structures, I still ended up explicitly listing out all of the directories that need to be scanned, but at least now if a new one isn't being done, it will be obvious why! I run both lupdate and pylupdate on all of them, and just combine the results of each. I don't have a great way of testing whether this really worked, though. The TS files are generated, but I don't know what sort of havoc this will wreak on CrowdIn.

luzpaz:
@chennes we need to add `src/Mod/Plot` as well, see https://github.com/FreeCAD/FreeCAD-translations/issues/60
 
I've started on Plot here, but it's not complete yet https://github.com/FreeCAD/FreeCAD/pull/5042

chennes:
Plot is omitted here because it doesn't have a Resources/translations directory. That is probably most appropriately added in #5042.

berndhahnebach:
would you rebase the PR to make it run on CI?

berndhahnebach:

FreeCAD-Tools:
The Crowdin build needs to be updated when this script is merged. For get new strings in FEM and other workbenches.

berndhahnebach:
FreeCAD CI-repository:
branch:
https://gitlab.com/berndhahnebach/FreeCAD/-/commits/PR_5037

pipeline:
https://gitlab.com/berndhahnebach/FreeCAD/-/pipelines/376765024

status:
success

yorikvanhavre (approved):
Ok I'll merge and do a crowdin update and we compare... It is always possible to revert if anything goes wrong. For reference, here is the situation before the change
Screenshot from 2021-09-29 14-17-55
:

yorikvanhavre:
Ok this seems consequent. This is after the merge:
Screenshot from 2021-09-29 14-28-52

yorikvanhavre:
Updated files in 77925c428e Everything seems good!
Your refactored script does not update the main src/Gui/translations/FreeCAD.ts file or am I mistaken?

chennes:
It does: that's the very first entry in the directories list, [line 56](https://github.com/FreeCAD/FreeCAD/blob/77925c428e82f3e2e76d0e4907ac06e4bd475134/src/Tools/updatets.py#L56).

yorikvanhavre:
Ok! I was fooled because the script didn't change a line of the ts file, but that's good :)

FreeCAD-Tools:
@yorikvanhavre why new strings not added to FEM workbench?

PR #5049, opened 2021-09-20
[Spreadsheet] Refactor and simplify paste code

chennes:
PR #5030 from @mwganson fixed a bug in Spreadsheet, resolving an infinite loop with pasting disconnected blocks of cells. Fixing this error revealed a further error that the paste algorithm missed pasting some cells. This PR refactors and simplifies the paste algorithm to ensure that, for example "cellUpdated()" is only called once for each cell, and that all of the necessary cells are pasted in.

---

- [X]  Bug fix
- [X]  All Spreadsheet unit tests are confirmed to pass
- [X]  No tracker ticket


chennes:
@mwganson When testing the new code I found that the last cell was being skipped when pasting multiple disconnected blocks. In trying to understand the old code, it seemed to me that cells were being visited and re-written multiple times by the old algorithm, which I guess was caused by a sort of disconnected series of bug fixes over time. I've basically dropped it back down to the simplest two loops, but I'm concerned that I may not be testing for all edge cases that had resulted in the previous algorithm. Can you give it a look?

hyarion:
@chennes I haven’t tested it but it almost  looks a bit too simple now! 👍

This feels like a good opportunity to look at the unit tests and see if anything seems to be missing. It might also be a good idea to write a test for the bug you found in the old version.

mwganson:
I conducted this test:

Using the original test file from https://forum.freecadweb.org/viewtopic.php?f=3&t=62207#p532911 I selected these cells:

A1, C1, B2, A3, C3

Snip macro screenshot-554fd5

Edit menu -> copy

Select A7

Edit menu -> paste

Result:
Snip macro screenshot-e5f6f8

Results of same test running 0.20 on windows:

OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.25645 (Git)
Build type: Release
Branch: master
Hash: 37d9757399b4c2ec30318eb88d7cd7c508246345
Python version: 3.8.6+
Qt version: 5.15.2
Coin version: 4.0.1
OCC version: 7.5.0


Snip macro screenshot-e40300

Clearly, as you found, there were already issues with copy/paste when using non-rectangular selection of cells.  Admittedly, I did not spend much time debugging or studying the original code.  My goal was to leave the algorithm as is and fix the infinite loop because I thought it was working correctly except for the infinite loop issue.

I also don't think in python there is a good way to determine which cells the user has selected.  In my EasyAlias macro I resort to using QtGui.QTableView.selectedIndexes()

```
    mw=FreeCADGui.getMainWindow()
    mdiarea=mw.findChild(QtGui.QMdiArea)
    subw=mdiarea.subWindowList()
    widgets = []
    for i in subw:
        if i.widget().metaObject().className() == "SpreadsheetGui::SheetView":
            widgets.append(i.widget())
    if len(widgets) > 1:
        FreeCAD.Console.PrintError("Having more than one spreadsheet view open at a time can confuse the macro.  Close the other sheet views and try again\n")
        return None
    elif len(widgets) == 1:
        return widgets[0].findChild(QtGui.QTableView).selectedIndexes()
    return None
```

chennes:
Oh, nice, I did't even think to test blank cells. I believe that error is actually in the "copyCells" code, just above this function: it's actually checking to see if there's a value in the cell and using that to determine whether to increment cell count. So for every range you have selected that has no content, one cell will get lost.

chennes:
@mwganson I think I figured out a way to make it work, I've pushed a new commit that no longer skips empty cells.

berndhahnebach:
Would you rebase, to make CI pass https://gitlab.com/berndhahnebach/FreeCAD/-/jobs/1618664760

berndhahnebach:

berndhahnebach:
FreeCAD CI-repository:
branch:
https://gitlab.com/berndhahnebach/FreeCAD/-/commits/PR_5049

pipeline:
https://gitlab.com/berndhahnebach/FreeCAD/-/pipelines/376765028

status:
success

chennes:
@donovaly would you mind testing this one? I'm worried there is some edge case that I've missed thinking about.

donovaly (commented):
The PR works as it should, the code is understandable.

donovaly:
> @donovaly would you mind testing this one? I'm worried there is some edge case that I've missed thinking about.

The PR is OK at it is. many thanks.

For a future PR, we should add a new dialog that warns you when you overwrite cells by pasting.
- take for example a spreadsheet in which cells 1,1 to 3,3 contain content
- now select 2 consecutive empty cells and copy them
- paste into e.g. cell 1,2
-> result: you overwrite a non-empty cell without any warning

--------------------

I think PR #5030 should be backported, and maybe this PR here too.

chennes:
Thanks for testing, let's get this into others' hands and see what breaks 😄 .

PR #5054, opened 2021-09-23
[BUILDFIX] Add missing includes

0penBrain:
@wwmayer could you please have a look at this as build is currently broken at least on some configurations ? Thx


berndhahnebach:
https://forum.freecadweb.org/viewtopic.php?f=10&t=62437

PR #5059, opened 2021-09-24
[Spreadsheet Import/Export] Add 3 new user parameters: ImportExportDe…

mwganson:
…limiter, ImportExportQuoteCharacter, and ImportExportEscapeCharacter

Currently, prior to this PR, users importing or exporting in the Spreadsheet workbench using the Gui could only export or import using tabs as the delimiter.  Quote character and escape character were also hard coded to " and \.

Now users will be able to edit these parameters, if desired.  The defaults are the same as what is currently being used.

ImportExportDelimiter accepts "tab", "\t", "comma", and "semicolon" as recognized tokens or the user may enter a single character, such as the comma (,), the semicolon(;), the colon (:), the bar (|), etc.  The other 2 parameters must be a single character, such as the double quote ("), or the single quote ('), the slash (/) or the backslash(\).

---


berndhahnebach:
FYI: CI passsed https://gitlab.com/berndhahnebach/FreeCAD/-/pipelines/376489236

berndhahnebach:

src/Mod/Spreadsheet/Gui/Command.cpp
```diff
@@ -197,9 +197,36 @@ void CmdSpreadsheetImport::activated(int iMsg)
     if (!fileName.isEmpty()) {
         std::string FeatName = getUniqueObjectName("Spreadsheet");
         Sheet * sheet = freecad_dynamic_cast<Sheet>(App::GetApplication().getActiveDocument()->addObject("Spreadsheet::Sheet", FeatName.c_str()));
+        ParameterGrp::handle group = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Mod/Spreadsheet");
+        QString delimiter = QString::fromStdString(group->GetASCII("ImportExportDelimiter","tab"));
+        QString quoteChar = QString::fromStdString(group->GetASCII("ImportExportQuoteCharacter","\""));
+        QString escapeChar = QString::fromStdString(group->GetASCII("ImportExportEscapeCharacter","\\"));
+        group->SetASCII("ImportExportDelimiter", Base::Tools::toStdString(delimiter).c_str()); //create parameters that do not already exist for user convenience
+        group->SetASCII("ImportExportQuoteCharacter", Base::Tools::toStdString(quoteChar).c_str());
+        group->SetASCII("ImportExportEscapeCharacter", Base::Tools::toStdString(escapeChar).c_str());
```
0penBrain, line 206: I don't like much this kind of code that is useful only once but executed every time.
To me, either it's only a parameter then document it in the Fine-tuning wiki page, or it's a preference then the parameter creation will be handled there.
But it may be only a personal opinion. :)

src/Mod/Spreadsheet/Gui/Command.cpp
```diff
@@ -197,9 +197,36 @@ void CmdSpreadsheetImport::activated(int iMsg)
...
+        ParameterGrp::handle group = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Mod/Spreadsheet");
+        QString delimiter = QString::fromStdString(group->GetASCII("ImportExportDelimiter","tab"));
+        QString quoteChar = QString::fromStdString(group->GetASCII("ImportExportQuoteCharacter","\""));
+        QString escapeChar = QString::fromStdString(group->GetASCII("ImportExportEscapeCharacter","\\"));
+        group->SetASCII("ImportExportDelimiter", Base::Tools::toStdString(delimiter).c_str()); //create parameters that do not already exist for user convenience
+        group->SetASCII("ImportExportQuoteCharacter", Base::Tools::toStdString(quoteChar).c_str());
+        group->SetASCII("ImportExportEscapeCharacter", Base::Tools::toStdString(escapeChar).c_str());
+        char delim = delimiter.size() == 1 ? delimiter[0].toLatin1() : '\0';
+        if (delimiter.compare(QString::fromStdString("tab"), Qt::CaseInsensitive) == 0 || delimiter.compare(QString::fromStdString("\\t"),Qt::CaseInsensitive) == 0){
```
0penBrain, line 208:
```suggestion
        if (delimiter.compare(QLatin1String("tab"), Qt::CaseInsensitive) == 0 || delimiter.compare(QLatin1String("\\t"),Qt::CaseInsensitive) == 0){

```

src/Mod/Spreadsheet/Gui/Command.cpp
```diff
@@ -197,9 +197,36 @@ void CmdSpreadsheetImport::activated(int iMsg)
...
+            delim = '\t';
+        }
+        if(delim != '\0' && quoteChar.size() == 1 && escapeChar.size() == 1){
+            sheet->importFromFile(Base::Tools::toStdString(fileName),  delim, quoteChar[0].toLatin1(), escapeChar[0].toLatin1());
```
0penBrain, line 212:
```suggestion
            sheet->importFromFile(filename.toStdString(),  delim, quoteChar[0].toLatin1(), escapeChar[0].toLatin1());

```

src/Mod/Spreadsheet/Gui/Command.cpp
```diff
@@ -197,9 +197,36 @@ void CmdSpreadsheetImport::activated(int iMsg)
...
+            sheet->execute();
+        } else {
+            std::stringstream errMsg("Invalid spreadsheet Import/Export parameter.\n");
+            if (delim == '\0') {
+                errMsg << "Unrecognized delimiter: " << Base::Tools::toStdString(delimiter).c_str() << " (recognized tokens: tab, semicolon, comma, or any single character)\n";
```
0penBrain, line 217:
```suggestion
                errMsg << "Unrecognized delimiter: " << delimiter.toStdString() << " (recognized tokens: tab, semicolon, comma, or any single character)\n";

```

src/Mod/Spreadsheet/Gui/Command.cpp
```diff
@@ -197,9 +197,36 @@ void CmdSpreadsheetImport::activated(int iMsg)
...
+            }
+            if (quoteChar.size() != 1){
+                errMsg << "Invalid quote character: " << Base::Tools::toStdString(quoteChar).c_str() << " (quote character must be one single character)\n";
```
0penBrain, line 220:
```suggestion
                errMsg << "Invalid quote character: " << quoteChar.toStdString() << " (quote character must be one single character)\n";

```

src/Mod/Spreadsheet/Gui/Command.cpp
```diff
@@ -197,9 +197,36 @@ void CmdSpreadsheetImport::activated(int iMsg)
...
+            }
+            if (escapeChar.size() != 1){
+                errMsg << "Invalid escape character: " << Base::Tools::toStdString(escapeChar).c_str() << " (escape character must be one single character)\n";
```
0penBrain, line 223:
```suggestion
                errMsg << "Invalid escape character: " << escapeChar.toStdString() << " (escape character must be one single character)\n";

```

0penBrain (commented):
Remarks on Import apply on Export as well.
Notice I didn't tested (compiled) suggestions thus it has to be checked before eventual merge -- sorry, not so much time --. I can probably do it later if needed.

src/Mod/Spreadsheet/Gui/Command.cpp
chennes: While in principle I agree, in actual practice it is VERY nice to have these variables automatically appear in the user.cfg and the Edit Parameters screen. Something like _half_ of #4787 would be unnecessary if we did this for all preferences!

0penBrain: @chennes In this case I'd be more in favor to have a mechanism at startup (or workbench load) that takes care of that only once. ;)

chennes: Right, I'd be fine with this moving/copying to the constructor (or loader, etc.) and only doing an access here, rather than the set.

0penBrain: I'm not sure of that as it would mean to duplicate the location where default value is defined (or using some kind of a #DEFINE to store it). IMO it would be too risky to get the values without setting the default in case user inadvertently delete them...

chennes: I'm not sure what you're suggesting, then. We basically _have_ to do the get here, in case the preference changes after the WB is loaded. And there's no risk to doing a "get" when the user has deleted the preference, the system will just return the default value that's given as part of the "get" command:
```
QString delimiter = QString::fromStdString(group->GetASCII("ImportExportDelimiter","tab"));
QString quoteChar = QString::fromStdString(group->GetASCII("ImportExportQuoteCharacter","\""));
QString escapeChar = QString::fromStdString(group->GetASCII("ImportExportEscapeCharacter","\\"));
```
The big downside is that those three lines have to be duplicated, and if one of them ever changes, you have to make sure to repeat that change wherever you are doing the set of the default. So the efficiency gain of not setting them every time comes at a price of code duplication. I don't feel strongly about it, neither solution is actually "good," that's just how our preference system works.

mwganson:
We should consider creating parameters on access with Get.  This will grow the config file, but will it be a noticeable reduction in peformance?  Perhaps this in itself could be a parameter.   Problem with that is once the config file is grown there is no ungrowing it without deleting it and rebuilding on restart.

There are parameters that the users know nothing about because either they haven't read the documentation or because the parameters are not documented.  For example, did you know there is a way to underline the active body or make it bold or italic?  It would be nice to be able to browse the parameters and know every option available is in there for a particular workbench.  Does Spreadsheet have other parameters available than these 3?  I have no idea without searching through the source code.

As to putting the creation in the module startup, if that is done then they are created with every initialization of the module.  This way they are only created when an import or export is done, which is very rarely for most users and is very little overhead compared to the actual import/export.  We could test to see if they are already there before creating, but that is probably just as costly in performance as just simply setting them.

There is merit to the notion that if they are worth creating there should be a settings page in preferences.  This is true, but reality is there isn't one until somebody creates one.  The delimiter, in particular, not so much the other two, is really needed for spreadsheet import/export in the Gui.  This was a fairly quick thing to code, much needed, and if not ideal, it is at least a step in the right direction.

0penBrain:
@mwganson that was just an "open remark" and moreover a personal opinion that I was unsure is a good one. :) 
For sure it should absolutely not prevent PR merging. And it doesn't affect the deep respect I have for anyone contributing too. 👍

chennes:
Agreed on all counts, and I'm happy to merge as-is, potentially with the other suggestions re:string handling that @0penBrain has made. @mwganson is it your plan to incorporate those other suggestions?

mwganson:
I will incorporate the suggestions on the string handling.

mwganson:
After this is merged I will add a preference page in edit -> preferences for this workbench.  But before it is merged I am wondering if we should have separate parameters for importing and exporting?

chennes:
Truthfully, IMO in an idea world we'd prompt the user for the import settings every time: I get data in all sorts of formats, for me I can't just set the variable once and be done with it. For export, I can, though.

(Also, I've got to ask... of all the changes @0penBrain suggested I thought the removal of the double-conversion of the "tab" text was the "for sure" change -- why did you decide to keep the `char*`->`std::string`->`QString` conversion?)

mwganson:
Just an oversight.

I'm making progress on the settings page.  I'm thinking a combobox is the way to go for the delimiter.  That way the user can select from "tab", "semicolon", "comma", or enter a custom character.  A line edit should suffice for the other two.  I don't think these should be translatable.  Maybe it would be better to use "\t", ";", and "," instead of whole words, like "tab", but then again "\t" is kind of a programmer's thing.  Tab is the only one that can't easily be entered in a text field.

I was also thinking about a dialog during import/export after finishing with the preferences page.  The current preference can be the default in the dialog.

mwganson:
I will go ahead and add the preference page to this commit rather than a new PR since it hasn't been merged yet.

chennes:
Sounds good, thanks. I'm going to set this PR as draft -- once you're happy with it, flip it back to "Ready for review" and I'll give it another look.

mwganson:
I'm getting this warning message, but I don't know if it's related to this PR or not.  I'm rebuilding from master to see if it still shows up.  If it does I'll assume it's unrelated.  I have to post an image because it's inside a virtual machine and I can't share clipboards for copying text.

Snip macro screenshot-b21c7b

I just made a copy of the spreadsheet create new spreadsheet icon to use in the preferences dialog.

0penBrain:
@mwganson I think I already saw these 2 lines of warning when using the master, so I don't think it's related to your PR. Can check later if needed. 

I also had a quick look at the code. Looks fine but I find that there is a lot of duplicated code between import and export. What would you think about factorizing in a function that would get the 3 delimiting chars from the prefs and eventually return an error message if needed?

mwganson:
I'm still getting the same warning after compiling from master, so I will assume it's unrelated.

src/Mod/Spreadsheet/Gui/CMakeLists.txt
```diff
@@ -93,6 +96,28 @@ SET(SpreadsheetGuiIcon_SVG
...
+set(SpreadsheetGui_UIC_SRCS
+    DlgSettings.ui
+)
+
+
+if (BUILD_QT5)
+    qt5_wrap_ui(SpreadsheetGui_UIC_HDRS ${SpreadsheetGui_UIC_SRCS})
+else()
+    qt4_wrap_ui(SpreadsheetGui_UIC_HDRS ${SpreadsheetGui_UIC_SRCS})
+endif()
```
chennes, line 119: We don't support Qt4 anymore, so you don't need this block.

src/Mod/Spreadsheet/Gui/Command.cpp
```diff
@@ -197,9 +197,38 @@ void CmdSpreadsheetImport::activated(int iMsg)
...
+            if (escapeChar.size() != 1){
+                errMsg << "Invalid escape character: " << escapeChar.toStdString() << " (escape character must be one single character)\n";
+            }
+            errMsg << "Import not done.\n";
+            Base::Console().Error(errMsg.str().c_str());
+            return;
 
-        sheet->importFromFile(Base::Tools::toStdString(fileName), '\t', '"', '\\');
-        sheet->execute();
+        }
```
chennes, line 231: Extract to its own function.

src/Mod/Spreadsheet/Gui/DlgSettingsImp.cpp
```diff
@@ -0,0 +1,115 @@
+/***************************************************************************
+ *   Copyright (c) 2002 Jürgen Riegel <juergen.riegel@web.de>              *
```
chennes, line 2: Probably not true 😄

src/Mod/Spreadsheet/Gui/DlgSettingsImp.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <Gui/PrefWidgets.h>
+#include <Base/Console.h>
+
+using namespace SpreadsheetGui;
+
+/* TRANSLATOR SpreadsheetGui::DlgSettingsImp */
+
+/**
+ *  Constructs a DlgSettingsImp which is a child of 'parent', with the
+ *  name 'name' and widget flags set to 'f'
```
chennes, line 41: I know it's copy-and-paste boilerplate... but it's still wrong, there is no "name" or "f" here.

src/Mod/Spreadsheet/Gui/DlgSettingsImp.h
```diff
@@ -0,0 +1,57 @@
+/***************************************************************************
+ *   Copyright (c) 2002 Jürgen Riegel <juergen.riegel@web.de>              *
```
chennes, line 2: Update

chennes (commented):
Nice update, looking forward to giving it a spin.

mwganson:
@0penBrain Yes, there is some code duplication.  I could put the validation code in a validation function in the SheetView class that would return a boolean value, something like bool SheetView::validateImportExportParameters() that would return false if the parameters are invalid and print the message to the console as an error if invalid.  There it could be also called in preferences, I think.

mwganson:
I based it on the Ray Tracing preferences.  Copy/Paste/Modify.

0penBrain:
@mwganson I was more thinking about something like `bool getCharsFromPrefs(char &délimiter, char &quote, char &escape, std::string &errMsg)` that would basically do the whole processing.

mwganson:
I think I've addressed all concerns.  Let me know if I overlooked anything or if any other improvements can be made.

chennes:
I feel like there is some issue with using Inkscape-generated SVG files, that they need to be exported as "plain SVG" to strip out some custom Inkscape formatting. Let me look around for documentation...

chennes:
Found it: https://wiki.freecadweb.org/Artwork_Guidelines#Recommended_recording_format

berndhahnebach:
FreeCAD CI-repository:
branch:
https://gitlab.com/berndhahnebach/FreeCAD/-/commits/PR_5059

pipeline:
https://gitlab.com/berndhahnebach/FreeCAD/-/pipelines/377336997

status:
success

0penBrain (approved):
I didn't compile but just do a reading code review.
Looks good to me.
Maybe the change in SpreadsheetView that is just 2 newlines can be removed, but it doesn't harm.

0penBrain:
@mwganson I'm sorry that I didn't see it before. Could you squash commits all together or by meaningful groups before merging?

mwganson:
I removed the whitespace after you pointed it out.

I'm afraid to try to squash.  Last time I tried to squash commits together I goofed it all up.  Isn't there an Squash and Merge option when merging?

chennes:
There is, but last time I did it I messed it up! 😄

0penBrain:
@mwganson Indeed the history in your branch is very weird (commits are mixed with the master commits). Looks like you didn't --rebase when you update against the upstream master :confused:

mwganson:
I used the button on github "update branch".  I think that's what did it, but maybe it was something else.  I was able to successfully rebase spreadsheet2 down from 3 commits to a single commit.  The trick is to leave the first pick in.  My local copy of this branch is a mess.  I don't want to push it.  I'll close this PR and submit a new one from the other branch.  Sorry for all the trouble.

PR #5068, opened 2021-09-27
 Request for remove more than 100 duplicate translations from Sketcher.ts

FreeCAD-Tools:
This request should remove more than 100 duplicate translations from Sketcher. ts. It is needed to reduce the number of duplicates in Crowdin. And for group many identical strings to translate into one string. Remarks is welcome.

Discussion on forum: https://forum.freecadweb.org/viewtopic.php?f=10&t=62499

This pull needs rebuild of Crowdin sources after merge 

- [x]  Discuss on forum
- [x]  Simple optimization
- [x]  Your pull request is well written and has a good description
- [x]  No ticket

### DIFF
--- a/src/Mod/Sketcher/Gui/Command.cpp
+++ b/src/Mod/Sketcher/Gui/Command.cpp
@@ -139,7 +139,7 @@ CmdSketcherNewSketch::CmdSketcherNewSketch()
     :Command("Sketcher_NewSketch")
 {
     sAppModule      = "Sketcher";
-    sGroup          = QT_TR_NOOP("Sketcher");
+    sGroup          = QT_TRANSLATE_NOOP("Workbench","Sketcher");
     sMenuText       = QT_TR_NOOP("Create sketch");
     sToolTipText    = QT_TR_NOOP("Create a new sketch.");
     sWhatsThis      = "Sketcher_NewSketch";
@@ -262,7 +262,7 @@ CmdSketcherEditSketch::CmdSketcherEditSketch()
     :Command("Sketcher_EditSketch")
 {
     sAppModule      = "Sketcher";
-    sGroup          = QT_TR_NOOP("Sketcher");
+    sGroup          = QT_TRANSLATE_NOOP("Workbench","Sketcher");
     sMenuText       = QT_TR_NOOP("Edit sketch");
     sToolTipText    = QT_TR_NOOP("Edit the selected sketch.");
     sWhatsThis      = "Sketcher_EditSketch";
@@ -292,7 +292,7 @@ CmdSketcherLeaveSketch::CmdSketcherLeaveSketch()
   : Command("Sketcher_LeaveSketch")
 {
     sAppModule      = "Sketcher";
-    sGroup          = QT_TR_NOOP("Sketcher");
+    sGroup          = QT_TRANSLATE_NOOP("Workbench","Sketcher");
     sMenuText       = QT_TR_NOOP("Leave sketch");
     sToolTipText    = QT_TR_NOOP("Finish editing the active sketch.");
     sWhatsThis      = "Sketcher_LeaveSketch";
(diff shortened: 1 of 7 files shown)


berndhahnebach:
FreeCAD CI-repository:
branch:
https://gitlab.com/berndhahnebach/FreeCAD/-/commits/PR_5068

pipeline for commit 67aeaba4c1476a:
https://gitlab.com/berndhahnebach/FreeCAD/-/pipelines/377906378

status:
success

yorikvanhavre:
That's really an impressive work you did! I'm all for it... the use of translate instead of tr is IMHO very good, the tr form is really not useful anywhere... It is always better to specify the context manually.

yorikvanhavre:
Just a note, for a next time, it'd be great if you could condensate or organize your commits better, and give more meaningful names than "Updating file XXXX". Otherwise if everybody does so we quickly pollute the code history and it makes it harder to get back to specific commits in time. This is hard to do directly from github, as it has limited git tools. Better to do it from your own computer ;) Many here will be able to help with that.

chennes:
@yorikvanhavre Based on @wwmayer's comments in the forum (https://forum.freecadweb.org/viewtopic.php?p=536443#p536443) do we need to revert this? Or just code up a new solution and commit that over the top?

FreeCAD-Tools:
> 
> 
> Just a note, for a next time, it'd be great if you could condensate or organize your commits better, and give more meaningful names than "Updating file XXXX". Otherwise if everybody does so we quickly pollute the code history and it makes it harder to get back to specific commits in time. This is hard to do directly from github, as it has limited git tools. Better to do it from your own computer ;) Many here will be able to help with that.

Im sorry. I will leave comments via the GitHub site, there is such an opportunity there. But, it seems that it is impossible to combine commits in github site.

wwmayer:
> Based on @wwmayer's comments in the forum (https://forum.freecadweb.org/viewtopic.php?p=536443#p536443) do we need to revert this? Or just code up a new solution and commit that over the top?

No, we don't have to. But IMO we should replace
`   sGroup          = QT_TRANSLATE_NOOP("Workbench","Sketcher");` with
`   sGroup          = "Sketcher";` because there is no point in repeating the context string over and over in all command classes.
A simpler solution is to write it exactly once for each group name (e.g. in Workbench.cpp of each module).

yorikvanhavre:
Sorry, I merged too soon..

> IMO we should replace sGroup = QT_TRANSLATE_NOOP("Workbench","Sketcher"); with sGroup = "Sketcher"; 

We still need to have the "Sketcher" string picked by lupdate somewhere, though? How to set the right context?

chennes:
I am in favor of @wwmayer's suggestion in the forums, of completely dropping the QT_TRANSLATE_NOOP for sGroup, and using the "dummy" calls to 
```
#if 0 // needed for Qt's lupdate utility
    qApp->translate("Command", "Something");
#endif
```

FreeCAD-Tools:
@chennes
> ETA: I figured that he and @FreeCAD-Tools were already working on that, though, so I didn't wade in to do any coding yet.

@wwmayer's make two commit:

https://github.com/FreeCAD/FreeCAD/commit/bc57ba6027

but it have not activated fragment.

>#if 0 // not yet activated
>    if (text == QString::fromLatin1(getGroupName()))
>        text = qApp->translate("Workbench", getGroupName());
>#endif

and a similar commit for other commands https://github.com/FreeCAD/FreeCAD/commit/fe7f4c4be7

FreeCAD-Tools:
> > IMO we should replace sGroup = QT_TRANSLATE_NOOP("Workbench","Sketcher"); with sGroup = "Sketcher";
> 
> We still need to have the "Sketcher" string picked by lupdate somewhere, though? How to set the right context?

@yorikvanhavre What do you think should be done?

PR #5069, opened 2021-09-27
[Arch] Fix LGTM warnings

chennes:
A number of small fixes addressing complaints from the LGTM static analyzer.

### DIFF
--- a/src/Mod/Arch/ArchCommands.py
+++ b/src/Mod/Arch/ArchCommands.py
@@ -416,7 +416,6 @@ def getCutVolume(cutplane,shapes,clip=False,depth=None):
         bb.add(sh.BoundBox)
     bb.enlarge(1)
     # building cutplane space
-    placement = None
     um = vm = wm = 0
     try:
         if hasattr(cutplane,"Shape"):
@@ -578,9 +577,8 @@ def meshToShape(obj,mark=True,fast=True,tol=0.001,flat=False,cut=True):
 
     name = obj.Name
     if "Mesh" in obj.PropertiesList:
-        faces = []
         mesh = obj.Mesh
-        plac = obj.Placement
+        #plac = obj.Placement
         solid = getShapeFromMesh(mesh,fast,tol,flat,cut)
         if solid:
             if solid.isClosed() and solid.isValid():
@@ -635,7 +633,7 @@ def removeShape(objs,mark=True):
                     v1 = dims[0].multVec(v1)
                     v2 = dims[0].multVec(v2)
                     line = Draft.makeLine(v1,v2)
-                    wal = ArchWall.makeWall(line,width=width,height=dims[3],name=name)
+                    ArchWall.makeWall(line,width=width,height=dims[3],name=name)
         else:
             if mark:
                 obj.ViewObject.ShapeColor = (1.0,0.0,0.0,1.0)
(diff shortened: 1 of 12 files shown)


berndhahnebach:
FreeCAD CI-repository:
branch:
https://gitlab.com/berndhahnebach/FreeCAD/-/commits/PR_5069

pipeline for commit 78ecda36632408:
https://gitlab.com/berndhahnebach/FreeCAD/-/pipelines/378167853

status:
success

src/Mod/Arch/ArchCommands.py
```diff
@@ -635,7 +633,7 @@ def removeShape(objs,mark=True):
                     v1 = dims[0].multVec(v1)
                     v2 = dims[0].multVec(v2)
                     line = Draft.makeLine(v1,v2)
-                    wal = ArchWall.makeWall(line,width=width,height=dims[3],name=name)
+                    _ = ArchWall.makeWall(line,width=width,height=dims[3],name=name)
```
yorikvanhavre, line 636: No actual need to catch the returned value I guess
```suggestion
                    ArchWall.makeWall(line,width=width,height=dims[3],name=name)
```

src/Mod/Arch/ArchCommands.py
```diff
@@ -1336,14 +1334,14 @@ def printMessage( message ):
     if FreeCAD.GuiUp :
         if sys.version_info.major < 3:
             message = message.decode("utf8")
-        reply = QtGui.QMessageBox.information( None , "" , message )
+        _ = QtGui.QMessageBox.information( None , "" , message )
```
yorikvanhavre, line 1337:
```suggestion
        QtGui.QMessageBox.information( None , "" , message )
```

src/Mod/Arch/ArchCommands.py
```diff
@@ -1336,14 +1334,14 @@ def printMessage( message ):
...
 
 def printWarning( message ):
     FreeCAD.Console.PrintMessage( message )
     if FreeCAD.GuiUp :
         if sys.version_info.major < 3:
             message = message.decode("utf8")
-        reply = QtGui.QMessageBox.warning( None , "" , message )
+        _ = QtGui.QMessageBox.warning( None , "" , message )
```
yorikvanhavre, line 1344:
```suggestion
        QtGui.QMessageBox.warning( None , "" , message )
```

src/Mod/Arch/ArchMaterial.py
```diff
@@ -21,7 +21,7 @@
 
 import FreeCAD
 if FreeCAD.GuiUp:
-    import FreeCADGui, Arch_rc, os
```
yorikvanhavre, line 24: Baware that importing the _rc file is needed to load the resources and be able to use any of the :/icons/ later on, in this case at L206:
`return ":/icons/Arch_Material_Group.svg"`

yorikvanhavre (changes requested):
I think just the removal of the Arch_rc module needs to be fixed (or changed somehow). The rest are mere suggestions...

src/Mod/Arch/ArchReference.py
```diff
@@ -183,14 +183,14 @@ def cleanShape(self,shapedata,obj,materials):
...
                     for soledge in solid.Edges:
-                        if found:
-                            break
                         if edge.hashCode() == soledge.hashCode():
                             found = True
                             break
+                    if found:
+                        break
+                if found:
+                    break
```
chennes, line 193: @yorikvanhavre can you double-check this change? It's not just a reformat, it is an algorithm change: the outermost for loop has an `else` clause, which only makes sense if there is a `break` statement that affects that outer loop. So I guessed at what that clause might be, but I'm not 100% sure of this one.

freecadci:
FreeCAD CI-repository:
branch:
https://gitlab.com/berndhahnebach/FreeCAD/-/commits/PR_5069

pipeline for commit 06a0e1e923ca01:
https://gitlab.com/berndhahnebach/FreeCAD/-/pipelines/378948281

status:
success

src/Mod/Arch/ArchReference.py
yorikvanhavre: Yes, you are right... Indeed my implementation was not good, the else clause would almost never be triggered when it needs. I think yours fixes the problem

yorikvanhavre:
I think this is good to go! I'll let you press the button unless you still want to add something

yorikvanhavre (approved):

chennes:
Got it, thanks for reviewing!

PR #5085, opened 2021-10-03
[Spreadsheet] fix preferences UI

donovaly:
- add missing layouts (Qt's designer warns too)
- add spacer to assure a compact layout on all screens

### DIFF
--- a/src/Mod/Spreadsheet/Gui/DlgSettings.ui
+++ b/src/Mod/Spreadsheet/Gui/DlgSettings.ui
@@ -6,181 +6,175 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>555</width>
-    <height>413</height>
+    <width>322</width>
+    <height>149</height>
    </rect>
+  </property>
+  <property name="sizePolicy">
+   <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
+    <horstretch>0</horstretch>
+    <verstretch>0</verstretch>
+   </sizepolicy>
   </property>
   <property name="windowTitle">
    <string>Spreadsheet</string>
   </property>
-  <widget class="QGroupBox" name="groupBox">
-   <property name="geometry">
-    <rect>
-     <x>10</x>
-     <y>10</y>
-     <width>531</width>
-     <height>161</height>
-    </rect>
-   </property>
-   <property name="title">
-    <string>Import/Export Settings</string>
-   </property>
-   <widget class="Gui::PrefLineEdit" name="quoteCharLineEdit">
-    <property name="geometry">
-     <rect>
-      <x>156</x>
-      <y>70</y>
-      <width>142</width>
-      <height>25</height>
-     </rect>
-    </property>
-    <property name="toolTip">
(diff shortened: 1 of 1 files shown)


donovaly:
@mwganson I stumbled over this while documenting the new spreadsheet preferences. Thanks for this helpful new feature!
Can you please have a look, if I documented it correctly: https://wiki.freecadweb.org/Spreadsheet_Preferences
(for example is the delimiter used for the import AND the export, or only the export?)

chennes:
Thanks for this, @donovaly -- I had added it to my "to-do" list, but not gotten to it yet.

chennes:
Can you add horizontal spacers between the labels and the fields, or make the fields expanding, so it matches our other preferences panes? I know it's a little odd on its own, but I think it's worthwhile for consistency.

donovaly:
> Can you add horizontal spacers between the labels and the fields, or make the fields expanding, so it matches our other preferences panes?

Done now.
(That was also the first thing I did but since I looked kind of lost in a big white space, I put the spacer behind the group box.)

mwganson:
Thanks.  I've also had documentation on the back burner.  QtCreator and I have never gotten along well.

mwganson:
All of the parameters are applied on both import and export.  I had considered separating them, but decided not to.  If you want to separate them now is the time before too many begin using them.

The logical next step for this would be to have a task dialog where the user can take these defaults and change them for that particular operation.  The import could possibly be its own preview.  Do the import and show it.  If the user changes a parameter, remove the sheet and do another import to show the changes.  On cancel, remove the sheet.   On OK just close the dialog. For export, maybe have a text field widget in the form to show a preview of the csv.

For the settings page now that the skeleton is there other preferences could be added, for example styles or alignment preferences.

donovaly:
> QtCreator and I have never gotten along well.

I use directly the Designer. This is the tool also the Creator uses for the forms. The advantage is that Designer is more strict and warns you about missing layouts.

> All of the parameters are applied on both import and export. I had considered separating them, but decided not to. If you want to separate them now is the time before too many begin using them.

Yes, having them separated would be nice. The reason is that I often get CSVs and sometimes their delimiter is ';', sometimes a TAB, a comma etc. To be able to use such a CSV to import as spreadsheet I think it is very valuable to set different import delimiter character. However for the usability, I think it is additionally necessary that one gets an import dialog asking for your the delimiter. So in the preferences one sets the delimiter that should be proposed as input delimiter for this import dialog.

So when I am dreaming ;-) , allow me to propose the following things:
- in the current import dialog, please add file filters. At the moment there is only "All". It would be useful to have there also "*.csv", and "*.tsv" and csv should be the default filter
- take an existing CSV import dialog from other OpenSource programs or a look have at their implementation. I use for example the program LabPlot: https://labplot.kde.org/
This is a Qt program and they have an excellent CSV import dialog. I think we don't need to reinvent the wheel since many programs have the same tasks  - besides the delimiters, exclude header/footer lines, omit some rows etc.

donovaly:
@chennes the PR should be ready to be merged, since what I discussed with mwganson is future stuff.

chennes:
Agreed, this looks good, and it sounds like there are some good ideas for future improvements. Merging...

PR #5098, opened 2021-10-07
Fix STEP import/export color problem

realthunder:
Forum thread https://forum.freecadweb.org/viewtopic.php?f=3&t=62103

The problem is caused by OCCT 7.5 (with [this](https://github.com/Open-Cascade-SAS/OCCT/commit/ba00aab7a0f) commit). Also see article [here](https://dev.opencascade.org/content/occt-3d-viewer-becomes-srgb-aware).

Note that in my opinion, OCCT's implementation is incorrect by always exporting STEP color in sRGB color space (see [here](https://github.com/Open-Cascade-SAS/OCCT/blob/ba00aab7a0f30b2a77d3faa4191249b32a9e03b0/src/STEPConstruct/STEPConstruct_Styles.cxx#L636)), while other CAD software (even OCCT <= 7.4) simply import the color as normal linear RGB color. This commit is more of a work around of this problem. Not sure if OCCT will change this in the future.

### DIFF
--- a/src/Mod/Part/Gui/ViewProviderExt.cpp
+++ b/src/Mod/Part/Gui/ViewProviderExt.cpp
@@ -454,7 +454,7 @@ void ViewProviderPartExt::onChanged(const App::Property* prop)
             float trans = Transparency.getValue()/100.0f;
             auto colors = DiffuseColor.getValues();
             for (auto &c : colors)
-                c.a = Transparency.getValue()/100.0f;
+                c.a = trans;
             DiffuseColor.setValues(colors);
 
             App::PropertyContainer* parent = ShapeMaterial.getContainer();


chennes:
Thanks, @realthunder . @Syres916, when you have a minute could you please re-test? It should work now.

realthunder:
@easyw you've discovered an unrelated bug, which affects per face color output when you export a single object with placement. The above commit fixes this problem. The bug does not affect exporting group/assembly.

Screenshot from 2021-10-09 16-11-45

realthunder:
@Syres916 can you please run the following command at your local repo, compile, run, and see if you can get the color? If it works fine, then it means OCCT < 7.5 does not support RGBA color in STEP, and I'll change the code to conditionally enable only for OCCT >= 7.5.

```
git checkout c19e1398f260fcd519cb75da4350108cecee0fdb
```

chennes:
@realthunder is this good to merge now?

chennes:
Thanks @realthunder for the PR, and @Syres916 and @easyw for the testing and comments, great work. Merging now...

Cronic-se:
Hi

Just want to share that the import function in 0.19.3 is malfunctioning.

I have attached pictures of the same file imported in 0.18 VS 0.19.3.
I did the exact same procedure. New file > Import > Selecting step-file > Ok.



FC_0 18_import
FC_0 19 3_import

luzpaz:
@Cronic-se v0.19.3 was not the branch this was merged in to. You need to install the [weekly build](https://github.com/FreeCAD/FreeCAD-Bundle/releases/tag/weekly-builds) (which updates every Sunday, or build it yourself) to take advantage of this update.

donovaly:
> @Cronic-se v0.19.3 was not the branch this was merged in to. 

Not, this PR was backported to FreeCAD 0.19.3, see our changelog: https://github.com/FreeCAD/FreeCAD/releases/tag/0.19.3#changelog

@Cronic-se, can you please bring this to the forum, post there the content of FreeCAD's About dialog and attach your STEP file.
We know FreeCAD 0.19.3 works if you also have the library OCC 7.5.2. If this does not work for you, wee need to investigate further.

luzpaz:
@donovaly oh, odd. I don't see it notated anywhere explicitly in this ticket. I guess I missed it.

donovaly:
> @donovaly oh, odd. I don't see it notated anywhere explicitly in this ticket. I guess I missed it.

You don't see in a PR if this was ported to other branches. Therefore we setup this list which I try to keep up to date:
https://forum.freecadweb.org/viewtopic.php?f=10&t=57659
This list is also the base for the changelog of the maintenance releases.

PR #5099, opened 2021-10-08
[Spreadsheet] Refactor keyboard handling

chennes:
LineEdit no longer actually handles motion, it simply indicates which action was taken to cause it to lose focus (e.g. which key was pressed). It's up to the client code to determine what this means. This allows significant consolidation of keyboard-handling logic, and the implementation of more extensive keyboard navigation features.

New keyboard shortcuts include a tab counter to implement auto-return, plus Ctrl->Arrow, End, Home, Ctrl-End, and Ctrl-Home, matching the behavior of OpenOffice, LibreOffice, etc.

Testers: please in particular ensure that keyboard navigation when entering and updating data is correct, and that the alias- and expression-completer appears and disappears as expected.

### DIFF
--- a/src/Mod/Spreadsheet/App/PropertySheet.cpp
+++ b/src/Mod/Spreadsheet/App/PropertySheet.cpp
@@ -982,6 +982,14 @@ void PropertySheet::getSpans(CellAddress address, int & rows, int & cols) const
     }
 }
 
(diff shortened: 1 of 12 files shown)


0penBrain:
@chennes hopefully I can give it another check. ATM due to limited time and energy, I focused on things that could be not discovered immediately by beta users doing functional testing.

PR #5104, opened 2021-10-10
[Gui] Migrate UrlLabel to use QDesktopServices

chennes:
UrlLabel predates QDesktopServices' ability to ask the host OS to launch a URL, and resorted to a Python workaround. This migrates that code to QDesktopServices, making the code more reliable across more OSes, and shortening it significantly.


chennes:
linkActivated only pertains to links within `<a>` tags in the text of the label, and the entire purpose of this label is to get a link that is *not* in `<a>` tags, so that it can be styled using the Qt stylesheet.

PR #5130, opened 2021-10-27
Draft: Fix merge layers

Roy-043:
The "Merge layer duplicates" Tree view context menu option for LayerContainers did not work.

Additionally:
- Using `openTransaction` and `commitTransaction` is required for the `merge_by_name` function but also for the `add_layer` function. Otherwise undo-redo does not work properly.
- Have replaced `translate("Draft", ...)` with `_tr(...)`.

Forum topic: https://forum.freecadweb.org/viewtopic.php?f=23&t=60843

---

### DIFF
--- a/src/Mod/Draft/draftviewproviders/view_layer.py
+++ b/src/Mod/Draft/draftviewproviders/view_layer.py
@@ -1,6 +1,7 @@
 # ***************************************************************************
 # *   Copyright (c) 2014 Yorik van Havre <yorik@uncreated.net>              *
 # *   Copyright (c) 2020 Eliud Cabrera Castillo <e.cabrera-castillo@tum.de> *
+# *   Copyright (c) 2021 FreeCAD Developers                                 *
 # *                                                                         *
 # *   This file is part of the FreeCAD CAx development system.              *
 # *                                                                         *
@@ -37,7 +38,7 @@ import FreeCAD as App
 import FreeCADGui as Gui
 
 from draftutils.messages import _msg
-from draftutils.translate import translate
+from draftutils.translate import _tr
 from draftobjects.layer import Layer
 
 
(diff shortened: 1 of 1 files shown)


chennes:
Thanks, looks good!

wwmayer:
> Have replaced translate("Draft", ...) with _tr(...).

@Roy-043  Note, that translate("Draft", ...) is not the same as _tr(...) because pyside2-lupdate won't add the strings to the .ts files any more.
As a quick test run `pyside2-lupdate view_layer.py -ts test.ts` and open the file in a text editor. None of the strings with _tr(...) are listed there. Insofar a clear regression.

I wonder what the function _tr() is good for as it only translates text but can't be used to pick them up. The same applies to the _qstr() function.

What seems to work is to rename _tr() to tr() and use `@default` as context string:
```
def tr(text):
    return translate("@default", text)
```

Roy-043:
I do not fully understand how `lupdate` should work. I am testing it now with an old version of `view_layer.py` and the produced TS file contains zero source texts.

See forum topic: https://forum.freecadweb.org/viewtopic.php?f=10&t=63456

Regarding the `_tr()` function: It is used elsewhere in the Draft code, and it seemed a more efficient way of doing things. If it does not work properly we must of course remove it from all Draft code.

wwmayer:
Have a look at https://doc.qt.io/qt-5/i18n-source-translation.html
Basically Qt offers these functions for translations:
* tr() : this marks strings for lupdate and translates them immediately. In C++ the context is the class name but in Python it seems to use the string `@default`
* QApplication::translate() : does the same as above but you must explicitly set the context
* QT_TR_NOOP : only marks the string for lupdate but doesn't translate it.
* QT_TRANSLATE_NOOP : does the same as above but you must explicitly set the context

The _tr() function is only able to translate strings but it doesn't mark them for lupdate. So, as soon as a string will be changed we won't be able any more to add them to the .ts files.

luzpaz:
CC @yorikvanhavre :point_up:

yorikvanhavre:
We should basically remove all _tr from Draft... I think the best strategy is to remove it from Draftutils so we make sure anything that still uses it fails. I'll take care of this

PR #3256, opened 2020-03-24
[Gui] Allow dot as decimal separator for locales not using it as thousand separator

0penBrain:
- [X] Branch rebased on latest master `git pull --rebase upstream master`
- [ ] Unit tests confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [X] Commit message is [well-written](https://chris.beams.io/posts/git-commit/)
- [X] Commit message includes `issue #<id>` or `fixes #<id>` where `<id>` is the [associated MantisBT](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) issue id if one exists

More details in the commit message

PR #5150, opened 2021-11-07
[Core] Improve decimal point substitution mechanism

0penBrain:
This PR improves the decimal point substitution mechanism that allows it to be conform to locale settings (introduced in PR #3256), when it is entered from the numerical pad : 
* The PR simplifies the mechanism. The current one is a bit weird because at the time I did it I observed a strange behavior of Qt modifiers() value. It happens that this weirdness comes from the OS, so it doesn't have to be fixed in Qt. Fixes exists in OS
* Following @wwmayer [post here](https://forum.freecadweb.org/viewtopic.php?f=8&t=62546&start=20#p544339), the mechanism has been extended to cope with substitution in a more general way

The algorithm is now simple : 
If comma or period key event coming from numerical keypad, compare with locale decimal point value. If different, tamper the event substituting with locale value.

### DIFF
--- a/src/Gui/GuiApplication.cpp
+++ b/src/Gui/GuiApplication.cpp
@@ -334,13 +334,13 @@ bool KeyboardFilter::eventFilter(QObject* obj, QEvent* ev)
 {
     if (ev->type() == QEvent::KeyPress || ev->type() == QEvent::KeyRelease) {
         QKeyEvent *kev = static_cast<QKeyEvent *>(ev);
-        QAbstractSpinBox *target = dynamic_cast<QAbstractSpinBox *>(obj);
-        if (kev->key() == Qt::Key_Period && target)
+        Qt::KeyboardModifiers mod = kev->modifiers();
+        int key = kev->key();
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5150">pipeline status</a> for feature branch [PR_5150](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5150). Pipeline [403877053](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/403877053) was triggered at [bf1e392](https://github.com/FreeCAD/FreeCAD/pull/5150/commits/bf1e392d2a6c48d69cdb719b1b8743c15d5c122e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
do we need to set the parameter for the substitution with this PR too or will it be on by default?

PR #2862, opened 2020-01-04
Configuration Table using Spreadsheet

realthunder:
Forum thread https://forum.freecadweb.org/viewtopic.php?f=17&t=42183

PR #5151, opened 2021-11-08
App: Merge realthunder PR 2862, stage 2

chennes:
This is the second stage of cherry-picking commits from @realthunder's PR #2862 -- Configuration Table using Spreadsheet. The main work of this set of commits is to introduce the `HIDDENREF()` function (formerly the `HREF()` function). To borrow from [realthunder's documentation](https://github.com/realthunder/FreeCAD_assembly3/wiki/Expression-and-Spreadsheet#href):

> `hiddenref()` stands for hidden reference, which accepts a single argument containing an property reference. This function hides any object reference inside the argument to work around cyclic dependency error. You need to be careful when using `hiddenref()` because skipping dependency checking may result in unstable recomputing order, and thus given unexpected result. As a rule of thumb, it will be safe if `hiddenref()` is referring (directly or indirectly) to some property that will only be changed by user instead of calculated by the object (e.g. with other expressions). When used right, `hiddenref()` can be a powerful tool. You can, for example, expose module parameters in a parent container (such as an `App::Part` or `PartDesign::Body`), and refer to these parameters inside any child features, e.g. `hiddenref(Part001.Length)`.

### DIFF
--- a/src/App/Document.cpp
+++ b/src/App/Document.cpp
@@ -444,13 +444,14 @@ void Document::exportGraphviz(std::ostream& out) const
                 // Create subgraphs for all documentobjects that it depends on; it will depend on some property there
                 auto i = expressions.begin();
                 while (i != expressions.end()) {
-                    std::set<ObjectIdentifier> deps;
+                    std::map<ObjectIdentifier,bool> deps;
 
                     i->second->getIdentifiers(deps);
 
-                    std::set<ObjectIdentifier>::const_iterator j = deps.begin();
-                    while (j != deps.end()) {
-                        DocumentObject * o = j->getDocumentObject();
+                    for(auto j=deps.begin(); j!=deps.end(); ++j) {
+                        if(j->second)
+                            continue;
+                        DocumentObject * o = j->first.getDocumentObject();
 
                         // Doesn't exist already?
                         if (o && !GraphList[o]) {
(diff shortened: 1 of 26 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5151">pipeline status</a> for feature branch [PR_5151](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5151). Pipeline [404206871](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/404206871) was triggered at [6c8f709](https://github.com/FreeCAD/FreeCAD/pull/5151/commits/6c8f709062d1025ea3d0910dfea682563ece64df). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

carlopav:
@chennes is there something we can test in this PR to speed up the merge process?

chennes:
No, we're making good progress here right now -- @realthunder said he plans on rebasing #2862 again now that this merge is complete. Then I should be able to quickly merge all but the very last commit of that PR because those intermediate commits are small and straightforward. That will leave just the one final large commit to dig into (if all goes as planned) -- I'll probably call in assistance for that one 😃 .

PR #5149, opened 2021-11-07
[PD] fixes loft visibility on creation and modification

donovaly:
- the PR fixes the issue reported here: https://forum.freecadweb.org/viewtopic.php?f=3&t=63252
- (also a MSVC code formatting fix)

PR #5155, opened 2021-11-08
[PD] Support adding solid faces for loft and pipe sections

AjinkyaDahale:
With these changes, one face per solid can be added as either the first
"profile" or subsequent sections in loft and pipe.

These changes depend on `App::PropertyXLinkSubList` preserving the order in which
sections are added.

A minor change this also adds is that when a solid's face is selected that face
is mentioned in the fields instead of the solid (eg `Box:Face1` instead of
`Box`).

### DIFF
--- a/src/Mod/PartDesign/App/FeatureLoft.cpp
+++ b/src/Mod/PartDesign/App/FeatureLoft.cpp
@@ -55,7 +55,7 @@ PROPERTY_SOURCE(PartDesign::Loft, PartDesign::ProfileBased)
 Loft::Loft()
 {
     ADD_PROPERTY_TYPE(Sections,(0),"Loft",App::Prop_None,"List of sections");
-    Sections.setSize(0);
+    Sections.setValue(0);
     ADD_PROPERTY_TYPE(Ruled,(false),"Loft",App::Prop_None,"Create ruled surface");
     ADD_PROPERTY_TYPE(Closed,(false),"Loft",App::Prop_None,"Close Last to First Profile");
 }
@@ -95,44 +95,56 @@ App::DocumentObjectExecReturn *Loft::execute(void)
     }
 
     try {
-        //setup the location
+        // setup the location
         this->positionByPrevious();
         TopLoc_Location invObjLoc = this->getLocation().Inverted();
-        if(!base.IsNull())
+        if (!base.IsNull())
             base.Move(invObjLoc);
 
-        //build up multisections
+        // build up multisections
         auto multisections = Sections.getValues();
-        if(multisections.empty())
+        if (multisections.empty())
             return new App::DocumentObjectExecReturn("Loft: At least one section is needed");
 
         std::vector<std::vector<TopoDS_Wire>> wiresections;
-        for(TopoDS_Wire& wire : wires)
+        for (TopoDS_Wire& wire : wires)
             wiresections.emplace_back(1, wire);
 
-        for(App::DocumentObject* obj : multisections) {
-            if(!obj->isDerivedFrom(Part::Feature::getClassTypeId()))
+        for (auto obj : multisections) {
+            if (!obj->isDerivedFrom(Part::Feature::getClassTypeId()))
                 return  new App::DocumentObjectExecReturn("Loft: All sections need to be part features");
+
+            // if the section is an object's face then take just the face
+            TopoDS_Shape shape;
+            if (obj->isDerivedFrom(Part::Part2DObject::getClassTypeId()))
+                shape = static_cast<Part::Part2DObject*>(obj)->Shape.getValue();
+            else {
+                auto subValues = Sections.getSubValues(obj);
+                if (subValues.empty())
+                    throw Base::ValueError("Loft: No valid subelement linked in Part::Feature");
+
+                shape = static_cast<Part::Feature*>(obj)->Shape.getShape(). getSubShape(subValues[0].c_str());
+            }
 
             TopExp_Explorer ex;
             size_t i=0;
-            for (ex.Init(static_cast<Part::Feature*>(obj)->Shape.getValue(), TopAbs_WIRE); ex.More(); ex.Next(), ++i) {
-                if(i>=wiresections.size())
+            for (ex.Init(shape, TopAbs_WIRE); ex.More(); ex.Next(), ++i) {
+                if (i>=wiresections.size())
                     return new App::DocumentObjectExecReturn("Loft: Sections need to have the same amount of inner wires as the base section");
                 wiresections[i].push_back(TopoDS::Wire(ex.Current()));
             }
-            if(i<wiresections.size())
+            if (i<wiresections.size())
                     return new App::DocumentObjectExecReturn("Loft: Sections need to have the same amount of inner wires as the base section");
 
         }
 
-        //build all shells
+        // build all shells
         std::vector<TopoDS_Shape> shells;
-        for(std::vector<TopoDS_Wire>& wires : wiresections) {
+        for (std::vector<TopoDS_Wire>& wires : wiresections) {
 
             BRepOffsetAPI_ThruSections mkTS(false, Ruled.getValue(), Precision::Confusion());
 
-            for(TopoDS_Wire& wire : wires)   {
+            for (TopoDS_Wire& wire : wires)   {
                  wire.Move(invObjLoc);
                  mkTS.AddWire(wire);
             }
@@ -149,7 +161,7 @@ App::DocumentObjectExecReturn *Loft::execute(void)
         TopoDS_Shape front = getVerifiedFace();
         front.Move(invObjLoc);
         std::vector<TopoDS_Wire> backwires;
-        for(std::vector<TopoDS_Wire>& wires : wiresections)
+        for (std::vector<TopoDS_Wire>& wires : wiresections)
             backwires.push_back(wires.back());
 
         TopoDS_Shape back = Part::FaceMakerCheese::makeFace(backwires);
@@ -158,7 +170,7 @@ App::DocumentObjectExecReturn *Loft::execute(void)
         sewer.SetTolerance(Precision::Confusion());
         sewer.Add(front);
         sewer.Add(back);
-        for(TopoDS_Shape& s : shells)
+        for (TopoDS_Shape& s : shells)
             sewer.Add(s);
 
         sewer.Perform();
@@ -166,7 +178,7 @@ App::DocumentObjectExecReturn *Loft::execute(void)
         //build the solid
         BRepBuilderAPI_MakeSolid mkSolid;
         mkSolid.Add(TopoDS::Shell(sewer.SewedShape()));
-        if(!mkSolid.IsDone())
+        if (!mkSolid.IsDone())
             return new App::DocumentObjectExecReturn("Loft: Result is not a solid");
 
         TopoDS_Shape result = mkSolid.Shape();
@@ -178,7 +190,7 @@ App::DocumentObjectExecReturn *Loft::execute(void)
 
         AddSubShape.setValue(result);
 
-        if(base.IsNull()) {
+        if (base.IsNull()) {
             if (getAddSubType() == FeatureAddSub::Subtractive)
                 return new App::DocumentObjectExecReturn("Loft: There is nothing to subtract from\n");
 
@@ -186,7 +198,7 @@ App::DocumentObjectExecReturn *Loft::execute(void)
             return App::DocumentObject::StdReturn;
         }
 
-        if(getAddSubType() == FeatureAddSub::Additive) {
+        if (getAddSubType() == FeatureAddSub::Additive) {
 
             BRepAlgoAPI_Fuse mkFuse(base, result);
             if (!mkFuse.IsDone())
@@ -204,7 +216,7 @@ App::DocumentObjectExecReturn *Loft::execute(void)
             boolOp = refineShapeIfActive(boolOp);
             Shape.setValue(getSolid(boolOp));
         }
-        else if(getAddSubType() == FeatureAddSub::Subtractive) {
+        else if (getAddSubType() == FeatureAddSub::Subtractive) {
 
             BRepAlgoAPI_Cut mkCut(base, result);
             if (!mkCut.IsDone())
--- a/src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
@@ -212,6 +212,8 @@ bool TaskLoftParameters::referenceSelected(const Gui::SelectionChanges& msg) con
             }
             else if (selectionMode == refRemove) {
                 if (f != refs.end())
+                    // Removing just the object this way instead of `refs.erase` and
+                    // `setValues(ref)` cleanly ensures subnames are preserved.
                     loft->Sections.removeValue(obj);
                 else
                     return false;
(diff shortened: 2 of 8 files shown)


src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
```diff
@@ -414,7 +414,7 @@ bool TaskPipeParameters::referenceSelected(const SelectionChanges& msg) const {
                     return false;
             }
 
-            static_cast<PartDesign::Pipe*>(vp->getObject())->Spine.setValue(
+            static_cast<PartDesign::Pipe*>(vp->getObject())->Spine.setValue(
```
donovaly, line 417: The linebreak has a strange character

src/Mod/PartDesign/Gui/TaskSketchBasedParameters.h
```diff
@@ -61,6 +61,8 @@ class TaskSketchBasedParameters : public PartDesignGui::TaskFeatureParameters,
     QVariant objectNameByLabel(const QString& label, const QVariant& suggest) const;
 
     QString getFaceReference(const QString& obj, const QString& sub) const;
+    QString make2DLabel(const App::DocumentObject* section,
```
donovaly, line 64: Please add a note what the function does. This helps in the IDE a lot since the IDEs will take this info and display it as tooltips

src/Mod/PartDesign/App/FeatureLoft.cpp
```diff
@@ -110,13 +110,24 @@ App::DocumentObjectExecReturn *Loft::execute(void)
         for(TopoDS_Wire& wire : wires)
             wiresections.emplace_back(1, wire);
 
-        for(App::DocumentObject* obj : multisections) {
+        for(auto obj : multisections) {
```
donovaly, line 113: Just a style note: after if, for etc. we have a space
(This occurs at several positions in the PR.)

src/Mod/PartDesign/App/FeatureLoft.cpp
```diff
@@ -110,13 +110,24 @@ App::DocumentObjectExecReturn *Loft::execute(void)
         for(TopoDS_Wire& wire : wires)
             wiresections.emplace_back(1, wire);
 
-        for(App::DocumentObject* obj : multisections) {
+        for(auto obj : multisections) {
             if(!obj->isDerivedFrom(Part::Feature::getClassTypeId()))
                 return  new App::DocumentObjectExecReturn("Loft: All sections need to be part features");
 
+            TopoDS_Shape shape;
```
donovaly, line 117: Please add comments why you do something. Now it is clear, but what about in 2 years when another person looks at it?

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
```diff
@@ -282,6 +281,7 @@ void TaskLoftParameters::indexesMoved()
 
 void TaskLoftParameters::clearButtons() {
 
+    ui->buttonProfileBase->setChecked(false);
```
donovaly, line 284: Why this addition?

donovaly (changes requested):
Many thanks" Looks good so far.
However, please add some comments to help others to understand it in future.

There is only one addition I don't understand yet.

Since PR #5149 is merged, could you maybe rebase your PR since this makes testing easier?

donovaly:
When you make a new PR, please delete the "Thank you for creating a pull request ..." code. This is only a checklist for you.

src/Mod/PartDesign/App/FeatureLoft.cpp
AjinkyaDahale: Thanks for the note, but I think looking at the rest of the document (for example the for loop just above in line 110), I'd like to keep this as is to maintain style consistency.

donovaly: >  I'd like to keep this as is to maintain style consistency.

We try to fix the style whenever we edit code. So when you see a "for(" you can just change it to "for (".

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
AjinkyaDahale: Otherwise, if we press the profile button and change the profile, the button remains pressed. It also remains pressed when the add/remove section buttons are pressed.
Screenshot at 2021-11-08 17-27-52

src/Mod/PartDesign/App/FeatureLoft.cpp
donovaly: OK, leave it as it is. I will do this when it is merged since MSVC does this automatically for me with just few clicks.

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
donovaly: OK, I only don't understand yet why this was not necessary before your PR.

AjinkyaDahale: It was. It was a minor fix so I forgot about it and left it in. It needs some more treatment similar to https://github.com/FreeCAD/FreeCAD/blob/3a5c5576ea269b8bae14325db844e4d9201a7fa4/src/Mod/Fem/Gui/TaskFemConstraintFixed.cpp#L249, because right now if you press, for example "remove" when "add" is pressed, "add" remains pressed. If you want I could make it a separate PR, though.

src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
AjinkyaDahale: It's the carriage return character. Probably a bug in emacs adding windows line ending where it's not needed.

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
donovaly: > because right now if you press, for example "remove" when "add" is pressed, "add" remains pressed. If you want I could make it a separate PR, though.

I see. Yes please make a separate PR. Then I can backport this for 0.19.

src/Mod/PartDesign/App/FeatureLoft.cpp
AjinkyaDahale: Looking at it again seems the entire file uses a mix of space and no space style, so just added the spaces.

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
AjinkyaDahale: OK change removed from here to be put in a separate PR.

donovaly: > OK change removed from here to be put in a separate PR.

Thanks. Please make the PR soon, because to be able to backport easily, I need to put that new PR in before this PR can go in.

donovaly:
There is a bug with the updating mechanism. Here is a screencast:

- at first I make a loft towards a face, then I add a new section. Bot operations are successful (I could have pressed OK and get a valid loft.)
- then I delete a section to get only the loft towards the face -> this is now suddenly invalid
- the second part of the screencast shows that the loft towards the face is fine, so the previous attempt should not have lead to an invalid loft.

qvc8MDHBLK

AjinkyaDahale:
@donovaly the bug should be fixed now.

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
```diff
@@ -250,8 +250,7 @@ void TaskLoftParameters::onDeleteSection()
         App::DocumentObject* obj = loft->getDocument()->getObject(data.constData());
         std::vector<App::DocumentObject*>::iterator f = std::find(refs.begin(), refs.end(), obj);
         if (f != refs.end()) {
-            refs.erase(f);
-            loft->Sections.setValues(refs);
+            loft->Sections.removeValue(obj);
```
donovaly, line 253: Can you please add a comment why you delete the object and not set the changed ref.

This is not straight forward and in two years... you know ;-)

donovaly (changes requested):
many thanks! I tested it well I think and the PR looks good, just add the comment and then this can go in.

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
AjinkyaDahale: Force-pushed.

IMO though deleting individual object sounds like the more intuitive way. If we have a method, why not use it?

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
0penBrain: @donovaly just a note : why require another PR? My own rule of thumb when I mix bug fix and feature enhancement is to have the bug fix in a separate commit (not a separate PR). Isn't that enough for backporting?

adrianinsaval:
since you're working on these tools, any chance of getting loft to point now that we have points in sketcher? datum points and arbitrary geometry vertex could be accepted too

AjinkyaDahale:
> since you're working on these tools, any chance of getting loft to point now that we have points in sketcher? datum points and arbitrary geometry vertex could be accepted too

I'll take a look if I can. I do plan to wander around a bit in these features because there seem to be quite a few reported bugs.

donovaly (approved):

donovaly:
Many thanks! I merged it. Can you please also document it in the Wiki and our release notes?

If you are not familiar with this yet, I can support you or do some of the work. Just tell me.

AjinkyaDahale:
> Many thanks! I merged it. Can you please also document it in the Wiki and our release notes?
> 
> If you are not familiar with this yet, I can support you or do some of the work. Just tell me.

I'd be happy to make the changes. I was trying to put this in the changelog for 0.20 but I can't seem to find the proper page. Have we moved it somewhere else?

AjinkyaDahale:
@donovaly just updated wiki. Please take a look.

wwmayer:
@AjinkyaDahale now that you have replaced the App::PropertyLinkList with App::PropertyXLinkSubList any existing projects with pipes or lofts cannot be fully loaded any more. So, please implement the function handleChangedPropertyType() for both features in order to support old project files.

donovaly:
> @AjinkyaDahale now that you have replaced the App::PropertyLinkList with App::PropertyXLinkSubList any existing projects with pipes or lofts cannot be fully loaded any more. So, please implement the function handleChangedPropertyType() for both features in order to support old project files.

I am to blame too, since as reviewer I should have seen this.

I thought I can quickly add it, but fail. The PR #5165 should do the hob, but apparently there is an underlying bug in the Link handling feature, see my comment in the PR. I spent an hour to figure it out without success and asked now @realthunder to have a look.

donovaly:
@wwmayer I played a bit and astonishingly without any handleChangedPropertyType() handling as we have in current master, I don't see issues. Here is a document I created with FC 0.19:
[lofttest-PR5155.zip](https://github.com/FreeCAD/FreeCAD/files/7514124/lofttest-PR5155.zip)

And this works fine when opening with master. Also other real-life document using lofts I have, work. Do you have a file for me that fails?

AjinkyaDahale:
@wwmayer thanks for pointing out the missing work, and @donovaly thanks for handling it while I was sleeping :grin:.

> @wwmayer I played a bit and astonishingly without any handleChangedPropertyType() handling as we have in current master, I don't see issues. Here is a document I created with FC 0.19: [lofttest-PR5155.zip](https://github.com/FreeCAD/FreeCAD/files/7514124/lofttest-PR5155.zip)
> 
> And this works fine when opening with master. Also other real-life document using lofts I have, work. Do you have a file for me that fails?

I would not be surprised if this works smoothly. IIRC the only thing I _had_ to do after changing `setSize(0)` in `Feature****.cpp` to `setValue(0)`, Beyond that `App::PropertyXLinkSubList` has some commonalities with `App::PropertyLinkList` even when there's no subclass relationship (it'll just not have a sublist).

AjinkyaDahale:
> @wwmayer I played a bit and astonishingly without any handleChangedPropertyType() handling as we have in current master, I don't see issues. Here is a document I created with FC 0.19: [lofttest-PR5155.zip](https://github.com/FreeCAD/FreeCAD/files/7514124/lofttest-PR5155.zip)
> 
> And this works fine when opening with master. Also other real-life document using lofts I have, work. Do you have a file for me that fails?

@donovaly on opening the file the 3D view is correct, but if you notice the `Section` attribute for both the loft and the sweep is empty. So we _do_ need to handle this.

donovaly:
> @donovaly on opening the file the 3D view is correct, but if you notice the `Section` attribute for both the loft and the sweep is empty. So we _do_ need to handle this.

Sure we need to handle this, I was only surprised to see no issues so far. Also for me, the section attribute for the loft and sweep is correct.

However, I hope to get a reply from realthunder soon since he wrote the Link code we rely on.

donovaly:
> Sure we need to handle this

@AjinkyaDahale found the bug that prevented the property upgrade via PR #5167. PR #5165 could therefore be merged. The the missing handling is now in master.

donovaly:
> @donovaly just updated wiki. Please take a look.

Done. This way I found a missing piece:

Loft has the preselection mode described here: https://wiki.freecadweb.org/PartDesign_AdditiveLoft#Selection-based_workflow
So you can preselect 3 sketches, then press the Loft button and get the loft.

This does not yet work for faces_
- select first a face, then two sketches -> works
- select first a sketch, then a face, then a sketch -> does not work, the face is ignored

Could you please have a look?

AjinkyaDahale:
> > @donovaly just updated wiki. Please take a look.
> 
> Done. This way I found a missing piece:
> 
> Loft has the preselection mode described here: https://wiki.freecadweb.org/PartDesign_AdditiveLoft#Selection-based_workflow So you can preselect 3 sketches, then press the Loft button and get the loft.
> 
> This does not yet work for faces_
> 
>     * select first a face, then two sketches -> works
> 
>     * select first a sketch, then a face, then a sketch -> does not work, the face is ignored
> 
> 
> Could you please have a look?

Please see PR #5168.

wwmayer:
> And this works fine when opening with master. Also other real-life document using lofts I have, work. Do you have a file for me that fails?

Without these changes the Sections property of e.g. the loft is empty. If you touch the loft and recompute the document you will get an error that: At least one section is needed

donovaly:
> Without these changes the Sections property of e.g. the loft is empty. If you touch the loft and recompute the document you will get an error that: At least one section is needed

I wast surprised that it worked for me - I did not get this error while AjinkyaDahale got it.
However, we already found the issue and the fix is in: https://github.com/FreeCAD/FreeCAD/pull/5155#issuecomment-965883978

wwmayer:
> I wast surprised that it worked for me

I am not sure you tested the right thing. When the Sections property is not restored then a recompute of the loft fails and the sweep will look different because a section has been removed.

Roy-043:
This PR has perhaps introduced an issue where sketches outside the body can be used as sections. I wonder if this is intended?
Forum topic:
https://forum.freecadweb.org/viewtopic.php?f=3&t=63835

AjinkyaDahale:
> This PR has perhaps introduced an issue where sketches outside the body can be used as sections. I wonder if this is intended? Forum topic: https://forum.freecadweb.org/viewtopic.php?f=3&t=63835

Replied in thread https://forum.freecadweb.org/viewtopic.php?f=3&t=63835&p=547606#p547606.

PR #5156, opened 2021-11-08
[Tux] fix navigation status bar width

donovaly:
- the PR adjusts the size of the navigation style status bar so that it is wide enough to contain the text + the space for the combobox button
- also sort the entries since as it is, it just reflects the order the styles were added in the development

With the fix:
FreeCAD_ft8GCErtQH

Without the fix:
bqxafRJiwA

The size adjustment is done using Qt's adjustSize() command and the space consumed be the combobox dropdown indicator is done by adding two spaces to the button text since on different OSes the necessary space might be different.

### DIFF
--- a/src/Mod/Tux/NavigationIndicatorGui.py
+++ b/src/Mod/Tux/NavigationIndicatorGui.py
@@ -53,6 +53,7 @@ class IndicatorButton(QtGui.QPushButton):
         if event.type() == QtCore.QEvent.LanguageChange:
             retranslateUi()
             onTooltip()
+            adjustSize()
         return super(IndicatorButton, self).changeEvent(event)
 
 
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5156">pipeline status</a> for feature branch [PR_5156](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5156). Pipeline [406327320](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/406327320) was triggered at [d7d23d5](https://github.com/FreeCAD/FreeCAD/pull/5156/commits/d7d23d5ef596a77e0debc267f00e7e3db06ef07e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

yorikvanhavre:
Ideally we could replace the button.setText("Title  ") (with 2 spaces) with something like self.setButtonText(button,"Title") for elegance (and ease of re-fix if some later Qt version changes something), but it's a good solution for now I think

PR #5158, opened 2021-11-09
[PD] fix sweep visibility bug

donovaly:
Currently:
- only sweep path is toggled correctly in its visibility, the profile is never shown
- the sections are always shown

This PR fixes this inconsistency reported here: https://forum.freecadweb.org/viewtopic.php?f=19&t=63544

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
@@ -1042,6 +1042,8 @@ bool TaskPipeScaling::referenceSelected(const SelectionChanges& msg) const {
         }
         else {
             if (f != refs.end())
+                // Removing just the object this way instead of `refs.erase` and
+                // `setValues(ref)` cleanly ensures subnames are preserved.
                 pipe->Sections.removeValue(obj);
             else
                 return false;
@@ -1082,8 +1084,9 @@ void TaskPipeScaling::onDeleteSection()
 
         // if something was found, delete it and update the section list
         if (f != refs.end()) {
-            refs.erase(f);
-            pipe->Sections.setValues(refs);
+            // Removing just the object this way instead of `refs.erase` and
+            // `setValues(ref)` cleanly ensures subnames are preserved.
+            pipe->Sections.removeValue(obj);
             clearButtons();
             recomputeFeature();
         }
--- a/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
@@ -118,7 +118,7 @@ TaskPipeParameters::TaskPipeParameters(ViewProviderPipe *PipeView, bool /*newObj
 
     //add initial values
     if (pipe->Profile.getValue())
-        ui->profileBaseEdit->setText(QString::fromUtf8(pipe->Profile.getValue()->Label.getValue()));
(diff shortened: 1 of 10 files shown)


AjinkyaDahale:
I think you got it wrong: profile is handled correctly, but sweep path is not. With these changes, the sections are hidden on confirming. However, on undoing or deleting the sweep, only the profile is correctly made visible. The path and sections remain invisible.

src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
```diff
@@ -147,13 +149,17 @@ TaskPipeParameters::~TaskPipeParameters()
             PartDesign::Pipe* pipe = static_cast<PartDesign::Pipe*>(vp->getObject());
             Gui::Document* doc = vp->getDocument();
 
-            //make sure the user sees all important things: the
-            //spine/auxiliary spine he already selected
+            // hide/show the elements when they were hidden/show before the pipe was opened
```
luzpaz, line 152: suggest:  
`// show/hide the elements when they were shown/hidden before the pipe was opened`

src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
```diff
@@ -108,20 +108,22 @@ TaskPipeParameters::TaskPipeParameters(ViewProviderPipe *PipeView, bool /*newObj
     PartDesign::Pipe* pipe = static_cast<PartDesign::Pipe*>(PipeView->getObject());
     Gui::Document* doc = PipeView->getDocument();
 
-    //make sure the user sees all important things: the
-    //spine/auxiliary spine he already selected
+    // make sure the user sees all important things and load the values
+    // the spine
```
luzpaz, line 112: suggest:
```
// make sure the user sees all the important things and load the values:
// starting with the spine
```

luzpaz (changes requested):

src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
donovaly: Thanks, I improved the comments now. One was even wrong.

donovaly:
> I think you got it wrong: profile is handled correctly, but sweep path is not.

No. At the moment the profile is not handled at all. The main thing I did is to add the code that exists already fro the spine to the profile.
And the spine handling is sensible:
- one checks the visibility status when the pipe is opened
- the spine is always shown
- on closing the pipe, the spine it kept visible or made invisible, depending on how it was before opening the pipe

With this PR this is now also the case with the profile. And that the profile was not made visible is annoying since you had to do this manually to keep the overview.

> With these changes, the sections are hidden on confirming.

Yes. There is no other handling possible since we cannot check their visibility status before the pipe was opened. The sections might have changed, deleted, more sections added or their number reduced.
So we have only 2 options: showing them on closing the pipe or opening them. The other sketch-based features like loft hide the sketches so it is consistent to hide them on closing for pipe too.

> However, on undoing or deleting the sweep, only the profile is correctly made visible.

I cannot reproduce this. I rebased my PR to current master (including your PR that I just merged). I get correctly this:

yJTUa3csJH

I also had a look at the code and cannot find that a deletion changes the visibility. When the sketches were invisible when the deletion was performed, they are correctly kept visible - like for the other PD sketch based objects like lofts.

donovaly:
Sorry, I killed this PR and cannot repair it. Closing it i  favor of the new #5163

PR #5159, opened 2021-11-09
[PD] Clear other taskbox buttons in loft and pipe when one pressed

AjinkyaDahale:
Without this, when you press, for example, the "Add Section" button, do nothing, then press "Remove Section" button, the first button remains pressed.

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
@@ -285,14 +285,18 @@ void TaskLoftParameters::indexesMoved()
     updateUI();
 }
 
-void TaskLoftParameters::clearButtons() {
-
-    ui->buttonRefAdd->setChecked(false);
-    ui->buttonRefRemove->setChecked(false);
+void TaskLoftParameters::clearButtons(const selectionModes notThis)
+{
+    if (notThis != refProfile)
+        ui->buttonProfileBase->setChecked(false);
+    if (notThis != refAdd)
+        ui->buttonRefAdd->setChecked(false);
+    if (notThis != refRemove)
+        ui->buttonRefRemove->setChecked(false);
 }
 
-void TaskLoftParameters::exitSelectionMode() {
-
+void TaskLoftParameters::exitSelectionMode()
+{
     selectionMode = none;
     Gui::Selection().clearSelection();
 }
@@ -314,23 +318,27 @@ void TaskLoftParameters::onRuled(bool val) {
 void TaskLoftParameters::onProfileButton(bool checked)
 {
     if (checked) {
+        clearButtons(refProfile);
         Gui::Selection().clearSelection();
         selectionMode = refProfile;
         //static_cast<ViewProviderLoft*>(vp)->highlightReferences(true, true);
     }
 }
 
-void TaskLoftParameters::onRefButtonAdd(bool checked) {
+void TaskLoftParameters::onRefButtonAdd(bool checked)
+{
     if (checked) {
+        clearButtons(refAdd);
         Gui::Selection().clearSelection();
         selectionMode = refAdd;
         //static_cast<ViewProviderLoft*>(vp)->highlightReferences(true, true);
     }
 }
 
-void TaskLoftParameters::onRefButtonRemove(bool checked) {
-
+void TaskLoftParameters::onRefButtonRemove(bool checked)
+{
     if (checked) {
+        clearButtons(refRemove);
         Gui::Selection().clearSelection();
         selectionMode = refRemove;
         //static_cast<ViewProviderLoft*>(vp)->highlightReferences(true, true);
--- a/src/Mod/PartDesign/Gui/TaskLoftParameters.h
+++ b/src/Mod/PartDesign/Gui/TaskLoftParameters.h
@@ -63,6 +63,8 @@ private Q_SLOTS:
     void indexesMoved();
 
 protected:
+    enum selectionModes { none, refAdd, refRemove, refProfile };
+
     void changeEvent(QEvent *e);
 
 private:
@@ -70,14 +72,13 @@ private:
     void updateUI();
     bool referenceSelected(const Gui::SelectionChanges& msg) const;
     void removeFromListWidget(QListWidget*w, QString name);
-    void clearButtons();
+    void clearButtons(const selectionModes notThis=none);
     void exitSelectionMode();
 
 private:
     QWidget* proxy;
     std::unique_ptr<Ui_TaskLoftParameters> ui;
 
-    enum selectionModes { none, refAdd, refRemove, refProfile };
     selectionModes selectionMode = none;
 };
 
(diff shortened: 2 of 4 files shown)


AjinkyaDahale:
@donovaly that PR you requested. We still need to figure out how to link all the buttons in pipe though. Also, perhaps we could use [`QButtonGroup`](https://doc.qt.io/archives/qt-5.6/qbuttongroup.html#details) to streamline this process a little.

donovaly:
Many thanks.

As funny as it is I made the same fix today for sweep and now I see your PR. So fine with me, however I cannot merge it because of merge conflicts. Can you please resolve them with current master?

> We still need to figure out how to link all the buttons in pipe though. Also, perhaps we could use QButtonGroup to streamline this process a little.

QButtonGroup only works for one UI, not across several UIs. Correct me when I am wrong.

However, although this might be strange, it is basically correct since you are allowed to use the same sketch for the spine and section:

feqELZdT0H

AjinkyaDahale:
> Many thanks.
> 
> As funny as it is I made the same fix today for sweep and now I see your PR. So fine with me, however I cannot merge it because of merge conflicts. Can you please resolve them with current master?
> 

Funny, but I suppose it makes you more appropriate to verify the PR :). I have fixed the merge conflicts.

> > We still need to figure out how to link all the buttons in pipe though. Also, perhaps we could use QButtonGroup to streamline this process a little.
> 
> QButtonGroup only works for one UI, not across several UIs. Correct me when I am wrong.
> 

As it is implemented all the task boxes are contained within the same task _dialog_ (please correct my terminology if I'm wrong). So something like the following works.

```
TaskDlgPipeParameters::TaskDlgPipeParameters(ViewProviderPipe *PipeView,bool newObj)
   : TaskDlgSketchBasedParameters(PipeView)
{
    assert(PipeView);
    parameter    = new TaskPipeParameters(PipeView,newObj);
    orientation  = new TaskPipeOrientation(PipeView,newObj);
    scaling      = new TaskPipeScaling(PipeView,newObj);

    Content.push_back(parameter);
    Content.push_back(orientation);
    Content.push_back(scaling);

    QButtonGroup *buttonGroup = new QButtonGroup();
    buttonGroup->setExclusive(true);

    buttonGroup->addButton(parameter->ui->buttonProfileBase);
    buttonGroup->addButton(parameter->ui->buttonRefAdd);
    buttonGroup->addButton(parameter->ui->buttonRefRemove);
    buttonGroup->addButton(parameter->ui->buttonSpineBase);

    buttonGroup->addButton(orientation->ui->buttonRefAdd);
    buttonGroup->addButton(orientation->ui->buttonRefRemove);
    buttonGroup->addButton(orientation->ui->buttonProfileBase);

    buttonGroup->addButton(scaling->ui->buttonRefAdd);
    buttonGroup->addButton(scaling->ui->buttonRefRemove);
}
```

I tested this in a stash on this branch and this works, though I haven't yet extensively checked the order in which the toggles work. Like with this (and removing the `clearButtons` calls from individual buttons) if I press "Add" and the add flag is set, but by toggling, say, "Remove", all flags are cleared nothing will work.

> However, although this might be strange, it is basically correct since you are allowed to use the same sketch for the spine and section:

That may be true, but if someone does want to use the same sketch for spine and section, they can do it exclusively

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5159">pipeline status</a> for feature branch [PR_5159](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5159). Pipeline [406343197](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/406343197) was triggered at [6d2edfd](https://github.com/FreeCAD/FreeCAD/pull/5159/commits/6d2edfd745ef837e0574399b1838c32667cf8acc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> > QButtonGroup only works for one UI, not across several UIs. Correct me when I am wrong.
> 
> As it is implemented all the task boxes are contained within the same task _dialog_ (please correct my terminology if I'm wrong). So something like the following works.

But this is what I meant - I think it only works within one dialog. The pipe task starts 3 dialogs and ideally we should assure that only one button is pressed, no matter in what of the 3 dialogs.

I'm putting this PR in and if you find a solution to check the button state across the 3 dialogs, please make a further PR to test.

donovaly:
Only for information: I don't want to backport this fix since it is cosmetic and we never got a complaint about it so far. Thus it is not that pressing in my opinion.

AjinkyaDahale:
> But this is what I meant - I think it only works within one dialog. The pipe task starts 3 dialogs and ideally we should assure that only one button is pressed, no matter in what of the 3 dialogs.

Correction: the pipe task starts 1 dialog with 3 task boxes. So the dialog works as the "common UI" for the buttons.

donovaly:
> Correction: the pipe task starts 1 dialog with 3 task boxes. So the dialog works as the "common UI" for the buttons.

Fine. Then let's try this out.

AjinkyaDahale:
> > Correction: the pipe task starts 1 dialog with 3 task boxes. So the dialog works as the "common UI" for the buttons.
> 
> Fine. Then let's try this out.

Tried it out. Pressing one button toggles the other off, but unfortunately there doesn't appear to be a way to toggle all buttons off. 
From https://doc.qt.io/qt-5/qbuttongroup.html#exclusive-prop
> In an exclusive group, the user cannot uncheck the currently checked button by clicking on it; instead, another button in the group must be clicked to set the new checked button for that group.

wwmayer:
A solution is to make our own button group that allows to have selected no button in exclusive mode. A solution is provided here (last answer): https://stackoverflow.com/questions/15177771/alternative-to-qbuttongroup-that-allows-no-selection

What's still not working is when I press a button twice e.g. "Add section" in the loft dialog then internally the mode is still set to refAdd (instead of none) while the button is unchecked. Now when I click on a sketch it will be added to the loft but this is unexpected behaviour.

AjinkyaDahale:
> A solution is to make our own button group that allows to have selected no button in exclusive mode. A solution is provided here (last answer): https://stackoverflow.com/questions/15177771/alternative-to-qbuttongroup-that-allows-no-selection
> 
> What's still not working is when I press a button twice e.g. "Add section" in the loft dialog then internally the mode is still set to refAdd (instead of none) while the button is unchecked. Now when I click on a sketch it will be added to the loft but this is unexpected behaviour.

Interesting. I'll need to check this out but perhaps it's because toggled is not called when a button is turned off.

AjinkyaDahale:
@wwmayer which implementation did you use from the SO answers? Could you share it if you have it ready?

wwmayer:
@AjinkyaDahale it's the very last answer that suggests to create a sub-class myQButtonGroup. This is the most elegant solution because it doesn't require to create custom buttons.

> Interesting. I'll need to check this out but perhaps it's because toggled is not called when a button is turned off.

The reason simply is that the slot function e.g. onRefButtonAdd() only handles the case if the button is toggled but not if it's untoggled. So, it should be checked if with a custom button group we can solve all the issues and simplify the code.

AjinkyaDahale:
> The reason simply is that the slot function e.g. onRefButtonAdd() only handles the case if the button is toggled but not if it's untoggled. So, it should be checked if with a custom button group we can solve all the issues and simplify the code.

I suppose one solution could be to unify the enums used within all the task boxes `TaskPipe****:selectionModes` into a common enum in `TaskDlgPipe...` and use that as the id for the buttons in our custom button group.

PR #5163, opened 2021-11-09
[PD] fix sweep visibility bug

donovaly:
Currently:
- only sweep path is toggled correctly in its visibility, the profile is never shown
- the sections are always shown

This PR fixes this inconsistency reported here: https://forum.freecadweb.org/viewtopic.php?f=19&t=63544

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
@@ -108,20 +108,23 @@ TaskPipeParameters::TaskPipeParameters(ViewProviderPipe *PipeView, bool /*newObj
     PartDesign::Pipe* pipe = static_cast<PartDesign::Pipe*>(PipeView->getObject());
     Gui::Document* doc = PipeView->getDocument();
 
-    //make sure the user sees all important things: the
-    //spine/auxiliary spine he already selected
+    // make sure the user sees all important things and load the values
+    // also save visibility state to reset it later when pipe is closes
+    // first the spine
     if (pipe->Spine.getValue()) {
-        auto* svp = doc->getViewProvider(pipe->Spine.getValue());
-        spineShow = svp->isShow();
-        svp->setVisible(true);
-    }
-
-    //add initial values
-    if (pipe->Profile.getValue())
-        ui->profileBaseEdit->setText(make2DLabel(pipe->Profile.getValue(), pipe->Profile.getSubValues()));
-    if (pipe->Spine.getValue())
+        auto* spineVP = doc->getViewProvider(pipe->Spine.getValue());
+        spineShow = spineVP->isShow();
+        spineVP->setVisible(true);
         ui->spineBaseEdit->setText(QString::fromUtf8(pipe->Spine.getValue()->Label.getValue()));
-
+    }
+    // the profile
+    if (pipe->Profile.getValue()) {
+        auto* profileVP = doc->getViewProvider(pipe->Profile.getValue());
+        profileShow = profileVP->isShow();
+        profileVP->setVisible(true);
+        ui->profileBaseEdit->setText(make2DLabel(pipe->Profile.getValue(), pipe->Profile.getSubValues()));
+    }
+    // the spine edges
     std::vector<std::string> strings = pipe->Spine.getSubValues();
     for (std::vector<std::string>::const_iterator it = strings.begin(); it != strings.end(); ++it) {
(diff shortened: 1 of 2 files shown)


AjinkyaDahale:
@donovaly in the previous PR you said:

> I also had a look at the code and cannot find that a deletion changes the visibility. When the sketches were invisible when the deletion was performed, they are correctly kept visible [sic] - like for the other PD sketch based objects like lofts.

I believe you meant "kept invisible"? If that's the case, I don't agree that this should be the expected behavior. As such, sketches should go back to being visible again, or ideally put in their own visibility.

I got the behavior (visible profile) by undoing rather than deleting the sweep.

donovaly:
> I believe you meant "kept invisible"?

I mean when deleting objects in PD the underlying sketches keep the visibility status, when they are visible, they stay visible, when then are invisible, they stay invisible. And sweep does not acct differently.

> If that's the case, I don't agree that this should be the expected behavior. As such, sketches should go back to being visible again, or ideally put in their own visibility.

Ideally yes, But this needs to changed then for all sketch-based PD objects.

> I got the behavior (visible profile) by undoing rather than deleting the sweep.

Undo is still a mystery to me. There are many small issues around with it. Since I did not yet understand how Undo works, I cannot fix them, here if for example a failed attempt from me: https://forum.freecadweb.org/viewtopic.php?f=10&t=63530

donovaly:
Does it the job it should do your you as well?
(The PR does not change anything to the undo mechanism.)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5163">pipeline status</a> for feature branch [PR_5163](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5163). Pipeline [406343205](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/406343205) was triggered at [e762bdc](https://github.com/FreeCAD/FreeCAD/pull/5163/commits/e762bdcd1aa068e74447e5351cd17cd066d73d16). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
> Does it the job it should do your you as well? (The PR does not change anything to the undo mechanism.)

I'm not sure I get what you mean here.

donovaly:
> > Does it the job it should do your you as well? (The PR does not change anything to the undo mechanism.)
> 
> I'm not sure I get what you mean here.

This:
- no matter if the sketches used in the pipe are visible or not when the Pipe is open, they will be shown when the Pipe is open
- When the pipe is closed, the profile and path sketch gets the visibility status they had before the pipe opening
- When the pipe is closed, the section sketches are all made invisible

AjinkyaDahale:
> > > Does it the job it should do your you as well? (The PR does not change anything to the undo mechanism.)
> > 
> > 
> > I'm not sure I get what you mean here.
> 
> This:
> 
>     * no matter if the sketches used in the pipe are visible or not when the Pipe is open, they will be shown when the Pipe is open
> 
>     * When the pipe is closed, the profile and path sketch gets the visibility status they had before the pipe opening
> 
>     * When the pipe is closed, the section sketches are all made invisible

Can confirm these 3 things work.

donovaly:
> Can confirm these 3 things work.

Thanks. @chennes if you agree the PR is OK, can you please merge it?

Since the fixes are cosmetic more or less, I don't plan to backport this PR.

donovaly:
Oh, there is now a merge conflict -> fixing it now

chennes:
OK, I think this is fine to merge then. Probably the undo issue just needs to be put into tracker, if it's not already there.

wwmayer:
This makes undo/redo working now: fb9a3eecff

The problem was that visibility was changed inside the destructor but at this point no transaction is active any more and thus changes cannot be recorded. So, I moved the code to a single function and call it inside the accept() method.

donovaly:
> The problem was that visibility was changed inside the destructor but at this point no transaction is active any more and thus changes cannot be recorded.

Many thanks!

As I wrote I miss info about undo, for example that actions in the destructor are not recorded. Here is another undo case where I failed: https://forum.freecadweb.org/viewtopic.php?f=10&t=63530

Could you maybe write in the forum threads the basics about the auto-transaction mechanism? I would the transfer this info to a Wiki page.

adrianinsaval:
Thank you guys for all the recent improvements in PD, you're awesome!

PR #5164, opened 2021-11-10
[PD] add feature to set direction for pockets

donovaly:
This PR adds the functionality of pads to pad along either a custom direction or along an edge to pockets.

So with this PR Pad and Pocket are on the same level of functionality.

Here is a screencast of what packet can now do:
MSJAOB41G2

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePocket.cpp
+++ b/src/Mod/PartDesign/App/FeaturePocket.cpp
@@ -67,6 +67,10 @@ Pocket::Pocket()
     Type.setEnums(TypeEnums);
     ADD_PROPERTY_TYPE(Length,(100.0),"Pocket",App::Prop_None,"Pocket length");
     ADD_PROPERTY_TYPE(Length2,(100.0),"Pocket",App::Prop_None,"P");
+    ADD_PROPERTY_TYPE(UseCustomVector, (false), "Pocket", App::Prop_None, "Use custom vector for pocket direction");
+    ADD_PROPERTY_TYPE(Direction, (Base::Vector3d(1.0, 1.0, 1.0)), "Pocket", App::Prop_None, "Pocket direction vector");
+    ADD_PROPERTY_TYPE(ReferenceAxis, (0), "Pocket", App::Prop_None, "Reference axis of direction");
+    ADD_PROPERTY_TYPE(AlongSketchNormal, (true), "Pocket", App::Prop_None, "Measure pocket length along the sketch normal direction");
     ADD_PROPERTY_TYPE(UpToFace,(0),"Pocket",App::Prop_None,"Face where pocket will end");
     ADD_PROPERTY_TYPE(Offset,(0.0),"Pocket",App::Prop_None,"Offset from face in which pocket will end");
     static const App::PropertyQuantityConstraint::Constraints signedLengthConstraint = {-DBL_MAX, DBL_MAX, 1.0};
@@ -84,6 +88,10 @@ short Pocket::mustExecute() const
         Length.isTouched() ||
         Length2.isTouched() ||
         Offset.isTouched() ||
+        UseCustomVector.isTouched() ||
+        Direction.isTouched() ||
+        ReferenceAxis.isTouched() ||
+        AlongSketchNormal.isTouched() ||
         UpToFace.isTouched())
         return 1;
     return ProfileBased::mustExecute();
@@ -130,6 +138,7 @@ App::DocumentObjectExecReturn *Pocket::execute(void)
 
     // get the Sketch plane
     Base::Placement SketchPos    = obj->Placement.getValue();
+    // get the normal vector of the sketch
     Base::Vector3d  SketchVector = getProfileNormal();
 
     // turn around for pockets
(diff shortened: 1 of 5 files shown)


src/Mod/PartDesign/App/FeaturePocket.cpp
```diff
@@ -141,7 +150,78 @@ App::DocumentObjectExecReturn *Pocket::execute(void)
 
         base.Move(invObjLoc);
 
-        gp_Dir dir(SketchVector.x,SketchVector.y,SketchVector.z);
+        Base::Vector3d pocketDirection;
+
+        if (!UseCustomVector.getValue()) {
+            if (ReferenceAxis.getValue() == nullptr) {
```
chennes, line 156: Why switch syntax between these two conditionals? Core Guidelines prefer your style on 155 to the style on 156, but is there some reason to really highlight the `nullptr` comparison on 156?

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
```diff
@@ -291,6 +353,193 @@ void TaskPocketParameters::onOffsetChanged(double len)
...
+    bool oldVal_blockUpdate = blockUpdate;
+    blockUpdate = true;
+
+    if (axesInList.empty()) {
+        ui->directionCB->clear();
+        // add sketch normal
+        PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
+        Part::Part2DObject* pcSketch = dynamic_cast<Part::Part2DObject*>(pcFeat->Profile.getValue());
+        if (pcSketch)
+            addAxisToCombo(pcSketch, "N_Axis", QObject::tr("Sketch normal"));
```
chennes, line 367: Why qualify `tr` here? You don't on the next line.

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
```diff
@@ -291,6 +353,193 @@ void TaskPocketParameters::onOffsetChanged(double len)
...
+
+    try {
+        recomputeFeature();
+    }
+    catch (const Base::Exception& e) {
+        e.ReportException();
+    }
+
+    // disable AlongSketchNormal when the direction is already normal
+    if (num == 0)
```
chennes, line 456: Is there a way to refactor to use an `enum` for these `num` comparisons, or even just some named constants, instead of bare numbers here?

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
```diff
@@ -291,6 +353,193 @@ void TaskPocketParameters::onOffsetChanged(double len)
...
+        ui->checkBoxAlongDirection->setEnabled(false);
+    else
+        ui->checkBoxAlongDirection->setEnabled(true);
+    // if custom direction is used, show it
+    if (num == 2) {
+        ui->checkBoxDirection->setChecked(true);
+        PartDesign::Pocket* pcPocket = static_cast<PartDesign::Pocket*>(vp->getObject());
```
chennes, line 463: This line looks redundant to me.

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
```diff
@@ -291,6 +353,193 @@ void TaskPocketParameters::onOffsetChanged(double len)
...
+    this->axesInList.emplace_back(new App::PropertyLinkSub);
+    App::PropertyLinkSub& lnk = *(axesInList.back());
+    lnk.setValue(linkObj, std::vector<std::string>(1, linkSubname));
+}
+
+void TaskPocketParameters::onDirectionCBChanged(int num)
+{
+    PartDesign::Pocket* pcPocket = static_cast<PartDesign::Pocket*>(vp->getObject());
+
+    if (axesInList.empty() || !pcPocket)
```
chennes, line 426: Is it actually possible for `pcPocket` to be null here? You have this same basic structure in several other callbacks in which you don't check for null.

chennes (approved):
Good feature addition, and excellent code. I've left a few comments, but no deal-breakers. I will leave to you to merge when you feel like it's ready to go, but overall this looks great.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5164">pipeline status</a> for feature branch [PR_5164](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5164). Pipeline [406343211](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/406343211) was triggered at [4d67b84](https://github.com/FreeCAD/FreeCAD/pull/5164/commits/4d67b8424f4183a9d449612520f3e9beac8c6110). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeaturePocket.cpp
donovaly: The first one is a bool, the second one a pointer therefore I checked for a nullptr to check that there is no axis. What are the guidelines in this case?
(I really miss a defined guideline that I can check myself.)

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
donovaly: Thanks. Fixed now

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
donovaly: Indeed. Thanks. Fixed.

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
donovaly: > Is it actually possible for pcPocket to be null here?

No. So this can go for consistency reasons. Done now.

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
donovaly: The `num` is an int because I number the possible settings according to the combobox entries:
0 - sketch normal
1 - entry to select a reference
2 - custom direction
3, 4, 5... - the different edges the user had selected in the 3D model

For the latter I refer to the int to get the particular axis. Since the list of edges has no limit, I cannot use an enum.
Do you have another idea?

src/Mod/PartDesign/App/FeaturePocket.cpp
chennes: I'm referring to: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-if

donovaly: > I'm referring to: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-if

Many thanks. So can we just use these guidelines for FC? Should I start a forum thread to make this decision?

I improved the code now accordingly.

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
chennes: You could still use an enum if you wanted to, it's not really much different than defining three independent constants. Maybe something like:
```
enum class ComboboxMapping {
  SKETCH_NORMAL,
  SELECTED_REFERENCE,
  CUSTOM_DIRECTION,
  SELECTED_EDGES_OFFSET // Could omit this one if there's no use for it -- do you ever add/subtract 3 to get the index?
};
```
You could also just:
```
const int SKETCH_NORMAL = 0;
const int SELECTED_REFERENCE = 1;
// etc.
```

donovaly:
> I will leave to you to merge when you feel like it's ready to go, but overall this looks great.

Many thanks for your review!
I improved the PR. However, please decide if it can go in and if so, please merge.

src/Mod/PartDesign/App/FeaturePocket.cpp
donovaly: I meant to make the style guideline mandatory. I remember we had a discussion this year and this guideline was mentioned by Werner as the one he follows too. But it would be good to have this info for new developers.

So what I can do is to add it to the Wiki and writ to the forum. OK?

src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
donovaly: I don't understand yet. What I do is tho read the current combobox entry. I take the entry index as number and this is the also the index number for the axes list

> do you ever add/subtract 3 to get the index?

No. Since I return the current combobox index, I already have this index number ans the index number of the axes list.

So assuming there are 4 axes from the 3D model, then the combo has 7 entries. When the user sets the combo to index number 5, I know this is the axis axesInList[5], see line 429.

chennes: Maybe a more concrete example: I'd rather see code like this:
```
if (num == SKETCH_NORMAL)
```
or this:
```
if (num == ComboboxMapping::SKETCH_NORMAL)
```
than this
```
if (num == 0)
```
It's not critical, it just makes it a little easier to read and reason about, and is self-documenting.

PR #5165, opened 2021-11-10
[PD] handle changed property for loft and pipe

donovaly:
- this was missing in PR #5155

This should be correct, but it crashes FreeCAD. The reason for the crash is line https://github.com/FreeCAD/FreeCAD/blob/master/src/App/PropertyLinks.cpp#L4322
When I uncomment it, it does not crash, but it does also not to the property value restoration.

@realthunder Can you please have a look? I cannot find the cause.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureLoft.cpp
+++ b/src/Mod/PartDesign/App/FeatureLoft.cpp
@@ -256,3 +256,14 @@ PROPERTY_SOURCE(PartDesign::SubtractiveLoft, PartDesign::Loft)
 SubtractiveLoft::SubtractiveLoft() {
     addSubType = Subtractive;
 }
+
+void Loft::handleChangedPropertyType(Base::XMLReader& reader, const char* TypeName, App::Property* prop)
+{
+    // property Sections had the App::PropertyLinkList and was changed to App::PropertyXLinkSubList
+    if (prop == &Sections && strcmp(TypeName, "App::PropertyLinkList") == 0) {
+        Sections.upgrade(reader, TypeName);
+    }
+    else {
+        ProfileBased::handleChangedPropertyType(reader, TypeName, prop);
+    }
+}
--- a/src/Mod/PartDesign/App/FeatureLoft.h
+++ b/src/Mod/PartDesign/App/FeatureLoft.h
@@ -54,6 +54,10 @@ public:
     }
     //@}
 
+protected:
+    // handle changed property
+    virtual void handleChangedPropertyType(Base::XMLReader& reader, const char* TypeName, App::Property* prop);
+
 private:
     //static const char* TypeEnums[];
     //static const char* SideEnums[];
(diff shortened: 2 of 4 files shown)


donovaly:
@realthunder here is a FC file to trigger the crash when you apply the PR to your master:

[lofttest-PR5155.zip](https://github.com/FreeCAD/FreeCAD/files/7514124/lofttest-PR5155.zip)

AjinkyaDahale:
The crash was actually in line https://github.com/FreeCAD/FreeCAD/blob/d86a73f07ef2ad631f88359e5064c5ae6f8ef3f0/src/Mod/PartDesign/Gui/ViewProviderAddSub.cpp#L238. The nameless temp property somehow reaches the view-provider and that caused the issue. Pushed PR #5167 to fix.

donovaly:
> The nameless temp property somehow reaches the view-provider and that caused the issue. Pushed PR #5167 to fix.

Many thanks. I merged that, and now this PR works as it should -> merging

The changed property type handling should now be done.

realthunder:
Looks like you guys have already solved problem. Just a question @AjinkyaDahale, are there any reason prefer `PropertyLinkSubList` over `PropertyXLinkSubList`? `PropertyXLinkSubList` added functionality to link object from external document, but `PartDesign` features don't even support linking outside its body.

donovaly:
>  are there any reason prefer `PropertyLinkSubList` over `PropertyXLinkSubList`?

But he uses now PropertyXLinkSubList for the sections in Loft and Pipe.

AjinkyaDahale:
> Looks like you guys have already solved problem. Just a question @AjinkyaDahale, are there any reason prefer `PropertyLinkSubList` over `PropertyXLinkSubList`? `PropertyXLinkSubList` added functionality to link object from external document, but `PartDesign` features don't even support linking outside its body.

I am currently using `PropertyXLinkSubList`. I suppose just `PropertyLinkSubList` could do the job, and is likely more appropriate, but it does not have `upgrade` implemented for itself which needs to be done before it can be used.

wwmayer:
> I am currently using PropertyXLinkSubList. I suppose just PropertyLinkSubList could do the job

@AjinkyaDahale with b2475b228a there is now an upgrade() method in PropertyLinkSubList, too. So, I suggest to do the replacement.

AjinkyaDahale:
> @AjinkyaDahale with [b2475b2](https://github.com/FreeCAD/FreeCAD/commit/b2475b228a006aab071704f8e57187f83359bb1f) there is now an upgrade() method in PropertyLinkSubList, too. So, I suggest to do the replacement.

@wwmayer thanks for implementing the `PropertyLinkSubList::upgrade` method. Unfortunately, I was naively assuming just replacing `PropertyXLinkSubList` with `PropertyLinkSubList` will wotk. It appears it needs some more methods, like `addValue` that I was using. I'll take a look if they can also be implemented or replaced.

wwmayer:
> It appears it needs some more methods, like addValue that I was using

0469e277ab

AjinkyaDahale:
@wwmayer 

> @AjinkyaDahale with [b2475b2](https://github.com/FreeCAD/FreeCAD/commit/b2475b228a006aab071704f8e57187f83359bb1f) there is now an upgrade() method in PropertyLinkSubList, too. So, I suggest to do the replacement.

> [0469e27](https://github.com/FreeCAD/FreeCAD/commit/0469e277ab672d8ecfb54a4a25ef6bd55425bcf5)

See PR #5176.

PR #5167, opened 2021-11-11
[PD] Check if property is named before `strcmp`

AjinkyaDahale:
`ViewProviderAddSub::updateData` assumed that the property provided was named.


donovaly:
Many thanks!
It is interesting that the MSVC debugger threw the exception earlier. Therefore I did not reach the actual causing code.

PR #5168, opened 2021-11-11
[PD] Allow body faces when making loft by pre-selection

AjinkyaDahale:
Support for solid faces while building loft by pre-selecting profile and sections was missing in PR #5155.

### DIFF
--- a/src/Mod/PartDesign/Gui/Command.cpp
+++ b/src/Mod/PartDesign/Gui/Command.cpp
@@ -1019,10 +1019,12 @@ void prepareProfileBased(PartDesign::Body *pcActiveBody, Gui::Command* cmd, cons
             std::vector<Gui::SelectionObject> selection = cmd->getSelection().getSelectionEx();
             if (selection.size() > 1) { //treat additional selected objects as sections
                 for (std::vector<Gui::SelectionObject>::size_type ii = 1; ii < selection.size(); ii++) {
-                    if (selection[ii].getObject()->isDerivedFrom(Part::Part2DObject::getClassTypeId())) {
-                        auto objCmdSection = Gui::Command::getObjectCmd(selection[ii].getObject());
-                        FCMD_OBJ_CMD(Feat, "Sections += [" << objCmdSection << "]");
-                    }
+                    // Add subvalues even for sketches in case we just want points
+                    auto objCmdSection = Gui::Command::getObjectCmd(selection[ii].getObject());
+                    std::ostringstream ss;
+                    for (auto &s : selection[ii].getSubNames())
+                        ss << "'" << s << "',";
+                    FCMD_OBJ_CMD(Feat, "Sections += [(" << objCmdSection << ", [" << ss.str() << "])]");
                 }
             }
         }


donovaly:
Many thanks. It works well, I only don't understand the comment you added.

Nevertheless it is now merged, I can update the comment later.

AjinkyaDahale:
> Many thanks. It works well, I only don't understand the comment you added.
> 
> Nevertheless it is now merged, I can update the comment later.

It is work towards supporting points for section. So when the user selects a point, loft only to that point instead of the whole sketch. I imagine it would be more relevant for sweep: say you want to sweep along a path and then end at it's last point.

donovaly:
> It is work towards supporting points for section. So when the user selects a point, loft only to that point instead of the whole sketch.

I assumed this since you picked this bug: https://tracker.freecadweb.org/view.php?id=4607
But it does not work yet, at least I failed to extrude a sketch with a circle to a sketch only containing a point. Of to I have to take use method?

My point with the comment is just that we should not put in comments that can be confusing or refer to npt yet implemented features.

AjinkyaDahale:
> > It is work towards supporting points for section. So when the user selects a point, loft only to that point instead of the whole sketch.
> 
> I assumed this since you picked this bug: https://tracker.freecadweb.org/view.php?id=4607 But it does not work yet, at least I failed to extrude a sketch with a circle to a sketch only containing a point. Of to I have to take use method?
> 
> My point with the comment is just that we should not put in comments that can be confusing or refer to npt yet implemented features.

Looking at this again, perhaps I should have let it be as it is, or leave a comment with a TODO flag. Please remind me to remove the it if I forget when the PR is ready.

donovaly:
> Looking at this again, perhaps I should have let it be as it is, or leave a comment with a TODO flag. Please remind me to remove the it if I forget when the PR is ready.

Leave it now as it is. I am looking forward to your PR since lofting to a point would be a very usefule feature.

p.s. when you cite posts, please only cite the sentences you are referring to. This helps to keep the overview for others who want to follow the discussion.

AjinkyaDahale:
> Leave it now as it is. I am looking forward to your PR since lofting to a point would be a very usefule feature.

See #5170. For now it's just vertices the existing solid, but I plan to add support for datum points and sketch vertices (or maybe just sketches with a single vertex). Let me know if any other things should be included.

adrianinsaval:
> I plan to add support for datum points and sketch vertices (or maybe just sketches with a single vertex).
I vote for just sketches with a single vertex

AjinkyaDahale:
> > I plan to add support for datum points and sketch vertices (or maybe just sketches with a single vertex).
> 
> I vote for just sketches with a single vertex

It's reasonable, but right now, if we make a sketch with a single vertex, that sketch is not visible outside of sketch view, and thus can't be selected. If someone were able to make it visible that would be great. @abdullahtahiriyo, would you have time to help in this regard?

adrianinsaval:
but that was exactly why I brought this feature up the other day, no you _can_ have a sketch with a single point visible
[sketchPointVisible.zip](https://github.com/FreeCAD/FreeCAD/files/7529728/sketchPointVisible.zip)
after creating the point you can toggle construction mode off for it
https://wiki.freecadweb.org/Sketcher_CreatePoint#Options

AjinkyaDahale:
> no you _can_ have a sketch with a single point visible [sketchPointVisible.zip](https://github.com/FreeCAD/FreeCAD/files/7529728/sketchPointVisible.zip) after creating the point you can toggle construction mode off for it https://wiki.freecadweb.org/Sketcher_CreatePoint#Options

If that's the case perhaps I can make it work. I'll check out your example when my internet issues go away.

donovaly:
> I plan to add support for datum point

Thant would contradict the concept of datums. Datums are not part of objects. This is also sensible to me.

As @adrianinsaval said, we have the point object in sketches.

What should also be possible is to select a vertex of a 3D object (e.g. of a pad) as end/start point of a loft. 

And of course also the case of having a vertex as end and startpoint should be handled :-))

AjinkyaDahale:
> > I plan to add support for datum point
> 
> Thant would contradict the concept of datums. Datums are not part of objects. This is also sensible to me.
> 

I see. In that case I can leave datum points alone. 

> As @adrianinsaval said, we have the point object in sketches.
> 

That we have, but I was not able to make them visible in the 3D view.

> What should also be possible is to select a vertex of a 3D object (e.g. of a pad) as end/start point of a loft.
> 

That is already possible for use as the "last section". Their use in the first section, the "profile" should be straightforward but I just haven't gotten to it.

> And of course also the case of having a vertex as end and startpoint should be handled :-))

As I mentioned in the PR itself, vertices can _only_ be used as start and end points.

donovaly:
> > And of course also the case of having a vertex as end and startpoint should be handled :-))
> 
> As I mentioned in the PR itself, vertices can _only_ be used as start and end points.

That was just a joke ;-) - a vertex as start and end, without an intermediate section, is a line. Yes, I am a nerd.

abdullahtahiriyo:
@AjinkyaDahale 

I did not try to understand this thread. I am just answering the question. In the following I am not implying you should support single points in sketches at all. I am just answering the question on whether points can be made "defining points".

As @adrianinsaval rightly points out, since end of last year it is indeed possible to have points visible in the 3D view, outside of edit mode in a Sketch. Unlike the rest of the geometries, that are by default "defining" (normal geometry that define the profile of a sketch) unless you have the geometry insertion mode toggled to insert construction geometry, points are always enter as construction points regardless of the insertion mode. 

This apparently inconsistent behaviour is justified because: 1) 99.9% of the times a user expects a construction point regardless of the insertion mode; and 2) the introduction of the feature is meant to respect the legacy operation of the sketcher, in which there were only construction points.

However, if it is relevant, this behaviour is only possible for single points (geometric points created with the geometry point). It is not possible for "points" that are not geometric points, but endpoints of a geometric object. The reason is that geometric objects are either normal or construction, but not partly normal, partly construction.

To have a defining point: 1) select the point creation tool in the sketcher, 2) create a point, 3) Select the point, 4) click the toggle construction button. The color of the button will turn white (from typical construction point color, red). If you now exit edit mode, the profile will show you a point.

AjinkyaDahale:
Thank you for your explanation, @abdullahtahiriyo.

> As @adrianinsaval rightly points out, since end of last year it is indeed possible to have points visible in the 3D view, outside of edit mode in a Sketch.

It appears this is true for only construction points. "Defining" points are not visible outside of sketch edit mode.

> ...points are always enter as construction points regardless of the insertion mode.

This might be a bug in itself. Currently, without the insertion mode set to construction, newly inserted points are "defining".

> ...It is not possible for "points" that are not geometric points, but endpoints of a geometric object. The reason is that geometric objects are either normal or construction, but not partly normal, partly construction.
> 

I believe I can work with that. Regardless of the point being "normal" or "construction", I only need them to be able to yield a `TopoDS_Shape` to pass on to the relevant OCCT algorithm.

donovaly:
> It appears this is true for only construction points. 

Sure. Like for every other geometry. If you have a circle in construction mode, it will also not be visible when you leave the sketch. Points don't behave differently in this respect.

Here is an example where the last sketch just contains a dot:
[lofttest-PR5155.zip](https://github.com/FreeCAD/FreeCAD/files/7532558/lofttest-PR5155.zip)

The idea is that the user can select this sketch with the single point and this point will then be the end/start point of the loft.

> This might be a bug in itself. Currently, without the insertion mode set to construction, newly inserted points are "defining".

As abdullah wrote, the behavior that points are by default construction elements is there on purpose. It is also described in the Wiki: https://wiki.freecadweb.org/Sketcher_CreatePoint#Options

When your PR #5170 to use single dots as points for lofts and pipes are implemented, we could change this. As I wrote, using single-point sketches should be usable for holes as well since the current hole mechanism is confusing for new users. Then many users need single points as normal geometry.

AjinkyaDahale:
> > It appears this is true for only construction points.
> 
> Sure. Like for every other geometry. If you have a circle in construction mode, it will also not be visible when you leave the sketch. Points don't behave differently in this respect.

Ah finally I put two and two together: construction points are red and "defining" points are white. What must have gotten me confused was the fact that the end points of edges are always red.

> When your PR #5170 to use single dots as points for lofts and pipes are implemented, we could change this. As I wrote, using single-point sketches should be usable for holes as well since the current hole mechanism is confusing for new users. Then many users need single points as normal geometry.

abdullahtahiriyo:
> Ah finally I put two and two together: construction points are red and "defining" points are white. What must have gotten me confused was the fact that the end points of edges are always red.

Because they are not "single points"/"geometric points". "Single points" that are defining "get written" into the TopoShape defining the sketch profile in non-edit mode. The rest only exist (as points) during edit mode. This happens in sketch.cpp in Sketch::toShape().

Let me know if you have any question.

AjinkyaDahale:
> Let me know if you have any question.

Thanks for yet another explanation. One question: how do we tell if the sketch has these geometric points and how many?

abdullahtahiriyo:
If you can access to the Geometry property of SketchObject, this is a std::vector<Part::Geometry *>. You want to test each pointer in the array if the TypeId is of type GeomPoint::getClassId(). Doing this, you can know which indices are GeomPoints and how many GeomPoints there are. Let me know if this is not helpful.

AjinkyaDahale:
> If you can access to the Geometry property of SketchObject, this is a std::vector<Part::Geometry *>. You want to test each pointer in the array if the TypeId is of type GeomPoint::getClassId().

I see. Thanks yet again. I went with getting the Sketch's `TopoDS_Shape` and searching for vertices within it. 
https://github.com/FreeCAD/FreeCAD/blob/5d1e4d9a2926dab311385a64c5a2b9f286ad5e89/src/Mod/PartDesign/App/FeatureLoft.cpp#L152
I did so since it's a bit more general and should also encompass cases where we select points on a body. Please let me know if you have an opinion on the particular implementation, preferably in that PR itself.

abdullahtahiriyo:
> I see. Thanks yet again. I went with getting the Sketch's `TopoDS_Shape` and searching for vertices within it.

Ah! I see. I did not realise you were dealing with the the TopoDS itself. Searching for a Vertex sounds like the most reasonable thing to do.

PR #5170, opened 2021-11-12
[PD] Support for point sections in loft

AjinkyaDahale:
The `BRepOffsetAPI_ThruSections` method that forms the crux of loft supports the first and the last sections to be vertices instead of wires.

Addresses issue#4607.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureLoft.cpp
+++ b/src/Mod/PartDesign/App/FeatureLoft.cpp
@@ -153,7 +153,7 @@ App::DocumentObjectExecReturn *Loft::execute(void)
                 profilePoint = ex.Current();
             }
             if (i > 1)
-                return new App::DocumentObjectExecReturn("Loft: Only one isolated point is needed if using a sketch with isolated points for section");
+                return new App::DocumentObjectExecReturn("Loft: When using points for profile/sections, the sketch should have a single point");
         }
 
         bool isLastSectionVertex = false;
@@ -179,7 +179,7 @@ App::DocumentObjectExecReturn *Loft::execute(void)
                         wires.push_back(ex.Current());
                 }
                 if (j > 1)
-                    return new App::DocumentObjectExecReturn("Loft: Only one isolated point is needed if using a sketch with isolated points for section");
+                    return new App::DocumentObjectExecReturn("Loft: When using points for profile/sections, the sketch should have a single point");
             }
             if (!isLastSectionVertex && numWiresAdded < wiresections.size())
                 return new App::DocumentObjectExecReturn("Loft: Sections need to have the same amount of inner wires as the base section");
--- a/src/Mod/PartDesign/App/FeatureLoft.cpp
+++ b/src/Mod/PartDesign/App/FeatureLoft.cpp
@@ -74,17 +74,48 @@ short Loft::mustExecute() const
 
 App::DocumentObjectExecReturn *Loft::execute(void)
 {
+    auto getSectionShape =
+        [](App::DocumentObject* feature, const std::vector<std::string> &subs) -> TopoDS_Shape {
+            if (!feature ||
+                !feature->isDerivedFrom(Part::Feature::getClassTypeId()))
+                throw Base::TypeError("Loft: Invalid profile/section");
+
+            auto subName = subs.empty() ? "" : subs.front();
+
+            // only take the entire shape when we have a sketch selected, but
+            // not a point of the sketch
+            if (feature->isDerivedFrom(Part::Part2DObject::getClassTypeId()) &&
+                !(subName.size() > 6 && subName.substr(0,6) == "Vertex"))
+                return static_cast<Part::Part2DObject*>(feature)->Shape.getValue();
+            else {
+                if(subName.empty())
+                    throw Base::ValueError("No valid subelement linked in Part::Feature");
+                return static_cast<Part::Feature*>(feature)->Shape.getShape().getSubShape(subName.c_str());
+            }
+        };
+
+    auto addWiresToWireSections =
+        [](TopoDS_Shape& section,
+           std::vector<std::vector<TopoDS_Shape>>& wiresections) -> size_t {
+            TopExp_Explorer ex;
+            size_t i=0;
+            bool initialWireSectionsEmpty = wiresections.empty();
+            for (ex.Init(section, TopAbs_WIRE); ex.More(); ex.Next(), ++i) {
+                // if profile was just a point then this is where we can first set our list
+                if (i>=wiresections.size()) {
+                    if (initialWireSectionsEmpty)
+                        wiresections.emplace_back(1, ex.Current());
+                    else
+                        throw Base::ValueError("Loft: Sections need to have the same amount of inner wires (except profile and last section, which can be points)");
+                }
+                else
+                    wiresections[i].push_back(TopoDS::Wire(ex.Current()));
+            }
+            return i;
+        };
 
     std::vector<TopoDS_Wire> wires;
-    try {
-        wires = getProfileWires();
-    } catch (const Base::Exception& e) {
-        return new App::DocumentObjectExecReturn(e.what());
-    }
-
-    TopoDS_Shape sketchshape = getVerifiedFace();
-    if (sketchshape.IsNull())
-        return new App::DocumentObjectExecReturn("Loft: Creating a face from sketch failed");
+    TopoDS_Shape profilePoint;
 
     // if the Base property has a valid shape, fuse the pipe into it
     TopoDS_Shape base;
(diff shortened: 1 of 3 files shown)


donovaly:
What is the status of this PR?

Concerning the feature, it should support to use vertices of 3D objects as well as sketches just consisting of a point object but not datum points.
The sketch point support might be tricky since I remember a discussion once I proposed to change the Hole feature so that one only has to provide a point. At the moment one has to produce a circle, but only the center pint of it is used as position of the hole).
So if it is possible to have only a sketcher point, I would try if I can change the Hole feature according to your solution for lofts.
(Maybe this is not that tricky, I simply never looked into this yet.)

However, I could not resist to merge already the 2 typos you corrected in the PR.

AjinkyaDahale:
> What is the status of this PR?
> 

I answered some questions in https://github.com/FreeCAD/FreeCAD/pull/5168#issuecomment-967729903. In summary, we can use vertices of the solid object as the last section for a loft, and I'm working on trying to support 

> ...The sketch point support might be tricky since I remember a discussion once I proposed to change the Hole feature so that one only has to provide a point. At the moment one has to produce a circle, but only the center pint of it is used as position of the hole). So if it is possible to have only a sketcher point, I would try if I can change the Hole feature according to your solution for lofts. (Maybe this is not that tricky, I simply never looked into this yet.)

If anything this support forces us to address supporting vertex-only sketches, or only vertices from sketches, because for loft (and pipe) vertices and circles mean different things.

> However, I could not resist to merge already the 2 typos you corrected in the PR.

That's perfectly fine :)

AjinkyaDahale:
Started a forum thread to keep the PR thread clean: https://forum.freecadweb.org/viewtopic.php?f=19&t=63697. @donovaly @adrianinsaval perhaps we should move the discussion there.

AjinkyaDahale:
@donovaly @adrianinsaval there are a few things remaining to be done, but overall I would call this ready for review.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5170">pipeline status</a> for feature branch [PR_5170](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5170). Pipeline [408635926](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/408635926) was triggered at [08e6c44](https://github.com/FreeCAD/FreeCAD/pull/5170/commits/08e6c44463e41d8386e78ac7b90503292fc08264). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5170">pipeline status</a> for feature branch [PR_5170](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5170). Pipeline [409255305](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/409255305) was triggered at [557193c](https://github.com/FreeCAD/FreeCAD/pull/5170/commits/557193c4113c22aaacd6a7c6f13c218a41cbb151). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5170">pipeline status</a> for feature branch [PR_5170](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5170). Pipeline [411472361](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/411472361) was triggered at [f9a50d8](https://github.com/FreeCAD/FreeCAD/pull/5170/commits/f9a50d8068010203478760c657d7df1a271ea031). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
Adding back the "do not merge" tag, because after rebasing with PR #5176, the "front" face is not built.

AjinkyaDahale:
> Adding back the "do not merge" tag, because after rebasing with PR #5176 merged, the "front" face is not built.

This bug should now be fixed.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5170">pipeline status</a> for feature branch [PR_5170](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5170). Pipeline [413453485](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/413453485) was triggered at [f8da88c](https://github.com/FreeCAD/FreeCAD/pull/5170/commits/f8da88c1b45cf884552e1797fb062ebaeaefa272). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureLoft.cpp
```diff
@@ -102,80 +133,109 @@ App::DocumentObjectExecReturn *Loft::execute(void)
...
-                    return new App::DocumentObjectExecReturn("Loft: Sections need to have the same amount of inner wires as the base section");
-                wiresections[i].push_back(TopoDS::Wire(ex.Current()));
+            size_t i = 0;
+            for (ex.Init(profileShape, TopAbs_VERTEX); ex.More(); ex.Next(), ++i) {
+                profilePoint = ex.Current();
             }
-            if (i<wiresections.size())
-                    return new App::DocumentObjectExecReturn("Loft: Sections need to have the same amount of inner wires as the base section");
+            if (i > 1)
+                return new App::DocumentObjectExecReturn("Loft: Only one isolated point is needed if using a sketch with isolated points for section");
```
donovaly, line 156: Misleading, better
("Only single-point sketches are allowed for loft sections"

src/Mod/PartDesign/App/FeatureLoft.cpp
```diff
@@ -102,80 +133,109 @@ App::DocumentObjectExecReturn *Loft::execute(void)
...
+        }
 
+        bool isLastSectionVertex = false;
+
+        for (auto &subSet : multisections) {
+            if (!subSet.first->isDerivedFrom(Part::Feature::getClassTypeId()))
+                return new App::DocumentObjectExecReturn("Loft: All sections need to be part features");
```
donovaly, line 163: I would omit the prefix "Loft" since these messages only appear while editing/creating lofts.

src/Mod/PartDesign/Gui/Command.cpp
```diff
@@ -1004,18 +1004,32 @@ void prepareProfileBased(PartDesign::Body *pcActiveBody, Gui::Command* cmd, cons
...
-            FCMD_OBJ_CMD(Feat,"Profile = (" << objCmd << ", [" << ss.str() << "])");
-        }
 
-        //for additive and subtractive lofts allow the user to preselect the sections
+        auto runProfileCmd =
+            [=]() {
+                FCMD_OBJ_CMD(Feat,"Profile = " << objCmd);
+            };
+
+        auto runProfileCmdWithSubs =
```
donovaly, line 1013: It is not your code but while you are at it, please add a comment since this code block is hard to get.

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
```diff
@@ -105,13 +105,15 @@ TaskLoftParameters::TaskLoftParameters(ViewProviderLoft *LoftView, bool /*newObj
     if (profile) {
         Gui::Application::Instance->showViewProvider(profile);
 
+        // TODO: if it is a single vertex of a sketch, use that subshape's name
```
donovaly, line 108: I don't understand this comment. When I use a single-vertex sketch, the label is correctly the sketch. Like it is the case for all other cases in which sketches form a section. Therefore I don't see the need to use the vertex name, especially since you already assured that only single-vertex sketches can be used.

donovaly (approved):
Excellent work.
I stress-tested the feature and it works nicely. I have only minor comments/clarifications.

donovaly:
Many thanks for this PR. This improves the loft feature a lot since pyramids are not that rare (own experience).

Please have a look at my very minor comments. I will merge it then and then you can continue to bring Sweep/Pipe to the same functionality ;-)

(By the way, after sweep/pipe can handle vertices as well, I propose to make a PR to uniform the naming. At the moment the tern "sweep" is mixed with "pipe" in the code. We should only use one term and "pipe" is for non-native speakers like me even misleading.)

src/Mod/PartDesign/App/FeatureLoft.cpp
adrianinsaval: Are they not shown if you recompute and it fails? It may not be immediately clear from were the error comes from in that case, further more, to me it's better to be always clear what is sending output to the report view, you can then look back on previous messages and know were they came from even if you are working on something else now. In truth I would like it if the output mentioned wich specific feature in three prompted it but that's another discussion.

src/Mod/PartDesign/App/FeatureLoft.cpp
AjinkyaDahale: I'd say that's still misleading. Because then people can believe regular sketches (that make faces) are not supported. That said, a better string would be well appreciated.

src/Mod/PartDesign/App/FeatureLoft.cpp
donovaly: OK, then keep it.

src/Mod/PartDesign/App/FeatureLoft.cpp
donovaly: I don't agree since the message will only occur when a sketch has more than one vertex. Then the messages makes clear that single-point sketches are allowed but multi-point sketches are not.

donovaly:
I now found a test case that fails:

FreeCAD_MdqpV6Szrx

Here is the test file: 
[loftbug.zip](https://github.com/FreeCAD/FreeCAD/files/7577649/loftbug.zip)


So the loft is nicely created but then not visible. Maybe this is because a body cannot contain of several volume shapes and in my test case they form one but the most thin point is a vertex?
However, this would not be a stopper for the PR.

src/Mod/PartDesign/App/FeatureLoft.cpp
AjinkyaDahale: I'd agree with @adrianinsaval here. There doesn't appear to be much negative to keeping the "Loft: " in.

src/Mod/PartDesign/Gui/Command.cpp
AjinkyaDahale: Are you talking about like 1010 or 1013?

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
AjinkyaDahale: I think you misunderstood. If the user selects a single vertex from a sketch (even if it has other elements), the code uses that instead of the whole sketch. It's not much relevant here (so I can remove it if I have to), but very useful for pipes, where we can choose one end point of a path as the first section and make a "horn" like shape.

src/Mod/PartDesign/Gui/Command.cpp
donovaly: the code block that begins with 1013

AjinkyaDahale:
> I now found a test case that fails:
> ...
> So the loft is nicely created but then not visible. Maybe this is because a body cannot contain of several volume shapes and in my test case they form one but the most thin point is a vertex? However, this would not be a stopper for the PR.

That is probably exactly what's happening. Could you try deleting the box and using a sketch instead?

src/Mod/PartDesign/Gui/TaskLoftParameters.cpp
donovaly: > If the user selects a single vertex from a sketch (even if it has other elements), 

Aha, cool feature, I was not yet aware of. Then it would indeed be useful to see the subname of the taken vertex and not the whole sketch as label. However, this is cosmetic. So leave the TODO in for now.

donovaly:
> That is probably exactly what's happening. Could you try deleting the box and using a sketch instead?

This works, here an example where the begin and end is a single-point sketch:
FreeCAD_Z23ccfTCcM

However, what I wanted to test is if a vertex from a surface could be taken. Therefore i need an object like the box.

AjinkyaDahale:
> However, what I wanted to test is if a vertex from a surface could be taken. Therefore i need an object like the box.

try one of the vertices on the left face. There should be some 2D (at least) intersection with the solid, or so I would imagine.

src/Mod/PartDesign/App/FeatureLoft.cpp
AjinkyaDahale: Changed the string to try making it a little clearer. Let me know if it is still ambiguous.

src/Mod/PartDesign/Gui/Command.cpp
AjinkyaDahale: Comment added.

donovaly: thanks

donovaly:
Merging now, since the comments and wording changes are in.

The issue I mentioned in https://github.com/FreeCAD/FreeCAD/pull/5170#issuecomment-974995086
is due to the fact that PD cannot have two 3D objects within one body yet. This is something @realthunder has in his branch, so in future this would be possible. For now, it is sufficient to mention the special case in the Wiki.

PR #5172, opened 2021-11-13
[PD] uniform pad and pocket handling

donovaly:
when using the option "through all" pads are 10.0 long while pocket is "only" 5.0. This PR uniforms this to 10.0


adrianinsaval:
AFAIK this doesn't affect through all mode, it's just the default values for the length, it's still a good idea to use the same value in both though

donovaly:
> 
> 
> AFAIK this doesn't affect through all mode, it's just the default values for the length, it's still a good idea to use the same value in both though

Oh, indeed: https://wiki.freecadweb.org/PartDesign_Pocket#Through_all

However, it does not harm to use the same initial value.

I plan to sort out the code pad and pocket shares, but at first I have to fix some issues I found. And while doing this, I found this inconsistent initial length.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5172">pipeline status</a> for feature branch [PR_5172](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5172). Pipeline [408636512](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/408636512) was triggered at [93ca8ff](https://github.com/FreeCAD/FreeCAD/pull/5172/commits/93ca8ff22010894a8c1280a8e9c1871df28b30ec). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
Making things uniform is nice and good but if the code of the two functions is identical for 99% then the next logical step is to avoid the code duplication and move the part starting from
`PartDesign::Body *pcActiveBody = PartDesignGui::getBody(true);`
to a global function. This function can be called
`void prepareProfileBased(Gui::Command* cmd, const std::string& which, double length)`

The length argument should go the captures section of the lambda expression and that can be changed to:
```
...
FCMD_OBJ_CMD(Feat,"Length = " << length);
...
```

Then the Pad command calls this function with
`prepareProfileBased(this, "Pad", 10.0);`
and the Pocket command with
`prepareProfileBased(this, "Pocket", 5.0);`

donovaly:
> then the next logical step is to avoid the code duplication

Yes, as I wrote today, this is my plan. But before starting this action, I want to merge the pending bugfix PR and to have the same code base, so this PR. Then the plan is to request in the forum further tests.

I just want to avoid double work. So once I know pocket is working as fine as pad, I can concentrate on the code shifting.

PR #2758, opened 2019-12-05
Accumulated minor fixes and improvements

realthunder:
These patches are split out from #2723

PR #5188, opened 2021-11-21
Do not auto rename when adding dynamic property

realthunder:
Do not auto rename when adding dynamic property with an invalid or existing name, but throw exception instead, because there is no easy way for Python code to find out the auto selected name. Parameter `BaseApp/Preferences/Document/AutoNameDynamicProperty` can be used to revert back to old behavior.

@chennes This is splitted out from #2758. I'll see if I can split out others.

### DIFF
--- a/src/App/DynamicProperty.cpp
+++ b/src/App/DynamicProperty.cpp
@@ -37,7 +37,7 @@
 #include <Base/Exception.h>
 #include <Base/Tools.h>
 
-FC_LOG_LEVEL_INIT("DynamicProperty",true,true)
+FC_LOG_LEVEL_INIT("Property",true,true)
 
 
 using namespace App;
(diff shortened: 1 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5188">pipeline status</a> for feature branch [PR_5188](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5188). Pipeline [413453490](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/413453490) was triggered at [55db4b7](https://github.com/FreeCAD/FreeCAD/pull/5188/commits/55db4b7fb173d250954922941e75bbd10bf0ed8c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Thanks, that's very helpful, I appreciate it.

wwmayer:
It's considered bad practice (code smell) to use user parameters to control the behaviour of low-level classes. If there is no real use case (any more) to auto-rename the passed property name then completely remove the code to do so.

If it should still be allowed to auto rename properties then add an argument to the function to make the intention clear. Currently, if a programmer wants to do it, first of all he must even know this parameter exists, then he must change the value and as last step restore the old value.

But the general problem is that the behaviour is almost unpredictable because depending on the value of the parameter custom Python features may work on one system and fail on another system.

A further special problem is the use of the static keyword:
```
    static ParameterGrp::handle hGrp = GetApplication().GetParameterGroupByPath(
            "User parameter:BaseApp/Preferences/Document");
```
The object hGrp will be used for the whole lifetime of the application and will never be read again. However, if in the parameter editor the group is renamed by the user, a new group "Document" is created and the user changes "AutoNameDynamicProperty" there, then DynamicProperty::addDynamicProperty() keeps reading values from the wrong parameter group.

realthunder:
> It's considered bad practice (code smell) to use user parameters to control the behaviour of low-level classes. If there is no real use case (any more) to auto-rename the passed property name then completely remove the code to do so.

Well, although FreeCAD has 'system.cfg', but no code is currently using it, and the parameter editor didn't expose it either. The way I see it, parameter editor in FreeCAD is like registry in Windows and sys filesystem in Linux, which exposes low level settings for power users (and other modules through API) to control low-level system behavior.

I discovered the auto renaming problem because I made a mistake in my addon module by adding property without checking the existence first. And I only discovered the problem when I saw many auto renamed properties in one of the user files while dealing with an unrelated bug. This makes me think that the silent renaming is a bad behavior. To bring this to other addon author's attention, I choose to throw exception. But I also realize that this may have some implication on existing addon modules. The parameter is added here for backward compatibility reason.

> A further special problem is the use of the static keyword:

This problem applies to the parameter observers as well, which is why I think there is [code](https://github.com/FreeCAD/FreeCAD/blob/81f4bcce66e2d2fda09a13b95699ab3186f84b0f/src/Base/Parameter.cpp#L887) there to not remove the group if there is any reference to it. I haven't tested it, though. I know this because I actually recently made a few changes to the parameter API in my branch related to this problem.

realthunder:
Just realized that you're talking about renaming the parameter group, not removing. Yeah, it does cause problem like you mentioned. I guess adding property is not really performance critical, so changing to non static is not that a big deal.

Still though, the renaming problem applies to those parameter observers as well. Maybe we should implement rename as copy, remove, and then add.

wwmayer:
> Well, although FreeCAD has 'system.cfg', but no code is currently using it,

The system.cfg is a different story and it was added by the project founder. I don't know what the original intention was but I assume it was supposed to put it into a directory where users only have read access (such as All Users on Windows or /etc/ on Linux) and that only contains application-specific settings. The way it is now it's pretty useless.

> The way I see it, parameter editor in FreeCAD is like registry in Windows and sys filesystem in Linux, which exposes low level settings for power users (and other modules through API) to control low-level system behavior.

With the consequence of all sorts of side-effects. I don't how many hours I spent to locate problems with functions that are affected by environment variables and to make things consistently working on all platforms, e.g. all the locale specific settings.

So, IMO for all low-level stuff we should try to avoid possible side-effects to guarantee consistent behaviour.

> I discovered the auto renaming problem because I made a mistake in my addon module by adding property without checking the existence first. And I only discovered the problem when I saw many auto renamed properties in one of the user files while dealing with an unrelated bug. This makes me think that the silent renaming is a bad behavior.

I fully agree on this and because you said it's impossible inside Python to determine the actually used property name the affected custom Python features are pretty useless. That's why I think the whole renaming stuff can be entirely removed sooner or later.

Existing project files with unexpected property names should still load fine because the names are stored inside Document.xml and the __init__ function of a Python feature with the addProperty() calls won't be called anyway.

> This problem applies to the parameter observers as well, which is why I think there is code there to not remove the group if there is any reference to it. I haven't tested it, though. I know this because I actually recently made a few changes to the parameter API in my branch related to this problem.

> Just realized that you're talking about renaming the parameter group, not removing.

Both can cause problems. I have seen use cases where two config files are merged into one by loading (a part of) a second file into a group of the already loaded file. If the group is not empty and there are attached observers then very funny things can happen directly inside the DOM document of the xercesc library. In some cases some removed nodes still somehow have a link to the document and when saving the document it can corrupt the XML structure.

> Still though, the renaming problem applies to those parameter observers as well. Maybe we should implement rename as copy, remove, and then add.

Two years ago I fixed several of these issues but before I could make more changes my PC died. Luckily, I still have a back-up of these changes and I should have a look of what can be merged.

Most of the currently used observers are only attached temporarily and thus the chance for problems is rather low but if a group is assigned to a static object the chance for problems is much higher.

> Yeah, it does cause problem like you mentioned. I guess adding property is not really performance critical, so changing to non static is not that a big deal.

If we should keep the possibility to restore the old behaviour then we have to consider two cases:
* should it be possible to decide for each container to allow the renaming then a non-static class member would be needed
* does it suffice to either enable it for all or for none then a static boolean should be used and set only once

PR #5189, opened 2021-11-21
[GUI] [PD] Create a custom button group and use in pipe task

AjinkyaDahale:
Some discussion here: https://github.com/FreeCAD/FreeCAD/pull/5159#issuecomment-966384207. `Gui` looks like an appropriate place to keep the custom group itself, so that it can be used in other workbenches.

### DIFF
--- a/src/Gui/ClearableButtonGroup.cpp
+++ b/src/Gui/ClearableButtonGroup.cpp
@@ -1,0 +1,53 @@
+ /**************************************************************************
+ *   Copyright (c) 2021 FreeCAD Developers                                 *
+ *   Author: Ajinkya Dahale <dahale.a.p@gmail.com>                         *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This library is free software; you can redistribute it and/or         *
+ *   modify it under the terms of the GNU Library General Public           *
+ *   License as published by the Free Software Foundation; either          *
+ *   version 2 of the License, or (at your option) any later version.      *
+ *                                                                         *
+ *   This library  is distributed in the hope that it will be useful,      *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+
+
+#include "PreCompiled.h"
+#ifndef _PreComp_
+# include <QButtonGroup>
+# include <QAbstractButton>
+#endif
+
+#include "ClearableButtonGroup.h"
+
+using namespace Gui;
+
+ClearableButtonGroup::ClearableButtonGroup(QObject *parent)
+    : QButtonGroup(parent)
+{
+    _bExclusive = true;
+    QButtonGroup::setExclusive(false);
+    connect(this, SIGNAL(buttonClicked(QAbstractButton *)), SLOT(buttonClicked(QAbstractButton *)));
+}
+
+void ClearableButtonGroup::buttonClicked(QAbstractButton *button)
+{
+    if (_bExclusive) {
+        // just uncheck all other buttons regardless of the state of the clicked button
+        QList<QAbstractButton *> buttonlist = buttons();
+        for (auto &iBtn : buttonlist)
+            if (iBtn && iBtn != button && iBtn->isCheckable())
+                iBtn->setChecked(false);
+    }
+}
--- a/src/Gui/ClearableButtonGroup.h
+++ b/src/Gui/ClearableButtonGroup.h
@@ -1,0 +1,58 @@
+ /**************************************************************************
+ *   Copyright (c) 2021 FreeCAD Developers                                 *
+ *   Author: Ajinkya Dahale <dahale.a.p@gmail.com>                         *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This library is free software; you can redistribute it and/or         *
+ *   modify it under the terms of the GNU Library General Public           *
+ *   License as published by the Free Software Foundation; either          *
+ *   version 2 of the License, or (at your option) any later version.      *
+ *                                                                         *
+ *   This library  is distributed in the hope that it will be useful,      *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+
+
+#ifndef GUI_CLEARABLEBUTTONGROUP_H
+#define GUI_CLEARABLEBUTTONGROUP_H
+
+#include <QButtonGroup>
+#include <QAbstractButton>
+
+namespace Gui {
+
+/**
+ * This class extends QButtonGroup to allow toggling off all buttons when set to
+ * `exclusive`.
+ * Uses code from https://stackoverflow.com/a/59910329.
+ */
+class ClearableButtonGroup : public QButtonGroup
+{
+    Q_OBJECT
+
+public:
+    ClearableButtonGroup(QObject *parent = nullptr);
+
+    void setExclusive(bool bExclusive) { _bExclusive = bExclusive; }
+    bool exclusive() const { return _bExclusive; }
+
+protected Q_SLOTS:
+    void buttonClicked(QAbstractButton *button);
+
+protected:
+    bool _bExclusive;
+};
+
+}
+
+#endif // GUI_CLEARABLEBUTTONGROUP_H
--- a/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
@@ -43,15 +43,16 @@
 #include <App/Document.h>
 #include <App/Origin.h>
 #include <App/Part.h>
+#include <Base/Console.h>
 #include <Gui/Application.h>
 #include <Gui/Document.h>
 #include <Gui/BitmapFactory.h>
 #include <Gui/ViewProvider.h>
 #include <Gui/WaitCursor.h>
-#include <Base/Console.h>
 #include <Gui/Selection.h>
 #include <Gui/CommandT.h>
 #include <Gui/MainWindow.h>
+#include <Gui/ClearableButtonGroup.h>
 #include <Mod/PartDesign/App/FeaturePipe.h>
 #include <Mod/Sketcher/App/SketchObject.h>
 #include <Mod/PartDesign/App/Body.h>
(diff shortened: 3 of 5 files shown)


AjinkyaDahale:
Some work needs to be done on the PD front to merge the functions called through `toggled` slot. Most of the buttons work the same, but highlight different things. Exception is `TaskPipeParameters::onProfileButton(bool checked)` which does something extra. Possibly just connecting that button separately will do the job. Problem however is maintaining states across the three task "boxes".

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189">pipeline status</a> for feature branch [PR_5189](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189). Pipeline [413453961](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/413453961) was triggered at [fb84047](https://github.com/FreeCAD/FreeCAD/pull/5189/commits/fb84047ea99935f094a28bfc8e33b857950feaa0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189">pipeline status</a> for feature branch [PR_5189](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189). Pipeline [414384462](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/414384462) was triggered at [4b1fee5](https://github.com/FreeCAD/FreeCAD/pull/5189/commits/4b1fee5cc3b6f8105426e18f74f06368dd0b2499). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189">pipeline status</a> for feature branch [PR_5189](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189). Pipeline [417926688](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/417926688) was triggered at [f9cf768](https://github.com/FreeCAD/FreeCAD/pull/5189/commits/f9cf76874a1eab5633508dba77c7edfd9d98499b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/ClearableButtonGroup.cpp
```diff
@@ -0,0 +1,53 @@
+ /**************************************************************************
+ *   Copyright (c) 2021 FreeCAD Developers                                 *
+ *   Author: Ajinkya Dahale                                                *
```
donovaly, line 3: please add a mail address - all other devs have this since copyright holders must be reachable somehow

src/Gui/ClearableButtonGroup.h
```diff
@@ -0,0 +1,58 @@
+ /**************************************************************************
+ *   Copyright (c) 2021 FreeCAD Developers                                 *
+ *   Author: Ajinkya Dahale                                                *
```
donovaly, line 3: same

src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
```diff
@@ -87,12 +87,12 @@ TaskPipeParameters::TaskPipeParameters(ViewProviderPipe *PipeView, bool /*newObj
             this, SLOT(onProfileButton(bool)));
     connect(ui->comboBoxTransition, SIGNAL(currentIndexChanged(int)),
             this, SLOT(onTransitionChanged(int)));
-    connect(ui->buttonRefAdd, SIGNAL(toggled(bool)),
-            this, SLOT(onButtonRefAdd(bool)));
-    connect(ui->buttonRefRemove, SIGNAL(toggled(bool)),
-            this, SLOT(onButtonRefRemove(bool)));
-    connect(ui->buttonSpineBase, SIGNAL(toggled(bool)),
-            this, SLOT(onBaseButton(bool)));
+    // connect(ui->buttonRefAdd, SIGNAL(toggled(bool)),
```
donovaly, line 90: why do you keep them? They are not needed anymore and a comment why there are no connections should be sufficient.

src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
```diff
@@ -610,12 +558,12 @@ TaskPipeOrientation::TaskPipeOrientation(ViewProviderPipe* PipeView, bool /*newO
 
     connect(ui->comboBoxMode, SIGNAL(currentIndexChanged(int)),
             this, SLOT(onOrientationChanged(int)));
-    connect(ui->buttonRefAdd, SIGNAL(toggled(bool)),
-            this, SLOT(onButtonRefAdd(bool)));
-    connect(ui->buttonRefRemove, SIGNAL(toggled(bool)),
-            this, SLOT(onButtonRefRemove(bool)));
-    connect(ui->buttonProfileBase, SIGNAL(toggled(bool)),
-            this, SLOT(onBaseButton(bool)));
+    // connect(ui->buttonRefAdd, SIGNAL(toggled(bool)),
```
donovaly, line 561: dito

donovaly (commented):
looks good I have only minor comments

can you please rebase? This would save me a lot of compile time on testing.

AjinkyaDahale:
@donovaly rebased and should be resolved.

donovaly:
I cannot compile and get this error:
`4>  Creating library D:/FreeCAD-build/src/Mod/PartDesign/Gui/Release/PartDesignGui.lib and object D:/FreeCAD-build/src/Mod/PartDesign/Gui/Release/PartDesignGui.exp
4>TaskPipeParameters.obj : error LNK2019: unresolved external symbol "public: __cdecl Gui::ClearableButtonGroup::ClearableButtonGroup(class QObject *)" (??0ClearableButtonGroup@Gui@@QEAA@PEAVQObject@@@Z) referenced in function "public: __cdecl PartDesignGui::TaskDlgPipeParameters::TaskDlgPipeParameters(class PartDesignGui::ViewProviderPipe *,bool)" (??0TaskDlgPipeParameters@PartDesignGui@@QEAA@PEAVViewProviderPipe@1@_N@Z)
4>D:\FreeCAD-build\Mod\PartDesign\PartDesignGui.pyd : fatal error LNK1120: 1 unresolved externals`

AjinkyaDahale:
> I cannot compile and get this error:

I'm not sure what's happening here, because it's compiling fine for me. Perhaps I used something compiler/OS specific?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189">pipeline status</a> for feature branch [PR_5189](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189). Pipeline [443189931](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443189931) was triggered at [cb8683e](https://github.com/FreeCAD/FreeCAD/pull/5189/commits/cb8683e7fc1e1677d33ce789eb201c68e905e727). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
@donovaly my best guess short of compiling it myself using your build is that use of `QT_NULLPTR` is causing issues. Replaced it with `nullptr` which is used everywhere else. A similar problem is seen [here](https://stackoverflow.com/questions/14448963/error-lnk2019-unresolved-external-symbol-public-thiscall-useruservoid) but the solution is not confirmed.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189">pipeline status</a> for feature branch [PR_5189](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5189). Pipeline [443255615](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443255615) was triggered at [4f7721f](https://github.com/FreeCAD/FreeCAD/pull/5189/commits/4f7721f9b73ffb37c653ce1284fa659a1a439d65). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I still cannot compile. i had a look and it is not clear to me why you implement a new QButtonGroup.

Therefore I took your changes but omitted the customClearableButtonGroup, used instead directly QButtonGroup and this works for me. Here is the PR: #5354 
Please give it a try.

By the way, Widgets.cpp has already a custom button group that could perhaps be taken if necessary.

AjinkyaDahale:
> I still cannot compile. i had a look and it is not clear to me why you implement a new QButtonGroup.

Please see the relevant comment from #5159 I mentioned in the PR: https://github.com/FreeCAD/FreeCAD/pull/5159#issuecomment-966384207. Problem with the original QButtonGroup is that if set to exclusive, then one button always remains pressed (initially all are unpressed when the task starts, but try say pressing any "Add" twice), which we do not want.

> Therefore I took your changes but omitted the customClearableButtonGroup, used instead directly QButtonGroup and this works for me. Here is the PR: #5354 Please give it a try.

I am skeptical it won't have the issue I described earlier, but I'm willing to give it a try if you still want me to.

> By the way, Widgets.cpp has already a custom button group that could perhaps be taken if necessary.

Let me take a look.

donovaly:
> > Therefore I took your changes but omitted the customClearableButtonGroup, used instead directly QButtonGroup and this works for me. Here is the PR: #5354 Please give it a try.
> 
> I am skeptical it won't have the issue I described earlier, but I'm willing to give it a try if you still want me to.

Yes, please try it out. Maybe I missed something but for me the buttons behave as they should.

AjinkyaDahale:
> > By the way, Widgets.cpp has already a custom button group that could perhaps be taken if necessary.
> 
> Let me take a look.
Update: I see that @wwmayer implemented the custom group exactly for this purpose one day after I started this PR (792277a848c5bb5d5cfbdf592e8c9991b7fa4889). I expect I can make changes accordingly.

donovaly:
> Update: I see that @wwmayer implemented the custom group exactly for this purpose one day after I started this PR ([792277a](https://github.com/FreeCAD/FreeCAD/commit/792277a848c5bb5d5cfbdf592e8c9991b7fa4889)). I expect I can make changes accordingly.

OK. Nevertheless I would like if you could try out my PR. I spent some time with this and want to learn what is not working with my (the direct Qt) approach.

AjinkyaDahale:
Replaced by #5356. Closing.

AjinkyaDahale:
> OK. Nevertheless I would like if you could try out my PR. I spent some time with this and want to learn what is not working with my (the direct Qt) approach.

It's taking some time to compile on my PC, probably because the change is so deep in the dependency. But try the following: press a button in the task view for pipe (doesn't matter which), and then try pressing that button again without doing anything else.

donovaly:
> But try the following: press a button in the task view for pipe (doesn't matter which), and then try pressing that button again without doing anything else.

OK, now I got it. Yes, this does not work with QButtonGroup directly.

wwmayer:
The class declaration lacks of the export macro. That's why it causes a linking error on Windows. So change it to:
`class GuiExport ClearableButtonGroup : public QButtonGroup`

> Update: I see that @wwmayer implemented the custom group exactly for this purpose one day after I started this PR

@AjinkyaDahale 
I needed this button group for something else too and didn't realize your PR at that time. If you like we can also use your ClearableButtonGroup but then I suggest to use the lambda expression as I did it. This is because the overloaded connect() function that uses the SIGNAL() and SLOT() macro shouldn't be used any more in new code. The reason is that with the modern connect() function you already get an error at compile time in case a (slot) function is missing.

donovaly:
> If you like we can also use your ClearableButtonGroup 

Ajinkya already made a PR: #5356 
(the button group works there but there is another issue)

AjinkyaDahale:
> The class declaration lacks of the export macro. That's why it causes a linking error on Windows. So change it to: `class GuiExport ClearableButtonGroup : public QButtonGroup`

Thanks for catching it!

> If you like we can also use your ClearableButtonGroup...

No, I think we're fine with using your implementation within `Widgets.h`. It makes sense to keep these classes together where they already are, and with the lambda expression `ClearableButtonGroup` will look the same anyway. If there's any change necessary I'll make them on your implementation.

AjinkyaDahale:
> > If you like we can also use your ClearableButtonGroup
> 
> Ajinkya already made a PR: #5356 (the button group works there but there is another issue)

We could try adding the export macro and see how it behaves. If the problem is introduced while I switched to wmayer's implementation, that reduces our search space.

PR #5191, opened 2021-11-22
Show green color for linked property

realthunder:
To distinguish the property that owned by the Link from those belong to the linked object.

Screenshot from 2021-11-22 10-25-34

### DIFF
--- a/src/Gui/propertyeditor/PropertyItem.cpp
+++ b/src/Gui/propertyeditor/PropertyItem.cpp
@@ -553,6 +553,9 @@ QVariant PropertyItem::data(int column, int role) const
 {
     // property name
     if (column == 0) {
+        if (role == Qt::ForegroundRole && linked)
+            return QVariant::fromValue(QColor(0,0x80,0));
+
         if (role == Qt::BackgroundRole || role == Qt::ForegroundRole) {
             if(PropertyView::showAll()
                 && propertyItems.size() == 1


chennes:
I think we should move away from hard-coded colors -- In this case we can't really use a StatefulLabel without a more significant refactoring (probably need a delegate with an overridden `paint()` for that), so maybe for now just add a user preference? I'm open to other ideas though.

realthunder:
Yeah, I thought about that. But then there are quite a few other places with hard coded color here and there. So I guess it would be better to fix them together in a dedicated PR. I think it would be easier to do with user preferences.

chennes:
Yes, fixing all those is one of the items on my to-do list -- so I was hoping to not add more work to it! 😄

realthunder:
So, have you decided which way to go? Using preference? I can change it here.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5191">pipeline status</a> for feature branch [PR_5191](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5191). Pipeline [414384466](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/414384466) was triggered at [6b3487a](https://github.com/FreeCAD/FreeCAD/pull/5191/commits/6b3487af8c553be67932506c0df4c64c655acbac). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

yorikvanhavre:
My 2 cents, I would try to make them qss-stylable. There are many, many colors that would need to be added to the preferences, which are already huge... Since this is not something of direct impact to the user workflow, there is probably little need to customize them very often. The main use here would be to not look bad in dark themes, so that would be totally a job for stylesheets.
Maybe we should start documenting styling options somewhere too...

adrianinsaval:
I would like to have the option to have it in a parameter, I use FreeCAD with the system theme and that likely wouldn't provide any FreeCAD specific colors, or maybe use some standard color name that most themes will provide?

chennes:
I don't know that there's really an appropriate `QPalette` entry for this use: I guess maybe `QPalette::Highlight` or `QPalette::BrightText`?

adrianinsaval:
maybe Link or LinkVisited? or will blue/magenta be confusing?

realthunder:
So, in summarize, the ideal solution should provide a configurable color setting from both stylesheet and FC preference/parameter, and yet be able to provide a default value that is usable in both dark and light theme.

Here is my proposed solution:

For each color option, there shall be two FC parameters, one integer (or string) for storing the color, the other boolean to let user choose whether the color can be overridden in stylesheet. The two parameters can have the same name (or not).

Expose the color option to stylesheet by defining a `qproperty` in the relevant widget. It is actually quite tricky to determine whether the current stylesheet contains any setting for a `qproperty`. If not handled correctly, then the color from the last stylesheet may be wrongly retained when switching to another stylesheet witch didn't have the setting. I made a recent [commit](https://github.com/realthunder/FreeCAD/commit/14062dcce9d6d9547d1b0364d87b485a166791d6) in my branch in attempt to solve this problem. It works like this. Two boolean flags are used. One used in [property setter](https://github.com/realthunder/FreeCAD/blob/14062dcce9d6d9547d1b0364d87b485a166791d6/src/Mod/Part/Gui/TaskAttacher.cpp#L467) to flag whether the property has been written. The other flag [copies](https://github.com/realthunder/FreeCAD/blob/14062dcce9d6d9547d1b0364d87b485a166791d6/src/Mod/Part/Gui/TaskAttacher.cpp#L1121) the first flag when handling `QEvent::StyleChange` event, and then reset the first flag.

The actual color is applied in response to `QEvent::PaletteChange` with a timer, because of another potential complication, which is to provide a default color that is usable in both dark and light theme. My commit above (which deals with the green or red message in TaskAttacher) also attempts to solve this problem by [checking](https://github.com/realthunder/FreeCAD/blob/14062dcce9d6d9547d1b0364d87b485a166791d6/src/Mod/Part/Gui/TaskAttacher.cpp#L425) on the button text color. If the color hsv value is greater than 128, then consider it as bright text, and the default green message color is lighten to make it more visible in dark theme. This method may or may not be applicable to auto tuning of other widget color.

chennes:
That's the basics of it, yes -- though I should note that the few stylesheet authors I spoke with when I created StatefulLabel were of the opinion that it was the stylesheet author's responsibility to make sure to set all of the appropriate qproperties, so I didn't bother ensuring that each one was set. If you want to implement the more robust solution I won't stop you, though 😄 .

dirkolbrich:
I came here from the @FreeCADNews tweet and would like to add a comment about **Accessibility** to this PR.
Making the list property distinguishable _only_ by color could have a negative impact for color vision impaired users. If it becomes an important workflow feature over time, these users are left stranded. 

Have you guys thought about not merely changing the colour of a list property, but also making the list entry distinguishable by some other form, e.g. by adding additional characters around the name like using  angled brackets `<Map Mode>` instead of `Map Mode`?

luzpaz:
Sorry to go off-topic but can anyone volunteer to make Accessibility stylesheet(s) for color vision impaired users?  I saw a QGIS YT clip talking about the new features in the latest QGIS and snapped a screenshot that was relevant. 
image

yorikvanhavre:
> Have you guys thought about not merely changing the colour of a list property, but also making the list entry distinguishable by some other form, e.g. by adding additional characters around the name like using angled brackets for <Map Mode> instead of Map Mode?

I think that would quickly become cluttered as there are different colors/cases. But at some point a kind of icon column could become cool, a bit like the visibility icon (I think) @realthunder has been working on in his branch.

> Sorry to go off-topic but can anyone volunteer to make Accessibility stylesheet(s) for color vision impaired users?

Seems like a good idea, but this should go to the forum...

> For each color option, there shall be two FC parameters, one integer (or string) for storing the color, the other boolean to let user choose whether the color can be overridden in stylesheet. The two parameters can have the same name (or not).

Is that not a bit overkill? I mean if you (as a user) use a stylesheet, it's somehow implied that the stylesheet will override anything, having been allowed or not. Indeed there might be problems with stylesheets, and some new elements won't be properly themed by older stylesheets, but that's just a matter of fixing them afterwards...

IMHO these colors are not so important that they need very fine tuning like you propose... In most cases, users will want the stylesheet to solve all the interface colors without the need to tweak anything.

I would propose to split the problem into two. Merge this PR now as-is, with its hard-coded green color (will work on both light and dark themes so it's good enough for now, I suppose), and then we work on a better solution that address all the tree colors at once...

yorikvanhavre:
I think this has been committed in master already... (Check the file conflict below..). Ok to close this, @realthunder ?

realthunder:
Yes, sure

PR #5193, opened 2021-11-22
 [PD] Support for point sections in pipe/sweep

AjinkyaDahale:
Similar changes to PR #5170, except uses `BRepOffsetAPI_MakePipeShell` instead.

Not ready for merge, but in good enough condition for tests. Certain bugs exist which I will describe in comments as I find them.

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePipe.cpp
+++ b/src/Mod/PartDesign/App/FeaturePipe.cpp
@@ -71,8 +71,8 @@
 
 using namespace PartDesign;
 
-const char* Pipe::TypeEnums[] = {"FullPath","UpToFace",NULL};
-const char* Pipe::TransitionEnums[] = {"Transformed","Right corner", "Round corner",NULL};
+const char* Pipe::TypeEnums[] = {"FullPath", "UpToFace", NULL};
+const char* Pipe::TransitionEnums[] = {"Transformed", "Right corner", "Round corner", NULL};
 const char* Pipe::ModeEnums[] = {"Standard", "Fixed", "Frenet", "Auxiliary", "Binormal", NULL};
 const char* Pipe::TransformEnums[] = {"Constant", "Multisection", "Linear", "S-shape", "Interpolation", NULL};
 
@@ -112,26 +112,54 @@ short Pipe::mustExecute() const
 
 App::DocumentObjectExecReturn *Pipe::execute(void)
 {
-    std::vector<TopoDS_Wire> wires;
-    try {
-        wires = getProfileWires();
-    } catch (const Base::Exception& e) {
-        return new App::DocumentObjectExecReturn(e.what());
-    }
+    auto getSectionShape =
+        [](App::DocumentObject* feature, const std::vector<std::string> &subs) -> TopoDS_Shape {
+            if (!feature ||
+                !feature->isDerivedFrom(Part::Feature::getClassTypeId()))
+                throw Base::TypeError("Pipe: Invalid profile/section");
 
-    TopoDS_Shape sketchshape = getVerifiedFace();
-    if (sketchshape.IsNull())
-        return new App::DocumentObjectExecReturn("Pipe: No valid sketch or face as first section");
-    else {
-        //TODO: currently we only allow planar faces. the reason for this is that with other faces in front, we could
-        //not use the current simulate approach and build the start and end face from the wires. As the shell
-        //begins always at the spine and not the profile, the sketchshape cannot be used directly as front face.
-        //We would need a method to translate the front shape to match the shell starting position somehow...
-        TopoDS_Face face = TopoDS::Face(sketchshape);
-        BRepAdaptor_Surface adapt(face);
-        if (adapt.GetType() != GeomAbs_Plane)
-            return new App::DocumentObjectExecReturn("Pipe: Only planar faces supported");
-    }
+            auto subName = subs.empty() ? "" : subs.front();
+
+            // only take the entire shape when we have a sketch selected, but
+            // not a point of the sketch
+            if (feature->isDerivedFrom(Part::Part2DObject::getClassTypeId()) &&
+                !(subName.compare(0, 6, "Vertex") == 0))
+                return static_cast<Part::Part2DObject*>(feature)->Shape.getValue();
+            else {
+                if(subName.empty())
+                    throw Base::ValueError("Pipe: No valid subelement linked in Part::Feature");
+                return static_cast<Part::Feature*>(feature)->Shape.getShape().getSubShape(subName.c_str());
+            }
+        };
+
+    auto addWiresToWireSections =
+        [](TopoDS_Shape& section,
+           std::vector<std::vector<TopoDS_Shape>>& wiresections) -> size_t {
+            TopExp_Explorer ex;
+            size_t i=0;
+            bool initialWireSectionsEmpty = wiresections.empty();
+            for (ex.Init(section, TopAbs_WIRE); ex.More(); ex.Next(), ++i) {
+                // if profile was just a point then this is where we can first set our list
+                if (i>=wiresections.size()) {
+                    if (initialWireSectionsEmpty)
+                        wiresections.emplace_back(1, ex.Current());
+                    else
+                        throw Base::ValueError("Pipe: Sections need to have the same amount of inner wires (except profile and last section, which can be points)");
+                }
+                else
+                    wiresections[i].push_back(TopoDS::Wire(ex.Current()));
+            }
+            return i;
+        };
+
+    // TODO: currently we can only allow planar faces, so add that check.
+    // The reason for this is that with other faces in front, we could not use the
+    // current simulate approach and build the start and end face from the wires.
+    // As the shell begins always at the spine and not the profile, the sketchshape
+    // cannot be used directly as front face. We would need a method to translate
+    // the front shape to match the shell starting position somehow...
+    std::vector<TopoDS_Wire> wires;
+    TopoDS_Shape profilePoint;
 
     // if the Base property has a valid shape, fuse the pipe into it
     TopoDS_Shape base;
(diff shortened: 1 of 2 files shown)


AjinkyaDahale:
See attached files. In `lofttest-5193-1` the profile is a point and it has two more sections, with "orientation mode" set to "Fixed". This pipe succeeds. However, in `lofttest-5193-2`, with the same cross sections, but reversed (so the hexagon is the profile and the other two are added as multi-sections), the "front" face is not formed.

Additionally, with "orientation mode" set to "Standard" or "Frenet" the ends of the pipe do not end in a plane and again the face is not formed.

AjinkyaDahale:
I think I found the cause for the bug and fixed it in c74f47cb2508e08877ec3083e788c6bfd4e885fd. @donovaly can you try this out whenever you're free. I'm sure there are a whole bunch of cases where things won't work and I would like some help by getting concrete test cases.

donovaly:
> I think I found the cause for the bug and fixed it in [c74f47c](https://github.com/FreeCAD/FreeCAD/commit/c74f47cb2508e08877ec3083e788c6bfd4e885fd). @donovaly can you try this out whenever you're free. I'm sure there are a whole bunch of cases where things won't work and I would like some help by getting concrete test cases.

I won't have time right now. There were too many new things added to PD recently and these need to be sorted out. I am also busy in releasing FC 0.19.3. So please give me some days.

AjinkyaDahale:
> I won't have time right now. There were too many new things added to PD recently and these need to be sorted out. I am also busy in releasing FC 0.19.3. So please give me some days.

That's OK. I don't think this is blocking things I'm planning right now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5193">pipeline status</a> for feature branch [PR_5193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5193). Pipeline [414384473](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/414384473) was triggered at [46e5f65](https://github.com/FreeCAD/FreeCAD/pull/5193/commits/46e5f653cc6e67aebc966ea7a42812896d56d5dc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I will find time for this tomorrow.

Can you please rebase the PR to latest master since there were a lot of changes to PD the last days and maybe you could squash the different commits to one?

What is the status? Is the PR ready to be reviewed or where do you see issues?

AjinkyaDahale:
> I will find time for this tomorrow.

Thanks!
 
> Can you please rebase the PR to latest master since there were a lot of changes to PD the last days and maybe you could squash the different commits to one?

Will take a look maybe squash is possible.
 
> What is the status? Is the PR ready to be reviewed or where do you see issues?

I think the issues are solved last I recall.

donovaly:
Can you please rebase you PR to current master? As it is it is based on master from Monday and I cannot compile because of merge conflicts.

src/Mod/PartDesign/App/FeaturePipe.cpp
```diff
@@ -314,7 +374,7 @@ App::DocumentObjectExecReturn *Pipe::execute(void)
         //result.Move(invObjLoc);
         AddSubShape.setValue(result);
 
-        if (base.IsNull()) {
+        if (base.IsNull()) {
```
donovaly, line 377: You added a strange character

src/Mod/PartDesign/App/FeaturePipe.cpp
```diff
@@ -272,31 +325,38 @@ App::DocumentObjectExecReturn *Pipe::execute(void)
...
+            }
+            for (auto& s : shells)
                 sewer.Add(s);
 
             sewer.Perform();
             mkSolid.Add(TopoDS::Shell(sewer.SewedShape()));
         } else {
             // shells are already closed - add them directly
-            for (TopoDS_Shape& s : shells) {
+            for (auto& s : shells) {
```
donovaly, line 359: Why this change? the explicit type name helps others to understand the code. Therefore as long as the type name is there, we can keep it.

src/Mod/PartDesign/Gui/Command.cpp
```diff
@@ -1048,15 +1049,22 @@ void prepareProfileBased(PartDesign::Body *pcActiveBody, Gui::Command* cmd, cons
...
+                 which.compare("SubtractivePipe") == 0) {
+            // for additive and subtractive pipes set subvalues even for sketches
+            // to support point sections
+            auto subName = subs.empty() ? "" : subs.front();
+
+            // `ProfileBased::getProfileShape()` and other methods will return
+            // just the sub-shapes if they are set. So when whole sketches are
+            // desired, don not set sub-values.
+            if (feature->isDerivedFrom(Part::Part2DObject::getClassTypeId()) &&
+                !(subName.size() > 6 && subName.substr(0,6) == "Vertex"))
```
donovaly, line 1062: Where does the 6 comes from? Can you please add a comment?

nit-pick: our style is to have a space after commas.

donovaly (changes requested):
Looks good. As I wrote you must rebase since I cannot compile.

I have only minor annotations.

src/Mod/PartDesign/Gui/Command.cpp
AjinkyaDahale: 6 is for the characters in "Vertex". A snippet copied from elsewhere in PD, which is why the space-less comma.

AjinkyaDahale:
> Can you please rebase you PR to current master? As it is it is based on master from Monday and I cannot compile because of merge conflicts.

As far as I can tell it is already rebased on to FC master.

Screenshot at 2021-11-28 11-20-48

src/Mod/PartDesign/App/FeaturePipe.cpp
AjinkyaDahale: That would be my emacs again. This isn't the first time it happened and unfortunately won't be the last.

src/Mod/PartDesign/App/FeaturePipe.cpp
AjinkyaDahale: I think this one can stay. The for loop already says that `s` runs through every element of `shells`.

src/Mod/PartDesign/Gui/Command.cpp
donovaly: You use the same code for lofts.

But is this check safe? What subNames could we have?

src/Mod/PartDesign/App/FeaturePipe.cpp
donovaly: This is nit-picking, I know but once the explicit type name is there, just keep it.

donovaly:
> As far as I can tell it is already rebased on to FC master.

Indeed, I deleted now my local branch for your PR and recreated it and now its history shows it is based on master. It seems that the tool "GitHub Desktop" I use has some issues with stashing.

src/Mod/PartDesign/Gui/Command.cpp
AjinkyaDahale: From https://www.cplusplus.com/reference/string/string/substr/:

> len
>    Number of characters to include in the substring (if the string is shorter, as many characters as possible are used).

So we probably don't need the length check.

Looking at that site, I see a clearer (and probably faster and safer) method https://www.cplusplus.com/reference/string/string/compare/.

As for what other subNames we can have, it includes "Edge<number>" and "Face<number>". I expect after topo-naming is merged the <number> part will be replaced by a description of where those entities are within the solid but the first part remain the same. If that's not the case, we'll have to address it in the future.

donovaly:
I could now compile and there is only one missing feature -> for a further PR:

- Take this file:
[loftbug.zip](https://github.com/FreeCAD/FreeCAD/files/7613956/loftbug.zip)
- open there the sweep and make it a multisection
- select as first section the vertex, then the circle

result: you correctly chatched this mistake case:
FreeCAD_WJ6InnY043

Now the user knows what to do but he cannot move the order of the sections in the dialog like he can for the sections of lofts:

opSGxOh06e

src/Mod/PartDesign/Gui/Command.cpp
donovaly: Since you also check for "Vertex", you can omit the length check I think.
Please decide.

From my point of view it is then ready to be merged.

AjinkyaDahale:
> I could now compile and there is only one missing feature -> for a further PR:
> 
>     * Take this file:
>       [loftbug.zip](https://github.com/FreeCAD/FreeCAD/files/7613956/loftbug.zip)
> 
>     * open there the sweep and make it a multisection
> 
>     * select as first section the vertex, then the circle
> 
> 
> result: you correctly chatched this mistake case: 

So the mistake was caught by OCCT (that's the only place where the term "punctual section" is used). For some reason the algorithm requires that only the first and last added sections can be punctual sections.

src/Mod/PartDesign/App/FeaturePipe.cpp
AjinkyaDahale: If you insist. I'll revert this change and force push shortly.

src/Mod/PartDesign/Gui/Command.cpp
AjinkyaDahale: So this is what I'll do:
1. In `FeaturePipe` the comparison only relates to pipe, so I'll change to using `compare`
2. In `Command.cpp` and other places where there are other occasions where this snippet is used, I'll instead make a separate PR.

Does that sound good?

donovaly: Very good.

AjinkyaDahale:
> Now the user knows what to do but he cannot move the order of the sections in the dialog like he can for the sections of lofts:

For other (i.e. not "punctual") sections, the algorithm is fairly forgiving. For example, the pipe will still work the same regardless of whether you add the rectangle first or the circle (still compiling to confirm this, but fairly sure of it). My guess is the algo guesses an appropriate position for the section along the path.

donovaly:
I'll merge the PR now.

Concerning the multisection ordering feature, I will make a PR for this

donovaly:
> Concerning the multisection ordering feature, I will make a PR for this

Here is the PR: https://github.com/FreeCAD/FreeCAD/pull/5207
Can you please test it?

AjinkyaDahale:
> Concerning the multisection ordering feature, I will make a PR for this

Alright. I'll leave it to you.

donovaly:
> Alright. I'll leave it to you.

Seems we had a mid-air collision since the same time you sent your post I sent one with the PR. Please try this out.

src/Mod/PartDesign/Gui/Command.cpp
AjinkyaDahale: See https://github.com/FreeCAD/FreeCAD/pull/5208.

PR #5199, opened 2021-11-24
[PD] Helix: Offer profile's normal as axis

jbaehr:
Previously, only the vertical and horizontal axis of the profile was
selectable in the task panel of the additive/subtractive helix. Now the
profile's normal axis can be selected, too. This now allows to create
helical extrusions or "twisted pockets".

The order of the GUI entries was chosen to be in line with the order of
axis selection of the multi-transform parameters.
The actual feature's implementation needed adaption for this special case
as in some places a unit vector (`gp_Dir`) was derived from the cross-
product of axis and profile normal -- a null-vector when the axis *is*
the normal. This caused the gp_Dir's ctor to throw.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -371,6 +371,20 @@ TopoDS_Shape Helix::generateHelixPath(void)
 
     Base::Vector3d normal = getProfileNormal();
     Base::Vector3d start = v.Cross(normal);  // pointing towards the desired helix start point.
+
+    // if our axis is (nearly) aligned with the profile's normal, we're only interested in the "twist"
+    // of the helix. The actual starting point, and thus the radius, isn't important as long as it's
+    // somewhere in the profile's plane: an arbitrary vector perpendicular to the normal.
+    if (start.IsNull()) {
+        auto hopefullyNotParallel = Base::Vector3d(1.0, 2.0, 3.0);
+        start = normal.Cross(hopefullyNotParallel);
+        if (start.IsNull()) {
+            // bad luck
+            hopefullyNotParallel = Base::Vector3d(3.0, 2.0, 1.0);
+            start = normal.Cross(hopefullyNotParallel);
+        }
+    }
+
     gp_Dir dir_start(start.x, start.y, start.z);
 
     // Find out in what quadrant relative to the axis the profile is located, and the exact position.
@@ -447,6 +461,12 @@ TopoDS_Shape Helix::generateHelixPath(void)
 // this function calculates self intersection safe pitch based on the profile bounding box.
 double Helix::safePitch()
 {
+    Base::Vector3d v = Axis.getValue();
+    Base::Vector3d n = getProfileNormal();
+    Base::Vector3d s = v.Cross(n);  // pointing towards the desired helix start point.
+    if (s.IsNull())
+        return Precision::Confusion(); // if the axis orthogonal to the profile, any pitch >0 is safe
+
     // Below is an approximation. It is possible to do the general way by solving for the pitch
     // where the helix is self intersecting.
 
@@ -461,14 +481,11 @@ double Helix::safePitch()
 
     double X = Xmax - Xmin, Y = Ymax - Ymin, Z = Zmax - Zmin;
 
-    Base::Vector3d v = Axis.getValue();
     gp_Dir dir(v.x,v.y,v.z);
     gp_Vec bbvec(X, Y, Z);
 
     double p0 = bbvec*dir; // safe pitch if angle=0
 
-    Base::Vector3d n = getProfileNormal();
-    Base::Vector3d s = v.Cross(n);  // pointing towards the desired helix start point.
     gp_Dir dir_s(s.x, s.y, s.z);
 
     if (tan(abs(angle))*p0 > abs(bbvec*dir_s))
--- a/src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
@@ -202,6 +202,7 @@ void TaskHelixParameters::fillAxisCombo(bool forceRefill)
         PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
         Part::Part2DObject* pcSketch = dynamic_cast<Part::Part2DObject*>(pcFeat->Profile.getValue());
         if (pcSketch){
+            addAxisToCombo(pcSketch,"N_Axis",QObject::tr("Normal sketch axis"));
             addAxisToCombo(pcSketch,"V_Axis",QObject::tr("Vertical sketch axis"));
             addAxisToCombo(pcSketch,"H_Axis",QObject::tr("Horizontal sketch axis"));
             for (int i=0; i < pcSketch->getAxisCount(); i++) {
--- a/src/Mod/PartDesign/Gui/TaskHelixParameters.ui
+++ b/src/Mod/PartDesign/Gui/TaskHelixParameters.ui
@@ -72,6 +72,11 @@
        </item>
        <item>
         <property name="text">
+         <string>Normal sketch axis</string>
+        </property>
+       </item>
+       <item>
+        <property name="text">
          <string>Select reference...</string>
         </property>
        </item>


donovaly:
Many thanks! Good feature addition. The code looks god so far but I cannot proper review it until Sunday.

I requested now a review from the author of the helix feature: @davidosterberg .

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -450,6 +464,12 @@ double Helix::safePitch()
     // Below is an approximation. It is possible to do the general way by solving for the pitch
     // where the helix is self intersecting.
 
+    Base::Vector3d v = Axis.getValue();
+    Base::Vector3d n = getProfileNormal();
+    Base::Vector3d s = v.Cross(n);  // pointing towards the desired helix start point.
+    if (s.IsNull())
+        return 0.0; // if the axis orthogonal to the profile, any pitch >0 is safe
```
0penBrain, line 471: It's inconsistent between the comment and the returned value.

src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
```diff
@@ -202,6 +202,7 @@ void TaskHelixParameters::fillAxisCombo(bool forceRefill)
         PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
         Part::Part2DObject* pcSketch = dynamic_cast<Part::Part2DObject*>(pcFeat->Profile.getValue());
         if (pcSketch){
+            addAxisToCombo(pcSketch,"N_Axis",QObject::tr("Normal sketch axis"));
```
0penBrain, line 205: Shall also be added in retranslateUI()

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: Indeed, at the (only) usage of this "safePitch" value, in `TaskHelixParameters::updateUI()`, it is compared against the current pitch value like this:
```cpp
        if (pcHelix->safePitch() > propPitch->getValue())
            status = "Warning: helix might be self intersecting";
```
So an actual pitch value of exactly 0.0 can slip though, i.e. without a warning status message. A zero pitch is used then creating a flat spiral, but this again doesn't make any sense with an axis normal to the profile.

Would `Precision::Confusion()` be a sensible value to return instead?

0penBrain: Actually I don't know which value is sensible. Maybe 0.0 is. Just do not at the same time 'return 0.0' and comment 'any value strictly greater than 0 is OK'. ;)

src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
jbaehr: Can you please give a bit more details on this? It's my first contact with FreeCAD's source and I'm still learning. My previous C++/Qt experiences are [long ago](https://github.com/KDE/krusader).

Do you mean adding this "Normal sketch axis" entry to the respective combobox in the TaskHelixParameters.ui file? What effect does it have and how can I verify it?

0penBrain: In the same file there is a function named 'retranslateUi()' that is called on language change. UI items shall also be there. ;)

jbaehr: I can only find this reference
```cpp
void TaskHelixParameters::changeEvent(QEvent *e)
{
    TaskBox::changeEvent(e);
    if (e->type() == QEvent::LanguageChange) {
        ui->retranslateUi(proxy);
    }
}
```
which made me come up the question about the .ui file, where the class behind the `ui` pointer is generated from.

0penBrain: @jbaehr Yes sorry, I messed a bit things up. Indeed it shall be added in the UI file so it is added in the retranslateUi() function at compile time. ;)

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: I decided to go for `Precision::Confusion()`. The alternative, i.e. returning 0.0 and change the usage from `>` to `>=` (and maybe rename to `safePitchLimit`), seems more risky to me.

src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
jbaehr: done

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5199">pipeline status</a> for feature branch [PR_5199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5199). Pipeline [417423101](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/417423101) was triggered at [08e2687](https://github.com/FreeCAD/FreeCAD/pull/5199/commits/08e268786b73f26b9937b911e3f4104be1c80239). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Many thanks again. This is a valuable feature.

The feature works but is not yet complete. What is missing is:
- select a face of a part
- click the Additive Helix button
result: under 'Axis' there is no entry to select the face's normal vector

Then there is an issue and a bug:
- select the sketch in the attached file: [hexagon-pad.zip](https://github.com/FreeCAD/FreeCAD/files/7612178/hexagon-pad.zip)
- click the Additive Helix button
result: you get an error since the vertical direction is preselected:
FreeCAD_x7NN0r5xJ7
Having the vertical direction as preselection for sketches is not sensible in my opinion soince as user I want to get a sensible preview of a helix, and not by default get no preview since the default direction is geometrically not sensible. Therefore the normal direction should be the default.

However, now
- change the direction to normal
result: you get a valid helix and preview of it, but the dialog still states there is an error:
FreeCAD_cDbiQ6T1WI

Can you please have a look? It is OK to merge feature additions in steps of several PRs. So you can postpone the handling of faces to a further PR. But to merge this PRs, the dialog preset and wrong error message should be fixed.

adrianinsaval:
that default is not sensible for that specific sketch but we can't expect FreeCAD to pick defaults depending on each sketch, considering the original intente of the helix feature the vertical sketch axis is a sensible default, just move your circle out of the vertical axis and you would see.

donovaly:
> that default is not sensible for that specific sketch but we can't expect FreeCAD to pick defaults depending on each sketch, considering the original intente of the helix feature the vertical sketch axis is a sensible default, just move your circle out of the vertical axis and you would see.

I know. But since I introduced recently FreeCAD to new users I kwon what they expect - a visial feedback upon that they can react. They are lost when they don't see anything and thus don't know what the mistake is.

So maybe I was not clear, the default as it is should stay but for cases where the default leads to an invalid object since this axis is perpendicular to the sketch normal, the sketch normal should be used as default.
We do this check for example in the pad feature, see https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/PartDesign/App/FeaturePad.cpp#L132

Previously we could not do this but thanks to your work we have the normal as fallback default.

jbaehr:
First of all: Thank you all for taking the time to review!

@donovaly, regarding the missing "face normal" entry, I'll take a look. I usually avoid basing anything on generated geometry, so I hadn't this on my radar.

The bug, that the status message is not updated seems unrelated to my addition; I can reproduce it in an older v0.20.26155. I can try to look at this, too, but I'd prefer to handle this independently of this PR.

The same holds true for changing the default: I completely agree that the default values must not result in an invalid shape or similar error. However, this issue is also independent of my PR and existed previously, too.

My addition now enables an "aways working" (i.e. never self-intersecting) default: The profile's normal. As @adrianinsaval points out, this would be a pretty unexpected axis then the aim is to create spring-like objects. So such a change should be discussed with the community (in "open discussion" forum) first and not mixed with this PR.

An alternative I could imagine is to determine the default axis based on the location of the profile's center-of-mass: If it's "close" to the origin, use the normal; if it's close to the V-axis, take the H-axis; otherwise use the V-axis. I'll try to implement a prototype but I prefer to keep this separate of this PR here.

donovaly:
> The bug, that the status message is not updated seems unrelated to my addition; I can reproduce it in an older v0.20.26155. I can try to look at this, too, but I'd prefer to handle this independently of this PR.

Then let's do it in different PRs:
- I will now merge this one.
- please make a PR to support also faces
- try to have a look why there is the wrong error message
- I will have a look for a better default direction

OK?

src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
wwmayer: It's pointless to add any items to the combo box in the .ui file because when opening the helix dialog the function fillAxisCombo() is called that clears the combo box and re-adds the items. At runtime when selecting a reference axis further items might be added.

The function changeEvent() is not fully implemented to really support a language change. At the moment when you change the language in the preferences while the helix dialog is open it will clear the combo box and only re-adds the items of the .ui file. However, all items that are added by reference axes will be gone and this can cause serious problems because then the lengths of the combo box and the internal array `axesInList` don't match any more.

If you e.g. select an edge of the model as reference axis, change the language and select the same reference axis again then the current index of the combo box is set to -1 because it cannot find the item any more. Now if -1 is used to access the `axesInList` a segmentation fault is the consequence.

In order to fully support a language change a few things must be done:
* save the current index of the combo box
* save the items that come from reference axes
* call retranslateUi()
* restore the items
* restore the current index

0penBrain: @wwmayer my whole mistake. Thanks for jumping in.
@jbaehr sorry for the useless noise on this

wwmayer: 41efee1b4c implements the language change. However, when the user has selected several reference axes then maximum one of them can be preserved. Trying to preserve all of them would lead to ugly code and since this issue is anyway quite uncommon I think we can leave like this.

jbaehr: Thank you.

wwmayer:
> result: under 'Axis' there is no entry to select the face's normal vector

What should this be good for? For a helix you cannot only have a direction vector but you also need a base point to get the reference axis. For a face there is not a unique base point that could be used.

donovaly:
I think we should distinguish the PR from the, let's say "improvable" ;-) state of Helix in general.
I merged the PR since it just uses the existing infrastructure. We also discussed that there should be next steps to fix several issues we encountered.

Many thanks Werner to improve the infrastructure!

I first looked yesterday closely to the Helix feature and I saw a lot of issues but did not yet have the time, therefore I only fixed some easy things.

>> result: under 'Axis' there is no entry to select the face's normal vector
> 
> What should this be good for? For a helix you cannot only have a direction vector but you also need a base point to get the reference axis. For a face there is not a unique base point that could be used.

Why is a sketch different in this respect? The PR allows to use the sketch normal as direction but the axis always runs through the origin. That is a fair deal in my opinion.
When I take now a sketch of my choice, loft it and eventually select now the base face of the loft object, I have the same 2D geometry for the Helix feature as with the Sketch itself. Therefore also faces should have the normal direction entry.
Here is an example file: [Helix-Tests.zip](https://github.com/FreeCAD/FreeCAD/files/7622030/Helix-Tests.zip)
I expect there the same functionality when I select either "Face1" of the Loft or the Sketch. But as it is I cannot helix along the normal of Face1.

-------------------------

I encounter now a crash and reported in the forum: https://forum.freecadweb.org/viewtopic.php?f=10&t=64160

jbaehr:
> I think we should distinguish the PR from the, let's say "improvable" ;-) state of Helix in general.

Yes, I also wanted to enhance the descriptions of the properties, right now, they only repeat their names.
I also want to create tests for the helix, which are entirely missing. Till now I failed for other reasons: My self-build FreeCAD has some issues with Shiboken2 ("<class 'RuntimeError'>: Failed to wrap widget").

> > > result: under 'Axis' there is no entry to select the face's normal vector
> > 
> > 
> > What should this be good for? For a helix you cannot only have a direction vector but you also need a base point to get the reference axis. For a face there is not a unique base point that could be used.
> 
> Why is a sketch different in this respect? The PR allows to use the sketch normal as direction but the axis always runs through the origin. That is a fair deal in my opinion. When I take now a sketch of my choice, loft it and eventually select now the base face of the loft object, I have the same 2D geometry for the Helix feature as with the Sketch itself. Therefore also faces should have the normal direction entry.

Each (flat) face has (a) normal vector, but there is no well-defined base to make an axis from that vector. The "normal axis" of a sketch has such a base: the sketch's origin. But for a face, there is no "origin". A seemingly obvious choice would be the face's center of mass. However, for all but symmetrical profiles this is highly arbitrary, sensitive to changes in the profile's shape, and lacks a way for the user to control it. (Introducing ways to control it would essentially recreate what we already have: attach a datum axis to the face and use the attachment offset to control it.)  
The more I think about, the more I come to the conclusion that it doesn't make much sense; esp. as it's "only" for a workflow that's not recommended anyway: Basing features on generated geometry.

donovaly:
> Each (flat) face has (a) normal vector, but there is no well-defined base to make an axis from that vector. The "normal axis" of a sketch has such a base: the sketch's origin. But for a face, there is no "origin".

We have the Part's coordinate system so we can use the same as we do for sketches - take the normal vector and fix it so that it runs through the coordinate center.

But OK, as you and also @wwmayer think this provides no benefit, then we can leave it as it is and do not provide a normal for faces.

PR #5201, opened 2021-11-25
[Gui] UserEditMode : add option in Preferences + fix for PartDesign objects

0penBrain:


adrianinsaval:
maybe you mixed some things? the preference you added seems to be about the decimal substitution not the user edit mode

0penBrain:
@adrianinsaval definitely mixed things. Too much at the same time... Closing, will reopen 2 PR appropriately. Thanks for the catch.

PR #5207, opened 2021-11-28
[PD] allow to move sweep sections

donovaly:
Now that we have the feature that sweeps can have vertices as end or begin of a sweep, there is the need for the feature to move sections. For example vertices may only be the last section.

Here is a use case:
t35WnSwed6

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskPipeParameters.cpp
@@ -938,6 +938,9 @@ TaskPipeScaling::TaskPipeScaling(ViewProviderPipe* PipeView, bool /*newObj*/, QW
     ui->listWidgetReferences->setContextMenuPolicy(Qt::ActionsContextMenu);
     connect(remove, SIGNAL(triggered()), this, SLOT(onDeleteSection()));
 
+    connect(ui->listWidgetReferences->model(),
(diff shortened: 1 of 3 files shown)


donovaly:
@AjinkyaDahale

AjinkyaDahale:
> @AjinkyaDahale GitHub doesn't let me to request you to review the PR the "normal way", thus I inform you this way.

I don't know how someone is added as a reviewer. I guess @luzpaz may have some idea.

AjinkyaDahale:
Tested with a couple examples and appears to work well.

donovaly:
Thanks for testing.

From the code point of view it uses the same code we have already to move loft sections -> merging

donovaly:
OK, to be able to request a review from you, it seems one must explicitly ask for this. Can you therefore please open this site and there on the button "Join Request"?:
https://github.com/orgs/FreeCAD/teams/reviewers/members

AjinkyaDahale:
> OK, to be able to request a review from you, it seems one must explicitly ask for this. Can you therefore please open this site and there on the button "Join Request"?: https://github.com/orgs/FreeCAD/teams/reviewers/members

Gives me a 404. Can only go up to https://github.com/orgs/FreeCAD/ which points to the FC Github page.

donovaly:
So neither 
https://github.com/orgs/FreeCAD/teams/reviewers/
nor
https://github.com/orgs/FreeCAD/teams/

But no problem, I requested myself to become a reviewer. Maybe then I am allowed to add you. I also asked that the maintainers add you directly.

PR #5208, opened 2021-11-28
[PD] Use `compare` to compare strings instead of `substr`

AjinkyaDahale:
Used when finding subelements of a feature. Many of the comparisons used to also
check for string lengths, but as far as I can tell they are not strictly
necessary (see https://www.cplusplus.com/reference/string/string/substr/) and
just `substr` can be used without them. However, `compare` explicitly is for
comparing, and does not make a new object that `substr` does.


donovaly:
Many thanks. merging, since I triggered/requested this PR.

AjinkyaDahale:
> Many thanks. Merging since I triggered/requested this PR.

Thanks! Should probably have gone through a bit more testing, though, since it makes the change everywhere in PD. Then again, the changes are relatively minor. Guess we'll deal with reports when they come :laughing:.

donovaly:
The changes are in my opinion straight forward and I think this is better code. I compiled it and tested the result on a file.

PR #5209, opened 2021-11-28
Add defaults for the 3DConnexion SpaceExplorer

broettge:
I modified the src/gui/3DConnexion/3DConnexion.xml file to support the SpaceExplorer.
This provides default button mappings consistent with other models already supported.


luzpaz:
Need any git help?

PR #5211, opened 2021-11-28
[PD] allow negative helix growth

donovaly:
Helices that become smaller with every turn are geometrically perfectly valid. Therefore we cannot forbid this.

(For example when creating a helix from a face you often cannot move it so that you can apply a positive growth.)

Here is a use case: [Helix-Tests.zip](https://github.com/FreeCAD/FreeCAD/files/7614657/Helix-Tests.zip)

FreeCAD_m8ocjYwv0D

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -134,13 +134,12 @@ App::DocumentObjectExecReturn *Helix::execute(void)
         if (Turns.getValue() < Precision::Confusion())
             return new App::DocumentObjectExecReturn("Error: turns too small!");
         if ((Height.getValue() < Precision::Confusion())
-            && (Growth.getValue() < Precision::Confusion()))
+            && (abs(Growth.getValue()) < Precision::Confusion()))
             return new App::DocumentObjectExecReturn("Error: either height or growth must not be zero!");
         Pitch.setValue(Height.getValue()/Turns.getValue());
     } else {
         return new App::DocumentObjectExecReturn("Error: unsupported mode");
     }
-
 
     TopoDS_Shape sketchshape;
     try {
@@ -407,7 +406,7 @@ TopoDS_Shape Helix::generateHelixPath(void)
     bool growthMode = std::string(Mode.getValueAsString()).find("growth") != std::string::npos;
     double radiusTop;
     if (growthMode)
-        radiusTop = radius + turns*growth;
+        radiusTop = radius + turns * growth;
     else
         radiusTop = radius + height * tan(Base::toRadians(angle));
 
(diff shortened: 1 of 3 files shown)


donovaly:
@0penBrain and @AjinkyaDahale could you please test this? I think it is straight forward, but I might have overseen something.

AjinkyaDahale:
I think the biggest issue is of self intersection (as the self intersection warning already says). For example, increasing the turns to more than 4 starts the turns to grow in negative direction and make complications. That said, even having too low heights can cause the issue with zero or positive "growth", so a negative growth shouldn't be a discouraged on those grounds. 

Note that in the height-turns-angle configuration we already can choose a negative angle and get the same effect as negative growth.

It may probably also be a good idea to also look into the logic of why the warning comes up. This particular case is clearly not self intersecting, so perhaps logic can be improved if handled by FC and not OCCT.

donovaly:
> Note that in the height-turns-angle configuration we already can choose a negative angle and get the same effect as negative growth.

I am aware of this. The Growth mode is new in FC 0.20 and I already fixed some of its issues today. With this PR I only want to make it allowing what is geometrically possible.

Currently the warning about self-intersecting does not work correctly of the Growth mode and this is on my ToDo list. However step by step.
With @jbaehr  I have agreed that he also takes a look at the helix feature's error handling. Therefore I won't change too much to avoid interferences with his work.

Concerning this PR here, does it work for you as it should and is the code OK in your opinion?

AjinkyaDahale:
> Concerning this PR here, does it work for you as it should and is the code OK in your opinion?

The code appears OK. However, with the example you provided, even with an integer number of turns the end faces of the helix do not appear to be parallel (more prominent when setting turns to 4). That is counterintuitive, but if that behavior is correct you are good to go.

donovaly:
> Note that in the height-turns-angle configuration we already can choose a negative angle and get the same effect as negative growth.

Hmm, I had a look at this again and cannot get this to work. Only with the growth mode, I can create a helix in a plane and according to our Wiki that feature was the reason the growth mode was added.

Therefore I am even more convinced that we must allow negative growths.

donovaly:
> The code appears OK. However, with the example you provided, even with an integer number of turns the end faces of the helix do not appear to be parallel (more prominent when setting turns to 4). That is counterintuitive, but if that behavior is correct you are good to go.

Thanks for this hint. This is indeed interesting. However, this has nothing to do with my PR but with the way the helix is calculated:
When the end is close to the center, the end is no longer on the same plane as the start. So for example even with 6 turns, but starting more outside of the center works:
firefox_q7sYWu8kiD

In this example first with 7 turns, the end is no longer on the same plane as the start.

However, as I wrote, there is something in general not correct with the calculation since also with positive growth it tells me for non-intersecting helices, that it would intersect:
FreeCAD_WycpIeJs9Y

donovaly:
Indeed, the bug always occurs if the helix is close to the center. This is independent on the growth:
FreeCAD_XzGjhlNGd5


So there is no reason not to allow negative growth -> merging the PR now.

I will have a look why the helix calculation fails in growth mode.

0penBrain:
@donovaly Saw you requested a review from me but not so available ATM. Sorry for that, nice you merged anyway. ;)

AjinkyaDahale:
> > Note that in the height-turns-angle configuration we already can choose a negative angle and get the same effect as negative growth.
> 
> Hmm, I had a look at this again and cannot get this to work. Only with the growth mode, I can create a helix in a plane and according to our Wiki that feature was the reason the growth mode was added.

I think you got it wrong. This is what I was talking about: 
Screenshot at 2021-11-29 13-27-40

Of course, this would only have the same effect if we wanted non-zero height (in your example it was 1mm). As height -> 0 I imagine this angle -> -90, but of course that is not, and should not be, supported.

donovaly:
But as I wrote the Growth mode is to allow helices within one plane. There is no other solution to get a helix in a plane but this ia what I referred to.

AjinkyaDahale:
> But as I wrote the Growth mode is to allow helices within one plane. There is no other solution to get a helix in a plane but this ia what I referred to.

Yep. I get that. When I said "same effect", I meant 1) it can replicate the negative growth case with the given non-zero height; and 2) can also face the same intersection issues. So if we're allowing negative angles, there's no reason to not have negative growth.

PR #5217, opened 2021-12-01
[PATH] rotation generator and tests

sliptonic:

---

### DIFF
--- a/src/Mod/Path/Generators/rotation_generator.py
+++ b/src/Mod/Path/Generators/rotation_generator.py
@@ -1,0 +1,205 @@
+# -*- coding: utf-8 -*-
+# ***************************************************************************
+# *   Copyright (c) 2021 sliptonic <shopinthewoods@gmail.com>               *
+# *                                                                         *
+# *   This program is free software; you can redistribute it and/or modify  *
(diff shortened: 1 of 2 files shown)


src/Mod/Path/Generators/rotation_generator.py
```diff
@@ -0,0 +1,205 @@
...
+
+    if compound is False, axis moves will be broken out to individual commands
+
+    The normalVector input from a typical face (f) can be obtained like this:
+
+        u, v = f.ParameterRange[:2]
+        n = f.normalAt(u,v)
+        plm = obj.getGlobalPlacement()
+        rot = plm.Rotation
+        normalVector = rot.multVec(n
```
Russ4262, line 152: If the expectation is that this method should be used to obtain the normal vector for a target face, and it is assumed this method might be used more than once in PathScripts to work with rotation, then I suggest making this set of instructions into a simple @classmethod here as a helper/utility method available for usage.

sliptonic: That's a good idea.  I hadn't thought of a class method but was considering some other options.  I don't know if it belongs in the generator module or a PathUtils or somewhere else.  I held off because I found the described solution didn't work in some cases and I didn't have time to track it down.

PR #5221, opened 2021-12-03
[Gui] QuantitySpinBox always append 'mm' after expression unitless value

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=10&t=64050

I set the PR as Draft ATM because a better solution would be that the expression result is evaluated immediately, without waiting the engine to do it later. But I didn't found a solution yet. 

If one confirms that the better solution isn't possible, or that this one is a good temporary solution though, I'll set it OK to be merged.

### DIFF
--- a/src/Gui/QuantitySpinBox.cpp
+++ b/src/Gui/QuantitySpinBox.cpp
@@ -64,13 +64,14 @@ namespace Gui {
 class QuantitySpinBoxPrivate
 {
 public:
-    QuantitySpinBoxPrivate() :
+    QuantitySpinBoxPrivate(QuantitySpinBox *q) :
       validInput(true),
       pendingEmit(false),
       unitValue(0),
       maximum(DOUBLE_MAX),
       minimum(-DOUBLE_MAX),
-      singleStep(1.0)
+      singleStep(1.0),
+      q_ptr(q)
     {
     }
     ~QuantitySpinBoxPrivate()
@@ -89,6 +90,8 @@ public:
 
     bool validate(QString& input, Base::Quantity& result) const
     {
+        Q_Q(const QuantitySpinBox);
+
         // Do not accept empty strings because the parser will consider
         // " unit" as "1 unit" which is not the desired behaviour (see #0004104)
         if (input.isEmpty())
@@ -108,7 +111,7 @@ public:
         else if (state == QValidator::Intermediate) {
             tmp = tmp.trimmed();
             tmp += QLatin1Char(' ');
-            tmp += unitStr;
+            tmp += q->isBound()?QStringLiteral("mm"):unitStr; // In expressions, default unit is always "mm"
             Base::Quantity res2 = validateAndInterpret(tmp, pos, state);
             res2.setFormat(quantity.getFormat());
             if (state == QValidator::Acceptable) {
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5221">pipeline status</a> for feature branch [PR_5221](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5221). Pipeline [423701042](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/423701042) was triggered at [5724b9b](https://github.com/FreeCAD/FreeCAD/pull/5221/commits/5724b9b9f95141a489486fb1cd2638745dfd5d6f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer (commented):
I understand the motivation to make things a bit clearer for the mentioned use case but in general this PR makes things even worse.

The assumption that the unit of a dimensionless expression is mm is simply wrong and leads to very strange situations. If a unitless value is passed it depends on the context and the spin box cannot make any assumptions about it. 

Example:
Instead of a rectangle draw two non-parallel lines and define an angle between them. Let's say the calculated angle is set to 40.564 deg. Double-click the constraint and in the dialog click the expression button and enter 45, confirm with OK.
The spin box correctly shows 45.000 but as soon as the widget loses focus the value jumps back to 40.564 deg. This is because the spin box expects a quantity in deg and not mm and thus switches back to the last correct value.

Now if you click the OK button of the dialog the constraint accepts the 45 deg. This is because it doesn't use the returned value of the dialog but that of the expression.

0penBrain:
@wwmayer Not sure it's actually worse but for sure not far better. :) I'm also not really happy to have to introduce a Q pointer.
As I said in the PR original comment, the best solution would be that when the expression dialog is accepted, it is immediately evaluated and the quantity spinbox is updated. I had not time till now to crawl into the code to know if it's possible, but if you can achieve this I'd be more than happy to kill this PR. ;)

wwmayer:
6aec424aa5

0penBrain:
@wwmayer looks excellent. Could you please kill this PR if it makes no more sense after your commit?

wwmayer:
Obsolete PR.

adrianinsaval:
@wwmayer the issue has not been solved here I get 13" displayed when it should be 13mm
Screenshot_20220107_200652
[wrongDisplayedUnit.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/7831758/wrongDisplayedUnit.FCStd.zip)

> OS: Manjaro Linux (KDE/plasma)
> Version: 0.20.26960 (Git)
> Build type: Release
> Branch: master
> Hash: d76f3b3c0cf231ba2d36cfdc24c2cd73d3f6bc86
> Python version: 3.10.1
> Qt version: 5.15.2
> Coin version: 4.0.1
> OCC version: 7.5.3
> Locale: English/United Kingdom (en_GB

wwmayer:
@adrianinsaval 
> the issue has not been solved here I get 13" displayed when it should be 13mm (or 1.024 ")

How was the behaviour before? All what I did is to override setExpression() and calling the same code as in focusOutEvent().

adrianinsaval:
before switching focus the field is shown without unit but after changing focus it incorrectly displayed 13", the reported issue was the incorrect unit, now the unit is shown immediately but is still wrong

adrianinsaval:
@wwmayer another instance of this bug https://forum.freecadweb.org/viewtopic.php?f=3&t=65230 validateInput should check what unit is appropriate to use or convert to the configured display unit

wwmayer:
The problem is that QuantitySpinBox is too permissive. https://github.com/FreeCAD/FreeCAD/blob/master/src/Gui/QuantitySpinBox.cpp#L108 

In case the input string is not valid but in a supposed intermediate state the unit of the last valid input is appended and then re-interpreted. This way the modified input will be accepted.

adrianinsaval:
Maybe there it should check if it has an expression providing the value and then get the correct default units for that? Or validateAndInterpret should append the unit and set it as acceptable when an expression is there? I'm not sure how these things work really.

0penBrain:
Would that be an option that QuantitySpinBox is not that permissive in case it's bound to an expression?

wwmayer:
> Would that be an option that QuantitySpinBox is not that permissive in case it's bound to an expression?

Exactly. See 9d37d27229

0penBrain:
@wwmayer Back to a Q-pointer :rofl: Sounds good, thanks for this

PR #5224, opened 2021-12-05
PD: Enhance the helix property docs

jbaehr:
Up to now the property descriptions (e.g. shown as tool tip in the property grid) just repeat the property's name, which was not very helpful. With this PR most strings are based on descriptions from the [wiki](https://wiki.freecadweb.org/PartDesign_AdditiveHelix). Some have been extended where I saw need (once accepted, I can adapt the wiki, too). Note that I'm not a native speaker, though.

I also plan to make those property descriptions [translatable](https://tracker.freecadweb.org/view.php?id=0002524) and use them as tooltip in the helix' task panel. As you seem to prefer several smaller PRs I did not include this here. It may also requires some more discussion.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -77,22 +77,42 @@ Helix::Helix()
 {
     addSubType = FeatureAddSub::Additive;
 
-    ADD_PROPERTY_TYPE(Base, (Base::Vector3d(0.0, 0.0, 0.0)), "Helix", App::Prop_ReadOnly, "Base");
-    ADD_PROPERTY_TYPE(Axis, (Base::Vector3d(0.0, 1.0, 0.0)), "Helix", App::Prop_ReadOnly, "Axis");
-    ADD_PROPERTY_TYPE(Pitch, (10.), "Helix", App::Prop_None, "Pitch");
-    ADD_PROPERTY_TYPE(Height, (30.0), "Helix", App::Prop_None, "Height");
-    ADD_PROPERTY_TYPE(Turns, (3.0), "Helix", App::Prop_None, "Turns");
+    const char* group = "Helix";
+    ADD_PROPERTY_TYPE(Base, (Base::Vector3d(0.0, 0.0, 0.0)), group, App::Prop_ReadOnly,
+        "The center point of the helix' start.");
+    ADD_PROPERTY_TYPE(Axis, (Base::Vector3d(0.0, 1.0, 0.0)), group, App::Prop_ReadOnly,
+        "The helix' direction.");
+    ADD_PROPERTY_TYPE(Pitch, (10.), group, App::Prop_None,
+        "The axial distance between two turns.");
+    ADD_PROPERTY_TYPE(Height, (30.0), group, App::Prop_None,
+        "The height of the helix' path, not accounting for the extent of the profile.");
+    ADD_PROPERTY_TYPE(Turns, (3.0), group, App::Prop_None,
+        "The number of turns in the helix.");
     Turns.setConstraints(&floatTurns);
-    ADD_PROPERTY_TYPE(LeftHanded, (long(0)), "Helix", App::Prop_None, "LeftHanded");
-    ADD_PROPERTY_TYPE(Reversed, (long(0)), "Helix", App::Prop_None, "Reversed");
-    ADD_PROPERTY_TYPE(Angle, (0.0), "Helix", App::Prop_None, "Angle");
-    ADD_PROPERTY_TYPE(Growth, (0.0), "Helix", App::Prop_None, "Growth");
+    ADD_PROPERTY_TYPE(LeftHanded, (long(0)), group, App::Prop_None,
+        "Sets the turning direction to left handed,\n"
+        "i.e. counter-clockwise when moving along its axis.");
+    ADD_PROPERTY_TYPE(Reversed, (long(0)), group, App::Prop_None,
+        "Determines whether the helix points in the opposite direction of the axis.");
+    ADD_PROPERTY_TYPE(Angle, (0.0), group, App::Prop_None,
+        "The angle of the cone that forms a hull around the helix.\n"
+        "Non-zero values turn the helix into a conical spiral.\n"
+        "Positive values make the radius grow, nevatige shrink.");
     Angle.setConstraints(&floatAngle);
-    ADD_PROPERTY_TYPE(ReferenceAxis, (0), "Helix", App::Prop_None, "Reference axis of revolution");
-    ADD_PROPERTY_TYPE(Mode, (long(0)), "Helix", App::Prop_None, "Helix input mode");
-    ADD_PROPERTY_TYPE(Outside, (long(0)), "Helix", App::Prop_None, "Outside");
-    ADD_PROPERTY_TYPE(HasBeenEdited, (long(0)), "Helix", App::Prop_None, "HasBeenEdited");
+    ADD_PROPERTY_TYPE(Growth, (0.0), group, App::Prop_None,
+        "The growth of the helix' radius per turn.\n"
+        "Non-zero values turn the helix into a conical spiral.");
+    ADD_PROPERTY_TYPE(ReferenceAxis, (0), group, App::Prop_None,
+        "The reference axis of the helix.");
+    ADD_PROPERTY_TYPE(Mode, (long(0)), group, App::Prop_None,
+        "The helix input mode specifies which properties are set by the user.\n"
+        "Dependent properties are then calculated.");
     Mode.setEnums(ModeEnums);
+    ADD_PROPERTY_TYPE(Outside, (long(0)), group, App::Prop_None,
+        "If set, the result will be the intersection of the profile and the preexisting body.");
+    ADD_PROPERTY_TYPE(HasBeenEdited, (long(0)), group, App::Prop_None,
+        "If false, the tool will propose an initial value for the pitch based on the profile bounding box,\n"
+        "so that self intersection is avoided.");
 }
 
 short Helix::mustExecute() const


src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
...
+    ADD_PROPERTY_TYPE(Axis, (Base::Vector3d(0.0, 1.0, 0.0)), group, App::Prop_ReadOnly, "The helix' direction.");
+    ADD_PROPERTY_TYPE(Pitch, (10.), group, App::Prop_None, "The axial distance between two turns.");
+    ADD_PROPERTY_TYPE(Height, (30.0), group, App::Prop_None, "The total length of the helix (not accounting for the extent of the profile)");
+    ADD_PROPERTY_TYPE(Turns, (3.0), group, App::Prop_None, "The number of turns in the helix (does not need to be a whole number).");
     Turns.setConstraints(&floatTurns);
-    ADD_PROPERTY_TYPE(LeftHanded, (long(0)), "Helix", App::Prop_None, "LeftHanded");
-    ADD_PROPERTY_TYPE(Reversed, (long(0)), "Helix", App::Prop_None, "Reversed");
-    ADD_PROPERTY_TYPE(Angle, (0.0), "Helix", App::Prop_None, "Angle");
-    ADD_PROPERTY_TYPE(Growth, (0.0), "Helix", App::Prop_None, "Growth");
+    ADD_PROPERTY_TYPE(LeftHanded, (long(0)), group, App::Prop_None, "Determins whether the turning direction of helix follows the \"left-hand rule\", i.e. counter-clockwise when moving along its axis.");
```
donovaly, line 87: Determin_**e**_s

However I don't find this info helpful since one need to know that the rule is. What about just

"Sets turning direction to left-handed"

Those who deal with threads and helices know the rule so we can keep the info short.

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
 {
     addSubType = FeatureAddSub::Additive;
 
-    ADD_PROPERTY_TYPE(Base, (Base::Vector3d(0.0, 0.0, 0.0)), "Helix", App::Prop_ReadOnly, "Base");
-    ADD_PROPERTY_TYPE(Axis, (Base::Vector3d(0.0, 1.0, 0.0)), "Helix", App::Prop_ReadOnly, "Axis");
-    ADD_PROPERTY_TYPE(Pitch, (10.), "Helix", App::Prop_None, "Pitch");
-    ADD_PROPERTY_TYPE(Height, (30.0), "Helix", App::Prop_None, "Height");
-    ADD_PROPERTY_TYPE(Turns, (3.0), "Helix", App::Prop_None, "Turns");
+    const char* group = "Helix";
+    ADD_PROPERTY_TYPE(Base, (Base::Vector3d(0.0, 0.0, 0.0)), group, App::Prop_ReadOnly, "The center point of the helix' start.");
```
donovaly, line 81: Just a question; Why can't the Base be changed?

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
...
     Turns.setConstraints(&floatTurns);
-    ADD_PROPERTY_TYPE(LeftHanded, (long(0)), "Helix", App::Prop_None, "LeftHanded");
-    ADD_PROPERTY_TYPE(Reversed, (long(0)), "Helix", App::Prop_None, "Reversed");
-    ADD_PROPERTY_TYPE(Angle, (0.0), "Helix", App::Prop_None, "Angle");
-    ADD_PROPERTY_TYPE(Growth, (0.0), "Helix", App::Prop_None, "Growth");
+    ADD_PROPERTY_TYPE(LeftHanded, (long(0)), group, App::Prop_None, "Determins whether the turning direction of helix follows the \"left-hand rule\", i.e. counter-clockwise when moving along its axis.");
+    ADD_PROPERTY_TYPE(Reversed, (long(0)), group, App::Prop_None, "Determins whether the helix points in the opposite direction of the reference axis.");
+    ADD_PROPERTY_TYPE(Angle, (0.0), group, App::Prop_None, "If non-zero, the profile's path wraps around a cone with this angle, instead of a cylinder. Positive values make the radius grow, nevatige shrink.");
+    ADD_PROPERTY_TYPE(Growth, (0.0), group, App::Prop_None, "The growth of the helix' radius per turn. Non-zero values turn the helix into a conical spiral.");
```
donovaly, line 90: I think just

"The radial growth of the helix' radius per turn."

is sufficient

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
...
+    ADD_PROPERTY_TYPE(Pitch, (10.), group, App::Prop_None, "The axial distance between two turns.");
+    ADD_PROPERTY_TYPE(Height, (30.0), group, App::Prop_None, "The total length of the helix (not accounting for the extent of the profile)");
+    ADD_PROPERTY_TYPE(Turns, (3.0), group, App::Prop_None, "The number of turns in the helix (does not need to be a whole number).");
     Turns.setConstraints(&floatTurns);
-    ADD_PROPERTY_TYPE(LeftHanded, (long(0)), "Helix", App::Prop_None, "LeftHanded");
-    ADD_PROPERTY_TYPE(Reversed, (long(0)), "Helix", App::Prop_None, "Reversed");
-    ADD_PROPERTY_TYPE(Angle, (0.0), "Helix", App::Prop_None, "Angle");
-    ADD_PROPERTY_TYPE(Growth, (0.0), "Helix", App::Prop_None, "Growth");
+    ADD_PROPERTY_TYPE(LeftHanded, (long(0)), group, App::Prop_None, "Determins whether the turning direction of helix follows the \"left-hand rule\", i.e. counter-clockwise when moving along its axis.");
+    ADD_PROPERTY_TYPE(Reversed, (long(0)), group, App::Prop_None, "Determins whether the helix points in the opposite direction of the reference axis.");
```
donovaly, line 88: "Flips the direction of the helix axis"

Because we don't have the term "reference axis" yet and I think the shorter the better - when it provides the same info.

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
...
+    ADD_PROPERTY_TYPE(Angle, (0.0), group, App::Prop_None, "If non-zero, the profile's path wraps around a cone with this angle, instead of a cylinder. Positive values make the radius grow, nevatige shrink.");
```
donovaly, line 89: "If non-zero, " can be omitted. It is clear that it is the angle, and if zero, there is no angle at all.

src/Mod/PartDesign/Gui/TaskHelixParameters.ui
```diff
@@ -230,7 +230,7 @@
      <item>
       <widget class="QLabel" name="labelGrowth">
        <property name="text">
-        <string>Growth:</string>
+        <string>Radius growth:</string>
```
donovaly, line 233: I would say "Radial growth"

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
...
+    ADD_PROPERTY_TYPE(Growth, (0.0), group, App::Prop_None, "The growth of the helix' radius per turn. Non-zero values turn the helix into a conical spiral.");
     Angle.setConstraints(&floatAngle);
-    ADD_PROPERTY_TYPE(ReferenceAxis, (0), "Helix", App::Prop_None, "Reference axis of revolution");
-    ADD_PROPERTY_TYPE(Mode, (long(0)), "Helix", App::Prop_None, "Helix input mode");
-    ADD_PROPERTY_TYPE(Outside, (long(0)), "Helix", App::Prop_None, "Outside");
-    ADD_PROPERTY_TYPE(HasBeenEdited, (long(0)), "Helix", App::Prop_None, "HasBeenEdited");
+    ADD_PROPERTY_TYPE(ReferenceAxis, (0), group, App::Prop_None, "The reference axis of the helix.");
```
donovaly, line 92: "The axis the helix winds around \r this is the same as the 'Axis'."

Because the Axis shows already the vector component values of it.

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
...
+    ADD_PROPERTY_TYPE(Mode, (long(0)), group, App::Prop_None, "The helix input mode.");
```
donovaly, line 93: I would say it is more the definition mode than the input, but I am also no native speaker.

donovaly (changes requested):
Thanks for the PR.

I would describe it in a different way, see my comments.

Sure, one could discuss much about the wording. I think our rule of thumb (looking at other FC features) is
"provide necessary info in the shortest way"

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
...
     addSubType = FeatureAddSub::Additive;
 
-    ADD_PROPERTY_TYPE(Base, (Base::Vector3d(0.0, 0.0, 0.0)), "Helix", App::Prop_ReadOnly, "Base");
-    ADD_PROPERTY_TYPE(Axis, (Base::Vector3d(0.0, 1.0, 0.0)), "Helix", App::Prop_ReadOnly, "Axis");
-    ADD_PROPERTY_TYPE(Pitch, (10.), "Helix", App::Prop_None, "Pitch");
-    ADD_PROPERTY_TYPE(Height, (30.0), "Helix", App::Prop_None, "Height");
-    ADD_PROPERTY_TYPE(Turns, (3.0), "Helix", App::Prop_None, "Turns");
+    const char* group = "Helix";
+    ADD_PROPERTY_TYPE(Base, (Base::Vector3d(0.0, 0.0, 0.0)), group, App::Prop_ReadOnly, "The center point of the helix' start.");
+    ADD_PROPERTY_TYPE(Axis, (Base::Vector3d(0.0, 1.0, 0.0)), group, App::Prop_ReadOnly, "The helix' direction.");
```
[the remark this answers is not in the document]
jbaehr: No, unfortunately not. What you mean is the *property* `ReferenceAxis`, which is called *in the UI of the task panel* just "Axis". The *property* `Axis` is only the direction. It is derived from `ReferenceAxis` (just like `Base`) and because of this read-only.

I know, it's a complete mess, but unfortunately we cannot change the property names without breaking existing models (at least AFAIK).

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -77,21 +77,22 @@ Helix::Helix()
...
+    ADD_PROPERTY_TYPE(Pitch, (10.), group, App::Prop_None, "The axial distance between two turns.");
+    ADD_PROPERTY_TYPE(Height, (30.0), group, App::Prop_None, "The total length of the helix (not accounting for the extent of the profile)");
```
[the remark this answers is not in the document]
jbaehr: You're right. I'Ve copied it directly from the wiki but this one clearly needs to be changed. Will do.

Regarding the overlong lines, you're right, too. I'm currently working on a patch that makes the property docs translatable, in order to use them as tooltips in the UI, too. In there I've already changed it.
Perhaps it's better to put it all together in this PR...

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: In my view, the tool tips are especially for those who *don't* know; users who do, don't need a tool tip at all. So in that sense, a tool tip for "Foo" which just tells "sets the Foo" is barely helpful.

What I tried to do with this text here is
a) to let the property name make sense: It's named `LeftHanded` *in reference to* the left-hand rule.
b) to explain what is does: It make the helix turn counter-clockwise when moving along its axis.

(`IsLeftHanded` would be a better property name, but it's too late to change this)

The typo gets fixed, of course. Thanks for pointing out.

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: See response to your first comment regarding `Axis`: `Base` is derived from the (changeable) `ReferenceAxis` (which again is called "Axis" in the task panel).

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: > "The radial growth of the helix' radius per turn."

The radial growth of the radius? I guess you mean I should just omit the second sentence... ;-)
I'm fine with that, although I don't see why it hurts. In the wiki we should add the term "conical spiral" though, to make it "googleable".

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: We *do* have a `ReferenceAxis`. Together with `Axis` and `Base` it is the very same as in the Revolution or Groove features.

Revolution has a property description of "Reverse extrusion direction".

src/Mod/PartDesign/Gui/TaskHelixParameters.ui
jbaehr: Fine with me.

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: The Revolution feature has "The reference axis of revolution" here. Also, as `Axis` is derived from `ReferenceAxis` I would not make the references in the docs point the other way round. Especially as I plan to use the very same property description as tooltip in the task panel, where the `Axis` property is not known at all: What is called "Axis" in the task panel is actually the `ReferenceAxis` property.

And it is *not* "the same as the 'Axis'": The `Axis` holds only the directional part of `RefernceAxis`.

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: Indeed, that not very helpful. In fact I only changed the previous doc string to become a complete sentences.

Better would be: "The helix input mode specifies which properties are set by the user. Dependent properties are then calculated."

src/Mod/PartDesign/App/FeatureHelix.cpp
donovaly: I meant
"The radial growth of the helix' per turn.""

src/Mod/PartDesign/App/FeatureHelix.cpp
donovaly: You explain too left-hand with left-hand.

"counter-clockwise when moving along its axis."

is only helpful for those who knows what left-had means. To make it clear one must mention the axis direction and from what direction one looks at the system. It is in my opinion impossible to explain left-handed in a tooltip and as i wrote, those who deal with helices know it. If not the "turning direction" makes clear what it is about and since there re only 2 possibilities, users can just check them out.

src/Mod/PartDesign/App/FeatureHelix.cpp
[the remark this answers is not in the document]
donovaly: Aha. So then the tooltip for "Reference axis" should reflect this.

src/Mod/PartDesign/App/FeatureHelix.cpp
donovaly: Thanks. So making the base changeable would improve the feature a lot for the normal direction. Then one could take a normal and move it around. This saves time in comparison of creating a special axis in normal direction at the base one wants.

src/Mod/PartDesign/App/FeatureHelix.cpp
donovaly: Yes, but
"Reverses the direction of the reference axis"
is much shorter than
"Determines whether the helix points in the opposite direction of the reference axis."

donovaly:
I added now the missing tooltips for the UI file:
https://github.com/FreeCAD/FreeCAD/commit/1af3755a175fa11d9d04ee51582d14848d8b45d0

I must admit, finding the wording is tricky. Therefore please report back if you think it could be improved.

src/Mod/PartDesign/Gui/TaskHelixParameters.ui
donovaly: Resolved by commit 1af3755

jbaehr:
> I added now the missing tooltips for the UI file: [1af3755](https://github.com/FreeCAD/FreeCAD/commit/1af3755a175fa11d9d04ee51582d14848d8b45d0)
> 
> I must admit, finding the wording is tricky. Therefore please report back if you think it could be improved.

I'm a bit disappointed that you already pushed your variant to master even though we're in the middle of a discussion. If we, both not native speakers, cannot come to a consensus, we would still ask in the "UX/UI Design" forum for assistance...
I also stated several times that I plan to set the tooltips in the GUI based on the property descriptions. Now we have to maintain the information in two places again, in TaskHelixParameters.ui and FeatureHelix.cpp.

Regarding your tooltips from https://github.com/FreeCAD/FreeCAD/commit/1af3755a175fa11d9d04ee51582d14848d8b45d0:
There are some variants which I indeed like more than mine. For example the one for the cone angle. In general I do not understand your motivation to make the properties description/tooltips as short as possible, up to the point where it's not even a complete sentence any more. You said, that those how already know do not need more information. True. But in my view the descriptions/tooltips should help explicitly those who do not know.

donovaly:
> I'm a bit disappointed that you already pushed your variant to master even though we're in the middle of a discussion.

This surprizes me. We agreed in your fix to the UI file. So to make the next step, I put this in and since the UI had no tooltips at all, I added some. In your commit there was only one change for the UI file.

> I also stated several times that I plan to set the tooltips in the GUI based on the property descriptions. 

I missed this, sorry. However, the properties contain more settings than the UI and have to be different. For example there Axis and Base represent each a part of the Reference Axis.
So the idea is to focus with this PR on the properties so that you can make a new commit with focus on this and with what we discussed.

donovaly:
> I do not understand your motivation to make the properties description/tooltips as short as possible

As I wrote, as short as possible but as long as necessary. We don't have strict rules, I learned meanwhile that the tooltips are intended as "tips" so they should give you the idea. For details we have our Wiki and to access the Wiki we have https://wiki.freecadweb.org/Std_WhatsThis

jbaehr:
> We agreed in your fix to the UI file.

In my eyes the thing we agreed on was the "Radial growth" instead of "Radius growth".

> So to make the next step, I put this in and since the UI had no tooltips at all, I added some.

Ok, makes sense under the light that you missed my intend to use the descriptions as tooltips as well.

> > I also stated several times that I plan to set the tooltips in the GUI based on the property descriptions.
> 
> I missed this, sorry. However, the properties contain more settings than the UI and have to be different. For example there Axis and Base represent each a part of the Reference Axis. So the idea is to focus with this PR on the properties so that you can make a new commit with focus on this and with what we discussed.

Indeed, let's first finish this one. But not Today ;-)

None the less I've already completed the work on using the property descriptions as tooltips. I think it works quite well. You're right that the UI does not expose all properties. However, for those that are exposed, I see no reason to use something different than the property description as tool tip. In fact, the only place a user sees the property descriptions are again as tooltip in the property grid. I think the same property of a feature (i.e. the height) should have *the same* tool tip, independently wether the current editor for it is the task panel of the property grid. 

As the code is based on this PR, I have not filed a new one for this change, but here it is, if you want to take a look: https://github.com/jbaehr/FreeCAD/commits/feature/introduce-helix-taskpanel-tooltips

donovaly:
> In my eyes the thing we agreed on was the "Radial growth" instead of "Radius growth".

Yes and this is now in.

> I see no reason to use something different than the property description as tool tip.

Sure.
Base, Axis, and Reference axis and Reversed are different.
For Growth, Height and Pitch we agreed or I used your proposal.
For the Angle you like my proposal.

So in fact only the Left handed property is to be discussed, right?

The label for Growth has to be renamed in the Properties to "Radial Growth".

Can you please rebase and make a new commit where you address the properties?

src/Mod/PartDesign/App/FeatureHelix.cpp
wwmayer: I don't think that "If non-zero" should be dropped because otherwise the rest "instead of a cylinder" doesn't make things clear.

> and if zero, there is no angle at all.

Why is an angle of 0 degree not an angle?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5224">pipeline status</a> for feature branch [PR_5224](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5224). Pipeline [423702770](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/423702770) was triggered at [d189f34](https://github.com/FreeCAD/FreeCAD/pull/5224/commits/d189f34a8ce89ea7a14a452ca9cab1464fc24656). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureHelix.cpp
donovaly: > Why is an angle of 0 degree not an angle?

Like a distance of 0 is actually no distance ;-)

However, I think we found already a suitable wording. The one I use in the dialog:
"Angle of the cone that forms a hull around the helix."

Then maybe a second sentence that a positive angle make the cone wider along the helix axis. So maybe:

"Angle of the cone that forms a hull around the helix. A positive angle widens the cone along the axis."

src/Mod/PartDesign/App/FeatureHelix.cpp
jbaehr: I think this particular property should not be defined here in the first place; it seems to shadow the `Reversed` property from the base class. I'll take a look after this PR.

jbaehr:
> Can you please rebase and make a new commit where you address the properties?

Done.

I also shortened a bit and finally opted for your first part for the `LeftHanded` property. I extended what you've set as tool tips in the GUI to make complete sentences (e.g. "Height of helix" -> "The height of the helix.") and added bits of information that I consider relevant (e.g. what left-handed means or that the height does not take the profile into account)

Overlong lines are broken, in the source as well as in the property descriptions themselves (i.e. when they get displayed as tool tips)

donovaly:
Merci.

PR #5226, opened 2021-12-05
[PD] fix calculation of possible Helix intersections

donovaly:
the current routine fails for the growth mode.

This PR fixes this by handling this case separately from the other modes.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -473,8 +473,29 @@ double Helix::safePitch()
 {
     Base::Vector3d axisVec = Axis.getValue();
     Base::Vector3d startVec = axisVec.Cross(getProfileNormal()); // pointing towards the helix start point
-    if (startVec.IsNull())
-        return Precision::Confusion(); // if the axis orthogonal to the profile, any pitch > 0 is safe
+    HelixMode mode = static_cast<HelixMode>(Mode.getValue());
+    double growthValue = Growth.getValue();
+    double turnsValue = Turns.getValue();
+
+    // handle case if axis is orthogonal to profile
+    // since startVec.IsNull() fails sometimes, for e.g. (0.0, 0.0, -0.0)
+    // we take the precision into account
+    if (startVec.Length() < Precision::Confusion()) {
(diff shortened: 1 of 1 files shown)


donovaly:
@jbaehr can you please give this a try? This works my all my test cases now, but I might have missed something.

donovaly:
@AjinkyaDahale can you please also give this a try?

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -473,8 +473,29 @@ double Helix::safePitch()
...
+        }
+        // if growth is not zero, there will in many cases be intersections
+        // when the turn is >= 1, thus return an 'infinite' pitch
+        // Note: The resulting helix body is in this case often garbage since
+        // the OCC algorithm to create the helix fails.
+        // Nevertheless, the result is a valid body so it should be valuable for users
+        // to get this correct warning anyway.
+        else {
+            if (abs(turnsValue) >= 1.0 && abs(growthValue) > 0.0)
+                return INT_MAX;
```
wwmayer, line 496: Wouldn't Precision::Infinite() be more appropriate? This also makes the intention clearer that "inf" should be returned.

AjinkyaDahale: Additionally a function returning `double` shouldn't be returning an integer anyway (please correct me if I'm missing something here).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5226">pipeline status</a> for feature branch [PR_5226](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5226). Pipeline [423702789](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/423702789) was triggered at [6f252fc](https://github.com/FreeCAD/FreeCAD/pull/5226/commits/6f252fcd962eac401df2c1641cace78b3a1fe764). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks for the revision. I changed it accordingly.

PR #5229, opened 2021-12-06
[PATH]  Refactoring drilling op to use generators

sliptonic:

---


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5229">pipeline status</a> for feature branch [PR_5229](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5229). Pipeline [423704161](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/423704161) was triggered at [3ddd575](https://github.com/FreeCAD/FreeCAD/pull/5229/commits/3ddd5751120c9a27f6638e251ea613ed5a5b39b9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Path/PathFeedRate.py
```diff
@@ -0,0 +1,95 @@
...
+                ToolController.HorizRapid.Value
+                if command in rapidcommands
+                else ToolController.HorizFeed.Value
+            )
+
+        params = command.Parameters
+        params["F"] = rate
+        command.Parameters = params
+
+        machine.addCommand(command)
```
iromero91, line 93: I think this would be a bit simpler and easier to understand like this. It does not use an auxiliary function nor builds too many extra objects.

```python
for command in commandlist:
    start_point = machine.getPosition()
    machine.addCommand(command)
    end_point = machine.getPosition()
    move_vec = end_point-start_point
    move_vec.z = 0 # Project onto the XY plane
    is_vertical = (move_vec.Length < PathGeom.Tolerance)
    if command.Name in feedcommands:
        if is_vertical:
            feed = ToolController.VertFeed.Value
        else:
            feed = ToolController.HorizFeed.Value
        command.Parameters["F"] = feed
    elif command.Name in rapidcommands:
        if is_vertical:
            feed = ToolController.VertRapid.Value
        else:
            feed = ToolController.HorizRapid.Value
        command.Parameters["F"] = feed
```

Also it makes sure to pass along all commands to the machine, this for example would handle correctly cases where there is a mixture of relative and absolute positioning for example (if/when the machine implements that). The vertical checking part is simple vector math, get movement vector, project to XY plane, check if there is XY motion.

A potential wrinkle is the semantics definition for the vertical feed. Some users may expect the vertical feed to be used as a limit on the Z axis speed, regardless of the angle of motion (say, a ramp). In that case we should be looking at the RATIO of motion in XY vs Z, if that ratio is is less than the ratio between horizontal and vertical feed in the TC the move should be considered vertical or not. so Z speed can never exceed the vertical feed value, regardless of how it ramps into it.

src/Mod/Path/PathFeedRate.py
```diff
@@ -0,0 +1,95 @@
...
+    machine = PathMachineState.MachineState()
+
+    for command in commandlist:
+        if command.Name not in feedcommands + rapidcommands:
+            continue
+
+        if _isVertical(machine.getPosition(), command):
+            rate = (
+                ToolController.VertRapid.Value
+                if command in rapidcommands
```
iromero91: Bug, it should be checking `command.Name`!

iromero91 (commented):
Test detailed code review. Let me know if this kind of feedback is useful or not. I only checked PathFeedRate so far.

src/Mod/Path/PathFeedRate.py
```diff
@@ -0,0 +1,95 @@
...
+            if "Z" in command.Parameters.keys()
+            else currentposition.z
+        )
+        endpoint = FreeCAD.Vector(x, y, z)
+        if currentposition == endpoint:
+            return True
+        return PathGeom.isVertical(Part.makeLine(currentposition, endpoint))
+
+    feedcommands = ["G01", "G1", "G2", "G3", "G02", "G03", "G81", "G82", "G83"]
+    rapidcommands = ["G0", "G00"]
```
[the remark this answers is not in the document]
sliptonic: agreed.  will change.

src/Mod/Path/PathFeedRate.py
```diff
@@ -0,0 +1,95 @@
...
+"""
+
+if True:
+    PathLog.setLevel(PathLog.Level.DEBUG, PathLog.thisModule())
+    PathLog.trackModule(PathLog.thisModule())
+else:
+    PathLog.setLevel(PathLog.Level.INFO, PathLog.thisModule())
+
+
+def setFeedRate(commandlist, ToolController):
```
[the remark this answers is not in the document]
sliptonic: Yep.  Thanks.

src/Mod/Path/PathFeedRate.py
sliptonic: I like it.  To clarify, the 'machine' here isn't the physical machine or even an internal representation of it.  It's a helper concept for knowing the current machine state.  Gcode is often ambiguous so given the last command, you have to read backwards, potentially to the beginning to know exactly where an axis value is.  For example given the command "G0 X10 Y5"   what is your current Z position?  Is the coolant turned on?  To know, you have to read backwards.  This object just keeps track of those kinds of things so you can query it more easily.

I don't think we should ever implement relative positioning internally either in this machinestate helper or the in the general Path Command structure.  Absolute positioning can be applied by the postprocessor if that is the output the user wants.

src/Mod/Path/PathFeedRate.py
sliptonic: good catch

src/Mod/Path/PathFeedRate.py
iromero91: Yeah, that is what I understood too. it is the gcode state machine. I still think it's a good idea to just feed the command to the machine and then ask the machine how it moved. That moves the concern of interpreting motion commands correctly to the gcode interpreter vs doing that again at this function. I could go as far as perhaps add an additional state variable to the machine class to just get the motion vector from the last command. I can really see this machinestate class being very useful in post processors.

sliptonic: >  I still think it's a good idea to just feed the command to the machine and then ask the machine how it moved.

What machine?

> I could go as far as perhaps add an additional state variable to the machine class to just get the motion vector from the last command. 

A motion vector isn't enough.  That gets us our translation but not rotation elements.

src/Mod/Path/PathScripts/PathDrilling.py
```diff
@@ -23,130 +23,235 @@
...
+            "None",
+            "Drill Tip",
+            "2x Drill Tip",
+        ]  # Canned Cycle Return Level
 
     def circularHoleExecute(self, obj, holes):
-        '''circularHoleExecute(obj, holes) ... generate drill operation for each hole in holes.'''
+        """circularHoleExecute(obj, holes) ... generate drill operation for each hole in holes."""
         PathLog.track()
+        machine = PathMachineState.MachineState()
```
Russ4262, line 142: A new module and class, MachineState.  Appears to be a solution to model the user's machine state based upon the code output from the operation.  Makes sense.  A couple of things...

1.) At the moment, this `machine` object dies when the `circularHoleExecute()` method returns.  This suggests this is something to be employed in future improvements to this operation.  Yes?
 
2.) Would it not be better to move this `MachineState()` usage upstream to the `PathOp` module, and then simply do a simple for loop over the path commands after `opExecute()` returns `results` and the path command list is complete.  As stands with this implementation, there are many `machine.addCommand()` calls.  A simple loop in `PathOp.execute()` would be cleaner and also include any pre and post `opExecute()` commands added in the parent `execute()` method there.  Also, it would apply this machine state idea to all existing ops based on PathOp module.
 
3.) How is the machine state going to be updated if a user inspects the gcode for an operation, such as this drilling op, and modifies the code.  That will then potentially make machine state and the operation's gcode asychronous.  Based on this, this same machine state updating should be performed when a user manually inspects and changes the operation code.  Or, is the machine state only intended to be used during code generation, since it only lives in the `opExecute()` method?

Just some thoughts.

src/Mod/Path/PathScripts/PathDrilling.py
```diff
@@ -23,130 +23,235 @@
...
+            repeat = 1  # technical debt:  Add a repeat property for user control
+
+            try:
+                drillcommands = generator.generate(edge, dwelltime, peckdepth, repeat)
+
+            except ValueError as e:  # any targets that fail the generator are ignored
+                PathLog.info(e)
+                continue
+
+            self.commandlist.extend(drillcommands)
```
Russ4262, line 219: Should there be a loop here to add each command also to the machine state?

src/Mod/Path/PathScripts/PathDrilling.py
```diff
@@ -23,130 +23,235 @@
...
+
+        # Cancel canned drilling cycle
+        self.commandlist.append(Path.Command("G80"))
```
Russ4262, line 222: Does this command need to be added to the machine state?

src/Mod/Path/PathFeedRate.py
sliptonic: Looking at this again, there's a couple problems. 
First, with this solution, the command is added before the feedrate is applied so the machine doesn't correctly reflect the feed rate.  This can be resolved by adding the command again at the end but that seems a bit hack-y.

The command 
```
command.Parameters["F"] = feed
```
won't work.  See here:
```
>>> p = obj.Path.Commands[8]
>>> p
Command G3 [ F:0.8333 I:0 J:-2.5 K:0 X:-1.5 Y:76.038 Z:6 ]
>>> p.toGCode()
'G3 F0.833300 I0.000000 J-2.500000 K0.000000 X-1.500000 Y76.037957 Z6.000000'
>>> p.Parameters['F'] = 200
>>> p
Command G3 [ F:0.8333 I:0 J:-2.5 K:0 X:-1.5 Y:76.038 Z:6 ]
>>> p.Parameters
{'F': 200, 'I': 0.0, 'J': -2.5, 'K': 0.0, 'X': -1.5, 'Y': 76.037957, 'Z': 6.0}
>>> p.toGCode()
'G3 F0.833300 I0.000000 J-2.500000 K0.000000 X-1.500000 Y76.037957 Z6.000000'
>>> params = p.Parameters
>>> params['F'] = 200
>>> p.Parameters = params
>>> p
Command G3 [ F:200 I:0 J:-2.5 K:0 X:-1.5 Y:76.038 Z:6 ]
>>> p.toGCode()
'G3 F200.000000 I0.000000 J-2.500000 K0.000000 X-1.500000 Y76.037957 Z6.000000'
>>> 
```

src/Mod/Path/PathScripts/PathDrilling.py
sliptonic: This is not intended to be a state persistent representation of the machine.  It is more of a scratch pad for the operation to keep track of the machine state when generating commands internally.  That's why it's thrown away at the end.
  
Moving it upstream would make it very easy to get the final machine state that results from the operation, but it makes it useless to the operation itself.  

For example, if the operation is iterating over a bunch of drilling commands and needs to decide after each hole whether or not a rotation is necessary, it can query the machinestate to see what the current position is and then compare that position with the required position.  Without the helper, the operation needs to iterate backwards through the current commands to determine when each axis moved and to what position.

I expect this will also be useful for postprocessors.

If I were a smarter man, I'd incorporate this into the C++ of the Path object.  You could add commands to the Path or query the path to get the current machine position and rotation.

src/Mod/Path/PathScripts/PathDrilling.py
sliptonic: It could but it won't make any difference.  A G80 doesn't cause any motion so the machine position won't change.

src/Mod/Path/PathScripts/PathDrilling.py
sliptonic: Yes, that's a good point.  That's going to take some thought.  The XY words can definitely affect position.  The Z word is the drilling depth but the position where the machine ends up after the move is the old Z prior to the move, not the Z coordinate.

src/Mod/Path/PathFeedRate.py
iromero91: DOH, that's the problem with mutable properties in wrapped objects, seems to always be a 50/50 if they are going to work properly or not. I'm a bit tempted to call this a bug in the Path.Command object, you should be expected to modify a mutable property and have it, you know, change. The workaround using the temp variable should be ok but we should consider fixing the wrapping later. There are a few solutions I would deem more acceptable than what we got now there.

- Remove the property interface and use something like `getParameters()` and `setParameters()` instead. To make it clear that objects are passed by-copy and you cannot mutate them.
- Make `Parameters` return a proxy object so that mutating it actually does the changes expected.
- Make `Parameters` return an immutable object and have a function like `setParameter()` to be able to write to them
- Remove the extra indirection and implement the parameters dictionary on the command class itself via `__getitem__` and `__setitem__`, that way you could do `p["F"] = 200`. Also possible to implement the rest of the basic dictionary interface with `keys()` and `items()`

src/Mod/Path/PathScripts/PathCircularHoleBase.py
```diff
@@ -46,9 +48,11 @@ def translate(context, text, disambig=None):
...
 
 
-PathLog.setLevel(PathLog.Level.INFO, PathLog.thisModule())
-# PathLog.trackModule(PathLog.thisModule())
-
+if False:
+    PathLog.setLevel(PathLog.Level.DEBUG, PathLog.thisModule())
+    PathLog.trackModule(PathLog.thisModule())
+else:
+    PathLog.setLevel(PathLog.Level.INFO, PathLog.thisModule())
```
benj5378, line 55: What is the purpose here? Won't it always result to False?

src/Mod/Path/PathScripts/PathDrilling.py
```diff
@@ -23,130 +23,235 @@
...
 
 
-PathLog.setLevel(PathLog.Level.INFO, PathLog.thisModule())
-# PathLog.trackModule(PathLog.thisModule())
-
+if False:
+    PathLog.setLevel(PathLog.Level.DEBUG, PathLog.thisModule())
+    PathLog.trackModule(PathLog.thisModule())
+else:
+    PathLog.setLevel(PathLog.Level.INFO, PathLog.thisModule())
```
benj5378, line 50: Purpose?

src/Mod/Path/PathScripts/PathUtils.py
```diff
@@ -231,31 +162,42 @@ def horizontalFaceLoop(obj, face, faceList=None):
...
+    """filterArcs(Edge) -used to split arcs that over 180 degrees. Returns list"""
     PathLog.track()
     s = arcEdge
     if isinstance(s.Curve, Part.Circle):
         splitlist = []
         angle = abs(s.LastParameter - s.FirstParameter)
         # overhalfcircle = False
         goodarc = False
-        if (angle > math.pi):
+        if angle > math.pi:
```
benj5378, line 185: What does `angle` contain? The angle in radians or which units? If the angle bigger than 3.14, I don't currently understand

benj5378 (commented):
;)

src/Mod/Path/PathFeedRate.py
```diff
@@ -0,0 +1,99 @@
...
+__title__ = "Feed Rate Helper Utility"
+__author__ = "sliptonic (Brad Collette)"
+__url__ = "https://www.freecadweb.org"
+__doc__ = "Helper for adding Feed Rate to Path Commands"
+
+"""
+TODO:  This needs to be able to handle feedrates for axes other than X,Y,Z
+"""
+
+if True:
```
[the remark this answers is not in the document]
iromero91: These are debugging boilerplate. I am not falling in love with them but there are way worse ways to do it....

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5229">pipeline status</a> for feature branch [PR_5229](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5229). Pipeline [439262132](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/439262132) was triggered at [4a99ce9](https://github.com/FreeCAD/FreeCAD/pull/5229/commits/4a99ce98c7adc1c7946e8066c11bdc55cb57f68c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5229">pipeline status</a> for feature branch [PR_5229](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5229). Pipeline [439823871](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/439823871) was triggered at [2bfcc37](https://github.com/FreeCAD/FreeCAD/pull/5229/commits/2bfcc37dd078c7e9f905e4d04a4898e312f10083). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Path/PathScripts/PathDrilling.py
```diff
@@ -23,130 +23,235 @@
...
+        """circularHoleExecute(obj, holes) ... generate drill operation for each hole in holes."""
         PathLog.track()
+        machine = PathMachineState.MachineState()
 
         self.commandlist.append(Path.Command("(Begin Drilling)"))
 
         # rapid to clearance height
-        self.commandlist.append(Path.Command('G0', {'Z': obj.ClearanceHeight.Value, 'F': self.vertRapid}))
+        command = Path.Command(
+            "G0", {"Z": obj.ClearanceHeight.Value, "F": self.vertRapid}
```
agren, line 148: Do I understand correctly that the F value is not used. And gets replaced in the call to PathFeedRate.setFeedRate on line 228?

src/Mod/Path/PathScripts/drillableLib.py
```diff
@@ -0,0 +1,235 @@
...
+        # Finds the vertical seam edge in a cylinder
+
+        for e in candidate.Edges:
+            if isinstance(e.Curve, Part.Line):  # found the seam
+                return e
+
+    if not candidate.ShapeType == "Face":
+        raise TypeError("expected a Face")
+
+    if not str(candidate.Surface) == "<Cylinder object>":
```
agren, line 58: Is there a reason for converting to string here?
Would `isinstance(candidate.Surface, Part.Cylinder)` work as well? Since isinstance is used on line 52.

src/Mod/Path/PathScripts/PathDrilling.py
sliptonic: Correct.  Good catch.  Thank you.

src/Mod/Path/PathScripts/drillableLib.py
sliptonic: I think you're correct.  I'll change it to use isInstance().   I think it's a bit of copy pasta left.

src/Mod/Path/PathScripts/PathCircularHoleBase.py
sliptonic: debug boilerplate

src/Mod/Path/PathScripts/PathDrilling.py
sliptonic: Updated PathMachineState to handle drillcommands correctly. 
implemented the loop here as suggested

src/Mod/Path/PathFeedRate.py
sliptonic: Added a bug to mantis.  https://tracker.freecadweb.org/view.php?id=4831
Otherwise it's outside the scope of this PR.  I'll use the original implementation for now.  We can revisit if/when the bug is resolved.

src/Mod/Path/PathScripts/PathUtils.py
sliptonic: Good point.  I've added a test case for filterArcs and refactored this function to be more readable.
Caught an error in the process!

PR #5231, opened 2021-12-07
TD: replace usage of raw pointers of Vertex and Face with shared poin…

wwmayer:
…ters.

This fixes issue 4741: Broken File After Using Landmark Dimension in TechDraw

---

### DIFF
--- a/src/Mod/TechDraw/App/Cosmetic.cpp
+++ b/src/Mod/TechDraw/App/Cosmetic.cpp
@@ -1121,13 +1121,13 @@ std::pair<Base::Vector3d, Base::Vector3d> CenterLine::calcEndPoints2Points(DrawV
 
     double scale = partFeat->getScale();
 
-    std::vector<TechDraw::Vertex*> points;
+    std::vector<TechDraw::VertexPtr> points;
     for (auto& vn: vertNames) {
         if (TechDraw::DrawUtil::getGeomTypeFromName(vn) != "Vertex") {
             continue;
         }
         int idx = TechDraw::DrawUtil::getIndexFromName(vn);
-        TechDraw::Vertex* v = partFeat->getProjVertexByIndex(idx);
+        TechDraw::VertexPtr v = partFeat->getProjVertexByIndex(idx);
         if (v != nullptr) {
             points.push_back(v);
         }
--- a/src/Mod/TechDraw/App/CosmeticExtension.cpp
+++ b/src/Mod/TechDraw/App/CosmeticExtension.cpp
@@ -115,7 +115,7 @@ TechDraw::CosmeticVertex* CosmeticExtension::getCosmeticVertexBySelection(std::s
         return result;
     }
     int idx = DrawUtil::getIndexFromName(name);
-    TechDraw::Vertex* v = dvp->getProjVertexByIndex(idx);
+    TechDraw::VertexPtr v = dvp->getProjVertexByIndex(idx);
     if (v == nullptr) {
         return result;
     }
--- a/src/Mod/TechDraw/App/DrawDimHelper.cpp
+++ b/src/Mod/TechDraw/App/DrawDimHelper.cpp
@@ -126,8 +126,8 @@ void DrawDimHelper::makeExtentDim(DrawViewPart* dvp,
     std::vector<std::string> cvTags;
     std::string tag0;
     std::string tag1;
-    TechDraw::Vertex* v0 = nullptr;
-    TechDraw::Vertex* v1 = nullptr;
+    TechDraw::VertexPtr v0;
+    TechDraw::VertexPtr v1;
     if (subElements.size() > 1) {
         int idx0 = DrawUtil::getIndexFromName(subElements[0]);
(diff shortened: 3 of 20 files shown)


chennes:
Is there something in particular you want reviewed here, or just normal testing?

wwmayer:
> Is there something in particular you want reviewed here

Well, TD is not my domain and I didn't want to simply push a fix that modifies many files that others might not be happy with. That's why I wanted to hear a second opinion.

The technical aspect:
The main problem with TD is that in many cases there is no clearly defined ownership of objects so that either some objects are never deleted or other objects twice. The bug report was about the latter.

Because of this I replaced the affected raw pointers to shared pointers to avoid double deletion and fix possible memory leaks.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5231">pipeline status</a> for feature branch [PR_5231](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5231). Pipeline [425424877](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/425424877) was triggered at [89d57d4](https://github.com/FreeCAD/FreeCAD/pull/5231/commits/89d57d4ece2c652220df035c9877400c7020d5d5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
If nobody has any reservations I will merge it shortly.

chennes:
I'm always a fan of smart pointers in ownership situations, so I think this is an excellent thing.

donovaly:
> The main problem with TD is that in many cases there is no clearly defined ownership of objects so that either some objects are never deleted or other objects twice.

Yes. And TD in general needs a thorough review. I once worked to fix the mess with pages and its displayed childs but I felt I could only cure symptoms, not the main causes.
In my opinion the main issues with TD are:
- TD is slow. Even with relatively simple assemblies and parts, it takes quickly ages (Wandererfan and Klaus, the maintainer of A2plus, once tried to improve the situation but both can no loner work for FC and I don't know enough about the underlying problem.)
- no other WB is recompiled so often when code in FC changes by MSVC. So there must be a lot of unnecessary dependencies.
- there are many hacks in the code Wandererfan was also aware of but never had time to look at. He told me that he took over the code from someone else and often does not understand these hacks as well.

Long story short, when you see strange code, any improvement like this PR is highly appreciated.
I can help with testing PRs since I have a loft of TD test files collected by time.
(I am by far no TD master. I mainly worked on the UI and fixed bugs. So I more or less only improved TD using the existing infrastructure.)

wwmayer:
> Yes. And TD in general needs a thorough review. I once worked to fix the mess with pages and its displayed childs but I felt I could only cure symptoms, not the main causes.

A general rework at this point is difficult and actually not needed because the basic design still looks OK to me. A good idea always is code-refactoring where big functions are split into smaller ones. This automatically reduces duplicated code and may fix inconsistencies.

> TD is slow....

For this we should have a good test file that eventually could be added to the examples directory. In a second step a profiler must be used to identify the code parts that are called too often.

> no other WB is recompiled so often

TD depends on many other extension modules like Measure, Part, Spreadsheet, Drawing and Import. And of course it depends on the changes of the other modules.
A re-compile may only be needed if there are changes in header files but if only source files are changed then a re-linking is sufficient.

> there are many hacks in the code

I have seen many cases where inside the constructor of a class certain things are checked and if not fulfilled the further initialization is aborted and a message is printed. To avoid possible crashes the affected methods are full of checks not to call the null pointer member variables. In some cases the better alternative would have been to throw an exception in the constructor in order to clean-up the whole instance.

wwmayer:
OK, I press the red button now :)

donovaly:
Do you think this can be backported?

donovaly:
> > TD is slow....
> 
> For this we should have a good test file that eventually could be added to the examples directory. In a second step a profiler must be used to identify the code parts that are called too often.

@kbwbe provided last year some good test cases. I searched the forum but cannot find the thread where we once discussed with @WandererFan .

> > no other WB is recompiled so often
> 
> TD depends on many other extension modules like Measure, Part, Spreadsheet, Drawing and Import. And of course it depends on the changes of the other modules. A re-compile may only be needed if there are changes in header files but if only source files are changed then a re-linking is sufficient.

What I mean is that when TD has to be recompiled because something changed in the code of FC, almost all TD files are recomputed. This is not the case for the other WBs.
Therefore I assume there are some unnecessary dependencies.

edi271:
This PR causes errors in the files CommandExtensionDims.cpp and CommandExtensionPack.cpp used in PR5144.

donovaly:
> This PR causes errors in the files CommandExtensionDims.cpp and CommandExtensionPack.cpp used in PR5144.

This PR is not yet merged and since this PR here is in and changed some D basics, you must rebase your PR to current master and then adapt your code in necessary.
I know that this is additional work for you.

As another reviewers already wrote in your PR, the best is to make PRs that each change a small portion. This assures it will get reviewed quickly and then merged.

PR #5233, opened 2021-12-07
Draft: Fix Part_Line offset issue

Roy-043:
The Draft_Offset command did not work with Part_Lines.

Additionally (offset.py):
`"Draft.offset: Unable to create an offset"` message was printed even if `newobj` was created.

Forum topic:
https://forum.freecadweb.org/viewtopic.php?f=23&t=64346

### DIFF
--- a/src/Mod/Draft/draftfunctions/offset.py
+++ b/src/Mod/Draft/draftfunctions/offset.py
@@ -44,8 +44,8 @@ from draftmake.make_bspline import makeBSpline
 
 def offset(obj, delta, copy=False, bind=False, sym=False, occ=False):
     """offset(object,delta,[copymode],[bind])
-    
-    Offset the given wire by applying the given delta Vector to its first 
+
+    Offset the given wire by applying the given delta Vector to its first
     vertex.
 
     Parameters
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5233">pipeline status</a> for feature branch [PR_5233](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5233). Pipeline [425424893](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/425424893) was triggered at [246f292](https://github.com/FreeCAD/FreeCAD/pull/5233/commits/246f292b56b2f14d30a414d95e438915f3e29178). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Roy-043:
@donovaly Thanks for testing.
This issue is not related to this PR. A default line created with the Part_Primitives command goes from 0,0,0 to 10,10,10. It is not in the global XY plane. If the Draft working plane is set to that plane you will encounter this problem. It does not matter if the line is a Part_Line or a Draft_Line. The object selected for the Draft_Offset command must in the current Draft working plane.

yorikvanhavre:
Yes, and later on there could also be a test in the command to warn the user (or to fail the command) if the base object is not in the WP plane.

I think this can be merged... I let you press the button if OK for you @donovaly

donovaly:
I understand, however from the user's perspective the important point is that my report window is filled with tons of warning messages. This way I cannot quickly see what happened before.

So maybe it can be improved that the warning is only issued once?

donovaly:
> I think this can be merged... 

Done.

donovaly:
I also backported this for the next 0.19 release

PR #5232, opened 2021-12-07
[Core] Fix unit management in QuantitySpinBox

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=3&t=64261

PR #5235, opened 2021-12-08
[Core] Fix Quantity construction with value + unit as string

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=3&t=64261&start=20#p552178

This PR fixes construction of quantity with 'Quantity(double, QString)' that because of Unit implicit conversion leads to weird : 
`Quantity(1.0, QStringLiteral("m")) ==> "1 mm"`

Now : 
* Unit(QString) is made explicit so no conversion can implicitly happen. Notice that behavior is unchanged so `Unit(QStringLiteral("m")).getUnit() ==> "mm"` but this looks correct to me as Unit has internally only unit dimensional awareness (no multiplier awareness)
* Quantity has a dedicated Quantity(double, QString) constructor that tooks into account multiplier induced by unit so now `Quantity(1.0, QStringLiteral("m")) ==> "1 m"`

This PR has impact on #5232 so should be merged or rejected first.

I checked that currently nowhere in the master was used the problematic 'Quantity(double, QString)' construction and that's fine (no current use).


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5235">pipeline status</a> for feature branch [PR_5235](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5235). Pipeline [425425698](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/425425698) was triggered at [2a4452f](https://github.com/FreeCAD/FreeCAD/pull/5235/commits/2a4452fc5f37d10429c779149a57935df4b8cdd4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5235">pipeline status</a> for feature branch [PR_5235](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5235). Pipeline [427881618](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/427881618) was triggered at [3f41c0b](https://github.com/FreeCAD/FreeCAD/pull/5235/commits/3f41c0beebc4ee99a22901151afb4ec11dedaad7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer (approved):
`catch(...)` should be replaced with `catch(const Base::Exception&)`

wwmayer:
Merged this PR as 2fef77e2673 with the above mentioned change.
fb84f715fae0 extends the Python wrapper and adds some unit tests

0penBrain:
> `catch(...)` should be replaced with `catch(const Base::Exception&)`

I (stupidly) copied it from https://github.com/FreeCAD/FreeCAD/blob/fb84f715fae0ffe2f2dbdf7146306e795fd37b75/src/Base/Unit.cpp#L110
Should it be changed there also ?
Can you give some more explanation on this in order to improve next contributions ?

ANyway thx for merging, will now update #5232 accordingly and unleash it ;)

wwmayer:
> Should it be changed there also ?

Yes. Thanks for the hint. Searching for `catch(...)` in the current code base shows that it's used too often and this should be reworked where not appropriate.

> Can you give some more explanation on this in order to improve next contributions ?

In many articles about coding guidelines it's considered bad practice to always handle exceptions with `catch(...)` because it may hide a deeper problem. The expression parser throws a Base::Exception (*) in case of a failure and only this should be handled. In case any other exception type is thrown the `catch(...)` swallows it and since it's not reported one would not even notice that there is a problem somewhere.

(*) IIRC the parser in the past has thrown Base::Exception but now it throws Base::ParserError. So, in the corresponding catch handles we should use Base::ParserError instead.

0penBrain:
@wmayer Thx for details

PR #5237, opened 2021-12-09
[StdCmdSendToPythonConsole] make python console visible, set focus

mwganson:
The only reason you would send an object to the python console is if you intend to type into the console, so it makes sense to make the console visible if it's not already visible and to set the keyboard focus to the console.  This way when entering Ctrl+Shift+P one can simply continue typing without need to click with the mouse to set the keyboard focus or to open the console if it is currently closed.

---

### DIFF
--- a/src/Gui/CommandFeat.cpp
+++ b/src/Gui/CommandFeat.cpp
@@ -24,10 +24,14 @@
 #include "PreCompiled.h"
 
 #include <App/DocumentObject.h>
+#include <QAction>
+#include <QDockWidget>
 
 #include "Application.h"
 #include "CommandT.h"
+#include "DockWindowManager.h"
 #include "Document.h"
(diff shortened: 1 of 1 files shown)


src/Gui/CommandFeat.cpp
```diff
@@ -24,10 +24,14 @@
 #include "PreCompiled.h"
 
 #include <App/DocumentObject.h>
+#include <QAction>
+#include <QDockWidget>
```
chennes, line 28: Consider
```suggestion
#ifndef _PreComp_
# include <QAction>
# include <QDockWidget>
#endif
```

wwmayer:
6c065300a8

wwmayer:
Plus code refactoring:
055bcbbe1a

PR #5240, opened 2021-12-09
ReverseEngineering: translate doxygen from DE to EN 

luzpaz:
For the purpose of making the source documentation uniform, source comments in this file were translated to english.

CC wmayer (original author) for review


wwmayer:
Thanks. I made a few small corrections. Merged as d87b45b214

PR #5242, opened 2021-12-09
[PATH]  Fix bug with translated UI 

sliptonic:
https://forum.freecadweb.org/viewtopic.php?f=15&t=64336
Many files were touched in this PR.
All the Black reformatting was consolidated to a single commit.
Many of the .UI files have had certain comboboxes marked not translatable to avoid wasted effort.  All these changes are in a single commit.
The remaining commits each fix a single module with the currentText/currentData  bug.  

---

### DIFF
--- a/src/Gui/DockWindowManager.cpp
+++ b/src/Gui/DockWindowManager.cpp
@@ -254,9 +254,13 @@ void DockWindowManager::activate(QWidget* widget)
         par = par->parentWidget();
     }
 
-    if (dw && !dw->toggleViewAction()->isChecked()) {
+    if (!dw) return;
+
+    if (!dw->toggleViewAction()->isChecked()) {
         dw->toggleViewAction()->activate(QAction::Trigger);
     }
+
+    dw->raise();
 }
 
 /**


src/Mod/Path/PathScripts/PathDressupPathBoundaryGui.py
```diff
@@ -147,7 +175,10 @@ def setupFromExisting():
             elif PathJobGui.StockFromExistingEdit.IsStock(self.obj):
                 setupFromExisting()
             else:
-                PathLog.error(translate('PathJob', "Unsupported stock object %s") % self.obj.Stock.Label)
+                PathLog.error(
+                    translate("PathJob", "Unsupported stock object %s")
```
chennes, line 179: Not necessarily for this PR, but long-term -- from a translation point of view, it's best if you can separate the string that must be translated from the part that translators shouldn't edit. So in this case I suggest something like 
```translate("PathJob","Unsupported stock object") + " %s" % self.obj.Stock.Label```
or
```translate("PathJob","Unsupported stock object") + f" {self.obj.Stock.Label}"```
This prevents a translator from accidentally removing the "%s", extra spaces, punctuation, etc. from the string.

File: src/Mod/Path/PathScripts/PathAdaptiveGui.py [hunk 2]
```diff
@@ -45,16 +45,22 @@ def getForm(self):
         formLayout.addRow(QtGui.QLabel("Coolant Mode"), form.coolantController)
 
         # cut region
+
+        form.direction.clear()
         form.Side = QtGui.QComboBox()
-        form.Side.addItem("Inside")
-        form.Side.addItem("Outside")
+        form.side.addItem(
+            QtCore.QT_TRANSLATE_NOOP("Path", "Inside"), "Inside"
```
chennes, line 52: Not for this PR (obviously!) but someday it might be nice to make these Enums instead of strings so that a linter/typechecker can catch mistakes.

src/Mod/Path/PathScripts/PathPocketBaseGui.py
```diff
@@ -22,50 +22,91 @@
...
+
         if not FeatureFacing & self.pocketFeatures():
             form.facingWidget.hide()
             form.clearEdges.hide()
 
         if FeaturePocket & self.pocketFeatures():
             form.extraOffset_label.setText(translate("PathPocket", "Pass Extension"))
-            form.extraOffset.setToolTip(translate("PathPocket", "The distance the facing operation will extend beyond the boundary shape."))
+            form.extraOffset.setToolTip(
+                translate(
```
chennes, line 105: If in this codebase this is a synonym for `QtCore.QT_TRANSLATE_NOOP` I suggest standardizing on one or the other. You'd probably save yourself some vertical space if you used the shorter one, if you're using Black.

src/Mod/Path/PathScripts/PathPreferencesPathJob.py
```diff
@@ -292,25 +398,24 @@ def browseDefaultJobTemplate(self):
         path = self.form.leDefaultJobTemplate.text()
         if not path:
             path = self.bestGuessForFilePath()
-        foo = QtGui.QFileDialog.getOpenFileName(QtGui.QApplication.activeWindow(),
-                "Path - Job Template",
-                path,
-                "job_*.json")[0]
+        foo = QtGui.QFileDialog.getOpenFileName(
+            QtGui.QApplication.activeWindow(), "Path - Job Template", path, "job_*.json"
```
chennes, line 402: I think these window titles need a `translate()` (here and below)

src/Mod/Path/PathScripts/PathSlotGui.py
```diff
@@ -190,37 +254,37 @@ def updateVisibility(self, sentObj=None):
         if hideFeatures:
             # reset values
             self.CATS = [None, None]
-            self.selectInComboBox('Start to End', self.form.pathOrientation)
+            self.selectInComboBox("Start to End", self.form.pathOrientation)
```
chennes, line 257: I have not looked at how `self.selectInComboBox()` works, but you might need to audit your usage of it to ensure that it's going to work with the new user data.

chennes (commented):
I've added some comments below: some of them are clearly "future work", but this seemed like a good opportunity to add them to the radar.

src/Mod/Path/PathScripts/PathSurfaceGui.py
```diff
@@ -46,21 +46,60 @@ def initPage(self, obj):
...
     def getForm(self):
         """getForm() ... returns UI"""
-        return FreeCADGui.PySideUic.loadUi(":/panels/PageOpSurfaceEdit.ui")
+        form = FreeCADGui.PySideUic.loadUi(":/panels/PageOpSurfaceEdit.ui")
+
+        form.boundBoxSelect.clear()
+        form.boundBoxSelect.addItem(QtCore.QT_TRANSLATE_NOOP("Path", "Stock"), "Stock")
+        form.boundBoxSelect.addItem(
+            QtCore.QT_TRANSLATE_NOOP("Path", "BaseBoundBox"), "BaseBoundBox"
+        )
```
Russ4262, line 55: If I may, I suggest a little different approach to this type of solution.  Ideally, translations would be available in the CL version of FreeCAD also.  Making the proposed changes does not allow for this.

I suggest modifying modifying the enumerations lists, or the parent method containing them, to contain these translate commands.  I am referring to https://github.com/sliptonic/FreeCAD/blob/ed945419d416ac6f2d98317aef76ec385d6371f6/src/Mod/Path/PathScripts/PathSurface.py#L202-L215 for this particular operation.   The `opPropertyEnumerations()` referenced with the link returns a list of enumerations so CL users can query the class and identify the options available.  Having these translations there makes sense, as the tooltips for the parent properties are available in the tuples stored in the `opPropertyDefinitions()` method, and those tooltips are translated there.  
 
Putting these proposed translation commands of the enumeration elements within the base module, rather than the GUI module, and then automating the `addItem()` population task for each combobox item makes the translations readily available across the operation.  As the proposed method stands, the enumeration element translations are only available via the GUI, whereas moving these to the appropriate methods in the base operation module will make for a more robust and manageable set of code, and shorter code since we could replace all the `addItem()` calls with `for e in enumerationProps:` iteration handler for a list of tuples that identify combobox items and their respective enumeration properties.

Not all ops make use of the database-style methods, `opPropertyEnumerations()` and similar `opPropertyDefinitions()`, that I put in place when editing Slot, Profile, Surface, and Waterline.  These methods make certain other actions possible for users.

Hope this makes sense.  Just my thoughts.

sliptonic: I've moved the PR to draft.  It looks like we have a constellation of translation related issues.  We need to figure out exactly what to use before deploying the full-scale copy-paste that I tried in this PR.

Russ, I think I understand what you're getting at.  Can you take one module and do a concrete implementation so we can see it?

Russ4262: I was able to do that tonight, though I need to do some lesson plans for my Algebra 1 class now!  Haha.

Refer to https://github.com/sliptonic/FreeCAD/pull/20.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5242">pipeline status</a> for feature branch [PR_5242](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5242). Pipeline [427881631](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/427881631) was triggered at [ed94541](https://github.com/FreeCAD/FreeCAD/pull/5242/commits/ed945419d416ac6f2d98317aef76ec385d6371f6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #5254, opened 2021-12-12
PD: Fix helix property visibility/writability

jbaehr:
With this PR minor tweaks are done to the additive/subtractive helix' properties:
- The `HasBeenEdited` property is only used internally to check wether to
  fill certain other parameters with initial values. So there is no need
  show it in the property grid at all.
- The `Outside` property is only used in the subtractive helix and thus
  hidden for the additive helix.
- Those properties which are calculated according to the current input mode (i.e. not directly entered by the user) are now set read-only. Previously they have been writable but their value was ignored and overwritten at the next recompute
- The input mode and reference axis have been placed further up in the list of properties to become more prominent.

This is a bundle of three independent commits, so I suggest to review them independently. They all belong to the same topic and are each rather small, so I've chosen to put them in a single PR -- I hope you don't mind.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -110,7 +110,7 @@ Helix::Helix()
     Mode.setEnums(ModeEnums);
     ADD_PROPERTY_TYPE(Outside, (long(0)), group, App::Prop_None,
         "If set, the result will be the intersection of the profile and the preexisting body.");
-    ADD_PROPERTY_TYPE(HasBeenEdited, (long(0)), group, App::Prop_None,
+    ADD_PROPERTY_TYPE(HasBeenEdited, (long(0)), group, App::Prop_Hidden,
         "If false, the tool will propose an initial value for the pitch based on the profile bounding box,\n"
         "so that self intersection is avoided.");
 }
@@ -609,9 +609,11 @@ void Helix::handleChangedPropertyType(Base::XMLReader& reader, const char* TypeN
 PROPERTY_SOURCE(PartDesign::AdditiveHelix, PartDesign::Helix)
 AdditiveHelix::AdditiveHelix() {
     addSubType = Additive;
+    Outside.setStatus(App::Property::Hidden, true);
 }
 
 PROPERTY_SOURCE(PartDesign::SubtractiveHelix, PartDesign::Helix)
 SubtractiveHelix::SubtractiveHelix() {
     addSubType = Subtractive;
+    Outside.setStatus(App::Property::Hidden, false);
 }
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -80,9 +80,15 @@ Helix::Helix()
 
     const char* group = "Helix";
     ADD_PROPERTY_TYPE(Base, (Base::Vector3d(0.0, 0.0, 0.0)), group, App::Prop_ReadOnly,
-        "The center point of the helix' start.");
+        "The center point of the helix' start; derived from the reference axis.");
     ADD_PROPERTY_TYPE(Axis, (Base::Vector3d(0.0, 1.0, 0.0)), group, App::Prop_ReadOnly,
-        "The helix' direction.");
+        "The helix' direction; derived from the reference axis.");
+    ADD_PROPERTY_TYPE(ReferenceAxis, (0), group, App::Prop_None,
+        "The reference axis of the helix.");
+    ADD_PROPERTY_TYPE(Mode, (long(initialMode)), group, App::Prop_None,
+        "The helix input mode specifies which properties are set by the user.\n"
+        "Dependent properties are then calculated.");
+    Mode.setEnums(ModeEnums);
     ADD_PROPERTY_TYPE(Pitch, (10.), group, App::Prop_None,
         "The axial distance between two turns.");
     ADD_PROPERTY_TYPE(Height, (30.0), group, App::Prop_None,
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5254">pipeline status</a> for feature branch [PR_5254](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5254). Pipeline [427882327](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/427882327) was triggered at [35d515d](https://github.com/FreeCAD/FreeCAD/pull/5254/commits/35d515d3fba0dbd1ae14906ebfa1454c28ffde04). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -606,6 +609,71 @@ void Helix::handleChangedPropertyType(Base::XMLReader& reader, const char* TypeN
...
+{
+    switch (inputMode)
+    {
+    case HelixMode::pitch_height_angle:
+        // primary input:
+        Pitch.setStatus(App::Property::ReadOnly, false);
+        Height.setStatus(App::Property::ReadOnly, false);
+        Angle.setStatus(App::Property::ReadOnly, false);
+        // derived props:
+        Turns.setStatus(App::Property::ReadOnly, true);
```
donovaly, line 633: Why? I think the idea of your second PR is to calculate the derived values automatically. Then we could keep it that one can change any parameter. This would easy the helix handling since one does not need to take care of the definition mode. One changes for example the growth and the angle is calculated accordingly, no matter what mode the user is in.

jbaehr: If you, as a user, change a parameter that is an output according to the current input mode, that change gets overwritten on the next recompute. To prevent such moments of surprise, I set the calculated properties to read-only.

Take this example: *Input mode* is "pitch-height-angle", your *pitch* is 10mm, your *height* is 30mm; *turns*  is calculated to be 3.0 (= *height* / *pitch* = 30mm / 10mm). Now change *turns* to be 4.0.
- Current master accepts the 4.0, but is has no effect: On the next recompute (usually triggered exactly by the *turns* change already) it gets reset to 3.0 because the current input mode tells *turns*  = *height* / *pitch*.
- With my proposed change this is prevented, the user cannot enter 4.0 for *turns*. The rule is simple: Output properties are not writable for the user. (You are not allowed to change *Base* and *Axis* because they are calculated from the *Reference Axis*. You are not allowed change the *Placement* if it's calculated by the attachment engine, ...)

I fail to see how we can safely allow changes to any property. What do you expect to happen in the above example? Change *pitch* to 7.5mm and keeping a *height* of 30mm? Change *height*  to 40mm and keep the original pitch? Change both to distribute the additional turn equally? In my eyes, neither of these are sane options. I'm open for suggestions, though.

(#5255 does not change this. It just adds the calculation for Angle and Growth, which apparently was forgotten when the growth-mode was introduced)

donovaly: > Take this example: _Input mode_ is "pitch-height-angle", your _pitch_ is 10mm, your _height_ is 30mm; _turns_ is calculated to be 3.0 (= _height_ / _pitch_ = 30mm / 10mm). Now change _turns_ to be 4.0.

Here we must decide what has to be changed. We can a make a survey in the forum. Should I start a survey?
(I bet one beer the outcome is to change the height, not the pitch. :-) )

I would like to keep the fields changeable because when I change the turns, this is obviously the user's intention. So something should then happen. I mean if not the right thing happened, I can change this in a second step.
But sure, this is a matter of taste.

As it is, it should be improved, so either your way of forbidding the change or my way of doing something according to rules we think are in most cases the user's will. My argument is that not doing anything does in 100% of the cases not follow the user's will. Doing something would please him at least by 50 % ;-) But of course, when we do nothing,we no nothing wrong :-)

jbaehr: > > Take this example: _Input mode_ is "pitch-height-angle", your _pitch_ is 10mm, your _height_ is 30mm; _turns_ is calculated to be 3.0 (= _height_ / _pitch_ = 30mm / 10mm). Now change _turns_ to be 4.0.
> 
> Here we must decide what has to be changed. We can a make a survey in the forum. Should I start a survey?

Sure. It's always good to have more feedback. But please include the option of not allowing the change (i.e. this PR), too. Personally, I still think that's the way to go -- but that's maybe just my taste ;-).

jbaehr: Here is another thing to consider: When you keep outputs (according to the current input mode) writable and want those changes to propagate to the respective inputs, you get in trouble if the user has assigned expressions to the inputs. Take the above example again:
> *Input mode* is "pitch-height-angle", your *pitch* is 10mm, your *height* is 30mm; *turns* is calculated to be 3.0 (= *height* / *pitch* = 30mm / 10mm).

Now you want to keep *turns* writable and want to keep the *pitch* stable and change *height* in case of a write to the output *turns*. What do you expect to happen if *height* is backed by an expression?

donovaly: > Now you want to keep _turns_ writable and want to keep the _pitch_ stable and change _height_ in case of a write to the output _turns_. What do you expect to happen if _height_ is backed by an expression?

Very good point. So you are right, we should keep these parameters that are not changeable directly as read-only.

However for me as merger, it would be good to have your other PR in first (with the requested UI changes). When you updated this PR to be merged, can you please rebase this one here?
This way I hope to test everything properly.

jbaehr: Thanks. Rebased to the latest master.
Or did you mean I should rebase this one (i.e. #5254) onto the rebased and fixed other (i.e. #5255) ?

PR #5255, opened 2021-12-12
PD: Fix assignment of helix' Angle/Growth properties

jbaehr:
Depending on the input mode, either `Growth` or `Angle` is used to create a conical or flat spiral. The respective other can easily be calculated to give consistent view. This is already done for the other properties, e.g. `Height` is calculated from `Pitch` and `Turns`.
With this PR, the same is done for `Angle` and `Growth`.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -136,6 +136,7 @@ App::DocumentObjectExecReturn* Helix::execute(void)
         if (Height.getValue() < Precision::Confusion())
             return new App::DocumentObjectExecReturn("Error: height too small!");
         Turns.setValue(Height.getValue() / Pitch.getValue());
+        Growth.setValue(Pitch.getValue() * tan(Base::toRadians(Angle.getValue())));
     }
     else if (mode == HelixMode::pitch_turns_angle) {
         if (Pitch.getValue() < Precision::Confusion())
@@ -143,6 +144,7 @@ App::DocumentObjectExecReturn* Helix::execute(void)
         if (Turns.getValue() < Precision::Confusion())
             return new App::DocumentObjectExecReturn("Error: turns too small!");
         Height.setValue(Turns.getValue() * Pitch.getValue());
+        Growth.setValue(Pitch.getValue() * tan(Base::toRadians(Angle.getValue())));
     }
     else if (mode == HelixMode::height_turns_angle) {
         if (Height.getValue() < Precision::Confusion())
@@ -150,6 +152,7 @@ App::DocumentObjectExecReturn* Helix::execute(void)
         if (Turns.getValue() < Precision::Confusion())
             return new App::DocumentObjectExecReturn("Error: turns too small!");
         Pitch.setValue(Height.getValue() / Turns.getValue());
+        Growth.setValue(Pitch.getValue() * tan(Base::toRadians(Angle.getValue())));
     }
     else if (mode == HelixMode::height_turns_growth) {
         if (Turns.getValue() < Precision::Confusion())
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5255">pipeline status</a> for feature branch [PR_5255](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5255). Pipeline [427882329](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/427882329) was triggered at [05384bf](https://github.com/FreeCAD/FreeCAD/pull/5255/commits/05384bf0b16ddfd3077169d21eb0b30616f64e98). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks for the PR.
It works but it is incomplete. Also when changing the mode the calculation must be performed.
- Take for example this file: [Helix-Tests.zip](https://github.com/FreeCAD/FreeCAD/files/7700083/Helix-Tests.zip)
- open the helix
- change to Pitch-Height-Angle mode
result: the angle is 0°

Also the other way round does not work:
- in the example change to Pitch-Height-Angle mode and set there an angle of 40°
- go to the growth mode
result: the growth is not recalculated according to the angle.

jbaehr:
> It works but it is incomplete. Also when changing the mode the calculation must be performed.
> 
> * Take for example this file: [Helix-Tests.zip](https://github.com/FreeCAD/FreeCAD/files/7700083/Helix-Tests.zip)
> * open the helix
> * change to Pitch-Height-Angle mode
>   result: the angle is 0°

When changing only the mode, a recalculation of the derived values should not be necessary -- they should have the right value already. For me it looks like your issue is the following:
- The last recompute of your model was done *before* applying my patch (that's why *Angle* is still 0°)
- Now, the first change *after* applying my patch is the change of the mode. This change triggers a recompute and now the wrong 0° *Angle*  is a primary input, so the *Growth* is correctly calculated to 0 mm.

I guess there is not much I can do here. I cannot change an inconsistency from the past. A single recompute (prior to changing the *mode*) fixes this by ensuring all properties to be consistent again.

Performing the calculation when the mode is changed won't help here: I will only see the new mode, but to resolve the original inconsistency I would need the old mode, too. Without this, I don't know whether to trust your *angle* or your *growth*.

> Also the other way round does not work:
> 
> * in the example change to Pitch-Height-Angle mode and set there an angle of 40°
> * go to the growth mode
>   result: the growth is not recalculated according to the angle.

This is something I cannot reproduce here.
* I load your example. It has a *mode* of "height-turns-growth", a *growth* of 20mm and an inconsistent *angle* value of 0.00° (because last recompute happened without my patch; should be 45°)
* I change the *mode* to "pitch-height-angle" (having #5254 applied, too, this makes *Angle* writable). *Growth* gets calculated to 0.0mm; which is correct, because *angle* is (wrongly) 0.0°.
* Now I set the *angle* to 40°.
* *Growth* is calculated to 16.78mm as expected.

May it be that you have automatic recomputes disabled? Without a recompute, derived values are not calculated yet -- obviously. Does an explicit recompute of the helix fixes the problem for you?

donovaly:
> This is something I cannot reproduce here.

I cannot reporduce that you cannot reporduce ;-). The point is that you don't update when the mode is changed. Here a screencast of what I mean, look at the angle
- we start with 45°
- I change the groth, so as the result the angle must be greater than 45°
- I switch the mode and indded the angle is still at 45°
- first when I accept tha change the angle is calculated correctly
a2G4Bl5NrH

Here is the file I used for the screencast:
[Helix-Tests.zip](https://github.com/FreeCAD/FreeCAD/files/7723992/Helix-Tests.zip)

jbaehr:
> I cannot reproduce that you cannot reproduce ;-). The point is that you don't update when the mode is changed. Here a screencast of what I mean, look at the angle

Ah, now I see. We've been talking about two different things: I did the changes in the property grid, where a change in the mode *does* trigger a recompute (like all property changes). You did them in the task panel, where obviously is something missing.
Thanks. I'll take a look.

jbaehr:
Rebased and fixed.

Actually, the calculations have been performed when using the task panel, too. Just the UI update was missing.

Beware though, that when loading an old model, where the last recompute was done *without* my patch, you will still experience the inconsistency as explained in my first comment. The first recompute *with* my patch fixes this, though, and after that also the task panel's UI is now updated as it should.

donovaly:
> Rebased and fixed.

Thanks, merged.

PR #5144, opened 2021-11-04
Next31TechDrawTools

edi271:
Thank you for creating a pull request to contribute to FreeCAD! To ease integration, we ask you to conform to the following items. Pull requests which don't satisfy all the items below might be rejected. If you are in doubt with any of the items below, don't hesitate to ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10)!

- [ ]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [ ]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [ ]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [ ]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [ ]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [ ]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the [FreeCAD bug tracker issue number](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) in case a particular commit solves or is related to an existing issue on the tracker. Ex: `Draft: fix typos - fixes #0004805`

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.20 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=56135).

---

PR #5257, opened 2021-12-14
[Sketcher] ViewProvider refactor

abdullahtahiriyo:
======================================================

Creation of EditModeCoinManager class and helpers.

In a nutshell:
- EditModeCoinManager gets most of the content of struct EditData
- Drawing is partly outsourced to EditModeCoinManager
- EditModeCoinManager gets a nested Observer class to deal with parameters
- A struct DrawingParameters is created to store all parameters used for drawing
- EditModeCoinManager assume responsibility for determining the drawing size of the Axes
- Preselection detection responsibility is moved to EditModeCoinManager.
- Generation of constraint nodes and constraint drawing is moved to EditModeCoinManager.
- Constraint generation parameters are refactored into ConstraintParameters.
- Text rendering functions are moved to EditModeCoinManager.
- Move HDPI resolution responsibility from VPSketch to EditModeCoinManager
- Move responsibility to create the scenograph for edit mode to EditModeCoinManager
- Move full updateColor responsibility to EditModeCoinManager
- Allows for mapping N logical layers (LayerId of GeometryFacade) to M coin Layers (M<N). This
is convenient as, unless the representation must be different, there is no point in creating coin
layers (overhead).

Refactoring of geometry drawing:
- Determination of the curve values to draw are outsourced to OCC (SRP and remove code duplications).
- Refactor specific drawing of each geometry type into a single template method, based on classes of geometry.
- Drawing of geometry and constraints made agnostic of scale factors of BSpline weights so that a uniform treatment can be provided.

Refactoring of Overlay Layer:
- A new class EditModeInformationOverlayConverter is a full rewrite of the previous overlay routines.

ViewProviderSketch:
- Major cleanup due to migration of functionalities to EditModeCoinManager
- Reduce public api of ViewProviderSketch due to refactor of DrawSketchHandler
- Major addition of documentation
- ShortcutListener implementation using new ViewProvider Attorney
- Gets a parameter handling nested class to handle all parameters (observer)
- Move rubberband to smart pointer
- Refactor selection and preselection into nested classes
- Removal of SEL_PARAMS macro. This macro was making the code unreadable as it "captured" a local stringstream that appeared unused. Substituted by local private member functions.
- Remove EditData
- Improve documentation
- Refactor Preselection struct to remove magical numbers
- Refactor Selection mechanism to remove hacks

ViewProviderSketchDrawSketchHandlerAttorney:
- new Attorney to limit access to ViewProviderSketch and reduce its public interface
- In order to enforce a certain degree of encapsulation and promote a not too tight coupling, while still allowing well
defined collaboration, DrawSketchHandler accesses ViewProviderSketch via this Attorney class.
-DrawSketchHandler has the responsibility of drawing edit temporal curves and markers necessary to enable visual feedback
to the user, as well as the UI interaction during such edits. This is its exclusive responsibility under the Single
Responsibility Principle.
- A plethora of speciliased handlers derive from DrawSketchHandler for each specialised editing (see for example all the
handlers for creation of new geometry). These derived classes do * not * have direct access to the
ViewProviderSketchDrawSketchHandlerAttorney. This is intentional to keep coupling under control. However, generic
functionality requiring access to the Attorney can be implemented in DrawSketchHandler and used from its derived classes
by virtue of the inheritance. This promotes a concentrating the coupling in a single point (and code reuse).

EditModeCoinManager:
- Refactor of updateConstraintColor
- Multifield - new struct to identify a single element in a multifield field per layer
- Move geometry management to delegate class EditModeCoinGeometryManager
- Remove refactored code that was never used in the original ViewProviderSketch.

CommandSketcherBSpline:
- EditModeCoinManager automatically tracks parameter change and triggers the necessary redraw, rendering an explicit redraw obsolete and unnecessary.

Rebase on top of master:
- Commits added to master to ViewProviderSketch applied to EditModeCoinManager.
- Memory leaks - wmayer
- Constraint Diameter Symbol - OpenBrain
- Minor bugfix to display angle constraints - syres

---

### DIFF
--- a/src/Mod/Part/App/Geometry.cpp
+++ b/src/Mod/Part/App/Geometry.cpp
@@ -298,7 +298,7 @@ bool Geometry::hasExtension(Base::Type type) const
     return false;
 }
 
-bool Geometry::hasExtension(std::string name) const
+bool Geometry::hasExtension(const std::string & name) const
 {
     for( auto ext : extensions) {
         if(ext->getName() == name)
@@ -318,7 +318,7 @@ std::weak_ptr<GeometryExtension> Geometry::getExtension(Base::Type type)
     throw Base::ValueError("No geometry extension of the requested type.");
 }
 
-std::weak_ptr<GeometryExtension> Geometry::getExtension(std::string name)
+std::weak_ptr<GeometryExtension> Geometry::getExtension(const std::string & name)
 {
     for( auto ext : extensions) {
         if(ext->getName() == name)
@@ -333,7 +333,7 @@ std::weak_ptr<const GeometryExtension> Geometry::getExtension(Base::Type type) c
     return const_cast<Geometry*>(this)->getExtension(type).lock();
 }
 
-std::weak_ptr<const GeometryExtension> Geometry::getExtension(std::string name) const
+std::weak_ptr<const GeometryExtension> Geometry::getExtension(const std::string & name) const
 {
     return const_cast<Geometry*>(this)->getExtension(name).lock();
 }
@@ -371,7 +371,7 @@ void Geometry::deleteExtension(Base::Type type)
         extensions.end());
 }
 
-void Geometry::deleteExtension(std::string name)
+void Geometry::deleteExtension(const std::string & name)
 {
     extensions.erase(
         std::remove_if( extensions.begin(),
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -1425,6 +1425,22 @@ std::vector<Part::Geometry *> Sketch::extractGeometry(bool withConstructionEleme
     return temp;
 }
 
+GeoListFacade Sketch::extractGeoListFacade() const
+{
+    std::vector<GeometryFacadeUniquePtr> temp;
+    temp.reserve(Geoms.size());
+    int internalGeometryCount = 0;
+    for (std::vector<GeoDef>::const_iterator it=Geoms.begin(); it != Geoms.end(); ++it) {
+        auto gf = GeometryFacade::getFacade(it->geo->clone(), true); // GeometryFacade is the owner of this allocation
+        if(!it->external)
+            internalGeometryCount++;
+
+        temp.push_back(std::move(gf));
+    }
+
+    return GeoListFacade::getGeoListModel(std::move(temp), internalGeometryCount);
+}
+
 void Sketch::updateExtension(int geoId, std::unique_ptr<Part::GeometryExtension> && ext)
 {
     geoId = checkGeoId(geoId);
--- a/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
+++ b/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
@@ -289,7 +289,7 @@ public:
             setPositionText(onSketchPos, text);
 
             EditCurve[1] = onSketchPos;
-            sketchgui->drawEdit(EditCurve);
+            drawEdit(EditCurve);
             if (seekAutoConstraint(sugConstr2, onSketchPos, onSketchPos - EditCurve[0])) {
                 renderSuggestConstraintsCursor(sugConstr2);
                 return;
@@ -306,7 +306,7 @@ public:
         }
         else {
             EditCurve[1] = onSketchPos;
-            sketchgui->drawEdit(EditCurve);
+            drawEdit(EditCurve);
             Mode = STATUS_End;
         }
         return true;
@@ -353,7 +353,7 @@ public:
             tryAutoRecomputeIfNotSolve(static_cast<Sketcher::SketchObject *>(sketchgui->getObject()));
 
             EditCurve.clear();
-            sketchgui->drawEdit(EditCurve);
+            drawEdit(EditCurve);
 
             bool continuousMode = hGrp->GetBool("ContinuousCreationMode",true);
             if(continuousMode){
@@ -484,7 +484,7 @@ public:
                 EditCurve[4] = EditCurve[0];
             }
 
-            sketchgui->drawEdit(EditCurve);
+            drawEdit(EditCurve);
             if (seekAutoConstraint(sugConstr2, onSketchPos, Base::Vector2d(0.0,0.0))) {
                 renderSuggestConstraintsCursor(sugConstr2);
                 return;
@@ -512,7 +512,7 @@ public:
                 EditCurve[2] = onSketchPos;
                 EditCurve[1] = Base::Vector2d(onSketchPos.x ,EditCurve[0].y);
                 EditCurve[3] = Base::Vector2d(EditCurve[0].x,onSketchPos.y);
-                sketchgui->drawEdit(EditCurve);
+                drawEdit(EditCurve);
                 Mode = STATUS_End;
             }
             else if (constructionMethod == CenterAndCorner) {
@@ -521,7 +521,7 @@ public:
                 EditCurve[2] = onSketchPos;
                 EditCurve[3] = Base::Vector2d(onSketchPos.x,EditCurve[0].y);
                 EditCurve[4] = EditCurve[0];
-                sketchgui->drawEdit(EditCurve);
+                drawEdit(EditCurve);
                 Mode = STATUS_End;
             }
         }
@@ -659,7 +659,7 @@ public:
             // This code enables the continuous creation mode.
                 Mode=STATUS_SEEK_First;
                 EditCurve.clear();
-                sketchgui->drawEdit(EditCurve);
+                drawEdit(EditCurve);
                 EditCurve.resize(5);
                 applyCursor();
                 /* this is ok not to call to purgeHandler
@@ -855,7 +855,7 @@ public:
             text.sprintf(" (%.1fR %.1fX %.1fY)", radius, lengthX, lengthY);
             setPositionText(onSketchPos, text);
 
-            sketchgui->drawEdit(EditCurve);
+            drawEdit(EditCurve);
             if (seekAutoConstraint(sugConstr2, onSketchPos, Base::Vector2d(0.f, 0.f))) {
                 renderSuggestConstraintsCursor(sugConstr2);
                 return;
@@ -1027,7 +1027,7 @@ public:
                 // This code enables the continuous creation mode.
                 Mode = STATUS_SEEK_First;
                 EditCurve.clear();
-                sketchgui->drawEdit(EditCurve);
+                drawEdit(EditCurve);
                 EditCurve.resize(37);
                 applyCursor();
                 /* this is ok not to call to purgeHandler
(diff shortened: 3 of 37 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5257">pipeline status</a> for feature branch [PR_5257](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5257). Pipeline [429647368](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/429647368) was triggered at [5f44039](https://github.com/FreeCAD/FreeCAD/pull/5257/commits/5f440392f1294caac6c3257647bb9f1637fb70c4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
Once this branch is merged the function GetCircleCenter should be removed and instead Part::Geom2dCircle::getCircleCenter should be used.

This is because in PR #5144 on old version of the function has been copied to TD that lacks of an important bug fix.

See 955fe455cb

abdullahtahiriyo:
@wwmayer 

Thanks for the heads up. 

I had to rebase the PR to master due to a conflict with a new commit, so I took the opportunity to remove the function and use its new Part counterpart.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5257">pipeline status</a> for feature branch [PR_5257](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5257). Pipeline [432835521](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/432835521) was triggered at [8237a20](https://github.com/FreeCAD/FreeCAD/pull/5257/commits/8237a20d839f1e1f55bb2b5a007c7b2413653654). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/App/Sketch.cpp
```diff
@@ -1425,6 +1425,22 @@ std::vector<Part::Geometry *> Sketch::extractGeometry(bool withConstructionEleme
     return temp;
 }
 
+GeoListFacade Sketch::extractGeoListFacade() const
+{
+    std::vector<GeometryFacadeUniquePtr> temp;
+    temp.reserve(Geoms.size());
+    int internalGeometryCount = 0;
+    for (std::vector<GeoDef>::const_iterator it=Geoms.begin(); it != Geoms.end(); ++it) {
```
0penBrain, line 1433: Prefer a ranged for loop if possible

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
```diff
@@ -6887,7 +6844,8 @@ namespace SketcherGui {
 
         virtual void deactivated(ViewProviderSketch *sketchgui)
         {
-            sketchgui->setAxisPickStyle(true);
+            Q_UNUSED(sketchgui);
```
0penBrain, line 6847:
```suggestion
            (void) sketchgui;
```
I'm always a bit disturbed when using Qt macro when there is no strong link to Qt (not inside a QObject or other). But maybe that's personal taste only. ;)

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
```diff
@@ -6703,7 +6659,8 @@ class DrawSketchHandlerExternal: public DrawSketchHandler
 
     virtual void deactivated(ViewProviderSketch *sketchgui)
     {
-        sketchgui->setAxisPickStyle(true);
+        Q_UNUSED(sketchgui);
```
0penBrain, line 6662:
```suggestion
            (void) sketchgui;
```
I'm always a bit disturbed when using Qt macro when there is no strong link to Qt (not inside a QObject or other). But maybe that's personal taste only. ;)

src/Mod/Sketcher/Gui/EditModeCoinManager.h
```diff
@@ -0,0 +1,293 @@
...
+     *
+     * In other words valid values are 0,1,2,... for normal geometry and -3,-4,-5,... for external geometry
+     *
+     * Cross: Axes and RootPoint values as defined in the enum class.
+     *
+     */
+    struct PreselectionResult {
+        enum SpecialValues {
+            InvalidPoint = -1,
+            InvalidCurve = -1,
```
0penBrain, line 162: Suspect value, -1 is used twice in an enum

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.h
```diff
@@ -0,0 +1,253 @@
...
+ *  @details    To be documented.
+ *
+ */
+class SketcherGuiExport EditModeConstraintCoinManager
+{
+private:
+    /// Coin Node indices for constraints
+    enum class ConstraintNodePosition {
+        MaterialIndex = 0,
+        DatumLabelIndex = 0,
```
0penBrain, line 77: Suspect value, 0 used twice in an enum

src/Mod/Sketcher/Gui/ViewProviderSketch.h
```diff
@@ -76,37 +81,320 @@ namespace Sketcher {
...
+      * Ids are zero-indexed points and curves.
+      *
+      * The DragPoint indexing matches PreselectPoint indexing.
+      *
+      */
+    class Drag {
+    public:
+        enum SpecialValues {
+            InvalidPoint = -1,
+            InvalidCurve = -1
```
0penBrain, line 242: Suspect value

src/Mod/Sketcher/Gui/ViewProviderSketch.h
```diff
@@ -76,37 +81,320 @@ namespace Sketcher {
...
+      *
+      * The PreselectPoint indexing matches DragPoint indexing (it further includes negative edges, which are
+      * not meaningful for Dragging).
+      *
+      */
+    class Preselection {
+    public:
+        enum SpecialValues {
+            InvalidPoint = -1,
+            InvalidCurve = -1,
```
0penBrain, line 296: Suspect value

0penBrain (commented):
Made a very fast code review as no spare time ATM.
I marked some enum values as suspect but they can be legitimate (there are ones I'm sure are).
Hopefully helpful.

src/Mod/Part/App/Geometry.cpp
```diff
@@ -318,7 +318,7 @@ std::weak_ptr<GeometryExtension> Geometry::getExtension(Base::Type type)
     throw Base::ValueError("No geometry extension of the requested type.");
 }
 
-std::weak_ptr<GeometryExtension> Geometry::getExtension(std::string name)
+std::weak_ptr<GeometryExtension> Geometry::getExtension(const std::string & name)
```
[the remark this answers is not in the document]
abdullahtahiriyo: Thanks. Const correctness is very important to me.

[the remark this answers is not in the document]
abdullahtahiriyo: Wait. But getExtension is const overloaded to be most const correct:
```
    std::weak_ptr<const GeometryExtension> getExtension(Base::Type type) const;
    std::weak_ptr<const GeometryExtension> getExtension(const std::string & name) const;
    std::weak_ptr<GeometryExtension> getExtension(Base::Type type);
    std::weak_ptr<GeometryExtension> getExtension(const std::string & name);
```
If the object on which is called is const, then a const weak_ptr is returned. If the object on which is called is not const, then a normal weak_ptr is returned. A normal weak_ptr can be used to change the object (the extension). A const weak_ptr emphasises that the object is not supposed to be changed without violating const correctness.

Do I make sense?

src/Mod/Sketcher/App/Sketch.cpp
abdullahtahiriyo: I must admit I am still conflicted by the "prefer range loop". 

Until now I have adopted the view that it is a matter of style and preference.

We don't have containers with copy-on-write semantics (to my knowledge), so I guess we could opt for range loops. Even with them, it is still possible to use std::as_const (c++17) to get away with it. However, the strong argument for readability no longer holds when one needs to write ```for(const auto & v : std::as_const(container))```.

Let me think about it.

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
abdullahtahiriyo: Q_UNUSED is pervasive in FreeCAD. I am not a fan, but it is the "Werner approved" way. 

Personally, I think we should start using the ```maybe_unused``` attribute (since c++17), as it is a language feature.

See:
https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

src/Mod/Sketcher/Gui/EditModeCoinManager.h
abdullahtahiriyo: It is valid and intended. 

However, this is something I want to refactor further at a second stage. I have "improved" it, so that it is clear what the magic numbers are (and what the intention is). But, I still do not like the fact that preselection goes one way (separate cross, different values), selection another way, and this lack of unity makes every struct has its "Special Values".

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.h
abdullahtahiriyo: It is valid and intended.

Some constraints are built having children 0 to be the MaterialIndex and others the DatumLabelIndex. This removes quite some magic numbers and awful macros. However, when the Constraints handling in coin is refactored (this is Werner's comment on the forum), it may be possible to improve it further.

src/Mod/Sketcher/Gui/ViewProviderSketch.h
abdullahtahiriyo: Idem.

src/Mod/Sketcher/Gui/ViewProviderSketch.h
abdullahtahiriyo: Idem

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
wwmayer: > I'm always a bit disturbed when using Qt macro when there is no strong link to Qt (not inside a QObject or other).

Well, we can also define our own macro FC_UNUSED.

> Q_UNUSED is pervasive in FreeCAD. I am not a fan

I prefer it over (void) because IMO it's more readable and because of capital letters hardly to overlook.

> Personally, I think we should start using the maybe_unused attribute (since c++17), as it is a language feature.

That's an option, too.

PR #5259, opened 2021-12-15
[MainWindow] Add toolbar name as toolbar's  tooltip

mwganson:
forum topic:
https://forum.freecadweb.org/viewtopic.php?f=34&t=64528

Adds the name of the toolbar as a tooltip to the toolbar in between braces.  Example, if toolbar name is File, then the tooltip is [File].  Can be disabled in user parameters:  BaseApp/Preferences/MainWindow/ToolBarNameAsToolTip (boolean, default = true).

### DIFF
--- a/src/Gui/ToolBarManager.cpp
+++ b/src/Gui/ToolBarManager.cpp
@@ -205,10 +205,10 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
                                         toolbarName.c_str())); // i18n
             toolbar->setObjectName(name);
             if (nameAsToolTip){
-                std::ostringstream tooltip;
-
-                tooltip << "[" << toolbarName << "]";
-                toolbar->setToolTip(QApplication::translate("Workbench", tooltip.str().c_str()));
+                auto tooltip = QChar::fromLatin1('[')
+                    + QApplication::translate("Workbench", toolbarName.c_str())
+                    + QChar::fromLatin1(']');
+                toolbar->setToolTip(tooltip);
             }
             toolbar->setVisible(visible);
             toolbar_added = true;
--- a/src/Gui/ToolBarManager.cpp
+++ b/src/Gui/ToolBarManager.cpp
@@ -186,6 +186,8 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
 
     ParameterGrp::handle hPref = App::GetApplication().GetUserParameter().GetGroup("BaseApp")
                                ->GetGroup("MainWindow")->GetGroup("Toolbars");
+    bool nameAsToolTip = App::GetApplication().GetUserParameter().GetGroup("BaseApp")
+            ->GetGroup("Preferences")->GetGroup("MainWindow")->GetBool("ToolBarNameAsToolTip",true);
     QList<ToolBarItem*> items = toolBarItems->getItems();
     QList<QToolBar*> toolbars = toolBars();
     for (QList<ToolBarItem*>::ConstIterator it = items.begin(); it != items.end(); ++it) {
@@ -202,6 +204,12 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
                 QApplication::translate("Workbench",
                                         toolbarName.c_str())); // i18n
             toolbar->setObjectName(name);
+            if (nameAsToolTip){
+                std::ostringstream tooltip;
+
+                tooltip << "[" << toolbarName << "]";
+                toolbar->setToolTip(QApplication::translate("Workbench", tooltip.str().c_str()));
+            }
             toolbar->setVisible(visible);
             toolbar_added = true;
         }


0penBrain (approved):
Looks good

luzpaz:
Sweet!  
@mwganson can we get a screenshot of how this looks?

src/Gui/ToolBarManager.cpp
```diff
@@ -202,6 +204,9 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
                 QApplication::translate("Workbench",
                                         toolbarName.c_str())); // i18n
             toolbar->setObjectName(name);
+            if (nameAsToolTip){
+                toolbar->setToolTip(QLatin1String("[") + name + QLatin1String("]"));
```
donovaly, line 208: The 'name' is hereby the technical/internal toolbar name. Sine you expose it as tooltip, it must be translatable.

This line instead fixes the issue for me:

`toolbar->setToolTip(QApplication::translate("Workbench", toolbarName.c_str()));`

I would omit the brackets in the tooltip but this is of course a matter of taste.

donovaly (changes requested):

donovaly:
Many thanks, good idea.

>  Can be disabled in user parameters: BaseApp/Preferences/MainWindow/ToolBarNameAsToolTip (boolean, default = true).

Please don't forget to document this in the Wiki.

src/Gui/ToolBarManager.cpp
mwganson: Okay, I'll make it translated.  I wanted the brackets to distinguish it from other tooltips for the various commands.

mwganson:
Snip macro screenshot-c6dcbc

src/Gui/ToolBarManager.cpp
```diff
@@ -202,6 +204,11 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
                 QApplication::translate("Workbench",
                                         toolbarName.c_str())); // i18n
             toolbar->setObjectName(name);
+            if (nameAsToolTip){
+                std::stringstream tooltip;
```
0penBrain, line 208:
```suggestion
                std::ostringstream tooltip;

```

0penBrain:
@mwganson would you be able to squash the commits ? Thx

adrianinsaval:
what even is the usecase for disabling a tooltip? That's overkill IMO

luzpaz:
The original request in the forum showed the toolbar name bolded, which I kind of thought made sense. Could we try that ?

0penBrain:
@luzpaz This is boring to have to manage string formatting in tooltips while it doesn't bring anything useful IMO. Not worth the effort. ;)

luzpaz:
Fair enough. Not worth arguing the point about personal taste/preference.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5259">pipeline status</a> for feature branch [PR_5259](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5259). Pipeline [431387985](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/431387985) was triggered at [9d0bcd1](https://github.com/FreeCAD/FreeCAD/pull/5259/commits/9d0bcd1d5c96aef6c98d98b6829d30de749a7aa7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/ToolBarManager.cpp
```diff
@@ -202,6 +204,12 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
                 QApplication::translate("Workbench",
                                         toolbarName.c_str())); // i18n
             toolbar->setObjectName(name);
+            if (nameAsToolTip){
+                std::ostringstream tooltip;
+
+                tooltip << "[" << toolbarName << "]";
+                toolbar->setToolTip(QApplication::translate("Workbench", tooltip.str().c_str()));
```
donovaly, line 211: This doesn't work because you introduce a new string to be translated first. Change your UI language to see that.
The aim is to reduce the translation work, therefore we should take the name that is already translated and concatenate the brackets around it.

donovaly (changes requested):

donovaly:
Fix applied -> merging

luzpaz:
Regular tooltips: 
image

Tooltip for toolbar:
image

donovaly:
> Regular tooltips:

What do you mean, what is wrong in your opinion?

luzpaz:
To me, It doesn't look uniform with the regular toolbar. IMHO, think it's worth bolding the toolbar name, remove the square paranthesis and append the word `toolbar` to it

adrianinsaval:
It might look better but I think it will bring unnecessary work for translators as it's probably not enough to separately translate the workbench name and append the translation of toolbar, for starter the correct order for them is different depending on language.

luzpaz:
I usually agree with not creating unnecessary work for the translators, but in this case IMHO falls in to that exception. If there is a grammatical concern we can use a `Toolbar: NameOfToolBar` approach?

donovaly:
To be honest this is a bit too much into details and not worth it.
We have the new feature to show the toolbar name. That is the real benefit and that the tooltip is different is sensible to distinguish it from the other tooltips.

We can debate about stylistics but I doubt that users care about this.

adrianinsaval:
> We can debate about stylistics but I doubt that users care about this.
I personally don't care much for this one in particular, but I would like to point out that users definitively do care about the application's style. @luzpaz I think your best bet would be to make a PR yourself if you can, one that does not require many additional translations is preferred IMO not just because it's more work for the translators but also because they might translate the toolbar name differently in the tooltip rendering the tooltip useless as it won't correspond to the checkboxes to hide them

mwganson:
It could be as simple (I haven't tried it) as embedding html into the tooltip. [<b> + toolBarName + </b>] or leave out the brackets if you don't want them.  I don't have strong feelings either way.  I do think it needs to be easily distinguishable from the tooltips for the various commands.

PR #5262, opened 2021-12-16
[PATH]  Bug/translation

sliptonic:
This is a a fix for the issue described here: https://forum.freecadweb.org/viewtopic.php?f=15&t=64336

This is a limited implementation of several translation fixes and just for the profile op/gui.  The same changes will need to be applied across much of Path.

The specific things addressed here:
translation uses QtCore.QT_TRANSLATE_NOOP
Combobox enums are defined in the operation code

Suggestion for how to regularize translation across path are very welcome.

---

### DIFF
--- a/src/Mod/Path/PathScripts/PathOpGui.py
+++ b/src/Mod/Path/PathScripts/PathOpGui.py
@@ -22,7 +22,7 @@
 
 import FreeCAD
 import FreeCADGui
-import PathGui as PGui # ensure Path/Gui/Resources are loaded
+import PathGui as PGui  # ensure Path/Gui/Resources are loaded
 import PathScripts.PathGeom as PathGeom
 import PathScripts.PathGetPoint as PathGetPoint
 import PathScripts.PathGui as PathGui
@@ -52,11 +52,11 @@ def translate(context, text, disambig=None):
 
 
 class ViewProvider(object):
-    '''Generic view provider for path objects.
+    """Generic view provider for path objects.
     Deducts the icon name from operation name, brings up the TaskPanel
     with pages corresponding to the operation's opFeatures() and forwards
     property change notifications to the page controllers.
-    '''
+    """
 
     def __init__(self, vobj, resources):
         PathLog.track()
@@ -79,12 +79,12 @@ class ViewProvider(object):
         return
 
     def deleteObjectsOnReject(self):
-        '''deleteObjectsOnReject() ... return true if all objects should
+        """deleteObjectsOnReject() ... return true if all objects should
         be created if the user hits cancel. This is used during the initial
         edit session, if the user does not press OK, it is assumed they've
-        changed their mind about creating the operation.'''
+        changed their mind about creating the operation."""
         PathLog.track()
-        return hasattr(self, 'deleteOnReject') and self.deleteOnReject
+        return hasattr(self, "deleteOnReject") and self.deleteOnReject
 
     def setDeleteObjectsOnReject(self, state=False):
         PathLog.track()
@@ -92,7 +92,7 @@ class ViewProvider(object):
         return self.deleteOnReject
 
     def setEdit(self, vobj=None, mode=0):
-        '''setEdit(vobj, mode=0) ... initiate editing of receivers model.'''
+        """setEdit(vobj, mode=0) ... initiate editing of receivers model."""
         PathLog.track()
         if 0 == mode:
             if vobj is None:
@@ -101,14 +101,16 @@ class ViewProvider(object):
             page.setTitle(self.OpName)
             page.setIcon(self.OpIcon)
             selection = self.getSelectionFactory()
-            self.setupTaskPanel(TaskPanel(vobj.Object, self.deleteObjectsOnReject(), page, selection))
+            self.setupTaskPanel(
+                TaskPanel(vobj.Object, self.deleteObjectsOnReject(), page, selection)
+            )
             self.deleteOnReject = False
             return True
         # no other editing possible
         return False
 
     def setupTaskPanel(self, panel):
-        '''setupTaskPanel(panel) ... internal function to start the editor.'''
+        """setupTaskPanel(panel) ... internal function to start the editor."""
         self.panel = panel
         FreeCADGui.Control.closeDialog()
         FreeCADGui.Control.showDialog(panel)
@@ -120,7 +122,7 @@ class ViewProvider(object):
             PathLog.info("did not find no job")
 
     def clearTaskPanel(self):
-        '''clearTaskPanel() ... internal callback function when editing has finished.'''
+        """clearTaskPanel() ... internal callback function when editing has finished."""
         self.panel = None
         job = self.Object.Proxy.getJob(self.Object)
         if job:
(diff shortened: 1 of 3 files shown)


src/Mod/Path/PathScripts/PathOpGui.py
```diff
@@ -184,182 +190,225 @@ def setupContextMenu(self, vobj, menu):
...
+        """selectInComboBox(name, combo) ...
+        helper function to select a specific value in a combo box."""
+        combo.blockSignals(True)
+        cnt = combo.count()
+        index = combo.currentIndex()  # Save initial index
+
+        # Search using currentData and return if found
+        while cnt > 0:
+            cnt -= 1
+            combo.setCurrentIndex(cnt)
```
chennes, line 380: This is going to emit `currentIndexChanged()` every time it's called. I'd probably use `findData()` instead.

src/Mod/Path/PathScripts/PathOpGui.py
```diff
@@ -184,182 +190,225 @@ def setupContextMenu(self, vobj, menu):
...
+    #         combo.blockSignals(False)
+
     def selectInComboBox(self, name, combo):
-        '''selectInComboBox(name, combo) ...
-        helper function to select a specific value in a combo box.'''
-        index = combo.findText(name, QtCore.Qt.MatchFixedString)
-        if index >= 0:
+        """selectInComboBox(name, combo) ...
+        helper function to select a specific value in a combo box."""
+        combo.blockSignals(True)
```
chennes, line 373: Consider using `QSignalBlocker` instead, so you don't have to have all these block/unblock calls.

src/Mod/Path/PathScripts/PathProfile.py
```diff
@@ -56,8 +56,8 @@
 
 
 # Qt translation handling
-def translate(context, text, disambig=None):
-    return QtCore.QCoreApplication.translate(context, text, disambig)
+def tr(text):
+    return QtCore.QT_TRANSLATE_NOOP("PathProfile", text)
```
iromero91, line 60: I don't know if this is correct. Wouldn't this just pass through the text untranslated?

iromero91: Looked into it further, this will not play well with `lupdate`. it does static analysis only, so the function HAS to be called QT_TRANSLATE_NOOP and include the context as an argument, otherwise it will not get extracted with the correct context. the good news is that you dont have to fully qualify it, so you could just bring it into the namespace at the top of the file like so:

```python
from PySide.QtCore import QT_TRANSLATE_NOOP
```

Another shorthand possibility is to use that function once at the top of the file to assign the string to a variable and then use that variable in the body of the code, specially if the string is repeated many times.

I am still not sure how the GUI knows which context to use, but it evidently works on other workbenches.

iromero91: Another shorthand option is if the build system is set up to add an alias when running lupdate so the function can be called with a shorter name like `-tr-function-alias QT_TRANSLATE_NOOP+=_TN` and then in the code
```python
from PySide.QtCore import QT_TRANSLATE_NOOP as _TN
```
or somesuch

chennes: @iromero91 is correct -- `lupdate` is basically just a simple text parser. Also, I thought our discussion was that you needed to use `QtCore.QCoreApplication.translate()` here: you are trying to get the actual translated string here, right? Not just mark the text for translation with a later call to `translate()`?

iromero91: Nevermind that, the python version of lupdate does not seem to have aliases in qt5. In any case i confirmed that these changes make all the translations end up in the `@Default` context

src/Mod/Path/PathScripts/PathOpGui.py
Russ4262: Cool.  I was wondering if there was a better way to parse the user data for each enumeration than the method I used, loading each enumeration and reading its user data value.

We can make this improvement.  Thanks!

src/Mod/Path/PathScripts/PathProfile.py
iromero91: @chennes actually I think we need the nop here as the translate call is done by the freecad gui.

chennes: Not the way it's being used here. First, that won't apply to any of the console output (though I was under the impression we didn't translate console output). Second, take the following code:
```
             "JoinType": [
                 (tr("Round"), "Round"),
                 (tr("Square"), "Square"),
                 (tr("Miter"), "Miter"),
             ],
```
It's critical in this use that the first string of the pair be the translated string in the internal code (not just as displayed in the UI). So this cannot use the NOOP, it must actually query the translation system.

sliptonic: I'm getting more confused by the minute.  The lupdate stuff makes sense, so I'll unwrap those calls.  
Everything else has my head in a twist.   

Realthunder said (in Telegram)  "I notice in existing Path code, some Path object pass translate()'ed string as documentation for properties of newly created object, this means, the those translated string are saved into file, another user in a different language will not be able to read"   Example: PathToolBit.py:140

This makes sense but the alternative is that the documentation is stored in english and untranslated so another user in a different language still can't read them.  


I grepped through everything in /MOD looking for how other workbenches are doing this.  There are a million calls to QT_TRANSLATE_NOOP  and only TEN calls to QCoreApplication.translate and all of those are in two files in the Draft wb.   Is Path the only workbench that's actually trying to do translation in the GUI?

> (though I was under the impression we didn't translate console output). 

Are there written standards for how translation should be handled?

chennes: If you're passing the string into something that's going to get `translate()` called on it later, you use NOOP so that the string gets extracted by `lupdate`, but the untranslated string gets passed around. This is common when dealing with Command, for example. Command itself calls `translate()` as it loops over all the commands puts them into the UI, but since it's working with variables rather than string literals, you have to make sure there is a `QT_TRANSLATE_NOOP` someplace with the literal in it so that the string exists in the translation database when `translate()` goes to find it. Property is similar: your docstrings there should use NOOP because the Property classes call translate() later on when actually showing those strings to the user. Since Command and Property make up like 95% of FreeCAD, that's why you see such prevalence of the NOOP.

In addition, for most of the UI there are functions associated with each dialog called `retranslate()` or `retranslateUi()` (or something along those lines) -- those functions call `translate()` on each widget (though often with a string literal, so it actually doesn't matter what strings you use elsewhere in the code, they just get overwritten!).

In your case you're doing things a bit differently: before this PR you were relying on the text content of the widget to function as an enum. Of course, as soon as translation happens that all breaks. So instead you are now attaching an extra, non-translated bit of data to each widget, so you don't care about the text that's displayed. You can either do the translation yourself when you create the widget, or you can have a `retranslate()` function that handles it later. As long as the user data doesn't get translated, it doesn't matter which approach you take.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5262">pipeline status</a> for feature branch [PR_5262](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5262). Pipeline [431387988](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/431387988) was triggered at [5f00154](https://github.com/FreeCAD/FreeCAD/pull/5262/commits/5f001548a1a6163d1829358e0ee5419aa86c38ad). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5262">pipeline status</a> for feature branch [PR_5262](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5262). Pipeline [432253341](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/432253341) was triggered at [21765de](https://github.com/FreeCAD/FreeCAD/pull/5262/commits/21765dec9cb713390c30ce86fbe04df7c09bec0c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #5267, opened 2021-12-17
Spreadsheet: Realthunder sheetbind stage4

chennes:
This is the penultimate of the merge PRs for @realthunder's PR "Configuration Table using Spreadsheet" (#2862). The main features implemented in this batch are cell binding and improved cut-copy-and-paste UI. Most needed right now are simply testers for those two features on various operating systems. The cell binding mechanism is found in the context menu and will open a dialog box. Copy and paste are accessed through the usual mechanisms.

### DIFF
--- a/src/App/PropertyStandard.cpp
+++ b/src/App/PropertyStandard.cpp
@@ -207,7 +207,7 @@ void PropertyPath::setValue(const char * Path)
     hasSetValue();
 }
 
-boost::filesystem::path PropertyPath::getValue() const
+const boost::filesystem::path &PropertyPath::getValue(void) const
 {
     return _cValue;
 }
--- a/src/App/Range.cpp
+++ b/src/App/Range.cpp
@@ -149,7 +149,7 @@ int App::validRow(const std::string &rowstr)
     char * end;
     int i = strtol(rowstr.c_str(), &end, 10);
 
-    if (i <0 || i >= CellAddress::MAX_ROWS || *end)
+    if (i <=0 || i > CellAddress::MAX_ROWS || *end)
         return -1;
 
     return i - 1;
@@ -234,24 +234,28 @@ App::CellAddress App::stringToAddress(const char * strAddress, bool silent)
   * @returns Address given as a string.
   */
 
-std::string App::CellAddress::toString(bool noAbsolute) const
+std::string App::CellAddress::toString(bool noAbsolute, bool r, bool c) const
 {
     std::stringstream s;
 
-    if(_absCol && !noAbsolute)
-        s << '$';
-    if (col() < 26)
-        s << (char)('A' + col());
-    else {
-        int colnum = col() - 26;
+    if(c) {
+        if(_absCol && !noAbsolute)
+            s << '$';
+        if (col() < 26)
+            s << (char)('A' + col());
+        else {
+            int colnum = col() - 26;
 
-        s << (char)('A' + (colnum / 26));
-        s << (char)('A' + (colnum % 26));
+            s << (char)('A' + (colnum / 26));
+            s << (char)('A' + (colnum % 26));
+        }
     }
 
-    if(_absRow && !noAbsolute)
-        s << '$';
-    s << (row() + 1);
+    if(r) {
+        if(_absRow && !noAbsolute)
+            s << '$';
+        s << (row() + 1);
+    }
 
     return s.str();
 }
--- a/src/Gui/ExpressionCompleter.cpp
+++ b/src/Gui/ExpressionCompleter.cpp
@@ -159,10 +159,10 @@ public:
         int docSize = (int)docs.size()*2;
         int objSize = 0;
         int propSize = 0;
-        std::vector<App::Property*> props;
+        std::vector<std::pair<const char*, App::Property*> > props;
         App::Document *doc = 0;
         App::DocumentObject *obj = 0;
-        App::Property *prop = 0;
+        const char *propName = nullptr;
         if(idx>=0 && idx<docSize)
             doc = docs[idx/2];
         else {
@@ -184,13 +184,13 @@ public:
                     idx -= objSize;
                     if(info.d.doc<0)
                         row = idx;
-                    cobj->getPropertyList(props);
+                    cobj->getPropertyNamedList(props);
                     propSize = (int)props.size();
                     if(idx >= propSize)
                         return;
                     if(idx>=0) {
                         obj = cobj;
-                        prop = props[idx];
+                        propName = props[idx].first;
                     }
                 }
             }
@@ -200,8 +200,8 @@ public:
                 *count = docSize + objSize + propSize;
             if(idx>=0 && v) {
                 QString res;
-                if(prop)
-                    res = QString::fromLatin1(prop->getName());
+                if(propName)
+                    res = QString::fromLatin1(propName);
                 else if(obj) {
                     if(idx & 1)
                         res = QString::fromUtf8(quote(obj->Label.getStrValue()).c_str());
@@ -248,18 +248,18 @@ public:
 
         if(noProperty)
             return;
-        if(!prop) {
+        if(!propName) {
             idx = row;
-            obj->getPropertyList(props);
+            obj->getPropertyNamedList(props);
             propSize = (int)props.size();
             if(idx<0 || idx>=propSize)
                 return;
-            prop = props[idx];
+            propName = props[idx].first;
             if(count)
                 *count = propSize;
         }
-        if(v)
-            *v = QString::fromLatin1(prop->getName());
+        if(v) 
+            *v = QString::fromLatin1(propName);
         return;
     }
 
--- a/src/Gui/ExpressionCompleter.cpp
+++ b/src/Gui/ExpressionCompleter.cpp
@@ -54,22 +54,21 @@ using namespace Gui;
 
 class ExpressionCompleterModel: public QAbstractItemModel {
 public:
-    ExpressionCompleterModel(QObject *parent, const App::DocumentObject *obj, bool noProperty)
+    ExpressionCompleterModel(QObject *parent, bool noProperty)
         :QAbstractItemModel(parent), noProperty(noProperty)
     {
-        setDocumentObject(obj);
     }
 
     void setNoProperty(bool enabled) {
         noProperty = enabled;
     }
 
-    void setDocumentObject(const App::DocumentObject *obj) {
+    void setDocumentObject(const App::DocumentObject *obj, bool checkInList) {
         beginResetModel();
         if(obj) {
             currentDoc = obj->getDocument()->getName();
             currentObj = obj->getNameInDocument();
-            if(!noProperty)
+            if(!noProperty && checkInList)
                 inList = obj->getInListEx(true);
         } else {
             currentDoc.clear();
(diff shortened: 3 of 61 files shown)


donovaly:
> The cell binding mechanism is found in the context menu and will open a dialog box. Copy and paste are accessed through the usual mechanisms.

I volunteer to test but need more info how to test properly. What could fail? Are there special constructs / use cases I should try?

chennes:
If I tell you what to test it will be a test I've already done 😄 -- just create a spreadsheet, put some data in it, and set up some bindings. Copy and paste some stuff. Cut and paste some stuff. Try to come up with creative ways to crash FreeCAD. The usual drill.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5267">pipeline status</a> for feature branch [PR_5267](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5267). Pipeline [432253535](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/432253535) was triggered at [68f46dd](https://github.com/FreeCAD/FreeCAD/pull/5267/commits/68f46dd834c606d4e1bc7168d845a40867aa375d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> Copy and paste some stuff. Cut and paste some stuff. 

The point is that I feel like a monkey looking at a gearbox - absolutely no clue twhat the feature is about. For example, I right-click on cell C6, the I get this dialog:

FreeCAD_odtkcY6oXL

This is meaning less for me. I see there 4 cells, but twice the same. But in fact I only selected C6, why is there C1? Why C1 and not B6 or A1?
When I do nothing in the dialog, but click OK, I see that cell C6 gets a blue frame. But why was in the dialog twice cell C1? And why did I not get a frame around all cells in the range C1 - C6?
Why can't I just select ranges as I can in LibreOffice, Excel and the like?

Then what is bound to the cells?

Then I don't understand how the selection of the spreadsheet works. So obviously one can bind something (but what) to another spreadsheet. But what to what cell and sheet?

I am completely lost and don't even have an idea what the feature is meant for. Therefore I need pointers. Sorry.

----------

I get a blur frame, but where can I change the color in case I don't like blue or I use already blue background and thus can't see the frame properly?

donovaly:
> I get a blue frame, but where can I change the color in case I don't like blue or I use already blue background and thus can't see the frame properly?

This is indeed an issue:
grafik
and
FreeCAD_9gJvO99KlP

The binding is identical, just the background is different.

donovaly:
When I set accidentally an expression which id not a string of a cell name, I get no warning. The dialog just lets me enter any property and clicking OK in the dialog:

FreeCAD_7OqNO1L9OO

Most of our other dialogs check for mistakes ideally don't let you enter wrong things. If not, then they try to either don't let you click OK or at least highlight the field with the mistake and set e.g. a tooltip of what is wrong.

The point is that even simple mistakes like a lowercase letter is not caught:
FreeCAD_ZHYEsSOfrX

This is problematic because I have in TechDraw a table view and because of the lowercase letter TD lets me add a table but it won't be shown. First when i correct my mistake, I get the table shown in the TD page. This should be fixed because as normal user you end up quickly to have no clue what is going on, that a lowercase letter causes your drawing page to break.

donovaly:
Next issue: I might have accidentally added a letter too much so that the cell range is exceeded,
- I get no info that there is a cell range, meaning cell ZZ4 exists, but not ZZZ4:

FreeCAD_grCJR7MmD5
(so one has 16k rows but only 52 columns?)

- when I enter e.g. "ZZZ4", I get in the status bar the info "Recompute failed! Please check report view." But when I do so, I see there
`Recompute failed! Please check report view.`
So report view tells me that I should look at report view ;-)

chennes:
> I am completely lost and don't even have an idea what the feature is meant for. Therefore I need pointers. Sorry.
No need to apologize: the reason I didn't give pointers was because this is exactly the kind of feedback that's needed. I happen to agree with your assessment of the interface, but didn't want to color anyone's testing by being heavy-handed with instructions.

donovaly:
> No need to apologize: the reason I didn't give pointers was because this is exactly the kind of feedback that's needed. I happen to agree with your assessment of the interface, but didn't want to color anyone's testing by being heavy-handed with instructions.

But how to proceed then? Will @realthunder have a look at my posts?

Besides the UI things I reported, I need to understand the feature itself. Is there a Wiki page or a forum post explaining the intention of the bind feature?

chennes:
There is a forums discussion about the original PR: https://forum.freecadweb.org/viewtopic.php?f=17&t=42183

chennes:
Regarding how to proceed, we have a couple of options:

1. @realthunder can add another commit revising the UI himself
2. We can merge this as-is and someone else can redesign the UI at a later date
3. We can modify this PR to hide the dialog entirely for the time being: the important part of this PR is really the under-the-hood work in preparation for the "real" UI, the configuration table stuff.
4. Someone else can redesign the Bind UI right now and include it as part of this PR. Now accepting volunteers 😉 ...

donovaly:
> 2. We can merge this as-is and someone else can redesign the UI at a later date

I opt for this option. However, it would be nice if @realthunder could have a look at the UI once the complete spreadsheet feature is in. I think for him it is much easier than for me to implement the checks to prevent invalid inputs than for me who first would need to read into the code.

realthunder:
The `cells binding` function is a meant to bind a range of cells to another range. You are meant to select two range of cells and then click the dialog. I think this is probably a low-level function that not many user will use. It is used to realize the higher level function of the `Configuration table`, so that changing one cell can change an entire row of cells.

I'll see what I can do to make the binding dialog more user friendly.

chennes:
> I'll see what I can do to make the binding dialog more user friendly.

I wouldn't worry about it for now, the config table stuff is more important and I don't want to hold this PR up. Maybe just add it to your "future items" to-do list...

donovaly:
> The `cells binding` function is a meant to bind a range of cells to another range.

Thanks. Now I get it.

My feedback should not holding back this PR. I only would like that once all the Spreadsheet features are fully merged, you have a look and change the dialogs so that they check for invalid input. Typos happen all the time unwantedly and the users should not be lost just because of a typo.

@chennes fine with me to merge since copy and pasting around worked for me so far.

@realthunder, many thanks for all your great feature additions.

luzpaz:
Heads up to the packagers, can we generate a new build for testers to test this latest merge?  
@looooo (AppImage weekly-build)  
@ppd (Snap edge)  
@hfiguiere (Flatpak, BTW is there a nightly flatpak?)

src/App/Range.cpp
```diff
@@ -234,24 +234,28 @@ App::CellAddress App::stringToAddress(const char * strAddress, bool silent)
   * @returns Address given as a string.
   */
 
-std::string App::CellAddress::toString(bool noAbsolute) const
+std::string App::CellAddress::toString(bool noAbsolute, bool r, bool c) const
```
wwmayer, line 237: @chennes 
This is a code smell called _boolean blindness_

For a reviewer or programmer it's not obvious what the arguments are supposed to do and one has to go to the declaration or documentation. Since it's not even documented one has to dive into the code to get a rough idea.

Therefore I suggest to replace the three boolean arguments with an enum class that supports bit-wise OR. Then in client code some speaking identifiers must be used that increases readability a lot.

ppd:
@luzpaz The snap picks up the latest commit every night a UTC 00:00. I still kicked off a build: https://github.com/ppd/freecad-ppd/actions/runs/1611311277
But generally, it's not necessary to do so as the next build is often just a few hours away.

src/Mod/Spreadsheet/App/PropertySheetPyImp.cpp
```diff
@@ -23,14 +23,31 @@
...
-#include "PropertySheet.h"
+#include "Mod/Spreadsheet/App/PropertySheet.h"
 
 // inclusion of the generated files (generated out of PropertySheetPy.xml)
 #include "PropertySheetPy.h"
 #include "PropertySheetPy.cpp"
 
 using namespace Spreadsheet;
 
+struct PropertySheetPyInit {
```
wwmayer, line 34: We do not need this hack to support Python's mapping protocol. We can properly define it in PropertySheetPy.xml and then implement the function mapping_subscript. As an example VectorPy.xml can be used.

src/Gui/ExpressionCompleter.cpp
```diff
@@ -564,28 +557,33 @@ void ExpressionCompleter::slotUpdate(const QString & prefix, int pos)
     }
 }
 
-ExpressionLineEdit::ExpressionLineEdit(QWidget *parent, bool noProperty, bool requireLeadingEqualSign)
+ExpressionLineEdit::ExpressionLineEdit(QWidget *parent, bool noProperty, char checkPrefix, bool checkInList)
```
wwmayer, line 560: If should be checked if the two boolean noProperty and checkInList should be replaced with an enum class.

src/App/PropertyStandard.cpp
```diff
@@ -207,7 +207,7 @@ void PropertyPath::setValue(const char * Path)
     hasSetValue();
 }
 
-boost::filesystem::path PropertyPath::getValue() const
+const boost::filesystem::path &PropertyPath::getValue(void) const
```
wwmayer, line 210: Adding void to the argument is considered deprecated.

hfiguiere:
> Heads up to the packagers, can we generate a new build for testers to test this latest merge?
>
>     * [ ]  @hfiguiere (Flatpak, BTW is there a nightly flatpak?)

Flathub doesn't do nightlies by policy. However it something that could be setup on CI with Github actions (just have never done it)
In the past I have build pre-releases on beta with the same revisions as the pre-release builds.

hfiguiere:
I have spun a build for testing. See https://github.com/flathub/org.freecadweb.FreeCAD/pull/52

Not sure how to test this properly, but at least it does build.

src/Mod/PartDesign/App/ShapeBinder.cpp
```diff
@@ -612,13 +747,23 @@ void SubShapeBinder::slotRecomputedObject(const App::DocumentObject& Obj) {
...
-        update();
+        try {
+            update();
+        } catch (Base::Exception &e) {
+            e.ReportException();
+        }
     }
 }
 
+static const char _GroupPrefix[] = "Configuration (";
```
wwmayer, line 758: This causes a compiler warning: unused variable _GroupPrefix [-Wunused-const-variable]
This is possibly a copy&paste issue since it's also declared in Link.cpp

src/App/Range.cpp
wwmayer: See https://github.com/FreeCAD/FreeCAD/pull/5360

chennes: That looks excellent, thanks for adding that enum bitmask metaprogramming.

wwmayer: It's basically the same as Qt's QFlags class but I didn't want to pull-in a Qt dependency on this class. And the bitmask is something that we may need more often for non-Qt classes.

If you have a minute you can try to compile it on Windows, otherwise I will do it the next time I am on Windows. If MSVC accepts it the PR can be merged.

chennes: Compiling now...

PR #5269, opened 2021-12-18
[PartDesign] for chamfer and fillet, add SelectAllEdges option in the…

mwganson:
… context menu while in add or remove mode
Snip macro screenshot-41f19f

If not in add or remove mode the shortcut is disabled.  When it is selected all the edges of the base object are selected, emulating what would happen if the user selected them all manually.

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
@@ -96,6 +96,9 @@ TaskChamferParameters::TaskChamferParameters(ViewProviderDressUp *DressUpView, Q
     createDeleteAction(ui->listWidgetReferences, ui->buttonRefRemove);
     connect(deleteAction, SIGNAL(triggered()), this, SLOT(onRefDeleted()));
 
+    createAddAllEdgesAction(ui->listWidgetReferences);
+    connect(addAllEdgesAction, &QAction::triggered, this, &TaskChamferParameters::onAddAllEdges);
+
     connect(ui->listWidgetReferences, SIGNAL(currentItemChanged(QListWidgetItem*, QListWidgetItem*)),
         this, SLOT(setSelection(QListWidgetItem*)));
     connect(ui->listWidgetReferences, SIGNAL(itemClicked(QListWidgetItem*)),
@@ -244,6 +247,12 @@ void TaskChamferParameters::onRefDeleted(void)
         ui->buttonRefRemove->setEnabled(false);
         ui->buttonRefRemove->setToolTip(tr("There must be at least one item"));
     }
+}
+
+void TaskChamferParameters::onAddAllEdges(void)
+{
+    TaskDressUpParameters::addAllEdges(ui->listWidgetReferences);
+    ui->buttonRefRemove->setEnabled(true);
 }
 
 void TaskChamferParameters::onTypeChanged(int index)
--- a/src/Mod/PartDesign/Gui/TaskDressUpParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskDressUpParameters.cpp
@@ -65,6 +65,7 @@ TaskDressUpParameters::TaskDressUpParameters(ViewProviderDressUp *DressUpView, b
     , proxy(0)
     , DressUpView(DressUpView)
     , deleteAction(nullptr)
+    , addAllEdgesAction(nullptr)
     , allowFaces(selectFaces)
     , allowEdges(selectEdges)
 {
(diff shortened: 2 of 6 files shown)


mwganson:
https://forum.freecadweb.org/viewtopic.php?f=17&t=64611

src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
```diff
@@ -96,6 +97,9 @@ TaskChamferParameters::TaskChamferParameters(ViewProviderDressUp *DressUpView, Q
     createDeleteAction(ui->listWidgetReferences, ui->buttonRefRemove);
     connect(deleteAction, SIGNAL(triggered()), this, SLOT(onRefDeleted()));
 
+    createSelectAllEdgesAction(ui->listWidgetReferences);
+    connect(selectAllEdgesAction, SIGNAL(triggered()), this, SLOT(onSelectAllEdges()));
```
wwmayer, line 101: In new code the overloaded version of connect() that uses the SIGNAL and SLOT macros should be avoided. Instead use the version where you function pointers are used.
`
connect(selectAllEdgesAction, &QAction::triggered, this, &TaskChamferParameters::onSelectAllEdges);
`
The advantage is that in case a signal or slot function isn't defined the compiler already raises an error.

src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
```diff
@@ -194,6 +198,31 @@ void TaskChamferParameters::clearButtons(const selectionModes notThis)
     DressUpView->highlightReferences(false);
 }
 
+// By adding the edges to the Gui.Selection in python, then onSelectionChanged() gets called
+// where the edges get added to the list in the task dialog and highlighted in the 3d view
+
+void TaskChamferParameters::onSelectAllEdges(void)
```
wwmayer, line 204: TaskChamferParameters::onSelectAllEdges and TaskFilletParameters::onSelectAllEdges are almost identical. The only difference is the different cast but because only the Base property is accessed that belongs to PartDesign::DressUp the whole function can be moved to the base class TaskDressUpParameters. This will avoid code duplication.

src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
```diff
@@ -194,6 +198,31 @@ void TaskChamferParameters::clearButtons(const selectionModes notThis)
...
+{
+    PartDesign::Chamfer* pcChamfer = static_cast<PartDesign::Chamfer*>(DressUpView->getObject());
+    if (pcChamfer) {
+        App::DocumentObject* base = pcChamfer->Base.getValue();
+        if (base) {
+            std::string baseName = base->getNameInDocument();
+            std::ostringstream command;
+            command << "from PySide import QtGui,QtCore\n";
+            command << "QtGui.QApplication.setOverrideCursor(QtCore.Qt.WaitCursor)\n";
```
wwmayer, line 213: I wonder whether it's a good idea to put the override cursor into the macro code. In case an exception is raised before restoring the old cursor the interpreter jumps out of the function and will never restore the old cursor.

Actually it should be sufficient to only show the wait cursor if the dialog is opened and thus it can be done in C++ by using the class Gui::WaitCursor.

src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
```diff
@@ -194,6 +198,31 @@ void TaskChamferParameters::clearButtons(const selectionModes notThis)
...
+            command << "for _ii in range(0, len(App.ActiveDocument.getObject(\'" << baseName << "\').Shape.Edges)):\n";
```
wwmayer, line 214: For the recorded Python code do not use App.ActiveDocument because the user can have opened several documents where a document is activated that doesn't belong to the opened dialog. Thus, using App.ActiveDocument will be executed for the wrong document. Instead use DocumentObjectT::getObjectPython() or one of the helper functions defined inside CommandT.h.

src/Mod/PartDesign/Gui/TaskChamferParameters.cpp
```diff
@@ -194,6 +198,31 @@ void TaskChamferParameters::clearButtons(const selectionModes notThis)
...
+            command << "    Gui.Selection.addSelection(App.ActiveDocument.getObject(\'" << baseName << "\'), \'Edge\'+str(_ii+1))\n";
+            command << "    if _ii % 20:\n";
```
wwmayer, line 216: isn't it extremely slow to add each edge separately? Why not directly updating the QListWidget without the detour of the selection?

src/Mod/PartDesign/Gui/TaskDressUpParameters.cpp
```diff
@@ -232,9 +237,28 @@ void TaskDressUpParameters::itemClickedTimeout() {
     wasDoubleClicked = false;
 }
 
+void TaskDressUpParameters::createSelectAllEdgesAction(QListWidget* parentList)
+{
+    // creates a context menu, a shortcut for it and connects it to e slot function
+
+    selectAllEdgesAction = new QAction(tr("Select all edges in the 3D view"), this);
+    selectAllEdgesAction->setShortcut(tr("Ctrl+Shift+A"));
```
wwmayer, line 245: Do not use tr("Ctrl+Shift+A") because there is the risk that it won't be correctly translated and thus might break the shortcut. Instead use QKeySequence(QString::fromLatin1("Ctrl+Shift+A")) since Qt already provides the correct translation.

src/Mod/PartDesign/Gui/TaskDressUpParameters.cpp
```diff
@@ -263,6 +287,10 @@ bool TaskDressUpParameters::KeyEvent(QEvent *e)
                 kevent->accept();
                 return true;
             }
+            if (selectAllEdgesAction && kevent->key() == (Qt::Key_Shift | Qt::Key_Control | Qt::Key_A)){
```
wwmayer, line 290: A bit-wise OR of different enum types is deprecated with C++20. Instead create a QKeySequence object as above and compare it with the kevent pointer. Qt provides a suitable comparison operator for this.

mwganson:
Thanks for the feedback and for providing the reasoning for each point.

mwganson:
I think I addressed all of your concerns.  If I missed anything it was by mistake.  This is a much faster way instead of using the selection.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5269">pipeline status</a> for feature branch [PR_5269](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5269). Pipeline [432835523](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/432835523) was triggered at [60e3770](https://github.com/FreeCAD/FreeCAD/pull/5269/commits/60e37709b228c9c38e8a0ad4c43d500064dff3b3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
Looks very good now. Thanks!

marioalexis84:
Crash reported in the forum: https://forum.freecadweb.org/viewtopic.php?f=19&t=65335

wwmayer:
Fixed with 2d1c083d9a

PR #5271, opened 2021-12-19
[PATH] change discretize value to fix issues

sundtek:
Discretize of 0.0001 is okay for inches, but for metric <1microns would be quite small, this fixes some issues when generating paths for curves that look inward.
Since the internal unit is mm 0.0001 must have been set when having inch in mind.

I found following post in the forum: https://forum.freecadweb.org/viewtopic.php?f=15&t=42755
I have experienced similar issues with other curves in my object.
This is a sample where I use bsplines for creating extensions to a face (I also had that issue when generating adaptive paths - with the updated accuracy it does not happen anymore):
https://streamable.com/jfbin6

---

### DIFF
--- a/src/Mod/Path/PathScripts/PathAdaptive.py
+++ b/src/Mod/Path/PathScripts/PathAdaptive.py
@@ -33,6 +33,7 @@ import json
 import math
 import area
 
+from FreeCAD import Units
 from PySide import QtCore
 
 # lazily loaded modules
@@ -104,7 +105,13 @@ def sceneClean():
 
 
 def discretize(edge, flipDirection=False):
-    pts = edge.discretize(Deflection=0.0001)
+    val=Units.Quantity("1mm").getUserPreferred()
+    if len(val)==3 and val[2]=='mm':
+        deflection=0.002
+    else:
+        deflection=0.0001
+
+    pts = edge.discretize(Deflection=deflection)
     if flipDirection:
         pts.reverse()
 
--- a/src/Mod/Path/PathScripts/PathAdaptive.py
+++ b/src/Mod/Path/PathScripts/PathAdaptive.py
@@ -33,7 +33,6 @@ import json
 import math
 import area
 
-from FreeCAD import Units
 from PySide import QtCore
 
 # lazily loaded modules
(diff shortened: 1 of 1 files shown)


sliptonic:
The original implementation has some logic for determining a Deflection value.  The logic was trivial but it selected between two different values. That logic may have been flawed but the developer included it for a reason.  Unfortunately, the logic was buried within a larger function and had no unit test corresponding to his expected use-case.

This PR removes that logic and replaces it with a hard-coded value.  Now the function will perform as you expect but may break for use-cases he had in mind that you are not considering.

This isn't a criticism of your programming ability, it's a plea for more testable code.  The better solution would be to factor the Discretization code out to a single function.  Then write a unit test that returns the correct value under the original and new cases.  Then, in the future, when this code is modified again to accommodate another use case, it will not break for the cases you're considering now.  The unit tests will exist to prevent that.

sundtek:
I changed it from hardcoded (0.0001) to some logic (detecting metric vs inches) back to hard-coded (0.002).
The logic was due to the missing knowledge that the internal scale is always metric.

Russ4262:
I understand this is merged, but if the issue is in the discretizing of an edge, and there seems to be only one crazy, irregular, incorrect, lateral path generated, perhaps the error is further upstream, say in the discretize function itself.  Do we have any tests for the discretize function?  Perhaps the discretize function is failing with these very small values?  Did we attempt the Adaptive with smaller values than the one listed?  Did we attempt with something other than a factor of 10, say 0.00000234? That I know of, no one has found the source of this path generation irregularity within the Adaptive source code.  Do we know what is tripping up the Adaptive algorithm with these very small values?

I am not convinced the true error is in the discretize function for the simple fact that if you take that model that has the bad path and increase the `stock to leave` value of the operation to a small amount, say 0.001, then the irregularity disappears (for all the tests I have run).  So, this leads me to think that the Adaptive is applying an offset equal to the `stock to leave` value first, then discretizing that output, or simply using the offset wire directly.  If using the offset wire directly, what is that discretize factor set to internally?

If we are changing this discretize value, we need to document this minimum value, and perhaps that it is hard-coded, in the wiki for technical purposes.  As we know, the Adaptive op is widely used.

Thank you,
Russell

Russ4262:
If this is the fix we are going with, I suggest adding this to the 0.19.4 backport list.

sundtek:
In case someone wants to investigate the issue, I have added a face to the forum where the problem happens with 0.0001.
https://forum.freecadweb.org/viewtopic.php?f=15&t=42755&p=554736#p554736
oclbug.FCStd

I'm just fine with the 0.002 tolerance, possibly the other discretize calls should also be evaluated as Russel mentioned. I only did a grep through the code and saw a few places but I did not have a closer look at it.

PR #5276, opened 2021-12-22
[PATH] Fix adaptive glitch by filtering nearby values

sundtek:
more details about it are in the forum:
https://forum.freecadweb.org/viewtopic.php?f=15&t=42755&start=10#p555566

---

### DIFF
--- a/src/Mod/Path/libarea/Adaptive.cpp
+++ b/src/Mod/Path/libarea/Adaptive.cpp
@@ -146,6 +146,38 @@ inline DoublePoint GetPathDirectionV(const Path &pth, size_t pointIndex)
 	const IntPoint &p2 = pth.at(pointIndex);
 	return DirectionV(p1, p2);
 }
+	
+bool doFilter(Path &ppg) {
+        size_t i;
+        size_t b;
+        size_t start;
+        size_t end;
+        for(i=0;i<ppg.size();i++) {
+            if (i==ppg.size()-1) {
+                start=0;
+                end=ppg.size()-2;
+            } else {
+                start=i+1;
+                end=ppg.size();
+            }
+            for(b=start;b<end;b++) {
+                if (abs(ppg[i].X-ppg[b].X)<2 && abs(ppg[i].Y-ppg[b].Y)<2) {
+                    ppg.erase(ppg.begin()+b);
+                    return true;
+                }
+            }
+        }
+        return false;
+}
+
+void filterCloseValues(Paths &ppg) {
+        size_t i;
+        for(i=0;i<ppg.size();i++) {
+            while(doFilter(ppg[i]));
+        }
+}
+
+	
 
 //*****************************************
 // Utils
@@ -1654,9 +1686,11 @@ void Adaptive2d::ApplyStockToLeave(Paths &inputPaths)
 		clipof.Clear();
 		clipof.AddPaths(inputPaths, JoinType::jtRound, EndType::etClosedPolygon);
 		clipof.Execute(inputPaths, -1);
+		filterCloseValues(inputPaths);
 		clipof.Clear();
 		clipof.AddPaths(inputPaths, JoinType::jtRound, EndType::etClosedPolygon);
 		clipof.Execute(inputPaths, 1);
+		filterCloseValues(inputPaths);
 	}
 }
 


src/Mod/Path/libarea/Adaptive.cpp
```diff
@@ -146,6 +146,38 @@ inline DoublePoint GetPathDirectionV(const Path &pth, size_t pointIndex)
...
+                    return true;
+                }
+            }
+        }
+        return false;
+}
+
+void filterCloseValues(Paths &ppg) {
+        size_t i;
+        for(i=0;i<ppg.size();i++) {
```
chennes, line 175: In new code it's best to use a range-based for loop.

src/Mod/Path/libarea/Adaptive.cpp
```diff
@@ -146,6 +146,38 @@ inline DoublePoint GetPathDirectionV(const Path &pth, size_t pointIndex)
 	const IntPoint &p2 = pth.at(pointIndex);
 	return DirectionV(p1, p2);
 }
+	
+bool doFilter(Path &ppg) {
```
chennes, line 150: I suggest using a different variable name, since this one below refers to a vector of paths, rather than a single path.

src/Mod/Path/libarea/Adaptive.cpp
```diff
@@ -146,6 +146,38 @@ inline DoublePoint GetPathDirectionV(const Path &pth, size_t pointIndex)
...
+                end=ppg.size();
+            }
+            for(b=start;b<end;b++) {
+                if (abs(ppg[i].X-ppg[b].X)<2 && abs(ppg[i].Y-ppg[b].Y)<2) {
+                    ppg.erase(ppg.begin()+b);
+                    return true;
+                }
+            }
+        }
+        return false;
```
chennes, line 170: There are a lot of C-style array accesses here, with the attendant overrun dangers. Your intent would be expressed more clearly using std::find_if().

src/Mod/Path/libarea/Adaptive.cpp
```diff
@@ -146,6 +146,38 @@ inline DoublePoint GetPathDirectionV(const Path &pth, size_t pointIndex)
...
+        for(i=0;i<ppg.size();i++) {
+            if (i==ppg.size()-1) {
+                start=0;
+                end=ppg.size()-2;
+            } else {
+                start=i+1;
+                end=ppg.size();
+            }
+            for(b=start;b<end;b++) {
+                if (abs(ppg[i].X-ppg[b].X)<2 && abs(ppg[i].Y-ppg[b].Y)<2) {
```
chennes, line 164: The 2 seems arbitrary here: through what mechanism is "2" defined as "close"?

chennes (commented):
Overall this code block is very C-heavy, and requires significant effort to evaluate for bounds safety, etc. I suggest refactoring to use modern C++ constructs.

src/Mod/Path/libarea/Adaptive.cpp
sundtek: if the nearby coordinate is the same or only differs by 1 unit just remove it, the vertexes are upscaled to a relatively high number (clipper uses integer only) so they can just be removed. I'm using Path Adaptive a lot and I was also able to trigger this problem easily as well.

src/Mod/Path/libarea/Adaptive.cpp
sundtek: If you have a better idea here please propose

                if (abs(ppg[i].X-ppg[b].X)<2 && abs(ppg[i].Y-ppg[b].Y)<2) {
                    ppg.erase(ppg.begin()+b);
                    return true;
                }

actually I think there could be an else / break as well because only direct nearby values in a row should matter if a similar value occurs at another position it should be left untouched, also I think I'd rather use modulo to go over the array.

I have never used std::find_if, I think it's too generic for that isn't it? (it might possibly describe the current filter, but it might have an issue with what I just wrote in the previous block).

src/Mod/Path/libarea/Adaptive.cpp
sundtek: ok

src/Mod/Path/libarea/Adaptive.cpp
sundtek: hmm I need that position information and also the next position index; how would you do that?

src/Mod/Path/libarea/Adaptive.cpp
chennes: I suggest something like (untested!):
```
bool doFilter(Path& p) {
	for (auto start = p.begin(); start != p.end(); ++start) {
		auto IsClose = [start](const IntPoint& testPoint) -> bool {
			return (abs(start->X - testPoint.X) < 2 && abs(start->Y - testPoint.Y) < 2);
		};
		auto closePoint = std::find_if(start, p.end(), IsClose);
		if (closePoint != p.end()) {
			p.erase(closePoint); // In C++20 we should replace with erase_if
			return true;
		}
	}
	return false;
}
```

sundtek: 1. you will have to loop over to check against the first item as well, that's why I'd rather use modulo for the iteration actually.
2. this would remove any nearby value within the entire array (yes the same way as my original code does, but I think there's no need to)
eg. X 0,1,1,2,1,1,1,3
this is very simplified because the code pairs x and y; but the code will remove all 1 from the list (I think about edges that loop back to a nearby position like eg. 8 where the center is not touching but very close)
only 0,x,x,2,1,x,x,3 should be left; don't remove the 1 after 2

chennes: Why would you have to loop over? If a is close to b, then b is close to a. No need to double-check every value.

sundtek: the last vertex needs to be checked against the first, second, etc. one as well. values that differ >1 are boundaries and shouldn't be crossed for removing (as with your code suggestion, and as with my patch -- you can consider both as incomplete)

chennes: The last vertex is checked against the first: when you check the first. This solution is sounding increasingly ad hoc, perhaps it's best to step back and rethink the goals.

sundtek: I will check within the next days, I still have to finish some other projects first.
I'll patch it into my local FreeCAD version and run my object through the code to see if things are okay.

I think I was a bit confused when I wrote that (I'm into a completely different topic here which requires my attention..)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5276">pipeline status</a> for feature branch [PR_5276](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5276). Pipeline [439262766](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/439262766) was triggered at [276bf84](https://github.com/FreeCAD/FreeCAD/pull/5276/commits/276bf8439f4a60e6adf3b77b4f9bf2b2402cb1b9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
Please review and address any outstanding concerns on this PR or close it.

sundtek:
maybe someone else can pick it up. My time is too limited ...

PR #5281, opened 2021-12-23
Spreadsheet: Fix trouble with clang and MacOS

ageeye:
Since the Spreadsheet: support cell binding commit, I got trouble by building freecad. I didn't find similar reports in install/compile forum. So I will only use the old version of typedef for macos. Better solutions are welcome.

See [trouble details](https://github.com/FreeCAD/FreeCAD/commit/68fca409833ca0c8c3811ad4efddcfcc3e2e4255#commitcomment-62364932)

### DIFF
--- a/src/App/PropertyExpressionEngine.h
+++ b/src/App/PropertyExpressionEngine.h
@@ -178,7 +178,7 @@ private:
     typedef std::pair<int, int> Edge;
     // Note: use std::map instead of unordered_map to keep the binding order stable
     #ifdef FC_OS_MACOSX
-    typedef boost::unordered_map<const App::ObjectIdentifier, ExpressionInfo> ExpressionMap;
+    typedef std::map<App::ObjectIdentifier, ExpressionInfo> ExpressionMap;
     #else
     typedef std::map<const App::ObjectIdentifier, ExpressionInfo> ExpressionMap;
     #endif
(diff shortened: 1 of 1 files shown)


donovaly:
@realthunder , could you please review this PR?

realthunder:
@ageeye Can you please try to change ` std::map<const App::ObjectIdentifier, ExpressionInfo>` to ` typedef std::map<App::ObjectIdentifier, ExpressionInfo>` and see if it can compile?

ageeye:
@realthunder , without const is the compiler happy. Should it general use or only for FC_OS_MACOSX? 

Thank you for the support.

realthunder:
It is actually not common or necessary to define map key as const. I did this only because the original `unordered_map` is declared as such. Nevertheless, the compiler should be fine with it. I think this is probably a bug of that particular version of `clang` that you are using.

donovaly:
It seams ready to be merged, right?

@ageeye , can you please rebase and force-push that the PR gets a single clean commit.

jbaehr:
I can confirm that this works on my 10.14 machine with "Apple clang version 11.0.0 (clang-1100.0.33.17)".
However, I think it's a bit unorthodox to remove a `const` just for a specific platform. I fear that opens the door for subsequent compilation failures in the future.

dirkolbrich:
Can confirm the initial error and the fix by this patch on Apple Silicon M1 with `Compiler: AppleClang, version: 13.0.0.13000029` and `-- Standard: Requires C++17` under `MacOSX12.1`.

And I agree with removing a `const` just for a specific platform could lead to compiling issues in the future.

PR #5284, opened 2021-12-23
[GUI] don't show report view on warnings by default

donovaly:
I introduced FreeCAD to several people and see that especially newbies are confused with all the warnings they get but cannot understand.
The solution was to change that the report view panel is not automatically shown on every warning.
This PR does this.

More experienced users can anytime enable the option. Thus this PR has not much impact but obviously makes life easier for beginners.

### DIFF
--- a/src/Gui/DlgReportView.ui
+++ b/src/Gui/DlgReportView.ui
@@ -134,9 +134,6 @@ on-screen while displaying the warning</string>
         <property name="text">
          <string>Show report view on warning</string>
         </property>
-        <property name="checked">
-         <bool>true</bool>
-        </property>
         <property name="prefEntry" stdset="0">
          <cstring>checkShowReportViewOnWarning</cstring>
         </property>
@@ -304,7 +301,7 @@ on-screen while displaying the log message</string>
           <property name="text">
            <string/>
           </property>
-          <property name="color" stdset="0">
+          <property name="color">
            <color>
             <red>0</red>
             <green>0</green>
@@ -362,7 +359,7 @@ on-screen while displaying the log message</string>
           <property name="text">
            <string/>
           </property>
-          <property name="color" stdset="0">
+          <property name="color">
            <color>
             <red>0</red>
             <green>0</green>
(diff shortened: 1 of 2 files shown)


0penBrain (changes requested):
At least following change is missing https://github.com/FreeCAD/FreeCAD/blob/a157b8808495a12a0429233dc800d7e5aa65b0e1/src/Gui/ReportView.cpp#L264

donovaly:
> At least following change is missing

Thanks for testing!
I fixed this now.

0penBrain:
@donovaly to be transparent I didn't test. 🤣 Just read the code changes.

0penBrain (approved):
Looks good to me (from a coding point of view). I have no opinion if this is better than current behavior.

Syres916:
Is there no forum discussion regarding this change, there are some important warnings which won't get reported by new users making triage even more difficult? The real solution is to get the correct error/warning/log setting at each point in the code IMHO.

donovaly:
> Is there no forum discussion regarding this change, there are some important warnings which won't get reported by new users making triage even more difficult?

I describe myself as experienced user but even I can in many cases not figure out what a warning means.
Warnings are no errors, therefore it is not necessary to bother every user with them. And by having them shown by default is exactly this - bothering.
Those who want to see them can do so, by for example just leaving the report view open. But when i decided to close the report view, it is super annoying that it pops up on every warning.

You can bring this to the forum if you like. Please notify me in this case.

> The real solution is to get the correct error/warning/log setting at each point in the code IMHO.

We should distinguish errors from warnings.

Concerning warnings:
The best solution is to issue warnings when the users needs them. The PartDesign Helix feature is a good example. You are warned on creation of a helix that it can be self-intersecting. You only get this warning, if this is geometrically possible.

Concerning errors:
The optimum is to design the UI so that the user cannot do anythings wrong, e.g. not enter negative values when this leads to errors.

Personally, I try to achieve both and made a lot of PRs to improve the UI in this respect. However I am pretty sure we will never reach the optimum. Take for example the addons. Every addon can issue a warning and we cannot take care of the UI of every addon.

Syres916:
OK, let's see how it goes after it's merged.

donovaly:
> OK, let's see how it goes after it's merged.

OK, merging then.

adrianinsaval:
note that the warnings can still be seen in the status bar when they show up, I think this is a good choice, I always end up disabling this because it's just annoying.

PR #5285, opened 2021-12-23
[Mesh] rename gmsh -> Gmsh

donovaly:
I was informed that Gmsh names itself with a capital G. I had a look and in FEM it is already named Gmsh, but not in the Mesh WB. Therefore this PR.

(The other UI file changes than the pure renaming were automatically done by Qt's designer.)


donovaly:
No objections arose and since this is a simple renaming -> merging

PR #5288, opened 2021-12-24
[Draft] change static to non-static declared variables

MariwanJ:
Static declared variable which should be non-static. The same variable is created as a mistake with self.
This is related to the discussion
https://forum.freecadweb.org/viewtopic.php?f=22&t=64727

### DIFF
--- a/src/App/ObjectIdentifier.cpp
+++ b/src/App/ObjectIdentifier.cpp
@@ -1138,6 +1138,8 @@ void ObjectIdentifier::getDep(Dependencies &deps, bool needProps, std::vector<st
     catch (Py::Exception& e) {
         e.clear();
     }
+    catch (Base::Exception &) {
+    }
 }
 
 /**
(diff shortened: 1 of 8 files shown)


luzpaz:
Please rename commit to `Draft: change static to non-static declared variables`  
You can use `git commit --amend`

donovaly:
Many thanks for the PR!

To keep the overview, can you please squash your commits to one single commit? Please also avoid the merge commits. A PR should ideally only contain either on commit with all relevant changes or several commits in case different things are changed.
This allows us to review the PRs faster.

In your case, I assume that rebasing on current master and then force-pushing would lead to the state we need.

MariwanJ:
I close this and redo the PR. I cannot change it I tried that.

luzpaz:
@MariwanJ is there a technical reason why you couldn't change it? Or was it because you're still understanding `git`?

luzpaz:
Closed in favor of #5292

MariwanJ:
I couldn't change it using the command. I needed to re-base ..etc .
remaking the whole thing was better . thanks.

PR #5289, opened 2021-12-24
[core] Add new parameter to BaseApp/Preferences/Dialog called DontUse…

mwganson:
…NativeColorDialog, defaults to true.

This would hopefully resolve a crash with some systems where the native color dialog causes FreeCAD to crash.

https://forum.freecadweb.org/viewtopic.php?f=3&t=64291


luzpaz:
@mwganson are we documenting these parameters anywhere?

adrianinsaval:
should probably be documented in [Fine-tuning](https://wiki.freecadweb.org/Fine-tuning)

mwganson:
I will document in fine tuning.  I was waiting, of course, until it was merged before documenting.

PR #5292, opened 2021-12-25
Draft: change static to non-static declared variables

MariwanJ:
Static declared variable which should be non-static. The same variable is created as a mistake with self.
This is related to the discussion
https://forum.freecadweb.org/viewtopic.php?f=22&t=64727

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_trackers.py
+++ b/src/Mod/Draft/draftguitools/gui_trackers.py
@@ -699,7 +699,7 @@ class ghostTracker(Tracker):
         obj.ViewObject.show()
         self.finalize()
         sep = self.getNode(obj)
-        Tracker.__init__(self, children=[self.sep])
+        Tracker.__init__(self, children=[sep])
         self.on()
         obj.ViewObject.hide()
 
--- a/src/Mod/Draft/drafttaskpanels/task_shapestring.py
+++ b/src/Mod/Draft/drafttaskpanels/task_shapestring.py
@@ -77,7 +77,8 @@ class ShapeStringTaskPanel:
         self.fileSpec = Draft.getParam("FontFile", "")
         self.point = App.Vector(0.0, 0.0, 0.0)
         self.pointPicked = False
-        self.oldValueBuffer = False
+        # Default for the "DontUseNativeFontDialog" preference:
+        self.font_dialog_pref = False
 
         QtCore.QObject.connect(self.task.fcFontFile, QtCore.SIGNAL("fileNameSelected(const QString&)"), self.fileSelect)
         QtCore.QObject.connect(self.task.pbReset, QtCore.SIGNAL("clicked()"), self.resetPoint)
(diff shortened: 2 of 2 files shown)


MariwanJ:
I hope it is ok now.

donovaly:
It looks OK now, thanks.

@Roy-043 would you review this PR?

Roy-043:
Regarding `task_shapestring.py`:
This is OK of course.
Side note: the name of the variable `oldValueBuffer` (not changed in this PR) is not every descriptive.

Regarding `gui_trackers.py`:
AFAICT the `update` function that has been modified is not used anywhere. The code is a bit strange anyway. Hiding the `ViewObject` of the selected object would be confusing for the user. Also it would remain invisible when the tracker finishes.
I am currently looking at the `ghostTracker` class in relation to switching to and from subelement mode for Draft commands such as Draft_Move. In that context an `update` function can make sense, but it would have to be rewritten.

MariwanJ:
Hi, How would you say it is not changed?
The current version is a static variable while it should be non-static.
You cannot use self. for a static variable. It is a bug. Look at the discussion I had at the forum. 
I didn't look on what these functions do, I found the bugs and fixed them. What these functions do was not my goal. 
I searched for static variable that has then been used like non-static.. 
Even path has that problem and might be in other python library. It is a huge problem.

Roy-043:
> How would you say it is not changed?

You have misunderstood my 'side-note' comment. All I wanted to say was that I would have chosen a different name for `self.oldValueBuffer`.

MariwanJ:
Ok. My point in the PR is make it clear that the bug is in static - non-static representation of variables. I wanted to minimize the amount of changes I do to highlight the issue itself.  I had a huge problem with similar bug in my WB. And I have Draft in my WB so I needed to fix this also

luzpaz:
@Roy-043 what do you propose as a more ideal description name for `oldValueBuffer` ?

Roy-043:
'Ideal' is a big word, but something like `self.font_dialog_pref` would be clearer. Adding a comment when the variable is introduced is of course also useful:

```
        # Default for the "DontUseNativeFontDialog" preference:
        self.font_dialog_pref = False
```

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5292">pipeline status</a> for feature branch [PR_5292](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5292). Pipeline [439264150](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/439264150) was triggered at [a11fb77](https://github.com/FreeCAD/FreeCAD/pull/5292/commits/a11fb77fd040f9c4776d325fd1665ec3d065f458). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
@MariwanJ what do you think, would it be amenable to you to make these recommended changes by Roy?

MariwanJ:
Please you can do it on behalf of me if that possible. I think it is possible? 
Since my goal was to fix the issue.. other than that is not my intention.

donovaly:
> Please you can do it on behalf of me if that possible.

This is our workflow:
- you make a PR
- it is reviewed
- depending on the feedback the PR is adjusted either by the PR creator or in special cases later by an admin. The latter case is the exception for huge PRs

Therefore please do the suggested change and we can merge your PR.

MariwanJ:
Changed as per the request

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5292">pipeline status</a> for feature branch [PR_5292](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5292). Pipeline [442014163](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/442014163) was triggered at [865bf66](https://github.com/FreeCAD/FreeCAD/pull/5292/commits/865bf664cffaaeab5699e539532dbd4f7a313281). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thank you!

PR #5296, opened 2021-12-26
[Part::CheckGeometry] add edwilliams16's code for some global propert…

mwganson:
…ies to the advanced shape content

https://forum.freecadweb.org/viewtopic.php?f=3&t=64653&start=10

### DIFF
--- a/src/Mod/Part/BasicShapes/ShapeContent.py
+++ b/src/Mod/Part/BasicShapes/ShapeContent.py
@@ -1,0 +1,81 @@
+#! python
+# -*- coding: utf-8 -*-
+# ShapeContent.py
+# 2021, by Mark Ganson <TheMarkster>
+# LGPL 2.1 or later
+
+# this file is called from c++ in TaskCheckGeometry.cpp
+# from buildShapeContent()
+
+import FreeCAD as App
+
+def roundVector(v,dec):
+    return str([round(v[0],dec), round(v[1],dec), round(v[2],dec)])
+
+def buildShapeContent(basename, decimals=2, advancedShapeContent=True):
+    obj = basename[basename.index('.')+1:] #note: obj is a string, so is doc
+    doc = basename[:basename.index(obj)-1]
+    shp = App.getDocument(doc).getObject(obj).Shape
+    typeStr = str(shp.ShapeType)
+    result = doc+'.'+App.getDocument(doc).getObject(obj).Label+' ('+obj+'):\n'
+    result += 'Shape type:  '+typeStr+'\n'
+    result += 'Vertices:  '+str(len(shp.Vertexes))+'\n'
+    result += 'Edges:  '+str(len(shp.Edges))+'\n'
+    result += 'Wires:  '+str(len(shp.Wires))+'\n'
+    result += 'Faces:  '+str(len(shp.Faces))+'\n'
+    result += 'Shells:  '+str(len(shp.Shells))+'\n'
+    result += 'Solids:  '+str(len(shp.Solids))+'\n'
+    result += 'CompSolids:  '+str(len(shp.CompSolids))+'\n'
+    result += 'Compounds:  '+str(len(shp.Compounds))+'\n'
+    result += 'Shapes:  '+str(len(shp.Vertexes+shp.Edges+shp.Wires+shp.Faces+shp.Shells+shp.Solids+shp.CompSolids+shp.Compounds))+'\n'
+    if advancedShapeContent:
+        result += '----------\n'
+        if hasattr(shp,'Area') and not 'Wire' in typeStr and not 'Edge' in typeStr and not 'Vertex' in typeStr:
+            result += 'Area:  '+str(round(shp.Area, decimals))+'\n'
+        if hasattr(shp,'Volume') and not 'Wire' in typeStr and not 'Edge' in typeStr and not 'Vertex' in typeStr and not 'Face' in typeStr:
+            result += 'Volume:  '+str(round(shp.Volume, decimals))+'\n'
+        if hasattr(shp,'Mass'):
+            result += 'Mass:  '+str(round(shp.Mass, decimals))+'\n'
+        if hasattr(shp,'Length'):
+            result += 'Length:  '+str(round(shp.Length, decimals))+'\n'
+        if hasattr(shp,'Curve') and hasattr(shp.Curve,'Radius'):
+            result += 'Radius:  '+str(round(shp.Curve.Radius, decimals))+'\n'
+        if hasattr(shp,'Curve') and hasattr(shp.Curve,'Center'):
+            result += 'Curve center:  '+str([round(vv,decimals) for vv in shp.Curve.Center])+'\n'
+        if hasattr(shp,'Curve') and hasattr(shp.Curve,'Continuity'):
+            result += 'Continuity:  '+str(shp.Curve.Continuity)+'\n'
+        if hasattr(shp,'CenterOfMass'):
+            result += 'CenterOfMass:  '+roundVector(shp.CenterOfMass,decimals)+'\n'
+        if hasattr(shp,'normalAt'):
+            try:
+                result += 'normalAt(0):  '+str([round(vv,decimals) for vv in shp.normalAt(0)]) +'\n'
+            except Exception:
+                try:
+                    result += 'normalAt(0,0):  '+str([round(vv,decimals) for vv in shp.normalAt(0,0)]) +'\n'
+                except Exception:
+                    pass
+        if hasattr(shp, 'isClosed') and ('Wire' in typeStr or 'Edge' in typeStr):
+            result += 'isClosed:  '+str(shp.isClosed())+'\n'
+        if hasattr(shp, 'Orientation'):
+            result += 'Orientation:  '+str(shp.Orientation)+'\n'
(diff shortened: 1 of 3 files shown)


edwilliams16 (changes requested):
`        cmdstream << "            if 'AxisOfInertia' in _p:" << std::endl;
        cmdstream << "                _result += 'Global ' + str(_p)+':  '+str([round(vv," << decimals << ") for vv in _rpl.multVec(_props[_p])]) +'\\n'" << std::endl;
        cmdstream << "else:" << std::endl;`

should be

`        cmdstream << "            if 'AxisOfInertia' in _p:" << std::endl;
        cmdstream << "                _result += 'Global ' + str(_p)+':  '+str([round(vv," << decimals << ") for vv in _rot.multVec(_props[_p])]) +'\\n'" << std::endl;
        cmdstream << "else:" << std::endl;`

per [https://forum.freecadweb.org/viewtopic.php?p=556564#p556564](url)

wwmayer:
Wouldn't it make more sense to move the whole Python-code from buildShapeContent() to a real .py file? This would considerably simplify any future changes as it's easier to modify a Python file instead of a complicated string.

edwilliams16:
I have no idea about how this fits into the code architecture.   However the hurdle of being able to compile and debug C++ is significant.

mwganson:
I will work on making this change.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5296">pipeline status</a> for feature branch [PR_5296](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5296). Pipeline [439264987](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/439264987) was triggered at [73ce93b](https://github.com/FreeCAD/FreeCAD/pull/5296/commits/73ce93b419043d06d252ceb111fd8dea1b63976f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

mwganson:
@edwilliams16 If you can't compile on mac os, then I suggest to look into installing virtual box and getting one of the ubuntu virtual machine files from osboxes.org.  This is how I do it on Windows rather than fight with Windows compiling.  You still need to configure the build system in the ubuntu vm, but it's much easier, IMO, than on Mac or Windows.

edwilliams16:
@mwganson   Good idea.  Ubuntu 18.04.6?  - or should I go with 20.04.3?

mwganson:
Might as well go with the newest version.  I recently updated from 18.04
LTS to 20.04 LTS.  In Virtual Box there is the option to create a snapshot,
which I did.  In the event I wasn't able to get everything working as
desired I had the option to revert back to the snapshot.

wwmayer:
> I have no idea about how this fits into the code architecture. However the hurdle of being able to compile and debug C++ is significant.

The built-in Python code in buildShapeContent() you cannot debug anyway. So, the steps are:
* move the code to ShapeContent.py file and there to a function e.g. buildShapeContent
* it suffices to pass a shape to this function. No need to pass the baseName as it's currently done for the C++ variant. Additionally pass the number of decimals and the AdvancedShapeContent parameter key. Maybe you should define a default value for them.
* adjust the CMakeLists.txt file to copy the file when running the build/install target
* In TaskCheckGeometryResults::buildShapeContent you import the ShapeContent.py module and call the function buildShapeContent

Assuming you put ShapeContent.py into Part/BasicShapes you can implement the last point with:
```
....
Base::PyGILStateLocker lock;
try {
    PyObject* module = PyImport_ImportModule("BasicShapes.ShapeContent");
    if (!module) {
        throw Py::Exception();
    }

    Py::Tuple args(3);
    args.setItem(0, Py::asObject(new Part::TopoShapePy(new Part::TopoeShape(shape))));
    args.setItem(1, Py::Long(decimals));
    args.setItem(2, Py::Boolean(advancedShapeContent));
    Py::Module shapecontent(module, true);
    Py::String result(callMemberFunction("buildShapeContent", args));
    shapeContentString = result.as_std_string("ascii");
}
catch (Py::Exception& e) {
    e.clear();
    // report an error...
}
```

mwganson:
Excellent.  This is a great little code snippet.

We still need the basename as a string in order for Ed's code to work in order to have obj for obj.Placement and obj.getGlobalPlacement().  

But if we pass the string in, then we don't really need to pass in the shape because it can be trivially gotten as obj.Shape inside the script.  Another issue I am running into with the shape as an argument is then it seems to have no PrincipalProperties attribute whereas if the shape is obtained in the script it does have it.  That could be a bug perhaps somewhere else in FreeCAD.  For these reasons I will do away with the shape argument and pass in the basename as a string instead unless you have a better solution.

I think runStringWithKey() is no longer used anywhere if we don't use it here anymore.  It could be removed or it could be kept as something to use potentially later.

This is a much neater implementation than my earlier way of doing it.

wwmayer:
> We still need the basename as a string in order for Ed's code to work in order to have obj for obj.Placement and obj.getGlobalPlacement().

OK, then.

> But if we pass the string in, then we don't really need to pass in the shape because it can be trivially gotten as obj.Shape inside the script.

Yes, when passing the string the shape is not needed.

> Another issue I am running into with the shape as an argument is then it seems to have no PrincipalProperties attribute whereas if the shape is obtained in the script it does have it. That could be a bug perhaps somewhere else in FreeCAD.

No, it's not a bug. The attribute PrincipalProperties is only set for the shape sub-types TopoShape[Edge|Wire|Face|ShellSolid]Py. So, it's a matter of using the right type inside TaskCheckGeometryResults::buildShapeContent. The Python wrapper can be created as `Py::asObject(Part::TopoShape(shape).getPyObject())`

> For these reasons I will do away with the shape argument and pass in the basename as a string instead unless you have a better solution.

Nevertheless, I am going to merge your PR. Feel free to open a new PR if needed.

> I think runStringWithKey() is no longer used anywhere if we don't use it here anymore. It could be removed or it could be kept as something to use potentially later.

We can keep it for now.

Thanks and Happy New Year!

mwganson:
I see.  Thanks for the explanation.

It seems to me:

`Py::asObject(new Part::TopoShapePy(new Part::TopoShape(shape)));`

and 

`Py::asObject(Part::TopoShape(shape).getPyObject());`

should be just 2 different ways to construct the same object.  I believe you when you say they don't, but it seems to me they *should* produce the same object.  Otherwise it could lead to some very subtle bugs difficult to fix.

I forgot to remove an include I had added for the shape parameter (TopoShapePy.h, I think, just going from memory).

For benefit of anyone else reading, the new function can now be called from python:

```
from BasicShapes import ShapeContent as SC
SC.buildShapeContent('Unnamed.Cylinder', decimals = 2, advancedShapeContent = True)
```

It returns the string found in the shape content panel when check geometry is run.
Defaults for decimals and advancedShapeContent was a good idea.  Passing in document name dot object name is a bit peculiar, but it works.

edwilliams16:
@mwganson I spent a few hours with Virtualbox on a Mac and gave up.  I could get a virtual machine OK, but in order to have it talk to the host - in particular, getting cut and paste between host and guest working requires hacking the Mac more than I wanted to.  Apparently, you have to disable SIP (system integrity protection) to install the necessary drivers - following no official documentation I could find.  I do have Parallels on that Mac and so I tried that.  The Ubuntu instructions in the wiki worked great and although the compile took for ever I got working FreeCAD0.20_dev on my first attempt.  Still have to try a debug version and couple it to VSCode though.

wwmayer:
> should be just 2 different ways to construct the same object. I believe you when you say they don't, but it seems to me they should produce the same object.

It is not exactly the same. The former creates TopoShapePy that is the base class of TopoShapeVertexPy, TopoShapeEdgePy, ..., TopoShapeSolidPy and the latter creates the corresponding Python object to the underlying shape type.

> Otherwise it could lead to some very subtle bugs difficult to fix.

The point is that OCC itself isn't very strict in this regard. There are many functions that e.g. only accept a TopoDS_Face but in the function interface it uses TopoDS_Shape. As a convenience the Part module offers a function cast_to_shape that returns the correct Python wrapper to the underlying shape type.

> Passing in document name dot object name is a bit peculiar, but it works.

Maybe it's easier to not pass a string but directly the object itself:
`SC.buildShapeContent(App.ActiveDocument.Cylinder, decimals = 2, advancedShapeContent = True)`

mwganson:
@edwilliams16 I never got copy/paste working either.  It's an inconvenience, but not a deal breaker.  It also has the benefit of preserving something copied into the clipboard in the VM, for example a github key.  I am able to easily transfer files to/from the VM, so if I really need to copy/paste something I can do it via a text file.

I haven't used Parallels, but it probably is just as good as Virtual Box.  You might need to tweak the settings to allocate more RAM or more cores.  Also, when compiling if I use make -j4 it's much faster than simply using make.  I have 4 cores.

mwganson:
> > Passing in document name dot object name is a bit peculiar, but it works.
> 
> Maybe it's easier to not pass a string but directly the object itself: `SC.buildShapeContent(App.ActiveDocument.Cylinder, decimals = 2, advancedShapeContent = True)`

https://github.com/FreeCAD/FreeCAD/pull/5334

PR #5300, opened 2021-12-29
Mouse navigation: change unnecesary warnings to messages

adrianinsaval:
Sent when trying to set rotation center without an object under the cursor and can be annoying if report view is set to show with warnings. Would it be ok to just remove the message entirely? I think other navigation modes (besides gesture and maya that are adressed in this PR) don't show it. Also, why do we even need an object under the cursor?


wwmayer:
IMO, they could even be degraded to log messages so that they by default don't appear in the Output Window.

adrianinsaval:
agreed, do you want another PR later or you'll just do it yourself? I put them as messages in case somebody wanted them to be displayed (there must have been a reason why the warning was put there to begin with)

wwmayer:
> I put them as messages in case somebody wanted them to be displayed (there must have been a reason why the warning was put there to begin with)

I don't think that there was a valid reason for it. It's OK not to set the rotation center when clicking into the void but extra writing a message is really not needed.

3d359e2e24

PR #5302, opened 2021-12-29
Path: fix source comment typo

luzpaz:


benj5378 (approved):
This is good to go too. If you want, you could also run flake8 and black to clean up the document meanwhile

PR #5304, opened 2021-12-29
Tools: small bug fix for showing the correct test icon

mdkus:
The genereation of a new cpp module should work analogue to the generation of a python module. When generating and compiling a new cpp module, an error is shown while loading the module after starting FC: "Cannot find icon: Test1"

### DIFF
--- a/src/Tools/_TEMPLATE_/Gui/Command.cpp
+++ b/src/Tools/_TEMPLATE_/Gui/Command.cpp
@@ -47,7 +47,7 @@ Cmd_TEMPLATE_Test::Cmd_TEMPLATE_Test()
     sToolTipText  = QT_TR_NOOP("_TEMPLATE_ Test function");
     sWhatsThis    = "_TEMPLATE__Test";
     sStatusTip    = QT_TR_NOOP("_TEMPLATE_ Test function");
-    sPixmap       = "Test1";
+    sPixmap       = "_TEMPLATE_Workbench";
     sAccel        = "CTRL+H";
 }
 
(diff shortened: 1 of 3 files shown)


chennes:
As a general rule when saving an SVG from Inkscape for use in FreeCAD, you should use a "Plain SVG" rather than an "Inkscape SVG". It's unlikely to matter in this case, but if this is intended as a template for people to look at when creating their workbenches it's probably good to follow best-practices here.

mdkus:
> As a general rule when saving an SVG from Inkscape for use in FreeCAD, you should use a "Plain SVG" rather than an "Inkscape SVG". It's unlikely to matter in this case, but if this is intended as a template for people to look at when creating their workbenches it's probably good to follow best-practices here.

Ok, I see. I just copied the svg from existing file
[here](https://github.com/FreeCAD/FreeCAD/blame/master/src/Tools/_TEMPLATEPY_/Resources/icons/_TEMPLATEPY_Workbench.svg) and renamed it. Shall we take there also a plain svg?

chennes:
For now let's just do it here, to keep this PR simple and compact. That change can be made later (for exactly this reason, so when someone copies it again later they get a clean SVG).

mdkus:
Ok. Do you have an example plain svg? Shall I make a new PR after converting the svg into a plain svg?

chennes:
If you just open it in Inkscape and re-save, choosing "Plain SVG" as the option, you'll get the right thing. Then you can add a new commit to the branch you are working on with the new SVG file. When you push that commit, it will show here as part of the PR.

mdkus:
Ok, thank you.

mdkus:
Added the replacement of Inkscape svg with plain svg to the PR.

wwmayer:
Merged as a single commit: e4c2cc3c76

PR #5307, opened 2021-12-29
Draft: Fix projectPoint method

marioalexis84:
see https://forum.freecadweb.org/viewtopic.php?f=23&t=59903&sid=b6fcc8f0fae011760ddf1cd858ba354a

---

### DIFF
--- a/src/Mod/Draft/WorkingPlane.py
+++ b/src/Mod/Draft/WorkingPlane.py
@@ -34,6 +34,7 @@ YZ, and XZ planes.
 #  in FreeCAD and a couple of utility functions.
 
 import math
+from sys import float_info
 import lazy_loader.lazy_loader as lz
 
 import FreeCAD
@@ -226,7 +227,7 @@ class Plane:
             direction = self.axis
         return direction.dot(self.position.sub(p))
 
-    def projectPoint(self, p, direction=None):
+    def projectPoint(self, p, direction=None, force_projection=True):
         """Project a point onto the plane, by default orthogonally.
 
         Parameters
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5307">pipeline status</a> for feature branch [PR_5307](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5307). Pipeline [439265240](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/439265240) was triggered at [fcd4386](https://github.com/FreeCAD/FreeCAD/pull/5307/commits/fcd438646ef36c02a984e3ef4b27ff0771dca4a5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Roy-043:
I can confirm that this code fixes the issue in the forum topic. Thanks for that.

Repeating this test: https://forum.freecadweb.org/viewtopic.php?p=557743#p557743 with the new code shows the improved accuracy. The deviation of the local Z from zero goes from ca. 1e-5 to ca. 1e-15!

I am not that experienced with Python, but why do you use:
`if direction is not None:`
Instead of just:
`if direction:`
?

Two small remarks:
1. If `direction` is orthogonal to `axis` the function should return `None` IMO. You should not use the `axis` in this case.
2. If `direction` is a null vector there is a division by zero error. I am not sure if, or how, that should be handled.

marioalexis84:
> I am not that experienced with Python, but why do you use: `if direction is not None:` Instead of just: `if direction:` ?

Although effectively in the case of Base.Vector we can use `if direction:` without inconvenience, the code style guide proposes avoiding the use of booleans in a situation like this : https://www.python.org/dev/peps/pep-0008/#programming-recommendations

 
>     1. If `direction` is orthogonal to `axis` the function should return `None` IMO. You should not use the `axis` in this case.

Using `None`  generates a segmentation fault in Coin. Another option is to use a Vector made up of `nan`, but this generates some unexpected behavior in the 3D view. Therefore, to avoid these problems, I decided to use axis as the direction of the projection in this specific case since it allows a good workflow when the direction of the view is changed.
 
>     2. If `direction` is a null vector there is a division by zero error. I am not sure if, or how, that should be handled.

It should be the responsibility of the programmer not to pass a null vector as the direction of the projection.

Roy-043:
Thanks for your answers.

> Using `None` generates a segmentation fault in Coin.

Yes, the calling code would have to catch the None return. Implementing that would be more involved. But it is perhaps possible to create an independent project_point function in `DraftVecUtils.py` (?) and use that function here?

marioalexis84:
There are only two functions (in gui_selectplane.py and gui_snapper.py) that use the direction argument of projectPoint, so we can use `None` as return value when direction is orthogonal to axis, and catch it in those functions.
I will do these changes and force push again.

marioalexis84:
@Roy-043 can you test again?
I modified the method to take into account the case of orthogonality. A new mode `force_projection` is added. If `false` the return value is `None` when the direction is orthogonal to axis. If `True`, the direction is modified to point towards a distant point on the plane.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5307">pipeline status</a> for feature branch [PR_5307](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5307). Pipeline [444168608](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/444168608) was triggered at [1cd0124](https://github.com/FreeCAD/FreeCAD/pull/5307/commits/1cd01248e0fb42baace06ae7ff2475e6dd60c68d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

yorikvanhavre:
Thanks for the work on this guys! Also very high-level math for me :sweat_smile:

yorikvanhavre:
BTW @Roy-043 I have invited you to the "FreeCAD Reviewers" group here on github (you should see an invite somewhere) so you can be asked a review, if you don't mind of course.

Roy-043:
@marioalexis84 Thanks. The code works fine.

Roy-043:
@yorikvanhavre My Python knowledge is still limited. But I do not mind testing code.

yorikvanhavre:
Thanks for testing! Ok to merge @marioalexis84 ?

Roy-043:
I am not sure how important this is, but I notice that the epsilon value is very small (2.2e-16). It is much smaller than the tolerance used elsewhere. F.e. App.Rotation() uses a tolerance of 1e-7. Shorter vectors are ignored.

marioalexis84:
> I am not sure how important this is, but I notice that the epsilon value is very small (2.2e-16)

The class Base::Vector also use this precision value. No need to use OCCT precision.

> Thanks for testing! Ok to merge @marioalexis84 ?

It's ok to me.

marioalexis84:
It is worth noting that this particular case (orthogonality) occurs only if we project in parallel to WorkingPlane, which is not a useful case.

yorikvanhavre:
In any case indeed it would deserve to have a better discussion over overall Draft precision... Okay, let's merge already!

PR #5334, opened 2022-01-04
[Part::CheckGeometry] pass object to buildShapeContent() rather than …

mwganson:
…string as docname.objname

### DIFF
--- a/src/Mod/Part/Gui/TaskCheckGeometry.cpp
+++ b/src/Mod/Part/Gui/TaskCheckGeometry.cpp
@@ -451,7 +451,7 @@ void TaskCheckGeometryResults::goCheck()
         checkedCount++;
         checkedMap.Clear();
 
-        buildShapeContent(baseName, shape);
+        buildShapeContent(sel.pObject, baseName, shape);
 
         BRepCheck_Analyzer shapeCheck(shape);
         if (!shapeCheck.IsValid())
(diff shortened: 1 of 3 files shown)


src/Mod/Part/Gui/TaskCheckGeometry.cpp
```diff
@@ -603,7 +603,7 @@ void TaskCheckGeometryResults::buildShapeContent(const QString &baseName, const
             throw Py::Exception();
         }
         Py::Tuple args(3);
-        args.setItem(0, Py::String(baseName.toStdString().c_str()));
+        args.setItem(0, Py::Object(pObject->getPyObject()));
```
wwmayer, line 606: Almost perfect. However this line should be (see Py::Object vs Py::asObject):
`args.setItem(0, Py::asObject(pObject->getPyObject()));`

getPyObject() always returns an object with a new reference (i.e. incremented reference counter) and adding it to the tuple increments the reference a second time and that will result into a memory leak.

But don't worry I can merge the PR locally and do this minor fix myself.

wwmayer:
Merged as c93c0e6aaf

src/Mod/Part/Gui/TaskCheckGeometry.cpp
mwganson: Okay.  Thanks for letting me know about the fix.

PR #5335, opened 2022-01-05
[PD] get rid of a CI compiler warning

donovaly:
The CI compiler outputs this warning:
`FeaturePad.cpp:101:20: warning: variable 'obj' set but not used [-Wunused-but-set-variable]`

This PR is to see if the CI compiler warning vanishes.

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePad.cpp
+++ b/src/Mod/PartDesign/App/FeaturePad.cpp
@@ -100,7 +100,8 @@ App::DocumentObjectExecReturn *Pad::execute()
 
     TopoDS_Shape sketchshape;
     try {
-        Part::Feature* obj = getVerifiedObject();
+        getVerifiedObject();
+
         sketchshape = getVerifiedFace();
     }
     catch (const Base::Exception& e) {
--- a/src/Mod/PartDesign/App/FeaturePad.cpp
+++ b/src/Mod/PartDesign/App/FeaturePad.cpp
@@ -98,10 +98,9 @@ App::DocumentObjectExecReturn *Pad::execute()
     Midplane.setReadOnly(hasReversed);
     Reversed.setReadOnly(hasMidplane);
 
-    Part::Feature* obj = 0;
     TopoDS_Shape sketchshape;
     try {
-        obj = getVerifiedObject();
+        Part::Feature* obj = getVerifiedObject();
         sketchshape = getVerifiedFace();
     }
     catch (const Base::Exception& e) {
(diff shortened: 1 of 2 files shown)


src/Mod/PartDesign/App/FeaturePad.cpp
```diff
@@ -98,10 +98,9 @@ App::DocumentObjectExecReturn *Pad::execute()
     Midplane.setReadOnly(hasReversed);
     Reversed.setReadOnly(hasMidplane);
 
-    Part::Feature* obj = 0;
     TopoDS_Shape sketchshape;
     try {
-        obj = getVerifiedObject();
+        Part::Feature* obj = getVerifiedObject();
```
0penBrain, line 103:
```suggestion
        getVerifiedObject();

```
Here you still have an unused 'obj' variable ;)

src/Mod/PartDesign/App/FeaturePocket.cpp
```diff
@@ -98,10 +98,9 @@ App::DocumentObjectExecReturn *Pocket::execute()
     if ((std::string(Type.getValueAsString()) == "TwoLengths") && (L < Precision::Confusion()))
         return new App::DocumentObjectExecReturn("Pocket: Second length of pocket too small");
 
-    Part::Feature* obj = 0;
     TopoDS_Shape profileshape;
     try {
-        obj = getVerifiedObject();
+        Part::Feature* obj = getVerifiedObject();
```
0penBrain, line 103:
```suggestion
        getVerifiedObject();

```

src/Mod/PartDesign/App/FeaturePad.cpp
donovaly: Then I don't understand the compiler warning. The idea is to check if
`getVerifiedObject();`
fails. To do this,  must store the result of to a variable so the variable is used.

Do you know a way to silence the compiler?

donovaly: > I fear this can trigger warnings with other compilers.

Hmm, it seems to be OK also with the super strict MSVC settings I have. But OK, MSVC did not output the unused variable warning.

0penBrain: No you don't need to store into a variable, hence my suggestion just to call the function without any assignment.
The idea here is to call 'getVerifiedObject' into a 'try' block because the function will throw an error if it finds a problem (that will then jump the code into the 'catch'. ;)

donovaly:
OK googling around it appears this is safe. -> since the change is trivial, merge to trigger the CI

PR #5339, opened 2022-01-05
Addon manager dependency resolver

chennes:
The main work of this PR is to add a basic dependency checker that will tell the user installing a package what missing dependencies they have. It does not currently offer to install them, just presents them in a list that can either be ignored (installing the package anyway) or cancelled (cancelling the installation).

As a secondary function, this PR cleans up various display items:
1) it checks for an internet connection on startup, and presents a (hopefully clear) error message if one is not found
2) The packages list is correctly filtered on the first refresh
3) The packages list should flicker less during its initial load
4) Very long lists of failed and succeeded updates are consolidated so they don't exceed maximum dialog size

### DIFF
--- a/src/Mod/AddonManager/AddonManager.py
+++ b/src/Mod/AddonManager/AddonManager.py
@@ -85,6 +85,7 @@ class CommandAddonManager:
         "load_macro_metadata_worker",
         "update_all_worker",
         "update_check_single_worker",
+        "dependency_installation_worker",
     ]
 
     lock = threading.Lock()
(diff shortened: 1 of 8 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5339">pipeline status</a> for feature branch [PR_5339](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5339). Pipeline [443191057](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443191057) was triggered at [b846f5d](https://github.com/FreeCAD/FreeCAD/pull/5339/commits/b846f5d97dcaee3892cc9e2afae28cfd8b0e29d9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #5340, opened 2022-01-06
[PartDesign Fillet/Chamfer] add UseAllEdges boolean property

mwganson:
Adds a new boolean property called "UseAllEdges" to Part Design fillets and chamfers.  This will make these features robust against topological naming problem if this property is set to True, because then no matter if the edges of the base object change this dressup feature will still be using all of the edges.

Pros:

- Robust against topological naming if set to True

- Convenient way to fillet/chamfer all edges instead of opening the dialog

Cons:

-If the user has set UseAllEdges to true and then tries to edit the edges in the list, then the changes to the list will have no effect, possibly confusing and frustrating some users.

There is some minor code duplication, but I think it's better not to put the code in FeatureDressup.cpp because that class is also used by Draft and Thickness, which will not get the UseAllEdges property.

We could set this property to true when a new fillet/chamfer is created and the user has not selected any edges instead of populating the Base property with all the edges, but I decided against this for a couple reasons: 1) it will confuse users who will expect to be able to edit the edges in the dialog, but if the user sets the property himself, then this will be less likely; and 2) files created in the new version will still be able to be opened in 0.19, which will know nothing of the UseAllEdges property, but this will not be the case if the Base property is left empty and the UseAllEdges property is set to true.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureChamfer.cpp
+++ b/src/Mod/PartDesign/App/FeatureChamfer.cpp
@@ -76,7 +76,7 @@ Chamfer::Chamfer()
     Angle.setConstraints(&floatAngle);
 
     ADD_PROPERTY_TYPE(FlipDirection, (false), "Chamfer", App::Prop_None, "Flip direction");
-    ADD_PROPERTY_TYPE(UseAllEdges, (false), "Base", App::Prop_None,
+    ADD_PROPERTY_TYPE(UseAllEdges, (false), "Chamfer", App::Prop_None,
              "Chamfer all edges if true, else use only those edges in Base property.\n"
              "If true, then this overrides any edge changes made to the Base property or in the dialog.\n");
 
--- a/src/Mod/PartDesign/App/FeatureChamfer.cpp
+++ b/src/Mod/PartDesign/App/FeatureChamfer.cpp
@@ -76,6 +76,9 @@ Chamfer::Chamfer()
     Angle.setConstraints(&floatAngle);
 
     ADD_PROPERTY_TYPE(FlipDirection, (false), "Chamfer", App::Prop_None, "Flip direction");
+    ADD_PROPERTY_TYPE(UseAllEdges, (false), "Base", App::Prop_None,
+             "Chamfer all edges if true, else use only those edges in Base property.\n"
+             "If true, then this overrides any edge changes made to the Base property or in the dialog.\n");
 
     updateProperties();
 }
@@ -115,8 +118,20 @@ App::DocumentObjectExecReturn *Chamfer::execute(void)
     }
 
     std::vector<std::string> SubNames = std::vector<std::string>(Base.getSubValues());
+
+    if (UseAllEdges.getValue()){
+        SubNames.clear();
+        std::string edgeTypeName = Part::TopoShape::shapeName(TopAbs_EDGE); //"Edge"
+        int count = TopShape.countSubElements(edgeTypeName.c_str());
+        for (int ii = 0; ii < count; ii++){
+            std::ostringstream edgeName;
+            edgeName << edgeTypeName << ii+1;
+            SubNames.push_back(edgeName.str());
+        }
+    }
+
     std::vector<std::string> FaceNames;
-  
+
     getContinuousEdges(TopShape, SubNames, FaceNames);
 
     if (SubNames.size() == 0)
--- a/src/Mod/PartDesign/App/FeatureFillet.cpp
+++ b/src/Mod/PartDesign/App/FeatureFillet.cpp
@@ -53,10 +53,10 @@ const App::PropertyQuantityConstraint::Constraints floatRadius = {0.0,FLT_MAX,0.
 
 Fillet::Fillet()
 {
-    ADD_PROPERTY(Radius,(1.0));
+    ADD_PROPERTY_TYPE(Radius, (1.0), "Fillet", App::Prop_None, "Fillet radius.");
     Radius.setUnit(Base::Unit::Length);
     Radius.setConstraints(&floatRadius);
-    ADD_PROPERTY_TYPE(UseAllEdges, (false), "Base", App::Prop_None,
+    ADD_PROPERTY_TYPE(UseAllEdges, (false), "Fillet", App::Prop_None,
       "Fillet all edges if true, else use only those edges in Base property.\n"
       "If true, then this overrides any edge changes made to the Base property or in the dialog.\n");
 }
--- a/src/Mod/PartDesign/App/FeatureFillet.cpp
+++ b/src/Mod/PartDesign/App/FeatureFillet.cpp
@@ -56,6 +56,9 @@ Fillet::Fillet()
     ADD_PROPERTY(Radius,(1.0));
     Radius.setUnit(Base::Unit::Length);
     Radius.setConstraints(&floatRadius);
+    ADD_PROPERTY_TYPE(UseAllEdges, (false), "Base", App::Prop_None,
+      "Fillet all edges if true, else use only those edges in Base property.\n"
+      "If true, then this overrides any edge changes made to the Base property or in the dialog.\n");
 }
 
 short Fillet::mustExecute() const
@@ -73,7 +76,19 @@ App::DocumentObjectExecReturn *Fillet::execute(void)
     } catch (Base::Exception& e) {
         return new App::DocumentObjectExecReturn(e.what());
     }
-    std::vector<std::string> SubNames = std::vector<std::string>(Base.getSubValues());
+    std::vector<std::string> SubNames = std::vector<std::string>(Base.getSubValues());  
+
+    if (UseAllEdges.getValue()){
+        SubNames.clear();
+        std::string edgeTypeName = Part::TopoShape::shapeName(TopAbs_EDGE); //"Edge"
+        int count = TopShape.countSubElements(edgeTypeName.c_str());
+        for (int ii = 0; ii < count; ii++){
+            std::ostringstream edgeName;
+            edgeName << edgeTypeName << ii+1;
+            SubNames.push_back(edgeName.str());
+        }
+    }
+
     getContinuousEdges(TopShape, SubNames);
 
     if (SubNames.size() == 0)
--- a/src/Mod/PartDesign/App/FeatureChamfer.h
+++ b/src/Mod/PartDesign/App/FeatureChamfer.h
@@ -44,6 +44,7 @@ public:
     App::PropertyQuantityConstraint Size2;
     App::PropertyAngle Angle;
     App::PropertyBool FlipDirection;
+    App::PropertyBool UseAllEdges;
 
     /** @name methods override feature */
     //@{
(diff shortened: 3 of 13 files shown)


src/Mod/PartDesign/App/FeatureFillet.cpp
```diff
@@ -56,6 +56,9 @@ Fillet::Fillet()
     ADD_PROPERTY(Radius,(1.0));
     Radius.setUnit(Base::Unit::Length);
     Radius.setConstraints(&floatRadius);
+    ADD_PROPERTY_TYPE(UseAllEdges,(false),"Base", App::Prop_None,
+      "Fillet all edges if true, else if false use only those edges in Base property\n"
+      "If true, then this overrides any edge changes made to the Base property or in the dialog.\n");
```
AjinkyaDahale, line 61:
```suggestion
    ADD_PROPERTY_TYPE(UseAllEdges,(false),"Base",App::Prop_None,
      "Fillet all edges if true, else use only those edges in Base property.\n"
      "If true, this overrides any edge changes made to the Base property or in the dialog.\n");

```

src/Mod/PartDesign/App/FeatureFillet.cpp
```diff
@@ -73,7 +76,18 @@ App::DocumentObjectExecReturn *Fillet::execute(void)
...
+    std::vector<std::string> SubNames = std::vector<std::string>(Base.getSubValues());  
+
+    if (UseAllEdges.getValue()){
+        SubNames.clear();
+        int count = TopShape.countSubElements("Edge");
+        for (int ii = 0; ii < count; ii++){
+            std::ostringstream edgeName;
+            edgeName << "Edge" << ii+1;
+            SubNames.push_back(edgeName.str());
+        }
```
AjinkyaDahale, line 88: I expect this to be invalid once topo-naming is merged---edges will start having names like `Edge_Between_Face1_and_Face2`. It would be better to add loop through the edges themselves and add their names. I'll see if I can't find a snippet for this.

AjinkyaDahale (changes requested):

src/Mod/PartDesign/App/FeatureFillet.cpp
mwganson: Okay, I like that wording better.  I need spaces after the commas, too.

src/Mod/PartDesign/App/FeatureFillet.cpp
adrianinsaval: Is that really how the topo naming algorithm works? That's not how I understood it.

mwganson: If this happens it will break a lot of stuff, especially addons depending on these subshape names.

I have found a way to get the type name (e.g. "Edge", "Vertex", "Face") from the sub shape object.  I will update the PR to use that instead of hard-coding "Edge".

AjinkyaDahale: @adrianinsaval I won't pretend to understand if that that's how it works so I might be wrong, but I do believe running through the edges and adding their names to the list (or the edges themselves) is better than creating labels which may or may not work.

src/Mod/PartDesign/App/FeatureFillet.cpp
AjinkyaDahale: Thanks for making the changes. I have seen that when using a macro, commas are often not used. However, it seems they are used in some other places, so that's okay I suppose.

src/Mod/PartDesign/App/FeatureChamfer.cpp
```diff
@@ -76,6 +76,9 @@ Chamfer::Chamfer()
     Angle.setConstraints(&floatAngle);
 
     ADD_PROPERTY_TYPE(FlipDirection, (false), "Chamfer", App::Prop_None, "Flip direction");
+    ADD_PROPERTY_TYPE(UseAllEdges, (false), "Base", App::Prop_None,
+             "Chamfer all edges if true, else use only those edges in Base property\n"
```
AjinkyaDahale, line 80: There should be a full stop before the newline here.

src/Mod/PartDesign/App/FeatureFillet.cpp
adrianinsaval: I guess it's safer but my understanding is that we will still have EdgeN as the names for edges, there's a separate mapped name that get's the history stuff and it's a bunch of letters and numbers not long descriptive names like that, the algorithm then finds which edge number corresponds to the mapped name.

AjinkyaDahale: I see. If so that's good.

src/Mod/PartDesign/App/FeatureChamfer.cpp
mwganson: What do you mean?  A period?

AjinkyaDahale: Yes. It's also called a period.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5340">pipeline status</a> for feature branch [PR_5340](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5340). Pipeline [443190369](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443190369) was triggered at [85adbf6](https://github.com/FreeCAD/FreeCAD/pull/5340/commits/85adbf6b915a24fe8ab1db42e6b1ee18dc3a1a08). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5340">pipeline status</a> for feature branch [PR_5340](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5340). Pipeline [446453070](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/446453070) was triggered at [c7c2282](https://github.com/FreeCAD/FreeCAD/pull/5340/commits/c7c228259c622daf2cd739163cfb6616f0b40338). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Why not adding the new property to the task dialog? This will resolve the main con:

- select e.g. a PD pad and then chamfer
- all edges will already be chamfered
- when now this PR would be extended to set in this case a new dialog option that all edges are chamfered, it is clear for the user, why the selection list is not editable - would then be disabled.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5340">pipeline status</a> for feature branch [PR_5340](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5340). Pipeline [462151169](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/462151169) was triggered at [ee1bc90](https://github.com/FreeCAD/FreeCAD/pull/5340/commits/ee1bc90088cf31f685d6e470c8f7b8a585240a8a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

mwganson:
> Why not adding the new property to the task dialog? This will resolve the main con:
> 
> * select e.g. a PD pad and then chamfer
> * all edges will already be chamfered
> * when now this PR would be extended to set in this case a new dialog option that all edges are chamfered, it is clear for the user, why the selection list is not editable - would then be disabled.

Okay.  I'll see if I can add it to the dialogs.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5340">pipeline status</a> for feature branch [PR_5340](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5340). Pipeline [462378322](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/462378322) was triggered at [8ccbd7f](https://github.com/FreeCAD/FreeCAD/pull/5340/commits/8ccbd7fbda6c0b299d8c5c9c26b5d0fb5cf763ed). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks for the update. It looks good.
I have 2 issues:

- the context menu must be removed since this feature is now outsourced to the option: 
XiulSuls8q

- the new property is a propertiy of the chamfer feature, therefore it must not be in the Base section, but should appear as last entry of the Chamfer section:
InkedFreeCAD_MwCRT72JcS_LI

mwganson:
The context menu is still useful.  For example, suppose you have an object with 100 edges and you wish to chamfer all but 4 of those edges.  Then the easiest workflow would be to add all the edges in the context menu, and then manually remove the 4 edges not to be chamfered rather than manually add 96 edges.

mwganson:
When the user selects only the base object, and then clicks the fillet/chamfer toolbar icon, currently (in this PR) all the edges are added to the list and the UseAllEdges property is set to True.  This is useful if the user wants only to do for example 96/100 edges.  He can disable UseAllEdges in the dialog and remove the 4 unwanted edges manually.  But would it be better to just add Edge1 in such cases and set UseAllEdges = True?

Here is my reasoning.  UseAllEdges = True offers protection from TNP, but if the user makes changes and now the base object has fewer edges than it had when the dressup was first created and the user toggles UseAllEdges to False, then the dressup will fail until the now no longer existing edges are removed from the list.  If only Edge1 is in the list box, then when the user toggles UseAllEdges to False the dressup will not fail (usually).  We must have at least one edge or else there can be a crash.  This is why you cannot remove all the edges in the dialog; you must always leave at least 1.

In the case where the user wants to chamfer 96/100 edges he can still use the add all edges context menu option, and then remove the other 4 in the dialog.

adrianinsaval:
sorry for the off topic but what causes the crash when there's no edge selected? Can't we put something like if edges=0 break before the crash and allow removing that damned edge? It's annoying that you have to add another edge before you remove the last edge

donovaly:
> Here is my reasoning.

Thanks. Merging now.

donovaly:
> sorry for the off topic but what causes the crash when there's no edge selected?

Where do you see a crash?

donovaly:
@mwganson , please don't forget to describe the new feature in the Wiki pages (for chamfer/fillet but also the 0.2 release notes).

adrianinsaval:
mwganson mentioned there would be a crash if we don't have at least one edge selected and that's the reason it's forbidden to remove all edges

donovaly:
> mwganson mentioned there would be a crash if we don't have at least one edge selected and that's the reason it's forbidden to remove all edges

Ah, yes, I remember. Therefore I had for example to assure that on deletion of edges, at least one remains. However, this was with OCC 7.3, now it should no longer crash but popup an error. However, for many users this was super annoying and beginners did not know what the error is about.

adrianinsaval:
Ok that makes sense, but can't we catch the absence of an edge before the OCC calls?

PR #5348, opened 2022-01-06
TopoShape::analyze() SetRunParallel(true) to do the bopcheck in paral…

mwganson:
…lel mode

This should improve performance of running check(True) on a shape, such as:

`obj.Shape.check(True)`

from the Python console.  In a test with a difficult file without running in parallel mode, it took about 40.5 seconds, in parallel mode only 30.8 seconds.

### DIFF
--- a/src/Mod/Part/App/TopoShape.cpp
+++ b/src/Mod/Part/App/TopoShape.cpp
@@ -1661,6 +1661,7 @@ bool TopoShape::analyze(bool runBopCheck, std::ostream& str) const
 #endif
 #if OCC_VERSION_HEX >= 0x060900
             BOPCheck.SetParallelMode(true); //this doesn't help for speed right now(occt 6.9.1).
+            BOPCheck.SetRunParallel(true); //performance boost, use all available cores
             BOPCheck.TangentMode() = true; //these 4 new tests add about 5% processing time.
             BOPCheck.MergeVertexMode() = true;
             BOPCheck.CurveOnSurfaceMode() = true;


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5348">pipeline status</a> for feature branch [PR_5348](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5348). Pipeline [443190921](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443190921) was triggered at [aa23ffb](https://github.com/FreeCAD/FreeCAD/pull/5348/commits/aa23ffb6c930440d4c5429f34917f7a77757b86e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/TopoShape.cpp
```diff
@@ -1661,6 +1661,7 @@ bool TopoShape::analyze(bool runBopCheck, std::ostream& str) const
 #endif
 #if OCC_VERSION_HEX >= 0x060900
             BOPCheck.SetParallelMode(true); //this doesn't help for speed right now(occt 6.9.1).
```
luzpaz, line 1663: @mwganson is this comment still relavent?

mwganson: I think not.

PR #5349, opened 2022-01-06
[PD] Pocket: add missing apply code

donovaly:
adds apply code that Pad has but not also Pocket

It seems this was forgotten when the direction feature was added to Pocket or it was lost during the Pad-Pocket code refactoring.

Nevertheless I wonder why the missing code has no visible effect with my test files. Werner, since you did the Pad-Pocket code refactoring, could you please have a look if this PR is necessary?

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskPocketParameters.cpp
@@ -203,7 +203,12 @@ void TaskPocketParameters::apply()
     ui->lengthEdit->apply();
     ui->lengthEdit2->apply();
 
-    FCMD_OBJ_CMD(obj,"Type = " << getMode());
+    FCMD_OBJ_CMD(obj, "UseCustomVector = " << (getCustom() ? 1 : 0));
+    FCMD_OBJ_CMD(obj, "Direction = ("
+        << getXDirection() << ", " << getYDirection() << ", " << getZDirection() << ")");
+    FCMD_OBJ_CMD(obj, "ReferenceAxis = " << getReferenceAxis());
+    FCMD_OBJ_CMD(obj, "AlongSketchNormal = " << (getAlongSketchNormal() ? 1 : 0));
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5349">pipeline status</a> for feature branch [PR_5349](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5349). Pipeline [443190922](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443190922) was triggered at [2205d2c](https://github.com/FreeCAD/FreeCAD/pull/5349/commits/2205d2c2d87d83e853fd1f70424fe7a4f603f7cd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> It seems this was forgotten when the direction feature was added to Pocket or it was lost during the Pad-Pocket code refactoring.

Nothing was lost during the refactoring. This is how it is implemented in v0.19:
https://github.com/FreeCAD/FreeCAD/blob/releases/FreeCAD-0-19/src/Mod/PartDesign/Gui/TaskPadParameters.cpp#L557
https://github.com/FreeCAD/FreeCAD/blob/releases/FreeCAD-0-19/src/Mod/PartDesign/Gui/TaskPocketParameters.cpp#L486

> Nevertheless I wonder why the missing code has no visible effect with my test files.

Because it assigns the same values to the properties again. The values are immediately set when the user interacts with the dialog. The code in the apply() function is actually only there for macro recording.

Your PR makes TaskPadParameters::apply() and TaskPocketParameters::apply() identical and thus the function should go to the base class TaskExtrudeParameters

donovaly:
> Your PR makes TaskPadParameters::apply() and TaskPocketParameters::apply() identical and thus the function should go to the base class TaskExtrudeParameters

Thanks. Here is the PR that does this: #5365

PR #5350, opened 2022-01-07
Path: PathArray: use python library 're' instead of 'regex'

jimzim111:
The python standard library includes 're' but not 'regex'.  'regex' has features that don't benefit here but has to be installed separately.  So, prefer 're', since it is available everywhere.

### DIFF
--- a/src/Mod/Path/PathScripts/post/grbl_post.py
+++ b/src/Mod/Path/PathScripts/post/grbl_post.py
@@ -31,7 +31,7 @@ import argparse
 import datetime
 import shlex
 import PathScripts.PathUtil as PathUtil
-import regex
+import re
 
 
 TOOLTIP = """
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5350">pipeline status</a> for feature branch [PR_5350](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5350). Pipeline [443190925](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443190925) was triggered at [34415c6](https://github.com/FreeCAD/FreeCAD/pull/5350/commits/34415c6946a6a32a2d82b4e3eb88c7c3dce7c09a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #5354, opened 2022-01-07
[PD] Pipe: use button groups

donovaly:
almost same PR as #5189 but use QButtonGroup directly, no custom implementation


donovaly:
Closing in favor of #5354

PR #5355, opened 2022-01-08
[PD] Pad/Pocket: sort out duplicated code

donovaly:
the checkboxes are almost identical and it avoids work to have them in one location

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
@@ -25,10 +25,10 @@
 
 #ifndef _PreComp_
 # include <sstream>
-# include <QRegExp>
-# include <QTextStream>
 # include <Precision.hxx>
+# include <QRegExp>
 # include <QSignalBlocker>
+# include <QTextStream>
 #endif
 
 #include "ui_TaskPadPocketParameters.h"
@@ -387,6 +387,82 @@ void TaskExtrudeParameters::addAxisToCombo(App::DocumentObject* linkObj, std::st
     // store the face normal as sublink
     if (hasSketch)
         lnk.setValue(linkObj, std::vector<std::string>(1, linkSubname));
+}
+
+void TaskExtrudeParameters::setCheckboxes(int index, Modes mode)
+{
+    // disable/hide everything unless we are sure we don't need it
+    // exception: the direction parameters are in any case visible
+    bool isLengthEditVisible = false;
+    bool isLengthEdit2Visible = false;
+    bool isOffsetEditVisible = false;
+    bool isOffsetEditEnabled = true;
+    bool isMidplaneEnabled = false;
+    bool isMidplaneVisible = false;
+    bool isReversedEnabled = false;
+    bool isFaceEditEnabled = false;
+
+    if (mode == Modes::Dimension) {
+        isLengthEditVisible = true;
+        ui->lengthEdit->selectNumber();
+        QMetaObject::invokeMethod(ui->lengthEdit, "setFocus", Qt::QueuedConnection);
+        isMidplaneVisible = true;
+        isMidplaneEnabled = true;
+        // Reverse only makes sense if Midplane is not true
+        isReversedEnabled = !ui->checkBoxMidplane->isChecked();
+    }
+    else if (mode == Modes::ThroughAll) {
+        isOffsetEditVisible = true;
+        isOffsetEditEnabled = false; // offset may have some meaning for through all but it doesn't work
+        isMidplaneEnabled = true;
+        isReversedEnabled = !ui->checkBoxMidplane->isChecked();
+    }
+    else if (mode == Modes::ToFirst) {
+        isOffsetEditVisible = true;
+        isReversedEnabled = true;
+    }
+    else if (mode == Modes::ToLast) {
+        isOffsetEditVisible = true;
+        isReversedEnabled = true;
+    }
+    else if (mode == Modes::ToFace) {
+        isOffsetEditVisible = true;
+        isReversedEnabled = true;
+        isFaceEditEnabled = true;
+        QMetaObject::invokeMethod(ui->lineFaceName, "setFocus", Qt::QueuedConnection);
+        // Go into reference selection mode if no face has been selected yet
+        if (ui->lineFaceName->property("FeatureName").isNull())
+            onButtonFace(true);
+    }
+    else if (mode == Modes::TwoDimensions) {
+        isLengthEditVisible = true;
+        isLengthEdit2Visible = true;
+        isReversedEnabled = true;
+    }
+
+    ui->lengthEdit->setVisible(isLengthEditVisible);
+    ui->lengthEdit->setEnabled(isLengthEditVisible);
+    ui->labelLength->setVisible(isLengthEditVisible);
+    ui->checkBoxAlongDirection->setVisible(isLengthEditVisible);
+
+    ui->lengthEdit2->setVisible(isLengthEdit2Visible);
+    ui->lengthEdit2->setEnabled(isLengthEdit2Visible);
+    ui->labelLength2->setVisible(isLengthEdit2Visible);
+
+    ui->offsetEdit->setVisible(isOffsetEditVisible);
+    ui->offsetEdit->setEnabled(isOffsetEditVisible && isOffsetEditEnabled);
+    ui->labelOffset->setVisible(isOffsetEditVisible);
+
+    ui->checkBoxMidplane->setEnabled(isMidplaneEnabled);
+    ui->checkBoxMidplane->setVisible(isMidplaneVisible);
+
+    ui->checkBoxReversed->setEnabled(isReversedEnabled);
+
+    ui->buttonFace->setEnabled(isFaceEditEnabled);
+    ui->lineFaceName->setEnabled(isFaceEditEnabled);
+    if (!isFaceEditEnabled) {
+        onButtonFace(false);
+    }
 }
 
 void TaskExtrudeParameters::onDirectionCBChanged(int num)
--- a/src/Mod/PartDesign/Gui/TaskPadParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskPadParameters.cpp
@@ -80,70 +80,8 @@ void TaskPadParameters::updateUI(int index)
 {
     // update direction combobox
     fillDirectionCombo();
-    
-    // disable/hide everything unless we are sure we don't need it
-    // exception: the direction parameters are in any case visible
-    bool isLengthEditVisible  = false;
-    bool isLengthEdit2Visible = false;
-    bool isOffsetEditVisible  = false;
-    bool isMidplaneEnabled    = false;
-    bool isMidplaneVisible    = false;
-    bool isReversedEnabled    = false;
-    bool isFaceEditEnabled    = false;
-
-    Modes mode = static_cast<Modes>(index);
-
-    if (mode == Modes::Dimension) {
-        isLengthEditVisible = true;
-        ui->lengthEdit->selectNumber();
-        QMetaObject::invokeMethod(ui->lengthEdit, "setFocus", Qt::QueuedConnection);
-        isMidplaneEnabled = !ui->checkBoxReversed->isChecked();
-        isMidplaneVisible = true;
-        // Reverse only makes sense if Midplane is not true
-        isReversedEnabled = !ui->checkBoxMidplane->isChecked();
-    }
-    else if (mode == Modes::ToLast || mode == Modes::ToFirst) {
-        isOffsetEditVisible = true;
-        isReversedEnabled = true;
-    }
-    else if (mode == Modes::ToFace) {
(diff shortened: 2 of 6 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5355">pipeline status</a> for feature branch [PR_5355](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5355). Pipeline [443882266](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443882266) was triggered at [5345e58](https://github.com/FreeCAD/FreeCAD/pull/5355/commits/5345e58a3d456a6caa0d15d58420f99b17199eed). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
```diff
@@ -388,6 +390,94 @@ void TaskExtrudeParameters::addAxisToCombo(App::DocumentObject* linkObj, std::st
...
+    bool isOffsetEditEnabled = true;
+    bool isMidplaneEnabled = false;
+    bool isMidplaneVisible = false;
+    bool isReversedEnabled = false;
+    bool isFaceEditEnabled = false;
+
+    TaskPadParameters::Modes modePad;
+    TaskPocketParameters::Modes modePocket;
+
+    if (isPocket)
```
wwmayer, line 409: TaskExtrudeParameters is the base class of TaskPocketParameters and TaskPadParameters. Now using in a class stuff from derived class is not a great idea.

What you can do instead is moving the enum class Modes of TaskPocketParameters and TaskPadParameters to TaskExtrudeParameters. Now there is one enum entry different of the two classes and you can solve it this way:
```
    enum class Modes {
        Dimension,
        ThroughAll,
        ToLast = ThroughAll,
        ToFirst,
        ToFace,
        TwoDimensions
    };
```

This way you don't need to have two variables modePad & modePocket but only one.

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
```diff
@@ -388,6 +390,94 @@ void TaskExtrudeParameters::addAxisToCombo(App::DocumentObject* linkObj, std::st
         lnk.setValue(linkObj, std::vector<std::string>(1, linkSubname));
 }
 
+void TaskExtrudeParameters::setCheckboxes(int index, bool isPocket)
```
wwmayer, line 393: It may sound nitpicking but the boolean here is a code smell (called boolean blindness) because in places where the function is used you see this `setCheckboxes(index, true);` or this `setCheckboxes(index, false);` and for a reviewer it's not immediately clear what the intention is, hence boolean blindness. So he has to go to the class declaration to get its meaning.

A better way is to use an enum that can be declared as:
```
enum class Type {
  Pad,
  Pocket
};
```
Then in the code you will have to write `setCheckboxes(index, Type::Pocket);` or setCheckboxes(index, Type::Pad);`

donovaly: > It may sound nitpicking but the boolean here is a code smell (called boolean blindness) because in places where the function is used you see this `setCheckboxes(index, true);` or this `setCheckboxes(index, false);` and for a reviewer it's not immediately clear what the intention is, hence boolean blindness. 

I can change this to an enum. However I don't understand the term "blindness". I named the Boolean "isPocket", so it is clear what it is about. The IDEs like KDevelop, MSVC etc. show you thin info why typing. Also when you right. click on it, you get the info about the procedure definition. If I would have added a comment in the .h file, the IDE will also show me this. So with one click, the coders get the info.
When you don't have an IDE but use a text editor, then I agree an enums helps for readability.

I must admit I use bools because this is quicker ceded than adding an enum, since I write a procedure and in the IDE I can right-click on it to ad its definition in the header file. Therefore I only need to write once, the rest is automatically done by the IDE. (Only the position in in private/public etc. needs sometimes to be adapted.)

So should I in future prefer enums over bools?

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
donovaly: This was my first attempt but since the 2 enums are different for Pad and Pocket, I did not find a suitable solution. Through all gets the "through all" length and this is longer than when to the last face. Therefore the 2 enums are different.

Maybe I have a think here? If not, is there another way?

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
donovaly: The enums are now in.

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
wwmayer: > This was my first attempt but since the 2 enums are different for Pad and Pocket, I did not find a suitable solution. Through all gets the "through all" length and this is longer than when to the last face. Therefore the 2 enums are different.

Sorry, but I don't get it. 
Currently we have the two enums
```
    enum class Modes {
        Dimension,
        ToLast,
        ToFirst,
        ToFace,
        TwoDimensions
    };
```
with Dimension=0, ToLast=1, ToFirst=2, ToFace=3, TwoDimensions=4 and
```
    enum class Modes {
        Dimension,
        ThroughAll,
        ToFirst,
        ToFace,
        TwoDimensions
    };
```
with Dimension=0, ThroughAll=1, ToFirst=2, ToFace=3, TwoDimensions=4. The only difference is ThroughAll and ToLast but because both have the same value a combined enum class can be created:
```
    enum class Modes {
        Dimension,
        ThroughAll,
        ToLast = ThroughAll,
        ToFirst,
        ToFace,
        TwoDimensions
    };
```
with Dimension=0, ThroughAll=1, ToLast=1, ToFirst=2, ToFace=3, TwoDimensions=4. So, the important thing is that all enum labels still have the same value and ThroughAll has the same value as ToLast.

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
wwmayer: > I can change this to an enum. However I don't understand the term "blindness". I named the Boolean "isPocket", so it is clear what it is about.

It only becomes clear if somebody who reads the switches to the class declaration. And it's only clear if the value is true, but what if it's false? Somebody who is not familiar with the code won't know that it will be a pad then.

To better understand the issue have a look at: https://doc.qt.io/archives/qq/qq13-apis.html#thebooleanparametertrap

> I must admit I use bools because this is quicker ceded than adding an enum, since I write a procedure and in the IDE I can right-click on it to ad its definition in the header file.

But you don't write the code only for you but for everybody else. And thus the client code should be as clear as possible so that there is no need to always jump between client code and class declaration.

Then with a boolean you only have two cases but what if in the future a new case will be added? Do you then add another boolean? An example of this bad practice is https://github.com/FreeCAD/FreeCAD/blob/f1d6c78cbe3c17a0b219e0f7b1443ec53463b80a/src/Mod/PartDesign/Gui/ReferenceSelection.h#L49 which in client code then e.g. looks like this https://github.com/FreeCAD/FreeCAD/blob/249c2c79705f3fa11c0f5c2d5d2c6fa0cfad4f3d/src/Mod/PartDesign/Gui/TaskDraftParameters.cpp#L215
Having six boolean parameters makes it really difficult to understand the code immediately.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5355">pipeline status</a> for feature branch [PR_5355](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5355). Pipeline [446453075](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/446453075) was triggered at [f473f3b](https://github.com/FreeCAD/FreeCAD/pull/5355/commits/f473f3b4b6c1dfae60610b80a82fbe989da9bac1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
donovaly: >>Through all gets the "through all" length and this is longer than when to the last face. Therefore the 2 enums are different.
>>
>> Maybe I have a think here? If not, is there another way?

;-) well, this was already my thinko. That the modes _do_ something different does not mean they need a different index.
So yes, sure, the enum can be sorted out to the helper class too. I updated the PR accordingly and I think the PR should now be ready to merge.

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
donovaly: > To better understand the issue have a look at: https://doc.qt.io/archives/qq/qq13-apis.html#thebooleanparametertrap

Thanks, interesting article.

> An example of this bad practice is
> 
> https://github.com/FreeCAD/FreeCAD/blob/f1d6c78cbe3c17a0b219e0f7b1443ec53463b80a/src/Mod/PartDesign/Gui/ReferenceSelection.h#L50

I see.
I think the point is that everybody uses an IDE with proper information delivery. For example here with MSVC I just have too move the mouse over a 'true' and get its type and when I want also the whole procedure definition. But it is exactly this feature doesn't make you think about code readability for those without a proper IDE. I'll try to think about this more.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5355">pipeline status</a> for feature branch [PR_5355](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5355). Pipeline [447017160](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/447017160) was triggered at [513420d](https://github.com/FreeCAD/FreeCAD/pull/5355/commits/513420d0deef0fb9b93f99ffeca0bc37ba62e62e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
wwmayer: An IDE doesn't help at all when reading the code on-line.

src/Mod/PartDesign/Gui/TaskPadParameters.cpp
```diff
@@ -80,70 +80,8 @@ void TaskPadParameters::updateUI(int index)
...
-    ui->lengthEdit2->setEnabled( isLengthEdit2Visible );
-    ui->labelLength2->setVisible( isLengthEdit2Visible );
-
-    ui->buttonFace->setEnabled( isFaceEditEnabled );
-    ui->lineFaceName->setEnabled( isFaceEditEnabled );
-    if (!isFaceEditEnabled) {
-        onButtonFace(false);
-    }
+    // set and enable checkboxes
+    setCheckboxes(index, static_cast<Modes>(index));
```
wwmayer, line 84: This doesn't make any sense. Why is index passed twice now?

In my code example the second argument was supposed to be new enum class type:
```
enum class Type {
  Pad,
  Pocket
};```

The called function would then be: `setCheckboxes(index, Type::Pad);` and of course the signature of setCheckboxes is: `void setCheckboxes(int, Type);`

File: src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp [hunk 4]
```diff
@@ -389,6 +389,82 @@ void TaskExtrudeParameters::addAxisToCombo(App::DocumentObject* linkObj, std::st
...
+        isOffsetEditVisible = true;
+        isOffsetEditEnabled = false; // offset may have some meaning for through all but it doesn't work
+        isMidplaneEnabled = true;
+        isReversedEnabled = !ui->checkBoxMidplane->isChecked();
+    }
+    else if (mode == Modes::ToFirst) {
+        isOffsetEditVisible = true;
+        isReversedEnabled = true;
+    }
+    else if (mode == Modes::ToLast) {
```
wwmayer, line 424: ToLast is equal to ThroughAll and thus this else block is dead code. So, what's still needed is the extrude type (Pad or Pocket) to make sure ThroughAll is only handled for Pockets and ToLast only for Pads.

wwmayer:
See 66864bbe7b

src/Mod/PartDesign/Gui/TaskPadParameters.cpp
donovaly: > This doesn't make much sense. Why is index passed twice now?

Good questions. I have no clue why I force-pushed this. Before my last push the enum was correctly passed.

I see now that you already fixed and merged. Many thanks.

PR #5357, opened 2022-01-08
[PD] add feature to create tapered Pad / Pocket

donovaly:
This PR adds support for tapered extrudes.

This was from time to time requested in the forum since other CAD programs and also Pad's Extrusion provides this feature.

Forum thread to discuss it: https://forum.freecadweb.org/viewtopic.php?f=17&t=65124

Here is it in action:
FreeCAD_phD8DANHBv

The PR treats Pad and Pocket the same way, so Pocket has the same feature:
FreeCAD_fzqdB1cbDr


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [443915890](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443915890) was triggered at [9148d80](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/9148d8024a1e6c4ca7ac0ba2e9412092a870b76a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [443984655](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/443984655) was triggered at [8a85e15](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/8a85e15cb16c7a3a4703c3928aa86a0524e4c869). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@berndhahnebach, I made another commit to fix the CI build failure, can you therefore please trigger another CI run on this PR?

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -666,6 +705,154 @@ void ProfileBased::generatePrism(TopoDS_Shape& prism,
...
+    else {
+        throw Base::RuntimeError("Only a wire or a face is supported");
+    }
+
+    if (sourceWire.IsNull()) {
+        throw Base::RuntimeError("Offset for base shape could not made");
+    }
+
+    // calculate lateral offset and vectors for top and bottom shape of the loft
+    double lateralOffset = tan(angle * M_PI / 180.0) * Length;
```
jbaehr, line 746: I propose `Base::toRadians(angle)` instead of `angle * M_PI / 180.0`

donovaly: Thanks, I will change this.
At first the PR must compile for the CI.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [444014761](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/444014761) was triggered at [c59a1ba](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/c59a1baf7f4a3bda8de33e4c008675ffa3fede5c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -666,6 +705,154 @@ void ProfileBased::generatePrism(TopoDS_Shape& prism,
...
+        xp.Next();
+    }
+    // Pad / Pocket does not only support closed wires, thus only 1 edge is not valid
+    // if this might be supported in future, see Extrusion::makeDraft how to handle 1 edge
+    if (numEdges == 1)
+        throw Base::RuntimeError("Base shape consists of only one edge");
+
+    // to store the sections for the loft
+    std::list<TopoDS_Wire> list_of_sections;
+    BRepBuilderAPI_MakeWire sectionResult;
```
jbaehr, line 766: Without precompiled headers, this requires an `# include <BRepBuilderAPI_MakeWire.hxx>` here on apple-clang-11.0

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -666,6 +705,154 @@ void ProfileBased::generatePrism(TopoDS_Shape& prism,
...
+    }
+
+    // since the order is important store the other sections after bottom
+    list_of_sections.push_back(sourceWire);
+    // create the top section
+    createTaperedPrismOffset(sourceWire, vecOffset, lateralOffset, Length, false, sectionResult);
+    list_of_sections.push_back(sectionResult);
+    
+    // build the loft using the created sections
+    BRepOffsetAPI_ThruSections mkGenerator(Standard_True, /*ruled=*/Standard_True);
```
jbaehr, line 782: Without precompiled headers, this requires an `# include <BRepOffsetAPI_ThruSections.hxx>` here on apple-clang-11.0

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -666,6 +705,154 @@ void ProfileBased::generatePrism(TopoDS_Shape& prism,
...
+    gp_Trsf tempTransform;
+    tempTransform.SetTranslation(translation);
+    TopLoc_Location loc(tempTransform);
+    TopoDS_Wire movedSourceWire = TopoDS::Wire(sourceWire.Moved(loc));
+
+    TopoDS_Shape offsetShape;
+    if (fabs(offset) > Precision::Confusion()) {
+        TopLoc_Location wireLocation;
+        TopLoc_Location edgeLocation;
+        BRepOffsetAPI_MakeOffset mkOffset;
```
jbaehr, line 818: Without precompiled headers, this requires an `# include <BRepOffsetAPI_MakeOffset.hxx>` here on apple-clang-11.0

jbaehr (changes requested):
There seem to be some missing headers when not using `_PreComp_`.

src/Mod/PartDesign/App/FeatureSketchBased.cpp
donovaly: thanks, added

src/Mod/PartDesign/App/FeatureSketchBased.cpp
donovaly: thanks, added

src/Mod/PartDesign/App/FeatureSketchBased.cpp
donovaly: thanks, added

donovaly:
@berndhahnebach I need another CI run please
What compiler is run be the CI? Does it use precompiled headers?

src/Mod/PartDesign/App/FeatureSketchBased.cpp
donovaly: done now

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [444168620](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/444168620) was triggered at [3413c5e](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/3413c5e921151d41fbadad7eee4623295c374f59). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureExtrude.h
```diff
@@ -24,12 +24,17 @@
...
+#include <App/PropertyGeo.h>
 #include <App/PropertyStandard.h>
 #include <App/PropertyUnits.h>
+#include <Base/Tools.h>
 #include "FeatureSketchBased.h"
 
 namespace PartDesign
 {
+    static const App::PropertyQuantityConstraint::Constraints signedLengthConstraint = { -DBL_MAX, DBL_MAX, 1.0 };
+    static const double maxAngle = 90 - Base::toDegrees<double>(Precision::Angular());
```
jbaehr, line 36: By moving this from the .cpp to the .h file (compared to the [initial commit](https://github.com/FreeCAD/FreeCAD/commit/821b330722a361e71632fa49a17e2cc76e1e6e69)), you now need `#include <Precision.hxx>` in the header, too. Otherwise it fails to compile here using apple-clang-11.0

donovaly: OK, so not it is clear to me that the CI does not use precompiled headers. Since I do I don't see the compilation failures.
Now it should work, please report back.

donovaly:
@berndhahnebach give me another CI run please

I realized now that the CI does not use precompiled headers. Since I do, I don't see the compilation failures due to missing headers.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [444253172](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/444253172) was triggered at [3e8adc1](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/3e8adc1abdbd089dc40f6637df42a49b811664c5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [444427600](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/444427600) was triggered at [6bcefcf](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/6bcefcf6b8377bac5e4872a58e8f1857f970280e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureExtrude.cpp
```diff
@@ -51,18 +51,20 @@
...
 
 #include "FeatureExtrude.h"
 
 using namespace PartDesign;
 
-
 PROPERTY_SOURCE(PartDesign::FeatureExtrude, PartDesign::ProfileBased)
 
 App::PropertyQuantityConstraint::Constraints FeatureExtrude::signedLengthConstraint = { -DBL_MAX, DBL_MAX, 1.0 };
+const double maxAngle = 90 - Base::toDegrees<double>(Precision::Angular());
```
wwmayer, line 63: Avoid global variables. You can declare it as a static variable and move it to the constructor of FeatureExtrude.

src/Mod/PartDesign/App/FeatureExtrude.h
```diff
@@ -41,13 +43,16 @@ class PartDesignExport FeatureExtrude : public ProfileBased
...
+    App::PropertyAngle       TaperAngle;
+    App::PropertyAngle       TaperAngle2;
     App::PropertyBool        UseCustomVector;
     App::PropertyVector      Direction;
     App::PropertyBool        AlongSketchNormal;
     App::PropertyLength      Offset;
     App::PropertyLinkSub     ReferenceAxis;
 
     static App::PropertyQuantityConstraint::Constraints signedLengthConstraint;
+    App::PropertyAngle::Constraints floatAngle;
```
wwmayer, line 55: Should be static as it's equal for all instances of pad/pocket.

src/Mod/PartDesign/App/FeaturePad.cpp
```diff
@@ -75,6 +75,10 @@ Pad::Pad()
     ADD_PROPERTY_TYPE(UpToFace, (0), "Pad", App::Prop_None, "Face where pad will end");
     ADD_PROPERTY_TYPE(Offset, (0.0), "Pad", App::Prop_None, "Offset from face in which pad will end");
     Offset.setConstraints(&signedLengthConstraint);
+    ADD_PROPERTY_TYPE(TaperAngle, (0.0), "Pad", App::Prop_None, "Taper angle");
+    TaperAngle.setConstraints(&floatAngle);
+    ADD_PROPERTY_TYPE(TaperAngle2, (0.0), "Pocket", App::Prop_None, "Taper angle for 2nd direction");
```
wwmayer, line 80: Should be "Pad", not "Pocket"

donovaly:
> The main question to me is do we really need this feature directly for pad/pockets?

Sure, otherwise I would not have spent so much time on this ;-).
Since I am active in FC people ask why FC lacks features so-called "CAD standard software" provides. They refer to Fusion 360:
https://help.autodesk.com/view/fusion360/ENU/?guid=SLD-EXTRUDE-SOLID
and Fusion appears indeed to be a standard in the maker scene. As I implemented it, it works almost the same as in Fusion.

In my working field Solidworks is used by most companies around and this also provides the feature:
https://help.solidworks.com/2019/english/SolidWorks/sldworks/r_extrude_propertymanager.htm
its workflow is close to the of Fusion.

You spoke about a "god" class, but I only implemented what part Extrude for years. And I could understand the users who asked why Part has a basic feature that PD does not provide.
(As you can see in Solidworks, there it is a real "god" feature because it also includes the Thickness feature.
Writing this after a full day of real-life construction work, I understand that speed matters and when you can setup a feature in one dialog and later change it in the tree in one place, saves time. I guess SW designed the UI as they did to speedup the workflow.)

Therefore implementing this was for a long time on my Todo-list. While working on it during the whole last week, I realized its power. One can save a lot of time in constructing because one saves the work to make lofts.
(Drafts are no option for me for real-life document since all the time the face numbers change and then the drafts go to the wrong faces.)
The last hour I took the freedom to use this PR on a real-life document and play around. So despite I was skeptic about such a feature for a long time (you can e.g ask chrisB who tried for a while to change my mind ;-)), I see its power now. Any having the same functionality in Part and PD also helps.

Here is the thread that eventually triggered my action: https://forum.freecadweb.org/viewtopic.php?f=3&t=64795&start=30

p.s. no fear. I am not planning to include the thickness feature to Pad/Pocket.

src/Mod/PartDesign/App/FeatureExtrude.h
donovaly: i know but could not get it to work. Either it works for me with MSVC then, I got a build failure of the CI. When the CI succeeds, I get a linking mistake. it cost me a while to realize that the definition of maxAngle causes this. When I use in maxAngle just a plain number, e.g.
"89.999"
then everything is fine, but as soon as I try to use
"90 - Base::toDegrees<double>(Precision::Angular()"
I got the problems.

I'll try later today if I find a solution that pleases the linker of MSVC and the CI compiler.

src/Mod/PartDesign/App/FeatureExtrude.h
```diff
@@ -41,13 +43,16 @@ class PartDesignExport FeatureExtrude : public ProfileBased
     App::PropertyEnumeration Type;
     App::PropertyLength      Length;
     App::PropertyLength      Length2;
+    App::PropertyAngle       TaperAngle;
+    App::PropertyAngle       TaperAngle2;
```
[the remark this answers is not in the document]
donovaly: I used the same naming scheme as we already have:
length   - length in normal direction
length2 - length in opposite direction

Thus taper is for normal, taper2 for opposite direction. I can change it to a different name, but then it won't fit to the length/length2 scheme. What is your opinion or proposal?

src/Mod/PartDesign/App/FeatureExtrude.cpp
donovaly: Thanks. I made it now a stat class member.

src/Mod/PartDesign/App/FeatureExtrude.h
donovaly: OK, I found a solution, it is now a static member.

src/Mod/PartDesign/App/FeaturePad.cpp
donovaly: Thanks, fixed now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [446453930](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/446453930) was triggered at [e6b6f64](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/e6b6f641a9041954e58db6b5311519042ab2eab0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> and Fusion appears indeed to be a standard in the maker scene. As I implemented it, it works almost the same as in Fusion.
>...
> In my working field Solidworks is used by most companies around and this also provides the feature:

Well, if it's considered as state of the art I can accept the functionality but this is independent on how it's implemented.

For example, to avoid to put all possible options into a single feature class one could split things into one or several DocumentObjectExtension classes or just normal helper classes. This is a bit of extra work but in the end pays off because it reduces code duplication and makes code more reusable.

> You spoke about a "god" class, but I only implemented what part Extrude for years.

God class is one of many possible code smells (see https://de.wikipedia.org/wiki/Code_smell) and it is independent from the functionality of an application that the end user sees.

Especially when looking at your changes of ProfileBased::generatePrism() there are now another two arguments (and the eight we had before were already more than enough) but inside the implementation there is a big if-block where all the new code goes to and the else-block contains the old code. This doesn't only make the function fatter by a factor of 3 but also complicates the its use for possible new features because having ten parameters it's not easy to figure out the valid values.

All this could be easily avoided by creating a new function.

The second thing is that ProfileBased::generateTaperedPrism() is almost identical to the code of Extrusion::makeDraft(). So again a big block of duplicated code that should be avoided. This can be fixed by moving the code to a helper class that can be reused by Part::Extrusion and PartDesign::FeatureExtrude.

donovaly:
> Especially when looking at your changes of ProfileBased::generatePrism() there are now another two arguments 
> 
> The second thing is that ProfileBased::generateTaperedPrism() is almost identical to the code of Extrusion::makeDraft(). So again a big block of duplicated code that should be avoided. This can be fixed by moving the code to a helper class that can be reused by Part::Extrusion and PartDesign::FeatureExtrude.

I agree with both. In the first case I would even go away from making a prism at all as the existing comment where the prism is made suggests. I asked for this in my initial PR forum post but got no reply yet.

For the second case I would have done that initially but recently I asked you on reusing code form the FEM WB to detect Gmsh to avid having two routines doing the same and you replied that WBs should not share code.
Therefore I treated here the taper code in PD only, despite it is just the existing code from the Part WB.
However, I will change the PR now accordingly. Just one question, where should the helper class resides? Directly in App?

p.s. I have to travel the next days, so I won't have time the next week to update this PR.

wwmayer:
> For the second case I would have done that initially but recently I asked you on reusing code form the FEM WB to detect Gmsh to avid having two routines doing the same and you replied that WBs should not share code.

Because FEM and Mesh don't have a dependency and the function from FEM doesn't justify to add a dependency.
However, Part contains a lot of core functionality of PartDesign and there it's recommended not to duplicate code.

> Just one question, where should the helper class resides? Directly in App?

The helper class would be under src/Mod/Part/App.

wwmayer:
> p.s. no fear. I am not planning to include the thickness feature to Pad/Pocket.

What a pity! It's no joke: [https://forum.freecadweb.org/viewtopic.php?p=561879#p561879](Feature request thin element for Part and PartDesign workbench)

donovaly:
> The helper class would be under src/Mod/Part/App.

Thanks. I will implement this the next days.

I will also take care of the different pending PRs to act in a planned matter:
- #5379 seems the be a suitable OCC building method -> should be used at first for PD Loft
- this PR will be remodeled for the helper class so that PD Pad/Pocket and Part Extrude share the same code
- eventually there will be a further PR to add the functionality that Extrude can handle inner wires

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -665,6 +708,154 @@ void ProfileBased::generatePrism(TopoDS_Shape& prism,
...
+
+    // create a new shape which is the input shape with the offset defined by the TaperAngle
+    // at first inspect the from shape
+    int numEdges = 0;
+    TopExp_Explorer xp(sourceWire, TopAbs_EDGE);
+    while (xp.More()) {
+        numEdges++;
+        xp.Next();
+    }
+    // Pad / Pocket does not only support closed wires, thus only 1 edge is not valid
```
davidosterberg, line 762: does only

adrianinsaval: Wait a second, what if I'm padding a circle? wouldn't that be just one edge but still closed? 1 edge should be valid, and isn't the shape already validated to be closed before this?

donovaly: > what if I'm padding a circle?

This is a special case because OCC has a bug in making offsets of circles.

However, this PR has to be remodeled to take the code from Part Extrude. -> I'll put it on hold until I found the time to write the helper class to achieve this.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [454035627](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454035627) was triggered at [76119f4](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/76119f4f536af0cdeb542ff35b4cf094f779d407). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> > The helper class would be under src/Mod/Part/App.
> 
> Thanks. I will implement this the next days.

Done. @wwmayer , can you please have it a look that I did it right? For example I kept the function in the helper class in the Part namespace, but maybe it is better not to include it to any namespace?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [454241669](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454241669) was triggered at [0526763](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/0526763ce5a86c76b17fcf24aebbf917d89dfd9e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@berndhahnebach , can you please give me a further CI run please? (I am struggling here again that I use precompiled headers and thus don't see the missing includes.)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [454359135](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454359135) was triggered at [1861c50](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/1861c506d874d2dc9e8f219d689b6e44fbec9d0d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/ExtrusionHelper.h
```diff
@@ -0,0 +1,54 @@
...
+
+class PartExport ExtrusionHelper
+{
+public:
+    ExtrusionHelper();
+
+    /**
+     * @brief makeDraft: creates a drafted extrusion shape out of the input 2D shape
+     */
+    static void makeDraft(const TopoDS_Shape& shape,
```
wwmayer, line 42: IMO, the class ExtrusionParameters in the original code was a good thing because it kept the list of arguments very small. Thus, I recommend to move the class ExtrusionParameters from Extrusion to this class but I would drop the "Extrusion" and just call it Parameters.

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -628,12 +636,56 @@ void ProfileBased::generatePrism(TopoDS_Shape& prism,
                 throw Base::ValueError("Cannot create a pocket with a depth of zero.");
         }
 
-        // Its better not to use BRepFeat_MakePrism here even if we have a support because the
-        // resulting shape creates problems with Pocket
-        BRepPrimAPI_MakePrism PrismMaker(from, Ltotal * gp_Vec(dir), 0, 1); // finite prism
-        if (!PrismMaker.IsDone())
-            throw Base::RuntimeError("ProfileBased: Length: Could not extrude the sketch!");
-        prism = PrismMaker.Shape();
+        if (fabs(angle) > Precision::Angular() || fabs(angle2) > Precision::Angular()) {
```
wwmayer, line 639: There is still no point in having all this inside generatePrism(). Revert the changes and add two additional functions:
```
bool hasTaperedAngle(double angle1, double angle2) const {
   return fabs(angle) > Precision::Angular() || fabs(angle2) > Precision::Angular();
}
```
```
void ProfileBased::generateTaperedPrism(...)
{
   // move all the code from inside the if-block
}
```

The code in Extrusion::execute() then will become:
```
if (hasTaperedAngle(TaperAngle.getValue(), TaperAngle2.getValue())) {
    generateTaperedPrism(...);
}
else {
    generatePrism(...);
}
```

src/Mod/PartDesign/App/FeatureSketchBased.cpp
```diff
@@ -628,12 +636,56 @@ void ProfileBased::generatePrism(TopoDS_Shape& prism,
...
-        prism = PrismMaker.Shape();
+        if (fabs(angle) > Precision::Angular() || fabs(angle2) > Precision::Angular()) {
+            // we have to make tapered prism
+
+            if (reversed)
+                dirTaper.Reverse();
+ 
+            std::list<TopoDS_Shape> drafts;
+            bool isSolid = true; // in PD we only generate solids, while Part Extrude can also create only shells
+            if (method == "ThroughAll")
```
wwmayer, line 647: In order to further reduce the size of this function the determination of the parameters for the tapered prism (L, L2, angle, and angle2) could be moved to a new function.
```
void ProfileBased::determineTaperedPrismParameters(const std::string& method, bool midplane, double& L, double& L2, double& Angle, double& Angle2)
{
    if (method == "ThroughAll") {
      L = getThroughAllLength();
      L2 = 0.0;
      Angle2 = 0.0;
    }
    else if (method == "TwoLengths") {
       // nothing to do
    }
    else if (method == "Length") {
        if (midplane) {
            L = L / 2;
            L2 = L;
            Angle2 = Angle;
        }
        else {
            L2 = 0.0;
            Angle2 = 0.0;
        }
    }
}
```
And generateTaperedPrism() will become
```
void ProfileBased::generateTaperedPrism(...)
{
...
    Part::ExtrusionHelper::makeDraft(sketchshape, dirTaper, L, L2,
                                                         Base::toRadians(angle), Base::toRadians(angle2), isSolid, drafts);
    if (drafts.empty()) {
...
    }
...
}
```

src/Mod/Part/App/ExtrusionHelper.h
donovaly: Yes and I wanted to keep this. I failed because in Line 648 ff. I need to call the Part::ExtrusionHelper::makeDraft in 4 different ways. 
Therefore the many arguments are looking weird but I don't know a better way. Do you have an idea?

src/Mod/PartDesign/App/FeatureSketchBased.cpp
donovaly: > There is still no point in having all this inside generatePrism(). Revert the changes and add two additional functions:

Thanks, I will do so.

src/Mod/PartDesign/App/FeatureSketchBased.cpp
donovaly: > In order to further reduce the size of this function the determination of the parameters for the tapered prism (L, L2, angle, and angle2) could be moved to a new function.

My new plan is to do this as Part Extrude does: check all parameters then store them in an enum. The enum is then input to either the draft prism or straight prism creation void.

I write new plan because since the community (forum) wants the taper feature to handle from the beginning also inner wires, I first need to merge PR #5367. This PR is from my point of view ready to be merged. I need to adapt the PR here then to the new part code.
Therefore it would help me when you could also review PR #5367.

src/Mod/PartDesign/App/FeatureSketchBased.cpp
jbaehr: Please note that `Precision::Angular()` returns a limit in radians, you seem to compare it with degrees.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [455773677](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/455773677) was triggered at [2b11bb1](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/2b11bb109737fb8d5426209e590047ed3dbadfc9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks @wwmayer  for the review. I refactored the code as you suggested but in a slightly different way:
I imagined how I would read the code as outsider. So I
- first Extrude
- in the Extrude function the given shape parameters are first evaluated
- then it is known what shape has to be created, either a linear or tapered shape

The term"prism" was misleading, because only the liner shape will become a prism. I think the code is now more logic and better understandable

src/Mod/PartDesign/App/FeatureSketchBased.cpp
donovaly: > Please note that Precision::Angular() returns a limit in radians, you seem to compare it with degrees.

Good catch! Fixed now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [456431037](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/456431037) was triggered at [81567f7](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/81567f7e39c36874ba95b511eda18b85c1c6e133). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [457435756](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/457435756) was triggered at [5c0460f](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/5c0460f0bfbe5f1a7a446d4e6eaa41040bf917aa). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [458320633](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/458320633) was triggered at [f62bcc0](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/f62bcc09188313042671723afe314b10efe6de63). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [459054342](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/459054342) was triggered at [ed777ef](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/ed777ef3cdeac9567fe5e69f2b126b1ead4229cf). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [459527778](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/459527778) was triggered at [7a2bb51](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/7a2bb5178424c68b94b54addf934d960950e6bf7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
OK, this PR is now adapted to current master. The functionality nothing has changes, the PR inherits Part's Extrude code for tapered extrusion. However, as wished in the forum, it does now support inner structures as Part does since today.

Anyone, please give it a try.
@mwganson

mwganson:
Looks good.  We even get cones and planes for the faces.

Only thing I got an error message opening ExtrusionHelper.cpp:
Snip macro screenshot-5dfe96

mwganson:
That was opening in QtCreator.

donovaly:
> Only thing I got an error message opening ExtrusionHelper.cpp: Snip macro screenshot-5dfe96

Thanks. I hope this encoding issue is now fixed. Can you please try again?

mwganson:
That seems to have fixed it.

donovaly:
> That seems to have fixed it.

Thanks.

I found meanwhile a minor issue, fixed this and rebased to trigger the CI.

donovaly:
Here is a Windows build with the PR on top of the current master to try out the PR without the need to compile:

https://github.com/donovaly/FreeCADInstProj/releases/tag/0.20r27327%2B1

jbaehr:
@donovaly Thanks for your work here!

I have a general question though: From a brief look at the diffs I get the impression that the `Extrude` (and its private `generate...Prism`) methods of FeatureSketchBased are only used in Pad and Pocket. And those have their own base in between FeatureSketchBased: FeatureExtrude. Wouldn't it be cleaner to move the code there?

donovaly:
> I have a general question though: From a brief look at the diffs I get the impression that the `Extrude` (and its private `generate...Prism`) methods of FeatureSketchBased are only used in Pad and Pocket.

It seems so. However, any further code shifting should be done in a separate PR. This PR here is already huge and the important thing is the new functionality of tapered extrudes. Therefore the focus is on that code part.

Later I plan to shift more code around, for example Part and PD have the same features to pad along a certain direction and this code can later be put into the new helper file I introduced with this PR.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [460324278](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/460324278) was triggered at [730b112](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/730b1123a5c83240277cd61c1bb4aab0513b67d2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357">pipeline status</a> for feature branch [PR_5357](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5357). Pipeline [460535351](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/460535351) was triggered at [10375db](https://github.com/FreeCAD/FreeCAD/pull/5357/commits/10375db8563a902067086965e3bcf142ac7c6d79). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
OK, I think it is time to make the move to merge.

PR #5358, opened 2022-01-08
PartDesign: Silence noisy FeatureTransformed

davidosterberg:
Remove several report view messages that are not necessary for everyday use.

By request from the forum: 
https://forum.freecadweb.org/viewtopic.php?f=3&t=64900&start=20#p558962


donovaly:
Thanks!

PR #5362, opened 2022-01-09
[PD] Pad/Pocket: make a variable static

donovaly:
as suggested in https://github.com/FreeCAD/FreeCAD/commit/e2f883fcc4d6708bc8fc30aa19b827dc6a9f8a1d#r63143914


wwmayer:
No. Look at: f1a265479d

signedLengthConstraint still should be a class member but it's initialized then in the source file.

PR #5367, opened 2022-01-09
[Part] handle inner wires for Extrude

donovaly:
this is a PR to make Extrude handle inner wires of sketches.

Forum thread for this PR: https://forum.freecadweb.org/viewtopic.php?f=10&t=65171

### DIFF
--- a/src/Mod/Part/App/FeatureExtrusion.cpp
+++ b/src/Mod/Part/App/FeatureExtrusion.cpp
@@ -26,14 +26,21 @@
 # include <cmath>
 # include <BRepAdaptor_Surface.hxx>
 # include <BRepAdaptor_Curve.hxx>
+# include <BRepAlgoAPI_Cut.hxx>
 # include <BRepBuilderAPI_Copy.hxx>
+# include <BRepBuilderAPI_MakeFace.hxx>
+# include <BRepBuilderAPI_MakeSolid.hxx>
 # include <BRepBuilderAPI_MakeWire.hxx>
+# include <BRepBuilderAPI_Sewing.hxx>
+# include <BRepClass3d_SolidClassifier.hxx>
+# include <BRepGProp.hxx>
 # include <BRepLib_FindSurface.hxx>
 # include <BRepOffsetAPI_MakeOffset.hxx>
 # include <BRepOffsetAPI_ThruSections.hxx>
 # include <BRepPrimAPI_MakePrism.hxx>
 # include <gp_Pln.hxx>
 # include <gp_Trsf.hxx>
+# include <GProp_GProps.hxx>
 # include <Precision.hxx>
 # include <ShapeAnalysis.hxx>
 # include <ShapeFix_Wire.hxx>
@@ -45,14 +52,12 @@
 #endif
 
 #include "FeatureExtrusion.h"
+#include <App/Application.h>
 #include <Base/Tools.h>
 #include <Base/Exception.h>
 #include "Part2DObject.h"
 
-
-
 using namespace Part;
-
 
 PROPERTY_SOURCE(Part::Extrusion, Part::Feature)
 
@@ -315,7 +320,6 @@ TopoShape Extrusion::extrudeShape(const TopoShape& source, const Extrusion::Extr
     if (result.IsNull())
         throw NullShapeException("Result of extrusion is null shape.");
     return TopoShape(result);
-
 }
 
 App::DocumentObjectExecReturn* Extrusion::execute(void)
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [444279336](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/444279336) was triggered at [2aed4ae](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/2aed4ae413c4717b3fe90b40c714d2079c7772d7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

mwganson:
I needed to add these includes to get it to compile.  I can't copy/paste text out of the VM, so I took a screenshot.

Snip macro screenshot-5f981e

donovaly:
> I needed to add these includes to get it to compile. 

Thanks, now fixed

donovaly:
I updated the PR to the method that was successful for PartDesign.

This still fails bit there is only one issue we need to resolve:
When you have a sketch with inner wires, the location of the inner wires is wrong. On making the offset, the location is correctly transformed the bug is that the initial location of the shape when making the offset is actually not the location of the shape in the document.

I fail to find out why this happens not how to restore the correct location.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [444427604](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/444427604) was triggered at [3b92b01](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/3b92b0152187a681a6d11c89ad04effbe9be0f46). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I think we are facing a OCC bug:
https://dev.opencascade.org/content/brepoffsetapimakeoffset-wire-and-face-odd-occt-740

I use OCC 7.5.3 and whenever I create an offset of an inner wire, the result is placed quite random in space. I tried to store the location of the wires before I offset them to reset the offset wire later to the original position but cannot get this to work. Meaning, whatever i store, it is not the position of the wire.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [446453933](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/446453933) was triggered at [39ef22a](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/39ef22a01dd2d412d85d7d14dee0b0a5d22ee37d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
Example file?

donovaly:
> I think we are facing a OCC bug: https://dev.opencascade.org/content/brepoffsetapimakeoffset-wire-and-face-odd-occt-740
> 
> I use OCC 7.5.3 and whenever I create an offset of an inner wire, the result is placed quite random in space. 

I am stupid. This is a known bug in OCC and there is already a workaround in the code for this:
https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Part/App/FeatureExtrusion.cpp#L375

In the PR I removed this workaround and clear that it then fails.

donovaly:
> I am stupid. This is a known bug in OCC and there is already a workaround in the code for this: https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Part/App/FeatureExtrusion.cpp#L375
> 
> In the PR I removed this workaround and clear that it then fails.

I am not that stupid. The case if s single-edge wire that is handled there is also handled by the PR. But what is not handled are in general circular inner wires.

So the PR does what it should - as long as you don't have inner circles. This is caused by the OCC bug.

I hope I'll find a workaround.

donovaly:
> I hope I'll find a workaround.

I found it: Single-edge wires (circles) must be treated different and also different when being an outer or an inner wire
So the PR works now eventually!

Here 2 screencasts showing this:

FreeCAD_VRzMP3esJg
FreeCAD_3peSPJqXL4

donovaly:
The PR is now ready to be stress-tested

0penBrain:
Shouldn't this be discussed with community?

donovaly:
> Shouldn't this be discussed with community?

I linked the discussion thread in the initial post of this PR.

Besides this, if you have time, please review the PR and test it out. The plan is to merge this PR as first step in a series. When this is merged, I will sort our the extrusion code to a helper class that Part and PD can take the same code. Then @davidosterberg 's proposal can go in to change the loft creation mechanism (PR #5379) can be rewritten to affect the new helper class and I will remodel PR #5357 to use the new helper class as well.

0penBrain:
> > Shouldn't this be discussed with community?
> 
> I linked the discussion thread in the initial post of this PR.

See that, but I guess no normal user really reads in the "Developer corner" subforum. ;) Anyway this was just a remark, I think this evolution can really harm.

> Besides this, if you have time, please review the PR and test it out.
Not really a lot of time ATM, so don't expect too much from me on this. Will write here if I find time to review. ;)

adrianinsaval:
Did you mean _can't_ really harm? I think it might break old files that used extrude with taper with inner wires but I think it's worth the breaking (how many of them are there anyways?). I'll see if I can compile an test that next week if it's not merged yet.

donovaly:
> Did you mean _can't_ really harm? I think it might break old files that used extrude with taper with inner wires

Extrudes with inner wires never worked since the result are several 3D bodies stacked into each other. Or you can say that just the shape of the outer wire was in effect used for the outermost and thus really resulting shape.
With this PR it will be the first time that extrudes with inner wires actually work. Therefore I don't think that anything can be broken.

src/Mod/Part/App/FeatureExtrusion.cpp
```diff
@@ -337,154 +342,316 @@ App::DocumentObjectExecReturn* Extrusion::execute(void)
 
 void Extrusion::makeDraft(const ExtrusionParameters& params, const TopoDS_Shape& shape, std::list<TopoDS_Shape>& drafts)
 {
-    double distanceFwd = tan(params.taperAngleFwd) * params.lengthFwd;
-    double distanceRev = tan(params.taperAngleRev) * params.lengthRev;
+    std::vector<std::vector<TopoDS_Shape>> wiresections;
+
+    auto addWiresToWireSections =
+        [&shape, &wiresections](std::vector<std::vector<TopoDS_Shape>>& wiresections) -> size_t {
```
adrianinsaval, line 348:
```suggestion
        [&shape](std::vector<std::vector<TopoDS_Shape>>& wiresections) -> size_t {

```
because I got the following build error:
```
FreeCAD/src/Mod/Part/App/FeatureExtrusion.cpp:348:73: error: lambda parameter ‘wiresections’ previously declared as a capture
  348 |         [&shape, &wiresections](std::vector<std::vector<TopoDS_Shape>>& wiresections) -> size_t {
      |                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
```

blindly following my uninformed interpretation of what is said here: https://stackoverflow.com/questions/61559765/checking-presence-of-an-element-in-a-list-using-2-params-and-a-lambda-function-i
I did this change and it looks like it compiled that file, no clue if it's still doing what's intended like this since I don't even understand what lambda parameters, capture and all those things mean. No clue if it will actually finish compiling either but I want to go to sleep now and my shitty notebook will still take a long time to finish compiling so I'm just going to leave this comment now hoping it's useful.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [454040375](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454040375) was triggered at [085d9e2](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/085d9e28fd0139bca9f7c78c76bc01660fc90076). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/FeatureExtrusion.cpp
donovaly: > because I got the following build error:

Thanks. That's why i need the CI. I use MSVC and this did not issue even a warning despite the obvious mistake "lambda parameter 'wiresections' previously declared as a capture"

I fixed this now and rebased to trigger another CI run.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [454241671](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454241671) was triggered at [2b397ad](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/2b397ad570d1bc2e38022e07722432417fa6d198). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
> > Did you mean _can't_ really harm? I think it might break old files that used extrude with taper with inner wires
> 
> Extrudes with inner wires never worked since the result are several 3D bodies stacked into each other. Or you can say that just the shape of the outer wire was in effect used for the outermost and thus really resulting shape. With this PR it will be the first time that extrudes with inner wires actually work. Therefore I don't think that anything can be broken.

This file was first created with freecad 0.19.3 then Extrude002 and Extrude003 were recomputed with my build of your changes (from when I left that comment about the compile error)
Screenshot_20220121_002147
For some reason Extrude003 has a bugged face
Also, it's not obvious but you can see the ones that aren't solid have a different shape because in 0.19.3 inner and outer faces are both offset in the same direction while new one have them opposed. I like this behavior but this change in behavior should be noted.
[extrudeExample.zip](https://github.com/FreeCAD/FreeCAD/files/7921168/extrudeExample.zip)

donovaly:
> For some reason Extrude003 has a bugged face

Many thanks for this test case! It shows an interesting behavior depending on the extrusion length:
- with 10 mm the triangle is closed while the other 2 structures are open
- with 100 mm only the circle is open
- with 1000 mm only the triangle is open

-> there is obviously a rounding error for the shell height since when the height is just a tiny bit too small FaceMakerCheese cannot detect this shell as inner wire. I'll have a look

> Also, it's not obvious but you can see the ones that aren't solid have a different shape because in 0.19.3 inner and outer faces are both offset in the same direction while new one have them opposed. I like this behavior but this change in behavior should be noted.

This change was explicitly requested in the forum by several people. The rationale for this is sensible - tapered extrudes are made for molded parts of molds and to de-mold parts the inner taper must have the opposite angle than the outer ones. With the same angle for inner wires, the taper feature is quite useless for practice.
Of course the change will be documented. However, this only affects when extruding as non-solid since for solid inner wires never worked..

donovaly:
I found the issue but also that my whole concept fails. Taper-extruding such a shape must work too but cannot work with my concept:
FreeCAD_d_nPY9oKr4f6


I have an idea how to achieve this, but need more time -> setting this PR as WIP for now

donovaly:
OK, my idea worked:
FreeCAD_ILU1Ul6mRN

Please give this PR now another try.

donovaly:
OK, my idea worked:
FreeCAD_ILU1Ul6mRN

Please give this PR now another try.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [454557975](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454557975) was triggered at [8229fa7](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/8229fa7f3248e112bed491bccf915d99f6479a42). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [455229696](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/455229696) was triggered at [04541ae](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/04541ae526f3610f8c083dbfa231bac2a65e47ae). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [455414443](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/455414443) was triggered at [8f8e118](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/8f8e118ff6d7f4132a8775e16c39f0b64f5a0b0b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [456431052](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/456431052) was triggered at [0ef2be4](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/0ef2be4a554924f963637f0d53190627833a0d39). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [457435762](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/457435762) was triggered at [3c56db5](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/3c56db537fa1f36fc6ca2237d9e07aee3105619e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [458320636](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/458320636) was triggered at [6261622](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/626162252664df4bf85dc27ddc39dbe104518c54). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [459054345](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/459054345) was triggered at [c335de8](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/c335de8d52d3aa708a00f80a6b81967cf4d702af). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367">pipeline status</a> for feature branch [PR_5367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5367). Pipeline [459213926](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/459213926) was triggered at [af0d902](https://github.com/FreeCAD/FreeCAD/pull/5367/commits/af0d902506c72fa3a8ec5ea21149dc135f04e4b2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I tested the PR now every day, also with real-life documents in which I replaced PD Pas by Part Extrude.

It is time to make progress since 2 other PRs are depending on this. Since to my best knowledge the PR works and also since it does not change anything else than tapered extrusions -> merging it now
(I will also provide a new build that more users can test it.)

src/Mod/Part/App/FeatureExtrusion.cpp
```diff
@@ -337,154 +341,361 @@ App::DocumentObjectExecReturn* Extrusion::execute(void)
...
+        Standard_Failure::Raise("Extrusion: Input must not only consist if a vertex");
+
+    // to store the sections for the loft
+    std::list<TopoDS_Wire> list_of_sections;
+    std::vector<std::vector<TopoDS_Shape>> extrusionSections;
+
+    // we need for all found wires an offset copy of them
+    // we store them in an array
+    TopoDS_Wire offsetWire;
+    for (auto& wire : wiresections)
```
marioalexis84, line 383: There is a warning due to unused `wire` variable.

src/Mod/Part/App/FeatureExtrusion.cpp
```diff
@@ -337,154 +341,361 @@ App::DocumentObjectExecReturn* Extrusion::execute(void)
...
-        TopExp_Explorer xp(sourceWire, TopAbs_EDGE);
-        while (xp.More()) {
-            numEdges++;
-            xp.Next();
+    // now subtract them
+    // if the moment of inertia changes, we have an inner wire
+    GProp_GProps tempProperties;
+    Standard_Real momentOfInertiaInitial;
+    Standard_Real momentOfInertiaFinal;
+    std::vector<bool>::iterator itInner = isInnerWire.begin();
```
marioalexis84, line 413: Same here due to `itInner`.

src/Mod/Part/App/FeatureExtrusion.cpp
donovaly: I need `wire ` as counter. I don't know how to silence the compiler since I don't get this warning.

src/Mod/Part/App/FeatureExtrusion.cpp
donovaly: Thanks, fixed now in master.

src/Mod/Part/App/FeatureExtrusion.cpp
marioalexis84: Add `(void)wire;` inside the loop.
or use `std::generate` over the resized `extrusionSections` and fill it with a lambda :
`std::vector<std::vector<TopoDS_Shape>> extrusionSections(wiresections.size());`
` std::generate(extrusionSections.begin(), extrusionSections.end(), []{return std::vector<TopoDS_Shape>();});`

marioalexis84: Actually, no need the loop:
`std::vector<std::vector<TopoDS_Shape>> extrusionSections(wiresections.size(), std::vector<TopoDS_Shape>());`

donovaly: Thanks! Fixed with commit 8eab4a397e

PR #5378, opened 2022-01-13
[Part] (Loft, Extrude, Sweep, Revolve) allow to skip task di…

mwganson:
…alog when object properties can be set by inferring from the user selection.

For example, with a loft the user can select the profiles before clicking the Loft toolbar icon.  The loft will be done without showing the dialog.  If necessary, the user can set properties as desired afterwards for the Loft object.

No changes are made to the dialogs.  They are either skipped or they work exactly as before.

Loft: Select the profiles in the tree or in the 3d view, doesn't matter, and then click the toolbar icon or menu item to make the loft.

Sweep: Select the profile(s) in the tree view.  Select the spine edges in the 3d view.

Extrude: Select the profile in the tree, an edge reference in the 3d view.  The edge reference must be a line segment.  The edge will be the DirLink property and DirMode property is set to "Edge".  If no edge is selected, then you get the dialog as before.

Revolve: Select the profile in the tree, edge reference in the 3d view.  The edge reference can be a line segment or an arc.  The edge is used to set the AxisLink property.

This is implemented in a python file in BasicShapes called SkipGui.py.  In case of a runtime script error the dialog is shown.  If in the script it is determined the selection is invalid, then the dialog is shown.

### DIFF
--- a/src/Mod/Part/BasicShapes/SkipGui.py
+++ b/src/Mod/Part/BasicShapes/SkipGui.py
@@ -1,0 +1,250 @@
+#! python
+# -*- coding: utf-8 -*-
+# SkipGui.py
+# 2022, by Mark Ganson <TheMarkster>
+# LGPL 2.1 or later
+
+# this file is called from c++ in Mod/Part/Gui/Command.cpp
+# from various functions when these commands are activated:
+#   Part::Loft
+#   Part::Sweep
+#   Part::Extrude
+#   Part::Revolve
+# Then if the user has preselected certain objects and subobjects
+# the Gui can be skipped and the document objects directly created
+
+# If False is returned, then the Gui dialog should be shown
+# Else if True is returned, then the dialog can be skipped
+
+import FreeCAD as App
+import FreeCADGui as Gui
+
+def getActStr(document):
+    return "App.getDocument(\'" + document.Name + "\').ActiveObject"
+
+def getDocStr(document):
+    return "App.getDocument(\'" + document.Name +"\')"
+
+def getObjStr(document, object):
+    return getDocStr(document) + ".getObject(\'" + object.Name + "\')"
+
+def getVisStr(document, objs):
+    ret = ""
+    for obj in objs:
+        ret += getDocStr(document) + ".getObject(\'" + obj.Name + "\').ViewObject.Visibility = False\n"
+    return ret
+
+def getSetPropListStr(document, propName, objList):
+    ret = getActStr(document) + "." + propName + " = ["
+    for obj in objList:
+        ret += getDocStr(document) + ".getObject(\'" + obj.Name + "\'), "
+    ret += "]\n"
+    return ret
+
+def getAddObjStr(document, typeId, name):
+    ret = getDocStr(document)
+    ret += ".addObject(\'" + typeId + "\', \'" + name + "\')\n"
+    return ret
+
+def getSubListStr(names):
+    ret = "["
+    for name in names:
+        ret += "\'" + name + "\', "
+    ret += "]"
+    return ret
+
+
+def makeExtrude():
+
+    selx = Gui.Selection.getSelectionEx()
+    if len(selx) == 0:
+        App.Console.PrintMessage(f"makeExtrude() len(selx) = {len(selx)} (must be > 0 to skip dialog)\n")
+        return False
+
+    selected = [sel.Object for sel in selx if bool(sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1) or
+                sel.Object.isDerivedFrom("Sketcher::SketchObject")) and
+                sel.HasSubObjects == False]
+    if len(selected) != 1:
+        App.Console.PrintMessage(f"makeExtrude() len(selected) = {len(selected)} (must be 1 to skip dialog)\n")
+        return False #need exactly 1 profile for Extrude
+
+    solid = "True"
+    base = selected[0]
+    if hasattr(base.Shape, "Wire1"):
+        if not base.Shape.Wire1.isClosed():
+            solid = "False"
+    else:
+        if not bool(hasattr(base,"Shape") and len(base.Shape.Vertexes) == 1):
+            return False #no Wire1
+        else:
+            solid = "False"
+
+    spines = [sel for sel in selx if sel.HasSubObjects]
+    if len(spines) != 1:
+        return False #all edges must be from the same object
+
+    spineEdges = spines[0].SubElementNames
+    subNames = [name for name in spineEdges if "Edge" in name]
+    if len(subNames) != 1:
+        return False
+
+    dirLink = spines[0].Object.getSubObject(subNames[0])
+    if dirLink.Curve.TypeId != "Part::GeomLine":
+        return False
+
+    if solid == "False":
+        App.Console.PrintWarning("Solid = False\n")
+
+    doc = selected[0].Document
+    cmd = getAddObjStr(doc, "Part::Extrusion", "Extrude")
+    cmd += getActStr(doc) + ".Base = " + getObjStr(doc, selected[0]) + "\n"
+    cmd += getActStr(doc) + ".Solid = " + solid + "\n"
+    cmd += getActStr(doc) + ".DirMode = \'Edge\'\n"
+    cmd += getActStr(doc) + ".Reversed = False\n"
+    cmd += getActStr(doc) + ".Symmetric = False\n"
+    cmd += getVisStr(doc, selected)
+    cmd += getActStr(doc) + ".DirLink = (" + getObjStr(doc, spines[0].Object) + ", \'" + subNames[0] + "\')\n"
+    Gui.doCommand(cmd)
+    doc.recompute()
+    return True
+
+
+def makeRevolve():
+    selx = Gui.Selection.getSelectionEx()
+    if len(selx) == 0:
+        return False #nothing selected, False = don't skip Gui
+
+    selected = [sel.Object for sel in selx if bool(sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1) or
+                sel.Object.isDerivedFrom("Sketcher::SketchObject")) and
+                sel.HasSubObjects == False]
+    if len(selected) != 1:
+        return False #need exactly 1 profile for Revolve
+
+    solid = "True"
+    base = selected[0]
+    if hasattr(base.Shape, "Wire1"):
+        if not base.Shape.Wire1.isClosed():
+            solid = "False"
+    else:
+        if not bool(hasattr(base,"Shape") and len(base.Shape.Vertexes) == 1):
+            return False #no Wire1
+        else:
+            solid = "False"
+
+    spines = [sel for sel in selx if sel.HasSubObjects]
+    if len(spines) != 1:
+        return False #all edges must be from the same objects
+
+    spineEdges = spines[0].SubElementNames
+    subNames = [name for name in spineEdges if "Edge" in name]
+    if len(subNames) != 1:
+        return False
+
+    if solid == "False":
+        App.Console.PrintWarning("Solid = False\n")
+
+    doc = selected[0].Document
+    doc.openTransaction("Revolve")
+    cmd = getAddObjStr(doc, "Part::Revolution", "Revolve")
+    cmd += getActStr(doc) + ".Source = " + getObjStr(doc, selected[0]) + "\n"
+    cmd += getActStr(doc) + ".Solid = " + solid + "\n"
+    cmd += getActStr(doc) + ".Symmetric = False\n"
+    cmd += getActStr(doc) + ".Angle = 360.0\n"
+    cmd += getVisStr(doc, selected)
+    cmd += getActStr(doc) + ".AxisLink = (" + getObjStr(doc, spines[0].Object) + ", \'" + subNames[0] + "\')\n"
+    Gui.doCommand(cmd)
+    doc.commitTransaction()
+    doc.recompute()
+    return True
+
+
+def makeLoft():
+
+    selx = Gui.Selection.getSelectionEx()
+    if len(selx) == 0:
+        return False #nothing selected, False = don't skip Gui
+
+    selected = [sel.Object for sel in selx if sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                sel.Object.isDerivedFrom("Sketcher::SketchObject") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1)]
+    if len(selected) < 2:
+        return False #need at least 2 objects for Loft
+
+    #filter out sketches with multiple wires
+    #should be removed if/when multiple wires are supported for Part::Loft
+    filtered = [sel for sel in selected if len(sel.Shape.Wires) < 2]
+    if len(filtered) < 2:
+        return False
+
+    if len(filtered) != len(selected):
+        return False #user will not get expected result since some profiles are not supported
+
+    doc = filtered[0].Document
+    cmd = getAddObjStr(doc, "Part::Loft", "Loft")
+    cmd += getSetPropListStr(doc, "Sections", filtered)
+    cmd += getActStr(doc) + ".Solid = True\n"
+    cmd += getActStr(doc) + ".Closed = False\n"
+    cmd += getActStr(doc) + ".Ruled = False\n"
+    cmd += getVisStr(doc, filtered)
+    Gui.doCommand(cmd)
+    doc.recompute()
+    return True
+
+
+def makeSweep():
+
+    selx = Gui.Selection.getSelectionEx()
+    if len(selx) == 0:
+        return False #nothing selected, False = don't skip Gui
+
+    selected = [sel.Object for sel in selx if bool(sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1) or
+                sel.Object.isDerivedFrom("Sketcher::SketchObject")) and
+                sel.HasSubObjects == False]
+    if len(selected) < 1:
+        return False #need at least 1 profile for Sweep
+
+    #filter out sketches with multiple wires
+    #should be removed if/when multiple wires are supported for Part::Sweep
+    filtered = [sel for sel in selected if len(sel.Shape.Wires) < 2]
+    if len(filtered) < 1:
+        return False #must have at least 1 profile to sweep
+
+    if len(filtered) != len(selected):
+        return False #user will not get expected result since some profiles are not supported
+
+    solid = "True"
+    base = filtered[0]
+    if hasattr(base.Shape, "Wire1"):
+        if not base.Shape.Wire1.isClosed():
+            solid = "False"
+
+    spines = [sel for sel in selx if sel.HasSubObjects]
+    if len(spines) != 1:
+        return False #all edges must be from the same object
+
+    spineEdges = spines[0].SubElementNames
+    subNames = [name for name in spineEdges if "Edge" in name]
+    if len(subNames) == 0:
+        return False
+
+    if solid == "False":
+        App.Console.PrintWarning("Solid = False\n")
+
+    doc = filtered[0].Document
+    cmd = getAddObjStr(doc, "Part::Sweep", "Sweep")
+    cmd += getSetPropListStr(doc, "Sections", filtered)
+    cmd += getActStr(doc) + ".Solid = " + solid + "\n"
+    cmd += getActStr(doc) + ".Frenet = False\n"
+    cmd += getActStr(doc) + ".Transition = \'Right corner\'\n"
+    cmd += getVisStr(doc, filtered)
+    cmd += getActStr(doc) + ".Spine = (" + getObjStr(doc, spines[0].Object) + ", " + getSubListStr(subNames) + ")\n"
+    Gui.doCommand(cmd)
+    doc.recompute()
+    return True
--- a/src/Mod/Part/BasicShapes/SkipGui.py
+++ b/src/Mod/Part/BasicShapes/SkipGui.py
@@ -20,37 +20,30 @@ import FreeCAD as App
 import FreeCADGui as Gui
 
 def getActStr(document):
-    return "App.getDocument(\'" + document.Name + "\').ActiveObject"
+    return f"App.getDocument('{document.Name}').ActiveObject"
 
 def getDocStr(document):
-    return "App.getDocument(\'" + document.Name +"\')"
+    return f"App.getDocument('{document.Name}')"
 
 def getObjStr(document, object):
-    return getDocStr(document) + ".getObject(\'" + object.Name + "\')"
+    return f"{getDocStr(document)}.getObject('{object.Name}')"
 
 def getVisStr(document, objs):
     ret = ""
     for obj in objs:
-        ret += getDocStr(document) + ".getObject(\'" + obj.Name + "\').ViewObject.Visibility = False\n"
+        ret += f"{getDocStr(document)}.getObject('{obj.Name}').ViewObject.Visibility = False\n"
     return ret
 
 def getSetPropListStr(document, propName, objList):
-    ret = getActStr(document) + "." + propName + " = ["
+    ret = f"{getActStr(document)}.{propName} = ["
     for obj in objList:
-        ret += getDocStr(document) + ".getObject(\'" + obj.Name + "\'), "
+        ret += f"{getDocStr(document)}.getObject('{obj.Name}'), "
     ret += "]\n"
     return ret
 
 def getAddObjStr(document, typeId, name):
     ret = getDocStr(document)
-    ret += ".addObject(\'" + typeId + "\', \'" + name + "\')\n"
-    return ret
-
-def getSubListStr(names):
-    ret = "["
-    for name in names:
-        ret += "\'" + name + "\', "
-    ret += "]"
+    ret += f".addObject('{typeId}', '{name}')\n"
     return ret
 
 
@@ -58,17 +51,18 @@ def makeExtrude():
 
     selx = Gui.Selection.getSelectionEx()
     if len(selx) == 0:
-        App.Console.PrintMessage(f"makeExtrude() len(selx) = {len(selx)} (must be > 0 to skip dialog)\n")
+        App.Console.PrintLog(f"makeExtrude() nothing selected, not skipping dialog\n")
         return False
 
     selected = [sel.Object for sel in selx if bool(sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::Face") or
                 sel.Object.isDerivedFrom("Part::FeaturePython") or
                 bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1) or
                 sel.Object.isDerivedFrom("Sketcher::SketchObject")) and
                 sel.HasSubObjects == False]
     if len(selected) != 1:
-        App.Console.PrintMessage(f"makeExtrude() len(selected) = {len(selected)} (must be 1 to skip dialog)\n")
-        return False #need exactly 1 profile for Extrude
+        App.Console.PrintLog(f"makeExtrude() need exactly 1 selected profile to skip dialog, not skipping\n")
+        return False
 
     solid = "True"
     base = selected[0]
@@ -77,21 +71,25 @@ def makeExtrude():
             solid = "False"
     else:
         if not bool(hasattr(base,"Shape") and len(base.Shape.Vertexes) == 1):
+            App.Console.PrintLog(f"makeExtrude() no wire in shape and not a vertex (not skip dialog)\n")
             return False #no Wire1
         else:
             solid = "False"
 
     spines = [sel for sel in selx if sel.HasSubObjects]
     if len(spines) != 1:
-        return False #all edges must be from the same object
+        App.Console.PrintLog(f"makeExtrude() all edges not from same object, not skipping dialog\n")
+        return False
 
     spineEdges = spines[0].SubElementNames
     subNames = [name for name in spineEdges if "Edge" in name]
     if len(subNames) != 1:
+        App.Console.PrintLog(f"makeExtrude() only 1 edge supported for DirLink, not skipping dialog\n")
         return False
 
     dirLink = spines[0].Object.getSubObject(subNames[0])
     if dirLink.Curve.TypeId != "Part::GeomLine":
+        App.Console.PrintLog(f"makeExtrude() selected edge must be of TypeId Part::GeomLine, is TypeId = {dirLink.Curve.TypeId}, not skipping dialog\n")
         return False
 
     if solid == "False":
@@ -99,13 +97,14 @@ def makeExtrude():
 
     doc = selected[0].Document
     cmd = getAddObjStr(doc, "Part::Extrusion", "Extrude")
-    cmd += getActStr(doc) + ".Base = " + getObjStr(doc, selected[0]) + "\n"
-    cmd += getActStr(doc) + ".Solid = " + solid + "\n"
-    cmd += getActStr(doc) + ".DirMode = \'Edge\'\n"
-    cmd += getActStr(doc) + ".Reversed = False\n"
-    cmd += getActStr(doc) + ".Symmetric = False\n"
+    actStr = getActStr(doc)
+    cmd += actStr + ".Base = " + getObjStr(doc, selected[0]) + "\n"
+    cmd += actStr + ".Solid = " + solid + "\n"
+    cmd += actStr + ".DirMode = 'Edge'\n"
+    cmd += actStr + ".Reversed = False\n"
+    cmd += actStr + ".Symmetric = False\n"
     cmd += getVisStr(doc, selected)
-    cmd += getActStr(doc) + ".DirLink = (" + getObjStr(doc, spines[0].Object) + ", \'" + subNames[0] + "\')\n"
+    cmd += actStr + ".DirLink = (" + getObjStr(doc, spines[0].Object) + ", '" + subNames[0] + "')\n"
     Gui.doCommand(cmd)
     doc.recompute()
     return True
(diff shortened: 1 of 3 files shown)


mwganson:
https://forum.freecadweb.org/viewtopic.php?f=8&t=65291

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378">pipeline status</a> for feature branch [PR_5378](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378). Pipeline [447601517](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/447601517) was triggered at [e7538a8](https://github.com/FreeCAD/FreeCAD/pull/5378/commits/e7538a8389abdf6cfd9a9ba0a452757f3618ed66). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
May this behavior (skipping) be made optional? 
From code point of view, wouldn't that be better to cancel the current transaction in the 'catch' blocks?

mwganson:
I will add a boolean user parameter called NeverSkipDialog in BaseApp/Preferences/Mod/Part, defaulting to False.  If True, then the dialogs are always shown.

In the second question you refer to the document transaction for undo/redo?  Open the transaction in c++, then call the python code and close it after python returns and also in the catch block in the event of a runtime exception?

0penBrain:
> In the second question you refer to the document transaction for undo/redo? Open the transaction in c++, then call the python code and close it after python returns and also in the catch block in the event of a runtime exception?

My wild guess ATM is that "opening the dialog" will open a transaction, so probably in case of failure/exception it's better to cancel the current one so it won't appear in the undo stack. It has to be checked, I didn't have a careful look at the code. Consider it as a remark right now. :)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378">pipeline status</a> for feature branch [PR_5378](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378). Pipeline [454036788](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454036788) was triggered at [2f24093](https://github.com/FreeCAD/FreeCAD/pull/5378/commits/2f24093159d79ac486438297cd77b5e990730912). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
@mwganson did some rebase goes wrong? The commit list is insane. 🤣

mwganson:
Yes, I struggle with these for some reason.  I sent you a PM.  Maybe you can get me sorted out with these.

mwganson:
Marked as draft.  I want to work on the transaction issue.  My plan is to open the transaction in c++ instead of python, then call the python script, then close as normally done after the return from python.  In the exception handler I will abort the transaction.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378">pipeline status</a> for feature branch [PR_5378](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378). Pipeline [454557978](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454557978) was triggered at [1fbd424](https://github.com/FreeCAD/FreeCAD/pull/5378/commits/1fbd4249cee6abca3308616292b42d427255566c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378">pipeline status</a> for feature branch [PR_5378](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378). Pipeline [455229703](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/455229703) was triggered at [9c63cc7](https://github.com/FreeCAD/FreeCAD/pull/5378/commits/9c63cc720a38188421fd9baa372ff4bcd56a46a2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1329,7 +1329,38 @@ CmdPartExtrude::CmdPartExtrude()
 void CmdPartExtrude::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    Gui::Control().showDialog(new PartGui::TaskExtrusion());
+    /** If the user selects objects prior to executing the Extrude command
+     *  we can use those objects rather than opening the dialog
+     */
+
+    bool neverSkipDialog = App::GetApplication().GetUserParameter().GetGroup("BaseApp/Preferences/Mod/Part")->GetBool("NeverSkipDialog", false);
```
wwmayer, line 1336: You 4x almost the identical code. The only difference is the name of the Python function, the undo/redo text and the dialog class. To avoid the code duplication you can move this to a separate template function (the template parameter will be the name of the task dialog class, e.g. PartGui::TaskExtrusion) and two function arguments for the Python function and the text.

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1329,7 +1329,38 @@ CmdPartExtrude::CmdPartExtrude()
...
+            Py::Boolean result(skipgui.callMemberFunction("makeExtrude"));
+            bool skip = result.as_bool();
+            if (!skip){
+                doc->openTransaction("Make Extrude");
+                Gui::Control().showDialog(new PartGui::TaskExtrusion());
+                doc->commitTransaction();
+            }
+        }
+        catch (Py::Exception&) {
+            doc->abortTransaction();
```
wwmayer, line 1355: The call of abortTransaction() here is quite problematic because you don't know where inside the try-block the exception was raised. 
You check for Py::Exception and most likely this will happen when running the function of the script, i.e. before you actually open the transaction.
The next thing is that you open a transaction before opening the dialog but after running the script. So, if the script was running successfully the new object won't be part of the transaction.

To address all the issues you should move the execution + transaction handling into a separate function or lambda function:
```
auto callFunction = [](const Py::Object& module, const std::string& function, App::Document* doc, const char* text) {
    try {
        doc->openTransaction(text);
        Py::Boolean result(skipgui.callMemberFunction(function));
        bool ok = result.as_bool();
        ok ? doc->commitTransaction() : doc->abortTransaction();
        return ok;
    }
    catch (const Py::Exception&) {
        doc->abortTransaction();
        throw; // re-throw the exception
    }
};
```

mwganson:
I have made the changes for another review, but I might have made some mistake with git.  Let me know if I need to close this and resubmit as another PR in addition to whatever other changes might be needed with the code.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378">pipeline status</a> for feature branch [PR_5378](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378). Pipeline [456431068](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/456431068) was triggered at [c4a5daa](https://github.com/FreeCAD/FreeCAD/pull/5378/commits/c4a5daae27268f366cdaae25cd8504afb7da62aa). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
@mwganson Indeed commit history is a bit weird. You can try `git pull --rebase upstream master` then check with `git log` how it looks. If it's correct you can `git push`.

If it's not correct there, you have to look at `git log` what is the SHA of your commit then you can try following :
`git checkout master # got to master branch
git pull --ff-only upstream master # update
git checkout -b tempBranch # create a temporary branch
git cherry-pick SHA # cherry pick your commit in the temp branch
git push --force origin tempBranch:part_loft # overwrite PR branch with temp branch`

At this step the PR should be OK. Then a it of clean up : 
`git checkout part_loft # switch to old branch
git branch -d tempBranch # delete temporary branch
git fetch origin part_loft # fetch remote branch to index
git reset --hard origin/part_loft # reset working copy`

HTH

mwganson:
I think that got it fixed.  Thanks.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+# the Gui can be skipped and the document objects directly created
+
+# If False is returned, then the Gui dialog should be shown
+# Else if True is returned, then the dialog can be skipped
+
+import FreeCAD as App
+import FreeCADGui as Gui
+
+def getActStr(document):
+    return "App.getDocument(\'" + document.Name + "\').ActiveObject"
```
0penBrain, line 23: I think a formatted string is more readable.

I didn't look at the function usage, but beware of what `activeObject` really is.

I think you don't need to escape single quotes as your string is enclosed in double quotes.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+
+def getDocStr(document):
+    return "App.getDocument(\'" + document.Name +"\')"
```
0penBrain, line 26: Formatted string. No single quote escape needed.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+
+def getObjStr(document, object):
+    return getDocStr(document) + ".getObject(\'" + object.Name + "\')"
```
0penBrain, line 29: Formatted string. No single quote escape needed.

Why use `getDocStr` here and not in `getActStr` ?

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+
+def getVisStr(document, objs):
+    ret = ""
+    for obj in objs:
+        ret += getDocStr(document) + ".getObject(\'" + obj.Name + "\').ViewObject.Visibility = False\n"
```
0penBrain, line 34: Formatted string. No single quote escape needed.

Name of the function is a bit weird as it eventually not only get things, but set Visibility to False.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    return ret
+
+def getSetPropListStr(document, propName, objList):
+    ret = getActStr(document) + "." + propName + " = ["
+    for obj in objList:
+        ret += getDocStr(document) + ".getObject(\'" + obj.Name + "\'), "
+    ret += "]\n"
```
0penBrain, line 41: Maybe simpler to build a Python list, then append its `repr()` to the base string.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    return ret
+
+def getAddObjStr(document, typeId, name):
+    ret = getDocStr(document)
+    ret += ".addObject(\'" + typeId + "\', \'" + name + "\')\n"
```
0penBrain, line 46: Formatted string. No single quote escape needed.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    return ret
+
+def getSubListStr(names):
+    ret = "["
+    for name in names:
+        ret += "\'" + name + "\', "
+    ret += "]"
+    return ret
```
0penBrain, line 54: Use native `repr()` of Python ?

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def makeExtrude():
+
+    selx = Gui.Selection.getSelectionEx()
+    if len(selx) == 0:
+        App.Console.PrintMessage(f"makeExtrude() len(selx) = {len(selx)} (must be > 0 to skip dialog)\n")
```
0penBrain, line 61: Do we really want this kind of message to appear regularly ? Isn't `PrintLog` enough ?

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+        return False
+
+    selected = [sel.Object for sel in selx if bool(sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1) or
```
0penBrain, line 66: I don't understand this line. This is a lone vertex, but that won't be a vertex object ???

Also the filter will skip Draft 2D objects, faces, and much more. This looks very limited to me.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+                sel.Object.isDerivedFrom("Sketcher::SketchObject")) and
+                sel.HasSubObjects == False]
+    if len(selected) != 1:
```
0penBrain, line 69: Doing so (with previous filter) means that if I select 3 faces + 1 sketch, the dialog will be skipped but only the sketch will be extruded and the 3 faces will be disregarded. This looks a weird behavior to me.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    if hasattr(base.Shape, "Wire1"):
+        if not base.Shape.Wire1.isClosed():
+            solid = "False"
+    else:
+        if not bool(hasattr(base,"Shape") and len(base.Shape.Vertexes) == 1):
+            return False #no Wire1
+        else:
+            solid = "False"
+
+    spines = [sel for sel in selx if sel.HasSubObjects]
```
0penBrain, line 84: Don't much understand? If all edges are from same object, `getSelectionEx` will return only one object.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    if len(spines) != 1:
+        return False #all edges must be from the same object
+
+    spineEdges = spines[0].SubElementNames
+    subNames = [name for name in spineEdges if "Edge" in name]
```
0penBrain, line 89: Would me more proper to use `SubObjects` then filter using `TypeId` of `isInstance`

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    if solid == "False":
+        App.Console.PrintWarning("Solid = False\n")
+
+    doc = selected[0].Document
+    cmd = getAddObjStr(doc, "Part::Extrusion", "Extrude")
+    cmd += getActStr(doc) + ".Base = " + getObjStr(doc, selected[0]) + "\n"
+    cmd += getActStr(doc) + ".Solid = " + solid + "\n"
+    cmd += getActStr(doc) + ".DirMode = \'Edge\'\n"
+    cmd += getActStr(doc) + ".Reversed = False\n"
+    cmd += getActStr(doc) + ".Symmetric = False\n"
```
0penBrain, line 106: Maybe storing `getActStr(doc)` in a temporary string may be beneficial to prevent repeated calls.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    cmd += getVisStr(doc, selected)
+    cmd += getActStr(doc) + ".DirLink = (" + getObjStr(doc, spines[0].Object) + ", \'" + subNames[0] + "\')\n"
+    Gui.doCommand(cmd)
+    doc.recompute()
+    return True
+
+
+def makeRevolve():
```
0penBrain, line 114: Roughly same comments as in `makeExtrude`

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    cmd += getActStr(doc) + ".Angle = 360.0\n"
+    cmd += getVisStr(doc, selected)
+    cmd += getActStr(doc) + ".AxisLink = (" + getObjStr(doc, spines[0].Object) + ", \'" + subNames[0] + "\')\n"
+    Gui.doCommand(cmd)
+    doc.commitTransaction()
+    doc.recompute()
+    return True
+
+
+def makeLoft():
```
0penBrain, line 165: Roughly same comments as in `makeExtrude`

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,250 @@
...
+    cmd += getActStr(doc) + ".Solid = True\n"
+    cmd += getActStr(doc) + ".Closed = False\n"
+    cmd += getActStr(doc) + ".Ruled = False\n"
+    cmd += getVisStr(doc, filtered)
+    Gui.doCommand(cmd)
+    doc.recompute()
+    return True
+
+
+def makeSweep():
```
0penBrain, line 199: Roughly same comments as in `makeExtrude`

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1309,6 +1309,57 @@ bool CmdPartBoolean::isActive(void)
...
+    try {
+        doc -> openTransaction(text);
+        Py::Boolean result(module.callMemberFunction(function));
+        bool ok = result.as_bool();
+        ok ? doc->commitTransaction() : doc->abortTransaction();
+        return ok;
+    }
+    catch (const Py::Exception&) {
+        doc->abortTransaction();
+        throw; //re-throw the exception
```
0penBrain, line 1331: According the usage of the function, wouldn't that be better to return `false` rather than throwing an exception ?

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1309,6 +1309,57 @@ bool CmdPartBoolean::isActive(void)
...
+void checkSkipDialog (Gui::TaskView::TaskDialog *dlg, const std::string& function, const char* text) {
+
+    bool neverSkipDialog = App::GetApplication().GetUserParameter().GetGroup("BaseApp/Preferences/Mod/Part")->GetBool("NeverSkipDialog", false);
+    if (!neverSkipDialog) {
+        App::Document* doc = App::GetApplication().getActiveDocument();
+        Base::PyGILStateLocker lock;
+        try {
+            PyObject* module = PyImport_ImportModule("BasicShapes.SkipGui");
+            if (!module) {
+                throw Py::Exception();
```
0penBrain, line 1344: Probably good to add a reason string for exception throwing.

0penBrain (commented):
@mwganson Lots of comments, sorry. :D 

All in all, adding some comments about the intentions (especially in makeXXX functions) would be good.

Admittedly, I'm not sure this function will be used so much. May you have a poll to community in order to know if we should enable or disable it by default ?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378">pipeline status</a> for feature branch [PR_5378](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378). Pipeline [457077909](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/457077909) was triggered at [f62dc7b](https://github.com/FreeCAD/FreeCAD/pull/5378/commits/f62dc7b6fe061ead9a27d6cd56570f763f8b0f2c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: By formatted string you mean something like:

`return f"App.getdocument('{document.Name}').ActiveObject"`

Definitely, that looks easier to read.  And I think now python 3.6 is minimum supported version, so it should be okay to use.

ActiveObject is the most recently created/edited object, I believe.  I preferred this to adding a temporary variable.

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: Where I know the object name I prefer to use it over ActiveObject.  When adding a new object with addObject() the name is returned, but I didn't want to save to a temporary variable and pollute the namespace unnecessarily.

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: All of these begin with get because they are getting a string value, here the string it gets is the string used to set visibility to false.  I think of it has getVisibilityStringForSettingVisibilityToFalse(), but this was less typing.

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: Good idea.  I will use that instead of getSubListStr() function.  Here I think this is better, but with using formatted strings instead for easier readability.

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: Done.

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: No, that was for debugging and I neglected to remove it.  But I like the idea of logging.

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: Did you try this?  When I select 2 sketches and 1 face I get the dialog and all 3 are extruded.

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: Yes, getSelectionEx() only returns 1 object if all edges are from the same object, but we don't know that to be the case yet.  The user might have selected multiple edges from multiple objects, in which case we'll not skip the dialog.

HasSubObjects ensures these objects were selected in the 3d view and not in the tree.  For extrude this will be used for the DirLink property which is of type App::PropertyLinkSub, which only accepts a single object.  It would only be sensible for 1 object with this property anyway since it defines the direction and distance to extrude, but for Sweep it could be sensible to use App::PropertyLinkSubList instead, to support edges from multiple objects to define the path.  But I stray off topic.

So we need to check here that len(spines) = 1 so we can use spines[0] in the next bit of code.

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: Perhaps if we were only interested in the subobjects, but we need the subname not the subobject so we can pass the name to doCommand().

src/Mod/Part/BasicShapes/SkipGui.py
mwganson: Good idea.

src/Mod/Part/Gui/Command.cpp
mwganson: I think this way you get a better error message of what went wrong in the script rather than simply returning false and showing the dialog.

src/Mod/Part/Gui/Command.cpp
mwganson: I think this only gets thrown if the SkipGui.py file can not be found or possibly if the lock is being held by another thread.  The reason gets shown in the catch block does it not?

mwganson:
Thanks for all the feedback.  I will incorporate some of these suggestions and submit the changes.

mwganson:
To try to tidy things up a bit I will mark all conversations as resolved, but feel free to bring up again anything not satisfactory.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378">pipeline status</a> for feature branch [PR_5378](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5378). Pipeline [457435781](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/457435781) was triggered at [30dc0a1](https://github.com/FreeCAD/FreeCAD/pull/5378/commits/30dc0a1f0fe30bb0cfcc6843d45bf1b79318bd64). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

mwganson:
I added log messages to each return of False.  That should help understand what the code is doing and also aid in debugging when unexpectedly the dialog is shown.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+    if len(selx) == 0:
+        App.Console.PrintLog(f"makeExtrude() nothing selected, not skipping dialog\n")
+        return False
+
+    selected = [sel.Object for sel in selx if bool(sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::Face") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1) or
+                sel.Object.isDerivedFrom("Sketcher::SketchObject")) and
+                sel.HasSubObjects == False]
```
0penBrain, line 62: Should that be took into account ? It will force user to take care of selecting the whole object rather than just one of it's subelements ? I have no strong opinion on this.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+    if len(selx) == 0:
+        App.Console.PrintLog(f"makeRevolve() nothing selected, not skipping dialog\n")
+        return False
+
+    selected = [sel.Object for sel in selx if bool(sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::Face") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1) or
+                sel.Object.isDerivedFrom("Sketcher::SketchObject")) and
+                sel.HasSubObjects == False]
```
0penBrain, line 124: Same as `makeExtrude`

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+            solid = "False"
+
+    spines = [sel for sel in selx if sel.HasSubObjects]
+    if len(spines) != 1:
+        App.Console.PrintLog(f"makeRevolve() all edges must be from the same object, not skipping dialog\n")
+        return False
+
+    spineEdges = spines[0].SubElementNames
+    subNames = [name for name in spineEdges if "Edge" in name]
+    if len(subNames) != 1:
```
0penBrain, line 148: Does Part/Revolve works if AxisLink isn't a Line but a Curve ?

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+    selected = [sel.Object for sel in selx if sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::Face") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                sel.Object.isDerivedFrom("Sketcher::SketchObject") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1)]
+    if len(selected) < 2:
+        App.Console.PrintLog(f"makeLoft() need at least 2 objects selected, not skipping dialog\n")
+        return False
+
+    #filter out sketches with multiple wires
```
0penBrain, line 186: Does Part/Extrude and Part/Revolve support sketch with multiple wire ?

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+    if len(selx) == 0:
+        App.Console.PrintLog(f"makeSweep() nothing selected, not skipping dialog\n")
+        return False
+
+    selected = [sel.Object for sel in selx if bool(sel.Object.isDerivedFrom("Part::2DObject") or
+                sel.Object.isDerivedFrom("Part::Face") or
+                sel.Object.isDerivedFrom("Part::FeaturePython") or
+                bool(hasattr(sel.Object,"Shape") and len(sel.Object.Shape.Vertexes) == 1) or
+                sel.Object.isDerivedFrom("Sketcher::SketchObject")) and
+                sel.HasSubObjects == False]
```
0penBrain, line 221: Same as `makeExtrude`

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1329,7 +1380,7 @@ CmdPartExtrude::CmdPartExtrude()
 void CmdPartExtrude::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    Gui::Control().showDialog(new PartGui::TaskExtrusion());
+    PartGui::checkSkipDialog(new PartGui::TaskExtrusion(), "makeExtrude", "Extrude");
```
0penBrain, line 1383: Text used for Undo transactions are generally translated

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1309,6 +1309,57 @@ bool CmdPartBoolean::isActive(void)
...
+            }
+        }
+        catch (Py::Exception&) {
+            Base::PyException e;
+            e.ReportException();
+            Gui::Control().showDialog(dlg); //show the dialog on runtime python error
+        }
+    } else {
+        Gui::Control().showDialog(dlg); //user parameter set to never skip the dialog
+    }
```
0penBrain, line 1358: Maybe you could return if `callFunction` succeeds, then `showDialog` only once as a default behavior at the end of he function.

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1409,7 +1460,7 @@ CmdPartRevolve::CmdPartRevolve()
 void CmdPartRevolve::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    Gui::Control().showDialog(new PartGui::TaskRevolution());
+    PartGui::checkSkipDialog(new PartGui::TaskRevolution(), "makeRevolve", "Revolve");
```
0penBrain, line 1463: Text used for Undo transactions are generally translated

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1590,7 +1641,7 @@ CmdPartLoft::CmdPartLoft()
 void CmdPartLoft::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    Gui::Control().showDialog(new PartGui::TaskLoft());
+    PartGui::checkSkipDialog(new PartGui::TaskLoft(), "makeLoft", "Loft");
 }
```
0penBrain, line 1645: Text used for Undo transactions are generally translated

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1619,7 +1670,7 @@ CmdPartSweep::CmdPartSweep()
 void CmdPartSweep::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    Gui::Control().showDialog(new PartGui::TaskSweep());
+    PartGui::checkSkipDialog(new PartGui::TaskSweep(), "makeSweep", "Sweep");
```
0penBrain, line 1673: Text used for Undo transactions are generally translated

0penBrain (commented):
@mwganson Sorry I owe you this review for a pretty long time. It's done now. :)

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+# the Gui can be skipped and the document objects directly created
+
+# If False is returned, then the Gui dialog should be shown
+# Else if True is returned, then the dialog can be skipped
+
+import FreeCAD as App
+import FreeCADGui as Gui
+
+def getActStr(document):
+    return f"App.getDocument('{document.Name}').ActiveObject"
```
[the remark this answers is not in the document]
mwganson: I don't think the object is available when this one is called.  This is because I prefer not to create a variable in the console if it can be avoided.  No, I didn't check if it was available in FreeCAD.  I presumed it wasn't.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+
+def getDocStr(document):
+    return f"App.getDocument('{document.Name}')"
+
+def getObjStr(document, object):
+    return f"{getDocStr(document)}.getObject('{object.Name}')"
+
+def getVisStr(document, objs):
```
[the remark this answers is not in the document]
mwganson: Done.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+    ret = ""
+    for obj in objs:
+        ret += f"{getDocStr(document)}.getObject('{obj.Name}').ViewObject.Visibility = False\n"
+    return ret
+
+def getSetPropListStr(document, propName, objList):
+    ret = f"{getActStr(document)}.{propName} = ["
+    for obj in objList:
+        ret += f"{getDocStr(document)}.getObject('{obj.Name}'), "
```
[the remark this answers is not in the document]
mwganson: Done.

src/Mod/Part/BasicShapes/SkipGui.py
```diff
@@ -0,0 +1,267 @@
...
+def getObjStr(document, object):
+    return f"{getDocStr(document)}.getObject('{object.Name}')"
+
+def getVisStr(document, objs):
+    ret = ""
+    for obj in objs:
+        ret += f"{getDocStr(document)}.getObject('{obj.Name}').ViewObject.Visibility = False\n"
+    return ret
+
+def getSetPropListStr(document, propName, objList):
```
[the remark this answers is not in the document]
mwganson: Wouldn't that infer that sometimes the list is not sent?  This is not the case because (unless there is a bug) it is always sent.

src/Mod/Part/Gui/Command.cpp
mwganson: Not here:

https://github.com/FreeCAD/FreeCAD/blob/796efc4c6b76b93ea36905eb013b79bb197dbbef/src/Mod/Sketcher/Gui/TaskSketcherValidation.cpp#L161

https://github.com/FreeCAD/FreeCAD/blob/853454a13920788120234d5e0f64b172a9e27315/src/Mod/Mesh/Gui/Segmentation.cpp#L108

https://github.com/FreeCAD/FreeCAD/blob/853454a13920788120234d5e0f64b172a9e27315/src/Mod/Points/Gui/Command.cpp#L340

https://github.com/FreeCAD/FreeCAD/blob/5d49bf78de785a536f941f1a6d06d432582a95d3/src/Mod/Part/Gui/Mirroring.cpp#L135

https://github.com/FreeCAD/FreeCAD/blob/5d49bf78de785a536f941f1a6d06d432582a95d3/src/Mod/Part/Gui/DlgRevolution.cpp#L356

src/Mod/Part/Gui/Command.cpp
mwganson: Done.

src/Mod/Part/BasicShapes/SkipGui.py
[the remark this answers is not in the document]
mwganson: I misunderstood at first.  Done.

src/Mod/Part/Gui/Command.cpp
chennes: You'll find untranslated `openTransaction` calls throughout the codebase because it didn't used to be supported (at least, that's my understanding of a discussion with @wwmayer at https://forum.freecadweb.org/viewtopic.php?f=35&t=65436). I believe it's now allowed to translate that text.

mwganson:
I thought I could delete it and recreate it without closing the PR.  I guess I have to open another one for this.

src/Mod/Part/Gui/Command.cpp
mwganson: Okay.  I will add it then.  QT_TR_NOOP() is used in other places in that file, so I will use that, too.

chennes: I don't know if the NOOP variant is correct there: does FreeCAD internally call translate() on the Undo menu item? Honestly the undo system is still a bit of a mystery to me, we may have to just test it.

mwganson: I guess we should move the conversation to the newest PR.

https://github.com/FreeCAD/FreeCAD/pull/6380

PR #5379, opened 2022-01-13
PartDesign: Taper angle in Pad

davidosterberg:
Make Pad support tapered pads with the help of BRepOffsetAPI_DraftAngle

This works for pads that are created from line segments and arcs.
It does not work for profiles that contain bspline segments. It also cannot pad profiles with arcs in other
directions than the normal direction.

- The advantage is that the implementation is very simple.
- TNP friendly since the face numbering is not dependent on if taper angle is 0 or not 0.
- Produces simple faces (Planar faces and cones).
- Inner and outer wires are seamlessly supported and behaves consistently

This PR should be seen as a complement to #5357 and #5367.
In particular this PR does not contain any UI code. 

Forum discussion:
https://forum.freecadweb.org/viewtopic.php?f=10&t=65171

image

Behavior is a bit different than the implementation in #5357. In this PR there is only one Taper angle and it will apply to both directions
image

### DIFF
--- a/src/Mod/PartDesign/App/FeatureExtrude.h
+++ b/src/Mod/PartDesign/App/FeatureExtrude.h
@@ -46,6 +46,7 @@ public:
     App::PropertyBool        AlongSketchNormal;
     App::PropertyLength      Offset;
     App::PropertyLinkSub     ReferenceAxis;
+    App::PropertyAngle       TaperAngle;
 
     static App::PropertyQuantityConstraint::Constraints signedLengthConstraint;
 
--- a/src/Mod/PartDesign/App/FeaturePad.cpp
+++ b/src/Mod/PartDesign/App/FeaturePad.cpp
@@ -23,8 +23,10 @@
 
 #include "PreCompiled.h"
 #ifndef _PreComp_
+# include <BRepOffsetAPI_DraftAngle.hxx>
 # include <BRep_Builder.hxx>
 # include <BRep_Tool.hxx>
+# include <BRepTools.hxx>
 # include <BRepAlgoAPI_Common.hxx>
 # include <BRepAlgoAPI_Fuse.hxx>
 # include <BRepAdaptor_Surface.hxx>
@@ -44,6 +46,8 @@
 # include <TopoDS_Solid.hxx>
 # include <TopoDS_Wire.hxx>
 # include <TopExp_Explorer.hxx>
+# include <GeomLProp_SLProps.hxx>
+# include <BRepGProp_Face.hxx>
 #endif
 
 #include <App/Document.h>
@@ -75,6 +79,7 @@ Pad::Pad()
     ADD_PROPERTY_TYPE(UpToFace, (0), "Pad", App::Prop_None, "Face where pad will end");
     ADD_PROPERTY_TYPE(Offset, (0.0), "Pad", App::Prop_None, "Offset from face in which pad will end");
     Offset.setConstraints(&signedLengthConstraint);
+    ADD_PROPERTY_TYPE(TaperAngle, (0.0), "Pad", App::Prop_None, "Pad taper angle");
 
     // Remove the constraints and keep the type to allow to accept negative values
     // https://forum.freecadweb.org/viewtopic.php?f=3&t=52075&p=448410#p447636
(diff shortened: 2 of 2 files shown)


AjinkyaDahale:
Thanks for the PR!

I wonder why we would like to do it this way rather than making a pad and then make a draft. Draft itself already uses `BRepOffsetAPI_DraftAngle` internally and should have the same capabilities and limitations, apart from being able to only draft some of the faces.

If it is still convenient to have a taper while making the pad, I would say it should use the pre-existing draft and auto-add all the faces in the command.

davidosterberg:
> I wonder why we would like to do it this way rather than making a pad and then make a draft. Draft itself already uses `BRepOffsetAPI_DraftAngle` internally and should have the same capabilities and limitations, apart from being able to only draft some of the faces.

 FeatureDraft is operating with FreeCAD face numbering. Topological naming problem makes FeatureDraft unusable for anything that will be parametric. 


> If it is still convenient to have a taper while making the pad, I would say it should use the pre-existing draft and auto-add all the faces in the command.

You still need the logic to identify the neutral plane, and identify which are the new faces. This is not part of FeatureDraft. This logic is 95% of this PR.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5379">pipeline status</a> for feature branch [PR_5379](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5379). Pipeline [447916854](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/447916854) was triggered at [f430b92](https://github.com/FreeCAD/FreeCAD/pull/5379/commits/f430b921e109124e3a52b2e9115e03a5032ce894). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks for the PR. However, we need to act in a planned matter. I will point this out again in the forum.
Here in short:
- I agreed with wmayer that Part Extrude and PD Pad/Pocket should share the same code.
- What you propose seems to work and should therefore first go into the Loft feature because there we don't have angles but have the issue of the shells being always splines.

donovaly:
Hi @davidosterberg , as I wrote here:
https://forum.freecadweb.org/viewtopic.php?p=562438#p562438

The plan is the following:
1. finish #5367

This way we have code that can handle inner wires.

2. using the solution of this PR for Part Extrude. If an Extrude cannot be made with your drafting solution, we take the lofting existing method as fallback

Would this be fine and reasonable with you?

davidosterberg:
> The plan is the following:
> 
> 1. finish #5367
> 
> This way we have code that can handle inner wires.
> 
> 2. using the solution of this PR for Part Extrude. If an Extrude cannot be made with your drafting solution, we take the lofting existing method as fallback
> 
> Would this be fine and reasonable with you?

Sounds good

donovaly:
The mentioned PRs are in.

davidosterberg:
Tapered pad has been merged in master with a different implementation. Unlike the implementation in this PR, that implementation supports having two separate taper angles Forward and Reverse.  Implementing this with the method here would involve a lot of additional code and it would not be elegant. Therefore I have decided to close this PR and work on other things.

donovaly:
> Tapered pad has been merged in master with a different implementation.

What do you mean with different? As I always wrote, the Part Extrude feature offered tapered extrude in two directions so we cannot drop this since this will break existing documents. All I did was to use the existing Part Extrude code also for PartDesign. I did not reinvent or rewrote anything.

davidosterberg:
It was not meant as a complaint. Only as info.  What I mean is that the method in this PR can only deal with one taper angle, not two. This is because it works by taking the prism, then making it tapered. You are right that it was known for long how the other implementation was working. But only now did I actually have time to have a look and try to integrate it into ExtrudeHelper, I conclude that it is not worth doing.

donovaly:
OK. I understand. Thank you for clarification.

donovaly:
Wait, I think we are too fast. We can of course rewrite the code so that tapering is done the following way:
- a prism is created
- if there is a second direction, a second prism is created
- the prism(s) are tapered using Draft
- if there is a second direction, the two drafted prism is merged to one solid

What do you think?

Note that as it is, sketches containing splines cannot be taper-extruded anyway. so with the proposed method, I don't see a regression.

adrianinsaval:
appears to work but only with negative angles
[splineTaper.zip](https://github.com/FreeCAD/FreeCAD/files/8063033/splineTaper.zip)

donovaly:
Since David said he will not continue on this, closing this PR. It can be reopened if necessary or a new PR can be made.

PR #5382, opened 2022-01-14
Part: Fix TangentPlane mode of attachment

marioalexis84:
For some surfaces (BSpline surface for example) the partial derivatives used to calculate the normal could be null, so the normal can't be computed using the cross product between the first partial derivatives on the parameters of the surface at this point. Another similar situation is ∂S/∂u parallel to ∂S/∂v.
To deal with this case an alternative method is added to compute the normal using second order derivatives.
An example is trying to attach a plane to the sphere poles. At this point  tangentU is not defined and ∂S/∂u is null, then the current code fails to do the attachment.
Note: In the sphere case, `LProp_SLProps::Normal()` still compute the normal at the poles (tangentU is null), but this is completly accidental due to diferences in the internal precision of the methods.

See https://git.dev.opencascade.org/gitweb/?p=occt.git;a=blob;f=src/CSLib/CSLib.cxx;h=f526dd54984396e6666c6e912ab44359ec96a9dc;hb=HEAD#l30
and 
https://git.dev.opencascade.org/gitweb/?p=occt.git;a=blob;f=src/LProp/LProp_SLProps.gxx;h=4cb74531f71113a44dffa371ec9d17f1a740da85;hb=HEAD

  

---

### DIFF
--- a/src/Mod/Part/App/Attacher.cpp
+++ b/src/Mod/Part/App/Attacher.cpp
@@ -62,6 +62,7 @@
 
 #include "Attacher.h"
 #include "AttachExtension.h"
+#include "Tools.h"
 #include <Base/Console.h>
 #include <App/OriginFeature.h>
 #include <App/Application.h>
@@ -1177,7 +1178,6 @@ Base::Placement AttachEngine3D::calculateAttachedPlacement(const Base::Placement
         if (vertex.IsNull())
             throw Base::ValueError("Null vertex in AttachEngine3D::calculateAttachedPlacement()!");
 
-        BRepAdaptor_Surface surf (face);
         Handle (Geom_Surface) hSurf = BRep_Tool::Surface(face);
         gp_Pnt p = BRep_Tool::Pnt(vertex);
 
@@ -1185,19 +1185,33 @@ Base::Placement AttachEngine3D::calculateAttachedPlacement(const Base::Placement
         double u, v;
         if (projector.NbPoints()==0)
             throw Base::ValueError("AttachEngine3D::calculateAttachedPlacement: projecting point onto surface failed.");
+
         projector.LowerDistanceParameters(u, v);
-
-        BRepLProp_SLProps prop(surf,u,v,1, Precision::Confusion());
-        SketchNormal = prop.Normal();
-
+        BRepAdaptor_Surface surf(face);
+        BRepLProp_SLProps prop(surf, u, v, 1, Precision::Confusion());
         gp_Dir dirX;
-        prop.TangentU(dirX); //if normal is defined, this should be defined too
+        Standard_Boolean done;
+
+        Tools::getNormal(face, u, v, Precision::Confusion(), SketchNormal, done);
+
+        if (!done)
+            throw Base::ValueError("AttachEngine3D::calculateAttachedPlacement: finding normal to surface at projected point failed.");
+
+        // if getNormal succeeds, at least one of the tangent is defined
+        if (prop.IsTangentUDefined()) {
+            prop.TangentU(dirX);
+            if (face.Orientation() == TopAbs_REVERSED)
+                dirX.Reverse();
+        }
+        // here the orientation of dirX is managed by SketchNormal orientation
+        else {
+            gp_Dir dirY;
+            prop.TangentV(dirY);
+            dirX = dirY.Crossed(SketchNormal);
+        }
+
         SketchXAxis = gp_Vec(dirX).Reversed();//yields upside-down sketches less often.
 
-        if (face.Orientation() == TopAbs_REVERSED) {
-            SketchNormal.Reverse();
-            SketchXAxis.Reverse();
-        }
         if (bThruVertex) {
             SketchBasePoint = p;
         } else {
--- a/src/Mod/Part/App/TopoShapeFacePyImp.cpp
+++ b/src/Mod/Part/App/TopoShapeFacePyImp.cpp
@@ -521,23 +521,20 @@ PyObject* TopoShapeFacePy::normalAt(PyObject *args)
 {
     double u,v;
     if (!PyArg_ParseTuple(args, "dd",&u,&v))
-        return 0;
+        return nullptr;
 
     const TopoDS_Face& f = TopoDS::Face(getTopoShapePtr()->getShape());
-    BRepAdaptor_Surface adapt(f);
+    Standard_Boolean done;
+    gp_Dir dir;
 
-    BRepLProp_SLProps prop(adapt,u,v,2,Precision::Confusion());
-    if (prop.IsNormalDefined()) {
-        gp_Pnt pnt; gp_Vec vec;
-        // handles the orientation state of the shape
-        BRepGProp_Face(f).Normal(u,v,pnt,vec);
-        vec.Normalize();
-        return new Base::VectorPy(new Base::Vector3d(vec.X(),vec.Y(),vec.Z()));
-    }
-    else {
+    Tools::getNormal(f, u, v, Precision::Confusion(), dir, done);
+
+    if (!done) {
         PyErr_SetString(PartExceptionOCCError, "normal not defined");
-        return 0;
+        return nullptr;
     }
+
+    return new Base::VectorPy(new Base::Vector3d(dir.X(),dir.Y(),dir.Z()));
 }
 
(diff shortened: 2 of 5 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5382">pipeline status</a> for feature branch [PR_5382](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5382). Pipeline [448702643](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/448702643) was triggered at [47769a5](https://github.com/FreeCAD/FreeCAD/pull/5382/commits/47769a54af93fdfb8891c27da9cc56f59a6f7f8d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/Attacher.cpp
```diff
@@ -1185,13 +1186,36 @@ Base::Placement AttachEngine3D::calculateAttachedPlacement(const Base::Placement
...
+            CSLib_NormalStatus stat;
+            CSLib::Normal(prop.D1U(), prop.D1V(), prop.D2U(), prop.D2V(), prop.DUV(),
+                Precision::Confusion(), done, stat, SketchNormal);
+            if (!done) {
+                throw Base::ValueError("AttachEngine3D::calculateAttachedPlacement: finding normal to surface at projected point failed.");
+            }
+            if (prop.IsTangentUDefined()) {
+                prop.TangentU(dirX);
+            }
+            else {
```
wwmayer, line 1212: Shouldn't you check here if IsTangentVDefined()?

wwmayer:
Do you have a case where the standard way to compute the normal fails?
I wonder if it would be better to move this into code into a separate helper function/class because using the alternative normal computation would be helpful in other use cases than attachments, too.

marioalexis84:
Here is an example:
Solid is build from a shell formed by (Surface001, Surface002, Face).
Attach Plane in TangentPlane mode in the following cases:
- Reference1 Surface001.Face1 , Reference2 Vertex1: Fails due to undefined Normal and undefined TangentV (∂S/∂V is null).

- Reference1 Solid.Face1, Reference2 Vertex1: Undefined TangentV. Normal defined (wrong value), but it is due to rounding error in the values of ∂S/∂V (on the order of 1e-16).
- Reference1 Sphere.Face1, Reference2 Vertex1: Undefined TangentU. Normal defined (correct value), but it is due accidental correction due to rounding error in the values of ∂S/∂U (in the order of 1e-16). 
[test_attachment.FCStd.gz](https://github.com/FreeCAD/FreeCAD/files/7877707/test_attachment.FCStd.gz)

src/Mod/Part/App/Attacher.cpp
marioalexis84: The method needs at least one of the tangents (U or V) to be non-zero as a necessary condition to compute a non-zero normal. So if the method successfully finds the normal, if IsTangentUDefined() is false, IsTangentVDefined() is necessarily true.

marioalexis84:
> I wonder if it would be better to move this into code into a separate helper function/class because using the alternative normal computation would be helpful in other use cases than attachments, too.

`GeomSurface::normal` should use it.

wwmayer:
> GeomSurface::normal should use it. I will prepare the request.

Thanks! That's exactly what I had in mind.

marioalexis84:
To avoid code duplication, the `getNormal` function has been added to the `Part::Tools` class (tell me if there is a better place).
In `Tools::getNormal`, the flipping of the normal in the particular case mentioned in the code comment is due to considering the correct sign of the increment in the Taylor series used by `CSLib::Normal`.

wwmayer:
> To avoid code duplication, the getNormal function has been added to the Part::Tools class (tell me if there is a better place).

Tools.h is totally fine.

src/Mod/Part/App/Attacher.cpp
```diff
@@ -1177,21 +1179,34 @@ Base::Placement AttachEngine3D::calculateAttachedPlacement(const Base::Placement
...
         double u, v;
         if (projector.NbPoints()==0)
             throw Base::ValueError("AttachEngine3D::calculateAttachedPlacement: projecting point onto surface failed.");
+
         projector.LowerDistanceParameters(u, v);
+        GeomLProp_SLProps prop(hSurf, u, v, 1, Precision::Confusion());
+        gp_Dir dirX;
+        Standard_Boolean done;
 
-        BRepLProp_SLProps prop(surf,u,v,1, Precision::Confusion());
```
wwmayer, line 1190: Be careful here! There is a difference between shapes and geometries. The former can have a placement (or in OCC jargon TopLoc_Location) or orientation and both affect the normal vector.

Example:
```
f=Part.makePlane(1,1)
f.normalAt(0,0) # => (0,0,1)
f.Surface.normal(0,0) # => (0,0,1)
f.reverse()
f.normalAt(0,0) # => (0,0,-1)
f.Surface.normal(0,0) # => (0,0,1)

f.Placement.Rotation.Angle=1
f.Placement.Rotation.Axis=(1,1,1)
f.normalAt(0,0) # => (-0.639, 0.332, -0.693)
f.Surface.normal(0,0) # => (0.639, -0.332, 0.693)
```

The last line needs some more explanation. When running `f.Surface` then internally BRepAdaptor_Surface is used that applies the transformation of the placement to the returned surface. That's why the normal vector is not (0,0,1).
However, when accessing the original surface with `BRep_Tool::Surface` then you will still get the normal (0,0,1).

src/Mod/Part/App/TopoShapeFacePyImp.cpp
```diff
@@ -521,23 +521,24 @@ PyObject* TopoShapeFacePy::normalAt(PyObject *args)
...
-    BRepLProp_SLProps prop(adapt,u,v,2,Precision::Confusion());
-    if (prop.IsNormalDefined()) {
-        gp_Pnt pnt; gp_Vec vec;
-        // handles the orientation state of the shape
-        BRepGProp_Face(f).Normal(u,v,pnt,vec);
-        vec.Normalize();
-        return new Base::VectorPy(new Base::Vector3d(vec.X(),vec.Y(),vec.Z()));
-    }
-    else {
+    Tools::getNormal(surf, u, v, Precision::Confusion(), dir, done);
```
wwmayer, line 531: The same issue as in AttachEngine3D

wwmayer:
> Ok! Any sugestion to deal with that? I could overload the function to works with Handle(Geom_Surface) and BRep_AdaptorSurface

Overloading getNormal() is a good idea. This could be done with  BRep_AdaptorSurface or TopoDS_Face. Then in this version I would also put the handling of the orientation to revert the normal direction so that in client code this doesn't need to be done.

Additionally, some unit tests would be good to be sure the changes don't cause a regression.

src/Mod/Part/App/Attacher.cpp
marioalexis84: Thanks for the aclaration. If I understand correctly, `BRep_Tool::Surface(const TopoDS_Face&, TopLoc_Location& L)` returns the surface without the tranformation applied but the overload `BRep_Tool::Surface(const TopoDS_Face&)` returns a copy of the surface with the transformation applied. The difference I see between working with the `BRepAdaptor_Surface` and the `Handle(Geom_Surface)` returned by `BRep_Tool::Surface(const TopoDS_Face&)` is the restriction on the parameters performed by `BRepAdaptor_Surface`. Is there any error in these statements? What is the impact of this restriction?

donovaly:
I could add the missing header but I don't know it better to OpenCascadeAll.h or to PreCompiled.h. I guess to OpenCascadeAll.h but in PreCompiled.h I see 3 OCC-specific headers.

marioalexis84:
In PreCompiled.h similarly to [this](https://github.com/FreeCAD/FreeCAD/commit/302568d177ddde171c28eaca89e8b4a49466974f#diff-0b567b5e982fffc582d9052719816c29d6ec0445b838580baaf19fd307129917)?

donovaly:
I added it now to OpenCascadeAll.h: commit 39ed257eb

PR #5385, opened 2022-01-16
PartDesign: prevent crash in thickness

mwganson:
addAllEdgesAction is null for Thickness and Draft, so we need to test against this before accessing.

https://forum.freecadweb.org/viewtopic.php?f=19&t=65335


wwmayer:
Ha, ha, ha. I was faster :)

See 2d1c083d9a

mwganson:
You even beat me to closing it.

PR #5386, opened 2022-01-16
PD: Add tests for InvoluteGear

jbaehr:
There are some basic tests to cover the gear profile and its parameters and some integration tests for the interaction with PD Pad and Pocket.

I followed the same pattern as the other PartDesign tests, e.g. not using PEP8 style method names. Eventually some helper methods could be extracted and moved to the common base framework in Mod/Test, but that's another step.

--------

---

### DIFF
--- a/src/Mod/PartDesign/PartDesignTests/TestInvoluteGear.py
+++ b/src/Mod/PartDesign/PartDesignTests/TestInvoluteGear.py
@@ -1,0 +1,136 @@
+#***************************************************************************
+#*   Copyright (c) 2021 Jonas Bähr <jonas.baehr@web.de>                    *
+#*                                                                         *
+#*   This program is free software; you can redistribute it and/or modify  *
+#*   it under the terms of the GNU Lesser General Public License (LGPL)    *
+#*   as published by the Free Software Foundation; either version 2 of     *
+#*   the License, or (at your option) any later version.                   *
+#*   for detail see the LICENCE text file.                                 *
+#*                                                                         *
+#*   This program is distributed in the hope that it will be useful,       *
+#*   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+#*   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+#*   GNU Library General Public License for more details.                  *
+#*                                                                         *
+#*   You should have received a copy of the GNU Library General Public     *
+#*   License along with this program; if not, write to the Free Software   *
+#*   Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  *
+#*   USA                                                                   *
+#*                                                                         *
+#***************************************************************************
+
+import unittest
+
+import FreeCAD
+from Part import makeCircle, Precision
+import InvoluteGearFeature
+
+class TestInvoluteGear(unittest.TestCase):
+    def setUp(self):
+        self.Doc = FreeCAD.newDocument("PartDesignTestInvoluteGear")
+
+    def tearDown(self):
+        FreeCAD.closeDocument(self.Doc.Name)
+
+    def testDefaultGearProfile(self):
+        InvoluteGearFeature.makeInvoluteGear('TestGear')
+        gear = self.Doc.getObject('TestGear')
+        self.assertSuccessfulRecompute(gear)
+        self.assertClosedWire(gear.Shape)
+
+    def testDefaultInternalGearProfile(self):
+        gear = InvoluteGearFeature.makeInvoluteGear('InvoluteGear')
+        gear.ExternalGear = False
+        self.assertSuccessfulRecompute(gear)
+        self.assertClosedWire(gear.Shape)
+
+    def testLowPrecisionGearProfile(self):
+        gear = InvoluteGearFeature.makeInvoluteGear('InvoluteGear')
+        gear.HighPrecision = False
+        self.assertSuccessfulRecompute(gear)
+        self.assertClosedWire(gear.Shape)
+
+    def testLowPrecisionInternalGearProfile(self):
+        gear = InvoluteGearFeature.makeInvoluteGear('InvoluteGear')
+        gear.ExternalGear = False
+        gear.HighPrecision = False
+        self.assertSuccessfulRecompute(gear)
+        self.assertClosedWire(gear.Shape)
+
+    def testCustomizedGearProfile(self):
+        gear = InvoluteGearFeature.makeInvoluteGear('InvoluteGear')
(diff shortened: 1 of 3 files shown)


wwmayer:
You also have to modify the corresponding CMake file as otherwise the new file won't be copied/installed.

jbaehr:
> You also have to modify the corresponding CMake file as otherwise the new file won't be copied/installed.

Thanks. Didn't notice that as I worked with symlinks and wasn't recompiling when only working on python stuff.

src/Mod/PartDesign/PartDesignTests/TestInvoluteGear.py
```diff
@@ -0,0 +1,118 @@
+#   (c) Jonas Bähr (jonas.baehr@web.de) 2021                  LGPL        *
```
luzpaz, line 1: Line 1 needs a line of `#**************************` identical to Line 20
Line 2: 
`#    Copyright (c) 2022 Jonas Bähr <jonas.baehr@web.de>                          *`

No need to specify LGPL on line 2 because it's mentioned below

luzpaz (changes requested):

src/Mod/PartDesign/PartDesignTests/TestInvoluteGear.py
wwmayer: It's a copy&paste issue. As far a I see all PD test scripts use this format of the license text. So, it should be fixed in a separate commit.

wwmayer:
When running the unit tests I get a failure in the function testCustomizedGearProfile. Error message is:
```
TestInvoluteGear.py", line 70, in testCustomizedGearProfile
    self.assertAlmostEqual(outer_diameter, bb.XLength, delta=0.01, msg="Wrong horizontal extend.")
  File "/usr/lib/python3.6/unittest/case.py", line 878, in assertAlmostEqual
    raise self.failureException(msg)
AssertionError: 14 != 13.980091909670271 within 0.01 delta : Wrong horizontal extend.
```

Tests with the (OCC) bounding box expansions are a bit tricky because it's not guaranteed that they exactly fit.

jbaehr:
> Tests with the (OCC) bounding box expansions are a bit tricky because it's not guaranteed that they exactly fit.

I see... I also observed the bounding box slightly smaller than expected but thought this was due to the fact that the gear's top land is generated as a straight line instead of an arc. That's why I used this 0.01 tolerance, which is already way larger than `Precision::Confusion`. Didn't know that OCCs BB is somewhat loose and not portable/reliable.

I'll change this to a test with intersecting circles: Below the root circle and above the tip circle there should not be an intersection, but the pitch circle should. 

 (btw: Are the methods/constants from OCC's Precision header exposed to python somewhere?)

wwmayer:
Here some background information about bounding boxes: https://opencascade.blogspot.com/2009/04/all-inclusive-bounding-box.html

> (btw: Are the methods/constants from OCC's Precision header exposed to python somewhere?)

No, I don't think so. But it's good idea to add them, too. Just wait a little bit...

wwmayer:
... here we are: 20b86e55b8

src/Mod/PartDesign/PartDesignTests/TestInvoluteGear.py
wwmayer: With 6461f9f6d2 I made the license text uniform of the test modules. So can you adjust it in your file too, please?

jbaehr: done

jbaehr:
Thanks for all the replies. I've now changed the bounding box based check to an intersection based one. And thereby already used the new Precision export.

I'm not sure about the internal precision of the `distToShape` method, though. If the results are obtained analytically, then the check with `Precision::Intersection` is fine, maybe even required to ensure tangents with similar curvature are correctly handled. However, if there is some numerical algorithm behind all this, my check for intersection may need refinement. -- Anyway, for the specific application in the gear's test it's irrelevant, but this could become interesting when extracting these helpers and reusing them elsewhere.

benj5378:
I believe that `testInvoluteGear` should be added to __init__.py? I guess it's an issue for all the others modules too...

wwmayer:
@benj5378 
> I believe that testInvoluteGear should be added to __init__.py? I guess it's an issue for all the others modules too...

Why? At least the way it's done the test system can import all needed modules.

wwmayer:
> It's an improvement suggestion. Let me know if I am incorrect.

OK, I see. Yes, it would be an option to adjust __init__.py accordingly. But I will address it after merging the PR.

wwmayer:
f9b37729fe

benj5378:
@wwmayer Cool, that sounds good. I agree that this mistake should not block the PR since it is minor and espacially as its missing in many FreeCAD python modules.

Without being sure, I believe the file should also be named `__init__.py`

jbaehr:
> If you need to use `PartDesignTests`, you currently need to write: `from PartDesignTests import testInvoluteGear, test2, test3, test4, etc`. By creating the `__init__.py` file correctly, all test from `PartDesignTests` can be imported by `import PartDesignTests`
> 

Honestly, I don't see a need to this. Feel free to convince me otherwise, though. Here are my arguments:

First, from my understanding, the python way prefers explicit over implicit. Having a very slim, often empty, `__init__.py` is common practice; even recommended, e.g. by the famous [Hitchhiker’s Guide to Python](https://docs.python-guide.org/writing/structure/#packages) (which gives some rationales for this, too).

Then, for whom does this makes life easier? I mean, who needs to consume test classes other than the test runner. We already to add them to `TestPartDesignApp.py` (or `TestPartDesignGui.py` in the case they need a GUI). I mean, where are all the `import PartDesignTests` statement you need to optimise?

> f9b3772

Now we have again one more place to remember where new tests have to be added, too. I suggest to revert this, unless I've overseen some benefits that justify this double maintenance.

wwmayer:
> Cool, that sounds good. I agree that it should not block the PR since it is minor and espacially as its already missing in many FreeCAD python modules.

I don't think there is anything wrong with leaving the __init__.py files empty. Many other Python packages leave it empty, too.

> Now we have again one more place to remember where new tests have to be added, too. I suggest to revert this, unless I've overseen some benefits that justify this double maintenance.

That's actually a valid point.

After all I don't mind if we have the stuff inside __init__.py or not. But I also agree that I don't see a real benefit to have it.

benj5378:
@jbaehr

I think that you have some good points. Though having done some searches on Google, I don't know if it's considered good practice, it seems more like people have different opinions and it changes from case to case.

I think you have some good points.

PR #5389, opened 2022-01-17
Sketcher: task tool settings widget

PaddleStroke:
This feature adds a widget to Task in sketcher.
This enables user to type dimensions of geometries and create constraints accordingly.
The goal is to speed up sketching.

Feature discussed here : https://forum.freecadweb.org/viewtopic.php?f=9&t=65279
Video showing the feature : https://youtu.be/K8oTQNpI8CQ

One thing to discuss : What key to use for polyline mode switch. 'M' can't be used anymore because it's typed in the spinbox. So I changed to shift. But shift may be in use by others?

Please let me know if you have any question/feedback.


src/Mod/Sketcher/Gui/TaskSketcherTool.h
```diff
@@ -0,0 +1,111 @@
+/***************************************************************************
+ *   Copyright (c) 2009 Jürgen Riegel <juergen.riegel@web.de>              *
```
davidosterberg, line 2: Your name. Jurgen has left the project. He does not need more files in his name.

src/Mod/Sketcher/Gui/TaskSketcherTool.h
```diff
@@ -0,0 +1,111 @@
...
+
+public:
+    TaskSketcherTool(ViewProviderSketch *sketchView);
+    ~TaskSketcherTool();
+    /// Observer message from the Selection
+    void OnChange(Gui::SelectionSingleton::SubjectType &rCaller,
+                  Gui::SelectionSingleton::MessageType Reason);
+    SketcherToolWidget* widget;
+
+//public Q_SLOTS:
```
davidosterberg, line 94: No point in having commented out code here.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1014,6 +1014,8 @@ bool ViewProviderSketch::mouseMove(const SbVec2s &cursorPos, Gui::View3DInventor
 {
     // maximum radius for mouse moves when selecting a geometry before switching to drag mode
     const int dragIgnoredDistance = 3;
```
davidosterberg, line 1016: Maybe this should be configurable. You can add a parameter that can be set with the parameter editor.

davidosterberg (changes requested):
Looks good. Some minor comments. The question of the shift key for the polyline tool is a bit tricky.

src/Mod/Sketcher/Gui/TaskSketcherTool.h
luzpaz: > He does not need more files in his name.  

I don't disagree with the point but I think the tone could be perceived as a little irreverent.

PaddleStroke:
Thanks.
I made the modifications mentioned and solved the over-constraint bug with perpendicular mode of polyline that can be seen in the video.

The last question is for polyline mode shortcut.
For shift : is there other shortcuts using it beside shift + RMB for rotation? If it's the only issue then we can try to catch the +RMB in the polyline tool and cancel the mode change of the tool if + RMB is done.
Beside we can use another key than shift. What about the F3-F12 keys? Or Screenlock or pause (though most laptop won't have those). Other key ideas?

PaddleStroke:
It seems the shortcuts are currently being refactored. So it might be better to keep shift for now until the global shortcut refactoring is done.
Beside the only conflict is with rotating with shift + rmb which is currently not working properly in sketcher edit mode. So it's kind of niche issue and might be solved anyway by the refactoring.

src/Mod/Sketcher/Gui/TaskSketcherTool.h
davidosterberg: That was not the intention. He is just not the author of this file. He probably authored the template this was based on. But the main contributor is Paddle so he should be mentioned. Also the year should probably be updated.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5389">pipeline status</a> for feature branch [PR_5389](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5389). Pipeline [454040201](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454040201) was triggered at [cbc579a](https://github.com/FreeCAD/FreeCAD/pull/5389/commits/cbc579acd56379d8da54d45ad5dd24e0d1a0f7bd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
any particular reason for closing this one? :(

PaddleStroke:
It's because I build constrain contextually on top of ToolSettings. And I made further improvement of toolsettings on that ConstrainContextuall branch. 
So this branch was kind of obsolete. All should be merged in the other branch.

adrianinsaval:
considering past statements from the maintainers, I think they would prefer to merge it separately as this feature is independent of the other but better to wait for what they have to say. I think abdullah is who usually reviews sketcher stuff but he's not around for now.

luzpaz:
@abdullahtahiriyo has had IRL (work etc...) stuff going on. But he pops in here and there. Hopefully he can comment some of @PaddleStroke's work soon

PaddleStroke:
> considering past statements from the maintainers, I think they would prefer to merge it separately as this feature is independent of the other but better to wait for what they have to say. I think abdullah is who usually reviews sketcher stuff but he's not around for now.

Yes I know but I had to make lot of modification and fixes to toolsettings while developping constrain contextually. And I couldn't checkout of branches everytime I needed to make a small change (because it takes 1 hour to rebuild everytime). So I ended up making the changes to ToolSettings along the development of Constrain contextually. Which in the end was a lot. 
And as Constrain Contextually builds on top of ToolSettings, it had to integrate that branch in the first place.

adrianinsaval:
are you mixing stuff within the same commit? if you have separate commits for stuff only related to this widget it would be possible to cherry pick only those. You are developing on windows right? Is there really no equivalent to ccache on windows? That would be a shame, with ccache compilation should require just a few minutes if the difference isn't big.

PaddleStroke:
I didn't paid attention to the commit's content unfortunately.
With visual studio it builds only what you modified (so 1-2min for sketcher
modif).
But when I checkout of branches it rebuilds the whole thing. Making
checkouts painful.

abdullahtahiriyo:
Ok, I will handle it with the other commit. It gets some time before learning how to use git effectively.

PR #5391, opened 2022-01-17
[FEM] Add reference highlighting for displacement constraints

AjinkyaDahale:
This PR when ready is intended to highlight the geometries on which a particular FEM constraint is applied, when in the constraint's taskview.

Forum post: https://forum.freecadweb.org/viewtopic.php?f=18&t=65412.

### DIFF
--- a/src/Mod/Fem/Gui/CMakeLists.txt
+++ b/src/Mod/Fem/Gui/CMakeLists.txt
@@ -42,6 +42,7 @@ link_directories(${SMESH_LIB_PATH})
 set(FemGui_LIBS
     Fem
     FreeCADGui
+    PartGui
 )
 
 
@@ -220,6 +221,8 @@ SET(FemGui_SRCS_ViewProvider
     FemSelectionGate.h
     ViewProviderFemConstraint.cpp
     ViewProviderFemConstraint.h
+    ViewProviderFemConstraintOnBoundary.cpp
+    ViewProviderFemConstraintOnBoundary.h
     ViewProviderFemConstraintBearing.cpp
     ViewProviderFemConstraintBearing.h
     ViewProviderFemConstraintFixed.cpp
--- a/src/Mod/Fem/Gui/TaskFemConstraintOnBoundary.cpp
+++ b/src/Mod/Fem/Gui/TaskFemConstraintOnBoundary.cpp
@@ -59,10 +59,12 @@ using namespace Gui;
 TaskFemConstraintOnBoundary::TaskFemConstraintOnBoundary(ViewProviderFemConstraint *ConstraintView, QWidget *parent, const char* pixmapname)
     : TaskFemConstraint(ConstraintView, parent, pixmapname)
 {
+    ConstraintView->highlightReferences(true);
 }
 
 TaskFemConstraintOnBoundary::~TaskFemConstraintOnBoundary()
 {
+    ConstraintView->highlightReferences(false);
 }
 
 void TaskFemConstraintOnBoundary::_addToSelection(bool checked)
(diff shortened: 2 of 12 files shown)


AjinkyaDahale:
@wwmayer tagging you since you recently created the class `ReferenceHighlighter` that I am using here, and which I'm not able to import properly.

AjinkyaDahale:
Possibly fixed the issue by adding `PartDesignGui` to the libraries in `/Mod/Fem/Gui/CMakelists.txt`.

AjinkyaDahale:
> Now when looking at the implementation of ReferenceHighlighter there is actually nothing that would forbid it to move it to the Part module. 

That thought did occur to me trying to fix the issue. Do you want to make the move yourself? Otherwise I can make a PR to precede this one.

> And since it does not even use any particular GUI stuff it could be even moved to Part/App.

I would vote against that, since this appears to be a very GUI-centric item and the dependency on Part/Gui already exists.

wwmayer:
OK I will move it to Part/Gui, then.

AjinkyaDahale:
> OK I will move it to Part/Gui, then.

Thank you. Could you also add the methods to highlight vertices? Some FEM constraints can work on vertices.

wwmayer:
See af868ed079

wwmayer:
> Could you also add the methods to highlight vertices? Some FEM constraints can work on vertices.

What exactly do you mean?

AjinkyaDahale:
> > Could you also add the methods to highlight vertices? Some FEM constraints can work on vertices.
> 
> What exactly do you mean?

Like say `ReferenceHighlighter::getVertexColors(subSet.second, colors)`.

wwmayer:
Ah, OK.

AjinkyaDahale:
I think this is ready for review apart from a few kinks, which I can add as additional commits.

wwmayer:
Here we go: 74fe1fa4db

But note the code is not tested yet because I didn't have a sandbox to test it. The function getVertexColors() is supposed to accept the sub-elements Vertex, Edge, Wire and Face. So the array of string can e.g. be: {Vertex1, Vertex2, Edge4, Wire1, Face7}

AjinkyaDahale:
> Here we go: [74fe1fa](https://github.com/FreeCAD/FreeCAD/commit/74fe1fa4db96ce9ba63a156515ddf816117c7a6e)
 
Thanks. Rebased.

> But note the code is not tested yet because I didn't have a sandbox to test it.

I should be able to try it with this PR's branch.

> The function getVertexColors() is supposed to accept the sub-elements Vertex, Edge, Wire and Face. So the array of string can e.g. be: {Vertex1, Vertex2, Edge4, Wire1, Face7}

Sounds useful, albeit I can't try it with this PR. Why are we not including Shells and Solids, though?

wwmayer:
> Why are we not including Shells and Solids, though?

Because shells and solids are not supported by the selection mechanism.

wwmayer:
Merged as ec00d35c5f

AjinkyaDahale:
> Merged as [ec00d35](https://github.com/FreeCAD/FreeCAD/commit/ec00d35c5f9033f1e252416937bbe5266e2af94b)

Thanks. I see you made some changes in `PROPERTY_SOURCE` for the cpp's. I tried the same, but I think I was having errors opening any file after that. Were you able to fix that? If yes how?

wwmayer:
> but I think I was having errors opening any file after that

I have opened the project files from the examples directory without problems.

> Were you able to fix that? If yes how?

Changing the inheritance of the types shouldn't cause any problems w.r.t. existing project files because this information is not stored inside the Document.xml/GuiDocument.xml there.

In case you can reproduce problems then please open a forum topic and ping me there.

AjinkyaDahale:
@wwmayer thanks for the reply

> In case you can reproduce problems then please open a forum topic and ping me there.

I sure will. So far I cannot reproduce the issue but you'll be the first to know if I do.

PR #5398, opened 2022-01-20
[Sketcher] Disable constraints selection when geometry/constraint handler is active

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=10&t=65465

Set as Draft PR time it's tested

### DIFF
--- a/src/Mod/Sketcher/Gui/EditModeCoinManager.cpp
+++ b/src/Mod/Sketcher/Gui/EditModeCoinManager.cpp
@@ -627,6 +627,11 @@ void EditModeCoinManager::updateColor(const GeoListFacade & geolistfacade)
     pEditModeConstraintCoinManager->updateConstraintColor(constraints);
 }
 
+void EditModeCoinManager::setConstraintSelectability(bool enabled /* = true */)
+{
+    pEditModeConstraintCoinManager->setConstraintSelectability(enabled);
+}
+
 void EditModeCoinManager::createEditModeInventorNodes()
 {
     // 1 - Create the edit root node
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -364,6 +364,7 @@ void ViewProviderSketch::activateHandler(DrawSketchHandler *newHandler)
     sketchHandler = std::unique_ptr<DrawSketchHandler>(newHandler);
     Mode = STATUS_SKETCH_UseHandler;
     sketchHandler->sketchgui = this;
+    sketchHandler->preActivated(this);
     sketchHandler->activated(this);
 
     // make sure receiver has focus so immediately pressing Escape will be handled by
@@ -382,10 +383,11 @@ void ViewProviderSketch::deactivateHandler()
         drawEdit(editCurve); // erase any line
         resetPositionText();
         sketchHandler->deactivated(this);
+        sketchHandler->postDeactivated(this);
         sketchHandler->unsetCursor();
         sketchHandler = nullptr;
     }
-
+    DrawSketchHandler::postDeactivatedAlwaysRun(this);
     Mode = STATUS_NONE;
 }
 
@@ -3288,6 +3290,11 @@ QIcon ViewProviderSketch::mergeColorfulOverlayIcons (const QIcon & orig) const
 
 
 /*************************** functions ViewProviderSketch offers to friends such as DrawHandlerSketch ************************/
+
+void ViewProviderSketch::setConstraintSelectability(bool enabled /*= true*/)
+{
+    editCoinManager->setConstraintSelectability(enabled);
+}
 
 void ViewProviderSketch::setPositionText(const Base::Vector2d &Pos, const SbString &text)
 {
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -467,6 +467,11 @@ bool ViewProviderSketch::keyPressed(bool pressed, int key)
             }
             return false;
         }
+    case SoKeyboardEvent::LEFT_SHIFT:
+        if (Mode < STATUS_SKETCH_UseHandler) {
+            editCoinManager->setConstraintSelectability(!pressed);
+            return true;
+        }
     default:
         {
             if (isInEditMode() && sketchHandler)
--- a/src/Mod/Sketcher/Gui/DrawSketchHandler.cpp
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandler.cpp
@@ -61,6 +61,11 @@ using namespace Sketcher;
 
 /************************************ Attorney *******************************************/
 
+inline void ViewProviderSketchDrawSketchHandlerAttorney::setConstraintSelectability(ViewProviderSketch &vp, bool enabled /*= true*/)
+{
+    vp.setConstraintSelectability(enabled);
+}
+
 inline void ViewProviderSketchDrawSketchHandlerAttorney::setPositionText(ViewProviderSketch &vp, const Base::Vector2d &Pos, const SbString &txt)
 {
     vp.setPositionText(Pos,txt);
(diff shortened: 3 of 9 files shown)


PaddleStroke:
Works great!

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5398">pipeline status</a> for feature branch [PR_5398](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5398). Pipeline [454039957](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454039957) was triggered at [53c9c7a](https://github.com/FreeCAD/FreeCAD/pull/5398/commits/53c9c7ab8ae75049d655c573ce5de5656f16b213). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
builds, runs and appears to work as intended, it's a good improvement. I don´t know of any circumstance where you would want to select or preselect constraints whit a tool active. Is there anything specific I should test?

0penBrain:
Push another commit that also disables constraint selection while user is holding "Shift" pressed.

@adrianinsaval Sorry for late reply, nothing specific to test :)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5398">pipeline status</a> for feature branch [PR_5398](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5398). Pipeline [462996914](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/462996914) was triggered at [cf137a3](https://github.com/FreeCAD/FreeCAD/pull/5398/commits/cf137a342cf678c25e051b81c18521ccaf2fd410). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/EditModeCoinManager.cpp
```diff
@@ -627,6 +627,16 @@ void EditModeCoinManager::updateColor(const GeoListFacade & geolistfacade)
     pEditModeConstraintCoinManager->updateConstraintColor(constraints);
 }
 
+void EditModeCoinManager::setConstraintSelectability(bool enabled /* = true */)
+{
+    if (enabled) {
```
abdullahtahiriyo, line 632: I think this code should be moved to a function with the same name in EditModeConstraintCoinManager and this function should call that one. 

The reason is the SRP (single responsibility principle). EditModeConstraintCoinManager is responsible for all node management relating to constraints.

src/Mod/Sketcher/Gui/EditModeCoinManager.cpp
```diff
@@ -745,6 +755,7 @@ void EditModeCoinManager::createEditModeInventorNodes()
 
     // coin nodes for the constraints +++++++++++++++++++++++++++++++++++++++++++++++++++
     pEditModeConstraintCoinManager->createEditModeInventorNodes();
+    setConstraintSelectability(); // Ensure default value;
```
abdullahtahiriyo, line 758: Idem. Initialisation is the responsibility of EditModeConstraintCoinManager.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -366,6 +366,8 @@ void ViewProviderSketch::activateHandler(DrawSketchHandler *newHandler)
     sketchHandler->sketchgui = this;
     sketchHandler->activated(this);
 
+    editCoinManager->setConstraintSelectability(false);
```
abdullahtahiriyo, line 369: I tend to think that it is not the ViewProvider's call to decide on constraint selectability. Managing the interaction with the rest of the scenograph is a responsibility of DrawSketchHandler (and the corresponding derived class implementing the command). 

Surely DrawSketchHandler can set a default that can be overriden by any derived class that finds it useful to select a constraint. For example, PaddleStroke's contextual constraining might benefit from the ability to edit the value of a datum constraint during execution (for example to correct a previously wrongly introduced one).

If you agree, a private function setConstraintSelectability could be created in ViewProviderSketch (we do not want anyone with a pointer tweaking functionality that under SRP belongs to DrawSketchHandler). Then there is an attorney class to regulate the level of friendship of DrawSketchHandler and ViewProviderSketch (ViewProviderSketchDrawSketchHandlerAttorney) that should implement this. DrawSketchHandler accesses private stuff of ViewProviderSketch via this attorney. I put in place this mechanism to force SRP between VP and DrawSketchHandler. The classes are very much related and should be partially coupled, but as loosest as possible. This attorney is there to kill the tentation of mixing responsibilities.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -385,7 +387,7 @@ void ViewProviderSketch::deactivateHandler()
         sketchHandler->unsetCursor();
         sketchHandler = nullptr;
     }
-
+    editCoinManager->setConstraintSelectability(true);
```
abdullahtahiriyo, line 390: Idem

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -465,6 +467,9 @@ bool ViewProviderSketch::keyPressed(bool pressed, int key)
             }
             return false;
         }
+    case SoKeyboardEvent::LEFT_SHIFT:
```
abdullahtahiriyo, line 470: This is handled when there is a DrawSketchHandler active and when there is not?

One might think that LEFT_SHIFT is too precious to be hardcoded at VP level. I would tend to think that it should be spared for DrawSketchHandler (and its offspring). Maybe I am wrong. Feel free to ping me in the forum.

abdullahtahiriyo (commented):
I think the functionality at EditCoinManager is desirable. I think that DrawSketchHandler should take over the responsibility of deciding when to make constraints pickable and when not. I think we should discuss whether LEFT_SHIFT should be spared for DrawSketchHandler to decide what exactly to do when pressed.

0penBrain:
@abdullahtahiriyo Thx for review. Will have a deeper look and update the PR (hopefully in coming days).

src/Mod/Sketcher/Gui/EditModeCoinManager.cpp
0penBrain: Done

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
0penBrain: I guess you're right. My new proposal is based on `Mode`. Let's see if you like it. :)

0penBrain:
> I think the functionality at EditCoinManager is desirable. I think that DrawSketchHandler should take over the responsibility of deciding when to make constraints pickable and when not. I think we should discuss whether LEFT_SHIFT should be spared for DrawSketchHandler to decide what exactly to do when pressed.

OK, I finished taking into account review (which by all points seems rightful to me). Admittedly, attorneys and the whole shebang pushes complexity beyond what I can easily handle, so please review again carefully the new proposal. ;)

abdullahtahiriyo:
I will make a separate commit to make the DrawSketchHandler pre post functions into an NVI pattern. Thanks for taking the time to modify it. :)

abdullahtahiriyo:
Take a look here:
https://github.com/FreeCAD/FreeCAD/blob/d8c052c666ca9394914588c9107fbb192881f2fd/src/Mod/Sketcher/Gui/DrawSketchHandler.h#L100

https://github.com/FreeCAD/FreeCAD/blob/d8c052c666ca9394914588c9107fbb192881f2fd/src/Mod/Sketcher/Gui/DrawSketchHandler.h#L131

https://github.com/FreeCAD/FreeCAD/blob/d8c052c666ca9394914588c9107fbb192881f2fd/src/Mod/Sketcher/Gui/DrawSketchHandler.cpp#L124

https://github.com/FreeCAD/FreeCAD/blob/d8c052c666ca9394914588c9107fbb192881f2fd/src/Mod/Sketcher/Gui/DrawSketchHandler.cpp#L132

The NVI is really appropiate here, because the pre and post functions were leaking implementation details in ViewProviderSketch. ViewProviderSketch should not care about how DrawSketchHandler manages its default behaviour. NVI forces a given execution sequence (because the public function is not virtual). The steps of the sequence are virtual and private, so they do not generate interface, yet allow the inheritance to override them.

PR #5407, opened 2022-01-21
Sketcher: Add 'Constraint Contextually' feature

PaddleStroke:
Add a constrain tool that adapts the constraint based on selected geometries.
See the forum post for videos and discussion : 
https://forum.freecadweb.org/viewtopic.php?p=563753#p563753

Note it builds on top of my Tool Widget branch which is a pending PR.
https://tracker.freecadweb.org/view.php?id=1416

---


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [454040377](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454040377) was triggered at [bfadfd8](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/bfadfd808df830ae38f9ab4d5170ab518514c1e8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [454285309](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/454285309) was triggered at [1edb431](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/1edb4314978da054459c210d64b698092f6d5a30). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [455229713](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/455229713) was triggered at [9b1c6c8](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/9b1c6c8d5e777ae23139d80c4004b094c731e25d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [456256881](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/456256881) was triggered at [03dec5c](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/03dec5c36bdec2082c5eb0c4bd8e0cc3c7b1d2e5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [456431088](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/456431088) was triggered at [27a676a](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/27a676adf0002d23a4b04161ed154cfdb6bbd3ad). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [457077917](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/457077917) was triggered at [6f439c9](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/6f439c9250293dc3e83c0b96f6d173a7867c75cc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [458320637](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/458320637) was triggered at [e9a8454](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/e9a8454d225fa70416dd04db3736af3905ef2cb9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [458653541](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/458653541) was triggered at [6e23540](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/6e235401aad0cf7b6dcb2e13b3808215a480bc90). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/TaskDlgEditSketch.cpp
```diff
@@ -64,6 +64,8 @@ TaskDlgEditSketch::TaskDlgEditSketch(ViewProviderSketch *sketchView)
     Content.push_back(Constraints);
     Content.push_back(Elements);
 
+    //TO DO: find a way to showGroupBox to be able to collapse/expand from tasksketchertool.cpp
```
luzpaz, line 67: `s/TO DO/TODO/`

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [459476269](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/459476269) was triggered at [15e8088](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/15e80886e1d5d09d9e10e0e6d65a3aaf810651fb). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [461151106](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/461151106) was triggered at [a288624](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/a288624b48e030dbf673f370bfd49c10356f5ced). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
I have taken a look to the modifications to ViewProviderSketch. 

1) The modifications originating in OpenBrain's commit should consider the review of OpenBrain's PR. That PR will be merged before this. This PR will need to be rebased on that one.

2) I do not favour data members with public access specifier (save for exceptions by design, such as properties). This has the potential to lead to very strong coupling and kills separation of responsibilities. It further leaks implementation details.

3) Although I have not looked into the whole with the necessary detail, it appears as the ViewProvider is being used as a carrier to store variables it should not even be aware of.

These issues are not problems that cannot be solved. However, they are serious issues that need to be tackled before merge. After OpenBrains' PR is merged and this is rebased, we will take a good look how we can solve them.

@PaddleStroke 

I think it would be great if you take a look to the comments I have made in OpenBrain's PR.

I think it is important that we all have a common understanding of the responsibilities of each component and the architectural barriers in place to prevent these responsibilities from spreading over different components (and where no specific barriers are set for flexibility and avoid a too complex design and mostly YANGNI, but special care should be taken so that responsibilities and delegations are mantained, this is the case of EditCoinManager and EditConstraintCoinManager, for example).

src/Mod/Sketcher/Gui/ViewProviderSketch.h
```diff
@@ -535,6 +542,7 @@ class SketcherGuiExport ViewProviderSketch : public PartGui::ViewProvider2DObjec
     /// signals if the elements list has changed
     boost::signals2::signal<void ()> signalElementsChanged;
     //@}
+    TaskSketcherTool* toolSettings;
```
abdullahtahiriyo, line 545: Strong code smell. Why does the ViewProvider need to store a pointer to a dialog in the taskbar? Worse, why does it need to expose it via its public interface to anybody with a pointer?

It appears that the VP does not need this information and it is only here as a convenient container.

If you need to signal something within the responsibility of the VP to a dialog (for example in the taskbar), you may use a boost::signals2 (see signalSetup in ViewProviderSketch for an example). But it does not appear to be here the case.

src/Mod/Sketcher/Gui/ViewProviderSketch.h
```diff
@@ -523,6 +527,9 @@ class SketcherGuiExport ViewProviderSketch : public PartGui::ViewProvider2DObjec
     virtual bool mouseButtonPressed(int Button, bool pressed, const SbVec2s& cursorPos, const Gui::View3DInventorViewer* viewer) override;
     //@}
 
+    /// moves a selected constraint
```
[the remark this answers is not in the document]
abdullahtahiriyo: See below for the removal of access specifier

src/Mod/Sketcher/Gui/ViewProviderSketch.h
```diff
@@ -635,9 +643,6 @@ class SketcherGuiExport ViewProviderSketch : public PartGui::ViewProvider2DObjec
     /// snap points x,y (mouse coordinates) onto grid if enabled
     void snapToGrid(double &x, double &y);
 
-    /// moves a selected constraint
-    void moveConstraint(int constNum, const Base::Vector2d &toPos);
```
abdullahtahiriyo, line 639: Ha! Trying to bypass the checkpoint? ;) 

If this is for DrawHandler, you need to use the attorney.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -276,6 +276,8 @@ SbVec2s ViewProviderSketch::DoubleClick::prvClickPos; //used by double-click-det
 SbVec2s ViewProviderSketch::DoubleClick::prvCursorPos;
 SbVec2s ViewProviderSketch::DoubleClick::newCursorPos;
 
+SbVec2s ViewProviderSketch::prvMoveCursorPos;
```
abdullahtahiriyo, line 279: Same comments as in .h

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -366,6 +368,8 @@ void ViewProviderSketch::activateHandler(DrawSketchHandler *newHandler)
     sketchHandler->sketchgui = this;
     sketchHandler->activated(this);
 
+    editCoinManager->setConstraintSelectability(false);
```
abdullahtahiriyo, line 371: Refer to comments on OpenBrain's PR.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -385,7 +389,7 @@ void ViewProviderSketch::deactivateHandler()
         sketchHandler->unsetCursor();
         sketchHandler = nullptr;
     }
-
+    editCoinManager->setConstraintSelectability(true);
```
abdullahtahiriyo, line 392: Refer to comments on OpenBrain's PR.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1014,6 +1020,8 @@ bool ViewProviderSketch::mouseMove(const SbVec2s &cursorPos, Gui::View3DInventor
 {
     // maximum radius for mouse moves when selecting a geometry before switching to drag mode
     const int dragIgnoredDistance = 3;
+    prvMoveCursorPos = cursorPos;
+    prvMoveViewer = viewer;
```
abdullahtahiriyo, line 1024: Same comments and in .h for these two.

There is a further aspect. When someone looks at this code, it appears to do nothing. This gets stored and never used. The coupling with another class via the public interface is not obvious. This also affects code readability, maintainability and increases the chances that somebody else, not understanding why this is here, may removing it or change it and break the functionality on the other end.

abdullahtahiriyo (commented):
I have done a partial review of ViewProvider.

I have tried to make explicit the reasons why it looks wrong. The reason is that I want to help you understand what I think is important. You do a lot of right things and I think taking the time now will be very helpful for the future. I probably did worse when I started ;)

src/Mod/Sketcher/Gui/ViewProviderSketch.h
```diff
@@ -38,6 +38,7 @@
 #include <QCoreApplication>
 #include <Gui/Document.h>
 #include "ShortcutListener.h"
+#include "TaskSketcherTool.h"
```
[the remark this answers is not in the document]
PaddleStroke: Thanks for your feedbacks! 
For most points you raised I was expecting it. I did those because I didn't knew how it should be done properly.
For this it is because I didn't knew how to add my TaskSketcherTool class to the task bar from within sketchHandlers. Initially I wanted to open add the widget when a tool is launched and destroy it when it closes. But I couldn't find how to do it. So I did as the other widgets are added and declared the class in VP in order to be able to access it from the SketchHandler.

src/Mod/Sketcher/Gui/ViewProviderSketch.h
```diff
@@ -407,6 +408,9 @@ class SketcherGuiExport ViewProviderSketch : public PartGui::ViewProvider2DObjec
     App::PropertyString EditingWorkbench;
     //@}
 
+    static SbVec2s prvMoveCursorPos;
+    Gui::View3DInventorViewer * prvMoveViewer;
```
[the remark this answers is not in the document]
PaddleStroke: I made these two because I want to trigger a mouseMove of the SketchHandler from within the ToolSettings Widget to update the tool when something is typed in.
See TaskSketcherTool.cpp line 141 : 
sketchView->mouseMove(sketchView->prvMoveCursorPos, sketchView->prvMoveViewer);
It felt so bad doing this but I was at a loss. It's linked to the previous thing that the widget class is declared in VP.

src/Mod/Sketcher/Gui/ViewProviderSketch.h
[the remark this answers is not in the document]
PaddleStroke: Yes it's because I want DrawSketchHandler onMouseMove to move the constraint. I don't know what is an attorney and encapsulation doesn't ring a bell either.

src/Mod/Sketcher/Gui/ViewProviderSketch.h
PaddleStroke: It is totally as a convenient container. It's because I need to access the dialog from the DrawSketchHandler.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
PaddleStroke: Right. I'll remove this commit from my branch when his PR merges and will rebase on master.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
PaddleStroke: I was pretty sure it wasn't correct and that you would catch it :) 
But I couldn't find a proper solution at the time.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [464825515](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/464825515) was triggered at [3a70145](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/3a701451986ce07d871c63063e5ca98940e3a957). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [465402034](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/465402034) was triggered at [9ad25a9](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/9ad25a971ce5e38749ba3a7add6acce03f0f77f3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/ViewProviderSketch.h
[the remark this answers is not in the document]
abdullahtahiriyo: Let's start here:
https://forum.freecadweb.org/viewtopic.php?style=3&f=10&t=66095

src/Mod/Sketcher/Gui/ViewProviderSketch.h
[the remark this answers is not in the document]
abdullahtahiriyo: I am still unsure about the tool in the taskbar. However, in order to do what you want, you do not need to modify the VP. You can access the task dialog from anywhere via:
```
Gui::TaskView::TaskDialog *dlg = Gui::Control().activeDialog();
TaskDlgEditSketch *sketchDlg = qobject_cast<TaskDlgEditSketch *>(dlg); 
```
So you can access the tool or some functionality of the tool via this pointer.

If the tools are going to be able to use this dialog, then it makes sense to give DrawSketchHandler the responsibility of registering/deregistering with the tool dialog. It can then provide a default implementation, which gets overriden by the specific DrawSketchHandler that wants to make use of the dialog.

By doing it this way, any tool gains the ability to use the tool widget and adapt its contents to the specific tool.

Sounds ok?

src/Mod/Sketcher/Gui/ViewProviderSketch.h
abdullahtahiriyo: Then, the smell was right. Refer to the solution above to get the access you need.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
abdullahtahiriyo: I need to understand first what is exactly needed. It appears to me that the solution here is a new private function in VP (possibly reusing code of the move function), with a clear name that one knows is to force a mouse move, that is exposed to DrawSketchHandler via the corresponding attorney. Then, DrawSketchHandler takes over this functionality of forcing mouse move.

The part that I understand less, is that you say that you want the tool to move the mouse in the ViewProvider. Without looking at the actual code, that sounds like code smell (if a sound can smell...odd).  The Tool dialog, is representation and input. If upon input a reaction is necessary, it should notify the specific DrawSketchHandler (either via pointer of via signal, such as a boost signal). The specific DrawSketchHandler is the one who has the responsibility to cause information to be shown in the dialog and decide what to do with the input from the dialog. The specific DrawSketchHandler should use the functionality inherited from DrawSketchHandler to trigger this in the ViewProvider via the collaboration interface.

But, I am not sure to have understood the whole picture.

src/Mod/Sketcher/Gui/ViewProviderSketch.h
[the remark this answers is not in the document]
PaddleStroke: Oh thanks I spent so much time looking for this... And I ended up doing this dirty hack.

You mean give it to the main DrawSketchHandler  class rather than each class? So I should declare this pointer in the main DrawSketchHandler  class, then every sub class will be able to access the pointer directly correct?

PaddleStroke:
I think we need a "onParameterSet()" function in the drawsketchhandler then
connect the valuechanged Signal to this function. Right?

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
PaddleStroke: What is required is that upon input a reaction is necessary as you said. When user enter a box length, then press enter, the mousemove updates geometry by redrawing the drawEdit curve.
So when a parameter is set in the dialog, we need to be able to trigger the mouseMove function of the current DrawSketchHandler subclass.
Could you please advise how to do?
I think we could create a "onParameterSet" function, which could trigger when a spinbox has value changed. Then this function could call mouse move or do something else. 
But how to catch the signal from the dialog?

src/Mod/Sketcher/Gui/ViewProviderSketch.h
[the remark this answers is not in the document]
PaddleStroke: I tried to do that and I'm trying to initialize dlg as follow in DrawSketchHandler constructor, but it's not working I'm getting an illegal storage access when a tool starts (on first attempt to reach for toolSettings).

Is that not correct?

`DrawSketchHandler::DrawSketchHandler() : sketchgui(0), dlg(Gui::Control().activeDialog()), toolSettings(qobject_cast<TaskSketcherTool*>(dlg)) {}`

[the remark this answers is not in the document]
PaddleStroke: I think Gui::Control().activeDialog() may be the issue. Because the widget is not particularly active when you launch a tool. So why would it be considered to be the active dialog amongs all the Task widgets.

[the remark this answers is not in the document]
PaddleStroke: Ok I misunderstood. So what I need is : 

`DrawSketchHandler::DrawSketchHandler() : sketchgui(0), toolSettings(qobject_cast<TaskDlgEditSketch*>(Gui::Control().activeDialog())->toolSettings) {}`
But this tells me that toolSettings is inacessible (because it's declared in protected.) So I can fix it in putting the declare in TaskDlgEditSketch.h public section. But I guess you're not going to like it :)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [467795259](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/467795259) was triggered at [a87ad88](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/a87ad8827740faa42cf68921451582f72c46957b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [469288396](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/469288396) was triggered at [e2da048](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/e2da048d4925f2929f47056b2172e9e162f83bc0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
```diff
@@ -0,0 +1,911 @@
...
+    //sketchView->toolSettings->hideGroupBox();
+
+    //Give the focus back to the viewproviderSketcher
+    //Gui::MDIView* mdi = Gui::Application::Instance->activeDocument()->getActiveView();
+    //mdi->setFocus();
+    QMdiArea* mdi = qobject_cast<QMdiArea*>(Gui::MainWindow::getInstance()->centralWidget());
+    if (!mdi) return;
+    mdi->activeSubWindow()->widget()->setFocus();
+
+    switch (toolSelected) {
```
abdullahtahiriyo, line 270: This kind of solution is not very flexible for changes. For example, there is a checkbox signal exclusively for chamfer.

Ideally, this widget would have generic code only and the specific code would be provided by the specific DrawHandler (or a specific class associated to the DrawHandler).

src/Mod/Sketcher/Gui/Resources/icons/obsolete/Sketcher_CreatePointPolyChamfer.svg
```diff
@@ -0,0 +1,515 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
abdullahtahiriyo, line 1: Why does this PR add so many "obsolete" icons? If the icons are not necessary, they should be removed from the PR.

PaddleStroke:
Thanks for feedback.
Ok so the idea is to remove this setsettings function and just have a
resetsettings function here. And move the setsettings in each
drawsketchhandler?

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
```diff
@@ -211,6 +213,33 @@ unsigned long DrawSketchHandler::getCrosshairColor()
     return color;
 }
 
+bool DrawSketchHandler::distanceXYorPointOnObject(bool distanceXZeroYOne, int geoId, Sketcher::PointPos posId, double distance) {
```
abdullahtahiriyo: I cannot find any reason why a function declared in DrawSketchHandler.h would be defined in CommandCreateGeo.cpp. 

Either the code is generic at DrawSketchHandler level or it belongs elsewhere (helper class, intermediate derived class from DrawSketchHandler,...).

What are you trying to achieve?

PaddleStroke:
I'm not sure actually. I wanted to make to function available for other
classes than just constrain contextually as i thought it could be useful.
But this place doesn't make much sense I agree.

PaddleStroke:
I understand and was expecting it.
Tell me what needs to be done about the structure and I'll fix it.

Regarding the order, I would do :
1- tool setting
2- chamfer
3- slot

Because chamfer and slots are adapted to work with tool settings.
Especially chamfer for point preservation and advanced stuff. So it would
need to rewrite back to the original chamfer tool branch I have, which was
not ideal on the first place. So it would need a lot of work.

What do you think?

Regarding tool setting you mean we should start by applying it only to one
tool? Rectangle for instance?

abdullahtahiriyo:
> Regarding the order, I would do : 1- tool setting 2- chamfer 3- slot Because chamfer and slots are adapted to work with tool settings. Especially chamfer for point preservation and advanced stuff. So it would need to rewrite back to the original chamfer tool branch I have, which was not ideal on the first place. So it would need a lot of work. What do you think?

My original choice was precisely to avoid having a chamfer implementation adapted to use the specific interface of the tool settings. However, if it is just the toolwidget and the champfer tool (just that one), then it may be small enough.

The current implementation has problems limiting the growth of the toolwiget and the flexibility it can offer to different tools (aside from coupling and encapsulation problems).

- The logic for the toolwidget (how to react to a given tool) does not belong in the taskbar toolwidget class. It is part of the implementation of a specific DrawSketchHandler (which can be common for different commands). The taskbar toolwidget should only have generic code.
- The toolwidget in the taskbar is best, when it is just a placeholder that shows a widget provided dynamically. The widget associated to the DrawSketchHandler. This way, different tools with different needs can set a different adapted widget (for example, chamfer could provide one with an extra checkbox for corner preservation).

For this I am contemplating a solution in the lines of using a widget factory. 

In concept, you register a widget associated with a type or string (it is basically a dictionary / std::map). The factory creates and returns a widget associated to the string it is provided with. 

DrawSketchHandler has a virtual function getWidget returning an empty string (which corresponds to a tool that has no widget), each child that wants a widget overloads this function to return the string associated to the widget it wants.

When activating a DrawSketchHandler, generic code can retrieve this string and use it to create an appropriate widget (if not empty). This widget can be stored in a pointer in DrawSketchHandler, and the pointer passed (for example via a boost signal) to the taskbar. The taskbar substitutes the widget dynamically.

This means that the child of DrawSketchHandler has access to the widget and can directly interact with it.

If I did not fail in my design (it is just in my head), this would solve all the other architectural problems:
- The taskbar knows about widgets (and works with a generic class without knowing anything about the implementation).
- DrawSketchHandler does not need to know anything about the widget implementation either.
- The specific child of DrawSketchHandler, which implements the tool, can interact directly with the widget (via its public interface and can attach specific signals if necessary). This is the first object that needs to know the specific type of the widget. However, this is all right, because this is an implementation detail of the DrawSketchHandler child. This detail is encapsulated and nobody else should care about it.
- All ViewProvider necessary interaction goes through the collaboration interface of DrawSketchHandler.

When DrawSketchHandler is purged, the taskbar is signaled to remove and delete the dynamic widget.

Does this sound ok?

P.S: I think we have a widget factory somewhere in FreeCAD. I do not have time at this moment to have a look.

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
PaddleStroke: I reviewed this question and found out that I put the declaration here because this function is used by many DrawSketchHandlers.
To the question why would it be defined in CommandCreateGeo.cpp, I just replicated how other functions are done : setCrosshairColor, getCrosshairColor.
Though it might be better if it is implemented as functions in "/* helper functions" ?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [471759211](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/471759211) was triggered at [b9ced0e](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/b9ced0edd1b6070e735176de2c91d1d0458ce349). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
> Sounds good, I think I understand the general idea, but implementing this idea might be far over my head, reading your text raised a lot of '???' in my mind. I'm good at replicating but making something new like this is something else. Though I'm quite sick today so that might not help !
> Perhaps if you could implement the general structure then I can do the dirty job and adapt to all tools? If not then I'll have plenty of questions :)

It has come to my attention that you are good at replicating... now you have to become good at creating :) 

Follow the script. Prepare a very small PR, just the tool widget and the chamfer. This will be the start point. We start from there with questions and answers.

> My first questions :
> 
>     * What about toolwidget class ? Should it still be created, declared and appended to 'Content' in TaskDlgEditSketch?
>       In my understanding yes it should be created with an empty widget, then the widget is replaced by the one produced by the widget factory?

You need a class in the toolbar.

This class is the one that gets the information about the widget. 

In my half-cooked idea, this class will have a public function (boost slot) taking a string as parameter (the information about the widget). DrawSketchHandler will trigger the signal (e.g. signalActiveTool) and pass the string of the tool.

This class will take the string, call the factory to create the widget.

For the signal, there is a similar thing here:

https://github.com/FreeCAD/FreeCAD/blob/95f583a6774f1e77506f7d47ef7c453e26f37f94/src/Mod/Sketcher/Gui/TaskSketcherMessages.h#L50

https://github.com/FreeCAD/FreeCAD/blob/a46592d45ad146d2a42af313311779775d57838c/src/Mod/Sketcher/Gui/TaskSketcherMessages.cpp#L62

https://github.com/FreeCAD/FreeCAD/blob/a46592d45ad146d2a42af313311779775d57838c/src/Mod/Sketcher/Gui/ViewProviderSketch.h#L534

Do not worry too much, I will help you out.

> 
>     * DrawSketchHandler onactivate calls the widget factory, which returns a widget. How is it passed into the task dialog?

There are two options, a) to create the widget in DrawSketchHandler and pass it via a signal and b) create the widget in the class of the toolbar and pass a string as a signal.

From an interface point of view I favour b) because any function that wants to rely on which function is selected can easily subscribe to the signal and use the string. The widget is only useful for the specific case of the taskbar. This also unloads responsibilities from DrawSketchHandler.

Then your question morphs into how do I get the pointer to the widget that the class in the taskbar has created. Well, I have to give it to it some more thought, but it is probably ok to create a function in the TaskDlg, in the public interface, getActiveToolWidget(), to get a pointer from to the widget.

The signal will be in VPSketch. It will be triggered from DrawSketchHandler via the collaboration interface (and thus delegating the triggering to DrawSketchHandler).

The pointer from the taskbar can be requested from VPSketch  from a private function which is called from DrawSketchHandler via the collaboration interface. That will keep DrawSketchHandler and the Taskbar decoupled. The only change to the public interface is for ViewProviderSketch, that now has a signal to which anybody may subscribe to be informed of the tool that is selected.

We might also come up with a better solution as we advance.

We can work on like this:
1. Small PR widget+chamfer.
2. Add the signal
3. Modify the widget so that when signaled, it loads the widget of the string.
4. Add the code to get the pointer/reference of the widget back

Sounds ok?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407">pipeline status</a> for feature branch [PR_5407](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5407). Pipeline [472746964](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/472746964) was triggered at [15e0a80](https://github.com/FreeCAD/FreeCAD/pull/5407/commits/15e0a80b1d5e0af38ab580da0902fa1babbdcb6e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
@PaddleStroke 

With my time availability, I need to handle a PR after the other. 

There will be a feature freeze soon, so this will be merged in the next development cycle. That is not a problem, we can continue working on it and getting it right by then.

I understand that deciding on a design is not easy. To me the most important thing is that you understand why some things are a "no go" and which solutions can be applied to fix that. This will make your future commits more likely to be directly merged. There is also never just a single solution. I do not claim to have come with the best solution. Just one where (at least) I do not see particular problems. 

Let me know when you have you starting PR. We will answer questions one by one.

abdullahtahiriyo:
@PaddleStroke 

I have made some time to create a first approximation to a Framework for the toolwidget:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/tree/paddle_widget

Please try to just finish the DrawSketchHandlerLine, so that it does (all) what you want it to do. But just that one handler.

Some code is scaffolding.

You do not need to push the result here. Just push it to your fork and mention me. I will check the code then. We will iterate until the framework is solid.

PaddleStroke:
> 

Hi,
Awesome !! I'll jump in now.

abdullahtahiriyo:
@PaddleStroke 

Do not worry. Most likely I won't have time to review before the next weekend. We will cross the bridge when we get there.

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
```diff
@@ -0,0 +1,967 @@
...
+        else if (keyEvent->key() == Qt::Key_Space)
+        {
+            sketchView->keyPressed(1, SoKeyboardEvent::SPACE);
+            return true;
+        }
+    }
+    return false;
+}
+
+void SketcherToolWidget::emitTogglePreserveFilletChamferCorner(bool val) {
```
abdullahtahiriyo, line 136: This is just some comment intended for the new framework.

The ToolDefaultWidget in the new framework is generic. It is not a problem if you need to add some checkboxes. However, these checkboxes need to be generic as the Parameters, something like checkboxOne, checkboxTwo.

If a tool needs a totally different widget, then a new widget should be get from the factory (that will eventually be implemented).

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
```diff
@@ -0,0 +1,967 @@
...
+    ui->checkBoxPreserveFilletChamferCorner->onSave();
+}
+void SketcherToolWidget::emitSetparameter(double val)
+{
+    isSettingSet[0] = 1;
+    toolParameters[0] = val;
+    if (unitChangeTriggeredValueChanged) {
+        unitChangeTriggeredValueChanged = 0;
+    }
+    else setParameterFocus(1);
```
abdullahtahiriyo, line 146: We do not want to have magic numbers in the code. It is difficult to read and error-prone. General solutions are constants or enums. 

I think I have left this in the code temporarily, but will need to be changed.

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
```diff
@@ -0,0 +1,967 @@
...
+            return true;
+        }
+    }
+    return false;
+}
+
+void SketcherToolWidget::emitTogglePreserveFilletChamferCorner(bool val) {
+    ui->checkBoxPreserveFilletChamferCorner->onSave();
+}
+void SketcherToolWidget::emitSetparameter(double val)
```
abdullahtahiriyo, line 139: a slot does not emit, rather catches. It is important to use meaningful names.

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
```diff
@@ -0,0 +1,967 @@
...
+}
+
+void SketcherToolWidget::loadSettings()
+{
+    ui->checkBoxPreserveFilletChamferCorner->onRestore();
+}
+
+void SketcherToolWidget::setparameter(double val, int i)
+{
+    if (i == 0) {
```
abdullahtahiriyo, line 745: switch case is best adapted to this kind of structures checking an int for a value. It is much more readable and easy to maintain.

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
```diff
@@ -0,0 +1,967 @@
...
+    }
+    else setParameterFocus(4);
+
+    if (isWidgetActive) {
+        sketchView->mouseMove(sketchView->prvMoveCursorPos, sketchView->prvMoveViewer);
+    }
+}
+void SketcherToolWidget::emitSetparameterFive(double val)
+{
+    isSettingSet[4] = 1;
```
abdullahtahiriyo, line 194: If these variables are just to keep the state, then it is a bad solution. The widget can get the values directly from the widget, there is no need to cache them and synchronise them.

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
```diff
@@ -0,0 +1,967 @@
...
+    widget->setSettings(0);
+}
+
+TaskSketcherTool::~TaskSketcherTool()
+{
+    Gui::Selection().Detach(this);
+    //Content.erase(Content.begin());
+}
+
+void TaskSketcherTool::onChangedSketchView(const Gui::ViewProvider& vp,
```
abdullahtahiriyo, line 943: I am unsure if you have a planned use for this. If you do not, then we need to clean up the code.

It also happens to the framework I shared. There are headers that simply are not needed in some files.

src/Mod/Sketcher/Gui/TaskSketcherTool.h
```diff
@@ -0,0 +1,113 @@
...
+
+    void setparameter(double val, int i);
+    void setLabel(const QString& atext, int i);
+    void setUnit(Base::Unit unit, int i);
+    void setParameterActive(bool val, int i);
+    void setParameterVisible(bool val, int i);
+    void setParameterFocus(int i);
+    void setSettings(int toolSelected);
+    bool isCheckBoxChecked(int i);
+    bool isWidgetActive;
```
abdullahtahiriyo, line 66: Now you can see if the widget is active from the DrawSketchHandler, just looking if the pointer to the tool widget is nullptr or not.

src/Mod/Sketcher/Gui/TaskSketcherTool.h
```diff
@@ -0,0 +1,113 @@
...
+    std::vector<double> toolParameters;
+    std::vector<bool> isSettingSet;
+    bool unitChangeTriggeredValueChanged;
+
+//Q_SIGNALS:
+public Q_SLOTS:
+    void emitSetparameter(double val);
```
abdullahtahiriyo, line 73: Oh, yes the slots. I do not know exactly how you will want to use these. I have left them in the code, but you may want consider to connect them to other class, or to signal to another class within the slot. 

Code in the default widget should be generic enough to allow any use of parameters.

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
PaddleStroke: This is not to keep the state. This is to keep either the user typed a value at least once or not. See this example : 
`virtual void mouseMove(Base::Vector2d onSketchPos)
    {
        if (Mode == STATUS_SEEK_First) {
            setPositionText(onSketchPos);
            if (seekAutoConstraint(sugConstr1, onSketchPos, Base::Vector2d(0.f, 0.f))) {...}
            if (toolSettings->widget->isSettingSet[0] + toolSettings->widget->isSettingSet[1] == 2) {
                pressButton(onSketchPos);
                releaseButton(onSketchPos);
            }
        }`
You see if the user typed in the first and second parameter, then DrawSketchHandler automatically goes to STATUS_SEEK_Second by calling mouseMove which release a pressButton/releaseButton.

I'm not sure how to implement this with the new architecture.

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
PaddleStroke: I don't know what it does and just left if after copy/paste the other task widget class.

src/Mod/Sketcher/Gui/TaskSketcherTool.h
PaddleStroke: Slots are doing : `isSettingSet[i] = 1;` which let drawSketchHandler if user input a value or not.
How should this be handled now? Should I put back isSettingSet bool vector in the default widget?

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
PaddleStroke:
```
            EditCurve[0] = onSketchPos;
            if (toolSettings->widget->isSettingSet[0] == 1) {
                EditCurve[0].x = toolSettings->widget->toolParameters[0];
            }
            if (toolSettings->widget->isSettingSet[1] == 1) {
                EditCurve[0].y = toolSettings->widget->toolParameters[1];
            }
```
So I'm quite lost with the new architecture. How do the basic operation of the widget : knowing if a parameters has been typed in (isSettingSet) and getting its value (toolParameters).

src/Mod/Sketcher/Gui/TaskSketcherTool.h
PaddleStroke: > Oh, yes the slots. I do not know exactly how you will want to use these. I have left them in the code, but you may want consider to connect them to other class, or to signal to another class within the slot.

They have 4 functions : 1- register that the parameter has been set at least once in the bool vector : isSettingSet.
2- set focus to next spinbox
3- Trigger a mouseMove of the drawsketchHandler. This enable to update the geometry and go to next step (STATUS_SEEK_Second...) automatically if user typed values.
4- When spinbox unit is changed (for angles for instance) valueChanged is triggered, which triggers isSettingSet. So then the DrawSketchHandler thinks we entered a value. Though this might not be an issue when the widget factory is used.

Please advise how 3- should be handled now. Thanks

```
void SketcherToolWidget::emitSetparameter(double val)
{
    isSettingSet[0] = 1;
    toolParameters[0] = val;
    if (unitChangeTriggeredValueChanged) {
        unitChangeTriggeredValueChanged = 0;
    }
    else setParameterFocus(1);

    //Make a mousemove to update geometry when parameter is set.
    if(isWidgetActive) {
        sketchView->mouseMove(sketchView->prvMoveCursorPos, sketchView->prvMoveViewer);
    }
}
```

PaddleStroke: Is it intended that the widget is private to toolWidgetManager?
Then nothing can be used directly from the handler. 
For instance if I want to do 
toolWidget->setParameterFocus(x1);
I need to create a function in toolWidgetManager? What is the point it feels like it just makes things convoluted. Or is it to prepare for widget factory?

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
PaddleStroke: > @PaddleStroke
> 
> Do not worry. Most likely I won't have time to review before the next weekend. We will cross the bridge when we get there.

See here the updated version. It is now working but with several issues: https://github.com/PaddleStroke/FreeCAD/tree/paddle_widget

abdullahtahiriyo:
@PaddleStroke 

I will try to give it a look today or tomorrow

PR #5427, opened 2022-01-24
[Draft] Grid size as quantity rather than double

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=8&t=26277

### DIFF
--- a/src/Mod/Draft/Resources/ui/preferences-draftsnap.ui
+++ b/src/Mod/Draft/Resources/ui/preferences-draftsnap.ui
@@ -445,7 +445,7 @@
          </spacer>
         </item>
         <item>
-         <widget class="Gui::PrefDoubleSpinBox" name="gui::prefdoublespinbox_3">
+         <widget class="Gui::PrefQuantitySpinBox" name="gui::prefdoublespinbox_3">
           <property name="enabled">
            <bool>true</bool>
           </property>
@@ -779,7 +779,7 @@
    <header>Gui/PrefWidgets.h</header>
   </customwidget>
   <customwidget>
-   <class>Gui::PrefDoubleSpinBox</class>
+   <class>Gui::PrefQuantitySpinBox</class>
    <extends>QDoubleSpinBox</extends>
    <header>Gui/PrefWidgets.h</header>
   </customwidget>


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5427">pipeline status</a> for feature branch [PR_5427](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5427). Pipeline [455231825](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/455231825) was triggered at [6157147](https://github.com/FreeCAD/FreeCAD/pull/5427/commits/6157147b907a123e0977039861ab83c1d572e439). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Roy-043:
This needs to be reverted. Changing the gridSpacing from double to quantity is more involved than changing the ui file only. The code that uses this preference has to be adapted as well.

See:
https://forum.freecadweb.org/viewtopic.php?f=23&t=68675

0penBrain:
Why not fixing rather than reverting?

Roy-043:
The fact that this was never tested is not good of course...

Changing a preference from float to string is something that should not be taken lightly.

Some remarks/questions:
1. I do not understand why this was only done for this preference. For consistency many other Draft preferences should also be updated. See the "texts and dimensions" tab f.e.
2. There should be code to transfer the old float value. Users that switch from V0.19 to V0.20 have to enter a new value now, and end up with two gridSpacing entries in the user.cfg file.
3. Since the quantity ends up as a string value there may need to be checks in place to handle faulty strings?
4. Isn't is possible to keep the stored value as a float and only present it to the user with the correct unit?

My suggestion is to revert this for now and readdress this during the V0.21 dev cycle. And then look at ALL Draft preferences.

There is also the issue that Draft preferences have inconsistent default values in the source code. That is something that may then also be tackled.

Roy-043:
There is a problem with the default value:
1. Start FreeCAD without user.cfg.
2. Switch to the Draft WB.
3. Call Std_DlgParameter.
4. Don't change anything, just press OK.
5. The gridSpacing preference is now set to "0 mm". This should be 1mm.

0penBrain:
@donovaly May you have an opinion about this ? May you know how much time we have to work on this before 0.20 is released ? Thx

yorikvanhavre:
iirc a value of zero disables the grid... maybe this should be changed (there are other, better ways to disablethe grid), but then, what value should we use instead?

In case the grid spacing value is not set, a value of 1 is used: https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Draft/draftguitools/gui_trackers.py#L1197

Maybe that should become a general rule?

I think what happens here is that once the pref dialog is open, all values are initialized. So the default of 1 above is not used anymore, but the actual default value of the preference ui file, which i suppose is zero. Maybe there, we should set something like 1mm?

0penBrain:
@yorikvanhavre I'd rather suspect @Roy-043 to be true. The preferences write a string, but Draft should still be using 'GetFloat' thus fails.

yorikvanhavre:
strange, i thought it was saved as a float too (the quantityspinbox widget is an extension of doublespinbox).. in any case, it certainly doesn't fail with setting a value of zero?
I can have a better look tomorrow

0penBrain:
> strange, i thought it was saved as a float too (the quantityspinbox widget is an extension of doublespinbox).

I thought too but it's indeed saved as a string.
@wwmayer Is there a reason/rationale why quantities are saved (in parameters) as string rather than as float in mm ?

yorikvanhavre:
then we could already set somewhere that if the value is zero, either don't try to draw the grid or set the value to 1.. maybe issue a warning to the user too

wwmayer:
> @wwmayer Is there a reason/rationale why quantities are saved (in parameters) as string rather than as float in mm ?

I once copied the code from the class InputField which saved the raw user input. The idea was that a user can e.g. enter `1 in` and immediately save it via the context menu. If you open the context menu again the entry is `1 in` and not the computed value according to the unit settings.

Roy-043:
> then we could already set somewhere that if the value is zero, either don't try to draw the grid or set the value to 1.. maybe issue a warning to the user too

Yes, and we should do that for this preference (and other preferences as well). But the reported bug should also be addressed. The default value in the ui file is ignored in V0.20:
https://github.com/FreeCAD/FreeCAD/blob/a564d5ced7aecf320a0ac2d73f824a3dc7861eb3/src/Mod/Draft/Resources/ui/preferences-draftsnap.ui#L464-L466

yorikvanhavre:
The above commit fixes the error, but yes indeed the issue of the default double value that isn't used should be fixed too

0penBrain:
Hi all, sorry, not so much spare time as I'd like ATM. May one test the PR #6899 to check it fixes issues and doesn't introduce other regressions ? Thx

Roy-043:
There is an issue with Yorik's commit. I'll do a PR for that.

After that PR there is one important issue that remains:
1. Switch Units to US customary.
2. Start Draft_SelectPlane.
3. In the task panel enter 1 (without a unit) for the grid spacing.
4. Press the Close button.
5. Result: grid spacing is 1mm.
6. You can verify this by drawing a line and snapping to 2 grid points.
7. But if you check the grid spacing in the preferences you will now see 1" (1 inch).

Roy-043:
Apart from leaving out the unit in the task panel it seems you can enter any unit you want. F.e. if you have caps lock on and enter 1 M (instead of 1 m), this is interpreted as 0,016667 ° and the grid spacing will then be 0,016667 mm.

0penBrain:
> 1. Switch Units to US customary.
> 2. Start Draft_SelectPlane.
> 3. In the task panel enter 1 (without a unit) for the grid spacing.
> 4. Press the Close button.
> 5. Result: grid spacing is 1mm.

This is inherent to `InputField` type and if should be modified, that shall be done with much care as would have have a wide impact.
This said, you have exactly the same case in eg. Spreadsheet where a unitless value is always considered mm (for length).
To me, I wouldn't change anything regarding this and just mention that "as a general rule, a unitless value is always considered being default base unit (mm for length, deg for angle, ...)". ;)
> Of course several other, already mentioned, details still remain.

If something has to be done, good to have a clear list of these points. Eventually better do it in #6899 
> Apart from leaving out the unit in the task panel it seems you can enter any unit you want. F.e. if you have caps lock on and enter 1 M (instead of 1 m), this is interpreted as 0,016667 ° and the grid spacing will then be 0,016667 mm. Again this was found while testing #6899.

Solved by new commit in #6899 ;)

0penBrain:
> There is a problem with the default value:
> 
> 1. Start FreeCAD without user.cfg.
> 2. Switch to the Draft WB.
> 3. Call Std_DlgParameter.
> 4. Don't change anything, just press OK.
> 5. The gridSpacing preference is now set to "0 mm". This should be "1mm".

Solved by #6991

PR #5440, opened 2022-01-25
Path: Path Array dressup compatibility

voneiden:
This commit makes PathArray able to discover the original base feature ToolController recursively through (nested) dressups. I've hard capped the recursion depth limit to 10 to avoid possible issues with cyclic references (if such are possible?). 

I've tested this with a nested Tag Dressup -> Ramp Dressup -> Profile and it seems to work fine.

I'd appreciate further testing from other users. I've tested this (+ ran unit tests) with Linux weekly AppImage build 27078.

Wiki page will need to be updated once this goes live.

Some discussion on this limitation 
* https://forum.freecadweb.org/viewtopic.php?t=26120 (2018)
* https://forum.freecadweb.org/viewtopic.php?f=15&t=54766 (2021)

Wiki page (section Limitations)
* https://wiki.freecadweb.org/Path_Array

### DIFF
--- a/src/Mod/Path/PathScripts/PathArray.py
+++ b/src/Mod/Path/PathScripts/PathArray.py
@@ -25,6 +25,7 @@ import FreeCADGui
 import Path
 import PathScripts
 from PathScripts import PathLog
+from PathScripts.PathDressup import toolController
 from PySide import QtCore
 import math
 import random
@@ -280,7 +281,7 @@ class ObjectArray:
         if len(base) == 0:
             return
 
-        obj.ToolController = base[0].ToolController
+        obj.ToolController = toolController(base[0])
 
         # Do not generate paths and clear current Path data if operation not
         if not obj.Active:
@@ -381,9 +382,12 @@ class PathArray:
                 return
             if not b.Path:
                 return
-            if not b.ToolController:
+
+            b_tool_controller = toolController(b)
+            if not b_tool_controller:
                 return
-            if b.ToolController != base[0].ToolController:
+
+            if b_tool_controller != toolController(base[0]):
                 # this may be important if Job output is split by tool controller
                 PathLog.warning(
                     translate(
@@ -512,13 +516,8 @@ class CommandPathArray:
         }
 
     def IsActive(self):
-        if bool(FreeCADGui.Selection.getSelection()) is False:
-            return False
-        try:
-            obj = FreeCADGui.Selection.getSelectionEx()[0].Object
-            return isinstance(obj.Proxy, PathScripts.PathOp.ObjectOp)
-        except (IndexError, AttributeError):
-            return False
+        selections = [sel.isDerivedFrom("Path::Feature") for sel in FreeCADGui.Selection.getSelection()]
+        return selections and all(selections)
 
     def Activated(self):
 


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5440">pipeline status</a> for feature branch [PR_5440](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5440). Pipeline [456257552](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/456257552) was triggered at [26e47f2](https://github.com/FreeCAD/FreeCAD/pull/5440/commits/26e47f2bbcf3f0892b09913f73bf70a9221d1d3c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
This is a good idea, but shouldn't it be a generalized fix beyond PathArray?  How about implementing a helper function in PathUtils that returns the tool controller for an operation/dressup?  The helper could use the same recursive search but be called by PathArray or any other dressup.

sliptonic:
In fact, I think this could even be an improvment to PathUtils.findToolController()

That function currently is used to pick  a default TC for a new operation.  But it could have an optional argument of Operation=None.   If Operation isn't none, it would recursively search the op/dressup until it found one.   Just a thought.

voneiden:
Seems reasonable. I wasn't aware of the existing (somewhat?) related code in PathUtils. For the time being just consider this as a proof of concept that patharray's work just fine with dressup's as long as a TC is provided.

I'm unlikely able to extend the scope of this PR any time soon as I'm not at all familiar with the codebase and quite tight on development free time. If somebody wants to implement this in a more general way, absolutely feel free to do so. Just drop a comment here so we don't end up accidentally doing duplicate work.

Russ4262:
@voneiden
Evening.  I just found your PR here.  I just created PR #6549, unaware of your proposal here.  Any changes or updates for this PR?

@sliptonic Is there any progress on this?  I will gladly close my PR seeing that this one is open.

voneiden:
I'm fine with closing this PR and letting @Russ4262 wrap it up if they can spare the time.

Russ4262:
@voneiden Please forgive my poor communcation. I am in no way suggesting you close your current PR here. You had this PR in place well before mine. You and Sliptonic clearly have some dialogue happening related to this PR. I will gladly close mine,  #6549  and support your effort here. I was simply unaware of your PR when I put mine together yesterday. I have been out of touch with FreeCAD for a number of months.

I am excited that more users like yourself are contributing to this project!

sliptonic:
The two PRs look very similar.  @voneiden's has the advantage of recursing.   I still wish this was abstracted out of PathArray and made part of Path Utils but I understand that won't happen.
There's a minor merge conflict.  If you'll fix this and rebase, I'll merge this one.

voneiden:
Alright alright, no promises, but I'll try to find some time this weekend to refactor it into path utils.

Russ4262:
Do we need recursion? I am thinks ng about how the post processing works. Does it not only see the parent operation,  in this case a dressup? Even if there are five dressups stacked one on top of the other, each dressup should have a ToolController and likely a Coolant properties, yes? When a dressup is applied, it should just adopt the TC and Coolant objects/values of the base operation it is modifying, yes?    For this reason, I question the need for recursion.  I am not convinced we do not, or will not ever need recursion, just trying to understand what we need.

voneiden:
I probably had a good justification for myself to use recursion back in January but the real reason may have been just inexperience with the codebase. I looked around a bit more now and became aware of the PathDressup.toolController. I rewrote using that and it would seem to work just fine (and is a much simpler). I also only now, after rewriting, peeked at the actual changes in #6549 @Russ4262 implemented.. and somewhat amusingly apart for variable names and `IsActive`, we're exactly on the same page. 

Any thoughts on `IsActive`? As I'm not familiar with the codebase I have no idea if I've taken a naughty shortcut with my approach.

Russ4262:
Regarding `IsActive`, I do not think my modification is best practice.  I made some attempts to compare classes of objects, but the string conversion method is the best I could come up with.  Your 'all-in' selection method might be better, as it should allow any Path object to be the base for an Array.  For example, I did have not tested either your or my PR with a Custom operation as a base.  In theory, a user might want to do so.  At the moment, your version of `IsActive` will likely allow it, whereas mine is only targeted at dressups.

All that to say, I really don't know.  I think your `IsActive` version looks fine, but my programming experience and knowledge is much less than others involved in this project.  Perhaps Sliptonic or one of the senior devs can way in.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5440">pipeline status</a> for feature branch [PR_5440](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5440). Pipeline [490566557](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/490566557) was triggered at [d2e8921](https://github.com/FreeCAD/FreeCAD/pull/5440/commits/d2e89213305f1ba89feccdc5417ecafb3b217401). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
> Any thoughts on `IsActive`? As I'm not familiar with the codebase I have no idea if I've taken a naughty shortcut with my approach.

all() returns true if the list is empty.  Doesn't that mean IsActive is going to return True if nothing is selected?

voneiden:
Good point @sliptonic . It is indeed normal behaviour with dressups - most of them are active with empty selection. However now that you point it out, I notice that unlike the dressups, PathArray `Activated` function does not check for empty selection and therefore doesn't print an error. So I've opted to inactivate it with empty selection.

I've updated `IsActive` to check for empty list and rebased the branch on top of master.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5440">pipeline status</a> for feature branch [PR_5440](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5440). Pipeline [490728508](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/490728508) was triggered at [af8b6fb](https://github.com/FreeCAD/FreeCAD/pull/5440/commits/af8b6fba07c77630bed0e7455e90924f7e5fed47). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #5495, opened 2022-02-04
Base matrix computation fixes & improvements

0penBrain:


0penBrain:
Set as draft. Rebasing will be tough. :)

0penBrain:
@wwmayer May you help. When I try in Python console `App.Matrix().scale(5)` it returns error saying
> TypeError: function takes exactly 3 arguments (1 given)
The above exception was the direct cause of the following exception:
SystemError: <built-in method scale of Base.Matrix object at 0x4472888> returned a result with an error set

I'm definitely missing something regarding Python binding...

marioalexis84:
if `PyArg_ParseTuple` fails, returns `false` and raise an exception. The error must be cleared in the next block with `PyErr_Clear()`.

0penBrain:
@marioalexis84 Brilliant catch. Thanks a lot.
Releasing PR.

marioalexis84:
`PyErr_Clear` must be called before the following code:
`if (PyArg_ParseTupe(...)) {`     <- fail here
`...}`
`else {`
`PyErr_Clear()`
`...`
`}`

0penBrain:
@marioalexis84 I checked as it is and works fine. ;) And actually is a dup of what is done in other `else if` blocks (that I previously missed).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5495">pipeline status</a> for feature branch [PR_5495](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5495). Pipeline [464174145](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/464174145) was triggered at [20cead1](https://github.com/FreeCAD/FreeCAD/pull/5495/commits/20cead1afb053e7546656623c16334f263b9f9d6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
The structure of MatrixPy::move() and MatrixPy::scale() is actually wrong. I don't know who originally wrote it that way but doing
```
if (PyArg_ParseTuple(args, ...)) {
...
}
else if (PyArg_ParseTuple(args, ...)) {
   PyErr_Clear();
...
}
else if (PyArg_ParseTuple(args, ...)) {
   PyErr_Clear();
...
}
```

doesn't work reliably. Once you called PyArg_ParseTuple() and it returns false you cannot immediately call it again with different arguments. It will work in most cases but in some rare cases it will fail when it should succeed.

The two functions (I haven't checked if there are more) should be restructured to:
```
if (PyArg_ParseTuple(args, ...)) {
   return ...;
}

PyErr_Clear();
if (PyArg_ParseTuple(args, ...)) {
   return ...;
}

PyErr_Clear();
if (PyArg_ParseTuple(args, ...)) {
   return ...;
}
```

0penBrain:
@wwmayer makes sense. The problem with the If..return structure is that it get rid of some interesting code factorization.
So I'd like to propose this implementation using a cascaded if. I know it's a bit controversial as an implementation, let me know.

@marioalexis84 sorry, looks like you were very right. :+1:

src/Base/MatrixPyImp.cpp
```diff
@@ -301,27 +308,39 @@ PyObject* MatrixPy::scale(PyObject * args)
...
+        // clears the error from previous PyArg_ParseTuple()
+        PyErr_Clear();
+        if (PyArg_ParseTuple(args, "O!:three floats or a vector is needed", &(Base::VectorPy::Type), &pcVecObj)) {
+            Base::VectorPy  *pcObject = static_cast<Base::VectorPy*>(pcVecObj);
+            Base::Vector3d* val = pcObject->getVectorPtr();
+            vec.Set(val->x,val->y,val->z);
+            break;
+        }
+        // clears the error from previous PyArg_ParseTuple()
         PyErr_Clear();
```
marioalexis84, line 340: Returning `nullptr` without setting an exception raises a `SystemError` exception.
This is the case if the arguments passed to the function do not meet the requirements in any of the previous `PyArg_ParseTuple` calls.
So, don't call `PyErr_Clear()` here, or instead set a specific error with `PyErr_SetString` before returning `nullptr`:

`PyErr_SetString(PyExc_TypeError, "Expected three floats or tuple with three floats or Base.Vector");`
`return nullptr;`

If you decide to use this approach, remove the error message `"three floats or a vector is needed"` from the format string in `PyArg_ParseTuple`,  which should actually be set using a semicolon, not a colon.

src/Base/MatrixPyImp.cpp
```diff
@@ -265,27 +265,34 @@ PyObject* MatrixPy::move(PyObject * args)
...
-        Base::Vector3d* val = pcObject->getVectorPtr();
-        vec.Set(val->x,val->y,val->z);
-        // clears the error from the first PyArg_ParseTuple()6
+            Base::VectorPy  *pcObject = static_cast<Base::VectorPy*>(pcVecObj);
+            Base::Vector3d* val = pcObject->getVectorPtr();
+            vec.Set(val->x,val->y,val->z);
+            break;
+        }
+        // clears the error from previous PyArg_ParseTuple()
         PyErr_Clear();
```
marioalexis84, line 292: Same as in next comment

src/Base/MatrixPyImp.cpp
0penBrain: Good catch. Thanks for fast review.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5495">pipeline status</a> for feature branch [PR_5495](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5495). Pipeline [464825517](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/464825517) was triggered at [3036743](https://github.com/FreeCAD/FreeCAD/pull/5495/commits/3036743b35e9e1699d52068f171e9db2a9a3ccf3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> So I'd like to propose this implementation using a cascaded if. I know it's a bit controversial as an implementation, let me know.

The do/while loop is fine and that's how I do it sometimes, too. The alternative would be to use nested if-else blocks that will very quickly lead to unreadable code.

PR #5498, opened 2022-02-04
Addon Manager: Add branch switching GUI

chennes:
For users with git and GitPython installed, a new button is added to the details view allowing them to change the branch they are on. Some other minor UI changes are made to accommodate this new behavior.

### DIFF
--- a/src/Mod/AddonManager/AddonManager.py
+++ b/src/Mod/AddonManager/AddonManager.py
@@ -757,6 +757,8 @@ class CommandAddonManager:
         if repo.status() == AddonManagerRepo.UpdateStatus.UPDATE_AVAILABLE:
             self.packages_with_updates.append(repo)
             self.enable_updates(len(self.packages_with_updates))
+        elif repo.status() == AddonManagerRepo.UpdateStatus.PENDING_RESTART:
+            self.restart_required = True
 
     def enable_updates(self, number_of_updates: int) -> None:
         """enables the update button"""
--- a/src/Mod/AddonManager/AddonManagerOptions.ui
+++ b/src/Mod/AddonManager/AddonManagerOptions.ui
@@ -285,6 +285,34 @@ of the line after a space (e.g. https://github.com/FreeCAD/FreeCAD master).</str
     </layout>
    </item>
    <item>
+    <widget class="QGroupBox" name="advanced">
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
+       <horstretch>0</horstretch>
+       <verstretch>0</verstretch>
+      </sizepolicy>
+     </property>
+     <property name="title">
+      <string>Advanced Options</string>
+     </property>
+     <layout class="QVBoxLayout" name="verticalLayout_3">
+      <item>
+       <widget class="Gui::PrefCheckBox" name="guiprefcheckboxShowBranchSwitcher">
+        <property name="text">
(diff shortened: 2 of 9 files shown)


luzpaz:
Can we provide a warning about changing branches and possibly destabilizing models since branches have experimental code ?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5498">pipeline status</a> for feature branch [PR_5498](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5498). Pipeline [464176217](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/464176217) was triggered at [6e26348](https://github.com/FreeCAD/FreeCAD/pull/5498/commits/6e26348373f4558de4e64b57b6c22bf9805ac5fa). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5498">pipeline status</a> for feature branch [PR_5498](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5498). Pipeline [464178221](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/464178221) was triggered at [330abdc](https://github.com/FreeCAD/FreeCAD/pull/5498/commits/330abdc3f5cf96274db36ae5b3d9b7650afd2214). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
Any feedback you can provide here @davesrocketshop ? You mentioned some points on twitter and was hoping you could shed any light as a addon dev perspective?

chennes:
I thought about it, but still haven't decided on a best path forward. A few possibilities are:
1) Have a default-false preference checkbox to show this button, so only advanced users will even see it
2) Add a warning dialog when switching branches the first time
3) Add a warning dialog when switching branches every time
4) Allow devs to specify which branches to show (maybe in a package.xml file, though that can be ugly because it is itself in git)

Probably other options out there, but I wanted to start simple.

davesrocketshop:
> I thought about it, but still haven't decided on a best path forward. A few possibilities are:
> 
>     1. Have a default-false preference checkbox to show this button, so only advanced users will even see it

I like this.

> 
>     2. Add a warning dialog when switching branches the first time
> 
>     3. Add a warning dialog when switching branches every time

I think it needs to be shown every time. Switching probably won't be frequent and the risks are high.

> 
>     4. Allow devs to specify which branches to show (maybe in a package.xml file, though that can be ugly because it is itself in git)

I don't think this should be much of a problem, and a good idea. It should be possible to specify a data model version and perhaps a list of compatible model versions - both forward and backwards. It could also flag whether the data model is unstable (not finalized).

> 
> 
> Probably other options out there, but I wanted to start simple.

Glad you're thinking of this. I think this will be a great feature.

chennes:
@luzpaz and @davesrocketshop I've pushed a change to this PR that marks this as an "Advanced" feature that requires deliberate enabling in the preferences, and that presents a warning dialog when switching branches. Please let me know if these changes address your concerns.

PR #5505, opened 2022-02-05
Path: Add Gcode for Spindle Stop (M5) before a Tool Change (M6)

jimzim111:
Updates the Path Tool Controller to insert gcode for a M5 spindle stop before M6 tool change.  This helps to preserve fingers when performing manual tool changes.

### DIFF
--- a/src/Mod/Path/Generators/toolchange_generator.py
+++ b/src/Mod/Path/Generators/toolchange_generator.py
@@ -66,6 +66,7 @@ def generate(
     commands = []
 
     commands.append(Path.Command(f"({toollabel})"))
+    commands.append(Path.Command("M5"))
     commands.append(Path.Command("M6", {"T": int(toolnumber)}))
 
     if spindledirection is SpindleDirection.OFF:


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5505">pipeline status</a> for feature branch [PR_5505](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5505). Pipeline [464340230](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/464340230) was triggered at [14220d2](https://github.com/FreeCAD/FreeCAD/pull/5505/commits/14220d20dc9321bb1b41ae221847f1b3c05b9ede). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
This change looks good but you need to revise the unit tests as well.
TestPathToolChangeGenerator

This should do it...

`
class TestPathToolChangeGenerator(PathTestUtils.PathTestBase):
    def test00(self):
        """Test Basic Tool Change Generator Return"""

        args = {
            "toolnumber": 1,
            "toollabel": "My Label",
            "spindlespeed": 500,
            "spindledirection": SpindleDirection.OFF,
        }

        results = generator.generate(**args)

        # Get a label
        self.assertTrue(len(results) == 3)
        commentcommand = results[0]
        self.assertTrue(isinstance(commentcommand, Path.Command))
        self.assertTrue(commentcommand.toGCode() == "(My Label)")

        # Stop the spindle
        toolcommand = results[1]
        self.assertTrue(toolcommand.Name == "M5")

        # Get a tool command
        toolcommand = results[2]
        self.assertTrue(toolcommand.Name == "M6")

        # Turn on the spindle
        args["spindledirection"] = SpindleDirection.CW
        results = generator.generate(**args)
        self.assertTrue(len(results) == 4)

        speedcommand = results[3]
        self.assertTrue(speedcommand.Name == "M3")
        self.assertTrue(speedcommand.Parameters["S"] == 500)

        # speed zero with spindle on
        args["spindlespeed"] = 0
        results = generator.generate(**args)
        self.assertTrue(len(results) == 3)
        PathLog.track(results)

        # negative spindlespeed
        args["spindlespeed"] = -10
        self.assertRaises(ValueError, generator.generate, **args)
`

sliptonic:
Hmm.  Now that I think about it, this needs more work.
Consider the scenario where the user is running the same tool with two different feeds/speeds.  One for roughing and another for finishing.  (I've actually seen this in woodworking on a router)

The spindle is started CW at 2000RPM. 
One or more operations is processed with a feedrate of 500IPM.
The spindle is then sped up to 4500RPM
several more operations are processed with feedrate of 800IPM.

With the current solution, the spindle will be stopped and then restarted without a toolchange.  That's not the worst thing in the world but if the controller senses current spindle speed, it will wait until the spindle comes to a complete stop, then start again and wait until it reaches target speed.

It seems this generator needs more context to do the right thing.

sliptonic:
Thinking about this more.  I don't think we should do the M5 here at all.
Even though the Path Commands are very nearly gcode, they are not, in fact, gcode.  The Path Commands are an intermediate structure that will become gcode at postprocessing. 

At the time that the tool change commands are added there are two things that are required:
We need to know WHICH tool we're changing to (tool number)
We need to know what to do with the the spindle (On/off,  forward/reverse,  how fast)
Both of these come from the Tool Controller.  

At post-processing, those things get translated into gcode that is specific to the machine. The specific gcode commands will likely invole many other things like: stopping the spindle,  transiting to a change location,  retracting to the top, etc etc.  

For this reason,  I think the M5 command to stop the spindle is a machine-specific thing that should be added by the postprocessor.  For example,  my machine doesn't need an m5.  If a tool change is called, it automatically stops the spindle.

On the other hand, if we do an M5 here, it means the generator of the commands has to know the context.  That means keeping some kind of state object around and passing it from operation to operation and tool change to tool change.  That will get ugly pretty fast.

Thoughts?

jimzim111:
This seems reasonable.  The design/planning portion of Path WB doesn't
know/care much about machine capabilities.  The post step is what makes
actual motion commands for the machine so it makes sense to deal with the
spindle control there.

I'll add it as an option in post.  Do you think it should be the default to
stop spindle for TC, or just make it available as an option?  (Tradeoff
between how many users/machines will benefit from the new behavior vs. the
change to the default user experience)

Also, would it be better to close this PR and make a new one when ready?
Or just leave it while I'm working on it?

sliptonic:
@jimzim111, if you're willing to work on it, I would suggest a 'slightly' larger scope.

We should create a designated 'default' post processor.  It should be very clean and well documented. It shouldn't have any machine-unique features.  It should default to being explicit (line numbers, comments).   And it should be safe (When in doubt, add the M5 or other possibly redundant commands.)  This should be the post that the user gets if he doesn't select anything specifically.

Eventually, I'd like to refactor the post processors so there's less code duplication and its simpler to add machine-specific features but I'm not talking about that now.  I'm just talking about a new post that can serve two purposes:

1) Be the one the user gets if he doesn't select one on purpose (I think right now he gets Centroid)
2) Be the starting point for new posts that are created by the user.  I think most users start with linuxcnc but that makes us reluctant to include linuxcnc specific features that don't apply to other machines. 

If you're willing to work on that, then close this PR and start a forum topic.  As soon as you have something, create a new PR in draft mode.  I started looking at this a while back but haven't touched it in a while.  You might find something useful in this branch.  https://github.com/sliptonic/freecad/tree/feature/defaultpost

PR #5517, opened 2022-02-07
[Sketcher] Fix segfault inserting knot in old b-spline

AjinkyaDahale:
See https://forum.freecadweb.org/viewtopic.php?f=19&t=64962&p=569081#p568723.

~This PR will stay as a draft until the issue is created and discussed.~ See #6324

### DIFF
--- a/src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
+++ b/src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
@@ -126,7 +126,7 @@ void EditModeConstraintCoinManager::updateVirtualSpace()
 
     bool isshownvirtualspace = ViewProviderSketchCoinAttorney::isShownVirtualSpace(viewProvider);
 
-    if(constrlist.size() == vConstrType.size()) {
+    if (constrlist.size() == vConstrType.size()) {
 
         editModeScenegraphNodes.constrGroup->enable.setNum(constrlist.size());
 
@@ -172,7 +172,7 @@ Restart:
 
         Base::Vector3d normal;
         try {
-            if(!(curve && curve->normalAt(pointoncurve, normal))) {
+            if (!(curve && curve->normalAt(pointoncurve, normal))) {
                 normal = Base::Vector3d(1,0,0);
             }
         }
@@ -199,7 +199,7 @@ Restart:
             SoSeparator *sep = static_cast<SoSeparator *>(editModeScenegraphNodes.constrGroup->getChild(i));
             const Constraint *Constr = *it;
 
-            if(Constr->First < -extGeoCount || Constr->First >= intGeoCount
+            if (Constr->First < -extGeoCount || Constr->First >= intGeoCount
                     || (Constr->Second!=GeoEnum::GeoUndef
                         && (Constr->Second < -extGeoCount || Constr->Second >= intGeoCount))
                     || (Constr->Third!=GeoEnum::GeoUndef
@@ -313,9 +313,9 @@ Restart:
                                 } else
                                     break;
 
-                                if( geo->getTypeId() == Part::GeomEllipse::getClassTypeId() ||
+                                if (geo->getTypeId() == Part::GeomEllipse::getClassTypeId() ||
                                     geo->getTypeId() == Part::GeomArcOfEllipse::getClassTypeId() ||
-                                    geo->getTypeId() == Part::GeomArcOfHyperbola::getClassTypeId() ){
+                                    geo->getTypeId() == Part::GeomArcOfHyperbola::getClassTypeId()){
 
                                     Base::Vector3d majDir, minDir, rvec;
                                     majDir = Base::Vector3d(cos(angle),sin(angle),0);//direction of major axis of ellipse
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5517">pipeline status</a> for feature branch [PR_5517](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5517). Pipeline [465403039](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/465403039) was triggered at [83a5e53](https://github.com/FreeCAD/FreeCAD/pull/5517/commits/83a5e536ad568fbe3109373a8b3c08637155ff56). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
```diff
@@ -126,7 +126,7 @@ void EditModeConstraintCoinManager::updateVirtualSpace()
 
     bool isshownvirtualspace = ViewProviderSketchCoinAttorney::isShownVirtualSpace(viewProvider);
 
-    if(constrlist.size() == vConstrType.size()) {
+    if (constrlist.size() == vConstrType.size()) {
```
abdullahtahiriyo, line 129: I am not very happy about changing the format of the whole file by introducing an extra space after each if.

I do not enforce code styling in PRs in contributed code (unless there are severe readability problems). But it is different to just change the styling of a whole file.

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
```diff
@@ -1919,8 +1919,10 @@ void EditModeConstraintCoinManager::drawConstraintIcons(const GeoListFacade & ge
 
         if (dynamic_cast<SoZoomTranslation *>(translationPtr))
             absPos = static_cast<SoZoomTranslation *>(translationPtr)->abPos.getValue();
-        else
+        else if (translationPtr)
```
abdullahtahiriyo, line 1922: It would be great to track who is calling drawConstraintIcons when this pointer is null.

The reason why this pointer is null should be related to the type of the constraints having changed. If this is the case, a better fix is to just stop the execution there, if the type of the constraints is different (same fix as last time).

If for any reason, that is not possible and no other solution is obvious, I would move this if block four lines up so that it also catches a null pointer when using the SoZoomTranslation (symmetry constraints).

In any case, thank you very much for tracking this one down. :)

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
AjinkyaDahale: > I am not very happy about changing the format of the whole file by introducing an extra space after each if.
> 

This is kind of a "damned if I do, damned if I don't" situation :smile: . There was another PR where a reviewer (I believe @donovaly) made me add all the spaces before they merged. It was part design though.

> I do not enforce code styling in PRs in contributed code (unless there are severe readability problems). But it is different to just change the styling of a whole file.

IIRC there were parts of the file where `if (...` was used and parts where `if(...` was, so the commit doesn't change the styling of the whole file as much as make the styling consistent. In any case I intentionally made a separate commit for it, and can drop the commit if you wish.

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
abdullahtahiriyo: For the second you are absolutely right about dynamic_cast.

For the first, the thing is that if the constraint type does not match, we should break out the loop and stop processing constraint drawing. If that solves the issue, is a much better solution than checking all the translation pointers over the whole function.

So maybe I should have written: Could you check if checking the type of the constraint (as in the other commit) fixes the issue? If yes, could you please bring this solution in? If no, just let me know.

Sorry for the back and forth and thanks for taking care of it.

AjinkyaDahale: > So maybe I should have written: Could you check if checking the type of the constraint (as in the other commit) fixes the issue? If yes, could you please bring this solution in? If no, just let me know.

Got it. It does seem to fix the issue so used that instead and rebased.

> Sorry for the back and forth and thanks for taking care of it.

No problem. Thanks for reviewing.

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
AjinkyaDahale: I was referring to https://github.com/FreeCAD/FreeCAD/pull/5155#discussion_r745132975.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5517">pipeline status</a> for feature branch [PR_5517](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_5517). Pipeline [470441381](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/470441381) was triggered at [81f455b](https://github.com/FreeCAD/FreeCAD/pull/5517/commits/81f455b8cc35a6258798341c98d012557a8a2178). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #6213, opened 2022-02-07
Part: Move Part_SectionCut command to Part menu item list

marioalexis84:
Since the command alters the document, `AlterDoc`is masked to 1.

---

### DIFF
--- a/src/Gui/Workbench.cpp
+++ b/src/Gui/Workbench.cpp
@@ -651,7 +651,7 @@ MenuItem* StdWorkbench::setupMenuBar() const
     *view << "Std_ViewCreate" << "Std_OrthographicCamera" << "Std_PerspectiveCamera" << "Std_MainFullscreen" << "Separator"
           << stdviews << "Std_FreezeViews" << "Std_DrawStyle" << "Std_SelBoundingBox"
           << "Separator" << view3d << zoom
-          << "Std_ViewDockUndockFullscreen" << "Std_AxisCross" << "Std_ToggleClipPlane" << "Part_SectionCut"
+          << "Std_ViewDockUndockFullscreen" << "Std_AxisCross" << "Std_ToggleClipPlane"
           << "Std_TextureMapping"
 #ifdef BUILD_VR
           << "Std_ViewVR"
--- a/src/Mod/Part/Gui/Command.cpp
+++ b/src/Mod/Part/Gui/Command.cpp
@@ -2484,7 +2484,7 @@ CmdPartSectionCut::CmdPartSectionCut()
     sToolTipText = QT_TR_NOOP("Creates a persistent section cut of visible part objects");
     sWhatsThis = "Part_SectionCut";
     sStatusTip = sToolTipText;
-    eType = Alter3DView;
+    eType = AlterDoc | Alter3DView;
 }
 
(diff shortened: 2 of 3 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6213">pipeline status</a> for feature branch [PR_6213](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6213). Pipeline [465494754](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/465494754) was triggered at [5cd76d8](https://github.com/FreeCAD/FreeCAD/pull/6213/commits/5cd76d8823e59d5d110f90a4aeac0c2f58457b57). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I cannot compile right now but it seem that this PR is not doing the right thing:

The main idea of the section cut feature is that it is always in the View menu, no matter in what WB you are. This is because the main usage would be assemblies, and when assembling, you switch WBs all the time and of course when e.g. being in Assembly4, you must be able to start the section cut feature from the view menu, otherwise it won't work as it is designed.

Therefore please bring it back the View menu for all WBs,

donovaly:
> The menu bar is recreated every time a workbench is activated, so a command added to the StdWorkbench menu bar (similar to Part_ColorPerFace, for example) will be lost with every workbench change.

Yes, and this must in this case NOT happen, otherwise the concept of the Section cut  will fail. It will only work in a useful way when you can start it from the View menu, no matter in what WB you are in.

> If the command needs to be accessed from workbenches other than Part, the easiest way is to add the command to these specific cases. For example, for Python-based workbenches: `import PartGui` --> add `"Part_SectionCut"` to workbench command list.

Since I am away and cannot do coding, could you please make such a PR that bring the Section cut feature back to all WBs? Thanks in advance.

marioalexis84:
I understand the idea of making the command visible to other workbenches, but as I say in the other comment, other modules should import PartGui and add Part_SectionCut to their list of commands.

wwmayer:
> but it seem that this PR is not doing the right thing:

It does the right thing but I already assumed you are not happy with the result. :)

> Therefore please bring it back the View menu for all WBs,

In order to do so it requires some little redesign of the core Workbench class. The way you did it was definitely wrong because it violated the design principle that the core system shouldn't know anything about implementation details of extension modules.

People who has set a start up workbench that doesn't automatically load PartGui always see an annoying error message that Part_SectionCut couldn't be found.

As a workaround for the time being you can add a "Global" toolbar and add there the command. This toolbar will be then visible for all workbenches.

donovaly:
> As a workaround for the time being you can add a "Global" toolbar and add there the command. This toolbar will be then visible for all workbenches.

How is this done?
However, this is not what I want. I don't want to bother every user with an icon that he might not want to use ever. I want it therefore to have it in the View menu for all WBs.

donovaly:
Until there is a way to bring it to the View menu for all WBs, I at least brought it back for PD and Part:
https://github.com/FreeCAD/FreeCAD/commit/68a5c7e438d

PR #4118, opened 2020-12-08
[Part] feature to make persistent section cut views

donovaly:
* one aim is to have a feature to make real cuts, not hollow ones like with clipping
* the second one is to have a persistent view, so an object in the tree that can be further exported, meshed etc.

For a detailed description with examples see https://forum.freecadweb.org/viewtopic.php?f=17&t=52441

PR #6322, opened 2022-02-07
Partially revert #4118 and implement intended fix

luzpaz:
`Part_SectionCut` was erroneously added to `src/Gui/Workbench.cpp` which led to a warning `Unknown command 'Part_SectionCut'` when starting up FreeCAD. This commit removes it and adds it to the appropriate place in `src/Mod/Part/Gui/Workbench.cpp`

Thread: https://forum.freecadweb.org/viewtopic.php?p=569276#p569276

### DIFF
--- a/src/Gui/Workbench.cpp
+++ b/src/Gui/Workbench.cpp
@@ -651,7 +651,7 @@ MenuItem* StdWorkbench::setupMenuBar() const
     *view << "Std_ViewCreate" << "Std_OrthographicCamera" << "Std_PerspectiveCamera" << "Std_MainFullscreen" << "Separator"
           << stdviews << "Std_FreezeViews" << "Std_DrawStyle" << "Std_SelBoundingBox"
           << "Separator" << view3d << zoom
-          << "Std_ViewDockUndockFullscreen" << "Std_AxisCross" << "Std_ToggleClipPlane" << "Part_SectionCut"
+          << "Std_ViewDockUndockFullscreen" << "Std_AxisCross" << "Std_ToggleClipPlane"
           << "Std_TextureMapping"
 #ifdef BUILD_VR
           << "Std_ViewVR"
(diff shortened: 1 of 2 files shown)


donovaly:
Thanks, however, this is not right. The cutting feature must be in the View menu and visible no matter what workbench you are in. Otherwise it is not that useful - the main application would be assemblies. So In fact, I was right to add it to the general Gui Workbench.cpp. It seems it needs more thoughts.

marioalexis84:
Same approach on this pull request: https://github.com/FreeCAD/FreeCAD/pull/6213

The menu bar is recreated every time a workbench is activated, so a command added to the StdWorkbench menu bar (similar to Part_ColorPerFace, for example) will be lost with every workbench change.
For Python-based workbenches, things are slightly different: the StdWorkbench menu is created the first time the module is activated, and future activations use a copy of the StdWorkbench menu bar created the first time.
So, in any case, the command is removed from the View menu bar with each activation.
If the command needs to be accessed from workbenches other than Part, the easiest way is to add the command to these specific cases.
For example, for Python-based workbenches:
`import PartGui`
--> add `"Part_SectionCut"` to workbench command list.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6322">pipeline status</a> for feature branch [PR_6322](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6322). Pipeline [465732752](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/465732752) was triggered at [d2697ad](https://github.com/FreeCAD/FreeCAD/pull/6322/commits/d2697ada3e8b469306b053634e788da0fa51f4bb). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
Closing in favor of #6213

Issue #6323, opened 2022-02-07
[Sketcher] bugs in feature "Inserting knots in B-Spline"


donovaly:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Forums discussion

#5311

### Version

0.20 (Development)

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.27422 (Git)
Build type: Release
Branch: master
Hash: d938733eaf2c2ce7cb18d1cbb56147185c473530
Python version: 3.8.6+
Qt version: 5.15.2
Coin version: 4.0.1
OCC version: 7.5.3
Locale: German/Germany (de_DE)
```

### Subproject(s) affected?

Sketcher

### Issue description

There are bugs in this feature (PR #5311):
- it is still not yet documented in our Wiki ;-), please do so
- the dialog always shows "0.5000000", so it ignores the system-wide decimal place settings
- the dialog's spin edit step is too large. It is set to 1.0. This is too large because using 1.5 or -0.5 leads for most cases to errors.
- the tooltip says that the feature will insert a knot with the given parameter and multiplicity, but it is not clear what the parameter is about. It is also not possible to specify the multiplicity.

### Anything else?

@AjinkyaDahale 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

luzpaz:
PR for fix at #5517

AjinkyaDahale:
@luzpaz #5517 does not fix the issues mentioned here

AjinkyaDahale:
@donovaly thanks for the comments. I already inserted a line in the changelog for 0.20 about this feature but will wait till more features are added (like clicking on the curve to pick a point if possible). For the rest of the issues you mentioned I'll try to fic them over the coming days, but you are welcome to suggest fixes if you have the time.

AjinkyaDahale:
@donovaly regarding point 2 of your issue:
1. Where can I find the decimal information?
2. Apart from the number of digits after the separator, are you also referring to the decimal separator itself? Because I'm not sure how I can change the latter.

AjinkyaDahale:
See #6326 

>     * it is still not yet documented in our Wiki ;-), please do so

There is a statement in [0.20 release notes](https://wiki.freecadweb.org/Release_notes_0.20#Further_Sketcher_improvements).
The wiki page itself is on the TODO list. Some features I'd like to add before writing it up:
[...]

PR #5311, opened 2022-01-01
[Sketcher] [Part] Support for inserting knots in B-Spline

AjinkyaDahale:
PR for adding knots in existing b-splines. The insertion itself uses `Geom_BSplineCurve::InsertKnot()` in the back-end.

Note: Please do not squash. The commits have changes in multiple workbenches.

PR #6326, opened 2022-02-08
[Sketcher] Fix some points mentioned in #6323

AjinkyaDahale:
Sets the spinbox step for knot insertion dialog to 0.001.

Removes the mention of multiplicity option from the tooltip.

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
+++ b/src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
@@ -32,6 +32,7 @@
 #endif
 
 #include <Base/Console.h>
+#include <Base/UnitsApi.h>
 #include <App/Application.h>
 #include <Gui/Application.h>
 #include <Gui/Document.h>
@@ -979,7 +980,7 @@ CmdSketcherInsertKnot::CmdSketcherInsertKnot()
     sAppModule      = "Sketcher";
     sGroup          = "Sketcher";
     sMenuText       = QT_TR_NOOP("Insert knot");
-    sToolTipText    = QT_TR_NOOP("Inserts knot at given parameter with the given multiplicity. If a knot already exists at that parameter, it's multiplicity is increased by the value.");
+    sToolTipText    = QT_TR_NOOP("Inserts knot at given parameter. If a knot already exists at that parameter, it's multiplicity is increased by one.");
     sWhatsThis      = "Sketcher_BSplineInsertKnot";
     sStatusTip      = sToolTipText;
     sPixmap         = "Sketcher_BSplineInsertKnot";
@@ -1039,7 +1040,7 @@ void CmdSketcherInsertKnot::activated(int iMsg)
         const auto& mults = bsp->getMultiplicities();
         QInputDialog knotDialog(Gui::getMainWindow());
         knotDialog.setInputMode(QInputDialog::DoubleInput);
-        knotDialog.setDoubleDecimals(8);
+        knotDialog.setDoubleDecimals(Base::UnitsApi::getDecimals());
         knotDialog.setWindowTitle(QObject::tr("Knot parameter"));
         // use values from `knots` and `weights` and insert in label
         std::stringstream knotList;
@@ -1055,6 +1056,8 @@ void CmdSketcherInsertKnot::activated(int iMsg)
                                 .arg(QString::fromUtf8(multList.str().c_str())));
         // use an appropriate middle value from `knots`
         knotDialog.setDoubleValue(0.5 * (knots.front() + knots.back()));
+        // set step size dependent on the knot range
+        knotDialog.setDoubleStep(0.001 * (knots.back() - knots.front()));
         // use min/max from `knots`
         knotDialog.setDoubleRange(knots.front(), knots.back());
 


src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
```diff
@@ -1040,6 +1040,7 @@ void CmdSketcherInsertKnot::activated(int iMsg)
         QInputDialog knotDialog(Gui::getMainWindow());
         knotDialog.setInputMode(QInputDialog::DoubleInput);
         knotDialog.setDoubleDecimals(8);
```
donovaly, line 1042: This is a mistake. It must be set to the user-specified number of decimals.

Search in The Part/Gui code for "XDirectionEdit" so see how it is done there.

donovaly (changes requested):

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
AjinkyaDahale: Thanks for the pointer. Done.

donovaly: > Done.

I don't see a new commit from you.

AjinkyaDahale: I meant in my local machine :). This was a minor edit so i didn't expect it to take this long to compile and test.

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
```diff
@@ -1055,6 +1055,8 @@ void CmdSketcherInsertKnot::activated(int iMsg)
                                 .arg(QString::fromUtf8(multList.str().c_str())));
         // use an appropriate middle value from `knots`
         knotDialog.setDoubleValue(0.5 * (knots.front() + knots.back()));
+        // set step size dependent on the knot range
+        knotDialog.setDoubleStep(0.001 * (knots.back() - knots.front()));
```
donovaly, line 1059: Is it really useful to have 1000 steps? I think 100 steps for the spin box is more practical. (you can enter very value directly.)

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
```diff
@@ -1055,6 +1055,8 @@ void CmdSketcherInsertKnot::activated(int iMsg)
...
         // use min/max from `knots`
         knotDialog.setDoubleRange(knots.front(), knots.back());
```
donovaly, line 1061: It seems you must explicitly set the Min and Max value. Because as it is, I can enter e.g. "3000". The spin edit should only allow to enter geometrically valid values.

donovaly (changes requested):
I found another issue that one can enter senseless values. Can you please have a look?

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
AjinkyaDahale: I believed that was handled by `setDoubleRange`. I'll change it if needed.

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
AjinkyaDahale: I don't have a preference either way but I'll wait for other people for input on this. @abdullahtahiriyo, opinion?

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
donovaly: A rebase to latest master helps me in these cases. Then when I switch between the rebased branches, I really only need to recompile the files changed in the PR.

In case you use MSVC, I recently found this new CMake flag:
FREECAD_USE_MP_COMPILE_FLAG
(listed here: https://wiki.freecadweb.org/Compile_on_Windows#Options_for_the_build_process
 @chennes , many thanks for adding this info, this sped up my workflow a lot)

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
AjinkyaDahale: I can confirm you cannot enter values outside the range. If you press a key that would make the number invalid or have more decimal points, the value doesn't change. Same happens if you paste an invalid value.

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
AjinkyaDahale: I don't use MSVC but thanks for letting me know. I do use `make -j$(nproc --ignore=2)` on Linux.

AjinkyaDahale: I guess I must have rebased to latest master recently but the already compiled version was not rebased so that caused everything to be compiled? I'm not really sure why this happens.

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
donovaly: Does not work for me, here is a screencast:

I can enter any value there into the dialog but I should only be able to enter sensible values

AjinkyaDahale: As such I am not particularly happy about the restricted decimal places but this appears to be a Qt issue: either we can't enter a value with more decimal places or all the unwanted decimal places are shown at all times.

donovaly: Hmm, GitHub seems to have problems with images. However, just pen the dialog and enter there some random value
-> result you can do this. If it would be correct and e.g. the range is between 0 and 1 you would not be able to enter a value like "4362"

AjinkyaDahale: > Does not work for me, here is a screencast:

I can't see the screencast. But if that's the case it might be due to difference in Qt versions. Can you try replacing `knotDialog.setDoubleRange(knots.front(), knots.back());` with `setDoubleMaximum/Minimum` in your branch?

donovaly: > As such I am not particularly happy about the restricted decimal places but this appears to be a Qt issue: either we can't enter a value with more decimal places or all the unwanted decimal places are shown at all times.

I don't understand. When the user decided to have 2 decimals, it is his decision. When he realizes, he needs more, he can change this. For example I use 3 digits.

donovaly: > it might be due to difference in Qt versions. Can you try replacing `knotDialog.setDoubleRange(knots.front(), knots.back());` with `setDoubleMaximum/Minimum` in your branch?

I cannot compile until next week since I am traveling. Here is my FC version:
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.27422 (Git)
Build type: Release
Branch: master
Hash: d938733eaf2c2ce7cb18d1cbb56147185c473530
Python version: 3.8.6+
Qt version: 5.15.2
Coin version: 4.0.1
OCC version: 7.5.3

AjinkyaDahale: > > As such I am not particularly happy about the restricted decimal places but this appears to be a Qt issue: either we can't enter a value with more decimal places or all the unwanted decimal places are shown at all times.
> 
> I don't understand. When the user decided to have 2 decimals, it is his decision. When he realizes, he needs more, he can change this. For example I use 3 digits.
The reported values and values used for calculations should be different, specifically the latter should store more digits. This can add unwanted rounding errors. For example say for whatever reason we have 
1. we have precision set to 1 digit (for exaggeration but we can conjure this issue with higher precision),
2. `v1 = 0.5`, `v2 = 0.6`, `vmid = (v1 + v2)/2`,
2. `vmid` is also stored in a user visible field.
Now, say we wanted `vmid**2`. Then, if `vmid` is stored precisely, the value (which is exactly `0.3025`) should be shown as `0.3`. But if `vmid` is stored incorrectly we are stuck with `vmid=0.6` and `vmid**2=0.36~0.4`.

In our case with a precision of 2 decimals a parameter change of 0.01 can make a lot of difference in the position of the point itself.

Additionally, we might want different precision for different types of values, like wanting length accurate to 0.01 mm but weight to 0.1 g.

I think I'm already dragging this discussion to a totally different tangent so if this is still relevant in the future I'll make a forum post/issue/PR.

AjinkyaDahale: > > it might be due to difference in Qt versions. Can you try replacing `knotDialog.setDoubleRange(knots.front(), knots.back());` with `setDoubleMaximum/Minimum` in your branch?
> 
> I cannot compile until next week since I am traveling.

So I presume you are using the daily builds for testing?

I cannot replicate your issue with the following:
OS: Manjaro Linux (GNOME/gnome)
Word size of FreeCAD: 64-bit
Version: 0.20.27456 +1 (Git)
Build type: Release
Branch: sketcher-fix-6323
Hash: d23e445699cc94096116e4e7ce2512884d5fa098
Python version: 3.10.2
Qt version: 5.15.2
Coin version: 4.0.1
OCC version: 7.5.3

The differences then (in your version) are:
OS: Windows 10 Version 2009
Version: 0.20.27422 (Git)
Branch: master
Hash: d938733eaf2c2ce7cb18d1cbb56147185c473530
Python version: 3.8.6+

I don't have a windows build ready, bet I'll update once I've compiled with d938733eaf2c2ce7cb18d1cbb56147185c473530. I think the quickest thing you can try on your end is changing locale (if en_US is available)?

AjinkyaDahale: Update: I can confirm arbitrary values can be inserted with d938733eaf2c2ce7cb18d1cbb56147185c473530. When I look closely I notice this is before #5517 was merged, so the permitted values are still [`-DBL_MAX`, 'DBL_MAX`]. The range should be fixed by 0.20.27456, but reopen this conversation and let me know if it isn't.

donovaly:
Thanks for the updated PR. I think the step size of the spin buttons can be changed anytime later.
Also thanks for adding the new feature to the Wiki.

Apropos Wiki, here are the main limitations I have with splines:
https://wiki.freecadweb.org/B-Splines#Limitations
If you are in the mood, being able to insert and remove control points would be very valuable.

Merging it now.

AjinkyaDahale:
> Thanks for the updated PR. I think the step size of the spin buttons can be changed anytime later. Also thanks for adding the new feature to the Wiki.

No problem. The wiki page is effectively a stub, though, and I'm too ignorant to properly incorporate it properly into the rest of the wiki. If someone has the time to handle that that'd be great.

> Apropos Wiki, here are the main limitations I have with splines: https://wiki.freecadweb.org/B-Splines#Limitations If you are in the mood, being able to insert and remove control points would be very valuable.

I am more than willing to improve (and in the process, learn about) B-Splines, especially inserting control points. But there are some limitations:
1. I'm basically only exposing routines already available in OCC (from [`Geom_BSplineCurve`](https://dev.opencascade.org/doc/refman/html/class_geom___b_spline_curve.html) which doesn't really have an insert pole function.
2. Even if there were such a function (or if we were to make one), this will result in the shape of the B-Spline changing unpredictably. 

However, addition/removal of control points correspondingly needs an addition/removal of knots in the knot vector, so as such both of these are equivalent. Addition of knots (including increasing multiplicity) can be done safely without altering the shape of the spline, but this also changes the control points that controlled the affected part of the curve. There is a link to some resources for how it's done in #5311 or related pages. By "[inserting] and [removing] control points", what you really want is to be able to change only some parts of the B-Spline, but this is an inherent limitation of B-Splines that we have little control on the actual points _on_ the curve. There are some other curves that allow a bit more control on that (don't quote me but possibly Chebyshev polynomials), but no doubt with their own drawbacks.

PR #6333, opened 2022-02-10
Gui: add option to set application font size

realthunder:
Forum thread https://forum.freecadweb.org/viewtopic.php?p=569808#p569808


src/Gui/ViewParams.cpp
```diff
@@ -59,3 +65,32 @@ ViewParams *ViewParams::instance() {
...
 }
 
+void ViewParams::init() {
+    instance()->onDefaultFontSizeChanged();
+}
+
+int ViewParams::appDefaultFontSize() {
+    static int defaultSize;
+    if (!defaultSize) {
+        QFont font;
```
0penBrain, line 75: Looks like it's not initialized.

realthunder: defaultSize? It's static variable, which is auto init to zero.

0penBrain: No, `font` variable ;)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6333">pipeline status</a> for feature branch [PR_6333](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6333). Pipeline [467796961](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/467796961) was triggered at [39f3a0e](https://github.com/FreeCAD/FreeCAD/pull/6333/commits/39f3a0eb9c6a384b1a52764000af5c8b26d4bf43). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/ViewParams.cpp
0penBrain: @chennes Know that. Probably a matter of taste, but I prefer that it is explicit that we call empty constructor in order to get application default. ;)
Anyway nothing important there.

yorikvanhavre (approved):
This looks perfect to me and highly welcome. Could one of you guys more proficient with C++ than me approve too? Then I'll merge.

chennes:
I am basically categorically opposed to adding any new macro code to our codebase: in C++17 there's basically no reason to do so. I defer to Stroustrup and Co. on this, the reasoning is powerful, and he has given many talks on it. However, I acknowledge that our codebase already has many complex macros, so rather than reject it outright, I leave it to others to evaluate.

wwmayer:
> I am basically categorically opposed to adding any new macro code to our codebase

+1

> However, I acknowledge that our codebase already has many complex macros, so rather than reject it outright, I leave it to others to evaluate.

A lot of macro has been added with the Link3 merge. At that time several developers complained about the excessive use of unreadable and unmaintainable macro code. And there are good reasons why many companies that care about code quality disallow macros in their developer guidelines.

realthunder:
I'll remove the macro usage and make an amendment soon.

realthunder:
The Macros introduced in this PR is for class `ViewParams`. My intention for such a class is to centralize the definition and accessing of various parameters, especially with their default values, which is diffcult to keep in sync with the current way.  In my branch, there are many classes like `ViewParams` for managing parameters in various groups. The reason of using Macro is to generate direct accessing code for each parameter so that it can be called even in performance critical places like rendering. I am not aware of any way to achieve this repetitive code generation using C++ template.

With the above two commits, the code is now generated using Python package [Cog](https://nedbatchelder.com/code/cog). `Cog` allows to embed the code generator inside the source code. But since we need the same set of parameter definitions in both the header and cpp file, I use a separate file  [ViewParams.py](https://github.com/realthunder/FreeCAD/blob/AppFontSize/src/Gui/ViewParams.py) to define parameters. The code template is defined in [Base/params_utils.py](https://github.com/realthunder/FreeCAD/blob/AppFontSize/src/Base/params_utils.py).

We could use cmake rules to auto run `Cog` when the parameter definition file changes. But I feel it is not necessary right now, as it will introduce a new compile time dependency. The generated class document contains [instructions](https://github.com/realthunder/FreeCAD/blob/feb7228ff75695f53ab9b39f943c11197304126c/src/Gui/ViewParams.h#L38) on how to do it manually.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6333">pipeline status</a> for feature branch [PR_6333](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6333). Pipeline [472746969](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/472746969) was triggered at [feb7228](https://github.com/FreeCAD/FreeCAD/pull/6333/commits/feb7228ff75695f53ab9b39f943c11197304126c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
I don't understand the benefit of cog. To me, it looks like just another preprocessor, with all the downsides that entails. Is there some reason we don't just add `std::variant` getters (and probably templated setters) to the main Parameter class?

realthunder:
For performance, so that for example I can access the parameters at rendering loop. Can you please elaborate on the downsides it bring, other than verbosity I guess?

realthunder:
Oops, wrong click.

wwmayer:
> I am not aware of any way to achieve this repetitive code generation using C++ template.

Over the weekend I was playing to find a solution based on templates: https://github.com/wwmayer/FreeCAD/commits/ParameterObserver

There is a base class that handles as much as possible and the relevant getter and setter functions must be defined in a sub-class. As an example I did it for ViewParams:
https://github.com/wwmayer/FreeCAD/commit/ece4e264b750c0e75ffea9dec268a655f457dbe3

For the class declaration there is no way to make the code any smaller but that's actually good because it keeps it readable. However, for the definition there are still a few pitfalls which can be problematic if developers blindly copy & paste code without making the needed adjustments. The pitfalls are:
* For a getter function (e.g. getUseNewSelection) the template parameter for getValue() must be equal to the return type of the function. I couldn't find an elegant way to achieve that. What works is to write
```
bool ViewParams::getUseNewSelection() const
{
    return getValue<decltype(getUseNewSelection()>("UseNewSelection");
}
```
but I am not sure if this is anyhow better.
* The next issue is that the string passed to getValue or setValue must match with the function name. Theoretically one could use the macro `__func__` and remove the prefix set or get to get the name of the parameter. But again is this the better way?
* Then in the setup() function there are the calls of addParameter() where you define the parameter names, its type and the default value. Here again it's up to the developer to make sure that the parameter names with that of the setter/getter function match and additionally that the used types match with the types of the functions. In the example code the latter is achieved by the static_assert so that you will get an error at compile time.
If you remove the static_assert and you used the wrong types then at least you will get an exception at runtime that exactly tells you which function uses the wrong type.

wwmayer:
> I don't understand the benefit of cog.

The advantage is that it at least generates readable C++ code which compared to the current situation is already a big improvement and allows to easily extend things.

Instead of cog we could also write a generator script with our generateDS framework that we already use for most of our Python bindings.

Now it seems that none of the Linux distributions offer packages for cog but I found its repository here https://github.com/nedbat/cog and it seems we need only two or three files that we could copy to our repository.

wwmayer:
> With the above two commits, the code is now generated using Python package cog.

A few things should still be improved:
* for the sake of consistency the getter functions should use get as prefix, e.g. getUseNewSelection()
* for the sake of clean design I don't see why there should be a static method in DlgGeneralImp. The call of `ViewParams::init()` has to go to another function. If needed create a new class with only an init() function that calls `ViewParams::init()`
* I don't get the point of `ViewParams::onDefaultFontSizeChanged()`. The class ViewParams is about user parameters to control the appearance of view providers but this function changes the font size application wide. This function doesn't have to belong to this class
* If cog allows it then there should be a warning that the following is generated code and shouldn't be touched. Instead the corresponding .py file should be modified.

chennes:
I believe this PR is almost ready, with just the few comments from @wwmayer above that still need addressing. @realthunder  do you have time work on this over the next couple of weeks, to make it into 0.20? Or should we defer to 0.21?

realthunder:
I'll make changes according to suggestion soon.

realthunder:
> * for the sake of consistency the getter functions should use get as prefix, e.g. getUseNewSelection()
> * If cog allows it then there should be a warning that the following is generated code and shouldn't be touched. Instead the corresponding .py file should be modified.

@wwmayer I have made changes according to suggestions

> * for the sake of clean design I don't see why there should be a static method in DlgGeneralImp. The call of `ViewParams::init()` has to go to another function. If needed create a new class with only an init() function that calls `ViewParams::init()`

There are other related changes in my branch about how to handle preference option changes. which I plan to submit as PR in the future. Instead of relying on the user to click that `Apply` button, it will respond to parameter changes. Hence the static function named `attachObserver()` is added to `DlgGeneralImp` and also other preference dialog (see [here](https://github.com/realthunder/FreeCAD/blob/3d159204f13685f8742c29b518f4869c7f19ce88/src/Gui/DlgGeneralImp.cpp#L443) for an example). User can choose whether to see immediate effect on changing options in preference dialog or batch the changes with `Apply` button. This also allows Python code change of preference option applicable without needing to restart FreeCAD, and correctly handle the synchronization between the parameter (which maybe changed by user code) and widgets in preference dialog.

> * I don't get the point of `ViewParams::onDefaultFontSizeChanged()`. The class ViewParams is about user parameters to control the appearance of view providers but this function changes the font size application wide. This function doesn't have to belong to this class

I thought the group name `View` in `BaseApp/Preferences/View` has more general meanings like various views such as property view, 3D view etc, more or less UI related. So, where do you propose we put this `DefaultFontSize` parameter, `BaseApp/Preferences/General` maybe?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6333">pipeline status</a> for feature branch [PR_6333](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6333). Pipeline [495111145](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/495111145) was triggered at [b61e3a3](https://github.com/FreeCAD/FreeCAD/pull/6333/commits/b61e3a3aa1937c3037187dc44131a483560322cf). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6333">pipeline status</a> for feature branch [PR_6333](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6333). Pipeline [496115009](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/496115009) was triggered at [09a12dd](https://github.com/FreeCAD/FreeCAD/pull/6333/commits/09a12dd81ccba3e85d295cd8d4885ed96ebed07a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
We have to postpone this since we are in pure bugfix mode.

adrianinsaval:
this has been sitting here untouched for months, is something still missing or just nobody got around to merging it?

donovaly:
@realthunder , I would like to merge this. Can you therefore please:
- remove the code re factorization (this should go to a separate PR)
- then rebase to fix the merge conflicts

realthunder:
I'll create a new smaller PR for this later.

Issue #6330, opened 2022-02-09
Number formatting customization


0penBrain:
The last info I have is that this issue is not resolved:
- take me as a German, my Windows is a German one, therefore it uses the comma as decimal separator.
- I use FreeCAD in English and therefore want the dot as decimal separator

As it is I have no setting to achieve this.

_Originally posted by @donovaly in https://github.com/FreeCAD/FreeCAD/issues/5871#issuecomment-1031895248_

donovaly:
As in other program, I expect that when I change the language to English, I get
- English menus ->works
- I enter e.g. "5.2 m" I get 5.2 m, but I getting 5200 m -> works not

I am currently working with another person on a sketch and since he cannot speak German, he sets the language to English. It is annoying for him, that he nevertheless mus use the comma to input the decimal sign. (We only have German Windows machines here.)

> FreeCAD to apply same regional number format as selected language in preferences ?

Yes. This is also the intuition.

> An option to enforce usage of C local ?

What do you mean with "C local"? I never heard about this term yet.

0penBrain:
> * I enter e.g. "5.2 m" I get 5.2 m, but I getting 5200 m -> works not

What is the relationship with language or number formatting ? Unit scaling has nothing to do with that. ;)

> > FreeCAD to apply same regional number format as selected language in preferences ?
> 
> Yes. This is also the intuition.

OK, this may be a bit picky hence my below proposal. Will look what can be done.
> > An option to enforce usage of C local ?
> 
> What do you mean with "C local"? I never heard about this term yet.

Roughly "C" (aka "POSIX") locale is a regional setting close to English, but with some differences like simplified number formatting.
I proposed this simplified setting because I think this is main encountered issue. People wanting to enforce usage of "English like" number formatting. C locale has the advantage to be standard across platforms.

0penBrain:
@donovaly I pushed PR #6335 to allow enforcing C locale usage in FreeCAD. You can eventually test it.
In between it's merged, most probably you can solve the problem of your colleague by typing into Python console
```
from Qt import QtCore
QtCore.QLocale.setDefault(QtCore.QLocale.c())
```
Will work on setting number format according selected UI language later on.
HTH

adrianinsaval:
> > * I enter e.g. "5.2 m" I get 5.2 m, but I getting 5200 m -> works not
> 
> What is the relationship with language or number formatting ? Unit scaling has nothing to do with that. ;)

[...]

PR #6335, opened 2022-02-10
Gui: add preference to enforce C locale usage

0penBrain:
This PR brings a first level answer to #6330 by allowing user to enforce C/POSIX locale usage inside FreeCAD.
I think it covers the vast majority of issues regarding number (& date, ...) format incompatibility issues.

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -2099,6 +2099,9 @@ void Application::runApplication(void)
         mainApp.installEventFilter(filter);
     }
 
+    if (hGrp->GetBool("UseCNumberFormat", false))
+        QLocale::setDefault(QLocale::c());
+
     // set text cursor blinking state
     int blinkTime = hGrp->GetBool("EnableCursorBlinking", true) ? -1 : 0;
     qApp->setCursorFlashTime(blinkTime);
--- a/src/Gui/DlgGeneral.ui
+++ b/src/Gui/DlgGeneral.ui
@@ -105,6 +105,23 @@
            </property>
           </widget>
          </item>
+         <item row="1" column="1">
+          <widget class="Gui::PrefCheckBox" name="UseCFormat">
+           <property name="toolTip">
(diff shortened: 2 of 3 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6335">pipeline status</a> for feature branch [PR_6335](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6335). Pipeline [468063577](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/468063577) was triggered at [b1e62a5](https://github.com/FreeCAD/FreeCAD/pull/6335/commits/b1e62a535a461791f2e4a3fa8de2c75b840350cd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
This does not seem to do the right thing because what we need is what e.g. LibreOffice does:
Let's take for example LibreOffice:
Inkedsoffice bin_JcXHowWOVp_LI

The language defines the rules. So German uses comma as decimal separator, English dots. etc. So all we need is an option when one uses FC in English, that one can also takes the decimal sign the language/locale defines.

What you do is to use in every case a special English style and not the decimal character defined by the language. For example in Farsi or Arabic, the decimal separator is neither a comma nor a dot.

So we need in my opinion a simple lookup list. We have the info here:
https://en.wikipedia.org/wiki/Decimal_separator#Usage_worldwide

In the preferences, we add an option. By default it is on, then the language's decimal separator from the lookup list is used according to the chosen language. If the option is turned off, the dot is used when the lookup list says for the language it would be a comma or U+066B. If it would be according to the language lookup list a dot, then the comma is used.

donovaly:
I hope I did not upset you. I only made a proposal which I think would fit. I am open for any discussion and opinions about this.

PR #6344, opened 2022-02-11
Sketcher: bugfix + improvements

0penBrain:
This PR is 3 commits :
 1- [BUGFIX] I found 2 functions where GeoId for external wasn't correctly computed. @abdullahtahiriyo may you check carefully this one, and eventually know if same flaw may exist elsewhere in sketcher ?
2 - Improvement that adds 'construction' status of point in element list. Possiblity for point to be normal or construction has been introduced some months ago IIRC, so this now makes sense
3 - Improvement that displays the GeoId of geometries in element list. This greatly helps user to make a link between constraints extended naming, and geometries extended naming


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6344">pipeline status</a> for feature branch [PR_6344](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6344). Pipeline [469288907](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/469288907) was triggered at [c09a999](https://github.com/FreeCAD/FreeCAD/pull/6344/commits/c09a9994fbdb1e25291af2c154e8840d27195bfa). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
@0penBrain 

Commit 1 is not correct because you are assuming an order in the complete geometry array different from the actual one.

When generating the complete geometry array, the external geometry elements are added using a reverse iterator:
https://github.com/FreeCAD/FreeCAD/blob/f179acf0625354dd79af7cc2d298206702beaede/src/Mod/Sketcher/App/SketchObject.cpp#L7192

I think you modified the code assuming that the element stored next to the last element of internal geometry is the external geometry with index -3, but this is not the case. The last element of the complete geometry array is GeoId = -1. So the third from the end is actually the one with index=-3.

Does this make sense?

abdullahtahiriyo:
@0penBrain 

The last commit then will not work.

For the last commit I will use an index that gets updated in the loop to add the GeoId to the string. One for the internal in the internal loop and one for the external in the external loop.

0penBrain:
> Does this make sense?

Your explanation yes. Implementation choice ... far less

> For the last commit I will use an index that gets updated in the loop to add the GeoId to the string. One for the internal in the internal loop and one for the external in the external loop.
OK. In the absolute, I find it horrible to not rely on geometry handling in SketchObject. Because if something needs change there, it can virtually break everything. However I'm trusting you that this is the best solution right now. PR is fixed. Thanks for review.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6344">pipeline status</a> for feature branch [PR_6344](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6344). Pipeline [470452295](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/470452295) was triggered at [799a5ea](https://github.com/FreeCAD/FreeCAD/pull/6344/commits/799a5eae8c0e41d91a5a9ac0ecd8cc08d5803ca7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
> OK. In the absolute, I find it horrible to not rely on geometry handling in SketchObject. Because if something needs change there, it can virtually break everything. However I'm trusting you that this is the best solution right now. PR is fixed. Thanks for review.

Yes, I acknowledge that it is far from ideal and it couples to a given implementation, which is in general undesirable. On the other hand, if we really change the sketcher to such an extent, this will be the lesser of the problems of the Elements Widget.

Taking into account the state of affairs, it is an acceptable solution. It is coupled but encapsulated. We can improve it later on.

PR #6348, opened 2022-02-11
[Sketcher] Use `DrawSketchHandler` when inserting a B-Spline knot

AjinkyaDahale:
This set of commits is supposed to be able to give the user some feedback of where the new point will be.

I am having some issues as I try to get this going so someone with better Qt knowledge can suggest how to get around them.

Current issues:
1. ~When I use a `knotDialog.show()` or `knotDialog.open()` (either in `CmdSketcherInsertKnot::activated()` or `DrawSketchHandlerBSplineInsertKnot::activated()`), the dialog is not visible. However with `knotDialog.exec()` it is visible. I need the `show()` method to be able to let the user see the viewport.~ Never mind found out it was a scope issue.
2. ~I do not know how to use signals and slots when some of the relevant objects are not subclasses of `QObject`, which is the case with commands and `DrawSketchHandler`'s.~ Deciding to skip the dialog box altogether for now.

This commit will probably remain a draft until @PaddleStroke's tool settings taskbox is merged since that is probably the best way to ask for the knot value. However, theoretically, it does appear we can work with a `QInputDialog` till then so we can work concurrently.

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
+++ b/src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
@@ -972,6 +972,97 @@ bool CmdSketcherCompModifyKnotMultiplicity::isActive(void)
     return isSketcherBSplineActive(getActiveGuiDocument(), false);
 }
 
+class DrawSketchHandlerBSplineInsertKnot: public DrawSketchHandler
+{
+public:
+    DrawSketchHandlerBSplineInsertKnot(const Part::GeomBSplineCurve* _bsp)
+        : bsp(_bsp)
+    {}
+
+    ~DrawSketchHandlerBSplineInsertKnot() override {}
+
+    void activated(ViewProviderSketch *) override
+    {
+        // TODO: create the dialog box here
+        const auto& knots = bsp->getKnots();
+        const auto& mults = bsp->getMultiplicities();
+        knotDialog = new QInputDialog(Gui::getMainWindow());
+        knotDialog->setModal(false);
+        knotDialog->setInputMode(QInputDialog::DoubleInput);
+        knotDialog->setDoubleDecimals(Base::UnitsApi::getDecimals());
+        knotDialog->setWindowTitle(QObject::tr("Knot parameter?"));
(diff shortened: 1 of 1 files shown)


AjinkyaDahale:
@donovaly can you help out with any of the issues?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6348">pipeline status</a> for feature branch [PR_6348](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6348). Pipeline [469448453](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/469448453) was triggered at [d2d7f69](https://github.com/FreeCAD/FreeCAD/pull/6348/commits/d2d7f698576a2af13b32321f79803c421f1e5fb1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I propose that you make the feature a normal task dialog. I mean it actually is a task where you set parameter and get a live preview. Also with a task dialog you can benefit from an almost perfect code infrastructure.

AjinkyaDahale:
> I propose that you make the feature a normal task dialog. I mean it actually is a task where you set parameter and get a live preview. Also with a task dialog you can benefit from an almost perfect code infrastructure.

It does make sense. However we are already in the edit sketch task. Got any Idea how to call a task within a task?

donovaly:
You can stack task dialogs. The Sketch task dialog is already an example - in consists of several task dialogs. So you could add another one when the knot insertion is called and minimize the other open ones.

However, this might have side effects I don't know. @abdullahtahiriyo , what do you think?

AjinkyaDahale:
> You can stack task dialogs. The Sketch task dialog is already an example - in consists of several task dialogs. So you could add another one when the knot insertion is called and minimize the other open ones.

Oh you meant task _boxes_ . I think that will be the eventual way, mostly when @PaddleStroke's tool settings task box is merged

PR #6360, opened 2022-02-14
[Part] improve SectionCutting code

donovaly:
all changes as suggested in https://github.com/FreeCAD/FreeCAD/pull/4118/files/05a9c88ab2dad40d4393203bed5aa861e5587d8d


wwmayer (approved):
Security-wise it looks OK now.

However, what could still be improved is reducing the code duplication. The onCut*valueChanged and onFlip*clicked functions are all very similar and with some extra effort these six functions can be replaced by two more general functions.

donovaly:
Thanks for having a look.

Concerning onFlipClicked I already thought about it. Since it is nested - X needs more code than Y and this less than Z - one does not get a better overview I thought and therefore did not do this..

However, I will have a look the next days and teen also for onCutvalueChanged. I will then make a new PR.

PR #6361, opened 2022-02-14
Addon Manager: interface and backend to disable addons

chennes:
This PR adds the ability to disable the loading of an Addon without actually uninstalling it. This is useful for scripted testing, as well as for disabling memory-intensive addons when they aren't needed. The mechanism is designed to be simple enough that a shell script can easily achieve it: a simple "ADDON_DISABLED" file is created in the Addon's root directory, and checked for when FreeCAD starts. The file's contents are not examined, only its existence matters. A simple GUI interface is added to the Addon Manager to create and delete this file, and the interface is updated to display a message when an Addon is disabled.

Fixes https://github.com/FreeCAD/FreeCAD-addons/issues/204

### DIFF
--- a/src/App/FreeCADInit.py
+++ b/src/App/FreeCADInit.py
@@ -31,232 +31,243 @@
 import FreeCAD
 
 def removeFromPath(module_name):
-	"""removes the module from the sys.path. The entry point for imports
-		will therefore always be FreeCAD.
-		eg.: from FreeCAD.Module.submodule import function"""
-	import sys
-	paths = sys.path
-	for path in paths:
-		if module_name in path:
-			sys.path.remove(path)
-			return
-	Wrn(module_name + " not found in sys.path\n")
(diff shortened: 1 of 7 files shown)


luzpaz:
This is sweet for helping to troubleshoot issues for users as well! Thanks!

yorikvanhavre:
Looks great!
Is there anything changed in FreeCADInit.py? Because of the tab->spaces change it's hard to know...

chennes:
Yes, sorry, I intended to put that in a different commit, and forgot until it was too late to easily revert. There are some added lines there to search for the stopfile: actually, identical changes to those in FreeCADInitGui.py, with just an additional line that prints out a diagnostic message.

wwmayer:
To ignore all white-space changes add `?w=1` at the end of the url.

luzpaz:
Also via the UI  
image

PR #6364, opened 2022-02-14
Gui: add preference to use selected language number formatting

0penBrain:
If not enabled (default), will defaults to C/POSIX formatting

Currently, the preference is disabled so on a blank config it will default to C locale.
This is the safest option as C is supported cross-platform, but could be changed anyway.


adrianinsaval:
Thanks for this, what's the danger with having te option on as default? I think that would be more intuitive for most regular people using FreeCAD with a different language, I personally will leave it off as dot is more convenient to me but default config should be decided with the intent of being comfortable for regular users.

0penBrain:
@adrianinsaval There's not much precision in Qt website and I can't be sure that after calling `setlocale`, we can safely use `QLocale::setDefault` in all situations.
https://doc.qt.io/qt-5/qcoreapplication.html#locale-settings

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6364">pipeline status</a> for feature branch [PR_6364](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6364). Pipeline [470442992](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/470442992) was triggered at [94743cc](https://github.com/FreeCAD/FreeCAD/pull/6364/commits/94743ccc68d29366162e527d8738c8e127fcf67e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/Application.cpp
```diff
@@ -1947,13 +1947,11 @@ void Application::runApplication(void)
     // http://forum.freecadweb.org/viewtopic.php?f=3&t=15540
     mainApp.setAttribute(Qt::AA_DontShowIconsInMenus, false);
 
-#ifdef Q_OS_UNIX
```
wwmayer, line 1950: Have you tested it that dropping this check doesn't cause problems on other OSes?

src/Gui/Language/Translator.cpp
```diff
@@ -251,6 +251,19 @@ std::string Translator::locale(const std::string& lang) const
     return loc;
 }
 
+bool Translator::setLocale(const std::string& language) const
+{
+    auto loc = QLocale::c(); //Defaulting to POSIX locale
+    auto bcp47 = Translator::instance()->locale(language);
```
wwmayer, line 257: The call of `Translator::instance()` should be dropped as calling the function setLocale() is done for a valid Translator object.

src/Gui/Application.cpp
0penBrain: Nope. I checked in possible extent that the function exists on other OSes.
I wish the CI is building on all OSes but as it's not the case anymore, have to bet on beta users. :)
Do you think we should ask to alpha users ?

src/Gui/Language/Translator.cpp
0penBrain: Indeed. Copy-paste syndrome. Thanks for the catch. Will update soon.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6364">pipeline status</a> for feature branch [PR_6364](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6364). Pipeline [471759217](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/471759217) was triggered at [fc6c60d](https://github.com/FreeCAD/FreeCAD/pull/6364/commits/fc6c60d6ada9d49c15cdfe0461c7bd95599992b6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
@donovaly may you have some time to test this PR on Windows? Thx

donovaly:
> @donovaly may you have some time to test this PR on Windows? Thx

Many thanks for this PR! This is what I wanted. I can now set the language to English and enter then dimensions using the dot as decimal separator - as in many other programs.

I only checked the functionality, not the code.

0penBrain:
> I only checked the functionality, not the code.

Yes, this was it. Actually the concern was about a system function call that could be troublesome on certain OSes.
Can I ask an extra test ? Could you delete the 'Preferences/General/Language' parameter and see if everything goes well when starting up FC without it (I'd expect your language to the one of the OS --German-- and the numbers to be formatted accordingly) ? Sorry I forgot this in my 1st request. Thanks

donovaly:
> Could you delete the 'Preferences/General/Language' parameter and see ...

I will do so later. Since recent changes, the compilation takes ages. Could you therefore please rebase?

donovaly:
- I removed the 'Preferences/General/Language' parameter
- I used:
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.27547 +1 (Git)
Build type: Release
Branch: pr/6364
Hash: fc6c60d6ada9d49c15cdfe0461c7bd95599992b6
Python version: 3.8.6+
Qt version: 5.15.2
Coin version: 4.0.1
OCC version: 7.5.3

- I started FC, then selected English as language, and use the option to use the local number format.

Result:
FreeCAD_XYRnyRHDEh


So with every click into a quantity field I get 3 digits extra. This issue also occurs when I don't use the option to use the option to use the local number format.

0penBrain:
@donovaly Many thanks. I wouldn't have catch this as I think it's specific to German (or any language using comma as decimal separator).
Do you remember if it still happens after restarting FC? 
Setting PR draft.

0penBrain:
@donovaly It's hopefully fixed with new commit. May you have a check please -- I can't really reproduce -- ? I rebased against master so compilation time should be better. :) Thax

@wwmayer : what is the way to update `customwidgets.cpp` please ?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6364">pipeline status</a> for feature branch [PR_6364](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6364). Pipeline [478983772](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/478983772) was triggered at [9662dd0](https://github.com/FreeCAD/FreeCAD/pull/6364/commits/9662dd088c8a1598b40c73467d860936b291da93). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> Do you remember if it still happens after restarting FC?

This happened also after a restart.

donovaly:
> May you have a check please

Thanks! This works now fine. I tested it as best as I could. Since @wwmayer is a German too, we can test it maybe under Linux. (I have a Linux but no build system on it.)

0penBrain:
> Thanks! This works now fine. I tested it as best as I could. Since @wwmayer is a German too, we can test it maybe under Linux. (I have a Linux but no build system on it.)

According to fix in 2nd commit, this shouldn't be OS dependent (only language dependent). I originally asked you for another change that could have been troublesome on Windows. For this latter, I already tested on Linux. ;)
Many thanks for helping reviewing, I'm happy you screened a bug before merging.
Now only to find how to update customwidgets. :)

donovaly:
> Now only to find how to update customwidgets. :)

@wwmayer ?

Besides this, can you please change the PR status to normal, meaning ready to be merged? In my opinion this should go in for FC 0.20.

wwmayer:
> Now only to find how to update customwidgets

You mean the file for the plugins for Qt designer? As long as no new widget has been added nothing needs to be done there. The plugin actually only offers the widgets for their visual representation but some of them are dysfunctional because otherwise it needed a dependency to FreeCADBase

0penBrain:
> You mean the file for the plugins for Qt designer? 

I guess it is. I was wondering because I saw code of QuantitySpinBox is duplicated in it. So I supposed it has to be updated.

donovaly:
So the commit is ready to be merged as it is, right?

0penBrain:
@wwmayer I talk about this file (especially these lines) https://github.com/FreeCAD/FreeCAD/blob/420fbb3d28deb6cbb7dd5792d453fc74a1c0ea63/src/Tools/plugins/widget/customwidgets.cpp#L655-L861 where the code I modified is copied.
So I guessed there was a script or something like that so it is updated. Let me know please.

chennes:
@donovaly I think this can go into 0.20 (in fact with no further changes), but I leave it to you to make the final determination.

donovaly:
> @donovaly I think this can go into 0.20 (in fact with no further changes), but I leave it to you to make the final determination.

@wwmayer reviewed the PR, I only tested it ;-) So Werner, can you please do the merge?

donovaly:
I tested it again, this time with the FEM module. The PR works nicely and helps me as German users to use FEM at all. (There is a long-standing bug that the FEM material settings don't accepts comma as decimal separator.)

However, since FC 0.20 comes close, I would like to have it in now that more people can test it using the weekly build. So @wwmayer if you don't see issues I would merge it next evening.

donovaly:
OK, I will now merge. It is time that this goes into our weekly build, especially to get feedback as soon as possible also from MacOS users.

donovaly:
I did not test well enough:
For all App:PropertyFloat the display is broken. Take for example the number of turns in a PD Helix:
FreeCAD_pxxYAkRlaA

For other properties like Length, Angle, it is correct.
@0penBrain , can you please have a look?

donovaly:
The one above is only cosmetic it seems, but here is another one where I cannot use the dot:

- I use English menus in FC and use the option to use the language's number format
- in dialogs that use the Gui::InputField component, I still need to use the comma to get the decimal separator.

donovaly:
there is another one where I cannot use the dot:
in dialogs that use the Gui::InputField component, I still need to use the comma to get the decimal separator.

So there are 2 places, where it does not work:
- App::PropertyFloat
- Gui::InputField

However, since Gui::InputField relies on App::PropertyFloat, there is probably only one case to fix

0penBrain:
@donovaly will look at this during next week. What are you're settings? 'English' + 'Use selected language formatting'?

donovaly:
> @donovaly will look at this during next week. What are you're settings? 'English' + 'Use selected language formatting'?

Yes, exactly.

0penBrain:
@donovaly bit strange here. I'm now stuck with any input field, whatever is its type.
Even QuantitySpinBox behaves strangely when I use "English" and "Use selected language format".
For example, when entering a QuantitySpinBox, only the integer part is selected, and I cannot use 'dot' but have to use 'comma'.
Roughly numbers are well displayed, but I have to use your system locale format to input them. Same on your side ?

donovaly:
> For example, when entering a QuantitySpinBox, only the integer part is selected, and I cannot use 'dot' but have to use 'comma'. Roughly numbers are well displayed, but I have to use your system locale format to input them. Same on your side ?

No, for me there is no problem with QuantitySpinBox. I checked right now using PD Pad. (I compile at the moment without precompiled headers but don't think this makes a difference here.

PR #6370, opened 2022-02-14
Fem: enable multi-threading for Gmsh

atorger:
This patch will add the General.NumThreads = <cpu core count> setting to the shape2mesh.geo configuration file for gmsh. This will cause gmsh to run with multiple threads (it uses OMP internally). It does not affect gmsh binaries without OMP support.

Great speedups can are had, in a project I'm working on currently the meshing takes 2,5 minutes with a single thread, reduced to only 40 seconds when multi-threaded.

### DIFF
--- a/src/Mod/Fem/femmesh/gmshtools.py
+++ b/src/Mod/Fem/femmesh/gmshtools.py
@@ -733,6 +733,13 @@ class GmshTools():
         geo = open(self.temp_file_geo, "w")
         geo.write("// geo file for meshing with Gmsh meshing software created by FreeCAD\n")
         geo.write("\n")
+
+        cpu_count = os.cpu_count()
+        if cpu_count != None and cpu_count > 1:
+            geo.write("// enable multi-core processing\n")
+            geo.write(f"General.NumThreads = {cpu_count};\n")
+            geo.write("\n")
+
         geo.write("// open brep geometry\n")
         # explicit use double quotes in geo file
         geo.write('Merge "{}";\n'.format(self.temp_file_geometry))
--- a/src/Mod/Fem/femtest/app/support_utils.py
+++ b/src/Mod/Fem/femtest/app/support_utils.py
@@ -228,17 +228,21 @@ def compare_files(
     file1 = open(file_name1, "r")
     f1 = file1.readlines()
     file1.close()
+
+    # TODO: add support for variable values in the reference file
+    # instead of using this workaround
+
     # workaround to compare geos of elmer test and temporary file path
     # (not only names change, path changes with operating system)
     lf1 = [l for l in f1 if not (
-        l.startswith('Merge "') or l.startswith('Save "') or l.startswith("// ")
+        l.startswith('Merge "') or l.startswith('Save "') or l.startswith("// ") or l.startswith("General.NumThreads")
     )]
     lf1 = force_unix_line_ends(lf1)
     file2 = open(file_name2, "r")
     f2 = file2.readlines()
     file2.close()
     lf2 = [l for l in f2 if not (
-        l.startswith('Merge "') or l.startswith('Save "') or l.startswith("// ")
+        l.startswith('Merge "') or l.startswith('Save "') or l.startswith("// ") or l.startswith("General.NumThreads")
     )]
     lf2 = force_unix_line_ends(lf2)
     import difflib
--- a/src/Mod/Fem/femtest/data/elmer/group_mesh.geo
+++ b/src/Mod/Fem/femtest/data/elmer/group_mesh.geo
@@ -1,5 +1,8 @@
 // geo file for meshing with Gmsh meshing software created by FreeCAD
 
+// enable multi-core processing
+General.NumThreads = X;
+
 // open brep geometry
 Merge "/tmp/tmp0TVZbM.brep";
 


luzpaz:
OP forgot to add forum thread: https://forum.freecadweb.org/viewtopic.php?f=18&t=66241 :wink:

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370">pipeline status</a> for feature branch [PR_6370](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370). Pipeline [470917027](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/470917027) was triggered at [fd6ac4b](https://github.com/FreeCAD/FreeCAD/pull/6370/commits/fd6ac4b08d622fd1bfdeffc1b2e82ff6c0a3d640). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Fem/femmesh/gmshtools.py
```diff
@@ -733,6 +733,13 @@ def write_geo(self):
         geo = open(self.temp_file_geo, "w")
         geo.write("// geo file for meshing with Gmsh meshing software created by FreeCAD\n")
         geo.write("\n")
+
+        cpuCount = os.cpu_count()
```
plaes, line 737: From the docs of `os.cpu_count()`:
> This number is not equivalent to the number of CPUs the current process can use. The number of usable CPUs can be obtained with len(os.sched_getaffinity(0))

atorger: Locking processes to specific CPU:s may be done on a realtime server, but I doubt that anyone does that on a computer that runs regular applications like CAD software. So I'm not sure it's worthwhile to make the code less readable for a use case that doesn't really exist. In addition multiprocessing.cpu_count() is already used when starting Calculix multi-threaded (which is a wrapper of os.cpu_count())

plaes: There are people who run these processes under containerized workflows.

atorger: *  sched_affinity is a unix/posix-specific function, as said in the docs "note: This method is only available on some UNIX platforms."
* cpu_count() is already used in FreeCAD.
* nothing will break if you in some special situation assign more threads than CPUs, it's just somewhat less optimal.

plaes: Ok, just confirmed that `os.cpu_count()` actually shows correct number of allowed cores in the lxd container.

atorger: Thanks, great to know, I did not know that.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370">pipeline status</a> for feature branch [PR_6370](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370). Pipeline [471760423](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/471760423) was triggered at [fc737e0](https://github.com/FreeCAD/FreeCAD/pull/6370/commits/fc737e0f1b179daa2cd3675b42030c7754ed5aed). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Fem/femmesh/gmshtools.py
AjinkyaDahale: You could still add a comment mentioning where this might cause problems. Additionally, even for people using it on a "computer that does regular operations", we may still want to have some free cores (like we do when compiling with `make -j$(nproc --ignore=2)`). I say `cpuCount=None` can be an argument (also please keep to snake case and use `cpu_count` unless that causes issues) so the user has an option to set it when running.

AjinkyaDahale:
Among the failing tests, one is just because of the new input file that shall be created.
```
  File "/builds/freecad/FreeCAD-CI/build/Mod/Fem/femtest/app/test_solver_elmer.py", line 162, in test_box_static_0_mm
    self.assertFalse(ret, "GMSH geo write file test failed.\n{}".format(ret))
AssertionError: 'Comparing /builds/freecad/FreeCAD-CI/build/Mod/Fem/femtest/data/elmer/group_mesh.geo to /tmp/FEM_unittests/solver_elmer_box_static_0_mm_b03f821d7bf5/group_mesh.geo failed!\n--- \n+++ \n@@ -0,0 +1,2 @@\n+\n+General.NumThreads = 4;\n' is not false : GMSH geo write file test failed.
Comparing /builds/freecad/FreeCAD-CI/build/Mod/Fem/femtest/data/elmer/group_mesh.geo to /tmp/FEM_unittests/solver_elmer_box_static_0_mm_b03f821d7bf5/group_mesh.geo failed!
--- 
+++ 
@@ -0,0 +1,2 @@
+
+General.NumThreads = 4;
```

I believe this can be corrected by changing the reference file, but use your discretion before doing so.

src/Mod/Fem/femmesh/gmshtools.py
atorger: If there is some policy for FreeCAD to not use all cores I think you should implement that top-down with a global setting in preferences. There is no such setting. Until then I think it's much better to use all available cores rather than just one like now. When I run gmsh I want it to finish as quickly as possible.

I don't think it's necessary with a comment. Using all cores for OMP processing is standard procedure. It's not like the computer will go suddenly unresponsive as there's no I/O processing, it's just pure maths calculations i RAM which the OS has no issues scheduling together with user interface tasks.

I'll change the patch to snake case.

atorger:
I have now updated the CI test. It's a bit ugly but was already before as the compare_files() function excludes lines with variable content rather than have support for such values, so I just added one other exception for the NumThreads setting (which will vary depending on core count on the test platform).

I haven't learnt how to run the CI locally though so I've done the change in the blind so I'm not 100% sure it will pass...

src/Mod/Fem/femmesh/gmshtools.py
AjinkyaDahale: > If there is some policy for FreeCAD to not use all cores I think you should implement that top-down with a global setting in preferences. There is no such setting. Until then I think it's much better to use all available cores rather than just one like now. When I run gmsh I want it to finish as quickly as possible.

FreeCAD as a program itself is for the most part serial and calls for more cores only at certain times. You are correct that we need some such setting so that it doesn't hog all processing power for a long time. It is also possible to want GMSH/Calculix and other external programs to use more cores, or run on a different computer altogether.

> 
> I don't think it's necessary with a comment. Using all cores for OMP processing is standard procedure. It's not like the computer will go suddenly unresponsive as there's no I/O processing, it's just pure maths calculations i RAM which the OS has no issues scheduling together with user interface tasks.

You would be surprised. "Just pure math calculations" is basically the crux of all scientific computing that can up millions of cores. One only needs to give a mesh size too small and they're stuck with no response for a long time.

> 
> I'll change the patch to snake case.

Thanks.

AjinkyaDahale:
> I haven't learnt how to run the CI locally though so I've done the change in the blind so I'm not 100% sure it will pass...

For Linux the command is `bin/FreeCAD -t 0` (or whatever the command is). You can also replace `0` with the name of a test to just run that one.

src/Mod/Fem/femmesh/gmshtools.py
atorger: I don't think we should be arguing about how scheduling of threads work. These type of calculations in RAM with no I/O using all available cores will not lead to an unresponsive computer unless you run out of memory, in which case the computer gets unresponsive even with just one thread. This is not at all the same as parallel compilation for example which is a very I/O-intensive process. With I/O-intensive operations multiple threads can indeed bog down the computer. With no I/O, the OS kernel has no trouble scheduling the interactive foreground processes responsively even with 100% load on all cores.

Running the mesher and solver multi-threaded is a big time saver with virtually no drawbacks.

I think that desktop software should be as performant as possible and should not be made artificially slower just for some special use case that very few users will be interested in. And indeed, FreeCAD's default setting for Calculix is to run with all available cores. How many users goes in and changes that number? (Well, actually I did change it to match my core count because I didn't realize that the default value 1 means use all cores rather than just one core, but that's another issue).

I'm not against having multi-core configuration in the preferences for these rare special needs, but it's not good like now have it only for Calculix (and not well-formulated what the setting actually does), so I think one should file a separate bug/issue about multi-core configuration to consolidate it into one. I would myself consider that a low priority feature though, while actually running multi-threaded per default is a high priority and a very easy fix which should be a no-brainer. In my current FEM/modeling project I have run the mesher 100s of times, I save almost 2 minutes each time (2:30 to down to 40 seconds).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370">pipeline status</a> for feature branch [PR_6370](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370). Pipeline [472748072](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/472748072) was triggered at [553fb54](https://github.com/FreeCAD/FreeCAD/pull/6370/commits/553fb54b3409c1ad7c13958f1c4127256d1c01ca). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Fem/femmesh/gmshtools.py
atorger: To add to the discussion, for at least like the past 15 years computers have got increasingly parallel with more cores as it has not been possible to scale up the speed of a single core as much as was possible back in say the 1990s when I started out. This means that to make efficient desktop software you should write the software to use as many cores as possible when it improves performance.

As an example, the open-source raw converter rawtherapee that I contributed to over a few years we had a guy whose only job was to parallelize and optimize existing code to make the algorithms use as many cores as possible as efficiently as possible. It had results and the software became very responsive. No user ever complained about the software using too many cores, but instead rejoiced over the performance improvements.

I have used OMP and other parallelism APIs myself in many software projects to get more responsive software, it's a natural part of modern software development. These days you even look at putting parallel algorithms on the graphics card (GPU) when possible.

If FreeCAD is truly intended to be a single-threaded software for most part and that is seen as a desirable feature rather than something that limits performance, I would be most surprised.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370">pipeline status</a> for feature branch [PR_6370](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370). Pipeline [474609633](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/474609633) was triggered at [e6e3ff7](https://github.com/FreeCAD/FreeCAD/pull/6370/commits/e6e3ff784aa2fdd7a7e1adcbb319ae0b094d0166). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

atorger:
I'm not sure what this patch is stuck on... is there something more I should do?

It's a really nice patch to have when doing trial/error search for failed tets... saves lots and lots of time.

luzpaz:
We haven't heard from @berndhahnebach the FEM lead dev

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370">pipeline status</a> for feature branch [PR_6370](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370). Pipeline [476917241](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/476917241) was triggered at [a67a715](https://github.com/FreeCAD/FreeCAD/pull/6370/commits/a67a7154d865eee383883523b365ad9d593acd98). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I will have a look next weekend. The patch looks good and since I contributed some FEM stuff for 0.20 I should be able to review the PR.
Can you just please do me the favor and rebase the PR against current master. (This saves compilation time for me when testing.)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370">pipeline status</a> for feature branch [PR_6370](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6370). Pipeline [490566560](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/490566560) was triggered at [5d32afc](https://github.com/FreeCAD/FreeCAD/pull/6370/commits/5d32afc7616b51ecc9542688e402c562b85f51e1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Fem/femtest/app/support_utils.py
```diff
@@ -228,17 +228,21 @@ def compare_files(
     file1 = open(file_name1, "r")
     f1 = file1.readlines()
     file1.close()
+
+    # TODO: add support for variable values in the reference file
```
donovaly, line 232: I don't understand this comment. What is missing and why can't it be added right now?

donovaly (commented):
Thanks for the PR. I tested it but it has no effect. GMSH is still only calculating on one core.

Also the long-standing issue that while GMSH is running FC becomes unresponsive and one cannot stop GMSH.

What do I have to do to active the feature of the PR to have GMSH run on several CPU kernels?

This is my test setup:
```
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.28030 +1 (Git)
Build type: Release
Branch: pr/6370
Hash: 5d32afc7616b51ecc9542688e402c562b85f51e1
Python 3.8.6+, Qt 5.15.2, Coin 4.0.1, OCC 7.5.3
Installed mods: 
  * A2plus 0.4.54b
  * Assembly4 0.11.10
  * fasteners 0.3.38
  * Manipulator
```

This is my GMSH version:
```
Version       : 4.9.2
License       : GNU General Public License
Build OS      : Windows64
Build date    : 20211223
Build host    : gmsh.info
Build options : 64Bit ALGLIB[contrib] ANN[contrib] Bamg Blas[petsc] Blossom Cgns DIntegration DomHex Eigen[contrib] Fltk Gmm[contrib] Hxt Jpeg Kbipack Lapack[petsc] MathEx[contrib] Med Mesh Metis[contrib] Mmg Mpeg Netgen NoSocklenT ONELAB ONELABMetamodel OpenCASCADE OpenCASCADE-CAF OpenGL OpenMP OptHom PETSc Parser Plugins Png Post QuadMeshingTools QuadTri Solver TetGen/BR Voro++[contrib] WinslowUntangler Zlib
FLTK version  : 1.4.0
PETSc version : 3.15.0 (real arithmtic)
OCC version   : 7.6.0
MED version   : 4.1.0
Packaged by   : nt authority system
Web site      : https://gmsh.info
Issue tracker : https://gitlab.onelab.info/gmsh/gmsh/issues
```

donovaly:
> This patch will add the General.NumThreads = setting to the shape2mesh.geo configuration file for gmsh. This will cause gmsh to run with multiple threads (it uses OMP internally). It does not affect gmsh binaries without OMP support.

In my shape2mesh.geo I don't see `NumThreads ` so Python cannot access the number of CPU kernels it seems.

src/Mod/Fem/femtest/app/support_utils.py
atorger: I just added it as a suggestion if someone wants to improve the test case in the future. The current test case just removes any lines with variables in them ('Merge' and 'Save'), and as NumThreads is also a variable that is removed too.

donovaly:
OK, the issue was on my build system. Now it works and I am running tests.

atorger:
The ccxtools.py uses the same technique (getting the cpu_count in python) for the Solver, so it's no the first time this is used. I'm using os.cpu_count() instead of multi_processing.cpu_count() though as the former is today more commonly used and use the os module which is already imported.

donovaly:
OK. The patch is straight forward and works.

For the records: with my test case (2nd order mesh) meshings is about 43 % faster.

luzpaz:
Any volunteers for adding this to https://wiki.freecad.org/Release_notes_0.20 ?

donovaly:
> Any volunteers for adding this to https://wiki.freecad.org/Release_notes_0.20 ?

I put it already in there.

PR #6378, opened 2022-02-15
Gracefully handle op creation abort when no TC is available or selected.

mlampert:
Assuming a job with more than one TC but no previous op yet - the user gets presented with a dialog to select one TC for the op. If the user cancels the selection the op creation "half way fails" with an exception.

Similarly if the op to be created has tool constraints (implements `isToolSupported`) and no suitable tool is in the job's Tools group the creation fails with an exception traceback in the report view. Something about `None` not understanding some method ....

Both these cases now throw an exception which is handled in the op creation algorithm. The former case generates a legible warning since the abort was most likely user initiated anyway, and the latter results in an error telling the user what the problem is.

### DIFF
--- a/src/Mod/Path/PathScripts/PathOp.py
+++ b/src/Mod/Path/PathScripts/PathOp.py
@@ -68,6 +68,13 @@ FeatureDiameters = 0x4000  # Turning Diameters
 
 FeatureBaseGeometry = FeatureBaseVertexes | FeatureBaseFaces | FeatureBaseEdges
 
+class PathNoTCException(Exception):
+    '''PathNoTCException is raised when no TC was selected or matches the input
+    criteria. This can happen intentionally by the user when they cancel the TC
+    selection dialog.'''
+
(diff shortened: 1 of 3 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6378">pipeline status</a> for feature branch [PR_6378](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6378). Pipeline [471996784](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/471996784) was triggered at [09c10da](https://github.com/FreeCAD/FreeCAD/pull/6378/commits/09c10dada05dfd214d231a89d69eabb663351126). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #6379, opened 2022-02-16
Part: allow to skip some dialogs when user desire can be inferred bas…

mwganson:
…ed on object selection

Accidentally closed the other PR for this.

https://github.com/FreeCAD/FreeCAD/pull/5378

So, I started this one.

With this PR users can skip some dialogs in Part workbench: Extrude, Sweep, Loft, and Revolve, but only if a proper preselection of objects is done before clicking the toolbar icon.  There is also a user parameter for disabling this behavior (always showing the dialog).

Extrude: select a sketch (or other suitable profile) in the tree and an edge in the 3d view.  The edge will be used to determine the direction and length of the extrusion.

Loft: select 2 or more profiles (whether in the tree or 3d view doesn't matter for loft).

Sweep: select 1 or more profiles in the tree and one or more edges in the 3d view.

Revolve: select 1 profile in the tree and a line segment or arc in the 3d view.  The selected subobject will be used as the axis of the revolve.

### DIFF
--- a/src/Mod/Part/Gui/Command.cpp
+++ b/src/Mod/Part/Gui/Command.cpp
@@ -56,25 +56,24 @@
 #include <Gui/View3DInventor.h>
 #include <Gui/View3DInventorViewer.h>
 #include <Gui/WaitCursor.h>
-#include <Mod/Part/App/Part2DObject.h>
-#include <Mod/Part/App/PartFeature.h>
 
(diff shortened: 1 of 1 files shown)


mwganson:
There were some unintended changes I shall need to fix.  Switching to draft mode for now.

PR #6381, opened 2022-02-16
[PD] change default FaceMaker to Bullseye

### LINKED ISSUE #6366: [bug] Part Design Pad/Pocket rejection nested profiles despite there are other solids assuring there won't be several bodies as result
### Forums discussion

starting here https://forum.freecadweb.org/viewtopic.php?f=10&t=66281

### Version

0.20 (Development)

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.27428 (Git)
Build type: Release
Branch: master
Hash: 27460358508a2057e0ec57a418641435f12628dd
Python version: 3.8.6+
Qt version: 5.15.2
Coin version: 4.0.1
OCC version: 7.5.3
Locale: Spanish/Paraguay (es_PY)
```

### Subproject(s) affected?

PartDesign

### Idea description

While double nested profiles would result in multiple solids if using on a pad (or other feature) in an empty body, it can very well be used to produce a single solid if there is an existing shape in the body. Currently PD rejects this type of sketch but IMO it is not necessary to reject it, we can see in Part Extrude that his kind of shapes work without issues. See attached example were attempting to pad Sketch001 is rejected but wouldn't lead to multiple solid if accepted and works without issues in Extrude003
[extrudeExample-singleSolid.zip](https://github.com/FreeCAD/FreeCAD/files/8061178/extrudeExample-singleSolid.zip)

Example revolves around pad but this possibly applies to most other PD features.

donovaly:
Part does already use this FaceMaker since it can also handle islands.

The PR therefore
- uniforms the behavior of PD and Part
- fixes #6366
- enables to use nested sketches for the various PD features like Revolve

Here is an example file what becomes now possible:
[revolveExample.zip](https://github.com/FreeCAD/FreeCAD/files/8076715/revolveExample.zip)
FreeCAD_Ifx8RosEme

### DIFF
--- a/src/Mod/PartDesign/App/FeatureSketchBased.cpp
+++ b/src/Mod/PartDesign/App/FeatureSketchBased.cpp
@@ -72,18 +72,18 @@
 
 #endif
 
+#include <App/OriginFeature.h>
+#include <App/Document.h>
 #include <Base/Exception.h>
 #include <Base/Parameter.h>
 #include <Base/Reader.h>
 #include <Base/Console.h>
 #include <Base/Tools.h>
 #include <App/Application.h>
-#include <App/OriginFeature.h>
-#include <App/Document.h>
 #include <Mod/Part/App/FaceMakerCheese.h>
-#include "FeatureSketchBased.h"
-#include "DatumPlane.h"
 #include "DatumLine.h"
+#include "DatumPlane.h"
+#include "FeatureSketchBased.h"
 
 using namespace PartDesign;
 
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6381">pipeline status</a> for feature branch [PR_6381](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6381). Pipeline [471998063](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/471998063) was triggered at [56961f6](https://github.com/FreeCAD/FreeCAD/pull/6381/commits/56961f6365ba960b0ba5afa60713f0b6b3f19f62). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #6389, opened 2022-02-17
[Sketcher] Fix some issues when b-spline knot points coincide with other points

AjinkyaDahale:
There are certain issues that happen when points corresponding to knots of a b-spline coincide with other points, particularly the end points of the curve or control points of the spline. This PR attempts to fix them.

The commit currently in is ready to be merged but I want to add another, which would make sure the control points that can only be selected as knots can also be moved.

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -160,9 +160,9 @@ void showNoConstraintBetweenFixedGeometry()
 {
     QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Wrong selection"),
                          QObject::tr("Cannot add a constraint between two fixed geometries. "
-                                     "Fixed geometries involve external geometry, "
-                                     "blocked geometry or special points "
-                                     "as B-spline knot points."));
+                                     "Fixed geometries include external geometry, "
+                                     "blocked geometry, and special points "
+                                     "such as B-spline knot points."));
 }
 
 /// Makes a simple tangency constraint using extra point + tangent via point
@@ -529,7 +529,7 @@ int SketchSelection::setUp(void)
         // if one selectetd, only sketch allowed. should be done by activity of command
         if(!selection[0].getObject()->getTypeId().isDerivedFrom(Sketcher::SketchObject::getClassTypeId()))
         {
-            ErrorMsg = QObject::tr("Only sketch and its support is allowed to select");
+            ErrorMsg = QObject::tr("Only sketch and its support are allowed to be selected.");
             return -1;
         }
 
@@ -540,7 +540,7 @@ int SketchSelection::setUp(void)
             SketchObj = static_cast<Sketcher::SketchObject*>(selection[0].getObject());
             // check if the none sketch object is the support of the sketch
             if(selection[1].getObject() != SketchObj->Support.getValue()){
-                ErrorMsg = QObject::tr("Only sketch and its support is allowed to select");
+                ErrorMsg = QObject::tr("Only sketch and its support are allowed to be selected.");
                 return-1;
             }
             // assume always a Part::Feature derived object as support
@@ -552,7 +552,7 @@ int SketchSelection::setUp(void)
             SketchObj = static_cast<Sketcher::SketchObject*>(selection[1].getObject());
             // check if the none sketch object is the support of the sketch
             if(selection[0].getObject() != SketchObj->Support.getValue()){
-                ErrorMsg = QObject::tr("Only sketch and its support is allowed to select");
+                ErrorMsg = QObject::tr("Only sketch and its support are allowed to be selected.");
                 return -1;
             }
             // assume always a Part::Feature derived object as support
@@ -561,7 +561,7 @@ int SketchSelection::setUp(void)
             SupportSubNames = selection[0].getSubNames();
 
         } else {
-            ErrorMsg = QObject::tr("One of the selected has to be on the sketch");
+            ErrorMsg = QObject::tr("One of the selected has to be on the sketch.");
             return -1;
         }
     }
@@ -936,7 +936,7 @@ void CmdSketcherConstrainHorizontal::activated(int iMsg)
             const Part::Geometry *geo = Obj->getGeometry(GeoId);
             if (geo->getTypeId() != Part::GeomLineSegment::getClassTypeId()) {
                 QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Impossible constraint"),
-                                     QObject::tr("The selected edge is not a line segment"));
+                                     QObject::tr("The selected edge is not a line segment."));
                 return;
             }
 
@@ -1003,7 +1003,7 @@ void CmdSketcherConstrainHorizontal::activated(int iMsg)
     }
     else { // vertex mode, fixedpoints > 1
         QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Impossible constraint"),
-                             QObject::tr("There are more than one fixed point selected. Select a maximum of one fixed point!"));
+                             QObject::tr("There are more than one fixed points selected. Select a maximum of one fixed point!"));
         return;
     }
     // finish the transaction and update
@@ -1031,7 +1031,7 @@ void CmdSketcherConstrainHorizontal::applyConstraint(std::vector<SelIdPair> &sel
                 const Part::Geometry *geo = Obj->getGeometry(CrvId);
                 if (geo->getTypeId() != Part::GeomLineSegment::getClassTypeId()) {
                     QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Impossible constraint"),
-                                        QObject::tr("The selected edge is not a line segment"));
+                                        QObject::tr("The selected edge is not a line segment."));
                     return;
                 }
 
@@ -1174,7 +1174,7 @@ void CmdSketcherConstrainVertical::activated(int iMsg)
             const Part::Geometry *geo = Obj->getGeometry(GeoId);
             if (geo->getTypeId() != Part::GeomLineSegment::getClassTypeId()) {
                 QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Impossible constraint"),
-                                     QObject::tr("The selected edge is not a line segment"));
+                                     QObject::tr("The selected edge is not a line segment."));
                 return;
             }
 
@@ -1267,7 +1267,7 @@ void CmdSketcherConstrainVertical::applyConstraint(std::vector<SelIdPair> &selSe
             const Part::Geometry *geo = Obj->getGeometry(CrvId);
             if (geo->getTypeId() != Part::GeomLineSegment::getClassTypeId()) {
                 QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Impossible constraint"),
-                                     QObject::tr("The selected edge is not a line segment"));
+                                     QObject::tr("The selected edge is not a line segment."));
                 return;
             }
 
@@ -1978,7 +1978,7 @@ void CmdSketcherConstrainCoincident::activated(int iMsg)
 
     if (SubNames.size() < 2) {
         QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Wrong selection"),
-            QObject::tr("Select two or more vertexes from the sketch."));
+            QObject::tr("Select two or more vertices from the sketch."));
         return;
     }
 
@@ -1988,7 +1988,7 @@ void CmdSketcherConstrainCoincident::activated(int iMsg)
         getIdsFromName(*it, Obj, GeoId, PosId);
         if (isEdge(GeoId,PosId)) {
             QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Wrong selection"),
-                QObject::tr("Select two or more vertexes from the sketch."));
+                QObject::tr("Select two or more vertices from the sketch."));
             return;
         }
     }
@@ -2134,7 +2134,7 @@ void CmdSketcherConstrainDistance::activated(int iMsg)
         }
         else {
             QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Wrong selection"),
-                                 QObject::tr("Select vertexes from the sketch."));
+                                 QObject::tr("Select vertices from the sketch."));
         }
         return;
     }
@@ -2368,7 +2368,7 @@ void CmdSketcherConstrainDistance::applyConstraint(std::vector<SelIdPair> &selSe
         }
         else {
             QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Wrong selection"),
-                QObject::tr("This constraint does not make sense for non-linear curves"));
+                QObject::tr("This constraint does not make sense for non-linear curves."));
         }
 
         return;
@@ -2559,7 +2559,7 @@ void CmdSketcherConstrainPointOnObject::activated(int iMsg)
 
                 if( geom && isBsplinePole(geom)) {
                     QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Wrong selection"),
-                                        QObject::tr("Select an edge that is not a B-spline weight"));
+                                        QObject::tr("Select an edge that is not a B-spline weight."));
                     abortCommand();
 
                     continue;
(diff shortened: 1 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6389">pipeline status</a> for feature branch [PR_6389](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6389). Pipeline [473508955](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/473508955) was triggered at [634817b](https://github.com/FreeCAD/FreeCAD/pull/6389/commits/634817b53f8c8f2b6162741119603313596eca6b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6389">pipeline status</a> for feature branch [PR_6389](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6389). Pipeline [474146647](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/474146647) was triggered at [c40e766](https://github.com/FreeCAD/FreeCAD/pull/6389/commits/c40e76682c5566f2af3855e164d6027995a98d98). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/CommandSketcherBSpline.cpp
```diff
@@ -646,45 +646,72 @@ void CmdSketcherIncreaseKnotMultiplicity::activated(int iMsg)
...
     if (isSimpleVertex(Obj, GeoId, PosId)) {
-        const std::vector<Sketcher::Constraint *> &vals = Obj->Constraints.getValues();
+        int splineGeoId;
+        int knotIndexOCC;
 
-        for (std::vector<Sketcher::Constraint *>::const_iterator it= vals.begin(); it != vals.end(); ++it) {
-            if ((*it)->Type == Sketcher::InternalAlignment
-                && (*it)->First == GeoId
-                && (*it)->AlignmentType == Sketcher::BSplineKnotPoint)
+        for (auto const constraint : Obj->Constraints.getValues())
```
abdullahtahiriyo, line 652: Please add braces to this for.

We omit braces generally only when it is a single line within the scope of the for. This is way longer and it is difficult to read where it is intended to end.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -484,7 +484,7 @@ void SketcherGui::notifyConstraintSubstitutions(const QString & message)
 {
     Gui::Dialog::DlgCheckableMessageBox::showMessage(   QObject::tr("Sketcher Constraint Substitution"),
                                                         message,
-                                                        QLatin1String("User parameter:BaseApp/Preferences/Mod/Sketcher/General"),
+                                                        QLatin1String("User parameter: BaseApp/Preferences/Mod/Sketcher/General"),
```
abdullahtahiriyo, line 487: Not sure this will work with the space. I will check later. Just so that I do not forget.

src/Mod/Sketcher/Gui/Utils.cpp
```diff
@@ -209,6 +209,9 @@ bool SketcherGui::isSimpleVertex(const Sketcher::SketchObject* Obj, int GeoId, P
     const Part::Geometry *geo = Obj->getGeometry(GeoId);
     if (geo->getTypeId() == Part::GeomPoint::getClassTypeId())
         return true;
+    else if (geo->getTypeId() == Part::GeomBSplineCurve::getClassTypeId() &&
```
abdullahtahiriyo, line 212: I do not quite understand this.

IsSimpleVertex (the name may be misleading), refers to an unconnected point.

For example an unconnected point cannot be used to apply a tangency point-to-point or a perpendicular point-to-point. B-Splines admit this type of constraints (without this commit). Select one endpoint of an arc-of-circle one endpoint of a bspline and select the perpendicular or the tangency constraint. I have not compiled this PR yet, but unless I am missing something, this functionality will be gone.

Do I make sense?

abdullahtahiriyo (commented):
I am not sure that it is possible to modify the multiplicity of an endpoint of a non-periodic B-Spline. I think I remember that this will not be supported by GCS. I have to look into the code to ensure this is accurate and remember why there was this limitation.

If the limitation can be lifted, then this PR is ok, it just applies the multiplicity modification to an endpoint.

EDIT: Ok. Now I understood you are limiting the functionality to periodic b-splines. There is no GCS limitation for periodic b-splines. So this PR is just fine (when the outstanding issue is resolved).

src/Mod/Sketcher/Gui/Utils.cpp
AjinkyaDahale: > IsSimpleVertex (the name may be misleading), refers to an unconnected point.
> 
> For example an unconnected point cannot be used to apply a tangency point-to-point or a perpendicular point-to-point.

I see. If I understand this correctly, this is what is happening:
- `isSimpleVertex` is created to reject "simple vertices" from being used in tangent/perpendicular constraints;
- knots are visualized as "simple vertices" that lie on the B-Spline;
- `isSimpleVertex` is used to accept knots in `CmdSketcherIncrease/DecreaseKnotMultiplicity::activated()`, but end points of B-splines are not accepted. 

This doesn't cause any issues in non-periodic splines since we can't practically change multiplicity of the end (i.e. first and last) knots (it is already at the highest and OCC does not allow us to decrease it). However, knot multiplicity of first and last knots in periodic spline (which are by it's nature the same) can be changed, and that is not reflected in Sketcher.

I might have to use a different function for what I intend to do in "Increase/Decrease Knot Multiplicity" command. Perhaps more appropriate would be to use `isBSplineKnot` (and make changes there to ensure end points of splines are also treated as knots). 

> B-Splines admit this type of constraints (without this commit). Select one endpoint of an arc-of-circle one endpoint of a bspline and select the perpendicular or the tangency constraint. I have not compiled this PR yet, but unless I am missing something, this functionality will be gone.
> 
> Do I make sense?

I am a little confused about how _something we cannot do with unconnected points_ relates to _something we can do with b-splines_.

abdullahtahiriyo: Ok. I understand your point.

I agree that knots are non-connected points (at least within the meaning of the sketcher, one could argue they are connected points in the same that they connect the piecewise parts of the B-Spline). However, it is my understanding that start and endpoint are not knots. I would say that they are coincident with knots. Endpoints do not have a InternalAlignment constraint of type knot or a extension attribute (SketchGeometryExtension, GeometryFacade) of type knot.

If you want to extend the change of multiplicity to endpoints, I see no problem in using them, as this is what the user will select, but under the understanding that we are not selecting knots, but a point coincident with them. Then we are taking this information to apply the operation to the right knot.

In this understanding, I think it is acceptable to use (to reject): isSimpleVertex || (bspline && ( start || end) ), which you may well encapsulate it in a free function (isKnotOrBSplineEnd). I would not change isBSplineKnot because of what is explained above.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6389">pipeline status</a> for feature branch [PR_6389](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6389). Pipeline [474609640](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/474609640) was triggered at [d404dab](https://github.com/FreeCAD/FreeCAD/pull/6389/commits/d404dab1e63d9939f08e90be4ae0e8e1272aa293). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/CommandConstraints.cpp
abdullahtahiriyo: It does not work.

Reproduce:
1. In sketcher settings uncheck the "Notify automatic constraint substitution" (you do not want to be notified).
2. Make two circle arcs
2. Make coincident between two endpoints of the arcs
3. Apply a tangency constraint to the edges of the two arcs
4. You will see the pop-up (which is the default behaviour if the parameter is not set/found)

src/Mod/Sketcher/Gui/Utils.cpp
abdullahtahiriyo: > I would suggest we try to add said constraint, but the endpoint can be common to multiple curves and that is a question we would want to address in the future (especially when OCC allows decreasing multiplicity of non-periodic end knots).

Ah, I see. We do have knots that are knots exactly at that position:
Peek 18-02-2022 20-52

I have deleted the poles at the endpoints so that it becomes clearer. There are three "simpleVertices". Two are coincident with the curve endpoints, but are rendered under the endpoints, so they are never picked. These knots do have the constraint and the extension.

You could theoretically select the knot point in the elements widget and trigger the tool (that won't work because it is unsupported by OCC, and currently also by GCS, the endpoint tangency and perpendicular will break, together with all models making use of them, as the current implementation at GCS is based on the property of non-periodic BSplines with degree multiplicity at the endpoints, that the pole is coincident with the endpoint).

The problem with coincident points, I think it is a general problem of selection which requires a solution for enhancing selection. I am unsure if RT had an improvement in his branch for this...

AjinkyaDahale: > We do have knots that are knots exactly at that position: 

That is odd. I used `gdb` to run `getCoincidentPoints` (or whatever the appropriate function is called), but it returned empty. That was my first idea.

> You could theoretically select the knot point in the elements widget and trigger the tool (that won't work because it is unsupported by OCC, and currently also by GCS, the endpoint tangency and perpendicular will break, together with all models making use of them, as the current implementation at GCS is based on the property of non-periodic BSplines with degree multiplicity at the endpoints, that the pole is coincident with the endpoint).

If you want I can add another check for non-periodic b-spline that shows a Qt message saying this is not supported.

> The problem with coincident points, I think it is a general problem of selection which requires a solution for enhancing selection. I am unsure if RT had an improvement in his branch for this...

If that exists my work is kind of moot :grin:

AjinkyaDahale: @realthunder in your branch is there a fix/workaround for sketcher cases where certain points (possibly different type) are inherently coincident on one another (like a bspline knot, pole and/ or end point), and an operation can only happen on one type? For example we can only increase multiplicity of knots but end points are selected.

AjinkyaDahale: > The problem with coincident points, I think it is a general problem of selection which requires a solution for enhancing selection.

Do you think we could filter out certain kinds of points by making a `DrawSketchHandler` for these tools, especially after #5398?

realthunder: Yes, there is a general solution in my branch to pick coincident geometry, even constraints. I actually attemtped a PR for that a few days ago because some user requested it. But it requires a larger change then I expected, many of them are in the core actually. In addition, because of the recent huge refactor of ViewProviderSketch, those changes are not well tested. So I decided to wait. If you guys really need it now, I can try again.

AjinkyaDahale: @realthunder it's not particularly urgent but good to have. This PR set out to do the exact thing but on a case by case basis. If your implementation is general it'll save quite a bit of duplication of effort.

realthunder: Maybe you can first give a try using my [release](https://github.com/realthunder/FreeCAD_assembly3/releases). If you are on Linux, you can also try the more recent [snap build](https://snapcraft.io/freecad-realthunder). And see if it covers your use case. Just use mouse move to highlight the geometry or constraint, and presse 'G, G' shortchut to bring up the geometry picker menu.

AjinkyaDahale: > Maybe you can first give a try using my [release](https://github.com/realthunder/FreeCAD_assembly3/releases).

Thanks for the suggestion. I tried it and it does seem like what I need. There are some other details though:
1. There is no information of the kind of vertex (or any other entity), i.e. is it an end point, control point or knot.
2. If something is selected and a constraint is applied, We need it to make a better decision about that. Like knot multiplicity change will only happen on knot points, and we can only drag a high degree knot because it is coincident on a control point.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6389">pipeline status</a> for feature branch [PR_6389](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6389). Pipeline [474829594](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/474829594) was triggered at [ed06eed](https://github.com/FreeCAD/FreeCAD/pull/6389/commits/ed06eed0a8f860d19c65ba7746a0565ebe0cbf25). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/Utils.cpp
abdullahtahiriyo: @AjinkyaDahale 



> Do you think we could filter out certain kinds of points by making a `DrawSketchHandler` for these tools, especially after #5398?

Not without a substantial amount of effort. I am not aware of a way to apply a pickstyle to parts of a pointset in coin. Rather, the solution may be to get everything crossing the pick ray and allow the user to select what he means.

At DrawSketchHandler level, you could make a gate filter, like many DrawSketchHandlers use. However that would prevent from selecting certain points, without enabling you to pick points that are under.

If RT has a solution, then it makes sense to see if it can be implemented. However, we are close to a feature freeze, and I trust RT's impact assessment. If he is of the opinion that it involves substantial core changes, then it may be reasonable to wait until the next release cycle.

AjinkyaDahale: > If RT has a solution, then it makes sense to see if it can be implemented. However, we are close to a feature freeze, and I trust RT's impact assessment. If he is of the opinion that it involves substantial core changes, then it may be reasonable to wait until the next release cycle.

In that case there may still be point in making the workarounds that I intend to.

realthunder: > 1. There is no information of the kind of vertex (or any other entity), i.e. is it an end point, control point or knot.

@AjinkyaDahale SketchObject can override new API [getHigherElements()](https://github.com/realthunder/FreeCAD/blob/8f08caed351bcf2e7c621bd834b931693e92c72c/src/Mod/Sketcher/App/SketchObject.cpp#L9508) to add new entries to the `Picked geometry` menu. The existing implemention checks if the picked vertex corresponds to a coincidence constraint to allow user to select those constraints instead. Same thing can be done for knots I think.

> 2\. If something is selected and a constraint is applied, We need it to make a better decision about that. Like knot multiplicity change will only happen on knot points, and we can only drag a high degree knot because it is coincident on a control point.

The above function makes it possible to select new types of elements, e.g. Knot1, you can then implement your command to check for that in selection notification.

> If he is of the opinion that it involves substantial core changes, then it may be reasonable to wait until the next release cycle.

Yeah, it does require quite some core changes. I'd suggest to wait.

AjinkyaDahale:
> EDIT: Ok. Now I understood you are limiting the functionality to periodic b-splines. There is no GCS limitation for periodic b-splines. So this PR is just fine (when the outstanding issue is resolved).

It is not that "I am limiting the functionality to periodic b-splines" as much as "OCC doesn't allow this functionality on non-periodic b-splines". If in the future OCC changes this (which is possible since it isn't explicitly mentioned in the docs, even in [`Geom_BSplineCurve::RemoveKnot`](https://dev.opencascade.org/doc/refman/html/class_geom___b_spline_curve.html#a4d22cafccfeb8b12cfaf5e9c3d4e6c3d)), we will need to explicitly restrict this possibility on our end or support it in GCS.

abdullahtahiriyo:
@AjinkyaDahale 

Yes, what I meant is that the functionality is only brought to periodic b-splines.

PR #6393, opened 2022-02-18
Sketch: Fix algorithm seeking for missing coincidences in Validate

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=8&t=62936

### DIFF
--- a/src/Mod/Sketcher/App/SketchAnalysis.cpp
+++ b/src/Mod/Sketcher/App/SketchAnalysis.cpp
@@ -87,6 +87,15 @@ private:
     double tolerance;
 };
 
+struct SketchAnalysis::VertexID_Less
+{
+    bool operator()(const VertexIds& x,
+                    const VertexIds& y) const
+    {
+        return (x.GeoId < y.GeoId || ((x.GeoId == y.GeoId) && (x.PosId < y. PosId)));
+    }
+};
+
 struct SketchAnalysis::Vertex_EqualTo
 {
     Vertex_EqualTo(double tolerance) : tolerance(tolerance){}
@@ -142,7 +151,9 @@ private:
 
 int SketchAnalysis::detectMissingPointOnPointConstraints(double precision, bool includeconstruction /*=true*/)
 {
-    std::vector<VertexIds> vertexIds;
+    std::vector<VertexIds> vertexIds; // Holds a list of all vertices in the sketch
+
+    // Build the list of sketch vertices
     const std::vector<Part::Geometry *>& geom = sketch->getInternalGeometry();
     for (std::size_t i=0; i<geom.size(); i++) {
         auto gf = GeometryFacade::getFacade(geom[i]);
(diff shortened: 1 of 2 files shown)


src/Mod/Sketcher/App/SketchAnalysis.cpp
```diff
@@ -87,6 +87,15 @@ struct SketchAnalysis::Vertex_Less
     double tolerance;
 };
 
+struct SketchAnalysis::VertexID_Less
+{
+    bool operator()(const VertexIds& x,
+                    const VertexIds& y) const
+    {
+        return (x.GeoId < y.GeoId || ((x.GeoId == y.GeoId) && (x.PosId < y. PosId)));
```
AjinkyaDahale, line 95: Isn't PosId now an enum class?

0penBrain: @AjinkyaDahale yes it is

0penBrain: Not a problem. 'LESS' comparing functions should just ensure a transitive ordering in all cases, and that when 2 equal objects are compared, it returns false in both argument orders. ;)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6393">pipeline status</a> for feature branch [PR_6393](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6393). Pipeline [474829595](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/474829595) was triggered at [d20c44b](https://github.com/FreeCAD/FreeCAD/pull/6393/commits/d20c44b1631ea256380bab53a325cf9ff3783983). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/App/SketchAnalysis.cpp
```diff
@@ -222,67 +233,120 @@ int SketchAnalysis::detectMissingPointOnPointConstraints(double precision, bool
...
+                v1.GeoId = coincidence->First;
+                v1.PosId = coincidence->FirstPos;
+                v2.GeoId = coincidence->Second;
+                v2.PosId = coincidence->SecondPos;
+
+                // Look if coincident vertices are in the group of adjacent ones we are processing
+                auto nv1 = vertexGrp.extract(v1);
+                auto nv2 = vertexGrp.extract(v2);
+                if (nv1.empty() && nv2.empty()) continue;
+
```
wwmayer, line 292: At this point nv1 or nv2 could be empty but not both

wwmayer:
The PR causes a crash when using chrisb's sketch example. For more details see above comments.

src/Mod/Sketcher/App/SketchAnalysis.cpp
0penBrain: Actually both can be empty. ;) I have a test file for that. But the problem is that continuing isn't the right solution in all cases. :| I have to polish this PR.

src/Mod/Sketcher/App/SketchAnalysis.cpp
```diff
@@ -222,67 +233,120 @@ int SketchAnalysis::detectMissingPointOnPointConstraints(double precision, bool
...
+                // Look if one of the constrained vertices is already in a group of coincident vertices
+                for (std::set<VertexIds, VertexID_Less> grp:coincVertexGrps) {
+                    if ( (grp.find(v1) != grp.end()) || (grp.find(v2) != grp.end()) ) {
+                        // If yes add them to the existing group
+                        grp.insert(nv1.value());
+                        grp.insert(nv2.value());
+                        continue;
```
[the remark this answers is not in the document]
0penBrain: I'd expect using nodes (since C++17) would offer some tolerance. However it's a bit of a guess. Will check with debugger and fix if needed.

0penBrain:
> The PR causes a crash when using chrisb's sketch example. For more details see above comments.

Actually the crash comes from the VertexID_Less comparator (which probably is called with an empty thing). Debugger is warming up. Thx for review.

src/Mod/Sketcher/App/SketchAnalysis.cpp
0penBrain: Actually, when both empty, in no cases it is good idea to continue. We should always check. It's fixed now.

src/Mod/Sketcher/App/SketchAnalysis.cpp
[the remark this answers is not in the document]
0penBrain: You were very right. The insert needs the comparator (obviously in a set), which generously crashes when a value is empty. :) Also fixed.

0penBrain:
@wwmayer Should everything be OK now. I tested on more test cases (actually I missed the simplest ones, which actually were the most significant).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6393">pipeline status</a> for feature branch [PR_6393](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6393). Pipeline [480439209](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/480439209) was triggered at [1b65c81](https://github.com/FreeCAD/FreeCAD/pull/6393/commits/1b65c81881ec787fb6da05c18d537994f7d5d1bc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/App/SketchAnalysis.cpp
wwmayer: > Actually both can be empty. ;)

Sure, but in your first version you had a `continue;` in the loop when nv1 and nv2 were empty. So, after that line only nv1 or nv2 could be empty but not both and this fact wasn't checked in the further code.

wwmayer:
> Let you edit the title of the PR when you think it's right. ;)

I will have a new look in the next couple of days.

wwmayer:
Works as expected now. Here is the mandatory unit test to see that current master will fail: 6e4c61881b

0penBrain:
> Here is the mandatory unit test to see that current master will fail: [6e4c618](https://github.com/FreeCAD/FreeCAD/commit/6e4c61881be3ba23f3796682f9e30d4b03c4de8b)

Please next time just let me know that a test is expected. Actually I didn't add it because I (wrongly) was pretty sure that this function can't be called from Python. I'll add another one that allowed me to find another flaw in my algorithm. ;)

wwmayer:
> Please next time just let me know that a test is expected.

A unit test is basically always expected. Although Qt has a test module to test GUI intersections I don't expect a function for this kind of stuff.

> Actually I didn't add it because I (wrongly) was pretty sure that this function can't be called from Python.

I didn't know about the exposed function either and so I simply searched in the code base. If there weren't one I would have written a wrapper for it.

PR #6408, opened 2022-02-21
Addon manager: Auto-add Macro toolbar button 

chennes:
The main work of this commit is to add a workflow following the installation of a new macro that asks if you would like to add this macro to a toolbar. The exact behavior will depend on whether you already have custom toolbars or not: if you do, then you'll be prompted as to whether you want to install the button in one of them, or to create a new toolbar. If you don't have any custom toolbars this is skipped, and one is automatically created. If the macro provides an icon, that is used in the toolbar. If not, the current behavior is to display the name of the macro. I am evaluating whether auto-creation of an icon is feasible/desirable, and welcome your comments in this regard.

Several other minor changes to macro installation behavior are also implemented in support of this new feature: most notably, two additional methods for supplying the macros' icon are implemented. First, it was noted that several macros attempted to use an online URL for their icon, even though this was not supported. Now, if an http or https link is provided for the `__icon__` field, that file is downloaded and used as the macro's icon both in the Addon Manager and when installed to a toolbar. Second, if either `__icon__` or `__xpm__` contain a triple-quoted string, that string is treated as XPM data and written into a *.xpm file. If the data results in a valid XPM-formatted icon, that icon is then used both in the Addon Manager and in toolbars as the icon for that macro. Note that this data **must** be triple-quoted, and triple quotes are explicitly disallowed for any other metadata fields.

A few other minor fixes include macros that specified their version as a reference to the date, e.g.:
```
__version__ = __date__
```
This was previously unsupported, as the metadata parsing does not actually execute the macro and this `__date__` was meaningless. This particular idiom was relatively common, however, so the specific use-case was added. Note that in general it is still required that all metadata fields be string literals, this case is handled as a special exception.

Other macros specify the version as a number, rather than a string literal. Again, this was not previously supported, but in this particular case if a number is detected, it is converted to a string and stored as the version.

TESTERS: Please backup your user.cfg file prior to testing this PR. If something is wrong it will probably mess up your custom toolbars.

### DIFF
--- a/src/Gui/CommandPy.xml
+++ b/src/Gui/CommandPy.xml
@@ -117,5 +117,15 @@ Returns True if something was removed, or False if not.
         </UserDocu>
       </Documentation>
     </Methode>
+    <Methode Name="findCustomCommand" Static='true'>
+      <Documentation>
+        <UserDocu>Find the name of a custom command, given a macro name
+findCustomCommand(name) -> Optional[str]
+--
+Given the name of a macro, return the name of the custom command for that macro, or
+None if there is no command matching that macro script name.
+        </UserDocu>
+      </Documentation>
+    </Methode>
   </PythonExport>
 </GenerateModel>
--- a/src/Gui/CommandPyImp.cpp
+++ b/src/Gui/CommandPyImp.cpp
@@ -384,6 +384,31 @@ PyObject* CommandPy::removeCustomCommand(PyObject* args)
     }
 }
 
+PyObject* CommandPy::findCustomCommand(PyObject* args)
+{
+    const char* macroScriptName = nullptr;
+    if (!PyArg_ParseTuple(args, "s", &macroScriptName))
+        return nullptr;
+
+    CommandManager& commandManager = Application::Instance->commandManager();
+    std::vector<Command*> macros = commandManager.getGroupCommands("Macros");
+
+    auto action = std::find_if(macros.begin(), macros.end(), [macroScriptName](const Command* c) {
+        if (auto mc = dynamic_cast<const MacroCommand*>(c))
+            if (std::string(mc->getScriptName()) == std::string(macroScriptName))
+                return true;
+        return false;
+        });
+
+    if (action != macros.end()) {
+        return PyUnicode_FromString((*action)->getName());
+    }
+    else {
+        Py_INCREF(Py_None);
+        return Py_None;
+    }
+}
+
 PyObject *CommandPy::getCustomAttributes(const char* /*attr*/) const
 {
     return 0;
--- a/src/Mod/AddonManager/AddonManager.py
+++ b/src/Mod/AddonManager/AddonManager.py
@@ -41,6 +41,10 @@ import AddonManager_rc
 from package_list import PackageList, PackageListItemModel
 from package_details import PackageDetails
 from AddonManagerRepo import AddonManagerRepo
+from install_to_toolbar import (
+    ask_to_install_toolbar_button,
+    remove_custom_toolbar_button,
+)
 
 from NetworkManager import HAVE_QTNETWORK, InitializeNetworkManager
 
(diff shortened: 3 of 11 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6408">pipeline status</a> for feature branch [PR_6408](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6408). Pipeline [475439146](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/475439146) was triggered at [5fc9a19](https://github.com/FreeCAD/FreeCAD/pull/6408/commits/5fc9a1944395e0a5c9a2e7e9ed342e76be91dd8c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

kisolre:
I tried Bevel macro. It got properly installed, selected toolbar entry was created but the icon was not installed/used/...? Even after restart. When the macro was uninstalled toolbar entry was also removed.

Word size of FreeCAD: 64-bit
Version: 0.20.27679 +2 (Git)
Build type: Release
Branch: (HEAD detached at chennes/addonManagerMacroButton)
Hash: 5fc9a1944395e0a5c9a2e7e9ed342e76be91dd8c
Python 3.8.8, Qt 5.15.2, Coin 4.0.1, OCC 7.5.3
Locale: Bulgarian/Bulgaria (bg_BG)

chennes:
The Bevel macro does not appear to specify an icon -- is there a version of it someplace that does?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6408">pipeline status</a> for feature branch [PR_6408](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6408). Pipeline [476342489](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/476342489) was triggered at [7154908](https://github.com/FreeCAD/FreeCAD/pull/6408/commits/7154908d32788084231ac18f5862466196b6a5c1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

mwganson:
Which macro has some example code for defining __xpm__ as a triple quote string or from a url?  I could host the xpm file on github and link to that, I suppose, or upload to the wiki.

chennes:
I just coded an XPM locally for testing, no one has implemented such a thing yet. For the URL-based icons, check out:

- CloneConvert
- Airfoil Import & Scale
- CenterOfMass
- Compound Plus
- Fonts Win10 PYMP

chennes:
Here's the XPM data I was using:
```
__xpm__    = """/* XPM */
static char *_437320bbc4c47d0888b3b4e40649215Bhhxcm4IoeuC3njx[] = {
/* columns rows colors chars-per-pixel */
"64 64 2 1 ",
"  c None",
". c red",
/* pixels */
"                                                                ",
"                                                                ",
"                                                                ",
"                                                                ",
"                                                                ",
"                                                                ",
"                                    ..........                  ",
"                              ......................            ",
"                           ............................         ",
"                        ................................        ",
"                      .................................         ",
"                    ...................................         ",
"                   ....................................         ",
"                  .....................................         ",
"                 .....................................          ",
"               .......................................          ",
"               .......................................          ",
"              ........................................          ",
"             ........................        ........           ",
"            ......................                ...           ",
"            ...................                                 ",
"           ...................                                  ",
"           ..................                                   ",
"          ..................                                    ",
"          .................                                     ",
"         .................                                      ",
"         .................                                      ",
"         ................                                       ",
"         ................                                       ",
"         ...............                                        ",
"        ................                                        ",
"        ................                                        ",
"        ................                                        ",
"        ................                                        ",
"        ................                                        ",
"        ................                                        ",
"        ................                                        ",
"        ................                                        ",
"        .................                                       ",
"        .................                                       ",
"         ................                                       ",
"         .................                                      ",
"         .................                                      ",
"         ..................                                     ",
"          ..................                                    ",
"          ...................                                   ",
"           ...................                                  ",
"           .....................                                ",
"            ......................                 ..           ",
"            ..........................       ........           ",
"             ........................................           ",
"              .......................................           ",
"              ........................................          ",
"               .......................................          ",
"                ......................................          ",
"                  ....................................          ",
"                   ...................................          ",
"                     ..................................         ",
"                       ................................         ",
"                         ..............................         ",
"                            ........................            ",
"                                  ............                  ",
"                                                                ",
"                                                                "
};"""
```

luzpaz:
pardon my ignorance, isn't XPM not ideal in the UI because of HiDPI. Instead the solution is to use SVG (which scales better)?

chennes:
XPM is not a great format: but it's a *simple* format, which counts for a lot. I could add the same idea to support inline SVGs, but SVGs aren't usually so compact. If you want an SVG, it's best to host it online someplace and put the URL in __icon__. Besides, a 64x64 XPM looks OK at the icon size we are using in the toolbar, even in HiDPI.

mwganson:
Bevel uses xpm, but it's complicated.  It creates the .py file from the .FCMacro text, which is commented out in the FCMacro file.  In short, the FCMacro file parses itself, strips off the # symbols, and writes the text to a new file, the .py file, on first run.  I do it this way to allow for distribution on the wiki with a single file.  The xpm definition won't show up if you parse the .FCMacro file because it's commented out.  This is no problem because I can simply define the __xpm__ string and reference it that way or even just duplicate it twice if need be.

With xpm I can have a dynamic toolbar icon by changing the xpm text and signalling the view provider:

fp.ViewObject.signalChangeIcon()

This triggers a new call to view provider getIcon(), which returns the new xpm string.  Parametric Defeaturing makes use of this.  The tree icon gets a red background when the defeaturing has failed.

Snip macro screenshot-9796d9



Would such a dynamic icon be possible with toolbar icons?  Just curious, not a big deal either way.

chennes:
We'd have to add a new Python function to allow custom commands to update their icon: it's not difficult, but I don't think it exists at this time (though @wwmayer may be able to say otherwise).

wwmayer:
> We'd have to add a new Python function to allow custom commands to update their icon: it's not difficult, but I don't think it exists at this time

From any command you can access its QAction object(s) with getAction() and call setIcon() on it/them.
https://github.com/FreeCAD/FreeCAD/blob/master/src/Gui/CommandPyImp.cpp#L294

chennes:
NIce. So then, if I'm understanding correctly, from within your macro you'd do something like:
```
command_name = findCustomCommand("my_macro.FCStd")
if command_name:
    command = FreeCADGui.Command.get(command_name)
    action = command.getAction()
    action.setIcon()
```

wwmayer:
_action_ is a list because a command may use an action group internally.

```
command_name = findCustomCommand("my_macro.FCStd")
if command_name:
    command = FreeCADGui.Command.get(command_name)
    action = command.getAction()
    action[0].setIcon(icon)
```

PR #6412, opened 2022-02-21
Remove empty __init__.py files

donovaly:
I don't know if this can have side effects. Can a Pythionist please have a look?


marioalexis84:
The \_\_init\_\_.py file is necessary for the package to be a _regular package_. Without the \_\_init\_\_.py file it is a _namespace package_.
Namespace packages need not all come from the same directory, while in regular package all parts are  in the same directory hierarchy.
Since these packages are not part of a package namespace, it is preferable to keep the empty \_\_init\_\_.py file.

chennes:
Yes, as @marioalexis84 notes, these files do have special meaning to Python, and affect its treatment of `import` statements. So as strange as it seems, in at least some cases these empty files are necessary. It's possible that some of them are not needed, but that will require a more careful review of each individual case.

donovaly:
Thank you for explanation. Closing this PR

PR #6427, opened 2022-02-22
Gui: Add splash and about image override

chennes:
This PR adds two preference items (currently in BaseApp/Preferences/General, but not exposed in the Gui) that allow a graphic designer working on the splashscreen and about images to override the current resource images, so that they can easily test their work.


chennes:
Oops, made the branch off the wrong head. Closing.

PR #6428, opened 2022-02-22
Gui: Add splash and about image override

chennes:
This PR adds two preference items (currently in BaseApp/Preferences/General, but not exposed in the Gui) that allow a graphic designer working on the splashscreen and about images to override the current resource images, so that they can easily test their work.

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -1449,10 +1449,22 @@ void MainWindow::stopSplasher(void)
 
 QPixmap MainWindow::aboutImage() const
 {
-    // See if we have a custom About screen image set
     QPixmap about_image;
+
+    // For graphic designers, provide a way to override the real About image via
+    // the preferences, so they can test their work before deploying it.
+    ParameterGrp::handle hGrp = App::GetApplication().GetUserParameter().
+    GetGroup("BaseApp")->GetGroup("Preferences")->GetGroup("General");
+    auto imageOverride = hGrp->GetASCII("AboutImageOverride");
+    if (!imageOverride.empty()) {
+        QFileInfo fi(QString::fromStdString(imageOverride));
+        if (fi.isFile() && fi.exists())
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6428">pipeline status</a> for feature branch [PR_6428](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6428). Pipeline [476343787](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/476343787) was triggered at [5846f6c](https://github.com/FreeCAD/FreeCAD/pull/6428/commits/5846f6c273925f7936bb92a9e3357cacc4cfca0a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
The possibility to test custom about images and splash screens already exists. All what needs to be done is putting the files about_image.png and splash_image.png into the directory `App.getUserAppDataDir()+'Gui/images'`

luzpaz:
huh? first i've heard of this  
It should be documented on https://wiki.freecad.org/branding

chennes:
Oh, that's cool -- is that a feature of Qt's resource system, or is it specific to FreeCAD? Does it work for all resources?

wwmayer:
The possibility to use custom splash screens already exists for many years. Might be 10 years or longer.

> Oh, that's cool -- is that a feature of Qt's resource system, or is it specific to FreeCAD? Does it work for all resources?

We use Qt's facility to set up search paths for a prefix as explained here: https://doc.qt.io/qt-5/qdir.html#setSearchPaths
For images the paths are defined here: https://github.com/FreeCAD/FreeCAD/blob/master/src/Gui/Application.cpp#L1982

This is a very nice feature of Qt that we use for many other resources, too.

PR #6440, opened 2022-02-23
Addon Manager: Fix corrupted ZIP downloads

chennes:
The code that followed redirects was resulting in a double-download of the data. This corrects the problem by manually re-queueing a new network request for each redirect, instead of trying to tell the QNetworkAccessManager to follow them automatically.


ryanthara:
Works.

PR #6449, opened 2022-02-25
[App] Expression and Extension: remove unused includes

donovaly:


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6449">pipeline status</a> for feature branch [PR_6449](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6449). Pipeline [479177507](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/479177507) was triggered at [6ae7acf](https://github.com/FreeCAD/FreeCAD/pull/6449/commits/6ae7acf30b71131c0994b48b73a6692358f5cef2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
```
In file included from ../../../src/App/ExtensionContainer.h:30,
                 from ../../../src/App/TransactionalObject.h:27,
                 from ../../../src/App/DocumentObject.h:28,
                 from ../../../src/App/Annotation.h:27,
                 from ../../../src/App/Annotation.cpp:29:
../../../src/App/Extension.h:92:21: error: 'PropertyData' in namespace 'App' does not name a type
   92 |   static const App::PropertyData * extensionGetPropertyDataPtr(void); \
      |                     ^~~~~~~~~~~~
```

src/App/Enumeration.cpp
```diff
@@ -23,8 +23,8 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <cassert>
-# include <cstring>
-# include <cstdlib>
+# include <string>
```
wwmayer, line 26: No need to include string and vector here. It's already included in the header file. cstring and string are totally different things. The former is to offer all C like string-related functions and the later about the std::string class

src/App/Expression.cpp
```diff
@@ -31,35 +31,24 @@
...
-#include <stdio.h>
-#include <stack>
-#include <deque>
-#include <algorithm>
+
 #include "ExpressionParser.h"
-#include <Base/Unit.h>
-#include <App/PropertyUnits.h>
-#include <App/ObjectIdentifier.h>
-#include <boost/math/special_functions/round.hpp>
```
wwmayer, line 61: boost::math::round is used and will cause a build failure

src/App/Extension.h
```diff
@@ -24,11 +24,11 @@
 #ifndef APP_EXTENSION_H
 #define APP_EXTENSION_H
 
-#include "PropertyContainer.h"
```
wwmayer, line 27: Removing PropertyContainer.h causes the build failure.
There is also a cyclic include of Extension.h and ExtensionContainer.h. It should be checked if it can be avoided to include Extension.h in ExtensionContainer.h and instead a forward declaration be used.

wwmayer:
b4fff07d9e

donovaly:
Thanks.
However, I don't know how to improve these PRs since I compiled FC from scratch without PCH and first when it compiled without any errors, I made the PR.

PR #6456, opened 2022-02-26
[App] Feature: remove unused includes

donovaly:

PR #6460, opened 2022-02-26
[App] GeoFeature etc.: remove unused includes

donovaly:


wwmayer:
Squashed with #6456

donovaly:
Many thanks, also for your patience.

Every PR takes me roughly about an hour and I recompile FC with a clean MSVC solution before I make a PR. It is still not clear for me what PRs that compiles fine for me fail with the CI.

PR #6463, opened 2022-02-27
[Sketcher] Some tools for more control over b-splines at creation

AjinkyaDahale:
New tool(s):
- Press "D" to open a dialog to set the degree.


AjinkyaDahale:
@PaddleStroke I intend to use your tool settings for this in the future, when setting degree and weights, say. Is that possible?

PaddleStroke:
> ool settings for this in the future, when setting degree and weights, say. Is that possible?

Yes I think that will be a great way to do it. but it will need to wait until it gets merged :)

PR #6467, opened 2022-02-27
[App] remove dependency on FCGlobal

donovaly:
This PR might be intrusive. I simply gave it a try, compiled FC from scratch and surprisingly I don't get any compilation error but a quicker compilation.

@wwmayer can you please have a look since I recently saw you adding this dependency.

### DIFF
--- a/src/App/Material.h
+++ b/src/App/Material.h
@@ -30,7 +30,7 @@
 
 #include <sstream>
 #include <iomanip>
-#include <FCGlobal.h>
+
 
 namespace App
 {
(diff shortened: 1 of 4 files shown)


wwmayer:
It's not about compilation because the export macros are used to be defined in the PreCompiled.h. So, the compiler never complained about them.

The FCGlobal.h and all the SomethingGlobal.h serve the purpose to avoid that code checkers always highlight the class declarations because they don't know about the export macros and thus think it's an invalid syntax.

This is very annoying when using other IDE like QtCreator.

Since the files are small they don't affect much the build time and thus there is no need to remove them.

luzpaz:
@donovaly  maybe worth adding a note next to each include for other devs to be aware?

donovaly:
Thanks for the explanation.

PR #6475, opened 2022-02-28
fix uninitialized variable

mosfet80:

---

### DIFF
--- a/src/Mod/TechDraw/Gui/CommandExtensionPack.cpp
+++ b/src/Mod/TechDraw/Gui/CommandExtensionPack.cpp
@@ -1446,7 +1446,7 @@ void CmdTechDrawExtensionPositionSectionView::activated(int iMsg) {
             QObject::tr("Selection is empty"));
         return;
     }
-    float xPos, yPos;
+    float xPos=NAN, yPos= NAN;
     TechDraw::DrawViewPart* baseView;
     auto objFeat = selection[0].getObject();
     if (objFeat->isDerivedFrom(TechDraw::DrawViewSection::getClassTypeId())) {


src/Mod/TechDraw/Gui/CommandExtensionPack.cpp
```diff
@@ -1446,7 +1446,7 @@ void CmdTechDrawExtensionPositionSectionView::activated(int iMsg) {
             QObject::tr("Selection is empty"));
         return;
     }
-    float xPos, yPos;
+    float xPos=NAN, yPos= NAN;
```
chennes, line 1449: If we're going to initialize these to NaN, shouldn't we also check for it before assigning, below? It looks like there is indeed a branch where it's possible they aren't set, and I'm not sure it's a good idea to let this code deliver NaN as a result for the sectionView position.

WandererFan:
setting to NaN will fix the "uninitialized" warning, but a better solution would be: 
```
 auto objFeat = selection[0].getObject();
    if (objFeat->isDerivedFrom(TechDraw::DrawViewSection::getClassTypeId())) {
        TechDraw::DrawViewSection* sectionView = dynamic_cast<TechDraw::DrawViewSection*>(objFeat);
        TechDraw::DrawViewPart* baseView = sectionView->getBaseDVP();
        double xPos = baseView->X.getValue();
        double yPos = baseView->Y.getValue();
        if (baseView->isDerivedFrom(TechDraw::DrawProjGroupItem::getClassTypeId())) {
            std::vector<App::DocumentObject*> parentViews = baseView->getInList();
            if (!parentViews.empty()) {
                TechDraw::DrawProjGroup* groupBase = dynamic_cast<TechDraw::DrawProjGroup*>(parentViews[0]);
                xPos = groupBase->X.getValue();
                yPos = groupBase->Y.getValue();
            }
        }
        std::string direction = sectionView->SectionDirection.getValueAsString();
        if ((direction == "Right") || (direction == "Left"))
            sectionView->Y.setValue(yPos);
        else if ((direction == "Up") || (direction == "Down"))
            sectionView->X.setValue(xPos);
    }
```

src/Mod/TechDraw/Gui/CommandExtensionPack.cpp
wwmayer: I also think it's better to initialize them with 0.0 instead.

wwmayer:
90659c131c6

wwmayer:
> setting to NaN will fix the "uninitialized" warning, but a better solution would be:

There are a few issues with the code block (besides the initialization):
* the dynamic_cast of objFeat is not needed here because it's already checked that it's a DrawViewSection
* the dynamic_cast of `parentViews[0]` is fine but then the returned value must be checked for null
* `baseView` could be null which is not checked. Thus initializing xPos/yPos this way won't work because it must be done in the if-block

PR #6477, opened 2022-02-28
[BUGFIX] PartDesign: fix crash in MultiTransform if moving on empty list

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=3&t=66713

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskMultiTransformParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskMultiTransformParameters.cpp
@@ -451,6 +451,8 @@ void TaskMultiTransformParameters::moveTransformFeature(const int increment)
     PartDesign::MultiTransform* pcMultiTransform = static_cast<PartDesign::MultiTransform*>(TransformedView->getObject());
     std::vector<App::DocumentObject*> transformFeatures = pcMultiTransform->Transformations.getValues();
 
+    if (transformFeatures.empty()) return;
+
     App::DocumentObject* feature = transformFeatures[row];
     transformFeatures.erase(transformFeatures.begin() + row);
     QListWidgetItem* item = new QListWidgetItem(*(ui->listTransformFeatures->item(row)));


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6477">pipeline status</a> for feature branch [PR_6477](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6477). Pipeline [481498938](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/481498938) was triggered at [45d6d2b](https://github.com/FreeCAD/FreeCAD/pull/6477/commits/45d6d2bdedf76bd1be96996f521d3614e720e79a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskMultiTransformParameters.cpp
```diff
@@ -451,6 +451,8 @@ void TaskMultiTransformParameters::moveTransformFeature(const int increment)
     PartDesign::MultiTransform* pcMultiTransform = static_cast<PartDesign::MultiTransform*>(TransformedView->getObject());
     std::vector<App::DocumentObject*> transformFeatures = pcMultiTransform->Transformations.getValues();
 
+    if (transformFeatures.empty()) return;
```
donovaly, line 454: Thanks for the fix.
However, we put everything after the closing parentheses in a new line. In general, as far as I can see in our code, a return statement has its own line.

0penBrain: > However, we put everything after the closing parentheses in a new line. In general, as far as I can see in our code, a return statement has its own line.

Really ?
```
~/dev/FreeCAD$ grep -rE "^\s*if.*return.*;" src/Base src/Gui src/App src/Mod/Part | wc -l
355
```
> I also see that a comment would be good because without it, it is not clear why we return/check the emptyness.

Do we really need a comment explaining why we don't want to dereference an empty vector  (that is initialized line above and then used line below) ???

donovaly: Why not. A comment never harms. When someone new looks at this in 2 years, it is not clear to him, why we just return and not e.g. issue also a message, error etc.

That there are occurrences where return is in the same line than if, does not mean this should not be improved. I think the consensus is that if has its own line and in this case this automatically means return has its own line too ;-)

wwmayer: > Really ?

Well, in many coding guidelines it's considered good practice to put every statement into its own line. This improves readability and avoids to do "unexpected" things because a statement may have been overlooked.

donovaly: >A comment can of course harm because it may explain the wrong thing or something irrelevant.

Sure, but you got my point.

> Everybody who claims to be a developer should be able to read and understand the code in the next block. 

I don't like what the statement implies. I pointed out that comments help to understand quicker and of course also non-developers look at code. I program also a lot at work and for example a colleague who never did UI programming, only scripting can have a look and even find mistakes. I see that it helps people like him to document more than I would do. So a general statement about the capabilities of future people looking at code is, well, not fair.

However, here I thought it would not harm to write why no message is output or why the case that leads to the empty array is not prevented. And this is my main point. I did not have time to have a closer look why we allow the UI action to get the empty array. There might be a good reason, but for me, just looking at the code snippet, it is not immediately clear. And that's why I think comments help. Then everything is quickly clear at the first look.

I am not insisting on a comment. When I propose to add a comment, it is not that I require it, it is to improve the situation and not to annoy people.

adrianinsaval: open source also attracts amateurs that sometimes learn from what they see in the codebase, I know this is not a priority but good comments on code are certainly helpful fur us amateurs. I like to thing that even my amateurish remarks are sometimes useful, they have resulted in some corrections. I do share werner's concerns that the code might be updated and the comment not, I've seen it quite a few times and been confused. In this particular case I think it's pretty self explanatory "if nothing to act on then don't try to act"

PR #6353, opened 2022-02-13
Part: Loft, Sweep, Extrude, Revolution, update dialogs to work with A…

mwganson:
…pp::Link objects

https://forum.freecadweb.org/viewtopic.php?f=10&t=66159

PR #6478, opened 2022-02-28
Part: Improve support for Links

mwganson:
I accidentally closed the other PR for this, so I am creating this one.  There haven't been any changes to the code except to resolve some conflicts with a recent commit (removing some includes).

Link to closed PR: https://github.com/FreeCAD/FreeCAD/pull/6353

Link to forum discussions:

https://forum.freecadweb.org/viewtopic.php?f=10&t=66159
https://forum.freecadweb.org/viewtopic.php?f=17&t=66449

It's a fairly big PR and the feature freeze is coming, so this probably will be for 0.21.  But I think it would be nice to improve Links support in Part for 0.20, if possible.

### DIFF
--- a/src/Mod/Part/Gui/Command.cpp
+++ b/src/Mod/Part/Gui/Command.cpp
@@ -37,6 +37,7 @@
 #include <Base/Console.h>
 #include <Base/Exception.h>
 #include <Base/Tools.h>
+#include <App/Link.h>
 #include <Gui/Action.h>
 #include <Gui/Application.h>
 #include <Gui/BitmapFactory.h>
@@ -250,6 +251,38 @@ bool checkForSolids(const TopoDS_Shape& shape)
     }
 
     return true;
+}
+/*
+ * returns vector of Part::TopoShapes from selected Part::Feature derived objects,
+ * App::Links linked to Part::Features, or App::Part containers with visible Part::Features
+ */
+std::vector<Part::TopoShape> getShapesFromSelection()
+{
+    std::vector<App::DocumentObject*> objs = Gui::Selection().getObjectsOfType(App::DocumentObject::getClassTypeId());
+    std::vector <Part::TopoShape> shapes;
+    for (std::vector<App::DocumentObject*>::iterator it = objs.begin(); it != objs.end(); ++it) {
+        Part::TopoShape shp = Part::Feature::getTopoShape(*it);
+        if (!shp.isNull()){
+            shapes.push_back(shp);
+        }
+    }
+    return shapes;
+}
+/*
+ * returns true if selected objects contain valid Part::TopoShapes.
+ * Objects can be Part::Features, App::Links, or App::Parts
+ */
+bool hasShapesInSelection()
+{
+    bool hasShapes = false;
+    std::vector<App::DocumentObject*> docobjs = Gui::Selection().getObjectsOfType(App::DocumentObject::getClassTypeId());
+    for (std::vector<App::DocumentObject*>::iterator it = docobjs.begin(); it != docobjs.end(); ++it) {
+        if (!Part::Feature::getTopoShape(*it).isNull()) {
+            hasShapes = true;
+            break;
+        }
+    }
+    return hasShapes;
 }
 }
 
@@ -1270,7 +1303,7 @@ void CmdPartReverseShape::activated(int iMsg)
 bool CmdPartReverseShape::isActive(void)
 {
     return Gui::Selection().countObjectsOfType
-        (Part::Feature::getClassTypeId()) > 0;
+        (Part::Feature::getClassTypeId(), 0, 3) > 0;
 }
 
 //===========================================================================
@@ -1518,11 +1551,10 @@ void CmdPartCrossSections::activated(int iMsg)
     Q_UNUSED(iMsg);
     Gui::TaskView::TaskDialog* dlg = Gui::Control().activeDialog();
     if (!dlg) {
-        std::vector<App::DocumentObject*> obj = Gui::Selection().getObjectsOfType
-            (Part::Feature::getClassTypeId());
+        std::vector<Part::TopoShape> shapes = PartGui::getShapesFromSelection();
         Base::BoundBox3d bbox;
-        for (std::vector<App::DocumentObject*>::iterator it = obj.begin(); it != obj.end(); ++it) {
-            bbox.Add(static_cast<Part::Feature*>(*it)->Shape.getBoundingBox());
+        for (std::vector<Part::TopoShape>::iterator it = shapes.begin(); it != shapes.end(); ++it) {
+            bbox.Add((*it).getBoundBox());
         }
         dlg = new PartGui::TaskCrossSections(bbox);
     }
@@ -1531,8 +1563,8 @@ void CmdPartCrossSections::activated(int iMsg)
 
 bool CmdPartCrossSections::isActive(void)
 {
-    return (Gui::Selection().countObjectsOfType(Part::Feature::getClassTypeId()) > 0 &&
-            !Gui::Control().activeDialog());
+    bool hasShapes = PartGui::hasShapesInSelection();
+    return (hasShapes && !Gui::Control().activeDialog());
 }
 
 //===========================================================================
@@ -1643,9 +1675,16 @@ CmdPartOffset::CmdPartOffset()
 void CmdPartOffset::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    auto shapes = getSelection().getObjectsOfType(Part::Feature::getClassTypeId(),nullptr,3);
-    if(shapes.empty())
+    std::vector<App::DocumentObject*> docobjs = Gui::Selection().getObjectsOfType(App::DocumentObject::getClassTypeId());
+    std::vector<App::DocumentObject*> shapes;
+    for (std::vector<App::DocumentObject*>::iterator it = docobjs.begin(); it != docobjs.end(); ++it) {
+        if (!Part::Feature::getTopoShape(*it).isNull()) {
+           shapes.push_back(*it);
+        }
+    }
+    if (shapes.size() != 1) {
         return;
+    }
     App::DocumentObject* shape = shapes.front();
     std::string offset = getUniqueObjectName("Offset");
 
@@ -1668,9 +1707,15 @@ void CmdPartOffset::activated(int iMsg)
 
 bool CmdPartOffset::isActive(void)
 {
-    Base::Type partid = Base::Type::fromName("Part::Feature");
-    bool objectsSelected = Gui::Selection().countObjectsOfType(partid,nullptr,3) == 1;
-    return (objectsSelected && !Gui::Control().activeDialog());
+    {
+        bool hasShapes = PartGui::hasShapesInSelection();
+        std::vector<App::DocumentObject*> docobjs = Gui::Selection().getObjectsOfType(App::DocumentObject::getClassTypeId());
+        return (hasShapes && !Gui::Control().activeDialog() && docobjs.size() == 1);
+    }
+
+//    Base::Type partid = Base::Type::fromName("Part::Feature");
+//    bool objectsSelected = Gui::Selection().countObjectsOfType(partid,0,3) == 1;
+//    return (objectsSelected && !Gui::Control().activeDialog());
 }
 
 
@@ -1695,9 +1740,17 @@ CmdPartOffset2D::CmdPartOffset2D()
 void CmdPartOffset2D::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    auto shapes = getSelection().getObjectsOfType(Part::Feature::getClassTypeId(),nullptr,3);
-    if(shapes.empty())
+    std::vector<App::DocumentObject*> docobjs = Gui::Selection().getObjectsOfType(App::DocumentObject::getClassTypeId());
+    std::vector<App::DocumentObject*> shapes;
+
+    for (std::vector<App::DocumentObject*>::iterator it = docobjs.begin(); it != docobjs.end(); ++it) {
+        if (!Part::Feature::getTopoShape(*it).isNull()) {
+           shapes.push_back(*it);
+        }
+    }
+    if (shapes.size() != 1) {
         return;
+    }
     App::DocumentObject* shape = shapes.front();
     std::string offset = getUniqueObjectName("Offset2D");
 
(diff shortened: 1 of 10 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478">pipeline status</a> for feature branch [PR_6478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478). Pipeline [481498940](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/481498940) was triggered at [a02ccc0](https://github.com/FreeCAD/FreeCAD/pull/6478/commits/a02ccc020b2641fdedaaba5cbabf819d156ef3a5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
> I accidentally closed the other PR for this, so I am creating this one.

You can re-open a closed PR, right? :confused:

mwganson:
It kept showing 0 commits even though I added commits to it.  Maybe after reopening they would have shown up?

donovaly:
@realthunder , can you please review this PR?

realthunder:
@mwganson Hi, I've gone through your changes. Here are my thoughts.

First, `DocumentObject::getLinkedObject()` by default will recursively resolve to the final linked object. If the object is not a link, it will return itself. So it is always safe to just call `getLinkedObject()` directly. And normally, you shouldn't explicitly check of `App::Link::getClassTypeId()`, as the core functionality of link is implemented as an extension `App::LinkBaseExtension` and can be install to any object. It is enough to simply use `getLinkedObject()` to check for link.

Second, If you have removed `Part::Feature::getClassTypeId()` checking in the relevant part features, you don't need to check it in the relevant command or dialog either. You can simply check if `Part.getShape()` returns a valid shape, which makes it possible to support other non `Part::Feature`, such as `App::Part` container. My general idea is to put as little type checking as possible. Since all that's needed as input for most part features is the shape, then that's all we need to check, instead of the input object type. I haven't done so already because of one complication. It is currently very difficult to obtain per face color from non part features, especially for containers which may be nested. There will be a general solution to this problem once the new topo naming gets merged. 

Third, speaking of topo naming, although it is not relevant now, it would be better to use `Part::TopoShape` whenever possible, and the companion `Part::Feature::getToposhape()`, instead of `TopoDS_Shape` or `Part::getShape()`, because future topo naming information will be stored inside `Part::TopoShape`.

So my advice is that, if this PR is aiming for 0.20 release, then please make some changes to the usage of `getLinkedObject()`, and remove App::Link::getClassTypeId() check. If it is for the next development cycle, then I'd suggest to remove most of the type checking and check for input shape directly.

mwganson:
@realthunder

Thanks for the information.

I will revise based on your feedback.

mwganson:
@realthunder 

I'm making some good progress, and I think it will be useful to be able to use App:Part objects similarly to using App::Link objects, but I have discovered something.  If the App::Part container contains a sketch I can extrude it by selecting the Part and clicking Extrude toolbar icon.  But if the sketch is later made invisible the Extrude object fails to recompute.

Same is true for Check Geometry function.  If I have for example a Cube and Cylinder in the App::Part container, select the container, and click the Check Geometry toolbar icon, it checks both the Cube and the Cylinder (and might show some self-intersections if they overlap), but if Cube is made invisible it does not get checked, same for Cylinder.

The behavior seems to be if the Part::Features within the App::Part are made invisible they are not returned as part of the shape with:
//docobj is a pointer to some App::Part document object.
Part::TopoShape topoShape = Part::Feature::getTopoShape(*docobj->getLinkedObject());
topoShape.isNull(); //true if all objects in the App:Part are hidden, false if any are visible

This can be demonstrated in python without need to build the PR.

Create a App::Part container and put a Part::Cylinder into it.  Select the Part and press Ctrl+Shift+P.  In the python console:

shp = obj.getLinkedObject().Shape

This works if the Cylinder is visible, but fails with App.Part has not attribute 'Shape' if the Cylinder is made invisible.

Is this the expected behavior or a bug?  I can see how it could be useful to be able to control which shapes are in the compound that represents the App::Part container's shape by making them visible or invisible, but this will also be a source of confusion for many users.  Visibility does not normally matter when it comes to the shape of an object.

mwganson:
Converting to draft because I still have a few more dialogs to revise.  Ones I have done so far:  ruled surface, check geometry, extrude, thickness, offset2d, offset 3d, cross-sections, and revolution.

mwganson:
I am using Part::TopoShape where possible, but my knowledge of OCC is not great, so where TopoDS_Shape is already being used it is trivial to convert from TopoShape to TopoDS_Shape.

Part::TopoShape topoShape = Part::Feature::getTopoShape(docobj->getLinkedObject());
TopoDS_Shape shape = topoShape.getShape();

This will hopefully make it a bit easier if the TopoShape is needed later for topo naming mitigation since it will be there already.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478">pipeline status</a> for feature branch [PR_6478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478). Pipeline [491259460](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/491259460) was triggered at [86db251](https://github.com/FreeCAD/FreeCAD/pull/6478/commits/86db2513c319f6c2db87370142b375a021beae16). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

realthunder:
`Part::Feature::getShape/getTopoShape()` will handle link inside. In fact, if the link is at a different placement, calling `getTopoShape(obj->getLinkedObject())` will obtain the shape at the wrong location.

realthunder:
> The behavior seems to be if the Part::Features within the App::Part are made invisible they are not returned as part of the shape with:

I have an very old pending PR that address this problem, by having the `ExportMode` attribute in group feature to control how to export shape of a group. The default the export by visibility. There is also option to control export by injecting property into each child for individual control. I'll see if I can refresh this PR.

mwganson:
@realthunder 
Okay.  I will use Part::Feature::getTopoShape(*docobj) instead of Part::Feature::getTopoShape(*docobj -> getLinkedObject())

I have discovered another anomaly when using the App::Part containers.

Create a new App::Part.  Add a sketch to it with a rectangle inside the sketch.  Make the sketch visible.  Select the App::Part, press Ctrl+Shift+P.  In the python console:

obj.Shape
<compound> (as expected)

Extrude the sketch.  Add the extrude to the App::Part.  So now you have:

```
Part
    > Origin
    >  Extrude
            Sketch
```

Now make Extrude invisible and Sketch visible.  Select the App::Part, Ctrl+Shift+P.  In the console:

obj.Shape
Attribute error: 'App.Part' object has no attribute 'Shape'.

When selecting the App::Part, the sketch appears green in the 3d view, indicating it has been selected.  It looks like what is happening is only the top level shapes are exported, and only then if they are visible.

mwganson:
Some things I did not touch because we are in the feature freeze and already this is a relatively large PR.  Among the untouched: Fillets, Chamfers, Project to face, Shape builder, Defeaturing.

realthunder:
> When selecting the App::Part, the sketch appears green in the 3d view, indicating it has been selected. It looks like what is happening is only the top level shapes are exported, and only then if they are visible.

This is by design. The part container (actually `App::GeoFeatureGroupExtension`) will auto add all dependency of its children. But those dependency are not claimed in the tree view. The logic is that since those dependency are used to create the shape of some child object in the container, it shouldn't be part of the compound shape of the container. Think of a fusion in a Part container. It does not make sense to include the tool shapes of the fusion as they'll occupy the same 3D space. It makes even less sense for a feature like cut.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478">pipeline status</a> for feature branch [PR_6478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478). Pipeline [492097089](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/492097089) was triggered at [204dbed](https://github.com/FreeCAD/FreeCAD/pull/6478/commits/204dbed0a1c2b026555e8b486ff772155e03fc19). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478">pipeline status</a> for feature branch [PR_6478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478). Pipeline [493289350](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/493289350) was triggered at [62f885a](https://github.com/FreeCAD/FreeCAD/pull/6478/commits/62f885a9e737c1e0fc89bb6a01bdf878b12c4d8a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

mwganson:
I think it's ready to go unless changes are required upon review.  I have rebased locally.  After I finish compiling and running the tests I will force push.

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1516,11 +1517,17 @@ void CmdPartCrossSections::activated(int iMsg)
...
-        std::vector<App::DocumentObject*> obj = Gui::Selection().getObjectsOfType
-            (Part::Feature::getClassTypeId());
+        std::vector<App::DocumentObject*> objs = Gui::Selection().getObjectsOfType(App::DocumentObject::getClassTypeId());
+        std::vector <Part::TopoShape> shapes;
+        for (std::vector<App::DocumentObject*>::iterator it = objs.begin(); it != objs.end(); ++it) {
+            Part::TopoShape shp = Part::Feature::getTopoShape(*it);
+            if (!shp.isNull()){
+                shapes.push_back(shp);
+            }
+        }
```
realthunder, line 1527: This piece of code is used in several place. Maybe create a static function for it.

src/Mod/Part/Gui/Command.cpp
```diff
@@ -1529,8 +1536,15 @@ void CmdPartCrossSections::activated(int iMsg)
...
-    return (Gui::Selection().countObjectsOfType(Part::Feature::getClassTypeId()) > 0 &&
-            !Gui::Control().activeDialog());
+    bool hasShapes = false;
+    std::vector<App::DocumentObject*> docobjs = Gui::Selection().getObjectsOfType(App::DocumentObject::getClassTypeId());
+    for (std::vector<App::DocumentObject*>::iterator it = docobjs.begin(); it != docobjs.end(); ++it) {
+        if (!Part::Feature::getTopoShape((*it)).isNull()){
+            hasShapes = true;
+            break;
+        }
+    }
```
realthunder, line 1546: Same here. Maybe use a static function.

src/Mod/Part/Gui/TaskSweep.cpp
```diff
@@ -83,11 +84,16 @@ class SweepWidget::Private
         }
         bool allow(App::Document* /*pDoc*/, App::DocumentObject*pObj, const char*sSubName)
         {
-            if (pObj->getTypeId().isDerivedFrom(Part::Feature::getClassTypeId())) {
+            if (pObj->getTypeId().isDerivedFrom(Part::Feature::getClassTypeId()) ||
+                pObj->getTypeId().isDerivedFrom(App::Link::getClassTypeId())) {
```
realthunder, line 88: There is no need to check for object type here.

src/Mod/Part/Gui/TaskSweep.cpp
```diff
@@ -277,9 +287,16 @@ bool SweepWidget::accept()
         return false;
     Gui::SelectionFilter edgeFilter  ("SELECT Part::Feature SUBELEMENT Edge COUNT 1..");
     Gui::SelectionFilter partFilter  ("SELECT Part::Feature COUNT 1");
+    Gui::SelectionFilter linkFilter  ("SELECT App::Link COUNT 1");
```
realthunder, line 290: Consider replacing `SelectionFilter` here. Just obtain the selected shape and check for the shape type.

realthunder (changes requested):
The PR is mostly okay. Just a few minor changes to make it more flexible.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478">pipeline status</a> for feature branch [PR_6478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478). Pipeline [497808538](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/497808538) was triggered at [a07b7cb](https://github.com/FreeCAD/FreeCAD/pull/6478/commits/a07b7cb8ae46fb6045dcab98a42465174c9c964f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

mwganson:
Sorry, I did not get a notification of the new reviews.  This is now tagged for 0.21 anyway.  But I suppose it could be argued it is a bug fix and not a new feature.

donovaly:
> But I suppose it could be argued it is a bug fix and not a new feature.

OK, putting it then on the 0.20 milestone list.

src/Mod/Part/Gui/Command.cpp
mwganson: Okay.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478">pipeline status</a> for feature branch [PR_6478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6478). Pipeline [508810306](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/508810306) was triggered at [7853243](https://github.com/FreeCAD/FreeCAD/pull/6478/commits/7853243c3d791bb9e6776ad197b37f5aea92d356). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@mwganson what is your time frame for this PR? Since it is in the 0.20 milestone, it should be merged soon, let's say it should be ready within a week? Would that be possible?

mwganson:
I think it's ready now.  Sooner the better so more people can build and test and report any issues.

donovaly:
@realthunder , is the PR OK to be merged in your opinion?

realthunder:
Looks fine to me.

donovaly:
Great, so one PR less from the 0.20 milestone.
Merging now.

donovaly:
@mwganson , don't forget to add documentation to the Wiki and to add it to our release notes: https://wiki.freecadweb.org/Release_notes_0.20

PR #6483, opened 2022-03-01
[Base] Parameter and Matrix: remove unused includes

donovaly:


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6483">pipeline status</a> for feature branch [PR_6483](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6483). Pipeline [481535212](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/481535212) was triggered at [4d1ddf3](https://github.com/FreeCAD/FreeCAD/pull/6483/commits/4d1ddf3eeb5046a152cc6d29116637f28b8d9d91). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Base/Matrix.cpp
```diff
@@ -23,15 +23,14 @@
 
 #include "PreCompiled.h"
 #ifndef _PreComp_
-# include <memory>
-# include <cstring>
 # include <sstream>
+# include <string>
```
wwmayer, line 27: Why do you add string when it's already included in the header?

donovaly: Because I got a compilation error. I found out yesterday, that the MSVC compiler use different files, not necessarily the one in the header, see my next comment.

donovaly: Yes.

src/Base/Matrix.cpp
```diff
@@ -23,15 +23,14 @@
 
 #include "PreCompiled.h"
 #ifndef _PreComp_
-# include <memory>
```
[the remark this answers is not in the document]
donovaly: But why does it then compile for me? I remove the include, it compiles. MSVC tells me that it loads vcruntime_string.h, but how is then vcruntime_string.h loaded?

PR #6506, opened 2022-03-04
Gui: introduce ShortcutManager for shortcut management and conflict resolving

realthunder:
Forum post https://forum.freecadweb.org/viewtopic.php?f=17&t=66821

Add new class ShortcutManager to unify shortcut management. Implement through global event filter to support longest key sequence match with user defined delay (configurable through 'Customize -> Keyboard -> Key sequence delay'). Also supports user defined priority to resolve shortcut conflict through 'Customize -> Keyboard'.

Refactored 'Customize -> Keyboard' page for easier command search and better synchronization.

https://user-images.githubusercontent.com/1207888/156695515-1e6abfa2-04ca-4a92-8a69-80835b255f21.mp4

### DIFF
--- a/src/Gui/Action.cpp
+++ b/src/Gui/Action.cpp
@@ -209,11 +209,12 @@ void Action::setToolTip(const QString & s, const QString & title)
                 title.isEmpty() ? _action->text() : title, 
                 _action->font(),
                 _action->shortcut().toString(QKeySequence::NativeText),
-                this));
+                _pcCmd));
 }
 
-static QString & actionTitle(QString & title)
+QString Action::cleanTitle(const QString & text)
 {
+    QString title(text);
     // Deal with QAction title mnemonic
     static QRegularExpression re(QStringLiteral("&(.)"));
     title.replace(re, QStringLiteral("\\1"));
@@ -223,19 +224,63 @@ static QString & actionTitle(QString & title)
     return title;
 }
 
-static inline QString actionTitle(const QString &title)
+QString Action::commandToolTip(const Command *cmd, bool richFormat)
 {
-    QString text(title);
-    return actionTitle(text);
+    if (!cmd)
+        return QString();
+
+    if (richFormat) {
+        if (auto action = cmd->getAction())
+            return action->_action->toolTip();
+    }
+
+    QString title, tooltip;
+    if (dynamic_cast<const MacroCommand*>(cmd)) {
+        if (auto txt = cmd->getMenuText())
+            title = QString::fromUtf8(txt);
+        if (auto txt = cmd->getToolTipText())
+            tooltip = QString::fromUtf8(txt);
+    } else {
+        if (auto txt = cmd->getMenuText())
+            title = qApp->translate(cmd->className(), txt);
+        if (auto txt = cmd->getToolTipText())
+            tooltip = qApp->translate(cmd->className(), txt);
+    }
+
+    if (!richFormat)
+        return tooltip;
+    return createToolTip(tooltip, title, QFont(), cmd->getShortcut(), cmd);
+}
+
+QString Action::commandMenuText(const Command *cmd)
+{
+    if (!cmd)
+        return QString();
+    
+    QString title;
+    if (auto action = cmd->getAction())
+        title = action->text();
+    else if (dynamic_cast<const MacroCommand*>(cmd)) {
+        if (auto txt = cmd->getMenuText())
+            title = QString::fromUtf8(txt);
+    } else {
+        if (auto txt = cmd->getMenuText())
+            title = qApp->translate(cmd->className(), txt);
+    }
+    if (title.isEmpty())
+        title = QString::fromUtf8(cmd->getName());
+    else
+        title = cleanTitle(title);
+    return title;
 }
 
 QString Action::createToolTip(QString _tooltip,
                               const QString & title,
                               const QFont &font,
                               const QString &sc,
-                              Action *act)
+                              const Command *pcCmd)
 {
-    QString text = actionTitle(title);
+    QString text = cleanTitle(title);
 
     if (text.isEmpty())
         return _tooltip;
@@ -259,17 +304,18 @@ QString Action::createToolTip(QString _tooltip,
             text.toHtmlEscaped(), shortcut.toHtmlEscaped());
 
     QString cmdName;
-    auto pcCmd = act ? act->_pcCmd : nullptr;
     if (pcCmd && pcCmd->getName()) {
         cmdName = QString::fromLatin1(pcCmd->getName());
-        if (auto groupcmd = dynamic_cast<GroupCommand*>(pcCmd)) {
-            int idx = act->property("defaultAction").toInt();
-            auto cmd = groupcmd->getCommand(idx);
-            if (cmd && cmd->getName())
-                cmdName = QStringLiteral("%1 (%2:%3)")
-                    .arg(QString::fromLatin1(cmd->getName()))
-                    .arg(cmdName)
-                    .arg(idx);
+        if (auto groupcmd = dynamic_cast<const GroupCommand*>(pcCmd)) {
+            if (auto act = pcCmd->getAction()) {
+                int idx = act->property("defaultAction").toInt();
+                auto cmd = groupcmd->getCommand(idx);
+                if (cmd && cmd->getName())
+                    cmdName = QStringLiteral("%1 (%2:%3)")
+                        .arg(QString::fromLatin1(cmd->getName()))
+                        .arg(cmdName)
+                        .arg(idx);
+            }
         }
         cmdName = QStringLiteral("<p style='white-space:pre; margin-top:0.5em;'><i>%1</i></p>")
             .arg(cmdName.toHtmlEscaped());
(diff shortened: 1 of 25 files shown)


luzpaz:
This looks awesome!

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506">pipeline status</a> for feature branch [PR_6506](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506). Pipeline [485159467](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/485159467) was triggered at [62fce1c](https://github.com/FreeCAD/FreeCAD/pull/6506/commits/62fce1cbb04a290858a733d9d9605985b8cea6fd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/Action.cpp
```diff
@@ -183,21 +190,121 @@ QString Action::statusTip() const
...
+}
+
+QString Action::createToolTip(QString _tooltip,
+                              const QString & title,
+                              const QFont &font,
+                              const QString &sc,
+                              Action *act)
+{
+    QString text = title;
+    text.remove(QLatin1Char('&'));;
```
chennes, line 220: This will remove even "escaped" ampersands (&&).

src/Gui/Action.cpp
```diff
@@ -1243,4 +1350,206 @@ void WindowAction::addTo ( QWidget * w )
...
+    QIcon icon;
+    bool iconChecked = false;
+};
+static std::vector<CmdInfo> _Commands;
+static int _CommandRevision;
+
+class CommandModel : public QAbstractItemModel
+{
+public:
+
```
chennes, line 1368:
```suggestion
```

src/Gui/Action.cpp
```diff
@@ -1243,4 +1350,206 @@ void WindowAction::addTo ( QWidget * w )
...
+        if (index.row() < 0 || index.row() >= (int)_Commands.size())
+            return QVariant();
+
+        auto &info = _Commands[index.row()];
+
+        switch(role) {
+        case Qt::DisplayRole:
+        case Qt::EditRole:
+            if (info.text.isEmpty()) {
+#if QT_VERSION>=QT_VERSION_CHECK(5,2,0)
```
chennes, line 1408: We don't support older versions of Qt anymore, so this should be unnecessary.

src/Gui/Action.cpp
```diff
@@ -1243,4 +1350,206 @@ void WindowAction::addTo ( QWidget * w )
...
+                info.tooltip = QString::fromLatin1("%1: %2").arg(
+                        QString::fromLatin1(info.cmd->getName()),
+                        qApp->translate(info.cmd->className(), info.cmd->getMenuText()));
+                QString tooltip = qApp->translate(info.cmd->className(), info.cmd->getToolTipText());
+                if (tooltip.size())
+                    info.tooltip += QString::fromLatin1("\n\n") + tooltip;
+            }
+            return info.tooltip;
+
+        case Qt::UserRole:
```
chennes, line 1440: I think it's usually clearer to make a constant for the `Qt::UserRole` with a more descriptive name.

src/Gui/Action.cpp
```diff
@@ -1243,4 +1350,206 @@ void WindowAction::addTo ( QWidget * w )
...
+
+// --------------------------------------------------------------------
+
+CommandCompleter::CommandCompleter(QLineEdit *lineedit, QObject *parent)
+    : QCompleter(parent)
+{
+    this->setModel(new CommandModel(this));
+#if QT_VERSION>=QT_VERSION_CHECK(5,2,0)
+    this->setFilterMode(Qt::MatchContains);
+#endif
```
chennes, line 1473:
```suggestion
    this->setFilterMode(Qt::MatchContains);
```

src/Gui/Action.cpp
```diff
@@ -1243,4 +1350,206 @@ void WindowAction::addTo ( QWidget * w )
...
+    this->setCaseSensitivity(Qt::CaseInsensitive);
+    this->setCompletionMode(QCompleter::PopupCompletion);
+    this->setWidget(lineedit);
+    connect(lineedit, SIGNAL(textEdited(QString)), this, SLOT(onTextChanged(QString)));
+    connect(this, SIGNAL(activated(QModelIndex)), this, SLOT(onCommandActivated(QModelIndex)));
```
chennes, line 1478: Prefer new-style connections.

src/Gui/Action.cpp
```diff
@@ -1243,4 +1350,206 @@ void WindowAction::addTo ( QWidget * w )
...
+            && (o == this->widget() || o == this->popup()))
+    {
+        QKeyEvent * ke = static_cast<QKeyEvent*>(ev);
+        switch(ke->key()) {
+        case Qt::Key_Escape: {
+            auto edit = qobject_cast<QLineEdit*>(this->widget());
+            if (edit && edit->text().size()) {
+                edit->setText(QString());
+                popup()->hide();
+                return true;
```
chennes, line 1494: Can you refactor to avoid repeating yourself here?

src/Gui/Action.cpp
```diff
@@ -1243,4 +1350,206 @@ void WindowAction::addTo ( QWidget * w )
...
+
+void CommandCompleter::onCommandActivated(const QModelIndex &index)
+{
+    QByteArray name = completionModel()->data(index, Qt::UserRole).toByteArray();
+    Q_EMIT commandActivated(name);
+}
+
+void CommandCompleter::onTextChanged(const QString &txt)
+{
+    if (txt.size() < 3 || !widget())
```
chennes, line 1543: The 3 needs some documentation here, otherwise it looks like a magic number.

src/Gui/Command.cpp
```diff
@@ -239,6 +258,12 @@ void Command::addTo(QWidget *pcWidget)
 {
     if (!_pcAction) {
         _pcAction = createAction();
+#ifdef FC_DEBUG
+        // Accelerator conflict can now be dynamically resolved in ShortcutManager
+        //
+        // printConflictingAccelerators();
+#endif
```
chennes, line 265: No reason to leave this in
```suggestion
```

src/Gui/Command.cpp
```diff
@@ -255,6 +280,12 @@ void Command::addToGroup(ActionGroup* group)
...
```
chennes, line 287:
```suggestion
```

src/Gui/CommandView.cpp
```diff
@@ -254,14 +254,17 @@ class StdCmdFreezeViews : public Gui::Command
...
+
 protected:
-    virtual void activated(int iMsg);
-    virtual bool isActive(void);
-    virtual Action * createAction(void);
-    virtual void languageChange();
+    virtual void activated(int iMsg) override;
+    virtual bool isActive(void) override;
+    virtual Action * createAction(void) override;
+    virtual void languageChange() override;
```
chennes, line 267: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final
```suggestion
   void activated(int iMsg) override;
    bool isActive(void) override;
    Action * createAction(void) override;
    void languageChange() override;

```

src/Gui/DlgKeyboardImp.cpp
```diff
@@ -71,6 +77,231 @@ DlgCustomKeyboardImp::DlgCustomKeyboardImp( QWidget* parent  )
...
+                                            QAbstractButton *buttonDown)
+{
+    QStringList labels;
+    labels << tr("Name") << tr("Title");
+    priorityList->setHeaderLabels(labels);
+    priorityList->header()->hide();
+    priorityList->header()->setSectionResizeMode(0, QHeaderView::ResizeToContents);
+    priorityList->header()->setSectionResizeMode(1, QHeaderView::ResizeToContents);
+
+    auto updatePriorityList = [priorityList](bool up) {
```
chennes, line 203: Using a boolean here feels pretty kludgy. I can live with it here since it's in such an isolated use-case, but I think it would be better to use a more explicit UP/DOWN specifier.

src/Gui/DlgKeyboardImp.cpp
```diff
@@ -71,6 +77,231 @@ DlgCustomKeyboardImp::DlgCustomKeyboardImp( QWidget* parent  )
...
+    initCommandCompleter(editCommand, comboGroups, commandTreeWidget);
+    auto conn = initCommandList(commandTreeWidget, comboGroups);
+
+    if (priorityList && buttonUp && buttonDown) {
+        initPriorityList(priorityList, buttonUp, buttonDown);
+
+        auto timer = new QTimer(priorityList);
+        timer->setSingleShot(true);
+        if (currentShortcut)
+            QObject::connect(currentShortcut, &QLineEdit::textChanged, timer, [timer](){timer->start(200);});
```
chennes, line 256: How did you arrive at the 200ms for these?

src/Gui/ShortcutManager.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+void ShortcutManager::OnChange(Base::Subject<const char*> &src, const char *reason)
+{
+    if (hSetting == &src) {
+        if (boost::equals(reason, "ShortcutTimeout"))
+            timeout = hSetting->GetInt("ShortcutTimeout");
+        return;
+    }
+
+    if (busy)
```
chennes, line 94: Where is `busy` ever set to `true`?

src/Gui/ShortcutManager.cpp
```diff
@@ -0,0 +1,455 @@
...
+        lastFocus = nullptr;
+        onTimer();
+        return true;
+    }
+
+    lastFocus = focus;
+    pendingSequence = key;
+
+    // Qt's shortcut state machine favors shortest match (which is ridiculous,
+    // unless I'm mistaken?). We'll do longest match. We've disabled all
```
chennes, line 249: I don't know if you're mistaken, but I agree with you that longest match should be the correct behavior.

src/Gui/ShortcutManager.cpp
```diff
@@ -0,0 +1,455 @@
...
+    return res;
+}
+
+void ShortcutManager::setPriorities(const std::vector<QByteArray> &actions)
+{
+    if (actions.empty())
+        return;
+    // Keep the same top priority of the given action, and adjust the rest. Can
+    // go negative if necessary
+    int current = 0;
```
chennes, line 363: If negative priorities are allowed, does it make sense to initialize `current` to zero?

src/Gui/ShortcutManager.cpp
```diff
@@ -0,0 +1,455 @@
...
+            if (it->key.shortcut != k) {
+                QKeyEvent *kev = new QKeyEvent(QEvent::KeyPress, s, Qt::ControlModifier, 0, 0, 0);
+                QApplication::postEvent(lastFocus, kev);
+                kev = new QKeyEvent(QEvent::KeyRelease, s, Qt::ControlModifier, 0, 0, 0);
+                QApplication::postEvent(lastFocus, kev);
+                break;
+            }
+        }
+    }
+    timer.stop();
```
chennes, line 452: It seems like this would make more sense at the top of the function, rather than the bottom.

src/Gui/ShortcutManager.h
```diff
@@ -0,0 +1,171 @@
...
+        }
+    };
+    struct ActionData {
+        ActionKey key;
+        intptr_t pointer;
+        QPointer<QAction> action;
+
+        ActionData(QAction *action, const char *name = "")
+            : key(action->shortcut(), name)
+            , pointer(reinterpret_cast<int64_t>(action))
```
chennes, line 131: Type name mismatch, you are using `intptr_t` above.

src/Gui/Workbench.cpp
```diff
@@ -311,6 +311,8 @@ void Workbench::setupCustomToolbars(ToolBarItem* root, const Base::Reference<Par
 
 void Workbench::setupCustomShortcuts() const
 {
+    // Now managed by ShortcutManager
+#if 0
```
chennes, line 315: I don't see any reason to leave more than just the comment here.

chennes (commented):
Overall I really like the feature, and the code looks good. Just a few comments:
1) The user interface for setting the multi-key delay makes it look like we are storing a delay per-keystroke. I don't have a great suggestion for improving it, but I don't know if we can include it as-is.
2) I think there needs to be a timeout, after which an in-progress keystroke sequence is cancelled even if the user presses no more keys. 
3) I'm not familiar enough with the Boost multi-index containers to really evaluate, I assume the reinterpret casts are related to the usage there? Obviously those set off red flags and warning sirens when reviewing code.

src/Gui/Action.cpp
realthunder: Noted. I do use new connection now. These code were added before we fully switch to Qt5.

src/Gui/Action.cpp
realthunder: There is subtle difference in handling of these two conditions. The ESC is supposed to clear the line edit text regardless of popup visibility. And if the edit is empty and popup is hidden, the ESC key shouldn't be filtered.

src/Gui/Action.cpp
realthunder: Comments added. To improve response time by not activate completer if text is too short.

src/Gui/Command.cpp
realthunder: Let's leave these there for now, because these are the only places printConflictingAccelerators() are used. They can be remove once the ShortcutManager is proven reliable.

src/Gui/DlgKeyboardImp.cpp
realthunder: Just to prevent refreshing priority list too often, since it gets triggered in a few signals.

realthunder:
> I think there needs to be a timeout, after which an in-progress keystroke sequence is cancelled even if the user presses no more keys.

The timer in ShortcutManager is used for this purpose. In fact, the `Multi-key delay` specifies the time to wait for more keystorke, after which the pending action will be triggered.

> I'm not familiar enough with the Boost multi-index containers to really evaluate, I assume the reinterpret casts are related to the usage there? Obviously those set off red flags and warning sirens when reviewing code.

I intend to use `QAction*` as map key for easy lookup, but the action may be deleted somewhere else. I use `QPointer` for safe accessing, and because `QPointer` is not immutable, it can't be used for key. I deliberately change the key type to `intptr_t` to prevent accidentally using it to access the action.

chennes:
> The timer in ShortcutManager is used for this purpose. In fact, the `Multi-key delay` specifies the time to wait for more keystorke, after which the pending action will be triggered.

It only does so if there *is* a pending action: if there isn't a match it just waits forever.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506">pipeline status</a> for feature branch [PR_6506](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506). Pipeline [485593433](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/485593433) was triggered at [b56a169](https://github.com/FreeCAD/FreeCAD/pull/6506/commits/b56a169d54c4a44ff65f540110a4865e35c0dfc0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

realthunder:
I see what you mean here. The thing is that `ShortcutManager` is intended to just help resolving the conflict of shortcuts, instead of replacing Qt's own private shortcut state machine entirely. That's why it only filters shortcut event instead of key press event. Otherwise, it will be much more difficult to implement. The current behavior kind of aligns with Qt's, which is to wait forever for key sequence to finish if there is partial match.

src/Gui/ShortcutManager.cpp
realthunder: I suppose it can work. It's just that the priority needs to rise from some base level. If the base is the lowest, then we'd have bunch of huge negative integers in the parameters, which would be confusing.

The reason the priority is allow to go negative is because I try to preserve the top priority (among actions with the same shortcuts) while reordering. This is a partial solution to a corner case where the user may have set priority in the past for some action in a workbench that hasn't been loaded yet. Keep the top priority unchanged can provide some degree of stability in this case.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506">pipeline status</a> for feature branch [PR_6506](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506). Pipeline [485782593](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/485782593) was triggered at [2d50b09](https://github.com/FreeCAD/FreeCAD/pull/6506/commits/2d50b091305b9180290e11859a4f95a24e86aa41). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506">pipeline status</a> for feature branch [PR_6506](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506). Pipeline [485787311](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/485787311) was triggered at [2cb2bac](https://github.com/FreeCAD/FreeCAD/pull/6506/commits/2cb2bac7fc417e3411d4eaf7cb1fa579ee202f1c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/Action.cpp
```diff
@@ -183,21 +192,135 @@ QString Action::statusTip() const
...
+    _tooltip = s;
+    _title = title;
+    _action->setToolTip(createToolTip(s,
+                title.isEmpty() ? _action->text() : title, 
+                _action->font(),
+                _action->shortcut().toString(QKeySequence::NativeText),
+                this));
+}
+
+static QString & actionTitle(QString & title)
```
wwmayer, line 215: In modern C++ it's recommended to put global functions into anonymous namespace, i.e.:
```
namespace {
...
}
```

src/Gui/Command.h
```diff
@@ -871,6 +881,12 @@ class GuiExport CommandManager
     void addCommandMode(const char* sContext, const char* sName);
     void updateCommands(const char* sContext, int mode);
 
+    /// Return a revision number to check for addition or removal of any command
+    int getRevision() { return _revision; }
```
wwmayer, line 885: Should be a const method.

src/Gui/DlgToolbarsImp.cpp
```diff
@@ -32,6 +32,7 @@
 #endif
```
wwmayer, line 33: The changes on this class cause some regressions:

 * Whenever opening the customize dialog and switching to the Toolbars tab it only shows the "Separator" item on the left side. In order to fill up this list one has to manually select a category, first.

* When the list is filled the "Separator" item is gone.

wwmayer:
Additionally, the Keyboard and Toolbars panel suffer from a further regression. When adding a new macro command in the Macros tab it isn't automatically added to the Keyboard and Toolbars panel while removing a macro still works.

The Commands panel is not affected.

realthunder:
@wwmayer New commits added for suggested changes and bug fix.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506">pipeline status</a> for feature branch [PR_6506](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506). Pipeline [490728511](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/490728511) was triggered at [9f40b81](https://github.com/FreeCAD/FreeCAD/pull/6506/commits/9f40b819cb801944bce709fcecf7d23559792c4e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506">pipeline status</a> for feature branch [PR_6506](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6506). Pipeline [492097097](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/492097097) was triggered at [1a8d6fe](https://github.com/FreeCAD/FreeCAD/pull/6506/commits/1a8d6fef3d554a6f33b86d7793f9e3ec8cd7599b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
We will have to postpone this to 0.21 to make 0.20 ready.

@realthunder maybe you have a fix or could review pending PRs for the 0.20 milestone?: https://github.com/FreeCAD/FreeCAD/milestone/2

chennes:
@realthunder if you are bored with working on the TNP mega-merge and want a break 😆, my impression of this PR is that it's ready for a final review once its conflicts are resolved.

realthunder:
Great. I'll work on it soon.

wwmayer:
@chennes I guess we all still agree in merging this PR? If yes, I will start to rebase it and fix the merge conflicts because this PR offers some functions that are needed in another PR.

chennes:
Yes, I think this is good functionality--go for it!

wwmayer:
Thanks, I will start with it now...

wwmayer:
#7771 replaces this PR.

PR #6513, opened 2022-03-05
Gui: fix property view synchronization of property status

realthunder:
Issue reported in https://forum.freecadweb.org/viewtopic.php?f=10&t=66810

### DIFF
--- a/src/Gui/propertyeditor/PropertyModel.cpp
+++ b/src/Gui/propertyeditor/PropertyModel.cpp
@@ -333,6 +333,8 @@ void PropertyModel::buildUp(const PropertyModel::PropertyList& props)
             // TODO: is it necessary to make sure the item has no pending commit?
             item->setPropertyData(jt->second);
         }
+        else
+            item->updateData();
     }
 
     // Third step, signal item insertion and movement.


luzpaz:
@realthunder is it worth adding comment above the code referencing this issue: `https://github.com/FreeCAD/FreeCAD/pull/6513`

realthunder:
I don't think it's necessary., because it won't help understanding why that missing call needs to be there. The more important one is the logic of PropertyModel incremental update ([PropertyModel::buildUp](https://github.com/FreeCAD/FreeCAD/blob/fe3feb30df8ec2ba22cafb0828c8bd5b00e4a465/src/Gui/propertyeditor/PropertyModel.cpp#L279)), which I have already put comments (not sure if that's enought, though). Incremental update means update existing on change instead of rebuild from scratch. The missing call is to account for updates when reusing an existing property item.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6513">pipeline status</a> for feature branch [PR_6513](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6513). Pipeline [485163255](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/485163255) was triggered at [39fbbdb](https://github.com/FreeCAD/FreeCAD/pull/6513/commits/39fbbdb58f854ea95e3eb2946e87f8e22be14e08). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
@WandererFan since you are the one who discovered the original regression, can you review?

WandererFan (approved):
This change fixes the problem where a status change was not applied to the property.

PR #6519, opened 2022-03-06
[App] add a missing return statement

donovaly:
as discussed here: https://github.com/FreeCAD/FreeCAD/commit/c96f61b0a1e3ab15cc2cb198fdbe58ce4a5d8c59#commitcomment-68066635

### DIFF
--- a/src/App/AutoTransaction.h
+++ b/src/App/AutoTransaction.h
@@ -32,7 +32,7 @@ class AppExport AutoTransaction {
 private:
     /// Private new operator to prevent heap allocation
     void* operator new(size_t size) {
-        return (void*)size;
+        return nullptr;
     };
 
 public:
@@ -123,7 +123,7 @@ public:
 private:
     /// Private new operator to prevent heap allocation
     void* operator new(size_t size) {
-        return (void*)size;
+        return nullptr;
     };
 
 private:
--- a/src/App/AutoTransaction.h
+++ b/src/App/AutoTransaction.h
@@ -31,7 +31,9 @@ class Application;
 class AppExport AutoTransaction {
 private:
     /// Private new operator to prevent heap allocation
-    void* operator new(size_t size);
+    void* operator new(size_t size) {
+        return (void*)size;
+    };
 
 public:
     /** Constructor
@@ -120,7 +122,9 @@ public:
 
 private:
     /// Private new operator to prevent heap allocation
-    void* operator new(size_t size);
+    void* operator new(size_t size) {
+        return (void*)size;
+    };
 
 private:
     bool active;


src/App/AutoTransaction.h
```diff
@@ -31,7 +31,9 @@ class Application;
 class AppExport AutoTransaction {
 private:
     /// Private new operator to prevent heap allocation
-    void* operator new(size_t size);
+    void* operator new(size_t size) {
+        return (void*)size;
```
chennes, line 35: I know it's academic since the point is that this can't be called, but it seems that just returning `nullptr` would be more idiomatic.

donovaly: nullptr is also what MSVC proposed but here: https://stackoverflow.com/questions/5909636/overloading-operator-new

someone wrote "The operator new should return a buffer of the size specified as the first argument."

chennes: Right -- if we were actually implementing this operator, it would return a buffer whose size is equal to the size passed in. Again, this is totally academic, the point of the code is that it can't be run, and trying to call it will give a compiler error about a private operator new. But it's a little jarring as a reviewer to see the location of size being returned.

donovaly: Fine with me of course. I changed it now to a nullptr. If this is now fine, please squash-merge.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6519">pipeline status</a> for feature branch [PR_6519](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6519). Pipeline [485593483](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/485593483) was triggered at [08530e2](https://github.com/FreeCAD/FreeCAD/pull/6519/commits/08530e25f6d96b003c7c9035fd35b70f545ecc77). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
OK, the CU run well and I return the requested nullptr -> merging.

src/App/AutoTransaction.h
wwmayer: I don't think you understood the essence of the link I posted. Making the new operator private only serves the purpose to disallow in client code to create an object on the heap and thus it suffices to **declare** it but a **definition** is not needed.

The error you got was because of changing the order of includes and thus the compiler got AutoTransaction.h as first that doesn't include a header file the pulls in the header where "size_t" is defined (because it's only a typedef of the actual type).

In the first answer of the SO question exactly this was said.

Now this PR makes things even worse because it creates hundreds of warnings:
```
AutoTransaction.h:35:9: warning: 'operator new' should not return a null pointer unless it is declared 'throw()' or 'noexcept' [-Wnew-returns-null]
AutoTransaction.h:34:31: warning: unused parameter 'size' [-Wunused-parameter]
AutoTransaction.h:126:9: warning: 'operator new' should not return a null pointer unless it is declared 'throw()' or 'noexcept' [-Wnew-returns-null]
AutoTransaction.h:125:31: warning: unused parameter 'size' [-Wunused-parameter]
```

donovaly: > Making the new operator private only serves the purpose to disallow in client code to create an object on the heap and thus it suffices to declare it but a definition is not needed

I understood the first part. But when no definition is needed, why does MSVC want one? Maybe I trust MSVC too much and MSVC proposed exactly what I merged.

> The error you got was because of changing the order of includes and thus the compiler got AutoTransaction.h as first that doesn't include a header file the pulls in the header where "size_t" is defined

Yes, that's why I realized I have to have Application.h as header before AutoTransaction.h. This part is already fixed. However, since MSVC kept saying me that it wants a definition block, I made this PR.

> Now this PR makes things even worse because it creates hundreds of warnings:

Sh... I don't get them. Do they disappear when you return `(void*)size`? If not, I will revert the change.

donovaly: I see them in the CI and reverted the PR.

wwmayer: > But when no definition is needed, why does MSVC want one?

What does it say exactly? And does it say for every cpp file that includes AutoTransaction.h directly or indirectly?

> Maybe I trust MSVC too much and MSVC proposed exactly what I merged.

Maybe yes. With newer versions it's not as bad as with older versions that reported many false-positives that we had to suppress with pragmas.

> Yes, that's why I realized I have to have Application.h as header before AutoTransaction.h. This part is already fixed.

And this is not needed because AutoTransation.h requires cstddef instead for the type size_t

donovaly: The warning only appear in AutoTransaction.h:
ReItjRrLHL

> And this is not needed because AutoTransation.h requires cstddef instead for the type size_t

Since the operator is defined within class Application I thought, Application.h is nevertheless necessary. However, I know my limitations of never learned C++, only learning by doing. Therefore the best is that I don't act here further.

---------------
Concerning my work with the headers. I am ready for FC 0.20 and also won't have tome to look at the yet unreviewed SoF* files in GUI the next days. I have 2 pending PRs because I am not sure if the CI compiler will bring errors or warnings.

wwmayer: > Since the operator is defined within class Application I thought, Application.h is nevertheless necessary.

No, it's not. In C++ it's possible to implement a new allocator for each class individually. And that you can see that the function is declared inside the class declaration.

Well, to me this MSVC hint is a false-positive because the code checker apparently doesn't realize the purpose of it. I wonder what the code checker says when you e.g. open the file Writer.h. Does it also complain about the lack of the implementation of the copy constructor and assignment operator of the class Writer?

donovaly: > Does it also complain about the lack of the implementation of the copy constructor and assignment operator of the class Writer?

No, it accepts all:
devenv_ZAYhS0akww

wwmayer: In order to make the disallowed use of the new function more explicit you could write:
`void* operator new (std::size_t) = delete;`

donovaly: > In order to make the disallowed use of the new function more explicit you could write: `void* operator new (std::size_t) = delete;`

This works. here is the PR: #6526

PR #6525, opened 2022-03-07
[Part] rename Section Cut feature

donovaly:
As discussed here: https://forum.freecadweb.org/viewtopic.php?p=577457#p577457
it would be an advantage to name the Section Cut feature with the prefix "Std_" despite it is technically a Part feature. I agree with @Roy-043.

In my experience users don't care about the internals, that see the feature in the View menu and will use it for assemblies.


wwmayer (commented):
See reply in forum

donovaly:
Closed in favor of #6531

PR #6534, opened 2022-03-08
[Sketcher] Fix "toggle internal geo" on multi-selection

AjinkyaDahale:
Some deleted GeoIds can stay in selection and GeoIds can change as some objects
are deleted.

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
+++ b/src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
@@ -983,50 +983,67 @@ void CmdSketcherRestoreInternalAlignmentGeometry::activated(int iMsg)
 
     getSelection().clearSelection();
 
+    // Return GeoId of the SubName only if it is an edge
+    auto getEdgeGeoId = [&Obj](const std::string& SubName) {
+        int GeoId;
+        Sketcher::PointPos PosId;
+        getIdsFromName(SubName, Obj, GeoId, PosId);
+        if (PosId == Sketcher::PointPos::none)
+            return GeoId;
+        else
+            return (int)GeoEnum::GeoUndef;
+    };
+
+    // Tells if the geometry with given GeoId has internal geometry
+    auto noInternalGeo = [&Obj](const auto& GeoId) {
+        const Part::Geometry *geo = Obj->getGeometry(GeoId);
+        bool hasInternalGeo = geo &&
+            (geo->getTypeId() == Part::GeomEllipse::getClassTypeId() ||
+             geo->getTypeId() == Part::GeomArcOfEllipse::getClassTypeId() ||
+             geo->getTypeId() == Part::GeomArcOfHyperbola::getClassTypeId() ||
+             geo->getTypeId() == Part::GeomArcOfParabola::getClassTypeId() ||
+             geo->getTypeId() == Part::GeomBSplineCurve::getClassTypeId());
+        return !hasInternalGeo; // so it's removed
+    };
+
+    std::vector<int> SubGeoIds(SubNames.size());
+    std::transform(SubNames.begin(), SubNames.end(), SubGeoIds.begin(), getEdgeGeoId);
+
+    // Handle highest GeoIds first to minimize GeoIds changing
+    // TODO: this might not completely resolve GeoIds changing
+    std::sort(SubGeoIds.begin(), SubGeoIds.end(), std::greater<int>());
+    // Keep unique
+    SubGeoIds.erase(std::unique(SubGeoIds.begin(), SubGeoIds.end()), SubGeoIds.end());
+
+    // Only for supported types and keep unique
+    SubGeoIds.erase(std::remove_if(SubGeoIds.begin(), SubGeoIds.end(), noInternalGeo),
+                    SubGeoIds.end());
+
     // go through the selected subelements
-    for (std::vector<std::string>::const_iterator it=SubNames.begin(); it != SubNames.end(); ++it) {
-        // only handle edges
-        if ((it->size() > 4 && it->substr(0,4) == "Edge") ||
-            (it->size() > 12 && it->substr(0,12) == "ExternalEdge")) {
-            int GeoId;
-            if (it->substr(0,4) == "Edge")
-               GeoId = std::atoi(it->substr(4,4000).c_str()) - 1;
-            else
-               GeoId = -std::atoi(it->substr(12,4000).c_str()) - 2;
+    for (const auto& GeoId : SubGeoIds) {
+        int currentgeoid = Obj->getHighestCurveIndex();
 
-            const Part::Geometry *geo = Obj->getGeometry(GeoId);
-            // Only for supported types
-            if (geo->getTypeId() == Part::GeomEllipse::getClassTypeId() ||
-                geo->getTypeId() == Part::GeomArcOfEllipse::getClassTypeId() ||
-                geo->getTypeId() == Part::GeomArcOfHyperbola::getClassTypeId() ||
-                geo->getTypeId() == Part::GeomArcOfParabola::getClassTypeId() ||
-                geo->getTypeId() == Part::GeomBSplineCurve::getClassTypeId()) {
+        try {
+            Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Exposing Internal Geometry"));
+            Gui::cmdAppObjectArgs(Obj, "exposeInternalGeometry(%d)", GeoId);
 
-                int currentgeoid = Obj->getHighestCurveIndex();
+            int aftergeoid = Obj->getHighestCurveIndex();
 
-                try {
-                    Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Exposing Internal Geometry"));
-                    Gui::cmdAppObjectArgs(Obj, "exposeInternalGeometry(%d)", GeoId);
-
-                    int aftergeoid = Obj->getHighestCurveIndex();
-
-                    if(aftergeoid == currentgeoid) { // if we did not expose anything, deleteunused
-                        Gui::cmdAppObjectArgs(Obj, "deleteUnusedInternalGeometry(%d)", GeoId);
-                    }
-                }
-                catch (const Base::Exception& e) {
-                    Base::Console().Error("%s\n", e.what());
-                    Gui::Command::abortCommand();
-
-                    tryAutoRecomputeIfNotSolve(static_cast<Sketcher::SketchObject *>(Obj));
-
-                    return;
-                }
-
-                Gui::Command::commitCommand();
-                tryAutoRecomputeIfNotSolve(static_cast<Sketcher::SketchObject *>(Obj));
+            if(aftergeoid == currentgeoid) { // if we did not expose anything, deleteunused
+                Gui::cmdAppObjectArgs(Obj, "deleteUnusedInternalGeometry(%d)", GeoId);
             }
         }
+        catch (const Base::Exception& e) {
+            Base::Console().Error("%s\n", e.what());
+            Gui::Command::abortCommand();
+
+            tryAutoRecomputeIfNotSolve(static_cast<Sketcher::SketchObject *>(Obj));
+
+            return;
+        }
+
+        Gui::Command::commitCommand();
+        tryAutoRecomputeIfNotSolve(static_cast<Sketcher::SketchObject *>(Obj));
     }
 }
 


AjinkyaDahale:
Forum post: https://forum.freecadweb.org/viewtopic.php?f=10&t=66937

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6534">pipeline status</a> for feature branch [PR_6534](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6534). Pipeline [487733651](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/487733651) was triggered at [89a606e](https://github.com/FreeCAD/FreeCAD/pull/6534/commits/89a606e51d9792374104679b84277e8d25e5ab17). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -983,50 +983,61 @@ void CmdSketcherRestoreInternalAlignmentGeometry::activated(int iMsg)
...
-               GeoId = -std::atoi(it->substr(12,4000).c_str()) - 2;
-
-            const Part::Geometry *geo = Obj->getGeometry(GeoId);
-            // Only for supported types
-            if (geo->getTypeId() == Part::GeomEllipse::getClassTypeId() ||
-                geo->getTypeId() == Part::GeomArcOfEllipse::getClassTypeId() ||
-                geo->getTypeId() == Part::GeomArcOfHyperbola::getClassTypeId() ||
-                geo->getTypeId() == Part::GeomArcOfParabola::getClassTypeId() ||
-                geo->getTypeId() == Part::GeomBSplineCurve::getClassTypeId()) {
+    std::vector<int> SubGeoIds(SubNames.size());
```
0penBrain, line 986: As you later do the elements unique and sort them "reversed", that would be probably be more efficient to use directly a `std::set` with `std::greater` as a compare operator.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -983,50 +983,61 @@ void CmdSketcherRestoreInternalAlignmentGeometry::activated(int iMsg)
...
+    std::transform(SubNames.begin(), SubNames.end(), SubGeoIds.begin(),
+                   [Obj](const std::string& SubName) {
+                       int GeoId;
+                       Sketcher::PointPos PosId;
+                       getIdsFromName(SubName, Obj, GeoId, PosId);
+                       return GeoId;
+                   });
```
0penBrain, line 993: Could you define the lambda out of the call (and add a small comment about what it does) to improve readability ?

Also I wonder if capturing `Obj` by reference won't be stack friendly (and overall improve performance) ???

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -983,50 +983,61 @@ void CmdSketcherRestoreInternalAlignmentGeometry::activated(int iMsg)
...
+                       [Obj](const auto& GeoId) {
+                           const Part::Geometry *geo = Obj->getGeometry(GeoId);
+                           bool hasInternalGeo = geo &&
+                               (geo->getTypeId() == Part::GeomEllipse::getClassTypeId() ||
+                                geo->getTypeId() == Part::GeomArcOfEllipse::getClassTypeId() ||
+                                geo->getTypeId() == Part::GeomArcOfHyperbola::getClassTypeId() ||
+                                geo->getTypeId() == Part::GeomArcOfParabola::getClassTypeId() ||
+                                geo->getTypeId() == Part::GeomBSplineCurve::getClassTypeId());
+                           return !hasInternalGeo; // so it's removed
+                       }),
```
0penBrain, line 1009: Same as above comment.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -983,50 +983,61 @@ void CmdSketcherRestoreInternalAlignmentGeometry::activated(int iMsg)
...
-                int currentgeoid = Obj->getHighestCurveIndex();
+    // go through the selected subelements
+    for (const auto& GeoId : SubGeoIds) {
+        int currentgeoid = Obj->getHighestCurveIndex();
 
-                try {
-                    Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Exposing Internal Geometry"));
-                    Gui::cmdAppObjectArgs(Obj, "exposeInternalGeometry(%d)", GeoId);
+        try {
+            Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Exposing Internal Geometry"));
```
0penBrain, line 1021: Having this here means you will create a transaction for each geometry, which will makes undoing horrible. :) Should be done before for loop.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
AjinkyaDahale: Will update the lambda in the next force push. `Obj` itself is a pointer so I don't understand expect much change, but capture by reference makes more sense, I suppose.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
AjinkyaDahale: This behavior is the same as before, though. Are you sure we want to change this?

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
0penBrain: No, you're right. As it's a pointer, AFAIK it will be captured as a reference to the value. ;)

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
0penBrain: Actually I didn't check. :) Just make a "reading" code review.
If I'm correct (and actually a transaction is created for each geometry), I think it makes no sense. Only one transaction shall be stored. ;)

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
AjinkyaDahale: I'm not sure I know how to do that. If you have a snippet I could try it out. I don't expect much difference in the two approaches but with set it might be cleaner.

As such far too often I expect only one element will be selected.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
AjinkyaDahale: https://github.com/FreeCAD/FreeCAD/blob/d1fc159f4c05feb455b324b4e32821817c30852a/src/Mod/Sketcher/Gui/CommandSketcherTools.cpp#L1008

AjinkyaDahale: @0penBrain you were correct that only one transaction is stored (though why I do not understand).

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
AjinkyaDahale: Changes made. I still went with `[&Obj]` even if it makes no difference.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
0penBrain: Here it is : https://github.com/0penBrain/FreeCAD/commit/43f948ac0f2f40f6e96226fc04376dd5a1f4e171

Sidenote : `std::erase_if` has only been introduced in C++20 so now using a check at beginning of for loop to keep compatibility with C++17

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6534">pipeline status</a> for feature branch [PR_6534](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6534). Pipeline [488398027](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/488398027) was triggered at [d6882f3](https://github.com/FreeCAD/FreeCAD/pull/6534/commits/d6882f3db120aab4e6008ee1245453540f201d00). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
AjinkyaDahale: That looks clean. if you want you can make a PR into my branch.

0penBrain: Done

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6534">pipeline status</a> for feature branch [PR_6534](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6534). Pipeline [489543192](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/489543192) was triggered at [f00cdae](https://github.com/FreeCAD/FreeCAD/pull/6534/commits/f00cdaeb23f5075b944bb715ad06839e50b79586). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
Reverted PR that uses `std::set` since it proved to be slower. Also fixed an issue where running unique before sort would only remove consecutive copies.

0penBrain:
@AjinkyaDahale from the GH diff, `sort` still is after `unique` ;)

AjinkyaDahale:
> @AjinkyaDahale from the GH diff, `sort` still is after `unique` ;)

Silly me. See now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6534">pipeline status</a> for feature branch [PR_6534](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6534). Pipeline [490119745](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/490119745) was triggered at [c52a3cc](https://github.com/FreeCAD/FreeCAD/pull/6534/commits/c52a3ccc868f82ae3e1250fa4849dc8af1ccbbb7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I can confirm that the PR fixes the crash. However, I haven't looked at the code.
@abdullahtahiriyo will have a look soon I think.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -983,50 +983,67 @@ void CmdSketcherRestoreInternalAlignmentGeometry::activated(int iMsg)
...
 
-                int currentgeoid = Obj->getHighestCurveIndex();
+    std::vector<int> SubGeoIds(SubNames.size());
+    std::transform(SubNames.begin(), SubNames.end(), SubGeoIds.begin(), getEdgeGeoId);
 
-                try {
-                    Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Exposing Internal Geometry"));
-                    Gui::cmdAppObjectArgs(Obj, "exposeInternalGeometry(%d)", GeoId);
+    // Handle highest GeoIds first to minimize GeoIds changing
+    // TODO: this might not completely resolve GeoIds changing
```
AjinkyaDahale, line 1013: A small observation here: it can be guaranteed that geo-ids don't change when the operation is done in descending order _if_ it is guaranteed that newly created b-splines do not use  pre-existing circles for poles. For now this is not guaranteed since we can always just add the internal constraint in python.

abdullahtahiriyo: Thanks. I understood the comment and I am aware of the limitation.

I think that the only way to overcome this limitation is to use the hash of the Part::Geometry to ensure that it has not changed, and if it has changed to retrieve the new geoid.

What do you think?
Feel free

PR #6556, opened 2022-03-12
[CXX] fix compiler warnings at MSVC

donovaly:
patch by @wwmayer: https://forum.freecadweb.org/viewtopic.php?p=578294#p578294

### DIFF
--- a/src/CXX/Python3/Objects.hxx
+++ b/src/CXX/Python3/Objects.hxx
@@ -1606,8 +1606,8 @@ namespace Py
             using iterator_category = std::random_access_iterator_tag;
             using value_type = const Object;
             using difference_type = int;
-            using pointer = const Object*;
-            using reference = const Object&;
+            using pointer = value_type*;
+            using reference = value_type&;
 
         protected:
             friend class SeqBase<T>;
--- a/src/CXX/Python3/Objects.hxx
+++ b/src/CXX/Python3/Objects.hxx
@@ -1442,8 +1442,14 @@ namespace Py
         }
 
         class PYCXX_EXPORT iterator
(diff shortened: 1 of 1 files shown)


wwmayer:
When building the smesh sources don't you get warnings because of SMDS_StdIterator?

For the patch I realized the pointer and reference types of const_iterator should be changed to:
```
            using pointer = value_type*;
            using reference = value_type&;
```
just like for iterator. This is how it's usually done for type definitions of template classes.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6556">pipeline status</a> for feature branch [PR_6556](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6556). Pipeline [490566676](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/490566676) was triggered at [a82befe](https://github.com/FreeCAD/FreeCAD/pull/6556/commits/a82befec565557a1088625ce1952777cb240f9a8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> When building the smesh sources don't you get warnings because of SMDS_StdIterator?

No. Therefore I did not include this to the PR. However, your suggested change to SMESH also works and created no errors or warnings. Since my knowledge is limited in this respect, I cannot do more than testing your suggestions.

donovaly:
> For the patch I realized the pointer and reference types of const_iterator should be changed to:

Done, tested and committed.

wwmayer:
4547e989ea

PR #6578, opened 2022-03-13
Redraw constraints on wheel event

abdullahtahiriyo:
https://forum.freecadweb.org/viewtopic.php?p=578857#p578857


wwmayer (approved):
All fine for me.

PR #6596, opened 2022-03-17
Retire qassistant

yorikvanhavre:
This PR disables the use of QAssistant to display help pages, which now relies on the Help addon.

If the Help module is present, it is used to handle the display of help pages. If not, an info message is displayed to inform the user of what they must do. The interpreter loadModule() function has been modified to allow to not raise an error if a module cannot be loaded, so we can use it to poke if a module is loadable or not.

Bulky QHelp files have been removed from the sources. We should still remove them form the history, but that requires a history rewrite, and it's better to plan that better before doing it.

The assistant.cpp & assistant.h in src/Gui can also be removed, but I'm not sure it's a good idea to make that entirely vanish, be it for history/documentation only? Same with the src/Tools/offline scripts, these can be removed but might be also kept somewhere as they could be useful to others... In any case they don't cause harm so for the time being I kept them here.

@wwmayer @chennes I marked you guys for review, but it's just for you guys to be aware of the changes and give a chance to object if you feel so ;) Tell me if the idea looks OK.

Forum discussion at https://forum.freecadweb.org/viewtopic.php?p=578362#p578362


adrianinsaval:
when you say rewrite history do you mean a force push onto master removing the files from the commit history? (what about the other branches?) If so would this be a good time to carry out the line end homogenization that was once proposed but rejected because a force push was needed, or for including the old svn history as was proposed as well? Or are those more problematic than this?

wwmayer:
> We should still remove them form the history, but that requires a history rewrite, and it's better to plan that better before doing it.

Why is this needed? Does it shrink the size of the local git repo or what's the point?

> or for including the old svn history as was proposed as well? 

I wouldn't do that because 99% of all devs don't need it anyway. Bernd has a branch on his repo for that purpose.

wwmayer (changes requested):
See comments below...

luzpaz:
> https://forum.freecadweb.org/viewtopic.php?f=8&t=67198

I don't have much say here admittedly, but it would be cool to have the whole FreeCAD codebase in its entirety in one place

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6596">pipeline status</a> for feature branch [PR_6596](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6596). Pipeline [495111158](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/495111158) was triggered at [8d154d6](https://github.com/FreeCAD/FreeCAD/pull/6596/commits/8d154d6498c5589aeb163622d6f24ae4d862f4d3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
> See comments below...

I don't actually see any comments -- did they get lost?

src/Gui/MainWindow.cpp
```diff
@@ -692,18 +692,18 @@ void MainWindow::whatsThis()
...
         QUrl url(help);
         if (url.scheme().isEmpty()) {
-            QString page;
-            page = QString::fromUtf8("%1.html").arg(help);
-            d->assistant->showDocumentation(page);
+            //QString page;
+            //page = QString::fromUtf8("%1.html").arg(help);
+            //d->assistant->showDocumentation(page);
+            QMessageBox::critical(getMainWindow(), tr("Help addon needed!"),
+            tr("The Help system of FreeCAD is now handled by the \"Help\" addon."
```
chennes, line 705: Missing space:
```suggestion
            tr("The Help system of FreeCAD is now handled by the \"Help\" addon. "
```
(Does the C++ lupdate extraction work correctly with strings formed like this?)

src/Gui/MainWindow.cpp
```diff
@@ -692,18 +692,18 @@ void MainWindow::whatsThis()
...
+                "Install it with menu Tools > Addons Manager"));
```
chennes, line 706: What do you think about a button that opens the Addon Manager, possibly even with a filter on "Help"?

yorikvanhavre: Excellent idea! How would I start the addons manager that way?

chennes: Good question... I thought it was possible to get a reference to the actual `CommandAddonManager` class using `Command.get("Std_AddonMgr")`, but it doesn't work the way I thought it did. There are some cheesy ways of doing it (by directly setting some of Addon Manager's preferences, which it uses to store the last filter information), but that's pretty fragile, and has your addon digging into the internals of the Addon Manager, which is obviously not a good idea. Do you know how/if it's possible to get the actual python object that is the instance of class `CommandAddonManager`?

yorikvanhavre: All this has no python implementation (yet) i think... probaply better to hack something for now.. but it's not super urgent, it works already, just less smooth

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6596">pipeline status</a> for feature branch [PR_6596](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6596). Pipeline [496115049](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/496115049) was triggered at [aad4e77](https://github.com/FreeCAD/FreeCAD/pull/6596/commits/aad4e7768bb301ccd5678fc81eeb9ed01c0aa7cf). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> I don't actually see any comments -- did they get lost?

They are still there. I added them as "Start Review". There were two comments:
1. about Interpreter().loadModule: No need to change the function and use this instead:
```
PyObject* module = PyImport_ImportModule("Help");
if (!module) {
    PyErr_Clear();
...
}
```
In case it's preferred to change loadModule() then the second argument shouldn't be a boolean but an enum class instead. This is to make sure about the purpose in client code.
2. To do branding right _FreeCAD_ shouldn't be directly used. Instead use `qApp->applicationName()`

yorikvanhavre:
I did some serious shit on this branch I think! :sweat_smile: I'll close it and reopen another PR

yorikvanhavre:
Reopened the PR in #6634

PR #6601, opened 2022-03-18
[Macros] change text of Addons... button to Download to help new people

adrianinsaval:
It may not be immediately clear to some newbies that the Addons... button is used to download macros. Suggested among the flame war in https://forum.freecadweb.org/viewtopic.php?p=580097#p580083
also added a couple of tooltips, the rest of the buttons are self explanatory IMO so I didn't add tooltips.

### DIFF
--- a/src/Gui/DlgMacroExecute.ui
+++ b/src/Gui/DlgMacroExecute.ui
@@ -279,6 +279,9 @@
          <property name="enabled">
           <bool>false</bool>
          </property>
+         <property name="toolTip">
+          <string>Launch a guide on how to set up a macro in a custom global toolbar.</string>
+         </property>
          <property name="text">
           <string>Toolbar</string>
          </property>
@@ -302,8 +305,11 @@
          <property name="enabled">
           <bool>true</bool>
          </property>
+         <property name="toolTip">
+          <string>Open Addon Manager where macros created by the community and other addons can be downloaded.</string>
+         </property>
          <property name="text">
-          <string>Addons...</string>
+          <string>Download</string>
          </property>
         </widget>
        </item>


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6601">pipeline status</a> for feature branch [PR_6601](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6601). Pipeline [495275042](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/495275042) was triggered at [2468579](https://github.com/FreeCAD/FreeCAD/pull/6601/commits/2468579d01af29321d633b6e6edc5fa4434f662a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/DlgMacroExecute.ui
```diff
@@ -302,8 +305,11 @@
          <property name="enabled">
           <bool>true</bool>
          </property>
+         <property name="toolTip">
+          <string>Open Addon Manager were macros created by the community and other addons can be downloaded.</string>
```
chennes, line 309:
```suggestion
          <string>Open Addon Manager where macros created by the community and other addons can be downloaded.</string>
```

adrianinsaval: done

adrianinsaval:
If button name change is rejected I think the tool tips should go in anyways. I can of course push a commit for just that if it comes to that.

chennes:
I have no objection to the change. If no one else minds I'll merge it later today.

adrianinsaval:
@chennes would it be possible to launch the addon manager directly showing the macros category from this dialog, this would be more intuitive IMO

chennes:
Yes, if you set the user preference integer `PackageTypeSelection` in `BaseApp/Preferences/Addons` to 2, when the AM launches it will display Macros.

adrianinsaval:
but that would be globally then, right? I meant specifically when launched with that button.

chennes:
That variable is used to "remember" the user's last selection: it gets set whenever the user changes their view, so setting it before launching the AM is exactly equivalent to letting the AM launch as before, and having the user then change the view to Macros.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6601">pipeline status</a> for feature branch [PR_6601](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6601). Pipeline [496447119](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/496447119) was triggered at [227c717](https://github.com/FreeCAD/FreeCAD/pull/6601/commits/227c7178965d1b14fc7789cc043d37be89fed195). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@chennes , I think we should decide right now if we do the renaming or not. You wrote you will merge. If yes, can you please do so today?

chennes:
Yeah, let's go ahead with it, I think @adrianinsaval is right, and this is clearer.

PR #6615, opened 2022-03-20
[FEM] Refactor `TaskFemConstraintOnBoundary`

AjinkyaDahale:
Now the behavior is (slightly more) consistent with behavior of PD fillets, for example.
See https://forum.freecadweb.org/viewtopic.php?f=18&t=67135#p580192. In the
future it may be possible to reuse some code from there.

What isn't yet implemented:
- If opening a constraint with selection, the constraint is not applied to that selection already. I expect this will need to be done for each of the subclasses. Or possibly it can be done in the superclass itself with some NVI?
- The last element can be deleted. We may want to keep this behavior in case the user wants to select, say, edges when some faces are already selected.

### DIFF
--- a/src/Mod/Fem/Gui/ViewProviderFemConstraintOnBoundary.cpp
+++ b/src/Mod/Fem/Gui/ViewProviderFemConstraintOnBoundary.cpp
@@ -56,16 +56,14 @@ void ViewProviderFemConstraintOnBoundary::highlightReferences(const bool on)
     const auto& subSets = pcConstraint->References.getSubListValues();
 
     for (auto& subSet : subSets) {
-        // if somehow the subnames are empty, we have nothing to do
-        if (subSet.second.empty())
-            continue;
-
-        Part::Feature* base = static_cast<Part::Feature*>(subSet.first);
+        Part::Feature* base = dynamic_cast<Part::Feature*>(subSet.first);
+        if (!base) continue;
         PartGui::ViewProviderPart* vp = dynamic_cast<PartGui::ViewProviderPart*>(
             Gui::Application::Instance->getViewProvider(base));
-        if (vp == nullptr) continue;
+        if (!vp) continue;
 
-        if (on) {
+        // if somehow the subnames are empty, clear any existing colors
+        if (on && !subSet.second.empty()) {
             // identify the type of subelements
             // TODO: Assumed here the subelements are of the same type.
             // It is a requirement but we should keep safeguards.
@@ -107,21 +105,52 @@ void ViewProviderFemConstraintOnBoundary::highlightReferences(const bool on)
             }
         }
         else {
-            if (subSet.second[0].find("Vertex") != std::string::npos &&
-                !originalPointColors[base].empty()) {
+            if (!originalPointColors[base].empty()) {
                 vp->PointColorArray.setValues(originalPointColors[base]);
                 originalPointColors[base].clear();
             }
-            else if (subSet.second[0].find("Edge") != std::string::npos &&
-                     !originalLineColors[base].empty()) {
+            else if (!originalLineColors[base].empty()) {
                 vp->LineColorArray.setValues(originalLineColors[base]);
                 originalLineColors[base].clear();
             }
-            else if (subSet.second[0].find("Face") != std::string::npos &&
-                     !originalFaceColors[base].empty()) {
+            else if (!originalFaceColors[base].empty()) {
                 vp->DiffuseColor.setValues(originalFaceColors[base]);
                 originalFaceColors[base].clear();
             }
+        }
+    }
+
+    if (subSets.empty()) {
+        // there is nothing selected but previous selection may have highlighting
+        // reset that highlighting here
+        for (auto& ogPair : originalPointColors) {
+            if (ogPair.second.empty()) continue;
+            PartGui::ViewProviderPart* vp = dynamic_cast<PartGui::ViewProviderPart*>(
+                Gui::Application::Instance->getViewProvider(ogPair.first));
+            if (!vp) continue;
+
+            vp->PointColorArray.setValues(ogPair.second);
+            ogPair.second.clear();
+        }
+
+        for (auto& ogPair : originalLineColors) {
+            if (ogPair.second.empty()) continue;
+            PartGui::ViewProviderPart* vp = dynamic_cast<PartGui::ViewProviderPart*>(
+                Gui::Application::Instance->getViewProvider(ogPair.first));
+            if (!vp) continue;
+
+            vp->LineColorArray.setValues(ogPair.second);
+            ogPair.second.clear();
+        }
+
+        for (auto& ogPair : originalFaceColors) {
+            if (ogPair.second.empty()) continue;
+            PartGui::ViewProviderPart* vp = dynamic_cast<PartGui::ViewProviderPart*>(
+                Gui::Application::Instance->getViewProvider(ogPair.first));
+            if (!vp) continue;
+
+            vp->DiffuseColor.setValues(ogPair.second);
+            ogPair.second.clear();
         }
     }
 }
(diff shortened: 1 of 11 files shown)


donovaly:
Good work! I stress-tested it and it works nicely. The PR fixes the issue I reported. So when the CI succeeds, this can be merged.

> If opening a constraint with selection, the constraint is not applied to that selection already. I expect this will need to be done for each of the subclasses. Or possibly it can be done in the superclass itself with some NVI?

That's what I meant with taking existing code from the fillet dialog ;-) The preselection is already implemented there.
But no problem. This can be added anytime later.

donovaly (approved):

AjinkyaDahale:
> That's what I meant with taking existing code from the fillet dialog ;-) The preselection is already implemented there. But no problem. This can be added anytime later.

I found some significant differences between the codes of fillet/other similar PD features' and FEM constraints' tasks. Also, there's a significant difference that the model itself changes for PD stuff, but for FEM stuff it remains the same, just changing what's highlighted and what's not.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6615">pipeline status</a> for feature branch [PR_6615](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6615). Pipeline [496447207](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/496447207) was triggered at [35b7a4e](https://github.com/FreeCAD/FreeCAD/pull/6615/commits/35b7a4e46bede5a16e2027bffa1c7e377d5a5de0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Fem/Gui/ViewProviderFemConstraintOnBoundary.cpp
```diff
@@ -56,16 +56,13 @@ void ViewProviderFemConstraintOnBoundary::highlightReferences(const bool on)
     const auto& subSets = pcConstraint->References.getSubListValues();
 
     for (auto& subSet : subSets) {
-        // if somehow the subnames are empty, we have nothing to do
-        if (subSet.second.empty())
-            continue;
-
         Part::Feature* base = static_cast<Part::Feature*>(subSet.first);
```
wwmayer, line 59: This static_cast is problematic because there is no guarantee that the objects in the References property are all Part::Feature. A dynamic_cast and a null check is needed here.

src/Mod/Fem/Gui/ViewProviderFemConstraintOnBoundary.cpp
```diff
@@ -107,22 +104,56 @@ void ViewProviderFemConstraintOnBoundary::highlightReferences(const bool on)
...
             }
         }
     }
+
+    if (subSets.empty()) {
+        // there is nothing selected but previous selection may have highlighting
+        // reset that highlighting here
+        for (auto& ogPair : originalPointColors) {
+            if (ogPair.second.empty()) continue;
+            Part::Feature* base = static_cast<Part::Feature*>(ogPair.first);
```
wwmayer, line 127: Here no static_cast is needed because the key of the originalPointColors map already is a Part::Feature.

src/Mod/Fem/Gui/ViewProviderFemConstraintOnBoundary.cpp
```diff
@@ -107,22 +104,56 @@ void ViewProviderFemConstraintOnBoundary::highlightReferences(const bool on)
...
+                Gui::Application::Instance->getViewProvider(base));
+            if (vp == nullptr) continue;
+
+            vp->PointColorArray.setValues(ogPair.second);
+            ogPair.second.clear();
+        }
+
+        for (auto& ogPair : originalLineColors) {
+            if (ogPair.second.empty()) continue;
+            Part::Feature* base = static_cast<Part::Feature*>(ogPair.first);
```
wwmayer, line 138: Same here

src/Mod/Fem/Gui/ViewProviderFemConstraintOnBoundary.cpp
```diff
@@ -107,22 +104,56 @@ void ViewProviderFemConstraintOnBoundary::highlightReferences(const bool on)
...
+                Gui::Application::Instance->getViewProvider(base));
+            if (vp == nullptr) continue;
+
+            vp->LineColorArray.setValues(ogPair.second);
+            ogPair.second.clear();
+        }
+
+        for (auto& ogPair : originalFaceColors) {
+            if (ogPair.second.empty()) continue;
+            Part::Feature* base = static_cast<Part::Feature*>(ogPair.first);
```
wwmayer, line 149: Same here

src/Mod/Fem/Gui/ViewProviderFemConstraintOnBoundary.cpp
AjinkyaDahale: Check added.

src/Mod/Fem/Gui/ViewProviderFemConstraintOnBoundary.cpp
AjinkyaDahale: Done.

AjinkyaDahale:
@donovaly @wwmayer I had added another change since your comments, to use a `Gui::ButtonGroup` to handle the add/remove switch. If you're not comfortable with it yet we can drop that for now. @wwmayer's changes are rebased onto the previous commit.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6615">pipeline status</a> for feature branch [PR_6615](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6615). Pipeline [496539293](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/496539293) was triggered at [c47997b](https://github.com/FreeCAD/FreeCAD/pull/6615/commits/c47997b952d51f1ebc73e190cc1b85c57589a5f8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@wwmayer thanks for the review, I did not noticed them. I would only see this from inside an IDE. So next time I will not just look at the written code.

@AjinkyaDahale many thanks. I will test it now again.

wwmayer:
Code looks good so far. However, I haven't tested the changes. So if everything works as expected then go ahead and merge it.

donovaly:
I tested and found only a minor issue:

- directly after removing a surface, it is still highlighted (until I leave the Remove mode). Ideally it should directly become unhighlighted after you clicked it (and this way removed it):
FreeCAD_sJmZQK7c5K


However, this can be addressed in another PR or even postponed to FC 0.21.

Merging this one now.

AjinkyaDahale:
>     * directly after removing a surface, it is still highlighted (until I leave the Remove mode). Ideally it should directly become unhighlighted after you clicked it (and this way removed it):
>       FreeCAD_sJmZQK7c5K

What is happening is that the face remains selected (light green). Can you try clicking outside the object in the viewport. I imagine that should disappear.

donovaly:
> What is happening is that the face remains selected (light green). 

Yes, that is what I mean. When I click somewhere in the 3D void, the removed surface looses its highlighting. However, I thought it should no longer be highlighted after one clicked at it to remove it since is is then just a normal surface.
But maybe there are good reasons to keep it highlighted?

AjinkyaDahale:
> But maybe there are good reasons to keep it highlighted?

I don't expect there to be much confusion, since the color is exactly that of any normal selection. If it were, say, a shade of pink (i.e. reference), or the color didn't go away, that would be a cause for concern. It might be something to revisit after we have unified the PD and any other such tasks.

donovaly:
> since the color is exactly that of any normal selection.

Yes, but after the removal, the surface is just a normal one. Why should it be highlighted anymore? At least for me this was unintuitive.

AjinkyaDahale:
> > since the color is exactly that of any normal selection.
> 
> Yes, but after the removal, the surface is just a normal one. Why should it be highlighted anymore? At least for me this was unintuitive.

I don't have a preference either way, so we can add a `clearSelection()` at the end. With the select-then-add/remove options it made a little bit of a sense since then one can immediately click remove/add to revert the choice. Without it there doesn't appear to be much difference.

PR #6628, opened 2022-03-20
[JtReader] remove unused headers and Xerces

donovaly:


donovaly:
Closed in favor of https://github.com/FreeCAD/FreeCAD/commit/04ff5ae34 since JtReader is no longer used anyway

PR #6632, opened 2022-03-21
Path: Block operation recompute on task panel cancel [Bug]

Russ4262:
This fix applies to standard operations and most dressups.  I think that the fix applied here must be applied to all other ViewProvider enabled actions if the cancel button is to block a recompute for those actions.  The implementation is fairly simple.  This implementation is triggers the cancel only on the GUI side.  There is not really an intended use for cancelling recomputes in scripting or command-line usage of FreeCAD.

This fix effectively flags the `execute()` method of the Proxy to return prematurely before updating any `obj` properties, leaving the source `obj` unchanged from its state prior to the aborted edit transaction started and ended in the task panel.

This is a long-standing bug or nuisance, especially with the often long-lasting 3D Surface computations.

Forum mention at:
* [3Dsurface still recomputes on Cancel.](https://forum.freecadweb.org/viewtopic.php?f=15&t=67326&sid=29b613681f528b052ca058c458ee3c7b)
* [Problem on millface and adaptive clearance. Stops responding](https://forum.freecadweb.org/viewtopic.php?f=15&t=64753)
* [Paths falsely marked as altered.](https://forum.freecadweb.org/viewtopic.php?f=15&t=63993)
* [Couple of potential bugs...](https://forum.freecadweb.org/viewtopic.php?f=15&t=24881&start=20)

Again, this PR does not fix the issue for all actions in Path workbench, just the standard operations based on PathOp & PathOpGui modules.  Dressups and other actions will have to be fixed individually.

Thanks to @J-Dunn, @mlampert, and many others for mentioning this bug on numerous occasions.  I think this PR will be the beginning of the end for recomputes on cancel in task panels!

Steps to reproduce bug:
* Create a big part, or small part if you have 3D Surface active.
* Create an operation (3D Surface or Adaptive work great for this) that takes a while (10 - 30 seconds) to compute, and save it.
* Double click on the operation in the object tree to edit using the task panel.
* Make changes if you like and apply them, but that might take some time (haha!).  ... Or do not for this example.
* Click the Cancel button.
* The operation will recompute upon exiting the task panel.

---

### DIFF
--- a/src/Mod/Path/PathScripts/PathDressupAxisMap.py
+++ b/src/Mod/Path/PathScripts/PathDressupAxisMap.py
@@ -109,6 +109,10 @@ class ObjectDressup:
 
     def execute(self, obj):
 
+        if PathUtils.isDressupCancelled(obj):
+            PathLog.debug("Axis Map Dressup cancelled")
+            return
+
         inAxis = obj.AxisMap[0]
         outAxis = obj.AxisMap[3]
         d = 0.1
@@ -176,6 +180,8 @@ class TaskPanel:
 
     def reject(self):
         FreeCAD.ActiveDocument.abortTransaction()
+        # Set flag to cancel dressup execution
+        PathUtils.cancelExecution(self.obj.Name)
         FreeCADGui.Control.closeDialog()
         FreeCAD.ActiveDocument.recompute()
 
--- a/src/Mod/Path/PathScripts/PathDressupDogbone.py
+++ b/src/Mod/Path/PathScripts/PathDressupDogbone.py
@@ -887,6 +887,10 @@ class ObjectDressup(object):
         return commands, bones
 
     def execute(self, obj, forReal=True):
+        if PathUtils.isDressupCancelled(obj):
+            PathLog.debug("Dogbone Dressup cancelled")
+            return
+
         if not obj.Base:
             return
         if forReal and not obj.Base.isDerivedFrom("Path::Feature"):
@@ -1133,6 +1137,8 @@ class TaskPanel(object):
 
     def reject(self):
         FreeCAD.ActiveDocument.abortTransaction()
+        # Set flag to cancel dressup execution
+        PathUtils.cancelExecution(self.obj.Name)
         FreeCADGui.Control.closeDialog()
         FreeCAD.ActiveDocument.recompute()
         FreeCADGui.Selection.removeObserver(self.s)
--- a/src/Mod/Path/PathScripts/PathDressupDragknife.py
+++ b/src/Mod/Path/PathScripts/PathDressupDragknife.py
@@ -373,6 +373,10 @@ class ObjectDressup:
         return (results, replace)
 
     def execute(self, obj):
+        if PathUtils.isDressupCancelled(obj):
+            # PathLog.debug("Dragknife Dressup cancelled")  # PathLog module unavailable
+            return
+
         newpath = []
         global currLocation
 
@@ -512,6 +516,8 @@ class TaskPanel:
 
     def reject(self):
         FreeCAD.ActiveDocument.abortTransaction()
+        # Set flag to cancel dressup execution
+        PathUtils.cancelExecution(self.obj.Name)
         FreeCADGui.Control.closeDialog()
         FreeCAD.ActiveDocument.recompute()
 
--- a/src/Mod/Path/PathScripts/PathDressupHoldingTags.py
+++ b/src/Mod/Path/PathScripts/PathDressupHoldingTags.py
@@ -1230,6 +1230,10 @@ class ObjectTagDressup:
         return (tags, positions, disabled)
 
     def execute(self, obj):
+        if PathUtils.isDressupCancelled(obj):
+            PathLog.debug("Holding Tags Dressup cancelled")
+            return
+
         # import cProfile
         # pr = cProfile.Profile()
         # pr.enable()
--- a/src/Mod/Path/PathScripts/PathDressupLeadInOut.py
+++ b/src/Mod/Path/PathScripts/PathDressupLeadInOut.py
@@ -179,6 +179,10 @@ class ObjectDressup:
         obj.IncludeLayers = True
 
     def execute(self, obj):
+        if PathUtils.isDressupCancelled(obj):
+            PathLog.debug("LeadInOut Dressup cancelled")
+            return
+
         if not obj.Base:
             return
         if not obj.Base.isDerivedFrom("Path::Feature"):
@@ -633,6 +637,21 @@ class ObjectDressup:
 class TaskDressupLeadInOut(SimpleEditPanel):
     _transaction_name = "Edit LeadInOut Dress-up"
     _ui_file = ":/panels/DressUpLeadInOutEdit.ui"
+
+    def abort(self):
+        """abort() overwritten here to include task panel cancel action"""
+        FreeCAD.ActiveDocument.abortTransaction()
+        # Set flag to cancel dressup execution
+        PathUtils.cancelExecution(self.obj.Name)
+        self.cleanup(True)
+
+    def reject(self):
+        """reject() overwritten here to include task panel cancel action"""
+        FreeCAD.ActiveDocument.abortTransaction()
+        # Set flag to cancel dressup execution
+        PathUtils.cancelExecution(self.obj.Name)
+        FreeCADGui.Control.closeDialog()
+        FreeCAD.ActiveDocument.recompute()
 
     def setupUi(self):
         self.connectWidget("LeadIn", self.form.chkLeadIn)
--- a/src/Mod/Path/PathScripts/PathDressupPathBoundary.py
+++ b/src/Mod/Path/PathScripts/PathDressupPathBoundary.py
@@ -103,6 +103,10 @@ class DressupPathBoundary(object):
         return True
 
     def execute(self, obj):
+        if PathUtils.isDressupCancelled(obj):
+            PathLog.debug("Boundary Dressup cancelled")
+            return
+
         pb = PathBoundary(obj.Base, obj.Stock.Shape, obj.Inside)
         obj.Path = pb.execute()
 
--- a/src/Mod/Path/PathScripts/PathDressupPathBoundaryGui.py
+++ b/src/Mod/Path/PathScripts/PathDressupPathBoundaryGui.py
@@ -36,6 +36,7 @@ else:
 
 
 translate = FreeCAD.Qt.translate
+PathUtils = PathDressupPathBoundary.PathUtils
 
 
 class TaskPanel(object):
@@ -88,6 +89,8 @@ class TaskPanel(object):
 
     def reject(self):
         FreeCAD.ActiveDocument.abortTransaction()
+        # Set flag to cancel dressup execution
+        PathUtils.cancelExecution(self.obj.Name)
         self.cleanup(True)
 
     def accept(self):
--- a/src/Mod/Path/PathScripts/PathDressupTagGui.py
+++ b/src/Mod/Path/PathScripts/PathDressupTagGui.py
@@ -69,7 +69,7 @@ class PathDressupTagTaskPanel:
             self.jvoVisible = jvoVisibility
         self.pt = FreeCAD.Vector(0, 0, 0)
 
-        self.isDirty = True
+        self.isDirty = False  # True
         self.buttonBox = None
         self.tags = None
         self.Positions = None
@@ -95,10 +95,16 @@ class PathDressupTagTaskPanel:
 
     def abort(self):
         FreeCAD.ActiveDocument.abortTransaction()
+        if self.isDirty:
+            # Set flag to cancel dressup execution
+            PathUtils.cancelExecution(self.obj.Name)
         self.cleanup(False)
 
     def reject(self):
         FreeCAD.ActiveDocument.abortTransaction()
+        if self.isDirty:
+            # Set flag to cancel dressup execution
+            PathUtils.cancelExecution(self.obj.Name)
         self.cleanup(True)
 
     def accept(self):
--- a/src/Mod/Path/PathScripts/PathDressupZCorrect.py
+++ b/src/Mod/Path/PathScripts/PathDressupZCorrect.py
@@ -154,6 +154,10 @@ class ObjectDressup:
 
     def execute(self, obj):
 
+        if PathUtils.isDressupCancelled(obj):
+            PathLog.debug("Z Correct Dressup cancelled")
+            return
+
         sampleD = obj.SegInterpolate.Value
         curveD = obj.ArcInterpolate.Value
 
@@ -237,6 +241,8 @@ class TaskPanel:
 
     def reject(self):
         FreeCAD.ActiveDocument.abortTransaction()
+        # Set flag to cancel dressup execution
+        PathUtils.cancelExecution(self.obj.Name)
         FreeCADGui.Control.closeDialog()
         FreeCAD.ActiveDocument.recompute()
 
--- a/src/Mod/Path/PathScripts/PathJob.py
+++ b/src/Mod/Path/PathScripts/PathJob.py
@@ -105,6 +105,7 @@ Notification = NotificationClass()
 
 class ObjectJob:
     def __init__(self, obj, models, templateFile=None):
+        FreeCAD.pathCancelExecution = []
         self.obj = obj
         self.tooltip = None
         self.tooltipArgs = None
@@ -483,6 +484,7 @@ class ObjectJob:
                     ops.Label = label
 
     def onDocumentRestored(self, obj):
+        FreeCAD.pathCancelExecution = []
         self.setupBaseModel(obj)
         self.fixupOperations(obj)
         self.setupSetupSheet(obj)
--- a/src/Mod/Path/PathScripts/PathJobGui.py
+++ b/src/Mod/Path/PathScripts/PathJobGui.py
@@ -620,6 +620,7 @@ class TaskPanel:
             self.obj, self.form.jobBox.widget(1)
         )
         self.name = self.obj.Name
+        self.doNotRecomputeOps = False
 
         vUnit = FreeCAD.Units.Quantity(1, FreeCAD.Units.Velocity).getUserPreferred()[2]
         self.form.toolControllerList.horizontalHeaderItem(1).setText("#")
@@ -725,6 +726,9 @@ class TaskPanel:
 
     def cleanup(self, resetEdit):
         PathLog.track()
+        if self.doNotRecomputeOps:
+            self.doNotRecomputeOps = False
+            PathUtils.cancelAllOps(self.obj)
         FreeCADGui.Control.closeDialog()
         if resetEdit:
             FreeCADGui.ActiveDocument.resetEdit()
@@ -1024,6 +1028,8 @@ class TaskPanel:
                     name=tool.Label, tool=tool, toolNumber=toolNum
                 )
                 self.obj.Proxy.addToolController(tc)
+                # Cancel recompute for all ops, since new tool controller does not affect ops
+                # self.doNotRecomputeOps = True
 
             FreeCAD.ActiveDocument.recompute()
             self.updateToolController()
--- a/src/Mod/Path/PathScripts/PathOp.py
+++ b/src/Mod/Path/PathScripts/PathOp.py
@@ -767,6 +767,10 @@ class ObjectOp(object):
         """
         PathLog.track()
 
+        if PathUtils.isOperationCancelled(obj):
+            PathLog.debug("Operation cancelled")
+            return
+
         if not obj.Active:
             path = Path.Path("(inactive operation)")
             obj.Path = path
--- a/src/Mod/Path/PathScripts/PathOpGui.py
+++ b/src/Mod/Path/PathScripts/PathOpGui.py
@@ -1211,6 +1211,9 @@ class TaskPanel(object):
             except Exception as ee:
                 PathLog.debug("{}\n".format(ee))
             FreeCAD.ActiveDocument.commitTransaction()
+        else:
+            PathUtils.cancelExecution(self.obj.Name)
+
         self.cleanup(resetEdit)
         return True
 
--- a/src/Mod/Path/PathScripts/PathUtils.py
+++ b/src/Mod/Path/PathScripts/PathUtils.py
@@ -827,7 +827,7 @@ def RtoIJ(startpoint, command):
     perp = chord.cross(Vector(0, 0, 1))
 
     # use pythagoras to get the perp length
-    plength = math.sqrt(radius**2 - (chord.Length / 2) ** 2)
+    plength = math.sqrt(radius ** 2 - (chord.Length / 2) ** 2)
     perp.normalize()
     perp.scale(plength, plength, plength)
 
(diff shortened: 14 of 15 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [496806791](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/496806791) was triggered at [4038d71](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/4038d7111f2463f5ef2334a3d1a34db1887d7d86). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

J-Dunn:
Many thanks for picking this up. This is a major PITA with the excellent 3D paths some of which take several minutes to compute.

Russ4262:
I just pushed a fix for the command-line compatibility, which is what was causing the failure.

Russ4262:
I changed up the method used to block the recompute.  This new method seems more in line with current GUI callback methods and is simpler to implement moving forward.

I also added blockage of recomputes to two dressups and the job object.  However, the job object is unlike the operation object in how it handles execution of all the related sub-elements.  It just doesn't as far as I can tell looking at the code base.  Therefore, the Job Cancel button fails in some ways that this PR does not attempt to fix.

mlampert:
I think this only fixes the issue in a very specific use case and introduces a regression for all other use cases. The short circuit only does the right thing if the user hasn't modified anything before hitting cancel. If they did make modifications and then press Cancel this will make Path inconsistent. Because the transaction is aborted the objects get reset to their state before the OP was edited - the `Path` of the object however does not get reset.

I don't see a way to make this work in the current (FC) update schema. The only option I see is to decouple Path generation from the FC update mechanism and let the user manually press a button to generate the actual Path. I would vote for that change in the next release - for this release the window for changes like this has closed.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [497808549](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/497808549) was triggered at [f916fc3](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/f916fc391dcd35c25278e66d8116a56d4af86d9e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

J-Dunn:
It is true that once the user has hit Apply , cancelling out is not as simple. 
Without introducing significant changes that mlampert suggests for later, there is still useful improvement that can be done ( I think ).  It requires a binary dirtypath flag to monitor whether changes have been made. Cancel as Russ has implemented here would then be safe if dirtypath flag is false. 

Cancel without any edit is desirable. A user may open the edit to examine parameters, eg to check which tool is used or verify cutting convention. ( Yes this is accessible in treeview but it is a lot less readable and accessible than the dlg. ) This should be possible without penalty. 

Cancel with parameter edits but without applying changes which alters the path. You start messing with something and prefer to keep it as it was. Cancel without penalty should be possible. That should work  : 

>  Because the transaction is aborted the objects get reset to their state before the OP was edited 

The problematic case is where user has hit apply and FCupdate has changed the path object. So it seems monitoring whether this has happened with a boolean would allow Russ's nice little mod to eliminate a lot of the unnecessary re-computations without needing more complex changes. 

For the path object , wouldn't it be sufficient to make a copy of that object on entry and restore it on cancel ? ( That maybe naively optimistic since I don't have much understanding of the depths of the update mechanism ).

Russ4262:
@mlampert Thanks for the awesome technical feedback. I am bit bummed about the shortcomings of my attempt, but alas, I shall not quit.  I did run some tests in the gui where I would edit an operation, click apply, then, click cancel. The property values appeared to be unchanged in the Data tab, and the paths appeared to be reverted back to originals in the viewport. Admittedly, I did not inspect the path data at anytime. I was working under the assumption that because the other property values were effectively cancelled, that the path data would be also.  From your explanation and review, it appears that there is another level of complexity here that I am missing.  Good to know!  Many thanks!

I might try to catch you on Gitter to ask for some help on this one  Or, if you like, feel free to apply your greater understanding and take over this endeavor. It might be more than I can handle with my lesser understanding of the Gui-to-nonGui interactions. Haha.

Looks like I have another round of digging and learning to do.

Again, so glad to have you involved! :)

Russ4262:
@mlampert @J-Dunn
Evening.  I have been re-reading your previous comments and thinking about the situation.  A couple items have surfaced.  In no particular order:

One ...
> Because the transaction is aborted the objects get reset to their state before the OP was edited - the Path of the object however does not get reset.

Please elaborate about the `Path` of the object not being reset.  More tests on my end, using the Inspect G-code tool before and after editing with apply and cancel, show the g-code for the OP to be reset correctly, or rather untouched.  This is also verified by sending the OP to the Python Console before and after the "edit, apply, cancel" action and running `obj.Path` to get a very simple summary of size and length.

Two ...
> If they did make modifications and then press Cancel this will make Path inconsistent.

What do you mean by `make Path inconsistent`?


Three ...
It has occurred to me that perhaps we are viewing this set of three buttons with different interpretations of intended behaviors.  My thinking is that the task panel is opened in order to edit the OP.  Once open, my thinking is that the three buttons have the following intended behaviors:
`Apply` - apply any changes made in the task panel in order to `preview` the resulting changes, without saving them to the source OP.
`Cancel` - cancel any and all changes made during the task panel edit session, leaving the OP in its pre-task-panel-edit state, and then close the task panel.
`OK` - accept all changes that might have been made, applying any unapplied changes, and then exit the task panel.

However, after reading your comments and those of J-Dunn, it seems there is an alternate interpretation of the buttons' intended actions.  Here is what I think they might be intended to do for some users.
`Apply` - apply any changes made in the task panel to the OP in order to preview the resulting changes, with the intention that these applied changes are saved to the original OP.
`Cancel` - (as I am not certain here) maybe cancel the last `Apply` and reset task panel values to pre-Apply values as well.  Maybe do not revert anything, and simply exit the task panel - in which case I suggest the label be changed from `Cancel` to `Exit`.
`OK` - accept all changes that might have been made, applying any unapplied changes, and then exit the task panel.

So, in comparing the two differing interpretations of the button actions, I think that this PR is on track for my interpretation.  With my interpretation, a user is still able to make changes, click `Apply` and have them saved so long as they click `OK` to commit the changes.

Four ...
I am not familiar with the FC Update feature or workflow or set of functions.  Can you please point me to some documentation that will help educate me about it?

@J-Dunn 
On your `Case 2` above, are you stating that perhaps a fourth button be added, say a `Reset` button to reset property values in the task panel that have been changed, but not applied?  For example, if a user has changed the Step Over and the Tool Controller, but not clicked `Apply`, then they could click a `Reset` button to reset the Step Over and Tool Controller values to previous values before being changed after the last `Apply` click.

Thanks for your feedback.
Russell

mlampert:
I agree with your interpretation of `Apply`, `Cancel` and `OK` - and it seems I still don't understand how the FC update works. Please ignore my ignorance and continue to rely on facts.

J-Dunn:
Russ:
> Apply - apply any changes made in the task panel in order to preview the resulting changes, without saving them to the source OP.

From my subjective understanding, I don't think that is the case. _Apply_ does actually apply the changes fully, this is reflected by the fact that the 3D view updates. ( This also relates to my complaints about recalculation being repeated on _Exit_ after _Apply_ has been done. This is totally redundant and another big time waster for 3D paths. Maybe these related issues will solve one another. )

This is apparently ( from your thorough testing evidence ) fully undone and the path **object** is restored up on cancel. Is some kind of update process needed to update the 3D view when the path object is replaced? Maybe that is the role of the FCupdate process.

> On your Case 2 above, are you stating that perhaps a fourth button be added, say a Reset button to reset property values in the task panel that have been changed, but not applied?

No, I had not thought of a reset button. That could be done but if exiting with Cancel was cost free, user can just Cancel out and reopen. I see no real need for another button. 

I see no concept of "preview" here, when user hits Apply it does exactly that.  This seems pretty standard behaviour across different platforms where such apply buttons are used. I doubt anyone expects different.

J-Dunn:
Hmm, this FCupdate question seems to have deeper consequences. 

I have a 3Dsurface on cylindrical model. I just went in to sketch editor to check the diameter. I looked at the sketch and hit close, without touching anything. It got bogged down recomputing the 3D path ! 

It seems there is a belt and braces policy of redoing everything, every time even if nothing is changed , just to be sure. 

While FC only had the relatively simple and fast 2.5D tools that was an expedient choice which makes sense in terms of rapid design. A nuanced approach now seems to be needed. 

I've also seen this re-computation problem when opening the job editor and closing with no changes. The whole job gets recomputed. 

It seems a better understanding of the cascade of update events that are trigger is needed. 
This is a deeper issue but will have wider reaching improvements once it's only doing the work which is required. 

I'm glad this is getting some attention, this is the fly in the ointment for 3D paths.

Russ4262:
@mlampert
> I think this only fixes the issue in a very specific use case and introduces a regression for all other use cases. The short circuit only does the right thing if the user hasn't modified anything before hitting cancel. If they did make modifications and then press Cancel this will make Path inconsistent. Because the transaction is aborted the objects get reset to their state before the OP was edited - the `Path` of the object however does not get reset.
> ...

I am trying to understand the situation related to this PR a little better.  Will you please provide me with a piece of code to run, or a pointer to an object to inspect, before and after an editing session with the task panel.  I am looking for such a test process to use both with this PR applied, and without, so that I can see the resulting inconsistency or regression created by this PR. 

I am still not understanding what object or other state of an operation or job is broken or otherwise left in an undesirable or incomplete condition after the fix in this PR is applied.  I am looking to learn more about the pieces to this complex puzzle and how this PR is affecting them and failing to address them.

Thanks for your time and help.

J-Dunn:
Earlier mlampert said the path object was not restored. I thought his last statement was an admission he maybe mistaken about that. I've just downloaded you fork, I'll try to find time to do some testing. 
Thanks for your work on this. 
Is your it your impression that this fully works as expected unless someone can prove a case where it does not ?

Russ4262:
@J-Dunn 
>... Is your it your impression that this fully works as expected unless someone can prove a case where it does not ?

No. My tests suggest that it is producing the intended and desired results.  However, I am not completely aware of the various layers and objects involved in the background that may be negatively impacted by the suggested changes.  So far my tests are not demonstrating any negative side effects, but I cannot declare this PR is 100% safe due to my ignorance of the aforementioned layers of code that I might be negatively impacting with the changes.

If there is good news that gives me confidence, it is that this PR does not attempt to delete any objects, or write to any.  It is relying on the `abortTransaction()` usage and the fact that all changes in the editing task panel are contained within a single transaction.  The Cancel button aborts the parent transaction, and the changes in this PR block the already-put-in-play execution of the operation's `Proxy.execute()` method.

I would like devs with a greater knowledge base of the involved objects (GUI and non-gui) to determine what breakage this PR `might` be causing.  I admit that even my tests might be too shallow to identify breakage deeper within the process and objects affected.  I am not seeing breakage (or noticeable effects of breakage), but I also know that just because I am not seeing breakage, does not mean it is not happening somewhere.  I repeat, I am a novice coder in this BIG project.

Thanks for the dialogue, help, and testing.

J-Dunn:
Hi Russ, 
here is a test file with 3Dsurface+boundary dressup on a cylinder. ( FCStd changed to zip to get around github filter ). 
[cylinder.zip](https://github.com/FreeCAD/FreeCAD/files/8348018/cylinder.zip)

When I edit the dressup ( without change ) and cancel, I get about 4s delay without the usual twirling cursor. 
If I edit the 3D path, I get about 7s with the twirling cursor plus 4s without. 
This seems a couple of seconds longer than recent master (counting in my head). There is a slight difference in the way the task panel shows during second part. Master shows the waiting cursor if I cancel the dressup. 

Unless I've messed up my build trees , this your fork I pulled yesterday:
[code]
OS: Fedora 34 (Thirty Four) (LXQt//usr/share/xsessions/lxqt)
Word size of FreeCAD: 64-bit
Version: 0.20.28344 (Git)
Build type: Release
Branch: master
Hash: 2eb3447694a11e95400dbdba4aa269b79af0490d
Python 3.9.10, Qt 5.15.2, Coin 4.0.0a, OCC 7.5.0
Locale: C/Default (C)
[/code]



Master:
[code]
OS: Fedora 34 (Thirty Four) (LXQt//usr/share/xsessions/lxqt)
Word size of FreeCAD: 64-bit
Version: 0.20.28285 (Git)
Build type: Release
Branch: master
Hash: 043e6c8ba34529ea5f40c1b1aaf555f4f1410259
Python 3.9.10, Qt 5.15.2, Coin 4.0.0a, OCC 7.5.0
Locale: C/Default (C)
Installed mods: 
  * FreeCAD_Assembly4 0.11.10
  * FreeCAD_assembly3
[/code]

Russ4262:
@J-Dunn 
Running some additional tests today using your situation with a dressup on an operation has yielded the fact that when the base op is cancelled, the dressup still recomputes.  Therefore, I need to add another attribute to the base op class as a flag for dressups.  Then, add a check in the dressup that looks at the base op's Proxy attributes to see if the cancel flag was thrown.

So, yes, your example above about editing the 3D Surface op that is the base, but just cancelling does indeed cancel the 3D Surface op, but does not cancel the parent Dressup on top.  I will see if I can get to this, perhaps this evening. 

Thanks for working on this with me, Sir.  Much appreciated.

Russ4262:
Modified and added code to allow dressups to check base ops for cancellation, and block recompute accordingly.  I also rebased on master since I pushed commits.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [501759913](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/501759913) was triggered at [aaa8370](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/aaa83701c6f035c5bcdefcc0448456ef9c8dc2fd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

J-Dunn:
I'm glad to be of assistance. 
Could you check my build version above it appears I last build after this upload. 

> Resolving deltas: 100% (316/316), completed with 90 local objects.
> From https://github.com/Russ4262/FreeCAD
>  + f916fc391d...aaa83701c6 fix/block_recompute_on_cancel -> origin/fix/block_recompute_on_cancel  (forced update)
> Already up to date.
> 

[code]
OS: Fedora 34 (Thirty Four) (LXQt//usr/share/xsessions/lxqt)
Word size of FreeCAD: 64-bit
Version: 0.20.28394 (Git)
Build type: Release
Branch: master
Hash: b219c50ed6be6e4c2c454d0b1594e6243ea9620a
Python 3.9.10, Qt 5.15.2, Coin 4.0.0a, OCC 7.5.0
Locale: C/Default (C)
[/code]

I'm still seeing about 8s delay when I open the 3dSurface path and cancel without change, as deated in my last post.

Russ4262:
@J-Dunn 
I am not very familiar with GitHub and comparing versions.  I clicked on the hash link in your post, then set the left comparison branch to master.  The resulting commits looked like it was the newest set of changes.

I am rebuilding, and will then run tests again, with and without dressup on operations.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [506484987](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/506484987) was triggered at [3703c9c](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/3703c9cb16210c7913449bc8764af3ea5b91e292). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Russ4262:
Update - 2022-03-31 22:26 CST
Ran some more tests of different type, which helped expose a different perspective and a problem with the first proposed version.  Consequently, I have completely changed the proposal.  The changes are much easier to apply!  Thanks for helping with testing and feedback.  This new method is dependent on the use of the `open`, `commit`, and `abortTransaction()` functions for this new strategy to work.  Thankfully, these transaction function calls appear to be in place already. 

I have only applied this new strategy to GUI modules that I feel reasonably confident are intended to cancel changes made when the `Cancel` button is clicked to exit an open Task Panel.  I have extended this new strategy to more modules.

I removed all the previous commits to replace them with the single new one.  I have rebased also.

Russ4262:
Just rebased also.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [506488708](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/506488708) was triggered at [8054849](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/8054849f2359c6077ef1d049e1bc58224d94d52a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

J-Dunn:
Great, this new strategy sounds like the right way to exploit the existing code. Many thanks for the continued effort improving this. It's one of the major workflow problems for your excellent 3D tool paths.  I'm building your repo and I'll do some testing later.

J-Dunn:
Strange, automatics checks seem OK, but I'm getting a build error on current state of this repo.

```
[ 41%] Building CXX object src/Base/CMakeFiles/FreeCADBase.dir/Parameter.cpp.o
/svn/FreeCADR/src/Base/Parameter.cpp:51:10: fatal error: Parameter.inl: No such file or directory
   51 | #include "Parameter.inl"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [src/Base/CMakeFiles/FreeCADBase.dir/build.make:651: src/Base/CMakeFiles/FreeCADBase.dir/Parameter.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....

```

Russ4262:
@J-Dunn The GitLab build succeeded.  It builds successfully for me.  Perhaps a corrupted file, or other issue is hindering your build.  I had to battle failing builds for a couple months myself before recently getting a successful build.  I had to decompress my Windows LibPack, start a fresh directory for build, and go from scratch.  Then it took more failures to realize that I had a CMake PYTHON_EXECUTABLE attribute missing, that I was sure was there automatically in the past.  Just glad to have successful builds again.

Russ4262:
@mlampert 
Is this new strategy used here more along what you were thinking earlier?  This new approach is much simpler and seems to be working as intended, whereas I found an issue with the first approach related to updating ops using changes to properties in the Data tab.  Thanks for your testing and imput.

J-Dunn:
Parameter.cpp 
```
#include "Parameter.h"
#include "Parameter.inl"
#include "Console.h"
#include "Exception.h"

```
`
find FreeCADR -iname "P*.inl"`
Nada, this file does not exist. 

I have no idea why this is compiling without error elsewhere.

J-Dunn:
Using my cylinder.FCStd model of hemispherical 3Dsurface with boundary dress-up, I do not see any difference between recalculation on cancel with no change and Apply with a change on boundary params. 
Not seeing any change here. Am I missing something?

mlampert:
@Russ4262 - I don't think you're gonna find a happy path while staying within the FC update framework. But you have proven me wrong more than once ;)
Splitting path generation from FC update requires a bigger discussion though.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [507882079](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/507882079) was triggered at [7632a68](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/7632a68d484d2a9d1bd01404b561aa8b6af0c910). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Russ4262:
Just pushed version 3 of this PR.  Back to a version similar to the first, but has ditched the problems identified in the first version with the property updates via the Data tab.  This new version is also easy to implement.  A part of the problem with version 2 was that I do not know the inner workings of FC update and just how the `touch` and `purgeTouched` play in updating objects, and nested objects.  Another problem with both of the first versions is that they were tied inside local transactions, so some of the flags were not being set correctly to original objects, rather temp objects in the aborted transactions.  This new version takes flags to global level, outside the object-transaction scope.  It does make use of the first method presented in this PR, simply cancelling the `execute()` method of the object.  So far tests are looking good.

I did remove the blocking of Job updates on cancel because the Job is not structured or organized like an operation of dressup.  Applying cancelling of Job updates will require some additional thought and analysis.

@J-Dunn Yes, that second version using the `purgeTouched()` appeared to not behave as consistently as expected.  Since I am not familiar with the `touch` system and FC Update, I am choosing to let it play out, but just cancel the `execute()` methods now by returning without executing any instructions therein.

@mlampert  This third version appears stable and reliable so far.  I am attaching the test file I have been playing with.  It contains a few ops, and a three-tier dressup-dressup-op situation to test multi-tier dressups.

[PartDesignExample.zip](https://github.com/FreeCAD/FreeCAD/files/8404847/PartDesignExample.zip)
I changed the extension from FCStd to zip for upload purposes.

Russ4262:
Just pushed fix for the Dogbone unit test, adding a needed `Name` attribute to the `TestFeature` object.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [507938782](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/507938782) was triggered at [1d46d07](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/1d46d074a5c6ccf1b36e876cfb1d8577d1e930fb). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

J-Dunn:
Russ, could you please look at this and make sure it has nothig to do with recent changes you have made ?
https://forum.freecadweb.org/viewtopic.php?f=15&t=67745

Thx

Russ4262:
@J-Dunn 

> Russ, could you please look at this and make sure it has nothig to do with recent changes you have made ? https://forum.freecadweb.org/viewtopic.php?f=15&t=67745
> 
> Thx

I have not been touching drilling, and I almost run the other direction from post-processing, as that is an area I have yet to even investigate as pertaining to the code and structure.

J-Dunn:
Thanks Russ. There is something going on there which is affecting g83 before it gets to the post-processor. It seems that the retraction depth is being ignored. This is new because I've used this in the last few months and it used to work.  I wanted to check that it was nothing to do with spin-edit changes.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [517427442](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/517427442) was triggered at [179f520](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/179f5206b2581098af01ad956b6ab571893a4b18). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Russ4262:
Just pushed refined version of latest fix strategy.  Moved new helper functions to PathUtils module.  Ran some successful tests with cancelling operations recomputes when new TC added to the job, but have disabled this feature since the Job code is not as linear and straight-forward as the operation code.

I have tested this code with a double dressup scenario where two dressups are applied on top of an operation.  I have tested it with a single dressup on top of an operation.  So far, test results look good.  I am not seeing any data loss or other negative affects so far.  Unit tests are passing.

Will review this PR's description above to verify that it is accurately describing the fix proposed.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [517778167](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/517778167) was triggered at [0b3dd14](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/0b3dd14e0f91e8c5c524878d2c25c8f90bbe654a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Russ4262:
Just rebased on master.  No changes.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632">pipeline status</a> for feature branch [PR_6632](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6632). Pipeline [533569405](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/533569405) was triggered at [e23918e](https://github.com/FreeCAD/FreeCAD/pull/6632/commits/e23918e17cf669b4ca1470e2f6105b8e76bb30f0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
We need to decide whether to merge, delay to 0.21, or add to the 0.20 milestone. Please comment

Russ4262:
I am.not convinced this PR will be an acceptable means to the targeted end.

PR #6634, opened 2022-03-21
Retire qassistant (try 2)

yorikvanhavre:
Follow-up of #6596 which I screwed up :)

@wwmayer I added your suggested changes, it compiles but crashes when pressing the F1 key to open the doc. I suppose I did some stupid mistake somewhere, any idea?


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6634">pipeline status</a> for feature branch [PR_6634](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6634). Pipeline [497519324](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/497519324) was triggered at [e0e4bf1](https://github.com/FreeCAD/FreeCAD/pull/6634/commits/e0e4bf1810af758faf9e133bfdcfea755b2b92dc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
I forgot to mention that before the call of PyImport_ImportModule you have to write `Base::PyGILStateLocker lock;`
```
void MainWindow::showDocumentation(const QString& help)
{
    Base::PyGILStateLocker lock;
    PyObject* module = PyImport_ImportModule("Help");
    if (module) {
        Py::Module helpmod(module, true);
        helpmod.callMemberFunction("show", Py::TupleN(Py::String(help.toStdString())));
    }
    else {
        PyErr_Clear();
        QUrl url(help);
        if (url.scheme().isEmpty()) {
            QMessageBox::critical(getMainWindow(), tr("Help addon needed!"),
                                  tr("The Help system of %1 is now handled by the \"Help\" addon. "
                                     "Install it with menu Tools > Addons Manager").arg(qApp->applicationName()));
        }
        else {
            QDesktopServices::openUrl(url);
        }
    }
}

```

yorikvanhavre:
Thanks @wwmayer this works perfectly! Just out of curiosity:

1. I understand we need to lock the interpreter because it cannot execute more than one python script at a time. But we don't unlock it afterwards. Is it automatically done when the function exits?
2. In your code above you have changed the way the `Help.show()` method is used, so it doesn't reflect in the Py console anymore... Do you find it better? I was hesitating about this, and ended up thinking it was better to explicitly print it in the py console so people learn how to use it, but it's no strong opinion

wwmayer:
> I understand we need to lock the interpreter because it cannot execute more than one python script at a time. But we don't unlock it afterwards. Is it automatically done when the function exits?

The class PyGILStateLocker uses the RAII idiom. It locks the GIL in its constructor and releases it in its destructor. This is then guaranteed exception-safe.

> Do you find it better? ... but it's no strong opinion

I have no strong opinion either.

yorikvanhavre:
Thanks for the comments @wwmayer !
I ended up leaving the explicit print to the console for now...

wwmayer:
> I ended up leaving the explicit print to the console for now...

But then `module` must be Py_DECREF'ed to avoid a reference leak.

yorikvanhavre:
> But then `module` must be Py_DECREF'ed to avoid a reference leak.
Done. Thanks!!

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6634">pipeline status</a> for feature branch [PR_6634](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6634). Pipeline [499644152](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/499644152) was triggered at [ac663cd](https://github.com/FreeCAD/FreeCAD/pull/6634/commits/ac663cd0c4fe4bbf5770f6f55ebe89d793292d6e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
Woah! Huge paradigm shift here. Thanks yorik!

PR #6640, opened 2022-03-21
Gui: fix some weird behavior of number format option

0penBrain:
Fixes problems seen on #6364 implementation

I should say that I didn't perfectly understand what was failing, but it seems that when using `QLocale::setDefault` after application has been started, the `LocaleChange` event isn't well propagated or handled.
Sending it manually afterwards solves the issue.

### DIFF
--- a/src/Gui/Language/Translator.cpp
+++ b/src/Gui/Language/Translator.cpp
@@ -258,6 +258,10 @@ bool Translator::setLocale(const std::string& language) const
     if (!bcp47.empty())
         loc  = QLocale(QString::fromStdString(bcp47));
     QLocale::setDefault(loc);
+    // Need to manually send the event so locale change is fully took into account on widgets
+    auto ev = QEvent(QEvent::LocaleChange);
+    qApp->sendEvent(qApp, &ev);
+
 #ifdef FC_DEBUG
     Base::Console().Log("Locale changed to %s => %s\n", qPrintable(loc.bcp47Name()), qPrintable(loc.name()));
 #endif


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6640">pipeline status</a> for feature branch [PR_6640](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6640). Pipeline [497519943](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/497519943) was triggered at [1197575](https://github.com/FreeCAD/FreeCAD/pull/6640/commits/11975750bb2fe0e9cdc6511d7e2d60f6bf6d988b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks! I wanted to quickly test the PR but it is not based on recent master and thus it takes ages to compile :-(

donovaly:
I tested the PR now by applying its diff upon today's master. Unfortunately, I don't see any fix. In these fields it still doesn't work:
*    App::PropertyFloat
*    Gui::InputField

Since you use an old master, can you please rebase your master and try again?

0penBrain:
@donovaly Hmm the branch was rebased at the time of the PR. It's now only 16 commits behind master. Are you sure you compiled the right thing? I'll rebase though

donovaly:
>Hmm the branch was rebased at the time of the PR.

It contained e.g. the iistaskpanel stuff that was removed a week ago.

However, as I wrote, the PR does not fix the issue for me. I compiled it now again to be sure.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6640">pipeline status</a> for feature branch [PR_6640](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6640). Pipeline [499644155](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/499644155) was triggered at [b830fe5](https://github.com/FreeCAD/FreeCAD/pull/6640/commits/b830fe56f3fec3933d56db8de18d2ff3d1e3fe7f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
@donovaly I switched my whole system to German. Can confirm that bug you saw is solved by this PR in my case. Hardly sure what you really compiled according the weirdness of your build env. :confused:

0penBrain:
@donovaly this has nothing to do with bug you recently mentioned (and actually nothing to do with these PR).
However this is the reason for this behavior that you already reported where eg. 10.00m is inadvertently converted to 1000m. I already explained several times that it's due to thousand separator removal (which is also done in QuantitySpinBox for example). I've for long been in favor of removing this, but it needs to be done with much care. Especially doing this at only one place as you did is IMO risky, and should be carefully discussed and tested. 

Regarding PropertyFloat, it's actually more special 
First saying 'PropertyFloat' is a bit weird. These values are exposed to user with a stock QDoubleSpinBox. So it's unlikely to be directly modified. ;)

wwmayer:
> I found the fix for Gui::InputField: 

That's not a bug fix but only cures the symptom because the actual problem is that for some widgets the group separator is a comma and decimal point a dot while at the same time for some other widgets it's the opposite. The behaviour mainly depends on the system settings.

So, if on my system LANG=de_DE.UTF-8 and I change language in FreeCAD to English then comma is supposed to be the group separator but for some InputField widgets the group separator is incorrectly set to dot.
And if I set LANG=C before starting FreeCAD and change language to German then I can observe the very same behaviour with commas that are incorrectly used as group separator.

This PR fixes exactly that!

wwmayer:
> I found the fix for Gui::InputField:

It does not only not fix the issue but adds another regression. To test:
* set language to German with the option to adjust number format
* create an InputField widget and set its content to `1.000` that is supposed to be one thousand
* after leaving the focus the value will turn into `1,000`.

This is because the quantity parser receives `1.000` as input and converts it into `1`

adrianinsaval:
@wwmayer is it possible to make all of those accept comma or dot interchangeably as a decimal separator as the expression (and therefore spreadsheet too) field does?

donovaly:
> That's not a bug fix but only cures the symptom 

As I wrote the PR does not fix the bug for me. Now, that you reverted my commit, the bug is back and breaks for me all InputFields:

I recompiled now and after your reversion of my commit, the bug is back:

FreeCAD_qJ75F9oh4K

Before I made the commit I tested using these 4 states:
- English locale
* with use selected number format
* without use selected number format
- German locale
* with use selected number format
* without use selected number format

And with my commit, values I enter e.g. in the FEM Gmsh dialog were correctly handled.

So I am now a bit lot. We cannot leave the situation as it is. Now still all Gui::InputField and App::PropertyFloat fields don't work for me on a German Win 10 with English menus and the "use number format option" on.

> This PR fixes exactly that!

OK. For me it did not make any difference.

0penBrain:
@donovaly I'll have another careful check at InputField beginning of next week. ;) Hopefully it's not OS specific or due to some Python caching. We'll keep updated

donovaly:
There is further info in this forum thread starting from this post: https://forum.freecadweb.org/viewtopic.php?p=583208#p583208

wwmayer:
What seems to fix the issue on Windows is to add this line to the constructor of InputField:
`setLocale(QLocale());`

When debugging the Qt code then it internally sets the Qt::WA_SetLocale and this apparently sets up the widget with the correct locale.

donovaly:
> What seems to fix the issue on Windows is to add this line to the constructor of InputField: `setLocale(QLocale());`
> 
> When debugging the Qt code then it internally sets the Qt::WA_SetLocale and this apparently sets up the widget with the correct locale.

Many thanks! It works for me.
Now the last missing part is that in App::Property float and English menus with the use number format option on, the dot is removed, so an input of "1.1" becomes "11":

FreeCAD_UuSsWWvNbC

0penBrain:
> Many thanks! It works for me. Now the last missing part is that in App::Property float and English menus with the use number format option on, the dot is removed, so an input of "1.1" becomes "11":

IMO this is a "drawback" of @wwmayer's fix. I think (just a thought right now, will dive in the code a bit later) that doing so, it "breaks" the propagation (that seems to not work well) so I guess we have to handle the `LocaleChange` event to correctly reset the locale. Will test and report. ;)

0penBrain:
@donovaly Just to be sure it's not highly specific, can you reproduce the exact same behavior with for example, the x/y/z values of Placement/Axis (which also are unitless floats) ? Thx

donovaly:
> can you reproduce the exact same behavior with for example, the x/y/z values of Placement/Axis

Yes:
FreeCAD_oC4gxfnKFb

[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.28459 (Git)
Build type: Release
Branch: master
Hash: c447f5a31d07adda818524b08207130e4eb555c9
Python 3.8.6+, Qt 5.15.2, Coin 4.0.1, OCC 7.5.3
[/code]

0penBrain:
@donovaly @wwmayer I think I finally found something. At least what looks to me as a far better solution.
Could you please test PR #6669 and tell me if it works for you ?
It is does, I would find very more comfortable to leave with this solution that doesn't look like a dirty patch. :)
The only drawback I found ATM is for `Gui::InputField` where you need to enter+leave them after a locale change so that new format is applied (entering a value correctly took into account new locale). This looks very minor to me (compared to issues we met till now) and will be probably easy to fix in a proper way.
I found another glitch where format isn't reset to C/POSIX if "use selected language number format" is disabled while it previously was enabled. Will fix this in another PR.
Thx for feedback (and precious testing and insights).

PR #6684, opened 2022-03-30
[GUI] add missing apply to color gradients dialog

donovaly:
an attempt according to: https://forum.freecadweb.org/viewtopic.php?p=583897#p583897

@WandererFan , what do I do wrong? The compilation fails on this line

`QObject::connect(dlg, SIGNAL(dlg.ModelChange(int index)), this, SLOT(onModelChanged(int index)));`

### DIFF
--- a/src/Gui/DlgSettingsColorGradientImp.cpp
+++ b/src/Gui/DlgSettingsColorGradientImp.cpp
@@ -23,6 +23,7 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <cmath>
+# include <QComboBox>
 # include <QDoubleValidator>
 # include <QFontMetrics>
 # include <QLocale>
@@ -54,7 +55,10 @@ DlgSettingsColorGradientImp::DlgSettingsColorGradientImp( QWidget* parent, Qt::W
     ui->floatLineEditMin->setValidator(fMinVal);
 
     QFontMetrics fm(ui->floatLineEditMax->font());
-    ui->floatLineEditMax->setMinimumWidth(QtTools::horizontalAdvance(fm, QString::fromLatin1("-1000.000000")));
+    ui->floatLineEditMax->setMinimumWidth(QtTools::horizontalAdvance(fm, QString::fromLatin1("-1000.0")));
+
+    connect(ui->comboBoxModel, qOverload<int>(&QComboBox::currentIndexChanged),
+        this, &DlgSettingsColorGradientImp::onModelChanged);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


chennes:
First, I'd change to using the new-style connect, like this:
```
QObject::connect(dlg, &DialogClassnameHere::ModelChange, this, &ThisClassnameHere::onModelChanged);
```
and see if that's enough to fix it. I don't know the details of the ModelChange function, if you might need some additional cleverness to deal with argument types, etc.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6684">pipeline status</a> for feature branch [PR_6684](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6684). Pipeline [504495098](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/504495098) was triggered at [bc6f5aa](https://github.com/FreeCAD/FreeCAD/pull/6684/commits/bc6f5aad2a4d293282bb91a3f25c5f00a9298024). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/DlgSettingsColorGradientImp.cpp
```diff
@@ -54,7 +55,10 @@ DlgSettingsColorGradientImp::DlgSettingsColorGradientImp( QWidget* parent, Qt::W
     ui->floatLineEditMin->setValidator(fMinVal);
 
     QFontMetrics fm(ui->floatLineEditMax->font());
-    ui->floatLineEditMax->setMinimumWidth(QtTools::horizontalAdvance(fm, QString::fromLatin1("-1000.000000")));
+    ui->floatLineEditMax->setMinimumWidth(QtTools::horizontalAdvance(fm, QString::fromLatin1("-1000.0")));
```
wwmayer, line 58: It is on purpose to use a number with six decimals after the dot to guarantee that the dialog is wide enough. With one of your recent changes you have changed the width in the UI file and every time I opened the dialog I had to manually change its size in order to avoid that numbers are truncated.

Reverting this fix again will be an annoying usability regression.

donovaly: > It is on purpose to use a number with six decimals after the dot to guarantee that the dialog is wide enough.

But this is a hack. A minimum for the width in the UI is the right approach, Alternatively one can set in the UI to expand if the numbers become bigger. I will have at the UI file and make this change.

> Reverting this fix again will be an annoying usability regression.

I obviously missed something. What do you refer to? Did I accidentally remove the minimal width settings from this dialog? Or what do you mean with the fix?

donovaly: > I will have at the UI file and make this change.

This is the result: #6695 I found it annoying that one had to deal with huge numbers because the LineEdit did not handle scientific notation. But most simulation results need this. Can you please give it a try if this change if fine for you as well?

donovaly:
Obsolete now, thanks to https://github.com/FreeCAD/FreeCAD/commit/027fb0774364669

PR #6686, opened 2022-03-30
Gui: if expression is emptied, pressing Enter discards expression

0penBrain:
https://forum.freecadweb.org/viewtopic.php?t=67471

### DIFF
--- a/src/Gui/DlgExpressionInput.cpp
+++ b/src/Gui/DlgExpressionInput.cpp
@@ -118,6 +118,14 @@ QPoint DlgExpressionInput::expressionPosition() const
 
 void DlgExpressionInput::textChanged(const QString &text)
 {
+    if (text.isEmpty()) {
+        ui->okBtn->setDisabled(true);
+        ui->discardBtn->setDefault(true);
+        return;
+    }
+
+    ui->okBtn->setDefault(true);
+
     try {
         //resize the input field according to text size
         QFontMetrics fm(ui->expression->font());


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6686">pipeline status</a> for feature branch [PR_6686](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6686). Pipeline [504695972](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/504695972) was triggered at [fbd2fd6](https://github.com/FreeCAD/FreeCAD/pull/6686/commits/fbd2fd6b4c02bf44477463a13926926bc00fe572). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
If @wwmayer agrees, this could in my opinion go in for 0.20 since it is kind of a bugfix.

wwmayer:
The PR causes a regression: it allows to set a numerical constraint with the wrong unit

* Create a sketch and add a line
* set the length constraint and open the expression editor
* replace the unit (e.g. mm) with s
* the dialog allows to accept this wrong quantity

Suggestions:
* move the call of `ui->okBtn->setDisabled(true);` into the block `if (text.isEmpty())` (to make it clear that this is only needed for an empty string)
* re-add the call of `ui->okBtn->setDisabled(true);` into the catch block

0penBrain:
@wwmayer Thx for review. Will have a look soon. However I think it's safe to ignore this PR for 0.20 if there is better things to do. ;)

0penBrain:
@wwmayer Done. Sorry for this, I should have caught it.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6686">pipeline status</a> for feature branch [PR_6686](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6686). Pipeline [507917960](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/507917960) was triggered at [8dfcdc5](https://github.com/FreeCAD/FreeCAD/pull/6686/commits/8dfcdc56e0fa234c58bcb969d300b8dfb6d927c6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
From my point of view this could be merged for v0.20. It's not a real bug fix but an improvement in usability.

Btw, the Clear button uses mnemonics but it depends on the translation if and which letter is used and I don't know if this works on all platforms.

0penBrain:
> Btw, the Clear button uses mnemonics but it depends on the translation if and which letter is used and I don't know if this works on all platforms.

Not sure what you're exactly talking, but specifically regarding this PR it works independently of the language.

wwmayer:
> Not sure what you're exactly talking about, but specifically regarding this PR it works independently of the language.

Yes, your PR makes it independent of all that but that was not my point.

With the current version of master and English as language the C of the text Clear is underlined. So, it's possible to invoke it with ALT+C. If language is set to German the button is called Löschen where the L is underlined (ALT+L).

So, all what I am saying is that at least on some systems it's already possible now to use a shortcut but that it depends on the language and if the translator was careful enough to not ignore the ampersand of the English text. Not more and not less.

0penBrain:
> So, all what I am saying is that at least on some systems it's already possible now to use a shortcut but that it depends on the language and if the translator was careful enough to not ignore the ampersand of the English text. Not more and not less.

Crystal clear. Also some times ago I fixed an issue that when 'Clear' button was selected using Tab then Enter was pressed, the dialog was actually accepted. So from there it's also possible to clear expression with just Tab then Enter. ;)

donovaly:
What is the status/outcome? The PR is in the 0.20 milestone and should therefore me merged soon. To me it reads as if the PR is almost but not 100% ready, right?

0penBrain:
@donovaly it's ready to be merged. ;)

PR #6695, opened 2022-03-31
[GUI] fix color gradient dialog

donovaly:
- large numbers as they appear of most simulations were not correctly displayed
- remove non-functional help button

Stress values like 210e9 Pa are standard simulation results. However, in FC 0.19 the user gets this back:
FreeCAD_hjF2VrHufq

So the 210 MPa are expanded with precision mistakes and the dialog is not wide enough for this.

This PR fixes it by using scientific notation:
FreeCAD_SsPUl95ZNh

### DIFF
--- a/src/Gui/DlgSettingsColorGradient.ui
+++ b/src/Gui/DlgSettingsColorGradient.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>260</width>
+    <width>255</width>
     <height>313</height>
    </rect>
   </property>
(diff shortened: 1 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6695">pipeline status</a> for feature branch [PR_6695](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6695). Pipeline [505456931](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/505456931) was triggered at [01596b6](https://github.com/FreeCAD/FreeCAD/pull/6695/commits/01596b69d98f34bba126b3e2351e4fa3e448661b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
I tested today with current master on another Win 10 PC and there, the scientific notation was correctly handled. However, here at my development PC this is not the case and I need to set the validator with scientific notation. Doing so does not harm.

Issue #6692, opened 2022-03-30
[Feature Request] Enable KeyboardInterrupt (ie Ctrl-C) to work in the Python console to break out of infinite loops


goatchurchprime:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=22&t=67587&p=584414

### Subproject(s) affected?

Core

### Idea description

The Python console of FreeCAD treats Ctrl-C as a copy selection instead of a KeyboardInterrupt.  

This makes it impossible to break out of an infinite loop or long running process as you might have to do when you are hacking Python macro code. 

(There is some process in the system which detects that the main GUI thread has frozen and pops up a dialog box with the option to kill the entire FreeCAD process or continue to wait.  It would be nicer if this dialog box also had the option to send a signal.CTRL_C_EVENT into the Python code.)

It seems that the Python macro code is running directly in the single GUI thread, which is why everything completely freezes.  A work-around [mentioned in the forum](https://forum.freecadweb.org/viewtopic.php?t=12678#p101749) is to call back out to some function in the FreeCADGui class to poll a Cancel button in a popup dialog.  But this is no use if your code is stuck in a loop that doesn't include this function.

 

### Anything else?

![freecadcrash](https://user-images.githubusercontent.com/677254/160895374-58b3e3f3-aaf9-47d8-9653-695b1e5482dc.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MicobyteMichael:
I believe I've addressed this issue in PR #6720 by moving execution of Python code to a background thread and adding a Ctrl+C listener to the GUI. Please see the PR for details.

goatchurchprime:
Thinking a bit more about this, it might be a good idea to implement this with a pop-up window (instead of the Ctrl-C) similar to the "FreeCAD is not responding" one, but to include the option to interrupt/cancel the running Python macro instead of Force quit the whole application.

Such a popup would automatically appear when a macro runs for more than five seconds and it would show a timer for how long it had been running.

(As a side-note, in one CADCAM system I once worked on we had to remove the progress timer because too many users used it to measure the calculation speed in the next version and complained if it ran one second longer: "Why am I paying this much money for an upgrade that's slower?".  Unlike proprietary software customers, open source users can handle the truth because they know how to send back useful information to the developers rather than just refuse to pay the invoice!)

[...]

PR #6720, opened 2022-04-07
Gui: Fix app not responding when executing long-running Python code

MicobyteMichael:
**What does this PR do?**
This pull request addresses issue #6692. In summary, the issue reports that the entire FreeCAD program hangs when a long-running statement (such as an infinite loop) in entered via the interactive Python console. This is caused by the long-running function monopolizing the main thread and preventing Qt from updating / processing events. I have resolved this by moving the execution of Python code from the console into a background thread. I have also added a way to terminate Python code running in the console by adding a key listener to the console which listens for the Ctrl+C key combination and triggers a `KeyboardInterrupt` in the background thread, which kills whatever Python code is currently running. Please see my commit messages for further details on implementation.

**Filling out the standard Pull Request form**

- [ ]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum ***Note: Is this small enough to qualify? If not, please let me know, I am glad to discuss it further.***
- [ ]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the issue ID number from our [Issues database](https://github.com/FreeCAD/FreeCAD/issues) in case a particular commit solves or is related to an existing issue. Ex: `Draft: fix typos - fixes #4805` ***Note: I did not realize this was a requirement until I saw this bullet point, I hope that me including the issue ID in the PR description is sufficient***

---


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [510935656](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/510935656) was triggered at [8055688](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/80556881a2030c8809421e0290f1ee7010cf932e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Thanks for the PR -- I just wanted to let you know that we are in a feature freeze pending the release of 0.20 right now, so there will be a delay before we get a chance to review this PR. We appreciate your patience!

MicobyteMichael:
No worries! Thanks for bearing with me, I wasn't aware of the freeze.

0penBrain:
@MicobyteMichael your commit history is really weird. Could you squash them altogether ?

Is this fix also working for macros manually ran by user ?

wwmayer:
I was curios about this PR and tried to build it. Using threads.h is not portable since on Windows compilers lack of that file. An option there is to use pthreads.h. Even on Linux it requires an extra package to get this header (I guess it's musl-dev).

But there are methods to get the thread id with C++. On https://stackoverflow.com/questions/7432100/how-to-get-integer-thread-id-in-c11 a method is provided that does:
```
    std::stringstream ss;
    ss << std::this_thread::get_id();
    threadID = std::stol(ss.str());
```
and this seems to work.

At least on Linux QThread::currentThreadId() will work too. However, the return value is a void* that first must be converted into a long. Not sure if that trick will work on Windows too but there is still the Win API function GetCurrentThreadID().

After having made the changes of the 1st method to get it compiled I started FreeCAD and entered this into the console:
`App.newDocument()`
and it exactly happened what I have feared. FreeCAD crashed because Qt raised a fatal error and aborted the process.
The reason is that Qt's GUI is not thread-safe and thus it's not allowed to modify it from within another thread other than the GUI thread.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [511781931](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/511781931) was triggered at [c0d5837](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/c0d58375d972a698fb4922061392a29ccaddc6b3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

MicobyteMichael:
Hi folks - thanks for all these suggestions and reviews. I'll begin addressing them shortly. I've also squashed all my commits as requested and fast-forwarded to the new latest commit on master. Sorry for the weirdness, SourceTree did something strange with my previous rebase. Also, as for the macro question, I haven't tested that quite yet. Suffice it to say this a bit more complex of an issue than I initially anticipated.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [511786256](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/511786256) was triggered at [803884f](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/803884f72fcf0b9576bce66c14b712899fb10e02). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

goatchurchprime:
Can this also block any edits/changes to the Model Tree in addition to blocking the execution of a second concurrent Python script? This would prevent me from deleting an object that the executing Python code has a pointer to.  We need to make sure that the user can only do the right thing: first stop a running Python script, and then do some editing through the GUI -- not the other way round.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [519223105](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/519223105) was triggered at [dbe85b5](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/dbe85b567834ca77c6df429cbce7af72b8eb94ef). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [519224666](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/519224666) was triggered at [5c9050a](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/5c9050ae3b18af6fb0fe6191494395938dd97e4b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

MicobyteMichael:
I was really hoping to include more in this update, but unfortunately I'm a bit stuck. I've made the change cross-platform using @wwmayer's suggestion of using <thread> instead of <threads.h>. But I'm still trying to figure out how to make the Python code still be able to interact with the GUI properly despite executing in a separate thread. The only thing I can think of is somehow wrapping all the GUI-related calls in something that runs the GUI code on the main thread then passes the results back to the worker. But that would be a pretty major change. I'm open to any suggestions.

wwmayer:
> The only thing I can think of is somehow wrapping all the GUI-related calls in something that runs the GUI code on the main thread then passes the results back to the worker.

This isn't possible because one can e.g. use PySide2 code and there you cannot change the C++ implementation.

> I'm open to any suggestions.

Instead of using a separate thread would using the mechanism of a Python debugger be an option?
See https://docs.python.org/3/c-api/init.html#profiling-and-tracing

MicobyteMichael:
> Instead of using a separate thread would using the mechanism of a Python debugger be an option?
That seems to work! I've re-implemented my changes using this mechanism instead of a background thread, and it seems to work! See my latest commit for details.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [520057609](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/520057609) was triggered at [45f8691](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/45f86917e6cef6d8c08dec26ede91ac84ddd429a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

MicobyteMichael:
Ok, looks like there's a very odd issue with this change. For some reason, when I close out the last open document from the close button in the GUI (the one in the documents bar). It causes a segfault most of the time in the destructor for `View3DInventor` but sometimes in some random smartpointer class from Boost? It's very odd. I was convinced it was some kind of threading thing but I think I've ruled that out by adding a printout if the thread the python callback is operating in isn't the main GUI thread.

The other strange thing is that it works fine when you close the last active document from the File->Close menu at the top, and it also works file when you use `App.closeDocument(...)` from the Python console. I've attached two backtraces that I've managed to capture in case anyone else can lend a second set of eyes and help me try to figure out what the cause of this is. My next step is probably to investigate how File->Close differs from the close button in the documents bar at the bottom in hopes of figuring out what about closing the last document from that bottom bar is causing this crash.

[boost_backtrace.txt](https://github.com/FreeCAD/FreeCAD/files/8518243/boost_backtrace.txt)
[widget_backtrace.txt](https://github.com/FreeCAD/FreeCAD/files/8518244/widget_backtrace.txt)
.

wwmayer:
> `PyEval_SetTrace(&pythonTraceFunc, NULL);`

Why is the trace function set at program start and kept for the whole runtime? I think that this can cause a regression on speed. Actually it should be enough to set it only tmp. when the user enters something into the Python console.

```

int pythonTraceFunc(PyObject *obj, PyFrameObject *frame, int what, PyObject *arg) {
    static std::chrono::time_point<std::chrono::steady_clock> lastCalledTime = std::chrono::steady_clock::now();
	std::chrono::time_point<std::chrono::steady_clock> currTime = std::chrono::steady_clock::now();
	long ms = (currTime - lastCalledTime) / 1ms;

	if(ms >= 200) {
		lastCalledTime = currTime;
		QGuiApplication::processEvents();
	}

    return 0;
}
```
Why does it force to process pending events every 200 ms? Doing this can be very dangerous and I guess that's the reason for the crashes. The problem is that when forcing the processing of events the execution order could change and thus leads to unexpected behaviour.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [520819056](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/520819056) was triggered at [112ac1b](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/112ac1b07f50e967874f396a04d128f21e7dff8d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

MicobyteMichael:
> Actually it should be enough to set it only tmp. when the user enters something into the Python console.

I've implemented this change.

> Why does it force to process pending events every 200 ms?

This was just a heuristic for updating the GUI "often enough", to keep the app responsive while the code is running. Now that I've changed it to only run while there is code running in the console (i.e. when the main thread is monopolized and not processing events) it seems like it's not running the event processing "too often" now, as the crashes I was experiencing before are no longer occurring. That being said, if there is a better way to keep the app responding, I am always open to suggestions.

wwmayer:
> That being said, if there is a better way to keep the app responding, I am always open to suggestions.

I don't know of any other way to keep the application responsive. IMO, updating every 200 ms is still too often because on complex models in the 3d view this can considerably slow down performance. So, instead of using a hard-coded value it's better to use a user-defined value which when set to -1 could even be used to disable the update.

MicobyteMichael:
The user can now select the interval at which the profiler runs in the preferences dialog (including setting it to zero to disable it).

Capture

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [522824361](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/522824361) was triggered at [919f5f4](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/919f5f496d48d83804933217d2780f7a8936ba68). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720">pipeline status</a> for feature branch [PR_6720](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6720). Pipeline [522824954](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/522824954) was triggered at [fecdc46](https://github.com/FreeCAD/FreeCAD/pull/6720/commits/fecdc46459b2a3f3535729a5a72d2fbe3e9718a1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
This PR seemed to be making very good progress and then went silent.  Bumping it for an update.

wwmayer:
Replaced with https://github.com/FreeCAD/FreeCAD/pull/10587

PR #6740, opened 2022-04-13
Part: Cross-section, fix issue with using Links that have been scaled…

mwganson:
… and/or transformed

https://forum.freecadweb.org/viewtopic.php?f=3&t=67961

### DIFF
--- a/src/Mod/Part/Gui/CrossSections.cpp
+++ b/src/Mod/Part/Gui/CrossSections.cpp
@@ -200,7 +200,7 @@ void CrossSections::apply()
     std::vector<App::DocumentObject*> obj;
     for (std::vector<App::DocumentObject*>::iterator it = docobjs.begin(); it != docobjs.end(); ++it){
         if (!Part::Feature::getTopoShape(*it).isNull()) {
-            obj.push_back((*it)->getLinkedObject());
+            obj.push_back((*it));
         }
     }
 


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6740">pipeline status</a> for feature branch [PR_6740](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6740). Pipeline [515515261](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/515515261) was triggered at [32fff59](https://github.com/FreeCAD/FreeCAD/pull/6740/commits/32fff59e4f11fc09e200124d514721a5c1290936). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@realthunder , could you please also review this PR? If OK, it can go in for FC 0.20

realthunder:
It's good.

mwganson:
> It's good.

Can you take a look at this thread?

https://forum.freecadweb.org/viewtopic.php?f=3&t=67960

What is the best way to get the subshape when the document object is app link sublink?

realthunder:
You can use `Part::Feature::getTopoShape(obj, subname, /*needSubElement*/ true)` to obtain the sub shape from a sub link. It works on all types of objects, and can be used as a general replacement for `TopoShape::getSubShape()`. Note that the `GUIApplication::notify` error reported in the thread indicates a leaked exception that needs to be addressed, as `TopoShape::getSubShape()` may throw exception. There maybe other places in `Part` that calls this function without catching exception or passing the `silent` option. `Part::Feature::getTopoShape()` does not throw exception, but return null shape if failed.

PR #6761, opened 2022-04-17
Part: Link support -- replace TopoShape.getSubShape() with Part::Feat…

mwganson:
…ure::getTopoShape(obj, elementname, true).getShape() in order to support use of App::Link sublinks.

This change, as suggested by @realthunder, seems to correct the issue of sublinks not working.  But there are a couple of issues I discovered during testing.

1) Sweep will crash with a certain file.  I will upload the file to the forum and create a topic, and then link to it after I submit this PR.

2) Ruled surface gets moved to the wrong position.  This is due to a workaround, which might no longer be needed when using the getTopoShape().  To address this I have added a new boolean property to ruled surfaces called "ResetPlacement", defaulting to false.  If set true, you get the workaround, if false the workaround is not done.

### DIFF
--- a/src/Mod/Part/App/FeatureExtrusion.cpp
+++ b/src/Mod/Part/App/FeatureExtrusion.cpp
@@ -100,7 +100,7 @@ bool Extrusion::fetchAxisLink(const App::PropertyLinkSub& axisLink, Base::Vector
 
     TopoDS_Shape axEdge;
     if (axisLink.getSubValues().size() > 0 && axisLink.getSubValues()[0].length() > 0) {
-        axEdge = Feature::getTopoShape(linked).getSubShape(axisLink.getSubValues()[0].c_str());
+        axEdge = Feature::getTopoShape(linked, axisLink.getSubValues()[0].c_str(), true /*need element*/).getShape();
     }
     else {
         axEdge = Feature::getShape(linked);
--- a/src/Mod/Part/App/FeatureRevolution.cpp
+++ b/src/Mod/Part/App/FeatureRevolution.cpp
@@ -94,7 +94,7 @@ bool Revolution::fetchAxisLink(const App::PropertyLinkSub &axisLink,
 
     TopoDS_Shape axEdge;
     if (axisLink.getSubValues().size() > 0  &&  axisLink.getSubValues()[0].length() > 0){
-        axEdge = Feature::getTopoShape(linked).getSubShape(axisLink.getSubValues()[0].c_str());
+        axEdge = Feature::getTopoShape(linked, axisLink.getSubValues()[0].c_str(), true /*need element*/).getShape();
     } else {
         axEdge = Feature::getShape(linked);
     }
--- a/src/Mod/Part/App/PartFeatures.cpp
+++ b/src/Mod/Part/App/PartFeatures.cpp
@@ -61,6 +61,8 @@ RuledSurface::RuledSurface()
     ADD_PROPERTY_TYPE(Curve2,(nullptr),"Ruled Surface",App::Prop_None,"Curve of ruled surface");
     ADD_PROPERTY_TYPE(Orientation,((long)0),"Ruled Surface",App::Prop_None,"Orientation of ruled surface");
     Orientation.setEnums(OrientationEnums);
+    ADD_PROPERTY_TYPE(ResetPlacement,(false),"Ruled Surface",App::Prop_None,
+                      "Whether to reset placement");
 }
 
 short RuledSurface::mustExecute() const
@@ -70,6 +72,8 @@ short RuledSurface::mustExecute() const
     if (Curve2.isTouched())
         return 1;
     if (Orientation.isTouched())
+        return 1;
+    if (ResetPlacement.isTouched())
         return 1;
     return 0;
 }
@@ -100,6 +104,7 @@ App::DocumentObjectExecReturn* RuledSurface::getShape(const App::PropertyLinkSub
 
     if (!part.getShape().IsNull()) {
         if (!element[0].empty()) {
+            //shape = Part::Feature::getTopoShape(obj, element[0].c_str(), true /*need element*/).getShape();
             shape = part.getSubShape(element[0].c_str());
         }
         else {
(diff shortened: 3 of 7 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6761">pipeline status</a> for feature branch [PR_6761](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6761). Pipeline [518487412](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/518487412) was triggered at [a9c135d](https://github.com/FreeCAD/FreeCAD/pull/6761/commits/a9c135d33783db9367af5405ee17c812c16a420b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

mwganson:
https://forum.freecadweb.org/viewtopic.php?f=17&t=68077

Direct link to sweep crash file:

https://forum.freecadweb.org/download/file.php?id=186806

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6761">pipeline status</a> for feature branch [PR_6761](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6761). Pipeline [533569411](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/533569411) was triggered at [5b02da7](https://github.com/FreeCAD/FreeCAD/pull/6761/commits/5b02da75f3b2659ba496349390cea495f5ff2c2a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@realthunder can you please review this Link-specific PR?

realthunder (commented):
I am fine with the code, except the `ResetPlacement` part. I think this should default to `True` for backward compatibility. If you think `False` is the preferred the behavior, then add a function `RuledSurface::setupObject()` and set it to `False` there, so that only the newly created object has the new behavior.

mwganson:
The workaround works for new objects, but fails for some reason with existing objects.  I created a Part::Box in 0.19, transformed it, and made a ruled surface from 2 adjacent edges.  Then when opening that saved file in this compiled PR the ruled surface moves to the wrong location upon recompute.  But creating a new ruled surface from the same edges produces the expected behavior.

realthunder:
The default behavior in 0.19 is as if `ResetPlacement = True` if I am not mistaken, but then you chose `False` as the default value of `ResetPlacement`, which means all legacy RuledFace feature created in older version FreeCAD will have a different behavior once recomputed.

donovaly:
For info: this will be the final week to finish this PR otherwise it cannot go in for 0.20.

mwganson:
> For info: this will be the final week to finish this PR otherwise it cannot go in for 0.20.

Okay.  I will work on reverting back to what it was before adding the support for App::Link.  The current situation is not acceptable, with the Ruled Surface sometimes appearing in the wrong location.  Perhaps it can be re-examined during 0.21 cycle.

mwganson:
> > For info: this will be the final week to finish this PR otherwise it cannot go in for 0.20.
> 
> Okay. I will work on reverting back to what it was before adding the support for App::Link. The current situation is not acceptable, with the Ruled Surface sometimes appearing in the wrong location. Perhaps it can be re-examined during 0.21 cycle.

After doing this I have found a bug that I could not figure out also pre-existed before adding the support for App::Link to Ruled Surface.  The bug is if the user moves the ruled surface object by changing its placement, then upon recompute the ruled surface moves back to where it was.

To reproduce (using 0.19) in Part workbench create a new Cube.  Transform the cube to x=5, y=5, z=0.  Select 2 edges of the cube and create a ruled surface.  Transform the ruled surface to x=5, y=5, z=5 (from z=0).  Right click the document label in the combo view, select mark to recompute, click refresh.  Upon recompute the ruled surface moves back to its original position.

mwganson:
I feel this is now ready to be merged.

One thing to consider is some models created with dev version of 0.20 might now be broken if they contain ruled surfaces created using app::links.  But I think the previous situation with the ruled surfaces not appearing the proper location is unacceptable.  If we cannot make app::links work properly with ruled surfaces then it is better to not support them.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6761">pipeline status</a> for feature branch [PR_6761](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6761). Pipeline [539562004](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/539562004) was triggered at [534b0cd](https://github.com/FreeCAD/FreeCAD/pull/6761/commits/534b0cd8c0a2820fce8383066781a5adbb1b1eec). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@realthunder , is the PR also ready to merge from your side?

realthunder:
@mwganson I have just taken a closer look at this 'reset placement' workaround. I don't think it is necessary. The root cause of the whole problem is that the original workaround of an [OCC bug](https://github.com/FreeCAD/FreeCAD/blob/45a0e6eb205a41abe5bd0649d14eee3334ed3873/src/Mod/Part/App/PartFeatures.cpp#L143) is kind of wrong. The bug can be worked around by simply [copy](https://github.com/realthunder/FreeCAD/blob/8dedb73c21d301eb10290bfee380b0eb13a12cdf/src/Mod/Part/App/PartFeatures.cpp#L148) the sub-shape. See my suggested fix [here](https://github.com/realthunder/FreeCAD/commit/8dedb73c21d301eb10290bfee380b0eb13a12cdf)

mwganson:
@realthunder If you can submit that PR I can close this one.  Would that resolve the issue?

donovaly:
For information: it seems we have now a working build system for Windows. This was the last known blocker for a release. Therefore I would close the merging next Monday.
I hope that this or another PR is ready before so that I can merge.

donovaly:
Closing in favor of #6920

donovaly:
obviously the closing on 27 May did not work. Closing again.

PR #6779, opened 2022-04-22
Spreadsheet: fix extreme slow down with large empty cells

realthunder:
https://forum.freecadweb.org/viewtopic.php?p=590151#p590151

### DIFF
--- a/src/Mod/Spreadsheet/App/Cell.cpp
+++ b/src/Mod/Spreadsheet/App/Cell.cpp
@@ -288,6 +288,8 @@ void Cell::setContent(const char * value)
     clearException();
     if (value) {
         if (owner->sheet()->isRestoring()) {
+            if (value[0] == '\0' || (value[0] == '\'' && value[1] == '\0'))
+                return;
             expression.reset(new App::StringExpression(owner->sheet(), value));
             setUsed(EXPRESSION_SET, true);
             return;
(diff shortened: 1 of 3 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6779">pipeline status</a> for feature branch [PR_6779](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6779). Pipeline [522592456](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/522592456) was triggered at [9f5d3ca](https://github.com/FreeCAD/FreeCAD/pull/6779/commits/9f5d3caf2e84764ffef22ed8a117a39fcb317281). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Very nice -- this works as advertised, and I didn't note any regressions. I will merge it as-is, but I have a question: the change you made to the paste code, that switches to the `rangeUpdated` -- can that idea also be applied to other changes? For example, when selecting an entire column and centering it, it's quite slow, and I'm not sure if it's just the actual act of updating the cell formatting that's slow, or if we're doing something similar there, and calling `cellUpdated` every time instead of one call to `rangeUpdated` at the end.

realthunder:
I checked. I guess you are using debug build, because the direct cause of slow down on row alignment change is due to log output. Release build is not affected as it has lower default log output level. I have submitted a [PR](https://github.com/FreeCAD/FreeCAD/pull/6782) to change log level for those two places.

However, there is also indirect cause. The row alignment change causes the empty cells to be created for the entire row. They are not exposed as properties because of this PR, but the cells are still created causing bloated file and processing overhead. I think we should address this issue after 0.20 release. Maybe add some row/column default style settings.

PR #6806, opened 2022-04-27
Draft: Substitute diacritic string Bézier for Bezier

luzpaz:
Closes https://github.com/FreeCAD/FreeCAD-translations/issues/122

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_beziers.py
+++ b/src/Mod/Draft/draftguitools/gui_beziers.py
@@ -24,7 +24,7 @@
 # ***************************************************************************
 """Provides GUI tools to create BezCurve objects.
 
-In particular, a cubic Bézier curve is defined, as it is one of the most
+In particular, a cubic Bezier curve is defined, as it is one of the most
 useful curves for many applications.
 
 See https://en.wikipedia.org/wiki/B%C3%A9zier_curve
@@ -50,7 +50,7 @@ from draftutils.translate import translate
 
 
 class BezCurve(gui_lines.Line):
-    """Gui command for the Bézier Curve tool."""
+    """Gui command for the Bezier Curve tool."""
 
     def __init__(self):
         super(BezCurve, self).__init__(wiremode=True)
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6806">pipeline status</a> for feature branch [PR_6806](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6806). Pipeline [526315568](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/526315568) was triggered at [bcc960d](https://github.com/FreeCAD/FreeCAD/pull/6806/commits/bcc960dc696571320d1f621d3aebdbf0fca790e7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

yorikvanhavre:
I'm not sure what to do here... I think if the correct English way is to write "Bézier", we should leave "Bézier" and fix the encoding. But since Bézier is a famous name and English language doesn't have é, maybe it's more correct to write ¨Bezier"? Any linguist around? :)

luzpaz:
IMHO, I think it's probably best to merge since the diacritic spelling of the word Is the exception currently within the FC source base. We can come back later and fix the encoding issue and then decide if we want to purposely go into the source and alter the string to add the diacritic mark

luzpaz:
The more I think about it, the more we should merge this. Pinging any merge dev

yorikvanhavre:
Honestly I still think it's best to fix the ts files. Let me see if I can get to something...

luzpaz:
@yorikvanhavre 

Compare: 
https://github.com/FreeCAD/FreeCAD/search?p=6&q=bezier
Vs.  
https://github.com/FreeCAD/FreeCAD/search?q=B%C3%A9zier

yorikvanhavre:
I know, this is inconsistent... Let me try to find a better fix, and if I can't then we merge this as is.

yorikvanhavre:
Ok the problem here is pyside2-lupdate. Using pylupdate5, there is no more encoding problem

yorikvanhavre:
@chennes BTW, both pylupdate5 and pyside2-lupdate still give a reduction of the number of strings after your last update of updatets.py (just doing `grep "<message>" src/Gui/Language/FreeCAD.ts | wc -l`)
| file | before | after |
| --- | --- | --- |
| FreeCAD.ts | 2147 | 2096 |
| Draft.ts | 1466 | 1415 | 
| Arch.ts | 1254 | 1219 |

Do you get a different result?

chennes:
@yorikvanhavre Those are the results I'm seeing too -- how can we find a string that's been removed to see if it's correct?

chennes:
@yorikvanhavre I extracted just the source strings from FreeCAD.ts (new and old), sorted them, and ran diff on the result. I've attached the diff file: it looks to me like the removed strings are being removed correctly, but I only spot-checked a few.
[diff.txt](https://github.com/FreeCAD/FreeCAD/files/8621182/diff.txt)

yorikvanhavre:
Yes, I did a more thorough check now, and indeed everything I could find are removal of duplicates. I pushed the new ts files to ab5a6d83a1 and doing a crowdin merge cycle now, let's see...

chennes:
Nice! That's a big update at CrowdIn, I bet it closes a lot of the open FreeCAD-translations issues.

luzpaz:
Thank fellas!

PR #6808, opened 2022-04-28
Offscreen rendering

wwmayer:

---

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -77,6 +77,7 @@
 #include "MDIViewPy.h"
 #include "SoFCDB.h"
 #include "Selection.h"
+#include "SoFCOffscreenRenderer.h"
 #include "SplitView3DInventor.h"
 #include "TaskView/TaskView.h"
 #include "TaskView/TaskDialogPython.h"
@@ -456,6 +457,10 @@ Application::Application(bool GUIenabled)
             break;
         Py_DECREF(descr);
     }
+
+    SoQtOffscreenRendererPy::init_type();
+    Base::Interpreter().addType(SoQtOffscreenRendererPy::type_object(),
+        module,"SoQtOffscreenRenderer");
 
     App::Application::Config()["COIN_VERSION"] = COIN_VERSION;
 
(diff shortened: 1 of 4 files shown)


luzpaz:
@wwmayer this looks interesting, what does it do?

wwmayer:
It renders a scene off-screen :)

Coin has an integrated off-screen rendering mechanism but it always lacked the possibility to support anti-aliasing or correctly handle transparent backgrounds and for several years it has stopped working for me completely.

So, the alternative we have implemented several years ago is using Qt's off-screen rendering capabilities combined with framebuffer objects.

Now this PR only exposes the corresponding C++ class to Python. This allows it to create high-quality images of a scene in true headless mode and thus could be used as the groundwork for a thumbnailer for XFCE that supports any file format that FreeCAD supports.

wwmayer:
What do you mean with alternative GUI?

luzpaz:
> What do you mean with alternative GUI?

Meaning, not using the FreeCAD regular GUI, but like a unix pipe (if you will) to a dialog of some sort showing the model being manipulated through the headless instance of FreeCAD

wwmayer:
You could show the rendered image in a widget and use some buttons to control the headless FreeCAD. But an interaction with the scene won't be possible and also the rendering might be a bit slow.

yorikvanhavre:
Fantastic! There are a lot of possibilities there...

I also think it would be impractical to run FreeCAD in another GUI that way, as it would need to re-render the scene 30 times per second or something like that, no sure that would be achievable... Better path would be to export the coin scene directly to some other format (webgl, etc...) and let the other app do the rendering

luzpaz:
CC @ickby (you may be interested in this since you've been working on https://github.com/OpenCollaborationPlatform/CollaborativeFC)

adrianinsaval:
does that mean it could be also used to generate thumbnails even for FCStd files that don't have the embedded thumbnail?

wwmayer:
> does that mean it could be also used to generate thumbnails even for FCStd files that don't have the embedded thumbnail?

Basically yes. However, it's not my primary goal to implement a general thumbnailer. I just wanted to make the real off-screen rendering working again. The thumbnailer is just a side-product of it.

For a thumbnailer there are a few points to consider:
* The thumbnailer would have to load FreeCAD in headless mode and if there is a directory with hundreds of files I am not sure how much it would freeze the system
* If a STEP, STL or any other data file is relatively big then loading it can be quite time consuming. So, there should be a certain limit of the file size for that an image is created.

PR #6839, opened 2022-05-06
Gui: improve NaviCube rendering

realthunder:
Related #6689

Issue #6689, opened 2022-03-30
[Bug] NaviCube does not respond to MSAA settings in preferences.


obelisk79:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=34&t=67442

### Version

0.20 (Development)

### Full version info

Unable to paste the full version info.

I have tested on Linux and Windows and using several weekly builds from the last 6 months 
to verify this issue using both a laptop and desktop computer.

### Issue description

In the UI on 1080p screens it is readily apparent that the NAVICUBE does not respond to the MSAA settings in preferences. This renders the face text difficult to read at off-angles and the unrefined edges of the navicube are incongruent with the more refined appearance of the rest of the UI.

![image](https://user-images.githubusercontent.com/4140247/165527571-f90ee8dd-baef-4251-891f-5c8a6f8272e6.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

obelisk79:
So, I did some experimenting after looking up the QPainter QT5 class, and there is the setRenderHints() method which seems to have antialiasing options.

However, when I tried adding the following at line 356 just below paint.begin(&image);

QPainter paint.setRenderHints(QPainter::Antialiasing, true) 

I don't see any anti-aliasing being applied. is there a better approach for this?

donovaly:
This is a known issue and also in FreeCAD 0.19

luzpaz:
![Screenshot_20220427_094704](https://user-images.githubusercontent.com/4140247/165533471-1c42c505-71f6-45f8-af4a-2b5f06559ed6.png)
Kinda looks the same on @realthunder's branch as well.

realthunder:
@luzpaz Did you turn on MSAA. Increasing border width will make the effect more obvious. Just compare it with the axis line. The border of the cube is rendered as texture in upstream if I remembered correctly, which is why it does not respond to MSAA setting. I have changed it to OpenGL line rendering.

https://user-images.githubusercontent.com/1207888/165647225-a8963957-885b-47af-9c81-b0e6bde0d11c.mp4

donovaly:
@realthunder , can you please make a PR to fix the issue also for master? This PR should go in for 0.20 is possible.

PR #6848, opened 2022-05-08
Gui: improve NaviCube rendering (simplified)

realthunder:
Related #6689 #6839

### DIFF
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -85,6 +85,7 @@ public:
 	GLuint m_TextureId;
 	QColor m_Color;
 	int m_PickId;
+	int m_PickTexId;
 	GLuint m_PickTextureId;
 	int m_RenderPass;
 	Face(
@@ -92,6 +93,7 @@ public:
 		 int vertexCount,
 		 GLuint textureId,
 		 int pickId,
+		 int pickTexId,
 		 GLuint pickTextureId,
 		 const QColor& color,
 		 int  renderPass
@@ -101,6 +103,7 @@ public:
 		m_VertexCount = vertexCount;
 		m_TextureId = textureId;
 		m_PickId = pickId;
+		m_PickTexId = pickTexId;
 		m_PickTextureId = pickTextureId;
 		m_Color = color;
 		m_RenderPass = renderPass;
@@ -131,7 +134,7 @@ private:
 	void setHilite(int);
 
 	void initNaviCube(QtGLWidget*);
-	void addFace(const Vector3f&, const Vector3f&, int, int, int, bool flag=false);
+	void addFace(float gap, const Vector3f&, const Vector3f&, int, int, int, bool flag=false);
 
 	GLuint createCubeFaceTex(QtGLWidget* gl, float gap, const char* text, int shape);
 	GLuint createButtonTex(QtGLWidget*, int);
@@ -204,11 +207,13 @@ public:
 	QColor m_HiliteColor;
 	QColor m_ButtonColor;
 	QColor m_FrontFaceColor;
+	QColor m_BorderColor;
 	int m_HiliteId = 0;
 	bool m_MouseDown = false;
 	bool m_Dragging = false;
 	bool m_MightDrag = false;
-    NaviCube::Corner m_Corner = NaviCube::TopRightCorner;
+	double m_BorderWidth;
+	NaviCube::Corner m_Corner = NaviCube::TopRightCorner;
 
 	QtGLFramebufferObject* m_PickingFramebuffer;
 
@@ -217,6 +222,7 @@ public:
 	vector<GLubyte> m_IndexArray;
 	vector<Vector2f> m_TextureCoordArray;
 	vector<Vector3f> m_VertexArray;
+	vector<Vector3f> m_VertexArray2;
 	map<int,GLuint> m_Textures;
 	vector<Face*> m_Faces;
 	vector<int> m_Buttons;
@@ -269,6 +275,8 @@ NaviCubeImplementation::NaviCubeImplementation(
 	OnChange(*hGrp, "HiliteColor");
 	OnChange(*hGrp, "ButtonColor");
 	OnChange(*hGrp, "CubeSize");
+	OnChange(*hGrp, "BorderWidth");
+	OnChange(*hGrp, "BorderColor");
 
 	m_PickingFramebuffer = nullptr;
 	m_Menu = createNaviCubeMenu();
@@ -301,7 +309,11 @@ void NaviCubeImplementation::OnChange(ParameterGrp::SubjectType &rCaller, Parame
 		m_ButtonColor.setRgba(rGrp.GetUnsigned(reason, QColor(226, 233, 239, 128).rgba()));
 	} else if (strcmp(reason,"CubeSize") == 0) {
 		m_CubeWidgetSize = (rGrp.GetInt(reason, 132));
-	}
+	} else if (strcmp(reason,"BorderWidth") == 0) {
+		m_BorderWidth = rGrp.GetFloat("BorderWidth", 1.5);
+	} else if (strcmp(reason,"BorderColor") == 0) {
+		m_BorderColor.setRgba(rGrp.GetUnsigned(reason, QColor(50, 50, 50, 255).rgba()));
+    }
 }
 
 char* NaviCubeImplementation::enum2str(int e) {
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6848">pipeline status</a> for feature branch [PR_6848](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6848). Pipeline [533777842](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/533777842) was triggered at [a06d76a](https://github.com/FreeCAD/FreeCAD/pull/6848/commits/a06d76a4a2d67893ab07ce8368c3293ccfafad98). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Many thanks! I will merge on Wednesday when I am back home.

I will also check if @Jolbas suggestions from #6839 can be applied too.

src/Gui/NaviCube.cpp
```diff
@@ -269,6 +275,8 @@ NaviCubeImplementation::NaviCubeImplementation(
 	OnChange(*hGrp, "HiliteColor");
 	OnChange(*hGrp, "ButtonColor");
 	OnChange(*hGrp, "CubeSize");
+	OnChange(*hGrp, "BorderWidth");
+	OnChange(*hGrp, "BorderColor");
```
donovaly, line 279: Why is the BorderColor important to fix the rendering? The PR fixes the issue also without a border color.

I committed this now nevertheless but please let us only change things that have to be changed to fix an issue.

src/Gui/NaviCube.cpp
```diff
@@ -301,7 +309,11 @@ void NaviCubeImplementation::OnChange(ParameterGrp::SubjectType &rCaller, Parame
 		m_ButtonColor.setRgba(rGrp.GetUnsigned(reason, QColor(226, 233, 239, 128).rgba()));
 	} else if (strcmp(reason,"CubeSize") == 0) {
 		m_CubeWidgetSize = (rGrp.GetInt(reason, 132));
-	}
+	} else if (strcmp(reason,"BorderWidth") == 0) {
+		m_BorderWidth = rGrp.GetFloat("BorderWidth", 1.5);
```
donovaly, line 313: This looked clumsy for me, so I had to change it from 1.5 to 1.1

donovaly:
I committed this now plus some minor fixes and some of the suggestions by @jolbas: commit f70329390cea.

However, we now have a minor regression:
- set the cube size in the preferences to 500
- result:
FreeCAD_TDV64QZpsd

So the corner frame is thicker than the square lines. It should have the same width. Can you please have a look? I could not yet find a solution.

donovaly:
> I committed this now plus some

-> closing this PR as being merged

donovaly:
> I committed this now plus some minor fixes and some of the suggestions by @Jolbas: commit [f703293](https://github.com/FreeCAD/FreeCAD/commit/f70329390ceaac95200d6a58b1a6ff2c4e9f3ed5).
> 
> However, we now have a minor regression:
> 
>     * set the cube size in the preferences to 500
> 
>     * result:
>       FreeCAD_TDV64QZpsd
> 
> 
> So the corner frame is thicker than the square lines. It should have the same width. Can you please have a look? I could not yet find a solution.

It is really weird because when I restart FC, I get every time a different width for the corner frame. Sometimes it is correct for a 500 px cube, but not for a 150 px cube, sometimes it is the opposite. Therefore I think we cannot do much.

realthunder:
The corner border is still drawn with texture. When the cube size is changed, the texture is not recreated, but just scaled to match the new cube size. That's why you see the corner border got thickened. When you open a new document, the texture size is determined by the new cube size, so no scaling there. This is one of the problem fixed in my first NaviCube PR.

In theory, we can use opengl to draw the triangle borders using various cube face vertices. I didn't do that, because I thought the triangle corner face is too small to click easily, especially if it is not front facing. So I reverted it back to the circular face as in 0.19. Please note that I did all these a while ago, so is the border color addition. I just wanted to minimize changes (against my existing implementation) to avoid unexpected error. This is the reason I included the corner shape change in my second NaviCube PR.

Allow me to make some complain here. Lack of support of anti-alias is arguably not a bug in the first place. I don't understand why you have to be so strict about bug vs. feature distinguish, while in reality many changes don't have clear cut differentiation. I doubt any one will complain about unfairness because of those changes. I thought this feature close thingy is kind of flexible. You don't see me complain when some of my feature PRs didn't make it despite being submitted in time.

src/Gui/NaviCube.cpp
```diff
@@ -380,8 +393,6 @@ GLuint NaviCubeImplementation::createCubeFaceTex(QtGLWidget* gl, float gap, cons
 		QPainterPath pathSquare;
 		pathSquare.addRect(QRectF(gapi, gapi, (qreal)texSize - 2.0 * gapi, (qreal)texSize - 2.0 * gapi));
 		paint.fillPath(pathSquare, Qt::white);
-		paint.setPen(pen);
-		paint.drawPath(pathSquare);
 	}
 	else if (shape == SHAPE_CORNER) {
 		QPainterPath pathCorner;
```
Jolbas, line 398: Couldn't comment on line 399 but try change line 399 to
```
                qreal midTex = (qreal)texSize / 2.0;
		QRectF rectCorner = QRectF(midTex - sqrt(0.5) * gapi, midTex - sqrt(2/3) * gapi, sqrt(2) * gapi, sqrt(1.5) * gapi);
```
or possibly
```
                qreal midTex = (qreal)texSize / 2.0;
		QRectF rectCorner = QRectF(midTex - sqrt(0.5) * gapi, midTex - sqrt(1/6) * gapi, sqrt(2) * gapi, sqrt(1.5) * gapi);
```
Some of these two should be right, but I can't test by myself.

donovaly: The first leads to this:
FreeCAD_SB1ac5f1fy

the second one to this:
FreeCAD_nsJOFeAvt7

donovaly:
> When you open a new document, the texture size is determined by the new cube size, so no scaling there

Thanks, I understand. But since the cube size is rarely changed, we can live with this I think.

> because I thought the triangle corner face is too small to click easily, especially if it is not front facing. So I reverted it back to the circular face as in 0.19. 

Allow me to complain here ;-): There was a decision for 0.20 to use corner triangles because this way one cannot accidentally click the faces from behind. This was a problem of the old cube. The change was discussed and now 0.20 is almost done. Therefore I don't want to have a change that reverts a decision. My approach is to discuss these changes with the community. What I learned the hard way during the 0.19 release process is that there are many displays around and since different users have different needs, decisions that seem obvious, can quickly introduce issues for other users.
Therefore every design change should be discussed. (We have users of RTL languages, different resolutions, color blind persons, trackball users... )

> Lack of support of anti-alias is arguably not a bug in the first place.

This issue popped up since 0.18 when I remember correctly. Also during the new cube development is popped up and now again. And I can confirm the look was ugly on small Hi-DPI laptop screens. So it is an issue that affect different people and it is something you immediately notice when starting FC the first time, there fore also a UX issue.

> I doubt any one will complain about unfairness because of those changes. 

I hope so. My background is 10 years of the project lyx.org and there people often went crazy when even small features were introduces in freeze times. Now it is the first time I am a release manager at all, so I have to learn.
I am happy that you are not upset that not all your PRs went in despite sent in time.

src/Gui/NaviCube.cpp
Jolbas: Then we have to leave it. I did test the two versions out of scope along with the original. And my first version returns very similar results to the original as long as the gap is set to 0.12. Clearly there is some side effects I don't understand. Thank you. I need to set up my own build system before I go further.

adrianinsaval:
This is kinda not the place to discuss it but I think I agree with realthunder here, perhaps there is space to be a little more flexible for cases like this were the line is blurred and it's hard to separate fix from feature, I totally understand uwe's reservations considering how sensitive some people can get though. (And i'll take this opportunity to once again thanks both of you for all the hard work you put into this project)

Regarding the corner shape, IIRC the problem was that the chamfered edges didn't have filling and weren't clickable, and once they were filled you preferred triangles over circles because they fit the shape better, I could easily be wrong here but I don't remember any technical reason why triangles were preferred over circles, personally I like this recent request to make the vertex hexagonal instead triangular: https://forum.freecadweb.org/viewtopic.php?f=34&t=68544 they are more functional than triangles and IMO look better than circles and triangles, further discussion is probably better done either there or on a new thread so the non dev part of the community also engages.

donovaly:
> I don't remember any technical reason why triangles were preferred over circles

Geometry - the problem was that the circles can accidentally clicked from behind since they will always protrude over the actual cube. I had a look what other CAD programs do and this is for example how it is done in OCC's CAD Assistant:
firefox_EqPYiDex4S

This is also a proposal that should do the job:
https://forum.freecadweb.org/viewtopic.php?p=593838#p593838

However, 0.20 has to be done and this is feature-froozen.

PR #6899, opened 2022-05-18
Draft: fix issues with grid spacing parameter

0penBrain:
Fixes issues introduced with #5427 // 6157147

https://forum.freecadweb.org/viewtopic.php?f=23&t=68675

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_selectplane.py
+++ b/src/Mod/Draft/draftguitools/gui_selectplane.py
@@ -96,7 +96,10 @@ class Draft_SelectPlane:
 
         # Fill values
         self.taskd.form.checkCenter.setChecked(self.param.GetBool("CenterPlaneOnView", False))
-        q = FreeCAD.Units.Quantity(self.param.GetFloat("gridSpacing", 1.0), FreeCAD.Units.Length)
+        try:
+            q = FreeCAD.Units.Quantity(self.param.GetString("gridSpacing", "1 mm"))
+        except ValueError:
+            q = FreeCAD.Units.Quantity("1 mm")
         self.taskd.form.fieldGridSpacing.setText(q.UserString)
         self.taskd.form.fieldGridMainLine.setValue(self.param.GetInt("gridEvery", 10))
         self.taskd.form.fieldGridExtension.setValue(self.param.GetInt("gridSize", 100))
@@ -477,7 +480,7 @@ class Draft_SelectPlane:
         except Exception:
             pass
         else:
-            self.param.SetFloat("gridSpacing", q.Value)
+            self.param.SetString("gridSpacing", q.UserString)
             if hasattr(FreeCADGui, "Snapper"):
                 FreeCADGui.Snapper.setGrid()
 
--- a/src/Mod/Draft/Resources/ui/TaskSelectPlane.ui
+++ b/src/Mod/Draft/Resources/ui/TaskSelectPlane.ui
@@ -171,7 +171,7 @@ will be moved to the center of the view</string>
         <string>The spacing between the smaller grid lines</string>
        </property>
        <property name="unit" stdset="0">
-        <string notr="true"/>
+        <string notr="true">mm</string>
        </property>
       </widget>
      </item>
--- a/src/Mod/Draft/draftguitools/gui_trackers.py
+++ b/src/Mod/Draft/draftguitools/gui_trackers.py
@@ -1210,7 +1210,10 @@ class gridTracker(Tracker):
 
     def reset(self):
         """Reset the grid according to preferences settings."""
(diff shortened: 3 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6899">pipeline status</a> for feature branch [PR_6899](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6899). Pipeline [541767601](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/541767601) was triggered at [aaeabc9](https://github.com/FreeCAD/FreeCAD/pull/6899/commits/aaeabc9c6d49c2f4132dc81d56fa4a53d57d6ebf). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Draft/draftguitools/gui_selectplane.py
```diff
@@ -96,7 +96,10 @@ def Activated(self):
 
         # Fill values
         self.taskd.form.checkCenter.setChecked(self.param.GetBool("CenterPlaneOnView", False))
-        q = FreeCAD.Units.Quantity(self.param.GetFloat("gridSpacing", 1.0), FreeCAD.Units.Length)
+        try:
+            q = FreeCAD.Units.Quantity(self.param.GetString("gridSpacing", "1 mm"))
+        except ValueError:
+            q = FreeCAD.Units.Quantity("1 mm")
```
yorikvanhavre, line 102: I just saw that it's possible to have two values with the same name:
Screenshot from 2022-05-19 13-10-09
What about this:
```suggestion
        try:
            q = FreeCAD.Units.Quantity(self.param.GetString("gridSpacing", "1 mm"))
        except ValueError:
            # older format, for backwards compatibility
            q = FreeCAD.Units.Quantity(self.param.GetFloat("gridSpacing", 1.0), FreeCAD.Units.Length)
```
Since the value is then saved as string, next time everything will be fine. But maybe it's not worth it, after all it's not terrible if the grid is reset to 1... 

This needs to be changed too: https://github.com/yorikvanhavre/BIM_Workbench/blob/master/BimSetup.py#L144 and L343

0penBrain: > What about this:

No strong opinion on my side. :) I'm not really a Draft user. Generally, I don't like to have to maintain code written just for transitional things.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6899">pipeline status</a> for feature branch [PR_6899](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6899). Pipeline [547594156](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/547594156) was triggered at [4474db7](https://github.com/FreeCAD/FreeCAD/pull/6899/commits/4474db7f50d7aa4f30ef5b7dbd8dc1c61e91bde6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Roy-043:
For the gridSpacing parameter it would be acceptable to start with the default value when switching to V0.20, and ask the user to change it if desired. But there are several (more than 10) other 'distance type' Draft parameters (f.e. textheight) which I assume will see a similar update (float -> quantity string) in the near future for reasons of consistency.

Instead of handling this throughout the code why not create a single update_params function (in utils.py) to be called from the Initialize method of the workbench?

```
def update_params():
    """Update parameters when a new FreeCAD version requires it.
    To be called from the Initialize method of the workbench.
    For now only 1 parameter is handled.
    """
    
    draft_params = "User parameter:BaseApp/Preferences/Mod/Draft"
    p = App.ParamGet(draft_params)

    # "gridSpacing" float to string (V0.19 -> V0.20).
    if ((not "gridSpacing" in p.GetStrings())
            and "gridSpacing" in p.GetFloats()):
        if p.GetFloat("gridSpacing") > 0:
            val = str(p.GetFloat("gridSpacing")) + " mm"
        else:
            val = "1 mm"
        p.SetString("gridSpacing", val)
```

yorikvanhavre:
The problem is that it'll be hard to track all the ones that have changed...
I propose to merge this one now, the way it is, as

1) it solves the issue for now, with minimal change before the release
2) the "grid reset" that will happen is not severe
3) I guess all other areas affected by the preferences field changes are not too important either, so it's not clear if the code proposed by @Roy-043 above is really necessary... I propose we look further into it later

PR #6916, opened 2022-05-25
Fix and simplify column number string to number conversion

berniev:
```
Before refactor:

    In  Silent      Expect      Actual   Comment
             0   exception          26  * FAIL *
     A       0           0           0          
     Z       0          25          25          
    AA       0          26          26          
    AB       0          27          27          
    AZ       0          51          51          
    BA       0          52          52          
    CB       0          79          79          
    ZA       0         676         676          
    ZZ       0         701         701          
   AAA       0         702          26  * FAIL *
   AAZ       0         727          51  * FAIL *
   CBA       0        2080        1404  * FAIL *
   AZA       0        1352         676  * FAIL *
   ZZA       0       18252       17576  * FAIL *
   ZZZ       0       18277       17601  * FAIL *
   ALL       0         999         323  * FAIL *
    Ab       0   exception   exception          
  ABCD       0   exception         757  * FAIL *
             1          -1          26  * FAIL *
    Ab       1          -1          -1          
  ABCD       1          -1         757  * FAIL *

After refactor: All tests pass.
```


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6916">pipeline status</a> for feature branch [PR_6916](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6916). Pipeline [547474035](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/547474035) was triggered at [5b32c1f](https://github.com/FreeCAD/FreeCAD/pull/6916/commits/5b32c1fc19eef542a2c660ba736286e04cadbf7f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6916">pipeline status</a> for feature branch [PR_6916](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6916). Pipeline [548458941](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/548458941) was triggered at [0a8b875](https://github.com/FreeCAD/FreeCAD/pull/6916/commits/0a8b875751c9469a52548730539dfd4b4fc51acc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
```shell
git rebase upstream master
git push origin --force
```

upstream: FreeCAD master
origin: your GitHub clone of upstream master

berniev:
Replaces PR 6783. Trying (unsuccessfully so far) to fix problem I created in git.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6916">pipeline status</a> for feature branch [PR_6916](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6916). Pipeline [550578197](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/550578197) was triggered at [0d12273](https://github.com/FreeCAD/FreeCAD/pull/6916/commits/0d12273fde48c1d18d58f2007b399659f5a2f0b2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berniev:
This is not the result I wanted. I've got all this crap in master and it keeps biting me. Maybe I should delete all and start again.

luzpaz:
Don't get too frustrated :stuck_out_tongue_winking_eye:  PM me on the forum I'm 'kunda1' there.

chennes:
@berniev am I correct in understanding that 0a8b875751c9469a52548730539dfd4b4fc51acc is the only commit that matters here?

chennes:
@berniev How are you testing this code, given that MAX_COLS is hardcoded to 26*26+26 (e.g. only one- and two-letter columns allowed)?

chennes:
No worries about the extra commits, it's easy enough for us to cherry-pick the one you intended. And it's clear that your code is superior (which is to say, it actually works!) for three-character column references. But right now FreeCAD does not permit three-character columns, it's capped at two, so I'm mostly curious about what use-case justifies the modification. We're in a feature-freeze right now in the lead-up to next Monday's release of 0.20, and I'm hesitant to merge anything that doesn't directly address a known bug.

berniev:
In terms of pure bug, if it can be guaranteed this code never sees more than 2chr input, then the only bug is the case of empty string, so no rush to merge. Might be able to do a better fix later.

But what’s the point in having a function that tests if a string is valid?

Having the conversion broken into two parts caught my eye as a code smell. Just looked too complicated. Comments tried to give names to parameters that are just strings, and the function names did not express what they do. And why is there even a check necesary every time a string is converted? Quite a mess. 

I did see the limit stated somewhere, but there appears no way to determine if it is relevant. In any event the code produced results for 3chr string (and 4chr)  even if they were wrong.

MacOS’s Numbers goes up to “ALL” (999) which is cute. Excel I think goes to 16384, so ZZ (701) seemed small.

Better if column label be an immutable (value) object containing conversion, memoisation etc, then this all goes away.

donovaly:
> In terms of pure bug, if it can be guaranteed this code never sees more than 2chr input, then the only bug is the case of empty string, so no rush to merge. 

OK, setting its milestone to 0.21.

chennes:
@wwmayer this seems like an obvious thing to add to our unit test suite, but I don't see a clear path forward for doing so, the decode function doesn't seem to be directly accessible to our Python code and I am not sure what code I could write that would test it. Do you have a suggestion for how to go about it?

donovaly:
What is the status of this. Can it be merged to master?

berniev:
Would you mind if I replace this PR with a new one I have ready to go? (I messed up my git, slowly learning).

chennes:
@berniev That's fine, just reference this one when you submit it.

berniev:
replaced by #7250

wwmayer:
@chennes 

> @wwmayer this seems like an obvious thing to add to our unit test suite, but I don't see a clear path forward for doing so. The decode function doesn't seem to be directly accessible to our Python code and I am not sure what code I could write that would test it. Do you have a suggestion for how to go about it?

In the forum it has been discussed to add the infrastructure for TDD to FreeCAD and IMO this would be the ideal way. However, this will take some time to have it done and in the meantime we must do it via Python.

Because I don't like it to directly expose the used functions to Python (the functions are not supposed to be used from Python) the best way for now is to implement a test feature class (e.g. FeatureTestRange) and control the input and output parameters via properties.

wwmayer:
0dddd3336e

PR #6927, opened 2022-05-28
Gui: fix number display issue of color bar

wwmayer:

---

### DIFF
--- a/src/Gui/SoFCColorBar.cpp
+++ b/src/Gui/SoFCColorBar.cpp
@@ -23,9 +23,11 @@
 #include "PreCompiled.h"
 
 #ifndef _PreComp_
+# include <Inventor/actions/SoGetBoundingBoxAction.h>
 # include <Inventor/actions/SoGLRenderAction.h>
 # include <Inventor/events/SoMouseButtonEvent.h>
 # include <Inventor/nodes/SoEventCallback.h>
+# include <Inventor/nodes/SoOrthographicCamera.h>
 # include <Inventor/nodes/SoSwitch.h>
 # include <QApplication>
 # include <QMenu>
@@ -76,6 +78,32 @@ void SoFCColorBarBase::GLRenderBelowPath(SoGLRenderAction *  action)
         setViewportSize(size);
     }
     SoSeparator::GLRenderBelowPath(action);
+}
+
+float SoFCColorBarBase::getBoundingWidth(const SbVec2s& size)
+{
+    // These are the same camera settings for front nodes as defined in the 3d view
+    SoOrthographicCamera* cam = new SoOrthographicCamera;
+    cam->position = SbVec3f(0, 0, 5);
+    cam->height = 10;
+    cam->nearDistance = 0;
+    cam->farDistance = 10;
+
+    SoGroup* group = new SoGroup();
+    group->ref();
+    group->addChild(cam);
+    group->addChild(this);
+
+    SbViewportRegion vpr(size);
+    SoGetBoundingBoxAction bbact(vpr);
+    bbact.apply(group);
+    SbBox3f box = bbact.getBoundingBox();
+    SbVec3f minPt, maxPt;
+    box.getBounds(minPt, maxPt);
+    group->unref();
+
+    float boxWidth = maxPt[0] - minPt[0];
+    return boxWidth + 0.2f;
 }
 
 // --------------------------------------------------------------------------
--- a/src/Gui/SoFCColorGradient.cpp
+++ b/src/Gui/SoFCColorGradient.cpp
@@ -110,18 +110,22 @@ void SoFCColorGradient::setMarkerLabel(const SoMFString& label)
 
 void SoFCColorGradient::setViewportSize(const SbVec2s& size)
 {
-    // don't know why the parameter range isn't between [-1,+1]
-    float fRatio = ((float)size[0]) / ((float)size[1]);
+    float fRatio = static_cast<float>(size[0]) / static_cast<float>(size[1]);
     float fMinX =  4.0f, fMaxX = 4.5f;
     float fMinY = -4.0f, fMaxY = 4.0f;
 
     if (fRatio > 1.0f) {
-        fMinX = 4.0f * fRatio;
+        fMinX = 5.0f * fRatio;
         fMaxX = fMinX + 0.5f;
     }
     else if (fRatio < 1.0f) {
         fMinY = -4.0f / fRatio;
         fMaxY =  4.0f / fRatio;
+    }
+
+    float boxWidth = getBoundingWidth(size);
+    if (fRatio < 1.0f) {
+        boxWidth *= fRatio;
     }
 
     // search for the labels
@@ -139,7 +143,7 @@ void SoFCColorGradient::setViewportSize(const SbVec2s& size)
             if (labels->getChild(j)->getTypeId() == SoTransform::getClassTypeId()) {
                 if (first) {
                     first = false;
-                    static_cast<SoTransform*>(labels->getChild(j))->translation.setValue(fMaxX + 0.1f, fMaxY - 0.05f + fStep, 0.0f);
+                    static_cast<SoTransform*>(labels->getChild(j))->translation.setValue(fMaxX + 0.1f - boxWidth, fMaxY - 0.05f + fStep, 0.0f);
                 }
                 else {
                     static_cast<SoTransform*>(labels->getChild(j))->translation.setValue(0, -fStep, 0.0f);
@@ -148,7 +152,7 @@ void SoFCColorGradient::setViewportSize(const SbVec2s& size)
         }
     }
 
-    _bbox.setBounds(fMinX, fMinY, fMaxX, fMaxY);
+    _bbox.setBounds(fMinX - boxWidth, fMinY, fMaxX - boxWidth, fMaxY);
     modifyPoints(_bbox);
 }
 
--- a/src/Gui/SoFCColorBar.h
+++ b/src/Gui/SoFCColorBar.h
@@ -112,6 +112,12 @@ public:
   virtual const char* getColorBarName() const = 0;
 
 protected:
+  /** Returns the width of the color bar and labels
+   *
+   * Computes the occupied width of the color bar and its labels.
+   * It therefore determines the bounding box.
+   */
+  float getBoundingWidth(const SbVec2s& size);
   /**
    * Sets the current viewer size to recalculate the new position.
    *
--- a/src/Gui/SoFCColorLegend.cpp
+++ b/src/Gui/SoFCColorLegend.cpp
@@ -190,13 +190,12 @@ void SoFCColorLegend::setMarkerValue(const SoMFString& value)
 
 void SoFCColorLegend::setViewportSize(const SbVec2s& size)
 {
-    // don't know why the parameter range isn't between [-1,+1]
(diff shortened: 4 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6927">pipeline status</a> for feature branch [PR_6927](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6927). Pipeline [550341794](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/550341794) was triggered at [4a4e0db](https://github.com/FreeCAD/FreeCAD/pull/6927/commits/4a4e0db01ce7316ca53b8f8e60447b23668be66c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

FajnaKafka:
Hi!
Maybe it's time to think about changing the logo? A bit different, but similar :) What do you think about it?

( https://drive.google.com/file/d/1deCJLBUHLkCV6Uq9ZA8oCn8OUSabQfXa/view?usp=sharing )

FC logo

luzpaz:
@FajnaKafka Please do not post off-topic content in the bugtracker! There is a [forum thread](https://forum.freecadweb.org/viewtopic.php?f=34&t=28941) dedicated for that. Also changing logos/branding is not something a project should do willy-nilly.

src/Gui/SoFCColorBar.cpp
```diff
@@ -78,6 +80,32 @@ void SoFCColorBarBase::GLRenderBelowPath(SoGLRenderAction *  action)
     SoSeparator::GLRenderBelowPath(action);
 }
 
+float SoFCColorBarBase::getBoundingWidth(const SbVec2s& size)
+{
+    // These are the same camera settings for front nodes as defined in the 3d view
+    SoOrthographicCamera* cam = new SoOrthographicCamera;
+    cam->position = SbVec3f(0, 0, 5);
```
donovaly, line 87: At first many thanks! Tested in a 15 in Hi-DPI laptop and a big screen (normal DPI) and it works fine.

However, I would like to get the numbers documented.  For example why is Z 5 here?

src/Gui/SoFCColorBar.cpp
```diff
@@ -78,6 +80,32 @@ void SoFCColorBarBase::GLRenderBelowPath(SoGLRenderAction *  action)
...
+    cam->height = 10;
```
donovaly, line 88: Same here, why do we need height 10 or would do any height > 0 do the job?

src/Gui/SoFCColorBar.cpp
```diff
@@ -78,6 +80,32 @@ void SoFCColorBarBase::GLRenderBelowPath(SoGLRenderAction *  action)
...
+    cam->nearDistance = 0;
+    cam->farDistance = 10;
```
donovaly, line 90: same here

src/Gui/SoFCColorBar.cpp
```diff
@@ -78,6 +80,32 @@ void SoFCColorBarBase::GLRenderBelowPath(SoGLRenderAction *  action)
...
+    SbViewportRegion vpr(size);
+    SoGetBoundingBoxAction bbact(vpr);
+    bbact.apply(group);
+    SbBox3f box = bbact.getBoundingBox();
+    SbVec3f minPt, maxPt;
+    box.getBounds(minPt, maxPt);
+    group->unref();
+
+    float boxWidth = maxPt[0] - minPt[0];
+    return boxWidth + 0.2f;
```
donovaly, line 106: why the 0.2f extra? I thought the routine should return the actual width and the final adjustment is later done in `SoFCColorGradient::setViewportSize`

src/Gui/SoFCColorGradient.cpp
```diff
@@ -110,20 +110,24 @@ void SoFCColorGradient::setMarkerLabel(const SoMFString& label)
 
 void SoFCColorGradient::setViewportSize(const SbVec2s& size)
 {
-    // don't know why the parameter range isn't between [-1,+1]
-    float fRatio = ((float)size[0]) / ((float)size[1]);
+    float fRatio = static_cast<float>(size[0]) / static_cast<float>(size[1]);
     float fMinX =  4.0f, fMaxX = 4.5f;
```
donovaly, line 114: Independent of the PR, I  did not understand where the 4.5f comes from It seems it was found by try and error or do you know the reason?

src/Gui/SoFCColorGradient.cpp
```diff
@@ -110,20 +110,24 @@ void SoFCColorGradient::setMarkerLabel(const SoMFString& label)
...
 {
-    // don't know why the parameter range isn't between [-1,+1]
-    float fRatio = ((float)size[0]) / ((float)size[1]);
+    float fRatio = static_cast<float>(size[0]) / static_cast<float>(size[1]);
     float fMinX =  4.0f, fMaxX = 4.5f;
     float fMinY = -4.0f, fMaxY = 4.0f;
 
     if (fRatio > 1.0f) {
-        fMinX = 4.0f * fRatio;
+        fMinX = 5.0f * fRatio;
```
donovaly, line 118: same here

donovaly (commented):
Great, many thanks!
I have only questions to understand what the code does.
The while testing I encountered one very special window positions this:
FreeCAD_kc1cLK8iZ0

(one has to play a while to reproduce this and a minimal change of the windows size delivers a correct result, so I think we can neglect it as artifact.)

src/Gui/SoFCColorBar.cpp
donovaly: When I omit the 0.2f extra it looks better here since there is then less whitespace until the right toolbar begins. I will try out tomorrow on all PCs I can access and report back.

src/Gui/SoFCColorGradient.cpp
donovaly: OK, I got it now, the bar has 0.5f width. I will document this.

donovaly:
As reported here: there was a jump when the ration went from < 1 to >1. With these changes there is no jump and I documented the code to make it easier to understand:

```
@@ -110,17 +110,18 @@ void SoFCColorGradient::setMarkerLabel(const SoMFString& label)
 
 void SoFCColorGradient::setViewportSize(const SbVec2s& size)
 {
+    // ratio of window height / width
     float fRatio = static_cast<float>(size[0]) / static_cast<float>(size[1]);
-    float fMinX =  4.0f, fMaxX = 4.5f;
-    float fMinY = -4.0f, fMaxY = 4.0f;
+    float baseYValue = 4.0f;
+    float barWidth = 0.5f;
+    float fMinX = 5.0f * fRatio; // must be scaled with the ratio to assure it stays at the right
+    float fMaxX = fMinX + barWidth;
+    float fMinY = -baseYValue, fMaxY = baseYValue; // bar has the height of almost whole window height
 
-    if (fRatio > 1.0f) {
-        fMinX = 5.0f * fRatio;
-        fMaxX = fMinX + 0.5f;
-    }
-    else if (fRatio < 1.0f) {
-        fMinY = -4.0f / fRatio;
-        fMaxY =  4.0f / fRatio;
+    if (fRatio < 1.0f) {
+        // height must be adjusted to assure bar stays smaller than window height
+        fMinY = -baseYValue / fRatio;
+        fMaxY = baseYValue / fRatio;
     }
 
     float boxWidth = getBoundingWidth(size);
@@ -143,6 +144,7 @@ void SoFCColorGradient::setViewportSize(const SbVec2s& size)
             if (labels->getChild(j)->getTypeId() == SoTransform::getClassTypeId()) {
                 if (first) {
                     first = false;
+                    // set the labels with a small space 0f 0.1f besides the bar
                     static_cast<SoTransform*>(labels->getChild(j))->translation.setValue(fMaxX + 0.1f - boxWidth, fMaxY - 0.05f + fStep, 0.0f);
                 }
                 else {
@@ -152,6 +154,7 @@ void SoFCColorGradient::setViewportSize(const SbVec2s& size)
         }
     }
 
+    // the bar is shifted to the left by the width of the labels to assure tha labels are in every case visible
     _bbox.setBounds(fMinX - boxWidth, fMinY, fMaxX - boxWidth, fMaxY);
     modifyPoints(_bbox);
 }
```

With this change I get a smooth display even in extreme window size ratios:
FreeCAD_xMhUkzJYSb

donovaly:
It turned out I need a proper test build (installer) to be able to test on the different machines tomorrow quickly . Therefore merging this right now.

Nevertheless it would be nice to have a comment about the camera settings to understand these numbers.

FajnaKafka:
Forgive me, please. 🙏🏻
It was a bad idea. I won't do that anymore. You are doing a really good job.



____________________________
FAJNA KAFKA
kubki z fajnymi nadrukami
... i nie tylko.
°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°°
wysłano z 791 447 004

niedz., 29 maj 2022, 23:00 użytkownik luzpaz ***@***.***>
napisał:

src/Gui/SoFCColorBar.cpp
wwmayer: It's the same camera configuration as here: https://github.com/FreeCAD/FreeCAD/blob/master/src/Gui/View3DInventorViewer.cpp#L386

The Z value of 5 is irrelevant and any other value > 0.0 probably works, too. The more important value is `cam->height = 10;` because this means that the normalized coordinates are in the range [-5, +5] x [-5*ratio, +5*ratio] if ratio > 1 and [-5*ratio, +5*ratio] x [-5, +5] if ratio < 1. This explains the numbers like 4.0, 4.5 or 5.0 in the code.

src/Gui/SoFCColorBar.cpp
wwmayer: See above. For `height` no arbitrary value should be used because the hard-coded values for the color bar must be relative to the height value. Since in the viewer it's set to 10 we must use 10 here, too.

src/Gui/SoFCColorBar.cpp
wwmayer: The near and far distances are not that important either because the geometry nodes that are inside the foreground nodes of the viewer usually use z=0.

src/Gui/SoFCColorBar.cpp
wwmayer: > why the 0.2f extra? I thought the routine should return the actual width and the final adjustment is later done in SoFCColorGradient::setViewportSize

This is a small offset to avoid that the text is too close to the right border. It doesn't need to be 0.2 it can also be 0.1 or 0.05 or so.

src/Gui/SoFCColorGradient.cpp
wwmayer: The explanation above. The y values of the normalized coordinates are in the range [-5, 5] because of the `height = 10`. Because the ratio > 1 the x values are in the range [-5*ratio, +5*ratio]. In the past 4.0 was used that worked quite well for all of my monitors and until now there were never any complaints :)

src/Gui/SoFCColorBar.cpp
donovaly: Many thanks. Now I understand the code better. I documented this now: https://github.com/FreeCAD/FreeCAD/commit/537acfdddd

Issue #6969, opened 2022-06-05
[Bug] [Sketcher] Deleting arc of parabola doesn't delete internal geometry


AjinkyaDahale:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=19&t=69200

### Version

0.20 (Development)

### Full version info

```shell
OS: Manjaro Linux (GNOME/gnome)
Word size of FreeCAD: 64-bit
Version: 0.20.29035 +3 (Git)
Build type: Debug
Branch: sketcher-split-join-bsplines
Hash: 76d2847f5ce962f6a212c134fba0694d08a58e8d
Python 3.10.4, Qt 5.15.4, Coin 4.0.1, OCC 7.5.3
Locale: English/United States (en_US)
Installed mods: 
  * Curves 0.4.4
  * ProDarkThemePreferencePack 1.0.0
```

### Issue description

When deleting the geometry, internal geometries of (arcs of) ellipses, circles, and hyperbolas are also deleted. The case is different for Parabolas.

### Anything else?

In `src/Mod/Sketcher/App/Constraint.h:66`, there is a lot of internal geometry info for ellipses and hyperbolas, but for parabola only internal geometry is focus.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

AjinkyaDahale:
@abdullahtahiriyo is there a reason to keep the middle construction segment?

AjinkyaDahale:
This should be fixed with the #6971, which is merged.

PR #6494, opened 2022-03-02
Sketcher: Split B-Splines

AjinkyaDahale:
With these changes the split tool in sketcher also works on b-splines.

Some changes also had to be made in `Part::GeomBSplineCurve::Trim()` to split periodic b-splines.

Issue #5835, opened 2022-02-07
Support for trimming arcs of hyperbola and arcs of parabola


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=3684

* **Reporter:** abdullah
* **Date submitted:** 11/12/2018
* **FreeCAD version:** 0.18
* **Category:** Feature
* **Status:** assigned
* **Tags:** 

# Original report text

It is not supported yet, so support it.

# Other bug information

* **Priority:** normal
* **Severity:** minor
* **Category:** Feature
* **Updated:** 2/6/2021

Roy-043:
This has been implemented in V0.21 AFAICT. OK to close this one?

luzpaz:
@AjinkyaDahale ping (since abdullah is AFK)

AjinkyaDahale:
@luzpaz I had mentioned in #6971 that this is fixed. There is a difference between the splitting in that PR and the "trimming" that this issue requests, but as far as I know trimming was already done in at least by the time 0.20 was released.

PR #6507, opened 2022-03-04
Sketcher: Join curves

AjinkyaDahale:
Rudimentary tool to join two b-splines into a single curve. Theoretically it is possible to join other curves as well after converting them to NURBS.

What is not complete:
* Thorough testing: I was trying to get this in time for the 0.20 feature freeze so this ended up being a last minute rush.
* ~No icons: I didn't design an icon for this tool. I made a temp with just a red "J" but I can't get it to appear on the toolbar. All that's visible is a placeholder.~ @bitacovir provided the icons and I figured out I was using the wrong name in command's pixmap field.

UI forum post: https://forum.freecadweb.org/viewtopic.php?f=34&t=66785

PR #6971, opened 2022-06-05
Sketcher: Split more curves

AjinkyaDahale:
Supercedes PR #6494. I expect to make some major changes to the method for a bit more code-reuse, ~so it's best to keep that PR as is~. 

UPDATE [2 July 2022]: After the recent removal of `DrawSketchHandlers` from `CommandCreateGeo.cpp` into their own classes, I chose to drop #6494 and only rebase this one.

Solves #5835.

Known issues:
1. Splitting parabolas makes constraints malformed. See issue #6969.

### DIFF
--- a/src/Mod/Part/App/Geometry.cpp
+++ b/src/Mod/Part/App/Geometry.cpp
@@ -1602,19 +1602,9 @@ void GeomBSplineCurve::Trim(double u, double v)
     };
 
     try {
-        if(!isPeriodic()) {
-            splitUnwrappedBSpline(u, v);
-        }
-        else { // periodic
-            if( v < u ) { // wraps over origin
-                v = v + 1.0; // v needs one extra lap (1.0)
-
-                splitUnwrappedBSpline(u, v);
-            }
-            else {
-                splitUnwrappedBSpline(u, v);
-            }
-        }
+        if (isPeriodic() && (v - u  <= Precision::PConfusion()))
+            v = v + myCurve->LastParameter() - myCurve->FirstParameter(); // v needs one extra lap
+        splitUnwrappedBSpline(u, v);
     }
     catch (Standard_Failure& e) {
         THROWM(Base::CADKernelError,e.GetMessageString())
--- a/src/Mod/Part/App/Geometry.cpp
+++ b/src/Mod/Part/App/Geometry.cpp
@@ -1602,7 +1602,7 @@ void GeomBSplineCurve::Trim(double u, double v)
     };
 
     try {
-        if (isPeriodic() && (v - u  <= Precision::PConfusion()))
+        if (isPeriodic() && (v < u))
             v = v + myCurve->LastParameter() - myCurve->FirstParameter(); // v needs one extra lap
         splitUnwrappedBSpline(u, v);
     }
--- a/src/Mod/Sketcher/App/SketchObject.cpp
+++ b/src/Mod/Sketcher/App/SketchObject.cpp
@@ -2975,286 +2975,224 @@ int SketchObject::split(int GeoId, const Base::Vector3d &point)
     std::vector<Constraint *> newConstraints;
 
     Base::Vector3d startPoint, endPoint, splitPoint;
-    double radius, startAngle, endAngle, splitAngle=0.0;
     double startParam, endParam, splitParam=0.0;
     unsigned int longestPart = 0;
 
+    auto createGeosFromPeriodic =
+        [&](const Part::GeomCurve* curve,
+            auto getCurveWithLimitParams,
+            auto createAndTransferConstraints) {
+            // find split point
+            curve->closestParameter(point, splitParam);
+            double period = curve->getLastParameter() - curve->getFirstParameter();
+            startParam = splitParam;
+            endParam = splitParam + period;
+
+            // create new arc and restrict it
+            auto newCurve = getCurveWithLimitParams(curve, startParam, endParam);
+            int newId(GeoEnum::GeoUndef);
+            newGeometries.push_back(newCurve);
+            newId = addGeometry(newCurve);
+            if (newId >= 0) {
+                newIds.push_back(newId);
+                setConstruction(newId, GeometryFacade::getConstruction(curve));
+                exposeInternalGeometry(newId);
+
+                // transfer any constraints
+                createAndTransferConstraints(GeoId, newId);
+                return true;
+            }
+
+            return false;
+        };
+
+    auto createGeosFromNonPeriodic =
+        [&](const Part::GeomBoundedCurve* curve,
+            auto getCurveWithLimitParams,
+            auto createAndTransferConstraints) {
+            startPoint = curve->getStartPoint();
+            endPoint = curve->getEndPoint();
+
+            // find split point
+            curve->closestParameter(point, splitParam);
+            startParam = curve->getFirstParameter();
+            endParam = curve->getLastParameter();
+            // TODO: Using parameter difference as a poor substitute of length.
+            // Computing length of an arc of a generic conic would be expensive.
+            if (endParam - splitParam > splitParam - startParam) {
+                longestPart = 1;
+            }
+
+            // create new curves
+            auto newCurve = getCurveWithLimitParams(curve, startParam, splitParam);
+            int newId(GeoEnum::GeoUndef);
+            newGeometries.push_back(newCurve);
+            newId = addGeometry(newCurve);
+            if (newId >= 0) {
+                newIds.push_back(newId);
+                setConstruction(newId, GeometryFacade::getConstruction(curve));
+                exposeInternalGeometry(newId);
+
+                // the "second" half
+                newCurve = getCurveWithLimitParams(curve, splitParam, endParam);
+                newGeometries.push_back(newCurve);
+                newId = addGeometry(newCurve);
+                if (newId >= 0) {
+                    newIds.push_back(newId);
+                    setConstruction(newId, GeometryFacade::getConstruction(curve));
+                    exposeInternalGeometry(newId);
+
+                    // TODO: Certain transfers and new constraint can be directly made here.
+                    // But this may reduce readability.
+                    // apply appropriate constraints on the new points at split point and
+                    // transfer constraints from start and end of original spline
+                    createAndTransferConstraints(GeoId, newIds[0], newIds[1]);
+                    return true;
+                }
+            }
+
+            return false;
+        };
 
     bool ok = false;
     if (geo->getTypeId() == Part::GeomLineSegment::getClassTypeId()) {
-        const Part::GeomLineSegment *lineSegm = static_cast<const Part::GeomLineSegment *>(geo);
-
-        startPoint = lineSegm->getStartPoint();
-        endPoint = lineSegm->getEndPoint();
-        splitPoint = point.Perpendicular(startPoint, endPoint - startPoint);
-        if ((endPoint - splitPoint).Length() > (splitPoint - startPoint).Length()) {
-            longestPart = 1;
-        }
-
-        Part::GeomLineSegment *newLine = static_cast<Part::GeomLineSegment *>(lineSegm->copy());
-        newGeometries.push_back(newLine);
-
-        newLine->setPoints(startPoint, splitPoint);
-        int newId = addGeometry(newLine);
-        if (newId >= 0) {
-            newIds.push_back(newId);
-            setConstruction(newId, GeometryFacade::getConstruction(geo));
-
-            newLine = static_cast<Part::GeomLineSegment*>(lineSegm->copy());
-            newGeometries.push_back(newLine);
-
-            newLine->setPoints(splitPoint, endPoint);
-            newId = addGeometry(newLine);
-            if (newId >= 0) {
-                newIds.push_back(newId);
-                setConstruction(newId, GeometryFacade::getConstruction(geo));
-
+        ok = createGeosFromNonPeriodic(
+            static_cast<const Part::GeomBoundedCurve *>(geo),
+            [](const Part::GeomCurve* curve, double startParam, double endParam) {
+                auto newArc = static_cast<Part::GeomLineSegment *>(curve->clone());
+                newArc->setRange(startParam, endParam);
+                return newArc;
+            },
+            [this, &newConstraints](int GeoId, int newId0, int newId1) {
                 Constraint* joint = new Constraint();
                 joint->Type = Coincident;
-                joint->First = newIds[0];
+                joint->First = newId0;
                 joint->FirstPos = PointPos::end;
-                joint->Second = newIds[1];
+                joint->Second = newId1;
                 joint->SecondPos = PointPos::start;
                 newConstraints.push_back(joint);
 
-                transferConstraints(GeoId, PointPos::start, newIds[0], PointPos::start, true);
-                transferConstraints(GeoId, PointPos::end, newIds[1], PointPos::end, true);
-                ok = true;
-            }
-        }
+                transferConstraints(GeoId, PointPos::start, newId0, PointPos::start);
+                transferConstraints(GeoId, PointPos::end, newId1, PointPos::end);
+            });
     }
     else if (geo->getTypeId() == Part::GeomCircle::getClassTypeId()) {
-        const Part::GeomCircle *circle = static_cast<const Part::GeomCircle *>(geo);
-
-        Base::Vector3d center(circle->getLocation());
-        Base::Vector3d dir(point - center);
-        radius = circle->getRadius();
-
-        splitAngle = atan2(dir.y, dir.x);
-        startAngle = splitAngle;
-        endAngle = splitAngle + M_PI*2.0;
-
-        splitPoint = Base::Vector3d(center.x + radius*cos(splitAngle), center.y + radius*sin(splitAngle));
-        startPoint = splitPoint;
-        endPoint = splitPoint;
-
-        Part::GeomArcOfCircle *arc = new Part::GeomArcOfCircle();
-        newGeometries.push_back(arc);
-
-        arc->setLocation(center);
-        arc->setRadius(radius);
-        arc->setRange(startAngle, endAngle, false);
-        int arcId = addGeometry(arc);
-        if (arcId >= 0) {
-            newIds.push_back(arcId);
-            setConstruction(arcId, GeometryFacade::getConstruction(geo));
-
-            transferConstraints(GeoId, PointPos::mid, arcId, PointPos::mid);
-            ok = true;
-        }
+        ok = createGeosFromPeriodic(
+            static_cast<const Part::GeomCurve *>(geo),
+            [](const Part::GeomCurve* curve, double startParam, double endParam) {
+                auto newArc = new Part::GeomArcOfCircle(Handle(Geom_Circle)::DownCast(curve->handle()->Copy()));
+                newArc->setRange(startParam, endParam, false);
+                return newArc;
+            },
+            [this](int GeoId, int newId) {
+                transferConstraints(GeoId, PointPos::mid, newId, PointPos::mid);
+            });
     }
     else if (geo->getTypeId() == Part::GeomEllipse::getClassTypeId()) {
-        const Part::GeomEllipse *curve = static_cast<const Part::GeomEllipse *>(geo);
-
-        // find split point
-        curve->closestParameter(point, splitParam);
-        double period = curve->getLastParameter() - curve->getFirstParameter();
-        startParam = splitParam;
-        endParam = splitParam + period;
-
-        // create new arc
-        auto newArc = new Part::GeomArcOfEllipse(Handle(Geom_Ellipse)::DownCast(curve->handle()->Copy()));
-        newArc->setRange(startParam, endParam, false);
-        int newId = addGeometry(newArc);
-        if (newId >= 0) {
-            newIds.push_back(newId);
-            setConstruction(newId, GeometryFacade::getConstruction(geo));
-            exposeInternalGeometry(newId);
-
-            transferConstraints(GeoId, PointPos::mid, newId, PointPos::mid);
-            ok = true;
-        }
+        ok = createGeosFromPeriodic(
+            static_cast<const Part::GeomCurve *>(geo),
+            [](const Part::GeomCurve* curve, double startParam, double endParam) {
+                auto newArc = new Part::GeomArcOfEllipse(Handle(Geom_Ellipse)::DownCast(curve->handle()->Copy()));
+                newArc->setRange(startParam, endParam, false);
+                return newArc;
+            },
+            [this](int GeoId, int newId) {
+                transferConstraints(GeoId, PointPos::mid, newId, PointPos::mid);
+            });
     }
     else if (geo->getTypeId() == Part::GeomArcOfCircle::getClassTypeId()) {
-        const Part::GeomArcOfCircle *arc = static_cast<const Part::GeomArcOfCircle *>(geo);
-
-        startPoint = arc->getStartPoint();
-        endPoint = arc->getEndPoint();
-
-        Base::Vector3d center(arc->getLocation());
-        radius = arc->getRadius();
-        arc->getRange(startAngle, endAngle, false);
-
-        Base::Vector3d dir(point - center);
-        splitAngle = atan2(dir.y, dir.x);
-        if (splitAngle < startAngle) {
-            splitAngle += M_PI*2.0;
-        }
-        if (endAngle - splitAngle > splitAngle - startAngle) {
-            longestPart = 1;
-        }
-
-        splitPoint = Base::Vector3d(center.x + radius*cos(splitAngle), center.y + radius*sin(splitAngle));
-        startPoint = splitPoint;
-        endPoint = splitPoint;
-
-        Part::GeomArcOfCircle *newArc = static_cast<Part::GeomArcOfCircle *>(arc->copy());
-        newGeometries.push_back(newArc);
-
-        newArc->setRange(startAngle, splitAngle, false);
-        int newId = addGeometry(newArc);
-        if (newId >= 0) {
-            newIds.push_back(newId);
-            setConstruction(newId, GeometryFacade::getConstruction(geo));
-
-            newArc = static_cast<Part::GeomArcOfCircle*>(arc->copy());
-            newGeometries.push_back(newArc);
-
-            newArc->setRange(splitAngle, endAngle, false);
-            newId = addGeometry(newArc);
-            if (newId >= 0) {
-                newIds.push_back(newId);
-                setConstruction(newId, GeometryFacade::getConstruction(geo));
-
+        ok = createGeosFromNonPeriodic(
+            static_cast<const Part::GeomBoundedCurve *>(geo),
+            [](const Part::GeomCurve* curve, double startParam, double endParam) {
+                auto newArc = static_cast<Part::GeomArcOfCircle *>(curve->clone());
+                newArc->setRange(startParam, endParam, false);
+                return newArc;
+            },
+            [this, &newConstraints](int GeoId, int newId0, int newId1) {
                 Constraint* joint = new Constraint();
                 joint->Type = Coincident;
-                joint->First = newIds[0];
+                joint->First = newId0;
                 joint->FirstPos = PointPos::end;
-                joint->Second = newIds[1];
+                joint->Second = newId1;
                 joint->SecondPos = PointPos::start;
                 newConstraints.push_back(joint);
 
                 joint = new Constraint();
                 joint->Type = Coincident;
-                joint->First = newIds[0];
+                joint->First = newId0;
                 joint->FirstPos = PointPos::mid;
-                joint->Second = newIds[1];
+                joint->Second = newId1;
                 joint->SecondPos = PointPos::mid;
                 newConstraints.push_back(joint);
 
-                transferConstraints(GeoId, PointPos::start, newIds[0], PointPos::start, true);
-                transferConstraints(GeoId, PointPos::mid, newIds[0], PointPos::mid);
-                transferConstraints(GeoId, PointPos::end, newIds[1], PointPos::end, true);
-                ok = true;
-            }
-        }
+                transferConstraints(GeoId, PointPos::start, newId0, PointPos::start, true);
+                transferConstraints(GeoId, PointPos::mid, newId0, PointPos::mid);
+                transferConstraints(GeoId, PointPos::end, newId1, PointPos::end, true);
+            });
     }
     else if (geo->isDerivedFrom(Part::GeomArcOfConic::getClassTypeId())) {
-        const Part::GeomArcOfConic *arc = static_cast<const Part::GeomArcOfConic *>(geo);
-
-        startPoint = arc->getStartPoint();
-        endPoint = arc->getEndPoint();
-
-        // find split point
-        arc->closestParameter(point, splitParam);
-        startParam = arc->getFirstParameter();
-        endParam = arc->getLastParameter();
-        // TODO: Using parameter difference as a poor substitute of length.
-        // Computing length of an arc of a generic conic would be expensive.
-        if (endParam - splitParam > splitParam - startParam) {
-            longestPart = 1;
-        }
-
-        // create new arcs
-        auto newArc = static_cast<Part::GeomArcOfConic *>(arc->clone());
-        int newId(GeoEnum::GeoUndef);
-        newGeometries.push_back(newArc);
-        newArc->setRange(startParam, splitParam, /*emulateCCW=*/true);
-        newId = addGeometry(newArc);
-        if (newId >= 0) {
-            newIds.push_back(newId);
-            setConstruction(newId, GeometryFacade::getConstruction(geo));
-            exposeInternalGeometry(newId);
-
-            // the "second" half
-            auto newArc2 = static_cast<Part::GeomArcOfConic *>(arc->clone());
-            int newId2(GeoEnum::GeoUndef);
-            newArc2->setRange(splitParam, endParam, /*emulateCCW=*/true);
-            newId2 = addGeometry(newArc2);
-            if (newId2 >= 0) {
-                newIds.push_back(newId2);
-                setConstruction(newId2, GeometryFacade::getConstruction(geo));
-                exposeInternalGeometry(newId2);
-
+        ok = createGeosFromNonPeriodic(
+            static_cast<const Part::GeomBoundedCurve *>(geo),
+            [](const Part::GeomCurve* curve, double startParam, double endParam) {
+                auto newArc = static_cast<Part::GeomArcOfConic *>(curve->clone());
+                newArc->setRange(startParam, endParam);
+                return newArc;
+            },
+            [this, &newConstraints](int GeoId, int newId0, int newId1) {
                 // apply appropriate constraints on the new points at split point
                 Constraint* joint = new Constraint();
                 joint->Type = Coincident;
-                joint->First = newIds[0];
+                joint->First = newId0;
                 joint->FirstPos = PointPos::end;
-                joint->Second = newIds[1];
+                joint->Second = newId1;
                 joint->SecondPos = PointPos::start;
                 newConstraints.push_back(joint);
 
                 // TODO: Do we apply constraints on center etc of the conics?
 
-                // transfer constraints from start and end of original spline
-                transferConstraints(GeoId, PointPos::start, newIds[0], PointPos::start, true);
-                transferConstraints(GeoId, PointPos::end, newIds[1], PointPos::end, true);
-                ok = true;
-            }
-        }
+                // transfer constraints from start and end of original
+                transferConstraints(GeoId, PointPos::start, newId0, PointPos::start, true);
+                transferConstraints(GeoId, PointPos::end, newId1, PointPos::end, true);
+            });
     }
     else if (geo->getTypeId() == Part::GeomBSplineCurve::getClassTypeId()) {
         const Part::GeomBSplineCurve *bsp = static_cast<const Part::GeomBSplineCurve *>(geo);
 
-        // find split point
-        bsp->closestParameter(point, splitParam);
-
         // what to do for periodic b-splines?
         if (bsp->isPeriodic()) {
-            Part::GeomBSplineCurve* newBsp;
-            int newId(GeoEnum::GeoUndef);
-            newBsp = static_cast<Part::GeomBSplineCurve *>(bsp->clone());
-            newGeometries.push_back(newBsp);
-            newBsp->Trim(splitParam, splitParam);
-            newId = addGeometry(newBsp);
-            if (newId >= 0) {
-                newIds.push_back(newId);
-                setConstruction(newId, GeometryFacade::getConstruction(geo));
-                exposeInternalGeometry(newId);
-
-                // no constraints to transfer here, and we assume the split is to "break" the b-spline
-                ok = true;
-            }
+            ok = createGeosFromPeriodic(
+                static_cast<const Part::GeomCurve *>(geo),
+                [](const Part::GeomCurve* curve, double startParam, double endParam) {
+                    auto newBsp = static_cast<Part::GeomBSplineCurve *>(curve->clone());
+                    newBsp->Trim(startParam, endParam);
+                    return newBsp;
+                },
+                [](int, int) {
+                    // no constraints to transfer here, and we assume the split is to "break" the b-spline
+                });
         }
         else {
-            // create new b-splines
-            Part::GeomBSplineCurve* newBsp;
-            int newId(GeoEnum::GeoUndef);
-            newBsp = static_cast<Part::GeomBSplineCurve *>(bsp->clone());
-            newGeometries.push_back(newBsp);
-            newBsp->Trim(newBsp->getFirstParameter(), splitParam);
-            newId = addGeometry(newBsp);
-            if (newId >= 0) {
-                newIds.push_back(newId);
-                setConstruction(newId, GeometryFacade::getConstruction(geo));
-                exposeInternalGeometry(newId);
-
-                // the "second" half
-                newBsp = static_cast<Part::GeomBSplineCurve *>(bsp->clone());
-                newGeometries.push_back(newBsp);
-                newBsp->Trim(splitParam, newBsp->getLastParameter());
-                newId = addGeometry(newBsp);
-                if (newId >= 0) {
-                    newIds.push_back(newId);
-                    setConstruction(newId, GeometryFacade::getConstruction(geo));
-                    exposeInternalGeometry(newId);
-
+            ok = createGeosFromNonPeriodic(
+                static_cast<const Part::GeomBoundedCurve *>(geo),
+                [](const Part::GeomCurve* curve, double startParam, double endParam) {
+                    auto newBsp = static_cast<Part::GeomBSplineCurve *>(curve->clone());
+                    newBsp->Trim(startParam, endParam);
+                    return newBsp;
+                },
+                [this, &newConstraints](int GeoId, int newId0, int newId1) {
                     // apply appropriate constraints on the new points at split point
                     Constraint* joint = new Constraint();
                     joint->Type = Coincident;
-                    joint->First = newIds[0];
+                    joint->First = newId0;
                     joint->FirstPos = PointPos::end;
-                    joint->Second = newIds[1];
+                    joint->Second = newId1;
                     joint->SecondPos = PointPos::start;
                     newConstraints.push_back(joint);
 
                     // transfer constraints from start and end of original spline
-                    transferConstraints(GeoId, PointPos::start, newIds[0], PointPos::start, true);
-                    transferConstraints(GeoId, PointPos::end, newIds[1], PointPos::end, true);
-                    ok = true;
-                }
-            }
+                    transferConstraints(GeoId, PointPos::start, newId0, PointPos::start, true);
+                    transferConstraints(GeoId, PointPos::end, newId1, PointPos::end, true);
+                });
         }
     }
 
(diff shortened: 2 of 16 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [555971790](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/555971790) was triggered at [4e764bd](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/4e764bd9bfef7612bce76d8bde71e70533e56401). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [556042026](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/556042026) was triggered at [e033588](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/e033588e9ffe7775749fed5e15221b1fed03a91c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
This needs another rebase

AjinkyaDahale:
> This needs another rebase

Are you planning to merge it right now? If yes, @abdullahtahiriyo did mention he needed to confirm the changes didn't have adverse effects on DrawSketchHandlers.

Apart from that it's best to only rebase if we're about to merger or there are some conflicts that need to be manually resolved.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [578984148](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/578984148) was triggered at [75a2876](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/75a2876435f536654d617b26eb418aebd8b8979b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [579735890](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/579735890) was triggered at [a556329](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/a5563299300afff7574cf15848d3a56b51986689). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [581004675](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/581004675) was triggered at [91e3c5d](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/91e3c5de458680faf868dc980d43616ffb6d4e9f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [581729925](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/581729925) was triggered at [91e249b](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/91e249b7cfdde8ef75886b2150eef26db58f8748). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [583968089](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/583968089) was triggered at [b92b07a](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/b92b07adce8c7c2c6298274a6f9296bfaac7185b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [584002800](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/584002800) was triggered at [cf168b7](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/cf168b7008ef226d6e7386d4f57240008f58e623). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [584005584](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/584005584) was triggered at [b678800](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/b67880033f7962f2cd10737c9ae88f2e4e6034ec). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/Geometry.cpp
```diff
@@ -1596,19 +1596,9 @@ void GeomBSplineCurve::Trim(double u, double v)
...
-
-                splitUnwrappedBSpline(u, v);
-            }
-            else {
-                splitUnwrappedBSpline(u, v);
-            }
-        }
+        if (isPeriodic() && (v - u  <= Precision::PConfusion()))
+            v = v + myCurve->LastParameter() - myCurve->FirstParameter(); // v needs one extra lap
+        splitUnwrappedBSpline(u, v);
```
abdullahtahiriyo, line 1601: I am not sure I understand correctly the motivation for the "<= Precision::PConfusion()".

A BSpline is wrapped if the v parameter is less than the u parameter. A BSpline where v is a positive amount (Precision::PConfusion()) greater than u is not wrapped. In such a case v is greater than u. Am I missing anything?

Then, to unwrap it, without changing it, v needs a full lap (1.0). However, it appears to me that LastParameter()-FirstParameter() will only be one lap if the Bpsline has not been split before (if it is a full "turn"). I have not checked, maybe the split of a periodic b-spline is no longer a periodic b-spline (so the code is not executed). Even if that is the case, why not use 1.0 addition instead, which clearly conveys the intention? Maybe I am missing something?

src/Mod/Sketcher/Gui/DrawSketchHandlerSplitting.h
```diff
@@ -90,7 +93,10 @@ class DrawSketchHandlerSplitting: public DrawSketchHandler
             const Part::Geometry *geom = sketchgui->getSketchObject()->getGeometry(GeoId);
             if (geom->getTypeId() == Part::GeomLineSegment::getClassTypeId()
                 || geom->getTypeId() == Part::GeomCircle::getClassTypeId()
-                || geom->getTypeId() == Part::GeomArcOfCircle::getClassTypeId()) {
+                || geom->getTypeId() == Part::GeomEllipse::getClassTypeId()
+                || geom->getTypeId() == Part::GeomArcOfCircle::getClassTypeId()
```
abdullahtahiriyo, line 97: Isn't GeomArcOfCircle a redundant checking in view of isDerivedFrom(GeomArcOfConic)?

src/Mod/Sketcher/Gui/DrawSketchHandlerSplitting.h
```diff
@@ -54,7 +54,10 @@ class SplittingSelection : public Gui::SelectionFilterGate
             const Part::Geometry *geom = Sketch->getGeometry(GeoId);
             if (geom->getTypeId() == Part::GeomLineSegment::getClassTypeId()
                 || geom->getTypeId() == Part::GeomCircle::getClassTypeId()
-                || geom->getTypeId() == Part::GeomArcOfCircle::getClassTypeId()) {
+                || geom->getTypeId() == Part::GeomEllipse::getClassTypeId()
+                || geom->getTypeId() == Part::GeomArcOfCircle::getClassTypeId()
+                || geom->isDerivedFrom(Part::GeomArcOfConic::getClassTypeId())
```
abdullahtahiriyo, line 59: idem as below.

src/Mod/Part/App/Geometry.cpp
AjinkyaDahale: > I am not sure I understand correctly the motivation for the "<= Precision::PConfusion()".
> 
> A BSpline is wrapped if the v parameter is less than the u parameter. A BSpline where v is a positive amount (Precision::PConfusion()) greater than u is not wrapped. In such a case v is greater than u. Am I missing anything?

`PConfusion` refers to a small epsilon because of roundoff errors. Let's consider a periodic spline with period 1.0. If I want to split it at point 0.5, and I've for some reason provided two different values (maybe because these values had to be computed by different clicks, say---which is not our case in split, but this change is in part and general). Then, maybe u=0.5 and v=0.499999997, or v=0.50000000001. In the first case, with just "v<=u" the wrapping will be considered. However, in the latter case, we will get just a small arc from 0.5 to 0.50000000001.

Note that if I had needed to check for strictly "v<u", I would have instead used `(v - u  <= -Precision::PConfusion())`. Also, here I expect that `<=` can be replaced by `<`.

> Then, to unwrap it, without changing it, v needs a full lap (1.0). However, it appears to me that LastParameter()-FirstParameter() will only be one lap if the Bpsline has not been split before (if it is a full "turn"). 
> ...why not use 1.0 addition instead, which clearly conveys the intention? Maybe I am missing something?

The `LastParameter()-FirstParameter()` is for the most general case. I don't expect it would happen organically, but I could, for example, split a nonperiodic spline at 0.6, delete the second part, then join the ends of the first part. We could enforce that every periodic spline in FC should always have a period of 1.0, but I don't see any particular good in that.

> I have not checked, maybe the split of a periodic b-spline is no longer a periodic b-spline (so the code is not executed). 

The code is executed as if for a periodic b-spline. Only after the OCC algorithm is executed does it become non-periodic.

src/Mod/Sketcher/Gui/DrawSketchHandlerSplitting.h
AjinkyaDahale: I was thinking the same. Just procrastinated deciding whether to make a new commit or amend an old one.

src/Mod/Sketcher/Gui/DrawSketchHandlerSplitting.h
AjinkyaDahale: I'll push a commit.

Off-topic: first time I hear this word "idem".

src/Mod/Sketcher/Gui/DrawSketchHandlerSplitting.h
AjinkyaDahale: Changes pushed.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [584105070](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/584105070) was triggered at [1b6aec3](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/1b6aec3df16fc841f6b5f4607801113b2aca7c5f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/Geometry.cpp
abdullahtahiriyo: > The `LastParameter()-FirstParameter()` is for the most general case. I don't expect it would happen organically, but I could, for example, split a nonperiodic spline at 0.6, delete the second part, then join the ends of the first part. We could enforce that every periodic spline in FC should always have a period of 1.0, but I don't see any particular good in that.

I thought OCC would always enforce 1.0 for periodic B-Splines (including the non-periodic to periodic case). If the periodic B-spline changes after OCC upon trimming to non-periodic, then it is acceptable to me to have the subtraction of parameters.

> `PConfusion` refers to a small epsilon because of roundoff errors.

Agreed.

> Let's consider a periodic spline with period 1.0. If I want to split it at point 0.5, and I've for some reason provided two different values (maybe because these values had to be computed by different clicks, say---which is not our case in split, but this change is in part and general). Then, maybe u=0.5 and v=0.499999997, or v=0.50000000001. In the first case, with just "v<=u" the wrapping will be considered. However, in the latter case, we will get just a small arc from 0.5 to 0.50000000001.

But, isn't this a case of wrong input to the Part function? 

In a general case, if I am telling this function I want to trim between u=0.5 and v=0.5000000000001, I expect, as a user of this function, that it produces very small arc (for it is what I am asking for). Getting a long arc is not what I am asking for. That would look to me as unexpected behaviour (when looking at the input data I am providing).

I can follow the argument that in your case, rounding playing against your will, you may be feeding the wrong numbers to the function (the ones of the extra short arc). However, the function should not try to guess whether the user may have wanted something else that what he is asking for. In other words, if you expect the large arc, the client code calling the function has to ensure it provides the right numbers (that it corrects wrong input data before calling the function). In your example, being a mouse input, input data validation (at client code) should be responsible for this.

> Note that if I had needed to check for strictly "v<u", I would have instead used `(v - u <= -Precision::PConfusion())`

That would have the contrary effect, that would also be undefined behaviour.

> Also, here I expect that `<=` can be replaced by `<`

When the trimming distance (in term of parameter bounds) is under Confusion, the OCC algorithm will produce an exception. This will be caught and trigger a CADKernelError for the client code to handle. If the client code wants to avoid receiving an exception, it needs to ensure it is providing values that do not produce one. 

I tend to think that if a user is asking for an impossible getting an exception/error is what he should expect. It would be unexpected to get a valid result, when an invalid result is expected.

src/Mod/Sketcher/Gui/DrawSketchHandlerSplitting.h
abdullahtahiriyo: This kind, I generally make a new commit, interactive rebase to move it up after the commit that needs correction, and squash it together with it. Then I force-push the branch. It is the cleanest history line.

src/Mod/Sketcher/Gui/DrawSketchHandlerSplitting.h
abdullahtahiriyo: :)

abdullahtahiriyo:
Note to self: You still need to sit down and give a look to the split SketchObject algorithm.

src/Mod/Part/App/Geometry.cpp
AjinkyaDahale: > > Let's consider a periodic spline with period 1.0. If I want to split it at point 0.5, and I've for some reason provided two different values (maybe because these values had to be computed by different clicks, say---which is not our case in split, but this change is in part and general). Then, maybe u=0.5 and v=0.499999997, or v=0.50000000001. In the first case, with just "v<=u" the wrapping will be considered. However, in the latter case, we will get just a small arc from 0.5 to 0.50000000001.
> 
> But, isn't this a case of wrong input to the Part function?
> 
> In a general case, if I am telling this function I want to trim between u=0.5 and v=0.5000000000001, I expect, as a user of this function, that it produces very small arc (for it is what I am asking for). Getting a long arc is not what I am asking for. That would look to me as unexpected behaviour (when looking at the input data I am providing).

Off the top of my head, I can think of one way this can happen. Say we implement an "arc" tool that takes an arc on the curve between two points. A user quickly clicks around the same point to get the same effect as "splitting" the loop. Because of the error this can go any way. 

A slightly more practical case maybe trying to find union/intersection of tangential faces, though in that case we probably will also need a way to replace segments by just points. I'll skip this example at least for now.
 
> > Note that if I had needed to check for strictly "v<u", I would have instead used `(v - u <= -Precision::PConfusion())`
> 
> That would have the contrary effect, that would also be undefined behaviour.

Note that u and/or v can be negative. We just want to ensure that "v<u" or "v-u<0" _without any doubt_. So we sacrifice that much amount of confusion to ensure that they are definitely not too close to each other. 

> I tend to think that if a user is asking for an impossible getting an exception/error is what he should expect. It would be unexpected to get a valid result, when an invalid result is expected.

I'm fairly certain that the things requested here are valid.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [584315625](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/584315625) was triggered at [6396ed3](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/6396ed38e17c4f40e5060800ccb48cfbf0be1138). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/Geometry.cpp
AjinkyaDahale: > > We just want to ensure that "v<u" or "v-u<0" without any doubt.
> 
> If I do "if( v<u)" and it returns true, I have no doubt at all that v is smaller than u.

You can ensure that the _provided_ v is smaller than the _provided_ u, but the _intended_ values maybe too close to each other.

> 
> We must be having a communication problem here. I will try again with different wording.

It may be difference in opinion as to where the check for confusion should be done. The check for "v < u" is already done here before I changed the file, so I don't expect things to break anywhere else.

>     1. There are many user input cases susceptible of producing values of v and u that the user does not want. This I can agree. The portion of the application getting this input from the user is responsible for validating user input data. It can do as many checks as necessary and condition this data. It must decide what it wants to trim to get the effect it wants to achieve. Then it can use those in this trim function.

I'm not sure if it's a good idea to depend on all "callers" to do this. As such then we can even impose that "u<v" and assume that is true. 

>     2. I cannot understand why a function that offers the functionality to trim a bspline between u and v should "sacrifice" part of its working range and wrap something that does not need to be wrapped. I cannot understand why if I am asking to trim between 0.5-Confusion and 0.5, the function should return a trim between 0.5-Conf and 1.5.

This is unfortunately an eternal problem in computational geometry. Thing is, if we do ask a trim between 0.5-Confusion and 0.5, OCC is not guaranteed to provide the intended result. Pick any of the solved bugs in FC tagged "OCC", and take a look at the commit that fixed it. I am certain that most of the fixes involved incorporating confusion somewhere in their checks.

If someone really needs a higher precision than `Confusion`, this will most likely involve extensive changes in the algorithm and/or performance penalty because of the higher precision checks.

As such, since we are directly passing these parameters to OCC, the tolerance has to be at least as much OCC allows. Using a more sophisticated solver, we may eventually get to numbers like `0.000001*Precision::Confusion` but even then this check will be needed. The price for not making this sacrifice is instability.

Also take a look at https://dev.opencascade.org/doc/occt-7.6.0/refman/html/class_geom_convert.html#a6e7ddc155168bc9b9b2f21d4d5295d42. It clearly stats "Raised if Abs (FromU1 - ToU2) <= ParametricTolerance", i.e. if they are too close.

> BTW, the wrapped parameters are confusion in excess of one "lap"...

Not sure how the behavior would change in this case, but I expect this should be fine so long as the difference past the "lap" is not too much. Again, from the reference "Raised if FromU1 or ToU2 are out of the parametric bounds of the curve (_The tolerance criterion is ParametricTolerance_)", so I expect that there are safeguards against it.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [584395895](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/584395895) was triggered at [6b3d1b8](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/6b3d1b842e3a395ba7fc55ebfabef379758935cb). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [585272109](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/585272109) was triggered at [0340128](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/03401288ee3cb2f40df3707ed977fcdd310067ec). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [586258639](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/586258639) was triggered at [7d49057](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/7d490573455064c8e84708d078c65967ff17b68f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [669090198](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/669090198) was triggered at [c2bf14a](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/c2bf14ac16fe25a02ff4036826b31f35aa0c8bc4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
> Note to self: You still need to sit down and give a look to the split SketchObject algorithm.

@abdullahtahiriyo did you get the time to review this and #6507? These will also eventually be important for better B-spline support.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [691949395](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/691949395) was triggered at [89e42e5](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/89e42e50d92ff3e3b4826fce876757eabd2e760c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [711238972](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/711238972) was triggered at [5b6d0f6](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/5b6d0f6785a9bfe2dbacef6e65a992fe6e501d84). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
I restart now the review of this PR.

From what I see above (and faintly remember), we have a problem of understanding where a check affecting the 10-4 percent of the cases should be done. I will review the exchange and try to see how we can solve this.

@AjinkyaDahale Is there anything specific about this PR that you want me to look at in detail?

AjinkyaDahale:
@abdullahtahiriyo I don't believe I recall anything specific apart from the roundoff error handling.

src/Mod/Sketcher/App/SketchObject.cpp
```diff
@@ -2983,155 +2983,246 @@ int SketchObject::split(int GeoId, const Base::Vector3d &point)
...
+                },
+                [](int, int) {
+                    // no constraints to transfer here, and we assume the split is to "break" the b-spline
+                });
+        }
+        else {
+            ok = createGeosFromNonPeriodic(
+                static_cast<const Part::GeomBoundedCurve *>(geo),
+                [](const Part::GeomCurve* curve, double startParam, double endParam) {
+                    auto newBsp = static_cast<Part::GeomBSplineCurve *>(curve->clone());
```
abdullahtahiriyo, line 3186: Notes to self:
1. Who takes care of the lifetime of this just cloned pointer, specially if Trim exists due to an exception? (this applies to all geometry types, not just this one).
2. It is possible to pass here to trim parameters start=1 and end=1 (see calling code in lambda), client code should probably not allow this.

abdullahtahiriyo: Note to self:

For testing:
App.getDocument('hang').getObject('Sketch').addGeometry(Part.BSplineCurve([App.Vector(-50,60),App.Vector(-60,70),App.Vector(-70,80),App.Vector(-80,90),App.Vector(-90,100)],None,None,False,3,None,False),False)

ActiveSketch.solve()

App.getDocument('hang').getObject('Sketch').split(10,App.Vector(-89.9999999999999999999999999,99.9999999999999999999999999,0))

abdullahtahiriyo (commented):
Hi!

The most prominent issue I have found is the possible memory leaks due to not releasing naked pointers upon exception.

In order to move forward, please consider these extra commits:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/pull/new/sketcher-split-more-curves-fixed

I think I have not changed behaviour, or introduced extra bugs. But feel free to review the changes and come back with anything you do not like.

One aspect that I have seen, but I have decided not to change, is that it is possible to have calls to create new curves with a same start and end parameter (there is some code to reproduce in the notes-to-self). I think sketchobject's split() could do some extra checks to decide what exactly to do (create only a first curve if the second curve would lead to a zero range parameter; or create only the second curve if the first curve would lead to a zero range parameter; of course zero parameter range can be considered in the light of PConfusion, so basically do not create, and correctly handle the non-creation at the level of constraints too, if the range is under PConfusion). 

If you want to change this PR to do something about it, I am ok with that. If you want to have it merged and later handle it as a feature improvement. I am fine with that too.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [719247023](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/719247023) was triggered at [cd4072e](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/cd4072e3dda70ef2c6626aacb60394ab8d7265b6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
@abdullahtahiriyo in https://github.com/FreeCAD/FreeCAD/pull/6971#discussion_r1045181940, you say you can't accept [0.5-conf, 1.5]. I'm not following your logic here. If the difference is so much, the user clearly wants a long spline, cut at the specific point. Do you mean the second case is where the user intended for a segment with `conf` distance between the points, and that FC should tell the user they are wrong instead of assuming they want to wrap?

> In case that the client code is not direct user input, but results from user interaction with FC (e.g. via the UI), the client code needs to ensure that it is asking the closest to the intention of the user that is possible for FC.

I'm OK with this. So any such checks should probably happen in the UI side.

AjinkyaDahale:
> The most prominent issue I have found is the possible memory leaks due to not releasing naked pointers upon exception.

> In order to move forward, please consider these extra commits: https://github.com/abdullahtahiriyo/FreeCAD_sf_master/pull/new/sketcher-split-more-curves-fixed

> I think I have not changed behaviour, or introduced extra bugs. But feel free to review the changes and come back with anything you do not like.

From a cursory look I don't see an issue.

> One aspect that I have seen, but I have decided not to change, is that it is possible to have calls to create new curves with a same start and end parameter (there is some code to reproduce in the notes-to-self).

This I do not understand. Do you mean running the `trim` in python with the same parameter for start and end? For periodic this will split it into a non-periodic curve, and for non-periodic curves I guess it'll cause OCC error.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [720617552](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/720617552) was triggered at [d4d5370](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/d4d537094b06ba5738ab408775f05118360aac20). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
@abdullahtahiriyo could you rebase your fixes to this branch? There are some conflicts.

abdullahtahiriyo:
> you say you can't accept [0.5-conf, 1.5]. I'm not following your logic here. If the difference is so much, the user clearly wants a long spline, cut at the specific point. Do you mean the second case is where the user intended for a segment with `conf` distance between the points, and that FC should tell the user they are wrong instead of assuming they want to wrap?

I will try one more time. The user code wants to trim a curve to a very small segment and asks for this:
u= 0.5-conf
v= 0.5

For a periodic b-spline. Evaluation in your code of condition ```v - u  <= Precision::PConfusion()``` yields true:

Following ```v = v + myCurve->LastParameter() - myCurve->FirstParameter(); // v needs one extra lap```, v is converted into v=1.5

The ```splitUnwrappedBSpline(u, v);``` function is called with u=0.5-conf and v=1.5. In practice, this leads to a curve with u=0.5 and v=1.5, so a "complete curve".

The function we are talking about is for trimming, not for splitting. Splitting just uses the trimming functionality to fulfill its goal. I am not trying to think what a user of the split function may want as a result. That is the work of the split function, not of the trimming function. The trimming function should just trim. If it cannot trim, it must throw an exception. Here the situation is that a user of the trim function is asking for a very small curve that OCC cannot deliver. Instead an error, the user is getting a full lap.

The Python code is recreating a trimming between u=0.5-conf and v=1.5, as this is what the c++ code would ultimately ask the ```splitUnwrappedBSpline(u, v);``` function to do. The trim function in Python is not calling this trimming function of c++ (maybe that is the source of the confusion).

> I'm OK with this. So any such checks should probably happen in the UI

Depending on the check, it may also go in the c++ split function in sketchobject. It may also be that it does not check, but reacts to the exception. It depends on the intended behaviour of the split function and what is being checked. For example, there are macros that will call the Python directly (without the UI). The c++ split() function needs to decide what service should provide this macros for wrong input parameters, it may be to get back an error, exception, ... If the problem is just of translating UI input into sensible parameters for Python, then it is indeed the UI part that should take care of this.

> From a cursory look I don't see an issue.

In here:

```
ok = createGeosFromNonPeriodic(
                static_cast<const Part::GeomBoundedCurve *>(geo),
                [](const Part::GeomCurve* curve, double startParam, double endParam) {
                    auto newBsp = static_cast<Part::GeomBSplineCurve *>(curve->clone());
                  newBsp->Trim(startParam, endParam);
                    return newBsp;
                },
                [this, &newConstraints](int GeoId, int newId0, int newId1) {
                    // apply appropriate constraints on the new points at split point
                    Constraint* joint = new Constraint();
                    joint->Type = Coincident;
                    joint->First = newId0;
                    joint->FirstPos = PointPos::end;
                    joint->Second = newId1;
                    joint->SecondPos = PointPos::start;
                    newConstraints.push_back(joint);

                    // transfer constraints from start and end of original spline
                    transferConstraints(GeoId, PointPos::start, newId0, PointPos::start, true);
                    transferConstraints(GeoId, PointPos::end, newId1, PointPos::end, true);
                });
```
This line:
```auto newBsp = static_cast<Part::GeomBSplineCurve *>(curve->clone());```

creates a naked pointer via clone(). Memory is dynamically allocated for this clone and the caller is responsible for releasing it.

Next line:
```newBsp->Trim(startParam, endParam);```

is called with parameters u, v that cause an OCC exception . OCC throws an exception. Part catches the exception and throws a Kernel exception that is unhandled in the caller code of split(), this propagates back until core handles any unhandled exception of certain types, which produces an error. Who has deallocated the memory chunck generated by clone()? Nobody has. It is now a memory leak.

> This I do not understand. Do you mean running the `trim` in python with the same parameter for start and end? For periodic this will split it into a non-periodic curve, and for non-periodic curves I guess it'll cause OCC error.

It is not for periodic, but for non-periodic B-Splines that this happens. In the comment linked to the code, there is an example which calls the trim function with u=v (I think it is with u=v=1).

**Comment by AjinkyaDahale** [comment]
> I will try one more time...

> The Python code is recreating a trimming between u=0.5-conf and v=1.5, as this is what the c++ code would ultimately ask the splitUnwrappedBSpline(u, v); function to do. The trim function in Python is not calling this trimming function of c++ (maybe that is the source of the confusion).

I see your point now, thanks. There may be merit in letting OCC handle this even for periodic cases. If the case that I was thinking (user wants to wrap around but somehow we have `0 < v - u < confusion`) does arise, we can do the wrapping before generating the python code.

**Comment by AjinkyaDahale** [comment]
>> From a cursory look I don't see an issue.
>
> This line:... creates a naked pointer via clone().
> ...Part catches the exception and throws a Kernel exception that is unhandled in the caller code of split(),... Who has deallocated the memory chunck generated by clone()? Nobody has. It is now a memory leak.

I should have been more clear. I meant I do not see an issue with the changes you made. Could you rebase them again so that I can merge them in this PR branch?

**Comment by AjinkyaDahale** [comment]
> It is not for periodic, but for non-periodic B-Splines that this happens. In the comment linked to the code, there is an example which calls the trim function with u=v (I think it is with u=v=1).

I see. Let me run the code, but for now let's just document the behaviour and make a decision later, unless it causes some serious issues.

**Comment by AjinkyaDahale** [comment]
So this is what I tried:
1. Make a B-spline (perhaps any other curve will do or behave slightly differently)
2. Fix an end point with the lock constraint (for convenience).
3. Split the curve at the coordinates of that locked end point.

This leads to a kernel error exception.
```
Base.CADKernelError: NCollection_Array1::Create In void Part::GeomBSplineCurve::Trim(double, double) in src/Mod/Part/App/Geometry.cpp:1610
```

image

In my opinion that is the correct behavior, though an exception with a better explanation may be possible. The parameter itself is not known until within the `SketchObject::split()` method (only the coordinates of the point are passed). So any check will have to be done there. Further, The user will have to intentionally run this through console since it is impossible to pass the end point by pre-selection when in GUI.

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [721149973](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/721149973) was triggered at [8c48fac](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/8c48facad4e611456087fbeecceb22bbae364f60). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by abdullahtahiriyo** [comment]
> I should have been more clear. I meant I do not see an issue with the changes you made. Could you rebase them again so that I can merge them in this PR branch?

I will try to rebase later on. 

The change basically is to produce a smart pointer. This smart pointer holds the responsibility to release the dynamic memory when it goes out of scope. When no exception happens, this smart pointer gets copy-elided to the caller of the lambda (effectively as if it were moved). When the exception happens, the stack must be unwound. This means that all local variables go out of scope, which means that the destructor of the smart pointer is called, and the memory released.

> If the case that I was thinking (user wants to wrap around but somehow we have `0 < v - u < confusion`) does arise, we can do the wrapping before generating the python code.

Exactly. This is what I was referring to when I was saying that the split tool (the UI part in your example) is responsible for translating the intention to the user when using the split tool into the trimming commands necessary for executing it.

> In my opinion that is the correct behavior, though an exception with a better explanation may be possible. The parameter itself is not known until within the `SketchObject::split()` method (only the coordinates of the point are passed). So any check will have to be done there. Further, The user will have to intentionally run this through console since it is impossible to pass the end point by pre-selection when in GUI.

About the message: 
Now the CADKernelError exception goes all the way to core, because it is not handled in SketchObject::split. The message is then not so meaningful. What you may do, is to catch CADKernelError exceptions in split(), report the underlying exception to console (to keep track of where the whole issue is originating), and then either provide your own meaningful message to console and tidy up or throw another exception with a more meaningful message (which will appear as the one you see, but with a better message).

About the behaviour:

The split command uses the addGeometry method at different times during the execution. Geometry property is changed on each call. This contrasts with other methods of SketchObject, where all the changes to the geometry property are done on a std::vector array and at the end the changes to the property are submitted. The first approach is more flexible. The second more efficient. Notwithstanding, the first introduces intermediate changes to the Geometry property. This means that, if the function is exited by exception after the Geometry property was changed, the execution of the command does not offer a transactional guarantee (i.e. that the operation will be executed in full, or not executed at all). I mean that the SketchObject split() function does not guarantee this. The UI function that calls the Python code does open a command and if an exception reaches it, rolls back any change (aborts the command). So, you will only see what I mean from Python: 

```
ActiveSketch.addGeometry(Part.BSplineCurve([App.Vector(-50,60),App.Vector(-60,70),App.Vector(-70,80),App.Vector(-80,90),App.Vector(-90,100)],None,None,False,3,None,False),False)
 
ActiveSketch.solve()

ActiveSketch.addGeometry(Part.ArcOfCircle(Part.Circle(App.Vector(-34.880920,77.786903,0),App.Vector(0,0,1),16.017876),1.635223,2.547379),False)
 
ActiveSketch.split(0,App.Vector(-89.9999999999999999999999999,99.9999999999999999999999999,0))
```

If you execute the code above, which will fail and exit with an exception, you will see that you end up with the original B-Spline and another B-Spline. Both essentially the same. This is because the one trim operation that failed is the second "part". If the split position trims first the exception-throwing curve, then nothing will be added. The arc is there in the code just to make more clear which B-Spline objects are newly added to the end of the list.

Generally the split() function works well and we are talking here just about corner cases.

What could be done differently in these corner cases? Well, when splitting a non-periodic, it should be possible to check at Python level that the two curves can be produced (at least a confusion check), if one cannot be produced, then the function may produce an error message to the console and return failure. In this case, no matter which point is selected, the outcome is consistent, even when the split function is asked to do something that it cannot do. In this case the original object would remain, which is basically what the user wants if one of the curves that need to be split is so small that cannot be handled by OCC.

**Comment by abdullahtahiriyo** [comment]
I have now rebased my branch on top of yours.

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [721206330](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/721206330) was triggered at [abca185](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/abca185a5f8c3b57eaa57af3eeac7be64734fdce). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by AjinkyaDahale** [comment]
> What could be done differently in these corner cases? Well, when splitting a non-periodic, it should be possible to check at Python level that the two curves can be produced (at least a confusion check), if one cannot be produced, then the function may produce an error message to the console and return failure.

The problem is split behaves differently for different curves, and what we are doing is passing a curve and a point, not a curve and the parameter where it should cut. So currently the check can only be done in split. However, there is a "trim" functionality which does seem to be designed more curve-type agnostically, since it handles B-splines automatically even when wiki said it didn't (I believe I changed that after testing).

**Comment by AjinkyaDahale** [comment]
@abdullahtahiriyo the corner case should be resolved now.

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [721405459](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/721405459) was triggered at [5f4a5df](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/5f4a5dfecc095e763e73d2308679bb6e007143b4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

File: src/Mod/Sketcher/App/SketchObject.cpp [hunk 5]
```diff
@@ -2970,160 +2978,250 @@ int SketchObject::split(int GeoId, const Base::Vector3d &point)
...
+            // find split point
+            curve->closestParameter(point, splitParam);
+            startParam = curve->getFirstParameter();
+            endParam = curve->getLastParameter();
+            // TODO: Using parameter difference as a poor substitute of length.
+            // Computing length of an arc of a generic conic would be expensive.
+            if (endParam - splitParam < Precision::PConfusion() ||
+                splitParam - startParam < Precision::PConfusion()) {
+                THROWM(ValueError, "Split point is at one of the end points of the curve.");
+                return false;
```
abdullahtahiriyo, line 3031: There are some considerations here:
1. The SketchObject split function returns an **int**, with an error code being a negative error code. This lambda returns a **boolean**, so it cannot convey an error code. An exception may be used to convey a specific message (THROWM). In this case, we are throwing a c++ exception (via THROWM), so ``` return false``` is unreachable (the execution path will never arrive there). This should be removed.
 
2. There is a choice to be made, for a given error, that the function may be left via c++ exception or via error code, but it cannot be both at the same time (i.e. for the same specific error). 

3. If error code is favoured and a function that is called may return an exception (as in the case of trim() ), then the exceptions are caught at SketchObject level and the corresponding error code is returned.
 
4. If exception is favoured, then the caller of this function shall catch it. When the caller is the Python binding (SketchObjectPyImp), it must catch it before leaving c++ land towards Python land. 

I am not sure myself why exactly the C++ exception must be converted to a Python exception. I wanted to have a conversation with @wwmayer about it, but suddenly I was not available and then I forgot. This can be a good occasion to ask him about the limitations in place and the underlying reasons.

In practical terms, because in SketchObject some functions just return error codes, but other functions may leave by exception (at least for some errors), in SketchObjectPyImp.cpp there are different approaches to this.

**Way A)** pure error code land:
```
    int err=this->getSketchObjectPtr()->setDatum(Index, Datum);
    if (err) {
        std::stringstream str;
        if (err == -1)
            str << "Invalid constraint index: " << Index;
        else if (err == -3)
            str << "Cannot set the datum because the sketch contains conflicting constraints";
        else if (err == -2)
            str << "Datum " << (const char*)Quantity.getUserString().toUtf8() << " for the constraint with index " << Index << " is invalid";
        else if (err == -4)
            str << "Negative datum values are not valid for the constraint with index " << Index;
        else if (err == -5)
            str << "Zero is not a valid datum for the constraint with index " << Index;
        else if (err == -6)
            str << "Cannot set the datum because of invalid geometry";
        else
            str << "Unexpected problem at setting datum " << (const char*)Quantity.getUserString().toUtf8() << " for the constraint with index " << Index;
        PyErr_SetString(PyExc_ValueError, str.str().c_str());
        return nullptr;
```
So handle the error codes and create Python exceptions (PyErr_SetString). No c++ exception reach here (if any, they were caught at SketchObject level and converted into error codes).

When we do this (PyErr_SetString), and we called the function from SketcherGui, we end up with a c++ exception at the caller (when Python returns control to the c++ code invoking the Python command, the python exception generates a c++ exception):

```
            try {
                Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Split edge"));
                Gui::cmdAppObjectArgs(sketchgui->getObject(), "split(%d,App.Vector(%f,%f,0))",
                                      GeoId, onSketchPos.x, onSketchPos.y);
                Gui::Command::commitCommand();
                tryAutoRecompute(static_cast<Sketcher::SketchObject *>(sketchgui->getObject()));
            }
                Base::Console().Error("Failed to split edge: %s\n", e.what());
                Gui::Command::abortCommand();
            }
```
So, the transaction guarantees are enforced here and any change to properties reverted. The message also contains the string that was put with ```PyErr_SetString```.

**WAY B)** Exiting SketchObject function via exception or error code (for some specific causes one, or other specific causes the other)

Then we need to catch this c++ exception and convert it into a Python exception.

```        try {
            int ret = this->getSketchObjectPtr()->addCopy(geoIdList, vect, false, Base::asBoolean(clone)) + 1;

            if(ret == -1)
                throw Py::TypeError("Copy operation unsuccessful!");

            std::size_t numGeo = geoIdList.size();
            Py::Tuple tuple(numGeo);
            for (std::size_t i=0; i<numGeo; ++i) {
                int geoId = ret - int(numGeo - i);
                tuple.setItem(i, Py::Long(geoId));
            }

            return Py::new_reference_to(tuple);
        }
        catch(const Base::ValueError & e) {
            throw Py::ValueError(e.getMessage());
        }```

Here I am looking into the catch block, which throws a Python exception. I think this is equivalent to the PyErr_SetString, but @wwmayer knows better and may confirm or deny this assertion or point out to differences.

So in this case, error codes and c++ exceptions, end up becoming Python exceptions. If the caller of the Python command was c++, then the caller gets a c++ exception corresponding to the Python counterpart, which shall be used to provide the transactional guarantee of a command.

N.B.: Often in this  post I assume that the caller of the SketchObject c++ function is Python, as it is the cases here. It needs not be, it can be another c++ function, but that is a different discussion.

AjinkyaDahale: OK I see the issue here. I had just followed snippets elsewhere in `SketchObject.cpp` etc. I have no preference from my end.

abdullahtahiriyo: I think it makes sense in this case to exit by THROWM, then catch the exception in SketchObjectPyImp, and either set an error with PyErr_SetString, or raise a Py::ValueError exception. I think that is what is left from your side.

In parallel I am tweaking the mechanism for notification of the parabola migration. I hope to have final code on my side this week.

Then with the two this PR can be merged. I find it is a very nice functionality.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [726951213](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/726951213) was triggered at [d2e5cdd](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/d2e5cddd4f11392504661f5dd88ca994d322e86a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/App/SketchObject.cpp
AjinkyaDahale: Are you sure we should do that? I don't see try-catch being used a lot in `SketchObjectPyImp.cpp`.

wwmayer:
> I am not sure myself why exactly the C++ exception must be converted to a Python exception. I wanted to have a conversation with @wwmayer about it, but suddenly I was not available and then I forgot. This can be a good occasion to ask him about the limitations in place and the underlying reasons.

Python itself is written in C and doesn't have an exception handling like in C++. And since we do not know what happens inside the C code before and after it has invoked our callback function we cannot be sure that everything is in a consistent state when somewhere a C++ exception is raised and not handled by the callback function.

At least for older MinGW builds the application crashed in such scenarios.

So, the safe way is to convert a C++ exception into a Python exception in the callback function and then the calling instance (C++) that has invoked the function of the C Python API (in most cases) can check the returned value whether it's null or not.

> Here I am looking into the catch block, which throws a Python exception. I think this is equivalent to the PyErr_SetString, but @wwmayer knows better and may confirm or deny this assertion or point out to differences.

Correct. Internally PyCXX uses PyErr_SetString.

AjinkyaDahale:
> So, the safe way is to convert a C++ exception into a Python exception in the callback function and then the calling instance (C++) that has invoked the function of the C Python API (in most cases) can check the returned value whether it's null or not.

@wwmayer so this translates into a `try-catch` block. I see both `PyErrSetString` and `throw Python::ValueError` being used. Which one is appropriate?

wwmayer:
>  Which one is appropriate?

Both ways are fine. When using PyErr_SetString you have to make sure to immediately return from the function with `return nullptr`. Using a PyCXX exception (or a sub-class of Base::Exception) does it do automatically for you. 

If you e.g. throw a PyCXX exception in SketchObjectPy::addGeometry then it will be caught in SketchObjectPy::staticCallback_addGeometry and this returns with a nullptr.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [727953778](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/727953778) was triggered at [edf8154](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/edf8154ac3c8c1e6d04a471621e0a7b8780a9f69). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [728026928](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/728026928) was triggered at [544984a](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/544984a5ffcc51a5a03ace2acda3fa14bbf80088). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
@wwmayer 

Many thanks for the explanation!! :)

abdullahtahiriyo:
@AjinkyaDahale 

Thank you for adding the exception handling.

I have modified the code to provide notifications of the parabola migration. I have created a new branch with my new code and on top of it I have put the commits you have just pushed here. 

I have added just one commit to remove the "return true" after exception throwing, as it is unreachable code.

This is the branch:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/commits/sketcher-split-more-curves-fixed_3

Would you just push that branch here?

AjinkyaDahale:
@abdullahtahiriyo just done. I only see the `return true` removed. Not sure where's the code for parabola migration notification.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [728158563](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/728158563) was triggered at [b4365b6](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/b4365b66dac20f07c1b840a204aed9162103cfc4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971">pipeline status</a> for feature branch [PR_6971](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6971). Pipeline [728177214](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/728177214) was triggered at [887b263](https://github.com/FreeCAD/FreeCAD/pull/6971/commits/887b2635aa5f737e7bd784f62ea1846077c62990). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
@AjinkyaDahale 

Thank you for such a nice functionality.

AjinkyaDahale:
Great. One issue I observe only now: since adding #7484 the splines becoming rational makes dragging difficult. This happens after splitting/joining.


https://user-images.githubusercontent.com/4316933/208946916-ccf413f6-016c-42b2-95ca-f8969372bec4.mp4

Of course for originally rational splines this doesn't make the situation any worse.

One solution might be to somehow add back the constraints if there are any.

abdullahtahiriyo:
@AjinkyaDahale 

Yes, I think we had this on the wanted list. Where constraints can be add back, then it would be way much more user friendly to do it.

Issue #6959, opened 2022-06-03
[Bug] [FEM] editing histogram blocks access to FreeCAD's main window


donovaly:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=18&t=69162

### Version

0.20 (Development)

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.29011 (Git)
Build type: Release
Branch: master
Hash: e8f1bde36707b31e4f0d9a0a909d0fee8cb74f00
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, OCC 7.6.2
```

### Subproject(s) affected?

FEM

### Issue description

- open the CCX result dialog
- select there e.g. the result "Displacement X"
- the Click the Histogram button
- in the histogram click in the toolbar the button "Configure subplots" (number 3 counted from the right)
result: the configure dialog opens behind the histogram but since the histogram can't be moved when a subdialog is open and since FC is blocked too, you cannot do anything anymore
The only way is to take the hammer of the Windows Task-manager to kill the histogram subdialog, therefore this is a rather major bug.
-> setting to 0.20 milestone

I assume the fix is to bring the subdialog to front. Additionally FC should not be frozen when a plot is open.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
@FEA-eng can u help us find Linux testers for this?

donovaly:
@chennes tested it earlier today. I committed now a simpler fix than in the above diff.

donovaly:
Many thanks!
Note that I got 2 positive test results from people at Linux before I merged this.

I'll comment further in #6989.

0penBrain:
> Note that I got 2 positive test results from people at Linux before I merged this.

I guess so. :) And indeed it "works" on Linux too. Notice that actually, the issue didn't exist on Linux --at least on mine-- as "StayOnTopHint" is relying on the WM that widely differs. ;)

PR #6989, opened 2022-06-08
[FEM] Implement a better dialog layout management for histogram

0penBrain:
Relates to #6959 

An side advantage of using Tool flag is that dialog opens as a pure child of FC, without adding a tab in the OS task bar.

### DIFF
--- a/src/Mod/Fem/femtaskpanels/task_result_mechanical.py
+++ b/src/Mod/Fem/femtaskpanels/task_result_mechanical.py
@@ -503,13 +503,19 @@ class _TaskPanel:
 
         if len(plt.get_fignums()) > 0:
             plt.close()
+        plt.ioff() # disable interactive mode so we have full control when plot is shown
         plt.hist(res_values, bins=50, alpha=0.5, facecolor="blue")
         plt.xlabel(res_unit)
         plt.title(translate("FEM","Histogram of {}").format(res_title))
         plt.ylabel(translate("FEM","Nodes"))
         plt.grid(True)
         fig_manager = plt.get_current_fig_manager()
-        fig_manager.window.setWindowFlags(QtCore.Qt.WindowStaysOnTopHint)    # stay ontop
+        # Lines below tells Qt that plot widget/dialog should be kept on top of its parent,
+        # its parent being defined as FC main window
+        # "Tool" type also is non modal, and dialog doesn't add a tab in the OS task bar
+        # See Qt::WindowFlags for more details
+        fig_manager.window.setParent(FreeCADGui.getMainWindow())
+        fig_manager.window.setWindowFlag(QtCore.Qt.Tool)
 
     def update_colors_stats(self, res_values, res_unit, minm, maxm):
         QApplication.setOverrideCursor(Qt.WaitCursor)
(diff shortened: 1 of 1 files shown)


donovaly:
Many thanks. This looks better and can go in.

Just 2 things:

Super cool would be the following:
- user opened the histogram
- user changed the result type in the FreeCAD panel and instead of that the histogram window is closed, it will be refreshed using the new result type. This way one can easily step through the different result types and compare e.g. Tresca with Mises stress distribution
I think I an capable of implementing this, but have no time right now. Since you are at it already, maybe you can do this?

Please keep commenting ;-). It often take me a while to figure out what code does when I try to fix things. I would even say more than 50% of the work to fix bugs is to find out what the existing code does. I commented in your code.

src/Mod/Fem/femtaskpanels/task_result_mechanical.py
```diff
@@ -503,13 +503,15 @@ def result_selected(self, res_type, res_values, res_unit, res_title):
 
         if len(plt.get_fignums()) > 0:
             plt.close()
+        plt.ioff()
```
donovaly, line 506: What does this do? Can you please add a comment?

src/Mod/Fem/femtaskpanels/task_result_mechanical.py
```diff
@@ -503,13 +503,15 @@ def result_selected(self, res_type, res_values, res_unit, res_title):
...
         plt.hist(res_values, bins=50, alpha=0.5, facecolor="blue")
         plt.xlabel(res_unit)
         plt.title(translate("FEM","Histogram of {}").format(res_title))
         plt.ylabel(translate("FEM","Nodes"))
         plt.grid(True)
         fig_manager = plt.get_current_fig_manager()
-        fig_manager.window.setWindowFlags(QtCore.Qt.WindowStaysOnTopHint)    # stay ontop
+        fig_manager.window.setParent(FreeCADGui.getMainWindow())
+        fig_manager.window.setWindowFlag(QtCore.Qt.Tool)    # stay ontop
```
donovaly, line 514: Can you please add a comment? For me as non-Pythionist it is not clear why setting a flag "QtCore.Qt.Tool" puts the window on top

0penBrain: Nothing to do with Python. ;) It's pure Qt Window Flag : https://doc.qt.io/qt-5/qt.html#WindowType-enum
Will add a comment too.

0penBrain:
> Super cool would be the following:
> 
> * user opened the histogram
> * user changed the result type in the FreeCAD panel and instead of that the histogram window is closed, it will be refreshed using the new result type. This way one can easily step through the different result types and compare e.g. Tresca with Mises stress distribution
>   I think I an capable of implementing this, but have no time right now. Since you are at it already, maybe you can do this?

I thought the same, but I strongly believe it should be processed separately. ;) I has deeper impact on how things are implemented (architectured). I lack of time now too, but may give it a try lateR.

> Please keep commenting ;-). It often take me a while to figure out what code does when I try to fix things. I would even say more than 50% of the work to fix bugs is to find out what the existing code does. I commented in your code.

I would have said 75% :)

0penBrain:
@donovaly should be OK now

donovaly:
Many thanks! Merging.

PR #6995, opened 2022-06-08
Spreadsheet: improve range binding feature

realthunder:
Forum thread https://forum.freecadweb.org/viewtopic.php?f=3&t=69219

@donovaly


donovaly:
Awesome work!

So now only 2 minor issues are left, see my forum post.

PR #6997, opened 2022-06-09
Spreadsheet: fix range checking

realthunder:
Forum post: https://forum.freecadweb.org/viewtopic.php?f=3&t=69219&p=600767#p600835

@donovaly

### DIFF
--- a/src/App/Range.cpp
+++ b/src/App/Range.cpp
@@ -37,7 +37,7 @@ using namespace App;
 const int App::CellAddress::MAX_ROWS = 16384;
 const int App::CellAddress::MAX_COLUMNS = 26 * 26 + 26;
 
-Range::Range(const char * range)
+Range::Range(const char * range, bool normalize)
 {
     std::string from;
     std::string to;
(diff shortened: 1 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6997">pipeline status</a> for feature branch [PR_6997](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_6997). Pipeline [559297518](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/559297518) was triggered at [8eeba82](https://github.com/FreeCAD/FreeCAD/pull/6997/commits/8eeba82ac95b494ceaa377918f018449e06db57d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/App/Range.cpp
```diff
@@ -37,7 +37,7 @@ using namespace App;
 const int App::CellAddress::MAX_ROWS = 16384;
 const int App::CellAddress::MAX_COLUMNS = 26 * 26 + 26;
 
-Range::Range(const char * range)
+Range::Range(const char * range, bool normalize)
```
chennes, line 40: Per @wwmayer's various comments recently about "boolean blindness", it might be a good idea to make this an enum (or something along those lines) instead of a bool.

realthunder: But I thought he is referring to cases of multiple bool arguments? Does that mean we aren't supposed to use any bool as function argument, @wwmayer?

donovaly: I also understood it as @realthunder . A single bool does not harm in my opinion.

donovaly:
Tested and works well. Many thanks!

A general note: please add comments why is the code is as it is.
For example that the code logic is to have the range defined from left top to right bottom but users might do the opposite (for example users from countries with RTL languages). For another developer looking at this in one year it is hard to get what the normalization is about and why it is needed.

Merging now anyway. A PR with some code comments for the binding feature is welcome anytime ;-)

src/App/Range.cpp
wwmayer: If from the function name it's obvious what it does then a (single) boolean is perfectly allowed. E.g. for `setVisible(bool)` it is clear what the arguments `true` or `false` will do.

If the intention of a boolean is not clear and one has to look at the documentation in order to get its meaning then the boolean might be the wrong type. Here is an article about Qt design principles that addresses this issue too: https://wiki.qt.io/API_Design_Principles#Avoiding_Common_Traps

In the past we had some cases where functions accepted several boolean (in an extreme case up to six) which does not only make client code unreadable but also the implementation of that function.

PR #7013, opened 2022-06-13
GUI: Preliminary support for HiDPI splashscreen

chennes:
Unfortunately the 0.20 splashscreen was generated for the original pixel size of a non-HiDPI screen, but this chunk of code detects when we are on a device with a higher-than-one pixel ratio and switches to using an upscaled version of the image. In future releases we should generate both the 1x and the 2x images so both look their best.

In creating this PR I also attempted many different settings to try to get the font to look better, but could not find anything that worked. I've left in the part that I think will be required (using a floating point value for the font size), but have removed all of my other attempts. If anyone else can get that font rendering to look better I'd welcome your patch!


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7013">pipeline status</a> for feature branch [PR_7013](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7013). Pipeline [561910974](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/561910974) was triggered at [acae3f8](https://github.com/FreeCAD/FreeCAD/pull/7013/commits/acae3f843b461348c61ff901344725b4a0539360). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
The feedback was positive, so merge it now.

PR #7019, opened 2022-06-14
Rename FindOpenCasCade.cmake to FindOCC.cmake

adrianinsaval:
To fix a warning
```
CMake Warning (dev) at C:/msys64/mingw64/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (OCC) does
  not match the name of the calling package (OpenCasCade).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cMake/FindOpenCasCade.cmake:120 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cMake/FreeCAD_Helpers/SetupOpenCasCade.cmake:4 (find_package)
  CMakeLists.txt:62 (SetupOpenCasCade)
This warning is for project developers.  Use -Wno-dev to suppress it.
```
See: https://forum.freecadweb.org/viewtopic.php?f=4&t=69401#p601937


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7019">pipeline status</a> for feature branch [PR_7019](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7019). Pipeline [563630675](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/563630675) was triggered at [bcd7e3b](https://github.com/FreeCAD/FreeCAD/pull/7019/commits/bcd7e3b386f99cfe325d1a4a6bf6e8b7d41f08c0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Looks fine to me, however I lack experience if this change could have side effects for certain OSes.
-> @wwmayer can you please have a look?

luzpaz:
We should run it by the various FreeCAD packager, right?

adrianinsaval:
Just the compilers if I'm not mistaken, or not even that, just the configuration. It worked fine for me (and for the CI) with cmake on linux, would be nice to have someone testing on windows and Mac but this shouldn't change the outcome for any platform, it's the same script that gets called, just with a different name.

ageeye:
On MacOS with Homebrew this PR works fine.

chennes (approved):
This runs fine for me on Windows, and silences the cMake warning as intended. This is needed because cMake has recently been modified to be less tolerant of the name mismatch that we had in our script. I don't think there is any downside to this, but I'll wait a short while on merging in case Werner wants to comment.

PR #7031, opened 2022-06-15
Gui: correctly save and restore message type enabling in report view

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=3&t=69419

### DIFF
--- a/src/Base/ConsoleObserver.h
+++ b/src/Base/ConsoleObserver.h
@@ -70,6 +70,44 @@ private:
     void Log    (const char *sErr);
 };
 
+/** The ILoggerBlocker class
+ *  This class allows to temporary block then automatically restore arbitrary message types
+ *  on a particular console observer.
+ */
+class BaseExport ILoggerBlocker
+{
+public:
+    // Constructor that will block message types passed as parameter. By default, all types are blocked.
+    inline explicit ILoggerBlocker(const char* co, ConsoleMsgFlags msgTypes =
+        ConsoleSingleton::MsgType_Txt | ConsoleSingleton::MsgType_Log | ConsoleSingleton::MsgType_Wrn | ConsoleSingleton::MsgType_Err);
+    // Disable copy & move constructors
+    ILoggerBlocker(ILoggerBlocker const&) = delete;
+    ILoggerBlocker(ILoggerBlocker const &&) = delete;
+    // Disable assignment & move-assignment operator
+    ILoggerBlocker& operator=(ILoggerBlocker const&) = delete;
+    ILoggerBlocker& operator=(ILoggerBlocker const&&) = delete;
+    // Destructor that will restore message type settings.
+    inline ~ILoggerBlocker();
+private:
+    ConsoleMsgFlags msgTypesBlocked = 0; // Stores message types blocked by the blocker
+    const char* conObs; //  Stores console observer name that blocker acts on
+};
+
+ILoggerBlocker::ILoggerBlocker(const char* co, ConsoleMsgFlags msgTypes):
+conObs(co)
+{
+    msgTypesBlocked = Console().SetEnabledMsgType(conObs, msgTypes, false);
+}
+
+ILoggerBlocker::~ILoggerBlocker()
+{
+    auto debug = Console().SetEnabledMsgType(conObs, msgTypesBlocked, true);
+#ifdef FC_DEBUG
+    if (debug != msgTypesBlocked)
+        Console().Warning("Enabled message types have been changed while ILoggerBlocker was set\n");
+#endif
+}
+
 class BaseExport RedirectStdOutput : public std::streambuf
 {
 public:
--- a/src/Gui/BitmapFactory.cpp
+++ b/src/Gui/BitmapFactory.cpp
@@ -338,9 +338,10 @@ QPixmap BitmapFactoryInst::pixmapFromSvg(const QByteArray& originalContents, con
 
     QPainter p(&image);
     // tmp. disable the report window to suppress some bothering warnings
+    auto tmp = Base::Console().IsMsgTypeEnabled("ReportOutput", Base::ConsoleSingleton::MsgType_Wrn);
     Base::Console().SetEnabledMsgType("ReportOutput", Base::ConsoleSingleton::MsgType_Wrn, false);
     QSvgRenderer svg(contents);
-    Base::Console().SetEnabledMsgType("ReportOutput", Base::ConsoleSingleton::MsgType_Wrn, true);
+    Base::Console().SetEnabledMsgType("ReportOutput", Base::ConsoleSingleton::MsgType_Wrn, tmp);
     svg.render(&p);
     p.end();
 
--- a/src/Gui/BitmapFactory.cpp
+++ b/src/Gui/BitmapFactory.cpp
@@ -42,6 +42,7 @@
 
 #include <App/Application.h>
 #include <Base/Console.h>
+#include <Base/ConsoleObserver.h>
 
 #include "BitmapFactory.h"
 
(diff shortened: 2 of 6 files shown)


0penBrain:
@wwmayer : this is minimal fix. Do you want me to have something more generic (maybe something based on variable instanciation/deletion such as QSignalBlocker) ? Or is it fine as it for now ?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7031">pipeline status</a> for feature branch [PR_7031](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7031). Pipeline [564333258](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/564333258) was triggered at [932801f](https://github.com/FreeCAD/FreeCAD/pull/7031/commits/932801f522e1029b507dabdf5894700645dc009d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
Added a 2nd commit for another similar issue with normal messages (first commit is for warnings)
https://forum.freecadweb.org/viewtopic.php?f=3&t=69419&p=602530#p602511

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7031">pipeline status</a> for feature branch [PR_7031](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7031). Pipeline [565344474](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/565344474) was triggered at [e3db0a0](https://github.com/FreeCAD/FreeCAD/pull/7031/commits/e3db0a0d9d1591d4a190f6688f54f3cb42ec946a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
Ah class like QSignalBlocker would be good, maybe ILoggerBlock or so. The class should be added to ConsoleObserver.h and not Console.h to avoid to recompile big parts of the code.

TIA.

0penBrain:
> Ah class like QSignalBlocker would be good, maybe ILoggerBlock or so. The class should be added to ConsoleObserver.h and not Console.h to avoid to recompile big parts of the code.

Should be able to tackle this next week. In between you can merge this PR as is (I'll open a new one) or wait for the update. As you wish.

wwmayer:
> Could you comment on this first proposal so I know if it's roughly what you expected ?

Looks basically fine but there are a few things that can be improved:
* The int should be replaced with ConsoleMsgFlags (Although there is comment in the code to remove it you should use the typedef for the time being. Once it will be removed the compiler will tell us if we forget to adjust this class)
* You don't have to iterate over the enum values by yourself. Instead just do `msgTypesBlocked = Console().SetEnabledMsgType(conObs, msgType, false);` and in the destructor `Console().SetEnabledMsgType(conObs, msgTypesBlocked, true);`
* If a copy constructor and assignment operator is defined (or disabled in this case) it should also be done for the move constructor and move assignment operator

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7031">pipeline status</a> for feature branch [PR_7031](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7031). Pipeline [569310140](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/569310140) was triggered at [686bf59](https://github.com/FreeCAD/FreeCAD/pull/7031/commits/686bf5942f64d2a3ba9865b8e18549985728fbff). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
> * The int should be replaced with ConsoleMsgFlags (Although there is comment in the code to remove it you should use the typedef for the time being. Once it will be removed the compiler will tell us if we forget to adjust this class)
> * You don't have to iterate over the enum values by yourself. Instead just do `msgTypesBlocked = Console().SetEnabledMsgType(conObs, msgType, false);` and in the destructor `Console().SetEnabledMsgType(conObs, msgTypesBlocked, true);`

:confused: I'm half surprised and half shameful here. I expected this and don't know why didn't catch it was there so I used a "workaround" that actually wasn't needed. Anyway it's fixed now (and definitely better).
> * If a copy constructor and assignment operator is defined (or disabled in this case) it should also be done for the move constructor and move assignment operator

I'm not an expert, but according https://en.cppreference.com/w/cpp/language/move_constructor and https://en.cppreference.com/w/cpp/language/move_assignment it's not needed.
Also https://stackoverflow.com/questions/37276413/default-move-constructor-assignment-and-deleted-copy-constructor-assignment is interesting.
Anyway, I added them. Tell me if you prefer it stays or if I can remove it in the light of provided links.

I added a 4th commit just showing how I tested but this shall be deleted before merging. I'll do when it's ready. Thanks for review

wwmayer:
> I'm not an expert, but according

I am not an expert either. I only remembered the blog of a C++ expert who said that if copy constructor or assignment operators are declared then move constructor or move assignment must be declared, too. However, there obviously is a difference if the functions are disabled.

On the other hand if they are explicitly disabled then it is at least clearer.

0penBrain:
> I am not an expert either. I only remembered the blog of a C++ expert who said that if copy constructor or assignment operators are declared then move constructor or move assignment must be declared, too. However, there obviously is a difference if the functions are disabled.

Yes if I got it well, as soon as copy/assignment is user-defined, implicit move/move-assignment won't be generated so you need to tell the compiler what to do. Not needed in our case because not having them is what we want, but doesn't hurt to explicitly delete them.

Tell me when you're ready to merge (or have other remarks). I'll remove the 4th commit then rebase. Thx

wwmayer:
IMO it's not a great idea to mix client code with test code. Before we have set-up the infrastructure for the [TDD](https://forum.freecadweb.org/viewtopic.php?f=10&t=68799) the code of ILoggerBlockerTest should be moved to the Test workbench instead.

For the time being it could be added to AppTestGui.cpp and the function `testILoggerBlocker()` be added to the Module class. Then in BaseTests.py add a new test method to ConsoleTestCase:
```
def testILoggerBlocker(self):
    if FreeCAD.GuiUp:
        import QtUnitTest
        QtUnitTest.testILoggerBlocker()
```

0penBrain:
@wwmayer OK, here is what I ended up with (only 4th commit is changed compared to previous push).
Just I'm not comfortable with how I can make a test to fail in C++ without actually aborting other tests or even app. Could you please tell what's the right way ?

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7031">pipeline status</a> for feature branch [PR_7031](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7031). Pipeline [578985891](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/578985891) was triggered at [e3431f2](https://github.com/FreeCAD/FreeCAD/pull/7031/commits/e3431f2e983e7ab1ce9073ec3bbb6a784429a560). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> Just I'm not comfortable with how I can make a test to fail in C++ without actually aborting other tests or even app. Could you please tell what's the right way ?

Not quite sure what are the difficulties you have. So, if you want to have a C++ function that fails then follow this recipe:
* inside the C++ function throw an exception (preferred is a sub-class of Base::Exception)
* the Python wrapper function should already convert it into a suitable Python exception type. So, in general there is nothing you have to do here
* in the Python test function it's assumed that an exception is raised. So, write the code accordingly.
Example:
It's expected that the function raises a RuntimeError. So write this test code:
```
def testSomething(self):
    with self.assertRaises(RuntimeError):
        obj.callTheWrapperFunction() # this function is supposed to throw a RuntimeError
```

The test will pass if callTheWrapperFunction() throws the expected RuntimeError. The test will fail if either no exception or a different exception is thrown

0penBrain:
> Not quite sure what are the difficulties you have. So, if you want to have a C++ function that fails then follow this recipe:

Sorry if it was unclear. What I was unhappy with was to use a Py::RuntimeError here : https://github.com/FreeCAD/FreeCAD/blob/212b140d82cda570c20532c21534be9c3c137c45/src/Mod/Test/Gui/AppTestGui.cpp#L68
I felt like it wasn't optimal.
My problem is that if I use a `Base::RuntimeError`, the complete sequence of tests is stopped at fail time. While with `Py::RuntimeError`, the test fails but then the next tests in the sequence are still run (which is what I expect).
So I just wondered if there was a better (more proper) way to achieve this. ;)

PR #7096, opened 2022-06-25
[Path] Fix import error if camotics not installed.

sliptonic:
The camotics python module is provided when the camotics application is installed.  This is not a dependency for FreeCAD, just another applicaiton that the user can install. If it's installed (and recent enough), Path can use it for simulation.
There was already a test in place when the Path workbench loaded which determined whether to enable the button.  But the camotics gui was being imported anyway and was, in turn, trying to import the python module.  this would throw an error.

Commit wraps the import in another test to avoid the error.

---

### DIFF
--- a/src/Mod/Path/PathScripts/PathGuiInit.py
+++ b/src/Mod/Path/PathScripts/PathGuiInit.py
@@ -80,13 +80,14 @@ def Startup():
 
         # If camotics is installed and current enough, import the GUI
         try:
+            import camotics
             r = subprocess.run(
                 ["camotics", "--version"], capture_output=True, text=True
             ).stderr.strip()
             major, minor, patch = r.split(".")
             if int(major) >= 1 and int(minor) >= 2 and int(patch) >= 2:
                 from PathScripts import PathCamoticsGui
-        except FileNotFoundError:
+        except (FileNotFoundError, ModuleNotFoundError):
             pass
 
         Processed = True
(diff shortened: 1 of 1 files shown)


agren:
Might be an idea to also check that the camotics module exists.
I can see a case where:
A user who use camotics but does not care about camotics integration cannot run Path because FreeCAD can't find the camotics module.

Something like:
```python
try:
    import camotics
    r = subprocess.run(
    ...
except (FileNotFoundError, ImportError):
    pass
```

Might even be an idea to import and check the module version. But i guess there were no python module available before 1.2.2?
```python
    from camotics import VERSION as CAMOTICS_VERSION
```

sliptonic:
I think I see your point.  I'm assuming that if Camotics 1.2.2 is installed that the python module is also available.  I'll make the suggested change.

Yes, the python module is new.  There was an earlier version with camsim but I'm avoiding that.

Russ4262:
I did not see this PR until now, I can close #7104 if you like, since these two PRs address the same problem.  #7104 might be a little simpler, though it likely needs the dual error type catch as suggested here by @agren .

fra589:
Hi @sliptonic 
  
There is another error with this functionality, when the `camotics --version` don't give the 3 numbers major, minor and patch.

In my Debian, camotics is installed in v1.2.0 and the result of version give only `1.2`
So, the Path module don't load with the error: `not enough values to unpack (expected 3, got 2)`  

You should also add:
```
            except ValueError:
                pass
```
to the PathWorkbench Initialize()
  
Regards,
@++;
Gauthier

sliptonic:
> 
> There is another error with this functionality, when the `camotics --version` don't give the 3 numbers major, minor and patch.
> 


Hmm.  Third time's the charm?   Can you try this: https://github.com/FreeCAD/FreeCAD/pull/7111   
Thanks Gauthier.

PR #7098, opened 2022-06-25
Sketcher: Python converter

abdullahtahiriyo:
It does not add visible functionality.

Class to generate Python commands to add geometry and constraints.


abdullahtahiriyo:
@berndhahnebach 

Hi! I see the runner works fine for master. Are PRs not (any longer) handled by the runner?

PR #7100, opened 2022-06-26
FEM WB Elmer viscosity bug fix.

mist90:
FEM WB: error "Requested property: [Viscosity], not found" if used mixed solid and fluid simulation in Elmer is fixed.
When project uses thermal-flow simulation for solid and fluid together then Navier-Stokes equation is applied both for solid and fluid. But solid has not "Viscosity" property so Elmer prints warnings and calculation has very long time or may not converged.
This fix disables Stokes-Navier equation for Solid objects.

### DIFF
--- a/src/Mod/Fem/femsolver/elmer/writer.py
+++ b/src/Mod/Fem/femsolver/elmer/writer.py
@@ -694,6 +694,10 @@ class Writer(object):
             youngsModulus *= 1e3
         return youngsModulus
 
+    def _isMaterialFlow(self, body):
+        m = self._getBodyMaterial(body).Material
+        return "KinematicViscosity" in m
+
     def _handleFlow(self):
         activeIn = []
         for equation in self.solver.Group:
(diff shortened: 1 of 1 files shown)


donovaly:
Many thanks! However to understand the issue better and to find maybe further issues, can you please attach a small FreeCAD example file that demonstrate the problem?

In general, having a thread in the forum would be nice for the FEM community.

mist90:
Here is simple example:
[fem-elmer-simple.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/8991240/fem-elmer-simple.FCStd.zip)
Now calculation are made in fixed version FreeCAD. If run version without fix then it appears "Requested property: [Viscosity], not found" messages and solve will not converge in Elmer.
Thread in the forum here:
https://forum.freecadweb.org/viewtopic.php?f=17&t=69805

donovaly:
Many thanks!
Your example also unveiled some further bugs, see my forum post.

PR #7103, opened 2022-06-26
[PD] add preferences setting for transparency

donovaly:
- allow to specify the default  transparency for new shapes

### DIFF
--- a/src/Gui/PreferencePackTemplates/Window_Colors.cfg
+++ b/src/Gui/PreferencePackTemplates/Window_Colors.cfg
@@ -13,6 +13,7 @@
           <FCInt Name="BacklightIntensity" Value="100"/>
           <FCUInt Name="BoundingBoxColor" Value="4294967295"/>
           <FCUInt Name="DefaultShapeColor" Value="3435973887"/>
+          <FCInt Name="DefaultShapeTransparency" Value="0"/>
           <FCBool Name="EnableBacklight" Value="0"/>
           <FCBool Name="Gradient" Value="1"/>
           <FCUInt Name="HighlightColor" Value="4294907903"/>
--- a/src/Gui/ViewParams.h
+++ b/src/Gui/ViewParams.h
@@ -56,6 +56,7 @@ public:
     FC_VIEW_PARAM(DefaultShapeLineColor,unsigned long,Unsigned,421075455UL) \
     FC_VIEW_PARAM(DefaultShapeVertexColor,unsigned long,Unsigned,421075455UL) \
     FC_VIEW_PARAM(DefaultShapeColor,unsigned long,Unsigned,0xCCCCCC00) \
+    FC_VIEW_PARAM(DefaultShapeTransparency,int,Int,0)        \
     FC_VIEW_PARAM(DefaultShapeLineWidth,int,Int,2) \
     FC_VIEW_PARAM(DefaultShapePointSize,int,Int,2) \
     FC_VIEW_PARAM(CoinCycleCheck,bool,Bool,true) \
--- a/src/Gui/ViewProviderGeometryObject.cpp
+++ b/src/Gui/ViewProviderGeometryObject.cpp
@@ -70,18 +70,20 @@ ViewProviderGeometryObject::ViewProviderGeometryObject()
         b = (float)rand()/fMax;
     }
     else {
-        unsigned long shcol = hGrp->GetUnsigned("DefaultShapeColor",3435973887UL); // light gray (204,204,204)
+        unsigned long shcol = hGrp->GetUnsigned("DefaultShapeColor", 3435973887UL); // light gray (204,204,204)
         r = ((shcol >> 24) & 0xff) / 255.0;
         g = ((shcol >> 16) & 0xff) / 255.0;
         b = ((shcol >> 8) & 0xff) / 255.0;
     }
+
+    int initialTransparency = hGrp->GetInt("DefaultShapeTransparency", 0); 
 
     static const char *dogroup = "Display Options";
     static const char *sgroup = "Selection";
     static const char *osgroup = "Object Style";
 
     ADD_PROPERTY_TYPE(ShapeColor, (r, g, b), osgroup, App::Prop_None, "Set shape color");
-    ADD_PROPERTY_TYPE(Transparency, (0), osgroup, App::Prop_None, "Set object transparency");
+    ADD_PROPERTY_TYPE(Transparency, (initialTransparency), osgroup, App::Prop_None, "Set object transparency");
     Transparency.setConstraints(&intPercent);
     App::Material mat(App::Material::DEFAULT);
     ADD_PROPERTY_TYPE(ShapeMaterial,(mat), osgroup, App::Prop_None, "Shape material");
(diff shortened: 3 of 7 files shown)


donovaly:
Works fine for PartDesign, but not yet for Part.

donovaly:
Now it also works for Part

donovaly:
@chennes , can you please cross-check this PR? It works now fine for me
-> in the preferences you can specify a default transparency for new shapes in Part/PD.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7103">pipeline status</a> for feature branch [PR_7103](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7103). Pipeline [578984150](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/578984150) was triggered at [ffa57b8](https://github.com/FreeCAD/FreeCAD/pull/7103/commits/ffa57b8edc8ec4fca4e27c6fd5d88cdc56894aa6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes (approved):
This looks fine to me, and works as expected in my testing.

donovaly:
Many thanks for testing. merging then,

Roy-043:
Would it be OK to change singleStep from 10 to 1?
https://github.com/FreeCAD/FreeCAD/blob/5cc9bae455b0158a95f42a123cd5be4776dee2bb/src/Mod/Part/Gui/DlgSettingsObjectColor.ui#L98-L99

Compare:
Edit -> Preferences -> Display -> 3D view -> Relative size

adrianinsaval:
In any case I would go with 5% step on both, 1% change is not noticeable and if you _really_ need that much fine tuning you might as well write the number.

Roy-043:
I have edited my comment.

donovaly:
I Agree with Adrian here. a 1% change is not really visible the spin buttons are there to speed up the workflow, therefore having them at 1% would be no benefit at all.

In general it should be consistent. The question is to what value. Personally, I never had the need to set a transparency to e.g. 35%, 30 or 40% does the job. Therefore I vote for 10. Also because the use case I have daily is to make objects half-transparent. Then clicking 5 times is more suitable than clicking 10 times if the step would be 5%.

Roy-043:
The daily use argument can only apply to the Transparency in the Property editor which has a 1% step size. Using a 1% step there and a 10% step in the preferences makes no sense.

donovaly:
> The daily use argument can only apply to the Transparency in the Property editor which has a 1% step size. Using a 1% step there and a 10% step in the preferences makes no sense.

as I wrote, it should be consistent and 1% is not helpful.
So what is your opinion?

I don't understand your comment since in the appearances dialog we use 5%, in the preferences I implemented 10%. So we need a decision if we go with 5 or 10 %.

Roy-043:
In the [Property editor](https://wiki.freecadweb.org/Property_editor) the step size for the Transparency is 1%. Apparently you never change the transparency of objects there, but my guess is that many users do. I rarely use the appearance dialog. Note that the step size there is 3% (not 5%).

So we have 1%, 3% and 10%. The 1% step size may not be a deliberate choice on the part of the programmers, but in general step sizes are small rather than big. See f.e. the https://wiki.freecadweb.org/PartDesign_Preferences#Shape_view preferences (step sizes 0.01% and 0.5°).

I think 10% is too big for the transparency. But for consistency whatever is decided should also be applied to the Property editor.

donovaly:
> In the [Property editor](https://wiki.freecadweb.org/Property_editor) the step size for the Transparency is 1%. Apparently you never change the transparency of objects there, but my guess is that many users do. I rarely use the appearance dialog. Note that the step size there is 3% (not 5%).

Yes, I use the appearance dialog all the time. However for me the step is 5 %, I will have to investigate why it is 3% for you.

The property editor is sadly treated stepmotherly . For example for TechDraw the step sizes were almost different everywhere from the dialogs. For the dialogs people took care and used reasonable values, for the property editor only the default, which is "1", independent of the unit. So the property editor must be adopted to follow the counterpart in the dialogs. (For TechDraw I hope I fixed all cases, if you see there a mistake, please tell me.)
For Part and PartDesign checking the spin buttons step consistency is since a while on my ToDo list.

> So we have 1%, 3% and 10%. The 1% step size may not be a deliberate choice on the part of the programmers, but in general step sizes are small rather than big. See f.e. the https://wiki.freecadweb.org/PartDesign_Preferences#Shape_view preferences (step sizes 0.01% and 0.5°).

The step size must depend on the use case. For the angular deflection 0.5° make a difference while for the transparency 0.5% is not visible by human eyes.

Here with the transparency I think we agree that a step of 1% is not sensible because this difference cannot be seen and using the buttons dozens of times to get the desired result, leads to the effect that the buttons won't be used at all - users will use the slider (if there is any) or input the number directly. But the main benefit of the buttons is that you save the time to leave the mouse, move your hands to the keyboard and type.
What I learned from the ergonomics point of view that clicking a button 5 times is quicker than input a value while clicking it 10 times can be as time consuming as inputting the value directly.
Therefore in this case the question is still for me if we should go with 5 or 10%.

I can only tell from my use case and this is to make objects in most cases half-transparent, sometimes a bit more, sometimes a bit less. So with 10% I have to click 5 times while with 5% I would have to click 10 times and then would probably choose to input "50" directly and not using the buttons. But I think the step size should be so that the usage of the button has a benefit.
However, my use case is maybe special, so I will not insist on the 10% but that it should be at least 5%.

What is your use case, what transparency levels do you need in practice?

donovaly:
> It is possible to hold down the spin buttons, and also use the scroll wheel when a spinbox has the focus.

Cool. I did not know this yet.

> With a 1% step size you can hold down a spin button until you are close to the required value and then use a small number of separate clicks to reach the exact number you are after. With a 10% step size holding down a spin button does not make sense anymore. That alone rules out the 10% step IMO.

I see. I played now a but and think 5% would be a good compromise because:
- it suits noobs like me who don't know one can keep the button pressed
- those who know can easily set e.g. 25% or 75% transparency. (Besides 50% I assume these are the must used values)

If you agree, I will make a PR that sets all to 5%.

> Regarding the step size in the 'appearance dialog': it is indeed 5%, but if you scroll while hovering the sliders it is 3% (but that value may depend on your mouse settings).

Must be a Qt or mouse settings thing because here (window 10,) it is always 5%.

donovaly:
> I will make a PR that sets all to 5%.

Here it is: #7723

Roy-043:
If I understand correctly you have changed the step size in the Property editor as well. That is indeed consistent, but is also a change that might surprise other users. It may be wise to announce this in the Forum.

donovaly:
> If I understand correctly you have changed the step size in the Property editor as well. That is indeed consistent, but is also a change that might surprise other users. It may be wise to announce this in the Forum.

This change will not be backported. It will be listed in the release notes of 1.0 as change therefore I think its gets the necessary attention. The consistency between the dialog and property editor is an improvement in my opinion.

(I don't think for the property editor the devs took care but just took the default step. This is my experience with several WBs.)

To resolve the issue: I merge now, please check the next weekly and when you see further inconsistencies, please directly open an issue and CC me. Please also report any inconsistencies between dialog and property editor (all WBs) since this would help me to resolve this long-standing issue.

PR #7110, opened 2022-06-28
[Sketcher] Only move one piece of B-spline when dragging

AjinkyaDahale:
This is useful since splines can get very long.

fc-sketcher-move-piece

Not yet complete. Next steps include:
1. ~This currently doesn't work well with periodic B-splines.~ This is potentially fixed now. Could use more testing.
2. There should be options for both moving the entire B-spline or just one piece.

### DIFF
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -4021,6 +4021,66 @@ void Sketch::resetInitMove()
     isInitMove = false;
 }
 
+int Sketch::initBSplinePieceMove(int geoId, PointPos pos, const Base::Vector3d& firstPoint, bool fine)
+{
+    isFine = fine;
+
+    geoId = checkGeoId(geoId);
+
+    clearTemporaryConstraints();
+
+    // don't try to move sketches that contain conflicting constraints
+    if (hasConflicts()) {
+        isInitMove = false;
+        return -1;
+    }
+
+    // this is only meant for B-Splines
+    if (Geoms[geoId].type != BSpline || pos == PointPos::start || pos == PointPos::end) {
+        return -1;
+    }
+
+    // Find the closest knot
+    auto partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId].geo);
+    double uNear;
+    partBsp->closestParameter(firstPoint, uNear);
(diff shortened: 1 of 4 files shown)


AjinkyaDahale:
Marked as ready for review since it is working for periodic b-splines now.

abdullahtahiriyo:
It looks really nice!! 

I hope to have some time over the weekend to review the code.

abdullahtahiriyo:
Very nice. Thank you. Merged.

AjinkyaDahale:
> Very nice. Thank you. Merged.

Glad you liked it. Please take a look at [the forum post on this](https://forum.freecadweb.org/viewtopic.php?f=9&t=69882&p=606360#p606360), because I suspect there would be people who would want it the old way.

abdullahtahiriyo:
@AjinkyaDahale 

I am not sure if this renders impossible block dragging (by selection the poles). But, we could think of having both, maybe with a modifier key. 

Honestly, the functionality was so nice and useful that it is a clear improvement. If there is something we left behind, then we ask users for ideas on how to best get it back.

AjinkyaDahale:
> I am not sure if this renders impossible block dragging (by selection the poles).

What is "block dragging"? I tried to select multiple elements and drag them together, but that didn't do anything.

abdullahtahiriyo:
> What is "block dragging"? I tried to select multiple elements and drag them together, but that didn't do anything.

You are right. It is not possible anymore. 

We should probably have a way to drag as before ("as a block"). You may want to open a discussion on this topic in the forum. Let's see which ideas pop up.

AjinkyaDahale:
> > What is "block dragging"? I tried to select multiple elements and drag them together, but that didn't do anything.
> 
> You are right. It is not possible anymore.

So was this possible before? Looking at the code, it doesn't look as straightforward to make multi-item drag possible. I may be wrong here.

> We should probably have a way to drag as before ("as a block"). You may want to open a discussion on this topic in the forum. Let's see which ideas pop up.

I had mentioned it in the forum post, but it appears it didn't have much traction. Probably something specific for multi-drag would help.

AjinkyaDahale:
@abdullahtahiriyo 

> You may want to open a discussion on this topic in the forum. Let's see which ideas pop up.

See https://forum.freecadweb.org/viewtopic.php?f=8&t=70046.

PR #7124, opened 2022-07-01
Gui: QuantitySpinBox parser changed to Expression Parser

ConnorWorrell:
This pull request changes the parser used in the QuantitySpinBox for non expression dimension input to run through the expression parser.

This fixes issues with these inputs in sketcher datum edit box:
1mm + 1in => 1mm expected 26.4mm
1mm + 1mm => 1mm expected 2mm

I believe all functionality is maintained except:
1 + 1mm => Fails expected 2mm
2 - 1mm => Fails expected 1mm

These get caught by the exceptions:
Base::UnitsMismatchError("Quantity::operator +(): Unit mismatch in plus operation")
Base::UnitsMismatchError("Quantity::operator -(): Unit mismatch in minus operation")

I think this should be handled by the parser, I have a second spicier pull request to src/Base/Quantity.cpp to allow for this functionality. If this is accepted, I'll submit that.

Brief thread I made about this on the forum here:
[Dimension Parser Help](https://forum.freecadweb.org/viewtopic.php?f=8&t=69409)


luzpaz:
Thanks for the PR and your first code contribution to FreeCAD!  
Please indent your code with spaces not tabs.

ConnorWorrell:
Thanks for looking at this @luzpaz,
Tabs are no more.

adrianinsaval:
> 1 + 1mm => Fails expected 2mm

IMO this is not an issue, it's ok for the parser to have rules, specifying units for operands is ok IMO

How does this deal with object references? IMO it shouldn't parse stuff like `Sketch.Constraints[1]` as it may give the wrong impression of being a parametric relation when in fact it isn't.

ConnorWorrell:
> > 1 + 1mm => Fails expected 2mm
> 
> IMO this is not an issue, it's ok for the parser to have rules, specifying units for operands is ok IMO

Yes, but this is a regression from before, I could fix it here or in the parser. And currently I think allowing parser to assume this is a better option. But I could pre parse it before it goes to the expression parser, but that requires keeping track of two parsers. (If anything other than this is required for a functional input, I think pre parsing is better)

> How does this deal with object references? IMO it shouldn't parse stuff like `Sketch.Constraints[1]` as it may give the wrong impression of being a parametric relation when in fact it isn't.

Currently it evaluates any valid reference, Sketch.Constraints[1] would be valid. I think there are enough context clues for this to be intuitive. I could see someone using this to easily pull values from a spreadsheet into a dimension when there is no value in parametrically tying the two together Eg: Stock thickness, Typical drill sizing for slots.

It could be possible to automatically move this input to the expression dialog box much like how "=" works. Although I think this is a downgrade.

ConnorWorrell:
I noticed a few more regressions that occur when the default unit is not the parser unit that couldn't be solved with my proposed change to the expression parser. I added a short pre-parser to help with these, it also fixes this case:
>1 + 1mm => Fails expected 2mm 
>2 - 1mm => Fails expected 1mm

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7124">pipeline status</a> for feature branch [PR_7124](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7124). Pipeline [578984807](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/578984807) was triggered at [c17cc9e](https://github.com/FreeCAD/FreeCAD/pull/7124/commits/c17cc9e1b77da4d34da2f244178592798ae21528). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7124">pipeline status</a> for feature branch [PR_7124](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7124). Pipeline [588569487](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/588569487) was triggered at [cb65672](https://github.com/FreeCAD/FreeCAD/pull/7124/commits/cb65672ced0ed7518def94dcd18ecbfea5e3fff5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
Merged as 7cd01505fc

webmite (commented):
Line 140 of QuantitySpinBox.cpp introduces the use of 'eval()' function for user input.

This should be discouraged.  It is bad practice IMHO.

wwmayer:
> Line 140 of QuantitySpinBox.cpp introduces the use of 'eval()' function for user input.

This is a member function of the class Expression and is NOT supposed to execute arbitrary code like Python's eval() function.

webmite:
Thank you for the clarification.

PR #7128, opened 2022-07-03
Vector2d:fix rotate bug

SuLouOrange:
Base:Fix a vector2d Rotate function bug which use the intermediate result(x) as the basis for the next calculation(y).


SuLouOrange:
[https://en.wikipedia.org/wiki/Rotation_matrix](url)

SuLouOrange:
https://en.wikipedia.org/wiki/Rotation_matrix

donovaly:
Many thanks!

PR #7136, opened 2022-07-05
Prepend CMAKE_INSTALL_PREFIX if CMAKE_INSTALL_LIBDIR is not absolute

adrianinsaval:
This solves an issue where `import freecad` does not work if you don't set CMAKE_INSTALL_LIBDIR to an absolute path during compilation.

This should simplify packaging a little, not sure if this is the correct way of doing this but it works for me. Can someone explain what is done in the windows portion of this?

### DIFF
--- a/src/Ext/freecad/CMakeLists.txt
+++ b/src/Ext/freecad/CMakeLists.txt
@@ -10,7 +10,8 @@ if (WIN32)
                        REALPATH BASE_DIR "${CMAKE_INSTALL_PREFIX}")
     set( ${CMAKE_INSTALL_BINDIR})
 else()
-    set(FREECAD_LIBRARY_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR})
+    get_filename_component(FREECAD_LIBRARY_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}"
+                       REALPATH BASE_DIR "${CMAKE_INSTALL_PREFIX}")
 endif()
 
 configure_file(__init__.py.template ${NAMESPACE_INIT})


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7136">pipeline status</a> for feature branch [PR_7136](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7136). Pipeline [580672546](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/580672546) was triggered at [11c280b](https://github.com/FreeCAD/FreeCAD/pull/7136/commits/11c280bc675aa498a2fb6e29ea61cb9c065d8bc9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Ext/freecad/CMakeLists.txt
```diff
@@ -10,7 +10,11 @@ if (WIN32)
                        REALPATH BASE_DIR "${CMAKE_INSTALL_PREFIX}")
```
adrianinsaval, line 10: actually looking at https://cmake.org/cmake/help/latest/command/get_filename_component.html it seems the windows portion is already doing what I want, what's the reason for having this only for windows?

src/Ext/freecad/CMakeLists.txt
```diff
@@ -10,7 +10,11 @@ if (WIN32)
...
     set( ${CMAKE_INSTALL_BINDIR})
```
adrianinsaval, line 11: can someone explain what this does?

src/Ext/freecad/CMakeLists.txt
adrianinsaval: ok, it's not the same actually since it's using CMAKE_INSTALL_BINDIR, can someone explain why BINDIR is used for windows?

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7136">pipeline status</a> for feature branch [PR_7136](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7136). Pipeline [580697069](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/580697069) was triggered at [e440c42](https://github.com/FreeCAD/FreeCAD/pull/7136/commits/e440c42f4667b3f51cc2491f3c991ec40fb8d0e7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Ext/freecad/CMakeLists.txt
looooo: sorry I can't remember.

src/Ext/freecad/CMakeLists.txt
donovaly: This sets the path to the install folder of CMake. In my case to
`C:\Program Files\CMake\bin`

I think it is there for the case that users installed CMake without adding it to the global PATH variable. In this case thanks to CMAKE_INSTALL_BINDIR, the CMake executable is found nevertheless. Therefore It doesn't harm to keep it.

donovaly:
Thanks for the PR.

Can you please
- keep CMAKE_INSTALL_BINDIR for now
- rebase
- merge the 2 commits into one

then I will merge it.

src/Ext/freecad/CMakeLists.txt
adrianinsaval: I don't think it's related to the cmake installation, `CMAKE_INSTALL_BINDIR` is where the compiled binaries will get installed, in my case `/usr/lib/freecad/bin`
Looking at the [cmake documentation for the set command](https://cmake.org/cmake/help/latest/command/set.html) this line makes no sense, when you provide a sinlge argument it is supposed to be a variable identifier that would be unset, if you pass `${CMAKE_INSTALL_BINDIR}` I think it gets evaluated so it is actually receiving just a path as argument, which makes no sense. Plus I don't see why you would want to unset `CMAKE_INSTALL_BINDIR`

adrianinsaval:
will do tomorrow

adrianinsaval:
Done, the windows portion is weird, some day I'll set a conda env on windows and test it out.

adrianinsaval:
@donovaly could this be backported too? IMO it counts as a fix as it fixes running `import freecad` in python when freecad is compiled without specifying an absolute path for `CMAKE_INSTALL_LIBDIR`

donovaly:
Backported.

adrianinsaval:
Thanks!

PR #7160, opened 2022-07-10
[FEM] fix solver console bugs

donovaly:
- for Elmer and Z88 on Windows several window popup. This is maybe annoying and the user is wondering what is going on, but the main problem is that when you close them you break the solving process.
Therefore, on Windows only, hide the empty popup windows

### DIFF
--- a/src/Mod/Fem/femsolver/elmer/tasks.py
+++ b/src/Mod/Fem/femsolver/elmer/tasks.py
@@ -121,10 +121,18 @@ class Solve(run.Solve):
                 if os.path.isdir(solvpath):
                     os.environ["ELMER_HOME"] = solvpath
                     os.environ["LD_LIBRARY_PATH"] = "$LD_LIBRARY_PATH:{}/modules".format(solvpath)
-            self._process = subprocess.Popen(
+            # hide the popups on Windows
+            if system() == "Windows":
+                self._process = subprocess.Popen(
                 [binary], cwd=self.directory,
                 stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE)
+                stderr=subprocess.PIPE,
+                startupinfo=femutils.startProgramInfo("hide"))
+            else:
+                self._process = subprocess.Popen(
+                    [binary], cwd=self.directory,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE)
             self.signalAbort.add(self._process.terminate)
             output = self._observeSolver(self._process)
             self._process.communicate()
--- a/src/Mod/Fem/femsolver/elmer/writer.py
+++ b/src/Mod/Fem/femsolver/elmer/writer.py
@@ -33,6 +33,7 @@ import os
 import os.path
 import subprocess
 import tempfile
+from platform import system
 
 from FreeCAD import Console
 from FreeCAD import Units
@@ -216,7 +217,12 @@ class Writer(object):
                     unvPath,
                     "-scale", "0.001", "0.001", "0.001",
                     "-out", self.directory]
-            subprocess.call(args, stdout=subprocess.DEVNULL)
+            # hide the popups on Windows
+            if system() == "Windows":
+                subprocess.call(args, stdout=subprocess.DEVNULL,
+                    startupinfo=femutils.startProgramInfo("hide"))
+            else:
+                subprocess.call(args, stdout=subprocess.DEVNULL)
 
     def _writeStartinfo(self):
         path = os.path.join(self.directory, _STARTINFO_NAME)
--- a/src/Mod/Fem/femsolver/z88/tasks.py
+++ b/src/Mod/Fem/femsolver/z88/tasks.py
@@ -31,6 +31,7 @@ __url__ = "https://www.freecadweb.org"
 import os
 import os.path
 import subprocess
+from platform import system
 
 import FreeCAD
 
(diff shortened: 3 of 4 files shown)


donovaly:
@chennes , since I am no Pythionist, could you please check my coding here? It works as it should on Windows and the aim is that for non-Windows the PR changes nothing.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160">pipeline status</a> for feature branch [PR_7160](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160). Pipeline [584460870](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/584460870) was triggered at [f5ed423](https://github.com/FreeCAD/FreeCAD/pull/7160/commits/f5ed4235b48574604d75bec1712e031dd78e2d15). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Fem/femsolver/elmer/tasks.py
```diff
@@ -121,10 +121,18 @@ def run(self):
...
+            if system() == "Windows":
+                self._process = subprocess.Popen(
                 [binary], cwd=self.directory,
                 stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE)
+                stderr=subprocess.PIPE,
+                startupinfo=femutils.startProgramInfo("hide"))
+            else:
+                self._process = subprocess.Popen(
+                    [binary], cwd=self.directory,
```
luzpaz, line 133: Trailing whitespace

src/Mod/Fem/femsolver/elmer/writer.py
```diff
@@ -216,7 +217,13 @@ def _writeMesh(self):
                     unvPath,
                     "-scale", "0.001", "0.001", "0.001",
                     "-out", self.directory]
-            subprocess.call(args, stdout=subprocess.DEVNULL)
+            args.extend(["-out", self.directory])
+            # hide the popups on Windows
+            if system() == "Windows":
+                subprocess.call(args, stdout=subprocess.DEVNULL,
```
luzpaz, line 223: trailing whitespace

luzpaz (changes requested):

donovaly:
Thanks Luzpaz, I fixed the whitespace issues and hopefully also the encoding issue that break the CI tests.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160">pipeline status</a> for feature branch [PR_7160](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160). Pipeline [585272123](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/585272123) was triggered at [c76c9d5](https://github.com/FreeCAD/FreeCAD/pull/7160/commits/c76c9d560b1d691ff48c461700b74cb7fb1c9170). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Fem/femsolver/z88/tasks.py
```diff
@@ -117,23 +118,29 @@ def run(self):
...
-        self._process = subprocess.Popen(
-            [binary, "-c", "-" + solver],
-            cwd=self.directory,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE)
+        # starting normal because the user must see the z88 window
+        Solve.runZ88("-c", binary, solver, "normal", self)
+
+
+    def runZ88(command, binary, solver, state, self):
```
benj5378, line 129:
```suggestion
    def runZ88(self, command, binary, solver, state):
```
If this hasn't caused you any unexpected results, then the code hasn't been tested properly. (This would always use the wrong object as self.)

benj5378 (changes requested):

benj5378:
Also, can you link a forum thread? (What is those pop ups, what is written in them?)

src/Mod/Fem/femsolver/z88/tasks.py
donovaly: I am no Pythionist. Why does the order matter here?
I made your change now and it worked as well as before but would like to understand why you want self to be as first argument.

donovaly:
> Also, can you link a forum thread? (What is those pop ups, what is written in them?)

There is no forum thread. The popups occur because programs are started in a console window. So an empty console pops up. This is confusing for users. However, as you can see I only hide the empty popups not the ones the user should see.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160">pipeline status</a> for feature branch [PR_7160](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160). Pipeline [585411960](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/585411960) was triggered at [7d03a0c](https://github.com/FreeCAD/FreeCAD/pull/7160/commits/7d03a0ccbbb097c553aaf5c447ad05a64d403294). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160">pipeline status</a> for feature branch [PR_7160](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160). Pipeline [585421922](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/585421922) was triggered at [d45bc5a](https://github.com/FreeCAD/FreeCAD/pull/7160/commits/d45bc5abddb0773fe53aa6758eaec191e6aaaa7c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160">pipeline status</a> for feature branch [PR_7160](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7160). Pipeline [585422746](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/585422746) was triggered at [81a85bf](https://github.com/FreeCAD/FreeCAD/pull/7160/commits/81a85bfd902993198bd9fc756d9a357ec81d142f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Fem/femsolver/z88/tasks.py
chennes: I suggest making two changes, actually: as currently written, `runZ88` is being called as a C++ programmer would think of as a _static_ member function, with an additional manually-specified argument specifying an instance to work on (so why is the functions static?!) As originally written, then, the order really doesn't matter because `self` is being passed in completely manually. Calling it like this is _very_ non-idiomatic -- I cannot figure out why it's being done, there seems to be no particular advantage. In my opinion the call itself should be
```
self.runZ88("-c", binary, solver, "normal")
```
which would the necessitate the change suggested by @benj5378, since in this format `self` is passed implicitly to the function as its first argument. Is there some reason to use the `Solve.runZ88("-c", binary, solver, "normal", self)` call format?

donovaly: As I wrote I already adopted  the PR. I understand now that I just need "self" of the class, and as first argument, this is just passed.
I need/use the class' "self" in runZ88() for self._process

benj5378: @donovaly When creating a class, the class itself is always passed as the first parameter which we usually call self. Normally, this happens implicitly as the first argument. So when you write command as the first parameter, then this actually gets passed the class itself while self ends up being the state. `self._process` is expected to be "interpreted" as `Solve._process`, but since command was the first parameter, `command` is in reality the class. And because self was given implicitly, all the parameters shift so with `self.runZ88("-c", binary, solver, "normal")`, `self` would be assigned the string `"normal"`.

Or, this is how it normally goes on. @chennes writes that self is parsed manually so it doesn't have any meaning here - why I don't understand, but I believe he knows what he is talking about.

donovaly: I still don't understand what the problem is. I changed already the argument call order as you requested.

(I googled a bit around and it seems that also my initial PR is valid, only unusal. I also debugged it and the arguments were correctly passed.)

I merged the PR to move on. Feels free to improve the code further.

PR #7182, opened 2022-07-13
Fix Snap launch of OpenSCAD

chennes:
Patch provided by forums user 'lamikr': https://forum.freecadweb.org/viewtopic.php?f=10&t=40848

Note that I have not yet tested this patch, I am just submitting on behalf of its author.

### DIFF
--- a/src/Mod/OpenSCAD/OpenSCADUtils.py
+++ b/src/Mod/OpenSCAD/OpenSCADUtils.py
@@ -142,12 +142,31 @@ def getopenscadversion(osfilename=None):
             "User parameter:BaseApp/Preferences/Mod/OpenSCAD").\
             GetString('openscadexecutable')
     if osfilename and os.path.isfile(osfilename):
-        with subprocess.Popen([osfilename, '-v'],\
-            stdout = subprocess.PIPE,stderr=subprocess.PIPE, universal_newlines=True) as p:
+        env = os.environ.copy()
+        try:
+            env.pop('LD_LIBRARY_PATH')
+        except KeyError:
+            None
+        def check_output2(*args, **kwargs):
+            env = os.environ.copy()
+            try:
+                env.pop('LD_LIBRARY_PATH')
+            except KeyError:
+                None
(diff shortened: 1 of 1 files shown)


src/Mod/OpenSCAD/OpenSCADUtils.py
```diff
@@ -142,12 +142,31 @@ def getopenscadversion(osfilename=None):
...
+        try:
+            env.pop('LD_LIBRARY_PATH')
+        except KeyError:
+            None
+        def check_output2(*args, **kwargs):
+            env = os.environ.copy()
+            try:
+                env.pop('LD_LIBRARY_PATH')
+            except KeyError:
+                None
```
benj5378, line 155: Aren't line 151-155 duplicate of 146-150? The commands are already run once?

luzpaz:
Is this Snap as in the Snap packaging ?

adrianinsaval:
Also, is this something that we want for all distributions of FreeCAD? IMO this should be managed in the snap or at the very least be inside an if statement for the snap

adrianinsaval:
Looking at the previous discussion this was actually a fix for the appimage, is this relevant for the snap too as stated in the title? IMO this should preferably be a configurable option that is patched to default to true for the appimage/snap/flatpak? but defaults to false normally.

luzpaz:
Please advise

chennes:
There is another ongoing forums discussion with a related, though more comprehensive, fix: https://forum.freecadweb.org/viewtopic.php?f=10&t=70790

luzpaz:
Is there TL;DR summary somewhere ?

luzpaz:
bump

chennes:
No, this just fell off my radar. I don't think the code in this PR is quite right, it probably makes the most sense to just close it. We could open an issue if there isn't one already.

luzpaz:
@chennes please advise what to name ticket to replace this PR ?

chennes:
I guess just something like "Snap launch of OpenSCAD broken"

adrianinsaval:
IIRC appimages had this problem too

PR #7152, opened 2022-07-09
Github: basic check CI Run

Floriansimmer:
### This pull request adds a GitHub action that performs various checks for a pull request:
- check for trailing whitespaces (fixes #5731)
- check for uses of tabs 
- check for string based QT connections (#6166)
- codespell --> with latest dictionaries
#### static code analyzer:
- cpplint ( https://github.com/cpplint/cpplint) 
- pylint (https://pylint.pycqa.org/en/latest/)
- clang-tidy (https://clang.llvm.org/extra/clang-tidy/)
- clazy + separate checks for QT6 porting (https://github.com/KDE/clazy) (#7134)
#### code-formatter :
- black (https://github.com/psf/black)
- clang-format (https://clang.llvm.org/docs/ClangFormat.html)
#### after making the checks
- ~~it reports the logs as sticky-comment (see comment below)~~ not working because of insufficient permissions of the GITHUB_TOKEN ( [Token Settings](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository) ), a workaround would be to run a second workflow in the scope of the repo [example](https://github.com/marocchino/sticky-pull-request-comment/issues/298#issuecomment-836278838) 
- adds annotations in the file's changed tab
- export the logs and automated fixes (which can be applied with `clang-apply-replacements`) as a build artifact that can be downloaded
- most of the checks are configurable with variables on top of the workflow file
#### Notes:
- some logs get shorten to 500 lines in the report because GitHub has a limit of 65000 characters per comment
- it only makes 10 annotations per step / 50 in total --> GitHub limitation
- the old codespell action can be removed / disabled after merging this PR

#### Now the Questions:
- Which checks should be enabled / filtered? Especially the static code analyzers produce many warnings, most of them are best practice, modernization or style related.
- should we make a `.clang-tidy`  file instead of hard-coding the clang-tidy checks?
- should cpplint be enabled? --> is mostly to check C/C++ files for style issues following Google's C++ style guide.

Do you have any suggestions? 
Or additional checks or tools you would like to see added?

PR #7193, opened 2022-07-15
PD: more options for revolution/groove

### LINKED ISSUE #7179: [Feature Request] PD: add more options in revolve/groove feature (two angles, up to face/first/last)

### Forums discussion

[Forum thread](https://forum.freecadweb.org/viewtopic.php?p=609439#p609439)

### Subproject(s) affected?

PartDesign

### Idea description

it's useful for the design adding some options in revolve/groove feature like in pad/pocket one. I mean:

- Two directions (two different angles from revolving axis)
- up to face/first/last

### Anything else?

_No response_

### Code of Conduct

AjinkyaDahale:
Attempts to fix #7179.

I am trying to copy the code from pad/pocket but they are somewhat different.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureRevolution.cpp
+++ b/src/Mod/PartDesign/App/FeatureRevolution.cpp
@@ -92,10 +92,6 @@ App::DocumentObjectExecReturn *Revolution::execute()
 
     double angle2 = Base::toRadians(Angle2.getValue());
 
-    // Reverse angle if selected
-    if (Reversed.getValue() && !Midplane.getValue())
-        angle *= (-1.0);
-
     TopoDS_Shape sketchshape;
     try {
         sketchshape = getVerifiedFace();
@@ -131,22 +127,6 @@ App::DocumentObjectExecReturn *Revolution::execute()
 
         std::string method(Type.getValueAsString());
 
-        // Rotate the face by half the angle to get Revolution symmetric to sketch plane
-        if (Midplane.getValue()) {
-            gp_Trsf mov;
-            mov.SetRotation(gp_Ax1(pnt, dir), Base::toRadians<double>(Angle.getValue()) * (-1.0) / 2.0);
-            TopLoc_Location loc(mov);
-            sketchshape.Move(loc);
-        }
-        else if (method == "TwoAngles") {
-            gp_Trsf mov;
-            mov.SetRotation(gp_Ax1(pnt, dir), angle2 * (-1.0));
-            TopLoc_Location loc(mov);
-            sketchshape.Move(loc);
-
-            angle = angle + angle2;
-        }
-
         this->positionByPrevious();
         TopLoc_Location invObjLoc = this->getLocation().Inverted();
         pnt.Transform(invObjLoc.Transformation());
@@ -162,6 +142,7 @@ App::DocumentObjectExecReturn *Revolution::execute()
                 return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Revolve axis intersects the sketch"));
         }
 
+        // Create a fresh support even when base exists so that it can be used for patterns
         TopoDS_Shape result;
         TopoDS_Face supportface = getSupportFace();
         supportface.Move(invObjLoc);
@@ -175,43 +156,25 @@ App::DocumentObjectExecReturn *Revolution::execute()
             else
                 throw Base::RuntimeError("ProfileBased: Revolution up to first/last is not yet supported");
 
+            // TODO: This method is designed for extrusions. needs to be adapted for revolutions.
             getUpToFace(upToFace, base, sketchshape, method, dir);
 
-            // TODO: Make enum
-            int mode = 2;
-            BRepFeat_MakeRevol RevolMaker;
-            for (TopExp_Explorer xp(sketchshape, TopAbs_FACE); xp.More(); xp.Next()) {
-                RevolMaker.Init(base, xp.Current(), supportface, gp_Ax1(pnt, dir), mode, Standard_True);
-                RevolMaker.Perform(upToFace);
+            TopoDS_Face supportface = getSupportFace();
+            supportface.Move(invObjLoc);
 
-                if (!RevolMaker.IsDone())
-                    throw Base::RuntimeError("ProfileBased: Up to face: Could not revolve the sketch!");
+            if (Reversed.getValue())
+                dir.Reverse();
 
-                base = RevolMaker.Shape();
-                if (mode == 2)
-                    mode = 1;
-            }
-
-            result = base;
+            TopExp_Explorer Ex(supportface,TopAbs_WIRE);
+            if (!Ex.More())
+                supportface = TopoDS_Face();
+            RevolMode mode = RevolMode::None;
+            generateRevolution(result, method, base, sketchshape, supportface, upToFace, gp_Ax1(pnt, dir), mode, Standard_True);
         }
         else {
-            // TODO: Make enum
-            int mode = 2;
-            BRepFeat_MakeRevol RevolMaker;
-            for (TopExp_Explorer xp(sketchshape, TopAbs_FACE); xp.More(); xp.Next()) {
-                RevolMaker.Init(base, xp.Current(), supportface, gp_Ax1(pnt, dir), mode, Standard_True);
-                RevolMaker.Perform(angle);
-
-                if (!RevolMaker.IsDone())
-                    throw Base::RuntimeError("ProfileBased: Could not revolve the sketch!");
-
-                base = RevolMaker.Shape();
-                if (mode == 2)
-                    mode = 1;
-            }
-
-            if (RevolMaker.IsDone())
-                result = RevolMaker.Shape();
+            bool midplane = Midplane.getValue();
+            bool reversed = Reversed.getValue();
+            generateRevolution(result, sketchshape, method, gp_Ax1(pnt, dir), angle, angle2, midplane, reversed);
         }
 
         if (!result.IsNull()) {
@@ -270,6 +233,97 @@ void Revolution::updateAxis()
 
     Base.setValue(base.x,base.y,base.z);
     Axis.setValue(dir.x,dir.y,dir.z);
+}
+
+void Revolution::generateRevolution(TopoDS_Shape& revol,
+                                    const TopoDS_Shape& sketchshape,
+                                    const std::string& method,
+                                    const gp_Ax1& axis,
+                                    const double angle,
+                                    const double angle2,
+                                    const bool midplane,
+                                    const bool reversed)
+{
+    if (method == "Angle" || method == "TwoAngles" || method == "ThroughAll") {
+    double angleTotal = angle;
+    double angleOffset = 0.;
+
+    if (method == "TwoAngles") {
+        // Rotate the face by `angle2`/`angle` to get "second" angle
+        angleTotal += angle2;
+        angleOffset = angle2 * -1.0;
+    }
+    else if (midplane) {
+        // Rotate the face by half the angle to get Revolution symmetric to sketch plane
+        angleOffset = -angle / 2;
+    }
+
+    if (fabs(angleTotal) < Precision::Angular())
+        throw Base::ValueError("Cannot create a revolution with zero angle.");
+
+    gp_Ax1 revolAx(axis);
+    if (reversed) {
+        revolAx.Reverse();
+    }
+
+    TopoDS_Shape from = sketchshape;
+    if (method == "TwoAngles" || midplane) {
+        gp_Trsf mov;
+        mov.SetRotation(revolAx, angleOffset);
+        TopLoc_Location loc(mov);
+        from.Move(loc);
+    }
+
+    // revolve the face to a solid
+    // BRepPrimAPI is the only option that allows use of this shape for patterns.
+    // See https://forum.freecadweb.org/viewtopic.php?f=8&t=70185&p=611673#p611673.
+    BRepPrimAPI_MakeRevol RevolMaker(from, revolAx, angleTotal);
+
+    if (!RevolMaker.IsDone())
+        throw Base::RuntimeError("ProfileBased: RevolMaker failed! Could not revolve the sketch!");
+    else
+        revol = RevolMaker.Shape();
+    }
+    else {
+        std::stringstream str;
+        str << "ProfileBased: Internal error: Unknown method '"
+            << method << "' for generateRevolution()";
+        throw Base::RuntimeError(str.str());
+    }
+}
+
+void Revolution::generateRevolution(TopoDS_Shape& revol,
+                                    const std::string& method,
+                                    const TopoDS_Shape& baseshape,
+                                    const TopoDS_Shape& profileshape,
+                                    const TopoDS_Face& supportface,
+                                    const TopoDS_Face& uptoface,
+                                    const gp_Ax1& axis,
+                                    RevolMode Mode,
+                                    Standard_Boolean Modify)
+{
+    if (method == "UpToFirst" || method == "UpToFace" || method == "UpToLast") {
+        BRepFeat_MakeRevol RevolMaker;
+        TopoDS_Shape base = baseshape;
+        for (TopExp_Explorer xp(profileshape, TopAbs_FACE); xp.More(); xp.Next()) {
+            RevolMaker.Init(base, xp.Current(), supportface, axis, Mode, Modify);
+            RevolMaker.Perform(uptoface);
+            if (!RevolMaker.IsDone())
+                throw Base::RuntimeError("ProfileBased: Up to face: Could not revolve the sketch!");
+
+            base = RevolMaker.Shape();
+            if (Mode == RevolMode::None)
+                Mode = RevolMode::FuseWithBase;
+        }
+
+        revol = base;
+    }
+    else {
+        std::stringstream str;
+        str << "ProfileBased: Internal error: Unknown method '"
+            << method << "' for generateRevolution()";
+        throw Base::RuntimeError(str.str());
+    }
 }
 
 }
--- a/src/Mod/PartDesign/App/FeatureRevolution.cpp
+++ b/src/Mod/PartDesign/App/FeatureRevolution.cpp
@@ -81,13 +81,16 @@ short Revolution::mustExecute() const
 App::DocumentObjectExecReturn *Revolution::execute()
 {
     // Validate parameters
-    double angle = Angle.getValue();
-    if (angle > 360.0)
+    // All angles are in radians unless explicitly stated
+    double angleDeg = Angle.getValue();
+    if (angleDeg > 360.0)
         return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Angle of revolution too large"));
 
-    angle = Base::toRadians<double>(angle);
+    double angle = Base::toRadians<double>(angleDeg);
     if (angle < Precision::Angular())
         return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Angle of revolution too small"));
+
+    double angle2 = Base::toRadians(Angle2.getValue());
 
     // Reverse angle if selected
     if (Reversed.getValue() && !Midplane.getValue())
@@ -126,12 +129,22 @@ App::DocumentObjectExecReturn *Revolution::execute()
         if (sketchshape.IsNull())
             return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Creating a face from sketch failed"));
 
+        std::string method(Type.getValueAsString());
+
         // Rotate the face by half the angle to get Revolution symmetric to sketch plane
         if (Midplane.getValue()) {
             gp_Trsf mov;
             mov.SetRotation(gp_Ax1(pnt, dir), Base::toRadians<double>(Angle.getValue()) * (-1.0) / 2.0);
             TopLoc_Location loc(mov);
             sketchshape.Move(loc);
+        }
+        else if (method == "TwoAngles") {
+            gp_Trsf mov;
+            mov.SetRotation(gp_Ax1(pnt, dir), angle2 * (-1.0));
+            TopLoc_Location loc(mov);
+            sketchshape.Move(loc);
+
+            angle = angle + angle2;
         }
 
         this->positionByPrevious();
@@ -150,17 +163,18 @@ App::DocumentObjectExecReturn *Revolution::execute()
         }
 
         TopoDS_Shape result;
+        TopoDS_Face supportface = getSupportFace();
+        supportface.Move(invObjLoc);
 
-        std::string method(Type.getValueAsString());
-        if (method == "UpToFace") {
-            TopoDS_Face supportface = getSupportFace();
-            supportface.Move(invObjLoc);
-
+        if (method == "UpToFace" || method == "UpToFirst" || method == "UpToLast") {
             TopoDS_Face upToFace;
             if (method == "UpToFace") {
                 getFaceFromLinkSub(upToFace, UpToFace);
                 upToFace.Move(invObjLoc);
             }
+            else
+                throw Base::RuntimeError("ProfileBased: Revolution up to first/last is not yet supported");
+
             getUpToFace(upToFace, base, sketchshape, method, dir);
 
             // TODO: Make enum
@@ -181,8 +195,20 @@ App::DocumentObjectExecReturn *Revolution::execute()
             result = base;
         }
         else {
-            // revolve the face to a solid
-            BRepPrimAPI_MakeRevol RevolMaker(sketchshape, gp_Ax1(pnt, dir), angle);
+            // TODO: Make enum
+            int mode = 2;
+            BRepFeat_MakeRevol RevolMaker;
+            for (TopExp_Explorer xp(sketchshape, TopAbs_FACE); xp.More(); xp.Next()) {
+                RevolMaker.Init(base, xp.Current(), supportface, gp_Ax1(pnt, dir), mode, Standard_True);
+                RevolMaker.Perform(angle);
+
+                if (!RevolMaker.IsDone())
+                    throw Base::RuntimeError("ProfileBased: Could not revolve the sketch!");
+
+                base = RevolMaker.Shape();
+                if (mode == 2)
+                    mode = 1;
+            }
 
             if (RevolMaker.IsDone())
                 result = RevolMaker.Shape();
--- a/src/Mod/PartDesign/App/FeatureRevolution.cpp
+++ b/src/Mod/PartDesign/App/FeatureRevolution.cpp
@@ -197,6 +197,9 @@ App::DocumentObjectExecReturn *Revolution::execute()
         else
             return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Could not revolve the sketch!"));
 
+        // eventually disable some settings that are not valid for the current method
+        updateProperties(method);
+
         return App::DocumentObject::StdReturn;
     }
     catch (Standard_Failure& e) {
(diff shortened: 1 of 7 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [589171920](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589171920) was triggered at [7afce09](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/7afce09a3f5b2463f49f1d426ffd5c6a02209e83). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [589181900](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589181900) was triggered at [27ca041](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/27ca041731a638f78a8c8714c5538d59727a0539). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [589438502](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589438502) was triggered at [a29735f](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/a29735f7b9b37fc090ac988f544e3f649a345015). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [589471191](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589471191) was triggered at [0afa79f](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/0afa79f5c5853249cefa39b92fba19e3c71a0be3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [589503164](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589503164) was triggered at [5da34f0](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/5da34f00bd4f7bd432595b361334459f3f7c97f9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [589674054](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589674054) was triggered at [04acc91](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/04acc91747a90a569af249c0113a66598474624b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [589784824](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589784824) was triggered at [0c256d8](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/0c256d8c5079f8b3f1258f88d6dab963c88c5d45). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureRevolution.cpp
```diff
@@ -117,13 +129,23 @@ App::DocumentObjectExecReturn *Revolution::execute(void)
         if (sketchshape.IsNull())
             return new App::DocumentObjectExecReturn("Creating a face from sketch failed");
 
+        std::string method(Type.getValueAsString());
```
wwmayer, line 132: You shouldn't repeat the same flaws from the pad/pocket feature. Instead of a std::string to handle the different methods use an enum class because this is faster and more secure because in case of a typo the compiler will complain.

src/Mod/PartDesign/App/FeatureRevolution.cpp
```diff
@@ -138,16 +160,64 @@ App::DocumentObjectExecReturn *Revolution::execute(void)
         for (;xp.More(); xp.Next()) {
             if (checkLineCrossesFace(gp_Lin(pnt, dir), TopoDS::Face(xp.Current())))
                 return new App::DocumentObjectExecReturn("Revolve axis intersects the sketch");
-        }
```
wwmayer, line 141: The execute() function is by far too long and should be refactored.

src/Mod/PartDesign/App/FeatureRevolution.cpp
AjinkyaDahale: Sounds right. But I notice that the enumeration is also a property that can be changed outside of the task view. Do we sacrifice that or are there other options?

src/Mod/PartDesign/App/FeatureRevolution.cpp
AjinkyaDahale: Possibly yes. It would also help to make a superclass like pad/pocket. Thing is I still can't wrap my head around the exact algorithm. Should try pen and paper first :grimacing:.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [589847111](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589847111) was triggered at [81d67ef](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/81d67ef6f2f0c18f12c2ca568812ed244a86319b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureRevolution.cpp
AjinkyaDahale: @wwmayer do you have a suggestion of what to do about the enum being a property? It's straightforward to make an enum class instead of strings, but I need to know what to do about this being a property. Is there an example where `enum class` was used with `App::PropertyEnumeration`?

src/Mod/PartDesign/App/FeatureRevolution.cpp
AjinkyaDahale: @wwmayer I moved some of the logic to separate methods.

src/Mod/PartDesign/App/FeatureRevolution.cpp
wwmayer: > do you have a suggestion of what to do about the enum being a property? ...

I am not talking about the PropertyEnumeration using an enum. There you still have to use a list of strings. I am talking about the feature's execute() method where you use a string to handle the different modes.
```
std::string method = ... // assigned from property
if (method == "Method1") {
...
}
else if (method == "Method2") {
...
}
```

Now a string comparison is slow and error-prone because if you make a typo and write
```
if (method == "MEthod1") { // should be Method1
...
}
else if (method == "Method2") {
...
}
```
then the compiler won't be able to find the bug. You are only able to find that at runtime. Now my suggestion is to convert the string of the property to an enum and therefore you need a helper function:
```
enum class Method {
    Mehod1,
    Method2
};

Method methodFromString(const std::string& method)
{
    if (method == "Method1")
        return Method::Method1;
    if (method == "Method2")
        return Method::Method2;

    throw Base::ValueError("No such method");
}

void Feature::execute()
{
    Method method = methodFromString(enumProperty.getValue());
    if (method == Method::MEthod1) { // this typo will cause a build failure
...
    }
    else if (method == Method::Method2) {
...
    }
}
```

Of course, you still need a function where you compare strings but it's only one single function and not several functions and thus the likelihood of a typo is much lower.

AjinkyaDahale: Thanks. I could probably reuse these enums in `TaskRevolutionParameters` as well.

AjinkyaDahale: This should be done now. It's all still pretty messy, and groove is now completely broken.

AjinkyaDahale:
So right now, the feature should be working for revolutions with two dimensions and up to face. It's all still pretty messy, and groove is now completely broken.

For up to first/last, I still need to make the method that finds first//last face.

AjinkyaDahale:
CI is failing with the error:

```
Error: /home/runner/work/FreeCAD/FreeCAD/src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp:39:10: error: 'ui_TaskRevolutionParameters.h' file not found [clang-diagnostic-error]
```
Anyone has an idea why?

adrianinsaval:
That's a problem in the CI itself because it's not actually running a full compilation, it just runs some checks (see #7152)
actual compilation CI is the one running on gitlab

AjinkyaDahale:
After doing the work for revolution, the changes for groove came relatively easily. The code is still pretty messy, but it's ready for testing.

adrianinsaval:
do you want a snap build of this? @luzpaz should be able to create one if necessary

AjinkyaDahale:
> do you want a snap build of this? @luzpaz should be able to create one if necessary

It would be fine. However, I expect this will need a good amount of change before merge, so there will be a few iterations of the snap.

adrianinsaval:
I built and tested, the reverse checkbox appears to have no effect when two angles are selected (and possibly other modes). Also having the first angle default to 360 makes the whole thing weird to use as it doesn't behave like pad or pocket, where each side is extruded according to the specified length, here instead the second angle substracts material from the first angle.
 This is a detail but the fields for the 2nd angle and face are visible all the time

AjinkyaDahale:
> I built and tested, the reverse checkbox appears to have no effect when two angles are selected (and possibly other modes). 

Reversed is not supposed to work under certain circumstances (for example when "symmetric to plane" is checked). Not sure if we want that for "two-angle mode", but I see it's available in pad/pocket, so I'll add that. It should still be possible to change the values of both angles though to have the same effect

Also note that "to first/last mode" is not yet supported (though "through all" groove is).

> Also having the first angle default to 360 makes the whole thing weird to use as it doesn't behave like pad or pocket, where each side is extruded according to the specified length, here instead the second angle substracts material from the first angle.

What do you feel is a good default? I just kept the old default, but it won't hurt to change it I believe.

The second angle can be negative, but not sure what the rule is but in the case I'm testing the revolution simply doesn't happen if the total angle (angle + angle2) is greater than 360 degrees.

> This is a detail but the fields for the 2nd angle and face are visible all the time

That is one thing I need to implement.

Apart from these, did you happen to come across any instabilities? Any issues changing the modes/values outside of the task view (like when in comboview or by console)? Do any values suddenly change going in and out of the edit view?

adrianinsaval:
No such issue that I noticed but note that I did really minimal testing, perhaps something like 90 or 180 could be a good default for two angles, the problem is that 360 is actually a good default for single angle and it wouldn't be nice to change that as it's probably the more common mode, is it possible to have different defaults depending on the mode?
Regarding reverse IMO it is a valid and useful feature for all modes except through all, and I'm not sure if we need that mode for revolutions at all since it would be the same as a 360deg angle mode.

AjinkyaDahale:
> ...the problem is that 360 is actually a good default for single angle and it wouldn't be nice to change that as it's probably the more common mode, is it possible to have different defaults depending on the mode?

You hit the nail on the head here. It would indeed be good, but then question is (how) do we track if the default should be applied or an edit has been made. This is relevant because without these smarts changing modes can end up changing values unnecessarily. IMO it's not worth implementing such smarts since that will probably involve storing values twice and keeping them coordinated.

> Regarding reverse IMO it is a valid and useful feature for all modes except through all,

It is, I suppose, depending on how we define "mode". If "midplane" cases are considered different modes it's not valid there.

> and I'm not sure if we need that mode for revolutions at all since it would be the same as a 360deg angle mode.

"Through all" only exists for grooves (subtractive revolve). For additive (just called revolve), it's replaced by "to last", which is actually non-trivial. This distinction I copied from pad/pocket, where making a pocket through the last face is simpler than padding to that last face, but still non-trivial since the length of the pocket depends on the size of the body (we pocket till the end of the bounding box). Technically we don't need it for grooves either, since we can just set the groove to 360. But removing it disturbs the enum commonality between additive and subtractive forms. Probably can be removed with better design patterns.

adrianinsaval:
> It is, I suppose, depending on how we define "mode". If "midplane" cases are considered different modes it's not valid there.

Symmetric does make reverse irrelevant but with modes I was referring to the options from the dropdown, it is relevant for all of them. On the other hand symmetric is only really relevant for single angle mode.

> IMO it's not worth implementing such smarts since that will probably involve storing values twice and keeping them coordinated.

Agreed but perhaps we can reconsider what the second angle means then, because the current behavior feels wonky IMO. Perhaps the second angle should follow the same direction as the first? Or perhaps we should rethinnk how the revolution is performed such that increasing the second angle doesn't remove material when there is overlap with the first angle. My logic is if I tell the program to revolve 300deg on one direction and 70deg on the other direction I should get a full revolution, not just 10deg.

AjinkyaDahale:
> Or perhaps we should rethinnk how the revolution is performed such that increasing the second angle doesn't remove material when there is overlap with the first angle. My logic is if I tell the program to revolve 300deg on one direction and 70deg on the other direction I should get a full revolution, not just 10deg.

The behavior when revolution is beyond 360 depends entirely on OCC. But I could probably add a check before passing the values to `BRep...`. The code is here if you want to suggest a better logic: https://github.com/AjinkyaDahale/FreeCAD/blob/1ad48ed7bf210db347356115fe401449f1c8e8d2/src/Mod/PartDesign/App/FeatureRevolution.cpp#L269

AjinkyaDahale:
@adrianinsaval two-angle reverse should work now. Also options change based on method.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [600422108](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600422108) was triggered at [de06527](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/de06527b6b59613233e081224788c6a4d03dff11). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [607006512](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/607006512) was triggered at [9d58a88](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/9d58a88091f636c98766d699ddd920d30e3deee3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [618200651](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/618200651) was triggered at [2068085](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/2068085752233e1fd614cfdb7bcf058a1cf221cb). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
@AjinkyaDahale do you want a dedicated experimental Snap package so testers can test this PR ?

AjinkyaDahale:
> @AjinkyaDahale do you want a dedicated experimental Snap package so testers can test this PR ?

Yes. But also depends on if testers want it.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [626253360](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/626253360) was triggered at [b5ee2e4](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/b5ee2e4166ec32e84fd13e3e23c90d6366ce18f2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [640021295](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/640021295) was triggered at [4a7c03a](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/4a7c03ad079188b1c0313cdb3b7a4cfd7b4fd046). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@AjinkyaDahale , can you please rebase the PR
@wwmayer , I think Ajinkya did what you requested. Could you please have a final look at this PR?

AjinkyaDahale:
@donovaly I'll do it once I'm back on my computer.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [712716234](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/712716234) was triggered at [7cae2fd](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/7cae2fdd7edfe96156cda4f72645dee1b74298f6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
@donovaly rebased now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [712730035](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/712730035) was triggered at [352db77](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/352db77f6352418888a55f32c52de55374ac79fa). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [712831374](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/712831374) was triggered at [250f8a8](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/250f8a8c53d2c2f29dafe167a15a09be95d16844). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [712857613](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/712857613) was triggered at [08783bb](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/08783bba4d68a6036b4e3cdd589931f3106aa4a7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
Currently we have a design flaw with the feature classes Revolution and Groove. The class interface of both of them is identical and the only difference in their implementations is that the former adds material and the latter removes it. The result is that we have two implementations where 90% of the code is identical.

I think that the two classes should be refactored by creating a common base class with the same class interface.

As consequence all the code duplication can be avoided and the dialog class could be greatly simplified too because it's not needed any more to check for two concrete types.

We once had the very same problem with Pad/Pocket where a common base class was created.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193">pipeline status</a> for feature branch [PR_7193](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7193). Pipeline [713634934](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/713634934) was triggered at [e3e8a53](https://github.com/FreeCAD/FreeCAD/pull/7193/commits/e3e8a53a573246e6a99211b5ce8abe93af09afb5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
@wwmayer I did notice this but did not prioritize combining the classes till now. It's best to make it a new PR, possibly before this. I'm currently working with Abdullah to merge sketcher stuff so I won't have the time this week and possibly the next.

One question that came past me while considering making the common class: what should it be called? Pad and Pocket have a superclass named Extrude, but for Additive Revolve and Groove the "classical" name is "Revolve", but that's already taken.

adrianinsaval:
maybe revolution? is there even a problem in using revolve?

wwmayer:
> but for (additive) Revolve and (subtractive) Groove the "classical" name is "Revolve", but that's already taken.

Is it? Where? The term Revolve I had in mind too but couldn't find that it's already used as class name. Alternatively FeatureRevolve could be used or according to OCC's documentation a suitable term is RevolvedSweep (see BRepPrimAPI_MakeRevol.hxx).

AjinkyaDahale:
> > but for (additive) Revolve and (subtractive) Groove the "classical" name is "Revolve", but that's already taken.
> 
> Is it? Where?

By "classical" I meant in other CAD design tools where both additive and subtractive features are made with the same tool. However, I should add a caveat (should have added earlier itself) that this is from memory and I haven't really used any tools recently.

> The term Revolve I had in mind too but couldn't find that it's already used as class name. Alternatively FeatureRevolve could be used or according to OCC's documentation a suitable term is RevolvedSweep (see BRepPrimAPI_MakeRevol.hxx).

I believe FeatureRevolve might be taken by some other classes. RevolvedSweep might work. Other alternatives are AddSubRevolve or RevolveGroove I guess.

AjinkyaDahale:
> Now this PR has some merge conflicts that should be solved. But I guess we are still in feature freeze, no?

I should be able to address the merge conflicts this week. Still, no need to make a feature freeze exception since AFAIK we're planning to release within July.

AjinkyaDahale:
Tangential question if someone has a quick answer: how to find in git blame when a line was deleted? Case in point: deleting a parameter from `ProfileBased::getUptoFace()`. For this particular case I found it because it was a method used elsewhere: 69aba3abe1f7f6d6192c8f832a25a291d464dbc1.

adrianinsaval:
Using the Two dimensions mode starting from the default value of Angle is still a jarring experience. What can we do about it? How do other CAD deal with this?

AjinkyaDahale:
@adrianinsaval define "jarring". Overall, Part Design is one of the most useful branches we have, but (perhaps because of the usage) also comes as very unpolished.

maxwxyz:
@FlachyJoe you are working on multiple faces as boundary selection in Pad, is this something for you or can be implemented in all additive and subtractive features?

FlachyJoe:
@maxwxyz I don't take a look on Revolution code yet so I don't know. Either way, I'll wait until this PR is merged.

maxwxyz:
great, does this PR need anything or what is holding it back?

AjinkyaDahale:
> great, does this PR need anything or what is holding it back?

@maxwxyz It seems we're waiting for @adrianinsaval's input on the "jarring". Apart from that there's a naming issue which we plan to handle in a different PR.

adrianinsaval:
In two dimensions mode the first angle cannot be set 0, but this is a perfectly valid value in that mode.

There is also a bug with how the task dialog is populated when starting the task on an existing revolution:

https://github.com/FreeCAD/FreeCAD/assets/36372335/5b7bca66-b88d-4b5b-bab6-c37f8bda7d31

it just launches with the options for dimension despite being set to two dimensions.

Besides that I still believe that a default angle of 360° is detrimental for the two dimensions mode and that the behavior when there is overlap between the angles is undesirable/confusing.

So I propose the following changes:

- Get rid of "Dimension" mode as it doesn't provide anything more than "Two Dimensions", present "Two Dimensions" as default mode with `Angle=Angle2=180°` as default values. This means a full revolution is still the default but we don't get an unreasonable default for two dimensions mode. It's also a more useful mode even when you want to make a full revolution as you can play around with the values to control where the seam line will be
- Make it such that whenever `Angle + Angle2 >= 360°` it's a full revolution, not a partial revolution like this:
image
Since these values might get used to control where the seam line will be the `Angle` value should be respected and `Angle2` ignored and use whatever value to make it a full revolution. If possible we could even restrict `Angle2` to have only values up to `360°-Angle`

adrianinsaval:
Are you open to making those changes @AjinkyaDahale ? Or do you want me to take over where you left? I can pick this up next week if you're not available

AjinkyaDahale:
> Are you open to making those changes @AjinkyaDahale ? Or do you want me to take over where you left? I can pick this up next week if you're not available

I'd be taking it over as well since I haven't touched it in a while. But it shouldn't be too complicated a fix. I'll give it a try but might be unavailable for a few days next week (because of Diwali).

adrianinsaval:
it might be a good idea to use shapes instead of faces as limits here as in #11136

maxwxyz:
> it might be a good idea to use shapes instead of faces as limits here as in #11136

both ideally

adrianinsaval:
> > it might be a good idea to use shapes instead of faces as limits here as in #11136
> 
> both ideally

don't worry, a face is also a shape

AjinkyaDahale:
> > > it might be a good idea to use shapes instead of faces as limits here as in #11136
> > 
> > 
> > both ideally
> 
> don't worry, a face is also a shape

I can think of cases where we need to specify a face because up to shape is basically up to first.

adrianinsaval:
> I can think of cases where we need to specify a face because up to shape is basically up to first.

It doesn't have to be the entire shape. The idea is to be able to pass more than one face as limit.

AjinkyaDahale:
> > I can think of cases where we need to specify a face because up to shape is basically up to first.
> 
> It doesn't have to be the entire shape. The idea is to be able to pass more than one face as limit. However it's probably best to wait until that PR gets merged before investing time here.

Oh! So internally it is creating a compound?

adrianinsaval:
Haven't reviewed yet but I think the plan was something like that

sliptonic:
@adrianinsaval will merge after one more review

adrianinsaval:
there are still problems with the UI when opening the task for an existing feature:
image
here two dimensions mode is set but the second dimension field is not shown

AjinkyaDahale:
> the bugs I reported are still there
> 
> ```yaml
> OS: Arch Linux (KDE/plasma)
> Version: 0.22.0dev.35022 +8 (Git)
> Build type: None
> Branch: pd-more-revol-options
> Hash: 131c74045a7e1d1a99f1cb5ee6a47b8bd80f93b8
> Python 3.11.5, Qt 5.15.11, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
> Locale: C/Default (C) [ OS: English/United Kingdom (en_GB) ]
> Installed mods: 
> ```

Here's what I propose:
1) Adding "up to shape" should be treated as a separate issue altogether.
2) If it takes too long (let's say past the Monday merge meeting) to add fixes for https://github.com/FreeCAD/FreeCAD/pull/7193#issuecomment-1795191219, that also be treated as a separate issue. I say this because I have had to fix merge conflicts twice for this PR already, and every time I do it I'm afraid I'm adding even more errors.

AjinkyaDahale:
@adrianinsaval thanks for merging. Have you also created an issue? If not I can.

adrianinsaval:
create the issues please, for the UI stuff I think what's missing is passing the method index to UpdateUI

adrianinsaval:
up to shape is certainly something for later, we are still figuring it out for Pad

chennes:
FYI MSVC compiler users -- this PR introduces an error in the precompiled headers, fixed by this PR:
https://github.com/FreeCAD/FreeCAD/pull/11470

PR #7191, opened 2022-07-15
[PD] Fix button behavior in Pad up-to-face

AjinkyaDahale:
The previous behavior was somewhat erratic when trying to change the end face.

File used for testing attached (needs to be extracted first).
[base-model.zip](https://github.com/FreeCAD/FreeCAD/files/9118068/base-model.zip)

PR #7201, opened 2022-07-17
[PD] improve face selection for Pad/Pocket

donovaly:
- this PR extends PR #7191 as described in this forum thread: https://forum.freecadweb.org/viewtopic.php?f=3&t=70327

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
@@ -220,7 +220,7 @@ void TaskExtrudeParameters::connectSlots()
         this, &TaskExtrudeParameters::onReversedChanged);
     connect(ui->changeMode, qOverload<int>(&QComboBox::currentIndexChanged),
         this, &TaskExtrudeParameters::onModeChanged);
-    connect(ui->buttonFace, &QPushButton::clicked,
+    connect(ui->buttonFace, &QPushButton::toggled,
         this, &TaskExtrudeParameters::onButtonFace);
     connect(ui->lineFaceName, &QLineEdit::textEdited,
         this, &TaskExtrudeParameters::onFaceName);
@@ -255,7 +255,7 @@ void TaskExtrudeParameters::onSelectionChanged(const Gui::SelectionChanges& msg)
                 ui->lineFaceName->setProperty("FeatureName", QByteArray(msg.pObjectName));
                 ui->lineFaceName->setProperty("FaceName", QByteArray(msg.pSubName));
                 // Turn off reference selection mode
-                onButtonFace(false);
+                ui->buttonFace->setChecked(false);
             }
             else {
                 clearFaceName();
@@ -468,7 +468,7 @@ void TaskExtrudeParameters::setCheckboxes(Modes mode, Type type)
         QMetaObject::invokeMethod(ui->lineFaceName, "setFocus", Qt::QueuedConnection);
         // Go into reference selection mode if no face has been selected yet
         if (ui->lineFaceName->property("FeatureName").isNull())
-            onButtonFace(true);
+            ui->buttonFace->setChecked(true);
     }
     else if (mode == Modes::TwoDimensions) {
         isLengthEditVisible = true;
@@ -507,7 +507,7 @@ void TaskExtrudeParameters::setCheckboxes(Modes mode, Type type)
     ui->buttonFace->setEnabled(isFaceEditEnabled);
     ui->lineFaceName->setEnabled(isFaceEditEnabled);
     if (!isFaceEditEnabled) {
-        onButtonFace(false);
+        ui->buttonFace->setChecked(false);
     }
 }
 
@@ -681,16 +681,18 @@ void TaskExtrudeParameters::getReferenceAxis(App::DocumentObject*& obj, std::vec
     }
 }
 
-void TaskExtrudeParameters::onButtonFace(const bool pressed)
+void TaskExtrudeParameters::onButtonFace(const bool checked)
 {
+    if (!checked && ui->lineFaceName->text().isEmpty())
+        ui->lineFaceName->setPlaceholderText(tr("No face selected"));
+    else if (checked && ui->lineFaceName->text().isEmpty())
+        handleLineFaceName(); // sets placeholder text
+
     // to distinguish that this is the direction selection
     selectionFace = true;
 
     // only faces are allowed
-    TaskSketchBasedParameters::onSelectReference(pressed ? AllowSelection::FACE : AllowSelection::NONE);
-
-    // Update button if onButtonFace() is called explicitly
-    ui->buttonFace->setChecked(pressed);
+    TaskSketchBasedParameters::onSelectReference(checked ? AllowSelection::FACE : AllowSelection::NONE);
 }
 
 void TaskExtrudeParameters::onFaceName(const QString& text)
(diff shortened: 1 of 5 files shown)


src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
```diff
@@ -681,16 +681,18 @@ void TaskExtrudeParameters::getReferenceAxis(App::DocumentObject*& obj, std::vec
     }
 }
 
-void TaskExtrudeParameters::onButtonFace(const bool pressed)
+void TaskExtrudeParameters::onButtonFace(const bool checked)
 {
+    if (!checked && ui->lineFaceName->text().isEmpty())
+        ui->lineFaceName->setPlaceholderText(tr("No face selected"));
+    else if (checked && ui->lineFaceName->text().isEmpty())
```
AjinkyaDahale, line 688: Sounds good, but note that it is possible to just type the face "name" here.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7201">pipeline status</a> for feature branch [PR_7201](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7201). Pipeline [589791405](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589791405) was triggered at [9a32ad8](https://github.com/FreeCAD/FreeCAD/pull/7201/commits/9a32ad89f97f0a797695b3e850d62e0ea751437f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskPadParameters.cpp
```diff
@@ -104,6 +104,11 @@ void TaskPadParameters::onModeChanged(int index)
         break;
     case Modes::ToFace:
         pcPad->Type.setValue("UpToFace");
+        // check button if there is no face yet
+        if (ui->lineFaceName->text().isEmpty()) {
+            ui->buttonFace->setChecked(true);
```
AjinkyaDahale, line 109: I suppose this is your main change? If so that's reasonable. As such the button an the text box should work in tandem I believe: while the box is focused the button should be checked and vice versa.

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
```diff
@@ -681,16 +681,18 @@ void TaskExtrudeParameters::getReferenceAxis(App::DocumentObject*& obj, std::vec
     }
 }
 
-void TaskExtrudeParameters::onButtonFace(const bool pressed)
+void TaskExtrudeParameters::onButtonFace(const bool checked)
 {
+    if (!checked && ui->lineFaceName->text().isEmpty())
+        ui->lineFaceName->setPlaceholderText(tr("No face selected"));
+    else if (checked && ui->lineFaceName->text().isEmpty())
+        ui->lineFaceName->setPlaceholderText(tr("Click on a face in the model"));
```
AjinkyaDahale, line 689: This text should be as small as possible. Alternatively, maybe move this to this to the bottom info bar?
Screenshot from 2022-07-17 19-35-28

src/Mod/PartDesign/Gui/TaskPadParameters.cpp
donovaly: Yes. This is just a logic thing to guide the user.

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
donovaly: That depends strongly on the font the user uses, the screen resolution and the width he uses for the task panels.

The user can widen the task panel if he needs to read everything. but you only need to read it once, then you get the idea how the workflow is. So in your screenshot is OK in my opinion.

Concerning the width, when you e.g. add a PD hole you will also have to widens your task panel to see everything and there it is even mandatory otherwise you cannot access all edits.

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
donovaly: You can do this already:
- select a face
- change its name

The solution is simple in my opinion: make the line edit read-only because it is not meant to be that the user changes there something on his own.

I did this now.

AjinkyaDahale: I was just referring to the "select a face" placeholder. Making the linedit readonly is one way to add consistency.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7201">pipeline status</a> for feature branch [PR_7201](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7201). Pipeline [589802708](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589802708) was triggered at [3a0c7d6](https://github.com/FreeCAD/FreeCAD/pull/7201/commits/3a0c7d68ff8eb684b748e2d56dd4df6bb8aa9aaf). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
AjinkyaDahale: This one is fine, but if you can add a tooltip to it that might be good. As such this should be handled in Qt itself I suppose.  I haven't worked with hole, but if it is as described it's a design flaw and it must be changed.

src/Mod/PartDesign/Gui/TaskPadParameters.cpp
wwmayer: In Pad and Pocket you have the exact same code block. The result is that the identical string for translation appears 3x in the .ts files. To fix this and the code duplication move this block to a protected function in TaskExtrudeParameters and call it here (and in Pocket).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7201">pipeline status</a> for feature branch [PR_7201](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7201). Pipeline [589830918](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589830918) was triggered at [51032e9](https://github.com/FreeCAD/FreeCAD/pull/7201/commits/51032e910da220437d585d1baa501649db76f8d0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskPadParameters.cpp
donovaly: Thanks. Done now.

src/Mod/PartDesign/Gui/TaskExtrudeParameters.cpp
adrianinsaval: since you are setting a placeholder text, do you really need to check if the lineedit is empty? once you have text the placeholder text shouldn't be displayed anymore.

donovaly: What do you mean? The placeholder is only shown if there is now text. That is the idea of the placeholder.

adrianinsaval: that's my point, qt already only shows the palceholder when there is no text, so why would you need to check if it's empty? It doesn't harm I guess but I don't understand why the check is there.

donovaly: Ah, this is indeed no longer necessary. I think in my initial PR it was different and later I did not review it thoroughly.
I fixed this now and also simplified the code a bat: https://github.com/FreeCAD/FreeCAD/commit/98dae1b2

PR #7202, opened 2022-07-17
[PD] remove superfluous nullptr checks

donovaly:


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7202">pipeline status</a> for feature branch [PR_7202](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7202). Pipeline [589792012](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/589792012) was triggered at [8caec80](https://github.com/FreeCAD/FreeCAD/pull/7202/commits/8caec807d8635aa30c21c91245ec4e2be2d880d2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
It's all OK.

src/Mod/PartDesign/Gui/ReferenceSelection.cpp
```diff
@@ -340,7 +340,7 @@ bool getReferencedSelection(const App::DocumentObject* thisObj, const Gui::Selec
 
 QString getRefStr(const App::DocumentObject* obj, const std::vector<std::string>& sub)
 {
-    if (obj == nullptr)
+    if (!obj)
```
benj5378, line 343: I'm just curious for some TechDraw refactoring. Is this the same as a nullptr check and why is this preferred?

donovaly: !obj means there is no object this is the same as if obj == nullptr but is seen as not the optimal coding style.

--------
For your TechDraw review, please check for unused includes. TechDraw has a lot of them. I already reviewed the ViewProviders few months back, but the more important part are the files in App and of course the .h files.

benj5378: Hum, how can I know if an include is used or not, without removing them one at a time and checking if the compiler fails? @donovaly

wwmayer: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-if

donovaly: > how can I know if an include is used or not, without removing them one at a time and checking if the compiler fails?

Exactly this way ;-)
But assure that you don't use precompiled headers.

I do it this way:
- search for the name of the include. Fro example for `DocumentObject`
- if I have no hits, I comment the include
- I repeat this for all includes
- from time to time I run the compiler.
- depending on the compiler output I re-activate some.

Time by time one becomes quicker, because for e.g. Exception.h one must also search for "assert".

This is a lot of work, so a project for weeks. 2 weeks ago I did this for the App part of the Part WB and it took me 2 weeks, but was also relaxing because after a stressy day just doing half an hour something repetitive is something I don't have usually during the day.

PR #7216, opened 2022-07-18
[PD Pad/Pocket] Fix bug with midplane usage in TwoLengths mode

adrianinsaval:
properly fix issue reported here: https://forum.freecadweb.org/viewtopic.php?f=3&t=70266

also reverted a previous commit that auto unchecked the midplane option in the task panel since it is now unnecessary with the fix and may result annoying if you want to go back to length mode expecting it to remain checked, also it is inconsistent since it only gets unchecked for two lengths but not the other modes were it is not a valid option either. IMO this was more a workaround than an actual fix for the bug as it could still be encountered through the property view or in old files.

This should be backported to 0.20 branch also

### DIFF
--- a/src/Mod/PartDesign/App/FeatureExtrude.cpp
+++ b/src/Mod/PartDesign/App/FeatureExtrude.cpp
@@ -144,12 +144,9 @@ void FeatureExtrude::generatePrism(TopoDS_Shape& prism,
 
 
         if (method == "TwoLengths") {
-            // midplane makes no sense here
             Ltotal += L2;
             if (reversed)
                 Loffset = -L;
-            else if (midplane)
-                Loffset = -0.5 * (L2 + L);
             else
                 Loffset = -L2;
         }
@@ -277,7 +274,7 @@ void FeatureExtrude::generateTaperedPrism(TopoDS_Shape& prism,
 void FeatureExtrude::updateProperties(const std::string &method)
 {
     // disable settings that are not valid on the current method
-    // disable everything unless we are sure we don't need it
+    // disable everything unless we are sure we need it
     bool isLengthEnabled = false;
     bool isLength2Enabled = false;
     bool isOffsetEnabled = false;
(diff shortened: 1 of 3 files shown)


adrianinsaval:
@donovaly

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7216">pipeline status</a> for feature branch [PR_7216](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7216). Pipeline [590739657](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/590739657) was triggered at [dc80a06](https://github.com/FreeCAD/FreeCAD/pull/7216/commits/dc80a061a7025add9c9a4664a350b036c1c7bfa0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly (changes requested):
Thanks. Handling this in App is indeed better. However, you must set midplane to false nevertheless. With your PR I get this when changing to "two dimensions":

FreeCAD_2OQG4ffhvo

So I see that symmetric is true but since two different lengths are used, it cannot be symmetric.

src/Mod/PartDesign/App/FeatureExtrude.cpp
```diff
@@ -277,7 +274,7 @@ void FeatureExtrude::generateTaperedPrism(TopoDS_Shape& prism,
 void FeatureExtrude::updateProperties(const std::string &method)
 {
     // disable settings that are not valid on the current method
-    // disable everything unless we are sure we don't need it
```
donovaly, line 280: Thanks, I fixed this now in master.

adrianinsaval:
But by that logic we should set midplane to false for everything except length, IMO there is no advantage in doing that, all them ignore it. In any case you may merge only the fix in App and skip the commit I reverted (but then it would be inconsistent for all the other modes)

donovaly:
> But by that logic we should set midplane to false for everything except length, IMO there is no advantage in doing that

That was my initial point but you stated tat the logic is also important for the App part. As it is, is shows symmetric while the model is apparently not symmetric.

Merging since I agree that App is not that important.

adrianinsaval:
The point I tried to make was that in App it should not care about midplane value when doing twolengths so I changed that, I never cared about what value midplane displayed. Thanks for merging! Please backport to 0.20 too.

donovaly:
> Please backport to 0.20 too.

Done.

PR #7219, opened 2022-07-19
[Part] Bugfix Primitives correction of decimal..

Syres916:
...places on creation
See https://forum.freecadweb.org/viewtopic.php?f=3&t=70339

If accepted, could this please be backported to 0.20 branch.

---

### DIFF
--- a/src/Mod/Part/Gui/DlgPrimitives.cpp
+++ b/src/Mod/Part/Gui/DlgPrimitives.cpp
@@ -726,6 +726,7 @@ void DlgPrimitives::on_buttonCircleFromThreePoints_clicked()
 
 QString DlgPrimitives::createPlane(const QString& objectName, const QString& placement) const
 {
+    Base::QuantityFormat format(Base::QuantityFormat::Fixed, Base::UnitsApi::getDecimals());
     return QString::fromLatin1(
         "App.ActiveDocument.addObject(\"Part::Plane\",\"%1\")\n"
         "App.ActiveDocument.%1.Length=%2\n"
@@ -733,15 +734,16 @@ QString DlgPrimitives::createPlane(const QString& objectName, const QString& pla
         "App.ActiveDocument.%1.Placement=%4\n"
         "App.ActiveDocument.%1.Label='%5'\n")
         .arg(objectName)
-        .arg(Base::UnitsApi::toNumber(ui->planeLength->value()))
-        .arg(Base::UnitsApi::toNumber(ui->planeWidth->value()))
+        .arg(Base::UnitsApi::toNumber(ui->planeLength->value(), format))
+        .arg(Base::UnitsApi::toNumber(ui->planeWidth->value(), format))
         .arg(placement)
         .arg(tr("Plane"));
 }
 
 QString DlgPrimitives::createBox(const QString& objectName, const QString& placement) const
 {
-   return QString::fromLatin1(
+    Base::QuantityFormat format(Base::QuantityFormat::Fixed, Base::UnitsApi::getDecimals());
+    return QString::fromLatin1(
         "App.ActiveDocument.addObject(\"Part::Box\",\"%1\")\n"
         "App.ActiveDocument.%1.Length=%2\n"
         "App.ActiveDocument.%1.Width=%3\n"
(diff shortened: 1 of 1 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7219">pipeline status</a> for feature branch [PR_7219](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7219). Pipeline [591100092](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/591100092) was triggered at [303cba8](https://github.com/FreeCAD/FreeCAD/pull/7219/commits/303cba8ca327d4fdf1951e3f756b0d084e35a3db). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@0penBrain , can you please review if this fix works with the new decimal separator settings on FC 0.20?

0penBrain (commented):
@donovaly No problem with decimal separator management. This latter only acts on presentation and interpretation in widgets, but not to internal value representation (a float is a float, it's not language aware). So no problem with that.

However, looking at the solution, it's seems pretty imperfect to me.
First it calls a lot of functions, but not the biggest problem to me.

The important thing is that, for users working with imperial units, it creates a rounding that leads to precision issue. Precisely, it converts (with rounding) to a unitless value, considered as mm. Worse, the precision issue is hidden to the user, as the value is rounded again to 'in' to be presented (eg. in the property pane).
Also the presentation to the user in the Python console when object is created is not so neat as you have these weird unitless numbers.

Let's take an example. Setup units as imperial in FC preferences, then create a 0.39" side box. Notice 0.39" is precisely 0.39*25.4 = 9.906 mm.
Here what user sees in the Python console :
```
>>> App.ActiveDocument.addObject("Part::Box","Box")
>>> App.ActiveDocument.Box.Length=9.91
>>> App.ActiveDocument.Box.Width=9.91
>>> App.ActiveDocument.Box.Height=9.91
>>> App.ActiveDocument.Box.Placement=App.Placement(App.Vector(0.00,0.00,0.00),App.Rotation(App.Vector(0.00,0.00,1.00),0.00))
>>> App.ActiveDocument.Box.Label='Box'
```
As said, not so neat to have those '9.91' and you see it has been rounded.
Also you can confirm there has been a precision loss.
```
>>> App.ActiveDocument.Box.Length.getValueAs('mil')/1000
0.39015748031496067
```

I would then suggest a different approach, considering that quantity values such as box Length/Width/Height can be set using a quantity string.
Here's what I'd propose, demonstrated on the box creation function (would have to be done for all functions) : 
```
QString DlgPrimitives::createBox(const QString& objectName, const QString& placement) const
{
    Base::QuantityFormat format(Base::QuantityFormat::Fixed, Base::UnitsApi::getDecimals());
    return QString::fromLatin1(
        "App.ActiveDocument.addObject(\"Part::Box\",\"%1\")\n"
        "App.ActiveDocument.%1.Length='%2'\n"
        "App.ActiveDocument.%1.Width='%3'\n"
        "App.ActiveDocument.%1.Height='%4'\n"
        "App.ActiveDocument.%1.Placement=%5\n"
        "App.ActiveDocument.%1.Label='%6'\n")
        .arg(objectName)
        .arg(ui->boxLength->value().getUserString())
        .arg(ui->boxWidth->value().getUserString())
        .arg(ui->boxHeight->value().getUserString())
        .arg(placement)
        .arg(tr("Box"));
}
```

With same example, here what user sees in the Python console : 
```
>>> App.ActiveDocument.addObject("Part::Box","Box")
>>> App.ActiveDocument.Box.Length='0.39 in'
>>> App.ActiveDocument.Box.Width='0.39 in'
>>> App.ActiveDocument.Box.Height='0.39 in'
>>> App.ActiveDocument.Box.Placement=App.Placement(App.Vector(0.00,0.00,0.00),App.Rotation(App.Vector(0.00,0.00,1.00),0.00))
>>> App.ActiveDocument.Box.Label='Box'
```
which seems better to me.

Also confirm no precision loss (beyond normal float rounding) : 
```
>>> App.ActiveDocument.Box.Length.getValueAs('mil')/1000
0.39000000000000007
```

HTH

Syres916:
@0penBrain I knew I should never have got involved in this. Firstly any real engineer who is working to 1/32" (0.03125) would never use only two decimal places, it would be five. Secondly I used the implementation from the equivalent PartDesign file https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/PartDesign/Gui/TaskPrimitiveParameters.cpp#L687
, therefore I was trying to implement some consistency but it would appear that's faulty too, This had been changed in the past due to a bug where large sizes such as 1023m got rounded down to 1020m after Werner tried to fix yet another issue handling very small sizes like 3 thou. Maybe all the task dialogs in Part and Part Design should have a consistent way to implement such a common input mechanism which has one test to pass??

0penBrain:
> @0penBrain I knew I should never have got involved in this.

Sorry for this. :) don't get me wrong, I value the PR. But I found it suboptimal by some points, so just reported and proposed.
> Firstly any real engineer who is working to 1/32" (0.03125) would never use only two decimal places, it would be five.

The purpose of my example isn't to show something 'real', just to demonstrate the problem I see.

> Secondly I used the implementation from the equivalent PartDesign file https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/PartDesign/Gui/TaskPrimitiveParameters.cpp#L687 , therefore I was trying to implement some consistency but it would appear that's faulty too, 

I have no doubt you wanted to do right. I think there is a problem also there in PartDesign. @wwmayer comment on the commit that introduces the change there says it used 'full internal precision', but actually it doesn't as I just showed on my previous post. ;)
> This had been changed in the past due to a bug where large sizes such as 1023m got rounded down to 1020m after Werner tried to fix yet another issue handling very small sizes like 3 thou.

I admittedly didn't test my proposal with very large or very tiny value. Will do and report.
> Maybe all the task dialogs in Part and Part Design should have a consistent way to implement such a common input mechanism which has one test to pass??

Sure. Let's find something that works perfectly and we'll propagate it widely in FC.

donovaly:
Many thanks to you both!

Let's do it in a two stage process:
- I will merge the PR since it fixes the main issue and Part is then on the state of PD
- @0penBrain , can you please make a further PR addressing Part and PD? Your solution looks good to me.

> I knew I should never have got involved in this.

Why? It is good that you had a look and fixed the issue. And criticism is part of the game, especially because FreeCAD lives from the community and often one finds the optimal solutions in discussions.

PR #7222, opened 2022-07-20
Make image scaling command aware of user units

0penBrain:
This PR makes the "distance" in Image Scaling command aware of user unit, i.e. it will : 
* Propose a default value using and showing user base unit
* If a unitless value is entered, it will consider it is given as user base unit

To do so, the QLineEdit has been replaced by a InputField (so that user also benefit from the checker). This is 2nd commit.

However to do so, I had to make a general change in InputField class (this is 1st commit).
This commit will let a InputField consider a unitless value is given as user base unit, if and only if value has been entered by user in the GUI (programmatically set values are still considered in FC core base unit if unitless).
This change will not only affect Image Scaling command, but all InputField across FC (eg. grid size in the working plane panel), hence I'd like feedback about people implied in Draft dev before this is merged. @yorikvanhavre @Roy-043

### DIFF
--- a/src/Gui/InputField.cpp
+++ b/src/Gui/InputField.cpp
@@ -642,6 +642,19 @@ void InputField::focusInEvent(QFocusEvent *event)
 
 void InputField::focusOutEvent(QFocusEvent *event)
 {
+    try {
+        if (Quantity::parse(this->text()).getUnit().isEmpty()) {
+            // if user didn't enter a unit, we virtually compensate
+            // the multiplication factor induced by user unit system
+            double factor;
+            QString unitStr;
+            actQuantity.getUserString(factor, unitStr);
+            actQuantity = actQuantity * factor;
+        }
+    }
+    catch (const Base::ParserError& e) {
+        // do nothing, let apply the last known good value
+    }
     this->setText(actQuantity.getUserString());
     QLineEdit::focusOutEvent(event);
 }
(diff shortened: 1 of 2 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7222">pipeline status</a> for feature branch [PR_7222](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7222). Pipeline [592414920](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/592414920) was triggered at [ca675ee](https://github.com/FreeCAD/FreeCAD/pull/7222/commits/ca675eee114bc8e033b5d9e23554595096c434f9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
many thanks for this fix!

I will merge now.

but I see room for UX improvement:
FreeCAD_t0M7CJd5cu

- the OK button should not be active once no image plane was yet clicked
- the dialog should not accept zero as input because the user could accidentally licked OK and then wonders why he does not see anything. Zero as scaling is also not sensible. Therefore I would even say, the OK button should not be active unless a value > 0 was entered.

0penBrain:
> but I see room for UX improvement: !

Can only agree. :) 
> Can this fix be backported or do you count it more as new feature?

IMO not a bug fix, but not a new feature either. Let's classify this as "improvement". ;)

PR #7239, opened 2022-07-23
[App] remove check for Python 2

donovaly:


donovaly:
I think this is pretty simple -> merging

PR #7240, opened 2022-07-23
[FEM] remove checks for Python 2

donovaly:
should be straight forward I think


donovaly:
> should be straight forward I think

Yes, therefore merging now

PR #7241, opened 2022-07-23
[Part] remove code for < GCC 4.8

donovaly:
we don't support such old GCC versions anymore so I think this can be removed

### DIFF
--- a/src/Mod/Part/App/GeometryDefaultExtension.cpp
+++ b/src/Mod/Part/App/GeometryDefaultExtension.cpp
@@ -72,24 +72,11 @@ std::unique_ptr<Part::GeometryExtension> GeometryDefaultExtension<T>::copy(void)
 
     copyAttributes(cpy.get());
 
-    #if (defined(__GNUC__) && __GNUC__ < 7 ) || defined(_MSC_VER)
-        return std::move(cpy); // GCC 4.8 and MSC do not support automatic move constructor call if the compiler fails to elide
+    #if defined(_MSC_VER)
+        return std::move(cpy); // MSC does not support automatic move constructor call if the compiler fails to elide
     #else
         return cpy; // all the others do automatic move constructor if RVO optimization not possible.
     #endif
-    // Advise from Scott Meyers Effective Modern c++:
-    //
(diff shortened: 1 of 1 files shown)


donovaly:
@wwmayer can you please have a quick look if the call is still necessary for MSVC?

wwmayer:
It's best to assign it to @abdullahtahiriyo because he once added the comment.

donovaly:
@abdullahtahiriyo, can you please have a quick look? I think it is straight-forward but maybe we can remove the call also for MS Visual C.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7241">pipeline status</a> for feature branch [PR_7241](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7241). Pipeline [600445261](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600445261) was triggered at [88b60d0](https://github.com/FreeCAD/FreeCAD/pull/7241/commits/88b60d07b4b0bd685479b1052e0cc9d2191b6cfc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
No response for a while. I think this is save since GCC 4.8 is really old

abdullahtahiriyo:
There is a while I do not so such an old version of GCC. Reducing the conditional compilation is not much of an improvement IMO. It would be cleaning up if we could get rid of the MSVC part too. I am not a MSVC user, so I do not know when/if MS has decided to follow the standard and elide. If MSVC users can reasonably say that nowadays MSVC works without, then we could remove it. All in all, I am ok with the merge.

donovaly:
> I am not a MSVC user, so I do not know when/if MS has decided to follow the standard and elide. If MSVC users can reasonably say that nowadays MSVC works without, then we could remove it.

Tanks for having a look. I am a MSVC user. How can I test if the compiler works now correctly?
I created now a PR that removes the MSVC code: #7613

PR #7262, opened 2022-07-27
[FEM] fix problem of several visible pipelines

donovaly:
When there are several visible pipelines, they are not drawn over each other because this is technically not possible (see comment on the PR). Therefore it is for the user like a bug that he has e.g. 3 pipelines but always see the one listed first in the Tree View.

After some discussions it turned out that the best way is to show the selected pipeline and make all other invisible.

Working on real-life simulations showed that there is indeed no use case where one likes/needs to have painted 2 pipelines over each other and see both.

### DIFF
--- a/src/Mod/Fem/Gui/ViewProviderFemPostObject.cpp
+++ b/src/Mod/Fem/Gui/ViewProviderFemPostObject.cpp
@@ -55,6 +55,7 @@
 #include <Gui/SoFCColorBar.h>
 #include <Gui/TaskView/TaskDialog.h>
 #include <Mod/Fem/App/FemPostFilter.h>
+#include <Mod/Fem/App/FemPostPipeline.h>
 
 #include "ViewProviderFemPostObject.h"
 #include "TaskPostBoxes.h"
@@ -761,11 +762,34 @@ void ViewProviderFemPostObject::onSelectionChanged(const Gui::SelectionChanges &
     // color bar.
     // But don't do this if the object is invisible because other objects with a
     // color bar might be visible and the color bar is then wrong.
+    // For pipelines, hide other visible ones and make the current visible since there
+    // is technically no way to paint them over each other and showing the one one likes
+    // Corresponding info from Realthunder:
+    // "Because on top rendering delays rendering to the last, where the OpenGL depth buffer
+    // is already populated. On top rendering is done with depth write disabled and only with
+    // transparency painting on top is somewhat visually acceptable."
     if (sel.Type == sel.AddSelection) {
         Gui::SelectionObject obj(sel);
         if (obj.getObject() == this->getObject()) {
+            if (this->getObject()->isDerivedFrom(Fem::FemPostPipeline::getClassTypeId()))
+                this->getObject()->Visibility.setValue(true);
             if (this->getObject()->Visibility.getValue())
                 WriteColorData(true);
+            if (!this->getObject()->isDerivedFrom(Fem::FemPostPipeline::getClassTypeId()))
+                    return;
+            // hide other pipelines
+            auto docGui = Gui::Application::Instance->activeDocument();
+            if (!docGui)
+                return;
+            auto doc = docGui->getDocument();
+            std::vector<App::DocumentObject *> ObjectsList = doc->getObjects();
+            for (auto it = ObjectsList.begin(); it != ObjectsList.end(); ++it) {
+                if ((*it)->Visibility.getValue()
+                    && (*it)->isDerivedFrom(Fem::FemPostPipeline::getClassTypeId())
+                    && (*it) != obj.getObject()) {
+                    (*it)->Visibility.setValue(false);
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Mod/Fem/Gui/ViewProviderFemPostObject.cpp
```diff
@@ -758,11 +759,35 @@ void ViewProviderFemPostObject::onSelectionChanged(const Gui::SelectionChanges &
...
+    // "Because on top rendering delays rendering to the last, where the OpenGL depth buffer
+    // is already populated. On top rendering is done with depth write disabled and only with
+    // transparency painting on top is somewhat visually acceptable."
     if (sel.Type == sel.AddSelection) {
         Gui::SelectionObject obj(sel);
         if (obj.getObject() == this->getObject()) {
+            if (this->getObject()->isDerivedFrom(Fem::FemPostPipeline::getClassTypeId()))
+                this->getObject()->Visibility.setValue(true);
+            this->getObject()->Visibility.setValue(true);
             if (this->getObject()->Visibility.getValue())
```
benj5378, line 774: If you set visibility to true, what is then the point on the next line to check whether the visibility is true or not?

donovaly: Oops, sorry. Seems a copy/paste mistake.
Now it is correct.

wwmayer:
> When there are several visible pipelines, they are not drawn over each other because this is technically not possible (see comment on the PR).

This is not quite true. See https://github.com/coin3d/coin/blob/master/src/nodes/SoAnnotation.cpp#L34 and https://mevislabdownloads.mevis.de/docs/current/MeVis/ThirdParty/Documentation/Publish/OpenInventorReference/classSoAnnotation.html

If you have an FEM project with two pipeline objects then the objects might be ResultPipeline and ResultPipeline001 and you can only see the former.

One hackish solution could be to move the second pipeline object from the scene graph and insert it in front of the first pipeline object.

Another solution is to use the mentioned SoAnnotation node. As mentioned in the links rendering of geometric objects may cause some weird artifacts and this happened here too. However, by adjusting the SoShapeHints node I get the rendering working correctly.

Here is the Python test code:
```
from pivy import coin

view = Gui.ActiveDocument.ActiveView.getViewer()
node = view.getSceneGraph()
child = node.getChild(1).getChild(0)

anno = coin.SoAnnotation()
child.addChild(anno)

vp = App.ActiveDocument.ResultPipeline001.ViewObject.RootNode
anno.addChild(vp)

# fix the rendering
action = coin.SoSearchAction()
action.setInterest(action.FIRST)
action.setSearchingAll(False)
action.setType(coin.SoShapeHints.getClassTypeId())
action.apply(vp)
path = action.getPath()
hints = path.getTail()
hints.vertexOrdering = hints.CLOCKWISE
hints.shapeType = hints.SOLID
```

donovaly:
> This is not quite true.

These are good news.

> Another solution is to use the mentioned SoAnnotation node. 
> 
> Here is the Python test code:
> 
> ```
> hints = path.getTail()
> ```

At this code line I get a Python error:

```
>>> hints = path.getTail()
AttributeError: 'NoneType' object has no attribute 'getTail'
```

Since you know the code and I am not helpful as intermediate player, I would be happy when you make a commit or PR that I can further test with my documents. (I don't even know how to translate your test code to C++.)

benj5378:
> At this code line I get a Python error:
> 
> ```
> >>> hints = path.getTail()
> Traceback (most recent call last):
>   File "<input>", line 1, in <module>
> AttributeError: 'NoneType' object has no attribute 'getTail'
> ```
> 
> Since you know the code and I am not helpful as intermediate player, I would be happy when you make a commit or PR that I can further test with my documents. (I don't even know how to translate your test code to C++.)

@donovaly This is because `path` is NoneType (null). Pronably because NoneType (null) was returned when assigning the value of `path` by `action.getPath()`

wwmayer:
> This is because path is NoneType (null). Pronably because NoneType (null) was returned when assigning the value of path by action.getPath()

This means for some reason the SoSearchAction couldn't find an SoShapeHints in the root node of the view provider. Maybe the action was applied on the wrong object.

wwmayer:
Consider it as a proof of concept: https://github.com/wwmayer/FreeCAD/tree/fem_annotate

If it doesn't work for your project then upload it here, please.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7262">pipeline status</a> for feature branch [PR_7262](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7262). Pipeline [600453104](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600453104) was triggered at [e46e03c](https://github.com/FreeCAD/FreeCAD/pull/7262/commits/e46e03ca070f72a8b0c89febcdd94ed978a91224). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> Consider it as a proof of concept: https://github.com/wwmayer/FreeCAD/tree/fem_annotate

Perfect! This is exactly what I had in mind. Works perfectly except of a minor issue I commented on. Please put this in soon if possible.

donovaly:
Closing, better fix committed by https://github.com/FreeCAD/FreeCAD/commit/8bb85d2b7cb

PR #7265, opened 2022-07-28
[Sketcher] fix icon name

donovaly:
- fix a FIXME, Sketcher_CreateEllipse depicts already the right thing, it just has to be renamed


0penBrain:
@donovaly had a very quick look. Looks to me like the SVG names aren't really consistent with others used in the same directory. ;) 
No damage though, that will work fine.

donovaly:
Thanks. I used the name of the file that was already in the Wiki. Changing this would have been more impact. I think Abdullah will do this if he likes to.

0penBrain:
> Thanks. I used the name of the file that was already in the Wiki. Changing this would have been more impact.

That's an excellent rational. 👍

PR #7273, opened 2022-07-29
Base: Fix Placement.rotate() to match documentation

mfro0:
Documentation for Placement.rotate() claims this function would be
compatible to TopoShape.rotate() which isn't exactly correct as the
generated placements differ because of backwards multiplication.

This commit adds a "comp" (as in "compatible") keyword argument that -
when "True" (default is False) - produces a Placement that is interchangable with the one
that is generated from TopoShape.rotate().

---

### DIFF
--- a/src/Base/PlacementPy.xml
+++ b/src/Base/PlacementPy.xml
@@ -63,14 +63,19 @@ Alias to move(), to be compatible with TopoShape.translate().\n
 vector : Base.Vector\n    Vector by which to move the placement.</UserDocu>
 			</Documentation>
 		</Methode>
-        <Methode Name="rotate">
+        <Methode Name="rotate" Keyword="true">
         <Documentation>
-            <UserDocu>rotate(center, axis, angle) -> None\n
+            <UserDocu>rotate(center, axis, angle, comp) -> None\n
 Rotate the current placement around center and axis with the given angle.
-This method is compatible with TopoShape.rotate().\n
+This method is compatible with TopoShape.rotate() if the (optional) keyword
+argument comp is True (default=False).
+
 center : Base.Vector, sequence of float\n    Rotation center.
 axis : Base.Vector, sequence of float\n    Rotation axis.
-angle : float\n    Rotation angle in degrees.</UserDocu>
+angle : float\n    Rotation angle in degrees.
+comp : bool\n    optional keyword only argument, if True (default=False),
+behave like TopoShape.rotate() (i.e. the resulting placements are interchangeable).
+</UserDocu>
         </Documentation>
         </Methode>
                 <Methode Name="multiply" Const="true">
--- a/src/Base/PlacementPyImp.cpp
+++ b/src/Base/PlacementPyImp.cpp
@@ -163,22 +163,36 @@ PyObject* PlacementPy::translate(PyObject * args)
     return move(args);
 }
 
-PyObject* PlacementPy::rotate(PyObject *args) {
-    PyObject *obj1, *obj2;
+PyObject* PlacementPy::rotate(PyObject *args, PyObject *kw) {
     double angle;
-    if (!PyArg_ParseTuple(args, "OOd", &obj1, &obj2, &angle))
+    char *keywords[] =  { "center", "axis", "angle", "comp", nullptr };
+    Vector3d center;
+    Vector3d axis;
+    PyObject* pyComp = Py_False;
+    
+    if (!PyArg_ParseTupleAndKeywords(args, kw, "(ddd)(ddd)d|$O!", keywords, &center.x, &center.y, &center.z,
+                                         &axis.x, &axis.y, &axis.z, &angle, &PyBool_Type, &pyComp))
         return nullptr;
-
+    
     try {
-        Py::Sequence p1(obj1), p2(obj2);
-        Vector3d center((double)Py::Float(p1[0]),
-                        (double)Py::Float(p1[1]),
-                        (double)Py::Float(p1[2]));
-        Vector3d axis((double)Py::Float(p2[0]),
-                      (double)Py::Float(p2[1]),
-                      (double)Py::Float(p2[2]));
-        (*getPlacementPtr()) *= Placement(
-                Vector3d(),Rotation(axis,toRadians<double>(angle)),center);
+        /*
+         * if comp is False, we retain the original behaviour that - contrary to the documentation - generates
+         * Placements different from TopoShape.rotate() to ensure compatibility for existing code
+         */
+        bool comp = Base::asBoolean(pyComp);
+
+        if (!comp) {
+            (*getPlacementPtr()) *= Placement(
+                        Vector3d(),Rotation(axis,toRadians<double>(angle)),center);
+        } else 
+        {
+            // multiply new Placement the same way TopoShape.rotate() does
+        
+            Placement p = *getPlacementPtr();
+            *getPlacementPtr() = Placement(
+                        Vector3d(),Rotation(axis,toRadians<double>(angle)),center) * p;
+        }
+        
         Py_Return;
     }
     catch (const Py::Exception&) {


src/Base/PlacementPy.xml
```diff
@@ -63,14 +63,19 @@ Alias to move(), to be compatible with TopoShape.translate().\n
 vector : Base.Vector\n    Vector by which to move the placement.</UserDocu>
 			</Documentation>
 		</Methode>
-        <Methode Name="rotate">
+        <Methode Name="rotate" Keyword="true">
         <Documentation>
-            <UserDocu>rotate(center, axis, angle) -> None\n
+            <UserDocu>rotate(center, axis, angle, kwargs) -> None\n
```
Roy-043, line 68: Shouldn't `kwargs` be `comp` here?

src/Base/PlacementPy.xml
```diff
@@ -63,14 +63,19 @@ Alias to move(), to be compatible with TopoShape.translate().\n
...
 Rotate the current placement around center and axis with the given angle.
-This method is compatible with TopoShape.rotate().\n
+This method is compatible with TopoShape.rotate() if the (optional) keyword
+argument comp is True (default=False).
+
 center : Base.Vector, sequence of float\n    Rotation center.
 axis : Base.Vector, sequence of float\n    Rotation axis.
-angle : float\n    Rotation angle in degrees.</UserDocu>
+angle : float\n    Rotation angle in degrees.
+comp : bool\n    optional keyword only argument, if Ture (default=False),
```
Roy-043, line 76: Typo: `Ture` should be `True`.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273">pipeline status</a> for feature branch [PR_7273](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273). Pipeline [600469899](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600469899) was triggered at [69b1f19](https://github.com/FreeCAD/FreeCAD/pull/7273/commits/69b1f1930eef8fb0d4ea19ff7110dea635497a01). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Base/PlacementPy.xml
mfro0: I'm no Python expert and not sure about that, but I believe that's how keyword arguments are commonly documented. With the introduction of keyword arguments, all arguments can be given by keyword, not just comp. Only positional arguments, if given, have to be specified in the order shown. So you can write

    rotate(center, angle=45, axis=(0,0,1), comp=False)
or

    rotate(center, axis, angle)

but not

    rotate(center, axis, angle, True)

src/Base/PlacementPy.xml
mfro0: fixed with d2d228dc89fcec750eb6226e64996dd3babe09a1

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273">pipeline status</a> for feature branch [PR_7273](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273). Pipeline [600567116](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600567116) was triggered at [c007ec2](https://github.com/FreeCAD/FreeCAD/pull/7273/commits/c007ec2fdbd71b88ea786a3c61b84ce00fc571f4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273">pipeline status</a> for feature branch [PR_7273](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273). Pipeline [600578560](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600578560) was triggered at [04fcbf6](https://github.com/FreeCAD/FreeCAD/pull/7273/commits/04fcbf661d668ddcc78b49339249ee7b52c4b386). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273">pipeline status</a> for feature branch [PR_7273](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273). Pipeline [600751323](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600751323) was triggered at [bd9e901](https://github.com/FreeCAD/FreeCAD/pull/7273/commits/bd9e9010b55824621305ba6fdda7631dfe814607). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Base/PlacementPy.xml
Roy-043: OK, thanks for explaining (I am not a Python expert either).

Roy-043:
@mfro0 I am not able to help with the requested review: I do not have C++ skills and do not compile FreeCAD myself.

mfro0:
> @mfro0 I am not able to help with the requested review: I do not have C++ skills and do not compile FreeCAD myself.

then please excuse my bothering. Thank you anyway!

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273">pipeline status</a> for feature branch [PR_7273](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273). Pipeline [600851980](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600851980) was triggered at [2143a9c](https://github.com/FreeCAD/FreeCAD/pull/7273/commits/2143a9cfade7df993d9aeb098c796c32ce55bcbe). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Base/PlacementPy.xml
benj5378: `**kwargs` is when you have an unknown number of keyword arguments. `*args` when you have an unknown number of non-keyword arguments. Since `comp` is a known argument with a default value, it should just be `comp` in the argument list.

benj5378: `**kwargs` is when you have an unknown number of keyword arguments. `*args` when you have an unknown number of non-keyword arguments. Since `comp` is a known argument with a default value, it should just be `comp` in the argument list. @Roy-043 @mfro0

src/Base/PlacementPyImp.cpp
```diff
@@ -163,22 +163,33 @@ PyObject* PlacementPy::translate(PyObject * args)
     return move(args);
 }
 
-PyObject* PlacementPy::rotate(PyObject *args) {
-    PyObject *obj1, *obj2;
+PyObject* PlacementPy::rotate(PyObject *args, PyObject *kw) {
     double angle;
-    if (!PyArg_ParseTuple(args, "OOd", &obj1, &obj2, &angle))
+    bool comp = false;
```
benj5378, line 168: This should be an un-assigned bool, since the default value is given in the XML file - if you change it to `rotate(center, axis, angle, comp=False) -> None`

benj5378 (commented):
I am very unsure about how the Python to C++ connections work, so all my comments are regarding how it would be done in case of pure Python - not sure if it applies to here too, but I believe so. Maybe @chennes knows more

src/Base/PlacementPy.xml
benj5378: @chennes

src/Base/PlacementPyImp.cpp
mfro0: I guess that's a matter of taste.

My taste is: C++ code should always rely on itself, never on anything else. If - for whatever reason - the XML assignment would fail, the C++ code would fail as well.

mfro0: Documentation: comp explicitly mentioned in the parameter list, although it's a keyword (only) argument.
Fixed with 2143a9cfade7df993d9aeb098c796c32ce55bcbe

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273">pipeline status</a> for feature branch [PR_7273](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273). Pipeline [600933626](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600933626) was triggered at [6bee41b](https://github.com/FreeCAD/FreeCAD/pull/7273/commits/6bee41bcf71529b511a36baea988a3a3ad697c71). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

marioalexis84:
Please, squash the commits to only one. There is no need to have multiple commits for this pull request.

src/Base/PlacementPyImp.cpp
marioalexis84: > his should be an un-assigned bool, since the default value is given in the XML file - if you change it to `rotate(center, axis, angle, comp=False) -> None` as I recommended

No, the <UserDocu> tag in the xml file only is used to generate the docstrings, there is no relation with the code implementation.

marioalexis84: > My understanding of "keyword=true", btw was that this just says the method takes keyword arguments?

Yes, the attribute Keyword on Methode elements "tells" to the generator which parameters the PyCFunction callback should take and sets the appropriate flags.

src/Base/PlacementPy.xml
marioalexis84: There is an attempt to standardize the documentation, so the format should be
rotate(center, axis, angle, comp=False) -> None\n
In the argument description, please start it with capital, add four spaces in each line and limit the size (something like 80 characters) so it will look like:
```
angle : float
    Rotation angle in degrees.
comp : bool
    Optional keyword only argument, if True (default=False), behave like
    TopoShape.rotate() (i.e. the resulting placements are interchangeable).
```
And add the `</UserDocu>` tag to the end of the last string. Putting in the next line generate a blank line at the end of the tooltip.

src/Base/PlacementPyImp.cpp
```diff
@@ -163,22 +163,33 @@ PyObject* PlacementPy::translate(PyObject * args)
...
-        (*getPlacementPtr()) *= Placement(
-                Vector3d(),Rotation(axis,toRadians<double>(angle)),center);
+        /*
+         * if comp is False, we retain the original behaviour that - contrary to the documentation - generates
+         * Placements different from TopoShape.rotate() to ensure compatibility for existing code
+         */
+        if (!comp) {
+            (*getPlacementPtr()) *= Placement(
+                        Vector3d(),Rotation(axis,toRadians<double>(angle)),center);
+        } else {
```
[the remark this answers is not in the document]
mfro0: Thanks for your (really useful) comments, @marioalexis84 . Should be fixed now.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273">pipeline status</a> for feature branch [PR_7273](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273). Pipeline [601182851](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/601182851) was triggered at [8c0612c](https://github.com/FreeCAD/FreeCAD/pull/7273/commits/8c0612cc5db5de493a747783aeae6416e907f4b5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

benj5378:
@mfro0 Do you have Discord? Could you please add bensay#5141 and I can help you rebase all these commits (change them into one).

mfro0:
> 

umm, thought I did it in the morning already, but somehow my fingers seemed to refuse the required forced push ;)

Thanks for the heads-up. Should be good now.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273">pipeline status</a> for feature branch [PR_7273](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7273). Pipeline [601240388](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/601240388) was triggered at [a7654e6](https://github.com/FreeCAD/FreeCAD/pull/7273/commits/a7654e6f4c51b96093b8f67adc211ba002e27b41). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

yorikvanhavre:
Although this is correct in theory, of course, since placement rotations are used all over FreeCAD, this might make thousands of FreeCAD models suddenly break all over the world with no real clue for users of what happened, and how to fix. I think we need to be very careful here and if we do this, do thorough testing and make up a strategy to help users to recover the mess...

Roy-043:
The new method has an extra `comp` argument. Are you aware of that?:
`rotate(center, axis, angle, comp=False)`

If comp is not provided (or `False`) the old algo will be used.

wwmayer:
I had a look at the code base and found only two occurrences of **this** rotate() method. It's in circ_placements and polar_placements which both are functions of Draft's array module.

So, I wonder whether we should really keep the faulty implementation (added with 3fcbf71fb5280) which from a mathematical point of view doesn't make too much sense or at least has a weird behaviour. And since its documentation explicitly mentions `This method is compatible with TopoShape.rotate()` but in fact isn't it must be considered as bug.

Given that there is no other function using rotate() I think we should fix it and adjust circ_placements and polar_placements accordingly.

Then there should be written some unit tests to ensure the compatibility with TopoShape.rotate:
```
# Use this as the base for a proper unit test
import math
from FreeCAD import Base

box = Part.makeBox(1, 1, 1)
box.Placement.Base = Base.Vector(10, 10, 10)
box.Placement
box.rotate((0, 0, 0), (0, 0, 1), 90)
box.Placement

p1 = Base.Placement()
p1.Base = Base.Vector(10, 10, 10)

p2 = Base.Placement()
p2.Rotation.Angle = math.radians(90)

# p3 is equal to box.Placement
p3 = p2 * p1
box.Placement

p4 = p1 * p2

p5 = p1.copy()
p5.rotate((0, 0, 0), (0, 0, 1), 90)

# p5 is equal to p4 (but actually should be equal to p3)
p4
p5
```

mfro0:
> I had a look at the code base and found only two occurrences of **this** rotate() method. It's in circ_placements and polar_placements which both are functions of Draft's array module.

a quick `grep` through my Mod directory showed that there seem to be quite some add on workbenches that use `Placement.rotate()`.

> Then in a future release we could remove the support for the old behaviour (but leave and ignore the comp parameter).

Not sure how this migration path should look like. Note that the proposed change has the old (wrong) behaviour as default, that means workbench developers had to do a two step migration (first adapt to the new behaviour '`comp=True`', then in a second step, drop the `comp` parameter (or revert its default).

My feeling is that it's way too late to reintroduce the correct behaviour as documented and we should rather document the new behaviour as the intended one.

wwmayer:
> A quick grep through my Mod directory showed that there seem to be quite some add on workbenches that use Placement.rotate(), however.

Do you talk about FreeCAD's Mod directory or that of  the Addons? In the FreeCAD code I found many calls of rotate() but except for two cases it everywhere is TopoShape.rotate() -- unless I have overlooked something.

> Not sure how this migration path should look like.

It looks like it's obsolete now. See my comment below.

> My belief is that it's way too late to reintroduce the correct behaviour as documented and we should rather document the new behaviour as the intended one.

In the meantime I have read the corresponding forum thread and Edward gave some examples. Although they are not really convincing they are valid. So, I agree that we shouldn't risk to break any existing Python scripts out there.

mfro0:
> Do you talk about FreeCAD's Mod directory or that of the Addons?

Addons. Lattice2 and Curves among the most prominent. I didn't check if they would be affected by the invasive ("correct") change, however.

wwmayer:
> Addons. Assembly 4, Lattice2 and Curves

Alright, merging it now...

PR #7284, opened 2022-07-30
Bump PACKAGE_VERSION_PATCH for next patch release

adrianinsaval:
This gets used here https://github.com/FreeCAD/FreeCAD/blob/5d49bf78de785a536f941f1a6d06d432582a95d3/src/XDGData/org.freecadweb.FreeCAD.appdata.xml.in#L61 and is the version picked up by flatpak so it should be updated too. Personally I don't understand why we have a PACKAGE_VERSION_PATCH and FREECAD_VERSION_PATCH as separate variables but I keep that way for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ project(FreeCAD)
 set(PACKAGE_VERSION_NAME "Vulcan")
 set(PACKAGE_VERSION_MAJOR "0")
 set(PACKAGE_VERSION_MINOR "20")
-set(PACKAGE_VERSION_PATCH "0")
+set(PACKAGE_VERSION_PATCH "1")
 set(PACKAGE_VERSION_SUFFIX "") # either "dev" for development snapshot or "" (empty string)
 set(FREECAD_VERSION_PATCH "1") # number of patch release (e.g. "4" for the 0.18.4 release)
 


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7284">pipeline status</a> for feature branch [PR_7284](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7284). Pipeline [600984407](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/600984407) was triggered at [0b99c07](https://github.com/FreeCAD/FreeCAD/pull/7284/commits/0b99c07818dceb4bd4f7af45b6653d4ece332ff1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Why for master? I think this should only go to the 0-20 branch, right?

adrianinsaval:
Notice the PR is for FreeCAD-0-20 branch, not master

donovaly:
> Notice the PR is for FreeCAD-0-20 branch, not master

GitHub tells me it would be merged to the "base branch". I cannot see the 0-20, therefore I was asking.

adrianinsaval:
Weird, on mobile it's shown here and I think on desktop too
Screenshot_2022-07-30-23-44-16-709_com brave browser_1

donovaly:
Yes, after the merge. But I wanted to see it before I merge where it goes.

adrianinsaval:
For me it also showed before merge, this is bellow the title of the PR. I can see in other PRs that it says something like `X wants to merge # commits into FreeCAD:branch from X:xbranch` so it's not something just shown to the person submitting.

Thanks for merging, I'll probably make a PR to master to unify those two so we don't have to change the number in two places for no reason, it's too prone to error.

PR #7287, opened 2022-07-31
[Arch] remove unneeded checks for Python 3

donovaly:
since we only support Python >=3, these checks can go

### DIFF
--- a/src/Mod/Arch/ArchCommands.py
+++ b/src/Mod/Arch/ArchCommands.py
@@ -20,7 +20,6 @@
 #*                                                                         *
 #***************************************************************************
 
-import sys
 import FreeCAD,Draft,ArchComponent,DraftVecUtils
 from FreeCAD import Vector
 if FreeCAD.GuiUp:
@@ -60,8 +59,6 @@ def string_replace(text, pattern, replacement):
     u'abc mm \xc2\xb3'
     ```
     """
-    if sys.version_info.major < 3:
-        text = text.encode("utf8")
     return text.replace(pattern, replacement)
 
 
@@ -937,8 +934,6 @@ def survey(callback=False):
                     if FreeCAD.SurveyObserver.totalLength:
                         u = FreeCAD.Units.Quantity(FreeCAD.SurveyObserver.totalLength,FreeCAD.Units.Length)
                         t = u.getUserPreferred()[0]
-                        if sys.version_info.major < 3:
-                            t = t.encode("utf8")
                         msg += " Length: " + t
                     if FreeCAD.SurveyObserver.totalArea:
                         u = FreeCAD.Units.Quantity(FreeCAD.SurveyObserver.totalArea,FreeCAD.Units.Area)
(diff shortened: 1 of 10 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7287">pipeline status</a> for feature branch [PR_7287](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7287). Pipeline [601048441](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/601048441) was triggered at [ccab19a](https://github.com/FreeCAD/FreeCAD/pull/7287/commits/ccab19a0822cd1d3dc0a2199e8e3f73f6cca329a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

benj5378:
@luzpaz I guess we got our answer. Only Python3 is used in FreeCAD

src/Mod/Arch/ArchCommands.py
```diff
@@ -1110,10 +1103,7 @@ def exportCSV(self):
         if rows:
             filename = QtGui.QFileDialog.getSaveFileName(QtGui.QApplication.activeWindow(), translate("Arch","Export CSV File"), None, "CSV file (*.csv)");
             if filename:
-                if sys.version_info.major < 3:
-                    mode = 'wb'
-                else:
-                    mode = 'w'
+                mode = 'w'
                 with open(filename[0].encode("utf8"), mode) as csvfile:
```
benj5378, line 1107:
```suggestion
                with open(filename[0].encode("utf8"), "w") as csvfile:
```

src/Mod/Arch/Dice3DS/dom3ds.py
```diff
@@ -21,8 +21,7 @@
 import numpy
 
 # xrange is not available in python3
```
benj5378, line 23:
```suggestion
```

src/Mod/Arch/Dice3DS/util.py
```diff
@@ -7,11 +7,9 @@
 """
 
 import numpy
-import sys
 
 # xrange is not available in python3
```
benj5378, line 11:
```suggestion
```

benj5378 (approved):

src/Mod/Arch/ArchCommands.py
donovaly: Thanks for your review!

your changes are further beyond what the PR does. I'll merge it now. Please make a new PR for further code simplification

benj5378: I don't agree. The only reason for the mode variable was changing between `w` and `wb` between Python 2 and 3. Removing the code changing those should also cause the removal of the variable.

My other comments were about removal of code explaining the code you just deleted. If you remove some code, you should also remove the comments that belonged to it.

donovaly: > I don't agree. The only reason for the mode variable was changing between w and wb between Python 2 and 3. 

Hmm, OK. I committed this change now.

> My other comments were about removal of code explaining the code you just deleted. If you remove some code, you should also remove the comments that belonged to it. How does that not belong to this PR

I don't agree hereby. why is range copied to xrange? There must be a comment for this why this is done since this is not obvious.
When there should not be a comment, the whole xrange variable should got. I made a PR for this, please review:
https://github.com/FreeCAD/FreeCAD/pull/7309

benj5378: > I don't agree hereby. why is range copied to xrange? There must be a comment for this why this is done since this is not obvious.
True

PR #7299, opened 2022-08-02
Fixed external SMESH cmake scripts - correct issue with bad FEM mesh generation with VTK9+

scottmudge:
Numerous reports (and personal past experiences on Windows + Linux) with corrupted/bad meshes being generated when FreeCAD is compiled with:

Internal SMESH
VTK >= 9
Either with GMSH or Netgen, the meshes come out looking like this:

image

(From: https://forum.freecadweb.org/viewtopic.php?style=10&p=609411)

I'm not sure what is incompatible in the internal SMESH module which causes, and why compiling against VTK 8.2 instead of VTK >= 9 fixes the issue, but many platforms now only have available VTK 9 by default. So when compiling FreeCAD from sources on these systems, mesh generation with the internal SMESH module does not work correctly.

I even saw this on the last LibPack I compiled for this project for Windows. I thought at the time the issue was with Netgen or GMSH, but it turns out this wasn't the case. It was the internal SMESH implementation.

The internal SMESH is based on SMESH 7, while the current external version is version 9. Compiling FreeCAD against VTK 9 with external SMESH fixes this mesh generation issue.

However, currently the FreeCAD CMake scripts do not properly link against or add the proper SMESH variables and configurations to the build process.

This commit fixes the CMake scripts to allow compiling against an external SMESH, and fixes the linker errors seen when attempting to link against external SMESH on 64-bit systems (this was due to `SALOME_USE_64BIT_IDS` not being defined, leading to FreeCAD expecting to link against SMESH functions with 32-bit arguments, while SMESH libraries exported functions with 64-bit arguments).

On my linux/Fedora machine, I was experiencing the same mesh generation issues with (both GMSH and Netgen) when compiling FreeCAD against VTK 9 with the internal SMESH enabled. Netgen would not even produce anything, it would just crash the module.

With external SMESH now properly working, the issue goes away and the FEM meshes are properly generated.

This is intended as a fix for VTK 9 compatibility, until the internal SMESH modules is updated to the latest SMESH sources.

### DIFF
--- a/cMake/FindSMESH.cmake
+++ b/cMake/FindSMESH.cmake
@@ -5,28 +5,55 @@
 # SMESH_INCLUDE_DIR   - where the Salome SMESH include directory can be found
 # SMESH_LIBRARIES     - Link this to use Salome SMESH
 #
- 
 
+# SMESH needs VTK
+find_package(VTK REQUIRED)
+
+# If this definition is not set, linker errors will occur against SMESH on 64 bit machines.
+if(CMAKE_SIZEOF_VOID_P STREQUAL 8)
+	add_definitions(-DSALOME_USE_64BIT_IDS)
+endif(CMAKE_SIZEOF_VOID_P STREQUAL 8)
+ 
 IF (CMAKE_COMPILER_IS_GNUCC)
     FIND_PATH(SMESH_INCLUDE_DIR SMESH_Mesh.hxx
     # These are default search paths, why specify them?
-    # /usr/include
-    # /usr/local/include
-    PATH_SUFFIXES smesh
+    PATH_SUFFIXES smesh SMESH smesh/SMESH
     )
-    FIND_LIBRARY(SMESH_LIBRARY SMESH
-    #  /usr/lib
-    #  /usr/local/lib
-    )
+    FIND_LIBRARY(SMESH_LIBRARY SMESH)
 ELSE (CMAKE_COMPILER_IS_GNUCC)
     # Not yet implemented
 ENDIF (CMAKE_COMPILER_IS_GNUCC)
+
+IF(SMESH_INCLUDE_DIR)
+	SET(SMESH_INC_ROOT "${SMESH_INCLUDE_DIR}/..")
+	# Append extra include dirs
+	SET(SMESH_INCLUDE_DIR 
+	"${SMESH_INCLUDE_DIR};
+	${SMESH_INC_ROOT}/Controls;
+	${SMESH_INC_ROOT}/Driver;
+	${SMESH_INC_ROOT}/DriverDAT;
+	${SMESH_INC_ROOT}/DriverGMF;
+	${SMESH_INC_ROOT}/DriverSTL;
+	${SMESH_INC_ROOT}/DriverUNV;
+	${SMESH_INC_ROOT}/Geom;
+	${SMESH_INC_ROOT}/Kernel;
+	${SMESH_INC_ROOT}/MEFISTO2;
+	${SMESH_INC_ROOT}/MeshVSLink;
+	${SMESH_INC_ROOT}/Netgen;
+	${SMESH_INC_ROOT}/NETGENPlugin;
+	${SMESH_INC_ROOT}/SMDS;
+	${SMESH_INC_ROOT}/SMESHDS;
+	${SMESH_INC_ROOT}/SMESHUtils;
+	${SMESH_INC_ROOT}/StdMeshers;")
+ELSE(SMESH_INCLUDE_DIR)
+	message(FATAL_ERROR "SMESH include directories not found!")
+ENDIF(SMESH_INCLUDE_DIR)
 
 SET(SMESH_FOUND FALSE)
 IF(SMESH_LIBRARY)
   SET(SMESH_FOUND TRUE)
   GET_FILENAME_COMPONENT(SMESH_LIBRARY_DIR ${SMESH_LIBRARY} PATH)
-  set(SMESH_LIBRARIES
+  set(SMESH_LIBRARIES 
     ${SMESH_LIBRARY_DIR}/libDriver.so
     ${SMESH_LIBRARY_DIR}/libDriverDAT.so
     ${SMESH_LIBRARY_DIR}/libDriverSTL.so
@@ -36,5 +63,8 @@ IF(SMESH_LIBRARY)
     ${SMESH_LIBRARY_DIR}/libSMESHDS.so
     ${SMESH_LIBRARY_DIR}/libStdMeshers.so
   )
+  set(EXTERNAL_SMESH_LIBS ${SMESH_LIBRARIES})
+ELSE(SMESH_LIBRARY)
+	message(FATAL_ERROR "SMESH libraries NOT FOUND!")
 ENDIF(SMESH_LIBRARY)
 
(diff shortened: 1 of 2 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7299">pipeline status</a> for feature branch [PR_7299](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7299). Pipeline [603042933](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/603042933) was triggered at [8352b54](https://github.com/FreeCAD/FreeCAD/pull/7299/commits/8352b54e39efe37d2f4a5bb8474458ffb05fe3ba). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

looooo:
@scottmudge I used a patch for external smesh:
https://github.com/looooo/freecad-feedstock/blob/master/recipe/patches/smesh.patch

I guess your fix will replace this patch.

CMakeLists.txt
```diff
@@ -55,6 +55,9 @@ if(NOT FREECAD_LIBPACK_USE OR FREECAD_LIBPACK_CHECKFILE_CLBUNDLER)
         # OPENGL_gl_LIBRARY empty that results into linker errors
         SetupOpenGL()
     endif(BUILD_GUI)
+    if (FREECAD_USE_EXTERNAL_SMESH)
+    	find_package(SMESH REQUIRED)
+    endif(FREECAD_USE_EXTERNAL_SMESH)
```
looooo, line 60: why not move this to SetupSalomeSMESH?

scottmudge:
> @scottmudge I used a patch for external smesh: https://github.com/looooo/freecad-feedstock/blob/master/recipe/patches/smesh.patch
> 
> I guess your fix will replace this patch.

This patch does not fix the linker errors I was encountering. It also lacked the search prefix/hint for the default installation directory for headers in the SMESH 9 cmake script, at least on my system.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7299">pipeline status</a> for feature branch [PR_7299](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7299). Pipeline [603830804](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/603830804) was triggered at [3b6ece4](https://github.com/FreeCAD/FreeCAD/pull/7299/commits/3b6ece4f46ff6edd5e841f7c507a7e86ab70543f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

CMakeLists.txt
scottmudge: One final change, I had excluded `BUILD_SMESH` if `FREECAD_USE_EXTERNAL_SMESH` was enabled, but changed this to allow both in the event unit testing or something else was desired when both were on.

This is identical to the previous logic anyway.

Also tabs->spaces.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7299">pipeline status</a> for feature branch [PR_7299](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7299). Pipeline [603844875](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/603844875) was triggered at [e03d8c0](https://github.com/FreeCAD/FreeCAD/pull/7299/commits/e03d8c0f7e40ae4d043887689184864c5e8971ad). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

CMakeLists.txt
wwmayer: Inside SetupSalomeSMESH.cmake we have the line `find_package(SMESH CONFIG)` (line 121) that searches for a file SMESHConfig.cmake or smesh-config.cmake.

In my local (external) SMESH build there is a SMESHConfig.cmake file. Don't you have this file or why do you need the `find_package(SMESH REQUIRED)`?

wwmayer:
> Unfortunately that patch does not fix the linker errors I was encountering.

I guess your PR fixes them. Does setting `SALOME_USE_64BIT_IDS` do the trick?

CMakeLists.txt
scottmudge: When I attempted to build using the existing CMake script when `FREECAD_USE_EXTERNAL_SMESH=ON` was set, it did not seem to appropriately initialize the required SMESH variables. I had to add that extra case.  Perhaps `BUILD_SMESH` was not being set. The variable is a bit ambiguous, as when `FREECAD_USE_EXTERNAL_SMESH` is enabled, SMESH is indeed usually *not* being built during the build process.

scottmudge:
> > Unfortunately that patch does not fix the linker errors I was encountering.
> 
> I guess your PR fixes them. Does setting `SALOME_USE_64BIT_IDS` do the trick?

Yes this is required.

cMake/FindSMESH.cmake
```diff
@@ -5,28 +5,55 @@
...
 # SMESH_LIBRARIES     - Link this to use Salome SMESH
 #
- 
 
+# SMESH needs VTK
+find_package(VTK REQUIRED)
+
+# If this definition is not set, linker errors will occur against SMESH on 64 bit machines.
+if(CMAKE_SIZEOF_VOID_P STREQUAL 8)
+	add_definitions(-DSALOME_USE_64BIT_IDS)
```
wwmayer, line 14: @looooo didn't you run into linker errors with your SMESH build from trelau's repo when this define was set?

cMake/FreeCAD_Helpers/SetupSalomeSMESH.cmake
```diff
@@ -1,6 +1,10 @@
 macro(SetupSalomeSMESH)
 # -------------------------------- Salome SMESH --------------------------
 
+    # Allow using external SMESH
```
wwmayer, line 4: CMake supports two search modes. If  you write `find_package(SMESH CONFIG)` then it's in config mode and either expects a file SMESHConfig.cmake or smesh-config.cmake.
If you write `find_package(SMESH REQUIRED)` then it's in module search mode and uses the FindSMESH.cmake file.

Because we already have the line `find_package(SMESH CONFIG)` that has worked quite well for many people the `find_package(SMESH REQUIRED)` should be used as fallback solution. Something like this:
```
find_package(SMESH CONFIG)
if (...) # successful
... # set some variables
else()
    find_package(SMESH REQUIRED)
    if (...) # successful
...
    else()
# raise an error
    endif()
endif()
```

cMake/FindSMESH.cmake
looooo: Yes, true. I solved it by commenting out this line in the smesh recipe :)

luzpaz:
Hi @scottmudge please follow-up to wmayer's responses above :point_up: (thanks!)

luzpaz:
bumping

donovaly:
@scottmudge, can you please rebase the PR and look at the code review comments?

adrianinsaval:
is this still relevant?

looooo:
It's merged in the related pr. So I guess it's ok to close this pr

adrianinsaval:
Thanks, @donovaly could you close it?

PR #7323, opened 2022-08-06
[Python console] prevent crash

mwganson:
Forum topic: https://forum.freecadweb.org/viewtopic.php?f=8&t=70732

Crash occurs when pasting the following into the python console:

```
if True:
    print('ß')
    print('ß')
```

There might be a better way to fix the issue, which seems almost certainly related to the use of non-ASCII characters.

### DIFF
--- a/src/Gui/PythonConsole.cpp
+++ b/src/Gui/PythonConsole.cpp
@@ -190,7 +190,9 @@ PyObject* InteractiveInterpreter::compile(const char* source) const
     PyObject* eval = PyObject_CallObject(func,args);  // must decref later
 #endif
 
-    Py_DECREF(args);
+    if (args) {
+        Py_DECREF(args);
+    }
     Py_DECREF(func);
 
     if (eval){


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7323">pipeline status</a> for feature branch [PR_7323](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7323). Pipeline [606464892](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/606464892) was triggered at [99b42a5](https://github.com/FreeCAD/FreeCAD/pull/7323/commits/99b42a5dd38470505b0c90a9a7b54dfcfa12827b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@wwmayer , I would like to have this in for 0.20.1. It looks straight forward, but you might want to have a look if more needs to be done.

al-right:
I am really not into the code base, but maybe this is of some help:
Could the toLatin1() that mfro mentions in:
https://forum.freecadweb.org/viewtopic.php?p=615500#p615500
, be related to back when FreeCAD used Python 2 prior to version 0.19?

Python 3 uses UTF-8 as default: https://peps.python.org/pep-3120/
Again - I am just guessing. Thanks :)

wwmayer:
Fixes the crash and the UTF-8 issue: c4547c03e1
@donovaly The change should be backported

donovaly:
Thanks!
Backported.

PR #7368, opened 2022-08-15
Arch: removing unnecessary 'if' conditional check

ebrahimraeyat:


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7368">pipeline status</a> for feature branch [PR_7368](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7368). Pipeline [613390211](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/613390211) was triggered at [c3e724d](https://github.com/FreeCAD/FreeCAD/pull/7368/commits/c3e724d78dce89459e5d446748ac4a602f372e04). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Arch/ArchRebar.py
```diff
@@ -468,28 +468,18 @@ def execute(self,obj):
...
-                shapes.append(bar)
-                w = basewire.copy()
-                w.Placement = obj.PlacementList[i]
-                self.wires.append(w)
+            bar = bar.copy()
+            bar.Placement = obj.PlacementList[i]
+            shapes.append(bar)
+            w = basewire.copy()
+            w.Placement = obj.PlacementList[i]
+            self.wires.append(w)
```
benj5378, line 482: Why can the specific logic for `i == 0` be removed?

ebrahimraeyat: > Why can the specific logic for `i == 0` be removed?

bar is a Part shape. we can simply copy it for each bar that must be add. The first one is not something special.

luzpaz:
@ebrahimraeyat please amend commit 

`git commit --amend -m "Arch: removing unnecessary 'if' conditional check" -m "Fix for ArchRebar.py"`

luzpaz:
CC @amrit3701 (rebar wb dev)

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7368">pipeline status</a> for feature branch [PR_7368](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7368). Pipeline [620950341](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/620950341) was triggered at [216ab71](https://github.com/FreeCAD/FreeCAD/pull/7368/commits/216ab71ade1db765cac31f70eb3581fafaaa878e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Arch/ArchRebar.py
amrit3701:
```suggestion
           if i != 0:
               bar = bar.copy()
               basewire = basewire.copy()
            bar.Placement = obj.PlacementList[i]
            shapes.append(bar)
            basewire.Placement = obj.PlacementList[i]
            self.wires.append(basewire)
```

I think we can use above logic. @ebrahimraeyat with your logic you are creating new `bar` and `basewire` objects  for `i=0` but objects are already there. Yes, Python has good garbage collector but here we can control object creation.

amrit3701 (changes requested):

ebrahimraeyat:
I didn't check the performance, but for creating stirrups we must check i value each time we create an stirrup, while bar and basewire are Part shape and not object and I think this is not important to create 1 extra copy of bar shape.

amrit3701 (approved):
LGTM!

yorikvanhavre:
I agree the extra object creation is not really important... shape.copy() is fast enough in this case (IIRC @realthunder discovered an even faster way but I can't remember) and the extra object should indeed be garbage-collected afterwards

PR #7415, opened 2022-08-29
Gui: Use auto (#1)

berniev:

---
Rather than present the whole gui folder at once ...

### DIFF
--- a/src/Gui/Action.cpp
+++ b/src/Gui/Action.cpp
@@ -247,7 +247,7 @@ void ActionGroup::addTo(QWidget *w)
     // and adding this action to the widget doesn't work.
     if (_dropDown) {
         if (w->inherits("QMenu")) {
-            QMenu *menu = new QMenu(w);
+            auto *menu = new QMenu(w);
             QAction* action = qobject_cast<QMenu*>(w)->addMenu(menu);
             action->setMenuRole(_action->menuRole());
             menu->setTitle(_action->text());
@@ -259,7 +259,7 @@ void ActionGroup::addTo(QWidget *w)
             tb->setPopupMode(QToolButton::MenuButtonPopup);
             tb->setObjectName(QString::fromLatin1("qt_toolbutton_menubutton"));
             QList<QAction*> acts = _group->actions();
-            QMenu* menu = new QMenu(tb);
+            auto* menu = new QMenu(tb);
             menu->addActions(acts);
             tb->setMenu(menu);
             //tb->addActions(_group->actions());
@@ -468,7 +468,7 @@ void WorkbenchComboBox::onActivated(int i)
 {
     // Send the event to the workbench group to delay the destruction of the emitting widget.
     int index = itemData(i).toInt();
-    WorkbenchActionEvent* ev = new WorkbenchActionEvent(this->actions().at(index));
+    auto *ev = new WorkbenchActionEvent(this->actions().at(index));
     QApplication::postEvent(this->group, ev);
     // TODO: Test if we can use this instead
     //QTimer::singleShot(20, this->actions()[i], SLOT(trigger()));
@@ -529,18 +529,18 @@ void WorkbenchGroup::addTo(QWidget *w)
 {
     refreshWorkbenchList();
     if (w->inherits("QToolBar")) {
-        QToolBar* bar = qobject_cast<QToolBar*>(w);
-        QComboBox* box = new WorkbenchComboBox(this, w);
+        auto *bar = qobject_cast<QToolBar *>(w);
+        QComboBox *box = new WorkbenchComboBox(this, w);
         box->setIconSize(QSize(16, 16));
         box->setToolTip(_action->toolTip());
         box->setStatusTip(_action->statusTip());
         box->setWhatsThis(_action->whatsThis());
         box->addActions(_group->actions());
-        connect(_group, SIGNAL(triggered(QAction*)), box, SLOT(onActivated (QAction*)));
+        connect(_group, SIGNAL(triggered(QAction *)), box, SLOT(onActivated(QAction *)));
         bar->addWidget(box);
     }
     else if (w->inherits("QMenu")) {
-        QMenu* menu = qobject_cast<QMenu*>(w);
+        auto *menu = qobject_cast<QMenu *>(w);
         menu = menu->addMenu(_action->text());
         menu->addActions(_group->actions());
     }
@@ -613,7 +613,7 @@ void WorkbenchGroup::refreshWorkbenchList()
 void WorkbenchGroup::customEvent( QEvent* e )
 {
     if (e->type() == QEvent::User) {
-        Gui::WorkbenchActionEvent* ce = (Gui::WorkbenchActionEvent*)e;
+        auto* ce = (Gui::WorkbenchActionEvent*)e;
         ce->action()->trigger();
     }
 }
@@ -826,8 +826,8 @@ void RecentFilesAction::restore()
         _group->addAction(QLatin1String(""))->setVisible(false);
     std::vector<std::string> MRU = hGrp->GetASCIIs("MRU");
     QStringList files;
-    for (std::vector<std::string>::iterator it = MRU.begin(); it!=MRU.end();++it)
-        files.append(QString::fromUtf8(it->c_str()));
+    for (auto & it : MRU)
+        files.append(QString::fromUtf8(it.c_str()));
     setFiles(files);
 }
 
@@ -969,9 +969,9 @@ void RecentMacrosAction::activateFile(int id)
     }
     else {
         if (QApplication::keyboardModifiers() == Qt::ShiftModifier){ //open for editing on Shift+click
-            PythonEditor* editor = new PythonEditor();
+            auto* editor = new PythonEditor();
             editor->setWindowIcon(Gui::BitmapFactory().iconFromTheme("applications-python"));
-            PythonEditorView* edit = new PythonEditorView(editor, getMainWindow());
+            auto* edit = new PythonEditorView(editor, getMainWindow());
             edit->setDisplayName(PythonEditorView::FileName);
             edit->open(filename);
             edit->resize(400, 300);
@@ -1212,7 +1212,7 @@ WindowAction::~WindowAction()
 
 void WindowAction::addTo ( QWidget * w )
 {
-    QMenu* menu = qobject_cast<QMenu*>(w);
+    auto* menu = qobject_cast<QMenu*>(w);
     if (!menu) {
         if (!_menu) {
             _menu = new QMenu();
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -208,7 +208,7 @@ FreeCADGui_subgraphFromObject(PyObject * /*self*/, PyObject *args)
     std::string vp = obj->getViewProviderName();
     SoNode* node = nullptr;
     try {
-        Base::BaseClass* base = static_cast<Base::BaseClass*>(Base::Type::createInstanceByName(vp.c_str(), true));
+        auto* base = static_cast<Base::BaseClass*>(Base::Type::createInstanceByName(vp.c_str(), true));
         if (base && base->getTypeId().isDerivedFrom(Gui::ViewProviderDocumentObject::getClassTypeId())) {
             std::unique_ptr<Gui::ViewProviderDocumentObject> vp(static_cast<Gui::ViewProviderDocumentObject*>(base));
             std::map<std::string, App::Property*> Map;
@@ -221,8 +221,8 @@ FreeCADGui_subgraphFromObject(PyObject * /*self*/, PyObject *args)
                 static_cast<App::PropertyPythonObject*>(pyproxy)->setValue(Py::Long(1));
             }
 
-            for (std::map<std::string, App::Property*>::iterator it = Map.begin(); it != Map.end(); ++it) {
-                vp->updateData(it->second);
+            for (const auto & it : Map) {
+                vp->updateData(it.second);
             }
 
             std::vector<std::string> modes = vp->getDisplayModes();
@@ -268,7 +268,7 @@ FreeCADGui_exportSubgraph(PyObject * /*self*/, PyObject *args)
     void* ptr = nullptr;
     try {
         Base::Interpreter().convertSWIGPointerObj("pivy.coin", "SoNode *", proxy, &ptr, 0);
-        SoNode* node = static_cast<SoNode*>(ptr);
+        auto* node = static_cast<SoNode*>(ptr);
         if (node) {
             std::string formatStr(format);
             std::string buffer;
@@ -695,10 +695,10 @@ void Application::exportTo(const char* FileName, const char* DocName, const char
             std::stringstream str;
             std::set<App::DocumentObject*> unique_objs;
             str << "__objs__=[]" << std::endl;
-            for (std::vector<App::DocumentObject*>::iterator it = sel.begin(); it != sel.end(); ++it) {
-                if (unique_objs.insert(*it).second) {
+            for (const auto &it : sel) {
+                if (unique_objs.insert(it).second) {
                     str << "__objs__.append(FreeCAD.getDocument(\"" << DocName << "\").getObject(\""
-                        << (*it)->getNameInDocument() << "\"))" << std::endl;
+                        << it->getNameInDocument() << "\"))" << std::endl;
                 }
             }
 
@@ -760,7 +760,7 @@ void Application::slotNewDocument(const App::Document& Doc, bool isMainDoc)
     std::map<const App::Document*, Gui::Document*>::const_iterator it = d->documents.find(&Doc);
     assert(it==d->documents.end());
 #endif
-    Gui::Document* pDoc = new Gui::Document(const_cast<App::Document*>(&Doc),this);
+    auto* pDoc = new Gui::Document(const_cast<App::Document*>(&Doc),this);
     d->documents[&Doc] = pDoc;
 
     // connect the signals to the application for the new document
@@ -779,7 +779,7 @@ void Application::slotNewDocument(const App::Document& Doc, bool isMainDoc)
 
 void Application::slotDeleteDocument(const App::Document& Doc)
 {
-    std::map<const App::Document*, Gui::Document*>::iterator doc = d->documents.find(&Doc);
+    auto doc = d->documents.find(&Doc);
     if (doc == d->documents.end()) {
         Base::Console().Log("GUI document '%s' already deleted\n", Doc.getName());
         return;
@@ -810,9 +810,9 @@ void Application::slotDeleteDocument(const App::Document& Doc)
 
 void Application::slotRelabelDocument(const App::Document& Doc)
 {
-    std::map<const App::Document*, Gui::Document*>::iterator doc = d->documents.find(&Doc);
+    auto doc = d->documents.find(&Doc);
 #ifdef FC_DEBUG
-    assert(doc!=d->documents.end());
+    assert(doc != d->documents.end());
 #endif
 
     signalRelabelDocument(*doc->second);
@@ -821,9 +821,9 @@ void Application::slotRelabelDocument(const App::Document& Doc)
 
 void Application::slotRenameDocument(const App::Document& Doc)
 {
-    std::map<const App::Document*, Gui::Document*>::iterator doc = d->documents.find(&Doc);
+    auto doc = d->documents.find(&Doc);
 #ifdef FC_DEBUG
-    assert(doc!=d->documents.end());
+    assert(doc != d->documents.end());
 #endif
 
     signalRenameDocument(*doc->second);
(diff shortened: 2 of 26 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7415">pipeline status</a> for feature branch [PR_7415](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7415). Pipeline [625382535](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/625382535) was triggered at [6d9a691](https://github.com/FreeCAD/FreeCAD/pull/7415/commits/6d9a6918766ac7e2578931a77d88f7f644367361). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

benj5378:
Why does it need to be auto?

berniev:
@benj5378 Part of a larger effort to update the code base closer to C++17 best practise.

benj5378:
> @benj5378 Part of a larger effort to update the code base closer to C++17 best practise.

Why is it better practice? Can you link any standard recommending it?

chennes:
A quick glance over these changes suggests this is mostly a "DRY" change -- if you already state the type on the line once, you shouldn't repeat yourself.

berniev:
Every time a type is typed there is a possibility of error, unnecessary extra cognitive load, duplication of effort for no actual gain (at least sine c++11 when auto was introduced).
Maybe these say it better than I ever could:
[microsoft](https://docs.microsoft.com/en-us/cpp/cpp/auto-cpp?view=msvc-170)
[cppreference](https://en.cppreference.com/w/cpp/language/auto)
[packt](https://subscription.packtpub.com/book/programming/9781786465184/1/ch01lvl1sec5/using-auto-whenever-possible)
[bogotobogo](https://www.bogotobogo.com/cplusplus/C11/C11_Type_Inference_auto_range_based_for_loop.php)

berniev:
@chennes I think it goes deeper than that. A fundamental and welcome change in the language at c++11 and improved through 14 and 17.

src/Gui/Action.cpp
```diff
@@ -529,18 +529,18 @@ void WorkbenchGroup::addTo(QWidget *w)
 {
     refreshWorkbenchList();
     if (w->inherits("QToolBar")) {
-        QToolBar* bar = qobject_cast<QToolBar*>(w);
-        QComboBox* box = new WorkbenchComboBox(this, w);
+        auto *bar = qobject_cast<QToolBar *>(w);
+        QComboBox *box = new WorkbenchComboBox(this, w);
```
0penBrain, line 533: Why are you randomly moving the star aside the type or aside the variable? I could understand such moves to achieve consistency, but it's not the case across this PR.

berniev: Thankyou OpenBrain. Within the FreeCAD code base the position of the * is not consistent. Of course c++ doesn't give a rat's bum where it is. I have been progressively using auto formatting in my IDE (Clion) using the clang-format file recently introduced into FreeCAD in an attempt to achieve consistency. Apologies if I havn't reformatted all touched code areas. If the question is whether the * should be to the left or the right by all means lets have the discussion to set the standard going forward. If there exists some FreeCAD coding guidelines please advise. I have not seen any.

X-Ryl669:
The main issue with a large C++ code base full of `auto` is that you don't know the type your are manipulating without digging into the code (which you might not have), example:
```
 auto f = some_func();
```
If you don't look at `some_func` signature, you don't know what `f` is, so you don't know if you can write `f->` or `f.`.
It increases the burden since you have to jump everywhere to find out what was this type. 
Sometimes you have an IDE that can resolve the type, but most of the time, you don't (or it's making errors).

Worst, the return type of `some_func` can be hidden also, so you might have to dig into a function you haven't written to understand what it does and whatsoever (if it's using `-> decltype(some expression)` for its declaration, good luck figuring out what it is).

`auto` is very useful when it's local (doesn't leak outside of a scope), or to write generic code (when dealing with STL's internal meta programming).  In this PR, I don't see why it's a benefit, since it's not coherent (why are you replacing `A* b = new A*` but not `C* c = (C*)ptr` ?) and it's partial (so you end up with a mix of different behavior, some functions with auto , some without in the code base).

src/Gui/Action.cpp
X-Ryl669: Why auto for the former and not the later type ? It there a reason using box as a `QComboBox` wouldn't work as a `WorkbenchComboBox` later on ? It doesn't seem consistent.

src/Gui/ApplicationPy.cpp
```diff
@@ -786,13 +784,11 @@ PyObject* Application::sExport(PyObject * /*self*/, PyObject *args)
 
 
             SoGetPrimitiveCountAction action;
-            action.setCanApproximate(true);
+            action.setCanApproximate(1);
```
X-Ryl669, line 787: Why change this ? The signature is a setCanApproximate(SbBool) so it clearly except a boolean (even if SbBool is an int, only 0 and 1 is allowed here, use C++'s true and false here)

src/Gui/Action.cpp
berniev: The first case states specifically object_cast<QToolBar*> so the type is obvious. No need to repeat (DRY).
The second case is new WorkbenchComboBox but the type is stated as QComboBox. Without knowing or caring why, there is a difference, so not a direct candidate for auto.

src/Gui/ApplicationPy.cpp
berniev: In the absence of any coding guidelines here it seemed a good idea to follow best practise warnings from clang-tidy. I can understand why that warning exists, but it also seems this is not a popular change here.

berniev:
@X-Ryl669 The SO page you refer to predates c++11 approval! Maybe look at the links I provided above.

What you say is largely a truism, but I disagree it applies here. If you have any more specific examples I'd be pleased to address them.

berniev:
Perhaps the seminal text re auto: [sutter
](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)

src/Gui/ApplicationPy.cpp
X-Ryl669: I'm not a FreeCAD developer. So this is my opinion here, I'll let the official dev kicks in to correct. Anyway, this code relies on OpenInventor library which is a mixed C & C++ library. So they "redefined" bool type since it doesn't exist in C90. They named it `SbBool` which is quite clear about their intend. 

When you rely on a code analyzer, it doesn't care about the semantic and figure out that SbBool is a int, and you are feeding a boolean to it. The code analyzer here is wrong, because, clearly, only value 0 and 1 are expected here (it wouldn't choke if you had written `setCanApproximate(42)`. You don't know inside the function if the code is doing `if (param == TRUE)` or `if (param)`. The former would fail to work so you have to dig into the code to know that (and since it's a dependency, it's not in FreeCAD code base, so it's hard to find, you might not have it on your computer).

It's a perfect example of what I've written earlier, this kind of change increases the complexity for the developer for no actual benefit. 

Reading the new line of code triggers the query "is 1 some kind of threshold ? What is the unit ?"
while you should ask "ok, I'm allowing to make approximation".

Anyway, it's my 2 pence, FreeCAD developers will decide what they prefer.

X-Ryl669:
Look, I don't disagree with you. I disagree with Herb Sutter which is, in my opinion, correct for what he does, but not for real world applications. Sure, in a STL only codebase, or a codebase that's truely generic (like a library) you should use auto everywhere you can. If you write a C++ application from scratch, only using generic C++ dependencies like STL or boost, then yes, he's completely right. 

But for a real world large application that has to deal with mixed C / C++ library (like FreeCAD), with binding between Python, Qt, OpenCascade, OpenInventor, etc.., you have ten different coding convention where **none** of them are generic (maybe Qt is the closest one to being generic).  

In all cases, his message isn't "use auto everywhere" (since it's impossible), but "using auto makes your code more generic/maintainable/beautiful". 

In this PR, you're using `auto *` for pointers but not for all of them. This is inconsistent. Either use `auto` for all variable (inc. pointers) but make sure to cast to the appropriate type. Then you'll realize that the DRY principle doesn't really apply:
`QComboBox * p = new WidgetCombo(x)` => `auto p = (QComboBox*)new WidgetCombo(x)` isn't any better IMHO.

What HS said here would  have been to write `auto p = WidgetCombo{x}` or `auto p = make_unique<WidgetComboBox>(x)` or even better `auto p = WidgetComboBox::make(x)`. You don't need to know what WidgetComboBox is here, and you *guess* that `make` is a static function that's creating a `unique_ptr<WidgetComboBox>`.  Please notice that all these rewrites change the type to manipulate from a `QComboBox *` to `unique_ptr<WidgetComboBox>` which isn't storable in a `QComboBox *` array anymore. Unless you use a `std::vector` and so on, you're STL'izing your complete codebase. 

See what I mean ? This is a lot more work, since Qt isn't written this way. 

Anyway, there are multiple place in this PR where `auto` make sense and code easier to read, but many where it's not the case too and I invite you to fix them, not by rewriting the lines where the variables are declared but the code that makes use of the variable so it can deal with auto. One example is for the `QComboBox*` above, where it likely possible that a `WidgetComboBox *` variable would work as well (so it avoid a useless cast).

src/Gui/ApplicationPy.cpp
berniev: As already stated, I understand the argument for allowing implicit conversions. I know Werner supports their use and has said that maybe clang-tidy settings should be changed. But that hasn't happened yet.

berniev:
auto * p = ptr and auto p = ptr are exactly the same. The * is entirely optional. Personally I think it is helpful to add the * to make it clearer its a ptr type. I don't catch them all in my changes and the existing code base uses auto * form rarely if at all. clang-tidy does suggest this. 

I don't understand your remark re new WidgetComboBox. If the programmer's intention is QComboBox type then that should be left well enough alone and not changed to auto, which I believe is what the pr does.

berniev:
@wwmayer @chennes  It seems this PR is not pleasing the Gods. If it is simply unacceptable, then please let me know and I
ll kill it and try according to your advice. I think the basic cause to introduce auto is worthwhile pursuing. Did I cloud the issue too much?

chennes:
Even Sutter agrees that "almost always auto" isn't quite right (though I can't find the video now, I thought it was last year's cppcon, but I must be misremembering). IMO, if your PR were limited to purely the DRY applications of AAA it would be a clear improvement. If the type is already completely 100% obvious from the rest of the line, then auto is great. If it's not completely clear to the reader what the variable type is, and it _matters_ (that is, it's not intended to be generic code that might be called with a bunch of different types), then auto isn't an improvement, and may even be a regression.

berniev:
@chennes I agree completely with what you say. I _**believe**_ my PR follows that rule. Every instance was evaluated manually (as are all my PR's). If I have failed to comply then I should be slapped, but there hasn't been a criticism yet that points to an example.

src/Gui/Action.cpp
```diff
@@ -826,8 +826,8 @@ void RecentFilesAction::restore()
         _group->addAction(QLatin1String(""))->setVisible(false);
     std::vector<std::string> MRU = hGrp->GetASCIIs("MRU");
     QStringList files;
-    for (std::vector<std::string>::iterator it = MRU.begin(); it!=MRU.end();++it)
-        files.append(QString::fromUtf8(it->c_str()));
+    for (auto & it : MRU)
```
chennes, line 829: Is there a reason not to use "const auto &" here? (I recognize the old code was also non-const, but as long as we're improving the line anyway...)

src/Gui/Action.cpp
```diff
@@ -247,7 +247,7 @@ void ActionGroup::addTo(QWidget *w)
     // and adding this action to the widget doesn't work.
     if (_dropDown) {
         if (w->inherits("QMenu")) {
-            QMenu *menu = new QMenu(w);
+            auto *menu = new QMenu(w);
```
chennes, line 250: I don't see any really compelling reason to keep the "*" here -- there's no conceivable way `new` is going to return a non-pointer type. I don't feel really strongly about this, but my overall feeling is that with auto we should use the simplest version of it possible, unless the addition provides some actual benefit (e.g. `const` or `&`).

src/Gui/Application.cpp
```diff
@@ -821,27 +821,27 @@ void Application::slotRelabelDocument(const App::Document& Doc)
...
+    assert(doc != d->documents.end());
 #endif
 
     signalRenameDocument(*doc->second);
 }
 
 void Application::slotShowHidden(const App::Document& Doc)
 {
-    std::map<const App::Document*, Gui::Document*>::iterator doc = d->documents.find(&Doc);
+    auto doc = d->documents.find(&Doc);
```
chennes, line 834: I think this type is too opaque, you have to know too much about `d->documents` -- I'd switch this one back.

src/Gui/Application.cpp
```diff
@@ -821,27 +821,27 @@ void Application::slotRelabelDocument(const App::Document& Doc)
...
+    assert(doc != d->documents.end());
 #endif
 
     signalShowHidden(*doc->second);
 }
 
 void Application::slotActiveDocument(const App::Document& Doc)
 {
-    std::map<const App::Document*, Gui::Document*>::iterator doc = d->documents.find(&Doc);
+    auto doc = d->documents.find(&Doc);
```
chennes, line 844: See previous comment.

src/Gui/Application.cpp
```diff
@@ -1117,27 +1117,25 @@ void Application::setActiveDocument(Gui::Document* pcDocument)
...
+Gui::Document *Application::getDocument(const char *name) const
 {
-    App::Document* pDoc = App::GetApplication().getDocument( name );
-    std::map<const App::Document*, Gui::Document*>::const_iterator it = d->documents.find(pDoc);
-    if ( it!=d->documents.end() )
-        return it->second;
-    else
-        return nullptr;
+    App::Document *pDoc = App::GetApplication().getDocument(name);
+    auto doc = d->documents.find(pDoc);
```
chennes, line 1127: See previous comment.

src/Gui/Application.cpp
```diff
@@ -1170,12 +1168,12 @@ void Application::detachView(Gui::BaseView* pcView)
 void Application::onUpdate()
 {
     // update all documents
-    std::map<const App::Document*, Gui::Document*>::iterator It;
-    for (It = d->documents.begin();It != d->documents.end();++It)
-        It->second->onUpdate();
+    std::map<const App::Document *, Gui::Document *>::iterator it;
```
chennes, line 1171: Is `it` getting used someplace I don't see here?

src/Gui/Application.cpp
```diff
@@ -1207,27 +1205,27 @@ void Application::updateActions(bool delay)
...
 {
     e->setAccepted(getMainWindow()->closeAllDocuments(false));
-    if(!e->isAccepted())
+    if (!e->isAccepted())
         return;
 
     // ask all passive views if closable
-    for (std::list<Gui::BaseView*>::iterator It = d->passive.begin();It!=d->passive.end();++It) {
-        e->setAccepted((*It)->canClose());
+    for (auto &it : d->passive) {
```
chennes, line 1215: Can this one be const?

src/Gui/Application.cpp
```diff
@@ -1914,8 +1914,8 @@ void Application::runApplication()
         // opens them
         QDir cwd = QDir::current();
         std::list<std::string> files = App::Application::getCmdLineFiles();
-        for (std::list<std::string>::iterator jt = files.begin(); jt != files.end(); ++jt) {
-            QString fn = QString::fromUtf8(jt->c_str(), static_cast<int>(jt->size()));
+        for (const auto &file : files) {
+            QString fn = QString::fromUtf8(file.c_str(), static_cast<int>(file.size()));
```
chennes, line 1918: Sort of an aside, but why in the world are we providing a size here? Feels more like an opportunity to introduce an error than to prevent one.

src/Gui/Application.cpp
```diff
@@ -2005,49 +2005,50 @@ void Application::runApplication()
...
+    else {
         mw.setWindowTitle(mainApp.applicationName());
     }
 
     QObject::connect(&mainApp, SIGNAL(messageReceived(const QList<QByteArray> &)),
                      &mw, SLOT(processMessages(const QList<QByteArray> &)));
 
     ParameterGrp::handle hDocGrp = WindowParameter::getDefaultParameter()->GetGroup("Document");
-    int timeout = hDocGrp->GetInt("AutoSaveTimeout", 15); // 15 min
+    int timeout = hDocGrp->GetInt("AutoSaveTimeout", 15);// 15 min
```
chennes, line 2031: This particular formatting change seems like noise to me. I'd omit it from future PRs, it just adds review overhead.

src/Gui/Application.cpp
```diff
@@ -2147,12 +2148,11 @@ void Application::runApplication()
     // Activate the correct workbench
     std::string start = App::Application::Config()["StartWorkbench"];
     Base::Console().Log("Init: Activating default workbench %s\n", start.c_str());
-    std::string autoload = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/General")->
-                           GetASCII("AutoloadModule", start.c_str());
+    std::string autoload = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/General")->GetASCII("AutoloadModule", start.c_str());
```
chennes, line 2151: I'm not thrilled with the line length here, I think we should pull a reference to the group on a separate line.

src/Gui/ApplicationPy.cpp
chennes: We can clean up clang-tidy later, let's drop this change from the PR.

src/Gui/ApplicationPy.cpp
```diff
@@ -786,13 +784,11 @@ PyObject* Application::sExport(PyObject * /*self*/, PyObject *args)
...
             SoGetPrimitiveCountAction action;
-            action.setCanApproximate(true);
+            action.setCanApproximate(1);
             action.apply(sep);
 
             bool binary = false;
-            if (action.getTriangleCount() > 100000 ||
-                action.getPointCount() > 30000 ||
-                action.getLineCount() > 10000)
+            if (action.getTriangleCount() > 100000 || action.getPointCount() > 30000 || action.getLineCount() > 10000)
```
chennes, line 791: I like the old formatting better.

src/Gui/ApplicationPy.cpp
```diff
@@ -1494,20 +1490,18 @@ PyObject* Application::sCreateViewer(PyObject * /*self*/, PyObject *args)
...
-    else if (num_of_views == 1) {
-        View3DInventor* viewer = new View3DInventor(nullptr, nullptr);
-        if (title)
-            viewer->setWindowTitle(QString::fromUtf8(title));
-        Gui::getMainWindow()->addWindow(viewer);
-        return viewer->getPyObject();
-    }
-    else {
-        SplitView3DInventor* viewer = new SplitView3DInventor(num_of_views, nullptr, nullptr);
+    if (num_of_views == 1) {
```
chennes, line 1493: This change is unrelated to using auto, and isn't an automated formatting change.

chennes (commented):
I didn't get through the whole thing, just the first couple files, so I've left some comments on those.

On the whole I am fine with the vast majority of the "auto" changes. I find the included unrelated formatting changes frustrating to review, but I got pretty good at ignoring them as I went through. But there are quite a few other changes being made to address linter concerns, all rolled into a PR that says it's about auto. IMO a PR of this size (+861/-800) needs to be a bit more focused, otherwise it's a lot of work to review.

berniev:
@chennes Thank you so much for having a look. Its disconcerting to hear nothing.

The takeaway for me is that there are no glaring mistakes but accept the drive-by format changes make review more difficult. 

In the context of achieving thousands of auto improvements, I'm sure a few 'could be const' would not be a show stopper.

The other take-away is the clang-format uncertainties that, if attended to, will make life better for everyone. In just a few short months we have come a long way with format and correctness tools.

I can't see a way to live edit a commit comparison so will start again and do a new PR.

chennes:
I think that is right. I personally especially prefer that the pure formatting changes be in their own commits so they can be added to a git-blame-ignore-revs file. Also, I can review pure formatting changes in a few minutes on a mobile device, whereas actual code changes need a bit more consideration.

berniev:
All my replies disppeared?

src/Gui/ApplicationPy.cpp
berniev: @wwmayer There are two available options for `readability-implicit-bool-conversion` [here](https://clang.llvm.org/extra/clang-tidy/checks/readability/implicit-bool-conversion.html) being: `
AllowIntegerConditions

    When true, the check will allow conditional integer conversions. Default is false.

AllowPointerConditions

    When true, the check will allow conditional pointer conversions. Default is false.
` Would false to the first, and true to the second be the way to go?

PR #7431, opened 2022-09-02
github: add a few more words for codespell to ignore

luzpaz:
This is specific to toponaming branch


luzpaz:
@realthunder apologies, I made a mistake with this PR. For some reason i didn't rebase it correctly, tried to correct it but didn't.

PR #7495, opened 2022-09-15
App: Use auto and range based for

berniev:

---

### DIFF
--- a/src/App/Application.cpp
+++ b/src/App/Application.cpp
@@ -503,7 +503,7 @@ Document* Application::newDocument(const char * Name, const char * UserName, boo
 
 bool Application::closeDocument(const char* name)
 {
-    map<string,Document*>::iterator pos = DocMap.find( name );
+    auto pos = DocMap.find( name );
     if (pos == DocMap.end()) // no such document
         return false;
 
@@ -550,9 +550,9 @@ App::Document* Application::getDocument(const char *Name) const
 
 const char * Application::getDocumentName(const App::Document* doc) const
 {
-    for (std::map<std::string,Document*>::const_iterator it = DocMap.begin(); it != DocMap.end(); ++it)
-        if (it->second == doc)
-            return it->first.c_str();
+    for (const auto & it : DocMap)
+        if (it.second == doc)
+            return it.first.c_str();
 
     return nullptr;
 }
@@ -560,8 +560,8 @@ const char * Application::getDocumentName(const App::Document* doc) const
 std::vector<App::Document*> Application::getDocuments() const
 {
     std::vector<App::Document*> docs;
-    for (std::map<std::string,Document*>::const_iterator it = DocMap.begin(); it != DocMap.end(); ++it)
-        docs.push_back(it->second);
+    for (const auto & it : DocMap)
+        docs.push_back(it.second);
     return docs;
 }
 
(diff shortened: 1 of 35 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7495">pipeline status</a> for feature branch [PR_7495](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7495). Pipeline [640878736](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/640878736) was triggered at [df32a91](https://github.com/FreeCAD/FreeCAD/pull/7495/commits/df32a91aec2a14c7dae38f96ab843137e359179b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
I have learned from my previous review experience: could you please re-submit without combining the two things? Make one PR for the ranged for loops, and one for the auto? (Actually, the auto may itself need to be two -- it would be easier to review if one of the PRs were truly all 100% single-line DRY things, and the other were the slightly more complex cases that require multiple lines of context to determine the type)

berniev:
Sure thing. But can we clarify a few things first?

1. Looking at existing code base, senior devs often apply auto - inside a class - to class variables. e.g:
   `740: auto it = _pendingDocMap.find(name);` in `src/App/Application.cpp` with 
   `607: std::map<std::string,std::vector<std::string> > _pendingDocMap;` in `Application.h`
   I think these are OK, but conflict with our discussions.

1a. Similar to 1. (above) but regards the result of a function call. e.g:
 `763: auto doc = openDocumentPrivate(path, name.c_str(), label, isMainDoc, createView, std::move(objNames));`

3. There are cases in the code where an iterator has been used, but const_iterator could be used instead, but to change it might be considered beyond the scope of a refactoring PR.

4. In some cases making range-based loop from a auto'd loop that has been set to cbegin() and cend(), my IDE (CLion) pushes `auto var`, whilst in others it happily makes `const auto & var`. I'm struggling to understand **_why_** (or is IDE just stupid), but I have read that in some cases (int, short string, even pointers?) and depending on what is done in the loop, this actually **_saves_** some copying and hence is to be preferred. I ran a further tidy run for things that could be const and it left them alone. I see many quite recent cases in the existing codebase by senior devs that are not so fussy about const-ness on auto. 
 
Confusing.

chennes:
Sure thing: in order...
1. First, the core developers do not all agree perfectly about `auto`: I prefer more of it, and Werner less, for example. Your example shows a case where I believe we do all agree, however, and that is with C++'s (extremely verbose) iterator types. For other uses, though, it's a tough sell to suggest a sweeping, multi-thousand line code change that has only lukewarm support from reviewers (and in some cases active opposition), and only marginal benefit to the codebase, considering how much time it takes to review those changes. A case like 1a is much harder to reach consensus on, so a big change to that sort of thing is probably a non-starter. I'd be fine with that line, but I suspect Werner would not, for example.
3. I'd love to see whole series of PRs dedicated entirely to `const`-correctness. It is far, far easier to review highly focused refactoring changes, so don't combine those with anything else.
4. Your suspicion about why there is sometimes that difference in the loop construct (that sometimes it's cheaper to make the copy) is correct, or at least it is often _true_ for small POD types. Whether that's what your IDE is doing or not I don't know, but CLion has a very good reputation, so I'd guess so.

So it boils down to this: huge changes with low perceived benefit basically have to be non-controversial, and to make reviewing easier (and more likely to actually happen) they shouldn't be mixed together. Every time I have to "context switch" when reviewing slows me down, and on a 2000 line PR that really matters. You can see the timestamps on that last review: it literally took me all day (I mean, I had customers, and taught a class, etc., but it was still a lot of code to look at).

berniev:
It might be unfair to speak for Werner. He writes heaps of auto, and even exactly like 1a. See Application.cpp just six months ago:
`654: auto docs = openDocuments(filenames, nullptr, nullptr, nullptr, createView);`

PR #7506, opened 2022-09-17
Draft: Fix several path array issues

Roy-043:
This PR fixes several issues:
* The hiding/unhiding of properties was not always correct.
* "Align Mode" Frenet did not ignore "Force Vertical" and "Vertical Vector".
* A binormal vector would be calculated from two vectors that were effectively parallel.
* The "Extra Translation" property, which should be applied to each element in the array, was applied to the array as a whole.

Remarks:
* The `placements_on_path` functions calculates all placements in a single loop. This ensues that the elements in the array are created in a logical order. This is important as a link array can be expanded. The element at the end of an open path would be the 2nd element in the old code.
* The `calculate_placement` function was changed to make better use of `App.Rotation()`. There is no need to calculate a binormal vector. Note that the main alignment options remain unchanged. But some problematic placements are handled differently. Where the old code would return a placement with a default rotation, the new code will use the available tangent vector. For an example of such a problematic placement see the first element in PathArray001 in the file attached to the 2nd forum topic. For each problematic placement there is still a warning.

Forum topics:
https://forum.freecadweb.org/viewtopic.php?f=23&t=59174
https://forum.freecadweb.org/viewtopic.php?f=23&t=59198
https://forum.freecadweb.org/viewtopic.php?f=23&t=59222

---

### DIFF
--- a/src/Mod/Draft/draftobjects/patharray.py
+++ b/src/Mod/Draft/draftobjects/patharray.py
@@ -336,34 +336,39 @@ class PathArray(DraftLink):
         self.show_and_hide(obj, prop)
 
     def show_and_hide(self, obj, prop):
-        """Show and hide the properties depending on the touched property."""
-        # The minus sign removes the Hidden property (show)
-        if prop == "Align":
+        """Show and hide the properties depending on the touched property.
+
+        Note that when the array is created, some properties will change
+        more than once in a seemingly random order.
+        """
+        # The minus sign removes the Hidden property (show).
+        if prop in ("Align", "AlignMode"):
+
+            # Check if all referenced properties are available:
+            for pr in ("Align", "AlignMode", "ForceVertical",
+                       "VerticalVector", "TangentVector"):
+                if not hasattr(obj, pr):
+                    return
+
             if obj.Align:
-                for pr in ("AlignMode", "ForceVertical", "VerticalVector",
-                           "TangentVector"):
-                    obj.setPropertyStatus(pr, "-Hidden")
+                obj.setPropertyStatus("AlignMode", "-Hidden")
+
+                if obj.AlignMode == "Frenet":
+                    for pr in ("ForceVertical", "VerticalVector"):
+                        obj.setPropertyStatus(pr, "Hidden")
+                else:
+                    for pr in ("ForceVertical", "VerticalVector"):
+                        obj.setPropertyStatus(pr, "-Hidden")
+
+                if obj.AlignMode == "Tangent":
+                    obj.setPropertyStatus("TangentVector", "-Hidden")
+                else:
+                    obj.setPropertyStatus("TangentVector", "Hidden")
+
             else:
-                for pr in ("AlignMode", "ForceVertical", "VerticalVector",
-                           "TangentVector"):
+                for pr in ("AlignMode", "ForceVertical",
+                           "VerticalVector", "TangentVector"):
                     obj.setPropertyStatus(pr, "Hidden")
-
-        if prop == "AlignMode":
-            if obj.AlignMode == "Original":
-                for pr in ("ForceVertical", "VerticalVector"):
-                    obj.setPropertyStatus(pr, "-Hidden")
-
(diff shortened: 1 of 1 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7506">pipeline status</a> for feature branch [PR_7506](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7506). Pipeline [642980502](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/642980502) was triggered at [9142d4a](https://github.com/FreeCAD/FreeCAD/pull/7506/commits/9142d4ac5eae91a42d46b6a006522f2f417645b4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Roy-043:
@donovaly This PR significantly changes portions of the code. I'd like to see this thoroughly tested during the V0.21/V1.0 development cycle. So please do not backport.

Overruler:
Looks good but you should apply a change in tangent vector to ExtraTranslation as well, like is done with final_rotation. I have this code that I've been testing:

```
        base_xlate = obj.ExtraTranslation
        final_xlate = base_xlate

        if (obj.Align and obj.AlignMode == "Tangent"
                and hasattr(obj, "TangentVector")):
            Xaxis = App.Vector(1.0, 0.0, 0.0)  # default TangentVector

            if not DraftVecUtils.equals(Xaxis, obj.TangentVector):
                # make rotation from TangentVector to X
                pre_rotation = App.Rotation(obj.TangentVector, Xaxis)
                final_rotation = base_rotation.multiply(pre_rotation)
                # apply rotation to ExtraTranslation
                final_xlate = pre_rotation.multVec(base_xlate)
```

Roy-043:
Thanks for looking at this, and good catch! This makes more sense. Please submit a PR.

Personally I find both the TangentVector and the ExtraTranslation only semi useful. They make sense in a 2D model, but less so in a 3D context. And you can achieve the same effect by nesting the Base object in a Std_Part.

If others want to check the fix: the attached zip contains the revised patharray.py file and a test file (PathArray_ExtraTranslation.FCStd).
[patharray.zip](https://github.com/FreeCAD/FreeCAD/files/9863247/patharray.zip)

Minor detail: the `base_xlate` variable can be avoided:
`final_xlate = obj.ExtraTranslation`
and
`final_xlate = pre_rotation.multVec(final_xlate)`

Roy-043:
It is also possible to solve this by changing the `calculate_placement` function instead. That solution is perhaps preferable?

https://github.com/FreeCAD/FreeCAD/blob/71d0490dab51c10e0cc99ddd6cf31156b5d2dd07/src/Mod/Draft/draftobjects/patharray.py#L491
Becomes:
```
    placement.Base = RefPt + placement.Rotation.multVec(xlate)
```

https://github.com/FreeCAD/FreeCAD/blob/71d0490dab51c10e0cc99ddd6cf31156b5d2dd07/src/Mod/Draft/draftobjects/patharray.py#L552-L554
Becomes:
```
    placement.Rotation = newRot.multiply(globalRotation)
    placement.Base = RefPt + placement.Rotation.multVec(xlate)
```

Overruler:
Changing `calculate_placement` instead does seem cleaner to me, yes.

My motivation to look into Path Link Array came from adding new properties that'd let me control exactly where copies are placed. For example: to place 85, 5, and 10 copies within the half-open, closed, and open intervals `[`0,850mm`)`, `[`850,900mm`]`, `(`900,1000mm`)`, respectively, by using 3 Path Link Arrays on one closed path that's itself the result of an offset operation and whose length could thus be something like 1004.932mm, which isn't a nice round number.

I lack an intuition on what the math to move between local and global coordinate frames should look like, so I just tried things until I got a Box primitive centered and aligned to the path 😃

PR #7478, opened 2022-09-12
[For testing] Sketcher: optionally merge 'coincident' within 'point-on-object' constraint

0penBrain:
The merge is activated with `User parameter:BaseApp/Preferences/Mod/Sketcher/General/MergePointOnObjectWithCoincident` boolean switch

The PR is essentially done to have a test between 2 flavors (merged or not) and decide what is to be kept.

The "feature" is only the last commit. 3 first commits are independent fixes and improvements.

PR #7522, opened 2022-09-24
Sketcher: various fixes in constraint commands

0penBrain:
This PR has been extracted from #7478 

Commit 1 : in current state, even if condition aren't OK to apply 'point-on-object' (for example selected vertex belongs to selected curve) we try to substitute constraints (typically coincident + curve/curve tangency => smooth joint). This can lead to strange effect, like "command not working" but still "something else happens". Commit prevent trying constraint substitution if conditions aren't OK for the command to process

Commit 2 : fixes an indentation problem that makes code hard to understand properly

Commit 3 : improves command (undo transaction) management. In current code, constraint substitution in 'coincident' generates its own transaction. With commit it no more does and delegate the transaction management to the calling method, which is more optimal and also consistent with what happens in 'point-on-object' constraint. This change was also needed for the last commit of #7478 to work.

@donovaly here it is :)

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -465,10 +465,10 @@ void SketcherGui::doEndpointTangency(Sketcher::SketchObject* Obj,
             std::swap(PosId1,PosId2);
         }
         // GeoId1 is the B-spline now
-        } // end of code supports simple B-spline endpoint tangency
+    } // end of code supports simple B-spline endpoint tangency
 
-        Gui::cmdAppObjectArgs(Obj, "addConstraint(Sketcher.Constraint('Tangent',%d,%d,%d,%d)) ",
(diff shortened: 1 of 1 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7522">pipeline status</a> for feature branch [PR_7522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7522). Pipeline [649556040](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/649556040) was triggered at [d9c2efc](https://github.com/FreeCAD/FreeCAD/pull/7522/commits/d9c2efccc02020036a462c6c19236f36d816b58f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks!
I am going to backport the first two then.

PR #7524, opened 2022-09-25
Fix for boost with C++20

berniev:
I found this necessary to build FC with C++20.
If nothing else, maybe useful later?

### DIFF
--- a/cMake/FreeCAD_Helpers/ConfigureCMakeVariables.cmake
+++ b/cMake/FreeCAD_Helpers/ConfigureCMakeVariables.cmake
@@ -34,4 +34,7 @@ macro(ConfigureCMakeVariables)
     message(STATUS "includedir: ${CMAKE_INSTALL_INCLUDEDIR}")
     message(STATUS "libdir: ${CMAKE_INSTALL_LIBDIR}")
     message(STATUS "cmake: ${CMAKE_VERSION}")
+
+    add_compile_definitions(BOOST_NO_CXX98_FUNCTION_BASE)
+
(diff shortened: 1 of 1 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7524">pipeline status</a> for feature branch [PR_7524](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7524). Pipeline [649930800](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/649930800) was triggered at [658c7ee](https://github.com/FreeCAD/FreeCAD/pull/7524/commits/658c7ee449253199fd864a4e41b90ce635e844b4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
https://www.boost.org/doc/libs/1_69_0/libs/config/doc/html/boost_config/boost_macro_reference.html

BOOST_NO_CXX98_FUNCTION_BASE:
> The standard library no longer supports std::unary_function and std::binary_function. They were deprecated in C++11 and is removed from C++14.

So, should be safe to merge it.

wwmayer:
The more suitable place for this define is the CompilerChecksAndSetups macro. Then in order to use add_compile_definitions CMake 3.12 is required which is not yet available on all supported platforms.

See 608fb15d40

berniev:
I'll take your word for it, sir!

PR #7529, opened 2022-09-26
App and Addon Manager: Add min python version tag

chennes:
Add support for a `<pythonmin>` tag in the package.xml metadata file, and add auto detection of the required Python version to Addon Manager Developer Mode.


luzpaz:
Downstream related ticket: https://github.com/FreeCAD/FreeCAD-addons/issues/215

PR #7531, opened 2022-09-27
Remove redundant reference to C++14

berniev:
C++14 has been previously deprecated for FreeCAD

### DIFF
--- a/cMake/FreeCAD_Helpers/CompilerChecksAndSetups.cmake
+++ b/cMake/FreeCAD_Helpers/CompilerChecksAndSetups.cmake
@@ -46,8 +46,6 @@ macro(CompilerChecksAndSetups)
         set(CMAKE_CXX_STANDARD 20)
     elseif(${BUILD_ENABLE_CXX_STD} MATCHES "C\\+\\+17")
         set(CMAKE_CXX_STANDARD 17)
-    elseif(${BUILD_ENABLE_CXX_STD} MATCHES "C\\+\\+14")
-        set(CMAKE_CXX_STANDARD 14)
     endif()
 
     # Log the compiler and version


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7531">pipeline status</a> for feature branch [PR_7531](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7531). Pipeline [651498186](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/651498186) was triggered at [5894ec7](https://github.com/FreeCAD/FreeCAD/pull/7531/commits/5894ec785a868a61e03c7a64aa9c96a23e373ff8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
@wwmayer wasn't there some random system out there that we were keeping this around for? Some flavor of MacOS maybe? I don't know if it's still relevant.

wwmayer:
> wasn't there some random system out there that we were keeping this around for? Some flavor of MacOS maybe?

I am not aware of such a situation. And since we already use some C++17 features I can hardly imagine that with the global C++14 option FreeCAD still builds.

However, what you probably mean is that for some modules (Path, Fem, ...) we use external libraries that do not support C++17 yet so that for these modules C++14 must be activated. Usually we do this inside the CMakeLists.txt files with:
```
    set_target_properties(Path PROPERTIES CXX_STANDARD_REQUIRED ON)
    set_target_properties(Path PROPERTIES CXX_STANDARD 14)
```

PR #7536, opened 2022-09-29
Cmake Final Report changes

berniev:
Make the final report code a bit cleaner, a bit simpler and introducing baby steps to Qt5 to Qt6 changeover.

In the codebase there is lots of hard-coding variables into names e.g. python3, pyside2, qt5 etc etc which can be counter-productive.


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7536">pipeline status</a> for feature branch [PR_7536](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7536). Pipeline [654005180](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/654005180) was triggered at [2b78a6f](https://github.com/FreeCAD/FreeCAD/pull/7536/commits/2b78a6f667ce4a9e5c039fa702fae8afeb4864f2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
Have you dropped the output of Qt5XmlPatterns on purpose? Note, in Qt5 the modules QtXml and QtXmlPatterns are two different things. In Qt6 it has been dropped.

In Qt5 a module Qt5Core5Compat doesn't exist so it's pointless to print a version number.

Qt WinExtras is a Windows-specific extension module. So, for UNIX-like systems it shouldn't be printed and for Windows only if it was found. And according to this table https://doc.qt.io/qt-6/qtmodules.html it's not available any more for Qt6.

berniev:
1. Re xmlPatterns. Logic was if its not there it will just not show, no big deal? 
2. Qt5CoreCompat. Of course! Will correct later if that's ok.
3. QtWinExtras so will will just show blank. No big deal??

Trying to establish a simplified report that can be in current master and work for current master and qt6 experimental branch (like I have). Rightly or wrongly I figured the current is a little OTT in some (but not all) areas.

wwmayer:
> 2. ...

I think we can completely drop the Compat stuff. In the forum I wrote that we only use two classes QTextCodec and QRegExp.
So, instead of using this module we should directly port the code to Qt6

> 3. ...

QtWinExtras is used optionally and it would be fine if you re-added `if (${Qt5WinExtras_FOUND})`.

PR #7539, opened 2022-09-30
[Bugfix]Gui: TinkerCAD mouse mode was always stopping RMB propagation

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=3&t=72174

Bugs comes from TinkerCAD mode using RMB dragging to rotate view.
So it has to itself handle when a user makes a short "single" click with RMB.
There was a problem with how conditions was nested, so actually it was always catching the mouse event -- stopping propagation -- even when the viewer was in Edit mode (which it shouldn't).
This for example prevents Sketcher tools to be exited with RMB.
The PR brings a correct behavior in case of short "single" RMB click : if viewer is not in Edit mode, the event is caught and processed ; if viewer is in Edit mode, the event is ignored and let free to propagate.

### DIFF
--- a/src/Gui/TinkerCADNavigationStyle.cpp
+++ b/src/Gui/TinkerCADNavigationStyle.cpp
@@ -114,7 +114,7 @@ SbBool TinkerCADNavigationStyle::processSoEvent(const SoEvent * const ev)
         const auto event = (const SoMouseButtonEvent *) ev;
         const int button = event->getButton();
         const SbBool press = event->getState() == SoButtonEvent::DOWN ? true : false;
-        SbBool canOpenPopupMenu = false;
+        SbBool shortRMBclick = false;
 
         switch (button) {
         case SoMouseButtonEvent::BUTTON1:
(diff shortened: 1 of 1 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7539">pipeline status</a> for feature branch [PR_7539](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7539). Pipeline [655090203](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/655090203) was triggered at [01570dd](https://github.com/FreeCAD/FreeCAD/pull/7539/commits/01570ddf84620044a5af59e86db8f972f485cb3a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Good catch. I encountered the issue too, thought it was more fundamental, not only in Tinker style.

PR #7543, opened 2022-10-02
[Base, Sketcher] remove redundant boolean checks

donovaly:
As the title says.

If someone could re-check, this would be nice.

### DIFF
--- a/src/Base/StackWalker.cpp
+++ b/src/Base/StackWalker.cpp
@@ -261,7 +261,7 @@ public:
     // SymInitialize
     if (szSymPath != NULL)
       m_szSymPath = _strdup(szSymPath);
-    if (this->pSI(m_hProcess, m_szSymPath, FALSE) == FALSE)
+    if (!this->pSI(m_hProcess, m_szSymPath, FALSE))
       this->m_parent->OnDbgHelpErr("SymInitialize", GetLastError(), 0);
       
     DWORD symOptions = this->pSGO();  // SymGetOptions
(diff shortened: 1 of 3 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7543">pipeline status</a> for feature branch [PR_7543](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7543). Pipeline [656391490](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/656391490) was triggered at [634f928](https://github.com/FreeCAD/FreeCAD/pull/7543/commits/634f928699a427852fcd61252ca196dce853d53b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes (approved):
I think I checked all of them and they are all correct -- the formatting changes mixed in with real changes make it a little tricky to be sure, though. Also, I note in StackWalker.cpp a lot of comparisons to NULL -- do you plan on making the same change to those?

donovaly:
Thanks for the review!

> the formatting changes mixed in with real changes make it a little tricky

I know. These changes are made automatically by Visual Studio's auto clang formatting. Since this uses our clang formatting, I leave those but sure, for a review this is not optimal.

> Also, I note in StackWalker.cpp a lot of comparisons to NULL -- do you plan on making the same change to those?

These comparisons are in many places in the code. I don't know yet if we can handle them like we do for nullptr. Can we, meaning would this be safe for all compilers?

PR #7547, opened 2022-10-03
Improve text color management in Report view + Python console + Macro editor

0penBrain:
Both problems solved here come when you apply an OS native dark theme -- keeping default FC theme --.
In this case you are provided with a dark background for widgets, but it's not ideally managed in FC.

1st commit solves a problem that when you have a blank config, the color of the normal messages color in the Report view is chosen by Qt. With a dark OS theme, this text turns to white -- good --. The problem is that when opening the Preferences editor, the default value is always initialized to black, so if you save your preferences (for any reason), you will now get black text over dark background, which is unreadable. With this commit, the default value in the Preferences editor is initialized with the current text color used in the FC app.

2nd commit solves another similar problem existing in Python console & Macro editor. In these 2 widgets, the default color for code instructions is always black, making it unreadable on dark themes. With this commit, the color defaults to the one currently used in the FC app for normal texts.

### DIFF
--- a/src/Gui/SyntaxHighlighter.cpp
+++ b/src/Gui/SyntaxHighlighter.cpp
@@ -22,6 +22,9 @@
 
 #include "PreCompiled.h"
 
+#include <QApplication>
+#include <QPalette>
+
 #include "SyntaxHighlighter.h"
 
 
@@ -33,11 +36,16 @@ class SyntaxHighlighterP
 public:
     SyntaxHighlighterP()
     {
-        cNormalText.setRgb(0, 0, 0); cComment.setRgb(0, 170, 0);
-        cBlockcomment.setRgb(160, 160, 164); cLiteral.setRgb(255, 0, 0);
-        cNumber.setRgb(0, 0, 255); cOperator.setRgb(160, 160, 164);
-        cKeyword.setRgb(0, 0, 255); cClassName.setRgb(255, 170, 0);
-        cDefineName.setRgb(255, 170, 0); cOutput.setRgb(170, 170, 127);
+        cNormalText = qApp->palette().windowText().color();
+        cComment.setRgb(0, 170, 0);
+        cBlockcomment.setRgb(160, 160, 164);
+        cLiteral.setRgb(255, 0, 0);
+        cNumber.setRgb(0, 0, 255);
+        cOperator.setRgb(160, 160, 164);
+        cKeyword.setRgb(0, 0, 255);
+        cClassName.setRgb(255, 170, 0);
+        cDefineName.setRgb(255, 170, 0);
+        cOutput.setRgb(170, 170, 127);
         cError.setRgb(255, 0, 0);
     }
 
--- a/src/Gui/DlgEditorImp.cpp
+++ b/src/Gui/DlgEditorImp.cpp
@@ -68,7 +68,7 @@ DlgSettingsEditorImp::DlgSettingsEditorImp( QWidget* parent )
 
     d = new DlgSettingsEditorP();
     QColor col;
-    col = Qt::black;
+    col = qApp->palette().windowText().color();
     unsigned int lText = (col.red() << 24) | (col.green() << 16) | (col.blue() << 8);
     d->colormap.push_back(QPair<QString, unsigned int>
         (QString::fromLatin1(QT_TR_NOOP("Text")), lText));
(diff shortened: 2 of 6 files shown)


0penBrain:
1st note : could be a good improvement that for these 2 preferences (basically "normal text"), there could be an "Auto" setting that would just use the default color set by Qt, so it will always adjust to theme even is user changes it over time.

0penBrain:
2nd note : during this modification, I found the color map for Python highlighting to be defined 4 times : 
 - https://github.com/FreeCAD/FreeCAD/blob/f827de4d4d03d5e1df3e57b3cf40bd8865c100f3/src/Gui/DlgEditorImp.cpp#L71-L130
 - https://github.com/FreeCAD/FreeCAD/blob/f827de4d4d03d5e1df3e57b3cf40bd8865c100f3/src/Gui/PythonConsole.cpp#L103-L116
 - https://github.com/FreeCAD/FreeCAD/blob/f827de4d4d03d5e1df3e57b3cf40bd8865c100f3/src/Gui/TextEdit.cpp#L198-L212
 - https://github.com/FreeCAD/FreeCAD/blob/f827de4d4d03d5e1df3e57b3cf40bd8865c100f3/src/Gui/SyntaxHighlighter.cpp#L36-L41

This is really tough to maintain.
@wwmayer what would you think of centralizing the definition for example in the `SyntaxHighlighter` then provide in this same class some functions that other classes may call to get required data ?

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7547">pipeline status</a> for feature branch [PR_7547](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7547). Pipeline [657149918](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/657149918) was triggered at [0b373a1](https://github.com/FreeCAD/FreeCAD/pull/7547/commits/0b373a15af509f02e964ca1fcd8ff08627972a8d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Do you want me to hold off on merging this while discussing your proposed improvements?

src/Gui/SyntaxHighlighter.cpp
```diff
@@ -33,12 +36,12 @@ class SyntaxHighlighterP
...
-        cNumber.setRgb(0, 0, 255); cOperator.setRgb(160, 160, 164);
-        cKeyword.setRgb(0, 0, 255); cClassName.setRgb(255, 170, 0);
-        cDefineName.setRgb(255, 170, 0); cOutput.setRgb(170, 170, 127);
-        cError.setRgb(255, 0, 0);
+        cNormalText = qApp->palette().windowText().color();
+        cComment.setRgb(0, 170, 0); cBlockcomment.setRgb(160, 160, 164);
+        cLiteral.setRgb(255, 0, 0); cNumber.setRgb(0, 0, 255);
+        cOperator.setRgb(160, 160, 164); cKeyword.setRgb(0, 0, 255);
+        cClassName.setRgb(255, 170, 0); cDefineName.setRgb(255, 170, 0);
+        cOutput.setRgb(170, 170, 127); cError.setRgb(255, 0, 0);
```
chennes, line 44: While I appreciate the effort to maintain the original formatting (which normally makes reviews easier and the code more consistent), in this case since you're already touching every line to rearrange for line length, I'd seriously consider reformatting to give each their own line.

0penBrain: I hesitated actually. Will push an amended commit on tomorrow.

0penBrain:
> Do you want me to hold off on merging this while discussing your proposed improvements?

You can merge as soon as these commits are OK.  I will eventually code refactoring in another PR (this may take some time).

src/Gui/SyntaxHighlighter.cpp
0penBrain: Done

0penBrain:
@chennes ready to be reviewed/merged on my side

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7547">pipeline status</a> for feature branch [PR_7547](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7547). Pipeline [657631520](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/657631520) was triggered at [3da6cca](https://github.com/FreeCAD/FreeCAD/pull/7547/commits/3da6ccac28aef5d515f7b338fb6a0ed0cca7e376). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #7559, opened 2022-10-05
[Bugfix]Base: Python method 'isEqual' of Vector shall be Const

0penBrain:
https://forum.freecadweb.org/viewtopic.php?t=72287


luzpaz:
@0penBrain while you're in this file would you mind also clearing the trailing whitespaces in it as well ? :stuck_out_tongue_closed_eyes:

0penBrain:
@luzpaz Done.

PR #7561, opened 2022-10-05
Gui: optionally disable wheel event when Gui::InputField doesn't have focus

0penBrain:
Reuse "ComboBoxWheelEventFilter" switch that already acts on combo and spin boxes

https://forum.freecadweb.org/viewtopic.php?f=10&t=30772&p=630804#p630788

### DIFF
--- a/src/Gui/InputField.cpp
+++ b/src/Gui/InputField.cpp
@@ -77,7 +77,12 @@ InputField::InputField(QWidget * parent)
     SaveSize(5)
 {
     setValidator(new InputValidator(this));
-    setFocusPolicy(Qt::WheelFocus);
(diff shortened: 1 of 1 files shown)


src/Gui/InputField.cpp
```diff
@@ -699,6 +704,9 @@ void InputField::keyPressEvent(QKeyEvent *event)
 
 void InputField::wheelEvent (QWheelEvent * event)
 {
+    if (!(hasFocus() || (focusPolicy() & 0x4)))
```
chennes, line 707: What's the magic number?

0penBrain: Because it doesn't work. :) 

However let me have another look. Maybe using focus is enough that will make things more readable.

0penBrain: OK, I checked and `hasFocus()` is enough. ;)

chennes:
Thanks for checking! Merging now...

PR #7417, opened 2022-08-30
Sketcher : allow to select points & other elements from the backside view

0penBrain:
https://forum.freecadweb.org/viewtopic.php?f=3&t=71482

This achieved by a camera sensor at viewprovider level that detects side change, inverse all Z layer values used for rendering and redraws.

PR #7618, opened 2022-10-18
[Bugfix]Sketcher/BackEdit : Backedit lost if user changed camera mode

0penBrain:
- in current implementation, backedit feature is lost if user changes camera mode (for example from orthographic to perspective, or oppositely). This happens because when changing camera mode, Coin actually do not just change a parameter of the camera node, but delete the current camera node and add a new one to the scenegraph. Thus the SoFieldSensor currently used (attached to camera orientation) is just detached and sensing is left -- notice is would be same with a SoNodeSensor attached to camera node --. The solution is to attach the sensor at scenegraph root node so it isn't lost when camera mode changed. No pre-check (such as checking if camera node changed since last call) has been added currently because it would not actually reduce callback execution time (and even worsen it).

### DIFF
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -3157,13 +3157,15 @@ void ViewProviderSketch::setEditViewer(Gui::View3DInventorViewer* viewer, int Mo
 
     viewer->setupEditingRoot();
 
-    cameraSensor.setData(new VPCam{this, viewer->getSoRenderManager()->getCamera()});
-    cameraSensor.attach(&viewer->getSoRenderManager()->getCamera()->orientation);
+    cameraSensor.setData(new VPRender{this, viewer->getSoRenderManager()});
+    cameraSensor.attach(viewer->getSoRenderManager()->getSceneGraph());
 }
 
 void ViewProviderSketch::unsetEditViewer(Gui::View3DInventorViewer* viewer)
 {
-    delete static_cast<VPCam*>(cameraSensor.getData());
+    auto dataPtr = static_cast<VPRender*>(cameraSensor.getData());
+    dataPtr->attached = false;
+    delete dataPtr;
     cameraSensor.detach();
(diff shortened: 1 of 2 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7618">pipeline status</a> for feature branch [PR_7618](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7618). Pipeline [670114980](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/670114980) was triggered at [ed5b6cb](https://github.com/FreeCAD/FreeCAD/pull/7618/commits/ed5b6cb30f53094555aadee265432510b36c1f04). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
> Sorry to be stupid here - what is "backedit"?

You're not stupid, probably it's not so obvious if you don't know. This is name I gave to feature introducing the ability to seamlessly edit a sketch from its back side. As usual, there were some problems in the initial implementation. :)

abdullahtahiriyo:
I have tested the PR. The issue is fixed. The changes with respect to master are reasonable to me.

0penBrain:
@abdullahtahiriyo Thx. While you're at this and if you have some time, maybe you can have a look at the PR #7417 that introduced the feature. Maybe you can find other flaws or suboptimal things by reviewing it.
You'll see there that it introduces `viewOrientationFactor` member variable to `ViewProviderSketch` class. Currently it's only an int toggling between +1 and -1 depending on the side, and that is then use to invert all offsets used for sketcher elements display.
My further hope is that it can later be moved to a float that will allow a finer control of these offsets taking into account camera zoom and skew -- angle between sketch plane and camera axis -- thus improving rendering for tiny scale drawings (as seen in https://forum.freecadweb.org/viewtopic.php?f=10&t=72501).

PR #7622, opened 2022-10-20
Sketcher: Solver - Improvement of popularity contest and bug fix

abdullahtahiriyo:
================================================================

Master has a problem in that internal alignment constraints are suggested to the user for removal.

This is fundamentally wrong, as an internal alignment constraint are an inherent part of the geometry. They cannot be the ones suggested for removal.

The popularity contest algorithm is an heuristic algorithm that determines which redundant/conflicting constraints should be proposed for removal.

Basically, the algorithm works on groups of redundant/conflicting constraints detected via the QR decomposition. A constraint may belong to more than one group.

The algorithm runs some heuristics, each constraint scoring a value, the one constraint from each group scoring the highest is proposed (is more popular and wins the contest).

This PR documents the algorithm, and adds a further condition, that internal alignment constraints are never proposed.

As the solver works with solver constraints as opposed to the sketcher, which works with sketcher constraints, information about whether a solver constraint originated from a sketcher constraint that is internal alignment is necessary. So the solver constraint is extended to accomodate this piece of information.

As a bonus, it fixes a bug. Solver constraints carry information of the ID of the corresponding sketcher constraint in their tag. Knots are not currently implemented as constraints. However, the tag index was not being update. This caused the popularity contest to provide wrong suggestions despite good detection.

### DIFF
--- a/src/Mod/Sketcher/App/planegcs/GCS.h
+++ b/src/Mod/Sketcher/App/planegcs/GCS.h
@@ -230,7 +230,7 @@ namespace GCS
         void removeConstraint(Constraint *constr);
 
         // basic constraints
-        int addConstraintEqual(double *param1, double *param2, int tagId=0, bool driving = true);
+        int addConstraintEqual(double *param1, double *param2, int tagId=0, bool driving = true, Constraint::Alignment internalalignment = Constraint::Alignment::NoInternalAlignment);
         int addConstraintProportional(double *param1, double *param2, double ratio, int tagId, bool driving = true);
         int addConstraintDifference(double *param1, double *param2,
                                     double *difference, int tagId=0, bool driving = true);
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -3283,6 +3283,8 @@ int Sketch::addInternalAlignmentKnotPoint(int geoId1, int geoId2, int knotindex)
 
         // no constraint is actually added, as knots are fixed geometry in this implementation
         // indexing is added here.
+        // However, we need to advance the tag, so that the index after a knot constraint is accurate
+        ConstraintsCounter++;
 
         b.knotpointGeoids[knotindex] = geoId2;
 
--- a/src/Mod/Sketcher/App/planegcs/Constraints.cpp
+++ b/src/Mod/Sketcher/App/planegcs/Constraints.cpp
@@ -37,7 +37,7 @@ namespace GCS
 ///////////////////////////////////////
 
 Constraint::Constraint()
-: origpvec(0), pvec(0), scale(1.), tag(0), pvecChangedFlag(true), driving(true)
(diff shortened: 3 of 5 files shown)


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7622">pipeline status</a> for feature branch [PR_7622](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7622). Pipeline [671934053](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/671934053) was triggered at [4582408](https://github.com/FreeCAD/FreeCAD/pull/7622/commits/458240849adc2586c0ceb62d88e105af72e645bc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Looks like the CI failure is just codespell:
`src/Mod/Sketcher/App/planegcs/GCS.cpp:4497: prefered ==> preferred`

src/Mod/Sketcher/App/planegcs/GCS.h
```diff
@@ -230,7 +230,7 @@ namespace GCS
         void removeConstraint(Constraint *constr);
 
         // basic constraints
-        int addConstraintEqual(double *param1, double *param2, int tagId=0, bool driving = true);
+        int addConstraintEqual(double *param1, double *param2, int tagId=0, bool driving = true, bool internalalignment = false);
```
chennes, line 233: I've been keeping an eye out for issues of "boolean blindness" recently after @wwmayer introduced me to the idea -- I notice when you actually set this variable to true you are even including a comment to indicate what it is, and I wonder if the `enum` solution might be a good approach here instead.

abdullahtahiriyo: Actually, I just saw the driving boolean and run in autopilot. 

I think you are right. For both of them. I will review and revert.

abdullahtahiriyo: I decided not to change the driving boolean at this time, but only change the internal alignment to an enum, so as not to interfere with the PR of BSpline improved solver support. That can be done separately after merging that other PR in a dedicated PR.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7622">pipeline status</a> for feature branch [PR_7622](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7622). Pipeline [673217838](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/673217838) was triggered at [77fc513](https://github.com/FreeCAD/FreeCAD/pull/7622/commits/77fc513358fa16c8a6192656dbff0657d0fe93dd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #7642, opened 2022-10-25
[PD] FeatureHole.cpp: list deprecated types at last 

donovaly:
- currently we list the deprecated hole cut types before the normed ones.
  This PR puts the deprecated types to the end since they only force the user to
  scroll down every time to use normed ones

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHole.cpp
+++ b/src/Mod/PartDesign/App/FeatureHole.cpp
@@ -2150,23 +2150,42 @@ void Hole::addCutType(const CutDimensionSet& dimensions)
     const std::string& name = dimensions.name;
 
     std::vector<std::string>* list;
+    int counter = -1;
+
     switch (thread) {
     case CutDimensionSet::Metric:
         HoleCutTypeMap.emplace(CutDimensionKey("ISOMetricProfile", name), dimensions);
         list = &HoleCutType_ISOmetric_Enums;
+        // search for first entry that contains "deprecated"
+        for (auto it = HoleCutType_ISOmetric_Enums.begin();
+             it != HoleCutType_ISOmetric_Enums.end(); it++) {
+            counter += 1;
+            if ((*it).find("deprecated") != std::string::npos)
+                break;
+        }
         break;
     case CutDimensionSet::MetricFine:
         HoleCutTypeMap.emplace(CutDimensionKey("ISOMetricFineProfile", name), dimensions);
         list = &HoleCutType_ISOmetricfine_Enums;
+        for (auto it = HoleCutType_ISOmetricfine_Enums.begin();
+             it != HoleCutType_ISOmetricfine_Enums.end(); it++) {
+            counter += 1;
+            if ((*it).find("deprecated") != std::string::npos)
+                break;
+        }
         break;
     default:
         return;
     }
     // add the collected lists of JSON definitions to the lists
     // if a name doesn't already exist in the list
+    // sort them hereby in before the deprecated entries
     if (std::all_of(list->begin(), list->end(),
-                [name](const std::string& x) { return x != name; }))
-        list->push_back(name);
+                    [name](const std::string &x) { return x != name; })) {
+        auto listPos = list->begin() + counter;
+        auto newIt = list->insert(listPos, name);
+        counter += 1;
+    }
 }
 
 bool Hole::isDynamicCounterbore(const std::string& thread,


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7642">pipeline status</a> for feature branch [PR_7642](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7642). Pipeline [675868170](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/675868170) was triggered at [4a7d918](https://github.com/FreeCAD/FreeCAD/pull/7642/commits/4a7d9188f23f505ee9ceffa21ceec3471a186e14). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7642">pipeline status</a> for feature branch [PR_7642](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7642). Pipeline [675871876](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/675871876) was triggered at [3bc490b](https://github.com/FreeCAD/FreeCAD/pull/7642/commits/3bc490b255a99a0cbaeb7fa229dd5d8d01256f2b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> does it make sense to keep these deprecated types around?

I think so. We have no idea yet who used the deprecated types. By removing them we would destroy backwards compatibility. Since with this PR they appear as last entries, it does not harm to keep them there.

src/Mod/PartDesign/App/FeatureHole.cpp
```diff
@@ -2150,23 +2150,42 @@ void Hole::addCutType(const CutDimensionSet& dimensions)
...
+             it != HoleCutType_ISOmetric_Enums.end(); it++) {
+            counter += 1;
+            if ((*it).find("deprecated") != std::string::npos)
+                break;
+        }
         break;
     case CutDimensionSet::MetricFine:
         HoleCutTypeMap.emplace(CutDimensionKey("ISOMetricFineProfile", name), dimensions);
         list = &HoleCutType_ISOmetricfine_Enums;
+        for (auto it = HoleCutType_ISOmetricfine_Enums.begin();
```
stefanklug, line 2170: Writing a helper function
```
int getIndexOfFirstDepricatedEntry( std::vector<std::string>& list) {
....
} 
```
could reduce some code duplication here

donovaly:
> I'm new to the codebase so excuse me if I misinterpret things.

That's how I incorporate you to FreeCAD ;-)

All I did here is to change the map. As it is, the manual (non-XML) entries are the beginning of the types map. Then all collected XML entries are mapped behind. How the XML is collected was not changed by the PR. I keep the collection as it is, only insert it to another place in the map.

donovaly:
> So assuming someone created a document with FreeCAD 0.18 and HoleType "Cheesehead". He will never be able to open that document with FreeCAD 0.19+ as there is no entry "Cheesehead" anymore. It is now called "Cheesehead (depricated)".
So how does the backwards compatibility work?

Good point. Do you have a FC 0.18 hole file for me to test please? (I have no FC 0.18 anymore)

However, if this is an issue, it is independent of this PR.

src/Mod/PartDesign/App/FeatureHole.cpp
donovaly: I also thought so but the helper function + its usage was more code lines than two times the call. And since there will for sure no further thread type, we can assume it will stay with those cases.
But OK, I will try a lambda.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7642">pipeline status</a> for feature branch [PR_7642](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7642). Pipeline [677052328](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/677052328) was triggered at [b16f826](https://github.com/FreeCAD/FreeCAD/pull/7642/commits/b16f8266e3fe9abae5901989c3cc1304f96bd709). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

stefanklug:
Here is a testfile created with the 0.18.4 AppImage.
[Cheesehead_hole_fc18.zip](https://github.com/FreeCAD/FreeCAD/files/9866926/Cheesehead_hole_fc18.zip)

Summary:
At that point in time, only an index t the enum entry was stored, because the enum was not yet custom. So these old files depend on the enum order.
Loading the file in 0.19 leads to a hole type "None".

So it seems the backwards compatibility is broken and we should come up with a solution that is future proof (something like string keys, or actually constant numbers comes to mind). 
But for now, why not just remove the depricated entries as no one seems to use them anymore (Otherwise you would have heard).

src/Mod/PartDesign/App/FeatureHole.cpp
```diff
@@ -2150,23 +2150,42 @@ void Hole::addCutType(const CutDimensionSet& dimensions)
...
+            if ((*it).find("deprecated") != std::string::npos)
+                break;
+        }
         break;
     default:
         return;
     }
     // add the collected lists of JSON definitions to the lists
     // if a name doesn't already exist in the list
+    // sort them hereby in before the deprecated entries
```
luzpaz, line 2182: how about: `sort them here inserting them before the deprecated entries`

adrianinsaval:
can these deprecated modes not be mapped to one of the supported modes by adjusting some parameter? what are they anyways? IMO we should remove these if we're going for a 1.0 release of freecad

donovaly:
> Here is a testfile created with the 0.18.4 AppImage. [Cheesehead_hole_fc18.zip](https://github.com/FreeCAD/FreeCAD/files/9866926/Cheesehead_hole_fc18.zip)

Thanks.
I see that we cannot do anything to support these because the info on the file is just:
```
<Property name="HoleCutType" type="App::PropertyEnumeration">
                    <Integer value="3"/>
                </Property>
```
Since FC 0.19 the content would be this:
```
<Property name="HoleCutType" type="App::PropertyEnumeration">
                    <Integer value="3" CustomEnum="true"/>
                    <CustomEnumList count="14">
                        <Enum value="None"/>
                        <Enum value="Counterbore"/>
                        <Enum value="Countersink"/>
                        <Enum value="Cheesehead (deprecated)"/>
                        <Enum value="Countersink socket screw (deprecated)"/>
                        <Enum value="Cap screw (deprecated)"/>
                        <Enum value="DIN 7984"/>
                        <Enum value="ISO 10642"/>
                        <Enum value="ISO 14583"/>
                        <Enum value="ISO 14583 (partial)"/>
                        <Enum value="ISO 2009"/>
                        <Enum value="ISO 4762"/>
                        <Enum value="ISO 4762 + 7089"/>
                        <Enum value="ISO 7046"/>
                    </CustomEnumList>
                </Property>
```

So the backwards compatibility is broken since FC 0.19. Since I haven't seen any complaint yet, these hole types seems to be very rarely used.

Thus we can remove the deprecated types. And for the case that someone used this type in FC 0.19 or 0.20 we can transfer these holes to custom ones that the actual hole parameters are preserved. -> I'll make a PR.

This PR is hereby obsolete -> closing it.

donovaly:
> Thus we can remove the deprecated types. And for the case that someone used this type in FC 0.19 or 0.20 we can transfer these holes to custom ones that the actual hole parameters are preserved. -> I'll make a PR.

Here it is: #7654

PR #7581, opened 2022-10-08
Sketcher: Task panel, 'solver message' and 'Edit controls' rework.

PaddleStroke:
https://forum.freecadweb.org/viewtopic.php?f=8&t=72262

Thank you for creating a pull request to contribute to FreeCAD! To ease integration, we ask you to conform to the following items. Pull requests which don't satisfy all the items below might be rejected. If you are in doubt with any of the items below, don't hesitate to ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10)!

- [ ]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [ ]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [ ]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [ ]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [ ]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [ ]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the issue ID number from our [Issues database](https://github.com/FreeCAD/FreeCAD/issues) in case a particular commit solves or is related to an existing issue. Ex: `Draft: fix typos - fixes #4805`

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---

PR #7643, opened 2022-10-25
[Sketch] set readable tooltip for DoF selection 

donovaly:
- since the removal of the DOF toolbar button we only have the link to select DOFs. this has however, only the meaningless tooltip "#dof" (we have many users who use English menus but are no native speakers because FC is not translated to their language. For them the term "#dof" is quite meaningless)
- also change the string "DoF" for the same reason

@chennes , @PaddleStroke

### DIFF
--- a/src/Mod/Sketcher/Gui/TaskSketcherMessages.cpp
+++ b/src/Mod/Sketcher/Gui/TaskSketcherMessages.cpp
@@ -111,26 +111,26 @@ void TaskSketcherMessages::slotSetUp(const QString& state, const QString& msg, c
 
 void TaskSketcherMessages::on_labelConstrainStatusLink_linkClicked(const QString &str)
 {
-    if( str == QString::fromLatin1("#conflicting"))
+    if (str == QString::fromLatin1("#conflicting"))
         Gui::Application::Instance->commandManager().runCommandByName("Sketcher_SelectConflictingConstraints");
     else
-    if( str == QString::fromLatin1("#redundant"))
+    if (str == QString::fromLatin1("#redundant"))
         Gui::Application::Instance->commandManager().runCommandByName("Sketcher_SelectRedundantConstraints");
     else
-    if( str == QString::fromLatin1("#dofs"))
+    if (str == QString::fromLatin1("Select geometrical elements where solver\n"
+                                    "detects unconstrained degrees of freedom"))
         Gui::Application::Instance->commandManager().runCommandByName("Sketcher_SelectElementsWithDoFs");
     else
-    if( str == QString::fromLatin1("#malformed"))
+    if (str == QString::fromLatin1("#malformed"))
         Gui::Application::Instance->commandManager().runCommandByName("Sketcher_SelectMalformedConstraints");
     else
-    if( str == QString::fromLatin1("#partiallyredundant"))
+    if (str == QString::fromLatin1("#partiallyredundant"))
         Gui::Application::Instance->commandManager().runCommandByName("Sketcher_SelectPartiallyRedundantConstraints");
-
 }
 
 void TaskSketcherMessages::on_autoUpdate_stateChanged(int state)
 {
-    if(state==Qt::Checked) {
+    if (state==Qt::Checked) {
         sketchView->getSketchObject()->noRecomputes=false;
         ui->autoUpdate->onSave();
     }
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -3009,8 +3009,9 @@ void ViewProviderSketch::UpdateSolverInformation()
     } else if (dofs > 0) {
         signalSetUp(QString::fromUtf8("under_constrained"),
             tr("Under constrained:"),
-            QString::fromUtf8("#dofs"),
-            tr("%n DoF(s)","",dofs));
+                    QString::fromUtf8("Select geometrical elements where solver\n"
+                                      "detects unconstrained degrees of freedom"),
+            tr("%n Degree(s) of freedom", "", dofs));
     }
     else {
         signalSetUp(QString::fromUtf8("fully_constrained"), tr("Fully constrained"), QString(), QString());


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7643">pipeline status</a> for feature branch [PR_7643](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7643). Pipeline [675884945](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/675884945) was triggered at [2827200](https://github.com/FreeCAD/FreeCAD/pull/7643/commits/2827200bde6859aee0e32cc001a0d1f139516125). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes (approved):
Much better! Thanks.

luzpaz:
Sorry for my ignorance but will this tooltip be translated?

PaddleStroke:
For such UI modifications please provide a screenshot of the modifications. It would be much better to get feedback.
Having to build just to see how it looks like is counter productive. I'm currently rebuilding the project and loosing 1 hour just to check, because VS somehow decided to rebuild the whole project for no good reason.

Adding a proper tooltip is welcome. 
But if you changed the string from DoF to 'Degree of freedom' this is not. There is no reason to change the string, especially since it increase a lot the width of the widget. When we had the DOF tool in the toolbar, the icon wasn't more explicit than a 'DoF' string. So I don't understand why we would need it here.

PaddleStroke:
Ok so those who wondered here what it looks like : 
image

This is not acceptable. It's totally killing the width of the task view for no good reasons. It's almost as wide as the current constraint widget (which isn't going to stay like this for long for that very reason).

chennes:
> Sorry for my ignorance but will this tooltip be translated?

Yes (what you are looking for in the code is that the string is within a `tr()` call).

adrianinsaval:
I think this message was shortened to be just DoF on purpose, the tooltip is good but the message itself should stay like this IMO.

src/Mod/Sketcher/Gui/TaskSketcherMessages.cpp
```diff
@@ -111,26 +111,26 @@ void TaskSketcherMessages::slotSetUp(const QString& state, const QString& msg, c
 
 void TaskSketcherMessages::on_labelConstrainStatusLink_linkClicked(const QString &str)
 {
-    if( str == QString::fromLatin1("#conflicting"))
+    if (str == QString::fromLatin1("#conflicting"))
         Gui::Application::Instance->commandManager().runCommandByName("Sketcher_SelectConflictingConstraints");
     else
-    if( str == QString::fromLatin1("#redundant"))
+    if (str == QString::fromLatin1("#redundant"))
```
adrianinsaval, line 117: why not also add more useful tooltips for these too?
image

chennes: I think we can't, the "tooltip" there is the link it follows, and has to be that exact text (or at least, cannot be translated).

donovaly: Yes, and this is a problem. As I wrote for a non-native speaker "DoF" is not clear as being an abbreviation nor what it stands for,
But also adding the tooltip does not help as it should when it is not translatable. @adrianinsaval is right, the other "#..." tooltips are also not useful for all users and not translatable.

Unfortunately I don't have a good idea. In my opinion the link should in general be replaced by a button to follow our general design of dialogs. Maybe when doing this, the tooltip can be fixed? @abdullahtahiriyo ?

PaddleStroke: We can replace the link url label by the corresponding button we had in the toolbar. And the number of constraints could be put in the 'Redundant constraints:' label.
So it would be : 'Redudant constraints: 5' label +  SelectRedundantToolButton.
This way we solve all the problems: 
- 'DoF' string meaning
- Tooltip not working
- width of the text will be smaller.

PaddleStroke:
I heard recurring complaints about the fact that the task view was too large. And I can understand them, because the more 3D view for the same functionality the better.

Regarding the DoF meaning. It's not clear for non english speakers obviously. But for english speakers DoF is also meaningless. Until you read somewhere about it, which any user, unrelated to their language would end up doing.

**What about this suggestion :** 
image
The label becomes a normal translatable label without URL. It writes plainly Degree Of Freedom. Even the label can have a tooltip explaining what are degree of freedom.
The DoF button is added next to the label. And this button can have the tooltip. 

Besides before going further on this, to avoid conflicts, @donovaly do you want to look at this PR ? 
https://github.com/FreeCAD/FreeCAD/pull/7581
It is an attempt to make the solver message more visible and seemed to be very well accepted in the forum discussion.

PaddleStroke:
Last small comment. 'Degree(s)' : we could catch programatically if there's 1 or more and just write Degree or Degrees.

src/Mod/Sketcher/Gui/TaskSketcherMessages.cpp
donovaly: Sounds good. Could you make an according PR?

chennes:
> Last small comment. 'Degree(s)' : we could catch programatically if there's 1 or more and just write Degree or Degrees.

Actually this is already done in the code (via the third argument to `tr()`) -- it's just that no translator has added the English translation for the string, so it's using the default.

donovaly:
> Last small comment. 'Degree(s)' : we could catch programatically if there's 1 or more and just write Degree or Degrees.

Yes, I agree, we should use "Degrees". This saves 2 characters and this is more worth than having it grammatically correct for the case of 1 degree.

donovaly:
> I heard recurring complaints about the fact that the task view was too large. And I can understand them, because the more 3D view for the same functionality the better.
> 
> Regarding the DoF meaning. It's not clear for non english speakers obviously. But for english speakers DoF is also meaningless. Until you read somewhere about it, which any user, unrelated to their language would end up doing.
> 
> **What about this suggestion :** image The label becomes a normal translatable label without URL. It writes plainly Degree Of Freedom. Even the label can have a tooltip explaining what are degree of freedom. The DoF button is added next to the label. And this button can have the tooltip.
> 
> Besides before going further on this, to avoid conflicts, @donovaly do you want to look at this PR ? #7581 It is an attempt to make the solver message more visible and seemed to be very well accepted in the forum discussion.

OK. Can you rebase that PR that I can see it with the recent changes in master?
Only in general, let's please do it step by step (the big change but in small portions) exactly to keep track of the changes (translations, backwards compatibility, different screen resolutions etc.)

abdullahtahiriyo:
I moved the discussion about both PRs back to the forum for further input:
https://forum.freecadweb.org/viewtopic.php?p=638962#p638962

abdullahtahiriyo:
I have merged improved tooltips. A sample can be found here:
https://forum.freecad.org/viewtopic.php?p=677307#p677307

adrianinsaval:
we can consider this PR obsolete now right?

abdullahtahiriyo:
@donovaly 

Are you satisfied with the current solution?

PR #7655, opened 2022-10-27
Sketcher: Gui: force-hide toolbars + Split sketcher toolbar

PaddleStroke:
The offered change is to :
Commit 1 - split the main sketcher toolbar in 2. One for tools that are only in edit mode. One for tools that are only in non-edit-mode.

image

Commit 2 - This adds the systemwide possibility to hide toolbars at anytime while in a workbench.
Show/hide automatically toolbars depending if we are in edit mode or not.
More specifically it hides : Sketcher-edit-mode, Geometry, constraint, tools, virtualSpace toolbars when in sketcher-not-edit-mode.
And it hides : Sketcher and Structure toolbar when in sketcher-edit-mode
Video presentation : https://youtu.be/5HF8f648JlU

[See discussion : https://forum.freecadweb.org/viewtopic.php?f=8&t=72946](https://forum.freecadweb.org/viewtopic.php?f=17&t=72978)

---

### DIFF
--- a/src/Gui/ToolBarManager.cpp
+++ b/src/Gui/ToolBarManager.cpp
@@ -36,11 +36,11 @@
 
 using namespace Gui;
 
-ToolBarItem::ToolBarItem()
+ToolBarItem::ToolBarItem() : forceHide(HideStyle::VISIBLE)
 {
 }
 
-ToolBarItem::ToolBarItem(ToolBarItem* item)
+ToolBarItem::ToolBarItem(ToolBarItem* item, HideStyle forcehide) : forceHide(forcehide)
 {
     if ( item )
         item->appendItem(this);
@@ -196,7 +196,6 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
         this->toolbarNames << name;
         QToolBar* toolbar = findToolBar(toolbars, name);
         std::string toolbarName = (*it)->command();
-        bool visible = hPref->GetBool(toolbarName.c_str(), true);
         bool toolbar_added = false;
 
         if (!toolbar) {
@@ -210,15 +209,16 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
                     + QChar::fromLatin1(']');
                 toolbar->setToolTip(tooltip);
             }
-            toolbar->setVisible(visible);
             toolbar_added = true;
         }
         else {
-            toolbar->setVisible(visible);
-            toolbar->toggleViewAction()->setVisible(true);
             int index = toolbars.indexOf(toolbar);
             toolbars.removeAt(index);
         }
+
+        bool visible = hPref->GetBool(toolbarName.c_str(), true) && (*it)->forceHide == ToolBarItem::HideStyle::VISIBLE;
+        toolbar->setVisible(visible);
+        toolbar->toggleViewAction()->setVisible((*it)->forceHide == ToolBarItem::HideStyle::VISIBLE);
 
         // setup the toolbar
         setup(*it, toolbar);
@@ -254,7 +254,7 @@ void ToolBarManager::setup(ToolBarItem* toolBarItems)
         QByteArray toolbarName = (*it)->objectName().toUtf8();
         if (!(*it)->toggleViewAction()->isVisible())
             continue;
-        hPref->SetBool(toolbarName.constData(), (*it)->isVisible());
+        //hPref->SetBool(toolbarName.constData(), (*it)->isVisible());
         (*it)->hide();
         (*it)->toggleViewAction()->setVisible(false);
     }
@@ -308,6 +308,8 @@ void ToolBarManager::saveState() const
     for (QStringList::ConstIterator it = this->toolbarNames.begin(); it != this->toolbarNames.end(); ++it) {
         QToolBar* toolbar = findToolBar(toolbars, *it);
         if (toolbar) {
+            if (!toolbar->toggleViewAction()->isVisible())
+                continue; //if toggleViewAction is not visible it means that the toolbar is forceHide. In which case we do not save settings.
             QByteArray toolbarName = toolbar->objectName().toUtf8();
             hPref->SetBool(toolbarName.constData(), toolbar->isVisible());
         }
@@ -384,4 +386,28 @@ QList<QToolBar*> ToolBarManager::toolBars() const
     }
 
     return tb;
+}
+
+void ToolBarManager::setToolbarVisibility(bool show, const QList<QString>& names) {
+    for (auto& name : names) {
+        setToolbarVisibility(show, name);
+    }
+}
+
+void ToolBarManager::setToolbarVisibility(bool show, const QString& name) {
+
+    ParameterGrp::handle hPref = App::GetApplication().GetUserParameter().GetGroup("BaseApp")->GetGroup("MainWindow")->GetGroup("Toolbars");
+
+    QToolBar* tb = findToolBar(toolBars(), name);
+    if (tb) {
+        if (show) {
+            if(hPref->GetBool(name.toStdString().c_str(), true))
+                tb->show();
+            tb->toggleViewAction()->setVisible(true);
+        }
+        else {
+            tb->hide();
+            tb->toggleViewAction()->setVisible(false);
+        }
+    }
 }
--- a/src/Gui/ToolBarManager.h
+++ b/src/Gui/ToolBarManager.h
@@ -35,8 +35,13 @@ namespace Gui {
 class GuiExport ToolBarItem
 {
 public:
+    enum class HideStyle {
+        VISIBLE,
+        FORCE_HIDE // Force a toolbar to be hidden. For when all elements are disabled at some point in a workbench.
+    };
+
     ToolBarItem();
-    ToolBarItem(ToolBarItem* item);
+    ToolBarItem(ToolBarItem* item, HideStyle forceHide = HideStyle::VISIBLE);
     ~ToolBarItem();
 
     void setCommand(const std::string&);
@@ -55,6 +60,8 @@ public:
     ToolBarItem& operator << (ToolBarItem* item);
     ToolBarItem& operator << (const std::string& command);
     QList<ToolBarItem*> getItems() const;
+
+    HideStyle forceHide;
 
 private:
     std::string _name;
@@ -81,6 +88,8 @@ public:
     void retranslate() const;
 
     void setMovable(bool movable) const;
+    void setToolbarVisibility(bool show, const QList<QString>& names);
+    void setToolbarVisibility(bool show, const QString& name);
 
 protected:
     void setup(ToolBarItem*, QToolBar*) const;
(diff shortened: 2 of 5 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655">pipeline status</a> for feature branch [PR_7655](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655). Pipeline [678333802](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/678333802) was triggered at [d2c5861](https://github.com/FreeCAD/FreeCAD/pull/7655/commits/d2c5861e3ee83cb97ac325544a2b8c7c4023e672). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655">pipeline status</a> for feature branch [PR_7655](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655). Pipeline [679791465](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/679791465) was triggered at [79917e3](https://github.com/FreeCAD/FreeCAD/pull/7655/commits/79917e3788c912fc95fdf5328e6f53a748def80b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/ToolBarManager.cpp
```diff
@@ -36,11 +36,11 @@
 
 using namespace Gui;
 
-ToolBarItem::ToolBarItem()
+ToolBarItem::ToolBarItem() : forceHide(false)
 {
 }
 
-ToolBarItem::ToolBarItem(ToolBarItem* item)
+ToolBarItem::ToolBarItem(ToolBarItem* item, bool forcehide) : forceHide(forcehide)
```
chennes, line 43: As a general rule we should prefer not to introduce these trailing boolean arguments: it creates a problem called "boolean blindness." Basically, we will have some function call in the code with a random boolean appended to it. What does that boolean mean? No one knows! So you can instead structure your code to force the calling code to be a little more verbose by introducing an enumeration. In your case, something like (untested):
```
enum class HideStyle {
    ALWAYS_VISIBLE,
    AUTO_HIDE // Auto-hide when all elements are disabled
};
...
ToolBarItem::ToolBarItem(ToolBarItem* item, HideStyle forcehide) : forceHide(forcehide)
```
so that a call looks like:
```
auto toolBarItem = ToolBarItem(otherItem, ToolBarItem::HideStyle::AUTO_HIDE);
```
There's no mistaking what that final argument is doing now (well, you have to know what AUTO_HIDE means, but that's a separate problem 😄).

PaddleStroke:
Rebased and change the boolean. 
I confirm it still works as expected.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655">pipeline status</a> for feature branch [PR_7655](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655). Pipeline [681151997](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/681151997) was triggered at [668cbb3](https://github.com/FreeCAD/FreeCAD/pull/7655/commits/668cbb37955e1af86f32681c6ad2867915422dda). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/ToolBarManager.cpp
```diff
@@ -385,3 +387,27 @@ QList<QToolBar*> ToolBarManager::toolBars() const
 
     return tb;
 }
+
+void ToolBarManager::setToolbarVisibility(bool show, const QList<QString>& names) {
+    for (size_t i = 0; i < names.size(); i++) {
+        setToolbarVisibility(show, names[i]);
+    }
```
chennes, line 394: Untested, but this looks to me like it should be a range-based for loop:
```suggestion
    for (const auto &name : names) {
        setToolbarVisibility(show, name);
    }

```

PaddleStroke: I'm not sure to understand the difference. Of course I can fix it here, but should I always use this kind of for loop instead of the more 'traditional' one?

chennes: Short answer: yes, always use range-based `for` loops. If you want more info, the C++ Core Guidelines are a good place to start: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-for-range

PaddleStroke: Ok thanks. I'll have a look if there's any other loops to change.
Is there any other issue ?

chennes: I'm waiting for my compile to finish, I'll let you know shortly.

chennes:
I'm seeing doubled-up buttons when I access Sketcher via PartDesign:
Screenshot 2022-11-03 at 3 32 15 PM

PaddleStroke:
Your screenshot is on part design.
From your screenshot what I see is that the 'sketcher-Not-Edit-mode' toolbar doesn't disappear (or appears) when you revert from sketcher to PartDesign.
Is that correct?

Did you rebuild entirely the project? It seems that modifying ToolBarItem class in anyway (here adding the boolean, or enum class now) requires a full rebuild or the build is broken.
Though to get to the point where you are, I would already have had to rebuild entirely.

PaddleStroke:
Ok I can replicate the issue. 'Sketcher' toolbar appears in part design when leaving Sketcher back to PartDesign

 I didn't caught it because on my testing the sketcher toolbar was appearing out of screen.
I'll have a look into this bug and fix it asap.

PaddleStroke:
@chennes Ok it's fixed. The snippet of code to show/force-hide toolbar in the unsetEdit function was called at the end of the function instead of the beginning. So it force-showed the sketcher toolbar in partdesign.
I also fixed the for loop.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655">pipeline status</a> for feature branch [PR_7655](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655). Pipeline [685211796](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/685211796) was triggered at [52f3cc2](https://github.com/FreeCAD/FreeCAD/pull/7655/commits/52f3cc21287e870e858615013140f426d48b5e71). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Excellent, thanks. The last thing (I think!) is to eliminate the code duplication where you have two identical string lists constructed in separate functions. That should be refactored so those lists are only generated in one place. I'd suggest you write two functions (in an anonymous namespace) right above `setupCommandBars`, something like:
```
namespace
{
    inline const QStringList editModeToolbarNames()
    {
        return QStringList {QString::fromLatin1("Sketcher Edit Mode"),
                            QString::fromLatin1("Sketcher geometries"),
                            QString::fromLatin1("Sketcher constraints"),
                            QString::fromLatin1("Sketcher tools"),
                            QString::fromLatin1("Sketcher B-spline tools"),
                            QString::fromLatin1("Sketcher virtual space")};
    }

    inline const QStringList nonEditModeToolbarNames() 
    {
        return QStringList {QString::fromLatin1("Structure"), 
                            QString::fromLatin1("Sketcher")};
    }
}
```
then make your calls look like:
```
    Gui::ToolBarManager::getInstance()->setToolbarVisibility(true, editModeToolbarNames());
    Gui::ToolBarManager::getInstance()->setToolbarVisibility(false, nonEditModeToolbarNames());
```
(and the reverse below, of course).

PaddleStroke:
So I add this nameless namespace in sketcher workbench.cpp, above Workbench::setupCommandBars().
- How do I get access to those functions in ViewProviderSketcher.cpp then? Should I include workbench.cpp in ViewProviderSketcher.cpp?
- What is the point of the nameless namespace ?

PaddleStroke:
I think you meant 'right below ViewProviderSketch::setupContextMenu'
And not 'right above setupCommandBars' as setupCommandBars is only in workbench and from my reading nameless namespace is meant for function not to be accessible from outside the file. So adding them to workbench.cpp doesn't seem to make sense.

I pushed like this. If it's not what you had in mind please let me know.

PaddleStroke:
I just forced push with this latest change. I squashed the fix commits so that there are only 2 commits left.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655">pipeline status</a> for feature branch [PR_7655](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7655). Pipeline [685456477](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/685456477) was triggered at [2b6e121](https://github.com/FreeCAD/FreeCAD/pull/7655/commits/2b6e12120b2d9adb495cae40a157bd9b5d096f5c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
Great thanks for taking your time here !

src/Gui/ToolBarManager.cpp
wwmayer: > Of course I can fix it here, but should I always use this kind of for loop instead of the more 'traditional' one?

Whenever possible use the range-based for loop as this is more efficient than the traditional way. If you write
`for (auto it = container.begin(); it != container.end(); ++it)` then for each iteration step the function end() must be called. The range-based for loop is syntactic sugar where the compiler creates code to call end() only once.
For more details have a look at: http://antonym.org/2014/02/c-plus-plus-11-range-based-for-loops.html

Hint: Sometimes you need two variables when iterating over a container, the iterator variable and an index. Thanks to boost you can also use the range-based for loop (https://www.fluentcpp.com/2018/10/26/how-to-access-the-index-of-the-current-element-in-a-modern-for-loop/):
```
    for (auto const& it : container | boost::adaptors::indexed(0)) {
        it.value(); // the element of the container
        it.index(); // the index variable
    }
```

PaddleStroke: Thanks for the details. I was wondering about the index, I'll make that down somewhere.

wwmayer:
> What is the point of the nameless namespace ?

The traditional way is to use static functions because they are only visible within a translation unit.
See: https://stackoverflow.com/questions/15235526/the-static-keyword-and-its-various-uses-in-c

The modern way is to use anonymous namespaces where the compiler creates unique namespace names to avoid name clashes.

PR #7667, opened 2022-10-29
[PartDesign] Revolution Show Sketch...

### LINKED ISSUE #7629: [Feature Request] [PartDesign] Keep sketch visible during revolution (and possibly other) tasks

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=8&t=72779

### Subproject(s) affected?

PartDesign

### Idea description

During the revolution task (additive or subtractive), if you want to select a part of the sketch you are using for the task itself, you currently have to go to the tree view and turn on the sketches visibility, because the sketch will get hidden as soon as the task is started.

I suggest to keep the sketch visible until the task is closed to prevent unnecessary steps (or make the sketch visible when "select reference…" is chosen, but just keeping the sketch visible seems simpler to me).

Given that the revolution preview will be hidden as soon as "select reference…" is selected, there is no risk of accidentally selecting parts of the revolution. I don't see any potential downsides, apart from a slightly less clean preview I guess.

I haven't checked other operations yet, but if the active sketches can be selected for things, I think they should stay visible until the task is completed in these operations as well.

### Anything else?

_No response_

### Code of Conduct

Syres916:
...when 'Select reference...' is used, see discussion https://forum.freecadweb.org/viewtopic.php?f=8&t=72779

PD: Revolution fixes #7629

- [x]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the issue ID number from our [Issues database](https://github.com/FreeCAD/FreeCAD/issues) in case a particular commit solves or is related to an existing issue. Ex: `Draft: fix typos - fixes #4805`

---

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
@@ -250,6 +250,10 @@ void TaskRevolutionParameters::onAxisChanged(int num)
     App::PropertyLinkSub &lnk = *(axesInList[num]);
     if (!lnk.getValue()) {
         // enter reference selection mode
+        auto sketch = dynamic_cast<Part::Part2DObject*>(pcRevolution->Profile.getValue());
+        if (sketch) {
+            FCMD_OBJ_SHOW(sketch);
+        }
         TaskSketchBasedParameters::onSelectReference(AllowSelection::EDGE |
                                                      AllowSelection::PLANAR |
                                                      AllowSelection::CIRCLE);


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7667">pipeline status</a> for feature branch [PR_7667](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7667). Pipeline [680534865](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/680534865) was triggered at [f0a0e56](https://github.com/FreeCAD/FreeCAD/pull/7667/commits/f0a0e563509b52c46fcc8db8dd07fbc835194676). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks for the PR. However this would introduce inconsistencies. Since I cannot compile right now, allow me some questions:

- will the sketch only be visible if the sketch was visible when the Revolution dialog was called? In my opinion it should only be visible if it was visible before, otherwise we don't follow the user's will.
- what about Groove? To be consistent, Groove is a subtractive Revolution therefore it should behave the same.
- what about Helix (additive and subtractive? A helix is a "Revolution with additional sweep along an axis" ;-). Therefore I would find it suitable if it also shows the sketch if it was visible when the Helix dialog was activated.

Syres916:
> will the sketch only be visible if the sketch was visible when the Revolution dialog was called?

 There's already existing code that hides the sketch when the Revolution/Groove is executed so it's exactly the same as it is now.

> what about Groove?

 If you actually compile the code you would see that the one change applies to both Revolution and Groove

I wanted the requester to try the new functionality on Revolution and Groove and if really needed then others could be PR'd in the near future. I really don't see why all the drama over such a small change that can be backed out with such ease but if you're still not happy then I'll close it.

adrianinsaval:
> I really don't see why all the drama

you were just asked some questions, let's not get carried away. Please don't assume ill intent out of the gate.

donovaly:
> There's already existing code that hides the sketch when the Revolution/Groove is executed so it's exactly the same as it is now.

Thanks. I was asking because it the moment the sketch is hidden in every case, no matter if it was visible before or not.

I will return back home on Saturday, can then compile and try it out.

> > what about Groove?
> 
> If you actually compile the code you would see that the one change applies to both Revolution and Groove

Very good.

> I wanted the requester to try the new functionality on Revolution and Groove and if really needed then others could be PR'd in the near future.

Fine with me. Regarding Helix, do you agree that it should get the same functionality or not?

> I really don't see why all the drama 

What drama? I do the review work, in this case the actual code change is OK, I only have to take care of consistency.

donovaly:
I could now compile and see that you only show the sketch when the user enters the selection mode.

According to #7629 the sketch should ideally been visible all the time. I agree and in my opinion we should have this:
- if sketch was visible before calling the dialog, it stays visible, otherwise it stays invisible
- if it was invisible, but the user goes to selection mode, it will get visible (what this PR already does)

donovaly:
Why do you close this PR?

wwmayer:
The PR was basically fine. As soon as the user wants to select a reference the sketch must become visible as otherwise the workflow becomes a nightmare. Once in selection mode it doesn't work any more to switch to the tree view and make the sketch visible (at least on some OS the selection of the tree item is disabled). So, the user had to abort the whole operation or choose another reference in order to be able to select the tree item and this will result in superfluous computation time.

So, I restored this PR but slightly modified it: 18781bcf63

Issue #7673, opened 2022-10-31
[Feature Request] Prevent losing selection when clicking empty space when Ctrl is pressed


pathfinder555:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=8&t=73056
https://forum.freecadweb.org/viewtopic.php?f=8&t=72435

### Subproject(s) affected?

Other (specify in description)

### Idea description

When selecting features in the main view, while keeping Ctrl pressed, accidentally clicking empty space instead of a feature shouldn't make you lose your selection.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chennes:
Closed by https://github.com/FreeCAD/FreeCAD/pull/7699 -- thanks to @PaddleStroke for coding this up.

PR #7699, opened 2022-11-04
Base: + Sketcher: disable geometry deselection when user holds Ctrl.

PaddleStroke:
This is solving the deselection problem for sketcher and for general 3D view.

https://github.com/FreeCAD/FreeCAD/issues/7673
https://forum.freecadweb.org/viewtopic.php?f=8&t=73056
https://forum.freecadweb.org/viewtopic.php?f=8&t=72435

---

### DIFF
--- a/src/Gui/NavigationStyle.cpp
+++ b/src/Gui/NavigationStyle.cpp
@@ -1479,8 +1479,10 @@ SbBool NavigationStyle::processEvent(const SoEvent * const ev)
     // check for left click without selecting something
     if ((curmode == NavigationStyle::SELECTION || curmode == NavigationStyle::IDLE)
             && !processed) {
-        if (SoMouseButtonEvent::isButtonReleaseEvent(ev,SoMouseButtonEvent::BUTTON1))
-            Gui::Selection().clearSelection();
+        if (SoMouseButtonEvent::isButtonReleaseEvent(ev, SoMouseButtonEvent::BUTTON1)) {
+            if (!(QApplication::keyboardModifiers() == Qt::ControlModifier))
+                Gui::Selection().clearSelection();
+        }
     }
 
     return processed;
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -889,9 +889,10 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                     }
                     Mode = STATUS_NONE;
                     return true;
-                case STATUS_SKETCH_StartRubberBand: // a single click happened, so clear selection
+                case STATUS_SKETCH_StartRubberBand: // a single click happened, so clear selection unless user hold control.
+                    if (!(QApplication::keyboardModifiers() == Qt::ControlModifier))
+                        Gui::Selection().clearSelection();
                     Mode = STATUS_NONE;
-                    Gui::Selection().clearSelection();
                     return true;
                 case STATUS_SKETCH_UseRubberBand:
                     doBoxSelection(DoubleClick::prvCursorPos, cursorPos, viewer);


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7699">pipeline status</a> for feature branch [PR_7699](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7699). Pipeline [685489225](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/685489225) was triggered at [3b950a1](https://github.com/FreeCAD/FreeCAD/pull/7699/commits/3b950a10cd21e5eb49ee55ea1e252b5e2589f796). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7699">pipeline status</a> for feature branch [PR_7699](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7699). Pipeline [685640843](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/685640843) was triggered at [9795999](https://github.com/FreeCAD/FreeCAD/pull/7699/commits/9795999a9ce73b0fdfbbccc053ad428c661378f3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Thanks for the quick PR -- looks great, nice improvement to the behavior.

PaddleStroke:
Thanks for looking into it ! :)

src/Gui/NavigationStyle.cpp
```diff
@@ -1479,8 +1479,10 @@ SbBool NavigationStyle::processEvent(const SoEvent * const ev)
     // check for left click without selecting something
     if ((curmode == NavigationStyle::SELECTION || curmode == NavigationStyle::IDLE)
             && !processed) {
-        if (SoMouseButtonEvent::isButtonReleaseEvent(ev,SoMouseButtonEvent::BUTTON1))
-            Gui::Selection().clearSelection();
+        if (SoMouseButtonEvent::isButtonReleaseEvent(ev, SoMouseButtonEvent::BUTTON1)) {
+            if (!(QApplication::keyboardModifiers() == Qt::ControlModifier))
```
wwmayer, line 1483: @PaddleStroke I think this is the wrong check. If you additionally press another key like Shift or Alt then it still clears the selection. Btw you can also use Inventor API: `if (!ev->wasCtrlDown())`

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -889,9 +889,10 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                     }
                     Mode = STATUS_NONE;
                     return true;
-                case STATUS_SKETCH_StartRubberBand: // a single click happened, so clear selection
+                case STATUS_SKETCH_StartRubberBand: // a single click happened, so clear selection unless user hold control.
+                    if (!(QApplication::keyboardModifiers() == Qt::ControlModifier))
```
wwmayer, line 893: @PaddleStroke This suffers from the same problem but the Inventor API can't be used here. So, the check should be: `if (!(QApplication::keyboardModifiers() & Qt::ControlModifier))`

PaddleStroke:
Thanks for the comments.
Why using inventor API rather than the same at both places?
Do you want me to open a PR for that or do you want to commit directly to
fix it?

wwmayer:
> Why using inventor API rather than the same at both places?

Actually, in all places when we have an SoEvent we use the methods this class provides and avoid to mix two APIs too much.

The next thing is that the signature of `mouseButtonPressed(int Button, bool pressed, const SbVec2s &cursorPos, const Gui::View3DInventorViewer *viewer)` is a design flaw which leads to the use of magic numbers (e.g. for Button) and it suppresses some extra information that might be needed here and there.
So, long story short: at some point this must be cleaned-up and an SoMouseEvent should be directly passed. Then one could consistently use the Inventor API in the implementations of the sub-classes.

> Do you want me to open a PR for that or do you want to commit directly to fix it?

If you agree that the code should be adjusted I can do it directly.

PaddleStroke:
Got it.
Yes of course please go ahead.

wwmayer:
d119c1c7a5

src/Gui/NavigationStyle.cpp
0penBrain: Seconds @wwmayer here using Inventor API for code consistency.
Also the 2 'if' can be joigned together with `&&` for clarity.

PR #7719, opened 2022-11-05
[PD] TaskDlgRevolutionParameters: fix some CI warnings

donovaly:
- reported here: https://github.com/FreeCAD/FreeCAD/pull/7667/commits/f0a0e563

@wwmayer , I don't know what to do with the first 2 warnings there. Can you please have a look?

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
@@ -47,32 +47,31 @@ using namespace Gui;
 /* TRANSLATOR PartDesignGui::TaskRevolutionParameters */
 
 TaskRevolutionParameters::TaskRevolutionParameters(PartDesignGui::ViewProvider* RevolutionView, QWidget *parent)
-    : TaskSketchBasedParameters(RevolutionView, parent, "PartDesign_Revolution", tr("Revolution parameters"))
-    , ui(new Ui_TaskRevolutionParameters)
+    : TaskSketchBasedParameters(RevolutionView, parent, "PartDesign_Revolution", tr("Revolution parameters")),
+      ui(new Ui_TaskRevolutionParameters), proxy(new QWidget(this))
 {
     // we need a separate container widget to add all controls to
-    proxy = new QWidget(this);
     ui->setupUi(proxy);
     QMetaObject::connectSlotsByName(this);
     this->groupLayout()->addWidget(proxy);
 
-    //bind property mirrors
-    PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
-    if (pcFeat->isDerivedFrom(PartDesign::Revolution::getClassTypeId())) {
-        PartDesign::Revolution* rev = static_cast<PartDesign::Revolution*>(vp->getObject());
+    // bind property mirrors
+    if (auto *rev = dynamic_cast<PartDesign::Revolution *>(vp->getObject())) {
         this->propAngle = &(rev->Angle);
         this->propMidPlane = &(rev->Midplane);
         this->propReferenceAxis = &(rev->ReferenceAxis);
         this->propReversed = &(rev->Reversed);
         ui->revolveAngle->bind(rev->Angle);
-    } else {
-        assert(pcFeat->isDerivedFrom(PartDesign::Groove::getClassTypeId()));
-        PartDesign::Groove* rev = static_cast<PartDesign::Groove*>(vp->getObject());
+    }
+    else if (auto *rev = dynamic_cast<PartDesign::Groove *>(vp->getObject())) {
         this->propAngle = &(rev->Angle);
         this->propMidPlane = &(rev->Midplane);
         this->propReferenceAxis = &(rev->ReferenceAxis);
         this->propReversed = &(rev->Reversed);
         ui->revolveAngle->bind(rev->Angle);
+    }
+    else {
+        throw Base::TypeError("The ViewProvider is not a Groove or a Revolution.");
     }
 
     ui->checkBoxMidplane->setChecked(propMidPlane->getValue());
@@ -88,15 +87,16 @@ TaskRevolutionParameters::TaskRevolutionParameters(PartDesignGui::ViewProvider* 
 
     setFocus ();
 
-    //show the parts coordinate system axis for selection
+    // show the parts coordinate system axis for selection
     PartDesign::Body * body = PartDesign::Body::findBodyOf ( vp->getObject () );
-    if(body) {
+    if (body) {
         try {
             App::Origin *origin = body->getOrigin();
-            ViewProviderOrigin* vpOrigin;
-            vpOrigin = static_cast<ViewProviderOrigin*>(Gui::Application::Instance->getViewProvider(origin));
+            auto *vpOrigin = static_cast<ViewProviderOrigin*>(
+                Gui::Application::Instance->getViewProvider(origin));
             vpOrigin->setTemporaryVisibility(true, false);
-        } catch (const Base::Exception &ex) {
+        }
+        catch (const Base::Exception &ex) {
             ex.ReportException();
         }
      }
@@ -109,17 +109,15 @@ void TaskRevolutionParameters::fillAxisCombo(bool forceRefill)
     if (axesInList.empty())
         forceRefill = true;//not filled yet, full refill
 
-    if (forceRefill){
+    if (forceRefill) {
         ui->axis->clear();
-
-        for(size_t i = 0; i < axesInList.size(); i++){
-            delete axesInList[i];
-        }
-        this->axesInList.clear();
+        axesInList.clear();
 
         //add sketch axes
-        PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
-        Part::Part2DObject* pcSketch = dynamic_cast<Part::Part2DObject*>(pcFeat->Profile.getValue());
+        auto *pcFeat = dynamic_cast<PartDesign::ProfileBased*>(vp->getObject());
+        if (!pcFeat)
+            throw Base::TypeError("The ViewProvider is not ProfileBased.");
+        auto *pcSketch = static_cast<Part::Part2DObject*>(pcFeat->Profile.getValue());
         if (pcSketch){
             addAxisToCombo(pcSketch,"V_Axis",QObject::tr("Vertical sketch axis"));
             addAxisToCombo(pcSketch,"H_Axis",QObject::tr("Horizontal sketch axis"));
@@ -176,23 +174,23 @@ void TaskRevolutionParameters::addAxisToCombo(App::DocumentObject* linkObj,
                                               QString itemText)
 {
     this->ui->axis->addItem(itemText);
-    this->axesInList.push_back(new App::PropertyLinkSub());
+    this->axesInList.emplace_back();
     App::PropertyLinkSub &lnk = *(axesInList[axesInList.size()-1]);
     lnk.setValue(linkObj,std::vector<std::string>(1,linkSubname));
 }
 
 void TaskRevolutionParameters::connectSignals()
 {
-    connect(ui->revolveAngle, SIGNAL(valueChanged(double)),
-            this, SLOT(onAngleChanged(double)));
-    connect(ui->axis, SIGNAL(activated(int)),
-            this, SLOT(onAxisChanged(int)));
-    connect(ui->checkBoxMidplane, SIGNAL(toggled(bool)),
-            this, SLOT(onMidplane(bool)));
-    connect(ui->checkBoxReversed, SIGNAL(toggled(bool)),
-            this, SLOT(onReversed(bool)));
-    connect(ui->checkBoxUpdateView, SIGNAL(toggled(bool)),
-            this, SLOT(onUpdateView(bool)));
+    connect(ui->revolveAngle, qOverload<double>(&QuantitySpinBox::valueChanged), this,
+            &TaskRevolutionParameters::onAngleChanged);
+    connect(ui->axis, qOverload<int>(&QComboBox::activated), this,
+            &TaskRevolutionParameters::onAxisChanged);
+    connect(ui->checkBoxMidplane, &QCheckBox::toggled, this,
+        &TaskRevolutionParameters::onMidplane);
+    connect(ui->checkBoxReversed, &QCheckBox::toggled, this,
+            &TaskRevolutionParameters::onReversed);
+    connect(ui->checkBoxUpdateView, &QCheckBox::toggled, this,
+            &TaskRevolutionParameters::onUpdateView);
 }
 
 void TaskRevolutionParameters::updateUI()
@@ -353,9 +351,7 @@ TaskRevolutionParameters::~TaskRevolutionParameters()
         ex.ReportException();
     }
 
-    for (size_t i = 0; i < axesInList.size(); i++) {
-        delete axesInList[i];
-    }
+    axesInList.clear();
 }
 
 void TaskRevolutionParameters::changeEvent(QEvent *e)
--- a/src/Mod/PartDesign/Gui/TaskRevolutionParameters.h
+++ b/src/Mod/PartDesign/Gui/TaskRevolutionParameters.h
@@ -70,7 +70,6 @@ private Q_SLOTS:
 protected:
     void onSelectionChanged(const Gui::SelectionChanges& msg) override;
     void changeEvent(QEvent *e) override;
-    bool updateView() const;
     void getReferenceAxis(App::DocumentObject *&obj, std::vector<std::string> &sub) const;
     double getAngle() const;
     bool getMidplane() const;
@@ -88,8 +87,8 @@ private:
     void updateUI();
 
 private:
-    QWidget* proxy;
     std::unique_ptr<Ui_TaskRevolutionParameters> ui;
+    QWidget *proxy;
 
     /**
      * @brief axesInList is the list of links corresponding to axis combo; must
@@ -99,7 +98,7 @@ private:
      * It is a list of pointers, because properties prohibit assignment. Use new
      * when adding stuff, and delete when removing stuff.
      */
-    std::vector<App::PropertyLinkSub*> axesInList;
+    std::vector<std::unique_ptr<App::PropertyLinkSub>> axesInList;
 };
 
 /// simulation dialog for the TaskView


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719">pipeline status</a> for feature branch [PR_7719](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719). Pipeline [686544858](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/686544858) was triggered at [8f3f17f](https://github.com/FreeCAD/FreeCAD/pull/7719/commits/8f3f17f855d325d6045ff25968058c67714cb719). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719">pipeline status</a> for feature branch [PR_7719](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719). Pipeline [686553473](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/686553473) was triggered at [353fb5f](https://github.com/FreeCAD/FreeCAD/pull/7719/commits/353fb5fd8dcf1b2907ade3011fd8e16418770b0d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719">pipeline status</a> for feature branch [PR_7719](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719). Pipeline [686554640](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/686554640) was triggered at [daaed49](https://github.com/FreeCAD/FreeCAD/pull/7719/commits/daaed4987b7d0b299582843515e2057abf97b4ff). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -56,18 +56,18 @@ TaskRevolutionParameters::TaskRevolutionParameters(PartDesignGui::ViewProvider*
     QMetaObject::connectSlotsByName(this);
     this->groupLayout()->addWidget(proxy);
 
-    //bind property mirrors
-    PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
+    // bind property mirrors
+    auto pcFeat = dynamic_cast<PartDesign::ProfileBased *>(vp->getObject());
```
wwmayer, line 60: Use `auto*` instead of `auto` to make it clear that a pointer is used.

Replacing static_cast with dynamic_cast here is OK because from the view provider type `PartDesignGui::ViewProvider` it's not obvious that the object type should be `PartDesign::ProfileBased`. But you should also check that `pcFeat` is not null as otherwise a segfault is still possible.

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -56,18 +56,18 @@ TaskRevolutionParameters::TaskRevolutionParameters(PartDesignGui::ViewProvider*
     QMetaObject::connectSlotsByName(this);
     this->groupLayout()->addWidget(proxy);
 
-    //bind property mirrors
-    PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
+    // bind property mirrors
+    auto pcFeat = dynamic_cast<PartDesign::ProfileBased *>(vp->getObject());
     if (pcFeat->isDerivedFrom(PartDesign::Revolution::getClassTypeId())) {
-        PartDesign::Revolution* rev = static_cast<PartDesign::Revolution*>(vp->getObject());
+        auto *rev = dynamic_cast<PartDesign::Revolution *>(vp->getObject());
```
wwmayer, line 62: No need to replace the static_cast here because with the previous line this is already asserted.

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -56,18 +56,18 @@ TaskRevolutionParameters::TaskRevolutionParameters(PartDesignGui::ViewProvider*
...
         this->propAngle = &(rev->Angle);
         this->propMidPlane = &(rev->Midplane);
         this->propReferenceAxis = &(rev->ReferenceAxis);
         this->propReversed = &(rev->Reversed);
         ui->revolveAngle->bind(rev->Angle);
     } else {
         assert(pcFeat->isDerivedFrom(PartDesign::Groove::getClassTypeId()));
-        PartDesign::Groove* rev = static_cast<PartDesign::Groove*>(vp->getObject());
+        auto *rev = dynamic_cast<PartDesign::Groove *>(vp->getObject());
```
wwmayer, line 70: Same as above, no need to replace the static_assert. But this whole block should be written as:
```
...
else if (pcFeat->isDerivedFrom(PartDesign::Groove::getClassTypeId())) {
    auto* rev = static_cast<PartDesign::Groove*>(vp->getObject());
....
}
```

The alternative way would be dynamic_cast without the previous type check:
```
else if (auto *rev = dynamic_cast<PartDesign::Groove *>(vp->getObject())) {
...
}
```

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -88,13 +88,14 @@ TaskRevolutionParameters::TaskRevolutionParameters(PartDesignGui::ViewProvider*
...
 
-    //show the parts coordinate system axis for selection
+    // show the parts coordinate system axis for selection
     PartDesign::Body * body = PartDesign::Body::findBodyOf ( vp->getObject () );
     if(body) {
         try {
             App::Origin *origin = body->getOrigin();
-            ViewProviderOrigin* vpOrigin;
-            vpOrigin = static_cast<ViewProviderOrigin*>(Gui::Application::Instance->getViewProvider(origin));
+            ViewProviderOrigin* vpOrigin = nullptr;
```
wwmayer, line 96: vpOrigin should be directly initialized with
`ViewProviderOrigin* vpOrigin = static_cast<ViewProviderOrigin *>(...`

The dynamic_cast is not needed here because it won't happen that an App::Origin has a view provider different to ViewProviderOrigin

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -112,14 +113,17 @@ void TaskRevolutionParameters::fillAxisCombo(bool forceRefill)
     if (forceRefill){
         ui->axis->clear();
 
-        for(size_t i = 0; i < axesInList.size(); i++){
+        int i = 0;
```
wwmayer, line 116: That's a weird way iterating over the array.

I recommend to replace `std::vector<App::PropertyLinkSub*> axesInList;` with
`std::vector<std::unique_ptr<App::PropertyLinkSub>> axesInList;` and adjust the implementation accordingly. Then you don't have to iterate over the array when clearing it.

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -112,14 +113,17 @@ void TaskRevolutionParameters::fillAxisCombo(bool forceRefill)
...
             delete axesInList[i];
         }
+
         this->axesInList.clear();
 
         //add sketch axes
-        PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
-        Part::Part2DObject* pcSketch = dynamic_cast<Part::Part2DObject*>(pcFeat->Profile.getValue());
+        auto pcFeat = dynamic_cast<PartDesign::ProfileBased*>(vp->getObject());
+        auto pcSketch = dynamic_cast<Part::Part2DObject*>(pcFeat->Profile.getValue());
```
wwmayer, line 126: Same thing about auto and dynamic_cast

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: Yes, but the CI issues this warning nevertheless. What would be the disadvantage of using dynamic_cast?

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: done

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: Since we only have 2 cases and check for a nullptr, I thought keeping the existing if...else is fine. Why do you prefere an else if here?

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: done

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: Done, but this is a new C++ field for me (never used std::make_unique yet), can you therefore please check if I did it right?

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: done

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
wwmayer: IMO we shouldn't blindly follow what the CI says. It only follows some static rules but doesn't know much about the context.  If the check `if (pcFeat->isDerivedFrom(PartDesign::Revolution::getClassTypeId()))` is true then it's already guaranteed that it's the right type.

The downside of using dynamic_cast here is that two expensive type checks are performed instead of only one.

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
wwmayer: Who gives you give the guarantee that a `PartDesign::ProfileBased` must either be a `PartDesign::Revolution` or a `PartDesign::Groove`? If for some reason the wrong type is passed to the constructor of TaskRevolutionParameters then it will cause a crash. So, ideally the constructor should throw a Base::TypeError if the wrong type is passed:
```
if (/*is revolution*/) {
...
}
else if (/*is groove*/) {
...
}
else {
    throw Base::TypeError("Add some useful text here");
}
```
Doing it this way requires one further small change: the creation of the proxy widget must be moved to the initialization list to avoid a memory leak.

donovaly: Thanks. I made this now, but I am not sure I did it right,

donovaly:
I rebased the PR and combined all changes into one commit. Can you please have another look?

Line 182 is suspicious because it compiles without warning but the MSVC IDE tells me sometimes (sic!) that the arguments of push_back contain also a destructor and therefor does not match the content of the vector.

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: I understand but wanted to know how expensive this is. So my idea is maybe wrong: accepting a single additional type check for the sake of not being annoyed by warnings. (That catch attention of devs and in the end cost us unnecessary spare time.)

wwmayer: > I understand but wanted to know how expensive this is

How expensive exactly I don't know. It depends on how the compiler implements it and the only way to check is write a test program and measure the time

wwmayer: A simple test application shows that dynamic_cast is by a factor 4-5 slower than static_cast -- but it's still in the range of 10ns.

Since isDerivedFrom() is slower than dynamic_cast by  factor of 2.4 the most efficient way is to write:
```
if (auto *rev = dynamic_cast<PartDesign::Revolution *>(vp->getObject()) {
...
}
```
The second best way is to write
```
if (pcFeat->isDerivedFrom(PartDesign::Revolution::getClassTypeId())) {
    auto* rev = static_cast<PartDesign::Revolution*>(vp->getObject());
}
```
and the worst way is to write
```
if (pcFeat->isDerivedFrom(PartDesign::Revolution::getClassTypeId())) {
    auto* rev = dynamic_cast<PartDesign::Revolution*>(vp->getObject());
}
```

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -47,33 +47,37 @@ using namespace Gui;
...
     ui->setupUi(proxy);
     QMetaObject::connectSlotsByName(this);
     this->groupLayout()->addWidget(proxy);
 
-    //bind property mirrors
-    PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
+    // bind property mirrors
+    auto *pcFeat = dynamic_cast<PartDesign::ProfileBased*>(vp->getObject());
+    if (!pcFeat)
+        return;
```
wwmayer, line 61: Returning here is a very bad way because it leaves a half-initialized object. Instead an exception must raised so that the compiler cleans it up again.

But because further down the actual type checks are done the lines 59-61 can be replaced with: `auto* pcFeat = vp->getObject()`

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -109,17 +114,15 @@ void TaskRevolutionParameters::fillAxisCombo(bool forceRefill)
...
-        }
-        this->axesInList.clear();
+        axesInList.clear();
 
         //add sketch axes
-        PartDesign::ProfileBased* pcFeat = static_cast<PartDesign::ProfileBased*>(vp->getObject());
-        Part::Part2DObject* pcSketch = dynamic_cast<Part::Part2DObject*>(pcFeat->Profile.getValue());
+        auto *pcFeat = dynamic_cast<PartDesign::ProfileBased*>(vp->getObject());
+        if (!pcFeat)
+            return;
```
wwmayer, line 124: Same here. Do not silently return from the constructor.

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -176,23 +179,23 @@ void TaskRevolutionParameters::addAxisToCombo(App::DocumentObject* linkObj,
                                               QString itemText)
 {
     this->ui->axis->addItem(itemText);
-    this->axesInList.push_back(new App::PropertyLinkSub());
+    this->axesInList.push_back(std::move(std::make_unique<App::PropertyLinkSub>()));
```
wwmayer, line 182: Write `this->axesInList.emplace_back();`

src/Mod/PartDesign/Gui/TaskRevolutionParameters.h
```diff
@@ -99,7 +99,7 @@ private Q_SLOTS:
      * It is a list of pointers, because properties prohibit assignment. Use new
      * when adding stuff, and delete when removing stuff.
```
wwmayer, line 100: To fix the above warning you can add:
```
TaskRevolutionParameters(const TaskRevolutionParameters&) = delete;
TaskRevolutionParameters(TaskRevolutionParameters&&) = delete;
void operator= (const TaskRevolutionParameters&) = delete;
void operator= (TaskRevolutionParameters&&) = delete;
```

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: Interesting. many thanks for this info.

I updated the code accordingly now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719">pipeline status</a> for feature branch [PR_7719](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719). Pipeline [689324297](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/689324297) was triggered at [21d8c81](https://github.com/FreeCAD/FreeCAD/pull/7719/commits/21d8c81a7dac532c9b47ee05a488a6717d1f0606). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
donovaly: I se but the whole pcFeat is not necessary at all. I removed it therefore now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719">pipeline status</a> for feature branch [PR_7719](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7719). Pipeline [689332454](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/689332454) was triggered at [418f5bd](https://github.com/FreeCAD/FreeCAD/pull/7719/commits/418f5bda405e38d6e56cbc5a60b77c147a9568c4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -47,33 +47,32 @@ using namespace Gui;
...
+    }
+    else if (auto *rev = dynamic_cast<PartDesign::Groove *>(vp->getObject())) {
         this->propAngle = &(rev->Angle);
         this->propMidPlane = &(rev->Midplane);
         this->propReferenceAxis = &(rev->ReferenceAxis);
         this->propReversed = &(rev->Reversed);
         ui->revolveAngle->bind(rev->Angle);
     }
+    else {
+        throw Base::TypeError("The ViewProvider is not a Groove or a Revolution.");
```
wwmayer, line 74: > The ViewProvider is not a Groove or a Revolution.

And will never be :) It's about the document object type.

wwmayer:
I think it's OK to merge now. I will change the wording of the exception and some of the reported warnings afterwards.

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
wwmayer: I realized a problem with the above hint. It should be `this->axesInList.emplace_back(new App::PropertyLinkSub());`because now it creates a unique pointer with null. This will result into a crash. I will fix it.

PR #7801, opened 2022-11-15
Discussion: Refactor some legacy features in splashscreen

berniev:

---

For discussion.

 I take Splashscreen file pair and describe some features I would describe as "legacy", and offer
 possible solutions. Untested. Achieved by many small refactorings.

 1. A block of code repeated 18 times (!), with varying data.

 2. Another block of code repeated multiple times almost identically.

 3a. makeSummaryReport() has many (and tangled) responsibilities.

 3b. makeSummaryReport() pulls information from all over the place.

 4. SplashObserver involved in text color and alignment.

 5. UPPER_CASE constexpr looks like a macro.

 6. Lots of type juggling esp std <==> Q <==> char.

 7. App::Application::Config() returns a variable which is operated on all over the place.

To address these (making minimum changes to overall functioning of code):

 1. Move the data part into an array (which could ultimately come from anywhere), and create a
    single function.

 2. Make one common function with paramters to handle minor variations.

 3a. Many refactorings to untangle responsibilities into sub-scopes, using closures . It became
     obvious that most of these responsibilities actually belonged elsewhere.
     The core purpose of the function was now exposed: Prepare a string. Only.

 3b. The dependencies should be moved out. Interim step create SummaryReport class.

 4. Move format stuff out of observer. It is an unwanted extra responsibility.

 5. Lodged issue on FC github, with view to PR (assuming no objection).

 6. Once responsibilities are in discrete scopes they often become much simpler and many type
    juggles superfluous because they were previously pandering to surrounding code. Often simple
    std types were much more readable and direct. As a result much code was suddenly more readable.

 7. Not tackled, but it seems obvious config should be a separate class or struct with associated
    operations included.

 Suggested reading: https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29

### DIFF
--- a/src/Gui/Splashscreen.cpp
+++ b/src/Gui/Splashscreen.cpp
@@ -23,21 +23,21 @@
 #include "PreCompiled.h"
 
 #ifndef _PreComp_
-# include <cstdlib>
-# include <QApplication>
-# include <QClipboard>
-# include <QFile>
-# include <QLocale>
-# include <QMutex>
-# include <QProcessEnvironment>
-# include <QRegularExpression>
-# include <QRegularExpressionMatch>
-# include <QScreen>
-# include <QSysInfo>
-# include <QTextBrowser>
-# include <QTextStream>
-# include <QWaitCondition>
-# include <Inventor/C/basic.h>
+#include <Inventor/C/basic.h>
+#include <QApplication>
+#include <QClipboard>
+#include <QFile>
+#include <QLocale>
+#include <QMutex>
+#include <QProcessEnvironment>
+#include <QRegularExpression>
+#include <QRegularExpressionMatch>
+#include <QScreen>
+#include <QSysInfo>
+#include <QTextBrowser>
+#include <QTextStream>
+#include <QWaitCondition>
+#include <cstdlib>
 #endif
 
 #include <App/Application.h>
@@ -45,75 +45,39 @@
 #include <Base/Console.h>
 #include <CXX/WrapPython.h>
 
-#include <boost/filesystem.hpp>
 #include <LibraryVersions.h>
+#include <boost/filesystem.hpp>
 #include <zlib.h>
 
+#include "MainWindow.h"
 #include "Splashscreen.h"
 #include "ui_AboutApplication.h"
-#include "MainWindow.h"
 
+#include <sstream>
+#include <strstream>
 
 using namespace Gui;
 using namespace Gui::Dialog;
 namespace fs = boost::filesystem;
 
-namespace Gui {
+namespace Gui
+{
+
+
 /** Displays all messages at startup inside the splash screen.
  * \author Werner Mayer
  */
-class SplashObserver : public Base::ILogger
+class SplashObserver: public Base::ILogger
 {
 public:
     SplashObserver(const SplashObserver&) = delete;
     SplashObserver(SplashObserver&&) = delete;
-    SplashObserver& operator= (const SplashObserver&) = delete;
-    SplashObserver& operator= (SplashObserver&&) = delete;
+    SplashObserver& operator=(const SplashObserver&) = delete;
+    SplashObserver& operator=(SplashObserver&&) = delete;
 
-    explicit SplashObserver(QSplashScreen* splasher=nullptr)
-      : splash(splasher)
-      , alignment(Qt::AlignBottom|Qt::AlignLeft)
-      , textColor(Qt::black)
+    explicit SplashObserver(QSplashScreen& splasher) : splash {splasher}
     {
         Base::Console().AttachObserver(this);
-
-        // allow to customize text position and color
-        const std::map<std::string,std::string>& cfg = App::Application::Config();
-        auto al = cfg.find("SplashAlignment");
-        if (al != cfg.end()) {
-            QString alt = QString::fromLatin1(al->second.c_str());
-            int align=0;
-            if (alt.startsWith(QLatin1String("VCenter"))) {
-                align = Qt::AlignVCenter;
-            }
-            else if (alt.startsWith(QLatin1String("Top"))) {
-                align = Qt::AlignTop;
-            }
-            else {
-                align = Qt::AlignBottom;
-            }
-
-            if (alt.endsWith(QLatin1String("HCenter"))) {
-                align += Qt::AlignHCenter;
-            }
-            else if (alt.endsWith(QLatin1String("Right"))) {
-                align += Qt::AlignRight;
-            }
-            else {
-                align += Qt::AlignLeft;
-            }
-
-            alignment = align;
-        }
-
-        // choose text color
-        auto tc = cfg.find("SplashTextColor");
-        if (tc != cfg.end()) {
-            QColor col; col.setNamedColor(QString::fromLatin1(tc->second.c_str()));
-            if (col.isValid()) {
-                textColor = col;
-            }
-        }
     }
     ~SplashObserver() override
     {
@@ -123,6 +87,7 @@ public:
     {
         return "SplashObserver";
     }
+
     void SendLog(const std::string& msg, Base::LogStyle level) override
     {
 #ifdef FC_DEBUG
@@ -134,7 +99,7 @@ public:
         }
 #endif
     }
-    void Log (const char * text)
+    void Log(const char* text)
     {
         QString msg(QString::fromUtf8(text));
         QRegularExpression rx;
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7801">pipeline status</a> for feature branch [PR_7801](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7801). Pipeline [694833165](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/694833165) was triggered at [f13e693](https://github.com/FreeCAD/FreeCAD/pull/7801/commits/f13e693889855970072e1c710daf9aa1d17fb0ca). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/Splashscreen.cpp
```diff
@@ -482,6 +525,10 @@ void AboutDialog::showLicenseInformation()
     }
 }
 
+/**
+ * A whole function just for additional license information?
+ * Is this really necessary?
+  */
```
luzpaz, line 531: formatting (is offset by 1 whitespace)

src/Gui/Splashscreen.h
```diff
@@ -66,7 +76,7 @@ class GuiExport AboutDialogFactory
     static void setDefaultFactory(AboutDialogFactory *factory);
 
 private:
-    static AboutDialogFactory* factory;
+    static AboutDialogFactory* factory {};
```
wwmayer, line 79: That is not allowed. A static member variable must be initialized in the source file the way how it was before.

src/Gui/Splashscreen.h
```diff
@@ -86,6 +96,17 @@ class GuiExport LicenseView : public Gui::MDIView
     QTextBrowser* browser;
 };
 
+class SummaryReport{
+public:
+    SummaryReport();
+    ~SummaryReport();
+    void addItem(const std::string item);
+    std::string asStdString();
```
wwmayer, line 104: Method should be declared as const.

src/Gui/Splashscreen.cpp
```diff
@@ -148,76 +113,143 @@ class SplashObserver : public Base::ILogger
...
+void SplashScreen::drawContents(QPainter* painter)
 {
     QSplashScreen::drawContents(painter);
 }
 
 // ------------------------------------------------------------------------------
 
-AboutDialogFactory* AboutDialogFactory::factory = nullptr;
-
-AboutDialogFactory::~AboutDialogFactory()
```
wwmayer, line 199: What happened to the definition of the destructor? I cannot find it anywhere. Since AboutDialogFactory has a virtual method and doesn't inherit from another class it must have a virtual destructor.

src/Gui/Splashscreen.cpp
```diff
@@ -235,10 +267,12 @@ void AboutDialogFactory::setDefaultFactory(AboutDialogFactory *f)
  *  The dialog will be modal.
  */
 AboutDialog::AboutDialog(bool showLic, QWidget* parent)
-  : QDialog(parent), ui(new Ui_AboutApplication)
+    : QDialog(parent)
 {
     Q_UNUSED(showLic);
 
+    ui = new Ui_AboutApplication;
```
wwmayer, line 274: Nope! https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-initialize

src/Gui/Splashscreen.cpp
```diff
@@ -235,10 +267,12 @@ void AboutDialogFactory::setDefaultFactory(AboutDialogFactory *f)
  *  The dialog will be modal.
  */
 AboutDialog::AboutDialog(bool showLic, QWidget* parent)
-  : QDialog(parent), ui(new Ui_AboutApplication)
+    : QDialog(parent)
 {
     Q_UNUSED(showLic);
 
+    ui = new Ui_AboutApplication;
+    summaryReport new SummaryReport;
```
wwmayer, line 275: Invalid syntax. `summaryReport = new SummaryReport;`
Like above this should go to the initialization list

src/Gui/Splashscreen.cpp
```diff
@@ -493,343 +540,397 @@ QString AboutDialog::getAdditionalLicenseInformation() const
...
-        QLatin1String("FreeType"),
-        baseurl + QLatin1String("#_TocFreeType"),
-        QLatin1String("https://freetype.org"),
-        QString::fromLatin1(FC_FREETYPE_VERSION)
+    auto deskInfo = [&sysEnv] {
+        std::string deskEnv = sysEnv("XDG_CURRENT_DESKTOP");
+        std::string deskSess = sysEnv("DESKTOP_SESSION");
+
+        std::string result {};
+        if (!(deskEnv.empty() && deskSess.empty())) {
```
wwmayer, line 706: According to the cited [Clean Code Rules](https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29) negative conditionals should be avoided. Instead: `if (!deskEnv.empty() || !deskSess.empty()) {`

src/Gui/Splashscreen.cpp
```diff
@@ -493,343 +540,397 @@ QString AboutDialog::getAdditionalLicenseInformation() const
...
+            result += " (" + deskEnv;
```
wwmayer, line 707: Are you sure? How do you know that deskEnv is not empty? So far it has only be verified that both deskEnv and deskSess are not empty but one of them can still be empty.

src/Gui/Splashscreen.cpp
```diff
@@ -148,76 +113,143 @@ class SplashObserver : public Base::ILogger
...
+}
+
+std::string SplashScreen::fromConfig(const std::map<std::string, std::string>& config,
+                                     const std::string value)
+{
+    auto al = config.find(value);
+    return al == config.end() ? "" : al->second.c_str();
+}
+
+QColor SplashScreen::colorFromString(const std::string str)
```
wwmayer, line 187: Shouldn't it be `const std::string& str`? From what I have read a while ago in C++11 you can also declare it `std::string str` but then in the implementation you should use `std::move(str)` to avoid to create a tmp. copy.

berniev:
@wwmayer Thanks for the feedback! I did mention untested (not meant as mergeable), and you know I'm learning C++ so all tips appreciated. But the focus was not the language minutae.

The purpose is to demonstrate refactoring techniques to move towards testable code.

Based on negative forum reactions, "legacy code" is not a widely understood term at FC, much less acknowledged, despite being used industry wide in various forms for many years. There is much literature.

I can but hope this helps promote some awareness and positive discussion.

wwmayer:
> The purpose is to demonstrate refactoring techniques to move towards testable code.

Thanks for this endeavor. 

> Based on negative forum reactions, "legacy code" is not a widely understood term at FC, much less acknowledged, despite being used industry wide in various forms for many years.

Well, this is a difficult topic and it won't happen over night to convince the people and maintainers to apply new programming techniques. When reading the corresponding forum thread I am not sure how wide spread e.g. TDD is. At least from my side I can claim that the companies I work(ed) for this has zero relevance.

So, as already discussed some months ago in the forum the best is to set up the basic infrastructure for TDD and pick a few example classes to migrate. For the rest of the code a long breath is needed which I guess will take several years.

> There is much literature.

Is there a good book that you can recommend? It should provide some good tips and tricks to transform legacy code into modern code.

berniev:
Thank you for being candid, and your kind words. Its been a difficult journey.

BTW I did reply to each of your points and did push back on a few, but it seems I still havn't learnt to press save or whatever.

There's no "quick fix". You have to use a new set of principles with an ultimate goal of testability.

I found the following useful (some big names in there):

- Working Effectively with Legacy Code [2005, Michael C. Feathers]
- The Legacy Code Programmer's Toolbox [2019, Jonathan Boccara]

-  Refactoring. Improving the design of legacy code [2019, Martin Fowler]

- Test-Driven Development By Example [2003, Kent Beck] (Might be a bit ambitious, but there is sound logic there).

I do have a commit ready for basic GoogleTest structure but didn't want to appear presumptious.
 
Mentioning testing leads straight to another concept that is not well understood at FC: Unit testing. Much confused with Integration Testing or Functional Testing, among other names. Whilst TDD uses unit testing, unit testing does not necessariuly imply TDD.

For what constitutes modern best practise see any of the Uncle Bob (Martin) books, videos etc or any discussion about SOLID principles. There's heaps out there.

wwmayer:
> I found the following useful (some big names in there):

Many thanks for the list.

> Mentioning testing leads straight to another concept that is not well understood at FC: Unit testing. Much confused with Integration Testing or Functional Testing, among other names.

True, and you can put me on this list too. Maybe I am wrong but I think that the term "unit testing" has changed its meaning somewhat over the years because from the very beginning on we focused on Python's testing framework and I was not aware of terms like Integration Testing or Functional Testing at that time but everything was just called Unit Testing.

Anyway, I am always eager to extend my horizon and one of my long-term goals is to integrate real unit testing into FreeCAD because with the current integration tests it's really difficult to achieve a good test coverage.

> For what constitutes modern best practise see any of the Uncle Bob (Martin) books, videos etc or any discussion about SOLID principles. There's heaps out there.

Yes, some years ago I watched some of his videos on YT and they are very inspiring.

berniev:
As Jonathan Boccara points out, there's a lot of legacy code out there!! It is not evil (fragile, perhaps), just something to deal with. I stand in awe of your ability to stay across such a complex app as FC.

Years ago the penny dropped for me and transformed the way I look at code. It really was a Eureka moment. I wrote a dependency injection container (with auto wiring) and refactored that application completely inside out. Difficult to do in C++ though (no reflection).

Going forward though, how many FC devs get what we're talking about? Based on serious push-back by senior devs recently, I suspect none. Understandable in the sense that legacy code will attract legacy coders, but was a shock for me.

How to go forward? I suggest while absorbing all this the first priority is uniform formatting. With refactoring there are loads of code changes and realistically the only way out is hit the format button. Formatting is totally unproductive and yet draws so much attention. We have tools for that, and IMO more important things to worry about.

wwmayer:
> How to go forward? I suggest while absorbing all this the first priority is uniform formatting. With refactoring there are loads of code changes and realistically the only way out is hit the format button. Formatting is totally unproductive and yet draws so much attention. We have tools for that, and IMO more important things to worry about.

As long as the topo naming branch isn't merged yet we should avoid to change the whole code base as this only causes a lot of extra work for the maintainers of the branch. But parts of the code base that aren't affected by this branch could already be changed now.
But before doing that it's probably good to see the output of some formatted files because there might be some formatting options that not everybody is happy with and for them we have to reach consensus.

berniev:
From what I've read the best way is all at once, so yes, you want to be sure the settings are what you want. Also have to consider how to fit into CI workflow. There was a forum discussion a while ago, but it died. Also, maybe need a test file.

What is the TN timeline looking like?

wwmayer:
> What is the TN timeline looking like?

I don't know. But I guess that it requires a major code-refactoring because there were already complaints that everything is too tightly coupled.

berniev:
I have created format test file #7858

berniev:
Closed as has met its purpose and generated meaningful discussion about modern programming techniques.

PR #7902, opened 2022-11-29
Base: Simplify repetitious code

berniev:

---
Don't repeat yourself (DRY)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7902">pipeline status</a> for feature branch [PR_7902](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7902). Pipeline [707612328](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/707612328) was triggered at [ea175a7](https://github.com/FreeCAD/FreeCAD/pull/7902/commits/ea175a771a77e40b90340f9ccd7b20fe8cfa4d34). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
@berniev how about having no default case and keep the latest line of the function returning it ? So if in the future one needs a case which just break, it works directly ?

berniev:
I thought about that, and indeed then just get rid of the switch but that would require an additional variable (which is almost always a bad thing) or replicating the function call getNodeType() which is often a bad thing. Also, code checkers will look at a switch and detect if there is missing options (eg if the test is a enum), so the case / default works well (IMHO).

src/Base/Parameter.cpp
```diff
@@ -1526,34 +1526,14 @@ DOMPrintFilter::FilterAction DOMPrintFilter::acceptNode(const DOMNode* node) con
...
-        break;
-    }
+        case DOMNode::DOCUMENT_TYPE_NODE:
+        case DOMNode::DOCUMENT_NODE: {
+            return DOMNodeFilter::FILTER_REJECT;// no effect
+        }
+        default: {
+            return DOMNodeFilter::FILTER_ACCEPT;
+        }
     }
```
0penBrain, line 1536: I was thinking to something like that :
```suggestion
        default: {
            // Let return function default value
        }
    }
    return DOMNodeFilter::FILTER_ACCEPT;
```

berniev: But why create an empty default and then separately a return? What is gained by the extra code?

0penBrain: It's clearer what is the function default return.
And if one want to add a new case like 
```
case foo:
    if (bar):
        break;
    else:
       foobar();
    [[fallthrough]]
```
That directly works.

berniev:
Thank you for reviewing this PR, but I can't see your suggestions offer any technical improvement.

Input is a NodeType. NodeType is an enum of 12 elements.
Output is a FilterAction. FilterAction is an enum of 3 elements. Only two are used.

The code specifies a result for some, but not all, of the possible inputs.
Not all possible enum cases are defined so a default is essential.

As a default is essential, and there are only two possible output values, the logic can be reduced to specifying only the _non-default_ items. That's just two NodeTypes.

if-then-else-if-else etc could be used but gets ugly pretty quick and either a) repeats the retrieval of the input, or b) requires an additional variable, and denies code checkers the ability to ensure all enum elements are covered.

if-else with booleans isn't much better.

A switch is the perfect structure here as the input is evaluated once and all logic is contained entirely within the one switch statement. Code checkers can warn if you miss cases. Cases can be easily added etc. Cases of like output are simply placed together (no break between) and fallthrough to the same code. Simple and concise.

The original has an additional return after the switch statement. All cases have a return so this return is unreachable code. Note also the redundant break statements in the original.

The alternative is add an intermediate output result variable but that is unnecessary complexity.

donovaly:
To me the PR looks good as it is - it removes the repetitions and keeps the functionality.

berniev:
@donovaly Thanks, but I think I've messed up (I cant prevent all those commits being sent). If you don't mind I'll come back afresh.

berniev:
See #7952

PR #7955, opened 2022-12-04
Unify version patch number management in cmake

adrianinsaval:
Current setup is confusing and inconsistent, see #7284
this can lead to incorrect version info in some places

I also remove a couple of files that look unused, one (src/Main/freecad.rc) looks like a generated file to me and the other (src/Build/BuildVersion.bat) is 2011 single line script that probably nobody uses anymore and even has a hardcoded path to an executable, it's probably from back when FreeCAD used SVN.

@donovaly I suggest backporting this to the 0.20 branch (at least the first commit) the third commit was cherry-picked from the 0.20 branch ddbcc0b3d58db90b9217ed3b8e2781338184646a


adrianinsaval:
what's missing is output of point release info when calling `FreeCAD.Version()` but that could potentially break scripts that use it

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7955">pipeline status</a> for feature branch [PR_7955](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7955). Pipeline [712917563](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/712917563) was triggered at [0d67775](https://github.com/FreeCAD/FreeCAD/pull/7955/commits/0d67775b931d67ed9e84ba933fd7878eaf7606c9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

CMakeLists.txt
```diff
@@ -18,11 +18,10 @@ project(FreeCAD)
 set(PACKAGE_VERSION_NAME "Vulcan")
 set(PACKAGE_VERSION_MAJOR "0")
 set(PACKAGE_VERSION_MINOR "21")
-set(PACKAGE_VERSION_PATCH "0")
+set(PACKAGE_VERSION_PATCH "0") # number of patch release (e.g. "4" for the 0.18.4 release)
 set(PACKAGE_VERSION_SUFFIX "dev") # either "dev" for development snapshot or "" (empty string)
-set(FREECAD_VERSION_PATCH "0") # number of patch release (e.g. "4" for the 0.18.4 release)
```
adrianinsaval, line 23: does somebody know if this is something that may be used by other programs that link freecad as a library?

donovaly: I just checked today around because I had the same idea as you and wondered why there is FREECAD_VERSION_PATCH and PACKAGE_VERSION_PATCH.

I found no reason,, so let's give it a try. However, to be safe, not for the 0.20 branch.

donovaly:
Oops, I was too rush. The file freecad.rc is necessary because otherwise the compiled .exe on Windows does not have version info. I will bring it back.

adrianinsaval:
but isn't that generated by cmake from freecad.rc.cmake?

donovaly:
Yes, but we need the file as such meaning already being generated. I will have a closer look once 0.20.2 is uploaded and I return to master. Maybe I just have to adapt a build script.

adrianinsaval:
> Yes, but we need the file as such meaning already being generated. I will have a closer look once 0.20.2 is uploaded and I return to master. Maybe I just have to adapt a build script.

the one in the repo does not have the info in it, it also just references the cmake variables (and is not the same as freecad.rc.cmake) , doesn't cmake create this file when it's run? there is no freecadCmd.rc file in the repo for example, but there is a freecadCmd.rc.cmake

adrianinsaval:
indeed:
https://github.com/FreeCAD/FreeCAD/blob/9869a2c3d7bdacdfac605db318c73e3b2d71245c/src/Main/CMakeLists.txt#L4-L5
https://cmake.org/cmake/help/latest/command/configure_file.html
that file is not used, it just gets replaced by the configured file generated by cmake

donovaly:
OK, so the removal was OK then.

PR #7969, opened 2022-12-07
Draft: fix make_sketch bugs

Roy-043:
This PR fixes 3 make_sketch bugs:

1. The last segment of an unrotated Rectangle was not constrained horizontally/vertically.
2. The `isAligned` function worked without a tolerance. This meant that none of the edges of a 90 degree rotated Rectangle was constrained horizontally/vertically.
3. Filleted Wires, Rectangles and Polygons where skipped. [Forum topic](https://forum.freecadweb.org/viewtopic.php?f=23&t=74194).

---

### DIFF
--- a/src/Mod/Draft/draftgeoutils/general.py
+++ b/src/Mod/Draft/draftgeoutils/general.py
@@ -148,32 +148,30 @@ def isAligned(edge, axis="x"):
 
     The axis can be 'x', 'y' or 'z'.
     """
+    def is_same(a, b):
+      return round(a, precision()) == round(b, precision())
+
     if axis == "x":
         if isinstance(edge, Part.Edge):
             if len(edge.Vertexes) == 2:
-                if edge.Vertexes[0].X == edge.Vertexes[-1].X:
-                    return True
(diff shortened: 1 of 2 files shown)


src/Mod/Draft/draftgeoutils/general.py
```diff
@@ -148,32 +148,30 @@ def isAligned(edge, axis="x"):
 
     The axis can be 'x', 'y' or 'z'.
     """
+    def is_same(a, b):
+      return round(a, precision()) == round(b, precision())
```
0penBrain, line 152: Any reason to not use something like ` return abs(a-b) < precision()` ?

Roy-043: If I am not mistaken it would then have to be:
`return abs(a-b) < 10 ** (-precision() - 1)`

In other parts of the file `round` is used though. I believe you would call this 'local consistency'... But both solutions are fine by me.

Roy-043: What I don't like about the `isAligned` function is the repetition of (almost) the same code block for x, y and z.

PR #7979, opened 2022-12-07
[MeshPart] improve PreCompiled.h handling

donovaly:


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7979">pipeline status</a> for feature branch [PR_7979](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7979). Pipeline [716473064](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/716473064) was triggered at [0df2255](https://github.com/FreeCAD/FreeCAD/pull/7979/commits/0df22551e2f193c17c073926a598bc61c007ca49). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

looooo:
@donovaly boost/python is optional here. FreeCAD can be build with pybind11 aswell.

donovaly:
> @donovaly boost/python is optional here. FreeCAD can be build with pybind11 aswell.

I compile always with pybind11 and also on 2 machines, on one with PCH, on the other one without. I get no problems. I use boost 1.67.

wwmayer:
I think this change should be reverted because it's a regression.

MeshFlatteningBoostPython.cpp is only built if the option BUILD_FLAT_MESH is set and FREECAD_USE_PYBIND11 is off. In this case there is a check for the boost python package while now boost python headers are always included even if not needed and more important without searching for the corresponding package.

It still works fine on systems where boost/python.hpp is in a standard include path but on system where the headers are located in non-standard paths the build will fail.

donovaly:
> I think this change should be (partially) reverted because it's a regression.

OK then. Just please only revert the conflicting part since the other changes of this PR cost me some time.

wwmayer:
70a85872a9

PR #7990, opened 2022-12-08
[Image] App: remove precompiled header

donovaly:
- there is not a single header that could be precompiled, thus remove PreCompiled.cpp/h

@wwmayer , can you please have a quick look if this change is OK?


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7990">pipeline status</a> for feature branch [PR_7990](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_7990). Pipeline [717646870](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/717646870) was triggered at [1d20874](https://github.com/FreeCAD/FreeCAD/pull/7990/commits/1d2087407acb5dbaa91a9453f37f1e66b74ac0d2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
There is no reason to handle things differently to all other modules. The overhead for the compiler to open an extra file is so small that you probably won't be able to measure it.

donovaly:
> There is no reason to handle things differently to all other modules. The overhead for the compiler to open an extra file is so small that you probably won't be able to measure it.

OK, I enabled Precompiled headers now like for all other WBs: #7999

PR #8028, opened 2022-12-12
TopoNaming: fix lost of element reference tracking on undo/redo

realthunder:
Fixed potential lost of element reference tracking after undo/redo.

@donovaly @chennes please check.


donovaly:
Thanks, looks good.

Issue #8018, opened 2022-12-11
[Feature] CAM: Implement G84 Canned Tapping Cycle


luvtofish:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=15&t=74171

### Subproject(s) affected?

Path

### Idea description

Tapping on a CNC is every bit as crucial as the drilling cycles. Rigid tapping aside, just having the ability to perform a rudimentary G84 canned cycle with a floating tap holder tool would close a huge functionality gap in Path.

Needed functionality would include: 

- ability to enter tap size (eg. 1/4-20, M6-1, etc)
- user inputted spindle speed (coupled with tap size, used to calculate feed rate in/pm & mm/pm)
- Z depth
- R Plane (safeHeight) and optional O plane (clearanceHeight) for clearing obstacles
- optional dwell entry in seconds.
- define G84/G80/ and optional dwell code, and all xyz movements. 

Required spindle stop / reverse function is handled by the control and is triggered by the G84 call. 

### Anything else?

typical GCODE example:
![image](https://user-images.githubusercontent.com/20867664/206919686-2d681f82-96b9-4839-b60f-49f86dd395ed.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

sliptonic:
Thanks for the feature request.  A couple of thoughts about this off the top of my head:

A user would create a tap tool using the toolbit system.  We don't have a standard tap tool shape but the toolbit system will support everything needed to create one.  This is pretty simple and is a good milestone task.

Spindle speed is declared in the tool controller so nothing should be needed there.  Calculating the feed rate is also an independent unit of work from the tapping operation. I would consider it out of scope for this feature.

The R plane and O plane should be derived from the safe height and clearance height properties common to all operations.  Some logic to derive it may be necessary but we should not need to add any properties to the operation.

That means most of what's required is already done.   A tapping operation should work on essentially the same kind of geometry as a drilling operation.  So we can copy src/Mod/Path/Path/Op/Drilling.py and use it as the starting point.  

[...]

Issue #5714, opened 2022-02-07
Path: Add Tapping tool


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=3099

* **Reporter:** Konstantin
* **Date submitted:** 6/24/2017
* **FreeCAD version:** 0.18
* **Category:** Feature
* **Status:** acknowledged
* **Tags:** 

# Original report text

I want Tapping tool.

* I see an ability to add a tap in Tool Table, but there's no Pitch option.
* There's a few comands for rigid tapping:

* Tapping cycle (mostly G84)
* Counter tapping cycle (on some machines G84 with M4, on my Fanuc G74, maybe others commands used to, not sure)
* Peck tapping (on some controllers only need to add Q depth to G84, on some used different command)
* Counter peck tapping (same as previous)

* On some controllers F is a pitch of a tap, on some it must be set as S*pitch. So, there must be an option for that in Postprocessor.

# Other bug information

* **Priority:** normal
* **Severity:** feature
* **Category:** Feature
* **Updated:** 2/6/2021

# Discussion from Mantis ticket

### Comment by Kunda1 2017-06-28 21:45
Changed to 'Feature'

---

### Comment by @luzpaz 2017-07-22 09:57
Related forum thread: https://forum.freecadweb.org/viewtopic.php?f=15&t=23127&start=10#p183405

---

### Comment by @sliptonic 2017-11-28 00:36
feature moving to 0.18 version target

---

### Comment by @luzpaz 2019-01-23 00:51
bumped to 0.19

---

### Comment by @sliptonic 2021-02-03 18:47
bumped to 0.20.  

This has gotten some discussion on the gitter channel in recent weeks. The operation itself shouldn't be too difficult since much of it can build on the circularhole base classes.  

The more interesting items will be whether it can detect threading information from the geometry and whether it can use toolbit attributes to ensure a proper tool is selected.

luzpaz:
@luvtofish does #8069 impact this ticket ?

luvtofish:
8069 will take care of that ticket, yes.

sliptonic:
Toolbit functionality allows creation of tapping tools.  Threadmill operation is also new since this was created.

PR #8069, opened 2022-12-21
Path: Add G84/G74 Tapping Operation 

luvtofish:

---


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [728431989](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/728431989) was triggered at [3fafa7d](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/3fafa7d70c378f126d8dc9cac7cc5f7236f84ed2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [728970388](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/728970388) was triggered at [87b2116](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/87b21163667a30a81883c285ce61c2070ad460db). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [733120296](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/733120296) was triggered at [c54320e](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/c54320e82772b7520b342f7ae2998269dfad6c7a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [733640333](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/733640333) was triggered at [2c6925c](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/2c6925c3ede2714d790357da6dca783a9d8288a9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luvtofish:
@sliptonic I’m at a point where I need to enter signals/slots for 3 new widgets on the tapping gui. They are:

Thread Type - combo box with “right hand”, “left hand” items
Feed rate - QLineEdit field with input mask limiting input.
RPM - QlineEdit field with input mask limiting input.

I’m struggling with where / how to update the defined tapping tool controller spindle speed, vertical feed, and spindle direction properties. Any insight you can provide would be much appreciated. TIA. 

Dan

ondsel-brad:
Actually, I don't think you need any of them in the operation.
The thread type should be an attribute of the tap.  The operation can query the assocated toolcontroller/tool to see what it's thread type is.

Same thing with the Feed Rate and spindle speed.  Those are already defined on the TC.

Otherwise, this is looking great.  

Please put the tapping icon together with drilling the way vcarve/deburr/engrave work.   I'm very protective of the toolbar and screen real estate.

sliptonic:
Actually, I don't think you need any of them in the operation.
The thread type should be an attribute of the tap.  The operation can query the assocated toolcontroller/tool to see what it's thread type is.

Same thing with the Feed Rate and spindle speed.  Those are already defined on the TC.

Otherwise, this is looking great.  

Please put the tapping icon together with drilling the way vcarve/deburr/engrave work.   I'm very protective of the toolbar and screen real estate.

luvtofish:
Thanks for the kind words. I’m going to rework the icon a bit to make it
emphasize the threads better. As suggested, I’ll work on moving it to the
drill icon as a drop down.

I’d like to make the spindle speed and feed at least visible on the tapping
operation. They could be made readonly and pulled from the TC. Those two
items are so integral to the tapping motion. A bad calculation here equals
a broken tap and the potential for a scrapped part. I’m concerned the
endusers will miss the criticality of this.

sliptonic:
> I’d like to make the spindle speed and feed at least visible on the tapping operation. They could be made readonly and pulled from the TC. Those two items are so integral to the tapping motion. A bad calculation here equals a broken tap and the potential for a scrapped part. I’m concerned the endusers will miss the criticality of this.

Understood.  But resist the temptation to solve a problem you don't yet have.
If you feel strongly that the user needs to see this information while editing the operation, Then display it in a label. If you put it into a read-only control, the user will invariably ask to be able to change it here as well, rather than going 'all the way' to the TC.  That will lead to more coupling and poor separation of concerns.

Under no conditions should the operation have duplicate properties or store the same information again.

luvtofish:
Okay this all makes sense. And as I think about it, none of the other
operations do this. I think I might be trying to solve a problem that
doesn’t yet exist.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [737197126](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/737197126) was triggered at [8e9cc4c](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/8e9cc4c0a028fc5e20b387e8b08bdc46942ab672). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [737443625](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/737443625) was triggered at [c9cb670](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/c9cb670a441ce741eb25f0f961b8f8994e963d81). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [738515415](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/738515415) was triggered at [b45caeb](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/b45caebd37964cb3a5e4b93826c5e4ff8fd8990e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luvtofish:
@sliptonic — I’ve removed the entries for rpm, thread type and feed rate from the tapping panel as suggested. I’ve created an enumeration attribute on the tap shape file which provides the ability to set the thread type to Right Hand” or “Left Hand” with RH being the default. I now need to be able to check this setting and issue either a G74 or G84 cmd via the tapping.py generate function. I’m scratching my head trying to figure out how to check this attribute in the generate function of tapping.py.  I’ll place a conditional statement which will append the correct G-Code depending on the value contained in this attribute. Thoughts?

sliptonic:
Awesome.  I'll look closer at it tonight.  Getting the attribute should be straightforward.
It will appear like any other property.  If you get a reference to the tool controller, you could do something like
`
>>> tc.Name
'TC__mytap'
>>> print(tc.Tool.Rotation)
Right Hand
>>> 
`

luvtofish:
@sliptonic, I found a way to grab the value contained in the tool bit "rotation" attribute. This code works, though there is likely a more elegant way of accomplishing the same result.

To the tapping.py generator I added the following items:

import FreeCAD

 # Determine if tapping tool is RH or LH via roation attribute. 
    handed = FreeCAD.ActiveDocument.Tapping.ToolController.Tool.Rotation
    if handed == "Left Hand":
        cmd = "G74"
    else:
        cmd = "G84"

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [739570110](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/739570110) was triggered at [068df48](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/068df48acfbac49d383f9d3b91a6d4715b95387c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
Nope.  Don't want to do this.  The generator shouldn't know anything about the upstream tools/controllers.  Have the operation, query the TC/tool and get the handedness and pass it into the generator as an argument.

luvtofish:
I thought that may be the case. Okay, i’ll keep testing ideas. Thanks.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [739590263](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/739590263) was triggered at [9b69e52](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/9b69e52c0ebe35787e6f2ad25af40137442f6133). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
> I thought that may be the case. Okay, i’ll keep testing ideas. Thanks.
This should be really easy.  You have a reference to the TC in the operation.  I'd make an optional argument to the generator like righthand=True.  This way, if the argument isn't passed, it's assumed to be 'True'.

Then in the operation, I'd determine the value like this:
`
rh = getattr(obj.Tool, "Rotation", "Right Hand") == "Right Hand"
`
This attempts to get the attribute from the tool.  If it isn't present, it's assumed to be "Right Hand".  Then it's compared to get a boolean.  This way,  rh should always be a boolean and should only be False if the tool contains the attribute and the value is  not "Right Hand"

luvtofish:
> rh = getattr(obj.Tool, "Rotation", "Right Hand") == "Right Hand"

You are some kind of special 👍
I was struggling with getting a handle on the tool object within the tapping.py generator. The answer all along was to grab it from the tapping.py operation and pass it along through the generate call.  I've coded this and and compiling now to for testing. Thanks you sir!

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [739657170](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/739657170) was triggered at [00633b8](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/00633b8fcffdbab22519df1c1e9730c21e841d3c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luvtofish:
I needed to add in obj.ToolController.Tool to the getattr function -- it was not finding "obj.Tool". This is working now, updates have been pushed up. @sliptonic thank you for the guidance.  I still need to perform some cleanup before it's ready for review. 

Also -- the addition of the "G74" code presents a problem with the newly merged dynapath_4060_post. Since the Dynapath control doesn't use G74 (G84 only for either direction), I'll need to create a new PR for that post processor to add a simple change so that it only uses G84.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [739667425](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/739667425) was triggered at [3e823db](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/3e823db990c8843a3d052a99f481013e2ec7dd5d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
It's looking pretty good.  Do your cleanup and remove the draft status from the PR so it starts running tests.

Also, pick a better name for the handedness boolean.   Maybe isRightHanded or something.  

run Black() to format everything.

luvtofish:
Will do. Was wondering if I should add a test condition for left hand tap
with wrong spindle rotation or the inverse?

sliptonic:
> Will do. Was wondering if I should add a test condition for left hand tap with wrong spindle rotation or the inverse?


Yep.  Good idea

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [741117471](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/741117471) was triggered at [b5664ef](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/b5664efd2482f977dcaf54c0075353c3a3cb6825). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [741152826](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/741152826) was triggered at [a3eefa7](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/a3eefa7ede924b940c35ef017f9e989f03d7f2ff). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
Note: @sliptonic, upon merge of this PR, the following ticket #8018 and #5714 should be closed. TIA

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [741176607](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/741176607) was triggered at [c51ad66](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/c51ad6612d44bd5cdb9a5a40cdd3e1ada95137a6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [743548182](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/743548182) was triggered at [6384eff](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/6384eff2b1c7347f5739880aa464f6429ff8f84b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [743572101](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/743572101) was triggered at [5d28f9a](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/5d28f9add52f1afc4c8df71861950476ca418f86). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luvtofish:
> The current iteration of the generator takes both a spindle direction and a 'handedness' of the tool. What's the use case where they differ? When would you use a lefthanded tap but turn the spindle clockwise, for example? Or is this just for catching bad configurations?

It’s for catching a bad configuration and reporting a valueError in the Report View window. It doesn’t prevent the user from proceeding, just warns them of a mismatch.

sliptonic:
> It’s for catching a bad configuration and reporting a valueError in the Report View window. It doesn’t prevent the user from proceeding, just warns them of a mismatch.

Hm.  I like the idea of catching the misconfiguration but I'm not sure it belongs in the tapping generator to do it. There's at least three places it might make more sense.   

First would be the operation itself.  The op knows what TC is selected and it is where most of the op-specific logic lives. 

I suppose someone could make the argument that it belongs in the TC itself. After all, the TC checks if the tool should allow power.
https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Path/Path/Tool/Controller.py#L269-L276

The toolchange generator actually acts on that test.  expanding it to test for 'handedness' is possible but it seems like we're going down a slippery slope.  
https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Path/Path/Base/Generator/toolchange.py#L47-L63

I would suggest that it's the operation's job to check if the tool is appropriate to the task and throw an error if it isn't. It should never call the generator with an inappropriate set of arguments.

luvtofish:
Okay I’ll poke around some more and see what I can come up with.  Ty

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [744993223](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/744993223) was triggered at [7bbc1c5](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/7bbc1c5b646f0d5e7f9c9befb5b0e5037a21de00). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luvtofish:
I would argue it belongs with the tool controller operation. The logic fits here since this is where the spindle direction is being set based on the value in SpindleDir. 

The timing of the error occurs while making changes to the tool controller. This context is perfect for corrective action by the user. 

See latest update. Thanks

luvtofish:
This is all well and good however, the need to understand the handedness of
the tool is a critical detail needed to formulate the correct tapping cycle
(G74/G84). I assume this places a HOLD on this PR until the other is fully
developed and threshed out?

sliptonic:
Not at all.  In fact I'd like to push forward with this.  The tapping generator should take an argument for either handedness or spindle direction (not both) and generate the correct tapping cycle.  
The op should get that from the TC spindle direction.  The op could (maybe should) check that the selected tool is, in fact, a tap but shouldn't worry about whether the selected direction matches the tool.  That part is the abstracted stuff which we can push off to a later PR.

luvtofish:
Okay I’m tracking now.  I’ll remove the mismatch Catch update provided in
the last push since this needs to be part of a larger discussion. We have
the handedness being passed to the generator now and is working for setting
the correct gcode.

I haven’t added anything to test for a tap. I did think about this and
wondered how quickly it would break due to the creation of a user defined
tapping tool.

sliptonic:
We do similar kinds of tests for other tools.  I think v-carve checks for a v-bit. 
You could do it in a general way.  Since the operation and generator need attributes in the tapping tool, you could insist that the tool be of that type.   Alternatively, you could check for specific attributes of the tool like thread pitch.  That would be more flexible for user-defined tools. But what's the real use case for that?  Why would a user need to create a new tapping tool vs using the existing one?  

Don't overthink it.  It's just code. We're smart guys. We can fix it if it breaks.

luvtofish:
I was going down a similar thought pattern. I was thinking an attribute could be added to the tap shape which could then be checked against to confirm tool type. The only unique item currently with the tap shape is “Rotation” which is not really unique to tapping as it could apply for any tool shape.  A “pitch” attribute would certainly work but of course would only be informational since we don’t currently compute the feed rate for tapping. 

I’ll add a “pitch” attribute and test if you agree this is a good approach. Thanks

sliptonic:
It should be impossible to create a tapping operation with a drillbit TC or an endmill TC.  The tapping operation should check to see if you're using a tap tool.   That alone is probably enough for now.

I'm not an expert on tapping but it seems like there are exactly two criteria that the user cares about.  Two things that make a tap a 'tap';  The diameter and the thread pitch.  Yes, there are others like shank, flutes, etc. but these are the two critical ones.

**Thread Pitch**
The thread pitch should absolutely be included.  This attribute needs to be defined in freecad units and represented in user units so that the display follows the unit schema.  Take a look at how the other attributes work.  
I can imagine that In the future,  someone might want to improve the tap toolshape so that it renders accurate 'threads' in the solid model.  They would need the thread pitch to do it.

Moreover,  the thread pitch can be compared to the feed rate and spindle speed for a sanity check on rigid tapping. 

**Diameter**
I guess this is the major diameter?  1/4"  for a 1/4x20 tap or a 6mm for an M6x1.  This should be (and is) the diameter attribute in the sketch.

Looking forward (No need to act on any of this.  Let's get a first cut of the tapping operation done and merged.  But this is stuff to  think about):

The thing we don't know and can't predict is how the users will prefer to model tapped holes.  
Most will just add a cylindrical hole feature to the solid.  The tapping operation should work with this.  However, What if the selected hole is too small?  What if it's too big?  What if the model contains hundreds of holes of different sizes?  It seems like the tapping operation might need to get smarter about auto-selecting its targets.

On the other hand, some users will use the "hole feature" in part design.  This will let them specify the thread pitch, counterbore, diameter, profile, etc etc etc.  It will render accurate threads if the user wants to see them. 

The first iteration of the tapping op won't don't do anything with this information but it's an area for future improvement.  If there's more information available, the tapping operation can do more with it.  It could, for example, automatically select any M6 holes holes (and only threaded holes) if you select an M6 tap.

luvtofish:
For empirical threads, threads per inch (TPI) is a key detail. One does not
consider thread pitch when selecting an empirical tap, it’s diameter and
TPI (eg. 1/4-20). Pitch can be easily determined however by dividing TPI by
1. A 1/4-20 tap has a pitch of .050”. That is one revolution of the tap
moves up or down .050”.

Metric threads do not use TPI, only pitch. An M6-1 tap refers to diameter
and the pitch (1). In this case one revolution of the tap moves up or down
1mm.

I can see the need to add both pitch and TPI attributes. Pitch for metric
and TPI for empirical.

sliptonic:
> I can see the need to add both pitch and TPI attributes. Pitch for metric and TPI for empirical.

I don't see two different things.   They are expressed differently but are the same thing;  The distance moved down for one rotation. Let's call that 'pitch' 

For user-friendliness, you might need to display it differently in the task panel.  If the user selected preference is imperial, you'll show the pitch as TPI.  If it's metric, you'll show it as pitch.

luvtofish:
@sliptonic, please have a look at the latest and provide your thoughts. Thanks

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [764462028](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/764462028) was triggered at [9fc1e13](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/9fc1e1394fccba836a8165b0a80963ed745ac7e0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [764518286](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/764518286) was triggered at [09dd98a](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/09dd98aa3ae9e01a890a4e6dc3cd08a3e869b93b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069">pipeline status</a> for feature branch [PR_8069](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8069). Pipeline [764521961](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/764521961) was triggered at [de0043d](https://github.com/FreeCAD/FreeCAD/pull/8069/commits/de0043d14aa3e4457d4cf126e9280a15d3b220f4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
This feature needs to be rebased after the Path/CAM rename.

maxwxyz:
@luvtofish will you rebase this PR?

luvtofish:
> This feature needs to be rebased after the Path/CAM rename.

Hi @sliptonic, how do I initiate the rebase?

sliptonic:
> > This feature needs to be rebased after the Path/CAM rename.
> 
> Hi @sliptonic, how do I initiate the rebase?

Best practice is to check out your branch locally.  Rebase on master.  This depends on how you have your remotes set up.  I usually set mine up so 'origin' is my fork on github and 'upstream' is github.com/freecad/freecad

`git rebase upstream\master`

Hopefully there won't be any merge conflicts.  If there are, you'll need to resolve them.

Once the rebase completes normally.  Push your branch back to your github fork with a --force

The PR will automatically update and you should see the CI tests start running again.

sliptonic:
Oh, I see there are merge conflicts.  This appears to be mostly related to the PATH->CAM renaming

luvtofish:
> Oh, I see there are merge conflicts. This appears to be mostly related to the PATH->CAM renaming
The first conflict I'm seeing is a change from pathtest. ---> test. Not yet sure what the resolution is for this...

sliptonic:
Since this is mostly new functionality, it's just a matter of getting the files into the right place.  The small changes to cmake files and selection are straightforward.

Let me know if you can't get this done and I can help.  No hurry since we are in feature freeze for 1.0 and this is too late to make it.

luvtofish:
I’ve started the process of applying the updates to a fresh pull of the FreeCAD source. The new updates for CAM do not seem to egregious. I’ll ping with any questions I have. Thank you!
________________________________

luvtofish:
Okay -- I believe the PR has now been successfully rebased to the latest Main stream. The major change that needed work was the name change from Path to CAM and the subsequent changes to folder names this entailed. I've tested the Tapping functionality provided by this PR and it seems to work well as expectedl. I'll go ahead and mark this PR ready for review so we can get more eyes on it. I can provide screenshots and evidence of its function by request.  -- Dan H

sliptonic:
Since this PR will sit until after 1.0 release, it will need periodic rebasing.  I encourage you to use rebase instead of merging the main into it.  It will keep the history cleaner.

luvtofish:
Will do, thx. BTW, do have a rough date on when 1.0 will release?
________________________________

sliptonic:
> Will do, thx. BTW, do have a rough date on when 1.0 will release?

We're in feature-freeze now.  We froze UI changes last week.
The tracker is showing 128 issues tagged for the 1.0 release: https://github.com/FreeCAD/FreeCAD/issues?q=is%3Aopen+is%3Aissue+milestone%3A1.0

Not all of those will be resolved but we won't release with any that are labeled 'blocker' still open.

luvtofish:
Thanks.
________________________________

maxwxyz:
@luvtofish could you please resolve the conflicts? TIA

maxwxyz:
Looked at this at the merge meeting but it still has conflicts. Could you please resolve them?

luvtofish:
> Looked at this at the merge meeting but it still has conflicts. Could you please resolve them?

Working on it now. This PR has been awaiting merge for many moons.  After this conflict fix, can we go ahead and merge before additional conflicts occur? TIA

maxwxyz:
hopefully. There are a lot of PRs in the queue due to feature freeze for 1.0

PaddleStroke:
@sliptonic is this good to go?

PR #8077, opened 2022-12-23
add fmt library

berniev:

---

[github](https://github.com/fmtlib/fmt)

fmt is a fantastic library for formatting. It makes it possible to simplify much code.

fmt is being incorporated in STL and while it is supposed to be in C++20, is not generally available.

As it is not known how long before STL catches up, it is worth using fmt in the meantime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ if(NOT FREECAD_LIBPACK_USE OR FREECAD_LIBPACK_CHECKFILE_CLBUNDLER)
     SetupXercesC()
     find_package(ZLIB REQUIRED)
     find_package(PyCXX REQUIRED)
+    find_package(fmt REQUIRED)
     SetupOpenCasCade()
     if(BUILD_GUI)
         # Do this before the check for SMESH because it depends on vtk
--- a/src/App/CMakeLists.txt
+++ b/src/App/CMakeLists.txt
@@ -291,7 +291,7 @@ endif(FREECAD_USE_PCH)
 
 add_library(FreeCADApp SHARED ${FreeCADApp_SRCS})
 
-target_link_libraries(FreeCADApp ${FreeCADApp_LIBS})
+target_link_libraries(FreeCADApp fmt::fmt ${FreeCADApp_LIBS})
 
 add_dependencies(FreeCADApp fc_version)
 
--- a/src/Base/CMakeLists.txt
+++ b/src/Base/CMakeLists.txt
@@ -387,7 +387,7 @@ endif(FREECAD_USE_PCH)
 
 add_library(FreeCADBase SHARED ${FreeCADBase_SRCS})
 
-target_link_libraries(FreeCADBase ${FreeCADBase_LIBS})
+target_link_libraries(FreeCADBase fmt::fmt ${FreeCADBase_LIBS})
 
 SET_BIN_DIR(FreeCADBase FreeCADBase)
 
--- a/src/Gui/CMakeLists.txt
+++ b/src/Gui/CMakeLists.txt
@@ -1260,7 +1260,7 @@ if (EIGEN3_NO_DEPRECATED_COPY)
 endif ()
 
 add_library(FreeCADGui SHARED ${FreeCADGui_SRCS})
-target_link_libraries(FreeCADGui ${FreeCADGui_LIBS})
+target_link_libraries(FreeCADGui fmt::fmt ${FreeCADGui_LIBS})
 
 SET_BIN_DIR(FreeCADGui FreeCADGui)
 


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8077">pipeline status</a> for feature branch [PR_8077](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8077). Pipeline [730489834](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/730489834) was triggered at [d520470](https://github.com/FreeCAD/FreeCAD/pull/8077/commits/d520470148081459543ef9056a4a16f88b655742). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
why include the entire source code in our repo? at most it should be a submodule IMO but even then why would we need the source code here?

berniev:
@adrianinsaval Because I don't know what I'm doing. The instructions I saw said just copy the whole thing in so I did. But it didn't work so I pulled PR back to draft. Cmake is definitely not my expertise, but fmt is really worthwhile pursuing.

It does come with header files and I guess they are really all that is needed, if someone can help with the how that would be fantastic.

berniev:
@adrianinsaval 

> why include the entire source code in our repo? at most it should be a submodule IMO but even then why would we need the source code here?

I'm back to looking at this but am confused by your post. You were quite adamant.

There's a precedent in libE57Format. Google tests also recomment building into project build. 

Care to help a learner understand? (Ignore that my original version didn't work - that's why it was immediately draft).

adrianinsaval:
> You were quite adamant.

sorry if I gave you the wrong impression but I am not a maintainer so my opinion is not authoritative, it is just that, my opinion. And I am in no way an expert either, just an amateur, that's also why I formulated these as a question not a statement, I'm open to being corrected.

> There's a precedent in libE57Format

yes and also smesh and other things, but personally I never liked this, we then also have to manage this third party code in our repo, why not just use available packages for these? Or if we really need the source code here and to compile ourselves, why not add it as a git submodule? This seems like a much more reasonable way to manage 3rd party repos in our codebase IMO.

berniev:
No worries. You have given the clue (use existing package), which is great!

libfmt does seem widely available. I see it on ubuntu, and in homebrew. So I'll go that way. It works great. Will force push new attempt shortly.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8077">pipeline status</a> for feature branch [PR_8077](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8077). Pipeline [736559984](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/736559984) was triggered at [0426375](https://github.com/FreeCAD/FreeCAD/pull/8077/commits/0426375841c481b0c8f48b7cdb123a6bbc1c9670). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berniev:
@adrianinsaval Turns out (werner) Windows is an issue until fmt is added to FC LibPack. So 3rd attempt will be just header files in 3rdParty.

luzpaz:
@berniev so the hold-up is adding `fmt` to libpack? That's donovaly's territory. Should we ping him ?

abdullahtahiriyo:
What a small world, I am also adding libfmt as part of the console refactor (to handle printf statements necessary for console). I am also confused by the documentation. CMake is also not my cup of tea.

This is what I did:
https://github.com/FreeCAD/FreeCAD/pull/8350

Basically:
1. Add to CMakeList ```find_package(fmt)```
2. For header only code, then the macro needs to be defined:
```
#define FMT_HEADER_ONLY
#include <fmt/printf.h>
```
Looking at this PR, I am mising the REQUIRED in the ```find_package```

As far as I understood, there are two ways of adding libfmt, either (a) as a header only library (which is what I did), or (b) pre-compiled. 

In the header only version, I do not think it is necessary to use "target_link_libraries", as there is nothing to link against (it is header only, it gets compiled together with the code).

In the pre-compiled version, there is indeed a shared library to link together with the code. I read that this reduces the compilation time quite a lot (as opposed to using the header only version).

In a project the size of FreeCAD, where libfmt has the potential to be used in several places, it would probably make sense to use the precompiled version (at least for Windows, which currently uses pre-compiled headers). 

On the issue of the libpack, if we are going for the header-only, the libpack (or outside the libpack the actual code) needs to be available for the compiler to compile it. If we are going for a pre-compiled version, then I guess the pre-compiled version needs to be added to the libpack.

I am unsure if we will end up with several copies of the object code of libfmt if we use the header only version at different modules.

@wwmayer 

Werner, as you also advocate to add the fmt library, would you like to weigh in. Do you have any expectation or additional requirement or input?

berniev:
Concensus so far is that pkg is the way to go, but more difficult on Windows if not in FC Libpack. Is vcpkg not able to provide libfmt and/or gtest? (I am on. Mac).

abdullahtahiriyo:
libfmt-dev in Ubuntu includes the source code and in addition an static library ```libfmt.a```. This caused some confusion to me. There is no shared library file (as a package).

If the packages are similar in other distributions/windows/macos, then the installed package should also have such a library. Then doing the header only or linking to the static library should be the same (the only difference being compilation speed). Am I wrong?

However, the docs say it is possible to build it as a shared library:
https://fmt.dev/latest/usage.html

This is what made me doubt whether we should create a shared library and distribute it together (in those distributions/os where no such shared library exists).

berniev:
[ubuntu packages](https://packages.ubuntu.com/search?keywords=libfmt-dev)
[SO on installing](https://askubuntu.com/questions/1205765/c-fmt-library-installation-is-not-working)

abdullahtahiriyo:
[SO on header-only with reference to vcpkg](https://stackoverflow.com/questions/66944554/how-to-use-fmt-library-in-the-header-only-mode)

[SO on why it is not header-only](https://stackoverflow.com/questions/67993134/how-come-the-fmt-library-is-not-header-only)

berniev:
Here's my understanding ...

fmt comes in three different forms:
- source. Can be built to linkable object files
- package: includes compiled, linkable object files built for certain target
- header-only header files. Available from source or package

The headers can be used without linking. Just copy them to the project. #include directly to format.h. But there be dragons I don't fully understand. I can verify you do need FMT_HEADER_ONLY to silence weird errors. Werner pointed out that including format in a header file can cause lots of linking as headers load nested dependencies.

Object files need linking to the project and can be:
- created from source
- provided pre-built by a package manager(brew, vcpkg, etc)
- in the case of FC, provided by so-called libpack, pre-built for Windows (atm not fmt).

Package also includes config files for cmake's find_package.

My first PR attempt was source, rejected as don't want fmt source in FC.
My second PR attempt was package, rejected because fmt notin FC libpack.

I have used both methods, but technically it seems linking to a package is the way to go. The missing link is Windows support. Hence my question whether vcpkg is a suitable alternative source for Windows fmt package. No response here or on forum.

abdullahtahiriyo:
@berniev 

Basically yes. 

I do have quite some interest in adding this library too. Let's see if we can get some answers to be able to move forward.

Some clarifications/summary:
- Including the source in FC (either for header-only or to generate a shared or static library) should not be done.
- The package, at least the Ubuntu version, includes both the source and a static library (not a shared library, but a statically linked one, .a not .so). This means that, the package can be used directly in Freecad either for header-only (Option A) or for statically linked code (Option B). The package could also be used indirectly to generated a shared library as in the link above (Option C). However, no shared library is provided as a package, at least, in Ubuntu.

> The headers can be used without linking. Just copy them to the project. #include directly to format.h. But there be dragons I don't fully understand. I can verify you do need FMT_HEADER_ONLY to silence weird errors. Werner pointed out that including format in a header file can cause lots of linking as headers load nested dependencies.

The use of FMT_HEADER_ONLY is [indicated](https://github.com/fmtlib/fmt/issues/596) by the creator of the library. As linked above, according to the creator, the only difference between statically linked and header-only is compilation time (which is always precious). 

However, in any of these two options, it is indeed a concern for me that several separate copies of object code (compiled from source or statically linked library) may be ultimately be present in the binary. This prompted my observation and why I mentioned the option to compile the source to make a shared library (Option C).

Option C would require in Linux (and macos I think) to compile libfmt as a shared library and install it, which may also have strong packaging implications. We could invite Kurt to the discussion it we would seriously consider Option C (I do not see why bother him, if for other reasons we are not going in this direction).

I am not emotionally attached to Option A. For my specific case it was the lowest modification option to introduce libfmt at a single point to fix the specific issue with Console I was addressing, while waiting to see which was the agreed best possible way to use it globally. Now we have arrived to this point, so Options B and C are left.

Option B is straight forward (as in this PR).

This said, I can live happily ever after with any option, but, of course, I prefer the one that better suits our needs. This we need to resolve first. We need Werner's guidance:

@wwmayer Could you please weigh in on which Option you favour?

In practical terms:
- No matter the option in Ubuntu (and I think macos too), the package is the way to go.
- I am on your very same footing regarding Windows. I see the package can be obtained from vcpkg. However, I also see that in Windows developers use the Libpack. I do not have Windows. We need additional information of how our community is approaching this kind of issues (it cannot be it is the first time). Let's try to get some guidance here:

@chennes @donovaly 

Could you please weigh in on which the Windows requirements are regarding this library inclusion? Do we need every single library into the libpack or we can live with a mix of libpack and vcpkg?

Hopefully, with those two questions answered, we can move forward. You can made the changes to your PR, if any necessary, get it merged and I can rebase on top and mark mine as non-draft for merge too...

berniev:
>  it is indeed a concern for me that several separate copies of object code (compiled from source or statically linked library) may be ultimately be present in the binary

You are suggesting that multiple copies of THE SAME object code will get into the binary? Is that right?

abdullahtahiriyo:
> You are suggesting that multiple copies of THE SAME object code will get into the binary? Is that right?

No. This won't happen. The linker will add more or less from the library, but there will be no duplicates.

Option B is fine with me. It is the most convenient option as there are not shared libraries available in most systems.

We just need to get clarification of whether Windows users can live with getting the package via vcpkg or need to be made part of the libpack. Hopefully we will get an answer from Windows users.

chennes:
Does `fetch_content` not work in this case? e.g.
```
FetchContent_Declare(fmt
       GIT_REPOSITORY https://github.com/fmtlib/fmt
       GIT_TAG 9.1.0
)
FetchContent_MakeAvailable(fmt)
```

(or don't specify the specific version, use `master`, etc.

berniev:
@chennes I know the idea works - I used it or similar for gtest at one stage, but it requires cmake 3.11 which we achieved only  just recently. You are probably the best person to actually try your idea!

If only someone from Windows land would step in! IF vcpg works then we have packages for all FC supported platforms and we can use simple find_package etc.

waebbl:
> * The package, at least the Ubuntu version, includes both the source and a static library (not a shared library, but a statically linked one, .a not .so). This means that, the package can be used directly in Freecad either for header-only (Option A) or for statically linked code (Option B). The package could also be used indirectly to generated a shared library as in the link above (Option C). However, no shared library is provided as a package, at least, in Ubuntu.

The Ubuntu package also includes the shared library. You only installed the -dev package, which, AFAIK never includes shared libraries. See the list of files at https://packages.ubuntu.com/lunar/amd64/libfmt9/filelist

[Repology](https://repology.org/project/fmt/versions) shows, that almost every linux distribution has a package for fmt, so at least for the linux platform using a system provided package should be preferred to keep distributions happy.

wwmayer:
> The Ubuntu package also includes the shared library.

They do this since Jammy (22.04). All older versions only provided a static version.

chennes:
I'm going to close this as handled by #8508 - I don't know if that PR was the "final" solution (if there is such a thing), but we've got fmt into master now. So fmt away.

PR #8103, opened 2022-12-30
[Sketcher] Add necessary include

AjinkyaDahale:
See https://forum.freecadweb.org/viewtopic.php?f=10&t=74890.


AjinkyaDahale:
Closing in favour of #8106.

Issue #8064, opened 2022-12-20
[Problem] CAM: Missing letter when V-carving a string


mhindi2:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.20 (Release)

### Full version info

```shell
OS: Fedora Linux 35 (Workstation Edition) (KDE/plasmax11)
Word size of FreeCAD: 64-bit
Version: 0.20.1.Unknown
Build type: Unknown
Python 3.10.6, Qt 5.15.2, Coin 4.0.0, Vtk 9.0.3, OCC 7.5.0
Locale: English/United States (en_US)
Installed mods: 
  * Pyramids-and-Polyhedrons
  * fasteners 0.4.25
  * GDML 2.0.0
  * Manipulator 1.4.9
  * BIM 2021.12.0
  * Assembly4 0.12.4
```

### Subproject(s) affected?

Path

### Issue description

![kitchen](https://user-images.githubusercontent.com/95139464/208580671-76291802-9480-419d-977b-31b71dd5fd37.jpg)
One of the letters in a vcarve of a Shape string is not rendering at all (the "e"  in Kitchen, please see attached image.)
The font is urw-base35/Z003-MediumItalic.otf; the size is 10.00 mm high. I've also observed a "B" missing in a different font, but not this one.

### Anything else?

![kitchen](https://user-images.githubusercontent.com/95139464/208581280-efdc31d4-ab84-44a5-a6a7-3cb88c5073c7.jpg)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
@mhindi2 can you please zip the .FCStd file and attach it to the ticket. Can you x-post this to the Path forum.freecadweb.org to make sure this isn't a 3rd party dependency or user-error ?

mhindi2:
[missing_letter.zip](https://github.com/FreeCAD/FreeCAD/files/10269803/missing_letter.zip)
Here it is. The tool is a 60 degree V-bit, with a 3 mm diameter.

mhindi2:
Discussions about this issue in the Path forum:
https://forum.freecadweb.org/viewtopic.php?f=15&t=74600&p=648804#p648804

mlampert:
First off, when I tested this last night vcarve on the latest master the letter `e` is carved correctly, however the letter `h` is missing. As already indicated by Gene in the forum discussion, whatever happens here is data driven.

[...]

PR #8111, opened 2023-01-01
[hold] [Path] bugfix/8064-vcarve-skipping-certain-letters

mlampert:
- Rounding Voronoi input values after scaling for numeric stability

This seems to fix issue #8064 - however - I have currently no explanation as to why and how. Using an integer cast instead of `round` does _not_ fix the issue. It would be nice to understand what is going on and why this should fix it - hence the hold.


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8111">pipeline status</a> for feature branch [PR_8111](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8111). Pipeline [735721969](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/735721969) was triggered at [c8dfc0a](https://github.com/FreeCAD/FreeCAD/pull/8111/commits/c8dfc0a84e1d6fb98fba6d918c88548de3e3a2b7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

br4243:
Just following up on this from the #8064 discussion.  The problematic letters occur when the path discretizer aims for the terminal vertex (presumably it could be any vertex?) and misses slightly.  insert_many_wires() closes the letter's outline polygon with a little short segment.

If Vcarve round()s like your fix, that slight miss vertex and the target vertex end up identical so the short segment added by insert_many_wires() goes to zero length, which doesn't hurt anything as long as orthognalProjection() checks for zero length segments.

Boost's Voronoi sweep line algorithm truncates the coordinates it gets to integer values, so there's a reasonable chance the two vertices that were close in FC's floating point coordinate system end up separated by one or two Manhattan distance abscissa values in Boost's coordinate system.  There's a further chance that the separation induced by that truncation makes the letter's outline polygon in Boost's coordinate space contain self-crossings, resulting in garbage out of VoronoiBuilder.

sliptonic:
Ping @mlampert  
Can we get an update here?  If this is pending some future work, can we convert it to draft?

mlampert:
Closing this PR since it didn't actually fix the issue.

Issue #6878, opened 2022-05-12
[Part] `BOPTools/ShapeMerge.py` code appears to indicate it's inefficient


luzpaz:
https://github.com/FreeCAD/FreeCAD/blob/5d49bf78de785a536f941f1a6d06d432582a95d3/src/Mod/Part/BOPTools/ShapeMerge.py#L100

AjinkyaDahale:
We can replace
https://github.com/FreeCAD/FreeCAD/blob/5d49bf78de785a536f941f1a6d06d432582a95d3/src/Mod/Part/BOPTools/ShapeMerge.py#L99-L102
with
```
l_groups = len(groups)
groups = [groups[i_group] for i_group in connected_to if i_group < l_groups]
```
This removes the unnecessary search in `connected_to`

luzpaz:
bump

AjinkyaDahale:
I can make the PR, but I need someone to back up what's happening.

luzpaz:
@DeepSOIC this is code you added. Could you lend a moment of your time to weigh-in? :pray:

DeepSOIC:
I don't quite remember why i thought it's inefficient. The odds of connected_to being long are pretty low - I can't think up an example of how it can be longer than 2 for manifold shapes (except compsolids, maybe). So, i think it is safe to remove this comment. On the other hand, i think it may be possible to improve the performance of the whole algo significantly... i'm actually having a hard time even estimating the O() complexity of it, because these aren't just plain random lists - topology is involved.

AjinkyaDahale:
@DeepSOIC thanks for the input. I had still made this: https://github.com/AjinkyaDahale/FreeCAD/tree/part-fix-6878. Probably an overkill for the list sizes you mentioned. We can discuss further what we can do instead.

luzpaz:
@AjinkyaDahale what's the fate of this ticket ?

AjinkyaDahale:
> @AjinkyaDahale what's the fate of this ticket ?

Let's close it. There may be need for a larger scale overhaul but this one snippet is already changed and merged.

PR #8113, opened 2023-01-01
[Part] Optimize snippet of code in `ShapeMerge.py`

AjinkyaDahale:
Some notes:

Earlier, when adding to the list `connected_to`. the numbers are added in order from `range(len(groups)`, so we are already in order and every `i_group` is unique.

As per https://github.com/FreeCAD/FreeCAD/issues/6878#issuecomment-1368330972 this might not be necessary. However, @DeepSOIC also mentions the entire method could be optimized so I can work on that here.

### DIFF
--- a/src/Mod/Part/BOPTools/ShapeMerge.py
+++ b/src/Mod/Part/BOPTools/ShapeMerge.py
@@ -77,6 +77,7 @@ def splitIntoGroupsBySharing(list_of_shapes, element_extractor, split_connection
         shape_elements.difference_update(split_connections)
         #search if shape is connected to any groups
         connected_to = []
+        not_in_connected_to = []
         for iGroup in range(len(groups)):
             connected = False
             for element in shape_elements:
@@ -84,6 +85,10 @@ def splitIntoGroupsBySharing(list_of_shapes, element_extractor, split_connection
                     connected_to.append(iGroup)
                     connected = True
                     break
+            else:
+                # `break` not invoked, so `connected` is false
+                not_in_connected_to.append(iGroup)
+
         # test if we need to join groups
         if len(connected_to)>1:
             #shape bridges a gap between some groups. Join them into one.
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8113">pipeline status</a> for feature branch [PR_8113](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8113). Pipeline [735677308](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/735677308) was triggered at [1148d5e](https://github.com/FreeCAD/FreeCAD/pull/8113/commits/1148d5e200ff6cde0507747db49198b01d2e59a5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8113">pipeline status</a> for feature branch [PR_8113](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8113). Pipeline [735747207](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/735747207) was triggered at [1148d5e](https://github.com/FreeCAD/FreeCAD/pull/8113/commits/1148d5e200ff6cde0507747db49198b01d2e59a5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
Can mark this for `0.22` since I don't plan to look into it right now.

src/Mod/Part/BOPTools/ShapeMerge.py
```diff
@@ -77,13 +77,18 @@ def splitIntoGroupsBySharing(list_of_shapes, element_extractor, split_connection
...
+        not_in_connected_to = []
         for iGroup in range(len(groups)):
             connected = False
             for element in shape_elements:
                 if element in groups[iGroup][1]:
                     connected_to.append(iGroup)
                     connected = True
                     break
+            else:
+                if not connected:
```
0penBrain, line 89: This line seems useless as induced by the rest of the code.

AjinkyaDahale: Do you mean `else` part will always be triggered after `connected` is `True`?

0penBrain: I mean as it is a `for...else` loop, `else` will be executed only if `break` isn't called, so only if `connected` is `False`

AjinkyaDahale: You seem correct here. Sorry I mixed up the behavior of `for...else`. I don't seem to recall why this exact structure is here.

AjinkyaDahale: It wasn't. But it was only a redundant check so no harm.

PR #8120, opened 2023-01-02
Notification Area - Non-Intrusive notifications

abdullahtahiriyo2:
DRAFT PR! DO NOT MERGE! Not intended for code review!

The intention of this PR is to be a kind of proof of concept and provide a basis for discussion of whether we need something like this or not.

Problems:
1. Users have complained in the past that they are not notified of issues that may impact the ability to open with an older version of FreeCAD a file saved with a newer version of FreeCAD (Forward compability issues).
2. While running FreeCAD, many users prefer not to have the report window open (as it takes quite some space), but then they may miss notifications about errors and warnings that may be useful.
3. While running FreeCAD, many notifications of errors where the user can do nothing, but press ok, are modal (blocking) and interrupt the workflow.
4. In many errors/warnings it is not clear where the error originates. This is specially problematic when opening a file.

Solution:
(a) FreeCAD has a error/warning/message mechanism in place (Console based).
(b) FreeCAD has a user Message signaling mechanism which can be used to require confirmation from a user for (1).

The solution here seeks to implement a Notification Area in the statusbar at Application level, so that it can gather notifications from different documents [to solve (4)]. It extends (a) and (b), so that both can provide a "notifier" field. This notifier field can be the fullname of a Document or DocumentObject, or any string usable for identifying the notifier within the Application [to solve (4)]. Both (a) and (b) feed the Notification Area. The notification Area is a single button in the tool bar which makes accessible all notifications upon clicking on it [in order to address the space issue in (2)]. New notifications (error/warning/messages) are briefly shown, in a look similar to a tooltip, via a non-intrusive non-modal auto-closing widget over the Notification Area [addressing (2) and (3)]. The idea is to also integrate user messages into this system and do away with blocking modal dialogs (the Sketcher is full of them) [addressing (3)]. Well, "do away" may mean "provide the user with a preference to choose between non-intrusive non-blocking and intrusive blocking".

The solution described is an early basic implementation for discussion.

[NotificationArea.webm](https://user-images.githubusercontent.com/121720825/210188173-d328f9bb-e73c-425b-9f87-68424bd29ab2.webm)

Tooltip alike non-intrusive auto-closing notification:
image

Notifications widget:
image

### DIFF
--- a/src/Base/Console.h
+++ b/src/Base/Console.h
@@ -489,7 +489,7 @@ public:
 
     /** Used to send a Log message at the given level.
      */
-    virtual void SendLog(const std::string& msg, LogStyle level) = 0;
+    virtual void SendLog(const std::string& notifiername, const std::string& msg, LogStyle level) = 0;
 
     virtual const char *Name(){return nullptr;}
     bool bErr,bMsg,bLog,bWrn;
@@ -515,24 +515,32 @@ public:
  */
 class BaseExport ConsoleSingleton
 {
-
 public:
     static const unsigned int BufferSize = 4024;
     // exported functions goes here +++++++++++++++++++++++++++++++++++++++
     /// Prints a Message
-    virtual void Message ( const char * pMsg, ... );
+    void Message ( const char * pMsg, ... );
     /// Prints a warning Message
-    virtual void Warning ( const char * pMsg, ... );
+    void Warning ( const char * pMsg, ... );
     /// Prints a error Message
-    virtual void Error   ( const char * pMsg, ... );
+    void Error   ( const char * pMsg, ... );
     /// Prints a log Message
-    virtual void Log     ( const char * pMsg, ... );
+    void Log     ( const char * pMsg, ... );
+    
+    /// Prints a Message with source indication
+    void MessageS ( const std::string &, const char * pMsg, ... );
+    /// Prints a warning Message with source indication
+    void WarningS ( const std::string &, const char * pMsg, ... );
+    /// Prints a error Message with source indication
+    void ErrorS   ( const std::string &, const char * pMsg, ... );
+    /// Prints a log Message with source indication
+    void LogS     ( const std::string &, const char * pMsg, ... );
 
     // observer processing
-    void NotifyMessage(const char *sMsg);
-    void NotifyWarning(const char *sMsg);
-    void NotifyError  (const char *sMsg);
-    void NotifyLog    (const char *sMsg);
+    void NotifyMessage(const char *sMsg, const std::string & notifiername = "");
+    void NotifyWarning(const char *sMsg, const std::string & notifiername = "");
+    void NotifyError  (const char *sMsg, const std::string & notifiername = "");
+    void NotifyLog    (const char *sMsg, const std::string & notifiername = "");
 
     /// Attaches an Observer to FCConsole
     void AttachObserver(ILogger *pcObserver);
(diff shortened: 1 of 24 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8120">pipeline status</a> for feature branch [PR_8120](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8120). Pipeline [735848086](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/735848086) was triggered at [3ef4c71](https://github.com/FreeCAD/FreeCAD/pull/8120/commits/3ef4c7110101b87334a65e2f5347fd65cc2dc15b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
This is amazing work! Auto-closing/dissolving messages are what I've been waiting for a [long time](https://forum.freecadweb.org/viewtopic.php?style=1&t=53543)!

chennes:
This looks excellent. +1 from me for sure.

src/Base/Console.h
```diff
@@ -489,7 +489,7 @@ class BaseExport ILogger
 
     /** Used to send a Log message at the given level.
      */
-    virtual void SendLog(const std::string& msg, LogStyle level) = 0;
+    virtual void SendLog(const std::string& notifiername, const std::string& msg, LogStyle level) = 0;
```
abdullahtahiriyo2, line 492: Notes:
ILogger interface is extended to include a new string "notifiername". This would ultimately enable Console messaging to offer notifier identification, which is specially useful to discern between Documents and Document Objects where an notification may originate, while still allowing other, client defined, identification where the object is neither a Document or a DocumentObject (such as sketch.cpp).  In part, this is done to separate the actual error messages and notifier (which could be treated differently by different observers, and may lead to reusing the same message for different observers despite how the handle notifier information) and, in part, to support multi-document projects.

This implies a change in the API. The majority of the rest of the changes in this commit are to deal with this change of API. If there is client code that links against FreeCAD and uses this interface, that client code would need to be adapted too. In those changes, I have not made a decision yet on how to best convey the notifier information.

src/Base/Console.h
```diff
@@ -515,24 +515,32 @@ class BaseExport ILogger
  */
 class BaseExport ConsoleSingleton
 {
-
 public:
     static const unsigned int BufferSize = 4024;
     // exported functions goes here +++++++++++++++++++++++++++++++++++++++
     /// Prints a Message
-    virtual void Message ( const char * pMsg, ... );
+    void Message ( const char * pMsg, ... );
```
abdullahtahiriyo2, line 522: Note:

Legacy Message/Warning/Error/Log functions stop being virtual. New MessageS/WarningS/ErrorS/LogS counterparts are created, which take an additional notifier parameter (also non-virtual). Both this are then handled via a virtual private interface (NVI pattern) for extension (see MessageV/WarningV/ErrorV/LogV).

The Console client user has then the option to emit regular legacy message (not identifying the notifier), or use the new versions identifying the notifier.

abdullahtahiriyo2:
Thanks.

It would be great if somebody could check how it runs on Windows, as the auto-closing notification has some windows specific code, and I have no Windows.

This file will trigger parabola migration, and partially redundant constraint warnings:
[test2.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/10332690/test2.FCStd.zip)

PR #8165, opened 2023-01-11
Part/Sketcher: fix #8143

### LINKED ISSUE #8143: [Bug] [Sketcher] Periodic BSpline calculation hangs

### Version

0.21 (Development)

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.0.31432 (Git)
Build type: Release
Branch: master
Hash: 63a2855751d638bf27ed969dc60654bda2b54baf
Python 3.10.8, Qt 5.15.4, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: French/France (fr_FR)
```

### Subproject(s) affected?

Sketcher

### Problem description

Hello,
I defining the degree of a Periodic BSpline with the shortcut by the D key, I enter the value 8 for example with 6 points and FreeCAD enters an infinite calculation.

I tested with a BSpline by control points and the result is correct. The degree displayed is 8.

### Anything else?

_No response_

### Code of Conduct

AjinkyaDahale:
The infinite loop already doesn't happen, but the generated B-spline is still ill-formed if the degree is set to too high.

### DIFF
--- a/src/Mod/Part/App/BSplineCurvePyImp.cpp
+++ b/src/Mod/Part/App/BSplineCurvePyImp.cpp
@@ -1226,7 +1226,7 @@ PyObject* BSplineCurvePy::buildFromPolesMultsKnots(PyObject *args, PyObject *key
                 if (knots != Py_None) { number_of_knots = PyObject_Length(knots); }
                 else { //guess number of knots
                     if (Base::asBoolean(periodic)) {
-                        if (number_of_poles < degree) {degree = number_of_poles+1;}
+                        if (number_of_poles < degree) {degree = number_of_poles;}
                         number_of_knots = number_of_poles+1;
                     }
                     else {
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerBSpline.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerBSpline.h
@@ -394,6 +394,8 @@ private:
 
             int currentgeoid = getHighestCurveIndex();
 
+            unsigned int maxDegree = ConstrMethod == 0 ? (BSplinePoles.size()-1) : (BSplinePoles.size());
+
             try {
                 //Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Add B-spline curve"));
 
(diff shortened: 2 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8165">pipeline status</a> for feature branch [PR_8165](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8165). Pipeline [744408436](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/744408436) was triggered at [98eca6a](https://github.com/FreeCAD/FreeCAD/pull/8165/commits/98eca6a3945665c33e1cad8217a5262086515f52). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/BSplineCurvePyImp.cpp
```diff
@@ -1226,7 +1226,7 @@ PyObject* BSplineCurvePy::buildFromPolesMultsKnots(PyObject *args, PyObject *key
                 if (knots != Py_None) { number_of_knots = PyObject_Length(knots); }
                 else { //guess number of knots
                     if (Base::asBoolean(periodic)) {
-                        if (number_of_poles < degree) {degree = number_of_poles+1;}
+                        if (number_of_poles < degree) {degree = number_of_poles;}
```
abdullahtahiriyo, line 1229: @AjinkyaDahale 
OCC does not set a constraint on degree for periodic B-Splines:
https://dev.opencascade.org/doc/refman/html/class_geom___b_spline_curve.html#a00d969dca532b7cd71594f805608fc7d

However, several algorithms fail with ```number_of_poles+1```, among them ```GCPnts_AbscissaPoint::Length```. They do not fail with ```degree = number_of_poles```.

I can only reach the conclusion that ```degree = number_of_poles + 1``` was an arbitrary limitation for a periodic B-Splines (possibly from non-periodic B-Splines). Given that algorithms fail, it could be treated as fixing a bug (although I would have welcome a more scientific rationale).

I do not foresee problems with existing projects, as what is serialised will be de-serialised to the same result.

AjinkyaDahale: In theory a periodic B-spline should be able to have arbitrary degree if you're ready to repeat the poles (and knots) indefinitely.

I don't understand what you mean by "(de-)serialized".

PR #5321, opened 2022-01-02
PD: Make helix property descriptions translatable and reuse them as task panel tooltips

jbaehr:
By reusing the property docs as tooltip we avoid maintaining the same information in two places. The property descriptions have been made translatable, too, to ensure a fully translatable UI.

If we agree on this in general, I can adapt the other PartDesign features/taskpanels as well. However, before putting more work into this, I want to hear your comments.

I followed the methodology @wwmayer and @yorikvanhavre have introduced long time ago, cf. [Issue 0002524](https://tracker.freecadweb.org/view.php?id=0002524). It is used at least in the Arch, Draft and Path WBs and I think (not only) the PartDesign WB can benefit from this work, too.

----------
Thank you for creating a pull request to contribute to FreeCAD! To ease integration, we ask you to conform to the following items. Pull requests which don't satisfy all the items below might be rejected. If you are in doubt with any of the items below, don't hesitate to ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10)!

- [x]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [ ]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [x]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [x]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [x]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [ ]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the [FreeCAD bug tracker issue number](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) in case a particular commit solves or is related to an existing issue on the tracker. Ex: `Draft: fix typos - fixes #0004805`

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.20 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=56135).

---

PR #8184, opened 2023-01-13
PD: Extend InvoluteGear to make the tooth length adjustable

jbaehr:
This PR exposes the addendum (tooth height above pitch circle), dedendum (tooth height from its root to the pitch circle) and root fillet radius as properties of PartDesign's Involute Gear profile. The values are all exposed as coefficients, i.e. normalized by the gear's module.

This PR addresses the following use cases:
- Adjust the tooth hight for internal gears to avoid interferences or non-involute flanks
- Create gears using others than the "full-deptsh system", e.g. with [stub teeth](https://khkgears.net/new/gear_knowledge/gear-nomenclature/stub-teeth.html)
- Use the profile for [involute splines](https://en.wikipedia.org/wiki/Spline_(mechanical)), i.e. splined shafts and hubs

image

When loading documents with gears created in earlier versions, the new properties are setup with values matching the previously hard-coded ones. This way the generated profile does not change. Unit tests for the backward compatibility are provided with fixtures generated by official mac build of FreeCAD-0.20.1.

For review, I suggest to look at the commits individually. They are carefully crafted to address a specific concern each and do not mix refactors/cleanups with functional changes. For this reason, please *do not squash* them at merge.


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8184">pipeline status</a> for feature branch [PR_8184](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8184). Pipeline [747073365](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/747073365) was triggered at [edea040](https://github.com/FreeCAD/FreeCAD/pull/8184/commits/edea040f82c168be0b440e2f79912b942be1446c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/fcgear/involute.py
```diff
@@ -157,7 +185,11 @@ def CreateInternalGear(w, m, Z, phi, split=True):
     if (Ra > Rb):         # start profile at top of fillet (if its greater)
         tipToPitchAngle -= genInvolutePolar(Rb, Ra)
     pitchToFilletAngle = genInvolutePolar(Rb, Rf) - baseToPitchAngle;
-    filletAngle = 1.414*clearance/Rf  # // to make fillet tangential to root
+    filletAngle = 1.414*(fRad/1.5)/Rf  # to make fillet tangential to root
+                                       # TODO: This and/or the Rf calculation doesn't seem quite
+                                       # correct. Keep it for compat, though. In the future, may
+                                       # use the special filletCoeff of 0.375 as marker to switch
+                                       # between "compat" or "correct/truely tangential"
```
luzpaz, line 192: pls fix typo truely ==> truly

src/Mod/PartDesign/InvoluteGearFeature.py
```diff
@@ -76,28 +75,53 @@ class _InvoluteGear:
...
-            involute.CreateExternalGear(w, obj.Modules.Value,obj.NumberOfTeeth, obj.PressureAngle.Value, obj.HighPrecision)
-        else:
-            involute.CreateInternalGear(w, obj.Modules.Value,obj.NumberOfTeeth, obj.PressureAngle.Value, obj.HighPrecision)
+        generator_func = involute.CreateExternalGear if obj.ExternalGear else involute.CreateInternalGear
+        generator_func(w, obj.Modules.Value, obj.NumberOfTeeth, obj.PressureAngle.Value,
+            split=obj.HighPrecision, addCoeff=obj.AddendumCoefficient, dedCoeff=obj.DedendumCoefficient,
+            filletCoeff=obj.RootFilletCoefficient)
         gearw = Part.Wire([o.toShape() for o in w.wire])
         obj.Shape = gearw
         obj.positionBySupport();
```
luzpaz, line 127: This is not related to your PR but do you mind adding a separate commit to remove this unnecessary semicolon?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8184">pipeline status</a> for feature branch [PR_8184](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8184). Pipeline [747309730](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/747309730) was triggered at [4e753e2](https://github.com/FreeCAD/FreeCAD/pull/8184/commits/4e753e2dcf43feaab96f646f4d70a8d0e27244f7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/CMakeLists.txt
```diff
@@ -55,6 +55,11 @@ set(PartDesign_TestScripts
     PartDesignTests/TestInvoluteGear.py
 )
 
+set(PartDesign_TestFixtures
+    PartDesignTests/Fixtures/InvoluteGear_v0-20-1.FCStd
```
donovaly, line 59: Can you please use a generic name. These files will stay also for FC 0.21 or 1.0

src/Mod/PartDesign/InvoluteGearFeature.ui
```diff
@@ -173,6 +173,66 @@
...
+     </property>
+     <property name="value">
+      <double>1.000000000000000</double>
+     </property>
+    </widget>
+   </item>
+   <item row="6" column="0">
+    <widget class="QLabel" name="label_2">
+     <property name="text">
+      <string>Dedendum Coefficient</string>
```
donovaly, line 199: Please also add tooltips. These will help especially newbies a lot.

donovaly (changes requested):
Many thanks. Looks good.
I only have these easy to resolve remarks:
- since I cannot test this right now, have you tested that files created with FreeCAD 0.20 will work fine (no warnings) also with your changes?
- please use generic file names without referencing a FreeCAD version
- please add tooltips to the UI file
- can you please merge your different commits to one

src/Mod/PartDesign/InvoluteGearFeature.ui
jbaehr: Yes, that's on my TODO list. My plan is to use the property descriptions as tool tips, in order to not maintain two times the same string. And make them translatable. Basically the same as what I've done for the Helix, cf. #5321.
I thought it would better be done as a separate PR -- separation if concerns, you know ;-)

src/Mod/PartDesign/CMakeLists.txt
jbaehr: The name is on purpose. Those files are used by unit tests to ensure we can 
a) load the old files
b) recalculate them successfully
b) come to the same result as with earlier versions. 

So if I have a unit test, asserting compat with v0.20.1, I do not want a generic name, but a name that makes clear that those are files from v0.20.1

jbaehr:
Thanks for the review.

> * since I cannot test this right now, have you tested that files created with FreeCAD 0.20 will work fine (no warnings) also with your changes?

Yes, as stated in he PR's description. I also added unit tests for this scenario to backup my claim and avoid future regressions. Since the InvoluteGear has not seen any changes lately, I guess using only v0.20.1 as reference is enough, though.

> * please use generic file names without referencing a FreeCAD version

Why? Those files are explicitly from v0.20.1, to ensure compat with that very version. If I add new features after we have released a, lets say 1.0, and I add new features for a potential 1.1, I will add additional fixtures generated with 1.0 which will then have "v1-0-0" in their name. And once we decide that backward compat with v0.20 is not needed any more, we can remove the "v0-20-1" files.

> * please add tooltips to the UI file

Will do, but on a dedicated PR, together with translatability and single-source for property docs and task panel tool tips. As there are no tool tips for all the previous fields as well, I decided to group my PRs that way.
Is that OK for you?

> * can you please merge your different commits to one

Here I oppose. I have carefully crafted my commits to address one specific concern each. I explicitly did not mix refactors/cleanups with functional changes. And I stated the motivation behind each change in the respective commit messages.  
I see a benefit of *not squashing everything together*. Not only to ease your review, but also to enable sensible bisection in case of regressions or make efficient use of git's history (e.g. via blame/annotate) in the future.
To not repeat what others have already put together very nicely, please see [this excellent article by Chris Beams](https://cbea.ms/git-commit/) with which I couldn't agree more.

src/Mod/PartDesign/InvoluteGearFeature.ui
donovaly: Fine with me.

src/Mod/PartDesign/CMakeLists.txt
donovaly: Well, only experts will understand this. Especially the reference to 0.20.1 (the ".1") is a problem. Here you add a new feature, so it will not be backported. Therefore, if you insist to include the version name, please use then only 0.20.

donovaly:
> I see a benefit of not squashing everything together.

Especially when running a Git blame this is a problem. It is often to find the PR that introduced the problem, rather than the particular commit This saves time for maintainers like me and usually the PR submitter has then already enought knowledge to check his PR.
Therefore we usually keep the commits per PR low. If we think having several commits is more suitable, we make several PRs. Sure this is a matter of taste, as a reviewer I only follow here what we usually do. If you insist to have the different commits, then I request you to split the PR to one PR in which are the commits where you add the actual feature and one where you refactor the code. Would that be OK for you?

jbaehr:
> > I see a benefit of not squashing everything together.
> 
> Especially when running a Git blame this is a problem. It is often to find the PR that introduced the problem, rather than the particular commit This saves time for maintainers like me and usually the PR submitter has then already enought knowledge to check his PR.

Well, then consider it that way: If you come back to me later, because my PR caused a problem, I definitely prefer digging though a few smaller well described  changes, then a single big one. I strive to write the git history I'd like to read.

> Therefore we usually keep the commits per PR low. If we think having several commits is more suitable, we make several PRs. Sure this is a matter of taste, as a reviewer I only follow here what we usually do.

I'm totally with when you want to keep the changes per PR low. That's also a reason why I choose not to throw in the tooltip/translation stuff right now. But just replacing a few smaller commits with a single big one, doesn't affect the changes made in the PR at all. It just makes them harder to understand.

> If you insist to have the different commits, then I request you to split the PR to one PR in which are the commits where you add the actual feature and one where you refactor the code. Would that be OK for you?

Well, "insisting" is maybe a bit too strong. In my eyes it's just a pity to throw away valuable information that may help future developers (myself included) understanding the reasoning behind certain changes. If feels like I'm forced to make my PR worse, not better. That's my main reason to be reluctant.

Sure I could split it in two or three dedicated PRs, but what do you gain by this? The changes are the same, you just get additional merges in between (and thus lose a bit of context).

src/Mod/PartDesign/CMakeLists.txt
jbaehr: I feel we still have a misunderstanding here. This is not about backporting, it's about ensuring backward compatibility.

I wrote automated tests that ensure the current version (0.21-alpha, 1.0, what ever the future brings) is still able to correctly handle gears created with FreeCAD 0.20.1. That was my safety net during development and kinda insurance for my older projects against potential regressions in the future.

As the gear's code hasn't changed in the recent past, it shouldn't really matter whether I'd use 0.20.0, 0.20.1 or 0.20.2 -- I just happen to have a 0.20.1 around to create those reference files. And because it was a 0.20.1 I suffixed them with "_v0-20-1".

Is your proposal to recreate the reference files with the original 0.20.0 and name them "..._v0-20" instead? Or is having the patch level in file name the problem? If so, why?
Or is it still unclear what I want to achieve here?

chennes:
Just as another data point: I prefer the current commit structure to one with any merging. I think @jbaehr did a very nice job structuring this PR, and I'd encourage other developers to follow this good example.

donovaly:
> Sure I could split it in two or three dedicated PRs, but what do you gain by this? The changes are the same, you just get additional merges in between (and thus lose a bit of context).

Well, once I started to do some merging work, I was explicitly told to keep the number of commits per PR low, if there are additions and refactorings, ideally the refactoring should be sorted out to a separate PR. I thought this is the case here.
You had good reasons that this is not necessary and personally, I don't have a real strong opinion, just tried to act as I usually do.
So let's keep the PR as it is, except of the naming, see my next post

In general, there is no rule how to handle the commits within a PR, so there is no right or wrong.

donovaly:
> As the gear's code hasn't changed in the recent past, it shouldn't really matter whether I'd use 0.20.0, 0.20.1 or 0.20.2 -- I just happen to have a 0.20.1 around to create those reference files. And because it was a 0.20.1 I suffixed them with "_v0-20-1".

But this is my point. the "-1" implies for other or future developers that there was a change in the 0.20 code line. But since this was not the case, the naming can be confusing (a future developer might spend time to onvestigate what might happened between 0.20.0, 0.20.1 0.20.2 etc.)

Therefore in my opinion the naming should be just "_v0-20".

> Is your proposal to recreate the reference files with the original 0.20.0 and name them "..._v0-20" instead?

Not completely. Since there were no changes in the 0.20 code line for the gear code, the name should be "..._v0-20" but there is no need to use 0.20.2 to recreate them, you can use 0.20.1 or 0.20.2.
 Or is having the patch level in file name the problem? If so, why? Or is it still unclear what I want to achieve here?

src/Mod/PartDesign/CMakeLists.txt
jbaehr: (Answering to [this comment](https://github.com/FreeCAD/FreeCAD/pull/8184#issuecomment-1399381491), which somehow got out of context)

> > As the gear's code hasn't changed in the recent past, it shouldn't really matter whether I'd use 0.20.0, 0.20.1 or 0.20.2 -- I just happen to have a 0.20.1 around to create those reference files. And because it was a 0.20.1 I suffixed them with "_v0-20-1".
> 
> But this is my point. the "-1" implies for other or future developers that there was a change in the 0.20 code line. But since this was not the case, the naming can be confusing (a future developer might spend time to onvestigate what might happened between 0.20.0, 0.20.1 0.20.2 etc.)
> 
> Therefore in my opinion the naming should be just "_v0-20".

This is a reasoning I can follow. So I kept the version number in the file name, but removed patch-level.

donovaly:
Thanks for the naming change.

-------------------
Since there internal discussions why I requested a PR split or commit merge. Here if only for information was what I initially had in mind when I see the PR the first time:
I feel like being brought to court in this respect. But following the way I review PRs, I would have done like this:
- commit 1 is one PR since it adds a testcase that can also be backported
- commit 2 and 3 can become another PR since this is code refactoring
- while commit 4 to end add a new feature and could be a separate PR

donovaly:
The last step is to update the Wiki. Can you therefore please add the new feature to our announcement list:
https://wiki.freecadweb.org/Release_notes_1.0
and update the Wiki pages describing the feature:
https://wiki.freecadweb.org/PartDesign_InvoluteGear

--------------
Please also make a PR with the tooltip addition.

jbaehr:
Thank you.

I will address those three points over the upcoming week.

PR #8223, opened 2023-01-21
Global refactoring of CI

0penBrain:
This PR is the initial submission of a global refactoring of the CI aiming at improving it both structurally and technically.

There is still much (smaller) improvements to be done, but this initial release is complete enough to be rolled into production and live.

It should not cause any regression compared to actual CI and give following improvements :
* Maintainable and scalable structure with a master workflow calling reusable workflows
  * Improves readability of workflow
  * Improves execution time by splitting workflow across different runners
  * Allows to add more jobs in parallel in the future, such as Windows building and testing
  * Currently Ubuntu 20.04 with GCC and Ubuntu 22.04 with Clang are provided
* Factorization of common tasks with local actions
  * Improves maintainability of actions
  * Currently especially FC build testing (unit test run) is provided as local actions
* Following technical improvements/fixes to current workflow are also present
  * Use of concurrency for better resource usage
  * Checks how old a PR rebase is, and optionally fail CI if PR is too old
  * Detection of files changed by the patch has been fixed
  * Ubuntu jobs (20.04 and 22.04) do tests both on CLI mode and GUI mode
  * Ubuntu 22.04 job manually install pivy to not disturb testing
  * Lint job detects files with non-Unix line endings and optionally can fail the CI if such files are found
  * Forcefully run the unit tests in a line-buffered console environment to keep log/error lines order
  * Improves a bit readability of output report

---

1st commit brings the new CI
2nd commit disables the current CI
3rd commit temporarily allows the Ubuntu 20.04 job to fail because it reveals a bug in PathAdaptative that was ignored by current CI. It shall be reverted as soon a bug is fixed

---

You can see an example of the workflow here: https://github.com/FreeCAD/FreeCAD/actions/runs/3976019257
Note : seems a bit strange that this worflow is run while it's only a PR right now...

---

I plan to continue improving the CI with further steps, but want it to be rolled out in this state so it can live and also it can then be enhanced with smaller commits.


.github/workflows/CI_master.yml
```diff
@@ -0,0 +1,67 @@
+# ***************************************************************************
+# *   Copyright (c) 2023 0penBrain                               *
```
luzpaz, line 2: fix whitespace

.github/workflows/actions/linux/generateCacheKey/action.yml
```diff
@@ -0,0 +1,42 @@
+# ***************************************************************************
+# *   Copyright (c) 2023 0penBrain                               *
```
luzpaz, line 2: whitespace

.github/workflows/actions/runTests/action.yml
```diff
@@ -0,0 +1,64 @@
+# ***************************************************************************
+# *   Copyright (c) 2023 0penBrain                               *
```
luzpaz, line 2: whitespace

.github/workflows/sub_prepare.yml
```diff
@@ -0,0 +1,165 @@
+# ***************************************************************************
+# *   Copyright (c) 2023 0penBrain                               *
```
luzpaz, line 2: whitespace

.github/workflows/sub_wrapup.yml
```diff
@@ -0,0 +1,114 @@
+# ***************************************************************************
```
luzpaz, line 1: whitespace

luzpaz (changes requested):

.github/workflows/CI_master.yml
0penBrain: It was copied from other files. All headers are like this in FC.
If this needs a fix, let's do it in a batch.

.github/workflows/sub_prepare.yml
```diff
@@ -0,0 +1,165 @@
...
+          echo "This CI run is performed on a Pull Request" | tee -a ${{env.reportdir}}${{ env.reportfilename }}
+          echo "Base SHA is $baseSha, Head SHA is $headSha" | tee -a ${{env.reportdir}}${{ env.reportfilename }}
+      - name: Check if PR has been recently rebased
+        if: env.isPR == 'true'
+        run: |
+          baseDate=$(curl -H "Accept: application/vnd.github+json" -H "authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" $GITHUB_API_URL/repos/$GITHUB_REPOSITORY/commits/$baseSha | jq -r '.commit.committer.date')
+          dateDiff=$(( ( $(date +%s) - $(date -d $baseDate +%s) ) / 3600 ))
+          echo "Pull request is based on a $dateDiff hour-old commit" | tee -a ${{env.reportdir}}${{ env.reportfilename }}
+          # Exit the step with appropriate code
+          if [ $dateDiff -le ${{ inputs.maxRebaseHours }} ]
```
0penBrain, line 96: Note to my self :)
```suggestion
          if [ $dateDiff -gt ${{ inputs.maxRebaseHours }} ]
```
Will fix when CI run of PR is over as it is used as example

.github/workflows/CI_master.yml
luzpaz: Good to know. Copied from where?

chennes:
This is a thing of beauty - thanks for taking this on.

.github/workflows/CI_master.yml
0penBrain: Another file of the codebase. :) Can't remember which one I had opened when I did that. ;)

.github/workflows/sub_prepare.yml
0penBrain: Fixed

0penBrain:
Thanks @chennes :blush:

adrianinsaval:
Impressive work!

the Path test error reported in the 2004 CI I'm also experience on archlinux

adrianinsaval:
Consider using https://github.com/actions/upload-artifact to make the compiled binaries available for download

0penBrain:
> Impressive work!
> 
> the Path test error reported in the 2004 CI I'm also experiencing on archlinux

Thanks. Which compiler are you using ? From my (limited) tests, it happens only with GCC, and only when testing on installed binary. Recently there was another error with PathAdaptative test but it seems fixed now.

> Consider also using `actions/upload-artifact` to make the compiled binaries available for download, are there storage constraints for the uploaded artifacts?

AFAIK there isn't a limit (in size, there is one in time) but it's important to me to make a fair use of resources.
What would be the goal ? Because when you get only binaries, except having the exact same environment, you won't be able to run it properly (and still you'll miss all Python files). :confused:

adrianinsaval:
> > the Path test error reported in the 2004 CI I'm also experiencing on archlinux
> 
> Thanks. Which compiler are you using ? From my (limited) tests, it happens only with GCC, and only when testing on installed binary. Recently there was another error with PathAdaptative test but it seems fixed now.

normally with gcc but I now tried compiling with clang and still have the same problem

> What would be the goal ? Because when you get only binaries, except having the exact same environment, you won't be able to run it properly (and still you'll miss all Python files).

I meant the entire thing, a similar ubuntu environment should be easy to reproduce no? I guess the snaps we currently use are more practical

chennes:
I expected the "allowedToFail: true" setting for the Build2004 CI to result in a green check, even with that one test failing. It doesn't seem to do that, so what _does_ it do?

0penBrain:
> I meant the entire thing, a similar ubuntu environment should be easy to reproduce no? I guess the snaps we currently use are more practical
Not so much IMO. Because you need to have the environment at the build time. If new versions of libs are rolled out, it can vary from one run to the other and this will be hard to know exactly what the environment was when the CI ran.
That's why package managers are so great. :smile: 

> I expected the "allowedToFail: true" setting for the Build2004 CI to result in a green check, even with that one test failing. It doesn't seem to do that, so what _does_ it do?

I should say I'm a bit surprised here too. "allowedToFail" does from my perspective what I expect. :smile: It allows the workflow run to be considered successful while the job fails. You can see it here for example (green check on the "master' workflow) : https://github.com/FreeCAD/FreeCAD/actions/runs/3976400963
The problem is that I didn't expect GH to consider all sub-workflows as individual checks, but to present only the result of the master workflow and use (only) this to mark the commit on which it is run.
Don't know much what to do with that. Have to deeper think about it.

chennes:
What about disabling that particular workflow for now? I don't want to merge if we are going to be showing a failing test for every single commit going forward.

0penBrain:
Sure we could. But still I won't be satisfied how the result is handled and displayed in GH. :)

chennes:
Do you want to mark this PR as on-hold until the build issue is resolved? Or change the structure so GitHub displays it better? Or what?

0penBrain:
> Do you want to mark this PR as on-hold until the build issue is resolved? Or change the structure so GitHub displays it better? Or what?

Needed some thoughts about that. :) I think the PR brings a quite significant amount of improvements so at now I disabled the 20.04 sub-workflow as a temporary measure (job is forcefully skipped).
I will report the problem to GH a bit later and see if one there is interested in improving things.
Depending on that, I'll eventually make changes to better handle this.

chennes:
How do you feel about merging this as-is, @0penBrain ?

0penBrain:
@chennes yes, no problem to merge as is.

PR #8228, opened 2023-01-22
Expressions: path auto-completion fixes

acpopescu:
I've added support to do path completions after a property is selected in the auto completer line edit.

This allows:

image

image

It also fixes some UX issues I had with the completer - chaining completions when pressing Enter should work.

It is not perfect, as it cannot complete sub-path components, and has other limitations.
- `Sketch.Placement.` will propose `Base.x`, `Base.y`,... `Rotation.Angle`, `Rotation.Axis.x`
- `Sketch.Placement.Rotation.A` will fail to complete as the path split "Sketch/Placement/Rotation/A" is not in the completion tree, only Sketch/Placement/Rotation.Axis.x is 
- I don't believe arrays will work, if there isn't a named structure in place the completer will not have that in the completion tree: `Object.Prop[10].Something` will not work, as `Prop[10]` is not in the tree. 

Tested:
- Constraints formula line edit
- Spreadsheet 
- x64 Windows Debug & Release builds

Video:
https://youtu.be/y-3bDFy8Pgc

---

### DIFF
--- a/src/Gui/ExpressionCompleter.cpp
+++ b/src/Gui/ExpressionCompleter.cpp
@@ -105,6 +105,24 @@ public:
     //
     // QModelIndex of a non-root property item has doc field as the document
     // index, and obj field as the object index.
+    //
+    // An item is uniquely identified by the pair (row, father_link) in the QModelIndex
+    // 
+    // The completion tree structure created takes into account the current document and object
+    //
+    // It is done as such, in order to have contextual completion (prop -> object -> files):
+    // * root (-1,-1)
+    // |
+    // |----- documents 
+    // |----- current documents' objects [externally set]
+    // |----- current objects' props  [externally set]
+    // 
+    // This complicates the decoding schema for the root, where the childcount will be
+    // doc.size() + current_doc.Objects.size() + current_obj.Props.size().
+    // 
+    // This is reflected in the complexity of the DATA function.
+    //
+
     union Info {
         struct {
             qint32 doc;
@@ -149,8 +167,25 @@ public:
         return v;
     }
 
+   // The completion tree structure created takes into account the current document and object
+    //
+    // It is done as such:
+    // * root (-1,-1)
+    // |
+    // |----- documents
+    // |----- current documents' objects [externally set]
+    // |----- current objects' props  [externally set]
+    //
+    // This complicates the decoding schema for the root, where the childcount will be
+    // doc.size() + current_doc.Objects.size() + current_obj.Props.size().
+    //
+    // this function is called in two modes: 
+    // - obtain the count of a node identified by Info,row => count != nullptr, v==nullptr
+    // - get the text of an item. This text will contain separators but NO full path
     void _data(const Info &info, int row, QVariant *v, int *count, bool sep=false) const {
         int idx;
+        // identify the document index. For any children of the root, it is given by traversing
+        // the flat list and identified by [row]
         idx = info.d.doc<0?row:info.d.doc;
         const auto &docs = App::GetApplication().getDocuments();
         int docSize = (int)docs.size()*2;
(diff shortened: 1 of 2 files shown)


acpopescu:
Thanks @luzpaz  for editing the description!

adrianinsaval:
This is fixing an old and annoying regression so maintainers if possible please give this priority

acpopescu:
I've some more cases I found where it didn't work as expected. 
`<<documentName>>#` nor `<<documentName>>` will not complete due to the parser throwing errors from a part of the code I did not modify 

` App::ObjectIdentifier p = ObjectIdentifier::parse(
                    currentObj.getObject(), path);`

`<<documentName>>` is a valid completion tree entry, but we're sending an empty path list in `splitPath `because of the parser errors

src/Gui/ExpressionCompleter.cpp
```diff
@@ -246,75 +391,207 @@ class ExpressionCompleterModel: public QAbstractItemModel {
...
     }
 
+    // function returns how many children the QModelIndex parent has
     int rowCount(const QModelIndex & parent = QModelIndex()) const override {
         Info info;
         int row = 0;
         if(!parent.isValid()) {
-            info.d.doc = -1;
-            info.d.obj = -1;
+            // root.
```
chennes, line 581: Did you mean to leave this here?

src/Gui/ExpressionCompleter.cpp
```diff
@@ -144,14 +231,32 @@ class ExpressionCompleterModel: public QAbstractItemModel {
...
+                << info.contextualHierarchy << "," << index.row()
                 << ": " << v.toString().toUtf8().constData());
         return v;
     }
 
+   // The completion tree structure created takes into account the current document and object
+    //
+    // It is done as such:
+    // * root (-1,-1)
+    // |
```
acpopescu, line 244: This order causes issues with contextual completion. 
If you have a document, or an object with the same name as a property, you will not be able to use completion to reach that property, because the completer's contextual solve will pick up the object rather than the prop.

Even swapping these around will still make it ambiguous.  (did you mean the prop or the object)

The Assembly 4 workbench has an issue that the model contains an 'Constraints' Object :(. As such any local object referencing "Constraints" won't be resolved properly.

src/Gui/ExpressionCompleter.cpp
```diff
@@ -399,36 +679,75 @@ QStringList ExpressionCompleter::splitPath ( const QString & input ) const
...
     int retry = 0;
-    std::string trim;
+    std::string lastElem;  // used to recover in case of parse failure after ".". 
+    std::string trim;      // used to delete ._self added for another recovery path 
     while(1) {
         try {
+            // this will not work for incomplete Tokens at the end
+            // "Sketch." will fail to parse and complete.
+
             App::ObjectIdentifier p = ObjectIdentifier::parse(
```
acpopescu, line 689: This code will collapse any global references to the current object to the sub path.

It causes issues when you have objects or props with the same name as documents, or other objects. 

Example `Constraints` object in Assembly4 blocks completion for a local `Constraints`. Looking at solutions.

src/Gui/ExpressionCompleter.cpp
acpopescu: It may work to swap these around, it depends if the split path code parser doesn't collapse the current document references.

Maybe there is a workaround (with ._self)? but I don't know FreeCAD very well.

chennes:
Do you want to switch this to Draft status and continue working on it, or merge it as-is and submit follow-up PRs as you make further incremental improvements? My inclination is to merge this as-is because it provides and improvement, even if there is still work to be done.

acpopescu:
I agree - let's merge this as is - Other fixes can come in later.

chennes:
Are you attached to the commit structure you've got? In this instance I'm not sure having all these separate commits in the final makes much sense, I'm inclined to squash and merge.

acpopescu:
Not really attached to the commits, no, we can squash it - do you need me to do anything else?

chennes:
I am seeing a small regression in the behavior when I use the arrow keys to select a completion from the drop-down menu. Even in the case of selecting a "final" component of a completion, a dot is appended (which then must be deleted). Is this a known issue? I don't see the same behavior in master.

acpopescu:
@chennes  - what's the case you're checking? Are you finishing on a property, or an object? Is it a contextual one (from currentDoc or current Obj)?

acpopescu:
I found an issue where the property would add a "." at the end because it wouldn't check correctly in all cases that there are no paths accessible

chennes:
Perfect, that addressed the issue I was seeing. Given that this addresses a very longstanding complaint and pain-point for FreeCAD users I'm going to merge it now and we'll have to stay tuned for any further regression reports from beta-testers (fortunately we have lots of those!).

acpopescu:
Awesome! Thanks for this! I'll keep an eye out for regressions!

donovaly:
This PR introduces these new compiler warnings:

```
/home/runner/work/FreeCAD/FreeCAD/src/Gui/ExpressionCompleter.cpp:451:32: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<App::ObjectIdentifier>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  451 |             if (idx < 0 || idx >= paths.size()) {
      |                            ~~~~^~~~~~~~~~~~~~~
/home/runner/work/FreeCAD/FreeCAD/src/Gui/ExpressionCompleter.cpp: In member function ‘bool ExpressionCompleterModel::modelIndexToParentInfo(QModelIndex, ExpressionCompleterModel::Info&) const’:
/home/runner/work/FreeCAD/FreeCAD/src/Gui/ExpressionCompleter.cpp:529:35: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<App::Document*>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  529 |                 if (element.row() >= docs.size()*2) {
      |                     ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
/home/runner/work/FreeCAD/FreeCAD/src/Gui/ExpressionCompleter.cpp:540:29: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<App::DocumentObject*>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  540 |                     if (idx < cdoc->getObjects().size()*2) {
      |                         ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/FreeCAD/FreeCAD/src/Gui/ExpressionCompleter.cpp:538:26: warning: unused variable ‘objsSize’ [-Wunused-variable]
  538 |                     auto objsSize = cdoc->getObjects().size()*2;
      |                          ^~~~~~~~
```

Can you please have a look?

wwmayer:
#8259

donovaly:
@acpopescu , additionally to the fixes made by #8259, There are further issues. Can you therefore please have a look at the reports from [GitHub actions](https://github.com/FreeCAD/FreeCAD/pull/8259/commits/b0a87025c5da629783a5f08648a10d1894f42f87)?

From these, especially these things are in my opinion important:
- please change the connect() to the Functor-Based Connection as we do in FreeCAD now in all places, because this is necessary for Qt6
- there are magic numbers. For example in ` d_enc = ((quint64(info.doc+1) & ((1ULL << 23) - 1)) << 41ULL)`
Can you please add comments why there is e.g. the 23 to make it better to understand for others?

acpopescu:
Hi @donovaly, I'll take a look!

@wwmayer Thanks for pushing that cleanup PR! - Visual Studio does not report these warnings when compiling Debug or Release.

donovaly:
Thanks. I am also on Visual studio and have the same problem that it does not know much about Qt. Therefore once I made a PR I started to look at the GitHub action reports. Note hereby that some of its messages can be ignored, so you are not forced to make them all disappear.

By the way, since a while, I can compile a debug build, but when I ran the debug build and add then a using the Part WB to add a box object, it crashes. I spent about 2 hours with compiling to assure I have not made a mistake and the problem persists. If you get the same problem, I will report this to the developer's forum as regression.

acpopescu:
Ah, right - I saw an old forum post about a crash on delete for debug builds.

The issue seems to be that the CMAKE install for the debug configuration or the FREECAD_COPY_* CMAKE options actually put the Release DLLS from the LIBPACK instead of the debug ones (TK*.dlls). 

I had to manually copy over `LibPack-OCC76-V2\bind` and `LibPack-OCC76-V2\libd` over in the installation folder, and that fixed the issue.

donovaly:
> I had to manually copy over `LibPack-OCC76-V2\bind` and `LibPack-OCC76-V2\libd` over in the installation folder, and that fixed the issue.

Many thanks! That fixes it. I will describe this now in https://wiki.freecadweb.org/Compile_on_Windows#Running_and_installing_FreeCAD

donovaly:
Note that the latest LibPack is version 2.8.2: https://github.com/FreeCAD/FreeCAD-LibPack/releases
(if you find there bugs, please report them in the developer's forum and mention me (I am in the forum "uwestoehr" and I maintain the LibPack at the moment)

acpopescu:
@donovaly  - I've created the cleanup Pull Request here: https://github.com/FreeCAD/FreeCAD/pull/8262

PR #8234, opened 2023-01-22
Draft: V0.18 VisGroups were not handled

Roy-043:
Added code to handle V0.18 VisGroups. It is a bit after the fact, but it is also strange to not handle them considering other code that is in place to handle V0.18 objects.

---


donovaly:
@Roy-043 , please don't merge PRs for which your get CI errors. Here you deleted a .py file but it is still in the CMake file

Roy-043:
Most CI issues are either irrelevant or point to problems with other PRs. This is an exception to that rule.

donovaly:
> Most CI issues are either irrelevant or point to problems with other PRs. This is an exception to that rule.

I cannot second that. it often helped me to find issues because the CIs use bath another compiler and another compiler than I do. That's the main idea, to assure it compiles for everyone. It also shows me typos and what I like very much are the reports how I can improve the code. This is also the case your your PR, check the report of [GitHub actions](https://github.com/FreeCAD/FreeCAD/pull/8234/commits/82663021f538020af35b24e7325907f386d4b37b)

Besides this, we have the policy not to merge own PRs. This can however be done if no other developer has time or you are the only only understanding a certain field. So there is no strict rule but you should give non-trivial things (new features etc.) some days before you merge.

Roy-043:
The CI messaged I see are mostly about lines that are too long, or about properties that are references before they are assigned. GitHub CI apparently is not aware that objects can already exist and can already have properties.

This PR is hardly controversial. But if I follow the new protocol (create an issue first, discuss, then PR) I am looking at a 7 weeks delay (4 weeks waiting for a discussion that does not happen, and then 3 weeks for the PR to be approved). Not very inspiring needless to say.

PR #8236, opened 2023-01-22
[Sketch] Constraints.h: fix 2 compiler warnings

donovaly:
- the type change fixes this error: \src\Mod\Sketcher\App\planegcs\Constraints.cpp(509,54): warning C4267: '=': conversion from 'size_t' to 'unsigned int', possible loss of data
- the scone change removes a non-existing definition

### DIFF
--- a/src/Mod/Sketcher/App/planegcs/Constraints.h
+++ b/src/Mod/Sketcher/App/planegcs/Constraints.h
@@ -251,10 +251,9 @@ namespace GCS
         void rescale(double coef=1.) override;
         double error() override;
         double grad(double *) override;
-        void setupInputs();
         size_t numpoints;
         BSpline& bsp;
-        unsigned int startpole;
+        size_t startpole;
     };
 
     // Difference


AjinkyaDahale:
If it removes the compiler warning I'm fine. The variable `startpole` is fine as `size_t` since it is the index of the first pole affecting the relevant segment of the spline.

AjinkyaDahale (approved):
(See comment outside review)

donovaly:
> If it removes the compiler warning I'm fine. The variable `startpole` is fine as `size_t` since it is the index of the first pole affecting the relevant segment of the spline.

Thanks, so this can go in.

For info from https://stackoverflow.com/questions/502856/whats-the-difference-between-size-t-and-int-in-c:
`The actual type of size_t is platform-dependent; a common mistake is to assume size_t is the same as unsigned int, which can lead to programming errors, particularly as 64-bit architectures become more prevalent.`

donovaly:
I see that GitHub actions report several thing we might need to fix. Can you please have a look at them?:
[actions output](https://github.com/FreeCAD/FreeCAD/pull/8236/commits/12a52646a2d46e58f1293850d9b13f81c5c42cdb)

AjinkyaDahale:
@donovaly have to file them under "above my pay grade". I think there was a discussion about the macros as well right, for the first declaration (`PROTECTED_UNLESS...`)?

PR #8249, opened 2023-01-24
README: fix bugtracker link

luzpaz:
* Fix bugtracker link at the top of the README

---


donovaly:
Thanks for the fix.
However, the change to centered is not yet clear to me. Personally, I like justified more because the monitor size does then not change too much. But sure matter of taste.
Since the fix should go in and the other change will maybe raise a debate, can you please make a separate PR for the fix?

luzpaz:
@donovaly I've reverted the centering and just kept the bugtracker link fix.  
JFYI, the monitor size doesn't change. I just didn't take the screenshot accurately.

PR #8251, opened 2023-01-24
[Part] fixes for the section cutting feature

donovaly:
- a typical use-case is to open a document with an existing cut and only this cut is visible. When now opening the section cutting tool, the user got the fault message that there were no visible objects to be cut.
- the transparency for the cut was not explicitly set. Therefore the default transparency for new objects was used instead of the transparency of the objects to be cut.
- workaround for a graphics issue: when cutting objects intersection each other (only then) the transparency setting is often ignored. The fix is simply to change the default color slightly. (By the way, this issue is independent on the color that is set in the Part preferences as color for new objects, seems to be a  graphics driver or OCC issue.)

- also avoid to recompute all cuts on color change. This it is not necessary and speeds up the workflow for large assemblies.

- besides this, avoid code duplication by using a lambda function

### DIFF
--- a/src/Mod/Part/Gui/SectionCutting.cpp
+++ b/src/Mod/Part/Gui/SectionCutting.cpp
@@ -1576,9 +1576,39 @@ void SectionCut::onFlipZclicked()
 // changes the cutface color
 void SectionCut::onCutColorclicked()
 {
-    // re-cut to change the color of all cut boxes
     if (ui->groupBoxX->isChecked() || ui->groupBoxY->isChecked() || ui->groupBoxZ->isChecked())
-        startCutting();
+        changeCutBoxColors();
+}
+
+// changes cutbox colors
+void SectionCut::changeCutBoxColors()
+{
+    // a lambda to set box color and transparency
+    auto setColorTransparency = [&](App::DocumentObject* boxObject) {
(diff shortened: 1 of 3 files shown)


src/Mod/Part/Gui/SectionCutting.cpp
```diff
@@ -147,6 +147,20 @@ SectionCut::SectionCut(QWidget* parent)
                 return;
             }
             BoundCompound = pcCompound->Shape.getBoundingBox();
+            // make parent objects of links visible to handle the case that
+            // the cutting is started when only an existing cut was visible
+            std::vector<App::DocumentObject*> compoundObjects;
+            pcCompound->Links.getLinks(compoundObjects);
+            for (auto itCompound = compoundObjects.begin(); itCompound != compoundObjects.end();
```
chennes, line 154: Is there a reason not to use a range-based for loop here?

donovaly: Done now. Since this PR is meant to fix issues and to be backported to 0.20.x. I will make the general loop code modernization in #8252

donovaly:
@chennes can this be merged? if so I could move on a bit further with SectionCutting.

PR #8252, opened 2023-01-24
[Part] extend SectionCutting feature for intersecting objects

donovaly:
- it is a well-known and often feed-backed missing feature that SectionCuttings fails for intersecting objects. To resolve this, the objects must be put int a BooleanFragments object.
- this PR adds this functionality as option.
- since a BooleanFragments objects has a specific color, the different colors of the objects cannot be preserved. Because of this disadvantage, the BooleanFragments option will not be the default.
- the PR includes the fixes from #8251

Here is the result with master:
FreeCAD_EIBfDUZpZ3

And here with the PR:
FreeCAD_asdMXu5w6v

### DIFF
--- a/src/Mod/Part/Gui/SectionCutting.cpp
+++ b/src/Mod/Part/Gui/SectionCutting.cpp
@@ -110,9 +110,9 @@ SectionCut::SectionCut(QWidget* parent)
         return;
     }
     // now store those that are currently visible
-    for (auto it = ObjectsList.begin(); it != ObjectsList.end(); ++it) {
-        if ((*it)->Visibility.getValue())
-            ObjectsListVisible.emplace_back(*it);
+    for (auto anObject : ObjectsList) {
+        if (anObject->Visibility.getValue())
+            ObjectsListVisible.emplace_back(anObject);
     }
 
     // lambda function to set color and transparency
@@ -125,8 +125,8 @@ SectionCut::SectionCut(QWidget* parent)
             cutColor = vpBox->ShapeColor.getValue();
             cutTransparency = vpBox->Transparency.getValue();
             ui->CutColor->setColor(cutColor.asValue<QColor>());
-            ui->CutTransparency->setValue(cutTransparency);
-            ui->CutTransparency->setToolTip(QString::number(cutTransparency)
+            ui->CutTransparencyHS->setValue(cutTransparency);
+            ui->CutTransparencyHS->setToolTip(QString::number(cutTransparency)
                                             + QString::fromLatin1(" %"));
         }
     };
@@ -137,21 +137,56 @@ SectionCut::SectionCut(QWidget* parent)
     Base::BoundBox3d BoundCutBox;
     if (doc->getObject(BoxXName) || doc->getObject(BoxYName) || doc->getObject(BoxZName)) {
         // automatic coloring must be disabled
-        ui->AutoCutfaceColor->setChecked(false);
+        ui->autoCutfaceColorCB->setChecked(false);
+        ui->autoBFColorCB->setChecked(false);
         if (doc->getObject(CompoundName)) {
+            // get the object with the right name
             auto compoundObject = doc->getObject(CompoundName);
-            Part::Compound* pcCompound = dynamic_cast<Part::Compound*>(compoundObject);
-            if (!pcCompound) {
-                Base::Console().Error(
-                    "SectionCut error: compound is incorrectly named, cannot proceed\n");
-                return;
+            // to later store the childs
+            std::vector<App::DocumentObject*> compoundChilds;
+
+            // check if this is a BooleanFragments or a Part::Compound
+            // Part::Compound is the case when there was only one object
+            Part::Compound* pcCompoundPart = dynamic_cast<Part::Compound*>(compoundObject);
+            if (!pcCompoundPart) {
+                // for more security check for validity accessing its ViewProvider
+                auto pcCompoundBF = Gui::Application::Instance->getViewProvider(compoundObject);
+                if (!pcCompoundBF) {
+                    Base::Console().Error(
+                        "SectionCut error: compound is incorrectly named, cannot proceed\n");
+                    return;
+                }
+                auto property = compoundObject->getPropertyByName("Shape");
+                Part::PropertyPartShape* objectShape =
+                    static_cast<Part::PropertyPartShape*>(property);
+                BoundCompound = objectShape->getBoundingBox();
+                // for BooleanFragments we also need to set the checkbox, transparency and color
+                ui->groupBoxIntersecting->setChecked(true);
+                auto pcCompoundBFGO = dynamic_cast<Gui::ViewProviderGeometryObject*>(pcCompoundBF);
+                if (pcCompoundBFGO) {
+                    App::Color compoundColor = pcCompoundBFGO->ShapeColor.getValue();
+                    ui->BFragColor->setColor(compoundColor.asValue<QColor>());
+                    int compoundTransparency = pcCompoundBFGO->Transparency.getValue();
+                    ui->BFragTransparencyHS->setValue(compoundTransparency);
+                    ui->BFragTransparencyHS->setToolTip(QString::number(compoundTransparency)
+                                                        + QString::fromLatin1(" %"));
+                    // Part::Cut ignores the cutbox transparency when it is set
+                    // to zero and the BooleanFragments transparency is not zero
+                    // therefore limit the cutbox transparency to 1 in this case
+                    if (compoundTransparency > 0)
+                        ui->CutTransparencyHS->setMinimum(1);
+                    else
+                        ui->CutTransparencyHS->setMinimum(0);
+                }
+                compoundChilds = pcCompoundBF->claimChildren();
             }
-            BoundCompound = pcCompound->Shape.getBoundingBox();
+            else {
+                BoundCompound = pcCompoundPart->Shape.getBoundingBox();
+                pcCompoundPart->Links.getLinks(compoundChilds);                
+            }
             // make parent objects of links visible to handle the case that
             // the cutting is started when only an existing cut was visible
-            std::vector<App::DocumentObject*> compoundObjects;
-            pcCompound->Links.getLinks(compoundObjects);
-            for (auto aCompoundObj : compoundObjects) {
+            for (auto aCompoundObj : compoundChilds) {
                 App::Link* pcLink = dynamic_cast<App::Link*>(aCompoundObj);
                 auto LinkedObject = pcLink ? pcLink->getLink() : nullptr;
                 if (LinkedObject) {
@@ -273,9 +308,17 @@ SectionCut::SectionCut(QWidget* parent)
     connect(ui->flipZ, &QPushButton::clicked, this, &SectionCut::onFlipZclicked);
     connect(ui->RefreshCutPB, &QPushButton::clicked, this, &SectionCut::onRefreshCutPBclicked);
     connect(ui->CutColor, &QPushButton::clicked, this, &SectionCut::onCutColorclicked);
-    connect(ui->CutTransparency, &QSlider::sliderMoved, this,
-            &SectionCut::onTransparencySliderMoved);
-    connect(ui->CutTransparency, &QSlider::valueChanged, this, &SectionCut::onTransparencyChanged);
+    connect(ui->CutTransparencyHS, &QSlider::sliderMoved,
+            this, &SectionCut::onTransparencyHSMoved);
+    connect(ui->CutTransparencyHS, &QSlider::valueChanged,
+            this, &SectionCut::onTransparencyHSChanged);
+    connect(ui->groupBoxIntersecting, &QGroupBox::toggled,
+            this, &SectionCut::onGroupBoxIntersectingToggled);
+    connect(ui->BFragColor, &QPushButton::clicked, this, &SectionCut::onBFragColorclicked);
+    connect(ui->BFragTransparencyHS, &QSlider::sliderMoved,
+            this, &SectionCut::onBFragTransparencyHSMoved);
+    connect(ui->BFragTransparencyHS,&QSlider::valueChanged,
+            this, &SectionCut::onBFragTransparencyHSChanged);
 
     // if there is a cut, perform it
     if (hasBoxX || hasBoxY || hasBoxZ) {
@@ -314,7 +357,6 @@ void SectionCut::startCutting(bool isInitial)
         onRefreshCutPBclicked();
 
     App::DocumentObject* anObject = nullptr;
-    std::vector<App::DocumentObjectT>::iterator it;
 
     // lambda function to delete objects
     auto deleteObject = [&](const char* objectName) {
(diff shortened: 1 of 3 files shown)


src/Mod/Part/Gui/SectionCutting.cpp
```diff
@@ -137,16 +137,65 @@ SectionCut::SectionCut(QWidget* parent)
...
+                                                        + QString::fromLatin1(" %"));
+                    // Part::Cut ignores the cutbox transparency when it is set
+                    // to zero and the BooleanFragments transparency is not zero
+                    // therefore limit the cutbox transparency to 1 in this case
+                    if (compoundTransparency > 0)
+                        ui->CutTransparencyHS->setMinimum(1);
+                    else
+                        ui->CutTransparencyHS->setMinimum(0);
+                }
+                compoundChilds = pcCompoundBF->claimChildren();
```
chennes, line 181: The CI complaint about cognitive complexity is primarily due to the deeply-nested if statements. You could consider splitting this block into its own function, which would probably about cut the complexity in half.

chennes: Regarding where I see it: when you view the "Files" section of the PR, the CI's checker warnings display inline in the code. They are only warnings, so don't cause the checks to fail.

donovaly: > I have not seen any modern software design advice advocating for **ten** nesting levels. Are you sure you aren't seeing references to 10 conditionals, rather than 10 nested levels?

They are not advocating it, say one should keep it low and not exceed more than 10. (which is obvious since this is then no longer to get by a human)

> It certainly improves the code's _testability_, but I am less convinced that it is improving its _understandability_. But perhaps it is, I'm too close to the code to tell for sure.

This is also my issue. I try to code so that others but also myself can understand what is done in e.g. 2 years. Therefore I use a lot of comments and split to functions and lambdas for repeated tasks.
I had a look and here i would keep the if structure because the logic is not trivial and by having it in a single procedure keeps the reading flow to understand what is done and why. I fear that when i sort out some ifs, another one (and myself in future of course), has to scroll around to see what is done in the functions that are called. This is not helpful to get the logic.
So for me the question is what would be the benefit. The code execution is not quicker, memory space is not saved, it is a bit harder to understand the code logic.
However, I might be too deep into it at the moment. I rely on your judgement here.

donovaly: > Regarding where I see it: when you view the "Files" section of the PR, the CI's checker warnings display inline in the code. They are only warnings, so don't cause the checks to fail.

So you mean GitHub Actions? The report is
https://clang.llvm.org/extra/clang-tidy/checks/readability/function-cognitive-complexity.html
I never heard of this yet. To me the limit of 25 is rather arbitrary. As a physicist I always ask myself why not 24 or 26. If there is no good explanation, I am sceptical.
Sure the nesting level should not be to great but nesting is hereby just one factor. The besting depth is in my case 5, but the main cutting void is large. It is, but it covers the whole cutting process. My idea was hereby that I can follow to read this void and can step through the whole process and understand it. Now that I haven't touched the cutting code for months, it helped me to find the issues. I stepped through the code and this way could identify the position in the process where an issue occurs/is created.
But again, I am the author of that code, what is great for me, can for others be a nightmare. That's why I will rely on your judgement.

However, in the CMake build report of the CI I see 
`warning: unused variable ‘compoundTransparency’`
so there is a bug in my code. I will have a look.

donovaly: > so there is a bug in my code. I will have a look.

Now fixed. I also modernized now all for loops

chennes: Certainly the 25 limit is arbitrary -- so is the definition of "complexity" itself. Linters like this are meant to be advice, not a hard-and-fast rule. Clean Code would argue that a complexity of 25 is itself absurdly high, but I bet a linter run on all of FreeCAD yields thousands of functions over that limit. Which is to say: this code is certainly no worse than other vast swaths of our legacy codebase, and until we are ready to start adding unit tests for real, it's sort of academic. Break it up or not, your call. That said, if you ever have free time for some "light" reading, Clean Code is a very highly regarded treatise on the subject, even if you don't follow Agile/TDD/etc.

donovaly: OK. For now and for the given reasons I would keep the structure as it is. One this has been merged, I expect that there are still issues under special circumstances and I would like to resolve them before I refactor the code. I expect the issues because SectionCutting is complex in terms of a high possible number of object combinations (normal objects, links, parts, part childs, links of links etc.

donovaly:
I found 2 issues that I could fix. 
From my perspective the PR is ready to go in.

src/Mod/Part/Gui/SectionCutting.cpp
wwmayer: >  I fear that when i sort out some ifs, another one (and myself in future of course), has to scroll around to see what > is done in the functions that are called. This is not helpful to get the logic.

As Uncle Bob has already been mentioned in this context. One of the aspects of clean code and SRP is that a function should do exactly one thing. Mixing levels of abstraction inside a function should be avoided because it's confusing as you have to keep in mind the higher level logic and at the same time get right the lower level logic. By splitting this into several small functions or even classes you can separate the different levels of abstractions.
The function names should be carefully chosen so that just by reading the name a reader already knows what it is supposed to do. Don't hesitate to use lengthy function names.

> So for me the question is what would be the benefit. The code execution is not quicker, memory space is not saved, it is a bit harder to understand the code logic.

Forget about execution speed or memory. Compilers nowadays are smart enough to optimize the assembler code better than an average programmer could do. The idea is to split the entire complexity into smaller less complex parts and the advantage is that the code becomes much more readable, maintainable and extendable. And of course it will be much easier to write test functions.

Things might sound too abstract to you. So, I give you a concrete example of some refactoring I did a few weeks ago. The implementation of CmdPartDesignNewSketch in the past was this big mess:
https://github.com/FreeCAD/FreeCAD/blob/674bd3d6ec59a46dbcfa2f9891eb085e12976e44/src/Mod/PartDesign/Gui/Command.cpp#L494

After doing some iterations of code refactoring it has become a two-liner:
```
PartDesignGui::SketchWorkflow creator(getActiveGuiDocument());
creator.createSketch();
```

The whole logic has been moved to a new class SketchWorkflow and the entry function is createSketch() -- https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/PartDesign/Gui/SketchWorkflow.cpp#L642

donovaly: > Forget about execution speed or memory. Compilers nowadays are smart enough to optimize the assembler code better than an average programmer could do. The idea is to split the entire complexity into smaller less complex parts and the advantage is that the code becomes much more readable, maintainable and extendable. And of course it will be much easier to write test functions.

As I wrote, I am open for a revision. However, at first, I would like to have the PR before I start a revision. This would be much easier for me.
Besides this, the SectionCutting code is now in for a while and you once approved it. It was a month-long process of getting it ready to be merged. In that time I could have written it in the way you prefer ;-).

I think that I take care of that my code is understandable. It is a personal preference that I like to follow a logic line by line without the need to scroll around to see what the different called functions do. For repeated tasks I already use lambdas whenever possible. When this is not suitable to be able to write test functions, then I need to learn how long a void can be.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252">pipeline status</a> for feature branch [PR_8252](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252). Pipeline [761016507](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/761016507) was triggered at [27e817b](https://github.com/FreeCAD/FreeCAD/pull/8252/commits/27e817bc0a53bbb3c05bb566d8fa9da1a6acfa0a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252">pipeline status</a> for feature branch [PR_8252](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252). Pipeline [770699464](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/770699464) was triggered at [a6fa0fa](https://github.com/FreeCAD/FreeCAD/pull/8252/commits/a6fa0fafa57a6d9e73600bf0c178d7442e976fc8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252">pipeline status</a> for feature branch [PR_8252](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252). Pipeline [772153562](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/772153562) was triggered at [397c0cc](https://github.com/FreeCAD/FreeCAD/pull/8252/commits/397c0ccaa781a025784fada62e65322e93a75d1a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/Gui/SectionCutting.cpp
```diff
@@ -137,21 +137,56 @@ SectionCut::SectionCut(QWidget* parent)
...
+            std::vector<App::DocumentObject*> compoundChilds;
+
+            // check if this is a BooleanFragments or a Part::Compound
+            // Part::Compound is the case when there was only one object
+            Part::Compound* pcCompoundPart = dynamic_cast<Part::Compound*>(compoundObject);
+            if (!pcCompoundPart) {
+                // for more security check for validity accessing its ViewProvider
+                auto pcCompoundBF = Gui::Application::Instance->getViewProvider(compoundObject);
+                if (!pcCompoundBF) {
+                    Base::Console().Error(
```
wwmayer, line 155: There are many places in this class where you write an error and return from the function. A better way is to throw an exception and handle it in the calling instance. Especially for the overdue refactoring this will make things much easier.

src/Mod/Part/Gui/SectionCutting.cpp
```diff
@@ -110,9 +110,9 @@ SectionCut::SectionCut(QWidget* parent)
         return;
```
wwmayer, line 110: The "ui" member is a raw pointer and not deleted. It's best to convert it into a std::unique_ptr.

wwmayer:
The code basically works and I haven't spot a bug or regression. So, from my point of view it can be merged. But as said after the merge a refactoring of almost all methods must be done.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252">pipeline status</a> for feature branch [PR_8252](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252). Pipeline [775334259](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/775334259) was triggered at [35df3bd](https://github.com/FreeCAD/FreeCAD/pull/8252/commits/35df3bda199e6c7ab550e52e6603fb4d8197e408). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252">pipeline status</a> for feature branch [PR_8252](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252). Pipeline [775337084](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/775337084) was triggered at [1e925f7](https://github.com/FreeCAD/FreeCAD/pull/8252/commits/1e925f7fbdff8856808fd851d940d9751a58f148). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/Gui/SectionCutting.cpp
donovaly: Thanks, I fixed this now.

src/Mod/Part/Gui/SectionCutting.cpp
donovaly: > A better way is to throw an exception and handle it in the calling instance

OK. I lack here knowledge - one one hand how it is done and on the other hand the technical insights about the advantage.
Maybe you have a pointer/link for me to read in?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252">pipeline status</a> for feature branch [PR_8252](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8252). Pipeline [775341199](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/775341199) was triggered at [f5a665e](https://github.com/FreeCAD/FreeCAD/pull/8252/commits/f5a665ec4d7a5b95deeb800595c7de297b318553). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/Gui/SectionCutting.cpp
wwmayer: Example:
Assuming you have a big function f() and after refactoring you extracted a further functions g(). Inside g() you want to stop the further execution for some reason. With the current approach the code will have this structure:
```
void f()
{
...
   if (g() < 0)
       return;
...
}

int g()
{
...
    if (...) {
        PrintError();
        return -1;
    }
...
}
```
Imagine after a full refactoring you could split some functions into many small and cascaded functions. Now if you want to stop in the innermost function you have to transport the error code to  the calling instance, do this for the next calling instance and so on.
Here error handling is too much bundled with the actual business logic and thus it becomes hardly to maintain. 

By using exceptions you can concentrate on the business logic and handle the exceptions in the outermost function.
```
void f()
{
   try {
      g();
   }
   catch (const exception& e) {
      PrintError(e.what());
   }
}

void g()
{
...
    if (...) {
        throw exception("bla bla bla");
    }
...
}
```
For more details see:
https://docs.oracle.com/javase/tutorial/essential/exceptions/advantages.html
https://www.geeksforgeeks.org/c-exception-handling-question-1/?ref=rp

donovaly: > Example: Assuming you have a big function f() and after refactoring you extracted a further functions g(). Inside g() you want to stop the further execution for some reason...

Many thanks! This is an excellent description and I understood this.
So I will apply this when refactoring.
Due to the pending funded FEM project, I can however, probably first do this in April.

PR #8255, opened 2023-01-24
Remove misleading message

berniev:
From commit message: (added by luzpaz)
> " Consider setting Qt5_DIR or Qt6_DIR (as needed)."
>Qt5_DIR is only used in libpacks
>Qt6_DIR is never referenced.

---

### DIFF
--- a/cMake/FreeCAD_Helpers/ChooseQtVersion.cmake
+++ b/cMake/FreeCAD_Helpers/ChooseQtVersion.cmake
@@ -77,7 +77,7 @@ macro(ChooseQtVersion)
       find_package(Qt6 QUIET COMPONENTS Core)
       if (NOT Qt6_FOUND)
         message(FATAL_ERROR
-          "Could not find a valid Qt installation. Consider setting Qt5_DIR or Qt6_DIR (as needed).")
+          "Could not find a valid Qt installation.")
       endif ()
       set(_FREECAD_QT_VERSION 6)
     endif ()


donovaly:
> Thank you for creating a pull request to contribute to FreeCAD! 

Can you please replace this generic text when you make PRs and describe what the PR is about?

Concerning the PR, for me the reference to Qt 6 makes sense since the next major FC release will also be compilabe using Qt6.
Also Qt5_DIR and Qt6_DIR are already possible.

berniev:
Detailed explanation is in commit message

chennes:
Both of those are used by Qt's cMake files to detect the location of Qt's installation on the system: I set them regularly. I don't know what you mean that they are unused, they are part of Qt itself.

berniev:
> cMake looks for the cMake variable "Foo_DIR" and takes that as the location search for Foo.cMake.

I could be learning something here today (thanks!) so if I may ... Is this a basic feature of find_package? I can find no documentation.

luzpaz:
@donovaly  I've amended the OP for convenience.

donovaly:
> Detailed explanation is in commit message

Yes, but it is about to help the mergers. it helps a lot when on e can just open a PR and immediately see if one can review or merge it or not. When you look at other PRs all other PR submitter therefore add the info. Please do so as well. When I make a PR, the commit info is automatically added to the PR description field, so there is no extra work, but you help others to save some time.

chennes:
> Is this a basic feature of find_package? I can find no documentation.

Yes: there is brief mention here -- https://cmake.org/cmake/help/book/mastering-cmake/chapter/Finding%20Packages.html

> CMake creates a cache entry called <Package>_DIR to store the location found or allow the user to set it.

berniev:
@chennes 

Beautiful! Thank you. I have read that a dozen times and it didn't sink in. 

So to be clear <Package>_DIR is the location of the config file, yes?

I don't recall seeing this in FC documentation.

I'll close this PR now. Thanks again for indulging my ignorance.

chennes:
I think your understanding is correct (depending on what you mean by "config file"): when I go looking for what I need to set, e.g. "Qt6_DIR" too, I search for the file "Qt6.cMake".

berniev:
This is my understanding: find_package operates in one of two modes: module or config. In config mode cmake searches for either <lowercasePackageName>-config.cmake or <PackageName>Config.cmake
<Package>_DIR is set to reflect where a config is found, or if set in user env is prepended(?) to the search path. If not found there the search continues regardless.

Issue #8232, opened 2023-01-22
[Issue] IDE does not handle clang formatting correctly


WandererFan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

n/a

### Version

0.21 (Development)

### Full version info

```shell
n/a
```

### Subproject(s) affected?

Other (specify in description)

### Issue description

As reported here https://github.com/FreeCAD/FreeCAD/commit/2b7de8315088a4806f588b5647d724c93045bd93#comments the use of some clang-format options causes issues in Microsoft Visual Studio - specifically during cut and paste operations.

The cause of these issues should be investigated and resolved or work around developed.

Possible solutions:
- remove problematic options from the project level clang-format file
- change clang formatter version and/or MSVS options to support problematic options

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
Related to #7858 ?

WandererFan:
see #7858 and 2b7de83150

WandererFan:
Forum user @abdullah reports that KDEvelop also struggles with clang-format.
" It has the option to set the clang formater (and I have configured it so), but it does not appear to work. "

chennes:
I believe we mean @abdullahtahiriyo :)

abdullahtahiriyo:
Yes, indeed. It is is a pity that Kdevelop is not applying the clang format file.

It apparently is known:
https://stackoverflow.com/questions/59595816/how-to-set-up-code-formatting-in-kdevelop-with-clang-format

I have even tried that hacky approach without success...

WandererFan:
I haven't noticed any issues in QtCreator when "format on save" is in effect.

I wonder if we should run clang-format as part of CI instead of asking contributors to format before submission.   That assumes that running clang-format as a stand-alone would avoid the issues with MSVS and KDevelop.

berniev:
> I wonder if we should run clang-format as part of CI 

At the risk of repeating myself, that is, and always has been, my stated aim. Right now is a testing phase. Its a bit tricky in that you can't just format all the whole project's files and wait for a reaction as that messes up git history. It needs people to try for themselves.

Once in CI it matters not what individuals do in their local environment.

All that remains is to get concensus on style. The current version is looking pretty close (after adding back braces).

If you can't use clang-format in your IDE, or don't use an IDE it is easy to run from the command line.

WandererFan:
[...]

PR #8259, opened 2023-01-25
Cleanup issues of PR 8228

wwmayer:

---

### DIFF
--- a/src/Gui/ExpressionCompleter.cpp
+++ b/src/Gui/ExpressionCompleter.cpp
@@ -253,7 +253,7 @@ public:
         return result;
     }
 
-   // The completion tree structure created takes into account the current document and object
+    // The completion tree structure created takes into account the current document and object
     //
     // It is done as such:
     // * root (-1,-1)
@@ -284,8 +284,9 @@ public:
         App::Property* prop = nullptr;
         // check if the document is uniquely identified: either the correct index in info.doc
         // OR if, the node is a descendant of the root, its row lands within 0...docsize
-        if(idx>=0 && idx<docSize)
+        if(idx>=0 && idx<docSize) {
             doc = docs[idx/2];
+        }
         else {
             // if we're looking at the ROOT, or the row identifies one of the other ROOT elements
             // |----- current documents' objects, rows: docs.size             ... docs.size + objs.size
@@ -337,16 +338,16 @@ public:
         // the item is the ROOT or a CHILD of the root
         if(info.doc<0) {
             // and we're asking for a count, compute it
-            if(count)
+            if(count) {
                 // note that if we're dealing with a valid DOC node (row>0, ROOT_info)
                 // objSize and propSize will be zero because of the early exit above
                 *count = docSize + objSize + propSize;
+            }
             if(idx>=0 && v) {
                 // we're asking for this child's data, and IT's NOT THE ROOT
                 QString res;
                 // we resolved the property
-                if (propName)
-                {
+                if (propName) {
                     res = QString::fromLatin1(propName);
                     // resolve the property
                     if (sep && !noProperty && retrieveSubPaths(prop).size() != 0)
@@ -360,7 +361,8 @@ public:
                         res = QString::fromLatin1(obj->getNameInDocument());
                     if(sep && !noProperty)
                         res += QLatin1Char('.');
-                } else {
+                }
+                else {
                     // the document has been resolved, use the saved idx to figure out quotation or not.
                     if(idx & 1)
                         res = QString::fromUtf8(quote(doc->Label.getStrValue()).c_str());
@@ -425,11 +427,11 @@ public:
                 }
                 if (v) {
                     QString res = QString::fromLatin1(propName);
-                   
+
                     // check to see if we have accessible paths from this prop name?
                     if (sep && retrieveSubPaths(prop).size() != 0)
                         res += QLatin1Char('.');
-                    
+
                     *v = res;
                 }
                 return;
--- a/src/Gui/ExpressionCompleter.cpp
+++ b/src/Gui/ExpressionCompleter.cpp
@@ -450,7 +450,7 @@ public:
             }
             
             // check to see if this is a valid path
-            if (idx < 0 || idx >= paths.size()) {
+            if (idx < 0 || idx >= static_cast<int>(paths.size())) {
                 return;
             }
 
@@ -523,12 +523,13 @@ public:
             if (parentInfo.doc < 0) {
                 // need special casing to properly identify this model's object
                 const auto& docs = App::GetApplication().getDocuments();
+                auto docsSize = static_cast<int>(docs.size()*2);
 
                 info.doc = element.row();
 
                 // if my element is a contextual descendant of root (current doc object list, current object prop list)
                 // mark it as such
-                if (element.row() >= docs.size()*2) {
+                if (element.row() >= docsSize) {
                     info.contextualHierarchy = 1;
                 }
             } else if (parentInfo.contextualHierarchy) { 
@@ -537,9 +538,9 @@ public:
                 auto cdoc = App::GetApplication().getDocument(currentDoc.c_str());
 
                 if (cdoc) {
-                    auto objsSize = cdoc->getObjects().size()*2;
+                    auto objsSize = static_cast<int>(cdoc->getObjects().size()*2);
                     int idx = parentInfo.doc - docs.size();
-                    if (idx < cdoc->getObjects().size()*2) {
+                    if (idx < objsSize) {
                         //  |-- Parent (OBJECT)        - (row 4, [-1,-1,-1,0]) = encode as element => [parent.row,-1,-1,1]
                         //      |- element (PROP)            - (row 0, [parent.row,-1,-1,1])  = encode as element => [parent.row,-1,parent.row,1]
 


src/Gui/ExpressionCompleter.cpp
```diff
@@ -284,8 +284,9 @@ class ExpressionCompleterModel: public QAbstractItemModel {
         App::Property* prop = nullptr;
         // check if the document is uniquely identified: either the correct index in info.doc
         // OR if, the node is a descendant of the root, its row lands within 0...docsize
-        if(idx>=0 && idx<docSize)
+        if(idx>=0 && idx<docSize) {
             doc = docs[idx/2];
+        }
```
donovaly, line 289: Just for curiosity, why are curly braces necessary for single-line ifs? Up to now i did not add them because I saw no need and also because most of FC's code don't have them.

chennes: Forums discussion of clang-format: https://forum.freecadweb.org/viewtopic.php?f=10&t=72183&p=629874

The gist of it is that it's widely considered a best practice due to some very famous bugs that would have been prevented had this formatting strategy been used.

donovaly: Thanks for the pointer. In this discussion I miss the info what can be harmful of omitting the braces. I don't mean when there are comments and then a single statement, but just a single statement alone, so
```
if (blah)
 do something;
else if (blub)
 do another thing;
...
```

Especially when having "else if" this way improves in my opinion the readability, especially on smaller screen since it saves lines.

Besides this, I recently reverted the automatic addition of braces, not because I am opposed of requiring the braces, but it was a nightmare to work with the VC IDE with this. At first, the addition seemed convenient but on longer coding sessions VC is not able to do this right when copy/pasting code around. Often it adds a closing brace despite it is already there. This cost a lot of time since the compilation error messages don't help. So one has to carefully re-read to find additions braces. On the other hand sometimes it does not add the closing brace. Still a mystery for me. What about you, are you on VC 2022 as well? When you enabled automatic brace insertion, and code around for a while, do you encounter problems? I had a look today in forums but cannot find a way to use a clang file and only overwrite one of its setting.
I see no workaround because I cannot turn of the formatter because then the code would not be formatted at all. And having a custom clang file only for me is error-prone (different PCs, keeping it up to date.)

berniev:
Interesting to see the experimental clang-format spoken of like it is official. Nice!

Just a couple of cautions though: 

It was previously pointed out that broadscale reformatting should be avoided for now to avoid further complicating the pending toponaming merge.

Format will eventually (I believe) be included in CI. After that local issues in the IDE become irrelevant. Do whatever you want! The point is that local issues are temporary and should not block achieving the best final outcome.

berniev:
>  This cost a lot of time since the compilation error messages don't help

I dont use VC but CLion and VSCode both instantly alert mismatched braces when I mess up. Is VC2022 is not similar? Odd  these get through to compile.

Maybe: [How do I specify a clang-format file](https://stackoverflow.com/questions/46373858/how-do-i-specify-a-clang-format-file)

donovaly:
> Interesting to see the experimental clang-format spoken of like it is official. Nice!

The point is that it is not experimental. Since VC 2019 the clang file of a project is the default formatter. You cannot disable this, except of
- turning of the formatter completely -> then your code is not formatted at all. Another user had the same problems as I and turned off the formatter completely, see https://github.com/FreeCAD/FreeCAD/pull/8262#discussion_r1087270242 and the result is that his code has no uniform formatting
- using your own clang file -> I tried this as well but sine I am the stable release manager I have different PCs now (currently 4). It consumes time to maintain its own clang file and to keep it in sync on all PCs. Moreover, it would contradict the principle that everybody out using VC should have its own clang file just to be able to be productive.

berniev:
At the very least you could identify your problem and maybe others have something to solve. Give some example etc. So far all we've heard is you don't like what happens.

donovaly:
> all we've heard is you don't like what happens.

Where did you hear this? I even approved the change to add auto brace insertion.

It took me a lot of time to realize what is going on, why I got frequent and hard to understand compilation errors. Thus I reverted the particular change to be productive again and wrote about this in the commit. I know, this was not according to our rules. But following the rules would have meant that I would not be productive for days, have to start a debate, bring examples etc. And as you see others were affected too. Thus it was quicker to "just resolve" the issue.
I was hardly criticized for my action. I can understand this. However, it is our spare time and when a simple change that only aims to add convenience for developers turn out to introduce troubles, one should quickly act. And here no code was actually changed.

berniev:
> It took me a lot of time to realize what is going on, why I got frequent and hard to understand compilation errors

You need to seriously revise your workflow. If you are using a moden IDE there is no way a format problem should remain undiscovered until compilation.

donovaly:
> You need to seriously revise your workflow. If you are using a moden IDE there is no way a format problem should remain undiscovered until compilation.

So Visual Studio 2022 is not modern? Exactly because I knew that I will get statements like this:
it is about you, you need to check...
I made the change, became productive again and others too. Yes, apparently VC 2022 has a bug, but why should all its users suffer for now?

src/Gui/ExpressionCompleter.cpp
adrianinsaval: from what you are saying it sounds like it's messing up the file _while_ you are coding, that does sound annoying and error prone, can't you run clang format just before committing?

adrianinsaval: > Thanks for the pointer. In this discussion I miss the info what can be harmful of omitting the braces

my first guess is that later someone might forget to put braces when trying to add more lines inside the if statement, I've certainly done so myself 🤦‍♂️

wwmayer:
> Just for curiosity, why are curly braces necessary for single-line ifs?

If you have some if-else if checks on the same level and any of it requires braces then by most style guidelines it's strongly recommended to add curly braces to all of them.

Example:
```
if (check1)
   do_something_1;
else {
   do_something_2;
   do_something_3;
}
```
should become
```
if (check1) {
   do_something_1;
}
else {
   do_something_2;
   do_something_3;
}
```

If you have an if/else if/else with a comment before the actual statement then it's strongly recommended to add curly braces.
```
if (check1)
    // this is a comment
   do_something;
```
should become
```
if (check1) {
    // this is a comment
   do_something;
}
```

If you have nested if-checks then for the outer if's curly braces are strongly recommended.
```
if (check1)
    if (check2)
       do_something;
```
should become
```
if (check1) {
    if (check2)
        do_something;
}
```
or (preferred)
```
if (check1) {
    if (check2) {
        do_something;
    }
}
```

berniev's rule is even simpler because he recommends to always use curly braces. Unfortunately a lot style guidelines say something else where for single line statement the curly braces should be dropped.

donovaly:
> If you have some if-else if checks on the same level and any of it requires braces then by most style guidelines it's strongly recommended to add curly braces to all of them....

Many thanks. What you describe is how most of FC's code is written and also how I try to code.
Therefore I don't see why this must be changed.

The problems with changing the rules are:
- When copying code around braces will be added also single-line statements like
```
if (bla)
    return something;
```

This gives you a lot of changes in PRs that have nothing to do with the actual intended code change, making it harder for the reviewers or you have to undo the automatical additions to keep the diffs small. Both ways consume spare time either of the coder or the reviewer.

- on VC 2022 sometimes only the opening brace is added, but sometimes also a closing one even when there is already a closing one. This is apparently not only annoying for me but for other coders too.

All in all, the whole issue consumes a lot of our spare time, for the discussion and also to handle the code. Therefore I don't see why we need to change or style when it gives us no benefit, but additional work. We should better use our time to focus to write good code (good to understand for others, less memory consumption, fast etc.)

src/Gui/ExpressionCompleter.cpp
chennes: In my opinion, format-as-you-type is more trouble than it is worth, in _any_ IDE. My strategy is to literally not worry about the format at all while I am coding, and just make the code look how I like it to look. Then, before committing, I select and auto-format (or in the case of Python, use Black as a pre-commit hook).

wwmayer:
> on VC 2022 sometimes only the opening brace is added, but sometimes also a closing one even when there is already a closing one. This is apparently not only annoying for me but for other coders too.

I don't use it a lot any more but sounds like the IDE is not very smart. Maybe you should open a new thread in the forum to discuss this issue.

berniev:
Re single line statements, a slight clarification:
```
if(bla) doThis();
```
and
```
if(bla)
    doThis();
```
Should the first version be allowed? I would argue it should not be allowed as it hides content and makes debugging more difficult. If it is allowed then perhaps braces are superfluous.

The second case it seems has concensus for always brace.

berniev:
I tried Visual Studio 2022 for Mac, but deleted it. No support for C++. I think VS priority was always C#.

chennes:
Does CLion do a good job with autoformatting? I'm getting ready to migrate (I really dislike using C++ in Xcode).

berniev:
CLion is, like other Jetbrains products I have used for years, very slick. Given the huge hours we spend at the screen, a hundred bucks or whatever is peanuts. I continue to be amazed how good they are. **BUT** CLion has some indexing/speed issues with FreeCAD - I'm talking sometimes wait 45 minutes with all cores flat out. This renders much of the refactoring assistance (one of my favourite features) unusable. I spent heaps on extra RAM - made no difference. I have several reports in. Maybe on your M1 you'll do better.
Formatting is not intrusive, works great.

WandererFan:
related: #8232

berniev:
@chennes clarification re CLion: The formatting is not entirely 'auto', more 'helpful'. Still hit format button from time to time.

donovaly:
> Should the first version be allowed?

In my opinion not. As I also argued in the forum, I don't see the need to change the formatting when > 95% of your code use already the linebreak.

> The second case it seems has concensus for always brace.

Where do you see this?
Personally I am against this since it brings us no benefit. I vote to keep the status we already more ore less have and this is that: https://github.com/FreeCAD/FreeCAD/pull/8259#issuecomment-1405018752

src/Gui/ExpressionCompleter.cpp
WandererFan: "In this discussion I miss the info what can be harmful of omitting the braces."

The most famous case is the "Apple goto fail".   There is also the python mantra "explicit is better than implicit".   From a maintainability standpoint, anything that reduces the chances of misinterpreting code is a good thing.

PR #8266, opened 2023-01-26
[Gui] allow to specify font for NaviCube

### LINKED ISSUE #8082: Navigation Cube - breaks upon larger size.

### Forums discussion

_No response_

### Version

0.20 (Release)

### Full version info

```shell
OS: Windows 10 Version 2004
Word size of FreeCAD: 64-bit
Version: 0.20.2.29177 +426 (Git)
Build type: Release
Branch: (HEAD detached from 0.20.2)
Installed mods: 
  * 3DfindIT 1.2.0
  * 3D_Printing_Tools
  * A2plus 0.4.60g
  * Assembly4 1.20.6
  * boltsfc 2022.11.5
  * CfdOF 1.20.6
  * CurvedShapes 1.0.4
  * Curves 0.12.5
  * Defeaturing 1.2.0
  * Design456 0.0.1
  * DynamicData 2.46.0
  * EM
  * Estimate 0.1.2
  * ExplodedAssembly
  * fasteners 0.4.27
  * fcgear 1.0.0
  * FEMbyGEN 1.2.0
  * FEM_FrontISTR 0.1.0
  * freecad.gears 1.0.0
  * Help 1.0.3
  * InventorLoader 1.3.0
  * lattice2 1.0.0
  * Launcher
  * Manipulator 1.4.9
  * offline-documentation 1.0.0-alpha
  * OpticsWorkbench 1.0.9
  * parts_library
  * pcb
  * PieMenu 1.2.4
  * POV-Ray-Rendering
  * QuickMeasure 2022.10.28
  * Silk 1.0.0
  * slic3r-tools
  * SlopedPlanesMacro
  * symbols_library
  * ThreadProfile 1.84.0
  * toSketch 1.0.1
```

### Subproject(s) affected?

None

### Issue description

Edit > Preferences > Display > Navigation.
Cube size: 132  <default>
Cube size: 142 <breaks>

Navigation Cube Breaks on enlarging size for the reverse spot in upper right.

If you adjust size of cube up from default the reverse side button won't work.

 Example: Default size = 132; Change size = 142.  Cube will not work when the reverse spot is selected.

### Anything else?

_No response_

### Code of Conduct

donovaly:
- the problem is that depending on the OS, the font of the NaviCube is hardly readable. For example under Windows 11, there is no Helvetica font and therefore an ugly replacement font is used. On some laptop screens the font is too large or too small etc.
- as solution this PR add the change the NaviCube's font and font size

- the PR also fixes an issue that the position (corner) of the NaviCube was not respected when the NaviCube is recreated. This fix is necessary for the PR therefore included

- as by-product the PR fixes #8082 since every change in the preferences now properly recreates the NaviCube

- the PR also removes the strange and unused class "HuuhaaClassPy"

Here is the new NaviCube settings implemented by this PR:
FreeCAD_g5i4BUqQct

### DIFF
--- a/src/Gui/NaviCube.h
+++ b/src/Gui/NaviCube.h
@@ -41,22 +41,18 @@ public:
         BottomLeftCorner,
         BottomRightCorner
     };
-    NaviCube(Gui::View3DInventorViewer* viewer) ;
+    NaviCube(Gui::View3DInventorViewer* viewer);
     virtual ~NaviCube();
     void drawNaviCube();
     void createContextMenu(const std::vector<std::string>& cmd);
     bool processSoEvent(const SoEvent* ev);
     void setCorner(Corner);
+    static QFont getDefaultSansserifFont();
     static void setNaviCubeCommands(const std::vector<std::string>& cmd);
     static void setNaviCubeLabels(const std::vector<std::string>& labels);
+
 private:
     NaviCubeImplementation* m_NaviCubeImplementation;
-};
-
-class HuuhaaClassPy : public Py::PythonExtension<HuuhaaClassPy> {
-public:
-    Py::Object huuhaa(const Py::Tuple&);
-    static void init_type() ;
 };
 
 #endif /* SRC_GUI_NAVICUBE_H_ */
--- a/src/Gui/DlgSettingsNavigation.cpp
+++ b/src/Gui/DlgSettingsNavigation.cpp
@@ -36,6 +36,7 @@
 #include "DlgSettingsNavigation.h"
 #include "ui_DlgSettingsNavigation.h"
 #include "MainWindow.h"
+#include "NaviCube.h"
 #include "NavigationStyle.h"
 #include "View3DInventor.h"
 #include "View3DInventorViewer.h"
@@ -72,7 +73,8 @@ void DlgSettingsNavigation::saveSettings()
     // where we set some attributes afterwards
     ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath
         ("User parameter:BaseApp/Preferences/View");
-    QVariant data = ui->comboNavigationStyle->itemData(ui->comboNavigationStyle->currentIndex(), Qt::UserRole);
+    QVariant data = ui->comboNavigationStyle->itemData(ui->comboNavigationStyle->currentIndex(),
+        Qt::UserRole);
     hGrp->SetASCII("NavigationStyle", (const char*)data.toByteArray());
 
     int index = ui->comboOrbitStyle->currentIndex();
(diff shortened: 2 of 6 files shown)


src/Gui/NaviCube.h
```diff
@@ -41,22 +41,18 @@ class GuiExport NaviCube {
...
     void setCorner(Corner);
+    static QFont getDefaultSansserifFont();
     static void setNaviCubeCommands(const std::vector<std::string>& cmd);
     static void setNaviCubeLabels(const std::vector<std::string>& labels);
+
 private:
     NaviCubeImplementation* m_NaviCubeImplementation;
 };
 
-class HuuhaaClassPy : public Py::PythonExtension<HuuhaaClassPy> {
```
adrianinsaval, line 56: what was this?? @nyholku

nyholku: I wondered how long it would take for someone to notice ;) It was an experiment that I forgot to remove from the code, it does nothing and should not be there.

nyholku: 'huuhaa' is a Finnish expression that could be translate to approximately 'hogwash'.

adrianinsaval: it took 5 years! hahaha

adrianinsaval:
I confirmed this fixed the bug with the navicube size, it seems it changed something (no clue what) on how the navicube font setting is interpreted though, after configuring the font and fontsize to my satisfaction on this PR's version:
image

opening back on master or in 0.20.2 resulted in undesired results:
image

since settings are shared across versions it would be ideal to keep them compatible

donovaly:
> I confirmed this fixed the bug with the navicube size, 

Thanks for testing. The bug occurred because the cube was not repainted. it was only made bigger and then the arrows bere covered by a transparent layer.

> opening back on master or in 0.20.2 resulted in undesired results: image
> 
> since settings are shared across versions it would be ideal to keep them compatible

This is impossible. In FC 0.20.x only a single font is used in a specific size which cannot be changed. That is the reason why I acted because the font selection mechanism fail with Windows 11. Since we have the policy not to backport features, only bugfixes, this PR cannot be backported.

adrianinsaval:
yeah I tried again and on 0.20.2 it looks like that with default settings, I guess I had fixed that using cube menu that's why I expected it to look good so all good with the PR I think

IMO even after merging this PR we should try to find better defaults for the cube, like a bigger default size according to screen size and resolution and a font size such that the text fits inside the cube by default

donovaly:
> so all good with the PR I think

Thanks. So I can merge.

> IMO even after merging this PR we should try to find better defaults for the cube, like a bigger default size according to screen size and resolution and a font size such that the text fits inside the cube by default

You did already great work with the cube, so please move on ;-)

PR #8269, opened 2023-01-26
Part: fix crash in SectionCutting dialog

wwmayer:
It segfaults when executing the script below:
```
doc = App.ActiveDocument
box = doc.addObject("Part::Box", "SectionCutBoxX")
comp = doc.addObject("Part::Compound", "SectionCutCompound")
comp.Links = box
grp = doc.addObject("App::DocumentObjectGroup", "SectionCutX")
grp.addObject(comp)
doc.recompute()
```


chennes:
Thanks for the fix -- is this something we could add a regression test for in the new C++ testing framework?

wwmayer:
#8286

PR #8270, opened 2023-01-26
Fem: move from string based connect() to overloaded variant based on …

wwmayer:
…meta methods

---

### DIFF
--- a/src/Mod/Fem/Gui/TaskPostBoxes.cpp
+++ b/src/Mod/Fem/Gui/TaskPostBoxes.cpp
@@ -32,6 +32,7 @@
 
 # include <QApplication>
 # include <QMessageBox>
+# include <QMetaMethod>
 # include <QToolTip>
 #endif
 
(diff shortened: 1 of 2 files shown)


src/Mod/Fem/Gui/TaskPostBoxes.cpp
```diff
@@ -241,8 +242,9 @@ void TaskDlgPost::connectSlots()
 {
     // Connect emitAddedFunction() with slotAddedFunction()
     QObject* sender = nullptr;
+    int indexSignal = 0;
```
chennes, line 245: Why this move?

wwmayer: The code has this structure:
```
// search for signal
for () {
...
}

if (signal) {
    // search for slot
    for () {
...
        if (slot found) {
            // use indexSignal and indexSlot
...
         }
    }
}
```
That's why indexSignal cannot be declared inside the first loop any more but must be declared outside. Since the connection is done inside the second loop indexSlot can be declared inside that loop.

An alternative is to use nested for loops but this makes the code less readable and harder to understand.

chennes: Ah! I have to stop reviewing code on GitHub, the way the code is collapsed I didn't even see that the first loop was separate. Doh! Thanks for the clarification.

PR #8277, opened 2023-01-27
[Gui] NaviCube: improve default font

donovaly:
- it turned out that DejaVu Sans is not the best font since its letters are wide. Bahnschrift was designed to run narrow and it at least on all Win 10 since 2017. Therefore use this, if possible

### DIFF
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -270,13 +270,18 @@ QFont NaviCube::getDefaultSansserifFont()
     QFont font(QString::fromLatin1("Helvetica"));
     if (font.styleHint() & QFont::SansSerif)
         return font;
-    // on Windows 11 there is no longer a Helvetia font
+    // on Windows 10 or newer there is no Helvetia font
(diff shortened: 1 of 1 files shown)


adrianinsaval:
There are other places in code where we directly use helvetica too
https://github.com/FreeCAD/FreeCAD/search?p=1&q=helvetica
would it be worth making a global function to get a default font and use that in all these? Although some of the uses are in coin so this exact method may not work.

In some places the font family is specified like this: `Arial, Helvetica, sans-serif` does qt not support this?
In some places it's specified like this though: `Arial, Helvetica, sans` which I think it's wrong as it would be a different style of font when arial and helvetica are not available.

donovaly:
> In some places the font family is specified like this: `Arial, Helvetica, sans-serif` does qt not support this?

I don't know.

> Another tangent, as I understand it the text in the navicube is now done with Qt as a texture,

It has always been done like this. The only difference was that a specific font and size was hardcoded. This hardcoding fails on Win 11 and that's why I acted.

PR #8280, opened 2023-01-27
[Gui] register NaviCube parameters to view

donovaly:
- there is a bug: when deleting any NaviCube parameter in the parameters dialog, the background color suddenly changes the fix is to register the NaviCube parameters to the view

- the PR also fixes the issue that the default font size was not properly calculated but hardcoded

### DIFF
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -112,6 +112,7 @@ class NaviCubeImplementation : public ParameterGrp::ObserverType {
 public:
     explicit NaviCubeImplementation(Gui::View3DInventorViewer*);
     ~NaviCubeImplementation() override;
+    int getDefaultFontSize();
     void drawNaviCube();
     void createContextMenu(const std::vector<std::string>& cmd);
 
@@ -214,8 +215,10 @@ public:
     bool m_MightDrag = false;
     double m_BorderWidth;
     NaviCube::Corner m_Corner = NaviCube::TopRightCorner;
+    bool m_RotateToNearest = true;
+    int m_NaviStepByTurn = 8;
     int m_CubeTextSize = 0;
-    std::string m_CubeTextString;
+    std::string m_CubeTextFont;
 
     QtGLFramebufferObject* m_PickingFramebuffer;
 
@@ -263,6 +266,31 @@ void NaviCube::setCorner(Corner c) {
     m_NaviCubeImplementation->m_PrevHeight = 0;
 }
 
+void NaviCube::setSize(int size)
+{
+    m_NaviCubeImplementation->m_CubeWidgetSize = size;
+}
+
+void NaviCube::setNaviRotateToNearest(bool toNearest)
+{
+    m_NaviCubeImplementation->m_RotateToNearest = toNearest;
+}
+
+void NaviCube::setNaviStepByTurn(int steps)
+{
+    m_NaviCubeImplementation->m_NaviStepByTurn = steps;
+}
+
+void NaviCube::setFont(std::string font)
+{
+    m_NaviCubeImplementation->m_CubeTextFont = font;
+}
+
+void NaviCube::setFontSize(int size)
+{
+    m_NaviCubeImplementation->m_CubeTextSize = size;
+}
+
 // sets a default sansserif font
 // the Helvetica font is a good start for most OSes
 QFont NaviCube::getDefaultSansserifFont()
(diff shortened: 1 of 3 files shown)


donovaly:
@wwmayer , please have a look and merge if OK

src/Gui/NaviCube.cpp
```diff
@@ -214,6 +215,8 @@ class NaviCubeImplementation : public ParameterGrp::ObserverType {
     bool m_MightDrag = false;
     double m_BorderWidth;
     NaviCube::Corner m_Corner = NaviCube::TopRightCorner;
+    bool m_RotateToNearest = true;
```
wwmayer, line 218: Where is this member ever used? There are only methods to set the value but nowhere in the code the value is used.

src/Gui/NaviCube.cpp
```diff
@@ -214,6 +215,8 @@ class NaviCubeImplementation : public ParameterGrp::ObserverType {
...
+    int m_NaviStepByTurn = 8;
```
wwmayer, line 219: Same here.

src/Gui/NaviCube.cpp
```diff
@@ -263,6 +266,31 @@ void NaviCube::setCorner(Corner c) {
     m_NaviCubeImplementation->m_PrevHeight = 0;
 }
 
+void NaviCube::setSize(int size)
+{
+    m_NaviCubeImplementation->m_CubeWidgetSize = size;
```
wwmayer, line 271: Having a setter for this property is cumbersome. On the one hand the method can be used to set a value but on the other hand the NaviCubeImplementation retrieves the value directly from the parameter group. Since now we have mixed responsibility it is not clear who sets values which will cause some undesired side-effects.

IMO, the NaviCubeImplementation class shouldn't access the parameter group directly but provide member variables and the corresponding setter methods. It will be up to the View3DInventor class to call the setter methods.

src/Gui/View3DInventorViewer.cpp
```diff
@@ -1171,6 +1171,36 @@ void View3DInventorViewer::setNaviCubeCorner(int c)
         naviCube->setCorner(static_cast<NaviCube::Corner>(c));
 }
 
+void View3DInventorViewer::setNaviCubeSize(int size)
```
wwmayer, line 1174: All these methods are not really needed. The View3DInventor class can directly access the NaviCube and then invoke the corresponding methods directly.

src/Gui/NaviCube.cpp
donovaly: Now it is used.

src/Gui/NaviCube.cpp
donovaly: Now it is used.

src/Gui/NaviCube.cpp
donovaly: OK. I improved the situation now a bit. I will not go further in the PR since this PR is meant to fix the missing handling of the NaviCube parameters.

I have not written the NaviCube feature. It was obviously not well maintained. In my previous merged PR, I deleted for example a year old class that was never used. In the NaviCube code i see further things like this: FontStretch, OffsetX settings etc. IMO these can be removed since they are nowhere set, except an expert user sets them.

src/Gui/View3DInventorViewer.cpp
donovaly: Thanks, now fixed.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8280">pipeline status</a> for feature branch [PR_8280](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8280). Pipeline [761011213](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/761011213) was triggered at [0576b9b](https://github.com/FreeCAD/FreeCAD/pull/8280/commits/0576b9bcfa3b653cafc8e72bf92631776b3f1388). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@wwmayer , can this now go in?

donovaly:
OK, I am going to merge this since it fixes the issue it should fix and there is an open discussion about the NaviCube running and people should be able to test the new settings possibility via the next weekly.
The mentioned necessary cube recreation definitely fixes #8082 as reported today.

I am of course open for any code refactoring, despite I am here not the author of the cube feature as such.

PR #8292, opened 2023-01-29
Further enhancements to Final Report

berniev:
Further efficiencies to Summary Report

Changed existing function to macro to access list of lines
Added conditional macro to save most of the if() else() endif()
Output sorted for improved comprehension
Moved build message out (nothing to do with report)
SLOC reduced ~30%

---


cMake/FreeCAD_Helpers/PrintFinalReport.cmake
```diff
@@ -1,182 +1,128 @@
...
-        report(PySideTools "not found")
-    endif()
-
+        simple(QtWebKitWidgets "not needed")
+    endif()
+    conditional(Shiboken SHIBOKEN_FOUND "not found" "${SHIBOKEN_VERSION} Base: [${SHIBOKEN_BASEDIR}]")
+    conditional(PySide PYSIDE_FOUND "not found" "${PYSIDE_VERSION} [${PYSIDE_INCLUDE_DIR}]")
+    conditional(PySideTools PYSIDE_TOOLS_FOUND
+                "not found"
+                "uic: [${PYSIDE_UIC_EXECUTABLE}] rcc: [${PYSIDE_RCC_EXECUTABLE}]"
```
chennes, line 85: This line exposes a problem in the code that's finding PySideTools. On Windows, the backslash character is retained in these two executable paths: until now that was harmless (and IMO actually sort of useful). However, that is breaking your macro above, because it gets interpreted as an escape character. All other Windows paths are being  modified to use a (forward) slash as a separator, which I believe is the standard way of doing it in cMake. In order to merge the above change, that needs to be addressed first.

cMake/FreeCAD_Helpers/PrintFinalReport.cmake
```diff
@@ -1,182 +1,128 @@
...
     endif()
+    conditional(Tests BUILD_TEST "OFF" "ON")
+
+    ############## End report data ####################
+
+    message(STATUS "\n==============\n"
+                   "Summary Report\n"
+                   "==============\n")
+
+    list(SORT simpleLines CASE INSENSITIVE)
```
chennes, line 123: I suspect this is going to be unpopular -- at the very least we should have a separate section at the top that lists the most important things first, and together. That is, the four compiler settings, and the version of Python. You might be better off manually sorting above and dropping this line.

chennes (changes requested):
Overall a good change: two comments below.

cMake/FreeCAD_Helpers/PrintFinalReport.cmake
chennes: I'd appreciate a double-check on #8293, which will correct this.

berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8292">pipeline status</a> for feature branch [PR_8292](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8292). Pipeline [761011215](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/761011215) was triggered at [276fced](https://github.com/FreeCAD/FreeCAD/pull/8292/commits/276fcedee422f32adecf2ae5a1652f8df424b789). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

cMake/FreeCAD_Helpers/PrintFinalReport.cmake
berniev: Go for it! I commented in your pr.

cMake/FreeCAD_Helpers/PrintFinalReport.cmake
berniev: Respectfully, I ask that we leave sorted for now as is. Its easy to change if real problems arise. 

I have been using this report quite a bit lately during various experiments and unsorted was driving me nuts. In the best traditions of 'data' I scratched the itch and sorted the result. Much better!

In unordered data it makes sense to have the most used at the top, but that is not to say unordered was the right choice in the first place. We are educated to expect sorting. The only important item to the reader is the one being sought, which sorting facilitates (ha ha).

Special cases almost always hint at underlying issues and are the first step to chaos. Every item is important. Being 'special' is subjective and circumstantial. E.g. during initial setup the python version might be of special interest, but after that not.

There are some alternatives:

The names of 'special' cases could simply be prefaced with e.g. '_' to move them up, or could be moved up by other means, but this is a slippery slope. It frustrates the reader by requiring recognition that there are two different systems in place. The first few items is where the brain automatically and subconsiously looks to determine if sorted.

Sort personal choice could be a cache flag. Seems OTT.

chennes: I certainly agree that the libraries should be sorted alphabetically -- to me it's not so much that any one item is "special" as that this is really two _different_ lists that by virtue of our old formatting _appeared_ to be one. I view this as list of compiler settings, and a list of library versions. 

That said, it is of course possible that simply sorting the whole thing and then being more careful with how we label things (e.g. "Flags" should really be "Compiler Flags"), will still achieve the desired effect. Let me ping a few other Maintainers and see if anyone else has an opinion. If not, I'll merge it as-is.

chennes: Well, seems like no one else cares (yet 😉). I'm going to merge as-is, with the caveat that I'd still like to find a way to keep the actual compiler settings close to one another.

berniev:
Thanks Chris!

> keep the actual compiler settings close to one another

I'm realising that the summary is all about `results`, but just as important is `cause`.  Mainly this is settings, and then mainly option settings. I'm thinking they should (if not silly quantity) also have a place in the report page.

berniev:
I am listening!

With a small tweak this is possible:
```
==============
System
==============

-- Compiler:            /usr/local/opt/llvm/bin/clang++ (15.0.7)
-- cmake:               3.25.1
-- prefix:              /Users/bernie/FreeCAD.app/Contents
-- bindir:              bin
-- datadir:             share
-- docdir:              share/doc/FreeCAD
-- includedir:          include
-- libdir:              /Users/bernie/FreeCAD.app/Contents/lib
-- 

==============
Config
==============

-- Flags:               -Wall -Wextra -Wpedantic -Wno-write-strings  -Wno-undefined-var-template
-- BuildType:           Release
-- Tests:               ON
-- FREECAD_USE_FREETYPE FREECAD_USE_FREETYPE
-- BUILD_VR:            -undefined-
-- 

==============
Summary
==============

-- Boost:               1.81.0
-- Coin3D:              4.0.0 [/usr/local/lib/libCoin.dylib] [/usr/local/include]
-- Coin3D_DOC:          not found
-- DesignerPlugin:      not built (BUILD_DESIGNER_PLUGIN is OFF)
...
```
If you like the idea let me know what you want where and I'll PR

chennes:
I think that looks like a good suggestion.

PR #8293, opened 2023-01-29
cMake: Correct PySideTools paths on Windows

chennes:
I had been using "TO_NATIVE_PATH" to get nice "copy-and-paste"-able Windows paths for these, but that is not the cMake Way, and causes problems later. Switch back to allowing cMake to use / as the path separator, even on Windows.

---


berndhahnebach:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8293">pipeline status</a> for feature branch [PR_8293](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8293). Pipeline [761011216](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/761011216) was triggered at [c4fbbac](https://github.com/FreeCAD/FreeCAD/pull/8293/commits/c4fbbaca8bf9d6de53e2ffa81514125654505dc9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berniev:
Oops! Yes, good idea, if everything else still works.

fyi:

I see 17 occurences of TO_NATIVE_PATH. Do some/all of these have to be looked at too? 
4 in SetupShiboken, SetupPySide uic,rcc
2 in SetupShibokenAndPySide (but this is dead code anyway)
11 others

> The sub-commands RELATIVE_PATH, TO_CMAKE_PATH and TO_NATIVE_PATH has been superseded.

chennes:
This is the minimal fix to get #8292 working, but it certainly wouldn't be a bad idea to look through the other uses of TO_NATIVE_PATH, especially if cMake doesn't want it used at all. I assume this runs on your system without issue? I've only tested it on Windows, but I expect that to be the only system that it affected.

berniev:
I was unaware of an issue til you pointed it out. MacOS only.
But my system is a bit broken lately with Python/PySide/Shipoken experiments.

> TO_NATIVE_PATH, especially if cMake doesn't want it used at all
There is a replacement

chennes:
I'm going to go ahead and merge this: no doubt if it breaks somebody's install or workflow they will let me know 😄 .

PR #8304, opened 2023-01-30
[Part] Tweak mesh generation

pmjdebruijn:
Mesh generation parameters are taken from PrusaSlicer's STEP importer, which does a just-in-time STEP->STL conversion in memory.

Resulting meshes have better geometry, but are a little larger as a result.

---

### DIFF
--- a/src/Mod/Part/App/Tools.cpp
+++ b/src/Mod/Part/App/Tools.cpp
@@ -639,7 +639,7 @@ Handle (Poly_Triangulation) Part::Tools::triangulationOfFace(const TopoDS_Face& 
     TopoDS_Shape shape = mkBuilder.Shape();
     shape.Location(loc);
 
-    BRepMesh_IncrementalMesh(shape, 0.1);
+    BRepMesh_IncrementalMesh(shape, 0.005, false, 1, true);
     return BRep_Tool::Triangulation(TopoDS::Face(shape), loc);
 }
 
@@ -665,7 +665,7 @@ Handle(Poly_Polygon3D) Part::Tools::polygonOfEdge(const TopoDS_Edge& edge, TopLo
     TopLoc_Location inv = loc.Inverted();
     shape.Location(inv);
 
-    BRepMesh_IncrementalMesh(shape, 0.1);
+    BRepMesh_IncrementalMesh(shape, 0.005, false, 1, true);
     TopLoc_Location tmp;
     return BRep_Tool::Polygon3D(TopoDS::Edge(shape), tmp);
 }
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8304">pipeline status</a> for feature branch [PR_8304](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8304). Pipeline [761961360](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/761961360) was triggered at [8fc0fc4](https://github.com/FreeCAD/FreeCAD/pull/8304/commits/8fc0fc4b1848c0b11f09ecf465b3fc869f710a9c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
Instead of harcoding that value isn't it better to use a config parameter? We a default part deviation parameter, should that be used or is this parameter something else?

pmjdebruijn:
In my fairly simple designs I've seen an increase of roughly 5%

chennes:
Thanks -- in general I think that people who care a great deal about this are going to be using the Mesh workbench and specifying the parameters carefully themselves. To me, 5% is not a big deal. I'm going to leave this open a while longer in case anyone else wants to discuss it, but otherwise I'm inclined to merge as-is (with the caveat that if you're looking for a project, making this a "hidden" parameter would be a pretty easy modification).

pmjdebruijn:
I've also been experimenting with:

-    BRepMesh_IncrementalMesh(shape, 0.1);
+    BRepMesh_IncrementalMesh(shape, 0.005, false, 0.1, true);

This results in even better geometry in somecases, with a filesize increase of about 10% in my fairly simple models.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8304">pipeline status</a> for feature branch [PR_8304](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8304). Pipeline [763309957](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/763309957) was triggered at [6ff03ad](https://github.com/FreeCAD/FreeCAD/pull/8304/commits/6ff03ad143b1fb896d67c5c4e5ea522b943379b4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
This problem actually came up in the discussions at FreeCAD Day today -- so thanks for preemptively fixing it so I could say that a PR was in the works already! I'm going to merge this updated parameter set now.

donovaly:
> This results in even better geometry in somecases, with a filesize increase of about 10% in my fairly simple models.

But only in simple ones. I totally agree that the next PR should be to add a parameter that users can change the preset. Could you make such a PR?
If not, @adrianinsaval , are you interested? I think it could be a small and nice project for you.

PR #8330, opened 2023-02-02
Start: Added option to hide start page scrollbars

yorikvanhavre:
This adds an option to the Start preferences that allow to hide the scrollbars on the start page. They are drawn in a different style than the other FreeCAD scrollbars, so they are ugly :) Scrolling with the mousewheel still works.

---

### DIFF
--- a/src/Mod/Start/Gui/DlgStartPreferences.ui
+++ b/src/Mod/Start/Gui/DlgStartPreferences.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>482</width>
-    <height>795</height>
+    <width>537</width>
+    <height>942</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8330">pipeline status</a> for feature branch [PR_8330](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8330). Pipeline [765717417](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/765717417) was triggered at [5403802](https://github.com/FreeCAD/FreeCAD/pull/8330/commits/5403802f3e24f795309bbfafa310a69d21beef55). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Start/Gui/DlgStartPreferences.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>482</width>
-    <height>795</height>
+    <width>537</width>
+    <height>942</height>
```
donovaly, line 10: 150 px more height for an option seems wrong.

However in the application the height is recalculated anyway. So I will merge and have a look at the ui file later.

donovaly (commented):
Thanks
please don't forget to update the Wiki and the release notes

chennes:
This PR causes us to fail CI because "normal" is not a valid CSS value for the display property. @yorikvanhavre can you submit a PR to correct? See https://developer.mozilla.org/en-US/docs/Web/CSS/display for valid options.

src/Mod/Start/Gui/DlgStartPreferences.ui
yorikvanhavre: yes at some point that page will need to be splitted I guess. For the time being, I think it still think it looks best as a single, long list I guess

PR #8350, opened 2023-02-04
[NotificationArea pre-requisite] Console refactor and extension

abdullahtahiriyo:
This PR is the first of a series to introduce NotificationArea.

This PR is mostly directed to refactor Console and Extend it to new messages. It could have been great to first refactor and then extend, but actually the refactor became advisable during the extension.

The core of the refactor/extension lies in Console.h and Console.cpp. The rest of the changes are a consequence of it. When reviewing it makes sense to start there.

Console:
* For refactor:
- variadic functions are substituted with template functions with parameter packs.
- To provide support for the c style printf, fmt library is used.
- Template functions with parameter packs were moved to header file.

* For extension:
- Messages are extended to include a Notifier string, to indicate the originator of the message/notification.
- CriticalMessages are added. This is intended to substitute the signal-slot mechanism for critical messages in a subsequent PR.
- Notifications and TranslatedNotification messages are added. These are intended to convey user notifications, both already translated ones or to be translated ones (see commit description). These will be used by the NotificationArea PR.

### DIFF
--- a/src/App/DocumentObject.cpp
+++ b/src/App/DocumentObject.cpp
@@ -282,6 +282,16 @@ std::string DocumentObject::getFullName() const {
     return name;
 }
 
+std::string DocumentObject::getFullLabel() const {
+    if(!getDocument())
+        return "?";
+
+    auto name = getDocument()->Label.getStrValue();
+    name += "#";
+    name += Label.getStrValue();
+    return name;
+}
+
 const char *DocumentObject::getNameInDocument() const
 {
     // Note: It can happen that we query the internal name of an object even if it is not
--- a/src/Base/Console.h
+++ b/src/Base/Console.h
@@ -25,6 +25,7 @@
 #define BASE_CONSOLE_H
 
 // Std. configurations
+#include <array>
 #include <chrono>
 #include <map>
 #include <set>
@@ -32,11 +33,13 @@
 #include <sstream>
 #include <FCGlobal.h>
 
+#include <fmt/printf.h>
+
 // Python stuff
 using PyObject = struct _object;
 using PyMethodDef = struct PyMethodDef;
 
-//FIXME: ISO C++11 requires at least one argument for the "..." in a variadic macro
+//FIXME: Even with parameter packs this is necessary for MSYS2
 #if defined(__clang__)
 # pragma clang diagnostic push
 # pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
@@ -361,24 +364,24 @@ using PyMethodDef = struct PyMethodDef;
         _instance.prefix(_str,_file,_line) << _msg;\
         if(_instance.add_eol) \
             _str<<std::endl;\
-        Base::Console()._func(_str.str().c_str());\
+        Base::Console()._func("",_str.str().c_str());\
         if(_instance.refresh) Base::Console().Refresh();\
     }\
 }while(0)
 
 #define _FC_PRINT(_instance,_l,_func,_msg) __FC_PRINT(_instance,_l,_func,_msg,__FILE__,__LINE__)
 
-#define FC_MSG(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_MSG,NotifyMessage,_msg)
-#define FC_WARN(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_WARN,NotifyWarning,_msg)
-#define FC_ERR(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_ERR,NotifyError,_msg)
-#define FC_LOG(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_LOG,NotifyLog,_msg)
-#define FC_TRACE(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_TRACE,NotifyLog,_msg)
+#define FC_MSG(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_MSG,Notify<Base::LogStyle::Message>,_msg)
+#define FC_WARN(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_WARN,Notify<Base::LogStyle::Warning>,_msg)
+#define FC_ERR(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_ERR,Notify<Base::LogStyle::Error>,_msg)
+#define FC_LOG(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_LOG,Notify<Base::LogStyle::Log>,_msg)
+#define FC_TRACE(_msg) _FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_TRACE,Notify<Base::LogStyle::Log>,_msg)
 
-#define _FC_MSG(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_MSG,NotifyMessage,_msg,_file,_line)
-#define _FC_WARN(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_WARN,NotifyWarning,_msg,_file,_line)
-#define _FC_ERR(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_ERR,NotifyError,_msg,_file,_line)
-#define _FC_LOG(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_LOG,NotifyLog,_msg,_file,_line)
-#define _FC_TRACE(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_TRACE,NotifyLog,_msg,_file,_line)
+#define _FC_MSG(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_MSG,Notify<Base::LogStyle::Message>,_msg,_file,_line)
+#define _FC_WARN(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_WARN,Notify<Base::LogStyle::Warning>,_msg,_file,_line)
+#define _FC_ERR(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_ERR,Notify<Base::LogStyle::Error>,_msg,_file,_line)
+#define _FC_LOG(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_LOG,Notify<Base::LogStyle::Log>,_msg,_file,_line)
+#define _FC_TRACE(_file,_line,_msg) __FC_PRINT(FC_LOG_INSTANCE,FC_LOGLEVEL_TRACE,Notify<Base::LogStyle::Log>,_msg,_file,_line)
 
 #define FC_XYZ(_pt) '('<<(_pt).X()<<", " << (_pt).Y()<<", " << (_pt).Z()<<')'
 #define FC_xy(_pt) '('<<(_pt).x<<", " << (_pt).y<<')'
@@ -445,11 +448,6 @@ using PyMethodDef = struct PyMethodDef;
 
 #endif //FC_LOG_NO_TIMING
 
-//TODO: Get rid of this forward-declaration
-namespace Base {
-    class ConsoleSingleton;
-} // namespace Base
-
 //TODO: Get rid of this typedef
 using ConsoleMsgFlags = unsigned int;
 
@@ -470,7 +468,10 @@ enum class LogStyle{
     Warning,
     Message,
     Error,
-    Log
+    Log,
+    Critical,               // Special message to mark critical notifications
+    Notification,           // Special message for notifications to the user (e.g. educational)
+    TranslatedNotification, // Special message for already translated notifications to the user (e.g. educational)
 };
 
 /** The Logger Interface
@@ -484,23 +485,57 @@ class BaseExport ILogger
 {
 public:
     ILogger()
-        :bErr(true),bMsg(true),bLog(true),bWrn(true){}
+    :bErr(true), bMsg(true), bLog(true), bWrn(true), bCritical(true), bNotification(false), bTranslatedNotification(false){}
     virtual ~ILogger() = 0;
 
     /** Used to send a Log message at the given level.
      */
-    virtual void SendLog(const std::string& msg, LogStyle level) = 0;
+    virtual void SendLog(const std::string& notifiername, const std::string& msg, LogStyle level) = 0;
+
+    /**
+     * Returns whether a LogStyle category is active or not
+     */
+    bool isActive(Base::LogStyle category) {
+        if(category == Base::LogStyle::Log) {
+           return bLog;
+        }
+        else
+        if(category == Base::LogStyle::Warning) {
+            return bWrn;
+        }
+        else
+        if(category == Base::LogStyle::Error) {
+            return bErr;
+        }
+        else
+        if(category == Base::LogStyle::Message) {
+            return bMsg;
+        }
+        else
+        if(category == Base::LogStyle::Critical) {
+            return bCritical;
+        }
+        else
+        if(category == Base::LogStyle::Notification) {
+            return bNotification;
+        }
+        else
+        if(category == Base::LogStyle::TranslatedNotification) {
+            return bTranslatedNotification;
+        }
+        return false;
+    }
 
     virtual const char *Name(){return nullptr;}
-    bool bErr,bMsg,bLog,bWrn;
+    bool bErr, bMsg, bLog, bWrn, bCritical, bNotification, bTranslatedNotification;
 };
 
 
 /** The console class
  *  This class manage all the stdio stuff. This includes
- *  Messages, Warnings, Log entries and Errors. The incoming
- *  Messages are distributed with the FCConsoleObserver. The
- *  FCConsole class itself makes no IO, it's more like a manager.
+ *  Messages, Warnings, Log entries, Errors, Criticals, Notifications and
+ *  TranslatedNotifications. The incoming Messages are distributed with the
+ *  FCConsoleObserver. The FCConsole class itself makes no IO, it's more like a manager.
  *  \par
  *  ConsoleSingleton is a singleton! That means you can access the only
  *  instance of the class from every where in c++ by simply using:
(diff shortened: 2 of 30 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350">pipeline status</a> for feature branch [PR_8350](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350). Pipeline [767638986](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/767638986) was triggered at [ca93116](https://github.com/FreeCAD/FreeCAD/pull/8350/commits/ca93116159165a5d860850b2c1ecd232e21d6c33). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
@berndhahnebach 

GitLab-ci is failing because I have added a new library (libfmt). This I have added to the Dockerfile (see last commit). I might not have done it right, or maybe there is some other manual step to be made?

luzpaz:
> GitLab-ci is failing because I have added a new library (libfmt). This I have added to the Dockerfile (see last commit). I might not have done it right, or maybe there is some other manual step to be made?

@abdullahtahiriyo is it the same fmt in #8077  ?

abdullahtahiriyo:
Yes, it is the same library. Thanks for cross posting.

.github/workflows/sub_buildUbuntu2204.yml
```diff
@@ -135,7 +135,8 @@ jobs:
                             shiboken2                                   \
                             swig                                        \
                             ccache                                      \
-                            xvfb
+                            xvfb                                        \
+                            libfmt-dev
```
luzpaz, line 139: Are these dependencies that we need to let downstream packagers know about once this PR is merged?

abdullahtahiriyo: It has been the subject of quite some thought.

Finally, @chennes managed to have cmake download the library from github, build it, and then build FreeCAD:
https://github.com/FreeCAD/FreeCAD/pull/8508

libfmt, when available as a package, is generally a library to be statically linked (not shared), so it becomes part of the code of the binary.

I imagine (do not hold me for an expert) downstream packagers will just provide a source package with what we have, including the magic in cmake, and will build the binary from it. The only issue I can think about is, if the computer building the packages does not have access to internet for cmake (some kind of limitation/protection) and cannot download the dependency. We may ask Kurt, he will know at least the Debian side of the story.

abdullahtahiriyo:
When Chris' PR about lib fmt is merged, I will rebase this PR on top of his.

chennes:
> When Chris' PR about lib fmt is merged, I will rebase this PR on top of his.

I just clicked the button, it's all yours.

abdullahtahiriyo:
> I just clicked the button, it's all yours.

It is a pleasure to work with you.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350">pipeline status</a> for feature branch [PR_8350](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350). Pipeline [782184547](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/782184547) was triggered at [8d20988](https://github.com/FreeCAD/FreeCAD/pull/8350/commits/8d209882ea77e7180213dac1564840be259664c7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/App/DocumentObject.cpp
```diff
@@ -282,6 +282,16 @@ std::string DocumentObject::getFullName() const {
     return name;
 }
```
wwmayer, line 284: > If the real name of the document is in foreign languages, such as JP, it is not correctly coded (document name appears as underscores).

It has nothing to do with incorrect encoding but the fact that for Python identifiers only a small number of characters were allowed. That's why all other characters are replaced with underscores. Now with Py3 any unicode characters are allowed and there is already a request to lift the current restrictions.

src/Base/Console.h
```diff
@@ -667,11 +746,150 @@ class BaseExport LogLevel {
...
+{
+    std::string format = fmt::sprintf(pMsg, args...);
+
+    if (connectionMode == Direct) {
+        Notify<category>(notifiername,format);
+    }
+    else {
+        FreeCAD_ConsoleMsgType type;
+
+        if constexpr(category == Base::LogStyle::Log) {
```
wwmayer, line 863: An alternative of using the if-else blocks is to use an unordered_map that you have to fill once where LogStyle is the key. But this cannot be declared as constexpr. But then std::array can be used which supports constexpr since C++17. IN order to access the element the category must be cast to an init or size_t.

wwmayer:
I have looked through the code and looks all good so far. I will checkout your branch and play a bit with it.

src/Base/ConsoleObserver.h
```diff
@@ -57,15 +57,16 @@ class BaseExport ConsoleObserverStd: public ILogger
...
 private:
-    void Warning(const char *sWarn);
-    void Message(const char *sMsg);
-    void Error  (const char *sErr);
-    void Log    (const char *sErr);
+    void Warning        (const char *sWarn);
+    void Message        (const char *sMsg);
+    void Error          (const char *sErr);
+    void Log            (const char *sLog);
+    void CriticalMessage(const char *sCriticalMsg);
```
wwmayer, line 69: What about calling this just Critical()? For error or log messages we also only call them Error() or Log()

wwmayer:
I was curious about the price of longer compile times and it's not that bad. As a test I built the version before and after this PR  with a single thread only.
Before: 131 min
After: 141 min
which means it's an increase of 7.6 %.

src/Base/Console.h
wwmayer: This is how it could be written with std::array. Important is to add the types in right order
```
constexpr std::array msgTypes {
    FreeCAD_ConsoleMsgType::MsgType_Wrn,
    FreeCAD_ConsoleMsgType::MsgType_Txt,
    FreeCAD_ConsoleMsgType::MsgType_Err,
    FreeCAD_ConsoleMsgType::MsgType_Log,
    FreeCAD_ConsoleMsgType::MsgType_CriticalTxt,
    FreeCAD_ConsoleMsgType::MsgType_Notification,
    FreeCAD_ConsoleMsgType::MsgType_TranslatedNotification
};

FreeCAD_ConsoleMsgType type = msgTypes.at(static_cast<std::size_t>(category));
```

src/Base/ConsoleObserver.h
abdullahtahiriyo: It is ok. I can bring this change when rebasing. Let me know when you think it is ready.

abdullahtahiriyo:
> I was curious about the price of longer compile times and it's not that bad. As a test I built the version before and after this PR with a single thread only. Before: 131 min After: 141 min which means it's an increase of 7.6 %.

We did know there was going to be an increase. The author is open about it:
https://github.com/fmtlib/fmt

Hopefully, we will also cash an increase in speed.

wwmayer:
> Hopefully, we will also cash an increase in speed.

We have several places where strings are created with string streams. Replacing this with format() causes a considerable increase in speed. That's how we actually came to this library.

src/Base/ConsoleObserver.h
wwmayer: > Let me know when you think it is ready.

From my point of view the PR has no serious issues. However, before merging it we should make sure that it also compiles on Windows (MSVC and MSYS2).

src/Base/Console.h
abdullahtahiriyo: This I will encapsulate in a separate function:

```
constexpr ConsoleSingleton::FreeCAD_ConsoleMsgType ConsoleSingleton::getConsoleMsg(Base::LogStyle style)
{
    constexpr std::array msgTypes { // In order of Base::LogStyle
        FreeCAD_ConsoleMsgType::MsgType_Wrn,
        FreeCAD_ConsoleMsgType::MsgType_Txt,
        FreeCAD_ConsoleMsgType::MsgType_Err,
        FreeCAD_ConsoleMsgType::MsgType_Log,
        FreeCAD_ConsoleMsgType::MsgType_CriticalTxt,
        FreeCAD_ConsoleMsgType::MsgType_Notification,
        FreeCAD_ConsoleMsgType::MsgType_TranslatedNotification
    };

    return msgTypes.at(static_cast<std::size_t>(style));
}
```

abdullahtahiriyo:
@wwmayer 

I have for pushed the branch. As changes:
- Rebased on today's master.
- The commit message of the first commit makes no longer reference to the underscore issue.
- The branching in Console.h is substituted by a constexpr function encapsulating the mapping.
- All references to Critical Message have been amended to Critical

I have run the GUI tests locally. 

I cannot compile for MSVC and MSYS2. Is it something you would do? Should I ask @donovaly?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350">pipeline status</a> for feature branch [PR_8350](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350). Pipeline [795443171](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/795443171) was triggered at [d36b9b3](https://github.com/FreeCAD/FreeCAD/pull/8350/commits/d36b9b31e2ed3fb81238f92a9e2825c26f0afad7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
Interesting. 

The PR builds locally in Ubuntu 20-04.

CI for Ubuntu 22-04 is failing because:

```/home/runner/work/FreeCAD/FreeCAD/src/Base/Console.h:712:26: error: no viable constructor or deduction guide for deduction of template arguments of 'array'
    constexpr std::array msgTypes { // In order of Base::LogStyle
                         ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/tuple:1595:45: note: candidate function template not viable: requires 1 argument, but 7 were provided
  template<typename _Tp, size_t _Nm> struct array;
                                            ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/tuple:1595:45: note: candidate function template not viable: requires 0 arguments, but 7 were provided```

However, there is a deduction guide for std::array for parameter packs in C++17:
https://en.cppreference.com/w/cpp/container/array/deduction_guides

**Comment by wwmayer** [comment]
> CI for Ubuntu 22-04 is failing because:

Maybe `#include <array>` is missing?

> I cannot compile for MSVC

Just finished the MSVC build and it requires these changes:
* add `#include <array>` in Console.h
* Because the template function ConsoleSingleton::Notify() is implemented in the source it needs an explicit export (although the class is already exported).
```
In Console.cpp:
template <Base::LogStyle category> BaseExport
void ConsoleSingleton::Notify(const std::string & notifiername, const std::string & msg) 
{

In Console.h:
template <LogStyle> 
BaseExport void Notify(const std::string & notifiername, const std::string & msg); 
```
* In GuiConsole.cpp the signature of GUIConsole::SendLog wasn't adjusted:
```
// "const std::string& notifiername" was missing
void GUIConsole::SendLog(const std::string& notifiername, const std::string& msg, Base::LogStyle level)
```

> and MSYS2

This build is running at the moment. I will write an extra post when it's done.

**Comment by abdullahtahiriyo** [comment]
> Maybe `#include <array>` is missing?

I thought of that, because the deduction rule is included in that file. However, it did not make much sense to me, as the compiler "knew" about the 0 and 1 parameters (from the error message).

I have done this and all the other windows fixes in a local branch, but I wait for MSYS2, as it often requires some template trickery.

**Comment by wwmayer** [comment]
> I thought of that, because the deduction rule is included in that file. However, it did not make much sense to me, as the compiler "knew" about the 0 and 1 parameters (from the error message).

On Windows FreeCADBase was building fine but I got failures in other modules because array was not included.

**Comment by wwmayer** [comment]
> This build is running at the moment. I will write an extra post when it's done.

Unfortunately, an earlier commit in master broke the MSYS2 build. Now FreeCADApp causes tons of linking errors related to some boost classes. I have to figure out this bad commit, first.

**Comment by abdullahtahiriyo** [comment]
> Unfortunately, an earlier commit in master broke the MSYS2 build. Now FreeCADApp causes tons of linking errors related to some boost classes. I have to figure out this bad commit, first.

Thanks for building.

I have just pushed the previous fixes, as I will be updating my PC.

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350">pipeline status</a> for feature branch [PR_8350](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350). Pipeline [795686255](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/795686255) was triggered at [15c7752](https://github.com/FreeCAD/FreeCAD/pull/8350/commits/15c7752075724d811f90d8a8a5776800488904d8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by wwmayer** [comment]
> Unfortunately, an earlier commit in master broke the MSYS2 build.

master is fine. There is somehow a collision with fmtlib and boost's graphviz that causes hundreds of linking errors with Document.cpp. If I comment out the two lines in Console.h related to fmtlib everything builds and links fine.

So, I have to do further investigations...

**Comment by abdullahtahiriyo** [comment]
> master is fine. There is somehow a collision with fmtlib and boost's graphviz that causes hundreds of linking errors with Document.cpp. If I comment out the two lines in Console.h related to fmtlib everything builds and links fine.
> 
> So, I have to do further investigations...

That is quite unexpected... now it builds on everything but MSYS2.

I tried to search on Internet for MSYS2 boost graphviz linking errors and nothing obvious pops out.

If boost graphviz is not using libfmt, what may cause the collision? May it relate to the type of fmt library being statically linked? POSITION_INDEPENDENT_CODE? ...

... not being able to test I may actually be  making more trouble for you than helping...

**Comment by wwmayer** [comment]
> If boost graphviz is not using libfmt, what may cause the collision?

Don't know what it is but I found a way to avoid it. All what needs to be done is moving the graphviz related code to a separate source file. Since Document.cpp is anyway too large it's actually not a bad thing to split the file.

The next problem then occurs in RobotGui where the linker complains about some multiple defined symbols like `fmt::v9::detail::dragonbox::to_decimal`. This could be fixed by creating a shared library of fmt.

**Comment by wwmayer** [comment]
https://github.com/FreeCAD/FreeCAD/pull/8737

**Comment by abdullahtahiriyo** [comment]
> #8737

Just looked at it.

So basically the "fix" is moving the dependency ```#include <boost/graph/graphviz.hpp>``` to a separate file. 

Then for the Robot issue, I grep-ed the source for boost:

```
grep -r -i boost *
App/kdl_cp/CMakeLists.txt:TARGET_LINK_LIBRARIES(orocos-kdl ${Boost_LIBRARIES})
App/kdl_cp/utilities/svd_eigen_Macie.hpp:#ifndef SVD_BOOST_MACIE
App/kdl_cp/utilities/svd_eigen_Macie.hpp:#define SVD_BOOST_MACIE
App/CMakeLists.txt:    ${Boost_INCLUDE_DIRS}
Gui/CMakeLists.txt:    ${Boost_INCLUDE_DIRS}
```

There is not a single Boost reference, yet boost is linked in? Perhaps removing it would lessen the "interference"? I dislike this sourcery guessin, but it is what we get with MSYS2.

**Comment by wwmayer** [comment]
> Then for the Robot issue, I grep-ed the source for boost:

In RobotGui boost isn't the problem. Fore some reason some fmt functions are defined multiple times:
fmt::v9::detail::dragonbox::to_decimal, fmt::v9::detail::dragonbox::decimal_fp, fmt::v9::detail::thousands_sep_result, fmt::v9::detail::throw_format_error, fmt::v9::detail::is_printable

**Comment by abdullahtahiriyo** [comment]
> In RobotGui boost isn't the problem. Fore some reason some fmt functions are defined multiple times:
> fmt::v9::detail::dragonbox::to_decimal, fmt::v9::detail::dragonbox::decimal_fp, fmt::v9::detail::thousands_sep_result, fmt::v9::detail::throw_format_error, fmt::v9::detail::is_printable

Before this PR MSYS2 built. The changes are to Base/Console.h and the inclusion of fmt as a statically linked library. 

My guess is that if we comment these lines:
```
AppRobotGui.cpp:#include <Base/Console.h>
AppRobotGui.cpp:        PyErr_SetString(PyExc_ImportError, "Cannot load Gui module in console application.");
AppRobotGui.cpp:    Base::Console().Log("Loading GUI of Robot module... done\n");
TaskDlgEdge2Trac.cpp:#include <Base/Console.h>
TaskDlgEdge2Trac.cpp:        Base::Console().Warning("TaskDlgEdge2Trac::clicked(): %s\n", e.what());
TaskDlgEdge2Trac.cpp:        Base::Console().Warning("TaskDlgEdge2Trac::accept(): %s\n", e.what());
ViewProviderRobotObject.cpp:    //Base::Console().Message("M %f %f %f\n", M[3][0], M[3][1], M[3][2]);
```

Then it will build and link. 

If that is the case, I have strong problems to follow the duplicated symbol saga, as KDL or other Robot dependency does not have fmt as dependency (that I managed to find).

I tried to focus today on where those multiple symbols might be coming from, an found this post showing how to use obj2yaml interesting:
https://github.com/msys2/MINGW-packages/issues/6855#issuecomment-680859662

On a sourcery-catching attempt, would it make any difference to declare the Send template function in Console.h inline?

File: src/Base/Console.h [hunk 5]
```diff
@@ -25,23 +25,20 @@
...
 #include <FCGlobal.h>
 
+#include <fmt/printf.h>
+
 // Python stuff
 using PyObject = struct _object;
 using PyMethodDef = struct PyMethodDef;
 
-//FIXME: ISO C++11 requires at least one argument for the "..." in a variadic macro
-#if defined(__clang__)
```
wwmayer, line 40: Unfortunately, this block (and the corresponding pop at the end) shouldn't be dropped because on MSYS2 clang raises this warning for each file that includes this header.

wwmayer:
> On a sourcery-catching attempt, would it make any difference to declare the Send template function in Console.h inline?

Have you changed it?  I can test it tomorrow again to see what happens.

wwmayer:
At the moment it doesn't compile neither with MSVC nor under MSYS2 because the export macro is missing:
```
    template <LogStyle>
    void Notify(const std::string & notifiername, const std::string & msg);
```
==>
```
    template <LogStyle>
    BaseExport void Notify(const std::string & notifiername, const std::string & msg);
```

Besides that a nicer way that doesn't require the extra export macro is to add a private function 
`void NotifyPrivate(LogStyle category, const std::string& notifiername, const std::string& msg);`
and the implementation of Notify can go to the header file:
```
template <LogStyle category>
void Notify(const std::string& notifiername, const std::string& msg) {
    NotifyPrivate(category, notifiername, msg);
}
```

src/Base/Console.h
abdullahtahiriyo: Ok. 
I removed it because it was referring to the "..." in a variadic macro and now there is only ... as part of parameter packs. 

I will reinstate it.

abdullahtahiriyo:
> Have you changed it? I can test it tomorrow again to see what happens.

No I have not, as I do not want to keep triggering the CI while trying to blindly fix the issue. But I will do when making the additional changes indicated today, as any way it appears better to have it inline.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350">pipeline status</a> for feature branch [PR_8350](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350). Pipeline [797333448](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/797333448) was triggered at [e77b026](https://github.com/FreeCAD/FreeCAD/pull/8350/commits/e77b026f119dd7a71cde373fa704dae717322335). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350">pipeline status</a> for feature branch [PR_8350](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8350). Pipeline [797349466](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/797349466) was triggered at [18e866f](https://github.com/FreeCAD/FreeCAD/pull/8350/commits/18e866fed9d9c060412f9b09c5e7b076628c3949). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
So I have:
- reinstated the clang preprocessor directives for variadic macros. I have clarified the comment for future reference.
- declared inline all parameter pack functions and the ones from the same class using them.
- Moved Notify to the header, drop the exporting macro and use a private function.

As yet another idea of possible trial is to try to use the master of fmt instead of the 9.1.0 tag version. I see that between August 2022 (release date of the tag) and today there are quite a few commits.

In the past, fmt was fixed for the issue of multiply defined symbols, but that fix is already in 9.1.0:
https://github.com/fmtlib/fmt/issues/1306

I am sorry to be loading you with this...

wwmayer:
> In the past, fmt was fixed for the issue of multiply defined symbols, but that fix is already in 9.1.0:

I don't know why it creates the multiple symbols in RobotGui but nowhere else. But as I said I got it working by creating a shared library of fmt.

abdullahtahiriyo:
> I don't know why it creates the multiple symbols in RobotGui but nowhere else. But as I said I got it working by creating a shared library of fmt.

Yes, I got that. However in Ubuntu 20.04 there is no shared library in the repositories.

What is the way forward then? Are we going to switch the current FC code to a shared library?

wwmayer:
> However in Ubuntu 20.04 there is no shared library in the repositories.

If Ubuntu 20.04 works with the static library there is no need to switch to a shared library. The trick to build the shared version from the sources was to add this block to SetupLibFmt.cmake
```
if (MINGW)
    set (BUILD_SHARED_LIBS ON)
endif()
```

But BUILD_SHARED_LIBS is a global CMake variable that affects all calls of add_library() to always create a shared library.

abdullahtahiriyo:
> But BUILD_SHARED_LIBS is a global CMake variable that affects all calls of add_library() to always create a shared library.

The sentence starts with "But", but I fail to see a negative consequence. For MINGW there is no alternative, so it is what we want. For the rest the conditional block does not apply and a static library will be created.

My concern is more about what will happen with the shared library we are creating. Will we install it?

Perhaps a working alternative is to require the fmt package in MSYS2 and do not try to build a shared library at all. The package has a dll file (in addition to the .a), which I presume is a shared library:
https://packages.msys2.org/package/mingw-w64-x86_64-fmt

What do you think?

abdullahtahiriyo:
Ok. 

So wrapping up... Should I merge this PR as it is now or do you prefer to defer it until after another action?

wwmayer:
> So wrapping up... Should I merge this PR as it is now or do you prefer to defer it until after another action?

From my point of view it's ready for merging.

abdullahtahiriyo:
Thank you.

PR #8384, opened 2023-02-07
[FEM] Elmer: better sorting of equations

donovaly:
- the elctricforce equation is actually a postprocessor and thus should not be first in the list, but after the electrostatics equation (where it belongs to technically)
- also change menu name to be consistent with the other FEM menus

### DIFF
--- a/src/Mod/Fem/Gui/Command.cpp
+++ b/src/Mod/Fem/Gui/Command.cpp
@@ -1355,7 +1355,7 @@ CmdFEMCompEmEquations::CmdFEMCompEmEquations()
 {
     sAppModule = "Fem";
     sGroup = QT_TR_NOOP("Fem");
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8384">pipeline status</a> for feature branch [PR_8384](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8384). Pipeline [769856493](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/769856493) was triggered at [c0b014b](https://github.com/FreeCAD/FreeCAD/pull/8384/commits/c0b014b6afc63057135a201f46a336ae7f428c0d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #8412, opened 2023-02-10
[FEM] change hiding logic on calling task panels

donovaly:
- the typical workflow is to change e.g. a constraint setting or material property and to check how and if this change changes the simulation result. It is therefore very annoying that on every call of a task panel the results are hidden.
- This PR changes this to hide only meshes and the filter functions but to keep the results. If a user don't like to see results for example to select faces, he hides results anyway.

### DIFF
--- a/src/Mod/Fem/femviewprovider/view_base_femobject.py
+++ b/src/Mod/Fem/femviewprovider/view_base_femobject.py
@@ -81,23 +81,15 @@ class VPBaseFemObject(object):
             # avoid edit mode by return False
             # https://forum.freecadweb.org/viewtopic.php?t=12139&start=10#p161062
             return False
-        if hide_mesh is True:
-            # hide all FEM meshes and VTK FemPost* objects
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8412">pipeline status</a> for feature branch [PR_8412](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8412). Pipeline [773303538](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/773303538) was triggered at [98d2261](https://github.com/FreeCAD/FreeCAD/pull/8412/commits/98d226149768a0efa92e7a665dbba140835fa680). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #8464, opened 2023-02-12
[FEM] improve messages in mesh generation

donovaly:
- no longer output an error on mesh from a BooleanFragments. There are cases when this fails, but other cases where one must use a BoopleanFragments compound and this works well, see https://forum.freecad.org/viewtopic.php?t=71070. Therefore a warning is sufficient.
- disable log messages not important for the average users.
- avoid code duplication

### DIFF
--- a/src/Mod/Fem/femmesh/gmshtools.py
+++ b/src/Mod/Fem/femmesh/gmshtools.py
@@ -394,11 +394,11 @@ class GmshTools():
                     self.group_elements[ge] = new_group_elements[ge]
                 else:
                     Console.PrintError("  A group with this name exists already.\n")
-        else:
-            Console.PrintMessage("  No Group meshing for analysis.\n")
+        #else:
+        #    Console.PrintMessage("  No Group meshing for analysis.\n")
 
-        if self.group_elements:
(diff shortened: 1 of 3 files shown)


donovaly:
Here an example where a BooleanFragments object works nicely and a compound filter doesn't work: 
[MagneticFieldWire.zip](https://github.com/FreeCAD/FreeCAD/files/10717652/MagneticFieldWire.zip)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8464">pipeline status</a> for feature branch [PR_8464](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8464). Pipeline [775367300](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/775367300) was triggered at [ea784f4](https://github.com/FreeCAD/FreeCAD/pull/8464/commits/ea784f4926a434b5ae9bcb570cbb4c2ce62ce2d6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #8467, opened 2023-02-13
CI: Tweaks to pylintrc

### LINKED ISSUE #8367: [Problem] pylint config file may encourage worse code

### Version

0.21 (Development)

### Full version info

```shell
N/A
```

### Subproject(s) affected?

None

### Problem description

The current pylint configuration file needs to be vetted to ensure it is not encouraging coding practices that are actually worse than the code it flags. As with any linter, it contains a large number of heuristic checks with somewhat arbitrary cutoffs. We should ensure those cutoffs are dialed in for the way we expect FreeCAD’s code to look.

As an example, because we use Qt and QThread it is common to have a class with only a single public function (`run`). So I think we should disable entirely the too-few-public-methods check. I’m sure there are others, like how many class members is too many, etc.

### Anything else?

_No response_

### Code of Conduct

chennes:
A few small changes to our pylint configuration:
* Correct exception type specification
* Don't require more than one public method
* Make max attributes match max locals
* Warn on use of PySide2 (rather than PySide)

Fixes #8367.

---

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -3,7 +3,7 @@
 # Analyse import fallback blocks. This can be used to support both Python 2 and
 # 3 compatible code, which means that the block might have code that exists
 # only in one or another interpreter, leading to false positives when analysed.
-analyse-fallback-blocks=no
+analyse-fallback-blocks=yes
 
 # Load and enable all available extensions. Use --list-extensions to see a list
 # all available extensions.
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8467">pipeline status</a> for feature branch [PR_8467](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8467). Pipeline [775450214](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/775450214) was triggered at [7e7893f](https://github.com/FreeCAD/FreeCAD/pull/8467/commits/7e7893f62709640d2b6fbeedeb6e14c97406bf54). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #8478, opened 2023-02-14
[FEM] improve visualization of pvtu files

donovaly:
- To speed up analyses one calculates on several CPU cores. The result is a partial VTU (*.ptvu) file. Unfortunately when applying a transparency the boundaries of the volumes computed by each CPU core are always visible. This makes it often unusable for visualizations. The solution is to filter the results the same way a clip filter does.

### DIFF
--- a/src/Mod/Fem/Gui/ViewProviderFemPostObject.cpp
+++ b/src/Mod/Fem/Gui/ViewProviderFemPostObject.cpp
@@ -38,6 +38,7 @@
 # include <vtkCellArray.h>
 # include <vtkCellData.h>
 # include <vtkDoubleArray.h>
+# include <vtkImplicitFunction.h>
 # include <vtkPointData.h>
 
 # include <QApplication>
@@ -55,6 +56,8 @@
 #include <Gui/SelectionObject.h>
 #include <Gui/SoFCColorBar.h>
 #include <Gui/TaskView/TaskDialog.h>
+#include <Gui/View3DInventor.h>
+#include <Gui/View3DInventorViewer.h>
 #include <Mod/Fem/App/FemPostFilter.h>
 
 #include "ViewProviderFemPostObject.h"
@@ -265,8 +268,8 @@ void ViewProviderFemPostObject::attach(App::DocumentObject* pcObj)
     (void)setupPipeline();
 }
 
-SoSeparator* ViewProviderFemPostObject::getFrontRoot() const {
-
+SoSeparator* ViewProviderFemPostObject::getFrontRoot() const
+{
     return m_colorRoot;
 }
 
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8478">pipeline status</a> for feature branch [PR_8478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8478). Pipeline [776555033](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/776555033) was triggered at [d77a70b](https://github.com/FreeCAD/FreeCAD/pull/8478/commits/d77a70bcd59a34dbdb87119002e60186b9f9e682). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8478">pipeline status</a> for feature branch [PR_8478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8478). Pipeline [776567831](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/776567831) was triggered at [8b969cc](https://github.com/FreeCAD/FreeCAD/pull/8478/commits/8b969cc3090994fa3099a8dfd5d5e4a56b81ba63). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8478">pipeline status</a> for feature branch [PR_8478](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8478). Pipeline [778361071](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/778361071) was triggered at [3e58561](https://github.com/FreeCAD/FreeCAD/pull/8478/commits/3e58561c15dc5dfa5d264406a49c0e34de7b2964). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

marioalexis84:
@donovaly can you review this changes?
With this commit I get a crash when trying to change the `DisplayMode` property on a `PostPipeline` object.
The crash is specifically in the `WriteColorData` function since  the `vtkDataArray* data` variable is set to `nullptr`.
(I guess it come from an undefined behavior)

donovaly:
> With this commit I get a crash when trying to change the `DisplayMode` property on a `PostPipeline` object. The crash is specifically in the `WriteColorData` function since the `vtkDataArray* data` variable is set to `nullptr`. (I guess it come from an undefined behavior)

Is this with latest master?
I cannot reproduce this here. (I played with the feature for about a week and never got a crash nor a VTK warning.)
What happens when you recreate the Pipeline (e.g. by re-running the calculation)?

Do you have a testcase for me that you can upload here as ZIP?

donovaly:
I forgot to ask, what is your VTK version?

donovaly:
In general I see that there is no check of there is data:

`vtkDataArray* data = pd->GetPointData()->GetArray(array);`

I recently made a PR do add such checks thanks to Werner why informed me about this. So i will have to add checks in ViewProviderFemPostObject as well.

nevertheless, it would be good to be able to reproduce why you don't get data. I took now some old simulations and cannot reproduce the crash. So a file from you and a recipe would be very helpful.

donovaly:
OK, I can now reproduce. It only happens when
- pipeline is visible, Field is not None and the DisplayMode is Surface
- then changing DisplayMode to Nodes in the parameter settings, work, but doing the change in the dialog crashes.
So there is something in the dialog handling. I will check as soon as possible.

marioalexis84:
> OK, I can now reproduce. It only happens when
    pipeline is visible, Field is not None and the DisplayMode is Surface

I get a crash in most of the cases if `Field` is not `None`, even switching between nodes and wireframe in the property view.
> then changing DisplayMode to Nodes in the parameter settings, work, but doing the change in the dialog crashes.
So there is something in the dialog handling.

For me, the change in the property view lead a crash inmediatly as weel in the Result dialog, but there is a depedency on the workflow (if the initial DisplayMode is Surface, the crash don't happen inmediatly), so I don't think the problem is in the dialog.

Here is the stack trace on Linux: (can you post the stack trace on Windows?)
```

Program received signal SIGSEGV, Segmentation fault.
#0  /usr/lib/libc.so.6(+0x38f50) [0x7f10c1451f50]
#1  0x7f0fbf973729 in vtkDataArray::GetRange(double*, int)
#2  0x7f0fbf970213 in FemGui::ViewProviderFemPostObject::WriteColorData(bool)
#3  0x7f0fbf971681 in FemGui::ViewProviderFemPostObject::show()
#4  0x7f10c74e83a5 in Gui::ViewProviderDocumentObject::onChanged(App::Property const*) 
#5  0x7f0fbf970eab in FemGui::ViewProviderFemPostObject::onChanged(App::Property const*) 
#6  0x7f10c5574990 in App::Property::hasSetValue() 
#7  0x7f10c55fabc2 in App::PropertyBool::setValue(bool) 
#8  0x7f0fbf9706f1 in FemGui::ViewProviderFemPostObject::filterArtifacts(vtkDataObject*) 
#9  0x7f0fbf970c5f in FemGui::ViewProviderFemPostObject::setupPipeline() 
#10  0x7f0fbf96ee0a in FemGui::ViewProviderFemPostObject::updateVtk() 
#11  0x7f0fbf96ed14 in FemGui::ViewProviderFemPostObject::setDisplayMode(char const*) 
#12  0x7f10c74e9d15 in Gui::ViewProviderDocumentObject::setActiveMode() 
#13  0x7f10c74e830a in Gui::ViewProviderDocumentObject::onChanged(App::Property const*) 
#14  0x7f0fbf970eab in FemGui::ViewProviderFemPostObject::onChanged(App::Property const*) 

```

donovaly:
Here is the fix: #8519

I relied on the existing code that it is checked in data exist when it is called. Thanks to wmayer, I already added checks few days back: #8488 Now I added two more missing checks.

That there is on some calls no data is correct since setupPipeline() is called for every field when there is a VectorMode change, not only for the currently active field. When it is called for the "none" field, there can of course not be any filter result. And we can just return in this case.

marioalexis84:
>That there is on some calls no data is correct since setupPipeline() is called for every field when there is a VectorMode change, not only for the currently active field.

OK!

PR #8487, opened 2023-02-14
[Gui] font setting fixes for NaviCube

donovaly:
- it turned out that the current check for default font failed because of wrong check for the hinting
- fix bug that when the FontSize is not yet stored in the parameters, the dialog shows a wrong size value


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8487">pipeline status</a> for feature branch [PR_8487](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8487). Pipeline [777753587](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/777753587) was triggered at [3b31963](https://github.com/FreeCAD/FreeCAD/pull/8487/commits/3b31963e8aada0db4e0ae1be6400dcd0ed818237). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/NaviCube.cpp
```diff
@@ -296,24 +296,21 @@ void NaviCube::setFontSize(int size)
 // the Helvetica font is a good start for most OSes
 QFont NaviCube::getDefaultSansserifFont()
 {
-    QFont font(QString::fromLatin1("Helvetica"));
-    if (font.styleHint() & QFont::SansSerif)
+    // On Windows 10 or newer there is no Helvetica font and Arial will be used
+    // On Win 11 sometimes the first font registered to Windows is used.
+    // Both can lead to ugly, even unreadable results
```
Roy-043, line 301: These 3 lines of comment describe an older situation and are therefore hard to understand for readers unaware of that situation.

src/Gui/NaviCube.cpp
```diff
@@ -296,24 +296,21 @@ void NaviCube::setFontSize(int size)
...
-        return font;
-    font.setFamily(QString::fromLatin1("DejaVu Sans"));
-    if (font.styleHint() == QFont::SansSerif || font.styleHint() == QFont::AnyStyle)
-        return font;
-    return font; // We failed, but return whatever we have anyway
+
+    // now check for Helvetica
+    font = QString::fromLatin1("Helvetica");
+    font.setStyleHint(QFont::SansSerif);
+    font.defaultFamily();
```
Roy-043, line 312: `font.defaultFamily()` does not do anything and can be removed.

src/Gui/NaviCube.cpp
donovaly: No, this is the current situation and the reason why I implemented the setting of the font. On Win 11 (at least my new laptop) the simple check for Helvetica delivers nothing and then the first font in Windows list is used. On Win 10 Arial is used. This is the current situation. This comment should make clear why we first select Bahnschrift for Win, then check for Helvetica.

donovaly:
@Roy-043 does the PR work for you? (if you have a non-Windows OS to test)

src/Gui/NaviCube.cpp
Roy-043: Well, then the text does not make sense to me. 'On Windows 10 or newer' includes Windows 11 obviously. It is confusing that you then go on the mention win 11 separately.

Why not something like:
Windows versions since 2017 have the 'Bahnschrift' font (a condensed sans serif font), we first check for that. If that font is not found we let Qt look for 'Helvetica' or a default sans serif font.

donovaly: Thanks. I updated the comment now accordingly.

However, does the PR work for you?

adrianinsaval:
this is how the navicube looks by default (clean config) for me on linux:
image
but this is what I get if I go into preferences
image
C059 is just the first font in the list, I don't think it's what gets matched normally because once I click ok on the preference page the navicube changes to this:
image

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8487">pipeline status</a> for feature branch [PR_8487](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8487). Pipeline [779762586](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/779762586) was triggered at [ee5cb3b](https://github.com/FreeCAD/FreeCAD/pull/8487/commits/ee5cb3b45855af098122b91bcc9ec814d9d3b15a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> this is how the navicube looks by default (clean config) for me on linux:

Many thanks for your help.
Excellent! So the fixed algorithm now finds correctly a Sansserif font.
Also the wrong displayed fontsize is now correct.

> C059 is just the first font in the list,

This is the same issue as was with the foontsize - when the parameter is not yet set, the dialog is not initialized and displays the default. I fixed this now.

> btw, I have no clue how hard to do this would be, but it would be nice if we had the option to set the font to be bold

It can be done, but we should limit the settings in the dialog. I have already requests to change the color of the cube, the color of the arrows, to and an option to draw lines around the arrows, even to change the color of the cube lines.
If desired these can be implemented as parameter-only. Then experts can change the parameter accordingly.

So I will merge the PR now. Please check if everything works for you now when starting with a fresh config.

adrianinsaval:
thanks, I triggered a build from master and will test later today
> It can be done, but we should limit the settings in the dialog.

if you mean for UI reasons we could leverage [QFontDialog](https://doc.qt.io/qt-5/qfontdialog.html) instead of cluttering the page with our own implementation of these widgets, anyways this is not very important for me now, I found a font that is bold and condensed in the list and am happily using it now! (Open Sans Condensed)

donovaly:
> I have already requests to change the color of the cube, the color of the arrows, to and an option to draw lines around the arrows, even to change the color of the cube lines. If desired these can be implemented as parameter-only. Then experts can change the parameter accordingly.

Today I realized that we have these features already, there were just undocumented. I did this now:
https://wiki.freecad.org/Navigation_Cube#Cube_Layout

donovaly:
> if you mean for UI reasons we could leverage [QFontDialog](https://doc.qt.io/qt-5/qfontdialog.html) instead of cluttering the page with our own implementation of these widgets,

That would be a nice solution. So you see there is always something to code on ;-)

adrianinsaval:
I still see the same situation with:
> OS: Arch Linux (KDE/plasma)
> Version: 0.21.0.31922 (Git)
> Build type: None
> Branch: master
> Hash: 08b3bf4462f6264375f1cb4119457ff7452be524
> Python 3.10.9, Qt 5.15.8, Coin 4.0.1, Vtk 9.1.0, OCC 7.6.3
> Locale: English/United Kingdom (en_GB)

A kinda oversized sans font at first launch and the first font in the list once I open the preference page

src/Gui/NaviCube.cpp
```diff
@@ -296,24 +296,18 @@ void NaviCube::setFontSize(int size)
 // the Helvetica font is a good start for most OSes
 QFont NaviCube::getDefaultSansserifFont()
```
adrianinsaval, line 297: btw, shouldn't this be called `getDefaultSansFont`? or just `getDefaultFont`?

donovaly:
> A kinda oversized sansserif font at first launch and the first font in the list once I open the preference page

Shit. So the font selection works. But for a reason I don't understand yet, this is not taken to the dialog.

But we should be close to the solution, if you help me:
- delete all NaviCube parameters and Save to disk in the parameter dialog
- close the parameter dialog
- open the preferences dialog
- now cancel the preferences dialog (pressing its X button, not OK or apply
- open the parameter dialog
What do you see as set parameters? (send a screenshot please)

adrianinsaval:
Interesting, this is the result of doing that:
image

donovaly:
This is the desired result (if there is a font called "Helvetica" on the system). So I don't understand why Helvetica is then not in the dialog as the current font.

Do you have a font called "Helvetica" installed? If not, the font selection mechanism still fails, only outputs the font we were searching font.

adrianinsaval:
No I couldn't find Helvetica in my system, in theory the system should use Nimbus Sans as replacement for it

adrianinsaval:
I checked setting Nimbus Sans in the preferences and it looks identical to what is shown on first launch with a clean config
image
it's shown wierdly in the dialog though
image

donovaly:
OK. What happens if your comment out line 308 in NaviCube.cpp?
(starting without set parameters and canceling the preferences dialog)
What do you then get as new set parameter?

donovaly:
For info: when I comment out the check for Bahnschrift, then I get for the check for Helvetica at least on Windows 10 "Arial". So the font check works here as it should.

Besides this, Nimbus Sans is apparently no good font as it runs too wide.

So to find a good solution, before we perform the generic check for Helvetica, for what font could we check? Meaning what font runs at size 95 narrow enough that "BOTTOM" is inside the cube's face?

adrianinsaval:
image
"Open Sans Condensed" seems to fit best at size 95 but it's still slightly oversized, I'm not sure if dpi settings affect this.

I'll build and test commenting out that line tomorrow

donovaly:
> I'll build and test commenting out that line tomorrow

Thanks. What is the result?

adrianinsaval:
image
same behavior overall but I get this font instead of helvetica if I fon't click ok when entering the preferences page

donovaly:
Thanks. So I have to give up. The right font is actually chosen but we cannot get the info. I see that the code to get the info works on Win 10 but fails on Win 11 the same way as for you. -> I will ask in the forum.

donovaly:
Here is the forum thread: https://forum.freecad.org/viewtopic.php?p=662040#p662040
And here is a new PR to test: #8553

PR #8507, opened 2023-02-16
[FEM] make pvtu file filtering optional

donovaly:
- it is sensible to filter by default, however, it turned out that on some complex models Elmer fails to compute if the CPU mesh volume regions are too small (in most cases) or at a certain mesh region. By disabling the filtering, one gets at least for the latter case a visual feedback where the mesh volume of the different CPU are (by also setting a transparency to the result pipeline).

### DIFF
--- a/src/Mod/Fem/Gui/DlgSettingsFemElmer.ui
+++ b/src/Mod/Fem/Gui/DlgSettingsFemElmer.ui
@@ -6,15 +6,15 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>400</width>
-    <height>203</height>
+    <width>350</width>
+    <height>259</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Elmer</string>
   </property>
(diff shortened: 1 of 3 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8507">pipeline status</a> for feature branch [PR_8507](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8507). Pipeline [779173071](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/779173071) was triggered at [7ef9b3b](https://github.com/FreeCAD/FreeCAD/pull/8507/commits/7ef9b3b5fb4d686db6bf1ba2b0f295f69f611f87). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #8522, opened 2023-02-17
Fix issue #8494

### LINKED ISSUE #8494: [Bug] Two unequal placements reported equal.

### Forums discussion

https://forum.freecad.org/viewtopic.php?t=75448

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Ubuntu 22.04.1 LTS (XFCE/xubuntu)
Word size of FreeCAD: 64-bit
Version: 0.21.0.31906 (Git)
Build type: debug
Branch: master
Hash: 99a7262e60f1b5d79cf169b95be764b045499cdd
Python 3.10.6, Qt 5.15.3, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.1
[/code]
```

### Subproject(s) affected?

Core

### Issue description

1. Load movecube.FCStd attached to the forum posting
2. Select (bottom) cube under Part001
3. Run this code snippet 
```
sel = Gui.Selection.getSelectionEx('',0)[0]
path = sel.SubElementNames[0]
shp = sel.Object.getSubObject(path)
gpl1 = sel.Object.getSubObject(path, retType = 3)
gpl = App.Placement()
for obj in sel.Object.getSubObjectList(path):
    gpl = gpl*obj.Placement

print(f'gpl1 = {gpl1}')
print(f'gpl = {gpl}')
print(f'Same? {gpl.isSame(gpl1, 1E-7)}')
print(gpl.Rotation.isSame(gpl1.Rotation, 1e-7))
print(gpl.Base.isEqual(gpl1.Base, 1e-7))
```
4. Output is
```
>>> print(f'gpl1 = {gpl1}')
gpl1 = Placement [Pos=(0,0,20), Yaw-Pitch-Roll=(106.26,0,0)]
>>> print(f'gpl = {gpl}')
gpl = Placement [Pos=(0,0,20), Yaw-Pitch-Roll=(90,0,0)]
>>> print(f'Same? {gpl.isSame(gpl1, 1E-7)}')
Same? True
>>> print(gpl.Rotation.isSame(gpl1.Rotation, 1e-7))
True
>>> print(gpl.Base.isEqual(gpl1.Base, 1e-7))
True
```
6. gpl and gpl1 shouldn't be considered equal

### Anything else?

_No response_

### Code of Conduct

Jolbas:
When created a Rotation from a scaled matrix, the quaternion become malicious. This filters out the scale from the matrix if there is one. Fixes #8494

### DIFF
--- a/src/Base/Matrix.cpp
+++ b/src/Base/Matrix.cpp
@@ -833,34 +833,43 @@ ScaleType Matrix4D::hasScale(double tol) const
     if (tol == 0.0)
         tol = 1e-9;
 
+    // check if the absolute values are proportionally close or equal
+    auto closeAbs = [&](double a, double b) {
+        double c = fabs(a);
+        double d = fabs(b);
+        if (d>c) return (d-c)/d <= tol;
+        else if (c>d) return (c-d)/c <= tol;
+        return true;
+    };
+
     // get column vectors
-    double dx = Vector3d(dMtrx4D[0][0],dMtrx4D[1][0],dMtrx4D[2][0]).Sqr();
-    double dy = Vector3d(dMtrx4D[0][1],dMtrx4D[1][1],dMtrx4D[2][1]).Sqr();
-    double dz = Vector3d(dMtrx4D[0][2],dMtrx4D[1][2],dMtrx4D[2][2]).Sqr();
+    double dx = getCol(0).Sqr();
+    double dy = getCol(1).Sqr();
+    double dz = getCol(2).Sqr();
     double dxyz = sqrt(dx * dy * dz);
 
     // get row vectors
-    double du = Vector3d(dMtrx4D[0][0],dMtrx4D[0][1],dMtrx4D[0][2]).Sqr();
-    double dv = Vector3d(dMtrx4D[1][0],dMtrx4D[1][1],dMtrx4D[1][2]).Sqr();
-    double dw = Vector3d(dMtrx4D[2][0],dMtrx4D[2][1],dMtrx4D[2][2]).Sqr();
+    double du = getRow(0).Sqr();
+    double dv = getRow(1).Sqr();
+    double dw = getRow(2).Sqr();
     double duvw = sqrt(du * dv * dw);
 
     double d3 = determinant3();
 
     // This could be e.g. a projection, a shearing,... matrix
-    if (fabs(dxyz - d3) > tol && fabs(duvw - d3) > tol) {
+    if (!closeAbs(dxyz, d3) && !closeAbs(duvw, d3)) {
         return ScaleType::Other;
     }
 
-    if (fabs(duvw - d3) <= tol && (fabs(du - dv) > tol || fabs(dv - dw) > tol)) {
+    if (closeAbs(duvw, d3) && (!closeAbs(du, dv) || !closeAbs(dv, dw))) {
         return ScaleType::NonUniformLeft;
     }
 
-    if (fabs(dxyz - d3) <= tol && (fabs(dx - dy) > tol || fabs(dy - dz) > tol)) {
+    if (closeAbs(dxyz, d3) && (!closeAbs(dx, dy) || !closeAbs(dy, dz))) {
         return ScaleType::NonUniformRight;
     }
 
-    if (fabs(dx - 1.0) > tol) {
+    if (fabs(d3 - 1.0) > tol) {
         return ScaleType::Uniform;
     }
 
--- a/src/Base/Rotation.cpp
+++ b/src/Base/Rotation.cpp
@@ -220,32 +220,55 @@ void Rotation::setValue(const double q[4])
 
 void Rotation::setValue(const Matrix4D & m)
 {
-    double trace = (m[0][0] + m[1][1] + m[2][2]);
+    
+    auto type = m.hasScale();
+    if (type == Base::ScaleType::Other) {
+        THROWM(Base::ValueError, "setValue(matrix): Could not determine the rotation.");
+    }
+    Matrix4D mc(m);
+    if (type != Base::ScaleType::NoScaling) {
+        mc.setCol(3, Vector3d(0.0, 0.0, 0.0));
+        if (type == Base::ScaleType::NonUniformRight) {
+            mc.transpose();
+        }
+        double sx = 1.0 / mc.getRow(0).Length();
+        double sy = 1.0 / mc.getRow(1).Length();
+        double sz = 1.0 / mc.getRow(2).Length();
+        mc.scale(sx, sy, sz);
+        if (type == Base::ScaleType::NonUniformRight) {
+            mc.transpose();
+        }
+        if (mc.determinant3() < 0.0) {
+            mc.scale(-1.0, -1.0, -1.0);
+        }
+    }
+    // Extract quaternion
+    double trace = (mc[0][0] + mc[1][1] + mc[2][2]);
     if (trace > 0.0) {
         double s = sqrt(1.0+trace);
         this->quat[3] = 0.5 * s;
         s = 0.5 / s;
-        this->quat[0] = ((m[2][1] - m[1][2]) * s);
-        this->quat[1] = ((m[0][2] - m[2][0]) * s);
-        this->quat[2] = ((m[1][0] - m[0][1]) * s);
+        this->quat[0] = ((mc[2][1] - mc[1][2]) * s);
+        this->quat[1] = ((mc[0][2] - mc[2][0]) * s);
+        this->quat[2] = ((mc[1][0] - mc[0][1]) * s);
     }
     else {
         // Described in RotationIssues.pdf from <http://www.geometrictools.com>
         //
         // Get the max. element of the trace
         unsigned short i = 0;
-        if (m[1][1] > m[0][0]) i = 1;
-        if (m[2][2] > m[i][i]) i = 2;
+        if (mc[1][1] > mc[0][0]) i = 1;
+        if (mc[2][2] > mc[i][i]) i = 2;
 
         unsigned short j = (i+1)%3;
         unsigned short k = (i+2)%3;
 
-        double s = sqrt((m[i][i] - (m[j][j] + m[k][k])) + 1.0);
+        double s = sqrt((mc[i][i] - (mc[j][j] + mc[k][k])) + 1.0);
         this->quat[i] = s * 0.5;
         s = 0.5 / s;
-        this->quat[3] = ((m[k][j] - m[j][k]) * s);
-        this->quat[j] = ((m[j][i] + m[i][j]) * s);
-        this->quat[k] = ((m[k][i] + m[i][k]) * s);
+        this->quat[3] = ((mc[k][j] - mc[j][k]) * s);
+        this->quat[j] = ((mc[j][i] + mc[i][j]) * s);
+        this->quat[k] = ((mc[k][i] + mc[i][k]) * s);
     }
 
     this->evaluateVector();
@@ -752,6 +775,8 @@ bool Rotation::isSame(const Rotation& q, double tol) const
     // This term can be simplified to
     // 2 - 2*(x1*y1 + ... + x4*y4) so that for the equality we have to check
     // 1 - tol/2 <= x1*y1 + ... + x4*y4
+    // This simplification only work if both quats are normalized
+    // Is it safe to assume that?
     // Because a quaternion (x1,x2,x3,x4) is equal to (-x1,-x2,-x3,-x4) we use the
     // absolute value of the scalar product
     double dot = q.quat[0]*quat[0]+q.quat[1]*quat[1]+q.quat[2]*quat[2]+q.quat[3]*quat[3];
--- a/src/Mod/Test/BaseTests.py
+++ b/src/Mod/Test/BaseTests.py
@@ -253,6 +253,35 @@ class AlgebraTestCase(unittest.TestCase):
         self.assertEqual(m2, m3*m4   , "Wrong multiplication order")
         self.assertNotEqual(m2, m4*m3, "Wrong multiplication order")
 
+
+    def testRotationFromMatrix(self):
+        m1 = FreeCAD.Matrix()
+        m1.rotateZ(.2)
+        m1.scale(0.5)
+        m1.rotateY(.2)
+        m1.scale(3)
+        m1.move(10,5,-3)
+        r1 = FreeCAD.Rotation(m1)
+        m2 = FreeCAD.Matrix()
+        m2.rotateZ(.2)
+        m2.rotateY(.2)
+        m2.move(10,5,-3)
+        r2 = FreeCAD.Rotation(m2)  
+        self.assertTrue(r1.isSame(r2, 1e-7), 'Scale on matrix influenced rotation')
+        m3 = FreeCAD.Matrix()
+        m3.scale(-1)
+        r3 = FreeCAD.Rotation(m3)
+        r0 = FreeCAD.Rotation()
+        self.assertTrue(r3.isSame(r0, 1e-7), 'Scale on matrix influenced rotation')
+        m4 = FreeCAD.Matrix()
+        m4.scale(1.25,1.0,0.25)
+        m4.move(4,5,6)
+        r24 = FreeCAD.Rotation(m2*m4)
+        r42 = FreeCAD.Rotation(m4*m2)
+        self.assertTrue(r2.isSame(r24, 1e-7), 'Scale on matrix influenced rotation')
+        self.assertTrue(r2.isSame(r42, 1e-7), 'Scale on matrix influenced rotation')
+
+        
     def testRotation(self):
         r=FreeCAD.Rotation(1,0,0,0) # 180 deg around (1,0,0)
         self.assertEqual(r.Axis, FreeCAD.Vector(1,0,0))
@@ -526,6 +555,11 @@ class MatrixTestCase(unittest.TestCase):
         self.mat.setRow(1, FreeCAD.Vector(1,2,3))
         self.mat.setRow(2, FreeCAD.Vector(1,2,3))
         self.assertEqual(self.mat.hasScale(), FreeCAD.ScaleType.Other)
+        self.mat.unity()
+        self.mat.scale(-1.)
+        self.assertEqual(self.mat.hasScale(), FreeCAD.ScaleType.Uniform)
+        self.mat.scale(-2.)
+        self.assertEqual(self.mat.hasScale(), FreeCAD.ScaleType.Uniform)
 
     def testShearing(self):
         self.mat.setRow(1, FreeCAD.Vector(0,1,1))


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [781298819](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/781298819) was triggered at [72cdaa7](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/72cdaa783fff2a85bb5d7ad89c002d7ad9135d32). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Jolbas:
Found some issues with this PR that I have to fix.

Jolbas:
False alarm. Made a mistake in my test.

chennes:
This looks like an absolutely ideal sort of code to test with a unit test, now that we've got Google Test incorporated into the main development branch. Are you at all familiar with that test framework?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [782165579](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/782165579) was triggered at [cb77cca](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/cb77cca600fcdee818a2fb8eaa585c1dd57da5ca). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Jolbas:
No, I'm not. Sorry. Just learned to make a python unittest.

chennes:
Ok, no problem, thanks for making the Python test. I might add a C++ test later.

src/Mod/Test/BaseTests.py
```diff
@@ -253,6 +253,22 @@ def testRotAndMoveMatrix(self):
...
         self.assertNotEqual(m2, m4*m3, "Wrong multiplication order")
 
+
+    def testRotAndScaleMatrix(self):
+        m1 = FreeCAD.Matrix()
+        m1.rotateZ(.2)
+        m1.scale(0.5)
+        m1.rotateY(.2)
+        m1.scale(3)
+        m1.move(10,5,-3)
```
chennes, line 263: Is the move a necessary component of the test?

Jolbas: There where cases I found during developing this fix where the move affected the output. The move is part of the matrix that is passed to the function so it surely could be the source for a future bug.

chennes:
Can I get another set of math-oriented eyes on this to make sure I'm not missing something? @0penBrain @AjinkyaDahale @abdullahtahiriyo ?

abdullahtahiriyo:
@chennes 
I am new to quaternions, so I doubt I can be of much help.

As I understand the problem in the linked issue, a determination has been made that the algorithm in isSame includes a simplification that is only true for uniform quaternions. Generally speaking there appears to be two possible solutions: (a) modify the algorithm so that it does not use the simplification when the quaternion is scaled, or (b) make the quaternion uniform.

If one can actually make a quaternion uniform by scaling as in the PR (which I ignore), the approach seems plausible to my untrained eye and the code seems ok. But I can only say this with the authority and confidence of a Cessna pilot set to land a 747 on a small airport. I may certainly be missing the whole picture. Sorry.

I know now that quaternions are used in physics, so perhaps Uwe has had more contact with the underlying math and understands its properties @donovaly

donovaly:
> I know now that quaternions are used in physics, so perhaps Uwe has had more contact with the underlying math and understands its properties @donovaly

I am a studied physicist but learned about quaternions first when I joined FreeCAD. So sorry.

@wwmayer , you know this well, could you please review this PR?

Jolbas:
The code is not tested on matrices representing combination of rotations and non uniform scaling of links. Maybe I should add a `this->normalize();` near the end as in `void Rotation::setValue(const double q[4])´ to ensure the quaternion is normalized.
As a double protection for the linked issue would be to replace the algorithm in `isSame()` with one that handles non normalized quaternions. There is one  mentioned in the code comment that seems to work.

Jolbas:
It didn't work if the uniform scale was negative. The `Matrix::hasScale()` couldn't handle negative uniform scale. I think I have fixed it. Also made the tolerance proportional to the scale to work better on very big or very small scale. 

Found out that there is no point trying to figure out anything else than uniform scale. The matrix get ambiguous as soon as there is both rotation and non uniform scale. The only consistent case is if there is no rotation and non uniform scale. So returning a basic norm rotation on every other case. Is that ok? Added tests for negative uniform scale too.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [782370423](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/782370423) was triggered at [56ade76](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/56ade76e7e0856d892eff1c544e85eb64848821e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Base/Rotation.cpp
```diff
@@ -220,32 +220,48 @@ void Rotation::setValue(const double q[4])
...
+    auto type = mc.hasScale();
+    if (type == Base::ScaleType::Uniform) {
+        // Scaled matrices results in a erroneus quaternion
+        // Removing scale make it work if scale is uniform
+        // or if there is no rotation
+        double d = mc.determinant3();
+        double s = 1.0/copysign(mc.getCol(0).Length(), d);
+        mc.scale(s, s, s);
+    } else if (type != Base::ScaleType::NoScaling) {
+        // Any other scale type are ambiguous about rotation
```
0penBrain, line 234: I'm not skilled enough to say if this is completely true or not. :)
But this is good way to clearly drop cases that we don't correctly handle.
This is necessary though IMO to throw an exception in this case so caller is informed something went wrong.

src/Base/Rotation.cpp
```diff
@@ -752,6 +768,8 @@ bool Rotation::isSame(const Rotation& q, double tol) const
     // This term can be simplified to
     // 2 - 2*(x1*y1 + ... + x4*y4) so that for the equality we have to check
     // 1 - tol/2 <= x1*y1 + ... + x4*y4
+    // This simplification only work if both quats are normalized
+    // Is it safe to assume that?
```
0penBrain, line 772: I didn't check again but AFAIR from working in Rotation.cpp yes, the quaternion are always intended to be normalized.

0penBrain (commented):
OK, had a quick review. Not an expert though. :)
The thing is that probably FreeCAD isn't first software trying to do so, and it would be cool to find an implementation handling all cases.
Did one has a look to https://github.com/blender/blender/blob/756538b4a117cb51a15e848fa6170143b6aafcd8/source/blender/blenlib/intern/math_rotation.c#L272-L396 for example ?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [782432786](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/782432786) was triggered at [559e563](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/559e56342751fab61113171a50fb327749a2eeb9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Base/Rotation.cpp
Jolbas: I try to filter out the cases when there is a uniform scale but no rotation and throw an exception for more complex cases.

src/Base/Matrix.cpp
```diff
@@ -832,37 +832,42 @@ ScaleType Matrix4D::hasScale(double tol) const
     // Only in case of uniform scaling it doesn't make a difference.
     if (tol == 0.0)
         tol = 1e-9;
+    
+    // compare absolute values propotional to size
+    auto close = [&](double a, double b) {
+        return fabs(fabs(a/b)-1.0) <= tol;
+    };
```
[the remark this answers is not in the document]
Jolbas: Ah. Good point. When the scale was very large the tolerance didn't work well. Maybe I can scale the tolerance instead. Or just go back to absolute compare.

Jolbas:
> Did one has a look to https://github.com/blender... for example ?
Good idea to look at their code but that code doesn't handle a matrix with scale either. Maybe it is more efficient but I don't have resources to figure that out easily.

src/Base/Rotation.cpp
Jolbas: Ok. I leave it as is then.

src/Base/Matrix.cpp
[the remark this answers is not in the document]
wwmayer: Another weakness of this expression is that for a=1 and b=-1 it returns true.

For a relative comparison have a look at: https://c-faq.com/fp/fpequal.html

src/Mod/Test/BaseTests.py
wwmayer: Now as we have set up the basic infrastructure for real unit tests there should also be written some tests in C++. See tests/src/Base/.

src/Base/Matrix.cpp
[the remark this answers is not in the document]
Jolbas: @wwmayer That was intentional, to make it return Uniform even if scale is negative.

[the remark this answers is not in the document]
wwmayer: OK, but then it might be good to choose a lambda name that makes it more obvious like closeAbsolute or so.

src/Base/Matrix.cpp
```diff
@@ -832,37 +832,42 @@ ScaleType Matrix4D::hasScale(double tol) const
...
         return ScaleType::NonUniformLeft;
     }
 
-    if (fabs(dxyz - d3) <= tol && (fabs(dx - dy) > tol || fabs(dy - dz) > tol)) {
+    if (close(dxyz, d3) && (!close(dx, dy) || !close(dy, dz))) {
         return ScaleType::NonUniformRight;
     }
 
-    if (fabs(dx - 1.0) > tol) {
+    if (abs(d3 - 1.0) > tol) {
```
wwmayer, line 868: Be careful with fabs vs abs (or std::abs). While for fabs it's guaranteed to work as expected for floating point numbers it depends with std::abs. There are several overloaded functions implemented and if not the right header is included the floating point numbers are cast to int and thus std::abs leads to unexpected results.
In this case it works because cmath is included by Vector3D.h

In the past I have read several times that fabs should be replaced with abs but according to https://en.cppreference.com/w/cpp/numeric/math/fabs it's still a  valid function.

Jolbas: Ops. I did read about that and was aware, but I did it anyway.

src/Base/Matrix.cpp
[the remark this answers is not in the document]
Jolbas: Sure

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [783360604](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/783360604) was triggered at [77786ca](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/77786ca71f7b0274fae0768149761143129b6420). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [783365574](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/783365574) was triggered at [a2cc9f0](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/a2cc9f03ae7993f7cb919dd004839bc18377cb27). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Base/Rotation.cpp
wwmayer: Obviously you tried to get the inverse scaling factors of a non-uniform scaling but then run into problems. Anyway there must be a way to also handle non-uniform scaling but it must be distinguished whether the scaling matrix has been multiplied from the left or the right side.

Here is an example in Python:
```
from math import copysign

# this is a pure rotation matrix
mat = App.Matrix()
mat.rotateX(1.234)
mat.rotateY(-4.321)
mat.rotateZ(1.423)

# this is a pure scaling matrix. The expected values for the inverse scaling are 1, 1/2 and 1/3
scale = App.Matrix()
scale.scale(1,2,3)

# multiply from left
mat1 = scale * mat
mat1.hasScale()

# and right
mat2 = mat * scale
mat2.hasScale()

# Scale type=NonUniformLeft:
# to get the inverse scaling factor one must use the rows of the matrix
# according to the changes of the first commit this would be
mat3 = App.Matrix(mat1)
dx = copysign(mat3.row(0).Length, mat3.A11)
dy = copysign(mat3.row(1).Length, mat3.A22)
dz = copysign(mat3.row(2).Length, mat3.A33)
mat3.scale(1/dx, 1/dy, 1/dz)
mat3.hasScale()

# mat3 now is indeed a pure rotation matrix which you can confirm with
mat3.col(0) * mat3.col(1) # = 0
mat3.col(0) * mat3.col(2) # = 0
mat3.col(1) * mat3.col(2) # = 0
mat3.determinant() # = 1

# But unfortunately it's a completely different rotation. This happens because dx=-1 and dz=-3 instead of the expected values of 1 and 3
If a scale factor is negative then it means it also does a mirroring and the question is whether this should be considered.

# Scale type=NonUniformRight:
mat4 = App.Matrix(mat2)
dx = copysign(mat4.col(0).Length, mat4.A11)
dy = copysign(mat4.col(1).Length, mat4.A22)
dz = copysign(mat4.col(2).Length, mat4.A33)
mat4.scale(1/dx, 1/dy, 1/dz)
mat4.hasScale()

```

src/Base/Matrix.cpp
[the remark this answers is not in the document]
wwmayer: I wonder what to do with negative scaling because effectively this is a (positive) scaling and a mirroring. If the  analysis gives a number of negative scaling factors of 2 then it can be considered as a positive scaling with two mirrorings. But two mirroring operations can also be considered as a rotation.

src/Base/Rotation.cpp
Jolbas: Yes. Obviously :)
Ok. I see. So it is possible to solve the rotation as long as the scale type is not `Other´. That doesn't seem too hard to do. I'll try to night.

Jolbas: @wwmayer I'm thinking though before I start. Is it really relevant to extract the rotation in those cases, in particular the NonUniformLeft case? Maybe better to create a resolve method in Matrix class to extract scale, pos, rotation, etc?

wwmayer: > Is it really relevant to extract the rotation in those cases, in particular the NonUniformLeft case?

If it's doable why not?

> Maybe better to create a resolve method in Matrix class to extract scale, pos, rotation, etc?

I agree. Doing it in the Rotation class is the very wrong place.

Your PR does it solve the issues you wanted to handle? If yes I can do a review of the current solution and once merged I can do the refactoring, extend the analysis to non-uniform scaling and write the C++ unit tests.

Jolbas: > If it's doable why not?

Afrer some more thought I agree

> Your PR does it solve the issues you wanted to handle? If yes I can do a review of the current solution and once merged I can do the refactoring, extend the analysis to non-uniform scaling and write the C++ unit tests.

If there is no rush I would really like to try the extended analysis again, but yes it solves the particular case from the forum thread.

wwmayer: OK, I can wait :)

wwmayer:
Yesterday I did another test of this PR: If the absolute scale factors are equal but have different signs e.g. (1, -1, 1) then the function claims that the scaling is uniform.

For a uniform scaling it's known that the matrix product is commutative, i.e. S * M is equal to M * S. But if S is not uniform then the product is not commutative. And if the diagonal of S = (1, -1, 1) then it can be easily seen that S * M and M * S give different results.

Jolbas:
@wwmayer 
 As I think you noted in the code snippet before there is no way to distinguish between scale(-1,1,1) and scale(-1,-1,-1) and a 180 degree rotation.

```
from math import pi
m1 = App.Matrix()
m1.scale(-1,-1,-1)
m1
m2 = App.Matrix()
m2.scale(-1,1,1)
m2.rotateX(pi)
m2
# m1 and m2 are (almost) equal

```
 I suggest we presume all scale factors have the same sign

```
mat3 = App.Matrix(mat1)
scale_sign = mat3.A11*mat3.A22*mat3.A33
sx = copysign(mat3.row(0).Length, scale_sign)
sy = copysign(mat3.row(1).Length, scale_sign)
sz = copysign(mat3.row(2).Length, scale_sign)
mat3.scale(1/dx, 1/dy, 1/dz)
if mat3.determinant() < 0:
    sx = -sx
    sy = -sy
    sz = -sz
    mat3.scale(-1)
mat3.hasScale()
'Detcted scale:', dx, dy, dz

```

wwmayer:
> As I think you noted in the code snippet before there is no way to distinguish between scale(-1,1,1) and scale(-1,-1,-1) and a 180 degree rotation.

Good point! I try to solve this apparent contradiction.

Of course, a scale(-1,1,1) is a uniform scaling in the sense that it doesn't change the relations of lengths of a geometry but mirrors it. If we have (-1,-1,-1) then any two mirroring can also be considered as a rotation around a coordinate axis by 180 degree. So, it's a rotation and a mirroring. The geometric meaning is uniform scaling.

But for the algebra it seems to make a difference. If you write this you can easily see that both results are the same (it's easy to prove this generally):
```
M=App.Matrix()
M.rotateX(1)
M.rotateY(1)
M.rotateZ(1)

S=App.Matrix()
S.scale(-1,-1,-1)

M * S
S * M
```

Now if you write this then you will see that the results are different:
```
M=App.Matrix()
M.rotateX(1)
M.rotateY(1)
M.rotateZ(1)

S=App.Matrix()
S.scale(-1,1,1)

M * S
S * M
```

So now let's try to solve this contradiction:
If we have a scale(-1,1,1) then it's equal to a scale(-1,-1,-1) and a rotation. Let's call them T and R. Because T is uniform we know that S = T * R = R * T.

Consider M * S:
M * S = M * (T * R) = M * (R * T) = (M * R) * T

Consider S * M:
S * M = (T * R) * M = T * (R * M)

So, in other words when multiplying T from left then another matrix is used than multiplying T from right.

Resume:
It's correct to handle equal absolute scale factors with different signs as uniform scaling.

Jolbas:
But from pratests it seems to hold for non uniform scales too. 

```
# pure rotation matrix
mat_r = App.Matrix()
mat_r.rotateX(pi/5)
mat_r.rotateY(-2*pi/3)
mat_r.rotateZ(pi/3)
App.Placement(mat_r)

# pure scaling matrix.
mat_s = App.Matrix()
mat_s.scale(1,2,-3)

# multiply from left
mat1 = mat_s * mat_r
mat1.hasScale()

# scale detect algorithm
mat3 = App.Matrix(mat1)
sx = mat3.row(0).Length
sy = mat3.row(1).Length
sz = mat3.row(2).Length
mat3.scale(1/sx, 1/sy, 1/sz)
if mat3.determinant() < 0:
    sx = -sx
    sy = -sy
    sz = -sz
    mat3.scale(-1)

# result
mat3.hasScale()
'Detcted scale:', sx, sy, sz

mat3.rotateZ(pi)
App.Placement(mat3)
App.Placement(mat_r)
# equal
```

T * R = R * T isn't true but it works anyway. At leas t on the scales and rotations I tried. But I can't prove it as you did.

wwmayer:
> T * R = R * T isn't true but it works anyway.

If T does uniform scaling and that's what I have demanded then T * R = R * T. In this context R doesn't need to be a rotation matrix but can be any matrix.
T is of the form
```
T =
|t 0 0|
|0 t 0|
|0 0 t|

and R =
|r11 r12 r13|
|r21 r22 r23|
|r31 r32 r33|

If you do T * R or R * T you can easily see that in both cases the product is
|t*r11 t*r12 t*r13|
|t*r21 t*r22 t*r23|
|t*r31 t*r32 t*r33|

which is equal to t * R             (t is a scalar in R) 
```

Jolbas:
I ended up with this algorithm to find a pure rotation matrix. Does it seem correct?

```
def findRotation(matrix):
    scale_type = matrix.hasScale()
    if scale_type == ScaleType.Other:
        raise ValueError
    m = App.Matrix(matrix)
    if scale_type == ScaleType.NonUniformRight:
        m.transpose()
    m.scale(*(1 / m.row(i).Length for i in range(3)))
    if scale_type == ScaleType.NonUniformRight:
        m.transpose()
    if m.determinant() < 0:
        m.scale(-1)
    return m
```

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [787496407](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/787496407) was triggered at [9fe98c9](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/9fe98c9ba33f8e78f12ba029d9fb46b5d3d1c84d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> Does it seem correct?

Yes, it looks good.

Jolbas:
> @Jolbas can you please squash the commits altogether ?

I don't know how to do that. I can close this PR and create a new one. Is that what you mean?

0penBrain:
> I don't know how to do that

Not lazyness, but best is too look on the web about "squashing git commits". ;) That's not so hard.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [788589184](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788589184) was triggered at [0ebccf7](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/0ebccf782cc7c4af9a87e9930d052fc300b68966). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [788594847](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788594847) was triggered at [ff510ce](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/ff510cec9ba826c6f0c4be90d773d736e8d2a93f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
```shell
git rebase -i HEAD~13
git push --force
```

Jolbas:
> Not lazyness, but best is too look on the web about "squashing git commits". ;) That's not so hard.

I tried to look on the web but the commits didn't squash. ;)

Actually it doesn't seem to work because I have repeatedly pressed the "update branch" button to merge changes from master.

>Here's how to squash commits + better to use rebase over using merge

Ah. Thanks. I did just find something similar. I think I managed to do it.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [788610999](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788610999) was triggered at [225c014](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/225c0142879c061a1cbd7a5727e28f4ff6eff83e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522">pipeline status</a> for feature branch [PR_8522](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8522). Pipeline [788619254](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788619254) was triggered at [bad52df](https://github.com/FreeCAD/FreeCAD/pull/8522/commits/bad52dfb6f6d0cd3f3dec1d76859689a9f7ec571). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #8530, opened 2023-02-18
[Sketcher] Command for B-spline by knot

### LINKED ISSUE #8529: [Feature Request] Construct B-spline by interpolation in Sketcher

### Forums discussion

https://forum.freecad.org/viewtopic.php?t=75950

### Subproject(s) affected?

Sketcher

### Idea description

It is desirable to have a tool in Sketcher that can draw a B-spline between user-definable points by directly providing said points (instead of by defining control points as is done presently).

The forum post describes two approaches to do this. This issue will track discussions around "direction 1".

### Anything else?

_No response_

### Code of Conduct

AjinkyaDahale:
Adds a new `DrawSketchHandler` for creating B-splines by interpolation. Each provided point is treated as a knot.

Fixes #8529.

This project is being funded by the Open Toolchain Foundation under the title "Open Toolchain Foundation - Curve drawing tool in Sketcher Workbench"

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
+++ b/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
@@ -984,10 +984,10 @@ Gui::Action * CmdSketcherCompCreateBSpline::createAction()
     periodicbspline->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline"));
 
     QAction* bsplinebyknot = pcAction->addAction(QString());
-    bsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_BSpline_By_Knot"));
+    bsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSplineByInterpolation"));
 
     QAction* periodicbsplinebyknot = pcAction->addAction(QString());
-    periodicbsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline_By_Knot"));
+    periodicbsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation"));
 
     _pcAction = pcAction;
     languageChange();
--- a/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
+++ b/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
@@ -951,35 +951,35 @@ bool CmdSketcherCreateBSplineByInterpolation::isActive()
     return isCommandActive(getActiveGuiDocument());
 }
 
-// /// @brief Macro that declares a new sketcher command class 'CmdSketcherCreatePeriodicBSplineByInterpolation'
-// DEF_STD_CMD_AU(CmdSketcherCreatePeriodicBSplineByInterpolation)
+/// @brief Macro that declares a new sketcher command class 'CmdSketcherCreatePeriodicBSplineByInterpolation'
+DEF_STD_CMD_AU(CmdSketcherCreatePeriodicBSplineByInterpolation)
 
-// CmdSketcherCreatePeriodicBSplineByInterpolation::CmdSketcherCreatePeriodicBSplineByInterpolation()
-// : Command("Sketcher_CreatePeriodicBSplineByInterpolation")
-// {
-//     sAppModule      = "Sketcher";
-//     sGroup          = "Sketcher";
-//     sMenuText       = QT_TR_NOOP("Interpolate periodic B-spline");
-//     sToolTipText    = QT_TR_NOOP("Create a periodic B-spline by interpolation, i.e. via knots in the sketch.");
-//     sWhatsThis      = "Sketcher_Create_Periodic_BSplineByInterpolation";
-//     sStatusTip      = sToolTipText;
-//     sPixmap         = "Sketcher_Create_Periodic_BSplineByInterpolation";
-//     sAccel          = "G, B, O"; // TODO: Finalize this. What about Periodic?
-//     eType           = ForEdit;
-// }
+CmdSketcherCreatePeriodicBSplineByInterpolation::CmdSketcherCreatePeriodicBSplineByInterpolation()
+: Command("Sketcher_CreatePeriodicBSplineByInterpolation")
+{
+    sAppModule      = "Sketcher";
+    sGroup          = "Sketcher";
+    sMenuText       = QT_TR_NOOP("Interpolate periodic B-spline");
+    sToolTipText    = QT_TR_NOOP("Create a periodic B-spline by interpolation, i.e. via knots in the sketch.");
+    sWhatsThis      = "Sketcher_Create_Periodic_BSplineByInterpolation";
+    sStatusTip      = sToolTipText;
+    sPixmap         = "Sketcher_Create_Periodic_BSplineByInterpolation";
+    sAccel          = "G, B, O"; // TODO: Finalize this. What about Periodic?
+    eType           = ForEdit;
+}
 
-// CONSTRUCTION_UPDATE_ACTION(CmdSketcherCreatePeriodicBSplineByInterpolation,"Sketcher_CreatePeriodicBSplineByInterpolation")
+CONSTRUCTION_UPDATE_ACTION(CmdSketcherCreatePeriodicBSplineByInterpolation,"Sketcher_CreatePeriodicBSplineByInterpolation")
 
-// void CmdSketcherCreatePeriodicBSplineByInterpolation::activated(int iMsg)
-// {
-//     Q_UNUSED(iMsg);
-//     ActivateHandler(getActiveGuiDocument(),new DrawSketchHandlerBSplineByInterpolation(1) );
-// }
+void CmdSketcherCreatePeriodicBSplineByInterpolation::activated(int iMsg)
+{
+    Q_UNUSED(iMsg);
+    ActivateHandler(getActiveGuiDocument(),new DrawSketchHandlerBSplineByInterpolation(1) );
+}
 
-// bool CmdSketcherCreatePeriodicBSplineByInterpolation::isActive()
-// {
-//     return isCommandActive(getActiveGuiDocument());
-// }
+bool CmdSketcherCreatePeriodicBSplineByInterpolation::isActive()
+{
+    return isCommandActive(getActiveGuiDocument());
+}
 
 
 /// @brief Macro that declares a new sketcher command class 'CmdSketcherCompCreateBSpline'
@@ -1015,9 +1015,9 @@ void CmdSketcherCompCreateBSpline::activated(int iMsg)
     else if (iMsg == 2) {
         ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(0));
     }
-    // else if (iMsg == 3) {
-    //     ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(1));
-    // }
+    else if (iMsg == 3) {
+        ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(1));
+    }
     else {
         return;
     }
@@ -1046,8 +1046,8 @@ Gui::Action * CmdSketcherCompCreateBSpline::createAction()
     QAction* bsplinebyknot = pcAction->addAction(QString());
     bsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSplineByInterpolation"));
 
-    // QAction* periodicbsplinebyknot = pcAction->addAction(QString());
-    // periodicbsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation"));
+    QAction* periodicbsplinebyknot = pcAction->addAction(QString());
+    periodicbsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation"));
 
     _pcAction = pcAction;
     languageChange();
@@ -1073,14 +1073,14 @@ void CmdSketcherCompCreateBSpline::updateAction(int mode)
             a[0]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSpline"));
             a[1]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline"));
             a[2]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSplineByInterpolation"));
-            // a[3]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation"));
+            a[3]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation"));
             getAction()->setIcon(a[index]->icon());
             break;
         case Construction:
             a[0]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSpline_Constr"));
             a[1]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline_Constr"));
             a[2]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSplineByInterpolation_Constr"));
-            // a[3]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation_Constr"));
+            a[3]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation_Constr"));
             getAction()->setIcon(a[index]->icon());
             break;
     }
@@ -1107,10 +1107,10 @@ void CmdSketcherCompCreateBSpline::languageChange()
     bsplinebyknot->setText(QApplication::translate("Sketcher_CreateBSplineByInterpolation","B-spline by knots"));
     bsplinebyknot->setToolTip(QApplication::translate("Sketcher_CreateBSplineByInterpolation","Create a B-spline by knots"));
     bsplinebyknot->setStatusTip(QApplication::translate("Sketcher_CreateBSplineByInterpolation","Create a B-spline by knots"));
-    // QAction* periodicbsplinebyknot = a[3];
-    // periodicbsplinebyknot->setText(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Periodic B-spline by knots"));
-    // periodicbsplinebyknot->setToolTip(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Create a periodic B-spline by knots"));
-    // periodicbsplinebyknot->setStatusTip(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Create a periodic B-spline by knots"));
+    QAction* periodicbsplinebyknot = a[3];
+    periodicbsplinebyknot->setText(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Periodic B-spline by knots"));
+    periodicbsplinebyknot->setToolTip(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Create a periodic B-spline by knots"));
+    periodicbsplinebyknot->setStatusTip(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Create a periodic B-spline by knots"));
 }
 
 bool CmdSketcherCompCreateBSpline::isActive()
@@ -2011,7 +2011,7 @@ void CreateSketcherCommandsCreateGeo()
     rcCmdMgr.addCommand(new CmdSketcherCreateBSpline());
     rcCmdMgr.addCommand(new CmdSketcherCreatePeriodicBSpline());
     rcCmdMgr.addCommand(new CmdSketcherCreateBSplineByInterpolation());
-    // rcCmdMgr.addCommand(new CmdSketcherCreatePeriodicBSplineByInterpolation());
+    rcCmdMgr.addCommand(new CmdSketcherCreatePeriodicBSplineByInterpolation());
     rcCmdMgr.addCommand(new CmdSketcherCompCreateBSpline());
     rcCmdMgr.addCommand(new CmdSketcherCreateLine());
     rcCmdMgr.addCommand(new CmdSketcherCreatePolyline());
--- a/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
+++ b/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
@@ -934,7 +934,7 @@ CmdSketcherCreateBSplineByInterpolation::CmdSketcherCreateBSplineByInterpolation
     sWhatsThis      = "Sketcher_CreateBSplineByInterpolation";
     sStatusTip      = sToolTipText;
     sPixmap         = "Sketcher_CreateBSplineByInterpolation";
-    sAccel          = "G, B, I"; // TODO: Finalize this. What about Periodic?
+    sAccel          = "G, B, I";
     eType           = ForEdit;
 }
 
@@ -964,7 +964,7 @@ CmdSketcherCreatePeriodicBSplineByInterpolation::CmdSketcherCreatePeriodicBSplin
     sWhatsThis      = "Sketcher_Create_Periodic_BSplineByInterpolation";
     sStatusTip      = sToolTipText;
     sPixmap         = "Sketcher_Create_Periodic_BSplineByInterpolation";
-    sAccel          = "G, B, O"; // TODO: Finalize this. What about Periodic?
+    sAccel          = "G, B, O";
     eType           = ForEdit;
 }
 
--- a/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
+++ b/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
@@ -921,6 +921,66 @@ bool CmdSketcherCreatePeriodicBSpline::isActive()
     return isCommandActive(getActiveGuiDocument());
 }
 
+/// @brief Macro that declares a new sketcher command class 'CmdSketcherCreateBSplineByInterpolation'
+DEF_STD_CMD_AU(CmdSketcherCreateBSplineByInterpolation)
+
+CmdSketcherCreateBSplineByInterpolation::CmdSketcherCreateBSplineByInterpolation()
+: Command("Sketcher_CreateBSplineByInterpolation")
+{
+    sAppModule      = "Sketcher";
+    sGroup          = "Sketcher";
+    sMenuText       = QT_TR_NOOP("Interpolate B-spline");
+    sToolTipText    = QT_TR_NOOP("Create a B-spline by interpolation, i.e. via knots in the sketch.");
+    sWhatsThis      = "Sketcher_CreateBSplineByInterpolation";
+    sStatusTip      = sToolTipText;
+    sPixmap         = "Sketcher_CreateBSplineByInterpolation";
+    sAccel          = "G, B, I"; // TODO: Finalize this. What about Periodic?
+    eType           = ForEdit;
+}
+
+CONSTRUCTION_UPDATE_ACTION(CmdSketcherCreateBSplineByInterpolation,"Sketcher_CreateBSplineByInterpolation")
+
+void CmdSketcherCreateBSplineByInterpolation::activated(int iMsg)
+{
+    Q_UNUSED(iMsg);
+    ActivateHandler(getActiveGuiDocument(),new DrawSketchHandlerBSplineByInterpolation(0) );
+}
+
+bool CmdSketcherCreateBSplineByInterpolation::isActive()
+{
+    return isCommandActive(getActiveGuiDocument());
+}
+
+// /// @brief Macro that declares a new sketcher command class 'CmdSketcherCreatePeriodicBSplineByInterpolation'
+// DEF_STD_CMD_AU(CmdSketcherCreatePeriodicBSplineByInterpolation)
+
+// CmdSketcherCreatePeriodicBSplineByInterpolation::CmdSketcherCreatePeriodicBSplineByInterpolation()
+// : Command("Sketcher_CreatePeriodicBSplineByInterpolation")
+// {
+//     sAppModule      = "Sketcher";
+//     sGroup          = "Sketcher";
+//     sMenuText       = QT_TR_NOOP("Interpolate periodic B-spline");
+//     sToolTipText    = QT_TR_NOOP("Create a periodic B-spline by interpolation, i.e. via knots in the sketch.");
+//     sWhatsThis      = "Sketcher_Create_Periodic_BSplineByInterpolation";
+//     sStatusTip      = sToolTipText;
+//     sPixmap         = "Sketcher_Create_Periodic_BSplineByInterpolation";
+//     sAccel          = "G, B, O"; // TODO: Finalize this. What about Periodic?
+//     eType           = ForEdit;
+// }
+
+// CONSTRUCTION_UPDATE_ACTION(CmdSketcherCreatePeriodicBSplineByInterpolation,"Sketcher_CreatePeriodicBSplineByInterpolation")
+
+// void CmdSketcherCreatePeriodicBSplineByInterpolation::activated(int iMsg)
+// {
+//     Q_UNUSED(iMsg);
+//     ActivateHandler(getActiveGuiDocument(),new DrawSketchHandlerBSplineByInterpolation(1) );
+// }
+
+// bool CmdSketcherCreatePeriodicBSplineByInterpolation::isActive()
+// {
+//     return isCommandActive(getActiveGuiDocument());
+// }
+
 
 /// @brief Macro that declares a new sketcher command class 'CmdSketcherCompCreateBSpline'
 DEF_STD_CMD_ACLU(CmdSketcherCompCreateBSpline)
@@ -955,9 +1015,9 @@ void CmdSketcherCompCreateBSpline::activated(int iMsg)
     else if (iMsg == 2) {
         ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(0));
     }
-    else if (iMsg == 3) {
-        ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(1));
-    }
+    // else if (iMsg == 3) {
+    //     ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(1));
+    // }
     else {
         return;
     }
@@ -986,8 +1046,8 @@ Gui::Action * CmdSketcherCompCreateBSpline::createAction()
     QAction* bsplinebyknot = pcAction->addAction(QString());
     bsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSplineByInterpolation"));
 
-    QAction* periodicbsplinebyknot = pcAction->addAction(QString());
-    periodicbsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation"));
+    // QAction* periodicbsplinebyknot = pcAction->addAction(QString());
+    // periodicbsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation"));
 
     _pcAction = pcAction;
     languageChange();
@@ -1012,11 +1072,15 @@ void CmdSketcherCompCreateBSpline::updateAction(int mode)
         case Normal:
             a[0]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSpline"));
             a[1]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline"));
+            a[2]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSplineByInterpolation"));
+            // a[3]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation"));
             getAction()->setIcon(a[index]->icon());
             break;
         case Construction:
             a[0]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSpline_Constr"));
             a[1]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline_Constr"));
+            a[2]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_CreateBSplineByInterpolation_Constr"));
+            // a[3]->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSplineByInterpolation_Constr"));
             getAction()->setIcon(a[index]->icon());
             break;
     }
@@ -1039,6 +1103,14 @@ void CmdSketcherCompCreateBSpline::languageChange()
     periodicbspline->setText(QApplication::translate("Sketcher_Create_Periodic_BSpline","Periodic B-spline by control points"));
     periodicbspline->setToolTip(QApplication::translate("Sketcher_Create_Periodic_BSpline","Create a periodic B-spline by control points"));
     periodicbspline->setStatusTip(QApplication::translate("Sketcher_Create_Periodic_BSpline","Create a periodic B-spline by control points"));
+    QAction* bsplinebyknot = a[2];
+    bsplinebyknot->setText(QApplication::translate("Sketcher_CreateBSplineByInterpolation","B-spline by knots"));
+    bsplinebyknot->setToolTip(QApplication::translate("Sketcher_CreateBSplineByInterpolation","Create a B-spline by knots"));
+    bsplinebyknot->setStatusTip(QApplication::translate("Sketcher_CreateBSplineByInterpolation","Create a B-spline by knots"));
+    // QAction* periodicbsplinebyknot = a[3];
+    // periodicbsplinebyknot->setText(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Periodic B-spline by knots"));
+    // periodicbsplinebyknot->setToolTip(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Create a periodic B-spline by knots"));
+    // periodicbsplinebyknot->setStatusTip(QApplication::translate("Sketcher_Create_Periodic_BSplineByInterpolation","Create a periodic B-spline by knots"));
 }
 
 bool CmdSketcherCompCreateBSpline::isActive()
@@ -1938,6 +2010,8 @@ void CreateSketcherCommandsCreateGeo()
     rcCmdMgr.addCommand(new CmdSketcherCreateArcOfParabola());
     rcCmdMgr.addCommand(new CmdSketcherCreateBSpline());
     rcCmdMgr.addCommand(new CmdSketcherCreatePeriodicBSpline());
+    rcCmdMgr.addCommand(new CmdSketcherCreateBSplineByInterpolation());
+    // rcCmdMgr.addCommand(new CmdSketcherCreatePeriodicBSplineByInterpolation());
     rcCmdMgr.addCommand(new CmdSketcherCompCreateBSpline());
     rcCmdMgr.addCommand(new CmdSketcherCreateLine());
     rcCmdMgr.addCommand(new CmdSketcherCreatePolyline());
--- a/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
+++ b/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
@@ -56,6 +56,7 @@
 #include "DrawSketchHandlerArcOfHyperbola.h"
 #include "DrawSketchHandlerArcOfParabola.h"
 #include "DrawSketchHandlerBSpline.h"
+#include "DrawSketchHandlerBSplineByInterpolation.h"
 #include "DrawSketchHandlerCarbonCopy.h"
 #include "DrawSketchHandlerCircle.h"
 #include "DrawSketchHandlerEllipse.h"
@@ -947,9 +948,14 @@ void CmdSketcherCompCreateBSpline::activated(int iMsg)
 {
     if (iMsg == 0) {
         ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSpline(iMsg));
-    } else if (iMsg == 1) {
+    }
+    else if (iMsg == 1) {
         ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSpline(iMsg));
-    } else {
+    }
+    else if (iMsg == 2) {
+        ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(0));
+    }
+    else {
         return;
     }
 
@@ -973,6 +979,9 @@ Gui::Action * CmdSketcherCompCreateBSpline::createAction()
 
     QAction* periodicbspline = pcAction->addAction(QString());
     periodicbspline->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline"));
+
+    QAction* bsplinebyknot = pcAction->addAction(QString());
+    bsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_BSplineByKnot"));
 
     _pcAction = pcAction;
     languageChange();
--- a/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
+++ b/src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
@@ -955,6 +955,9 @@ void CmdSketcherCompCreateBSpline::activated(int iMsg)
     else if (iMsg == 2) {
         ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(0));
     }
+    else if (iMsg == 3) {
+        ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSplineByInterpolation(1));
+    }
     else {
         return;
     }
@@ -981,7 +984,10 @@ Gui::Action * CmdSketcherCompCreateBSpline::createAction()
     periodicbspline->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline"));
 
     QAction* bsplinebyknot = pcAction->addAction(QString());
-    bsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_BSplineByKnot"));
+    bsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_BSpline_By_Knot"));
+
+    QAction* periodicbsplinebyknot = pcAction->addAction(QString());
+    periodicbsplinebyknot->setIcon(Gui::BitmapFactory().iconFromTheme("Sketcher_Create_Periodic_BSpline_By_Knot"));
 
     _pcAction = pcAction;
     languageChange();
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
@@ -201,7 +201,8 @@ public:
             SplineDegree = QInputDialog::getInt(
                 Gui::getMainWindow(),
                 QObject::tr("B-Spline Degree"),
-                QObject::tr("Define B-Spline Degree, between 1 and %1:")
+                QObject::tr("NOTE: For construction by interpolation, this degree is ignored for now.\n"
+                            "Define B-Spline Degree, between 1 and %1:")
                 .arg(QString::number(Geom_BSplineCurve::MaxDegree())),
                 SplineDegree, 1, Geom_BSplineCurve::MaxDegree(), 1);
             // FIXME: Pressing Esc here also finishes the B-Spline creation.
@@ -263,8 +264,7 @@ public:
                 return;
             }
         }
-        // TODO: On pressing, say, W, modify last pole's weight
-        // TODO: On pressing, say, M, modify next knot's multiplicity
+        // TODO: On pressing, say, M, modify the knot's multiplicity
 
         return;
     }
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
@@ -391,26 +391,33 @@ private:
 
             unsigned int myDegree = 3;
 
-            BSplineMults.front() = myDegree + 1; // FIXME: This is hardcoded until degree can be changed
-            BSplineMults.back() = myDegree + 1; // FIXME: This is hardcoded until degree can be changed
+            if (ConstrMethod == 0) {
+                BSplineMults.front() = myDegree + 1; // FIXME: This is hardcoded until degree can be changed
+                BSplineMults.back() = myDegree + 1; // FIXME: This is hardcoded until degree can be changed
+            }
 
             std::vector<std::stringstream> streams;
 
-            for (size_t i = 0; i < BSplineKnots.size(); ++i) {
-                if (!streams.empty())
-                    streams.back() << "App.Vector(" << BSplineKnots[i].x << "," << BSplineKnots[i].y << "),";
-                if (BSplineMults[i] >= myDegree && i != (BSplineKnots.size() - 1)) {
+            // The first point
+            streams.emplace_back();
+            streams.back() << "App.Vector(" << BSplineKnots.front().x << "," << BSplineKnots.front().y << "),";
+            // Middle points
+            for (size_t i = 1; i < BSplineKnots.size() - 1; ++i) {
+                streams.back() << "App.Vector(" << BSplineKnots[i].x << "," << BSplineKnots[i].y << "),";
+                if (BSplineMults[i] >= myDegree) {
                     streams.emplace_back();
                     streams.back() << "App.Vector(" << BSplineKnots[i].x << "," << BSplineKnots[i].y << "),";
                 }
             }
+            // The last point
+            streams.back() << "App.Vector(" << BSplineKnots.back().x << "," << BSplineKnots.back().y << "),";
 
             std::vector<std::string> controlpointses; // Gollum gollum
             controlpointses.reserve(streams.size());
             for (auto & stream: streams) {
                 // TODO: Use subset of points between C0 knots.
                 controlpointses.emplace_back(stream.str());
-
+
                 auto & controlpoints = controlpointses.back();
 
                 // remove last comma and add brackets
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
@@ -592,6 +592,26 @@ private:
                 tryAutoRecomputeIfNotSolve(
                     static_cast<Sketcher::SketchObject*>(sketchgui->getObject()));
 
+                ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath(
+                    "User parameter:BaseApp/Preferences/Mod/Sketcher");
+                bool continuousMode = hGrp->GetBool("ContinuousCreationMode", true);
+
+                if (continuousMode) {
+                    // This code enables the continuous creation mode.
+                    resetHandlerState();
+
+                    drawCursorToPosition(position);
+
+                    /* It is ok not to call to purgeHandler
+                     * in continuous creation mode because the
+                     * handler is destroyed by the quit() method on pressing the
+                     * right button of the mouse */
+                }
+                else {
+                    sketchgui
+                        ->purgeHandler();// no code after this line, Handler get deleted in ViewProvider
+                }
+
                 return false;
             }
 
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
@@ -97,6 +97,7 @@ public:
 
         if (Mode == STATUS_SEEK_FIRST_POINT) {
             BSplineKnots.push_back(onSketchPos);
+            BSplineMults.push_back(1); // NOTE: not strictly true for end-points
 
             Mode = STATUS_SEEK_ADDITIONAL_POINTS;
 
@@ -135,6 +136,7 @@ public:
         }
         else if (Mode == STATUS_SEEK_ADDITIONAL_POINTS) {
             BSplineKnots.push_back(onSketchPos);
+            BSplineMults.push_back(1); // NOTE: not strictly true for end-points
 
             // check if coincident with first knot
             for(auto & ac : sugConstr.back()) {
@@ -208,6 +210,21 @@ public:
             // FIXME: Pressing Esc here also finishes the B-Spline creation.
             // The user may only want to exit the dialog.
         }
+        if (SoKeyboardEvent::M == key && pressed) {
+            // TODO: On pressing, say, M, modify the knot's multiplicity
+            if (BSplineMults.size() > 1) {
+                BSplineMults.back() = QInputDialog::getInt(
+                    Gui::getMainWindow(),
+                    QObject::tr("Set knot multiplicity"),
+                    QObject::tr("NOTE 1: For construction by interpolation, the de facto maximum is currently 3.\n"
+                                "NOTE 2: Under certain circumstances (details WIP), this value may be ignored.\n"
+                                "Set knot multiplicity at the last point provided, between 1 and %1:")
+                    .arg(QString::number(SplineDegree)),
+                    BSplineMults.back(), 1, SplineDegree, 1);
+            }
+            // FIXME: Pressing Esc here also finishes the B-Spline creation.
+            // The user may only want to exit the dialog.
+        }
         // On pressing Backspace delete last knot
         else if (SoKeyboardEvent::BACKSPACE == key && pressed) {
             // when mouse is pressed we are in a transitional state so don't mess with it
@@ -264,7 +281,6 @@ public:
                 return;
             }
         }
-        // TODO: On pressing, say, M, modify the knot's multiplicity
 
         return;
     }
@@ -315,6 +331,7 @@ private:
         sugConstr.clear();
         knotGeoIds.clear();
         BSplineKnots.clear();
+        BSplineMults.clear();
 
         eraseEditCurve();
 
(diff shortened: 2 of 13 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [782079746](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/782079746) was triggered at [c95887e](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/c95887e3b1f20241f0d6cba78f4e2c0e4dbb9667). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
```diff
@@ -947,9 +947,14 @@ void CmdSketcherCompCreateBSpline::activated(int iMsg)
...
-    } else if (iMsg == 1) {
+    }
+    else if (iMsg == 1) {
         ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSpline(iMsg));
-    } else {
+    }
+    else if (iMsg == 2) {
+        ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerBSpline(iMsg));
+    }
+    else {
```
AjinkyaDahale, line 957: @abdullahtahiriyo I could use some input here. I can either extend `DrawSketchHandlerSpline` as I'm doing now, or create a new DSH especially for interpolation. Both have similarilities in how they need to be input, but there's a major difference that when defining poles we draw circles, and otherwise we draw vertices.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [782088517](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/782088517) was triggered at [63a4dcc](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/63a4dcc35ea5563feb1b53876a3a959162374463). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
abdullahtahiriyo: Both are possible. If the differences are significant, I think you would be better served by a new DSH. Each individual DSH will be simpler and easier to read and understand. Future developers reading the code will also be happier.

AjinkyaDahale: My only concern is that making a different class could result in code duplication. I guess we can work on that later, reducing the duplication rather than trying to split documents.

abdullahtahiriyo: It is always a trade-off. A DSH does not have that many parts. If the mouse-flow and representation is different and the command execution is different there is about nothing in common. But, yes, we can look to code duplication and what exactly to do about it when we know how much we are actually duplicating, if any.

As an example of what happens when a lot of adaptations of behaviour and command execution are implemented in a DSH, you may look at the lineset DSH (polyline). It is implemented as an all-in-one DSH. The complexity frightens developers eyeing it.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [782680918](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/782680918) was triggered at [d921776](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/d921776edc5111cbf97b0a843546cf7389a1bf59). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/CommandCreateGeo.cpp
abdullahtahiriyo: The current DSH framework offers support for different creation modes, which is the parameter provided on activation. If the input were the same, (for example an array of points, each point taken after one click) one could use a generic name for that array (e.g. inputPoints), without indicating if it is poles or interpolation points. Then it is possible to reuse all the input code. The command execution will be different. Two different member functions, called depending on the creation mode can be created and called appropriately. Small exceptions may be made for some differentiation between modes (e.g. less amount of steps,). But I think that is about the limit.

So, yes, we want to do code reuse in every instance where it makes sense. We want to keep separate things that change for different reasons (not unexpectedly introducing bugs on unrelated functionality while fixing a bug, just because it is tightly coupled despite being a separate thing). At the end, we want the code to be maintainable which is kind of the setting of the trade-off. We want to avoid duplication because duplicating creates problems to maintainability, we want to keep unrelated functionality separate, because mixing unrelated functionality creates problems to maintainability. There is where the skill of the developer enters into play ;)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [783360623](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/783360623) was triggered at [a664343](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/a664343f9dab3c9699ca892e5617bfa57453095a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [784271854](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/784271854) was triggered at [8c829c2](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/8c829c25c87f68a8199b7959be4a06504ce846ce). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
@abdullahtahiriyo Do you have time to make a review? I have a minimal product ready, which so far only utilizes existing OCCT functionality.

https://user-images.githubusercontent.com/4316933/220364571-64264bc9-de04-435f-a991-6a370a8c27c5.mp4

abdullahtahiriyo:
I hope to be able by the end of the week, I have to PRs I must deal with before that.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [786808629](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/786808629) was triggered at [1aee6b3](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/1aee6b30af5e0ea0858b1176e4aaa2a62cc9468c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,512 @@
...
+        prevCursorPosition = onSketchPos;
+
+        MousePressMode = MOUSE_PRESSED;
+
+        if (Mode == STATUS_SEEK_FIRST_POINT) {
+            BSplineKnots.push_back(onSketchPos);
+
+            Mode = STATUS_SEEK_ADDITIONAL_POINTS;
+
+            // insert circle point for knot, defer internal alignment constraining.
```
abdullahtahiriyo, line 103: circle?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,512 @@
...
+                return false;
+            }
+
+            //Gui::Command::commitCommand();
+
+            //static_cast<Sketcher::SketchObject *>(sketchgui->getObject())->solve();
+
+            // add auto constraints on knot
+            if (!sugConstr.back().empty()) {
+                createAutoConstraints(sugConstr.back(), knotGeoIds.back(), Sketcher::PointPos::mid, false);
```
abdullahtahiriyo, line 128: mid?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,512 @@
...
+                Gui::Command::abortCommand();
+
+                static_cast<Sketcher::SketchObject *>(sketchgui->getObject())->solve();
+
+                return false;
+            }
+
+            // add auto constraints on knot
+            if (!sugConstr.back().empty()) {
+                createAutoConstraints(sugConstr.back(), knotGeoIds.back(), Sketcher::PointPos::mid, false);
```
abdullahtahiriyo, line 179: mid?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: Will take a look. Most of these are discrepancies in the comments so they shouldn't have an effect on the result, but of course should be fixed to avoid confusion.

AjinkyaDahale: Reworded all mentions of circles and poles.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [786848232](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/786848232) was triggered at [23739d9](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/23739d9350ff37342c6a6a0c635166c7ca44a937). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
Ok. You were fast fixing the mid to start and the typos.

I did not do a detailed review. Just had a quick look and tested it.

The command delivers on its description. If interpolating per knot is what is intended, it is okeish. 

As it is now, weights are not constrained to non-rational, which leads to lots of DoFs. Even after constraining weights to non-rational, it leads to more DoFs than necessary for interpolation.

Aside from the issues of weights becoming near zero and lack of convergence associated to unconstrained poles and the excess of DoFs, it does what it promises by the title.

I ran the same example of interpolating through 5 fixed points, with this tool (result is 7 poles) and generating a normal 5 pole BSpline and constraining it point on object. The number of DoFs is lower with the second method. It is easier to work with, if interpolation is all that is needed.

I will give it another go later on or during the weekend.

Do you have a roadmap of how many different new construction methods you want to implement?

AjinkyaDahale:
> 



> Ok. You were fast fixing the mid to start and the typos.

Thanks. It was more of something I missed rather than a bug to hunt so it was straightforward.

> The command delivers on its description. If interpolating per knot is what is intended, it is okeish.

Indeed. OK-ish is an apt description. It delivers on the expectation, but the results might not be exactly what the user intended. `Part::BSplineCurve::interpolate` supports the providing parameters at the interpolated points, which can be used to get some better curves (say by making centripetal Catmull-Rom splines). 
 
> As it is now, weights are not constrained to non-rational, which leads to lots of DoFs. Even after constraining weights to non-rational, it leads to more DoFs than necessary for interpolation.

I believe there I once modified a command (in Sketcher App) that automatically applies the weight equality constraints. I'll dig into that and see if it can be used. Indeed there are more DoFs than necessary, and I don't know what is the typical method people employ for that in software with more mature spline support.

It would be restrictive to not allow all possible curves satisfying the constraints. I think the best idea here would be to take users' input once this is ready.

> I ran the same example of interpolating through 5 fixed points, with this tool (result is 7 poles) and generating a normal 5 pole BSpline and constraining it point on object. The number of DoFs is lower with the second method. It is easier to work with, if interpolation is all that is needed.

The problem here is that it is intended that the user only provide the interpolation points and the software take care from there. In that case, how do you suggest we differentiate between which of the provided points becomes knots and which become just point-on-object?

> Do you have a roadmap of how many different new construction methods you want to implement?

For this particular project I just intend to add this and a periodic version. The word of the project is technically satisfied, but the results are not fully satisfactory. Apart from the possibility of non-uniformity (i.e. unequal gaps between knot parameters), I also want to add the ability to specify the degree of continuity at every knot.

In the future, this construction method may also be able to allow specifying tangents, but that's outside the scope of the present PR.

abdullahtahiriyo:
@AjinkyaDahale 

Ok. Yes, I think we could benefit quite a lot from additional input from users. 

- We both understand that a fit can be done with a lesser amount of DoFs. 
- We both understand that not constraining poles is "dangerous", as convergence to near zero weights leads to unexpected behaviour. 
- It is unclear to both of us which are the user expectations, and to this respect whether more DoFs may be preferable in some (or majority) of applications or the contrary.

AjinkyaDahale:
Here's a list of tasks remaining. I've divided them into things that can be done with present OCCT functionality, and things that may need a new algorithm.

With OCCT functionality:
- [ ] Clamp the weights (say by equality constraints) so they don't collapse to zero.
- [ ] Create non-uniform B-splines
- [ ] Fix Periodic B-splines. They are currently possible but the results are not great.

New functionality needed:
- [ ] Let user choose degree.
- [ ] Let user choose multiplicity of each knot.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [788853139](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788853139) was triggered at [3eceedd](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/3eceedde2c0d4aeecacfd6995b232c8fe0bb571e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [788861225](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788861225) was triggered at [2bd191e](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/2bd191e7ce6ae6cd39af9aa0db31986ce7eb120e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Intuos5:
@AjinkyaDahale which user feedback are you loiking for in particular? Do you have any specific questions/ examples you could show? Please feel free to tag me for your spline work and I'll try to help as a user (don't know the particular msth on this though).

AjinkyaDahale:
@Intuos5 I'm looking for what the user's expectations are when using this tool. Some questions
- This tool is right now only useful to create a B-spline from scratch (as opposed to connecting two edges), but within that constraint is the curve exactly what you expected? Should it look any different?
- How often do you need other degrees, or discontinuities in the knots?
- We are leaving a lot of DoF's in the created B-spline; are you comfortable with that?

I can only provide source and snaps right now. Appimages would be great but right now only main branch has them.

Intuos5:
>     * This tool is right now only useful to create a B-spline from scratch (as opposed to connecting two edges), but within that constraint is the curve exactly what you expected? Should it look any different?
It looks as something I would expect. Could you provide a DXF export with the places where you clicked for the CVs (poles) and the spline itself? I could see how it differs from Rhino if that's of interest to you.
An inkscape like input method or Blendcurve would be helpful too, but that all depends on your scope of the project. 

>     * How often do you need other degrees, or discontinuities in the knots?
That depends mostly on the constraints in a sketch, so if I wers to make it tangency or curvature constrained, that would involve degree changes to make it fully constrained or to add a CV so I can adjust the middle portion of the spline. 
>     * We are leaving a lot of DoF's in the created B-spline; are you comfortable with that?
I don't think any parametric package out there enables you to change knot values and such (weighting), so I would rather have a slightly more restrictive spline and no DoF's than lots of DoF's as I use the number to inform how I go about constraining the sketch. I only see people change such things in Rhino, because you can, but I doubt most users are even aware that that's even possible. Nor do I see it's use in my personal workflow (adjusted weight splines look odd to me). 

> I can only provide source and snaps right now. Appimages would be great but right now only main branch has them.
I am on Windows, but can generally tell how it works by the look of it. You could export the files you make so I could cross-reference it with commercial apps. 

Hope this helps! If you have any further questions, please do not hesitate to tag me.

abdullahtahiriyo:
> Here's a list of tasks remaining. I've divided them into things that can be done with present OCCT functionality, and things that may need a new algorithm.
> 

Thank you.

> With OCCT functionality:
> 
>     * [ ]  Clamp the weights (say by equality constraints) so they don't collapse to zero.

Is the algorithm intended to produce rational B-Splines? 

If not, definitely one weight constrained to 1 and equality constraints would be a very nice improvement. It would also drive those DoFs down.

If the yes, then it may still be useful to constrain those circles, when all poles are different, then perhaps by fixing the weight of each.

> 
>     * [ ]  Create non-uniform B-splines

I must admit I am curious about what this delivers.

What are you eyeing, during creation edition (DSH level) or after creation?

> New functionality needed:
> 
>     * [ ]  Let user choose degree.

You mean that ```GeomAPI_Interpolate``` has this limitation? Which degree does it produce? (I browsed through the documentation, but it says nothing about the degree and I am not familiar with the function itself)

I am unsure about user requirements here. It would seem to me that deciding a degree during creation (DSH) would be difficult for users. I would expect users to "prefer" to be given a default which can later on be modified if the default does not match the outcome (for example by increasing the degree with the current tools). But, for sure, ask them, they use these tools.

> 
>     * [ ]  Let user choose multiplicity of each knot.

During creation? This one appears to me it may make more sense during creation, as it modifies the curvature and that is probably something users may have in mind. There must be more than one user that can comment on this.

AjinkyaDahale:
@Intuos5 
 
> Could you provide a DXF export with the places where you clicked for the CVs (poles) and the spline itself? I could see how it differs from Rhino if that's of interest to you.

I didn't save any particular list, so you're free to pick your own points and let me know which you used.

> > How often do you need other degrees, or discontinuities in the knots?
> 
> That depends mostly on the constraints in a sketch, so if I wers to make it tangency or curvature constrained, that would involve degree changes to make it fully constrained or to add a CV so I can adjust the middle portion of the spline.

So since you mentioned only tangency and curvature, you just need cubic B-splines?

> I don't think any parametric package out there enables you to change knot values and such (weighting), so I would rather have a slightly more restrictive spline and no DoF's than lots of DoF's as I use the number to inform how I go about constraining the sketch.

Noted.

> I am on Windows, but can generally tell how it works by the look of it. You could export the files you make so I could cross-reference it with commercial apps.

OK. We seriously need to look for test builds for all platforms.

AjinkyaDahale:
> > ```
> > * [ ]  Clamp the weights (say by equality constraints) so they don't collapse to zero.
> > ```
> 
> Is the algorithm intended to produce rational B-Splines?
> 
> If not, definitely one weight constrained to 1 and equality constraints would be a very nice improvement. It would also drive those DoFs down.

I believe it makes non-rational uniform B-splines. Is there a quick command that applies these constraints?

 
> > ```
> > * [ ]  Create non-uniform B-splines
> > ```
> 
> I must admit I am curious about what this delivers.
> 
> What are you eyeing, during creation edition (DSH level) or after creation?

I think after creation for the moment. In the future the spline itself should be visible (like arcs of conics are now), so you can judge if you need to edit the previously added points.
 
> > ```
> > * [ ]  Let user choose degree.
> > ```
> 
> You mean that `GeomAPI_Interpolate` has this limitation? Which degree does it produce? 

I believe it gives cubic splines. As far as I can tell, the `interpolate` method just gives the spline and then sketcher simply adds it.
 
> I am unsure about user requirements here. It would seem to me that deciding a degree during creation (DSH) would be difficult for users. I would expect users to "prefer" to be given a default which can later on be modified if the default does not match the outcome (for example by increasing the degree with the current tools). But, for sure, ask them, they use these tools.
> 
> > ```
> > * [ ]  Let user choose multiplicity of each knot.
> > ```
> 
> During creation? This one appears to me it may make more sense during creation, as it modifies the curvature and that is probably something users may have in mind. There must be more than one user that can comment on this.

Intuos5:
> So since you mentioned only tangency and curvature, you just need cubic B-splines?

Yes (and bezier curves).

And to come back to degrees and tools, there are two approaches I am aware of:
1. Offer the spline tool in-command options to change the degree prior to creation
2.. Give the user a degree 3 and a degree 5 option for the spline. 
Degree 3 is the default for tangency constraints and is what you need to be a close match for SubD geometry (the NURBS based version of Opensubdiv), which is mathematically only able to maintain curvature continuity with itself (it is degree 3 math).
Degree 5 is what you need for curvature continuity. 

1. Gives the user more flexibility. For instance, in Rhino, I set my default alias to access a spline to CV and degree 5. I can access any degree by typing C3, ..., C9. 
2. Simplifies the interface for more novice users. The downside, however, is that if you need to do some hefty degree changes, you'll need give more input than with option 1. 

Then there is another creation method that I haven't brought up which is a spline through points (an interpolate through the clicked points spline). In Rhino, this type of spline is converted back to a cubic B-spline after creation. You'll lose the points you clicked, but you keep the same type of spline. From the top of my head I don't know whether there is another type of spline that lets you interpolate through the clicked points and that is capable of being converted to a cubic B-spline later (by the user). For most people, this is much easier to draw, because it gives most local control (as opposed to the CV Hull based cubic B-spline, which changes shape slightly as the CV you manipulate also affects the portion of the spline at the next CV).

Intuos5:
> > New functionality needed:
> > ```
> > * [ ]  Let user choose degree.
> > ```
> 
> You mean that `GeomAPI_Interpolate` has this limitation? Which degree does it produce? (I browsed through the documentation, but it says nothing about the degree and I am not familiar with the function itself)
> 
> I am unsure about user requirements here. It would seem to me that deciding a degree during creation (DSH) would be difficult for users. I would expect users to "prefer" to be given a default which can later on be modified if the default does not match the outcome (for example by increasing the degree with the current tools). But, for sure, ask them, they use these tools.

Well, the downside to the current command is that it does not let you add a CV along the spline where you would need one to change it's shape. Yes, I know this alters the entire spline, but in some cases it is better to have the existing CVs stay put (e.g. to maintain tangency with other curves).
What it also doesn't let you do is increase the degree of the curve, but retain its shape. This is mostly used in surface modelling, where you try to keep the degree of splines to the bare minimum. If, however, you are unable to obtain the desired shape due to lack of CVs, you increase the degree and manipulate the CVs to get the desired shape. The advantage of this method is that you can clearly see what the additional vertex gives you in terms of control for the same (albeit approximated) spline. So you know whether or not the additional CV is worth the hassle. 

I think we already discussed this on the FreeCAD forum topic.

abdullahtahiriyo:
> I believe it makes non-rational uniform B-splines. Is there a quick command that applies these constraints?

Only as part of the creation to my knowledge (but I forget things). If the function is just necessary (i.e. you just need it) in Sketcher/Gui, you may add one free function in utils.h (take a look first).

> Indeed, I think most users will want cubic splines. There are apparently some cases were G3/C3 continuity is used, but we can wait till a user comes with this request. In any case, increasing degree is trivial, though the shape will end up modified if more smoothness is needed. Decreasing degree is NOT trivial, and OCCT leads to a whole lot of knots, so if someone wants 2 or 1 degree B-splines for whatever reason, that needs to be done at creation time.

Understood. Thanks.

> This is indeed the most sensible point in the list. I have some idea of how to get it done, but for now it seems a little hacky: apply fake tangency at C1 knots, and create multiple splines and join if C0.

If I am understanding you well, I would create a temporary BSpline at DSH level, containing all the user entered selection (knot multiplicity included). This temporary, I will use to paint it on screen in Edit Mode, then the user reacts to a real shape, so the DSH can assist the user to modify the multiplicity (or even come back to a previous piece connection) to change it. When the user is satisfied, we can finally create the last BSpline as usual.

I think I have coded this kind of things... yup, look in DrawSketchHandler:
```void drawEdit(const std::vector<Part::Geometry *> &geometries);```

You can throw it  some geometries, and it will draw them during edit mode.

Then, you also have PythonConverter to generate the commands (but B-Splines will likely not be supported yet).

See if anything of this works for you. Let me know.

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [791374952](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/791374952) was triggered at [142c923](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/142c923b63a322d9e9218bbaf9311afd4338b624). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [791571527](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/791571527) was triggered at [3f338c8](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/3f338c8134e39ba99cd2ff0c69a51c04a8551a68). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by AjinkyaDahale** [comment]
> > I believe it makes non-rational uniform B-splines. Is there a quick command that applies these constraints?
> 
> Only as part of the creation to my knowledge (but I forget things). If the function is just necessary (i.e. you just need it) in Sketcher/Gui, you may add one free function in utils.h (take a look first).

@abdullahtahiriyo 

Found it in `exposeInternalGeometry`. Just added a change small change here which could become a separate PR by itself. Basically constrains the first pole if it is created and let the rest of the mechanism take care. For rational splines making a clamp is also an option, but it won't apply retroactively anyway.

**Comment by abdullahtahiriyo** [comment]
@AjinkyaDahale 

With that change to ```exposeInternalGeometry```, if a user restores internal geometry (e.g. a first pole inadvertently deleted), it is going to restore a pole and constraint it with weight=1 (if the first pole has weight=1). So the change in behaviour is that before it was not constrained and now it is. I think we can leave with it, specially taking into account that I plan to remove the possibility of deleting internal geometry.

Do you currently have the case that the resulting B-Spline of an interpolation is rational?

**Comment by AjinkyaDahale** [comment]
> I think we can leave with it, specially taking into account that I plan to remove the possibility of deleting internal geometry.

@abdullahtahiriyo do you mean "live with it"? If so it's fine. When removing the possibility to delete internal geometry we'll also need to take care of the side effects of `exposeInternalGeometry`, and deal with the possibility of setting some geometries to internal, like we do in the DSH for B-spline.

> Do you currently have the case that the resulting B-Spline of an interpolation is rational?

I don't think that's the case at all. However, as we add more constraints while keeping the curvature continuity, we will need to work with more equations and so allow weights to change so we have more DOFs.

**Comment by abdullahtahiriyo** [comment]
Yup! typo.

> When removing the possibility to delete internal geometry we'll also need to take care of the side effects of `exposeInternalGeometry`, and deal with the possibility of setting some geometries to internal, like we do in the DSH for B-spline.

There should not be any side effects. This function checks what is already there and adds the rest. All the geometries are set to internal. Not sure what you mean.

**Comment by AjinkyaDahale** [comment]
> There should not be any side effects. This function checks what is already there and adds the rest. All the geometries are set to internal. Not sure what you mean.

The question is about the constraints on the poles. That is done in addition to creating the poles. Biggest issue is to disallow adding weight constraints when they are not desired.

**Comment by abdullahtahiriyo** [comment]
> The question is about the constraints on the poles. That is done in addition to creating the poles. Biggest issue is to disallow adding weight constraints when they are not desired.

- With the legacy code, if the initial pole is not constrained, nothing is constrained.
- With the modified code, the first pole is constrained to weight 1 if it is 1. The rest are constrained if the first is constrained and they are the same value.

If the legacy code is fed a rational B-Spline, it won't constrain any pole.
If the modified code is fed a rational B-Spline, it may at best constrain the first pole to weight 1.

If you see an issue with rational B-Splines, then perhaps we should not do the modification. 

An alternative is to create a new function which does not constrain at all and move there all the common code. The current function will then be refactored to call that other function and constrain (either legacy or modified).

The caller is responsible for calling one or the other, depending on what is to be achieved.

We may also make the change now, and if the problem arises, then refactor. Give it a thought.

**Comment by AjinkyaDahale** [comment]
> If you see an issue with rational B-Splines, then perhaps we should not do the modification.

> We may also make the change now, and if the problem arises, then refactor. Give it a thought.

Given the very specific conditions when the constraints may be applied undesirably, it seems making the changes now and then refactoring if needed is a good idea.

**Comment by AjinkyaDahale** [comment]
> Could you provide a DXF export with the places where you clicked for the CVs (poles) and the spline itself? I could see how it differs from Rhino if that's of interest to you.

@Intuos5 I see we didn't follow along this. Here are some points to test

```
[App.Vector(-36.0539,-8.32013),App.Vector(-24.6434,11.8067),App.Vector(10.3804,12.916),App.Vector(14.3423,-17.0364),App.Vector(-17.1949,-21.7908),App.Vector(-17.6704,-5.46751)]
```

Sorry I wasn't able to get you a DXF but this will open fine with any version of FreeCAD.

[test-spline.zip](https://github.com/FreeCAD/FreeCAD/files/10874327/test-spline.zip)

**Comment by Intuos5** [comment]
@AjinkyaDahale 
I did a test with Rhino. I have extracted the Spline's CV and drew a degree 3 and degree 5 spline in Rhino, this is the result:
afbeelding
It's quite a bit different than FreeCAD's (you can see this by checking the layer colours on the right).

Here's a curvature graph comparison of each. The curvature comb looks quite OK for this spline. But, you are using different math nonetheless.

https://user-images.githubusercontent.com/63149655/222894873-231055ac-5180-4c16-8319-c88be3684180.mp4

**Comment by AjinkyaDahale** [comment]
> @AjinkyaDahale I did a test with Rhino. I have extracted the Spline's CV and drew a degree 3 and degree 5 spline in Rhino, this is the result:...

Degree 5 B-spline will of course be different. With creating from the CV I am not surprised the curves are different, since apparently there is some play OCCT does to make the curves "better" (with some definition of that word). Could you try creating a cubic, C2 continuous (or curvature continuous if that setting doesn't exist) spline passing through the points I mentioned, viz:

> `[App.Vector(-36.0539,-8.32013),App.Vector(-24.6434,11.8067),App.Vector(10.3804,12.916),App.Vector(14.3423,-17.0364),App.Vector(-17.1949,-21.7908),App.Vector(-17.6704,-5.46751)]`

**Comment by AjinkyaDahale** [comment]
> ...Rhino's degree 5 spline is smoother looking in the curvature comb. But both FreeCAD's spline and Rhino's 3rd degree spline have continuity issues (they are multi-span after all). So being able to draw a 5th degree spline would already improve the curvature as you could see based on Rhino's 5th degree spline.

@Intuos5 There will always be continuity issues in a spline if you go to a high enough derivative. Degree 3 splines can only guarantee that the curvatures will be _continuous_, but not `differentiable (smooth)`, which is why there is a "corner" in the curvature graph. Discontinuity would mean the two lines do not meet at all, like so:
Screenshot from 2023-03-04 16-31-24


If you still need higher degree B-splines by interpolation, that was indeed one of the questions I asked and I'd be happy to try it some time. Unfortunately I haven't gotten to algorithm for higher order/degree splines right now (nor the time to implement it). That said, you can always create the 5-degree B-spline in freecad. For this:
1. Use the B-spline creation tool.
2. Press "D" and enter 5.
3. Provide the "CV" points (note they are called poles or control points here).

**Comment by Intuos5** [comment]
@AjinkyaDahale all splines use the same CVs (did IGES export) so it's not like I used different points for them. Rhino only offers additional knot parameterisation options for interpolation splines ('through point') , no C2/G2 options for either construction type, see: https://docs.mcneel.com/rhino/7/help/en-us/commands/curve.htm

Let me know if there's something else you want to see. Or any of the specific knot oarameterisation options (never touched them myself). 

Ah, I see cubic also refers to the 3rd  degree only. 😅 So in that case my earlier answer for Cubic B-splines should be supplemented with quintic B-splines 😉

**Comment by Intuos5** [comment]
> @Intuos5 There will always be continuity issues in a spline if you go to a high enough derivative. Degree 3 splines can only guarantee that the curvatures will be _continuous_, but not `differentiable (smooth)`, which is why there is a "corner" in the curvature graph. 

That's why I brought it up, so we could draw higher degree splines from within the same command that would be curvature continuous with itself (meaning: larger single span curves, CV wise). Rhino does this by default if the degree is high enough, I assume that's why it is not exposed to the user.

**Comment by AjinkyaDahale** [comment]
> @AjinkyaDahale all splines use the same CVs (did IGES export) so it's not like I used different points for them. Rhino only offers additional knot parameterisation options for interpolation splines ('through point') , no C2/G2 options for either construction type, see: https://docs.mcneel.com/rhino/7/help/en-us/commands/curve.htm

I'm not surprised. I presume when just providing poles the default is a uniform B-spline (equally placed knots).

> Let me know if there's something else you want to see. Or any of the specific knot oarameterisation options (never touched them myself).

I believe there should be a method to get the knot parametrization, but I can't find it right now. For now, could you just try using the "through point" splines and then using the aforementioned points, leaving all other options default?

> Ah, I see cubic also refers to the 3rd degree only, instead of 3 and up. sweat_smile So in that case my earlier answer for Cubic B-splines should be supplemented with quintic B-splines wink

Thanks. Though as I said, no promises yet :sweat_smile:.

**Comment by Intuos5** [comment]
Here's the spline text through points, you can see the creation specifics and the different splines by different knot values in the video:

https://user-images.githubusercontent.com/63149655/222954938-f9415e97-b9e4-45ce-8304-6ef27fa3f5a5.mp4

afbeelding

**Comment by AjinkyaDahale** [comment]
> @AjinkyaDahale Here's the spline text through points, you can see the creation specifics and the different splines by different knot values in the video:

Thanks! I was suspecting OCCT must be using the "sqrt(chord)" spacing, but apparently I was wrong. Or possibly there are still two more degrees of freedom and that's where the difference lies.

> If you would like to get an IGES file for any of the splines, just let me know.

Thanks for the offer. Maybe for the "sqrt (chord)" one?

**Comment by Intuos5** [comment]
@AjinkyaDahale here you go:
[sqrt (chord).zip](https://github.com/FreeCAD/FreeCAD/files/10891687/sqrt.chord.zip)

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [796777782](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/796777782) was triggered at [f016841](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/f0168415a0b1a7fae5e5315ece9f14b08f9db322). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [799010049](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/799010049) was triggered at [e2189cd](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/e2189cdd83f11d879f01e975af028f5c8ebac199). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [799509727](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/799509727) was triggered at [6365e18](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/6365e18340412f3dab7f70cf445cca2e46964e8e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [799889273](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/799889273) was triggered at [ba3b436](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/ba3b436b126e949eaffe930fd24f829074f3b441). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [799910893](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/799910893) was triggered at [1493577](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/1493577046df0e4fd937bdf04f815f751218965d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by AjinkyaDahale** [comment]
Some new features have been added.

1. Now, the B-spline created will be automatically constrained to be non-rational.
2. The periodic B-spline by interpolation tool is disabled for now.
3. The option to set the degree is disabled, only cubic B-splines will be created. 
4. There is now an option to set multiplicity of the last knot inserted. To do this, press "M" and a dialog box will appear to set the multiplicity of the last knot. Note that this value will be ignored under certain cases:
   * For the first and last points, the multiplicity is always 4.
   * If the points just before and just after have multiplicities >=3, the piece
between these two is fully continuous, and this (middle) point will only be
constrained with point-on-object.


https://user-images.githubusercontent.com/4316933/223747381-0ad61ea3-5975-4c17-9725-9b5415db8f89.mp4

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [801320049](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/801320049) was triggered at [7f24db0](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/7f24db075281359b1bcd6887b649ffc7342325e4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by AjinkyaDahale** [comment]
> @AjinkyaDahale
> 
> So, shall I understand by the change from Draft that you intend the merge?

Yes. I'm making some changes that shouldn't affect the [intended things](https://github.com/FreeCAD/FreeCAD/pull/8530#issuecomment-1460289121). Just don't test periodic splines (or beter yet please try to fix it :D).

**Comment by AjinkyaDahale** [comment]
Note to self: The issue with periodic B-splines is not unique to those created by interpolation.

To reproduce:
1. Create a periodic B-spline.
2. Increase the multiplicity of the first knot (currently red point) to make it C1.
3. Insert a knot in the first segment/piece.
Expected: Just that new knot is added, and all other knots are same.
Observed: A "knot" appears in the middle of the first two poles, which may not be on the curve at all.

image

My guess is that I kept a wrong assumption in the GCS solver, not accounting for non-uniform periodic B-splines under certain circumstances.

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [801541854](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/801541854) was triggered at [a1a0d07](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/a1a0d0747e2238cb9401524e59263e84971fe76e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by freecadci** [comment]
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [801774573](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/801774573) was triggered at [5cd1e0e](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/5cd1e0eb586ef31a6e954ede2a13bdf096468615). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

**Comment by abdullahtahiriyo** [comment]
@AjinkyaDahale 

Ok. Let's try to separate issues:
A) Working on this you have realised there is a bug that when knot separation gets non-uniform, at least sometimes, a knot is not properly constrained to the curve.
B) You are limiting the scope of this PR to non-periodic splines because this issue becomes much more apparent with the new functionality

If I understood it all correctly, I think (A) should be a separate PR. I think the decision (B) is a good one until (A) is fixed.

C) From what I see in the video, the result of changing the multiplicity is not apparent until the B-Spline is added to the Sketch. While editing, there is only the white line connecting the knots, which shows no information about the decisions made. This, I think reduces the value of this tool. If there were a way to show the effect, the added value would be higher. Perhaps, something could be achieved with the drawEdit as indicated here:
https://github.com/FreeCAD/FreeCAD/pull/8530#issuecomment-1445371802

That is also something that can be done in a separate PR if you so wish.

So. How do you want to proceed? Do you want the current code reviewed or should I defer it?

**Comment by AjinkyaDahale** [comment]
> So. How do you want to proceed? Do you want the current code reviewed or should I defer it?

@abdullahtahiriyo I believe I already fixed (A), but it's part of this PR (commit tagged with planegcs). I can make a quick PR that separates this if you want, but the fix is relatively simple so I think we can work with it.

I have removed the limitation in (B). Looking into the bug I realized it also affects non-periodic B-splines (if you set some knot to be C1), so now (B) should be unnecessary.

I think part (C) can be handled in a different PR altogether. This also affects our regular B-splines so we can handle them both together.

**Comment by abdullahtahiriyo** [comment]
@AjinkyaDahale 


Ok. Thanks for fixing (A) and (B).

> I think part (C) can be handled in a different PR altogether. This also affects our regular B-splines so we can handle them both together. 

Absolutely.

> EDIT: I managed to miss that you're talking about multiplicity change as well: for now we can add some text next to the points when in edit mode. Do you want me to put that?

No, it is not necessary. We both have understood that drawing the splines would be helpful and that another PR can take over that. I think that other PR would do the user a much better service than a quick text next to the points. We leave it like this.

Now this PR in on my roof (watch out if I drop a tool from up there :) )

File: src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h [hunk 5]
```diff
@@ -0,0 +1,606 @@
...
+                Gui::Command::abortCommand();
+
+                static_cast<Sketcher::SketchObject *>(sketchgui->getObject())->solve();
+
+                return false;
+            }
+
+            //Gui::Command::commitCommand();
+
+            //static_cast<Sketcher::SketchObject *>(sketchgui->getObject())->solve();
```
abdullahtahiriyo, line 125: Clean up this commented code?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,606 @@
...
+                if( ac.Type == Sketcher::Coincident && ac.GeoId == knotGeoIds[0] && ac.PosId == Sketcher::PointPos::start ) {
+                    IsClosed = true;
+                }
+            }
+
+            if (IsClosed) {
+                Mode = STATUS_CLOSE;
+
+                if (ConstrMethod == 1) { // if periodic we do not need the last pole
+                    // FIXME: Look at `interpolate` and check if we need to do this for knots. Periodic splines are as of writing not tested.
```
abdullahtahiriyo, line 152: Is this FIXME still true?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,606 @@
...
+        if (SoKeyboardEvent::M == key && pressed) {
+            // TODO: On pressing, say, M, modify the knot's multiplicity
+            if (BSplineMults.size() > 1) {
+                BSplineMults.back() = QInputDialog::getInt(
+                    Gui::getMainWindow(),
+                    QObject::tr("Set knot multiplicity"),
+                    QObject::tr("NOTE 1: For construction by interpolation, the de facto maximum is currently 3.\n"
+                                "NOTE 2: Under certain circumstances (details WIP), this value may be ignored.\n"
+                                "Set knot multiplicity at the last point provided, between 1 and %1:")
+                    .arg(QString::number(SplineDegree)),
```
abdullahtahiriyo, line 222: Consider if you really want these notes here.

Perhaps these notes are part of the documentation of the tool. The user can do nothing about it. In other words, the user will have to enter 1-3, because 3 is our degree and we cannot go beyond. So he will enter a number 1-3.  The explanations do not quite help the choice IMO.

If the user does not understand the tool, then the user will go to the documentation. There it can be explained more in detail that the maximum is the degree of the BSpline and that any change to the multiplicity of endpoints will be ignored. The latter is actually what most users would expect.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,606 @@
...
+        //         QObject::tr("NOTE: For construction by interpolation, this degree is ignored for now.\n"
+        //                     "Define B-Spline Degree, between 1 and %1:")
+        //         .arg(QString::number(Geom_BSplineCurve::MaxDegree())),
+        //         SplineDegree, 1, Geom_BSplineCurve::MaxDegree(), 1);
+        //     // FIXME: Pressing Esc here also finishes the B-Spline creation.
+        //     // The user may only want to exit the dialog.
+        // }
+        if (SoKeyboardEvent::M == key && pressed) {
+            // TODO: On pressing, say, M, modify the knot's multiplicity
+            if (BSplineMults.size() > 1) {
```
abdullahtahiriyo, line 215: remove TODO?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,606 @@
...
+
+    void eraseEditCurve()
+    {
+        drawEdit(std::vector<Base::Vector2d>());
+    }
+
+    bool finishCommand(Base::Vector2d position)
+    {
+        if (Mode==STATUS_CLOSE) {
+            // TODO: make modifications for construction by knots
```
abdullahtahiriyo, line 388: remove TODO?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,606 @@
...
+                streams.back() << "App.Vector(" << BSplineKnots[i].x << "," << BSplineKnots[i].y << "),";
+                if (BSplineMults[i] >= myDegree) {
+                    streams.emplace_back();
+                    streams.back() << "App.Vector(" << BSplineKnots[i].x << "," << BSplineKnots[i].y << "),";
+                }
+            }
+            // The last point
+            streams.back() << "App.Vector(" << BSplineKnots.back().x << "," << BSplineKnots.back().y << "),";
+
+            std::vector<std::string> controlpointses; // Gollum gollum
```
abdullahtahiriyo, line 415: Hope Smeagol does not jump on me. Any purpose for that comment?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,606 @@
...
+            std::vector<bool> isBetweenC0Points(BSplineKnots.size(), false);
+            for (size_t i = 1; i < BSplineKnots.size()-1; ++i) {
+                if (BSplineMults[i-1] >= myDegree && BSplineMults[i+1] >= myDegree)
+                    isBetweenC0Points[i] = true;
+            }
+
+            int currentgeoid = getHighestCurveIndex();
+
+            try {
+                //Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Add B-spline curve"));
```
abdullahtahiriyo, line 443: Leftover?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,606 @@
...
+                    else {
+                        // TODO: Adjust for periodic
+                        Gui::Command::runCommand(Gui::Command::Gui, "_finalbsp_poles.extend(_bsps[-1].getPoles()[1:])");
+                        Gui::Command::runCommand(Gui::Command::Gui, "_finalbsp_knots.extend([_finalbsp_knots[-1] + i for i in _bsps[-1].getKnots()[1:]])");
+                        Gui::Command::runCommand(Gui::Command::Gui, "_finalbsp_mults[-1] = 3"); // FIXME: Hardcoded
+                        Gui::Command::runCommand(Gui::Command::Gui, "_finalbsp_mults.extend(_bsps[-1].getMultiplicities()[1:])");
+                    }
+                }
+
+                // TODO: Join the pieces of the B-splines
```
abdullahtahiriyo, line 477: Could you just search for TODO in this file and check that those are actually TODOs at this stage?

abdullahtahiriyo (commented):
I work as expected in my tests.

Take look at the comments.

If you are in a position to execute clang-format version 15 or higher, please do it for the NEW files you have created (clang-format -i filename). The other files will eventually be changed. If you are not in a position, then leave it and they will be converted with the rest.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: It does remind the reader that this this is a vector of strings that represent lists. I can be more explicit here.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: Leftover. Though in the undo history it reads as "create knot" or something, like for the vanilla tool. Should do something about that.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: Sorry for tiring you out.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
```diff
@@ -0,0 +1,606 @@
...
+                // {"poles", "mults", "knots", "periodic", "degree", "weights", "CheckRational", NULL};
+                Gui::cmdAppObjectArgs(sketchgui->getObject(), "addGeometry(Part.BSplineCurve"
+                                      "(_finalbsp_poles,_finalbsp_mults,_finalbsp_knots,%s,%d,None,False),%s)",
+                                      ConstrMethod == 0 ?"False":"True",
+                                      myDegree,
+                                      geometryCreationMode==Construction?"True":"False");
+                currentgeoid++;
+
+                // TODO: Confirm we do not need to delete individual elements
+                Gui::Command::runCommand(Gui::Command::Gui, "del(_bsps)\n");
```
AjinkyaDahale, line 487: @abdullahtahiriyo I think I knot the answer already, but can you confirm if deleting just `_bsps` deletes each element of `_bsps`?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: Needs some work to answer for sure. I'll get back over the weekend.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: Oh! that is interesting. Where does it happen other than endpoints? (as this could help looking at a better message)

Ok, let's try to look into alternatives:
"Set knot multiplicity at the last entered point, between 1 and %1."
"Note that multiplicity will be ignored for endpoints, which will have %1 multiplicity"

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: It must be a reference to something I ignore. Gollum gollum says nothing to me. If I try hard I can imagine the array trying to rule all other arrays in FreeCAD.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: Please see https://github.com/FreeCAD/FreeCAD/pull/8530#issuecomment-1460289121 for examples. For periodic B-splines there is also an added complication that the first knot is multiplicity 2 (unnecessary in my opinion, but implemented in OCCT). Lots to test here. Something like "Note that multiplicity may be ignored under certain circumstances. Please refer to documentation for details".

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: Just that Gollum often makes plural of plurals, "ringses".

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: I have checked that undo works in my tests, but I did not look into the string itself. It is always good to be expressive in the naming of the undo history, but be practical. 

If the problem you are facing is that you have created a command for the knot that then you would like to include the B-Spline creation in it, perhaps we should look into extending the framework to enable renaming of an open not yet committed transaction (it may even be possible and me not knowing it).

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: No problem. You are methodical and that is very good. I am just doing a review and try to think the information the future developer reading the code needs. The TODOs that are yet to be done must remain. The TODOS that are done will be confusing for that future developer.

AjinkyaDahale: TODOs no longer necessary are removed.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: The problem is that if once the spline is added, the undo history still only says "add a knot", instead of "add a B-spline". Ideally, when editing, undo should remove knots, but when editing is done, undo should remove the whole B-spline.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: Please, think of the our dear new Developer from Antarctica that is not familiar with this kind of references. She is just down there at -60 Celsius  trying to understand what you wrote ;)

AjinkyaDahale: OK I'll change it. Who is this legend though?

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: My understanding is that del will decrease the reference counter of each element. In this case there is no other variable name referencing those element, then python will delete the elements.

So it would be a problem if you had:

```
mycoolbspline = Part.BSplineCurve...
mycoolarray = [mycoolbspline]
del(mycoolarray)
```

In this case, if you do not delete mycoolbspline that object remains in memory.

But do not hold me for a Python expert. I may be mistaken.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: Ok. I agree.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: Didn't you meet her yet? LoL

AjinkyaDahale: Not really.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: Thought so too: this is as far as my knowledge stands. Does any of this change for our use case where there is some C++ involved as well? I'd imagine not, but you never know. Anyone we can ask?

abdullahtahiriyo: The c++ here is just calling the python. The objects are created in python as if you typed the code in the console. The fact that c++ code is writing to the python console is irrelevant.

Where it may be an issue is when you are writing the python bindings for a c++ object, then you need to decide whether the Python object takes a copy of the c++ object or not and then also decide on the ownership (we got leaks in the past there as it is indeed difficult to follow). But that is totally different scenario.

If you are unconvinced, feel free to post in the developers corner.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
abdullahtahiriyo: She is very friendly and a real team player. She know her c++ too. I am sure you will like her.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: OK. I think we should be fine so long as the C++ side itself doesn't create dangling objects.

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: Introduce us, then. Half expecting this person to actually be a penguin and good at sea++.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [805629123](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/805629123) was triggered at [8fd5d34](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/8fd5d342322b8bd7e85de04039271666f19c9f6d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/DrawSketchHandlerBSplineByInterpolation.h
AjinkyaDahale: removed

abdullahtahiriyo:
I hope to be able to do a final review over the weekend.

abdullahtahiriyo:
@AjinkyaDahale 

Testing I have come to some weird behaviour.

For the periodic B-Spline, if I try to interpolate two straight lines:
1- Click first line point
2- Click second line point
3- Click third line point.

Then it works fine:
image

If, however, I continue:
4- Click on the first point again

The command fails, the DSH becomes in some dangling state in which you cannot cancel the DSH with the mouse. You may cancel it clicking another DSH (another tool).

Would you take a look?

AjinkyaDahale:
@abdullahtahiriyo I'll take a look once I have some time this evening. Though this kind of failure probably needs some better management. There have been plenty of corner cases (which I believe should be fixed now) that result im the final behaviour as you mentioned.

AjinkyaDahale:
> Testing I have come to some weird behaviour.
...
> If, however, I continue: 4- Click on the first point again
> Would you take a look?

I think I have an idea of what maybe happening: somehow the last point may not be deleted, and that causes two very-close points in the interpolation. OCCT doesn't like this and causes an issue.

AjinkyaDahale:
@abdullahtahiriyo OK I know where the issue is: in the try-catch in `finishCommand`, the continuous mode logic is not triggered if an exception is caught (we just return `false`). In my opinion this should be solved by following the "open-closed" principle where the parent class handles the main logic, so that new DSH's cannot suddenly break things this way.

AjinkyaDahale:
@abdullahtahiriyo the issue should be fixed now: the cryptic error is still there, but the command should move on. However, there is another issue and I don't know if it's related to this DSH or the general UI.

To reproduce:
0. Create two line segments as before (I suspect one will do)
1. Begin creating periodic B-spline starting with one of the points of a segment as starting.
2. Click the other points and end with a click on the first point.
3. Right click to (try to) create B-spline
4. The creation will fail. Right click again and try moving the segments.

Result: The end vertices of the line-segments disappear.

abdullahtahiriyo:
> Result: The end vertices of the line-segments disappear.

With tricks like that one you may consider a side-job as illusionist :-D 

I have just tested with your latest commits and I cannot reproduce it. I cannot reproduce the previous issue either (you have fixed that one).

Can you still reproduce it?

abdullahtahiriyo:
I mean, I when not closing he endpoint it works fine:
[Periodic_Not_Last_Point_Closing_Works.webm](https://user-images.githubusercontent.com/5249162/226199257-65641867-87c5-48bb-bbd0-2358673aa80d.webm)

When, closing the last endpoint it does not work, it does not break big time like before,no magic making lines disappear:
[Periodic_Last_Closing_Not_working.webm](https://user-images.githubusercontent.com/5249162/226199339-5071de4e-456b-4581-ad0c-74f925e2a7dd.webm)

The thing is that many people will close periodic B-Splines by clicking in the initial point, as in this exercise...

AjinkyaDahale:
> With tricks like that one you may consider a side-job as illusionist :-D

I was just watching a video about multiple income streams. Maybe I just might if I figure out how to reliably produce it.
 
> Can you still reproduce it?

Not always. But it's difficult to tell when. Usually when this happens only one phrase rings in my head: race conditions. But as far as I know there's no parallelism in the drawing.

AjinkyaDahale:
> The thing is that many people will close periodic B-Splines by clicking in the initial point, as in this exercise...

Sounds like a relatively simple fix of checking for coincidences. Some enterprising people could actually just choose to click on the previous point twice and cause the same issue.

abdullahtahiriyo:
> Not always. But it's difficult to tell when. Usually when this happens only one phrase rings in my head: race conditions. But as far as I know there's no parallelism in the drawing.

We mostly do not have those. We get not initialised values. Dangling pointers usually end up in a crash... 

There the whole geometry coin layer is gone. Only the Edit line stays. Perhaps it could be explained by the combination of "malfuntion of the geometry creation/bspline interpolation causing the geometry property to end up empty" + "the DSH entered an out of control state and the edit mode curve is not cleared (because it did not finish)".

Perhaps you could try to get the geometry property in the Python console when this happens, and see if it is empty. If so, then we have bigger problems. I am not in a good position to help you not being able to reproduce it...

sliptonic:
@abdullahtahiriyo  I'm not qualified to comment on the technical details of this PR (not by a long shot).  My comment is more about our overall PR process.

While the feature isn't perfect, is it stable enough that we can merge this and create a new issue for the outstanding corner-cases?  I would like to see us moving to a faster and more iterative cadence where we get features out in the wild (or at least the development master branch) and get feedback sooner so we can improve it.

What are your thoughts?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [812838321](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/812838321) was triggered at [fc225e3](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/fc225e3f3a7aad7d2f6009fdbf22bafc3a425a30). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
@sliptonic 

There are two issues with this PR:
- The coincidence clicking to close B-Splines.
- The ghost issue with lines disappearing.

For the first, @AjinkyaDahale  indicates that it is an easy case to check coincidences. I think this one is important because that is what many users will do. I see he has pushed some commits, but I did not get feedback from him. I do not know if the commits address the issue or not. This issue is stopping the merge because the feature is not stable.

For the second one, I cannot reproduce and I won't stop the PR for something I cannot reproduce. It was Ajinkya raising the issue. I just tried to help him out to track the cause.

I also see that one CI (Ubuntu 22.04) has failed (but I have not checked exactly why).

Do we agree?

AjinkyaDahale:
> There are two issues with this PR:
> 
>     * The coincidence clicking to close B-Splines.
> ...
> For the first, @AjinkyaDahale indicates that it is an easy case to check coincidences. I think this one is important because that is what many users will do. I see he has pushed some commits, but I did not get feedback from him. I do not know if the commits address the issue or not. This issue is stopping the merge because the feature is not stable.

By "easy", here I meant the solution seems known. However, there may be implementation level detail as I try to make it happen. 

The commit I pushed only intends to clear the DSH to fail this construction and go forward if continuous mode is on (or end otherwise). Since it reliably fails and doesn't cause other issues I think we can go forward with the merge (but create an issue to keep track).

Another reason I think the solution should be a separate PR is that this is also possible in the regular B-spline (where we specify control points). Simply repeat your steps with the periodic B-spline tool instead of Periodic B-spline by interpolation tool. When selecting the (intended) last point, the DSH will not close the spline and keep seeking next points. This does not cause error reports but it's still not intended behaviour.

Coming to the second issue, I think that's yet another deeper issue, since I tried just making the two points, without invoking the B-spline tool, and the vertices still (unreliably) disappeared.

AjinkyaDahale:
> I cannot compile the branch right now to check what we have.
> 
> If that is fixed and you gentlemen insist, we could go a two steps approach here.

Looked into it and the solution indeed was straightforward. With the latest commit periodic B-splines should automatically close.

> I do not think the two situations are the same.

We are using slightly different definitions of "same" here. The behaviour is same to the extent that the closure does not happen. As you can see in the latest commit, the solutions also are the same. The only difference is in the response from OCCT after the command eventually reaches it, which you could rightly argue reflects differently to the user. Do note that in case of the regular tool, the resultant B-spline has repeat control points (which are coincident) instead of the last point being deleted.

> > Coming to the second issue, I think that's yet another deeper issue, since I tried just making the two points, without invoking the B-spline tool, and the vertices still (unreliably) disappeared.
> 
> Then this is for sure a separate issue. Hopefully somebody who can reproduce will debug and come with a solution.

I'll make a issue then.
 
> Could you check the error of the CI?
From what I could tell it's from a test failure not related to this PR. In [another branch in a fork](https://github.com/Ondsel-Development/FreeCAD/actions/runs/4479891724/jobs/7874459570) also the test failed, but the same Branch succeeded in [its PR](https://github.com/FreeCAD/FreeCAD/actions/runs/4479892266/jobs/7874461740?pr=8923).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530">pipeline status</a> for feature branch [PR_8530](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8530). Pipeline [813413127](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/813413127) was triggered at [5d760a4](https://github.com/FreeCAD/FreeCAD/pull/8530/commits/5d760a4619fa53fe60196797e758ade1d5fffd18). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
> Do we agree?

You guys are all over it and doing an awesome job!  Thank you.

I will keep pushing for faster iteration in the future.  It's unfortunate that we've caused people to rely on the development branch being stable.  It's making us be too tentative about merging things.  Faster iteration solves a lot of our problems and creates very few new ones.

AjinkyaDahale:
@abdullahtahiriyo disappearing vertices issue reported: #8978.

abdullahtahiriyo:
> I will keep pushing for faster iteration in the future. It's unfortunate that we've caused people to rely on the development branch being stable. It's making us be too tentative about merging things. Faster iteration solves a lot of our problems and creates very few new ones.

Brad, here we have diverging views. 

I think it is very fortunate that people rely on master being reasonably stable. Because this ensures we get timely feedback. If master became unstable, (a lot of) people would stop using it and a lot of feedback would only come from (yearly) releases. Intensive testing of master aside, within release cycle corrections are vital for FreeCAD success.

New features are shaped by the fast feedback only because there is feedback. There is feedback because a substantial amount of users understand they can rely on FreeCAD master being reasonably stable.

Of course there is a balance between quality and merge rate. As often in life, virtue lies on the right balance.

Specific of this PR, Ajinkya generally produces good quality code. Some corner cases with unacceptable behaviour were detected. Other unrelated issues were detected. We (mostly Ajinkya) separated ones from the others. Fixed the ones relating to this feature (I presume). And as soon as I  have time (will definitely happen before the weekend), I will test and merge.

Some guidelines remain (this has nothing to do with this PR, it is just general information):
- A PR with well written code has much better chances of being merged early. It also puts a lower burden on the maintainer. For this, I recommend that every contributor critically reads their own code before making the PR.
- A well tested PR has much better chances of being merged early. It also puts a lower burden on the maintainer. For this, the contributor should have some minimal testing in place. Perhaps builds can be given for volunteer testers for some contributions as needed.

My approach may be perceived as egoistic as it leans toward maintainer time. Reality is that what I want is to merge right away. Review code once, test once and hit the button. This leaves more time available for me to review other PRs and to code my own ideas.

This does not mean I do no want to be consulted early if the developer sees impeding difficulties or have doubts about the soundness of the architecture being used. I should be consulted early to avoid loss to effort and frustration. It means that, when the situation is balanced, I can make an optimum allocation of resources, and often, life allowing, I can be reasonably responsive to all.

For times were I cannot be responsive, and other maintainers are not available or prefer to defer to me due to the nature of the PR, I would propose that contributors run development of different features in parallel, so that they are not stopped waiting. If the features are of different WBs, it may be better for diversification, but also several features of a single WB could work.

PR #7255, opened 2022-07-25
Part: better quantity values handling for primitives

0penBrain:
Following PR #7219

[Bugfix] : fixes bug declared here : https://forum.freecadweb.org/viewtopic.php?f=3&t=73321

PR #8565, opened 2023-02-20
[Bugfix] Gui: fix bug in expression completer tokenizing

0penBrain:
https://forum.freecad.org/viewtopic.php?t=76239

The problem is that in the tuple returned by `tokenizeExpression`, the std::string is converted to QString, but the character position is still stored as computed in the std::string. This breaks when using UTF-8 chars as their representation size isn't same in std::string and QString.
The PR fixes the problem by recomputing the character position in the QString representation.

### DIFF
--- a/src/Gui/ExpressionCompleter.cpp
+++ b/src/Gui/ExpressionCompleter.cpp
@@ -820,9 +820,12 @@ void ExpressionCompleter::slotUpdate(const QString & prefix, int pos)
         std::transform(result.cbegin(),
                        result.cend(),
                        std::back_inserter(tokens),
-                       [](const std::tuple<int, int, std::string>& item) {
+                       [&](const std::tuple<int, int, std::string>& item) {
                            return std::make_tuple(
-                               get<0>(item), get<1>(item), QString::fromStdString(get<2>(item)));
+                                get<0>(item),
+                                QString::fromStdString(expr.toStdString().substr(0,get<1>(item))).size(),
+                                QString::fromStdString(get<2>(item))
+                                );
                        });
         return tokens;
     };
@@ -883,7 +886,9 @@ void ExpressionCompleter::slotUpdate(const QString & prefix, int pos)
     }
 
     // Not an unclosed string and the last character is a space
-    if (!stringing && !prefix.isEmpty() && prefix[prefixEnd-1] == QChar(32)) {
+    if (!stringing && !prefix.isEmpty() &&
+            prefixEnd > 0 && prefixEnd <= prefix.size() &&
+            prefix[prefixEnd-1] == QChar(32)) {
         if (auto itemView = popup())
             itemView->setVisible(false);
         return;


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8565">pipeline status</a> for feature branch [PR_8565](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8565). Pipeline [783494557](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/783494557) was triggered at [338f23a](https://github.com/FreeCAD/FreeCAD/pull/8565/commits/338f23a60a7bab55d5c28096f2885af9c8a13218). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer (changes requested):
Your PR fixes the reported problem.

When testing current master (i.e. without your fix) the application even crashes for me. This happens in line 886 because of this incorrect check:
`if (!stringing && prefix.isEmpty() && prefix[prefixEnd-1] == QChar(32))`

So, it would be good to add this change to your PR too:
`if (!stringing && prefix.size() >= prefixEnd && prefix[prefixEnd-1] == QChar(32))`

In order to later find the reasons of a change it would be good to add the forum link as comment into the code.

Another point -- but this would be too much for this PR -- is to refactor the function slotUpdate() and move the whole tokenizing logic to its own class. This class could be added to FreeCADApp and would allow it write some unit tests.

0penBrain:
> When testing current master (i.e. without your fix) the application even crashes for me.

Yes, I guess because (as I do), you're using a debug build. In this case Qt has a strong assert about not iterating over the QString size. In Release builds it only fails silently.

> So, it would be good to add this change to your PR too: `if (!stringing && prefix.size() >= prefixEnd && prefix[prefixEnd-1] == QChar(32))`

I hesitated but at the end thought it doesn't add anything useful. Qt already checks that and, for example in this case, it was actually very useful that it wasn't implemented. Made the bug easy to find. :) But if you think it's needed, no problem to add it.

> In order to later find the reasons of a change it would be good to add the forum link as comment into the code.

How about creating an issue in GH (pointing to the forum) and adding a "fixes #XXXX" in the commit message?

wwmayer:
> In Release builds it only fails silently.

And isn't this undefined behaviour, then? According to the Qt code in release mode it accesses a memory address outside the array and then everything can happen. Most of the time the program will continue but sometimes it really crashes.

> How about creating an issue in GH (pointing to the forum) and adding a "fixes #XXXX" in the commit message?

Either this or add the forum link to the commit message. Do as you like.

0penBrain:
> And isn't this undefined behaviour, then? According to the Qt code in release mode it accesses a memory address outside the array and then everything can happen.

Indeed. I had to check the Qt source code too. I thought that in release mode, the `[]` operator would either throw an exception (that it doesn't do for sure) or return an empty `QChar()` if trying to access out of boundaries. But actually it indeed just blindly access the memory. :)

0penBrain:
> So, it would be good to add this change to your PR too:
> `if (!stringing && prefix.size() >= prefixEnd && prefix[prefixEnd-1] == QChar(32))`

OK, made the change. This said, the fix in the PR shall already prevent this.
So if we are at having this kind of checks, shouldn't we also check `prefixEnd > 0` ?

wwmayer:
> So if we are at having this kind of checks, shouldn't we also check prefixEnd > 0 ?

Yes. As I see you did this already the PR can be merged now.

0penBrain:
@donovaly to backported ?

donovaly:
> @donovaly to backported ?

Yes. But first this regression must be fixed: https://github.com/FreeCAD/FreeCAD/pull/7255#issuecomment-1442656426
(I noticed this directly after backporting and this way also broke the stable branch :-( )

0penBrain:
> > @donovaly to backported ?
> 
> Yes. But first this regression must be fixed: [#7255 (comment)](https://github.com/FreeCAD/FreeCAD/pull/7255#issuecomment-1442656426)

I propose to backport this + #8612 + #8624
I can do it as a batch when I'll merge the 2 latter ones later in the day.
Is is OK for you ?

wwmayer:
Yesterday I had a quick look at it and almost copied the entire function to the v0.20 branch. However, there still the same problem can be observed. Since there are made some further fixes recently we should consider copying over the whole class.

donovaly:
> > > @donovaly to backported ?
> > 
> > 
> > Yes. But first this regression must be fixed: [#7255 (comment)](https://github.com/FreeCAD/FreeCAD/pull/7255#issuecomment-1442656426)
> 
> I propose to backport this + #8612 + #8624 I can do it as a batch when I'll merge the 2 latter ones later in the day. Is is OK for you ?

Sure. Thanks in advance. Don't forget to update the backporting list in the forum. #8624 is already backported

donovaly:
I backported this now.
Concerning #8612, I could not backport because too many things were changed in master in the meantime.

PR #8553, opened 2023-02-20
[Gui] attempt to fix getting NaviCube font name

donovaly:
- as discussed here: https://forum.freecad.org/viewtopic.php?t=76233 maybe the conversion to ".toLatin1()" is unsafe on some systems for non-ASCII characters on the font names

PR #8567, opened 2023-02-21
Improve navicube default font matching on linux

adrianinsaval:
-set several font families for better matching on multiple systems
-set SansSerif stylehint again as it is needed for proper matching
-remove hack that saves the default font to parameters even when cancelling on DlgSettingsNavigation
-change return type of getDefaultSansserifFont to QString as only the font family is really used from this.

related to #8553 and https://forum.freecad.org/viewtopic.php?p=662155#p662155

@donovaly 

@chennes I would like to know what gets matched by this on macOS if you can test, if building is too much trouble getting the result from [this python snippet](https://forum.freecad.org/viewtopic.php?p=662267#p662267) could also help me.

---

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -346,6 +346,17 @@ MainWindow::MainWindow(QWidget * parent, Qt::WindowFlags f)
 
     // accept drops on the window, get handled in dropEvent, dragEnterEvent
     setAcceptDrops(true);
+
+    // setup font substitutions for NaviCube
+    // Helvetica usually gives good enough results on mac & linux
+    // in rare cases Helvetica matches a bad font on linux
+    // Nimbus Sans Narrow and Open Sans Condensed added as fallback
+    // Bahnschrift is a condensed font available on windows versions since 2017
+    // Arial added as fallback for older version
+    auto substitutions = QStringLiteral("Bahnschrift,Helvetica,Nimbus Sans Narrow,Open Sans Condensed,Arial,Sans");
+    auto family = QStringLiteral("FreeCAD NaviCube");
+    QFont::insertSubstitutions(family, substitutions.split(QLatin1Char(',')));
+
     statusBar()->showMessage(tr("Ready"), 2001);
 }
 
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -299,15 +299,8 @@ void NaviCube::setButtonColor(QColor ButtonColor)
 
 QString NaviCube::getDefaultSansserifFont()
 {
-    // Windows versions since 2017 have the 'Bahnschrift' font (a condensed
-    // sans serif font, optimized for readability despite being condensed),
-    // we first check for that.
-    QFont font(QStringLiteral("Bahnschrift"));
-    if (!font.exactMatch())
-        // On systems without 'Bahnschrift' we check for 'Helvetica' or its closest match
-        // as default sans serif font. (For Windows 7 this will e.g. result in 'Arial'.)
-        font = QFont(QStringLiteral("Helvetica"));
-
+    // "FreeCAD NaviCube" family susbtitutions are set in MainWindow::MainWindow
+    QFont font(QStringLiteral("FreeCAD NaviCube"));
     font.setStyleHint(QFont::SansSerif);
     // QFontInfo is required to get the actually matched font family
     return QFontInfo(font).family();


adrianinsaval:
archlinux package available for testing: https://github.com/adrianinsaval/pacman-repo/releases/download/archlinux-x86_64/freecad-git-0.21.0.31976-PR8567-x86_64.pkg.tar.zst

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8567">pipeline status</a> for feature branch [PR_8567](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8567). Pipeline [783754500](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/783754500) was triggered at [f175517](https://github.com/FreeCAD/FreeCAD/pull/8567/commits/f1755172a633f1c5e177e08f2ba96d6989e69e32). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
ugggggh setFamilies was introduced in qt 5.13 :(

chennes:
Well, this would still improve things for the majority of our users: so maybe just add a fallback for earlier Qt versions?

adrianinsaval:
I'll see how I can do that in the coming day(s), if you have a suggestion let me know, I'm not very experienced at this stuff.

I think the only system were we would realistically be dealing with qt<5.13 would be these older linux distros (and most people don't run new software from repos on those anyways), so I think adding a fallback to just "Sans"  should be enough

src/Gui/NaviCube.cpp
```diff
@@ -292,22 +292,21 @@ void NaviCube::setFontSize(int size)
...
-    // On systems without 'Bahnschrift' we check for 'Helvetica' or its closest match
-    // as default sans serif font. (For Windows 7 this will e.g. result in 'Arial'.)
-    font = QString::fromLatin1("Helvetica");
+    // Condensed fonts are prefered
+    // Helvetica usually gives good results on mac & linux
+    // font matching on windows seems to be very bad as of 20/02/23
+    // so a couple windows font that would be a exact match are set
+    // Bahnschrift is a condensed font available on windows
+    // versions since 2017
+    auto str = QString::fromLatin1("Bahnschrift,Helvetica Condensed,Arial Narrow,Nimbus Sans Narrow");
```
0penBrain, line 303: Use QLatin1String or QStringLiteral

0penBrain:
> I'll see how I can do that in the coming day(s), if you have a suggestion let me know, I'm not very experienced at this stuff

You can just use 'setFamily("Helvetica")' as a fallback. For example 'Nimbus' is of 'Helvetica' family so having it in the list is useless. ;)

adrianinsaval:
> You can just use 'setFamily("Helvetica")' as a fallback. For example 'Nimbus' is of 'Helvetica' family so having it in the list is useless. ;)

I added Nimbus Sans **Narrow** because I want specifically a narrow font, doesn't seem to be possible to get this by writing Helvetica, note that just using Helvetica has been reported as problematic in the past

0penBrain:
@adrianinsaval What should work is to add a custom substitution list with https://doc.qt.io/qt-5/qfont.html#insertSubstitutions using a custom family name (such as "FreeCAD NaviCube") then use this family name to get the font.

adrianinsaval:
I'll look into it, is the substitution recursive? That is, is none of the substitution families we set are available does it search the substitutions available for those families?

0penBrain:
I'm not sure but at least in this case it will use the "font matching algorithm" that in most cases should find an appropriate font.

wwmayer:
Build failure
```
/home/runner/work/FreeCAD/FreeCAD/src/Gui/NaviCube.cpp: In static member function ‘static QString NaviCube::getDefaultSansserifFont()’:
/home/runner/work/FreeCAD/FreeCAD/src/Gui/NaviCube.cpp:306:10: error: ‘class QFont’ has no member named ‘setFamilies’; did you mean ‘setFamily’?
  306 |     font.setFamilies(str.split(QLatin1Char(',')));
      |          ^~~~~~~~~~~
      |          setFamily
```
Handle it with
```
#if QT_VERSION < QT_VERSION_CHECK(5,13,0)
// older than Qt 5.13
#else
// Qt 5.13 or newer
#endif
```

0penBrain:
> Build failure

Using 'insertSubstitutions' should work for all versions. ;) If it works.

adrianinsaval:
@0penBrain should I also run removeSubstitutions after having used this or is it ok to leave it there?

src/Gui/NaviCube.cpp
```diff
@@ -292,22 +292,22 @@ void NaviCube::setFontSize(int size)
...
+    auto substitutions = QStringLiteral("Bahnschrift,Helvetica,Nimbus Sans Narrow,Arial,Sans");
+    auto family = QStringLiteral("FreeCAD NaviCube");
+    QFont::insertSubstitutions(family, substitutions.split(QLatin1Char(',')));
+    QFont font;
     font.setStyleHint(QFont::SansSerif);
-    return font;
+    font.setFamily(family);
+
+    // QFontInfo is required to get the matched font name
+    return QFontInfo(font).family();
```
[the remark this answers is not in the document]
adrianinsaval: Setting the sylehint in the QFont seems to be important for the font matching algorithm so I can't do that, I tried (in python) on windows and on Linux and if you don't set the sylehint a generic font is always picked up "MS Shell Dog 2" on Windows which is a bad one, on Linux at least it's Not Sans which is pretty good

src/Gui/NaviCube.cpp
```diff
@@ -292,22 +292,22 @@ void NaviCube::setFontSize(int size)
...
-    // as default sans serif font. (For Windows 7 this will e.g. result in 'Arial'.)
-    font = QString::fromLatin1("Helvetica");
+    // Helvetica usually gives good results on mac & linux
+    // in rare cases Helvetica matches a bad font on linux
+    // Nimbus Sans Narrow and Sans added as fallback
+    // Bahnschrift is a condensed font available on windows versions since 2017
+    // Arial added as fallback for older version
+    auto substitutions = QStringLiteral("Bahnschrift,Helvetica,Nimbus Sans Narrow,Arial,Sans");
+    auto family = QStringLiteral("FreeCAD NaviCube");
+    QFont::insertSubstitutions(family, substitutions.split(QLatin1Char(',')));
```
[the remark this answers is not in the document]
adrianinsaval: I'll see if I can do it but unfortunately I probably won't have time until the weekend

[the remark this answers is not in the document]
0penBrain: Oh yes. Please keep as is and ignore.

src/Gui/NaviCube.cpp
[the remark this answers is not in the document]
0penBrain: Oh yes. Please keep as is and ignore.

src/Gui/NaviCube.cpp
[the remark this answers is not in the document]
0penBrain: I don't think it's a problem. The concerned bug isn't causing hard crash and data loss. :)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8567">pipeline status</a> for feature branch [PR_8567](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8567). Pipeline [787723266](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/787723266) was triggered at [db249a6](https://github.com/FreeCAD/FreeCAD/pull/8567/commits/db249a62fe31feaadd3a66f11d64c2adb282e314). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Can we please go back a step, because:
- we know that on NaviCube.cpp, we find on all systems a font. To search for something better than Helvetica for e.g. Arch should be doe in a separate PR)
- we know that the found font is not correctly displayed in DlgSettingsNavigation.cpp. Therefore we need a PR that fixes this, meaning only changes DlgSettingsNavigation.cpp.

Otherwise we will circle around more days.
So we should separate the GUI part from the font selection mechanism.

@adrianinsaval , for curiosity, can you please compile latest master and report back what you now see in the preferences dialog when starting without defined font parameters?

adrianinsaval:
@0penBrain is this what you meant? looks a little disconnected to put this in mainwindow but it seems to work. I'm also not a fan of having "FreeCAD NaviCube" as just independent strings on each, is there a way I can store it somewhere and access it from navicube.cpp?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8567">pipeline status</a> for feature branch [PR_8567](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8567). Pipeline [788473633](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788473633) was triggered at [f464c32](https://github.com/FreeCAD/FreeCAD/pull/8567/commits/f464c32281e4269983a85cbff1d66ab836e2664e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
> Therefore we need a PR that fixes this, meaning only changes DlgSettingsNavigation.cpp.
ok I'll see about making this more minimalistic but this is delaying it even more IMO

> @adrianinsaval , for curiosity, can you please compile latest master and report back what you now see in the preferences dialog when starting without defined font parameters?

I'll try but it's not going to change anything, the issue is not in the string conversion, the issue is in using QFont when we need QFontInfo

src/Gui/NaviCube.cpp
```diff
@@ -292,22 +292,14 @@ void NaviCube::setFontSize(int size)
...
-    // Windows versions since 2017 have the 'Bahnschrift' font (a condensed
-    // sans serif font, optimized for readability despite being condensed),
-    // we first check for that.
-    QFont font(QString::fromLatin1("Bahnschrift"));
-    if (font.exactMatch())
-        return font;
-
-    // On systems without 'Bahnschrift' we check for 'Helvetica' or its closest match
-    // as default sans serif font. (For Windows 7 this will e.g. result in 'Arial'.)
-    font = QString::fromLatin1("Helvetica");
```
adrianinsaval, line 308: I just realized this was completely wrong! font is not even a QFont here

0penBrain:
> @0penBrain is this what you meant?

Yes

> looks a little disconnected to put this in mainwindow but it seems to work. I'm also not a fan of having "FreeCAD NaviCube"

To me it's not a problem. It's like if (in MainWindow) you create a new "generic" font called 'FreeCAD NaviCube', then later you use it in Navicube

0penBrain:
> > Therefore we need a PR that fixes this, meaning only changes DlgSettingsNavigation.cpp.
> 
> ok I'll see about making this more minimalistic but this is delaying it even more IMO (and changing only DlgSettingsNavigation.cpp is not smart as the flaw is in getDefaultSansserifFont)

@donovaly is that worth? The PR seems OK now. (I didn't check it solves the issue though, but it should).

adrianinsaval:
made alternative PR #8638 , choose whichever your prefer, this one I like more as it results in a better font in my system (it's still a little too wide and I want to check what the hell is going on with font sizes later)

adrianinsaval:
in DlgSettingsNavigation this is also saving the parameter even if you close the preference page without saving, is this the correct way of doing this? this does not affect this PR, just asking
```c++
    if (hGrp->GetInt("FontSize", 0) == 0) {
        // the "4" is the hardcoded m_OverSample from getDefaultFontSize()
        hGrp->SetInt("FontSize", int(0.18 * 4 * ui->prefCubeSize->value()));
        ui->naviCubeFontSize->onRestore();
```

donovaly:
> @donovaly is that worth?

Sadly yes. I already spent hours (sic!) in testing because Win 11 behaved so different (from laptop to laptop). Now we found a selection solution that works, therefore the first step should be to fix the UI part.
I tested now #8638 and this works on the 2 Win PCs I could access right now. Therefore I put this in. Thanks Adrián.

Now let's make the next step and adapt the font selection mechanism to font a narrow font on Arch. I will invest time to test the this further PR on my Win 11 machines.
Basically, what I can say for sure: If the font selection mechanism searches at first for the exact match of "Bahnschrift", it worked on all Win systems I tested. All other methods (e.g. checking for Helvetica before Bahnschrift etc.) failed on some machines

donovaly:
> in DlgSettingsNavigation this is also saving the parameter even if you close the preference page without saving, is this the correct way of doing this?

I don't know if this is correct, but at least this was my intention ;-). I tested a lot and this method assures the parameter is immediately written after opening the preferences dialog.

adrianinsaval:
I understand it writes the parameter immediately, but do we really want that? Why are we writing parameters that the user didn't touch?

0penBrain:
> I understand it writes the parameter immediately, but do we really want that? Why are we writing parameters that the user didn't touch?

No of course this isn't normal behavior. I saw someone on the forum that lose the Navicube texts very frequently, most probably about this mechanism.
We shall not save a preference without explicit user action.
BTW, we probably shall add a mechanism in the function seeking for the font. If the font matching done from the user preference returns no result, we shall automatically return the default font instead.

donovaly:
> No of course this isn't normal behavior. 

it is normal:
- delete all NaviCube parameters
- open the preferences dialog, do nothing but close it by pressing OK

result: the parameters are created and saved

> I saw someone on the forum that lose the Navicube texts very frequently, most probably about this mechanism.

The missing text Wandererfan and others observed was due to the bugs in the font selection. They should now be resolved.

0penBrain:
> it is normal:

No it isn't. ;)
https://github.com/FreeCAD/FreeCAD/blob/e5cdd9a80ea8018ef2c06b5fa5cd42f2d706917b/src/Gui/DlgSettingsNavigation.cpp#L209-L212
Here you are in the `loadSettings` function, so it means a parameter will be saved even if the user clicks "Cancel".
This is some kind of workaround, but not ideal and making the code harder to maintain.

adrianinsaval:
I agree, saving a parameter immediately upon opening preference is not ideal.

> BTW, we probably shall add a mechanism in the function seeking for the font. If the font matching done from the user preference returns no result, we shall automatically return the default font instead.

As far as I've seen QFontInfo always returns something, just a default font if nothing matches so I think Qt is already taking care of this mechanism and we don't need anything more.

donovaly:
> https://github.com/FreeCAD/FreeCAD/blob/e5cdd9a80ea8018ef2c06b5fa5cd42f2d706917b/src/Gui/DlgSettingsNavigation.cpp#L209-L212
> 
> Here you are in the `loadSettings` function, so it means a parameter will be saved even if the user clicks "Cancel".
> This is some kind of workaround, but not ideal and making the code harder to maintain.

This code has nothing to do with saving, but with loading.
This is the case why it is necessary:

- fresh FC installation, no NaviCube parameters written to the database
- user opens the first time the preferences

Then what font size do we show? Since there is no info available yet, we must get the info and that is what the code does.

To avoid doing this every time, we write the parameter also when user cancels the dialog so that from the next dialog call on we can read from the parameter.
What is wrong with this in your opinion?

adrianinsaval:
we get the default value and show it but don't save it

donovaly:
> we get the default value and show it but don't save it

Why? what is wrong with writing the default to the database and reading in future from it?

adrianinsaval:
because we are writing it with no user action, even with explicit user action to not save stuff. I've had trouble with this type of behavior before, when the default locations for user config and data was changed on linux the macro location directory was stored as a parameter despite being the default and me never having touched it, I had to manually change to the new default.

donovaly:
> because we are writing it with no user action, even with explicit user action to not save stuff. I've had trouble with this type of behavior before, when the default locations for user config and data was changed

But these cases are different.  I cannot see a problem of writing the font size we are using anyway also to the parameters and if the user would leave the preferences with the OK button than with the Cancel button, he gets the parameters written as well.

So to be precise, what exactly do you fear?

adrianinsaval:
> if the user would leave the preferences with the OK button than with the Cancel button, he gets the parameters written as well.

this is not ideal either but that's harder to change

> So to be precise, what exactly do you fear?

if we later change the default font size (say we make it adaptive to the screen resolution, or adaptive so the text fits inside the cube by default) an user that never touched this won't get the correct default in the future and would have to manually change the value, it is not a big deal but it's not good practice.

adrianinsaval:
rebased this PR, can you test if it works on windows @donovaly? would also appreciate input from more people using linux, does this look better?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8567">pipeline status</a> for feature branch [PR_8567](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8567). Pipeline [788902806](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788902806) was triggered at [d650fae](https://github.com/FreeCAD/FreeCAD/pull/8567/commits/d650fae76ba48f30d65edd162c352d883b3a02b6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/MainWindow.cpp
```diff
@@ -346,6 +346,17 @@ MainWindow::MainWindow(QWidget * parent, Qt::WindowFlags f)
...
     // accept drops on the window, get handled in dropEvent, dragEnterEvent
     setAcceptDrops(true);
+
+    // setup font substitutions for NaviCube
+    // Helvetica usually gives good enough results on mac & linux
+    // in rare cases Helvetica matches a bad font on linux
+    // Nimbus Sans Narrow and Open Sans Condensed added as fallback
+    // Bahnschrift is a condensed font available on windows versions since 2017
+    // Arial added as fallback for older version
+    auto substitutions = QStringLiteral("Bahnschrift,Helvetica,Nimbus Sans Narrow,Open Sans Condensed,Arial,Sans");
```
donovaly, line 356: very minor: we use spaces after comma

donovaly:
> rebased this PR, can you test if it works on windows @donovaly? would also appreciate input from more people using linux, does this look better?

I will test on Win 11 and report back -> will maybe be first tomorrow

src/Gui/MainWindow.cpp
0penBrain: Here would be a bad idea. This isn't parameter list. ;)

adrianinsaval: the comma is a separator to later split this into a QStringList, if I add spaces the second QString in the list will become `" Helvetica"` (with a leading space) instead of just `"Helvetica"` no? I'm not sure what effect that would have

donovaly:
On Win 10 it works.

wwmayer:
> Why? what is wrong with writing the default to the database and reading in future from it?

1. Because it does things a user (or programmer) doesn't expect. A function that does unexpected things is always bad.
2. Because it can cause inconsistencies. The parameter is written when opening the preferences but it is not written when directly creating a 3D view after starting the application. So, here the NaviCube uses a certain default font and when the user opens the preferences now you overwrite the font which the NaviCube immediately applies and here it can happen that Qt delivers a different font. So, for the user the situation can be: start FreeCAD > create 3D view where a certain font is used > open the preferences > the font has changed

> if the user would leave the preferences with the OK button than with the Cancel button, he gets the parameters written as well.

This is not possible. When he clicks the OK button the dialog is closed and he cannot do anything else.

donovaly:
>     1. Because it does things a user (or programmer) doesn't expect. A function that does unexpected things is always bad.

I don't understand because please try it out:
- remove all NaviCube parameters
- restart FC
- open the preferences
- close the preferences by pressing OK
-> the NaviCube parameters are written

This is the case also in FC 0.20.

>  So, for the user the situation can be: start FreeCAD > create 3D view where a certain font is used > open the preferences > the font has changed

No, the preferences take the same font  used by the 3D View creation. So what you describe can only happen when there is a bug. Looking at the dialog uses the same font selection code from NaviCube.cpp. So always the same code is executed.

> > if the user would leave the preferences with the OK button than with the Cancel button, he gets the parameters written as well.
> 
> This is not possible. When he clicks the OK button the dialog is closed and he cannot do anything else.

But as I wrote above, this is the case. Calling the preferences and do nothing but clicking OK writes parameters.

adrianinsaval:
> So what you describe can only happen when there is a bug.

which is exactly what happened and had to be fixed by these PR (in the case of font family), this method is more error prone so it should be avoided just for that to begin with. And beyond that , there's still the case where an user gets stuck with a value he never set even if we later change the default and has to manually fix it. For example, anyone who opened the preference page on linux now has a nonexistent "Helvetica" font set in their preference, my PRs won't fix it for them without manual intervention, because we wrote a parameter even if they didn't want to. This is just not good practice and should be avoided whenever possible.

> But as I wrote above, this is the case. Calling the preferences and do nothing but clicking OK writes parameters.

this is also not correct IMO, but not as easy to solve. It has been problematic in the past as I explained with the macro path parameter as well as some techdraw paths and likely other parameters.

0penBrain:
> But as I wrote above, this is the case. Calling the preferences and do nothing but clicking OK writes parameters.

Here we are talking that the parameter is written even if user does nothing and clicks **Cancel**.

wwmayer:
> I don't understand because please try it out:

Here the NaviCube gets the font name by reading the parameter and if the key doesn't exist getDefaultSansserifFont() is used. 
https://github.com/FreeCAD/FreeCAD/blob/ef29cc7657c4aa0552d7ec6f6b34201c25482a71/src/Gui/NaviCube.cpp#L408

Here the DlgSettingsNavigation checks the parameter key and if it's empty it will overwrite it. But an existing empty key is not the same as a non-existing key. 

> close the preferences by pressing OK
-> the NaviCube parameters are written

But that's not the point. The point is: close the preferences by pressing Cancel
-> some NaviCube parameters are written

donovaly:
> But an existing empty key is not the same as a non-existing key.

But this discussion is about the font size not the font: https://github.com/FreeCAD/FreeCAD/blob/ef29cc7657c4aa0552d7ec6f6b34201c25482a71/src/Gui/DlgSettingsNavigation.cpp#L213-L217

So the particular issue is that a font size is already used but we need to make the dialog display the size when the user calls it and no parameter exists yet.
I am open for any solution.

adrianinsaval:
Can't we do it like we do with the font now? Get what the default should be, set it in the UI but don't touch the parameters

wwmayer:
> But this discussion is about the font size not the font: 

Nobody talked ever about font size. It's all about the font name. Click on the links in my previous posting.

adrianinsaval:
The discussion started with me pointing out this behaviour withe the font size: https://github.com/FreeCAD/FreeCAD/pull/8567#issuecomment-1444431599

This was also the behaviour for the font family but that has already been solved by my other PR. IMO it doesn't really matter if it's font size or name, it is not desired behaviour.

wwmayer:
> The discussion started with me pointing out this behaviour withe the font size

Oh, I haven't seen this one. And yes writing the font size is also bad practice.

donovaly:
OK, how can we proceed?

I have no idea yet because when I take your concerns int account, then the whole preferences handling is not optimal:
- take FC 0.20
- assure you have no NaviCube parameters
- change to the FEM WB
- - open the preferences and make there a change (e.-g change an option in the CalculiX settings)

result: The NaviCube parameters are written (also some non-font. You argued that this was done by pressing "OK" but in fact the user did not see the NaviCube settings.
In my opinion this is no problem but you say yes.

As I wrote I am open for any solution, it is just that I have no idea at the moment. What I can do is not to write on pressing the cancel button but this does not solve the problem at all.

src/Gui/MainWindow.cpp
donovaly: OK then. To avoid that somebody changes this , I recommend to add a comment.

donovaly:
Back to the PR. I tested now on Win 11 and it works there.

So I vote for this:
- merge this PR
- start a discussion thread in the forum about the parameter writing because we should not clutter a PR with a slight off-topic discussion

adrianinsaval:
The PR does not touch parameter handling so yes, it can be merged independent of our parameter writing discussion.

src/Gui/MainWindow.cpp
adrianinsaval: I will add it, although I think it would be strange for any dev to change the contents of a harcoded string without considering the implications.

donovaly: I feared the auto-formatters But sure, they should not change a string value. So let's keep it as it is without a comment.
Sorry for the noise.

donovaly:
> The PR does not touch parameter handling so yes, it can be merged independent of our parameter writing discussion.

OK, so let's do this.

donovaly:
As I wrote yesterday, here is a PR that at least avoids that the font size parameter is written on loading the preferences: #8699
So when the user cancels the dialog the size won't be written to the parameters.

However, that the parameter is written when the user closes the preferences dialog with OK (no matter what he actually changed there) is an issue I have no clue how to resolve.

0penBrain:
> However, that the parameter is written when the user closes the preferences dialog with OK (no matter what he actually changed there) is an issue I have no clue how to resolve.

This would be a huge huge effort to do. :)

PR #8580, opened 2023-02-22
[Material] fix bugs for external material files

donovaly:
- handle case that material cards are chosen from folders outside of the standard material folders

- also reduce singleStep size for spin edits for more convenience

### DIFF
--- a/src/Mod/Material/MaterialEditor.py
+++ b/src/Mod/Material/MaterialEditor.py
@@ -61,7 +61,7 @@ class MaterialEditor:
         self.iconPath = (filePath + "Resources" + os.sep + "Icons" + os.sep)
 
         # load the UI file from the same directory as this script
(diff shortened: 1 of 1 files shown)


src/Mod/Material/MaterialEditor.py
```diff
@@ -738,6 +745,8 @@ def matProperWidget(parent=None, matproperty=None, Type="String", Value=None,
         # the magnetic permeability is the parameter for which many decimals matter
         # the most however, even for this, 6 digits are sufficient
         widget.setDecimals(6)
+        # for almost all Float parameters of materils a step of 1 would be too large
```
luzpaz, line 748: `materials`

donovaly: thanks, fixed

Issue #8545, opened 2023-02-19
[Feature Request] Add new property type for list of Part::TopoShape 


WandererFan:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

n/a

### Subproject(s) affected?

Part

### Idea description

In order to detect corrupt dimension references, TechDraw needs to maintain a copy of the shape pointed to by each reference (1-3 references per dimension).

It would be convenient if these copies could be in the form of Part::TopoShape or TopoDS_Shape.  

Currently there are properties for a single TopoShape (PropertyPartShape) and a list of low level geometry (PropertyGeometryList), but there is no property for a collection of TopoShapes.  

Putting the reference shapes into a Compound and storing in a PropertyPartShape would make it difficult to connect them back to the dimension reference.  Extracting the reference shape's low level geometry and storing it in PropertyGeometryList is an option, but would require converting the low level geometry back to a TopoShape/TopoDS_Shape whenever a comparison is required and from TopoShape/TopoDS_Shape to Part::Geometry whenever the dimension reference changes.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #8593, opened 2023-02-22
[Part]add PropertyTopoShapeList (FR 8545)

WandererFan:
This PR implements a solution to Feature Request #8545.  It adds PropertyTopoShapeList to the Part module.

---

### DIFF
--- a/src/Mod/Part/App/PropertyTopoShapeList.cpp
+++ b/src/Mod/Part/App/PropertyTopoShapeList.cpp
@@ -22,12 +22,13 @@
 
 #include "PreCompiled.h"
 
+#include <BRepBuilderAPI_Copy.hxx>
+
 #include <Base/Console.h>
 #include <Base/Reader.h>
 #include <Base/Writer.h>
 
 #include "TopoShapePy.h"
-#include "Part2DObject.h"
 
 #include "PropertyTopoShapeList.h"
 
@@ -53,14 +54,10 @@ PropertyTopoShapeList::PropertyTopoShapeList()
 
 PropertyTopoShapeList::~PropertyTopoShapeList()
 {
-    for (std::vector<TopoShape*>::iterator it = _lValueList.begin(); it != _lValueList.end(); ++it)
-        if (*it) delete *it;
 }
 
 void PropertyTopoShapeList::setSize(int newSize)
 {
-    for (unsigned int i = newSize; i < _lValueList.size(); i++)
-        delete _lValueList[i];
     _lValueList.resize(newSize);
 }
 
@@ -69,42 +66,67 @@ int PropertyTopoShapeList::getSize() const
     return static_cast<int>(_lValueList.size());
 }
 
-void PropertyTopoShapeList::setValue(TopoShape* lValue)
+// this verion of setValue is for the ADD_PROPERTY_TYPE macro to use when creating a
+// new PropertyTopoShapeList.  In other list properties that use pointers (instead of
+// references) a nullptr is passed to setValue to initialize the empty list, but we can
+// not do that with references.
+void PropertyTopoShapeList::setValue()
 {
-    if (lValue) {
-        aboutToSetValue();
-        _lValueList.resize(1);
-        _lValueList[0] = lValue;
-        hasSetValue();
-    }
+    aboutToSetValue();
+    clear();
+    hasSetValue();
 }
 
-void PropertyTopoShapeList::setValues(const std::vector<TopoShape*>& lValue)
+void PropertyTopoShapeList::setValue(const TopoShape &ts)
+{
+    aboutToSetValue();
+    _lValueList.resize(1);
+    _lValueList[0] = ts;
+    hasSetValue();
+}
+
+void PropertyTopoShapeList::setValues(const std::vector<TopoShape>& lValue)
 {
     aboutToSetValue();
     _lValueList.resize(lValue.size());
-    for (unsigned int i = 0; i < lValue.size(); i++)
+    for (unsigned int i = 0; i < lValue.size(); i++) {
         _lValueList[i] = lValue[i];
+    }
     hasSetValue();
+}
+
+// clean out the list
+void PropertyTopoShapeList::clear()
+{
+    _lValueList.clear();
+    _lValueList.resize(0);
+}
+
+// populate the lists with the TopoShapes that have now finished restoring
+void PropertyTopoShapeList::afterRestore()
+{
+    _lValueList.clear();
+    for (auto entry : m_restorePointers) {
+        _lValueList.push_back(*entry);
+        delete entry;     //??
+    }
+    m_restorePointers.clear();
+    App::PropertyLists::afterRestore();
 }
 
 PyObject *PropertyTopoShapeList::getPyObject()
 {
     PyObject* list = PyList_New(getSize());
     for (int i = 0; i < getSize(); i++)
-        PyList_SetItem( list, i, _lValueList[i]->getPyObject());
+        PyList_SetItem( list, i, _lValueList[i].getPyObject());
     return list;
 }
 
-
 void PropertyTopoShapeList::setPyObject(PyObject *value)
 {
-    // check container of this property to notify about changes
-    Part2DObject* part2d = dynamic_cast<Part2DObject*>(this->getContainer());
-
     if (PySequence_Check(value)) {
         Py_ssize_t nSize = PySequence_Size(value);
-        std::vector<TopoShape*> values;
+        std::vector<TopoShape> values;
         values.resize(nSize);
 
         for (Py_ssize_t i=0; i < nSize; ++i) {
@@ -115,19 +137,14 @@ void PropertyTopoShapeList::setPyObject(PyObject *value)
                 throw Base::TypeError(error);
             }
 
-            values[i] = static_cast<TopoShapePy*>(item)->getTopoShapePtr();
+            values[i] = *static_cast<TopoShapePy*>(item)->getTopoShapePtr();
         }
-
         setValues(values);
-        if (part2d)
-            part2d->acceptGeometry();
     }
     else if (PyObject_TypeCheck(value, &(TopoShapePy::Type))) {
         TopoShapePy  *pcObject = static_cast<TopoShapePy*>(value);
-        setValue(pcObject->getTopoShapePtr());
-        if (part2d)
-            part2d->acceptGeometry();
-    }
+        setValue(*pcObject->getTopoShapePtr());
+   }
     else {
         std::string error = std::string("type must be 'Shape' or list of 'Shape', not ");
         error += value->ob_type->tp_name;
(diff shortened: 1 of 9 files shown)


src/Mod/Part/App/PropertyTopoShapeList.h
```diff
@@ -0,0 +1,99 @@
...
+    void Save(Base::Writer &writer) const override;
+    void Restore(Base::XMLReader &reader) override;
+
+    App::Property *Copy() const override;
+    void Paste(const App::Property &from) override;
+
+    unsigned int getMemSize() const override;
+
+private:
+    std::vector<TopoShape*> _lValueList;
```
wwmayer, line 91: Is this on purpose to use a pointer here? If not then don't do it. Since the ownership of a TopoShape allocated on the heap is not always clear you cannot guarantee that PropertyTopoShapeList is the only owner of these objects. The result will be that several PropertyTopoShapeList own the same TopoShape instances which sooner or later will lead to a crash.

If you wanted to guarantee solely ownership you must change all methods to create a copy on the heap and them into the std::vector but then you can also just a TopoShape directly.

Don't worry about creating copies of a TopoShape because they are cheap. The class members of a TopoShape are a TopoDS_Shape, a long and an atomic int. The TopoDS_Shape itself is only a handle around the actual TShapes. So writing this won't create deep copies of the shape data:
```
TopoDS_Shape shape1 = ...
TopoDS_Shape shape2 = shape1;
TopoDS_Shape shape3 = shape2;
```

src/Mod/Part/App/PropertyTopoShapeList.h
```diff
@@ -0,0 +1,99 @@
...
+     * A more elaborate description of the destructor.
+     */
+    ~PropertyTopoShapeList() override;
+
+    void setSize(int newSize) override;
+    int getSize() const override;
+
+    /** Sets the property
+     */
+    void setValue(TopoShape*);
```
wwmayer, line 67: If you decide to use `std::vector<TopoShape> _lValueList` as member then also adjust the methods to work with a const-reference of TopoShape.

src/Mod/Part/App/PropertyTopoShapeList.cpp
```diff
@@ -0,0 +1,193 @@
...
+    reader.readEndElement("ShapeList");
+
+    // set the list to the restored values
+    setValues(std::move(values));
+}
+
+
+App::Property *PropertyTopoShapeList::Copy() const
+{
+    PropertyTopoShapeList *p = new PropertyTopoShapeList();
```
wwmayer, line 174: Here you should consider creating deep copies of the shapes with BRepBuilderAPI_Copy. This is also done in PropertyPartShape::Copy()

src/Mod/Part/App/PropertyTopoShapeList.cpp
```diff
@@ -0,0 +1,193 @@
...
+                std::string error = std::string("types in list must be 'Shape', not ");
+                error += item->ob_type->tp_name;
+                throw Base::TypeError(error);
+            }
+
+            values[i] = static_cast<TopoShapePy*>(item)->getTopoShapePtr();
+        }
+
+        setValues(values);
+        if (part2d)
```
wwmayer, line 122: I don't think that this acceptGeometry() stuff is needed here. It looks like a hack copied over from PropertyGeometryList.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8593">pipeline status</a> for feature branch [PR_8593](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8593). Pipeline [787723273](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/787723273) was triggered at [428e5d2](https://github.com/FreeCAD/FreeCAD/pull/8593/commits/428e5d29f94d906700320ec00f93d2891eed94a9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/PropertyTopoShapeList.h
WandererFan: Changed to const TopoShape& most places.   Could not get Restore to work without using pointers.  See PropertyTopoShapeList::Restore and PropertyTopoShapeList::afterRestore.

src/Mod/Part/App/PropertyTopoShapeList.cpp
WandererFan: Done.

src/Mod/Part/App/PropertyTopoShapeList.cpp
WandererFan: Removed.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8593">pipeline status</a> for feature branch [PR_8593](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8593). Pipeline [788929818](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788929818) was triggered at [9e060b2](https://github.com/FreeCAD/FreeCAD/pull/8593/commits/9e060b2ca7b46d0195413cb2a6092581276a9b67). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/PropertyTopoShapeList.h
wwmayer: OK I will have a deeper look on that...

WandererFan: Thanks.  No big rush. The new code that depends on this isn't quite ready.

wwmayer: I had a closer look at your implementation and it's fine. However, you should replace the raw pointer with shared pointers to be sure that no memory leak occurs.
There are a few places where `m_restorePointers.clear()` is called without actually deleting the pointers. So, with shared pointers we are safe.

What I have found are two missing aboutToSetValue() / hasSetValue() and for the Restore() function we can use TopoShape::Restore()

Here is the diff file of my modifications:
```
diff --git a/src/Mod/Part/App/PropertyTopoShapeList.cpp b/src/Mod/Part/App/PropertyTopoShapeList.cpp
index 9f4da48bdc..02791630ce 100644
--- a/src/Mod/Part/App/PropertyTopoShapeList.cpp
+++ b/src/Mod/Part/App/PropertyTopoShapeList.cpp
@@ -98,18 +98,22 @@ void PropertyTopoShapeList::setValues(const std::vector<TopoShape>& lValue)
 // clean out the list
 void PropertyTopoShapeList::clear()
 {
+    aboutToSetValue();
     _lValueList.clear();
     _lValueList.resize(0);
+    hasSetValue();
 }
 
 // populate the lists with the TopoShapes that have now finished restoring
 void PropertyTopoShapeList::afterRestore()
 {
+    aboutToSetValue();
     _lValueList.clear();
     for (auto entry : m_restorePointers) {
         _lValueList.push_back(*entry);
-        delete entry;     //??
     }
+    hasSetValue();
+
     m_restorePointers.clear();
     App::PropertyLists::afterRestore();
 }
@@ -170,14 +174,9 @@ void PropertyTopoShapeList::Restore(Base::XMLReader &reader)
     m_restorePointers.clear();      // just in case
     m_restorePointers.reserve(count);
     for (int i = 0; i < count; i++) {
-        reader.readElement("TopoShape");
-        std::string file (reader.getAttribute("file") );
-        if (!file.empty()) {
-            // add file to reader's list of files to be loaded by each newShape's RestoreDocFile()
-            TopoShape* newShape = new TopoShape();
-            m_restorePointers.push_back(newShape);
-            reader.addFile(file.c_str(), newShape);
-        }
+        auto newShape = std::make_shared<TopoShape>();
+        newShape->Restore(reader);
+        m_restorePointers.push_back(newShape);
     }
     reader.readEndElement("ShapeList");
 }
diff --git a/src/Mod/Part/App/PropertyTopoShapeList.h b/src/Mod/Part/App/PropertyTopoShapeList.h
index e1bbb388c0..add43270e9 100644
--- a/src/Mod/Part/App/PropertyTopoShapeList.h
+++ b/src/Mod/Part/App/PropertyTopoShapeList.h
@@ -97,7 +97,8 @@ private:
     // holds the new TopoShapes between their creation in Restore, and the completion of all
     // individual RestoreDocFile calls.  Once the restore is complete, the new TopoShapes are
     // inserted into _lValueList;
-    std::vector<TopoShape*> m_restorePointers;
+    using TopoShapePtr = std::shared_ptr<TopoShape>;
+    std::vector<TopoShapePtr> m_restorePointers;
 };
 
 } // namespace Part

```

What would be good too is to add a unit test to see that the property works. Here is some unpolished code that must be brought in form. If you don't know how to achieve that then ping me and I will write a proper test after this PR is merged.
```
doc = App.newDocument()
obj = doc.addObject("App::FeaturePython", "Object")
obj.addProperty("Part::PropertyTopoShapeList", "Shapes")
box = Part.makeBox(1,1,1)
box2 = Part.makeBox(1,1,2)
box3 = Part.makeBox(1,1,3)
obj.Shapes = [box, box2, box3]

doc.saveAs("/tmp/test.FCStd")
App.closeDocument(doc.Name)

doc = App.openDocument("/tmp/test.FCStd")
boxes = doc.ActiveObject.Shapes
boxes[0].Volume # ~1
boxes[1].Volume # ~2
boxes[2].Volume # ~3

twoboxes = [boxes[1], boxes[2]]

doc.openTransaction("Change shapes")
doc.ActiveObject.Shapes = twoboxes
len(doc.ActiveObject.Shapes) == 2
doc.commitTransaction()

doc.undo()

len(doc.ActiveObject.Shapes) == 3
```

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8593">pipeline status</a> for feature branch [PR_8593](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8593). Pipeline [791990761](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/791990761) was triggered at [6621d27](https://github.com/FreeCAD/FreeCAD/pull/8593/commits/6621d27893444e81dc137a15551fecd048b63245). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
The tests fails because when running the test from cmd then Undo/Redo of the document must be explicitly switched on.
Nevertheless I will merge the PR and fix the script afterwards.

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/8709

WandererFan:
doc.UndoMode = 1

This must be already set when using the Test Framework? 

In any case, thanks for your help with this.  I learned a lot.

wwmayer:
The point is when running FC in cmd mode then the creation of a document doesn't consider user settings (while it does when creating it in GUI mode). So, undo/redo is always disabled and must be explicitly enabled after the creation with doc.UndoMode = 1

> This must be already set when using the Test Framework?

No, there is nothing to be set beforehand. If needed UndoMode must be set for each created instance of a Document

PR #8601, opened 2023-02-23
Demonstrating separation of code and data

berniev:

---

For consideration / discussion.

That which can often be categorised as mere data is, in FC, often represented by/in code. E.g. if there is a data item "freddy", there is a function getFreddy() and enum.freddy etc etc.

This makes code very brittle. The slightest change to the data means changing code. Data is by definition subject to change (even if you don't initially think so, Murphy is everywhere!), and making it easy to change the stored representation is A Good Thing.

This PR amends License class to demonstrate data stored in exactly one place in a constexpr and not codified further. The data is immutable, locked-in at compile time. Behavior is unaffected by data change. Of course that is not to say some code that uses License won't make decisions based on content, but at least License itself is clean.

Note the removal of "Other" from the base data as it more correctly is something that is NOT in the set of data, an error condition.

Unit tests included.

Concerns: Compile time may be increased. How much? Don't know.


wwmayer:
Your PR looks basically good to me:

The PR fails to build with the CI. The failure for Ubuntu 20.04 is expected because some code in FreeCADGui must be adjusted.

However, the build for Ubuntu 22.04 already stops in FreeCADApp with tons of cryptic error messages. I don't know what exactly the problem should be. Maybe the compiler doesn't like the `using TLicenseArr = std::array<char*, 3>;` and prefers `using TLicenseArr = std::array<const char*, 3>;` which anyway is the better choice.

On my local Xubuntu 22.04 I can build it with clang (v14) and gcc (v11.3). The compiler used on the CI system is gcc 12.0

So, when building the code locally it breaks in DlgSettingsDocumentImp.cpp and DlgProjectInformationImp.cpp because this line is no longer valid:
`if (license.getType() == App::License::Type::Other)`
To fix this I suggest to add the method
```
bool License::isOtherLicense() const
{
    return strcmp(license.at(0), noLicense.at(0)) == 0;
}
```
and client code can be changed to `if (license.isOtherLicense())`

A further point that must be checked is that the code also compiles with MSVC and clang on MSYS2 because the experience shows that these compilers often struggle with constexpr.

berniev:
Thanks for having a look!

The primary purpose of the PR is obviously promotion of the concept of separating code and data.

Yes, am aware that there were some build issues from 

> if (license.getType() == App::License::Type::Other)

Maybe I havn't pushed latest, yet, to fix those.

I am super impressed, however,  that you accept the basic premise of the PR! Thank you!

wwmayer:
> I am super impressed, however, that you accept the basic premise of the PR! Thank you!

Your solution is pretty smart and very easy to extend by further entries. The only thing what could happen is that one forgets to adjust countOfLicenses but then the compiler will tell you that. So, no risk to break anything.

berniev:
"Other" is a contrivance for not-a-license! Changed to "".
`constexpr char* noLicenseStr = "";` Pedantic removal of magic string.

Added:
```
template<typename T>
constexpr bool isLicense(T arg)
{
    return findLicense(arg).at(0) != noLicenseStr;
}
```
The template is to handle int and long types. TODO expunge int and long as they rely on the ordering of licenses in the set.

getLicenses doesn't belong in an instance of License as it relates to all licenses. Moved to a constexpr function outside the class. Return a std::array instead of std::vector for efficiency and to reflect fixed size.
```
constexpr std::array<char*, countOfLicenses> getLicenses()
{
    std::array<char*, countOfLicenses> arr {};
    for(int i = 0; i < countOfLicenses; i++){
        arr.at(i) = licenseItems.at(i).at(1);
    }
    return arr;
};
```

TODO: "Type" and "License" I keep finding confusing. Are these standard words for license titles? If not then perhaps "ShortName" and "Name" would be more appropriate?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601">pipeline status</a> for feature branch [PR_8601](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601). Pipeline [787717287](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/787717287) was triggered at [eef993d](https://github.com/FreeCAD/FreeCAD/pull/8601/commits/eef993dd69f445c0b13a1c8ce97c6137c6b417a7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> TODO: "Type" and "License" I keep finding confusing. Are these standard words for license titles? If not then perhaps "ShortName" and "Name" would be more appropriate?

I don't mind. Your suggestions is OK for me. Another option is to use the term SPDX which e.g. is used in the Linux kernel where the SPDX identifier is used instead of the 20 lines long license text in every file.

https://en.wikipedia.org/wiki/Software_Package_Data_Exchange
https://spdx.org/licenses/

berniev:
Thanks! "Identifier" and "FullName" it is. Much better!

> The failure for Ubuntu 20.04 is expected because some code in FreeCADGui must be adjusted.

So can I safely ignore this for now? I just spent some time on this before coming back to your sage advice.

berniev:
`using TLicenseArr = std::array<const char*, 3>`

The hurrier I go the behinder I get. I'll investigate this. Thanks.

wwmayer:
> So can I safely ignore this for now?

I saw that you already adjusted one of the dialogs but it fails for another reason now. The iterator variable already is a const char*, so `c_str()` doesn't exist. But that's trivial to fix once you think the PR is ready.

About the failure with Ubuntu 22.04 you should try to switch from char* to const char* as recommended. Here I found some information about it: https://www.enseignement.polytechnique.fr/informatique/INF478/docs/Cpp/en/cpp/language/string_literal.html

See the second last point in the notes section:
> In C, string literals are of type char[], and can be assigned directly to a (non-const) char*. C++03 allowed it as well (but deprecated it, as literals are const in C++). C++11 no longer allows such assignments without a cast. 

Good possible that "older" compilers were still too permissive to allow it but now not any more.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601">pipeline status</a> for feature branch [PR_8601](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601). Pipeline [788636303](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788636303) was triggered at [df2a313](https://github.com/FreeCAD/FreeCAD/pull/8601/commits/df2a313ff715fb19a65b57754cfc6d0d181db5bc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601">pipeline status</a> for feature branch [PR_8601](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601). Pipeline [788644355](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788644355) was triggered at [91a1af7](https://github.com/FreeCAD/FreeCAD/pull/8601/commits/91a1af7b0602534036eaff5286ad3739bc7dbe61). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berniev:
const worked! Thank you!

It would be really good to see the numeric license identifiers replaced by the standard strings, but I'm not confident enough  to attempt that just yet.

wwmayer:
> It would be really good to see the numeric license identifiers replaced by the standard strings

For Document.cpp this is easy. You only have to change the parameter type from long to string, i.e. use GetASCII instead of GetInt. So, in the config file two strings are used.

For DlgSettingsDocumentImp.cpp and  DlgProjectInformationImp.cpp:
Therefore the license class must be changed to return identifier and url.

When filling the combo box pass two arguments to addItem(). The first one is the translated string as it's done now and the second parameter is the url. For "Other" pass an empty url.

In the slot function onLicenseTypeChanged() you get an int and with this you can access the second argument of addItem() with itemData(). Pass the value of itemData() to the line edit for the url.
As you can see the License class is then not even needed in onLicenseTypeChanged().

berniev:
Fact check pls:

In document is stored `License` and gives an example licenseIdentifier. I have amended the comment to `/// The license identifier, e.g. "CC_BY_40". See also https://spdx.org/licenses/`

If this is correct, the var could be better name licenseIdentifier.

Also in document is `LicenseUrl` with comment
> /// License description/contract URL

which is contradictory. Is the comment right or wrong? Is it meant to be editable?

The license ComboBox is given a list of licenseIdentifiers. Currently signal currentIndexChanged is utilised which sends an int. Could currentTextChanged be used instead as valid licenseIdentifier would be presented (assuming the text is not editable).

ui->prefLicenseType would be better named ui->prefLicenseTypes (plural) as it is a collection. Grrr.

Originally I thought licenseFullName was used, but it now appears that is wrong. Only licenseIdentifier (and url). Very confusing!

How am I doing?

wwmayer:
> Is the comment right or wrong? Is it meant to be editable?

Don't know. When looking at the code it seems there was some code not properly merged because the properties were registered twice which I fixed in 4036613fd29e. So, good possible that there is some leftover.

> Could currentTextChanged be used instead as valid licenseIdentifier would be presented (assuming the text is not editable).

But how would you then get the corresponding url? In order to get it an expensive string comparison must be done which is not even safe because the string the combo box delivers is translated. This means it will be rather complicated to find the matching url.

berniev:
The translation  you mean language, I assume. Does it make sense to translate a licenseIdentifier? From the string I was thinking just initialise a License. Now integer is no longer required in License.

wwmayer:
> Does it make sense to translate a licenseIdentifier?

Sorry for the confusion. I meant the long name of the license. This of course will be at least partially translated.

berniev:
As I understand them, these are the design goals for license setting (some of which I have drifted away from):

`License` is now called `LicenseFullName`.
`Type` is now called `LicenseIdentifier`.

`LicenseIdentifier` is not used.
ComboBox displays LicenseFullNames, translated. Immutable.
Document stores FullName, translated.
ComboBox options to include "Other"
Url is lookup from license data. Immutable except In the case of "Other" where it is set blank and editable.

Notes:
Some comments/documentation needs correcting.
Example files need correcting. Most (not all) show LicenseIdentifier.

****************************************
Ureka!!!! The answer was staring me in the face the whole time.
Access `licenseItems` directly. No loading, copying, instantiation etc. Fast and simple.
License.cpp can go completely. As can much of License.h
Who was it said "Writing complicated code is easy. Writing simple code is hard!"
****************************************
Add in License.h:
```
constexpr int posnOfIdentifier = 0;
constexpr int posnOfFullName = 1;
constexpr int posnOfUrl = 2;
```
Then:
```
void DlgSettingsDocumentImp::addLicenseTypes()
{
    auto add = [&](const char* what) {
        ui->prefLicenseType->addItem(
            QApplication::translate("Gui::Dialog::DlgSettingsDocument", what));
    };

    ui->prefLicenseType->clear();
    for (const auto& licenseItem : App::licenseItems) {
        add(licenseItem.at(App::posnOfFullName));
    }
    add("Other");
}

/**
 * Fix Url according to changed type
 */
void DlgSettingsDocumentImp::onLicenseTypeChanged(int index)
{
    if (index < App::countOfLicenses) {
        // existing license
        const char* url {App::licenseItems.at(index).at(App::posnOfUrl)};
        ui->prefLicenseUrl->setText(QString::fromLatin1(url));
        ui->prefLicenseUrl->setReadOnly(true);
    }
    else {
        // Other
        ui->prefLicenseUrl->clear();
        ui->prefLicenseUrl->setReadOnly(false);
    }
}
```
and
```
void DlgProjectInformationImp::onLicenseTypeChanged(int index)
{
    const char* url {index < App::countOfLicenses ? App::licenseItems.at(index).at(App::posnOfUrl)
                                                  : _doc->LicenseURL.getValue()};

    ui->lineEditLicenseURL->setText(QString::fromLatin1(url));
}
```
I'm not at all sure about this though in Document constructor. What is it doing?:
```
    // license stuff
    auto paramGrp {App::GetApplication().GetParameterGroupByPath(
        "User parameter:BaseApp/Preferences/Document")};
    auto index = static_cast<int>(paramGrp->GetInt("prefLicenseType", 0));
    const char* name = App::licenseItems.at(index).at(App::posnOfFullName);
    const char* url = App::licenseItems.at(index).at(App::posnOfUrl);
    const char* licenseUrl = (paramGrp->GetASCII("prefLicenseUrl", url)).c_str();

    ADD_PROPERTY_TYPE(License, (name), 0, Prop_None, "License string of the Item");
    ADD_PROPERTY_TYPE(
        LicenseURL, (licenseUrl), 0, Prop_None, "URL to the license text/contract");

```
Similarly not sure about:
```
void DlgProjectInformationImp::accept()
{
    ...
    QByteArray license = ui->comboLicense->itemData(ui->comboLicense->currentIndex()).toByteArray();
    if (license.isEmpty())
        license = ui->comboLicense->itemText(ui->comboLicense->currentIndex()).toUtf8();
    _doc->License.setValue(license);
    _doc->LicenseURL.setValue(ui->lineEditLicenseURL->text().toUtf8());

```

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601">pipeline status</a> for feature branch [PR_8601](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601). Pipeline [791697069](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/791697069) was triggered at [efeef2d](https://github.com/FreeCAD/FreeCAD/pull/8601/commits/efeef2d8da7f7ca817cdd28f64b8146a666df516). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> I'm not at all sure about this though in Document constructor. What is it doing?:

What it is supposed to do is using "All rights reserved" as the default license if the user hasn't specified another license in the settings.

> Similarly not sure about:

If the user opens this dialog then the currently set license and url of the document is set in the dialog. Here he can change the license and its url and when pressing OK it will be written to the properties (License & LicenseURL) of the document.

berniev:
Thank you Sir. Will check it out.

berniev:
AFAICT the behaviour should be the same as it was.

wwmayer:
Is this the final state of how you want it to be?

berniev:
I think so! Does it look ok to you?

wwmayer:
> I think so! Does it look ok to you?

I will have a look tomorrow. And if you look at the integration tests there are several failures now and it must be checked whether it's related to this PR or something else.

berniev:
My build is too broken ATM to run the integration tests, but if you give me the details I'll have a look.

wwmayer:
The failures are related to this PR and this line is the culprit:
`const char* licenseUrl = (paramGrp->GetASCII("prefLicenseUrl", url)).c_str();`

GetASCII() returns a tmp. std::string object and it will be destroyed after the call of c_str(). This means that licenseUrl is a dangling pointer and may point to a garbage value. Writing this garbage will invalidate the Document.xml file of an FCStd project so that it cannot be opened any more. This is then the reason for the failing tests.

The line must be corrected to:
`std::string licenseUrl = (paramGrp->GetASCII("prefLicenseUrl", url));`

berniev:
and L:1577 to use licenseUrl.c_str() to pass in a C string from the now std::string.

Made the changes, but struggling to come up with a rule of thumb to avoid this in future. Any hints?

berniev:
push is not working at the moment. will try later

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601">pipeline status</a> for feature branch [PR_8601](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8601). Pipeline [798073489](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/798073489) was triggered at [15baf45](https://github.com/FreeCAD/FreeCAD/pull/8601/commits/15baf45c585c8ddca2c001cdbc99da3d01f1de8a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> Made the changes, but struggling to come up with a rule of thumb to avoid this in future. Any hints?

The `const char*` is a raw pointer owned by the std::string. So, accessing a pointer of a tmp. object usually results in dangling pointers.

PR #8646, opened 2023-02-25
[Gui] add UI to change the NaviCube button size

donovaly:
- it was frequently requested and people now even published videos since I documented this hidden feature in the Wiki. However, for a user it is a nightmare without a UI and users who like a bright background need to change the button color to keep the buttons visible.

### DIFF
--- a/src/Gui/DlgSettingsNavigation.cpp
+++ b/src/Gui/DlgSettingsNavigation.cpp
@@ -57,6 +57,7 @@ DlgSettingsNavigation::DlgSettingsNavigation(QWidget* parent)
     , q0(0), q1(0), q2(0), q3(1)
 {
     ui->setupUi(this);
+    ui->naviCubeButtonColor->setAllowTransparency(true);
     retranslate();
 }
 
@@ -93,6 +94,7 @@ void DlgSettingsNavigation::saveSettings()
     ui->naviCubeToNearest->onSave();
     ui->prefCubeSize->onSave();
     ui->naviCubeFontSize->onSave();
+    ui->naviCubeButtonColor->onSave();
 
     bool showNaviCube = ui->groupBoxNaviCube->isChecked();
     hGrp->SetBool("ShowNaviCube", showNaviCube);
@@ -141,6 +143,7 @@ void DlgSettingsNavigation::loadSettings()
     ui->naviCubeToNearest->onRestore();
     ui->prefCubeSize->onRestore();
     ui->naviCubeFontSize->onRestore();
+    ui->naviCubeButtonColor->onRestore();
 
     ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath
         ("User parameter:BaseApp/Preferences/View");
--- a/src/Gui/DlgSettingsNavigation.ui
+++ b/src/Gui/DlgSettingsNavigation.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>500</width>
-    <height>394</height>
+    <width>432</width>
+    <height>423</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -117,7 +117,7 @@
         </item>
        </widget>
       </item>
-      <item row="2" column="0" colspan="2">
+      <item row="1" column="0" colspan="2">
        <widget class="Gui::PrefCheckBox" name="naviCubeToNearest">
         <property name="toolTip">
          <string>Rotates to nearest possible state when clicking a cube face</string>
@@ -136,14 +136,14 @@
         </property>
        </widget>
       </item>
-      <item row="2" column="3">
-       <widget class="QLabel" name="TextLabel3">
+      <item row="1" column="3">
+       <widget class="QLabel" name="FontNameLabel">
         <property name="text">
          <string>Font name:</string>
         </property>
        </widget>
       </item>
-      <item row="2" column="4">
+      <item row="1" column="4">
        <widget class="Gui::PrefComboBox" name="naviCubeFontName">
         <property name="toolTip">
          <string>Font name of the navigation cube</string>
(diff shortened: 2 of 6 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8646">pipeline status</a> for feature branch [PR_8646](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8646). Pipeline [788749479](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/788749479) was triggered at [cfc9594](https://github.com/FreeCAD/FreeCAD/pull/8646/commits/cfc95945a9f32bda3213bfcc8817d80c9d432554). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
the commit message is wrong but I guess it's too late to change now right? or is it allowed to force push to master?

donovaly:
Too late. But not a big deal anyway

wwmayer:
> or is it allowed to force push to master?

No because allowing this would be really bad because you would break master branch for anybody else.

> the commit message is wrong but I guess it's too late to change now right?

But the color handling seems to me wrong too because in View3DInventor.cpp an unsigned long is directly passed to a QColor while in NaviCube it's split into its RGBA components and then assigned to a QColor. This will result into different color values.

Qt handles an unsigned int as ARGB while FreeCAD handles it like Coin3D as RGBA

```
>>> from PySide import QtGui
>>> col = 3806916480
>>> color = QtGui.QColor(col)
>>> (col >> 24) & 0xff
226
>>> color.red()
232
>>> (col >> 16) & 0xff
232
>>> color.green()
239
>>> (col >> 8) & 0xff
239
>>> color.blue()
128
>>> col & 0xff
128
>>> color.alpha()
255
>>> color2 = QtGui.QColor((col >> 24) & 0xff, (col >> 16) & 0xff, (col >> 8) & 0xff, col & 0xff)
>>> color.red()
232
>>> color2.red()
226
>>> color.green()
239
>>> color2.green()
232
>>> color.blue()
128
>>> color2.blue()
239
>>> color.alpha()
255
>>> color2.alpha()
128
>>> 
```
Additionally when assigning an unsigned int (aka QRgb) then the alpha channel will be ignored and always set to 255:
https://doc.qt.io/qt-6/qcolor.html#QColor-3

donovaly:
> But the color handling seems to me wrong too because in View3DInventor.cpp an unsigned long is directly passed to a QColor while in NaviCube it's split into its RGBA components and then assigned to a QColor.

I don't understand. In View3DInventor.cpp the same default value is passed as in NaviCube .cpp. This unsigned long is in the form of RRGGBBAA.
On reading in the unsigned long, the shift from RGBA to ARBG is made in NaviCube.cpp.

When you take latest master and assure you don't have in the parameters "ButtonColor", you get for new cubes the correct whitish color.
When you open the preferences, you can change the color using the new color button and get the color that you selected.

The only bug I see at the moment is that in Qt Designer the Gui::ColorButton property `allowTransparency` cannot be set. I made now this PR: #8651
Since I don't know yet much about the widgets, can you please review the PR and merge if OK?

wwmayer:
Just run my test code in Python and you will see they are different.

A color as unsigned long/unsigned int in a parameter is considered to be RGB(A) but Qt considers them to be as ARGB.
Additionally the constructor that accepts an unsigned int (which QRgb is) will discard the alpha value.

In View3DInventor.cpp  you pass the unsigned long directly which is wrong:
` _viewer->getNavigationCube()->setButtonColor(rGrp.GetUnsigned("ButtonColor", 3806916480));`

Inside the NaviCube class it's done right.

Currently I am preparing a PR that provides some convenience functions to simplify the conversion.

donovaly:
> In View3DInventor.cpp you pass the unsigned long directly which is wrong: ` _viewer->getNavigationCube()->setButtonColor(rGrp.GetUnsigned("ButtonColor", 3806916480));`
> 
> Inside the NaviCube class it's done right.

Ah, now I see. I will make a PR to fix this.

donovaly:
> Ah, now I see. I will make a PR to fix this.

done: #8663

adrianinsaval:
the failure to restore transparency happens with a color set from the UI. To reproduce set a button color with some transparency value, click ok and the navicube is correctly updated, open the navicube preferences again and click ok, transparency value has been lost.

Actually it even happens with the default color, open freecad with a clean config and the buttons are slightly transparent, go to navicube preferences and click ok without chanign anything, the buttons are now fully opaque

adrianinsaval:
the change in how the color is interpreted is problematic when sharing the same config with 0.20, consider changing the parameter name if the new interpretation is a must or backporting the change to interpret the colors correctly. After having just opened and click ok on the navicube preference page on a 0.21 install my 0.20 install now looks like this:
image

donovaly:
> the failure to restore transparency happens with a color set from the UI. 

I can reproduce. I am 100% sure it worked once I implemented it. I see that the color handling was improved, so i will have to look why the transparency is no longer respected.

donovaly:
> the change in how the color is interpreted is problematic when sharing the same config with 0.20

Yes but this cannot be changed. In FC 0.20 the transparency was the first of the color code tuple, not it is the last because for the UI support we rely on QColor.

donovaly:
Here is a workaround: #8812

But this does not fix the general bug that onRestore() does not read the transparency.

adrianinsaval:
> > the change in how the color is interpreted is problematic when sharing the same config with 0.20
> 
> Yes but this cannot be changed. In FC 0.20 the transparency was the first of the color code tuple, not it is the last because for the UI support we rely on QColor.

That's why I suggest changing the parameter path in 0.21, so it doesn't interfere with earlier versions as they are not compatible with the new format

donovaly:
> But this does not fix the general bug that onRestore() does not read the transparency.

OK, here is a PR that fixes this general issue: #8813

PR #8654, opened 2023-02-25
Gui: Eliminate variable reuse

chennes:
Coverity complains about USE_AFTER_FREE here (CID 436695), and I found it was difficult to reason about the code to determine whether the analyzer was correct. To help address this, refactor to use different variables for the different parameter groups. Also change the function call to use a non-reference to `Reference`, for the same reason. @wwmayer I don't have a good sense of when I should be using a C++ reference to Base::Reference and when to just let the ref-counted copy be made, so I'd welcome any advice you have there. I don't think the changed function signature is an important part of the PR, and could be rolled back.

---

### DIFF
--- a/src/Gui/Workbench.cpp
+++ b/src/Gui/Workbench.cpp
@@ -226,14 +226,15 @@ void Workbench::setName(const std::string& name)
 void Workbench::setupCustomToolbars(ToolBarItem* root, const char* toolbar) const
 {
     std::string name = this->name();
-    ParameterGrp::handle hGrp = App::GetApplication().GetUserParameter().GetGroup("BaseApp")
-        ->GetGroup("Workbench");
+    const auto workbenchGroup {
+        App::GetApplication().GetUserParameter().GetGroup("BaseApp")->GetGroup("Workbench")
+    };
     // workbench specific custom toolbars
-    if (hGrp->HasGroup(name.c_str())) {
-        hGrp = hGrp->GetGroup(name.c_str());
-        if (hGrp->HasGroup(toolbar)) {
-            hGrp = hGrp->GetGroup(toolbar);
-            setupCustomToolbars(root, hGrp);
+    if (workbenchGroup->HasGroup(name.c_str())) {
+        const auto customGroup = workbenchGroup->GetGroup(name.c_str());
+        if (customGroup->HasGroup(toolbar)) {
+            const auto customToolbarGroup = customGroup->GetGroup(toolbar);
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8654">pipeline status</a> for feature branch [PR_8654](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8654). Pipeline [789013942](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/789013942) was triggered at [d910abb](https://github.com/FreeCAD/FreeCAD/pull/8654/commits/d910abbc1cedaee5685cbddf354c61560da682c1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer (commented):
It's a good opportunity to clean up the code but I don't think it makes the reported issue to disappear because IMO it's a false-positive.

A Base::Reference is similar to a std::shared_ptr with the difference that the template argument already provides its own reference counter.

When the Base::Reference calls unref() and the counter reaches 0 the handled object will be destroyed. I guess the analyzer finds this fact but it doesn't know anything about the current value of the reference counter and simply claims that the object is used after it has been freed.

chennes:
> I don't think it makes the reported issue to disappear because IMO it's a false-positive.

I agree, and I suspect you are right about what it's seeing. My thinking with this PR was that by reducing the variable re-use we could get Coverity's analysis path to be easier to understand (it's 19 calls deep right now), but it's not terribly important.

> A Base::Reference is similar to a std::shared_ptr with the difference that the template argument already provides its own reference counter

With a `shared_ptr` I'd never pass it by reference since the copy is so cheap: is the same true of Base::Reference? It doesn't look like it's done much in the code, maybe a dozen times for a `Base::Reference<ParameterGrp>`.

wwmayer:
> With a shared_ptr I'd never pass it by reference since the copy is so cheap: is the same true of Base::Reference?

The Base::Reference uses an QAtomiInt of the handled type which probably has the same cost as a shared_ptr. In this case it doesn't make a big difference because the function is not called that often but in general passing shared_ptr by value has its price.

Here you will find the summary of a video of Nicolai Josuttis https://www.reddit.com/r/cpp/comments/3azi39/the_real_price_of_shared_pointers_in_c_nico/ where it's pointed out how much the cost will be.

PR #8658, opened 2023-02-26
Part: SectionCutting cutTransparency initialization

chennes:
Addresses Coverity [CID 356563](https://scan8.scan.coverity.com/reports.htm#v44798/p11555/fileInstanceId=71281803&defectInstanceId=12830295&mergedDefectId=356563): Uninitialized scalar. `cutTransparency` is later used, from the static analyzer's point of view possibly uninitialized, in a comparison operation. I don't think it's actually possible for that to happen given the logic of the code, but initialize the POD type anyway, for good measure. I chose a value that it can't be given so that if this analysis is incorrect, it should be obvious what happened when stepping through the code.

---

- [X]  This Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

### DIFF
--- a/src/Mod/Part/Gui/SectionCutting.cpp
+++ b/src/Mod/Part/Gui/SectionCutting.cpp
@@ -603,7 +603,7 @@ void SectionCut::startCutting(bool isInitial)
 
     // store color and transparency of first object
     App::Color cutColor;
-    int cutTransparency;
+    int cutTransparency {0};
     bool autoColor = true;
     bool autoTransparency = true;
     auto vpFirstObject = dynamic_cast<Gui::ViewProviderGeometryObject*>(


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8658">pipeline status</a> for feature branch [PR_8658](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8658). Pipeline [789065309](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/789065309) was triggered at [22555e5](https://github.com/FreeCAD/FreeCAD/pull/8658/commits/22555e5b94b45f82a8fe965ea34cb84034d09e4b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8658">pipeline status</a> for feature branch [PR_8658](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8658). Pipeline [789066184](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/789066184) was triggered at [376c2fb](https://github.com/FreeCAD/FreeCAD/pull/8658/commits/376c2fbdfd6b7d2dfbe1a6d68cf719cf81099cdd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/Gui/SectionCutting.cpp
```diff
@@ -603,7 +603,7 @@ void SectionCut::startCutting(bool isInitial)
 
     // store color and transparency of first object
     App::Color cutColor;
-    int cutTransparency;
+    int cutTransparency {-1};
```
donovaly, line 606: Why not
`int cutTransparency = 0;`

I think we should initialize to sensible values just in case the initialization will be used in an unforeseen case.

chennes: My reasoning was that if that variable is ever used without getting set, something has gone wrong. Initializing it to zero , which is a valid value, would mask the error when debugging. If you are not concerned with that eventuality I will adjust the PR.

donovaly: I don't understand the problem at all because few lines below the variable is initialized by reading the transparency of an object. We already check if that object exists.
If the valid object has for whatever reason no transparency, the int would have the value 0. So we can safely initialize it as o.
However, what I learned from C++ is that a new int gets internally already the value 0. Therefore no special initialization is necessary if 0 is a usable value.

chennes: > However, what I learned from C++ is that a new int gets internally already the value 0. Therefore no special initialization is necessary if 0 is a usable value.

That is not true: the compiler is not required to initialize POD types unless they have static storage duration, and when compiled in Release mode usually does not do so. Because of this subtlety it is widely regarded as a best practice to initialize all POD variables in C++. See, e.g. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es20-always-initialize-an-object

donovaly: Thanks. so I will in future initialize these variables.

Here I would prefer to initialize to 0.

chennes: OK.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8658">pipeline status</a> for feature branch [PR_8658](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8658). Pipeline [789096186](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/789096186) was triggered at [b985774](https://github.com/FreeCAD/FreeCAD/pull/8658/commits/b9857744939c2e2f7d2ed663d2df6f38d584071e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Why didn't you assign the zero?:
I mean why not
`int cutTransparency = 0;`
but
`int cutTransparency {0};`

donovaly (approved):

chennes:
> Why didn't you assign the zero?:

Although I personally prefer the assignment-style initializing, I think that's just because I'm a decrepit old man who's been coding in C++ since 1998. Kids these days tell me I am supposed to be using {}-initializers, and I am trying to learn! https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax

PR #8683, opened 2023-02-26
[App] attempt to fix compiler warnings in Property.h

donovaly:
- "conversion from 'size_t' to 'int', possible loss of data" in for loops by modernizing the for loops
- fix warning about uninitialized variable

### DIFF
--- a/src/App/Property.h
+++ b/src/App/Property.h
@@ -550,24 +550,31 @@ public:
 
 protected:
 
-    void setPyValues(const std::vector<PyObject*> &vals, const std::vector<int> &indices) override 
+    void setPyValues(const std::vector<PyObject*>& vals, const std::vector<int>& indices) override
     {
+        int i = 0;
         if (indices.empty()) {
-            ListT values;
+            ListT values {};
             values.resize(vals.size());
-            for (std::size_t i=0,count=vals.size();i<count;++i)
-                values[i] = getPyValue(vals[i]);
+            i = 0;
+            for (auto valsContent : vals) {
+                values[i] = getPyValue(valsContent);
+                i++;
+            }
             setValues(std::move(values));
             return;
         }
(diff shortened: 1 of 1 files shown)


donovaly:
@chennes , can you please review and merge if OK?

chennes:
It looks to me like it's technically correct: I think it could be cleaned up a bit by using `reserve` instead of `resize` for the vector, then use `push_back` for each element, and get rid of the integer altogether. If you want to keep the int for some reason, instead of calling `resize`, make the size the argument to the constructor. If you'd prefer we can merge it as-is and I can make those changes in a second pass.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8683">pipeline status</a> for feature branch [PR_8683](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8683). Pipeline [789445975](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/789445975) was triggered at [1c8e607](https://github.com/FreeCAD/FreeCAD/pull/8683/commits/1c8e607696ec982f4bc5d2610f20d842a6afeabe). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> If you'd prefer we can merge it as-is and I can make those changes in a second pass.

Yes, this would be good because I won't have more time now and my technical capabilities are limited. (I only acted to get rid of the compiler warnings.)
By the way, DONT update to MSVC 17.5 yet! With this you get tons of Boost compilation issues. For me the only way to be productive again was to go back to version 17.4.5. (At least the MS guys wrote that they are aware now and work to make a fix release soon)

PR #8686, opened 2023-02-27
Crowdin translations

yorikvanhavre:


jbaehr:
Out of curiosity: why do we put the `*.qm` files into the repo? Those binaries are compiled from the respective `*.ts` files, right? As such, it should be possible to generate them at build time.

Having large, often-changing binaries in a git repo is kind of a time bomb as git is not designed as a repository for binary artefacts. And in that particular case, it's not even needed to have them in the repo, as they are compiler output. Or did I miss some important bit?

yorikvanhavre:
They are currently not built at compile time AFAIK. But indeed that's something we need to do at some point.. There was a trial at doing that somewhere I think in https://github.com/FreeCAD/FreeCAD-translations but I can't find it back...

PR #8699, opened 2023-02-28
[Gui] NaviCube: don't write font size on loading preferences

donovaly:
- as discussed here: https://github.com/FreeCAD/FreeCAD/pull/8567#issuecomment-1445283127 and following)

### DIFF
--- a/src/Gui/DlgSettingsNavigation.cpp
+++ b/src/Gui/DlgSettingsNavigation.cpp
@@ -211,8 +211,9 @@ void DlgSettingsNavigation::loadSettings()
     // but not accessible if there is no cube yet drawn
     if (hGrp->GetInt("FontSize", 0) == 0) {
         // the "4" is the hardcoded m_OverSample from getDefaultFontSize()
-        hGrp->SetInt("FontSize", int(0.18 * 4 * ui->prefCubeSize->value()));
-        ui->naviCubeFontSize->onRestore();
+        ui->naviCubeFontSize->setValue(int(0.18 * 4 * ui->prefCubeSize->value()));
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8699">pipeline status</a> for feature branch [PR_8699](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8699). Pipeline [790986226](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/790986226) was triggered at [8651174](https://github.com/FreeCAD/FreeCAD/pull/8699/commits/865117470f9d977b1fdd71c5638c856f0459013d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/DlgSettingsNavigation.cpp
```diff
@@ -211,8 +211,9 @@ void DlgSettingsNavigation::loadSettings()
     // but not accessible if there is no cube yet drawn
```
adrianinsaval, line 211: do you know why it isn't available? This is of course nothing that should stop this PR as it is a previous problem but duplicating the logic to calculate the font size means that we may forget to update one of them if we ever decide to change this.

donovaly: > do you know why it isn't available?

From where should the info come? There is no parameter yet written thus the info is not yet available.

> This is of course nothing that should stop this PR

OK, merging then.

PR #8716, opened 2023-03-01
[Sketcher] Layers Basic Support

abdullahtahiriyo:
This PR provides basic layers functionality in the Sketcher. It is intended to provide an alternative to Rendering Order, which is removed. This enables to continue developing layers, as rendering order interferes with the implementation.

Basically this PR provides a basic naive implementation of visual layers with 3 fixed layers, which enables to hide geometries, so as to allow to access those that are behind.

Layers is intended to be substantially more powerful, but it still requires quite some effort on SketchObject to manage User Layers, on the UI side, as well as making a new entity to handle the responsibility for managing Visual Layers. This PR advances the implementation, provides a work-around for the rendering order, and enables to do away with the task panel edit controls. This further allows to move forward in the pipeline of Sketcher PRs.

[Peek 01-03-2023 23-12.webm](https://user-images.githubusercontent.com/5249162/222277291-fa33b9a3-7276-4a5e-acf9-512719c068a3.webm)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8716">pipeline status</a> for feature branch [PR_8716](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8716). Pipeline [793453770](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/793453770) was triggered at [aa73ee3](https://github.com/FreeCAD/FreeCAD/pull/8716/commits/aa73ee3b763a6c59085b8dbbeb18961dcc878d62). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
So cool! Sketcher has improved so much this dev cycle and the good stuff just keeps coming, many thanks abdullah

src/Mod/Sketcher/Gui/SketcherSettings.cpp
```diff
@@ -49,11 +48,6 @@ SketcherSettings::SketcherSettings(QWidget* parent)
     : PreferencePage(parent), ui(new Ui_SketcherSettings)
 {
     ui->setupUi(this);
-    QGridLayout* gridLayout = new QGridLayout(ui->placeholder);
-    gridLayout->setSpacing(0);
-    gridLayout->setContentsMargins(0, 0, 0, 0);
-    form = new SketcherGeneralWidget(ui->placeholder);
```
wwmayer, line 55: The class SketcherSettings still has a private class member "form" that isn't used any more.

luzpaz: FTR 94d89e53c285c90c7c11dcfa4fd58f0011f63f4c

adrianinsaval:
@abdullahtahiriyo could you explain a little more this feature? I can assign elements to a different layer but I don't know what this does, in your video the element in layer 1 gets a different display style but I don't know how to set this. Now that there is visibility control for the geometry elements it might make sense to port over the visibility controls form the constraint widget (the eye icon and the checkbox to bind visibility to filters).
did this fully remove the edit controls widget? I tried enabling it in the preference page and it doesn't show. If it was removed the option should be removed from the preference page.

abdullahtahiriyo:
Yes. The feature is a very basic of  visual layers. It will be much more powerful in the future. The reason why we get this now is twofold: (a) I need to remove the rendering order to improve the visualisation, (b) Removing the rendering order without providing anything else would be harder for users when dealing with overlapping geometry.

The basic solutions is to provide 3 layers. Layer 1 which is the normal Layer. Layer 2 which has a discontinuous line style. Layer 3 which is hidden (so the Hidden Layer). In terms of FreeCAD ability any layer has the functionality to hide and show itself, but this is not exposed to the user yet (will be in the future). So, in this interim basic solution, when you want to hide a geometry, you assign it to the "Hidden Layer". This can be done via the checkbox (as in constraints), or via contextual menu (there is an entry called "Layers" where you can select 1, 2 or Hidden).

Yes, I have realised that the constraints of the geometry elements stay. I have noted this for when I implement the extended support. I am not sure exactly how it will work yet. 

The Elements Widget will need to be modified before that to introduce User Layers. User Layers is a logical grouping by the user of geometry, which will help some automations, such as keeping geometry imported via Carbon Copy in a User Layer group. Possibly providing some automatisation based on grouping such as synchronisation in the future (many other things need to come together first, but there is that intention). Then, the interface for Visual Layers has to be modified to actually leverage all the potential of that feature

Yes. The edit control widget is gone for good. The Grid is now in the toolbar, the autoconstraints and redundant avoiding in the constraints widget and the rendering order is gone as explained above. So there was nothing left there.

Thank you for pointing out that the preference page still has a left over option. I will remove it.

adrianinsaval:
I added this to release notes: https://wiki.freecad.org/Release_notes_1.0#Sketcher_Workbench

abdullahtahiriyo:
Thank you Adrián,

PR #8735, opened 2023-03-04
FEM: Add cylinder filter function

marioalexis84:
it is now possible to work with a cylindrical filter function.

image

---

### DIFF
--- a/src/Mod/Fem/App/AppFem.cpp
+++ b/src/Mod/Fem/App/AppFem.cpp
@@ -187,6 +187,7 @@ PyMOD_INIT_FUNC(Fem)
 
     Fem::FemPostFunction                      ::init();
     Fem::FemPostFunctionProvider              ::init();
+    Fem::FemPostCylinderFunction              ::init();
     Fem::FemPostPlaneFunction                 ::init();
     Fem::FemPostSphereFunction                ::init();
 
--- a/src/Mod/Fem/App/FemPostFunction.cpp
+++ b/src/Mod/Fem/App/FemPostFunction.cpp
@@ -59,6 +59,44 @@ DocumentObjectExecReturn* FemPostFunction::execute() {
     return DocumentObject::StdReturn;
 }
 
+// ***************************************************************************
+// cylinder function
+PROPERTY_SOURCE(Fem::FemPostCylinderFunction, Fem::FemPostFunction)
+
+FemPostCylinderFunction::FemPostCylinderFunction() : FemPostFunction()
+{
+    ADD_PROPERTY(Center, (Base::Vector3d(0.0, 0.0, 0.0)));
+    ADD_PROPERTY(Axis, (Base::Vector3d(0.0, 0.0, 1.0)));
+    ADD_PROPERTY(Radius, (5.));
+
+    m_cylinder = vtkSmartPointer<vtkCylinder>::New();
+    m_implicit = m_cylinder;
+
+    m_cylinder->SetAxis(0., 0., 1.);
+    m_cylinder->SetCenter(0., 0., 0.);
+    m_cylinder->SetRadius(5.);
+}
+
+FemPostCylinderFunction::~FemPostCylinderFunction()
+{
+}
+
+void FemPostCylinderFunction::onChanged(const Property* prop)
+{
+    if (prop == &Axis) {
+        const Base::Vector3d& vec = Axis.getValue();
+        m_cylinder->SetAxis(vec[0], vec[1], vec[2]);
+    }
+    else if (prop == &Center) {
+        const Base::Vector3d& vec = Center.getValue();
+        m_cylinder->SetCenter(vec[0], vec[1], vec[2]);
+    }
+    else if (prop == &Radius) {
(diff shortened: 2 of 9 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8735">pipeline status</a> for feature branch [PR_8735](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8735). Pipeline [796041266](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/796041266) was triggered at [63522ef](https://github.com/FreeCAD/FreeCAD/pull/8735/commits/63522ef455f7c85d614e24b93d3b77bb0ac1d74a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
@berndhahnebach do you have a moment to take a look at this PR ?

wwmayer:
From my point of view the code looks all good.

donovaly:
Nice! Looks good to me. As also approved by @wwmayer -> merging.

I see here and there some inconsistencies in the formatting (can be neglected) and will have a look.

@marioalexis84 : I have ta similar thing on my ToDo list: a box (cube) as filter function to have section views in two planes. Since you are already at it, do you think you can implement a box filter as well?

donovaly:
OK, the fix is just to initialize a vector properly: #8784

donovaly:
> @donovaly Yes, we can use two planes with variable angle to each other. I'll work on it.

Cool. However, instead of using two planes, I thought we can use the vtkCubeSource class:
https://kitware.github.io/vtk-examples/site/Cxx/GeometricObjects/Cube1/
and the SoHandleBoxDragger class manipulator
https://www.openinventor.com/reference-manuals/NewRefMan9912/RefManJava/com/openinventor/inventor/draggers/SoHandleBoxDragger.html

For the cylinder function you use the SoJackManip class as manipulator. This is a manipulator for a plane. What is the reason?
I think we should use the SoHandleBoxDragger class.

donovaly:
> For the cylinder function you use the SoJackManip class as manipulator. This is a manipulator for a plane. What is the reason?
I think we should use the SoHandleBoxDragger class.

OK, I see now that the SoHandleBoxManip would be better for the cylinder function. But with this oval cylinders are possible:
FreeCAD_FCHvljR2ng
which is not possible with VTK. The the SoJackManip is no bad choice. However, the initial cylinder is too large and the initial center is not in the center of the object.

marioalexis84:
>This is a manipulator for a plane. What is the reason?

I choose the SoJackDragger since is possible change the orientation of the cylinder. At first I also think on the SoHandleBoxDragger
>However, the initial cylinder is too large and the initial center is not in the center of the object.

Feel free to optimize these values.

> However, instead of using two planes, I thought we can use the vtkCubeSource class

Ok, we can use vtkBox (which inherits from vtkImplicitFunction) any time soon.
With the planes I meant to use 3 planes to obtain a more versatile implicit function by adding the possibility of changing the normal of each plane and obtain this type of implicit function:
image

donovaly:
> I choose the SoJackDragger since is possible change the orientation of the cylinder.

This is also possible with the SoTransformBoxManip, and meanwhile I think we should ise SoTransformBoxManip instead of SoJackDragger because with SoJackDragger you always jet a plane larger than the cylinder. This is misleading for the user.

> > However, the initial cylinder is too large and the initial center is not in the center of the object.
> Feel free to optimize these values.

I give it a try but failed. The problem is this:
FreeCAD_x3DtJq1pVN

So the center is not at the center of the object. This is however the case for a plane function.
The initial radius is OK, but the height of the cylinder should not be twice as large as the object. The same size as the object would look way more nice.
If you have an idea how to achieve this, please make a PR. (I tried 2 hours now and don't get it properly to work.)

> Ok, we can use vtkBox (which inherits from vtkImplicitFunction) any time soon. With the planes I meant to use 3 planes to obtain a more versatile implicit function by adding the possibility of changing the normal of each plane and obtain this type of implicit function: image

Hmm, isn't this overkill? I mean analyses are usually done in Cartesian, cylindrical or spherical coordinates. For the last 2 we have now a function. For Cartesian you need a cube whose faces are perpendicular to each other. I cannot imagine when trapezoidal cubes would be necessary. Can you?

donovaly:
OK, forget what I wrote about SoHandleBoxManip. SoJackDragger is OK. Only the initial position and size should be adjusted.

donovaly:
OK, here is a PR that improves the initial cylinder: #8789

This is better as in master but the optimum would be that the radius of the cylinder would be independent on the drawn height (the blue lines)

marioalexis84:
I will do a pull request with the Cube function and then work on the size of the cylinder.

donovaly:
> I will do a pull request with the Cube function and then work on the size of the cylinder.

Thanks. But note that I already made a PR for the Cylinder height: #8789
With this the height is a bit larger than the object by default to assure the cylinder borders are always visible.

PR #8776, opened 2023-03-08
App: Toponaming indexed name

### LINKED ISSUE #8765: [Issue] TopoNaming Phase 1 (card 1 of 4) - create IndexedName class

### Forums discussion

_No response_

### Version

0.21 (Development)

### Full version info

```shell
[code]
Word size of FreeCAD: 64-bit
Build type: Debug
Branch: master
[/code]
```

### Subproject(s) affected?

None

### Issue description

> The toponaming algorithm would benefit from a type to easily store and manipulate the short reference names of the elements.

## description

Create an IndexedName.cpp class that is used to save the name of an element. It will later by used by the `MappedElement` class which will be used by the `ElementMap` class.

One benefit of this change is to "simplify" the renaming/numbering of faces/etc. For example, it provides a `+=` operator to increment the number part.

It stores the "short name" of the element, such as "Face1".

For context, see issue https://github.com/FreeCAD/FreeCAD/issues/8432

## notes

Instead of a simple string, it is a string/integer combination. The string is the "type" of element. The integer is an number on it's suffix. So, instead of storing "face3", this class stores "face" and 3.

The `toString()` function resolves the class back into a simple string.

suggested private data members in the class:

```cpp
    const char* ;
    int;
```

most of the work has been done already in the branch at `src/App/MappedElement.h`

mostly, just extract this into a separate file and add good unit tests to ensure that these core functions are very reliable.

### Anything else?

_No response_

### Code of Conduct

chennes:
Fixes #8765 -- extracts IndexedName into its own header and source files, performs clang-format and clang-tidy cleanups on it, refactors to clarify and correct some issues, and adds unit tests.

- [X]  This Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

### DIFF
--- a/src/App/IndexedName.cpp
+++ b/src/App/IndexedName.cpp
@@ -1,0 +1,127 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+/****************************************************************************
+ *   Copyright (c) 2022 Zheng, Lei (realthunder) <realthunder.dev@gmail.com>*
+ *                                                                          *
+ *   This file is part of FreeCAD.                                          *
+ *                                                                          *
+ *   FreeCAD is free software: you can redistribute it and/or modify it     *
+ *   under the terms of the GNU Lesser General Public License as            *
+ *   published by the Free Software Foundation, either version 2.1 of the   *
+ *   License, or (at your option) any later version.                        *
+ *                                                                          *
+ *   FreeCAD is distributed in the hope that it will be useful, but         *
+ *   WITHOUT ANY WARRANTY; without even the implied warranty of             *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU       *
+ *   Lesser General Public License for more details.                        *
+ *                                                                          *
+ *   You should have received a copy of the GNU Lesser General Public       *
+ *   License along with FreeCAD. If not, see                                *
+ *   <https://www.gnu.org/licenses/>.                                       *
+ *                                                                          *
+ ***************************************************************************/
+
+#include "PreCompiled.h"
+
+#ifndef _PreComp_
+# include <cstdlib>
+# include <unordered_set>
+#endif
+
+#include <QHash>
+
+#include "IndexedName.h"
+
+using namespace Data;
+
+struct ByteArray
+{
+    ByteArray(const QByteArray& b)
+        :bytes(b)
+    {}
+
+    ByteArray(const ByteArray& other)
+        :bytes(other.bytes)
+    {}
+
+    ByteArray(ByteArray&& other)
+        :bytes(std::move(other.bytes))
+    {}
+
+    void mutate() const
+    {
+        QByteArray copy;
+        copy.append(bytes.constData(), bytes.size());
+        bytes = copy;
+    }
+
+    bool operator==(const ByteArray& other) const {
+        return bytes == other.bytes;
+    }
+
+    mutable QByteArray bytes;
+};
+
+struct ByteArrayHasher
+{
+    std::size_t operator()(const ByteArray& bytes) const
+    {
+        return qHash(bytes.bytes);
+    }
+
+    std::size_t operator()(const QByteArray& bytes) const
+    {
+        return qHash(bytes);
+    }
+};
+
+void IndexedName::set(
+    const char* name,
+    int len,
+    const std::vector<const char*>& types,
+    bool allowOthers)
+{
+    static std::unordered_set<ByteArray, ByteArrayHasher> NameSet;
+
+    if (len < 0)
+        len = static_cast<int>(std::strlen(name));
+    int i;
+    for (i = len - 1; i >= 0; --i) {
+        if (name[i] < '0' || name[i]>'9')
+            break;
+    }
+    ++i;
+    this->index = std::atoi(name + i);
+
+    for (int j = 0; j < i; ++j) {
+        if (name[j] == '_'
+            || (name[j] >= 'a' && name[j] <= 'z')
+            || (name[j] >= 'A' && name[j] <= 'Z'))
+            continue;
+        this->type = "";
+        return;
+    }
+
+    for (const char* type : types) {
+        int j = 0;
+        for (const char* n = name, *t = type; *n; ++n) {
+            if (*n != *t || j >= i)
+                break;
+            ++j;
+            ++t;
+            if (!*t) {
+                this->type = type;
+                return;
+            }
+        }
+    }
+
+    if (allowOthers) {
+        auto res = NameSet.insert(QByteArray::fromRawData(name, i));
+        if (res.second)
+            res.first->mutate();
+        this->type = res.first->bytes.constData();
+    }
+    else
+        this->type = "";
+}
--- a/src/App/IndexedName.cpp
+++ b/src/App/IndexedName.cpp
@@ -2,6 +2,7 @@
 
 /****************************************************************************
  *   Copyright (c) 2022 Zheng, Lei (realthunder) <realthunder.dev@gmail.com>*
+ *   Copyright (c) 2023 FreeCAD Project Association                         *
  *                                                                          *
  *   This file is part of FreeCAD.                                          *
  *                                                                          *
@@ -21,6 +22,7 @@
  *                                                                          *
  ***************************************************************************/
 
+// NOLINTNEXTLINE
 #include "PreCompiled.h"
 
 #ifndef _PreComp_
@@ -28,100 +30,93 @@
 # include <unordered_set>
 #endif
 
-#include <QHash>
-
 #include "IndexedName.h"
 
 using namespace Data;
 
-struct ByteArray
+/// Check whether the input character is an underscore or an ASCII letter a-Z or A-Z
+inline bool isInvalidChar(char test)
 {
-    ByteArray(const QByteArray& b)
-        :bytes(b)
-    {}
+    return test != '_' && (test < 'a' || test > 'z' ) && (test < 'A' || test > 'Z');
+}
 
-    ByteArray(const ByteArray& other)
-        :bytes(other.bytes)
-    {}
-
-    ByteArray(ByteArray&& other)
-        :bytes(std::move(other.bytes))
-    {}
-
-    void mutate() const
-    {
-        QByteArray copy;
-        copy.append(bytes.constData(), bytes.size());
-        bytes = copy;
-    }
-
-    bool operator==(const ByteArray& other) const {
-        return bytes == other.bytes;
-    }
-
-    mutable QByteArray bytes;
-};
-
-struct ByteArrayHasher
+/// Get the integer suffix of name. Returns a tuple of (suffix, suffixPosition). Calling code
+/// should check to ensure that suffixPosition is not equal to nameLength (in which case there was no
+/// suffix).
+///
+/// \param name The name to check
+/// \param nameLength The length of the string in name
+/// \returns An integer pair of the suffix itself and the position of that suffix in name
+std::pair<int,int> getIntegerSuffix(const char *name, int nameLength)
 {
-    std::size_t operator()(const ByteArray& bytes) const
-    {
-        return qHash(bytes.bytes);
-    }
+    int suffixPosition {nameLength - 1};
 
-    std::size_t operator()(const QByteArray& bytes) const
-    {
-        return qHash(bytes);
+    for (; suffixPosition >= 0; --suffixPosition) {
+        // When we support C++20 we can use std::span<> to eliminate the clang-tidy warning
+        // NOLINTNEXTLINE cppcoreguidelines-pro-bounds-pointer-arithmetic
+        if (!isdigit(name[suffixPosition])) {
+            break;
+        }
     }
-};
+    ++suffixPosition;
+    int suffix {0};
+    if (suffixPosition < nameLength) {
+        // When we support C++20 we can use std::span<> to eliminate the clang-tidy warning
+        // NOLINTNEXTLINE cppcoreguidelines-pro-bounds-pointer-arithmetic
+        suffix = std::atoi(name + suffixPosition);
+    }
+    return std::make_pair(suffix, suffixPosition);
+}
 
 void IndexedName::set(
     const char* name,
-    int len,
-    const std::vector<const char*>& types,
+    int length,
+    const std::vector<const char*>& allowedNames,
     bool allowOthers)
 {
+    // Storage for names that we weren't given external storage for
     static std::unordered_set<ByteArray, ByteArrayHasher> NameSet;
 
-    if (len < 0)
-        len = static_cast<int>(std::strlen(name));
-    int i;
-    for (i = len - 1; i >= 0; --i) {
-        if (name[i] < '0' || name[i]>'9')
-            break;
+    if (length < 0) {
+        length = static_cast<int>(std::strlen(name));
     }
-    ++i;
-    this->index = std::atoi(name + i);
+    // Name typically ends with an integer: find that integer
+    auto [suffix, suffixPosition] = getIntegerSuffix(name, length);
+    if (suffixPosition < length) {
+        this->index = suffix;
+    }
 
-    for (int j = 0; j < i; ++j) {
-        if (name[j] == '_'
-            || (name[j] >= 'a' && name[j] <= 'z')
-            || (name[j] >= 'A' && name[j] <= 'Z'))
-            continue;
+    // Make sure that every character is either an ASCII letter (upper or lowercase), or an
+    // underscore. If any other character appears, reject the entire string.
+    // When we support C++20 we can use std::span<> to eliminate the clang-tidy warning
+    // NOLINTNEXTLINE cppcoreguidelines-pro-bounds-pointer-arithmetic
+    if (std::any_of(name, name+suffixPosition, isInvalidChar)) {
         this->type = "";
         return;
     }
 
-    for (const char* type : types) {
-        int j = 0;
-        for (const char* n = name, *t = type; *n; ++n) {
-            if (*n != *t || j >= i)
-                break;
-            ++j;
-            ++t;
-            if (!*t) {
-                this->type = type;
-                return;
-            }
+    // If a list of allowedNames was provided, see if our set name matches one of those allowedNames: if it
+    // does, reference that memory location and return.
+    for (const auto *typeName : allowedNames) {
+        if (std::strncmp(name, typeName, suffixPosition) == 0) {
+            this->type = typeName;
+            return;
         }
     }
 
+    // If the type was NOT in the list of allowedNames, but the caller has set the allowOthers flag to
+    // true, then add the new type to the static NameSet (if it is not already there).
     if (allowOthers) {
-        auto res = NameSet.insert(QByteArray::fromRawData(name, i));
-        if (res.second)
-            res.first->mutate();
+        auto res = NameSet.insert(ByteArray(QByteArray::fromRawData(name, suffixPosition)));
+        if (res.second /*The insert succeeded (the type was new)*/) {
+            // Make sure that the data in the set is a unique (unshared) copy of the text
+            res.first->ensureUnshared();
+        }
         this->type = res.first->bytes.constData();
     }
-    else
+    else {
+        // The passed-in type is not in the allowed list, and allowOthers was not true, so don't
+        // store the type
         this->type = "";
+    }
 }
--- a/src/App/IndexedName.h
+++ b/src/App/IndexedName.h
@@ -2,6 +2,7 @@
 
 /****************************************************************************
  *   Copyright (c) 2022 Zheng, Lei (realthunder) <realthunder.dev@gmail.com>*
+ *   Copyright (c) 2023 FreeCAD Project Association                         *
  *                                                                          *
  *   This file is part of FreeCAD.                                          *
  *                                                                          *
@@ -21,9 +22,8 @@
  *                                                                          *
  ***************************************************************************/
 
-
-#ifndef _IndexedName_h_
-#define _IndexedName_h_
+#ifndef APP_INDEXEDNAME_H
+#define APP_INDEXEDNAME_H
 
 #include <cassert>
 #include <cstring>
@@ -32,6 +32,7 @@
 #include <vector>
 
 #include <QByteArray>
+#include <QHash>
 
 #include "FCGlobal.h"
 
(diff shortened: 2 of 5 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [799421813](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/799421813) was triggered at [f250ee2](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/f250ee238489380ccd8198ce24abc00197c57857). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [801918024](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/801918024) was triggered at [fb9e05c](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/fb9e05cf7658c90105f552ac3264a065e99f1a3c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [802052203](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/802052203) was triggered at [482e2c4](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/482e2c46fca8349f92d0b4fc134a63870a4e4407). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [802058827](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/802058827) was triggered at [a776ff4](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/a776ff4de93ec6ab87325f4814250dadc5f1fc65). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [802946053](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/802946053) was triggered at [cdcfe88](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/cdcfe88a156fdd6d762bf148e3836bf566bf165b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/App/IndexedName.h
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    // Pre-increment operator: increases the index of this element by one.
+    IndexedName & operator++()
+    {
+        ++this->index;
+        return *this;
+    }
+
+    // Pre-decrement operator: increases the index of this element by one. Must not make the index
```
chennes, line 159: Decreases.

src/App/IndexedName.cpp
```diff
@@ -0,0 +1,170 @@
...
+// Get the integer suffix of name. Returns a tuple of (suffix, suffixPosition). Calling code
+// should check to ensure that suffixPosition is not equal to length (in which case there was no
+// suffix).
+std::pair<int,int> getIntegerSuffix(const char *name, int length)
+{
+    int suffixPosition {length - 1};
+    for (; suffixPosition >= 0; --suffixPosition) {
+        // When we support C++20 we can use std::span<> to eliminate the clang-tidy warning
+        // NOLINTNEXTLINE cppcoreguidelines-pro-bounds-pointer-arithmetic
+        if (name[suffixPosition] < '0' || name[suffixPosition]>'9') {
```
chennes, line 105: Should use `isdigit`.

src/App/IndexedName.cpp
```diff
@@ -0,0 +1,170 @@
...
+    // underscore. If any other character appears, reject the entire string.
+    // When we support C++20 we can use std::span<> to eliminate the clang-tidy warning
+    // NOLINTNEXTLINE cppcoreguidelines-pro-bounds-pointer-arithmetic
+    if (std::any_of(name, name+suffixPosition, isInvalidChar)) {
+        this->type = "";
+        return;
+    }
+
+    // If a list of types was provided, see if our set name starts with one of those types: if it
+    // does, truncate it to that type and return
```
chennes, line 147: Outdated comment.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [803046427](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803046427) was triggered at [be83c15](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/be83c15aacb9da7548d092fd9d1cd85e575102c4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [803052400](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803052400) was triggered at [d7f8070](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/d7f807098f40248cae4fda1210e04c2c830ec030). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/App/IndexedName.cpp
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    std::size_t operator()(const QByteArray& bytes) const
+    {
+        return qHash(bytes);
+    }
+};
+
+inline bool isInvalidChar(char test)
+{
```
wwmayer, line 92: Check this for potential problems with std::isalpha: https://en.cppreference.com/w/cpp/string/byte/isalpha
If you want to make sure to have a char in the range a-z or A-Z it might be better to do that directly.

src/App/IndexedName.cpp
```diff
@@ -0,0 +1,171 @@
...
+    return test != '_' && (std::isalpha(test) == 0);
+}
+
+// Get the integer suffix of name. Returns a tuple of (suffix, suffixPosition). Calling code
+// should check to ensure that suffixPosition is not equal to length (in which case there was no
+// suffix).
+std::pair<int,int> getIntegerSuffix(const char *name, int length)
```
wwmayer, line 99: I guess `length` is the length of `name`. Maybe this should be documented to avoid to wrongly use this function.

tests/src/App/IndexedName.cpp
```diff
@@ -0,0 +1,243 @@
...
+TEST_F(IndexedNameTest, isNull)
+{
+    auto indexedName = Data::IndexedName();
+    EXPECT_TRUE(indexedName.isNull());
+    indexedName = Data::IndexedName("",1);
+    EXPECT_TRUE(indexedName.isNull());
+    indexedName = Data::IndexedName("EDGE",1);
+    EXPECT_FALSE(indexedName.isNull());
+    indexedName = Data::IndexedName("INVALID42NAME",1);
+    EXPECT_TRUE(indexedName.isNull());
```
wwmayer, line 211: IMO, what's missing here is a test with a string with non-letter characters, e.g. `indexedName = Data::IndexedName(".EDGE",1);`

chennes: Good idea, thanks.

src/App/IndexedName.cpp
chennes: Is it the locale that might be the issue, or the missing cast to unsigned char? I don't mind switching back to the direct comparison, I was thinking that using isalpha made it clearer (and shorter), but it's a very minor thing.

chennes: (I switched it to a direct comparison in d4d9ce7c18176d113713c330882ca4a10b68d55a).

src/App/IndexedName.cpp
chennes: I renamed the variable and added clearer documentation in d4d9ce7c18176d113713c330882ca4a10b68d55a

tests/src/App/IndexedName.cpp
chennes: Added in d4d9ce7c18176d113713c330882ca4a10b68d55a

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [803369254](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803369254) was triggered at [d4d9ce7](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/d4d9ce7c18176d113713c330882ca4a10b68d55a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/App/IndexedName.cpp
AjinkyaDahale: If locale is an issue I imagine `isalpha` or a `Qt::String` equivalent would be better since it might check beyond the English alphabet. Or this same property could make it worse if only a-zA-Z are to be allowed.

wwmayer: The (unsigned)char is probably not a problem because the come from a string. But according to the example code isalpha() can theoretically be a problem when it allows characters that shouldn't be allowed like German umlauts.

But I have to admit that on my system the test code neither for C nor de_DE locale the `ß` was accepted.

chennes:
OK, I've cleaned up the commit history here: my plan is to keep @realthunder's original code intact in the first commit, and apply a single commit on top of that making the changes we've discussed. The original code is not _quite_ exactly RTs because I let the autoformatter run on it, but all the substantial refactoring work, linter cleanup, and bug fixes are separated.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [803710368](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803710368) was triggered at [756b251](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/756b25151b2503a11683f03718f1189b2c33461d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

tests/src/App/IndexedName.cpp
```diff
@@ -0,0 +1,248 @@
...
+}
+
+TEST_F(IndexedNameTest, compoundAssignmentOperator)
+{
+    constexpr int base{42};
+    constexpr int offset{10};
+    // Only += is defined for this class
+    auto indexedName = Data::IndexedName("EDGE",base);
+    indexedName += offset;
+    EXPECT_EQ(indexedName.getIndex(), base+offset);
```
JohnAD, line 124: This will come off a being picky, sorry about that. In assertions it is good practice to avoid operations _just in case_ there is some non-obvious operator overloading somewhere.  So, it might better as `EXPECT_EQ(indexedName.getIndex(), 52);` . But that is paranoia speaking, feel free to leave it as is :).

tests/src/App/IndexedName.cpp
```diff
@@ -0,0 +1,248 @@
...
+    auto indexedName = Data::IndexedName();
+    EXPECT_FALSE(static_cast<bool>(indexedName));
+    indexedName = Data::IndexedName("",1);
+    EXPECT_FALSE(static_cast<bool>(indexedName));
+    indexedName = Data::IndexedName("EDGE",1);
+    EXPECT_TRUE(static_cast<bool>(indexedName));
+    indexedName = Data::IndexedName("INVALID42NAME",1);
+    EXPECT_FALSE(static_cast<bool>(indexedName));
+    indexedName = Data::IndexedName(".EDGE",1);  // Invalid name
+    EXPECT_FALSE(static_cast<bool>(indexedName));
```
JohnAD, line 228: To help future programmers,  you might also add a test showing it's nature in use as one of the tests. aka,

```cpp
indexedName = Data::IndexedName(".EDGE",1);  // Invalid name
if (indexedName) {
  FAIL() << "Should have resolved to FALSE.";
}
```

JohnAD:
Reviewing the testing, the two new(-ish) functions don't have a test: `toString` and `appendToStringBuffer`

tests/src/App/IndexedName.cpp
chennes: Pickiness is welcome here, these couple of classes are the core of the whole thing - IMO it's worth the time to get them right. I wrote it like that mostly to shut up the linter (which complains about the magic number otherwise), but I can just as easily add a NOLINT directive. Really, tests are littered with magic numbers, I am inclined to add that ignore to the entire file.

JohnAD:
And, something for open-ended discussion. Over the years, I've started always marking test code with comments marking the AAA pattern. It acts as a quick "visual" hint to future me about "where to look for things".

So, instead of

```cpp
// Names must not contain numbers in the middle:
TEST_F(IndexedNameTest, constructionNumberInName)
{
    const int testIndex {42};
    std::string testName;
    testName += "Test" + std::to_string(testIndex) + "Name";
    auto indexedName = Data::IndexedName(testName.c_str(), testIndex);
    EXPECT_STREQ(indexedName.getType(), "");
}
```
I'd add a few lines like:

```cpp
// Names must not contain numbers in the middle:
TEST_F(IndexedNameTest, constructionNumberInName)
{
    // Arrange
    const int testIndex {42};
    std::string testName;
    testName += "Test" + std::to_string(testIndex) + "Name";

    // Act
    auto indexedName = Data::IndexedName(testName.c_str(), testIndex);

    // Assert
    EXPECT_STREQ(indexedName.getType(), "");
}
```

Visually, my eye "jumps" to the `// Act` comment in the middle to first to see what is happening. If needed I'll then look upwards for what is being setup for the test (`// Arrange`). Otherwise, my eye moves down to see what is being tested that caused the failure (`// Assert`). Not all tests have the the Arrange and Assert parts. That is OK. But it always has an "Act" part. 

This isn't a required thing at all. Test-driven development works just fine without it. It is just a suggestion to talk about.

chennes:
I've used the given-when-then pattern before -- it sounds like AAA is much the same thing. Typically in GWT I'd create a function that sets up the givens, and another that does the asserts, rather than using comments. It felt clunky in this application, though. Maybe using comments like that strikes a good balance.

tests/src/App/IndexedName.cpp
chennes: Good idea, will do.

JohnAD:
I've seen GWT terms when using tools like _SpecFlow_. In theory it is subtly different from AAA, but I've never seen anyone really explain how. I've always used them in the same manner.

I suspect the real difference is that GWT doesn't use a fancy word like "assert". Since tools like SpecFlow are meant to be readable by upper management and, possibly, the court system system should somebody try to sue a software tool.

Thankfully, the open-source licenses shields us from that complication.

chennes:
OK, I have addressed @JohnAD's comments in two new commits: the first just adds the missing tests and tweaks a couple of existing tests. The second commit adds "Arrange-Act-Assert" comments to most of the tests. Comments?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [803916238](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803916238) was triggered at [5439d6f](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/5439d6fce9e461adcc89ac96d2328db9cb12311b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

tests/src/App/IndexedName.cpp
```diff
@@ -0,0 +1,421 @@
...
+{
+    // Arrange
+    auto indexedName1 = Data::IndexedName("EDGE42");
+    auto indexedName2 = Data::IndexedName("EDGE42");
+
+    // Assert
+    EXPECT_EQ(indexedName1.compare(indexedName2), 0);
+    EXPECT_TRUE(indexedName1 == indexedName2);
+    EXPECT_FALSE(indexedName1 != indexedName2);
+    EXPECT_FALSE(indexedName1 < indexedName2);
```
JohnAD, line 218: A few possiblities for the block starting on 214. Either mark the block comment as `// Act & Assert`. Which is reasonable. Or parlay it out like

```cpp
    // Act
    auto compareResult = indexedName1.compare(indexedName2)
    auto eqResult = (indexedName1 == indexedName2);
    auto neResult = (indexedName1 != indexedName2);
    auto ltResult = (indexedName1 < indexedName2);

    // Assert
    EXPECT_EQ(compareResult, 0);
    EXPECT_TRUE(eqResult);
    EXPECT_FALSE(neResult);
    EXPECT_FALSE(ltResult);
```

Either is good IMO.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [803940949](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803940949) was triggered at [1ba61ae](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/1ba61ae88d16feda8f600da7a1175f988b3f3854). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Right now, the IndexedName.h code has very good test coverage -- my tool says 94% of lines and 90% of branches are covered, and the ones that aren't are the various asserts. The major thing that's missing from the tests now is anything to do with the ByteArray and its hasher, since those are private to the cpp file and can't be accessed by the tests. I am also not testing the static NameMap operation -- that's really an implementation detail of the `set()` function, I wasn't sure if it was reasonable to test it (e.g. by ensuring that the same memory locations are used for two identical names).

This leads to two questions, though: first, should I refactor to make ByteArray accessible to testing code? And second, should I test the operation of the NameMap?

JohnAD:
Testing private functions is something of a judgement call. Generally speaking, the public functions "test" those by using them behind-the-scenes. That is, they are tested indirectly by producing good results.

In this particular case, however, I don't see a clean way to verify repeated memory locations without creating test-only public functions.

I'm inclined to simply make a note about this in the cards and push this concern to the Optimization phase of all this. We might have different perspective by then.

JohnAD:
Regarding NameMap, mind giving me a file and line number reference?

chennes:
Yes, sorry, `NameSet` -- it would be possible for it to be not working at all and for everything on the outside to look like it is (except taking up more memory than expected). 

Getting `ByteArray` under test is just a matter of moving it into the header file, but getting is used in `NameSet` tested probably requires making a public interface for that functionality.

JohnAD:
In the past, when I've encountered this kind of thing, I either:

1) Create weird `test_only_{methods}` public methods.
2) Extrapolate the code into a new class and have the first class call the second class.

Certainly (1) would work with some odd `test_only_get_memory_addr_of_...` methods. Messy, but functional.

And option (2) seems extreme for so little code only used by one class.

But, I just learned of this: https://github.com/google/googletest/blob/main/docs/advanced.md#testing-private-code

Basically, there is a `FRIEND_TEST` that allows access to the private methods etc. I've not tried it, but that might be a good solution.

JohnAD:
I like it. The more I think about it, the more I want to know if some future change causes the memory reuse to stop working.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [804943314](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/804943314) was triggered at [1595239](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/1595239f7fe0f580308b123dcc261086c9fad55f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
OK, I've re-squashed all of this discussion so we're back down to the two planned commits. I think this is ready to go in, but I'll wait until later today or tomorrow to actually pull the trigger on the merge in case anyone else want to weigh in. It does include a small change to the public interface compared to @realthunder's original code that will have to be propagated through the remainder of the PRs, and it introduces what I hope will become the standard construction of the tests we implement during this merge (@Pesc0 you might give it a look to see if you have any feedback there).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [805000729](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/805000729) was triggered at [6d93c3e](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/6d93c3e6e56d859a107d949bd6a500ecc6f966e0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Pesc0:
Hey I'm not the expert here, best I can do is take note and try to replicate in the other PR!

I have only one question:
```
char operator[](int input) const
    {
        assert(input >= 0);
        assert(input < static_cast<int>(std::strlen(this->type)));
        return this->type[input];
    }
```
Is it enough to use assert? Whats the thought process behind this kind of decision? Wouldn't it be more robust to throw an exception?

chennes:
Speed. The asserts will compile out in release mode. It would certainly be more robust to put in "real" error handling, I have not looked closely at how often this operator is used.

AjinkyaDahale (changes requested):

src/App/IndexedName.h
```diff
@@ -0,0 +1,302 @@
...
+// consist of upper- and lower-case ASCII characters and the underscore ('_') character. The index
+// must be a positive integer. The string representation of this IndexedName is the name followed by
+// the index, with no spaces between: an IndexedName may be constructed from this string. For
+// example "EDGE1" or "FACE345" might be the names of elements that use an IndexedName. If there is
+// then an "EDGE2", only a pointer to the original stored name "EDGE" is retained.
+//
+// The memory efficiency of the class comes from re-using the same character storage for names that
+// match, while retaining their differing indices. This is achieved by either using user-provided
+// const char * names (provided as a list of typeNames and presumed to never be deallocated), or by
+// maintaining an internal list of names that have been used before, and can be re-used later.
```
AjinkyaDahale, line 55:
```suggestion
/// The IndexedName class provides a very memory-efficient data structure to hold a name and an index
/// value, and to perform various comparisons and validations of those values. The name must only
/// consist of upper- and lower-case ASCII characters and the underscore ('_') character. The index
/// must be a positive integer. The string representation of this IndexedName is the name followed by
/// the index, with no spaces between: an IndexedName may be constructed from this string. For
/// example "EDGE1" or "FACE345" might be the names of elements that use an IndexedName. If there is
/// then an "EDGE2", only a pointer to the original stored name "EDGE" is retained.
///
/// The memory efficiency of the class comes from re-using the same character storage for names that
/// match, while retaining their differing indices. This is achieved by either using user-provided
/// const char * names (provided as a list of typeNames and presumed to never be deallocated), or by
/// maintaining an internal list of names that have been used before, and can be re-used later.
```

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776">pipeline status</a> for feature branch [PR_8776](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8776). Pipeline [805729909](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/805729909) was triggered at [ff34611](https://github.com/FreeCAD/FreeCAD/pull/8776/commits/ff34611f3d0dfbe4c7e903ce5732521a0d060ffb). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/App/IndexedName.cpp
```diff
@@ -0,0 +1,121 @@
...
+    return std::make_pair(suffix, suffixPosition);
+}
+
+void IndexedName::set(
+    const char* name,
+    int len,
+    const std::vector<const char*>& types,
+    bool allowOthers)
+{
+    // Storage for names that we weren't given external storage for
```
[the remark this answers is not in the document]
chennes: In most cases IndexedName is constructed using the `fromConst` method. In a few instances, like:
https://github.com/FreeCAD/FreeCAD/blob/140236c7747aecaedd7e199a05ee81c1081378d9/src/Mod/Sketcher/App/SketchObject.cpp#L9606
It is used with a vector of types, but is instructed _not_ to allow non-matching types. And here is an example where the internal storage is allowed:
https://github.com/FreeCAD/FreeCAD/blob/140236c7747aecaedd7e199a05ee81c1081378d9/src/Mod/Sketcher/App/SketchObject.cpp#L9581

src/App/IndexedName.h
```diff
@@ -0,0 +1,302 @@
...
+#include <QByteArray>
+#include <QHash>
+
+#include "FCGlobal.h"
+
+
+namespace Data
+{
+
+// The IndexedName class provides a very memory-efficient data structure to hold a name and an index
```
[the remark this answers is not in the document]
chennes: Yeah, I guess so. I'll have to make another pass through all the comments and reformat. Thanks!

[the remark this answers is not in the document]
chennes: Done, thanks!

src/App/IndexedName.h
```diff
@@ -0,0 +1,339 @@
...
+ *   Lesser General Public License for more details.                        *
+ *                                                                          *
+ *   You should have received a copy of the GNU Lesser General Public       *
+ *   License along with FreeCAD. If not, see                                *
+ *   <https://www.gnu.org/licenses/>.                                       *
+ *                                                                          *
+ ***************************************************************************/
+
+
+#ifndef IndexedName_h_
```
AjinkyaDahale, line 26: This variable is named different than others in the same folder. Shouldn't it be `APP_INDEXED_NAME_H` or similar?

chennes: Yes, I agree. I will change it.

Issue #8220, opened 2023-01-21
[Issue] Sketcher angle constraint is not displayed where expected in 0.20.2 Windows LibPack build


Almisuifre:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.20 (Release)

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.2.29177 +426 (Git)
Build type: Release
Branch: (HEAD detached from 0.20.2)
Hash: 930dd9a76203a3260b1e6256c70c1c3cad8c5cb8
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: French/France (fr_FR)
```

### Subproject(s) affected?

Sketcher

### Problem description

Hi,
I came to see you because I loaded the version 0.20.2.
I notice a nice bug positioning on the constraints of angles. See for yourself!
They seem unarmed attached to the central axis instead of the usual position. This is rather embarrassing because we no longer know what the constraint applies to.
![image001](https://user-images.githubusercontent.com/8313541/213866627-857b247c-7db1-4a95-9421-a12b014e3a90.png)
![image](https://user-images.githubusercontent.com/8313541/213866630-46b7c771-cf28-45c4-90d7-c671a32ea38b.png)
![image002](https://user-images.githubusercontent.com/8313541/213866635-35a8330a-1b6d-4ec4-b8f2-78276b7b47e1.png)

_The steps that lead to the bug:_
Start FreeCAD 0.20.2
New tab
New body
Select a starting face (construction line)
Switch to sketcher mode
Draw a picture
Place the angle rib. It appears in the wrong place.

This produces a bad experience because we no longer know what the constraint is attached to. It seems to be frozen between the central axes and the angle we want to set. It seems to stand out from the selection.

### Anything else?

https://discord.com/channels/870877411049357352/873944627806339082/1065907714242387978

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
Can you please share an example file?

Almisuifre:
Okay, so here’s a video that shows the problem, as well as a file generated during the video.
[Youtube Bug contraint](https://youtu.be/S9CqsQkTgdU)
[...]

PR #7888, opened 2022-11-27
Gui: overlay dock widgets user interface

realthunder:
Closes #7756 

Forum thread https://forum.freecadweb.org/viewtopic.php?f=34&t=73280

PR #8781, opened 2023-03-08
[NotificationArea] Notificationarea

abdullahtahiriyo:
After the pre-requisites are merged, this PR brings the actual NotificationArea.

After this PR, there are two other PRs left for NotificationArea: 
- A first one relates to general functions to help WB to move from blocking QMessage::warning to the  non-intrusive framework.
- A first migration of the sketcher constraint creation to the new framework using those general functions.

What is the Notifications Area?

It consists of:
(a) Non-intrusive messages
(b) A widget showing errors, warnings, critical messages and notifications

How does it look like?

[Peek 2023-03-08 15-33.webm](https://user-images.githubusercontent.com/5249162/223741009-d9f267a6-0ea1-4a57-9590-e740c22e3fd2.webm)

### DIFF
--- a/src/Base/Observer.cpp
+++ b/src/Base/Observer.cpp
@@ -1,0 +1,34 @@
+/***************************************************************************
+ *   Copyright (c) 2023 Abdullah Tahiri <abdullah.tahiri.yo@gmail.com>     *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU Library General Public License (LGPL)   *
+ *   as published by the Free Software Foundation; either version 2 of     *
+ *   the License, or (at your option) any later version.                   *
+ *   for detail see the LICENCE text file.                                 *
+ *                                                                         *
+ *   FreeCAD is distributed in the hope that it will be useful,            *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with FreeCAD; if not, write to the Free Software        *
+ *   Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  *
+ *   USA                                                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "PreCompiled.h"
+
+#include "Observer.h"
+
+
+namespace Base {
+
+template class BaseExport Observer<const char*>;
+template class BaseExport Subject<const char*>;
+
+} //namespace Base
--- a/src/Gui/NotificationArea.cpp
+++ b/src/Gui/NotificationArea.cpp
@@ -1,0 +1,993 @@
+/***************************************************************************
+ *   Copyright (c) 2022 Abdullah Tahiri <abdullah.tahiri.yo@gmail.com>     *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This library is free software; you can redistribute it and/or         *
+ *   modify it under the terms of the GNU Library General Public           *
+ *   License as published by the Free Software Foundation; either          *
+ *   version 2 of the License, or (at your option) any later version.      *
+ *                                                                         *
+ *   This library  is distributed in the hope that it will be useful,      *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "PreCompiled.h"
+
+#ifndef _PreComp_
+#include <QAction>
+#include <QActionEvent>
+#include <QApplication>
+#include <QEvent>
+#include <QHBoxLayout>
+#include <QHeaderView>
+#include <QMenu>
+#include <QMessageBox>
+#include <QStringList>
+#include <QThread>
+#include <QTimer>
+#include <QTreeWidget>
+#include <QWidgetAction>
+#include <memory>
+#include <mutex>
+#endif
+
+#include <App/Application.h>
+#include <Base/Console.h>
+
+#include "Application.h"
+#include "BitmapFactory.h"
+#include "MDIView.h"
+#include "MainWindow.h"
+#include "NotificationBox.h"
+
+#include "NotificationArea.h"
+
+using namespace Gui;
+
+using Connection = boost::signals2::connection;
+
+namespace bp = boost::placeholders;
+
+class NotificationAreaObserver;
+
+namespace Gui
+{
+/** PImpl idiom structure having all data necessary for the notification area */
+struct NotificationAreaP
+{
+    // Structure holding all variables necessary for the Notification Area.
+    // Preference parameters are updated by NotificationArea::ParameterObserver
+
+    /** @name Non-intrusive notifications parameters */
+    //@{
+    /// Parameter controlled
+    int maxOpenNotifications = 15;
+    /// Parameter controlled
+    unsigned int notificationExpirationTime = 10000;
+    /// minimum time that the notification will remain unclosed
+    unsigned int minimumOnScreenTime = 5000;
+    /// Parameter controlled
+    bool notificationsDisabled = false;
+    /// Control of confirmation mechanism (for Critical Messages)
+    bool requireConfirmationCriticalMessageDuringRestoring = true;
+    //@}
+
+    /** @name Widget parameters */
+    //@{
+    /// Parameter controlled - maximum number of message allowed in the notification area widget (0
+    /// means no limit)
+    int maxWidgetMessages = 1000;
+    /// User notifications get automatically removed from the Widget after the non-intrusive
+    /// notification expiration time
+    bool autoRemoveUserNotifications;
+    //@}
+
+    /** @name Notification rate control */
+    //@{
+    /* Control rate of updates of non-intrusive messages (avoids visual artifacts when messages are
+     * constantly being received) */
+    // Timer to delay notification until a minimum time between two consecutive messages have lapsed
+    QTimer inhibitTimer;
+    // The time between two consecutive messages forced by the inhibitTimer
+    const unsigned int inhibitNotificationTime = 250;
+    //@}
+
+    // Access control
+    std::mutex mutexNotification;
+
+    // Pointers to widgets (no ownership)
+    QMenu* menu;
+    QWidgetAction* notificationaction;
+
+    /** @name Resources */
+    //@{
+    /// Base::Console Message observer
+    std::unique_ptr<NotificationAreaObserver> observer;
+    Connection finishRestoreDocumentConnection;
+    /// Preference Parameter observer
+    std::unique_ptr<NotificationArea::ParameterObserver> parameterObserver;
+    //@}
+};
+
+}// namespace Gui
+
+
+/******************* Resource Management *****************************************/
+
+/** Simple class to manage Notification Area Resources*/
+class ResourceManager
+{
+
+private:
+    ResourceManager()
+    {
+        error = BitmapFactory().pixmapFromSvg(":/icons/edit_Cancel.svg", QSize(16, 16));
+        warning = BitmapFactory().pixmapFromSvg(":/icons/Warning.svg", QSize(16, 16));
+        critical = BitmapFactory().pixmapFromSvg(":/icons/critical-info.svg", QSize(16, 16));
+        info = BitmapFactory().pixmapFromSvg(":/icons/info.svg", QSize(16, 16));
+    }
+
+    inline static const auto& getResourceManager()
+    {
+        static ResourceManager manager;
+
+        return manager;
+    }
+
+public:
+    inline static auto ErrorPixmap()
+    {
+        auto rm = getResourceManager();
+        return rm.error;
+    }
+
+    inline static auto WarningPixmap()
+    {
+        auto rm = getResourceManager();
+        return rm.warning;
+    }
+
+    inline static auto CriticalPixmap()
+    {
+        auto rm = getResourceManager();
+        return rm.critical;
+    }
+
+    inline static auto InfoPixmap()
+    {
+        auto rm = getResourceManager();
+        return rm.info;
+    }
+
+private:
+    QPixmap error;
+    QPixmap warning;
+    QPixmap critical;
+    QPixmap info;
+};
+
+/******************** Console Messages Observer (Console Interface) ************************/
+
+/** This class listens to all messages sent via the console interface and
+   feeds the non-intrusive notification system and the notifications widget */
+class NotificationAreaObserver: public Base::ILogger
+{
+public:
+    NotificationAreaObserver(NotificationArea* notificationarea);
+    ~NotificationAreaObserver() override;
+
+    /// Function that is called by the console interface for this observer with the message
+    /// information
+    void SendLog(const std::string& notifiername, const std::string& msg,
+                 Base::LogStyle level) override;
+
+    /// Name of the observer
+    const char* Name() override
+    {
+        return "NotificationAreaObserver";
+    }
+
+private:
+    NotificationArea* notificationArea;
+};
+
+NotificationAreaObserver::NotificationAreaObserver(NotificationArea* notificationarea)
+    : notificationArea(notificationarea)
+{
+    Base::Console().AttachObserver(this);
+    bLog = false;                  // ignore log messages
+    bMsg = false;                  // ignore messages
+    bNotification = true;          // activate user notifications
+    bTranslatedNotification = true;// activate translated user notifications
+}
+
+NotificationAreaObserver::~NotificationAreaObserver()
+{
+    Base::Console().DetachObserver(this);
+}
+
+void NotificationAreaObserver::SendLog(const std::string& notifiername, const std::string& msg,
+                                       Base::LogStyle level)
+{
+    // 1. As notification system is shared with report view and others, the expectation is that any
+    // individual error and warning message will end in "\n". This means the string must be stripped
+    // of this character for representation in the notification area.
+    // 2. However, any message marked with the QT_TRANSLATE_NOOT macro with the "Notifications"
+    // context, shall not include
+    // "\n", as this generates problems with the translation system. Then the string must be
+    // stripped of "\n" before translation.
+
+    auto simplifiedstring =
+        QString::fromStdString(msg)
+            .trimmed();// remove any leading and trailing whitespace character ('\n')
+
+    // avoid processing empty strings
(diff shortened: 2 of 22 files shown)


abdullahtahiriyo:
@donovaly 

Could you at some time  compile it for Windows to ensure it is not breaking anything Windows related? Thanks.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781">pipeline status</a> for feature branch [PR_8781](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781). Pipeline [799905530](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/799905530) was triggered at [90e9236](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/90e92362c0eb4f25c42612b9b839caa123d6a83b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
Build failures are about FEM tests. I will try to rebase and retrigger the CI.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781">pipeline status</a> for feature branch [PR_8781](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781). Pipeline [800209513](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/800209513) was triggered at [7aa33cb](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/7aa33cb5326ca8583da6d1a335ea12c6b74cb44f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
I need to investigate an unexpected issue:

```
The femelement_table needs to be calculated.
QObject::startTimer: Timers cannot be started from another thread
```

donovaly:
> Could you at some time compile it for Windows to ensure it is not breaking anything Windows related? 

I can do so but at the moment it also doesn't work for the CIs, see the Gui unit test results.

abdullahtahiriyo:
@donovaly 

Thank you Uwe. 

There was an issue with multithreading in FEM gui unit tests. That I have just fixed.

Whenever you have time.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781">pipeline status</a> for feature branch [PR_8781](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781). Pipeline [801338113](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/801338113) was triggered at [b86e1a3](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/b86e1a38e6d20ab88cd50d2ca90b664dd21b25d4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
> Whenever you have time.

I tried now. I get this linking error:

```
2>------ Build started: Project: FreeCADGui, Configuration: Release x64 ------
2>Automatic MOC and UIC for target FreeCADGui
2>FreeCADBase.lib(FreeCADBase.dll) : error LNK2005: "public: virtual __cdecl Base::Observer<char const *>::~Observer<char const *>(void)" (??1?$Observer@PEBD@Base@@UEAA@XZ) already defined in NotificationArea.obj
2>FreeCADBase.lib(FreeCADBase.dll) : error LNK2005: "public: __cdecl Base::Observer<char const *>::Observer<char const *>(class Base::Observer<char const *> const &)" (??0?$Observer@PEBD@Base@@QEAA@AEBV01@@Z) already defined in NotificationArea.obj
2>FreeCADBase.lib(FreeCADBase.dll) : error LNK2005: "public: class Base::Observer<char const *> & __cdecl Base::Observer<char const *>::operator=(class Base::Observer<char const *> const &)" (??4?$Observer@PEBD@Base@@QEAAAEAV01@AEBV01@@Z) already defined in NotificationArea.obj
2>FreeCADBase.lib(FreeCADBase.dll) : error LNK2005: "public: __cdecl Base::Observer<char const *>::Observer<char const *>(void)" (??0?$Observer@PEBD@Base@@QEAA@XZ) already defined in NotificationArea.obj
2>   Creating library D:/FreeCAD-build/src/Gui/Release/FreeCADGui.lib and object D:/FreeCAD-build/src/Gui/Release/FreeCADGui.exp
2>D:\FreeCAD-build\bin\FreeCADGui.dll : fatal error LNK1169: one or more multiply defined symbols found
2>Done building project "FreeCADGui.vcxproj" -- FAILED.
```

I tried to recompile from scratch (deleted CMake cache etc.) but the linking error persists.

abdullahtahiriyo:
> I tried to recompile from scratch (deleted CMake cache etc.) but the linking error persists.

Thank you. I will try to make sense of that tomorrow.

donovaly:
> I will try to make sense of that tomorrow.

On MSVC caching is often a problem. So one gets linking errors because of something in the compiler cache. Therefore on linking errors, I first try to recompile a target from scratch and if nothing helps kill the CMake cache and then recompile everything from scratch (which takes about 25 min :-( )

abdullahtahiriyo:
@donovaly 

Thank you. I understand I am taking your time. Sorry for that.

I have pushed some headers that were missing, but I do not quite understand why there are multiply defined symbols here.

If you can, please try to compile it. If it does not, then I fear I have to try to get Werner's help.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781">pipeline status</a> for feature branch [PR_8781](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781). Pipeline [802528098](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/802528098) was triggered at [a8c817d](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/a8c817d70de9919c9b7690efe4ad535f5e640bd5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
Note for multiply defined symbols:
- The MSVC linker is complaining about the nested class ParameterObserver. 
- This class derives from template class ParameterGrp::ObserverType  which is ParameterGrp::Observer<_MessageType>.
- That parent is declared BaseExport.
- NotificationArea has its header guard, which is unique

I do not understand why MSVC is complaining.

As an idea, I could try to move the nested class to the cpp file as  separate class, o perhaps as part of NotificationAreaP (given that the observed parameters are ultimately stored there). I do not know if that would fix the issue or not. It would be blindly trying to fix it without understanding what is causing it.

donovaly:
> Thank you. I understand I am taking your time. Sorry for that.

Why? What we do is normal  work. You don't steal my time, I purposely spend my spare time fro FreeCAD ;-)

If you want to help me, please rebase your PR to recent master. This would save me about 10 min compilation time.

abdullahtahiriyo:
@donovaly 

Your wish, my command :)

abdullahtahiriyo:
@wwmayer 

Would you take a look to these multiply defined symbols?

abdullahtahiriyo:
@donovaly 

Thanks

wwmayer:
Maybe this is the problem: https://github.com/FreeCAD/FreeCAD/commit/7dc14634cd100fbf50e86237ea8dd42c69cc2721

donovaly:
Thanks, I will try this later today.
(@abdullahtahiriyo , as today, could you please rebase to save me some minutes compilation time?)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781">pipeline status</a> for feature branch [PR_8781](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781). Pipeline [803167501](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803167501) was triggered at [5c9747b](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/5c9747b7a1d64e0ce31e82d1496603ad9b568a40). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
It was introduced because of multiply defined symbols:
https://github.com/FreeCAD/FreeCAD/pull/7888#issuecomment-1330108694
https://github.com/FreeCAD/FreeCAD/pull/7888#issuecomment-1330740693

This Stackoverflow question has a good answer (second answer) regarding the extern use in the header file (section Solutions to the object redefinition problem):
https://stackoverflow.com/questions/2351148/explicit-template-instantiation-when-is-it-used

The developer, I believe, tried to apply solution 3 listed in that question, while saving the includer to have to define the extern version by using the macros.

This is the only instance in FreeCAD in which the "extern template" solution is applied.

What I do not understand is why this needs to be OS dependent. It would appear to me that it is best to follow a uniform approach (whatever the final decision on this should be)

To try to help debug this I have:
1. Rebased on master
2. Pushed the branch with 3 new commits:
- [59c3d50](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/59c3d50ba36b5ec4fcb0e988d624ca0d78992ada) removes the explicit template instantiation (first compilation check indicated by Werner).
-  [5b9bdf9](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/5b9bdf9062a8ce2aae6e6c43ecedc66deaf0ca36) moves the explicit instantiation to the cpp file (second compilation check indicated by Werner).
- [5c9747b](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/5c9747b7a1d64e0ce31e82d1496603ad9b568a40) the traditional solution for non OS dependent explicit instantiation in the cpp file (this is what we have applied elsewhere in FreeCAD)

@donovaly 

Would you when you have some time?

src/Gui/NotificationArea.cpp
```diff
@@ -0,0 +1,991 @@
...
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
```
wwmayer, line 22: `#include "PreCompiled.h"` is missing as the very first line. And `#include <QWidgetAction>` must be added to QtAll.h

src/Gui/NotificationArea.cpp
```diff
@@ -0,0 +1,991 @@
...
+    }
+}
+
+/************************* Notification Area *****************************************/
+
+NotificationArea::NotificationArea(QWidget* parent)
+    : QPushButton(parent)
+{
+    // QPushButton appearance
+    setText(0);
```
wwmayer, line 670: This raises an error that `QString(const char*)` is private. Should be `setText(QString())`

src/Base/Observer.cpp
```diff
@@ -0,0 +1,33 @@
...
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with FreeCAD; if not, write to the Free Software        *
+ *   Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  *
+ *   USA                                                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+
```
wwmayer, line 24: #include "PreCompiled.h" is missing as the very first line.

wwmayer:
I have tested this PR on Windows and I neither get a linking error with MSVC nor MSYS2. Also when removing the template instantiation block from Observer.h doesn't cause a failure for me.

But with RT's implementation I realized a problem:
```
#if defined (FC_OS_WIN32) || defined(FC_OS_CYGWIN)
#   ifdef FCBase
template class BaseExport Observer<const char*>;
template class BaseExport Subject<const char*>;
#   else
extern template class BaseExport Observer<const char*>;
extern template class BaseExport Subject<const char*>;
#   endif
#endif
```
For every translation unit of FreeCADBase that includes Observer.h the symbols are recreated by the compiler (confirmed with MSYS2). To avoid this the explicit template instantiation must move to Observer.cpp and the code in Observer.h should become:
```
extern template class BaseExport Observer<const char*>;
extern template class BaseExport Subject<const char*>;
``` 
This works fine with MSYS2 but now fails with MSVC when building FreeCADBase because the compiler dislikes the export macro. So, a workaround that works with MSVC and MSYS2 is
``` 
// Workaround for MSVC
#ifdef FreeCADBase_EXPORTS
#  define Base_EXPORT
#else
#  define Base_EXPORT  BaseExport
#endif

extern template class Base_EXPORT Observer<const char*>;
extern template class Base_EXPORT Subject<const char*>;
``` 
With MSYS2 I have confirmed that exclusively for Observer.cpp the symbols are defined by the compiler.

abdullahtahiriyo:
> For every translation unit of FreeCADBase that includes Observer.h the symbols are recreated by the compiler (confirmed with MSYS2)

Indeed. Thank you for the explanation and for taking the time to look into it.

I have just made all the changes and pushed the branch.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781">pipeline status</a> for feature branch [PR_8781](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8781). Pipeline [803322130](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803322130) was triggered at [da83bee](https://github.com/FreeCAD/FreeCAD/pull/8781/commits/da83bee51645aeffc19d61500ecee2bd99c64440). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
@donovaly 

Werner can build this one in MSVC and MSYS2.

The thing is that he could before the latest changes too. Could you confirm it builds ok for you?

donovaly:
> The thing is that he could before the latest changes too. Could you confirm it builds ok for you?

I will.  The next hours I have to build a new build for 0.20.2 because of #8220). So give me some time.

abdullahtahiriyo:
@donovaly 

Sure. When you can.

donovaly:
It compiles and links now for me.

wwmayer:
Something weird happened to me. After pulling master I did a make clean on my MSYS2 build and now I get a linking error. But here I have already a fix: https://github.com/FreeCAD/FreeCAD/pull/8852

src/Gui/NotificationArea.cpp
```diff
@@ -0,0 +1,993 @@
...
+
+    // If we cannot show more messages, we do no need to update the non-intrusive notification
+    if (currentlyshown < pImp->maxOpenNotifications) {
+        // There is space for at least one more notification
+        // We update the message with the most recent up to maxOpenNotifications
+
+        QString msgw =
+            QString::fromLatin1(
+                "<style>p { margin: 0 0 0 0 } td { padding: 0 15px }</style>                     \
+        <p style='white-space:nowrap'>                                                                                      \
```
wwmayer, line 885: @abdullahtahiriyo 
While looking at issue https://github.com/FreeCAD/FreeCAD/issues/8860 I got a very long error message. Because the HTML code uses nowrap the label occupied the whole width of the screen and only a part of the message was shown. Wouldn't it make more sense to use `<p style='white-space:wrap'>`?

src/Gui/NotificationArea.cpp
```diff
@@ -0,0 +1,993 @@
...
+
+    auto simplifiedstring =
+        QString::fromStdString(msg)
+            .trimmed();// remove any leading and trailing whitespace character ('\n')
+
+    // avoid processing empty strings
+    if(simplifiedstring.isEmpty())
+        return;
+
+    if (level == Base::LogStyle::TranslatedNotification) {
```
wwmayer, line 237: @abdullahtahiriyo 
If level is TranslatedNotification then the text is not translated and otherwise it will be. Shouldn't it be the other way round?

src/Gui/NotificationArea.cpp
abdullahtahiriyo: I remember there were visualisation problems when not using "nowrap".

I will investigate this further.

src/Gui/NotificationArea.cpp
wwmayer: OK, I see. Thanks!

src/Gui/NotificationArea.cpp
abdullahtahiriyo: Qt offers for white-space normal, pre, nowrap, pre-wrap.

With normal, and pre-wrap short messages become strongly multiline. Example pre-wrap:
image

Example normal:
image

With pre, the notification gets too big:
image

nowrap is the one providing the best visual results (except when a message is too long):
image
image
image

There are consequences from wrapping. The vertical space occupied by accumulated messages is controlled by the maximum number of simultaneous non-intrusive communications (defaults to 15). However, if message extend over different lines, then the vertical space can be higher.

I am still looking for a solution:
- The best looking nowrap does not have a fixed length for the qlabel, which adapts to the text. That is why fixing the size (width), which would also need to take into account the screen size, has some limitations.
- Then I thought of providing a user preference for the maximum text length, so that the user can configure it. When the maximum text length is surpassed, the notificationbox temporarily changes mode, or somehow introduces </br>... it does not look very nice though...

I keep thinking...

wwmayer: In the Python console I entered:
```
Err("""Running the Python command 'MeshPart_CreateFlatMesh' failed:
  File "/Applications/FreeCAD.app/Contents/Resources/Mod/MeshPart/MeshFlatteningCommand.py", line 58, in Activated
    boundaries = flattener.getFlatBoundaryNodes()

Unable to convert function return value to a Python type! The signature was
	(self: flatmesh.FaceUnwrapper) -> std::__1::vector<Eigen::Matrix<double, -1, 3, 0, -1, 3>, std::__1::allocator<Eigen::Matrix<double, -1, 3, 0, -1, 3> > >

Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.""")
```
and got this very wide label. In the mentioned line I simply removed the "no" from nowrap and then for this case the label text looked more useful to me:

But if there is no better alternative than nowrap for the general case then it would be good to not let the label occupy the whole width of the screen but maybe up to 30 % or so.

abdullahtahiriyo: I will check with "wrap". I did not found that option documented:
https://doc.qt.io/qt-6/richtext-html-subset.html
https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

abdullahtahiriyo: I will push a PR with:
- Fixed width notification, which is configurable from preferences.
- Normal white-space, which means that the text will wrap to fit the contents.
- Enforcing via <BR> of any \n existing in the message. This preserves the intended structure of the message (unless the message is too big for the fixed width).

Then your Error message renders like this in my screen:
image

abdullahtahiriyo: But then looking at it, I realise that it needs some escaping... will come back...

abdullahtahiriyo: Better now:
image

abdullahtahiriyo: https://github.com/FreeCAD/FreeCAD/pull/8917

PR #8783, opened 2023-03-08
Gui: fix possibly garbled toolbars and menus

wwmayer:
See also Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=986821 For now enable this line only for Windows builds

---

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -1875,7 +1875,7 @@ void Application::runApplication()
 #if QT_VERSION < QT_VERSION_CHECK(6,0,0)
         QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
-#if QT_VERSION >= QT_VERSION_CHECK(5,14,0)
+#if QT_VERSION >= QT_VERSION_CHECK(5,14,0) && defined(Q_OS_WIN)
         QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::PassThrough);
 #endif
     }


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8783">pipeline status</a> for feature branch [PR_8783](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8783). Pipeline [800050980](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/800050980) was triggered at [d32bc0d](https://github.com/FreeCAD/FreeCAD/pull/8783/commits/d32bc0dbaff972a9a96493cfbd9ebbe84fbc782e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

xtemp09:
Can I ask here a question closely related to this kind of style?

Would it be better to define global constants to simplify the code?

Like this:
```c++
#ifdef Q_OS_WIN
    bool isOSWindows = true;
    bool isOSLinux = false;
    bool isOSAndroid = false;
#endif

//<some code>
namespace Base::Qt {
    int major = 5;
    int minor = 15;
    int patch = 7;
}
```

The code you've just sent becomes:
```c++
if constexpr ( isOSWindows && (Base::Qt::major == 5) && (Base::Qt::minor > 14) )
        QGuiApplication::setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::PassThrough);
```

Therefore, [this fragment in](https://github.com/FreeCAD/FreeCAD/blob/0d77da1c816d8eba84663c1c54e85aaa39f33ab6/src/Mod/Fem/Gui/DlgSettingsFemElmerImp.cpp#L115-L135) `/src/Mod/Fem/Gui/DlgSettingsFemElmerImp.cpp`
<details><summary>fragment</summary>
<p>


```c++
#if defined(FC_OS_WIN32)
    // name ends with "_mpi.exe"
    if (!FileName.endsWith(QLatin1String("_mpi.exe"))) {
        QMessageBox::warning(this, tr("FEM Elmer: Not suitable for multithreading"),
            tr("Wrong Elmer setting: You use more than one CPU core.\n"
                "Therefore an executable with the suffix '_mpi.exe' is required."));
        ui->sb_elmer_num_cores->setValue(1);
        ui->sb_elmer_num_cores->setMaximum(1);
        return;
    }
#elif defined(FC_OS_LINUX) || defined(FC_OS_CYGWIN) || defined(FC_OS_MACOSX) || defined(FC_OS_BSD)
    // name ends with "_mpi"
    if (!FileName.endsWith(QLatin1String("_mpi"))) {
        QMessageBox::warning(this, tr("FEM Elmer: Not suitable for multithreading"),
            tr("Wrong Elmer setting: You use more than one CPU core.\n"
                "Therefore an executable with the suffix '_mpi' is required."));
        ui->sb_elmer_num_cores->setValue(1);
        ui->sb_elmer_num_cores->setMaximum(1);
        return;
    }
#endif
```

</p>
</details>




<details><summary>becomes:</summary>
<p>


```c++
const QString ending = [] (bool os) {
	if (os) return QLatin1String("_mpi.exe");
	else    return QLatin1String("_mpi"); } (isOSWindows);

if ( !FileName.endsWith(ending) ) {
	QMessageBox::warning( this, tr("FEM Elmer: Not suitable for multithreading"),
	tr("Wrong Elmer setting: You use more than one CPU core.\n"
	"Therefore an executable with the suffix '%1' is required.").arg( ending ) );
	ui->sb_elmer_num_cores->setValue(1);
	ui->sb_elmer_num_cores->setMaximum(1);
    }
```

</p>
</details>

Usage of `constexpr` also would be good.

wwmayer:
Will such code still compile with Qt6 then? Some of the Qt::AA attributes are not available in Qt6 any more and thus I don't think that `if constexpr ( Base::QInfo::Version.major < 6){` will work because constexpr does not guarantee that the code will be evaluated at compile time. It _can_ but doesn't have to.

xtemp09:
Well, it requires some digging. If FreeCAD is being ported to Qt 6, might I ask how QWinTaskbarButton is going to be replaced? It [was removed](https://doc.qt.io/qt-6.4/extras-changes-qt6.html) in Qt 6.

I can speak for all Windows users that observing progress of heavy tasks in task bar is incredibly good feature.

wwmayer:
Apparently there are plans to provide this feature for all platforms. So, it will be part of the official Qt API. But I don't know when this will happen.

Issue #8766, opened 2023-03-07
[Issue] TopoNaming Phase 1 (card 2 of 4) - create MappedName class


JohnAD:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Pop!_OS 22.04 LTS (pop:GNOME/pop)
Word size of FreeCAD: 64-bit
Version: 0.20.14555 (Git shallow) Snap 222
Build type: Debug
Branch: master
Hash: d4feb6578398d51eaf56cd794e8d3e2e000bea64
Python 3.10.6, Qt 5.15.3, Coin 4.0.0, Vtk 9.1.0, OCC 7.5.1
Locale: English/United States (en_US)
[/code]
```

### Subproject(s) affected?

None

### Issue description

> The toponaming algorithm would benefit from a type to easily store and manipulate the name serialized from the element mapping.

## description

Create MappedName.cs to store and manipulate the more-complex mapped-name used by elements (and the toponaming algorithm).

For context, see issue https://github.com/FreeCAD/FreeCAD/issues/8432

## notes

private data members are defined as follow in the branch:

```cpp
    QByteArray data; 
    QByteArray postfix; 
    bool raw;
```

where `data` could contain something like "Face2" which is either the parent (?? or the actual name if not postfix ??)
and `postfix` could contain something like ";:M2;FUS;:Hb4d:8,F" which is a "description" how we got here.

The methods of determining the contents of those fields is done in phase 2 in the ElementMap class; this class should be simpler.

I personally recommend using a more descriptive name than `data`, but those are private elements so the naming is not critical.

Most of the work already done in `src/App/MappedElement.*`

mostly, just extract this into a separate file and add unit tests to ensure that these core functions are very reliable.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chennes:
@realthunder the `MappedName` class includes some functionality to make a copy of a "slice" of the mapped name, potentially overlapping from the data to the postfix, e.g.
```
    auto temp = Data::MappedName::fromRawData(QByteArray("TESTTEST", 8));
    temp.append("ABCDEFGHIJKLM"); //postfix
    auto mappedName = Data::MappedName::fromRawData(temp, 2, 13);
```
[...]

PR #8785, opened 2023-03-08
Toponaming: Added MappedName class

Pesc0:

---

I'm giving a shot at #8766 

I've basically isolated the MappedName class into its own files and renamed a couple of variables to make the code more understandable, but apart from that I haven't changed anything.

I feel like this class is needlessly long and complicated for what it does, so before writing tests I'm asking if/how it could be simplified.

Left to do:
- simplification and formatting
- write tests
- uncomment some parts that reference IndexedName (will have to wait until it gets merged)

### DIFF
--- a/src/App/CMakeLists.txt
+++ b/src/App/CMakeLists.txt
@@ -262,6 +262,7 @@ SET(FreeCADApp_CPP_SRCS
     ComplexGeoDataPyImp.cpp
     Enumeration.cpp
     IndexedName.cpp
+    MappedName.cpp
     Material.cpp
     MaterialPyImp.cpp
     Metadata.cpp
@@ -279,6 +280,7 @@ SET(FreeCADApp_HPP_SRCS
     ComplexGeoData.h
     Enumeration.h
     IndexedName.h
+    MappedName.h
     Material.h
     Metadata.h
 )
--- a/src/App/MappedName.cpp
+++ b/src/App/MappedName.cpp
@@ -33,13 +33,12 @@
 using namespace Data;
 
 
-void MappedName::compact() const
+void MappedName::compact()
 {
-    auto self = const_cast<MappedName*>(this);
 
     if (this->raw) {
-        self->data = QByteArray(self->data.constData(), self->data.size());
-        self->raw = false;
+        this->data = QByteArray(this->data.constData(), this->data.size());
+        this->raw = false;
     }
 
 #if 0
--- a/src/App/MappedName.cpp
+++ b/src/App/MappedName.cpp
@@ -1,0 +1,54 @@
+/****************************************************************************
+ *   Copyright (c) 2020 Zheng, Lei (realthunder) <realthunder.dev@gmail.com>*
+ *                                                                          *
+ *   This file is part of the FreeCAD CAx development system.               *
+ *                                                                          *
+ *   This library is free software; you can redistribute it and/or          *
+ *   modify it under the terms of the GNU Library General Public            *
+ *   License as published by the Free Software Foundation; either           *
+ *   version 2 of the License, or (at your option) any later version.       *
+ *                                                                          *
+ *   This library  is distributed in the hope that it will be useful,       *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of         *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the          *
+ *   GNU Library General Public License for more details.                   *
+ *                                                                          *
+ *   You should have received a copy of the GNU Library General Public      *
+ *   License along with this library; see the file COPYING.LIB. If not,     *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,          *
+ *   Suite 330, Boston, MA  02111-1307, USA                                 *
+ *                                                                          *
+ ****************************************************************************/
+
+#include "PreCompiled.h"
+
+#ifndef _PreComp_
+# include <unordered_set>
+#endif
+
+//#include <boost/functional/hash.hpp>
+
+#include "MappedName.h"
+
+using namespace Data;
+
+
+void MappedName::compact() const
+{
+    auto self = const_cast<MappedName*>(this);
+
+    if (this->raw) {
+        self->data = QByteArray(self->data.constData(), self->data.size());
+        self->raw = false;
+    }
+
+#if 0
+    static std::unordered_set<QByteArray, ByteArrayHasher> PostfixSet;
+    if (this->postfix.size()) {
+        auto res = PostfixSet.insert(this->postfix);
+        if (!res.second)
+            self->postfix = *res.first;
+    }
+#endif
+}
+
--- a/src/App/MappedName.h
+++ b/src/App/MappedName.h
@@ -44,6 +44,12 @@ namespace Data
 class AppExport MappedName
 {
 public:
+    /// Create a MappedName from a C string, optionally prefixed by an element map prefix, which
+    /// will be omitted from the stored MappedName.
+    ///
+    /// \param name The new name. A deep copy is made.
+    /// \param size Optional, the length of the name string. If not provided, the string must be
+    /// null-terminated.
     explicit MappedName(const char* name, int size = -1)
         : raw(false)
     {
@@ -57,6 +63,10 @@ public:
         data = size < 0 ? QByteArray(name) : QByteArray(name, size);
     }
 
+    /// Create a MappedName from a C++ std::string, optionally prefixed by an element map prefix,
+    /// which will be omitted from the stored MappedName.
+    ///
+    /// \param name The new name. A deep copy is made.
     explicit MappedName(const std::string& nameString)
         : raw(false)
     {
@@ -69,6 +79,9 @@ public:
         data = QByteArray(name, static_cast<int>(size));
     }
 
+    /// Create a MappedName from an IndexedName. If non-zero, the numerical part of the IndexedName
+    /// is appended as text to the MappedName. In that case the memory is *not* shared between the
+    /// original IndexedName and the MappedName.
     explicit MappedName(const IndexedName& element)
         : data(element.getType()),
           raw(false)
@@ -87,6 +100,11 @@ public:
     // FIXME if you pass a raw MappedName into these constructors they will
     // reset raw to false and things will break. is this intended?
 
+    /// Copy constructor with start position offset and optional size. The data is *not* reused.
+    ///
+    /// \param other The MappedName to copy
+    /// \param startPosition an integer offset to start the copy from
+    /// \param size the number of bytes to copy. If not specified
     MappedName(const MappedName& other, int startPosition, int size = -1)
         : raw(false)
     {
--- a/src/App/MappedName.h
+++ b/src/App/MappedName.h
@@ -89,11 +89,12 @@ public:
     /// is appended as text to the MappedName. In that case the memory is *not* shared between the
     /// original IndexedName and the MappedName.
     explicit MappedName(const IndexedName& element)
-        : data(element.getType()),
-          raw(false)
+        : data(QByteArray::fromRawData(element.getType(), qstrlen(element.getType()))),
+          raw(true)
     {
         if (element.getIndex() > 0) {
-            data += QByteArray::number(element.getIndex());
+            this->data += QByteArray::number(element.getIndex());
+            this->raw = false;
         }
     }
 
@@ -664,8 +665,13 @@ public:
     /// continuing through postfix. No bounds checking is performed when compiled in release mode.
     char operator[](int index) const
     {
-        // FIXME overflow underflow checks?
+        if (index < 0) {
+            index = 0;
+        }
         if (index >= this->data.size()) {
+            if (index - this->data.size() > this->postfix.size() - 1) {
+                index = this->postfix.size() - 1;
+            }
             return this->postfix[index - this->data.size()];
         }
         return this->data[index];
@@ -708,7 +714,7 @@ public:
     }
 
     /// Ensure that this data is unshared, making a copy if necessary.
-    void compact() const;
+    void compact();
 
     /// Boolean conversion is the inverse of empty(), returning true if there is data in either the
     /// data or postfix, and false if there is nothing in either.
--- a/src/App/MappedName.h
+++ b/src/App/MappedName.h
@@ -80,6 +80,9 @@ public:
 		: data(other.data), postfix(other.postfix), raw(other.raw)
 	{}
 
+    //FIXME if you pass a raw MappedName into these constructors they will 
+    //reset raw to false and things will break. is this intended?
+
 	MappedName(const MappedName & other, int startpos, int size = -1)
         : raw(false)
     {
@@ -276,6 +279,7 @@ public:
 
     void append(const char * d, int size = -1)
     {
+        //FIXME raw not assigned?
         if (d && size) {
             if (size < 0)
                 size = qstrlen(d);
@@ -288,26 +292,41 @@ public:
 
     void append(const MappedName & other, int startpos = 0, int size = -1)
     {
+        // enforce 0 <= startpos <= other.size
         if (startpos < 0)
             startpos = 0;
         else if (startpos > other.size())
             return;
-        if (size < 0 || size + startpos > other.size())
+
+        // enforce 0 <= size <= other.size - startpos
+        if (size < 0 || size > other.size() - startpos)
             size = other.size() - startpos;
 
-        int count = size;
-        if (startpos < other.data.size()) {
-            if (count > other.data.size() - startpos)
+       
+        if (startpos < other.data.size()) // if starting inside data
+        { 
+            int count = size;
+            //make sure count doesn't exceed data size and end up in postfix
+            if (count > other.data.size() - startpos) 
                 count = other.data.size() - startpos;
+                
+            //if this is empty append in data else append in postfix
             if (startpos == 0 && count == other.data.size() && this->empty()) {
                 this->data = other.data;
                 this->raw = other.raw;
             } else
                 append(other.data.constData() + startpos, count);
+
+            //setup startpos and count to contiune appending the remainder to postfix
             startpos = 0;
             size -= count;
-        } else
+        } 
+        else //else starting inside postfix
+        {
             startpos -= other.data.size();
+        }
+
+        //if there is still data to be added to postfix
         if (size) {
             if (startpos == 0 && size == other.postfix.size()) {
                 if (this->empty())
--- a/src/App/MappedName.h
+++ b/src/App/MappedName.h
@@ -103,9 +103,6 @@ public:
 
     MappedName(const MappedName& other) = default;
 
-    // FIXME if you pass a raw MappedName into these constructors they will
-    // reset raw to false and things will break. is this intended?
-
     /// Copy constructor with start position offset and optional size. The data is *not* reused.
     ///
     /// \param other The MappedName to copy
@@ -364,7 +361,6 @@ public:
     /// \param size The number of bytes to copy. If omitted, dataToAppend must be null-terminated.
     void append(const char* dataToAppend, int size = -1)
     {
-        // FIXME raw not assigned?
         if (dataToAppend && (size != 0)) {
             if (size < 0) {
                 size = static_cast<int>(qstrlen(dataToAppend));
@@ -690,6 +686,8 @@ public:
     }
 
     /// Returns true if this is shared data, or false if a unique copy has been made.
+    /// It is safe to access data only if it has been copied prior. To force a copy 
+    /// please \see compact()
     bool isRaw() const
     {
         return this->raw;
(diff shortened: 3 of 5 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8785">pipeline status</a> for feature branch [PR_8785](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8785). Pipeline [800120267](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/800120267) was triggered at [719cc8e](https://github.com/FreeCAD/FreeCAD/pull/8785/commits/719cc8e3236e744bb13031a98184f56672358445). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Pesc0:
Thanks for helping out, being a newbie this advice is much appreciated. 
I've started writing tests, but soon realized I can't build anymore. I'm on another computer and I'm running fedora 37, which from a quick google seems to be the cause of the problem.

```
/usr/include/c++/12/bits/stl_vector.h:1206: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::front() [with _Tp = cmGccStyleDependency; _Alloc = std::allocator<cmGccStyleDependency>; reference = cmGccStyleDependency&]: Assertion '!this->empty()' failed.
gmake[2]: *** [src/Mod/PartDesign/Gui/CMakeFiles/PartDesignGui.dir/build.make:1451: src/Mod/PartDesign/Gui/CMakeFiles/PartDesignGui.dir/depend] Aborted (core dumped)
gmake[1]: *** [CMakeFiles/Makefile2:4830: src/Mod/PartDesign/Gui/CMakeFiles/PartDesignGui.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
```

Trying to debug this at the moment...

Pesc0:
Update: a clean build fixed the issue. No clue what it was

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8785">pipeline status</a> for feature branch [PR_8785](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8785). Pipeline [802950491](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/802950491) was triggered at [07989dd](https://github.com/FreeCAD/FreeCAD/pull/8785/commits/07989ddf8792bc4de750ec77f314a0e120806f3e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
It looks to me like all of those Qt version checks can be removed: FreeCAD only supports Qt 5.12 and up now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8785">pipeline status</a> for feature branch [PR_8785](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8785). Pipeline [803311455](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803311455) was triggered at [19a01a5](https://github.com/FreeCAD/FreeCAD/pull/8785/commits/19a01a5c37468cd7bf01cbeabc22d5c5e151e425). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Pesc0:
I'm almost done adding tests. I've run into issues with two test cases, called `fromRawDataCopyStartposAndSize` and `rfind`. Could you please have a look at them, to make sure I've not made mistakes?

As i said in the beginning to me it looks like this class performs a simple function in a complicated way. So then i have a few questions (perhaps i should ping realthunder?):
- Whats the point in having two separate variables `QBytesArray data, postfix` when great effort is spent in almost all functions to treat them as a single concatenate one? I think a great simplification could be made by changing the class to have a single `QBytesArray name;` and then an index `int postfixStartIdx;` to signal the postfix starting position.
- Whats the role of the `raw` flag? As i understand only `data` can be raw. But then I'm confused by the `compact()` function which just discards raw and sets it to false? Some constructors also just set raw to false.
- Is a `QBytesArray` really needed? Could it be replaced by `std::vector<uint8_t>` or some other structure to drop the Qt dependency? 

Finally, while @sliptonic suggestion of writing tests first is really useful when it comes to improving the implementation of the class, it provides little insight when it comes to improving its interface. How is this class used? Are all these overloads really needed? Can parts of the interface be discarded to make the class simpler?

Or perhaps this is premature optimization and adding tests is enough for now?

chennes:
Thank you (so very much!) for taking this on -- I really appreciate it. If you haven't already, I suggest giving @realthunder's TNP writeup a read (at least the first couple of sections), it helps make sense of what these classes are being used for: https://github.com/realthunder/FreeCAD_assembly3/wiki/Topological-Naming

I should have time to look at your tests tonight and see if I can help figure out why you're getting those strange results.

Pesc0:
I have skimmed through it but I'm definitely going to give it a proper read when I have a bit of time. Thank you for your help!

chennes:
For your purposes, the first section explains why he's added both the data and the postfix, and goes through the trouble of maintaining both. My overall feeling is that the current interface is too general, and makes it too easy to shoot yourself in the foot. I pinged RT in the issue to see if he has a use case for the sort of slicing you are testing in that failing test -- I stepped through the code in a debugger for a couple of hours last night and couldn't isolate what the exact failure is. It seems like the QByteArray is at one point making a data copy that I find surprising, and then later on not making a copy when I would have expected it to. I ran out of steam at that point, but will revisit it tonight.

Pesc0:
Ok so after a bit of digging and reading I think I have some answers:

From the Qt documentation of [fromRawData](https://doc.qt.io/qt-6/qbytearray.html#fromRawData):
> The bytes are _not_ copied. The [QByteArray](https://doc.qt.io/qt-6/qbytearray.html) will contain the data pointer.

> A subsequent attempt to modify the contents of the returned [QByteArray](https://doc.qt.io/qt-6/qbytearray.html) or any copy made from it will cause it to create a deep copy of the data array before doing the modification. This ensures that the raw data array itself will never be modified by [QByteArray](https://doc.qt.io/qt-6/qbytearray.html).

> Warning: A byte array created with fromRawData() is not '\0'-terminated, unless the raw data contains a '\0' byte at position size. While that does not matter for [QDataStream](https://doc.qt.io/qt-6/qdatastream.html) or functions like [indexOf](https://doc.qt.io/qt-6/qbytearray.html#indexOf)(), passing the byte array to a function accepting a const char * expected to be '\0'-terminated will fail.

So:
- Probably this is the reason for the unexpected/missing copy you have found
- The role of `raw` is to keep track of the `\0` termination. Not to signal that `\0` might be in the string (as I initially thought). The tests should be updated then.
- QByteArray is used to avoid copying data, for performance reasons.
- Having two separate variables for data and postfix is probably for the same reason (avoid copying data, or rather copy only whats needed and leave alone the other variable).

Perhaps smart pointers could come in handy to better express ownership?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8785">pipeline status</a> for feature branch [PR_8785](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8785). Pipeline [805124785](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/805124785) was triggered at [98068b6](https://github.com/FreeCAD/FreeCAD/pull/8785/commits/98068b6a0d1d256d9f64c885ceee966ebe04af31). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
I just merged #8776 so you'll need to manually merge in the CMake changes, the you can uncomment the IndexedName stuff in your PR

chennes:
@Pesc0 if you can rebase to current master I have a PR for your branch that runs a clang-tidy pass on the code: https://github.com/chennes/FreeCAD/commit/f6628a726820865e0ccd3edf4d862a26bc13ce13

Pesc0:
Hopefully I did that right. Sorry for the absence, university is taking most of my time. I'll try to complete work on this in the next few days, but since I dont want to hold back development feel free to take over at any time.

chennes:
I've just submitted a PR to your repo with two commits: one is a clang-tidy cleanup of the code, and the second starts adding Doxygen comments. I was in the middle of adding those comments, so I can continue doing so if you want to focus on the code itself, or we can reverse that, or whatever. School takes priority, of course!! 😁 

https://github.com/Pesc0/FreeCAD/pull/1

Pesc0:
Thank you! Since I've never done this, should I merge with merge commit or with a rebase? In this case it doesn't matter right? 
It might be a good idea to split work on comments / code, both because I dont have much time, and because english is not my language, so your comments will be more understandable to others, which is the important part

chennes:
I suggesting using a "Rebase and merge", which will give us less cleanup work to do later. I will continue working on adding comments, and will send you another PR when I've finished them. Let me know if there are functions you are planning on changing the interface to, or deleting, and I won't comment those 😄 .

Pesc0:
Done. 
> Let me know if there are functions you are planning on changing the interface to, or deleting

I can't really comment on that right now, I need to study a bit how the class is actually used in code. In general I'm thinking of not touching the interface at all, as it may be better to do that at a later time. I am pretty sure some functions are not even used though, and those could as well be deleted. Still, I need to check first.

As for the code: so it is fine if I substitute many of the checks, for example
```
        if (startpos < 0)
            startpos = 0;
```
with asserts? I havent coded any tests that verify that those checks are implemented. 
In general I am free to change things as long as the test passes right?

chennes:
Any idea why `operator==` doesn't just delegate to the `data` and `postfix` members, but instead jumps through hoops that may include constructing a whole new `QByteArray`?

chennes:
Regarding replacing the bounds checks with asserts: in a technical sense that's fine, of course, but I'm not sure there's any reason to do it -- it's probably premature optimization at this point. It may be that @realthunder had benchmarks indicating that the additional comparisons added an unacceptable time increment in `IndexedName` but didn't have the same for `MappedName`. You could add a timed test to find out, if you want (don't forget to run in release mode, though!).

Pesc0:
Because, as you can see in the test `comparisonOperators`:
```
Data::MappedName mappedName1 = Data::MappedName(Data::MappedName("TEST"), "POSTFIXTEST");
Data::MappedName mappedName3 = Data::MappedName(Data::MappedName("TESTPOST"), "FIXTEST");

EXPECT_EQ(mappedName1 == mappedName3, true);
```

It treats `data` and `postfix` as a single concatenate variable. It would not be the same as 
```
this->data == other.data && this->postfix == other.postfix;
```

chennes:
Thanks for the clarification. The code feels disjointed, but we're pretty close to the metal here, I'm not sure that's really avoidable.

Regarding `toConstString` -- as far as I can see this is only used once in the code, and to me it feels like it would be better represented as `operator[]`. Calling it `toConstString` makes it seem like it will behave like `toString`, when in fact it's totally different. What do you think?

chennes:
Another one:
```
MappedName copy() const
    {
        if (!this->raw) {
            return *this;
        }
        MappedName res;
        res.data.append(this->data.constData(), this->data.size());
        res.postfix = this->postfix;
        return res;
    }
```
As far as I can tell this is the only piece of code in the entire class that actually does anything with the value of `raw`, and it's not clear to me why. If `raw` is false, then we create an implicitly shared copy of both `data` and `postfix`. If it is true, then we only share `postfix`, but create a new copy of `data`. Why does "raw" require that `data` be copied instead of shared?

chennes:
In your comments you have:
```
void append(const char* dataToAppend, int size = -1)
    {
        // FIXME raw not assigned?
```
I believe that not assigning `raw` in this function is correct. If `this->data` is currently empty, `raw` will already be false, and the data is copied via `QByteArray`'s `append(char *)` function. If `this->data` is NOT empty, then we don't know (or care) if it is raw, because we only append to `this->postfix`.

Pesc0:
> Regarding toConstString -- as far as I can see this is only used once in the code

If it's a single instance I think its smarter to delete it completely and rework the code at the call site. One less thing to worry about. It does not behave like `toString` and neither as `operator[]`

> Why does "raw" require that data be copied instead of shared?

The role of raw is to keep track of the \0 termination. If raw is true then the array is not `\0` terminated, so special care is needed when handling the variable, in particular the size of it, as it can't be determined in a conventional string-like way of looking for the termination char. Thats why you see `this->data.size()`

> I believe that not assigning raw in this function is correct.

I think you're right on this.

--- 

I think you're getting what I was trying to say some time ago: I feel this class is needlessly complicated for what it does. Many things are convoluted and hard to understand, which should not be the case for a simple storage structure. True, it needs to be fast, but I think there's room for great improvement. I think we have to make a choice right now

- Keep the interface as is, reorganize the code a bit to make it more readable (in the limits of whats possible), and perhaps come back later in the optimization phase to refactor and improve things.

- Attempt a rework of the interface and class itself: This will require:
  - Understanding well how the class is used in code
  - Many changes to future code that requires this class (which may not be a bad thing, if a cleaner interface means cleaner future code as well)

What I think can be improved:
- Functions that aren't absolutely necessary can be straight up deleted. 
- Function implementations should be in the cpp file, not in the header.
- I would love to see `raw` disappear if its not absolutely needed. I feel it adds unnecessary complexity. Reason it is present now: to keep track of the `\0` termination, because QByteArrays created with FromRawData remove the `\0`. Why is FromRawData used? To avoid copying data.
- On the topic of data copies, I think a choice needs to be made: either this class holds (owns) the data, OR it holds a pointer to data stored somewhere else. The current behavior of copying data _sometimes_ is just confusing, and bug-prone (see `fromRawDataCopyStartposAndSize`)
- I dont think having two distinct variables is a good idea when they are treated in all cases as a single concatenate string. As I already proposed, a single variable with an index to split the `data` and `postfix` segments would be much better.
- I dont like the Qt dependency: right now it is used for this avoid-copying behavior, but if the logic of the class is going to be reworked i would like to see some stdlib storage class used instead. For example `std::vector<uint8_t>` or `std::string`.
- To avoid copies (after deciding who owns the data) smart pointers could be used.

Thoughts on this?

chennes:
The interface is the important thing right now: let's get that right and worry about the implementation details in another PR. I'll look more closely at how this is getting used in the toponaming class.

chennes:
OK, I've got a good first draft of all the Doxygen comments in place now: https://github.com/Pesc0/FreeCAD/pull/2

realthunder:
> Any idea why `operator==` doesn't just delegate to the `data` and `postfix` members, but instead jumps through hoops that may include constructing a whole new `QByteArray`?

The use of `data` and `postfix` in MappedName` is considered as an implementation details. The reason of the separation is to improve memory efficiency. The same postfix may be attached to many, and sometimes all, resulting subshapes of a boolean operation. To most of outside caller, the `MappedName` is supposed to be treated as a single string. So, two `MappedName's are considered the same if their `data+postfix` are the same. One case where two `MappedName` can have different postfix is when the name is converted to a normal string and stored inside a property link, and then read back and converted to `MappedName` to query topological data, where the `postfix` information is lost.


> I think actually raw is used in the QByteArray sense: it purely indicates whether a copy has been made. The Compact function is used to ensure a copy, which is why it then sets raw false, even though it will happily copy null characters.

Yes, that is correct. In other word, `raw` means the data is not owned, and is not safe to store in the element map.

chennes:
How are you feeling about the tests you've got now @Pesc0 ? Are we ready to mark this as 'ready for review' so we can get it merged? I know there's more work to be done, but I don't want to hold up the other two cards working on implementation details here, which can happen behind-the-scenes once the tests are solid.

Pesc0:
I've made a draft of the changes I proposed yesterday on another [branch](https://github.com/Pesc0/FreeCAD/blob/mappedNameExperimental/src/App/MappedName.h). The interface and implementation has been simplified drastically. I'm trying to separate out the logic required to avoid copies: My idea is to use another class `MappedNameRef`, and copies will need to be explicitly made with `copy(MappedNameRef other)`. I have yet to study how the class is used by actual code, so I haven't thought how to layout the logic handling copies.
Perhaps all these changes are not useful, but hey it doesn't hurt to try. If you have any thoughts on this I would be glad to hear them.

Meanwhile, thanks to the insight from realthunder, I've fixed the broken tests. Next I'm going to refine what we're currently working on. Surely `raw` and `compact()` are going to need a rename.

Just a couple of tests are missing: the hash function and the functions that make use of `indexedName` are not tested. Since you worked on IndexedName it would be awesome if you could take care of those. The rest of the test suite is going to need a proper review to make sure I haven't made mistakes. Other than that I think the current state looks good.

JohnAD:
@Pesc0 and @chennes ,

I really like how this is coming out! 

Pesc0 you had asked about the two options, the first one being:

> Keep the interface as is, reorganize the code a bit to make it more readable (in the limits of whats possible), and
> perhaps come back later in the optimization phase to refactor and improve things.

IMO, I agree with you and suggest this path. I find it ironic that it might be in the optimization phase that we come back and see if we can make it simpler and less optimized :smile: .

Setting up good performance tests will be easier when we have the `ElementMap` class fully flushed out. That one does much of the name mangling IIRC.

I'll take a deeper dive into your code and tests tonight and tomorrow; and see if I can help any.

chennes:
Thanks @JohnAD -- I'll get the tests for the NamedIndex functions written tonight (hopefully).

chennes:
@Pesc0 I just sent you a PR with those new tests.

Pesc0:
Thank you John and thank you Chris!

Pesc0:
@chennes can I revert the two blocks you deleted? Those check that the moved object is properly reset. It helped me find some issues while I was messing in the other branch so they're not useless.

chennes:
You can't legally access an object once it has been moved: it's Undefined Behavior.

Pesc0:
> You can't legally access an object once it has been moved: it's Undefined Behavior.

Oh! 

I added the remaining tests, so I can mark this as ready for review. 
One of your tests was failing, could you please check if the fix I made to `MappedName(const IndexedName& element)` is correct? To have the data shared QByteArray needs to be initialized with `fromRawData`, so I did that.

chennes:
I think this is looking good -  I'll wait for @JohnAD to give it another look an then let's get it merged! Nice teamwork here, thanks for stepping to the plate @Pesc0. I know you mentioned you are a university student: just a heads-up that FreeCAD is a Google Summer of Code sponsoring organization, so if you want to continue working on this or any other FreeCAD project, you should consider applying for some money :).

adrianinsaval:
It's so very nice to see actual progress on this and multiple people involved. Thank you all for your work

Pesc0:
Oh cool, I didn't know you could get paid for this, thanks for letting me know! I gave that a quick look, it looks like it requires a significant time commitment, and from what I understand follows a pretty strict schedule. Did I get that right? With many things going on in life I'm not sure I would be able to follow a regular schedule and so on, since I'm doing this for fun anyway. I'm going to properly research the topic in the next days.
Regardless of how that goes I would love to keep helping here, since this has been fun and you all have been very welcoming!

JohnAD:
Things look good!

Only one quick item jumped out: I don't see a test for `const char* constPostfix() const`. But given that just returns a pointer, there is not much to test. I guess simply that it does not return a nullptr when there is a legit constData.

JohnAD:
I have also been making some notes for the future. Given that this class is very narrowly used, there might be some refactorings later on.

For example, the behavior of `append` where it *sometimes* modifies `postfix` based on whether the `data` is empty is ... IMO ... oddly confusing. It works, but I wonder if a pair of more-specific renamed routines might make for cleaner/clearer code. 

But I'll refrain from suggesting anything on these notes at least until we are done with the mapping algo in Phase 2.

JohnAD:
BTW, I agree you @Pesc0 that almost all of this code should be in the .cpp file. That way things like the `appendToBuffer` function can be made strictly private. But, lets do that in "optimization" phase later.

chennes:
@Pesc0 I'm going to merge this as-is so that the remaining two cards in this series have access to this class. If you want to submit another PR with more tests or additional polish let me know, or I can take care of writing the remaining test. Thanks @JohnAD for your insights here, I think there is definitely some future work on this class, but this an excellent start.

Regarding Google Summer of Code - yes, it's a significant commitment (typically a couple of months) -- it's intended to be an "internship"-like experience where you have an opportunity to really dive deeply into an open source project. Of course, there are other opportunities besides GSoC, so please feel free to continue working on code, participating in Issues discussions, etc. Thanks again for jumping right in on this card!

Pesc0:
@chennes  Alright, I'll come back to refine this class when more pieces are in place and I have a bit more context. Besides the missing test, I don't think there's much more to do with the tests (for now). The missing test can be added for completeness but it's pretty useless IMO: `postfixBytes()` gets checked in almost every test, `constPostfix()` is just the const char* version of that.

Issue #8708, opened 2023-03-01
[Problem] Misleading conflict shortcuts error message


Fat-Zer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.20 (Release)

### Full version info

```shell
[code]
OS: Gentoo Linux (Trinity/trinity)
Word size of FreeCAD: 64-bit
Version: 0.21.0.32083 (Git)
Build type: Debug
Branch: master
Hash: 85ad88d101fb2f5eccf62e096f65a79f348d5f65
Python 3.11.1, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.7.0
Locale: Russian/Russia (ru_RU)
Installed mods: 
  * slic3r-tools
  * BIM 2021.12.0
  * ExtremeProDark 2.7.0
  * Glass (Disabled)
[/code]
```

### Subproject(s) affected?

None

### Problem description

When some workbench has conflicting shortcuts an error message is shown that says:

```
The key sequence 'W, A' is ambiguous. Use 'Configure Keyboard Shortcuts'
from the 'Settings' menu to solve the ambiguity.
No action will be triggered.
```

But there is neither a menu called 'Settings' nor a 'Configure Keyboard Shortcuts' entry. At least I couldn't found such. There is a tab in `Tools->Customize->Keyboard`, but (1) it suggests that one shortcut should be prioritized over another in case of ambiguity and (2) the field 'Current shortcut' as well as 'Assign' button are appear to be disabled, so I see no way to change a shortcut.

So to recap the issues are:

1. A misleading (probably obsolete) error message
2. (Probably) Incorrect conflicting shortcut handling
3. <s>GUI doesn't provide an ability to modify a shortcut</s>

----
 
![img8](https://user-images.githubusercontent.com/1060057/222139673-cf7c4d31-ed5f-47b6-9886-6a26ba01f355.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

----------

## Update

The bug is a bit more specific to surroundings when I thought initially, so I'll add some more concrete instructions to reproduce it. 

### How to reproduce

1. Install some kde libraries, namely `kxmlgui` and its designer plugin `plugins/designer/kxmlgui5widgets.so`. In Debian they will be respectively in `libkf5xmlgui5` and `libkf5xmlgui5-dev`packages.

      `apt-get install libkf5xmlgui5 libkf5xmlgui5-dev`

[...]

PR #8810, opened 2023-03-10
Gui: Prevent UiLoader from loading 3rd-party Qt plugins.

Fat-Zer:
This PR workarounds problems in QUiLoader/kde and fixes https://github.com/FreeCAD/FreeCAD/issues/8708. See the bug for details.

Due to a flaw in the QUiLoader, UiLoader were loading all designer plugins it can find in QApplication::libraryPaths(). This in general a bad practice and leads to bugs due to some plugins may perform some unexpected actions upon load which may interfere with FreeCAD's functionality.

To avoid such problems reset the libraryPaths before creation of a UiLoader object and restore afterwards.

Also:
- move setLanguageChangeEnabled(true) into constructor due to it's called every time it's being instanced anyway.
- Add a small tooltip on how to modify a shortcut.

### DIFF
--- a/src/Gui/DlgKeyboard.ui
+++ b/src/Gui/DlgKeyboard.ui
@@ -103,6 +103,9 @@
          </item>
          <item row="1" column="2" colspan="2">
           <widget class="Gui::AccelLineEdit" name="accelLineEditShortcut">
+           <property name="toolTip">
+            <string>To change a current shortcut enter the new shortcut in the field below and press 'Assign'.</string>
+           </property>
            <property name="readOnly">
             <bool>true</bool>
            </property>
--- a/src/Gui/UiLoader.cpp
+++ b/src/Gui/UiLoader.cpp
@@ -24,6 +24,7 @@
 #ifndef _PreComp_
 # include <QAction>
 # include <QActionGroup>
+# include <QCoreApplication>
 # include <QDir>
 # include <QFile>
 # include <QLayout>
@@ -488,10 +489,20 @@ QString QUiLoader::errorString() const
 UiLoader::UiLoader(QObject* parent)
   : QUiLoader(parent)
 {
-    // do not use the plugins for additional widgets as we don't need them and
-    // the application may crash under Linux (tested on Ubuntu 7.04 & 7.10).
-    clearPluginPaths();
     this->cw = availableWidgets();
+    setLanguageChangeEnabled(true);
+}
+
+std::unique_ptr<UiLoader> UiLoader::newInstance(QObject *parent)
+{
+    QCoreApplication *app=QCoreApplication::instance();
+    QStringList libPaths= app->libraryPaths();
+
+    app->setLibraryPaths(QStringList{}); //< backup library paths, so QUiLoader won't load plugins by default
+    std::unique_ptr<UiLoader> rv{new UiLoader{parent}};
+    app->setLibraryPaths(libPaths);
+
+    return rv;
 }
 
 UiLoader::~UiLoader()
@@ -544,8 +555,8 @@ void UiLoaderPy::init_type()
 }
 
 UiLoaderPy::UiLoaderPy()
+    : loader{UiLoader::newInstance()}
 {
-    loader.setLanguageChangeEnabled(true);
 }
 
 UiLoaderPy::~UiLoaderPy()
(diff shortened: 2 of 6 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8810">pipeline status</a> for feature branch [PR_8810](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8810). Pipeline [801995728](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/801995728) was triggered at [773c51e](https://github.com/FreeCAD/FreeCAD/pull/8810/commits/773c51e8f2fc21dd0eb3edc57baa01cb6ef95af8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/DlgKeyboard.ui
```diff
@@ -103,6 +103,9 @@
          </item>
          <item row="1" column="2" colspan="2">
           <widget class="Gui::AccelLineEdit" name="accelLineEditShortcut">
+           <property name="toolTip">
+            <string>To change a current shortcut enter the new shortcut in the field bellow and press 'Assign'.</string>
```
luzpaz, line 107: s/bellow/below/

src/Gui/UiLoader.cpp
```diff
@@ -488,10 +489,20 @@ QString QUiLoader::errorString() const
...
     this->cw = availableWidgets();
+    setLanguageChangeEnabled(true);
+}
+
+std::unique_ptr<UiLoader> UiLoader::newInstance(QObject *parent)
+{
+    QCoreApplication *app=QCoreApplication::instance();
+    QStringList libPaths= app->libraryPaths();
+
+    app->setLibraryPaths(QStringList{}); //< backup library paths, so QUiLoader won't plugins by default
```
luzpaz, line 501: won't **load** plugins by default

adrianinsaval:
doesn't this mean that shortcuts defined in kde won't work on freecad? I'm not sure if this is what we want

src/Gui/DlgKeyboard.ui
Fat-Zer: fixed, thanks

src/Gui/UiLoader.cpp
Fat-Zer: fixed, thanks

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8810">pipeline status</a> for feature branch [PR_8810](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8810). Pipeline [803062064](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803062064) was triggered at [4cf9fd1](https://github.com/FreeCAD/FreeCAD/pull/8810/commits/4cf9fd1bd6264b826848a21d364a6b3cea22d563). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Fat-Zer:
> doesn't this mean that shortcuts defined in kde won't work on freecad? I'm not sure if this is what we want

Nope, it doesn't... Assuming you are talking about common action shortcuts like "Open file", they are handled by a different plugin type (`platformthemes` AFAIR) this PR prevents `designer` plugins  being loaded (which provides 3rd-party widgets for dynamic ui-forms). Those plugins were disabled anyway due to [`clearPluginPaths()`][1] call, but the shared libraries were loaded into the process.

PS: The bug is a bit more specific when I initially assumed, so I've added some steps on how to reproduce to the initial report... not so much to convince to try it, but to show off that it requires pretty specific circumstances to be triggered... i.e. kde dev-libs installed on a non-kde system, switch to a python-driven workbench, concurrent shortcuts combinations (provided out of the box on the (external) BIM workbench)
PPS: sorry for a bit late answer, I thought I've posted this comment yesterday.


[1]: https://github.com/FreeCAD/FreeCAD/blob/master/src/Gui/UiLoader.cpp#L493

PR #8811, opened 2023-03-10
[PD] add tooltip for Hole feature

donovaly:
- tooltip makes clear what the depth means for countersinks
- add comment why a calculation cannot be performed
- shorten some long lines

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHole.cpp
+++ b/src/Mod/PartDesign/App/FeatureHole.cpp
@@ -1654,7 +1654,6 @@ App::DocumentObjectExecReturn* Hole::execute()
 {
      TopoDS_Shape profileshape;
     try {
-
         profileshape = getVerifiedFace();
     }
     catch (const Base::Exception& e) {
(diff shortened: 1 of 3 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8811">pipeline status</a> for feature branch [PR_8811](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8811). Pipeline [802013499](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/802013499) was triggered at [43df324](https://github.com/FreeCAD/FreeCAD/pull/8811/commits/43df3244cb8b390c1549ccb93d7ac3847bbbcd53). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/PartDesign/App/FeatureHole.cpp
```diff
@@ -1740,23 +1742,29 @@ App::DocumentObjectExecReturn* Hole::execute()
...
                 countersinkAngle = Base::toRadians(90.0);
             }
 
-            if ( isCountersink) {
-                holeCutDepth = 0;
+            if (isCountersink) {
+                holeCutDepth = 0.0;
+                // We cannot recalculate the HoleCutDiameter because the previous HoleCutDepth
+                // is unknown. Therefore we cannot know with what HoleCutDepth the current
+                // HoleCutDiameter was calculated.
```
adrianinsaval, line 1755: shouldn't we also reset the holeCutDepth property then?

donovaly: You mean we should block to set the depth? This would indeed be sensible but let me think about this further

PR #8819, opened 2023-03-10
Do not allow an empty identifier in `getIdentifier`

galou:

---

An empty string is not a valid identifier, so `Base::Tools::getIdentifier` should not return an empty string. This commit fixes this, `getIdentifer` returns `"_"` when the input is empty.

The Python API apparently doesn't use the function `getIdentifier` because `App.ActiveDocument.Box.addProperty('App::PropertyFloat', '', 'Test', 'help')` still raises a `NameError` exception.


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8819">pipeline status</a> for feature branch [PR_8819](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8819). Pipeline [802449577](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/802449577) was triggered at [de42289](https://github.com/FreeCAD/FreeCAD/pull/8819/commits/de42289c12dc401608d97924959a6b6486a01966). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
why `"_"`? isn't raising an error the correct behavior for this? What's the usecase you want to fix?

galou:
I don't think `getIdentifier` is supposed to raise an error. "_" is use as replacement in the same function. I don't have any use-case, I just saw this and I considered it a flaw so I corrected it.

chennes:
In almost all calls to `getIdentifier` the input is explicitly checked a priori, before calling the function, and in no cases that I could find is there any sort of exception handling that could take a reasonable action if this instead threw an exception. So I don't think this change is going to do any harm, whereas throwing an exception could maybe possibly do ugly things. I don't have a strong opinion about whether to make the change or not, so I'm inclined to just merge it as a bit of "just in case" input sanitization. Any objections?

chennes (approved):

PR #8826, opened 2023-03-10
Consistency of task button positions

ronnystandtke:
I noticed that most task dialogs have their buttons north. I also noticed that the angular measure task had its buttons south despite the linear measure task was already changed to have them north. This is now fixed.

I also removed a redundant implementation of the close button in the display properties task (View -> Appearance...). This way it now also has its close button north per default.

The "Edit -> Placement" task still has the buttons at the south position but I am not sure if it is by design because of the additional reset button...

### DIFF
--- a/src/Gui/DlgDisplayProperties.ui
+++ b/src/Gui/DlgDisplayProperties.ui
@@ -434,13 +434,6 @@
      </layout>
     </widget>
    </item>
-   <item row="3" column="0">
-    <widget class="QDialogButtonBox" name="buttonBox">
-     <property name="standardButtons">
-      <set>QDialogButtonBox::Close</set>
-     </property>
-    </widget>
-   </item>
   </layout>
  </widget>
  <layoutdefault spacing="6" margin="11"/>
@@ -528,22 +521,6 @@
     <hint type="destinationlabel">
      <x>98</x>
      <y>404</y>
-    </hint>
-   </hints>
-  </connection>
-  <connection>
-   <sender>buttonBox</sender>
-   <signal>rejected()</signal>
-   <receiver>Gui::Dialog::DlgDisplayProperties</receiver>
-   <slot>reject()</slot>
-   <hints>
-    <hint type="sourcelabel">
-     <x>150</x>
-     <y>461</y>
-    </hint>
-    <hint type="destinationlabel">
-     <x>144</x>
-     <y>243</y>
     </hint>
    </hints>
   </connection>
--- a/src/Gui/DlgDisplayPropertiesImp.cpp
+++ b/src/Gui/DlgDisplayPropertiesImp.cpp
@@ -192,11 +192,6 @@ void DlgDisplayPropertiesImp::setupConnections()
     connect(d->ui.buttonColorPlot, &ColorButton::clicked, this, &DlgDisplayPropertiesImp::onButtonColorPlotClicked);
 }
 
-void DlgDisplayPropertiesImp::showDefaultButtons(bool ok)
-{
-    d->ui.buttonBox->setVisible(ok);
-}
-
 void DlgDisplayPropertiesImp::changeEvent(QEvent *e)
 {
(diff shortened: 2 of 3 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8826">pipeline status</a> for feature branch [PR_8826](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8826). Pipeline [802726945](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/802726945) was triggered at [7580878](https://github.com/FreeCAD/FreeCAD/pull/8826/commits/75808785c909acebd8a0283668ce683673f0d266). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
Thank you, the display properties dialog already has the button on the north side in 0.21, but I think it's also good to remove unnecessary code. I had not noticed the angular dimension discrepancy, which is odd as I had searched the entire code for the "South" keyword before. 

If you have the time, could you also look into moving the buttons for the clipping and persistent section cut dialogs?
image

also the edit->placement dialog:
image

donovaly:
> @ronnystandtke can you provide screenshot(s) of the change you're proposing ?

Yes, this would be good since I cannot see the difference at a first look.

luzpaz:
Does this change tab clicking order (using tab button to navigate the dialog box)?

ronnystandtke:
> @ronnystandtke can you provide screenshot(s) of the change you're proposing ?

OK, here is a screenshot comparing the situation before and after this pull request:
FreeCAD_Angular_Measure_Task_Button_Position_Patch

> Does this change tab clicking order (using tab button to navigate the dialog box)?

I noticed that before the patch it was impossible to move the focus to the close button by pressing the tab button. After applying the patch it was possible.


Regarding the other questions: I will take a look at other parts of FreeCAD in the future, but please don't mix up too many different things in this pull request. I would be glad if we could first merge this PR and continue the discussion in later PRs.

src/Gui/DlgDisplayProperties.ui
```diff
@@ -531,21 +524,5 @@
     </hint>
    </hints>
   </connection>
-  <connection>
-   <sender>buttonBox</sender>
-   <signal>rejected()</signal>
```
donovaly, line 536: Why did you remove this signal?

donovaly:
> If you have the time, could you also look into moving the buttons for the clipping and persistent section cut dialogs?

This has nothing to do with the PR as far as I can see. Since I am the author of the section cutting dialog, please CC me in a new forum thread to discuss what could be improved.

donovaly:
> OK, here is a screenshot comparing the situation before and after this pull request: 

Thanks, now I get what is meant by North and South (strange wording since depending on where I look towards, south is up, left, right or down. Thus North/south on a PC screen is misleading). I know it is a technical term you are just use it.

The PR is fine with me except of the removal of the rejected() signal (which I don't understand yet).

src/Gui/DlgDisplayProperties.ui
ronnystandtke: I removed the whole buttonBox widget. It is redundant because the same thing is already implemented in DlgDisplayPropertiesImp.cpp with:
```
QDialogButtonBox::StandardButtons TaskDisplayProperties::getStandardButtons() const
{
    return QDialogButtonBox::Close;
}
```

donovaly:
@chennes could you please have a look at this PR? Looks OK for me but maybe I miss something.

chennes:
> Regarding the other questions: I will take a look at other parts of FreeCAD in the future, but please don't mix up too many different things in this pull request. I would be glad if we could first merge this PR and continue the discussion in later PRs.

Agreed! Let's keep these PRs tight and focused (as this one is). Everything looks great to me, thanks for the PR.

PR #8838, opened 2023-03-11
Sketcher: Solver Message Widget simplification.

PaddleStroke:
@abdullahtahiriyo Solver Message as we discussed.

---

### DIFF
--- a/src/Mod/Sketcher/Gui/TaskSketcherMessages.h
+++ b/src/Mod/Sketcher/Gui/TaskSketcherMessages.h
@@ -39,20 +39,19 @@ namespace SketcherGui {
 
 class ViewProviderSketch;
 
-class TaskSketcherMessages : public Gui::TaskView::TaskBox
+class TaskSketcherMessages : public QWidget
 {
     Q_OBJECT
 
 public:
-    explicit TaskSketcherMessages(ViewProviderSketch *sketchView);
+    explicit TaskSketcherMessages(ViewProviderSketch* sketchView);
     ~TaskSketcherMessages() override;
 
-    void slotSetUp(const QString &state, const QString &msg, const QString& link, const QString& linkText);
+    void slotSetUp(const QString& state, const QString& msg, const QString& link, const QString& linkText);
 
 private Q_SLOTS:
     void on_labelConstrainStatusLink_linkClicked(const QString &);
-    void on_autoUpdate_stateChanged(int state);
-    void on_autoRemoveRedundants_stateChanged(int state);
+    void onAutoUpdateStateChanged();
     void on_manualUpdate_clicked(bool checked);
     
 protected:
@@ -60,7 +59,6 @@ protected:
     Connection connectionSetUp;
 
 private:
-    QWidget* proxy;
     std::unique_ptr<Ui_TaskSketcherMessages> ui;
 };
 
--- a/src/Mod/Sketcher/Gui/TaskDlgEditSketch.cpp
+++ b/src/Mod/Sketcher/Gui/TaskDlgEditSketch.cpp
@@ -56,8 +56,6 @@ TaskDlgEditSketch::TaskDlgEditSketch(ViewProviderSketch *sketchView)
     Content.push_back(Constraints);
     Content.push_back(Elements);
 
-    if (!hGrp->GetBool("ExpandedMessagesWidget",true))
-        Messages->hideGroupBox();
     if (!hGrp->GetBool("ExpandedSolverAdvancedWidget",false))
         SolverAdvanced->hideGroupBox();
     if (!hGrp->GetBool("ExpandedConstraintsWidget",true))
@@ -96,7 +94,6 @@ bool TaskDlgEditSketch::accept()
 bool TaskDlgEditSketch::reject()
 {
     ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Mod/Sketcher");
-    hGrp->SetBool("ExpandedMessagesWidget",Messages->isGroupVisible());
     hGrp->SetBool("ExpandedSolverAdvancedWidget",SolverAdvanced->isGroupVisible());
     hGrp->SetBool("ExpandedConstraintsWidget",Constraints->isGroupVisible());
     hGrp->SetBool("ExpandedElementsWidget",Elements->isGroupVisible());
--- a/src/Mod/Sketcher/Gui/TaskSketcherMessages.cpp
+++ b/src/Mod/Sketcher/Gui/TaskSketcherMessages.cpp
@@ -21,6 +21,9 @@
  ***************************************************************************/
 
 #include "PreCompiled.h"
+#ifndef _PreComp_
+# include <QAction>
+#endif
 
 #include <Gui/Application.h>
 #include <Gui/BitmapFactory.h>
@@ -36,29 +39,16 @@ using namespace SketcherGui;
 using namespace Gui::TaskView;
 namespace bp = boost::placeholders;
 
-TaskSketcherMessages::TaskSketcherMessages(ViewProviderSketch *sketchView) :
-    TaskBox(Gui::BitmapFactory().pixmap("document-new"), tr("Solver messages"), true, nullptr),
+TaskSketcherMessages::TaskSketcherMessages(ViewProviderSketch* sketchView) :
+    QWidget(),
     sketchView(sketchView),
     ui(new Ui_TaskSketcherMessages)
 {
-    // we need a separate container widget to add all controls to
-    proxy = new QWidget(this);
-    ui->setupUi(proxy);
-    QMetaObject::connectSlotsByName(this);
-
-    this->groupLayout()->addWidget(proxy);
+    ui->setupUi(this);
 
     connectionSetUp = sketchView->signalSetUp.connect(boost::bind(&SketcherGui::TaskSketcherMessages::slotSetUp, this, bp::_1, bp::_2, bp::_3, bp::_4));
 
     ui->labelConstrainStatus->setOpenExternalLinks(false);
-
-    ui->autoUpdate->onRestore();
-    ui->autoRemoveRedundants->onRestore();
-
-    if(ui->autoUpdate->isChecked())
-        sketchView->getSketchObject()->noRecomputes=false;
-    else
-        sketchView->getSketchObject()->noRecomputes=true;
 
     // Set up the possible state values for the status label
     ui->labelConstrainStatus->setParameterGroup("User parameter:BaseApp/Preferences/Mod/Sketcher/General");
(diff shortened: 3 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8838">pipeline status</a> for feature branch [PR_8838](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8838). Pipeline [803196059](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803196059) was triggered at [ecfa7f2](https://github.com/FreeCAD/FreeCAD/pull/8838/commits/ecfa7f2c3b8d2f2ae47ee2a427746afcc20deec9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/TaskSketcherMessages.h
```diff
@@ -39,28 +39,26 @@ namespace SketcherGui {
...
 class ViewProviderSketch;
 
-class TaskSketcherMessages : public Gui::TaskView::TaskBox
+class TaskSketcherMessages : public QWidget
 {
     Q_OBJECT
 
 public:
-    explicit TaskSketcherMessages(ViewProviderSketch *sketchView);
+    explicit TaskSketcherMessages(ViewProviderSketch* sketchView, QWidget* parent = nullptr);
```
abdullahtahiriyo, line 47: I am unsure why have you done this.

At first I thought you had realised that now TaskSketcherMessages is a QWidget. You know that Widgets are generally added to the parent in QT's memory model and then the parent is responsible for deleting them. So you added this field.

However, in the cpp file you are instantiating the object with only one parameter (without parent). So it has no effect.

Then, I start to think maybe you are doing it for a different reason. But I cannot find one, as passing a nullptr as parent is the same as passing nothing at all (a QWidget's default constructor has parent defaulted to nullptr).

In any case:

Observation 1: If you did this to use the QT's memory management, the constructor is taking a default argument nullptr. If the parent is not passed, the object is created, but who is responsible for the deletion? Nobody.

Observation 2: A TaskBox is a widget, but a TaskBox does not pass the parent as parameter. If this is working fine, the TaskBox must have a different memory  model (which may not follow the QT's memory model). If we look into that, the widgets of a TaskDialog are added to the Contents array of the TaskDialog class. If you look into the destructor of TaskDialog, you will see that the pointers inside Contents are destroyed there. So TaskDialog has a different memory model in place.

Observation 3: Because the TaskSketcherMessages object is added to the Contents array, you potentially have two memory models running in parallel (the one of QT via a parent being different from nullptr, and the one provided by TaskDialog). 

I tend to think that adding this parameter here is highly misleading for future developers, as they are invited to think the object may be used in a way that is not intended.

Is there any reason I have not considered for which this should stay?

PaddleStroke: For this I reused the same as I used in previous PR. Now why I did that in the first place last october I'm not sure. I think it's because I copied the SketcherGeneralWidget as a base for this solver widget, and mixed it with the things that taskBox does not pass the parent. So it's an innaccuracy, I will remove it.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8838">pipeline status</a> for feature branch [PR_8838](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8838). Pipeline [803287292](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803287292) was triggered at [6967035](https://github.com/FreeCAD/FreeCAD/pull/8838/commits/6967035c2a844b2bec6ebcdebb4c743fa0e8fa82). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/TaskSketcherMessages.h
abdullahtahiriyo: Thanks for fixing it.

Try to always read the code before PR-ing it and questioning what looks odd. Aside from possibly fixing all mistakes, it will help you review code of others. Eventually you will need to do that too.

donovaly:
This breaks the dialog for me visibly:
FreeCAD_k2K4ACR9hr

The green script is hardly visible on our default background color. Can you please have a look?

donovaly:
What is the intention of the PR? Opening the UI file in Qt's designer I get a totally scrambles UI:
designer_fDOdffDYPb

I need to make a new Windows build for master but think this PR needs top be fixed before.
So am i allowed to revert the PR, to make then the weekly build and you meanwhile have a look and improve the PR?

abdullahtahiriyo:
@donovaly 

Do you have a style sheet?

Without style sheet it does not look ugly here:
image

But with style sheets it does look even uglier than yours:
image

If the weekly cannot be made on the previous PR, then please revert the PR.

abdullahtahiriyo:
@PaddleStroke 

Could you take a look what may be causing this?

donovaly:
I have no style sheet. And also when I reset my preferences I get the same ugly result.

donovaly:
> If the weekly cannot be made on the previous PR, then please revert the PR.

The weekly can of course be built (technically) but all Windows users will get an ugly dialog and as it seems also everybody with a style sheet.

PaddleStroke:
Indeed it seems to look weird depending on stylesheet. If I'm not mistaken,
Uwe has windows without style sheet.

Yes it can be reverted I will improve it on Monday.

abdullahtahiriyo:
> The weekly can of course be built (technically) but all Windows users will get an ugly dialog and as it seems also everybody with a style sheet.

I miswrote and the message I wanted to convey did not get through.

If Paddle is not going to work on it until Monday, it only makes sense to revert it.

donovaly:
> Indeed it seems to look weird depending on stylesheet. If I'm not mistaken, Uwe has windows without style sheet. Yes it can be reverted I will improve it on Monday.

Thanks for having a look. I can also text the next PR in advance before it is merged.

abdullahtahiriyo:
@donovaly 

Thanks for reverting. 

Yes, I would appreciate if you look into this one when it is fixed and before merge.

donovaly:
> I miswrote and the message I wanted to convey did not get through.

I found meanwhile out that I can locally reset my master to a certain commit. I did so and built my weekly from this.

I reverted now nevertheless since usually the Conda weekly builds are prepared during the weekend.

adrianinsaval:
If I'm not mistaken that ugly gradien is added by us so IMO rather than not having this PR we should look into removing that useless gradient.

adrianinsaval:
the stuff here looks suspicious as there are mentions of gradient and it's used in the task panel
https://github.com/FreeCAD/FreeCAD/tree/master/src/Gui/QSint/actionpanel
https://github.com/FreeCAD/FreeCAD/blob/57aac275c77ecfd0295bc6ab78f77e57fbbbcda5/src/Gui/TaskView/TaskView.h#L115

donovaly:
> If I'm not mistaken that ugly gradien is added by us so IMO rather than not having this PR we should look into removing that useless gradient.

The bug is not the gradient or background color but that there is no colored frame anymore like for the other sub-panels (the user is free to choose any background color he likes). The second bug is that the refresh icon is sometimes hidden as in Abdullah's screenshot.
I also get a scrambled UI file in Designer and Designer also complains about missing settings. So all in all the UI file is somehow mangled.

adrianinsaval:
the point of the PR is to not have a frame anymore, all sane stylesheets provide a plain color for the task panel where text could be readable, it's only due to that colored gradient background that it doesn't work by default, the gradient should be removed.

adrianinsaval:
Googling a little on the subject:
https://marketplace.qt.io/products/qsint

> QSint::Actions module contains additional Qt widgets which maintain look-and-feel of Windows XP Task Panels, Windows Vista Actions etc.

we're using something that is meant to emulate the style of OS versions we don't even support anymore, no wonder it looks outdated and out of place even on windows.

adrianinsaval:
I created an issue for the task panel fiasco, if you have responses to this subject please direct them to the issue: #8845

PaddleStroke:
After thinking about it, I think a frame would be good. Because if RT transparancy feature get merged, then it will look very bad without some kind of frame.
So I'm thinking that I'm going to make a frame that will look like a 'minimal taskbox' so that style is consistent.

donovaly:
> I created an issue for the task panel fiasco, if you have responses to this subject please direct them to the issue: #8845

There is no "fiasco". Please see the UI file. This is mangled and that's the problem. For example I can take the PR as it is but use the UI file from master and it looks at least OK that I can see all options and the button properly.

You can open the general debate about the default colors but it is already running for the background color.
Once the background color is changed we can change the other colors. In the discussion I was convinced that we should not change any color before the background because it determines the other colors.
In every case, please don't mix in a general issue in a particular PR.

adrianinsaval:
I don't see anything scrambled in your screenshot, what you see is what was intended, text directly on the panel and a toolbutton besides it, no frame is intended. So the only problem that makes this not look OK there is the gradient background.

donovaly:
> I don't see anything scrambled in your screenshot

Please open the UI file with Qt's designer and see by yourself. (Also save the UI file and look at the diff) Also lee Abdullah's second screenshot and mine from Designer.

adrianinsaval:
The only problem I see in your Qt designer screenshot is that the toolbutton can be larger than it should be allowed, but this is not a problem during actual use as can be seen in your first screenshot it has a reasonable size, abdullah's second screenshot is a bug in FreeCAD bundled stylesheets that can be observed in any other toolbutton in the task panel wether it is inside a frame or not regardless of what the UI file has, it is due to a workaround that is explicitly applied to those stylesheets. I'll search and link the problematic line in the stylesheets later. So again, look at the screenshot which you first brought up saying it isn't right, you are seeing exactly what this PR intended, the only problem being the gradient.

adrianinsaval:
this is the customization that makes toolbuttons look ugly when using a stylesheet
https://github.com/FreeCAD/FreeCAD/blob/57aac275c77ecfd0295bc6ab78f77e57fbbbcda5/src/Gui/Stylesheets/Darker-blue.qss#L1585-L1598

might need this hack applied but I'm not sure how we can limit that to only the sketcher toolbuttons as apparently toolbuttons are also used in other places without a menu (I wonder why those aren't pushbuttons?) and look odd with the 20px padding only on the right
https://github.com/FreeCAD/FreeCAD/blob/57aac275c77ecfd0295bc6ab78f77e57fbbbcda5/src/Gui/Stylesheets/Darker-blue.qss#L1967-L1971

PaddleStroke:
@adrianinsaval Thanks for finding that. I was starting to wonder where this button size bug could come from.

PaddleStroke:
So I have been trying things this morning. One nice thing is that I found
taskBox can be set without header.
So we can keep using taskBox for the solver message, and just disable the
header.
That will help keeping the style consistent and provide a frame.

The only problem I have is that the height of the frame is a bit too big,
so it's a bit ugly. I have been trying to play with the size hints, margins
and more but can't find what is building up the height so far. To be
continued...

PaddleStroke:
So, in my many test I can't find how to shrink this height. So here how it looks with and without header : 
image

After looking at it for some time, I think that the version with header looks better. It takes a little bit more vertical space but the headerless box looks odd (even if it had a normal height).

What do you guys think ?

adrianinsaval:
the header is not useful here as I don't think you would want to hide the message and this kind of prevents having the message fixed always at the top, but this task dialog has been cleaned enough that this is not important anymore, even with the header this will likely fit entirely in the task panel without a scrollbar for me so I'm happy with either.

donovaly:
> the header is not useful here as I don't think you would want to hide the message and this kind of prevents having the message fixed always at the top

One might not want to collapse this subdialog but the header itself is useful. Don't forget new users. Therefore the header text should stay to make clear what this is about.

PR #8855, opened 2023-03-12
[NotificationArea] Gui: Notifications - Convenience functions to send intrusive / non-intrusive notifications

abdullahtahiriyo:
Functions to transparently send errors, warnings, notifications and translatednotifications, either as intrusive messages (modal pop-up QMessageBox style) or as non-intrusive messages (notificationarea), depending on the preferences of the user.

It is intended to substitute calls to QMessageBox with these functions in WBs, including the Sketcher WB.

Example:

QMessageBox::warning(Gui::getMainWindow(), QObject::tr("Wrong selection"),
                      QObject::tr("Cannot add a constraint between two external geometries."));

Can be rewritten as:

 Gui::TranslatedNotification(obj,
               QObject::tr("Wrong selection"),
               QObject::tr("Cannot add a constraint between two external geometries."));

or

 Gui::TranslatedNotification("Sketcher",
               QObject::tr("Wrong selection"),
               QObject::tr("Cannot add a constraint between two external geometries."));

where the first parameter (obj or "Sketcher") is the notifier string. If obj is an App::DocumentObject (or a derived class of it), internally the function transparently calls getFullLabel() to get the notifier string.

### DIFF
--- a/src/Gui/CMakeLists.txt
+++ b/src/Gui/CMakeLists.txt
@@ -1171,6 +1171,7 @@ SET(FreeCADGui_SRCS
     Macro.h
     MergeDocuments.h
     MetaTypes.h
+    Notifications.h
     PreCompiled.cpp
     PreCompiled.h
     QtAll.h
--- a/src/Gui/Notifications.h
+++ b/src/Gui/Notifications.h
@@ -1,0 +1,219 @@
+/***************************************************************************
+ *   Copyright (c) 2023 Abdullah Tahiri <abdullah.tahiri.yo@gmail.com>     *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This library is free software; you can redistribute it and/or         *
+ *   modify it under the terms of the GNU Library General Public           *
+ *   License as published by the Free Software Foundation; either          *
+ *   version 2 of the License, or (at your option) any later version.      *
+ *                                                                         *
+ *   This library  is distributed in the hope that it will be useful,      *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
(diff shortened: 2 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8855">pipeline status</a> for feature branch [PR_8855](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8855). Pipeline [803700881](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803700881) was triggered at [dfc2ac1](https://github.com/FreeCAD/FreeCAD/pull/8855/commits/dfc2ac170ec298c5ef99e007bc23baced60c224f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
which is more advisable to use, Notification or TranslatedNotification?

abdullahtahiriyo:
For Gui (which is the namespace of this file), at least generally, the translated notification is better. From a translation framework point of view, with a TranslatedNotification the translation itself is made at the source of the notification. This means that the developer has full control on the size of the context used for translation. Also, there are many translations that are already there (with whichever context the developers decided). Changing it to "Notifications" may not be advisable (I am not knowledgeable enough of Crowding, but changing a context perhaps makes troubles for translators).

Notifications, where translation is advisable (IMO always), share the context all over FreeCAD ("Notifications"). This means the context will be as big as all the possible notifications. However, this may cover a use case where it may not be acceptable to add a dependency on the translation framework, while marking for translation is perfectly acceptable. This is much more important though for the App namespace than for Gui.

Although the question is about notifications, the notification area listens to warnings and errors (as the report view does). These are not translated. If these are marked for translation with the "Notifications" context, the notification area will translate them. So, we can have the regular untranslated messages in the Report View, while having translated versions in the Notification Area.

I wanted it to be fully legacy compatible while providing a new flexible framework. So there are different possibilities. We will see how users and developers end up using it (I am often surprised by how features end up being used).

abdullahtahiriyo:
@donovaly 

When you can. If you can. It is rebased.

wwmayer:
Compilation of this PR won't fail because the header is not included by any source file

abdullahtahiriyo:
@wwmayer 

Yes. I had gone for a walk and I was thinking exactly that. Thanks.

PR #8849, opened 2023-03-12
fix #8847 sketcher dialog tool buttons bad styling

adrianinsaval:
Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [x]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---
Fixes #8847 
the buttons and filter menu now look like this:
image
instead of:
image

PR #8864, opened 2023-03-12
Sketcher: Solver message taskbox simplification

PaddleStroke:
Rework of Solver Message simplification regarding @donovaly remarks on the visibility issue on color gradient.

Here I kept the taskbox, only simplified the content.

Note: The button too small with style sheet is solved by @adrianinsaval in https://github.com/FreeCAD/FreeCAD/pull/8849

---


PaddleStroke:
Here is what it looks like : 
image
@abdullahtahiriyo @adrianinsaval @donovaly

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8864">pipeline status</a> for feature branch [PR_8864](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8864). Pipeline [803804991](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/803804991) was triggered at [02f0ddc](https://github.com/FreeCAD/FreeCAD/pull/8864/commits/02f0ddcfb41f14f17221a4835d1371cd5209edf8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
many thanks!

At first I have a general note: Can you in future please create your PRs from master?
I see there many compiler warnings that are already fixed in master since some days and by starting from latest master testers save a lot of compilation time (Compiling your PR took more than 20 minutes)
Before you continue to work on the PR, can you please rebase?

To the PR: the dialog looks now correct:
FreeCAD_uxtthkJ4cv

But the update button is not working. I can click on it and nothing happens:
FreeCAD_csv7lGElCM
So Auto-Update is always on, no matter if and how often I click the button.

PaddleStroke:
@donovaly I just rebased. 
Are you sure about the issue you have ? I doubled checked before and after rebasing and don't have these issues on windows.
Besides I haven't changed the manualUpdate button connection, are you sure it's not workign?
For the auto-update checkbox it works great on my side. Have you tried clearing preferences?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8864">pipeline status</a> for feature branch [PR_8864](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8864). Pipeline [804139014](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/804139014) was triggered at [858317d](https://github.com/FreeCAD/FreeCAD/pull/8864/commits/858317dc50401a61b21db59ac011887d8e6ea23a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
image

abdullahtahiriyo:
@donovaly 

Do you really want to keep the "solver messages" header?

I can go along with it and without it with no problems. The question is more in line with the aim of this PR to reduce the panel height. I think it is quite evident from any possible content of the panel that what is provided is a solver message.

Then I remember well we still have an issue with the Freiheitsgrad string and DoFs for after this PR, where I plan to work on a proper tooltip.

PaddleStroke:
Personally I think we should keep the header too since we can't keep it directly in the taskview because of the background color.

The headerless taskbox looks weird among the other taskboxes.

Another solution would be to put the solver message next to the new notification button. Then it doesn't need a taskbox.

abdullahtahiriyo:
@donovaly 

I wait for your input.

donovaly:
> Do you really want to keep the "solver messages" header?

Yes, because for newbies it is important to have a self-explaining GUI. The header tells you what the messages are about. This helps to immediately understand.
The panel does not have to be collapsible. However, then the whole dialog would be inconsistent (people might wonder when the first panel behaves differently). Since I never noticed a complaint, i would keep the collapsible feature for all panels of the dialog.

All this is of course only my personal opinion.

donovaly:
> @donovaly I just rebased. Are you sure about the issue you have ?

I provided a screencast.

I tested you PR now again (now just 35 seconds compilation time :-) )

You maybe misunderstand my bug report:
- open a sketch
- click on the triangle besides the update button in the sketch dialog

result: Auto-update is on but the user has no feedback that the update is on - the button does not indicate it is already pressed

- now click on the update button (not the triangle)

result: the button does no indicate a change. Click at the triangle besides the update button and you see that the Auto-update is still on.

So the issues are:
1. button state is not indicated in general
2. pressing the button does not change the state shown when pressing the triangle
3. since there is only one option now, a triangle besides the button is not necessary at all. It is only confusing. We should remove it
4. The tooltip of the button tells about recomputing the whole document. This is confusing for new users. The tooltip should be improved by adding more info (tooltips can have line breaks)

PaddleStroke:
@donovaly Ok I see now there is a misunderstanding : Clicking the button, and clicking the menu entry do not have the same effect.

Clicking the button is equivalent as current master button 'Update'.
Toggling the menu entry 'Auto update' is same as current master checkboxe 'auto update'.

So the button do not have 2 states. Every time you click it, it recompute (what the tooltip says).

donovaly:
> So the button do not have 2 states. Every time you click it, it recompute (what the tooltip says).

Hmm, but why do  you hide the option auto-update? This is a major feature and personally, I change it a lot when designing because on complex documents you need to turn it off temporarily.

You try to save the height of a single text line by hiding a major option people need to change often. At least for my personal workflow this is inconvenient.

Regarding the tooltip, it is technically correct, but for new users not enough. I will think about it and propose an extended one.

PaddleStroke:
Well it was discussed extensively and people seemed to agree on this ui decision. It is not so much hidden and it's in a logical position.

Besides what is the point of enabling auto-update? I mean you say on complex models you need to turn it off, which I understand. But when do you actually need to turn it on? What is the point of turning it on?

abdullahtahiriyo:
> Well it was discussed extensively and people seemed to agree on this ui decision. It is not so much hidden and it's in a logical position.
> 
> Besides what is the point of enabling auto-update? I mean you say on complex models you need to turn it off, which I understand. But when do you actually need to turn it on? What is the point of turning it on?

Welcome to community driven OSS development.

Uwe has it generally on. I have it 99% of the time off. To me (and I was the one putting the checkbox in the solver message taskbox in the first place), it is immaterial to have it as a drop-down on the update button. However, for Uwe it is important. I do see his point on not being able to know the state when it is under a toolbutton option. But then we can have the reduction in height people were asking. What about leaving it as it is now, but changing the icon of the button depending on the state of the auto-update feature? That might look odd to someone....

To me it is unimportant to have the header at all. In fact, I think that according to people's feedback they wanted to get rid of it and have the minimal amount of text. Uwe prefers to have the header and have it collapsible for UI coherence. I have problems to understand collapsed solver messages.

Hopefully we can agree that with the taskbox (any version), there is no visual artifact and at least from that point of view the PR is OK.

Gentlemen, we cannot have all at the same time. Either we find an agreement, or I propose to move this to the user forum for a short user input discussion to clarify the matter.

donovaly:
@abdullahtahiriyo , I think you mixed up some things. Let me clarify:

1. I want to keep the header like for new users because especially new users discover the Sketcher WB as first thing. An intuitive UI is very helpful.
2. I don't have a strong opinion about the collapsible feature. You can make it so that it can't be collapsed, but what is benefit? The height stays the same, only some users might wonder why one cannot collapse this while the other panels. Therefore I think, we can keep this as it is.
3. I have the auto-update feature on about 50% of the times. For example in my current project, I have about 30 different sketches. Some are simple (rectangles etc.) some are complex meander structures of circuit paths (several hundred constraints)). So depending of what sketch I am working on and the recompilation time, I toggle the update feature.

It is hard to please every user. This PR removes the height of 2 options. So even when we keep the auto-update option, we still save the height of one option.

donovaly:
> Gentlemen, we cannot have all at the same time.

But we can have a compromise. You know, compromises are something in the middle that does not please anybody by 100 % :-)
So what about this:
- we remove the redundants option but keep the auto-update option

PaddleStroke:
I still don't understand why you toggle it on. What is the use ? When you are editing a sketch, its feature is automatically hidden. So you don't even see the solid updating as you sketch. So I can't see any usecase where you want this enabled.

By compromising you mean having things your way. 
Having this auto-update feature in plain sight is poor UI. Having this setting available at all is already a bad UI decision. Just 1 day ago someone asked me for help on youtube, he couldn't sketch without heavy lag. Guess what his problem was? Auto-update enabled. (see comment of jocelyn linot in this video: https://www.youtube.com/watch?v=5HF8f648JlU)

So saying this should stay in plain sight as a compromise and/or as a good ui decision is wrong. This is just getting your way.

I advocate to remove it completely. So compromising is putting it in this menu.

Syres916:
All I'll say is read https://forum.freecad.org/viewtopic.php?style=3&p=563447#p563447 this is just one of many discussions in the past and there has never been a consensus to the best of my knowledge.

donovaly:
> I still don't understand why you toggle it on.

As I wrote, I design electrical circuits on 3D-printed parts. The sketched of the part are usually poly-lines thus have e.g. about 15 to 25 constraints. With them I want to see the change immediately -> turn on auto-update.

The sketches that define the later galvanized electrical paths are complex meander structures with up to 1000 constraints. Recalculation would often take minutes -> I turn it off.
So I change the option often. (Our layouts are usually relatively complex in general (custom medical engineering where 20 micron difference can matter.))

However, we cannot know what people do with FC. This is just my use case. If a majority says we should remove the option/put it under the button as in your PR, I have to accept it since we cannot please everybody.

abdullahtahiriyo:
I have dedicated quite some time since yesterday's exchange to reflect.

The first reflection is that it is suboptimal to arrive to this stage after a long discussion and reopen the discussion of something that seemed closed. 

The second reflection is that Uwe is right in noting that with the new arrangement it is not evident whether the feature is on or off. 

I am rather sure that users in the discussion wanted to save vertical space. I am unsure if users in the discussion realised Uwe's point.

New today is whether there is or there is not a use case. I think there is one. I do not come to it as often as Uwe. However I have been there. I oppose removing the setting altogether.

The last point relates to how often we change the setting. I seldom change the setting. Uwe does it often.

A last consideration is: how much more vertical space do we need to save? Having a large monitor, I lack perspective.

I need to make a decision. I will review this PR and, if it is ok, I will merge it. Then, I propose that we let some time for users to start using it. We will also get a grasp of what we have done. Then, we can restart the discussion if necessary.

abdullahtahiriyo:
@PaddleStroke 

I have merged Adrián's PR and rebased yours on top:
image

PaddleStroke:
Found the problem. manualUpdate already had some stylesheet entries before adrianisaval added his. So they are conflicting I guess.
What is your stylesheet on your screenshot by the way?

adrianinsaval:
I think I missed a comma on the qss files, but I don't quite get how it gets like that, I'll investigate later today

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8864">pipeline status</a> for feature branch [PR_8864](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8864). Pipeline [805819717](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/805819717) was triggered at [d5b294e](https://github.com/FreeCAD/FreeCAD/pull/8864/commits/d5b294e95592b17fc285f3ffbce7430f74ae28b8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
Ah yes looks like it. Missing comma after `QSint--ActionGroup QToolButton#filterButton`.
Do you want me to include the fix in my PR?

donovaly:
> I need to make a decision

Yes please. As I often wrote, in the end someone has to decide. We cannot debate forever and not everybody could be pleased.
I put in my perspective to the discussion and that is enough.

I did not want to reopen a discussion I even missed. I stumbled over it because did not understand the new dialog logic.

abdullahtahiriyo:
Thank you paddle.

PaddleStroke:
Thanks to you guys for reviewing!

Issue #8800, opened 2023-03-09
[Issue] 'Unit system' setting is hard to find in preferences


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
Last
```

### Subproject(s) affected?

Core

### Issue description

Currently to change the unit system, one has to go to preferences, then to the 'unit' tab, which is at the end of the list. So for new users it's hard to find it. 
It is also annoying to change manually for users that need to change between documents.

Suggestion : 
Find a more prominent place for this setting.
Enable document specific unit system on top of system default unit system.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

vernzimm:
Agree, and normally I switch decimal place at the same time. Would be great if the default could be set for that (maybe in current preference pane), and there was a drop-down somewhere accessible on main screen to swap unit profiles.

PaddleStroke:
I have 2 ideas to solve this : 
![image](https://user-images.githubusercontent.com/19984177/224271028-fa328432-16a5-4dca-8d16-5f3f3551d2da.png)
![image](https://user-images.githubusercontent.com/19984177/224271049-1a5c5749-c8d9-4b92-9fb0-9bdf505eab53.png)

@vernzimm @sliptonic What do you think? I think I like idea B better because it's less clutter. And it's probably intuitive to click on that to try to change unit.

Syres916:
This was discussed on the forum months ago, I even put a patch together using the TuxbWb, see https://forum.freecad.org/viewtopic.php?p=632572#p632572 but yorik decided to make the change in the Draft Wb so if you have Draft in Autoload preferences it's already available in any workbench.

sliptonic:
The solution from Draft is problematic. First, it is changing the current active schema AND changing the user's saved preference but it isn't clear to the user that this is what is happening.  In other words, the change isn't just for the session.  

Second, the draft selector doesn't give the full list of schemas.  It shows a subset of units.  This means it's impossible to select the Metric Small Parts and CNC schema this way or the FEM schema.  
That means there are two different ways to change the default schema but they work differently.  

Idea B is more economical with space but less intuitive.  I would advocate for idea A.   In the future, if real estate is needed in the status bar, it could be revised to idea B and the user would already be trained to look there.  

If you did want to go with idea B, the tooltip could show the name of the schema on hover.

[...]

PR #8870, opened 2023-03-13
Core: Add possibility to change the unit in the sizeLabel (bottom left).

PaddleStroke:
Fixe issue https://github.com/FreeCAD/FreeCAD/issues/8800

image

---

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -67,6 +67,7 @@
 #include <Base/FileInfo.h>
 #include <Base/Interpreter.h>
 #include <Base/Stream.h>
+#include <Base/UnitsApi.h>
 #include <DAGView/DAGView.h>
 #include <TaskView/TaskView.h>
 
@@ -152,7 +153,7 @@ private:
 // Pimpl class
 struct MainWindowP
 {
-    QLabel* sizeLabel;
+    QToolButton* sizeLabel;
     QLabel* actionLabel;
     QTimer* actionTimer;
     QTimer* statusTimer;
@@ -297,8 +298,9 @@ MainWindow::MainWindow(QWidget * parent, Qt::WindowFlags f)
     d->status = new StatusBarObserver();
     d->actionLabel = new QLabel(statusBar());
     // d->actionLabel->setMinimumWidth(120);
-    d->sizeLabel = new QLabel(tr("Dimension"), statusBar());
-    d->sizeLabel->setMinimumWidth(120);
+
+    initialiseSizeLabel();
+
     statusBar()->addWidget(d->actionLabel, 1);
     QProgressBar* progressBar = Gui::SequencerBar::instance()->getProgressBar(statusBar());
     statusBar()->addPermanentWidget(progressBar, 0);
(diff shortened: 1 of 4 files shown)


src/Gui/MainWindow.cpp
```diff
@@ -2079,6 +2081,51 @@ void MainWindow::setPaneText(int i, QString text)
...
+    ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Units");
+    int userSchema = hGrp->GetInt("UserSchema", 0);
+
+    for (int i = 0; i < num; i++) {
+        QAction* action = new QAction(qApp->translate("Gui::Dialog::DlgSettingsUnits", Base::UnitsApi::getDescription(static_cast<Base::UnitSystem>(i))), this);
+        action->setCheckable(true);
+        action->setChecked(i == userSchema);
+        d->sizeLabel->addAction(action);
+
+        QObject::connect(qAsConst(d->sizeLabel)->actions()[i], &QAction::changed,
```
wwmayer, line 2107: Isn't `d->sizeLabel)->actions()[i]` the same as `action`?

src/Gui/MainWindow.cpp
```diff
@@ -2079,6 +2081,51 @@ void MainWindow::setPaneText(int i, QString text)
...
+            this, [this, i, hGrp] {
```
wwmayer, line 2108: Using a QActionGroup makes probably more sense here because if set to exclusive it will automatically uncheck all other actions.

PaddleStroke: Thanks I will look into it.

src/Gui/MainWindow.cpp
PaddleStroke: Yes this is correct thanks.

src/Gui/MainWindow.cpp
```diff
@@ -2079,6 +2081,45 @@ void MainWindow::setPaneText(int i, QString text)
...
+    actionGrp->setExclusive(true);
+
+    for (int i = 0; i < num; i++) {
+        QAction* action = new QAction(qApp->translate("Gui::Dialog::DlgSettingsUnits", Base::UnitsApi::getDescription(static_cast<Base::UnitSystem>(i))), this);
+        action->setCheckable(true);
+        action->setChecked(i == userSchema);
+        actionGrp->addAction(action);
+
+        QObject::connect(action, &QAction::changed,
+            this, [this, i, hGrp] {
```
chennes, line 2110: Linter warning here: you don't need `this`.

src/Mod/Sketcher/Gui/EditModeCoinManager.cpp
```diff
@@ -275,6 +277,11 @@ void EditModeCoinManager::ParameterObserver::updateColor(SbColor &sbcolor, const
     Client.updateInventorColors();
 }
 
+void EditModeCoinManager::ParameterObserver::updateUnit(const std::string &parametername)
```
chennes, line 280: Can you use `Q_UNUSED` (or `NOLINT`) to eliminate this warning?

PaddleStroke:
Thanks @chennes !
I have fixed the commit history, added your comments and rebased on master.

chennes:
Great addition to FreeCAD, thanks for doing it. I'll merge when CI finishes this round.

src/Gui/MainWindow.cpp
wwmayer: Out of curiosity: wasn't it an option to use a QActionGroup instead?

PaddleStroke: Got it. I'll make a PR for this next time I change of branch.

src/Gui/MainWindow.cpp
```diff
@@ -297,8 +298,9 @@ MainWindow::MainWindow(QWidget * parent, Qt::WindowFlags f)
     d->status = new StatusBarObserver();
     d->actionLabel = new QLabel(statusBar());
     // d->actionLabel->setMinimumWidth(120);
-    d->sizeLabel = new QLabel(tr("Dimension"), statusBar());
-    d->sizeLabel->setMinimumWidth(120);
+
+    initialiseSizeLabel();
```
0penBrain, line 302:
```suggestion
    initializeSizeLabel();
```

src/Gui/MainWindow.cpp
```diff
@@ -2079,6 +2081,44 @@ void MainWindow::setPaneText(int i, QString text)
...
+        QAction* action = new QAction(qApp->translate("Gui::Dialog::DlgSettingsUnits", Base::UnitsApi::getDescription(static_cast<Base::UnitSystem>(i))), this);
+        action->setCheckable(true);
+        action->setChecked(i == userSchema);
+        actionGrp->addAction(action);
+
+        QObject::connect(action, &QAction::changed,
+            this, [i, hGrp] {
+                // Set and save the Unit System
+                Base::UnitsApi::setSchema(static_cast<Base::UnitSystem>(i));
+                hGrp->SetInt("UserSchema", i); //Note : saving parameter will trigger EditModeCoinManager::ParameterObserver
```
0penBrain, line 2112: There should be no comment about things that Gui::MainWindow class is not aware.
You can set a general comment such as `Impacted classes have to register an observer` but this is even not needed as this should already be the case.

0penBrain (commented):
Well. @PaddleStroke
On the good side, the code is very clean. :)
On the bad side, just reviewing the code, it suffers a major flaw that at least should as been discovered by some decent functional testing : the newly introduced widget doesn't update when the user changes unit schema from preferences or through Python, which leads to a strong inconsistency.
Can you please take care of a fixing PR ?

PR #8892, opened 2023-03-15
Path: add File source to CustomOp

morganrallen:
original functionality is the default behavior under Text Source


src/Mod/Path/Path/Op/Custom.py
```diff
@@ -47,21 +49,77 @@ def opFeatures(self, obj):
...
+            # probably absolute, just return
+            return filename
+
+        doc_path = os.path.dirname(FreeCAD.ActiveDocument.FileName)
+        prospective_path = os.path.join(doc_path, filename)
+
+        if os.path.exists(prospective_path):
+            return prospective_path
 
     def opExecute(self, obj):
```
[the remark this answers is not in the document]
cam72cam: You could probably dump the output of gcode_file into self.Gcode and keep the same commandlist code that existed before.  That also allows someone to switch it to "Text" mode and edit the path.

src/Mod/Path/Path/Op/Custom.py
```diff
@@ -47,21 +49,77 @@ def opFeatures(self, obj):
...
                 self.commandlist.append(newcommand)
+        elif obj.Source == "File" and len(obj.GcodeFile) > 0:
+            gcode_file = self.findGcodeFile(obj.GcodeFile)
+
+            # could not determine the path
+            if not gcode_file: return
+
+            with open(gcode_file) as fd:
+                for l in fd.readlines():
+                  newcommand = Path.Command(str(l))
```
cam72cam, line 121: This worries me a bit, FreeCAD has a fairly strict definition on what GCode Ops are supported and in what format (line numbers/comments/etc...).  When a custom op is normally created from a file, it uses the behavior in Post/scripts/gcode_pre.py to only include the "whitelisted" functionality that the post_processors can handle.

On the other hand, it is allowed to put any custom gcode in here regardless.  It just may have some odd side effects.

I guess my main concern is that it differs from the "standard" import workflow.

src/Mod/Path/Path/Op/Custom.py
```diff
@@ -47,21 +49,77 @@ def opFeatures(self, obj):
...
+        )
+
         obj.addProperty(
             "App::PropertyStringList",
             "Gcode",
             "Path",
             QT_TRANSLATE_NOOP("App::Property", "The gcode to be inserted"),
         )
 
+        obj.Source = [ "Text", "File" ]
```
cam72cam, line 73: I don't believe this is correct, it should be done through opPropertyEnumerations if I understand correctly.

obj.Source is the value, not the potential valid values and should probably default to "Text".

cam72cam (changes requested):
Overall, I'm looking forward to using this feature.  I think the code might need some work on the edge cases, but it's on the right track.

src/Mod/Path/Path/Op/Custom.py
morganrallen: So far I've been approaching this with the goal of minimizing changes in default behavior so I just copied how it was done previously. I guess the bigger picture question is, **should all internal gcode be parsed to this standard format?**

This could also be put behind a Property switch to enable strict behavior without changing the default.

src/Mod/Path/Path/Op/Custom.py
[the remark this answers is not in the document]
morganrallen: editing is an interesting idea but handling divergence worries me, what would your behavior look like if the user has made an edit, them the Op is recomputed?

src/Mod/Path/Path/Op/Custom.py
morganrallen: I cannot remember which Op I was referencing when I wrote this, but it's done in several places. `Main/Job.py#525` for example. I'm looking into `opPropertyEnumberations` now, looks like I can pass this off to the Base class if used right.

luzpaz:
Please prepend `Path:` to commit title before it gets merge. TIA

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8892">pipeline status</a> for feature branch [PR_8892](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8892). Pipeline [812838350](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/812838350) was triggered at [942a72f](https://github.com/FreeCAD/FreeCAD/pull/8892/commits/942a72f32bd985ab53d8a0c51f94c24ac35dc789). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Path/Path/Op/Custom.py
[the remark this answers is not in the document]
cam72cam: That is a fair point, it's 'source' instead of 'import'.

src/Mod/Path/Path/Op/Custom.py
cam72cam: It's a tricky thing to answer.  Right now gcode_pre is *very* limited, but that's by design.  You are correct that it's a bigger question of how we should handle this scenario in general.

CustomOp is currently loaded with foot-guns, maybe we keep that for now and don't worry about it here.

src/Mod/Path/Path/Op/Custom.py
morganrallen: I've submitted PR #9165. If merged it will allow `opPropertyEnumerations` to be overridden on a subclass to return it's own enumeration values.

sliptonic: #9165 has been merged.

morganrallen: Great I'll update this PR and test to ensure it works as expected.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8892">pipeline status</a> for feature branch [PR_8892](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8892). Pipeline [843831163](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/843831163) was triggered at [ff4388a](https://github.com/FreeCAD/FreeCAD/pull/8892/commits/ff4388a610c77abc2f9339be9a6db3c1b0ae86c9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

morganrallen:
@sliptonic this should be good to merge now

PR #8745, opened 2023-03-05
[Ready for review] Sketcher: Circle to circle distance constraints

FlachyJoe:
I'm trying to continue the work of @davidosterberg started here: #5439

PR #8896, opened 2023-03-15
Sketcher : Circle to circle distance constraints

FlachyJoe:
Implements a circle to circle distance sketcher constraint, including concentric case.

Replace #8745
Forum discussion: https://forum.freecad.org/viewtopic.php?p=564607

@abdullahtahiriyo I took your remarks into account but I broke my git branch somewhere so this is the new place to commit.

### DIFF
--- a/src/Mod/Sketcher/App/ConstraintPyImp.cpp
+++ b/src/Mod/Sketcher/App/ConstraintPyImp.cpp
@@ -224,14 +224,6 @@ int ConstraintPy::PyInit(PyObject* args, PyObject* /*kwd*/)
         if (PyNumber_Check(index_or_value)) { // can be float or int
             SecondIndex = any_index;
             Value = PyFloat_AsDouble(index_or_value);
-            //if (strcmp("Distance",ConstraintType) == 0) {
-            //    this->getConstraintPtr()->Type   = Distance;
-            //    this->getConstraintPtr()->First  = FirstIndex;
-            //    this->getConstraintPtr()->Second = SecondIndex;
-            //    this->getConstraintPtr()->Value  = Value;
-            //    return 0;
-            //}
-            //else
             if (strcmp("Angle",ConstraintType) == 0) {
                 if (PyObject_TypeCheck(index_or_value, &(Base::QuantityPy::Type))) {
                     Base::Quantity q = *(static_cast<Base::QuantityPy*>(index_or_value)->getQuantityPtr());
@@ -240,6 +232,13 @@ int ConstraintPy::PyInit(PyObject* args, PyObject* /*kwd*/)
                 }
                 this->getConstraintPtr()->Type   = Angle;
                 this->getConstraintPtr()->First  = FirstIndex;
+                this->getConstraintPtr()->Second = SecondIndex;
+                this->getConstraintPtr()->setValue(Value);
+                return 0;
+            }
+            else if (strcmp("Distance",ConstraintType) == 0) {
+                this->getConstraintPtr()->Type = Distance;
+                this->getConstraintPtr()->First = FirstIndex;
                 this->getConstraintPtr()->Second = SecondIndex;
                 this->getConstraintPtr()->setValue(Value);
                 return 0;
@@ -481,7 +480,7 @@ std::string ConstraintPy::representation() const
         case Coincident         : result << "'Coincident'>";break;
         case Horizontal         : result << "'Horizontal' (" << getConstraintPtr()->First << ")>";break;
         case Vertical           : result << "'Vertical' (" << getConstraintPtr()->First << ")>";break;
-        case Block            	: result << "'Block' (" << getConstraintPtr()->First << ")>";break;
+        case Block              : result << "'Block' (" << getConstraintPtr()->First << ")>";break;
         case Radius             : result << "'Radius'>";break;
         case Diameter           : result << "'Diameter'>";break;
         case Weight             : result << "'Weight'>";break;
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -1797,6 +1797,20 @@ int Sketch::addConstraint(const Constraint *constraint)
                                         constraint->Second,constraint->SecondPos,
                                         c.value,c.driving);
         }
+        else if (constraint->FirstPos == PointPos::none &&
+            constraint->SecondPos == PointPos::none &&
+            constraint->Second != GeoEnum::GeoUndef &&
+            constraint->Third == GeoEnum::GeoUndef) {        // circle to circle, circle to arc, etc.
+
+            c.value = new double(constraint->getValue());
+            if(c.driving)
+                FixParameters.push_back(c.value);
+            else {
+                Parameters.push_back(c.value);
+                DrivenParameters.push_back(c.value);
+            }
+            rtn = addDistanceConstraint(constraint->First, constraint->Second,c.value,c.driving);
+        }
         else if (constraint->Second != GeoEnum::GeoUndef) {
             if (constraint->FirstPos != PointPos::none) { // point to line distance
                 c.value = new double(constraint->getValue());
@@ -1806,8 +1820,7 @@ int Sketch::addConstraint(const Constraint *constraint)
                     Parameters.push_back(c.value);
                     DrivenParameters.push_back(c.value);
                 }
-                rtn = addDistanceConstraint(constraint->First,constraint->FirstPos,
-                                            constraint->Second,c.value,c.driving);
+                rtn = addDistanceConstraint(constraint->First,constraint->FirstPos,constraint->Second,c.value,c.driving);
             }
         }
         else {// line length
@@ -2735,6 +2748,19 @@ int Sketch::addDistanceConstraint(int geoId1, PointPos pos1, int geoId2, PointPo
 
         int tag = ++ConstraintsCounter;
         GCSsys.addConstraintP2PDistance(p1, p2, value, tag, driving);
+        return ConstraintsCounter;
+    }
+    return -1;
+}
+
+// circle-circle offset distance constraint
+int Sketch::addDistanceConstraint(int geoId1, int geoId2, double * value, bool driving)
+{
+    if ((Geoms[geoId1].type == Circle)  && (Geoms[geoId2].type == Circle)) {
+        GCS::Circle &c1 = Circles[Geoms[geoId1].index];
+        GCS::Circle &c2 = Circles[Geoms[geoId2].index];
+        int tag = ++ConstraintsCounter;
+        GCSsys.addConstraintC2CDistance(c1, c2, value, tag, driving);
         return ConstraintsCounter;
     }
     return -1;
--- a/src/Mod/Sketcher/App/planegcs/Constraints.cpp
+++ b/src/Mod/Sketcher/App/planegcs/Constraints.cpp
@@ -2585,5 +2585,109 @@ double ConstraintEqualLineLength::grad(double *param)
     return deriv*scale;
 }
 
+// ConstraintC2CDistance
+ConstraintC2CDistance::ConstraintC2CDistance(Circle &c1, Circle &c2, double *d)
+{
+    this->d = d;
+    pvec.push_back(d);
+
+    this->c1 = c1;
+    this->c1.PushOwnParams(pvec);
+
+    this->c2 = c2;
+    this->c2.PushOwnParams(pvec);
+
+    origpvec = pvec;
+    pvecChangedFlag = true;
+    rescale();
+}
+
+void ConstraintC2CDistance::ReconstructGeomPointers()
+{
+    int i=0;
+    i++; // skip the first parameter as there is the inline function distance for it
+    c1.ReconstructOnNewPvec(pvec, i);
+    c2.ReconstructOnNewPvec(pvec, i);
+    pvecChangedFlag = false;
+}
+
+ConstraintType ConstraintC2CDistance::getTypeId()
+{
+    return C2CDistance;
+}
+
+void ConstraintC2CDistance::rescale(double coef)
+{
+    scale = coef * 1;
+}
+
+void ConstraintC2CDistance::errorgrad(double *err, double *grad, double *param)
+{
+    if (pvecChangedFlag) ReconstructGeomPointers();
+
+    DeriVector2 ct1 (c1.center, param);
+    DeriVector2 ct2 (c2.center, param);
+
+    DeriVector2 vector_ct12 = ct1.subtr(ct2);
+
+    double length_ct12, dlength_ct12;
+    length_ct12 = vector_ct12.length(dlength_ct12);
+
+    // outer case (defined as the centers of the circles are outside the center of the other circles)
+    // it may well be that the circles intersect.
+    if( length_ct12 >= *c1.rad &&
+        length_ct12 >= *c2.rad ) {
+        if (err) {
+            *err = length_ct12 - (*c2.rad + *c1.rad + *distance());
+        }
+        else if (grad) {
+            double drad = (param == c2.rad || param == c1.rad || param == distance())?-1.0:0.0;
+            *grad = dlength_ct12 + drad;
+        }
+    }
+    else {
+        double * bigradius = (*c1.rad >= *c2.rad)?c1.rad:c2.rad;
+        double * smallradius = (*c1.rad >= *c2.rad)?c2.rad:c1.rad;
+
+        double smallspan = *smallradius + length_ct12 + *distance();
+
+        if (err) {
+            *err = *bigradius - smallspan;
+        }
+        else if (grad) {
+            double drad = 0.0;
+
+            if(param == bigradius) {
+                drad = 1.0;
+            }
+            else if(param == smallradius) {
+                drad = -1.0;
+            }
+            else if(param == distance()) {
+                drad = (*distance()<0.)?1.0:-1.0;
+            }
+
+            *grad = -dlength_ct12 + drad;
+        }
+    }
+}
+
+double ConstraintC2CDistance::error()
+{
+    double err;
+    errorgrad(&err,nullptr,nullptr);
+    return scale * err;
+}
+
+double ConstraintC2CDistance::grad(double *param)
+{
+    if ( findParamInPvec(param) == -1 )
+        return 0.0;
+
+    double deriv;
+    errorgrad(nullptr, &deriv, param);
+
+    return deriv*scale;
+}
 
 } //namespace GCS
--- a/src/Mod/Sketcher/App/planegcs/GCS.h
+++ b/src/Mod/Sketcher/App/planegcs/GCS.h
@@ -315,6 +315,8 @@ namespace GCS
                                    bool flipn1, bool flipn2,
                                    int tagId, bool driving = true);
 
+        int addConstraintC2CDistance(Circle &c1, Circle &c2, double *dist, int tagId, bool driving = true);
+
         // internal alignment constraints
         int addConstraintInternalAlignmentPoint2Ellipse(Ellipse &e, Point &p1, InternalAlignmentType alignmentType, int tagId=0, bool driving = true);
         int addConstraintInternalAlignmentEllipseMajorDiameter(Ellipse &e, Point &p1, Point &p2, int tagId=0, bool driving = true);
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -2150,7 +2150,7 @@ CmdSketcherConstrainDistance::CmdSketcherConstrainDistance()
     sAppModule      = "Sketcher";
     sGroup          = "Sketcher";
     sMenuText       = QT_TR_NOOP("Constrain distance");
-    sToolTipText    = QT_TR_NOOP("Fix a length of a line or the distance between a line and a vertex");
+    sToolTipText    = QT_TR_NOOP("Fix a length of a line or the distance between a line and a vertex or between two circles");
     sWhatsThis      = "Sketcher_ConstrainDistance";
     sStatusTip      = sToolTipText;
     sPixmap         = "Constraint_Length";
@@ -2160,7 +2160,8 @@ CmdSketcherConstrainDistance::CmdSketcherConstrainDistance()
     allowedSelSequences = {{SelVertex, SelVertexOrRoot}, {SelRoot, SelVertex},
                            {SelEdge}, {SelExternalEdge},
                            {SelVertex, SelEdgeOrAxis}, {SelRoot, SelEdge},
-                           {SelVertex, SelExternalEdge}, {SelRoot, SelExternalEdge}};
+                           {SelVertex, SelExternalEdge}, {SelRoot, SelExternalEdge},
+                           {SelEdge, SelEdge}};
 }
 
 void CmdSketcherConstrainDistance::activated(int iMsg)
@@ -2289,6 +2290,55 @@ void CmdSketcherConstrainDistance::activated(int iMsg)
             return;
         }
     }
+    else if (isEdge(GeoId1,PosId1) && isEdge(GeoId2,PosId2))  { // circle to circle distance
+        const Part::Geometry *geom1 = Obj->getGeometry(GeoId1);
+        const Part::Geometry *geom2 = Obj->getGeometry(GeoId2);
+        if (geom1->getTypeId() == Part::GeomCircle::getClassTypeId()
+        && geom2->getTypeId() == Part::GeomCircle::getClassTypeId() ) {
+            auto circleSeg1 = static_cast<const Part::GeomCircle*>(geom1);
+            double radius1 = circleSeg1->getRadius();
+            Base::Vector3d center1 = circleSeg1->getCenter();
+
+            auto circleSeg2 = static_cast<const Part::GeomCircle*>(geom2);
+            double radius2 = circleSeg2->getRadius();
+            Base::Vector3d center2 = circleSeg2->getCenter();
+
+            double ActDist = 0.;
+
+            Base::Vector3d intercenter = center1 - center2;
+            double intercenterdistance = intercenter.Length();
+
+            if( intercenterdistance >= radius1 &&
+                intercenterdistance >= radius2 ) {
+
+                ActDist = intercenterdistance - radius1 - radius2;
+            }
+            else {
+                double bigradius = std::max(radius1,radius2);
+                double smallradius = std::min(radius1,radius2);
+
+                ActDist = bigradius - smallradius - intercenterdistance;
+            }
+
+            openCommand(QT_TRANSLATE_NOOP("Command", "Add circle to circle distance constraint"));
+            Gui::cmdAppObjectArgs(selection[0].getObject(),
+                "addConstraint(Sketcher.Constraint('Distance',%d,%d,%f)) ",
+                GeoId1,GeoId2,ActDist);
+
+            if (arebothpointsorsegmentsfixed || constraintCreationMode==Reference) { // it is a constraint on a external line, make it non-driving
+                const std::vector<Sketcher::Constraint *> &ConStr = Obj->Constraints.getValues();
+
+                Gui::cmdAppObjectArgs(selection[0].getObject(),
+                     "setDriving(%i,%s)",
+                     ConStr.size()-1,"False");
+                finishDatumConstraint (this, Obj, false);
+            }
+            else
+                finishDatumConstraint (this, Obj, true);
+
+            return;
+        }
+    }
     else if (isEdge(GeoId1,PosId1)) { // line length
         if (GeoId1 < 0 && GeoId1 >= Sketcher::GeoEnum::VAxis) {
             Gui::TranslatedNotification(Obj,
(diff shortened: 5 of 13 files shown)


src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -2514,6 +2513,61 @@ void CmdSketcherConstrainDistance::applyConstraint(std::vector<SelIdPair> &selSe
 
         return;
     }
+    case 8: // {SelEdge, SelEdge}
```
FlachyJoe, line 2516: This case aims to handle Two Edges selection but the One Edge case is triggered first and the selection is flushed.

PaddleStroke:
Great to see this is going on :) 
Let me know when you need a windows tester.

src/Mod/Sketcher/App/planegcs/Constraints.cpp
```diff
@@ -2585,5 +2585,112 @@ double ConstraintEqualLineLength::grad(double *param)
...
+
+    origpvec = pvec;
+    pvecChangedFlag = true;
+    rescale();
+}
+
+void ConstraintC2CDistance::ReconstructGeomPointers()
+{
+    int i=0;
+    i++; // there is an inline function for distance
```
xtemp09, line 2608: What does this fragment mean?

FlachyJoe: pvec with index 0 is distance ( d at line 2591) it's skipped here as it doesn't need a reconstruction.
This pvec has an getter defined at src/Mod/Sketcher/App/planegcs/Constraints.h:755

i is not initialized at 1 in order to explicit the skip.

@abdullahtahiriyo You confirm ?

FlachyJoe, line 2608:
```suggestion
    i++; // skip d as there is the inline function distance for it
```

src/Mod/Sketcher/App/ConstraintPyImp.cpp
```diff
@@ -244,6 +244,13 @@ int ConstraintPy::PyInit(PyObject* args, PyObject* /*kwd*/)
                 this->getConstraintPtr()->setValue(Value);
                 return 0;
             }
+            else if (strcmp("Distance",ConstraintType) == 0) {
+                this->getConstraintPtr()->Type = Distance;
+                this->getConstraintPtr()->First = FirstIndex;
+                this->getConstraintPtr()->Second = SecondIndex;
+                this->getConstraintPtr()->setValue(Value);
+                return 0;
+            }
```
0penBrain, line 253: Same block already present (commented) a bit above. Some cleanup needed.

src/Mod/Sketcher/App/Sketch.cpp
```diff
@@ -1797,6 +1797,20 @@ int Sketch::addConstraint(const Constraint *constraint)
                                         constraint->Second,constraint->SecondPos,
                                         c.value,c.driving);
         }
+        else if (constraint->FirstPos == PointPos::none &&
+            constraint->SecondPos == PointPos::none &&
+            constraint->Second != GeoEnum::GeoUndef &&
+            constraint->Third == GeoEnum::GeoUndef) {        // circle to circle, circle to arc, etc.
+
+            c.value = new double(constraint->getValue());
```
0penBrain, line 1805: @abdullahtahiriyo do you know if this heap allocations are correctly freed ?

src/Mod/Sketcher/App/Sketch.cpp
```diff
@@ -1797,6 +1797,20 @@ int Sketch::addConstraint(const Constraint *constraint)
...
+            constraint->Third == GeoEnum::GeoUndef) {        // circle to circle, circle to arc, etc.
+
+            c.value = new double(constraint->getValue());
+            if(c.driving)
+                FixParameters.push_back(c.value);
+            else {
+                Parameters.push_back(c.value);
+                DrivenParameters.push_back(c.value);
+            }
+            rtn = addDistanceConstraint(constraint->First, constraint->Second,c.value,c.driving);
```
0penBrain, line 1812:
```suggestion
            rtn = addDistanceConstraint(constraint->First,constraint->Second,c.value,c.driving);

```
For consistency

src/Mod/Sketcher/App/Sketch.cpp
```diff
@@ -2740,6 +2754,40 @@ int Sketch::addDistanceConstraint(int geoId1, PointPos pos1, int geoId2, PointPo
...
+        //~ GCSsys.addConstraintOffsetRadius(c1, a2, value, tag, driving);
+        //~ return ConstraintsCounter;
+    //~ }
+    //~ else if ((Geoms[geoId1].type == Arc) && (Geoms[geoId2].type == Circle)) {
+        //~ GCS::Arc &a1 = Arcs[Geoms[geoId1].index];
+        //~ GCS::Circle &c2 = Circles[Geoms[geoId2].index];
+        //~ int tag = ++ConstraintsCounter;
+        //~ GCSsys.addConstraintOffsetRadius(a1, c2, value, tag, driving);
+        //~ return ConstraintsCounter;
+    //~ }
```
0penBrain, line 2787: What is that ?

src/Mod/Sketcher/App/planegcs/Constraints.cpp
```diff
@@ -2585,5 +2585,112 @@ double ConstraintEqualLineLength::grad(double *param)
...
+            *err = *bigradius - smallspan;
+        }
+        else if (grad) {
+            double drad = 0;
+
+            if(param == bigradius) {
+                drad = 1.0;
+            }
+            else if(param == smallradius) {
+
```
0penBrain, line 2664: Add a comment here to explain why nothing is done.

src/Mod/Sketcher/App/planegcs/Constraints.cpp
```diff
@@ -2585,5 +2585,112 @@ double ConstraintEqualLineLength::grad(double *param)
...
+        double * bigradius = (*c1.rad >= *c2.rad)?c1.rad:c2.rad;
+        double * smallradius = (*c1.rad >= *c2.rad)?c2.rad:c1.rad;
+
+        double smallspan = *smallradius + length_ct12 + *distance();
+
+        if (err) {
+            *err = *bigradius - smallspan;
+        }
+        else if (grad) {
+            double drad = 0;
```
0penBrain, line 2658:
```suggestion
            double drad = 0.0;
```

src/Mod/Sketcher/App/planegcs/Constraints.cpp
```diff
@@ -2585,5 +2585,112 @@ double ConstraintEqualLineLength::grad(double *param)
...
+            }
+            else if(param == smallradius) {
+
+            }
+            else if(param == distance()) {
+                drad = (*distance()<0.)?1.0:-1.0;
+            }
+            else {
+                drad = 0.0;
+            }
```
0penBrain, line 2671: Not needed as `drad` is initialized to `0.0`

src/Mod/Sketcher/App/planegcs/GCS.h
```diff
@@ -315,6 +315,9 @@ namespace GCS
                                    bool flipn1, bool flipn2,
                                    int tagId, bool driving = true);
 
+        int addConstraintC2CDistance(Circle &c1, Circle &c2, double *dist, int tagId, bool driving = true);
+
+
```
0penBrain, line 320: Remove one newline ?

src/Mod/Sketcher/Gui/CommandConstraints.cpp
0penBrain: Yes, this is a current limitation. You can't have `{SelEdge}` and `{SelEdge, SelEdge}` for the same command as the former will prevent the later to happen.
This can be improved in the future though. :)

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
```diff
@@ -654,25 +655,33 @@ void EditModeConstraintCoinManager::processConstraints(const GeoListFacade & geo
...
                                 const Part::GeomLineSegment *lineSeg = static_cast<const Part::GeomLineSegment *>(geo);
                                 Base::Vector3d l2p1 = lineSeg->getStartPoint();
                                 Base::Vector3d l2p2 = lineSeg->getEndPoint();
                                 // calculate the projection of p1 onto line2
                                 pnt2.ProjectToLine(pnt1-l2p1, l2p2-l2p1);
                                 pnt2 += pnt1;
+
+                            } else if (geo->getTypeId() == Part::GeomCircle::getClassTypeId()) { // circle to circle distance
+                                const Part::Geometry *geo1 = geolistfacade.getGeometryFromGeoId(Constr->First);
+                                auto circleSeg1 = static_cast<const Part::GeomCircle*>(geo1);
```
0penBrain, line 672: Even if it should be OK, I'm always worrying a bit about static_casting without checking type first. :)

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
```diff
@@ -654,25 +655,33 @@ void EditModeConstraintCoinManager::processConstraints(const GeoListFacade & geo
...
                                 break;
                         } else if (Constr->FirstPos != Sketcher::PointPos::none) {
                             pnt2 = geolistfacade.getPoint(Constr->First, Constr->FirstPos);
                         } else if (Constr->First != GeoEnum::GeoUndef) {
-                            const Part::Geometry *geo = geolistfacade.getGeometryFromGeoId(Constr->First);
-                            if (geo->getTypeId() == Part::GeomLineSegment::getClassTypeId()) {
-                                const Part::GeomLineSegment *lineSeg = static_cast<const Part::GeomLineSegment *>(geo);
+                            const Part::Geometry *geo1 = geolistfacade.getGeometryFromGeoId(Constr->First);
+                            if (geo1->getTypeId() == Part::GeomLineSegment::getClassTypeId()) { // segment distance
+                                const Part::GeomLineSegment *lineSeg = static_cast<const Part::GeomLineSegment *>(geo1);
```
0penBrain, line 684: Why ?

src/Mod/Sketcher/Gui/Utils.cpp
```diff
@@ -327,6 +327,48 @@ double SketcherGui::GetPointAngle(const Base::Vector2d& p1, const Base::Vector2d
...
+    double radius1 = circle1->getRadius();
+    double radius2 = circle2->getRadius();
+    Base::Vector3d v = center2 - center1;
+    double length = v.Length();
+    auto p1 = Base::Vector3d(0., 0., 0.);
+    auto p2 = Base::Vector3d(0., 0., 0.);
+    bool retval = false;
+
+    v = v.Normalize();
+    if (length == 0) { //concentric case
```
0penBrain, line 345: Shall we manage `Confusion` here ?

src/Mod/Sketcher/Gui/Utils.h
```diff
@@ -119,6 +119,9 @@ inline bool isEdge(int GeoId, Sketcher::PointPos PosId)
 // Return counter-clockwise angle from horizontal out of p1 to p2 in radians.
 double GetPointAngle (const Base::Vector2d &p1, const Base::Vector2d &p2);
 
+// Set the two points on circles at minimal distance
+bool GetCirclesMinimalDistance(const Part::GeomCircle *circle1, const Part::GeomCircle *circle2, Base::Vector3d *point1, Base::Vector3d *point2);
```
0penBrain, line 123: The return value is never used.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -2160,7 +2160,8 @@ CmdSketcherConstrainDistance::CmdSketcherConstrainDistance()
     allowedSelSequences = {{SelVertex, SelVertexOrRoot}, {SelRoot, SelVertex},
                            {SelEdge}, {SelExternalEdge},
                            {SelVertex, SelEdgeOrAxis}, {SelRoot, SelEdge},
-                           {SelVertex, SelExternalEdge}, {SelRoot, SelExternalEdge}};
+                           {SelVertex, SelExternalEdge}, {SelRoot, SelExternalEdge},
+                           {SelEdge, SelEdge}};
```
0penBrain, line 2164: Command tooltip a few lines above shall be updated.

0penBrain (commented):
Just a code review, no testing ATM.

src/Mod/Sketcher/App/Sketch.cpp
abdullahtahiriyo: Two lines below, they are added to FixedParameters or Parameters. These are freed in clear().

src/Mod/Sketcher/App/planegcs/Constraints.cpp
abdullahtahiriyo: This is a mistake.

```
        double smallspan = *smallradius + length_ct12 + *distance();

        if (err) {
            *err = *bigradius - smallspan;
        }
```

partial(err)/partial(smallradius) = -1, so drad should be -1 for this case.

src/Mod/Sketcher/App/planegcs/Constraints.cpp
abdullahtahiriyo: This is fair. It can be removed.

abdullahtahiriyo:
@0penBrain 

Thank you OpenBrain. Your auditing is as always very good. I have replied to some comments that corresponded to my code. You make very good points in general, I leave those for @FlachyJoe.

I have just built the branch. I have detected a small issue in the minimal testing I have done:

If you create a first circle and fully constraint it. Then create a second circle inside but not concentric. Do not constraint its radius. Now set a distance constraint edge to edge (circle-to-circle). Now make the center of the circles concentric. The solver gives an overconstrained message. However, if you invert the order and do the coincident first and then the circle-to-circle it works fine.

Then, there is something that I do not know what is best for the user exactly. If you constrain two circles and set the circle-to-circle to reference constraint, you can overlap the edges of the two circles, distances will be negative. However, this cannot be enforced with a negative value (FreeCAD refuses). I do not know if this is good or bad. It looks odd though...

src/Mod/Sketcher/Gui/Utils.h
FlachyJoe: Could be used elsewhere, maybe in the coming line to circle distance or for the arc generalization. It has no charge to keep it.

0penBrain: It doesn't harm the code. But it can bring confusion to one reading the code. Very minor though. :)

FlachyJoe:
> Just a code review, no testing ATM.

Thank you ! Some test artifacts were still presents.
It's my first substantial C++ PR and I don't feel comfortable yet with so many files for one feature. I'm almost lost sometimes. :)

0penBrain:
@FlachyJoe CI is failing because of a missing `;` ;)

FlachyJoe:
> CI is failing because of a missing `;` ;)

:crying_cat_face: 
I don't understand why this C++ end line obligation still exists. Does anybody know one case where the precompiler can't decide if it's the end or not ?

0penBrain:
@FlachyJoe still failing :)

0penBrain:
> It's my first substantial C++ PR and I don't feel comfortable yet with so many files for one feature. I'm almost lost sometimes. :)

All in all it's a very good PR. And yes, not a simple one as touching most parts of the Sketcher (I'm myself not able to in-depth review the solver part). A good IDE helps handling this if you didn't installed one yet. ;)

@abdullahtahiriyo may you have some insights on the still unresolved comments please ?

@FlachyJoe then it will be time for squashing. Are you comfortable with this ?

FlachyJoe:
@0penBrain Done with local check.

@abdullahtahiriyo 
> If you create a first circle and fully constraint it. Then create a second circle inside but not concentric. Do not constraint its radius. Now set a distance constraint edge to edge (circle-to-circle). Now make the center of the circles concentric. The solver gives an overconstrained message.

I can't reproduce with the last changes.

0penBrain:
> @0penBrain Done with local check.

Not sure what you're talking about. :)

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
abdullahtahiriyo: You are right.

We need either type checking and static casting, or dynamic casting. I would use type checking and static_casting for consistency (geo is checked, geo1 is not checked).

FlachyJoe:
> Not sure what you're talking about. :)

edited

> it will be time for squashing. Are you comfortable with this ?

I don't know how to keep abdullahtahiriyo as co-author in the squashed commit else it's ok

0penBrain:
> I don't know how to keep abdullahtahiriyo as co-author in the squashed commit else it's ok

If you squash all commits together, you should end up with a commit co-authored by @abdullahtahiriyo and you.
My thought was that maybe you could squash the fix commits only with the first commit if the fixes only touch it specifically. But it's a bit more complex to achieve. :)

0penBrain:
> My thought was that maybe you could squash the fix commits only with the first commit if the fixes only touch it specifically. But it's a bit more complex to achieve. :)

OK, just tried but doesn't work straight forward. Probably the simplest solution will be to squash commits altogether. :)

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
abdullahtahiriyo: For example:

```
const Part::Geometry *geo1 = geolistfacade.getGeometryFromGeoId(Constr->First);

if(geo1->getTypeId() == Part::GeomCircle::getClassTypeId()) {
    auto circleSeg1 = static_cast<const Part::GeomCircle*>(geo1);
    auto circleSeg2 = static_cast<const Part::GeomCircle*>(geo);
    GetCirclesMinimalDistance(circleSeg1, circleSeg2, &pnt1, &pnt2);
}
```

FlachyJoe:
@abdullahtahiriyo I just find an annoying case
1. Full constraint a circle
2. Create a circle inside
3. Create a circle to circle distance
4. Drag the small circle center out of the big one
5. Solver complains for invalid solution while the mouse is in a forbidden zone = only solution for negative radius

NB: I'm sorry but your contribution was hidden when I squashed.

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
FlachyJoe:
```suggestion
                                if (geo1->getTypeId() == Part::GeomCircle::getClassTypeId()
                                    auto circleSeg1 = static_cast<const Part::GeomCircle*>(geo1);
                                else
                                    break;
```

I anticipate arc distance can be implemented here with a GeomCircle created with the GeomArcOfCircle parameters.

src/Mod/Sketcher/Gui/Utils.cpp
abdullahtahiriyo: I would have said that length == 0 is always false, if it weren't because the UI paints that case. So the solver is that good in my simple case.

However, what would happen if length is not zero, but very close to zero. Well, v will be a normalized vector and it will be treated under the corresponding inner case. That looks like ok.

src/Mod/Sketcher/Gui/Utils.h
abdullahtahiriyo, line 123: It is unconventional to pass stack parameters by pointer (convention is by reference). A developer reading the code tends to think that point1 and point2 are dynamically allocated in the heap. It is thus a little bit misleading.

```..., Base::Vector3d & point1, Base::Vector3d& point2);```

is what we are looking for.

Comment (abdullahtahiriyo) [reply src/Mod/Sketcher/Gui/Utils.h:123]: This function could difficultly be reused for an arc o a point as type is hardcoded, and there is no inheritance relationship. At best it might be templated.

I think YAGNI, but if you insist, I would at least update the documentation (use three slashes instead of two and indicate what it returns), then in most IDEs the developer using the code will be notified by the IDE.

**Comment by abdullahtahiriyo** [comment]
> I can't reproduce with the last changes.

Yes, it was the small radius differential that @0penBrain realised it was missing.

**Comment by abdullahtahiriyo** [comment]
> I don't know how to keep abdullahtahiriyo as co-author in the squashed commit else it's ok

Do not worry. Just squash everything together under your name. I was only helping around.

**Comment by abdullahtahiriyo** [comment]
> 5\. Solver complains for invalid solution while the mouse is in a forbidden zone = only solution for negative radius

Somehow it does not quite work well around the overlapping area. This is what I was commenting in the morning. Sometimes you can do this, but not always:

image

I would go through the code trying to find somewhere where a negative value is not allowed, because I think the solver should take it... but I am too tired now and need to sleep. For another day...

File: src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp (hunk 11 above)
Comment (abdullahtahiriyo) [reply src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp]: You are going to have some issues there. You are going to need a specifically constructed temporary object. Circles are not directly constructable from ArcOfCircles, so you will have to init it (at least via the OCC handle which is not the cleanest).

The code you propose is odd, circleSeg1 is never used. It goes out of scope just when going out of the "if". C++ is a strongly typed language. You cannot "change type". You can fake it to some extent with std::variant, but it has its own caveats too.

In any case, please always use braces for if-else. We didn't use to and lately we have changed this approach, because it leads to many more errors which reviews often miss.

If you want to extend this to arcs (and I guess combinations of arcs and circles), here or in another PR, I would make:

```
template <T1, T2>
bool GetCirclesMinimalDistance(T1 circle_or_arc1, T2 circle_or_arc2, &pnt1, &pnt2);
```

because the API of both is common, you may get away without much adaptation for type.

However, you will still need to make sure to call that function with the right cast-ed types (static o dynamic cast-ed).

File: src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp [hunk 16]
```diff
@@ -669,10 +669,15 @@ void EditModeConstraintCoinManager::processConstraints(const GeoListFacade & geo
...
+                                const Part::GeomCircle *circleSeg1;
+                                if (geo1->getTypeId() == Part::GeomCircle::getClassTypeId())
+                                    circleSeg1 = static_cast<const Part::GeomCircle*>(geo1);
+                                else
+                                    break;
+
                                 auto circleSeg2 = static_cast<const Part::GeomCircle*>(geo);
 
-                                GetCirclesMinimalDistance(circleSeg1, circleSeg2, &pnt1, &pnt2);
+                                GetCirclesMinimalDistance(circleSeg1, circleSeg2, pnt1, pnt2);
```
abdullahtahiriyo, line 680: I do not see an advantage of this over a single if statement as I indicated in the other comment. Because circleSeg1 is already of type GeomCircle, it cannot be reused in the else branch for an GeomArcOfCircle.

I think it is less readable like this, as if is not immediately clear when the circleSeg2 and the function are called.

If you really want to leave arc ready code in this PR as an exercise, you could template the function and try to write compiling code for Circles and ArcsOfCircle here. I think you will soon realise the limitations.

Yet an alternative, in Part::GeomCurve (Geometry.h and Geometry.cpp in Part Mod), you write a member function:
```void closestParameters(const GeomCurve * other, double&u1, double&u2)```

Then using OCCT you calculate the parameters in the curve and in the other curve (u1 and u2) corresponding to the closest points.

This u1 and u2, you pass them via Part::pointAtParameter (of the corresponding geomcurve). Now you have support for all geomcurves, and you can just write here:

if(geo is GeomCurve and geo1 is GeomCurve) 
call those two functions to get the points.

If you want to do this, take a look to ```GeomCurve::intersect``` which uses OCCT ```GeomAPI_ExtremaCurveCurve ```. However, this will also come with additional caveats, as extrema (read the docs) sometimes does not provide the closest distance between curves, and for this specific use case, you would need distance between basis curves (non-trimmed curves) anyway...

So, probably now you are in a best position to select:
(a) I will keep this PR minimal and do a minimal readable call to the function I have.
(b) I want to learn function templates, I was waiting for this opportunity for a long time.
(c) I am getting bored and this OCCT world looks amasing...

In any case, I would still choose (a) and do another PR targeting (b) or (c), if you feel like  :)

File: src/Mod/Sketcher/Gui/Utils.cpp [hunk 17]
```diff
@@ -328,45 +328,36 @@ double SketcherGui::GetPointAngle(const Base::Vector2d& p1, const Base::Vector2d
...
             if (radius1 > radius2){
-                p1 = center1 + v * radius1;
-                p2 = center2 + v * radius2;
+                point1 += v * radius1;
+                point2 += v * radius2;
             } else {
-                p1 = center1 - v * radius1;
-                p2 = center2 - v * radius2;
+                point1 += -v * radius1;
+                point2 += -v * radius2;
```
abdullahtahiriyo, line 354: There is operator -=, but actually the way it is here is more visual, as one can spot the differences between calls.

BTW, I like the refactoring.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1429,10 +1429,14 @@ void ViewProviderSketch::moveConstraint(int constNum, const Base::Vector2d &toPo
 
             } else if (geo->getTypeId() == Part::GeomCircle::getClassTypeId()) { // circle to circle distance
                 const Part::Geometry *geo1 = GeoList::getGeometryFromGeoId (geomlist, Constr->First);
-                const Part::GeomCircle *circleSeg1 = static_cast<const Part::GeomCircle *>(geo1);
-                const Part::GeomCircle *circleSeg2 = static_cast<const Part::GeomCircle *>(geo);
+                const Part::GeomCircle *circleSeg1;
+                if (geo1->getTypeId() == Part::GeomCircle::getClassTypeId())
+                    circleSeg1 = static_cast<const Part::GeomCircle *>(geo1);
+                else
+                    return;
```
abdullahtahiriyo, line 1436: Same comment as above applies here.

abdullahtahiriyo:
I have added a very small commit. It allows to set negative distances. It gives some consistency to switching from reference constraints (which necessarily accept negative values).

Give it a try and let me know.

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
FlachyJoe: I follow your wise advice.

FlachyJoe:
Circle error when drag is not related to this PR

https://user-images.githubusercontent.com/10371513/226168854-ad99219b-5872-40ac-be18-16371184f247.mp4

abdullahtahiriyo:
> Circle error when drag is not related to this PR

Definitely not. It is often that circles want to become points in FreeCAD and even take negative space. :D

abdullahtahiriyo:
If you are happy with the behaviour, squash it and then I will merge.

FlachyJoe:
> If you are happy with the behaviour, squash it and then I will merge.

Done !
Thank you very much for the escort.

0penBrain:
Congrats @FlachyJoe :)

luzpaz:
Well done!

PaddleStroke:
Yeah ! Thanks guys for the welcome addition.

adrianinsaval:
Awesome! Please remember to add this to the release notes wiki page

FlachyJoe:
@abdullahtahiriyo 
The small radius differential fix broke the concentric case. This small circle shouldn't be fully constraint:
Capture d’écran_2023-03-19_20-53-36

I have a fix, do you approve:
```
        else if (grad) {
            double drad = 0.0;

            if(param == bigradius) {
                drad = 1.0;
            }
            else if(param == smallradius) {
                drad = -1.0;
            }
            else if(param == distance()) {
                drad = (*distance()<0.)?1.0:-1.0;
            }
            if (length_ct12>1e-13) {
                *grad = -dlength_ct12 + drad;
            } else { // concentric case
                *grad = drad;
            }
        }
```

0penBrain:
> The small radius differential fix broke the concentric case. This small circle shouldn't be fully constraint:

Not so obvious. :)
Actually your sketch has only 1 remaining DoF. So the logics says one of the 2 circles is necessarily fully constrained. :) That's the mathematical side.
But, from a user point of view, both circles can still be moved. So none is fully constrained. That's the human side.

FlachyJoe:
> Actually your sketch has only 1 remaining DoF.

Yes

> So the logics says one of the 2 circles is necessarily fully constrained.

No ;)
Circles are in relation but, as you see with lines, there is no reason for the small one to be blocked.
With this configuration you can't add the small diameter constraint.

> May you open a PR with the fix so it's easier ?

Done with #8967

FlachyJoe:
@adrianinsaval 

> Please remember to add this to the release notes wiki page

Done with minimal words, is more needed ?

adrianinsaval:
looks sufficient to me, other relevant wiki pages may need updating too. Like https://wiki.freecad.org/Sketcher_Workbench and https://wiki.freecad.org/Sketcher_ConstrainDistance if you don't have the time to do this please ask on the wiki forum for help.

Roy-043:
> 3\. Do functional tests on it. Look if both ways of triggering the constraint (selection+button and button first+click) work fine.

Post-selection (=button first) does not seem to work yet.
```
Word size of FreeCAD: 64-bit
Version: 0.21.0.32637 (Git)
Build type: Release
Branch: master
Hash: 7fa7cb7f0b222864be7416c683166876eab190bc
Python 3.10.10, Qt 5.15.8, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: English/United States (en_US) [ OS: Dutch/Netherlands (nl_NL) ]
Installed mods:
[/code]

```

0penBrain:
> Post-selection (=button first) does not seem to work yet.

Post-selection is not possible because of the 'one-click' line segment constraining. It "eats" the click to any sketch curve.
We could improve something there. :)

FlachyJoe:
> I see that implementing this for arcs has been discussed. That would indeed make sense IMO.

I'm working on this.

PR #8904, opened 2023-03-16
[Lint] remove checks for line endings

donovaly:
- there is no need to check for certain OS line endings, git handles them well apparently

@0penBrain , this is the mentioned PR to get rid of the lint line ending warnings


0penBrain:
@donovaly ATM please don't remove code but instead set the default value of 'checkLineendings' input to 'false'. No more is needed.

donovaly:
> @donovaly ATM please don't remove code but instead set the default value of 'checkLineendings' input to 'false'. No more is needed.

Thanks. Done: #8907

PR #8906, opened 2023-03-16
[FEM] improve deletion warning

donovaly:
- don't popup a warning if all childs of an object are selected too when deleting the object

### DIFF
--- a/src/Mod/Fem/Gui/ViewProviderAnalysis.cpp
+++ b/src/Mod/Fem/Gui/ViewProviderAnalysis.cpp
@@ -31,6 +31,7 @@
 # include <Inventor/nodes/SoSeparator.h>
 #endif
 
+#include <App/Document.h>
 #include <App/MaterialObject.h>
 #include <App/TextDocument.h>
 #include <Gui/ActionFunction.h>
@@ -38,6 +39,8 @@
 #include <Gui/Control.h>
 #include <Gui/Document.h>
 #include <Gui/MainWindow.h>
+#include <Gui/Selection.h>
+#include <Gui/SelectionObject.h>
 #include <Mod/Fem/App/FemAnalysis.h>
 #include <Mod/Fem/App/FemConstraint.h>
 #include <Mod/Fem/App/FemMeshObject.h>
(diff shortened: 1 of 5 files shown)


src/Mod/Fem/Gui/ViewProviderSolver.cpp
```diff
@@ -55,11 +59,29 @@ std::vector<std::string> ViewProviderSolver::getDisplayModes() const
...
+                    break;
+                }
+            }
+            if (!found)
+                break;
+        }
+        if (found)// all childs are selected too
+            return true;
+
         // generate dialog
```
wwmayer, line 85: The same code block exists four times. Consider moving it to a helper function.

donovaly: I thought this as well - after I made the PR. The question is where to put the code. I will made a proposal.

src/Mod/Fem/Gui/ViewProviderSolver.cpp
```diff
@@ -55,11 +59,29 @@ std::vector<std::string> ViewProviderSolver::getDisplayModes() const
...
+                if (std::string(child->getNameInDocument())
+                    == std::string(selection.getFeatName())) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found)
+                break;
+        }
+        if (found)// all childs are selected too
```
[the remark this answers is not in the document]
donovaly: According to dictionaries yes, but in real-life it is quite common (I read it even in scientific papers). I also think we can be more flexible since people understand the comment and "childs" is shorter than "children". (language lives ;-) )

src/Mod/Fem/Gui/ViewProviderSolver.cpp
donovaly: I updated the PR accordingly, the shared function is in ViewProviderAnalysis.

src/Mod/Fem/Gui/ViewProviderAnalysis.cpp
```diff
@@ -251,15 +254,41 @@ void ViewProviderFemAnalysis::dropObject(App::DocumentObject *obj)
...
+}
+
+bool ViewProviderFemAnalysis::checkSelectedChildren(const std::vector<App::DocumentObject*> objs,
+                                                    std::string objectName)
+{
+    // warn the user if the object has unselected children
     if (!objs.empty()) {
+        // check if all children are in the selection
+        bool found = false;
+        auto docGui = Gui::Application::Instance->activeDocument();
```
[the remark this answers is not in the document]
donovaly: I updated the PR accordingly.
However, for curiosity, how can one make a selection to be deleted while not being in the active document?

PR #8920, opened 2023-03-16
App: Fix missing python addProperty and removeProperty of Document object

FlachyJoe:
The python Document class didn't have addProperty nor removeProperty but such behavior was available throw GUI.

Forum discussion (french): https://forum.freecad.org/viewtopic.php?t=76855

### DIFF
--- a/src/App/DocumentPyImp.cpp
+++ b/src/App/DocumentPyImp.cpp
@@ -40,6 +40,37 @@
 using namespace App;
 
 
+PyObject*  DocumentPy::addProperty(PyObject *args)
+{
+    char *sType,*sName=nullptr,*sGroup=nullptr,*sDoc=nullptr;
+    short attr=0;
+    std::string sDocStr;
+    PyObject *ro = Py_False, *hd = Py_False;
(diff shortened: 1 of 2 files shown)


src/App/DocumentPyImp.cpp
```diff
@@ -40,6 +40,37 @@
...
+
+    if (sDoc) {
+        sDocStr = sDoc;
+        PyMem_Free(sDoc);
+    }
+
+    getDocumentPtr()->addDynamicProperty(sType,sName,sGroup,sDocStr.c_str(),attr,
+            Base::asBoolean(ro), Base::asBoolean(hd));
+
+    return Py::new_reference_to(this);
```
0penBrain, line 61: @wwmayer Is it sensible to return a reference to the object (Document in this case) rather than to the created property ?
I saw it's same in DocumentObjectPy implementation, so it's a general question. :)

wwmayer: We don't have Python wrappers for Property classes. When trying to access a property of a document(object) you get directly its value. So, yes, it's sensible to return a reference of itself.

PR #8923, opened 2023-03-17
[Core] Create `ElementMap.h` and tests

### LINKED ISSUE #8768: [Issue]

### Forums discussion

_No response_

### Version

0.21 (Development)

### Full version info

```shell
[code]
Word size of FreeCAD: 64-bit
Build type: Debug
Branch: master
[/code]
```

### Subproject(s) affected?

None

### Issue description

> The toponaming is more easily worked on if the descriptive elements of it are gathered in one place for easy reference.

## description

The goal here is to create a common place for the consts and other names that will be used by toponaming as a whole.

`ElementMap.h`

At first, this will be mainly the strings used for serialization of the MappedName's elements. As common point of reference are discovered related to the toponaming ElementMap algorithm, this will be a good place to place them.

Later, in phase 2, this header file will be used with the associated ElementMap.cpp class.

For context, see issue https://github.com/FreeCAD/FreeCAD/issues/8432

## notes

Go ahead and move the contents of src/Mod/Part/App/TopoShapeOpCode.h to here. Include the conditional directives.

For the following items, see ComplexGeoData.cpp, TopoShapeEx.cpp.

| postfix (elementMap key) | meaning       |
|----------------|---------------|
| `:H`           | tag (local)   |
| `:C`           | child element |
| `:X`           | external tag  |
| `:I`           | index         |
| `D`            | duplicate     |
| `:T`           | deprecated tag (uses decimal) |
| `:M`           | mod           |
| `:MG`            | modgen        |
| `:G`           | gen           |
| `:U`           | upper         |
| `:L`           | lower         |

(above edited based on RealThunder's feedback)

For example,

```cpp
const std::string &ComplexGeoData::tagPostfix() {
    static std::string postfix(elementMapPrefix() + ":H");
    return postfix;
}
```

in ComplexGeoData.cpp chould be moved here to simply be something like:

```cpp
const std::string POSTFIX_FOR_TAG = ";:H";
```

I recommend we also go ahead and move `elementMapPrefix()` to a const as it simply returns a semicolon (always).

### Anything else?

_No response_

### Code of Conduct

AjinkyaDahale:
`ElementMap.h` for now contains a list of constants which will be used in the `ElementMap` class in a future phase of toponaming solution.

Intended to fix #8768.

### DIFF
--- a/src/App/ElementMap.h
+++ b/src/App/ElementMap.h
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+/****************************************************************************
+ *   Copyright (c) 2018-2022 Zheng, Lei (realthunder)                       *
+ *   <realthunder.dev@gmail.com>                                            *
+ *   Copyright (c) 2023 FreeCAD Project Association                         *
+ *                                                                          *
+ *   This file is part of FreeCAD.                                          *
+ *                                                                          *
+ *   FreeCAD is free software: you can redistribute it and/or modify it     *
+ *   under the terms of the GNU Lesser General Public License as            *
+ *   published by the Free Software Foundation, either version 2.1 of the   *
+ *   License, or (at your option) any later version.                        *
+ *                                                                          *
+ *   FreeCAD is distributed in the hope that it will be useful, but         *
+ *   WITHOUT ANY WARRANTY; without even the implied warranty of             *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU       *
+ *   Lesser General Public License for more details.                        *
+ *                                                                          *
+ *   You should have received a copy of the GNU Lesser General Public       *
+ *   License along with FreeCAD. If not, see                                *
+ *   <https://www.gnu.org/licenses/>.                                       *
+ *                                                                          *
+ ***************************************************************************/
+
+#ifndef DATA_ELEMENTMAP_H
+#define DATA_ELEMENTMAP_H
+
+#include "FCGlobal.h"
+#include "IndexedName.h"
+
+namespace Data {
+
+static constexpr const char *POSTFIX_TAG = ";:H";
+static constexpr const char *POSTFIX_DECIMAL_TAG = ";:T";
+static constexpr const char *POSTFIX_EXTERNAL_TAG = ";:X";
+static constexpr const char *POSTFIX_CHILD = ";:C";
+static constexpr const char *POSTFIX_INDEX = ";:I";
+static constexpr const char *POSTFIX_UPPER = ";:U";
+static constexpr const char *POSTFIX_LOWER = ";:L";
+static constexpr const char *POSTFIX_MOD = ";:M";
+static constexpr const char *POSTFIX_GEN = ";:G";
+static constexpr const char *POSTFIX_MODGEN = ";:MG";
+static constexpr const char *POSTFIX_DUPLICATE = ";D";
+
+} // namespace data
+
+#endif // DATA_ELEMENTMAP_H
--- a/src/Mod/Part/App/TopoShapeOpCode.h
+++ b/src/Mod/Part/App/TopoShapeOpCode.h
@@ -1,0 +1,99 @@
+/****************************************************************************
+ *   Copyright (c) 2018 Zheng Lei (realthunder) <realthunder.dev@gmail.com> *
+ *                                                                          *
+ *   This file is part of the FreeCAD CAx development system.               *
+ *                                                                          *
+ *   This library is free software; you can redistribute it and/or          *
+ *   modify it under the terms of the GNU Library General Public            *
+ *   License as published by the Free Software Foundation; either           *
+ *   version 2 of the License, or (at your option) any later version.       *
+ *                                                                          *
+ *   This library  is distributed in the hope that it will be useful,       *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of         *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the          *
+ *   GNU Library General Public License for more details.                   *
+ *                                                                          *
+ *   You should have received a copy of the GNU Library General Public      *
+ *   License along with this library; see the file COPYING.LIB. If not,     *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,          *
+ *   Suite 330, Boston, MA  02111-1307, USA                                 *
+ *                                                                          *
+ ****************************************************************************/
+
+#ifndef PART_TOPOSHAPE_OPCODE_H
+#define PART_TOPOSHAPE_OPCODE_H
+
+/** Definition of commonly used TopoShape operational code
+ *
+ * The operational code (op code) is encoded into the mapped element name to
+ * provide more context and meaning to the  shape history when tracing it back
+ * to its predecessors. Some op codes can be passed to the generalized shape-
+ * making API TopoShape::makEBoolean() to make a shape.
+ */
+namespace Part {
+
+class PartExport OpCodes {
+public:
+    /** Element name encoding scheme version number
+     *
+     * Increase this version if there is major change in encoding scheme.
+     * Opening a document containing a mismatched version number will cause the
+     * element map to be regenerated after recompute
+     */
+
+    static constexpr const int Version = 15;
+
+    /** @name Op codes that are also accepted by TopoShape::makEBoolean() */
+    //@{
+    static constexpr const char *Fuse = "FUS";
+    static constexpr const char *Cut = "CUT";
+    static constexpr const char *Common = "CMN";
+    static constexpr const char *Section = "SEC";
+    static constexpr const char *Compound = "CMP";
+    static constexpr const char *Compsolid = "CSD";
+    static constexpr const char *Pipe = "PIP";
+    static constexpr const char *Shell = "SHL";
+    static constexpr const char *Wire = "WIR";
+    //@}
+
+    static constexpr const char *Tag = "TAG";
+    static constexpr const char *Copy = "CPY";
+    static constexpr const char *Transform = "XFM";
+    static constexpr const char *Gtransform = "GFM";
+    static constexpr const char *Face = "FAC";
+    static constexpr const char *FilledFace = "FFC";
+    static constexpr const char *Extrude = "XTR";
+    static constexpr const char *GeneralFuse = "GFS";
+    static constexpr const char *Refine = "RFI";
+    static constexpr const char *Boolean = "BOL";
+    static constexpr const char *Slice = "SLC";
+    static constexpr const char *Maker = "MAK";
+    static constexpr const char *Fillet = "FLT";
+    static constexpr const char *Chamfer = "CHF";
+    static constexpr const char *Thicken = "THK";
+    static constexpr const char *Offset = "OFS";
+    static constexpr const char *Offset2D = "OFF";
+    static constexpr const char *Revolve = "RVL";
+    static constexpr const char *Loft = "LFT";
+    static constexpr const char *Sweep = "SWP";
+    static constexpr const char *PipeShell = "PSH";
+    static constexpr const char *ShellFill = "SHF";
+    static constexpr const char *Solid = "SLD";
+    static constexpr const char *RuledSurface = "RSF";
+    static constexpr const char *Mirror = "MIR";
+    static constexpr const char *Sketch = "SKT";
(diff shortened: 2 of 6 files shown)


AjinkyaDahale:
As mentioned in [a comment in the issue](https://github.com/FreeCAD/FreeCAD/issues/8768#issuecomment-1472626983), I have not moved constants from `TopoShapeOpCode` since it did seem to belong to `Part`-specific items.

AjinkyaDahale:
@realthunder are there any other constants that can be stored here? Also, question, where are the mod/gen/upper/lower strings in your own branch?

JohnAD:
@AjinkyaDahale 

My thinking of moving the TopoShapeOpCode items was that future programmers would find it convenient to have a single place to "interpret the meaning" of a particular mapped name. Since since these are simple constants from a standard default workbench, it would not add much complication to the Part workbench later.

Is the new TopoShapeOpCode.h file more about the naming algorithm or more about the Part workbench? It is both; so there is not a clear answer for where to put them.

If you are keeping them separate, you might consider placing the `TopoShapeOpCode.h` into the Part workbench now, and place a comment reference in the `ElementMap.h` file to the TopoShapeOpCode for future programmers to follow.

chennes:
See https://github.com/KDE/clazy/blob/master/docs/checks/README-non-pod-global-static.md

src/App/ElementMap.h
```diff
@@ -0,0 +1,48 @@
...
+#ifndef DATA_ELEMENTMAP_H
+#define DATA_ELEMENTMAP_H
+
+#include "FCGlobal.h"
+#include "IndexedName.h"
+
+namespace Data {
+
+const std::string POSTFIX_TAG = ";:H";
+const std::string POSTFIX_DEC_TAG = ";:T";
```
adrianinsaval, line 35: Maybe call this POSTFIX_DECIMAL_TAG for more clarity? At least to me it's not immediately obvious what dec stands for, could just be me though. Might be worth adding a comment about this being there only for backwards compatibility.

src/App/ElementMap.h
```diff
@@ -0,0 +1,48 @@
...
+const std::string POSTFIX_EXTERNAL_TAG = ";:X";
+const std::string POSTFIX_CHILD = ";:C";
+const std::string POSTFIX_INDEX = ";:I";
+const std::string POSTFIX_UPPER = ";:U";
+const std::string POSTFIX_LOWER = ";:L";
+const std::string POSTFIX_MOD = ";:M";
+const std::string POSTFIX_GEN = ";:G";
+const std::string POSTFIX_MODGEN = ";:MG";
+const std::string POSTFIX_DUPLICATE = ";D";
```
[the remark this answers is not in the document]
AjinkyaDahale: I believe the answer to the first is there may be other meanings when the `:` is not used. As for why this particular postfix doesn't have the colon, I'm just taking it from the issue. I don't see it used in the `development/toponaming` branch, but RT may have just not merged it.

adrianinsaval:
`TopoShapeOpCode.h` has mostly Part WB specific stuff but I also see some mentions of sketcher and draft, presumably when implementing the algorithm on other WBs one would want to add the op codes for those there too so unless we're going to split those for each workbench I think it might make more sense to move it into App

Note that I'm very much an amateur so don't take me _too_ seriously.

AjinkyaDahale:
> `TopoShapeOpCode.h` has mostly Part WB specific stuff but I also see some mentions of sketcher and draft,...

AFAIK Part is used in those WB's as well.

I do expect there may be other such WB specific names. Can't think of operations right now, but I'm sure there may be operations that are defined for, say path, but not used in part.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923">pipeline status</a> for feature branch [PR_8923](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923). Pipeline [812852228](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/812852228) was triggered at [e3ced6c](https://github.com/FreeCAD/FreeCAD/pull/8923/commits/e3ced6ce2e8a6f1068ffd169f30d0c3ad48b5413). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/App/ElementMap.h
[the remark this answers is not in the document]
AjinkyaDahale: @JohnAD @realthunder could you pitch in?

AjinkyaDahale:
> See https://github.com/KDE/clazy/blob/master/docs/checks/README-non-pod-global-static.md

@chennes I've replaced the `std::string` global vars with `char*` now.

chennes:
Awesome. I'm thinking we'll hold the merge for a short time here to see if either @realthunder or @JohnAD can comment on the `:` syntax question above. That feels like something we want to make sure we understand before going too much further. When you have a minute can you take a look at #8974? There wasn't a lot to do there, so it should be a fast review.

src/Mod/Part/App/TopoShapeOpCode.h
```diff
@@ -0,0 +1,99 @@
...
+ *   Suite 330, Boston, MA  02111-1307, USA                                 *
+ *                                                                          *
+ ****************************************************************************/
+
+#ifndef PART_TOPOSHAPE_OPCODE_H
+#define PART_TOPOSHAPE_OPCODE_H
+
+/** Definition of commonly used TopoShape operaional code
+ *
+ * The operaional code (op code) is encoded into the mapped element name to
```
chennes, line 28:
```suggestion
 * The operational code (op code) is encoded into the mapped element name to
```

src/Mod/Part/App/TopoShapeOpCode.h
```diff
@@ -0,0 +1,99 @@
...
+ *   License along with this library; see the file COPYING.LIB. If not,     *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,          *
+ *   Suite 330, Boston, MA  02111-1307, USA                                 *
+ *                                                                          *
+ ****************************************************************************/
+
+#ifndef PART_TOPOSHAPE_OPCODE_H
+#define PART_TOPOSHAPE_OPCODE_H
+
+/** Definition of commonly used TopoShape operaional code
```
chennes, line 26:
```suggestion
/** Definition of commonly used TopoShape operational code
```

tests/src/App/ElementMap.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+class ElementMapTest : public ::testing::Test {
+};
+
+TEST_F(ElementMapTest, defaultConstruction)
+{
+  // Act
+
+  // Assert
+}
```
chennes, line 19: Needs to close the `NOLINTBEGIN` (or just eliminate it, since there aren't any magic numbers here 😁 ).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923">pipeline status</a> for feature branch [PR_8923](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923). Pipeline [813587061](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/813587061) was triggered at [125eacd](https://github.com/FreeCAD/FreeCAD/pull/8923/commits/125eacdee6f0e8d0a3fc55c444a89e2a61d2f8c6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Part/App/TopoShapeOpCode.h
```diff
@@ -0,0 +1,99 @@
...
+
+#ifndef PART_TOPOSHAPE_OPCODE_H
+#define PART_TOPOSHAPE_OPCODE_H
+
+/** Definition of commonly used TopoShape operational code
+ *
+ * The operational code (op code) is encoded into the mapped element name to
+ * provide more meaning shape history when tracking back. Some code can be
+ * passed to generalized shape making API TopoShape::makEBoolean() to make
+ * shape.
```
chennes, line 31: Some slight grammatical strangeness here. How about something like:
```suggestion
 * The operational code (op code) is encoded into the mapped element name to
 * provide more context and meaning to the  shape history when tracing it back
 * to its predecessors. Some op codes can be passed to the generalized shape-
 * making API TopoShape::makEBoolean() to make a shape.
```
Also: why do all these TopoShape::makE functions have a capital "E" in their function name? is this some kind of convention from OCC?

AjinkyaDahale: Probably @realthunder  would be better confirming if the change removes some semantics.

My best guess about `makE` is that it stands for `make E...` where E stands for something.

src/App/ElementMap.h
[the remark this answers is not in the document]
realthunder: My intention is for `;:` as the separator for a `postfix` that may be used for parsing later on. `;D` is not exposed in my original code as a `postfix`. It is internally used as a default way to rename duplicated element that may be overridden by derived class.

src/Mod/Part/App/TopoShapeOpCode.h
realthunder: `E` stand for `Element`. I choose the naming to distinguish the element mapping aware interface from others in the transition period. Once it gets matured, we can drop that convention.

realthunder:
The first priority of the topo naming encoding scheme is to generate unique and stable names for geometry elements. Decoding and history tracing is a `good to have` feature, which is why there is no strict syntax definition. Another reason is that, for many cases, there are ambiguities when doing history tracing. You can't get back the source element even with a strict syntax definition. What we should guarantee is that a model done by an old version syntax will work correctly and can continue modeling with new syntax in newer release. The syntax change may break history tracing, which may affect how good the software can auto fix common topological changing problem. But that's acceptable, as long as it can detect the change and report the failure to user in a meaningful way.

With that in mind, you'll see that `TopoShapeOpCodes` is just a convenience to centralize all the model operation code. It's exact content is not that important. I prefer to make it in `Part` workbench, because currently only `Part` and other workbenches that uses `Part` make use of it. It is good not to have duplicate op codes, but not a must. So spread the codes in other workbenches shouldn't be a problem either.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923">pipeline status</a> for feature branch [PR_8923](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923). Pipeline [813945472](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/813945472) was triggered at [f0e9d86](https://github.com/FreeCAD/FreeCAD/pull/8923/commits/f0e9d865f40a54dfaa34d65baefb7b3c22f07fe3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923">pipeline status</a> for feature branch [PR_8923](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923). Pipeline [814738102](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/814738102) was triggered at [2aca160](https://github.com/FreeCAD/FreeCAD/pull/8923/commits/2aca160f428163a73ddab38ee952689a706f8bbc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923">pipeline status</a> for feature branch [PR_8923](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8923). Pipeline [814755132](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/814755132) was triggered at [27a9879](https://github.com/FreeCAD/FreeCAD/pull/8923/commits/27a987949f660980ae0baa020bcc4647f0a3da65). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
@AjinkyaDahale this looks ready to merge to me -- anything else you want to do before that happens?

AjinkyaDahale:
@chennes no I think we're fine.

PR #8948, opened 2023-03-18
TD: move all XML query handling to a single function

wwmayer:

---

### DIFF
--- a/src/Mod/TechDraw/App/CMakeLists.txt
+++ b/src/Mod/TechDraw/App/CMakeLists.txt
@@ -175,6 +175,8 @@ SET(TechDraw_SRCS
     TechDrawExport.h
     ProjectionAlgos.cpp
     ProjectionAlgos.h
+    XMLQuery.cpp
+    XMLQuery.h
 )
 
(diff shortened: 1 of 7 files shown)


wwmayer:
This PR refactors the code where classes of the QtXmlPatterns module are used. The pattern of using QXmlQuery is always the same so that the code could be simplified to have a single function and the special handling is done with lambda expressions.

In the final step only the function XMLQuery::processItems() must be ported.

This PR already allows it to fully compile the TechDraw module with Qt6 and functionality-wise the only limitation at the moment is that editable text markers of SVG files cannot be replaced.

@WandererFan can you have a look, please?

WandererFan:
Works ok.  Thank you!

chennes:
The merged PR has a failing test in the Addon Manager code on Ubuntu 22.04 -- I can think of no reason this would have affected that, so I will look at that AM test and see why it is overly fragile.

wwmayer:
The CI tests of the PR have all passed and the error only occurred after merging it. And since after adding further PR no error is reported any more I guess there must be something that causes some random behaviour.

Issue #8949, opened 2023-03-18
[Issue] PartDesign: (extrude) Pad inward taper angle orientation selection


tbMaker1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.0.32362 (Git)
Build type: Release
Branch: master
Hash: 45401e30ce29c8a8c66bced6d2db216c96227289
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: French/France (fr_FR)
[/code]
```

### Subproject(s) affected?

PartDesign

### Issue description

(Issue was modified to Feature Request to avoid overlaping with other issue.)

Currently, in version 20.2, the outward pad taper angle can be set positive or negative.

The inward taper angle direction, on the other hand, is fixed in one direction.

Or it can be useful to be able to have inward taper angle either parallel or opposite to the outward taper angle direction.

An extra parameter in the Pad extrude menu could be useful to be able to select the direction for inward taper.

### Anything else?

![TaperAngleCorrected](https://user-images.githubusercontent.com/128159531/226114940-6f7af7d4-109d-4457-b4c9-166dac3e8ec7.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
isn't this a duplicate issue?

tbMaker1:
Yes and no. It was not only about the other bug (which is more than just single edge issue), it was about the possibility to adjust the orientation as needed.

adrianinsaval:
In that case this should have been made as exclusively about adjusting orientation

tbMaker1:
But adjusting orientation per internal feature would not have been the same topic either (because much more complex and probably not adequate).
Here, the simplification assumed was that all internal feature would be in one way or the other. 
Therefore the implication that the discrepency could/would be removed.

adrianinsaval:
sorry if I wasn't clear but that is not what I meant, what I mean is that we should leave the problem with circles in it's own issue as it is not related to this one. Make it exclusively about allowing to choose to have parallel or opposite orientation.

tbMaker1:
Ok

Issue #8587, opened 2023-02-22
PartDesign: Taper function error (in OCC)


FlubM:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

http://forum.freecadweb.org:80/viewtopic.php?f=13&t=75281

### Version

0.20 (Release)

### Full version info

```shell
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.2.29177 +426 (Git)
Build type: Release
Branch: (HEAD detached from 0.20.2)
Hash: 930dd9a76203a3260b1e6256c70c1c3cad8c5cb8
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: German/Switzerland (de_CH)
[/code]
```

### Subproject(s) affected?

PartDesign

### Issue description

The taper function has an error.
Depending on where the hole is, the taper jumps from negative to positive.
If you move the large hole to the left, the taper changes from positive to negative
It seems to be a bug in the FC software

### Anything else?

![Picture_1](https://user-images.githubusercontent.com/122787106/220684675-04e2fdbe-e5f9-4a93-bfb6-e46c13cc1b36.PNG)
![Picture_2](https://user-images.githubusercontent.com/122787106/220684716-f3f28211-0803-4012-85b5-dc9cc7c5c1cd.PNG)
[TaperError.zip](https://github.com/FreeCAD/FreeCAD/files/10805928/TaperError.zip)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
confirmed, maybe it's a bug when offsetting circles or when detecting if they are inside other profiles. Other shapes besides circles seem to work fine, even constructing a circle using two arcs works correctly

donovaly:
This is a long-stranding bug in OCC. I don't see how we could fix this. Here the 3 different outputs of OCC when you move only the upper point slightly:

![FreeCAD_XaOmSYEAOD](https://user-images.githubusercontent.com/1828501/221082001-251704b8-e68f-49bc-9cb5-69128ff5b3db.png)
![FreeCAD_fn6v8uAKUT](https://user-images.githubusercontent.com/1828501/221082002-48de63e1-d7a6-464a-a604-94982ec58c4f.png)
[...]

PR #8950, opened 2023-03-18
PartDesign: (extrude) Pad inward taper angle orientation

tbMaker1:
This PR corrects the discrepency of taper angle in the PD extrude function. (issue: #8949)

The single edge and multi-edge inner sketch had opposite direction taper angles. It adds a checkbox to enable to select if the inward and outward taper angles are opposite or parallel.

- [ x]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

---
TaperAngle

TaperAngleCorrected

CheckBox

### DIFF
--- a/src/Mod/Part/App/ExtrusionHelper.cpp
+++ b/src/Mod/Part/App/ExtrusionHelper.cpp
@@ -58,6 +58,7 @@ void ExtrusionHelper::makeDraft(const TopoDS_Shape& shape,
                                 const double LengthRev,
                                 const double AngleFwd,
                                 const double AngleRev,
+                                bool taperParallel,
                                 bool isSolid,
                                 std::list<TopoDS_Shape>& drafts,
                                 bool isPartDesign)
@@ -156,13 +157,10 @@ void ExtrusionHelper::makeDraft(const TopoDS_Shape& shape,
                     createTaperedPrismOffset(TopoDS::Wire(singleWire), vecRev, distanceRev, true, offsetWire);
                 }
                 else {
-                    // there is an OCC bug with single-edge wires (circles), see inside createTaperedPrismOffset
-                    if (numEdges > 1 || !isPartDesign)
-                        // inner wires must get the negated offset
-                        createTaperedPrismOffset(TopoDS::Wire(singleWire), vecRev, -distanceRev, true, offsetWire);
-                    else
-                        // circles in PartDesign must not get the negated offset
+                    if (taperParallel)
                         createTaperedPrismOffset(TopoDS::Wire(singleWire), vecRev, distanceRev, true, offsetWire);
+                    else
+                        createTaperedPrismOffset(TopoDS::Wire(singleWire), vecRev, -distanceRev, true, offsetWire);
                 }
                 if (offsetWire.IsNull())
                     return;
@@ -204,13 +202,10 @@ void ExtrusionHelper::makeDraft(const TopoDS_Shape& shape,
                     createTaperedPrismOffset(TopoDS::Wire(singleWire), vecFwd, distanceFwd, false, offsetWire);
                 }
                 else {
-                    // there is an OCC bug with single-edge wires (circles), see inside createTaperedPrismOffset
-                    if (numEdges > 1 || !isPartDesign)
-                        // inner wires must get the negated offset
-                        createTaperedPrismOffset(TopoDS::Wire(singleWire), vecFwd, -distanceFwd, false, offsetWire);
-                    else
-                        // circles in PartDesign must not get the negated offset
+                    if (taperParallel)
                         createTaperedPrismOffset(TopoDS::Wire(singleWire), vecFwd, distanceFwd, false, offsetWire);
+                    else
+                        createTaperedPrismOffset(TopoDS::Wire(singleWire), vecFwd, -distanceFwd, false, offsetWire);
                 }
                 if (offsetWire.IsNull())
                     return;
--- a/src/Mod/PartDesign/App/FeatureExtrude.cpp
+++ b/src/Mod/PartDesign/App/FeatureExtrude.cpp
@@ -63,8 +63,9 @@ short FeatureExtrude::mustExecute() const
         UseCustomVector.isTouched() ||
         Direction.isTouched() ||
         ReferenceAxis.isTouched() ||
-        AlongSketchNormal.isTouched() ||
-        Offset.isTouched() ||
+        AlongSketchNormal.isTouched() || 
+        Offset.isTouched() || 
+        InwardTaperParallel.isTouched() ||
         UpToFace.isTouched())
         return 1;
     return ProfileBased::mustExecute();
@@ -232,27 +233,28 @@ void FeatureExtrude::generateTaperedPrism(TopoDS_Shape& prism,
                                           const double L2,
                                           const double angle,
                                           const double angle2,
-                                          const bool midplane)
+                                          const bool midplane,
+                                          const bool taperParallel)
 {
     std::list<TopoDS_Shape> drafts;
     bool isSolid = true; // in PD we only generate solids, while Part Extrude can also create only shells
     bool isPartDesign = true; // there is an OCC bug with single-edge wires (circles) we need to treat differently for PD and Part
     if (method == "ThroughAll") {
         Part::ExtrusionHelper::makeDraft(sketchshape, direction, getThroughAllLength(),
-            0.0, Base::toRadians(angle), 0.0, isSolid, drafts, isPartDesign);
+            0.0, Base::toRadians(angle), 0.0, taperParallel, isSolid, drafts, isPartDesign);
     }
     else if (method == "TwoLengths") {
         Part::ExtrusionHelper::makeDraft(sketchshape, direction, L, L2,
-            Base::toRadians(angle), Base::toRadians(angle2), isSolid, drafts, isPartDesign);
+            Base::toRadians(angle), Base::toRadians(angle2), taperParallel, isSolid, drafts, isPartDesign);
     }
     else if (method == "Length") {
         if (midplane) {
             Part::ExtrusionHelper::makeDraft(sketchshape, direction, L / 2, L / 2,
-                Base::toRadians(angle), Base::toRadians(angle), isSolid, drafts, isPartDesign);
+                Base::toRadians(angle), Base::toRadians(angle), taperParallel, isSolid, drafts, isPartDesign);
         }
         else
             Part::ExtrusionHelper::makeDraft(sketchshape, direction, L, 0.0,
-                Base::toRadians(angle), 0.0, isSolid, drafts, isPartDesign);
+                Base::toRadians(angle), 0.0, taperParallel, isSolid, drafts, isPartDesign);
     }
 
     if (drafts.empty()) {
@@ -282,10 +284,12 @@ void FeatureExtrude::updateProperties(const std::string &method)
     bool isReversedEnabled = false;
     bool isUpToFaceEnabled = false;
     bool isTaperVisible = false;
+    bool isInwardTaperParallelVisible = false;
     bool isTaper2Visible = false;
     if (method == "Length") {
         isLengthEnabled = true;
         isTaperVisible = true;
+        isInwardTaperParallelVisible = true;
         isMidplaneEnabled = true;
         isReversedEnabled = !Midplane.getValue();
     }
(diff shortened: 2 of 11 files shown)


src/Mod/Part/App/ExtrusionHelper.cpp
```diff
@@ -156,13 +157,10 @@ void ExtrusionHelper::makeDraft(const TopoDS_Shape& shape,
                     createTaperedPrismOffset(TopoDS::Wire(singleWire), vecRev, distanceRev, true, offsetWire);
                 }
                 else {
-                    // there is an OCC bug with single-edge wires (circles), see inside createTaperedPrismOffset
-                    if (numEdges > 1 || !isPartDesign)
-                        // inner wires must get the negated offset
-                        createTaperedPrismOffset(TopoDS::Wire(singleWire), vecRev, -distanceRev, true, offsetWire);
-                    else
-                        // circles in PartDesign must not get the negated offset
```
adrianinsaval, line 164: are you sure this should be removed? don't we need an occt version check here? @donovaly thoughts on this?

tbMaker1: It works without. But I may not have cleaned enough.
Furthermore, I saw the issue #8587 and I am looking into it as it relates also to taper errors.

tbMaker1: (I did not correct the Part WB because I don't use it much. I did not implement the checkbox in Part and the discrepency of inner taper angle is still there.)

adrianinsaval: If you look at the comment it is saying this was done to workaround an occt bug, if this is not necessary anymore it likely means it was solved in occt, but we should identify in which version and then apply the workaround for the versions where it's needed

adrianinsaval: Looking again, this file is used in PartWB too, you can't modify it without taking into account it's use in Part WB

tbMaker1: For the PartWB, the checkbox implementation is needed.
For occt, I am looking into it for the #8587. Do you know how to display occt objects in the console? Is there a library for that? where is it?

donovaly:
I will try you PR when you tell me it is ready.
(I am very curious because the issue is in OCC not in FreeCAD.)

Regarding the new option introduced, we purposely don't have this. We discussed about this and there is no use case.
We have the policy not to fill the dialogs with everything that might be possible but is in effect not used.
So what is the use case of having the taper angle for holes the same at the outer wall?

src/Mod/Part/App/ExtrusionHelper.cpp
donovaly: > For the PartWB, the checkbox implementation is needed.

As I wrote we purposely don't have this option. To avoid that we open this discussion here, let's focus with this PR on the actual bug that OCC treats circles (single-line wire) in a different way than polyline wires .

>Do you know how to display occt objects in the console?

What do you mean? OCC (OpenCasCade) od the CAD library FreeCAD is based on. FreeCAD's Part and PartDesign WB are basically a frontend for OCC.

src/Mod/PartDesign/App/FeatureExtrude.cpp
```diff
@@ -63,8 +63,9 @@ short FeatureExtrude::mustExecute() const
         UseCustomVector.isTouched() ||
         Direction.isTouched() ||
         ReferenceAxis.isTouched() ||
-        AlongSketchNormal.isTouched() ||
-        Offset.isTouched() ||
+        AlongSketchNormal.isTouched() ||
```
donovaly, line 66: why this change?

donovaly:
Hmm, I see that you simply deleted the work around for the OCC bug. What is your OCC version? I have OCC 7.6.3.

Here is a FreeCAD test file:
[Taper-test.zip](https://github.com/FreeCAD/FreeCAD/files/11014223/Taper-test.zip)

This is how the test file looks for me:
FreeCAD_7KewAJk3jL

This is correct. Your changes break the file for me since then the taper angles for the circles are wrong.

src/Mod/Part/App/ExtrusionHelper.cpp
tbMaker1: I found out how to do that. For OCC objects, DumpJson() is the equivalent to ToString(). To output it to the console, I use         `
myOcctObject.DumpJson(aSStream);
std::string& resultstr = aSStream.str();
Base::Console().Message(resultstr.c_str());
`
My question was if there was librairie/functions in Freecad taylored to do something like that to ease the debug of OCC objects.

However, this method is not convenient. So I tried to compile Freecad in debug mode and use the occt.natvis visualizer for Visual Studio.

But now, I struggled to compile the OCC in debug and use this debug version in the Freecad debug version.
The cmake points to libpack occ version.

It would be great if the OCC debug version could be included in the libpack and a switch be implemented in the cmake to be able to compile and use both freecad and occ in debug mode.

tbMaker1:
> So what is the use case of having the taper angle for holes the same at the outer wall?

The use case is manufacturability when you make molds for objects. 
Right now, to make a box for example, that has the correct draft for molding:
* you make a cube, draft 4 sides, and make thickness

With the current extrude on pad, you could make the same but the taper does not allow to extract the box because of undercuts.

But you need both parallel and opposite depending on the feature of your objects.

tbMaker1:
> This is correct. Your changes break the file for me since then the taper angles for the circles are wrong.

Your image seems great to me, all tapers correct. You can unmold such an object.
I'll try your file to see if I am missing something. (But I broke my setup trying to compile and use OCC in debug.)

tbMaker1:
> This is correct. Your changes break the file for me since then the taper angles for the circles are wrong.

taper Test file

This is your test file with version 20.2. If you make a square instead of a circle, the 2 circles on the left have their taper switch side.

It didn't break anything. I solved a problem of manufacturability that needs the taper either parallel or opposite. 
The taper issue, reported in other threads #8587 is not solved. I am looking into it.

src/Mod/Part/App/ExtrusionHelper.cpp
tbMaker1: Ok, it seems it has the debug version of OCC. 
However I saw the thread: https://forum.freecad.org/viewtopic.php?f=10&t=47017, which does not mention that.
I'll verify.

donovaly:
> The use case is manufacturability when you make molds for objects. Right now, to make a box for example, that has the correct draft for molding:
> 
>     * you make a cube, draft 4 sides, and make thickness

I think you have a thinko here. The taper functionality wad added specially for molds. We made a survey in the forum and I made one in real-life (calling injection molding and sieve printing guys). The implementation as it is is exactly how molds are made:
FreeCAD_UtozgETqsJ

Can you show me a mold that is not done like this?

donovaly:
> This is your test file with version 20.2. If you make a square instead of a circle, the 2 circles on the right have their taper switch side.

Yes, this is the OCC bug. Again, please tell me what is your OCC version?

In every case deleting the existing workaround code for circles would break the taper feature for all users with OCC 7.4.x - 7.6.x and is therefore no option. In case you use OCC 7.7, we can add code that will only be used by this OCC version.

tbMaker1:
> I think you have a thinko here.

I think you are taking that personaly without understanding what I am saying.

You added a great function to pads.
I want to complete this function to make it more general.
There was a bug in your function. It is still there. I didn't add an extra one. (And, sorry, but it cannot be called a workaround if it doesn't work)
I am working to solve this bug.
If you don't want me to. I can stop. As I said, I am mainly using the PartDesign_draft function to make my molds.

I am using the Libpack2.6 with OCC 7.6.3.

This is a example with a box, and you cannot unmold it. With parallel taper, you can unmold it.
box
part and molds

Parallel and opposite
part and molds2

tbMaker1:
> ... the PartDesign_draft function to make my molds.

= molds / parts designed for molding

donovaly:
> I think you are taking that personaly without understanding what I am saying.

Why so aggressive?

I am not taking it personally. We had a week-long debate about this 1.5 years back or so. In your post the first 2 images show that you make a mistake: you used a positive taper angle. Change the taper angle sign and you can demold your parts. And this design is what I referred to as the standard mold design.

The design in your first 2 images does not exist in reality -that was the outcome of the debate.

> You added a great function to pads. I want to complete this function to make it more general. There was a bug in your function. It is still there. I didn't add an extra one.

Where is the bug? You delete the workaround code to fix at least some occurrences of the OCC bug. This will introduce the bug for all users with OCC 7.5.x or older. This is of course no option.
For OCC 7.6.3 I will need more time to check.

> I am using the Libpack2.6 with OCC 7.6.3.

LibPack 2.6 has OCC 7.6.2 not 7.6.3.
However, please use the latest LibPack 2.8.2: https://github.com/FreeCAD/FreeCAD-LibPack/releases/tag/2.8.2

Sine you compile using the LibPack, you fond all compilation info, also for OCC here:
https://wiki.freecad.org/Compile_on_Windows

adrianinsaval:
Uwe you might want to look at any plastic cup or similar geometries out there, these are injection moulded and have parallel angles for inner and outer side, the current behaviour was agreed upon as the most common expected/required behaviour not necessarily as the only one ever needed. Of course you could use a shell for these but that's known to be prone to failure, also you can always just make one pad and one pocket, which might be the most sane approach actually.

Regarding the bug with circle offset, it seems to not behave the same for all circles, I'll search for a recent report where we found bugs with this even with the workaround. So we might need to keep the workaround but need to investigate what conditionals it needs. I suspect we might need to check the circle orientation.

adrianinsaval:
it was already referenced somewhere here #8587 as can be seen the workaround is sometimes working sometimes not which suggest there is some other factor besides just the number of edges to take into account for this.

btw wasn't this exclusively about circles? is the number of edges a good way to identify those? what about closed bsplines and ellipses?

tbMaker1:
> Why so aggressive?

I am not aggressive. I am not a native english speaker, but I suspected "thinko" what not a compliment. So I guessed you were taking the current situation the wrong way, as I am trying to make Freecad better, as you are, I suppose.

> The design in your first 2 images does not exist in reality
??

> Where is the bug?

I showed you the image of your test file with the released Freecad 20.2. I did not contribute to this released version. The taper function, as shown, does not work properly.


> Change the taper angle sign and you can demold your parts.
If you change the taper sign, you reduce the width of your part, which, limits the height (and the thickness is not constant)
taper

> So we might need to keep the workaround

The bug will be there until it is solved.


> I suspect we might need to check the circle orientation.

That is why I asked for simple ways to output info to the console. Orientation can be simple to put to the console.

tbMaker1:
> btw wasn't this exclusively about circles? is the number of edges a good way to identify those? what about closed bsplines and ellipses?

it is not just about the circle but its seems that its position with regard to the outer contour, the geometry of the contour, the size of the circle, and probably other factors are involved.

adrianinsaval:
I meant that it only affected circles

donovaly:
> > Change the taper angle sign and you can demold your parts.

You are confusing me. The point is that there are 2 types of molds around:

- double-taper (e.g. sieve or paste printing):
FreeCAD_SwwLqeZJ9r
- straight outer frame edge and inner taper (e.g. injection molding):
FreeCAD_FFV2ySjXZG

Parallel taper does not exist. This was the outcome of the discussion and as I wrote I also celled mold manufacturers, sieve printers etc. You parallel-taper mold might exist, but is then a special case.
And this is the reason why we have no option for a parallel taper.

If you have a real-life application for a parallel taper, please share it.

tbMaker1:
When I wrote molds, as I commented afterwards, I implied not only molds but (more) the parts designed for molding.
Anyway, I don't insist on the checkbox. I'll go on using the PartDesign_draft. I'll close the issue and the pull request I made.

donovaly:
For the parallel taper, the thickness feature can be used.

I don't want to disappoint you but changed to the basic features (Pad/Pocket/revolve etc.) needs to be discussed anyway in the forum.
For the future please open a forum discussion about the feature you plan. If you get positive feedback, move on and code. This statement applies only for the basic features. For special features, one can usually directly start coding.

Concerning the OCC bug: https://github.com/FreeCAD/FreeCAD/issues/8587#issuecomment-1442733932
I checked and it is still on OCC 7.6.3. I don't see how we can fix it other than directly in OCC.
The sad thing is that the OCC fix was already reported to the OCC people but they don't care. (OCC is a company and when a bug does not affect their products, they don't give it a high priority). Therefore for the Windows builds (the LibPack) we include already our own set of OCC fixes (the ones from user "blobfish")

PR #8952, opened 2023-03-18
Dimension indicator improvement, improves #8870

0penBrain:
This PR aims at improving #8870 

Main improvements are :
* Use QPushButton for lighter style display and consistency in the status bar
* Fix issue that unit schema change in Preferences or through Python isn't too into account
* Live retranslate texts when user changes UI language

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -171,8 +171,7 @@ public:
         auto* actionGrp = new QActionGroup(menu);
         int num = static_cast<int>(Base::UnitSystem::NumUnitSystemTypes);
         for (int i = 0; i < num; i++) {
-            QAction* action = menu->addAction(qApp->translate("Gui::Dialog::DlgSettingsUnits",
-                    Base::UnitsApi::getDescription(static_cast<Base::UnitSystem>(i))));
+            QAction* action = menu->addAction(QStringLiteral("UnitSchema%1").arg(i));
             actionGrp->addAction(action);
             action->setCheckable(true);
             QObject::connect(action, &QAction::toggled, this, [this,i](bool checked) {
@@ -183,9 +182,10 @@ public:
                         // Update the application to show the unit change
                         Gui::Application::Instance->onUpdate();
                     }
-            } );
+                } );
         }
         setMenu(menu);
+        retranslateUi();
         unitChanged();
         getWindowParameter()->Attach(this);
     }
@@ -203,6 +203,16 @@ public:
         }
     }
 
+    void changeEvent(QEvent *event) override
+    {
+        if (event->type() == QEvent::LanguageChange) {
+            retranslateUi();
+        }
+        else {
+            QPushButton::changeEvent(event);
+        }
+    }
+
 private:
     void unitChanged(void)
     {
@@ -214,6 +224,17 @@ private:
         auto action = actions[userSchema];
         if(!action->isChecked()) { // Using a QSignalBlocker leads to issue
             action->setChecked(true);
+        }
+    }
+
+    void retranslateUi() {
+        auto actions = menu()->actions();
+        int maxSchema = static_cast<int>(Base::UnitSystem::NumUnitSystemTypes);
+        assert(actions.size() <= maxSchema);
+        for(int i = 0; i < maxSchema ; i++)
+        {
+            actions[i]->setText(qApp->translate("Gui::Dialog::DlgSettingsUnits",
+                    Base::UnitsApi::getDescription(static_cast<Base::UnitSystem>(i))));
         }
     }
 };
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -150,6 +150,74 @@ private:
     int _timeout;
 };
 
+/**
+ * The DimensionWidget class is aiming at providing a widget used in the status bar that will:
+ *  - Allow application to display dimension information such as the viewportsize
+ *  - Provide a popup menu allowing user to change the used unit schema (and update if changed elsewhere)
+ */
+class DimensionWidget : public QPushButton, WindowParameter
+{
+    Q_OBJECT
+
+public:
+    explicit DimensionWidget(QWidget* parent): QPushButton(parent), WindowParameter("Units")
+    {
+        setFlat(true);
+        setText(qApp->translate("Gui::MainWindow", "Dimension"));
+        setMinimumWidth(120);
+
+        //create the action buttons
+        auto* menu = new QMenu(this);
+        auto* actionGrp = new QActionGroup(menu);
+        int num = static_cast<int>(Base::UnitSystem::NumUnitSystemTypes);
+        for (int i = 0; i < num; i++) {
+            QAction* action = menu->addAction(qApp->translate("Gui::Dialog::DlgSettingsUnits",
+                    Base::UnitsApi::getDescription(static_cast<Base::UnitSystem>(i))));
+            actionGrp->addAction(action);
+            action->setCheckable(true);
+            QObject::connect(action, &QAction::toggled, this, [this,i](bool checked) {
+                    if(checked) {
+                        // Set and save the Unit System
+                        Base::UnitsApi::setSchema(static_cast<Base::UnitSystem>(i));
+                        getWindowParameter()->SetInt("UserSchema", i);
+                        // Update the application to show the unit change
+                        Gui::Application::Instance->onUpdate();
+                    }
+            } );
+        }
+        setMenu(menu);
+        unitChanged();
+        getWindowParameter()->Attach(this);
+    }
+
+    ~DimensionWidget()
+    {
+        getWindowParameter()->Detach(this);
+    }
+
+    void OnChange(Base::Subject<const char*> &rCaller, const char * sReason) override
+    {
+        Q_UNUSED(rCaller)
+        if (strcmp(sReason, "UserSchema") == 0) {
+            unitChanged();
+        }
+    }
+
+private:
+    void unitChanged(void)
+    {
+        int userSchema = getWindowParameter()->GetInt("UserSchema", 0);
+        auto actions = menu()->actions();
+        if(Q_UNLIKELY(userSchema < 0 || userSchema >= actions.size())) {
+            userSchema = 0;
+        }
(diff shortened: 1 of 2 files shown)


0penBrain:
@PaddleStroke may you please review ?

@wwmayer I wonder about architecture. Now I added the custom widget as an inline class in `MainWindow.cpp` -- trying to keep it short -- but maybe you'd prefer another way

src/Gui/MainWindow.cpp
```diff
@@ -150,11 +150,91 @@ class CustomMessageEvent : public QEvent
     int _timeout;
 };
 
+class DimensionWidget : public QPushButton, WindowParameter
+{
+    Q_OBJECT
+
+public:
+    DimensionWidget(QWidget* parent): QPushButton(parent), WindowParameter("Units")
```
wwmayer, line 158: Mark the constructor explicit

src/Gui/MainWindow.cpp
```diff
@@ -150,11 +150,91 @@ class CustomMessageEvent : public QEvent
...
+
+    ~DimensionWidget()
+    {
+        getWindowParameter()->Detach(this);
+    }
+
+    void OnChange(Base::Subject<const char*> &rCaller, const char * sReason) override
+    {
+        Q_UNUSED(rCaller)
+        if (strcmp(sReason, "UserSchema") == 0) unitChanged();
```
wwmayer, line 196: I think the unitChanged() should be moved to the next line. Lint then also requires curly braces.

src/Gui/MainWindow.cpp
```diff
@@ -150,11 +150,91 @@ class CustomMessageEvent : public QEvent
...
+            QPushButton::changeEvent(event);
+        }
+    }
+
+private:
+    void unitChanged(void)
+    {
+        int userSchema = getWindowParameter()->GetInt("UserSchema", 0);
+        auto actions = menu()->actions();
+        if(Q_UNLIKELY(userSchema < 0 || userSchema > actions.size())) userSchema = 0;
```
wwmayer, line 214: Move userSchema = 0 to next line

wwmayer:
> I wonder about architecture. Now I added the custom widget as an inline class in MainWindow.cpp -- trying to keep it short -- but maybe you'd prefer another way

OK for now. But the file already has more than 2000 lines and at some point we should consider moving the additional classes to their own files.

src/Gui/MainWindow.cpp
0penBrain: Well, that was an attempt to keep the number of lines pretty low. :)
To me, this way is far less error-prone than the 2-line way with no braces -- BTW the linter doesn't say anything here --.
So if you require a multi-line form, I'll necessarily add the curly braces. ;)

wwmayer: IIRC we once had this discussion to avoid to write this on the same line as the if condition because it's easier to overlook.

src/Gui/MainWindow.cpp
```diff
@@ -150,11 +150,91 @@ class CustomMessageEvent : public QEvent
...
+    {
+        setFlat(true);
+        setText(qApp->translate("Gui::MainWindow", "Dimension"));
+        setMinimumWidth(120);
+
+        //create the action buttons
+        auto* menu = new QMenu(this);
+        auto* actionGrp = new QActionGroup(menu);
+        int num = static_cast<int>(Base::UnitSystem::NumUnitSystemTypes);
+        for (int i = 0; i < num; i++) {
```
wwmayer, line 168: Maybe https://github.com/FreeCAD/FreeCAD/pull/8870#discussion_r1140461629

0penBrain: I thought to that too but that didn't really lighten the code actually.
I'm not sure to fully understand your comment there though as QActionGroup has no signal, so in any case we have to connect to each individual QAction signal.

wwmayer: > QActionGroup has no signal

It actually has two: hovered(QAction*) and triggered(QAction*)
It looks like the Qt docs for Qt6 isn't complete because they are not mentioned while the Qt5 docs show them. And both signals are declared in the QActionGroup for Qt5 and Qt6, so they definitely exist.

src/Gui/MainWindow.cpp
0penBrain: Done

src/Gui/MainWindow.cpp
0penBrain: Done

src/Gui/MainWindow.cpp
0penBrain: Done

src/Gui/MainWindow.cpp
0penBrain: Yep I was fooled by Qt6 doc.
I made the change in a last commit if you to check it matches your expectations.

src/Gui/MainWindow.cpp
```diff
@@ -150,11 +150,92 @@ class CustomMessageEvent : public QEvent
...
+        }
+    }
+};
+
 // -------------------------------------
 // Pimpl class
 struct MainWindowP
 {
-    QToolButton* sizeLabel;
+    QPushButton* sizeLabel;
```
PaddleStroke, line 238: Why not DimensionWidget here?

src/Gui/MainWindow.cpp
```diff
@@ -2250,3 +2293,4 @@ ActionStyleEvent::Style ActionStyleEvent::getType() const
 
 
 #include "moc_MainWindow.cpp"
+#include "MainWindow.moc"
```
PaddleStroke, line 2296: What is that? How is it related to this pr?

src/Gui/MainWindow.cpp
```diff
@@ -150,11 +150,92 @@ class CustomMessageEvent : public QEvent
     int _timeout;
 };
 
+class DimensionWidget : public QPushButton, WindowParameter
```
PaddleStroke, line 153: Maybe some documentation could be added before the class to explain briefly what is this widget.

PaddleStroke (commented):
Tested and works as expected. The button style looks more like it used to so that's good, it looks less intrusive. Though maybe users may not see it then.

src/Gui/MainWindow.cpp
0penBrain: Roughly, to help keep encapsulation under control. :)
The `MainWindow` class doesn't need to know more than it is using a QPushButton, then the rest is intended to stay internal stuff.
It has also the advantage that changing the `DimensionWidget` internals is "transparent" for the `MainWindow`.
However your interrogation seems legit and I'm not 100% sure to be right on this. @wwmayer may you please advise the right way ?

src/Gui/MainWindow.cpp
0penBrain: `DimensionWidget` uses the `Q_OBJECT` macro so Qt MOC will automatically generate (and need) this file so it's included here.

src/Gui/MainWindow.cpp
0penBrain: Yes, will add before merging.

src/Gui/MainWindow.cpp
wwmayer: I will have a look in one or two days as I am still working on the image plane/view migration to the core system.

src/Gui/MainWindow.cpp
wwmayer: As said we can leave it for now. Once we decide to move the class(es) we can do the same as Qt and create private headers.

src/Gui/MainWindow.cpp
PaddleStroke: Thanks for the explanation

src/Gui/MainWindow.cpp
0penBrain: OK, I will polish the PR and then let you merge if it's fine. ;)

src/Gui/MainWindow.cpp
0penBrain: @wwmayer I'm afraid you didn't get the point (maybe my mistake). :)
Simply the question is : is that OK for you to declare `sizeLabel` as `QPushButton` while it will actually be initialized as a `DimensionWidget` -- that inherits `QPushButton` -- taking as argument that `MainWindowP` doesn't need to know more than that ? Thx

src/Gui/MainWindow.cpp
0penBrain: Done

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8952">pipeline status</a> for feature branch [PR_8952](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8952). Pipeline [812842695](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/812842695) was triggered at [7040c5a](https://github.com/FreeCAD/FreeCAD/pull/8952/commits/7040c5a99e18c5e4f9f093f454b8109c8b4e785c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/MainWindow.cpp
0penBrain: @wwmayer May you tell when you'll review ? PR is ready.
If you have no time for this soon, I think I'll merge as is and then do another PR to take into account your further remarks if needed. Thx

0penBrain:
> does this make it look like the mouse model button?

Yes it does

src/Gui/MainWindow.cpp
wwmayer: > I'm afraid you didn't get the point (maybe my mistake). :)

Sorry, my bad. Using QPushButton here is totally fine.

src/Gui/MainWindow.cpp
wwmayer: > If you have no time for this soon, I think I'll merge as is and then do another PR to take into account your further remarks if needed.

The code looks good and using the QActionGroup this way is exactly what I had in mind. So, I think the code can be merged now.

Issue #8945, opened 2023-03-18
[Issue] Image workbench controls are scattered and hidden in a wb.


PaddleStroke:
### Forums discussion

https://forum.freecad.org/viewtopic.php?p=668562#p668562

### Issue description

**Currently :**
Image workbench :
- Image_Open: Open the img in a (bad) image viewer.
- Image_CreateImagePlane: Create a plane and put the selected img in it.
- Image_Scale: a Dialog opens to scale the image.
Core :
- File/Open + File/import : both call Image_Open.
- Double clicking on the Tree element triggers the transform enabling setting the image at correct position.

**Remarks :**
- Image_Open is useless. It only opens the picture in a bad image viewer. It is misleading and bad UI because what the user want is Image_CreateImagePlane.
- Having a workbench for only 1-2 tools is overkill + clutter the workbench selector.

**Suggestion :**
- Delete the Image workbench.
- Move the Image code to Core.
- The image viewer is removed. All tools from the toolbar are removed.
- File/Open + File/import + Paste with an image + drag drop an img, all call Image_CreateImagePlane.
- Double clicking the ImagePlane in tree + in 3DView opens a TaskBox containing the necessary for placement + transform + scale.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
> In the Tools menu we have the image creation function and there we can move the function to map an image onto a plane.

@wwmayer you mean `src\Tools\ImageTools\ImageConv` right ? So necessary files should be put in `src\tools\ImageTools\imagePlane` for instance?

But the command CmdCreateImagePlane, it should still be in one of the command tools in src/gui Like CommandStd.cpp?

Currently the Open command connects to the `Py::Object open` of the `ImageGui Module` in a generic way through : `StdCmdOpen` then  `Application::open` which then calls `Command::doCommand(Command::App, "%s.open(u\"%s\")", Module, unicodepath.c_str());`

So I'm unsure how to connect the CmdCreateImagePlane command in here. Should we add an `else if (File.hasExtension("png") || File.hasExtension("jpg") || ...)` in the `Application::open`?

wwmayer:
> you mean src\Tools\ImageTools\ImageConv right ?

No. I meant the function in FreeCAD: Tools > Save picture, that creates an image of the currently active view. I thought about adding a function to the Tools menu to also load an image file. Using the global open/import mechanism is an option too but it causes a lot of extra work.

[...]

Issue #8556, opened 2023-02-20
[Problem] Opening an image from the file menu is confusing and not useful


julia-m-stephenson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.20 (Release)

### Full version info

```shell
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.1.29410 (Git)
Build type: Release
Branch: releases/FreeCAD-0-20
Hash: f5d13554ecc7a456fb6e970568ae5c74ba727563
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/United Kingdom (en_GB)
Installed mods: 
  * Curves 0.6.5
  * fasteners 0.4.27
[/code]
```

### Subproject(s) affected?

None

### Issue description

When opening a PNG file the view seems to be constrained to a letter box. 

![image](https://user-images.githubusercontent.com/52968628/220046269-d2c4f180-2e01-4b42-862d-fb6f7320fc71.png)

If you zoom in and out using the scroll wheel it displays tiny.

![image](https://user-images.githubusercontent.com/52968628/220046639-d5ed7a6e-743e-4c95-9a05-82d07ad13a6c.png)

As a workaround I opened the file in inkscape and saved as SVG. This file was opened as a drawing and displayed correctly.

![image](https://user-images.githubusercontent.com/52968628/220047024-bff7a141-62a6-4d67-9cbe-fec6d3579c74.png)

Here is the PNG.
![DeckEyePlateInstructions](https://user-images.githubusercontent.com/52968628/220047133-67739b5a-7c90-4dea-a5b0-f46eef91c307.png)

Here is the SVG.
![DeckEyePlateInstructions](https://user-images.githubusercontent.com/52968628/220047135-417c5a11-f409-4df6-8f42-09fd836c3c1a.svg)

Thanks. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

WandererFan:
How are you "opening" the file?  What command? What workbench?

julia-m-stephenson:
> How are you "opening" the file? What command? What workbench?

File->Open

WandererFan:
I didn't know png was a valid import type.  I don't think this works in v0.21.

You could add the image to a TechDraw Page with [TechDraw_Image](https://wiki.freecad.org/TechDraw_Image).

[...]

PR #8955, opened 2023-03-18
Move image loading to core

wwmayer:

---


sliptonic:
Relates to Issue [8556](https://github.com/FreeCAD/FreeCAD/issues/8556)

sliptonic:
Unless you're anticipating something I don't know about, I would suggest NOT adding the menu item to the Tools menu.
All the other items in that menu feel the same - they launch a dialog.   This opens the file selector but then it is complete and so it doesn't behave like a 'tool'.  

Instead, it's an import function so triggering it from file->import is intuitive.  

Whenever we have multiple ways of triggering the exact same action, the user has to spend time understanding if they are the same thing or different.  I take my inspiration from the idea in zen of python;   "There should be one-- and preferably only one --obvious way to do it."

PaddleStroke:
Great, I will test it as soon as possible.

While you are at it, is it possible to also connect the Paste of image to this openImage function? So that pasting an image acts as opening one.

PaddleStroke:
On windows VS I have a build error : SoTextureCoordinate2 is undeclared.
Apparently the header is not in precompile. Moving `# include <Inventor/nodes/SoTextureCoordinate2.h>` out of the `#ifndef _PreComp_` solves the issue

wwmayer:
> Instead, it's an import function so triggering it from file->import is intuitive.

With File > Open/Import we already load the image into the 3D view and add it to the plane. So, it's not possible to use this mechanism to load an image into an image view.

wwmayer:
> On windows VS I have a build error : SoTextureCoordinate2 is undeclared.

Should be fixed now.

sliptonic:
> > Instead, it's an import function so triggering it from file->import is intuitive.
> 
> With File > Open/Import we already load the image into the 3D view and add it to the plane. So, it's not possible to use this mechanism to load an image into an image view.
> 
> > Whenever we have multiple ways of triggering the exact same action, the user has to spend time understanding if they are the same thing or different. I take my inspiration from the idea in zen of python; "There should be one-- and preferably only one --obvious way to do it."
> 
> These are two different actions. The File>Open/Import workflow adds the image to the 3D view. The command from the Tools menu creates a separate MDI image view. I added this because some people still wanted to have an image view.

I'm sorry.  I didn't see that.  I saw [your question asking if anyone used the image view](https://forum.freecad.org/viewtopic.php?t=76884#p668470).   I never saw anyone give a use-case for it.  Who was still using it?

PaddleStroke:
1- It compiles correctly now.
2- The tree icon is much better now.
3- drag&drop image + open + import all create an image plane directly. That's great.
4- Paste of a picture file works great.

Now what is left to do : 
a- Paste of raw pixel data doesn't work : If you open a picture in paint, then select only a part of the picture. Copy this part of picture. Then you can't paste this copied part of image.
b- A single taskbox to handle image settings. Triggers when double clicking the picture.
I posted a mockup in the issue : https://github.com/FreeCAD/FreeCAD/issues/8945
c- Back side of picture is black. People asked for back side to show reverse picture.
d- Remove image wb.

wwmayer:
> a- Paste of raw pixel data doesn't work : If you open a picture in paint, then select only a part of the picture. Copy this part of picture. Then you can't paste this copied part of image.

What is the paste target? An main window or the image view? The latter already works fine for me. If I open an image with GIMP, select and copy it or an area of it I can paste it into an open image view in FreeCAD. Or do you want that pasting into the main window should open an Image view?

> c- Back side of picture is black. People asked for back side to show reverse picture. They suggest using transparancy on the picture by default.

Nah. We can use transparency but we can also do it right. The idea is to implement two-side rendering: 9e18c54cc

> d- Remove image wb.

This will be the final step.

wwmayer:
The code is now ready for review. It implements the most important point but not every wish of everybody. This may be the task of future PRs.

sliptonic:
I definitely think the task panel should have a control to do the two-point scaling.  This was a core feature in image wb.

sliptonic:
Minor issue.  If you open the task panel and exit (with either ok or cancel) the imageplane node is left needing a recompute.

PaddleStroke:
Keep aspect ratio could be On by default I think.

Using image size in pixel doesn't make much sense in the context. Maybe displaying the picture size in the Unit System would make more sense.

sliptonic:
> The code is now ready for review.

I'm no judge of the code but the functionality looks great.  I'd still like to see a way to trigger the existing 2-point scaling but otherwise, it is a huge improvement. 

Any thoughts on unit testing?

src/App/ImagePlane.cpp
```diff
@@ -25,19 +25,36 @@
...
 
-using namespace Image;
 using namespace App;
 
-PROPERTY_SOURCE(Image::ImagePlane, App::GeoFeature)
+PROPERTY_SOURCE(App::ImagePlane, App::GeoFeature)
 
 
 ImagePlane::ImagePlane()
+    : XPixelsPerMeter{1000.0}
```
chennes, line 34: Where is this 1000 coming from? It feels sort of "magic-number-y".

src/Gui/CommandView.cpp
```diff
@@ -1969,6 +1971,51 @@ bool StdViewScreenShot::isActive()
...
+//===========================================================================
+// Std_ViewLoadImage
+//===========================================================================
+DEF_STD_CMD(StdViewLoadImage)
+
+StdViewLoadImage::StdViewLoadImage()
+  : Command("Std_ViewLoadImage")
+{
+    sGroup      = "Standard-View";
+    sMenuText   = QT_TR_NOOP("Load picture...");
```
chennes, line 1983: Any reason to use "picture" instead of "image"? It feels a little informal.

src/Gui/Camera.cpp
```diff
@@ -0,0 +1,200 @@
...
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
```
chennes, line 21: These files are still using the old-style copyright header, instead of the new one. Consider using the updated version in new files. Example here: https://github.com/FreeCAD/FreeCAD/blob/2d3953f65fff0db0bfa23d9877601a828ba24a7a/src/App/IndexedName.h#L1

src/Gui/FileHandler.h
```diff
@@ -1,50 +1,61 @@
...
+namespace App {
+class Document;
+}
+
+namespace Gui {
+
+class FileHandler
+{
+public:
+    FileHandler(const QString& filename);
```
chennes, line 38:
```suggestion
    explicit FileHandler(const QString& filename);
```

src/Gui/FreeCADGuiInit.py
```diff
@@ -291,6 +291,7 @@ def _SoGroup_init(self,*args):
 FreeCAD.addImportType("Inventor V2.1 (*.iv)","FreeCADGui")
 FreeCAD.addImportType("VRML V2.0 (*.wrl *.vrml *.wrz *.wrl.gz)","FreeCADGui")
 FreeCAD.addImportType("Python (*.py *.FCMacro *.FCScript)","FreeCADGui")
+FreeCAD.addImportType("Image formats (*.bmp *.jpg *.png *.xpm)","FreeCADGui")
```
chennes, line 294: Can we use Qt's list of support types here, instead of hard-coding it?

src/Gui/ImageView.cpp
```diff
@@ -0,0 +1,369 @@
...
+    imageLabel->resize(scaleFactor * imageLabel->pixmap(Qt::ReturnByValue).size());
+#else
+    imageLabel->resize(scaleFactor * imageLabel->pixmap()->size());
+#endif
+
+    adjustScrollBar(scrollArea->horizontalScrollBar(), factor);
+    adjustScrollBar(scrollArea->verticalScrollBar(), factor);
+}
+
+void ImageView::adjustScrollBar(QScrollBar *scrollBar, double factor)
```
chennes, line 108: Could be static.

src/Gui/ImageView.cpp
```diff
@@ -0,0 +1,369 @@
...
+{
+    scrollBar->setValue(int(factor * scrollBar->value()
+                            + ((factor - 1) * scrollBar->pageStep()/2)));
+}
+
+bool ImageView::canZoomIn() const
+{
+    return !isFitToWindow() && imageLabel->width() < 10000;
```
chennes, line 116: Magic number -- there are several around here (200 in canZoomOut, etc.). My clang-tidy complains about them, suggesting using named constants.

src/Gui/ImageView.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+    QAction* zoomOutAct = menu.addAction(tr("Zoom out"), this, &ImageView::zoomOut);
+    zoomOutAct->setEnabled(canZoomOut());
+
+    menu.exec(event->globalPos());
+}
+
+void ImageView::mousePressEvent(QMouseEvent* event)
+{
+    if (event->buttons() & Qt::MiddleButton) {
```
chennes, line 193: Technically this is an implicit conversion from an int to a bool.

src/Gui/ImageView.cpp
```diff
@@ -0,0 +1,369 @@
...
+    const QMimeData* data = event->mimeData();
+    if (data->hasUrls()) {
+        event->accept();
+    }
+    else {
+        event->ignore();
+    }
+}
+
+bool ImageView::isImageFormat(const QFileInfo& fileInfo) const
```
chennes, line 246: Could be static.

src/Gui/ImageView.cpp
```diff
@@ -0,0 +1,369 @@
...
+    auto it = std::find_if(supportedFormats.begin(), supportedFormats.end(), [suffix](const QByteArray& image) {
+        return (image == suffix);
+    });
+
+    return (it != supportedFormats.end());
+}
+
+void ImageView::loadImageFromUrl(const QList<QUrl>& urls)
+{
+    if (urls.isEmpty())
```
chennes, line 260: Consider adding braces.

src/Gui/ImageView.cpp
```diff
@@ -0,0 +1,369 @@
...
+        return;
+    QUrl url = urls.first();
```
chennes, line 262: Could be a const ref.

src/Gui/ImageView.cpp
```diff
@@ -0,0 +1,369 @@
...
+        setImage(image);
+    }
+}
+
+bool ImageView::canPasteImage() const
+{
+    return !imageFromClipboard().isNull();
+}
+
+QImage ImageView::imageFromClipboard() const
```
chennes, line 284: Could be static.

src/Gui/ImageView.cpp
```diff
@@ -0,0 +1,369 @@
...
+    QPixmap pixmap = QPixmap::fromImage(rawImage);
+    QRect rect = painter.viewport();
+    QSize size = pixmap.size();
+    size.scale(rect.size(), Qt::KeepAspectRatio);
+    painter.setViewport(rect.x(), rect.y(), size.width(), size.height());
+    painter.setWindow(pixmap.rect());
+    painter.drawPixmap(0, 0, pixmap);
+}
+
+bool ImageView::onMsg(const char* pMsg,const char**)
```
chennes, line 308: What is this second parameter? clang-tidy complains about its missing name.

src/Gui/ImageView.h
```diff
@@ -0,0 +1,99 @@
...
+class QUrl;
+
+namespace Gui {
+
+class GuiExport ImageView : public MDIView
+{
+    Q_OBJECT
+
+public:
+    ImageView(QWidget* parent);
```
chennes, line 41:
```suggestion
    explicit ImageView(QWidget* parent);
```

chennes (approved):
Code looks great -- a few suggestions below (some are just from clang-tidy, and nothing that's a big deal).

src/App/ImagePlane.cpp
wwmayer: Qt says "dots per meter" and for most of the tested images this value was 3750 or 3850.  Now I needed a sensible default value and because further down in the code the XPixelsPerMeter is divided by 1000 to get the size in mm it looked like a not totally wrong choice.

So, if for any reason the value couldn't be determined than we use `1000/1000` to convert px into mm.

But if you know how to get the "dots per meter" independent of the Qt API we can use that instead.

src/Gui/CommandView.cpp
wwmayer: Actually _image_ was the right term but the screen shot command above used the term _picture_ for consistency reasons. We can change then both commands to use the term _image_.

src/Gui/Camera.cpp
wwmayer: So, the copyright text is fine and only the SPDX identifier is missing?

src/Gui/FileHandler.h
wwmayer: lint told me that too but I forgot to add it.

src/Gui/FreeCADGuiInit.py
wwmayer: We needed to use PySide here but I don't want PySide to be a fix part of the core system. So, what we can do is: leave this code as is and after FreeCAD is fully initialized the further image types can be added.

src/Gui/ImageView.cpp
wwmayer: OK

src/Gui/ImageView.cpp
wwmayer: Yes.

src/Gui/ImageView.cpp
wwmayer: onMsg() can use the second parameter to report back a string to the calling instance. We have exactly one use case where we do this: View3DInventor::onMsg. It's used in StdViewDockUndockFullscreen::activated

wwmayer:
> I definitely think the task panel should have a control to do the two-point scaling. This was a core feature in image wb.

IMO the workflow is not very intuitive and error-prone. It took me quite some time to figure out how to use it correctly. The feature can be implemented but it must be easier to use.

> Any thoughts on unit testing?

There isn't much testable code. The only non-GUI class is ImagePlane.

PaddleStroke:
> IMO the workflow is not very intuitive and error-prone. It took me quite some time to figure out how to use it correctly. The feature can be implemented but it must be easier to use.

+1 the tool ui is 'Fancy' and could use an improvement.

sliptonic:
> > IMO the workflow is not very intuitive and error-prone. It took me quite some time to figure out how to use it correctly. The feature can be implemented but it must be easier to use.
> 
> +1 the tool ui is bad and could use an improvement.

Ok, then let's merge and immediately open an issue to redesign the tool.  I think that is the only remaining functionality keeping us from removing old image WB.  Scaling an image based on a depicted feature is a critical workflow.

src/Gui/Camera.cpp
chennes: Sorry, GitHub doesn't make it easy to link to the whole code block. You've still got the version that references a street address, "59 Temple Place, Suite 330, Boston MA" instead of the URL they suggest now.

src/App/ImagePlane.cpp
wwmayer: Nothing to do here right now. So I mark it as resolved.

src/Gui/Camera.cpp
wwmayer: Moved to new license header.

src/Gui/FreeCADGuiInit.py
wwmayer: Removed this line and added the image types in C++

src/Gui/ImageView.cpp
wwmayer: Fixed.

src/Gui/FileHandler.h
wwmayer: Fixed

src/Gui/ImageView.cpp
wwmayer: Added an identifier

src/Gui/ImageView.cpp
wwmayer: Switched to QFlags::testFlag

src/Gui/ImageView.cpp
wwmayer: Fixed

src/Gui/ImageView.cpp
wwmayer: Fixed

src/Gui/ImageView.cpp
wwmayer: Fixed

src/Gui/ImageView.cpp
wwmayer: Added a named argument

src/Gui/ImageView.h
wwmayer: Fixed

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8955">pipeline status</a> for feature branch [PR_8955](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8955). Pipeline [814655492](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/814655492) was triggered at [706b259](https://github.com/FreeCAD/FreeCAD/pull/8955/commits/706b259cee9385148a34e9ff233a3c388fa7645a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8955">pipeline status</a> for feature branch [PR_8955](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8955). Pipeline [814778213](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/814778213) was triggered at [649e5a6](https://github.com/FreeCAD/FreeCAD/pull/8955/commits/649e5a673ce68eb7a3e9586e67fb1c82b0b871d5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Roy-043:
I encounter a problem when importing an image if the path contains a single quote:
```
16:39:07  <string>(1)<class 'SyntaxError'>: ('unterminated string literal (detected at line 1)', ('<string>', 1, 86, "App.getDocument('Unnamed').ActiveObject.ImageFile = 'C:/Downloads/Foto's/IMG_8371.JPG'", 1, 86))
16:39:07  App.getDocument('Unnamed').ActiveObject.ImageFile = 'C:/Downloads/Foto's/IMG_8371.JPG'
```
```
Word size of FreeCAD: 64-bit
Version: 0.21.0.32785 (Git)
Build type: Release
Branch: master
Hash: 010e2ca37771fb9d89d513869baf520f89090204
Python 3.10.10, Qt 5.15.8, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: English/United States (en_US) [ OS: Dutch/Netherlands (nl_NL) ]
Installed mods:
```

chennes:
Oh, that's pretty bad, we're failing to sanitize that string someplace. Dangerous. Probably worth opening its own issue.

Issue #8940, opened 2023-03-18
[Core] Notifications are below the main window of FreeCAD


donovaly:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecad.org/viewtopic.php?p=668153#p668153

### Version

0.21

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.0.32357 (Git)
Build type: Release
Branch: master
Hash: d23b576c3361b7d593b8e65255a5f970e5954c39
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/United States (en_US) [ OS: German/Germany (de_DE) ]
```

### Subproject(s) affected?

Core

### Issue description

The message notification feature shows messages always below FreeCAD's main window.

Here is a screencast of a selection mistake. The error message appears below the main window:
![5YaSOFhAF0](https://user-images.githubusercontent.com/1828501/226076551-7589aeec-c90a-4277-97a9-db6c29571e4e.gif)

### Anything else?

@abdullahtahiriyo 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

donovaly:
@abdullahtahiriyo

abdullahtahiriyo:
I see.

In my PC this renders inside the frame. I will look into it.

donovaly:
Will be fixed with #8958

Issue #7882, opened 2022-11-26
[Problem] Toolbar customizations problem with RTL (right to left) locales 


mskf1383:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
OS: Debian GNU/Linux bookworm/sid (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.20.1.
Build type: Release
Python 3.10.7, Qt 5.15.4, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: English/United States (en_US) [ OS: Persian/Iran (fa_IR) ]
Installed mods: 
  * kicadStepUpMod.bak
  * pcb
  * kicadStepUpMod
```

### Subproject(s) affected?

Core

### Problem description

I customized toolbar to this:
![نماگرفت_۲۰۲۲۱۱۲۶_۱۴۳۱۰۵](https://user-images.githubusercontent.com/50908329/204085435-42591844-bce0-4a0e-8838-320342787a14.png)

But after close and reopening FreeCAD, It became this:
![نماگرفت_۲۰۲۲۱۱۲۶_۱۴۳۲۱۰](https://user-images.githubusercontent.com/50908329/204085465-a4c5d426-da5c-4003-a1ee-10a19b317bc1.png)

### Anything else?

My system language is Persian which is RTL (Right-To-Left). Also when I was in GNOME, FreeCAD was LTR anyway. But in KDE is true.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
Does the customization never work or is this a one time thing?

mskf1383:
It always happen. Also it is important that when I move a toolbox vertical, it save. but when I move it horizontal, It reset at next startup. I think this is a problem with RTL.

> Are other Qt applications working in this regard?

I don't know. QT apps I have are from KDE and don't let user to move toolboxes.

> Is this an installation from the debian repos?

Yes

luzpaz:
This is a bigger issue. This needs to be discussed on the UI/UX subforum as a dedicated thread for RTL UI/UX

luzpaz:
@mskf1383 can you open a forum thread to generate more exposure for this ticket, please?

mskf1383:
I created an account, but after signing in, there is no forum:

[...]

PR #8958, opened 2023-03-19
NotificationBox/NotificationArea: Restrict rendering area to main frame

abdullahtahiriyo:
========================================================================

The NotificationBox is extended to take the QRect in global coordinates. Then it will try to dimension the label within this area. If a fixed width is provided, that is enforced (take precedence).

The NotificationArea passes the QRect of the main window to the NotificationBox.

This is intended to fix:
https://github.com/FreeCAD/FreeCAD/issues/8940

### DIFF
--- a/src/Gui/NotificationArea.cpp
+++ b/src/Gui/NotificationArea.cpp
@@ -234,7 +234,7 @@ void NotificationAreaObserver::SendLog(const std::string& notifiername, const st
             .trimmed();// remove any leading and trailing whitespace character ('\n')
 
     // avoid processing empty strings
-    if(simplifiedstring.isEmpty())
+    if (simplifiedstring.isEmpty())
         return;
 
     if (level == Base::LogStyle::TranslatedNotification) {
@@ -827,7 +827,7 @@ void NotificationArea::pushNotification(const QString& notifiername, const QStri
     auto timer_thread = pImp->inhibitTimer.thread();
     auto current_thread = QThread::currentThread();
 
-    if(timer_thread == current_thread)
+    if (timer_thread == current_thread)
         pImp->inhibitTimer.start(pImp->inhibitNotificationTime);
 }
 
@@ -939,7 +939,8 @@ void NotificationArea::showInNotificationArea()
                     iconstr = QStringLiteral(":/icons/info.svg");
                 }
 
-                QString tmpmessage = convertFromPlainText(item->msg, Qt::WhiteSpaceMode::WhiteSpaceNormal);
+                QString tmpmessage =
+                    convertFromPlainText(item->msg, Qt::WhiteSpaceMode::WhiteSpaceNormal);
 
                 msgw +=
                     QString::fromLatin1(
@@ -990,11 +991,17 @@ void NotificationArea::showInNotificationArea()
 
         msgw += QString::fromLatin1("</table></p>");
 
+        // Calculate the main window QRect in global screen coordinates.
+        auto mainwindow = getMainWindow();
+        auto mainwindowrect = mainwindow->rect();
+        auto globalmainwindowrect =
+            QRect(mainwindow->mapToGlobal(mainwindowrect.topLeft()), mainwindowrect.size());
 
         NotificationBox::showText(this->mapToGlobal(QPoint()),
                                   msgw,
                                   pImp->notificationExpirationTime,
                                   pImp->minimumOnScreenTime,
+                                  globalmainwindowrect,
                                   pImp->notificationWidth);
     }
 }
--- a/src/Gui/NotificationBox.cpp
+++ b/src/Gui/NotificationBox.cpp
@@ -73,6 +73,8 @@ public:
     bool notificationLabelChanged(const QString& text);
     /// Place the notification at the given position
     void placeNotificationLabel(const QPoint& pos);
+    /// Set the windowrect defining an area to which the label should be constrained
+    void setTipRect(const QRect &restrictionarea);
 
     /// The instance
     static qobject_delete_later_unique_ptr<NotificationLabel> instance;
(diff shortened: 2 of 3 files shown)


abdullahtahiriyo:
@donovaly 

Give it a go to see if it fixes the Windows issue that the notification appears outside the main window frame. Let me know.

PaddleStroke:
I was about to report this behavior, great to see it's already being fixed :)

donovaly:
> Give it a go to see if it fixes the Windows issue that the notification appears outside the main window frame. Let me know.

Many thanks, this fixes the issue.

I realized another issue while testing:
The notification stay on top over all windows. So when I get a notification, then switch to another program, I still see the notification. This is confusing. The notification should be a subwindow of FC's main window not a windows of the OS.

xtemp09:
@abdullahtahiriyo, might I ask, why do you not want to pass the context object to [QTimer::singleShot()](https://doc.qt.io/qt-6/qtimer.html#singleShot-6) as Lint suggested? When the object is destroyed, the method will not be called, meaning that it's not necessary to check the item very existence.

What do you think about usage of multi-arg, as Lint also suggested?

PaddleStroke:
I agere with @donovaly , I have the messages still showing even when I go to another app, it's very intrusive

Another issue for me : All warnings and error show up as notifications. I don't know if it's my cmake config but I always have random warning triggering so it's also very intrusive.

xtemp09:
@abdullahtahiriyo, have you ever developed Right-to-left user interface? My question is related to the issue #7882.

Syres916:
I did report in the forum thread `all sorts of graphics 'left behind' from other apps as well as sub windows going behind FC and having to move windows around my desktop in order to retrieve them` but I have to say it did feel as though no-one was heeding it. If I'm 'sending' meshes from FreeCAD to Cura using a macro and I need to change one of the numerous settings then the Profile window in Cura becomes a right pain to keep getting access to it as FC appears to be pushing it to the bottom.
```
Word size of FreeCAD: 64-bit
Version: 0.21.0.32394 (Git)
Build type: Release
Branch: master
Hash: 431fafc5337976c43bf50aae7bc75b46ee8037c7
Python 3.8.10, Qt 5.12.8, Coin 4.0.0, Vtk 7.1.1, OCC 7.7.0
Installed mods: 
  * Silk 1.0.0
  * CurvedShapes 1.0.4
  * freecad.gears 1.0.0
  * Curves 0.6.8
  * PieMenu 1.2.4
  * Plot 2022.4.17
  * AirPlaneDesign 0.4.0
  * fasteners 0.4.54
  * sheetmetal 0.2.59

```

abdullahtahiriyo:
> I realized another issue while testing:
> The notification stay on top over all windows. So when I get a notification, then switch to another program, I still see the notification. This is confusing. The notification should be a subwindow of FC's main window not a windows of the OS.

The notification should disappear if you click after your configured minimum time. Does it?

The notification should disappear at any time if you click inside it. Does it?

Making the notification a subwindow of FC also has drawbacks. Because you may miss notifications. However, we can try to find solutions that work for each user, which may become preferences. It is a matter of reacting to the right combination of events. But because events do not work exactly the same over all OSs, we need to go one by one.

> Another issue for me : All warnings and error show up as notifications. I don't know if it's my cmake config but I always have random warning triggering so it's also very intrusive.

Please post in the forum the specific message. That warnings and errors show up in the NotificationArea is intentional. If errors and warnings are not meaningful, we need to filter them.

> I did report in the forum thread `all sorts of graphics 'left behind' from other apps as well as sub windows going behind FC and having to move windows around my desktop in order to retrieve them` but I have to say it did feel as though no-one was heeding it. If I'm 'sending' meshes from FreeCAD to Cura using a macro and I need to change one of the numerous settings then the Profile window in Cura becomes a right pain to keep getting access to it as FC appears to be pushing it to the bottom.

Where in the forum have you reported issues of NotificationArea (you may ping me in the forum)?

You are using Linux. Could you make a small video showing he issue so that I can try to understand it? 

I am using Linux (Ubuntu mate). I have FreeCAD open. I receive no notification at all from it. Perhaps, your experience helps me find a solution.

> I can only speak for myself but at the moment I get more messages/notifications/mentions than I can ever handle as a single persion.

Please open a ticket showing these messages. We need to know what they are.

The QTimer ones Werner pushed a fix for them. For common unnecessary QT messages Werner implemented a filter.

What is left?

abdullahtahiriyo:
For those who really can't leave with the notification area, you may disable it entirely in preferences. You will keep receiving your errors and warnings in the report view as before (the notification area did not take anything back from you). 

But even if you disable it, let us know that is annoying you exactly.

donovaly:
> The notification should disappear if you click after your configured minimum time. Does it?

Yes

> The notification should disappear at any time if you click inside it. Does it?

Yes

> Making the notification a subwindow of FC also has drawbacks. Because you may miss notifications. 

How? When I am e.g. viewing a PDF, how should I get in this time a FC note?

abdullahtahiriyo:
> How? When I am e.g. viewing a PDF, how should I get in this time a FC note?

Perhaps we can invert the question. If you are viewing a PDF, why is FC in your computer sending you a notification?

donovaly:
> Perhaps we can invert the question. If you are viewing a PDF, why is FC in your computer sending you a notification?

The issue is that I e.g. do things parallel. I start e.g. a simulation run, I get an error, but need to check e.g. what is written in the Elmer simulation tutorial to check why I get the error. So I change to Okular to read.
I learned now that I can click on the notification to get ris of it immediately but why not making the notification a subwindow? Then I could just change to Okular and doesn't have to click.

donovaly:
> So you do not care to miss any notification sent by FreeCAD during the time the FreeCAD window is not active.

Yes, when I leave FreeCAD, I have see he notification and once I left FC, there should not be notifications.

However, since I know now that I can click on a notification to make it vanish, this is OK for me. So please decide if it is too much troubles to make the notifications a subwindow or not.

What does @Syres916 and @PaddleStroke say?

abdullahtahiriyo:
https://github.com/FreeCAD/FreeCAD/pull/8977

Issue #8884, opened 2023-03-14
[Problem] C++ DXF importer never imports text (annotations)


kpemartin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Windows 7 Version 6.1 (Build 7601: SP 1)
Word size of FreeCAD: 64-bit
Version: 0.21.0.32303 (Git)
Build type: Debug
Branch: master
Hash: 45ee407645c147147656ec63cec9da5674516493
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/United States (en_US)
[/code]
```

### Subproject(s) affected?

None

### Problem description

The C++ DXF importer no longer imports text from the DXF file. This worked fine in the 0.20 importer, and still works fine with the legacy importer.

To reproduce, import the attached DXF file (contained in [Missing Text.zip](https://github.com/FreeCAD/FreeCAD/files/10970418/Missing.Text.zip)) with preferences: Do no use the legacy importer, import text and dimensions. Notice that the text in the file does not get imported. To compare, import with the legacy importer.

In fact, the C++ DXF importer ignores _any_ choices in the Preferences -> Import/Export -> DXF dialog; you are stuck with No texts and dimensions, do not group layers into blocks, scaling factor 1.0 (the only settings that seem to make any difference)

The problem is that this dialog places the preferences in `User parameter:BaseApp/Preferences/Mod/Draft` but the code loads the settings from `User parameter:BaseApp/Preferences/Mod/Draft`.

More specifically, the `ImpExpDxfRead` class ctor reads settings from `User parameter:BaseApp/Preferences/Mod/Draft` (finding the correct settings), but then method `readDXF` in class `Import::Module` in file AppImportPy.cpp forces settings to be loaded from `User parameter:BaseApp/Preferences/Mod/Import`. Because of the way `ImpExpDxfRead.setOptions` works, you get the class's default settings as modified by whatever is in `User parameter:BaseApp/Prefernces/Mod/Import` and the previous settings from `User parameter:BaseApp/Preferences/Mod/Draft` are discarded. Then because `User parameter:BaseApp/Preferences/Mod/Import` actually contains no settings, you always get the class's defaults which are: No text/dimension import, no grouping layers into blocks, scaling factor 1.0.

[...]

Issue #8806, opened 2023-03-09
[Problem] DXF import: "Get original colors from the DXF file" doesn't work in C++ importer


alafr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecad.org/viewtopic.php?p=665307#p665307

### Version

0.21 (Development)

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.0.31917 (Git)
Build type: Release
Branch: master
Hash: 4639283b459f9ff9093e289908cc003db8745b82
Python 3.10.9, Qt 5.15.6, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: French/France (fr_FR)
Installed mods: 
  * BIM
```

### Subproject(s) affected?

File formats

### Issue description

~~The option "Group layers into blocks" for DXF import does not work in weekly builds.~~ EDIT: see issue https://github.com/FreeCAD/FreeCAD/issues/8884 for this problem.

The option "Get original colors from the DXF file" doesn't work (nor in weekly nor in stable 0.20.2). All imported objects are black. Ideally when "Group layers into blocks" and "Get original colors from the DXF file" are active it should set on each created Part the layer's color.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

yorikvanhavre:
Both issues (group shapes + retrieve colors) confirmed. Looking into it...

alafr:
I searched the code history and although i have no C++ knowledge this commit looks suspicious for the broken "Group layers into blocks" option: https://github.com/FreeCAD/FreeCAD/commit/e1ebcef39e8f53ada00ad77ba800a0132aef61e8

```
>>> App.ParamGet("User parameter:BaseApp/Preferences/Mod/Draft").GetBool("groupLayers")
True
>>> App.ParamGet("User parameter:BaseApp/Preferences/Mod/Import").GetBool("groupLayers")
False
```

yorikvanhavre:
Yes it definitely looks like you found it. Now, we have two options: 

* Or we fix the Draft dialog at https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Draft/Resources/ui/preferences-dxf.ui but there is more to fix in the whole Draft code )legacy importer) to make sure we go fetch the preferences values at the new location,
* Or maybe an easier option:
[...]

PR #8959, opened 2023-03-19
Issues 8884, 8806. Change readDXF settings handling

kpemartin:
This PR is offered as a correction for issues #8884 and #8806

The change leaves loading the regular import options to the `ImpExpDXFRead` class object's ctor (where the code already existed). If the caller to `readDXF` passes a fourth argument specifying another options location, this will be given to the `ImpExpDXFRead` object to load from, but otherwise `readDXF` does not initiate any options loading.

The documentation string for readDXF has been updated to include the options_source argument, which was already in the code.

Because the removed code was causing the `ImpExpDXFRead` to load settings from a location where no settings had been stored, the settings from the dialog were being ignored, so no text was being imported, layers were never grouped into blocks, and the scaling factor was always 1.

### DIFF
--- a/src/Mod/Import/App/AppImportPy.cpp
+++ b/src/Mod/Import/App/AppImportPy.cpp
@@ -108,7 +108,7 @@ public:
             "export(list,string) -- Export a list of objects into a single file."
         );
          add_varargs_method("readDXF",&Module::readDXF,
-            "readDXF(filename,[document,ignore_errors]): Imports a DXF file into the given document. ignore_errors is True by default."
+            "readDXF(filename,[document,ignore_errors,option_source]): Imports a DXF file into the given document. ignore_errors is True by default."
         );
         add_varargs_method("writeDXFShape",&Module::writeDXFShape,
             "writeDXFShape([shape],filename [version,usePolyline,optionSource]): Exports Shape(s) to a DXF file."


yorikvanhavre:
I am not sure I would do this...
Originally, this C++ dxf importer/exporter was part of Draft. Then it has been moved to Import because it was more logical, made Draft again 100% python,  and also so it could be shared with other modules. TechDraw uses it already.
But since at the moment only Draft has preferences options implemented, all the pref settings have been kept within Draft.
But in the future this could (and maybe should, not sure) change. So I would think it makes sense to still default to read options from User parameter:BaseApp/Preferences/Mod/Import, even if it's currently unused. Be it only to not have to redo later something that was undone now...

kpemartin:
My feeling is somewhat opposite this.

There are several things to do to move the options to the new storage location, and the existing support code for the new location is the only part implemented. And a tiny part it is, changing one path string relating to the options storage location. If this were changed back to use the old location, in the future when a shift to the new location were done fully, it would be trivial to find and change back.

Right now it is just an impediment to other development, an unused stub of code that nevertheless needs to be maintained. If the change were harmless it would not be a problem, but right now the change breaks DXF import.

I would feel quite differently about this if there were a well laid-out plan in place to migrate the options location, with a list of where changes were required and what the changes might be, both in terms of code changes and UI changes. Would Import work differently depending on which workbench is active? What about "Open file"? Would there be separate options for Draft import and TechDraw import? Perhaps such a plan exists and I just haven't found it.

That being said, I do have a different fix for the problem, which changes `ImpExpDxfRead.setOptions` so that, when it is called, the settings that exist in the new options container will be loaded, but settings that do not exist in the new options container are left unchanged. The current behaviour is that settings that do not exist int the container revert to their hard-coded defaults. Because the current code is passing an empty container (`Mod/Import`), this forces all options to their hard-coded defaults. With this alternative proposed change, the options not in the passed container would stay at the settings read by the `ImpExpDxfRead` ctor from the `Mod/Draft` location.

yorikvanhavre:
Hmm you have a point there indeed. It may actually be good to force it that way: The import/exporter has no preferences on their own. It's the responsibility of the module using it to provide the UI and preferences. That turns it even more generic, actually...

Ideally then, the module should not provide a preferences path, but rather already all the needed values. So the UI/preferences could be arranged really as the module prefers. But that's a discussion for later :)

Ok I'm convinced. But before merging, I'd like to check to make sure TechDraw or OpenSCAD are not using or setting anything in User `parameter:BaseApp/Preferences/Mod/Import` 

@WandererFan please shout if I forgot something ;)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8959">pipeline status</a> for feature branch [PR_8959](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8959). Pipeline [815710882](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/815710882) was triggered at [62f2683](https://github.com/FreeCAD/FreeCAD/pull/8959/commits/62f26835801ce95b1932513af767bcb0a959f08c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8959">pipeline status</a> for feature branch [PR_8959](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8959). Pipeline [817382499](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817382499) was triggered at [20ab1f8](https://github.com/FreeCAD/FreeCAD/pull/8959/commits/20ab1f8eff62bf7fba7ce44de710d83f169a9497). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8959">pipeline status</a> for feature branch [PR_8959](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8959). Pipeline [817390153](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817390153) was triggered at [09f1458](https://github.com/FreeCAD/FreeCAD/pull/8959/commits/09f145819509a114e0667c59a7015f6752a374f2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8959">pipeline status</a> for feature branch [PR_8959](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8959). Pipeline [817392570](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817392570) was triggered at [cd9ca44](https://github.com/FreeCAD/FreeCAD/pull/8959/commits/cd9ca4496a6f33a5a6d756285c96f71186221735). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Thanks, merging

Issue #8091, opened 2022-12-27
[Issue] Error when entering dimensions as a fraction


CrashedAgain1859:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=3&t=74764

### Version

0.21 (Development)

### Full version info

```shell
OS: Ubuntu 22.04.1 LTS (MATE/mate)
Word size of FreeCAD: 64-bit
Version: 0.21.0.31391 (Git) AppImage
Build type: Release
Branch: master
Hash: 0ab2608a429f3a5ee9a63da9b66d13557655d7e4
Python 3.10.8, Qt 5.15.4, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: English/Canada (en_CA)
Installed mods: 
  * toSketch 1.0.1
  * ThreadProfile 1.84.0
  * A2plus 0.4.60g
  * freecad.gears 1.0.0
```

### Subproject(s) affected?

None

### Issue description

Sketcher: Entering imperial dimensions as a fraction (e.g. 1/4 in) or using a fractional entry to edit an existing dimension (eg. editing existing "0.123 in" by entering "/2" ("0.123 in /2")) gives an incorrect result. 
Part, Part Design, and other workbenches: Entering a fraction as a dimension in a data dialog box gives an incorrect calculation e.g. attempting to extrude a sketch to 1/2 inch length by entering "1/2" or "1/2 in" in the dialog box gives an extrusion of 0.2", not 0.500". 
It appears that it is now necessary to be very specific in making the entry, making sure to add a " symbol in just the right place in the entry.....for example entering 1"/2 works but 1/2" does not. 

At best this is just a workaround. 
This used to work but now does not.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
most likely introduced by #7124

webmite:
It appears [#7124](https://github.com/FreeCAD/FreeCAD/pull/7124) affects far more than just sketcher.


On 'part workbench' adding a cube and attempting to modify Length dimension with unit math in input text box fails spectacularly.

CrashedAgain1859:
Yes it does. There is some discussion of this in the forum conversation
referenced.

adrianinsaval:
it affects most gui spinboxes

> attempting to modify Length dimension with unit math in input text box fails spectacularly.

[...]

PR #8982, opened 2023-03-22
[GUI] QuantitySpinBox: more intuitive parsing

ConnorWorrell:
[GUI] QuantitySpinBox: more intuitive parsing before handoff to expression parser.
More robust parentheses, multiplication, division, exponent handling.

---

Contains modifications discussed here: https://github.com/FreeCAD/FreeCAD/issues/8091?notification_referrer_id=NT_kwDOATpYjLM1MTg0MzU2Mzk0OjIwNjAwOTcy#issuecomment-1457365758

This modifies input text to give more intuitive results, by extracting the units and moving them to give the intuitive result. 
Eg: 1/2in -> 1/2*(1in)

Also improves handling of parentheses, multiplication, division, exponentiation. Groups things between + and -, assumes units apply to the entire group. If multiple units are present in one group, no changes are made to that group.
Eg: 1/(2\*2mm) + 1/2in + 2^(2+1thou) -> 1/(2\*2)\*(1mm) + 1/2\*(1in) + 2^(2+1)\*(1thou)
Eg: 1mm/2in + 1/2mm-> 1mm/2in (no change) + 1/2\*(1mm)

Ignores from the unit selection operations listed in src/Base/QuantityParser.y (hard coded).
Eg: cos(1mm) -> cos(1)*(1mm)

### DIFF
--- a/src/Gui/QuantitySpinBox.cpp
+++ b/src/Gui/QuantitySpinBox.cpp
@@ -163,7 +163,7 @@ public:
         double value = min;
         bool ok = false;
 
-        QChar space = QLatin1Char(' '), period = QLatin1Char ('.'), plus = QLatin1Char('+'), minus = QLatin1Char('-'), division = QLatin1Char('/'), multiply = QLatin1Char('*'), exp = QLatin1Char('^'), par = QLatin1Char('(');
+        QChar plus = QLatin1Char('+'), minus = QLatin1Char('-');
 
         if (locale.negativeSign() != minus)
             copy.replace(locale.negativeSign(), minus);
@@ -171,34 +171,42 @@ public:
             copy.replace(locale.positiveSign(), plus);
 
         //Prep for expression parser
-        unsigned int length = copy.length(), shift;
-        bool noUnit = false;
-        for (unsigned int pos=0; pos < length; pos++) {
-            QChar num = copy[pos];
-            if (num == division || num == multiply || num == exp || num == par){
-                break; // Stop if anything odd happens
-            }
-            else if (num.isNumber()) {
-                noUnit = true;
-            }
-            else if (num != space && num != period && num != plus && num != minus) {
-                noUnit = false;
+        //This regex matches chunks between +,-,$,^ accounting for matching parenthesis.
+        QRegularExpression chunkRe(QString::fromUtf8("(?<=^|[\\+\\-])((\\((?>[^()]|(?2))*\\))|[^\\+\\-\n])*(?=$|[\\+\\-])"));
+        QRegularExpressionMatchIterator expressionChunk = chunkRe.globalMatch(copy);
+        unsigned int lengthOffset = 0;
+        while (expressionChunk.hasNext()) {
+            QRegularExpressionMatch matchChunk = expressionChunk.next();
+            QString origionalChunk = matchChunk.captured(0);
+            QString copyChunk = origionalChunk;
+            std::reverse(copyChunk.begin(), copyChunk.end());
+
+            //Reused regex patterns
+            static const std::string regexUnits = "sAV|VC|lim|nim|im|hpm|[mf]?bl|°|ged|dar|nog|″|′|rroT[uµm]?|K[uµm]?|A[mkM]?|F[pnuµm]?|C|S[uµmkM]?|zH[kMGT]?|H[nuµm]?|mhO[kM]?|J[mk]?|Ve[kM]?|V[mk]?|hWk|sW|lack?|N[mkM]?|g[uµmk]?|lm?|(?<=\\b|[^a-zA-Z])m[nuµmcdk]?|uoht|ni|\"|'|dy|dc|bW|T|t|zo|ts|twc|Wk?|aP[kMG]?|is[pk]|h|G|M|tfc|tfqs|tf|s";
+            static const std::string regexUnitlessFunctions = "soca|nisa|2nata|nata|hsoc|hnis|hnat|soc|nat|nis|pxe|gol|01gol";
+            static const std::string regexConstants = "e|ip|lomm|lom";
+            static const std::string regexNumber = "\\d+\\s*\\.?\\s*\\d*|\\.\\s*\\d+";
+
+            //Find units and replace 1/2mm -> 1/2*(1mm), 1^2mm -> 1^2*(1mm)
+            copyChunk.replace(QRegularExpression(QString::fromStdString("("+regexUnits+")(\\)|(?:\\*|(?:\\)(?:(?:\\s*(?:"+regexConstants+"|\\)(?:[^()]|(?R))*\\((?:"+regexUnitlessFunctions+")|"+regexNumber+"))|(?R))*\\(|(?:\\s*(?:"+regexConstants+"|\\)(?:[^()]|(?R))*\\((?:"+regexUnitlessFunctions+")|"+regexNumber+"))))+[\\/\\^](?!("+regexUnits+")))")), QString::fromUtf8(")\\11(*\\2"));
+
+            //Add default units to string if none are present
+            QRegularExpression unitsRe(QString::fromStdString("(?<=\\b|[^a-zA-Z])("+regexUnits+")(?=\\b|[^a-zA-Z])|°|″|′|\"|'|\\p{L}\\.\\p{L}|\\[\\p{L}"));
+
+            QRegularExpressionMatch match = unitsRe.match(copyChunk);
+            std::reverse(copyChunk.begin(), copyChunk.end());
+            if (!match.hasMatch() && !copyChunk.isEmpty()){ //If no units are found, use default units
+                copyChunk.append(QString::fromUtf8("*(1")+unitStr+QString::fromUtf8(")")); // Add units to the end of chunk *(1unit)
             }
 
-            if (noUnit
-                && (num == plus ||                       // 1+  -> 1mm+
-                    num == minus ||                      // 1-  -> 1mm-
-                    (pos == length - 1 && (pos += 1)))) {// 1EOL-> 1mmEOL
-                copy.insert(pos, unitStr);
-                pos += shift = unitStr.length();
-                length += shift;
-                noUnit = false;
-            }
+            copy.replace(matchChunk.capturedStart() + lengthOffset,
+                    matchChunk.capturedEnd() - matchChunk.capturedStart(), copyChunk);
+            lengthOffset += copyChunk.length() - origionalChunk.length();
         }
 
         ok = parseString(copy, res, value, path);
 
-        // If result has not unit: add default unit
+        // If result does not have unit: add default unit
         if (res.getUnit().isEmpty()){
             res.setUnit(unit);
         }


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982">pipeline status</a> for feature branch [PR_8982](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982). Pipeline [813991077](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/813991077) was triggered at [46a0107](https://github.com/FreeCAD/FreeCAD/pull/8982/commits/46a0107c7288cc28e509ae53bde6e4232dbf4350). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
Cos(1mm) should not be accepted as valid IMO, that is totally wrong syntax

ConnorWorrell:
The assumption is the user had intended to put the units at the end, but put them in the wrong spot, because cos(1mm) and 1/(2mm) never make sense.

As far as I can tell (And I might be ignorant) the quantity spin box takes inputs that are a distance, angle, or unitless, so this should be a safe assumption.

adrianinsaval:
Also how does it deal with references to other parameters? If I input `Parameters.Area/(2*5mm)` what does it interpret? Is it aware that the numerator might have units?

donovaly:
> Doesn't `cos` take only an angle as parameter?

I think it takes any quantity as argument, but don't know.

>Eitherway, this change is proposing to interpret `cos(1mm)` as `cos(1)*1mm`, is that also what you have in mind?

Yes, and I think this is sensible.

donovaly:
> for me it does not seem sensible to take units out of a parenthesis, in any situation.

I agree to this as I wrote.

Concerning the angular functions, you convinced me that their behavior should also not be changed, meaning cos(1mm) should not be accepted at all. However as I read it, this is at the moment possible?

Anyway, @ConnorWorrell how to you want to proceed?

adrianinsaval:
> However as I read it, this is at the moment possible?

you mean in master? no it's rejected

ConnorWorrell:
> AFAIK it can be any arbitrary unit supported by FreeCAD, say kg*m/s^2 

This is correct, I was trying to say that I have only seen quantity spinbox used for sketcher dimensions/constrains to input distances or angles or unitless (snells law) numbers, never more complex. If its used in areas with more complex units then this becomes more difficult.

> > 1/(2*2mm)
> 
> Changing this is in my opinion a mistake. The user explicitly put it in parentheses, so he might like to get an inverse length.

I agree, but pulling out the units here doesn't change anything. Its evaluated to 1/4 mm^-1 by the expression parser. I'm pretty sure the units are not what's expected, so it sets the units to what ever was last used, .25mm giving the same result.

I still think this is the easiest way to intuitively deal with this.  For example, nested parens: 1/(1/(1/2mm)) -> 1/(1/(2/2))\*(1mm). Did the user intend for 1/2mm to be .5 mm^-1? Probably not. And with the assumption that the only acceptable outputs are distances and angles, we should be fairly safe in assuming they meant millimeters. cos(1mm) and 1/(2\*2mm) are collateral in this assumption.

>what if I wrote `acos(Spreadsheet.something\*1mm)` and `something` had units of `mm^-1`? Is this going to detect that or will it move the units I explicitly placed inside a parenthesis?

In general any text that has more than 1 unique string of letters, (excluding cos, pi, and the like) won't be modified. This has 3 unique strings so it wont be modified. I feel good about handling it this way, previously you (@adrianinsaval ) said https://github.com/FreeCAD/FreeCAD/pull/7124#issuecomment-1172797321 variable inputs should be disabled outright for this type of input.

Currently the only time I see this failing is if something like "Spreadsheet.Spreadsheet" is entered, intending to pull a value of alias Spreadsheet. This is modified to ".*(1Spreadsheet)" which fails to parse. I could fix this by hard coding in all available units.

ConnorWorrell:
Ignoring units that occur inside of parenthesis is trivial, but I still think there is more utility in not.

I'm also thinking hard coding the units to match against is probably the better solution than grabbing the first string. This could help with some edge case senarios.

donovaly:
> > > 1/(2*2mm)
> 
> I agree, but pulling out the units here doesn't change anything. Its evaluated to 1/4 mm^-1 by the expression parser. I'm pretty sure the units are not what's expected, so it sets the units to what ever was last used

You cannot make assumptions what the user might want. In this case he explicitly puts the unit in the parentheses so he gets an inverse length as result. If the user does not want, this, he can move the unit out of the parentheses, but not the parser. One of the base rules is to respect the user's will, therefore we try to avoid assumptions what he might will, but do what he entered.

> I still think this is the easiest way to intuitively deal with this. For example, nested parens: 1/(1/(1/2mm)) -> 1/(1/(2/2))*(1mm). Did the user intend for 1/2mm to be .5 mm^-1? Probably not.

How can you know? The setting of the parentheses is up to the user.

adrianinsaval:
> and previously you (@adrianinsaval ) said [#7124 (comment)](https://github.com/FreeCAD/FreeCAD/pull/7124#issuecomment-1172797321) variable inputs should be disabled outright for this type of input.

Yes, but they aren't disabled!

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982">pipeline status</a> for feature branch [PR_8982](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982). Pipeline [817593916](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817593916) was triggered at [5028431](https://github.com/FreeCAD/FreeCAD/pull/8982/commits/50284318371e570dcd173b4dc49a00086268cfb8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

ConnorWorrell:
I have reworked this, I think these outputs are much more inline with what ya'll are expecting:

It now replaces the units with *(1units) in place. And its also more picky about what it replaces, targeting division (1/2in), exponents (1^2in), and parentheses with units (1+1)in

| Input                           | Output                                |
|---------------------------------|---------------------------------------|
| 1/2                             | 1/2*(1in)                             |
| 1/2 in                          | 1/2 *(1in)                            |
| 1/2"                            | 1/2"                                  |
| 1 + 1 in                        | 1 *(1in)+ 1 in                        |
| 1/(1in+1)                       | 1/(1in+1)                             |
| 3 in + 6 mm                     | 3 in + 6 mm                           |
| 1/(2*2mm) + 1/2in + 2^(2+1thou) | 1/(2\*2mm) + 1/2\*(1in) + 2^(2+1thou) |
| 1mm/2in + 1/2mm                 | 1mm/2in + 1/2*(1mm)                   |
| cos(1mm)                        | cos(1mm)                              |
| 1/(2mm)                         | 1/(2mm)                               |
| Parameters.Area/(2*5mm)         | Parameters.Area/(2*5mm)               |
| acos(Spreadsheet.something*1mm) | acos(Spreadsheet.something*1mm)       |
| cos(1) in                       | cos(1) *(1in)                         |
| 1/(1/(1/2mm))                   | 1/(1/(1/2*(1mm)))                     |
| ( 1 + 1 ) kg                    | ( 1 + 1 ) *(1kg)                      |
| 1^2in                           | 1^2*(1in)                             |

Thoughts?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982">pipeline status</a> for feature branch [PR_8982](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982). Pipeline [817609779](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817609779) was triggered at [cef4cae](https://github.com/FreeCAD/FreeCAD/pull/8982/commits/cef4cae4796f1a3c579823f768dde80a7f125d81). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Looks good.
Can you  please rebase and force-push to trigger another CI run (I made a merge mistake)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982">pipeline status</a> for feature branch [PR_8982](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982). Pipeline [817960862](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817960862) was triggered at [63c15a9](https://github.com/FreeCAD/FreeCAD/pull/8982/commits/63c15a9be3d955298fce53fb995270f8eb0f8ea2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982">pipeline status</a> for feature branch [PR_8982](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982). Pipeline [817962238](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817962238) was triggered at [cc1e4d8](https://github.com/FreeCAD/FreeCAD/pull/8982/commits/cc1e4d8fea8fb4c55b3e740f97e29a738aa01c49). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
@chennes , @wwmayer could you please also have a look?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982">pipeline status</a> for feature branch [PR_8982](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982). Pipeline [818327224](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/818327224) was triggered at [d2fd2b2](https://github.com/FreeCAD/FreeCAD/pull/8982/commits/d2fd2b214b9c243f986ced1326d82810f249b0e1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/QuantitySpinBox.cpp
```diff
@@ -163,42 +163,41 @@ class QuantitySpinBoxPrivate
...
+            QRegularExpressionMatch matchChunk = expressionChunk.next();
+            QString origionalChunk = matchChunk.captured(0);
+            QString copyChunk = origionalChunk;
+
+            //Find units and replace
+            copyChunk.replace(QRegularExpression(QString::fromUtf8("((?:(?:\\((?>(?:[\\+\\*\\-\\^]?\\ *(?:\\d+\\ *\\.?\\ *\\d*|\\.\\ *\\d+)|e|pi|acos|asin|atan2|atan|cosh|sinh|tanh|cos|tan|sin|exp|log|log10)+|(?R))*\\)|(?:[\\+\\*\\-\\^]?\\ *(?:\\d+\\ *\\.?\\ *\\d*|\\.\\ *\\d+)|e|pi|acos|asin|atan2|atan|cosh|sinh|tanh|cos|tan|sin|exp|log|log10)+)+\\ *[\\/\\^]\\ *(?:\\((?>(?:[\\+\\*\\-\\^]?\\ *(?:\\d+\\ *\\.?\\ *\\d*|\\.\\ *\\d+)|e|pi|acos|asin|atan2|atan|cosh|sinh|tanh|cos|tan|sin|exp|log|log10)+|(?R))*\\)|(?:[\\+\\*\\-\\^]?\\ *(?:\\d+\\ *\\.?\\ *\\d*|\\.\\ *\\d+)|e|pi|acos|asin|atan2|atan|cosh|sinh|tanh|cos|tan|sin|exp|log|log10)+)+)|\\)\\ *)(VAs|CV|mil|min|mi|mph|lb[mf]?|°|deg|rad|gon|″|′|[uµm]?Torr|[uµm]?K|[mkM]?A|[pnuµm]?F|C|[uµmkM]?S|[kMGT]?Hz|[nuµm]?H|[mk]?V|[kM]?Ohm|[mk]?J|[kM]?eV|kWh|Ws|k?cal|[mkM]?N|[uµmk]?g|m?l|[nuµmcdk]?m|thou|in|\"|'|yd|cd|Wb|T|t|oz|st|cwt|k?W|[kMG]?Pa|[pk]si|h|G|M|cft|sqft|ft|s)")), QString::fromUtf8("\\1*(1\\2)"));
```
chennes, line 184: This regex is a bit of a monster -- between the extreme line length and the code duplication, I am really hoping that something a little more understandable can be created. Do we really have to do the whole thing in one go?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982">pipeline status</a> for feature branch [PR_8982](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982). Pipeline [819739571](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/819739571) was triggered at [dec4a22](https://github.com/FreeCAD/FreeCAD/pull/8982/commits/dec4a22b31bd846aa23d85513322d6de4dcb2653). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/QuantitySpinBox.cpp
ConnorWorrell: Most of the length ends up handling the units, and functions, pushed an updated to break these out and generate the regex on the fly, this reduces the length by quite a bit.

If I run the regex from right to left it also reduces complexity because the positive look ahead  can replace ~half of the regex. Using a look ahead also ends up being slightly more robust. (Although it gets a bit less readable with everything being backwards)

The regex is fairly straightforward now, so I feel good about this:
(Parentheses) or
((not a unit) + (divide or exponent) + (numbers inside parentheses or numbers outside parentheses))
+ units

src/Gui/QuantitySpinBox.cpp
```diff
@@ -163,42 +163,50 @@ class QuantitySpinBoxPrivate
...
+            QRegularExpressionMatch matchChunk = expressionChunk.next();
+            QString origionalChunk = matchChunk.captured(0);
+            QString copyChunk = origionalChunk;
+            std::reverse(copyChunk.begin(), copyChunk.end());
+
+            //Reused regex patterns
+            static std::string regexUnits = "sAV|VC|lim|nim|im|hpm|[mf]?bl|°|ged|dar|nog|″|′|rroT[uµm]?|K[uµm]?|A[mkM]?|F[pnuµm]?|C|S[uµmkM]?|zH[kMGT]?|H[nuµm]?|mhO[kM]?|J[mk]?|Ve[kM]?|V[mk]?|hWk|sW|lack?|N[mkM]?|g[uµmk]?|lm?|(?<=\\b|[^a-zA-Z])m[nuµmcdk]?|uoht|ni|\"|'|dy|dc|bW|T|t|zo|ts|twc|Wk?|aP[kMG]?|is[pk]|h|G|M|tfc|tfqs|tf|s";
+            static std::string regexUnitlessFunctions = "soca|nisa|2nata|nata|hsoc|hnis|hnat|soc|nat|nis|pxe|gol|01gol";
+            static std::string regexConstants = "e|ip|lomm|lom";
+            static std::string regexNumber = "\\d+\\s*\\.?\\s*\\d*|\\.\\s*\\d+";
```
chennes, line 188: This helps the readability quite a lot, I think -- nice work. There's probably still some cleanup that could be done, but I think that can be done in a later PR. These can all be `const`, right?

ConnorWorrell: Ah yes, these should be static const.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982">pipeline status</a> for feature branch [PR_8982](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8982). Pipeline [819784702](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/819784702) was triggered at [98b1cbe](https://github.com/FreeCAD/FreeCAD/pull/8982/commits/98b1cbe3e5414f27f7f22a9f0f7172fef09aeae2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

vernzimm:
@ConnorWorrell first off, yeah, that regex was awesome, and don't let anyone tell you otherwise 🤣 

I pulled latest weekly (0.21.32785) but this doesn't seem to be functional yet? It has been killing me since this change took place, but I didn't bother to look it up at first thinking it would be fixed eventually.

Probably unpopular opinion, but are there existing libraries/modules that do this without having to reinvent the wheel?

Could activation of this code automatically pop a formula pane open?
20230412 FreeCAD numberic entry to formula box

1. Type formula style entry in regular box (is that QuantitySpinBox?)
     1.1 "formula style" meaning whatever triggers @ConnorWorrell's code
     1.2 Alternately, use something like the Excel "=" before an entry to imply a formula?
     1.3 Lose the weirdly placed (fx) button in that case, or in any case?
2. A transient formula entry pane opens above
     2.1 Pane is initially inactive or does it immediately pull mouse/cursor focus, etc?
     2.2 Dismiss with esc key etc?
     2.3 "Enter" or "Tab" to accept the result and close the pane?
3. Include a button to pop the pane out into the full formula box window.

ConnorWorrell:
> Probably unpopular opinion, but are there existing libraries/modules that do this without having to reinvent the wheel?

There might be, I took a brief look around before I started, but didn't see anything, although I don't do much with cpp so I could total have overlooked something obvious.

> Could activation of this code automatically pop a formula pane open? 20230412 FreeCAD numberic entry to formula box
> 
> 1. Type formula style entry in regular box (is that QuantitySpinBox?)
>    1.1 "formula style" meaning whatever triggers @ConnorWorrell's code
>    1.2 Alternately, use something like the Excel "=" before an entry to imply a formula?
>    1.3 Lose the weirdly placed (fx) button in that case, or in any case?
> 2. A transient formula entry pane opens above
>    2.1 Pane is initially inactive or does it immediately pull mouse/cursor focus, etc?
>    2.2 Dismiss with esc key etc?
>    2.3 "Enter" or "Tab" to accept the result and close the pane?
> 3. Include a button to pop the pane out into the full formula box window.

I might be misunderstanding what your saying here, but this seems outside of the scope of this PR. 

The goal of this code is to cleanup inputs that aren't a formula before they get evaluated to give the results one would expect after evaluation from the expression parser. This PR makes this cleanup more robust and fixes the 1/2in regression. The expression stuff is a different system entirely, and after the small changes made by this code it is also run through the expression parser to arrive at the final number.

Side tangent:

- Currently when editing a constraint in sketcher if "=" is pressed, it opens the formula editor box, similar to how you've noted. But outside of sketcher this behavior is less reliable. For example, quantity spin box also covers some inputs in the properties panel, but pressing equals on these won't open the formula editor box, so the fx button is required. 

  What your suggesting is interesting, and I would love for the "=" behavior to be consistent outside of sketcher, but it seems like this behavior is already implemented.
  
  fds

vernzimm:
@ConnorWorrell

> I might be misunderstanding what your saying here, but this seems outside of the scope of this PR.
> 
> The goal of this code is to cleanup inputs before they get evaluated to give the results one would expect after evaluation from the expression parser.

Agreed, sorry. I was spit balling the larger concept of when should an entry be considered a formula. I'm wondering if any/all entries that require interpretation could just dump directly into the formula entry box, rather than having the three types: 

1. explicit entry (0.500)
2. expression parsed entry (1.000/2)
3. "true" formula entry (formula input -> 1.000/2)

(note, this 1,2,3 is unrelated to my previous comment)

In my mind, 2 and 3 are the same thing, but 2 requires special handling with code, and 3 requires additional action by user. If 2 transitioned seamlessly and automagically to 3, then essentially there would only ever be 1 or 3.

Let me mentally take a step back here though. Maybe what you're working on applies to formulas as well? I was assuming that formula entry functioned differently and was handled by something else, so it seemed redundant to have 2 and 3. If what you're working on applies to both situations then it would narrow the scope of what I'm saying significantly. In that case it's mostly a UI/UX issue to make the formula entry seamless.

> 
> Currently when editing a constraint in sketcher if "=" is pressed, it opens the formula editor box, similar to how you've noted.
>   What your suggesting is interesting, and I would love for the "=" behavior to be consistent outside of sketcher, but if I'm understanding what your asking it seems like this behavior is already implemented.

🤦 this is not the first time I've learned that something I wished for already existed and I had never actually thought to try it first 🤣, sorry.

>   The fx button could also be considered more intuitive as an indicator that a formula can be entered. There are a few areas where the fx button can be used to define an expression where you wouldn't expect to be able to. ie: The Label (name) property of a feature can be defined in an expression.

I like the idea of there being some kind of indicator, but I've always found the button to look bad and the separate formula box window to be too much effort. It would be more intuitive and natural for the entry box to feel "live". Something like I was explaining in prev comment, or like the variable/function name auto-complete in an IDE.

TL:DR yes, largely unrelated to the PR, and me imagining trivial QOL changes that are way under the radar of anyone working on the application at the moment 😞, but thanks for humoring me. I appreciate what you're doing because it was sorely missed.

Issue #8878, opened 2023-03-14
[Problem] TechDraw Dimensions are corrupted on recompute


WandererFan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
any version.
```

### Subproject(s) affected?

Techdraw

### Problem description

TechDraw dimensions use the traditional geometry + index name ("EDGE5") to identify model elements.  These names can change when the base model changes (the "Topological Naming Problem") and also when the OCC Hidden Line Removal algorithm is executed.  

When these names change (more precisely, when "EDGE5" now refers to a different edge),  the dimension may become corrupt, or worse, report an incorrect value without warning.

A typical user reaction to this problem: https://forum.freecad.org/viewtopic.php?t=75364

This problem can be mitigated to some extent by using references to the 3d model instead of to lines on the drawing.  This is easier than it once was, but is still not as convenient as selecting an edge in the drawing. 

Ideally, each edge produced by the Hidden Line Removal algorithm would carry a persistent link to the original edge in the 3d model.  This would require major changes to the OCC algorithm.

An alternative would be for the dimension to maintain a copy of the original geometry and report when the saved copy does not match the current state of the model.  In this alternative, it would also be possible to "autocorrect" the dimension. 

### Anything else?

Forum discussion: https://forum.freecad.org/viewtopic.php?t=76815

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

WandererFan:
There are two aspects of the problem:
- first, we need to detect that the problem has occured and inform the user.
- second, the system should be able to recover from the broken state without user intervention.

In order to solve the first aspect, we need to know that the geometry being referenced by the dimension is not the geometry that was originally referenced.  This requires saving a copy of the referenced geometry for comparison.  Geometry is a relatively lightweight object, so saving a copy of an edge or a few vertices is not a problem.

Comparing geometry is a difficult undertaking - there is no "==" operator for geometry.  In a complete solution, we would need to consider changes in scale or positioning as well as actual changes in the geometry, such as a line becoming longer or a radius getting bigger. The problem has some similarities with feature detection in computer vision applications.

The general flow of the comparison process is: 
    geometric shapes > extract characteristics > compare characteristics > take action

[...]

PR #8989, opened 2023-03-22
[TD]Dimension Reference Error Detection and Correction Ph 1 [#8878]

WandererFan:
This PR implements the first phase of a solution for issue #8878.

This phase detects changes in referenced geometry and corrects the reference if an exact match is available.

Phase 2 will address partial matches.

---

### DIFF
--- a/src/Mod/TechDraw/App/AppTechDrawPy.cpp
+++ b/src/Mod/TechDraw/App/AppTechDrawPy.cpp
@@ -87,6 +87,7 @@ using Import::ImpExpDxfWrite;
 using TechDraw::ProjectionAlgos;
 
 using namespace std;
+using namespace Part;
 
 namespace TechDraw {
 
@@ -271,7 +272,7 @@ private:
         }
         else {
             for (auto& w : sortedWires) {
-                PyObject* wire = new TopoShapeWirePy(new TopoShape(w));
+                PyObject* wire = new TopoShapeWirePy(new Part::TopoShape(w));
                 result.append(Py::asObject(wire));
             }
         }
--- a/src/Mod/TechDraw/App/CMakeLists.txt
+++ b/src/Mod/TechDraw/App/CMakeLists.txt
@@ -24,6 +24,7 @@ set(TechDrawLIBS
     Part
     Spreadsheet
     Import
+    PartDesign
 )
 
 if(FREECAD_QT_MAJOR_VERSION EQUAL 5)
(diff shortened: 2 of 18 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8989">pipeline status</a> for feature branch [PR_8989](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8989). Pipeline [814596118](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/814596118) was triggered at [e35151c](https://github.com/FreeCAD/FreeCAD/pull/8989/commits/e35151c4091ddcc5aa9f361baae46bc37dcca4c2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

WandererFan:
Thanks for the comments - much appreciated.

fixExactMatch is used (@ 1428 in LTNPAutoCorrect3, 1430 in the newest LTNPAutoCorrect4) - that is where we fix the reference when we fix identical geometry in the pile.
handleNoExactMatch is used (@ 1484 / 1486) - the implementation for phase 1 doesn't do much, but it does clear the error flag.

I added the explict keyword you indicated and fixed some of the lint/tidy warning (at least the ones I understood!).

aapo-aapo:
Excellent!  I have now compiled this specific PR, and have confirmed that it works as expected.  I have used it to load some old files that have known corrupted dimensions (due to TechDraw evolution and TNP), and somehow the results with these files are better than before (less wrong dimensions)!  My main reason was to check against unexpected crashes with old files, and I know that old files should not have been improved with this PR, but somehow I got better results than I should have had.  However, my pet peeve, which is radically changing the ISO count for a dimensioned drawing still does not completely work, but seems to work better (as expected).  This is one of the difficult cases, where edges will be split into several pieces; and thus it's not even possible to find edge-to-edge correspondence for dimensions, just vertex-to-vertex mapping is possible.

I also noticed that there are some divisions by View->getScale() in the code and got suspicious about division-by-zero, but it seems that the getScale() function itself is protected against wrong scales:
`    if (!(result > 0.0)) {
        result = 1.0;
    }`

So, it compiles correctly, works, and I cannot find any problems with the code, so I'd say go for it!  :)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8989">pipeline status</a> for feature branch [PR_8989](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8989). Pipeline [815056724](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/815056724) was triggered at [1e450aa](https://github.com/FreeCAD/FreeCAD/pull/8989/commits/1e450aa7b8b05fc542b5d5459e9922a0b6109bb5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #8999, opened 2023-03-22
Fem: Add option to select default solver when creating Analysis

marioalexis84:

---

### DIFF
--- a/src/Mod/Fem/Gui/DlgSettingsFemGeneral.ui
+++ b/src/Mod/Fem/Gui/DlgSettingsFemGeneral.ui
@@ -364,6 +364,73 @@
        </layout>
       </widget>
      </item>
+     <item>
+      <widget class="QGroupBox" name="gb_6_defaults">
+       <property name="title">
+        <string>Defaults</string>
+       </property>
+       <layout class="QHBoxLayout" name="horizontalLayout_6">
+        <item>
+         <layout class="QGridLayout" name="gl_3">
+          <item row="0" column="0">
+           <widget class="QLabel" name="l_def_solver">
+            <property name="enabled">
+             <bool>true</bool>
+            </property>
+            <property name="text">
+             <string>Default solver</string>
+            </property>
+           </widget>
+          </item>
+          <item row="0" column="1">
+           <widget class="Gui::PrefComboBox" name="cmb_def_solver">
+            <property name="enabled">
+             <bool>true</bool>
+            </property>
+            <property name="editable">
+             <bool>false</bool>
+            </property>
+            <property name="toolTip">
+             <string>Default solver used by the analysis container</string>
+            </property>
+            <property name="prefEntry" stdset="0">
+             <cstring>DefaultSolver</cstring>
+            </property>
+            <property name="prefPath" stdset="0">
+             <cstring>Mod/Fem/General</cstring>
+            </property>
+            <item>
+             <property name="text">
+              <string>None</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>CalculiX</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>Elmer</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
+              <string>Mystran</string>
+             </property>
+            </item>
+            <item>
+             <property name="text">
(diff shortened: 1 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8999">pipeline status</a> for feature branch [PR_8999](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8999). Pipeline [814945724](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/814945724) was triggered at [7ca29ce](https://github.com/FreeCAD/FreeCAD/pull/8999/commits/7ca29ce283d1b59a965e7549ed6d46a6265a589e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Many thanks.
I miss in the PR the check for the binaries. One can e.g. set Mystran as default but it is not checked if FreeCAD can access a Mystran binary.

Can you please add the the binary check so that only those solvers can be selected that are available?

(For info: As far as i can judge from the code at the moment the FEM WB ony works properly with these 3 solvers: CCX, Elmer, Z88.
Mystran support is rudimentary. But OK, if a users wants to set it, he should be able to do so.)

donovaly:
Remark: by adding this preferences setting, you still have a default: CCX But the idea is not so promote a certain solver.

Can you therefore please have a look how current master is for you? For me it works as it should be: no solver is preferred:
- I get no automatic solver when adding a new analysis
- I can choose one of the solvers directly in the toolbar since also there no solver is preferred

marioalexis84:
> Therefore I don't think this PR is necessary since it doesn't improve the situation.
Note that for example I use by 50 % CCX and the other 50% Elmer, so having a default solver added on new analyses is something that annoyed me since a while

The option "No solver by default" can be added.
In this way If the user wishes, he can create an analysis without a default solver, and if he wants to have several solvers installed and use a specific one by default, he can also.

With the changes you made, the solvers are not considered at the same level: if only calculix is installed, then it is used by default, but if Elmer or Z88 are installed then not and it is no longer possible to have one by default.

I believe that the changes that I propose here with the addition of the "None default solver" option improve the situation

marioalexis84:
Anyway, I don't think checking the solver binary is that important (although I'm not against doing it). If someone selects a solver either by default or from one of the commands in the menu and this solver is not installed, at the time of doing the analysis they will receive an error, so it is obvious that the program must be installed.

donovaly:
> I believe that the changes that I propose here with the addition of the "None default solver" option improve the situation

But then the "None default solver" should be the default.

> I don't think checking the solver binary is that important

I don't agree here. I have had troubles by myself in installing e.g. z88. It is very valuable that FC tells you if it cannot find it. Therefore I implemented the binary checks in the preferences and thus they should be used.

marioalexis84:
> But then the "None default solver" should be the default.

No problem, that option can be the default.
>Therefore I implemented the binary checks in the preferences and thus they should be used.

Ok, You will implement the checks in the preferences?

marioalexis84:
@donovaly you can now add the check to preferences.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8999">pipeline status</a> for feature branch [PR_8999](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_8999). Pipeline [816501126](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/816501126) was triggered at [770e2c6](https://github.com/FreeCAD/FreeCAD/pull/8999/commits/770e2c626385f0fa46d52ada89bc3258e67e0f92). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly (changes requested):
Many thanks, just the binary check please

donovaly:
Here it is: #9031

@marioalexis84 , can you please do me the favor and delete all your branches that were already merged also remote? The point is that when I checkout a PR from you I get tons of branches because it seems you never deleted your remote branches.

PR #9002, opened 2023-03-23
Fem: Group solver commands on toolbar

marioalexis84:
The Icon file name for CalculiX changed from FEM_SolverStandard.svg to FEM_SolverCalculiX.svg
In the menu bar, the commands are intentionally left ungrouped due to the presence of the `CalculiX (new framework)` command.

---


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9002">pipeline status</a> for feature branch [PR_9002](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9002). Pipeline [815131640](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/815131640) was triggered at [5e43506](https://github.com/FreeCAD/FreeCAD/pull/9002/commits/5e435063a1819f17e006ed74cf22e4f146d296ba). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

donovaly:
Many thanks.
However I think we should leave the 3 solvers ungrouped in the toolbar. We purposely don't have the Mystran solver there and the experimental CCX solver in the toolbar. For convenience reason only the 3 main supported are there and I would like to keep that. I remember some discussions that lead to this, but can't find them right now. However, since people debated about this, I would no reopen this discussion. We would only save the space of 2 icons.

donovaly:
Another aspect: at the moment the CCX solver is promoted by FC ads kind of default. The goal is to treat all solvers the same. I am working currently on a PR that creates just a plain Analysis container if there is an Elmer of Z88 binary found. Then we have real equality - users can choose the solver they like.
However, when we group the solvers in the toolbar, the one solver will be visible the others are by default hidden.

So what about this?:
- when there is no Z88 or Elmer binary found, the solvers are grouped, otherwise not.
We have the information about the binaries, so it should be possible to decide in Workbench.cpp if we use the plain or Comp solver items.

marioalexis84:
>at the moment the CCX solver is promoted by FC ads kind of defaul

I don't think it's right to promote it as default because it's not included in FreeCAD. Even linux distribution packages do not include any default solvers.

> I am working currently on a PR that creates just a plain Analysis container

Please, see https://github.com/FreeCAD/FreeCAD/pull/8999.

donovaly:
> > at the moment the CCX solver is promoted by FC ads kind of defaul
> 
> I don't think it's right to promote it as default because it's not included in FreeCAD. Even linux distribution packages do not include any default solvers.

Yes, as I wrote, we should not promote a special solver. However, at least in the Conda and Windows packages CCX is included. Most users use these builds. Then it makes sense to set CCX.
I therefore made #9017 which only adds CCX of not other solver binary is found.

> Please, see #8999.

Oh, I missed this, will have a look.

marioalexis84:
>Then it makes sense to set CCX

For that reason I kept CalculiX as the default option in #8999 
>So what about this?:

Another option is to disable the command if the solver is not installed and add a message in the tool tip about it. I think it's better to change this in the python file and not add conditionals to the workbench.cpp, to keep it cleaner.

donovaly:
> Another option is to disable the command if the solver is not installed and add a message in the tool tip about it. I think it's better to change this in the python file and not add conditionals to the workbench.cpp, to keep it cleaner.

I re-thought and I think we should offer the different solvers plain as we already do in the toolbar, but we should omit this solvers that are not available. For example if Z88 is not installed, there is no reason to offer it in the toolbar.
So users just with CCX will only have one toolbar button, those with Elmer too, will have 2.

donovaly:
> I re-thought and I think we should offer the different solvers plain as we already do in the toolbar, but we should omit this solvers that are not available. For example if Z88 is not installed, there is no reason to offer it in the toolbar. So users just with CCX will only have one toolbar button, those with Elmer too, will have 2.

After some more thoughts, I still think this is the optimal solution. The vast majority of users only have one solver and end up with a single toolbar button.
users like me with 2 solvers, get 2 toolbar buttons.
Here is the PR that implements this: #9035

So compared to the current situation we save 2 toolbar buttons for most users.

donovaly:
Closing in favor of #9035

Issue #8692, opened 2023-02-27
[Problem] Workbench dropdown is difficult for a new user to navigate


chennes:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
OS: macOS 13.2
Word size of FreeCAD: 64-bit
Version: 0.21.0.31802 (Git)
Build type: Release
Branch: master
Hash: 7352512f6c7e312728bcb64ccf9e7c890df983cd
Python 3.10.9, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: English/United States (en_US)
```

### Subproject(s) affected?

None

### Problem description

This was discussed at FreeCAD Day 2023:

In a new installation of FreeCAD, the standard workbenches list is arranged alphabetically and contains about twenty workbenches. This makes it difficult for a new user to discover the standard workflows in FreeCAD, and leads to clutter in the user interface.

We discussed a number of possible solutions (these are not mutually exclusive):
1) A first-run dialog that helps a new user select which workbenches they expect to use the most, and displays those more prominently in the user interface.
2) An easier-to-discover system for editing the workbenches menu (for example, an option at the bottom of that menu called "Edit...")
3) The ability to create submenus to either group those workbenches in some predefined way, or to allow less-used workbenches to be tucked away and not clutter the main list.
4) An option to auto-sort the list so that the most-recently-used workbench is at the top of the list.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
I would combine No. 1 with a selector toolbar add-on style selector instead of the current genuinely annoying to use drop-down.

luzpaz:
@MisterMakerNL was working on a first run dialog.

ngocminhdao88:
I have an idea. How about a search box with fuzzy search. User types in few keywords and FreeCAD filters workbenches, which matched the search criteria. Maybe filter wb combobox contents when it focused and user typing is also a nice touch.

Intuos5:
> I have an idea. How about a search box with fuzzy search. User types in few keywords and FreeCAD filters workbenches, which matched the search criteria. Maybe filter wb combobox contents when it focused and user typing is also a nice touch.

That would inlo work well with an extensive command search, an extended version of Tripkus' launcher.

yorikvanhavre:
[...]

PR #9026, opened 2023-03-24
Preferences : Rework Workbench page by merging the 'customize' dialog capability

PaddleStroke:
- This PR adds the capability of enabling / disabling workbenches in the preference page.
- It also removes the customize Workbench page.
- Move startup workbench pref to the workbench page
- Move workbench selector position to the workbench page.

First step towards https://github.com/FreeCAD/FreeCAD/issues/8692

Video : https://youtu.be/KzKVq2Spizg

---


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [816878069](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/816878069) was triggered at [1dc92bf](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/1dc92bfb361a809db787033cff3ad40188c2620d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [816882244](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/816882244) was triggered at [b7f5923](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/b7f5923b16d87c6904793a16f154d14aee02b494). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
This is a much more intuitive approach to workbench customization.  Two suggestions:
The order of workbenches has two effects but only one is obvious to the user. 
- It affects the order they appear in the dropdown
- The first nine workbenches get accelerator keys  w,<n>

Reordering the workbenches by dragging is a really nice feature.  We should show the corresponding accelerator key in another column.  Maybe in the future, the user could even assign their own accelerator key.

Changing things requires a restart.  There's a handy note at the top indicating this.  In the addon workbench, exiting presents a dialog reminding the user that a restart is required and offering the option to do so immediately or later.  I suggest eliminating that part of the note and using the same workflow as addon manager

chennes:
It looks from your video like you were basing your dialog design on an old version of the Workbenches dialog -- a week or two ago we merged some code that shifted all of the columns around so that it was easier to see which workbench was being worked on. Have a look at https://github.com/FreeCAD/FreeCAD/commit/5e77365c0d

PaddleStroke:
No I'm based on master. The order was reversed indeed in my changes.
There are a lot of changes, notably the widget is not a Qtable anymore. So
it's now easy to see on which row the buttons are.
If you feel the buttons are still too far from the label, we can think
about a solution. However I don't think that a weird order is the correct
solution.

chennes:
I don't have strong feeling about it, I'm just hesitant to overwrite @ronnystandtke's work so soon after its addition. But that's the nature of the beast, after all! I'm happy to see this dialog getting some attention: it was originally added solely to reduce user complaints about missing preference pages for workbenches that hadn't been loaded yet. Making it actually do something productive is very nice.

ronnystandtke:
No problem. I'm still not a friend of huge gaps between controls that belong together but if the visibility of rows is improved so that it is still clear which button belongs to which workbench I'm absolutely fine with the changes.

adrianinsaval:
Very welcomed change, I think the fact the line gets highlighted when you click on it is enough to make it possible to distinguish what workbench the button will be acting on.

PaddleStroke:
I don't get it. Preferences do not customize the GUI? Preferences are by definition intended to customize the GUI. Having two separated dialogs is just splitting the settings in several places, which for the user is bad.
This is only a first step, but I intend to merge completely the customize into preference page.

xtemp09:
Maybe this should be called Addon Manager, while the other one should be Addon Downloader.

I think, a button to download additional workbenches should be present in the tab.

PaddleStroke:
I have done the following : 
- Move startup workbench pref to the workbench page
- Move workbench selector position to the workbench page.
- Made the load button closer to the labels.

So for the remaining improvement ideas : 
- Restart application message after clicking ok.
- Showing the shortcuts (w i).
- Adding a button to launch addon manager.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [818128242](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/818128242) was triggered at [8ce1ba5](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/8ce1ba583aec63057bb02665fc579369ca205679). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+/**
+ *  Constructs a DlgSettingsWorkbenchesImp
+ */
+DlgSettingsWorkbenchesImp::DlgSettingsWorkbenchesImp( QWidget* parent )
+    : PreferencePage( parent )
+    , ui(new Ui_DlgSettingsWorkbenches)
+{
+    ui->setupUi(this);
+    connect(ui->AutoloadModuleCombo, QOverload<int>::of(&QComboBox::activated), this, [this](int index) { onStartWbChangedClicked(index); });
```
0penBrain, line 57: Is using `qOverload` still a problem with MSVC ?
```suggestion
    connect(ui->AutoloadModuleCombo, qOverload<int>(&QComboBox::activated), this, [this](int index) { onStartWbChangedClicked(index); });
```

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+{
+    std::ostringstream enabledStr, disabledStr, autoloadStr;
+
+    for (int i = 0; i < ui->wbList->count(); i++) {
+        QWidget* widget = ui->wbList->itemWidget(ui->wbList->item(i));
+        if (!widget)
+            continue;
+        QCheckBox* enableBox = widget->findChild<QCheckBox*>(enableCheckboxStr);
+        if (!enableBox)
+            continue;
```
0penBrain, line 78: This looks super hacky. IMO better to define our own item model with appropriate methods.

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+        QCheckBox* autoloadBox = widget->findChild<QCheckBox*>(autoloadCheckboxStr);
+        if (!autoloadBox)
+            continue;
+
+        if (autoloadBox->isChecked()) {
+            if (!autoloadStr.str().empty())
+                autoloadStr << ",";
+            autoloadStr << widget->objectName().toStdString();
+        }
```
0penBrain, line 99: Hacky here again

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+        if (!widget)
+            continue;
+        QCheckBox* enableBox = widget->findChild<QCheckBox*>(enableCheckboxStr);
+        if (!enableBox)
+            continue;
+
+        if (enableBox->isChecked()) {
+            if (!enabledStr.str().empty())
+                enabledStr << ",";
+            enabledStr << widget->objectName().toStdString();
```
0penBrain, line 83: You could advantageously use a lambda for these cases.

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+    // replace load button with loaded indicator
+    for (int i = 0; i < ui->wbList->count(); i++) {
+        QWidget* widget = ui->wbList->itemWidget(ui->wbList->item(i));
+        if (widget && widget->objectName() == wbName) {
+            QWidget* loadLabel = widget->findChild<QWidget*>(loadLabelStr);
+            QWidget* loadButton = widget->findChild<QWidget*>(loadButtonStr);
+            loadButton->setVisible(false);
+            loadLabel->setVisible(true);
+            break;
```
0penBrain, line 168: Hacky too (guess it's copied from before).

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+    QStringList workbenches = Application::Instance->workbenches();
+    QStringList enabledWbs = getEnabledWorkbenches();
+    QStringList disabledWbs = getDisabledWorkbenches();
+
+    ui->wbList->setDragDropMode(QAbstractItemView::InternalMove);
+    ui->wbList->setSelectionMode(QAbstractItemView::SingleSelection);
+    ui->wbList->viewport()->setAcceptDrops(true);
+    ui->wbList->setDropIndicatorShown(true);
+    ui->wbList->setDragEnabled(true);
+    ui->wbList->setDefaultDropAction(Qt::MoveAction);
```
0penBrain, line 215: Why not done in the constructor ?

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+    // 1: Enable checkbox
+    auto enableCheckBox = new QCheckBox(this);
+    enableCheckBox->setToolTip(tr("If unchecked, %1 will not appear in the available workbenches.").arg(wbDisplayName));
+    enableCheckBox->setChecked(enabled);
+    enableCheckBox->setObjectName(enableCheckboxStr);
+    if (wbName.toStdString() == _startupModule) {
+        enableCheckBox->setChecked(true);
+        enableCheckBox->setEnabled(false);
+        enableCheckBox->setToolTip(tr("This is the current startup module, and must be enabled. See Preferences/General/Autoload to change."));
+    }
```
0penBrain, line 264: Reading this and rest of the code, I have a doubt if the list correctly updates if "load on startup" WB is changed (didn't test).

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+        if (widget) {
+            QCheckBox* enableCheckbox = widget->findChild<QCheckBox*>(enableCheckboxStr);
+            if (enableCheckbox && widget->objectName() == wbName) {
+                enableCheckbox->setEnabled(false);
+            }
+            else {
+                enableCheckbox->setEnabled(true);
+            }
+        }
+    }
```
0penBrain, line 472: OK, got the answer to previous point. However such way of doing creates a strong interconnexion between widgets in the page which isn't ideal.

src/Gui/DlgSettingsWorkbenchesImp.h
```diff
@@ -31,42 +31,59 @@ class QCheckBox;
...
-    explicit DlgSettingsLazyLoadedImp( QWidget* parent = nullptr );
-    ~DlgSettingsLazyLoadedImp() override;
+    explicit DlgSettingsWorkbenchesImp( QWidget* parent = nullptr );
+    ~DlgSettingsWorkbenchesImp() override;
 
     void saveSettings() override;
     void loadSettings() override;
 
+    static QStringList getEnabledWorkbenches();
+    static QStringList getDisabledWorkbenches();
```
0penBrain, line 54: Why are they static functions ?

src/Gui/QListWidgetDragBugFix.cpp
```diff
@@ -0,0 +1,58 @@
...
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "PreCompiled.h"
+#ifndef _PreComp_
+# include <QDragMoveEvent>
+#endif
+
+#include "QListWidgetDragBugFix.h"
```
0penBrain, line 28: Name is really not very good. :)
First you should never start with `Q` as this denotes a class genuinely provided by Qt.
So something like `ListWidgetDragDrop` ?

src/Gui/QListWidgetDragBugFix.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+}
+
+/* Qt has a recent bug (2023, https://bugreports.qt.io/browse/QTBUG-100128) 
+* where the items disappears in certain conditions when drag and dropping.
+* Here we prevent the situation where this happens.
+* 1 - If the item is dropped on the item below such that the item doesn't move (ie superior half of the below item)
+* 2 - The item is the last one and user drop it on the empty space below.
+* In both those cases the item widget was lost.
+ */
```
0penBrain, line 46: Place in the header file

src/Gui/QListWidgetDragBugFix.cpp
```diff
@@ -0,0 +1,58 @@
...
+void QListWidgetDragBugFix::dragMoveEvent(QDragMoveEvent *e)
+{
+    if ((row(itemAt(e->pos())) == currentRow() + 1) 
+        || (currentRow() == count() - 1 && row(itemAt(e->pos())) == -1)) {
+        e->ignore();
+    }
+    else {
+        QListWidget::dragMoveEvent(e);
+    }
```
0penBrain, line 55: I generally prefer to have
```suggestion
    if ((row(itemAt(e->pos())) == currentRow() + 1) 
        || (currentRow() == count() - 1 && row(itemAt(e->pos())) == -1)) {
        e->ignore();
        return;
    }
    QListWidget::dragMoveEvent(e);
```
But maybe it's personal taste only.

src/Gui/QListWidgetDragBugFix.h
```diff
@@ -0,0 +1,42 @@
...
+#include <QDragMoveEvent>
+#include <QListWidget>
+
+
+class QListWidgetDragBugFix : public QListWidget
+{
+    Q_OBJECT
+
+public:
+    QListWidgetDragBugFix(QWidget *parent);
```
0penBrain, line 35:
```suggestion
    explicit QListWidgetDragBugFix(QWidget *parent);
```

src/Gui/DlgSettingsWorkbenchesImp.cpp
PaddleStroke: Without qOverload or with your version it doesn't build on MSVC

0penBrain: OK, I remembered MSVC generated issues with this...
It's solved from MSVC 2019 IIRC.

src/Gui/DlgSettingsWorkbenchesImp.cpp
PaddleStroke: It was how it was done before for the load and auto-load checkboxes. I reused the same pattern here.
But yes it's true I can make a custom widget class with pointers to the different subwidgets. It would make more sense.

src/Gui/DlgSettingsWorkbenchesImp.h
PaddleStroke: I just copy pasted them from the customize dlg. I think they are static because they are used in Action.cpp.

src/Gui/QListWidgetDragBugFix.cpp
PaddleStroke: You are corerct I shouldn't have started it with Q. I dumbly reused the previously used 'QListWidgetCustom' and replaced Custom with a better name.
Regarding the name, I think it's a good name because it reminds us the exact purpose of this class, that is to fix the qt bug. And that later when Qt fix the bug this will not be necessary anymore.
I think `ListWidgetDragDrop` might be misleading in the way that it does not clearly state that this should be temporary until qt fixes the bug.

src/Gui/DlgSettingsWorkbenchesImp.cpp
0penBrain: You don't need pointers. You can use a customized QListView, or QItemDelegate to handle data.

src/Gui/DlgSettingsWorkbenchesImp.h
0penBrain: Good reason. :)

src/Gui/QListWidgetDragBugFix.cpp
PaddleStroke: Makes sense.

src/Gui/DlgSettingsWorkbenchesImp.cpp
PaddleStroke: No reasons, it was just where those were before. I'll move them to ctor it's better.

src/Gui/DlgSettingsWorkbenchesImp.cpp
PaddleStroke: The list updates correctly when 'start' wb is changed.
Note, the 'start' combobox shows only the wb that are activated.

src/Gui/DlgSettingsWorkbenchesImp.cpp
PaddleStroke: You mean as it's done in Element widget?
I know custom delegate are the best way to do widgets in list. But this is a lot of work that I don't think is justified here because we have a limited number of entries (vs element which can have hundreds of entries). Besides the custom delegate seems harder to maintain as can attest the recent checkbox addition to element widget which have (had?) selection issues. Besides this is not a core functionality that needs to be as efficient as possible.

So I suggest just making a custom widget to access the subwidget in a cleaner way than with the qobject names.

0penBrain: It's a bit tweaky regarding the situation, but you can also have a look at `QListWidget::setItemWidget`. Should be enough to do this.

PaddleStroke: setItemWidget is what I'm using now actually. I just forced push with changeds. I made a subclass of qwidget. Can you please check if it is acceptable like this?
Thanks

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+
+using namespace Gui::Dialog;
+
+const QString DlgSettingsWorkbenchesImp::iconLabelStr = QString::fromLatin1("iconLabel");
+const QString DlgSettingsWorkbenchesImp::nameLabelStr = QString::fromLatin1("nameLabel");
+const QString DlgSettingsWorkbenchesImp::loadLabelStr = QString::fromLatin1("loadLabel");
+const QString DlgSettingsWorkbenchesImp::loadButtonStr = QString::fromLatin1("loadButton");
+const QString DlgSettingsWorkbenchesImp::enableCheckboxStr = QString::fromLatin1("enableCheckbox");
+const QString DlgSettingsWorkbenchesImp::autoloadCheckboxStr = QString::fromLatin1("autoloadCheckbox");
```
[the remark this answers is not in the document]
PaddleStroke: Removed with the new qwidget subclass.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [818227903](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/818227903) was triggered at [f01acc0](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/f01acc0843cf5a81d8a79871891e9e7d786d0660). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/QListWidgetDragBugFix.cpp
sliptonic: Does the FreeCAD project have a published standard for variable naming in C++ code?  I know there's some general conventions and I've seen some forum posts about it but can't find an official standard.

AjinkyaDahale:
@PaddleStroke minor nitpick: please keep the summary of the commits short, at least so as not to trigger the `...` in Github. In case you were unaware, you can add more details by adding one blank line after the summary.

Currently, I think you could use short forms like `Prefs: WB:`.

src/Gui/QListWidgetDragBugFix.cpp
0penBrain: > Does the FreeCAD project have a published standard for variable naming in C++ code? 

Don't think so.

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,434 @@
...
+    iconLabel->setEnabled(checked);
+    textLabel->setEnabled(checked);
+    loadLabel->setEnabled(checked);
+    loadButton->setEnabled(checked);
+    autoloadCheckBox->setEnabled(checked);
+    if (!checked) //disabling wb disable auto-load.
+        autoloadCheckBox->setChecked(false);
+
+    // Reset the start combo items.
+    dlg->setStartWorkbenchComboItems();
```
0penBrain, line 147: Rather than breaking encapsulation here, you'd better have to emit a signal and declare `setStartWorkbenchComboItems` as a slot, so that you can connect them together.

src/Gui/DlgSettingsWorkbenchesImp.h
```diff
@@ -31,42 +31,71 @@ class QCheckBox;
...
-    std::map<QString, QCheckBox*> _autoloadCheckBoxes;
+    void saveWorkbenchSelector();
+    void loadWorkbenchSelector();
+
+    void onStartWbChangedClicked(int index);
+
+    std::unique_ptr<Ui_DlgSettingsWorkbenches> ui;
+};
+
+class wbListItem : public QWidget
```
0penBrain, line 75: I guess according usage scope of this class, it could have been directly defined in CPP file.

src/Gui/DlgSettingsWorkbenchesImp.h
```diff
@@ -31,42 +31,71 @@ class QCheckBox;
...
+
+public:
+    QCheckBox* enableCheckBox;
+    QCheckBox* autoloadCheckBox;
+    QLabel* iconLabel;
+    QLabel* textLabel;
+    QLabel* loadLabel;
+    QPushButton* loadButton;
+
+    DlgSettingsWorkbenchesImp* dlg;
```
0penBrain, line 95: Generally having public member variables is considered bad practice.
Better make them private and create just needed setters and getters. ;)

0penBrain:
@PaddleStroke CI fails (actually build fails) because of missing includes.

sliptonic:
> Personally I am not so sure about the real need for this change...

This comment is what the contributing policy refers to as a 'value judgement'.   It's a comment about  the value of a contribution.  These comments shouldn't be made in pull requests.  They should be made in the original issue, BEFORE development work is done.  The discussion about eliminating the redundancy of the 'customize' and 'preference' dialogs occured almost a week ago and before this was coded.  Nobody objected at that time.

0penBrain:
> The discussion about eliminating the redundancy of the 'customize' and 'preference' dialogs occured almost a week ago and before this was coded.

That would greatly help that we have a tag to identify tickets where decisions about changes are taking place. Now the ticket queue is a mess, I don't know how to sort it.

sasobadovinac:
> > Personally I am not so sure about the real need for this change...
> 
> This comment is what the contributing policy refers to as a 'value judgement'. It's a comment about the value of a contribution. These comments shouldn't be made in pull requests. They should be made in the original issue, BEFORE development work is done. The discussion about eliminating the redundancy of the 'customize' and 'preference' dialogs occured almost a week ago and before this was coded. Nobody objected at that time.

Oh, you mean that issue that has nothing to do with this and that comment that no one replayed to with a positive or negative opinion about it?

sliptonic:
> > > Personally I am not so sure about the real need for this change...
> > 
> > 
> > This comment is what the contributing policy refers to as a 'value judgement'. It's a comment about the value of a contribution. These comments shouldn't be made in pull requests. They should be made in the original issue, BEFORE development work is done. The discussion about eliminating the redundancy of the 'customize' and 'preference' dialogs occured almost a week ago and before this was coded. Nobody objected at that time.
> 
> Oh, you mean that issue that has nothing to do with this and that comment that no one replayed to with a positive or negative opinion about it?

Yep.  That's the one.  
I'll be the first to agree that we're not doing this well (yet).  We're still figuring out how to write good issues, when to split them, and how to get the right people involved.  
But we're getting better. 

We still shouldn't be re-opening the discussion and introducing the value judgments after the work is done.  If the contribution creates problems, we can revert it later.   The goal is faster cycle times, more iteration, and more experimentation.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [818743220](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/818743220) was triggered at [49f3318](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/49f33181e8ab14402fc8bc261cc215ad8f51e216). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/DlgSettingsWorkbenchesImp.h
0penBrain: Better use it. You asked to me a question in one of my PR just few days ago 😉

0penBrain: https://github.com/FreeCAD/FreeCAD/pull/8952#discussion_r1141282500

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [819091650](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/819091650) was triggered at [5c174dd](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/5c174dd0b46228ebe4f041191cd44ce4dc6ee570). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/DlgSettingsWorkbenchesImp.cpp
PaddleStroke: Yes, I removed the dlg pointer.

src/Gui/DlgSettingsWorkbenchesImp.h
PaddleStroke: Oh right. Sorry I forgot about this point!

src/Gui/DlgSettingsWorkbenchesImp.cpp
PaddleStroke: Done

src/Gui/DlgSettingsWorkbenchesImp.cpp
PaddleStroke: Not sure what your point was here. Please have a look at the new code as it changed a lot with other fixes.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [819343769](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/819343769) was triggered at [2483f23](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/2483f2381b9d89d9a5ae0c9ccc9ccf67eda48e65). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
I'm in the middle of the 'restart freecad' window in preferences. So maybe I'll just finish that as well.

For the copyright comment at the beginning of the file, can you please tell me if you want me to update it in all the files that I modified or only the files I created?

chennes:
I don't really care: I'm trying to update it when I touch files, but I don't always remember.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [820157716](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/820157716) was triggered at [2a6ecfd](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/2a6ecfdc29a2f77be03e092715e88305f4cf15fe). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
@chennes I have finished and replaced the copyright in the files that were modified as you mentioned.
Please let me know if you spot any issue in the last commit that adds the 'Require restart' to prefs.
After we can also use this 'require restart' elsewhere in the pref (example the comboview setting requires a reboot).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [820172611](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/820172611) was triggered at [7b956fd](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/7b956fd33d02b73bcdec5fad14d3120865b73d57). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026">pipeline status</a> for feature branch [PR_9026](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9026). Pipeline [820379305](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/820379305) was triggered at [d0bd202](https://github.com/FreeCAD/FreeCAD/pull/9026/commits/d0bd2022935b5b9b1304b447f0581ae8dc6a9b21). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
OK, think this looks good now -- please make sure to edit the wiki so that the documentation tracks with the new changes, I think we'll have to start to field questions about where the old customize options went as soon as the weeklies get rebuilt, it would be nice to be able to point folks to the updated documentation as soon as possible. There will surely be at least a few complaints.

PaddleStroke:
Thanks for reviewing ! :)

PaddleStroke:
For the wiki I'm unsure what to do. Because the production version is 0.20 so if I replace the picture in the wiki, regular 0.20 users are going to be confused.

wwmayer:
Problem 1:
Version 0.21 and 0.20 share the same user.cfg file. Since both builds have a different workbenches (e.g. Image exist in 0.20 but not any more in v0.21) this  can trigger some annoying warnings. For example when running v0.20 and afterwards v0.21 you get this warning `Ignoring unknown ImageWorkbench workbench found in user preferences.` when opening the preferences.

If the difference is more than one workbench then the warnings are all written to a single line.

Suggested fixes:
* the warning should be degraded to a log message because the user anyway cannot do anything
* the log message must contain a newline `\n` at the end

Problem 2:
When I change anything on the workbench tab (change order, disable or enable a wb) and click OK then a message appears to restart FreeCAD now or later. If I restart it now then a warning appears after it has restarted:
`File format not supported: ./bin/FreeCAD`

If I change again something in the workbench tab, click OK and confirm to restart the message after the restart will be:
```
File format not supported: /home/user/Projects/build_clang/bin/FreeCAD 
File format not supported: ./bin/FreeCAD 
```

Every time I restart the application after making changes in the workbench tab the warnings get longer and longer.

src/Gui/DlgPreferencesImp.cpp
```diff
@@ -471,6 +477,34 @@ void DlgPreferencesImp::applyChanges()
...
+
+        int exec = restartBox->exec();
+
+        if (exec == QMessageBox::Ok) {
+            //restart FreeCAD after a delay to give time to this dialog to close
+            QTimer::singleShot(1000, []() 
+            {
+                QStringList args = QApplication::arguments();
+                if (getMainWindow()->close())
+                    QProcess::startDetached(QApplication::applicationFilePath(), args);
```
wwmayer, line 502: Passing the unchanged `args` is incorrect because the first element is always the path to the executable. So, the first element of the list must be removed.

src/Gui/CommandStd.cpp
```diff
@@ -383,6 +385,7 @@ void StdCmdDlgPreferences::activated(int iMsg)
     if (cDlg.exec()) {
         cDlg.activeGroupPage(groupName, index);
     }
+    cDlg.isRebootRequired(); //The user may have applied first, then clicked the cancel button so it's not in the if(cDlg.exec())
```
wwmayer, line 388: This is bad design. If the dialog recommends a reboot or not is an implementation detail and should not be exposed as public function.

The correct way is to handle this inside DlgPreferencesImp::accept() and DlgPreferencesImp::reject()

src/Gui/DlgPreferencesImp.cpp
```diff
@@ -471,6 +477,34 @@ void DlgPreferencesImp::applyChanges()
     }
 }
 
+void DlgPreferencesImp::isRebootRequired()
```
wwmayer, line 480: The function name is confusing because one would assume it's a getter function but it does something and doesn't return a value. Also it doesn't _reboot_ but restarts the application.

src/Gui/PropertyPage.cpp
```diff
@@ -105,6 +107,18 @@ void PreferencePage::changeEvent(QEvent* event)
     QWidget::changeEvent(event);
 }
 
+bool PreferencePage::isRebootRequired()
```
wwmayer, line 110: Method should be const

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,576 @@
...
+    QStringList enabledWbs = getEnabledWorkbenches();
+    QStringList disabledWbs = getDisabledWorkbenches();
+
+    //First we add the enabled wbs in their saved order.
+    for (const auto& wbName : enabledWbs) {
+        if (workbenches.contains(wbName)) {
+            addWorkbench(wbName, true);
+        }
+        else {
+            Base::Console().Warning("Ignoring unknown %s workbench found in user preferences.", wbName.toStdString().c_str());
```
wwmayer, line 355: Should be a Log() message and the string lacks of `\n`

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,576 @@
...
+        }
+    }
+    //Second we add workbench in alphabetical order that are either Disabled, or !enabled && !disabled, ie newly added wb.
+    for (const auto& wbName : workbenches) {
+        if (disabledWbs.contains(wbName)) {
+            addWorkbench(wbName, false);
+        }
+        else if (!enabledWbs.contains(wbName)) {
+            Base::Console().Warning("Adding unknown %s workbench.", wbName.toStdString().c_str());
```
wwmayer, line 364: Same here

src/Gui/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -0,0 +1,576 @@
...
+
+QStringList DlgSettingsWorkbenchesImp::getEnabledWorkbenches()
+{
+    QString enabled_wbs;
+    QStringList enabled_wbs_list;
+    ParameterGrp::handle hGrp;
+    QString allWorkbenches = QString::fromLatin1("ALL");
+
+    hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Workbenches");
+    enabled_wbs = QString::fromStdString(hGrp->GetASCII("Enabled", allWorkbenches.toStdString().c_str()).c_str());
```
wwmayer, line 395: The c_str() at the end is not needed and creates an unnecessary tmp. copy of std::string

wwmayer:
There is also a bug now:
When a user installs a new addon and restarts FreeCAD then it can happen that the new wb is not added to the end but inserted anywhere. This makes it harder to find it.

When this happened to me then all workbenches that have been disabled all the sudden re-appeared in the workbench selector and in the settings they were marked as enabled. The workbench selector even showed the `<none>` workbench now.

What can happen too is that the new installed wb correctly appears in the workbench selector but is marked as disabled in the preferences. So, if the user opens the preferences dialog and closes with OK then after the next restart the newly added workbench is missing.

PaddleStroke:
@wwmayer I will look into those issue now and issue a PR asap.

src/Gui/DlgPreferencesImp.cpp
PaddleStroke: I copied that from the addon manager. So there is probably the same problem there. Should it be fixed in the same PR?

PaddleStroke: Note adding `args.pop_front();` removes correctly the `File format not supported: /home/user/Projects/build_clang/bin/FreeCAD`  message on restart.

src/Gui/CommandStd.cpp
PaddleStroke: Got it.

PaddleStroke:
@wwmayer regarding the issue about 

> Error at file "(memory)", line 167, column 64
> Message: invalid character encountered
> ...

I think this might be a problem linked to a partial rebuild. I had some issues when I have done only partial builds where the rebootRequired variable was not initialized in all the pref pages. But after a full rebuild the issue was gone. Can you please confirm if this needs to be further investigated or if it's gone?

wwmayer:
> I will have a closer look if I can find something but if not then we should leave it like that.

I couldn't find anything suspicious. 

> When you mean it's inserted anywhere, you mean in the wb selector, not in preferences right?

Yes.

> If you don't launch the preferences, then there is no reason that a bug would have been introduced.

But I have never seen this behaviour before.

> This needs further investigation.

OK.

> Any recommendation to the group name?

No, feel free to find a nice name :)

PaddleStroke:
Is there a reason why addonmanager does not register newly installed addons as 'enabled' (nor disabled), and that action.cpp has what feels like a workaround where displayed workbenches is not just enabled but total - disabled as to pickup workbenches that are neither enabled or disabled?

PaddleStroke:
Besides before my PR there was an inconsistency. If you add a new workbench without having customized active wbs, then the added wb is listed alphabetically. If you have customized active wbs, then it adds at the end of the list.

PaddleStroke:
I suggest removing the 'disabled' list preference. It should be binary, either enabled, either disabled. Not enabled, disabled or else.
Addon manager should register new wb as enabled.
This would simplify the logic here.

PaddleStroke:
Another inconsistency, addon-manager can disable workbenches in it's own way. I think this should be removed we should have only one enable/disable system.

@chennes any input on this?

wwmayer:
> Is there a reason why addonmanager does not register newly installed addons as 'enabled'

This is a question for @chennes 

> Besides before my PR there was an inconsistency. If you add a new workbench without having customized active wbs, then the added wb is listed alphabetically. 

Yeah, that's probably the issue I have seen.

> I suggest removing the 'disabled' list preference. It should be binary, either enabled, either disabled. Not enabled, disabled or else.

Makes sense.

chennes:
The Addon Manager is a glorified git frontend, it is attempting to behave the same way a user manually installing an Addon does. That said, I inherited that behavior and have no opposition to doing some sort of "registration."

Regarding the disable thing -- I wanted an option that a user could do manually, outside of FreeCAD, for when things are really going sideways. Since they can always just move the directory out of Mod, though, I don't mind changing to use the built-in enable/disable feature instead.

PR #9031, opened 2023-03-24
[FEM] improve new default solver setting

donovaly:
- fix bug if no solver should be default
- only provide solvers that are available to be the default
- also add commit opened transactions
- also remove unused include and correct an include

### DIFF
--- a/src/Mod/Fem/Gui/DlgSettingsFemCcxImp.cpp
+++ b/src/Mod/Fem/Gui/DlgSettingsFemCcxImp.cpp
@@ -29,7 +29,7 @@
 # include <QMessageBox>
 #endif
 
-#include <Gui/Application.h>
+#include <App/Application.h>
 
 #include "DlgSettingsFemCcxImp.h"
 #include "ui_DlgSettingsFemCcx.h"
--- a/src/Mod/Fem/Gui/DlgSettingsFemGeneral.ui
+++ b/src/Mod/Fem/Gui/DlgSettingsFemGeneral.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>425</width>
-    <height>447</height>
+    <width>411</width>
+    <height>450</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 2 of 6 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9031">pipeline status</a> for feature branch [PR_9031](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9031). Pipeline [817296109](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817296109) was triggered at [8e692a3](https://github.com/FreeCAD/FreeCAD/pull/9031/commits/8e692a33f0142b51903db7402f6135673f1439af). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

marioalexis84:
With these changes, on Linux, installed solvers no longer appear in the list of default solvers in preferences if the "Search known binary directories" checkbox is checked: It is necessary to add the path to the solver explicitly.
Even in that case, the resolver selected as default is not used, for example currently on my system only the Elmer resolver is shown in the preferences list (because I type the path manually in the Elmer's preferences) and if I select this as default, when creating the Analysis container, the CalculiX solver is added to the analysis.

marioalexis84:
Also the icons no longer appear despite having the solvers installed and must be selected from the menu bar

donovaly:
Thanks for the report. Can you please have a look and fix this since I only have windows and here I must always explicitly set the path to the binary. Therefore I cannot test well how the binaries are found.

But OK, the Python part of FEM detects the solvers via known paths, so the C++ part must be able to do so as well. I think the fix is simply to take the Python part's solution via a doCommand instead of checking the preferences property.

If you won't have time, please tell me and I will try to to have a look tomorrow.

marioalexis84:
Later I will create a PR with the fix.

donovaly:
Thanks.
Meanwhile I checked that the Python part's FEM binary check does it all right. The "search in known places" feature is in the Python part.

It was a big mistake that I did not realize this earlier, sorry.

donovaly:
I have a PR read, just doing tests. The idea is that QFileInfo does the job since it also checks the system variables of every OS.

marioalexis84:
since you asked me if I could correct it. I'm working on a PR for this.
Can you wait for me to finish it?

donovaly:
I am finished already. So let's both make the PR and then discuss what to use.

donovaly:
Here is my PR: #9051

it was surprisingly easy to code. I think with this we have nice C++ version of the binary_check we ave in Python.

PR #9043, opened 2023-03-25
[FEM] Elmer: fix transient analyses

donovaly:
- different time results were never output, only the final one the reason is that the Output Intervals intervals property was missing
- a second reason was that the result scaling solver must be executed every time an output should be generated
- the BDF Order property is a transient-only property

### DIFF
--- a/src/Mod/Fem/femsolver/elmer/solver.py
+++ b/src/Mod/Fem/femsolver/elmer/solver.py
@@ -101,10 +101,18 @@ class Proxy(solverbase.Proxy):
 
         obj.addProperty(
             "App::PropertyIntegerList",
+            "OutputIntervals",
+            "Timestepping",
+            "After how many time steps a result file is output"
+        )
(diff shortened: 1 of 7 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9043">pipeline status</a> for feature branch [PR_9043](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9043). Pipeline [817742406](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817742406) was triggered at [58652d8](https://github.com/FreeCAD/FreeCAD/pull/9043/commits/58652d812fe2ea02250abeca4225052d0b1178fa). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #9044, opened 2023-03-25
[WIP] Sketcher : Circle to Line distance constraint

FlachyJoe:
Implements Circle to Line distance constraint.


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044">pipeline status</a> for feature branch [PR_9044](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044). Pipeline [817748548](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817748548) was triggered at [5862a20](https://github.com/FreeCAD/FreeCAD/pull/9044/commits/5862a20caa0fc12843d55f6c5270f8254bb5bc59). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/App/planegcs/Constraints.cpp
```diff
@@ -2845,4 +2845,86 @@ double ConstraintC2CDistance::grad(double *param)
...
+
+    double u = -ct.x * v_line.y + ct.y * v_line.x + p1.x * p2.y - p2.x * p1.y; //center to line distance
+    double du = -ct.dx*v_line.y - ct.x * v_line.dy + ct.dy * v_line.dx + ct.y * v_line.dx
+                + p1.dx * p2.y + p1.x * p2.dy  - p2.dx * p1.y - p2.x * p1.dy;
+
+    double dv;
+    double v = v_line.length(dv);
+    double h = u / v;
+    double dh = (du * v - u * dv) / (v * v);
+
```
abdullahtahiriyo, line 2900: Please, try to use better names for the variables and/or put some comments about what exactly is each. Also some comments about what you are calculating would be helpful. (Yes, I know we have these elsewhere without comments, but we are trying to improve...)

This kind of vector algebra can generally be written using DeriVector2 is a clearer way (for the one reading the code).

src/Mod/Sketcher/App/planegcs/Constraints.cpp
```diff
@@ -2845,4 +2845,86 @@ double ConstraintC2CDistance::grad(double *param)
...
+    if (err) {
+        *err = h - *circle.rad - *distance();
+    }
+    else if (grad) {
+        *grad = dh;
+        if ( param == distance() || param == circle.rad) {
+            *grad -= 1.0;
+        }
```
abdullahtahiriyo, line 2908: I do not know if the grad is well calculated, but if it is, the *grad = dh should go as an else of the if.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -2336,6 +2336,35 @@ void CmdSketcherConstrainDistance::activated(int iMsg)
             else
                 finishDatumConstraint (this, Obj, true);
 
+            return;
+        } else if (geom1->getTypeId() == Part::GeomCircle::getClassTypeId()
+        && geom2->getTypeId() == Part::GeomLineSegment::getClassTypeId() ) { // circle to line distance
+            auto circleSeg = static_cast<const Part::GeomCircle*>(geom1);
```
abdullahtahiriyo, line 2342: Maybe I missed something, but it appears that you are not checking a case where geom1 is the line and geom2 is the circle (so the user clicked in the other order).

If it is not handled, look around for some checking and a call to std::swap, so that you are sure geom1 is the circle when arriving here.

abdullahtahiriyo:
Just left some comments from fast scrolling the code. I would welcome if anybody else could test and do a review. I am tight on time.

FlachyJoe:
> anybody else could test

It doesn't work yet has the solver doesn't converge !

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044">pipeline status</a> for feature branch [PR_9044](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044). Pipeline [817952571](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817952571) was triggered at [ce6f19a](https://github.com/FreeCAD/FreeCAD/pull/9044/commits/ce6f19a4408d56f945116035cfa73ba19a2e4af2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044">pipeline status</a> for feature branch [PR_9044](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044). Pipeline [817962240](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/817962240) was triggered at [95971d0](https://github.com/FreeCAD/FreeCAD/pull/9044/commits/95971d09c8a505543bc4306de668efa90f021dc6). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

abdullahtahiriyo:
Ok. Let us know.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044">pipeline status</a> for feature branch [PR_9044](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044). Pipeline [818208178](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/818208178) was triggered at [e300b4d](https://github.com/FreeCAD/FreeCAD/pull/9044/commits/e300b4d91297401b0c362534a1147f4c54d5a30e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

FlachyJoe:
This rainy Sunday lets me work on this PR :)
Ready for review !

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044">pipeline status</a> for feature branch [PR_9044](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044). Pipeline [818268035](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/818268035) was triggered at [63a7f2f](https://github.com/FreeCAD/FreeCAD/pull/9044/commits/63a7f2f1c932ed7a3bb1cefc4b4f9f51523f5e4e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
> This rainy Sunday lets me work on this PR :) Ready for review !

So please remove "[WIP]" from title. ;)

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044">pipeline status</a> for feature branch [PR_9044](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044). Pipeline [819466709](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/819466709) was triggered at [03b0a3a](https://github.com/FreeCAD/FreeCAD/pull/9044/commits/03b0a3a585f0084e452a71ac954061af6f1fce90). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044">pipeline status</a> for feature branch [PR_9044](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044). Pipeline [820792356](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/820792356) was triggered at [65d6b89](https://github.com/FreeCAD/FreeCAD/pull/9044/commits/65d6b89e46aecf3a16ba836b1fcdf38b4107f951). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044">pipeline status</a> for feature branch [PR_9044](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9044). Pipeline [820795437](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/820795437) was triggered at [66e5e21](https://github.com/FreeCAD/FreeCAD/pull/9044/commits/66e5e21ceec16a8ba62565e064fc53245a10513b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

jbaehr:
I suggest to add automated tests in `src/Mod/Sketcher/SketcherTests`, too. They not only demonstrate the new functionality, its accessibility through python, and which corner cases are already handled, but also ensure that stuff doesn't get broken in the future.

And once getting used to, this can also speed up your development: You can execute many scenarios with a simple command line instead of trying them out manually one my one and over and over again.

FlachyJoe:
@jbaehr could you help with test writing?

jbaehr:
> @jbaehr could you help with test writing?

Sure. In which way do you prefer getting help? Is there something specific you have in mind?

FlachyJoe:
@jbaehr I have no idea of what to check to be certain it works…

jbaehr:
Finally I put some first rudimentary integration tests together for this feature, cf. https://github.com/jbaehr/FreeCAD/commit/3ffb96d90a8a9dcc16aa985f8db8b911c17f52cf

To execute the sketcher tests only you can use the following command line
```sh
./build-pr9044/bin/FreeCAD -c -M FreeCAD-git/src/Mod/Sketcher -t TestSketcherApp.TestSketcherSolver
```
Replace `build-pr9044/bin/FreeCAD` by the path to your FreeCAD executable and `FreeCAD-git/` by the location of your source tree relatively to your current working dir.
`-c` (`--console`) avoids starting the GUI, which means you can run your tests *much* faster.
`-M <path>` (`--module-path <path>`) allows you to prepend a path to FreeCAD's module search path. I use this when working on python-only stuff to not having to recompile. You can omit this, but then you have to deploy your changed .py-files manually or you have to work outside of your source tree.
`-t <test case>` (`--run-test <test case>`) tells FreeCAD to execute the specified test case.

Note that in the first of the new tests, `testCircleToLineDistance_Driving_Passant`, the solver fails to converge.
Also note that I'm not yet fluent with sketcher scripting, so there is certainly room for improvement. I also commented on some behaviour I do not fully understand -- especially the test for the reference constraint is somewhat cumbersome.
@abdullahtahiriyo or @0penBrain do you know why `solve()` moves the geometry if there is no driving constraint, only a reference one? The added "Block" constraints prevent this but it feels like a hack. When using the GUI this issue is hidden by the fact that also non-driving constraints are initialized with a pre-calculated value.

luzpaz:
@FlachyJoe same request as https://github.com/FreeCAD/FreeCAD/pull/9257#issuecomment-1510513769

0penBrain:
> @abdullahtahiriyo or @0penBrain do you know why `solve()` moves the geometry if there is no driving constraint, only a reference one?

I have no clue about that. 😄

abdullahtahiriyo:
@jbaehr 
> Note that in the first of the new tests, `testCircleToLineDistance_Driving_Passant`, the solver fails to converge.
> Also note that I'm not yet fluent with sketcher scripting, so there is certainly room for improvement. I also commented on some behaviour I do not fully understand -- especially the test for the reference constraint is somewhat cumbersome.
> @abdullahtahiriyo or @0penBrain do you know why `solve()` moves the geometry if there is no driving constraint, only a reference one? The added "Block" constraints prevent this but it feels like a hack. When using the GUI this issue is hidden by the fact that also non-driving constraints are initialized with a pre-calculated value.

Reference constraints are added to the system. The non-driving datum sets the initial value for the solver. Being non-driving the value can change (as opposed to when it is driving, when it does not change). However, there is no guarantee that only that parameter will change during a call to solve. If fact, it could be that we have geometry with unmet driving constraints and other parameters need to change to. The only guarantee is that at the end of the solve, if it converges, all constraints will be met. However, any valid solution meeting them is a valid solution.

Your solution is the solution power user use to circumvent unwanted change.

Would you please PR those unit tests? I wish we all did more of these (and that includes me particularly).

abdullahtahiriyo:
It is a plan. Thank you.

I did not understand the solver was failing to converge. I misunderstood that the test was failing because of moving geometry. I will look into it with your PR. Thanks.

Is "getDatum" what you are looking for?

```
ActiveSketch.getDatum(1)
26.26 mm
```

jbaehr:
> Is "getDatum" what you are looking for?
> 
> ```
> ActiveSketch.getDatum(1)
> 26.26 mm
> ```

No, this can only request the datum of an already existing constraint, the one with index 1 in your example. I was looking for a way to "measure" e.g. a current distance in order to determine the initial value for  new constraint.
When I create a new datum constraint using the GUI, the constraint's value is initialized with a specific datum already which correspond to the current, e.g. distance. When using the Python API, I am supposed to provide this value. But where to I get it from?

Something like this:
```py
# c_idx and l_idx being the geometry index of a circle and a line
current_distance = ActiveSketch.getDatum('Distance', c_idx, l_idx) # missing API? the GUI provides this "magically"
ActiveSketch.addConstraint(Sketcher.Constraint('Distance', c_idx, l_idx, current_distance)) # existing API requires the datum to be known
```

abdullahtahiriyo:
ActiveSketch is SketchObject. If the constraint has not been added yet to the sketch (to the property), SketchObject knows nothing about it.

The flow is:
The UI command calculates the initial value, may ask the user for a value, later on it issues a Python call creating the constraint and adding it to the sketch. Then the sketch is solved.

There is currently no c++ interface to expose that temporal pre-calculated value (let alone Python).

abdullahtahiriyo:
I am failing to understand what did I check exactly, but obviously there are cases where this fails.

This is the FreeCAD macro version of the unit test that fails:

```
sketch = ActiveSketch
radius = 20
xy_normal = App.Vector(0,0,1)
center = App.Vector(0, 0, 0)
circle = Part.Circle(center, xy_normal, radius)
line = Part.LineSegment(App.Vector(-radius, 2*radius,0), App.Vector(radius, 2*radius, 0))
c_idx = sketch.addGeometry(circle)
l_idx = sketch.addGeometry(line)
# use a positive distance, other than the initial distance of the line
wanted_distance = radius/2
sketch.addConstraint(Sketcher.Constraint('Distance', c_idx, l_idx, wanted_distance))
ActiveSketch.solve()
```

In fact, even when it does not fail because of lack of convergence, the result is often no draggable despite DoFs.

abdullahtahiriyo:
I have taken a closer look, but do not have enough time to test now.

The test passes with this function:
```
void ConstraintC2LDistance::errorgrad(double *err, double *grad, double *param)
{
    if (pvecChangedFlag) ReconstructGeomPointers();

    DeriVector2 ct (circle.center, param);
    DeriVector2 p1 (line.p1, param);
    DeriVector2 p2 (line.p2, param);
    DeriVector2 v_line = p2.subtr(p1);
    DeriVector2 v_p1ct = ct.subtr(p1);

    //center to line distance (=h) and its derivative (=dh)
    double darea = 0.0;
    double area = v_line.crossProdNorm(v_p1ct, darea); //parallelogram oriented area

    double dlength;
    double length = v_line.length(dlength);

    // vector product (cross vector) has a magnitude corresponding to the area of
    // the parallelogram defined by the vectors above. The area of the triangle is
    // half the parallelogram area. The height of the triangle is the area divided by
    // the base, which is the distance from the center of the circle to the line.
    //
    // However, the vector (which points in z direction), can be positive or negative.
    // the area is the absolute value
    double h = std::abs(area) / length;

    // darea is the magnitude of a vector in the z direction, which makes the area vector
    // increase or decrease. If area vector is negative a negative value makes the area increase
    // and a positive value makes it decrease.
    darea = std::signbit(area) ? -darea : darea;

    double dh =  (darea - h * dlength) / length;

    if (err) {
        *err = *distance() + *circle.rad - h;
    }
    else if (grad) {
        if ( param == distance() || param == circle.rad) {
            *grad = 1.0;
        } else {
            *grad = -dh;
        }
    }
}

```

https://github.com/abdullahtahiriyo/FreeCAD_sf_master/tree/circle-to-line_tests

Anybody can give it some testing?

0penBrain:
> The test passes with this function:

Hmmm. The CI failed on your branch. Is it expected ?

> Anybody can give it some testing?

What kind of tests are you expecting ? ;)

abdullahtahiriyo:
The first test, which was failing, passes.

It is ok if somebody does some coverage tests, to ensure the behaviour is ok.

The second test (a secant) does not pass. In the morning I draw a secant line and it converged fine.

Here I do not know if the test is expected something unreasonable, or if the test rightfully fails. Reproducing the test (rewriting it as a macro) would allow to see if the test fails because the tool misbehaves, or there is something wrong with the test.

Any testing help is welcome. 

My objective is to come ensure this time the tool works at it should and the tests pass (either by changing the behaviour or by changing the tests, whatever is appropriate).

I will try to come back to it at the end of the day or tomorrow.

abdullahtahiriyo:
I remember now (very long day). 

When the line is secant, the proposed distance is negative. The sketcher does not accept a negative datum constraint. 

I understood that the wanted behaviour is that distances are not negative, and thus a negative distance leads to freecad refusing to constraint the line (at the end the distance is zero if it is secant). I think this is coherent with the circle to circle distance, where it refuses secant circles.

This is a user imposed limitation (as the solver can work with negative values). If I remember correctly, in the circle to circle, I indicated what needed to be changed if it was desirable, but it was decided not to.

I do not have a strong opinion and I welcome feedback.

jbaehr:
> When the line is secant, the proposed distance is negative.

Please note that this was the pre-existing behaviour. I first played around a bit with this new feature in the GUI and then distilled some very basic tests (passant, secant, driving, reference), which I poured into code.

I guess I hit a not-yet-covered corner here: To make the test as simple as possible, I placed the circle at the origin and a horizontal line on top, symmetric to the Y-axis. Only for this particular case the solver fails to converge -- if you tilt the line just a little bit, it works, e.g.
```diff
- line = Part.LineSegment(App.Vector(-radius, 2*radius,0), App.Vector(radius, 2*radius, 0))
+ line = Part.LineSegment(App.Vector(-radius, radius,0), App.Vector(radius, 2*radius, 0))
```
(but just changing the tests to not detect an error anymore is not the goal ;-))

> The sketcher does not accept a negative datum constraint.

This is not meant as a universal statement, right? For a point, negative horizontal and vertical constraints do work; e.g. for a point in the third quadrant. (to create the constraint, select only the point; not point and axis!)

For the circle-to-line distance, you get negative values in case of a secant. You can create such constraints when they are non-driving. It continues to work when changing it to driving later, but the Sketcher-GUI does not permit any changes to the value then (still works via the property panel). This feels somewhat inconsistent.

jbaehr:
> The flow is: The UI command calculates the initial value, may ask the user for a value, later on it issues a Python call creating the constraint and adding it to the sketch. Then the sketch is solved.
> 
> There is currently no c++ interface to expose that temporal pre-calculated value (let alone Python).

I see. This "The UI command calculates the initial value" was what I saw, but I have hoped this would also be possible via Python. Is this already on some todo-list? Otherwise I'll create a ticket.

abdullahtahiriyo:
> This is not meant as a universal statement, right? For a point, negative horizontal and vertical constraints do work; e.g. for a point in the third quadrant. (to create the constraint, select only the point; not point and axis!)
> 
> For the circle-to-line distance, you get negative values in case of a secant. You can create such constraints when they are non-driving. It continues to work when changing it to driving later, but the Sketcher-GUI does not permit any changes to the value then (still works via the property panel). This feels somewhat inconsistent.

Horizontal and Vertical distances are directional. Absolute distance is not. I did not insert this limitation myself, but I have some vague memory that it was intended behaviour (at least for lines). As said, it is not a solver limitation (at least for this constraint), but an imposed one:
https://github.com/FreeCAD/FreeCAD/blob/6f8e3f374f7dead4a373e1114ab4cc30785d8daa/src/Mod/Sketcher/App/SketchObject.cpp#L349

I raised this issue with the circle-to-circle distance. However, it seemed to me that there was no interest to allow distances when two circles are secant. From this point of view, the circle-to-line is consistent with the circle-to-circle.

Many behaviours are possible. It is possible not to allow negative distances for the constraint when toggling back to driving. It is possible to remove the restriction (but we have then to consult users if we change the behaviour of the normal point to distance constraint and ensure it does not break the normal work-flow).

> Please note that this was the pre-existing behaviour.

The pre-exiting implementation had some bugs. The constraint works by calculating the area of the triangle defined by the endpoints of the lines and one endpoint to the center of the circle, then dividing this area by the base length (giving the height of the triangle, which is the distance, and then correcting this distance with the constraint value and circle radius). The area is calculated using the geometric property of the vector product (cross product). While the absolute value of the magnitude is the area, the vector can point to -z or z, the differential was calculated without taking into account this. Depending on the direction of the vectors (the positions of the start and end points), the constraint would fail to converge.

> I see. This "The UI command calculates the initial value" was what I saw, but I have hoped this would also be possible via Python. Is this already on some todo-list? Otherwise I'll create a ticket.

I am not aware of any such effort or plan.

FlachyJoe:
> ```
> // darea is the magnitude of a vector in the z direction, which makes the area vector
>     // increase or decrease. If area vector is negative a negative value makes the area increase
>     // and a positive value makes it decrease.
>     darea = std::signbit(area) ? -darea : darea;
> ```

I thought this was done with std::copysign


> For the circle-to-line distance, you get negative values in case of a secant. You can create such constraints when they are non-driving. It continues to work when changing it to driving later, but the Sketcher-GUI does not permit any changes to the value then (still works via the property panel). This feels somewhat inconsistent.

I agree.


> This is the FreeCAD macro version of the unit test that fails:

I currently have a compiled version (based on 32540) which works well but it seems I lost the corresponding code ;-(

jbaehr:
Thanks for your extensive explanation -- highly appreciated!
This explains why initially `testCircleToLineDistance_Driving_Passant` failed to converge and `testCircleToLineDistance_Driving_Secant` passed, and with your recent change exactly the opposite happens.

What do you think about moving the discussion about the *wanted* behaviour to the forum? Personally I see great use of allowing the new constraint to handle secants to the circle (as well as intersecting circles). My view is limited, though, and seeing more opinions can make the final solution only better.

For getting access to the initial constraint value via python, I'll create an issue then. Maybe I'll give it a shot myself, one day...

0penBrain:
@abdullahtahiriyo Are you going to do a PR from your fix branch ?

abdullahtahiriyo:
> I'm sorry for the bad quality of this code. I would be happy if my next PRs was checked twice.

I did fail to test the original enough before merging. That is on me, as that is my task. 

Your code was not of bad quality. It had some bugs. When the bugs got revealed, we managed to give an appropriate response together. We all make FreeCAD. That includes documentalists, testers, reporters, developers and mantainers.

@0penBrain 
@FlachyJoe 
@jbaehr 

I would really like to have a set of working unit tests:

https://github.com/FreeCAD/FreeCAD/pull/9474

I understand that we should discuss whether we should support or not the secant case. However, that should be postponed for the 0.22 development cycle, as it may impact point to line distance. If we do not get that one right it would be a major regression without enough testing time before release.

Do you think we could comment out the secant test to move forward with the merge?

FlachyJoe:
@abdullahtahiriyo 
> it may impact point to line distance

I can't see how point to line distance can be negative.


> Do you think we could comment out the secant test to move forward with the merge?

IMHO secant case is an error and we shouldn't allow such constraint to exist as it doesn't fit to the distance definition we follow (=minus distance from each points of the two geometries). NB : it's also the case for two secant circles distance.
So, I think allowing it only as reference is good enough for now and it has not to be tested.

0penBrain:
> Do you think we could comment out the secant test to move forward with the merge?

No problem that's fine. We currently have known limitations and with minimal documentation it's perfectly OK.

FlachyJoe:
FYI I've implemented the positive distance for a point in a circle in #9559 as this case follow the above definition.

jbaehr:
> Do you think we could comment out the secant test to move forward with the merge?

I suggest to [skip](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) the test with an appropriate message instead of commenting it out.

0penBrain:
@jbaehr just to be clear, are you taking care of doing that ?

jbaehr:
> @jbaehr just to be clear, are you taking care of doing that ?

Yes. I'll update PR #9474 accordingly.

abdullahtahiriyo:
@jbaehr 

Thank you so much. I have merged it.

PaddleStroke:
I may be wrong, but it seems that you forgot to add the necessary code in `CmdSketcherConstrainDistance::applyConstraint` to support line-circle distance.

So the line-distance constraint works when you start distance tool then select. But it does not work when you start tool, then select geometries.

Besides in the workflow tool->selection, circle-circle distance doesn't work either.

0penBrain:
This is a (known) limitation of the continuous mode. The first selection got "eaten".
We have to deeply refactor the continuous mode to make it work.

PaddleStroke:
Is it? I guess that if you select a line first you get a distance, but if you select the circle first ?

There is a `case 8:// {SelEdge, SelEdge}` in the applyConstraint function that has been implemented that suggest that it should work.

Do you mean that when you select a first edge, then `case 2:// {SelEdge}` is triggered, which go to : 
```
            else if (geom->getTypeId() == Part::GeomCircle::getClassTypeId()) {
                // allow this selection but do nothing as it needs 2 circles or 1 circle and 1 line
            }
```
And that once this is done the selection is cleared so it is actually impossible de select 2 circles? In this case this comment and code are confusing and should be removed no?

0penBrain:
That's it. `case 2` resets the selection so `case 8` is actually (currently) unreachable.

PR #8862, opened 2023-03-12
Fix text encoding handling in C++ DXF importer

kpemartin:
This addresses part of [Issue #8704](https://github.com/FreeCAD/FreeCAD/issues/8704), relating to the C++ DXF importer. The Issue also relates to the legacy importer, where [a separate PR](https://github.com/yorikvanhavre/Draft-dxf-importer/pull/24) has been placed in the appropriate repository.

The fix analyzes the variables at the start of the DXF file to determine the appropriate encoding for text which occurs later in the file, and applies that encoding as required.

- [X]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Issue #8704, opened 2023-02-28
[Problem] DXF files produced by older AutoCAD versions where text contains non-ASCII characters cannot be imported/opened by either the legacy or non-legacy DXF importer.


kpemartin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.20 (Release)

### Full version info

```shell
OS: Windows 7 Version 6.1 (Build 7601: SP 1)
Word size of FreeCAD: 64-bit
Version: 0.20.2.29177 +426 (Git)
Build type: Release
Branch: (HEAD detached from 0.20.2)
Hash: 930dd9a76203a3260b1e6256c70c1c3cad8c5cb8
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/United States (en_US)

Also: AutoCAD version 13 (yes, it is that old)
```

### Subproject(s) affected?

None

### Problem description

Per the DXF format documentation at https://help.autodesk.com/view/OARX/2018/ENU/?guid=GUID-2553CF98-44F6-4828-82DD-FE3BC7448113 ("Storage of String Values" section), text in the DXF file is encoded as UTF-8 for versions 2007 and beyond ($ACADVER >= "AC1021"), but as "Plain ASCII" [sic] for versions 2004 and previous ($ACADVER <= "AC1018").

In the latter case, examination of the contents of the DXF file reveal that the encoding is not really "ASCII" because it contains bytes larger then 0x7F. Such characters can be entered using ALT+0ddd keyboarding on the numeric keypad or by pasting from another program such as Character Map, or dimension text for angular dimensions can contain a degree sign. These strings can also contain in-band encoding syntax \U+xxxx to select Unicode code points above 0xFF, a feature only useful if the text font was one that supported wide character sets.

![Autocad Screen.png](https://user-images.githubusercontent.com/4140247/222018160-e7013f7b-b30d-444c-884d-94fdcea223a0.png)

The way AutoCAD renders characters above 0x7F on the screen indicates that the values are actually ISO-6659-1 (or perhaps Windows-1252) with no encoding (i.e. raw bytes). Thus the byte 0xB0 in the text in the sample DXF file renders as a degree sign, and the two bytes 0xC2 0xB0, which are the bytes of the UTF-8 encoding for the degree sign (0xB0), render as an accented uppercase A and (somewhat coincidentally) a degree sign.

[...]

PR #9060, opened 2023-03-26
Fix encoding error for import from older DXF files

kpemartin:
This addresses #8704 and replaces #8862 which I accidentally closed while trying to clean up.

The C++ importer incorrectly treated the contents of all TEXT and MTEXT objects as beind encoded as UTF-8, but this is not true for DXF files before AutoCAD 2007, where the encoding is "plain ASCII" plus some in-band \U+dddd encoding. This would cause errors if the text contained non-ASCII characters such as the Degree Sign.

This change causes the correct encoding to be used.

### DIFF
--- a/src/Mod/Import/App/dxf/dxf.cpp
+++ b/src/Mod/Import/App/dxf/dxf.cpp
@@ -1750,7 +1750,7 @@ CDxfRead::CDxfRead(const char* filepath)
     memset( m_str, '\0', sizeof(m_str) );
     memset( m_unused_line, '\0', sizeof(m_unused_line) );
     m_fail = false;
-    m_aci = 0;
+    m_ColorIndex = 0;
     m_eUnits = eMillimeters;
     m_measurement_inch = false;
     strcpy(m_layer_name, "0");  // Default layer name
@@ -1841,7 +1841,7 @@ bool CDxfRead::ReadLine()
         switch(n){
             case 0:
                 // next item found, so finish with line
-                    DerefACI();
+                    ResolveColorIndex();
                     OnReadLine(s, e, hidden);
                     hidden = false;
                 return true;
@@ -1889,7 +1889,7 @@ bool CDxfRead::ReadLine()
                 case 62:
                 // color index
                 get_line();
-                ss.str(m_str); ss >> m_aci; if(ss.fail()) return false;
+                ss.str(m_str); ss >> m_ColorIndex; if(ss.fail()) return false;
                 break;
 
             case 100:
@@ -1908,7 +1908,7 @@ bool CDxfRead::ReadLine()
     }
 
     try {
-        DerefACI();
+        ResolveColorIndex();
         OnReadLine(s, e, false);
     }
     catch(...)
@@ -1939,7 +1939,7 @@ bool CDxfRead::ReadPoint()
         switch(n){
             case 0:
                 // next item found, so finish with line
-                    DerefACI();
+                    ResolveColorIndex();
                 OnReadPoint(s);
                 return true;
 
@@ -1967,7 +1967,7 @@ bool CDxfRead::ReadPoint()
                 case 62:
                 // color index
                 get_line();
-                ss.str(m_str); ss >> m_aci; if(ss.fail()) return false;
+                ss.str(m_str); ss >> m_ColorIndex; if(ss.fail()) return false;
                 break;
 
             case 100:
@@ -1987,7 +1987,7 @@ bool CDxfRead::ReadPoint()
     }
 
     try {
-        DerefACI();
+        ResolveColorIndex();
         OnReadPoint(s);
     }
     catch(...)
@@ -2022,7 +2022,7 @@ bool CDxfRead::ReadArc()
         switch(n){
             case 0:
                 // next item found, so finish with arc
-                    DerefACI();
+                    ResolveColorIndex();
                     OnReadArc(start_angle, end_angle, radius, c,z_extrusion_dir, hidden);
                     hidden = false;
                     return true;
@@ -2070,7 +2070,7 @@ bool CDxfRead::ReadArc()
                 case 62:
                 // color index
                 get_line();
-                ss.str(m_str); ss >> m_aci; if(ss.fail()) return false;
+                ss.str(m_str); ss >> m_ColorIndex; if(ss.fail()) return false;
                 break;
 
 
@@ -2094,7 +2094,7 @@ bool CDxfRead::ReadArc()
                 break;
         }
     }
-    DerefACI();
+    ResolveColorIndex();
     OnReadArc(start_angle, end_angle, radius, c, z_extrusion_dir, false);
     return false;
 }
@@ -2127,7 +2127,7 @@ bool CDxfRead::ReadSpline()
         switch(n){
             case 0:
                 // next item found, so finish with Spline
-                    DerefACI();
+                    ResolveColorIndex();
                 OnReadSpline(sd);
                 return true;
             case 8: // Layer name follows
@@ -2137,7 +2137,7 @@ bool CDxfRead::ReadSpline()
                 case 62:
                 // color index
                 get_line();
-                ss.str(m_str); ss >> m_aci; if(ss.fail()) return false;
+                ss.str(m_str); ss >> m_ColorIndex; if(ss.fail()) return false;
                 break;
             case 210:
                 // normal x
@@ -2275,7 +2275,7 @@ bool CDxfRead::ReadSpline()
                 break;
         }
     }
-    DerefACI();
+    ResolveColorIndex();
     OnReadSpline(sd);
     return false;
 }
@@ -2301,7 +2301,7 @@ bool CDxfRead::ReadCircle()
         switch(n){
             case 0:
                 // next item found, so finish with Circle
-                DerefACI();
+                ResolveColorIndex();
                 OnReadCircle(c, radius, hidden);
                 hidden = false;
                 return true;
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9060">pipeline status</a> for feature branch [PR_9060](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9060). Pipeline [818223830](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/818223830) was triggered at [b387a3c](https://github.com/FreeCAD/FreeCAD/pull/9060/commits/b387a3c9f6af160bf2166e12f97b1249752d7f9e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

yorikvanhavre:
This looks fine, but it's best not to use the word "autocad" (or any other commercial product) anywhere in FreeCAD... Just to be on the safe side. Could you change that?

kpemartin:
Hmmm, the only place I added this to any code is the name of the type eAutoCADVersion_t so I can rename this to refer to "DXF" instead.
In the two changed files that product is also mentioned in code I did not change, in a comment explaining what "aci" means. Should I take that out as well? I think I would rename some things at the same time to remove the cryptic "aci" and instead use "colorIndex"

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9060">pipeline status</a> for feature branch [PR_9060](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9060). Pipeline [818915422](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/818915422) was triggered at [dbf1737](https://github.com/FreeCAD/FreeCAD/pull/9060/commits/dbf17376b76778b780374fd7d6fff3eb34d17b8e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

yorikvanhavre:
yes that seems like an excellent idea... Thanks!

yorikvanhavre:
that would be perfect. Thanks!!

yorikvanhavre:
damn.. github acting strange today :sweat_smile:

yorikvanhavre:
Sure! you'll find my email address at the top of any of the Arch module files. But it's cool to discuss here too, it can always help somebody else in the same situation...
Don't worry, we all screwed our entire git tree a time or two at the beginning. The joices of working with Git :smile:
At some point if things are very bad it is often easier to extract a diff of your changes, make a new clean branch, and reapply your patch to that branch.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9060">pipeline status</a> for feature branch [PR_9060](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9060). Pipeline [820542013](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/820542013) was triggered at [3f195bb](https://github.com/FreeCAD/FreeCAD/pull/9060/commits/3f195bb46b982b3e6fb64d18cb37dd1192fd23ba). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

kpemartin:
@yorikvanhavre  Hmmm, maybe you had something with "GitHub is acting strange today". A "git push" that failed yesterday worked today, and I don't think I did anything to my local repository in the interim.

It was giving me:
```
 ! [remote rejected]   master -> master (failure)
error: failed to push some refs to 'https://github.com/kpemartin/FreeCAD.git
```
I had done a `git fetch` before the successful push, but that appeared to do nothing (no output, ran instantaneously).


Anyway, I think this PR is ready to merge now, assuming it passes all the checks.

I intend to do a separate PR to clean up some of the pre-existing lint that has shown up in the checks for this PR.

yorikvanhavre:
This looks good to me... @wwmayer could you have a quick look at the python code used here inside, if it looks OK to you?

src/Mod/Import/App/dxf/dxf.cpp
```diff
@@ -3271,7 +3280,118 @@ bool CDxfRead::ReadLayer()
...
+    }
+    else {
+        // Names may be of the form "ansi_1252" which we map to "cp1252" but we don't map "ansi_x3xxxx" which means "ascii"
+        if (strncmp(m_CodePage->c_str(), "ansi_", 5) == 0
+            && strncmp(m_CodePage->c_str(), "ansi_x3", 7) != 0) {
+            std::string* p = new std::string(*m_CodePage);
+            p->replace(0, 5, "cp");
+            m_encoding = p;
+        }
+        else
```
wwmayer, line 3349: The `if` has curly braces, so add it to the `else` branch too, please.

src/Mod/Import/App/dxf/dxf.cpp
```diff
@@ -3271,7 +3280,118 @@ bool CDxfRead::ReadLayer()
...
+            m_encoding = new std::string(*m_CodePage);
+        // At this point we want to recognize synonyms for "utf_8" and use the custom decoder function.
+        // This is because this is one of the common cases and our decoder function is a fast no-op.
+        // We don't actually use the decoder function we get from PyCodec_Decoder because to call it we have to convert the (char *) text into
+        // a 'bytes' object first so we can pass it to the function using PyObject_Callxxx(), getting the PYObject containing the
+        // Python string, which we then decode back to UTF-8. It is simpler to call PyUnicode_DecodeXxxx which takes a (const char *)
+        // and is just a direct c++ callable.
```
wwmayer, line 3356: Before using any Python functions you must lock the GIL. Do this by adding the line
`Base::PyGILStateLocker lock;`

src/Mod/Import/App/dxf/dxf.cpp
```diff
@@ -3271,7 +3280,118 @@ bool CDxfRead::ReadLayer()
...
+    }
+    return m_encoding != NULL;
+}
+const char* CDxfRead::UTF8ToUTF8(const char* encoded) const
+{
+    return encoded;
+}
+const char* CDxfRead::GeneralToUTF8(const char* encoded) const
+{
+    PyObject* decoded = PyUnicode_Decode(encoded, strlen(encoded), m_encoding->c_str(), "strict");
```
wwmayer, line 3377: Lock GIL here too

src/Mod/Import/App/dxf/dxf.cpp
kpemartin: If you insist I will, but I personally abhor unnecessary braces (in particular, the extra lines which they occupy, meaning less code fits on a screen)

src/Mod/Import/App/dxf/dxf.cpp
kpemartin: How much performance penalty does this have? Say, compared to the cost of the call(s) that it is protecting?

Just asking because, if it is very expensive (either in processor or real time) I might consider some other encoding conversion method. My assumption is that this asserts some sort of lock, so the only impact would be if garbage collection is in progress, there would be a real-time delay waiting for it to complete?

src/Mod/Import/App/dxf/dxf.cpp
kpemartin: Actually it is moot now as I've recoded it to not need braces at all.

However, for future reference, is there a document describing style guidelines?

Does that apply in the opposite, i.e. does
```
if (condition)
    simple statement
else {
    many statements
    on several lines
}
```
need braces around the `simple statement`?

What about a chain of `else if`s, does one only have to use braces after the first compound statement?

What about if the `simple statement` is so long that it is wrapped onto multiple lines?

src/Mod/Import/App/dxf/dxf.cpp
kpemartin: Locking code added

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9060">pipeline status</a> for feature branch [PR_9060](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9060). Pipeline [822108934](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/822108934) was triggered at [31f3458](https://github.com/FreeCAD/FreeCAD/pull/9060/commits/31f345839967178e1cf1c59a59fe78003e1589ea). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Import/App/dxf/dxf.cpp
wwmayer: In many code style guidelines there is a rule that if a branch of an if-else block requires curly braces then it should be added to all of them.

src/Mod/Import/App/dxf/dxf.cpp
wwmayer: No, it's not very expensive. If the GIL is not locked and the used Python function requires it then Python will call abort() and crashes the application.

yorikvanhavre:
Thanks all for the work and reviewing! I think this is good to merge then!

yorikvanhavre:
You've met "It's done when it's done" , official FreeCAD motto :)

PR #9083, opened 2023-03-27
CI: fix issue in cleanup workflow

0penBrain:
Fix an issue failing cleanup workflow when date has no decimal part for seconds


0penBrain:
OK, cleaned. Ready to merge. :)

PR #9104, opened 2023-03-28
Techdraw: finish Qt6 port

wwmayer:

---

### DIFF
--- a/src/Mod/TechDraw/App/CMakeLists.txt
+++ b/src/Mod/TechDraw/App/CMakeLists.txt
@@ -31,9 +31,6 @@ if(FREECAD_QT_MAJOR_VERSION EQUAL 5)
     include_directories(
         ${QtXmlPatterns_INCLUDE_DIRS}
     )
-else()
-    # This is deprecated as of Qt 5.13, and removed in Qt 6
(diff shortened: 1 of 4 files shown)


chennes:
Was this the last thing that we were waiting on for full Qt6 support?

wwmayer:
Yes, I think so.

However, we still have tons of slot functions of this form `on_objectname_signalname`  that are automatically connected with QMetaObject::connectSlotsByName. Since several signals have been renamed in Qt6 some automatic connections may fail.

wwmayer:
BTW, the algorithm for Qt6 can be easily tested with Qt5. Therefore replace `#if QT_VERSION < QT_VERSION_CHECK(6,0,0)` with `#if 0` in XMLQuery.cpp. When creating a TD page you will see the green bullets where you can change the editable fields.

Now if you comment out the line `processElements` inside `processItems` you will realize that the green bullets are missing.

wwmayer:
> Should we be eliminating the automatic connections and doing it manually?

This is what clazy recommends. I don't have a commit right now that shows the warning but you have already seen it for sure that clazy says that the automatic connections are error-prone. IMO, it's similar to the old-style connect with SIGNAL and SLOT which has raised many issues with the Qt6 port.

So, as a long-term goal we should migrate to the manual connection but this is not of high priority.

PR #9107, opened 2023-03-28
[App] Detect if file was modified externally before saving

### LINKED ISSUE #8924: [Problem] FreeCAD will happily overwrite a changed file

### Version

0.21 (Development)

### Full version info

```shell
any
```

### Subproject(s) affected?

Core

### Problem description

 The user has a document open and has made changes.  
Meanwhile the same document is opened in another instance of FreeCAD, changed, and saved.  
Then the first instance tries to save, FreeCAD happily overwrites the changes.  No warning.

FreeCAD should detect that the file on the disk has changed since it was opened and warn the user before overwriting the file.  

### Anything else?

_No response_

### Code of Conduct

AjinkyaDahale:
Proof of concept to fix #8924.

Some considerations:
1. Can we avoid creating a new attribute? Instead of `PropertyString LastModifiedDate` we could have something like `PropertyTimeStamp LastModifiedTime`.
2. The timestamps are only accurate up to a second.
3. What if the file is externally replaced by an older copy for some reason.
4. Is `std::chrono` fine or we need a method in `Base::TimeInfo`?
5. How this manifests in the various ways FC is used: GUI, CLI, and without interface. For GUI there's a mechanism that catches these exceptions and offers to save with a different name instead. For CLI the same could be used. However, when being used without an interface (through a script), do we want a way to override this?


src/App/Document.h
```diff
@@ -563,6 +564,9 @@ class AppExport Document : public App::PropertyContainer
 
     std::string oldLabel;
     std::string myName;
+
+    /// Time document was last modified
+    std::time_t myEditTime;
```
chennes, line 569: To me "edit time" sounds like "last time I made a change to my open document", but what you are really after is "lastSavedTime", right?

AjinkyaDahale: The name can be workshopped. I think the key is the "my" i.e. it is the time used as reference by the `Document` object.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9107">pipeline status</a> for feature branch [PR_9107](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9107). Pipeline [820775644](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/820775644) was triggered at [796c277](https://github.com/FreeCAD/FreeCAD/pull/9107/commits/796c27717fea0ab8d2fadd77e2845d0408a464e5). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Can you bypass the timestamp altogether and calculate the hash of the file instead?

AjinkyaDahale:
OK guess I'll have to address point 5. sooner rather than later.

AjinkyaDahale:
@chennes hash is possible. Do you know what library I should be using?

chennes:
No, I've only done it in Python, where's it's built-in.

chennes:
Do we already link against OpenSSL?

FlachyJoe:
If timestamp is problematic, what about to save an UUID for the current file version (uuid_current) and the last saved (uuid_saved) ?
If uuid_saved is not the same as the on disk one then this file is not the last place we saved and we can warn.
I think UUID generation spends less resource then hash of the whole file. See src/Base/Uuid.cpp

AjinkyaDahale:
@chennes what does openssl have to do with this? I see there's a `filesystem::hash_value`, but that is apparantly for the path rather than file content.

@FlachyJoe So UUID is guaranteed to change everytime I save over?

wwmayer:
An option is to use https://doc.qt.io/qt-6/qfilesystemwatcher.html
Here the second FC instance can be notified that the file has been changed outside.

Alternatively you can use a timer and regularly check the time stamp. If the time stamp is newer than the stored time stamp you can be sure that the file has been changed outside. This mechanism is e.g. used in the EditorView class.

chennes:
If you are still interested in using the hash, it turns out (of course) there is a Qt class: `QCryptographicHash hasher(QCryptographicHash::Sha1);`.

FlachyJoe:
> UUID is guaranteed to change everytime I save over?

UUID changes everytime it's generated. The DocumentObject.UUID is generated at the document creation and "never" changes but you can add another one which is regenerated at save time.

AjinkyaDahale:
@chennes thanks for digging out the hash method. IIRC sha1 is nowadays discouraged because of security issues, though again git itself uses sha1. Will that be a concern?

@FlachyJoe so are you suggesting that we create a new uuid everytime we save over? I think this might be an issue. What if it's saved over by an older version of FC (or even directly written over) that doesn't recreate a UUID? Additionally, is UUID meant to be regularly recreated?

@wwmayer `QFilesystemWatcher` seems like the most elegant solution to this problem, though it introduces more Qt dependency in App (not sure what the policy has been for that). Are there any other concerns you can think of?

AjinkyaDahale:
> > QFilesystemWatcher seems like the most elegant solution to this problem, though it introduces more Qt dependency in App (not sure what the policy has been for that). Are there any other concerns you can think of?
> 
> Do we really need this in App rather than in Gui?

I can very much see this happening in, for example `freecadcmd`.

FlachyJoe:
> a new uuid everytime we save over?

Yes


> that doesn't recreate a UUID

If the UUID is not here then you know the file was changed and the problem is solved: you have to warn.


> is UUID meant to be regularly recreated

Why not ? 128 bits UUID can have 3×10³⁶ different values, it gives a very small collision risk.

wwmayer:
> I can very much see this happening in, for example freecadcmd

FreeCADCmd doesn't have a running event loop and thus a timer or QFileSystemWatcher won't work because it's not possible to emit signals.

For the cmd line version I only see the option to store the time stamp when opening a file and before trying to write then a comparison is done. If the file is newer an exception is raised and the saving aborted.

Sometimes the cmd line version may want to override the file even if modified from outside. In this case we add a keyword argument `Force = True` to the function save()

chennes:
> @chennes thanks for digging out the hash method. IIRC sha1 is nowadays discouraged because of security issues, though again git itself uses sha1. Will that be a concern?

The concern would be that someone could maliciously craft a FCStd file with the same hash as the one you are working on (it basically can't happen accidentally), put it in place of the one you are working on while you are editing it, and then you'd fail to detect that it was modified and save over it without asking the user. I don't think this is a realistic scenario 😄 .

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9107">pipeline status</a> for feature branch [PR_9107](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9107). Pipeline [826720819](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/826720819) was triggered at [2e4e8eb](https://github.com/FreeCAD/FreeCAD/pull/9107/commits/2e4e8eb8d9df2d1480568681e272a2b202a3597d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9107">pipeline status</a> for feature branch [PR_9107](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9107). Pipeline [829875115](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/829875115) was triggered at [7e590e9](https://github.com/FreeCAD/FreeCAD/pull/9107/commits/7e590e95bf64fee9aafb6bb2e31ae1a3e8ced2b4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
I'm having some issues addressing the failing tests. So far this is what I observe:

1. On running `bin/FreeCAD -t Document`, tests fail.
2. On running just one of the tests `bin/FreeCAD -t Document.DocumentObserver.testSave` (the exact name might be different), the test succeeds.

My guess is that the tests use the same file, and modified time somehow is not updated correctly.

src/App/Document.cpp
```diff
@@ -856,6 +857,12 @@ Document::Document(const char* documentName)
                       0,
                       PropertyType(Prop_Hidden | Prop_ReadOnly),
                       "Link of the tip object of the document");
+
+    if (!myName.empty())
+        myEditTime = fs::last_write_time(fs::absolute(fs::path(FileName.getValue())));
+    else
+        myEditTime = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+
```
AjinkyaDahale, line 865: This check apparently is not enough, and possibly the cause for the failures. `myName` can be non-empty even for new files (for example `Unnamed`). Adding a check that this path exists also doesn't seem to help.

chennes:
@AjinkyaDahale if you are still working on this, can you please mark it as "Draft" status? thanks!

maxwxyz:
@AjinkyaDahale what's the status on this PR, will you come back to this?

AjinkyaDahale:
I'm not planning to get to this for 1.0, unless someone considers this as necessary.

maxwxyz:
@AjinkyaDahale will you pick up this PR?

AjinkyaDahale:
> @AjinkyaDahale will you pick up this PR?

If there's interest, sure.

PR #9120, opened 2023-03-29
Interactive image scale

wwmayer:

---

### DIFF
--- a/src/Gui/TaskView/TaskImageScale.cpp
+++ b/src/Gui/TaskView/TaskImageScale.cpp
@@ -24,13 +24,27 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <QDialog>
+# include <QPushButton>
 # include <map>
+# include <Inventor/events/SoLocation2Event.h>
+# include <Inventor/events/SoMouseButtonEvent.h>
+# include <Inventor/nodes/SoBaseColor.h>
+# include <Inventor/nodes/SoCoordinate3.h>
+# include <Inventor/nodes/SoLineSet.h>
+# include <Inventor/nodes/SoSeparator.h>
 #endif
 
+#include <Base/Console.h>
+#include <Base/Precision.h>
 #include <Base/Tools.h>
 #include <App/Document.h>
+#include <Gui/Application.h>
 #include <Gui/BitmapFactory.h>
 #include <Gui/Camera.h>
+#include <Gui/Document.h>
+#include <Gui/View3DInventor.h>
+#include <Gui/View3DInventorViewer.h>
+#include <Gui/ViewProviderDocumentObject.h>
 #include <Gui/TaskView/TaskView.h>
 
 #include "TaskImageScale.h"
@@ -39,6 +53,8 @@
 
 using namespace Gui;
 
+/* TRANSLATOR Gui::TaskImageScale */
+
 TaskImageScale::TaskImageScale(Image::ImagePlane* obj, QWidget* parent)
   : QWidget(parent)
   , ui(new Ui_TaskImageScale)
@@ -46,17 +62,30 @@ TaskImageScale::TaskImageScale(Image::ImagePlane* obj, QWidget* parent)
   , aspectRatio{1.0}
 {
     ui->setupUi(this);
+    ui->pushButtonCancel->hide();
     ui->spinBoxWidth->setValue(obj->getXSizeInPixel());
     ui->spinBoxHeight->setValue(obj->getYSizeInPixel());
 
     aspectRatio = obj->XSize.getValue() / obj->YSize.getValue();
 
-    connect(ui->spinBoxWidth, qOverload<int>(&QSpinBox::valueChanged), this, &TaskImageScale::changeWidth);
-    connect(ui->spinBoxHeight, qOverload<int>(&QSpinBox::valueChanged), this, &TaskImageScale::changeHeight);
+    connect(ui->spinBoxWidth, qOverload<int>(&QSpinBox::valueChanged),
+            this, &TaskImageScale::changeWidth);
+    connect(ui->spinBoxHeight, qOverload<int>(&QSpinBox::valueChanged),
+            this, &TaskImageScale::changeHeight);
+    connect(ui->pushButtonScale, &QPushButton::clicked,
+            this, &TaskImageScale::onInteractiveScale);
+    connect(ui->pushButtonCancel, &QPushButton::clicked,
+            this, &TaskImageScale::rejectScale);
 }
 
 TaskImageScale::~TaskImageScale()
 {
+    if (scale) {
+        if (scale->isActive()) {
+            scale->deactivate();
+        }
+        scale->deleteLater();
+    }
 }
 
 void TaskImageScale::changeWidth()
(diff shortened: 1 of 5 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9120">pipeline status</a> for feature branch [PR_9120](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9120). Pipeline [822213766](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/822213766) was triggered at [2992ac4](https://github.com/FreeCAD/FreeCAD/pull/9120/commits/2992ac4a1f0c42b125b1bfb381c25d96f515f145). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/TaskView/TaskImageScale.cpp
```diff
@@ -85,4 +112,196 @@ void TaskImageScale::changeHeight()
...
+        ui->spinBoxWidth->setValue(feature->getXSizeInPixel());
+        QSignalBlocker blockH(ui->spinBoxHeight);
+        ui->spinBoxHeight->setValue(feature->getYSizeInPixel());
+    }
+}
+
+void TaskImageScale::startScale()
+{
+    scale->activate();
+    ui->labelInstruction->setText(tr("Select two points in the 3d view"));
```
chennes, line 160: It took me a while to figure out how this was supposed to work. I think it might help to be more explicit about the tool's "thought process" -- maybe something like:
```suggestion
    ui->labelInstruction->setText(tr("Select two points on the image in the 3d view"));
```

src/Gui/TaskView/TaskImageScale.cpp
```diff
@@ -85,4 +112,196 @@ void TaskImageScale::changeHeight()
...
+    return nullptr;
+}
+
+void TaskImageScale::selectedPoints(size_t num)
+{
+    if (num == 1) {
+        ui->labelInstruction->setText(tr("Select second point"));
+    }
+    else if (num == 2) {
+        ui->labelInstruction->setText(tr("Enter distance value"));
```
chennes, line 135: Maybe:
```suggestion
        ui->labelInstruction->setText(tr("Enter physical distance between the two selected points"));
```

chennes (approved):
This compiles are runs fine (tested on Windows with CLion/MSVC). It took me some time to figure out what it was asking for, though, so I've suggested some minor wording changes that might help others.

src/Gui/TaskView/TaskImageScale.cpp
adrianinsaval: Haven't checked this so I don't know how crowded the UI is, if a more compact text than chennes' proposal is desired consider:
```suggestion
        ui->labelInstruction->setText(tr("Enter desired distance between the points"));
```

leoheck: Shorter and cleaner, `Enter the distance between the points` (the points were selected before by the user)

src/Gui/TaskView/TaskImageScale.cpp
leoheck: Maybe simpler and cleaner: `Select two points on the image`

src/Gui/TaskView/TaskImageScale.cpp
wwmayer: > Shorter and cleaner, Enter the distance between the points

The value that is set to the spin box after the second click is already the physical distance. The wording must be clear that the **desired** distance is meant. The ratio of desired and current distance gives the scaling factor.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9120">pipeline status</a> for feature branch [PR_9120](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9120). Pipeline [822577985](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/822577985) was triggered at [458f1fd](https://github.com/FreeCAD/FreeCAD/pull/9120/commits/458f1fdbee2462c93cdb58ea2c2c77e01a4590cd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/TaskView/TaskImageScale.cpp
chennes: But conceptually what we expect a user to be doing is clicking two points on the image whose physical separation distance they know: the fact that the implementation doesn't care if they click on the image or not probably doesn't matter to someone using the tool. I can't imagine a scenario where someone would be clicking outside their image when using this tool.

leoheck: @wwmayer I don't disagree. But when we are end users, we are looking for simple prompts. Freecad very often has some complicated vocabulary that does not help in the end-user land.

wwmayer: > But conceptually what we expect a user to be doing is clicking two points on the image whose physical separation distance they know:

The function just does what the old function did. Since I haven't heard of any complaints about this behaviour I don't know whether it's wise to suddenly change it.

The second point is that getting points from the image is way more expensive than getting it from the focal plane because you have to traverse the scene graph for every tiny mouse movement and depending on how many nodes there are the application may be dead slow. If you have ever worked with Coin's SoExtSelection node you will know what I mean.

chennes: OK. If you say it isn't reasonable I believe you 👍 -- I have not worked with SoExtSelection. If it makes sense to everyone else then I will adapt.

wwmayer: What could be done is adding an option where the user can decide if the line must completely lie on the plane. And in order to avoid the expensive traversal of the scene graph it could still compute the intersection point with the focal plane for mouse move events and for click events it does then compute the intersection with the image plane.

wwmayer: There is no an option to let the user decide. By default it requires the points to be inside the image.

leoheck: I use this feature a lot. I never thought or wanted to use it extrapolating the plane or the area of the image. I understand we don't need to do those things because the physical dimension the user is going to enter should be something known about something inside the image. If the user selects a point out of the image, the dimension they know is not meaningful anymore and then this tool loses its value.

wwmayer: How do you know that this is the only valid use case? What if a user wants an image to fit into a face of another object? In this case the user will click on the corner points of the face and they don't have to be necessarily inside the image plane.

chennes:
IMO we should merge this as it stands now (so that the users who depend on the feature have it back), and if there are further refinements to be made, then we can work on them separately.

wwmayer:
OK.

PaddleStroke:
@wwmayer are you done on the image work? If so I'll jump in.

wwmayer:
For now yes.

PR #9125, opened 2023-03-30
[Sketcher] [DO NOT MERGE] General Tangency with B-splines

AjinkyaDahale:
Generalizing `AngleViaPoint` to allow working with B-splines without nearest point. Instead this will take the same parameter as `PointOnObject`.

This would need one or two extra param to be stored ([depending] on whether one or both curves are B-splines or need this param otherwise).

@abdullahtahiriyo opinions?

### DIFF
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -3145,80 +3145,70 @@ int Sketch::addAngleAtPointConstraint(int geoId1,
     }
 
     int tag = -1;
+    // FIXME: Perform construction of any parameters where this method is called instead of here
     if (e2c) {
-        // increases ConstraintsCounter
-        tag = Sketch::addPointOnObjectConstraint(geoId1, pos1, geoId2, driving);
+        if (Geoms[geoId2].type == BSpline) {
+            GCS::Point &p1 = Points[getPointId(geoId1, pos1)];
+            auto* partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId2].geo);
+            double uNear;
+            partBsp->closestParameter(Base::Vector3d(*p1.x, *p1.y, 0.0), uNear);
+            double* pointparam = new double(uNear);
+            Parameters.push_back(pointparam);
+            tag = addPointOnObjectConstraint(geoId1, pos1, geoId2, pointparam, driving); //increases ConstraintsCounter
+            --ConstraintsCounter;
+            GCSsys.addConstraintAngleViaPointAndParam(*crv2, *crv1, p, pointparam, angle, tag, driving);
+        }
+        else {
+            // increases ConstraintsCounter
+            tag = Sketch::addPointOnObjectConstraint(geoId1, pos1, geoId2, driving);//increases ConstraintsCounter
+            GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
+        }
     }
     if (e2e) {
         tag = ++ConstraintsCounter;
         GCSsys.addConstraintP2PCoincident(p, *p2, tag, driving);
+        GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
     }
     if (avp) {
         tag = ++ConstraintsCounter;
+        if (Geoms[geoId1].type == BSpline || Geoms[geoId2].type == BSpline) {
+            if (Geoms[geoId1].type == BSpline && Geoms[geoId2].type == BSpline) {
+                GCS::Point &p3 = Points[getPointId(geoId3, pos3)];
+                auto* partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId1].geo);
+                double uNear;
+                partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
+                double* pointparam1 = new double(uNear);
+                Parameters.push_back(pointparam1);
+                addPointOnObjectConstraint(geoId3, pos3, geoId1, pointparam1, driving); //increases ConstraintsCounter
+                --ConstraintsCounter;
+                partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId2].geo);
+                partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
+                double* pointparam2 = new double(uNear);
+                addPointOnObjectConstraint(geoId3, pos3, geoId2, pointparam2, driving); //increases ConstraintsCounter
+                Parameters.push_back(pointparam2);
+                --ConstraintsCounter;
+                GCSsys.addConstraintAngleViaPointAndTwoParams(*crv1, *crv2, p, pointparam1, pointparam2, angle, tag, driving);
+            }
+            else {
+                if (Geoms[geoId1].type != BSpline) {
+                    std::swap(geoId1, geoId2);
+                    std::swap(pos1, pos2);
+                }
+                GCS::Point &p3 = Points[getPointId(geoId3, pos3)];
+                auto* partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId1].geo);
+                double uNear;
+                partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
+                double* pointparam = new double(uNear);
+                Parameters.push_back(pointparam);
+                addPointOnObjectConstraint(geoId3, pos3, geoId1, pointparam, driving); //increases ConstraintsCounter
+                --ConstraintsCounter;
+                GCSsys.addConstraintAngleViaPointAndParam(*crv1, *crv2, p, pointparam, angle, tag, driving);
+            }
+        }
+        else
+            GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
     }
 
-    if (Geoms[geoId1].type == BSpline || Geoms[geoId2].type == BSpline) {
-        if (Geoms[geoId1].type == BSpline && Geoms[geoId2].type == BSpline) {
-            GCS::Point& p3 = Points[getPointId(geoId3, pos3)];
-            auto* partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId1].geo);
-            double uNear;
-            partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
-            double* pointparam1 = new double(uNear);
-            Parameters.push_back(pointparam1);
-            addPointOnObjectConstraint(geoId3,
-                                       pos3,
-                                       geoId1,
-                                       pointparam1,
-                                       driving);  // increases ConstraintsCounter
-            --ConstraintsCounter;
-            partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId2].geo);
-            partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
-            double* pointparam2 = new double(uNear);
-            addPointOnObjectConstraint(geoId3,
-                                       pos3,
-                                       geoId2,
-                                       pointparam2,
-                                       driving);  // increases ConstraintsCounter
-            Parameters.push_back(pointparam2);
-            --ConstraintsCounter;
-            GCSsys.addConstraintAngleViaPointAndTwoParams(*crv1,
-                                                          *crv2,
-                                                          p,
-                                                          pointparam1,
-                                                          pointparam2,
-                                                          angle,
-                                                          tag,
-                                                          driving);
-        }
-        else {
-            if (Geoms[geoId1].type != BSpline) {
-                std::swap(geoId1, geoId2);
-                std::swap(pos1, pos2);
-            }
-            GCS::Point& p3 = Points[getPointId(geoId3, pos3)];
-            auto* partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId1].geo);
-            double uNear;
-            partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
-            double* pointparam = new double(uNear);
-            Parameters.push_back(pointparam);
-            int tag = addPointOnObjectConstraint(geoId3,
-                                                 pos3,
-                                                 geoId1,
-                                                 pointparam,
-                                                 driving);  // increases ConstraintsCounter
-            --ConstraintsCounter;
-            GCSsys.addConstraintAngleViaPointAndParam(*crv1,
-                                                      *crv2,
-                                                      p,
-                                                      pointparam,
-                                                      angle,
-                                                      tag,
-                                                      driving);
-        }
-    }
-    else {
-        GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
-    }
     return ConstraintsCounter;
 }
 
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -3041,7 +3041,6 @@ int Sketch::addAngleAtPointConstraint(int geoId1,
                                       ConstraintType cTyp,
                                       bool driving)
 {
-
     if (!(cTyp == Angle || cTyp == Tangent || cTyp == Perpendicular)) {
         // assert(0);//none of the three types. Why are we here??
         return -1;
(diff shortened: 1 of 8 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125">pipeline status</a> for feature branch [PR_9125](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125). Pipeline [823291631](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/823291631) was triggered at [66e333c](https://github.com/FreeCAD/FreeCAD/pull/9125/commits/66e333ca93f832329b8eeba15db7e7fc794d6964). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125">pipeline status</a> for feature branch [PR_9125](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125). Pipeline [826720974](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/826720974) was triggered at [4360173](https://github.com/FreeCAD/FreeCAD/pull/9125/commits/43601736a97cde9b609c703a99d94a74313926dc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125">pipeline status</a> for feature branch [PR_9125](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125). Pipeline [828847541](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/828847541) was triggered at [59b953c](https://github.com/FreeCAD/FreeCAD/pull/9125/commits/59b953caac7c1e393cbbe36763acc631ae872d7c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
We now have a POC ready. However, the behavior still is janky. Marking this as ready for review but not to merge.

src/Mod/Sketcher/App/Sketch.cpp
```diff
@@ -2691,7 +2690,42 @@ int Sketch::addAngleAtPointConstraint(
...
+            partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
+            double* pointparam = new double(uNear);
+            Parameters.push_back(pointparam);
+            int tag = addPointOnObjectConstraint(geoId3, pos3, geoId1, pointparam, driving); //increases ConstraintsCounter
+            --ConstraintsCounter;
+            GCSsys.addConstraintAngleViaPointAndParam(*crv1, *crv2, p, pointparam, angle, tag, driving);
+        }
+    }
+    else
+        GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
```
AjinkyaDahale, line 2728: This will move to the method that calls this method. This may mean having to split this method as well.

src/Mod/Sketcher/App/planegcs/Geo.cpp
```diff
@@ -663,6 +663,159 @@ DeriVector2 BSpline::CalculateNormal(const Point &p, const double* derivparam) c
...
+            / (flattenedknots[startpole + i + degree] - flattenedknots[startpole + i]);
+    }
+    double wslopesum = degree * BSpline::splineValue(
+        *param, startpole + degree, degree - 1, d, flattenedknots);
+
+    DeriVector2 result(wsum*xslopesum - wslopesum*xsum, wsum*yslopesum - wslopesum*ysum);
+
+    // get dx, dy of the normal as well
+    bool dpfound = false;
+    for (size_t i = 0; i < numpoints; ++i) {
```
AjinkyaDahale, line 721: It may be possible to exploit the Derivector tool but this way was much quicker.

src/Mod/Sketcher/App/planegcs/Constraints.h
```diff
@@ -700,6 +702,61 @@ namespace GCS
...
+        ConstraintAngleViaPointAndParam(Curve &acrv1, Curve &acrv2, Point p, double* param1, double* angle);
+        ~ConstraintAngleViaPointAndParam() override;
+        ConstraintType getTypeId() override;
+        void rescale(double coef=1.) override;
+        double error() override;
+        double grad(double *) override;
+    };
+
+    // TODO: Do we need point here at all?
+    class ConstraintAngleViaPointAndTwoParams : public Constraint
```
AjinkyaDahale, line 733: @abdullahtahiriyo can you comment on whether the point is needed in this case if I choose to only equalize slopes of both curves at their respective parameters?

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -3700,13 +3700,13 @@ void CmdSketcherConstrainPerpendicular::activated(int iMsg)
...
-                              QObject::tr("Perpendicular to B-spline edge currently unsupported."));
-                return;
-            }
+            // if( geom2 && geom2->getTypeId() == Part::GeomBSplineCurve::getClassTypeId() ){
+            //     // unsupported until normal to B-spline at any point implemented.
+            //     Gui::TranslatedNotification(Obj,
+            //                   QObject::tr("Wrong selection"),
+            //                   QObject::tr("Perpendicular to B-spline edge currently unsupported."));
+            //     return;
+            // }
```
AjinkyaDahale, line 3709: I've basically removed all restrictions for B-splines right now in GUI. Some may have to be put back up, or need other odifications

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125">pipeline status</a> for feature branch [PR_9125](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125). Pipeline [829875140](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/829875140) was triggered at [bf656fc](https://github.com/FreeCAD/FreeCAD/pull/9125/commits/bf656fc19b305bec6f63ebbd572f3c7fd744845c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
@0penBrain do you have knowledge about the "angle-via-point" mechanism in th solver, or know someone who does? Abdullah seems away for now.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125">pipeline status</a> for feature branch [PR_9125](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125). Pipeline [834640323](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/834640323) was triggered at [73154a0](https://github.com/FreeCAD/FreeCAD/pull/9125/commits/73154a01be854da6501cab9e5ae5e10bdf0d3f53). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
> @0penBrain do you have knowledge about the "angle-via-point" mechanism in th solver, or know someone who does? Abdullah seems away for now.

Sorry I have no knowledge about that. @abdullahtahiriyo told he'll be unavailable till mid of April IIRC.

AjinkyaDahale:
> Sorry I have no knowledge about that. @abdullahtahiriyo told he'll be unavailable till mid of April IIRC.

OK. Middle of April is close, so I can wait.

src/Mod/Sketcher/App/Sketch.cpp
```diff
@@ -2682,16 +2681,68 @@ int Sketch::addAngleAtPointConstraint(
...
+                Parameters.push_back(pointparam1);
+                addPointOnObjectConstraint(geoId3, pos3, geoId1, pointparam1, driving); //increases ConstraintsCounter
+                --ConstraintsCounter;
+                partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId2].geo);
+                partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
+                double* pointparam2 = new double(uNear);
+                addPointOnObjectConstraint(geoId3, pos3, geoId2, pointparam2, driving); //increases ConstraintsCounter
+                Parameters.push_back(pointparam2);
+                --ConstraintsCounter;
+                GCSsys.addConstraintAngleViaPointAndTwoParams(*crv1, *crv2, p, pointparam1, pointparam2, angle, tag, driving);
```
AjinkyaDahale, line 2725: Note that the parameter here has to be the same for point-on-object and angle-via-point, so these constraints are best combined. This makes other point-on-object constraints partially redundant (unsure why they are not marked completely redundant).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125">pipeline status</a> for feature branch [PR_9125](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125). Pipeline [835264836](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/835264836) was triggered at [78eb993](https://github.com/FreeCAD/FreeCAD/pull/9125/commits/78eb993ba4b3a26b738d7e1131ab8c91e0b74d1d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/App/planegcs/Constraints.cpp
```diff
@@ -2494,6 +2494,212 @@ double ConstraintAngleViaPoint::grad(double *param)
...
+double ConstraintAngleViaPointAndParam::error()
+{
+    if (pvecChangedFlag)
+        ReconstructGeomPointers();
+    double ang = *angle();
+    DeriVector2 n1 = crv1->CalculateNormal(cparam());
+    DeriVector2 n2 = crv2->CalculateNormal(poa);
+
+    // rotate n1 by angle
+    DeriVector2 n1r(n1.x * cos(ang) - n1.y * sin(ang), n1.x * sin(ang) + n1.y * cos(ang));
```
abdullahtahiriyo, line 2571: Do you think it makes sense to extend DeriVector2 to rotate by an angle, instead of hardcoding the rotation and derivative?

I am not sure exactly what is intended here, so I may be wrong. Please take a look at the sign of the derivative. If what is intended here is that the second argument is the partial of the first with respect to the angle, it appears the sign is reversed (dcos=-sin, dsin=cos).

AjinkyaDahale: No comments (yet) on the first paragraph. Derivector math needs some thinking. I recently happened to come across a video about "dual numbers" which they called "cousins" of complez numbers that have use in these" value + derivative" calculations.

In the line, the arguments are x and y components of the rotated vector. I believe derivative information is lost when creating n1r (take a look at the constructor).

abdullahtahiriyo: My bad. I know derivector constructor takes coordinates and then derivatives, but somehow I did not realise that it takes x and y and then their derivatives. The fact that sin and cos derivatives are cos and -sin did not help me realise my mistake. Sorry for that.

I remember DeepSOIC was into "dual numbers" when he created DeriVector2.

I am also sorry I did not manage to progress on this review. I will try to carry on tomorrow.

AjinkyaDahale:
So there's a lot happening here, and I'm trying to figure out how to approach this. Some thoughts:
1. We need to add a new parameter for the constraint, which for point-on-object became the value of the consttraint. If the constraint is between two splines and a point, this would have to be two values---not a problem since constraints can have two values. However, angle-via-point already has one value, viz. the angle between the curves.
2. The order in which the curves are provided matter. When calling the solver constraint creator, this order can get mixed up. So that also may be create unintended consequences.
3. There is currently an issue when drawing angle constraints: the angle is not drawn properly. This may be an ordering issue.
4. We need to tie point-on-object and angle-via-point so that they use the same curve parameter for the b-spline. So both constraints have to be made at the same time. However, this is different from how it works with other curves like hyperbolas. I've tried to avoid this as much as possible but this can still have unintended consequences, for example, if the same point is used for other constraints, some partial redundancy is detected.

abdullahtahiriyo:
I will try to provide some input:

> 5\. The constraint depends also on the direction of the curves. A B-spline and another exactly reverse B-spline will behave very differently.

We used to use a criteria to have curves go counter-clock-wise (CCW). 

There was a function to track preexisting curves and port them:
```SketchObject::port_reversedExternalArcs```

I only have a vague memory of all this. But I think it was at the same time as AngleViaPoint.

We did not try to enforce this on B-Splines to me knowledge. If it is a solution we may try to.

> 3\. There is currently an issue when drawing angle constraints: the angle is not drawn properly. This may be an ordering issue.

Do you mean that we have a problem with AngleViaPoint in master?

> 4\. We need to tie point-on-object and angle-via-point so that they use the same curve parameter for the b-spline. So both constraints have to be made at the same time. However, this is different from how it works with other curves like hyperbolas. I've tried to avoid this as much as possible but this can still have unintended consequences, for example, if the same point is used for other constraints, some partial redundancy is detected.

This is quite some issue. 

I have given this quite some thought, as the basis is that for B-Splines the point becomes an issue or even a problem, as, at least for B-Splines, we want to work with a curve parameter instead. Yet, the whole AngleViaPoint framework is based on using a Point. At Sketcher level the point is there and, although suboptimal, it is possible to add the point on object and the tangency separately. They are totally independent too. It feels like we are trying to squeeze  the B-Spline approach into the AngleViaPoint framework, when it appears not to be well suited for it. In doing so, we come to having differences in how things work (having to create the constraints simultaneously)... and on top, the issue of partial redundancy when reusing the point.

Ajinkya, you have walked the path and are in a much better position than me to know better. Do you think we should think of a different framework (from the one of AngleViaPoint)?

Something perhaps unifying:
- All curves in terms of params
- No points
- All solver constraints integrated into a single sketcher constraint, added at once. One single symbol. Solver implementation transparent to the user.
- Not sure exactly how the two params (one of each curve) remain at a single point in space though without the point. We should avoid, I believe, to have a hidden solver-only point.

Alternatively, given that you have done quite some work, we could try to see what is creating the partial redundancy, and try to remove the corners from the other issues.

Give it some thought. Should we build a new thing based on the knowledge gained, or try to fix the issues?

> 2\. The order in which the curves are provided matter. When calling the solver constraint creator, this order can get mixed up. So that also may be create unintended consequences.

The order matters because of the curve direction (reverse direction), or it is another reason? I am not following this one.

> 1. We need to add a new parameter for the constraint, which for point-on-object became the value of the consttraint. If the constraint is between two splines and a point, this would have to be two values---not a problem since constraints can have two values. However, angle-via-point already has one value, viz. the angle between the curves.

If I understand you correctly, this is only necessary on the solver side. Wouldn't it solve the problem to add one extra ```double *``` to ```struct ConstrDef```, I mean a "thirdvalue".

Please, come back with your remarks. This will help me better understand where you stand and how to move forward.

AjinkyaDahale:
Thanks for the input, @abdullahtahiriyo, and sorry for the late reply.

> > 5. ... 
> We used to use a criteria to have curves go counter-clock-wise (CCW)...
> We did not try to enforce this on B-Splines to me knowledge. If it is a solution we may try to.

Let me take a look at this.

> > 3. There is currently an issue when drawing angle constraints: the angle is not drawn properly. This may be an ordering issue.
> 
> Do you mean that we have a problem with AngleViaPoint in master?

This is an issue with the _drawing_ of the angle. You can probably tell from the following image, though I added a lot of clutter to make it confusing. The angle marked 30 degrees is definitely not that small: in fact, it is meant for another pair of edges.
Screenshot from 2023-04-12 22-38-04

> > 4. ... 
> This is quite some issue.
> 
> I have given this quite some thought,... 
> Ajinkya, you have walked the path and are in a much better position than me to know better. Do you think we should think of a different framework (from the one of AngleViaPoint)?
> 
> Something perhaps unifying:...
> 
> Alternatively, given that you have done quite some work, we could try to see what is creating the partial redundancy, and try to remove the corners from the other issues.
> 
> Give it some thought. Should we build a new thing based on the knowledge gained, or try to fix the issues?

If we are going to add that third value to `Sketcher::Constraint` it may make it more streamlined working with the parameters. However, we also need to refactor the solver for some other issues already. I hope
 
> > 2. The order in which the curves are provided matter. When calling the solver constraint creator, this order can get mixed up. So that also may be create unintended consequences.
> 
> The order matters because of the curve direction (reverse direction), or it is another reason? I am not following this one.

Need to study this deeper. Currently: spline, line, point works, but line, spline, point has some issues. However, the behavior I was talking to is the CCW angle value. These two cases should look totally different.

> 
> > 1. We need to add a new parameter for the constraint, which for point-on-object became the value of the consttraint. If the constraint is between two splines and a point, this would have to be two values---not a problem since constraints can have two values. However, angle-via-point already has one value, viz. the angle between the curves.
> 
> If I understand you correctly, this is only necessary on the solver side. Wouldn't it solve the problem to add one extra `double *` to `struct ConstrDef`, I mean a "thirdvalue".

Possible.

AjinkyaDahale:
> That was originally not my intention. Only to ConstrDef if necessary. Adding a third value to sketcher constraints has more implications.

If it is OK to only add it to `ConstrDef` and not `Sketcher::Constraint`, that's only more convenient.
 
> However, given that you raise the issue, what would this third value constraint when used?

I actually wonder what the _second_ value constrains now that Snell's law constraint is gone :D. If you recall Point-on-Object implementation in spline the value of the parameter was stored as the _first_ value of the constraint.

> At a first glance, it appears to me it may be intended as a freewheeling parameter. If so, it would then appear that instead of hiding a solver implementation detail we might be exposing it to higher hierarchy code. If we extend sketcher::Constraint it should be to support a constraint. A value that will be restricting when driving.

Your understanding here is correct. And indeed it can be made restricting when driving. The only thing it can constrain, though, is the position along the curve, something which is abstracted away from the user currently. You did address this in the next paragraph, so I'll add my thoughts about it there shortly. One thing to consider though, is that every value of a constraint can be made driving or non-driving separately, in theory. However, currently only the entire constraint can be set so. This may still be the model to keep: even when a constraint with multiple "values" is driving, only one of the values is actually made driving.

> I also tried to think about introducing "something" to embody the curve parameter at sketcher level, but nothing I have thought of makes sense in the whole picture.

The "something" will basically be a documentation first, such that the user knows how the parametrization is done. It's probably least ambiguous for B-splines, since there's no getting around parameters. I had thought that the knot contraint's value would actually be one such parameter, so that the discontinuity itself changes (the non-uniform part of NURBS) if the knot constraint is made non-driving. 

I see little use of simple point-on-object constraints being "driving", but again, there might be something there. Let's take an example of a circular arc. Consider two cases: when it's parameters are (0, angle) vs (startangle, endangle) w.r.t. the horizontal. If PoO is made driving, the behavior will be quite different. Which one do we want? Is there any use for constraining.

> During the release freeze we might have some more time to dedicate to this topic. I keep reading your input and giving it some thought.

Sure. The hard math itself isn't currently an issue so we can think of the structural aspects.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125">pipeline status</a> for feature branch [PR_9125](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9125). Pipeline [903564957](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903564957) was triggered at [c74c87f](https://github.com/FreeCAD/FreeCAD/pull/9125/commits/c74c87f52e49b7045678444a3c7439e16831054a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
@chennes how do I run pre-commit locally on the already existing commits? The developer handbook doesn't seem to have instructions for this.

FlachyJoe:
@AjinkyaDahale if you compile your working tree you run all the changes even the not committed ones. Maybe I misunderstood your question.

adrianinsaval:
try amending the latest commit

AjinkyaDahale:
> @AjinkyaDahale if you compile your working tree you run all the changes even the not committed ones. Maybe I misunderstood your question.

@FlachyJoe I was referring to this tool called `pre-commit`. See the comments starting from this one: https://github.com/FreeCAD/FreeCAD/pull/10294#issuecomment-1717259298.

@adrianinsaval Just did. Maybe I need to ammend all commits, or run clang-format on all changed files?

AjinkyaDahale:
@adrianinsaval I ran `clang-format` on every commit (effectively ammending all) and pre-commit still complains.

0penBrain:
> @chennes how do I run pre-commit locally on the already existing commits? The developer handbook doesn't seem to have instructions for this.

You can `pre-commit run --from-ref older_commit --to-ref HEAD` where `older_commit` is the older SHA to consider for file changing.
Notice that it will not run on "every commit", but run on "all files changed by the commit range".

FlachyJoe:
`pre-commit run --all-files` should do the job.

AjinkyaDahale:
@FlachyJoe @0penBrain both the commands you suggested fail locally exactly the same as the results here on github.

AjinkyaDahale:
OK so the method seems to be:
1. Run pre-commit, it may fail.
2. If there are new changes, commit them.
3. Run pre-commit again.

AjinkyaDahale:
Closing in favour of #11853.

PR #9126, opened 2023-03-30
Gui: prevent crashes on MacOS using boost::format

0penBrain:
This PR is an attempt to solve crashes on MacOS described here :
https://forum.freecad.org/viewtopic.php?t=75442
https://forum.freecad.org/viewtopic.php?t=77234

### DIFF
--- a/src/Mod/Part/Gui/TaskOffset.cpp
+++ b/src/Mod/Part/Gui/TaskOffset.cpp
@@ -174,8 +174,8 @@ bool OffsetWidget::accept()
         double offsetValue = d->ui.spinOffset->value().getValue();
         Gui::cmdAppObjectArgs(d->offset, "Value = %f", offsetValue);
         d->ui.spinOffset->apply();
-        Gui::cmdAppObjectArgs(d->offset, "Mode = %i", d->ui.modeType->currentIndex());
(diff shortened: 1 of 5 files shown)


0penBrain:
@wwmayer do you see any risk at merging this PR ? I just replaced all `%i` with `%d` which at worst shouldn't change anything.

wwmayer:
According to https://www.boost.org/doc/libs/1_66_0/libs/format/doc/format.html#printf_directives using "%i" should be valid. But since `%d` is also valid it's worth a try.

0penBrain:
Yes `%i` is valid as is `%d`. But while both converts to decimal numbers, there are cases where 'such as in `scanf` where they are not equivalent.
Hence my present try. :)
I'm gonna merge and we'll see. If it's not enough, I may try change all `%d` to `%.1d`.

wwmayer:
> Yes %i is valid as is %d. But while both converts to decimal numbers, there are cases where 'such as in scanf where they are not equivalent.

Sooner or later we should move from boost::format to libfmt because for `fmt::sprintf` this library does some proper run-time checks whether the passed arguments match with the format specifiers. (From what I have read it can even do some compile-time checks with C++20.)

When using `fmt::format` (what I prefer) then for each argument `{}` can be used which will simplify things a lot. The compiler will find the right type automatically, then.

Issue #9137, opened 2023-03-31
[Issue] Preference page 'Unit' is hard to find for user


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecad.org/viewtopic.php?t=77065&start=20

### Version

0.21 (Development)

### Full version info

```shell
last
```

### Subproject(s) affected?

Core

### Issue description

The Preference page 'Unit' is hard to find for user while it's a very important setting.

I suggest putting it on the front page.
- Unit + decimals can be on the same line
- fraction of inch can be hidden by default unless the only unit system that uses it is selected.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #9143, opened 2023-03-31
Pref: Remove Unit tab and move settings back to main page.

PaddleStroke:
This has been discussed on the forum and everyone seems to agree about it.
Also the main page just lost 2 lines since the wb pref PR merged. So now it's not over-crowded.

Description : 
- Moves content of Unit tab to General tab.
- Fraction of inch is invisible if the unit system is not US building instead of just deactivated.
- Number of decimal is on the same line as unit system, so the whole unit page fit on one line.
- Change the groupbox name from "Language" to "Language and number format".
- Change the QLabel "Change language" to "Language"

Issue : https://github.com/FreeCAD/FreeCAD/issues/9137

---


src/Gui/DlgGeneral.ui
```diff
@@ -71,13 +71,100 @@
...
+          <widget class="QLabel" name="unitSystemLabel">
+           <property name="text">
+            <string>Unit system:</string>
+           </property>
+          </widget>
+         </item>
+         <item row="1" column="1">
+          <widget class="QComboBox" name="comboBox_UnitSystem">
+           <property name="toolTip">
+            <string>Unit system that should be used for all parts the application</string>
```
chennes, line 83: "for all parts *of* the application"

src/Gui/DlgGeneralImp.cpp
```diff
@@ -462,4 +536,23 @@ void DlgGeneralImp::onLoadPreferencePackClicked(const std::string& packName)
     }
 }
 
+
+void DlgGeneralImp::on_comboBox_UnitSystem_currentIndexChanged(int index)
```
chennes, line 540: This style of function name makes it look like an auto-connected slot. If it is, it should not be: Clazy will complain about it, that use is no longer recommended. (Sorry, on a phone, it's hard to see for sure).

chennes (changes requested):
Thanks for the PR: couple of minor things.

src/Gui/DlgGeneralImp.cpp
PaddleStroke: In this PR I only copy pasted without having a more indepth look.
It appears that it is indeed auto-connected. I can change that.

chennes:
I haven't checked, but I assume this will require some updates on the Wiki.

PaddleStroke:
I guess so. However there might be further changes to the main page, so it
may not be useful to change now.

PR #9146, opened 2023-04-01
[PATH] Update getDrillableTargets() to evaluate all faces 1..<n>

jimzim111:
This fixes one cause of valid holes not getting included in Drill/Helix Ops due to a list iteraton off-by-one counter.  Detailed explanation at https://forum.freecad.org/viewtopic.php?style=5&t=77055 (the first cause discussed on that post, disregard the 2nd as I'm still investigating/working on that one)

---

### DIFF
--- a/src/Mod/Path/Path/Base/Drillable.py
+++ b/src/Mod/Path/Path/Base/Drillable.py
@@ -295,7 +295,7 @@ def getDrillableTargets(obj, ToolDiameter=None, vector=App.Vector(0, 0, 1)):
     shp = obj.Shape
 
     results = []
-    for i in range(1, len(shp.Faces)):
+    for i in range(1, len(shp.Faces)+1):
         fname = "Face{}".format(i)
         Path.Log.debug(fname)
         candidate = obj.getSubObject(fname)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9146">pipeline status</a> for feature branch [PR_9146](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9146). Pipeline [826746112](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/826746112) was triggered at [8b03616](https://github.com/FreeCAD/FreeCAD/pull/9146/commits/8b0361674daa5c8f54a0148cf454e93132438abd). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
I'm merging this because it passes tests and appears to resolve a real problem.  However, there is no github issue, only a forum topic.  
What concerns me is that the existing unit tests didn't detect the original problem and we've added no new unit tests to the suite.  
Why didn't the problem surface before?  Was it that the current unit tests only have a single base object in the job?  Was it that the Face0 or FaceN didn't have a hole?

jimzim111:
I think the bug could hit a single base object or any/all of multiple
bases.  It's just random luck of the draw- if a valid hole face happens to
be the highest face number of the entire object it would be missed.

In order to create a test case, do you know if there's any way to influence
the numbering of faces when objects are computed?  Or if not, maybe I can
try making an object with a valid hole and as few other faces as possible
to maximize the odds that the hole gets the highest face number.

Issue #9135, opened 2023-03-31
[Issue] Sketcher toolbars missing/appear on other workbenches


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecad.org/viewtopic.php?p=671911#p671911

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.0.32641 (Git)
Build type: Release
Branch: sketcher_toolbar_fix
Hash: 5329cede7f926d55aec396b8fee091889ef14767
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk , OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * A2plus 0.4.60k
  * Curves 0.6.8
[/code]
```

### Subproject(s) affected?

Sketcher

### Issue description

1. reset config, restart FreeCAD
2. set startup workbench to Sketcher. Restart FreeCAD.
so far so good, Sketcher workbench is loaded, with almost no toolbars
3. make a new sketch
4. while the sketch is still open, switch to another workbench, then back to sketcher
-> no geometry and constraints toolbars, can't do anything.
Reopening the sketch helps.

and then

5. restart FreeCAD
now, all geometry and constraints toolbars appear... which is not normal, but nothing bad really.
6. switch to Part workbench
now, all sketcher geometry and constraints toolbars are on Part! completely displacing all the Part workbench tools.
Switching back to Sketcher and then back to Part fixes things.
![image](https://user-images.githubusercontent.com/19984177/229150404-72a797be-d7f5-4fa6-841c-d82803a0672e.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
The first thing to note is that this issue happen when the user changes of workbench while being in edit mode. In this case, the toolbars are not the only weird thing, the UI looks pretty buggy: 
![image](https://user-images.githubusercontent.com/19984177/229151267-ace8d62f-bb3c-481e-962a-8080a3c1dbcb.png)

First question : Do we want to keep this buggy behavior ? My first thought was that leaving sketcher should close edit-mode. As it doesn't make sense to be in edit mode while in PartDesign or whatever else.

To this one valid counter argument was given : 
[...]

PR #9150, opened 2023-04-01
Sketcher: restore toolbars when changing between workbenches in edit …

wwmayer:
…mode

This partially fixes issue #9135

---

### DIFF
--- a/src/Gui/ToolBarManager.cpp
+++ b/src/Gui/ToolBarManager.cpp
@@ -43,8 +43,9 @@ ToolBarItem::ToolBarItem() : visibility(HideStyle::VISIBLE)
 
 ToolBarItem::ToolBarItem(ToolBarItem* item, HideStyle visibility) : visibility(visibility)
 {
-    if ( item )
+    if (item) {
         item->appendItem(this);
+    }
 }
 
 ToolBarItem::~ToolBarItem()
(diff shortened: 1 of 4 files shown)


src/Gui/ToolBarManager.cpp
```diff
@@ -327,15 +328,35 @@ void ToolBarManager::setup(ToolBarItem* item, QToolBar* toolbar) const
...
+            return true;
+        }
+
+        QVariant property = action->property("HideStyle");
+        if (property.isNull()) {
+            return false;
+        }
+
+        // If hide style is FORCE_HIDE then never save the state because it's
+        // alwas controlled by the application.
```
chennes, line 344:
```suggestion
        // always controlled by the application.
```

chennes:
Do you have an opinion about the
```
'auto it' can be declared as 'auto *it' [readability-qualified-auto]
```
linter warning? I'm not a huge fan of the unnecessary *, I feel like C++ has enough punctuation already and to me that doesn't make the code any clearer in most cases (though I grant that there are some exceptions where it does make sense to include the *). Anyway, I'd be in favor of disabling that lint check.

wwmayer:
> linter warning? I'm not a huge fan of the unnecessary *, I feel like C++ has enough punctuation already and to me that doesn't make the code any clearer in most cases (though I grant that there are some exceptions where it does make sense to include the *). Anyway, I'd be in favor of disabling that lint check.

Absolutely. We should disable the warning.

Another warning that I see frequently is that a function could be made static but the code checker completely ignores the fact that in many cases the function overrides a virtual function and thus it's not possible to make it static.

PR #9153, opened 2023-04-01
Path: make drilling cycles G98/G99 aware. 

J-Dunn:

---
This PR adds G99 retract mode awareness to drilling cycles G81/G83 , finally making it NIST conformant. 
This satisfies both the need to avoid multiple retractions and air cutting on unobstructed jobs, and respects safe height between holes when required, preventing the tool-crashes produced by fudging G98 to act like G99 and silently ignoring safe height as is current implemented. 

A new KeepToolDown checkbox is added to UI , per wsk8's previous suggestion. 
Old paths are automatically migrated to add the new object properties, so existing FCStd files will be compatible. 
3Dview also gets to be G98 compliant ( it previously ignored retract modes, so this should not be visible to existing paths ).


GeneGH:
I do not understand the use of "keep tool down" as a proxy for G99. The drill still needs to go up between holes. This terminology seems misleading. It is not consistent with the usage in other parts of the Path WB where the milling cutter does not rise at all.

J-Dunn:
Thanks. Fair point. I adopted that from wsk8's PR, maybe without due consideration. If this is used elsewhere in a difference sense, it could be changed.  I'll wait for other f/b and make a new PR if it's worth it. 

Should this be "Retract mode" with a drop down list of G98 and G99 ?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9153">pipeline status</a> for feature branch [PR_9153](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9153). Pipeline [826746117](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/826746117) was triggered at [0c43789](https://github.com/FreeCAD/FreeCAD/pull/9153/commits/0c437894cff7919c4116bcd393937aa56f7608c1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

J-Dunn:
master update and lint tidy up submitted as #9200

PR #9149, opened 2023-04-01
[GUI] Radial gradient implementation

xtemp09:
This commit implements radial gradient as background and adds the option to settings. It also renames "Color gradient" as "Linear gradient", keeping Linear gradient default. Internally, it remains unchanged for compatibility.

Radio gradient is more suitable for CAD, since it gives a more balanced color distribution across the screen, improving visibility of model and sketches with a halo-like effect.

This pull request closes #8297. I advise to set radial gradient default in the upcoming version of FreeCAD.

<details><summary>Options</summary>

Screenshot 1 - Options

</details>

<details><summary>Screenshot 2</summary>

Screenshot 2

</details>

<details><summary>Screenshot 3</summary>

Screenshot 3

</details>

<details><summary>Screenshot 4</summary>

Screenshot 4

</details>

<details><summary>ProDark</summary>

ProDark

</details>

<details><summary>ProDark2</summary>

ProDark 2

</details>

<details><summary>ProDark3</summary>

ProDark 3

</details>

The commit looks huge due to git, though it just adds several lines of code, mostly to `SoFCBackgroundGradient.cpp` and an option to the settings.

<details><summary>Standard form</summary>

Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [X]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---

</details>

PR #9172, opened 2023-04-03
Pref: Colors: Reorganize the ui.

PaddleStroke:
Following https://github.com/FreeCAD/FreeCAD/pull/9149
@chennes @MisterMakerNL please have a look if this looks good to you.

---


Pesc0:
Looks very good. I am thinking one thing could be changed: instead of completely hiding the middle color entry when the checkbox is cleared, just grey out the option. Also the checkbox could be placed inline with the middle color row

PaddleStroke:
Yes it definitely looks better. I just pushed the change.

PaddleStroke:
What reset color button? Do you mean the button above the checkbox? It's the switch button it swap the top and bottom colors.

Pesc0:
> It's the switch button it swap the top and bottom colors.

Oh, I thought it was to reset to the default color, sorry

PaddleStroke:
@chennes yeah it's true it looks like a refresh button it's unfortunate.
I replaced the icon with the 'button_sort' that was already in the icon base.
image

chennes:
I dunno, how do you feel about it? It's an improvement for sure, but "middle" kind of screws things up with the "switch" button. As usual I'm inclined to merge the improvement and file this away as something to look at again, but if you have other ideas right not, there's no time like the present...

PaddleStroke:
Hmm I can't see an obvious way to make it much clearer. Putting the middle
color at the end looks weird imo.

The best I can think of:
Remove the icon and name the button "switch top and bottom". Just "switch"
wasn't very clear anyway.

But I think current situation with this icon is kind of ok. Not worse than
the "switch" labeled button anyway. As long as the tooltip is informative
enough I think it's ok. Though maybe it's not informative enough?

Pesc0:
> Remove the icon and name the button "switch top and bottom"

I think this would be the best solution. In addition I would move the button somewhere else, perhaps below the bottom color, otherwise it looks like some property of the top color only

adrianinsaval:
how about something like this:
image

PaddleStroke:
or this
image

xtemp09:
Perhaps, preview will be better.

Something like this:
gradient-tool18

PaddleStroke:
It would be better, but this is a lot of work that I think is not necessary. This is not a everyday-used widget of freecad, just a setting for background color.
So I think our current solution is good enough.

chennes:
Yeah, I think @adrianinsaval's suggesting is excellent.

PaddleStroke:
There it is.

MisterMakerNL:
FreeCAD_qOUQ1qf5Mj
Looks awesome! Works awesome!

xtemp09:
May I propose some observations? Every color button has a label, but _Simple color_ does not. The checkbox to enable midcolor is to the right, while the _Switch colors_ button is to the left. I don't think the design follows standard guidelines.

chennes:
Awesome, very nice work.

PR #9098, opened 2023-03-28
Add Conda Environment and VS Code configuration.

oursland:
Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [x]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---

This PR adds a conda-based development environment and configuration to easily stand up a development environment in Visual Studio Code.

Documentation for installation and use of the conda-managed dependencies and VS Code is found at [this GitHub repository](https://github.com/oursland/FreeCAD-Build-Notes).

A video walkthrough is available on [YouTube](https://youtu.be/2ujlBmywx5g).

Functionality includes:

* consistent dependency management via conda, similar to how the weekly builds are made
* CMake configuration
* building of FreeCAD
* C++ run and interactive debug functionality
* Python interactive debugging for core and workbench
* code navigation and code completion

PR #9175, opened 2023-04-03
[Toponaming] create ElementMap class

### LINKED ISSUE #9115: [Issue] [WIP] TopoNaming Phase 2 (card 2 of 5) - create ElementMap.cpp (and more to ElementMap.h)

### Forums discussion

_No response_

### Version

0.20 (Release)

### Full version info

```shell
[code]
Word size of FreeCAD: 64-bit
Build type: Debug
Branch: master
[/code]
```

### Subproject(s) affected?

None

### Issue description

> The toponaming algorithm needs a place to generatively express the history of an item.

## description

This class embodies the postfix support in `src/App/ComplexGeoData.cpp`

It is a serializable sorted mapping of items which generatively describe the elements and it's history.

The unit testing should mock up a variety of scenarios that would be seen in real-world projects and assert expected strings. This is critical as future changes to the class should be as consistent and non-breaking as possible.

This step also adds the `elementMap`/`_elementMap` members seen in the the ComplexGeoData class; though it is left empty when FreeCAD is running.

For context, see issue #8432

## notes

Please don't confuse the ElementMap class with the MappedElement class, which stores the IndexedName and MappedName data items.

### Anything else?

_No response_

### Code of Conduct

Pesc0:

---

fixes #9115

### DIFF
--- a/src/App/ComplexGeoData.h
+++ b/src/App/ComplexGeoData.h
@@ -166,20 +166,12 @@ public:
 
     /** @name Element name mapping */
     //@{
-    /// Special prefix to mark the beginning of a mapped sub-element name
-    static const std::string &elementMapPrefix();
-    /// Special postfix to mark the following tag
-    static const std::string &tagPostfix();
-    /// Special postfix to mark the index of an array element
-    static const std::string &indexPostfix();
-    /// Special prefix to mark a missing element
-    static const std::string &missingPrefix();
     /// Check if a subname contains missing element
     static bool hasMissingElement(const char *subname);
-    /** Check if the name starts with elementMapPrefix()
+    /** Check if the name starts with ELEMENT_MAP_PREFIX
      *
      * @param name: input name
-     * @return Returns the name stripped with elementMapPrefix(), or 0 if not
+     * @return Returns the name stripped with ELEMENT_MAP_PREFIX, or 0 if not
      * start with the prefix
      */
     static const char *isMappedElement(const char *name);
--- a/src/App/ComplexGeoData.h
+++ b/src/App/ComplexGeoData.h
@@ -164,33 +164,6 @@ public:
     virtual bool getCenterOfGravity(Base::Vector3d& center) const;
     //@}
 
-    /** @name Element name mapping */
-    //@{
-    /// Check if a subname contains missing element
-    static bool hasMissingElement(const char *subname);
-    /** Check if the name starts with ELEMENT_MAP_PREFIX
-     *
-     * @param name: input name
-     * @return Returns the name stripped with ELEMENT_MAP_PREFIX, or 0 if not
-     * start with the prefix
-     */
-    static const char *isMappedElement(const char *name);
-
-    /// Strip out the trailing element name if there is mapped element name precedes it.
-    static std::string newElementName(const char *name);
-    /// Strip out the mapped element name if there is one.
-    static std::string oldElementName(const char *name);
-    /// Strip out the old and new element name if there is one.
-    static std::string noElementName(const char *name);
-
-    /// Find the start of an element name in a subname
-    static const char *findElementName(const char *subname);
-
-    static inline const char *hasMappedElementName(const char *subname) {
-        return isMappedElement(findElementName(subname));
-    }
-    //@}
-
 protected:
 
     /// from local to outside
--- a/src/App/ElementNamingUtils.h
+++ b/src/App/ElementNamingUtils.h
@@ -7,27 +7,33 @@ namespace Data
 {
 
 /// Special prefix to mark the beginning of a mapped sub-element name
-const std::string ELEMENT_MAP_PREFIX                = ";";
+constexpr const char* ELEMENT_MAP_PREFIX                = ";";
+constexpr size_t      ELEMENT_MAP_PREFIX_SIZE           = 1;
 
 /// Special prefix to mark a missing element
-const std::string MISSING_PREFIX                    = "?";
+constexpr const char* MISSING_PREFIX                    = "?";
 
-const std::string MAPPED_CHILD_ELEMENTS_PREFIX      = ELEMENT_MAP_PREFIX + ":R";
+// IMPORTANT: For all the constants below, the semicolon ";"
+// at the start is ELEMENT_MAP_PREFIX
+
+constexpr const char* MAPPED_CHILD_ELEMENTS_PREFIX      = ";:R";
 
 /// Special postfix to mark the following tag
-const std::string POSTFIX_TAG                       = ELEMENT_MAP_PREFIX + ":H";
-const std::string POSTFIX_DECIMAL_TAG               = ELEMENT_MAP_PREFIX + ":T";
-const std::string POSTFIX_EXTERNAL_TAG              = ELEMENT_MAP_PREFIX + ":X";
-const std::string POSTFIX_CHILD                     = ELEMENT_MAP_PREFIX + ":C";
+constexpr const char* POSTFIX_TAG                       = ";:H";
+constexpr size_t      POSTFIX_TAG_SIZE                  = 3;
+
+constexpr const char* POSTFIX_DECIMAL_TAG               = ";:T";
+constexpr const char* POSTFIX_EXTERNAL_TAG              = ";:X";
+constexpr const char* POSTFIX_CHILD                     = ";:C";
 
 /// Special postfix to mark the index of an array element
-const std::string POSTFIX_INDEX                     = ELEMENT_MAP_PREFIX + ":I";
-const std::string POSTFIX_UPPER                     = ELEMENT_MAP_PREFIX + ":U";
-const std::string POSTFIX_LOWER                     = ELEMENT_MAP_PREFIX + ":L";
-const std::string POSTFIX_MOD                       = ELEMENT_MAP_PREFIX + ":M";
-const std::string POSTFIX_GEN                       = ELEMENT_MAP_PREFIX + ":G";
-const std::string POSTFIX_MODGEN                    = ELEMENT_MAP_PREFIX + ":MG";
-const std::string POSTFIX_DUPLICATE                 = ELEMENT_MAP_PREFIX + "D";
+constexpr const char* POSTFIX_INDEX                     = ";:I";
+constexpr const char* POSTFIX_UPPER                     = ";:U";
+constexpr const char* POSTFIX_LOWER                     = ";:L";
+constexpr const char* POSTFIX_MOD                       = ";:M";
+constexpr const char* POSTFIX_GEN                       = ";:G";
+constexpr const char* POSTFIX_MODGEN                    = ";:MG";
+constexpr const char* POSTFIX_DUPLICATE                 = ";D";
 
 
 /// Check if a subname contains missing element
--- a/src/App/ElementNamingUtils.h
+++ b/src/App/ElementNamingUtils.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <string>
+#include "FCGlobal.h"
 
 
 namespace Data
@@ -37,7 +38,7 @@ constexpr const char* POSTFIX_DUPLICATE                 = ";D";
 
 
 /// Check if a subname contains missing element
-bool hasMissingElement(const char *subname);
+AppExport bool hasMissingElement(const char *subname);
 
 /** Check if the name starts with elementMapPrefix()
  *
@@ -45,21 +46,21 @@ bool hasMissingElement(const char *subname);
  * @return Returns the name stripped with elementMapPrefix(), or 0 if not
  * start with the prefix
  */
-const char *isMappedElement(const char *name);
+AppExport const char *isMappedElement(const char *name);
 
 /// Strip out the trailing element name if there is mapped element name preceeds it.
-std::string newElementName(const char *name);
+AppExport std::string newElementName(const char *name);
 
 /// Strip out the mapped element name if there is one.
-std::string oldElementName(const char *name);
+AppExport std::string oldElementName(const char *name);
 
 /// Strip out the old and new element name if there is one.
-std::string noElementName(const char *name);
+AppExport std::string noElementName(const char *name);
 
 /// Find the start of an element name in a subname
-const char *findElementName(const char *subname);
+AppExport const char *findElementName(const char *subname);
 
-const char *hasMappedElementName(const char *subname);
+AppExport const char *hasMappedElementName(const char *subname);
 
 
 }// namespace Data
--- a/src/App/ElementNamingUtils.h
+++ b/src/App/ElementNamingUtils.h
@@ -30,5 +30,30 @@ const std::string POSTFIX_MODGEN                    = ELEMENT_MAP_PREFIX + ":MG"
 const std::string POSTFIX_DUPLICATE                 = ELEMENT_MAP_PREFIX + "D";
 
 
+/// Check if a subname contains missing element
+bool hasMissingElement(const char *subname);
+
+/** Check if the name starts with elementMapPrefix()
+ *
+ * @param name: input name
+ * @return Returns the name stripped with elementMapPrefix(), or 0 if not
+ * start with the prefix
+ */
+const char *isMappedElement(const char *name);
+
+/// Strip out the trailing element name if there is mapped element name preceeds it.
+std::string newElementName(const char *name);
+
+/// Strip out the mapped element name if there is one.
+std::string oldElementName(const char *name);
+
+/// Strip out the old and new element name if there is one.
+std::string noElementName(const char *name);
+
+/// Find the start of an element name in a subname
+const char *findElementName(const char *subname);
+
+const char *hasMappedElementName(const char *subname);
+
 
 }// namespace Data
--- a/tests/src/App/ElementMap.cpp
+++ b/tests/src/App/ElementMap.cpp
@@ -3,16 +3,198 @@
 #include "gtest/gtest.h"
 
 #include "App/ElementMap.h"
+#include <App/Application.h>
 
-#include <sstream>
+// NOLINTBEGIN(readability-magic-numbers)
+
+
+// this is a "holder" class used for simpler testing of ElementMap in the context of a class
+class LessComplexPart
+{
+public:
+    LessComplexPart(long tag, const std::string& nameStr)
+        : elementMapPtr(std::make_shared<Data::ElementMap>())
+        , Tag(tag)
+        , name(nameStr)
+    {
+        // object also have Vertexes etc and the face count varies; but that is not important
+        // here since we are not testing a real model
+        // the "MappedName" is left blank for now
+        auto face1 = Data::IndexedName("Face", 1);
+        auto face2 = Data::IndexedName("Face", 2);
+        auto face3 = Data::IndexedName("Face", 3);
+        auto face4 = Data::IndexedName("Face", 4);
+        auto face5 = Data::IndexedName("Face", 5);
+        auto face6 = Data::IndexedName("Face", 6);
+        elementMapPtr->setElementName(face1, Data::MappedName(face1), Tag);
+        elementMapPtr->setElementName(face2, Data::MappedName(face2), Tag);
+        elementMapPtr->setElementName(face3, Data::MappedName(face3), Tag);
+        elementMapPtr->setElementName(face4, Data::MappedName(face4), Tag);
+        elementMapPtr->setElementName(face5, Data::MappedName(face5), Tag);
+        elementMapPtr->setElementName(face6, Data::MappedName(face6), Tag);
+    }
+
+    Data::ElementMapPtr elementMapPtr;
+    mutable long Tag;
+    Data::MappedName name;
+};
 
 class ElementMapTest: public ::testing::Test
 {
+protected:
+    static void SetUpTestSuite()
+    {
+        int argc = 1;
+        char* argv[] = {"FreeCAD"};
+        App::Application::Config()["ExeName"] = "FreeCAD";
+        App::Application::init(argc, argv);
+    }
+
+    void SetUp() override
+    {
+        App::GetApplication().newDocument("test", "testUser");
+        sids = &_sid;
+    }
+
+    // void TearDown() override {}
+
+    Data::ElementIDRefs _sid;
+    QVector<App::StringIDRef>* sids;
 };
 
 TEST_F(ElementMapTest, defaultConstruction)
 {
     // Act
+    Data::ElementMap elementMap = Data::ElementMap();
 
     // Assert
+    EXPECT_EQ(elementMap.size(), 0);
 }
+
+TEST_F(ElementMapTest, setElementNameDefaults)
+{
+    // Arrange
+    Data::ElementMap elementMap;
+    Data::IndexedName element("Edge", 1);
+    Data::MappedName mappedName("TEST");
+
+    // Act
+    auto resultName = elementMap.setElementName(element, mappedName, 0, sids);
+    auto mappedToElement = elementMap.find(element);
+
+    // Assert
+    EXPECT_EQ(resultName, mappedName);
+    EXPECT_EQ(mappedToElement, mappedName);
+}
+
+TEST_F(ElementMapTest, setElementNameWithHashing)
+{
+    // Arrange
+    Data::ElementMap elementMap;
+    std::ostringstream ss;
+    Data::IndexedName element("Edge", 1);
+    Data::MappedName elementNameHolder(element);// Will get modified by the encoder
+    const Data::MappedName expectedName(element);
+
+    // Act
+    elementMap.encodeElementName(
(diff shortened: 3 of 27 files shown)


Pesc0:
Uhm... I think I need some help
```
src/Base/Console.h:355:28: error: ‘_s_fclvl’ was not declared in this scope
  355 | #   define FC_LOG_INSTANCE _s_fclvl
      |                            ^~~~~~~~
```

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175">pipeline status</a> for feature branch [PR_9175](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175). Pipeline [826781146](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/826781146) was triggered at [50e15c8](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/50e15c83e32897d1b5072508796d6f65c04b1cf2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175">pipeline status</a> for feature branch [PR_9175](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175). Pipeline [828829178](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/828829178) was triggered at [b3e6712](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/b3e67120015992e26dfc023018b62fb7ec2c315d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
If you add `[skip ci]` to your commit message it will not run CI.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175">pipeline status</a> for feature branch [PR_9175](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175). Pipeline [828953481](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/828953481) was triggered at [dcb5b4f](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/dcb5b4f32fa58d5459d14712665cf6da76b5fbb0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175">pipeline status</a> for feature branch [PR_9175](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175). Pipeline [829000923](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/829000923) was triggered at [bff603f](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/bff603f469af2d7e9d89511b5468164856ab569c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175">pipeline status</a> for feature branch [PR_9175](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175). Pipeline [836875996](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/836875996) was triggered at [11b5da3](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/11b5da36f68d20038f3f4e1ddbf595364ea42152). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175">pipeline status</a> for feature branch [PR_9175](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175). Pipeline [838453245](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838453245) was triggered at [b063c58](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/b063c5814f1749de794f056f5d8178ff1486b8d4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175">pipeline status</a> for feature branch [PR_9175](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175). Pipeline [838898376](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838898376) was triggered at [6d22e3e](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/6d22e3e7477557b36e5d122f74e05a8a8a7b6e46). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175">pipeline status</a> for feature branch [PR_9175](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9175). Pipeline [842282405](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/842282405) was triggered at [7fda894](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/7fda89499b87f6ae4a7c41101c0d2baa73de3b3e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

.vscode/c_cpp_properties.json
```diff
@@ -1,12 +1,16 @@
...
+                "/usr/include/QtCore",
+                "/usr/include/python3.11"
+            ],
+            "cStandard": "c17",
+            "cppStandard": "c++17",
+            "configurationProvider": "ms-vscode.cmake-tools"
+        }
+    ],
+    "version": 4
+}
```
adrianinsaval, line 16: I don't think you should touch these files unless you know what youa re doing and actually want to change the default for everyone

Pesc0: Afaik this just adds the two paths so that vscode can find the proper include headers (on linux of course). I've intentionally named the commit [vscode stuff: drop or merge?](https://github.com/FreeCAD/FreeCAD/pull/9175/commits/cf4651588916d857454ca017a1a4cc5f50703a37), so I'm perfectly fine reverting these changes before merge

AjinkyaDahale: @adrianinsaval we should probably remove this folder itself and add it to `.gitignore`

Pesc0: It was added recently by #9098

Pesc0:
rebased on master

.vscode/c_cpp_properties.json
adrianinsaval: adding to gitignore would be a good idea, when you actually want to make changes there you can still explicitly add them to the commit. Here that commit should be dropped IMO

src/App/ComplexGeoData.h
```diff
@@ -166,20 +166,12 @@ class AppExport ComplexGeoData: public Base::Persistence, public Base::Handled
...
     static bool hasMissingElement(const char *subname);
-    /** Check if the name starts with elementMapPrefix()
+    /** Check if the name starts with ELEMENT_MAP_PREFIX
      *
      * @param name: input name
-     * @return Returns the name stripped with elementMapPrefix(), or 0 if not
+     * @return Returns the name stripped with ELEMENT_MAP_PREFIX, or 0 if not
      * start with the prefix
      */
     static const char *isMappedElement(const char *name);
```
AjinkyaDahale, line 177: If we move `isMappedElement`, `newElementName` and `oldElementName` to `ElementMap` or a smaller header, `SketchObject` is no longer depenedent on this header and thus lowering overhead.

Pesc0:
Thanks for the review. I'd argue it's best to get elementmap working for now, this stuff can be moved quickly later in the complexgeodata card. Opinions (from everyone) on this?

.vscode/c_cpp_properties.json
Pesc0: I agree the commit is not really part of this PR. Still these changes made vscode happy on both of my machines (arch and fedora). Not trying to force this or anything, its just my opinion since this has been useful to me. Unless someone is favourable on keeping this I will drop it before merge.

chennes: It probably doesn't make sense to have the include path changes in there, since they are system-specific. The rest of the changes are fine (though I agree that we should add the files to .gitignore).

AjinkyaDahale:
@Pesc0 if it is not done now that is slightly more work to change the includes in cards 4 (and 3, if these changes are done outside of that) whenever the changes are made. If you believe making these changes right now will need more time (because of testing etc.) than you have, I'm OK with delaying, but otherwise I think it's best done here.

Pesc0:
@AjinkyaDahale I had some time today to look more more in depth, here are my observations:
- All the functions you mentioned are static helpers, quite simple to move. 
- At the same time they seem to touch a lot of files, so it would be better to chose wisely: is ElementMap the right place to put them?
- Also they seem to be part of a block of helpers: you didn't mention `hasMissingElement`, `noElementName`, `findElementName`, `hasMappedElementName`. I think the block should be moved all together, not just the ones you mentioned. 

Given all of this, I was reluctant towards this change simply because we are struggling to merge this, so it did not seem a good idea to add more stuff on top of it. Since this seems a simple change, would you be able to submit a PR? I would do it myself but since I don't know the codebase very well it would take quite some time.

AjinkyaDahale:
>  Since this seems a simple change, would you be able to submit a PR? I would do it myself but since I don't know the codebase very well it would take quite some time.

@Pesc0, are you asking me to make a PR to your branch or directly to the main?. Also, at this point I believe you know better about this part of the code than I do.

Pesc0:
Alright I've managed to move them to PostfixStringReferences, as it seemed a better place to me (since they are static functions).  I'm thinking of renaming PostfixStringReferences to ElementNamingUtils. Pushing changes soon

Pesc0:
@AjinkyaDahale I did not have to touch `SketchObject` (that you mentioned would drop the complexgeodata dependency). Could you check that these changes actually do what you were thinking of?

AjinkyaDahale:
> @AjinkyaDahale I did not have to touch `SketchObject` (that you mentioned would drop the complexgeodata dependency). Could you check that these changes actually do what you were thinking of?

I think there was a miscommunication. ComplexGeoData isn't currently used in SketchObject in master, but it is planned in card 4.

Pesc0:
Ah right, I've missed that. Still, do these changes reflect what you expected?

src/App/ElementNamingUtils.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <string>
+
+
+namespace Data
+{
+
+/// Special prefix to mark the beginning of a mapped sub-element name
+const std::string ELEMENT_MAP_PREFIX                = ";";
```
AjinkyaDahale, line 10: Please see https://github.com/FreeCAD/FreeCAD/pull/8923#issuecomment-1476693963 and the link within it :D.

AjinkyaDahale:
@Pesc0 sorry for the late reply. I just checked and to answer your question, it indeed does reflect the changes. Thanks again.

src/App/ElementNamingUtils.h
Pesc0: Ah snap I've missed that. See also the discussion here https://github.com/FreeCAD/FreeCAD/commit/9af8e6a0556bd6def909a2787613ab0af698a06a#commitcomment-113554654.
Is there a better way to do this?

AjinkyaDahale: Not sure of the variable naming convention. I would suggest staying with the rest of FreeCAD there.

Coming to the optimization, Pretty sure that would be asking too much. It could consider that a future file may import and want to use the string related functions, so it'll create the whole object rather than just the string. Perhaps that's why it's a strict no for libraries. This is all thought experiment though.

JohnAD:
The latest PR I pushed to your `elementmap` branch has a fix for handling the singleton better. Essentially, before, all four tests would pass just fine when you ran one at a time. But if you ran all four of them in a row, the Setup/Teardown would create a run-time error. So, it now does a suite-wide setup for the `Application` singleton, and the individual test `Setup` only creates a new Document and leaves the singleton alone.

Pesc0:
@JohnAD Thanks! I've left a comment there

JohnAD:
We will probably want to merge to master soon as this branch and master are starting to diverge quite a bit.

I'd like to add quite a few more "mocked operation" tests to the unit testing. But, frankly, I think that could be done later after the merge.

It would, however, be good to get the remaining basic unit tests done; at _least_ a tiny one per public method (except the save/restore methods):

- [ ] MappedName addName(MappedName& name, const IndexedName& idx, const ElementIDRefs& sids, bool overwrite, IndexedName* existing);
- [ ] MappedName setElementName(const IndexedName& element, const MappedName& name, long masterTag, const ElementIDRefs* sid = nullptr, bool overwrite = false);
- [x] void encodeElementName(char element_type, MappedName& name, std::ostringstream& ss, ElementIDRefs* sids, long masterTag, const char* postfix = 0, long tag = 0, bool forceTag = false) const;
- [ ] MappedName hashElementName(const MappedName& name, ElementIDRefs& sids) const;
- [ ] MappedName dehashElementName(const MappedName& name) const;
- [ ] bool erase(const MappedName& name);
- [ ] bool erase(const IndexedName& idx);
- [x] unsigned long size() const;
- [ ] bool empty() const;
- [ ] IndexedName find(const MappedName& name, ElementIDRefs* sids = nullptr) const;
- [ ] MappedName find(const IndexedName& idx, ElementIDRefs* sids = nullptr) const;
- [ ] std::vector<std::pair<MappedName, ElementIDRefs>> findAll(const IndexedName& idx) const;
- [ ] bool hasChildElementMap() const;
- [ ] void hashChildMaps(long masterTag);
- [ ] void addChildElements(long masterTag, const std::vector<MappedChildElements>& children);
- [ ] std::vector<MappedChildElements> getChildElements() const;
- [ ] std::vector<MappedElement> getAll() const;

The only reason I think to push off the unused save/restore methods is that it really isn't very testable until the next card IMO (if not the next phase overall).

Does this sound like a finishing plan? @chennes ?

Pesc0:
I've moved things around a bit, hopefully this does not mess up what you were working on.
> at least a tiny one per public method

Note that I have made some methods private, since in RT's branch they seemed to be used only internally. See commit https://github.com/FreeCAD/FreeCAD/pull/9175/commits/5730c4e366ee358cb79ad7824046cfd41ef583da. Then seeing you had used `getAll()` in your test I've reverted some of those changes https://github.com/FreeCAD/FreeCAD/pull/9175/commits/61886b4847b00102d9a97e9ebe5a6f4466cfcdba. I think the simpler the interface the better, thats why I moved as much as possible to private. 

> this branch and master are starting to diverge quite a bit

Not true, I have been rebasing frequently ;)

JohnAD:
So that we don't accidentally duplicate effort: I would like to write the unit tests for hasChildElementMap, hashChildMaps, addChildElements, and getChildElements next.

Also, I'm increasing not liking the name `masterTag` that I came up with for the current tag parameter. How does `selfTag` or `currentTag` or `ownTag` sound to you?

Pesc0:
> So that we don't accidentally duplicate effort: I would like to write the unit tests for hasChildElementMap, hashChildMaps, addChildElements, and getChildElements next.

Before you do that could you please ensure that in realthunder's branch they are actually used externally? These are the same functions that I touched in the last commit I mentioned, and I'm not sure if they should be private or not. Not that testing them would hurt, but if they are just used internally we should test the public stuff first.

> Also, I'm increasing not liking the name masterTag that I came up with for the current tag parameter. How does selfTag or currentTag or ownTag sound to you?

Could you please describe briefly what exactly does the tag do / mean?

JohnAD:
> Could you please describe briefly what exactly does the tag do / mean?

I'm running from the context of the code, so it is possible my understanding is not full or complete. @chennes,  please correct me if wrong.

But it is my understanding that the tag is used to uniquely id an object/face/vector/etc .

In the simple case, it can be ignored mostly. But when a operation is involved, such as a union, this becomes more important.

So, if a cube (tag = 1) and another cube (tag = 2) are "fused" into a new shape (tag = 3); and you follow through the history of the new shape, you can identify the "current tag" (3) is different than the other objects (1 and 2). This helps prevent duplication of the same objects when serializing the toponame.

It's **very** possible that tags are used for other things also.

Complicating things further, elements of an object, such as faces, vectors(lines) and points also have unique tags. Again, it is to distinguish between the items involved in any operation.

JohnAD:
- `ComplexGeoData::getMappedChildElements` directly calls `getChildElements`

    -  however nothing calls `getMappedChildElements`. So, you can delete this one and leave a note in card 3's Issue if you would like.

- `ComplexGeoData::hasChildElementMap()` calls ElementMap -> `hasChildElementMap` and that is called by Mod/Part/App/PropertyTopoShape.cpp

- `ComplexGeoData::hashChildMaps()` calls ElementMap -> `hashChildMaps` and that is called by Mod/Part/App/PropertyTopoShape.cpp

- `ComplexGeoData::setMappedChildElements` directly calls `addChildElements`. That is called by `ComplexGeoData::setMappedChildElements`. That is called by Mod/Part/App/PropertyTopoShape.cpp

For these last three, it is possible that the ComplexGeoData methods might be removed if we add a direct ElementMapPtr getter. But that is for the next card to consider.

Pesc0:
Thanks for the explanation, it cleared things up a bit. I'm thinking `elementTag`? or `currentTag`. self / own make me think the tag is referring to the map itself.

As for the functions it looks like they should be public then. Thanks for investigating that.

Pesc0:
Looks like addName, hashElementName, dehashElementName are used only internally as well? Should I make them private?

JohnAD:
`addName` .... yes, private. `hashElementName` ... private. `dehashElementName` ... also private.

tests/src/App/ElementMap.cpp
```diff
@@ -3,16 +3,198 @@
 #include "gtest/gtest.h"
 
 #include "App/ElementMap.h"
+#include <App/Application.h>
```
AjinkyaDahale, line 6: Is there any reason this and the above headers are imported with different formats?

Pesc0:
When I have a bit of time I should be able to add the remaining tests, not including the ones John is working on of course.

AjinkyaDahale:
@Pesc0 at some point you should probably also look into squashing the commits to a reasonable number. For example, all fixes can get merged with the commit they fix.

Pesc0:
Except for a failing test that should be investigated I think this is now pretty much ready.

@AjinkyaDahale I would do that but at this point I dont even remember what I did and for what reason, and I'm really afraid of breaking stuff.

AjinkyaDahale:
> @AjinkyaDahale I would do that but at this point I dont even remember what I did and for what reason, and I'm really afraid of breaking stuff.

I would advise creating a new branch from this one and experimenting with it. It's more important if you can't recall your steps. This can tell if you missed something important, for example.

Pesc0:
I've just tried reorganizing and squashing the commits, but it did not work so well. The end result (comparing the branches before and after the operation) is that the files are not the same.

AjinkyaDahale:
@Pesc0 thanks for checking it out. @JohnAD do you have any opinion on the squashing?

Pesc0:
I think it's a good idea to reorganize and rename the commits. The difference between branches most likely comes from the fact that I've messed up something.

JohnAD:
I generally squash a PR into a single commit; letting the single commit message and the changes themselves express the meaning of the merge.

But if you are wanting to preserve commit history while still doing some squashes...that can get really complicated. Git isn't really designed to handle that easily.

JohnAD:
IIRC, squashing to a single commit is an option in GitHub's PR process. So, if you ware willing to have a single commit, write up the wording you want Chris to use when completing the PR.

AjinkyaDahale:
@JohnAD I would steer away from squashing into a single commit. If you do choose that, Id suggest keeping the text of all the commits.

chennes:
I don't know that there's a lot of value in preserving the commit history here -- my strategy with the Toponaming work has been to make one commit that is as close as possible to just real thunder's original code, and then a subsequent commit (possibly more than one) that is my additions and changes. So if this can be arranged like that then some squashing and cleanup may be in order. Otherwise, let's just squash the whole thing on merge.

chennes:
Any progress on that failing `ElementMapTest.eraseIndexedName` test? I obviously don't want to merge until we are passing CI...

Pesc0:
I've written the test intentionally failing because to me that's what makes the most sense. The question is if the elementmap code is correct, not the test. 

I would like to get an ACK from @JohnAD on the other tests I've added as well, just to ensure I didn't write anything completely bogus.

chennes:
I can't merge while CI fails -- what's the path forward here?

JohnAD:
On first examination, the `erase` is not erasing the indexed name and it's mapping. Instead, it is erasing the first matching type of indexed name. So, erasing "Edge1" will erase "Edge*". So erasing "Edge1" again will erase "Edge*" and if "Edge2" is in the list of indexed names, it will erase that one next.

I'm not done investigating. I'm going to try to "deduce meaning" from usage the original toponaming branch. Guessing, either:

1. it is badly named and should be renamed something like `eraseFirstMatchingNameType`.  (Where "type" means a string pointer to "Edge". I really don't like that `IndexedName` call the string prefix a `type`. But that is a different discussion.) or,
2. this never comes up as an issue in most scenarios, but we should add some extra code to make this work in all scenarios anyway.

I have a suspicion it is scenario 1. But I'll report back soon.

JohnAD:
Actually, even as I hit send on the above commend, this can't possibly be correct behavior. I'll do more homework :smile: .

JohnAD:
Yeah, it's just a bug. And, it was probably never caught because the `bool` that is returned is never used.

So, I'll going to remove the return value since it adds to value. But, i'll also add to the test to make sure the "same type" does not over-erase. I'll send a quick PR.

John

Pesc0:
Thanks John, I've mirrored the same changes to the other erase function, `erase(const MappedName& name)`

Pesc0:
Managed to squash things a bit. Perhaps last time I was too aggressive with the squashing.

Should be ready to merge, but before that please someone take a look at the tests in particular, let's not make the mistake of merging broken stuff that will bite us in the future.

Pesc0:
does this work? I am on linux so I'll have to wait for CI

JohnAD:
All pipeline checks pass! Unfortunately I don't have access to a Windows machine to try running it as a final check.

Pesc0:
CI reports no issues, I think there's no need for you to run it John, thanks for being available though.

chennes:
Let's do it, then!

luzpaz:
Heads-up, this PR added trailing whitespace

```
Warning: src/App/ElementMap.h:59:    /** Default constructor: hooks internal functions to \c signalSaveDocument and <-- trailing whitespace
Warning: src/App/ElementMap.h:60:     * \c signalStartRestoreDocument. This is related to the save and restore process <-- trailing whitespace
Warning: src/App/ElementMap.h:68:     * <-- trailing whitespace
Warning: src/App/ElementMap.h:76:     * @param s: serialized stream <-- trailing whitespace
Warning: src/App/ElementMap.h:80:    /** Deserialize and restore this map. This function restores \c childMaps and <-- trailing whitespace
Warning: src/App/ElementMap.h:81:     * \c postfixes from the stream, then calls the other (private) restore function with those <-- trailing whitespace
Warning: src/App/ElementMap.h:186:     * @param s: serialized stream <-- trailing whitespace
Warning: src/App/ElementMap.h:208:     * @param existing: out variable: if not overwriting, and \c name is already <-- trailing whitespace
Warning: src/App/ElementMap.h:216:     * if it was not present in the map. <-- trailing whitespace
Warning: src/App/MappedName.cpp:75:        pos = this->rfind(POSTFIX_DECIMAL_TAG); <-- trailing whitespace
Warning: src/App/MappedName.cpp:110:        //                     |              |   |   |  <-- trailing whitespace
```

PR #9176, opened 2023-04-03
CI: initial release of Windows build workflow

0penBrain:
Well, this is what I ended up as an initial Windows workflow.
It has following structure :
* Uses ccache as a caching program
* Builds with MSVC
* 2 actions are created so we can use a GH cached version of LibPack and CCache to reduce compilation time
* Only CLI tests (unit tests seems to not be here -- didn't investigate much -- and I didn't looked at GUI tests yet)
* No report is generated yet, just the status is reported in the summary (PowerShell isn't my friend actually)

I tried to optimize running delay but still it's generally longer than the Ubuntu builds (from about 5 to 15 min) and so will be the bottleneck of the CI.

One of things that takes some minutes is copying the libpack bin folder to build directory.
I tried to instead add the folder to PATH but tests fail. So it looks like it's not enough and actually there are some dependencies in it that aren't handled by the "COPY_DEPENDS" flag.

### DIFF
--- a/.github/workflows/CI_master.yml
+++ b/.github/workflows/CI_master.yml
@@ -43,13 +43,19 @@ jobs:
     needs: [Prepare]
     uses: ./.github/workflows/sub_buildUbuntu2004.yml
     with:
-      artifactBasename: Build2004-${{ github.run_id }}
+      artifactBasename: Ubuntu_20-04-${{ github.run_id }}
 
   Ubuntu_22-04:
     needs: [Prepare]
     uses: ./.github/workflows/sub_buildUbuntu2204.yml
     with:
-      artifactBasename: Build2204-${{ github.run_id }}
+      artifactBasename: Ubuntu_22-04-${{ github.run_id }}
+
+  Windows:
+    needs: [Prepare]
+    uses: ./.github/workflows/sub_buildWindows.yml
+    with:
+      artifactBasename: Windows-${{ github.run_id }}
 
   Lint:
     needs: [Prepare]
@@ -61,7 +67,7 @@ jobs:
       changedPythonFiles: ${{ needs.Prepare.outputs.changedPythonFiles }}
 
   WrapUp:
-    needs: [Prepare, Ubuntu_20-04, Ubuntu_22-04, Lint]
+    needs: [Prepare, Ubuntu_20-04, Ubuntu_22-04, Windows, Lint]
     if: always()
     uses: ./.github/workflows/sub_wrapup.yml
     with:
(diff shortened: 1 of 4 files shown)


xtemp09:
Which version of Windows is CI running? Are the artifacts downloadable?

0penBrain:
> Which version of Windows is CI running?

Windows Server 2022

> Are the artifacts downloadable?

There are no artifacts, so I'd say no :)

xtemp09:
Which version of MSVC? Is it possible to use clang?

0penBrain:
> Which version of MSVC?

You can read here https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md or look at the CMake configure step log.

Is it possible to use clang?

Probably it is, but as goal is to screen specific issue to MSVC, doesn't seem ideal.

chennes:
Wow, still going! It's been two hours so far...

0penBrain:
> Wow, still going! It's been two hours so far...

Yes, no cache exists yet. It could take 4 to 5 hours. :|

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9176">pipeline status</a> for feature branch [PR_9176](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9176). Pipeline [826781156](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/826781156) was triggered at [dd7af04](https://github.com/FreeCAD/FreeCAD/pull/9176/commits/dd7af041e287bb5d95bb4c182e1b4fe4018f0d56). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
> Wow, still going! It's been two hours so far...

Finally it was fast. :) 2.5 hours, compared to about 2 hours for Ubuntu to complete without cache. This is toughly the overhead I can see (+25% for Win build).

chennes:
Is this waiting on anything? Just someone to push the big red button?

0penBrain:
> Is this waiting on anything? Just someone to push the big red button?

Basically yes. Probably it should be pushed at a quiet moment because all CI checks will last 2.5 hours till the first completes and generates a cache. 😉

chennes:
OK, it's the middle of the night for most of our devs... let's merge this and see what happens :).

adrianinsaval:
> One of things that takes some minutes is copying the libpack bin folder to build directory.

maybe you can just create links for these instead of copying, I would expect that would be faster but I could be wrong

0penBrain:
> maybe you can just create links for these instead of copying, I would expect that would be faster but I could be wrong

I have a plan to test that is just to move the folder so it's just one inode to process that should be almost instantaneous. ;)

Issue #9179, opened 2023-04-04
[Feature Request] Add Keyboard Shortcuts to Preference Packs


howie-j:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

https://forum.freecad.org/viewtopic.php?t=75948

### Subproject(s) affected?

None

### Idea description

There is currently no way to export/backup/restore all keyboard shortcuts from the gui, other than backing up user.cfg.

It would be nice to add keyboard shortcuts to the preference pack templates, so they could be exported/imported/shared easily.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #9180, opened 2023-04-04
[Preference Packs] Add Keyboard Shortcuts Template

howie-j:
An attempt to solve issue #9179

I don't know if i understand pref packs correctly, but it seems like there needs to be a tag for every keyboard shortcut for pref packs to save modified shortcuts?

What i did was export all commands from a clean/unmodified freecad setup, with all default WB's enabled and loaded, and added those with its corresponding shortcut to an xml (.cfg) file.

---


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9180">pipeline status</a> for feature branch [PR_9180](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9180). Pipeline [827887396](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/827887396) was triggered at [81fdac8](https://github.com/FreeCAD/FreeCAD/pull/9180/commits/81fdac808e2ac6285624a841f3cfe8db967396b9). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Yes, it looks like this is correct. I will try to give it a test later today or tomorrow to be sure.

howie-j:
Thanks a lot @chennes!

Just a follow up question; Doesn't this approach with xml templates means that only "default" parameters can be saved in preference packs, not custom parameters such as custom toolbars for example?

So if i want to "export" custom toolbars, i have to do that outside of the pref pack manager using a separate macro/function (or manually copying from my user.cfg), and only importing would work via the pref pack gui?

chennes:
Yep. Exactly right. Future work might be to have the pref pack system also be able to run scripts to generate parts of the saved file.

PR #9199, opened 2023-04-06
Image: Merge taskboxes into one 'Image plane Settings'.

PaddleStroke:
- Allow user to position the image with a X / Y distance.
- Allow user to rotate the image on its plane.
- Adds Transparency.
- Image sizes change from PX to System unit as PX don't make sense for the user.
- Interactive tool reworked.
image

TODO: 
- See if we can get arrows similarly as in the 'Transform' taskbox to move the image on its XY plane.

---


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [829941677](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/829941677) was triggered at [929922b](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/929922bf69b5b55055091b535504c4a204c82209). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
@wwmayer I just wanted to double check with you, but is there a reason why imageplane is using px as size unit? I can't see it making any sense but maybe I'm missing the point?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [830133707](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/830133707) was triggered at [0318335](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/0318335f1c3a91c0009599f403810533cd0b9f63). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> I just wanted to double check with you, but is there a reason why imageplane is using px as size unit? I can't see it making any sense but maybe I'm missing the point?

I have looked at GIMP recently and there is an option to show the image size in px or mm or inches are some other format. So, I was thinking about to add such a combo box too where the user can switch between px and the preferred unit scheme. But maybe you are right and that there is no benefit for the user to know the size in pixels.

PaddleStroke:
These 2 first commits can be already reviewed/merged first I think. 
I'm moving on the 2 points length tool which is pretty independant from those 2 commits and will need some time to get ready. So I can push those in a second PR.

One question regarding those first changes. There's a restore() function where the widget are initialized. Is there a reason why to do it here rather than in the TaskImage ctor?

PaddleStroke:
The orientation part of the code was heavily modified to allow rotating the picture.
If you had other plans for this TaskOrientation, then we can just leave it here (rather than deleting it as I did) and not use it for TaskImage?

src/App/ImagePlane.h
```diff
@@ -43,6 +43,11 @@ class AppExport ImagePlane : public App::GeoFeature
     App::PropertyLength       XSize;
     App::PropertyLength       YSize;
 
+    double getXSize();
+    double getYSize();
```
wwmayer, line 47: Both methods should be declared `const`. But actually all the new methods are superfluous because they do nothing than directly accessing the property to get or set the value.

src/Gui/TaskView/TaskImage.cpp
```diff
@@ -48,38 +48,65 @@
...
+  , ui(new Ui_TaskImage)
   , feature(obj)
-  , aspectRatio{1.0}
+  , aspectRatio(1.0)
 {
     ui->setupUi(this);
     ui->pushButtonCancel->hide();
-    ui->spinBoxWidth->setValue(obj->getXSizeInPixel());
-    ui->spinBoxHeight->setValue(obj->getYSizeInPixel());
+
```
wwmayer, line 67: The constructor is fairly big and it's best to move parts of the code here to some functions:
* The initialization of the sliderTransparency can be moved to its own function
* The calls of connect() can be moved to its own function

src/Gui/TaskView/TaskImage.cpp
```diff
@@ -89,33 +116,51 @@ TaskImageScale::~TaskImageScale()
...
+    if (prop && prop->getTypeId().isDerivedFrom(App::PropertyInteger::getClassTypeId())) {
+        auto Transparency = static_cast<App::PropertyInteger*>(prop);
+        Transparency->setValue(val);
+
+        bool blocked = ui->spinBoxTransparency->blockSignals(true);
+        ui->spinBoxTransparency->setValue(val);
+        ui->spinBoxTransparency->blockSignals(blocked);
+        blocked = ui->sliderTransparency->blockSignals(true);
+        ui->sliderTransparency->setValue(val);
+        ui->sliderTransparency->blockSignals(blocked);
```
wwmayer, line 135: To keep the code clean and understandable use a QSignalBlocker

src/App/ImagePlane.h
PaddleStroke: Function removed.

wwmayer:
> If you had other plans for this TaskOrientation, then we can just leave it here (rather than deleting it as I did) and not use it for TaskImage?

Yes, please revert the removal.

When looking at the dialog I find it a bit odd that Transparency is part of the Placement group.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [830580396](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/830580396) was triggered at [171935f](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/171935f1e1b597eb5068cf3d81568c167d3d6928). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
I reversed the removal. 

Regarding transparency it's true that it's wrongly placed.
I removed the 'placement' groupbox and put the content directly in the taskbox. I think this groupbox was superfluous. 

Besides a question regarding the distance tool. I think it would look better if we use the same geometry as the sketcher distance constraint, the arrows with the distance writen in the middle, rather than a simple line.
I had a look at how it's done in sketcher, and it's implemented in sketchergui `SoDatumLabel`. I was wondering if there is something identical in core? Because I have seen there is a lot of SoXXX files such as `SoTextLabel`, and I don't know how to preview what each looks like. So I wondered maybe these double arrows are already here somewhere?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [830592180](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/830592180) was triggered at [c20ca5a](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/c20ca5a4b7905604fe8b13516a4e69631beb7e99). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
Instead of copy pasting and customizing SoDatumLabel, I moved it to core so that it can be reused anywhere.
I think it's best to share this kind of graphics accross the application for consistency (as of now every kind of measure tool has a different graphic).
@wwmayer @abdullahtahiriyo Can you please let me know if this is acceptable? And if it is done correctly (it builds and work on my side).

Note: SoDatumLabel.cpp had a `#include <Mod/Sketcher/SketcherGlobal.h>`. Removing it does not affect anything so it seems to be unecessary. So I removed it so that core do not rely on a mod.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [830965689](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/830965689) was triggered at [dcefc59](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/dcefc5996894f8fd801586a31873efddf0134b17). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> Can you please let me know if this is acceptable?

That should be fine since the class  doesn't depend on anything sketcher-specific.

> Removing it does not affect anything so it seems to be unnecessary.

Not quite. It provides the export macro SketcherGuiExport. You should replace the include with `#include <FCGlobal.h>` because otherwise when checking the lint warnings it complains about the syntax because it assumes GuiExport is the class name.

PaddleStroke:
Thanks for your feedback.

Ok so I finished implementing this nicer label for the interact tool.
Now I want to find a way to draw the spinbox directly on the 3Dview as in following mockup : 
image

I think the ability of doing this will be very handy in a lot of places (sketcher, partdesign...). So here is the perfect opportunity to setup this capability if it doesn't exist yet.

Do you perhaps have some hints how this should be done ? I mean the global architecture.

PaddleStroke:
Thanks for your feedback.

Ok so I finished implementing this nicer label for the interact tool.
Now I want to find a way to draw the spinbox directly on the 3Dview as in following mockup : 
image

I think the ability of doing this will be very handy in a lot of places (sketcher, partdesign...). So here is the perfect opportunity to setup this capability if it doesn't exist yet.

Do you perhaps have some hints how this should be done ? I mean the global architecture.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [831144892](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/831144892) was triggered at [1ef5663](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/1ef5663371f9acf3d0c79dc28560f66c6241d5aa). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
I have no hints, but that is exactly the user interface I imagined for this tool, I love it.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [831181501](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/831181501) was triggered at [60c6811](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/60c6811053e2f19601d7d18e3d22649cded1381d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
Ah so it's actually very simple : 
```
            Gui::MDIView* mdi = Gui::Application::Instance->activeDocument()->getActiveView();

            QuantitySpinBox* box = new QuantitySpinBox(mdi);
            box->setValue(50);
            box->show();
```

Then we 'just' need to set the position correctly.

PaddleStroke:
Hmm do you know how I can get the pixel position in the MDI widget of a point in the 3D view ?
image

I can move the widget, but it needs a position (d1,d2) in pixel from the top left corner of the MDI widget to position at the correct position.
The wanted position is known in the 3D model in (x,y,z).

Then there is the matter of updating the spinbox position based on the camera movement. So we need to connect camera movements to `updateSpinboxPosition()`

wwmayer:
Use View3DInventorViewer::getPointOnScreen to convert 3d into 2d coordinates. But note, that the returned SbVec2s is in Inventor coordinates that begin at bottom left, not at top left. So, to get it in Qt coordinates you have to compute `y = height - y`.

> Do you know what signal could be used to catch all camera movement (paning, zooming rotating)?

There is currently nothing implemented to notify widgets. An option is to implement a special SoEngine that emits a signal when one of the fields of an camera node has changed.

wwmayer:
> Do you perhaps have some hints how this should be done ? I mean the global architecture.

Adding a widget as a child of the 3d viewer works but you cannot rotate the widget as shown in your image. For that case you have to use the capabilities of the QGraphicsView.

In FreeCAD we have changed the Quarter code so that QuarterWidget doesn't inherit from QopenGLWidget but QGraphicsView which opens a lot more possibilities adding widgets to the 3d view. So, an alternative is to use a QGraphicsProxyWidget that contains the spin box and can be added to the QGraphicsScene.

However, when rotating the item then the rendered widget looks very poor. To improve the render quality one needs a specialized graphics proxy widget as used in this example: https://doc.qt.io/qt-6/qtwidgets-graphicsview-embeddeddialogs-example.html

chennes:
> you cannot rotate the widget as shown in your image.

I'm not sure it really needs to be rotated, though -- that seems like some needless complication that will actually get harder to use as the angle gets larger. @PaddleStroke are you committed to the rotation? Or can it just be level?

PaddleStroke:
actually I made it rotated in my mockup but that was an innacuracy. I think rotating it will complicate things a lot and will make it hard for the user to read. So it will be level.

PaddleStroke:
So View3DInventorViewer::getPointOnScreen works, but this function has a bug with windows when you use windows scaling: 
image

My screen being very big, I usually use 125% to scale things a little bit on windows.
The result is that the View3DInventorViewer::getPointOnScreen returned point is offset by 25% from where it whould really be.
Ie, say the mdi size is initially 1250px. With windows app scaling it is actually 1000px. But apparantly getPointOnScreen still consider that the mdi size is 1250px.

wwmayer:
> So View3DInventorViewer::getPointOnScreen works, but this function has a bug with windows when you use windows scaling:

OK, it looks like it's an issue with the device pixel ratio.

Currently, getPointOnScreen is only used in ViewProviderSketch::preselectAtPoint and ViewProviderSketch::moveCursorToSketchPoint. Can you check whether this works as expected, please?

wwmayer:
When I do `export QT_SCALE_FACTOR=1.25` before running FreeCAD, create a box and run this code
```
view.getPointOnScreen(App.Vector(0,0,0))
view.getCursorPos()
```
then the results are exactly off by 25%

PaddleStroke:
Note that I did not measure precisely, it looks like 25%
I'll test later the sketcher fcts

wwmayer:
If you start FreeCAD what does this give you:
```
from PySide import QtWidgets
QtWidgets.QApplication.instance().devicePixelRatio()
```
?

wwmayer:
After all the output of getPointOnScreen() is correct because the coordinates are taken of the viewport of the Inventor viewer. The device pixel ratio only must be considered when converting the position into a Qt coordinates.

Therefore I have implemented convenience functions toQPoint and fromQPoint that also flips the y coordinate.

And to clarify that getPointOnScreen() returns Inventor coordinates I have renamed the method to getPointOnViewport()

See https://github.com/FreeCAD/FreeCAD/pull/9216

wwmayer:
For moving the quantity spin box here is some test code to achieve this quite easily:
```
#include <Inventor/sensors/SoNodeSensor.h>
#include <Inventor/nodes/SoOrthographicCamera.h>


struct NodeData {
    View3DInventorViewer* viewer;
    Gui::QuantitySpinBox* sb;
    SbVec3f coord;
};

...

View3DInventorViewer* viewer = ...
Gui::QuantitySpinBox* sb = new Gui::QuantitySpinBox(viewer->parentWidget());
sb->show();

NodeData* info = new NodeData{viewer, sb, SbVec3f(0,0,0)};

SoNodeSensor* nodeSensor = new SoNodeSensor([](void * data, SoSensor * sensor) {
    NodeData* info = static_cast<NodeData*>(data);
    auto pos2d = info->viewer->getPointOnViewport(info->coord);
    auto pnt2d = info->viewer->toQPoint(pos2d);
    info->sb->move(pnt2d);
}, info);

nodeSensor->attach(viewer->getCamera());
```

The basic idea is to use a SoNodeSensor and attach it to the camera node. Every time when a field of the camera changes the lambda expression will be invoked. Inside there the new Qt position of the 3D point is computed.

I have tested it inside TaskImageScale::onInteractiveScale() and it works very well. 

This is the complete test code:
```

#include <Inventor/sensors/SoNodeSensor.h>
#include <Inventor/nodes/SoOrthographicCamera.h>

struct NodeData {
    View3DInventorViewer* viewer;
    Gui::QuantitySpinBox* sb;
    SbVec3f coord;
};

void TaskImageScale::onInteractiveScale()
{
    if (!feature.expired() && !scale) {
        View3DInventorViewer* viewer = getViewer();
        if (viewer) {
            auto vp = Application::Instance->getViewProvider(feature.get());
            scale = new InteractiveScale(viewer, vp);
            connect(scale, &InteractiveScale::selectedPoints,
                    this, &TaskImageScale::selectedPoints);
            Gui::QuantitySpinBox* sb = new Gui::QuantitySpinBox(viewer->parentWidget());
            sb->show();

            NodeData* info = new NodeData{viewer, sb, SbVec3f(0,0,0)};

            SoNodeSensor* nodeSensor = new SoNodeSensor([](void * data, SoSensor * sensor) {
                NodeData* info = static_cast<NodeData*>(data);
                auto pos2d = info->viewer->getPointOnViewport(info->coord);
                auto pnt2d = info->viewer->toQPoint(pos2d);
                info->sb->move(pnt2d);
            }, info);

            nodeSensor->attach(viewer->getCamera());
        }
    }

    if (scale) {
        if (scale->isActive()) {
            acceptScale();
        }
        else {
            startScale();
        }
    }
}

```

PaddleStroke:
@wwmayer thanks it works great ! And the bug with windows scaling is gone with the getPoint method. Please have a look at the functionality and let me know if you find any issues.

Remaining ideas of improvement : 
- Cancel interactive tool when right clicking or pressing ESC.
- Add arrows similar to Transform tool to move the image on X,Y and rotate.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [832005683](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/832005683) was triggered at [c673e85](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/c673e855651062cc0f868ee2a89430c6c87f627a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
Hmm actually it works properly only on X,Y plane.
The SoDatumLabel is not drawing correctly on other planes.
I thought setting the plane of the SoDatumLabel was done with the SoDatumLabel::norm variable, but I must be missing something else.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [832010690](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/832010690) was triggered at [a24b96b](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/a24b96bba5af075c873a060c6602f2f288782348). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
So it seems that `SoDatumLabel` is taking for granted that we are on XY plane.
For instance it calculates `normal = SbVec3f (-dir[1],dir[0],0);`.
So the points feeded to `SoDatumLabel `must not be 3d points, but actually 2d points.

I expected the `norm` variable to 'set' the plane to which to draw, but actually no.

So I'm not sure how sketcher is handling this 3D to 2D, but now that I'm thinking about it, it's doing it in a global way, all the coordinates used in sketcher are 2D.

wwmayer:
SoDatumLabel indeed only is on the XY plane. The `norm` field must be set but the rotation is achieved by an SoTransform node which e.g. is set to:
```
    Transform {
      translation 0 0 0
      rotation 1 0 0  1.5707964
      scaleFactor 1 1 1
      scaleOrientation 0 0 1  0
    }
```

Below you find an example of a sketch on the XZ plane with a single line and a constraint:
```
#Inventor V2.1 ascii

Separator {
    Transform {
      translation -0 -0 -0
      rotation 1 0 0  1.5707964
      scaleFactor 1 1 1
      scaleOrientation 0 0 1  0
    }
    Separator {
      renderCaching OFF
      Separator {
        Material {
          diffuseColor 1 1 1
        }
        Coordinate3 {
          point [ 23.793104 31.26437 0.0060000001,
              -23.793102 -15.287357 0.0060000001 ]
        }
        DrawStyle {
          lineWidth 3
          linePattern 0xffff
          linePatternScaleFactor 5
        }
        LineSet {
          numVertices 2
        }
      }
      Annotation {
        renderCaching OFF
        SoDatumLabel {
          fields [ MFString string, SFColor textColor, MFVec3f pnts, SFVec3f norm, SFName name, SFInt32 size, SFFloat lineWidth, SFEnum datumtype, SFFloat param1, SFFloat param2 ]
          string "66,5696 mm"
          textColor 1 0.14901961 0
          pnts [ 23.793104 31.26437 0.0089999996,
              -23.793102 -15.287357 0.0089999996 ]
          norm 0 -1 0
          size 13
          lineWidth 2
          datumtype DISTANCE
          param1 6.6666665
          param2 0
        }
      }
    }
}
```

PaddleStroke:
Thanks very much.
I can't find references to `SoTransform` in sketcher mod. Can you point me where this is done for the sketches?
I'm not sure to understand exactly how this `SoTransform` works. Does it affect the ImagePlane object ? Its placement ? Or does that affect only the visiual drawing of the inventor graphics ?

I mean in sketcher, SoDatumLabel is fed 2d points (so on XY plane) and yet draws to other planes thanks to this `SoTransform`. But the 2d points, are they 2d because the sketche placement is always XY plane and then the `SoTransform` changes the placement. Or the sketcher has it's placement (ZY) and convert the 3d points to 2d before feeding to SoDatumLabel and then SoDatumLabel graphics are `SoTransform`ed?

wwmayer:
> I can't find references to SoTransform in sketcher mod. Can you point me where this is done for the sketches?

This special node is direct part of the viewer class: https://github.com/FreeCAD/FreeCAD/blob/master/src/Gui/View3DInventorViewer.cpp#L437 and can be changed with the Document class: https://github.com/FreeCAD/FreeCAD/blob/master/src/Gui/Document.cpp#L456

> I'm not sure to understand exactly how this SoTransform works. Does it affect the ImagePlane object ? Its placement ? Or does that affect only the visiual drawing of the inventor graphics ?

This transform node should only affect the SoDatumLabel but not the image itself because it already has its own SoTransform node controlled by the placement.

> the sketcher has it's placement (ZY) and convert the 3d points to 2d before feeding to SoDatumLabel and then SoDatumLabel graphics are SoTransformed?

Yes, this is what happens. The magic for the sketcher happens inside getCoordsOnSketchPlane() that transforms world coordinates into local sketch coordinates

PaddleStroke:
```
void ViewProviderSketch::getCoordsOnSketchPlane(const SbVec3f &point, const SbVec3f &normal, double &u, double &v) const
{
    // Plane form
    Base::Vector3d R0(0, 0, 0), RN(0, 0, 1), RX(1, 0, 0), RY(0, 1, 0);

    // move to position of Sketch
    Base::Placement Plz = getEditingPlacement();
    R0 = Plz.getPosition();
    Base::Rotation tmp(Plz.getRotation());
    tmp.multVec(RN, RN);
    tmp.multVec(RX, RX);
    tmp.multVec(RY, RY);
    Plz.setRotation(tmp);
```

Hmm I have two remarks on `getCoordsOnSketchPlane`, but maybe I'm missing something.
- Why not using `tmp.multVec(RN)` ?
- Why `Plz.setRotation(tmp);` ? multVec does not seem to modify the rotation.

wwmayer:
> Please have a look at the functionality and let me know if you find any issues.

There are quite of few usability issues:
* The content of the spin box is truncated. Most of the time the unit is only partially visible
* The up/down arrows are missing
* If I hover over the spin box and scroll then the value is automatically accepted.
* If a second scaling is done then after the first click it shows the previous datum label
* Accepting a value is only possible by clicking inside the spin box and pressing Enter (or the odd behaviour to start scrolling) while rejecting is only possible by clicking the Cancel button. So, in order to be more consistent there should be buttons to accept or reject. Since the spin box is already inside the 3D view there could be added two push/tool buttons with an icon only (a green check mark for accepting and a red cross for rejecting). Additionally there should also be a way to reject the scaling by clicking the ESC button.

wwmayer:
> A - Pressing enter validate.

I don't think that it's for everyone immediately obvious that pressing Enter is the only way to accept a value.

> Then it would be intuitive enough to not have OK/Cancel buttons on view.

If you don't want them directly on the view then put Ok/Cancel to the context menu.

PaddleStroke:
I fixed the different issues. + I added ok cancel buttons. It looks like this : 
image

- Escape key does not escape before the spinbox appear. I tried to `this->installEventFilter(this);` but it doesn't work because InteractiveScale doesn't actually have the focus. Any idea how this should be done?
- Regarding to the spinbox size: It's now of a size big enough so that most of the time it won't hide text. I thought about resizing on text size when user type, but that would make the widget move around everytime user type in something. Which would probably look weird.
- Sometimes Enter doesn't register since I added the OK and cancel button. I can't pin-point a pattern. Do you see something wrong with my event filter?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [833791175](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/833791175) was triggered at [82efd84](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/82efd844671221f507e545d84479a86288449455). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> Escape key does not escape before the spinbox appear. I tried to this->installEventFilter(this); but it doesn't work because (I think) InteractiveScale doesn't have the focus. Any idea how this should be done?

You don't do this with the Qt API but the Inventor API. You add another callback function with addEventCallback() for a SoKeyboardEvent (or you can change the event type of getMouseClick() to SoButtonEvent and handle it there). Inside the callback function you check for a SoKeyboardEvent and check with getKey() for SoKeyboardEvent::ESCAPE. If the getState() is SoButtonEvent::UP you leave the mode.

> Regarding to the spinbox size: It's now of a size big enough so that most of the time it won't hide text.

Having a fixed size that is big enough is indeed better.

wwmayer:
> Note: I can't rebase on master because it doesn't compile on windows. I posted there I don't know if you've seen

https://github.com/FreeCAD/FreeCAD/commit/b543fe3bfe634c15eb3150612b29c2e2c1eab236

PaddleStroke:
I just pushed final changes. It looks ready to me. Please if you can have a look and if you have any further comments don't hesitate.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [834741826](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/834741826) was triggered at [89fe7ed](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/89fe7edb4f90c592fdb0745de3f924565f2c95b3). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [834762588](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/834762588) was triggered at [466aac1](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/466aac123b9cd1e76e77c611b36fd254f5086904). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
It was solely a UI decision it was working well. I didn't squash in case we want to go back to those buttons.

> The SoDatumLabel is misplaced if the image plane has an offset or if the image is child of a Part container.
> Setting a value by pressing Enter doesn't work any more. Instead the text of the spin box will be selected. Pressing ESC to cancel works flawlessly
> When performing a second scale operation then after the first click you still see SoDatumLabel of the previous operation

- Fixed the offset issue. I can't replicated the Part container issue, can you please elaborate when this happens?
- On my side pressing enter works flawlessly so I am not sure what to do here.
- I see you mean right after first click before moving the mouse, then the previous label is visible indeed. It's fixed.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [834997978](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/834997978) was triggered at [5e1a84f](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/5e1a84fde39d6bc2bb868219ac5baced6e6ca5e8). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
Transparency isn't intuitive.  If the image plane is the only thing in the document, changing the setting doesn't appear to change anything.  Set it to 100 and the plane is still visible.  Then if you add another object, the plane suddenly changes transparency and disappears.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [835143392](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/835143392) was triggered at [aed4113](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/aed41136411b6d0b98d5c5ab2d9dbdcddaac3ca4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
Let's talk more about selecting points outside the image boundary.  

In the vast majority of cases, the user is selecting two points inside the image.  For that reason, it's helpful to bound the selection points inside the image.  It makes it easier to position the points accurately.  

I can only imagine a single use-case where letting the user set points outside is valuable:   When the user has an image that is cropped so the target point they want to select isn't included but where the image still has enough information that the user can roughly guess where the endpoint should be.  That's a pretty rare case.   Are there others?

If there are valid cases, then 
1) We ALWAYS let the user select outside the image and this makes selection harder in the most common cases
2) We provide a control to turn on outside selection.  This takes up screen and the user has to understand what the control does.

If we're just including this feature because it's cool and _might_ be useful to someone in the future, I would suggest removing it until users ask for it.  The cognitive burden of understanding it is higher than the value of the feature at this point.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [835291238](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/835291238) was triggered at [e168be3](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/e168be33971772cc232d233c1f1a62ab11f80667). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
Oh actually there is also at line 246 on the initialization of scale.
It is now working

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [835328014](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/835328014) was triggered at [b10a7a6](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/b10a7a65db30cc13a820499257e368ceb7b45b49). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [835346956](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/835346956) was triggered at [7b30698](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/7b3069868d64564649f88315baeec10629b35578). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
Regarding the setting 'allow outside picture' I think that this settings should be gone. Either we:
- Always allow outside picture. Why restricting usability? If user click outside the picture that's because he wants it.
- Just remove the setting: user can select points only on imageplane. Selecting outside the image seems to have very little use to me.

Besides it is to be noted that the 'outside picture' mode is buggy: If the camera is not perpendicular to the imageplane. If the camera is tilted then the points are not correct. I'm not sure if I made this regression but I seem to recall that it was already the case before I started.

sliptonic:
> Besides it is to be noted that the 'outside picture' mode is buggy

That should settle it.  Let's remove it for now.   
Restrict the pointer to the image bounds.
If users want the additional feature, we can do that with a new issue and invested users to test.

sliptonic:
I don't know if it's related to this PR but I found a repeatable segfault.
open a document with an imageplane.  
Double-click to open the task panel.
Calibrate and select the first and second point.
Now close the document.
The task panel is still open.
Click ok or cancel at the task panel (not calibration), and it segfaults.

PaddleStroke:
This segfault is not linked to this PR, it can be repeated in current master : 
image

PaddleStroke:
But is that not already a bug the fact that closing the document does not dismiss the taskbox ?

wwmayer:
> But it doesn't crash when you click ok or cancel.

Inside the function that causes a crash all what you have to do beforehand is checking with `feature.expired()`

PaddleStroke:
Segfault fixed and outside picture mode removed.
By the way, we should do something about the fact that taskviews are not dismissed when a document is closed. This makes the app feel buggy.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199">pipeline status</a> for feature branch [PR_9199](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9199). Pipeline [839641329](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/839641329) was triggered at [b3e1d3d](https://github.com/FreeCAD/FreeCAD/pull/9199/commits/b3e1d3dc6a96bea07fed2445a824ac26f7dd69dc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
> we should do something about the fact that taskviews are not dismissed when a document is closed. This makes the app feel buggy.

https://github.com/FreeCAD/FreeCAD/issues/9297

chennes:
OK, this one looks good to me -- @PaddleStroke is there anything else you are doing to it before merging?

PaddleStroke:
No it's ok for me, nothing to add.

PaddleStroke:
Sorry I must have missed this last review I don't recall the issues you fixed here.

wwmayer:
> Sorry I must have missed this last review I don't recall the issues you fixed here.

Hmm, I added them as in-source comments but didn't explicitly add a @PaddleStroke So, maybe that's why you haven't got a notification.

PR #9209, opened 2023-04-06
[TechDraw] Add customizability for dimension line spacing

### LINKED ISSUE #8512: [Feature Request] TechDraw adjust dimension value position

### Forums discussion

https://forum.freecad.org/viewtopic.php?t=76122

### Subproject(s) affected?

Techdraw

### Idea description

The spacing between dimension value text and the dimension/extension lines is fixed.  In some cases it would be helpful to have an adjustment to increase this spacing. 

Note to programmer:  see QGIViewDimension::getDefaultIsoDimensionLineSpacing. default is 2*lineWidth.

### Anything else?

_No response_

### Code of Conduct

benj5378:

---

Fixes #8512

### DIFF
--- a/src/Mod/TechDraw/Gui/DlgPrefsTechDrawDimensions.ui
+++ b/src/Mod/TechDraw/Gui/DlgPrefsTechDrawDimensions.ui
@@ -559,6 +559,40 @@ Multiplier of 'Font Size'</string>
           </property>
          </widget>
         </item>
+        <item row="11" column="0">
+         <widget class="QLabel" name="label_18">
+          <property name="font">
+           <font>
+            <italic>true</italic>
+           </font>
+          </property>
+          <property name="toolTip">
+           <string>Controls the size of spacing between dimension line and dimension text for ISO dimensions.</string>
+          </property>
+          <property name="text">
+           <string>Line Spacing - ISO</string>
+          </property>
+         </widget>
+        </item>
+        <item row="11" column="2">
+         <widget class="Gui::PrefDoubleSpinBox" name="pdsbLineSpacingFactorISO">
+          <property name="toolTip">
+           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Controls the size of spacing between dimension line and dimension text. Value * linewidth is the line spacing.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+          </property>
+          <property name="alignment">
+           <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
+          </property>
+          <property name="value">
+           <double></double>
+          </property>
+          <property name="prefEntry" stdset="0">
+           <cstring>LineSpacingFactorISO</cstring>
+          </property>
+          <property name="prefPath" stdset="0">
+           <cstring>Mod/TechDraw/Dimensions</cstring>
+          </property>
+         </widget>
+        </item>
        </layout>
       </item>
      </layout>
--- a/src/Mod/TechDraw/Gui/QGIViewDimension.cpp
+++ b/src/Mod/TechDraw/Gui/QGIViewDimension.cpp
@@ -849,7 +849,7 @@ Base::Vector2d QGIViewDimension::getIsoRefOutsetPoint(const Base::BoundBox2d& la
 {
     return Base::Vector2d(right ? labelRectangle.MinX - getDefaultIsoReferenceLineOverhang()
                                 : labelRectangle.MaxX + getDefaultIsoReferenceLineOverhang(),
-                          labelRectangle.MinY - getDefaultIsoDimensionLineSpacing());
+                          labelRectangle.MinY - getIsoDimensionLineSpacing());
 }
 
 Base::Vector2d QGIViewDimension::getIsoRefJointPoint(const Base::BoundBox2d& labelRectangle,
@@ -1146,10 +1146,10 @@ bool QGIViewDimension::constructDimensionArc(
         double borderRadius = (labelRectangle.GetCenter() - arcCenter).Length();
 
         if (borderRadius > arcRadius) {
-            borderRadius = arcRadius + getDefaultIsoDimensionLineSpacing();
+            borderRadius = arcRadius + getIsoDimensionLineSpacing();
         }
         else if (borderRadius < arcRadius) {
-            borderRadius = arcRadius - getDefaultIsoDimensionLineSpacing();
+            borderRadius = arcRadius - getIsoDimensionLineSpacing();
         }
 
         // ISO oriented labels are symmetrical along their center axis
@@ -1576,7 +1576,7 @@ void QGIViewDimension::drawDistanceExecutive(const Base::Vector2d& startPoint,
             labelCenter
             + Base::Vector2d::FromPolar(
                 placementFactor
-                    * (labelRectangle.Height() * 0.5 + getDefaultIsoDimensionLineSpacing()),
+                    * (labelRectangle.Height() * 0.5 + getIsoDimensionLineSpacing()),
                 lineAngle + M_PI_2));
 
         // Compute the dimensional line start and end crossings with (virtual) extension lines
@@ -1786,7 +1786,7 @@ void QGIViewDimension::drawDistanceOverride(const Base::Vector2d& startPoint,
             labelCenter
             + Base::Vector2d::FromPolar(
                 placementFactor
-                    * (labelRectangle.Height() * 0.5 + getDefaultIsoDimensionLineSpacing()),
+                    * (labelRectangle.Height() * 0.5 + getIsoDimensionLineSpacing()),
                 lineAngle + M_PI_2));
 
         // Compute the dimensional line start and end crossings with (virtual) extension lines
@@ -1990,7 +1990,7 @@ void QGIViewDimension::drawRadiusExecutive(const Base::Vector2d& centerPoint,
         double lineAngle;
         double devAngle = computeLineAndLabelAngles(centerPoint, labelCenter,
                                                     labelRectangle.Height() * 0.5
-                                                        + getDefaultIsoDimensionLineSpacing(),
+                                                        + getIsoDimensionLineSpacing(),
                                                     lineAngle, labelAngle);
 
         // Is there point on the arc, where line from center intersects it perpendicularly?
@@ -2014,7 +2014,7 @@ void QGIViewDimension::drawRadiusExecutive(const Base::Vector2d& centerPoint,
 
             devAngle = computeLineAndLabelAngles(arcPoint, labelCenter,
                                                  labelRectangle.Height() * 0.5
-                                                     + getDefaultIsoDimensionLineSpacing(),
+                                                     + getIsoDimensionLineSpacing(),
                                                  lineAngle, labelAngle);
             lineAngle = DrawUtil::angleComposition(lineAngle, M_PI);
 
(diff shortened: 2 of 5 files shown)


src/Mod/TechDraw/Gui/DlgPrefsTechDrawDimensions.ui
```diff
@@ -559,6 +559,40 @@ Multiplier of 'Font Size'</string>
...
+          </property>
+          <property name="value">
+           <double></double>
+          </property>
+          <property name="prefEntry" stdset="0">
+           <cstring>LineSpacingISO</cstring>
+          </property>
+          <property name="prefPath" stdset="0">
+           <cstring>Mod/TechDraw/Dimensions</cstring>
+          </property>
```
benj5378, line 593: I have no idea what these lines define and if they are needed here (copied from GapASME in lines above)

src/Mod/TechDraw/Gui/ViewProviderDimension.cpp
```diff
@@ -86,6 +86,8 @@ ViewProviderDimension::ViewProviderDimension()
                       "Adjusts the gap between dimension point and extension line");
     ADD_PROPERTY_TYPE(GapFactorASME, (Preferences::GapASME()), group, App::Prop_None,
                       "Adjusts the gap between dimension point and extension line");
+    ADD_PROPERTY_TYPE(LineSpacingISO, (-1.0), group, App::Prop_None,
```
benj5378, line 89: Why do we need to define the property here too, if it was already defined in the ui file???

src/Mod/TechDraw/Gui/QGIViewDimension.cpp
```diff
@@ -2656,6 +2656,16 @@ double QGIViewDimension::getDefaultIsoDimensionLineSpacing() const
     return Rez::appX(m_lineWidth * 2.0);
 }
 
+double QGIViewDimension::getIsoDimensionLineSpacing() const
+{
+    auto vp = static_cast<ViewProviderDimension*>(getViewProvider(getViewObject()));
+    double customLineSpacing = vp->LineSpacingISO.getValue();
+    if (customLineSpacing == -1.0) {  // By default -1.0 in gui until user defined
```
benj5378, line 2663: Hardcoded by setting default gui value to -1.0. Any suggestions to how to better check if should use default or not? Using default needs to be an option if it needs to be relative to line width. Maybe a solution could be to set gui value to 2.0 and then returning `m_linewidth * customLineSpacing` in function?

benj5378:
> https://github.com/FreeCAD/FreeCAD/issues/8512#issuecomment-1472967170 : Harder than it sounds as dim text is positioned relative to origin, not the dimension line.
This is still an issue if it is an issue? Also, would preferably be fixed at a later point cause a lot of skill or refactoring is needed first.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9209">pipeline status</a> for feature branch [PR_9209](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9209). Pipeline [830574514](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/830574514) was triggered at [8c7b383](https://github.com/FreeCAD/FreeCAD/pull/9209/commits/8c7b38356c8a48ddab8bbcbc4f81cca8fde4f692). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/TechDraw/Gui/DlgPrefsTechDrawDimensions.ui
WandererFan: There are a bunch of widgets for updating preferences. This is how the "pref" widgets know where to write the parameter value.  Do you have the Designer plugin (https://wiki.freecad.org/Compile_on_Windows/en#Qt_Designer_plugin)?

src/Mod/TechDraw/Gui/ViewProviderDimension.cpp
WandererFan: The Properties in the ViewProvider are persistent,  they are saved with the document.  Properties in the UI apply to the widget.

src/Mod/TechDraw/Gui/QGIViewDimension.cpp
WandererFan: You made a new preference entry above, no?   The ViewProvider ctor would set the property for line spacing to the preference value using a getter from App/Preferences or Gui/PreferencesGui.   Same for initializing the UI - for create, take the value from preferences, for update, take the value from the ViewProvider.

src/Mod/TechDraw/Gui/DlgPrefsTechDrawDimensions.ui
benj5378: @WandererFan I have the designer plugin, but I avoided it as it wanted to reformat a lot of stuff in the UI file.

src/Mod/TechDraw/Gui/QGIViewDimension.cpp
benj5378: @WandererFan Not following along? If `-1.0`, it will use default instead.

WandererFan:
> > [#8512 (comment)](https://github.com/FreeCAD/FreeCAD/issues/8512#issuecomment-1472967170) : Harder than it sounds as dim text is positioned relative to origin, not the dimension line.
> 
> This is still an issue - if it is an issue in the first place? Also, would preferably be fixed at a later point cause a lot of skill or refactoring is needed first.

[](https://forum.freecad.org/viewtopic.php?t=77312)

src/Mod/TechDraw/Gui/DlgPrefsTechDrawDimensions.ui
WandererFan: The "pref" widgets create/update an entry in the FC parameter file (user.cfg). 

ADD_PROPERTY_TYPE macro adds a persistent property to an object.  The initial value of the property is set by looking up the value in user.cfg.  The look ups are centralized in Preferences/PreferencesGui.

src/Mod/TechDraw/Gui/QGIViewDimension.cpp
WandererFan: You don't decide about using the default here, you just use whatever value is in the ViewProvider's line spacing property.

The ViewProvider property is initialized to the default at creation.

src/Mod/TechDraw/Gui/DlgPrefsTechDrawDimensions.ui
benj5378: @WandererFan Right.

`          <property name="prefEntry" stdset="0">
           <cstring>LineSpacingISO</cstring>
          </property>
          <property name="prefPath" stdset="0">
           <cstring>Mod/TechDraw/Dimensions</cstring>
          </property>`
So this automatically sets the preferences. So I just have to add 2.0 to the ui file and we will always have a default preference in user.cfg where LineSpacingISO = 2.0?

benj5378: @WandererFan

src/Mod/TechDraw/Gui/QGIViewDimension.cpp
benj5378: No. If the value in viewprovider is -1, it uses the default ISO value. And that value can only be calculated from linewidth and not known beforehand

benj5378:
> 1. I'm changing Line Spacing ISO to 1000 and recomputing, but I don't see any change.  What am I missing?
va da fak, let me check, it worked previously.

> 2. is there a reason it is Line Spacing ISO?  is ASME different?
lol, why did I make an ISO function only.

As more customizable properties in ViewProvider is added, it should be made such that there is one dropdown choosing ASME or ISO for all the other properties, instead of having 2 propertyinputs for each property depending on whether it is ISO or ASME.

WandererFan:
https://freecad.github.io/DevelopersHandbook/technical/preferences.html
tell us what is missing in this section.  Better yet, submit a PR to update the handbook.
PR9209_saveRestore

WandererFan:
I don't know why I don't see a difference in these 3 dimensions yet.  Maybe I am not doing something right?  First glance at the code looks reasonable.

PR9209_LineSpacing

benj5378:
Yeah, well, it clearly works on the video? Can you send file plz @WandererFan

benj5378:
user1234:
> for me the moving itself works like in your video in the PR.

WandererFan:
.zip extension to satisfy github...

[PR9209_LineSpacing.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/11227178/PR9209_LineSpacing.FCStd.zip)

WandererFan:
I had ISO Referencing set as the default style.

- works for ISO Oriented
- doesn't work for ISO Referencing (but should work?)
- doesn't work for ASME Inlined (see below)
- doesn't work for ASME Referencing (see below)

My copy of ASME Y14.5-2009 Dimensioning and Tolerancing doesn't even show
examples of ASME Referencing style, so I think we can ignore the ASME styles until somebody complains.

benj5378:
Yeah, I just figured out. Will look at tomorrow, currently learning Vim so I can work faster and make you sweat even more 😘

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9209">pipeline status</a> for feature branch [PR_9209](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9209). Pipeline [837780607](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/837780607) was triggered at [7ae7408](https://github.com/FreeCAD/FreeCAD/pull/9209/commits/7ae7408434f62f0c88cd07c53ca0b0c38dc37324). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

benj5378:
@WandererFan So figured out why I never made an ASME line spacing function: ASME dimensions doesn't even have linespacing, likewise there are no logic using any ASME linespacing. ASME dimensions rather has a margin around text than linespacing.

> For the save/restore issue:

```
    ui->pdsbLineSpacingFactor->onSave();
    ui->pdsbLineSpacingFactor->onRestore()
```
Is there a reason that this isn't just done by default from the XML data?

```
double Rez::appX(double x)
{
   return x / getRezFactor();
}
```
What is `RezFactor`?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9209">pipeline status</a> for feature branch [PR_9209](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9209). Pipeline [837798356](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/837798356) was triggered at [97afb02](https://github.com/FreeCAD/FreeCAD/pull/9209/commits/97afb021e8339e397ddf7ae42eb002eb3556022e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

WandererFan:
"What is RezFactor"

We actually draw the graphics in 0.1 mm units.  If we don't use Rez, the graphics come out like 8 bit game graphics.

PR #9215, opened 2023-04-08
Gui: Add function in WindowParameter class to connect slot to ParameterManager signal

marioalexis84:
If a observer is connected directly to the signal of the ParameterManager, the observer is only notifed when a parameter value changes. So if is necessary to avoid recomputations, this strategy can by used instead the traditional `OnChange` function.

Sub-classes  of `WindowParameter` class can define a slot function with the appropiate signature:
`void MyClass::slot(ParameterGrp*, ParameterGrp::ParamType, const char*, const char*)`
and then connect using `setSlotParamChanged(&MyClass::slot, this)`

In `setGroupName` the group path is now checked using a regular expression.

 
---

### DIFF
--- a/src/Gui/Window.cpp
+++ b/src/Gui/Window.cpp
@@ -23,11 +23,13 @@
 
 #include "PreCompiled.h"
 
+#include <boost/algorithm/string.hpp>
 #include <qglobal.h>
-#include "Window.h"
 
 #include <App/Application.h>
-#include <Base/Console.h>
+
+#include "Window.h"
+
 
 using namespace Gui;
 
@@ -53,29 +55,40 @@ WindowParameter::WindowParameter(const char *name)
 
 WindowParameter::~WindowParameter()
 {
+    connParamChanged.disconnect();
 }
 
 /** Sets the group of the window to \a name */
 bool WindowParameter::setGroupName(const char* name)
 {
-  if (_handle.isValid())
-    return false; // cannot change parameter group
+    // cannot change parameter group
+    if (_handle.isValid()) {
+        return false;
+    }
 
-  assert(name);
-  std::string prefGroup = name;
-  if (prefGroup.compare(0,15,"User parameter:") == 0 ||
-      prefGroup.compare(0,17,"System parameter:") == 0)
-    _handle = App::GetApplication().GetParameterGroupByPath( name );
-  else
-    _handle = getDefaultParameter()->GetGroup( name );
-  return true;
+    assert(name);
+
+    std::string prefGroup = name;
+    const auto& list = App::GetApplication().GetParameterSetList();
+
+    auto found = std::find_if(list.begin(), list.end(), [prefGroup](auto item) {
+        return boost::starts_with(prefGroup, item.first);
+    });
+
+    if (found != list.end()) {
+        _handle = App::GetApplication().GetParameterGroupByPath(name);
+    }
+    else {
+        _handle = getDefaultParameter()->GetGroup(name);
+    }
+
+    return true;
 }
 
 void WindowParameter::OnChange(Base::Subject<const char*> &rCaller, const char * sReason)
 {
   Q_UNUSED(rCaller);
   Q_UNUSED(sReason);
-  Base::Console().Log("Parameter has changed and window (%s) has not overridden this function!",_handle->GetGroupName());
 }
 
 ParameterGrp::handle  WindowParameter::getWindowParameter()
--- a/src/Gui/Window.cpp
+++ b/src/Gui/Window.cpp
@@ -44,13 +44,13 @@ using namespace Gui;
 
 WindowParameter::WindowParameter(const char *name)
 {
-  // not allowed to use a Window without a name, see the constructor of a DockWindow or a other QT Widget
-  assert(name);
-  //printf("Instanceate:%s\n",name);
+    // not allowed to use a Window without a name, see the constructor
+    // of a DockWindow or a other QT Widget
+    assert(name);
 
-  // if string is empty do not create group
-  if ( strcmp(name, "") != 0 )
-    _handle = getDefaultParameter()->GetGroup( name );
+    // if string is empty do not create group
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9215">pipeline status</a> for feature branch [PR_9215](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9215). Pipeline [831617992](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/831617992) was triggered at [39f0852](https://github.com/FreeCAD/FreeCAD/pull/9215/commits/39f0852e84a9157b8497e0efaa57b4de8009b8db). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/Window.h
```diff
@@ -46,6 +46,19 @@ class GuiExport WindowParameter : public ParameterGrp::ObserverType
...
+  /// connect slot to ParameterManager signal
+  template<typename S, typename T>
+  void setSlotParamChanged(S slot, T* obsPtr)
+  {
+    namespace bp = boost::placeholders;
+    if (_handle->Manager()) {
+        connParamChanged = _handle->Manager()->signalParamChanged.connect(
+            boost::bind(slot, obsPtr, bp::_1 ,bp::_2 ,bp::_3 ,bp::_4));
+    }
+  }
```
chennes, line 58: Any reason to continue using `boost::bind`, now that we have `std::bind`? They are basically identical, I believe.

marioalexis84: No special reason. There is currently a mix between std::bind and boost::bind throughout FreeCAD code.
If you prefer to migrate from boost::bind to std::bind, I'll use the latter here.

chennes: I think at some point we should migrate, but it's not necessary to change this one unless you want to.

src/Gui/Window.cpp
```diff
@@ -53,29 +55,34 @@ WindowParameter::WindowParameter(const char *name)
...
+    return false;
+  }
 
   assert(name);
-  std::string prefGroup = name;
-  if (prefGroup.compare(0,15,"User parameter:") == 0 ||
-      prefGroup.compare(0,17,"System parameter:") == 0)
+
+  boost::regex ex("^(User|System)\\s{1}parameter:.*$");
+  if (boost::regex_match(name, ex)) {
```
chennes, line 72: It's probably not really important in this application, but isn't this much slower than the original?

src/Gui/Window.h
```diff
@@ -46,6 +46,19 @@ class GuiExport WindowParameter : public ParameterGrp::ObserverType
...
+
+  boost::signals2::connection connParamChanged;
```
chennes, line 60: Does this have to be public?

src/Gui/Window.cpp
marioalexis84: I propose to use a regex to avoid an error (and comparing strings can be error prone) due to the number of spaces between "User" ("System") and "parameter". With the regex it is made explicit that there is one and only one space.

src/Gui/Window.h
marioalexis84: Not at all. It should be private.
Do you agree with move `setSlotParamChanged` and `setGroupName` to protected?

chennes: Definitely on `setSlotParamChanged`. I don't feel strongly about changing `setGroupName`. Is there a compelling reason to put the function definition in the header instead of the cpp file?

marioalexis84: I put it there because it is a template. What alternative do you suggest?

chennes: lol, yep, that's a "compelling reason"! Sorry for the noise.

marioalexis84:
@chennes
I kept using `boost::bind`.
In another PR we can make the full switch to `std::bind`.
Moved `setGroupName` to protected along with the template.
The connection is declared as private.
The template is defined as inline outside the body of the class.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9215">pipeline status</a> for feature branch [PR_9215](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9215). Pipeline [842282423](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/842282423) was triggered at [d10bf73](https://github.com/FreeCAD/FreeCAD/pull/9215/commits/d10bf732d89f3dd6f6e68e88a8ae2435ebcac178). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/Window.cpp
wwmayer: This function only checks for the parameter managers ("User parameter" or "System parameter") provided by the core system. However, it's possible that (proprietary) extension modules register their own parameter managers.

In order to support  them the function should be implemented as:
```

/** Sets the group of the window to \a name */
bool WindowParameter::setGroupName(const char* name)
{
    // cannot change parameter group
    if (_handle.isValid()) {
        return false;
    }

    assert(name);
    std::string prefGroup = name;
    const auto & list = App::GetApplication().GetParameterSetList();

    auto found = std::find_if(list.begin(), list.end(), [prefGroup](auto item) {
        return boost::starts_with(prefGroup, item.first);
    });

    if (found != list.end()) {
        _handle = App::GetApplication().GetParameterGroupByPath(name);
    }
    else {
        _handle = getDefaultParameter()->GetGroup(name);
    }

    return true;
}
```

marioalexis84:
Change code accoriding @wwmayer  sugestion.
The second commit only set the indentation to four spaces in both files

chennes:
Sorry this fell off my radar...

PR #9218, opened 2023-04-08
Path WB: Correct bug causing incorrect feedrate to be used.

J-Dunn:
Recent changes to had G98/G99, ended up causing horizontal feedrate to be used these drilling operations.  This fix corrects that and does some fluke8 clean ups.

- [ x]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

---

### DIFF
--- a/src/Mod/Path/Path/Op/Drilling.py
+++ b/src/Mod/Path/Path/Op/Drilling.py
@@ -172,8 +172,12 @@ class ObjectDrilling(PathCircularHoleBase.ObjectOp):
             "Drill",
             QT_TRANSLATE_NOOP("App::Property", "How far the drilling depth is extended"),
         )
-        obj.addProperty("App::PropertyBool", "KeepToolDown", "Drill",
-            QT_TRANSLATE_NOOP("App::Property", "Apply G99 retraction: only retract to RetractHeight between holes in this operation"))
+        obj.addProperty(
+            "App::PropertyBool", "KeepToolDown", "Drill",
+            QT_TRANSLATE_NOOP(
+                "App::Property",
+                "Apply G99 retraction: only retract to RetractHeight between holes in this operation")
+        )
 
         for n in self.propertyEnumerations():
             setattr(obj, n[0], n[1])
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9218">pipeline status</a> for feature branch [PR_9218](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9218). Pipeline [831869726](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/831869726) was triggered at [f02460c](https://github.com/FreeCAD/FreeCAD/pull/9218/commits/f02460ca74db265b49443f279b49e2ef2ec6d88c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

sliptonic:
Merging.  Could a unit test be added to catch this regression in the future

J-Dunn:
Thanks for merging.  Is there a guide to writing test units. I have no idea about that. 

This was because I removed one line too many of the previous giggling.  G0 X Y is not needed since that is the first thing G81/83 does in the interpreter. But not doing an explicit G0 made FC regard it as non-vertical move and set horizontal feed rate.  That is probably wrong in itself but the quickest way to fix it is to keep the explicit  G0 X Y move.

sliptonic:
There isn't a separate guide though this might be a good addition to the Technical section of the [Developers Handbook](https://freecad.github.io/DevelopersHandbook/).

Python unit tests are pretty easy to write and there are a lot of them for Path already so it's easy to figure out how to add new ones.  The entire suite is under Mod/Path/PathTests 

Path tests can be run from Test workbench but I prefer to run from the command line like: 

`./bin/FreeCADCmd -t TestPathApp`

Don't worry so much about testing the UI.  The near-term goal is to test the lower level logic at a functional level.  Study the tests for the generators like TestPathDrillGenerator.py

Individual tests are small and self contained.  They have little or no branching/looping logic.  Writing these kinds of tests encourages the developer to write more functional and modular code because it's easier to write tests for them. Over time, the code becomes more testable which makes it easier to refactor.

J-Dunn:
Thanks, if I find time, I'll try to learn from those examples. 

In fact the test here should be " isVertical or ( cmd in [G81,G83] )"  but that would probably need some testing with a variety of post processors and their interpreters to check they are doing their job properly too. 

NIST states that the first thing G81/83 does is move to X,Y but maybe we'd need to make sure that is respected.

PR #9242, opened 2023-04-12
Draft: Improve color handling of SVG geometry importer

0penBrain:
Handle 'none' color => Default to black (+ print message)
 Handle #RRGGBBAA color => Return alpha (+ print, message)
 Handle 'rgba(float, float, float, float) => Return alpha (+ print message)
 Reorganize function to return ASAP + print warning if unknown color format

https://forum.freecad.org/viewtopic.php?p=675119#p675119

### DIFF
--- a/src/Mod/Draft/importSVG.py
+++ b/src/Mod/Draft/importSVG.py
@@ -244,36 +244,46 @@ def getcolor(color):
     (r, g, b, a)
         RGBA float tuple, where each value is between 0.0 and 1.0.
     """
+    if color == "none":
+        FreeCAD.Console.PrintMessage("Color defined as 'none', defaulting to black\n")
+        return (0.0, 0.0, 0.0, 0.0)
     if color[0] == "#":
-        # Color string '#12ab9f'
-        if len(color) == 7:
+        if len(color) == 7 or len(color) == 9: # Color string '#RRGGBB' or '#RRGGBBAA'
             r = float(int(color[1:3], 16) / 255.0)
             g = float(int(color[3:5], 16) / 255.0)
-            b = float(int(color[5:], 16) / 255.0)
-        # Color string '#1af'
-        elif len(color) == 4:
+            b = float(int(color[5:7], 16) / 255.0)
+            a = 1.0
+            if len(color) == 9:
+                a = float(int(color[7:9], 16) / 255.0)
+                FreeCAD.Console.PrintMessage(f"Non standard color format #RRGGBBAA : {color}\n")
+            return (r, g, b, 1-a)
+        if len(color) == 4: # Color string '#RGB'
             # Expand the hex digits
             r = float(int(color[1], 16) * 17 / 255.0)
             g = float(int(color[2], 16) * 17 / 255.0)
             b = float(int(color[3], 16) * 17 / 255.0)
-        return (r, g, b, 0.0)
-    # Color string 'rgb(0.12,0.23,0.3,0.0)'
-    elif color.lower().startswith('rgb('):
-        cvalues = color[3:].lstrip('(').rstrip(')').replace('%', '').split(',')
-        if '%' in color:
-            r, g, b = [int(float(cv)) / 100.0 for cv in cvalues]
-        else:
-            r, g, b = [int(float(cv)) / 255.0 for cv in cvalues]
-        return (r, g, b, 0.0)
-    # Color string 'MediumAquamarine'
-    else:
-        v = svgcolorslower.get(color.lower())
-        if v:
-            r, g, b = [float(vf) / 255.0 for vf in v]
             return (r, g, b, 0.0)
-        # for k, v in svgcolors.items():
-        #    if k.lower() == color.lower():
-        #        pass
+    if color.lower().startswith('rgb(') or color.lower().startswith('rgba('): # Color string 'rgb[a](0.12,0.23,0.3,0.0)'
+        cvalues = color.lstrip('rgba(').rstrip(')').replace('%', '').split(',')
+        if len(cvalues) == 3:
+            a = 1.0
+            if '%' in color:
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9242">pipeline status</a> for feature branch [PR_9242](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9242). Pipeline [835394091](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/835394091) was triggered at [2ab550a](https://github.com/FreeCAD/FreeCAD/pull/9242/commits/2ab550aa475ed9b5b5322db7ccaa8691c3dcfbe4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Draft/importSVG.py
```diff
@@ -244,36 +244,46 @@ def getcolor(color):
...
             # Expand the hex digits
             r = float(int(color[1], 16) * 17 / 255.0)
             g = float(int(color[2], 16) * 17 / 255.0)
             b = float(int(color[3], 16) * 17 / 255.0)
-        return (r, g, b, 0.0)
-    # Color string 'rgb(0.12,0.23,0.3,0.0)'
-    elif color.lower().startswith('rgb('):
+            return (r, g, b, 1.0)
+    if color.lower().startswith('rgb(') or color.lower().startswith('rgba('): # Color string 'rgb[a](0.12,0.23,0.3,0.0)'
         cvalues = color[3:].lstrip('(').rstrip(')').replace('%', '').split(',')
```
Roy-043, line 267: This does not work properly for 'rgba' strings.
Suggested improvement:
`cvalues = color.lstrip('rgba(').rstrip(')').replace('%', '').split(',')`

Roy-043: Sorry for the delay, I was unaware of the PR.

Just a question: I am right in concluding that `'#RGBA'` does not occur and that `'rgba(0.12,0.23,0.3,0.0)'` is a standard format?

src/Mod/Draft/importSVG.py
```diff
@@ -244,36 +244,46 @@ def getcolor(color):
...
         RGBA float tuple, where each value is between 0.0 and 1.0.
     """
+    if color == "none":
+        FreeCAD.Console.PrintMessage("Color defined to 'none', defaulting to black\n")
+        return (0.0, 0.0, 0.0, 1.0)
     if color[0] == "#":
-        # Color string '#12ab9f'
-        if len(color) == 7:
+        if len(color) == 7 or len(color) == 9: # Color string '#RRGGBB' or '#RRGGBBAA'
             r = float(int(color[1:3], 16) / 255.0)
```
Roy-043, line 252: Actually `float` is not necessary here.

Roy-043:
Sorry for the delay, I was unaware of the PR.

Just a question: I am right in concluding that `'#RGBA'` does not occur and that `'rgba(0.12,0.23,0.3,0.0)'` is a standard format?

src/Mod/Draft/importSVG.py
```diff
@@ -244,36 +244,46 @@ def getcolor(color):
     (r, g, b, a)
         RGBA float tuple, where each value is between 0.0 and 1.0.
     """
+    if color == "none":
+        FreeCAD.Console.PrintMessage("Color defined to 'none', defaulting to black\n")
+        return (0.0, 0.0, 0.0, 1.0)
```
Roy-043, line 249: The default Alpha value in FreeCAD is zero for some reason. So in 3 places `1.0` should become `0.0`.

Roy-043: Note that the color properties involved (ShapeColor, TextColor, and LineColor) actually ignore the Alpha value. So perhaps it is wiser to always return `0.0`?

yorikvanhavre: Indeed FreeCAD works with  "transparency" instead of "alpha" (inheritance from older times... several other apps and renderers work with transparency). transparency  = 1 - alpha

I would still store the transparency value here, just to unify... At some point we might use that for something

yorikvanhavre (approved):
Looks good to me, I agree with @Roy-043 's comments though

src/Mod/Draft/importSVG.py
0penBrain: Good catch

src/Mod/Draft/importSVG.py
0penBrain: I won't touch this now to keep uniformity

src/Mod/Draft/importSVG.py
0penBrain: > The default Alpha value in FreeCAD is zero for some reason. So in 3 places `1.0` should become `0.0`.

This isn't alpha so (as confirmed). ;) I changed 4 places to return 0.0, and 2 others to return 1-a.

src/Mod/Draft/importSVG.py
```diff
@@ -244,36 +244,46 @@ def getcolor(color):
     (r, g, b, a)
         RGBA float tuple, where each value is between 0.0 and 1.0.
     """
+    if color == "none":
+        FreeCAD.Console.PrintMessage("Color defined to 'none', defaulting to black\n")
```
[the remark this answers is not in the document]
chennes: I agree with @Roy-043

src/Mod/Draft/importSVG.py
Roy-043: > I would still store the transparency value here, just to unify... At some point we might use that for something

Is that really a good idea? It results in a color property value that the user cannot set via the GUI.

yorikvanhavre:
I think it's good to go then! Thanks everyone!

PR #9244, opened 2023-04-12
added i18n for navicube labels

ronnystandtke:
The navigation cube labels are user visible and therefore need i18n and L10n. This pull request adds the necessary i18n.

---

### DIFF
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -110,6 +110,7 @@ public:
 };
 
 class NaviCubeImplementation : public ParameterGrp::ObserverType {
+	Q_DECLARE_TR_FUNCTIONS(NaviCubeImplementation)
 public:
     explicit NaviCubeImplementation(Gui::View3DInventorViewer*);
     ~NaviCubeImplementation() override;
@@ -893,12 +894,19 @@ void NaviCubeImplementation::initNaviCube(QtGLWidget* gl) {
         labels.clear();
         ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath(
             "User parameter:BaseApp/Preferences/NaviCube");
-        labels.push_back(hGrp->GetASCII("TextFront", QT_TR_NOOP("FRONT")));
-        labels.push_back(hGrp->GetASCII("TextRear", QT_TR_NOOP("REAR")));
-        labels.push_back(hGrp->GetASCII("TextTop", QT_TR_NOOP("TOP")));
-        labels.push_back(hGrp->GetASCII("TextBottom", QT_TR_NOOP("BOTTOM")));
-        labels.push_back(hGrp->GetASCII("TextRight", QT_TR_NOOP("RIGHT")));
-        labels.push_back(hGrp->GetASCII("TextLeft", QT_TR_NOOP("LEFT")));
+        // see https://wiki.qt.io/Technical_FAQ#How_can_I_convert_a_QString_to_char.2A_and_vice_versa.3F
+        QByteArray frontByteArray = tr("FRONT").toLocal8Bit();
+        labels.push_back(hGrp->GetASCII("TextFront", frontByteArray.constData()));
+        QByteArray rearByteArray = tr("REAR").toLocal8Bit();
+        labels.push_back(hGrp->GetASCII("TextRear", rearByteArray.constData()));
+        QByteArray topByteArray = tr("TOP").toLocal8Bit();
+        labels.push_back(hGrp->GetASCII("TextTop", topByteArray.constData()));
+        QByteArray bottomByteArray = tr("BOTTOM").toLocal8Bit();
+        labels.push_back(hGrp->GetASCII("TextBottom", bottomByteArray.constData()));
+        QByteArray rightByteArray = tr("RIGHT").toLocal8Bit();
+        labels.push_back(hGrp->GetASCII("TextRight", rightByteArray.constData()));
+        QByteArray leftByteArray = tr("LEFT").toLocal8Bit();
+        labels.push_back(hGrp->GetASCII("TextLeft", leftByteArray.constData()));
     }
     // create the main faces
     m_Textures[TEX_FRONT] = createCubeFaceTex(gl, gap, labels[0].c_str(), SHAPE_SQUARE);
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -895,17 +895,17 @@ void NaviCubeImplementation::initNaviCube(QtGLWidget* gl) {
         ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath(
             "User parameter:BaseApp/Preferences/NaviCube");
         // see https://wiki.qt.io/Technical_FAQ#How_can_I_convert_a_QString_to_char.2A_and_vice_versa.3F
-        QByteArray frontByteArray = tr("FRONT").toLocal8Bit();
+        QByteArray frontByteArray = tr("FRONT").toUtf8();
         labels.push_back(hGrp->GetASCII("TextFront", frontByteArray.constData()));
-        QByteArray rearByteArray = tr("REAR").toLocal8Bit();
+        QByteArray rearByteArray = tr("REAR").toUtf8();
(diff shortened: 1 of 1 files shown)


chennes:
Thanks for the PR, good idea. This won't actually do the translation, though, it just adds the entry to the translation database. In this application you need the `tr()` function. Also, at some point we should probably consider making this class watch for language changes so that this updates without needing a restart.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9244">pipeline status</a> for feature branch [PR_9244](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9244). Pipeline [835508457](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/835508457) was triggered at [69dd7f2](https://github.com/FreeCAD/FreeCAD/pull/9244/commits/69dd7f294936b0402e8a5ab13ac61273e9fce868). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

ronnystandtke:
Replacing `QT_TR_NOOP()` with `tr()` triggers the compiler error `‘tr’ was not declared in this scope; did you mean ‘str’?`.
I am not yet very familiar with the FreeCAD codebase. If you point me to the right directions I will add the necessary changes for using the `tr()` function and watching for language changes.

chennes:
Sorry, I didn't notice Navicube wasn't a `QObject` -- that complicates things a little, because you need to both use the macro to mark the string for translation, and then use the static `QCoreApplication::translate()` function to do the actual translation. You can see Qt's documentation of it here: https://doc.qt.io/qt-6/i18n-source-translation.html#mark-translatable-data-text-strings

Don't worry about watching for the change signal right now, we can do that in a separate pass.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9244">pipeline status</a> for feature branch [PR_9244](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9244). Pipeline [835667408](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/835667408) was triggered at [9e46e4a](https://github.com/FreeCAD/FreeCAD/pull/9244/commits/9e46e4a7d1618a79e8eec264060a973db2943d8f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

ronnystandtke:
I used the `Q_DECLARE_TR_FUNCTIONS` macro to add translation support to a non-Qt class, see:
https://doc.qt.io/qt-5/i18n-source-translation.html#translating-non-qt-classes

Is this also OK? This variant looks easier than the one with `QCoreApplication::translate()`.

chennes:
No problem with using `Q_DECLARE_TR_FUNCTIONS` -- but in FreeCAD the standard is to use `toUtf8` (469 occurrences), rather than `toLocal8Bit` (8 occurrences). It's not clear to me that it ever matters (it would only affect Windows), but I'd be happier to use the far more common function here.

xtemp09:
`toLocal8Bit` is [buggy on Windows](https://bugreports.qt.io/browse/QTBUG-14887?jql=text%20~%20%22toLocal8Bit%22), at least in older versions of Qt.

I took a look at NaviCube implemented in Revit.

- FRONT is translated as into Russian as Спереди
- REAR as Сзади
- TOP as Сверху
- BOTTOM as Снизу
- RIGHT as Справа
- LEFT as Слева

_Front_ has five letters, while _Спереди_ has seven. _Top_ has three, while _Сверху_ has six. Will the labels condense to fit the face? Will they be trimmed?

@ronnystandtke, did you check it? Will kerning be the same? Does the font used contains other characters besides Latin?

ronnystandtke:
> No problem with using `Q_DECLARE_TR_FUNCTIONS` -- but in FreeCAD the standard is to use `toUtf8` (469 occurrences), rather than `toLocal8Bit` (8 occurrences). It's not clear to me that it ever matters (it would only affect Windows), but I'd be happier to use the far more common function here.

Hmm, interesting. I just followed the 'official' advice in the Qt Wiki technical FAQ, see:
https://wiki.qt.io/Technical_FAQ#How_can_I_convert_a_QString_to_char.2A_and_vice_versa.3F

> `toLocal8Bit` is [buggy on Windows](https://bugreports.qt.io/browse/QTBUG-14887?jql=text%20~%20%22toLocal8Bit%22), at least in older versions of Qt.

Is this still buggy? The issue was marked as 'resolved' 13 years ago...

> Will the labels condense to fit the face? Will they be trimmed?

The current navicube implementation just centers the labels and trims them if they are too large. You can test it yourself by increasing the font size of the navicube:
image


> @ronnystandtke, did you check it? Will kerning be the same? Does the font used contains other characters besides Latin?

Yes, I tested it with other languages. You can do so via `Tools -> Edit parameters...`:
image

Here is the result for Russian:
image


> You are proposing i18n of a non-Qt object. If it were a QToolButton or a QMessageBox I would not bother. Did you check the translation with RightToLeft layout?

Yes, I tested it with Hebrew and it looks god so far:
image

chennes:
Are you working on Windows? I'd be curious to know if using `toLocal8Bit()` gives the correct results for non-ASCII characters with a non-UTF-8 encoding. As far as I know, Coin supports UTF-8, I don't know what it will do with non-UTF-8 non-ASCII characters.

ronnystandtke:
> Are you working on Windows?

No, I am working on Linux.

chennes:
OK - as far as I can determine Coin expects UTF-8 encoding, and that's what we're sending it elsewhere. The Qt wiki's suggestion doesn't really apply here because it's not Qt we're worried about, it's Coin that's rendering those letters. My advice is to change to that encoding, and later tonight I will have access to a Windows machine that I can test it out on to confirm.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9244">pipeline status</a> for feature branch [PR_9244](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9244). Pipeline [836503747](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/836503747) was triggered at [37eb3ba](https://github.com/FreeCAD/FreeCAD/pull/9244/commits/37eb3ba2298cd9d0c09baa2691f481555260090b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

Jolbas:
> OK - as far as I can determine Coin expects UTF-8 encoding, and that's what we're sending it elsewhere. The Qt wiki's suggestion doesn't really apply here because it's not Qt we're worried about, it's Coin that's rendering those letters. My advice is to change to that encoding, and later tonight I will have access to a Windows machine that I can test it out on to confirm.

@chennes I think Qt is responsible for rendering the text using a Qpainter 
https://github.com/FreeCAD/FreeCAD/blob/3432d8796c57ab51a1070c14a2a12069a31b9304/src/Gui/NaviCube.cpp#L514

chennes:
OK, I double-checked on Windows. Here's the result with `toLocal8Bit`:
toLocal8Bit

And here's `toUtf8`:
toUtf8

So although @Jolbas is right, it is in fact Qt doing the font rendering, it still does seem to expect UTF-8 encoding. So, I'll merge this as it currently stands.

Jolbas:
Regarding translations of navigation cube labels. There is a qApp.translate() present at the following line? How does that work? Seem to be double work. Can I remove this call to qApp.translate()? 
https://github.com/FreeCAD/FreeCAD/blob/c93b5e27c90e5257e9cc67b5bb373a7ed41a7225/src/Gui/NaviCube.cpp#L537

chennes:
Yes, that's actively wrong (especially now) -- it would have attempted to translate user-provided strings from the parameters, if they were set, which is almost certainly not what a user would intend.

Jolbas:
Ok. I remove it in my upcoming PR

PR #9257, opened 2023-04-13
Sketcher: add a circle overlay for arcs

### LINKED ISSUE #9177: [Issue] Geometric constraints on Arcs are displayed away of supports

### Forums discussion

_No response_

### Version

0.21 (Development)

### Full version info

```shell
OS: Debian GNU/Linux 12 (bookworm) (XFCE/lightdm-xsession)
Word size of FreeCAD: 64-bit
Version: 0.21.0.32540 (Git)
Build type: Release
Branch: master
Hash: 07cf291516b8cf52b6c6c8d6dd2ed0ba84b2117c
Locale: French/France (fr_FR)
```

### Subproject(s) affected?

Sketcher

### Issue description

When inserting a geometric constraint (perpendicular, tangency or pointOnObject) with an arc as support, the icon is displayed at the supports intersection but it can be away the arc endpoints.

### Anything else?

Current view:
ConstraintsOnArc

I propose to add an helper below the arc with the constraints color:
ConstraintsOnArc_Helper

You can also see future issue with arc to something distance: https://github.com/FreeCAD/FreeCAD/pull/9166

### Code of Conduct

FlachyJoe:
Fix #9177

The circle overlay is enabled by default and removable by setting Mod/Sketcher/General/ArcCircleHelperVisible preference to False.

image
image

### DIFF
--- a/src/Mod/Sketcher/Gui/EditModeInformationOverlayCoinConverter.cpp
+++ b/src/Mod/Sketcher/Gui/EditModeInformationOverlayCoinConverter.cpp
@@ -58,21 +58,25 @@ EditModeInformationOverlayCoinConverter::EditModeInformationOverlayCoinConverter
 
 void EditModeInformationOverlayCoinConverter::convert(const Part::Geometry * geometry, int geoid) {
 
-    // at this point all calculations relate to BSplineCurves
-    assert(geometry->getTypeId() == Part::GeomBSplineCurve::getClassTypeId());
+    if (geometry->getTypeId() == Part::GeomBSplineCurve::getClassTypeId()){
+        // at this point all calculations relate to BSplineCurves
+        calculate<CalculationType::BSplineDegree>(geometry, geoid);
+        calculate<CalculationType::BSplineControlPolygon>(geometry, geoid);
+        calculate<CalculationType::BSplineCurvatureComb>(geometry, geoid);
+        calculate<CalculationType::BSplineKnotMultiplicity>(geometry, geoid);
+        calculate<CalculationType::BSplinePoleWeight>(geometry, geoid);
 
-    calculate<CalculationType::BSplineDegree>(geometry, geoid);
-    calculate<CalculationType::BSplineControlPolygon>(geometry, geoid);
-    calculate<CalculationType::BSplineCurvatureComb>(geometry, geoid);
-    calculate<CalculationType::BSplineKnotMultiplicity>(geometry, geoid);
-    calculate<CalculationType::BSplinePoleWeight>(geometry, geoid);
-
-    addUpdateNode(degree);
-    addUpdateNode(controlPolygon);
-    addUpdateNode(curvatureComb);
-    addUpdateNode(knotMultiplicity);
-    addUpdateNode(poleWeights);
-
+        addUpdateNode(degree);
+        addUpdateNode(controlPolygon);
+        addUpdateNode(curvatureComb);
+        addUpdateNode(knotMultiplicity);
+        addUpdateNode(poleWeights);
+    } else if (geometry->getTypeId() == Part::GeomArcOfCircle::getClassTypeId()){
+        // at this point all calculations relate to ArcOfCircle
+        calculate<CalculationType::ArcCircleHelper>(geometry, geoid);
+        addUpdateNode(circleHelper);
+    } else
+        assert(false /* Not implemented */);
 };
 
 void EditModeInformationOverlayCoinConverter::addToInfoGroup(SoSwitch * sw) {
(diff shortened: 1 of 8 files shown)


src/Mod/Sketcher/Gui/EditModeInformationOverlayCoinConverter.cpp
```diff
@@ -82,167 +86,191 @@ void EditModeInformationOverlayCoinConverter::addToInfoGroup(SoSwitch * sw) {
 
 template < EditModeInformationOverlayCoinConverter::CalculationType calculation >
 void EditModeInformationOverlayCoinConverter::calculate(const Part::Geometry * geometry, [[maybe_unused]] int geoid) {
-    const Part::GeomBSplineCurve *spline = static_cast<const Part::GeomBSplineCurve *>(geometry);
-
-    if constexpr (calculation == CalculationType::BSplineDegree ) {
-        clearCalculation(degree);
 
-        std::vector<Base::Vector3d> poles = spline->getPoles();
+    if (geometry->getTypeId() == Part::GeomBSplineCurve::getClassTypeId()){
```
benj5378, line 90: Would suggest turning this into a guard clause (flipping the condition and moving the else lines in here.)

FlachyJoe: The ArcOfConic case could be treated next (to show an elliptic helper) and the `else` statement will become `else if (geometry->getTypeId() == Part::GeomArcOfCircle::getClassTypeId())`

It could be better to add a default failing case with `assert(false /*not implemented*/)`.

luzpaz:
@FlachyJoe just a minor tweak: please amend the commit title to remove the space between `Sketcher` and the colon `:` so it reads `Sketcher: add a circle overlay for arcs`. TIA

0penBrain:
@FlachyJoe could you please change the screenshot in description to use default colors? I'll be curious to see how it looks when applied to a fully constrained arc too.

FlachyJoe:
@0penBrain Done !

luzpaz:
@FlachyJoe https://github.com/FreeCAD/FreeCAD/pull/9257#issuecomment-1510513769

adrianinsaval:
I don't like too much that this can only be disabled with an advanced parameter

0penBrain:
> @0penBrain Done !

Thanks

> I don't like too much that this can only be disabled with an advanced parameter

Well, I agree but what would you like.
On my side I'd like something similar to `Sketcher_BSplineComb`. That would eventually need some refactoring of the sketcher menus to fit somewhere.

Also what one would think that it displays as a dotted line (or dashed line, with very short dashs) ?

Finally would like to have @abdullahtahiriyo 's thought on this PR :smile:

adrianinsaval:
> On my side I'd like something similar to `Sketcher_BSplineComb`. That would eventually need some refactoring of the sketcher menus to fit somewhere.
> 
> Also what one would think that it displays as a dotted line (or dashed line, with very short dashs) ?

agreed

0penBrain:
@FlachyJoe do you feel like you can handle these improvements?

FlachyJoe:
> @FlachyJoe do you feel like you can handle these improvements?

I'll try ;-)

FlachyJoe:
> @FlachyJoe [#9257 (comment)](https://github.com/FreeCAD/FreeCAD/pull/9257#issuecomment-1510513769)

I'll re-title the commit with the next changes

FlachyJoe:
> Also what one would think that it displays as a dotted line (or dashed line, with very short dashs) ?

All Information Overlay Layer nodes share the same SoDrawStyle defined here: https://github.com/FreeCAD/FreeCAD/pull/9257/files#diff-7af3b478ff75c396d76c4d8354a757ea53673c2f3c2d63e682182f6b31a15da9L350-L355

I don't know if it's a good idea to create a specific style for this feature.

0penBrain:
> I don't know if it's a good idea to create a specific style for this feature.

In this case you can keep same style. ;)

abdullahtahiriyo:
I like the feature. 

I do not quite like that it cannot be hidden from the UI.

The feature is going to suffer from substantial backlash if merged as currently is.

I see two possible ways forward:
(a) merge as is but with the default off
(b) add code to enable and disable it from a toolbar

I do see however, that (b) involves refactoring the code, as information layer is now B-Spline specific in the toolbar (I did not see that one coming, my bad). Perhaps all visibility staff should have its own toolbar. This could also go along with separating the grid visibility and the snap controls out.

Because of this, I would propose to do minimal changes and merge (a) now.

Then, somebody would have to open a thread in the forum to discuss the toolbar issue.

Does this sound ok?

0penBrain:
> Does this sound ok?

I would prefer the toolbar rework is done before merging. I don't think we need a discussion for a first throw.

abdullahtahiriyo:
If the default is disabled, I could go along with merging without toolbar. 

If we must provide a toolbar, I do not think the BSpline toolbar is the right place for this. Yet, it is where the rest of the information layer icons are.

Any proposal?

0penBrain:
> Any proposal?

Probably we have to add another submenu to the Sketcher main menu, something like "Visibility" and place the new command there along with all ones dealing with B-Splines stuffs visibility. ;)

FlachyJoe:
Ok for default off. We maybe can temporarily add a checkbox in the Sketcher Setting Display page in order to let user enable this easily.

src/Mod/Sketcher/Gui/EditModeInformationOverlayCoinConverter.cpp
FlachyJoe: Ok, got it !

AjinkyaDahale:
Like B-splines, I would like it to be only visible for certain cases. B-spline combs tend to clutter the entire scene, so I feel it'd be better to set it such that only selected/pre-selected curves have their overlays shown. This could be a separate issue though.

src/Mod/Sketcher/Gui/EditModeInformationOverlayCoinConverter.cpp
```diff
@@ -58,21 +58,25 @@ EditModeInformationOverlayCoinConverter::EditModeInformationOverlayCoinConverter
...
+        addUpdateNode(controlPolygon);
+        addUpdateNode(curvatureComb);
+        addUpdateNode(knotMultiplicity);
+        addUpdateNode(poleWeights);
+    } else if (geometry->getTypeId() == Part::GeomArcOfCircle::getClassTypeId()){
+        // at this point all calculations relate to ArcOfCircle
+        calculate<CalculationType::ArcCircleHelper>(geometry, geoid);
+        addUpdateNode(circleHelper);
+    } else
+        assert(false /* Not implemented */);
```
0penBrain, line 79:
```suggestion
    } else {
        assert(false /* Not implemented */);
    }

```

src/Mod/Sketcher/Gui/EditModeInformationOverlayCoinConverter.cpp
```diff
@@ -82,167 +86,188 @@ void EditModeInformationOverlayCoinConverter::addToInfoGroup(SoSwitch * sw) {
...
-
-        controlPolygon.coordinates.clear();
-        controlPolygon.indices.clear();
-
-        size_t nvertices;
+        circleHelper.coordinates.reserve(ndiv);
+        for (int i = 0; i < ndiv; i++) {
+        	double param = i * std::atan(1) * 8 / ndiv;
+        	circleHelper.coordinates.emplace_back(circle.value(param));
+        }
```
0penBrain, line 106: Do we really need to use something tessellated ? Isn't Coin able to draw arc of circles directly ?

0penBrain (commented):
@FlachyJoe I did a quick review.
How are you feeling about implementing GUI command to toggle it on/off ?
Good, or do you prefer to get it merged as is ?

src/Mod/Sketcher/Gui/EditModeInformationOverlayCoinConverter.cpp
AjinkyaDahale: I tried this for B-splines once. Didn't have much success. I did see that being done in python and can dig out the forum. If we change this we should consider changing everything else as well in a separate PR.

abdullahtahiriyo: This one is "only" for information layer, it is not the regular one for geometry. 

Anyway, it is not an straight-forward change. For sure in a separate PR.

abdullahtahiriyo:
Are we including this PR in the current release or do we push it for the next development cycle?

0penBrain:
> Are we including this PR in the current release or do we push it for the next development cycle?

I think it's good to go. Safer enough to not generate regression. :)

abdullahtahiriyo:
Then, I would propose to merge it now as is.

I will treat another PR or PRs to complete the feature in the next days as a bug fix and within the release.

0penBrain:
> Then, I would propose to merge it now as is.

Agree :)

PR #9164, opened 2023-04-02
[Gui] Advise user OpenGL version is insufficient

Syres916:
This seems a rational approach to reduce unnecessary reports in the forum

Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [x]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---

Issue #8808, opened 2023-03-09
`QBackingStore::endPaint()` called with active painter (check w/ Qt 5.15.8 when released)


vernzimm:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecad.org/viewtopic.php?p=525171

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.0.32198 (Git)
Build type: Release
Branch: master
Hash: f51b2156f35399cab38eef1e957a59ad5a11de66
Python 3.8.16, Qt 5.15.6, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * Behave-Dark-Colors 0.1.1
  * FreeCAD_Info
  * ProDarkThemePreferencePack 1.0.0
[/code]
```

### Subproject(s) affected?

Sketcher

### Issue description

Not sure exactly what happened, but program began to bog down and report was spammed with
"QBackingStore::endPaint() called with active painter; did you forget to destroy it or call QPainter::end() on it?"

[report out.txt](https://github.com/FreeCAD/FreeCAD/files/10937087/report.out.txt)

Note that forum link is to a similar sounding issue, but not related to Sketcher, and I couldn't figure out what the common feature was or how to reproduce it.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

xtemp09:
It's a Qt bug, I don't know what causes it. Quick search showed that the message `QBackingStore::endPaint() called with active painter` introduced with [this patch](https://github.com/qt/qtbase/commit/2aa9908e24611fa6d321e694b8415ba7e8d364b0). `qt/qtbase 5.9.5 on Jan 12, 2018`

However, [it caused flooding](https://bugreports.qt.io/browse/QTBUG-66452) the console output with the aforementioned warning. `5.9.6, 5.10.1`
 
It was supposed to be fixed in [this](https://github.com/qt/qtdeclarative/commit/047f161b1a706e81034ccd8d4700a3126c16dfe4) patch. `5.11.0-beta1 on Feb 19, 2018`



[...]

Issue #7303, opened 2022-08-03
[Feature] FreeCAD detects graphics driver and outputs opengl32.dll workaround for user 


theleetbeagle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=4&t=36087&start=20

### Version

0.20 (Release)

### Full version info

```shell
[code]
OS: Windows 7 Version 6.1 (Build 7601: SP 1)
Word size of FreeCAD: 64-bit
Version: 0.20.29177 (Git)
Build type: Release
Branch: releases/FreeCAD-0-20
Hash: 68e337670e227889217652ddac593c93b5e8dc94
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.2
Locale: English/United States (en_US)
[/code]
```

### Issue description

Well known issue that older computers, often with integrated graphics, will crash shortly after load or when creating new or opening file. 

Common suggestion to select Use OpenGL in preferences does not work. Users are often on the lastest installed drivers, which are several years old and discontinued. Submitter, for instance, is using a Dell E6510 laptop with latest drivers being from 2011. 

Using Build-0-Matic's suggestion on page 3 of this thread yields immediate results. 

https://forum.freecadweb.org/viewtopic.php?f=4&t=36087&start=20

Also, simply renaming opengl32sw.dll to opengl32.dll in the \bin\ directory fixes the issue.

Suggest altering FreeCAD installer to correctly rename opengl32.dll.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Syres916:
This is not a bug, if the file was renamed as proposed that would mean all Windows users would have to use Software Accelerated graphics rather than the (sometimes very expensive) hardware acceleration.The renaming of the file in it's present state is for those relatively few machines that their hardware/drivers aren't OpenGL compatible.

theleetbeagle:
> This is not a bug, if the file was renamed as proposed that would mean all Windows users would have to use Software Accelerated graphics rather than the (sometimes very expensive) hardware acceleration.The renaming of the file in it's present state is for those relatively few machines that their hardware/drivers aren't OpenGL compatible.

When a user is having an issue where the app is crashing, checking the OpenGL box under Preferences doesn't fix the issue. The user should not have to dive into directories to manually rename files, even if the issue were documented, which it isn't.

Seems like a bug to me.

[...]

PR #9272, opened 2023-04-15
App: Warn on OpenGL too old

chennes:
Based on original code from PR #9164, optimized for easy translation.


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272">pipeline status</a> for feature branch [PR_9272](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272). Pipeline [838129644](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838129644) was triggered at [112d3dd](https://github.com/FreeCAD/FreeCAD/pull/9272/commits/112d3ddaccd00c0d3f67d1476799e972cd137c01). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272">pipeline status</a> for feature branch [PR_9272](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272). Pipeline [838134718](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838134718) was triggered at [6e0e18c](https://github.com/FreeCAD/FreeCAD/pull/9272/commits/6e0e18cdf0ea72aa50a828ed38ca123b469b6edc). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272">pipeline status</a> for feature branch [PR_9272](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272). Pipeline [838136285](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838136285) was triggered at [1ff7e3b](https://github.com/FreeCAD/FreeCAD/pull/9272/commits/1ff7e3bf1a65c19d3cc280618b922e71a52070e4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272">pipeline status</a> for feature branch [PR_9272](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272). Pipeline [838139362](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838139362) was triggered at [024084f](https://github.com/FreeCAD/FreeCAD/pull/9272/commits/024084fa44611812509c102737e028cadf94fe95). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

xtemp09:
I would have wrote full sentences, without `+ QString::fromUtf8(". ")` and without `\n\n` in the end to facilitate translation and, actually, reading the code.

Also, I would have added `#ifdef NDEBUG` before the QMessageBox. I think developers and maintainers compile on virtual machine, they already know about OpenGL, just another QMessageBox can be annoying.

chennes:
I don't see a good way around having at least `+ QString::fromUtf8(" ")` if I break it up into individual sentences, since I don't want to end those sentences with a single space. And I did drop the "\n\n" at the end, so you may need to refresh.

xtemp09:
Of course, I should have updated the page. [In Hindi](https://blogs.transparent.com/hindi/punctuation-in-hindi/) symbol "|" is used to end sentence, so the full stop should be inside of `QObject::tr()`.

chennes:
Thanks -- I've made the change to the full stops. 

> Also, I would have added `#ifdef NDEBUG` before the QMessageBox. I think developers and maintainers compile on virtual machine, they already know about OpenGL, just another QMessageBox can be annoying.

This is much more than an "annoying message box" -- line 2197 is basically shutting FreeCAD down if you don't have OpenGL 2.0 or higher. I'm not confident that's right, though: just how broken are things with such old versions of OpenGL. Should we really refuse to run at all? I'm inclined to remove that line and take your suggestion of adding the `#ifdef NDEBUG`.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272">pipeline status</a> for feature branch [PR_9272](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272). Pipeline [838443262](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838443262) was triggered at [874fd1b](https://github.com/FreeCAD/FreeCAD/pull/9272/commits/874fd1bebc579cf57445e7e4528b77a7f15883f2). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Gui/Application.cpp
```diff
@@ -2180,6 +2180,22 @@ void Application::runApplication()
...
+                    + QString::fromUtf8(" ")
+                    + version
+                    + QString::fromUtf8(" ")
+                    + QObject::tr("Invalid OpenGL Version");
+                auto message =
+                    QObject::tr("This system is running OpenGL %1.%2.").arg(major).arg(minor)
+                    + QString::fromUtf8(" ")
+                    + QObject::tr("FreeCAD requires OpenGL 2.0 or above.")
+                    + QString::fromUtf8(" ")
+                    + QObject::tr("Please upgrade your graphics driver and/or card as required.");
```
adrianinsaval, line 2195: This is not a big deal at all, but is there any real reason why these strings need to be separated in chunks? When translating I feel like it would be better if they are all together.

chennes: I agree, but the line length gets annoyingly long. Do you have a good technique for mitigating that?

luzpaz: The problem (from my perspective looking at translations on crowdin) with having large string chunks is that they include newlines. Translators sometimes get 'creative' with newlines or just ignore them altogether. So the strategy is to limit newlines in translations as best as possible. But obviously inevitable.

adrianinsaval: point taken, out of curiosity what do we do if we do need long text translated, would syntax like this work:
```c++
                QObject::tr("This system is running OpenGL %1.%2. "
                    "FreeCAD requires OpenGL 2.0 or above. "
                    "Please upgrade your graphics driver and/or card as required."
                    ).arg(major).arg(minor);
```

adrianinsaval:
If you have software opengl enabled, is this check relevant?

xtemp09:
QMessageBox can be annoying, perhaps, showing a message in Report view is sufficient.


> line 2197 is basically shutting FreeCAD down if you don't have OpenGL 2.0 or higher.

I think this is erroneous. If OpenGL version is insufficient, FreeCAD continues to work, though, some features may be displayed imperfectly, like NaviCube.

When OpenGL is buggy or insufficient it resorts to software rendering. All OSes support OpenGL 1.1. Linux supports OpenGL 3.3 via Mesa. Therefore, I see no reason to show QMessageBox or to shut down if FreeCAD works fine.

I don't know what exactly causes the issue #8808, but I suppose it results from switch from hardware rendering to software. In this case FreeCAD continues to work, showing warnings in Report view.

src/Gui/Application.cpp
chennes: Yes - it works fine actually. In Python it doesn't, but that's because the Python lupdate script doesn't actually run a Python interpreter, so doesn't correctly concatenate the strings. In C++ it works fine, so I have updated this PR.

chennes: > The problem (from my perspective looking at translations on crowdin) with having large string chunks is that they include newlines. Translators sometimes get 'creative' with newlines or just ignore them altogether. So the strategy is to limit newlines in translations as best as possible.

I've put the newline in this paragraph as a separate (non-translated) string, following our practice elsewhere. So I think the translators will be happy with what they see now.

chennes:
> Therefore, I see no reason to show QMessageBox or to shut down if FreeCAD works fine.

I agree, that seems like a bad course of action. I've removed the offending line of code, and made the whole thing a simple console warning that displays in release mode only.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272">pipeline status</a> for feature branch [PR_9272](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272). Pipeline [838906598](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838906598) was triggered at [562473b](https://github.com/FreeCAD/FreeCAD/pull/9272/commits/562473b14ece339051152f282111795081bd7044). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
A console warning is useless! The original reason for this PR was because we've had many users that can't go past the splash screen because they don't have proper OpenGL support but there is nothing to tell them this is the problem.

chennes:
So then we seem to have two choices: annoy developers with a "useless" error message in a message box, or fail to inform users to the reason their FreeCAD doesn't load might be because they have too old an OpenGL implementation. Can we split the difference and make this a failing error (with dialog), but only when compiled in Release?

xtemp09:
> A console warning is useless! The original reason for this PR was because we've had many users that can't go past the splash screen because they don't have proper OpenGL support but there is nothing to tell them this is the problem.

adrianinsaval, could you give some links to the issues or the forum topics describing such a problem? I couldn't find anything. If OpenGL is insufficient FreeCAD will work, you may take a look at the videos in #9287. OpenGL is 1.1, but FreeCAD works anyway.

If a user can't get past the splash screen, it's a problem of graphics card driver, not insufficient OpenGL.

src/Gui/Application.cpp
```diff
@@ -2180,6 +2180,21 @@ void Application::runApplication()
 
             int major = context.format().majorVersion();
             int minor = context.format().minorVersion();
+
+#ifdef NDEBUG
+            // In release mode, issue a warning to users that their version of OpenGL is
+            // very old.
+            if (major < 2) {
+                auto message =
+                    QObject::tr("This system is running OpenGL %1.%2. "
```
0penBrain, line 2189: Is it really what is wanted to use 'QObject::tr', which means text will have no context for translation?

chennes: That's what we're using elsewhere in this file. The class this is in isn't itself derived from `QObject`.

0penBrain: Yes I saw that. So generally (when using Qt) either 'QCoreApplication::translate' is used, or class is equipped with the tr function by adding to it the 'Q_DECLARE_TR_FUNCTIONS' macro. So translations will have a context.
This said if at other places in the file there are other translations using direct call to 'QObject::tr' better use it too. But maybe later we have to think about contextualizing all translations.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272">pipeline status</a> for feature branch [PR_9272](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272). Pipeline [842282442](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/842282442) was triggered at [05fdfe3](https://github.com/FreeCAD/FreeCAD/pull/9272/commits/05fdfe349f4965694458a98ae210ae9454b03b88). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
This PR as it currently stands issues a warning dialog for OpenGL < 2.0 when running in release mode, but allows the app to continue to load. I think one possible further refinement would be to make the warning message one-shot: once seen and dismissed once, it would never be shown again. What are your thoughts on this implementation, @xtemp09 and @adrianinsaval ?

xtemp09:
Now I'm thinking, the installer should check the OpenGL version.

adrianinsaval:
A "Don't show again" checkbox or button could be a good idea

adrianinsaval:
see #7303 is it worth mentioning available workarounds? We could also offer the option of enabling software opengl in these situations

0penBrain:
> This PR as it currently stands issues a warning dialog for OpenGL < 2.0 when running in release mode, but allows the app to continue to load. I think one possible further refinement would be to make the warning message one-shot: once seen and dismissed once, it would never be shown again. 

We have `DlgCheckableMessageBox` class for that. Are you willing to do the improvement before merge ?

chennes:
Yes, I will switch this to Draft until I have time to do that (as you say, it's a simple change since it's built in to Qt).

0penBrain:
> Yes, I will switch this to Draft until I have time to do that (as you say, it's a simple change since it's built in to Qt).

I'm not sure it's really "built in to Qt". I'm a bit unclear about this class as it apparently originated in QtCreator but was widely customized for FC (especially to use our pref system).

chennes:
Oh, I was thinking I'd use [QErrorMessage](https://doc.qt.io/qt-6/qerrormessage.html).

0penBrain:
> Oh, I was thinking I'd use [QErrorMessage](https://doc.qt.io/qt-6/qerrormessage.html).

https://github.com/FreeCAD/FreeCAD/pull/9272#issuecomment-1523888406 :smile:

chennes:
Ah, there is a problem with the strategy of using those dialogs: if the error we are concerned about is preventing FreeCAD from fully launching, then DlgCheckableMessageBox will never get shown either, because it doesn't display until FreeCAD finishes launching. What is the exact error case we are trying to handle here with this message?

0penBrain:
> if the error we are concerned about is preventing FreeCAD from fully launching, then DlgCheckableMessageBox will never get shown either, because it doesn't display until FreeCAD finishes launching

Hmmm. Did you investigate where this limitation comes from ?

chennes:
> Hmmm. Did you investigate where this limitation comes from ?

Yes, it was not a real effect, it was just my own startup process tricking me 😄 -- all is well now.

0penBrain:
OK works well except that because the message box used main window as parent but is displayed before the latter is created, it (the message box) is displayed on top left of screen and then splash screen is shifted there too.
I guess it's a minor inconvenience though and so no simple way to counter this.

chennes:
Strange, I don't see that behavior on Windows. I'll see if there is something simple to do about it. Which OS are you on?

adrianinsaval:
anything holding this back?

chennes:
Just time to make sure it's OK -- if someone else wants to merge that's fine, but it's low on my priority list right now.

adrianinsaval:
It's not high priority but I think we should try to merge it before release, it can save us time with help requests in the future and a lot of people will be using the stable release

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272">pipeline status</a> for feature branch [PR_9272](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9272). Pipeline [903673407](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903673407) was triggered at [e13b846](https://github.com/FreeCAD/FreeCAD/pull/9272/commits/e13b846bb359ef6e553aa380bd4f2f225086174f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
Perhaps we should merge this for the sake of the sanity of the community needing to troubleshoot this issue constantly?

PR #9285, opened 2023-04-15
Fix downcasting of SoNode to SoFCSelectionRoot

dodomorandi:

---

A SoNode* is not necessarily a SoFCSelectionRoot*, and when this assumption breaks the code causes UB (the comment related to one of the chunks explicitly says that it is safe, but unfortunately it is not).

Using `dynamic_cast` and checking for nullptr solves the issue.


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9285">pipeline status</a> for feature branch [PR_9285](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9285). Pipeline [838480786](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/838480786) was triggered at [030d3a2](https://github.com/FreeCAD/FreeCAD/pull/9285/commits/030d3a233903a05073d86efe38d42a4d5a66f867). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
Unfortunately, this breaks selection highlighting and pre-highlighting. In many cases that pointer really isn't a `SoFCSelectionRoot` but it still needs to be looked up as though it were. Not my design 🤷 .

dodomorandi:
Ok, I did not realize the breaking behavior. I am gonna try to use a different approach in order to maintain the old behavior without triggering UB.

dodomorandi:
Hi @chennes! I tried a different route.
The main idea is that the `stack` could contain things that are _just_ `SoNode*`s instead of _full_ `SoFCSelectionRoot*`. Therefore I changed the type of `stack` and I try to cast the node that is expected to be as `SoFCSelectionRoot*`, if this is not the case I behave like it did not exist. In this way it much less strict than my previous attempt, and obviously it does not trigger UB.

One relatively minor nitpick: I know _a priori_ that a pointer inside `stack` is _exactly_ a `SoFCSelectionRoot*` or just a generic `SoNode*` I am not really interested into. Therefore, just theoretically, instead of using a `SoNode*` and then doing dynamic casts, it could be possible to use a `std::variant<SoNode*, SoFSelectionRoot*>` and totally avoid RTTI. However doing so will surely decrease the readability of the code (`std::variant` is not exactly ergonomic to use), and at the same time we will double the amount of memory needed by `stack` because of a single bit of information, which would be a bit a shame.
I think that in this case maybe just using dynamic casts is good enough, but let me know what you think.

Thanks in advance!

src/Gui/SoFCSelectionContext.cpp
```diff
@@ -178,11 +178,18 @@ bool SoFCSelectionContextEx::isSingleColor(uint32_t &color, bool &hasTransparenc
...
+        SoFCSelectionContextBasePtr input, SoNode *node)
 {
     auto ctx = std::dynamic_pointer_cast<SoFCSelectionContextEx>(input);
+    SoFCSelectionRoot* selectionNode;
+    if (node == nullptr) {
+        selectionNode = nullptr;
+    } else {
+        selectionNode = dynamic_cast<SoFCSelectionRoot*>(node);
+    }
+
```
luzpaz, line 190: please fix trailing whitespace :wink:

PR #9302, opened 2023-04-18
i18n for CommandBody

ronnystandtke:
The labels of new bodies are not localized. This PR fixes this issue.

---

### DIFF
--- a/src/Mod/PartDesign/Gui/CommandBody.cpp
+++ b/src/Mod/PartDesign/Gui/CommandBody.cpp
@@ -198,7 +198,7 @@ void CmdPartDesignBody::activated(int iMsg)
     doCommand(Doc,"App.ActiveDocument.getObject('%s').Label = '%s'", bodyString, labelByteArray.constData());
     if (baseFeature) {
         if (partOfBaseFeature){
-            //withdraw base feature from Part, otherwise visibility mandess results
+            //withdraw base feature from Part, otherwise visibility madness results
             doCommand(Doc,"App.activeDocument().%s.removeObject(App.activeDocument().%s)",
                     partOfBaseFeature->getNameInDocument(), baseFeature->getNameInDocument());
         }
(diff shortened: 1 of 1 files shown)


src/Mod/PartDesign/Gui/CommandBody.cpp
```diff
@@ -189,41 +189,45 @@ void CmdPartDesignBody::activated(int iMsg)
     openCommand(QT_TRANSLATE_NOOP("Command", "Add a Body"));
 
     std::string bodyName = getUniqueObjectName("Body");
+    const char* bodyString = bodyName.c_str();
 
     // add the Body feature itself, and make it active
-    doCommand(Doc,"App.activeDocument().addObject('PartDesign::Body','%s')", bodyName.c_str());
+    doCommand(Doc,"App.activeDocument().addObject('PartDesign::Body','%s')", bodyString);
+    // set Label for i18n/L10N
+    QByteArray labelByteArray = QObject::tr("Body").toUtf8();
```
chennes, line 197: You should be able to use just `tr()` here, right? Rather than stripping off the context with `QObject`?

ronnystandtke: Unfortunately not. If I do so I get the following error message from the compiler:
`error: ‘tr’ was not declared in this scope; did you mean ‘tm’?`

chennes: OK, I thought this was a QObject, but it must not be.

ronnystandtke: Hmm, I'm pretty new to C++ and the whole FreeCAD code base, but wouldn't this exactly be the use case for the Q_DECLARE_TR_FUNCTIONS macro? We could declare the context in this macro and use `tr()` instead of `QObject::tr()`.

ronnystandtke: OK, I will try that here. How is the workflow now? Just adding another commit to this PR or creating a new PR?

chennes: Now that this is merged, you'll have to make a new PR for anything further you want to work on.

ronnystandtke: OK, will do so.

PR #9314, opened 2023-04-19
Add App to CrowdIn

chennes:
There are a few translatable strings in App: those are not currently being sent to CrowdIn. This PR updates the script that is used to communicate with CrowdIn to add them.

---


chennes:
@yorikvanhavre please give this a look.

yorikvanhavre:
Ok that seems good! I did manually a first "commit" of the file to crowdin already, so we should be good to go!
There might be an error later on when we download the strings and the script tries to update a non-existing qrc file, but better to see what happens first before trying to solve blindly here

PR #9325, opened 2023-04-19
Use smart pointer to handle GCS::Constraint in GCS::System.

andre-caldas:
Discussion:
https://forum.freecad.org/viewtopic.php?t=76724#p676575

According to @abdullahtahiriyo 's request, this is to be merged ONLY AFTER people profile it and discard significant impact on the solver.

I have commented on a piece of code where a "cache" is created so elements of a `std::vector` are not erased inside a (forward) iterator loop. The entries that are supposed to be removed are recorded in an auxiliary vector and deleted in a second loop. I think there is no need to create such a cache. I guess it is perfectly fine to erase an element if we use a backward iterator. I can do it if anyone wants me to. Please, let me know.

Also, the method `removeConstraint()` silently ignores when it is asked to remove a constraint it does not manage. I think this is a very strange behaviour. I made a comment on the code about this. I suppose it is just shadowing some bug that exists somewhere. Upon request, I can (do any of the below):
1. Just remove the comment.
2. Throw an exception, instead of ignoring the error.
3. Do not change the behaviour, but emit a warning message.

---

- [??]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

I do get an error when running the tests:
> ERROR: test_html_validates (TestStart.TestStartPage.TestStartPage.test_html_validates)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325">pipeline status</a> for feature branch [PR_9325](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325). Pipeline [842752902](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/842752902) was triggered at [36acd67](https://github.com/FreeCAD/FreeCAD/pull/9325/commits/36acd67f9e6ff3e431f4ae8c382b5c93f4be0edb). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
In order to do a fair job of profiling this code, I suggest writing some Google Test C++ unit tests: using the Python ones will distort the results because test setup and teardown is so slow. We can configure a gtest function to run a few million times and measure the results.

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -619,7 +618,7 @@ int System::addConstraint(Constraint *constr)
     if (constr->getTag() >= 0) // negatively tagged constraints have no impact
         hasDiagnosis = false;  // on the diagnosis
 
-    clist.push_back(constr);
+    clist.emplace_back(constr);
```
chennes, line 621: So this is taking a pointer to a constraint constructed elsewhere, and turning it into a unique_ptr, which will then deallocate it? I don't think it's reasonable to go from a non-owning pointer and simply declare that you own it now: where is this function getting called?

andre-caldas: I couldn't agree more. :-)

But the important fact is that the `GCS::System` is actually in charge of destroying this `Constraint` someone else has allocated. :-(
Destructor:
https://github.com/FreeCAD/FreeCAD/blob/1dcc58c4b1c51c50215639ca58326f6b69867743/src/Mod/Sketcher/App/planegcs/GCS.cpp#L568
Called by destructor (Clear):
https://github.com/FreeCAD/FreeCAD/blob/1dcc58c4b1c51c50215639ca58326f6b69867743/src/Mod/Sketcher/App/planegcs/GCS.cpp#L590
Called by "Clear" (free):
https://github.com/FreeCAD/FreeCAD/blob/1dcc58c4b1c51c50215639ca58326f6b69867743/src/Mod/Sketcher/App/planegcs/GCS.cpp#L5588

So, the `GCS::System` actually owns what is passed to it through `addConstraint` and stored in `clist`.

To follow recommendations, we should actually change `addConstraint` to take a `unique_ptr`.

andre-caldas: > where is this function getting called?

Sorry... I hadn't answered this!
It is called all over the same .cpp file. There are different addXXXXConstraints:
https://github.com/FreeCAD/FreeCAD/blob/1dcc58c4b1c51c50215639ca58326f6b69867743/src/Mod/Sketcher/App/planegcs/GCS.cpp#L662

They do create a new constraint and "forget about it". Well... if `addConstraint` was "private" it would be forgivable. :-)

andre-caldas:
The automated windows build in github is giving an error. I think there is something the other compilers see as an rvalue and the windows compiler doesn't.

How can I figure out the offending line of code?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325">pipeline status</a> for feature branch [PR_9325](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325). Pipeline [845320580](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/845320580) was triggered at [1ccfe56](https://github.com/FreeCAD/FreeCAD/pull/9325/commits/1ccfe56332945e895cbcd4eb14b7afe8646b4754). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325">pipeline status</a> for feature branch [PR_9325](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325). Pipeline [845459485](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/845459485) was triggered at [aec3954](https://github.com/FreeCAD/FreeCAD/pull/9325/commits/aec39541dd544ffad6fdb90a09b2d831ec7c104a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

andre-caldas:
I really do not understand the build errors. My branch does not show any errors.
https://github.com/andre-caldas/FreeCAD/actions/runs/4774473193

Maybe this message has anything to do with it...
> Pull request is based on a 76 hour-old commit
> Pull request should be rebasedChangeset is composed of 4 commit(s)

chennes:
Did you try to compile locally after your last rebase to upstream? Those errors are coming from the test code that was added a couple of hours ago, you will have to change the lambda that's being used to count constraints by ID.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325">pipeline status</a> for feature branch [PR_9325](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325). Pipeline [845547618](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/845547618) was triggered at [2ce52c4](https://github.com/FreeCAD/FreeCAD/pull/9325/commits/2ce52c4a55397423d74b9c6db7377f0949c334a4). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325">pipeline status</a> for feature branch [PR_9325](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325). Pipeline [845555012](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/845555012) was triggered at [0e27e3e](https://github.com/FreeCAD/FreeCAD/pull/9325/commits/0e27e3e89f3220475567bd25d72be3fec3d7c0f7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

andre-caldas:
Dear @chennes,

I guess the tests are supposed to work with `std::vector<Constraint*>` and `std::vector<std::unique_ptr<Constraint>`. Unless the tests need to access the value of the pointer, two things can be done to make things work on both:
1. Use `clist.cbegin()` and `clist.cend()`. (actually, probably I should make it `std::vector<const std::unique_ptr<Constraint>`... but it is a good idea, anyway, because you probably do not want to change the pointers)
2. Use `auto& constraint` in those "lambda functions", because `std::unique_ptr` cannot be copied.

andre-caldas:
>     (actually, probably I should make it `std::vector<const std::unique_ptr<Constraint>`... but it is a good idea, anyway, because you probably do not want to change the pointers)

Now that I --- like the pointers --- am getting smarter... :-)
... I realize that probably the `std::unique_ptr` inside the vector cannot be `const`, because it needs to be relocated when vector resizes. :-)

chennes:
Yes, modify that introspection code however you need to such that it delivers the expected result -- I'm partial to the `auto` solution, but don't have a strong opinion.

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -1613,16 +1619,12 @@ void System::initSolution(Algorithm alg)
             return;
     }
     std::vector<Constraint *> clistR;
-    if (!redundant.empty()) {
-        for (std::vector<Constraint*>::const_iterator constr = clist.begin(); constr != clist.end();
-             ++constr) {
-            if (redundant.count(*constr) == 0)
-                    clistR.push_back(*constr);
+    for (auto constr = clist.cbegin(); constr != clist.cend(); ++constr) {
```
0penBrain, line 1622: Use range for loop ? With const access ?

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -1613,16 +1619,12 @@ void System::initSolution(Algorithm alg)
...
+        if (redundant.empty() || redundant.count(constr->get()) != 0)
```
0penBrain, line 1623: Compared to actual code, this is opposite logic. Is that wanted ?

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -4632,7 +4634,7 @@ void System::makeReducedJacobian(Eigen::MatrixXd &J,
 
     int jacobianconstraintcount=0;
     int allcount=0;
-    for (std::vector<Constraint *>::iterator constr=clist.begin(); constr != clist.end(); ++constr) {
+    for (auto constr=clist.cbegin(); constr != clist.cend(); ++constr) {
```
0penBrain, line 4637: Use range for loop ? With const access ?

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -5378,10 +5380,9 @@ void System::identifyConflictingRedundantConstraints(
 
     std::vector<Constraint *> clistTmp;
     clistTmp.reserve(clist.size());
-    for (std::vector<Constraint *>::iterator constr=clist.begin();
-        constr != clist.end(); ++constr) {
-        if ((*constr)->isDriving() && skipped.count(*constr) == 0)
-            clistTmp.push_back(*constr);
+    for (auto constr=clist.cbegin(); constr != clist.cend(); ++constr) {
+        if ((*constr)->isDriving() && skipped.count(constr->get()) == 0)
```
0penBrain, line 5384: Use range for loop ? With const access ?

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -5470,8 +5471,8 @@ void System::identifyConflictingRedundantConstraints(
     }
 
     // remove tags represented at least in one non-redundant constraint
-    for (std::vector<Constraint *>::iterator constr=clist.begin(); constr != clist.end(); ++constr)
-        if (redundant.count(*constr) == 0)
+    for (auto constr=clist.cbegin(); constr != clist.cend(); ++constr)
```
0penBrain, line 5474: Use range for loop ? With const access ?

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -601,14 +600,13 @@ void System::invalidatedDiagnosis()
 
 void System::clearByTag(int tagId)
 {
+    // TODO: Check if we could simply remove in reverse order.
     std::vector<Constraint *> constrvec;
-    for (std::vector<Constraint *>::const_iterator
-         constr=clist.begin(); constr != clist.end(); ++constr) {
+    for (auto constr=clist.cbegin(); constr != clist.cend(); ++constr) {
```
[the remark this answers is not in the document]
andre-caldas: Actually, it seems that this whole `removeConstraint()` thing does not allow things to be done properly. I guess the ideal would be to use the so called "erase-remove" idiom. Calling consecutive `erase` on a vector is not good. This is the only place `removeConstraint()` is called by FreeCAD official repo.

Notice that we pass a `Constraint*`, and that is used later to find an iterator again! We could simply have passed the iterator.

But I do believe that:
1. `removeConstraint()` should be made private and renamed to something like `prepareForConstraintRemoval()`.
2. The iterator part should not be dealt with inside `prepareForConstraintRemoval()`.
3. In special, `prepareForConstraintRemoval()` should not in fact remove it from `clist`.
4. Use erase-remove idiom like in here: https://stackoverflow.com/a/56197450/1290711
```
clist.erase(std::remove_if(clist.cbegin(), clist.cend(), [&](auto& constr) {
    if (constr.getTag() == tagId) {
        prepareForConstraintRemoval(constr->get());
        return true;
    }
    return false;
}), clist.cend());

...

void prepareForConstraintRemoval(Constraint* constr)
{
    if (constr->getTag() >= 0)
        hasDiagnosis = false;
    clearSubSystems();

    VEC_pD constr_params = c2p[constr];
    for (VEC_pD::const_iterator param=constr_params.begin();
         param != constr_params.end(); ++param) {
        std::vector<Constraint *> &constraints = p2c[*param];
        auto i = std::find(constraints.begin(), constraints.end(), constr);
        constraints.erase(i);
    }
    c2p.erase(constr);
}
```

Would that be appropriate?

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -601,14 +600,13 @@ void System::invalidatedDiagnosis()
...
         if ((*constr)->getTag() == tagId)
-            constrvec.push_back(*constr);
+            constrvec.push_back(constr->get());
     }
-    for (std::vector<Constraint *>::const_iterator
-         constr=constrvec.begin(); constr != constrvec.end(); ++constr) {
+    for (auto constr=constrvec.cbegin(); constr != constrvec.cend(); ++constr) {
```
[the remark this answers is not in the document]
andre-caldas: The `prepareForConstraintRemoval()` would eliminate this loop.

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -632,12 +630,21 @@ int System::addConstraint(Constraint *constr)
...
-    it = std::find(clist.begin(), clist.end(), constr);
-    if (it == clist.end())
+    auto it = clist.cbegin();
+    for(; it != clist.cend(); ++it)
+    {
+        if(it->get() == constr)
+        {
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
andre-caldas: The `prepareForConstraintRemoval()` would eliminate this loop.

Notice that we iterate, find the iterator, pass the address, then, **for each match** we iterate again... and find it again...

src/Mod/Sketcher/App/planegcs/GCS.cpp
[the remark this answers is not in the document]
0penBrain: Huh. I didn't dig that deep to know. :)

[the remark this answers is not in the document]
andre-caldas: But, as a "FreeCAD" culture thing...
... would it be okay to entirely remove `removeConstraint()`? Or shall I leave it?
... would it be okay to create a new private `prepareForConstraintRemoval()`?

I was trying to change the least as possible. But I guess this was not the right choice. :-)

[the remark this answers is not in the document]
0penBrain: Improving is always welcome, whatever are the changes. But keep in mind that the bigger is change set, the harder it is to get the PR reviewed and merged. ;)

src/Mod/Sketcher/App/planegcs/GCS.cpp
andre-caldas: Done.

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -646,14 +652,11 @@ void System::removeConstraint(Constraint *constr)
     for (VEC_pD::const_iterator param=constr_params.begin();
          param != constr_params.end(); ++param) {
         std::vector<Constraint *> &constraints = p2c[*param];
-        it = std::find(constraints.begin(), constraints.end(), constr);
-        constraints.erase(it);
+        auto i = std::find(constraints.begin(), constraints.end(), constr);
```
0penBrain, line 655: Maybe use a more significant variable name.

src/Mod/Sketcher/App/planegcs/GCS.cpp
```diff
@@ -632,12 +628,22 @@ int System::addConstraint(Constraint *constr)
...
-    std::vector<Constraint *>::iterator it;
-    it = std::find(clist.begin(), clist.end(), constr);
-    if (it == clist.end())
+    auto it = std::find_if(clist.cbegin(), clist.cend(), [constr](auto& candidate){
+        return (candidate.get() == constr);
+    });
+    if (it == clist.cend())
+    {
+        // TODO: someone is removing a nonexistent constraint.
+        // Shouldn't we at least emit a warning?
```
0penBrain, line 637: Yes, probably good to emit warning, at least in debug build

0penBrain (commented):
LGTM from just code reviewing.
Then I don't know GCS so not able to check if implementation is functionally correct.

@abdullahtahiriyo Do you plan to have a look at this ?

@andre-caldas May you please clean up a bit the commits ? Now it's a bit of a mess with merge commits and all. Thx

abdullahtahiriyo:
@0penBrain 

Yes, I will review this PR before merge. I have shifted this to 0.22 though. So after release.

That does not mean the PR does not benefit from all your work. It does and I am grateful to you all.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325">pipeline status</a> for feature branch [PR_9325](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9325). Pipeline [903667659](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903667659) was triggered at [4da1151](https://github.com/FreeCAD/FreeCAD/pull/9325/commits/4da115174bb97316f804a112e88eb50bb5b9ba6f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

andre-caldas:
I don't know if changes to make whitespaces uniform is good or bad... but it surely causes conflicts that are hard to merge. The probability of unconsciously removing already commited code while resolving them is very high, IMO.

https://github.com/FreeCAD/FreeCAD/commit/2ae7cc4a94c1ab1dfc0264367c5da4716ea02bfe#diff-991727098558e5924b885b1d727bbfdc567e4e5a23b961679d274df2763ac73c

berniev:
or reformat your branch ...?

luzpaz:
Make whitespace formatting changes makes things more complicated. You can make them in a separate PR for faster reviewing, @andre-caldas

andre-caldas:
Dear all,

I am very sorry. If you don't mind, I think I want to withdraw this PR.

PR #9328, opened 2023-04-20
Sketcher: Fix Issue 6283

abdullahtahiriyo:


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9328">pipeline status</a> for feature branch [PR_9328](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9328). Pipeline [843783692](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/843783692) was triggered at [e786999](https://github.com/FreeCAD/FreeCAD/pull/9328/commits/e7869993fa91d007fd33ff3a9c3fdef858cf1a84). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/Sketcher/Gui/SoZoomTranslation.cpp
```diff
@@ -66,16 +66,16 @@ void SoZoomTranslation::initClass()
...
     SbViewVolume vv = SoViewVolumeElement::get(action->getState());
     float aspectRatio = SoViewportRegionElement::get(action->getState()).getViewportAspectRatio();
-    float scale = vv.getWorldToScreenScale(SbVec3f(0.f, 0.f, 0.f), 0.1f) / (5*aspectRatio);
-    return scale;
+    scaleFactor = vv.getWorldToScreenScale(SbVec3f(0.f, 0.f, 0.f), 0.1f) / (5*aspectRatio);
+    return scaleFactor;
 }
 
-SoZoomTranslation::SoZoomTranslation()
+SoZoomTranslation::SoZoomTranslation(): scaleFactor(0)
```
AjinkyaDahale, line 78: Default 0 for `scaleFactor` might not be the best idea, especially if there is division involved. 1 sounds better, depending on what the definition is.

src/Mod/Sketcher/Gui/SoZoomTranslation.cpp
```diff
@@ -96,7 +96,7 @@ void SoZoomTranslation::doAction(SoAction * action)
         SbVec3f absVtr = this->abPos.getValue();
         SbVec3f relVtr = this->translation.getValue();
 
-        float sf = this->getScaleFactor(action);
+        float sf = this->calculateScaleFactor(action);
         // For Sketcher Keep Z value the same
         relVtr[0] = (relVtr[0] != 0) ? sf * relVtr[0] : 0;
```
AjinkyaDahale, line 101: This is the same as `relVtr[0] *= sf;`

src/Mod/Sketcher/Gui/SoZoomTranslation.cpp
abdullahtahiriyo: The objective is to store whatever the ZoomTranslation calculates. 

Retrieving the scaleFactor before (internal) calculation will always result in a wrong value. This means that 1 will be as wrong as 0.

There is no risk of exception as the the scale factor is multiplied to a relative translation.

Then, perhaps 0 best serves a user of the client code to realise the mistake, if trying to retrieve a scale before the node is actually in running in the scenegraph.

src/Mod/Sketcher/Gui/SoZoomTranslation.cpp
```diff
@@ -116,7 +116,7 @@ void SoZoomTranslation::getMatrix(SoGetMatrixAction * action)
         SbVec3f absVtr = this->abPos.getValue();
         SbVec3f relVtr = this->translation.getValue();
 
-        float sf = this->getScaleFactor(action);
+        float sf = this->calculateScaleFactor(action);
         // For Sketcher Keep Z value the same
         relVtr[0] = (relVtr[0] != 0) ? sf  * relVtr[0] : 0;
```
AjinkyaDahale, line 121: See previous comment.

src/Mod/Sketcher/Gui/SoZoomTranslation.cpp
abdullahtahiriyo: Yes, you are indeed right. But I did not write that line...

I have also seen other things I do not like from this class (such as a hardcoded 5 in calculateScaleFactor). 

I have fought the temptation to make a full review of SoZoomTranslation internal workings at this time. 

Just adding a way to retrieve the scaling factor it uses, because that is what is necessary to fix the actual bug of the issue.

If you have time to do a full review and wish to do it, feel free to make a PR. It is welcome.

AjinkyaDahale: Maybe eventually. We could create an issue about it, but then again we'd need as many as the number of files qe have!

PR #9336, opened 2023-04-20
PD: Helix parameters status translation

chennes:
This probably fixes https://github.com/FreeCAD/FreeCAD-translations/issues/22, but it will require verification because the messages don't quite match the reported error.

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskHelixParameters.cpp
@@ -275,16 +275,17 @@ void TaskHelixParameters::updateStatus()
 {
     auto pcHelix = static_cast<PartDesign::Helix*>(vp->getObject());
     auto status = std::string(pcHelix->getStatusString());
+    QString translatedStatus;
     if (status.compare("Valid") == 0 || status.compare("Touched") == 0) {
-        if (pcHelix->safePitch() > pcHelix->Pitch.getValue())
-            status = "Warning: helix might be self intersecting";
-        else
-            status = "";
+        if (pcHelix->safePitch() > pcHelix->Pitch.getValue()) {
+            translatedStatus = tr("Warning: helix might be self intersecting");
+        }
     }
     // if the helix touches itself along a single helical edge we get this error
-    else if (status.compare("NCollection_IndexedDataMap::FindFromKey") == 0)
-        status = "Error: helix touches itself";
-    ui->labelMessage->setText(QString::fromUtf8(status.c_str()));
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9336">pipeline status</a> for feature branch [PR_9336](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9336). Pipeline [843948247](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/843948247) was triggered at [2be234f](https://github.com/FreeCAD/FreeCAD/pull/9336/commits/2be234f59b016bffed4081ec7cefbec17036f264). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
The message comes from: https://github.com/FreeCAD/FreeCAD/blob/2be234f59b016bffed4081ec7cefbec17036f264/src/Mod/PartDesign/App/FeatureHelix.cpp#L448

chennes:
We have a lot of these, where we are directly showing the text of an exception to end users: do you have any thoughts on a translation approach? Or just displaying a more generic error on-screen, with the exception text in Report View?

wwmayer:
* We can decide what context string is used for exceptions (e.g. "Exception")
* Mark then the text as `throw Base::ValueError(QT_TRANSLATE_NOOP("Exception", "Error: Result is self intersecting"));`
* In the dialog or wherever it's shown we translate it: `qApp->translate("Exception", exc.what())`

The whole point is to have a reliable context string

chennes:
Do you think we need to do this to all thrown `Base::Exception`s? Just `ValueError`? I don't have a sense of how often we are showing exception text directly to end users, and I couldn't think of a good way to search for it.

wwmayer:
A way is to search for what(). In PD there are 10 hits where an exception is shown in a dialog. From there we can find the source where the exception is thrown.

chennes:
In src/Gui/Command.cpp on line 489 we are popping up a QMessageBox having caught a Base::Exception -- to me this makes it seem like literally any exception throw from within a FreeCAD command could potentially be show to an end user.

wwmayer:
I think doing it this way is fine because it's up to the author of a command class to handle exceptions in a more meaningful way. The function Command::_invoke() is just the fallback to handle all exception types to avoid that they go through Qt code.

chennes:
OK - I'm going to merge this translation fix so at least this one small instance shows a translated error message, and then start to work on tracking down the exceptions that are getting shown to users and see if we can start getting those translated.

PR #9337, opened 2023-04-21
TechDraw: Translate leader line arrow shapes

chennes:
Fixes https://github.com/FreeCAD/FreeCAD-translations/issues/46. The base enum is unchanged by this operation (which uses NOOP to ensure the char * is unchanged). Only on actual display in the user interface is the translated value used. It appears that this was actually the original intent, but it didn't work because the translatable strings were never added to the database. Internally, only the numeric index of the current entry is used, so the actual displayed text can be anything.

### DIFF
--- a/src/Mod/TechDraw/Gui/DrawGuiUtil.cpp
+++ b/src/Mod/TechDraw/Gui/DrawGuiUtil.cpp
@@ -73,7 +73,7 @@ void DrawGuiUtil::loadArrowBox(QComboBox* qcb)
     qcb->clear();
     int i = 0;
     for (; i < ArrowPropEnum::ArrowCount; i++) {
-        qcb->addItem(tr(ArrowPropEnum::ArrowTypeEnums[i]));
+        qcb->addItem(QCoreApplication::translate("ArrowPropEnum", ArrowPropEnum::ArrowTypeEnums[i]));
         QIcon itemIcon(QString::fromUtf8(ArrowPropEnum::ArrowTypeIcons[i].c_str()));
         qcb->setItemIcon(i, itemIcon);
     }
--- a/src/Mod/TechDraw/App/ArrowPropEnum.cpp
+++ b/src/Mod/TechDraw/App/ArrowPropEnum.cpp
@@ -28,15 +28,16 @@
 namespace TechDraw {
 
 const int   ArrowPropEnum::ArrowCount = 8;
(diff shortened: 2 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9337">pipeline status</a> for feature branch [PR_9337](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9337). Pipeline [844076423](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/844076423) was triggered at [07f6cb6](https://github.com/FreeCAD/FreeCAD/pull/9337/commits/07f6cb650580bbb1337cbcf26bc17a2750de2c3d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

src/Mod/TechDraw/Gui/DrawGuiUtil.cpp
```diff
@@ -73,7 +73,7 @@ void DrawGuiUtil::loadArrowBox(QComboBox* qcb)
     qcb->clear();
     int i = 0;
     for (; i < ArrowPropEnum::ArrowCount; i++) {
-        qcb->addItem(tr(ArrowPropEnum::ArrowTypeEnums[i]));
+        qcb->addItem(QCoreApplication::translate("ArrowPropEnum", ArrowPropEnum::ArrowTypeEnums[i]));
```
WandererFan, line 76: If the strings are translated in the definition of the enum, they don't need to be retranslated here(?).

I didn't think it was possible to translate the data inside a variable.  This could be:
qcb->addItem(ArrowPropEnum::ArrowTypeEnums[i]);

chennes: The strings aren't translated in the enum definition: that array contains the original English language `const char *` data, the `QT_TRANSLATE_NOOP` function is a literal NOOP as far as the C++ compiler is concerned. The macro expands to the original string argument and nothing else. But `lupdat`e sees it what it scans the source code and extracts those strings. So now when translate is called, even though `lupdate` can't extract anything from the translate call (because it doesn't contain a string literal), the actual real translate function call still occurs at runtime, and the correct string is looked up in the database. This preserves the original English text being stored in the FCStd files, while still giving the user a translated interface.

PR #9358, opened 2023-04-22
Sketcher: Change behaviour of toggle of mixed group of edges and vertices

abdullahtahiriyo:
Forum:
https://forum.freecad.org/viewtopic.php?p=677304#p677304

A common operation is to toggle a group of geometry to/from construction. This happens often in connection with carbon copy operations.

When this happens, geometry is often selected using box selection.

If there are geometry points (which are generally construction), and selection box is used, this results in construction points being toggled to normal defining mode. This situation is undesirable in 99% of the cases. It requires to reselect only the points and toggle them back.

When construction points need be made defining geometry (for example to use them as attachments), the general workflow is to select only those points and toggle them.

In order to improve the efficiency, it has been decided to provide a special behaviour for construction toggling:

1. Vertices will only be toggled to/from construction IF ONLY vertices are selected.
2. If there is a mixture of edges and vertices, vertices will be ignored during the toggling operation.

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandAlterGeometry.cpp
+++ b/src/Mod/Sketcher/Gui/CommandAlterGeometry.cpp
@@ -155,14 +155,34 @@ void CmdSketcherToggleConstruction::activated(int iMsg)
         openCommand(QT_TRANSLATE_NOOP("Command", "Toggle draft from/to draft"));
 
         // go through the selected subelements
+        bool verticesonly = true;
+
+        for(const auto & subname : SubNames) {
+            if (subname.size() > 4 && subname.substr(0,4) == "Edge") {
+                verticesonly = false;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9358">pipeline status</a> for feature branch [PR_9358](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9358). Pipeline [845450634](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/845450634) was triggered at [f9cdd49](https://github.com/FreeCAD/FreeCAD/pull/9358/commits/f9cdd49b2b4f15800e40e843a44af11c35006b4a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #9361, opened 2023-04-22
Gui: added semi transparent face to origin planes

ronnystandtke:
The planes of the origin currently consist only of the surrounding lines, which are very difficult to distinguish visually, especially when rotating.

In this pull request I have added a semi transparent face to each plane, which improves their visibility.

---

### DIFF
--- a/src/Gui/ViewProviderOrigin.h
+++ b/src/Gui/ViewProviderOrigin.h
@@ -76,6 +76,9 @@ public:
 
     /// Returns default size. Use this if it is not possible to determine appropriate size by other means
     static double defaultSize();
+
+    // default color for origini: light-blue (50, 150, 250, 255 stored as 0xRRGGBBAA)
+    static const uint32_t defaultColor = 0x3296faff;
 protected:
     void onChanged(const App::Property* prop) override;
     bool onDelete(const std::vector<std::string> &) override;
--- a/src/Gui/ViewProviderOrigin.h
+++ b/src/Gui/ViewProviderOrigin.h
@@ -77,6 +77,9 @@ public:
     /// Returns default size. Use this if it is not possible to determine appropriate size by other means
     static double defaultSize();
 
+    // the factor by which the axes are longer than the planes
+    static constexpr float axesScaling = 1.5f;
+
     // default color for origini: light-blue (50, 150, 250, 255 stored as 0xRRGGBBAA)
     static const uint32_t defaultColor = 0x3296faff;
 protected:
--- a/src/Gui/ViewProviderPlane.cpp
+++ b/src/Gui/ViewProviderPlane.cpp
@@ -96,6 +96,11 @@ void ViewProviderPlane::attach ( App::DocumentObject *obj ) {
     shapeHints->shapeType = SoShapeHints::UNKNOWN_SHAPE_TYPE;
     faceSeparator->addChild(shapeHints);
 
+    // disable picking
+    auto pickStyle = new SoPickStyle();
+    pickStyle->style = SoPickStyle::UNPICKABLE;
+    faceSeparator->addChild(pickStyle);
+
     auto faceSet = new SoFaceSet();
     auto vertexProperty = new SoVertexProperty();
     vertexProperty->vertex.setValues(0, 4, verts);
--- a/src/Gui/ViewProviderPlane.cpp
+++ b/src/Gui/ViewProviderPlane.cpp
@@ -26,10 +26,14 @@
 #ifndef _PreComp_
 # include <Inventor/nodes/SoAsciiText.h>
 # include <Inventor/nodes/SoCoordinate3.h>
+# include <Inventor/nodes/SoFaceSet.h>
 # include <Inventor/nodes/SoIndexedLineSet.h>
+# include <Inventor/nodes/SoMaterial.h>
 # include <Inventor/nodes/SoPickStyle.h>
 # include <Inventor/nodes/SoSeparator.h>
+# include <Inventor/nodes/SoShapeHints.h>
 # include <Inventor/nodes/SoTranslation.h>
+# include <Inventor/SbColor.h>
 #endif
 
 #include "ViewProviderPlane.h"
@@ -72,6 +76,31 @@ void ViewProviderPlane::attach ( App::DocumentObject *obj ) {
     pLines->coordIndex.setNum(6);
     pLines->coordIndex.setValues(0, 6, lines);
     sep->addChild ( pLines );
+
+    // add semi transparent face
+    auto faceSeparator = new SoSeparator();
+    sep->addChild(faceSeparator);
+
+    auto material = new SoMaterial();
+    material->transparency.setValue(0.95f);
+    auto color = new SbColor();
+    float alpha = 0.0f;
+    color->setPackedValue(ViewProviderOrigin::defaultColor, alpha);
+    material->ambientColor.setValue(*color);
+    material->diffuseColor.setValue(*color);
+    faceSeparator->addChild(material);
+
+    // disable backface culling and render with two-sided lighting
+    auto shapeHints = new SoShapeHints();
+    shapeHints->vertexOrdering = SoShapeHints::COUNTERCLOCKWISE;
+    shapeHints->shapeType = SoShapeHints::UNKNOWN_SHAPE_TYPE;
+    faceSeparator->addChild(shapeHints);
+
+    auto faceSet = new SoFaceSet();
+    auto vertexProperty = new SoVertexProperty();
+    vertexProperty->vertex.setValues(0, 4, verts);
+    faceSet->vertexProperty.setValue(vertexProperty);
+    faceSeparator->addChild(faceSet);
 
     auto textTranslation = new SoTranslation ();
     textTranslation->translation.setValue ( SbVec3f ( -size * 49. / 50., size * 9./10., 0 ) );
--- a/src/Gui/ViewProviderOrigin.cpp
+++ b/src/Gui/ViewProviderOrigin.cpp
@@ -182,9 +182,9 @@ void ViewProviderOrigin::onChanged(const App::Property* prop) {
             if (vpPlaneXY) { vpPlaneXY->Size.setValue ( szXY ); }
             if (vpPlaneXZ) { vpPlaneXZ->Size.setValue ( szXZ ); }
             if (vpPlaneYZ) { vpPlaneYZ->Size.setValue ( szYZ ); }
-            if (vpLineX) { vpLineX->Size.setValue ( szX ); }
-            if (vpLineY) { vpLineY->Size.setValue ( szY ); }
-            if (vpLineZ) { vpLineZ->Size.setValue ( szZ ); }
+            if (vpLineX) { vpLineX->Size.setValue ( szX * axesScaling ); }
+            if (vpLineY) { vpLineY->Size.setValue ( szY * axesScaling ); }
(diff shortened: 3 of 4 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9361">pipeline status</a> for feature branch [PR_9361](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9361). Pipeline [845498473](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/845498473) was triggered at [c61d64c](https://github.com/FreeCAD/FreeCAD/pull/9361/commits/c61d64c72ddf0cbeb5fd07683bdc631e47719878). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
Would you mind adding a screenshot of what you've added?

src/Gui/ViewProviderOrigin.h
```diff
@@ -76,6 +76,9 @@ class GuiExport ViewProviderOrigin : public ViewProviderDocumentObject
 
     /// Returns default size. Use this if it is not possible to determine appropriate size by other means
     static double defaultSize();
+
+    // default color for origin (light-blue stored as 0xRRGGBBAA)
+    static const uint32_t defaultColor = (50<<24) + (150<<16) + (250<<8) + 255;
```
adrianinsaval, line 81: Is this really the best way to define this parameter?

adrianinsaval:
In general I like the feature, however it makes selecting the axis very hard when all are visible:

https://user-images.githubusercontent.com/36372335/234138174-2132668a-c20b-40c3-b5fc-653ab60cd5f6.mp4

consider making the axis bigger so that their label is outside the planes

adrianinsaval:
This also means that stuff behind the planes cannot be selected anymore, so I would like more community feedback on this before merging, could you create a forum topic for this?

src/Gui/ViewProviderOrigin.h
ronnystandtke: RGBA seems to be the only common ground between the FreeCAD API and the Coin3D API when you try to set a color with a single variable, see:
https://freecad.github.io/SourceDoc/d9/d0b/classApp_1_1PropertyColor.html
https://www.coin3d.org/Coin/html/classSbColor.html

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9361">pipeline status</a> for feature branch [PR_9361](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9361). Pipeline [848686025](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/848686025) was triggered at [4917018](https://github.com/FreeCAD/FreeCAD/pull/9361/commits/49170184a5137c756e7eb80fc28320b346526b78). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

ronnystandtke:
Thank you for your feedback. I'm glad you like the change.

> consider making the axis bigger so that their label is outside the planes

OK, I did so in the latest commit. This way the axes are always selectable. Could you take another look at it? If you then think that all is well, I would continue the discussion in the forum.

adrianinsaval:
> OK, I did so in the latest commit. This way the axes are always selectable.

Excellent! the concern I have left is with general selectability of objects, in master branch objects behind the planes are easily selected, but with this PR one may need to rotate the view to ensure the object is not obstructed by the planes. This is what i would like some feedback from the community, we may need to make this filling optional. Selecting objects coincident with the plane is also unreliable

https://user-images.githubusercontent.com/36372335/234366873-d9a392c7-f533-4d6c-9a0f-b8cd8a7ea149.mp4

ronnystandtke:
> Excellent! the concern I have left is with general selectability of objects, 

I would handle this like all situations with more complex geometry: just hide the origin and bring it back when needed. Do you think this would be acceptable?

ronnystandtke:
OK, forum topic is [here](https://forum.freecad.org/viewtopic.php?t=77911).

adrianinsaval:
> I would handle this like all situations with more complex geometry: just hide the origin and bring it back when needed. Do you think this would be acceptable?

I'm not quite sure, because the origin planes are made automatically visible when creating a sketch in Part Design, so I think it may end up being very inconvenient to have this face by default. If it's just for visualization purposes we could keep the face but make it unselectable, but then it's usefulness is questionable.

I'm not sure if there is a way to tell coin to prioritize selection of all other objects before the plane, if something like that is possible then that would be the ideal solution.

GeneGH:
This absolutely needs an "opt out". Hindering selection in any manner is a losing approach, regardless of it looks.

src/Gui/ViewProviderOrigin.h
```diff
@@ -76,6 +76,12 @@ class GuiExport ViewProviderOrigin : public ViewProviderDocumentObject
 
     /// Returns default size. Use this if it is not possible to determine appropriate size by other means
     static double defaultSize();
+
+    // the factor by which the axes are longer than the planes
+    static constexpr float axesScaling = 1.5f;
+
+    // default color for origin (light-blue stored as 0xRRGGBBAA)
+    static const uint32_t defaultColor = (50<<24) + (150<<16) + (250<<8) + 255;
```
adrianinsaval, line 84: maybe this works:
```suggestion
    static const uint32_t defaultColor = 0x3296faff;
```
I think storing as hex is more easily understood as this is standard format for colors

ronnystandtke:
We could make the planes unpickable with [SoPickStyle::Style::UNPICKABLE](https://www.coin3d.org/Coin/html/classSoPickStyle.html#a8c23088ad3a4eb7bdeaa93ae5a29b910).

If you think this is the right approach, I would add this and revert the last commit where I enlarged the axes. Should I?

src/Gui/ViewProviderOrigin.h
ronnystandtke: I think mere mortal humans are more familiar with the decimal system (RGB for light-blue = 50, 150, 250) but I am not obsessed with this. If you like to store it as a hex value, I don't mind.

adrianinsaval:
Let's wait for more feedback before acting on that front
in my personal opinion enlarging the axis seems like a good move regardless of plane selectability as it is now on master it is not easy to distinguish them when all is visible.

src/Gui/ViewProviderOrigin.h
adrianinsaval: the syntax with sums and << operators obscures this to regular people IMO, hex representations of color are very common on any graphic editing software

ronnystandtke:
OK, let's wait then.

By the way, Coin3D has the pick style `SHAPE_ON_TOP`. Unfortunately, the opposite seems to be missing, there is no `SHAPE_UNDERNEATH`.

I also took a more closer look at the implementation in Onshape and they do it really clever. The planes are visible and selectable but they are actually `SHAPE_UNDERNEATH`. Someone more familiar with the Coin3D API can hopefully give some tips and tricks how to do the same here...

adrianinsaval:
@realthunder can you give some input here? I know in your branch you have a mechanism to select stuff behind other stuff, is there a way to make a specific face have the lowest priority for selection?

realthunder:
I implemented the high priority picking group in my branch, not the low priority, but the same principle applies. The scene of a FreeCAD document is grouped under a `SoFCUnifiedSelection` node. The picking is initiated in `SoFCUnifiedSelection::handleEvent()` by traversing a `SoRayPick` action down the scene node. I added a new group node at the same hierarchy as the scene node, and perform a picking traversal through that node before the normal picking. Similarly, low priority picking can be done by adding another group node at the same hierarchy and perform an extra picking traversal. But, that would involve some changes to the core, which I think is outside the scope of this PR.

adrianinsaval:
Indeed outside the scope, I guess for now we make it unpickable (maybe optionally?) and then we'll see when we can work on implementing selection priorities in the core.

ronnystandtke:
OK, I made the planes unpickable and colored the axes the same way as the axis cross, the navicube and the coordinate system in the corner. This should make the origin more consistent with the other items and improve its visibility without affecting general selectability of objects. I am very curious to hear your feedback.

adrianinsaval:
very nice and I'm willing to merge as is, however a few improvements I would like if you have the time:
• instead of hardcoding the color of the transparent face it would be nice if it used the color property of the plane but just make it transparent. This way they won't look so odd if you manually change the color of the planes.
• it would be nice if the labels also had backside rendering so they don't look black on the backside
image

• I think we could try making the borders and labels use `SHAPE_ON_TOP` pick style, this may seem unnecessary on first thought but I think it can result useful when there are already other existing bodies that cover the entire plane, this should be done as a follow up PR however, not here.
   • in such PR we could consider adding and additional invisible frame that would be selectable to make it easier to hit the target than with just the thin borders as is now, the frame should be outside the bounds of the body bounding box so it won't interfere with object selection. This can be a nice stop gap while we explore implementing selection priorities.
   • we can also consider using the colors of the respective normal axis to identify the planes by color, but this is low priority and we would have to check if it's not just too much visual clutter in practice, needs some community discussion

adrianinsaval (changes requested):

src/Gui/ViewProviderPlane.cpp
```diff
@@ -73,6 +77,36 @@ void ViewProviderPlane::attach ( App::DocumentObject *obj ) {
...
 
+    // add semi transparent face
+    auto faceSeparator = new SoSeparator();
+    sep->addChild(faceSeparator);
+
+    auto material = new SoMaterial();
+    material->transparency.setValue(0.95f);
+    auto color = new SbColor();
+    float alpha = 0.0f;
+    color->setPackedValue(ViewProviderOrigin::defaultColor, alpha);
```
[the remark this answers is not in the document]
adrianinsaval: this doesn't actually work as it is not updated, for this we'll need to implement ViewProviderPlane::onChanged for ShapeColor property

adrianinsaval:
I've reworded and squashed some commits in more logical way for source control, once the CI passes I'll merge

wwmayer:
A minor issue is that the backside of the labels of the axes is black. If backlight color is enabled in the viewer settings then the backside of the green and blue labels might be rendered with the same color but the red label is still black.

Explicitly setting a light model to disable the lighting (SoLightModel::BASE_COLOR) fixes the issue: https://github.com/FreeCAD/FreeCAD/pull/9482

ronnystandtke:
Thank you very much for your support and merging the PR. The related forum discussion made me feel like this PR became a prime example for Ondsel's recent blog post:
https://ondsel.com/blog/freecad-approach-to-software-development-needs-to-change
I will still try to continue to contribute to the project.

xtemp09:
Dear participants,

Leak sanitizer reported the leak caused by [the line](https://github.com/FreeCAD/FreeCAD/blob/master/src/Gui/ViewProviderPlane.cpp#L85). Looking at the source, I suspect that the code may be intentionally written in this way.

[This commit](https://github.com/xtemp09/FreeCAD/commit/ff174b5099deb7229467a8123b9c421c20814b30) fixes the leak. Should I make the pull request?

wwmayer:
> Should I make the pull request?

Yes, make one, please.

Issue #9170, opened 2023-04-03
[Issue] CI: GTest failures show the wrong completion message and icon


chennes:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
n/a
```

### Subproject(s) affected?

None

### Issue description

When a Google Test fails, the CI correctly reports that as a failure in the final analysis, but still shows the message "GTest C++ unit test suite succeeded" with the "success" check mark in the Summary report.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
Hi @berniev do you mind taking a look at this?

berniev:
I think @OpenBrain is the expert on CI stuff?

chennes:
@0penBrain (with a zero for an O, just to confuse us all 😄 ).

berniev:
> just to confuse us 

Certainly worked! I gave up after a while, confident someone would put it right. That worked too!

0penBrain:
Could one give a link to a run with such failure?

0penBrain:
@berniev any example to give?

chennes:
Here is a PR with a failing test: https://github.com/FreeCAD/FreeCAD/pull/9367

berniev:
Creating a failing test is really really simple. I don't understand the complexity being expressed.

0penBrain:
> Here is a PR with a failing test: #9367

Thanks. I just pushed a PR. When CI ends, I'll merge and ping you if you can rebase the demonstrating PR just to check it's OK now. ;)

0penBrain:
@chennes I just merged. Could you please rebase and push #9367 please ?

Sorry BTW, I probably created merge conflicts with your CI PR... :confused:

PR #9367, opened 2023-04-23
Tests: Demonstrating a failed test

chennes:
This is a PR with a failing GTest for debugging Issue #9170.

ETA: Actually, it's demonstrating two problems! The first is that I recently added a new test target but didn't add it to the CI, and the second is the problem I meant to demo, a failing gTest.

### DIFF
--- a/tests/src/App/IndexedName.cpp
+++ b/tests/src/App/IndexedName.cpp
@@ -597,4 +597,9 @@ TEST_F(ByteArrayTest, moveAssignmentOperator)
     EXPECT_EQ(originalByteArrayLocation, newByteArray.bytes.constData());
 }
 
+TEST_F(ByteArrayTest, failingTestForTesting)
+{
+    FAIL() << "This is a failing gTest unit test";
+}
+
 // NOLINTEND(readability-magic-numbers)
--- a/tests/src/Mod/Sketcher/App/planegcs/GCS.cpp
+++ b/tests/src/Mod/Sketcher/App/planegcs/GCS.cpp
@@ -4,9 +4,11 @@
 
 #include "Mod/Sketcher/App/planegcs/GCS.h"
 
-class SystemTest : public GCS::System{
+class SystemTest: public GCS::System
+{
 public:
-    size_t getNumberOfConstraints(int tagID = -1) {
+    size_t getNumberOfConstraints(int tagID = -1)
+    {
         return _getNumberOfConstraints(tagID);
     }
 };
@@ -33,7 +35,7 @@ private:
     std::unique_ptr<SystemTest> _system;
 };
 
-TEST_F(GCSTest, clearConstraints) // NOLINT
+TEST_F(GCSTest, clearConstraints)// NOLINT
 {
     // Arrange
     const size_t numConstraints {100};
@@ -44,6 +46,22 @@ TEST_F(GCSTest, clearConstraints) // NOLINT
 
     // Act
     System()->clear();
+
+    // Assert
+    EXPECT_EQ(0, System()->getNumberOfConstraints());
+}
+
+TEST_F(GCSTest, clearConstraintsByTag)// NOLINT
+{
+    // Arrange
+    const size_t numConstraints {100};
+    for (size_t i = 0; i < numConstraints; ++i) {
+        System()->addConstraint(new GCS::Constraint());
+    }
+    ASSERT_EQ(numConstraints, System()->getNumberOfConstraints());
+
+    // Act
+    System()->clearByTag(1);
 
     // Assert
     EXPECT_EQ(0, System()->getNumberOfConstraints());


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9367">pipeline status</a> for feature branch [PR_9367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9367). Pipeline [845798938](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/845798938) was triggered at [a305f6b](https://github.com/FreeCAD/FreeCAD/pull/9367/commits/a305f6bd3f273ba931972488fc1f212886026068). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9367">pipeline status</a> for feature branch [PR_9367](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9367). Pipeline [845800036](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/845800036) was triggered at [c60a451](https://github.com/FreeCAD/FreeCAD/pull/9367/commits/c60a4512dff0979dc7c3a27bc1f926728f7294f7). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
I've got a PR in process to add the Sketcher unit tests to the CI: #9368. @0penBrain do you want me to refactor that now so that all the C++ unit tests are grouped together into a single workflow, or ca/should that wait until there are more than two? In the long run every Mod with C++ code should have an entry.

0penBrain:
> @0penBrain do you want me to refactor that now so that all the C++ unit tests are grouped together into a single workflow, or ca/should that wait until there are more than two?

I think you can start refactoring right now. 😉 You probably can't use a reusable workflow for this, but I think a "master" composite action that in turns calls a generic composite action several times with appropriate parameters should be possible.

berniev:
Please consider that many so-called "unit tests" are in fact "integration tests". I ask the two be kept separate.
Even the simplest search, wikipedia, will explain the difference. Unfortunately this is not yet well understood at FreeCAD but if not kept separate will lead to problems later.

adrianinsaval:
> but if not kept separate will lead to problems later.

what would those problems be?

berniev:
`Unit tests` come into their own when a large body of same are developed, each targeting a very small, isolated, "unit" of code. They run VERY fast. Because of the speed, a developer can run unit tests very frequently. This leads to the concept of TDD (Test Driven Development).

The problem though, is getting the code into so-called `testable` state. That is to say in a form where unit tests can be applied, which is dictated by dependencies, statics, singletons, complexity. Basically much of the stuff espoused by the likes of Uncle Bob Martin and others over the last 20 + years. FreeCAD is large, old legacy code, and in the main not `testable`.

As FreeCAD has not been written with unit tests in mind, the alternative is to perform so-called integration tests. These each encompass a range of classes, files etc in each test. Typically, due to the many dependencies, they run much, much slower.

FreeCAD has say 5000 classes etc. If each has say 20 tests, and they each take 20mS that is 33 minutes run time. They will not be run very frequently!

I have had well over 100 unit tests and cannot get the run time to register 1 mS. 

A **_single_** integration test introduced for metadata took variously from 8 to 120 mS.

Integration tests aare terrific, just keep them separate.

chennes:
I'm not sure why during development you'd ever run the full test suite -- why not run just the tests for the code you're working on?

berniev:
Google: "why unit tests must be fast". 137,000,000 answers. I politely suggest you study the subject a bit more.

You are not alone. Stating last year that FreeCAD is largely legacy code brought howls of derision from a senior FC person. Similarly, my suggestion last year that FreeCAD should have unit testing brought stern admonition from a different senior FC person "we already do unit testing". Old attitudes die hard.

The things that make unit testing possible (SOLID etc etc) are not generally understood or practised at FreeCAD, so understandably unit testing and its benefits are similarly difficult to absorb.

One of my aims in introducing unit testing was not only for new work, but to begin converting legacy code to the newer style demanded by unit tests.

OTOH, once code is testable the benefits of that, including unit testing, become apparent. It just may take a while.

chennes:
Feathers suggests a good target is 1/10th second per unit test max: that seems reasonable. In C++, compilation and linking time still dwarfs runtime for the tests for us, and will for quite some time. And if you're doing TDD, you're only running the tests on your units, so it will remain so. When you run the whole suite things like DLL load time can be amortized: for example, loading Xerces-C into memory does take a comparatively long time, something like 10ms, but it only has to happen once per executable.

berniev:
Feathers was written in 2005. He also makes a point of expecting all tests to go MUCH faster in future. He is also targeting unit tests.

For someone who is not literate in modern software design principles (SOLID etc --remember I've looked at your code), just started writing some integration tests which to this day you stilll call unit tests, never done TDD, you sure are determined to be the expert.

Again, I suggest you study the subject matter in much more detail.

0penBrain:
@berniev I'm gonna be honest. Your recurrent pedantic blabla looking like a schoolteacher having a superiority complex isn't going to help in any way.
As well for repeating the same again and again in any forum thread / issue / PR that has some direct or indirect concern with unit tests. It just clutters other's work.
I'd advise you place your efforts in improving things for real.

sliptonic:
> For someone who is not literate in modern software design principles (SOLID etc --remember I've looked at your code), just started writing some integration tests which to this day you stilll call unit tests, never done TDD, you sure are determined to be the expert.
> 

This comment is inappropriate.  It's rude, doesn't assume the best intentions of people, and is counterproductive. Consider yourself warned.  Future comments like this will receive formal reporting.

PR #8918, opened 2023-03-16
rebase removed files

mosfet80:
Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [ ]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---

PR #9376, opened 2023-04-24
Draft: Remove Drawing WB related code from DXF related files

Roy-043:
This PR removes the (hopefully) last bits of Drawing WB related code.

As mentioned here:
https://github.com/FreeCAD/FreeCAD/pull/8918#issuecomment-1493297591

Please note that both the old and the new code have their issues. Some of it seems semi obsolete. The new code works but `TechDraw.projectToDXF` does not always produce output that matches the view.

---

### DIFF
--- a/src/Mod/Draft/Resources/ui/preferences-dxf.ui
+++ b/src/Mod/Draft/Resources/ui/preferences-dxf.ui
@@ -584,11 +584,11 @@ If it is set to '0' the whole spline is treated as a straight segment.</string>
         <item>
          <widget class="Gui::PrefCheckBox" name="checkBox_5">
           <property name="toolTip">
-           <string>Drawing Views will be exported as blocks.
+           <string>TechDraw Views will be exported as blocks.
 This might fail for post DXF R12 templates.</string>
           </property>
           <property name="text">
-           <string>Export Drawing Views as blocks</string>
+           <string>Export TechDraw Views as blocks</string>
           </property>
           <property name="checked">
            <bool>true</bool>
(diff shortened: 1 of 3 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9376">pipeline status</a> for feature branch [PR_9376](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9376). Pipeline [846629724](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/846629724) was triggered at [b5a1c0a](https://github.com/FreeCAD/FreeCAD/pull/9376/commits/b5a1c0acfb430f07de270586e023261d99c10587). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

yorikvanhavre:
This looks good, but you also removed the ability to process groups? Maybe it does not make much sense actually, as there is never a group of TD views...

Roy-043:
In the Old Drawing WB you could drag and drop a group on a page and use it to organize your views. In the TD WB a group cannot be dropped on a page, so TD views in a group should not occur.

TD does have TechDraw_ClipGroups but these are not derived from `"App::DocumentObjectGroup"`, and would also require additional code for the actual clipping.

What is interesting to note is that the code in can create a DXF from a SVG based view. When creating a DXF with the TD export option SVG views, and also templates (frames and text) are currently not processed.

Issue #6029, opened 2022-02-07
Problem with negative offset "Placement"  when the preference are "Building US" units


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=4263

* **Reporter:** jylandry
* **Date submitted:** 2/10/2020
* **FreeCAD version:** 0.18
* **Category:** Bug
* **Status:** new
* **Tags:** #pending-forum, #ready-for-migration, units

# Original report text

https://forum.freecadweb.org/viewtopic.php?f=8&t=42766

When using "Placement" for an object and working with the unit in "Building US(ft-in/sqft/cuft)"

You can't work with negative number.  For example, «-1.5"» or «-1+1/2"» will translate to «-1/2»

The same for «-13'6"» that will become «-12'6"»

# Steps to reproduce

1) Put your preference in "Building US(ft-in/sqft/cuft)"

2) Create an empty document with an object like a cube (or anything will do)

3) try to change the "placement"  (Translation) -- any of X: Y: or Z:    -- Enter «-1.5"» or «-1+1/2"» or «-13'6"»

if you return in placement, the entry «-1.5"» should have been change to «-1/2"»
the entry «-13'6"» shall have been change by «-12'6"»

# FreeCAD Info

```
OS: Ubuntu 18.04.4 LTS
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.18.4.
Build type: Release
Python version: 3.6.8
Qt version: 5.9.5
Coin version: 4.0.0a
OCC version: 7.3.0
Locale: English/UnitedStates (en_US)

```

# Other bug information

* **Priority:** normal
* **Severity:** minor
* **Category:** Bug
* **Updated:** 2/5/2022

# Discussion from Mantis ticket

### Comment by @0penBrain 2020-02-10 11:00
Could that be a duplicate of #5909 ?

---

### Comment by jylandry 2020-02-11 02:16
I look at it. Some aspect of  #5909  might exhibit the same symptom.

However, their is additional problem with the Building US unit related to the interpretation of the fraction.

a) When you enter -10+1/8" and click OK on the "Placement box"... you get -9+7/8". This might be similar to #5909.

b) When you re-enter the "Placement Box", it does display  -8+1/8".  I don't think it is the same.

---

### Comment by jylandry 2020-02-11 02:17
if you look at the post	https://forum.freecadweb.org/viewtopic.php?f=8&t=42766

[...]

Issue #5909, opened 2022-02-07
Negative numbers with multiple units interpeted as expressions


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=3925

* **Reporter:** metatheorem
* **Date submitted:** 4/4/2019
* **FreeCAD version:** 0.18
* **Category:** Bug
* **Status:** confirmed
* **Tags:** units

# Original report text

When typing in a negative measurement, such as in the Draft Move tool, combinations of units are interpreted as an addition expression.  For example, the expression "-1m 10mm" is evaluated to "-990.0 mm" instead of "-1010.0 mm" as it should be.  This appears to happen regardless of the units used.

Forum disussion here: [Negative numbers in feet/inches interpreted wrong](https://forum.freecadweb.org/viewtopic.php?f=3&t=35401&sid=cf46dba2369d767772ccde09b3448bb8)

# Additional information

This can happen either when a text entry field loses focus or when a tool is applied.

Workarounds:
1. If possible, key in a single type of unit.
2. If not possible, such as when working with feet and inches, perform two operations that add up to the desired measurement.  This may not work for all operations.

OS: Linux
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.18.Unknown
Build type: Unknown
Python version: 2.7.15
Qt version: 4.8.7
Coin version: 3.1.3
OCC version: 6.9.1.oce-0.18
Locale: English/UnitedStates (en_US)

# Steps to reproduce

1. In the Draft workspace, create a two-point rectangle.
2. In any of the point value fields (x, y, or z), type in an expression such as "-1m 10mm".
3. Move keyboard focus away from the field, and the entry changes to "-990.0 mm".

# Other bug information

* **Priority:** normal
* **Severity:** minor
* **Category:** Bug
* **Updated:** 1/23/2022

# Discussion from Mantis ticket

### Comment by Kunda1 2019-04-24 11:19
In the future, please follow our guidelines (mentioned in the yellow banner at the top of the page)

How did you build FreeCAD?
Can you please try with a Qt5/Py3/OCC7.3 build?
https://github.com/FreeCAD/FreeCAD/releases/tag/0.18.1
or even better try the 0.19_pre 
https://github.com/FreeCAD/FreeCAD/releases/tag/0.19_pre

---

### Comment by Geoplace 2019-04-27 03:46
Hi

I've teste the version 18.1

But there is the same behavior.
[...]

PR #9397, opened 2023-04-26
[App] Expression parser allow multi unit inputs (2'2")

ConnorWorrell:

---

Adds support to the expression parser for multi unit inputs:
Ex: 2'2" or 2'2mm as 2'+2" or 2'+2mm

Added step:
1. Number Unit are combined
2. (Number Unit) (Number Unit) are combined (added step)
3. Parse remaining expression

Addresses an edge case in this issue https://github.com/FreeCAD/FreeCAD/issues/8091
Also addresses input in https://github.com/FreeCAD/FreeCAD/issues/6029 and https://github.com/FreeCAD/FreeCAD/issues/5909

Past behavior: 

- Prior to https://github.com/FreeCAD/FreeCAD/pull/7124 -2'2" was evaulated as -2'+2".
- After https://github.com/FreeCAD/FreeCAD/pull/7124 this input failed entirely.

This correctly evaluates both the -2'2" and 2'2" inputs as -(2'+2") and 2'+2".

Adding it to the expression parser is most appropriate to maintain consistency. Previously this was handled only in the quantity spin box, so these inputs would inconsistently work. eg: they did not work in Spreadsheet.

src/App/ExpressionParser.y is the file I made changes to,

GNU Bison regenerated these files, but I'm not sure what to do with them. Let me know if I should do something different with these.
- src/App/ExpressionParser.tab.c
- src/App/ExpressionParser.tab.h


src/App/ExpressionParser.tab.h
```diff
@@ -30,6 +31,10 @@
    This special exception was added by the Free Software Foundation in
    version 2.2 of Bison.  */
 
+/* DO NOT RELY ON FEATURES THAT ARE NOT DOCUMENTED in the manual,
+   especially those whose name start with YY_ or yy_.  They are
+   private implementation details that can be changed or removed.  */
+
```
AjinkyaDahale, line 37: Is this from GNU Bison, or from author of the PR? If former, we are still using these features and may be an issue. If the latter, could you elaborate why you're choosing this nomenclature?

AjinkyaDahale: OK I should have read this:
> GNU Bison regenerated these files, but I'm not sure what to do with them. Let me know if > I should do something different with these.
> * src/App/ExpressionParser.tab.c
> * src/App/ExpressionParser.tab.h

As to the question itself, I see there are items like `#line` directives, which may be very easily messed up if edited by hand by as much as a newline. It's best to let these be generated automatically.

adrianinsaval:
It works, though it allows some very odd inputs like `1in 1"1cm` or `0.5*pi rad 10°` which results in 95°, intuitively I would have expected 100° but I see why it applies the multiplication to the whole thing, it actually changes the expression to `0.5 * (pi rad + 10 °)` if you reopen the expression dialog. I don't think it would be common to use this type of syntax on anything else besides feets and inches though. Would it be complicated to limit this syntax to only those two units? I'm not sure if it's preferred to leave it available for all units or limit it to imperial units.

adrianinsaval:
I notice that spreadsheets still interpret `2'2"` as a string (it works if you use `=2'2"` to make it a expression), it is not necessary to solve this now but it would be a nice follow up PR

@wwmayer could you answer this:

> GNU Bison regenerated these files, but I'm not sure what to do with them. Let me know if I should do something different with these.
> 
> src/App/ExpressionParser.tab.c
> src/App/ExpressionParser.tab.h

ConnorWorrell:
> Would it be complicated to limit this syntax to only those two units? I'm not sure if it's preferred to leave it available for all units or limit it to imperial units.

I agree, I would have expected .5*pi rad + 10 deg.

Restricting this to only feet and inches should be straight forward. I think adding the many US-Building input formats to the expression parser is the path forward.

adrianinsaval:
I'm willing to merge as is or would you prefer to rework this to limit to imperial units?

ConnorWorrell:
Changes made to ExpressionParser.l and ExpressionParser.y, everything else is regenerated.

This limits the #'#" format to only ' and " characters, so there should only be 4 acceptable forms:
* #'#' -> #' + #'
* #'#" -> #' + #"
* #"#' -> #" + #'
* #"#" -> #" + #"

If I think parsing ' and " as interchangeable is ok and doesn't cause any unexpected results. If it does I can change it to only accept #'#".

src/App/ExpressionParser.y
```diff
@@ -130,7 +134,11 @@ cond: exp EQ exp                                { $$ = new OperatorExpression(Do
...
     ;
 
-unit_exp: UNIT                                  { $$ = new UnitExpression(DocumentObject, $1.scaler, $1.unitStr );                }
+us_building_unit: USUNIT                        { $$ = new UnitExpression(DocumentObject, $1.scaler, $1.unitStr );                }
+other_unit: UNIT                                { $$ = new UnitExpression(DocumentObject, $1.scaler, $1.unitStr );                }
+
+unit_exp: other_unit                            { $$ = $1; }
+        | us_building_unit                      { $$ = $1; }
         | unit_exp '/' unit_exp                 { $$ = new OperatorExpression(DocumentObject, $1, OperatorExpression::DIV, $3);   }
         | unit_exp '*' unit_exp                 { $$ = new OperatorExpression(DocumentObject, $1, OperatorExpression::MUL, $3);   }
```
ConnorWorrell, line 143: On a side note, does line 143 work for anyone? I was surprised to see this in here, because I specifically remember this input not working.

I would have expected 1mm*mm to eval to 1mm^2. But it fails.

Looks like it groups as (num * unit_expr)( unit_expr) because of incorrect precedence.
NUM_AND_UNIT has higher precedence than * / and %.

Did this ever work?

adrianinsaval:
sorry misscliked a button

src/App/ExpressionParser.y
adrianinsaval: I'll have to check if this is a regression, but I would expected 1kg*m and similar to work, but currently you need to input 1kg*1m

adrianinsaval:
> If I think parsing ' and " as interchangeable is ok and doesn't cause any unexpected results. If it does I can change it to only accept #'#", this requires tracking ' and " separately, and would be slightly longer.

yeah I think that's ok although it's probably not expected that it works. I'll test this again and merge within the week

src/App/ExpressionParser.y
adrianinsaval: I just found out how to use this on master, `1 (kg*m)` works

PR #9398, opened 2023-04-26
Fixed segmentation fault in PartDesign workbench

### LINKED ISSUE #9382: [Issue] Crash when trying to move object from part to another body

### Forums discussion

_No response_

### Version

0.20 (Release)

### Full version info

```shell
[code]
OS: Ubuntu 20.04.6 LTS (XFCE/xubuntu)
Word size of FreeCAD: 64-bit
Version: 0.20.1.
Build type: Release
Branch: unknown
Hash: 1397818c9664d158281339a84be59d3e9887f6ad
Python 3.8.10, Qt 5.12.8, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.2
Installed mods: 
  * Manipulator 1.4.9
[/code]


Also crash happens in nightly:

[code]
OS: Ubuntu 20.04.6 LTS (XFCE/xubuntu)
Word size of FreeCAD: 64-bit
Version: 0.21.
Build type: Release
Branch: unknown
Hash: 1282fdb7e8a721e9b411358d68daf8921b0bf50a
Python 3.8.10, Qt 5.12.8, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.2
Installed mods: 
  * Manipulator 1.4.9
[/code]
```

### Subproject(s) affected?

None

### Issue description

In the attached diagram I made a mistake, rather than creating a new part with multiple bodies, I created a body, and inside of it created a part, added a shape binder and created a pad using the shapebinder.

Tried to create another body to move the shape binder and the part to when I select the objects "MiniPC_sketch_ref" and "Pad" inside the "Mini_PC_Physical" right click and select "Move object to another body" FreeCAD segfaults with:

[see attached Segfault.zip file]

### Anything else?

[Mini_PC_Holder.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/11320908/Mini_PC_Holder.FCStd.zip)
[Segfault.zip](https://github.com/FreeCAD/FreeCAD/files/11320954/Segfault.zip)

### Code of Conduct

xtemp09:
Closes #9382. [usakhelo](https://github.com/usakhelo), the original author of the function `isFeatureMovable`, politely [refused](https://github.com/FreeCAD/FreeCAD/issues/9382#issuecomment-1522224360) to contribute to FreeCAD.

The cause of segmentation fault was attempted access to nonexistent `App::DocumentObject* const feat` (it was `nullptr`).

I concluded that if there is no object, so it cannot be moved. I can be wrong.

---

<details><summary>Standard form</summary>

</details>

### DIFF
--- a/src/Mod/PartDesign/Gui/Utils.cpp
+++ b/src/Mod/PartDesign/Gui/Utils.cpp
@@ -438,6 +438,9 @@ void relinkToBody (PartDesign::Feature *feature) {
 
 bool isFeatureMovable(App::DocumentObject* const feat)
 {
+    if (!feat)
+        return false;
+
     if (feat->getTypeId().isDerivedFrom(PartDesign::Feature::getClassTypeId())) {
         auto prim = static_cast<PartDesign::Feature*>(feat);
         App::DocumentObject* bf = prim->BaseFeature.getValue();


adrianinsaval:
thanks

0penBrain:
Sorry, a bit late to the party. Patch in the PR is good, but does one look if it's an expected behavior that the function is called with null parameter? I just worry it hides a deeper problem 😉

xtemp09:
I agree, it masks a deeper problem.

wwmayer:
There is indeed a deeper problem. The Pad uses a shape binder as profile instead of a sketch. However, inside isFeatureMovable() it's checked whether the passed object (the Pad) is profile based and if yes its sketch is checked. Because the pad is not based on a sketch the returned object is null. This null pointer is then passed to the recursive call of isFeatureMovable() and caused the crash.

There are two things to fix:
* instead of getVerifiedSketch() the method getVerifiedObject() should be used
* if the returned value is null then isFeatureMovable() shouldn't be called at all because otherwise it will say the pad cannot be moved which isn't necessarily true

xtemp09:
Actually, I replaced:

```c++
if (!feat)
        return false;
```

with

```c++
if (!feat)
        return true;
```

and the pad remained unmovable. It's good that you have found time to take a look at this.

wwmayer:
I wonder about the general usefulness of the Move command because it's not even able to handle simple cases.

1. Create a body
2. Create a 2nd body
3. Create a sketch with a rectangle
4. Create a pad
5. Move the pad to the 1st body. The sketch will be moved, too.
6. Modify the pad and confirm your changes
7. You get the warning: `PartDesign::Body: Link(s) to object(s) 'Pad' go out of the allowed scope 'Body001'. Instead, the linked object(s) reside within 'Body Body'.`

The problem is that the second body still has a reference to the moved pad (Tip property). After manually resetting the Tip in the property editor fixes the issue.

0penBrain:
@wwmayer Do you have an idea how it could be solved or should we create a ticket ?

0penBrain:
@xtemp09 May you take care of the tickets (I have a limited understanding of the exact issues ATM) ?
I think there are 2 :
* Problem with shapebinder-based sketch: https://github.com/FreeCAD/FreeCAD/pull/9398#issuecomment-1536397142
* Problem with remaining link between objects: https://github.com/FreeCAD/FreeCAD/pull/9398#issuecomment-1536452658

Issue #6992, opened 2022-01-27
Supporting Qt6 / PySide6 / Shiboken6


luzpaz:
This ticket tracks the Qt5 :arrow_right: Qt6 migration progress. See also dedicated [forum thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=65089)   

## Commits 

Qt6 commits to master: 
| commit | summary | committer |
| --------- | ------------- | ------------- |
| 339bb73e9f | Base: add convenience classes/functions that are marked as deprecated in Qt 5.15 and removed in Qt 6  | wwmayer | 

## Notes

* [Porting to Qt 6](https://doc.qt.io/qt-6/portingguide.html)
* [Porting from PySide2 to PySide6](https://doc.qt.io/qtforpython/porting_from2.html)
* A Clazy-based tool is available to facilitate porting from Qt 5 to Qt 6: [Porting C++ Applications to Qt 6 using Clazy checks](https://doc.qt.io/qt-6/porting-to-qt6-using-clazy.html). #7134

### Module level changes
- [ ] The content of the *QtOpenGL* module has been replaced. The class `QGLWidget` and related classes (`QGLContext`, `QGLFunctions`, `QGLShaderProgram`) have been removed. Parts of the *Open GL* functionality from *QtGui* have been extracted into this module, for example `QOpenGLBuffer` and `QOpenGLShaderProgram`. There is a new module *QtOpenGLWidgets* which contains the class `QOpenGLWidget`, a replacement for `QGLWidget`.

## Imports
- [ ] Change PySide2 -> PySide6
Ex. 
	```python
	from PySide2.QtWidgets import QApplication...
	from PySide2 import QtCore
	```
    to 
	```python
	from PySide6.QtWidgets import QApplication...
    from PySide6 import QtCore
	```
- [ ] Some classes are in a different module now, for example `QAction` and `QShortcut` have been moved from `QtWidgets` to `QtGui`.

### Class/Function Deprecations
- [ ] The High DPI scaling attributes `Qt.AA_EnableHighDpiScaling`, `Qt.AA_DisableHighDpiScaling` and `Qt.AA_UseHighDpiPixmaps` are **deprecated**. High DPI is by default enabled in Qt 6 and cannot be turned off.
- [ ] `QDesktopWidget` has been **removed**. `QScreen` should be used instead, which can be retrieved using `QWidget.screen()`, `QGuiApplication.primaryScreen()` or `QGuiApplication.screens()`.
[...]

PR #9404, opened 2023-04-26
[Bugfix]Plot: reintroduce patch removed by #9103

0penBrain:
While FC don't depend on PyQt anymore, patch is needed to have
 matplotlib working correctly

 https://github.com/FreeCAD/FreeCAD/commit/876601d34d7596fd61041263a9c56b611fd04777#r108628060

### DIFF
--- a/src/Mod/Plot/Plot.py
+++ b/src/Mod/Plot/Plot.py
@@ -30,7 +30,13 @@ try:
     import matplotlib
     matplotlib.use('Qt5Agg')
 
-    import matplotlib.pyplot as plt
+    # Force matplotlib to use PySide backend by temporarily unloading PyQt
+    if 'PyQt5.QtCore' in sys.modules:
+        del sys.modules['PyQt5.QtCore']
+        import matplotlib.pyplot as plt
+        import PyQt5.QtCore
+    else:
+        import matplotlib.pyplot as plt
 
     from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg as FigureCanvas
     from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT as NavigationToolbar


luzpaz:
Relevant thread: https://forum.freecad.org/viewtopic.php?t=77693

adrianinsaval:
I have no problem putting this back in at all, but just to explore alternatives: https://matplotlib.org/stable/api/backend_qt_api.html#qt-bindings perhaps we could set the QT_API env variable instead? We would have to have a check for qt5 or qt6 though

0penBrain:
> I have no problem putting this back in at all, but just to explore alternatives: https://matplotlib.org/stable/api/backend_qt_api.html#qt-bindings perhaps we could set the QT_API env variable instead? We would have to have a check for qt5 or qt6 though

I'm afraid rule 1 will apply before rule 2 and so this won't change anything. This said, this is pretty easy to test. Just I don't know a way to reproduce the issue. :confused:

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9404">pipeline status</a> for feature branch [PR_9404](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9404). Pipeline [903674379](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903674379) was triggered at [0a23f9b](https://github.com/FreeCAD/FreeCAD/pull/9404/commits/0a23f9b42c02ee2609dd63b99f9b0e8bde4a237f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
I need to check this against Qt6.

luzpaz:
Linking to #6992

adrianinsaval:
if this is also required for qt6 it would be an additional conditional, this PR won't make it worse for qt6 so I'm merging it

Issue #8426, opened 2023-02-11
Building US unit schema conflicts with expression parser, breaks when reevaluated for negative numbers


pcoe1959:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.20 (Release)

### Full version info

```shell
0.20.2
29177 +426 (Git)
2022/12/07 02:35:37
Ubuntu Core 20
64-bit
tag: 0.20.2
930dd9a
```

### Subproject(s) affected?

Part

### Issue description

My units are set to Building US (ft-in/sqft/cft). Create a cube in part and try to position it via the cubes properties->Placement->Position, x y and z are set to 0 change any one of them and set its  value to a negative to move it below its axis, leave the field and try to adjust it, it will change to something other than what it was when you first edited it forcing you to renter the entire number even though no edit has occurred yet, it appears that the fields On_Enter handler has gone amok. It is really frustrating when trying to fine tune positioning. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
Please add an example file

adrianinsaval:
To reproduce set unit schema to Building US and on almost any input field (say a Part Cube's Position x) input for example -1.5" and hit enter this will then be displayed as -1+1/2" when clicking again to edit this field this will be evaluated by the parser and change to -1/2" instead

adrianinsaval:
@ConnorWorrell may you have a look into this one too? I'm not sure if this is possible to solve on the parser since it is correct to interpret -1+1/2 as -1/2, maybe we can make the Building US scheme use parenthesis when dealing with negative numbers?

ConnorWorrell:
I see a few solutions,

1. Handle the division in the expression parser -1'1/2", this seems not great because there is no easy way to get a dimension on the bottom of the fraction, ex: 1in*1/2in -> .5in^2 instead of .5 unitless the current expected result.
2. Parentheses -(1+1/2) , this is straight forward, but since this change would happen in the scheme these parentheses would show up everywhere, and it annoys me because they look just like reference dimensions. This might also lead to bad inputs, I often edit a box and type +1" to add a bit of length to something, but here the result would be -(1+1+1/2) which is the opposite of what was intended.
3. Carry the negative through -1-1/2, this seems like the cleanest.

adrianinsaval:
fixed by #9408

PR #9408, opened 2023-04-27
[Base] US Building Schema correct sign on fraction

ConnorWorrell:

---

This makes a few changes to how US Building Scheme displays/calculates units:
1. Sign is carried through multiple units: -1'2.5" -> -1'2" - 1/2" https://github.com/FreeCAD/FreeCAD/issues/8426
2. Inches is notated with a quote to play better with the expression parser/spinbox 1'2 -> 1'2" and 2 -> 2"
3. A fraction is always preceded by a plus or minus to play better with the expression parser 1'1/2" -> 1'+1/2"

This relies on this PR to handle the #'#" format units US Building creates. https://github.com/FreeCAD/FreeCAD/pull/9397

### DIFF
--- a/src/Base/UnitsSchemaImperial1.cpp
+++ b/src/Base/UnitsSchemaImperial1.cpp
@@ -274,35 +274,32 @@ QString UnitsSchemaImperialBuilding::schemaTranslate(const Quantity &quant, doub
 
         // Process into string. Start with negative sign if quantity is less
         // than zero
+        char plusOrMinus;
         if( quant.getValue() < 0 )
+        {
             output << "-";
+            plusOrMinus = '-';
+        }
+        else plusOrMinus = '+';
 
+        bool trailingNumber = false;
         // Print feet if we have any
         if( feet!=0 )
         {
             output << feet << "'";
-
-            // if there is to be trailing numbers, add space
-            if( inches!=0 || num!=0 )
-            {
-                output << " ";
-            }
+            trailingNumber = true;
         }
-
-        // Three cases:
-        //   1. Whole inches, no fraction
-        //   2. Whole inches, fraction
-        //   3. Fraction only
-        if( inches>0 && num==0 ) // case 1.
+        // Print whole inches if we have any
+        if( inches!=0 )
         {
+            if (trailingNumber) output << " ";
             output << inches << "\"";
+            trailingNumber = true;
         }
-        else if( inches>0 && num!=0 ) // case 2
+        // Print fractional inches if we have any
+        if( num!=0 )
         {
-            output << inches << "+" << num << "/" << den << "\"";
-        }
-        else if( inches==0 && num!=0 ) // case 3
-        {
+            if (trailingNumber) output << " " << plusOrMinus << " ";
             output << num << "/" << den << "\"";
         }
 


adrianinsaval:
many thanks, looks good to me but I'll wait for some input from others, I'm not familiar with how US people expect their lengths displayed

yorikvanhavre:
I am OK with this, just a note:

> A fraction is always preceded by a plus or minus to play better with the expression parser 1'1/2" -> 1'+1/2"

People in the building industry specifically use 1'1/2" notations. I think we can work with this internally, though (is it not already so BTW?)
* As long as people are still able to insert or copy/paste 1'1/2" notations into the field, I think it's acceptable if it's then converted to 1'+1/2"
* Draft dimensions will show 1'+1/2" as 1'1/2" ( https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Draft/draftviewproviders/view_dimension.py#L577 ) We should check how it works in TechDraw, maybe worth doing the same there...

ConnorWorrell:
The copy paste of this doesn't work 1'1/2", the parser splits the numbers as such: (1'+1)/(2"), and gets upset about the foot + unitless conversion.

I found this snipped from The United States National CAD Standard V4 notes the format as #'-#-#/#". Anyone know if there are any other standards on this? 


Unfortunately, neither #'#-#/#" nor #'-#-#/#" play well with the parser. The minus signs cause a lot of issues.

Maybe the best path is to have the parser treat these as units? But this could lead to some undesirable behavior, ex: subtracting any fractional value will add it instead.
* #'-#-#/#"
* #'#-#/#"
* #-#/#"
* #'-#"
* #'#"

yorikvanhavre:
> Anyone know if there are any other standards on this?

@theoryshaw you're needed here!

adrianinsaval:
> 3\. A fraction is always preceded by a plus or minus to play better with the expression parser 1'1/2" -> 1'+1/2"

can you describe better what is the problem with 1'1/2"?

theoryshaw:
Not sure about standards, but the following is from Revit, so gives you a little sense of what large part of the industry is used to...

https://user-images.githubusercontent.com/507113/236915047-1055a2e5-ba8a-4961-acc4-61a37abafc17.mp4

ConnorWorrell:
> > 3\. A fraction is always preceded by a plus or minus to play better with the expression parser 1'1/2" -> 1'+1/2"
> 
> can you describe better what is the problem with 1'1/2"?

Currently the parser can't handle the 1'1 part of that. 

If patching in #9397, it has trouble with the units on the 1' + 1 unitless operation. I expect ignoring units it would end up with (1'1)/2 which is also unexpected.

It looks like Autodesk likes the X'-x x/x" format, which the parser could support. 

The x x x/x format seems ambiguous, but it could be supported in the parser or spin box, probably better in the spin box.

The 2'-6" format would cause issues, that would definitely need to happen in spin box. 

I've been wondering @theoryshaw how does AutoCAD handle numbers like 2.5222? Currently freecad would round to 2.5 when editing the dimensions. It seems to me that this is not a good behavior.

theoryshaw:
https://github.com/FreeCAD/FreeCAD/assets/507113/a39b9e59-4645-4473-bcc2-62c90574c24d

ConnorWorrell:
Interesting, it seems like AutoCAD also has this rounding on edit behavior, so I guess it must not be that big an issue. It still feels odd to display 2.5222 as 2'-6 1/4" (2.5208333'), but internally store the 2.52222 until the user edits the dimension, where its then restored as 2.5208333', decreasing the length by ~1/8", even though the user didn't directly edit the dimension.

@adrianinsaval how should this proceed? Should we create a feature request for bringing US-Building up to standard, and continue this discussion there? These changes will need PRs spanning multiple modules.

yorikvanhavre:
I agree storing 2'-6" could be catastrophically misleading :)

I think we need here to split what we represent internally in FreeCAD UI and what we can do on dimensions that appear on our documents. The internal/UI representation needs to be unequivocal.

I think the 2'+6" representation (or -2'-6") is good because it is logical mathematically. That's what we should use.

We should still permit that someone copy/pastes in values like 2'6" (which should be OK) or -2'6" (which should maybe be internally converted to -2'-6"?)

Someone inserting 2' - 6" is unclear... I'm not sure what we should do in such cases?

Now, on objects such as Draft dimensions, we could easily take the internal value and represent it as 2' - 6". or maybe use here a long dash? 2' — 6". That would make it unequivocal...

yorikvanhavre:
> that's a minus sign all over the world, using it for anything else in the context of measurements is beyond stupid IMO

It is a standard in a big part of the world, people don't have a choice to use it or not, we can't be calling that stupid.

adrianinsaval:
> It is a standard in a big part of the world, people don't have a choice to use it or not, we can't be calling that stupid.

I understand supporting it for _display_ in technical drawings or 3d annotations, but on input fields it's too big of a clash with the minus sign and math is more important than these odd notations.

yorikvanhavre:
Yes, that's what i am saying too.
Anyway, this is for another discussion i guess.

adrianinsaval:
Many thanks!

Roy-043:
There is an issue that seems to be related to the PR:
https://github.com/FreeCAD/FreeCAD/issues/11345#issuecomment-1808135862

PR #9413, opened 2023-04-27
Sheet: display an error in console when a cyclic dependency is detected

0penBrain:
https://forum.freecad.org/viewtopic.php?t=77448

### DIFF
--- a/src/Mod/Spreadsheet/App/Sheet.cpp
+++ b/src/Mod/Spreadsheet/App/Sheet.cpp
@@ -1005,8 +1005,9 @@ DocumentObjectExecReturn *Sheet::execute(void)
             std::list<Vertex> make_order;
             try {
                 boost::topological_sort(graph, std::front_inserter(make_order));
-            } catch (std::exception&) {
+            } catch (std::exception&) { //TODO: evaluate using a more specific exception (not_a_dag)
                 // Cycle detected; flag all with errors
+                Base::Console().Error("Cyclic dependency detected in spreadsheet : %s\n", *pcNameInDocument);
                 std::ostringstream ss;
                 ss << "Cyclic dependency";
                 int count = 0;


src/Mod/Spreadsheet/App/Sheet.cpp
```diff
@@ -1007,6 +1007,7 @@ DocumentObjectExecReturn *Sheet::execute(void)
                 boost::topological_sort(graph, std::front_inserter(make_order));
             } catch (std::exception&) {
```
AjinkyaDahale, line 1008: Could be specific, for e.g. [`not_a_dag`](https://www.boost.org/doc/libs/1_82_0/libs/graph/doc/exception.html#not_a_dag).

luzpaz:
An aside, for clarity purposes would it make sense to write out the full word 'spreadsheet' instead of 'sheet'. I'm also thinking about translators and wondering if 'sheet' is perhaps ambiguous in other languages ?

adrianinsaval:
> An aside, for clarity purposes would it make sense to write out the full word 'spreadsheet' instead of 'sheet'. I'm also thinking about translators and wondering if 'sheet' is perhaps ambiguous in other languages ?

I agree, sheet could be confused with something from sheet metal wb too when looking at the report view later on

yorikvanhavre:
+1 on the two proposed changes: Use a specific exception and sheet -> spreadsheet

0penBrain:
OK guys. I will over the game now. :) Was a funny one.

I think we can leverage this PR to question how we (as maintainers) deal with PR:
 * The remark from @AjinkyaDahale seconded by @yorikvanhavre isn't part of the changeset of this PR. So does it need to be addressed here ? I have no strong opinion. On one side, there is a risk with this PR that we have a message "cyclic dependency" while another type of exception was raised. But on the other side it's very low as only `topological_sort` is called in the `try` block -- we clearly use the exception as a function to detect cyclic dependency issue --. And is it worse than having a lone laconic "Recompute failed, check report view" ?
 * Does writing "spreadsheet" instead of "sheet" makes this PR unmergeable ? Again no strong opinion from me. But if yes, it should be clearly expressed. If no, well ... someone should have merged this and push the change in another PR.

Also remember than when a maintainer authors a PR, and according our guidelines, she/he is a contributor like others that should not merge her/his own PR. ;)

I'll do the text change and force-push. ;)

AjinkyaDahale:
>     * ...On one side, there is a risk with this PR that we have a message "cyclic dependency" while another type of exception was raised. But on the other side it's very low as only `topological_sort` is called in the `try` block -- we clearly use the exception as a function to detect cyclic dependency issue --. And is it worse than having a lone laconic "Recompute failed, check report view" ?

The title of the commit can be changed, for sure, if you feel it's an issue.

I do see that topological sort only gives this one exception, so we can assume any exception is this. One way I can see this assumption being broken is that someone in the future adds more lines to the `try` block (say an `assert(length < something)`, though I don't know if it gets caught). In that case, the user (most likely a sophomoric developer like me) is confused and furiously finding the cycle. To be on the safe side, I believe it's best to use the specific exception and then if need be another general "unhandled exception occurred".

As for the `Recompute failed, check report view` (often the only new content in the report view as well), that's for sure another issue. But so long as the recompute is failing, at least it is technically correct :smiley: (unless it isn't when we should be fixing that too).

yorikvanhavre:
Thanks for the fix! I have no strong opinion on the other fix, indeed it's not part of the intent of this PR, and it requires extra testing too. For me it's fine to merge this as it is now.

Maybe we could just add a TODO comment in the code saying that a better exception should be used here, so it's easy to find later on?

yorikvanhavre:
Thanks!

PR #9418, opened 2023-04-27
PATH: Avoid unnecessary mid-Op tool movements to origin: don't add the end …

jimzim111:
…vector of 0-length segments to the Op move list.

When an edge is evaluated to produce a path segment for tool movement, but no movement results, Path.fromShapes returns a default value (0,0,0) and the tool position remains unchanged.  So we shouldn't update the current end_vector to that (0,0,0).  (Otherwise this causes the next segment to start at (0,0,0) instead of the actual tool position.

The bug is caused when a face being profiled contains any holes that are too small for the tool to fit.  Those holes get checked to see if they should produce any gcode.  They don't produce gcode since the tool can't fit.  But the (0,0,0) still gets saved in self.end_vector which leads to the tool moving there before moving to the next hole where it does fit and actually does cut.

For example screenshots, see https://forum.freecad.org/viewtopic.php?p=666869#p666869.  (That forum post refers to a different PR which got more changes intermingled so I closed it.  This PR is just to address this one bug)

---


mlampert (approved):

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9418">pipeline status</a> for feature branch [PR_9418](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9418). Pipeline [903673408](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903673408) was triggered at [0c7ef7d](https://github.com/FreeCAD/FreeCAD/pull/9418/commits/0c7ef7d4c8aa61cea2a0257d10414aebf460266f). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
@sliptonic do you approve of this PR and what's the target for it (v0.22 ?)

PR #9430, opened 2023-04-28
Begin translating user-visible exception strings

chennes:
In many cases we directly display exception strings to users in dialog boxes explaining the cause of a failure. This PR begins the process of allowing translating those strings. This is a two-step process: when displaying the exception text in the dialog box, the code that displays that dialog must ask Qt to attempt to translate that string using `QApplication::translate` with the context "Exception". Second, when creating the text for those exceptions, developers must use the `QT_TRANSLATE_NOOP()` macro with the "Exception" context to ensure the text is added to the TS file for translation.

Fixes https://github.com/FreeCAD/FreeCAD-translations/issues/203 (though there are certainly many more exceptions that must be tagged -- these will be addressed in follow-on PRs).

---

### DIFF
--- a/src/Mod/Part/Gui/DlgExtrusion.cpp
+++ b/src/Mod/Part/Gui/DlgExtrusion.cpp
@@ -495,8 +495,10 @@ void DlgExtrusion::apply()
         throw;
     }
     catch (Base::Exception &err){
-        QMessageBox::critical(this, windowTitle(),
-            tr("Creating Extrusion failed.\n%1").arg(QString::fromUtf8(err.what())));
+        QMessageBox::critical(this,
+                              windowTitle(),
+                              tr("Creating Extrusion failed.\n%1")
+                                  .arg(QCoreApplication::translate("Exception", err.what())));
         return;
     }
     catch(...) {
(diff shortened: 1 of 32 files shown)


chennes:
@wwmayer is this what you had in mind for translating these user-visible strings?

yorikvanhavre:
I think this is pretty much what is done in other parts of FreeCAD...

But won't most error messages either come from other libs (OCCT,...) or contain some parts that are different each time like memory addresses or python line number?

wwmayer:
> is this what you had in mind for translating these user-visible strings?

Yes, exactly what I had in mind. For the  `QApplication::translate("Exception", e.what())` we could think about adding a convenience function.

The next step would be to add the sources files of App modules to the translation system. Here the question is if we should add a Resources/translations structure also to App modules or should rather the script that updates the .ts files be extended to search in the App directories. The second option is probably faster and easier to implement.

chennes:
> But won't most error messages either come from other libs (OCCT,...) or contain some parts that are different each time like memory addresses or python line number?

The half-dozen or so that translators specifically called out in GitHub issues all came from our code: I'm sure there are times when we show OCCT exception text to users, but over time we can catch those and rethrow translated text instead.

PR #9441, opened 2023-04-29
[TechDraw] Remove BaseGeom

benj5378:

---
https://forum.freecad.org/viewtopic.php?t=77616


benj5378:
Can only say, that this is very labour intensive
Screenshot from 2023-04-29 21-37-57

benj5378:
@WandererFan Are you still sure this is a good idea? Please look through and let me know if this was what were in your thoughts. It's only a draft.

WandererFan:
Still a good idea.  I was looking at your original commit and moved the geometry->QPainterPath stuff to a separate file: https://github.com/WandererFan/FreeCAD/tree/PR9441 

Still an issue with face outlines.  

I will look at your newest commit tomorrow. We probably want to do this in small chunks instead of one big PR.

There will be a feature freeze coming soon, so this likely won't make it into the v0.21 release.

benj5378:
> I will look at your newest commit tomorrow. We probably want to do this in small chunks instead of one big PR.

This probably won't be possible because it is so intervened. I can split up VertexGeom, FaceGeom and EdgeGeom. But I can't split up EdgeGeom by itself. Though it is very far from over and I might move and merge the code around A LOT.

benj5378:
@WandererFan I think this PR is beginning to get over our heads. I have used about 12 hours now fixing compiler error and I can only imagine at least as many fixing bugs. I suggest that we start skinnying down BaseGeom before removing it completely.

WandererFan:
OK.  Bin this and we'll move to Plan B.

PR #9443, opened 2023-04-30
Sketcher: element arc icons have start and end point switched

Roy-043:

---


AjinkyaDahale:
Looks like a straightforward change. However,
1. Please squash both commits into a single one, and in commit body explain why this change is needed. I assume it is to keep it ik line with how angles are measured counterclockwise.
2. Please make sure the svgs are plain and do not have the inkscape metadata.

Roy-043:
I have tried to squash.

Converting to plain SVG should be done in a separate PR for all Sketcher SVG files. Doing only these files would not be consistent.

Let's merge...

AjinkyaDahale:
Thanks for squashing. SVGs are already plain AFAIK so we don't want to add inkscape svgs in new merges. Yours already seems to be plain though so that's OK.

AjinkyaDahale (approved):

PR #9445, opened 2023-04-30
Draft: use convert_draft_texts in importDXF.py

Roy-043:
Replace `convertDraftTexts` with `convert_draft_texts` to prevent a warning.


yorikvanhavre:
Good catch :)

PR #9450, opened 2023-05-01
Fix black image

wwmayer:

---

### DIFF
--- a/src/Gui/View3DInventorViewer.cpp
+++ b/src/Gui/View3DInventorViewer.cpp
@@ -287,6 +287,42 @@ public:
  *
  * \section overview Overview
  * \todo Overview and complements for the 3D Viewer
+ *
+ * \section trouble Troubleshooting
+ * When it's needed to capture OpenGL function calls then the utility apitrace
+ * can be very useful: https://github.com/apitrace/apitrace/blob/master/docs/USAGE.markdown
+ *
+ * To better locate the problematic code it's possible to add custom log messages.
+ * For the prerequisites check:
+ * https://github.com/apitrace/apitrace/blob/master/docs/USAGE.markdown#
+ * emitting-annotations-to-the-trace
+ * \code
+ * #include <GL/glext.h>
+ * #include <Inventor/C/glue/gl.h>
+ *
+ * void GLRender(SoGLRenderAction* glra)
+ * {
+ *     int context = glra->getCacheContext();
+ *     const cc_glglue * glue = cc_glglue_instance(context);
+ *
+ *     PFNGLPUSHDEBUGGROUPPROC glPushDebugGroup = (PFNGLPUSHDEBUGGROUPPROC)
+ *     cc_glglue_getprocaddress(glue, "glPushDebugGroup");
+ *     PFNGLDEBUGMESSAGEINSERTARBPROC glDebugMessageInsert = (PFNGLDEBUGMESSAGEINSERTARBPROC)
+ *     cc_glglue_getprocaddress(glue, "glDebugMessageInsert");
+ *     PFNGLPOPDEBUGGROUPPROC glPopDebugGroup = (PFNGLPOPDEBUGGROUPPROC)
+ *     cc_glglue_getprocaddress(glue, "glPopDebugGroup");
+ *
+ *     glPushDebugGroup(GL_DEBUG_SOURCE_APPLICATION, 0, -1, __FUNCTION__);
+ * ...
+ *     glDebugMessageInsert(GL_DEBUG_SOURCE_APPLICATION, GL_DEBUG_TYPE_OTHER,
+ *                          0, GL_DEBUG_SEVERITY_MEDIUM, -1, "begin_blabla");
+ * ...
+ *     glDebugMessageInsert(GL_DEBUG_SOURCE_APPLICATION, GL_DEBUG_TYPE_OTHER,
+ *                          0, GL_DEBUG_SEVERITY_MEDIUM, -1, "end_blabla");
+ * ...
+ *     glPopDebugGroup();
+ * }
+ * \endcode
  */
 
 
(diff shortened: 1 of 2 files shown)


Jolbas:
Can you provide some information about the system where the black image plane occurs?
I'm thinking since navigation cube could heal this issue, the cause should be between `inherited::actualRedraw();` and `naviCube->drawNaviCube();` in `renderScene()`
Does the problem go away when using box selection?
When disable/enable axis cross?
Activating ENABLE_GL_DEPTH_RANGE in code?
Commenting away something else between the lines above?

wwmayer:
> Can you provide some information about the system where the black image plane occurs?

System: Xubuntu 22.04
Graphics card: GeForce GTX 680MX
Graphics driver: nouveau

> Does the problem go away when using box selection?

Yes.

> When disable/enable axis cross?

Using View > Toggle axis cross fixes it. Alternatively, adding any kind of geometry will fix it, too.
Using the small axis cross in the bottom right corner doesn't make a difference.

> Activating ENABLE_GL_DEPTH_RANGE in code?

No. This doesn't make a difference.

In the meantime I found another problem. When I create e.g. a mesh box then it's also completely black until I hover it in the 3d view. Creating a part box instead is rendered correctly. Luckily, this PR fixes this issue, too.

adrianinsaval:
Maybe unrelated but it seems similar, when the axis cross is enabled and you use box selection on an empty document the axis cross gets darker and stays like that until you add something. Maybe this PR fixes that too? Can't test right now

wwmayer:
> Maybe unrelated but it seems similar

I can confirm your observation but it's not directly related and this PR doesn't fix it either.

The axis cross gets darker because the light model has changed for some reasons. When creating the axis cross it has no shading (SoLightModel::BASE_COLOR) and after the selection it has a shading (SoLightModel::PHONG) and that's why it appears darker. After adding an object the shading will be reset.

wwmayer:
About the axis cross I have an AppImage of v0.17 and there the shading doesn't change whereas in v0.19 it changes. The AppImage of v0.18 doesn't work so I cannot test its behaviour.

wwmayer:
When I set the corners to (0,1),  (0,0), (1,1), (1,0) then it only draws the top right quarter of the background. The image plane is correctly rendered but there is still an overlap of the image plane and the part of the background.

When I choose a smaller part that doesn't overlap with the image plane (like (0.9,1),  (0.9,0.9), (1,1), (1,0.9)) then it's still rendered correctly. So, it doesn't matter whether the image plane overlaps with the part of the background or not.

Jolbas:
Yes. I suspected that. I just wanted the square to be drawn and still be able to see at least a part of the image if its drawn after the image plane. If you then move the `drawSingleBackground(col);` to directly after the `inherited::actualRedraw();` does it still work as a healer?

wwmayer:
> If you then move the drawSingleBackground(col); to directly after the inherited::actualRedraw(); does it still work as a healer?

I can move it but then it completely overrides the image. If I again draw a smaller area then the non-overlapping part of the image plane is still rendered correctly.

wwmayer:
> Maybe unrelated but it seems similar

As said above this is caused by enabling the lighting. When using the rubber band then before starting the actual selection the 3d view is rendered into a framebuffer object and converted into an image. Then while doing the selection the image is rendered with renderGLImage(). When the selection is done the viewer switches back to normal rendering but because renderGLImage() enables GL_LIGHTING the axis cross is rendered in shaded mode.

As there is actually no need to do that the line can be removed which fixes this issue.

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/9463

Jolbas:
I found that if i drew the rectangle without disabling GL_LIGHTING it was drawn black and simillar to your image it was fixed with View > Toggle axis cross. So I wonder. If you replace `drawSingleBackground(col);` with a single simple `glDisable(GL_LIGHTING);` Does that fix the black image issue equally good as your background function?

wwmayer:
> I managed to provoke my image plane to be black by playing with the light. Seems that sometimes the light is set to come from the top instead of from camera. Is that what happen when you get a black image? Or is it black even if you lean the image to let light from another direction fall on its surface.

The image will be black from all sides. It's like if the image plane doesn't emit a color (or black).

> If you replace drawSingleBackground(col); with a single simple glDisable(GL_LIGHTING); Does that fix the black image issue equally good as your background function?

Wow! This indeed fixes the problem with the image plane. However, explicitly disabling the lighting in renderScene() again causes a problem with meshes because it's rendered in a monotone grey color where you cannot observe any contours (i.e. no shading).

Btw:
With PR #9463 I have observed the problem that if a mesh is in the 3D view then it will be rendered with no shading after doing the box selection because it disabled the lighting. So, this requires a few more changes where we have to use glPushAttrib/glPopAttrib calls inside renderGLImage(), renderFramebuffer(), ...

Jolbas:
It seems then that it's the `inherited::actualRedraw();` that somehow makes a mess of the lighting. I suspect a `glPushAttrib(GL_LIGHTING_BIT);` right before it and a `glPopAttrib();` right after would reset the light settings and fix this issue too.

wwmayer:
inherited::actualRedraw() calls QuarterWidet::actualRedraw which calls: https://github.com/coin3d/coin/blob/master/src/rendering/SoRenderManager.cpp#L572
Coin3D doesn't set or reset the lighting there. Actually there are only a very few classes the (re)sets the lighting and I cannot see that this is somewhere messed up: https://github.com/search?q=repo%3Acoin3d%2Fcoin+gl_lighting&type=code

Jolbas:
Mess was a maybe a strong word. But if i surround actualRedraw with log messages like this:

```
        Base::Console().Message("Before: %i\n",glIsEnabled(GL_LIGHTING));
        inherited::actualRedraw();
        Base::Console().Message("After: %i\n",glIsEnabled(GL_LIGHTING));
```
I see that it is changing depending on if for example axis cross is enabled or not or if there is another object than an image plane.

wwmayer:
> I see that it is changing depending on if for example axis cross is enabled or not or if there is another object than an image plane.

I see. I guess it's the SoGLLazyElement class that sets the lighting attribute and is not supposed to unset it after the rendering.

PR #9464, opened 2023-05-03
Draft: Add l10n context for ambiguous string

### LINKED ISSUE #9406: Draft: Clarify specific translated string in `draftguitools/gui_selectplane.py` so that it's less ambiguous
https://github.com/FreeCAD/FreeCAD/blob/25157275f07075ccd40873837f9b0f48aef28c5a/src/Mod/Draft/draftguitools/gui_selectplane.py#L523

Most likely in context meant to be 'Direction'. But some translators might confuse with 'Directory'.

luzpaz:
Also tweaked code formatting  
Closes #9406

---

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_selectplane.py
+++ b/src/Mod/Draft/draftguitools/gui_selectplane.py
@@ -519,7 +519,7 @@ class Draft_SelectPlane:
         vdir += str(_vdir.x)[:4] + ','
         vdir += str(_vdir.y)[:4] + ','
         vdir += str(_vdir.z)[:4] + ')' + ' '
-        vdir *= translate("draft", "Dir", "Dir here means Direction, not Directory. Also shorten the translation because of available space in GUI")
+        vdir += translate("draft", "Dir", "Dir here means Direction, not Directory. Also shorten the translation because of available space in GUI")
         vdir = ': ' + vdir
         if type(arg).__name__ == 'str':
             self.wpButton.setText(arg + suffix)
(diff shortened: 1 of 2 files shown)


chennes:
Draft uses a custom version of translate: so you actually need to jump through another hoop to make this work, and add a default parameter to it in draftutils/translate.py

In that file line 46 should read:
```
def translate(context, text, comment=None):
```

And line 82 should read
```
    return Qtranslate(context, text, comment)
```

chennes:
(Well, or we could start stripping out the custom version an use `translate = FreeCAD.Qt.translate` instead.

luzpaz:
> (Well, or we could start stripping out the custom version and use `translate = FreeCAD.Qt.translate` instead.)

Lets do this so there is uniformity. Therefore I'll put this PR on the backburner

src/Mod/Draft/draftguitools/gui_selectplane.py
```diff
@@ -518,9 +518,9 @@ def display(self, arg, icon=None):
         vdir = '('
         vdir += str(_vdir.x)[:4] + ','
         vdir += str(_vdir.y)[:4] + ','
-        vdir += str(_vdir.z)[:4]
-        vdir += ')'
-        vdir = " " + translate("draft", "Dir") + ": " + vdir
+        vdir += str(_vdir.z)[:4] + ')' + ' '
+        vdir *= translate("draft", "Dir", "Dir here means Direction, not Directory. Also shorten the translation because of available space in GUI")
```
yorikvanhavre, line 522: Are you sure the *= will work here? AFAIR you can only multiply a string by a number....

yorikvanhavre:
> Important question: Does modifying the translate() function, 'reset' all the translations for Draft WB?

No, it has no impact. The important part for the strings to go to crowdin is that lupdate finds "translate(context,text)" in  the code, no matter how that translate() works internally

luzpaz:
> > (Well, or we could start stripping out the custom version and use `translate = FreeCAD.Qt.translate` instead.)
> 
> Lets do this so there is uniformity. Therefore I'll put this PR on the backburner

Lets just merge this PR and open an issue to do the above in a different ticket, @chennes  ?

PR #9490, opened 2023-05-05
Tools: Add exclusion list to lupdate calls

chennes:
The reason Path translations were not getting updated was because one of the files in Path was causing lupdate to hang. As a workaround, this PR adds an "excluded files" list. None of the files in the current list have any translatable strings in them, so this does not affect the UI.

---


sliptonic:
ok

Issue #9265, opened 2023-04-14
[Problem] Memory leak in TechDraw


xtemp09:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
OS: Ubuntu 22.04.2 LTS (ubuntu:GNOME/ubuntu)
Word size of FreeCAD: 64-bit
Version: 0.21.0.14555 (Git shallow)
Build type: Debug
Branch: master
Hash: 5f0b58449d674aadeadaba5b3f1634306eece8af
Python 3.10.6, Qt 5.15.3, Coin 4.0.0, Vtk , OCC 7.8.0.dev
Locale: English/United States (en_US)
```

### Subproject(s) affected?

Techdraw

### Problem description

I compiled FreeCAD with ASAN, started TechDraw, added a cosmetic line. ASAN reported:
```ruby
Direct leak of 576 byte(s) in 2 object(s) allocated from:
    #0 0x7f269fe12a92 in operator new(unsigned long) ../../../../src/libsanitizer/lsan/lsan_interceptors.cpp:248
    #1 0x7f265608a26a in TechDraw::CosmeticExtension::addCosmeticEdge[abi:cxx11](Base::Vector3<double>, Base::Vector3<double>)
FreeCAD/src/Mod/TechDraw/App/CosmeticExtension.cpp:155
    #2 0x7f26564d12f8 in TechDrawGui::TaskCosmeticLine::createCosmeticLine()
FreeCAD/src/Mod/TechDraw/Gui/TaskCosmeticLine.cpp:192
```
I looked at the code. Cosmetic lines are created using new, but memory is never deleted. For example, cosmetic edges are “_cleared_”: [src/Mod/TechDraw/App/DrawViewPart.cpp#L1494](https://github.com/FreeCAD/FreeCAD/blob/3977330012fe58f9ca3c42a0510725ed470acead/src/Mod/TechDraw/App/DrawViewPart.cpp#L1494)

```c++
void DrawViewPart::clearCosmeticEdges()
{
    std::vector<CosmeticEdge*> noEdges;
    CosmeticEdges.setValues(noEdges);
}
```

[...]

PR #9499, opened 2023-05-07
[TechDraw] Merge CosmeticLists

benj5378:

---
https://forum.freecad.org/viewtopic.php?t=78050


benj5378:
It's 03:30 and not a goal finishing tonight anymore. But as I said @WandererFan solely merging those three AppPropertyLists gets rid of a lot of bloated code

benj5378:
@WandererFan Do we know anyone who would be interested looking into the MSVC build issue? It's due to the explicit template specilization not being succesful. Would be easy to look at without needing to know anything about the FreeCAD codebase.

WandererFan:
We could try uwestoehr.  He is a developer that uses MSVC (I think.  I'm
not clear on the difference between MSVC and VisualStudio).  If all else
fails, maybe werner?  He seems to be equally at home on linux and windows.

Maybe a post in "developer corner"?

benj5378:
@WandererFan Should be quite easy, I just don't have MSVC to do it.

> I'm not clear on the difference between MSVC and VisualStudio

MSVC refers to the compiler itself (Microsoft Visual C++), while Visual Studio refers to the IDE that typically holds MSVC.

WandererFan:
Does this fix #9265 as a side effect?  If not, could we use shared pointers as part of this change?

benj5378:
@WandererFan I didn't fix any memory leaks, but I many and can only conclude they are everywhere across the code.

Everywhere, new pointers are created to pass in a function call. But after the function call, nearly nowhere is the pointer deleted, hence memoryleak.

https://stackoverflow.com/questions/26243640/is-the-pointer-to-data-on-the-heap-deleted-when-it-goes-out-of-scope

The second example in the answer to this question is basically TechDraw code everywhere.

benj5378:
@WandererFan Okay, I never tested the code, but now just functionally tested the code. It only had one bug, and that was due to an already present "bug" in the code. So during this pretty complicated large PR, I didn't introduce a single bug. That is absolutely insane.

The only thing there is missing from my side is the Windows build. If you want, you can start getting functional testers.

WandererFan:
I can't build this on linux.  Lots of complaints about the templates.

benj5378:
> I can't build this on linux. Lots of complaints about the templates.

Remove the `class TechDrawExport` at explicit template specilization and it will compile.
so it looks like this `template std::string CosmeticExtension::addCosmetic<CenterLine>(CenterLine* cosmetic);`
I will fool a lot around with the template specialization and use GitHub CI to make build tests for me...

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9499">pipeline status</a> for feature branch [PR_9499](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9499). Pipeline [903583280](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903583280) was triggered at [a89bfa3](https://github.com/FreeCAD/FreeCAD/pull/9499/commits/a89bfa308d8458bcffb38a61c2a28b2c79d5b96b). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

WandererFan:
Something wrong with Restore()?  Old files will still arrive with CosmeticEdge and Centerline properties so we need to handle that.    Face centerline created in this version come back on wrong face on Save/Restore.
mergedCosmeticTest

WandererFan:
Any chance we can clean up the commit history a bit?

benj5378:
@WandererFan Face centerline created in this version come back on wrong face on Save/Restore. I tried to reproduce this multiple times but couldn't. Plz send file

WandererFan:
Add centerline to square, add diagonal cosmetic edge to square, save, restore, centerline is in circle.

But this is a false alarm.  Same thing happens with master.   Must be face numbers changing when the diagonal creates 2 faces where there was only 1 before.   Maybe cosmetics should be excluded from face finding.

 
[cubeAndSphere_newCosmetic.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/12095186/cubeAndSphere_newCosmetic.FCStd.zip)

src/Mod/TechDraw/App/Cosmetic.h
```diff
@@ -95,25 +115,19 @@ class TechDrawExport CosmeticEdge : public Base::Persistence, public TechDraw::B
...
+    using Cosmetic::createNewTag;
 
+    using Cosmetic::tag;
 };
 
 //********** GeomFormat ********************************************************
 
 // format specifier for geometric edges (Edge5)
-class TechDrawExport GeomFormat: public Base::Persistence
+class TechDrawExport GeomFormat: public TechDraw::Cosmetic
```
WandererFan, line 132: Why would GeomFormat have Cosmetic as an ancestor?  GeomFormat is not a specialization of Cosmetic.   It is a LineFormat + a geometry index.

src/Mod/TechDraw/App/AppTechDraw.cpp
```diff
@@ -121,7 +121,7 @@ PyMOD_INIT_FUNC(TechDraw)
     TechDraw::GeomFormat            ::init();
     TechDraw::PropertyCenterLineList::init();
     TechDraw::CenterLine            ::init();
-    TechDraw::PropertyCosmeticEdgeList::init();
```
WandererFan, line 124: Can not remove PropertyCosmeticEdgeList as old document files will contain references to this.  Need to keep this around so old documents restore properly.

src/Mod/TechDraw/App/CosmeticExtension.cpp
```diff
@@ -42,7 +42,7 @@ CosmeticExtension::CosmeticExtension()
     static const char *cgroup = "Cosmetics";
 
     EXTENSION_ADD_PROPERTY_TYPE(CosmeticVertexes, (nullptr), cgroup, App::Prop_Output, "CosmeticVertex Save/Restore");
-    EXTENSION_ADD_PROPERTY_TYPE(CosmeticEdges, (nullptr), cgroup, App::Prop_Output, "CosmeticEdge Save/Restore");
```
WandererFan, line 45: This property needs to be retained for backwards compatibility.  We can stop using it for new documents.
- need to modify the Restore method of PropertyCosmeticEdgeList to create the equivalent "Cosmetic" and update PropertyCosmeticList
    - or maybe add a call in onDocumentRestored/onExtendedDocumentRestored to move any CosmeticEdges in PropertyCosmeticEdgeList to PropertyCosmeticList and clear PropertyCosmeticEdgeList

src/Mod/TechDraw/App/Cosmetic.h
benj5378: `Cosmetic` shouldn't be understood in terms of whether it's part of OCC or artificially made by the user. `Cosmetic` should be understood as an object that holds visual data. And we want the objects using `LineFormat` to use dynamic dispatch all the locations were we need to access `LineFormat` without having lots of `bg->source()` trees.

So essentially, GeomFormat inherits TechDraw::Cosmetic because it has been treated throughout the TechDraw code in the same way as CenterLine, CosmeticEdge. It has to do with GeomFormats usage inside the TechDraw code rather than what we as people (without knowledge of TechDraw internals) define as a cosmetic object

Here we see how GeomFormat in the exact same way as CenterLine and CosmeticEdge internally by the code. Therefore we want to see them as the same type and use dynamic dispatch for all of them. Regarding the code, `Cosmetic` should be understood as an object that holds visual data, not as something artificially made by the user.

https://github.com/FreeCAD/FreeCAD/blob/7ba145c36f161d0b25a7f6848cde620e06286cf7/src/Mod/TechDraw/Gui/TaskLineDecor.cpp#L111-L132

src/Mod/TechDraw/App/CosmeticExtension.cpp
benj5378: @WandererFan Do you know from where PropertyCosmeticList is called?

WandererFan: PropertyCosmeticList is your new property?  You would have to search for Cosmetics.getValues() and Cosmetics.setValues.

WandererFan:
This branch: https://github.com/WandererFan/FreeCAD/tree/PR9499Restore
shows how to use onExtendedDocumentRestore to move old style cosmetics to the new style.   

- add back the old propertylists and objects.
- rename new CosmeticEdge and CenterLine.  You could do this without changing the names, but it would be much harder.  This way old and new do not conflict.
- ensure ancestor onDocumentRestored gets executed in DrawView so onExtendedDocumentRestored get executed in CosmeticExtension.
- CosmeticExtension::onExtendedDocumentRestored converts any old propertylists to new propertylists.  It also deletes the old style items.

WandererFan:
closing as too stale.

PR #9501, opened 2023-05-07
fix library vulnerability

mosfet80:
fix 
[CVE-2023-0465](https://www.cve.org/CVERecord?id=CVE-2023-0465)
[CVE-2023-1255](https://www.cve.org/CVERecord?id=CVE-2023-1255)
[CVE-2023-0464](https://www.cve.org/CVERecord?id=CVE-2023-0464)

---

### DIFF
--- a/conda/environment.devenv.yml
+++ b/conda/environment.devenv.yml
@@ -75,7 +75,7 @@ dependencies:
 - ninja
 - numpy
 - occt==7.6.3
-- openssl==3.0.9
+- openssl==3.0.*
 - pcl
 - pip
 - pivy
--- a/conda/environment.devenv.yml
+++ b/conda/environment.devenv.yml
@@ -75,7 +75,7 @@ dependencies:
 - ninja
 - numpy
 - occt==7.6.3
-- openssl==3.0.8
+- openssl==3.0.9
 - pcl
 - pip
 - pivy


conda/environment.devenv.yml
```diff
@@ -75,7 +75,7 @@ dependencies:
 - ninja
 - numpy
 - occt==7.6.3
-- openssl==3.0.8
+- openssl==3.0.9
```
adrianinsaval, line 78: @oursland could you confirm this is ok? maybe we can specify openssl==3.0 so we just get the latest point release in this series rather than having to manually upgrade for minor releases?

oursland: Unfortunately, [conda-forge](https://anaconda.org/conda-forge/openssl/files) does not have openssl==3.0.9 at this time, so this PR would fail.

I attempted to specify `openssl==3.0`, but it resulted in an uninstallable set of conflicts.  I suspect that version is interpreted differently than what is expected.  Using `openssl<3.1.0` did install the highest version of openssl 3.0.x, which is currently 3.0.8.

I suggest we use `openssl<3.1.0` for the version specification.  Unfortunately, this will not resolve the CVEs until conda-forge adds OpenSSL 3.0.9.

adrianinsaval: is there a problem in using openssl 3.1?

oursland: Yes, all builds of OpenSSL 3.1.x for MacOS ARM result in an invalid instruction upon launch in debugging mode.  There may be other issues, but not being able to debug the build is a major issue that needs to be addressed by Conda Forge before we can move up.

Some discussion about this issue can be seen on this Stack Overflow: https://stackoverflow.com/questions/74059978/why-is-lldb-generating-exc-bad-instruction-with-user-compiled-library-on-macos

adrianinsaval: Can we write `3.0.*`?

oursland: Yes, for new installations, but without specifying specific versions a previous install will consider 3.0.8 as satisfying the solver.  That said, so would the `<3.1.0` suggestion I made yesterday.

mosfet80:
openssl 3.0.9 are now on conda https://anaconda.org/conda-forge/openssl/3.0.9/download/osx-arm64/openssl-3.0.9-h53f4e23_0.conda

adrianinsaval:
can you change the PR to specify 3.0.* instead of 3.0.9 so that we don't have to do this every time they make a point release?

PR #9515, opened 2023-05-08
Arch: Removed html from tooltips

yorikvanhavre:
HTML makes translation more difficult, so should be avoided when it's not needed.

Makes progress towards fixing https://github.com/FreeCAD/FreeCAD-translations/issues/105


chennes:
@yorikvanhavre a couple of these changes are adding HTML entities, rather than removing them. Is that intentional?

yorikvanhavre:
I only removed the <html> tags. But indeed some things like quotes were automatically changed to the equivalent html escape code... It's qtdesigner that did that I think. Not sure if that's an issue... I think it's standard XML? Like, I understand you cannot have a < character inside a text in xml...

Maybe it's better to just change the texts to not have them, actually.

yorikvanhavre:
Uh oh... messed things up. I'll close and redo

Issue #7746, opened 2022-11-08
User/Document/Feature level units selection [Feature Request] <title>


gotchoices:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=34&t=34834

### Subproject(s) affected?

Core

### Idea description

This would allow the user to specify preferred display units at three prioritized levels:

- User Preferences (most general default)
- Document (project/part file)
- Feature (individual constraint, dimension, etc)

This shouldn't disrupt the idea of storing feature dimensions internally in a single, standard way (mm) but with this enhancement, the author (not necessarily the current user) would be in charge of how units are displayed on dimensions, features and technical drawings.

This would also solve the matter of this [related feature request](https://github.com/FreeCAD/FreeCAD/issues/7395) and perhaps some others.

Objectives:
- If an author creates a file/model using a certain unit of measure, he can attach that preferred unit to the document (in the properties header) and that will carry along to anyone else who may open the document.
- If an author creates a dimension specifying a particular unit of measure, that unit will stick with the constraint/feature and it will always display with those units regardless of document or user settings.
- A model/drawing could contain multiple dimensions, each displaying with a different unit of measure.  For example, an architectural drawing might use primarily meters, but devolve to cm for smaller features.  A mechanical assembly can contain Imperial/US and metric components and display each differently with its native units.
- The user has a quick, efficient way to specify display units on individual constraints/features.  As an example, one might put a special symbol right after the units when entering (25.375 inch!).  The exclamation point would indicate that the input units should also be the display units.  The user should also be able to modify the display units for existing dimensions.
- Files which specify no unit will fall back on the user preference.  Dimensions that specify no unit will rely on the file default, or then the user preference default.  This should supply backward compatibility.

### Anything else?

I will drop a bounty on bountysource.com in case others also want to contribute to it.  For purposes of claiming the bounty, the standard will be to create the feature request, meet the objectives above, and get the feature integrated into the main codebase.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gotchoices:
[Here](https://app.bountysource.com/issues/113499254-user-document-feature-level-units-selection-feature-request-title) is the issue on bountysource.com.

If you're interested in supporting this feature, consider adding to the bounty.

AgCaliva:
hi is this available? i want to do this and other bounties. Where can i contact you?

gotchoices:
[...]

PR #9521, opened 2023-05-09
Adding "ProjectUnitSystem" support to project files.

AgCaliva:

---
Im doing PR to solve this issue:
https://github.com/FreeCAD/FreeCAD/issues/7746
It solves what is asked related to:
User Preferences (most general default)
Document (project/part file)

I will add another commit in a week or two for:
"Feature (individual constraint, dimension, etc)"
EDIT:
Will PR again for the last feature (individuall constraint) so its only this.

Thank you in advance and i wait for suggestions.


adrianinsaval:
Unit settings tab has been removed in master and it's content moved to the general tab, please adapt your PR to this

AgCaliva:
I see, im sory it seems i closed it by accident, i will correct it.
What it does is that it adds a setting in the General settings that lets you save per project the unit system used. If per project unit system is not specified or its ignored then it uses old user unit system preference.

adrianinsaval:
Does it switch units as you move from one file to another? Is the selected unit for the file changed when you change units with the file open or do you need to manually set a specific unit for it? Current version of master also has a tool button on the status bar to quickly switch between units, did you take this into account in your PR?

AgCaliva:
Yes it does switch units when you move from one file to another. It dont takes into account the new tool on the status bar, will see if it does problems but i think it should not. I guess we should add the "ignore ProjectUnitSystem" checkbox to the new tool on the status bar, so it makes sense.

AgCaliva:
I dont think the new tool in status bar will be needed after adding this and the per constraint/measure unit system feature. Now the unit system from the toolbar its being ignored.

PaddleStroke:
The tool in status bar has been implemented on purpose because people complained it was annoying to access it in preferences. So it's not going away.

Document-level unit system has (imo) nothing to do in the preference page. As it's not a system preference but a per-document property.

I think the correct place is in the status bar unit widget like in this mockup I just did : 
image

And only here, not in the preference page.

So to do this you to modify the label-button such that instead of showing directly the entries of the unit system, it shows a QWidget that contains 2 labels and 2 combobox.

I haven't review functionality or code.

AgCaliva:
For sure i can do that if its required, but i think people complained about have to going to settings because this feature and "per constraint" unit metrics was not existing. It will be easy to me to add that, anyway the functionality will not change a lot so you could say if its ok or not. Thank you.

adrianinsaval:
I can speak for myself that I would still find the status bar selector useful even if per project units where implemented

chennes:
I think maybe the status bar becomes the "set for this document" and in preferences you have "set default for new documents."

adrianinsaval:
> I think maybe the status bar becomes the "set for this document" and in preferences you have "set default for new documents."

tbh I don't want to lose the functionality to quickly set the unit, if I have to create a set of inch based parts it's easier to change the default for a little while rather than changing it for every new file. I would instead suggest that we use this statusbar menu to set the default when no file is active and to set the file units when one is active, and add a little button on the right side of every entry to set it as default.

adrianinsaval:
> Document-level unit system has (imo) nothing to do in the preference page. As it's not a system preference but a per-document property.

agreed, it should be available in the property tab when selecting the document in the tree IMO

> So to do this you to modify the label-button such that instead of showing directly the entries of the unit system as actions, it shows a QActionWidget that contains 2 labels and 2 combobox. If you need a qactionwidget example, have a look at the sketcher constraint taskbox filter button.

if you want to have a menu within a menu the best way would be to use sub menus rather than adding combobox inside the menu I would think. But like I said earlier I think the same menu could be used for both things.

sliptonic:
Maintainers:  According to contributing guidelines, we are not making value judgments about the PR.  We are asking if it does what the issue described and if the code quality is sufficient. 
I read the issue (and all associated comments) and this PR resolves the problem.

If the code quality is good enough, it should be merged.  We can address shortcomings in the UI implementation in subsequent PRs.

I'm not qualified to judge most of the code.  The only criticism I have is that the functionality should not be putting red errors/warnings in the report view for normal situations.  Opening an existing document that doesn't have  a stored unit system is a normal situation.  So this error should be resolved before merging.

`09:48:05  ProjectUnitSystem not found
09:48:05  Reading failed from embedded file: GuiDocument.xml (160628 bytes, 7576 bytes compressed)`

sliptonic:
@AgCaliva Can you add another commit resolving the error messages in the console when opening an existing document that doesn't have a schema?  

When that's done, I'll merge the PR unless there are code reviews or other concerns about code quality.

luzpaz:
@AgCaliva ping

AgCaliva:
Yes will complete in one or two days. Thanks!

sliptonic:
I'm still seeing notices in the report window.
image

AgCaliva:
Ok will comment the warning. And i have find out why it stills get thrown the exception even after being catch, im working on it probably will have it tomorrow. Sorry the delay.

AgCaliva:
I have been trying to solve this last error but it seems that progressive reading the xml doesnt allow to get back to previous point, and calling readEndElement() after some error will always throw another error even if first was caught.
Im adding to src/Base/Reader.cpp the feature to have optional elements, like its implemented in src/Base/Parameter.cpp since i see as bad design to add empty element to xml just to avoid this error.
So it should be possible to check from Reader.cpp if an element its present before calling readElement or readEndElement.

Will try to have it done tomorrow.

sliptonic:
Thanks for tackling this!  I didn't expect it to become a bigger issue but I'm glad you're chasing the bug upstream rather than letting it go.  I've seen these kinds of noisy errors/warnings before so this might clean up the report view for more than just the ProjectUnitSystem.

AgCaliva:
Sory i got a little busy but still working on this, will commit asap this week.

AgCaliva:
I think maybe it will take some more time than a week since its not so easy as i tought, im not sure yet what will be better, if do an inherited class of ParameterGrp, or create a new similar class for reading XML from zip files (FCSTD).

AgCaliva:
Already make it work, but im thinking what will be better for the code.

luzpaz:
bump

AgCaliva:
A little patience, i think i commit tomorrow.

AgCaliva:
Sory i have been a busy but i will upload what i have untill now in 2 hours, and will try to finish this weekend. Sory for the delay.

AgCaliva:
Will try to finish it this weekend, i would suggest to rename XMLReader from src/Base/Reader.cpp to XMLProgressiveReader since its what it implements from Xerces, then we can rename new class DocumentReader to XMLReader. What do you think? I should do it? TY

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9521">pipeline status</a> for feature branch [PR_9521](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9521). Pipeline [903583285](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903583285) was triggered at [52c55f5](https://github.com/FreeCAD/FreeCAD/pull/9521/commits/52c55f59a80947ab481e1b00ae4682999d2b9b77). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AgCaliva:
I have kinda finished but i wanted to use the two methods and i have come across the zipios folder because im having problems. Im curious, why is FreeCAD not using the zipios lib, and instead using the srcs from zipios folder?

AgCaliva:
The reading part seems to need a lot of rework, im not sure i can do it since i need to pay my rent and i need to finish this bounty to get paid, i would like to do it but since i cant get some part of the pay before, maybe someone can donate me 40$ so i can survive?  if not probably will have to let just the original workaround and advance with the original features needed by the bounty. Thank you and sorry to say this here.

sliptonic:
The PR is showing a merge conflict.  Can you please resolve this.  I will try to get the PR merged if there's nothing major standing in the way.

AgCaliva:
This last commit is the working one, but it should be tested since i have reworked a lot, the reading of GuiDocument.xml is being done with DocumentReader.cpp. There is still a merge conflict? Let me know if thats the case. TY.

sliptonic:
Yes, there's still a merge conflict with /src/Base/Reader.cpp

AgCaliva:
Whats going on? im not on date with those features in git, it says build failed? im building successfully in ubuntu 2204...

adrianinsaval:
the tests are failing, check the log and try to find what is the cause, you might need to adapt some tests to your changes or fix some regression introduced by your changes.

AgCaliva:
hummm im not sure whats wrong, im literally compiling and running on Ubuntu 22-04, im not being able to get the reason from the log yet it seems to be something related to function activeWindow() but i have not changed nothing there, the call from src/Gui/Application.cpp:slotActiveDocument() cant be affected by the code i added, maybe i need some help? i cant find whats wrong, Whats the test that its failing?

AjinkyaDahale:
This build fails the exact same way builds sometime fail in my local machine: all tests conclude, but there's a Qt window related segmentation fault.

AgCaliva:
@adrianinsaval im not fixing every issue in the project, im only implementing the features requested. I got pissed off anyway, i would like if you are amenable to split the original issue and close the part i solved since im not continuing, so i can get some part of the money. Thank you.

AjinkyaDahale:
@AgCaliva I haven't looked at the changes so can't comment on the code. I'll just mention the failing tests.

There are 3 tests that fail and a segmentation fault at the end, once all the tests are done. The segmentation fault is probably out of scope here. 2/3 failed dests are as follows. I'll update with the third test when I find it.
'''
FAILED: test_femobjects_open_de9b3fb438 (femtest.gui.test_open.TestObjectOpen)
...
FAILED: test_femobjects_open_head (femtest.gui.test_open.TestObjectOpen)
...
'''

adrianinsaval:
> @adrianinsaval im not fixing every issue in the project, im only implementing the features requested. I got pissed off anyway, i would like if you are amenable to split the original issue and close the part i solved since im not continuing, so i can get some part of the money. Thank you.

I'm not the one managing bounties, perhaps @sliptonic or @yorikvanhavre ? and I'm not sure if we can say that you solved something if it isn't merged/mergeable.
In hindsight with something as complex as this seems to be it might have been a better approach to do this with a grant so you can work on it and get paid at least something beforehand, it sounds to me that for a bounty you need to fix and get it merged first before getting paid. Actual FPA members might be able to give you better information on the monetary aspect.

About the technical problems you are having with this PR, I don't think I have enough knowledge to properly asses this.

If you believe the bundled zipios is not up to the task, there is a cmake option to use an external library, and you could update or ask for help to update the library in the source code. Can you give more details on what is problematic about the bundled version of zipios? (and for the record, I agree that maintaining our own copy of 3rd party code is not ideal)

You also say you are not fixing every issue on the project, and of course I understand that, however that is not what I was asking you to do. The tests seem to be passing without problems on the master branch so it would be your PR that is introducing some problems, when implementing a feature it is part of the task to not break the rest of the software with it.

Ajinkya and others might be able to help in identifying the problems your PR introduces or if it's not actually something from your PR, unfortunately I don't have the time to dig deep into this one.

AgCaliva:
I cant continue untill get some money, i have bills to pay, i guess someone can continue this if its needed. I love software but i dont live from air

sliptonic:
This was a bountysource bounty (linked in the original issue).  It's not managed by FPA or anyone here.  I suggested that the guy who offered the bounty pay 1/2 when the PR was created and the remaining 1/2 when it was merged.  I don't know if he followed through on the first half or not.

As far as I'm concerned the functionality is fine and has been for a long time.  I made some suggestions to suppress some noisy messages in the console and that apparently became more work than expected.  That's unfortunate. But we can't merge if it doesn't pay CI/CD tests.

@AgCaliva,   Adrian is right about grants.  The FPA has money to offer on grants if there's something you'd like to work on.  However, the grant criteria will work largely the same way:  payment in installments.  One part paid when a solid proposal is delivered, another when a PR is submitted, and the final when the PR is merged.  If you have something you'd like to work on, please consider applying.  https://fpa.freecad.org/programs/fpadf-announcement

AgCaliva:
I have talked with gotchoices a long time ago, he never answered again my mails, he dont wanted to get refund the bounty from bountysource and payme some part because he is 'busy', so yes its over.

sliptonic:
That's unfortunate.  I'm sorry it happened.  My own experience with software bounties has been that they are generally not effective.  In fact, we (Ondsel) published a blog post about them today.  The title is a bit opinionated but the article goes into good depth about why they don't work.   https://ondsel.com/blog/software-bounties-are-a-dumb-idea

AgCaliva:
I mean i could had just done a workaround and just let the code as bad as it is now. But im not that kind of developer. I do things correctly or dont do nothing.

sliptonic:
@AgCaliva If you're interested in finishing this for a an FPA development fund grant, please email me at brad 'at' ondsel.com

AgCaliva:
@sliptonic thank you so much! im talking now with gotchoices and he is willing to pay me some part, so will continue this, finish merge and finish the other feature probably. Its not needed to get more money.

AgCaliva:
@sliptonic Sory but it seems gotchoices its not going to do anything since he dont know how to manage the bountysources bounty or send/receive crypto... I dont see your email in ondsel, my mail is ramiel1390@gmail.com please write me TY.

AgCaliva:
Im fixing the tests and enhancing the code asap.

luzpaz:
bump

AgCaliva:
Im working on this, i have already fixed the issue but im trying to make sure that everything working as expected.

AgCaliva:
I needed to implement some things for having the tests working correctly,
More specifically i had to implement some Restore functions in Mod/Sketcher/Gui/PropertyVisualLayerList.cpp, so i will try to commit the working code in a day or two.

AgCaliva:
Im not sure whats going wrong with the Windows build, someone knows? i dont have much experience with visual studio

AgCaliva:
I think i have found whats wrong, im not sure but i think the merge from web in github its a little buggy, one of the merges content was ignored even when it says it modified files, so will revert those changes and do it via cli

adrianinsaval:
It's weird that every time you push changes it requires approval for the CI to run

chennes:
> It's weird that every time you push changes it requires approval for the CI to run

@AgCaliva will be a "first-time contributor" until we merge this PR -- after that we won't have to manually approve them.

AgCaliva:
Im setting up windows to test until solve.

AgCaliva:
Hi, someone knows why im getting this error?
fmt.lib(format.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in Axis.obj
I have run cmake from cli and i get those errors when trying to compile in visual studio. Thank you in advance.

AgCaliva:
Dont matter it seems i was using wrong version of Libpack, and now i understand the problem, some DOMDocument usage its being taked as DOMDocument from msxml.h.

AgCaliva:
The problem i have seems to be related to this:
https://stackoverflow.com/questions/12591894/c-domdocument-redefinition-error-with-xercesc
and to solve it i had to change the line were some headers were added. But that causes the libpack to fail since it seems some files were using headers from this proyect instead of having their own needed headers included. I will probably need to add some headers to Libpack.
This is a clear example of how using incorrectly the headers can lead you to really annoying bugs, so i have to mention that someday it will be needed to curate the headers used in all .cpp files, including the ones from Libpack.
Because of this i have been debugging 24 hours without stop until i figured what was wrong. So thats why its taking more than the time needed for the task.
I mention this since i have seen a more than some of not needed headers added in places were was not used, or headers missing in other places.

AgCaliva:
hummm im not able to reproduce this, in my pc its working

AgCaliva:
I have added an temporary compiler option so it prints the included headers while compiling, so even if i cant reproduce now it should tell me where its wrong. i will remove this option after fixing the bug, so it should not be automatically merged, i dont know if someone can avoid that.

chennes:
I don't know that it's much help, but this PR compiles without any problems on my Windows box, so whatever is going on seems to be specific to the CI environment. @0penBrain, any insight into the Windows build failures due to DOMDocument redefinition?
```
2023-09-09T22:46:10.6609216Z        "C:\FC\build\ALL_BUILD.vcxproj" (default target) (1) ->
2023-09-09T22:46:10.6609592Z        "C:\FC\build\src\Mod\TechDraw\Gui\TechDrawGui.vcxproj" (default target) (117) ->
2023-09-09T22:46:10.6609851Z        (ClCompile target) -> 
2023-09-09T22:46:10.6610533Z          C:\Program Files (x86)\Windows Kits\10\Include\10.0.20348.0\um\msxml.h(9010,1): error C2371: 'DOMDocument': redefinition; different basic types [C:\FC\build\src\Mod\TechDraw\Gui\TechDrawGui.vcxproj]
2023-09-09T22:46:10.6610645Z 
2023-09-09T22:46:10.6610746Z 
2023-09-09T22:46:10.6611111Z        "C:\FC\build\ALL_BUILD.vcxproj" (default target) (1) ->
2023-09-09T22:46:10.6611643Z        "C:\FC\build\src\Mod\PartDesign\Gui\PartDesignGui.vcxproj" (default target) (93) ->
2023-09-09T22:46:10.6612022Z        "C:\FC\build\src\Mod\Sketcher\Gui\SketcherGui.vcxproj" (default target) (95) ->
2023-09-09T22:46:10.6612697Z          C:\Program Files (x86)\Windows Kits\10\Include\10.0.20348.0\um\msxml.h(9010,1): error C2371: 'DOMDocument': redefinition; different basic types [C:\FC\build\src\Mod\Sketcher\Gui\SketcherGui.vcxproj]
2023-09-09T22:46:10.6612872Z 
2023-09-09T22:46:10.6612971Z 
2023-09-09T22:46:10.6613275Z        "C:\FC\build\ALL_BUILD.vcxproj" (default target) (1) ->
2023-09-09T22:46:10.6613643Z        "C:\FC\build\src\Mod\MeshPart\Gui\MeshPartGui.vcxproj" (default target) (88) ->
2023-09-09T22:46:10.6614374Z          C:\Program Files (x86)\Windows Kits\10\Include\10.0.20348.0\um\msxml.h(9010,1): error C2371: 'DOMDocument': redefinition; different basic types [C:\FC\build\src\Mod\MeshPart\Gui\MeshPartGui.vcxproj]
2023-09-09T22:46:10.6614482Z 
2023-09-09T22:46:10.6614582Z 
2023-09-09T22:46:10.6614892Z        "C:\FC\build\ALL_BUILD.vcxproj" (default target) (1) ->
2023-09-09T22:46:10.6615234Z        "C:\FC\build\src\Mod\Fem\Gui\FemGui.vcxproj" (default target) (72) ->
2023-09-09T22:46:10.6615920Z          C:\Program Files (x86)\Windows Kits\10\Include\10.0.20348.0\um\msxml.h(9010,1): error C2371: 'DOMDocument': redefinition; different basic types [C:\FC\build\src\Mod\Fem\Gui\FemGui.vcxproj]
```

AgCaliva:
There is no problem im fixing it i guess it does not trigger bugs because its something related to modules.

AgCaliva:
Hi there i need the workflow to be run please so i can finish

AgCaliva:
Again its not applying the changes from the merge done by the web so i have to revert that and do it via cli. I guess thats something github will never fix.

AgCaliva:
Hi, i have found that pre-commit hooks its reordering in alphabetical order the headers which its problematic for solving the DOMDocument redefinition error(the issue i have now.) i think i can avoid the precommit to run once, but probably when someone else will do other commit it will be the same, as you can see the file that has the error and pre-commit hook is getting wrong is: src/Mod/MeshPart/Gui/Command.cpp from my last commit.

0penBrain:
> Hi, i have found that pre-commit hooks its reordering in alphabetical order the headers which its problematic for solving the DOMDocument redefinition error(the issue i have now.) i think i can avoid the precommit to run once, but probably when someone else will do other commit it will be the same, as you can see the file that has the error and pre-commit hook is getting wrong is: src/Mod/MeshPart/Gui/Command.cpp from my last commit. Also it has potentially reintroduced the bug in others files, what we can do? what i can do about this? i need to solve this, this PR its taking forever please HELP.

That's a cool finding. This said, we should virtually be able to include the headers in any order without having problems. If there is one, then a protection against re-include is missing or something like that. Can you detail the header path that leads to the issue ?

WandererFan:
You could wrap your includes with clang-format tags to prevent  the SortIncludes:true option from sorting them.

// clang-format off
#include myfile.h
#include another.h
...
// clang-format on

AgCaliva:
I have already shared whats the issue:
https://stackoverflow.com/questions/12591894/c-domdocument-redefinition-error-with-xercesc
"no matter what you do your code will not compile. The compiler will tell you that you are redefining 'DOMDocument', which is actually whats happening. This is because of the way Microsoft wrote msxml.h."
I know it should not matter the order of headers. But Microsoft dont cares.
I will then use the clang-format tags, thank you so much for the help, i have never used hooks.

chennes:
> That's a cool finding. This said, we should virtually be able to include the headers in any order without having problems. If there is one, then a protection against re-include is missing or something like that. Can you detail the header path that leads to the issue ?

Unfortunately this is a rather common problem with Microsoft's headers, in various circumstances. The problem is a fundamental design flaw in the C preprocessor (that definitions are global, and "leak" out of included headers). There's really nothing @AgCaliva can do in this case.

AgCaliva:
Nice, fixed. Removing option for printing included files while compiling
set(CMAKE_CXX_FLAGS "/showIncludes ${CMAKE_CXX_FLAGS}")

adrianinsaval:
On an initial test seems to work, but it seems to be dependent on what is the active tab. In this example Unnamed has it's units set to US Customary and Unnamed1 is in SI. Despite this if Unnamed1 is active properties from Unnamed are still shown in mm:
image
and the inverse is also true, if Unnamed is the active tab properties from Unnamed1 are shown in inches
image

I don't think this needs to be addressed in this PR already but it would be good to eventually make these show the properties in the corresponding unit without depending on the active tab

AgCaliva:
Yes i can fix it and will do it. in a day or two. Thanks to share this.

AgCaliva:
Sorry are you sure it dont works correctly? As far as i can tell its working correctly i have set one of the projects to other Unit System then switched between them and they have their needed unit system. Maybe an step by step reproducing would help?

chennes:
OK, hopefully one last time, can you rebase and resolve those conflicts (presumably due to the formatting work that's being done)? Thanks!

AgCaliva:
Im sory for what i said before, im not sure what introduced the previous new error but i think it was because doing merge via web broke cmake compilation.
Anyway some words about the code in this PR:
DocumentReader was implemented because the other method of reading dont allows to fail, if it fails then it should start reading all over again, but even if thats the case i have not been able to run again the routine for start reading again in a sane way. Also previous method don´t allowed to have optional elements in XML, since i needed to add the projectUnitSystem element, but dont wanted to modify all files that dont include this option(previous version stdfc files) for just a dummy value because i think thats bad practice.
TODO some day:
-DocumentReader class should be used with another layer of abstraction that ease his use, now the DocumentReader class its being directly, but later it would be better to create something like Parameter.cpp, that should implement DocumentReader class. Parameter.cpp indeed should also use DocumentReader so the low level code used for reading the XML files its unified in one class. Im not promising but i would like to do that later with other bounty, one more short. Right now i dont have the time to do this since this implementation was something not related to the original issue. I still have to do another PR for the per-constraint unit system.

-(Suggestion)Related to the new unit system bar that was added while i was doing this PR:
By default/first time opening file it should use the projectUnitSystem value.
Add another element in this bar that use projectUnitSystem value and indicates that its the per project setting, by doing this then people will switch more clearly between per project unit system and arbitrary unit system, then to modify the per-project unit system you always need to do it from Preferences, and it can be used from the bar, or not.

sliptonic:
This is looking very good.  One thing I'm seeing that looks like a bug:  Changing the unit system should mark the document dirty so the user is prompted to save it when closing.  

That's something that can be fixed in a separate PR so I'm inclined to merge it.  I'll leave this until tomorrow to see if there's anymore feedback.

@AgCaliva if you do decide to fix that bug, let me know and I'll hold on it.

Another aside while I'm thinking about it.  Is anyone else bothered by this?  It's currently impossible to tell which schema is currently selected.  Clicking shows the list of schemas but doesn't indicate which is the current.  Easy solution would be to stick the name of the schema into the tooltip.
image

adrianinsaval:
> This is looking very good. One thing I'm seeing that looks like a bug: Changing the unit system should mark the document dirty so the user is prompted to save it when closing

Are you changing the unit system on the preference page or on the status bar widget? Last time I tried this PR the document level unit system could only be changed on the preference page, not on the status bar. Which is terrible UX but you've said this is something to be solved in further PRs, not here.

AgCaliva:
@sliptonic Sory i dont agree with this:
"Changing the unit system should mark the document dirty so the user is prompted to save it when closing"
That should only happen when the unit system was set by using the Preferences tab, since that really modifies the document. As far as i know the bar its not able to modify the document, it should only allow you to switch unit systems(i have not implemented the bar). And for a future development that setting could be saved as user metadata related to that specific project, that its saved in the PC of the user and not in the project files.
The bar was implemented before i add ProjectUnitSystem per project, so it probably only modifies the FreeCAD setting, not the project setting, and i think thats good indeed because per project unit system at least i think it should not be so easy to be changed.
Thats what i think that makes more sense at least, i would like to know what do you think after reading my thoughts about this.

chennes:
> That's something that can be fixed in a separate PR so I'm inclined to merge it.

I think this is a great and valuable development -- we should get it into users hands so they can play with it, and we can refine in subsequent PRs -- I'm happy to see it merged as it stands now, this was a lot of work from @AgCaliva. I'm sure there's more to be done, but it's a great start.

sliptonic:
ok.  Let's get some user feedback.  Improvement will be incremental.

luzpaz:
Trailing whitespace introduced: 
```
Warning: src/App/ExtensionContainer.h:214:    <-- trailing whitespace
Warning: src/App/PropertyContainer.cpp:478:		    <-- trailing whitespace
Warning: src/App/PropertyContainer.cpp:496:	    <-- trailing whitespace
Warning: src/App/PropertyContainer.cpp:526:	    <-- trailing whitespace
Warning: src/Gui/Application.cpp:934:        <-- trailing whitespace
Warning: src/Gui/Application.cpp:946:        <-- trailing whitespace
Warning: src/Gui/PreferencePages/DlgSettingsGeneral.ui:90:       <-- trailing whitespace
Warning: src/Gui/PreferencePages/DlgSettingsGeneral.ui:98:        <-- trailing whitespace
Warning: src/Gui/PreferencePages/DlgSettingsGeneral.ui:106:        <-- trailing whitespace
Warning: src/Gui/PreferencePages/DlgSettingsNavigation.cpp:119:        hGrp->SetASCII("FontString", ui->naviCubeFontName->currentText().toLatin1());        <-- trailing whitespace
Warning: src/Gui/PreferencePages/DlgSettingsNavigation.cpp:195:    <-- trailing whitespace
```

adrianinsaval:
> IMHO, changing in the preference should set what the default unit schema is for new documents. Changing it on the toolbar should change it for the current document and mark the document dirty. We can get the design guys to weigh in on it.

+1 althought I would add at least a right click menu action to set the default on the status bar.

>  The bar was implemented before i add ProjectUnitSystem per project, so it probably only modifies the FreeCAD setting, not the project setting, and i think thats good indeed because per project unit system at least i think it should not be so easy to be changed.

I disagree, changing the document unit system shouldn't be harder than changing the default.

AgCaliva:
Just to clarify a little what i think would be better taking the current state as reference, i would only change the following;
-The bar should not change any setting, but should be saved in a new user metadata file related exclusively to the project opened.
-Something should indicate in the bar if you are using current project setting.
By doing this some problems will be avoided:
-If user wants to change unit system in a temporary way he can by using the bar, resulting in not changing accidentally the project unit system setting.
-If project owner wants to set an project unit system he can without conflicting with other workers, by using Preferences.

sliptonic:
> -The bar should not change any setting, but should be saved in a new user metadata file related exclusively to the project opened.

By 'project' do you mean 'document'?   Why does this need to be a separate file rather than a property of the document?

> -Something should indicate in the bar if you are using current project setting.

The bar should show which unit schema is active for the document that is active. 

> By doing this some problems will be avoided:
> -If user wants to change unit system in a temporary way he can by using the bar, resulting in not changing accidentally the project unit system setting.

Temporary change without changing the document isn't necessary unless there's some big cost to changing schemas.  IMHO, changing schema is a low-cost operation.  I'm just changing how I see things.  If the user changes it and accidently saves the document with the other schema it's no big deal.  They can just change it back and save again, right?

> -If project owner wants to set an project unit system he can without conflicting with other workers, by using Preferences.

Again I don't understand what you mean by project as opposed to document.  Are you thinking about some larger collaboration context that doesn't exist yet?

Let's not overthink it.  

> +1 althought I would add at least a right click menu action to set the default on the status bar.

Not necessary.  User can set default preference in the preference tool and that's the right place to do it.  It's not like users change their default preference all the time.  It's a preference.  They set it and forget it.

AgCaliva:
In this case project and documents means the same, the larger collaboration context was kinda little suggested in the forum, i readed all the discusion there before and thats what i use as reference also adding my opinion. Anyway i just wanted to share my thoughts.

src/App/Document.cpp
```diff
@@ -88,6 +88,7 @@ recompute path. Also, it enables more complicated dependencies beyond trees.
 #include <Base/FileInfo.h>
 #include <Base/TimeInfo.h>
 #include <Base/Reader.h>
+#include <Base/DocumentReader.h>
```
wwmayer, line 91: This header is not used anywhere. So, it shouldn't be included.

src/Base/Parameter.h
```diff
@@ -421,11 +422,43 @@ class BaseExport ParameterManager : public ParameterGrp
     bool          gUseFilter            ;
     bool          gFormatPrettyPrint    ;
 
-private:
     ParameterManager();
     ~ParameterManager() override;
 };
 
+XERCES_CPP_NAMESPACE_USE
```
wwmayer, line 429: A `using namespace` statement inside a header file is a bad idea.

src/Base/XMLTools.h
```diff
@@ -28,16 +28,10 @@
...
 #include <Base/Exception.h>
-
-
-XERCES_CPP_NAMESPACE_BEGIN
-    class DOMNode;
-    class DOMElement;
-    class DOMDocument;
-XERCES_CPP_NAMESPACE_END
-
+XERCES_CPP_NAMESPACE_USE
```
wwmayer, line 34: Same here.

AgCaliva: Yes, my wrong, i got a little crazy trying to solve "DOMDocument redefinition error" at first.

wwmayer:
Until today I wasn't aware of this pending PR and all the sudden saw a lot of modified files when pulling the latest changes.

Since in this thread I couldn't see that there was ever done a proper code review I tried to go through the modified code to get an overview. So, the code **quality** is basically OK except of a few quirks that I hopefully have fixed with my PR:
https://github.com/FreeCAD/FreeCAD/pull/10964

However, I have some concerns about the **design**. What annoys me a lot is the huge amount of duplicated code because this results into a worse maintainability in the future.

After going through the changes a few times I still struggle to understand why this bloat is needed.

To me the major thing are the two methods Document::readUsing_DocumentReader and Document::readUsing_XMLReader. The former uses a DOM document and the latter a SAX parser but do basically the same. And all this work only to read in the additional XML element:
`<ProjectUnitSystem US="0" ignore="0"/>`

Really? Haven't you noticed that with the XMLReader class this is possible, too?

AgCaliva:
Hi, with XMLReader class you cant have optional fields in xml. If you try to read some element that its not present then it throws an exception and cant continue reading forcing you to have double try catch with this element and that you need to read it at last because you cant continue, at least not without starting all over as far as i know. Also you dont need to add ProjectUnitSystem element to files created before this was added. Meaning that if you dont use ProjectUnitSystem it is not added to xml.
Yes you are kinda right its duplicated but enhanced indeed because it dont do some checks that was not doing nothing in existent implementation in Parameter.cpp. And has some aditional functions. It should be later implemented in Parameter.cpp and use Parameter.cpp to read the files, being Parameter an wraper of DocumentReader. The other method i think it should be deprecated.

wwmayer:
> If you try to read some element that its not present then it throws an exception and cant continue reading forcing you to have double try catch with this element and that you need to read it at last because you cant continue, at least not without starting all over as far as i know

But then it's still better to extend the API of XMLReader than adding tons of new functions throughout the API and duplicating the code. And in this special case it doesn't matter to use a try/catch as workaround because it's the very last element to read.


> And has some aditional functions. It should be later implemented in Parameter.cpp and use Parameter.cpp to read the files

Parameter.cpp has nothing to do with (Gui)Document.xml because it's focused on reading and writing user config files. Non-related things shouldn't be mixed together only because they hold a DOM document. 

> XMLReader class should be deprecated, using the method used in DocumentReader gives you more control, and you dont have to check if element or field exists, you just get an null value from the request, replacing a function call with a simple if.

XMLReader uses a SAX parser and this works completely different compared to using a DOM document because you can only see the element the parser is currently processing but also there it's possible to see if an element is optional. All what's needed is a function to read in the start of the next element and return its name. Then the caller will see if the expected element exists or not.

And if we should ever decide to replace the SAX parser with a DOM document then everything should be handled inside XMLReader without affecting any of the other core classes.

sliptonic:
> Until today I wasn't aware of this pending PR and all the sudden saw a lot of modified files when pulling the latest changes.

Sorry @wwmayer.  I'll take blame for that.  The PR has been open since May and there's been more than 80 comments on it but I don't have a good sense when to request a code review from you.   
We're trying to use code reviews more tactically to get the right people weighing in early.  Can you give some guidance on what kinds of PRs you want to look at closer?

wwmayer:
> We're trying to use code reviews more tactically to get the right people weighing in early. Can you give some guidance on what kinds of PRs you want to look at closer?

For all parts where I am the maintainer. These are:
* the core system
* the Inspection module
* the Mesh/MeshPart modules
* the Part module 
* the Points module
* the Reverse Engineering module
* the Surface module
* the Test module

If a change is only about fixing typos or code beautifying then there you don't need to contact. But if you are in doubt then better ping me more often than less often.

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/10972/commits/1c3450dae5196bcb21998fb4cf44548d52e2360e extends the XMLReader class by a method to access the next element and check its current name afterwards. This allows it to check more conveniently for optional elements.

AgCaliva:
Yes but i dont think that two different methods for reading xml are needed, as far as i see  the other its better anyway the pr can be reversed and i can just put the part of the feature with the double try catch

AgCaliva:
Yes but i dont think that two different methods for reading xml are needed, as far as i see  the other its better anyway the pr can be reversed and i can just put the part of the feature with the double try catch

wwmayer:
The new method is incomplete because it implements the reading from XML only for a very small subset of properties and thus it cannot be considered as a replacement at all.

Another serious problem of this implementation is that it exposes private data (the DOMElement) of the reader class to the calling instance and requires the calling instance to correctly interact with the reader class. For more complex property classes this is quite error-prone.

AgCaliva:
The only bad thing about the new method its that it looks ugly because a wraper using DocumentReader its not implemented yet, thats something TODO that i have mentioned. Talking about capabilities and coherence its better. It can be not suited to be merged to master, but i think it goes in the right direction.

AgCaliva:
superflous?

luzpaz:
sounds like the superfluous was in regards to the 'very small subset of properties'. Why take offense to that?

AgCaliva:
./FreeCAD -t 0
Reading with DocumentReader once:
Ran 1311 tests in 2.93e+02s

Reading with XMLReader once:
Ran 1311 tests in 2.85e+02s

Not big differences in performance.

SAX Parser: (XMLReader class)
    Advantages:
        Serial access (flow of events).
        Low memory usage (only events are generated).
    Disadvantages:
        Backward navigation is not possible as it sequentially processes the document.(Right now XMLReader reads the whole file again to read backwards)

DOM Parser: (DocumentReader class)
    Advantages:
        (Object based)Tree data structure.
        Random access (in-memory data structure).
        To edit the document (processing the in-memory data structure).
        To process multiple times (document loaded in memory).
        Ease of navigation.
    Disadvantages:
        High memory usage (the document is loaded into memory, but its not so important since files being read are not big)

BOTH methods dont allows to read SAME file with multithread, but you can read different files, it is mentioned in the XERCES FAQ. This is not big issue but still an issue. For doing multithread reading an extra implementation its needed i think(i mean you dont have it out of the box, and without having to warn about how to do it to dev.

In Conclusion:
	Both methods are interesting but have their constraints. SAX Parser should be used only when you have really big files that cant be readed at once because of memory limits, which its not the case.
Using DOM Parser method its more dev friendly, and results in a more clear implementation of reading/writing xml. Also more complex xml searchs, node traversing, and more complex xml reading and writing operations in general can be done more easily.
	This is the best method for reading and storing data? Well, no. As far as i know JSON its the new cutting edge technology since Google made efforts optimizing libraries some time ago. Knowing this then the best we can do for the current implementation is:
	-Rename XMLReader to SAXParser and DocumentReader to DOMParser. Those names fit better with what they are.
	-Replace old DOM Parser method present in Base/Parameter.cpp with DOMParser class usage.
	-Replace all SAXParser usage with DOMParser class usage unless file size being read its too big.
	-DOMParser should be what its recommended for devs to use in Modules.
Also want to clarify that DocumentReader was thought as low level class, that should be wrapped in a high level class, thats why it looks raw at current state in my fork.
Also some changes where made in how DOMParser functions are used, and some little details in Master implementation that i tried to make the code more clear, when Restore() function its called, that were ignored. I will PR that later maybe.

Talking about the StatusBar issue, i have already said what it should be done. ProjectUnitSystem value should only be changed from Preferences. The status bar should use this value if set at first, but if you change it it should store in some local/user folder as user metadata, and be read from there instead.
Why? because i think most of the serious people that uses this software probably uses some version control system, and this would avoid to actually modify the project file when user temporarily changes unit system.
Also probably someday FreeCAD will like to have user metadata, so its a good place to start i think.
TY.

AgCaliva:
I forgot to mention that the comparison made by my self before in performance its not well done because its not using last version of XMLReader usage. So probably they are more similar, but if data its not serial then its exponentially slower since you have to read file again each time you check for some element existence, until getting to the same point that was before the error in file. So current implementation its not correct, that was my thought process on deprecating XMLReader.
Also i don think its correct to depend on serialization of data.
@wwmayer what do you think about this?

AgCaliva:
@wwmayer ? those lagging errors can be related to the changes you implemented to XMLReader, you should test that.

Issue #9479, opened 2023-05-04
[Bug] Workbench selector index issue


Roy-043:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecad.org/viewtopic.php?t=77280

### Version

0.21 (Development)

### Full version info

```
OS: Windows 8.1 Version 6.3 (Build 9600)
Word size of FreeCAD: 64-bit
Version: 0.21.0.32985 (Git)
Build type: Release
Branch: master
Hash: 5cd377f931312ede305e4c764ba2028b20eaacfd
Python 3.10.10, Qt 5.15.8, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: Dutch/Netherlands (nl_NL)
Installed mods:
```

### Subproject(s) affected?

Core

### Issue description

After disabling a workbench on the "Available Workbenches" tab in the [Preferences Editor](https://wiki.freecad.org/Preferences_Editor#Available_Workbenches) there is an index issue if you do not restart FreeCAD.
1. In the Preferences Editor go to Workbenches -> Available Workbenches.
2. Disable the first WB in the list (Arch WB in my case).
3. Press OK, but do not restart FreeCAD.
4. Select the 4th workbench from the selector.
5. Result: Correct WB is activated.
6. Select the 5th workbench from the selector.
7. Result: the 6th WB is activated.
8. Again select the 4th workbench from the selector.
9. Result: the 5th WB is activated.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
@PaddleStroke could you take a look please?

PaddleStroke:
Yes I'm looking into it today

PaddleStroke:
So what happens is that the disabled workbench does not disappear from the workbench selector Combobox on-the-fly. However the list of workbench is actually refreshed when you change of workbench.

It appears that `WorkbenchGroup::addTo` is called : 
- On app start for QToolBar and for QMenu.
- On every workbench change for QMenu only.

So when addTo is called for QMenu it calls `refreshWorkbenchList` and recreates the view/workbenches menu.

So actually the view/workbenches menu becomes correct (without the disabled wb) after this. And then it creates the index issue for the comboxbox (as the list is updated but not the combobox items).

What I'm not sure to understand is why addTo is called for the QMenu on every wb change. I seemed to understand that only the 'not-already-created' menus / toolbars are addTo.

[...]

PR #9534, opened 2023-05-10
Core: rework WorkbenchGroup & WorkbenchComboBox

PaddleStroke:
To solve : https://github.com/FreeCAD/FreeCAD/issues/9479

I came upon what seemed to be unecessarily convoluted code (please tell me if I'm missing something). So I simplified it and solved the mentioned issue. 
Also it now have the advantage of not requiring a restart of the application when the wb are activated / moved.

I may have missed some advanced usecase (such as changing of workbench using python) so if you know of them could you test if they are still ok?

---

### DIFF
--- a/src/Gui/Action.cpp
+++ b/src/Gui/Action.cpp
@@ -59,6 +59,7 @@
 #include "WhatsThis.h"
 #include "Widgets.h"
 #include "Workbench.h"
+#include "WorkbenchManager.h"
 #include "ShortcutManager.h"
 #include "Tools.h"
 
@@ -602,34 +603,8 @@ void ActionGroup::onHovered (QAction *act)
 
 // --------------------------------------------------------------------
 
-namespace Gui {
-
-/**
- * The WorkbenchActionEvent class is used to send an event of which workbench must be activated.
- * We cannot activate the workbench directly as we will destroy the widget that emits the signal.
- * @author Werner Mayer
- */
-class WorkbenchActionEvent : public QEvent
-{
-public:
-    explicit WorkbenchActionEvent(QAction* act)
-      : QEvent(QEvent::User), act(act)
-    { }
-    QAction* action() const
-    { return act; }
-private:
-    QAction* act;
-
-    Q_DISABLE_COPY(WorkbenchActionEvent)
-};
-}
-
 WorkbenchComboBox::WorkbenchComboBox(WorkbenchGroup* wb, QWidget* parent) : QComboBox(parent), group(wb)
 {
-    connect(this, qOverload<int>(&WorkbenchComboBox::activated),
-            this, qOverload<int>(&WorkbenchComboBox::onActivated));
-    connect(getMainWindow(), &MainWindow::workbenchActivated,
-            this, &WorkbenchComboBox::onWorkbenchActivated);
 }
 
 void WorkbenchComboBox::showPopup()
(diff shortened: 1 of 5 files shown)


adrianinsaval:
It's crashing while running the GUI tests, it only shows up for ubuntu because GUI tests are not run on the windows CI
given this line on the errors:
`#19  0x7f278896d419 in Gui::Application::sRemoveWorkbenchHandler(_object*, _object*) from /usr/local/lib/libFreeCADGui.so+0x119`
I think this is the test that is crashing:
https://github.com/FreeCAD/FreeCAD/blob/ecb72b8eadced103cc648990f8ebe251d64c3f43/src/Mod/Test/Workbench.py#L59-L78

it might be causing refreshWorkbenchList to access a null pointer or something like that

wwmayer:
This code snippet is extracted from the unit test and it crash at the line when it removes the currently active workbench:
```
import __main__
class UnitWorkbench(__main__.Workbench):
    MenuText = "Unittest"
    ToolTip = "Unittest"
    def Initialize(self):
        cmds = ["Test_Test"]
        self.appendToolbar("My Unittest",cmds)
    def GetClassName(self):
        return "Gui::PythonWorkbench"

FreeCADGui.addWorkbench(UnitWorkbench())
wbs=FreeCADGui.listWorkbenches()
FreeCADGui.activateWorkbench("UnitWorkbench")
FreeCADGui.updateGui()
FreeCADGui.activeWorkbench().name()=="UnitWorkbench"
FreeCADGui.removeWorkbench("UnitWorkbench") ### <<<<<<<<<<======== CRASH!
wbs=FreeCADGui.listWorkbenches()
```

Internally removeWorkbench() calls WorkbenchGroup::refreshWorkbenchList() and since the active workbench has been removed `WorkbenchManager::instance()->active()` return a null pointer which isn't checked and thus crashes when accessing the name() method.

wwmayer:
> I came upon what seemed to be unnecessarily convoluted code (please tell me if I'm missing something).

Things worked a bit differently at Qt3/Qt4 times and required some workarounds to avoid crashes when switching between workbenches.

PaddleStroke:
> Things worked a bit differently at Qt3/Qt4 times and required some workarounds to avoid crashes when switching between workbenches.

Thanks for the feedback. Then is my code proposal good / didn't I miss some usecases ?
I'm not sure to understand your explanations concerning the test crash so I will need to get back to this in more depth.

wwmayer:
> Then is my code proposal good

Except for the crash. Look at my comments...

> didn't I miss some usecases

1. Changing the wb from Python updates the combo box. OK
2. Changing the wb from the combo box updates the sub-menu. OK
3. Changing the wb from the sub-menu updates the combo box. OK

> I'm not sure to understand your explanations concerning the test crash so I will need to get back to this in more depth.

The call of removeWorkbench() internally calls WorkbenchGroup::refreshWorkbenchList() which gets the active workbench. But if the removed wb was the active wb then at this point there is no active wb any more. WorkbenchManager::instance()->active() returns a null pointer in this case but further down the null pointer will be de-referenced.

wwmayer (changes requested):
See comments in source code

sliptonic:
Is this now fully resolved?  Can it be merged?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9534">pipeline status</a> for feature branch [PR_9534](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9534). Pipeline [903667667](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903667667) was triggered at [c3c5f49](https://github.com/FreeCAD/FreeCAD/pull/9534/commits/c3c5f491ef62fad70169bfc59f7f717ce9dac959). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #9565, opened 2023-05-14
PD: fix out of scope warnings of Clone feature

wwmayer:
If the BaseFeature link of the Body of a Clone feature isn't set then do not update it. Forum thread: https://forum.freecad.org/viewtopic.php?p=675644#p675644

---

### DIFF
--- a/src/Mod/PartDesign/App/FeatureBase.cpp
+++ b/src/Mod/PartDesign/App/FeatureBase.cpp
@@ -73,18 +73,22 @@ App::DocumentObjectExecReturn* FeatureBase::execute() {
     return StdReturn;
 }
 
+void FeatureBase::trySetBaseFeatureOfBody()
+{
+    if (auto body = getFeatureBody()) {
+        if (BaseFeature.getValue()
+                && body->BaseFeature.getValue()
+                && body->BaseFeature.getValue() != BaseFeature.getValue()) {
+            body->BaseFeature.setValue(BaseFeature.getValue());
+        }
+    }
+}
+
 void FeatureBase::onChanged(const App::Property* prop) {
 
     // the BaseFeature property should track the Body BaseFeature and vice-versa
     if (prop == &BaseFeature) {
-
-        auto body = getFeatureBody();
-        if(!body)
-            return;
-
-        if (BaseFeature.getValue() && body->BaseFeature.getValue() != BaseFeature.getValue()) {
-            body->BaseFeature.setValue(BaseFeature.getValue());
-        }
+        trySetBaseFeatureOfBody();
     }
 
     Part::Feature::onChanged(prop);
--- a/src/Mod/PartDesign/App/FeatureBase.h
+++ b/src/Mod/PartDesign/App/FeatureBase.h
@@ -48,6 +48,9 @@ public:
     void onChanged(const App::Property* prop) override;
     App::DocumentObjectExecReturn* execute() override;
     void onDocumentRestored() override;
+
+private:
+    void trySetBaseFeatureOfBody();
 };
 
 } //namespace PartDesign


adrianinsaval:
This seems to work and IMO it is desired behavior, but could you explain to me what is going on the code here? It's not clear to me why `body->BaseFeature.getValue() != BaseFeature.getValue()` is meaningful, is this just a way of detecting a clone vs a normal basefeature? It's not clear to me why we need the out of scope warning for basefeatures either (or what even is the difference between clone and basefeature). btw I had made a forum post about this before too: https://forum.freecad.org/viewtopic.php?p=661409#p661409

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9565">pipeline status</a> for feature branch [PR_9565](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9565). Pipeline [903669771](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903669771) was triggered at [eb6623f](https://github.com/FreeCAD/FreeCAD/pull/9565/commits/eb6623f162b1b42ca71d3239aa03880393e426d1). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
Is there a point preventing to merge ?

adrianinsaval:
No I'm just hesitant to merge something that I don't understand at all but if you think this is good go ahead

Adrián Insaurralde
________________________________

0penBrain:
> No I'm just hesitant to merge something that I don't understand at all but if you think this is good go ahead

OK. I went to test it so. :)

@wwmayer I found 2 regressions that this code would introduce.

1/ Clone volatility

To reproduce:
* Create a Body with an additive box inside
* Select the box in the tree and create a PartDesign Clone
* Add a Part/Sphere
* Drag the Sphere to the "clone" Body => the Clone BaseFeature moved it's shape to the Sphere !!!

2/ Multiple different base features

To reproduce:
* Create a Body with an additive box inside
* Select the box in the tree and create a PartDesign Clone
* Add a Part/Sphere
* Add a BaseFeature with Python `fb = App.ActiveDocument.addObject("PartDesign::FeatureBase")`
* Move it to the "clone" Body `App.ActiveDocument.Body001.addObject(fb)` (adapt body name if needed)
* Change the `BaseFeature` property of the manually created BaseFeature so it is the Sphere => The "clone" Body has 2 different base features !!!

adrianinsaval:
1/ is no regresion at all, the same behavior can be observed in master and it seems like normal behavior, I think you are supposed to be able to set a basefeature by dragging the object to the body.
I'm not sure if 2/ could be considered a regression from this PR, it requires hacking around with python and it's not clear to me what you are saying is the difference in behavior in comparison to master.

0penBrain:
> 1/ is no regresion at all, the same behavior can be observed in master

Did you even tried ? The behavior is very different. With master, dragging the Sphere to the Body is rejected. With this PR, it is accepted and BaseFeature is changed.

> I'm not sure if 2/ could be considered a regression from this PR, it requires hacking around with python

My reproducing steps do not pretend to be the only way. Their only purpose is to demonstrate that the code changes may lead to a ambiguous situation

> it's not clear to me what you are saying is the difference in behavior in comparison to master.

If you do this with master, the 2 BaseFeature objects are kept in sync, so at the end the Body virtually has only one base feature. With code changes, they aren't anymore kept in sync and Body has 2 different base features.

adrianinsaval:
> Did you even tried ? The behavior is very different. With master, dragging the Sphere to the Body is rejected. With this PR, it is accepted and BaseFeature is changed.

my bad, I tried in 0.20 by mistake instead of master and the behavior is like that on 0.20, but not on master. So which should we consider a regression?

adrianinsaval:
> If you do this with master, the 2 BaseFeature objects are kept in sync, so at the end the Body virtually has only one base feature. With code changes, they aren't anymore kept in sync and Body has 2 different base features.

ok I see what you are saying, although ideally it shouldn't even be possible to add two basefeature objects into the same body.

0penBrain:
> my bad, I tried in 0.20 by mistake instead of master and the behavior is like that on 0.20, but not on master. So which should we consider a regression?

Past is past. :smile: I consider this changeset to be a regression because it brings a behavior that seems to me less expected than the current one. But subjective and open to discussion.



> ok I see what you are saying, although ideally it shouldn't even be possible to add two basefeature objects into the same body.

I think the good solution now is to open a general discussion if we want clone/basefeature to support or not being an internal feature of a body. Then we can implement a more global accordingly. I'm afraid now it's patches over patches, and not really sustainable anymore.

adrianinsaval:
> Past is past. 😄 I consider this changeset to be a regression because it brings a behavior that seems to me less expected than the current one. But subjective and open to discussion.

Dragging to set the BaseFeature property is documented behavior: https://wiki.freecad.org/PartDesign_Body#Base_Feature but I'm not quite sure if this was meant to happen only when BaseFeature wasn't already set or at any time. Personally, the latter makes more sense to me.

> I think the good solution now is to open a general discussion if we want clone/basefeature to support or not being an internal feature of a body

I'm not quite sure I understand what you are saying here. More than one BaseFeature does not make sense in any situation, whether they are in sync or not.

If you refer if we want to allow making a clone out of a specific feature in the tree from another body, yes! That's exactly why this PR exists. It already works quite well right now, the only problem is the annoying out of scope warning, which shouldn't be relevant in this case, because BaseFeature doesn't track the placement of the original solid and puts it in reference to the Body origin.

0penBrain:
> Dragging to set the BaseFeature property is documented behavior: https://wiki.freecad.org/PartDesign_Body#Base_Feature but I'm not quite sure if this was meant to happen only when BaseFeature wasn't already set or at any time. Personally, the latter makes more sense to me.

When it's a normal base feature, I don't really mind if you can change it or not by dragging a shape to the body (both behaviors may have pros and cons). But specifically when it's a clone, I think you shouldn't be able to overwrite it by dragging.

> If you refer if we want to allow making a clone out of a specific feature in the tree from another body, yes!

That's it. If we want that, maybe we will need deeper changes so we have no side-effects.

chennes:
@wwmayer what's your take on the status of this?

wwmayer:
> It's not clear to me why body->BaseFeature.getValue() != BaseFeature.getValue() is meaningful

If you look at the code you will see that this check is already in master. My change makes sure that the body's BaseFeature is only adjusted if it's already set to a value. This is to avoid that it's automatically set when e.g. restoring a project file where it wasn't set beforehand.

When you execute this code snippet you will see that the BaseFeature of Body is not set. Now, if you save the document and restore it you will realize that it's all the sudden set and this triggers the out-of.scope warnings.
```
doc = App.newDocument()

part = doc.addObject("PartDesign::Body", "Part")
box = doc.addObject("Part::Box", "Box")
part.Group = [box]

body = doc.addObject("PartDesign::Body", "Body")
base = doc.addObject("PartDesign::FeatureBase", "Clone")

base.BaseFeature = box
base.Placement = box.Placement
base.setEditorMode('Placement',0)

body.Group = [base]
body.Tip = base

doc.recompute()
```

> Did you even tried ? The behavior is very different. With master, dragging the Sphere to the Body is rejected. With this PR, it is accepted and BaseFeature is changed.

Adrian is right that this is not a regression and the behaviour depends on which master version you have used. If master is older than f4a8d94838 then the behaviour is different compared to after this change.

Before this change the creation of a clone did this:
```
doc = App.newDocument()

part = doc.addObject("PartDesign::Body", "Part")
box = doc.addObject("Part::Box", "Box")
part.Group = [box]

body = doc.addObject("PartDesign::Body", "Body")
base = doc.addObject("PartDesign::FeatureBase", "Clone")

base.BaseFeature = box
base.Placement = box.Placement
base.setEditorMode('Placement',0)

body.Group = [base]
body.Tip = base

doc.recompute()
```
And now it does
```
doc = App.newDocument()

part = doc.addObject("PartDesign::Body", "Part")
box = doc.addObject("Part::Box", "Box")
part.Group = [box]

body = doc.addObject("PartDesign::Body", "Body")
base = doc.addObject("PartDesign::FeatureBase", "Clone")

body.Group = [base]
body.Tip = base

base.BaseFeature = box
base.Placement = box.Placement
base.setEditorMode('Placement',0)

doc.recompute()
```

PR #9574, opened 2023-05-14
Sketcher: fix Circle to Line distance constraint display

FlachyJoe:
Fix Circle to Line distance constraint moving in the opposite direction when dragged.
https://forum.freecad.org/viewtopic.php?t=78244 first point

NB : the label still jump away when you start to drag but I don't understand why and how to fix this.


src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
```diff
@@ -677,16 +677,13 @@ void EditModeConstraintCoinManager::processConstraints(const GeoListFacade & geo
...
-                                        Base::Vector3d dir = pnt2;
+                                        pnt1.ProjectToLine(ct-l2p1, l2p2-l2p1); //project on the line translated to origin
+                                        Base::Vector3d dir = pnt1;
                                         dir.Normalize();
-                                        pnt1 = ct + dir * radius;
-                                        pnt2 += ct;
+                                        pnt2 = ct + dir * radius;
+                                        pnt1 += ct;
                                         Base::Vector3d d = l2p2 - l2p1;
                                         double ActDist = std::abs(-ct.x*d.y+ct.y*d.x+l2p1.x*l2p2.y-l2p2.x*l2p1.y) / d.Length() - radius;
```
0penBrain, line 686: As you can see from the CI, this variable is no more needed (or if it is, there is a problem. ;)

src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
```diff
@@ -677,16 +677,13 @@ void EditModeConstraintCoinManager::processConstraints(const GeoListFacade & geo
...
-                                        if (ActDist < 0) {
-                                            std::swap(pnt1, pnt2);
-                                        }
```
0penBrain, line 689: What was the original goal of this ? Isn't there any risk to remove it ?

0penBrain (commented):
Some remarks. Gonna test it.

Very impressed. I was actually debugging the problem and found myself almost lost in the VPSketch. If it works, it's brilliant.

BTW, could you update comment of this line to add it also concerns circle to line ? https://github.com/FreeCAD/FreeCAD/blob/c2fa4793b76ea59c2eabfe00c7b1a51da292498f/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp#L1388

0penBrain:
Well, not really solved actually (sorry).

There is always a significant offset between mouse cursor and actual value position.

Also I found that it may depends how line and circle are relatively located.

See here with horizontal line it works pretty OK (except offset):
hor

But here with vertical line it still fails:
ver

I didn't investigate more, it may not be due to line actually being horizontal or vertical, but to more complex context.

I think you'll have to play with VPSketch (around the line I already pointed out). :wink:

0penBrain:
I moved the PR to Draft so nobody may merge it ATM. @FlachyJoe feel free to set it back "Ready for review" when you think it's OK.

FlachyJoe:
> I think you'll have to play with VPSketch

You're still right I missed this file in my dev notes.

0penBrain:
@FlachyJoe Code is good to me. Can you squash the 2 commits together ? Thx

FlachyJoe:
@0penBrain done, merci pour le soutien ;-)

0penBrain:
> @0penBrain done, merci pour le soutien ;-)

De rien. Bien joué.

PR #9575, opened 2023-05-14
Change guessing algorithm of Rotation::makeRotationByAxes()

### LINKED ISSUE #9414: [Feature Request] Change guessing algorithm of Rotation::makeRotationByAxes()

### Forums discussion

https://forum.freecad.org/viewtopic.php?t=77092

### Subproject(s) affected?

Core

### Idea description

The wiki for [OXY attachment](https://wiki.freecadweb.org/O-X-Y_type_attachment_modes) describes how the second direction is guessed when omitted. I don't think the guessing process is very good and have room for improvement but I want your opinion before changing.
I started to look into this after reading a [post](https://forum.freecad.org/viewtopic.php?p=656069#p656069) by @ppemawm and the wiki by @edwilliams16.
I like to suggest that the "hint" direction instead uses the local axis corresponding to the last character. This is true now for OXZ, OYZ and OZY attachments. But not true for OXY, OYX and OZX. 

To achieve this I planned to target the rotation constructor Rotation(vX, vY, vZ, orderString) in Rotation.cpp which is used by the Attachment engine. I realize the change in some cases may brake models but I think it will make it more intuitive and beneficial for future works.

If the orderString is for example "YZX" the current implementation works like this to find a rotated CS:
- if vY (first character in "YZX") is not a zero vector, lock the target Y axis to this direction
- else try the same for vZ and then vX
- look for a hint direction in remaining vectors
- if no hint direction is found, use a hard coded local axis as hint depending on the main axis as described in the wiki link above

I would like to change the behavior to this:
- if vY is not a zero vector, lock the target Y axis to this direction
- else lock the target Y axis to the local Y
- if vZ (second character in "YZX") is not a zero vector and not equal to main direction, use as a hint for the new Z axis
- else use the local Z as hint direction
- the third vector is never used

I suggest that a `orderString` with only two characters is accepted too.

### Anything else?

_No response_

### Code of Conduct

Jolbas:
Fix #9414

Also follows the description in header file better by ignoring the third vector compleatly
This forum thread is also related: https://forum.freecad.org/viewtopic.php?t=78207

### DIFF
--- a/src/Base/Rotation.cpp
+++ b/src/Base/Rotation.cpp
@@ -532,159 +532,72 @@ Rotation
 Rotation::makeRotationByAxes(Vector3d xdir, Vector3d ydir, Vector3d zdir, const char* priorityOrder)
 {
     const double tol = Precision::Confusion();
-    enum dirIndex
-    {
-        X,
-        Y,
-        Z
-    };
-
-    // convert priorityOrder string into a sequence of ints.
-    if (strlen(priorityOrder) != 3) {
-        THROWM(ValueError, "makeRotationByAxes: length of priorityOrder is not 3");
+    // check priorityOrder and convert characters to integers.
+    int orderLen = strlen(priorityOrder);
+    if (orderLen < 2 || orderLen > 3) {
+        THROWM(ValueError, "makeRotationByAxes: length of priorityOrder is not 2 or 3");
     }
-    int order[3];
-    for (int i = 0; i < 3; ++i) {
-        order[i] = priorityOrder[i] - 'X';
-        if (order[i] < 0 || order[i] > 2) {
+    size_t mainDirIndex, hintDirIndex, lastDirIndex, parsedDir;
+    for (int i = 0; i < orderLen; ++i) {
+        parsedDir = priorityOrder[i] - 'X';
+        if (parsedDir < 0 || parsedDir > 2) {
             THROWM(ValueError,
                    "makeRotationByAxes: characters in priorityOrder must be uppercase X, Y, or Z. "
-                   "Some other character encountered.")
+                   "Some other character encountered.");
+        }
+        if ((i > 0 && parsedDir == mainDirIndex) || (i > 1 && parsedDir == hintDirIndex)) {
+            THROWM(ValueError,
+                   "makeRotationByAxes: there cannot be two identical characters in priorityOrder");
+        }
+        if (i == 0) {
+            mainDirIndex = parsedDir;
+        }
+        else if (i == 1) {
+            hintDirIndex = parsedDir;
         }
     }
-
-    // ensure every axis is listed in priority list
-    if (order[0] == order[1] || order[1] == order[2] || order[2] == order[0]) {
-        THROWM(ValueError, "makeRotationByAxes: not all axes are listed in priorityOrder");
-    }
-
-
-    // group up dirs into an array, to access them by indexes stored in @order.
+    lastDirIndex = 3 - mainDirIndex - hintDirIndex;
+    // group up dirs into an array, to access them by indexes stored in order.
     std::vector<Vector3d*> dirs = {&xdir, &ydir, &zdir};
 
-
-    auto dropPriority = [&order](int index) {
-        int tmp {};
-        if (index == 0) {
-            tmp = order[0];
-            order[0] = order[1];
-            order[1] = order[2];
-            order[2] = tmp;
+    auto pickDir = [&](size_t direction) {
+        Vector3d v = *(dirs[direction]);
+        if (v.Length() < tol) {
+            v = Vector3d();
+            v[direction] = 1.0;
         }
-        else if (index == 1) {
-            tmp = order[1];
-            order[1] = order[2];
-            order[2] = tmp;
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9575">pipeline status</a> for feature branch [PR_9575](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9575). Pipeline [903669775](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903669775) was triggered at [1e19a28](https://github.com/FreeCAD/FreeCAD/pull/9575/commits/1e19a28fb7d33f0fa6c17cc5fff0a8d40f6fb4d0). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
@edwilliams16 this came from a discussion that you and @Jolbas had in Issue #9414 -- do you have any further input on whether this should be merged or not? I'm not sure the issue meets the bar set in CONTRIBUTING.md for an "identified and agreed-on problem," and the related forums discussion didn't get a lot of feedback either. That said, the new algorithm does have the virtue of being considerably simpler than the old one, so from a code-review perspective this is an improvement.

edwilliams16:
@chennes  I think this issue is still being litigated. See [https://foru(m.freecad.org/viewtopic.php?t=80318](url)
The current API for App.Rotation(vecx, vecy, vecz, ‘XYZ’) can specify a general rotation.  The decision was apparently made to provide ‘sensible’ fallback for invalid arguments - when direction vectors have zero length or when orthogonal axes are impossibly mapped to ones that are parallel.  These fallbacks are somewhat obscure. I figured them out by experiment and attempted to document them in the wiki.
@jolbas, in this PR, has proposed fallbacks that are arguably more logical.

The downside of the PR, IMO, is there is very likely code out there that uses the fallbacks that will break and may be quite tricky to diagnose and fix.

An alternative is to provide an additional API that provides similar capability but errors out with invalid input, forcing the code author to deal with these error cases explicitly should they arise.  The discussion referenced above concerns what that might look like.  I prefer this approach - but there is yet to be agreement as to the cleanest way to do it.

Jolbas:
Regarding old documents and the attachment extension this will have an effect on type OXY, OYX and OZX where only 2 references are present. Those cases currently behaves as OXZ, OYZ and OZY respectively. It would be possible to catch those cases in `AttachExtension::onExtendedDocumentRestored()` and change them to the type that won't change when documents with ProgramVersion less than 0.22 are opened. That way we can be backwards compatible regarding documents.

With python code and macros using Rotation(vecX, vecY, vecZ, sequence) this won't be possible since there is no document version. There is several possible cases that will be affected but never when the vectors corresponding to the first two characters is supplied and not null. The current implementation falls back to the the next vector in sequence despite the documentation says the third character in sequence is ignored. All use cases I could find passes a null vector at the position corresponding to the third character and that leads to unexpected guesses when main direction is null.

For example if Rotation(vec1, vec2, nullVec, "XYZ") is called and vec1 happens to be null vector: 
The current implementation will align Y to vecY and Z to (1,0,0)!!
This new implementation will align X to (1,0,0) and Y to vecY.

If this PR is still a candidate to merge I'm willing to go through the FreeCAD and FreeCAD-macro repositorys and look for unambiguous code. A quick look suggests the function is not used in this form at FreeCAD-macro but in FreeCAD source I found some code that should be updated. For example:
https://github.com/FreeCAD/FreeCAD/blob/3e164a570068bcd91f7320db033efbd8c4f9514a/src/Mod/Arch/ArchFrame.py#L216
If we close this, then tooltip and documentation should be updated to not say the third vector is ignored

Jolbas:
The main change in c++ and python interface is that the sequence argument is no longer completely ignored if only one vector is provided

luzpaz:
bumping for status

Jolbas:
We haven't got any further with this. I have seen new people struggling with this on the forum but I have no solution yet without breaking backwards compatibility. I like to go on working with this but have currently and until end of year no time to spend on FreeCAD development. One possible first step solution could be to identify and rename current attachments with just two references to OX, OY and OZ since the last letter have no effect anyway.

maxwxyz:
@Jolbas how is this PR going, could we ping someone for help or did you revisit the last ideas?

Jolbas:
My day job has continued to be overwhelming this spring so I haven't had the opportunity to make this the ultimate solution. However, I can find time to update the wiki and tooltips and of cause resolve the conflicts if you think the behaviour change acceptable. I will also take time to keep an eye on the forum for users struggeling with it.

maxwxyz:
@Jolbas any chance you could revisit this PR?

Jolbas:
@maxwxyz Yes. It's just that I'm not sure what conclusions to draw from the conversation here and how to proceed. I should add some tests of course and maybe clarify the documentation further. But other than I think that I myself am satisfied as it is.

maxwxyz:
So if you are ready, change the status to ready for review. Currently the PR is in draft.

PR #9583, opened 2023-05-15
Save all decimals when property is edited

### LINKED ISSUE #9570: [Problem] Position/Angle placement with Euler too unprecise

### Version

0.21 (Development)

### Full version info

```shell
OS: Debian GNU/Linux 11 (bullseye) (X-Cinnamon/lightdm-xsession)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33177 (Git)
Build type: Release
Branch: master
Hash: 3c8ee2213f85b056a742a7e386620d8d2fc0abf9
Python 3.9.2, Qt 5.15.2, Coin 4.0.0, Vtk 9.0.1, OCC 7.7.1
Installed mods: 
  * QuickMeasure 2022.10.28
  * Assembly4 0.12.6
  * CurvedShapes 1.0.4
  * Curves 0.6.8
```

### Subproject(s) affected?

Core

### Problem description

see https://forum.freecad.org/viewtopic.php?t=76461

When more then one turning angle is included, while transforming (transform operation or via placement in the property panel), the position and angle values get too unprecise to work. This can affect to boolean operations, assemblies, as well later for TechDraw deviations. Also how much unprecise it gets depends on the digits settings.

[mwe.zip](https://github.com/FreeCAD/FreeCAD/files/11472640/mwe.zip)

[2023-02-28 12-39.webm](https://github.com/FreeCAD/FreeCAD/assets/109620914/cab04307-c194-4c61-992a-eff070d735c4)

### Anything else?

_No response_

### Code of Conduct

Jolbas:
---
When editing from property view or a properties associated task panel. Save all interpreted decimals to the internal property.
Fix #9570

### DIFF
--- a/src/Gui/propertyeditor/PropertyItem.cpp
+++ b/src/Gui/propertyeditor/PropertyItem.cpp
@@ -1044,7 +1044,7 @@ void PropertyFloatItem::setValue(const QVariant& value)
         if (!value.canConvert<double>())
             return;
         double val = value.toDouble();
-        QString data = QString::fromLatin1("%1").arg(val, 0, 'f', decimals());
+        QString data = QString::fromLatin1("%1").arg(val, 0, 'g', 16);
         setPropertyValue(data);
     }
 }
@@ -1113,7 +1113,7 @@ void PropertyUnitItem::setValue(const QVariant& value)
             return;
         const Base::Quantity& val = value.value<Base::Quantity>();
 
-        Base::QuantityFormat format(Base::QuantityFormat::Fixed, decimals());
+        Base::QuantityFormat format(Base::QuantityFormat::Default, 16);
         QString unit = Base::UnitsApi::toString(val, format);
         setPropertyValue(unit);
     }
@@ -1220,7 +1220,7 @@ void PropertyFloatConstraintItem::setValue(const QVariant& value)
         if (!value.canConvert<double>())
             return;
         double val = value.toDouble();
-        QString data = QString::fromLatin1("%1").arg(val, 0, 'f', decimals());
+        QString data = QString::fromLatin1("%1").arg(val, 0, 'g', 16);
         setPropertyValue(data);
     }
 }
@@ -1427,9 +1427,9 @@ void PropertyVectorItem::setValue(const QVariant& value)
         return;
     const Base::Vector3d& val = value.value<Base::Vector3d>();
     QString data = QString::fromLatin1("(%1, %2, %3)")
-                    .arg(val.x,0,'f',decimals())
-                    .arg(val.y,0,'f',decimals())
-                    .arg(val.z,0,'f',decimals());
+                    .arg(val.x,0,'g',16)
+                    .arg(val.y,0,'g',16)
+                    .arg(val.z,0,'g',16);
     setPropertyValue(data);
 }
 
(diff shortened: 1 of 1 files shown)


adrianinsaval:
Is the display value still rounded? If so, is the precision not lost when editing the property again?

Jolbas:
@adrianinsaval Yes the display value is still rounded. The other question need some elaboration. The precision isn't lost if the value isn't changed, but the value is still rounded after you started an editing action. I would like the full precision except trailing zeroes to be shown when editing. But when done editing it goes back to rounding according to user preferences. I also would like to be able to add more decimals than user preferences in float properties. It is possible on properties with an unit.

AjinkyaDahale:
Perhaps there should be an "editing_precision" vs "display_precision".

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9583">pipeline status</a> for feature branch [PR_9583](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9583). Pipeline [903671162](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903671162) was triggered at [a1c5a61](https://github.com/FreeCAD/FreeCAD/pull/9583/commits/a1c5a619ca6fd32646e41479069a39273ce32631). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

chennes:
@AjinkyaDahale Are you OK with the being merged as-is, possibly pending some future work in this area?

AjinkyaDahale:
> @AjinkyaDahale Are you OK with this being merged as-is, possibly pending some future work in this area?

I have no objection w.r.t. my comment. The changes here do not seem to make matters worse than they currently are, though of course storing floats in decimals has its antics (like `12.01` getting stored as `12.00999999999994` or something similar).

Jolbas:
> The changes here do not seem to make matters worse than they currently are, though of course storing floats in decimals has its antics (like 12.01 getting stored as 12.00999999999994 or something similar).

That is unlikely to happen as the values are still rounded to 16 significant figures.

AjinkyaDahale:
> > storing floats in decimals has its antics (like 12.01 getting stored as 12.00999999999994 or something similar).
> 
> That is unlikely to happen as the values are still rounded to 16 significant figures.

There should be exactly 16 digits in that number I wrote :). This issue happens because of how floats are stored in binary. What is a neatly ending decimal can very well have a recurring binary notation, which is truncated. When printing with all the decimal digits, this truncation can give different numbers.

To be clear, I don't see this as being a problem for the most part. This is just a limitation of computers.

chennes:
OK, I'm going to merge this before we make a release candidate then.

Jolbas:
@AjinkyaDahale @chennes  Did some testing now.
If rounding to 17 significant digits the the mentioned decimal->binary->decimal problem occurs on about 83% of the input numbers
When rounding to 16 digits it is reduced to 9% of the tested inputs. Never when the first digit is 4 or lower
With 15 significant digits the problem never occurs.
I tested to print fractional numbers created from strings with 1 to 4 significant digits and then formatted with the g format specifier and checked if the length was longer than necessary.
What do you think of using 15 instead of 16?

PR #9593, opened 2023-05-16
Draft: ShapeString: use FaceMakerBullseye instead of current face algo

Roy-043:
The current ShapeString face algorithm cannot handle characters with nested islands such as "©". Instead of fixing that code it made more sense to use `Part::FaceMakerBullseye` instead.

In almost all cases using `Part::FaceMakerBullseye` will work, else the code will use `Part::FaceMakerCheese` or, if  that fails, `Part::FaceMakerSimple`.

The `execute` function was also improved a little bit.

Forum topic:
https://forum.freecad.org/viewtopic.php?t=77733

---

### DIFF
--- a/src/Mod/Draft/draftobjects/shapestring.py
+++ b/src/Mod/Draft/draftobjects/shapestring.py
@@ -63,21 +63,15 @@ class ShapeString(DraftObject):
             self.props_changed_clear()
             return
 
-        import Part
-        # import OpenSCAD2Dgeom
         if obj.String and obj.FontFile:
+            import Part
+
             if obj.Placement:
                 plm = obj.Placement
 
-            CharList = Part.makeWireString(obj.String,obj.FontFile,obj.Size,obj.Tracking)
-            if len(CharList) == 0:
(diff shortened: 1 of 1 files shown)


luzpaz:
@Roy-043 is there an open ticket associated with this PR ?

yorikvanhavre:
Good move. Do you think it would make sense to expose the FaceMaker algo to the user? For ex as a PropertyEnum. So one could tweak manually if the defalt algo does not work well, even if it didn't fail

Roy-043:
I would not expose the FaceMaker. It would probably be very rare that FaceMakerBullseye fails. And if it fails the code will try 2 alternatives for each character.

The bad font I have used for testing (1CamBam_Stick_0.ttf) has issues with all FaceMakers depending on the characters in the string. So if the FaceMaker is exposed there should also be an 'Auto' option that tries all 3 FaceMakers as in my proposal.

yorikvanhavre:
Yes... Well we can go step by step and see how this one behaves before seeing if there's a need to expose

PR #9596, opened 2023-05-17
Gui: Mac version info hack

chennes:
Fixes https://github.com/FreeCAD/FreeCAD/issues/9481

---

- [X]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

PR #9603, opened 2023-05-18
Gui: Hack to get Windows 11 version

chennes:
As promised in #9596, here's the equivalent hack to get the proper Windows 11 version based on the current build number.

### DIFF
--- a/src/Gui/Splashscreen.cpp
+++ b/src/Gui/Splashscreen.cpp
@@ -34,6 +34,7 @@
 # include <QRegularExpression>
 # include <QRegularExpressionMatch>
 # include <QScreen>
+# include <QSettings>
 # include <QSysInfo>
 # include <QTextBrowser>
 # include <QTextStream>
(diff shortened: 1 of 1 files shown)


adrianinsaval:
is this also fixed in qt6 then?

adrianinsaval:
Also looking at the upstream issue I think the fix for mac is in qt 6.5 not 6.0, I didn't see anything about windows
https://bugreports.qt.io/browse/QTBUG-111114
https://codereview.qt-project.org/c/qt/qtbase/+/460158

Would windows 7 still be reported correctly with this method?

chennes:
This should only affect 64 bit Windows, so I think Win7 will be unaffected.

adrianinsaval:
Windows 7 64bit is widely used, possibly more so than the 32bit version. Not that I think anyone should use windows 7 but people do it

chennes:
OK, I pushed a more thorough hack that should get us back to Win7.

adrianinsaval:
unless we have reason to believe that this issue has been fixed in qt6 wouldn't it be better to put the windows if outside the qt6 if?

PR #9604, opened 2023-05-18
Sketcher: Obey non-intrusive preferences for modal pop-ups

abdullahtahiriyo:
==========================================================

This commit turns the remaining blocking pop-ups in which the user can only click "ok" into non-intrusive notifications, depending on the user preferences in settings.


abdullahtahiriyo:
This PR is currently meant as a working implementation for the discussion in:
https://forum.freecad.org/viewtopic.php?p=682600#p682600

While making pop-ups follow (non-intrusive) preferences is only logical, doing so raises some questions that should be answered to take the best approach.

PR #9605, opened 2023-05-18
Sketcher PlaneGCS clang-format and pre-commit

abdullahtahiriyo:
A first PR for Sketcher seeking to format the code and making use of the pre-commit framework.


abdullahtahiriyo:
@chennes 

Thank you for your framework.

I have installed pre-commit as per:
https://forum.freecad.org/viewtopic.php?t=77205
https://github.com/FreeCAD/DevelopersHandbook/blob/master/gettingstarted/index.md#setting-up-for-development

This is a first approach with the smallest directory we can migrate at this time.

I have added the directory to pre-commit as per the instructions. 

However, I am unsure about:
```As they have the time and opportunity to do so, Maintainers run clang-format and/or Black as appropriate on the code they are responsible for, in coordination with the other developers working on that code. The commits introducing those formatting changes are added to .git-blame-ignore-revs.```

My understanding is that we need to merge this PR first, then get the commit hash in master, and then do another PR. Am I right?

Let me know if something does not add up.

**Comment by chennes** [comment]
If we merge the PR via a merge commit (rather than a rebase and merge) then the commit hash will stay the same and you can just add the hash to the ignore file in this PR. If you'd prefer to merge with a rebase and merge, then the hash will change, and you need to do it in two steps.

File: .pre-commit-config.yaml [hunk 1]
```diff
@@ -7,6 +7,7 @@ files: |
         src/Mod/AddonManager|
         src/Tools|
         tests/src
```
chennes, line 9: You need a vertical pipe at the end of the previous line.

abdullahtahiriyo: Thanks for noticing.

abdullahtahiriyo:
> If we merge the PR via a merge commit (rather than a rebase and merge) then the commit hash will stay the same and you can just add the hash to the ignore file in this PR. If you'd prefer to merge with a rebase and merge, then the hash will change, and you need to do it in two steps.

I am the rebase and merge type. I can do it in a separate PR. Thanks.

abdullahtahiriyo:
Clang-format, with the current settings, reorders the headers (within a block of headers). Many Python binding files XXXPyImp.cpp, have the two includes (cpp and h) in a row:

```
#include "SketchObjectSFPy.h"
#include "SketchObjectSFPy.cpp"
```
Reordering them causes the compilation process to fail.

The solution I found is to leave a blank line to avoid the reorder:
```
#include "SketchObjectSFPy.h"

#include "SketchObjectSFPy.cpp"
```

I am sure this will be widely known. It was not for me, so I comment it here. Perhaps it is useful for somebody, or somebody knows a better/preferred alternative way.

PR #9613, opened 2023-05-20
Translate import export names

chennes:
Many of the importer/exporter strings that we display to users contain translatable elements ("STEP with colors", "Point formats", etc.). This PR exposes any importer/exporter text strings that contain potentially-translatable elements to translation, adding them to the "FileFormat" context. They are translated at point-of-use in the GUI when the dialog is presented. This fixes https://github.com/FreeCAD/FreeCAD-translations/issues/217

---

### DIFF
--- a/src/App/FreeCADInit.py
+++ b/src/App/FreeCADInit.py
@@ -30,6 +30,10 @@
 # imports the one and only
 import FreeCAD
 
+
+def QT_TRANSLATE_NOOP(_, txt):
+    return txt
+
 def removeFromPath(module_name):
     """removes the module from the sys.path. The entry point for imports
         will therefore always be FreeCAD.
@@ -690,7 +694,7 @@ except Exception as e:
     Err(traceback.format_exc())
     Err('-'*80+'\n')
 
-FreeCAD.addImportType("FreeCAD document (*.FCStd)","FreeCAD")
+FreeCAD.addImportType(QT_TRANSLATE_NOOP("FileFormat","FreeCAD document (*.FCStd)"), "FreeCAD")
 
 # set to no gui, is overwritten by InitGui
 App.GuiUp = 0
--- a/src/Gui/FileDialog.cpp
+++ b/src/Gui/FileDialog.cpp
@@ -882,7 +882,9 @@ SelectModule::SelectModule (const QString& type, const SelectModule::Dict& types
             module = module.left(match.capturedStart());
         }
 
-        button->setText(QString::fromLatin1("%1 (%2)").arg(filter, module));
+        auto filterText = QObject::tr(filter); // Filter should be in the QObject context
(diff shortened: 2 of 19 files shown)


chennes:
I considered that, but on the other hand this provides valuable context to the translators. @kaktusus what do you think?

kaktusus:
I have read what Yorik wrote are valuable comments about how the mechanism was created.

What I don't like is that based on the string the program takes various actions that may fail as presented by Yorik.
I propose to attach to each string some internal identifier on the basis of which the program will take an action. This will not disturb operation of the program if the translator introduces nonsense into the string or makes a mistake.

 In this way, the string will be able to contain full information also the file extension and the program will be able to function correctly regardless of the content of the string.

chennes:
OK, I will rework this so that mistakes from translators can't break it.

yorikvanhavre:
> I propose to attach to each string some internal identifier on the basis of which the program will take an action. This will not disturb operation of the program if the translator introduces nonsense into the string or makes a mistake.

The problem is that this system is done by Qt, not by us, we have no control over it. We would need to recode a file dialog system ourselves, which is of course much hassle. Too bad there is no way to add comments in a translate() call...

chennes:
> Too bad there is no way to add comments in a translate() call...

There is - on the line before a translate call you can add a comment that starts with `//:` -- that gets extracted by lupdate. I don't know how CrowdIn shows it though.

yorikvanhavre:
Oh great to know! That could be VERY useful.. If you are reformatting this PR, could you maybe add a couple so we can test?

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9613">pipeline status</a> for feature branch [PR_9613](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9613). Pipeline [903671168](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903671168) was triggered at [7f019ce](https://github.com/FreeCAD/FreeCAD/pull/9613/commits/7f019ce2551e19dd827b89f956ad035c56f5a87a). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

maxwxyz:
@chennes any news on this PR, the original mentioned issue was closed last year?
Otherwise you know the date: _ready for review_ by **Monday, 2024-06-03** or maybe in this case 2024-06-24 (UI) to be considered for 1.0

maxwxyz:
@chennes is your PR still relevant? The mentioned issue was fixed already and there is also https://github.com/FreeCAD/FreeCAD/pull/16417

chennes:
I'm just going to re-work this.

PR #9615, opened 2023-05-20
Gui: Don't show raw exception to user

chennes:
As discussed in https://github.com/FreeCAD/FreeCAD-translations/issues/184

---


abdullahtahiriyo:
This approach is acceptable in all cases where the string is untranslatable.

When the exception comes from Python (because the interpreter was called and a Python exception was generated), the Interpreter reported the exception to the report view.

When the exception does not come from Python, perhaps we should consider on a case by case basis if it makes sense to report the exception to the report view.

```e.ReportException()``` should do it.

However, that function will report only to the Developer after this other PR is merged:
https://github.com/FreeCAD/FreeCAD/pull/9652

as it includes a feature fix for the Console Extension Framework just merged into master.

Now it will report to the notification area too.

PR #9633, opened 2023-05-22
Surface: allow instant update of BlendCurve properties

tomate44:
Hello,
Since Surface::BlendCurve feature doesn't compute any automatic initial property values, I think it is better to allow live update of properties, for easier tweaking.
Thanks

---

### DIFF
--- a/src/Mod/Surface/App/Blending/FeatureBlendCurve.cpp
+++ b/src/Mod/Surface/App/Blending/FeatureBlendCurve.cpp
@@ -48,19 +48,19 @@ PROPERTY_SOURCE(Surface::FeatureBlendCurve, Part::Spline)
 
 FeatureBlendCurve::FeatureBlendCurve()
 {
-    ADD_PROPERTY_TYPE(StartEdge, (nullptr), "FirstEdge", App::Prop_None, "");
-    ADD_PROPERTY_TYPE(StartContinuity, (2), "FirstEdge", App::Prop_None, "");
+    ADD_PROPERTY_TYPE(StartEdge, (nullptr), "FirstEdge", App::Prop_None, "Edge support of the start point");
+    ADD_PROPERTY_TYPE(StartContinuity, (2), "FirstEdge", App::Prop_None, "Geometric continuity at start point");
     StartContinuity.setConstraints(&StartContinuityConstraint);
-    ADD_PROPERTY_TYPE(StartParameter, (0.0f), "FirstEdge", App::Prop_None, "");
+    ADD_PROPERTY_TYPE(StartParameter, (0.0f), "FirstEdge", App::Prop_None, "Parameter of start point along edge");
     StartParameter.setConstraints(&StartParameterConstraint);
-    ADD_PROPERTY_TYPE(StartSize, (1.0f), "FirstEdge", App::Prop_None, "");
+    ADD_PROPERTY_TYPE(StartSize, (1.0f), "FirstEdge", App::Prop_None, "Size of derivatives at start point");
 
-    ADD_PROPERTY_TYPE(EndEdge, (nullptr), "SecondEdge", App::Prop_None, "");
-    ADD_PROPERTY_TYPE(EndContinuity, (2), "SecondEdge", App::Prop_None, "");
+    ADD_PROPERTY_TYPE(EndEdge, (nullptr), "SecondEdge", App::Prop_None, "Edge support of the end point");
+    ADD_PROPERTY_TYPE(EndContinuity, (2), "SecondEdge", App::Prop_None, "Geometric continuity at end point");
     EndContinuity.setConstraints(&EndContinuityConstraint);
-    ADD_PROPERTY_TYPE(EndParameter, (0.0f), "SecondEdge", App::Prop_None, "");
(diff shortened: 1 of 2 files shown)


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9633">pipeline status</a> for feature branch [PR_9633](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9633). Pipeline [903660554](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903660554) was triggered at [1b0874d](https://github.com/FreeCAD/FreeCAD/pull/9633/commits/1b0874ddbad6d3e70a8126c0c24739173fe2de7e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

luzpaz:
@wwmayer do you mind weighing in on this ?

sliptonic:
Maintainers are struggling with whether to merge this .   @tomate44  can you please provide some more context on why this needs to be removed?  @wwmayer Do you see what's at stake with this PR?

tomate44:
Hello,
We used other features of the module as model, like [FeatureExtend](url=https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Surface/App/FeatureExtend.cpp#L153).
I don't know if this code block with really needed from the start.
But the instant update doesn't work if this Lock is present.
The goal is to allow instant update of the shape when editing a property, like a Part::Box primitive does.

wwmayer:
> We used other features of the module as model, like FeatureExtend.
I don't know if this code block was really needed from the start.

Actually not because there won't occur an infinite recursion but there is only up to one recursion if the value set to StartContinuity or EndContinuity isn't correct.

So, it's fine to entirely remove the "mutex" thing. Nevertheless, the code has some flaws which by accident could indeed lead to an infinite recursion. To avoid such an accidental change you should make the code more robust:

Instead of
```
        if (changedStartProp->getValue() > (maxDegree - 2 - EndContinuity.getValue())) {

            StartContinuity.setValue(maxDegree - 2 - EndContinuity.getValue());
        }
```
write
```
        long maxValue = maxDegree - 2 - EndContinuity.getValue();
        if (changedStartProp->getValue() > maxValue) {

            StartContinuity.setValue(maxValue);
        }
```

Then with `if (prop == &StartContinuity)` you already know that prop is the StartContinuity instance which makes the dynamic_cast of prop completely superfluous. Instead you should directly use StartContinuity to get the value.

So, the function should be simplified to
```

void FeatureBlendCurve::onChanged(const App::Property *prop)
{
    if (prop == &StartContinuity) {
        long maxValue = maxDegree - 2 - EndContinuity.getValue();
        if (StartContinuity.getValue() > maxValue) {
            StartContinuity.setValue(maxValue);
        }
    }
    else if (prop == &EndContinuity) {
        long maxValue = maxDegree - 2 - StartContinuity.getValue();
        if (EndContinuity.getValue() > maxValue) {
            EndContinuity.setValue(maxValue);
        }
    }
    Part::Spline::onChanged(prop);
}
```

> The goal is to allow instant update of the shape when editing a property, like a Part::Box primitive does.

The basic question is how fast or slow the recompute will be. If it's not super fast you shouldn't do an automatic recompute as it will easily become a serious bottleneck.

tomate44:
Thank you.
I'll check the needed fixes.

> The basic question is how fast or slow the recompute will be. If it's not super fast you shouldn't do an automatic recompute as it will easily become a serious bottleneck.

I don't think it is heavy. Is there a way I can check this ?
Is there a way to measure and report recompute time in C++ ?
I would compare with Part::Box

tomate44:
I didn't find the Part::Box file.
But I did some speed measurement of execute method against the 3 Part Primitives that have live-update properties (min - max) : 
Cone : 0.005 - 0.01
Cylinder : 0.001 - 0.003
Plane : 0.0004 - 0.0008
BlendCurve : 0.0003 - 0.0007

So BlendCurve is on par with Part::Plane. No speed problem.
I'll push the needed changes tomorrow.

wwmayer:
> I'll push the needed changes tomorrow.

OK.

wwmayer:
> I'll check the needed fixes.

Am I wrong or doesn't your pushed PR change anything?

tomate44:
I had 3 small commits ready.
But I had trouble rebasing the branch on my work PC.
So I pushed the "Update Branch" button on this page.
I am currently trying to solve this problem here at home ...

tomate44:
The branch seems up-to-date now.
So I'll be able to redo my commits tonight or tomorrow.

wwmayer:
OK

tomate44:
Last commit ("[cleanup properties instantiation](https://github.com/FreeCAD/FreeCAD/pull/9633/commits/a10f0b46b139f8889bffa40191bd129efc0ec7ee)") changes the property type of the 2 size properties from PropertyFloat to PropertyFloatConstraint to allow a 0.1 step that is much more practical than a 1.0 step.
I checked with a file saved before commit, and its size properties get automatically converted to PropertyFloatConstraint, so backward compatibility seems preserved.

PR #9634, opened 2023-05-22
[Base/Gui/Sketcher] Console improvement and application to Sketcher WB

abdullahtahiriyo:
This PR proposes an extension of the console interface intended to fix the limitations identified here:
https://github.com/FreeCAD/FreeCAD-translations/issues/223
https://forum.freecad.org/viewtopic.php?t=77673

In essence:
1. The console framework is extended to include additional information, the intended recipient and the translation status of the message. The TranslatedNotification message is removed, as it is covered by a notification where the translation status is "translated".
2. The notifications framework is adapted to the new console framework.
3. Base interpreter is adapted to report the exception with intended recipients being only Developers
4. The Notification Area is adapted. It won't show a message that is intended for developers or that is untranslatable (such a stack), unless in preferences the user has opted-in to receive developer messages.
5. The Notification Area preferences are adapted. While it is possible to opt-in developer warnings and errors, they are not show by default. User intended warnings and errors are an integral part part of the notification area and it is no longer possible to opt-out. Of course, it is always possible to disable the notification area as before.
6. The Report view will show errors and warnings intended for developers and not translated, i.e. in English. It won't show errors intended exclusively to users or messages that are translated.
7. The Sketcher Constraint Commands is adapted to the new framework and may serve as a reference for other portions of the code.

Is there any breaking change?

No. The API is the same as before, except that it is extended to have more options.

Is there code that should be adapted?

Errors and Warning messages should be curated. This means identifying the recipient, curating the text and using the extended API if appropriate.

What is the main goal?

I. To be able to differentiate between messages that are curated and translated and intended to be shown to the end user and messages which may be very useful for a power user (e.g. to debug macros or console statements) or for developers.

II. To stop showing stack traces in the Notification Area, which is a major reason of complaint.

Should this be merged for 0.21?

I think it would be advantageous to do so, if only because of the traces being reported in the Notification Area. However, I do not have a strong opinion.

### DIFF
--- a/src/Base/Console.h
+++ b/src/Base/Console.h
@@ -473,12 +473,28 @@ enum class LogStyle{
     Notification,           // Special message for notifications to the user (e.g. educational)
 };
 
+/** Used to indicate the intended recipient of a message (warning, error, ...).
+    While it is possible to create a custom message via Console().Send, console provides convenience
+    functions for most common combinations.
+
+    See documentation of ConsoleSingleton class for more details.
+
+    @see ConsoleSingleton
+    */
 enum class IntendedRecipient {
     All,                    // All recipients, Developers and Users (One notification covers all)
     User,                   // User only (notification intended only for a user)
     Developer,              // Developer only (notification intended only for a developer)
 };
 
+/** Used to indicate the translation state of a message (warning, error, ...).
+    While it is possible to create a custom message via Console().Send, console provides convenience
+    functions for most common combinations.
+
+    See documentation of ConsoleSingleton class for more details.
+
+    @see ConsoleSingleton
+    */
 enum class ContentType {
     Untranslated,           // Not translated, but translatable
     Translated,             // Already translated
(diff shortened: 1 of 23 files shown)


src/Base/Console.h
```diff
@@ -471,7 +471,18 @@ enum class LogStyle{
...
-    TranslatedNotification, // Special message for already translated notifications to the user (e.g. educational)
+};
+
+enum class IntendedRecipient {
+    All,                    // All recipients, Developers and Users (One notification covers all)
+    User,                   // User only (notification intended only for a user)
+    Developer,              // Developer only (notification intended only for a developer)
+};
+
+enum class ContentType {
```
chennes, line 482: I really like this design (thanks for the strong typing!) -- can you add a bit of documentation here explaining how a developer should use this? e.g.
```
/// Use the QT_TRANSLATE_NOOP("Notification", "Static error string here") when using `Untranslated`
/// Use tr() when using Translated so that the string is pre-translated.
```
(or something like that)

abdullahtahiriyo: Thank you. Sure!

chennes (approved):
I'm personally in favor of merging this now, since it will let me address some translation issues before the 0.21 release. That said, whether this is a "feature" or a "bug fix" is a little nebulous, so others may disagree with me.

abdullahtahiriyo:
Thank you.

I also think the benefits outweigh the possible drawbacks, and there is still some time to detect and fix any outstanding issue that might have been inadvertently introduced.

PR #9652, opened 2023-05-24
Sketcher console error split

abdullahtahiriyo:
Revisiting all console calls:
- Using the Notification Framework when the user needs to be notified
- Turning them into DeveloperError/DeveloperWarnings when messages are intended only for developers

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerArcOfHyperbola.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerArcOfHyperbola.h
@@ -25,6 +25,8 @@
 
 #include <boost/math/special_functions/fpclassify.hpp>
 
+#include <Gui/Notifications.h>
+
 #include "GeometryCreationMode.h"
 
 
@@ -224,7 +226,9 @@ public:
 
             if (boost::math::isnan(startAngle) || boost::math::isnan(endAngle)) {
                 sketchgui->purgeHandler();
-                Base::Console().Error("Cannot create arc of hyperbola from invalid angles, try again!\n");
+                Gui::NotifyError(sketchgui,
+                                    QT_TRANSLATE_NOOP("Notifications", "Error"),
+                                    QT_TRANSLATE_NOOP("Notifications", "Cannot create arc of hyperbola from invalid angles, try again!"));
                 return false;
             }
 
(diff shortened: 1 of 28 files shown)


src/Mod/Sketcher/Gui/DrawSketchHandlerArcOfHyperbola.h
```diff
@@ -224,7 +226,9 @@ class DrawSketchHandlerArcOfHyperbola : public DrawSketchHandler
 
             if (boost::math::isnan(startAngle) || boost::math::isnan(endAngle)) {
                 sketchgui->purgeHandler();
-                Base::Console().Error("Cannot create arc of hyperbola from invalid angles, try again!\n");
+                Gui::NotifyError(sketchgui,
+                                    QT_TRANSLATE_NOOP("Notifications", "Error"),
+                                    QT_TRANSLATE_NOOP("Notifications", "Cannot create arc of hyperbola from invalid angles, try again!"));
```
chennes, line 231: "Try again!" might be out of place here :). Then again, how often are these angles actually NaN? I gather this is a rare error...

chennes (approved):
I love the addition of `DeveloperError` and `DeveloperWarning`, that is _such_ a useful distinction. Thanks for this demonstration of how the system should work.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcOfHyperbola.h
abdullahtahiriyo: I thought that out a couple of times.

This kind of expressions are generally out of place. It may even appear as misleading: "If I did it once and got an error, if I do again I would get an error to, wouldn't I?". The problem is that... most probably not, because not the same exact points are going to be hit. So perhaps trying again is not a bad piece of advice after all.

That is why I left it that way. But as you say, I have never seen this error popping...

abdullahtahiriyo:
@chennes 

> I love the addition of `DeveloperError` and `DeveloperWarning`, that is _such_ a useful distinction. Thanks for this demonstration of how the system should work.

We are still shaping the system and perhaps it needs to be extended again. 

When I was moving to the previous framework, user-only messages could only be notifications ('i' icon). Something was actually dying inside me when I was applying it to the Sketcher. When the need for developer only messages arose, it was clear to me that the framework needed to be more flexible... but the requirements for users and developers where symmetric and both could be brought easily with the same architecture.

I still have left for the weekend the issue with the dynamic strings... that one is a tougher one...

PR #9654, opened 2023-05-24
Draft: fix getcoords in gui_snapper.py

Roy-043:
The function did not handle global and relative mode properly.

It was not updated when global mode was introduced. This primarily affected Arch commands. As a consequence it was impossible to place a column at the origin of a custom Draft working plane through keyboard input.

Forum topic:
https://forum.freecad.org/viewtopic.php?t=78538

---


yorikvanhavre:
I didn't remember this get_coords function..  I couldn't find any use of it anywhere else in Draft or Arch, so AFAICS this should be safe

PR #9667, opened 2023-05-25
Spreadsheet: fix string save/restore, fixes #6395

### LINKED ISSUE #6395: [Bug] Spreadsheet cell values are not correctly saved / restored

### Version

0.19 (Release) and 0.20 (Development)

### Full version info

```shell
OS: Linux (GNOME/gnome)
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.19.24366 (Git)
Build type: Release
Branch: makepkg
Hash: 6530e364184ce05ccff39501e175cf2237e6ee4b
Python version: 3.10.2
Qt version: 5.15.2
Coin version: 4.0.1
OCC version: 7.5.3
```
and
```
OS: Linux (GNOME/gnome)
Word size of FreeCAD: 64-bit
Version: 0.20.27630 (Git)
Build type: Unknown
Branch: master
Hash: 13fb62fc3f9d65daf5b2b14e250a9a61267d8281
Python 3.10.2, Qt 5.15.2, Coin 4.0.1, OCC 7.5.3
```

### Subproject(s) affected?

Spreadsheet

### Description

If I type `'3M` into a spreadsheet cell, it will be displayed as `3M` which is correct. If I save and reopen the document, it will now display `0.05 °` in the cell.

### How to reproduce

1. Open FreeCAD
2. File -> New
3. Select Workbench Spreadsheet
4. Create new Spreadsheet
5. Click on cell A1
6. Type `'3M`
7. Enter
8. Type `=<<3M>>`
9. Enter
10. Type `=<<%s>> % <<3M>>`
  - Spreadsheet looks like this: 
    Screenshot-1645214027
11. File -> Save
12. Close FreeCAD
13. Open previous saved file
14. Open spreadsheet
  - Now the spreadsheet looks like this: 
    Screenshot-1645215240
15. Cell content for A1 and A2 changed to `0.05 °`

### Additional info from python console

After step 10 the python console gives the following values:
```
>>> App.ActiveDocument.Spreadsheet.A1
'3M'
>>> App.ActiveDocument.Spreadsheet.A2
'3M'
>>> App.ActiveDocument.Spreadsheet.A3
'3M'
>>> App.ActiveDocument.Spreadsheet.cells.Content
'<Cells Count="3" xlink="1">\n    <XLinks count="0">\n    </XLinks>\n    <Cell address="A1" content="3M" />\n    <Cell address="A2" content="3M" />\n    <Cell address="A3" content="=&lt;&lt;%s&gt;&gt; % &lt;&lt;3M&gt;&gt;" />\n</Cells>\n'
```

### Document.xml content (extracted from saved file)

```
<!-- snip -->
<Cell address="A1" content="3M" />
<Cell address="A2" content="3M" />
<Cell address="A3" content="=&lt;&lt;%s&gt;&gt; % &lt;&lt;3M&gt;&gt;" />
<!-- snip -->
```

### Expected

If I change the cells in Document.xml to the following:
```
<!-- snip -->
<Cell address="A1" content="'3M" />
<Cell address="A2" content="=&lt;&lt;3M&gt;&gt;" />
<Cell address="A3" content="=&lt;&lt;%s&gt;&gt; % &lt;&lt;3M&gt;&gt;" />
<!-- snip -->
```
and open the document, it will correctly be opened and shows the values in the spreadsheet as before it was closed.

### Possible bug location

I am not sure, may this is completely misleading, but from having a short look at the code, I would expect the bug in `Cell::setContent` in line https://github.com/FreeCAD/FreeCAD/blob/7b7b76f1ec741151efda3af20b219d82a75f6632/src/Mod/Spreadsheet/App/Cell.cpp#L305

### Code of Conduct

0penBrain:
See also https://forum.freecad.org/viewtopic.php?t=78204

### DIFF
--- a/src/Mod/Spreadsheet/TestSpreadsheet.py
+++ b/src/Mod/Spreadsheet/TestSpreadsheet.py
@@ -1082,7 +1082,7 @@ class SpreadsheetCases(unittest.TestCase):
         self.doc.recompute()
         sheet.set('A3', '')
         sheet.set('A3', 'A1')
-        self.assertEqual(sheet.getContents('A3'), 'A1')
+        self.assertEqual(sheet.getContents('A3'), '\'A1')
 
     def testInsertRowsAlias(self):
         """ Regression test for issue 4429; insert rows to sheet with aliases"""
@@ -1447,6 +1447,21 @@ class SpreadsheetCases(unittest.TestCase):
         self.doc.recompute()
         self.assertEqual(sheet.getContents('A1'), '=1 == 1 ? 1 : 0')
         self.assertEqual(sheet.A1, 1)
+
+    def testIssue6395(self):
+        """ Testing strings are correctly saved and restored"""
+        sheet = self.doc.addObject('Spreadsheet::Sheet','Spreadsheet')
+        sheet.set('A1', '\'36C') # Use a string that may be parsed as a quantity
+        self.doc.recompute()
+
+        self.doc.saveAs(self.TempPath + os.sep + 'string.fcstd')
+        FreeCAD.closeDocument(self.doc.Name)
+
+        self.doc = FreeCAD.openDocument(self.TempPath + os.sep + 'string.fcstd')
+
+        sheet = self.doc.getObject('Spreadsheet')
+        self.assertEqual(sheet.getContents('A1'), '\'36C')
+        self.assertEqual(sheet.get('A1'), '36C')
 
     def tearDown(self):
         #closing doc
--- a/src/Mod/Spreadsheet/App/Cell.cpp
+++ b/src/Mod/Spreadsheet/App/Cell.cpp
@@ -250,12 +250,7 @@ bool Cell::getStringContent(std::string & s, bool persistent) const
             s = "=" + expression->toString(persistent);
         else if (freecad_dynamic_cast<App::StringExpression>(expression.get())) {
             s = static_cast<App::StringExpression*>(expression.get())->getText();
-            char * end;
-            errno = 0;
-            double d = strtod(s.c_str(), &end);
-            (void)d; // fix gcc warning
-            if (!*end && errno == 0)
-                s = "'" + s;
+            s = "'" + s;
         }
         else if (freecad_dynamic_cast<App::ConstantExpression>(expression.get()))
             s = "=" + expression->toString();


adrianinsaval:
I guess this has to be changed to compare to "'A1"

adrianinsaval:
Would be nice to add a test for #6395 too

0penBrain:
> This makes all strings start with a ' right?

Yes exactly. Any content detected as a constant string.

> I guess this has to be changed to compare to "'A1"
> 
> https://github.com/FreeCAD/FreeCAD/blob/a182c0140926e3682d1c305fceac632eee09d408/src/Mod/Spreadsheet/TestSpreadsheet.py#L1085

Yes, but I wanted to check a bit more in depth that it wasn't a sign that this PR would create regressions or side-effects.

> Would be nice to add a test for #6395 too

It's here now

src/Mod/Spreadsheet/TestSpreadsheet.py
```diff
@@ -1448,6 +1448,21 @@ def testParensAroundCondition(self):
...
+    def testIssue6395(self):
+        """ Testing strings are correctly saved and restored"""
+        sheet = self.doc.addObject('Spreadsheet::Sheet','Spreadsheet')
+        sheet.set('A1', '\'36C') # Use a string that may be parsed as a quantity
+        self.doc.recompute()
+
+        self.doc.saveAs(self.TempPath + os.sep + 'string.fcstd')
+        FreeCAD.closeDocument(self.doc.Name)
+
+        self.doc = FreeCAD.openDocument(self.TempPath + os.sep + 'string.fcstd')
```
adrianinsaval, line 1460: For this case I don't think save and restore is really required, in master this test would fail even if you skip this.

0penBrain: You asked for a test versus #6395

adrianinsaval: Yes, but while #6395 is more obviously noticeable after save an restore the root of the problem happens immediately, the leading ' is not stored when it should, but testing save and restore also probably doesn't hurt.
I guess `sheet.get('A1')` would give a different result than before after save+restore but the test would already fail in the previous line.

It might be a good idea to add other potential failures to the test too, like `''asd` and `'1/2m` in case someone later on wants to bring back auto removal of leading ' but only takes the 36C exception into account.

0penBrain: Well might be. Too tired with idea ATM. Will eventually come back to this later on.

chennes: Adding more tests would certainly be nice: I propose that we merge this PR as-is (CONTRIBUTING.md does not require 100% test coverage) and that those interested in the topic continue development as needed.

0penBrain: > I propose that we merge this PR

Beside tests, be aware that this PR will display all fixed string with a prepended single quote in the 'content' field.
To me it's acceptable, but YMMV.

chennes: It's not my favorite, but I think for the time being it's worth it to resolve the bug. Once it's been in people's hands for a while we will see if it's too irksome and we need an alternative or refinement to it.

adrianinsaval: I think it's better than the bug, we can try to improve later if there are complaints

adrianinsaval:
works as advertised, shouldn't cause any regression since the actual value of the expression is unaffected

PR #9668, opened 2023-05-25
[Core] Fix ordering confusion in backups

### LINKED ISSUE #9635: [Problem]  Backup file numbering is confusing

### Version

0.20 (Release)

### Full version info

```shell
N/A
```

### Subproject(s) affected?

Core

### Problem description

If the user sets the preference setting for more than one backup file to be created, the numbers assigned to backup files are confusing.  
For example, if the user sets the value to 3, the first backup is .FCStd1.  Then .FCStd2, etc.  When the available slots are filled, older files need to be renamed/deleted/overwritten.  You can end up with strange effects like this:
Nº1 is the newest
Nº3 is the next newest
Nº2 is the oldest.

### Anything else?

_No response_

### Code of Conduct

AjinkyaDahale:
Fixes #9635.

### DIFF
--- a/src/App/Document.cpp
+++ b/src/App/Document.cpp
@@ -1556,7 +1556,7 @@ public:
     BackupPolicy() {
         policy = Standard;
         numberOfFiles = 1;
-        useFCBakExtension = false;
+        useFCBakExtension = true;
         saveBackupDateFormat = "%Y%m%d-%H%M%S";
     }
     ~BackupPolicy() = default;
@@ -1928,7 +1928,7 @@ bool Document::saveToFile(const char* filename) const
             count_bak = -1;
         }
         bool useFCBakExtension = App::GetApplication().GetParameterGroupByPath
-            ("User parameter:BaseApp/Preferences/Document")->GetBool("UseFCBakExtension",false);
+            ("User parameter:BaseApp/Preferences/Document")->GetBool("UseFCBakExtension",true);
         std::string saveBackupDateFormat = App::GetApplication().GetParameterGroupByPath
             ("User parameter:BaseApp/Preferences/Document")->GetASCII("SaveBackupDateFormat","%Y%m%d-%H%M%S");
 
--- a/src/Gui/DlgSettingsDocument.ui
+++ b/src/Gui/DlgSettingsDocument.ui
@@ -511,6 +511,9 @@ Common sizes are 128, 256 and 512</string>
            <string>Backup files will get extension '.FCbak' and file names
 get date suffix according to the specified format</string>
           </property>
+          <property name="checked">
+           <bool>true</bool>
+          </property>
           <property name="text">
            <string>Use date and FCBak extension</string>
           </property>


AjinkyaDahale:
TODO: How to set default to `FCBak`? I know this is a routine method but I don't know the exact mechanism.

AjinkyaDahale:
There are a couple of corner cases when migrating the `FCStd#` system from the older ordering with the desired one. I'll comment more once I've slept on it.

AjinkyaDahale:
Found relevant settings to make the `FCBak` extension default, but it didn't reflect in the configs until I deleted the relevant line in `user.cfg`. @chennes @wwmayer could you tell when these settings entries are written there and how to check if it's done by default or by user configuration?

adrianinsaval:
> how to check if it's done by default or by user configuration?

There is no way to know, this is a big flaw in how the preference page works, once you hit ok every preference is written to the config file even if the user didn't touch them, so anyone who's opened the preference page before now has the defaults written in their config

AjinkyaDahale:
Did some basic testing and I can call this ready for review.

Here's a zip file with backups if you anyone wants to test
[simple-pad-with-backups.zip](https://github.com/FreeCAD/FreeCAD/files/11591625/simple-pad-with-backups.zip)
After saving the file with any necessary changes, the backups should be in intended order of recency. Some things to confirm (not exhaustive):
1. Renaming files doesn't change modified date. Might be platform-specific.
2. What happens when the number of backups is increased/decreased?
3. What happens if oldest/newest/middle backup is deleted/renamed.

I haven't yet changed the name of the variable yet to force `FCBak` extension.

sliptonic:
This PR was discussed at the contributor meetup on Sunday 6/4.   
The consensus recommendation is:
- Change the default setting to FCBak.  All new users will use FCBak with thte date format template as the default going forward
- Leave the .fcstd# sequence numbering unchanged and consider this backup naming scheme to be deprecated
- Add a warning to existing users whose preferences are using .fcstd#  that they are using a deprecated scheme.  
- Get these changes merged to v21
- remove .fcstd# sequence backup as an option in a future version.

chennes:
Is this then back in draft status? Or is it still ready to review?

AjinkyaDahale:
> Is this then back in draft status? Or is it still ready to review?

The first commit here will have to be dropped and another commit for the warning will need to be added. Draft it is.

AjinkyaDahale:
Earlier this weekend I dropped the commit (there's a backup in my local machine if its needed). However I have not started on the warning. Does anyone know of an example where this was done earlier?

adrianinsaval:
Is a warning in the software actually needed? Isn't it sufficient to mention this in the release notes?

AjinkyaDahale:
> Is a warning in the software actually needed? Isn't it sufficient to mention this in the release notes?

I think it's a good idea to have it in place. Not everyone will read this particular item.

There are two settings (3 if you count the timestamp format) relevant to backups: how many backups to keep, and whether or not to use `.FCBak`. The default number of backups is 0, and in that case it is better to quietly move to `.FCBak` and skip the warning.

luzpaz:
I think a warning is important (not just in Release Notes which our Attention Deficit brethren will surely skip reading)

adrianinsaval:
The default is 1 backup if I'm not mistaken

AjinkyaDahale:
> The default is 1 backup if I'm not mistaken

My bad I was mistaken. There is also one more checkbox to ask whether or not to use backups at all.

yorikvanhavre:
I would think of two options:

1. We simply display the warning as a yellow (warning) line in the report view, at FreeCAd startup. Unintrusive, but it's nagging: If a user wishes to keep the fcstd# format, they will see the message each and every time they start FreeCAD.
2. We pop up a dialog (a simple QMessageBox warning?) at FreeCAD startup, one and only one time telling the user this format will be deprecated in a the future and the user just has to press OK to dismiss.

I think I would do 2. It's a little bit more work, and you need to store a pref setting for FreeCAD to remember the dialog has been displayed already, but it seems more elegant

AjinkyaDahale:
> I would think of two options:
> 
>     1. We simply display the warning as a yellow (warning) line in the report view, at FreeCAd startup. Unintrusive, but it's nagging: If a user wishes to keep the fcstd# format, they will see the message each and every time they start FreeCAD.

At least it's not a dialog. 

>     2. We pop up a dialog (a simple QMessageBox warning?) at FreeCAD startup, one and only one time telling the user this format will be deprecated in a the future and the user just has to press OK to dismiss.

Ideally there should be an option to change this setting right in the dialog box. In the future there may be more deprecations, so either this setting will have to be saved for each of the values deprecated, or saved as a list (if that's an option), or the setting will have to be reset everytime a new deprecation is added.

There's also a factor where the user dismisses the warning once, but say, wants to go back to find what it was.

yorikvanhavre:
I wonder if we shouldn't provide a way for users to disable that warning... It might get really annoying if someone intents to keep using the fcstd# format

> Do we need this warning in CLI too?
I don't think so... The CLI is mostly made for automation so probably nobody would see it.
Make sure it is in the release notes! That's probably the most important

AjinkyaDahale:
> I wonder if we shouldn't provide a way for users to disable that warning... It might get really annoying if someone intents to keep using the fcstd# format

Starting 0.22/1.0 the fcstd# is already scheduled to go away. We can in fact do it the very first commit after releasing 0.21. That warning seems necessary.

yorikvanhavre:
Yes that's true...  if there are problems with the fcbak system we should fix them anyway so there is no disadvantage to use it over fcstd.

AjinkyaDahale:
Not sure why tests are failing. Doesn't seem directly related to the changes.

adrianinsaval:
Maybe the warning is showing up during test and messes up something else.

adrianinsaval:
@0penBrain do you know if this is a genuine problem in the PR or a problem in our CI? The message about the action timing out makes me suspicious.

0penBrain:
> @0penBrain do you know if this is a genuine problem in the PR or a problem in our CI? The message about the action timing out makes me suspicious.

I tend to believe it is a regression in the PR. It fails the same way already twice, and CLI tests on Ubuntu workflows are generally completed in about 1.5 min.

AjinkyaDahale:
> I tend to believe it is a regression in the PR. It fails the same way already twice, and CLI tests on Ubuntu workflows are generally completed in about 1.5 min.

Problem is I cannot reproduce the failure here. It's giving me a different test failure though.

0penBrain:
> Problem is I cannot reproduce the failure here. It's giving me a different test failure though.

To be clear there is no FC test failing per se. But the CI has a time limit to execute all CLI tests that is reached here. So it cancels the task and reports a failure.

AjinkyaDahale:
> > Problem is I cannot reproduce the failure here. It's giving me a different test failure though.
> 
> To be clear there is no FC test failing per se. But the CI has a time limit to execute all CLI tests that is reached here. So it cancels the task and reports a failure.

Windows succeeded after 205 minutes. So maybe that's what it needs.

0penBrain:
> Windows succeeded after 205 minutes. So maybe that's what it needs.

To be clear, the 10-minute limit is to execute the lone 'CLI tests' steps. Not the whole process.

0penBrain:
> The warning is in GUI only. So somehow changing defaults is causing this timeout?

I have absolutely not investigated (even look the purpose of your PR :stuck_out_tongue: ) but your changes makes the CLI tests last for more than 10 minutes while there normal execution time is around 1.5 minute.

AjinkyaDahale:
> I have absolutely not investigated (even look the purpose of your PR stuck_out_tongue ) but your changes makes the CLI tests last for more than 10 minutes while there normal execution time is around 1.5 minute.

Right. Sorry for presuming. The changes are not supposed to be significant and that is causing me a lot of confusion. The first commit is changing the default of a setting from false to true, and that is the only commit affecting the CLI. So now I'm totally confused how it can be that this change is causing the hang, sometimes without a single test running.

0penBrain:
OK, I fetch your PR locally and can't reproduce this effect. :confused: 
@AjinkyaDahale Do you want me to clear CI caches of your PR and re-run to check ?

AjinkyaDahale:
> @AjinkyaDahale Do you want me to clear CI caches of your PR and re-run to check ?

Yes. Let's try throwing everything we can at it.

AjinkyaDahale:
> OK, I fetch your PR locally and can't reproduce this effect. 😕 

Did any other test fail for you? workbench activation seemed to fail for me.

0penBrain:
> Didn't pay attention but IIRC 2 tests where failing, will re-test and report.

Bad memory. 2 tests are skipped, but all others succeed.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9668">pipeline status</a> for feature branch [PR_9668](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9668). Pipeline [903567741](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903567741) was triggered at [e5306af](https://github.com/FreeCAD/FreeCAD/pull/9668/commits/e5306af0ad010ee7877e7ac9b1af12fda9800a45). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9668">pipeline status</a> for feature branch [PR_9668](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9668). Pipeline [904526153](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/904526153) was triggered at [30577f4](https://github.com/FreeCAD/FreeCAD/pull/9668/commits/30577f4fc9f6af837dbb62301a402b7f28b23afb). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

AjinkyaDahale:
> May you temporarily add a third commit that would enable the test on build dir on the CI ? To do so you have to add (correctly indented): `testOnBuildDir: true` after this line
> 
> https://github.com/FreeCAD/FreeCAD/blob/127404d793d5c2081383b46a7b677426d9a0c5aa/.github/workflows/CI_master.yml#L46
> and this one
> 
> https://github.com/FreeCAD/FreeCAD/blob/127404d793d5c2081383b46a7b677426d9a0c5aa/.github/workflows/CI_master.yml#L52
> 

Sorry for the late reply. I just pushed this.

> You may also in same temporary commit revert your second commit as it not supposed to influence on CLI tests.

Didn't read this part when I made the commit. Maybe in the next try.

On another note, freecadcli is miraculously passing.

AjinkyaDahale:
@0penBrain so the test fails even with `testOnBuildDir: true` (unless I'm doing something wrong). I don't expect the second commit revert to do anything as well, but I'll try as soon as the CI finishes.

AjinkyaDahale:
> You may also in same temporary commit revert your second commit as it not supposed to influence on CLI tests.

Trying this now, albeit in a separate commit.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9668">pipeline status</a> for feature branch [PR_9668](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9668). Pipeline [904669911](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/904669911) was triggered at [6b328b9](https://github.com/FreeCAD/FreeCAD/pull/9668/commits/6b328b94f05c7c3a0786bb45cd889cf1aa26acff). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

0penBrain:
> Trying this now, albeit in a separate commit.

Ok. Everything fails consistently even when testing on build dir. Last test may be to finally revert 1st commit as well to see if situation is back to normal.

AjinkyaDahale:
> Last test may be to finally revert 1st commit as well to see if situation is back to normal.

Done. We wait patiently.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9668">pipeline status</a> for feature branch [PR_9668](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9668). Pipeline [904771964](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/904771964) was triggered at [5e8f96c](https://github.com/FreeCAD/FreeCAD/pull/9668/commits/5e8f96c11074f60b5675c9add8c08cd56bb5d731). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

adrianinsaval:
something seems off on github CI, on gitlab the tests run in 3min.

AjinkyaDahale:
OK. Making a new PR now. If that works we discard this one with its gremlins.

AjinkyaDahale:
> @AjinkyaDahale I'm running the same changeset as your 1st commit on my fork : https://github.com/0penBrain/FreeCAD/actions/runs/5325233183 There was no cache there so we'll see. Will take some time.

There's already this: https://github.com/Ondsel-Development/FreeCAD/tree/core-better-numbering-retry-2. Don't know if there was any cache there.

0penBrain:
> Don't know if there was any cache there.

https://github.com/Ondsel-Development/FreeCAD/actions?query=actor%3AAjinkyaDahale+branch%3Acore-better-backup-numbering

0penBrain:
Ok so my test makes it clear that enabling FCBak makes the CI CLI tests fail. Probably won't be so easy to investigate.

AjinkyaDahale:
Trying to zero on what changes are causing the issue.

AjinkyaDahale:
@0penBrain for whatever reason now more tests are happening before failure. Increasing the 10-minute timeout on Ubuntu tests might solve the problem, though why tests take 10x is a new problem altogether.

0penBrain:
> @0penBrain for whatever reason now more tests are happening before failure.

I already did see such case in some of your previous pushes. :wink: 

> Increasing the 10-minute timeout on Ubuntu tests might solve the problem,

Feel free to remove it on your branch for testing purposes. It's here (same on other Ubuntu workflow): https://github.com/FreeCAD/FreeCAD/blob/28f8abd3002ec15878359ab804ada52eafa833e2/.github/workflows/sub_buildUbuntu2204.yml#L187
You can remove/comment the line or set a bigger value (60 ?).

> though why tests take 10x is a new problem altogether.
Yes, I didn't have time to investigate yet. A wild guess would be that at some time it needs a user interaction (something like "Are you sure you want to overwrite the file ?") for some reason and so get stuck. But not sure as it doesn't reproduce on local build.

AjinkyaDahale:
> > @0penBrain for whatever reason now more tests are happening before failure.
> 
> I already did see such case in some of your previous pushes. 😉

I don't think I did. I don't recall even seeing the ascii logo in runs which I saw.

> > Increasing the 10-minute timeout on Ubuntu tests might solve the problem,
> 
> Feel free to remove it on your branch for testing purposes. It's here (same on other Ubuntu workflow):
> 
> https://github.com/FreeCAD/FreeCAD/blob/28f8abd3002ec15878359ab804ada52eafa833e2/.github/workflows/sub_buildUbuntu2204.yml#L187

Let me confirm with the local version first. If it takes the same time in CLI I'll remove the limit.

AjinkyaDahale:
Removed the CLI testing time limit. @0penBrain in case it keeps going for hours, how do I force stop this?

adrianinsaval:
If you click to see the workflow run there should be a cancel button on the top right, eventually GitHub will auto cancel it if it reaches 6h

AjinkyaDahale:
@adrianinsaval please cancel the job if you can. I can't find the button.

0penBrain:
> @adrianinsaval please cancel the job if you can. I can't find the button or don't have the authorization.

I just did. ;)

AjinkyaDahale:
I think I see what might be causing the delay, but I still don't know how. When backups are created in quick succession, they can have the same time-stamp up to the second. In this case, the new file is saved with a suffix.

Screenshot from 2023-06-29 00-33-51

Pushed a commit that should ensure this doesn't happen. Let's see how that goes.

AjinkyaDahale:
Experiments removing certain tests that were last reported before the infinite loop.
1. `TestIssue4823`: https://github.com/FreeCAD/FreeCAD/actions/runs/5444466949/jobs/9902401846?pr=9668
2. 1 and `TestMem`:

0penBrain:
@AjinkyaDahale sorry I'm not able to get time currently to help you with this. Good to see your tenacity. Feel free to ping of needed.

0penBrain:
OK, was a long journey but finally found the culprit.
The `isFile()` function of `FileInfo` uses an "optimistic" implementation that freezes FC on specific files.
When a file is saved using `FCBak`, the whole folder content list is tested (each item is tested to be a file or not).
In case of tests, this is the `/tmp/` folder and it happens that CI shall have in this folder some very specific file(s) that makes the `isFile()` function to freeze.
Didn't investigate deeper (enough time consumed :smile: ).
Once #9974 is merged, this one can be cleaned up and CI should pass. :wink:

sliptonic:
> OK, was a long journey but finally found the culprit. 

Wow!  Super impressive work OpenBrain!   How did you even find it?

0penBrain:
> Wow! Super impressive work OpenBrain! How did you even find it?

Brute force attack. :smile:

First find the difference between using or not "FCBak" => In `App/Document.cpp`, `applyTimeStamp()` is used instead of `applyStandard()`.
OK, replace `applyTimeStamp()` with `applyStandard` => No freeze. OK, it's in `applyTimeStamp()`

Then comment-dichotomy. Comment/Uncomment larger to smaller parts of code and run the CI, and find your way though the passes and failures.

In parallel, as soon as I knew that problem is in the loop that iterates over the directory content, I locally probed the code to see the item list returned by the current implementation, then compared to a quick and dirty re-implementation with `QDir` => `QDir` implementation returns less items (especially all folders with setuid bit set aren't listed).
Then launched my `QDir` implementation in the CI => Pass

At the end found `isFile()` is the cause. Went to it and found this "a-bit-strange" implementation.
Replaced by something more "standard" using `stat` (you can see that `isDir()` already used a similar implementation.
Run this new implementation in the CI => :relieved: passes

Make the PR, comment this one, breath :smile:

This was actually very long because it happened only in the CI, and it was a freeze not a crash (so no framestack). There you can't run a debugger, and each testing step requires a CI run which takes 15~20 min to now if CLI tests are passing or freezing. :wink:

AjinkyaDahale:
Force-pushed with rebase. Let's see how this goes.

0penBrain:
> Force-pushed with rebase. Let's see how this goes.

Congrats. 🎉

chennes:
All green! You want to do the honors, @0penBrain ?

0penBrain:
No please, I'm too much involved in this. :smile:

AjinkyaDahale:
> > Force-pushed with rebase. Let's see how this goes.
> 
> Congrats. 🎉

Good riddance, more like 😅

PR #9687, opened 2023-05-28
Arch: more consistent object naming and labelling

Roy-043:
Following the example of Arch_Wall:
- Fixed object name.
- Default label is translatable.
- Name argument should only be used for the label and not be translated.

---


yorikvanhavre:
That's a good cleaning!!! Thanks a lot @Roy-043 !

PR #9694, opened 2023-05-29
Sketcher/Gui: Extend toolbar framework and fFix toolbar visibility not preserved

### LINKED ISSUE #9208: [Issue] Sketcher Toolbar Configuration not Saved

### Forums discussion

_No response_

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Fedora Linux 37 (Workstation Edition) (GNOME/gnome)
Word size of FreeCAD: 64-bit
Version: 0.21.0.32764 (Git)
Build type: Release
Branch: master
Hash: 8ea5d9bac10698a9ae31735ad45a63688a6be58b
[/code]
```

### Subproject(s) affected?

Sketcher

### Issue description

_Sketcher edit mode toolbars_ position and visibility is not saved to user.cfg after modification.

Steps to reproduce:

- Switch to sketcher
- New sketch, open edit mode
- Rearrange toolbars
- Hide toolbars
- Close FreeCAD

When FreeCAD is restarted all sketcher toolbars are restored to default visibility and position.

### Anything else?

_No response_

### Code of Conduct

abdullahtahiriyo:
fixes #9208

Essentially:
- The regular mechanism to save toolbar state when changing from one WB to another is not designed to support changes within one WB (e.g. from edit mode and back).
- At creation time, toolbars can be initialised with default visible or default hidden state. Additionally, there is third configuration "Unavailable", which refers to a toolbar that is hidden, and the control to enable it is also hidden by default.
- The ToolBarManager is extended to enable to set the State of one or more toolbars.
- The State refers to changes to be effected on one or more toolbars by client code:
* ForceHidden allows to hide a toolbar and also hide its control (a toolbar not available in a mode).
* ForceAvailable allows to make a toolbar available by making its control visible, the toolbar itself is visible or not depending on user settings.
* RestoreDefault allows to bring the control visibility to the default of the toolbar, the toolbar itself is visible or not depending on user settings.
* SaveState allows to store the current visibility state of a toolbar. It enables client code to save the state when appropriate. It provides the only option for default "Unavailable" toolbars, which are fully managed by client code. It provides additional flexibility to save other toolbar visibility on request.

For the Sketcher this means:
- That edit mode toolbars are not shown outside edit mode.
- That edit mode toolbars and non-edit mode toolbars can be configured independently.
- that edit mode toolbars' state is saved when leaving edit mode if and only if, the workbench that is selected when leaving edit mode is the Sketcher WB.
- it won't save the state if the user manually selected another WB and then left edit mode (why? see limitation above).

Limitation:
- When switching to another WB while in edit mode, the other WB is activated before the current WB (sketcher WB) is deactivated. This means that at sketcher level, the sketcher has no chance to save states or do other tidy up actions before the tools of the other WB are activated.
- This, however, is understood as not relevant enough as to warrant changing the mechanisms in place.


adrianinsaval:
does it work correctly if you close FreeCAD while on sketcher edit mode?

howie-j:
@adrianinsaval I commented on the issue, but toolbar visibility seems to be saved to user.cfg even if you close freecad in edit mode.

luzpaz:
Typo in 1st commit title `fFix` shoud be `fix`  
Screenshot_20230530_100213

abdullahtahiriyo:
@luzpaz 

Thanks. I have fixed the typo now.

PR #9697, opened 2023-05-30
Core: Tree : Remove useless root item and hide description by default.

PaddleStroke:
This PR solve this : https://forum.freecad.org/viewtopic.php?t=78555&sid=6c1c30d9b4ae46d34c3b0d929d6ab34e&start=10

It does 2 things : 
A - Remove the useless "Application" root item.
B - Hide by default the seldomly used "Description" column. Also when it is hidden, the labels are now hidden too.
Below a comparaison picture : on the left after this PR, on the right before.
image

---

### DIFF
--- a/src/Gui/Tree.cpp
+++ b/src/Gui/Tree.cpp
@@ -520,6 +520,7 @@ TreeWidget::TreeWidget(const char* name, QWidget* parent)
         documentPartialPixmap.reset(new QPixmap(icon.pixmap(documentPixmap->size(), QIcon::Disabled)));
     }
     setColumnHidden(1, TreeParams::getHideColumn());
+    header()->setVisible(!TreeParams::getHideColumn());
 }
 
 TreeWidget::~TreeWidget()
--- a/src/Gui/Tree.cpp
+++ b/src/Gui/Tree.cpp
@@ -901,6 +901,27 @@ void TreeWidget::contextMenuEvent(QContextMenuEvent* e)
         subMenu.addActions(subMenuGroup.actions());
     }
 
+    // add a submenu to present the settings of the tree.
+    QMenu settingsMenu;
+    settingsMenu.setTitle(tr("Tree settings"));
+    contextMenu.addSeparator();
+    contextMenu.addMenu(&settingsMenu);
+
+    QAction* action = new QAction(tr("Show description column"), this);
+    action->setStatusTip(tr("Show an extra tree view column for item description. The item's description can be set by pressing F2 (or your OS's edit button) or by editing the 'label2' property."));
+    action->setCheckable(true);
+
+    ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/TreeView");
+    action->setChecked(!hGrp->GetBool("HideColumn", true));
+
+    settingsMenu.addAction(action);
+    QObject::connect(action, &QAction::triggered, this, [this, action, hGrp]() {
+        bool show = action->isChecked();
+        hGrp->SetBool("HideColumn", !show);
+        setColumnHidden(1, !show);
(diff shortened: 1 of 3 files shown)


sliptonic:
It's amazing how much better/cleaner that looks.  
As a future enhancement, perhaps we could check if any items in the tree have a description (label2) and show the column.  That would give the best of both world.  People who use the label would see it, everyone else would have clean tree.

xtemp09:
I use the description column. =/

PaddleStroke:
Note this is only the default preference. So you can still enable the preference if you want to use it.
Having said that this HideColumn preference doesn't seem to be anywhere in the UI and needs to be entered in the parameter editor. So perhaps it would be best to add it to the Preferences dialog?

sliptonic:
> I use the description column. =/

You're definitely in the minority.  I did a google image search for "FreeCAD tree" and couldn't find a single example where someone was using it.  Most people use descriptive names instead.

Not saying it doesn't have value, but really an unused feature.

yorikvanhavre:
I use it too. "Description" is a central item in IFC. But this solution is good to me, indeed the tree looks way less cluttered. Two observations, though:

- Indeed users needing the column might want to restore it and won't know how. Why not add a context menu option to do that?
- The changes in TreeParams.cpp are under a `// Auto generated code (Tools/params_utils.py:203)` comment. Isn't there a risk that they are overwritten if that script is executed? Shouldn't that script be updated too?

yorikvanhavre:
I also don't understand what that script is needed for...  @realthunder maybe you can help here?

realthunder:
The `HideColumn` parameter is declared [here](https://github.com/FreeCAD/FreeCAD/blob/a3c4f4bb01f5201daf666852444a8f9c5adbda8d/src/Gui/TreeParams.py#L72). Change the second parameter to `True`. Check out the [class document](https://github.com/FreeCAD/FreeCAD/blob/a3c4f4bb01f5201daf666852444a8f9c5adbda8d/src/Gui/TreeParams.h#L41) of `TreeParam` for instruction on code generation. If the change is brief like this one. You can manually change the code [here](https://github.com/FreeCAD/FreeCAD/blob/a3c4f4bb01f5201daf666852444a8f9c5adbda8d/src/Gui/TreeParams.cpp#L145) and [here](https://github.com/FreeCAD/FreeCAD/blob/a3c4f4bb01f5201daf666852444a8f9c5adbda8d/src/Gui/TreeParams.cpp#L359).

PaddleStroke:
Thanks for your explanations @realthunder !
So I add the modification to the python file for the future as well as the manual modifications so that we don't need to generate the code.

So now I'm adding a submenu to add 'Tree settings' to the contextual menu of the tree. In this submenu I will add this 'show/hide description column'. And in the future we can put more settings in here, such at text size and icon size as you implemented in linkstage.

PaddleStroke:
That's it. I cleaned commits history, added the contextual menu and rebased. How it looks : 
image

sliptonic:
This looks like high-value/low-risk.   Can we merge this for 0.21?

MisterMakerNL:
Looks really nice! And it's prime screenspace we are saving!

adrianinsaval:
perhaps you enabled the column by mistake?

martinproks:
As @sliptonic  wrote:
> ... As a future enhancement, perhaps we could check if any items in the tree have a description (label2) and show the column. That would give the best of both world...

I'm for this. I'm using Label2 for my comments too / ok I'm minority ;-)

Another note: The 2nd column is headlined as "Description", but the content is Label2. It could be confusing, especially with examples that adds property named "Description": [Create a FeaturePython object part I / Adding properties](https://wiki.freecad.org/Create_a_FeaturePython_object_part_I#Adding_properties)

PR #9559, opened 2023-05-13
Sketcher: Point to Circle Distance Constraints

FlachyJoe:
Implements Point to Circle distance constraint

Capture d’écran_2023-05-13_14-28-46

PR #9166, opened 2023-04-02
Sketcher: Arc to Other distance constraint

FlachyJoe:
Propagate circles distance constraint to arcs.

PR #9703, opened 2023-05-31
Sketcher: Add command to switch arc helpers

FlachyJoe:
This PR add a command to allow user to show/hide arc helpers (introduced by #9257) from GUI.

All the Overlay (aka. Information Layer) commands are grouped in a unique new CommandSketcherOverlay.cpp.

As the VirtualSpace toolbar has only one icon, I propose to rename it to Display and move here the BSpline Overlay drop down and the new Arc Overlay icon.

I asked for a new icon in the forum https://forum.freecad.org/viewtopic.php?t=78733 and I'll switch this PR from draft when chosen.


FlachyJoe:
CI fails in Path workbench :interrobang:

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9703">pipeline status</a> for feature branch [PR_9703](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9703). Pipeline [903580486](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903580486) was triggered at [155fa1f](https://github.com/FreeCAD/FreeCAD/pull/9703/commits/155fa1f77270634bb9b3e504e6d64505f842321d). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PaddleStroke:
Can you please rebase (build faster) and adds a screenshot of the new command (easier to get what is done)?
If possible can you do that for the #9559 and #9166 ?

PaddleStroke (commented):
Overall the code looks good to me.
One thing I'm not sure about are the changes to the .qrc file. After thinking about it, having multiple groups is not really useful so merging them makes them all in alphabetical order, which is maybe better. But I'm not sure that belongs in this PR.

Now on the UI side of things. I'm not sure that a toolbar entry is necessary for this setting. I'm a skeptical about using toolbar real estate for such settings.
I think a checkbox in the Element taskbox setting button + a preference setting in the sketcher preference page would have been enough.

PaddleStroke:
And another thing is that you moved the bspline commands to the new commandSketcherOverlay, but not the switch virtual space.

PaddleStroke:
@abdullahtahiriyo  recently added a toolbutton for snapping settings and one for rendering order and IMO this is going in the wrong direction. This would be 3 more buttons cluttering the toolbars. 
These would be best in the settings button. Or if we really wants them in the toolbar, we could create a 'Sketcher setting' toolbar command grouping all the settings.

PaddleStroke:
I just read and found out that this UI decision has been asked to you by the maintainers (@adrianinsaval , @abdullahtahiriyo, @0penBrain) in the previous PR so I'm a bit late for the party.
Still I don't think this kind of ui decision should be taken by 3 people, imo this is exactly where forum polls should come in and rule the best decision.

We have two setting buttons, one for element and one for constraint, which purpose is to show settings: 
image

I think all settings should be put at the same place or the user doesn't know where to find them and the plethora of tools makes these options all but invisible to new users. Even veterans who can handle more icons as they know them all already, have backlashed after @abdullahtahiriyo all but forced my hand making the snapping button (https://forum.freecad.org/viewtopic.php?t=77229&sid=de0aea30b4a7c5030b4a65634a0fa93b&start=10 ).

I don't see any reason why not to use the settings buttons. If they are deemed too hidden or not accessible enough, then the solution is to have a single settings button in the 'sketcher-edit-mode' toolbar where we put all the sketcher settings that we need at hand : snapping, rendering order, arc helper, auto-constraint...

PaddleStroke:
Besides we need a preference setting for the color of the helper such that it can be customized by themes.

FlachyJoe:
Hi @PaddleStroke, my current arc helper implementation follow the BSpline overlay one (at the time I wrote it, I don't read yet the recent changes if some of them concern this).
As BSpline tools have icons, arc helper should has one and none of them reside in the setting menu.

0penBrain:
@PaddleStroke that's a lot of value judgement. According our rules, it must not belong to a PR.
You can open a ticket, discuss there and commit another PR on top of this one if you feel like it.

Beyond that, having here the same behavior as the one you tirelessly rejected in the past when you was the PR's author seems very weird to me.

PaddleStroke:
Sure thing it does not belong to the PR. Let's make some back and forth
then.

You say I'm having the same behavior as I criticized but this is wrong. I
am not the one who told him to make a command, then changed my mind and
asked something else.
Besides that's just my opinion and arguments, I didn't say "you have to
change".

PaddleStroke:
Another thought: It would be best to not have any command/settings at all. And have the arc-helper appear only when needed, ie when a constrain is on empty air.

FlachyJoe:
@PaddleStroke 
> have the arc-helper appear only when needed, ie when a constrain is on empty air.

Feel free to contribute as I don't know how to do this.

sliptonic:
The conversation seems to have dried up.  We're going to merge this as is.   There really wasn't an initial issue so value judgements were appropriate in this case.  Please create an issue first next time.

If there are outstanding issues, please open a new issue specifically for them.

luzpaz:
Is someone willing to add this in the 1.0 release page on the wiki ?

0penBrain:
> There really wasn't an initial issue so value judgements were appropriate in this case.

This is ridiculous. As stated in the description, this PR is 2nd part of a 2-step PR that matches an issue. The 2-step process was acknowledged by the WB maintainer.

PR #9709, opened 2023-06-01
Exceptions: Enable translations of dynamic strings

abdullahtahiriyo:
This is to solve this kind of issues:
https://github.com/FreeCAD/FreeCAD-translations/issues/183

This PR is a mechanism to allow to mark for translation dynamic strings in Exceptions and in particular, enabling Python exceptions to use such dynamic strings.

A piece of code is worth a thousand words:
```
  PyError_SetEnhancedString(
      PyExc_ValueError,
      QT_TRANSLATE_NOOP("Notifications",
                        "Datum {} for the constraint with index {}  is invalid"),
      (const char*)Quantity.getUserString().toUtf8(),
      Index);
```

This is a ```PyErr_SetString``` substitute using fmt::format syntax, which stores the formatter string and the arguments separately as strings in the exception. In Python this is done with an exception dictionary object. In c++ Base::Exception is extended to store it. It also marks the flag that is it is a translatable string.

With ```PyErr_SetString``` only static strings (without substitutions are possible).

Because the Exception error message is untranslated, any exception reporting to the Report View at the Interpreter is provided in the original English language (but formatted by fmt::format).

At the other end, the Notifications Framework is extended so that it can take an exception as a notification message:
```
            Gui::NotifyUserError(sketch, QT_TRANSLATE_NOOP("Notifications", "Value Error"), e);
        }
```
It will translate the formatter string and arguments as necessary, assemble the translation (by fmt::format), and report it to the notification area or blocking pop-up as per user preferences.

### DIFF
--- a/src/Base/EnhancedPythonExceptions.h
+++ b/src/Base/EnhancedPythonExceptions.h
@@ -1,0 +1,96 @@
+/***************************************************************************
+ *   Copyright (c) 2023 Abdullah Tahiri <abdullah.tahiri.yo@gmail.com>     *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU Library General Public License (LGPL)   *
+ *   as published by the Free Software Foundation; either version 2 of     *
+ *   the License, or (at your option) any later version.                   *
+ *   for detail see the LICENCE text file.                                 *
+ *                                                                         *
+ *   FreeCAD is distributed in the hope that it will be useful,            *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with FreeCAD; if not, write to the Free Software        *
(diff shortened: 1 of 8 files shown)


src/Base/EnhancedPythonExceptions.h
```diff
@@ -0,0 +1,96 @@
...
+        ...
+        );
+
+        d.setItem("sformatterArguments",t);
+    }
+
+    d.setItem("btranslatable", Py::Boolean(true));
+
+    PyErr_SetObject(exceptiontype, Py::new_reference_to(d));
+}
```
abdullahtahiriyo, line 92: @wwmayer 

I would welcome a trained second set of eyes on this function and in particular on whether I may be leaking (ref counter issues).

abdullahtahiriyo:
I will not be available for some time, so I think it is best to turn this into draft.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9709">pipeline status</a> for feature branch [PR_9709](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9709). Pipeline [903580490](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903580490) was triggered at [9e16b48](https://github.com/FreeCAD/FreeCAD/pull/9709/commits/9e16b487f02480ed5f6eae33165131e16b60882c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

maxwxyz:
@abdullahtahiriyo just a heads up for the upcoming feature freeze: 

maxwxyz:
any news on this PR? Will someone pick it up?

PR #9712, opened 2023-06-02
[Import] DXF make codepage case insensitive

Syres916:
See discussion https://forum.freecad.org/viewtopic.php?t=78719

I don't believe https://github.com/FreeCAD/FreeCAD/issues/8704 is fixed even after this change, the example file Encoding-error.dxf which can be opened by VariCAD Viewer still opens error free in FreeCAD but has nothing in the 3D View.

---


Syres916:
@yorikvanhavre Could you please review this PR, TIA

yorikvanhavre (approved):
Looks good to me! I was wondering if there isn't a better way to convert a string to lowercase, you could [use boost](https://stackoverflow.com/questions/313970/how-to-convert-an-instance-of-stdstring-to-lower-case):

 ```
[Boost provides a string algorithm for this](https://www.boost.org/doc/libs/1_70_0/doc/html/boost/algorithm/to_lower.html):

#include <boost/algorithm/string.hpp>

std::string str = "HELLO, WORLD!";
boost::algorithm::to_lower(str); // modifies str

[Or, for non-in-place](https://www.boost.org/doc/libs/1_70_0/doc/html/boost/algorithm/to_lower_copy.html):

#include <boost/algorithm/string.hpp>

const std::string str = "HELLO, WORLD!";
const std::string lower_str = boost::algorithm::to_lower_copy(str);
```

But I leave that to you... If you prefer to stick with this version we merge it as is

Syres916:
Thanks for the feedback Yorik but I did try the boost method and failed miserably, so I'm happy to merge as is.

yorikvanhavre:
ok done!

PR #9724, opened 2023-06-04
Improved font handling on Navigation cube

### LINKED ISSUE #9672: [Problem] Replace uses of missing font family "FreeCAD NaviCube"

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: macOS 13.3.1
Word size of FreeCAD: 64-bit
Version: 0.21.0.33263 (Git)
Build type: Release
Branch: master
Hash: 864f99bf984c237ef38d9451fd6f3a5ec34e6eb4
Python 3.11.3, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.2, OCC 7.7.1
Locale: C/Default (C)
Installed mods: 
  * freecad.gears 1.0.0
  * A2plus 0.4.60k
  * kicadStepUpMod 10.17.1
  * toSketch 1.0.1
  * Assembly4 0.12.6
  * Assembly3 0.11.4
[/code]
```

### Subproject(s) affected?

None

### Problem description

The error message in the title appears in the report window:

1. after quite a few seconds (maybe 40) when I first ran this version of FreeCAD
2. on subsequent runs when I open Preferences (almost immediately)
3. on subsequent runs when I open a model file (almost immediately)

The full error message is:
Populating font family aliases took 270 ms. Replace uses of missing font family "FreeCAD NaviCube" with one that exists to avoid this cost.

### Anything else?

https://forum.freecad.org/viewtopic.php?p=684168#p684168

### Code of Conduct

Jolbas:
Adding anisotropy texture mapping for better readability on steep angles. Tip from aapo in this post https://forum.freecad.org/viewtopic.php?p=657880#p657880
Before:
Skärmavbild 2023-06-04 kl  10 20 38
After:
Skärmavbild 2023-06-04 kl  10 23 33

Don't save a font name to parameters if none is chosen in preferences
Skärmavbild 2023-06-04 kl  10 51 12

Fix #9672 because the "FreeCAD NaviCube" is no longer created.

### DIFF
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -401,12 +401,10 @@ int imageVerticalBalance(QImage p, int sizeHint) {
 
 void NaviCubeImplementation::createCubeFaceTextures() {
     int texSize = 192; // Works well for the max cube size 1024
-    // find font sizes
     QFont font;
-    if (!m_TextFont.empty()) {
-        QString fontString = QString::fromStdString(m_TextFont);
-        font.fromString(fontString);
-    }
+    if (m_TextFont.empty()) font.fromString(QStringLiteral("Arial"));
+    else font.fromString(QString::fromStdString(m_TextFont));
+    font.setStyleHint(QFont::SansSerif);
     if (m_FontWeight > 0) {
         font.setWeight(convertWeights(m_FontWeight));
     }
@@ -454,9 +452,10 @@ void NaviCubeImplementation::createCubeFaceTextures() {
             delete m_LabelTextures[pickId].texture;
         }
         m_LabelTextures[pickId].texture = new QOpenGLTexture(image.mirrored());
-        m_LabelTextures[pickId].texture->generateMipMaps();
+        m_LabelTextures[pickId].texture->setMaximumAnisotropy(4.0);
         m_LabelTextures[pickId].texture->setMinificationFilter(QOpenGLTexture::LinearMipMapLinear);
         m_LabelTextures[pickId].texture->setMagnificationFilter(QOpenGLTexture::Linear);
+        m_LabelTextures[pickId].texture->generateMipMaps();
     }
 }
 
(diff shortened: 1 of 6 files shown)


xtemp09:
@Jolbas, what do you think about addition of shadow or mirror effect for NaviCube? Can it easily be implemented?

Jolbas:
@xtemp09 I'm not sure what you mean and in which way it would improve the navigation cube.

xtemp09:
I meant something like this:

cube

Can you make the NaviCube cast shadow?

adrianinsaval:
I don't see a benefit to that and this is not the place to discuss feature requests

src/Gui/NaviCube.cpp
```diff
@@ -454,9 +452,10 @@ void NaviCubeImplementation::createCubeFaceTextures() {
             delete m_LabelTextures[pickId].texture;
         }
         m_LabelTextures[pickId].texture = new QOpenGLTexture(image.mirrored());
-        m_LabelTextures[pickId].texture->generateMipMaps();
```
adrianinsaval, line 457: do we now if this is safe to use for example with noveau, old cards, arm machines?

Jolbas: I think you ment the line with: `m_LabelTextures[pickId].texture->setMaximumAnisotropy(4.0);`?
There should be a check built in QT but there seems to be a bug reported twice but not resolved. https://bugreports.qt.io/browse/QTBUG-47091 and https://bugreports.qt.io/browse/QTBUG-44604
I could do this to be safer: 
```
if (m_LabelTextures[pickId].texture->hasFeature(QOpenGLTexture::AnisotropicFiltering))
    m_LabelTextures[pickId].texture->setMaximumAnisotropy(4.0);
```

Jolbas: The extension is listed as ubiquitous here: https://www.khronos.org/opengl/wiki/Ubiquitous_Extension

adrianinsaval: excellent, should be acceptable then

PR #9734, opened 2023-06-06
[TechDraw] allow to change the arrow size per dimension object in the property panel

grubertechnic:

---


src/Mod/TechDraw/Gui/QGIViewDimension.cpp
```diff
@@ -2624,7 +2626,13 @@ double QGIViewDimension::getDefaultArrowTailLength() const
 {
     // Arrow length shall be equal to font height and both ISO and ASME seem
     // to have arrow tail twice the arrow length, so let's make it twice arrow size
-    return QGIArrow::getPrefArrowSize() * 2.0;
+    auto arrowSize = PreferencesGui::dimArrowSize();
+    auto vp = static_cast<ViewProviderDimension*>(getViewProvider(getViewObject()));
+    if (vp) {
```
benj5378, line 2631: Note that vp can only be null pointer if `getViewProvider(getViewObject())` returns nullptr

benj5378: I never touched C++ before I started doing FreeCAD 😄

I don't think the nullptr check is any issue. static_cast is done at compile time, like I don't think it causes anymore instructions during runtime, so maybe it is more efficient than checking getViewProvider has returned nullptr... in any case, the performance gain is probably 0,01 ms.

Question is, if getViewObject can even return nullptr... but I don't see any issues with the code. Though I prefer explicitely putting the type of `arrowSize` instead of auto, as I can't decode it's type otherwise like I can with `vp`.

> Checking for nullptr seems to me not a clean solution, since vp is auto and when changing the type of getViewProvider(), this seems to also not to work.

I am not sure I completely get what you mean here... but for sure, TechDraw code has so many nullptr checks... sometimes I wonder, if the functions can even return nullptr...

WandererFan: a) pretty sure Gui::Document::getViewProvider(obj) can return nullptr if obj does not have a vp.

b) auto x = xxx_cast<type>(y);
seems to be the format that is preferred these days.  I guess using the IDE to tell us the type of x is expected.

PR #9740, opened 2023-06-07
PartDesign: Clean length validation for pad

FlachyJoe:
Very small fix for a exception text.
Zero and negative length is allowed for L2. The exception text was false.

Ref: https://forum.freecad.org/viewtopic.php?t=78862

---

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePad.cpp
+++ b/src/Mod/PartDesign/App/FeaturePad.cpp
@@ -71,11 +71,10 @@ App::DocumentObjectExecReturn *Pad::execute()
 {
     // Validate parameters
     double L = Length.getValue();
-    if ((std::string(Type.getValueAsString()) == "Length") && (L < Precision::Confusion()))
-        return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Length of pad too small"));
     double L2 = Length2.getValue();
-    if ((std::string(Type.getValueAsString()) == "TwoLengths") && (L < Precision::Confusion()))
-        return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Second length of pad too small"));
+    std::string type = Type.getValueAsString();
+    if ((type == "Length" || type == "TwoLengths") && (L < Precision::Confusion ()))
+        return new App::DocumentObjectExecReturn (QT_TRANSLATE_NOOP("Exception", "Length of pad too small"));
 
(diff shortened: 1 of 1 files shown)


src/Mod/PartDesign/App/FeaturePad.cpp
```diff
@@ -71,11 +71,10 @@ App::DocumentObjectExecReturn *Pad::execute()
 {
     // Validate parameters
     double L = Length.getValue();
-    if ((std::string(Type.getValueAsString()) == "Length") && (L < Precision::Confusion()))
-        return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Length of pad too small"));
     double L2 = Length2.getValue();
-    if ((std::string(Type.getValueAsString()) == "TwoLengths") && (L < Precision::Confusion()))
```
adrianinsaval, line 77: in any case I think here we want to check `L+L2 < Precision::Confusion()`

FlachyJoe: it already fails here:
https://github.com/FreeCAD/FreeCAD/blob/bf8c59ba2ab4a2e6bd2060544130a0a4693ba4d8/src/Mod/PartDesign/App/FeatureExtrude.cpp#LL168C1-L174C1

adrianinsaval: I that case I think we should just remove this check, IMO abs(L2)>0 and L==0 is a valid input

FlachyJoe: And L2 < L1 < 0 should be valid too.

adrianinsaval: I think L1 is restricted to positive values but not by checks

FlachyJoe: You're right, see f9d153818 which remove the positive constraint for L2.
Negative L1 is similar to reverse mode but it could be helpful when work with expressions.

adrianinsaval: I think it's a little too redundant with the "Reverse" option, would like to hear more opinions on the matter, if you think we should enable this please create an issue and let's discuss there

adrianinsaval:
I removed this validation at #9771 and also for pockets

About allowing negative numbers for `Length`, if you want to pursue that route please create an issue so that it may be discussed better and with others too

PR #9766, opened 2023-06-13
FEM: let gmsh infer format from file name

jedbrown:
I sometimes use the FEM workbench to create meshes for a problem that I'll solve with an external FE solver that doesn't yet have workbench integration, or to prepare a Gmsh file for tweaks from directly running Gmsh. The `.unv` format is pretty limited on technical grounds so I rename the file to `.msh` (can express everything Gmsh can) or a parallel-friendly format. Explicitly setting `Mesh.Format = 2` is confusing because this line also needs to be fixed (or deleted) when renaming the output file name.

## Alternative
I would consider going further and also removing the `Save "Shape_Mesh.unv"` line from the file, then invoking as `gmsh -3 path/to/shape2mesh.geo -o path/to/Shape_Mesh.unv`. The reason is that it's ready for a person/script to invoke `gmsh` directly or to load the `shape2mesh.geo` for interactive graphical work without forcing a save on each refresh.

---

---

### DIFF
--- a/src/Mod/Fem/femmesh/gmshtools.py
+++ b/src/Mod/Fem/femmesh/gmshtools.py
@@ -869,7 +869,6 @@ class GmshTools():
 
         # save mesh
         geo.write("// save\n")
-        geo.write("Mesh.Format = 2;\n")  # unv
         if self.group_elements and self.group_nodes_export:
             geo.write("// For each group save not only the elements but the nodes too.;\n")
             geo.write("Mesh.SaveGroupsOfNodes = 1;\n")
--- a/src/Mod/Fem/femtest/data/elmer/group_mesh.geo
+++ b/src/Mod/Fem/femtest/data/elmer/group_mesh.geo
@@ -41,7 +41,6 @@ Mesh  3;
 Coherence Mesh; // Remove duplicate vertices
 
 // save
-Mesh.Format = 2;
 // Ignore Physical definitions and save all elements;
 Mesh.SaveAll = 1;
 Save "/tmp/tmpjVhNNb.unv";


jedbrown:
Fixed the `group_mesh.geo` reference.

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9766">pipeline status</a> for feature branch [PR_9766](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9766). Pipeline [903574346](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903574346) was triggered at [ff787c9](https://github.com/FreeCAD/FreeCAD/pull/9766/commits/ff787c96722f568e3f5299a34976e5afc3a5cc7e). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

berndhahnebach:
I have not tested this. The FreeCAD gmsh tools needs unv as mesh format because FreeCAD can not read msh mesh files and on gmsh mesh tool the unv is read back into FreeCAD. If you wourl rename it to msh it can not be read back into FreeCAD AFAIK.

jedbrown:
The default is `Mesh.Format = 10` (auto), which writes unv format when the file name ends in `.unv`, as it does now. Deleting the line as proposed here has no effect on behavior in the workbench.

I have other use cases where I prepare a model using the workbench, then hack the `.geo` file to use with a solver that isn't yet supported by the workbench. (I plan to contribute mine, but lots of research codes support gmsh formats and some things are easier in FreeCAD than importing STEP directly into Gmsh.) That `Mesh.Format = 2` line has been a recurring hiccup for myself and students. The *Alternative* proposal (which I'd be happy to implement) would be nicer for automation since the `.geo` file would not have to change, and also faster for interaction using the Gmsh GUI.

berndhahnebach:
Ahh I did not know about this. Thanks for explaining. Your PR totally makes sense now.

BTW: all this geo was codes at a time when there was not python interface to gmsh. AFAIK up to date versions of gmsh could be scripted by python which could make things much easier than the use of a geo file. But this is another story.

berndhahnebach:
no noise anymore on new PRs like #9810

berndhahnebach:
merged

jedbrown:
Regarding using the Python interface to Gmsh, I think there is benefit to separate processes because I've sometimes crashed gmsh and you don't want that to also crash the freecad process. Meshing can be very expensive and you'd rather it not lock up the interface and that you can break it if it's taking too much time or memory. So you could use a subprocess to call Gmsh's Python interface, but I'm not sure how much benefit that has and it's nice to be able to drive Gmsh directly sometimes. (We can take this conversation elsewhere if it needs anything more.)

PR #9663, opened 2023-05-25
Improve VSCode integration

Pesc0:
This PR:

- Improves includepath configuration, can now be customized per-OS
- Makes configuration general, not tied to conda. 
- C++ and python debugging should now work automatically within VSCode. Tested only on linux: can someone test on windows and mac please?
- Tests (googletest) are automatically discovered and presented in VSCode. It is not recommended to use this feature, it's way faster to run the gtest executable directly. Python tests are run using freecad, so they cant be easily integrated in vscode, there's just a simple task for those.
- Upgrade integrated python debugger from ptvsd (deprecated) to debugpy. 

When all the things are fixed/approved I'll create a followup pr on the developers handbook on how to use all these features and get started coding in vscode.

PR #9773, opened 2023-06-14
cmake: enable BUILD_FLAT_MESH by default

adrianinsaval:

---

https://github.com/FreeCAD/FreeCAD/pull/9663#issuecomment-1565420374

it would seem it's not necessary to keep this off by default anymore

@wwmayer do we need to change something else? is pybind something that is not normally used? would it be ok to use it by default?


freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9773">pipeline status</a> for feature branch [PR_9773](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9773). Pipeline [903572416](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903572416) was triggered at [93f9a34](https://github.com/FreeCAD/FreeCAD/pull/9773/commits/93f9a34b408a8c595cd49483ea42f8a0784ff90c). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

wwmayer:
> do we need to change something else?

Should be fine.

> is pybind something that is not normally used?

pybind usually isn't installed on a user's machine.

The user has two options for the Python binding: pybind or boost.python and what is used depends on the option FREECAD_USE_PYSIDE

> would it be ok to use it by default?

pybind is more powerful and has less overhead than boost.python. So, for me it's fine to prefer pybind over boost.python by default. In this case FREECAD_USE_PYSIDE should then be set to ON by default.

adrianinsaval:
perhaps you meant to say `FREECAD_USE_PYBIND11`?

wwmayer:
> perhaps you meant to say FREECAD_USE_PYBIND11?

Thanks for the correction. Yes, I meant FREECAD_USE_PYBIND11

PR #9781, opened 2023-06-15
GUI: Update Splashscreen for 0.21

chennes:
New splasher!


yorikvanhavre:
The idea of the splashscreen not containing the "FreeCAD" name was for branding reasons ( https://wiki.freecad.org/Branding ). So to stick with this it might be better to remove the name from the splash image.

However, really it makes little sense for someone to rebrand FreeCAD, and still use the default splash screen. So we could consider indeed stop printing the name on top programmatically (IIRC we had that back-and-forth already). Also it permits people to design the way the name FreeCAD is shown, but I don't know if that's something we want or not (if we would prefer a "standard" way to write the FreeCAD name)

There's probably more discussion to be had there, maybe next time that's something we should consider in the splash requirements... For me we can use this as is now.

chennes:
My thought process was the same: if you want to rebrand, you are almost certainly going to replace the Splashscreen, so it was OK for us to include "FreeCAD" there (and it looks much nicer done that way, particularly on hiDPI screens). The branding stuff is starting to get a little ugly though, it's a bit hackish. It would be nice to revisit before 1.0.

yorikvanhavre:
Yes, probably we need to rethink a bit the whole branding thing, and reflect better on what we want it to be... Right now it feels a bit like something left there but not really either supported nor encouraged

freecadci:
<a href="https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9781">pipeline status</a> for feature branch [PR_9781](https://gitlab.com/freecad/FreeCAD-CI/-/commits/PR_9781). Pipeline [903567747](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines/903567747) was triggered at [b472f08](https://github.com/FreeCAD/FreeCAD/pull/9781/commits/b472f08f7cdacbb006c8f5de7cc1c090361a6e16). All CI [branches](https://gitlab.com/freecad/FreeCAD-CI/-/branches/all) and [pipelines](https://gitlab.com/freecad/FreeCAD-CI/-/pipelines?scope=branches).

PR #9783, opened 2023-06-16
Navicube: Fix rotation direction on north and south buttons

Jolbas:
https://forum.freecad.org/viewtopic.php?t=79030


Jolbas:
@Rexbas Can you confirm this is the best place to make the correction? I did check that ArrowNorth actually is above the cube and ArrowSouth is below.

Rexbas (approved):
Yes, it works, this is indeed the right place to change the direction for the rotations.

Issue #5582, opened 2022-02-07
Constraints creation "on the fly"


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=1416

* **Reporter:** Pauvres_honteux
* **Date submitted:** 2/16/2014
* **FreeCAD version:** 
* **Category:** Feature
* **Status:** new
* **Tags:** constraints, Solver

# Original report text

Hi, I have a wish of getting a constraints creation tool that consists/contains of all possible constraints commands avaible and starts running after double clicking a general constraints icon.

Similar principle behavior as in a well known french cad program.

I.E. after this general constraints tool has started you just click on the two lines/points you want to constrain and the following behavior is the same as now with the addition of a right click meny where one can select submatters to that specific constraint if applicable. And here's the beauty in it: you do not need to move your cursor up to and click the constraint icon for every constraint you want to create. This saves your hand from getting over strained and speeds up your work - a lot.

To stop it running one could click the general constraints icon again or press ESC.

Regards Pauvres_honteux

# Additional information

Had this in the "Open discussion" before but it was apparently the wrong place.

# Other bug information

* **Priority:** normal
* **Severity:** feature
* **Category:** Feature
* **Updated:** 6/9/2021

# Discussion from Mantis ticket

### Comment by Pauvres_honteux 2014-05-18 07:01
One more who finds this principle behavior to be of interest:
http://www.freecadweb.org/tracker/view.php?id=1518

---

### Comment by Kunda1 2017-05-22 13:45
Abdullah OP seems to be referencing a ticket you worked on #1518 as similar. Can you weigh in?

---

### Comment by abdullah 2017-05-22 22:00
Well, I did the Continuous geometry mode you refer to. Jxnd did the continuous constraint creation mode (you click the constraint icon and now you select elements one after the other to which you want to apply the constraint, or pairs, or triplets depending on the constraint).

I am not sure how that "french cad program" works. Apparently the reporter is asking for a generalisation of what jxnd did. 

I am not sure I understand the part of "a general constraints tool". If it is general, how does it know which constraint you want to apply? by guessing? The right click menu initially I thought was to change from one constraint to another, or select from a computer generated guess list. However, it says submatters, so it seems some kind fine tunning?

A while ago I had an idea (never implemented, it will be in the forum somewhere) called "constrainator" which was guessing constraints based on the relative position of the geometry. This ticket reminded me of that.

[...]

PR #9810, opened 2023-06-22
Sketcher: Constrain Contextual implementation V2

PaddleStroke:
You may remember this Constrain Contextual tool that was in the tool setting branch on which we worked with @abdullahtahiriyo. As tool setting branch is very big and has been pending for a long time, I decided to take this out of that branch as it does not really need tool settings and can merge separately.

So here is constrain contextual, rebased, simplified, finalized. Also, thanks to the experience I accumulated since, I reviewed it and fixed all kind of things.

Here is the forum discussion : https://forum.freecad.org/viewtopic.php?style=4&t=65521 (videos in the thread to have an idea what the tool is like)

**PR description (to make it easier to review) :** 
What it does : 
- Adds constrain contextual, a sketcher constraining tool. The idea is that you have a single tool that replaces all the individual dimensioning tools. The tool apply the correct constraint depending on what you click. This tool only works in the workflow tool->selection (user select tool first before selecting what he wants to constrain).
- Adds 2 settings in sketcher preferences : One for the user to choose if he wants this tool or separated (currently existing) tools. A second setting let him choose how Constrain Contextual behaves for circles and arcs.

How it works : 
- The tool initially opens a command. Then the user can select geometries. Every time a geometry is selected, the tool will check what constraint is available for the current selection and will add this constraint. If the user select an additional geometry, the command is aborted which remove previously added constraint and it opens a new command.

If the reviewer wants, we can video call so that I can walk him through the code to make it easier/faster to understand.

> Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.
> - [ ]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR
> 
> Please remember to update the Wiki with the features added or changed once this PR is merged.  
> **Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).
> 
> ---
>


src/Mod/Sketcher/Gui/SketcherSettings.ui
```diff
@@ -163,6 +163,56 @@ Requires to re-enter edit mode to take effect.</string>
...
+       <widget class="QLabel" name="dimensioningLabel">
+        <property name="text">
+         <string>Dimensioning constraints:</string>
+        </property>
+       </widget>
+      </item>
+	  <item row="0" column="1">
+       <widget class="QComboBox" name="dimensioningMode">
+        <property name="toolTip">
+         <string>Selection the type of dimensioning constraints for your toolbar:
```
luzpaz, line 187: Sounds wrong 'Selection the type..' maybe 'Select the type...'

src/Mod/Sketcher/Gui/SketcherSettings.ui
```diff
@@ -163,6 +163,56 @@ Requires to re-enter edit mode to take effect.</string>
...
+'Single tool': You will have a single tool for all dimensioning constraints : Distance, Distance X / Y, Angle, Radius.
+'Separated tools': You will have separated tools for each dimensioning constraint.
+'Both': You will have both the 'Constrain Contextual' tool and the individual tool.
```
luzpaz, line 190: maybe we can remove 'You will have...' ?

src/Mod/Sketcher/Gui/SketcherSettings.ui
PaddleStroke: Thanks !

PaddleStroke:
@adrianinsaval if you feel empty-handed now that 0.21 shipped 😄

howie-j:
I have used this tool/pr now for a few weeks, but only for very minor work and i have not tested it extensively.

The crash on sketch-exit-with-contextualconstraint-active is a major bug that must be fixed, but if i remember correctly the first PR also had this bug in the beginning, but were fixed later.

Overall the tool works VERY well, and is imho the most important feature sketcher is missing. It cleans up the gui nicely, simplifies workflow for everyone (both selection->command and command->selection workflows), enables simplifying the overwhelming sketcher shortcuts, and prevents misclicks from selecting the wrong constraint.

Feedback:
- The alternative constraints using SHIFT is perhaps not the most ideal (point-point gives dimension, horizontal and vertical, not coincidence etc), but i don't think it's a big issue and can be corrected with another PR later
- The icon is perhaps a bit overloaded. I believe other cad systems often use our "length" symbol for their dimensioning tools
- The name is a mouthful as said earlier :)
- IF, and only IF there is pushback on this pr, the old icons could be placed in a dropdown menu. Personally i dont feel the need for that.

So from a users perspective i would say great work @PaddleStroke, really hope this can get merged soon.

PaddleStroke:
Thanks for the feedback!

- For the name I agree. I think "Dimensioning" is a better name. I'll
change it.
- yes the bug must be fixed. I don't recall that we fixed it before. I need
to re-read the old discussion then.
- yes I removed many constraint from the shift key (such as 2 point
coincidence I seem to recall). The main reason was to simplify the code by
a lot, making it easier to review. Also I think what is really important is
the core of the functionality. The many shift-modes were secondary. And we
can add them later.

ifohancroft:
>     * IF, and only IF there is pushback on this pr, the old icons could be placed in a dropdown menu. Personally i dont feel the need for that.

I'm not entirely sure what this means, but just in-case it means what I think it means:

The icons/toolbar buttons for the regular constrain tools shouldn't be affected in anyway, whatsoever, by this tool, if you have no plans to use this tool and have it disabled from the settings, or be affected at all by it for that matter.

howie-j:
> ```
> IF, and only IF there is pushback on this pr, the old icons could be placed in a dropdown menu. Personally i dont feel the need for that.
> ```
> 
> The icons/toolbar buttons for the regular constraint tools shouldn't be affected in anyway, whatsoever, by this tool, if you have it disabled from the settings.

I just checked, and there is three options in preferences:
PR9810

So all parties should be happy. Forget my comment about dropdown icon.

ifohancroft:
Regarding the second option - choosing the diameter/radius mode - it only takes effect if the first option isn't set to 'Separate tools', right?
The setting shouldn't be affecting the default behavior if the tool is disabled.

ifohancroft:
Btw what is the current default regarding a single tool vs multiple?
If it's single - would that break existing configurations/toolbar arrangements?

PaddleStroke:
The default would be single tool.
Disabling by default would mean most users wouldn't even be aware of it. And this is a really big improvement. Having separated tools makes no sense, they are a very inferior solution. If we were a commercial app and I was making decision I would remove separate tools completely. But this is FreeCAD and you can't annoy the 0.1% of veterans who will still want separate tools because that's what they used for a long time.

howie-j:
Relevant GH Issue: https://github.com/FreeCAD/FreeCAD/issues/5582

PaddleStroke:
@wwmayer Would you perhaps have a bit of time to check this PR in terms of code architecture ?
This feature is in the waiting list for 1.5 year now and Abdullah is still away and will be swamped with things to do when he'll be back. That would be really appreciated.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -990,6 +990,1165 @@ void CmdSketcherConstraint::activated(int /*iMsg*/)
     getSelection().clearSelection();
 }
 
+
+// Dimension tool =======================================================
+
+class DrawSketchHandlerDimension : public DrawSketchHandler
```
wwmayer, line 996: All methods of this class (including destructor) that overrides methods of the base class should be declared as `override`. The keyword `virtual` is superfluous for these methods and should be removed.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -990,6 +990,1165 @@ void CmdSketcherConstraint::activated(int /*iMsg*/)
...
+        }
+    }
+
+    virtual bool pressButton(Base::Vector2d onSketchPos)
+    {
+        return true;
+    }
+
+    virtual bool releaseButton(Base::Vector2d onSketchPos)
+    {
```
wwmayer, line 1128: It doesn't harm to refactor this function.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -990,6 +990,1165 @@ void CmdSketcherConstraint::activated(int /*iMsg*/)
...
+            return false;
+        };
+
+        return !contains(selPoints, elem)
+            && !contains(selLine, elem)
+            && !contains(selCircleArc, elem)
+            && !contains(selEllipseAndCo, elem);
+    }
+
+    bool makeAppropriateConstraint(Base::Vector2d onSketchPos) {
```
wwmayer, line 1300: This method has almost 300 lines and must be refactored.

src/Mod/Sketcher/Gui/SketcherSettings.cpp
```diff
@@ -66,6 +66,42 @@ void SketcherSettings::saveSettings()
...
     ui->checkBoxNotifyConstraintSubstitutions->onSave();
     ui->checkBoxAutoRemoveRedundants->onSave();
+
+    // Dimensioning constraints mode
+    ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Mod/Sketcher/dimensioning");
+    bool singleTool = true;
+    bool SeparatedTools = false;
+    int index = ui->dimensioningMode->currentIndex();
+    switch (index) {
+    case 1:
```
wwmayer, line 76: Avoid magic numbers and instead choose some speaking enumerators. Same applies below...

src/Mod/Sketcher/Gui/CommandConstraints.cpp
PaddleStroke: This function is big but it is intended to be a 'map' of the constraint that you get depending on selection. It is intended to be looked at folded as such : 
image
Then you can open the case that you want to edit.

I don't really see how to refactor it in a way that would make more sense than this.

If you have an idea how this should be split let me know.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
PaddleStroke: Got it

src/Mod/Sketcher/Gui/CommandConstraints.cpp
PaddleStroke: Please let me know if it is now ok.

src/Mod/Sketcher/Gui/SketcherSettings.cpp
PaddleStroke: Please have a look at the commit [79bb68e](https://github.com/FreeCAD/FreeCAD/pull/9810/commits/79bb68ef142f9273753954fd356ec00f3a64a2af) to let me know if this is what you had in mind. I don't see how we can use modern class enum here so I used legacy enums. Let me know.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
berniev: Copy pasting the same code over and over makes it extremely difficult to comprehend or maintain (DRY = Don't Repeat Yourself, WET = Write Everything Twice). When the same code repeats then for sure the parameters can be placed in a simple container, perhaps constexpr, and have just one copy of the algorithm. Through the whole 300 odd lines there is IMHO much repetition that could be refactored out.

I can't remember who it was said something like "Show me code and my eyes will glaze. Show me the data and I won't need to look at the code."

src/Mod/Sketcher/Gui/SketcherSettings.cpp
wwmayer: Old-style enums are fine because it anyway is only an implementation detail, i.e. not part of the API this class.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
wwmayer: Much better now, thanks!

src/Mod/Sketcher/Gui/CommandConstraints.cpp
PaddleStroke: Hmm I'm still not sure how it can be refactored. Here it's convenient and easy to maintain. It's easy to find a bug. Say distance between 2 points has an issue, you go to the right section and find it easily.

I simplified it a little bit by removing unecessary tests in [00cb724](https://github.com/FreeCAD/FreeCAD/pull/9810/commits/00cb7243413ed2f514e790564116259e8980f09a)
image

PaddleStroke:
@wwmayer Once you confirm there is no major architecture flaws that require to rewrite it all, I will chase the crash that happen when edit-mode is left while tool active.

wwmayer:
> no major architecture flaws

There are no architecture flaws but I haven't looked very closely at the code. So, I don't know if there are any hidden bugs.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
berniev: > Hmm I'm still not sure how it can be refactored

On that basis I trust you are open to suggestions.

A first step could be to rationalise obvious repetitions (DRY principle) e.g.:
```
auto sp {selPoints.size()};
auto sl {selLine.size()};
auto sc {selCirCircleArc.size()};
auto se {selEllipseAndCo.size()};
```
I don't know about you, but my brain hurts if it has to read what appears to be the same thing over and over knowing subtle differences are easy to miss.

Next step, move duplicate condition code into lambdas, passing parameters. Something like
```
// reusable utility
auto matchGt = [](auto val, bool targ, bool gt){
    return gt ? val >= arg: val > targ;
};

auto checkSp = [&](bool spGt, auto spTarg, auto slTarg, auto scTarg, auto seTarg){
    return matchGt(sp, spGt, spTarg) && sl == slTarg && sc ==scTarg && se == seTarg;
}

// the work to be done
auto funcA = [](){};

if(sp > 0){
         if(checkSp(false, 1, 0, 0, 0)){ funcA(); }
    else if(checkSp(false, 2, 0, 0, 0)){ funcB(); }
    else if(checkSp(false, 1, 1, 0, 0)){ funcC(); }
    else if(checkSp(false, 3, 0, 0, 0)){ funcD(); }
    else if(checkSp(true,  4, 0, 0, 0)){ ... }
    else if(checkSp(false, 2, 1, 0, 0)){ ... }
    else if(checkSp(true,  3, 1, 0, 0)){ ... }
    else if(checkSp(true,  1, 0, 1, 0)){ ... }
    else if(checkSp(true,  1, 0, 0, 1)){ ... }
}
```
The parameters for the code, settings if you like, are being progressively moved away from the code. Code folding not required.

The conditions for checkSp can now be seen as values (set of params) which could be stored in e.g. constexpr container and looped over. More repetition is thus removed from the list of if/elses. Perhaps the loop could return an int that is used as the key of a switch to call the various funcs.

A next step could be to pass the func into the check so it is automatically called on a match. etc etc you get the idea.

With each refactor step the problem is broken into smaller and smaller chunks of reusable code. Rinse and repeat ..

As a bonus if lambdas are instead separate functions they become individually testable.

PaddleStroke: For the first point I agree it's a easy simplification that makes a lot of sense.

For the lambda, imo this is building complexity for not much gain and it is much less flexible. First any geo type has test that are not equality. So you would need not 1 but 4 boolean. Then what if someone wants to modify and he wants to put a case where he test if se <= sl?

Regarding splitting into individual functions each case, I'm ok with it. I'm not sure it's so great because that means there's going to be 20 more functions which is a lot. But yes if they are individually testable it's probably best

PaddleStroke: `makeAppropriateConstraint` has been refactored. @wwmayer please let me know if this is now acceptable. Thanks @berniev for the ideas

PaddleStroke:
The crash is fixed. It appears that it actually happened only when you exited by using the 'close' button. LeaveSketch command was working fine.
Looking at `CmdSketcherLeaveSketch` I found that the command purge the handle first while TaskDlgEditSketch::reject() didn't. Purging the handler first in Task reject solves the bug.

PaddleStroke:
> There are no architecture flaws but I haven't looked very closely at the code. So, I don't know if there are any hidden bugs.

@wwmayer  Great thanks. With a big feature like this it's going to be impossible to catch every single feature related bug. But it went under a lot of testing with many forum users so most feature bugs should be fixed already. If more appear I'll fix them asap.

So I would suggest just making sure there are no code-issues, like a memory leak or something not done properly (for instance in the deactivated and the restartCommand mechanism).

src/Mod/Sketcher/Gui/CommandConstraints.cpp
berniev: > though 2 letters variables are going to make the CI complain

Actually, I think you'll find two-letter are ok, one-letter vars is an issue with clang-tidy.

> Thanks @berniev for the ideas

No worries. Happy to encourage more modern approaches.

> what if someone wants to modify and he wants to put a case where he test if se <= sl

Deal with what you got. YDNIY. Can be added later as reqd.

wwmayer: It's in an acceptable state now. Nevertheless, one could reduce the complexity even a bit more by writing a little helper class that provides functions for the `if` conditions.
```
class GeomSelection
{
    size_t s_pts, s_lns, s_cir, s_ell;
public:
    GeomSelection(size_t s_pts, size_t s_lns, size_t s_cir, size_t s_ell)
...
   bool hasPoints() const {
       return s_pts > 0;
   }
...
   bool has1Point const {
       return s_pts == 1 && s_lns == 0 && s_cir == 0 && s_ell == 0;
   }
   bool has2Points const {
       return s_pts == 2 && s_lns == 0 && s_cir == 0 && s_ell == 0;
   }
...
};
```

PaddleStroke:
@wwmayer I have added this GeomSelectionSizes class as you described.
If this is all then perhaps we can finally merge this 😃

qewer33:
@PaddleStroke this might be better if the "Both" option (both contextual and other dimensioning tools) was enabled by default. Just having contextual in the toolbar might confuse some current users.

If this decision was made to save toolbar space, maybe a third option that combines all dimension tools into one dropdown button could be added. Just a suggessiton.

adrianinsaval:
the newly introduced circle to point distance constraint is not taken into account by this tool. Meaning one needs to go to the menu and search for the distance constraint in order to apply that constraint.

PaddleStroke:
Ah yes I will introduce that asap.

FlachyJoe:
@PaddleStroke it currently causes a crash with Arc+Something if arc radius is not set.

PaddleStroke:
@FlachyJoe I removed the arc+something mode

@qewer33 So yes it's a good question. I tought about making a comp command (dropdown) too. In the end I went with the preference setting so that user who don't like the new tool can go back to exactly how it was before.

Now regarding to the default of this preference, I think we must offer the best UI as the default. And with the both option there's a lot of tools doing the same thing, it will add confusion for the users.

PaddleStroke:
This is how a comp command would look like for reference : 
image

benj5378:
> this might be better if the "Both" option (both contextual and other dimensioning tools) was enabled by default. Just having contextual in the toolbar might confuse some current users.

Yes, I was very confused both yesterday and today.

And I am still confused; using the contextual dimension tool, I don't have a tool settings panel to input the dimension values?

image

ifohancroft:
> This is how a comp command would look like for reference : image

Sorry, what does comp command mean? I don't understand the comp part.

PaddleStroke:
@benj5378 yes the workflow changed compared to the video since Tool settings as not merged.
Though the new workflow is actually the same as other CAD software so I think it's better this way.

PaddleStroke:
@ifohancroft sorry this is a name that is used in the code, I think it's short for composition. It means a group command (dropdown). See https://github.com/FreeCAD/FreeCAD/pull/10463

adrianinsaval:
is this documented in the wiki?

PaddleStroke:
Not yet no. Perhaps it can wait a little bit that the feature stabilize ?
What do you think?

adrianinsaval:
I guess so, I'm just mentioning since I see some people confused about it here. Do you plan to do more changes to it in the short term?

PaddleStroke:
Hmm. Maybe additional modes (like arc-arc distance...) but that's not
critical for the wiki I guess. Yes perhaps doing the wiki now is a good
idea.
I'll check that on Monday unless a wiki specialist want to jump in.

ifohancroft:
The argument that this new tool should replace the other constrain tools by default (I.e. that the default setting for the constraint tools choice should be single) because otherwise new/unaware users would be missing out is invalid.

This argument relies on the assumption that this new tool is the superior choice vs the regular constraint tools. Obviously, this assumption is false, since it's not a fact but it's up to personal interpretation. Not everyone feels this way.

Therefore, the default option should either be to leave things like they were (with this tool disabled by default, and only the regular constraints showing) or a both option where there is just another icon for this tool in the regular constraint toolbar.

PaddleStroke:
> This argument relies on the assumption that this new tool is the superior choice vs the regular constraint tools

It is very superior. Separate tools make so little sense that **no other CAD softwares** use this workflow. Solidworks? No. Catia? No. Fusion? No. OnShape? No... Industry standard accross all the CAD softwares is to offer such single tool for dimensions constraints. And for good reason: separated tools for dimensions are bad : 
- Force the user to learn 7 different tools/icons/shortcut for no reason. The user knows what type of constraint he wants to add. Forcing him to choose a correct tool to do so makes no sense.
- Even when you know the commands, it often happen to click DistanceX instead of DistanceY and it's very frustrating.
- The toolbars are so cluttered it is a pain for new user to disover
- The toolbars are so cluttered it is hard to find the correct tool for all users.

The single and only advantage that these separated tools have is that they have been here for a long time. So veterans who have spent hours upon hours of pain hardcoding the correct shortcuts into their brain don't want to change it.

Changing the default UI to 'Both' would be a perfect example of why FreeCAD UI is bad and is so often criticized all over the internet. And I think (but I may be wrong of course) that those who ask for the separate tools to come back are the loud minority. If we run a poll I wouldn't be surprised that 75%+ want to keep as it is.

PR #9835, opened 2023-06-28
Arch: Fixed error on null window preset creation parameters

yorikvanhavre:
Fixes yorikvanhavre/BIM_Workbench#116

---

### DIFF
--- a/src/Mod/Arch/ArchWindowPresets.py
+++ b/src/Mod/Arch/ArchWindowPresets.py
@@ -53,7 +53,7 @@ def makeWindowPreset(windowtype,width,height,h1,h2,h3,w1,w2,o1,o2,placement=None
         o2 = float(o2)
         # h1, h2, w1, w2 cannot be null (for now)
         # TODO allow these to be null (don't create the component if so)
-        if not h1 or not h2 or not w1 or not w2:
+        if h1*h2*w1*w2 == 0:
             FreeCAD.Console.PrintError("H1, H2, W1 and W2 parameters cannot be zero. Aborting\n")
             return
         # small spacing to avoid wrong auto-wires in sketch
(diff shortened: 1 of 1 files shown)


src/Mod/Arch/ArchWindowPresets.py
```diff
@@ -51,6 +51,11 @@ def makeSketch(windowtype,width,height,h1,h2,h3,w1,w2,o1,o2):
         w2 = float(w2)
         o1 = float(o1)
         o2 = float(o2)
+        # h1, h2, w1, w2 cannot be null (for now)
+        # TODO allow these to be null (don't create the component if so)
+        if not h1 or not h2 or not w1 or not w2:
+            FreeCAD.Console.PrintError("H1, H2, W1 and W2 parameters cannot be zero. Aborting\n")
+            return
```
0penBrain, line 58: Why not something just like
```
if h1*h2*w1*w2 == 0:
    FreeCAD.Console.PrintError("H1, H2, W1 and W2 parameters cannot be zero. Aborting\n")
    return
```
at the very beginning of the function ?

yorikvanhavre: Indeed that's more elegant. But in any case, this test should vanish in the future (see the TODO comment above)

yorikvanhavre:
```suggestion
        # h1, h2, w1, w2 cannot be null (for now)
        # TODO allow these to be null (don't create the component if so)
        if h1*h2*w1*w2 == 0:
            FreeCAD.Console.PrintError("H1, H2, W1 and W2 parameters cannot be zero. Aborting\n")
            return
```

0penBrain:
Code is good to me (commits could be squashed, and introduced check could be done sooner in the function though).

But I won't dare to merge as I absolutely know nothing about Arch. :smile:

yorikvanhavre:
I think we can make an exception, I'll merge this myself :)

PR #9836, opened 2023-06-28
fix: Spreadsheet applies content/alias even when focus lost

### LINKED ISSUE #9801: [Problem] spreadsheet apply content or alias with loss of focus

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Debian GNU/Linux 12 (bookworm) (GNOME/gnome)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33263 (Git)
Build type: Unknown
Branch: master
Hash: 864f99bf984c237ef38d9451fd6f3a5ec34e6eb4
Installed mods: 
  * Rocket 3.0.0
  * IconThemes
  * freecad.gears 1.0.0
  * A2plus 0.4.60n
[/code]
```

### Subproject(s) affected?

Spreadsheet

### Problem description

The spreadsheet editor allows the aliases and content of a cell to be changed in the top bar. However, when editing these values enter has to be clicked in order for the changes to apply. This can be annoying when trying to enter multiple aliases in a row that are being copied from somewhere else (e.i. only one hand is on the keyboard, the other is on the mouse).

For this reason, it would be nice if the fields would apply whenever the focus of the field was lost as well. I don't think this is a big change but it could improve usability and speed up the spreadsheet workflow.

Please let me know if I've missed something!

### Anything else?

I have a change for the code that already works, but I think the FreeCAD guidelines say that only PRs that fix previously identified errors should be created, so I figured I would create an issue and only create the PR if this is something that should be changed according to other people.

If I can just add PRs for such small fixes and do not need to create issues for them, let me know for the future :)

### Code of Conduct

RobbeDGreef:
The original stylesheet behavior was that the alias or content fields could be edited, but only after enter was pressed the field would actually update. This change makes the field apply when focus is lost on top of when enter is pressed. This makes it easier to enter the alias of a lot of fields at once. 
Closes #9801 

---


adrianinsaval:
I'm not entirely sure if this is desired behavior for the content field since invalid expressions could be problematic, would like some user testing on that. Agreed on the alias field

chennes:
@RobbeDGreef is it your plan to remove the content field, or would you like this considered as-is? I'd prefer to merge it prior to the 0.21 release, if possible.

PR #9847, opened 2023-06-30
Rename R2, R3 and R4 in salomesmesh to fix building on armel.

petterreinholdtsen:
Fix compilation on Armel due to coincident variables R2, R3 and R4 in <sys/ucontext.h>.

Based on suggestion by Paul Brook in
<URL: http://lists.alioth.debian.org/pipermail/debian-science-maintainers/2011-October/009876.html >.

Patch by Anton Gladky <gladk@debian.org>.

Patch has been used by the Debian edition of FreeCAD since 2011.


chennes:
As far as I am aware the only thing holding up this merge is agreement from the Maintainer team that we should be patching a 3rd party library. As far as I am aware there is precedent for our doing so, so I think this is fine to merge.

PR #9853, opened 2023-07-01
Disable memory-check in SMDS because sysinfo.h is not working on BSD.

petterreinholdtsen:
The SMDS memory check cause the build to fail on on BSD-systems. It would be better to disable it only on BSD-platforms, or use http://code.google.com/p/sysinfo-bsd/source/browse/sysinfo.c for them.

This fixes Debian bug https://bugs.debian.org/623560 .

Patch by Anton Gladky <gladk@debian.org>.

Been part of the Debian edition of FreeCAD since 2011.

### DIFF
--- a/src/3rdParty/salomesmesh/src/SMDS/SMDS_Mesh.cpp
+++ b/src/3rdParty/salomesmesh/src/SMDS/SMDS_Mesh.cpp
@@ -61,13 +61,6 @@
 #include <iterator>
 using namespace std;
 
-#ifndef WIN32
-#if !(defined(__MACH__) && defined(__APPLE__))
-#include <sys/sysinfo.h>
-#endif
-#include <sys/wait.h>
-#endif
-
 // number of added entities to check memory after
 #define CHECKMEMORY_INTERVAL 100000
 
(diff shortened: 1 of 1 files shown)


luzpaz:
Just out of curiosity, has the patch ever been submitted upstream ?

petterreinholdtsen:
[luzpaz]
> Just out of curiosity, has the patch ever been submitted upstream ?

I have no idea.  As far as I see it, FreeCAD is upstream for the Debian
package, so I made sure it was sent here to avoid it just lingering in
Debian.

-- 
Happy hacking
Petter Reinholdtsen

luzpaz:
Note: upstream here is the SMESH package (links above).

adrianinsaval:
I have thrown trash on the floor before, that doesn't make it a good idea to do it again. (Note: not comparing this PR to trash, just pointing out the flaw in the logic)

We _could_ merge this patch, but I don't think that's a good idea. Plus the patch itself removes an entire check rather than fixing it for the platform where it fails, or at least skipping it only on that platform. I would guess the check was there for a reason so it shouldn't just be removed without thought.

petterreinholdtsen:
[Adrián Insaurralde Avalos]
> Plus the patch itself removes an entire check rather than fixing it
> for the platform where it fails, or at least skipping it only on that
> platform. I would guess the check was there for a reason so it
> shouldn't just be removed without thought.

I suggest you have another look at the code.

It is worth noting that the "check" in question as far as I can tell is
already disabled on all architectures (note the #if 0), and always
return 1000;

It can also be noted that sysinfo() on FreeBSD do not behave like the
disabled implementation expect, see
<URL: https://man.freebsd.org/cgi/man.cgi?query=sysinfo&sektion=2&manpath=SunOS+5.7 >
for the manual page for sysinfo(2) on FreeBSD.

Perhaps the Debian patch is no longer needed?  I see the '#if 0' was
added in 99bdef75867e0dd84c374721396d3a0191324055 to get rid of the
error message about SMDS_MemoryLimit not found, which would originate
from the stupid system("SMDS_MemoryLimit") call.

-- 
Happy hacking
Petter Reinholdtsen

chennes:
@adrianinsaval can you revisit this? Any further comment?

chennes:
OK, I took a look at this -- the code that's been removed was already deactivated, so the only real effect of this PR is to remove some includes at the top that were no longer needed (they were only used by the disabled code). So this seems like a simply merge to me.

PR #9854, opened 2023-07-01
Skip open_de9b3fb438 s390x to avoid OOM.

petterreinholdtsen:
The load of the current file in test_open_head is fine, but open_de9b3fb438 goes into an OOM even in huge (e.g. 10GB) systemd. This probably needs a proper fix by upstream in regard to the migration modules that load the old code, but until then (since the rest works on s390x) this unblocks the package self-test in Debian & Ubuntu.

Patch by Christian Ehrhardt <christian.ehrhardt@canonical.com>.

This is Debian bug https://bugs.debian.org/984952 and Ubuntu bug https://bugs.launchpad.net/ubuntu/+source/freecad/+bug/1918474.

Been part of the Debian edition of FreeCAD since 2021.

### DIFF
--- a/src/Mod/Fem/femtest/app/test_open.py
+++ b/src/Mod/Fem/femtest/app/test_open.py
@@ -25,6 +25,7 @@ __title__ = "Open files FEM App unit tests"
 __author__ = "Bernd Hahnebach"
 __url__ = "https://www.freecad.org"
 
+import platform
 import tempfile
 import unittest
 from os.path import join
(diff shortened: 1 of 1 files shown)


adrianinsaval:
this is just hiding the issue, ideally one should identify what is actually causing the high memory consumption.

petterreinholdtsen:
[Adrián Insaurralde Avalos]
> this is just hiding the issue, ideally one should identify what is
> actually causing the high memory consumption.

While I am sure you are correct, it is worth noting that the goal of me
submitting the patch was to ensure the upstream project were aware of
the patches used in the Debian edition of FreeCAD and to get a
discussion started.  I have no idea about the inner workings of FreeCAD,
so any search for the culprit here will have to be conducted by someone
else than me. :)

If someone got patches that might address the problem, I can test on the
s390x machine available for Debian Developers, but only if it can be
done non-interactively without X access, as my only access is via large
latency ssh. :)

-- 
Happy hacking
Petter Reinholdtsen

PR #9855, opened 2023-07-01
Change XDG desktop entry to only run single instance of FreeCAD.

petterreinholdtsen:
This ensure invoking it to open different files only start the program once, and load new files into the already running instance.

Patch by Håvard Flaget Aasen <haavard_aasen@yahoo.no> and Tobias Frost <tobi@debian.org>.

Related to https://bugs.debian.org/1029846 and
https://github.com/FreeCAD/FreeCAD/commit/c7a21ecbeecefe7c2dfc9e950b3d6bb42351d476 .

Been part of the Debian edition of FreeCAD since 2021.


chennes:
Is it correct here to migrate to a hardcoded absolute path?

petterreinholdtsen:
[Chris Hennes]
> Is it correct here to migrate to a hardcoded absolute path?

It depend on what you want. It is done in a Debian context to ensure the
package binary is used, and not some random binary in PATH.

-- 
Happy hacking
Petter Reinholdtsen

adrianinsaval:
I don't think we want that hardcoded, and specially not as in this PR considering that we don't install the binary as freecad but as FreeCAD. In any case you could use cmake to adapt this to the actual install path of the binary.

petterreinholdtsen:
I dropped the renaming and hardcoding of the binary, to allow focus to stay the change in command line arguments.

yorikvanhavre:
using a single instance seems sensible, indeed it's what we would want by default. The extra - is maybe not needed anymore seeing @wwmayer 's comment on the debian bug, but apparently it does no harm.

However: starting freecad with `freecad --single-instance` (or `freecad - --single-instance`) makes the start page not load anymore for me here. Can someone else check?

chennes:
> However: starting freecad with `freecad --single-instance` (or `freecad - --single-instance`) makes the start page not load anymore for me here. Can someone else check?

Works find for me (on Mac OS).

wwmayer:
> However: starting freecad with freecad --single-instance (or freecad - --single-instance) makes the start page not load anymore for me here. Can someone else check?

Works fine for me, too (on Xubuntu 22.04).

yorikvanhavre:
> @yorikvanhavre what system are you testing on?

Debian testing. I get this in the terminal if I use `--single-instance`:

*WARNING Qt AtSpiAdaptor: Accessible invalid:  QAccessibleInterface(0x55ace61c8440 invalid) "/org/a11y/atspi/accessible/2147483729"*

chennes:
@yorikvanhavre does this latest patch address your issue?

yorikvanhavre:
No, no change (The latest commits don't actually change anything).

Note that this is probably an issue with the --single-instance argument that went undetected since now, and not a problem with this PR particularly. However, if we merge this we might get a swarm of issues if this happens on other systems as well...

wwmayer:
Do you have installed the package qt-at-spi? If yes, remove and try again.
This package has already caused several problems in the past.

yorikvanhavre:
> Do you have installed the package qt-at-spi? If yes, remove and try again.

That package apparently does not exist anymore.. There is now a non-qt one called at-spi2-core and it is part of Gnome, but the error message here specifically references Qt... Apparently qt accessibility stuff is now integrated in QtGui

Does the --single-instance arg work normally for you @wwmayer ? (ie. the start page loads normally)

wwmayer:
> Does the --single-instance arg work normally for you @wwmayer ? (ie. the start page loads normally)

Yes, it works normally and the StartPage is loaded.

yorikvanhavre:
I think we can consider this as a problem on my machine only then... And since this has been in Debian for the last 2 years, we would have heard if that was a problem... I think we can safely merge it.

@adrianinsaval I don't know if you had any more concern, for me this is good to go, I let you press the button. I'll try to find what's wrong here :)

src/XDGData/org.freecadweb.FreeCAD.desktop
```diff
@@ -14,7 +14,7 @@ GenericName[de]=CAD-Anwendung
 GenericName[ko]=CAD 응용프로그램
 GenericName[pl]=Aplikacja CAD
 GenericName[ru]=Система автоматизированного проектирования
-Exec=FreeCAD %F
+Exec=FreeCAD - --single-instance %F
```
adrianinsaval, line 17: seems like there's an extra dash here?

petterreinholdtsen:
[Adrián Insaurralde Avalos]
> seems like there's an extra dash here?

Correct.  All explained in the Debian bugs reports leading up to this
patch.  Is FreeCAD fixed so it is no longer needed?

-- 
Happy hacking
Petter Reinholdtsen

Issue #9538, opened 2023-05-10
[Problem] sometimes can't deselect things in newly created sketches


dominikh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: NixOS 23.05 (Stoat)
Word size of FreeCAD: 64-bit
Version: 0.21.0.Unknown
Build type: Release
Python 3.10.10, Qt 5.15.8, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.2
Locale: English/United States (en_US)
Installed mods: 
  * fasteners 0.4.56
  * ProDarkThemePreferencePack 1.0.0
[/code]
```

### Subproject(s) affected?

Sketcher

### Problem description

When creating a new sketch in a certain way it is not possible to deselect objects such as lines by clicking on them in the 3D view.

One exact sequence of operations to reproduce the issue:

1. Start FreeCAD. Using an almost default config this starts me off in the Start workbench
2. Ctrl+N to start a new document
3. Click the _Create Part_ button in the toolbar
4. Switch to the PartDesign workbench
5. Click the _Create sketch_ button in the toolbar
6. Click on the XY-plane in the 3D view
7. Click the _Create rectangles_ button in the toolbar and draw any rectangle
8. Press ESC to exit out of the rectangle tool
9. Click on an edge to select it
10. Click on the same edge again to deselect it. It will not work.

After exiting out of the sketch and going back in, selected edges can now be deselected by clicking on them.

Note that the sequence I described is one of multiple sequences that lead to the same issue.

### Anything else?

Rebasing points to 3e33a91aa9fc81e9d5f4c6f99181273761ee0ab1.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
Rebasing points to https://github.com/FreeCAD/FreeCAD/commit/3e33a91aa9fc81e9d5f4c6f99181273761ee0ab1

do you mean that you are running the version from hat commit or what?

your version information is incomplete as it is missing the revision and commit hash information

dominikh:
Sorry, I typoed. I meant to say that *bisecting* pointed to 3e33a91aa9fc81e9d5f4c6f99181273761ee0ab1.

Sorry about the incomplete version information; it doesn't seem able to embed that information with the way I am building it (via Nix). I was running yesterday's master branch, but the commit that seemingly causes it is much older, anyway.

[...]

PR #9863, opened 2023-07-02
Fix #9792: PD tasks fail to launch if body is inside Part and activated

### LINKED ISSUE #9792: [Problem] PartDesign tasks fail to launch when body is inside Part and activated

### Version

0.21 (Development)

### Full version info

```shell
OS: Debian GNU/Linux 11 (bullseye) (X-Cinnamon/lightdm-xsession)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33411 (Git)
Build type: Release
Branch: master
Hash: 127404d793d5c2081383b46a7b677426d9a0c5aa
Python 3.9.2, Qt 5.15.2, Coin 4.0.0, Vtk 9.0.1, OCC 7.7.1
Installed mods:
```

### Subproject(s) affected?

PartDesign

### Problem description

-new document
-create Part
-create Body, body is automatically placed inside Part however it is not highlighted as activated
(if you try now to create a sketch in this moment it works without problems)
-double click on Body to activate it
-attempt creating a sketch on body (or launch any PD task really), the task is not launched and the error shown in report view

> 14:57:31  <Gui> Document.cpp(385): Invalid sub object 'Unnamed#Part.Body.Box.Body.'

double clicking on the created sketch or feature successfully launchs the task.

### Anything else?

originally reported at https://forum.freecad.org/viewtopic.php?p=688246#p688246

and here the example file provided there (remove .zip extension):
[Yoke.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/11782390/Yoke.FCStd.zip)

a3c4f4bb01f5201daf666852444a8f9c5adbda8d is mentioned as introducing this problem. @wwmayer got any ideas?

If we can't fix this before release this commit should be reverted.

### Code of Conduct

adrianinsaval:
introduced by PR #9538

---

### DIFF
--- a/src/Mod/PartDesign/Gui/Utils.cpp
+++ b/src/Mod/PartDesign/Gui/Utils.cpp
@@ -89,10 +89,9 @@ bool setEdit(App::DocumentObject *obj, PartDesign::Body *body) {
         return false;
     App::DocumentObject *parent = nullptr;
     std::string subname;
-    auto activeBody = activeView->getActiveObject<PartDesign::Body*>(PDBODYKEY,&parent,&subname);
+    auto activeBody = activeView->getActiveObject<PartDesign::Body*>(PDBODYKEY);
     if (activeBody != body) {
         parent = obj;
-        subname.clear();
     }
     else {
         parent = getParent(obj, subname);


adrianinsaval:
this fixes a critical bug that is blocking release, @0penBrain if you have time could you review?

grubertechnic:
A test here works flawless.

```
OS: Debian GNU/Linux 11 (bullseye) (X-Cinnamon/lightdm-xsession)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33487 (Git)
Build type: Release
Branch: PD9792
Hash: ab243719252bd6af6fdbeec15fa82bb94415a601
Python 3.9.2, Qt 5.15.2, Coin 4.0.0, Vtk 9.0.1, OCC 7.7.1
Installed mods: 
  * QuickMeasure 2022.10.28
  * Assembly4 0.12.6
  * CurvedShapes 1.0.4
  * Curves 0.6.8
```

0penBrain:
Tested but I can't see any difference between current and PR code. :confused: 
Actually I was not able to get any problem with current code, trying to understand what this PR changes.
Could one help by providing a way to reproduce the current issue that is solved here ?

adrianinsaval:
step by step instructions are at #9792

0penBrain:
> step by step instructions are at #9792

OK, I tested.

And yes, the PR fixes the error message (hence probably reach its target).

But well, fundamentally I don't think #9538 creates this bug, but actually revealed a deeper underlying one.

Look at this.

 * New document
 * Create a Part container -> OK, container is active
 * Create a Body -> Hmmm, it's created inside the Part but in a very weird state. It is displayed bold, but do not have the "active" background color :confused:

Still more strange
 * New document
 * Create a Part container
 * Create a Body
 * Double click the Body (take care to not deselect it) -> It goes to **inactive** state

compared to
 * New document
 * Create a Part container
 * Create a Body
 * Deselect the Body by clicking empty area of the tree then double-click on it -> It goes to **active** state

At this time I really feel like when a Body is created when a Part container is active, it is created in a very weird state.
I think this is what we have to solve now. Feeling like this PR is patching the consequence of the bug, not the bug itself.

adrianinsaval:
No I think there are two bugs at play here IMO the one fixed here is also a bug because it happens when the body is in the correct state of being active.

You mention a bug where bodies are created in a weird state and I agree but I don't think that's the root of the problem for the reported bug, but still should be fixed.

adrianinsaval:
https://forum.freecad.org/viewtopic.php?p=691905#p691905
werner says:
> What happens is that

> ```
> auto activeBody = activeView->getActiveObject<PartDesign::Body*>(PDBODYKEY,&parent,&subname);
> ```
> unexpectedly sets subname to "Body". Afterwards when calling getParent() then the string "Body.Sketch." is prepended and "." appended which results into the incorrect subname "Body.Sketch001.Body.".
> 
> Fixing the issues is easy: before calling getParent() the string subname must be cleared.

0penBrain:
> https://forum.freecad.org/viewtopic.php?p=691905#p691905 werner says:

Fully agree with him. I pushed to your branch a commit that implements this. This looks IMO simpler and cleaner.
Feel free to squash if OK.

What was the reason for you to implement something else ?

adrianinsaval:
These functions are very confusing to me, why is it unexpected that `subname` is set to Body, that seems like the logical outcome of line 92 to me, if we don't want this to modify `subname` why are we passing it as argument at all? Following that same line of thought, if we are going to get `parent` independently anyways why are we passing it as an argument to `getActiveObject` anyway? Perhaps in that line we should just let it use their default nullptr value and just always get parent and subname later.

The behavior of `getActiveObject` also seems strange to me, when you initially create the Part and the Body, parent=Body and subname is empty, but if you then double click on the Body parent=Part and subname=Body, this later behavior seems the most correct to me, but it's not handled correctly by setEdit. Maybe when you first create the Body something goes wrong and it doesn't properly register that it has a Part as parent, the fact the Body is not highlighted makes it suspicious too.

And why does a function called `getParent` modify `subname`?

src/Mod/PartDesign/Gui/Utils.cpp
```diff
@@ -90,9 +90,9 @@ bool setEdit(App::DocumentObject *obj, PartDesign::Body *body) {
     App::DocumentObject *parent = nullptr;
     std::string subname;
     auto activeBody = activeView->getActiveObject<PartDesign::Body*>(PDBODYKEY,&parent,&subname);
+    subname.clear();
     if (activeBody != body) {
```
adrianinsaval, line 94: I'm also not understanding in which situation this statement is true. Anybody knows?

adrianinsaval:
This seemed to work and makes a little more sense to me, what do you think?

src/Mod/PartDesign/Gui/Utils.cpp
0penBrain: I don't know too.

0penBrain:
> These functions are very confusing to me, why is it unexpected that `subname` is set to Body, that seems like the logical outcome of line 92 to me, if we don't want this to modify `subname` why are we passing it as argument at all? Following that same line of thought, if we are going to get `parent` independently anyways why are we passing it as an argument to `getActiveObject` anyway? Perhaps in that line we should just let it use their default nullptr value and just always get parent and subname later.

Agree. If codebase is correct, it should be possible (and better) to just call `auto activeBody = activeView->getActiveObject<PartDesign::Body*>(PDBODYKEY)` Would you mind testing ?

> The behavior of `getActiveObject` also seems strange to me, when you initially create the Part and the Body, parent=Body and subname is empty, but if you then double click on the Body parent=Part and subname=Body, this later behavior seems the most correct to me, but it's not handled correctly by setEdit. Maybe when you first create the Body something goes wrong and it doesn't properly register that it has a Part as parent, the fact the Body is not highlighted makes it suspicious too.
> 
> And why does a function called `getParent` modify `subname`? Given that this function is actually setting two variables, won't it be better to have it be something like:
> 
> ```c++
> getParentAndSubname(obj, &parent, &subname);
> ```
> 
> to be more explicit about what it does

Yes, not everything is ideally implemented, but here IMO it's too much for this PR to go fixing all this. :)

adrianinsaval:
> Agree. If codebase is correct, it should be possible (and better) to just call `auto activeBody = activeView->getActiveObject<PartDesign::Body*>(PDBODYKEY)` Would you mind testing ?

I did! It's working. The PR is that now.

> Yes, not everything is ideally implemented, but here IMO it's too much for this PR to go fixing all this. :)

yep, PR remains limited to the fix

PR #9865, opened 2023-07-03
[Core] Create EditableDatumLabel, a datumLabel that can be edited with a spinbox

PaddleStroke:
It actually moves the spinbox gestion outside of TaskImage such that it can be reused in other places, notably in sketcher.

This will be a prerequisit of 'tool setting' branch, to draw the parameters on the view directly rather than in the taskbox.

It does not change any functionality.

Dear reviewer, I draw your attention to the ~EditableDatumLabel dtor to make sure everything is handled/deleted properly as I'm not 100% sure.

> Thank you for creating a pull request to contribute to FreeC

> AD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.
> 
> - [ ]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR
> 
> Please remember to update the Wiki with the features added or changed once this PR is merged.  
> **Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).
> 
> ---
>

### DIFF
--- a/src/Gui/SoDatumLabel.cpp
+++ b/src/Gui/SoDatumLabel.cpp
@@ -97,6 +97,7 @@ SoDatumLabel::SoDatumLabel()
     this->imgWidth = 0;
     this->imgHeight = 0;
     this->glimagevalid = false;
+    this->textOffset = SbVec3f(0.f, 0.f, 0.f);
 }
 
 void SoDatumLabel::drawImage()
@@ -485,7 +486,7 @@ void SoDatumLabel::generateDistancePrimitives(SoAction * action, const SbVec3f& 
     img3 = SbVec3f((img3[0] * c) - (img3[1] * s), (img3[0] * s) + (img3[1] * c), 0.f);
     img4 = SbVec3f((img4[0] * c) - (img4[1] * s), (img4[0] * s) + (img4[1] * c), 0.f);
 
-    SbVec3f textOffset = midpos + normal * length + dir * length2;
+    textOffset = midpos + normal * length + dir * length2;
 
     img1 += textOffset;
     img2 += textOffset;
@@ -539,7 +540,7 @@ void SoDatumLabel::generateDiameterPrimitives(SoAction * action, const SbVec3f& 
     img3 = SbVec3f((img3[0] * c) - (img3[1] * s), (img3[0] * s) + (img3[1] * c), 0.f);
     img4 = SbVec3f((img4[0] * c) - (img4[1] * s), (img4[0] * s) + (img4[1] * c), 0.f);
 
-    SbVec3f textOffset = pos;
+    textOffset = pos;
 
     img1 += textOffset;
     img2 += textOffset;
@@ -585,7 +586,7 @@ void SoDatumLabel::generateAnglePrimitives(SoAction * action, const SbVec3f& p0)
     // p0 - vector for angle intersect
     SbVec3f v0(cos(startangle+range/2),sin(startangle+range/2),0);
 
-    SbVec3f textOffset = p0 + v0 * r;
+    textOffset = p0 + v0 * r;
 
     SbVec3f img1 = SbVec3f(-this->imgWidth / 2, -this->imgHeight / 2, 0.f);
     SbVec3f img2 = SbVec3f(-this->imgWidth / 2,  this->imgHeight / 2, 0.f);
@@ -820,8 +821,6 @@ void SoDatumLabel::GLRender(SoGLRenderAction * action)
 
     // Position for Datum Text Label
     float angle = 0;
-
-    SbVec3f textOffset;
 
     // Get the colour
     const SbColor& t = textColor.getValue();
--- a/src/Gui/CMakeLists.txt
+++ b/src/Gui/CMakeLists.txt
@@ -1152,6 +1152,7 @@ SET(FreeCADGui_CPP_SRCS
     DocumentPyImp.cpp
     DocumentObserver.cpp
     DocumentObserverPython.cpp
+    EditableDatumLabel.cpp
     ExpressionBinding.cpp
     ExpressionBindingPy.cpp
     GraphicsViewZoom.cpp
@@ -1183,6 +1184,7 @@ SET(FreeCADGui_SRCS
     DocumentModel.h
     DocumentObserver.h
     DocumentObserverPython.h
+    EditableDatumLabel.h
     ExpressionBinding.h
     ExpressionBindingPy.h
     ExpressionCompleter.h
--- a/src/Gui/EditableDatumLabel.cpp
+++ b/src/Gui/EditableDatumLabel.cpp
@@ -1,0 +1,301 @@
+/// SPDX-License-Identifier: LGPL-2.1-or-later
+/****************************************************************************
+ *                                                                          *
+ *   Copyright (c) 2023 Ondsel <development@ondsel.com>                     *
+ *                                                                          *
+ *   This file is part of FreeCAD.                                          *
+ *                                                                          *
+ *   FreeCAD is free software: you can redistribute it and/or modify it     *
+ *   under the terms of the GNU Lesser General Public License as            *
+ *   published by the Free Software Foundation, either version 2.1 of the   *
+ *   License, or (at your option) any later version.                        *
+ *                                                                          *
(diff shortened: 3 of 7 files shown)


luzpaz:
Just wondering, is there a POC screencast of this ?

PaddleStroke:
@luzpaz this PR doesn't change anything on the user side. It only moves code from taskimage to datumLabel so that it can be reused later.

The behavior that is concerned is the current 'interactive scale' tool of the 'task Image'. The taskbox that opens when you double click on imageplane objects.

PaddleStroke:
Switched to draft as this can wait and I'm putting more changes in as the new tool settings branch progress.

abdullahtahiriyo:
I am not knowledgeable about this (and best if more expert eyes review). However, because you have pointed me to this PR to try to make another PR compile, I will just comment.

The ultimate objective appears to be to create a new class EditableDatumLabel (which perhaps should be SoEditableDatumLabel to note that a Coin node is intended). 

This is created in SoDatumLabel.h/SoDatumLabel.cpp, for which a substantial amount of dependencies, which are not necessary for SoDatumLabel are added. In view of this, perhaps it would have made sense to create SoEditableDatumLabel.h/cpp instead.

SoDatumLabel class is modified by adding an offset as member function. This changes to SoDatumLabel should be a separate commit. In the commit message should have been explained why this change is introduced. At a first glance, I have no objection to introducing it. However, it would have been helpful for the reviewer to have that information. It would have been easier for the reviewer if different concerns are separated in different commits.

I am not familiar with TaskImage TaskView, it appears that TaskImage TaskView could be rewritten, reusing code, using the new class (So)EditableDatumLabel. If this is the case it is really fine. But perhaps it would have been easier for the reviewer if a second commit introduced the code necessary for SoEditableDatumLabel (in SoEditableDatumLabel.h/cpp files), using the changed SoDatumLabel of the first commit, without touching TaskImage TaskView at all. Then, a third commit would have tackled the code reuse (rewritting TaskImage TaskView in terms of SoEditableDatumLabel).

PaddleStroke:
Thanks for the review.
The image taskbox (and especially the interactive scale tool with the spinbox) was rewriten by myself few months ago and reviewed by @wwmayer.

Yes I could have split the commits indeed, but as I wrote it I moved the code out of taskImage bit by bit so it was hard to separate the commits as so. Though I guess we can still soft reset and separate as it affects different files it should be easy now.

For the files I felt it was best to keep them close together that's why I kept them in the same files. But it's a weak reason. So if necessary we can split in 2 files.

PaddleStroke:
About the textOffset, noted that it would be best in a separate commit. 
About the why, it is actually the center position of the label text. So this is the point where we want to spinbox to appear. So this is why this needs to be accessible.

By the way there is a small issue here, it is that this textOffset is calculated after the image is rendered. So when the editableDatumLabel position the spinbox, it is using the textOffset from previous position. Which is visible if the mouse moves fast.

src/Gui/SoDatumLabel.cpp
```diff
@@ -1030,3 +1036,265 @@ void SoDatumLabel::setPoints(SbVec3f p1, SbVec3f p2)
...
+    root->addChild(label);
+
+    setPlacement(plc);
+}
+
+EditableDatumLabel::~EditableDatumLabel()
+{
+    deactivate();
+    transform->unref();
+    root->unref();
```
wwmayer, line 1084: When adding `transform->unref()` you have to do `transform->ref();` in the constructor as otherwise the application will segfault.

PaddleStroke:
@wwmayer thanks for reviewing. If you can please let me know what you think about Abdullah's suggestion that this is split into 2 files?
There are so many files in src/gui so perhaps it's best not too.

wwmayer:
> If you can please let me know what you think about Abdullah's suggestion that this is split into 2 files?

Abdullah is right to move EditableDatumLabel to separate source files. The class EditableDatumLabel inherits from QObject and thus shouldn't be inside the same file as SoDatumLabel which is an Inventor node class.

> There are so many files in src/gui so perhaps it's best not too. I don't know what's best here so let me know.

Sure, but as discussed elsewhere (in the forum?) we should create some sub-directories in src/Gui and move the files there.

While writing this message I have seen that you have renamed the class to SoDatumLabelEditable which IMO makes only little sense since it's NOT an Inventor class. The old class name was fine.

PaddleStroke:
Got it, then I'll revert the name change and split the files

PaddleStroke:
- Name is revert to EditableDatumLabel
- Files are split.
- Commit history is cleaned as per Abdullah suggestion.

wwmayer:
> Is this all?

I will test your PR tomorrow.

PaddleStroke:
@wwmayer Could you test it ?

wwmayer:
There are several bugs:
* Memory leak: the instance of EditableDatumLabel is never destroyed
* When I select two points and accept the value then width and height of the image is set to 0
* A compiler warning: the order of initializing class members in the constructor of EditableDatumLabel is different to the declaration order (-Wreorder-ctor).

PaddleStroke:
Can you please check if the deletion of EditableDatumLabel is done properly?
I also fixed the initialisation order.

> When I select two points and accept the value then width and height of the image is set to 0

I cannot reproduce this. The tool works correctly on my side, the spinboxes 'Width' and 'Height' are behaving correctly.

wwmayer:
> Can you please tell me more about how and when this happen?

* Click the Calibration button
* Select two points on the image
* Press Enter (or click on the Accept button)
* The image disappears because width and height are set to 0

PaddleStroke:
Hmm I can't replicate that here on windows. You don't have any error messages?

wwmayer:
No error message so far. I will try the PR later today to see where the values are set to 0.

wwmayer:
> the spinboxes 'Width' and 'Height' are behaving correctly after I enter a value

I never talked about the spin boxes of the dialog. I talked about the spin box after drawing the line in the 3d view. When I simply accept the calculated value without changing it the width and height of the image object become 0.

wwmayer:
Now that the memory leak is fixed more quirks show up.

When accepting/rejecting the calibration line then EditableDatumLabel::stopEdit() is called at some point. Internally this triggers a delayed deletion of the quantity spin box. Then after a while the destructor of EditableDatumLabel is called that again calls stopEdit(). But the quantity spin box is already deleted and the pointer is dangling.

The second call of stopEdit() leads to a segmentation fault. To fix this crash either rewrite the method:
```
void EditableDatumLabel::stopEdit()
{
    if (spinBox) {
        spinBox->deleteLater();
        spinBox = nullptr;
    }
}
```
or use a QPointer for the class member spinBox.

wwmayer:
> Hmm I can't replicate that here on windows.

I cannot imagine that this works correctly on Windows. This is what happens:
* EditableDatumLabel is a class member `value` that is initialized with 0
* When doing the procedure as described above then TaskImage::acceptScale() is called
* TaskImage::acceptScale() calls InteractiveScale::getScaleFactor()
* Because both points are different InteractiveScale::getScaleFactor() calls EditableDatumLabel::getValue()
* The class member value hasn't been modified and is still 0
* Because the pass scale factor is 0 TaskImage::scaleImage() will set widget and height to 0

IMO it's a flaw that this class member `value` even exists. The current scale factor can also be directly queried from the quantity spin box when needed.

PaddleStroke:
Sorry I miss understood the issue. I thought you entered a value. I haven't tried to actually accept without changing the value. I'll post a fix asap

PaddleStroke:
I'm not sure why I put this value variable in the first place actually. But yes it's wrong.
I removed it and applied your fix to the stopEdit. So both bugs are gone now.

wwmayer:
OK, the code looks good so far. But can you clean-up the commit history a bit, please?

* Leave 1st commit
* Incorporate 4th and 5th into 2nd commit
* Leave 3rd commit

PaddleStroke:
Thanks @wwmayer
It is squashed.

PR #9867, opened 2023-07-03
Avoid dagerous move without tool height compensation while it is supposed to be active

### LINKED ISSUE #9866: [Problem] Avoid dangeruos first G-code move while G49 is active

### Version

0.20 (Release)

### Full version info

```shell
[code]
OS: Debian GNU/Linux 12 (bookworm) (KDE/lightdm-xsession)
Word size of FreeCAD: 64-bit
Version: 0.20.2.
Build type: Release
Python 3.11.1, Qt 5.15.8, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Installed mods: 
  * freecad.gears 1.0.0
[/code]
```

### Subproject(s) affected?

Path

### Problem description

I created a FreeCAD model and generated some Path Jobs to mill it out,
and was very surprised when the generated G-code ended up digging the
first cutter straight into the work table.  This is using the linuxcnc
post processor, but is unrelated to the postprocessor used.

The model is available from
https://codeberg.org/pere/thinkpad-x230-dc-socket-bracket .  I
used the G-code in the subdirectory linuxcnc-jobs in  commit
ff248e59b7c44bda3e6320e6499ee9c240156dae for this failure.

I tracked this down to the first move in the generated in the G-code
file, which is a "G0 Z<safe height>", which is no longer safe when the
15 cm tool height compensation has been wiped out with the G49 in the
preamble.

This is the generated G-code:

```
(Exported by FreeCAD)
(Post Processor: linuxcnc_post)
(Output Time:2023-07-03 14:24:06.190303)
(begin preamble)
G17 G54 G40 G49 G80 G90
G21
(begin operation: G54)
(machine units: mm/min)
G54 
G0 Z19.000         ; <-------- This one is dangerous while G49 is active
(finish operation: G54)
(begin operation: 6mm-endmill001)
(machine units: mm/min)
(6mm-endmill001) 
M5
M6 T3 
G43 H3 
M3 S2500 
(finish operation: 6mm-endmill001)
(begin operation: MillFace)
(machine units: mm/min)
(MillFace) 
G0 Z19.000 
```

I see from https://forum.freecad.org/viewtopic.php?t=37839 the issue
has been known since 2019, and was even mentioned as an issue when the
introduction of G43 was discussed in
https://forum.freecad.org/viewtopic.php?t=27180 .

It is unclear to me why the G49 is used in the preamble to begin with.
Is it not better to put it just before the G43, and avoid any G0 move
before G43 has been used to load the correct tool height compensation?

I had used the automatic tool height measurement in the machine to set
the correct tool height compensation for the tool holder and bit in
question, and was very surprised to see that it dived straight down as
the first move.  Was not able to stop it quickly enough to avoid
damage.

### Anything else?

_No response_

### Code of Conduct

petterreinholdtsen:
Changed all postprocessor using G43 to set tool height compensation after tool change to not set G49 before doing the first move to "safe" height after G54, as this would take an unexpected dive if the already active tool height was significant.

Fixes #9866.

### DIFF
--- a/src/Mod/Path/Path/Post/scripts/fanuc_post.py
+++ b/src/Mod/Path/Path/Post/scripts/fanuc_post.py
@@ -118,7 +118,7 @@ PRECISION = 3
 tapSpeed = 0
 
 # Preamble text will appear at the beginning of the GCODE output file.
-PREAMBLE = """G17 G54 G40 G49 G80 G90
+PREAMBLE = """G17 G54 G40 G80 G90
 """
 
 # Postamble text will appear following the last operation.
@@ -611,8 +611,11 @@ def parse(pathobj):
 
                 # add height offset
                 if USE_TLO:
+                    outstring.append("G49")
                     tool_height = "\nG43 H" + str(int(c.Parameters["T"]))
                     outstring.append(tool_height)
+                else:
+                    outstring.append("G49")
 
             if command == "message":
                 if OUTPUT_COMMENTS is False:
--- a/src/Mod/Path/Path/Post/scripts/linuxcnc_post.py
+++ b/src/Mod/Path/Path/Post/scripts/linuxcnc_post.py
@@ -110,7 +110,7 @@ CORNER_MAX = {"x": 500, "y": 300, "z": 300}
 PRECISION = 3
 
 # Preamble text will appear at the beginning of the GCODE output file.
-PREAMBLE = """G17 G54 G40 G49 G80 G90
+PREAMBLE = """G17 G54 G40 G80 G90
 """
 
 # Postamble text will appear following the last operation.
@@ -432,8 +432,11 @@ def parse(pathobj):
 
                 # add height offset
                 if USE_TLO:
+                    outstring.append("G49")
                     tool_height = "\nG43 H" + str(int(c.Parameters["T"]))
                     outstring.append(tool_height)
+                else:
+                    outstring.append("G49")
 
             if command == "message":
                 if OUTPUT_COMMENTS is False:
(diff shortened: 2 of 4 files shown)


petterreinholdtsen:
An alternative is to get rid of the G0 Z<height> move mentioned in #9866, but it seem more sensible to just make sure all moves are done with tool compensation when it is active.

Note, the patch is untested.

petterreinholdtsen:
The patch is now tested and found to be working.  Still not happy about leaving the G0 after G54 in place, and would love some feedback on the approach and suggestions how to get rid of it.

petterreinholdtsen:
I believe I found a way to only drop the G0 after G54, and am working on an updated patch.  Will test it before updating it here.

luzpaz:
CC @mlampert (would you have a moment to weigh-in?)

petterreinholdtsen:
[luzpaz]
> CC @mlampert (would you have a moment to weigh-in?)

I am on IRC, #freecad, if someone want to enjoy a more direct chat
channel with me. :)

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
I've updated and tested the patch to only drop the dangerous G0 move instead of changing how the various postprocessors work.

sliptonic:
> I am on IRC, #freecad, if someone want to enjoy a more direct chat channel with me. :)

I'm not on IRC anymore but available on Telegram, Element, Discord, Slack, Gitter, etc.

Seems most of the Path chat is going on in the Gitter/Element  room here:
https://matrix.to/#/#FreeCAD_Path:gitter.im

PR #9876, opened 2023-07-05
Aded AppImage build script

yorikvanhavre:
An all-in-one script meant to produce an AppImage using conda. Roughly the same as https://github.com/FreeCAD/FreeCAD-Bundle/blob/master/conda/linux/create_bundle.sh but it has been reworked to be portable, and I think it's good to have one in the FreeCAD source itself.

Later on we could try to make it build FreeCAD from the sources inside the conda environment...

---

### DIFF
--- a/tools/build/AppImage/.gitignore
+++ b/tools/build/AppImage/.gitignore
@@ -1,0 +1,4 @@
+AppDir
+*.txt
+*.AppImage
+qt.conf
--- a/tools/build/AppImage/build-with-conda.sh
+++ b/tools/build/AppImage/build-with-conda.sh
@@ -1,0 +1,207 @@
+#!/bin/sh
+
+# ***************************************************************************
+# *   Copyright (c) 2023 Yorik van Havre <yorik@uncreated.net>              *
+# *                      and FreeCAD maintainers (looooo, adrianinsaval )   *
+# *                                                                         *
+# *   This program is free software; you can redistribute it and/or modify  *
+# *   it under the terms of the GNU Lesser General Public License (LGPL)    *
+# *   as published by the Free Software Foundation; either version 2 of     *
+# *   the License, or (at your option) any later version.                   *
+# *   for detail see the LICENCE text file.                                 *
+# *                                                                         *
+# *   This program is distributed in the hope that it will be useful,       *
+# *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+# *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+# *   GNU Library General Public License for more details.                  *
+# *                                                                         *
+# *   You should have received a copy of the GNU Library General Public     *
+# *   License along with this program; if not, write to the Free Software   *
+# *   Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  *
+# *   USA                                                                   *
+# *                                                                         *
+# ***************************************************************************
+
+# adapted from https://github.com/FreeCAD/FreeCAD-Bundle/blob/master/conda/linux/create_bundle.sh
+
+# This script produces a FreeCAD AppImage locally using conda repositories. It is a all-in-one
+# script that can be run from anywhere and does not rely on the FreeCAD source code. It fetches
+# everything form conda repositories. Therefore, you need to make sure they contain the FreeCAD
+# package you want to build. An easy way to check is by searching for available FreeCAD packages:
+#
+#     conda search "freecad=0.21.0" -c adrianinsaval/label/dev -c freecad -c conda-forge
+#
+
+
+# To use this scriipt, you will need:
+#
+#     1) Mambaforge from https://github.com/conda-forge/miniforge#mambaforge
+#     if you choose to not install the conda environment when installing Mambaforge, 
+#     you will need to do it before running this script, by running the following
+#     in your terminal. This will make the "conda" and "mamba" commands available:
+#     eval "$(/home/yorik/Mambaforge/bin/conda shell.zsh hook)"
+#     (replace the Mambaforge path by yours and zsh by your shell name)
+#
+#     2) appimagetool from https://appimage.github.io/appimagetool/
+#     place it in your exec path so it can be found by this script
+#
+#     3) a gpg key to sign the package
+#
+#     4) verify or change the config values here below
+#
+#     5) If you are on Fedora, there are lines to uncomment below in the script
+#
+# When done, just run this script and you should get an appimage with the
+# corresponding sha hash in the current folder.
+#
+# To cleanup after build: Just delete the "AppDir" folder (not done automatically since
+# next builds can reuse the downloaded packages).
+
+# config
+
+# make sure you have a gpg key for this email
+gpg_key="yorik@freecad.org"
+# the FreeCAD version we're looking for
+target_version="0.21.0"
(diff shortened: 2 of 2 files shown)


tools/build/AppImage/build-with-conda.sh
```diff
@@ -24,10 +24,12 @@
 
 # adapted from https://github.com/FreeCAD/FreeCAD-Bundle/blob/master/conda/linux/create_bundle.sh
 
-# This script produces a FreeCAD AppImage locally using conda repositories. It is a all-in-one
-# script that can be run from anywhere and does not rely on the FreeCAD source code. It fetches
-# everything form conda repositories. Therefore, you need to make sure they contain the FreeCAD
-# package you want to build. An easy way to check is by searching for available FreeCAD packages:
+# This script produces a FreeCAD AppImage locally using conda repositories. It
+# is a all-in-one script that fetches FreeCAD and all dependencies from conda
```
luzpaz, line 28: is **an** all-in-one...

tools/build/AppImage/build-with-conda.sh
```diff
@@ -24,10 +24,12 @@
...
+# repo (it also extracts a couple of XDG files from the FreeCAD source). Everything
+# comes form conda repositories, therefore, you need to make sure they contain the
```
luzpaz, line 30: comes **from** conda...

adrianinsaval:
I don't think putting this kind of scripts on the main repo is a good idea, they tend to rot (see the docker stuff). Instead it would be a better idea to document the usage and tell people to go directly to the script in the bundle repo where it's much more likely to be kept up to date since it's used all the time and we are sure to hear about it if it starts having problems.

yorikvanhavre:
@luzpaz fixed, thanks!

@adrianinsaval I agree these things rot quickly, but I also think it's important to have all the build tools, or at least a minimal version of them, in the main repo. Maybe indeed the bulk needs to be kept elsewhere and only instructions be part of the main repo, not sure... The idea basically is to lower the entry barrier and allow more people to help working on the releases. Even if some stuff needs to be upgraded from time to time

Pesc0:
How would this work? you launch the script and it builds? 

@adrianinsaval how does this relate to the VSCode PR?

tools/build/AppImage/.gitignore
```diff
@@ -0,0 +1,4 @@
+AppDir
+*.txt
+*.AppImage
```
adrianinsaval, line 3: *.AppImage-SHA256.txt
```suggestion
*.AppImage-SHA256.txt
```

adrianinsaval:
> How would this work? you launch the script and it builds?
> 
> @adrianinsaval how does this relate to the VSCode PR?

No this creates an environment and then installs the freecad precompiled conda package on it and then makes an AppImage out of this environment. But you bring up a good idea, we could integrate this into the vscode PR and have a "Create AppImage" task.

I think it would be best to split this script into 2:

1. that creates a conda environment like @oursland described here: https://github.com/oursland/FreeCAD-Build-Notes [can be added as a "Create conda env" task in vscode). This script could have an optional argument to also install the precompiled freecad package if the intention is to make an appimage out of this
2. that creates an appimage (or potentially a dmg for mac) out of that environment

we then document that the steps to make an appimage is create an environment using `1` passing the corresponding flag if we want to install the precompiled package, then optionally build freecad in that environment and install it on the environment and lastly create an appimage out of this environment with `2`

tools/build/AppImage/build-with-conda.sh
```diff
@@ -0,0 +1,209 @@
...
+# When done, just run this script and you should get an appimage with the
+# corresponding sha hash in the current folder.
+#
+# To cleanup after build: Just delete the "AppDir" folder (not done automatically since
+# next builds can reuse the downloaded packages).
+
+# config
+
+# make sure you have a gpg key for this email
+gpg_key="yorik@freecad.org"
```
adrianinsaval, line 65: gpg signing should be optional and the key shouldn't be hardcoded, we could have this as an argument for the script or an env variable

tools/build/AppImage/build-with-conda.sh
```diff
@@ -0,0 +1,209 @@
...
+# the FreeCAD version we're looking for
+target_version="0.21.0"
+# make sure target_python matches the one FreeCAD is built with! Check with
+# conda search "freecad=0.21.0" -c adrianinsaval/label/dev -c freecad -c conda-forge
+target_python="3.10"
```
adrianinsaval, line 70: this could be made into optional arguments (that default to these values) so the user doesn't actually need to edit the script to use it.

tools/build/AppImage/build-with-conda.sh
```diff
@@ -0,0 +1,209 @@
...
+package_name="FreeCAD_${target_version}-${target_date}-conda-Linux-${arch}-py$(echo ${target_python} | sed 's/\.//g')"
+conda_env="AppDir/usr"
+
+# dependencies
+echo "\nCreating the environment"
+packages="freecad=${target_version} occt vtk python=${target_python} \
+blas=*=openblas numpy matplotlib-base scipy sympy pandas \
+six pyyaml pycollada lxml xlutils olefile requests blinker \
+opencv qt.py nine docutils calculix opencamlib ifcopenshell \
+appimage-updater-bridge"
```
adrianinsaval, line 127: I recommend that you update to the latest script in the FreeCAD-Bundle repo as it has been adapted to work on aarch64 architecture too, it skips some packages that aren't available on that platform yet.

tools/build/AppImage/build-with-conda.sh
adrianinsaval: the conda channel to be used could be a good thing to put as an argument for the script, this way people could create their own conda channel and generate appimages for their forks (e.g. realthunder, althought he seems to already have a good system in place)

oursland:
A general philosophical argument is to keep source, build scripts, and documentation in the same repository so that they may be in sync at all times.  When checking out a version of source, there's no inconsistency with that version and the necessary version in the build and documentation repos.  Consequently, a means of building the packages within the repository would be a benefit.

Duplicating external code, however increases developer overhead and the likelihood of a synchronization error.

I believe a reasonable approach to make cross-platform package building from within the source tree would be to:

1. bring over the `FreeCAD-Bundle` repository into the main repository
2. use CMake's [CPack External Generator](https://cmake.org/cmake/help/latest/cpack_gen/external.html#variable:CPACK_EXTERNAL_PACKAGE_SCRIPT) to call the platform-specific package generation script.

yorikvanhavre:
There are of course many, many ways to improve this. I expect @adrianinsaval's bundle work to remain the main framework used to build different FreeCAD release packages. This one, as you can see it was intentionally "reduced" into a single file, is much more aimed at being didactical than powerful. I think we should keep it as minimal as possible. It's more meant as a recipe than to be actually used.

If we ask users to go through another repo, with several separated scripts, the result is that very few people will take the time to do it. I myself needed several weeks to finally go dig into it 😅. As a consequence, very few know how to build an appimage, and the release process always falls on a couple of shoulders only.

This here is more an attempt to have the process documented and easy to try & learn. I don't think we should complexify it too much or we'll loose the main advantage.

I actually thought of making it a md file instead, but a script that you can actually run is way more interesting,

Maybe it needs more commenting, more than everything else?

yorikvanhavre:
I think we should also actually remove the .gitignore. Bad idea :)

PR #9878, opened 2023-07-06
Use python3 as the interpreter instead of python (ie v2) more places.

petterreinholdtsen:
This fixes lintian warning "unusual-interpreter".

Part of the Debian edition of FreeCAD since 2023.

- [X]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR.

### DIFF
--- a/src/Mod/Mesh/App/MeshTestsApp.py
+++ b/src/Mod/Mesh/App/MeshTestsApp.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
 #  Copyright (c) 2007 Jürgen Riegel <juergen.riegel@web.de>
--- a/src/Mod/Robot/KukaExporter.py
+++ b/src/Mod/Robot/KukaExporter.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
 # Kuka export lib (c) Jürgen Riegel 2009 LGPL 2.1 or higher
--- a/src/Mod/Robot/MovieTool.py
+++ b/src/Mod/Robot/MovieTool.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 import os,sys,string
 import FreeCAD,FreeCADGui,Robot,RobotGui
 


benj5378:
Is it even relevant having these lines, maybe they should be completely deleted? I think they are only relevant if the Python scripts are run through the Bash terminal.

petterreinholdtsen:
[Benjamin Bræstrup Sayoc]
> Is it even relevant having these lines, maybe they should be
> completely deleted? I think they are only relevant if the Python
> scripts are run through the Bash terminal.

I have no idea, but assumed they were there for a reason and should be
correct.  I assume you as the upstream know best there.

-- 
Happy hacking
Petter Reinholdtsen

chennes:
I think we can leave the lines in -- sometimes they provide value to people trying to test things without actually running FreeCAD, for example.

benj5378:
@petterreinholdtsen pretty sure they can be completely deleted: https://realpython.com/python-shebang/#:~:text=Therefore%2C%20use%20the%20shebang%20when,script%20with%20python%20or%20python3%20.&text=The%20highlighted%20line%20used%20to,as%20Python%20defaulted%20to%20ASCII

Sounds like it for running scripts in bash without writing the executer in front (so '# script.py' instead of '# python3 script.py'. And that isn't relevant here IMO.

benj5378:
> I think we can leave the lines in -- sometimes they provide value to people trying to test things without actually running FreeCAD, for example.

?? Which value? Lines basically tell it's a Python script, pretty obvious in my opinion.

petterreinholdtsen:
[Benjamin Bræstrup Sayoc]
> ?? Which value? Lines basically tell it's a Python script, pretty
> obvious in my opinion.

Given that the execute bit is missing, I suspect this value is
miniscule.

But it might help editors to correctly guess the program language, if
they are unable to figure it out from the .py ending.  Not sure if it is
the case for any relevant editor in use.

I do not care much if they are removed or kept, but if they are kept, I
would love for the to accurately point to python 3. :)
-- 
Happy hacking
Petter Reinholdtsen

benj5378:
@petterreinholdtsen would you like to do any of these? https://forum.freecad.org/viewtopic.php?t=79067

petterreinholdtsen:
[Benjamin Bræstrup Sayoc]
> @petterreinholdtsen would you like to do any of these?
> https://forum.freecad.org/viewtopic.php?t=79067

Nah, I just got involved with these patches because I updated the Debian
package and got annoyed with patches in the package not being already
sent upstream.  In FreeCAD I am more interested in the Path workbench to
be able to generate G-code for LinuxCNC than the modelling GUI.  I hang
out on the #freecad IRC channel, if someone want to chat.

-- 
Happy hacking
Petter Reinholdtsen

benj5378:
@petterreinholdtsen Can recommend using Discord instead as I will then be allowed to tag you.

PR #9895, opened 2023-07-08
fix wrong filesystem encoding in GUI with Python 3.11 on Linux

bdieterm:

---

The added initialization of the locales will prevent the wrong ascii-only filesystem encoding.

Further description of the problem, tracking down the cause and the solution can be seen at my topic in the forum:
https://forum.freecad.org/viewtopic.php?t=79192

I don't see a reason for the added line to break something. The same line I added has been in the main function of src/Main/MainCmd.cpp for more than 9 years.

### DIFF
--- a/src/Main/MainGui.cpp
+++ b/src/Main/MainGui.cpp
@@ -97,6 +97,7 @@ private:
 
 int main( int argc, char ** argv )
 {
+    setlocale(LC_ALL, ""); // use native environment settings
 #if defined (FC_OS_LINUX) || defined(FC_OS_BSD)
     // Make sure to setup the Qt locale system before setting LANG and LC_ALL to C.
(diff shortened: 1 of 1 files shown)


adrianinsaval:
@0penBrain you did the numbers locale stuff that's there IIRC, could you check this PR when you have time?

0penBrain:
What I did was essentially dealing with Qt things. :)
This said, I don't see why the added line may harm, but I may miss something.
Would be good to have some tests on various Win and MacOS OSes.

yorikvanhavre:
Looking at this with @chennes... @wwmayer could you have a look at this one?

adrianinsaval:
Probably a good idea to backport this to 0.21?

chennes:
I think so: can you either do that, or submit a backport request issue?

PR #9896, opened 2023-07-08
set toolbar icon size when starting FreeCAD GUI from Python

bdieterm:

---

When starting the FreeCAD GUI from a Python Interpreter (via `FreeCADGui.showMainWindow()`), some steps are missing in the function `setupMainWindow` compared to a normal start of FreeCAD. One of these missing steps is that the toolbar icon size is not set according to the user settings.
The added lines are copied from file src/Gui/Application.cpp and only slightly modified.

Ideas for further modifications are described in my topic in the FreeCAD forum:
https://forum.freecad.org/viewtopic.php?t=79283

### DIFF
--- a/src/Main/FreeCADGuiPy.cpp
+++ b/src/Main/FreeCADGuiPy.cpp
@@ -305,6 +305,12 @@ QWidget* setupMainWindow()
             mw->setWindowTitle(QString::fromLatin1(App::Application::Config()["ExeName"].c_str()));
         }
 
+        // set toolbar icon size
+        ParameterGrp::handle hGrp = Gui::WindowParameter::getDefaultParameter()->GetGroup("General");
+        int size = hGrp->GetInt("ToolbarIconSize", 0);
+        if (size >= 16) // must not be lower than this
+            mw->setIconSize(QSize(size,size));
+
         if (!SoDB::isInitialized()) {
             // init the Inventor subsystem
             SoDB::init();


src/Main/FreeCADGuiPy.cpp
```diff
@@ -305,6 +305,12 @@ QWidget* setupMainWindow()
             mw->setWindowTitle(QString::fromLatin1(App::Application::Config()["ExeName"].c_str()));
         }
 
+        // set toolbar icon size
+        ParameterGrp::handle hGrp = Gui::WindowParameter::getDefaultParameter()->GetGroup("General");
+        int size = hGrp->GetInt("ToolbarIconSize", 0);
+        if (size >= 16) // must not be lower than this
```
benj5378, line 311: And what happens if not?

bdieterm: Hi @benj5378,
I guess you are referring to the comment in the last line: "must not be lower than this".
Good question -- I also asked this myself. As I wrote in my third sentence, the code is just copied from another place in this repo. So let's have a look at this original code with exactly this comment.
Traversing the blame layers in this FreeCAD repository does not bring us back to the author because the history stops suddenly. But there is an older archive, and there we find it:
`https://github.com/berndhahnebach/All_FreeCAD/commit/8077736ad6eb4b17d034ac9474ea58bc68bcadb1`
So if you want to know the answer, you can ask wwmayer who authored this in 2011.
(I did not make the url a link and wwmayer a mention to avoid unnecessary messages.)

wwmayer: > And what happens if not?

If not then the icon size of the main window won't be overridden but the default value (24x24 px) will be kept.

chennes:
We left this tagged with the 0.21 milestone in Monday's review because it seems like a small adjustment that has minimal possibility for breaking things and essentially corrects an oversight in earlier code. If there are any objections to this being merged for 0.21 please post here, otherwise my inclination is to merge it.

PR #9910, opened 2023-07-11
[Stylesheets] Sliders and some cleanup.

MisterMakerNL:
I gave the sliders some love and found some missing checkboxes, I also added a border to some checkbox svg files Checkbox_undetermined had not border while the rest did. So basically a clean-up.

And made the check_light....actually light and not the exact same black as the check_dark.
afbeelding
afbeelding
afbeelding
afbeelding

PR can be tested by installing this addon:
https://github.com/MisterMakerNL/Freecad-Built-in-themes-beta

---


benj5378:
Without even reading the code, there are multiple formatting issues here. The svg files should really be plain svg without inkscape stuff : this conversion should be done by it's own commit.

There's a 6000 lines diff, can't see it as the diff is too big, but guess it's some kind of formatting. This type and this size of formatting is not allowed in this PR.

MisterMakerNL:
Nice friendly feedback....
There was a tab issue in 2 or 3 files, no clue how it got in. I removed the extra tabs.
Every stylesheet has the same edits so it is easy to check.
I do see that it again has ticked Cmakesettings, which is super annoying that I have too think about that every-time. So I ahve to fix that.
About the svg, I have no clue what you mean with inkscape and conversions care to explain?

src/Gui/Stylesheets/images_dark-light/checkbox_light.svg
```diff
@@ -2,66 +2,41 @@
...
+   viewBox="0 0 30 29.999999"
+   sodipodi:docname="checkbox_dark.svg"
+   inkscape:version="1.2-beta (1b65182ce9, 2022-04-05)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:dc="http://purl.org/dc/elements/1.1/">
```
luzpaz, line 18: The SVGs need to be saved without inkscape header info, is what I think @benj5378 is referring to. 
In inkscape it's something like: 
Open the save dialog [File > Save As...] and in the drop down list choose Optimised SVG

benj5378:
@MisterMakerNL 

> Nice friendly feedback...

Sorry, but I believe it is stated somewhere, that formatting is not allowed as part of PR (though now I can't find it myself). That doesn't mean I don't like you contributing to the project - I do. The reason that we don't want formatting changes for multiple reasons, both contribution stats and it creates a lot of noise in git-blame for instance. Non the less, if it's done, it should be a commit by itself for keeping track of actual changes and formatting noise. I don't know if someone has another opinion than me?

>> The svg files should really be plain svg without inkscape stuff : this conversion should be done by it's own commit.
>>
> Mean with inkscape and conversions care to explain

All attributes in SVG starting with `inkscape:` is not really SVG but just metadata convenient for Inkscape usage/editing, for instance `inkscape:pagecheckerboard="0"`. Therefore, we don't really want it to be included into our files. `inkscape:zoom="14.116667"` this for instance is data on your last zoom in Inkscape level I believe.

It's really not your fault as it were there to begin with, but it would be good to remove it. This is what I call "conversion", convert Inkscape SVG to Plain SVG. This can be done in Inkscape by using "Save As..." and then choosing Plain SVG as the filetype.

:) let me know if you have any more questions

src/Gui/Stylesheets/images_dark-light/checkbox_light.svg
benj5378: Plain SVG, not optimised XD else it will be all collapsed and stuff I believe. We usually save as Plain SVG in the repo as far as I know.

grubertechnic:
> Nice friendly feedback....

I assume, that is sarcasm. If it is, i do not see any unfriendliness, it is just notes.

Also a note from me, please change the the tile. Give a meaningful tile like

_[Stylesheets] Lorem ipsum_

See https://wiki.freecad.org/Source_code_management#Writing_good_commit_messages. Else when search commits or while any other git and code management operations, the handling is bad.

MisterMakerNL:
I fixed some svg files but I feel like this work was just useless busy work. I saved 300kb.... Also every single SVG file I opened in the folder has inkscape stuff.
So yeah, probably a bunch of time wasted.

benj5378:
> I fixed some svg files but I feel like this work was just useless busy work. I saved 300kb.... Also every single SVG file I opened in the folder has inkscape stuff. So yeah, probably a bunch of time wasted.

If you want to look at it that way, okay then. But for most regular developer, it's important as to avoid lots of noise throughout the repository. Writing code is easy. Maintaining code is difficult. Therefore we want to do ourselves that favor to keep things tidy.

I really only expected you to clean up the 3-4 files you had already begun editing. But nice that you did it with all of the files.

benj5378:
@MisterMakerNL 
> a bunch of time wasted.

If it was that bad, then for another time:
```
import os

for file in os.listdir("./"):
    if file.endswith(".svg"):
        os.system(f"./Inkscape-b0a8486-x86_64.AppImage --export-plain-svg ./{file} --export-filename=./{file}")
```

chennes:
> Sorry, but I believe it is stated somewhere, that formatting is not allowed as part of PR (though now I can't find it myself).

You are probably thinking of this section of the [Handbook](https://freecad.github.io/DevelopersHandbook), in the [Maintainer's Guide](https://freecad.github.io/DevelopersHandbook/maintainersguide/):
> as a best practice developers should work to eliminate any code formatting changes except in code they are actually changing (that is, they should not apply automatic code formatting across an entire file where they are only changing a few lines). Maintainers should err on the side of forgiveness in the event of inadvertent code format changes: if it's straightforward to remove the superfluous changes then it should be done, but a PR should not be rejected for simple formatting issues.

So that is to say, it would have been best if the tab issue was in a separate commit (because then we could add it to .git-blame-ignore-revs), but it's not a show-stopper here. It is nice to have the extra Inkscape stuff removed from the SVG files, so I do appreciate that as well. @MisterMakerNL if you are comfortable using git, I suggest squashing this into three commits: the original one, the one that implements the transparency fix, and the one that fixes the SVGs. If not, no big deal. We'll review this on Monday and decide if we can pull it in before 0.21.

MisterMakerNL:
I have absolutly no idea on how to split this up. I'm on windows and use github desktop client and vscode.

PaddleStroke:
This commit does not change the cmakefile which still looks like : 

> SET(PreferencePacks_Directories
> "Classic theme"
> "Dark theme"
> "Light theme"
> )

Though it works so I'm not sure what to think of it?

PaddleStroke:
Ah no it doesn't seem to appear. At least not anymore in my branch. Well whatever I'm on it.

PR #9937, opened 2023-07-15
[Everywhere] Remove Inkscape metadata

benj5378:

---


benj5378:
Well, now I am the one creating lots of noise.

Do you have any opinions of the sanitations @chennes? This is far from all files... some files are exported as optimized SVG others as plain SVG, we should probably have a guideline in the handbook

yorikvanhavre:
Thinking out loud here, honestly, I am wondering if this is a good move, in theory. 

This is source code, right? The files here are meant to be easily editable by others. By suppressing inkscape properties, we turn these files less easily editable. This is I guess not really a problem for our icons, but in more general meaning, is that a good move?

This inkscape stuff is not really heavy, and it's transparent to other apps too. So it's not really an issue that we're having and solving here. Plus, each time someone will update icons, there is a good chance the inkscape stuff will be in again. Will we each time go there and tell people to change their PR? Again, why would we do so if it does not really cause a problem...

If it is a file size problem, I'm not sure this really makes a big difference, but shouldn't it then be some script that runs at build time?

chennes:
Wasn't there some kind of issue with some part of FreeCAD not liking the Inkscape stuff? That may be in the distant past now, I don't remember exactly what it was.

yorikvanhavre:
Indeed we should turn this into some kind of official policy to avoid unnecessary work like this (sorry @benj5378 !) Not sure where, though...

PR #9945, opened 2023-07-17
App/Toponaming: Tweak ComplexGeoData destructor

chennes:
MSVC seems to prefer this destructor: in debug mode on Windows I get a segmentation fault on the destruction of a shared_ptr with the original code, and this eliminates the problem. @wwmayer do you have any insight?

### DIFF
--- a/src/App/ComplexGeoData.cpp
+++ b/src/App/ComplexGeoData.cpp
@@ -60,8 +60,6 @@ ComplexGeoData::ComplexGeoData()
 {
 }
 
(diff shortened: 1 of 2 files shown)


wwmayer:
> do you have any insight?

It very likely has something to do with the fact that multiple inheritance is used in this case but I haven't found an explanation about the behaviour of declaring the destructor as "override" or "virtual" in such a case. But good possible that it's simply a compiler bug in MSVC

PR #9950, opened 2023-07-18
Add FREEECAD_USE_CCACHE cmake option

adrianinsaval:
Defaults to ON except for conda compilation on windows, automatic ccache detection has caused me problems in conda compilations using github actions, therefore I add an option to disable it.

---


adrianinsaval:
merging since it's harmless and helps with packaging

adrianinsaval:
Ok, this was not harmless and I should not have merged this myself. CI now has no caching and takes almost 2h every time, the if statement happens before the option is set so it's always false. I will try to fix this asap but could use some help.

adrianinsaval:
@wwmayer do you know if it's required to find ccache so early in cmake or can I move this to `cMake/FreeCAD_Helpers/InitializeFreeCADBuildOptions.cmake`?

wwmayer:
> do you know if it's required to find ccache so early in cmake

Yes, it is: https://crascit.com/2016/04/09/using-ccache-with-cmake/

PR #9954, opened 2023-07-18
[Core] Warn user if using deprecated backup extension

AjinkyaDahale:
A minimal version of #9668. Only warns the user to manually change the backup system to `FCBak`.

#9668 fails only in the github CI (outside of tests), and only with the commit not included here.

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -2518,6 +2518,17 @@ void Application::setStyleSheet(const QString& qssFile, bool tiledBackground)
     }
 }
 
+void Application::checkForDeprecatedSettings()
+{
+    // From 0.21, `FCBak` will be the intended default backup format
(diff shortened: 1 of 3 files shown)


chennes:
Are we trying to merge this for 0.21?

AjinkyaDahale:
> Are we trying to merge this for 0.21?

@chennes yes. Github CI doesn't seem to like changing to fcbak, but local machines (mine and @0penBrain's) do not seem to be able to reproduce. So I propose we just add this message to the users to let them know the setting will be deprecated.

chennes:
@0penBrain you want to take this one?

0penBrain:
@chennes I lack time currently but I'm trying to find why the original PR fails on CI. Will take some more time.

chennes:
OK, I will merge this one as-is then so we at least have some warning, while work continues on the other.

PR #9909, opened 2023-07-11
Gui: Add rotation center indicator

Rexbas:
- [X]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---
Adds an option to show the rotation center while dragging. I personally like to see the center of rotation while dragging, maybe other people like it too. Besides, it could be useful for finding problems regarding the rotation center.

https://github.com/FreeCAD/FreeCAD/assets/14298143/ab540721-3a36-41bd-a7a1-252a301b8b00

PR #9969, opened 2023-07-22
Gui: Improve object center rotation mode

### LINKED ISSUE #9911: [Problem] Axis cross affects the object center rotation mode

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: KDE neon 5.27 (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33442 +3 (Git)
Build type: Release
Branch: rotationcenter
Hash: 1ae6a68ee0842d4e35b609556d5105bed3a6125f
Python 3.11.3, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.7.1
[/code]


The about page did not give me the right branch name and +n in version
```

### Subproject(s) affected?

None

### Problem description

Axis cross affects the "object center" rotation mode. Steps to reproduce (needs #9909 for visualization):
1. Ensure the axis cross is disabled.
2. Make a simple cube.
3. Set rotation mode to "Object center".
4. Start dragging. With the axis cross disabled, the center should be at the center of the object. This is expected.
5. Enable the axis cross.
6. Start dragging again. Now the rotation center is not shown at the center of the object but depending on the zoom level more towards the center of the axis cross. I would expect that the axis cross does not have an effect on the object center but apparently it does.

https://github.com/FreeCAD/FreeCAD/assets/14298143/8bad2e17-f622-487f-9f72-954dded24b01

### Anything else?

_No response_

### Code of Conduct

Rexbas:
---
This is my attempt to fix #9911. I created a group in `pcViewProviderRoot` called `objectGroup` which contains basically everything that is allowed to affect the rotation center. Things that should not affect the rotation center are directly added as a child of `pcViewProviderRoot`. The axis cross is directly added to `pcViewProviderRoot` and should no longer affect the rotation center. I also excluded the measure distance tool from `objectGroup` because I think it could affect the rotation center too much in some cases.

I am not sure if this is the best way to fix the object center rotation mode but I think it is a step in the right direction. If there is a better way to fix this or if there are more things that should be excluded from the `objectGroup` please let me know.

This PR does not depend on #9909 but I recommend it for testing.

### DIFF
--- a/src/Gui/View3DInventorViewer.cpp
+++ b/src/Gui/View3DInventorViewer.cpp
@@ -332,6 +332,7 @@ View3DInventorViewer::View3DInventorViewer(QWidget* parent, const QtGLWidget* sh
     : Quarter::SoQTQuarterAdaptor(parent, sharewidget)
     , SelectionObserver(false, ResolveMode::NoResolve)
     , editViewProvider(nullptr)
+    , objectGroup(nullptr)
     , navigation(nullptr)
     , renderType(Native)
     , framebuffer(nullptr)
@@ -350,6 +351,7 @@ View3DInventorViewer::View3DInventorViewer(const QtGLFormat& format, QWidget* pa
     : Quarter::SoQTQuarterAdaptor(format, parent, sharewidget)
     , SelectionObserver(false, ResolveMode::NoResolve)
     , editViewProvider(nullptr)
+    , objectGroup(nullptr)
     , navigation(nullptr)
     , renderType(Native)
     , framebuffer(nullptr)
@@ -477,6 +479,11 @@ void View3DInventorViewer::init()
     pcEditingRoot->addChild(pcEditingTransform);
     pcViewProviderRoot->addChild(pcEditingRoot);
 
+    // Create group for the physical object
+    objectGroup = new SoGroup();
+    objectGroup->ref();
+    pcViewProviderRoot->addChild(objectGroup);
+
     // Set our own render action which show a bounding box if
     // the SoFCSelection::BOX style is set
     //
@@ -564,6 +571,8 @@ View3DInventorViewer::~View3DInventorViewer()
     coinRemoveAllChildren(this->pcViewProviderRoot);
     this->pcViewProviderRoot->unref();
     this->pcViewProviderRoot = nullptr;
+    this->objectGroup->unref();
+    this->objectGroup = nullptr;
     this->backlight->unref();
     this->backlight = nullptr;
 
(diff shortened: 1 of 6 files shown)


0penBrain:
> I am not sure if this is the best way to fix the object center rotation mode

The main problem is lack of genericity.
The good way IMO would be to mimic `canAddToSceneGraph`. Add another virtual function (something like `virtual bool excludeFromPhysicalObject() const { return false; }` to `ViewProvider.h` so then any VP that needs to be excluded can override this function to return `true`.

Then keep the separate SoGroup as you did but decide if VP goes in or out based on the `excludeFromPhysicalObject` return.

Rexbas:
Sounds like a good idea. I will try to make that work.

src/Gui/View3DInventorViewer.cpp
```diff
@@ -115,6 +115,7 @@
 #include "ViewProvider.h"
 #include "ViewProviderDocumentObject.h"
 #include "ViewProviderLink.h"
+#include "ViewProviderMeasureDistance.h"
```
0penBrain, line 118: This shouldn't be needed anymore.

src/Gui/ViewProvider.h
```diff
@@ -142,6 +142,8 @@ class GuiExport ViewProvider : public App::TransactionalObject
     virtual SoSeparator* getBackRoot() const;
     ///Indicate whether to be added to scene graph or not
     virtual bool canAddToSceneGraph() const {return true;}
+    // Indicate whether to be added to scene graph root or object group root
```
0penBrain, line 145: The comment isn't very clear. Especially because we expect toe find the `true` value first then the `false` one.
```suggestion
    // Indicate whether to be added object group (true) or only to scene graph (false)
```

src/Gui/View3DInventorViewer.h
```diff
@@ -484,6 +484,7 @@ class GuiExport View3DInventorViewer : public Quarter::SoQTQuarterAdaptor, publi
     SoDirectionalLight* backlight;
 
     SoSeparator * pcViewProviderRoot;
+    SoGroup* objectGroup;
```
0penBrain, line 487: Adding a short comment to each line to get the purpose of each would be good.

0penBrain (commented):
LGTM except some remarks.

Just read the code, I didn't compile an test though.

Rexbas:
Thanks for the feedback. I applied your suggestions and added more comments. :)

0penBrain:
Welcome. I won't go further now as I think we gonna wait after 0.21 release to merge (except if one decides otherwise).
When it's time to merge, I can give it a try if needed.

yorikvanhavre:
Anything left here or is it good for merging?

Rexbas:
I think it is good, unless @0penBrain wants to test it first.

0penBrain:
Didn't re-check but nothing wrong from my side

luzpaz:
@yorikvanhavre :point_up_2:

luzpaz:
@yorikvanhavre is AFK perhaps @chennes can merge?   
@Rexbas can you rebase ?

chennes:
Sure, sounds like everyone is happy about it.

adrianinsaval:
should we backport this?

Rexbas:
I think backporting this would be a good idea.

PR #9972, opened 2023-07-23
allow reduction of Part ViewProvider tesselation resolution

bdieterm:

---

**Problem description**
To display a shape, the ViewProvider creates a mesh of the shape data. The resolution depends also on the two parameters "Angular Deflection" and "Deviation" in the View properties. (The values of these parameters are also updated for all objects when changing these settings in the Preferences dialog).
Changing the parameters from a high value to a lower value has an immediate effect on the displayed shape -- for example a circle will be displayed finer with more line segments. But changing the parameter back to the higher value (with less resolution) will keep the finer view-mesh instead of calculating a coarser one. To get back to the coarse view, the object has to be manually recomputed or the file re-opened.

**Solution**
I just set the meshing parameter `AllowQualityDecrease` to True. I guess that this was not implemented before because this parameter is quite new (introduced in OCC v7.5 in the year 2020).

The attached animation demonstrates the new behaviour: The resoultion of the view-mesh is now updated in both directions instead of only towards finer resolution.

demo_tesselation

### DIFF
--- a/src/Mod/Part/Gui/ViewProviderExt.cpp
+++ b/src/Mod/Part/Gui/ViewProviderExt.cpp
@@ -951,7 +951,19 @@ void ViewProviderPartExt::updateVisual()
 
         // create or use the mesh on the data structure
         Standard_Real AngDeflectionRads = AngularDeflection.getValue() / 180.0 * M_PI;
+
+#if OCC_VERSION_HEX >= 0x070500
+        IMeshTools_Parameters meshParams;
+        meshParams.Deflection = deflection;
+        meshParams.Relative = Standard_False;
+        meshParams.Angle = AngDeflectionRads;
+        meshParams.InParallel = Standard_True;
+        meshParams.AllowQualityDecrease = Standard_True;
+
+        BRepMesh_IncrementalMesh(cShape, meshParams);
+#else
         BRepMesh_IncrementalMesh(cShape, deflection, Standard_False, AngDeflectionRads, Standard_True);
+#endif
 
         // We must reset the location here because the transformation data
         // are set in the placement property


src/Mod/Part/Gui/ViewProviderExt.cpp
```diff
@@ -951,7 +951,19 @@ void ViewProviderPartExt::updateVisual()
...
+#else
+        IMeshTools_Parameters meshParams;
+        meshParams.Deflection = deflection;
+        meshParams.Relative = Standard_False;
+        meshParams.Angle = AngDeflectionRads;
+        meshParams.InParallel = Standard_True;
+        meshParams.AllowQualityDecrease = Standard_True;
+
+        BRepMesh_IncrementalMesh(cShape, meshParams);
+#endif
```
0penBrain, line 966: If works, something like this seems cleaner.
```suggestion
        IMeshTools_Parameters meshParams;
        meshParams.Deflection = deflection;
        meshParams.Relative = Standard_False;
        meshParams.Angle = AngDeflectionRads;
        meshParams.InParallel = Standard_True;
#if OCC_VERSION_HEX >= 0x070500
        meshParams.AllowQualityDecrease = Standard_True;
#endif
        BRepMesh_IncrementalMesh(cShape, meshParams);
```

0penBrain (changes requested):
Please at least squash your commits.
Dealing with intermediate faulty commits is a plague. :wink:

src/Mod/Part/Gui/ViewProviderExt.cpp
adrianinsaval: It seems IMeshTools_Parameters was introduced in 7.4 and our minimum supported version is 7.3, so I think the PR is correct as is.

ref: https://dev.opencascade.org/doc/occt-7.4.0/overview/html/occt_dev_guides__upgrade.html

bdieterm: Exactly, @adrianinsaval!

The `Ubuntu_20-04` Build uses OCC 7.3.0

`IMeshTools_Parameters` was introduced in OCC 7.4.0
https://old.opencascade.com/doc/occt-7.3.0/refman/html/class_b_rep_mesh___incremental_mesh.html
https://old.opencascade.com/doc/occt-7.4.0/refman/html/class_b_rep_mesh___incremental_mesh.html

`AllowQualityDecrease` was introduced in OCC 7.5.0
https://old.opencascade.com/doc/occt-7.4.0/refman/html/struct_i_mesh_tools___parameters.html
https://old.opencascade.com/doc/occt-7.5.0/refman/html/struct_i_mesh_tools___parameters.html

chennes:
Needs back port to 0.21

Roy-043:
Backported:
cf9ed2b070d8172933e03445fa65dd0431d19d5b

PR #9974, opened 2023-07-23
[Bugfix]Base: fix implementation of isFile() so it is safe to use

0penBrain:
Current implementation can freeze the app if called on specific files

This was detected in the CI of #9668.

#9668 CI should be retriggered and hopefully can be merged when this present PR is merged.

I didn't search which file / kind of files exactly triggers the problem, but it may be done if of interest.

### DIFF
--- a/src/Base/FileInfo.cpp
+++ b/src/Base/FileInfo.cpp
@@ -367,25 +367,21 @@ bool FileInfo::setPermissions(Permissions perms)
 
 bool FileInfo::isFile() const
 {
-#ifdef FC_OS_WIN32
     if (exists()) {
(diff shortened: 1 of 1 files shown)


src/Base/FileInfo.cpp
```diff
@@ -367,25 +367,21 @@ bool FileInfo::setPermissions(Permissions perms)
 
 bool FileInfo::isFile() const
 {
+        if (exists()) {
```
AjinkyaDahale, line 370: Indentation seems off.

0penBrain: Thanks

PR #9975, opened 2023-07-23
[Bugfix]Gui: detach field sensor of draggers and handle post-call of callbacks

### LINKED ISSUE #9465: [Problem]  Coin version on Mac Conda builds causes transform crash

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: macOS 10.16
Word size of FreeCAD: 64-bit
Version: 0.21.0.33040 (Git)
Build type: Release
Branch: master
Hash: 1e830a8800156097ccb3ef6124dced30d4002e76
Python 3.11.3, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.2, OCC 7.7.0
Locale: C/Default (C)
Installed mods: 
  * MeshRemodel 1.8919.0
  * DynamicData 2.46.0
  * lattice2.backup1682566094.626239 1.0.0 (Disabled)
  * fcgear 1.0.0
  * workfeature
  * Pyramids-and-Polyhedrons
  * QuickMeasure-main
  * GDML 1.8.0
  * offline-documentation
  * Manipulator 1.5.0
  * fasteners 0.4.56
  * lattice2 1.0.0
  * CurvedShapes 1.0.4
  * toSketch 1.0.1
  * Help 1.0.3
  * Curves 0.6.8
[/code]
```

### Subproject(s) affected?

None

### Problem description

Create a plane with Part|Create Primitives
Translate it  with the Transform Tool
FreeCAD crashes when the dialog is closed.
Seems to be MacOS specific.

### Anything else?

Forum link https://forum.freecad.org/viewtopic.php?t=78077

### Code of Conduct

0penBrain:
This is an attempt to fix #9465 

I can't test at all (no Mac) but AFAIK nobody really can as this only happens on prebuilt versions, and not on own build ones.
So at some time, only way to test may be to merge and see. :smile:

I did this because the reported framestack makes me think to symptoms of a Coin sensor that wouldn't be correctly detached.
Looking at the code, I indeed found the `fieldSensor` of `TDragger` and `RDragger` to not be detached.*They now are detached in instance dtor.

As sensor data is instance itself (`this`) and because Coin can still call the static callback after the instance is destructed, I make sure to nullify the sensor data in dtor and check for it being not null in sensor callback.

:crossed_fingers: for Mac users :rofl:

### DIFF
--- a/src/Gui/SoFCCSysDragger.cpp
+++ b/src/Gui/SoFCCSysDragger.cpp
@@ -128,7 +128,8 @@ TDragger::TDragger()
 
 TDragger::~TDragger()
 {
-
+    fieldSensor.setData(nullptr);
+    fieldSensor.detach();
 }
 
 void TDragger::buildFirstInstance()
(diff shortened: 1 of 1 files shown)


chennes:
It doesn't seem like there is any downside to this PR -- I propose that we merge it and then make sure to get a new Conda build ASAP. 0.21 is about ready and it would be good to have this bug resolved.

0penBrain:
At the worst I don't think it can break something. Anyway we can keep an eye on it and promptly revert if needed.

PR #9977, opened 2023-07-23
Qt6 port

wwmayer:
Fix deprecation warnings with version 6.3 or 6.4

---

### DIFF
--- a/src/App/Document.cpp
+++ b/src/App/Document.cpp
@@ -903,7 +903,11 @@ std::string Document::getTransientDirectoryName(const std::string& uuid, const s
     // Create a directory name of the form: {ExeName}_Doc_{UUID}_{HASH}_{PID}
     std::stringstream s;
     QCryptographicHash hash(QCryptographicHash::Sha1);
+#if QT_VERSION < QT_VERSION_CHECK(6,3,0)
     hash.addData(filename.c_str(), filename.size());
+#else
+    hash.addData(QByteArrayView(filename.c_str(), filename.size()));
+#endif
     s << App::Application::getUserCachePath() << App::Application::getExecutableName()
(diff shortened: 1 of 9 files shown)


chennes:
This PR resolves those deprecation warnings for me. FYI, when compiling against 6.6.0a1 I get a new deprecation. Here is an example:
```
/home/chennes/FreeCAD/src/Mod/Sketcher/Gui/TaskSketcherElements.cpp:1083:13: warning: ‘constexpr typename std::add_const<_Tp>::type& qAsConst(T&) [with T = QToolButton*; typename std::add_const<_Tp>::type = QToolButton* const]’ is deprecated: Use std::as_const() instead. [-Wdeprecated-declarations]
 1083 |     qAsConst(ui->filterButton)->addAction(action);
      |     ~~~~~~~~^~~~~~~~~~~~~~~~~~
/media/chennes/Qt/qt6-install/include/QtCore/qttypetraits.h:50:45: note: declared here
   50 | constexpr typename std::add_const<T>::type &qAsConst(T &t) noexcept { return t; }
      |                                             ^~~~~~~~
```

wwmayer:
TBH I don't understand the use of qAsConst in this context which was added with 15dd77512eb582.
IMO, it should be removed here.

wwmayer:
@chennes 
What about the line: `qAsConst(ui->filterButton)->addAction(action);` ? Shouldn't we replace it with `ui->filterButton->addAction(action);`

Btw, the same line occurs in TaskSketcherConstraints::TaskSketcherConstraints

chennes:
Yes, I figured we'd do the forward-looking deprecations in a separate PR -- what we have now should work fine with Qt 6.4, I was testing on a version of 6.6 from git in my comment above.

Issue #9976, opened 2023-07-23
[Problem] Face color in preferences not working


benj5378:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
OS: Ubuntu 22.10 (ubuntu:GNOME/ubuntu)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33631 (Git)
Build type: Debug
Branch: iss9
Hash: 2167fb6f7ec6ac5c2f6365d4c3035d2db6879b1e
Python 3.10.7, Qt 5.15.6, Coin 4.0.0, Vtk 7.1.1, OCC 7.6.3
Locale: English/United States (en_US)
```

### Subproject(s) affected?

Techdraw

### Problem description

If one sets a face color in preferences, it won't take effect in TechDraw

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #9978, opened 2023-07-23
[TD]fix face colour preference ignored issue #9976

WandererFan:
This PR implements a fix for issue #9976 

---

### DIFF
--- a/src/Mod/TechDraw/Gui/QGIFace.cpp
+++ b/src/Mod/TechDraw/Gui/QGIFace.cpp
@@ -81,8 +81,12 @@ QGIFace::QGIFace(int index) :
 
     getParameters();
 
+    // set up style & colour defaults
     m_styleDef = Qt::SolidPattern;
     m_styleSelect = Qt::SolidPattern;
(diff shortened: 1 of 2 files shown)


benj5378:
@WandererFan I have a better fix for this upcoming also simplifying stuff in Preferences. Just ran into some issues with base functionality

benj5378:
#9979

WandererFan:
So why is #9979 better?  

It brings in 2 changes that aren't related to issue #9976
- changing App::Color interpretation of transparency, which potentially affects everywhere App::Color is used, and 
- removes the ability to specify see-through faces (at least I don't see how to specify it) by removing the "transparent faces" option.

benj5378:
@WandererFan Transparency is now an alpha value inside the color picker. That IMO simplifies color handling in code.

> changing App::Color interpretation of transparency, which potentially affects everywhere App::Color is used, and

This change would have to be done anyway at some point.

WandererFan:
see comments about transparency in PR9979.  In this case,  transparency refers to an empty fill,  rather than a shaded colour.

PR #9979, opened 2023-07-23
[TechDraw + App] Make faces use preference color setting and fix handling of alpha value

benj5378:

---

### DIFF
--- a/src/App/Color.h
+++ b/src/App/Color.h
@@ -41,9 +41,9 @@ class AppExport Color
 public:
     /**
      * Defines the color as (R,G,B,A) whereas all values are in the range [0,1].
-     * \a A defines the transparency.
+     * @param A defines the transparency.
      */
-    explicit Color(float R=0.0,float G=0.0, float B=0.0, float A=0.0);
+    explicit Color(float R=0.0,float G=0.0, float B=0.0, float A=1.0);
 
     /**
      * Does basically the same as the constructor above unless that (R,G,B,A) is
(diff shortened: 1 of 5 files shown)


src/App/Color.h
```diff
@@ -41,9 +41,9 @@ class AppExport Color
 public:
     /**
      * Defines the color as (R,G,B,A) whereas all values are in the range [0,1].
-     * \a A defines the transparency.
+     * @param A defines the transparency.
      */
-    explicit Color(float R=0.0,float G=0.0, float B=0.0, float A=0.0);
+    explicit Color(float R=0.0,float G=0.0, float B=0.0, float A=1.0);
```
benj5378, line 46: One should evaluate this change very carefully!!

Issue is, that some functions App::Color considers `A = 1.0` as transparent while other functions considers `A = 0.0` as transparent.

I have here chosen, to redefine transparent as `A = 0` as it harmonizes how alpha values are handled elsewhere like QColor and HTML. Though this could break various stuff! Some code might need to be adopted to this change, as it will else cause objects to suddenly become transparent.

This is exactly what happened to me, as I tried implementing alpha value in the `asValue` function: suddenly all objects became transparent, as I had implemented the functionality according to documentation where transparency is defined as `A = 1`, while `asPackedRGBA` defined transparency as `A = 0` making all alpha values suddenly being retrieved opposite than saved.

WandererFan:
There are a few things here:
- we can't change a core component like App::Color without some prior discussion to understand the impact 
- we don't want to change App::Color just before a release.  It may need changing, I don't know, but not now
- transparency in this case refers to a using a transparent fill, not opacity of a colour fill.  Might be a cool addition, but again, not now.
- we have a code defect that needs fixing before release, so the fix should be small and focused.

I'm going to merge #9978 instead of this one.  You can reintroduce the App::Color and opacity stuff in v0.22.

benj5378:
> transparency in this case refers to a using a transparent fill, not opacity of a colour fill.

@WandererFan It is the same. Qt::Transparent currently used for transparent faces is just an alias for opaqueness = 0. https://doc.qt.io/qt-6/qt.html#GlobalColor-enum

WandererFan:
stale and CI errors.  closing.

Issue #9990, opened 2023-07-24
[Problem] TechDraw page returns incorrect width and height in Python


benj5378:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
OS: Ubuntu 22.10 (ubuntu:GNOME/ubuntu)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33650 (Git)
Build type: Debug
Branch: iss11
Hash: 0fee9c98f036d18762b6eceb4a6aecbc8f9f948d
Python 3.10.7, Qt 5.15.6, Coin 4.0.0, Vtk 7.1.1, OCC 7.6.3
Locale: English/United States (en_US)
```

### Subproject(s) affected?

Techdraw

### Problem description

If one tries to retrieve the page width and height before the TechDraw page has been opened in the gui, then one gets the value `0`.

Example code from @Roy-043 

```
import FreeCAD as App

doc = App.ActiveDocument
box = doc.addObject("Part::Box", "Box")

page = doc.addObject("TechDraw::DrawPage", "Page")
template = doc.addObject("TechDraw::DrawSVGTemplate", "Template")
template.Template = App.getResourceDir() + "Mod/TechDraw/Templates/A4_LandscapeTD.svg"
page.Template = template

view = doc.addObject("TechDraw::DrawViewPart", "View")
page.addView(view)
view.Source = [box]
view.Direction = (0.0, 0.0, 1.0)

doc.recompute()

page.PageWidth  // Returns 0
page.PageHeight  // Returns 0
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

benj5378:
https://forum.freecad.org/viewtopic.php?p=695067#p695067

luzpaz:
What's the status on this ?

WandererFan:
It is build on top of a PR that doesn't work.

WandererFan:
Should be fixed by commit 622df46971.

benj5378:
@WandererFan Is #9991 still relevant?

benj5378:
@WandererFan Is https://github.com/FreeCAD/FreeCAD/pull/9991 still relevant?

WandererFan:
The part about using only 1 property for the template would be good but hard to implement while preserving backwards compatibility.   Making PageResult hidden in the tree might solve most of the confusion around having 2 properties with only a 1 line change.

PR #9991, opened 2023-07-24
[TechDraw] Process template upon filepath specification (#9990)

benj5378:

---

### DIFF
--- a/src/Mod/TechDraw/App/DrawSVGTemplate.cpp
+++ b/src/Mod/TechDraw/App/DrawSVGTemplate.cpp
@@ -52,8 +52,7 @@ DrawSVGTemplate::DrawSVGTemplate()
 {
     static const char *group = "Template";
 
-    ADD_PROPERTY_TYPE(PageResult, (nullptr),  group, App::Prop_Output,    "Embedded SVG code for template. For system use.");   //n/a for users
-    ADD_PROPERTY_TYPE(Template,   (""), group, App::Prop_None, "Template file name.");
+    ADD_PROPERTY_TYPE(Filepath, (nullptr), group, App::Prop_Output, "Filename containing the SVG markup for the template");   //n/a for users
 
     // Width and Height properties shouldn't be set by the user
     Height.setStatus(App::Property::ReadOnly, true);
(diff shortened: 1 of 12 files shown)


benj5378:
Of what I tested it works fine, but it would be really sad to have introduced a bug into 21.0.

Also, we should probably write some more tests...

src/Mod/TechDraw/App/DrawSVGTemplate.cpp
```diff
@@ -181,17 +181,24 @@ double DrawSVGTemplate::getHeight() const
...
+    Base::FileInfo tfi(Filepath.getValue());
+    if (!tfi.isReadable()) {
+        // If there is an old absolute template file set use a redirect
+        const std::string newPath = App::Application::getResourceDir() + "Mod/Drawing/Templates/" + tfi.fileName();
+        tfi.setFile(newPath);
+        // Try the redirect
+        if (tfi.isReadable()) {
+            Filepath.setValue(newPath.c_str());
+            return;
+        }
```
benj5378, line 200: I haven't tested this

WandererFan:
Once upon a time, template worked like the DrawingWB version and the svg(PageResult) was modified with the editable fields when they changed.  Now we only merge in the editable field values as required in processTemplate(), so we don't need to keep the modified svg.

The v0.21 tag is supposed to be created this weekend(?), so we can merge this next week maybe.

WandererFan:
Add new page using default template (USLetter_Landscape_blank.svg) gives message:
20:21:12  DSVGT::checkFilepath(/home/wile/.cache/FreeCAD/Cache/FreeCAD_Doc_c90b97cc-6262-46ce-a7d8-fd2e7636952b_da39a3_1884894/USLetter_Landscape_blank.svg)

and again on changing FilePath property of Template to A4_Landscape_ISO7200TD.svg:
20:22:41  DSVGT::checkFilepath(/home/wile/.cache/FreeCAD/Cache/FreeCAD_Doc_c90b97cc-6262-46ce-a7d8-fd2e7636952b_da39a3_1884894/USLetter_Landscape_blank.svg)

clicking on a green edit handle and changing a field is not reflected on screen, but the EditableTexts property is updated.  Closing and reopening document updates the EditableTexts on screen.

Open pre-PR9991 document: 
Template.FilePath is empty.  Default template used instead of the template specified in the document.

WandererFan:
This PR is too stale and template processing has changed in the interim.  Closing.

PR #9998, opened 2023-07-26
fix use ccache by default on most systems

adrianinsaval:

---

to fix my mistake at #9950 I had tested that one locally and ccache was working but turns out it was because in my system ccache's binaries have higher priority in path than the true compiler anyways.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,12 @@ if (POLICY CMP0072)
     set(OpenGL_GL_PREFERENCE LEGACY)
 endif(POLICY CMP0072)
 
+if (BUILD_WITH_CONDA AND WIN32)
+    option(FREECAD_USE_CCACHE "Auto detect and use ccache during compilation" OFF)
+else()
+    option(FREECAD_USE_CCACHE "Auto detect and use ccache during compilation" ON)
+endif()
+
 if(FREECAD_USE_CCACHE)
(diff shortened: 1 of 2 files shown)


adrianinsaval:
I re ran the ubuntu 2204 build to actually test the cache and this makes it work again.

CMakeLists.txt
```diff
@@ -10,6 +10,12 @@ if (POLICY CMP0072)
     set(OpenGL_GL_PREFERENCE LEGACY)
 endif(POLICY CMP0072)
 
+if (BUILD_WITH_CONDA AND WIN32)
```
wwmayer, line 13: BUILD_WITH_CONDA is used now before it's defined in InitializeFreeCADBuildOptions.cmake. What about writing:
```
if (WIN32)
    option(FREECAD_USE_CCACHE "Auto detect and use ccache during compilation" OFF)
else()
    option(FREECAD_USE_CCACHE "Auto detect and use ccache during compilation" ON)
endif()
```
?

adrianinsaval: BUILD_WITH_CONDA has to be passed as an argument to cmake AFAIK so it should exist when relevant I think, I'll check again.

But ccache is not very effective on windows anyways, at least not when using msvc so we could disable it by default on windows, than if someone really wants to use it they can enable it.

PR #10000, opened 2023-07-27
TechDraw Check for valid selection before asking for destination page

domi4484:
When having multiple Tech Draw pages, and no one of them is the currently displayed page, on some operations the user get asked which is the destination page before checking that the selection is valid.

This PR propose to postpone the page choice after checking that the user selection is valid.


luzpaz:
Please amend commit title to `TechDraw: Check for valid selection before asking for destination page`
using `git commit --amend` for example

WandererFan:
This PR makes the change for ProjectionGroup and Draft, but not for the other view types that behave the same way.  Is there a reason to not make this change for Section, ComplexSection, Detail and Arch?

domi4484:
> This PR makes the change for ProjectionGroup and Draft, but not for the other view types that behave the same way. Is there a reason to not make this change for Section, ComplexSection, Detail and Arch?

Thanks I didn't noticed it. I'll change there as well.

domi4484:
@WandererFan for DrawView it was also missing.

I noticed that for "Detail" page was not used at all. I suppose that DrawViewPart can exist only inside a TechDraw?

luzpaz:
Just checking on the status of this PR

PR #10012, opened 2023-07-28
[TechDraw] seam line in HLR default false

grubertechnic:

- [ X]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

---

### DIFF
--- a/src/Mod/TechDraw/Gui/DlgPrefsTechDrawHLR.ui
+++ b/src/Mod/TechDraw/Gui/DlgPrefsTechDrawHLR.ui
@@ -260,7 +260,7 @@ Fast, but result is a collection of short straight lines.</string>
            <string>Show Seam Lines</string>
           </property>
           <property name="checked">
-           <bool>true</bool>
+           <bool>false</bool>
           </property>
           <property name="prefEntry" stdset="0">
            <cstring>SeamViz</cstring>
--- a/src/Mod/TechDraw/Gui/DlgPrefsTechDrawHLRImp.cpp
+++ b/src/Mod/TechDraw/Gui/DlgPrefsTechDrawHLRImp.cpp
@@ -58,7 +58,11 @@ void DlgPrefsTechDrawHLRImp::saveSettings()
 
 void DlgPrefsTechDrawHLRImp::loadSettings()
 {
+    // set defaults for HLR
+    bool visibleSeamDefault = false;
+    ui->pcbSeamViz->setChecked(visibleSeamDefault);
     ui->pcbSeamViz->onRestore();
+
     ui->pcbSmoothViz->onRestore();
     ui->pcbHardViz->onRestore();
     ui->pcbPolygon->onRestore();


grubertechnic:
@WandererFan : question, i do not get the italic of the preferences. I do not see where the italic changes in the code, besides i barely see a difference, even with my new glasses.

Is that already needed? Because this statement is not only true in TechDraw, it is true for all objects in FreeCAD.

Thanks in advance!

benj5378:
@grubertechnic As I understand it, properties marked italics in preferences will be used as the default for new object (whatever the preference value is set to).

grubertechnic:
Thanks. Since pretty everything is italic (when i can trust my eyes), then it makes not much sense. Also some parameters are not italic, but have a similar tooltip (example in `TechDraw --> General --> Label Font*`) or none, but still affected.

I have not looked yet, if this note is in the TechDraw preferences ui or in every tab singular. But anyway, it is not important.

WandererFan:
"I do not see where the italic changes in the code, besides i barely see a difference, even with my new glasses."

Can't help with the glasses.  I use QtDesigner:
italicsInDesigner

but if you really want to edit the ui code...
`class Ui_DlgPrefsTechDrawHLRImp
{
public:
...
    Gui::PrefCheckBox *pcbSeamViz;
...
        QFont font;
        font.setItalic(true);
...
        pcbSeamViz = new Gui::PrefCheckBox(gbMisc);
        pcbSeamViz->setObjectName(QString::fromUtf8("pcbSeamViz"));
        sizePolicy1.setHeightForWidth(pcbSeamViz->sizePolicy().hasHeightForWidth());
        pcbSeamViz->setSizePolicy(sizePolicy1);
        pcbSeamViz->setFont(font);  // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
        pcbSeamViz->setChecked(true);
        pcbSeamViz->setProperty("prefEntry", QVariant(QByteArray("SeamViz")));
        pcbSeamViz->setProperty("prefPath", QVariant(QByteArray("Mod/TechDraw/HLR")));

        gridLayout_6->addWidget(pcbSeamViz, 4, 0, 1, 1);

`

WandererFan:
"Because this statement is not only true in TechDraw, it is true for all objects in FreeCAD."

There are definitely properties where changing the preference will affect existing objects (may need a recompute).  LabelFont on the general tab is one example.

grubertechnic:
I mean, if a label is italic or not is barely readable (at least for me) in the GUI itself, especially when it sis not in a floating text.

> LabelFont on the general tab is one example.
Only on labels, not on the dimensions.

Anyway, it is not important now and for this.

The questions is more,

is the PR valid?
if yes, 0.21 or 0.22? (since this issue occurs regular)

WandererFan:
You don't need the change in DlgPrefsTechDrawHLRImp, the change to the ui will handle it. 
You do need a change in DrawViewPart to use the new default if the parameter is not set in user.cfg.   You need to change the true to false here:
    ADD_PROPERTY_TYPE(SmoothVisible, (Preferences::getPreferenceGroup("HLR")->GetBool("SmoothViz", true)),
        sgroup, App::Prop_None, "Show Visible Smooth lines");
    ADD_PROPERTY_TYPE(SeamVisible, (Preferences::getPreferenceGroup("HLR")->GetBool("SeamViz", true)),
        sgroup, App::Prop_None,
                      "Show Visible Seam lines");

benj5378:
I can easily see what is italic, even though it isn't as distinct as with other fonts. That said, I do acknowledge that we could improve this. Why do we even have to do stuff italic and then make a label saying "italic settings will become default for new objects"? We could just have had a label at top stating "These settings will be used as default for new object", when all setting are italic anyway. And in any case, two group boxes would probably be better than marking some some properties italic and others.

What do you think of this change? Should we make it?

WandererFan:
Some preferences are only used on new objects, some are used every time the object is recomputed/redrawn.  There needs to be some way of distinguishing these cases for the user.

We could use BOLD or an "*" or a different colour.  I'm not committed to italic.

benj5378:
I'm generally not into using the font for communicating this. What about this instead, where one can select if the setting should also be used when recomputing. This would allow people for opt-in and out.

```
+-------------------+--------------------+--------------------+
| Setting           | Use for new object | Use on recomputing |
+===================+====================+====================+
| Show smooth lines | X                  | X                  |
+-------------------+--------------------+--------------------+
| Show UV ISO lines | X                  |                    |
+-------------------+--------------------+--------------------+
| Show seam lines   | X                  | X                  |
+-------------------+--------------------+--------------------+```

WandererFan:
> I'm generally not into using the font for communicating this. What about this instead, where one can select if the setting should also be used when recomputing. This would allow people for opt-in and out.

But that is a much bigger change that just flipping a boolean preference.  Every place where showSmoothLines is used would have to change to use the 2nd preference instead of the object property.

WandererFan:
This commit and the changes mentioned in the comments are included in PR #10019

WandererFan:
closing in favour of #10019

benj5378:
> But that is a much bigger change that just flipping a boolean preference. Every place where showSmoothLines is used would have to change to use the 2nd preference instead of the object property.

That is true :/ ... though I still consider it an improvement. Consider it low priority

grubertechnic:
> You don't need the change in DlgPrefsTechDrawHLRImp, the change to the ui will handle it.

I know, but i think that is not really "clean" way. But also i am not a programmer.

> You do need a change in DrawViewPart to use the new default if the parameter is not set in user.cfg.

Thanks, maybe next time (if github let me, i will not do the 2FA, its BS).

Issue #7640, opened 2022-10-24
[Problem] Users don't understand stylesheets vs preference packs


MisterMakerNL:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=8&t=72889&p=635896#p635896

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.30345 (Git)
Build type: Release
Branch: master
Hash: f42a47aa0fd8c4f34ef63c185d9f29b9b6e84737
Python 3.10.6, Qt 5.15.4, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: Dutch/Netherlands (nl_NL)
Installed mods: 
  * Assembly3 0.11.3
  * Behave-Dark-Colors 0.1.1
  * BIM 2021.12.0 (Disabled)
  * Dracula 0.0.2
  * ExtremeProDark 2.3.2
  * ExtremeProDark.backup1665346471.659064 2.3.1 (Disabled)
  * ExtremeProLight 1.0.1
  * ExtremeProLight.backup1665559197.7999926
  * FeedsAndSpeeds 0.4.0
  * FEM_FrontISTR 0.1.0
  * IconThemes
  * lattice2 1.0.0
  * PieMenu 1.1.10
  * PieMenu.backup1664822562.2342157
  * ProDarkThemePreferencePack 1.0.0
  * scanObjects
  * SelectorToolbar
  * sheetmetal 0.2.58
  * SP_Startup
  * ToolbarStyle
[/code]
```

### Issue description

Ok so people get confused to when to use a style-sheet or when a preference pack.
Also the preference pack has colors that you would expect in the style-sheet. But who cares? (probably people who read this)
But as an end user you just want a good looking Freecad, and it all works like you expect.

Now the cool thing is, it all works very well already so the hard work is already done for 99%.

Now we only need to fine tune it a bit.
And my proposal is too rename "Preference packs" to "Theme's" and style-sheets too "Highlights".
We need to link the style-sheet to a preference pack since they work together. Like you can't use a light theme preference pack with a dark stylesheet that doesn't work.
It could look like this: 
![afbeelding](https://user-images.githubusercontent.com/29804962/197600437-898f608a-dac8-489f-9e57-8fd06743b80d.png)

[...]

PR #10015, opened 2023-07-30
[Preferences] Theme

PaddleStroke:
As discussed here : 
https://github.com/FreeCAD/FreeCAD/issues/7640

- Addons that are Themes specify it in the package.xml as @chennes suggested :

```
  <content>
    <preferencepack>
      <name>ProDarkThemePack</name>
      <type>Theme</type>
      <description>ProDark stylesheet and Colour scheme</description>
      <version>1.0.0</version>
      <tag>stylesheet</tag>
      <tag>dark</tag>
      <tag>professional</tag>
      <file>ProDark.qss</file>
    </preferencepack>
  </content>
```
- The 'Preference packs' widget does NOT show the preference packs of type 'Theme'.

- The "StyleSheet" combobox is moved to the new Theme pref page.

- A "Theme" combobox replace the StyleSheet in general: This combobox shows all the preference packs of type 'Theme'. When the user select a theme it applies the preference pack.

- Adds a secondary color in theme pref page that can be leverraged by stylesheets.

@MisterMakerNL @chennes @adrianinsaval 

---


MisterMakerNL:
Really cool stuff!
I already tested it!
Had some issues with applying a stylesheet. It seems to need a restart to get the full stylesheet working correctly.
Appling old preference-packs from preference pack menu also needs a restart.
afbeelding
afbeelding

PaddleStroke:
Your issue is when you change the stylesheet in the display/theme page ? Or when you change the theme in general?

I can't seem to replicate on my side so if you can give me more details.

Btw what do you think about this secondary color ? It's to replace the numerous stylesheets like 'dark-green' 'dark-blue' ... The color is saved as a preference. Can you access it from the stylesheet?

MisterMakerNL:
ok I just wanted to record it how it looks with the failed stylesheet update and it just worked.

MisterMakerNL:
Something else, could we add a filter on the stylesheets so that only dark stylesheets show up if a dark theme is selected? 
I can then add another theme to it so we have 3:
- light
- dark
- darker
Dark and darker will be the same for now, but would prevent some confusion to the user.

PaddleStroke:
I dunno, perhaps the stylesheets should be abstracted completely to the user. I think it would be best to remove the stylesheet altogether and only offer 'themes' directly.

Maybe we can keep it as it is now, then when other stylesheets are ready to be offered as themes, we can make the move.

MisterMakerNL:
I'm ok with how it is, since now it also doesn't filter so it's still better than what it was. 
People will definitely complain at first because it is different and some are used to it. But it looks more professional like this and for new users more clear.
If we can get the highlight color working I also think we can scrap the style-sheets, although most theme's have a highlight color and a focus color. So that would be a small detail. Although I can work around it if necessary.(can just make the focus a thicker border)

PaddleStroke:
You tell me what parameters you need in the Theme pref page for the theme/stylesheet customization and I'll implement them.

MisterMakerNL:
> You tell me what parameters you need in the Theme pref page for the theme/stylesheet customization and I'll implement them.

For now I would like to have:

- Highlight color
- Focus color
- Toolbutton padding in pixels (this was a request on the forum, if this is possible)

chennes:
I'm very happy to see progress being made on this front, thanks for the PR. Obviously this is something for the 1.0 release cycle, so I'll review it once we've release 0.21.

PaddleStroke:
Yes it needs more polish and time to get all the stylesheet/theme adapting so it's for after the release.

chennes:
There are two other places the metadata change needs to be reflected: the Python wrapper for the C++ class (MetadataPy.xml and MetadataPyImp.cpp), and in the Addon Manager's Metadata Python class (addonmanager_metadata.py).

MisterMakerNL:
I splitted the dark and darker and normal/modern stylesheets into preference packs:
https://github.com/FreeCAD/FreeCAD/pull/9910

benj5378:
@PaddleStroke Please stop nuking all my PRs and get on Discord. You need to get on Discord now! https://discord.gg/QhDVWS7PWc

adrianinsaval:
What are you talking about and why would this need to be discussed on discord?

benj5378:
@adrianinsaval Sorry man, expressed my self wrongly

obelisk79:
@PaddleStroke This PR has made activating Preference Packs impossible, only the defined 'theme' stylesheets can be applied.

PaddleStroke:
I think what you mean is that activating a stylesheet-modifying preference pack does not apply the stylesheet ? Other preference packs works correctly right?

Can you please confirm, say you activate 'prodarkThemePack' which is still a pref pack and not a theme: When you activate it, the stylesheet is not applied right? But if you restart the application, it restarts with the stylesheet applied right? This is how it behaves on my side.

If so, then what you say is that pref-packs that modify stylesheet now require a restart, when before it applied directly. Is that correct?

This happenned when the stylesheet combobox has moved to the theme page. 
Apparantly it's presence on the general page was applying the stylesheet when a preference pack was applied.

If we take a look, applying a pref page reloads the page : 

```
void DlgSettingsGeneral::onLoadPreferencePackClicked(const std::string& packName)
{
    if (Application::Instance->prefPackManager()->apply(packName)) {
        auto parentDialog = qobject_cast<DlgPreferencesImp*> (this->window());
        if (parentDialog)
            parentDialog->reload();
    }
}
```

I'm not sure, but perhaps reload was saving the pref page. By doing so it was saving the stylesheet that was just modified by the pref-pack. And thus applying the stylesheet directly.

@chennes what do you think about this? We could fix that by adding this :

```void DlgSettingsGeneral::onLoadPreferencePackClicked(const std::string& packName)
{
    if (Application::Instance->prefPackManager()->apply(packName)) {
        auto parentDialog = qobject_cast<DlgPreferencesImp*> (this->window());
        if (parentDialog)
            parentDialog->reload();

       // Set the StyleSheet
       QString sheet = QString::fromStdString(hGrp->GetASCII("StyleSheet"));
       bool tiledBackground = hGrp->GetBool("TiledBackground", false);
       Application::Instance->setStyleSheet(sheet, tiledBackground);
    }
}
```

PaddleStroke:
Oh I found the exact reason.

When you apply the pref-pack, it saves all preferences pages. So moving the stylesheet combobox should not have changed anything. 
However I have added a condition to save the stylesheet to prevent setting the stylesheet everytime you save the preferences, which seemed unefficient.

```
void DlgSettingsTheme::saveSettings()
{
    ui->ThemeSecondaryColor->onSave();
    ui->ThemeHighlightColor->onSave();
    ui->ThemeFocusColor->onSave();

    if (styleSheetChanged)
        saveStyleSheet();
}
```

I have done that because for the new theme combobox I added  the same : 
```
    if (themeChanged)
        saveThemes();
```
For the theme it was mandatory because else everytime you would change a setting it would apply the preference pack. Which would save the settings and so on.
Then I also saw that for stylesheet it would be more-efficient to not save it everytime.

Long story short, if we remove this `if (styleSheetChanged)` then pref-packs-apply changes the stylesheet directly as it was before.

However the stylesheet combobox is going to be deleted as soon as all the stylesheets are transfered to Themes. So this would only be a temporary solution.

obelisk79:
@PaddleStroke you're correct on the behavior, however the correct stylesheet does *not* apply after a program restart on my machine.

PaddleStroke:
Can you build FC and try if removing the if statement I mentioned before
from the DlgSettingsTheme solves the problem on your side ? So that we are
sure it's the same underlying issue.

MisterMakerNL:
Please note that I had similar issues when trying this pr but it only
happened first boot. Maybe some old settings?

Op di 8 aug. 2023 15:27 schreef PaddleStroke ***@***.***>:

> Can you build FC and try if removing the if statement I mentioned before
> from the DlgSettingsTheme solves the problem on your side ? So that we are
> sure it's the same underlying issue.
>
>

benj5378:
image
@PaddleStroke

PaddleStroke:
@obelisk79 what about changing the onLoadPreferencePackClicked (line 551  DlgSettingsGeneral) as such : 
```
void DlgSettingsGeneral::onLoadPreferencePackClicked(const std::string& packName)
{
    if (Application::Instance->prefPackManager()->apply(packName)) {
        auto parentDialog = qobject_cast<DlgPreferencesImp*> (this->window());
        if (parentDialog)
            parentDialog->reload();

       // Set the StyleSheet
       QString sheet = QString::fromStdString(hGrp->GetASCII("StyleSheet"));
       bool tiledBackground = hGrp->GetBool("TiledBackground", false);
       Application::Instance->setStyleSheet(sheet, tiledBackground);
    }
}
```

PaddleStroke:
@benj5378 oups I forget to remove this after copy pasting another pref page.

obelisk79:
@PaddleStroke 
"error: ‘hGrp’ was not declared in this scope"

PaddleStroke:
Oups, add this at the beginning of the function :
ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath("User
parameter:BaseApp/Preferences/MainWindow");

PaddleStroke:
This is a good question.
The colors are saved in the preferences. Can you access Freecad preference
from the stylesheet? I think that probably not so we need to find a way.

I think we probably have to edit the stylesheet file when we change the
preferences in freecad.
But for this to work, all stylesheet need to declare those colors at the
same place in the file.

Can you modify the stylesheet in such a way and tell me where the colors
are defined in the files? Then I'll implement in CPP the file edit.

Also make the theme for each different stylesheet (but not for all colors
so instead of darker-green, darker-blue... Just darker).

MisterMakerNL:
This has a bunch of options https://stackoverflow.com/questions/10898399/using-variables-in-qt-stylesheets

MisterMakerNL:
Added a QSS file with the highlight and focus color changed into 
  @ThemeHighlightColor
  @ThemeFocusColor
[Dark.qss.TXT](https://github.com/FreeCAD/FreeCAD/files/12315971/Dark.qss.TXT)

Intuos5:
Wait hold on, @PaddleStroke you want to move the highlight colour to the preferences?
What happens if I want to theme the toolbar button highlight and hover colours differently than elsewhere in the UI? 
I could currently do this in my own stylesheet, but does this mean that won't be possible in the future?

MisterMakerNL:
> Wait hold on, @PaddleStroke you want to move the highlight colour to the preferences? What happens if I want to theme the toolbar button highlight and hover colours differently than elsewhere in the UI? I could currently do this in my own stylesheet, but does this mean that won't be possible in the future?

It only replaces the any variable that you put in it, if you don't want to use it you can just not put in the variable. 
You can also use the variable on other things then highlight.

PaddleStroke:
Yes basically the idea was to offer to theme developpers color settings so
that user can customize the themes.
Basically the initial idea was to remove the need to have multiple themes
like "dark-blue", "dark-orange" ...
That doesn't mean that you have to use it.

Intuos5:
How is this meant to work then in case there are multiple colours in use for the highlight? 
I like the idea of exposing button padding, at least that enables people to adjust from FHD to High DPI displays more easily. But you would have to look at various types of buttons and I'm not sue one padding option covers all of them. I think I did multiple assignments there, at least to cover for the toolbar buttons (File, Edit, etc.).

MisterMakerNL:
> How is this meant to work then in case there are multiple colours in use for the highlight? I like the idea of exposing button padding, at least that enables people to adjust from FHD to High DPI displays more easily. But you would have to look at various types of buttons and I'm not sue one padding option covers all of them. I think I did multiple assignments there, at least to cover for the toolbar buttons (File, Edit, etc.).

Qt doesn't really work with a lot if colors since everything is used mixed together , so if you use a lot of different colors you get a very ugly GUI. But if you can prove me wrong please show an example where this could result into a usable GUI.

Intuos5:
I did this for the toolbar buttons in my theme. The highlight colours for the File row are different from the buttons below, for instance.  

https://github.com/FreeCAD/FreeCAD/assets/63149655/e65af047-4a25-4619-a83a-3869dbb0ede6

MisterMakerNL:
But if you replace teal with any other color and keep the gray just gray it would work perfectly!

PaddleStroke:
@obelisk79 what about changing the onLoadPreferencePackClicked ?

```
void DlgSettingsGeneral::onLoadPreferencePackClicked(const std::string& packName)
{
    if (Application::Instance->prefPackManager()->apply(packName)) {
        auto parentDialog = qobject_cast<DlgPreferencesImp*> (this->window());
        if (parentDialog)
            parentDialog->reload();

       // Set the StyleSheet
       ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/MainWindow");
       QString sheet = QString::fromStdString(hGrp->GetASCII("StyleSheet"));
       bool tiledBackground = hGrp->GetBool("TiledBackground", false);
       Application::Instance->setStyleSheet(sheet, tiledBackground);
    }
}
```

PaddleStroke:
@MisterMakerNL I'm not sure to understand exactly what is the goal of ThemeHighlightColor and ThemeFocusColor. When looking at the stylesheet I saw 'Dark-blue', 'Dark-green', 'Dark-organe'... So that's why initially I implemented ThemeSecondaryColor as a setting. But it seems that in the qss file you sent me you are not using this secondary color but the highlight and focus instead. Can you please tell me if the 'secondary color' and those are redundant ? If you have a screenshot to show me where each color is btw.

MisterMakerNL:
I only asked for highlight colors which is the color that is shown when hovering over buttons and tabs. And the line above tabs.
And the focus color that is uaed the show what element is currently active.
I also asked for a icon padding. But this isn't a color.
Some users like some spacing in their menu buttons, but some want more buttons on the screen. Although the last one could be just getting used too issues. You can already just make the icons bigger which would solve the issue too.
So tldr 2 colors and maybe 1 variable for icon spacing.

PaddleStroke:
Ok so did I understand correctly here : 
image
? 
So the stylesheet dark-Green is actually the dark with highlight color set to green right?

Ok so what I implemented as 'secondary color' was aiming to be this highlight color. So I will remove this 'secondary color' as it's redundant.

For the line spacing, does this actually have to be implemented in the stylesheet ? Can't it just be implement as a normal setting directly in cpp ?

MisterMakerNL:
You almost had it, not sure why your highlight color is blue in a green theme. It should be green.
afbeelding
We could indeed leave the icon spacing out of the style-sheets and move it in setting.

PaddleStroke:
Ok so it seems to me that what you asked for is very specific. Maybe other theme creators will want to customize other elements than those precise ones.
So I think it would be best that I make 2 colors available and call them 'secondary theme color' and 'third theme color'. Then on your side you use the colors for what you want.

MisterMakerNL:
Most elements have 4 states in QT:
Hover > Highlight color
Neutral > Default color often gray, often the same color in every stylesheet for a certain theme. 
Focus > When an element is active.
Pressed > This can have a color but it will just blink and nobody notice it.
afbeelding

PaddleStroke:
I see, but say for instance another theme creator wants to have the focus and highlight both the same color, and he wants to have a third color for something else. Say the menus or the buttons. I don't know just throwing ideas here. Are there actually other theme creators btw? I think @obelisk79 is planning something ? Any thoughts on it?

MisterMakerNL:
Oblisk is looking at the colors in the preference file. The cfg file. Which is very nice. I don't think he is planning any multi color theme. But I do appreciate his opinion since he got good taste 😁.

PaddleStroke:
@MisterMakerNL well let's keep the names as they are for now anyway. So I'm on it now, trying to make the change to the dark.qss file you sent before and see if I can get the preferences colors to work with the stylesheet.

Can you please provide me the .qss with variables for the other stylesheets ?

MisterMakerNL:
I can make them tomorrow!

obelisk79:
I am preparing for a sleep study tonight, but can think more on this tomorrow.

Axeia:
Perhaps it would be wise to model this after Microsoft's design. Windows 10 (and 11) let the user set an 'accent' color.
This color and some calculated variations of it, 6 to be exact.

1. SystemAccentColorLight3
2. SystemAccentColorLight2
3. SystemAccentColorLight1
4. SystemAccentColorDark1
5. SystemAccentColorDark2
6. SystemAccentColorDark3

Information about it can be found over here:
See https://learn.microsoft.com/en-us/windows/apps/design/style/color#accent-color 
It's probably a good idea to use it as the starting values on Windows as well.

How to get the accent color in C++ can be found over here:
https://github.com/res2k/Windows10Colors

Intuos5:
> I see, but say for instance another theme creator wants to have the focus and highlight both the same color, and he wants to have a third color for something else. Say the menus or the buttons. I don't know just throwing ideas here. 

I think this is difficult, for instance, for some buttons I don't use hover colours and make it the same as the background colour (to avoid the button jumping around), but do use a focus colour. 
For the most part, I think I may have used 4 different colours depending on where the button resides, but it's mostly grey, white, black or my highlight blue (or a darker tint thereof). Which means I can only really standardise the UI background colours for the toolbar, the primary and secondary background colour for main UI elements.
I have seven main UI colours actually, based it on Blender's:
afbeelding

adrianinsaval:
> Some users like some spacing in their menu buttons, but some want more buttons on the screen. Although the last one could be just getting used too issues. You can already just make the icons bigger which would solve the issue too.

This is simply not true. Padding is an important part of UIs too. Making the icons bigger does not have any effect on this problem. First off because this is not just a problem with the toolbar icons and secondly because even if the icons are bigger they would still have no spacing between them, they will just each take up more space.

adrianinsaval:
Also, shipping as many stylesheets as we do right now with no theme for them seems useless. I suggest we aim to get one very polished dark stylesheet with a corresponding light stylesheet, assign those to the corresponding theme and then relegate all the other stylesheets to addons, I don't think we can truly maintain so many stylesheets with actual quality. However we probably should not be discussing this on a PR, what would is the proper issue for this?

adrianinsaval:
a IMO big problem with this implementation is that choosing a theme overwrites any customization you might have made to sketcher or background colors and similar.

Also, I had mentioned this before but calling "Classic" to a theme that does not set a stylesheet only really fits well on windows where Qt uses the old looking Fusion theme. On Linux having no stylesheet means freecad inherits whatever is set at the system level which could be something that looks a heck of a lot more modern. And on macOS IIRC it uses a theme that still looks cohesive with the rest of the system.

MisterMakerNL:
> Perhaps it would be wise to model this after Microsoft's design. Windows 10 (and 11) let the user set an 'accent' color (and calculates some other shades of it for a total of 6)
> 
>     1. SystemAccentColorLight3
> 
>     2. SystemAccentColorLight2
> 
>     3. SystemAccentColorLight1
> 
>     4. SystemAccentColorDark1
> 
>     5. SystemAccentColorDark2
> 
>     6. SystemAccentColorDark3
> 
> 
> Information about it can be found over here: See https://learn.microsoft.com/en-us/windows/apps/design/style/color#accent-color It's probably a good idea to use it as the starting values on Windows as well.
> 
> How to get the accent color in C++ can be found over here: https://github.com/res2k/Windows10Colors

This is interesting, use system accent color would be cool!
Have too note these aren't 6 different colors but the same color but for 6 different themes in different tints. So windows just uses 1 highlight/accent color.

MisterMakerNL:
@adrianinsaval 
This pr would reduce the amount if stylesheets too 5.
I currently have not added sketcher colors to the cfg file so it doesn't overwrite these. Only the actual GUI stuff. Although if Obelisk manage to come up with good contrast colors for sketcher we probably should put those in.

PaddleStroke:
@adrianinsaval my plan is to remove the stylesheet combobox alltogether as soon as we 'moved' stylesheets from stylesheet to themes. ie create theme for each stylesheet.
In parallel we are reducing the number of stylesheet by adding this 'accent colors' to avoid having -orange -pink ...
Yes themes would override a lot of settings : view color, sketcher colors, console text color and so on. And it's the whole point of those themes : be able to give the user a good looking and cohesive ui out of the boxe. Some users (but not so many I'd guess) are already customizing a lot of colors such as sketcher colors and so on. But for those people they (probably) already use the preference packs to save their settings so they can just make a custom pref pack or apply theme then their pref packs. Anyway I think for the vast majority of users, having good looking themes that define all the system colors coherently would make sense.

Regarding the 'classic' theme, we can rename it 'System' perhaps ?

PaddleStroke:
@Axeia 'Accent colors' names make sense imo. As it's the idea here, to give a accent color to a theme. Let's use this name and make 3 accents colors for theme developper. I'll also check if the default can use windows' accent color for win10 users.

MisterMakerNL:
Accent colors sounds fancy. Lets do that :)

obelisk79:
This shift is likely not going to make everyone happy, but I generally agree, packaged themes with less settings crammed into preferences overall is probably a good thing. If the settings are still exposed in the properties editor, the most hardcore customizers can still have access to change and tweak.

Offering coherent themes makes more sense, and they should be scoped/limited to a simple light/dark with the option to add new ones to the list using the addon-manager.

Regarding the accent colors, I think this is a good idea and should move forward. The concept of algorithmically deriving variations of a single accent is interesting, but probably better off being explored after the initial implementation and tracked through a separate github issue after a merge.

obelisk79:
Regarding the new light/dark themes I'd like to offer some concepts applying my recent color research. I'd like to get something initial put together this weekend.

MisterMakerNL:
[Stylesheets-var.zip](https://github.com/FreeCAD/FreeCAD/files/12338011/Stylesheets-var.zip)

PaddleStroke:
Thanks. I'm on it, I got the variable to work already. Should be ready
tomorrow morning probably.

adrianinsaval:
> Regarding the 'classic' theme, we can rename it 'System' perhaps ?

yes that seems better

PR #10017, opened 2023-07-30
[Notif area] Fixes and disable for 0.21

PaddleStroke:
This does : 
- Fix a ui bug, if area disabled when you open the pref page, the rest of the parameters are not disabled. 
- remove unecessary qmessagebox
- Disable area by default for 0.21 bacause it is not polished enough yet.

I have no strong feeling for the disabling, someone told me that it's not polished enough yet to the point of being very annoying. So I suggest disabling it for the 0.21 release and re-enable it in next development cycle to enable polishing it.

---

### DIFF
--- a/src/Gui/DlgSettingsNotificationArea.cpp
+++ b/src/Gui/DlgSettingsNotificationArea.cpp
@@ -41,35 +41,24 @@ DlgSettingsNotificationArea::DlgSettingsNotificationArea(QWidget* parent)
     ui->setupUi(this);
 
     connect(ui->NotificationAreaEnabled, &QCheckBox::stateChanged, [this](int state) {
-        if (state == Qt::CheckState::Checked) {
-            ui->NonIntrusiveNotificationsEnabled->setEnabled(true);
-            ui->maxDuration->setEnabled(true);
-            ui->maxDuration->setEnabled(true);
-            ui->minDuration->setEnabled(true);
-            ui->maxNotifications->setEnabled(true);
-            ui->maxWidgetMessages->setEnabled(true);
-            ui->autoRemoveUserNotifications->setEnabled(true);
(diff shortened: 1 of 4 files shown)


sliptonic:
I concur.  The youtube video by Mango Jelly (Linked at the end of the release notes) praises the feature.  Many people like it because it permits turning off the report view and not having it repeatedly re-enabled. 
Other people dislike it and find it overly intrusive.  
I use i3 window manager and it doesn't behave well there.  
I have no strong feelings about enabling or disabling for 0.21.  All things being equal, I agree with @PaddleStroke and think it should be off for 0.21 and then re-enabled for further polishing.

PaddleStroke:
To the maintainer, if you decide not to disable it, the 3 first commits are still worth making it.

PaddleStroke:
Just clicked on the wrong button

FEA-eng:
I definitely agree that the new "notification area" feature still needs some polish and should be deactivated by default for now. I disabled it in weekly builds because I found it to be actually more intrusive than the standard notification system.

chennes:
Thanks for the PR. I don't mind the notification area, and as a developer it does make me consider more carefully that sorts of things I make "error" vs. "warning" vs. "message", which I think is a good thing in the long term. That said, I am OK with setting it to default to off in the 0.21 release. It feels a bit "untidy" to me to have a PR to disable it, and then basically another that immediately re-enables it. Maybe it would make more sense to take that commit and only apply it to the release branch (once it's been created) -- that way it's explicit that the change is only intended to apply to the release, and not to the development (master) branch. @sliptonic, @adrianinsaval, what do you think of that strategy?

adrianinsaval:
Sounds good!

chennes:
OK, the first three commits have been cherry-picked and pushed to master.

chennes:
I've created releases/0.21 and cherry-picked the last commit into it. Closing now.

PR #10024, opened 2023-07-31
Rename cmake flag for unit tests to BUILD_UNIT_TESTS

adrianinsaval:
To differentiate from the previously existing BUILD_TEST that controls building the Test WB. See https://forum.freecad.org/viewtopic.php?p=695667#p695667

@chennes 

---


chennes:
Rather than "UNIT_TEST" I'd suggest something less likely to generate bickering about what that phrase means -- maybe "ENABLE_GOOGLE_TESTS" or "ENABLE_GTEST"?

adrianinsaval:
I'm happy with anything so long as it's different from the wb! I was under the impression that gtest wasn't the only supported system so I went with what seemed more generic. I recall @wwmayer speaking of Qt tests and I think I've seen something about that in the code. Give me a name and I change it.

chennes:
OK, let's keep it super generic then: "ENABLE_DEVELOPER_TESTS"

wwmayer:
I support the idea to use a different CMake option to control the C++ unit tests. The name ENABLE_DEVELOPER_TESTS is OK.

When merging the PR make sure to use the "squash" option so that we have a single commit instead of five.

wwmayer:
> I only see the one commit (with five files changed).

Oops! You're right.

adrianinsaval:
so is it ok to merge then?

chennes:
Yes, sorry, I meant to do that earlier today and it fell off my radar. Merged now.

adrianinsaval:
please also cherry pick on release branch

PR #10064, opened 2023-08-04
add cstdint behind ifdef __GNUC__ for compiling under alpine linux

morganrallen:
note: in other headers there is a mix of including #ifdef __GNUC__

Sadly I did not catch this during the RC window but a patch in the aport can make this work for Alpine specifically, cannot say if this affects other systems.

### DIFF
--- a/src/Base/Builder3D.h
+++ b/src/Base/Builder3D.h
@@ -26,6 +26,10 @@
 
 // Std. configurations
 
+#ifdef __GNUC__
+# include <cstdint>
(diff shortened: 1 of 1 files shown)


chennes:
I don't have Alpine Linux to test this on, but if it's fixing a compilation issue for you, it matches what we're doing in many other places so seems fine to me.

morganrallen:
Yup, compiling went fine and I've got a PR into Alpine/aports

PR #10065, opened 2023-08-04
[TechDraw] User-defined presets for editable fields

benj5378:

---


src/Mod/TechDraw/Gui/DlgPrefsTechDrawTemplate.h
```diff
@@ -0,0 +1,53 @@
...
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+#ifndef GUI_DIALOG_DLGSETTINGSFONT
+#define GUI_DIALOG_DLGSETTINGSFONT
+
```
WandererFan, line 25: #include <memory> for std::unique_ptr

WandererFan:
Would prefer a solution where the client code does not need to know how to unpack and process the packed preferences.    There is ParameterGrp::GetASCIIs that returns all the string parameters in a group as a vector of strings. 

Can we make it so that prefTableWidget makes a bunch of key-value pairs instead of packing everything into a single string?

WandererFan:
Lets take the template changes out.   That should be a separate PR.

We should also build this PR on top of current master instead of being dependent on changing the template properties.

benj5378:
> Would prefer a solution where the client code does not need to know how to unpack and process the packed preferences. There is ParameterGrp::GetASCIIs that returns all the string parameters in a group as a vector of strings.
>
> Can we make it so that prefTableWidget makes a bunch of key-value pairs instead of packing everything into a single string?

Not really, because it might not be all implementations of prefTableWidget that has key-value pairs. I don't see how `ParameterGrp::GetASCIIs` would be more useful than a JSON array for instance.

> Would prefer a solution where the client code does not need to know how to unpack and process the packed preferences. 

I don't see anyway around that. Currently, the fillEditableFields need to know, that it is dealing with an array, that packs arrays of "key and value pairs". If we changed it to `ParameterGrp::GetASCIIs` for instance, then it would still be an array of key and value pairs; we would still need to know how to unpack it.

I don't see any format, where the unpacking process isn't needed to be known? No matter the format (JSON, XML etc.), we still need to know that there is a list of some sort (array) in the preference XML that has a name of an editable field together with it's a preference value (key-value pairs).

Do you have any suggestions, a hypothetical for instance?

If I should point to an issue with the format I am trying to introduce, then it's the human readability. Unless reading through how the values are packed/unpacked, one cannot manually make a preference value. Though it should be easily discoverable if one just runs the code once and examines the preference XML output.

Another issue is saving as ASCII. If we save JSON, we should really introduce a new datatype called setJSON, getJSON... but that's mainly naming than functionality

WandererFan:
Take a look at https://github.com/WandererFan/FreeCAD/tree/TemplateFields3

benj5378:
@WandererFan Okay, so it's easier to read, but we still need to know how the key values are packed: it's a preference group, where each preference entry is the editableKey and the value is the editableValue.

Issue it, that we can't use this format across all prefTableWidgets, and we also want to avoid using specific implementations.

What about saving like diz?

[
{"editableField": "Organization", "editableValue": "Joe's button factory"},
{"editableField": "Organization", "editableValue": "Joe's button factory"},

]

WandererFan:
https://github.com/WandererFan/FreeCAD/tree/TemplateFields4  uses a standard QTableWidget instead of PrefTableWidget.

Retrieval uses standard preference functions that are used everywhere in FC.

benj5378:
> uses a standard QTableWidget instead of PrefTableWidget.

I see it worse using QTableWidget over prefTableWidget; that would mean we would have to make implement it three different locations (multiple PRs need a tableWidget). If it's because you don't want to use Qt JSON framework, then we should change how prefTableWidget works rather than shotgun implementing QTableWidgets multiple locations.

> I don't really know what is the best solution. While your suggestion would only work for tables with two columns (and is simpler!), then when would we ever need more than two columns? So far, all my 2-3 implementations of prefTableWidget is 2 columns only. But rewriting it further down the road would also be sad.

Once again, it comes down to this. We could change prefTableWidget to work as you suggest but on the cost of might having to change it down the road.

> Retrieval uses standard preference functions that are used everywhere in FC.

I agree your suggestion using preference functions simplify things and I'm a supporter of that. But I don't get how using preference functions are better than Qt JSON framework just because they are used other locations.

> I don't really know what is the best solution. While your suggestion would only work for tables with two columns (and is simpler!), then when would we ever need more than two columns? So far, all my 2-3 implementations of prefTableWidget is 2 columns only. But rewriting it further down the road would also be sad.

So should we do this?

WandererFan:
Implementing PrefTableWidget isn't the goal here, though.  The goal is to create a retrievable list of key-value (fieldName+defaultValue) pairs and apply them to a template.

Consistency across the application is why I think the existing functions are a better choice.

WandererFan:
This PR has been OBE.  See Add-on Manager TitleBlockWB and Techdraw_FillTemplate.  Closing.

PR #10066, opened 2023-08-04
Bump build version to 0.22 and make it obvious it's a developer build

chennes:
This PR sets the version to 0.22.0dev. It then ensures that the "dev" suffix shows in the places where we show the version (including the title bar), and adds a prominent (translatable) warning label to the splashscreen.

Screenshot 2023-08-03 at 10 57 31 PM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ project(FreeCAD)
 
 set(PACKAGE_VERSION_NAME "Vulcan")
 set(PACKAGE_VERSION_MAJOR "0")
-set(PACKAGE_VERSION_MINOR "21")
+set(PACKAGE_VERSION_MINOR "22")
 set(PACKAGE_VERSION_PATCH "0") # number of patch release (e.g. "4" for the 0.18.4 release)
 set(PACKAGE_VERSION_SUFFIX "dev") # either "dev" for development snapshot or "" (empty string)
 set(PACKAGE_BUILD_VERSION "0") # used when the same FreeCAD version will be re-released (for example using an updated LibPack)
--- a/src/App/Application.cpp
+++ b/src/App/Application.cpp
@@ -2478,6 +2478,7 @@ void Application::initConfig(int argc, char ** argv)
         App::Application::Config()["BuildVersionMajor"  ] = FCVersionMajor;
         App::Application::Config()["BuildVersionMinor"  ] = FCVersionMinor;
         App::Application::Config()["BuildVersionPoint"  ] = FCVersionPoint;
+        App::Application::Config()["BuildVersionSuffix" ] = FCVersionSuffix;
         App::Application::Config()["BuildRevision"      ] = FCRevision;
         App::Application::Config()["BuildRepositoryURL" ] = FCRepositoryURL;
         App::Application::Config()["BuildRevisionDate"  ] = FCRevisionDate;
@@ -2557,21 +2558,23 @@ void Application::initConfig(int argc, char ** argv)
         // Remove banner if FreeCAD is invoked via the -c command as regular
         // Python interpreter
         if (!(mConfig["Verbose"] == "Strict"))
-            Base::Console().Message("%s %s, Libs: %s.%s.%sR%s\n%s",
+            Base::Console().Message("%s %s, Libs: %s.%s.%s%sR%s\n%s",
                               mConfig["ExeName"].c_str(),
                               mConfig["ExeVersion"].c_str(),
                               mConfig["BuildVersionMajor"].c_str(),
                               mConfig["BuildVersionMinor"].c_str(),
                               mConfig["BuildVersionPoint"].c_str(),
+                              mConfig["BuildVersionSuffix"].c_str(),
                               mConfig["BuildRevision"].c_str(),
                               mConfig["CopyrightInfo"].c_str());
         else
-            Base::Console().Message("%s %s, Libs: %s.%s.%sR%s\n",
+            Base::Console().Message("%s %s, Libs: %s.%s.%s%sR%s\n",
                               mConfig["ExeName"].c_str(),
                               mConfig["ExeVersion"].c_str(),
                               mConfig["BuildVersionMajor"].c_str(),
                               mConfig["BuildVersionMinor"].c_str(),
                               mConfig["BuildVersionPoint"].c_str(),
+                              mConfig["BuildVersionSuffix"].c_str(),
                               mConfig["BuildRevision"].c_str());
     }
     LoadParameters();
--- a/src/Build/Version.h.cmake
+++ b/src/Build/Version.h.cmake
@@ -1,9 +1,10 @@
 
 // Version Number
-#define FCVersionMajor "${PACKAGE_VERSION_MAJOR}"
-#define FCVersionMinor "${PACKAGE_VERSION_MINOR}"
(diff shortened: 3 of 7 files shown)


yorikvanhavre:
Looks good to me! I imagine users quickly asking for a way to force-turn this off, though.. but we can introduce a user parameter for that later on

sliptonic:
> Looks good to me! I imagine users quickly asking for a way to force-turn this off, though.. but we can introduce a user parameter for that later on

The correct answer to that request would be....'no'

chennes:
> > Looks good to me! I imagine users quickly asking for a way to force-turn this off, though.. but we can introduce a user parameter for that later on
> 
> 
> 
> The correct answer to that request would be....'no'
> 
> 

Technically they can turn it off by just not showing the splash screen. Or, you know, running the release version ;).

benj5378:
Honestly, warning message does not look visually good. Please place this warning somewhere else; I'd suggest the start page or a picture docked to the bottom, like the Python console.

sliptonic:
> Honestly, the warning message does not look visually good. Please place this warning somewhere else; I'd suggest the start page or a picture docked to the bottom, like the Python console.

That's exactly the point.  It should be big and ugly and in your face

benj5378:
> That's exactly the point. It should be big and ugly and in your face

Things can be clear and obvious without being ugly as I just came with suggestions to.

sliptonic:
I hear you  and understand but we're trying to change culture.  Historically we've done all kinds of things that make the development build attractive to users.  As a result users are attracted to the development build.  We're trying to change that.

chennes:
I went with the US Surgeon General's Warning style here, prioritizing legibility and contrast, as well as leveraging the cultural norm that warning has created. See, e.g.
e_cig_jpg

adrianinsaval:
I agree with having a warning but it's not something so critical that it needs to look that bad and in your face. We want people to be more aware that this is a dev version but we don't really want people to stop testing them!

chennes:
Let me see if I can work up a compromise solution...

adrianinsaval:
Something like this maybe?
image

chennes:
How about this? (minimum code change 😄)
Screenshot 2023-08-04 at 12 45 33 PM

sliptonic:
> @sliptonic It doesn't sound like you are reading what I'm writing. I am not writing it shouldn't be clear that it's a dev build or that it has risks, I just wrote it shouldn't be ugly. I came with a suggestion on how to make it clear without making it ugly.

I understood and agree that it should be clear.   No argument from me on that point.

> So no, it doesn't sound like you understand me - unless you mean you want to make users stop using dev version by literally making it too ugly for usage, rather than rationally explaining to the user, why they shouldn't use the dev version. 

I certainly wouldn't argue to change existing functionality to make it uglier.  That is silly.  My point is that we shouldn't take effort to make the warning more attractive.  The warning will come off when it's done.  Until then it's meant to be in your face and meant to be jarring.  What is the value of making it more attractive unless it is to "attract" users?

sliptonic:
There's a very real psychological effect called habituation or sensory adaptation.  The user will see something so often that they don't see it or notice it anymore.   Making things color-complementary, symmetric, proportional will speed the habituation.  Placing the warning at the bottom will not be effective since that spot is less important.  

Making it obscure the main graphic is more effective at avoiding habituation.  Putting it off center as original suggested is good too.  It's slightly annoying and hard to get used to.

benj5378:
> I agree with having a warning but it's not something so critical that it needs to look that bad and in your face. We want people to be more aware that this is a dev version but we don't really want people to stop testing them!

This also reflects what I am trying to point out. The goal is to inform people and avoid them from hurting themselves unpurposely using a dev version, when it's really not a good idea for their purposes. The goal is not to make the dev version unattractive for the sole goal of stopping people from using the dev version - because then, we could just have psychedelic colors and random jump scares inside FreeCAD.

> My point is that we shouldn't take effort to make the warning more attractive. The warning will come off when it's done. Until then it's meant to be in your face and meant to be jarring. What is the value of making it more attractive unless it is to "attract" users?

There is nothing wrong with a warning being attractive, if it fulfills it's goal: to warn users against using a dev version, where it's not appropriate and would hurt themselves. I want to do that through enlightment, where a user based on the information can choose whether they want to take that risk or not. If the user wants to, then okay fine! it's their choice, there is no reason to push people out by having ugly UI, that is like protecting people from themselves - and has nothing to do with the goal, avoiding people from hurting themselves unpurposely.

> There's a very real psychological effect called habituation or sensory adaptation. The user will see something so often that they don't see it or notice it anymore. Making things color-complementary, symmetric, proportional will speed the habituation.
>
> Making it obscure the main graphic is more effective at avoiding habituation. Putting it off center as original suggested is good too. It's slightly annoying and hard to get used to.

And what is wrong with that? The user has chosen - based on the warning they saw the first time - to use the dev version despite it's risks. If they want to take that choice, then again, fine. I find constantly reminding people in a obscure way on every startup above the goal we are trying to achieve. I still don't believe that to be reminded, things have to be ugly.

> As a result users are attracted to the development build. We're trying to change that.

I'm unaware if I just misunderstood you on your first couple of posts, because I read it like you wanted to have stuff ugly (unattractive) to keep people from using a dev version. In the new posts I read it as to have stuff ugly to keep reminding people it's a dev version.

We could probably go far with discussing this on something trivial (that I started, yes), but I can only say that I disagree with making it ugly and find it off goal that we are trying to fulfill. So no, I don't think any of that is good.

> image

I'm happy with this in terms of "beautiness". Though I want to elaborate, that if the goal is to make people aware of the risks using the dev version, then information on why it shouldn't be used in a production environment would be relevant. To avoid people from hurting themselves, I think that (an explanation) would be more effective than a warning where people don't understand it's relevance - why is it risky. What do you think about also appending "development version" to the window title bar?

Once again, I'm fine with merging this, but
- If the goal is to inform users why they might now want to use a dev version (that it has risks that could mess up one's model), then I'm not sure we are achieving this. Maybe it does. If not, add information on what the risks are.
- If the goal solely is to let people know it's a dev version, I feel that this PR now fulfills it.
- If the goal solely is to actively remind people it's a dev version, add the unhabituation effects.
- If the goal solely is to stop people from using the dev version, making it unattractive by adding the psychedelic colors and jumpscares.

Let's remember the goal behind making a warning in the first place.

chennes:
Truth be told, what I'm really trying to do is discourage our advanced users from telling newbies that they should use the weeklies. They don't really needer want an explanation, and probably wouldn't read one if I offered it 😄. This is just a gentle reminder that the master branch is going to start moving quite a lot faster, and there is some risk with that.

benj5378:
> This is just a gentle reminder that the master branch is going to start moving quite a lot faster, and there is some risk with that.

If it's just to make sure people have been warned and that dev version have a risk. Not trying to avoid people from hurting themselves (I don't say we need to), then yes, PR fulfills it. ;)

sliptonic:
I'm fine with it as is.  Though the idea of random jump scares in the app sounds pretty cool!

chennes:
OK, sounds like a rough consensus to me, and I want to get that version number incremented ASAP. I'm going to merge, but feel free to continue to discuss, we can always make more changes later.

benj5378:
image
@chennes

chennes:
Thanks, I'll add a check for that and drop the font size as needed.

Issue #10082, opened 2023-08-05
[Backport] Regex speed improvement on QuantitySpinBox


chennes:
### Original PR

https://github.com/FreeCAD/FreeCAD/pull/10067

### Description and Justification

Speed improvement to QuantitySpinBox when using large numbers of digits.

adrianinsaval:
@chennes was this ever backported?

chennes:
No, I don't think so.

Roy-043:
Backported.

PR #10067, opened 2023-08-04
[GUI] QuantitySpinBox: Regex optimization.

ConnorWorrell:
Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [x]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---
Addressing issue https://github.com/FreeCAD/FreeCAD/issues/9864, regex performance improvements.

PR #10110, opened 2023-08-08
[GUI] QuantitySpinBox Regex Optimization #10067 Ammendum

ConnorWorrell:

---

Addendum to #10067, A bug in line 210 where the unitStr variable didn't get reversed when line 213 was added, causing the appended string to be *(1tinu) instead of the correct *(1unit). 

https://github.com/FreeCAD/FreeCAD/issues/10082

### DIFF
--- a/src/Gui/QuantitySpinBox.cpp
+++ b/src/Gui/QuantitySpinBox.cpp
@@ -170,6 +170,9 @@ public:
         if (locale.positiveSign() != plus)
             copy.replace(locale.positiveSign(), plus);
 
+        QString reverseUnitStr = unitStr;
+        std::reverse(reverseUnitStr.begin(), reverseUnitStr.end());
+
         //Prep for expression parser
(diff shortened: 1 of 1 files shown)


chennes:
Absent developer tests it's pretty hard to evaluate the regex code: can you give me an input to give the box that demonstrates the failure?

ConnorWorrell:
Certainly!

Steps to reproduce:
1. Set unit system to standard.
1. Part workbench > insert primitive cube
2. Set length of one dimension to `1um` (This changes the default unit from mm to um)
3. Set length to `3`.
4. Length should update to `3.00 µm`, but does not.

chennes:
Sorry for the delay on this: I wanted to the the developer test framework in place for testing this class automatically. This business of tweaking and testing by hand is pretty fragile. I've added a couple of demonstration tests to tests/src/gui/QuantitySpinBox.cpp, if you'd like to see where this is going.

adrianinsaval:
This should also be backported I think

PR #10146, opened 2023-08-09
App: Fix crash and undefined behavior in StringHasherPy and StringIDPy

marioalexis84:
The `StringHasherPy::getID` function crash if the first argument is an integer less or equal to 0.
If the first argument is a string, the signature is changed to explicitly receive a boolean in the optional second argument.

In `StringHasherPy::isSame` and `StringID::isSame` currently returns a result with an exception set if the argument isn't `App.StringHasherPy` (`App.StringID`). This is undefined behavior and must be avoided.

Other changes:
- In `StringHasherPy::PyInit` explicitly avoid parameters.
- In `StringHasherPy::getCount` and `StringHasherPy::getSize` use `PyLong_FromSize_t` to get the `PyLongObject` from `size_t`.
- Change PyCXX `Py::Int` to `Py::Long`.

The second commit change indentation to four spaces.

---

### DIFF
--- a/src/App/Application.cpp
+++ b/src/App/Application.cpp
@@ -114,6 +114,8 @@
 #include "PropertyFile.h"
 #include "PropertyLinks.h"
 #include "PropertyPythonObject.h"
+#include "StringHasherPy.h"
+#include "StringIDPy.h"
 #include "TextDocument.h"
 #include "Transactions.h"
 #include "VRMLObject.h"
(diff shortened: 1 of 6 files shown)


chennes:
Thanks for the reviewing the code, and for submitting this PR! We haven't reviewed the Python TNP interface at all yet, so your looking at it is much appreciated. Can you add tests that demonstrate the problems in `getID` and `isSame`?

marioalexis84:
Of course, I can add a unit test.

PR #10151, opened 2023-08-10
Test: Begin QuantitySpinBox QTests

chennes:
This is a first try at adding tests for QuantitySpinBox, focused on getting the framework in place. The few tests are just proofs-of-concept. In the long run we should really exercise the regular expressions being run behind the scenes when validating this input.

@wwmayer I don't think I understand correctly how you planned on structuring the QTest tests -- I added this one the same as the InventorBuilder test, but it requires a fair bit of CMake boilerplate. Is there a better way of structuring this that doesn't require all that input for every test class? Do you have an idea for how to structure these (since in an ideal world there will be hundred of test classes).


wwmayer:
> I don't think I understand correctly how you planned on structuring the QTest tests

At the time when adding the Qt test framework it wasn't clear how to structure all the unit tests. That's why I have added to a separate directory.

But I think the Qt test files should go to the same directories as the Google test files, i.e. not having a separate directory structure. Then the goal should be to make it as easy as possible to choose between the two test frameworks.

So, this means that InventorBuilder.cpp should be moved from Qt to Base and the CMake function SETUP_TESTS to the tests/CMakeLists.txt file.

chennes:
@wwmayer I've rearrange things a bit, then -- is this what you had in mind?

berniev:
Minor niggle: In cmake `Typically functions use all-lowercase names` [doc]

You use CLion, check out cmake formatting - real easy.

wwmayer:
> is this what you had in mind?

Yes.

chennes:
> You use CLion, check out cmake formatting - real easy.

It doesn't rename pasted functions -- I'm guessing @wwmayer won't object to the change, though, so I'll update the PR.

wwmayer:
Several months ago I watched a video how to activate the Qt testing framework in CMake and for convenience this function in upper case was provided.

At https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html I found this rule:
> Use lowercase function names: CMake functions and macros can be called lower or upper case. Always use lower case. Upper case is for variables.

So, it's fine to change the function name.

berniev:
> It doesn't rename pasted functions

Likely your setup needs tweaking. Works fine for me.

Issue #10136, opened 2023-08-09
[Problem] Regressions with the new theme handling


wwmayer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
OS: Ubuntu 22.04.3 LTS (XFCE/xubuntu)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.33756 (Git)
Build type: debug
Branch: master
Hash: d150fa7164e07651b503e2fd16e6eb55a9849a77
Python 3.10.12, Qt 5.15.3, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.1
Locale: German/Germany (de_DE)
```

### Subproject(s) affected?

Core

### Problem description

There are several issues with the new theme handling:
* Illogical user interface: In the General tab there is  a setting to choose between Classic, Dark and Light themes. and in the Display section there is a separate Theme tab. For the user it's not obvious how they are related -- especially as they are can be combined.
* Whenever switching between Classic, Dark or Light theme the report view will become invisible
* The Classic, Dark or Light theme modifies the background color of the 3d view. If after switching to another theme and adjusting the background colors under Display > Colors then the colors won't be accepted and rejected to the default colors of the theme. The only way to get the custom colors accepted is to close the the preferences dialog after changing the theme and reopening it again. This is extremely annoying behaviour.
* Whenever choosing a style sheet in the Theme tab a warning "Hello" is written to the report view

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chennes:
@PaddleStroke

PaddleStroke:
- ui : The Theme page has a text to explain it's purpose 'This page let you customize your current theme. ...' I think the situation is now confusing because we still have the stylesheet combobox in there and this stylesheet combobox has been acting as 'kind-of-theme' for a long time. Besides the color pickers does nothing currently so that doesn't help.

So the general idea is that general page will have the Theme combobox to make theme selection proiminent. And the theme page will give the user theme-customization settings. Maybe the theme page can also have the theme combobox to make things clearer?

- report view hidding : I think this is because hidding the report view is included in the themes (they are preference packs). @MisterMakerNL can you confirm this?

- Indeed this is annoying. It's because once you changed the theme once, themeChanged bool is true so every-time you click on apply again, it will apply the theme again overriding the background color change. I will make a fix for this

- This is a coding leftover I'll remove it.

PaddleStroke:
[...]

PR #10153, opened 2023-08-10
Pref: Theme: Solves #10136

PaddleStroke:
Solves #10136

> - The Classic, Dark or Light theme modifies the background color of the 3d view. If after switching to another theme and adjusting the background colors under Display > Colors then the colors won't be accepted and rejected to the default colors of the theme. The only way to get the custom colors accepted is to close the the preferences dialog after changing the theme and reopening it again. This is extremely annoying behaviour.
> - Whenever choosing a style sheet in the Theme tab a warning "Hello" is written to the report view

---

### DIFF
--- a/src/Gui/PreferencePages/DlgSettingsGeneral.cpp
+++ b/src/Gui/PreferencePages/DlgSettingsGeneral.cpp
@@ -377,6 +377,8 @@ void DlgSettingsGeneral::saveThemes()
     QString sheet = QString::fromStdString(hGrp->GetASCII("StyleSheet"));
     bool tiledBackground = hGrp->GetBool("TiledBackground", false);
     Application::Instance->setStyleSheet(sheet, tiledBackground);
+
+    themeChanged = false;
 }
 
 void DlgSettingsGeneral::loadThemes()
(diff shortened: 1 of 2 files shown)


wwmayer:
The second point is fixed but the first point only partially.

1. Open preferences and click on the Theme combo box. You don't have to switch to another theme
2. Do NOT press the Apply button
3. Go to the Colors tab and change the background color
4. Press the Apply button
5. Result: your changes on the background color are overwritten

The colors are not overwritten if in step 2. the user has pressed the Apply button.

PaddleStroke:
I have added a commit to take care of the case where you click the combobox but don't actually change the theme.

If you have any idea how to handle to case where both theme and view color are changed at the same time let me know.

wwmayer:
> But this is kind of expected as the change of theme changes the 3Dview color so there must be a conflict if you apply both settings at the same time.

For a developer who knows the code it is expected. But for a normal user it's not because when he changes the theme it's not necessarily expected that this will affect the background color.

> But is that a real problem ?

No, it's only a theoretical problem now. With your latest changes I think it's fine and probably nobody will realize it.

chennes:
At the least this is an improvement on the current state: good to merge, or do you plan more work on it, @PaddleStroke ?

PaddleStroke:
No more things planed here.

sliptonic:
Does this actually close 10136 then?

PaddleStroke:
Hmm there's 

> Whenever switching between Classic, Dark or Light theme the report view will become invisible

that is not solved yet. This is because hidding the report view is apparantly included in the pref packs. Which it probably shouldn't.

PR #10168, opened 2023-08-11
Disable Robot, Inspection, Reverse engineering, OpenSCAD by default.

PaddleStroke:
Second step to solve #8692 by disabling some seldomly used workbenches : 
- Robot : Seldomly used
- Inspection : Still very early stage and probably seldomly used. Also it appears this workbench has a single command. It could probably be best in core than in a separate wb...
- Reverse Engineering: wiki says : "At the moment there is no functionality present in this workbench. It is used as a sandbox by the programmers." Though it seems to actually have a single tool that is mesh related.
- OpenSCAD : This requires an external installation to work so there's no reason to be enabled by default.

---

### DIFF
--- a/src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
+++ b/src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
@@ -424,7 +424,7 @@ QStringList DlgSettingsWorkbenchesImp::getDisabledWorkbenches()
     ParameterGrp::handle hGrp;
 
     hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Workbenches");
-    disabled_wbs = QString::fromStdString(hGrp->GetASCII("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench"));
+    disabled_wbs = QString::fromStdString(hGrp->GetASCII("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench,OpenSCADWorkbench"));
 #if QT_VERSION >= QT_VERSION_CHECK(5,15,0)
     unfiltered_disabled_wbs_list = disabled_wbs.split(QLatin1String(","), Qt::SkipEmptyParts);
 #else
--- a/src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
+++ b/src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
@@ -424,7 +424,7 @@ QStringList DlgSettingsWorkbenchesImp::getDisabledWorkbenches()
     ParameterGrp::handle hGrp;
 
     hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Workbenches");
-    disabled_wbs = QString::fromStdString(hGrp->GetASCII("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench"));
+    disabled_wbs = QString::fromStdString(hGrp->GetASCII("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench"));
 #if QT_VERSION >= QT_VERSION_CHECK(5,15,0)
     unfiltered_disabled_wbs_list = disabled_wbs.split(QLatin1String(","), Qt::SkipEmptyParts);
 #else
--- a/src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
+++ b/src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
@@ -424,7 +424,7 @@ QStringList DlgSettingsWorkbenchesImp::getDisabledWorkbenches()
     ParameterGrp::handle hGrp;
 
     hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Workbenches");
-    disabled_wbs = QString::fromStdString(hGrp->GetASCII("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench"));
+    disabled_wbs = QString::fromStdString(hGrp->GetASCII("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench"));
 #if QT_VERSION >= QT_VERSION_CHECK(5,15,0)
     unfiltered_disabled_wbs_list = disabled_wbs.split(QLatin1String(","), Qt::SkipEmptyParts);
 #else
--- a/src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
+++ b/src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
@@ -424,7 +424,7 @@ QStringList DlgSettingsWorkbenchesImp::getDisabledWorkbenches()
     ParameterGrp::handle hGrp;
 
     hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Workbenches");
-    disabled_wbs = QString::fromStdString(hGrp->GetASCII("Disabled", "NoneWorkbench,TestWorkbench"));
+    disabled_wbs = QString::fromStdString(hGrp->GetASCII("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench"));
 #if QT_VERSION >= QT_VERSION_CHECK(5,15,0)
     unfiltered_disabled_wbs_list = disabled_wbs.split(QLatin1String(","), Qt::SkipEmptyParts);
 #else
--- a/src/Mod/AddonManager/Addon.py
+++ b/src/Mod/AddonManager/Addon.py
@@ -610,7 +610,7 @@ class Addon:
         wbName = self.get_workbench_name()
 
         # Add the wb to the list of disabled if it was not already
-        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench")
+        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench,OpenSCADWorkbench")
         # print(f"start disabling {disabled_wbs}")
         disabled_wbs_list = disabled_wbs.split(",")
         if not (wbName in disabled_wbs_list):
@@ -641,7 +641,7 @@ class Addon:
     def remove_from_disabled_wbs(self, wbName: str):
         pref = fci.ParamGet("User parameter:BaseApp/Preferences/Workbenches")
 
-        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench")
+        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench,OpenSCADWorkbench")
         # print(f"start enabling : {disabled_wbs}")
         disabled_wbs_list = disabled_wbs.split(",")
         disabled_wbs = ""
--- a/src/Mod/AddonManager/Addon.py
+++ b/src/Mod/AddonManager/Addon.py
@@ -610,7 +610,7 @@ class Addon:
         wbName = self.get_workbench_name()
 
         # Add the wb to the list of disabled if it was not already
-        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench")
+        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench")
         # print(f"start disabling {disabled_wbs}")
         disabled_wbs_list = disabled_wbs.split(",")
         if not (wbName in disabled_wbs_list):
@@ -641,7 +641,7 @@ class Addon:
     def remove_from_disabled_wbs(self, wbName: str):
         pref = fci.ParamGet("User parameter:BaseApp/Preferences/Workbenches")
 
-        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench")
+        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench,RobotWorkbench,ReverseEngineeringWorkbench")
         # print(f"start enabling : {disabled_wbs}")
         disabled_wbs_list = disabled_wbs.split(",")
         disabled_wbs = ""
--- a/src/Mod/AddonManager/Addon.py
+++ b/src/Mod/AddonManager/Addon.py
@@ -610,7 +610,7 @@ class Addon:
         wbName = self.get_workbench_name()
 
         # Add the wb to the list of disabled if it was not already
-        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench")
+        disabled_wbs = pref.GetString("Disabled", "NoneWorkbench,TestWorkbench,InspectionWorkbench")
         # print(f"start disabling {disabled_wbs}")
         disabled_wbs_list = disabled_wbs.split(",")
         if not (wbName in disabled_wbs_list):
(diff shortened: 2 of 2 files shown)


sasobadovinac:
I don't agree with this, Inspection and Reverse Engineering have some very useful tools, it would be better to think how to develop this WBs further instead of hiding them...

FEA-eng:
> I don't agree with this, Inspection and Reverse Engineering have some very useful tools, it would be better to think how to develop this WBs further instead of hiding them...

Reverse Engineering wb has literally single tool and there are no plans to develop it further at the moment. IMO it should be an add-on at this point (unlike many other much more commonly used and feature-rich workbenches that are still add-ons).

Inspection workbench also has just a single tool (no further development perspectives at this point) and is even less commonly used. Such niche workbench should be an add-on too.

If not degraded to add-ons, those workbenches should be disabled by default, in my opinion. Very few people use them and they are not some standard modules that should be included for everyone.

PaddleStroke:
Reverse engineering has a single tool that approximate a bspline from a point cloud. If this wb does not have development planed, maybe it's best to move this tool to the Mesh workbench and remove Reverse Engineering alltogether.

Same for inspection, it has a single tool that could be integrated into Part. 

Besides this PR does not remove them from FreeCAD, it only disable them by default. You can re-enable them in the preferences.

FEA-eng:
> Reverse engineering has a single tool that approximate a bspline from a point cloud. If this wb does not have development planed, maybe it's best to move this tool to the Mesh workbench and remove Reverse Engineering alltogether.
> 
> Same for inspection, it has a single tool that could be integrated into Part.
> 

Right, ultimately, it would be best to remove those workbenches and add their tools to more developed modules in which they fit. Mesh and Part make sense here. But disabling them by default is also reasonable. Makes the list of workbenches more clear and reduces confusion among new users.

sasobadovinac:
There are at least a dozen tolls in Reverse engineering IMO the problem is more like it was with mesh WB in the past, this tools need icons, get placed in the toolbars and get more awareness on how to use them.

The single tool in Inspection could maybe be moved to a different WB but on the other hand additional tools for doing analysis and measurements could be added to this WB.

Hiding this WBs is IMO the wrong solution, there should be more awareness about the available functionality in this tools and they should be easier to find and use. What is generally useful and what not is very subjective, someone working in architecture could say that all part, assembly, fem, path... basically almost everything should be removed because no one is using that in architecture, same could be for someone that is doing mostly cam, or someone that is doing mostly mechanisms, etc.

This ongoing simplification of FC by somehow dumbing it down and moving from an general engineering tool to a "toy" is indeed somewhat bothering me.

chennes:
I believe Reverse Engineering is @wwmayer ’s project, so he can weigh in on that.

PaddleStroke:
> This ongoing simplification of FC by somehow dumbing it down and moving from an general engineering tool to a "toy" is indeed somewhat bothering me.

Please be more specific in your criticism. What exactly is dumbing freecad down to a toy so far? Is it disabling the 'Test' workbench ? Or perhaps the glorious 'None' wb? Or perhaps simplifying the piece of art that was the sketcher constraint taskbox ? Or even maybe removing the Image wb and its so-intuitive ui?

This is just disabling wbs that are seldomly used. We're not talking about disabling FEM. We're talking about disabling Robot. How many users actually use Robot ? And for those extremely rare users I don't think re-enabling it in the preference is a problem.

Could inspection have more tools and be a awesome hub for all kind of measurement tools and so on? Maybe but right now it's not. And it doesn't feel like it's going to happen anytime soon. So in the meantime disabling it is the sensible thing to do.

For reverse engineering indeed I just saw that there are indeed many tools in the menu. The wiki is outdated.
However I would still argue that 'Reverse engineering' feels like a undefined category. If you look at the different tools, the logic cohesion between them is weak. It feels like some should be in mesh, some in Points ...

FEA-eng:
> There are at least a dozen tools in Reverse engineering IMO the problem is more like it was with mesh WB in the past, this tools need icons, get placed in the toolbars and get more awareness on how to use them.

Are they even working ? They are not documented on the wiki and not accessible from the toolbar for a few years so either someone should continue the development of that workbench or hide it for now.

> Hiding this WBs is IMO the wrong solution, there should be more awareness about the available functionality in this tools and they should be easier to find and use. What is generally useful and what not is very subjective, someone working in architecture could say that all part, assembly, fem, path... basically almost everything should be removed because no one is using that in architecture, same could be for someone that is doing mostly cam, or someone that is doing mostly mechanisms, etc.

There are some standards in the general-purpose CAD software industry. Such software typically provides modules for:
- part and assembly modeling,
- technical drawings,
- surface modeling,
- sheet metal part modeling,
- weldments modeling,
- fasteners library, 
- BIM (usually in a separate software though), 
- FEM, 
- CAM,
- kinematic analysis/motion simulation,
- mesh modeling (typically highly limited if present at all).

and not much more than that by default. The remaining functionalities are considered specialized tools/add-ons. FreeCAD doesn't provide many of those workbenches (especially assembly) by default but it provides some niche workbenches like the ones discussed here or the ones already gone from 0.21.

> 
> This ongoing simplification of FC by somehow dumbing it down and moving from an general engineering tool to a "toy" is indeed somewhat bothering me.

With general engineering CAD tools defined above, I don't think that commercial CAD software worth tens of thousands of dollars can be considered a toy because it doesn't have some unmaintained workbenches, each offering a single tool for a very specific purpose. Besides, nobody wants to remove those functionalities from FreeCAD. Just not show them by default since, let's face it, a very small percentage of users will ever need them. FreeCAD has that weird characteristic that the most important and useful tools are add-ons while the least useful ones are built-in. Fortunately, it's starting to change in current releases. We got rid of some redundant workbenches and tools that nobody used but they were included by default for some reason.

wwmayer:
> Same for inspection, it has a single tool that could be integrated into Part.

Makes no sense to me because the Inspection wb is able to do a visual inspection of any kind of input geometry (shape, points, meshes). If a user wants to do an inspection of two meshes it would be very illogical having to switch to the Part wb for this.

Also from a SW architecture point of view this would be a regression. One of the strengths of the current architecture is that a package builder can only build the modules he is interested in. So, he could even disable all CAD functionality and build a reduced package that only provides the Mesh and Inspection modules. By merging unrelated things together we would lose a lot of flexibility. 

> There are at least a dozen tools in Reverse engineering IMO the problem is more like it was with mesh WB in the past, this tools need icons, get placed in the toolbars and get more awareness on how to use them.

Exactly. Some functions like the triangulation of a point cloud are only accessible when linking the pcl library and thus not visible to everybody.

At the moment the only way to create a part object out of a mesh is the function in the Part wb but this is a very naive solution and totally useless for really big meshes. And even for smaller meshes there is no real gain of information about the actual geometry. For example, if the mesh represents a cylinder then the created part object still doesn't know that it's a cylinder. It's just a set of trimmed planes.

The Reverse Engineering wb is supposed to fix that by providing functions to restore the information about the actual geometry. Now, IMO it doesn't make sense to move these functions to any other wb.

It's correct that there are still missing a lot of functions and the UI isn't very polished either. So, I don't mind if the Reverse Engineering wb is disabled by default BUT I do mind trying to move any of its functions to another wb.

sasobadovinac:
> Please be more specific in your criticism. What exactly is dumbing freecad down to a toy so far? Is it disabling the 'Test' workbench ? Or perhaps the glorious 'None' wb? I mean who don't want to switch to the None wb and have a nice little break. Or perhaps simplifying the piece of art that was the sketcher constraint taskbox ? Or even maybe removing the Image wb and its so-intuitive ui?

It was not a critique on your work or you personally, but on  what it feels is a general drive of the project in the past few years,but if I take your comment as an example... How much time did you spend working on this? For me absolutely none of them give any additional or improved capabilities to FC. Improved usability and experience yes, bot not capabilities. On the other hand your work on sketcher offsets etc do. Again it is not a critique on you and if this would be just about your work I would not have an issue, but for the past few years most of the developers including werner, realthunder,... so folks that are capable to do really useful things, including you, have spend, at least IMO, an unreasonable a lot of time working on things that  that made no real usable improvements to FC just because some folks feel that it should be a bit more prettier. This mostly started with new users from fusion and comparing FC to fusion... This is a screenshot of some of teh early version of fusion from somewhere around 2014 

fusion

It was a simple and functional cad, but very limited in its capabilities. To compare for example to Catia v5 had probably in its full version over 100 different workbenches at least a decade earlier... FC started with the inspiration to be something like Catia, but lately it feels like it is just about how to make it more and more simple for the "new" users that never had and never will have any need or inspiration to do real engineering and are unwilling to put any effort in to it. None and Test WB were maybe useless, did they really bother me? No. Am I supper happy and looking forward to your improvements to sketcher? Yes.

@FEA-eng mentions the missing of default assembly etc. and I agree, but my view on that is that it is because everyone seems to be working on everything else instead of the core functionalities.

> This is just disabling wbs that are seldomly used. We're not talking about disabling FEM. We're talking about disabling Robot. How many users actually use Robot ? And for those extremely rare users I don't think re-enabling it in the preference is a problem.

We will be adding now the new startup wizard that is meant to improve on this issues with overwhelming functionality for new users. Sorry but I really don't see the need of this PR or you and all of us spending time on it. 

> For reverse engineering indeed I just saw that there are indeed many tools in the menu. The wiki is outdated. However I would still argue that 'Reverse engineering' feels like a undefined category. If you look at the different tools, the logic cohesion between them is weak. It feels like some should be in mesh, some in Points ...

By this logic we could put everything in to Mesh and Part, since they are the two geometry types that everything else is based on... Download one or two version old FC and open the Mesh WB, it was just something like this few...

PaddleStroke:
> Makes no sense to me because the Inspection wb is able to do a visual inspection of any kind of input geometry (shape, points, meshes). If a user wants to do an inspection of two meshes it would be very illogical having to switch to the Part wb for this.

Then perhaps it's the same situation as image workbench, why not merging this in core ? It would make sense to have this in the Tools menu rather than in it's own workbench if it can inspect any kind of input.

Besides I have been playing with this tool to see what it actually does and it is full of bugs.
- Click on 'Inspection...' then you can't leave this mode. Even changing of workbench doesn't work. The cursor shape is stuck to the inspection logo and right click is stuck too.
- Inspection between 2 meshes seems to produce a result that is acceptable : 
image
But inspection between 2 solids (Part primitives for example) is bad :
image
image

sasobadovinac:
To understand the actual usability of this tool see some of this...
https://holocreators.com/blog/deviation-analysis/
http://catiadoc.free.fr/online/cfyugrvs_C2/cfyugrvsdeviationanalysis.htm
https://www.youtube.com/watch?v=y_nY4S5mvms

sasobadovinac:
> and not much more than that by default. The remaining functionalities are considered specialized tools/add-ons. FreeCAD doesn't provide many of those workbenches (especially assembly) by default but it provides some niche workbenches like the ones discussed here or the ones already gone from 0.21.

I doubt that I need too much explanations about cad systems... But do you understand that reverse engineering is actually exactly what you were asking for in regard of complex surface reconstruction from scans (meshes, point clouds,...) ?

FEA-eng:
> I doubt that I need too much explanations about cad systems... But do you understand that reverse engineering is actually exactly what you were asking for in regard of complex surface reconstruction from scans (meshes, point clouds,...) ?

I was thinking about very basic mesh handling capabilities of typical CAD software. Something like being able to convert STL to CAD geometry, make some repairs and remesh. Usually, not much more than that is offered. Reverse engineering is way broader and typically requires a separate specialized software. Of course, it would be great to have it in FreeCAD (current capabilities of the aforementioned workbench are very limited) but not necessarily as a separate default workbench because most users don’t need it and their interaction with meshes ends at those basic operations listed above. Also because the RE workbench is so far from being user-friendly (after all, it’s tools are not documented and normally visible in GUI). For me, it seems to be of add-on value, unlike the Curves wb, for instance.

wwmayer:
> Then perhaps it's the same situation as image workbench, why not merging this in core ? It would make sense to have this in the Tools menu rather than in it's own workbench if it can inspect any kind of input.

That's not possible because the module has a hard dependency to other modules. Although, it currently provides only two functions but there might be a request for more analysis functions in the future.

> But inspection between 2 solids (Part primitives for example) doesn't seem to make sense :

Your example is somewhat pointless because nobody will do an inspection between two totally different geometries. 

Anyway, for the comparison the algorithm needs points from the "actual" geometry to compare it with the "nominal" geometry. But for a Part object there is a difference to a mesh or point cloud because its naturally available points are its vertexes and they are by far too few in order to do a sensible comparison. That's why it's internal tessellation is used but usually such meshes still don't deliver many points.

An alternative idea is to sub-sample the Part object but one has to be very careful with choosing a sensible distance as otherwise the generation of the points would take forever.

PaddleStroke:
> That's not possible because the module has a hard dependency to other modules. Although, it currently provides only two functions but there might be a request for more analysis functions in the future.

A code limitation shouldn't force itself on the UI imo. Is there no way for a module to load a general menu rather than be a separate workbench? Yes there might be more inspection tools later, I can see how a workbench would then make sense, but it is not the case and the workbench has been around for 12+ years. So what is the likelyhood that it will never happen?

> Your example is somewhat pointless because nobody will do an inspection between two totally different geometries.
> 
> Anyway, for the comparison the algorithm needs points from the "actual" geometry to compare it with the "nominal" geometry. But for a Part object there is a difference to a mesh or point cloud because its naturally available points are its vertexes and they are by far too few in order to do a sensible comparison. That's why it's internal tessellation is used but usually such meshes still don't deliver many points.
> 
> An alternative idea is to sub-sample the Part object but one has to be very careful with choosing a sensible distance as otherwise the generation of the points would take forever.

Well that sounds to me that this tool is not as versatile as you mentioned before. If this tool uses vertexes, the results are probably not so good with any parametric solids no?

PaddleStroke:
Long story short, what about this PR to disable those 4 wbs ? Is this going in or not?
At the end of the day, this is for maintainers to decide, can you discuss about it between you ? @sliptonic @chennes @yorikvanhavre @adrianinsaval @0penBrain

wwmayer:
> Well that sounds to me that this tool is not as versatile as you mentioned before. If this tool uses vertexes, the results are probably not so good with any parametric solids no? So perhaps it should actually be included in Mesh.

When considering an inspection between _actual_ and _nominal_ model then the most common use cases are:
mesh vs mesh, points vs mesh, mesh vs part and points vs part.
Doing an inspection between two parts is only a theoretical scenario and will almost never happen in practice. That's why in practice it doesn't harm that an actual part doesn't provide many points for the comparison.

sliptonic:
Sorry, @sasobadovinac but I strongly disagree that spending time on the UI, workflow, and discoverability is a waste of time.   In fact, even more coordinated effort should be spent on it starting earlier in the process so FreeCAD isn't a hodgepodge of workbenches each with limited utility. 

The topic that came up repeatedly in Vancouver was getting away from the 1:1 correspondence between modules and workbenches.  For some reason, there's a temptation to develop every new feature as a workbench.  One result is functionality that is poorly maintained and the user is forced to do lots of switching to get anything done.  It's even worse in the addon ecosystem where we have "workbenches" like DynamicData and Yaml.  

Instead, the workbenches should be user-facing, task-oriented, and collectively maintained.   We should have a unified approach to workbenches that actually serves users and minimizes switching.  We should also try to get away from the idea that workbenches are 'owned' by a developer.  I don't like that I'm so closely associated with Path.  It makes me into a gate-keeper and limits contributions.

chennes:
Thanks, @sliptonic -- that still leaves us staring at this PR in particular, though. Is this part of a systematic analysis and rethinking of the Workbenches? Or just another rshoot-from-the-hip UI change?

PaddleStroke:
Good analysis @sliptonic. The main problem is maybe not that there're too may entries. Though even if it's not the core issue, I think that it doesn't help and would still be a step in the good direction.

After thinking about it some more, I think the idea of adding tooltips to each workbench in the combobox would really help new users and would be pretty easy to do. (https://github.com/FreeCAD/FreeCAD/issues/8692#issuecomment-1674233633).

PaddleStroke:
Having said that, I think this PR still has some valid points that need to be addressed, and that regardless to the main issue : 

- OpenSCAD : This requires an external installation, and is probably used by a very small fraction of users. This is IMO addon-material.
- Inspection : Single tool workbench. It would be best as a menu entry in tools/inspection/. Is it possible for mods to not be separate workbench and instead just add a menu entry?
- Reverse Engineering : This is a WIP that has few complex and very specific functions that are totally undocumented. We should not be showing to the user such WIP workbenches. So I would still be in favor of disabling it by default.

So I am still in favor of disabling OpenSCAD and RE. And moving Inspection single tool out of it's own workbench.

sliptonic:
This PR was discussed this morning at the merge meeting.  The consensus is that there are several related issues with different fixes.  So We're going to close this PR without merging and then revisit the original issue and ask that it be split so they can be discussed one at a time.

@PaddleStroke raises additional issues that fit in that framework regarding OpenSCAD, Inspection, and RevEng.

sliptonic:
Individual issues have been created in the tracker along with a broader project to look at the contents of the workbench selector in general.
https://github.com/orgs/FreeCAD/projects/16/views/1

adrianinsaval:
What's the conclusion regarding robot? That one sounds like something we can even disable building by default

sliptonic:
Added issue to discuss Robot
https://github.com/FreeCAD/FreeCAD/issues/10315

Issue #10165, opened 2023-08-10
[Feature] Start page new file templates


qewer33:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Full version info

```shell
[code]
OS: Fedora Linux 38 (KDE Plasma) (KDE/plasmax11)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33675 (Git) AppImage
Build type: Release
Branch: (HEAD detached at 0.21.0)
Hash: 41b058e2087de60dc8fef3d3e68c7d0129e13abf
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * Assembly4 0.50.2
  * Curves 0.6.12
  * DynamicData 2.46.0
  * fasteners 0.4.58
  * freecad.gears 1.0.0
  * Glass
  * lattice2 1.0.0
  * kicadStepUpMod 10.19.4
  * Manipulator 1.5.0
  * Behave-Dark-Colors 0.1.1
[/code]
```

### Subproject(s) affected?

Other (specify in description)

### Problem description

One of the problems of FreeCAD for newcomers is not knowing what to do when they create a new file for the first time, staring into the endless 3D void inside the Start workbench with icons that make no sense for any kind of 3D work. Of course this can be discarded with comments like "New users should just watch tutorials, they should look at the wiki before creating a new file" etc. But many newcomers (especially if they're from another CAD software) have the confidence to just dive in and start using the program. As an example, I myself came to FreeCAD from Solidworks (thus I knew how the feature and sketch based CAD workflow worked) and I remember having to google "FreeCAD Part Design vs Part" to actually determine which workbench I should use to create a simple part.

This idea proposes a simple "fix" for that problem. Adding new file templates to the start workbench will help these new users see what workbench they should use for which kind of project. It will also reduce clicks for advanced users (for example instead of creating a new file, switching to the PD workbench and pressing "Create Body" you can just click the Part Design file template).

Current template ideas:
- Parametric Part: Spawns the user inside the Part Design workbench with a pre created Part Design Body
- CSG Part: Spawns the user inside the Part workbench (potentially with a pre-created geometry such as a Cube or a Sketch)
- Assembly: This is for the future when FreeCAD gets a default assembly workbench
- 2D Drawing: Spawns the user inside the Draft workbench
- Architecture: Spawns the user inside the Arch workbench

[...]

PR #10171, opened 2023-08-11
StartPage: Add new file templates and fix javascript error

qewer33:

---

This PR adds new file templates to the StartPage. More details can be found on the Github issue and the Forum discussion:
https://github.com/FreeCAD/FreeCAD/issues/10165
https://forum.freecad.org/viewtopic.php?p=698862

A few notes:
- The plus icon on "Empty File" is a new icon created by me, the older one looked weird on dark themes. This has a light color with a dark outer stroke (following the art design guidelines on the wiki) so it should look great on both light and dark themes.
- There is a second commit that fixes a javascript error that occures when changing tabs, this behavior also exists on the current 0.21 release so it might be best to backport it.
- The HTML/CSS/JS files of the StartPage were very poorly formatted, I let VSCode's default formatter take care of them.
- I am not sure if I need to add anything to any translation files for the added texts (they're only added to the TranslatedTexts.py file).
- I've noticed a lot of inconsistencies and weird code in the HTML/CSS files. They don't follow basic web development best practices and are quite messy, I have some webdev experience and I am willing to fix them in a future PR.
- This is my first PR to FreeCAD, if I messed up anything please point out. I will try my best to fix them.

What each "template" does in detail:
- Empty File: Just creates a new file
- Import File: Creates a new file and spawns the import dialog
- Parametric Part: Creates a new files, switches to the Part Design workbench, creates a new PD Body and creates a sketch (the user chooses the sketch plane)
- CSG Part: Creates a new file and switches to the Part workbench
- 2D draft: Creates a new file, switches to the Draft workbench and aligns the view with the default draft grid
- Architecture: Creates a new file and switches to the Arch workbench

Below is a screenshot of the change:

Screenshot_20230811_194620


yorikvanhavre:
I like this a lot! Not sure exactly "what" needs to be behind the different "new" buttons, if it's just a matter of switching to the required WB, or if we need more there (explanations? tutorial? "where to click next"?), but IMHO this is the right approach and it's elegantly integrated in the current FreeCAD workflow

qewer33:
Personally, I don't think a tutorial after creating a new file is needed (as of yet at least, could be a future plan).

Also @yorikvanhavre, what do you think about the javascript error fix? It happens on the current 0.21 release too (try switching tabs in the StartPage and look at the new error popups on the bottom right). It might be a good idea to backport it to the next 0.21 bugfix release.

yorikvanhavre:
> Personally, I don't think a tutorial after creating a new file is needed (as of yet at least, could be a future plan).

Of course, this was just a theoretical question... In any case, it's absolutely something that can be added later, this PR can still totally be merged as is.

Regarding the javascript fix, before, the for loop would run over 1,2,3,4 now I guess it will run over 1,2,3 only. Did you check if everything still runs OK if you activate the fourth tab (the Blog tab)? I think line 12 maybe must then become `for (var i = 1; i < tabs + 1; i++) {` ?

luzpaz:
@MisterMakerNL just a heads up about this proposal and possible impact on stylesheets

MisterMakerNL:
I rather had these buttons in the sidebar of the start workbench, just like part design workbench has "new part".
Qewer proposed doing it in the start page and sidebar.
I have some doubts about that, although I am open for testing it out.
A counter idea for this is to rename "new file" too "new project" and then present the user these different options in the sidebar.

qewer33:
> Regarding the javascript fix, before, the for loop would run over 1,2,3,4 now I guess it will run over 1,2,3 only. Did you check if everything still runs OK if you activate the fourth tab (the Blog tab)? I think line 12 maybe must then become `for (var i = 1; i < tabs + 1; i++) {` ?

Yeah I did, it works totally fine. I'm pretty sure the reason for that is, the "Blog" tab is just a static link and not really a tab. It doesn't switch to anything, just sends the user to the blog web page.

qewer33:
> @MisterMakerNL just a heads up about this proposal and possible impact on stylesheets

I am not aware if stylesheets can theme the StartPage or not but it shouldn't have any impact on any existing stylesheets since StartPage has it's own color preferences. Doesn't have any impacts on Preference Packs either since there are no color parameter changes.

qewer33:
> I rather had these buttons in the sidebar of the start workbench when starting a new file, just like part design workbench has "new part". Qewer proposed doing it in the start page and sidebar. I have some doubts about that, although I am open for testing it out. A counter idea for this is to rename "new file" too "new project" and then present the user these different options in the sidebar.

The main reasons I don't like that idea:
- The tasks panel doesn't show when you start FreeCAD, by default
- The StartPage source code is more of a separate entity (compared to tasks being a part of the main C++) so tasks panel implementation would be significantly harder, not sure if anyone is willing to take that on
- The tasks panel is really inconsistent among workbenches, some planning needs to be done about it before it gets used further imo
- Showing some options in the main view and some on the task view is also an inconsistency for the Start WB
- I really like the StartPage (and from what I can tell, most other users do too) but it has visible design and code flaws. This is (hopefully) an entry point to fixing those and making the StartPage better. Don't get me wrong, the StartPage is already great but it can be *much better*

MisterMakerNL:
I mean yes it's harder too do, but I still feel it belongs there.
Now you also set the idea that a user will think he needs a new file for doing sheetmetal and spreadsheet.
So it is confusing.
Here a little quick mockup:
afbeelding

FEA-eng:
I would definitely prefer to have it directly on the start page. There's a lot of space that could be better used there. I don't think we need that much space for recent files and especially for the examples (which are likely rarely used). So I like the original idea. It feels like in other CAD software, much more user-friendly.

MisterMakerNL:
@FEA-eng But Freecad doesn't have a drawing and a assembly format. Making it look like it does is confusing.

FEA-eng:
> @FEA-eng But Freecad doesn't have a drawing and a assembly format. Making it look like it does is confusing.

This screenshot from SW is just to show the layout, the buttons would of course be different in FreeCAD. There would be no assembly (for now) but we could think about adding a button for technical drawings (as discussed on the forum - it would add a TechDraw page and take the user to Part Design to start modeling). But it's optional, the buttons proposed by @qewer33 are more important (Parametric Part, CSG Part, Draft and so on).

yorikvanhavre:
It's not too hard to add taskbar items when starting the Startpage using `FreeCADGui.Control.addTaskWatcher()` [example here](https://github.com/yorikvanhavre/BIM_Workbench/blob/master/InitGui.py#L642) But then all the "new file" items need to become FreeCAD commands, which is more work.

All in all, I like having everything on the start page. And indeed this is an easy change, nothing prevents from adding more stuff later on. For now, I think I would suggest:

1. Renaming "New file" to "New project" indeed, that seems like a good idea, specially because the options below say more about a certain workflow you want to follow, not really the contents of the file you'll be producing
2. Renaming "Create parametric part" to "Create body" maybe? Because the Part WB also does parametric objects, not less than PartDesign. For me the difference is that with Part you do more some sort of "free modelling" and with PartDesign "restricted modelling" and indeed it should already create a body
3. Starting a new blank file with Part WB loaded is actually not very useful and will not help new users much... But indeed if it stays so, naming it "CSG" is a good idea.
4. Rename Architecture  to BIM. If the BIM add-on is present (try for ex. importing BimCommands) switch to BIM instead of Arch
5. Remove the name "template" in the descriptions if there is no real template.

> Yeah I did, it works totally fine. I'm pretty sure the reason for that is, the "Blog" tab is just a static link and not really a tab. It doesn't switch to anything, just sends the user to the blog web page.

Ok that makes sense. In the future I'd like that blog tab to become a true tab and fetch the blog contents via json, but we can change that when that's done :)

FEA-eng:
> 2. Renaming "Create parametric part" to "Create body" maybe? Because the Part WB also does parametric objects, not less than PartDesign. For me the difference is that with Part you do more some sort of "free modelling" and with PartDesign "restricted modelling" and indeed it should already create a body
> 3. Starting a new blank file with Part WB loaded is actually not very useful and will not help new users much... But indeed if it stays so, naming it "CSG" is a good idea.

I would definitely like those names to make a clear distinction between Part Design and Part for new users, this will significantly reduce their confusion. The question is how to name them. Maybe "Standard Part" for Part Design wb and "CSG Part" for Part wb? This way, new users would pick Part wb only if they knew what CSG is and that they need it. Otherwise, they would go for Part Design which corresponds to what most people are looking for in CAD software.

yorikvanhavre:
There is an [ongoing discussion](https://discord.com/channels/870877411049357352/1139679868561412217) about Part/PartDesign naming. Maybe better wait until something comes out of it

FEA-eng:
> There is an [ongoing discussion](https://discord.com/channels/870877411049357352/1139679868561412217) about Part/PartDesign naming. Maybe better wait until something comes out of it

Well, waiting for such big changes typically extends to infinity (since usually no consensus is made and the discussion just stops at some point). And I think that those templates could be added independently since they don't have to use the workbenches' names directly - it might even make more sense to use different names for them (since one of their main purposes is to make it clear which workbench to choose for which task without having to rename the workbenches drastically). Then, we could always change their names slightly if needed after the workbenches are renamed (but I don't think it would be necessary).

yorikvanhavre:
Yes. makes sense indeed. Better merge this now and adapt later

qewer33:
@yorikvanhavre just made it work with the BIM workbench and tweaked descriptions to not not include "template"

obelisk79:
I'm in favor of this merge.

PaddleStroke:
I have not tested it, but from the screenshot it looks good, I think for new users it will help getting started. And for every user of part design it's nice to have one click for a new file / body / sketch.

Can the button have tooltips ? If so it would also be interesting to show tooltips explaining the different workflows here. For example : 'Parametric part', tooltip would show a quick summary of what is part design and how it works with sketcher... Basically a summary of the wiki.

qewer33:
> Can the button have tooltips ? If so it would also be interesting to show tooltips explaining the different workflows here. For example : 'Parametric part', tooltip would show a quick summary of what is part design and how it works with sketcher... Basically a summary of the wiki.

Great idea. To add on that, instead of adding just plain tooltips, imo maybe custom HTML tooltips with a clickable wiki link along with a simple and short animated GIF showing the workflow would be nice. But it would be better to add that after this is merged, I don't want this PR to be too big.

qewer33:
This last commit removes the "Use gradient for the new file button" option as I realized that's no longer needed since this PR removes the old new file button.

kaktusus:
> I don't think we need that much space for recent files and especially for the examples (which are likely rarely used)

My list of recent files contains 50 items. :wink:

onekk:
> > I don't think we need that much space for recent files and especially for the examples (which are likely rarely used)
> 
> My list of recent files contains 50 items. 😉

It is very useful to have "recent files" in Start Page, with some space, as I think it is what the "silent majority of users" have used till now.

Good idea to place however something similar to this way to create new "documents" with some settings already done to follow "design philosophies" ()even if CSG is the worst choice for a "Part WB" made part.

Syres916:
One quite important point that I can't see mentioned is this PR isn't Python 3.8 compatible (`LoadNew.py` use of `match`), `match` wasn't even in Python 3.9 so should a new PR be raised in order for it to work on Python 3.8 and above? For info error raised:

```
14:35:25  Traceback (most recent call last):
  File "<string>", line 2, in <module>
<class 'SyntaxError'>: invalid syntax (<string>, line 29)
```

qewer33:
> One quite important point that I can't see mentioned is this PR isn't Python 3.8 compatible (`LoadNew.py` use of `match`), `match` wasn't even in Python 3.9 so should a new PR be raised in order for it to work on Python 3.8 and above? For info error raised:
> 
> ```
> 14:35:25  Traceback (most recent call last):
>   File "<string>", line 2, in <module>
> <class 'SyntaxError'>: invalid syntax (<string>, line 29)
> ```

Hmm seems like I did mess up in that regard. It can easily be changed to be an if statement, I can open a new PR that fixes it.

Syres916:
> Although I wonder, what is the minimum supported Python version of FreeCAD?

I'm referencing https://wiki.freecad.org/FreeCAD_1.0_Development_Cycle unless advised otherwise

PR #10206, opened 2023-08-13
TD: Default object name translation

chennes:
@WandererFan This is a proposal for a technique to generate new document objects with a non-translated internal name, but a translated Label, retaining the unique suffix generated for the internal name. I've created a new function, private to Command.cpp, called `addObjectWithTranslatedLabel`. You provide the type of object and an untranslated name string (that when calling this function you add to the translation database with a NOOP call). It generates the unique name, the translated version of the unique name, creates the object, applies the translated name as the label, and returns a pointer to the document object it created.

### DIFF
--- a/src/Mod/TechDraw/Gui/Command.cpp
+++ b/src/Mod/TechDraw/Gui/Command.cpp
@@ -84,6 +84,29 @@ class Vertex;
 using namespace TechDrawGui;
 using namespace TechDraw;
 
+namespace {
+    // This function creates a document object that has a unique name based exactly on the provided baseName, but also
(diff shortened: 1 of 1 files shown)


WandererFan:
I'm working on this and #10202.   Trying to get the same approach working everywhere we add an object in both TechDraw/App and TechDraw/Gui.

WandererFan:
I'm going to close this PR.  There are some features (section and complexSection) where building the Label from the object name doesn't work.  There are also a couple of cases (DrawProjGroup) where the addObject happens on the App side, so a function in Command doesn't help.  There's also the other CommandXXX files that need tweaking.  

All my changes will be included in TD: Translate default labels #10202

Issue #5800, opened 2022-02-07
Renaming alias in spreadsheet changes the spreadsheet part of the reference from label to internal name


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=3560

* **Reporter:** lokomoko
* **Date submitted:** 8/15/2018
* **FreeCAD version:** 0.17
* **Category:** Bug
* **Status:** confirmed
* **Tags:** equation, expressions, formula, spreadsheet

# Original report text

Renaming an alias in a spreadsheet changes the spreadsheet-part of a reference to the alias from Label (e.g. "MyInputData") to internal name/Id (e.g. "Spreadsheet001").

Discussed in Forum under [Renaming alias in spreadsheet changes the name of the spreadsheet in any reference](http://forum.freecad.org/viewtopic.php?f=3&t=30294)

# Additional information

OS: Ubuntu 16.04.5 LTS
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.17.13528 (Git) AppImage
Build type: None
Branch: releases/FreeCAD-0-17
Hash: 5c3f7bf8ec51e2c7187789f7edba71a7aa82a88b
Python version: 2.7.6
Qt version: 4.8.6
Coin version: 4.0.0a
OCC version: 7.3.0
Locale: German/Germany (de_DE)

# Steps to reproduce

1. Create new document
2. In spreadsheet workbench create a spreadsheet by the name "MyData"
3. Create cell with an alias e.g. "BoxSize" and a value
4. In part workbench create a 3D-object (e.g. cube) and therein a reference to the alias (e.g. "=MyData.BoxSize").
5. In spreadsheet change name of alias to BoxLength

Actual Result: The reference in the 3D-Object is "=Spreadsheet001.BoxLength" now, that is the alias was changed, but also the name of the spreadsheet
Expected Result: The reference in the 3D-Object should be "=MyData.BoxLength"

# Other bug information

* **Priority:** normal
* **Severity:** minor
* **Category:** Bug
* **Updated:** 11/24/2021

# Discussion from Mantis ticket

### Comment by lokomoko 2018-08-15 12:41
Direct link to forum topic [Renaming alias in spreadsheet changes the name of the spreadsheet in any reference](https://forum.freecadweb.org/viewtopic.php?f=3&t=30294)

---

### Comment by wmayer 2018-11-11 11:27
[...]

PR #10222, opened 2023-08-14
Keep the label of a variable in expression on spreadsheet alias rename

djurodrljaca:
Fixes the bug #5800 

The bug fix makes sure that the original document object name and sub object name are not changed in an expression when an alias in a spreadsheet is renamed.


chennes:
Thanks for the PR -- sorry for the delay in letting the CI run, I've enabled it now. I'll take a look at it locally and we should be able to get it merged shortly.

src/App/Expression.cpp
```diff
@@ -2892,10 +2892,18 @@ bool VariableExpression::_renameObjectIdentifier(
...
+        ObjectIdentifier::String originalDocumentObjectName = var.getDocumentObjectName();
+        std::string originalSubObjectName = var.getSubObjectName();
         if(path.getOwner())
             var = it->second.relativeTo(path);
         else
             var = it->second;
+        if (originalHasDocumentObjectName) {
+            var.setDocumentObjectName(std::move(originalDocumentObjectName),
+                                      true,
+                                      std::move(originalSubObjectName));
```
chennes, line 2905: It's not a particularly important thing to fix, but the `std::move` of `originalSubObjectName` won't actually do a move here, because it's a `std::string`, not a `ObjectIdentifier::String` -- that means it's actually getting passed as a `const std::string &` to the constructor of `ObjectIdentifier::String`, which does not do a move, and ultimately copies the string.

PR #10227, opened 2023-08-14
Draft: Translate task titles for draft tools

chennes:
Fixes https://github.com/FreeCAD/FreeCAD-translations/issues/260

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_lines.py
+++ b/src/Mod/Draft/draftguitools/gui_lines.py
@@ -63,18 +63,21 @@ class Line(gui_base_original.Creator):
                 'MenuText': QT_TRANSLATE_NOOP("Draft_Line", "Line"),
                 'ToolTip': QT_TRANSLATE_NOOP("Draft_Line", "Creates a 2-point line. CTRL to snap, SHIFT to constrain.")}
 
-    def Activated(self, name="Line", icon="Draft_Line"):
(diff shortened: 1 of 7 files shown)


src/Mod/Draft/draftguitools/gui_lines.py
```diff
@@ -63,18 +63,21 @@ def GetResources(self):
                 'MenuText': QT_TRANSLATE_NOOP("Draft_Line", "Line"),
                 'ToolTip': QT_TRANSLATE_NOOP("Draft_Line", "Creates a 2-point line. CTRL to snap, SHIFT to constrain.")}
 
-    def Activated(self, name="Line", icon="Draft_Line"):
+    def Activated(self, name=QT_TRANSLATE_NOOP("draft","Line"), icon="Draft_Line", task_title=None):
```
Roy-043, line 66: It's a detail, but it would perhaps be clearer to change:
`    def Activated(self, name=QT_TRANSLATE_NOOP("draft","Line"), icon="Draft_Line", task_title=None):`
To:
`    def Activated(self, name="Line", icon="Draft_Line", title=translate("draft", "Line):`
You could then do away with:
```
        if task_title is None:
        ...
```

chennes: I think that will fail on a language change because of the way Python handles default arguments.

PR #10228, opened 2023-08-14
Draft: Translate label type combobox

chennes:
Fixes https://github.com/FreeCAD/FreeCAD-translations/issues/259


Roy-043 (commented):
The return value from `get_label_types` is also used for the `LabelType` property enumeration list. Translated strings would not work there.

Roy-043:
Aha, right. Forgot about that. Thanks for explaining. Merging then...

PR #10238, opened 2023-08-15
[Preferences] theme : Variables in stylesheet and more.

PaddleStroke:
- Add support for variables in stylesheets for accent colors.
- Rename the theme colors 'Accent colors'
- Rename themes to remove 'theme' from name.
- Create themes for Behave-Dark, ProDark, Dark-contrast, Dark-modern, Darker, Light-modern.

I created themes for all the stylesheets we supported before (I think). Though I am not sure we want to keep so many themes around. Maybe some should be removed from core and offered as addons. Ultimately I think we should offer 1 light and 1 dark themes so that they can be maintained well and of high quality. Now the question is which should it be, which is going to be a huge problem :)

Personally I think ProDark is my favorite. Though I have not used the Dark-modern yet.

@MisterMakerNL Do you know who made which stylesheet ? Which themes do you think should be removed from build-in if any?


MisterMakerNL:
Only one I would like to see go obsolete now is the high contrast since I hope other stylesheet have enough contrast that we don't need that version.
In the future hopefully only the modern versions stay and the rest go into addons. Now they are all in so people can get used to it and give feedback.

obelisk79:
The modern stylesheets have style guidelines. The zero padding, while fitting the most functions on the toolbar makes the toolbars look cluttered due to not enough space separating the already visually complex icons.

MisterMakerNL:
@obelisk79 that is exactly the feedback we need to make it better and why it's included. Although wrong place😄.

src/Gui/Application.cpp
```diff
@@ -2532,6 +2532,27 @@ void Application::setStyleSheet(const QString& qssFile, bool tiledBackground)
...
 
+QString Application::replaceVariablesInQss(QString qssText)
+{
+    //First we fetch the colors from preferences,
+    ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Themes");
+    unsigned long longAccentColor1 = hGrp->GetUnsigned("ThemeAccentColor1", 0);
+    unsigned long longAccentColor2 = hGrp->GetUnsigned("ThemeAccentColor2", 0);
+    unsigned long longAccentColor3 = hGrp->GetUnsigned("ThemeAccentColor3", 0);
+
+    //convert them to hex, Note, the ulong contains alpha channels so 8 hex character when we need 6 here.
```
luzpaz, line 2543: Recommend: 
```
//Convert them to hex
//Note: the ulong contains alpha channels so 8 hex character when we need 6 here.
```

MisterMakerNL:
I got it compiled!
First it is awesome!
Second there some bugs.
First is if you apply a theme in the general settings it doesn't change the stylesheet (advanced) settings. So if you change the colors and press ok it also changed the stylesheet to whatever it was set too.
FreeCAD_SLpkoVMCfE

I diffidently screwed up the hover buttons on the older styles.
afbeelding

PaddleStroke:
Just fixed.
There is still the problem raised by @wwmayer about the reportview hidding (and once I got the python console showing instead). Though I cannot find where this is coming from as the cfg (https://github.com/Ondsel-Development/FreeCAD/blob/pref_theme_2/src/Gui/PreferencePacks/Dark/Dark.cfg) doesn't seem to have that.

chennes:
Do you want to wait on this PR until you have solved that last problem (the report view hiding), or would you like to merge as-is and submit a new PR when you find that issue?

PaddleStroke:
I prefer merge as soon as it's reviewed. A following PR can come later.

MisterMakerNL:
I tested it, works fine now!
FreeCAD_g7FV22gw2C

MisterMakerNL:
> 

haha yes, but the power is restricted to what the stylesheet maker wants. Although you can still make all black buttons....

chennes:
Thanks for the PR -- please make sure that those three variables that can now be used in the stylesheets get documented someplace. I don't think we have a Handbook page that applies to Stylesheet authors right now, so I'm not sure where the most appropriate place is.

NXTler:
Would it be possible to set background colors (e.g. Toolbars, Combo View) with the variables?

PaddleStroke:
I'm no stylesheet expert but I think so, that's for the stylesheet designer to decide where he applies the accent colors.

NXTler:
Thanks, I might do a feature request about it for the default themes delivered with FreeCAD.

PR #10242, opened 2023-08-15
Gui: Change SelectionFilterPy to new PyCXX extension type

marioalexis84:
Move class to its own files.
Add `SelectionFilterPy::getFilter` method.
Change class name from `SelectionFilter` to `Gui.SelectionFilter`.
Use range-based loop in `SelectionFilterPy::result`

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -83,6 +83,7 @@
 #include "MDIViewPy.h"
 #include "SoFCDB.h"
 #include "Selection.h"
+#include "SelectionFilterPy.h"
 #include "SoFCOffscreenRenderer.h"
 #include "SplitView3DInventor.h"
 #include "TaskView/TaskView.h"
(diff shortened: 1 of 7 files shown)


wwmayer:
@marioalexis84 
The new-style class SelectionFilterPy causes a regression. 

The steps to reproduce:
1. Create a cube
2. Set this filter
```
filter = Gui.Selection.Filter('SELECT Part::Feature')
Gui.Selection.addSelectionGate(filter)
```
3. Hover with the mouse over the cube
4. Crash!

For some reason the data of the `Ast` object are corrupted. When restoring the old-style class everything works again.

wwmayer:
Found it. When printing the memory address of SelectionFilterPy inside its constructor then it is different to the address of the cast SelectionFilterPy in SelectionSingleton::sAddSelectionGate.

This is because the SelectionFilterPy is not the passed Python object itself but a component of it. That's why all members are corrupted. For SelectionFilterPy a converter method must be added to return the real SelectionFilterPy instance.

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/10433

PR #10258, opened 2023-08-17
Update translations

chennes:
Adds the ten units strings from Base as well.

ETA: I'm still working on figuring out why we've got all these translations being removed. I deleted the branch I had made on CrowdIn and thought that had fixed it, but a closer look shows it did not.

ETA2: I figured it out -- when creating the branch on CrowdIn their system suggested we change the Duplicates setting to "Show within a Version Branch". When I change that setting back to "Show, but Auto-translate" we get our strings back. @yorikvanhavre and @kaktusus can you give this a look and see if you see anything amiss?


kaktusus:
I reviewed several .ts files with translations (not only Polish) everything looks fine.
🥇

PR #10269, opened 2023-08-18
[Core] Custom Data Elements/Document-wide Variables

AjinkyaDahale:
See [this blog post](https://ondsel.com/blog/build-system-for-custom-data-elements/) and [this issue](https://github.com/Ondsel-Development/FreeCAD/issues/5) for context.

As opposed to the blog post, where we proposed creating a `CustomData` object to store these variables, they are now stored directly in the document. This functionality has already existed for some time, but is not "complete" in the sense that these variables cannot be accessed through the expressions engine. I'll attempt fixing this as this draft PR progresses, or make it a separate PR.

Currently, the only addition in this PR is the support for keyword arguments, so for example, one can do these examples:
```
my_doc = App.ActiveDocument()
my_doc.addProperty(“App::PropertyFloat”, name=”myDim”, doc="My Custom Dimension") # Setting a double with a docstring
my_doc.addProperty(“App::PropertyFloat”, name=”myDim2”, group="CustomDims") # Setting a double with group
```
with the remaining data given default values.

Some additions targeted:
1. Make these variables visible in the expression engine. 
2. Improve UI of the property panel, including renaming variables.
3. Make it possible to create variables within the expression engine. For example `myNewLength = 12.34 mm" should create a new variable `myNewLength` at the document level (or optionally in the document object with an alternative way to define at document level).


sliptonic:
[Problem: FreeCAD lacks a consistent user property system](https://trello.com/c/MBTDPH3T/14-problem-freecad-lacks-a-consistent-user-property-system)

Issue #10243, opened 2023-08-15
[Feature Request] Selection filters (select only faces, edges or vertices)


FEA-eng:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.21.0.33668 +7 (Git)
Build type: Release
Branch: (HEAD detached at 0.21)
Hash: 41b058e2087de60dc8fef3d3e68c7d0129e13abf
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: Polish/Poland (pl_PL)
```

### Subproject(s) affected?

Core

### Problem description

FreeCAD lacks an option to temporarily restrict selection to one type of geometric entities (faces, edges or vertices). It's common in other CAD software and very helpful when applying fillets or using other tools that require multiple manual selections. It would be also useful in the FEM workbench when defining constraints.

Such functionality is already available in the Assembly4 workbench but in my opinion, it deserves implementation in the core.

Forum discussion: https://forum.freecad.org/viewtopic.php?t=80457
Older discussion: https://forum.freecad.org/viewtopic.php?t=67783

There's a bounty offered for the implementation of this feature: https://forum.freecad.org/viewtopic.php?t=80395

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10271, opened 2023-08-18
[Gui] Add new Selection Filter functionality

Syres916:
This PR is intended to add the ability for the user to choose a Selection Filter based on _Vertex_, _Edge_ or _Face_ in the 3D View.
A dropdown toolbar has been created in order to minimise the space taken on the View toolbar.

Fixes https://github.com/FreeCAD/FreeCAD/issues/10243

I am already preparing documentation for this quite simple additional functionality.

### DIFF
--- a/src/Gui/CommandView.cpp
+++ b/src/Gui/CommandView.cpp
@@ -1359,6 +1359,230 @@ void StdCmdViewTop::activated(int iMsg)
     doCommand(Command::Gui,"Gui.activeDocument().activeView().viewTop()");
 }
 
+
+//===============================================================================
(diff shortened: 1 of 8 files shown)


FEA-eng:
I tested this new feature and it works as expected.

wwmayer:
Functionality-wise the PR is OK but having it in the core system is actually the wrong location because the selection filters only works for Part objects. Thus, the Part module would be the correct place.

Since your intention apparently was to make the functions available independent of the active workbench we need a mechanism to manipulate a workbench from another module. I will have a look on how to achieve that...

Issue #10264, opened 2023-08-18
[Problem] single quote in file path breaks opening from recent history


janvanderwatt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.2.29177 +426 (Git)
Build type: Release
Branch: (HEAD detached from 0.20.2)
Hash: 930dd9a76203a3260b1e6256c70c1c3cad8c5cb8
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/Australia (en_AU)
Installed mods: 
  * 3D_Printing_Tools
  * CurvedShapes 1.0.4
  * Curves 0.4.3
  * DynamicData 2.46.0
[/code]
```

### Subproject(s) affected?

None

### Problem description

If a file path contains a single quote (apostrophe), it can be opened using File->Open, and it will appear in the recent history displayed at startup; however, clicking on it will result in this error:
![image](https://github.com/FreeCAD/FreeCAD/assets/4944435/6e9f25a5-2db0-471a-9478-96cc9c526e7a)
This problem is new to 0.21, and does not occur in 0.20.2:
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.2.29177 +426 (Git)
Build type: Release
Branch: (HEAD detached from 0.20.2)
Hash: 930dd9a76203a3260b1e6256c70c1c3cad8c5cb8
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/Australia (en_AU)
Installed mods: 
  * 3D_Printing_Tools
  * CurvedShapes 1.0.4
  * Curves 0.4.3
  * DynamicData 2.46.0
[/code]

Removing the single quote from the directory name is a successful workaround to this problem.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
This was fixed in master and fix was also backported to v0.21
Props to @bdieterm

PR #10278, opened 2023-08-19
Core: escape filepath characters in FreeCAD.loadFile

bdieterm:
Currently, executing the Python command `FreeCAD.loadFile("/tmp/let's go.FCStd")` will fail because of the single quote sign in the path.

This pull request fixes the problem by properly escaping the special characters in the string.

The error was described in #10264 as happening when "opening from recent history". The root cause lies in function `FreeCAD.loadFile` and therefore the error also happens on FreeCAD version 0.20.2 when calling that function directly instead of via the history action.


luzpaz:
Warrants a backport ?

bdieterm:
I recommend a backport.

There is the possibility to run arbitrary Python code when someone opens your file a second time via the Start Page history, or when someone opens your file via `FreeCAD.loadFile`.

Example:

Create a FreeCAD file and give it this name:
`aa'+exec(bytes.fromhex('4775692e6765744d61696e57696e646f7728292e73657457696e646f775469746c652827434f44452045584543555445442729')))#'.FCStd`
This will execute `Gui.getMainWindow().setWindowTitle('CODE EXECUTED')`

luzpaz:
@bdieterm please open a PR against `releases/FreeCAD-0-21` for backport

luzpaz:
> There is the possibility to run arbitrary Python code when someone opens your file a second time via the Start Page history, or when someone opens your file via FreeCAD.loadFile.

How do you propose to avoid this ?

bdieterm:
> How do you propose to avoid this ?

This PR is already the fix, so the code execution vulnerability does not exist anymore in FreeCAD:master.

PR #10291, opened 2023-08-21
StartPage: Fix new file templates issues

qewer33:

---

This is an addition to my previous PR which implemented new file templates (better to call them quick-start buttons actually): https://github.com/FreeCAD/FreeCAD/pull/10171
Also see forum discussion: https://forum.freecad.org/viewtopic.php?t=80533

This PR makes 3 changes:
- Address the issue mentioned in this comment (https://github.com/FreeCAD/FreeCAD/pull/10171#issuecomment-1686346899)
- Change the Part Design template to not create a sketch by default (community feedback)
- Remove (comment out) the Part template because a suitable name is hard to find and it isn't that useful anyways, it can just be uncommented/re-enabled later when a proper name is found


src/Mod/Start/StartPage/LoadNew.py
```diff
@@ -26,32 +26,28 @@
 # template will be given before this script is run
 template_name = str(template)
 
-match template_name:
-    case "empty_file":
+if template_name == "empty_file":
         FreeCADGui.runCommand('Std_New')
         StartPage.postStart()
-    case "import_file":
+elif template_name == "import_file":
```
chennes, line 32: Can you please fix the indentation of this block, now that it's just conditionals and not a match/case?

qewer33: Done

PR #10294, opened 2023-08-21
[Core] Use keyword arguments in DocumentObjectPy `addProperty`

AjinkyaDahale:
On the lines of #10269. Keyword arguments may be used in other places as well to simplify the console interface.


sliptonic:
[PR 10294](https://trello.com/c/JVa7epKQ/2-pr-10294)

src/App/DocumentObjectPyImp.cpp
```diff
@@ -75,23 +75,39 @@ Py::Object DocumentObjectPy::getDocument() const
...
+PyObject*  DocumentObjectPy::addProperty(PyObject *args, PyObject *kwd)
 {
     char *sType,*sName=nullptr,*sGroup=nullptr,*sDoc=nullptr;
     short attr=0;
     std::string sDocStr;
     PyObject *ro = Py_False, *hd = Py_False;
-    if (!PyArg_ParseTuple(args, "s|ssethO!O!", &sType,&sName,&sGroup,"utf-8",&sDoc,&attr,
-        &PyBool_Type, &ro, &PyBool_Type, &hd))
+    PyObject* enumVals = nullptr;
+    static char *kwlist[] = {"type","name","group","doc","attr","read_only","hidden","enum_vals",nullptr};
```
chennes, line 85: This should now be changed to
```suggestion
    const std::array<const char *, 9> kwlist {"type","name","group","doc","attr","read_only","hidden","enum_vals",nullptr};
```

and the call to `PyArg_ParseTupleAndKeywords` replaced with a call to `Wrapped_ParseTupleAndKeywords` -- this is made for C++ compliance, and is required for Qt 6 on MSVC.

AjinkyaDahale: Will do. Did you also make these changes after merging #10269? Also, specifying the `9` seems inelegant.

chennes: I am taking care of all the ones in the current codebase, so you don't need to worry about those. And yes, it would be nice if the compiler would supply the 9 for us, but it cannot: although the **number** of elements must be constant in a `std::array`, their actual contents is not required to be, and you can default-initialize it "empty" which will default-init all of its elements.

AjinkyaDahale: Just pushed the changes, though I cannot confirm right now on my local machine if it compiles.

src/App/DocumentObjectPyImp.cpp
```diff
@@ -76,23 +76,39 @@ Py::Object DocumentObjectPy::getDocument() const
...
 {
     char *sType,*sName=nullptr,*sGroup=nullptr,*sDoc=nullptr;
     short attr=0;
     std::string sDocStr;
     PyObject *ro = Py_False, *hd = Py_False;
-    if (!PyArg_ParseTuple(args, "s|ssethO!O!", &sType,&sName,&sGroup,"utf-8",&sDoc,&attr,
-        &PyBool_Type, &ro, &PyBool_Type, &hd))
+    PyObject* enumVals = nullptr;
+    const std::array<const char *, 9> kwlist {"type","name","group","doc","attr","read_only","hidden","enum_vals",nullptr};
+    if (!Wrapped_ParseTupleAndKeywords(
```
chennes, line 87:
```suggestion
    if (!Base::Wrapped_ParseTupleAndKeywords(

```

yorikvanhavre:
This fixes adding enumeration properties to documents... Is it good for you @chennes ?

yorikvanhavre:
image
That's the pre-commit reason for failing.. It will likely fail for all PRs coming from Ondsel...

AjinkyaDahale:
> image That's the pre-commit reason for failing.. It will likely fail for all PRs coming from Ondsel...

The hint does indicate so. I don't see the difference between a PR from an individual vs an organization though.

yorikvanhavre:
I think it's because precommit actually adds a commit to your PR, and this is apparently not possible here because it has no right to push to your ondsel branch.

When I do a PR, it offers me the possibility to "let project maintainers push to my branch" or something like that. Is that option not available when making a PR from Ondsel?

AjinkyaDahale:
> When I do a PR, it offers me the possibility to "let project maintainers push to my branch" or something like that. Is that option not available when making a PR from Ondsel?

I am aware of the checkbox you mention. Usually I leave it on as a default in my PR, so probably this may be an Ondsel setting.

chennes:
@AjinkyaDahale is there a reason you can't run pre-commit locally? Then the fact that the CI can't add to your PR won't matter.

AjinkyaDahale:
> @AjinkyaDahale is there a reason you can't run pre-commit locally? Then the fact that the CI can't add to your PR won't matter.

How do I do that?

chennes:
https://github.com/FreeCAD/DevelopersHandbook/blob/master/gettingstarted/index.md

See item3 in "Setting up for Development".

luzpaz:
bumping

AjinkyaDahale:
> https://github.com/FreeCAD/DevelopersHandbook/blob/master/gettingstarted/index.md
> 
> See item3 in "Setting up for Development".

OK I'm not sure what just happened but I followed the instructions and pre-commit seems to succeed.

yorikvanhavre:
Merging as this seems ok to go now

PR #10296, opened 2023-08-21
Draft: remove getPlacementFromPoints function

Roy-043:
This somewhat convoluted function was only used by gui_trackers.


adrianinsaval:
Is this considered a public API? Could it maybe be in use by some macro?

Roy-043:
That is indeed an issue. I have meanwhile found that the BIM WB uses the function. Will convert this to Draft for now.

luzpaz:
CC @yorikvanhavre (BIM WB related)

Roy-043:
The BIM code (BIMBox) that relies on getPlacementFromPoints in WorkingPlane.py still works after this PR. Users that use the dev version will only see a `use_instead` warning.

PR #10328, opened 2023-08-23
fix: windows build need /utf-8

ppphp:
FreeCAD\src\Mod\TechDraw\Gui\CommandExtensionDims.cpp(174) has a utf-8 character


luzpaz:
https://github.com/FreeCAD/FreeCAD/blob/5a153e50ffb6046b6cc4dba498344032f6edb952/src/Mod/TechDraw/Gui/CommandExtensionDims.cpp#L174

PR #10330, opened 2023-08-23
[BASE] Removed dead code

mosfet80:


src/Base/Interpreter.cpp
```diff
@@ -467,7 +454,7 @@ bool InterpreterSingleton::loadModule(const char* psModName)
 {
     // buffer acrobatics
     //PyBuf ModName(psModName);
-    PyObject *module{};
```
chennes, line 470: Why drop the initialization here?

Issue #10099, opened 2023-08-07
[Problem] Open file dialog freezes on windows (upstream bug solved in Qt6) 


luzpaz:
This ticket is a placeholder for #5853 and depends on #6992

luzpaz:
In short when Qt6 for windows port is implemented, this ticket should be closed.

adrianinsaval:
Please add a proper description in this issue itself, our guidelines says that an issue should be understood without requiring to follow links

adrianinsaval:
Btw I didn't see any confirmation that this issue is actually solved in qt6, just speculation that it might be

chennes:
I'm hoping to have the Qt6 LibPack ready in a couple of weeks, so we should be able to find out for sure soon.

luzpaz:
@chennes can this ticket be closed ?

chennes:
> In short when Qt6 for windows port is implemented, this ticket should be closed.

If this is true, then yes. I don't know anything about this issue, and don't experience any problems with Qt's open file dialog in Qt6 (but I also didn't experience the issue with Qt5, so that doesn't really mean anything :) ).

PR #10337, opened 2023-08-24
Update for LibPack3

chennes:
v3 of the LibPack will be based on Qt 6.7, Boost 1.85, Python 3.12, and OCCT 7.8 -- this PR updates the FreeCAD source to allow compilation of this system using MSVC.

Qt6 on MSVC requires compilation with the `/permissive-` option (that is, it turns **off** all "permissive" options -- Qt6's cMake configuration will add automatically, it doesn't need to be in FreeCAD's setup) -- this has a number of consequences that require us to be a little more correct in our code, most notably with const-correct C-style strings, which represent the bulk of the code changes.

Other code changes:
* Some old workarounds for Boost errors no longer work, but are also no longer necessary
* The Gui::GuiNativeEvent::RawInputEventFilter interface has changed in Qt6
* References *must* be initialized
* The transition to the const-correct Wrapped_ParseTupleAndKeywords is now complete
* `QtPlatformHeaders/QWindowsWindowFunctions` no longer exists
* I had to move the Boost intrusive pointer into the GUI namespace -- I'm not clear on why this change was needed.

cMake changes:
* The new LibPack is versioned, so cMake can check for that now
* The code that searches for PySide6 and Shiboken6 has been moved into findXXX files, which seems to be more in keeping with "the cMake way" than the previous solution.


luzpaz:
@chennes is there documentation on how to update the libpack ?

chennes:
There will be when it is ready. I've scripted the process, but won't submit the PR until FreeCAD actually compiles against it.

chennes:
@0penBrain do you have any idea what's going on with that failing Windows CI run? CMake appears to be erroring out, but I don't see any errors in its output. (side note -- can we set up a Qt6 CI build?)

0penBrain:
The error is at line 363.

You also have call stacks (don't know how critical they are) at lines 93 and 106.

chennes:
Thanks @0penBrain -- I guess I just need another set of eyes on it!

luzpaz:
Please test #10099 after merging this PR

cMake/UseLibPack3.cmake
```diff
@@ -0,0 +1,75 @@
...
+set (Qt6Qml_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6Qml CACHE PATH "")
+set (Qt6Quick_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6Quick CACHE PATH "")
+set (Qt6QuickTest_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6QuickTest CACHE PATH "")
+set (Qt6QuickWidgets_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6QuickWidgets CACHE PATH "")
+set (Qt6Sql_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6Sql CACHE PATH "")
+set (Qt6Svg_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6Svg CACHE PATH "")
+set (Qt6Test_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6Test CACHE PATH "")
+set (Qt6UiPlugin_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6UiPlugin CACHE PATH "")
+set (Qt6UiTools_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6UiTools CACHE PATH "")
+set (Qt6Widgets_DIR ${Qt6_ROOT_DIR}/lib/cmake/Qt6Widgets CACHE PATH "")
```
Neumann-A, line 38: Why can this not use `find_package`?

cMake/UseLibPack3.cmake
```diff
@@ -0,0 +1,75 @@
+set(ENV{PATH} "${FREECAD_LIBPACK_DIR};$ENV{PATH}")
+set(ENV{CMAKE_PREFIX_PATH} ${FREECAD_LIBPACK_DIR})
```
Neumann-A, line 2: Why is this modifying CMAKE_PREFIX_PATH as an env?

cMake/UseLibPack3.cmake
```diff
@@ -0,0 +1,75 @@
...
+
+# Make sure we are using the static versions of Boost here: the LibPack includes both
+set(Boost_DIR ${FREECAD_LIBPACK_DIR}/lib/cmake CACHE PATH "" FORCE)
+set(Boost_USE_STATIC_LIBS OFF)
+find_package(Boost COMPONENTS filesystem program_options regex system thread date_time REQUIRED)
```
Neumann-A, line 7:
```suggestion
set(Boost_DIR ${FREECAD_LIBPACK_DIR}/lib/cmake CACHE PATH "" FORCE)
set(Boost_USE_STATIC_LIBS OFF)
find_package(Boost COMPONENTS filesystem program_options regex system thread date_time REQUIRED)
```

`find_package` also has a PATHS argument. So you might not need to set `Boost_DIR` explicitly

cMake/UseLibPack3.cmake
Neumann-A: you probably want `list(PREPEND CMAKE_PREFIX_PATH "${FREECAD_LIBPACK_DIR}")` (or `APPEND` if you want people to take control)

cMake/UseLibPack3.cmake
```diff
@@ -0,0 +1,75 @@
...
+if (XercesC_LIBRARY_RELEASE AND XercesC_LIBRARY_DEBUG)
+    set (XercesC_LIBRARIES debug ${XercesC_LIBRARY_DEBUG} optimized ${XercesC_LIBRARY_RELEASE})
+elseif (XercesC_LIBRARY_RELEASE)
+    set (XercesC_LIBRARIES ${XercesC_LIBRARY_RELEASE})
+elseif (XercesC_LIBRARY_DEBUG)
+    set (XercesC_LIBRARIES ${XercesC_LIBRARY_DEBUG})
+else()
+    message (FATAL_ERROR "Could not find XercesC library in the LibPack")
+endif()
+set (XercesC_FOUND TRUE)
```
Neumann-A, line 52: xercesc in vcpkg has `XercesCConfig.cmake` so why this manual search here? Also this belongs into its own FindXercesC.cmake the way it is provided. If you provide a Find module also provide targets the modern cmake way.

cMake/UseLibPack3.cmake
```diff
@@ -0,0 +1,75 @@
...
+if (COIN3D_LIBRARY_RELEASE AND COIN3D_LIBRARY_DEBUG)
+    set (COIN3D_LIBRARIES debug ${COIN3D_LIBRARY_DEBUG} optimized ${COIN3D_LIBRARY_RELEASE})
+elseif (COIN3D_LIBRARY_RELEASE)
+    set (COIN3D_LIBRARIES ${COIN3D_LIBRARY_RELEASE})
+elseif (COIN3D_LIBRARY_DEBUG)
+    set (COIN3D_LIBRARIES ${COIN3D_LIBRARY_DEBUG})
+else()
+    message (FATAL_ERROR "Could not find Coin3D library in the LibPack")
+endif()
+set (COIN3D_FOUND TRUE)
```
Neumann-A, line 65: same as xerces. Also provide cmake config files.

src/Gui/3Dconnexion/GuiNativeEventWin32.h
```diff
@@ -62,7 +62,11 @@ namespace Gui
 
     private:
         bool InitializeRawInput(HWND hwndTarget);
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
         static bool RawInputEventFilter(void* msg, long* result);
+#else
+        static bool RawInputEventFilter(void* msg, qintptr* result);
+#endif
```
Neumann-A, line 69: The other platforms also need to be adjusted -> GuiNativeEventLinuxX11

src/Gui/MainWindow.cpp
```diff
@@ -1705,7 +1705,7 @@ void MainWindow::loadWindowSettings()
     max ? showMaximized() : show();
 
     // make menus and tooltips usable in fullscreen under Windows, see issue #7563
-#if defined(Q_OS_WIN)
+#if defined(Q_OS_WIN) && QT_VERSION < QT_VERSION_CHECK(6,0,0)
```
Neumann-A, line 1708: Qt6 way:
```
#include <qpa/qplatformwindow_p.h>
    using namespace QNativeInterface::Private;
    if (auto *windowsWindow = dynamic_cast<QWindowsWindow*>(this->windowHandle()))
        windowsWindow->setHasBorderInFullScreen(true);
```
and
`target_link_libraries(FreeCADGui ${FreeCADGui_LIBS} Qt6::GuiPrivate)`

maxwxyz:
@chennes not sure if related to the feature freeze but just a heads up here as well

cMake/UseLibPack3.cmake
chennes: It can -- that was just copied from the old LibPack code, dropping it works just fine.

cMake/UseLibPack3.cmake
chennes: Copied from the old LibPack -- I've modified as you suggested.

cMake/UseLibPack3.cmake
chennes: Must be vestigial -- replacing with find_package().

cMake/UseLibPack3.cmake
chennes: Must be vestigial -- replacing with find_package().

cMake/UseLibPack3.cmake
chennes: Yes: I've modified to set the paths as part of find_package().

src/Gui/3Dconnexion/GuiNativeEventWin32.h
chennes: This has been resolved elsewhere for the other platforms, which (as of now) do not directly include a RawInputEventFilter method.

src/Gui/MainWindow.cpp
chennes: Thanks, changed.

chennes:
@oursland and @adrianinsaval I'm getting a CI build failure that I don't understand -- it's the same error on MacOS and Ubuntu 22.04 --
```
CMake Error at .conda/freecad/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Boost: Found unsuitable version "1.82.0", but required is
  exact version "1.85.0" (found
  /Users/runner/work/FreeCAD/FreeCAD/.conda/freecad/include, )
Call Stack (most recent call first):
  .conda/freecad/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:598 (_FPHSA_FAILURE_MESSAGE)
  .conda/freecad/share/cmake-3.28/Modules/FindBoost.cmake:2393 (find_package_handle_standard_args)
  .conda/freecad/lib/cmake/vtk-9.2/VTK-vtk-module-find-packages.cmake:491 (find_package)
  .conda/freecad/lib/cmake/vtk-9.2/vtk-config.cmake:152 (include)
  cMake/FreeCAD_Helpers/SetupSalomeSMESH.cmake:29 (find_package)
  CMakeLists.txt:72 (SetupSalomeSMESH)
```
I don't understand where the requirement for Boost 1.85 is coming from in VTK 9.2. Any ideas?

CMakeLists.txt
```diff
@@ -68,8 +68,8 @@ if(NOT FREECAD_LIBPACK_USE OR FREECAD_LIBPACK_CHECKFILE_CLBUNDLER)
         # OPENGL_gl_LIBRARY empty that results into linker errors
         SetupOpenGL()
     endif(BUILD_GUI)
-    SetupSalomeSMESH()
     SetupBoost()
```
oursland, line 71: Possibly this line re-order is causing the issue.  I do not see where the specific requirement for Boost 1.85.0 is made.  The dependencies between the failed build and recent successful builds on `main` match, so the issue is not due to a dependency upgrade.

SMESH is built with specific versions of Boost as a requirement.  The latest build of SMESH requires Boost 1.84.0 as a dependency within conda-forge.  (ref: [smesh-feedstock](https://github.com/conda-forge/smesh-feedstock/commits/main/))

chennes: I made the reordering to try to fix this issue (which it did not do, obviously).

cMake/FreeCAD_Helpers/SetupSalomeSMESH.cmake
```diff
@@ -26,7 +26,7 @@ macro(SetupSalomeSMESH)
 
         # check which modules are available
         if(UNIX OR WIN32)
-            find_package(VTK COMPONENTS vtkCommonCore REQUIRED NO_MODULE)
+            find_package(VTK REQUIRED NO_MODULE)
```
oursland, line 29: This is the culprit preventing builds on conda based platforms.

The file `.conda/freecad/lib/cmake/vtk-9.2/VTK-vtk-module-find-packages.cmake` has the following stanza:

    if (_vtk_module_find_package_enabled)
      set(_vtk_module_find_package_required)
      if (_vtk_module_find_package_is_required)
        set(_vtk_module_find_package_required REQUIRED)
      endif ()

      find_package(Boost
        1.85.0
        EXACT

        ${_vtk_module_find_package_quiet}
        ${_vtk_module_find_package_required}
        COMPONENTS
        OPTIONAL_COMPONENTS )
      if (NOT Boost_FOUND AND _vtk_module_find_package_fail_if_not_found)
        if (NOT ${CMAKE_FIND_PACKAGE_NAME}_FIND_QUIETLY)
          message(STATUS
            "Could not find the ${CMAKE_FIND_PACKAGE_NAME} package due to a "
            "missing dependency: Boost")
        endif ()
        set("${CMAKE_FIND_PACKAGE_NAME}_xdmf3_FOUND" 0)
        list(APPEND "${CMAKE_FIND_PACKAGE_NAME}_xdmf3_NOT_FOUND_MESSAGE"
          "Failed to find the Boost package.")
      endif ()
    endif ()

This chunk of code is evaluated when `VTK` is imported, but not when a subset of the components are imported.  It looks like the VTK component `vtkxdmfcore` is dependent upon Boost and may force this evaluation.

chennes: That's what I get for trying to clean up the VTK warning about renaming CommonCore! I can just roll that change back, it's not important. Thanks!

Jookia:
This builds on Ubuntu 22.04 with Qt5 and Arch Linux with Qt6. Re-running CMake, it's able to find PySide6 and Shiboken6 fine.

3x380V:
@chennes I'm sorry for not reading through this earlier...
The first commit moves `find_pip_package` macro while changing its output from `${PACKAGE}_LIBRARY` to `${PACKAGE}_LIBRARIES` (also renames includes). Unfortunately the callsites in `SetupShibokenAndPyside.cmake` are unchanged, so builds against pip installed PySide and Shiboken are broken. This fix is obvious, however internal logic of finding libraries inside `find_pip_package` changed as well and does not work on Linux (empty string is returned) and I do not see an intention. Could you help here?
A side note: There are shinny new `FindShiboken6.cmake` and `FindPySide6.cmake`, but I do not see them used anywhere...

chennes:
Can you make an Issue? Our discussion will get lost here now that this is merged.

Issue #10338, opened 2023-08-24
[Problem] 'Part' Object is confusing.


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
0.22 dev
```

### Subproject(s) affected?

Core

### Problem description

This has started to be discussed in this issue : https://github.com/FreeCAD/FreeCAD/issues/9278

The problem is that the Std_Part tool that create 'Part' objects is confusing for a lot of users. Many people are confused between this 'Part' and partDesign 'Body'. Besides the name does not carry out the meaning of the object. If we look at the wiki https://wiki.freecad.org/Std_Part  : 

> The Std Part element was developed to be the basic building block to create mechanical [assemblies](https://wiki.freecad.org/Assembly).

So how did we end up with the confusing name 'Part' that makes no sense?

@Zolko-123 mentioned 2 possible usecases : 

> well, I see 2 use-cases, but I'm not sure how relevant they are:
> 
> 1 - parts using booleans, where you design 2 shapes and subtract (or add or intersect) them together: in this case, you can use a Std_Part container for the final result, which is not an assembly but a complex yet standalone part
> 2 - for surface functions, where you design 3D surfaces and cut and intersect them, producing a Std_Part which is a general purpose 3D container

But I don't think those make enough sense to validate keeping the name Part. For case 1 I fail to see how you need a container for booleans as they already results in single tree entities. For case 2, it could be called an assembly of surfaces rather than a Part it would make more sense.

Anyone see reason why we shouldn't rename 'Part' object to 'Assembly'?

I also suggest changing the icon so that the big block and small block do not share the same color. To further differentiate with PartDesign body and show in a clearer fashion the fact that it's supposed to be an assembly.
![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/9c4b4908-6400-40e3-ae9b-916fa5820eca)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Zolko-123:
> The problem is that the Std_Part is confusing for a lot of users. Many people are confused between this and part design body

in what occasions does a user see the name "_Std_Part_ " ? And what has that to do with the confusion between _Part_ and _PartDesign_ WBs ? 

> Anyone see reason why we shouldn't rename it ?

rename _what_ to _what_ ?

[...]

Issue #9748, opened 2023-06-08
[Problem] PartDesign Body container is a confusing concept (with unfitting name)


adrianinsaval:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
not relevant
```

### Subproject(s) affected?

PartDesign

### Problem description

related to #9747

probably a poor choice of word from early development of PD WB, this name carries unfortunate connotations from common language and from it's use in other software in practice our body container behaves more like what is commonly called a Part in other software, this leads to confusion for many users that start their projects by creating a Part container then get confused when a command asks them to create a body to start modelling, the distinction is not immediately clear. It has also been confusing to some on their usage in assembly WBs (it was initially not supported on asm4 and up to this day some people believe they need to put the body inside a Part to use it).

The solution I favor is of dropping the body container in favor of using a customized Part container, similar to what is described in https://github.com/FreeCAD/FreeCAD/issues/9278#issuecomment-1509899507 for assemblies. The container would then be a variation of the Part container and offer greater flexibility, perhaps with a subgroup containing the cumulative/sequential features and potentially allowing other objects or separate feature histories within the same container (and LCS). A downside of this approach is that it would require a migration script for old files and would break forwards compatibility.

This would have to be evaluated when work on #9747 starts as it would potentially make this concept more confusing which made some prominent members of the community uneasy about such change.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
In another issue I also mentioned another possibility to alleviate this:

> in Part Design the term Body does not fit right and carries some unfortunate expectations from some users (single solid rule) that were needlessly added to the software as limitations, but we can't easily use the Part name because it's already used here... So I'm thinking perhaps we can call the current Part container just Container, that's very generic but still fits IMO.

Might be simpler to implement in software but might invalidate lots of documentation/tutorials and lead to great online confusion. But it's probably a decision we should make before a 1.0 release.

grubertechnic:
> in practice our body container behaves more like what is commonly called a Part in other software

No, it is different. The Part container in FreeCAD is that was in other Software used to be a Part. A Part most contains much more as only (psychical) Bodies. Just said that a Part can also have valid more the one Bodies (roller bearings, locking assemblies, shrink discs, compounds, welded parts.....).

[...]

PR #10342, opened 2023-08-24
Core: Rename Std_Part to Std_Assembly

PaddleStroke:
Fixes https://github.com/FreeCAD/FreeCAD/issues/10338
Provides a proper tooltip.
And paves the way for the integrated Assembly workbench.

Here is how it looks : 
image

Details : 
- The command Std_Part has been renamed Std_Assembly. 
- The cpp class StdCmdPart is renamed StdCmdAssembly
- The object name in the tree is now Assembly.
- The icon is modified to make it clearer that it's 2 separate solids in the icon.

Note : 
- It does not changes the internal "App::Part" name. I think it would be best if we rename this as well, but I fear that this is going to break a lot of things. Thoughts ?

People who might want to weight in : @adrianinsaval @chennes @Zolko-123 @sliptonic

### DIFF
--- a/src/Gui/CommandStructure.cpp
+++ b/src/Gui/CommandStructure.cpp
@@ -42,28 +42,31 @@ using namespace Gui;
 //===========================================================================
 // Std_Part
 //===========================================================================
-DEF_STD_CMD_A(StdCmdPart)
+DEF_STD_CMD_A(StdCmdAssembly)
 
-StdCmdPart::StdCmdPart()
-  : Command("Std_Part")
+StdCmdAssembly::StdCmdAssembly()
+  : Command("Std_Assembly")
 {
     sGroup        = "Structure";
-    sMenuText     = QT_TR_NOOP("Create part");
-    sToolTipText  = QT_TR_NOOP("Create a new part and make it active");
+    sMenuText     = QT_TR_NOOP("Create assembly");
+    sToolTipText = QT_TR_NOOP("Create a new assembly and make it active.\n"
+        "An assembly is meant to arrange objects like Part Primitives,\n"
+        "PartDesign Bodies, and other Assemblies. It provides an Origin object,\n"
+        "that can be used as reference to position the contained objects.");
     sWhatsThis    = "Std_Part";
     sStatusTip    = sToolTipText;
     sPixmap       = "Geofeaturegroup";
 }
 
-void StdCmdPart::activated(int iMsg)
+void StdCmdAssembly::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
 
-    openCommand(QT_TRANSLATE_NOOP("Command", "Add a part"));
-    std::string FeatName = getUniqueObjectName("Part");
+    openCommand(QT_TRANSLATE_NOOP("Command", "Add an assembly"));
+    std::string FeatName = getUniqueObjectName("Assembly");
 
     std::string PartName;
-    PartName = getUniqueObjectName("Part");
+    PartName = getUniqueObjectName("Assembly");
     doCommand(Doc,"App.activeDocument().Tip = App.activeDocument().addObject('App::Part','%s')",PartName.c_str());
     // TODO We really must set label ourselves? (2015-08-17, Fat-Zer)
     doCommand(Doc,"App.activeDocument().%s.Label = '%s'", PartName.c_str(),
@@ -75,7 +78,7 @@ void StdCmdPart::activated(int iMsg)
     updateActive();
 }
 
-bool StdCmdPart::isActive()
+bool StdCmdAssembly::isActive()
 {
     return hasActiveDocument();
 }
@@ -128,7 +131,7 @@ void CreateStructureCommands()
 {
     CommandManager &rcCmdMgr = Application::Instance->commandManager();
 
-    rcCmdMgr.addCommand(new StdCmdPart());
+    rcCmdMgr.addCommand(new StdCmdAssembly());
     rcCmdMgr.addCommand(new StdCmdGroup());
 }
 
--- a/src/Gui/Icons/Geofeaturegroup.svg
+++ b/src/Gui/Icons/Geofeaturegroup.svg
@@ -1,20 +1,20 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-   xmlns:svg="http://www.w3.org/2000/svg"
-   xmlns="http://www.w3.org/2000/svg"
-   xmlns:xlink="http://www.w3.org/1999/xlink"
-   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
-   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    width="64"
    height="64"
    version="1.1"
    viewBox="0 0 64 64"
    id="svg96"
    sodipodi:docname="Geofeaturegroup.svg"
-   inkscape:version="0.92.5 (2060ec1f9f, 2020-04-08)">
+   inkscape:version="1.1-beta1 (77e7b44db3, 2021-03-28)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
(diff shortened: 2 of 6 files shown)


Zolko-123:
This is in open violation of the agreed C4 developer procedure.

Also, can you please explain why we should even consider listening to your proposals concerning assemblies ? As far as I'm aware, you haven't shown any competence in assemblies, be it to build real-world assembly CAD models or having contributed to existing assembly WBs.

grubertechnic:
TBH, this is a dully issue and PR. This make no sense in a engineering, PDM, PLC and ERP and in real life work way. A part is no assembly (and in addition, a Body is also no part, see #9748) and many forum threads.

Also this was discussion out in many threads, making new side channels to discuss it alone to make it semi hidden is annoying and not helpful in any way.

PaddleStroke:
Stop raging @Zolko-123 please. Why should we listen to you when all you do is rage at everything and everyone all the time ? You're an angry kid and you have no arguments. It's not because you write something **IN CAPITALS AND IN BOLD** that it makes it true.

Please give some real argument about why the 'Part' terminology makes sense and is better than the offered 'Assembly' here. The only reason I see is that it does not goes the way you chose when you made A4.

You offered 2 'potential' usecase for a 'Part' container that would not be 'assemblies' but your examples are not correct. I note you did not reply to my comment about your example. Instead you just wrote that it makes no sense fixing terminology here if we don't fix it everywhere at once. Which makes no sense.

PaddleStroke:
@grubertechnic I don't understand what you mean. A 'Part' object as it is defined now is not a part in the common sense. It's a container for multiple parts. The wiki says very clearly that it is intended to be an assembly object. Please explain how the current 'Part' name makes any sense.

One my side let me compile you a list of forum post of people confused about what is this Std_Part tool.

sasobadovinac:
Assemblies are build from Parts... Nothing is wrong with the Part name or its functionality, this was discussed and agreed in the past, it is the users (and developers) that lack the knowledge about it and are pushing for unreasonable changes while ignoring the requests to make this work as it should. It is the Body that needs to be fixed not the Part.

FlachyJoe:
What about **Positioned Group**? This thing is not an assembly until some constraints are added.
This PR is too early as consensus is not found yet. Should be closed.

PaddleStroke:
> Assemblies are build from Parts...

This is correct in the conceptual sense of the word 'part'. But Std_Part is not a part in the sense of a solid designed by 'Part Design' or by Part wb. No Std_Part is a container. So exactly an assembly of solids.
So when you say 'Assemblies are build from Std_parts' you are saying is 'Assemblies are build from assemblies' which is totally wrong. You can build assemblies only from PartDesign bodies without ever using std_Part.

FlachyJoe:
> You can build assemblies only from PartDesign bodies without ever using std_Part.

So where is the reason to name Std_Part an Assembly?

PaddleStroke:
> So where is the reason to name Std_Part an Assembly?

Reason 1 : Part name is confusing and wrong.
Reason 2 : We are trying to add an integrated assembly wb to freecad and it needs an assembly object. And it makes sense to use this already existing std_part object as it is exactly an assembly object...

sasobadovinac:
Again, you clearly don't understand it and are trying to create a story for yourself that would make some sort of sense to you... This things avre very clearly defined in the industry, even when implementations in different cad systems are maybe somewhat differed they all work in the same way and comply with standards.

Zolko-123:
> Std_Part is not a part in the sense of a solid designed by 'Part Design' or by Part wb. No Std_Part is a container.

this is getting ridiculous: Std_Part is not an object, it's a command, a function :

```
>>> "Std_Part" in Gui.listCommands()
True
```

What `Std_Part` does is to create an `App::Part` object. Are you really so ignorant in FreeCAD code ?

PaddleStroke:
Zolko raging again and trying to discredit others based on nothing. Again your argument are empty and only aims at discrediting others. I know that std_part is the command [as I said in the issue](https://github.com/FreeCAD/FreeCAD/issues/10338#issuecomment-1691168210).

You all keep raging but none of you have any argument explaining why the term Part makes sense for a container. You can attacking me personally but none of you is giving real arguments (not to say that there isn't). You say I'm ignorant of industry standards, but why?
Is SolidWorks, Catia, Fusion... all using the term 'Part' for a container in which you can put several solids ?
Just give something real besides useless and pathetic personal attacks...

FlachyJoe:
> none of you have any argument explaining why the term Part makes sense for a container

I agree but Assembly is not better term IMHO.

sasobadovinac:
> Is SolidWorks, Catia, Fusion... all using the term 'Part' for a container in which you can put several solids ?

Yes they are. Look  I am sorry that this sort of hits on you, I have nothing against you, and for me you seem capable and inteligent and willing to learn. But you have to understand that some of us have spend literary years if not some even more then a decade researching this things and now some (I see this originally actually coming from others not you, so you are sort of now in this PR just a "messenger") try to push some changes that are clearly based on very little understanding of this things are are just looking for some quick solutions to complex problems. 

I don't want to be explain all this again here in some PR comments, if people are serious on understanding and solving this then we should create a workgroup of of knowledgeable users on this topic and do some real debate and review of this things.

Zolko-123:
> I wrongly said std_part to designate the App::Part

out of curiosity : you propose that the function `Std_Part` that creates an `App::Part` object be renamed to `Std_Assembly` to create the same `App::Part` object, and somehow this is going to be less confusing ? 

> none of you have any argument explaining why the term Part makes sense for a container

classical logical paradox : **YOU** want a change therefore **YOU** need to come up with **GOOD** arguments to justify the change.

PaddleStroke:
The name 'Part' would potentially make sense if PartDesign was using them instead of separate but semantically identical 'body' object. And if Part wb was using them as well to store the primitives.

Then I would understand that the Part container would make sense. You make Parts with either Part wb or PartDesign, then make an assembly with them. And potentially you can nest parts to make 'multi-body parts' without making an assembly.

But this is not the case today. So if you look at the current workflow of someone using PartDesign, for him the parts he creates are Bodies. Then for him the App::Part object doesn't make sense and is redundant.

sasobadovinac:
> The name 'Part' would potentially make sense if PartDesign was using them instead of separate but semantically identical 'body' object. And if Part wb was using them as well to store the primitives.

Yes, this is how it should be, everything should be able to use Part to put in there its geometry (even mesh if you want so) and it is why I said that it is the Body that needs to be fixed not the Part, and I said and explained that in detail back before PartDesign was merged in to master, and I said that without the needed correction it will open and endless confusion and endless workaround fixes that will actually only make things worse, I call it the "wheels on the roof" problem. But no one really bothered to understand it and was just pushing for the merge...

adrianinsaval:
As I had understood it the path we had agreed upon was that an assembly wb would have a new command that would create a part of type assembly (see how asm4 does it).

Not remove/rename the current Part command. Personally I don't really oppose this though.


Please stop the personal attacks as they contribute nothing to the discussion.

PaddleStroke:
Ok I see now @sasobadovinac . So yes the real issue is https://github.com/FreeCAD/FreeCAD/issues/9748
But until this is fixed the App::Part is confusing.
Anyway let's close this then.

sasobadovinac:
There are also other important things to understand to be able to full see why things are today in FC as they are and what should be the correct path forward. For example the traditional way in parametric cad is to have separate assembly and part "containers" and them being directly associated to files (.asm and .prt files types for example). But over time, with some newer technologies becoming more common, for example with moving from file based to database product management has become clear that some of this traditional "way" are actually more just unneeded "restrictions" that existed before because of the things where done and not because they where really strictly needed for the proper system architecture. So it was discussed and in some way decided (I say in some way because no final implementation was actually realized up to now), so that we could have a single file for the full assembly (in this way the file acts more like a database), but with the link functionality those users that want still can make the more traditional file structure (each part and assembly in its own file), so we gain flexibility but don't loose "traditional" functionality. A similar discussion was with Part/Assembly container, traditionally the are two different, but their core functionality from a technical perspective is actually the same and even in the "old documentations" of traditional cad systems this is acknowledged, while some, foe example NX have moved forward and removed this and this was also the basis when we discussed and decided (mostly in de development of Asm4 with Zolko) that this would be good also for FC, since we again only gain on the flexibility in this way and are again also not loosing on the "traditional" functionality. Original plan, back in 2012 (or even before?) when Part container was actually first implemented it was to go more in the old "traditional" way. Another thing to understand is that a lot of FC was originally "inspired" more by "Catia" then any other cad, so to someone who understand Catia everything in FC makes much more sense. etc

prokoudine:
> This thing is not an assembly until some constraints are added.

I don't understand this argument. If the user specifically chooses to create an assembly, it's safe to assume that they will add some constraints later on. You don't really expect the container to be renamed to 'assembly' once the first constraint is added, do you?

sasobadovinac:
> > This thing is not an assembly until some constraints are added.
> 
> I don't understand this argument. If the user specifically chooses to create an assembly, it's safe to assume that they will add some constraints later on. You don't really expect the container to be renamed to 'assembly' once the first constraint is added, do you?

True, it is a wrong argument, and it is also the primary reason why most of the existing Asm developments have failed, because they were focused on that. Assembly/Product structure is primary build from the Part/Asm containers, geometry and constraints that go in to them (or between) then bring context and functionality in to it.

FOSCODER:
I use part a lot for attachment and this would break all my files.

PR #10365, opened 2023-08-25
Wrap PyArg_ParseTupleWithKeywords

chennes:
The CPython function `PyArg_ParseTupleWithKeywords` signature is
```
int PyArg_ParseTupleAndKeywords(
    PyObject *args, 
    PyObject *kw, 
    const char *format, 
    char *keywords[], 
    ...)
```
This poses two issues for C++: first, the keywords array is a non-const `char *`, but is always created as a list of string literals in the code. This necessitates that we use a `const_cast<char **>` if compiling in strict-compliance mode -- `const_cast` is usually a bad idea, and clang-tidy will (and should) alert us to its uses, of which there would be hundreds.

A second, more minor, issue is that although the C function returns an int, it is really a boolean value, and we treat it as such. This causes clang-tidy to warn about implicit conversions from int to bool, which is most easily addressed via a wrapper that attempts to hide the C function call as much as possible, and provide a more C++-like interface to it.

This PR first adds a C++ wrapper, `Wrapped_ParseTupleWithKeywords` that takes a `const std::array<const char *,N>` for the keywords list. This allows us to silence clang-tidy warnings about use of C-style arrays when constructing the list, but necessitates a templated function to adapt to the varying array sizes. It allows the use of string literals in the list construction without issue, and encapsulates the single `const_cast<char**>` that is required in a single location.

It also adapts the integer return type from the C function to a true C++ boolean, to eliminate the clang-tidy warnings about implicit conversion.

All of this done, however, a single clang-tidy warning remains on all calls to the function: it still uses the C-style variadic parameters, and the clang-tidy flag `cppcoreguidelines-pro-type-vararg` causes this to be flagged on each call.


src/Base/PyWrapParseTupleAndKeywords.h
```diff
@@ -0,0 +1,70 @@
...
+    /// \arg (variadic) Pointers to the storage locations for the parameters
+    /// \returns boolean true on success, or false on failure
+    template<size_t arraySize>
+    bool Wrapped_ParseTupleAndKeywords(PyObject *args,
+                                       PyObject *kw,
+                                       const char *format,
+                                       const std::array<const char *, arraySize> keywords,
+                                       ...)
+    {
+        // NOTE: This code is from getargs.c in the Python source code (modified to use the public interface at
```
wwmayer, line 47: Maybe use a `static_assert(arraySize > 0, "Array must at least have a single element");` to guarantee that no empty array is passed.

chennes: Good idea, I've done so now. I'd love to also do a compile-time check on the last argument being a `nullptr`, but my metaprogramming skills are too weak.

wwmayer: > I'd love to also do a compile-time check on the last array entry being a nullptr, but my metaprogramming skills are too weak.

I don't think this is possible because how could the function make guarantees at compile time about the std::array it gets from elsewhere? All what you can do is to make this guarantee in the calling instance:

```
constexpr std::array<const char *, 3> keywords = {"1", "2", nullptr};
static_assert(keywords.back() == nullptr, "Last element must be a nullptr");

Wrapped_ParseTupleAndKeywords(args, kwds, "|i", keywords, &compression)
...
```
But having the static_assert at the some scope where the std::array is created is pointless, IMO.

Actually, it is sufficient that this check is done at runtime and in case it fails you should set a Python exception and return false. This way the exception handling of the calling instances doesn't need to be changed.
```
bool Wrapped_ParseTupleAndKeywords(PyObject *args,
                                       PyObject *kw,
                                       const char *format,
                                       const std::array<const char *, arraySize> keywords,
                                       ...)
{
...
    if (keywords.back()) {
        PyErr_SetString(PyExc_ValueError, "Last element of keywords array is not null");
        return false;
    }
...
}
```

chennes: > I don't think this is possible because how could the function make guarantees at compile time about the std::array it gets from elsewhere?

You're right -- metaprogramming is definitely not my strength!

> But having the static_assert at the some scope where the std::array is created is pointless, IMO.

Agree completely.

I've added a commit that implements your suggestion for making the error message better at runtime.

PR #10366, opened 2023-08-25
Test: add more unit tests

wwmayer:
Add unit tests of several more core classes


tests/src/Base/BoundBox.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+    Base::BoundBox3d box1, box2, box3;
+    EXPECT_EQ(box1.Intersect(box1), false);
+    box1.Add(Base::Vector3d(0, 0, 0));
+    EXPECT_EQ(box1.Intersect(box1), true);
+    EXPECT_EQ(box1.Intersect(box2), false);
+    EXPECT_EQ(box2.Intersect(box1), false);
+
+    box1.Add(Base::Vector3d(1, 1, 1));
+    box2.Add(Base::Vector3d(0.5, 0.5, 0.5));
```
chennes, line 128: I've been adding `// NOLINTBEGIN(readability-magic-numbers)` (and corresponding NOLINTEND) in the test files I work on, since tests often have these.

PR #10367, opened 2023-08-25
Core: escape filepath characters in FreeCAD.loadFile

bdieterm:
backport of #10278 
----
Currently, executing the Python command `FreeCAD.loadFile("/tmp/let's go.FCStd")` will fail because of the single quote sign in the path.

This pull request fixes the problem by properly escaping the special characters in the string.

The error was described in #10264 as happening when "opening from recent history". The root cause lies in function `FreeCAD.loadFile` and therefore the error also happens on FreeCAD version 0.20.2 when calling that function directly instead of via the history action.


chennes:
Thanks for the backport PR, it's much appreciated!

PR #10377, opened 2023-08-26
PartDesign: Add offset modes for patterns

kadet1090:
This PR adds support for providing not only overall length or angle for patterns like it works now, but also allows user to specify exact length or angle for spacing consecutive features. It's based on previous work of @sugol-a at https://github.com/sugol-a/FreeCAD/tree/lp-offset which I rebased and build upon.

Related forum topic: https://forum.freecad.org/viewtopic.php?p=596742 - I don't know If it is required to create issue for the PR - if it is, I'll make one based on that topic.

image
image

As for the nomenclature I chose to stick to **Offset**  and **Overall Length / Angle** for the mode names, but I'm not particularly proud of them. I can't however think of clearly better alternatives. Fusion 360 uses **"Spacing"** and **"Extent"**, but I don't think that they are obvious without corresponding icons. Few choices that I thought of for now:
- **Offset** / **Overall Length**
- **Space** / **Distribute evenly**
- **Offset** / **Length** (that was in original branch)

Possible issues:
- When `offset angle * occurrences > full circle` it is possible that features will overlap exactly and cause some trouble. I think that this is not a bug rather problem that should be solved by the user if it happens.
- I changed angle precision check to be done after calculating offset - not before. Potentially it could be breaking change, but If resulting offset is to small it should be probably fixed none the less in the design.
- As each field is displayed only for specific mode I've wrapped field with label with additional `QWidget` (QT Designer -> Layout -> Morph Into -> QWidget` so it could be hidden with only one line of code, ensuring that there will be no situation when only label or field is visible. I don't know if this is valid solution or maybe there are better possibilities - if so, I'll gladly change that part. 

For LinearPatterns `Length` and `Offset` properties are kept in sync to express same meaning. Corresponding value is updated only if values are set explicitly (i.e. no expression). I didn't implement such behavior for PolarPattern because **Overall Angle** is limited to full circle, whereas **Offset** is not. If we decide to add same constraint to Offset or remove maximum value from Overall Angle I'll implement analogous feature for the PolarPattern.

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskLinearPatternParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskLinearPatternParameters.cpp
@@ -198,7 +198,6 @@ void TaskLinearPatternParameters::setupUI()
     }
 
     adaptVisibilityToMode();
-    updateUI();
     connectSignals();
 }
 
@@ -241,6 +240,8 @@ void TaskLinearPatternParameters::adaptVisibilityToMode()
 
     ui->lengthWrapper->setVisible(mode == PartDesign::LinearPatternMode::length);
     ui->offsetWrapper->setVisible(mode == PartDesign::LinearPatternMode::offset);
+
+    updateUI();
 }
 
 void TaskLinearPatternParameters::onUpdateViewTimer()
--- a/src/Mod/PartDesign/Gui/TaskLinearPatternParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskLinearPatternParameters.cpp
@@ -127,8 +127,12 @@ void TaskLinearPatternParameters::connectSignals()
             this, &TaskLinearPatternParameters::onDirectionChanged);
     connect(ui->checkReverse, &QCheckBox::toggled,
             this, &TaskLinearPatternParameters::onCheckReverse);
+    connect(ui->comboMode, qOverload<int>(&QComboBox::activated),
+            this, &TaskLinearPatternParameters::onModeChanged);
     connect(ui->spinLength, qOverload<double>(&Gui::QuantitySpinBox::valueChanged),
             this, &TaskLinearPatternParameters::onLength);
+    connect(ui->spinOffset, qOverload<double>(&Gui::QuantitySpinBox::valueChanged),
+            this, &TaskLinearPatternParameters::onOffset);
     connect(ui->spinOccurrences, &Gui::UIntSpinBox::unsignedChanged,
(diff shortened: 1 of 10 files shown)


src/Mod/PartDesign/Gui/TaskLinearPatternParameters.cpp
```diff
@@ -210,15 +222,28 @@ void TaskLinearPatternParameters::updateUI()
         dirLinks.setCurrentLink(pcLinearPattern->Direction);
     }
 
-    // Note: These three lines would trigger onLength(), on Occurrences() and another updateUI() if we
+    // Note: These three lines would trigger related on*() and another updateUI() if we
```
luzpaz, line 225: Is this sentence correct ?

kadet1090: I'm not sure anymore ;) I've changed this comment to:
```
    // Note: This block of code would trigger change signal handlers (e.g. onOccurrences())
    // and another updateUI() if we didn't check for blockUpdate
```

Now it should be clear and easy to extend as it does not explicitly mention triggered methods.

chennes:
Thanks for the PR, @kadet1090 -- as a general rule we do ask contributors to submit a PR as a response to a discussion in a GitHub Issue, but that's a new process and we try not to be too pedantic about it right now 😄 . I've enabled the CI on this PR, we should have those results shortly. I appreciate you taking the time to port this over from @sugol-a's work.

kadet1090:
Thanks! 

> as a general rule we do ask contributors to submit a PR as a response to a discussion in a GitHub Issue

I think that this should be included in the  `Contribution Requirements`  section of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md). This requirement makes perfect sense, but IMHO it is not stated clearly enough. Most projects have clear distinction between fixes / code improvements (which does not necessarily require Issue) and features (which do require one) and it is clearly stated in the that every PR that introduces new feature MUST (or at least SHOULD) have related Issue where feature could be tracked and discussed. I know that "Process" section mentions it, but for me it seemed more like a way for users to create feature requests that contributors can later implement, and so I was not sure if this applies If I'm basically proposing changes not requesting them.

chennes:
This is a nice feature to have, I think it makes sense to merge it as-is to get it into beta testers hands. We can always tweak the wording or behavior based on their feedback. Thanks!

PR #10387, opened 2023-08-27
[Stylesheets] Replaced more static colors with the accent colors.

MisterMakerNL:
As the title says I replaced some static colors with the accent colors, also for ProDark and Behave-Dark and Dark-contrast.
I also added gradient to the modern styles, as you can now manually disable gradients by making the 2 colors the same.

We might need to look into some "reset to default" option for the theme's.

afbeelding
afbeelding
afbeelding
afbeelding


MisterMakerNL:
It's probably wise to add the default colors too the theme/preference pack cfg file, so when applying the theme it resets the colors.

PaddleStroke:
Yes you are correct. Accent colors should be defined in the themes to get an appropriate color

PR #10392, opened 2023-08-27
Improve Pad/Pocket Task Dialog

geolta:
Based on the discussion here https://forum.freecad.org/viewtopic.php?t=80413 I made the following changes to the Pad/Pocket task dialog:
- reorder UI elements
- hide "Select Face" if not necessary
- improve tab stop order

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskPadPocketParameters.ui
+++ b/src/Mod/PartDesign/Gui/TaskPadPocketParameters.ui
@@ -16,10 +16,46 @@
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
     <layout class="QGridLayout" name="gridLayout">
+     <item row="0" column="0">
+      <widget class="QLabel" name="textLabel1">
+       <property name="text">
+        <string>Type</string>
+       </property>
+      </widget>
+     </item>
+     <item row="0" column="1">
+      <widget class="QComboBox" name="changeMode">
+       <item>
+        <property name="text">
+         <string>Dimension</string>
+        </property>
+       </item>
+      </widget>
+     </item>
      <item row="1" column="0">
       <widget class="QLabel" name="labelLength">
        <property name="text">
         <string>Length</string>
+       </property>
+      </widget>
+     </item>
+     <item row="1" column="1">
+      <widget class="Gui::PrefQuantitySpinBox" name="lengthEdit">
+       <property name="keyboardTracking">
+        <bool>false</bool>
+       </property>
+       <property name="unit" stdset="0">
+        <string notr="true">mm</string>
+       </property>
+       <property name="minimum">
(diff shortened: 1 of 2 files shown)


src/Mod/PartDesign/Gui/TaskPadPocketParameters.ui
```diff
@@ -16,10 +16,34 @@
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
     <layout class="QGridLayout" name="gridLayout">
-     <item row="0" column="0">
-      <widget class="QLabel" name="textLabel1">
+     <item row="1" column="0">
```
PaddleStroke, line 19: It's not very important but I prefer when the items are sorted correctly in QGridLayouts. Rather than having 1 here, then 2 then 0. But yes this is probably not visible from QtDesigner.

PaddleStroke (commented):
Looks good. Does what is advertized.

obelisk79:
In the future, please ensure you are following the guidelines from the contributing document. It just helps organize and standardize the contribution process. This PR is pretty small so I don't think its a huge issue, but FreeCAD is trying to move forward with this process and it only works if we all follow it.

https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md

chennes:
Thanks for the PR, @geolta! I've enabled the CI on it, so we should get those results soon.

GeneGH:
> In the future, please ensure you are following the guidelines from the contributing document.

It would be useful to point out which portions are wrong or offensive.

src/Mod/PartDesign/Gui/TaskPadPocketParameters.ui
geolta: Should be fixed now (wasn't able to see this in Qt Designer unfortunately)

geolta:
> In the future, please ensure you are following the guidelines from the contributing document. It just helps organize and standardize the contribution process. This PR is pretty small so I don't think its a huge issue, but FreeCAD is trying to move forward with this process and it only works if we all follow it.
> 
> https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md

Yes, thanks for the hint; in theory I know about the contributing process, just thought to handle it a bit loosely due to the small changes :)

(Although it maybe might be a good opportunity to get into the process with this small change...)

chennes:
As a general rule it's good to start out with an Issue, so that when it comes time to actually write code you don't waste time working on something that gets rejected for reasons that could have been worked out before you spent the time. It's sort of a moot point here, there seems to be solid agreement that this is a good change (certainly it makes sense to me), but more complex/controversial UI changes definitely warrant an Issue before you actually get to coding. At any rate, I'm going to merge this one now. Thanks!

PR #10395, opened 2023-08-27
Main: Fix for Windows with Py3.10

wwmayer:
The fix for https://forum.freecad.org/viewtopic.php?t=79192 may have caused a regression with Py3.10 on Windows as reported here https://forum.freecad.org/viewtopic.php?t=80659

This PR activates the previous fix for Linux & UNIX systems only

### DIFF
--- a/src/Main/MainGui.cpp
+++ b/src/Main/MainGui.cpp
@@ -97,8 +97,9 @@ private:
 
 int main( int argc, char ** argv )
 {
-    setlocale(LC_ALL, ""); // use native environment settings
 #if defined (FC_OS_LINUX) || defined(FC_OS_BSD)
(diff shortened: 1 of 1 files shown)


src/Main/MainGui.cpp
```diff
@@ -97,8 +97,9 @@ class Redirection
 
 int main( int argc, char ** argv )
 {
-    setlocale(LC_ALL, ""); // use native environment settings
 #if defined (FC_OS_LINUX) || defined(FC_OS_BSD)
+    setlocale(LC_ALL, ""); // use native environment settings
```
adrianinsaval, line 101: what do we know about this for macOS? should it be included in that case too?

Issue #10221, opened 2023-08-14
StartWB: `/tmp/FreeCADStartThumbnails*` directories are not removed when FreeCAD terminates


berberic2:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
OS: Debian GNU/Linux 12 (bookworm) (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.33802 +2 (Git)
Build type: Release
Branch: Meins
Hash: c70ded0ad5c6dadc86d3731aa0233f291b2839fa
Python 3.11.2, Qt 5.15.8, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * Curves 0.6.13
  * Assembly4 0.50.2
  * Pyramids-and-Polyhedrons
  * sheetmetal 0.2.63
  * Silk 0.1.3
  * CurvedShapes 1.0.4
  * DynamicData 2.46.0
  * FreeCAD_assembly3 0.11.3
  * fcgear 1.0.0
  * ExplodedAssembly
  * A2plus 0.4.60n
  * fasteners 0.4.59
  * Render 2023.2.6
```

### Subproject(s) affected?

None

### Problem description

Whenever FreeCAD starts the `Start` workbench creates a folder in `/tmp`, named `/tmp/FreeCADStartThumbnails⍰⍰⍰⍰⍰⍰⍰`. These never get removed so over time a lot of theses folders pile up in `/tmp`.

The folder should be removed if FreeCAD terminates.

see: https://forum.freecad.org/viewtopic.php?p=697560&sid=c7efd2a32a269cdf1eceb2de29017068

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

luzpaz:
@yorikvanhavre this is Start WB related

xtemp09:
The `/tmp` directory is managed by OS and is automatically cleaned upon system reboot or on schedule. Might I ask why this should be removed when FreeCAD terminates? Is there a rule?

yorikvanhavre:
The fact that these files are not deleted by FreeCAD is useful as they can then be reused by a later session. But then indeed as discussed in #10414 one should put reusable files not in a temp directory but in a cache directory.

I think once things are clearly separated between disposable files (in temp) and cached files (in cache) then this very question stops being one, as we can let the system do garbage treatment in temp.

[...]

PR #10414, opened 2023-08-28
StartPage: purge unnecessary tempfolders (issue #10221)

qewer33:
Fix for https://github.com/FreeCAD/FreeCAD/issues/10221

Works pretty well and shouldn't have any performance issues. But one very specific problem I've noticed:
- Launch FreeCAD
- Launch another instance of FreeCAD
- The second instance removes the tempfolder that the first instance uses
- Error warnings show when the StartPage of the first instance is refreshed because it can't access it's thumbnail files

One idea I have to fix this is making it so that the folders are not purged if there is another currently running instance of FreeCAD. But this would require looping over the user's processes, I'm not sure if that's considered fine or not (or if it's cross platform). Would be glad to hear suggestions on how this could be fixed

### DIFF
--- a/src/Mod/Start/StartPage/StartPage.py
+++ b/src/Mod/Start/StartPage/StartPage.py
@@ -26,6 +26,7 @@
 
 import sys
 import os
+import shutil
 import tempfile
 import time
 import zipfile
@@ -329,10 +330,18 @@ def handle():
     import Start
     if hasattr(Start,"iconbank"):
         iconbank = Start.iconbank
-    if hasattr(Start,"tempfolder"):
+    if hasattr(Start,"tempfolder") and os.path.exists(Start.tempfolder):
         tempfolder = Start.tempfolder
     else:
         tempfolder = tempfile.mkdtemp(prefix="FreeCADStartThumbnails")
+
+    # purge old tempfolders
+
+    for entry in os.scandir(tempfile.gettempdir()):
+        if entry.is_dir():
+            if entry.name.startswith("FreeCADStartThumbnails"):
+                if entry.path != tempfolder:
+                    shutil.rmtree(entry.path)
 
     # build the html page skeleton
 


yorikvanhavre:
That's an interesting case... A solution could be to create unique tempfolders for each instance. But after all, why are we removing the tempfolder at all. The operating system should normally take care of this

qewer33:
> A solution could be to create unique tempfolders for each instance.

That will be a much cleaner and better implementation yes. Is there a Start WB function that runs when the workbench is switched/StartPage tab is closed? Also a callback for something that will run right before FreeCAD itself is terminating. Both should ideally be in Python. I'm not very familiar with how workbench code is structured, would appreciate some tips here

> But after all, why are we removing the tempfolder at all. The operating system should normally take care of this

After some research on this, it seems like in most Linux distros, it does get frequently auto cleaned. But Windows doesn't seem to have such feature enabled by default. So it can be argued both ways, I feel like it's still better for FreeCAD to do it's own housekeeping though.

qewer33:
@Syres916 would be great to hear your opinions/ideas on this

qewer33:
It just occured to me that FreeCAD has a cache directory, wouldn't it make sense for the StartPage to store it's thumbnail and icon banks there instead of /tmp? Similar to the Addon Manager.

image

Syres916:
There's at least one other temp file/folder issue that I'm keeping in mind which involves the creation of the Techdraw template page which at present only applies to certain Windows users and varies depending on the version of Qt.

For me what it boils down to is the inconsistency in the code used for various parts of FreeCAD for temporary file storage and it would be good for there to be just one temporary folder per FC process used by all parts of the app which is cleared down in `Application::destruct()`

It'll be made more complicated because I believe the Python processid lookup doesn't work in MacOS but I'll have a look into it later this week.

Syres916:
@qewer33 Got to the bottom of the Techdraw issue and it wasn't to do with the code so I can forget that. Therefore for me Linux and Windows can use the processid as a unique per session folder identifier and MacOS will have to use the existing function unless there's a method for Mac that I've not seen.

luzpaz:
bumping for activity

yorikvanhavre:
> It just occured to me that FreeCAD has a cache directory, wouldn't it make sense for the StartPage to store it's thumbnail and icon banks there instead of /tmp? Similar to the Addon Manager.

That's a good idea, actually. I understand a cache as something made to be reused, while a temp file is something to be discarded. Indeed that would make much more sense here.

adrianinsaval:
should this be closed now?

yorikvanhavre:
Yes, it is superseded by #10951

PR #10421, opened 2023-08-29
More Matrix test cases

wwmayer:
* Fix many clang-tidy warnings of Matrix class
* Fix incorrect method names about diagonal and trace of a matrix
* Add more unit tests

### DIFF
--- a/src/Base/Matrix.cpp
+++ b/src/Base/Matrix.cpp
@@ -63,9 +63,9 @@ Matrix4D::Matrix4D (double a11, double a12, double a13, double a14,
 {
 }
 
-Matrix4D::Matrix4D (const Matrix4D& rclMtrx) : Matrix4D()
+Matrix4D::Matrix4D (const Matrix4D& mat) : Matrix4D()
 {
-    (*this) = rclMtrx;
+    (*this) = mat;
 }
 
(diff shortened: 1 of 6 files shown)


Pesc0:
Question: why is this class present in freecad? why not use a library like eigen for example?

Pesc0:
@chennes may I ask you the same question?

wwmayer:
What would be the advantage?

Pesc0:
I'm just trying to learn how the codebase is structured, why things are the way they are. 

The advantage of using a library seems obvious to me: 
- dont need to maintain this
- no need to worry about possible bugs
- no need to write tests

the library maintainers take care of all of this. Less code, less bugs, less time "wasted"
So, asking again, is there a specific reason for re-implementing this in freecad when there exist implementations that already did the work?

wwmayer:
> So, asking again, is there a specific reason for re-implementing this in freecad when there exist implementations that already did the work?

Re-implemented? This part of the FreeCAD code is much older than the Eigen3 library. 

> no need to worry about possible bugs

That's wrong. Some years ago there was a regression in the Eigen3 code where things started to become extremely slow.

Also a problem of the Eigen3 library is a certain code bloat because all classes are template classes that usually leads to bigger and slower binaries.

PR #10425, opened 2023-08-29
Draft: Fix Draft_Mirror preview

Roy-043:
When using Draft_Mirror there is currently no preview when the user has picked the 1st point of the mirror line. This PR fixes that.


yorikvanhavre:
Nice!! LGTM!

PR #10428, opened 2023-08-29
[0.21] [Gui] Py311 Initialization of locales...

Syres916:
along with fix to bug in time formatting.
See discussion https://forum.freecad.org/viewtopic.php?t=79192


chennes:
Thanks for submitting a back port PR, I appreciate it.

Issue #10424, opened 2023-08-29
[Problem] Sketcher segfault on latest master


qewer33:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
OS: Fedora Linux 38 (KDE Plasma) (KDE/plasmax11)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34071 (Git)
Build type: Release
Branch: startpage-housekeeping
Hash: 932fd4638ba80671115fb2aa871ab3a273b6f673
Python 3.11.4, Qt 5.15.10, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * Glass
  * lattice2 1.0.0
  * kicadStepUpMod 10.19.4
  * Manipulator 1.5.0
  * Assembly4.backup1691951820.225657 0.50.2
  * Assembly4
  * Behave-Dark-Colors 0.1.1
  * Curves 0.6.13
  * DynamicData 2.46.0
  * fasteners 0.4.59
  * Freecad-Built-in-themes-beta 1.2.2
  * freecad.gears.backup1692223246.5590403 1.0.0
  * freecad.gears
  * ExtremeProLight 1.6.0
  * A2plus 0.4.60n
  * ExplodedAssembly
  * FreeCAD_DarkPreferencePAck 0.0.1
```

### Subproject(s) affected?

Sketcher

### Problem description

1. Open the example file provided below in a latest master build (unzip first):
[PartDesignExample.zip](https://github.com/FreeCAD/FreeCAD/files/12465220/PartDesignExample.zip)

2. Try to open Sketch001 or Sketch003
3. FreeCAD crashes with a segfault

### Anything else?

Terminal output on crash:

<details>
<summary>Terminal Output</summary>
<pre>
FreeCAD 0.22.0, Libs: 0.22.0devR34071 (Git)
© Juergen Riegel, Werner Mayer, Yorik van Havre and others 2001-2023
FreeCAD is free and open-source software licensed under the terms of LGPL2+ license.
FreeCAD wouldn't be possible without FreeCAD community.
  #####                 ####  ###   ####  
  #                    #      # #   #   # 
  #     ##  #### ####  #     #   #  #   # 
  ####  # # #  # #  #  #     #####  #   # 
  #     #   #### ####  #    #     # #   # 
  #     #   #    #     #    #     # #   #  ##  ##  ##
[...]

PR #10429, opened 2023-08-29
Sketcher: fix segfault on distance constraints without SecondPos

bdieterm:
As reported in issue #10424 FreeCAD crashes with a segmentation fault when opening/editing a sketch.

The cause was a change in #9559 that evaluated a geometry of the second constraint position before the check if the second constraint position was valid.
The crash then happend because the second position of a DistanceX constraint was `GeoEnum::GeoUndef` (-2000). This value means that the distance point is the origin of the sketch and not a geometry object.

This PR restores the old order, i.e. the geometry will be evaluated and assigned _after_ the check.
For me, the change of the order in #9559 seems to be a mishap -- I cannot grasp the reasoning behind that change.


chennes:
Thanks for the quick find and fix!

FlachyJoe:
> For me, the change of the order in #9559 seems to be a mishap

@bdieterm you're right it's my fault I didn't check the incidence correctly. Thank for the fix.

PR #10432, opened 2023-08-30
PartDesign: json is a 3rdparty library, move to 3rdparty

ppphp:
move json to 3rdparty folder for better maintenance


wwmayer:
You mean to replace this json lib with RapidJSON? I don't know.

Issue #10075, opened 2023-08-05
3MF files exported from FreeCAD don't work in PrusaSlicer


greggjaskiewicz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
OS: macOS 13.5
Word size of FreeCAD: 64-bit
Version: 0.21.0.33675 (Git)
Build type: Release
Branch: (HEAD detached at 0.21.0)
Hash: 41b058e2087de60dc8fef3d3e68c7d0129e13abf
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: C/Default (C)
Installed mods: 
  * A2plus 0.4.60k
  * sheetmetal 0.2.63
```

### Subproject(s) affected?

None

### Problem description

exciting that 3MF is now main export format. 
But when importing these files into PrusaSlicer (slic3r) - I get message like:
```
The size of the object is zero
Object size from file KentEVs-Kent EV logo.3mf appears to be zero.
This object has been removed from the model
```

I tried the same with multiple models. 

I don't think it's the PrusaSlicer, as it has been out for quite a while. 

Prusa slicer version 2.6.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

greggjaskiewicz:
I just noticed that some files work, others don't.

Attached is the file that does not work.

[Kent_EVs_fails_3mf.zip](https://github.com/FreeCAD/FreeCAD/files/12266710/Kent_EVs_fails_3mf.zip)

luzpaz:
@MisterMakerNL do you mind testing this issue ?

grubertechnic:
Does the silcer allow multiple meshes? Since you have 3 single solids in your model in a part container, they also exported as intersecting meshes. If you do a boolean union the 3 solids (Kent Borders Body, Text Body, base) and then export it, does then the silcer accept it?

greggjaskiewicz:
The whole point of formats like 3MF is to allow multiple solids, unlike STL. You can download prusa slicer and try it yourself, it's open source.

grubertechnic:
Please be explicit, IS multiple solids supported in the slicer? Because this 

> exciting that 3MF is now main export format

is not rue, FreeCAD have no main export format.


> can download

No, have no interest in it, just asking for triaging.

greggjaskiewicz:
[...]

PR #10446, opened 2023-08-30
Mesh: fixes issue #10075: 3MF files exported from FreeCAD don't work …

wwmayer:
…in PrusaSlicer

Add an option to force to always write a mesh as model type even if it's not a solid


khimaros:
are you planning to merge this? this is currently affecting a key workflow and blocking upgrade to 0.21.x releases. @zanovis

wwmayer:
Backported to v0.21: https://github.com/FreeCAD/FreeCAD/commit/cc93256813a0d9ab3f34c16e7521e3959e22570f

PR #10447, opened 2023-08-30
Draft: importDWG.py wrongly assumed that opt directory always exists on Linux

Roy-043:
Forum topic:
https://forum.freecad.org/viewtopic.php?p=703063#p703061


Roy-043:
Should also be backported.

Roy-043:
Merging this now so it can be backported to V0.21.1.

Issue #10443, opened 2023-08-30
[Problem] Sketcher_Dimension doest not allow selection->tool workflow


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
0.22 latest
```

### Subproject(s) affected?

Sketcher

### Problem description

I noticed that the newly introduced tool Sketcher_Dimension does not allow the workflow where you select the geos first, then starts the tool.

There's no reason why it shouldn't support this so I think it should be implemented.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

howie-j:
Yes, i also noticed this. I personally do not use that workflow, but that is how much of Freecad works so it should probably be implemented.

luzpaz:
@PaddleStroke please use `Fixes #10443` next time to auto-close tickets when their PRs get merged, TIA!

PaddleStroke:
Thanks I didn't knew !

luzpaz:
@PaddleStroke :+1:  
JFYI, see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #10449, opened 2023-08-30
Sketcher_Dimension : Implement selection->tool workflow

PaddleStroke:
Edit: It is now working and solves https://github.com/FreeCAD/FreeCAD/issues/10443

Initial text:

> Try to solve https://github.com/FreeCAD/FreeCAD/issues/10443
> However there is an interesting issue.
> I handle the initial selection in the DSH activated function.
> 
> ```
> class DrawSketchHandlerDimension : public DrawSketchHandler
> {
>     void activated() override
>     {
>         ....
>         handleInitialSelection();
>     }
> ```
> 
> It works, you can select a line, trigger the tool, then it creates the correct constraint.
> 
> However it appears that in this case the opened transation is commited immediately after.
> 
> Having a closer look, it does not come from my code. I think the open transaction is automatically closed after the command activated function is finished : 
> 
> ```
> void CmdSketcherDimension::activated(int iMsg)
> {
>     ...
>     ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerDimension(SubNames));
> }
> ```
> So it seems that I open a transaction in ActivateHandler. Then something is closing the transaction after CmdSketcherDimension::activated finishes its execution.
> 
> I also see that the selection is cleared automatically.
> 
> @wwmayer do you know what is happening here?

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -2297,6 +2297,7 @@ CmdSketcherDimension::CmdSketcherDimension()
 void CmdSketcherDimension::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
+    App::AutoTransaction::setEnable(false);
 
     // get the selection
     std::vector<Gui::SelectionObject> selection = getSelection().getSelectionEx();
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -1078,7 +1078,7 @@ public:
 class DrawSketchHandlerDimension : public DrawSketchHandler
 {
 public:
-    DrawSketchHandlerDimension()
+    DrawSketchHandlerDimension(std::vector<std::string>& SubNames)
         : specialConstraint(SpecialConstraint::None)
         , availableConstraint(AvailableConstraint::FIRST)
         , previousOnSketchPos(Base::Vector2d(0.f, 0.f))
@@ -1086,6 +1086,7 @@ public:
         , selLine({})
         , selCircleArc({})
         , selEllipseAndCo({})
+        , initialSelection(SubNames)
         , numberOfConstraintsCreated(0)
     {
     }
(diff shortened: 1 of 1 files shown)


wwmayer:
> do you know what is happening here?

I haven't checked your problem in detail but it sounds like an issue I had several times, too. The problem is caused by the auto-transaction mechanism and you should create a tmp. AutoTransaction instance at the top of your command class. This deactivates the auto-transaction of the calling instance.

So, just add this line: `App::AutoTransaction trans(nullptr);`

PaddleStroke:
Hmm this sounds like a perfect explanation. However adding this line does not change anything.
Maybe I did not add it where you meant : 

```
CmdSketcherDimension::CmdSketcherDimension()
    : Command("Sketcher_Dimension")
{
     //tried to add here too
    sAppModule = "Sketcher";
    sGroup = "Sketcher";
    sMenuText = QT_TR_NOOP("Dimension");
    sToolTipText = QT_TR_NOOP("Constrain contextually based on your selection.\n"
        "Depending on your selection you might have several constraints available. You can cycle through them using SHIFT key.\n"
        "Left clicking on empty space will validate the current constraint. Right clicking or pressing Esc will cancel.");
    sWhatsThis = "Sketcher_Dimension";
    sStatusTip = sToolTipText;
    sPixmap = "Constraint_Dimension";
    sAccel = "D";
    eType = ForEdit;
}

void CmdSketcherDimension::activated(int iMsg)
{
    Q_UNUSED(iMsg);
    App::AutoTransaction trans(nullptr);

    // get the selection
    std::vector<Gui::SelectionObject> selection = getSelection().getSelectionEx();
    std::vector<std::string> SubNames = {};

    // only one sketch with its subelements are allowed to be selected
    if (selection.size() == 1 && selection[0].isObjectTypeOf(Sketcher::SketchObject::getClassTypeId())) {
        SubNames = selection[0].getSubNames();
    }

    // tried here as well.
    ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerDimension(SubNames));
}
```

Is that not where you meant?

wwmayer:
> Maybe I did not add it where you meant :

It's there where I meant. I once did this for the StdCmdRefresh where by default no transaction is created because it makes undo/redo useless.

Then what else you can try is adding this line instead:
`App::AutoTransaction::setEnable(false);`

PR #10452, opened 2023-08-30
GitHub: Better layout and description for Problem template

### LINKED ISSUE #10448: [Problem]Issue display hides description

### Full version info

```shell
any
```

### Subproject(s) affected?

### Problem description

a) When viewing Issues, the description of the problem is hidden on initial presentation and requires scrolling before it is visible.  The description is a more significant piece of information than the data ahead of it in the current template.

b) When creating a new Issue, the text "File an problem report" is too cryptic and does not provide any guidance.  

### Anything else?

Forum discussion: https://forum.freecad.org/viewtopic.php?t=80789

### Code of Conduct

kadet1090:
This updates description of Problem report issue template to be more descriptive and to include what kinds of issues could be reported. Field with description is moved to be second one so it will be rendered on the top of issue instead of near the end.

Fixes #10448

### DIFF
--- a/.github/ISSUE_TEMPLATE/PROBLEM_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/PROBLEM_REPORT.yml
@@ -1,5 +1,5 @@
 name: Report a Problem
-description: File an problem report
+description: Have you found something that does not work well, is too hard to do or is missing altogether? Please create a Problem Report.
 title: "[Problem] "
 labels: ["needs triage"]
 body:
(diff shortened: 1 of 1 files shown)


chennes:
LGTM, but since @WandererFan submitted the Issue I'll leave it to him to comment.

.github/ISSUE_TEMPLATE/PROBLEM_REPORT.yml
```diff
@@ -1,5 +1,5 @@
 name: Report a Problem
-description: File an problem report
+description: Something does not work well, is too hard to do or missing at all? Please report a Problem.
```
WandererFan, line 2: Have you found something that does not work well, is too hard to do or is missing altogether? Please create a Problem Report.

kadet1090: Seems better to me, I've updated the file.

PR #10463, opened 2023-08-31
Sketcher_Dimension : Turn the command into a group-command

PaddleStroke:
Following the introduction of Sketcher_Dimension, many people support adding a dropdown to the new tool to show the individual constraints.

image

@wwmayer I remembered that you told me about GroupCommand before. So I want to use this here but there are some issues : 
- The commands are inactive and nothing seem to activate them.
- The commands don't change of color when you toggle Reference/driving.

Do you know if this can be solved or if I should use the same code as `CmdSketcherCompCreateRectangles` for instance.


wwmayer:
> addCommand(new CmdSketcherDimension());

You create independent copies of these commands and since they are not part of the command manager they cannot be updated. Instead you must override the method createAction() and take the commands from the command manager.

PaddleStroke:
Thanks for the feedback.
Though for reference overriding the createAction as so : 

```
    /*Gui::Action* createAction() {
        auto* pcAction = new Gui::ActionGroup(this, Gui::getMainWindow());
        pcAction->setMenuRole(QAction::NoRole);
        pcAction->setDropDownMenu(true);
        pcAction->setExclusive(false);
        pcAction->setCheckable(true);
        pcAction->setWhatsThis(QString::fromLatin1(sWhatsThis));

        Gui::CommandManager& rcCmdMgr = Gui::Application::Instance->commandManager();

        rcCmdMgr.getCommandByName("Sketcher_Dimension")->addToGroup(pcAction);
        pcAction->addAction(QString::fromLatin1(""))->setSeparator(true);
        rcCmdMgr.getCommandByName("Sketcher_ConstrainLock")->addToGroup(pcAction);
        rcCmdMgr.getCommandByName("Sketcher_ConstrainDistanceX")->addToGroup(pcAction);
        rcCmdMgr.getCommandByName("Sketcher_ConstrainDistanceY")->addToGroup(pcAction);
        rcCmdMgr.getCommandByName("Sketcher_ConstrainDistance")->addToGroup(pcAction);
        rcCmdMgr.getCommandByName("Sketcher_ConstrainDiameter")->addToGroup(pcAction);
        rcCmdMgr.getCommandByName("Sketcher_ConstrainRadius")->addToGroup(pcAction);
        rcCmdMgr.getCommandByName("Sketcher_ConstrainAngle")->addToGroup(pcAction);
        rcCmdMgr.getCommandByName("Sketcher_ConstrainSnellsLaw")->addToGroup(pcAction);

        pcAction->setProperty("defaultAction", QVariant(0));
        setup(pcAction);
        return pcAction;
    }*/
```

Was not working properly. The default action was not set to be the first added command and instead it was the comp command. Until one other command was activated.

But there is also a addCommand that takes in the command name instead of a new command. And there it works properly. I'll just add a 'addSeparator' command.
Then there's also the comp icon that doesn't change of color. I need to investigate a bit more.

PaddleStroke:
Ah well it's almost working now. But somehow the color of the icon is not changing from red to blue even though I added the line 

`    rcCmdMgr.addCommandMode("ToggleDrivingConstraint", "Sketcher_CompDimensionTools");`

Any idea why this is so? If we can solve this we can then simplify all the other sketcher comp like Sketcher_CompConstrainRadDia.

PaddleStroke:
Ok the updateAction function just need to be reimplemented.
So it's ready now.

wwmayer:
Currently, the toolbar button is toggable which doesn't make too much sense because it doesn't have a special meaning like a constraint command is active. So, it's best to switch this off:
```
    Gui::Action * createAction() override
    {
        Gui::Action* act = GroupCommand::createAction();
        act->setCheckable(false);
        return act;
    }
```

BTW: To reduce some extra warnings you should declare updateAction() as `override`.

PaddleStroke:
Thanks for the review! I just fixed this. (And also sGroup was wrong)

PaddleStroke:
By the way for the 'checkable' property, perhaps we could change CommandGroup so that it can be declared in the Ctor?
```
CmdSketcherCompDimensionTools()
        : GroupCommand("Sketcher_CompDimensionTools")
    {
        sAppModule = "Sketcher";
        sGroup = "Sketcher";
        sMenuText = QT_TR_NOOP("Dimension");
        sToolTipText = QT_TR_NOOP("Dimension tools.");
        sWhatsThis = "Sketcher_CompDimensionTools";
        sStatusTip = sToolTipText;
        eType = ForEdit;
        bCheckable= true;
```
It would be more elegant than having to reimplement when not needed.

wwmayer:
I have added this PR to make the configuration more flexible. So, you can write `setCheckable(false)` inside the constructor.

PaddleStroke:
I have done so. CI is going to fail because you PR hasn't merged yet.

wwmayer:
Yes, I have merged it now. Can you please squash your 2 commits into a single one and force push it?

PaddleStroke:
@wwmayer squashed and force pushed

PaddleStroke:
And rebased

wwmayer:
Great. Once the CI builds are finished we can merge the PR.

PaddleStroke:
Force pushed again because there was a small problem in updateAction, I did not take the separator into account so the icons where shifted.

howie-j:
I suspect this PR has caused a regression where FreeCAD crashes if you right click with dimension active.

Steps to reproduce:
- new sketch
- press D or click dimension
- right click mouse

Can anyone confirm?

wwmayer:
> Can anyone confirm?

Yes, I can.

> I suspect this PR has caused a regression where FreeCAD crashes if you right click with dimension active.

But I don't see how this PR can cause the crash because it only re-organizes the tool bar commands but doesn't change anything on the dimension tools.

wwmayer:
The PR to blame must be this one: https://github.com/FreeCAD/FreeCAD/pull/10449

PaddleStroke:
I just rebuild master after the 3 PR merged and I can't replicate the crash on windows. @howie-j on what platform are you?

howie-j:
Yes sorry, git commit c72c9eaaa3b8ff91b9c872ac9887fdf1de322766 introduced the crash.

I'm on linux, 

```
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34116 (Git)
Build type: Release
Branch: master
Hash: c7116599cec06d658c92b02be823ef220ec060b5
Python 3.11.4, Qt 5.15.10, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
```

PaddleStroke:
Ah sorry I'm mistaken, if the sketch is empty I also get the illegal storage access.

PaddleStroke:
Ok I found the reason. Opening PR asap.

wwmayer:
Yes, I can confirm that c72c9eaaa3b8ff91b causes the crash.

PR #10464, opened 2023-08-31
Bug fix: Allow Draft WB to import in FreeCAD headless docker container

amrit3701:
# Description

Not able to import  `Draft` workbench in FreeCAD docker image compiled without GUI element.

Here is the dockerfile: [https://github.com/amrit3701/docker-freecad-cli/blob/master/0.21/amd64/Dockerfile](https://github.com/amrit3701/docker-freecad-cli/blob/master/0.21/amd64/Dockerfile)

Image already published: [https://hub.docker.com/r/amrit3701/freecad-cli/tags](https://hub.docker.com/r/amrit3701/freecad-cli/tags)

### DIFF
--- a/src/Mod/Draft/draftmake/make_hatch.py
+++ b/src/Mod/Draft/draftmake/make_hatch.py
@@ -24,7 +24,8 @@
 
 import FreeCAD
 from draftobjects.hatch import Hatch
-from draftviewproviders.view_hatch import ViewProviderDraftHatch
+if FreeCAD.GuiUp:
+    from draftviewproviders.view_hatch import ViewProviderDraftHatch
 
 def make_hatch(baseobject, filename, pattern, scale, rotation):
 


src/Mod/Draft/draftviewproviders/view_hatch.py
```diff
@@ -26,7 +26,9 @@
 import PySide.QtCore as QtCore
 import PySide.QtGui as QtGui
 
-import FreeCADGui as Gui
+import FreeCAD as App
+if App.GuiUp:
+    import FreeCADGui as Gui
```
yorikvanhavre, line 31: view providers are normally never created when FreeCAD runs without the GUI. Are you sure this file is used? I think if yes, it might be good to find where it is called and fix that instead...

amrit3701: Hi @yorikvanhavre you are right, view provider was initiated in `src/Mod/Draft/draftmake/make_hatch.py`. Here is the new commit 9f8f7c5ab3118264d418f1f94f06bd8ec61f1a63

yorikvanhavre:
Perfect, thanks for the fix!

PR #10465, opened 2023-08-31
Assembly: Add Assembly to precommit

chennes:
After this merge, anyone working on Assembly should make sure they have installed pre-commit into their local clone of the repo.  https://freecad.github.io/DevelopersHandbook/gettingstarted/ (see item 3 in "Setting up for development")

### DIFF
--- a/src/Mod/Assembly/InitGui.py
+++ b/src/Mod/Assembly/InitGui.py
@@ -51,8 +51,7 @@ class AssemblyWorkbench(Workbench):
     def __init__(self):
         print("Loading Assembly workbench...")
         self.__class__.Icon = (
-            FreeCAD.getResourceDir()
-            + "Mod/Assembly/Resources/icons/AssemblyWorkbench.svg"
+            FreeCAD.getResourceDir() + "Mod/Assembly/Resources/icons/AssemblyWorkbench.svg"
         )
         self.__class__.MenuText = "Assembly"
         self.__class__.ToolTip = "Assembly workbench"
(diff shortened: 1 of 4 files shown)


src/Mod/Assembly/InitGui.py
```diff
@@ -51,8 +51,7 @@ class AssemblyWorkbench(Workbench):
     def __init__(self):
         print("Loading Assembly workbench...")
         self.__class__.Icon = (
-            FreeCAD.getResourceDir()
-            + "Mod/Assembly/Resources/icons/AssemblyWorkbench.svg"
+            FreeCAD.getResourceDir() + "Mod/Assembly/Resources/icons/AssemblyWorkbench.svg"
```
PaddleStroke, line 54: Nice the 100 characters limit seems more forgiving than the standard black that Brad applied.

PaddleStroke (commented):
Ok, to further work on assembly we must opt-in this pre-commit right ? Though I see that sketcher is in the list and I have installed nothing regarding this.

src/Mod/Assembly/InitGui.py
chennes: We don't have a system in place to force the pre-commit hook (eventually we'll add a CI step that requires the formatting be correct, but that's not implemented right now). So actually installing pre-commit in the repo is a step that every developer must take.

PaddleStroke: By the way regarding this pre-commit hook: It makes PR very hard to review as every file you touch get completely re-writen. I'm thinking about FlachyJoe's PR that were +-1500 lines when the real code was less than 100 lines.
So perhaps all files in the sources should be passed through a formating step now. Though that would make horrible conflicts to all the opened PR. Though the pre-commit hook only dilute the problem of making conflicts to other open PR.

PR #10368, opened 2023-08-26
[Material] Add Material WB

davesrocketshop:
**NOTE**: This PR introduces a new dependency on _yaml-cpp_ (https://github.com/jbeder/yaml-cpp). This is available through
 apt on Debian based distributions, but the LibPack image will need to be updated for Windows builds

First PR for Issue #10174 - the Materials system update.

- The editor is minimally functional. You can save changes but may need to restart for them to take effect
- The API supports new and old methods of access. See the unit tests in src/Mod/Material/TestMaterialApp.py for examples
- Uses an updated .FCMat format. The old editor won't read the new files, but the new editor will still read the old files.
- Defaults to the old editor. This can be changed in the Material Preferences pane
- System materials are in the new format. You can still use the old format files by copying them to a separate folder and adding that as a custom folder in the Material Preferences pan
- Moves the preferences from FEM to a separate Materials section. This was submitted as a PR prior to 0.21 but that PR has been closed to avoid conflicts

PR #10474, opened 2023-09-01
Update CI to use LibPack 0.21 V2.10

chennes:
This libpack adds yaml-cpp and fmt.


adrianinsaval:
what do we use yaml-cpp for?

chennes:
@adrianinsaval It's going to be in the new Material WB. See PR https://github.com/FreeCAD/FreeCAD/pull/10368

PR #10475, opened 2023-09-01
CI: Fix syntax in labeler

chennes:
Syntax error in GitHub YAML file.


luzpaz:
Sorry about that. Thanks for the fix

PR #10476, opened 2023-09-01
Arch: Fixed IFC export of compsolids

yorikvanhavre:
Simple fix to workaround face index errors.
Could be backported. I'll make a separate PR

### DIFF
--- a/src/Mod/Arch/exportIFC.py
+++ b/src/Mod/Arch/exportIFC.py
@@ -2344,7 +2344,10 @@ def getRepresentation(
                 i = 0
                 rgbt = []
                 for sol in obj.Shape.Solids:
-                    rgbt.append(diffusecolor[i])
+                    if i < len(diffusecolor):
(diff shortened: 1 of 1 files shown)


Roy-043:
This is a strange issue. I can imagine a Shape with a number of solids and some loose faces.

This code will return `True` for such a shape:
`len(obj.Shape.Faces)) > sum([len(sol.Faces) for sol in obj.Shape.Solids])`

But this PR seems to be dealing with the opposite:
`len(obj.Shape.Faces)) < sum([len(sol.Faces) for sol in obj.Shape.Solids])`

Do you have an example of such a Shape?

yorikvanhavre:
The problem is that compsolids have a number of solids, and they share some faces. So if you take the sum of faces of each solids, you'll end up with a bigger number of faces than the total.

This solution is not ideal because it arbitrarily attributes a color, but I saw no better solution for the time being...

PR #10477, opened 2023-09-01
Gui: make GroupCommand more flexible

wwmayer:
Currently the GroupCommand by default sets the action group as non-exclusive, checkable and having a drop-down menu which isn't always the desired behaviour in sub-classes. Thus, some new methods are added to let a sub-class in its constructor decide how it should behave.


PaddleStroke:
Looks good.

PR #10487, opened 2023-09-01
Fix slight grammar error in Windows installer (English)

evan-chaney:
If you have FreeCad running and start the installer it will produce a message that says "You must close FreeCAD at first!" This is not proper grammar. This PR removes "at" to make the message correct.


chennes:
Thanks for the PR! I'm not going to roll a new installer for 0.21.1, but I'll use this when I generate the installer for 0.21.2.

PR #10489, opened 2023-09-01
Sketcher_Dimension Replace SubNames by std::move(SubNames)

PaddleStroke:
Small fix, @wwmayer

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -1149,7 +1149,7 @@ public:
 class DrawSketchHandlerDimension : public DrawSketchHandler
 {
 public:
-    DrawSketchHandlerDimension(std::vector<std::string>& SubNames)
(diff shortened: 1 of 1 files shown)


src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -1157,7 +1157,7 @@ class DrawSketchHandlerDimension : public DrawSketchHandler
         , selLine({})
         , selCircleArc({})
         , selEllipseAndCo({})
-        , initialSelection(SubNames)
+        , initialSelection(std::move(SubNames))
```
wwmayer, line 1160: This change without the mandatory change of the argument list of the constructor is dangerous as it destroys the passed string of the calling instance.
The change is: `DrawSketchHandlerDimension(std::vector<std::string> SubNames)`, i.e. the ampersand must be removed.

PaddleStroke: Ah sorry about that. I just forced push

wwmayer:
Yes, it's correct now.

Issue #10462, opened 2023-08-31
`mesh.unite()` causes segementaion fault


mbenkmann:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
OS: Ubuntu 20.04.6 LTS
Word size of FreeCAD: 64-bit
Version: 0.21.0.33675 (Git) AppImage
Build type: Release
Branch: (HEAD detached at 0.21.0)
Hash: 41b058e2087de60dc8fef3d3e68c7d0129e13abf
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: English/United States (en_US) [ OS: English/United States (en_US) ]
Installed mods: 
  * Help 1.0.3
  * CubeMenu
```

### Subproject(s) affected?

None

### Problem description

PRECONDITION: Create a New document and add a Box.

**Works**: The following code works when pasted directly into the Python console. 

**Broken**: If saved in a macro `test.FCMacro` and executed via Macro/Macros.../Execute causes a SEGMENTATION FAULT.

The code worked fine in FreeCAD 0.20

```
import FreeCAD, Mesh, MeshPart

shape = FreeCAD.ActiveDocument.getObject("Box").Shape
mesh = Mesh.Mesh()
mesh2 = MeshPart.meshFromShape(Shape=shape, LinearDeflection=.1, AngularDeflection=3, Relative=False)
mesh = mesh.unite(mesh2)
```

This is the backtrace printed by FreeCAD:

```
Program received signal SIGSEGV, Segmentation fault.
#0  /lib/x86_64-linux-gnu/libc.so.6(+0x43090) [0x7ff482919090]
#1  0x7ff482c15c5b in std::ostream::sentry::sentry(std::ostream&) from /tmp/.mount_freecaMyqEOK/usr/bin/../lib/libstdc++.so.6+0x13
#2  0x7ff482c16d62 in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long) from /tmp/.mount_freecaMyqEOK/usr/bin/../lib/libstdc++.so.6+0x29
#3  0x7ff483bc445b in Base::Builder3D::Builder3D() from /tmp/.mount_freecaMyqEOK/usr/bin/../lib/libFreeCADBase.so+0x4b
[...]

PR #10499, opened 2023-09-02
Base: fix segfault in InventorBuilder constructor

bdieterm:
As described in issue #10462, the `mesh.unite(...)` command causes a crash with the error "segmentation fault".
I could also reproduce the crash by executing the single line `import Mesh; mesh = Mesh.Mesh(); mesh.unite(mesh)` in the FreeCAD GUI Python console. The problem is not only in v0.21.0 but also in master.

As far as I have found out, the problem boils down to this:
In the constructor of a derived class the initializer of the base class is called and given a reference to the derived class's member variable. Then, in the constructor of the base class, the referenced variable is accessed, but that variable was not yet constructed because it will be constructed in the derived class AFTER the base class was fully constructed.

The following code is a stand-alone example to demonstrate the problem
```
#include <iostream>
#include <sstream>

class InventorBuilder
{
public:
    explicit InventorBuilder(std::ostream& str);

private:
    std::ostream& result2;
};

InventorBuilder::InventorBuilder(std::ostream& output)
  : result2(output)
{
    result2 << "#Inventor V2.1 ascii \n\n";
}



class Builder3D : public InventorBuilder
{
public:
    Builder3D();

private:
    std::stringstream result;
};

Builder3D::Builder3D()
  : InventorBuilder(result)
{
}



int main()
{
    std::cout << "creating an object of only the base class" << std::endl;
    std::stringstream result;
    InventorBuilder aa(result);
    std::cout << "... worked normally!" << std::endl;

    std::cout << "creating an object of the derived class" << std::endl;
    Builder3D bb;
    std::cout << "... does not work -- segmentation fault" << std::endl;
    return 0;
}
```
See [here on stackoverflow](https://stackoverflow.com/questions/71679698/initialization-of-a-base-class-reference-from-a-derived-class-member) for further explanations.

To solve the problem, I changed the base class to not access the referenced variable in the constructor. Instead, the initialization will be done at the beginning of other member functions that access that referenced variable.

I will also create a backport for 0.21 after this PR is merged.

### DIFF
--- a/src/Base/Builder3D.cpp
+++ b/src/Base/Builder3D.cpp
@@ -915,7 +915,13 @@ void TransformItem::write(InventorOutput& out) const
 InventorBuilder::InventorBuilder(std::ostream& output)
   : result(output)
 {
-    result << "#Inventor V2.1 ascii \n\n";
+    /* We must not access variable "output" here in the constructor!
+     * When this constructor is called from the constructor of the
+     * derived class Builder3D::Builder3D() then output is a reference
+     * to Builder3D's member variable which is not yet valid because
+     * this base class is constructed before the derived class.
+    */
+    firstLineWritten = false;
 }
 
 InventorBuilder:: ~InventorBuilder() = default;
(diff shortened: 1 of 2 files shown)


wwmayer:
Thanks for the explanation and working on a solution. However, it's basically a problem of the class Builder3D and not its base class InventorBuilder. That's why I prefer a solution that fixes Builder3D.

So, instead of letting Builder3D derive from InventorBuilder the latter can be a class member of the former. Then Builder3D must implement the three methods addNode, beginSeparator and endSeparator.

wwmayer:
See https://github.com/FreeCAD/FreeCAD/pull/10508

PR #10505, opened 2023-09-02
Draft: Fix angle range issue of make_circle

### LINKED ISSUE #10504: [Problem] Draft: make_circle can only handle angles in -360 to 360 degree range

### Problem description

If make_circle receives angles that are outside the -360 to 360 degree range, -360 resp. 360 degrees are used instead.

### Full version info

```
OS: Windows 8 build 9600
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.33981 (Git)
Build type: Release
Branch: master
Hash: 5a153e50ffb6046b6cc4dba498344032f6edb952
Installed mods:
```

### Subproject(s) affected?

Draft

### Anything else?

_No response_

### Code of Conduct

Roy-043:
Fixes #10504.

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_arcs.py
+++ b/src/Mod/Draft/draftguitools/gui_arcs.py
@@ -342,8 +342,8 @@ class Arc(gui_base_original.Creator):
             end = math.degrees(self.firstangle + self.angle)
             if end < sta:
                 sta, end = end, sta
-            sta = sta % 360
-            end = end % 360
+            sta %= 360
+            end %= 360
 
             try:
                 Gui.addModule("Draft")
--- a/src/Mod/Draft/draftguitools/gui_arcs.py
+++ b/src/Mod/Draft/draftguitools/gui_arcs.py
@@ -342,7 +342,7 @@ class Arc(gui_base_original.Creator):
             end = math.degrees(self.firstangle + self.angle)
             if end < sta:
                 sta, end = end, sta
-            sta = sat % 360
+            sta = sta % 360
             end = end % 360
 
             try:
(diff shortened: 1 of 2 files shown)


Roy-043:
Trying to follow official protocols here... ;)

src/Mod/Draft/draftguitools/gui_arcs.py
```diff
@@ -342,13 +342,9 @@ def drawArc(self):
...
             if end < sta:
                 sta, end = end, sta
-            while True:
-                if sta > 360:
-                    sta = sta - 360
-                elif end > 360:
-                    end = end - 360
-                else:
-                    break
+            sta = sat % 360
```
bdieterm, line 345: there is a typo: `sat` should be `sta`
even better: `sta %= 360`

src/Mod/Draft/draftguitools/gui_arcs.py
```diff
@@ -342,13 +342,9 @@ def drawArc(self):
...
+            end = end % 360
```
bdieterm, line 346: `end %= 360` would be cleaner

bdieterm (commented):
General remark:
Actually, for objects the allowed range of angle properties is [-360 deg, 360 deg], see [here](https://github.com/FreeCAD/FreeCAD/blob/89579cff6e8834e296d5efe6292f019fadedce53/src/App/PropertyUnits.cpp#L41). It could be a bit confusing for the user that for example `obj = Draft.makeCircle(9, face=False, startangle=-50, endangle=50)` will create an object with only positive angles in the `FirstAngle` and `LastAngle` property although -50 deg is still in the allowed angle range. A short note in the docstring of `make_circle` would be nice.
It would also be possible to still allow the [-360 deg, 360 deg] range but to do the modulo when exceeding it, depending on the side: -370 --> -10 and +370 --> +10

Roy-043:
Thanks for your review. I acually missed it when I fixed the typo. All of your remarks make sense. I'll update the PR.

yorikvanhavre:
LGTM!

Issue #6117, opened 2022-02-07
FreeCAD Path needs a new Postprocessor architecture.


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=4468

* **Reporter:** sliptonic
* **Date submitted:** 10/18/2020
* **FreeCAD version:** 
* **Category:** Feature
* **Status:** new
* **Tags:** 

# Original report text

The current post-processor architecture requires individual .py files for each post.  This is flexible in allowing users to copy an existing post and modify it but results in lots of duplicated code and redundant work.
Changes and improvements made in one post don't become part of the other posts so bugs and incompatibilities persist.

# Additional information

A better architecture would allow code reuse while preserving the users ability to customize the post to their needs.

# Other bug information

* **Priority:** normal
* **Severity:** feature
* **Category:** Feature
* **Updated:** 2/6/2021

LarryWoestman:
I am currently working on refactoring the postprocessor code and the postprocessors.

I think it makes sense to assign this issue to me.

luzpaz:
@sliptonic :point_up:

luzpaz:
@sliptonic JFYI, this ticket is still open yet is labeled 'Done' in your CAM stuff list.

sliptonic:
Yeah, I left it open because the refactoring only touched the first part of post-processing.  The next step will be to start refactoring the individual posts to re-use code and to be easier to customize.

LarryWoestman:
Progress is being made on refactoring the individual posts, with more work to come.

luzpaz:
@LarryWoestman what's your verdict? can we close this now ?

LarryWoestman:
Not yet.  I would consider the process over half done, but still more work to go.  The main tasks left relate to adding back in enough customizability to make up for having lost the ability to change the shared code.  There are also over 10 more postprocessors that I would like to port to the "refactored" code base.  I suspect that I will learn about some additional customizability requirements from each one :-)

PR #10509, opened 2023-09-02
Path:  Major refactoring of the export_common function.

LarryWoestman:
This is part of the work for #6117.

This is one in a series of PR's where I am taking code from at least 4 postprocessor files partially written by different people, combining them into three files, then trying to make the code more consistent and simpler. The long term goal is to enable much simpler postprocessor files with much less copied code in each postprocessor file and much more "shared" code in the three Utils*.py files.

In this PR I refactored the "export_common" function to make it more maintainable.  This also required refactoring the "parse_a_path" and "parse_a_group" functions a bit to stay compatible.  I also removed "message" command handling since it didn't seem to be able to be used.

### DIFF
--- a/src/Mod/Path/Path/Post/UtilsArguments.py
+++ b/src/Mod/Path/Path/Post/UtilsArguments.py
@@ -504,10 +504,6 @@ def init_shared_values(values):
     #
     values["RAPID_MOVES"] = ["G0", "G00"]
     #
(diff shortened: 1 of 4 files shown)


luzpaz:
@LarryWoestman does #6117 depend on this ?

LarryWoestman:
Yes, it does.

PR #10515, opened 2023-09-03
Sketcher: fix several inconsistencies in Sketcher_CompDimensionTools

wwmayer:
* The menu text and tooltip of the Sketcher_ConstrainRadius command shouldn't be different if part of the group command or not. The explicit mentioning to change the weight of a B-Spline is pointless and confusing
* The order of Sketcher_ConstrainRadius and Sketcher_ConstrainDiameter has flippend when used in Sketcher_CompDimensionTools
* The command Sketcher_ConstrainRadiam is completely missing
* The command Sketcher_ConstrainSnellsLaw was wrongly re-added
* CmdSketcherCompConstrainRadDia uses a wrong context

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -1057,10 +1057,10 @@ public:
         addCommand("Sketcher_ConstrainDistanceX");
         addCommand("Sketcher_ConstrainDistanceY");
         addCommand("Sketcher_ConstrainDistance");
-        addCommand("Sketcher_ConstrainDiameter");
         addCommand("Sketcher_ConstrainRadius");
+        addCommand("Sketcher_ConstrainDiameter");
+        addCommand("Sketcher_ConstrainRadiam");
         addCommand("Sketcher_ConstrainAngle");
-        addCommand("Sketcher_ConstrainSnellsLaw");
     }
 
     void updateAction(int mode) override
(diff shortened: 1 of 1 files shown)


FEA-eng:
@wwmayer @PaddleStroke Maybe one more change could be added to this - the Snell's law constraint was removed from the toolbar in 0.21 since it's so niche and now I see that it's back. I would suggest removing it again.

wwmayer:
> the Snell's law constraint was removed from the toolbar in 0.21 since it's so niche and now I see that it's back. I would suggest removing it again

I have realized this too but don't have a strong opinion but you are probably right. I will adjust the PR.

FEA-eng:
> I have realized this too but don't have a strong opinion but you are probably right. I will adjust the PR.

Thanks. Pretty much nobody uses this constraint, it only clutters the toolbar so it was decided that it can be removed from the toolbar (and left only in the menu) in the previous release cycle.

PaddleStroke:
Ah yes sorry I added the snells' law because I thought here it wasn't
cluttering so much. While initially I was for the removal of the main
toolbar.
Anyway I have no strong opinion about it, it can be removed.

PLChris:
I remember too a broad consensus, that Snell's law should occur in the menu only.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -8526,7 +8527,7 @@ void CmdSketcherCompConstrainRadDia::languageChange()
     QAction* arc3 = a[2];
     arc3->setText(QApplication::translate("CmdSketcherCompConstrainRadDia",
```
luzpaz, line 8528: is `CmdSketcherCompConstrainRadDia` a typoe as well ?

PaddleStroke:
Note : I had not put Sketcher_ConstrainRadiam in the command group on purpose. Because the goal of the command group is to have access to individual commands if the user need to apply a specific constraint.

And Sketcher_ConstrainRadiam is not a specific constraint. It's actually a 'combo-tool' like Sketcher_Dimension. So there is no usecase where user will choose radiam in the command group.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
wwmayer: No, it's an abbreviation for **Rad**ius and **Dia**meter

PR #10518, opened 2023-09-03
Gui: Add cancel button to transform task UI dialog

### LINKED ISSUE #7517: [Feature Request] Add cancel button in the Transform task UI

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=34&t=71598

### Subproject(s) affected?

_No response_

### Idea description

Sometime we just move a part to compare with another, but to go back to previous position we need to click OK and then undo the changes.
But if the part is filled with details then it takes time to undo the changes. So with a "Cancel" button this can be prevented.
Also if you move the part in 6 step than you have to press ctrl+z 6 times to go to previous position.

With the cancel button that will be reduced and the UI consistency will also be maintained. 

image

Requested UI 
image

### Anything else?

_No response_

### Code of Conduct

didendron:
Fixes #7517


chennes:
Per that issue, it sounds like @wwmayer basically knew exactly what was needed: Werner, can you take a look and make sure this covers it?

luzpaz:
After review is done if i may offer the suggestion:  squash && `git commit --amend -m "Gui: Add cancel button to transform task UI dialog" -m "Fixes #7517"`

wwmayer:
> Werner, can you take a look and make sure this covers it?

It all looks good and works as expected.

luzpaz:
Congrats on your 1st PR @didendron! Thanks wwmayer and chennes for your participation as well!

didendron:
@luzpaz , @wwmayer , @chennes thanks for checking on me.

wwmayer:
Something went wrong with this PR because it now includes 120 files. I will prepare a new PR with the relevant changes.

chennes:
We just need to drop the commit added by the pre-commit.ci hook -- it's no longer relevant. This was the first PR to get pushed after that hook was enabled, and it turned out there were still some outstanding issues that pre-commit was catching. Those are now resolved.

wwmayer:
See https://github.com/FreeCAD/FreeCAD/pull/10529

So, I will close this PR.

wwmayer:
> We just need to drop the commit added by the pre-commit.ci hook -- it's no longer relevant. This was the first PR to get pushed after that hook was enabled, and it turned out there were still some outstanding issues that pre-commit was catching. Those are now resolved.

Oh, does this mean that there is now a mechanism active that automatically adds new commits that has auto-corrected the code? Isn't this dangerous because even when running the git hooks locally the code can get broken  by the automated ordering of header files.

With our *PyImp.cpp files this automated reordering of header files is quite annoying because it puts the *Py.cpp before the Py.h files and then the compiler fails to find the class declarations. Those things must be fixed by hand by putting a `clang-format off` / `clang-format on` around these includes.

chennes:
The hook's default setting is to add a cleanup commit to the PR -- this could be disabled (I think... you have to have "owner" permissions here to make the change) and just deliver a CI failure, but I'm not sure that's better. We should probably instead polish up the .clang-format settings to ensure they actually do what we want them to (or to add the protection lines you suggest).

Regarding the TS files: the only ones I saw were the translated files themselves, and the changes were removing trailing whitespace from some translation strings, rather than the source. If it proves to be a problem we can disable checking *.ts files, though.

wwmayer:
OK

Issue #5547, opened 2022-02-07
Objects names in Label & Attributes Tree View should be translatable


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=494

* **Reporter:** normandc
* **Date submitted:** 11/27/2011
* **FreeCAD version:** 
* **Category:** Feature
* **Status:** assigned
* **Tags:** #lowhangingfruit, translation

# Original report text

Currently, all the GUI tools can be localized in other languages than English, either in menus or the toolbars. But the names of the objects created with these tools, in the Tree view under Labels & Attributes remain in English.

It would avoid confusion to users and enhance the user experience if those objects' names could be translated as well.

EDIT: So, to be clear, this ticket is about translating default names for items to the local language on <b>object creation</b>. ( ~6953 )
This ticket is <b>not</b> about being able to translate said items post object creation.

# Additional information

EDIT: Also, at this time, only Part Primitive labels are translatable. All other FreeCAD objects are missing: 
Arch, 
Draft, 
PartDesign, and so on. 

# Other bug information

* **Priority:** normal
* **Severity:** minor
* **Category:** Feature
* **Updated:** 2/6/2021

# Discussion from Mantis ticket

### Comment by Jriegel 2011-11-27 17:56
Its perfectly possible to set the Label property with a translated 
text in the command which create the object. 
But we have to take into account that then its hard to exchange 
document with other people... They will see names which means nothing to them...

I would not do it for the 0.12!

---

### Comment by normandc 2011-11-27 18:09
I hadn't thought of that.

I wonder if those object text strings could be translated on the fly like the rest of the GUI? That they would appear in the language set in FreeCAD?

If I recall, that's how Solid Edge works, but I haven't used a French version of the software for a couple of years. I believe that the features names would be translated in the software locale, but I can't be sure.

Still, as it is, the users create features of which names mean nothing to them if they don't have a working knowledge of English.

---

### Comment by ian.rees 2016-04-06 07:01
Just trying to think through how this might work:

1) French user creates Cylindre and Cylindre001 in a new file, saves it and sends to English user
2) English user opens file and sees Cylinder and Cylinder001
3) English user fuses Cylinder and Cylinder001 together, to create Fusion.  Re-saves file and sends back to French user.

What does the French user see when they open the document and expand the object in it?

If the French user opens the file from the English user and saves it again without explicitly modifying anything, do we want the file to change?

[...]

PR #10519, opened 2023-09-03
FEM: removing "constraint" from the names

kaktusus:
Removing "constraint" from the names of most FEM workbench features.
https://github.com/FreeCAD/FreeCAD/discussions/10217

### DIFF
--- a/src/Mod/Fem/Gui/Command.cpp
+++ b/src/Mod/Fem/Gui/Command.cpp
@@ -177,8 +177,8 @@ CmdFemConstraintBearing::CmdFemConstraintBearing()
 {
     sAppModule      = "Fem";
     sGroup          = QT_TR_NOOP("Fem");
-    sMenuText       = QT_TR_NOOP("Constraint bearing");
-    sToolTipText    = QT_TR_NOOP("Creates a FEM constraint for a bearing");
+    sMenuText       = QT_TR_NOOP("Bearing constraint");
+    sToolTipText    = QT_TR_NOOP("Creates a bearing constraint");
     sWhatsThis      = "FEM_ConstraintBearing";
     sStatusTip      = sToolTipText;
     sPixmap         = "FEM_ConstraintBearing";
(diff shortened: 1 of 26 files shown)


chennes:
@FEA-eng I believe you are the local expert in this area -- would you mind reviewing?

FEA-eng:
> @FEA-eng I believe you are the local expert in this area -- would you mind reviewing?

I checked this and discussed with @kaktusus. Everything is fine, this PR can be merged.

chennes:
FYI @kaktusus, I fixed up your branch a bit: I used an interactive rebase (`git rebase -i`) to "drop" all but the single commit you actually wanted to keep, and then I rebased onto the latest HEAD of the master branch. That's what I've force-pushed here.

maxwxyz:
@kaktusus @FEA-eng Nice to see this change, unfortunately when creating the "constraints", the tree items still keep the name with "constraint":
grafik

luzpaz:
@maxwxyz can you please add your full about info so we know how you built it ?

kaktusus:
This problem coincides with an open ticket ...
was recently raised I think by you luzpaz _(or by chennes)_
can't find it now ...
will try to look for it again later :wink:

kaktusus:
it took me a while but I found it: https://github.com/FreeCAD/FreeCAD/issues/5547
so that we can link this post to that thread.

maxwxyz:
Not sure if #5547 is about the same. I thought in this PR here, the idea was to remove the prefix "constraint" in FEM. When adding the "constraints" the tree elements still have the name "constraint" in them, shouldn't they be removed in the tree items as well, to match the button names?

kaktusus:
> Not sure if #5547 is about the same. I thought in this PR here, the idea was to remove the prefix "constraint" in FEM. When adding the "constraints" the tree elements still have the name "constraint" in them, shouldn't they be removed in the tree items as well, to match the button names?

These are related topics for me I noticed another problem.

On the other hand, you are definitely right about the mainstream. :wink:

Issue #10451, opened 2023-08-30
[Feature] Custom CSS option for StartPage


qewer33:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
OS: Fedora Linux 38 (KDE Plasma) (KDE/plasmax11)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34071 (Git)
Build type: Release
Branch: startpage-housekeeping
Hash: 932fd4638ba80671115fb2aa871ab3a273b6f673
Python 3.11.4, Qt 5.15.10, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * Glass
  * lattice2 1.0.0
  * kicadStepUpMod 10.19.4
  * Manipulator 1.5.0
  * Assembly4.backup1691951820.225657 0.50.2
  * Assembly4
  * Behave-Dark-Colors 0.1.1
  * Curves 0.6.13
  * DynamicData 2.46.0
  * fasteners 0.4.59
  * Freecad-Built-in-themes-beta 1.2.2
  * freecad.gears.backup1692223246.5590403 1.0.0
  * freecad.gears
  * ExtremeProLight 1.6.0
  * A2plus 0.4.60n
  * ExplodedAssembly
  * FreeCAD_DarkPreferencePAck 0.0.1
```

### Subproject(s) affected?

Other (specify in description)

### Problem description

There are many power users who want to customize the StartPage for their own use cases. Preference packs might also want to customize the StartPage. Currently pref packs can do this with a stylesheet if "Use FreeCAD stylesheet" option under Start WB preferences is checked. There's also a "Custom template" option for power users. 

But I feel like adding a "custom CSS" option to Start preferences would make this significantly easier for power users and preference packs. It'll ideally be a text area where you can paste your custom CSS in and the StartPage will apply it on top of it's default CSS. And honestly if this is implemented, the custom template option can be removed. I have never heard of anyone in the community using it (please mention if you have) and never seen any alternative templates online.

I can open a PR for this myself, this is just a discussion issue in-case someone opposes it or suggests something more/else.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

yorikvanhavre:
[...]

PR #10520, opened 2023-09-03
StartPage: Add custom CSS option, re-organize preferences and further fixes

qewer33:
The changes made by this PR:
- Re-organize Start WB preferences and add "Custom CSS" option (closes https://github.com/FreeCAD/FreeCAD/issues/10451) + rename preference page widget names to make code more readable

Screenshot_20230903_192913
Screenshot_20230903_192944

- Turn the StartPage "Import" quickstart button into "Open" (uses Std_Open now instead of Std_New & Std_Import)
- Removes the "Hide scrollbars" preference, seems like this was originally added because the scrollbars looked ugly. They're styled according to the theme now and they're thinner so there shouldn't be any reason to keep this
- Minor CSS and HTML fixes


qewer33:
Not sure why the merge conflicts happened but they should be fixed now, hopefully CI builds fine

qewer33:
Ok I have no idea why the tests are failing. Ran locally on my machine and I get this output:
```
======================================================================
ERROR: test_html_validates (TestStart.TestStartPage.TestStartPage.test_html_validates)
----------------------------------------------------------------------
  File "/usr/lib64/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/usr/lib64/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
  File "/usr/lib64/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/home/qewer/Desktop/Git/freecad-build/Mod/Start/TestStart/TestStartPage.py", line 144, in test_html_validates
    StartPage.exportTestFile()
  File "/home/qewer/Desktop/Git/freecad-build/Mod/Start/StartPage/StartPage.py", line 665, in exportTestFile
    f.write(handle())
UnicodeEncodeError: 'ascii' codec can't encode character '\xfc' in position 22662: ordinal not in range(128)

```
"\xfc" seems to be the letter "ü" which does exist on my keyboard but I couldn't find any instance of it on the StartPage files. I'll investigate further.

qewer33:
Alright neverminds. This should fix it.

Syres916:
Don't ignore that Python 3.11 UnicodeDecodeError, they appear to be popping up in all sorts of places, I believe part of the test is specifically around the author field must be able to save non-ascii characters.

chennes:
Start is now in our pre-commit config -- please install pre-commit to fix up the formatting here. See Item 3 in "Setting Up for Development" here: https://freecad.github.io/DevelopersHandbook/gettingstarted/

yorikvanhavre:
This looks very good and apparently all issues are solved, so I think we can merge!

Syres916:
@yorikvanhavre please see https://forum.freecad.org/viewtopic.php?p=704090#p704090

PR #10530, opened 2023-09-04
Git: Add recent formatting commits to ignore-revs

chennes:
Adds recent clang-format and pre-commit formatting change commits to the .git-blame-ignore-revs file -- generated with:
```
git log --format="%H # %s" --grep="format" --since="2023-08-1" >> .git-blame-ignore-revs
git log --format="%H # %s" --grep="pre-commit" --since="2023-08-1" >> .git-blame-ignore-revs
```
(and then hand-edited to remove a few that shouldn't be ignored).

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -73,3 +73,28 @@ ec899d75c6abf45d7ee4dd91ea0f8141a9818446
 233a337ca5d693e4773fecf1de14325752afd658
 9271b6d433578f65df9b469131ece4e4bdbb19aa
 2637527e3a4e132506e8c074b752721d66ae146a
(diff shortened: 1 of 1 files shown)


luzpaz:
@chennes an aside, maybe it's worth our time and part of the workflow  to add the commit name as a comment  ?

chennes:
I'm not sure if GitHub supports that syntax: according to the git docs it should work, but I am testing it here.

chennes:
This format appears to work just fine here at GitHub, despite the scary red formatting of the .git-blame-ignore-revs file itself. I have no idea why it's showing like that, but apparently it is fair game for us to include comment lines (indicated with a #).

Issue #10524, opened 2023-09-03
[Problem] Sketcher code needs refactoring


FlachyJoe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

The sketcher workbench code needs a cleanup

- Gui/CommandConstraints.cpp is too big (10 000+ lines)
- quite same code is copied at three places for each constraints cases (in Gui/CommandConstraints.cpp , Gui/EditModeConstraintCoinManager.cpp and Gui/ViewProviderSketch.cpp)

### Full version info

```shell
[code]
OS: Debian GNU/Linux trixie/sid (XFCE/lightdm-xsession)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.33981 (Git)
Build type: Release
Branch: master
Hash: 5a153e50ffb6046b6cc4dba498344032f6edb952
Python 3.11.5, Qt 5.15.10, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: French/France (fr_FR)
[/code]
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
I was thinking a bit about the code structure of the commands actually.
I remember initially being very confused about where were commands, what were dsh and so on. Perhaps it would be best to have one file for each command.

Recently Abdullah split CommandCreateGeos because it was so huge. But he left CommandCreateGeo and just moved the DSH to individual files.
Perhaps it would be best to have individual files for commands : 
- CmdSketcherGeometryLine : include the command + the DSH
- CmdSketcherGeometryCircle
- CmdSketcherConstraintDistance
- CmdSketcherConstraintCoincident
- ...

But this also raise the question of uniformity of code between workbenches. Currently every workbench architecture is different, which makes it hard to discover. Maybe we should have structure rules that all workbenches have to follow. I think for contributors it would make life much easier. Which would result in more contributions and less confusion.

Rule example : 
- Commands should be in individual files with naming following the naming convention 'CmdWorkbenchName'. All commands files should be in the folder Gui/Commands. 
- The command files should contain all the code relative to this command (example : Taskbox, DSH...)

FlachyJoe:
> I would also add that it's not an issue, more like a discussion.

IMHO it's an issue for new contributors to understand the code and so it limits the possible improvement at user side.

> this also raise the question of uniformity of code between workbenches.

[...]

PR #10531, opened 2023-09-04
[Sketcher] Angle constraint refactor

PaddleStroke:
First step to solve https://github.com/FreeCAD/FreeCAD/issues/10524

This PR reduce code duplicity in the command angle and in sketcher_dimension.

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -186,6 +186,132 @@ bool isGeoConcentricCompatible(const Part::Geometry* geo)
     return (isEllipse(*geo) || isArcOfEllipse(*geo) || isCircle(*geo) || isArcOfCircle(*geo));
 }
 
+/// Makes an angle constraint between 2 lines
+void SketcherGui::makeAngleBetweenTwoLines(Sketcher::SketchObject* Obj,
+    Gui::Command* cmd,
+    int geoId1,
+    int geoId2)
+{
+    Sketcher::PointPos posId1 = Sketcher::PointPos::none;
+    Sketcher::PointPos posId2 = Sketcher::PointPos::none;
+    double actAngle;
+
+    if (!calculateAngle(Obj, geoId1, geoId2, posId1, posId2, actAngle)) {
+        return;
+    }
+
+    if (actAngle == 0.0) {
+        Gui::TranslatedUserWarning(
+            Obj,
+            QObject::tr("Parallel lines"),
+            QObject::tr("An angle constraint cannot be set for two parallel lines."));
+
+        return;
+    }
+
+    Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Add angle constraint"));
+    Gui::cmdAppObjectArgs(Obj,
+        "addConstraint(Sketcher.Constraint('Angle',%d,%d,%d,%d,%f))",
+        geoId1,
+        static_cast<int>(posId1),
+        geoId2,
+        static_cast<int>(posId2),
+        actAngle);
+
+    if (areBothPointsOrSegmentsFixed(Obj, geoId1, geoId2)
+        || constraintCreationMode == Reference) {
+        // it is a constraint on a external line, make it non-driving
+        const std::vector<Sketcher::Constraint*>& ConStr = Obj->Constraints.getValues();
+
+        Gui::cmdAppObjectArgs(Obj, "setDriving(%d,%s)", ConStr.size() - 1, "False");
+        finishDatumConstraint(cmd, Obj, false);
+    }
+    else {
+        finishDatumConstraint(cmd, Obj, true);
+    }
+}
+
+
+
+bool SketcherGui::calculateAngle(Sketcher::SketchObject* Obj, int& GeoId1, int& GeoId2, Sketcher::PointPos& PosId1, Sketcher::PointPos& PosId2, double& ActAngle)
+{
+    const Part::Geometry* geom1 = Obj->getGeometry(GeoId1);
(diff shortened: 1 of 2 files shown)


src/Mod/Sketcher/App/SketchObject.cpp
```diff
@@ -8500,6 +8500,83 @@ double SketchObject::calculateAngleViaPoint(int GeoId1, int GeoId2, double px, d
         throw Base::ValueError("Null geometry in calculateAngleViaPoint");
 }
 
+bool SketchObject::calculateAngle(int& GeoId1, int& GeoId2, Sketcher::PointPos& PosId1, Sketcher::PointPos& PosId2, double& ActAngle)
+{
+    const Part::Geometry* geom1 = getGeometry(GeoId1);
+    const Part::Geometry* geom2 = getGeometry(GeoId2);
+
+    if (geom1->getTypeId() == Part::GeomLineSegment::getClassTypeId() &&
+        geom2->getTypeId() == Part::GeomLineSegment::getClassTypeId()) {
```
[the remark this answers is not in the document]
PaddleStroke: You're right!

src/Mod/Sketcher/App/SketchObject.cpp
```diff
@@ -8500,6 +8500,84 @@ double SketchObject::calculateAngleViaPoint(int GeoId1, int GeoId2, double px, d
...
```
benj5378, line 8509: You changed the behaviour when trying to make the guard clause. You didn't invert the condition

PaddleStroke: Ah this is unfortunate. Fixed

howie-j:
Perhaps include arc angle also? https://github.com/FreeCAD/FreeCAD/issues/10598

PaddleStroke:
@howie-j I'll make a separate PR as it's not really related to this PR.

PaddleStroke:
@abdullahtahiriyo interested by this little refactor perhaps?
CI complains because of precommit.

PaddleStroke:
Build fails but I don't think it's due to this PR?

PaddleStroke:
@wwmayer care to take a quick look here ? It's just a little refactor because some code was copy-pasted in 3 places in CommandConstraint.

src/Mod/Sketcher/App/SketchObject.h
```diff
@@ -472,6 +472,12 @@ class SketcherExport SketchObject: public Part::Part2DObject
     static void appendMalformedConstraintsMsg(const std::vector<int>& malformed, std::string& msg);
 
     double calculateAngleViaPoint(int geoId1, int geoId2, double px, double py);
+    // Find the angle between two lines. Return false if geoIds are not lines.
+    bool calculateAngle(int& geoId1,
```
wwmayer, line 476: the method should be declared as `const` if possible.

wwmayer:
Please also consider the reported lint warnings (clang-diagnostic error, modernize C++, ...).

PaddleStroke:
I applied const and added #include <Gui/Command.h> that was giving a clang error.
Thanks

PaddleStroke:
Got it. Please let me know if the forward declaration is correctly done.

wwmayer:
Yes, it's fine.

abdullahtahiriyo:
@PaddleStroke 

Why does the member function added to SketchObject need to be in SketchObject? In other words, why it is not a free function in CommandConstraints.h instead?

I am tracking the dependencies visually, so I may be wrong, but it seems to me that that member function does not access anything private or protected of SketchObject. It is not used by any other part of the App namespace either, only Gui.  

SketchObject is already some kind of God object. To avoid it becoming unmaintainable, we should strive to only add there what is necessary. 

When functionality needed in App is to be added and there is no need for a member function, but a function using the interface is enough, one could consider either a free function in App namespace, or a new class (see SketchAnalysis as an example).

If for any reason, there is no dependency in SketchObject and there is no need for it in App namespace and CommandConstraints.h is not a good place (for example, because you will use it elsewhere in Gui), which I fail to see at this moment, then perhaps Sketcher/Gui/Utils.h could be an alternative place.

Let me know if in doubt.

PaddleStroke:
There is no necessity. I did that because I was following `calculateAngleViaPoint` pattern. I thought making code sharable was the correct way of doing things. Especially when what this function is doing (calculating angle between two lines) could potentially be used in other places of sketcher.

abdullahtahiriyo:
calculateAngleViaPoint could be refactored out of SketchObject. Probably the reason it has not been done yet is that it is used by the tangency locking function in SketchObject. It may be possible to refactor that one out too (but within App), as it is needed for the constraint addition mechanism.

It is definitely good to reuse code. As a free function, it is equally reusable (you just need to add the header where you want to reuse it). 

As for scope, it is not extremely strict. There is some leeway. I would go for the broadest where it makes sense now and avoid broader where YAGNI. For example, here it must be within the Sketcher module, because it makes use of Sketcher module classes (meaning not global App/Gui). Nowadays, I think it is only used in SketcherGui. Eventually it might perhaps be possible to use it at Sketcher namespace level. However, looking at what Sketcher namespace does now, it looks to me like a YAGNI scenario. For this reason, I would tend to maintain it within SketcherGui. If this is just used as a refactor of constraints, I think it makes sense to put it in CommandConstraints.

I am interested in exposing the rationale so that we understand why we do things one way and not the other.

PaddleStroke:
It's clear. So I will move it in CommandConstraint.

abdullahtahiriyo:
Let me know when you are happy with the code.

PaddleStroke:
I moved the function. Please confirm if this is ok.

PR #10535, opened 2023-09-04
[Start] Fix Python 3.11 Unicode Errors

Syres916:
On Python 3.11 (tested using Debian 12) UnicodeDecodeErrors are raised when running All Tests, this PR addresses those errors.

Example Error :
```
  File "/home/john/freecad-daily-build/Mod/Start/StartPage/StartPage.py", line 422, in handle
    ALTCSS = f.read()
             ^^^^^^^^
  File "/usr/lib/python3.11/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 421: ordinal not in range(128)
```

### DIFF
--- a/src/App/PropertyPythonObject.cpp
+++ b/src/App/PropertyPythonObject.cpp
@@ -87,6 +87,7 @@ std::string PropertyPythonObject::toString() const
             throw Py::Exception();
         Py::Callable method(pickle.getAttr(std::string("dumps")));
         Py::Object dump;
+#if PY_VERSION_HEX < 0x03b0000
         if (this->object.hasAttr("__getstate__")) {
             Py::Tuple args;
             Py::Callable state(this->object.getAttr("__getstate__"));
@@ -98,6 +99,9 @@ std::string PropertyPythonObject::toString() const
         else {
             dump = this->object;
         }
+#else
+        dump = this->object;
+#endif
 
         Py::Tuple args(1);
         args.setItem(0, dump);
--- a/src/App/PropertyPythonObject.cpp
+++ b/src/App/PropertyPythonObject.cpp
@@ -87,7 +87,6 @@ std::string PropertyPythonObject::toString() const
             throw Py::Exception();
         Py::Callable method(pickle.getAttr(std::string("dumps")));
         Py::Object dump;
-#if PY_VERSION_HEX < 0x03b0000
         if (this->object.hasAttr("__getstate__")) {
             Py::Tuple args;
             Py::Callable state(this->object.getAttr("__getstate__"));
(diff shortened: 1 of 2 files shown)


src/Mod/Start/StartPage/StartPage.py
```diff
@@ -789,9 +790,13 @@ def exportTestFile():
 
     "Allow to check if everything is Ok"
 
-    f = open(os.path.expanduser("~") + os.sep + "freecad-startpage.html", "w")
-    f.write(handle())
-    f.close()
+    with codecs.open(
+        os.path.expanduser("~") + os.sep + "freecad-startpage.html",
+        encoding="utf-8",
```
benj5378, line 795: Will this always be UTF-8?

chennes: It's considered a best practice in Python to always explicitly open files (for reading _and_ writing) with the encoding set. If that file might *not* be UTF-8, the fix is to ensure that the corresponding `open(..., "w")` statement has the encoding parameter set as well.

src/App/PropertyPythonObject.cpp
```diff
@@ -87,6 +87,7 @@ std::string PropertyPythonObject::toString() const
             throw Py::Exception();
         Py::Callable method(pickle.getAttr(std::string("dumps")));
         Py::Object dump;
+#if PY_VERSION_HEX < 0x03b0000
```
wwmayer, line 90: I wonder why this change would be needed. For Python features that implement the __getstate__/__setstate__ functions persistence might be broken.

Syres916: Mistaken commit, intended elsewhere.

wwmayer:
> On Python 3.11 (tested using Debian 12) UnicodeDecodeErrors are raised when running All Tests, this PR addresses those errors.

Today I have built latest master (80a54d51c) on Debian 12 and for me all tests have passed.

wwmayer:
What does this give for you: `import sys; sys.getfilesystemencoding()`

Syres916:
> What does this give for you: `import sys; sys.getfilesystemencoding()`

`'utf-8'`

wwmayer:
OK, that's correct.

yorikvanhavre:
Is this good to go @Syres916 @wwmayer ?

wwmayer:
I don't know. When running the self-tests on my Debian system I didn't run into unicode errors.

Syres916:
Yes, it's ready to merge, I don't understand why my Debian 12 had errors but Werner's didn't. It's backwards compatible to Python 3.8 anyway so I don't see what harm it can do.

yorikvanhavre:
I guess we'll see quickly if it gives problems for some people with very different locales. Let's merge then!

PR #10544, opened 2023-09-04
Proposal: Support TextSignature in PyExport.xml files

jbaehr:
The new optional attribute "TextSignature" allows to specify a signature for build-in methods. The signature is stored in the `__text_signature__` property and thus enables introspection via the `inspect.signature` module on build-in methods, too.
This is used e.g. by the `help()` function to display the signature instead of just "(...)".
The format has to follow [Python's "Argument Clinic"](https://docs.python.org/3/howto/clinic.html#how-to-override-the-generated-signature), e.g.
```xml
   <Methode Name="carbonCopy" TextSignature="($self, srcObjName, asConstruction=True)">
```
The value has to be a valid python signature and does not support return type annotations (as of now, i.e. Python-3.11).

Also note that I'm not sure if `__text__signature__` and the support for this "some_method(...)\n--\n\n" docstring prefix is "public API" or just support for Argument Clinic (which is explicitly declared internal-only for CPython). It has been around since Python-3.4 though, so for quite some time.

What are your thoughts about this? Is this worth perusing? Should we better discuss this in the forum?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,6 +20,7 @@ files: |
     )
 exclude: |
     (?x)^(
+        src/Tools/generateBase/generateModel_Module.py|
         .*vcproj.*|
         .*vcxproj.*
     )
--- a/src/Tools/generateBase/generateMetaModel_Module.xsd
+++ b/src/Tools/generateBase/generateMetaModel_Module.xsd
@@ -16,6 +16,15 @@
 										<xs:element ref="Parameter" minOccurs="0" maxOccurs="unbounded"/>
 									</xs:sequence>
 									<xs:attribute name="Name" type="xs:string" use="required"/>
+									<xs:attribute name="TextSignature" type="xs:string" use="optional">
+                                        <xs:annotation>
+                                            <xs:documentation>
+                                                Specifies the __text__signature__ for the method.
+                                                This is similar to Argument Clinic's @text_signature,
+                                                see https://docs.python.org/3/howto/clinic.html#how-to-override-the-generated-signature
+                                            </xs:documentation>
+                                        </xs:annotation>
+                                    </xs:attribute>
 									<xs:attribute name="Const" type="xs:boolean" use="optional"/>
 									<xs:attribute name="Keyword" type="xs:boolean" use="optional" default="false"/>
 									<xs:attribute name="Class" type="xs:boolean" use="optional" default="false"/>
--- a/src/Tools/generateBase/generateModel_Module.py
+++ b/src/Tools/generateBase/generateModel_Module.py
@@ -32,31 +32,28 @@ from xml.dom import Node
 # Support/utility functions.
 #
 
-
 def showIndent(outfile, level):
     for idx in range(level):
-        outfile.write("    ")
-
+        outfile.write('    ')
 
 def quote_xml(inStr):
     s1 = inStr
-    s1 = s1.replace("&", "&amp;")
-    s1 = s1.replace("<", "&lt;")
-    s1 = s1.replace('"', "&quot;")
+    s1 = s1.replace('&', '&amp;')
+    s1 = s1.replace('<', '&lt;')
+    s1 = s1.replace('"', '&quot;')
     return s1
-
 
 def quote_python(inStr):
     s1 = inStr
     if s1.find("'") == -1:
-        if s1.find("\n") == -1:
+        if s1.find('\n') == -1:
             return "'%s'" % s1
         else:
             return "'''%s'''" % s1
     else:
         if s1.find('"') != -1:
             s1 = s1.replace('"', '\\"')
-        if s1.find("\n") == -1:
+        if s1.find('\n') == -1:
             return '"%s"' % s1
         else:
             return '"""%s"""' % s1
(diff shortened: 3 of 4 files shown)


wwmayer:
Apparently excluding src/Tools/generateBase/generateModel_Module.py from auto-formatting doesn't work. And IMO, the applied auto-formatting is a regression of readability.

jbaehr:
> Apparently excluding src/Tools/generateBase/generateModel_Module.py from auto-formatting doesn't work.

hmm.. the unfortunate truth is, that I don't have much of a clou about this pre-commit stuff. maybe @chennes can weight in?

> And IMO, the applied auto-formatting is a regression of readability.

Do you mean the auto-formatting done by the pre-commit hook (commit 4a7e1b6) or my undoing of it (due to the code being generator-output, not hand-written. commit [aa8e19f](https://github.com/FreeCAD/FreeCAD/pull/10544/commits/aa8e19f0b8fc3e86f2c0c63dfb71ee46c5cdc617))?

Without the auto-formatting, the changes on the generated file become more apparent (or rather: visible at all): commit [7ec36f2](https://github.com/FreeCAD/FreeCAD/pull/10544/commits/7ec36f2db1455ea76a5e7684c0d7b13ed1abd382). That's why I split the change in two commits.
Of course we could apply the same auto-formatting on the regenerated file again, but I don't like messing around in generated files.

----

In all cases, this is only worth pursuing if you* consider `__text_signature__` as a feasible way for making our C++-implemented python methods inspectable. It comes with limitations (e.g. no return type annotations or no "overloads") but is fairly easy to add.
Then, once signatures can be requested for a significant part of our API, we can make use of them not only in `help()` but also e.g. in the auto-completion pop-up of our python console. Right now, we're in a no-chicken-no-egg situation.

*) not addressing you personally here, but rather the community as a whole

chennes:
> hmm.. the unfortunate truth is, that I don't have much of a clou about this pre-commit stuff. maybe @chennes can weight in?

According to the docs that exclusion should work:
https://pre-commit.com/#regular-expressions

luzpaz:
What's the path forward here ?

wwmayer:
It's still in Draft mode. As long as it's not ready for review there is nothing I can do.

jbaehr:
It's in draft mode as I'm unsure whether this is the right approach to support signatures for C++-implemented Python methods. I have hoped for some discussion about this prior to someone just hitting the merge button. My questions above go beyond the code only.

If your interpretation of "draft" is "not worth looking at it yet", I have no problems in marking this PR as "ready". Sorry if this caused misunderstandings.

wwmayer:
> If your interpretation of "draft" is "not worth looking at it yet"

For me draft means that a PR is still WIP, i.e. it's not finished yet.

> It's in draft mode as I'm unsure whether this is the right approach to support signatures for C++-implemented Python methods.

I think it's fine in cases where the additional markers can go to the generated file. But I have to test the PR to see that it behaves as described in https://devguide.python.org/development-tools/clinic/.

wwmayer:
Your PR does only the very first step but is far from being complete. In the script temlplazeClassPyExport.py are missing all the markers like `[clinic input]` or `[clinic start generated code]`

Then the Argument Clinic stuff is supposed to replace code where `PyArg_ParseTuple*` is currently used. Thus, the script should generate a different function body if `TextSignature` is set as attribute.

jbaehr:
It is not my intention, at least not now, to introduce argument-clinic. From my understanding, clinic solves the same problem as FC's python template and xml data: Generate the boiler plate code. Both approaches to this problem differ quite a lot, though.
* FreeCAD uses XML to describe the classes, methods and doc strings, then some python code generates headers and method stubs as separate files (out of the source tree). The developer then implements the actual methods, including all the argument parsing/conversion. It looks like this stuff predates argument-clinic, too.
* argument-clinic, on the other hand, leverages some DSL to describe the methods and its arguments in source code comments, which then gets expanded (as part of the source tree). The generated code also takes care about the argument parsing and conversion, so the developer "only" implements a method that already has the C equivalent of the python signature. I.e. the developer usually does not have to call `PyArg_ParseTuple` and friends any longer.

Again a little later, with Python-3.4, some extension has been introduced to smuggle a text version of the method's python signature into the definition of C-implemented methods without altering Python's C-API for method definition: If the provided "internal docstring" starts with "methodname( ... )\n--\n\n", the "methodname( ... )" part ends up in a new attribute `__text_signature__`, to be later used to generate a real Signature object for `__signature__` (cf. the [inspect](https://docs.python.org/3/library/inspect.html) module). The remaining part goes to `__doc__`, just as before. (For details you could use [`find_signature` in typeobject.c](https://github.com/python/cpython/blob/a254120f2f1dd99fa64f12594d1ed19c67df7d64/Objects/typeobject.c#L484) as an entry point)

What I now put to discussion, with this PR, is to extend our existing XML-based mechanism to feed Python's C-API with same "text signature as part of the internal docstring". Nothing more. This already allows some introspection for our "build-in" methods. But the argument parsing code still has to be written manually -- no change there.

I agree generating the argument parsing and conversion code would be very nice, too. But this is a *much* bigger undertaking. I have some vague ideas how this can be done with some backward-compatible extensions to our existing XML data, however the now proposed `TextSignature`-attribute is by far not enough for this. Maybe some more complex XML structure describing signatures and data conversion, maybe reuse as much of clinic's DSL as possible and just add a `<clinic>` text node. But we have so many existing XML descriptions that I don't think overthrowing this approach completely would be wise.

wwmayer:
> If the provided "internal docstring" starts with "methodname( ... )\n--\n\n",

Ah, now I start to understand its purpose. I manually edited the doc string of MatrixPy.move and there is indeed a `__text_signature__` attribute with the content of what's inside the braces. Other methods also have this attribute but for them their value is None.

> to be later used to generate a real Signature object for __signature__ (cf. the [inspect](https://docs.python.org/3/library/inspect.html) module).

And how does this work? Can you give an example, please?

>  But we have so many existing XML descriptions that I don't think overthrowing this approach completely would be wise.

I agree.

maxwxyz:
@jbaehr will you work on that PR after 1.0?

jbaehr:
> @jbaehr will you work on that PR after 1.0?

Yes, but it's not very high on my priority list

jbaehr:
closing this in favour of the proposal from #19516.

jbaehr:
@tritao the `__text__signature__` is
a) rather limited (e.g. no return type annotations or no "overloads"), and
b) *maybe* an CPython implementation detail, i.e. no public API.

Still it's in my eyes way better than nothing: For a large percentage of our python methods it would already allow sensible introspection. The most visible benefit is that it enables `help(some_func)` to give you some_func's signature instead of just "some_func(...)". For everything defined in Python this works out of the box, but when the actual code is C++, Python needs some help to get this information.

Now with your proposal from #19516, we don't need some additional xml attribute any longer -- we do have a python definition of the methods in question. That gives us a lot more options for future extensions. In the first step, after your proposal has been accepted, we can extend it with "text signatures", similar to what I've proposed here. But opposed to my proposal, we can do this entirely in the generator: We have the indented python signature already, so no need for additional input from the developer. Still we need to derive it and put it in the generated C++ code (where applicable), like I've shown in this PR here.

For more advanced scenarios, I guess we need to *evaluate* the python stubs instead of (only) looking at the AST. The inspect module can give us "real" [Signature](https://docs.python.org/3/library/inspect.html#inspect.Signature) objects which we can use to generate more advanced C++ stubs that already include the argument parsing with `PyArg_ParseTuple` and friends (similar to Argument-Clinic) what is currently done manually. And I'm sure we can use it generate code that provides those signatures at runtime as well, but I have not investigates this path yet.

Issue #10460, opened 2023-08-31
[Problem] App::PropertyPythonObject is not saving data 


Serince:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
[code]
OS: Pop!_OS 22.04 LTS (pop:GNOME/pop)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33675 (Git)
Build type: Release
Branch: (HEAD detached at 0.21.0)
Hash: 41b058e2087de60dc8fef3d3e68c7d0129e13abf
Python 3.11.4, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * aa.qrc
  * FEMbyGEN 1.5.0
  * Plot 2022.4.17
  * Ship 2022.4.11
  * CfdOF 1.19.4
[/code]
```

### Subproject(s) affected?

None

### Problem description

I use App::PropertyPythonObject to store some data in an object. When I save the file and open it again, no data comes with the created PropertyPythonObject  in 0.21 version. When I try the same process in version 0.20.2, it works fine.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
@Serince can you demonstrate the bug ?

Serince:
Sorry, I checked several times before I posted the error and there was indeed such a bug. Now when I try it on another computer to show it I don't have this problem, I think there is something wrong with my configuration. Let me reinstall freecad. I apologize for opening an issue without investigating the bug in more detail.

Serince:
Hello,

I catch the error again. The problem comes with python 3.11 . I have installed two conda environment to test it. The data is storing freecad 0.21.1 Python 3.10. But freecad 0.21.1 Python 3.11 has the bug. Not storing data in object as ProperyPythonObject. I add a small macro to store data. After running it, save the file and close / reopen the file. When you call the stored data it will return None.

import FreeCAD as App
doc=App.newDocument()
obj=doc.addObject("Part::Box","Box")
doc.recompute()
obj.addProperty("App::PropertyPythonObject", "Dictionary")
obj.Dictionary={"Stored data": [3,5,7]}
print(obj.Dictionary)

![before](https://github.com/FreeCAD/FreeCAD/assets/10951891/a98c1164-89ff-475e-9498-a9ad1639dc69)
[...]

PR #10551, opened 2023-09-05
[App] PropertyPythonObject Python 3.11 fix

Syres916:
Fixes https://github.com/FreeCAD/FreeCAD/issues/10460

Tested on both Debian 12 (Python 3.11.2) and Linux Mint (Python 3.8.10)

### DIFF
--- a/src/App/PropertyPythonObject.cpp
+++ b/src/App/PropertyPythonObject.cpp
@@ -87,6 +87,7 @@ std::string PropertyPythonObject::toString() const
             throw Py::Exception();
         Py::Callable method(pickle.getAttr(std::string("dumps")));
         Py::Object dump;
+#if PY_VERSION_HEX < 0x03b0000
         if (this->object.hasAttr("__getstate__")) {
             Py::Tuple args;
             Py::Callable state(this->object.getAttr("__getstate__"));
@@ -98,6 +99,9 @@ std::string PropertyPythonObject::toString() const
         else {
             dump = this->object;
         }
+#else
+        dump = this->object;
+#endif
 
         Py::Tuple args(1);
         args.setItem(0, dump);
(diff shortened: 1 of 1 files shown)


Syres916:
@wwmayer In my testing on Py3.11 versus Py3.8, on Py3.11 the `__getstate__` condition was being tripped but on Py3.8 with exactly the same data neither `__getstate__` nor `__dict__` were found. If you feel that in itself needs further investigation then let me know but the `Py::Callable method(pickle.getAttr(std::string("dumps")));` is proving difficult for me to evaluate, getting out of comfort zone.

wwmayer:
> Tested on both Debian 12 (Python 3.11.2)

I have a Debian 12 running too and when I ran the self-tests the last time all of them has succeeded. But I am not sure whether there is a specific unit test to check the `__getstate__/__setstate__` stuff. If not then one must be written.

I am very skeptical that with Py3.11 passing the object directly works reliably because there exists data types (like Coin3D nodes) that cannot be made persistent (at least with < Py3.11) and that's one of the reason of having `__getstate__/__setstate__`

So, here some more tests are required to see what works and what not.

Syres916:
I found this quote `Changed in version 3.11: Added the default implementation of the __getstate__() method in the [object](https://docs.python.org/3/library/functions.html#object) class.` taken from https://docs.python.org/3/library/pickle.html#pickle-inst

wwmayer:
Our first version to serialize Python objects was indeed based on pickle, hence the method names `__getstate__/__setstate__.` Because it's considered unsafe (you can execute arbitrary Python code when restoring an object) we switched to json but we left these two function names untouched.

wwmayer:
The problem is that since Py3.11 the very base class `object` now has the method `__getstate__` and that's why PropertyPythonObject uses this method when it shouldn't. I think the best what we can do is to rename our `__getstate__/__setstate__` to `__get_state__/__set_state__`

wwmayer:
This PR fixes the issue: https://github.com/FreeCAD/FreeCAD/pull/10769

Issue #10122, opened 2023-08-08
[Problem] 8514oem font results in error after FreeCAD start


KAPO62:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
DirectWrite: CreateFontFaceFromHDC() failed (Gibt einen Fehler in einer Eingabedatei an, zum Beispiel eine Schriftartdatei.) for QFontDef(Family="8514oem", pointsize=10, pixelsize=20, styleHint=5, weight=50, stretch=100, hintingPreference=0) LOGFONT("8514oem", lfWidth=0, lfHeight=-20) dpi=120
```

### Subproject(s) affected?

None

### Problem description

DirectWrite: CreateFontFaceFromHDC() failed (Gibt einen Fehler in einer Eingabedatei an, zum Beispiel eine Schriftartdatei.) for QFontDef(Family="8514oem", pointsize=10, pixelsize=20, styleHint=5, weight=50, stretch=100, hintingPreference=0) LOGFONT("8514oem", lfWidth=0, lfHeight=-20) dpi=120

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Syres916:
To assist in others finding this in future, could you please amend the title to something that includes _8514oem_ as this is the critical word in finding the solution which is https://forum.freecad.org/viewtopic.php?style=1&p=654799#p654799
The types of fonts allowed for the Macro Editor, Report View and Python Console have been slimmed down from previous versions.

G0MJW:
It is happening to me with a clean install of 0.21 in Windows 11

21:21:27  DirectWrite: CreateFontFaceFromHDC() failed (Indicates an error in an input file such as a font file.) for QFontDef(Family="8514oem", pointsize=10, pixelsize=20, styleHint=5, weight=50, stretch=100, hintingPreference=0) LOGFONT("8514oem", lfWidth=0, lfHeight=-20) dpi=144

I think this might have happened because I had to make the sketcher font larger as it is far too tiny to read with a high DPI monitor.

I see several posts but no actual solution, apart from a reference to an editor menu in preferences which does not appear.

OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.21.0.33668 +7 (Git)
Build type: Release
Branch: (HEAD detached at 0.21)
Hash: 41b058e2087de60dc8fef3d3e68c7d0129e13abf
[...]

Issue #10514, opened 2023-09-03
[Problem] Failed to compute left/right minimum bearings for "cursor.pcf"


ville-h:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

If I try to use the provided "Dark theme" or the "Light theme" preference packs and then try to open the "view > panels > report view" freecad starts reporting: `Failed to compute left/right minimum bearings for "cursor.pcf"` repeatedly. Eventually consumes 100% of a core and becomes unresponsive.

I also tried getting other themes with "tools > addon manager" and they exhibited the same behaviour.

Reason I'd like to use the "Dark theme" is that otherwise I'll have white text on white background in the spreadsheet rendering it unusable. Should this be reported separately?

### Full version info

```shell
[code]
OS: Arch Linux (openbox)
Word size of FreeCAD: 64-bit
Version: 0.21.1.33694 (Git)
Build type: Release
Branch: makepkg
Hash: f6708547a9bb3f71a4aaade12109f511a72c207c
Python 3.11.5, Qt 5.15.10, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Locale: English/United States (en_US)
[/code]
```

### Subproject(s) affected?

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
@MisterMakerNL have you encountered this before ?

luzpaz:
@ville-h what version of fontconfig is Arch using ?

ville-h:
On my system:
```
extra/fontconfig 2:2.14.2-1 [installed]
```

cfillion:
Replacing `<FCText Name="Font">cursor.pcf</FCText>` to `monospace` in ~/.config/FreeCAD/user.cfg under `<FCParamGroup Name="Editor">` seems to have fixed it for me. The log is no longer displayed in a huge font, and FreeCAD no longer freezes in an infinite loop of "Failed to compute left/right minimum bearings" errors.

ville-h:
Based on cfillion's message decided to take a closer look. Running `find / -name '*cursor.pcf' 2>/dev/null` on my system doesn't produce any results. Running `find / -name '*cursor.pcf.gz' 2>/dev/null` does. Removing `.pcf` and using `<FCText Name="Font">cursor</FCText>` works. So few observations:

0) Using a name like: `cursor.pcf` rather than `cursor` your themes rely on the font file being in an specific format on disk

[...]

PR #10555, opened 2023-09-05
[Preferences] Exclude 8514oem from Monospace fonts

Syres916:
Fixes https://github.com/FreeCAD/FreeCAD/issues/10122

The 8514oem font is added when Windows users change DPI to for example 125% see https://answers.microsoft.com/en-us/windows/forum/all/windows-add-font-8514oem-when-i-change-windows-dpi/2ec32e81-52d9-40a0-ab35-07a8ef9051b8
which in turn results in a warning in the Report View which makes little sense to the user.

### DIFF
--- a/src/Gui/PreferencePages/DlgSettingsEditor.cpp
+++ b/src/Gui/PreferencePages/DlgSettingsEditor.cpp
@@ -26,8 +26,6 @@
 # include <QFontDatabase>
 #endif
 
-#include <boost/algorithm/string.hpp>
-
 #include <App/Color.h>
 #include <Gui/PythonEditor.h>
 #include <Gui/Tools.h>
@@ -303,8 +301,7 @@ void DlgSettingsEditor::loadSettings()
     QStringList fixedFamilyNames;
     for (const auto &name : familyNames) {
         if (QFontDatabase().isFixedPitch(name)) {
-            std::string utf8_name = name.toUtf8().constData();
-            if (boost::iequals(utf8_name, "8514oem")) {
+            if (name.compare(QLatin1String("8514oem"), Qt::CaseInsensitive) != 0) {
                 fixedFamilyNames.append(name);
             }
         }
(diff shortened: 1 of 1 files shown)


src/Gui/PreferencePages/DlgSettingsEditor.cpp
```diff
@@ -301,15 +303,21 @@ void DlgSettingsEditor::loadSettings()
     QStringList fixedFamilyNames;
     for (const auto &name : familyNames) {
         if (QFontDatabase().isFixedPitch(name)) {
-            fixedFamilyNames.append(name);
+            std::string utf8_name = name.toUtf8().constData();
+            if (boost::iequals(utf8_name, "8514oem")) {
```
wwmayer, line 307: This filters out everything but 8514oem. So, it does the opposite. And since we already use QString I don't see the necessity to switch to another lib for string comparison.

IMO, the easiest solution would be to remove "8514oem" after the string list has been filled:
`fixedFamilyNames.removeAll(QString::fromLatin1("8514oem"));`

Syres916: Thanks Werner, I obviously need some sleep. I'm not relying on the font definitively staying in that case so prefer to have case insensitivity.

sliptonic:
@Syres916   Please update and resolve the conversation when done.

Syres916:
This is ready to merge, please.

luzpaz:
bumping for visibility

yorikvanhavre:
merging then!  ;)

adrianinsaval:
might be worth blacklisting "cursor.pcf" too: #10514

Manuelzwei:
See #12915: 
I set my DPI to 100% and the error was gone.
But my icons are now very small.
So I set the DPI to 125% and live with this error - since you fix it.

PR #10556, opened 2023-09-05
Draft: arcTracker did not take Working Plane rotation into account

Roy-043:
Steps to verify the issue:
1. Open FreeCAD.
2. Switch to Std_ViewTop.
3. Click one of the curved arrows of the Navigation Cube.
4. Switch to the Draft WB.
5. Make sure Draft_SelectPlane is set to "Auto".
6. Start Draft_Arc and click 2 points (for the radius and the start angle).
7. Result: The displayed arc does not match the start angle or the current point.

Additonally:
* Removed the normal argument from `__init__`. The tracker does not work properly with just a normal. AFAICT there is no code that specifies this normal.
* Minor improvement to `getAngle`: Switched the vectors.

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_trackers.py
+++ b/src/Mod/Draft/draftguitools/gui_trackers.py
@@ -530,7 +530,7 @@ class arcTracker(Tracker):
     """An arc tracker."""
 
     def __init__(self, dotted=False, scolor=None, swidth=None,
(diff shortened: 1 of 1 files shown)


yorikvanhavre:
I don't remember why that `normal` argument was there but if you checked and it's not used, then I guess it's fine to remove it. Indeed as you say it's not enough to get a correct positioning.

PR #10563, opened 2023-09-05
Update FindPySide2Tools.cmake

mosfet80:
Set miminum qt =5.15 into find pyside.cmake


chennes:
Where are you getting the 5.15 requirement? I though Ubuntu 20.04 only supported Qt 5.12?

mosfet80:
sorry!! I had checked only ubuntu 22 and windows. ... I will submit it patch in April 2025 (eol ubuntu20.04 :) )

PR #10573, opened 2023-09-06
Core: Selection : Add possibility to set the style of selection

PaddleStroke:
Add the possibility to set the style of selection between greedy and normal.
This is usefull for tools that require a greedy selection, like partDesign Fillet.
Also makes the python interface.

@wwmayer can you please have a look ?
There is one little thing that can be improved it's that currently you do 
`Gui.Selection.setSelectionStyle(0)`
when it would be best to have an enum to do 
`Gui.Selection.setSelectionStyle(Gui.Selection.SelectionStyle.GreedySelection)`
But I could not find how to implement this enum in python. I tried to look at `Gui.Selection.ResolveMode.NoResolve` but couldn't find where it's implemented.

### DIFF
--- a/src/Gui/FreeCADGuiInit.py
+++ b/src/Gui/FreeCADGuiInit.py
@@ -38,8 +38,7 @@ Gui = FreeCADGui
 Gui.listCommands = Gui.Command.listAll
 Gui.isCommandActive = lambda cmd: Gui.Command.get(cmd).isActive()
 
-# The values must match with that of the
-# C++ enum class ResolveMode
+# The values must match with that of the C++ enum class ResolveMode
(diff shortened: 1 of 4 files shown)


wwmayer (changes requested):
> But I could not find how to implement this enum in python. I tried to look at Gui.Selection.ResolveMode.NoResolve but couldn't find where it's implemented.

This is implemented in FreeCADGuiinit.py

wwmayer:
After adding this Python enum to FreeCADGuiInit.py it's fine to merge the PR.

PaddleStroke:
@wwmayer thanks. It is done.

PR #10576, opened 2023-09-06
Removed references to soqt (deprecated)

mosfet80:
Removed references to soqt (removed into freecad)


wwmayer:
Removing the stuff from FCConfig.h is OK. Removing the Python binding for the off-screen rendering is not OK. Although the class uses SoQt as prefix it has nothing to do with the SoQt library.

PR #10577, opened 2023-09-06
Update sub_buildWindows.yml

mosfet80:
increase size of ccache.
Freecad uses more than 2 GB


0penBrain:
AFAIK the rationale is false.
A "one-time" Windows Ccache is somewhere around 300 MB IIRC (with currently applied compression level).
So this PR seems counter-productive as it will significantly increase the Windows cache and thus reduce the number of PR caches we can keep in the buffer.

chennes:
I'm going to close this (unmerged). Feel free to re-open with an explanation if you think this is incorrect, @mosfet80.

PR #10579, opened 2023-09-06
format: add spaces before trailing comment

wwmayer:
As discussed in https://forum.freecad.org/viewtopic.php?p=703967#p703967


chennes:
Were you trying to use [skip ci] to avoid the pre-commit.ci hook? I'm surprised it ran with that tag on the commit.

wwmayer:
> Were you trying to use [skip ci] to avoid the pre-commit.ci hook?

Not to avoid the pre-commit hook but to avoid to compile the source code.

wwmayer:
Replaced with https://github.com/FreeCAD/FreeCAD/pull/10658

PR #10581, opened 2023-09-06
Test: add unit tests for PropertyPythonObject

wwmayer:
Adds some tests to demonstrate issue #10460 with Py3.11

### DIFF
--- a/src/App/PropertyPythonObject.cpp
+++ b/src/App/PropertyPythonObject.cpp
@@ -92,6 +92,14 @@ std::string PropertyPythonObject::toString() const
             Py::Callable state(this->object.getAttr("dumps"));
             dump = state.apply(args);
         }
+#if PY_VERSION_HEX < 0x030b0000
+        // support add-ons that use the old method names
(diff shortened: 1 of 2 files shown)


chennes:
@wwmayer do you want to hold off on merging this until there is a resolution for the bug, or is it OK to merge now?

wwmayer:
If we merge the PR now there will be failures with the self-tests for Py3.11 builds. A solution for the actual issue is to rename the methods to avoid a clash with the methods of Python's object class.

sliptonic:
We're making more use of Draft status now to keep the PR queue clean and let maintainers know what is ready for merge.
This is a different case.  Maybe we need to prepend the title with a [HOLD]  or add a label to PRs that are, themselves, ready to go but depend on something else to happen first.

wwmayer:
I don't mind if the PR is merged now because the people with a Py3.11 anyway already suffer from the problem that objects are not correctly restored.

wwmayer:
The PR is now based on https://github.com/FreeCAD/FreeCAD/pull/10769

Issue #10536, opened 2023-09-04
[Problem] Scaling a Part shape is awkward


WandererFan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Current situation:
In order to scale a shape, one must load the Draft workbench and create a Clone of the shape.  3 properties in the Clone must then be updated with the scale factor.
There is also a Scale tool in Draft, but it does not perform a simple "scale to 2x" operation.

Desired situation:
Select one or more source shapes, press a Scale button in the Part toolbar.  This would create a "Scaled" DocumentObject whose Shape property contains a scaled version of the source shapes.

### Full version info

```shell
v0.22
```

### Subproject(s) affected?

Part

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Roy-043:
You can also create a Link which has a `Scale` and a `Scale List` property.

luzpaz:
> I think I missed something for objects that are already selected when the tool is started.

WandererFan:
The current task dialog does not accept negative factors, but they can be set in the property editor afterwards.  I will change the dialog to accept negative numbers from the outset.

I don't know how to make the factors expression aware yet.

wwmayer:
A widget can only be bound to a property of a specific scale object. But because the scale objects will be created after clicking the Apply/OK button this isn't possible.

All what would be possible is to implement the edit function of the view provider of an existing scale object so that after its creation the properties can be bound to the widgets. Therefore the scale dialog must be reworked that an existing scale object can be passed and that in this state it will only allow to modify this object and not to create new scale objects.

WandererFan:
Task dialog has been updated to allow negative factors.

As werner pointed out, expressions can only be used once the object exists, so can't be applied in the creation dialog.  Once the object exists, you can apply an expression to the Scale object in the PropertyEditor.

PR #10583, opened 2023-09-06
[Part]add scale command & feature

WandererFan:
This PR provides a solution to #10536 Scaling a Part shape is awkward.  It adds a scale function to the Part module.   The code is largely based on FeatureExtrusion and related code for task dialog and viewprovider.

### DIFF
--- a/src/Mod/Part/App/FeatureScale.cpp
+++ b/src/Mod/Part/App/FeatureScale.cpp
@@ -1,0 +1,162 @@
+/***************************************************************************
+ *   Copyright (c) 2023 Wanderer Fan <wandererfan@gmail.com>               *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This library is free software; you can redistribute it and/or         *
+ *   modify it under the terms of the GNU Library General Public           *
+ *   License as published by the Free Software Foundation; either          *
+ *   version 2 of the License, or (at your option) any later version.      *
+ *                                                                         *
+ *   This library  is distributed in the hope that it will be useful,      *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "PreCompiled.h"
+#ifndef _PreComp_
+#include <BRepBuilderAPI_GTransform.hxx>
+#include <BRepBuilderAPI_Transform.hxx>
+#include <BRepBuilderAPI_Copy.hxx>
+#include <gp_Pnt.hxx>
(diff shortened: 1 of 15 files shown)


src/Mod/Part/App/FeatureScale.h
```diff
@@ -0,0 +1,90 @@
...
+    App::PropertyFloat YScale;
+    App::PropertyFloat ZScale;
+
+    /**
+     * @brief The ScaleParameters struct is supposed to be filled with final
+     * scale parameters, after resolving links, applying mode logic,
+     * reversing, etc., and be passed to scaleShape.
+     */
+    struct ScaleParameters {
+        bool uniform;
```
wwmayer, line 56: Since all other member variables are initialized this should be, too: `bool uniform{true};`

src/Mod/Part/Gui/DlgScale.cpp
```diff
@@ -0,0 +1,363 @@
...
+    ui->dsbXScale->setDecimals(Base::UnitsApi::getDecimals());
+    ui->dsbYScale->setDecimals(Base::UnitsApi::getDecimals());
+    ui->dsbZScale->setDecimals(Base::UnitsApi::getDecimals());
+    findShapes();
+}
+
+/*
+ *  Destroys the object and frees any allocated resources
+ */
+DlgScale::~DlgScale()
```
wwmayer, line 76: According to the rules to modernize C++ code this should be written as: `DlgScale::~DlgScale() = default;`
Alternatively, you can completely drop the destructor.

src/Mod/Part/App/FeatureScale.cpp
```diff
@@ -0,0 +1,162 @@
...
+    mat.SetValue(3,2,matScale[2][1]);
+    mat.SetValue(1,3,matScale[0][2]);
+    mat.SetValue(2,3,matScale[1][2]);
+    mat.SetValue(3,3,matScale[2][2]);
+    mat.SetValue(1,4,matScale[0][3]);
+    mat.SetValue(2,4,matScale[1][3]);
+    mat.SetValue(3,4,matScale[2][3]);
+
+    // this copy step seems to eliminate Part.OCCError: gp_GTrsf::Trsf() - non-orthogonal GTrsf
+    // which may to be related to the tesselation of the input shape.  See Github issue #9651
```
wwmayer, line 130: Typo checker says: `tesselation ==> tessellation`

src/Mod/Part/App/FeatureScale.h
WandererFan: fixed.

src/Mod/Part/Gui/DlgScale.cpp
WandererFan: fixed

src/Mod/Part/App/FeatureScale.cpp
WandererFan: fixed.

WandererFan:
Thanks for the review.

FlachyJoe:
I can't get this working, even with a simple cube. Do I miss something?
image

WandererFan:
You have to select "Cube" in the Shape list.   I think I missed something for objects that are already selected when the tool is started.

Roy-043:
The command now sits between Part_Extrude and Part_Revolve in the toolbar. IMO it would be more logical to place it after Part_Revolve since that command and Part_Extrude have some similarities.

wwmayer:
Agreed. I think we should move it after the Mirror command.

maxwxyz:
I have issues with this tool and created a new one here: https://github.com/FreeCAD/FreeCAD/issues/15832

PR #10602, opened 2023-09-08
[Part]correct handling of preselected objects

WandererFan:
This PR corrects the handling of preselected objects in the new FeatureScale.


wwmayer:
When I create a Part tube (a hollow cylinder) then the dialog will ignore it. Responsible is the method DlgScale::canScale with this check:
`xp.Init(shape, TopAbs_SHAPE);`
This is not a valid way to explore sub-shapes of a compound. You have to explicitly check for a type like TopAbs_EDGE. If there is at least one such element you can assume to have a scalable shape.

Issue #10467, opened 2023-08-31
[Problem] User can't indicate what type of constraint to add


sliptonic:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Full version info

N/A

### Subproject(s) affected?

None

### Problem description

A fundamental feature of the assembly workbench is for the user to constrain how one part moves relative to another.  Numerous different types of constraints are possible and it should be efficient for the user to indicate exactly what type of constraint to apply based on what geometry is selected and what else the user chooses.  For example, the user may select a single face from each of two solids. They may wish to make the two faces co-planar or simple parallel.

The user needs to be able to select geometry from two parts and apply an appropriate constraint between them.

This issue will start the discussion about how the user workflow will function for applying constraints.

### Anything else?

[constraint grid](https://help.spaceclaim.com/2015.0.0/en/Content/Printable/SpaceClaimAssemblyConstraintsChart.pdf)

[constraint type overview](https://help.alibre.com/articles/#!alibre-help-v27/assemblies-assembly-constraints-types-of-assembly-constraints-overview)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PaddleStroke:
There are different approach to this problem.
- Solidworks approach is similar to A3 and A2+ : You select features of the part (faces, edges, points) then you choose what kind of constraint you want to apply: parallel, distance, co-axial...
- OnShape/Fusion approach are using 'mate connectors' (a bit similar to A4) a short video will be more efficient than words : https://www.onshape.com/en/resource-center/videos/build-your-first-assembly

Maybe the first question is which workflow do we want to achieve ?

tobiasfalk:
Maybe a lost of benefits and drawbacks of each method would be good. Also is one method is preventing future features, like that two parts of a assembly cam be moved(in the programm) or that for example a FEM simulation is easier with one?

sliptonic:
Another good reference on constraint types.
https://forum.freecad.org/viewtopic.php?t=80819

FEA-eng:
> or that for example a FEM simulation is easier with one?

That won't be a problem. FEM completely ignores assembly constraints and a user has to define all mechanical constraints (also operating on degrees of freedom but in a more direct way) to be used in the analysis in the FEM workbench.

tobiasfalk:
@FEA-eng 
[...]

PR #10611, opened 2023-09-08
Assembly joints

PaddleStroke:
This is a basic implementation of a constraint system for the assembly wb.

The approach is the one that uses 'joint connectors'. It is a similar approach as A4 but the idea is to generate the placement based on the user selection. So for example you select the plane face4 of your body, the joint-connector placement is going to be the rotation of that plane, and the position of the face4 vertex that is the closer to your mouse. It is also similar to the approach of two well known commercial package.

The reason of this choice is that the solver developped by Dr Koh is following this logic to build assembly. Also this workflow is good and intuitive.

- 'Create Assembly' also create a 'joints' folder.
- Adds 'create joint xxx' commands. These commands prompt the user to select 2 features of 2 different parts/bodies.
- The command creates a 'joint' object that stores the selected objects and selected elements and generated placement in properties.

The joint-connectors are ready to be assembled. At this stage we now need to have the API of the solver ready and accessible to move forward and actually solve the assembly.

Note: 
- Commit 1 concerns core and has a PR already opened : #10573
- Issue https://github.com/FreeCAD/FreeCAD/issues/10467
Early feedback would be appreciated.

### DIFF
--- a/src/Mod/Assembly/CommandCreateJoint.py
+++ b/src/Mod/Assembly/CommandCreateJoint.py
@@ -1,0 +1,505 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# /****************************************************************************
+#                                                                           *
+#    Copyright (c) 2023 Ondsel <development@ondsel.com>                     *
+#                                                                           *
+#    This file is part of FreeCAD.                                          *
+#                                                                           *
+#    FreeCAD is free software: you can redistribute it and/or modify it     *
+#    under the terms of the GNU Lesser General Public License as            *
+#    published by the Free Software Foundation, either version 2.1 of the   *
+#    License, or (at your option) any later version.                        *
+#                                                                           *
+#    FreeCAD is distributed in the hope that it will be useful, but         *
+#    WITHOUT ANY WARRANTY; without even the implied warranty of             *
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU       *
+#    Lesser General Public License for more details.                        *
+#                                                                           *
+#    You should have received a copy of the GNU Lesser General Public       *
+#    License along with FreeCAD. If not, see                                *
+#    <https://www.gnu.org/licenses/>.                                       *
+#                                                                           *
+# ***************************************************************************/
+
+import os
+import FreeCAD as App
+
+from PySide.QtCore import QT_TRANSLATE_NOOP
+
+if App.GuiUp:
+    import FreeCADGui as Gui
+    from PySide import QtCore, QtGui, QtWidgets
+
+import JointObject
+import UtilsAssembly
+import Assembly_rc
+
+# translate = App.Qt.translate
+
+__title__ = "Assembly Commands to Create Joints"
+__author__ = "Ondsel"
+__url__ = "https://www.freecad.org"
+
+
+class CommandCreateJointFixed:
+    def __init__(self):
+        pass
+
+    def GetResources(self):
+
+        return {
+            "Pixmap": "Assembly_CreateJointFixed",
+            "MenuText": QT_TRANSLATE_NOOP("Assembly_CreateJointFixed", "Create Fixed Joint"),
+            "Accel": "F",
+            "ToolTip": QT_TRANSLATE_NOOP(
+                "Assembly_CreateJointFixed",
+                "<p>Create a Fixed Joint: Permanently locks two parts together, preventing any movement or rotation.</p>",
+            ),
+            "CmdType": "ForEdit",
+        }
+
+    def IsActive(self):
+        return UtilsAssembly.activeAssembly() is not None
+
+    def Activated(self):
+        assembly = UtilsAssembly.activeAssembly()
+        if not assembly:
+            return
+        view = Gui.activeDocument().activeView()
+
+        self.panel = TaskAssemblyCreateJoint(assembly, view, 0)
+        Gui.Control.showDialog(self.panel)
+
+
+class CommandCreateJointRevolute:
+    def __init__(self):
+        pass
+
+    def GetResources(self):
+
+        return {
+            "Pixmap": "Assembly_CreateJointRevolute",
+            "MenuText": QT_TRANSLATE_NOOP("Assembly_CreateJointRevolute", "Create Revolute Joint"),
+            "Accel": "R",
+            "ToolTip": QT_TRANSLATE_NOOP(
+                "Assembly_CreateJointRevolute",
+                "<p>Create a Revolute Joint: Allows rotation around a single axis between selected parts.</p>",
+            ),
+            "CmdType": "ForEdit",
+        }
+
+    def IsActive(self):
+        return UtilsAssembly.activeAssembly() is not None
+
+    def Activated(self):
+        assembly = UtilsAssembly.activeAssembly()
+        if not assembly:
+            return
+        view = Gui.activeDocument().activeView()
+
+        self.panel = TaskAssemblyCreateJoint(assembly, view, 1)
+        Gui.Control.showDialog(self.panel)
+
+
+class CommandCreateJointCylindrical:
+    def __init__(self):
+        pass
+
+    def GetResources(self):
+
+        return {
+            "Pixmap": "Assembly_CreateJointCylindrical",
+            "MenuText": QT_TRANSLATE_NOOP(
+                "Assembly_CreateJointCylindrical", "Create Cylindrical Joint"
+            ),
+            "Accel": "C",
+            "ToolTip": QT_TRANSLATE_NOOP(
+                "Assembly_CreateJointCylindrical",
+                "<p>Create a Cylindrical Joint: Enables rotation along one axis while permitting movement along the same axis between assembled parts.</p>",
+            ),
+            "CmdType": "ForEdit",
+        }
+
+    def IsActive(self):
+        return UtilsAssembly.activeAssembly() is not None
+
+    def Activated(self):
+        assembly = UtilsAssembly.activeAssembly()
+        if not assembly:
+            return
+        view = Gui.activeDocument().activeView()
+
+        self.panel = TaskAssemblyCreateJoint(assembly, view, 2)
+        Gui.Control.showDialog(self.panel)
+
+
+class CommandCreateJointSlider:
+    def __init__(self):
+        pass
+
+    def GetResources(self):
+
+        return {
+            "Pixmap": "Assembly_CreateJointSlider",
+            "MenuText": QT_TRANSLATE_NOOP("Assembly_CreateJointSlider", "Create Slider Joint"),
+            "Accel": "S",
+            "ToolTip": QT_TRANSLATE_NOOP(
+                "Assembly_CreateJointSlider",
+                "<p>Create a Slider Joint: Allows linear movement along a single axis but restricts rotation between selected parts.</p>",
+            ),
+            "CmdType": "ForEdit",
+        }
+
+    def IsActive(self):
+        return UtilsAssembly.activeAssembly() is not None
+
+    def Activated(self):
+        assembly = UtilsAssembly.activeAssembly()
+        if not assembly:
+            return
+        view = Gui.activeDocument().activeView()
+
+        self.panel = TaskAssemblyCreateJoint(assembly, view, 3)
+        Gui.Control.showDialog(self.panel)
+
+
+class CommandCreateJointBall:
+    def __init__(self):
+        pass
+
+    def GetResources(self):
+
+        return {
+            "Pixmap": "Assembly_CreateJointBall",
+            "MenuText": QT_TRANSLATE_NOOP("Assembly_CreateJointBall", "Create Ball Joint"),
+            "Accel": "B",
+            "ToolTip": QT_TRANSLATE_NOOP(
+                "Assembly_CreateJointBall",
+                "<p>Create a Ball Joint: Connects parts at a point, allowing unrestricted movement as long as the connection points remain in contact.</p>",
+            ),
+            "CmdType": "ForEdit",
+        }
+
+    def IsActive(self):
+        return UtilsAssembly.activeAssembly() is not None
+
+    def Activated(self):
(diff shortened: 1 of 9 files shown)


src/Mod/Assembly/JointObject.py
```diff
@@ -0,0 +1,227 @@
...
+    - If elt is a vertex, then vtx = elt And placement is vtx coordinates without rotation.
+    - if elt is an edge, then vtx = edge start/end vertex depending on which is closer. If elt is an arc or circle, vtx can also be the center. The rotation is the plane normal to the line positioned at vtx. Or for arcs/circle, the plane of the arc.
+    - if elt is a plane face, vtx is the face vertex (to the list of vertex we need to add arc/circle centers) the closer to the mouse. The placement is the plane rotation positioned at vtx
+    - if elt is a cylindrical face, vtx can also be the center of the arcs of the cylindrical face.
+    """
+
+    def findPlacement(obj, elt, vtx):
+        plc = obj.Placement
+
+    return plc
```
howie-j, line 181:
```suggestion
        return plc
```
return outside of method, gives error when loading workbench:
asm_error

PaddleStroke: Sorry about that typo !

PaddleStroke:
This is now ready for review.

src/Mod/Assembly/CommandCreateJoint.py
```diff
@@ -0,0 +1,482 @@
...
+
+    def GetResources(self):
+        tooltip = "<p>Create a revolute joint between parts of your assembly."
+        tooltip += "This will allow rotation along one axis.</p>"
+
+        return {
+            "Pixmap": "Assembly_CreateJointRevolute",
+            "MenuText": QT_TRANSLATE_NOOP("Assembly_CreateJointRevolute", "Create Revolute Joint"),
+            "Accel": "J,R",
+            "ToolTip": QT_TRANSLATE_NOOP("Assembly_CreateJointRevolute", tooltip),
```
chennes, line 85: This won't work, you can't use a variable in QT_TRANSLATE_NOOP.

src/Mod/Assembly/JointObject.py
```diff
@@ -0,0 +1,421 @@
...
+                self.joint.Object2, self.joint.Element2, self.joint.Vertex2
+            )
+        else:
+            self.joint.Object2 = None
+            self.joint.Element2 = ""
+            self.joint.Vertex2 = ""
+            self.joint.Placement2 = UtilsAssembly.activeAssembly().Placement
+
+    """
+    So here we want to find a placement that correspong to a local cordinate system that would be placed at the selected vertex.
```
chennes, line 173:
```suggestion
    So here we want to find a placement that corresponds to a local coordinate system that would be placed at the selected vertex.
```

src/Mod/Assembly/UtilsAssembly.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def getElementName(full_name):
+    # full_name is "Assembly.Assembly1.Assembly2.Assembly3.Box.Edge16"
+    # We want either Edge16.
+    parts = full_name.split(".")
+
+    if len(parts) < 3:
+        # At minimum "Assembly.Box.edge16". It shouldn't be shorter
+        return ""
```
chennes, line 92: As opposed to raising an exception?

src/Mod/Assembly/UtilsAssembly.py
```diff
@@ -0,0 +1,239 @@
...
+    points = []
+    for vtx in vertexes:
+        points.append(vtx.Point)
+    return points
+
+
+def findClosestPointToMousePos(candidates_points, mousePos):
+    point_min_length = None
+    i = 0
+    for point in candidates_points:
```
chennes, line 223: I'd use `enumerate` here.

src/Mod/Assembly/UtilsAssembly.py
```diff
@@ -0,0 +1,239 @@
...
+            point_min_length = length
+            closest_point_index = i
+        i += 1
+
+    return closest_point_index, point_min_length
+
+
+def findVertexNameInObject(vertex, obj):
+    i = 0
+    for vtx in obj.Shape.Vertexes:
```
chennes, line 235: Again, `enumerate` is more idiomatic.

howie-j:
### Non-code feedback

Note: I've mostly tested the workflow of adding joints, since that is what this PR adresses.

I find it very intuitive and easy to add joints, really great work so far. I also like the agile approach of making small PR's, quickly merging, getting early feedback and solving problems continually.

#### :+1: :+1: :+1:
- The LCS indicator snaps to edges, vertexes, faces and holes mostly as expected, and the preview is awesome
- Nice to be able to edit joint type in task view, and later in property view
- Quick to hide LCS's/joints by clicking "Joints" folder and space bar
- Overall i'm very impressed by the user experience so far. Not only miles ahead of the other wb's, but also on par with commercial systems.

#### :-1: 
- Joint/LCS snap has room for improvement, for example snap to mid line and center of face
- The LCS's / joints are quite big and intrusive, and should perhaps be hidden automatically after a joint is created
- Enter should confirm joint, to be consistent with task panels (PD Pad, Pocket etc)
- Double clicking on a joint in tree view should launch the joint task panel for easy editing
- No way of selecting and removing a joint connector in the task panel (by right clicking in the list for example)
- Selecting a joint connector in the task panel list and pressing delete will delete the last feature of the part, not remove the joint connector
- If you delete the "Joints" folder/group, you get an error when attempting to create a new joint. The create joint command should check if a Joints folder exists, and create a new one if not
- (Very personal opinion) The shortcuts follows the horrible sketcher model of multiple keys, when Assembly currently only has 10 commands. In my opinion it's better to only create default shortcuts for the most used commands, and let the user themself create shortcuts for less used commands. Much better to have 5 or 10 single key shortcuts that the user might remember, than 30 that the user wont.. :)

src/Mod/Assembly/CommandCreateJoint.py
```diff
@@ -0,0 +1,482 @@
...
+    def deactivated(self):
+        Gui.Selection.removeSelectionGate()
+        Gui.Selection.removeObserver(self)
+        Gui.Selection.setSelectionStyle(0)
+        Gui.Selection.clearSelection()
+        self.view.removeEventCallback("SoLocation2Event", self.callbackMove)
+
+    def createJointObject(self):
+        type_index = self.form.jointType.currentIndex()
+
```
howie-j, line 360: Check if group "Joints" exist i assembly, and create if if not

PaddleStroke:
> 1 - [ ] The LCS's / joints are quite big and intrusive, and should perhaps be hidden automatically after a joint is created
> 2 - [ ] Enter should confirm the joint, to be consistent with task panels (PD Pad, Pocket etc)
> 3 - [ ] Double clicking on a joint in tree view should launch the joint task panel for easy editing
> 4 - [ ] No way of selecting and removing a joint connector in the task panel (by right clicking in the list for example)
> 5 - [ ] Selecting a joint connector in the task panel list and pressing delete will delete the last feature of the part, not remove the joint connector
> 6 - [x] If you delete the "Joints" folder/group, you get an error when attempting to create a new joint. The create joint command should check if a Joints folder exists, and create a new one if not
> 7 - [ ] Joint/LCS snap has room for improvement, for example snap to mid line and center of face (nitpicking, not important atm)
> 8 - [x] (Very personal opinion) The shortcuts follows the horrible sketcher model of multiple keys, when Assembly currently only has 10 commands. In my opinion it's better to only create default shortcuts for the most used commands, and let the user themself create shortcuts for less used commands. Much better to have 5 or 10 single key shortcuts that the user might remember, than 30 that the user wont.. :)

- 1, 3, 4, 7 : Yes I agree. But each of them need more work. I was thinking that these could be subsequent issues / PR, as to involve more people in the development.
- 5 : Yes this is an undefined behavior. I think as delete is undefined in the list, the event goes back to the parent, until it reaches the comboview, and there it does the tree delete action. I would say it goes with 4.
- 6 and 8 done

- 2 : When you want to catch key press events in the 3D view, you need to use addEventCallback: 

```
        self.callbackKey = self.view.addEventCallback("SoKeyboardEvent", self.KeyboardEvent)
        self.view.removeEventCallback("SoLocation2Event", self.callbackKey)

    def KeyboardEvent(self, info):
        if info["State"] == "UP" and info["Key"] == "ESCAPE":
            # Create a qt key release event for the Escape key
            key_release_event = QtGui.QKeyEvent(
                QtGui.QKeyEvent.KeyPress, QtCore.Qt.Key_Escape, QtCore.Qt.NoModifier
            )
            # Send the key release event to the taskbox
            QtGui.QApplication.sendEvent(self.form, key_release_event)

        if info["State"] == "UP" and info["Key"] == "RETURN":
            if self.accept():
                Gui.Control.closeDialog()
```
The question is how to close the task. I found the 2 options above, first make a QKeyEvent to act as if user pressed esc in the taskbox and send it to the taskbox. Second call reject or accept, and if true then closeDialog.
@chennes could you please advise what is the proper way to dismiss a task when the 3d view has the focus?

Besides I found a weird bug with this. In ways it works ONCE but if you start the tool a second time again and try to do it again (give focus to view and click enter or escape), then it crash.

qewer33:
I also just gave the branch a test and I really like it! The LCS selection is incredibly intuitive. I agree with the points mentioned by howie-j but I'd like to add a few little things:
- I realized that some of the joint icons have improper padding making them appear small on the toolbar, that's my bad. I can fix it in a PR later when this is merged or you can also address it in this PR if you want @PaddleStroke
- It would be great to have alignment options for the part LCSs in the task view (I'm thinking like two groups, "Edit First Part LCS" and "Edit Second Part LCS" with options to rotate, flip, offset etc.). This might be better if it's implemented after this base PR is merged though, but would love to see that
- When will the Ondsel Solver be added so we can test the actual solver joint functionality?
- howie-j mentioned how LCSs would be better hidden, to add on that, it would be great if the LCSs and their attached vertices/features can be highlighted when the joint is selected in the tree view

PaddleStroke:
@qewer33 
- Please go ahead and make a PR, as it's separated there won't be conflicts.
- Alignment options to rotate / reverse the placement is something we need to implement indeed.
- I'm not sure when the API will be ready. Hopefully soon !
- Yes that's needed too

chennes:
> could you please advise what is the proper way to dismiss a task when the 3d view has the focus?

I think that in this case both of your suggestions are reasonable. I normally don't care for generating keypress events, but since yours is really _passing along_ a keypress event it seems like an acceptable path to me. I'd personally probably take the other route, and call accept or reject, but I don't feel strongly about it.

PaddleStroke:
Ok then I'll go with 

``` 
  if self.accept():
      Gui.Control.closeDialog()
```

Are you guys able to reproduce the crash I'm having when closing the tool with Escape for the second time ? If so do you have any logs on your side that could hint at what is wrong? I have no message on my end.

howie-j:
Yes, esc crashes freecad.
Steps to reproduce:
- Switch to assembly
- Create new assembly
- Click create joint (any)
- Click in the 3D view
- Double click escape key, freecad crashes

[FreeCAD.log](https://github.com/FreeCAD/FreeCAD/files/12640586/FreeCAD.log)

chennes:
That looks like it's trying to send an event to a Widget that has been deleted.

jpg87Fr:
Hello everyone.
I would like to know what to do to test the progress of the new Assembly WB.

qewer33:
> Hello everyone. I would like to know what to do to test the progress of the new Assembly WB.

You can clone the branch of this PR and build it but it currently doesn't have the solver integrated, you may want to test it after that's done.

chennes:
> I would like to know what to do to test the progress of the new Assembly WB.

Right now there's not much to see -- are you interested in testing the un-merged PR code, or the code that's already been integrated into the master branch?

PaddleStroke:
I just forced push with : 
- Fixing the crash. For reference, it was a copy paste error in the removeEventCallback of the keyboard event:
```
        self.view.removeEventCallback("SoLocation2Event", self.callbackMove)
        self.view.removeEventCallback("SoLocation2Event", self.callbackKey)
```
instead of 
```
        self.view.removeEventCallback("SoLocation2Event", self.callbackMove)
        self.view.removeEventCallback("SoKeyboardEvent", self.callbackKey)
```
- Improve the tooltips of the joint commands.

PaddleStroke:
A summary of what needs to be done in subsequent PRs:

- [ ] JCS are ugly.
- [ ] JCS should hide when the joint is not in edit.
- [ ] When Joint is selected in the tree, the JCS and selected feature should be highlighted.
- [ ] Joints should be editable. Double clicking should open the Task
- [ ] Delete a joint connector in the taskbox list should remove it from selection.
- [ ] JCS should also snap to line middle and face center of gravity.
- [ ] Add buttons to rotate / reverse JCS

If any of you want to pick up one of these let me know !
@qewer33 maybe you'd be interested to make a better design for the JCS? You can just make a mockup and I'll try to make it in coin. Or if you want to do it in coin directly you are more than welcome 😄

chennes:
OK, this looks ready to me then, I'm going to merge it.

qewer33:
> If any of you want to pick up one of these let me know ! @qewer33 maybe you'd be interested to make a better design for the JCS? You can just make a mockup and I'll try to make it in coin. Or if you want to do it in coin directly you are more than welcome 😄

Personally, I find the current LCS design usable. Just checked OnShape and Fusion360 and they seem to have a very similar graphic. They will definitely look better a little smaller though, can you test that?

qewer33:
Ok I couldn't hold myself and made a mockup anyway @PaddleStroke :sweat_smile: 

It's similar to the current one, just a little shorter and thicker lines with a translucent plane to show the local base XY plane.

image

Here's what OnShape has for reference but I don't think it's that great.

image

jpg87Fr:
_> are you interested in testing the un-merged PR code, or the code that's already been integrated into the master branch?_
I don't know anything about code. I will wait until new elements are present in the Weekly version.

Issue #10625, opened 2023-09-09
[Problem] ladybug in requirements.txt is the wrong package


eye942:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

The ladybug package in [requirements.txt](https://github.com/FreeCAD/FreeCAD/blob/577004c1eba55641e92ea085fa6a078d36f49a06/requirements.txt#L6) is incorrect.

It should either be changed to `ladybug-core` ([github](https://github.com/ladybug-tools/ladybug), [pypi](https://pypi.org/project/ladybug-core/)) or removed. 

The current ladybug package [pypi](https://pypi.org/project/ladybug/) doesn't do anything assumed by the [code](https://github.com/FreeCAD/FreeCAD/blob/577004c1eba55641e92ea085fa6a078d36f49a06/src/Mod/Arch/ArchSite.py#L314)

### Full version info

```shell
N/A - more of a code review.
```

### Subproject(s) affected?

Arch

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
I think this can be closed now?

PR #10626, opened 2023-09-09
Change ladybug to latest ladybug-core

eye942:
Addresses #10625

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ cog==0.6.1
 ConfigParser==5.3.0
 defusedxml==0.7.1
 ifcopenshell==0.7.0.230318
-ladybug==0.0.2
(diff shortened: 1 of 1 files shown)


luzpaz:
@eye942 have you tested this ?

eye942:
No, I haven't.

It looks like those code branches dependent on ladybug normally fail silently, so behavior would probably change.

chennes:
@yorikvanhavre it looks like this is your code -- can you confirm the correct ladybug library?

yorikvanhavre:
Is this requirements file actually used in anything? I can't remember.... Or is it only informative? If yes, I guess we can safely use the latest version. 

In any case, ladybug is only optional in Arch, I guess in dev version there is no problem bumping the version up, we'll fix it if anything goes wrong

PR #10638, opened 2023-09-10
Import: Apply clang format

wwmayer:


wwmayer:
black fails to reformat several files. The same issue happens locally, too.

PR #10643, opened 2023-09-10
Sketcher: refactor CommandConstraint.cpp

FlachyJoe:
Related to  #10524 : get cleaner code in CommandConstraint.cpp by changing 
`geom->getTypeId() == Part::GeomLineSegment::getClassTypeId()` to `geom->isLineSegment()` and so one.

ping @PaddleStroke


FlachyJoe:
split in 2 commits as they touch 2 WB.

PaddleStroke:
This is great!

Just FIY, it's going to conflict with the currently opened https://github.com/FreeCAD/FreeCAD/pull/10531

FlachyJoe:
@PaddleStroke I think we should work in a same branch in order to prepare a common conflict free PR.
Could we ask for a freeze of Sketcher until the refactor was ready?

@abdullahtahiriyo @wwmayer @howie-j @andre-caldas @AjinkyaDahale you have open PRs on Sketcher are you ready to deal with many conflicts to solve or do you want we wait for the merge of your PRs to refactor the code (and so we will handle the whole conflicts)?

andre-caldas:
No need to wait for me.

wwmayer:
Adding methods to a base class to make assumptions about a sub-class isn't good practice (code smell) and makes maintainability more difficult.

You can add a method
```
bool Geometry::is(Base::Type type) const
{
    return getTypeId() == type;
}
```
In client code you will write `geom->is(Part::GeomEllipse::getClassTypeId())`

If you want to avoid to write the `getClassTypeId()` you can add a template function to the Geometry class:
```
template <typename T>
bool is() const
{
    return getTypeId() == T::getClassTypeId();
}
```
In client code you will write `geom->is<Part::GeomEllipse>()`

AjinkyaDahale:
> ... you have open PRs on Sketcher are you ready to deal with many conflicts to solve or do you want we wait for the merge of your PRs to refactor the code (and so we will handle the whole conflicts)?

@FlachyJoe don't wait on my PR. Plenty of conflicts have already appeared since I last touched it.

wwmayer:
> What about some sketcher specific functions in Utils.cpp?

Having a set of these functions in the Sketcher module would be another good alternative.

PaddleStroke:
What about something like
```
bool Sketcher::getGeoCenter(const Part::Geometry* geom){
    if (isCircle(geom)){
         auto* circle = ....
         return Base::Vector2D(center.x, center.y)
    }
    else if (isArc(geom))
          ....

    return Base::Vector2D(0.,0.)
}
```
And similar things for end/start points, radius...

PaddleStroke:
> How do you know whether a passed geometry defines a valid center point? Defining (0, 0) as invalid center won't work because it can very well be a valid center point.

Yes but I think this would not be a real problem. This function would always be called where you know it has a center : 
```
if isCircle(geom) || isEllipse(geom) || isArc(geom) {
      center = getCenter(geom)
      ...
}
```

getCenter would not be called outside a ifCircle (or other type with center) because the developper wouldn't ask for the center of a random geometry.

FlachyJoe:
Is this the good place to discuss?
@PaddleStroke @wwmayer I invited you to collaborate on my fork so you can push on the sketcher_refactor branch.

PaddleStroke:
I don't know if making a big refactor all at once is the best thing because it will make merging difficult. I would suggest making small PRs step by step. Like this one is a first step.

wwmayer:
> This function would always be called where you know it has a center :

And what if not? Code will be changed all the time and it can easily happen that by accident the function is called for an inappropriate geometry and then all the sudden garbage is returned to the calling instance. If you don't want to return a boolean but a point you should throw an exception, instead.

```
if (isCircle(geom) || isEllipse(geom) || isArc(geom) {
      center = getCenter(geom)
      ...
}
```
There are for sure more than one place where you will write such code which violates the DRY rule.

Another problem is when later adding more curve types (e.g. if for some reason the focal point of a parabola is considered as center) you have to adjust all occurrences which is error-prone.

This can be avoided by implementing a function 
```
bool hasCenter(const Geometry* curve)
{
    if (isCircle(geom)) {
        return true;
    }
....

    return false;
}
```

wwmayer:
I don't know how much code you want to refactor (since in the Sketcher module there is a lot of code duplication) but I recommend small and focused PRs.

So, for the simplification to avoid to compare the types I suggest to make one commit that adds the new functions. In a second commit use the new functions where needed. This makes the review process  a lot easier.

After the PR is merged start with the next PR.

FlachyJoe:
Soooo, I've brake my branch and I don't know how to get it clean. I close this and I'll make a new PR.

PR #10644, opened 2023-09-10
Import: move several .py files that black fails to re-format to a new…

wwmayer:
… directory and exclude it


wwmayer:
> should we add src/Mod/Import/App/SCL_output to codespell ignore list ?

Yes, we can. As it's all generated code it doesn't make sense to run them through the spell checker.

Issue #10616, opened 2023-09-09
[Problem] Path: Sanity Report is not converting from asciidoc to html


VectorHasting:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

**Error/problem:** 

On Windows systems, the Path_Sanity option (used as a button, at least) does not display the Job Report, instead it displays:

`[WinError 2] The system cannot find the file specified`

**Background**:

I have installed asciidoctor as per [instructions found on their website](https://docs.asciidoctor.org/asciidoctor/latest/install/)

**To reproduce the problem**:

1. Create a Job in Path
2. Specify an output file under Job > Output > Output File
3. Attempt to use the Sanity Check Button. 

Expected Behavior:
Which is [documented on the FreeCAD wiki, at the bottom](https://wiki.freecad.org/Path_Sanity ) 
and [demonstrated in @sliptonic video here](https://www.youtube.com/watch?v=RFu-DQ04uZM&t=903s )

It should:

- generate a Job Report in .asciidoc format at the location specified in Output file, 
- convert that file to .html format, 
- display it in the system's default browser. 

Observed Behavior:

It seems to actually:

- generates a Job Report in .asciidoc format, 
- fail to convert that file to .html format, 
- display the error shown above in FreeCAD's Report View, 
- and launch (on Windows) Microsoft Edge browser, which either displays the home page (if the output file is on a local drive) or an error message (if the file is on a unmapped drive/url server, ie \\fileserver\etc)

**Workaround**:
After installing asciidoctor, manually convert the file using [their instructions here.](https://docs.asciidoctor.org/asciidoctor/latest/get-started/)

Note: the exact command I use is simply:

`asciidoctor  c:\filepath\filename.asciidoc`

It works whether the command is issued from a different directory or different drive. 

Error page screen follows:

![Screenshot 2023-09-08 145244](https://github.com/FreeCAD/FreeCAD/assets/51137712/72682bfa-a9be-4b00-bb1f-01991245dd09)

Thank you!

### Full version info

```shell
[code]
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.21.0.33668 +7 (Git)
Build type: Release
Branch: (HEAD detached at 0.21)
Hash: 41b058e2087de60dc8fef3d3e68c7d0129e13abf
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
[...]

PR #10654, opened 2023-09-11
[Path] Improve Sanity to generate html output without...

Syres916:
...the need for dependences.

Fixes https://github.com/FreeCAD/FreeCAD/issues/10616

There will be a follow up PR to do work on the internal style so it can changed easily and possibly have some Preferences set in the future but the main thing is to get some feedback from  the weekly Path user base.


Syres916:
Therefore if the tests pass this is ready for review and merge.

sliptonic:
Very nice work!  I'm thrilled to see the extra dependency eliminated.  Merging.

It would be nice to further separate the html and css from the other code but that can be another unit of work.  
thank you!

Issue #10603, opened 2023-09-08
[Problem] Sketcher_dimension : Angle between 2 lines should switch between supplementary


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Sketcher_dimension : Angle between 2 lines should switch between supplementary angles depending on the mouse position. Currently it does not.
Discussion on the forum : https://forum.freecad.org/viewtopic.php?t=80994

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10657, opened 2023-09-12
Sketcher_Dimension: Enable the user to select what angle he wants.

PaddleStroke:
Fixes https://github.com/FreeCAD/FreeCAD/issues/10603
In Sketcher_Dimension and also in general moving the constraint : Enable the user to select the complementary angle he wants based on the position of the mouse. Similarly as how he can choose distance/DistanceX/Y.
Forum discussions : 
https://forum.freecad.org/viewtopic.php?t=28321
https://forum.freecad.org/viewtopic.php?t=80994

### DIFF
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -1580,13 +1580,13 @@ void ViewProviderSketch::moveConstraint(int constNum, const Base::Vector2d& toPo
     if (!isInEditMode())
         return;
 
-    const std::vector<Sketcher::Constraint*>& constrlist =
-        getSketchObject()->Constraints.getValues();
+    Sketcher::SketchObject* obj = getSketchObject();
+    const std::vector<Sketcher::Constraint*>& constrlist = obj->Constraints.getValues();
     Constraint* Constr = constrlist[constNum];
 
 #ifdef FC_DEBUG
-    int intGeoCount = getSketchObject()->getHighestCurveIndex() + 1;
-    int extGeoCount = getSketchObject()->getExternalGeometryCount();
+    int intGeoCount = obj->getHighestCurveIndex() + 1;
+    int extGeoCount = obj->getExternalGeometryCount();
 #endif
 
     // with memory allocation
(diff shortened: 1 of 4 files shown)


PaddleStroke:
@chennes this one is pretty cool :)

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -1252,6 +1254,9 @@ class DrawSketchHandlerDimension : public DrawSketchHandler
         if (specialConstraint == SpecialConstraint::LineOr2PointsDistance)
             updateDistanceType(onSketchPos);
 
+        if (specialConstraint == SpecialConstraint::LinesAngle)
+            updateAngle(onSketchPos);
+
```
chennes, line 1259: Looks like there's something wrong with the pre-commit CI (or its application of clang-format) -- I thought we set it up so that curly braces were required even on one-liners.

chennes:
When I flip the angle to the "wrong" side, the constraint markers are mis-rendered:
Screenshot 2023-09-14 at 1 31 46 PM

howie-j:
Hmm, i'm getting quite unexpected results:
[pr_10657.webm](https://github.com/FreeCAD/FreeCAD/assets/82042580/fcdd1714-3128-440d-b499-79b7f334ece4)

`
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34260 (Git)
Build type: Release
Branch: pr_10657
Hash: 4b696429101bb7e687345c9e09aac95e38f78526
Python 3.11.5, Qt 5.15.10, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
`

howie-j:
Tried on X11 instead of wayland with a clean/default setup.
It works as expected with crossing lines, but not with coincident or separate lines:

[pr_10657_x11.webm](https://github.com/FreeCAD/FreeCAD/assets/82042580/a99c9f94-90be-4f59-a158-a2846e549636)

chennes:
> @chennes what did you meant by misrendered? Is it that the value is writen on top of the arc. Or is it that you expected the complementary angle (ie 360 - angle)? In this PR the goal was to make the supplementary angle (180-angle).

I think that without rendering the connecting lines, showing the supplementary angle doesn't make any sense -- in the animation shown in [this post](https://forum.freecad.org/viewtopic.php?p=231038&sid=acf322a7c835cb7e9d94d8c557ee05bc#p231038) there are lines indicating what angle is being displayed. In this case, the angle is just floating out in space, with no clear reference.

qewer33:
It might also make sense to implement this for the normal Angle tool as well when it's stable. For the users who prefer the separate toolset

PaddleStroke:
You are right @chennes. I was planning to try to improve the connecting lines for angles. Though I don't know if this belong to this PR as it's quite a separate part of the code.

PaddleStroke:
@qewer33 the normal angle tool does not offer a moment to do that. As it's a one shot command. No mouse move time to do it.
However what we can do is add to the datum edit dialog a 'Reverse angle' checkbox that would swap the angle.

PaddleStroke:
@chennes I fixed the problem that @qewer33 mentioned and I added the brackets to the ifs.
I will start looking into how to improve the arc SoDatumLabel. But I think it's best to separate the PRs.

howie-j:
Very nice, almost all bugs fixed in the last commit.

The only bug i could find now is when creating an angle between a line and X or Y axis:
When the mouse goes "negative" compared to the axis, the dimension snaps 180' opposite until the cursor passes 90' to the line:

[dim_angle_bug.webm](https://github.com/FreeCAD/FreeCAD/assets/82042580/8e7e9cb8-f10b-489a-9ba3-8d2aec650801)

(the blue construction line normal to the regular line is just for illustration)

chennes:
>  I will start looking into how to improve the arc SoDatumLabel. But I think it's best to separate the PRs.

Fair enough -- once @howie-j is satisfied with the PR I'll merge it.

PaddleStroke:
After a long fruitless investigation, I finally found that this bug is not due to my PR. It is in fact a bug of moving the constraint.
First it does not need to be an axis. This happen with any lines. Though in some cases, such as the axis, it's more visible. The axis is equivalent to a 1mm long line actually. If you try the same as you did with a random 1 mm horizontal line you'll have the same result.

But this is not due to sketcher_dimension : Make the angle and validate it first. Then try to click and drag the constraint as you did, the described behavior is achieved.

So I'm not sure what to do here. Do I have to fix it in this PR ?

The angle constraint 'drag and move' could benefit from a rework as we could work to enable the angle flipping during the mouse move rather than in sketcher dimension... Which would make this PR useless.

I will check tomorrow if I can move this code out of sketcher_dimension and into the moving of angle constraint

chennes:
> Do I have to fix it in this PR ?

No, certainly not -- if the fix were obvious and simple it would be fine for you to do so, but if it's not a regression due to this PR, there's no requirement that you fix it. You might consider submitting an Issue about it, though. Do you want to merge this as-is then?

howie-j:
For me it's more than fine to merge as-is, as long as we create an issue we can point to regarding the constraint flipping.
The PR provides a very nice improvement to creating angles that has annoyed me and probably many other users for years. Merci beaucoup Pierre!

PaddleStroke:
Ok so the good news is that it was fairly easy to move to 'moveConstraint' so that it now works not only on angle creation, but also when you move an angle around.
The bad news is that there is now an issue when the constraint has an expression. It should be easy and just adding/removing 180- from the expression. The issue is how to access the expression.
Cf my pseudo code here : 

```
                    // Edit the expression if any
                    /*
                    if (Constr->hasExpression()) {
                        std::string expression = Constr->getExpression();
                        if (expression.substr(0, 3) == "180-") {
                            expression = expression.substr(3);
                        }
                        else {
                            expression = "180-" + expression;
                        }
                        Constr->setExpression(expression);
                    }*/
```

@wwmayer do you perhaps know if there is an easy way to access the constraint expression? I tried to follow the logic in EditDatumDialog but I'm having trouble as it seems it use an expression binding to the widget there.

wwmayer:
> do you perhaps know if there is an easy way to access the constraint expression?

@abdullahtahiriyo is there no method to get the index (int) for a given constraint?

Assuming you know the index of the constraint then you can access a possible associated expression via the sketch object:
```
int index = ...
SketchObject* sketch = ...
App::ObjectIdentifier path = sketch->Constraints.createPath(index);

auto info = sketch->getExpression(path);
if (info.expression) {
    std::string expression = info.expression->toString();
    expression = ...;
    try {
        std::shared_ptr<App::Expression> expr(App::Expression::parse(sketch, text));
        // there is a bug in the SketchObject API because setExpression() is protected but public in DocumentObject
        App::DocumentObject* base = sketch;
        base->setExpression(path, expr);
    }
    catch (const Base::Exception&) {
    }
}
```

I don't know if there is a better way than this but it works.

> expression = "180-" + expression;

Manipulating the expression this way is fragile because you don't know how the expression looks like. Better do
`expression = "180 - (" + expression + ")";`
The parentheses are important because the old expression could be "75 - 15". Without them you would get `180 - 75 - 15` = 90 instead of `180 - (75 - 15)` = 120

Another thing to check is whether the old expression uses the degree symbol to avoid a unit mismatch when changing the expression.

PaddleStroke:
`   App::ObjectIdentifier path = getSketchObject()->Constraints.createPath(constNum);`
Is giving me a 'incomplete class is not allowed' error. What includes I need for it to work? I'm in ViewProviderSketch.cpp

wwmayer:
Some general remarks:
* You undid the auto-formatting done by the pre-commit hook. If you want to make sure that the changes persist you should put `// clang-format off` and `// clang-format on` around the code. But make sure that the lines won't be much longer than 100 characters.
* ViewProviderSketch::moveConstraint() is one of these huge functions with more than 250 lines of code. Take the opportunity and at least refactor the else-branch you are working at. This bad practice to bloat functions more and more must be stopped.

wwmayer:
```
#include <App/ObjectIdentifier.h>
#include <App/Expression.h>
```

wwmayer:
Sketcher::Constraint looks like the wrong place to me. Better put it into SketchObject or where it makes sense to DocumentObject.

sliptonic:
Moving to Draft.  Please reset when ready.

PaddleStroke:
Ok so I have split the move constraint function. And moved some code into SketchObject.
Is that all good now?

Side note: The problem raised by @howie-j is solved.

howie-j:
Tested, could not detect any misbehavior. Very nice work :1st_place_medal:

howie-j:
_Damn_ this is cool

[movable.webm](https://github.com/FreeCAD/FreeCAD/assets/82042580/f8279a63-883c-48a8-aab6-620f8a076ce9)

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1750,91 +1750,122 @@ void ViewProviderSketch::moveConstraint(int constNum, const Base::Vector2d& toPo
...
+            const auto* lineSeg1 = static_cast<const Part::GeomLineSegment*>(geo1);
+            const auto* lineSeg2 = static_cast<const Part::GeomLineSegment*>(geo2);
+
+            Base::Vector2d l1[2], l2[2];
+            l1[0] = Base::Vector2d(lineSeg1->getStartPoint().x, lineSeg1->getStartPoint().y);
+            l1[1] = Base::Vector2d(lineSeg1->getEndPoint().x, lineSeg1->getEndPoint().y);
+            l2[0] = Base::Vector2d(lineSeg2->getStartPoint().x, lineSeg2->getStartPoint().y);
+            l2[1] = Base::Vector2d(lineSeg2->getEndPoint().x, lineSeg2->getEndPoint().y);
+
+            // First we will check if the angle needs to be reversed to its suplementary
```
chennes, line 1792:
```suggestion
            // First we will check if the angle needs to be reversed to its supplementary
```

PaddleStroke:
Typo fixed + typeId test replaced by isLineSegment(*geo1) (line 1779, 1850, 1854).
And rebased.

Syres916:
@PaddleStroke shouldn't errors presented to the user be made available for translation?

PaddleStroke:
This error is not a user-facing error. It's not supposed to ever appear so I think it's not worth translating.

PR #10658, opened 2023-09-12
clang format SpacesBeforeTrailingComments

WandererFan:
This PR is a change to clang-format for SpacesBeforeTrailingComments as discussed here: https://forum.freecad.org/viewtopic.php?p=703970#p703970


WandererFan:
pre-commit.ci autofix

PR #10665, opened 2023-09-12
Part: move scale command

wwmayer:
See https://github.com/FreeCAD/FreeCAD/pull/10583#issuecomment-1716187490

### DIFF
--- a/src/Mod/Sketcher/Gui/Workbench.cpp
+++ b/src/Mod/Sketcher/Gui/Workbench.cpp
@@ -494,7 +494,7 @@ template<>
 inline void SketcherAddWorkbenchTools<Gui::ToolBarItem>(Gui::ToolBarItem& consaccel)
 {
     consaccel  //<< "Sketcher_SelectElementsWithDoFs" //rarely used, it is usually accessed by
-               //solver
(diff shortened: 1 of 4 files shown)


src/Mod/Sketcher/Gui/Workbench.cpp
```diff
@@ -494,7 +494,7 @@ template<>
 inline void SketcherAddWorkbenchTools<Gui::ToolBarItem>(Gui::ToolBarItem& consaccel)
 {
     consaccel  //<< "Sketcher_SelectElementsWithDoFs" //rarely used, it is usually accessed by
-               //solver
+               // solver
                // message.
```
luzpaz, line 498: Should line 498 be consolidated on to 497 ?

wwmayer: This was done by the pre-commit hook to avoid to exceed the 100 columns. But I don't know why three lines of comments are created instead of only 2.

luzpaz: weird, then the pre-commit hook is malfunctioning in some way.

PR #10666, opened 2023-09-12
Sketcher: add TypeId checking helpers

FlachyJoe:
in order to replace
```cpp
geo->getTypeId() == Part::GeomEllipse::getClassTypeId()
```

with
```
Sketcher::isEllipse(*geo)
```

### DIFF
--- a/src/Mod/Sketcher/Gui/Utils.cpp
+++ b/src/Mod/Sketcher/Gui/Utils.cpp
@@ -48,6 +48,51 @@ using namespace std;
 using namespace SketcherGui;
 using namespace Sketcher;
 
+bool Sketcher::isCircle(const Part::Geometry& geom)
+{
+    return geom.getTypeId() == Part::GeomCircle::getClassTypeId();
+}
+
+bool Sketcher::isArcOfCircle(const Part::Geometry& geom)
+{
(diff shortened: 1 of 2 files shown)


0penBrain:
Would be better to use reference rather than pointer as argument. Methods aren't supposed to handle null pointer (and actually fail to do). 😉

FlachyJoe:
>  use reference rather than pointer as argument.

Notice that the replaced lines don't check for null pointer but I'll follow your advise so the usage could be extend.

Some other functions in Utils.cpp should be changed too, for example:
```cpp
bool SketcherGui::isBsplinePole(const Sketcher::SketchObject* Obj, int GeoId)
{

    auto geom = Obj->getGeometry(GeoId);

    return isBsplinePole(geom);
}
```

0penBrain:
Maybe I was unclear. You can keep the references as const. 😉

FlachyJoe:
Fixed + Rebased + Squashed

chennes:
@wwmayer what are your thoughts on this? It's not quite the solution you proposed, but at least the methods are base class members now.

wwmayer:
> what are your thoughts on this?

As a first step this is totally fine. 

> It's not quite the solution you proposed

I have offered some more alternatives but this one is fine, too.

PR #10667, opened 2023-09-12
App/Toponaming: Add StringHasher to Document

chennes:
Before we can really begin [#10353](https://github.com/FreeCAD/FreeCAD/issues/10353), Document needs to be able to provide a `StringHasher`. This PR brings in the bare minimum code required to load and save the StringHasher in a Document. It does _not_ fully implement storing the naming information, nor does it implement the "forceXML" mechanism for the hasher.

The first commit is as close to @realthunder's original code as could be made to compile in current FreeCAD without making any significant changes to the code structure (in particular I did not implement the `pimpl` code in `PropertyContainer`).

Planned follow-up commits include some basic C++ modernization cleanup, and basic tests.

### DIFF
--- a/src/App/Document.cpp
+++ b/src/App/Document.cpp
@@ -64,6 +64,7 @@ recompute path. Also, it enables more complicated dependencies beyond trees.
 #endif
 
 #include <boost/algorithm/string.hpp>
+#include <boost/bimap.hpp>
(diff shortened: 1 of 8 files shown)


JohnAD (approved):
Looks good! Presumably, there will never be a good reason (or safe way) to remove a hasher :smile: .

PR #10674, opened 2023-09-13
Gui: determine the bounding box of an SoDatumLabel outside its GLRend…

wwmayer:
…er() method

See forum: https://forum.freecad.org/viewtopic.php?t=73443&start=40

### DIFF
--- a/src/Gui/SoDatumLabel.cpp
+++ b/src/Gui/SoDatumLabel.cpp
@@ -140,13 +140,311 @@ void SoDatumLabel::drawImage()
     Gui::BitmapFactory().convert(image, this->image);
 }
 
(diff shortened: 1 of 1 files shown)


wwmayer:
I don't know of a latency issue. This PR fixes an issue where a datum label couldn't be selected when its computeBBox() is called before its GLRender() method because in the latter the bounding box was computed.

PaddleStroke:
Ah it's a coincidence then. I was looking for a way to calculate the textOffset outside of the rendering. So that textOffset position could be calculated by the EditableDatumLabel.
Because currently EditableDatumLabel is using the textOffset calculated from the previous render. For task image it's not a problem as the spinbox appears only after the datum is fixed, so its not moving. But in my recent work on sketcher tool settings, the label is moving, and so it is noticeable, especially if the mouse goes fast.

PR #10676, opened 2023-09-13
Gui: Remove bounding box computation from SoDatumLabel::GLRender()

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/10674 is merged that is supposed to handle a selection issue of constraints.

This PR now cleans up some  code that shouldn't be needed any more but it's best to wait until we see that the merged PR 100% does its job.


wwmayer:
It shows that this PR also fixes a selection issue when zooming out. For more details see: https://forum.freecad.org/viewtopic.php?p=706500#p706500

abdullahtahiriyo:
Just to note that I have checked and endorse it.

I am not merging it because of the comment to best wait until the other (merged) PR settles.

PR #10681, opened 2023-09-14
Gui: make layout of Combo, Property and Tree view more flexible

wwmayer:
Rework the current behaviour of the ComboView to make it more flexible. It moves the Task view out of the ComboView and makes it a stand-alone dock widget. 

See https://forum.freecad.org/viewtopic.php?p=626732#p626732 and https://forum.freecad.org/viewtopic.php?t=79950

### DIFF
--- a/src/Gui/ComboView.cpp
+++ b/src/Gui/ComboView.cpp
@@ -23,15 +23,13 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <QEvent>
+# include <QGridLayout>
 # include <QSplitter>
 #endif
 
 #include "ComboView.h"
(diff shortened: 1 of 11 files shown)


PaddleStroke:
Ah great I was just about to ask you if you planed to pick this up !

obelisk79:
I look forward to seeing this merge. Thanks for the great work Werner!

M4xTheMan:
I've missed the discussion regarding this UI change. I'm uncertain how to dock the panels. Is the standard behavior to place the selection view, model and tasks on the left side? From my point of view, this results in to much vertical scrolling and a missing overview. That's why I'm asking for "best practice" examples.

PR #10684, opened 2023-09-14
[planegcs] Removed unused code.

mosfet80:
removed redefinition of pi
removed unused code
speed pi/18 calculations


luzpaz:
@mosfet80 since `removed unused code.` is pretty generic,  can you please amend the commit title to:

```bash
git commit --amend -m "[planegcs] Removed unused code" -m "removed redefinition of pi" -m "removed unused code"
-m "speed pi/18 calculations"
git push --force
```

mosfet80:
@chennes I mistakenly performed a rebase on the wrong branch. Now are ok.

src/Mod/Sketcher/App/planegcs/Geo.h
```diff
@@ -50,6 +48,9 @@ class Point
 };
 
 using VEC_P = std::vector<Point>;
+static const float pi = boost::math::constants::pi<float>();
+static const float pi_2 = boost::math::constants::pi<float>() / 2.0f;
+static const float pi_18 = boost::math::constants::pi<float>() / 18.0f;
```
chennes, line 53: Can these be `constexpr`?

mosfet80: The "Boost" mathematical constraints are already used: :https://github.com/FreeCAD/FreeCAD/blob/0bbb10864b11015e040860b549e93b1b708ab9b6/src/Gui/NaviCube.cpp#L633 .  
And ..in preparation for  #10650
M_PI  definition have a problem with library libe57 v3.1

chennes: It's not a big deal, it just seems to me that if the underlying boost call is constexpr, we can make our constants a bit more flexible by also marking them as constexpr.

xtemp09:
`pi`, `pi_2`, `pi_18` are _float_, while everywhere _double_ is used. Is there a reason behind that? Would it not be better to declare them as _double_ to avoid conversion and possible confusion?

Would it not be better to use `std::numbers` instead of boost? `std::numbers` are supported since gcc 8.

mosfet80:
@xtemp09  pi are not defined in into std::numbers c++17 or lower https://en.cppreference.com/w/cpp/numeric/constants .
pi, pi_2 ad pi18 are a float 32bit , previus definition of M_PI are a 21 bit of approximation.

xtemp09:
@mosfet80, I don't understand. You wrote:  

```c++
static const float pi = boost::math::constants::pi<float>();
static const float pi_2 = boost::math::constants::pi<float>() / 2.0f;
static const float pi_18 = boost::math::constants::pi<float>() / 18.0f;
```

Is there a logic behind usage of _float_? Everywhere in the code _double_ is used, for example:

```c++
double incrAngle = *(a.startAngle) < *(a.endAngle) ? pi_2 : -pi_2;
```

I don't understand what you wrote about usage of C++17. In my understanding, FreeCAD should support Ubuntu 22.04 and 20.04. The latter [has gcc 9.3.0 by default](https://distrowatch.com/table.php?distribution=ubuntu). Of course, anyone can install the latest compiler from repositories or backports, but gcc 9.3 is provided out-of-the-box. `std::numbers` are present in gcc 8 and newer.

I concluded that usage of C++ linguistic features of gcc 9.3 and older is possible, and, maybe, even encouraged, since it is the standard, while boost is an additional library.

src/Mod/Sketcher/App/planegcs/Geo.h
chennes: Can you make these doubles, since they are used in double-precision arithmetic later on?

wwmayer:
> std::numbers are present in gcc 8 and newer.

std::numbers have been added with C++20 but C++17 is still the default version to compile FreeCAD. On my Debian 12 I have recently tested to build FreeCAD with C++20 and it caused a build failure -- probably a bug in the gcc version. The same problem can be found here: https://forum.freecad.org/viewtopic.php?t=80375

chennes:
I've switched those constants to `constexpr` and `double` -- in all cases I found they were being used in double-precision arithmetic, and using `constexpr` makes them a little more flexible.

xtemp09:
I compiled FreeCAD with C++20 on Ubuntu 22.04 two weeks ago. If FreeCAD can be compiled with Visual C++, is it time to raise the bar? If so, some things from _boost_ and _libfmt_ become unnecessary.

---

Does FreeCAD has guideline for declaration of variables? I would have written terser:

```c++
static constexpr double pi = std::numbers::pi,
                        pi_2 = pi / 2.0,
                        pi_18 = pi / 18.0;
```

chennes:
We still support Ubuntu 20.04 LTS (since it is still in main support phase) -- my understanding is that the version of GCC there does not support C++20.

Regarding multiple variables in a single declaration: it's frowned upon...
https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html

xtemp09:
> Global variables and member variables are excluded.

Repetition of the same fragment can cause bugs since it hampers readability.
boost::math::constants::pi<double>() is repeated thrice, while in the second and the third is redundant
If you think each variable must be in new line, maybe this is better:

src/Mod/Sketcher/App/planegcs/Geo.h
```diff
@@ -50,6 +48,9 @@ class Point
 };
 
 using VEC_P = std::vector<Point>;
+static constexpr double pi = boost::math::constants::pi<double>();
+static constexpr double pi_2 = boost::math::constants::pi<double>() / 2.0;
+static constexpr double pi_18 = boost::math::constants::pi<double>() / 18.0;
```
chennes, line 53:
```suggestion
static constexpr double pi = boost::math::constants::pi<double>();
static constexpr double pi_2 = pi / 2.0;
static constexpr double pi_18 = pi / 18.0;
```

PR #10688, opened 2023-09-14
[Feature Request] [Core] [Preferences] [UI/UX] Buttons to reset default values of Pages/Groups

### LINKED ISSUE #10161: [Feature Request] Reset tab button in Preferences (only reset preferences in the active tab)

### Full version info

```shell
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Build type: Release
Branch: (HEAD detached at 0.21)
Locale: Polish/Poland (pl_PL)
```

### Subproject(s) affected?

Core

### Problem description

Currently, there's only a Reset button resetting all the preferences to their default values. But it would be very useful to have an additional "Reset tab" button that would reset the values of preferences only in an active tab. This would be particularly helpful e.g. for tessellation settings.

Forum discussion: https://forum.freecad.org/viewtopic.php?t=80253

### Anything else?

_No response_

### Code of Conduct

CalligaroV:
Fixes #10161

**Forum discussion**
https://forum.freecad.org/viewtopic.php?t=80253

Allows to reset the default values of a Preferences Page, as for the title of the issue, and also adds another button to reset all the Preference Pages of the chosen Group.


luzpaz:
Can you copy image into the OP to show what the final outcome of the PR looks like in the GUI ?

CalligaroV:
Not sure if you asked me to put it on my first post here or to put it in the forum topic but in the meantime I paste it here :wink:

Don't know if this will be the final version but it takes into account the tips from other users on the forum

preferencesDefaultButtons

src/Gui/DlgPreferencesImp.cpp
```diff
@@ -204,11 +220,38 @@ void DlgPreferencesImp::createPageInGroup(QTabWidget *tabWidget, const std::stri
...
+    int groupIndex = ui->listBox->currentRow();
+    QFontMetrics fm(font());
+    QString group = fm.elidedText(ui->listBox->item(groupIndex)->text(), Qt::ElideRight, ui->buttonResetGroup->width()-4);
+    
+    QTabWidget* tabWidget = static_cast<QTabWidget*>(ui->tabWidgetStack->currentWidget());
+    int tabIndex = tabWidget->currentIndex();
+
+
+    ui->buttonResetGroup->setText(tr(std::string("Reset Group\n" + group.toStdString()).c_str()));
+    ui->buttonResetTab->setText(tr(std::string("Reset Tab " + tabWidget->tabText(tabIndex).toStdString()).c_str()));
```
chennes, line 234: What you need here is to use the String `arg` function to construct your string, so that translators see the right thing. It will looks like `tr("Reset Group %1").arg(groupName)`. I wouldn't include the "\n" either, that will only work in English. So tell the button to wrap the text (you'll have to use a QToolButton), or something along those lines.

FEA-eng:
Looks great to me :-) It's a very nice enhancement.

CalligaroV:
@FEA-eng Thank you very much for the appreciation! :smile: 

@chennes Thanks for the suggestions!
About the `arg` usage I have no other questions ATM, most likely I will apply it in my next commit.

About the QToolButton, I'm not sure I understood your suggestion.
Could you please share an example from other parts of FreeCAD or from other software?

chennes:
> About the QToolButton, I'm not sure I understood your suggestion.

You have included a `\n` in the text of that button to manually wrap the line: that isn't going to work well in translated text. So I am proposing that you switch from using a `QPushButton` (which does not support text wrapping, as far as I remember) to a `QToolButton`, which does.

CalligaroV:
I see, so it's just to get rid of `\n`, not for particular features of the `QToolButton`.

However, how is the text wrapping enabled?
I did some searches and found that one method used is suggested here [https://stackoverflow.com/questions/8960233/subclassing-qlabel-to-show-native-mouse-hover-button-indicator/8960548#8960548](url)
Do you know easier ways? I'd prefer to remove the code I had to add to obtain the text wrapping

Besides that, I applied your other suggestion about the `arg` function, hope it's correct now.

I also noticed that, after changing the language, the entries in the groups list lose the vertical alignment so I added some lines of code to fix that too

yorikvanhavre:
You can apparently set a label as the widget of a pushbutton: https://stackoverflow.com/questions/8969086/is-there-a-way-to-enable-word-wrapping-of-text-on-some-simple-widgets-like-qpush

and then enable word wrap...

FEA-eng:
@CalligaroV Is this PR ready to merge?

CalligaroV:
@FEA-eng by now I can see only 2 reasons for not merging:

1. Haven't yet figured out how to handle the settings added in "non standard ways" (see my Side Note in the blog post https://forum.freecad.org/viewtopic.php?p=703098#p703098)
2. Don't know if someone else tested it

If this two items aren't an issue then for me it's ok to merge!

chennes:
I had planned on merging it at this morning merge meeting, but then I couldn't make the meeting... sorry for the delay.

wwmayer:
> * Morphed buttons from QPushButton to QToolButton

Why? The height of a tool button is different to the height of a push button and this makes the dialog looking bad. Also the behaviour when hovering or pressing a tool button is very different and this makes it more inconsistent.

Having three Reset buttons now is very confusing, IMO.  The buttons are somehow related but are located at very different places which makes it even more confusing.

I think a better and more polished way is to switch back to a single reset button which when clicking on it opens a sub-widget with the 3 options.

CalligaroV:
@chennes, thanks for merging!

@wwmayer, thanks for the feedback!

> Why? The height of a tool button is different to the height of a push button and this makes the dialog looking bad. Also the behaviour when hovering or pressing a tool button is very different and this makes it more inconsistent.

I saw some differences in the height of those buttons but in my case those differences were small enough to not make it an issue for me.
If you think it's better to fix it I'll apply the modifications needed.

About the hovering behavior, have you experienced something strange because of that or are you foreseeing potential issues?
I tried again right now but haven't spotted anything
> Having three Reset buttons now is very confusing, IMO. The buttons are somehow related but are located at very different places which makes it even more confusing.

I can understand your argument and I somehow agree.
The current implementation takes into account the requests/suggestions from other forum users.
If you think this point is worth of further discussion I can write a post about that on the forum!

> I think a better and more polished way is to switch back to a single reset button which when clicking on it opens a sub-widget with the 3 options.

I like this idea!
This may give yet another reason for using QToolButton.

At this point, if you agree, I'll expose these last two point on the forum.

chennes:
FYI there is a UI/UX working group forming right now (details to come once the organizers have a preliminary meeting) and I know that they are planning on working on the Preferences dialog. So you might want to reach out to @Obelisk79, who is organizing that effort.

obelisk79:
Use of QPushButton is more logical unless the QToolButton with dropdown is implemented. I would like to see the preferences dialog get a complete review and reorganization prior to the next full release, which is one of the reasons why I'll be eventually leading a discussion with the working group. But that will be a little bit further down the road.

wwmayer:
> If you think it's better to fix it I'll apply the modifications needed.

It's not only the height but its visual appearance. If you use the default desktop theme it's not that obvious but as soon as you activate one of the style sheets e.g. to Dark then the push buttons are flat and look like a label while the push buttons are raised. With other style sheets like "Behave-dark" it looks even worse because a tool button has a linear gradient while a push button hasn't it.

> About the hovering behavior, have you experienced something strange because of that or are you foreseeing potential issues?

No issues but visual appearance only. But IMO this change contradicts to the efforts (e.g. from paddle) to make the UI more user friendly and consistent.

> I like this idea!
This may give yet another reason for using QToolButton.

I don't mind what buttons you use for this sub-panel.

But IMO three reset buttons are way too prominent and this isn't something a user uses all the time.

> At this point, if you agree, I'll expose these last two point on the forum.

OK.

Btw, I have tested this function and doesn't work in 100% of all cases because not all setting pages use "PrefWidgets" but get and set the parameters by accessing the parameter groups via the Application. Some example where it fails are:
* DlgSettingsGeneral
* DlgSettingsWorkbench
* DlgSettingsTheme
* DlgSettingsEditor

The class DlgPreferencesImp isn't able to handle all eventualities and thus I think there should be a way to let a concrete settings class override the default behaviour. I can think of two alternatives to achieve this:
* Move restorePageDefaults to PreferencePage and make it virtual. Then a subclass can override the method and handle the tricky cases (or disallow to reset any parameters in case this makes sense anywhere)
* The DlgPreferencesImp class could check if the settings class has a slot function of the name restoreDefaults and if yes call it, otherwise use restorePageDefaults

The first option is more OOP and thus probably the better choice.

CalligaroV:
@chennes thanks for the info!

@obelisk79 I'm sorry I couldn't attend the meeting. Is the next one already scheduled?
About the dialog reorganization, you mean Kadet's the proposal of forum topic [https://forum.freecad.org/viewtopic.php?t=81609](url)?

@wwmayer, I just saw you already opened a pull request to restore the single reset button layout. Thanks!
About moving `restorePageDefaults`, I think I got your idea but I'd like to confirm:
by doing so the pages with all the parameters defined using `Pref*` Widgets will use that function but pages such as the ones you mentioned must be modified redefining `restorePageDefaults`, correct?

p.s. for everyone: sorry for the late reply!

wwmayer:
> correct?

Correct.

PR #10692, opened 2023-09-15
Sketcher: refactor CommandConstraint.cpp

FlachyJoe:
- use TypeId helpers
- use auto for variable declaration which use static cast
- change empty lines


PaddleStroke:
Great to see that his went through

PR #10693, opened 2023-09-15
Draft: gui_edit called displayPoint without mask

Roy-043:
Forum topic:
https://forum.freecad.org/viewtopic.php?t=81186


Roy-043:
Self-merging as this is a noncontroversial bug fix.

PR #10695, opened 2023-09-15
[TD]allow cosmetic deletion via DEL

WandererFan:
This PR implements a solution to a long standing problem where users press the delete key instead of using the CosmeticEraser command.  This deletes the entire view instead of the selected cosmetic element.


luzpaz:
backport ?

WandererFan:
IDK.  It won't merge into v0.21 as is.  It would have to be a cut and paste job.

PR #10699, opened 2023-09-15
Draft: Fix for PR10673

Roy-043:
Fixes an issue that was missed in #10673.


Roy-043:
Self-merging as this is a noncontroversial bug fix.

PR #10700, opened 2023-09-16
Fix substring operation into SetupShibokenAndPyside.cmake

mosfet80:
Fixed macro: The substring "Location:" contains 10 characters


mosfet80:
patch are into https://github.com/berniev/FreeCAD/tree/pip_files

PR #10703, opened 2023-09-16
Gui: Improve accuracy for navigation cube's rotate to nearest feature

Rexbas:
Makes the navigation cube's rotate to nearest feature more accurate. Previously the current camera orientation was used in the final result from the calculations. This led slightly different values every time a face was selected and resulted in some near zero values. Now, the fixed standard orientations are used in the final calculation. This makes sure the results are consistent and prevent some, but not all, near zero values. This PR is related to https://forum.freecad.org/viewtopic.php?t=81043

### DIFF
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -948,6 +948,12 @@ SbRotation NaviCubeImplementation::getNearestOrientation(PickId pickId) {
     SbVec3f standardZ;
     standardOrientation.multVec(SbVec3f(0, 0, 1), standardZ);
 
+    // Cleanup near zero values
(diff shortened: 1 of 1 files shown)


chennes:
@Roy-043 have you had a chance to look at this to make sure it solves the issues you were concerned about? In my testing it seems to.

Roy-043:
@Rexbas Thanks!

@chennes I do not compile myself. Please merge.

Rexbas:
I am not sure if a screenshot makes sense. This PR did not do much except change some calculations to be more accurate and consistent but did not change or add a new feature. This PR may have fixed some problems elsewhere. If it does, we should probably also backport this.

PR #10706, opened 2023-09-16
Gui: Transform dragger improvements

qewer33:
This PR improves the Transform tool dragger. It isn't based on an issue, I can open a discussion issue if necessary.

This PR:
- Improves the appearance of the dragger making it look more modern and user friendly:

image

- Implements additional sub-draggers to move object on a plane (see video below):

[vokoscreenNG-2023-09-17_00-06-27.webm](https://github.com/FreeCAD/FreeCAD/assets/69015181/3a016158-f053-4e82-bcbb-386480ddca68)

- Adds new parameters for "axis colors". Currently only the Transform dragger uses these colors but FreeCAD has quite a bit of axis elements and it would be great to unify their colors and make them configurable in `Preferences > Display > Colors` in the future. 

Any reviews and feedback is appreciated.

### DIFF
--- a/src/Gui/SoFCCSysDragger.cpp
+++ b/src/Gui/SoFCCSysDragger.cpp
@@ -35,6 +35,7 @@
 #include <Inventor/nodes/SoCone.h>
 #include <Inventor/nodes/SoCoordinate3.h>
 #include <Inventor/nodes/SoCylinder.h>
+#include <Inventor/nodes/SoCube.h>
 #include <Inventor/nodes/SoGroup.h>
 #include <Inventor/nodes/SoLineSet.h>
 #include <Inventor/nodes/SoOrthographicCamera.h>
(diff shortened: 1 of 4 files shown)


WandererFan:
Love seeing transform get some love.

My 2 big complaints about Transform (which I use a lot) are that the dragger obscures the object and the dragger is nowhere near the object (using Placement instead of object centroid?).  Not always, but sometimes.

qewer33:
> Love seeing transform get some love.

I also use it a lot, it doesn't get much improvements in general so I decided to give it a shot myself :D 

> My 2 big complaints about Transform (which I use a lot) are that the dragger obscures the object 

Well this isn't a problem that can be avoided but maybe making the dragger a bit smaller might help with this. Screenshot from Godot (FOSS game engine) for reference:

image

> and the dragger is nowhere near the object (using Placement instead of object centroid?). Not always, but sometimes.

I also get this sometimes (mostly happens with imported STEP files) and I honestly have no idea what causes it. One possible "solution" I've though about is putting "Transform Origin Offset X/Y" options in the Task dialog. This wouldn't necessarily solve that issue but at least it'll also help with rotating on points that are not the origin of the object. Not sure if it would be easy to implement though.

qewer33:
Another Transform dragger issue that I'd like to fix is (this issue exists in current versions too, not a regression caused by this PR): The rendering order doesn't take the viewing angle into account and just renders stuff based on the OpenInventor scene traversal. I've done some research but I have no idea how to fix it. Would appreciate help from anyone knowledgeable with OpenInventor/Coin/OpenGL.

image

FEA-eng:
The transform tool is very useful so I also appreciate it being improved. Would it be possible to add an option to locate the tool at the center of the mass or at the selected geometry (like "Transform at" available in Realthunder's branch) on this occasion ?

luzpaz:
> Another Transform dragger issue that I'd like to fix is (this issue exists in current versions too, not a regression caused by this PR): The rendering order doesn't take the viewing angle into account and just renders stuff based on the OpenInventor scene traversal. I've done some research but I have no idea how to fix it. Would appreciate help from anyone knowledgeable with OpenInventor/Coin/OpenGL.

@wwmayer any chance you can take a look at https://github.com/FreeCAD/FreeCAD/pull/10706#issuecomment-1722431251 ?

qewer33:
> wwmayer any chance you can take a look at [#10706 (comment)](https://github.com/FreeCAD/FreeCAD/pull/10706#issuecomment-1722431251) ?

Would be great if you could review this PR too while you're at it if you're free :sweat_smile:

qewer33:
> The transform tool is very useful so I also appreciate it being improved. Would it be possible to add an option to locate the tool at the center of the mass or at the selected geometry (like "Transform at" available in Realthunder's branch) on this occasion or is it too much for now ?

I have plans to attempt something like that after this is merged. Probably something like "Offset Transform Origin" with XYZ input + "Center Of Mass" and "Zero" buttons.

sliptonic:
> It isn't based on an issue, I can open a discussion issue if necessary.

That's ok this time.  In future, please do the issue first to keep discussion between problem and solution separate

wwmayer:
> any chance you can take a look at https://github.com/FreeCAD/FreeCAD/pull/10706#issuecomment-1722431251

The PR has been merged before I had the chance to look at the code. Basically the code looks good and seems to work as expected but there are a few issues that should be solved:
* Several empty newlines between two functions are removed. They should be kept.
* SoFCCSysDragger has a dependency to  the singleton ViewParams now which has a code smell because it has global behaviour. It's actually as bad as working with global variables. Instead you can add some public class members of type SbColor to SoFCCSysDragger and then set them inside ViewProviderDragger (or TaskCSysDragger)
* In SoFCCSysDragger don't use App::Color but SbColor

PR #10712, opened 2023-09-17
Draft: shape2dview: avoid issues with intersecting solids

Roy-043:
Add `Shape.SubShapes` instead of a copy of the shape. This avoids cut problems if there are intersecting solids.

Forum topic:
https://forum.freecad.org/viewtopic.php?t=81240


yorikvanhavre:
Hm interesting find.. That might solve other issues as well. I'll merge already, and give it a test spin

PR #10714, opened 2023-09-17
[Macro Execute Dialog] add file name and file content filtering

mwganson:
This adds 2 labels and 2 line edits above the macro files list box:

Find file: [LineEditFile] Find in files: [LineEditFindInFiles]
[macro file names list box]

LineEditFile text is used to search the filenames for matches to the filter the user has entered.  Example: if the user types in "a" (without the quotes) then only files with "a" in the filename get listed in the list box.  Regular expressions are supported.  A check is done if the text entered is a valid regex expression.  If it is not valid, then we use a simpler text search with QString::contains().

LineEditFindInFiles text is used to further filter the results of the first filter to only include files containing text entered in LineEditFindInFiles.  This is also a regular expression search unless it's an invalid regex expression, in which case QString::contains() is used instead.

In the filtering process we skip as many loops as possible.  For example, if there is no content filter then we don't bother loading and reading the files to search for matches to the empty string.  If there is no filename filter, then we just apply the content filter to the unfiltered filenames.

All searches are case-insensitive.
Snip macro screenshot-4e5b33

Notice in the screenshot, the regular expression calls for file names with an _ followed by a capital letter, and yet the results contain lower case matches.  I considered adding a checkbox for case-sensitive, but I don't want to clutter up the dialog.

### DIFF
--- a/src/Gui/DlgMacroExecuteImp.cpp
+++ b/src/Gui/DlgMacroExecuteImp.cpp
@@ -27,6 +27,7 @@
 # include <QMessageBox>
 # include <QComboBox>
 # include <QSignalBlocker>
+# include <QTextStream>
 #endif
 
 #include <App/Document.h>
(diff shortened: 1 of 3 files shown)


mwganson:
Forum link:

https://forum.freecad.org/viewtopic.php?t=81255

src/Gui/DlgMacroExecuteImp.cpp
```diff
@@ -188,6 +262,14 @@ void DlgMacroExecuteImp::onUserMacroListBoxCurrentItemChanged(QTreeWidgetItem* i
     }
 }
 
+void DlgMacroExecuteImp::onLineEditFindTextChanged(const QString &){
```
chennes, line 265: The linter is yelling here because in general it's recommended to name the variable, and then use the Q_UNUSED macro on it, so it's explicit that you're deliberately ignoring it.

mwganson: I've made this change.  Waiting for the unit tests to finish, and then I'll push the update.  There was another change I noticed, too.  QtDesigner added native=true or something to the file chooser.  I hate it when software does stuff like that behind my back.

chennes:
Works great, nice feature -- thanks!

mwganson:
Thanks for merging.  I'll go ahead and start working on the wiki documentation.

Syres916:
@mwganson could you please advise why the ... button at the bottom right hand corner next to Macros Location has changed from _Search for a Folder_ to _Search for a File_, surely a location is a folder and not a file???

mwganson:
> @mwganson could you please advise why the ... button at the bottom right hand corner next to Macros Location has changed from _Search for a Folder_ to _Search for a File_, surely a location is a folder and not a file???

QtDesigner made that change behind my back and I didn't notice it.  I'll submit a PR for it.

mwganson:
https://github.com/FreeCAD/FreeCAD/pull/11670

PR #10721, opened 2023-09-18
Sketcher: add points getter helpers

PaddleStroke:
Following on https://github.com/FreeCAD/FreeCAD/pull/10666

This enables to get centerpoints, startpoint and endpoints easily and will lighten the syntax a lot all over the place in sketcher.
Also it enables getting them in vector2d or Vector3d which avoids converting them which is also making syntax very heavy.
I can't remember how many times I wrote huge blocks of code like 
```
    if (isArcOfEllipse(*geom2)) {
        center = (static_cast<const Part::GeomArcOfEllipse*>(geom2))->getCenter();
    }
    else if (isEllipse(*geom2)) {
        center = (static_cast<const Part::GeomEllipse*>(geom2))->getCenter();
    }
    else if (isCircle(*geom2)) {
        center = (static_cast<const Part::GeomCircle*>(geom2))->getCenter();
    }
    else if (isArcOfCircle(*geom2)) {
        center = (static_cast<const Part::GeomArcOfCircle*>(geom2))->getCenter();
    }
    Base::Vector2d vec = p1 + Base::Vector2d(center.x, center.y);
```

@FlachyJoe @wwmayer @chennes


chennes:
It still feels like a lot of repetition, most of which is the conditionals -- would it make sense to have a single function that returned all three?

FlachyJoe:
@PaddleStroke 
SketchObject::getPoint return a Vector3d so it has to be changed too or your helpers should return Vector3d to be homogeneous.
Vector3d are used very often in CommandConstraint even for 2d calculation.

But SketchObject::getPoint already have all the needed stuff, take a look at https://github.com/FreeCAD/FreeCAD/blob/4293c4a65b0e69975dcdbbf68c80d8019a4f850d/src/Mod/Sketcher/App/SketchObject.cpp#L837-L901

@chennes
the given example is the current code and will become
```cpp
    Base::Vector2d vec = p1 + getCenter(geom2);
```

PaddleStroke:
@FlachyJoe ah I didn't knew of this getPoint function. Has it been there for a long time? I haven't seen it used yet.
Maybe it would be best in utils so that it does not require to type 'getSketchObject()->'
Then this PR needs a bit more of thinking.

PaddleStroke:
So here is my updated proposal.
This moves getPoint code to Utils. As it does not rely on sketchObject. The SketchObject::getPoint function is left to ensure nothing breaks.
Make functions to get the points either in 2d or 3d.
Also make small helpers to toggle between 2d & 3d.

FlachyJoe:
Utils resides in GUI, is it really the good place ?
Code in SketchObject could also be refactored with helpers if Utils is split in a GUI file and an App file. Or go to workbench top dir.
Maybe my next PR.

PaddleStroke:
Yes that was my thought as well. IsCircle and co are also useful in App not only gui. So it would make sense that Utils goes to top dir.

FlachyJoe:
In facts the helpers I introduce shouldn't reside here because all the other content are in the SketcherGui namespace. But some of the other are not Gui specific and could go to the Sketcher namespace and in an upper file.

PaddleStroke:
Or perhaps we just need a Utils file in App? As these helper functions feels like they should belong to App.

chennes:
@PaddleStroke can you fix that function call that's failing? I think it's just a typo.

FlachyJoe:
@chennes @PaddleStroke not just a typo. App shouldn't depend on the Gui code.

PaddleStroke:
I created Utils files in the App module where I moved all the new helper functions.

wwmayer:
`#include <QCursor>` is a GUI thing and cannot be part of an App module.

PaddleStroke:
Sorry, I forgot to clean the headers in the `#ifndef _PreComp_`

PaddleStroke:
It appears this is now stuck because App and Gui cannot share utils functions directly. I ran in the same problem in another wb. See https://forum.freecad.org/viewtopic.php?t=83974

PR #10723, opened 2023-09-19
TechDraw: create closure for concurrent thread context

andre-caldas:
We use a lambda function with a copy of variables that  might be destructed in the original calling thread, possibly producing dangling references.

See:
https://forum.freecad.org/viewtopic.php?t=81260

All tests have passed.

### DIFF
--- a/src/Mod/TechDraw/App/DrawComplexSection.cpp
+++ b/src/Mod/TechDraw/App/DrawComplexSection.cpp
@@ -281,11 +281,12 @@ void DrawComplexSection::makeSectionCut(const TopoDS_Shape& baseShape)
         connectAlignWatcher =
             QObject::connect(&m_alignWatcher, &QFutureWatcherBase::finished, &m_alignWatcher,
                              [this] { this->onSectionCutFinished(); });
-#if QT_VERSION < QT_VERSION_CHECK(6,0,0)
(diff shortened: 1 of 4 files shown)


luzpaz:
```bash
git commit --amend -m "TechDraw: create closure for concurrent thread context" -m "We use a lambda function with a copy of variables that might be destructed in the original calling thread, possibly producing dangling references." -m "See: https://forum.freecad.org/viewtopic.php?t=81260" -m "All tests have passed."
git push --force
```

andre-caldas:
> ```shell
> git commit --amend -m "TechDraw: create closure for concurrent thread context" -m "We use a lambda function with a copy of variables that might be destructed in the original calling thread, possibly producing dangling references." -m "See: https://forum.freecad.org/viewtopic.php?t=81260" -m "All tests have passed."
> git push --force
> ```

Thank you... and sorry! :-)
I tried to fix the comment but I could not figure how to do it.

WandererFan (commented):
LGTM, but I barely understand what is going on.  We should have somebody competent like Werner look at it.

wwmayer (commented):
I don't see that technically this makes a big difference. It just  makes it more explicit that the capture of the lambda copies the passed shape.

But the good thing about this PR is that it eliminates the differences between Qt5 & Qt6

Issue #10726, opened 2023-09-19
[Problem] Part HLR Projector Python API misses argument


pieterhijma:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

When applying HLR via the Part API, you will have to set the projector. However, there is a small bug in the Python wrapper code that prevents you to do so. To reproduce:

``` python
>>> origin = App.Vector(0, 0, 0)
>>> direction = App.Vector(0, 1, 0)
>>> xDirection = App.Vector(1, 0, 0)
>>> algo = Part.HLRBRep.Algo()
>>> algo.setProjector(origin, direction, xDirection)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
SystemError: more argument specifiers than keyword list entries (remaining format:'d')
>>> algo.setProjector(origin, direction, xDirection, 1.0)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
TypeError: function takes at most 3 arguments (4 given)
```

### Full version info

```shell
[code]
OS: Arch Linux
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34337 (Git)
Build type: Unknown
Branch: master
Hash: 4293c4a65b0e69975dcdbbf68c80d8019a4f850d
Python 3.11.3, Qt 5.15.10, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * osh-autodoc-workbench
[/code]
```

### Subproject(s) affected?

Part

### Anything else?

This [forum post](https://forum.freecad.org/viewtopic.php?p=699331#p699331) discusses the requested behavior. It is a very small problem, and I will submit a pull request that fixes this shortly.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

pieterhijma:
Yes, fully solved with PR #10727.

PR #10727, opened 2023-09-19
Part: Fix a missing argument for HLR projector

pieterhijma:
As discussed in #10726, this small pull request fixes it.

The following listing works as opposed in the issue

``` python
>>> origin = App.Vector(0, 0, 0)
>>> direction = App.Vector(0, 1, 0)
>>> xDirection = App.Vector(1, 0, 0)
>>> algo = Part.HLRBRep.Algo()
>>> algo.setProjector(origin, direction, xDirection)
>>> algo.setProjector(origin, direction, xDirection, 1.0)
```

### DIFF
--- a/src/Mod/Part/App/HLRBRep/HLRBRep_AlgoPyImp.cpp
+++ b/src/Mod/Part/App/HLRBRep/HLRBRep_AlgoPyImp.cpp
@@ -114,7 +114,7 @@ PyObject* HLRBRep_AlgoPy::setProjector(PyObject *args, PyObject *kwds)
     PyObject* xd = nullptr;
     double focus = std::numeric_limits<double>::quiet_NaN();
 
-    static const std::array<const char *, 4> kwlist {"Origin", "ZDir", "XDir", nullptr};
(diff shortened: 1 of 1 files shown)


luzpaz:
An aside, is there no API documentation for this functionality ? (I couldn't find an associated header or doxygen comments on it

pieterhijma:
> An aside, is there no API documentation for this functionality ? (I couldn't find an associated header or doxygen comments on it)
> 
> Edit: I opened a separate issue #10728 to address this

Wow you're fast, you beat me to it. Yes, that's right. I wanted to create that issue as well. If you want, you can assign it to me. Since I'm using the API, the various functionality is fresh to me, so it probably takes me less time than anyone else at the moment.

pieterhijma:
> I don't know how often perspective projection is used in practice, but it should be available if required.
> 
> LGTM!

Thanks for merging so fast! It's not only perspective projection that didn't work, the code below doesn't work either (see #10726):

``` python
>>> algo.setProjector(origin, direction, xDirection)
SystemError: more argument specifiers than keyword list entries (remaining format:'d')
```

PR #10732, opened 2023-09-19
Gui: Add Workbench Manipulator

wwmayer:
The WorkbenchManipulator class provides a mechanism to customize a workbench when it is about to be activated. For example, extra commands can be added to any workbench.

Background: PR #10271 added several selection filter classes to the core system but these commands only work for Part::Feature objects. Thus, it's a design-flaw to put them to the core system instead to the Part module.
But because these selection filter commands should be available independent of the active workbench a mechanism is needed to support this.

Commit 1: Implements the base class WorkbenchManipulator
Commit 2: Implements a derived class for the Part module that handles Part_SectionCut to be always shown in the Views menu
Commit 3: Handles the CmdSelectFilter classes
Commit 4: Moves the CmdSelectFilter classes to the Part module


sliptonic:
Should the developer handbook be updated with a concrete example of using this?

wwmayer:
> Should the developer handbook be updated with a concrete example of using this?

If this is needed for others to understand how it's done then go ahead. You can copy the code from PartGui::WorkbenchManipulator that handles the Std_ToggleClipPlane command.

luzpaz:
Just saying, but there may be some confusion for folks between this and the 3rd party addon Manipulator workbench[^1][^2]

[^1]: https://github.com/easyw/Manipulator
[^2]: https://wiki.freecad.org/Manipulator_Workbench

wwmayer:
> Just saying, but there may be some confusion for folks between this and the 3rd party addon Manipulator workbench

OMG! This is just the name of a class and it appears nowhere in the GUI. Really, I don't see a reason why this should lead to any confusion.

yorikvanhavre:
Do I undersand correctly @wwmayer ?

1. The Part WB creates a WB manipulator, and sets it up with the needed commands (filter command in this case)
2. The manipulator is FreeCAD-wide, it "stays there",  a bit like an observer
3. When a WB is activated, the existing WB manipulator(s) is/are applied

Only thing I didn't understand, the mechanism that creates the manipulator is in AppPartGui. But that is only loaded when Part is activated (or PartGui py module loaded) right? How then is the manipulator created without loading Part?

wwmayer:
> (filter command in this case)

The filter and section cut commands.

> The manipulator is FreeCAD-wide, it "stays there", a bit like an observer

It's not an observer in the sense of the design pattern. But yes a manipulator is permanent (unless it's explicitly removed).

> When a WB is activated, the existing WB manipulator(s) is/are applied

Exactly. It allows a module to add commands to the activated workbench which this workbench may don't know of because it's part of a different module.

The basic idea is to make the workbench concept more flexible and at the same time to keep the SW design clean.

> Only thing I didn't understand, the mechanism that creates the manipulator is in AppPartGui. But that is only loaded when Part is activated (or PartGui py module loaded) right? How then is the manipulator created without loading Part?

Of course, that's the whole idea and intention. The selection filter commands only work for Part objects and if no Part module is loaded then these commands are useless anyway.

Background:
The initial implementation of the selection filter put the commands to the core system. But these commands are focused on Part features and thus rely on the existence and details of the Part module. If a user-defined FreeCAD is built without the Part module these commands are totally useless.

wwmayer:
> Would you be interested in adding a python version? 😇 (I can try it too, I guess I'd look at how the python version of the doc observer is done)

Writing the class WorkbenchManipulatorPython that accepts a Py::Object and calls the relevant methods of the Python type is trivial but exchanging the information of what exactly should be changed is a bit more tricky. One has to define some kind of protocol of what should be changed on the structures.

yorikvanhavre:
You go too fast :fearful:

Issue #10733, opened 2023-09-19
[Problem] TechDraw: Expose SVG export edges


pieterhijma:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

With two functions exposed from TechDraw, Python code and workbenches in particular would have access to generate SVG images based on Part HLR without having to duplicate code that is in the TechDraw module.

### Full version info

```shell
[code]
OS: Arch Linux
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34337 (Git)
Build type: Unknown
Branch: master
Hash: 4293c4a65b0e69975dcdbbf68c80d8019a4f850d
Python 3.11.3, Qt 5.15.10, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * osh-autodoc-workbench
[/code]
```

### Subproject(s) affected?

Techdraw

### Anything else?

An example of what would be nice to have is this [forum post](https://forum.freecad.org/viewtopic.php?p=699331#p699331).

I will issue a pull request shortly that allows the example in the forum post.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
@pieterhijma is this ticket complete ?

pieterhijma:
Yes, by means of PR #10734.

PR #10734, opened 2023-09-19
TechDraw: Expose two functions for SVG export

pieterhijma:
This pull requests exposes two functions to Python that allows Python code and workbenches to generate SVG images. See #10733.

This pull request exposes `TechDraw.build3dCurves()` and `TechDraw.exportSVGEdges()` and it is a convenient addition for workbenches that want more control over SVG export, possibly in relation to Part HLR, see for example #10726.

This pull request is independent of but related to #10726. In order to test the code below with the Part HLR module, it is necessary to have the changes from #10727. With that code in place, the following is possible creating an SVG from Python like so:

``` python
# create object
doc = App.ActiveDocument
doc.addObject("Part::Box","Box")
obj = doc.ActiveObject
obj.Label = "Cube"
doc.recompute()
shape = obj.Shape

# find camera position
view.viewIsometric()
direction = view.getViewDirection()
xDirection = App.Vector(0.7, 0.7, 0)
origin = App.Vector(0, 0, 0)

# perform HLR
algo = Part.HLRBRep.Algo()
algo.add(shape)
algo.setProjector(origin, direction, xDirection)
algo.update()
algo.hide()
projShapes = Part.HLRBRep.HLRToShape(algo)

# create the SVG with help from TechDraw
import TechDraw as td
v = td.build3dCurves(projShapes.vCompound(shape))
edges = td.exportSVGEdges(v)
bb = v.BoundBox

with open('hlr.svg', 'w') as f:
    f.write("<svg\n")
    f.write(f"       width=\"{bb.XLength}\"\n")
    f.write(f"       height=\"{bb.YLength}\"\n")
    f.write(f"       viewBox=\"0 0 {bb.XLength} {bb.YLength}\"\n")
    f.write("	xmlns=\"http://www.w3.org/2000/svg\" version=\"1.1\"\n")
    f.write("	xmlns:freecad=\"http://www.freecadweb.org/wiki/index.php")
    f.write("?title=Svg_Namespace\">\n")
    f.write('<g stroke=\"#000000\"\n')
    f.write('   stroke-width=\"0.4px\"\n')
    f.write('   stroke-linecap=\"round\"\n')
    f.write('   stroke-linejoin=\"round\"\n')
    f.write('   fill=\"none\"\n')
    f.write(f'   transform=\"matrix(1, 0, 0, -1, {-bb.XMin}, {bb.YMax})\">\n')
    f.write(edges)
    f.write("\n</g>")
    f.write("\n</svg>")
```

### DIFF
--- a/src/Mod/TechDraw/App/ProjectionAlgos.cpp
+++ b/src/Mod/TechDraw/App/ProjectionAlgos.cpp
@@ -50,6 +50,18 @@ using namespace std;
 // ProjectionAlgos
 //===========================================================================
 
+namespace TechDraw {
+  //added by tanderson. aka blobfish.
+  //projection algorithms build a 2d curve(pcurve) but no 3d curve.
+  //this causes problems with meshing algorithms after save and load.
+  const TopoDS_Shape& build3dCurves(const TopoDS_Shape &shape)
+  {
+    TopExp_Explorer it;
+    for (it.Init(shape, TopAbs_EDGE); it.More(); it.Next())
+      BRepLib::BuildCurve3d(TopoDS::Edge(it.Current()));
+    return shape;
+  }
+}
 
 ProjectionAlgos::ProjectionAlgos(const TopoDS_Shape &Input, const Base::Vector3d &Dir)
   : Input(Input), Direction(Dir)
@@ -61,16 +73,6 @@ ProjectionAlgos::~ProjectionAlgos()
 {
 }
 
-//added by tanderson. aka blobfish.
-//projection algorithms build a 2d curve(pcurve) but no 3d curve.
-//this causes problems with meshing algorithms after save and load.
-static const TopoDS_Shape& build3dCurves(const TopoDS_Shape &shape)
-{
-  TopExp_Explorer it;
-  for (it.Init(shape, TopAbs_EDGE); it.More(); it.Next())
-    BRepLib::BuildCurve3d(TopoDS::Edge(it.Current()));
-  return shape;
-}
 
 void ProjectionAlgos::execute()
 {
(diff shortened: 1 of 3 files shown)


src/Mod/TechDraw/App/ProjectionAlgos.cpp
```diff
@@ -50,6 +50,18 @@ using namespace std;
 // ProjectionAlgos
 //===========================================================================
 
+namespace TechDraw {
+  //added by tanderson. aka blobfish.
+  //projection algorithms build a 2d curve(pcurve) but no 3d curve.
+  //this causes problems with meshing algorithms after save and load.
+  const TopoDS_Shape& build3dCurves(const TopoDS_Shape &shape)
```
WandererFan, line 57: Did you mean to remove "static" on build3dCurves?   If it is static, it can be used without a ProjectionAlgo instance.

I don't think this routine is widely used without the rest of ProjectionAlgo, so maybe it doesn't matter.

pieterhijma: Yes, this was intentionally, thanks for noticing, because I realize that I should have mentioned my design choices. I can imagine you would like to see things a bit differently.

I believe `static` here does not mean that `build3dCurves()` is a class function (that doesn't require an instance). It's not part of the `ProjectionAlgo` class. Instead, I think the author of the code wanted to express that this is a small helper function local to this translation unit, so no other compiled needs to link to this function. However, for my changes, I do want to link to this function (from `AppTechDrawPy.cpp`), so I needed to remove the `static` keyword.

In addition, I needed to add the `TechDraw` namespace to make it accessible. However, I can imagine the location of this function is a bit strange now. It used to be a helper function local to the file for the `ProjectionAlgo` functions. Now it is a function accessible from the `TechDraw` namespace. I chose to leave the function in this file because it originates from here, for easier review, and because I couldn't find a better location, but I can imagine that you prefer to relocate the function, perhaps in a TechDraw file with similar kinds of utility functions.

PR #10741, opened 2023-09-19
Arch: avoid intersecting solids in windows and doors with openings

Roy-043:
With the current code, if you place a simple door and change its Opening property to 50, the door leaf and the frame intersect. This PR fixes that issue.


yorikvanhavre:
Just tested, LGTM!

PR #10743, opened 2023-09-19
[Sketcher] Fix angle constraint expression regression

### LINKED ISSUE #10742: [Regression] Sketcher angle constraint expression handling

### Problem description

The new complementary angle functionality introduced in #10657 does not handle expressions correctly.

Test expression: 60
Complementary: 180 - 60
Back to original: 180 - (180 - 60)

Test expression 2: 60,00 °
Complementary: 180 - 60 ° _(Should be: 180 ° - 60 °)_

This gives an error:
`<class 'ArithmeticError'>: Quantity::operator -(): Unit mismatch in minus operation
in expression: 180 - 60 °
in property binding 'Unnamed#Sketch.Constraints'`

[angle_expr.webm](https://github.com/FreeCAD/FreeCAD/assets/82042580/6e353a66-c17a-425a-96d0-52d8740b100d)

### Full version info

```shell
[code]
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34352 (Git)
Build type: Release
Branch: master
Hash: ab4abab53d6cc81ffe834d7a34620f8fae4f3c6e
Python 3.11.5, Qt 5.15.10, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
[/code]
```

### Subproject(s) affected?

Sketcher

### Anything else?

@PaddleStroke 

### Code of Conduct

howie-j:
Fixes #10742.

Unsure if this is the correct way to do this, but it works as far as i can tell.


PaddleStroke:
Thanks for jumping in ! 
Does this also fix the case 

> Test expression: 60
> Complementary: 180 - 60
> Back to original: 180 - (180 - 60)

I can't tell for sure just reading the code

howie-j:
Yep, it fixes both cases👍

wwmayer:
The code for the fix (i.e. the modification of the string) should be moved to a helper function (e.g. a static public function in SketchObject) and then a unit test should be added to the Sketcher_tests_run target.

howie-j:
I'm in deep water when it comes to writing a unit test..

If i can get some help setting up TEST_F() then i can add some asserts myself :)

chennes:
> If i can get some help setting up TEST_F() then i can add some asserts myself :)

Sure, I can set up the test framework for you and push a change to your PR branch.

howie-j:
Thanks @chennes, go ahead!

chennes:
OK, I didn't actually put any asserts in, but I gave you a couple of example tests that show how you can set things up.

howie-j:
Thanks for the setup!

Shouldn't i be able to do something like `EXPECT_EQ("60", getObject()->getConstraintExpression(id));`?
It seems no matter what i do `getObject()->getConstraintExpression(id)` only returns `""`.

```
[ RUN      ] SketchObjectTest.testReverseAngleConstraintToSupplementaryExpressionSimple
/home/howie/git/fc_dev/tests/src/Mod/Sketcher/App/SketchObject.cpp:179: Failure
Expected equality of these values:
  "60"
  getObject()->getConstraintExpression(id)
    Which is: ""
[  FAILED  ] SketchObjectTest.testReverseAngleConstraintToSupplementaryExpressionSimple (1 ms)
[ RUN      ] SketchObjectTest.testReverseAngleConstraintToSupplementaryExpressionApplyAndReverse
[       OK ] SketchObjectTest.testReverseAngleConstraintToSupplementaryExpressionApplyAndReverse (0 ms)
[----------] 9 tests from SketchObjectTest (15 ms total)

```

wwmayer:
The idea of moving the code to a separate function and making it static was that you can test it directly without a dependency to anything else.

I still recommend to make the function static and pass the string by value and not by reference. 

Then your test functions can directly do what they are supposed to do:

```
TEST_F(SketchObjectTest, testReverseAngleConstraintToSupplementaryExpressionSimple)
{
    std::string expr = Sketcher::SketchObject::reverseAngleConstraintExpression("180 - (60)");
    EXPECT_EQ(expr, std::string("(60)"));
}

TEST_F(SketchObjectTest, testReverseAngleConstraintToSupplementaryExpressionApplyAndReverse)
{
    std::string expr = "180";
    expr = Sketcher::SketchObject::reverseAngleConstraintExpression(expr);
    expr = Sketcher::SketchObject::reverseAngleConstraintExpression(expr);
    EXPECT_EQ(expr, std::string("(180)"));
}
```

Of course, it should still be figured out why getConstraintExpression() returns an empty string but this is a separate problem.

wwmayer:
> why getConstraintExpression() returns an empty string

This is because the sketch object doesn't have an expression set. Before you can set a constraint expression you have to set an expression:
```
...
auto id = getObject()->addConstraint(constraint);

App::ObjectIdentifier path(App::ObjectIdentifier::parse(getObject(), "Constraints[0]"));
std::shared_ptr<App::Expression> shared_expr(App::Expression::parse(getObject(), "0"));
getObject()->setExpression(path, shared_expr);

getObject()->setConstraintExpression(id, "180");
...
```

howie-j:
Many thanks @wwmayer and @chennes, c++ is scary for a simple EE but i'm trying to learn 😅

`reverseAngleConstraintExpression()` is now pass by value.
Is it correct to use static in the header file declaration?

Also i went nuts and added many unit tests, and kept the two functional tests. Do we keep all or or should i try to keep only the minimum possible? I'm not very familiar with best practices..

At least all tests passes🥳

PR #10761, opened 2023-09-20
Toolbar View and Structure cleaning

### LINKED ISSUE #10755: [Problem] 'Bounding box' is seldomly used

### Problem description

The view toolbar is cluttered. And in general we have too many useless icons cluttering the view.

'Bounding box' is seldomly used and doesn't appear to have a clear usecase. So a position in the toolbar feels unecessary and its presence in the view menu should be enough.

Proposal : Remove it from the toolbar, keep it in menu.

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10756: [UI Problem] 'std_selBack' 'std_selForward' are seldomly used

### Problem description

'std_selBack' 'std_selForward' : Dubious utility. Though for super huge files I can somehow see the use. On top of a dubious utility, the name and tooltips are not clear on what they do. And these arrows icons are semantically identical to 'undo' and 'redo' making things confusing.

I think these should be in the view/Treeview actions menu instead.

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10757: [UI Problem]  'Go to link object' position in the view toolbar is questionable.

### Problem description

'Go to link object' command group : These would be better as right click contextual menu. Appearing only when selection is correct (Which is already the case).

I don't see that its more intuitive or usefull in the toolbar. Because you first need to select the item in the tree, then go to the command. If it was in context menu you would directly right click on it.

I propose to remove the command group from the toolbar.

See discussion : https://forum.freecad.org/viewtopic.php?t=81319

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10758: [UI Problem] Structure toolbar : 'Make link' should be in command group.

### Problem description

In structure toolbar we currently have 2 entries : 
- Make link
- Make sub-link group command.

I don't see why we don't have a single entry and Make Link becomes the first one of the group.

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Anything else?

_No response_

### Code of Conduct

PaddleStroke:
This PR does several things : 
- Put Link in the command group
- Give Link a proper tooltip (screenshot outdated: tooltips changed).
image

- Put sel_back and sel_forward in the treeview actions group rather than in root of view toolbar. This way they now appear in the view menu (they weren't before) and they don't appear in the root of view toolbar.
- Give them proper names and tooltips (screenshot outdated: tooltips changed)
image

- Remove 'Link Navigation' from the view toolbar
- Put 'Link Navigation' in the view menu where it was missing.
- Remove 'Bouding box' from the view toolbar (its still in view menu)
- Reorder the view toolbar so that the order makes sense.

image

Lastly it improves the tooltips of std_part and std_group.

Fixes #10755
Fixes #10756
Fixes #10757
Fixes #10758

### DIFF
--- a/src/Gui/CommandLink.cpp
+++ b/src/Gui/CommandLink.cpp
@@ -206,7 +206,11 @@ StdCmdLinkMake::StdCmdLinkMake()
 {
     sGroup        = "Link";
     sMenuText     = QT_TR_NOOP("Make link");
-    sToolTipText  = QT_TR_NOOP("Create a link to the selected object(s)");
+    sToolTipText  = QT_TR_NOOP("<p>An App::Link is a type of object that references or links"
+        " to another object, in the same document, or in another document. Unlike Clones, Link"
+        " references directly the original Shape, so it is more memory efficient which helps "
+        "with the creation of complex assemblies from smaller subassemblies, and from multiple"
+        " reusable components like screws, nuts, and similar fasteners.</p>");
     sWhatsThis    = "Std_LinkMake";
     sStatusTip    = sToolTipText;
     eType         = AlterDoc;
--- a/src/Gui/CommandLink.cpp
+++ b/src/Gui/CommandLink.cpp
@@ -214,7 +214,7 @@ StdCmdLinkMake::StdCmdLinkMake()
 }
 
 bool StdCmdLinkMake::isActive() {
-    return !!App::GetApplication().getActiveDocument();
+    return App::GetApplication().getActiveDocument();
 }
 
 void StdCmdLinkMake::activated(int) {
@@ -878,6 +878,7 @@ public:
         eType         = AlterDoc;
         bCanLog       = false;
 
+        addCommand(new StdCmdLinkMake());
         addCommand(new StdCmdLinkMakeRelative());
         addCommand(new StdCmdLinkReplace());
         addCommand(new StdCmdLinkUnlink());
@@ -898,7 +899,6 @@ namespace Gui {
 void CreateLinkCommands()
 {
     CommandManager &rcCmdMgr = Application::Instance->commandManager();
-    rcCmdMgr.addCommand(new StdCmdLinkMake());
     rcCmdMgr.addCommand(new StdCmdLinkActions());
     rcCmdMgr.addCommand(new StdCmdLinkMakeGroup());
     rcCmdMgr.addCommand(new StdCmdLinkSelectActions());
--- a/src/Gui/CommandLink.cpp
+++ b/src/Gui/CommandLink.cpp
@@ -206,11 +206,12 @@ StdCmdLinkMake::StdCmdLinkMake()
 {
     sGroup        = "Link";
     sMenuText     = QT_TR_NOOP("Make link");
-    sToolTipText  = QT_TR_NOOP("<p>An App::Link is a type of object that references or links"
-        " to another object, in the same document, or in another document. Unlike Clones, Link"
-        " references directly the original Shape, so it is more memory efficient which helps "
-        "with the creation of complex assemblies from smaller subassemblies, and from multiple"
-        " reusable components like screws, nuts, and similar fasteners.</p>");
+    static std::string toolTip = std::string("<p>")
+        + QT_TR_NOOP("A Link is an object that references or links to another object in the same "
+        "document, or in another document.Unlike Clones, Links reference the original Shape directly, "
+        " making them more memory efficient which helps with the creation of complex assemblies.")
+        + "</p>";
+    sToolTipText = toolTip.c_str();
     sWhatsThis    = "Std_LinkMake";
     sStatusTip    = sToolTipText;
     eType         = AlterDoc;
--- a/src/Gui/CommandStructure.cpp
+++ b/src/Gui/CommandStructure.cpp
@@ -49,7 +49,11 @@ StdCmdPart::StdCmdPart()
 {
     sGroup        = "Structure";
     sMenuText     = QT_TR_NOOP("Create part");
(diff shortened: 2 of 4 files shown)


src/Gui/CommandLink.cpp
```diff
@@ -206,15 +206,19 @@ StdCmdLinkMake::StdCmdLinkMake()
 {
     sGroup        = "Link";
     sMenuText     = QT_TR_NOOP("Make link");
-    sToolTipText  = QT_TR_NOOP("Create a link to the selected object(s)");
+    sToolTipText  = QT_TR_NOOP("<p>An App::Link is a type of object that references or links"
```
wwmayer, line 209: It's always cumbersome to put technical elements (here the HTML tags) into translatable text because as soon as a translator doesn't do it correctly the displayed string will be broken.

And I don't think we should mention App::Link in this text because it's a programming term and  a regular user shouldn't be bothered with such terms. Same with App::Part in the tooltip below.

PaddleStroke: What is the proper way then? Without it the tooltip is going to be all in one line. Then you can put \n to break into lines, but it doesn't look as good and you have the same problem as with <p>.
I thought about using `sToolTipText  = "<p>" + QT_TR_NOOP("abc") + "</p>"` but this is not accepted.

Roy-043:
The texts for the new tooltips seem to be directly copied from the Wiki. This does not necessarily work well. Some consitency with the other tooltips in the same menu/toolbar may also be desirable. I also think that a tooltip should be relatively short.

Example:

Wiki text (and current tooltip proposal):
"The Std SelBack command restores the previous recorded Tree view selection. Note that selections are only recorded if Tree RecordSelection mode is switched on."

Shorter alternative:
"Restore the previous Tree view selection. Only works if Tree RecordSelection mode is switched on."

PaddleStroke:
I don't think tooltips need to be particularly short.
Having said that you are right in your example that the command name is not necessary.

src/Gui/CommandLink.cpp
wwmayer: > I thought about using sToolTipText  = "<p>" + QT_TR_NOOP("abc") + "</p>" but this is not accepted.

This isn't possible. One way is using a static std::string and once constructed assign it to the sToolTipText
```
static std::string toolTip = ...
sToolTipText = toolTip.c_str();
```

PaddleStroke: That still doesn't build. 'expression must have integral or unscoped enum type'
```
    static std::string toolTip = "<p>" + QT_TR_NOOP("A Link is a type of object that references or links"
        " to another object, in the same document, or in another document. Unlike Clones, Link"
        " references directly the original Shape, so it is more memory efficient which helps "
        "with the creation of complex assemblies from smaller subassemblies, and from multiple"
        " reusable components like screws, nuts, and similar fasteners.") + "</p>";
    sToolTipText  = toolTip.c_str();
```

wwmayer: Ah of course, because all three parts are C strings and they cannot be concatenated. You have to write
`static std::string toolTip = std::string("<p>") + QT_TR_NOOP("A Link .... ") + "</p>";`

PaddleStroke: Thanks, I moved the HTML out of the translation then.

Roy-043:
My suggestion for the other tooltips (fixing some language issues and shortening a few):

A Link is an object that references or links to another object in the same document, or in another document. Unlike Clones, Links reference the original Shape directly, making them more memory efficient which helps with the creation of complex assemblies.

A Part is a general purpose container to keep together a group of objects so that they act as a unit in the 3D view. It is meant to arrange objects that have a Part TopoShape, like Part Primitives, PartDesign Bodies, and other Parts."

A Group is a general purpose container to group different types of objects in the Tree view, regardless of their data type. It is a simple folder to organize the objects in a model.

PaddleStroke:
Sounds good to me. I updated them.

mwganson:
I prefer having the selection arrows on the toolbar.  Is there a preference to keep them there?

qewer33:
> I prefer having the selection arrows on the toolbar. Is there a preference to keep them there?

You can customize the toolbars from `Tools > Customize` and you can even create custom toolbars for functions that you use regularly.

PaddleStroke:
I have not added a preference to keep them in as no one of the forum (https://forum.freecad.org/viewtopic.php?t=81319) seem to even know what these were doing.
So it feels like very few people actually use them. I think that custom toolbars are specifically designed for such niche tools.

adrianinsaval:
> I have not added a preference to keep them in as :
> 
>     * custom toolbar let you do this very easily.
> 
>     * no one of the forum (https://forum.freecad.org/viewtopic.php?t=81319) seem to even know what these were doing.
>       So it feels like very few people actually use them. IMO custom toolbars are specifically designed for such niche tools.

Yes, let's not keep polluting FreeCAD with very niche config options that serve no real purpose than a minor convenience for a small subset of users.

wwmayer:
After each start of FreeCAD these error messages appear in the console:
```
<Command> Command.cpp(1828): duplicate command Std_SelBack
<Command> Command.cpp(1828): duplicate command Std_SelForward

The extra creation of StdCmdSelBack and StdCmdSelForward should be removed from CreateViewStdCommands()

PaddleStroke:
Ah yes sorry about this. I will fix it asap.

wwmayer:
I have a PR.

PR #10764, opened 2023-09-21
[Assembly] Solve the assembly (Card 9).

### LINKED ISSUE #10479: [Assembly] User can't move/rotate parts easily

### Full version info

```shell
Latest 0.22
```

### Subproject(s) affected?

### Problem description

In most CADs when you are working on an assembly, you can move parts of the assembly by clicking and dragging. This let you :
- easily position parts roughly at the beginning
- Move parts in a constrained assembly to make 'real' motions.

short video from OnShape (3:04), I think it will make it clearer what I mean : https://www.onshape.com/en/resource-center/videos/build-your-first-assembly

It is also usual to be able to rotate parts by holding ctrl while clicking and moving mouse.

Forum discussion :
https://forum.freecad.org/viewtopic.php?t=80773

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10749: [Assembly] Card 5 : Joint should be editable.

### Problem description

Following the introduction of https://github.com/FreeCAD/FreeCAD/pull/10611 :

Currently joints cannot be edited. Double clicking them should open back the joint taskbox to enable its modification.

### Full version info

```shell
0.22 after 10611
```

### Subproject(s) affected?

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10751: [Assembly] Card 7 : JCS snapping not ideal

### Problem description

Following the introduction of https://github.com/FreeCAD/FreeCAD/pull/10611 :

Currently you cannot put your JCS to the middle of a line. Or to a face center. Which would be useful.
Maybe other cases are also missing / not ideal.

### Full version info

```shell
0.22 after 10611
```

### Subproject(s) affected?

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10752: [Assembly] Card 8 : Joint Taskbox, can't flip and rotate JCS

### Problem description

Following the introduction of https://github.com/FreeCAD/FreeCAD/pull/10611 :

Now there is no way to reverse the direction of the JCS nor to rotate it.
There should be : 
- A flip button.
- A rotate spinbox where you can specify a rotation angle. Next to it a 'turn 90°' button for convenience.

### Full version info

```shell
0.22 after 10611
```

### Subproject(s) affected?

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10753: [Assembly] Card 9 : Joints do nothing.

### Problem description

Following the introduction of https://github.com/FreeCAD/FreeCAD/pull/10611 :

Joint are currently doing nothing.

So now comes the actual solving. We need a way to solve the assembly using the joints.

Proposal : Make an API to the open 'Ondsel Solver' and connect assembly wb to it.
The big question is how to do it. I was thinking that the solving of the assembly would happen in the recompute of the assembly object, the App::Part. So that would mean making a subclass of App::Part in assembly, called Assembly::Assembly I guess. And in this class define the recompute to solve the assembly from its joints.

### Full version info

```shell
0.22 after 10611
```

### Subproject(s) affected?

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10754: [Assembly] Card 10 : 'Fix part' missing

### Problem description

Following the introduction of https://github.com/FreeCAD/FreeCAD/pull/10611 :

We need a way to 'ground' or 'fix' parts.
This should perhaps take the form of a Joint as well. But with only 1 object obviously and no JCS required.

The question is, where do we put this. Some softwares use the mouse right click. I guess we could use a a toolbar entry. Perhaps in the joints toolbar, at the begining. And perhaps with a separator between it and other joints.

@qewer33 we need one more icon here :) and we need it to not be too confusing with the 'joint fixed' icon.

### Full version info

```shell
0.22 after 10611
```

### Subproject(s) affected?

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10815: [Problem] Assembly workbench should also recognize Part WB objects/solids

### Problem description

Tinkering with it I notice that it does not see Part workbench solids a linkable.

If the source file has a Body, it shows it available for linking.
If the source file has a Part (Std_Part), it shows it available for linking.
If the source file has a Part Cut, Extrude, etc., it is not available for linking.

I realize it is very early in development, but figured I'd point this out.

Posted about this in the forum: [https://forum.freecad.org/viewtopic.php?t=81311](url)

### Full version info

```shell
[code]
OS: Debian GNU/Linux 12 (bookworm) (XFCE/xfce)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34410 (Git)
Build type: Release
Branch: master
Hash: 836c237d3c40d925a5447d1f52841cbb9c25986f
Installed mods: 
  * MnesarcoUtils 0.2.2
  * ZolkoBase 0.0.1
  * lattice2 1.0.0
  * Curves 0.6.13
  * Nodes 0.1.36
  * Help 1.0.3
  * QuickMeasure 2022.10.28
  * Manipulator 1.5.0
  * Assembly
  * Assembly4 0.50.2
  * freecad.gears 1.0.0
  * sheetmetal 0.3.0
  * Silk 0.1.5
  * kicadStepUpMod 10.19.4
  * A2plus 0.4.60n
  * DynamicData 2.47.0
  * workfeature-macro
  * fasteners 0.4.66
  * ExtremeProDark 2.9.0
  * Defeaturing 1.2.0
  * CurvedShapes 1.0.4
[/code]
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #11311: [Problem] Assembly: Joint is not JSON serializable

### Problem description

When a document containing an assembly and joints are saved, the following error is reported:
```
PropertyPythonObject::toString(): failed for <class 'JointObject.Joint'>
  File "/usr/lib64/python3.11/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
<class 'TypeError'>: Object of type FeaturePython is not JSON serializable
```

When reopening the file, the joint viewprovider is not loaded / joint is not visible and greyed out in the treeview. The following error is reported:

```
Loading Assembly workbench...
Initializing Assembly workbench...
Assembly workbench loaded
  File "/home/howie/git/fc_build/Mod/Assembly/JointObject.py", line 276, in attach
    self.draw_style.lineWidth = self.axis_thickness
                                ^^^^^^^^^^^^^^^^^^^
<class 'AttributeError'>: 'ViewProviderJoint' object has no attribute 'axis_thickness'
  File "/home/howie/git/fc_build/Mod/Assembly/JointObject.py", line 383, in updateData
    self.switch_JCS1.whichChild = coin.SO_SWITCH_ALL
    ^^^^^^^^^^^^^^^^
<class 'AttributeError'>: 'ViewProviderJoint' object has no attribute 'switch_JCS1'
  File "/home/howie/git/fc_build/Mod/Assembly/JointObject.py", line 391, in updateData
    self.switch_JCS2.whichChild = coin.SO_SWITCH_ALL
    ^^^^^^^^^^^^^^^^
<class 'AttributeError'>: 'ViewProviderJoint' object has no attribute 'switch_JCS2'
  File "/home/howie/git/fc_build/Mod/Assembly/JointObject.py", line 427, in getIcon
    if self.app_obj.getPropertyByName("JointType") == "Fixed":
       ^^^^^^^^^^^^
<class 'AttributeError'>: 'ViewProviderJoint' object has no attribute 'app_obj'
```

### Full version info

```shell
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34935 (Git)
Build type: Release
Branch: main
Hash: e235a1b795567a99d67e1887a1c7151e60ebe395
Python 3.11.6, Qt 5.15.10, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Installed mods: 
  * Assembly4 0.50.4
  * QuickMeasure 2022.10.28
  * OpenDark 2023.10.30
```

### Subproject(s) affected?

Assembly

### Anything else?

@PaddleStroke 

### Code of Conduct

### LINKED ISSUE #11397: [Problem] Assembly: UI/UX lacks of feedback when trying to add links to an unsaved assembly

### Problem description

When trying to add an external link to an assembly without saving it first, it will appear to silently fail.

Steps to reproduce:
* Have the report view hidden (with the new shiny non-intrusive notifications)
* Create a new document and an assembly, do not save
* Press the Insert Link button in the toolbar, the task panel will open
* Select a part from an external file and hit OK
* The task panel will close with nothing having happened and no error reported to the user

Note: if you then manually enable the report view you can see that the WB did print out the fact that the assembly needs to be saved in order to use external links, but this is not visible when using non-intrusive notifications

### Full version info

```shell
OS: Ubuntu Core 22 (GNOME/gnome-xorg)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34988 (Git) Snap 832
Build type: Release
Branch: main
Hash: dfcfdad0400d5635cc03fcaa4849a2246bbd3e14
Python 3.10.12, Qt 5.15.10, Coin 4.0.0, Vtk 7.1.1, OCC 7.7.1
Installed mods: 
  * kicadStepUpMod 10.17.1
  * sheetmetal 0.2.63
  * LCInterlocking
  * fasteners 0.4.62
```

### Subproject(s) affected?

### Anything else?

I suggest at least elevating the notification to a warning or error so that it will appear when the report view is hidden. Perhaps some user friendlier way to remind the user to save could be implemented, perhaps including a prompt to save.

### Code of Conduct

### LINKED ISSUE #11606: [Feature request] Assembly: allow sketches in assembly container

### Problem description

It is possible to create datum elements in an assembly container, as it should. However, it is impossible to create nor move a sketch at the assembly level. Joints can snap to sketch elements so they are well recognized as legitimate references. It would be very useful to allow sketches in assembly to create skeletons.

When implemented, it would be necessary to put the `PartDesign_CreateSketch` command in the assembly toolbar. Its name would be obsolete as well.

### Full version info

```shell
OS: Ubuntu 23.04 (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35249 (Git) AppImage
Build type: Release
Branch: main
Hash: 94ca51b7799b2ac60a2a56d7e6e753bb97a73671
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.6, OCC 7.6.3
Locale: French/France (fr_FR)
Installed mods: 
  * Render 2023.8.18
  * OpenDark 2023.12.1
```

### Subproject(s) affected?

### Anything else?

_No response_

### Code of Conduct

PaddleStroke:
Adds AssemblyObject and viewer in cpp that handle the solving by connecting to ondsel solver.
Fixes #10753
Fixes #10749
Fixes #10754
Fixes #10479
Fixes #11311
Fixes #10751
Fixes #11397
Fixes #10815
Fixes #11606
Fixes #10752


wwmayer:
I can't give a qualified answer about whether these classes are needed or not as I don't know much about assemblies and their requirements.

But for the new code you should use the new copyright header and also listen to the warnings reported by Lint.

PaddleStroke:
Assembly are very similar to sketches actually. You have elements and constraints and you need to solve.

Yes for copyrights and warning, it was just an early draft to open discussion.

wwmayer:
> Assembly are very similar to sketches actually. You have elements and constraints and you need to solve.

I know what the assembly is good for but not much beyond that.

PaddleStroke:
So my next step is to be able to create an AssemblyObject in python :
`App.ActiveDocument.addObject("Assembly::AssemblyObject", "Assembly")`
But it is not working yet. My binding must be missing something but I can't find what. Any idea ?

PaddleStroke:
> AssemblyObjectPy.cpp(20,25): error C2871: 'Assembly': a namespace with this name does not exist 

@wwmayer any idea what is wrong with the namespace ? I can't understand what is happening. In visual studio when I follow the error to the generated AssemblyObjectPy.cpp file, the namespace is error-highlighted, then after 2 seconds it becomes ok as if the error is resolved. I tried to follow PartPy example and can't find any difference.

FlachyJoe:
Another way could be to delegate the solve to a App::Part child (a solver object) who get the joints by parsing it's parent. The FEM solver is working as this.

wwmayer:
```
// clang-format off
// inclusion of the generated files (generated out of AssemblyObject.xml)
#include "AssemblyObjectPy.h"
#include "AssemblyObjectPy.cpp"
// clang-format on
```

wwmayer:
FYI, the workaround with clang-format off/on is not needed any more because the sorting of header file has been disabled: https://github.com/FreeCAD/FreeCAD/commit/b65ae56f1faa4f3d9390915178beb0bebe99a3f1

PaddleStroke:
Ah thanks ! 
I see now clang was reversing the .h and .cpp. Thanks very much. I'll continue on it now.

PaddleStroke:
I am still unable to create the object in python. Any idea what is missing here ?
Is there a template about how to add a DocumentObject class somewhere ?

wwmayer:
What are you trying to do?

PaddleStroke:
So I want to make a subclass of App::Part called Assembly::AssemblyObject.
This subclass will have a 'solve' function that will solve the assembly based on the joint objects.

So now my goal is to already make a dummy subclass of App::Part and be able to create this object from python : 
`App.ActiveDocument.addObject("Assembly::AssemblyObject", "Assembly")`

PaddleStroke:
Thanks for your feedback! I did those steps, though it still doesn't work. Perhaps I am not using the name AssemblyApp at the correct places ?

wwmayer:
The AssemblyObject is created now but what's missing is the creation of its view provider. Currently AssemblyGui is a normal shared library and in order to make it a Python extension module it must define the entry point PyInit_AssemblyGui. And the type ViewProviderAssembly must be registered.
https://github.com/wwmayer/FreeCAD/commit/7d7bb9056c73c3b7483abb43a41062569d3e7954

PaddleStroke:
Thanks very much, this is working properly now.
One other question. I need a getter to get the list of Joints of the assembly.
Joints are currently App::FeaturePython objects.

To get them : 

```
std::vector<App::FeaturePython*> AssemblyObject::getJoints()
{
    std::vector<App::FeaturePython*> joints = {};

    App::Document* doc = getDocument();
    App::DocumentObject* jointsGroup = doc->getObject("Joints");

    if (jointsGroup && jointsGroup->isDerivedFrom(App::DocumentObjectGroup::getClassTypeId())) {
        for (auto* obj : static_cast<App::DocumentObjectGroup*>(jointsGroup)->getObjects()) {

            if (obj /* && obj->isDerivedFrom(App::FeaturePython::getClassTypeId())*/) {
                joints.push_back(static_cast<App::FeaturePython*>(obj));
            }
        }
    }

    return joints;
}
```

However I am not sure how I can test to make sure its a Joint and not another random object that ended up in the joints group.
`obj->isDerivedFrom(App::FeaturePython::getClassTypeId())` does not compile, besides it would not tell if its a Joint or another random FeaturePython.

Any recommandation about the best way to handle things here ? Thanks

wwmayer:
```
#include <App/PropertyPythonObject.h>
#include <Base/Interpreter.h>

std::vector<App::DocumentObject*> Assembly::getJoints()
{
    std::vector<App::DocumentObject*> joints = {};

    App::Document* doc = getDocument();
    App::DocumentObject* jointsGroup = doc->getObject("Joints");

    Base::PyGILStateLocker lock;
    if (jointsGroup && jointsGroup->isDerivedFrom(App::DocumentObjectGroup::getClassTypeId())) {
        for (auto* obj : static_cast<App::DocumentObjectGroup*>(jointsGroup)->getObjects()) {
            App::PropertyPythonObject* proxy = obj ? dynamic_cast<App::PropertyPythonObject*>(obj->getPropertyByName("Proxy")) : nullptr;
            if (proxy) {
                Py::Object joint = proxy->getValue();
                if (joint.hasAttr("setJointConnectors")) {
                    joints.push_back(obj);
                }
            }
        }
    }

    return joints;
}
```

PaddleStroke:
@chennes What is the standard procedure to add a library to freecad ? I am now looking into how to include the OndselSolver and  I'm not sure how to proceed.
So now I have a OndselSolver.lib and I created a file SetupLibOndselSolver.cmake. What are the next steps?

wwmayer:
https://github.com/Ondsel-Development/OndselSolver/blob/main/LICENSE is GPLv2 only. 
Its usage in FreeCAD is not allowed [1] because it conflicts with the QSint library [2] that is LGPLv3 only.

[1] https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility
[2] https://marketplace.qt.io/products/qsint

PaddleStroke:
I think we will find a solution to this @sliptonic ?
Meanwhile my goal is to get things ready so what I want to do is mostly to add the library locally to test things out.

sliptonic:
> OndselSolver is GPLv2 only [1].

Should be LGPL.   I'll investigate and revise

wwmayer:
> Should be LGPL. I'll investigate and revise

Oops, it is LGPL. I missed the word "Library" in the license terms. Sorry for the noise!

wwmayer:
> So now I have a OndselSolver.lib and I created a file SetupLibOndselSolver.cmake. What are the next steps?

What about adding it as git sub-module?

sliptonic:
> Oops, it is LGPL. I missed the word "Library" in the license terms. Sorry for the noise!
> The old version (from 1991) of the license text use the word "library" while the newer version (from 1999) use the word "lesser" but both versions have the same meaning.

We'll still update to use the newer license boilerplate

chennes:
We currently have two "standard" procedures, and as @wwmayer points out, there is a third option that we don't use for anything but should consider here, because there are significant problems with trying to use those other two methods.

Our "normal" way of adding a dependency that can be provided by system package managers is to just start depending on it (for example cpp-yaml in Materials) and add it to our list of required libraries. We add it ourselves to Conda, Snap, Appimage, and LibPack builds, and rely on packagers and self-compilers to handle it on their own. This really only works for packages that are provided by system package managers, so Ondsel will have to undertake an effort in that direction before we can really even begin. We took a half-way approach with libfmt, where we first try system packages, and if that fails use the cMake fetch-content methods to clone the appropriate repo. Packagers weren't super happy with this, though, they don't like the network access during cMake.

Our second common method is to literally copy the library into our source tree: see e.g. zipios. that's great for libraries that are either unlikely to change in any significant way or have been abandoned, but isn't really suitable for libraries that are in active development.

I am in favor of Werner's suggestion that we use a git-submodule in this case. It will require FreeCAD developers to jump through one more hoop when cloning, but gives us complete separation between the library and the rest of our code, and also a very easy channel for updating to new versions when it's appropriate to do so.

PaddleStroke:
So I managed to include my .lib file locally first and I'm now debugging.

I have a c++ error that is driving me in circles, perhaps you know how to handle it @wwmayer ?
I'm using a `std::unordered_map<App::DocumentObject*, std::shared_ptr<MbD::ASMTPart>>` in AssemblyObject.
And I have an error when I try to iterate through the elements in the function `setNewPlacements ` :
`    for (auto& pair : objectPartMap) {`
This gives me a 'error invalid map<K, T> key'.

I saw a suggestion that it needs to define a hash for App::DocumentObject*, but my attempts so far are not working. For reference : 

```
    struct DocumentObjectHash
    {
        std::size_t operator()(const App::DocumentObject* obj) const
        {
            std::size_t hash = 0;

            std::string name = obj->getFullName();
            hash = std::hash<std::string> {}(name);

            return hash;
        }
    };

    std::unordered_map<App::DocumentObject*, std::shared_ptr<MbD::ASMTPart>, DocumentObjectHash> objectPartMap;
```

Any idea how I can iterate through my list ?
Thanks

wwmayer:
> I have a c++ error that is driving me in circles,

Based on the information you give me I have written this small test application:
```
#include <iostream>
#include <memory>
#include <string>
#include <unordered_map>

class DocumentObject
{
public:
    DocumentObject(const std::string& name) : name(name) {}
    std::string name;
};

class ASMTPart
{
public:
    ASMTPart(const std::string& part) : part(part) {}
    std::string part;
};

int main()
{

    std::unordered_map<DocumentObject*, std::shared_ptr<ASMTPart>> objectMap;
    objectMap[new DocumentObject("obj1")] = std::make_shared<ASMTPart>("part1");
    objectMap[new DocumentObject("obj2")] = std::make_shared<ASMTPart>("part2");
    for (auto& it : objectMap) {
        std::cout << it.first->name << ":" << it.second->part << '\n';
    }
}
```

which compiles and works fine. I guess the problem must be elsewhere.

PaddleStroke:
Ok so it is now mostly working ! 
Preview video : https://youtu.be/tkGzMTDHD_E
There are a few quirks to fix, such as the grounded part jumping to origin. But I think it can be started to be reviewed. Quirks will be solved before review is done considering the size.

Here's the .lib file and includes for testing. it's a bit huge so we need to figure out why:
https://www.dropbox.com/t/KLPopcX1vODVpsp1

sliptonic:
Moving back to draft pending the inclusion of the Ondsel Solver as a dependency

PaddleStroke:
Rebased on #11363

sliptonic:
Testing.   Config is based on current tip of  this PR with the solver version 84762ac

Start with a new file.  Create or import some solids
Create an assembly.  (It's ACTIVE by default)
Drag parts into assembly.
At this point, I cannot drag any parts.  I cannot toggle the assembly INACTIVE.
Save the file.  Close. Reopen
The Assembly is INACTIVE.  Toggling Active works.  Dragging Parts works. Toggling INACTIVE works.

PaddleStroke:
I can reproduce the same behavior. You don't actually have to close the file. For example creating a second assembly also solves the bug. I will investigate.

PaddleStroke:
@wwmayer there is one issue that I am not sure how to handle :

- Objects within the AssemblyObject cannot be deleted when the Assembly is activated/inEdit (assembly obj is inEdit when activated, such that we trigger the assembly VP so that the part dragging works). 
- When it is not in edit, deletion works OK
- The same happen with an App::Part, if you put it in edit by using transform. Then you cannot delete things inside.

I have been looking at `StdCmdDelete::activated`. So when in edit the block `if(vpedit)` is used.

So here I am not sure what is the best way to solve this. I see two possibilities : 
- override onDelete in VPAssemblyObject and handle deletion from the subnames.
- add a function :
```
ViewProviderDocumentObject::shouldActAsIfNotInEditOnDeletion { return false; }
ViewProviderAssemblyObject::shouldActAsIfNotInEditOnDeletion override { return true; }
```
that would prevent going into the block `if(vpedit)` by changing the condition to : 
`if(vpedit && vpedit->shouldActAsIfNotInEditOnDeletion())`

Or perhaps you have a better way in mind? Thanks

chennes:
Try getting rid of the byte-order marker.

chennes:
It still doesn't like the BOM: `1 invalid non-printable character U+FEFF` I suspect it will work fine when you remove that.

PaddleStroke:
Thanks @chennes I removed the BOM let's see if it's ok now !

pierreporte:
Just a question. There are more and more issues fixed, while cards were designed to be pretty much standalone, that it seems it will end up implementing more or less the whole thing at once. Will it be split again later? What’s the ETA for this PR to enter the review process?

sliptonic:
> What’s the ETA for this PR to enter the review process?

No ETA.  It's done when it's done.

PaddleStroke:
> Just a question. There are more and more issues fixed, while cards were designed to be pretty much standalone, that it seems it will end up implementing more or less the whole thing at once. Will it be split again later? What’s the ETA for this PR to enter the review process?

Yes there are more and more issues fixed, but there are also more and more issues introduced :)
I think the code likely need a little bit of time to be 'finished' and stabilized. For now you can try it on the assembly playground build that ondsel released, and maybe help reporting the bugs and weird behaviors on the discord channel.

src/Mod/Assembly/CommandSolveAssembly.py
```diff
@@ -0,0 +1,76 @@
...
+#    FreeCAD is distributed in the hope that it will be useful, but         *
+#    WITHOUT ANY WARRANTY; without even the implied warranty of             *
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU       *
+#    Lesser General Public License for more details.                        *
+#                                                                           *
+#    You should have received a copy of the GNU Lesser General Public       *
+#    License along with FreeCAD. If not, see                                *
+#    <https://www.gnu.org/licenses/>.                                       *
+#                                                                           *
+# ***************************************************************************/
```
luzpaz, line 22: The uniformity of the ascii box characters isn't correct. See line 2 and 22

Same for https://github.com/FreeCAD/FreeCAD/pull/10764/files#diff-08d614cbb4fb3a9554bc57f0639d31c6a5a5fcc83621114dd06bb9d7600a9ff5L2-R22

src/Mod/Assembly/Gui/ViewProviderAssemblyPyImp.cpp
```diff
@@ -0,0 +1,59 @@
...
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
```
luzpaz, line 21: forward slash should be under the above asterisk

src/Mod/Assembly/CommandSolveAssembly.py
luzpaz: Same for https://github.com/FreeCAD/FreeCAD/pull/10764/files#diff-ec35d691aafdacda9b015fd00e530d861d07a967356d4c57fadea06448bfb4e0R2-R22

PaddleStroke:
@chennes I think the PR can now be reviewed. It is a large chunk so let me know how I can make things easier.
- Do you want me to walk you through the code before we do so at a merge meeting?
- Alternatively we can check this out during the hackathon.
- Do you want me to squash all the commits into one? Now it's a bit big with 45 commits. So I will already squash meaningless commits.
- Anything else?

chennes:
No tests? For example, `GetAngleOriented` is an obvious candidate, just starting from the top.

PaddleStroke:
Yes tests are needed. I have not writen any tests yet so it will be my first.
But for a function like GetAngleOriented we need a test? It is a basic geometric function, it is not like it can start failing no? I will look into it soon

PaddleStroke:
Ok so I started to look into it, and indeed Vector3d GetAngle does have a test:

```
TEST(Vector, TestAngle)
{
    Base::Vector3d vec1(0, 0, 0.000001);
    Base::Vector3d vec2(0, 0.000001, 0);
    double angle = vec1.GetAngle(vec2);
    EXPECT_EQ(angle, Base::float_traits<double>::pi() / 2);
}
```
I'm highly skeptical about the usefulness of such tests, but I'm not going to rant for longer than it will take me to write a similar one.

PaddleStroke:
Hi @Nemesis81 , I have seen your video already. The weird behaviors are because the joints are creating a 'loop' so the dragging goes a bit crazy. Also some joints (cylindrical and many distance) don't have special drag modes which makes the dragging not so good. It's on my todo list.

PaddleStroke:
@chennes rebased & conflcit fixed

luisDias89:
Nice Workbench: FreeCAD has been waiting for this feature for a long time.
Please, whoever is developing it, try to make it easy and intuitive, like SolidWorks, for example.
When you select two faces, two disks, edges, or other object items, it will analyze all the possibilities that could be achieved with them and will automatically provide a menu. Initially, the menu will be semi-transparent. If the mouse cursor is moved over the menu, it becomes available; if the mouse moves away, the menu disappears.
Some Features for example for planing is make coincident, make paralel, distance betwen a range, distance only, lock, perpedicular, range of angle betwen boot, or fixed angle.

Look at this image as an example:
image

Other example is two circles: 
image
We have concentric, tangent, lock...

And so on, another example: among gears, you can define when one gear rotates in relation to another, making animations super realistic, with the teeth rotating perfectly meshed.

And many other options that can be integrated.

This is the most importante workbench for large assembly, this is what the freecad need to bee a great Software.

luisDias89:
Exemple of Gear Assembly
https://www.youtube.com/watch?v=JcOO6jcdZ_I

Lock at the mouse, we can able to move object with mouse and see in real time one animation

Issue #5603, opened 2022-02-07
[Problem] Arch: Working plane not reset under Arch when making new project


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=1823

* **Reporter:** lanciaux
* **Date submitted:** 11/10/2014
* **FreeCAD version:** 
* **Category:** Bug
* **Status:** assigned
* **Tags:** python

# Original report text

Current working plane settings are not reset after you create a new Project/Sheet.

# Steps to reproduce

1. Make anything you want
2. Go to Arch
3. Setup a plane setting by selecting any face of your model
4. Close this Project/Sheet without saving it
5. Create a new Project/Sheet

# Other bug information

* **Priority:** low
* **Severity:** minor
* **Category:** Bug
* **Updated:** 9/21/2021

# Discussion from Mantis ticket

### Comment by @yorikvanhavre 2014-11-10 23:55
Hm these are two independent things... I'm not sure if it would be a good thing or not to reset the work plane when opening a new model. You might have set it a certain way in another file, that might be convenient to keep it.

Besides, at the moment there is no mechanism to detect when a new document is opened from python, so this feature cannot be imlemented right now.

But let's leave it open, an idea could arise.

---

### Comment by @luzpaz 2017-06-03 15:47
yorik can you weigh in on this again?

---

### Comment by @yorikvanhavre 2017-06-05 15:46
Still no idea on how to detect when a new document has been created/opened :(

---

### Comment by @luzpaz 2017-10-18 11:03
[quote=yorik]Besides, at the moment there is no mechanism to detect when a new document is opened from python, so this feature cannot be imlemented right now.
[/quote]
wmayer care to weigh in?

---

### Comment by @yorikvanhavre 2018-12-05 12:58
This is now possible with the DocumentObserver. But since it can be disruptive, it's best to implement after the 0.18 release.

---

### Comment by @luzpaz 2020-12-08 01:41
Forum thread: https://forum.freecadweb.org/viewtopic.php?f=23&t=53021

---

### Comment by @yorikvanhavre 2021-09-21 08:56
***New proposal:*** implementing per-document working plane

*  The grid system is already per view, so each view can display a different grid already.
*  The python interface should stays the same. You just get and set the working plane and shouldn't worry about anything else
[...]

PR #10766, opened 2023-09-21
Draft: Introduction of the PlaneBase class

Roy-043:
This PR introduces the new PlaneBase class.

Most of its functions are rewritten versions of the functions in the original Plane class, with the exception of align_to_point_and_axis_svg.

Note that after this PR the original Plane class does not yet inherit from PlaneBase. This will be done in several following PRs. I will then also explain the reasons behind some of the revised functions in PlaneBase.

The introduction of the PlaneGui class will follow after that.

Issue: #5603.


yorikvanhavre:
Sorry for the delay. This seems perfect to me and the right way to go (rewrite it all then test then migrate)

Issue #10746, opened 2023-09-20
[Assembly] Card 2 : JCS should scale with zoom.


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Following the introduction of https://github.com/FreeCAD/FreeCAD/pull/10611 : 

Currently the JCS size is defined when the tool is launched. But it does not update if you zoom in or out. It would probably be best to have them to scale with zoom.

Note about this: we can't use the mouse wheel event as zoom can be achieved by other means. So it needs to connect to the camera change event. Not sure how this is done in python.

### Full version info

```shell
0.22 after #10611
```

### Subproject(s) affected?

Other (specify in description)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
Note : @qewer33 is currently attempting this one.

Issue #10745, opened 2023-09-20
[Assembly] Card 1 : JCS are ugly


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Following the introduction of https://github.com/FreeCAD/FreeCAD/pull/10611 : 
The created JCS are fairly ugly.
@qewer33 the mockup you posted seems good to me. Do you want to attempt to implement it in coin or do you want me to jump in ?
Mockup : ![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/8209d99d-2387-4eb4-96f4-a56b9909d5cf)

### Full version info

```shell
0.22 after https://github.com/FreeCAD/FreeCAD/pull/10611
```

### Subproject(s) affected?

Other (specify in description)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

qewer33:
> Following the introduction of #10611 : The created JCS are fairly ugly. @qewer33 the mockup you posted seems good to me. Do you want to attempt to implement it in coin or do you want me to jump in ?

I can give it a shot but I need some pointers. What classes/files do I need to modify, or do I need to create a brand new class? Python or C++?

Also here are new mockups (with a circle alternative) with the plane on the center:

![image](https://github.com/FreeCAD/FreeCAD/assets/69015181/f6703f14-49ea-4433-b520-3ee763d8142e)

PaddleStroke:
It's python.
The coin code is in the JointObject.py ViewProviderJoint class

attach() is the function that creates the coin objects. and assign it to the display mode. In this function you can see : 
`self.switch_JCS1 = self.JCS_sep(obj, self.transform1)`
This is one joint coordinate system. So you can see its made in JCS_sep function
This function creates JCS = coin.SoAnnotation() this is basically the JCS coin object. You can see that the axis get added to this.
The axis themselves are create in the line_sep function.

The line thickness is a property, so it is actually defined in __init__.
The line lengths are applied by the self.axisScale. So its determined in get_JCS_size. Basically if you want to make them shorter you need to change this line : camera.height.getValue() / 10.
replace 10 by 20 and length will halve.
For the rectangle, you will need to create a soFaceSet (I seem to recall that's the name) And you will want to create it in a similar fashion as the lines are created.

howie-j:
The circle one looks nice :100: 
It will also help visually to make sure the JCS is placed correctly, e.g in the center of a hole.

[...]

PR #10767, opened 2023-09-21
Assembly: Improve JCS appearance and implement autoscale

qewer33:
Completes https://github.com/FreeCAD/FreeCAD/issues/10745 and https://github.com/FreeCAD/FreeCAD/issues/10746

CC @PaddleStroke 

Preview video:

[vokoscreenNG-2023-09-21_19-30-07.webm](https://github.com/FreeCAD/FreeCAD/assets/69015181/603f1a3e-a20b-4d7f-ad4c-c64695311c3f)

### DIFF
--- a/src/Mod/Assembly/JointObject.py
+++ b/src/Mod/Assembly/JointObject.py
@@ -21,6 +21,8 @@
 #                                                                           *
 # ***************************************************************************/
 
+import math
+
 import FreeCAD as App
 import Part
 
@@ -256,13 +258,17 @@ class ViewProviderJoint:
         obj.color_X_axis = (1.0, 0.0, 0.0)
         obj.color_Y_axis = (0.0, 1.0, 0.0)
         obj.color_Z_axis = (0.0, 0.0, 1.0)
-        obj.axis_thickness = 2
+        obj.axis_thickness = 3
         self.x_axis_so_color = coin.SoBaseColor()
         self.x_axis_so_color.rgb.setValue(1.0, 0.0, 0.0)
         self.y_axis_so_color = coin.SoBaseColor()
         self.y_axis_so_color.rgb.setValue(0.0, 1.0, 0.0)
         self.z_axis_so_color = coin.SoBaseColor()
         self.z_axis_so_color.rgb.setValue(0.0, 0.0, 1.0)
+
+        camera = Gui.ActiveDocument.ActiveView.getCameraNode()
+        self.cameraSensor = coin.SoFieldSensor(self.camera_callback, camera)
+        self.cameraSensor.attach(camera.height)
 
         self.app_obj = app_obj
         obj.Proxy = self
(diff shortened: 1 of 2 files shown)


qewer33:
ok uhhh... what?

I have no idea how @chennes' TNP fixes ended up in my commit :joy: 

really weird,I think it might be caused by the pre-commit, I'll see how I can fix it

PaddleStroke:
Awesome work !
Maybe you did a rebase interactive with 1 commit too much? If so you can redo the rebase and drop chris commit.

qewer33:
> Looks like its squashed with a commit from chris ?

I didn't do any squashing or rebasing, that's why I am confused. Should be all good now though!

PaddleStroke:
Yes it looks better now.

Any other card holds your interest perhaps ? With such efficiency we'll be through them quickly😄

qewer33:
> Any other card holds your interest perhaps ? With such efficiency we'll be through them quickly😄

I'll see what I can do :sweat_smile:

luzpaz:
@qewer33 please `git commit --amend` the typo:  appereance - > appearance

tobiasfalk:
It would be nice if there were a seting for the size of JCS, if there is not already one.

PaddleStroke:
Another small thing regarding the JCS appearance : 
I initially put the axis colors as properties. But this is a mistake, there no point to have them as properties. The axis colors should be preferences instead.

 If you want to tackle that as well in this PR? Or it can come later.

qewer33:
> Another small thing regarding the JCS appearance : I initially put the axis colors as properties. But this is a mistake, there no point to have them as properties. The axis colors should be preferences instead.
> 
> If you want to tackle that as well in this PR? Or it can come later.

I used parameters as axis colors on my recent transform dragger improvements PR. The idea is to unify all the axis colors across FreeCAD (axis cross, corner coord system, transform dragger, navi cube axis's etc.) and make them configurable from Preferences. I'll open an issue and a PR for this later.

For now I can JCSs also use these parameters.

PaddleStroke:
Yes it needs to be removed. It's an inaccuracy.

qewer33:
@chennes ready for review/merge now

wwmayer:
Some remarks:
* In order to avoid possible problems with scripting you should consider moving the view provider code into a separate file
* In the view provider you shouldn't add a callback function that is linked to the camera of the active view because this is error prone. If you open a new 3D view from the View menu then auto-scaling doesn't work.

Instead you should use the class SoShapeScale:
```
from pivy import coin

cube = coin.SoCube()

type = coin.SoType.fromName("SoShapeScale")
scale = type.createInstance()
scale.setPart("shape", cube)

graph = Gui.ActiveDocument.ActiveView.getSceneGraph()
graph.addChild(scale)
scale.scaleFactor = 10
```

PR #10769, opened 2023-09-21
App: fixes #10460: App::PropertyPythonObject is not saving data

### LINKED ISSUE #10460: [Problem] App::PropertyPythonObject is not saving data

wwmayer:
For serialization of Python features the methods `__getstate__/__setstate__` are used optionally. With Py3.11 these methods have been added to the `object` class so that they exist for each Python class. But this way the class App::PropertyPythonObject will always call these methods when it shouldn't.

### DIFF
--- a/src/App/PropertyPythonObject.cpp
+++ b/src/App/PropertyPythonObject.cpp
@@ -87,9 +87,9 @@ std::string PropertyPythonObject::toString() const
             throw Py::Exception();
         Py::Callable method(pickle.getAttr(std::string("dumps")));
         Py::Object dump;
-        if (this->object.hasAttr("__getstate__")) {
+        if (this->object.hasAttr("dumps")) {
             Py::Tuple args;
-            Py::Callable state(this->object.getAttr("__getstate__"));
+            Py::Callable state(this->object.getAttr("dumps"));
             dump = state.apply(args);
         }
         else if (this->object.hasAttr("__dict__")) {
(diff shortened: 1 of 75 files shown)


chennes:
If we're going to move away from the  `__getstate__` naming, does it still make sense to write this as a dunder method? Could it just be named as a normal method instead, so it doesn't look like a mistake and we meant to use `__getstate__`?

wwmayer:
At the beginning we used Python's pickle module to save and restore data and this required to implement `__getstate__ & __setstate__`. But as pickle is dangerous because you can execute arbitrary code we moved to the JSON module but left the method names. Which method names do you suggest? Maybe `dumps` and `loads` as used by the JSON module?

wwmayer:
OK, I will re-write the PR then.

adrianinsaval:
how severe was this issue? should we attempt to backport this? when cherry picking the only conflict I actually get is on an Assembly file which is to be expected since it doesn't exist. So I think we could

wwmayer:
This is the bug report: https://github.com/FreeCAD/FreeCAD/issues/10460

The bug is worth to be backported because when linking FreeCAD with Py3.11 there is a naming clash with `__setstate__ and __getstate__` so that FreeCAD fails to restore parts of a project.

PR #10771, opened 2023-09-22
Fem: Apply clang-format

wwmayer:


chennes:
It looks like in FEM, pre-commit is catching some string issues in FEM.ts (not just the translated files). I'll clean those up in a follow-up PR.

PR #10781, opened 2023-09-22
Change default file saving preferences

dyylanhammond:
This PR:
 - Increases default compression level from 3 to 7
 - Increases default thumbnail size from 128px to 256px
 - Decreases autosave interval from 15 minutes to 5 minutes

In my testing, increasing compression level didn't significantly increase saving time, even on low-end machines, and compression performance plateaus at 7.
image

Increasing thumbnail resolution significantly increases quality even when displayed at a lower resolution.
image

### DIFF
--- a/src/App/Document.cpp
+++ b/src/App/Document.cpp
@@ -1895,7 +1895,7 @@ bool Document::saveToFile(const char* filename) const
     signalStartSave(*this, filename);
 
     auto hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Document");
-    int compression = hGrp->GetInt("CompressionLevel",3);
+    int compression = hGrp->GetInt("CompressionLevel",7);
     compression = Base::clamp<int>(compression, Z_NO_COMPRESSION, Z_BEST_COMPRESSION);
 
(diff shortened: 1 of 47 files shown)


chennes:
Thanks for the PR -- is there an Issue where these have been discussed? The first three seem pretty non-controversial, but I hesitate to change the autosave frequency without some sense of consensus on the matter.

dyylanhammond:
I didn't open an issue to discuss these changes. I don't see any scenario where increasing the autosave interval could cause issues. If necessary, this can be delayed for further discussion.

chennes:
Let's take out the commits that change the autosave interval, and the one that modifies all the TS files (those are autogenerated, you don't need to fix them by hand). Then we can merge what's left, and start a discussion about the autosave interval. Thanks!

dyylanhammond:
Files that are autogenerated should have a comment saying so.

chennes:
> Files that are autogenerated should have a comment saying so.

I agree - unfortunately the tool that creates them (Qt's `update`) doesn't add those, so we'll have to modify the tool we use to call `lupdate`, src/tools/updatets.py, to read in and modify those files. Please consider creating an Issue here on GitHub with this suggestion.

PR #10790, opened 2023-09-23
Gui: Add rotation center indicator color, transparency and size preferences

Rexbas:
The rotation center indicator is now customizable. The color, transparency and size can be changed in the preferences.


yorikvanhavre:
It seems this PR is removing the "show rotation center" checkbox from preferences. Is that intentional?

Rexbas:
Yes that is intentional, you can now disable it with the new checkbox.
image

yorikvanhavre:
Great, thanks! Let's merge then :)

Issue #10791, opened 2023-09-24
[Problem] Exception in LeadInOutDressup due to floating point inaccuracy


jffmichi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

The LeadInOutDressup throws an IndexError exception on line 212 due to "queue" only containing one element and trying to access the second element. This ins mainly caused by a floating point inaccuracy when comparing the z-position of the last and the current command on line 549. A file that demonstrates the issue will be attached to this report and a pull request ist on the way.

### Full version info

```shell
[code]
OS: Arch Linux (KDE/plasmawayland)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34275 (Git)
Build type: None
Branch: master
Hash: 222a2520b1eeae9ecd04affab91dc71dbee0453a
Python 3.11.5, Qt 5.15.10, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * better-tool-library 0.9.0
  * dodo 1.0.0
  * sheetmetal 0.2.63
  * DynamicData 2.46.0
  * fcgear 1.0.0
  * fasteners 0.4.58
  * fcgear.backup1663629032.4145591 1.0.0 (Disabled)
  * Assembly4 0.50.2
  * lattice2 1.0.0
[/code]
```

### Subproject(s) affected?

Path

### Anything else?

[leadinout_exception.zip](https://github.com/FreeCAD/FreeCAD/files/12707560/leadinout_exception.zip)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10792, opened 2023-09-24
Path: fix LeadInOutDressup throwing an exception due to floating poin…

jffmichi:
…t inaccuracy when comparing z-positions

This PR fixes issue #10791.

"Path.Geom.isRoughly" is used to make sure the current command moves down by more than the floating point inaccuracy and only considers "curCommand" to be a layer change if that is the case. This prevents single element "queue"s to be added to "layers" in the case demonstrated in the attached file in the issue.


luzpaz:
Should we consider backporting this as well ?

PR #10796, opened 2023-09-24
[Release 0.21] pre-commit: Avoid automatic commits for backport fixes

wwmayer:
For backport fixes the automatic changes by pre-commit should be avoided as this leads to hundreds of modified files.


luzpaz (approved):

chennes:
We could also add
```
ci:
    autofix_prs: false
```

PR #10799, opened 2023-09-24
FEM naming fix part 2 (removes remaining incorrect "constraint" usage)

### LINKED ISSUE #10786: [Problem] FEM naming fix part 2

### Problem description

After resolving #10134 with the help of @kaktusus, there are still a few names with "constraint" left where they shouldn't be present. I will try to clean this up. On this occasion, I will also consider removing "FEM" from some names where it's unnecessary.

### Full version info

```shell
[code]
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34367 (Git)
Build type: Release
Branch: master
Hash: e36e16de59dece2d67caf478d25e3a6b825102c6
Locale: Polish/Poland (pl_PL)
[/code]
```

### Subproject(s) affected?

FEM

### Anything else?

_No response_

### Code of Conduct

FEA-eng:
Follow-up on #10519, resolves #10786 - cleaning the remaining names that shouldn't use "constraint"

### DIFF
--- a/src/Mod/Fem/Gui/Command.cpp
+++ b/src/Mod/Fem/Gui/Command.cpp
@@ -275,7 +275,7 @@ CmdFemConstraintDisplacement::CmdFemConstraintDisplacement()
 {
     sAppModule = "Fem";
     sGroup = QT_TR_NOOP("Fem");
-    sMenuText = QT_TR_NOOP("Constraint displacement");
+    sMenuText = QT_TR_NOOP("Displacement boundary condition");
     sToolTipText =
         QT_TR_NOOP("Creates a displacement boundary condition for a geometric entity");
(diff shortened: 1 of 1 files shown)


chennes:
@wwmayer do you want to merge the formatting PR first, then rebase this one, or the reverse?

wwmayer:
In the end it doesn't matter because in both ways there will be merge conflicts. But it's probably less work to first merge this PR as otherwise I had to create a new PR.

Also this gives me the opportunity to rework my PR a bit, so that the changes of the resource files are in a separate commit. This makes it a bit easier to adjust the source strings where the superfluous whitespaces must be removed.

PR #10800, opened 2023-09-24
[Path] Sanity Follow up improvements...

Syres916:
... from https://github.com/FreeCAD/FreeCAD/pull/10654 and removal of remaining asciidoc generating code

Feedback https://forum.freecad.org/viewtopic.php?p=708324#p708324


Syres916:
Had some further feedback from the forum so more testing required.

PR #10805, opened 2023-09-25
Std_Transparency : Toggle transparency of objects.

### LINKED ISSUE #10770: [Feature request] Conveniently streamlined Transparency toggle

### Problem description

It would be nice to have a transparency toggle, when activated it switch a part between 0 and 63 transparency (63 is a 'good' value).
This way it would enable user to toggle transparency easily, without having to right click, appearance ... Which is too long if someone just want to quickly check what is behind a part.

Forum discuss : https://forum.freecad.org/viewtopic.php?t=81275

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Anything else?

_No response_

### Code of Conduct

PaddleStroke:
Add a tool to toggle transparency.
image

It works with single solids. But also with App::Parts containing links to multiple files.
Fixes #10770
Forum discuss : https://forum.freecad.org/viewtopic.php?t=81275

### DIFF
--- a/src/Gui/CommandView.cpp
+++ b/src/Gui/CommandView.cpp
@@ -48,6 +48,8 @@
 #include <App/DocumentObject.h>
 #include <App/GeoFeature.h>
 #include <App/GeoFeatureGroupExtension.h>
+#include <App/Part.h>
+#include <App/Link.h>
 #include <App/MeasureDistance.h>
 #include <Base/Console.h>
 #include <Base/Parameter.h>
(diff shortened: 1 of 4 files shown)


Roy-043:
Related issue: #10770.

Is this really a toggle?
Let's assume the user has a number of objects with different transparencies. Will this implematation allow the user to return to the original situation?

Reading the code I get the idea that this is not the case, and IMO this therefore needs some more thought.

PaddleStroke:
> Is this really a toggle? Let's assume the user has a number of objects (in a Std_Part) with different transparencies. Will this implementation allow the user to return to the original situation?
> 
> Reading the code I get the idea that this is not the case, and IMO this therefore needs some more thought. Maybe a proper transparency override should be the aim.

This is a real toggle, so there is no way to go back to original situation. So for example you have 2 transparent and 2 not-transparent solids, if you select them all and toggle they will all go to not-transparent, then toggle again they go to transparent. But if you want to go back to 2 transparent 2 not, then you have to do it manually. The same as the current kicad tool.

BUT you can revert by using ctrlZ.

What you describe is something like the 'Draw style', which override the transparency property at the viewprovider level. So a change that is not persistent, ie that do not save. IMO this would be more confusing for the user. Especially since you can ctrlZ I don't see the point of a temporary change. But that is up for discussion of course

sasobadovinac:
I agree with @Roy-043, changing the model and/or its properties in any way, just for some temporary changes/state/view is considered bad practice. Even for basic individual projects this is not okay, but could maybe be somewhat acceptable, but for bigger and especially group projects this is just bad. In more advanced setups where you would get access to only part of the model and have the rest read-only, this would not even work.

PaddleStroke:
I have no time to handle the asked feature modifications. Besides not everyone seems to agree on this modification request. But its mostly that I have to focus on assembly right now and on top of that Abdullah is back to work on sketcher to work on offset and co.
So I suggest that perhaps this is added like this, and an issue is opened to ask to change for a temporary transparency modification.

tigert:
I would guess it comes down to the usecase really.

I am still learning FreeCAD but I usually use transparency to see better what I am doing when the object itself obstructs some of the geometry behind itself, and I cannot, for example, see the sketch I am doing. While this is maybe something Sketcher could also handle better, I think at least this case would be handled with a View mode, like wireframe/solid/shaded/transparent (that could maybe even have some kind of an onion skin style thing, or a slider / mouse wheel modifier binding to adjust the transparency. 

The objects also do have transparency among other style properties, like if you make a part that actually is translucent. But I also think changing this property as an operation that both changes all selected objects but also resets them all to the same transparency level might be a bit of a firehose. 

I mean, to me it feels like a view mode that is implemented as a property change.

I might well have not thought it throughly yet, but this is my first impression.

JohnOCFII:
> While this is maybe something Sketcher could also handle better, I think at least this case would be handled with a View mode, like wireframe/solid/shaded/transparent (that could maybe even have some kind of an onion skin style thing, or a slider / mouse wheel modifier binding to adjust the transparency).

That capability exists.

Screenshot 2023-10-05 at 9 02 13 AM
Screenshot 2023-10-05 at 9 02 34 AM
Screenshot 2023-10-05 at 9 03 02 AM

maxwxyz:
Is there a possibility to integrate this with the concept of a show/no show space (or group of objects with currently 63% transparency or hidden objects [maybe identified with 0% transparency]). Then switch the visibility between all visible objects and all hidden objects.
[viewtopic.php?p=665961](https://forum.freecad.org/viewtopic.php?p=665961)

Reference from CATIA: http://catiadoc.free.fr/online/intug_C2/intugrf0220.htm

PaddleStroke:
It has been suggested that :

> "Could the the space-bar toggle of visibility be changed to toggle between visible->transp->hidden?"

This sounds good to me. What are your thoughts on that? Visibility is a persistent change. So it it toggle between transparent as well it would make sense for it to be persistent.

tigert:
Would it make sense to steer the object tree to look a bit more like a layers/objects dialog in other design apps? Something like this could work, so you could have a "x-ray" view mode and toggle visibility (that also works via space now) as mode indicator icons in the tree?

image

obelisk79:
After a rather lengthy discussion with the design working group the general recommendation we have to offer is this:

1. Toolbuttons should not be used to modify display parameters for single bodies. This should be restricted to global controls.
2. We recommend the existing functionality of using the space bar as a keyboard toggle for visibility be extended to support 3 visual states. "Normal" > "Transparent" > "Hidden"
3. One of the following options be implemented as the gui control for this toggle.
    - (preferred option) Implement a *single* icon located to the left of bodies/containers with the 3 states. Normal, transparent, and hidden. Different icons can be used for each state to indicate the setting.
    - (secondary option) add an entry to the context (right click) menu for treeview to directly toggle the transparency.
    
Below is an example of what is being proposed for the treeview, and if this is implemented, I can draft the 3 icons.
image

PaddleStroke:
I agree for the space key cycling through.
For the eye icons I'm not sure. For me its over-loading the tree. Linkstage has this I seem to remember and I thought it was a bit heavy.

Ah found it. Here's how it looks : 
image

Hmm so yes not a huge fan. I would rather just put the command in the right click menu.

obelisk79:
@PaddleStroke I agree that the linkstage implementation is quite cluttered, my proposal was more limiting. In your screenshot there are a few items of note. The icon is added to everything. Also, linkstage adds other additional icons (up to 3 in addition to the object's type) all over, which again, adds clutter. My proposal was to limit only one extra icon and have it *only* be present on bodies and container objects. This means: Body, Part, Group, other geometry objects ie Part WB only rather than with features, origin planes, axis etc.

In either case, menu or indicator in the tree, are acceptable solutions.

PaddleStroke:
Ah yes if icons are only on bodies and container it wouldn't be so bad.
Still this is more work that can happen in a separate PR.
Here I propose that we just remove the toolbar icon and instead add it to the cycling of space bar.
Then in a subsequent PR the right click context menu can be replace by the icons.

obelisk79:
I think that is a good plan.

maxwxyz:
Currently when selecting an object in the 3D view and hitting space bar - the last Part Design Feature is hidden, not the body or part. This makes it hard to see in the collapsed tree view, which items are visible or not. Changing visibility ob the body does not toggle the last feature of the body. 

Can this be tackled in this approach to toggle the visibility of the whole body, not the last Part Design feature of the selected body?

PaddleStroke:
So it appears that doing thy cycling on space bar is pausing many questions. For simple solids it's very simple indeed.
But for App::Parts, LinkGroups, Groups it becomes a mess.
Let's take App::Part for example. It has a visibility property, but no transparency property.

So currently to toggle the visibility of a App::Part, you change visibility of the App::Part, not of the subobjects.
But for transparency the toggle I made is changing the transparency of the subobjects.

It feels like there're so many cases that it's going to be a big project.

sliptonic:
Ok.  It looks like we're feature creeping.   Should we move this back to draft or merge and improve in a later PR?

PaddleStroke:
Well its for maintainers to decide I guess. I would say merge to avoid rebasing issues.

obelisk79:
Leave in right-click menu, exclude the qtoolbutton by default and merge would be my suggestion.

maxwxyz:
Does not work as expected, shall I create a new issue?

Footz355:
@PaddleStroke @obelisk79 How does "toggle transparency of objects" compare to LS3's "transparency on demand" after clicking on object? From my experience working with LS3 the most usefull feature of making body temporary transparent after selecting a face/edge is the clickthrough feature: you don't have to rotate the body/assembly all around to select all the features you want to modify, you just click through the body. You think that is a bad practice? While coming back to master It seems to me now like working with one eye closed due to to the lack of that feature. If this is a bad place to discuss this issue, should I continue it on discord or forum?

PaddleStroke:
I think you probably should open a feature request (github issue) describing what you want. This would be the best place to discuss this.

Roy-043:
Small detail: the position of the tool in the View menu seems strange. Why not place the tool after "Color per face". It would fit in that group of commands I think.

PaddleStroke:
This is true, I am not sure what came to my mind when I did this.

PR #10816, opened 2023-09-25
Arch: ArchSectionPlane.getSVG add linespacing argument

Roy-043:
Forum topic:
https://forum.freecad.org/viewtopic.php?t=81470


yorikvanhavre:
Thx! The build failures seem related to FEM, so I think this is safe to merge...

PR #9446, opened 2023-04-30
Gui: Refactor navigation animations

Rexbas:
Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [X]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---

I have been working to improve navigation animations. This PR contains mostly refactoring. I have added a `NavigationAnimator` that manages active `NavigationAnimation`s. An animation must be played through the animator. Currently, only animations to move to a selected view are refactored (e.g. when interacting with the NaviCube). These kinds of animations have been changed a little bit. They now have a fixed duration and the duration can be changed in the preferences. The option in preferences meets the demand described here: https://forum.freecad.org/viewtopic.php?t=76406. The PR also fixes #9415. 

`NavigationStyle` still contains some methods related to animations but I believe this is only for the infinite spin animation used by the Open Inventor style. I plan to refactor these animations in the future so all animations are played through the animator. I also want to add an option to the animations so they accelerate and decelerate instead of having a constant velocity.

PR #10828, opened 2023-09-26
Add nullptr check to prevent segfault

Rexbas:
Attempt to fix the segfault during test caused by #9446. Adding a nullptr check appeared to prevent the segfault when I ran the actions locally using act.


wwmayer:
This check is not sufficient. When using the turntable dialog then it still crashes when pressing the Play button twice.

PR #10835, opened 2023-09-27
Path: fix compiler warnings

wwmayer:
* fix -Wunused-parameter
* fix -Wunused-but-set-variable

### DIFF
--- a/src/Mod/Path/App/Area.cpp
+++ b/src/Mod/Path/App/Area.cpp
@@ -501,6 +501,8 @@ std::shared_ptr<Area> Area::getClearedArea(double tipDiameter, double diameter) 
 #define AREA_MY(_param) myParams.PARAM_FNAME(_param)
     PARAM_ENUM_CONVERT(AREA_MY, PARAM_FNAME, PARAM_ENUM_EXCEPT, AREA_PARAMS_OFFSET_CONF);
     PARAM_ENUM_CONVERT(AREA_MY, PARAM_FNAME, PARAM_ENUM_EXCEPT, AREA_PARAMS_CLIPPER_FILL);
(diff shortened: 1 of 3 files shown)


0penBrain:
I tried to re-run CI (once only failed test, once completely) but this PR seems to very reliably trigger the segfault that shew quite a lot since 4435a3974f64dd52829c0c24d1be0f14405d5218 :confused:

wwmayer:
I guess the crash came rather from this PR: https://github.com/FreeCAD/FreeCAD/pull/9446

wwmayer:
Rebased and force pushed.

0penBrain:
> I guess the crash came rather from this PR: #9446

It has been reverted here with ffd66cf
So with rebate should go well. We'll see

wwmayer:
> It has been reverted here with

Sure, but when you only trigger a rebuild without rebasing the branch is it then supposed to work?

Now there is no segfault any more but one of the tests fails: TimeInfo.TestDiffTime

0penBrain:
> Sure, but when you only trigger a rebuild without rebasing the branch is it then supposed to work?

It may if the problem is randomly appearing. But sure it compiles the exact same tree.

> Now there is no segfault any more but one of the tests fails: TimeInfo.TestDiffTime

Maybe another unlucky merge. Didn't look at it yet.

> But nevertheless, this PR only fixes some compiler warnings and doesn't break anything.

Exactly. Why it was interesting to use it to analyze an eventual random bug. ;)

0penBrain:
> Now there is no segfault any more but one of the tests fails: TimeInfo.TestDiffTime

Well, it's gone after re-running the CI. Hopefully it was just a glitch.

Issue #10834, opened 2023-09-27
[Problem] Macro Toolbar : Record and Stop should be the same tool.


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

The macro toolbar has : Record tool and Stop tool. Those tools are never active at the same time, because obviously one starts the recording and the other stops it.
So the 'stop recording' could be removed from the toolbar if the recording tool could handle the stop recording.

This would save one icon from the toolbars at no cost.

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10836, opened 2023-09-27
[Macro toolbar] Remove stop from toolbar and menu

PaddleStroke:
And adds the stop functionality to the Std_DlgMacroRecord command. ie if you start recording, the button becomes a stop button.
image

There might be some usecases where the user stops the recording by another mean than the command (in python for example?) and in this case the command wouldn't update. But I don't know if there are actually such usecases.

Fixes https://github.com/FreeCAD/FreeCAD/issues/10834

### DIFF
--- a/src/Gui/CommandMacro.cpp
+++ b/src/Gui/CommandMacro.cpp
@@ -21,9 +21,14 @@
  ***************************************************************************/
 
 #include "PreCompiled.h"
+#ifndef _PreComp_
+# include <QApplication>
+#endif
 
 #include "Command.h"
+#include "Action.h"
 #include "Application.h"
+#include "BitmapFactory.h"
 #include "DlgMacroExecuteImp.h"
 #include "DlgMacroRecordImp.h"
 #include "Macro.h"
@@ -54,13 +59,27 @@ StdCmdDlgMacroRecord::StdCmdDlgMacroRecord()
 void StdCmdDlgMacroRecord::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    Gui::Dialog::DlgMacroRecordImp cDlg(getMainWindow());
-    cDlg.exec();
+    if (!getGuiApplication()->macroManager()->isOpen()){
+        Gui::Dialog::DlgMacroRecordImp cDlg(getMainWindow());
+        if (cDlg.exec() && getAction()) {
+            getAction()->setIcon(Gui::BitmapFactory().iconFromTheme("media-playback-stop"));
+            getAction()->setText(QCoreApplication::translate("StdCmdDlgMacroRecord", "S&top macro recording"));
+            getAction()->setToolTip(QCoreApplication::translate("StdCmdDlgMacroRecord", "Stop the macro recording session"));
+        }
+    }
+    else {
+        getGuiApplication()->macroManager()->commit();
+        if (getAction()) {
+            getAction()->setIcon(Gui::BitmapFactory().iconFromTheme("media-record"));
+            getAction()->setText(QString::fromLatin1(sMenuText));
+            getAction()->setToolTip(QString::fromLatin1(sToolTipText));
+        }
+    }
 }
 
 bool StdCmdDlgMacroRecord::isActive()
 {
-    return ! (getGuiApplication()->macroManager()->isOpen());
+    return true;
 }
 
 //===========================================================================
--- a/src/Gui/CommandMacro.cpp
+++ b/src/Gui/CommandMacro.cpp
@@ -83,34 +83,6 @@ bool StdCmdDlgMacroRecord::isActive()
 }
 
 //===========================================================================
-// Std_MacroStopRecord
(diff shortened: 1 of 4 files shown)


src/Gui/CommandMacro.cpp
```diff
@@ -54,13 +56,27 @@ StdCmdDlgMacroRecord::StdCmdDlgMacroRecord()
 void StdCmdDlgMacroRecord::activated(int iMsg)
 {
     Q_UNUSED(iMsg);
-    Gui::Dialog::DlgMacroRecordImp cDlg(getMainWindow());
-    cDlg.exec();
+    if (!getGuiApplication()->macroManager()->isOpen()){
+        Gui::Dialog::DlgMacroRecordImp cDlg(getMainWindow());
+        if (cDlg.exec() && getAction()) {
+            getAction()->setIcon(Gui::BitmapFactory().iconFromTheme("media-playback-stop"));
+            getAction()->setText(QString::fromLatin1(QT_TR_NOOP("S&top macro recording")));
```
chennes, line 63: I don't think using the NOOP version is correct here, I think you need the actual translated text (so using `tr()` instead).

PaddleStroke: tr() is a QObject function. So it would be QObject::tr(...). Is that correct ? I pushed a commit. If that's ok let me know and I'll squash.

0penBrain:
> There might be some usecases where the user stops the recording by another mean than the command (in python for example?) and in this case the command wouldn't update. But I don't know if there are actually such usecases.

AFAIK there is no Python API to control the macro manager. You can limit the risk by also removing the loading of `Std_StopMacroRecorder` from here https://github.com/FreeCAD/FreeCAD/blob/fe4f1a4bc80f2cba04f436b1241d12c5882e1e72/src/Gui/Workbench.cpp#L828 because it currently makes the command available for a direct call in the Python Console.

chennes:
A direct call of the command will work properly, though, won't it? So no need to disable.

src/Gui/CommandMacro.cpp
chennes: If you directly call the QObject version doesn't it get the context wrong? I thought this class was derived from QObject (sorry, not at a computer right now to check).

0penBrain:
> A direct call of the command will work properly, though, won't it? So no need to disable.

That won't "break everything" for sure. But the icon won't be in sync with actual situation anymore.

src/Gui/CommandMacro.cpp
PaddleStroke: I don't know. Sorry I don't know the translation code at all so it is not clear when to use one or another. Though what is sure is that Gui::Command are not QObjects

PaddleStroke:
@0penBrain thanks, I removed it from there as well.

Alternatively what I can do is modify the stop command so that when it is activated it re-initialize the icon/texts of the record command ?
Or perhaps remove the stop recording command completely.

0penBrain:
> Alternatively what I can do is modify the stop command so that when it is activated it re-initialize the icon/texts of the record command ? Or perhaps remove the stop recording command completely.

After your change I don't see any purpose for the Stop command, so IMO you can remove it. Eventually better to do it in a dedicated commit so if one want to restore the command it will be easy.

PaddleStroke:
I removed the Std_MacroStopRecord command in a separate commit. (and removed some reference to it in some places)

PaddleStroke:
So the last standing question is the ` QObject::tr` vs `NOOP` @chennes.
After you confirm I will squash or drop and switch the PR to open.

0penBrain:
> So the last standing question is the ` QObject::tr` vs `NOOP` @chennes. After you confirm I will squash or drop and switch the PR to open.

Not really. 'NOOP' is sure not the good way. The question is 'QObject::tr' or 'qApp::translate' which allows a context. 😉

src/Gui/CommandMacro.cpp
chennes: Ah, sorry, I say the "Dlg" and immediately thought of the Dialog classes. So using QObject::tr() should be correct and put it in the right context. I'll make a note to check on the next translation cycle.

chennes:
> > So the last standing question is the ` QObject::tr` vs `NOOP` @chennes. After you confirm I will squash or drop and switch the PR to open.
> 
> Not really. 'NOOP' is sure not the good way. The question is 'QObject::tr' or 'qApp::translate' which allows a context. 😉

Yes, you're right, we need to set the context to the name of the class, so just `tr()` wasn't correct. Sorry!

src/Gui/CommandMacro.cpp
PaddleStroke: I have done so, squashed and and force pushed. Please have a look thanks.

chennes:
LGTM, let's wait for the CI to complete and then merge.

PaddleStroke:
Sorry missing header.

PR #10838, opened 2023-09-27
Draft: Update the Plane class (step 1)

Roy-043:
After this PR the `Plane` class inherits from the `PlaneBase` class.

Related issue:
#5603.

Previous PR:
#10766

The following functions have been updated:

`__init__`
* `u` argument can be a WP now.
* Input vectors are copied.
* Added the `weak` argument.
* the `doc` attribute was not used and has been removed.

`__repr__`
* Inherited.
* The `position` is now also shown.

`copy`
* Uses `match` from the parent class which uses the local `_get_prop_list` function. Again vectors are copied.

`offsetToPoint`
* I could not find evidence that this function is used anywhere.
* The `direction` argument did not make a lot of sense (only worked in 2D). It is now ignored.
* Removed the overly long docstring.
* The function now uses an external function: `DraftGeomUtils.distance_to_plane`.

`projectPoint`
* The function calls `super().project_point` which in turn calls `DraftGeomUtils.project_point_on_plane`.
* `from sys import float_info` is no longer required.

`projectPointOld`
* Obsolete and removed.

`align*`
* All align* functions now have an `offset` argument.
* All align* functions return `True` or `False`.

`alignToPointAndAxis`
* The function ignored `upvec` if `axis` was parallel to the global X axis. This meant that the "Set WP" button of the Draft_Wire command did not work properly if the WP was aligned with the Right or Left view.
* The function calls `align_to_point_and_axis` which determines the rotation with `FreeCAD.Rotation(Vector(), upvec, axis, "ZYX")`.

`alignToPointAndAxis_SVG`
* The function calls `super().align_to_point_and_axis_svg` which contains the code from the old function.

`alignToCurve`
* Was not implemented.

`alignToEdges`
* Original code required two parallel edges (for example from an axis system), else the `u` and `v` vectors would not be perpendicular.
* The function calls `super().align_to_edges_vertexes` now.

`alignToFace`
* The function now uses an external function: `DraftGeomUtils.placement_from_face`. The rotation algorithm for the `u` and `v` vectors has been moved there.

`alignTo3Points`
* Basically unchanged. `Part` is already imported at the top of the file.

`alignToSelection`
* Basically unchanged. Will be replaced in the `PlaneGui` class.

### DIFF
--- a/src/Mod/Draft/WorkingPlane.py
+++ b/src/Mod/Draft/WorkingPlane.py
@@ -35,13 +35,13 @@ YZ, and XZ planes.
 #  in FreeCAD and a couple of utility functions.
 
 import math
-from sys import float_info
 import lazy_loader.lazy_loader as lz
 
 import FreeCAD
 import DraftVecUtils
 from FreeCAD import Vector
 from draftutils import utils
+from draftutils.messages import _wrn
 from draftutils.translate import translate
 
 DraftGeomUtils = lz.LazyLoader("DraftGeomUtils", globals(), "DraftGeomUtils")
@@ -534,7 +534,7 @@ class PlaneBase:
         direction: Base.Vector, optional
             Defaults to `None` in which case the WP `axis` is used.
             Direction of projection.
-        force_projection: Bool, optional
+        force_projection: bool, optional
             Defaults to `True`.
             See DraftGeomUtils.project_point_on_plane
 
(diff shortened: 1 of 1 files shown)


yorikvanhavre:
@Roy-043 we had an interesting discussion yesterday night with @sliptonic and @adrianinsaval about your work on the Draft grid, and the possibility to have a single grid system shared between Draft and the Sketcher. I think we could all sit and talk about this and look at what would be possible... Interested?

Roy-043:
Of course I am interested. We could plan something this Sunday perhaps? Starting a GitHub discussion may also be a good idea. That way things are public and others can participate and comment.

yorikvanhavre:
This sunday is a good idea as it is the first sunday of the month, and therefore our monthly developers meeting! I added the idea to the agenda https://forum.freecad.org/viewtopic.php?t=81584

yorikvanhavre:
Related discussion: FreeCAD/DevelopersHandbook#65

Roy-043:
Note that this PR does not deal with the grid yet. IMO it can be merged.

yorikvanhavre:
Indeed! Let's do it

PR #10844, opened 2023-09-27
[cmake] some definitions do not matter build now

ppphp:
The scripts are 12 years ago. `CompilerChecksAndSetups.cmake` should implement the function


chennes:
Thanks for the PR -- do you know what version of MSVC (or cMake) was that last that required these lines?

ppphp:
`FCAPPxxxx` is not used in repo now. `limits.h` is now for a boost file `python22_fixed.h`, it is a c99 library. `-DHAVE_ACOSH -DHAVE_ASINH -DHAVE_ATANH` are also in c99 standard. C++11 and above support them well.

chennes:
Thanks -- all of the CI build systems work fine, and I can compile locally. Obviously we can't test every possible system, so if this change gives anyone trouble please message me on the Forums, Discord, or Telegram. Thanks again for the PR!

PR #10845, opened 2023-09-27
Draft: ImportDXF: CURRENTDXFLIB is not obsolete and should be kept

Roy-043:
Forum topics:
https://forum.freecad.org/viewtopic.php?t=81422
https://forum.freecad.org/viewtopic.php?t=81523


yorikvanhavre:
Hm I see... the version is important to tell FreeCAD it needs to update. But what if we removed that mechanism and relied only on the addon manager?

Roy-043:
I am not sure what you mean. Informing the user they need to update by manually starting the Addon Manager would be inconvenient (and would still require a version check).

yorikvanhavre:
You're right. In any case let's do this in different steps. So this can be merged now.

PR #10848, opened 2023-09-27
Gui: fixes #10807: By default keep the old tabulated layout of ComboV…

### LINKED ISSUE #10807: [UI Problem] Tasks panel separation, user doesn't get asked if they wish to keep the old tabulated layout

### Problem description

While I can fully understand that some/maybe even a majority of users would like to have the Tasks panel separated as it is now in 0.22dev master, it's not in the FreeCAD ethos IMHO to not give the user a choice on the first run after an upgrade.

For example, I was trying some fairly simple Techdraw work over the weekend and just had enough of the fact that I kept having to expand/move the panel around in order to work efficiently, I wasn't aware that the user could even move it over the Model panel to have it back to how it was previously.

As I'm regularly using clean config files for testing, I made a once run dialog box asking the user whether they want to retain the old Tabulated or the new Separated layout. https://github.com/Syres916/FreeCAD/commit/28894e9def6d2cffef7e479876221520adbcf490 This would be particularly of use to users with only 1024x768 resolution.

This maybe seen as unnecessary which is fine but at least I have a proposal alongside my gripe, sorry constructive feedback.

### Full version info

```shell
[code]
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34403 (Git)
Build type: Release
Branch: master
Hash: 37e477d3e7ed5af91eaa4e71ed2e12bc8a9c89ec
Python 3.8.10, Qt 5.12.8, Coin 4.0.0, Vtk 7.1.1, OCC 7.6.3
Installed mods: 
  * Silk 0.1.3
  * Defeaturing 1.2.0
  * CurvedShapes 1.0.4
  * freecad.gears 1.0.0
  * Curves 0.6.13
  * Help 1.0.3
  * fasteners 0.4.56
  * sheetmetal 0.3.0
[/code]
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

wwmayer:
…iew and TaskView

By default the Combo mode is activated again.

See also: https://forum.freecad.org/viewtopic.php?t=81472

### DIFF
--- a/src/Gui/DockWindowManager.cpp
+++ b/src/Gui/DockWindowManager.cpp
@@ -380,6 +380,20 @@ void DockWindowManager::setup(DockWindowItems* items)
         }
     }
 
+    // Don't always tabify after switching the workbench
+    static bool tabify = false;
+    if (!tabify) {
+        // tabify dock widgets for which "tabbed" is true and which have the same position
+        for (int i=0; i<4; i++) {
+            const QList<QDockWidget*>& dws = areas[i];
+            for (QList<QDockWidget*>::ConstIterator it = dws.begin(); it != dws.end(); ++it) {
+                if (*it != dws.front()) {
+                    getMainWindow()->tabifyDockWidget(dws.front(), *it);
+                    tabify = true;
+                }
+            }
+        }
+    }
 }
 
 void DockWindowManager::saveState()
--- a/src/Gui/DockWindowManager.cpp
+++ b/src/Gui/DockWindowManager.cpp
@@ -23,6 +23,7 @@
 
 #include "PreCompiled.h"
 #ifndef _PreComp_
+# include <array>
 # include <QAction>
 # include <QDockWidget>
 # include <QMap>
@@ -337,18 +338,17 @@ void DockWindowManager::setup(DockWindowItems* items)
                                ->GetGroup("MainWindow")->GetGroup("DockWindows");
     QList<QDockWidget*> docked = d->_dockedWindows;
     const QList<DockWindowItem>& dws = items->dockWidgets();
-    QList<QDockWidget*> areas[4];
-    for (QList<DockWindowItem>::ConstIterator it = dws.begin(); it != dws.end(); ++it) {
-        QDockWidget* dw = findDockWidget(docked, it->name);
-        QByteArray dockName = it->name.toLatin1();
-        bool visible = hPref->GetBool(dockName.constData(), it->visibility);
+    for (const auto& it : dws) {
+        QDockWidget* dw = findDockWidget(docked, it.name);
+        QByteArray dockName = it.name.toLatin1();
+        bool visible = hPref->GetBool(dockName.constData(), it.visibility);
 
         if (!dw) {
-            QMap<QString, QPointer<QWidget> >::Iterator jt = d->_dockWindows.find(it->name);
+            QMap<QString, QPointer<QWidget> >::Iterator jt = d->_dockWindows.find(it.name);
             if (jt != d->_dockWindows.end()) {
-                dw = addDockWindow(jt.value()->objectName().toUtf8(), jt.value(), it->pos);
+                dw = addDockWindow(jt.value()->objectName().toUtf8(), jt.value(), it.pos);
                 jt.value()->show();
-                dw->toggleViewAction()->setData(it->name);
+                dw->toggleViewAction()->setData(it.name);
                 dw->setVisible(visible);
             }
         }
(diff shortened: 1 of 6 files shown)


0penBrain:
Tested (quickly). Works fine.
Eventually the only remaining "problem" to me is that in default config, "Task" widget is raised.
IMO would be better that "Model" is raised. :wink:

src/Gui/DockWindowManager.cpp
```diff
@@ -380,6 +380,20 @@ void DockWindowManager::setup(DockWindowItems* items)
         }
     }
 
+    // Don't always tabify after switching the workbench
+    static bool tabify = false;
+    if (!tabify) {
+        // tabify dock widgets for which "tabbed" is true and which have the same position
+        for (int i=0; i<4; i++) {
```
0penBrain, line 387: Would that make sense to iterate up to `areas->length()` just in case it evolves in the future ?

src/Gui/DockWindowManager.cpp
```diff
@@ -380,6 +380,20 @@ void DockWindowManager::setup(DockWindowItems* items)
         }
     }
 
+    // Don't always tabify after switching the workbench
+    static bool tabify = false;
+    if (!tabify) {
+        // tabify dock widgets for which "tabbed" is true and which have the same position
+        for (int i=0; i<4; i++) {
+            const QList<QDockWidget*>& dws = areas[i];
+            for (QList<QDockWidget*>::ConstIterator it = dws.begin(); it != dws.end(); ++it) {
```
0penBrain, line 389: Are you against for range-based for loop ? :smile:

src/Gui/DockWindowManager.cpp
wwmayer: Not possible because areas is a C-array of QList's: `QList<QDockWidget*> areas[4];`
An option is to use a std::array instead.

src/Gui/DockWindowManager.cpp
wwmayer: No. In new code I use range-based loops where ever possible. This is just the old code that has been removed a year ago and I have put it back.

0penBrain: Right. `areas` is a local variable so we shouldn't expect much problems changing its type. Let you see what you think is best.

src/Gui/DockWindowManager.cpp
0penBrain: Right. `areas` is a local variable so we shouldn't expect much problems when changing its type. Let you see what you think is best.

0penBrain: Other solution : how about using a `constexpr` of `log2(Qt::AllDockWidgetAreas+1)` to define the array size and so the loop range ?

wwmayer:
> Eventually the only remaining "problem" to me is that in default config, "Task" widget is raised.
IMO would be better that "Model" is raised

Fixed now. The first dock widget of the list will be activated.

src/Gui/DockWindowManager.cpp
wwmayer: Replaced the C array with std::array. Also refactored the function setup() as it was becoming too long.

src/Gui/DockWindowManager.cpp
wwmayer: Replaced the iterators with a range-based loop. But this will have happened anyway when continuing my work to modernize the C++ code base.

src/Gui/DockWindowManager.cpp
```diff
@@ -380,6 +391,24 @@ void DockWindowManager::setup(DockWindowItems* items)
         }
     }
 
+    // Don't always tabify after switching the workbench
+    static bool tabify = false;
+    if (!tabify) {
```
0penBrain, line 396: @wwmayer I'm not clear about that.
The comment says `// Dont' always` but from the code the logic makes that it can happen only once : `tabify` can never become `false` again after it has been `true`. Can you tell what is the functional goal of this ?

Also if I'm correct, the rest of the function above does nothing real. So shouldn't we also prevent it to run if `tabify` is `true` ?
Thx

wwmayer: It is supposed to happen only once. When a user has modified the default layout he for sure don't want to restore it every time he switches to another workbench, no?

> Also if I'm correct,

Correct.

0penBrain: > It is supposed to happen only once. When a user has modified the default layout he for sure don't want to restore it every time he switches to another workbench, no?

Sure. I would have been no doubt about implementation if the comment said `only once`. :wink: The `Not always` make we wonder about the implementation correctness. ;)
One last point fo rmy full understanding. I guess that if "tabification" shall happen, it should do only at first call (when WB layout is initialized).
But with current implementation, `tabify` could stay `false` if WB initializes with no tabbed widgets. Well, OK, in this case `tabifyDockWidgets` should always do nothing, but it will be completely processed each time. Couldn't we make `tabify = true` unconditionally at the end of the function so it will return immediately after 1st call ?
Thx -- sorry for this :smile: --

wwmayer: >But with current implementation, tabify could stay false if WB initializes with no tabbed widgets.

Yes. During the startup phase of FreeCAD an empty workbench is used for the initialization because at this point the real workbench cannot be used yet. That's why the `tabify` variable is only changed if something has happened.

In the worst case what could happen (none of the workbenches have dock widgets that can be tabbed) the function will be called each time after switching a workbench and it will never do anything. But is this a problem? I don't think so.

> Couldn't we make tabify = true unconditionally at the end of the function so it will return immediately after 1st call ?

Sorry but I don't get your point?

0penBrain: > Yes. During the startup phase of FreeCAD an empty workbench is used for the initialization because at this point the real workbench cannot be used yet. That's why the `tabify` variable is only changed if something has happened.

Thanks, was precisely the information I was missing.

> In the worst case what could happen (none of the workbenches have dock widgets that can be tabbed) the function will be called each time after switching a workbench and it will never do anything. But is this a problem? I don't think so.

Not a problem per se. :) Just seeking for code meaningfulness to help future maintenance. With your above explanation, good as is. :+1:

> > Couldn't we make tabify = true unconditionally at the end of the function so it will return immediately after 1st call ?
> 
> Sorry but I don't get your point?

I thought it shall be call only once + at first call. With your above explanation doesn't make much sense anymore.

Thanks for the details. Going to merge.

PR #10850, opened 2023-09-27
[TD]fix missing reference geometry

WandererFan:
This PR addresses the handling of Dimension references that are missing their reference geometry.


luzpaz:
@WandererFan just wondering if there is a ticket open related to this fix ?

WandererFan:
No ticket.  Just something I noticed while debugging something else.

PR #10852, opened 2023-09-27
Mat: fix several compiler warnings:

wwmayer:
* -Winconsistent-missing-override
* -Wtautological-undefined-compare


chennes:
@davesrocketshop just making sure you saw this

davesrocketshop:
In the many refactors that will be the next merge, I saw the "#include <memory>" requirement. The rest are new to me but good catches.

PR #10855, opened 2023-09-28
Material: include some more required header files

wwmayer:
as otherwise the parser of the clang code model raises a lot of error messages


luzpaz:
CC @davesrocketshop just a soft heads-up

PR #10857, opened 2023-09-28
Add G43 tool height compensation to centroid post

djmdjm:
This adds G43 tool height compensation to the centroid CAM post-processor using a similar approach to the G43 support in the MACH3 post.

Specifically, G43 is emitted at every tool change by default and a new `--no-tlo` post argument can restore the old behaviour of no height compensation being emitted. These both match how the MACH3 post does it.


LarryWoestman (approved):
This looks reasonable to me.

djmdjm:
ah, I didn't realise there are regress tests for the centroid post.  This additional commit should fix it. Otherwise, I'll have to figure out how to build/run the tests locally

LarryWoestman (approved):
I don't believe that the TestCentroidPost.py test is run by default, but it is a good idea to keep it updated to match the code.

The changes to TestCentroidPost.py look good to me.

Issue #10858, opened 2023-09-28
[Problem] LeadInOutDressup produces incorrect path when axis words are missing in the base operation


jffmichi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

When the base operation used for LeadInOutDressup doesn't specify all axis words for every command, e.g.

G0 X0 Y100 Z0
G1 X100          <-----

instead of

G0 X0 Y100 Z0
G1 X100 Y100 Z0

LeadInOutDressup treats the missing axis words as being 0 instead of the last used value. In the above example on  the marked line, the dressup assumes Y is 0. This is happening when applying a LeadInOutDressup to an operation with a DressupPathBoundary. See attached images for the current behaviour and the expected result from the example file. A pull request that fixes this is on the way.

### Full version info

```shell
[code]
OS: Arch Linux (KDE/plasmawayland)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34447 (Git)
Build type: None
Branch: master
Hash: aadca6a4a81688555bd1f65a97ebb1adf6591a62
Python 3.11.5, Qt 5.15.10, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * better-tool-library 0.9.0
  * dodo 1.0.0
  * sheetmetal 0.2.63
  * DynamicData 2.46.0
  * fcgear 1.0.0
  * fasteners 0.4.58
  * fcgear.backup1663629032.4145591 1.0.0 (Disabled)
  * Assembly4 0.50.2
  * lattice2 1.0.0
[/code]
```

### Subproject(s) affected?

Path

### Anything else?

![leadinout_error](https://github.com/FreeCAD/FreeCAD/assets/14004750/d1818b77-3672-47bc-8d10-be2e0ed5515c)
![leadinout_fix](https://github.com/FreeCAD/FreeCAD/assets/14004750/8d535030-3781-4ffa-a72c-75539f17826d)
[leadinout_error.zip](https://github.com/FreeCAD/FreeCAD/files/12753034/leadinout_error.zip)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #10859, opened 2023-09-28
Path: fix LeadInOutDressup incorrectly treating missing axis words as…

jffmichi:
… being 0

This PR fixes issue  #10858.

To fix the issue all missing axis words are inserted when splitting the g-code of the base operation into "layers". When those operations are later used in "getLeadStart" and "getLeadEnd" all axis words are available and the dressup produces the expected result.


jffmichi:
I tried to rework the LeadInOutDressup (see https://forum.freecad.org/viewtopic.php?t=81591). I guess this PR still remains relevant since it fixes an issue with minimal changes (?)

PR #10860, opened 2023-09-28
TechDraw: Implements projected Length dimension

FlachyJoe:
When a line and a vertex are selected, the Length dimension currently use closest points of the finite edge. It's not the mathematical definition of the distance which require the projection of the point on the infinite line.

When selecting the upper right vertex and the bottom line:
Left the current implementation, Right the true length given by this PR
image

https://forum.freecad.org/viewtopic.php?t=81561

EDIT: update screenshot with corrected extension line

### DIFF
--- a/src/Mod/TechDraw/App/DimensionGeometry.cpp
+++ b/src/Mod/TechDraw/App/DimensionGeometry.cpp
@@ -41,6 +41,14 @@ pointPair::pointPair(const pointPair& pp)
 {
     first(pp.first());
     second(pp.second());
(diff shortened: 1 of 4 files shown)


WandererFan:
Are we implementing this as is, or waiting on the extension line change?

FlachyJoe:
@WandererFan I switch back to draft until the extension line is correct.

FlachyJoe:
@WandererFan can you take a look please?
I keep 2 distinct commits but they can also be squashed.

WandererFan:
Merci!

PR #10878, opened 2023-10-01
[Feature Request] [Draft] [UI/UX] new Draft_Hyperlink command

### LINKED ISSUE #7062: [Feature Request] Draft label with hyperlink to internal/external doc/pdf/txt

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=23&t=69528

### Subproject(s) affected?

Draft, Export

### Idea description

Is it possible to set an draft label on the 3D Part and set up an hyperlink, to an internal/external document, such as a pdf drawing or a text document?
Click on the draft label and the system viewer displays the document.
It would be great if such a function could be exported and used as well after an export as an webGL file.

### Anything else?

_No response_

### Code of Conduct

CalligaroV:
Fixes #7062

**Forum discussion**

https://forum.freecad.org/viewtopic.php?f=23&t=69528

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_hyperlink.py
+++ b/src/Mod/Draft/draftguitools/gui_hyperlink.py
@@ -1,0 +1,116 @@
+# ***************************************************************************
+# *   Copyright (c) 2023 <https://www.freecad.org>                          *
+# *                                                                         *
+# *   This program is free software; you can redistribute it and/or modify  *
+# *   it under the terms of the GNU Lesser General Public License (LGPL)    *
+# *   as published by the Free Software Foundation; either version 2 of     *
+# *   the License, or (at your option) any later version.                   *
+# *   for detail see the LICENCE text file.                                 *
+# *                                                                         *
+# *   This program is distributed in the hope that it will be useful,       *
+# *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+# *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+# *   GNU Library General Public License for more details.                  *
+# *                                                                         *
+# *   You should have received a copy of the GNU Library General Public     *
+# *   License along with this program; if not, write to the Free Software   *
+# *   Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  *
+# *   USA                                                                   *
+# *                                                                         *
+# ***************************************************************************
+
+"""Provides GUI tools to open hyperlinks to internal/external documents."""
+
+## @package gui_hyperlink
+# \ingroup draftguitools
+# \brief Provides GUI tools to open hyperlinks to internal/external documents
+
+## \addtogroup draftguitools
+# @{
+
+import FreeCAD
+import os, sys, subprocess
+import re
+import draftutils.utils as utils
+from draftutils.messages import _msg
+
+if FreeCAD.GuiUp:
+    import FreeCADGui
+    import Draft_rc
+def QT_TRANSLATE_NOOP(ctx,txt):
+    return txt
+translate = FreeCAD.Qt.translate
+
+from PySide import QtWidgets
+
+__title__ = "FreeCAD Draft Workbench GUI Tools - Hyperlinks tools"
+__author__ = ("")
+__url__ = "https://www.freecad.org"
+
+
+class Draft_Hyperlink:
+    """The Draft_Hyperlink FreeCAD command definition."""
+
+    def GetResources(self):
+        d = {'Pixmap': '',
+             'Accel': "",
+             'MenuText': QT_TRANSLATE_NOOP("Draft_Hyperlink", "Open hyperlinks"),
+             'ToolTip': QT_TRANSLATE_NOOP("Draft_Hyperlink", "Open linked documents")}
+        return d
+
+    def Activated(self):
+        self.find_hyperlinks()
+
+        if len(self.hyperlinks_list) > 1:
+            m = QtWidgets.QMessageBox()
+            m.setWindowTitle(translate("draft", "Opening multiple hyperlinks"))
+            m.setText(
+                translate(
+                    "draft",
+                    "Multiple hyperlinks found."
+                )
+            )
+            m.setInformativeText(
+                translate(
+                    "draft",
+                    "This may lead to the opening of various windows"
+                )
+            )
+            m.setStandardButtons(m.Ok | m.Cancel)
+            ret = m.exec_()
+
+        if len(self.hyperlinks_list) == 1 or ret == m.Ok:
+            for hyperlink in self.hyperlinks_list:
+                self.open_hyperlink(hyperlink)
+
+    def find_hyperlinks(self):
+        self.hyperlinks_list = []
+
+        for o in FreeCADGui.Selection.getCompleteSelection():
+            if hasattr(o.Object, "Text"):
+
+                for text in o.Object.Text:
+                    hyperlinks = re.findall(r"((%\w+%|/\w+|\w:\\|\w{3,4}://)[\w\\/: ]+\.[\w?$#=\\\.]+)", text)
+
+                    for hyperlink in hyperlinks:
+                        self.hyperlinks_list.append(hyperlink[0])
+            
+    def has_hyperlinks(self):
+        self.find_hyperlinks()
+        
+        return len(self.hyperlinks_list) > 0
+
+    def open_hyperlink(self, hyperlink):
+        _msg(translate("draft", "Opening hyperlink") + " " + hyperlink)
+        
+        match sys.platform:
+            case ("win32", "cygwin"):
+                os.startfile(hyperlink)
+            case "darwin":
+                subprocess.call(["open", hyperlink])
+            case "linux":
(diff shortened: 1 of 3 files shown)


src/Mod/Draft/draftguitools/gui_hyperlink.py
```diff
@@ -0,0 +1,116 @@
...
+    def open_hyperlink(self, hyperlink):
+        _msg(translate("draft", "Opening hyperlink") + " " + hyperlink)
+        
+        match sys.platform:
+            case ("win32", "cygwin"):
+                os.startfile(hyperlink)
+            case "darwin":
+                subprocess.call(["open", hyperlink])
+            case "linux":
+                subprocess.call(["xdg-open", hyperlink])
```
adrianinsaval, line 112: would this work in flatpak and snap environments?

yorikvanhavre:
There are a few details to fix but otherwise the general idea behind this PR seems good.

src/Mod/Draft/draftguitools/gui_hyperlink.py
CalligaroV: Just started some research about that but found nothing ATM.

Python docs doesn't say nothing about that, or maybe I simply missed it.

I'll dig deeper, worst case scenario I'll have to make and test those environments but I guess it will take me some time, never done this kind of tests before :sweat_smile:

CalligaroV:
O.T.: are the PR Meetings now for maintainers / veterans only? It's the second week I send the request to join without success.

src/Mod/Draft/draftguitools/gui_hyperlink.py
```diff
@@ -0,0 +1,116 @@
...
+                    for hyperlink in hyperlinks:
+                        self.hyperlinks_list.append(hyperlink[0])
+            
+    def has_hyperlinks(self):
+        self.find_hyperlinks()
+        
+        return len(self.hyperlinks_list) > 0
+
+    def open_hyperlink(self, hyperlink):
+        _msg(translate("draft", "Opening hyperlink") + " " + hyperlink)
```
Roy-043, line 104: There should be a check if the file exists:
`os.path.isfile(hyperlink)`

src/Mod/Draft/draftguitools/gui_hyperlink.py
```diff
@@ -0,0 +1,116 @@
...
+        
+        match sys.platform:
+            case ("win32", "cygwin"):
```
Roy-043, line 107: This does not work, but I do not know the correct syntax. I have used `case "win32":` in my tests.

src/Mod/Draft/draftguitools/gui_hyperlink.py
```diff
@@ -0,0 +1,116 @@
...
+                self.open_hyperlink(hyperlink)
+
+    def find_hyperlinks(self):
+        self.hyperlinks_list = []
+
+        for o in FreeCADGui.Selection.getCompleteSelection():
+            if hasattr(o.Object, "Text"):
+
+                for text in o.Object.Text:
+                    hyperlinks = re.findall(r"((%\w+%|/\w+|\w:\\|\w{3,4}://)[\w\\/: ]+\.[\w?$#=\\\.]+)", text)
```
Roy-043, line 93: You have mentioned certain issues with the regular expression (https://github.com/FreeCAD/FreeCAD/issues/7062#issuecomment-1742073962). And I am not sure what is meant by 'internal/external document' in the feature request. But here are some more examples that the regular expression does not handle:
```
text = "File on server: \\Server2\Share\Test\Foo.txt"
text = "Alternative Windows path: D:/path/file.ext"
text = "Remote file: https://forum.freecad.org/viewtopic.php?f=23&t=69528"
```
Sadly I am not a regex specialist (you know much more than I do). Maybe one of the other devs has a suggestion, or else you could ask in the [Developers corner](https://forum.freecad.org/viewforum.php?f=10)

Roy-043 (commented):
It may be worth investigating if this feature should support objects from other workbenches as well. Not sure how that can be done if the code is part of the Draft WB, but it would fit in with our goal of more consistency.

src/Mod/Draft/draftguitools/gui_hyperlink.py
CalligaroV: Totally agree, don't know why I missed it.
My fault.

Adding it to the next commit

CalligaroV: for completeness, as an hyperlink could refer to a webpage or other remote files, probably your suggestion will be expanded to handle also those cases

yorikvanhavre:
Nothing really blocking... An icon is missing, and I would maybe rely on QDesktopServices to open the default browser, or offer the possibility to open the link inside FreeCAD using the Web WB. But all this can also be added later.

src/Mod/Draft/draftguitools/gui_hyperlink.py
```diff
@@ -0,0 +1,123 @@
...
+            case ("win32" | "cygwin"):
+                os.startfile(hyperlink)
+            case "darwin":
+                subprocess.call(["open", hyperlink])
+            case "linux":
+                subprocess.call(["xdg-open", hyperlink])
+
+FreeCADGui.addCommand('Draft_Hyperlink', Draft_Hyperlink())
+
+## @}
```
CalligaroV, line 123: This line must be removed.
Also the final empty line

src/Mod/Draft/draftguitools/gui_hyperlink.py
adrianinsaval: for flatpak I can add a wrapper for xdg-open, snap might be out of luck

src/Mod/Draft/draftguitools/gui_hyperlink.py
```diff
@@ -0,0 +1,123 @@
...
+    def open_hyperlink(self, hyperlink):
+        file_hyperlink = len(re.findall(r"^(\w:[\\/]|%\w+%|\\\\\w+|/\w+)", hyperlink)) > 0
+
+        if file_hyperlink and not os.path.isfile(hyperlink):
+            _msg(translate("draft", "File not found:") + " " + hyperlink)
+            return
+
+        _msg(translate("draft", "Opening hyperlink") + " " + hyperlink)
+
+        match sys.platform:
```
adrianinsaval, line 113: found the real problem tripping up CI, match is only available on python >3.10, ubuntu 20.04 and teh windows libpack have python 3.8. Can't remember what was our minimum supported version but it sure isn't 3.10

CalligaroV:
@yorikvanhavre, as suggested I changed the management of the hyperlinks, now they will be opened using QDesktopServices.

Still to do: add an icon and open the web pages with Web WB

@adrianinsaval, as now we don't use anymore `sys.platform`, `os.startfile` and `subprocess.call` do you think the compatibility issues you raised will be fixed?

yorikvanhavre:
Excellent!

For the icon, I would suggest you use one of the existing icons in src/Gui/Icons (maybe a "web browser" or  "internet" icon) for now, and we can develop a specific one later. These icons are already bundled in FreeCAD's resources so you just need to specify a path like this ":/icons/someicon.svg"

adrianinsaval:
> @adrianinsaval, as now we don't use anymore `sys.platform`, `os.startfile` and `subprocess.call` do you think the compatibility issues you raised will be fixed?

I don't know, but I suspect it will still be a problem. But it's not a merge a blocker, if it doesn't work in those platforms we can figure out some workaround specific for that. By the way xdg-open does work within the flatpak environment, not sure about snaps.

Roy-043:
I notice that something has been overlooked. Object specific context menu items are normally added in the viewprovider of the object. See for example view_layer.py.

yorikvanhavre:
But in this case, this command is designed to work across different object types (anything with a "Text" property). So it makes sense to have it as a command, and handle it at workbench level IMHO.

Roy-043:
Only Draft_Text and Draft_Label objects are currently supported. These (and other annotations) all use the same parent class for their viewprovider (view_draft_annotation.py).

You now have the inconsistency of some objects having context menu options available before the Draft WB has been loaded (layer, wire) and this new option that requires that the WB is loaded. Also context menu items added in the viewprovider are presented at the top of the context menu, this new option is not.

yorikvanhavre:
I get your point...
Indeed then it could be converted. It's probably just a matter of removing the command from the context menu in the WB class and adding it in view_draft_annotation...

CalligaroV:
@Roy-043, for sure I agree with you about the inconsistency issue.

I implemented this new function using `gui_setstyle.py` as a staring point as it's in the context menu as well.

Therefore, as we all agree on moving this function, I'll start working on this.

A final thought:
as of now this function is available both on the tree view and on the 3D view but, if I'm not wrong, moving it to `view_draft_annotation` will make it available only from the tree view.

I'd like to ask, both to you and to @yorikvanhavre, a comment about this last thought.

TIA!

Roy-043:
We should look more carefully at the Draft context menus in general. In the past I have done a quick cleanup (#5108) but more thought and effort is required. With that in mind I am fine with leaving this command where it is now and addressing this in the next dev cycle.

PR #10881, opened 2023-10-01
Gui: Refactor navigation animations (attempt 2)

### LINKED ISSUE #9415: [Problem] Navigation animations don't always work

### Version

0.20 (Release)

### Full version info

```shell
OS: KDE Flatpak runtime (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.20.2.29603 (Git) (and also 0.21)
Build type: Release
Branch: (HEAD detached at 0.20.2)
Python 3.10.6, Qt 5.15.8, Coin 4.0.0, Vtk 8.2.0, OCC 7.6.3
```

### Subproject(s) affected?

None

### Problem description

Some navigation animations don't work with animations enabled. The animations don't work when using the two arrows above the NaviCube. Also, alternating between the corner and edge of the NaviCube results in maybe one or two frames of animation and feels like it immediately snaps to the selected view. I attached a video that demonstrates this.

I have been working on something to deal with different kinds of navigation animations. A first version, that also fixes the animations, is almost finished so expect a PR soon.

### Anything else?

https://user-images.githubusercontent.com/14298143/234894433-039e478c-4b04-48e6-bebf-c49a26bea4d4.mp4

### Code of Conduct

Rexbas:
- Animations are played through an Animator
- Standard animations have a fixed duration and inherit from QVariantAnimation
- The animation duration can be changed in the preferences
- Fix animations when using the two arrows above the NaviCube
- Start an animation or start and wait for an animation
- Replaces standard camera animations (e.g. by selecting standard views or NaviCube) with FixedTimeAnimation
- Replace View3DInventorViewer's CameraAnimation with FixedTimeAnimation
- Replace OpenInventor navigation style infinite spin animation with SpinningAnimation
- Stops an active animation when a new animation is started
- Stops an active animation when the user starts dragging, panning or zooming
- Refactor reorientCamera so it can be used in animations
- Enable animations by default
- Fixes #9415 

This is the second attempt of #9446.

@wwmayer, double clicking the play button in the View Turntable from the tools menu should not crash anymore. And the minimum and maximum speed are faster. Can you confirm this?

### DIFF
--- a/src/Gui/NavigationAnimation.cpp
+++ b/src/Gui/NavigationAnimation.cpp
@@ -39,7 +39,7 @@ void NavigationAnimation::updateCurrentValue(const QVariant& value)
     update(value);
 }
 
-void NavigationAnimation::onStop(bool finished)
+void NavigationAnimation::onStop([[maybe_unused]] bool finished)
 {}
 
 FixedTimeAnimation::FixedTimeAnimation(NavigationStyle* navigation, const SbRotation& orientation,
@@ -158,7 +158,7 @@ void SpinningAnimation::update(const QVariant& value)
 /**
  * @param finished True when the animation is finished, false when interrupted
  */
-void SpinningAnimation::onStop(bool finished)
+void SpinningAnimation::onStop([[maybe_unused]] bool finished)
 {
     if (navigation->getViewingMode() != NavigationStyle::SPINNING) {
         return;
--- a/src/Gui/NavigationAnimation.cpp
+++ b/src/Gui/NavigationAnimation.cpp
@@ -31,11 +31,6 @@ NavigationAnimation::NavigationAnimation(NavigationStyle* navigation)
     : navigation(navigation)
 {}
 
-void NavigationAnimation::startAnimation(QAbstractAnimation::DeletionPolicy policy)
-{
-    QAbstractAnimation::start(policy);
-}
-
 void NavigationAnimation::updateCurrentValue(const QVariant& value)
 {
     if (state() == QAbstractAnimation::State::Stopped) {
--- a/src/Gui/NavigationAnimation.cpp
+++ b/src/Gui/NavigationAnimation.cpp
@@ -90,8 +90,9 @@ void FixedTimeAnimation::update(const QVariant& value)
 
     SbRotation rotation(rotationAxis, angle - prevAngle);
 
+    camera->position = camera->position.getValue() - prevTranslation;
     navigation->reorientCamera(camera, rotation, rotationCenter);
-    camera->position = camera->position.getValue() + translation - prevTranslation;
+    camera->position = camera->position.getValue() + translation;
 
     prevAngle = angle;
     prevTranslation = translation;
--- a/src/Gui/NavigationAnimation.cpp
+++ b/src/Gui/NavigationAnimation.cpp
@@ -39,10 +39,8 @@ void NavigationAnimation::updateCurrentValue(const QVariant& value)
     update(value);
 }
 
-void NavigationAnimation::stopAnimation()
-{
-    QAbstractAnimation::stop();
-}
+void NavigationAnimation::onStop(bool finished)
+{}
 
 FixedTimeAnimation::FixedTimeAnimation(NavigationStyle* navigation, const SbRotation& orientation,
                                        const SbVec3f& rotationCenter, const SbVec3f& translation,
@@ -139,9 +137,8 @@ void SpinningAnimation::update(const QVariant& value)
     prevAngle = value.toFloat();
 }
 
-void SpinningAnimation::stopAnimation()
+void SpinningAnimation::onStop(bool finished)
 {
-    NavigationAnimation::stopAnimation();
     if (navigation->getViewingMode() != NavigationStyle::SPINNING) {
         return;
     }
--- a/src/Gui/NavigationAnimation.cpp
+++ b/src/Gui/NavigationAnimation.cpp
@@ -1,0 +1,155 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/****************************************************************************
+ *                                                                          *
+ *   Copyright (c) 2023 Bas Ruigrok (Rexbas) <Rexbas@proton.me>             *
+ *                                                                          *
+ *   This file is part of FreeCAD.                                          *
+ *                                                                          *
+ *   FreeCAD is free software: you can redistribute it and/or modify it     *
+ *   under the terms of the GNU Lesser General Public License as            *
+ *   published by the Free Software Foundation, either version 2.1 of the   *
+ *   License, or (at your option) any later version.                        *
+ *                                                                          *
+ *   FreeCAD is distributed in the hope that it will be useful, but         *
+ *   WITHOUT ANY WARRANTY; without even the implied warranty of             *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU       *
+ *   Lesser General Public License for more details.                        *
+ *                                                                          *
+ *   You should have received a copy of the GNU Lesser General Public       *
+ *   License along with FreeCAD. If not, see                                *
+ *   <https://www.gnu.org/licenses/>.                                       *
+ *                                                                          *
+ ***************************************************************************/
+
+#include "PreCompiled.h"
+#include "NavigationAnimation.h"
+#include <Inventor/nodes/SoCamera.h>
+
+using namespace Gui;
+
+NavigationAnimation::NavigationAnimation(NavigationStyle* navigation)
+    : navigation(navigation), started(false)
+{}
+
+void NavigationAnimation::startAnimation(QAbstractAnimation::DeletionPolicy policy)
+{
+    started = true;
+    QAbstractAnimation::start(policy);
+}
+
+void NavigationAnimation::updateCurrentValue(const QVariant& value)
+{
+    if (!started) {
+        return;
+    }
+    update(value);
+}
+
+void NavigationAnimation::stopAnimation()
+{
+    QAbstractAnimation::stop();
+}
+
+FixedTimeAnimation::FixedTimeAnimation(NavigationStyle* navigation, const SbRotation& orientation,
+                                       const SbVec3f& rotationCenter, const SbVec3f& translation,
+                                       int duration)
+    : NavigationAnimation(navigation)
+    , targetOrientation(orientation)
+    , targetTranslation(translation)
+    , rotationCenter(rotationCenter)
+{
+    setDuration(duration);
+    setStartValue(0.0);
+    setEndValue(duration * 1.0);
+}
+
+void FixedTimeAnimation::initialize()
+{
+    prevAngle = 0;
+    prevTranslation = SbVec3f(0, 0, 0);
+
+    // Find an axis and angle to rotate from the camera orientation to the target orientation using post-multiplication
+    SbVec3f rotationAxisPost;
+    float angle;
+    SbRotation(navigation->getCamera()->orientation.getValue().inverse() * targetOrientation).getValue(rotationAxisPost, angle);
+    if (angle > M_PI) {
+        angle -= 2 * M_PI;
+    }
+
+    // Convert post-multiplication axis to a pre-multiplication axis
+    navigation->getCamera()->orientation.getValue().inverse().multVec(rotationAxisPost, rotationAxis);
+
+    angularVelocity = angle / duration();
+    linearVelocity = targetTranslation / duration();
+}
+
+/**
+ * @param value The elapsed time
+ */
+void FixedTimeAnimation::update(const QVariant& value)
+{
+    SoCamera* camera = navigation->getCamera();
+    if (!camera) {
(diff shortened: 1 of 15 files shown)


Rexbas:
@wwmayer, have you had any time to check whether the problems and crash you described (https://github.com/FreeCAD/FreeCAD/pull/9446#issuecomment-1735503473) have been resolved? I think they are resolved but if not then I would like to take another look.

wwmayer:
> have you had any time to check whether the problems 

Unfortunately, not yet. I hope to have a look at it soon.

Rexbas:
I added some commits to simplify some things. I also fixed a problem with animations that combine rotation and translation where the final position was not correct.

src/Gui/View3DPy.cpp
```diff
@@ -777,10 +777,8 @@ Py::Object View3DInventorPy::getCameraOrientation()
 
 Py::Object View3DInventorPy::viewPosition(const Py::Tuple& args)
 {
-    PyObject* p=nullptr;
-    int steps = 20;
-    int ms = 30;
-    if (!PyArg_ParseTuple(args.ptr(), "|O!ii",&Base::PlacementPy::Type,&p,&steps,&ms))
+    PyObject* p = nullptr;
```
wwmayer, line 780: This change is problematic because it may break existing Python macros.

Rexbas: Yes true, we could either keep the steps and ms parameters and ignore them so it does not break, or we could break it. I feel like it is better to clean it up and make any breaking changes before a 1.0 release. But let me know what you think.

src/Gui/View3DInventorViewer.cpp
```diff
@@ -3117,18 +3082,49 @@ SbBool View3DInventorViewer::isAnimating() const
...
+void View3DInventorViewer::startAnimation(const SbRotation& orientation,
+                                          const SbVec3f& rotationCenter, const SbVec3f& translation, bool wait)
+{
+    // Currently starts a FixedTimeAnimation. If there is going to be an additional animation like
+    // FixedVelocityAnimation, check the animation type from a parameter and start the right animation
+
+    int duration = App::GetApplication()
+                       .GetParameterGroupByPath("User parameter:BaseApp/Preferences/View")
+                       ->GetInt("AnimationDuration", 250);
+
```
wwmayer, line 3101: IMO it's a bad idea to make this function dependent on a user parameter because it will be more complicated to run consecutive calls of viewPosition() with different speeds. Then instead of a boolean argument and int could be used to increase flexibility.

src/Gui/NavigationStyle.cpp
```diff
@@ -353,171 +323,71 @@ SbBool NavigationStyle::lookAtPoint(const SbVec2s screenpos)
 
     SbVec3f hitpoint;
     hitpoint = picked->getPoint();
-    lookAtPoint(hitpoint);
+    this->rotationCenterFound = false;
```
wwmayer, line 326: The public function `void lookAtPoint(const SbVec3f&)` shouldn't be simply removed. You can implement it as
```
void NavigationStyle::lookAtPoint(const SbVec3f& pos)
{
    this->rotationCenterFound = false;
    translateCamera(hitpoint - getFocalPoint());
}
```
and leave its call in `lookAtPoint(const SbVec2s screenpos)`

src/Gui/View3DInventorViewer.cpp
```diff
@@ -3117,18 +3082,49 @@ SbBool View3DInventorViewer::isAnimating() const
...
 {
-    navigation->startAnimating(axis, velocity);
+    auto animation = std::make_shared<SpinningAnimation>(navigation, axis, velocity);
+    navigation->getAnimator()->start(animation);
 }
 
 void View3DInventorViewer::stopAnimating()
 {
-    navigation->stopAnimating();
+    navigation->getAnimator()->stop();
```
wwmayer, line 3127: Have you ever heard of the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter)? You should keep the method NavigationStyle::stopAnimation() and inside there call `getAnimator()->stop()`.

src/Gui/View3DInventorViewer.cpp
```diff
@@ -3117,18 +3082,49 @@ SbBool View3DInventorViewer::isAnimating() const
...
+ *
+ * @param axis The rotation axis in screen coordinates
+ * @param velocity The angular velocity in radians per second
  */
-void View3DInventorViewer::startAnimating(const SbVec3f& axis, float velocity)
+void View3DInventorViewer::startSpinningAnimation(const SbVec3f& axis, float velocity)
 {
-    navigation->startAnimating(axis, velocity);
+    auto animation = std::make_shared<SpinningAnimation>(navigation, axis, velocity);
+    navigation->getAnimator()->start(animation);
```
wwmayer, line 3122: Law of Demeter: Adjust NavigationStyle::startAnimating() to accept the SpinningAnimation and inside there call `getAnimator()->start(animation)`

src/Gui/View3DInventorViewer.cpp
```diff
@@ -3117,18 +3082,49 @@ SbBool View3DInventorViewer::isAnimating() const
...
+    // FixedVelocityAnimation, check the animation type from a parameter and start the right animation
+
+    int duration = App::GetApplication()
+                       .GetParameterGroupByPath("User parameter:BaseApp/Preferences/View")
+                       ->GetInt("AnimationDuration", 250);
+
+    auto animation = std::make_shared<FixedTimeAnimation>(
+        navigation, orientation, rotationCenter, translation, duration);
+
+    if (wait) {
```
wwmayer, line 3105: Law of Demeter

src/Gui/View3DInventorViewer.cpp
```diff
@@ -3117,18 +3082,49 @@ SbBool View3DInventorViewer::isAnimating() const
...
+        navigation->getAnimator()->startAndWait(animation);
+    }
+    else {
+        navigation->getAnimator()->start(animation);
```
wwmayer, line 3109: Law of Demeter

src/Gui/View3DPy.cpp
wwmayer: In the past it was possible to write this Python code:
```

# Manually move to the desired position and then call...
pos1 = view.viewPosition()
# Manually move to the desired position and then call...
pos2 = view.viewPosition()
# Manually move to the desired position and then call...
pos3 = view.viewPosition()

view.viewPosition(pos1, 20, 2000)
view.viewPosition(pos2, 50, 5000)
view.viewPosition(pos3, 10, 1000)
```

To achieve a similar effect now the code must be changed to:
```

# Manually move to the desired position and then call...
pos1 = view.viewPosition()
# Manually move to the desired position and then call...
pos2 = view.viewPosition()
# Manually move to the desired position and then call...
pos3 = view.viewPosition()

par = App.ParamGet("User parameter:BaseApp/Preferences/View")
par.SetInt("AnimationDuration", 2000)
view.viewPosition(pos1)
par.SetInt("AnimationDuration", 5000)
view.viewPosition(pos2)
par.SetInt("AnimationDuration", 1000)
view.viewPosition(pos3)
```
The downside is that this has side-effects because changing AnmimationDuration parameter affects other functions. Therefore I recommend the following changes:
* Revert this change
* Add an overloaded method `View3DInventorViewer::moveCameraTo(const SbRotation& orientation, const SbVec3f& position, int duration)`
* Use this function here and pass "ms" as third parameter.
* "steps" will still be used for the Python interface but not used any more
* In startAnimation() replace the bool with an int and remove the access to the user parameters

Rexbas: Sounds like a good plan, working on it.

src/Gui/NavigationAnimation.cpp
```diff
@@ -0,0 +1,167 @@
...
+    SbRotation deltaRotation = SbRotation(rotationAxis, value.toFloat() - prevAngle);
+    navigation->reorientCamera(camera, deltaRotation);
+
+    prevAngle = value.toFloat();
+}
+
+/**
+ * @param finished True when the animation is finished, false when interrupted
+ */
+void SpinningAnimation::onStop(bool finished)
```
wwmayer, line 161: `finished` isn't used. Either use `Q_UNUSED(finished)` or declare it as `[[maybe_unused]]`

src/Gui/NavigationAnimation.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+void NavigationAnimation::updateCurrentValue(const QVariant& value)
+{
+    if (state() == QAbstractAnimation::State::Stopped) {
+        return;
+    }
+    update(value);
+}
+
+void NavigationAnimation::onStop(bool finished)
```
wwmayer, line 42: Unused parameter: Use `Q_UNUSED` or `[[maybe_unused]]`

wwmayer:
With the PR there is still the issue that independent of the current camera location the duration to the new location is always fix which is weird for very small movements. I think we can leave this for now but should address it if there are more user complaints.

wwmayer:
> have you had any time to check whether the problems and crash you described

Yes, the crash is fixed and the behaviour of the speed slider is as expected. Thanks!

src/Gui/View3DPy.cpp
Rexbas: I added a commit that makes `view.viewPosition()` work as previously. If the duration is not passed to `viewPosition()` then it will be -1 by default. I did not overload `View3DInventorViewer::moveCameraTo(...)`  but instead I added a duration parameter with default value -1. `View3DInventorViewer::startAnimation(...)` now also has a duration parameter with default value -1. `View3DInventorViewer::startAnimation(...)` checks if the duration is valid, meaning more or equal to 0. If it is not valid then the duration is overwritten with the "AnimationDuration" parameter. So, If the duration is not passed explicitly in one of the previously mentioned methods then the duration will default to "AnimationDuration".

I also did not replace the bool "wait" because for `viewPosition()` it is desired to wait until the animation is finished but when selecting a standard view or a face on the navicube, waiting cause some side-effects (I don´t remember what exactly).

src/Gui/View3DInventorViewer.cpp
Rexbas: > it will be more complicated to run consecutive calls of viewPosition() with different speeds.

I guess 61e419b88ffac8cdb9f874363270327a0ab3ed3c also fixes this.

src/Gui/NavigationStyle.cpp
Rexbas: Resolved in 26ea9614e64b923008f29cbb06d4b6f45efc0b37

src/Gui/View3DInventorViewer.cpp
Rexbas: Never heard of it until now. Resolved in f28e3325812b2e3ce9c024db367dde0f31108382

src/Gui/View3DInventorViewer.cpp
Rexbas: Resolved in f28e3325812b2e3ce9c024db367dde0f31108382

src/Gui/View3DInventorViewer.cpp
Rexbas: Resolved in f28e3325812b2e3ce9c024db367dde0f31108382

src/Gui/NavigationAnimation.cpp
Rexbas: Resolved in 2fec778d31a44790854e6ea224d0a436664d8096

src/Gui/NavigationAnimation.cpp
Rexbas: Resolved in 2fec778d31a44790854e6ea224d0a436664d8096

Rexbas:
> With the PR there is still the issue that independent of the current camera location the duration to the new location is always fix which is weird for very small movements. I think we can leave this for now but should address it if there are more user complaints.

If users start to complain then maybe I can add another type of animation with fixed velocity and let the user choose which type of animation to use.

There are also some other small improvements I have planned to make regarding navigation animations, but they will come later in small follow up PRs

src/Gui/View3DPy.cpp
wwmayer: > I also did not replace the bool "wait" because for viewPosition() it is desired to wait until the animation is finished

OK.

Rexbas:
> Out of curiosity what exactly was the reason for this crash? https://github.com/FreeCAD/FreeCAD/pull/9446#issuecomment-1734316113

There was a segfault in the tests. The camera was nullptr but I am not sure why. The segfault was gone after I added a nullptr check in FixedTimeAnimation::update().

src/Gui/View3DInventorViewer.cpp
wwmayer: > I guess 61e419b also fixed this.

It's OK.

wwmayer:
> There was a segfault in the tests. The camera was nullptr but I am not sure why.

Ah, OK. Yes it can happen that the camera is a nullptr but I don't know the exact circumstances either.

PR #10888, opened 2023-10-02
Updated README.md to include more background and setup information

XianJianZhang:
Added important information about installation and setup process. Background information about the fundamental concepts and purpose of FreeCAD included for developers who don't know much about 3d modelers


yorikvanhavre:
I think the setup part is incomplete (also creating a python environment is not mandatory), but in any case such section should be placed elsewhere, in this proposal it sits in the middle of the description section...

Also the technical description of parametric geometry seems very precise and specific for a README... Wouldn't it sit better on the wiki? for example on https://wiki.freecad.org/Feature_list

PR #10892, opened 2023-10-02
[Sketcher] tool settings

PaddleStroke:
This is initial draft PR. For discussion and enable people to test as well.


luzpaz:
Spelling issues:  
Warning: src/Mod/Sketcher/Gui/Doc/DrawSketchHandler_parameterValueChanged.plantuml:55: fulfiled ==> fulfilled
Warning: src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h:261: hander ==> handler
Warning: src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h:272: clases ==> classes, clashes, cases
Warning: src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h:344: clases ==> classes, clashes, cases
Warning: src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetHandler.h:316: fulfiled ==> fulfilled

PaddleStroke:
@abdullahtahiriyo the point DSH is ready for review. I will start on the offset tool now.

PaddleStroke:
@abdullahtahiriyo I just pushed initial implementation of Offset. I managed to fix the various crashes and most quirks.
There is one important issue now, its handling the auto-switching of arcs ends.
I join the curves (create coincident cstrs) of the offset in jointOffsetCurves. But somewhere after the command finish the arcs are switch. And so it ends up as a mess.
I tried to solve before calling the jointOffsetCurves function but that doesn't help.

Can you please confirm at which step sketcher is doing this switching of arcs? And how I can do it before jointOffsetCurves.
Thanks

abdullahtahiriyo:
@PaddleStroke 

When you create arcs that are CCW, you should not have any problem. The sketcher is coded to consider all geometry CCW.

This was introduced by DeepSOIC some 10 years ago. I do not have all the details with me ATM.

Have you ensured that your arcs, as you create them, are CCW? If yes, then we will have to debug. If no, the offset tool needs to consider this.

PaddleStroke:
Thanks for your feedback. I have not paid attention to this so this is probably the issue. I will check and feedback if it solves the problem.

luzpaz:
Trailing whitespace:  
Warning: src/Gui/QSint/actionpanel/actiongroup.cpp:188:  Q_UNUSED(event); <-- trailing whitespace

PaddleStroke:
That's it, I wasn't giving true to getStartPoint of arcs. Now the bug is gone.

So it is now ready for review.

adrianinsaval:
> So it is now ready for review.

If so then you should change it's status, it's still marked as draft

PaddleStroke:
Ah yes but since Abdullah reported that it was crashing on his end. So I
haven't changed the status yet.

PaddleStroke:
@Syres916 fixed the crash on linux.
Also rebased and removed solve/draw from executeCommand of offset.
@abdullahtahiriyo ping to let you know it updated.

abdullahtahiriyo:
I will be working on this in the following days.

abdullahtahiriyo:
@PaddleStroke 

Please cherry pick 03f58f7396a3f947468c4c5cc013c532fd6ac80f into main branch (master) and make a separate PR for that one.

abdullahtahiriyo:
@PaddleStroke 

Please cherry pick into current master and PR separately:
0c72708480c2bc20b104774ed1bb8d4b3ca09c8c 
7ae1e53761e591aad79e7964c87ff8866c1af193
754747a0ab31f0179cb3c227039780816dface1f

PaddleStroke:
@abdullahtahiriyo its done : 
https://github.com/FreeCAD/FreeCAD/pull/11029
https://github.com/FreeCAD/FreeCAD/pull/11028

abdullahtahiriyo:
@PaddleStroke 

Thank you. Those are already merged.

No need to rebase this PR at this moment.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3247,7 +3247,7 @@ bool ViewProviderSketch::setEdit(int ModNum)
     if (!sketchDlg)
         sketchDlg = new TaskDlgEditSketch(this);
 
-    connectionToolWidget = sketchDlg->registerToolWidgetChanged(boost::bind(&SketcherGui::ViewProviderSketch::slotToolWidgetChanged, this, bp::_1));
+    connectionToolWidget = sketchDlg->registerToolWidgetChanged(boost::bind(&SketcherGui::ViewProviderSketch::slotToolWidgetChanged, this, boost::placeholders::_1));
```
abdullahtahiriyo, line 3250: @PaddleStroke 

Using ```boost::placeholders``` was acceptable when we wrote this (actually, I think I did write it). However, in the meantime Werner moved ```boost::bind``` to ```std::bind```. See commit:
68d22d8

Then the good fix here when rebasing, was to move to std::bind.

No need to change anything right now, as I have to revisit that code and I will do it. I just wanted you to know that ```boost::bind``` should not be used and ```std::bind``` should be used instead.

PaddleStroke: Got it.

PaddleStroke:
Let me know when you want me to squash or rebase.

abdullahtahiriyo:
@PaddleStroke 

There are many unformatted commits (and pre-commit is also failing in this PR).

I do not know if it is because some commits were "old" or because you do not have pre-commit installed.

I am doing the conversion now that I am at it, so there is no need to change anything to this PR now. But if you do not have the pre-commit hook installed in your git, please take some minutes to install it.

PaddleStroke:
I have pre-commit installed. It's complaining about our old commits that I did not overwrite.

PaddleStroke:
I have not cleaned the commit history so that you could start from 'known ground' and see the intent of each commit. But yes the commit history needs to be squashed / cleaned.

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetHandler.h
```diff
@@ -989,7 +989,7 @@ class DrawSketchDefaultWidgetHandler: public DrawSketchDefaultHandler<HandlerT,
         }
 
         /* This function give the focus to a spinbox and register to which it gave it.*/
-        void setFocusToLabel(int labelindex)
+        void setFocusToLabel(size_t labelindex)
```
abdullahtahiriyo, line 992: size_t should not be used as type of parameter, unless one really means a size. Here it is an index, not a size.

to solve the iwarning there are two options. If no negative indices are envisaged, then the parameter can be made unsigned. If negative indices are envisaged, then the changes are within the function, you may static_cast the size_t to int, considering that the index will never go to the overflow region by design.

PaddleStroke: I make no changes for now right ?

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetHandler.h
```diff
@@ -644,16 +644,7 @@ class DrawSketchDefaultWidgetHandler: public DrawSketchDefaultHandler<HandlerT,
          */
         void onHandlerModeChanged()
         {
-            if constexpr (std::is_same_v<StateMachines::OneSeekEnd, SelectMode>) {
```
abdullahtahiriyo: Because Point tool uses OneSeekEnd and Point tool does require a behaviour different than default, here the choice was to remove the default behaviour, so as not to explicitly provide a behaviour different than default for point.

If another tool with OneSeekEnd is created, and uses the default behaviour, it will have to implement it, as it is not there anymore. This is inconsistent with other state machines.

There are two approaches: 
(i) we do not provide default behaviours at all and each handler will have to implement its own.
(ii) we provide consistent default behaviours for several state machines.

If we take (ii) is because we assume most tools would use a default behaviour, and it also helps as documentation of how the function should normally work. However, if we take (ii) the default behaviour needs to be consistent for all state machines and any tool using different shall provide its own.

If we take (i) is because there is not a "logical default pattern".

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetHandler.h
abdullahtahiriyo: Right. I will come back to you with another branch, and we discussed it from there. 

It is not worth that you invest time in rebasing and correcting this branch, as I need to do it in parallel to add some other changes. 

I am leaving the comments in this PR, as it is convenient so that we understand each other.

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetHandler.h
PaddleStroke: I think that actually it is better to do (i) and not provide defaults.
While it sounds like a good idea, in my experience it is not ideal:
- First, having writen most of the DSH before, there was actually very little use of the default if I recall correctly. We have not so many DSH, so in the end there wasn't two really similar. If we had 100 DSH it would make sense I guess.
- It increase complexity, making debugging harder. For the offset tool for example I had a bug that took me a long time to identify (which is the OneSeekEnd default I removed).

So yes I would rather remove the defaults.

abdullahtahiriyo: I am ok with that. But then there is a second step, where we have to decide from which functions we remove the defaults. It may be from all functions, or it may be from some.

I will be posting a working branch, as I have finished with the toolwidget signaling block. Then we can discuss over it.

src/Mod/Sketcher/Gui/TaskDlgEditSketch.cpp
```diff
@@ -80,6 +89,18 @@ TaskDlgEditSketch::~TaskDlgEditSketch()
     if (it == Content.end()) {
         Content.push_back(SolverAdvanced);
     }
+
+    connectionToolSettings.disconnect();
+}
+
+void TaskDlgEditSketch::slotToolChanged(const std::string& toolname)
+{
+    bool hidden = toolname == "DSH_None" || toolname == "DSH_Point";
```
abdullahtahiriyo, line 98: Special behaviour for the widget should be dictated by the tool, rather than hard-coded here.

This a kind of anti-pattern and we should always strive to avoid it.

src/Mod/Sketcher/Gui/TaskSketcherTool.cpp
```diff
@@ -0,0 +1,90 @@
...
+TaskSketcherTool::~TaskSketcherTool()
+{}
+
+void TaskSketcherTool::toolChanged(const std::string& toolname)
+{
+    // TODO: Implement a factory here to get an appropriate widget from the toolname
+
+    // At this stage, we add a Default tool widget for all tools with a defined name, but this needs
+    // to change
+    if (toolname != "DSH_None") {
```
abdullahtahiriyo, line 65: This is even a better example of the anti-pattern.

It is ok to provide some special behaviour for "DSH_None", as it will not change over time with the introduction of each tool (plus there is not a tool that is DSH_None, but rather the absence of a tool).

However, for the rest, this if-else branching is not the way to go. We want that the special behaviour of a tool resides with the tool. The code here should be generic and achieve its function while relying on information provided by the tool.

I have considered several options. The most generic would be a full fledged widget settings factory. However, because at a time there is a single tool active (at maximum), it was probably overkill/YAGNI. 

Then, I settled for providing an NVI interface for DSH Handlers, and exposing a minimal configuration interface in view provider.  This has a relatively low footprint by comparison, and it is generic enough to move the whole widget to whereever within Sketcher Edit mode (dockable independent dialog,...). 

At the same time, this API can provide additional information, such as the specific widget (for cases where the default widget may not be appropriate), whether a widget should be visible or not, icons/title texts,...

I will post the branch.

(i) A widget settings factory
(ii)



 a kind of anti-pattern and we should always strive to avoid it.

abdullahtahiriyo:
Moving forward. Look at this branch:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/commits/pw_tmp

We are aiming at agreeing on the first three commits: 
e07a57f8ed2eea7ef3383096decedd788e79c08e 
50aafa73bf8ca7761a720a94f3442d56ef508202
d980e354d9e5fbe3e6a3f0b337ba3cc8335c07d0

These provide the default widget and the framework for providing a tool widget in the taskbar. I think they are now totally separated from the rest of the PR. Then, I think we should merge these three in a first PR and continue with the rest.

The code is essentially what we had reworked in the last year, with the std::bind modification and the small changes in the last commit to fix the issues raised above.

Here I would like you to:
(1) Review these three commits.
(2) Do a minimal testing to ensure that the framework is still working as expected.
(3) Come back with comments or prepare a separate PR with these three commits for merge.

Now I move to the next topic, the new DSH Architecture.

src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h
```diff
@@ -0,0 +1,1003 @@
...
+    }
+
+    //@}
+
+protected:
+    // The initial size may need to change in some tools due to the configuration of the tool, so
+    // resetting may lead to a different number than the compiled time value
+    int initialEditCurveSize;
+
+    std::vector<Base::Vector2d> EditCurve;
```
abdullahtahiriyo, line 989: @PaddleStroke 

Here I have a question for you.

At the very beginning, we mirrored the approach of the preexisting DrawSketchHandlers, in which often an EditCurve was provided, which would be drawn at each step based on Vector2D.

Later on, we introduced the ShapeGeometry and the DrawShapeGeometry, where essentially we add Part::Geometry to ShapeGeometry, and it gets Drawn  via DrawShapeGeometry. This way, avoiding to have to deal with the Vector2ds of EditCurve, and leveraging PythonConverter to generate all commands necessary at the end.

My question is: Could you take a look at all the DSHs we (mostly you) have written and evaluate whether we still need this "EditCurve" interface?

What I seek is:
If we do not need it, then we can drop it, including from the template parameters. We set for a more modern, less error-prone framework. What you see during construction is what you get, because it is actually based on that very same geometry.

If we do need it, then we keep it.

Please, let me know.

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetHandler.h
PaddleStroke: I think we can remove them most for now. And if need be we can reimplement the one that make sense.

We can already have a look between the point DSH and the offset DSH, which are IIRC the only 2 tools which are OneSeekEnd. And see between these two tools for what function we use the default. I would guess that its stuff related to the construction mode combo box. Then we can keep these.

src/Mod/Sketcher/Gui/TaskDlgEditSketch.cpp
PaddleStroke: Got it.

src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h
PaddleStroke: For very basic tools, like line and circle, I seem to recall that it was still used. As those geometry are very simple. For example in the line tool instead of having a startPoint and endPoint, the editcurve was used to store the points.
But I think we can do without it. It will indeed probably be cleaner without it. We just make geometries and preview them.

PaddleStroke:
The changes you made look good. I'm building and will test to make sure its ok. If so I'll make a PR

PaddleStroke:
Stills works as expected and code is ok so here's the PR : https://github.com/FreeCAD/FreeCAD/pull/11048

abdullahtahiriyo:
To test the framework, it would be great to have something relatively more complicated, like the Rectangle tool.

**Would you implement the Rectangle DSH on top of that other branch?** (You can clone that branch, add the rectangle DSH, test, push it to your repo, then I can cherry-pick the commit).

If you do, it does not need to be from the beginning with the on-screen parameters. The one we had, adapted to the few changes of the branch can be a good starting point, and would already help me with testing. 

Then you may add the on-screen parameters on a second commit. When I see the common code with the point DSH, it would also help me see how the framework can be improved to facilitate porting other tools to the on-screen parameters.

PaddleStroke:
Hmm the rectangle tool is very big and complex. Especially since after your final review last year I also added 3 points rectangles and centered 3p rectangles. Which made the tool even more complex.
So I was thinking that this tool could be a single PR by its own.

But I can do it if you want. I just want to make sure that we avoid feature creeping in this PR. We can always validate this first, then on subsequent PR modify the framework if necessary.

Just to make sure I got it right, you want me to make a branch starting from your branch with the stuff that just merged right?

PaddleStroke:
Also rectangle DSH now also needs the angle datum label (for the 3p rectangle mode) which is not implemented yet.
My plan was to implement this editable angle datum label with a line tool as its simpler. Before moving to the rectangle tool.
One more thing, I am not very clear about what you included in your pw_tmp branch. Did you include all the changes from my branch that you all squashed together ?

abdullahtahiriyo:
@PaddleStroke 

Yes, the rectangle tool is big and complex, and that makes it a very good candidate for testing the framework. At the same time, it is restricted to a single file, which does not interfere with the rest of this PR. 

It can be merged together or separately as a PR, whatever is more convenient for us. Whatever the decision, adding it now would have helped debug the framework. The great thing about these DSHs is that they can be made just one commit at the end of the branch. Then, merging together or separately is a matter of pushing a PR with or without that commit, and that commit separately (where without). I do not see any problem of feature creeping while commits of framework and tools are kept separated (which is a must in any case).

**The problem we are aiming at solving is to provide a stable framework that can cover all use cases, while avoiding that fixing the framework to support DSH A breaks it for DSH B**. While bugs will likely persist in the framework after merging, which would require fixing, I would not like to turn every new DSH addition into a Framework redevelopment. The risk associated with pushing a not fully tested framework is exactly that. 

Moreover, as I move to other tasks, periodically coming back to framework review will affect my ability to concentrate on those other tasks, or I may prioritise those other tasks leading to delays in merging the DSHs.

I do understand that you do not want to invest yourself into adapting DSH tools to the on-screen system until you see the framework merged. However, there needs to be a trade off between your time investment and mine. None of us wants to put this task on hold again.

From my point of view, the most efficient way of doing this task, while minimising your risks, is by keeping adding DSH tools one by one to a development branch (at least those DSHs that do not require modifications beyond their single DSH file). After every DSH tool addition, we can check whether any modification is necessary for the framework, and how to best put it in place, while testing it for all the previously added tools. If we modified the framework, new DSHs will be added taking into account the modification (thereby reducing your risks). This may be more relevant at the beginning, as the probability of changes is higher at the beginning. At the end we would ideally end up in a branch with a few commits for the framework, and one commit per DSH tool. 

We merge the framework and the the tools separately, or if you do not have PR objectives, if everything is working fine, we could merge the branch at once. The important bit for both of us is that there is no meaningful additional work involved in doing it one way or the other.

This strategy is also best for those who help with testing independently. A user built this PR and basically noted, apparently deceived or disappointed, that the only thing that is in is a DSH Point. Their testing is not that efficient in small separated chunks. More importantly, if we move to a different task, the result of their testing are not as meaningful, as instead of a quick hot fix before merge, it may mean a ticket macerating in GH issues, or a PR longing for being merged (specially if the testing result says we need to do changes to the framework and all the tools already merged...).

I will have substantially more time available this week and at least part of the next week. Now it is a good opportunity to move this forward.

Replying to some of the questions that keep coming:

My branch is this PR, where I have reordered the commits per topic and squash them, minus the issues I have commented on this PR, plus the additions only to what is now already in master. So basically it is the same, or should be the same. The fact that commits were not clang formatted, make git differences less visual and difficult to accomplish. I hope not to have removed any meaningful bit. Hopefully, as we both use pre-commit now, it is a problem of the past.

If you prefer to do the Line DSH first, I am fine with that too. If you need the editable angle datum, feel free to implement it and add it as a separate commit. With this strategy, we should be able to test the editable angle datum, the framework and the Line DSH together. When the editable angle datum works fine, we apply the same strategy as in this PR, we cherry-pick the commits of the editable angle datum on top of master and we PR them separately. The idea is that the development branch is basically the DSH Framework with one commit per tool at the end. While we keep that rule, development and review will be fine.

For all clarity, the branch I added above can be fast-forward rebased onto master. I think it is the easiest for development is to not even do that. Just keep adding to that branch the commits you need for your tools and modifications to the framework (tools and modifications to the framework in separate commits).

PaddleStroke:
I am currently working on the line DSH. Fixing bugs here and there. Once this is done I will move to the rectangle tool then.

PaddleStroke:
Though I don't really understand your point.
You realize that in the past in the paddle_widget branch, we had all the tools and they were all working (mostly) correctly right? Which means that we already know that the framework is not going to need big changes. The only change is to move from the widget spinbox to the on-screen spinboxes.

I do not particularly have PR objectives, other than I don't want to end up with doing all the work, having it ready to merge and then you going away for one year. So if possible I prefer to merge bit by bit as soon as it's ready. So that if at some point you become unavailable for a long period of time, we can still carry on.

abdullahtahiriyo:
Regarding paddle_widget, not all tools were working properly if I remember correctly, I think OpenBrain did test at least some (maybe it even was the Rectangle DSH), and had remarks about the functionality (not referring about code). There is a question whether to address these, the framework needs adaptation, or whether the DSHs themselves were to blame (and the framework was just fine).  Then, there is the addition of the on-screen spinboxes, which was not tested back then.

If the framework passes both tests and require no big changes, then it will be merged soon. I want to avoid continued redevelopment of the framework after merge. That is why I am proposing the present approach. I am also not planning on requiring all DSHs to be adapted before merge, but some representative ones. 

The thing is that probably, when those work and the framework is tested and can be merged, doing the rest is not going to be much work for you, as the framework will be developed enough so that adapting them is easy (save for some special ones that require changes to SketchObject).

I do not plan to be overridden by a truck in the near time. However, the sooner things are in place, the lower the probability that it happens before I merge it.

PaddleStroke:
It's fine by me. So I'm on the line DSH, fixing things and will move to rectangle tool after. Hopefully I can wrap it today.

abdullahtahiriyo:
Thank you.

abdullahtahiriyo:
Thank you.

Please cherry-pick 5fcc4805d676ef7493807a3ac1b9f39ab7d98dc0 , consider the comment on it and modify/squash, and make a separate PR on that one. I can merge that one directly and we keep focusing here on the framework/tools.

I have forced pushed the branch:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/tree/pw_tmp

All your changes I have just reposition or directly squashed. Consider the comments I made on your branch. Aside, I have added some documentation to DrawSketchDefaultHandler as I go reviewing the code, and I have removed the editcurve template parameter as we discussed some post up here. I also corrected the authoring of the Offset DSH that was inadvertently changed during squashing.

BTW, very nice on-screen angle edit controls.

In a quick test, DSH Line, point-length-width, I had some issue that TAB would not switch control once. I could not reproduce in further very limited testing. Probably no action required from you ATM, but just so that we both know.

abdullahtahiriyo:
Any new development (e.g. Rectangle DSH), please do it on top of the new forced-pushed pw_tmp.

abdullahtahiriyo:
BTW, I have added a couple of methods to DrawSketchHandler for  convenience clearning editing curve and markers, see: 32f4c26d37e126126782ac62414c35c74bd5d1bd

PaddleStroke:
I have opened the PR for https://github.com/FreeCAD/FreeCAD/commit/32f4c26d37e126126782ac62414c35c74bd5d1bd

On my side I got your forced pushed branch and I am starting to work on it now. (so I will push here : https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp)

PaddleStroke:
Is `adaptDrawingToLabelChange` really necessary at all ? Because we are doing a `enforceWidgetParametersOnPreviousCursorPosition` just before anyway.
So what is the point of this function I cannot remember. Can you remember why this was implemented ?

PaddleStroke:
Erratum, `adaptDrawingToLabelChange` is in fact necessary in the offset tool as we can't do doEnforceWidgetParameters due to complexity of overriding the mouse position. But this is an exception I guess.

PaddleStroke:
I have added the rectangle DSH but it is NOT debugged / ready.
I have done adapting the code to the on-view parameters. But there are still a lot of bugs.

I'm posting already because the code adaptation is done so it should serve the purpose that you had in mind. But as I said, the architecture is the same as the line tool. Just much more modes that adds a lot of complexity.

abdullahtahiriyo:
@PaddleStroke 

Thank you.

I have just commented a minimal change on the Angle edit datum PR. It can be merged just afterwards. Sorry I did not realise that yesterday (including the big banner of Lint). There are couple other Lints where I do not quite agree, as changing the label, changes the object. You may suppress them with // NOLINT if you wish.

You raise an interesting question regarding the ```adaptDrawingToLabelChange```. That mirrors the equivalent of the old widget. For the old version, I do not fully remember, but I think it had to do with whom was requiring the change (and the corresponding automatisation). I am currently reviewing one step before that, at ```DrawSketchDefaultHandler```. I will assess it when I arrive to ```DrawSketchDefaultHandler```.

I appreciate that you invest the time in debugging the Rectangle DHS. I think it has the potential to test the framework.

Regarding the refactoring within Rectangle DSH. I am not sure if you mean refactoring ```DrawSketchHandlerRectangle``` or refactoring the framework. If you mean the former, it appears very reasonable. Feel free to do and see how it goes. I am not looking exactly at the code. However, if you could find a pattern within move1, move2, move3 (I do not know if this makes sense or not), and refactor the generic part out, that would be very interesting. In a further development, if part of this common generic part were of interest for other tools, then it could go up to the framework, and that would be quite a code reuse win.

I aim to have reviewed the framework before the end of this week (stage 1). Then, I think we would do good to look into the couple of handlers we have implemented, and see how to simplify/refactor the code (stage 2). If we manage to do it right, it has the potential to simplify your rework other DSHs.

PaddleStroke:
image

The view gets really crowded pretty fast. This looks bad and confusing I think.
I see several solutions : 
- When dsh goes to next mode : 
-> If a label is set : the spinboxes are removed, leaving only the soDatumLabel.
-> If a label is NOT set : the spinboxes are removed, AND the soDatumLabel is also removed.

What do you think?

PaddleStroke:
There is a problem with the construction mode combobox. Changing the mode does not go to the selected index, but iterate to the next mode. See in DrawSketchDefaultWidgetHandler.h : 
```
        void adaptDrawingToComboboxChange(int comboboxindex, int value)
        {
            Q_UNUSED(comboboxindex);
            Q_UNUSED(value);

            if constexpr (PFirstComboboxIsConstructionMethod == true) {

                if (comboboxindex == WCombobox::FirstCombo
                    && handler->ConstructionMethodsCount() > 1) {
                    handler->iterateToNextConstructionMethod();
                }
            }
        }
```

What we need is something like :
```
         void adaptDrawingToComboboxChange(int comboboxindex, int value)
        {
            if constexpr (PFirstComboboxIsConstructionMethod == true) {

                if (comboboxindex == WCombobox::FirstCombo
                    && handler->ConstructionMethodsCount() > 1) {

                    handler->setConstructionMethod(value);
                }
            }
        }
```
But it needs some thinking as ConstructionMethodMachine has currently no way to set by index.

PaddleStroke:
Please note that I am making changes to the framework. I will push changes later today

PaddleStroke:
@abdullahtahiriyo I am now done. Rectangle DSH is finished. It's quite complex so there might be a bug escaping my testing but overall seems good.
I also added some nice icons : 
image

There is still the bug I mentioned before about changing the mode from the combobox.
Please let me know if you have any question.

PaddleStroke:
@abdullahtahiriyo I have pushed some more changes cleaning the framework. At this stage I don't know if it is worth going through commits one by one. Especially since there are some changes that I undo after as they are no longer necessary. For example I just removed beforeFirstMouseMove that I introduced before.

The framework is much simpler now I think. So perhaps it is best that you squash my commits and review the thing like this. I will let unsquashed in case you want to look individually.

To summarize the standing issues : 
- adaptDrawingToComboboxChange iterate the mode instead of going to the correct mode.

abdullahtahiriyo:
@PaddleStroke 

Ok. Thank you. I will look into what you have done, and what you have written here, and will come back to you.

I do not think it is worth at this moment that you try to catch more (possible) bugs on DSH Line and Rectangle. The ball is on my roof now.

abdullahtahiriyo:
One observation:

```
void SketcherToolDefaultWidget::setCheckboxIcon(int checkboxindex, QIcon icon)
{
    if (checkboxindex < nCheckbox) {
        getCheckBox(checkboxindex)->setIcon(icon);
    }
}
```

setIcon is:
```
setIcon(const QIcon &icon)
```
Meaning that it has no move semantics (it is not QIcon &&icon).

That means that you are making a copy of QIcon when calling ```setCheckboxIcon``` , which would be useless, as setIcon will make its own copy inside.

Passing by value, as you have done, could make sense if setIcon had move semantics, and you could do setIcon(std::move(icon)). 

When we are not using an API with move semantics, and we want to avoid copies, input-only parameters should be ```const &```, unless they are basic types (int, bool, float). For input-output parameters we use a non-const reference (including cases where the parameter is output only, but an argument to the function).

No need to change anything, I will do will reviewing.

abdullahtahiriyo:
Please:
8128e3eb9d993d1ba761f07c4c2dfdd810c16a65
1727c153876e779ab72e5fcf340a3c28b79fd317
5a976ade2a24a2e6f996fede4c28fbeef532ff90

(I think those are all the new commits touching EditableDatumLabel)

in a separate PR.

When the checks are passed on the PR, please do look into any lint warning. I have been talking to Werner and the policy we should aim for is that of no lint warning on new code. Thank you.

abdullahtahiriyo:
After that separate PR, please do another separate PR for f21aaa3b7cace570eab81992c823326c01f9ec25 (SketcherToolDefaultWidget). This is because we have already merged that part and I think it is best to concentrate on Framework/Tools in the current effort.

When doing so, please do look into all the lint warnings, old and new ones. If you have doubts about how to solve them, let us know. Thanks.

abdullahtahiriyo:
> Writing :
> `dHandler->constructionMethod() == ConstructionMethod::ThreePoints`
> All over the place is cluttering the code. It would be good if we could just do :
> `isMethode(ThreePoints)`

There is a method ``ìsConstructionMethod``` in DSDH and DSDWH:

```
    virtual QString getCrosshairCursorSVGName() const override
    {
        if (isConstructionMethod(ConstructionMethod::OnePointLengthAngle)) {
            return QString::fromLatin1("Sketcher_Pointer_Create_Line_Polar");
        }
        else {
            return QString::fromLatin1("Sketcher_Pointer_Create_Line.svg");
        }
    }
```
I do not think it is possible to remove the ```ConstructionMethod'``` in the argument in a clean way and still checking types (which is the whole point of the using a enum class). I think I did read that in some situations it is possible to use a using declaration to omit the scope, but I no longer have the details. I would have to fetch that reference again. I will try to find it back...

abdullahtahiriyo:
BTW, testing the combo change with the rectangle I got into a situation where the tool produced a rombus/romboid. Is this intended?

abdullahtahiriyo:
> The view gets really crowded pretty fast. This looks bad and confusing I think. I see several solutions :
> 
>     * When dsh goes to next mode :
>       -> If a label is set : the spinboxes are removed, leaving only the soDatumLabel.
>       -> If a label is NOT set : the spinboxes are removed, AND the soDatumLabel is also removed.
> 
> 
> What do you think?
> 
> Edit: I have implemented like this and it is much much better.

I think it is good for view clutter. It may indirectly solve a different problem. When the parameters were in the taskbox, one could SHIFT+TAB into a parameter that belonged to a previous state in the state machine. This caused problems of synchronisation, that a user would expect already entered parameters of a future state to be taken into account when painting. Now, it is not possible to go back to change previous parameters (of a previous state of the state machine), leading to the perhaps less convenient facility in 5% of the times, but more straight-forward simpler behaviour (the user would have to cancel the execution if need come back for a parameter of a previous state machine and restart)...

PaddleStroke:
```
dHandler->constructionMethod() == ConstructionMethod::ThreePoints
dHandler->isConstructionMethod(ConstructionMethod::ThreePoints)
```
That doesn't help much yet. But renaming the function isMethod and if we can call without dHandler perhaps.

abdullahtahiriyo:
> That doesn't help much yet. But renaming the function isMethod and if we can call without dHandler perhaps. However this is not very important. Just a small code clarity if possible.

We are at a different class, therefore I do not see how we can get away without the instance of the object that has the method. 

The risk associated to reducing the number of characters in the name is that of code clarity. This is particularly important as isConstructionMethod is one the several parent classes of the parent class template of the parent class template. Semantically isMethod does not give much meaningful information... we do not even know we are talking about a construction method.

I am not opposed to changing the function name to something shorter that is still meaningful... will think about it...

abdullahtahiriyo:
> Yes it is. It's a new mode (3 points rectangle) that has been asked while you were away. So we have 4 construction methods now.

Ok.

abdullahtahiriyo:
> However the complexity of this DSH is now even more complex due to the onViewParameter that require positioning. Many functions are over 200 line. Refactoring of certain functions could help. For example onHandlerModeChanged could call onChangedToMove1, onChangedToMove2 and so on. Please comment.

I have tried to understand this one without success. That function is not that big:

```
template<>
void DrawSketchHandlerRectangleBase::ToolWidgetManager::onHandlerModeChanged()
{
    switch (handler->state()) {
        case SelectMode::SeekSecond: {
            if (dHandler->constructionMethod() == ConstructionMethod::ThreePoints) {
                onViewParameters[WLabel::Fourth]->setLabelType(Gui::SoDatumLabel::ANGLE);
            }
            else {
                onViewParameters[WLabel::Third]->setLabelType(Gui::SoDatumLabel::DISTANCEX);
                onViewParameters[WLabel::Fourth]->setLabelType(Gui::SoDatumLabel::DISTANCEY);
            }
        } break;
        case SelectMode::SeekThird: {

            if (dHandler->constructionMethod() == ConstructionMethod::Diagonal
                || dHandler->constructionMethod() == ConstructionMethod::CenterAndCorner) {

                if (dHandler->roundCorners) {
                    onViewParameters[WLabel::Fifth]->setLabelType(Gui::SoDatumLabel::RADIUS);
                }
            }
            else {
                onViewParameters[WLabel::Sixth]->setLabelType(Gui::SoDatumLabel::ANGLE);
            }
        } break;
        case SelectMode::SeekFourth: {
            if ((dHandler->constructionMethod() == ConstructionMethod::ThreePoints
                 || dHandler->constructionMethod() == ConstructionMethod::CenterAnd3Points)
                && dHandler->roundCorners) {
                onViewParameters[WLabel::Seventh]->setLabelType(Gui::SoDatumLabel::RADIUS);
            }
        } break;
        default:
            break;
    }
}
```
Perhaps it was big and then it changed. I also do not understand the reference to move1 move2.

Please let me know if this comment is no longer relevant, or try to clarify if it still is.

PaddleStroke:
This comment is no longer relevant you can disregard it. Because before I improved the framework to the new on-screen labels, there was 3 places where the spinbox position was done. Resulting in a lot of complexity. But with the latest commits its no longer true.

Btw it seems from your onHandlerModeChanged function that you are not currently on the latest commit. Because the 
`onViewParameters[WLabel::Fourth]->setLabelType(Gui::SoDatumLabel::ANGLE);`
have moved to the configureWidget in latest commit.

Just to make sure you have not missed some commits.

abdullahtahiriyo:
> Btw it seems from your onHandlerModeChanged function that you are not currently on the latest commit. Because the `onViewParameters[WLabel::Fourth]->setLabelType(Gui::SoDatumLabel::ANGLE);` have moved to the configureWidget in latest commit.
> 
> Just to make sure you have not missed some commits.

Indeed. Thank you for the heads  up. I will try to track that one down.

PaddleStroke:
The last commit is : e8b434778038daada23685a4bf650fdca78b5ec1
The commit that moved setLabelTypes to the configureToolWidget is 76d274c1c35a10beeab02e24c0fb925c9585f825
See in : https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp

abdullahtahiriyo:
@PaddleStroke 

Thanks. I had several changes of my own and in the meantime I did a branch comparison restricted to the Sketcher/Gui directory and brought the code that was missing.

I have also forced-pushed the branch (so now you should see there reflected the changes).

I am still reviewing DSDWH. Some parts I need to revisit, as I was missing some code...

abdullahtahiriyo:
@PaddleStroke 

I suspect that the requirements of on-screen parameters and in-widget spinners may be different.

Q1: Does any of your tools retain a spinner parameter in the widget (i.e. not on-screen)?
Q2: Does any of your tools retain a spinner parameter in the widget which depends on the mouse position (for example, as former x coordinate/y coordinate)?
Q3: Do we reasonably anticipate mouse-position dependent parameter in the widget?

PaddleStroke:
Q1: Some tools need widget spinboxes : number of sides of polygons. Number of copies of array / polar array.
Q2: There is at least one tool for which the mouse position is needed for a widget spinbox : the scale tool. The seek second of scale tool is determining the scale factor if you use the mouse. But it does not represent anything on the screen.
Q3: So yes.

abdullahtahiriyo:
Thank you.

Is the scale tool already made? Is it possible for you to add it on top of my branch?

https://github.com/abdullahtahiriyo/FreeCAD_sf_master/tree/pw_tmp

PaddleStroke:
I can port it to the new framework as ultimately the goal is to implement
it.
Though after thinking about it, maybe it will be better as a on screen
parameter anyway.

abdullahtahiriyo:
Generally speaking, it is highly unlikely that exposing the internals of a class is a sound decision. Here, data members where changed from private to public:

```
    public:
        Base::Vector2d prevCursorPosition = Base::Vector2d();
        Base::Vector2d lastWidgetEnforcedPosition = Base::Vector2d();

        bool firstMoveInit = false;
        int labelIndexWithFocus = 0;
```

This leads to quite a lack of encapsulation, and the possibility to blur the responsibilities of each class. That portion belongs to the ToolWidgetManager. One of the reasons to do that is so that the handler can access the internals of the inner class (TooWidgetManager), as in here:

```
    virtual void mouseMove(Base::Vector2d onSketchPos) override
    {
        if (!toolWidgetManager.firstMoveInit) {
            toolWidgetManager.firstMoveInit = true;
            toolWidgetManager.setModeLabels();
        }
        toolWidgetManager.enforceWidgetParameters(onSketchPos);
        updateDataAndDrawToPosition(onSketchPos);
        toolWidgetManager.adaptParameters(onSketchPos);
    }
```

However, how ToolWidgetManager should react to a first move is nothing the handler should decide, but ToolWidgetManager. The handler should not even know that there are "labels". Showing the labels is a responsibility of the ToolWidgetManager.

There are several ways to implement this. One is by creating a function in the ToolWidgetManager ``m̀ouseMoved```, and calling it here on every mouseMove, then the code can be moved the ToolWidgetManager, and the involved data members can be back to ```private```.

ToolWidgetManager controls the handler to make the widget added functionality happen. The handler is as agnostic as it can get from how the widget makes its functionality happen. It is only aware that there is something that controls it, for which it needs to provide inputs and eventually obey orders. That ensures encapsulation of ToolWidgetManager and provide a clear boundary for where the responsibility belongs.

The other example is ```onModeChanged()```, which also has a similar solution of moving the functionality to the toolWidgetManager, eventually creating new functions as needed.









One of the reasons

abdullahtahiriyo:
> Though after thinking about it, maybe it will be better as a on screen parameter anyway.

If it is going to be an on-screen parameter, it is up to you if you want to work on it now or not. It is not a requirement you do. I would be equally happy whatever you do, as it does not impact framework testing. At the same time, I have nothing against separated commits at the end for specific tools, as many as you feel like (one might hit an edge in the framework, although after the ones we have, it is less likely).

In any case, if you feel like it, I would rather have one DSH that has a parameter entry that is not mouse position dependent. That one would increase the coverage of testing (as we have none ATM). I do not anticipate problems, but one never knows.

Regarding the on-screen parameters vs mouse-dependent parameters,  I think that on-screen parameters have different requirements with regard to reacting to changes and repainting than (mouse related) widget parameters. However, the more I see how you have written the tools, and how you make use of the on-screen parameters, the more I think that a mouse-dependent parameter in the widget may be a YAGNI (because no matter what, this mouse dependent relationship is always going to be clearer on-screen). At the same time, given the complexity of the framework, it is unlikely to be a better moment than now to create an infrastructure to work with both, on-screen parameters and mouse-dependent widget mandated parameters. Here I hear you. Give it a thought. If you reasonably think we need it, we go for it. If you do not actually see a use-case and think it is a relic from the past, then perhaps it is better not to go for it.

abdullahtahiriyo:
One minimal unrelated detail I have realised. The widget of the Offset tool does not have the text (U) (V) for the shortcuts (unlike the rectangle DSH). No need to bring the change right away. When you start coding something for the branch, you may add it.

abdullahtahiriyo:
This is a real good candidate for a unique_ptr: ```std::vector<Gui::EditableDatumLabel*> onViewParameters;```. I will be bringing the change to the framework.

abdullahtahiriyo:
We have a tiny little problem of definitions. We are using WLabel for both on-view parameters and widget spinbox parameters. Until now, as only on-view parameters are used, the problem has not revealed itself. Even when there is no problem, the terminology is misleading (using the word label to refer to both). I will be working on this...

PaddleStroke:
After first thinking about it, I think that we can get rid of fistMouseMove
function. Instead we can initialize the labels inside the
configureToolWidget.

Initially I made this firstMouseMove so that the spinboxes and labels are
initialized only when the mouse moves on the canvas.
However thinking about it, I think it is best to initialize the labels on
the view at the origin, so that user can type values directly without
having to move the mouse.

PaddleStroke:
I think all the parameters that depend on mouse will be on screen. Scale is
the only example where the question could be asked. But after thinking
about it, you are correct is will be clearer on view that on the widget.

abdullahtahiriyo:
> Initially I made this firstMouseMove so that the spinboxes and labels are initialized only when the mouse moves on the canvas. However thinking about it, I think it is best to initialize the labels on the view at the origin, so that user can type values directly without having to move the mouse.

The only issue we might hit here is whether the representation would be good when the origin is not on the screen. It might even be that some tools accept it graciously while others do not.

Generally, I think it is good to have a mechanism to be able to react to a first movement in the framework. Whether to use it or not in specific circumstances is then open. 

I am restructuring the framework. I hope to finish between today and tomorrow.

abdullahtahiriyo:
@PaddleStroke 

In ```setModeLabels()``` what is the relevance of toolWidget in:
```onViewParameters[i]->startEdit(0.0, toolWidget);```

I see that it is an event filtering object. But what is exactly the motivation of using toolWidget? 

A related but different question, in a case where somebody writes a custom DSH, with on-screen parameters, but without any widget, who should be this class?

**Comment by PaddleStroke** [comment]
The motivation to use tool widget is to use tool widget event filtering for
the keys. You know there's a class there that you made so that M key passes
through unless user holds it for long.

Now having said that, I think we should probably move this filtering class
to dsdwh as the tool widget does not have the focus at all time as it did
before.

**Comment by PaddleStroke** [comment]
I think I have the solution.
We can use the middle of the 3dview in configure tool widget

**Comment by PaddleStroke** [comment]
To clarify my previous message, as I was on the phone I didn't make it very clear. So just to make sure there was no misunderstanding. It was about this` if firstMoveInit` block : 
```
    virtual void mouseMove(Base::Vector2d onSketchPos) override
    {
        if (!toolWidgetManager.firstMoveInit) {
            toolWidgetManager.firstMoveInit = true;
            toolWidgetManager.setModeLabels();
        }
        toolWidgetManager.enforceWidgetParameters(onSketchPos);
        updateDataAndDrawToPosition(onSketchPos);
        toolWidgetManager.adaptParameters(onSketchPos);
    }
```
I meant we can remove this if. So we have :

```
    virtual void mouseMove(Base::Vector2d onSketchPos) override
    {
        toolWidgetManager.enforceWidgetParameters(onSketchPos);
        updateDataAndDrawToPosition(onSketchPos);
        toolWidgetManager.adaptParameters(onSketchPos);
    }
```
And in configureToolWidget we add : 
 `toolWidgetManager.setModeLabels();`
And in `setModeLabels` we set the spinboxes points as 
```
p1 = origin
p2 = middle of the camera view projected on sketch plane.
```

**Comment by abdullahtahiriyo** [comment]
> And in configureToolWidget we add : `toolWidgetManager.setModeLabels();` And in `setModeLabels` the DSH choose the spinboxes points. Most likely most DSH will do :
> 
> ```
> p1 = origin
> p2 = middle of the camera view projected on sketch plane.
> ```

Ok. No hardcoded origin in the framework. Just left to the implementation of the specific handler. That is ok.

**Comment by abdullahtahiriyo** [comment]
> The motivation to use tool widget is to use tool widget event filtering for the keys. You know there's a class there that you made so that M key passes through unless user holds it for long.

Yes. The motivation was to allow to enter units in in-widget parameters. This worked and most probably still works (although we no longer have them, and it appears we might never have them, as now they are on-screen for good reason).

By registering with the widget, the on-view parameters inherit this shortcut filtering. This is probably desirable. However, it does not appear currently possible to have the typed unit taken with the on-screen parameters. Perhaps I do not know how it works. 

For DSHs not having a widget, then either we always construct a not-shown widget, or we lose the delay-mechanism.

Let's go one step at a time. **Can currently units be entered in the on-screen parameters?**

**Comment by abdullahtahiriyo** [comment]
Eventually, I am already expecting the ticket, people will want to be able to enter expressions for constraints. I am **NOT** saying we should do it now. I think it is well beyond this PR effort. Just noting it.

**Comment by PaddleStroke** [comment]
> However, it does not appear currently possible to have the typed unit taken with the on-screen parameters. Perhaps I do not know how it works.
> 
> For DSHs not having a widget, then either we always construct a not-shown widget, or we lose the delay-mechanism.
> 
> Let's go one step at a time. Can currently units be entered in the on-screen parameters?

I have to say that I forgot to check this. My goal at the time was to have the M key working, and passing toolWidget as filtering key events make the M key working. So I thought it also enabled typing the units. But I have actually not checked.

**Comment by abdullahtahiriyo** [comment]
Ok. Let's handle that one in another iteration after I am finished with the framework review.

However, to be able to test that one fully, we really need one DSH that has a parameter spinbox in the widget. In fact, best would be to have at least one spinbox in the widget and at least one on-view parameter. This would also enable to check the focus.

 Do you feel like adding one such DSH?

**Comment by PaddleStroke** [comment]
Ok I can work on the polygon DSH then. Should I start from your pw_tmp branch? You integrated all the changes in it?

**Comment by abdullahtahiriyo** [comment]
I have just forced-pushed the branch. Yes, please. Start on that branch. 

I know there is a bug somewhere in that branch in the Offset tool. It crashes. I do not know exactly when it started crashing. I planned to look into it when I and done with the framework. If you want to take a look that would also be helpful.

**Comment by abdullahtahiriyo** [comment]
Question, I am at:

```
/** Returns the state to which the widget parameter corresponds in the current construction method */
        auto getState(int parameterindex) const
        {
            Q_UNUSED(parameterindex);
            return handler->getFirstState();
        }
```
This function was originally written for widget parameters (the doc string still reflects that). Now this function is used for the on-view parameters. Essentially, it is a way for the tool to map the different parameters to different state machine states.

Now widget parameters cannot be associated to a state-machine state. Only OnViewParameters can be associated to a state-machine state.

In view all other DSHs that you are have or plan to code, do you see this as a limitation?

**Comment by PaddleStroke** [comment]
I don't think the widget spinbox will need to be binded to a specific mode.

**Comment by abdullahtahiriyo** [comment]
Ok.

**Comment by PaddleStroke** [comment]
Where have you force pushed ? On :  https://github.com/FreeCAD/FreeCAD/compare/main...abdullahtahiriyo:FreeCAD_sf_master:pw_tmp
it doesn't seems like you pushed recently.

**Comment by PaddleStroke** [comment]
On a side note, the function adaptParameters is now called after the mouse move. So it should not necessitate onSketchPos. As mouseMove usually registered the position to a meaningful variable.

**Comment by PaddleStroke** [comment]
Note: I have tried to type the unit in a onview spinbox and it works. If you type 100 then follows with letters then the letters are accepted.

**Comment by abdullahtahiriyo** [comment]
> it doesn't seems like you pushed recently.
Last commit is this:
8478c1cbf9dfb79f628ed9bac548e1a8e4cd6862

I forced pushed today but I had already done it yesterday, so no change. On top of that commit is ok.

**Comment by abdullahtahiriyo** [comment]
> On a side note, the function adaptParameters is now called after the mouse move. So it should not necessitate onSketchPos. As mouseMove usually registered the position to a meaningful variable.

Ok. Save this change for the next round.

**Comment by abdullahtahiriyo** [comment]
> Note: I have tried to type the unit in a onview spinbox and it works. If you type 100 then follows with letters then the letters are accepted.

Have you tried to type 10 m, or 10 cm? The input accepts letters, but it does not interpret them. Here it goes with 10cm ending up in 10 mm:

[Peek 2023-10-23 18-37.webm](https://github.com/FreeCAD/FreeCAD/assets/5249162/3d1792f7-bf50-4324-813d-0a0d3bd354ba)

**Comment by abdullahtahiriyo** [comment]
> As it appeared that you did not forced pushed your branch (last commit is from yesterday). I made the changes on my branch :
> [main...PaddleStroke:FreeCAD:pw_tmp](https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp)
> There are 2 new commits. Please check them.
> Thanks

This would be helpful:
1. Make a backup copy of your paddle  pw_tmp
2. clone my pw_tmp
3. cherry-pick your two polygon commits
4. Check that the polygon still works properly
5. If you can, take a look at why the offset tool crashes in this branch.

**Comment by PaddleStroke** [comment]
I have updated my https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp branch to your branch.

**Comment by PaddleStroke** [comment]
I will have a look at the offset tool, but as I have no idea what changes you made, and since you squashed it all, its going to be difficult to find the crash...

**Comment by abdullahtahiriyo** [comment]
It is a single commit about the construction geometry. It was being abused quite a lot in the handlers (lots of implicit conversions to int/booleans). It was not only in your handlers, but also in others. I plan to move this one out. But now I am stuck with other changes.

In that branch changes are mostly yours + the ones discussed above in this thread (or on the branch). But you can move to the ```sketcher/Gui``` directory and do a ```git diff branch1 branch2 .``` (the last dot is important as this restricts the changes to sketcher/Gui and thus any change having to do with rebasing to master).  Continuous rebasing to master is detrimental for running ```git diff``` (which I do quite a lot). 

I understand squashing makes tracking more difficult, but working with tens of development commits changing direction is not an easy task either.

I appreciate you taking the time to look into it.

**Comment by abdullahtahiriyo** [comment]
Oh god! You are right. There are quite some commits that do not belong even to us. I will force-push again when I have solved it.

**Comment by abdullahtahiriyo** [comment]
Ok. Now is forced pushed. Just drop your commits on top of b6ad90affc375682f3cd8f1e86b6772c1a7dc7c9. Sorry for that.

**Comment by abdullahtahiriyo** [comment]
I have forced pushed the branch to include the Polygon DSH and fixed it to some small changes in the framework. I have left the changes unsquashed so that you can see the small modifications.

I found the use of label/WLabel quite misleading. The W stands for Widget. The Labels in the widget could be the QLabels. That is the reason for the new name. I am going to rename functions having label within DSDWH for coherence.

In the Polygon DSH, in a quick test, I found a limitation that it has no angle snapping (despite having code for it). If you have anglesnapping in your branch, then it may be worth looking into why using ```angleSnappingControl``` instead of the ```onModeChangeHandler``` is not working. The former is called from DSDH (not DSDWH).

**Comment by abdullahtahiriyo** [comment]
Note: initNOnViewParameters hardcodes colors. Color of set parameter should follow that of constraints.

**Comment by PaddleStroke** [comment]
@abdullahtahiriyo I actually already rebased on your latest branch and implemented the polygon tool there. On my side the angle snapping is working properly.
I also made a few more commits please check https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp

I am now starting to look into offset crash.

**Comment by PaddleStroke** [comment]
You are right about the colors. Do you want me to do the change or have you already done it on your side?

**Comment by abdullahtahiriyo** [comment]
> You are right about the colors. Do you want me to do the change or have you already done it on your side?

Not at this time. It is more of a note not to forget in the next iteration.

**Comment by PaddleStroke** [comment]
I think I found the offset crash : 
You are calling angleSnappingControl after finish. And finish purge the handler in non-continuous mode.
```
    virtual void onModeChanged() override
    {
        finish();  // internally checks that state is SelectMode::End, and only finishes then.
        angleSnappingControl();
    };
```

**Comment by PaddleStroke** [comment]
offset crash is fixed. Please check my branch for the latest commits https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp

**Comment by abdullahtahiriyo** [comment]
> offset crash is fixed. Please check my branch for the latest commits [main...PaddleStroke:FreeCAD:pw_tmp](https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp)

Thank you. This was very helpful.

I have just forced-pushed my branch with all your today's additions. I have not squashed commits so that you can follow the changes. Most it in the the DSHs is about renaming. I am separating the code which handles the widget and the on-screen parameters (still with DSDHW). The idea is to be able to make DSDHW more modular, so that the framework can support custom widgets. The interface towards the specific DSHs will not change. This effort is just about tidying up the framework.

If you want to work on other tools, or modify the existing, please use this forced-pushed branch due to the renaming.

**Comment by abdullahtahiriyo** [comment]
> On my side the angle snapping is working properly.

Now it is work here to.

**Comment by PaddleStroke** [comment]
@abdullahtahiriyo I have just pushed 3 commits to solve the hard-coded color issue. It was before seeing your message so it is not on your rebased branch. Can you cherry pick them?

**Comment by PaddleStroke** [comment]
Btw regarding this hardcoded color fix, I wonder if I actually have to fetch the value from the parameters.
Or can I apply the DrawingParameter colors directly ? So like this instead : 

```
                SbColor sbcolor = DrawingParameters::DeactivatedConstrDimColor;

                // the returned is a naked pointer
                auto parameter = onViewParameters
                                     .emplace_back(std::make_unique<Gui::EditableDatumLabel>(
                                         viewer,
                                         placement,
                                         sbcolor,
                                         /*autoDistance = */ true))
                                     .get();
```

herostrat:
Sorry to disturb you guys, but I wanted to say something:
I really love the way you two work together, it is a pleasure to observe you working on new features.

Thanks for the hard work for freecad and thanks for the awesome teamwork you do.

abdullahtahiriyo:
> @abdullahtahiriyo I have just pushed 3 commits to solve the hard-coded color issue. It was before seeing your message so it is not on your rebased branch. Can you cherry pick them?
> 
> Edit: I decided to fix them now anyway because it looked bad in Open Dark new theme pack from Obelisk who's making a hard work to make a good color scheme.

It is ok. You did reasonably well and this can be transplanted easily. But look at the comment I left on the breaking of the architectural barriers and how to overcome that. You may well work in that branch on this topic and I will cherry-pick or directly copy the changes (it is code extending the framework, not modifying it). 

My comment about working on the new branch is more directed to whether you work on specific DSHs. In that case, the renaming will get in the middle.

abdullahtahiriyo:
@PaddleStroke 

I have just pushed my branch.

I am trying to cannibalise DrawSketchDefaultWidgetHandler into two separate classes, one for the controller and one for the handler. 

The reason for this, is that DrawSketchDefaultWidgetHandler is a do-all closed product that does not allow for custom widgets. Implementing a custom widget handler would amount to repeating 80% of the code there. This would lead to having to maintain a substantial amount of duplicated code per custom widget handler.

At this first stage, basically the handler part of DrawSketchDefaultWidgetHandler is made into DrawSketchControllableHandler and the controller part (ToolWidgetManager) into DrawSketchController. The changes to the actual specific handlers are only cosmetic (see last commit).

As a bonus, this architecture will work with a single handler (not two, one handler and one dHandler). It is now called dHandler, but I plan to rename it to handler. This is implementation type (DrawSketchHandlerPoint).

All the function interfaces remain the same and work exactly the same.

It works at my side, but I know that sometimes templates are handled in particular ways by MSVC, so I am letting the CI run this before continuing.

In a second stage, the defaultwidget specific parts will be moved to another controller deriving from DrawSketchController (that would be the controller class when working with a default widget). Then, an implementation requiring a custom widget will derive its own control from the stripped down DrawSketchController. This way, it will get all the common implementation for free, and will only have to add the specifics.

The nice thing is that all this reshufle is only internal so the DSH tools won't be affected (other than by cosmetic changes).

PaddleStroke:
Ok. If there are no bugs introduced that's fine for me. So is the framework ready to merge now? Aside from this hardcoded color issue.

abdullahtahiriyo:
I have just pushed my branch. All today's work squash. There is no DSDWH anymore. The DSH's are all migrated to the new controller-handler framework.

In my list:
- The headers of CommandSketcherTools we are discussing in the branch (move what is necessary to the offset tool and clean up there).
- Units in on-view parameters (they are not working as shown in the screenshot). Check the eventfilter mechanism.
- Harcoded colors issue
- Split DrawsketchController into the non-default-widget functionality and the default-widget functionality
- Lints (after pushing the PR, as I do not have them handy in Kdevelop)

So that I can do the split (which is rather straightforward), I need a solution for the units issue, as it will relate to the eventfilter issue. The eventfilter issue is problematic because the Editable labels take the default widget class. This precludes me from doing the splitting as the default widget belongs to B and the editable labels to A.

My lately very high availability is also reaching its end. This does not mean I am going, but it is going to diminish.

I would ask you to clone my branch and try to find a solution for the units issue. As we shortly discussed above, I am fine with moving the keyboard mechanism (of the "m") from SketcherDefaultWidget to DrawSketchController. That should have the highest priority as it is a blocking issue. 

You may also push a solution for the color issue. That should be straightforward with what we discussed.

Tomorrow evening I will look into the progress and will reply to any question.

The idea is fixing these issues and merging over the weekend (or earlier if it is done earlier).

PaddleStroke:
Headers and hardcoded colors are done and pushed.
I will look into the unit issue.

PaddleStroke:
I just checked and units are working correctly. On your video it doesn't work because you validate '10 cm mm' which is wrong. If you type '1 cm' and remove the original mm, then you get 10mm correctly.

PaddleStroke:
I have finished to move the KeyboadManager to the Controller. Please check latest commits : 
https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp

I am not entirely sure about how I managed the move of the Keyboard manager, so please feedback if its ok.

So now there is only the split you want to do and the lint messages.

PaddleStroke:
I have made a draft PR to see the LINT messages : https://github.com/FreeCAD/FreeCAD/pull/11174.
I pushed commits to solve most of Lint warnings. There are 2 kind of messages left that I detailed in 11174 that still need to be resolved.

So what is left to do : 
- Check the KeyboardManager commit and feedback what needs to be done.
- Check the 2 lint warning left.
- Split what you wanted to split.

abdullahtahiriyo:
> I just checked and units are working correctly. On your video it doesn't work because you validate '10 cm mm' which is wrong. If you type '1 cm', then you get 10mm correctly.

OK. One needs to type the cm and then delete the remaining mm. Somehow, I thought that the preexisting mm where separate (a separate control). If the user types no unit they stay, if the user types a unit they change. But, it is ok as it is if it reasonably works.

In previous tests often the execution of the DSH was often interrupted because the 'm' would be treated as shortcut "m". This added to the confusion as it appeared the tool was misbehaving.

I have not tested yet today's work (I have just tested this with yesterday's branch).

abdullahtahiriyo:
> Edit: It builds on windows but I made a draft PR and ubuntu builds are failing. I think this is because there is not the
> `#include "moc_DrawSketchController.cpp"` include that is needed since KeyboardManager is now a QObject (so that it can filter events). As DrawSketchController does not have a .cpp file I am not sure how to handle this.

Moving the keyboardManager to a separate header file, providing a cpp and adding the moc to the cpp. I have done it locally in my branch and it solves the issue.

abdullahtahiriyo:
@PaddleStroke 

In DSH Offset:
1. Checkboxes do not have the U/J shortcut indications.
2. If the addConstraint checkbox is selected, when I offset a rectangle I get redundant constraints.

abdullahtahiriyo:
> I have made a draft PR to see the LINT messages

After years of using Kdevelop as IDE, I am seriously considering moving to qtcreator. Apparently with qtcreator you can get the lints from cpplint directly on the editor.

PaddleStroke:
@abdullahtahiriyo I will check the offset issues

PaddleStroke:
Note I am also refactoring lot of things in offset DSH. As it was writen a long time ago there are some syntax that are not ideal and functions hard to read.

PaddleStroke:
@abdullahtahiriyo I have just pushed fixes to offset: https://github.com/FreeCAD/FreeCAD/compare/main...PaddleStroke:FreeCAD:pw_tmp
This fixes the redundant parallel constraint.
Adds UJ
Refactor things.

abdullahtahiriyo:
@PaddleStroke 

Thanks for looking into DSH Offset.

I did not have much time to test, but on the functional part, I am not fully satisfied with the mechanism for managing the keyboard when editing the units in the boxes. If I want to write 10 cm, it is often that I end up without on-view parameters because it takes a 'c' or 'm' as command. I am unsure if it is merely a matter of increasing the timeout. **Is you experience smooth?**

I did not have too much time today and I still have to check some functions on my side that are currently not working as they should... My branch has your work, including the DSH Offset, but it is not functional at the moment. If you want to look into anything, better stick for the time being with yours.

I am still looking into PR-ing this before the weekend. Let's see.

PaddleStroke:
I agree that the keyboard typing of units is not so smooth. Indeed maybe the timeout would benefit from being increased a little bit.
But I think even as it is, not being very smooth, it's still acceptable because changing the unit is not something that people will do often I think. It's really a edge case. Other softwares do not even show the unit in those spinboxes. So I think it is acceptable to keep as it is, with perhaps increasing the timeout by 50-100%.

abdullahtahiriyo:
> I agree that the keyboard typing of units is not so smooth. Indeed maybe the timeout would benefit from being increased a little bit.
> But I think even as it is, not being very smooth, it's still acceptable because changing the unit is not something that people will do often I think. It's really a edge case. Other softwares do not even show the unit in those spinboxes. So I think it is acceptable to keep as it is, with perhaps increasing the timeout by 50-100%.

Not saying it is a show stopper. Just saying that perhaps we could try to do a little bit better while the other issues get fixed. Would you do some tests with your branch to see where do you locate the sweet spot for the timeout and give it some extra thought?

BTW, I found my little issue with the splitting, and will fix it tomorrow.

PaddleStroke:
What about not showing the unit by default (for distances at least). So instead of `123 mm` you would have `123`. The advantage is that if you want to specify the unit (cm for example), then you just have to type de 'cm' and not remove the 'mm' that was there. It will also reduce clutter on the view.

abdullahtahiriyo:
> What about not showing the unit by default (for distances at least). So instead of `123 mm` you would have `123`. The advantage is that if you want to specify the unit (cm for example), then you just have to type 'cm' and not remove the 'mm' that was there. It will also reduce clutter on the view.

It is an option. I anticipate a ticket though. Some people work in their projects with meters (for example). No unit may be misleading (although great for the shortcut problem).

If I understood correctly, that control is used elsewhere in FreeCAD, so any change should not negatively affect there (though it could improve there).

Ok. Let's try to work some alternatives:
A. What about showing the unit by default, but removing it when starting editing (keyboard entry).

The problem with this, might be, that the user may think the unit is gone and forced to write it again. The user can enter another unit and then this one is taken. This option needs to find a sweet spot for the event filter.

B. What about showing the unit by default, but the widget removing any trailing second unit. So 10 cm mm => 10 cm. This option needs to find a sweet spot for the event filter.

C. What about splitting the control into two areas, one for the value and another one for the unit. Only the value has the focus (not the unit), but both are shown. If the user types a unit as part of the value being entered, then the typed unit is used

When we PR this, we can also ask for opinions to testers (if any).

PaddleStroke:
The unit could be hidden only when it is not the default unit. Similarly as dimension constraint has an option to hide only if its the default unit.

B - I think this one is a bit confusing. Beside as a user I wouldn't validate '10 cm mm' in the first place.

C - So basically we would have 2 widgets, a spinbox for the value, and a combobox or another spinbox for the unit.
This reminds me a draft PR I made to simplify the edit datum dialog : 
https://www.youtube.com/watch?v=W5sP93yWt8c&ab_channel=FreeCADDevDiary
So we could have a tiny spinbox without unit. Then an arrow, if you open it the widget is bigger and offer a unit combobox for example.
image

But I think this is maybe overkill. I would rather go with hidding the unit when its a default distance (so hide mm but not m or µm) and increase the timeout by 75%

But yes you are right. It's probably best to merge as is and ask users to give feedback.

abdullahtahiriyo:
> This reminds me a draft PR I made to simplify the edit datum dialog:

I do like this kind of solution. It also paves the way for expressions to be inserted. However, it still has the issue that one cannot move the mouse to click the arrow...

... an option could be to use a hotkey to open the extended dialog and while this dialog is open top tracking the mouse in the DSH until it is closed...

... or once the on-view parameter starts receiving numbers, it could stop tracking the mouse (to allow clicking on the arrow) until a hot key (TAB, ENTER, ESC) is hit, when it starts tracking the mouse again...

It is ok to go for it as a separate PR (if you wish to).

abdullahtahiriyo:
@PaddleStroke 

I have just pushed my branch.

Please clone it, do some testing and if everything works ok, PR-it.

maxwxyz:
@PaddleStroke Quick question, can I directly name the value (like a variable) or enter expressions in the overlay box?

Is this overlay directly invoked when drawing the geometry (to draw it at the right length and add dimension) or do I have to click on the dimension tool?

PaddleStroke:
Not yet no. But we are conscious that this will become a feature request pretty soon.

luzpaz:
@maxwxyz JFYI, checkout @mwganson's Dynamic Data addon (https://github.com/mwganson/DynamicData)

PR #10909, opened 2023-10-02
App/Toponaming: Add base64 encoding to XML

chennes:
This PR adds @realthunder's Base64 encoding/decoding support to Reader and Writer. The code in the first commit is nearly identical to that in the Toponaming branch, with the exception of the addition of two enum classes to replace boolean parameters. 
* `Base64ErrorHandling` (replaces a boolean indicating whether or not to throw an error on bad character data)
*  `CharStreamFormat` (replaces a boolean indicating whether to Base64-encode/-decode the data). `CharStreamFormat` is defined in the [src/Base/FileInfo.h](https://github.com/FreeCAD/FreeCAD/compare/master...chennes:toponamingPhase3Card1?expand=1#diff-6fba68d97e78796c20dbf9d25340188f74cf7014ec48c3e705090a704679af2a) header because it's needed by both `Reader` and `Writer` -- @wwmayer does that seem like an OK place for it?

Two simple tests are added to verify the encode and decode functionality.  The second commit applies some linter fixes -- it is not exhaustive. In particular it does not address the complaints about pointer arithmetic or array indexing.

### DIFF
--- a/src/Base/FileInfo.h
+++ b/src/Base/FileInfo.h
@@ -34,6 +34,15 @@
 namespace Base
 {
 
+/// When reading and writing a character stream, the incoming data can be dumped into the stream
+/// unaltered (if it contains only data that is valid in the current XML character set), or it can
+/// be Base64-encoded. This enum is used by Reader and Writer to distinguish the two cases.
+enum class CharStreamFormat {
+    Raw,
+    Base64Encoded
+};
+static constexpr int base64LineWidth {80};
+
 /** File name unification
   * This class handles everything related to file names
   * the file names are internal generally UTF-8 encoded on
--- a/src/Base/FileInfo.h
+++ b/src/Base/FileInfo.h
@@ -41,7 +41,6 @@ enum class CharStreamFormat {
     Raw,
     Base64Encoded
 };
-static constexpr int base64LineWidth {80};
 
 /** File name unification
   * This class handles everything related to file names
--- a/src/Base/Base64.cpp
+++ b/src/Base/Base64.cpp
@@ -1,128 +1,162 @@
-/* 
-   base64.cpp and base64.h
+/*
+base64.cpp and base64.h
 
-   Copyright (C) 2004-2008 René Nyffenegger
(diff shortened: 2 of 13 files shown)


wwmayer:
The amount of code of Base64 has become enormous and heavy but when looking at the repo of the original author it is still rather small: https://github.com/ReneNyffenegger/cpp-base64
In his repo he has some Google unit tests and it would be good to copy them too in order to make sure the algorithm still works correctly.

src/Base/FileInfo.h
```diff
@@ -34,6 +34,14 @@
 namespace Base
 {
 
+/// When reading and writing a character stream, the incoming data can be dumped into the stream
```
wwmayer, line 37: Putting this enum to FileInfo.h seems wrong to me as it's used in the Reader class.

chennes:
> The amount of code of Base64 has become enormous and heavy but when looking at the repo of the original author it is still rather small

Do you want me to move the new stream filter stuff to its own files? We could even consider re-synchronizing our copy with a new one from Nyffenegger, it looks as though he's continued working on it since we grabbed our copy.

wwmayer:
> Do you want me to move the new stream filter stuff to its own files?

If it's possible to cleanly separate the code then it is a good idea to move RT's code into separate files.

chennes:
> If it's possible to cleanly separate the code

I can easily move the actual filter structs into a new file, but the modifications to use a provided character buffer were changes to the original function, so they aren't really extractable. I could split the original function into two parts so that the modification is more obvious, but I'm not sure that provides a significant benefit.

chennes:
> In his repo he has some Google unit tests and it would be good to copy them too in order to make sure the algorithm still works correctly.

I added his tests (slightly refactored) in this latest commit. I removed all the output to stdout from them and split his single test up into the four individual tests that it was doing.

wwmayer:
All tests pass and I didn't find anything obviously wrong. So, it's good to merge.

chennes:
I missed marking a constructor as `explicit` -- I'll wait for the CI to complete then merge it. Thanks!

PR #10914, opened 2023-10-02
Path:  Added type annotations along with some refactoring.

LarryWoestman:
This is part of the work for #6117.

This is one in a series of PR's where I am taking code from at least 4 postprocessor files partially written by different people, combining them into three files, then trying to make the code more consistent and simpler.  The long term goal is to enable much simpler postprocessor files with much less copied code in each postprocessor file and much more "shared" code in the the three Utils*.py files.

In this PR I added type annotations along with some required refactoring.

### DIFF
--- a/src/Mod/Path/Path/Post/UtilsArguments.py
+++ b/src/Mod/Path/Path/Post/UtilsArguments.py
@@ -33,21 +33,30 @@ These are functions related to arguments and values for creating custom post pro
 import argparse
 import os
 import shlex
+from typing import Any, Callable, Dict
 
 from FreeCAD import Units
 
 import Path.Post.UtilsParse as PostUtilsParse
 
+# Define some types that are used throughout this file
+PathParameter = float
+PathParameters = Dict[str, PathParameter]
+Parser = argparse.ArgumentParser
+Values = Dict[str, Any]
+
+ParameterFunction = Callable[[Values, str, str, PathParameter, PathParameters], str]
+
 
 def add_flag_type_arguments(
     argument_group,
-    default_flag,
-    true_argument,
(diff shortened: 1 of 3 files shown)


src/Mod/Path/Path/Post/UtilsArguments.py
```diff
@@ -33,21 +33,30 @@
...
-    false_help,
-    visible=True,
-):
+    default_flag: bool,
+    true_argument: str,
+    false_argument: str,
+    true_help: str,
+    false_help: str,
+    visible: bool = True,
+) -> None:
```
luzpaz, line 59: Line 51: Too many arguments (7/5) (too-many-arguments)

LarryWoestman: At this stage in the refactoring I have too many arguments in several places.  This will require refactoring the data structures used in the existing code.

I am trying to follow the guidelines related to keeping the PR's smaller and doing more of them, so I am waiting for a later PR to do the data structure refactoring since that will cause a lot of lines of code to change.  Most likely I will rewrite many of the existing routines to use an interface class and change many of the existing functions into classes to clean up the overall structure.

This is part of the process of taking already existing code written by at least 4 different people and trying to make it more consistent and supportable.  That requires a lot of change overall.

Note that this code is work-in-progress and is only used by the "refactored*" postprocessors.  It shouldn't be depended on by any other code in FreeCAD.  The "refactored*" postprocessors should work but they aren't ready to depend on, yet, because they will still be changing a lot for quite a while into the future.

luzpaz:
@sliptonic can you review (or designate someone to review) this PR soon ?

PR #9097, opened 2023-03-28
[FEM] pressure constraint overhaul

donovaly:
- add missing binding to pressure value in dialog
- make pressure a property with unit to get rid of hacks

PR #10923, opened 2023-10-03
FEM: pressure constraint overhaul

marioalexis84:
see #9097


marioalexis84:
@sliptonic

kadet1090:
If you only rebased changes you should mark @donovaly as actual author via git, you can do that via:
```
git commit --author "Uwe <donovaly@users.noreply.github.com>"
```

Here you have more informaiton how to do it in rebase to also have your name as `Commiter` https://stackoverflow.com/questions/16217801/git-rebase-change-author - I don't know if this is required in this case, but it is useful git skill to have nonetheless :)

PR #10929, opened 2023-10-04
Draft: importSVG: handle viewBox in mm

Roy-043:
Forum topic:
https://forum.freecad.org/viewtopic.php?t=81696

### DIFF
--- a/src/Mod/Draft/importSVG.py
+++ b/src/Mod/Draft/importSVG.py
@@ -820,18 +820,16 @@ class svgHandler(xml.sax.ContentHandler):
                 if 'width' in data \
                         and 'height' in data \
                         and 'viewBox' in data:
-                    vbw = float(data['viewBox'][2])
-                    vbh = float(data['viewBox'][3])
-                    w = attrs.getValue('width')
-                    h = attrs.getValue('height')
-                    self.viewbox = (vbw, vbh)
                     if len(self.grouptransform) == 0:
                         unitmode = 'mm' + str(self.svgdpi)
                     else:
                         # nested svg element
                         unitmode = 'css' + str(self.svgdpi)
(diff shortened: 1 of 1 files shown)


yorikvanhavre:
LGTM!

nmschulte:
With this change, an SVG (created in Inkscape) that imported as I would expect in versions prior, now imports as ~3.77x the size.

image

image


```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->

<svg
   width="540mm"
   height="120mm"
   viewBox="0 0 540 120"
   version="1.1"
   id="svg5"
   inkscape:version="1.2.2 (b0a8486541, 2022-12-01)"
   sodipodi:docname="test.svg"
   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
   xmlns:xlink="http://www.w3.org/1999/xlink"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:svg="http://www.w3.org/2000/svg">

<!-- ... -->

</svg>
```

Roy-043:
@nmschulte without an actual file there is not much we can do with your report.

nmschulte:
@Roy-043, no worries; here's an example: a 100 mm x 100 mm canvas w/ a 50 mm x 50 mm un-stroked red filled square.
drawing
(zipped in case GitHub does something peculiar with the SVG: [drawing.zip](https://github.com/FreeCAD/FreeCAD/files/13061282/drawing.zip))

When I import into `main` with this commit:
image

When I import into `main` with this commit reverted:
image

PR #10931, opened 2023-10-04
Draft: Update the Plane class (step 2)

Roy-043:
Related issue:
#5603.

Previous PRs:
#10766
#10838

The remaining Plane class functions have been processed (some are unchanged):

`setup`
* Updated to use new functions.
* The `direction` etc. arguments are never used AFAICT.
* Added `from draftutils import gui_utils`.

`reset`
* Unchanged.

`setTop`
* Calls `super().set_to_top()`. That function does not use `align_to_point_and_axis` but specifies exact vectors for improved accuracy.
* `import FreeCADGui` already happens at the top of the file.
* The code block after `if FreeCAD.GuiUp:` was kept for now.

`setFront`
* Idem.

`setSide`
* This function sets the WP to the left side position, and not to the right side postion as in the Draft_SelectPlane command.
* Idem.

`getRotation`
* Revised.
* The function has an akward name as it returns a Placement.
* The special code for Arch active container is obsolete (was already commented out in other functions).

`getPlacement`
* The `rotated argument` is ignored now. It was only used in the context of the old `placement_from_face` algorithm. See `DraftGeomUtils.placement_from_face` for the new implementation.
* Calls `super().get_placement()`.

`getNormal`
* Basically unchanged.

`setFromPlacement`
* Revised to call `super()` functions.

`inverse`
* Unchanged.

`save`
* Unchanged.

`restore`
* More compact code, but basically unchanged.

`getLocalCoords`
* Calls `super().get_local_coords()` which does not rely on Draft code, but creates a matrix and returns `mtx.inverse().multVec(point)`.

`getGlobalCoords`
* Calls `super().get_global_coords()` which creates a matrix and returns `mtx.multVec(point)`.

`getLocalRot`
* Calls `super().get_local_coords()`.

`getGlobalRot`
* Calls `super().get_global_coords()`.

`getClosestAxis`
* Calls `super().get_closest_axis(vec)` which translates the vector to the WP coordinate system and then finds the closest axis by comparing the absolute coordinates of the vector.

`isGlobal`
* Not used anywhere AFAICT.
* Calls `super().is_global()` which also checks the `position` attribute.
* As before the function works without any tolerance.

`isOrtho`
* Not used anywhere AFAICT.
* The original algorithm did not work properly (there was a comment). All angles were measured relative to the global Y axis. If the WP was randomly rotated around that axis the function would still return `True`. The new algorithm, `super().is_ortho()`, uses rounded Yaw, Pitch and Roll angles to check for orthogonality.

`getDeviation`
* Not used anywhere AFAICT.
* Unchanged.

`getParameters`
* Calls `super().get_parameters()` which uses the local `_get_prop_list` function.

`setFromParameters`
* Calls `super().set_parameters()` which uses the local `_get_prop_list` function.

`_get_prop_list`
* Function used internally. Returns a list of attribute.

### DIFF
--- a/src/Mod/Draft/WorkingPlane.py
+++ b/src/Mod/Draft/WorkingPlane.py
@@ -40,6 +40,7 @@ import lazy_loader.lazy_loader as lz
 import FreeCAD
 import DraftVecUtils
 from FreeCAD import Vector
+from draftutils import gui_utils
 from draftutils import utils
 from draftutils.messages import _wrn
 from draftutils.translate import translate
@@ -881,7 +882,7 @@ class Plane(PlaneBase):
         """Set up the working plane if it exists but is undefined.
 
         If `direction` and `point` are present,
-        it calls `alignToPointAndAxis(point, direction, 0, upvec)`.
+        it calls `align_to_point_and_axis(point, direction, 0, upvec)`.
 
         Otherwise, it gets the camera orientation to define
         a working plane that is perpendicular to the current view,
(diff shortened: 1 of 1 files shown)


yorikvanhavre (approved):
didn't try it live yet but so far so good! Let's merge and see :)

PR #10932, opened 2023-10-04
Clang-tidy: enable check google-explicit-constructor

wwmayer:
Currently this check isn't enabled but when pushing a PR then the lint checker warns about non-explicit constructors which is a bit annoying because the PR must be updated to fix the warning or in most cases it won't be fixed at all.

By enabling this check the developer already sees the warning before pushing a PR.


chennes:
Good idea, thanks.

PR #10933, opened 2023-10-04
Gui: Python support of workbench manipulation

wwmayer:
See https://github.com/FreeCAD/FreeCAD/pull/10732#issuecomment-1746496249

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -124,6 +124,7 @@
 #include "WaitCursor.h"
 #include "Workbench.h"
 #include "WorkbenchManager.h"
+#include "WorkbenchManipulator.h"
 #include "WidgetFactory.h"
 
 
@@ -535,6 +536,7 @@ Application::~Application()
 {
     Base::Console().Log("Destruct Gui::Application\n");
     WorkbenchManager::destruct();
+    WorkbenchManipulator::removeAll();
     SelectionSingleton::destruct();
     Translator::destruct();
     WidgetFactorySupplier::destruct();
--- a/src/Gui/Application.h
+++ b/src/Gui/Application.h
@@ -348,6 +348,9 @@ public:
     static PyObject* sAddDocObserver           (PyObject *self,PyObject *args);
     static PyObject* sRemoveDocObserver        (PyObject *self,PyObject *args);
 
+    static PyObject* sAddWbManipulator         (PyObject *self,PyObject *args);
+    static PyObject* sRemoveWbManipulator      (PyObject *self,PyObject *args);
+
     static PyObject* sListUserEditModes        (PyObject *self,PyObject *args);
     static PyObject* sGetUserEditMode          (PyObject *self,PyObject *args);
     static PyObject* sSetUserEditMode          (PyObject *self,PyObject *args);
--- a/src/Gui/ApplicationPy.cpp
+++ b/src/Gui/ApplicationPy.cpp
@@ -66,6 +66,7 @@
 #include "WidgetFactory.h"
 #include "Workbench.h"
 #include "WorkbenchManager.h"
+#include "WorkbenchManipulatorPython.h"
 #include "Inventor/MarkerBitmaps.h"
 #include "Language/Translator.h"
 
@@ -355,6 +356,18 @@ PyMethodDef Application::Methods[] = {
    "removeDocumentObserver(obj) -> None\n"
    "\n"
    "Remove an added document observer.\n"
+   "\n"
+   "obj : object"},
+  {"addWorkbenchManipulator",  (PyCFunction) Application::sAddWbManipulator, METH_VARARGS,
+   "addWorkbenchManipulator(obj) -> None\n"
+   "\n"
+   "Add a workbench manipulator to modify a workbench when it is activated.\n"
+   "\n"
+   "obj : object"},
+  {"removeWorkbenchManipulator",  (PyCFunction) Application::sRemoveWbManipulator, METH_VARARGS,
+   "removeWorkbenchManipulator(obj) -> None\n"
+   "\n"
+   "Remove an added workbench manipulator.\n"
    "\n"
    "obj : object"},
   {"listUserEditModes", (PyCFunction) Application::sListUserEditModes, METH_VARARGS,
@@ -1528,6 +1541,32 @@ PyObject* Application::sRemoveDocObserver(PyObject * /*self*/, PyObject *args)
 
     PY_TRY {
         DocumentObserverPython::removeObserver(Py::Object(o));
+        Py_Return;
+    }
+    PY_CATCH;
+}
+
+PyObject* Application::sAddWbManipulator(PyObject * /*self*/, PyObject *args)
+{
+    PyObject* o;
+    if (!PyArg_ParseTuple(args, "O",&o))
+        return nullptr;
+
+    PY_TRY {
+        WorkbenchManipulatorPython::installManipulator(Py::Object(o));
+        Py_Return;
+    }
+    PY_CATCH;
+}
+
+PyObject* Application::sRemoveWbManipulator(PyObject * /*self*/, PyObject *args)
+{
+    PyObject* o;
+    if (!PyArg_ParseTuple(args, "O",&o))
+        return nullptr;
+
+    PY_TRY {
+        WorkbenchManipulatorPython::removeManipulator(Py::Object(o));
         Py_Return;
     }
     PY_CATCH;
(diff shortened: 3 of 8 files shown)


wwmayer:
> So it would be something like: Gui.manipulateMenu("insert","Std_Help", "Bim_Help")

No, that's not how it works. You implement a Python class
```
class WbManipulator:
   def modifyMenuBar(self):
       return [("insert","Std_Help", "Bim_Help")]
```

Then at some point you register it to the system with `Gui.installManipulator(WbManipulator())`

The method modifyMenuBar returns a list of tuples or list or dicts to allow to support several changes but we only changing one thing it could be allowed to return the inner tuple, list or dict directly.

But at this point it's only a proposal and things could still be changed when starting to implement it.

wwmayer:
I decided to return a dict or list of dicts to tell the calling instance what to change.
Here some code snippets how to manipulate things:

This manipulator
* removes the Quit command from the File menu
* adds the About command to the end of the Macro menu
* add the About command before the Parameter editor
* and adds the Windows command after the Parameter editor

By default the "insert" action puts a command before the reference command. If it should be added afterwards then the dict must contain the key "after". Its associated value is irrelevant.
```
class Manipulator:
  def modifyMenuBar(self):
    return [{"remove" : "Std_Quit"},
                  {"append" : "Std_About", "menuItem" : "Std_DlgMacroRecord"},
                  {"insert" : "Std_About", "menuItem" : "Std_DlgParameter"}
                  {"insert" : "Std_Windows", "menuItem" : "Std_DlgParameter", "after" : ""}]

manip = Manipulator()
Gui.addWorkbenchManipulator(manip)
```

This manipulator remove the whole sub-menu "Standard views" from the context-menu and adds the Windows command before the Measure toggle command.
```
class Manipulator:
   def modifyContextMenu(self, recipient):
     if recipient == "View":
       return [{"remove" : "Standard views"},
                     {"insert" : "Std_Windows", "menuItem" : "View_Measure_Toggle_All"}]

 manip = Manipulator()
 Gui.addWorkbenchManipulator(manip)
 ```

This manipulator removes the Macro toolbar and adds the Quit command at the end of the File toolbar
 ```
 class Manipulator:
   def modifyToolBars(self):
       return [{"remove" : "Macro"},
               {"append" : "Std_Quit", "toolBar" : "File"}]

 manip = Manipulator()
 Gui.addWorkbenchManipulator(manip)
 ```

**Comment by yorikvanhavre** [comment]
Excellent!!

**Comment by mwganson** [comment]
In experimenting with this I tried:

```
class Manipulator:
  def modifyToolBars(self):
      return [{"remove" : "PartDesign_ShapeBinder", "toolbar": "PartDesignHelper"},
              {"insert" : "Std_Part", "menuItem":"PartDesign_Point", "toolbar" : "PartDesignHelper"}]

manip = Manipulator()
Gui.addWorkbenchManipulator(manip)
wb = FreeCADGui.activeWorkbench()
wb.reloadActive()
```

It removes the ShapeBinder toolbar item, but doesn't add the Std_Part.  What am I doing wrong?  Is it not possible to both remove and insert at the same time?

**Comment by wwmayer** [comment]
> Is it not possible to both remove and insert at the same time?

Of course, it's possible.

> but doesn't add the Std_Part.

but I don't know what you try to do with the second action.

Did you want to do this?
```
class Manipulator:
  def modifyToolBars(self):
      return [{"remove" : "PartDesign_ShapeBinder", "toolbar": "PartDesignHelper"},
              {"insert" : "Std_Part", "toolItem":"PartDesign_Point"}]

manip = Manipulator()
Gui.addWorkbenchManipulator(manip)
wb = FreeCADGui.activeWorkbench()
wb.reloadActive()
```

**Comment by mwganson** [comment]
Thanks.  I see my mistake, using menuItem instead of toolItem.

Potentially, there is a bug with inserting.  I expected the Std_Part to be put in front of the datum point, but instead has been placed at the end.
Snip macro screenshot-157191

Initially I had tried to replace the shapebinder tool with a macro that also creates a shapebinder, but bypasses the dialog and handles the visibility of the base object differently.  When that failed I tried the Std_Part to see if that would work.

**Comment by wwmayer** [comment]
> Potentially, there is a bug with inserting. I expected the Std_Part to be put in front of the datum point, but instead has been placed at the end.

Yes, but it has nothing to do with the manipulator. It's caused by this: https://github.com/FreeCAD/FreeCAD/blob/main/src/Gui/ToolBarManager.cpp#L346

> Initially I had tried to replace the shapebinder tool with a macro that also creates a shapebinder,

You can also replace the command.

**Comment by alexneufeld** [comment]
> > Is it not possible to both remove and insert at the same time?
> 
> Of course, it's possible.
> 
> > but doesn't add the Std_Part.
> 
> but I don't know what you try to do with the second action.
> 
> Did you want to do this?
> 
> ```
> class Manipulator:
>   def modifyToolBars(self):
>       return [{"remove" : "PartDesign_ShapeBinder", "toolbar": "PartDesignHelper"},
>               {"insert" : "Std_Part", "toolItem":"PartDesign_Point"}]
> 
> manip = Manipulator()
> Gui.addWorkbenchManipulator(manip)
> wb = FreeCADGui.activeWorkbench()
> wb.reloadActive()
> ```

If I copy this code to a file called $HOME/.local/share/FreeCAD/Mod/SomeCustomModule/InitGui.py I get the following error:

```
During initialization the error "No active workbench
" occurred in $HOME/.local/share/FreeCAD/Mod/SomeCustomModule/InitGui.py
Please look into the log file for further information
```

The log file shows the following:
```
Log: ----------------------------------------------------------------------------------------------------
Log: Traceback (most recent call last):
  File "<string>", line 142, in RunInitGuiPy
  File "$HOME/.local/share/FreeCAD/Mod/SomeCustomModule/InitGui.py", line 35, in <module>
    wb = FreeCADGui.activeWorkbench()
AssertionError: No active workbench

Log: ----------------------------------------------------------------------------------------------------
Err: During initialization the error "No active workbench
" occurred in $HOME/.local/share/FreeCAD/Mod/SomeCustomModule/InitGui.py
Err: Please look into the log file for further information
```

I'm using this version of FreeCAD:
```
OS: EndeavourOS (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35370 (Git) AppImage
Build type: Release
Branch: main
Hash: 1066bb54ceec8f4f8376b5149253acda5677651d
Locale: English/Canada (en_CA)
Installed mods: 
  * FreeCAD-Flat-Icons
  * OpenDark 2023.12.1
  * TDAlignItems
```
@wwmayer any suggestions?

**Comment by alexneufeld** [comment]
To answer my own question:

Adapt this snippet from the [Macro at Startup](https://wiki.freecad.org/Macro_at_Startup) wiki page:

``` python
def runStartupMacros(name):
    # Do not run when NoneWorkbench is activated because UI isn't yet completely there
    if name != "NoneWorkbench":
        # Run macro only once by disconnecting the signal at first call
        FreeCADGui.getMainWindow().workbenchActivated.disconnect(runStartupMacros)

        # Following 2 lines shall be duplicated for each macro to run
        import MySuperMacro # <-\
        MySuperMacro.run()  # <-- replace these with the workbench manipulation commands

        # Eg. if a second macro shall be launched at startup
        # import MyWonderfulMacro
        # MyWonderfulMacro.run()

# The following 2 lines are important because InitGui.py files are passed to the exec() function...
# ...and the runMacro wouldn't be visible outside. So explicitly add it to __main__
import __main__
__main__.runStartupMacros = runStartupMacros

# Connect the function that runs the macro to the appropriate signal
FreeCADGui.getMainWindow().workbenchActivated.connect(runStartupMacros)
```

mnesarco:
Hello @wwmayer,  I have some questions about this:

```
 class Manipulator:
   def modifyMenuBar(self):
     return [{"remove" : "Std_Quit"},
             {"append" : "Std_About", "menuItem" : "Std_DlgMacroRecord"},
             {"insert" : "Std_About", "menuItem" : "Std_DlgParameter"}
             {"insert" : "Std_Windows", "menuItem" : "Std_DlgParameter", "after" : ""}]

```

1. What happens if Std_Quit appears in more than one menu?
2. What happens if Std_DlgMacroRecord appears in more than one menu?
3. What happens if Std_DlgParameter appears in more than one menu?



```
   def modifyToolBars(self):
       return [
             {"insert" : "Std_Cut", "toolItem" : "Std_New"}]
```

4. What happens if Std_New appears in more than one Toolbar?


The mentioned cases look a bit ambiguous, if exiting references exists in more than one place, are all removed, or just the first found are removed? what if i want to remove the one in a specific menu?

Also for insertion, an existing item is used as reference, but what if the existing command is in more than one menu/toolbar? is the new one inserted in all places where the reference is found?

Thank you.

PR #10940, opened 2023-10-05
[DO NOT MERGE] Sketcher: try to get a cleaner code

FlachyJoe:
@PaddleStroke this is a try to put the geometry selection logic out of `CmdSketcherConstrainDistance::activated`

I think your Dimension tool doesn't use this function so it's not really useful but I wrote it as an idea for the places with many nested if...elseif ( EditModeConstraintCoinManager.cpp and ViewProviderSketch.cpp ).

I would be happy to read your opinion.

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -712,18 +712,18 @@ int SketchSelection::setUp()
 }// namespace SketcherGui
 
 namespace{
-enum CONSTRAINT_DISTANCE_QUALIFIER{
-    POINT_TO_POINT_DISTANCE,
-    POINT_TO_LINE_DISTANCE,
-    POINT_TO_CIRCLE_DISTANCE,
-    CIRCLE_TO_CIRCLE_DISTANCE,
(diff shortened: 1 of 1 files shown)


src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -710,6 +711,58 @@ int SketchSelection::setUp()
 
 }// namespace SketcherGui
 
+namespace{
+enum CONSTRAINT_DISTANCE_QUALIFIER{
```
chennes, line 715: The C++ Core Guidelines recommend using `enum class` over `enum` in new code: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class

The also discourage all-caps names for things that aren't macros:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class

FlachyJoe: Thank you @chennes for the advices.

maxwxyz:
@FlachyJoe just a heads up for the upcoming feature freeze: 

PR #10951, opened 2023-10-06
 Start: fixes #10221: /tmp/FreeCADStartThumbnails* directories are no…

### LINKED ISSUE #10221: StartWB: `/tmp/FreeCADStartThumbnails*` directories are not removed when FreeCAD terminates

wwmayer:
…t removed when FreeCAD terminates

The commit adds some new functions:

* getThumbnailDir() Returns the path to the thumbnail directory which is a sub-directory of the Cache directory. There is no need to always create a unique directory after each restart because it doesn't harm if the thumbnail directoy contains deprecated files.

* createThumbnailDir() Creates the thumbnail directoy if it doesn't exist yet.

* getSha1Hash Helper function to compute a SHA-1 hash of a given path. If the same path is passed then the hash value will be the same. This way it can be avoided to create a different image file from a project file after each restart.

* getUniquePNG Computes the path of a PNG image file for a given project file. It's also possible to pass an arbitrary string as argument.

* useCachedPNG If the PNG image exists and if it's newer than the project file True is returned and False otherwise.

The changes in detail are:
1. For a given project file it is checked if the thumbnail directory already contains a cached image. If it's newer than the project file it will used, otherwise it will be re-created.

2. For the system icon to a given file format it is only checked if the icon already exists. If yes, it will be used.

### DIFF
--- a/src/Mod/Start/StartPage/StartPage.py
+++ b/src/Mod/Start/StartPage/StartPage.py
@@ -33,6 +33,7 @@ import re
 import FreeCAD
 import FreeCADGui
 import codecs
+import hashlib
 import urllib.parse
 from . import TranslationTexts
 from PySide import QtCore, QtGui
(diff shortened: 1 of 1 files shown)


yorikvanhavre (approved):
Amazing!!!! Thanks!

Issue #10924, opened 2023-10-03
[Problem] Material: Failing unit test


wwmayer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

When running the self-test TestMaterialsApp then the test testCalculiXSteel() (line 166) fails with the error message:
`7900,00000 kg/m^3 != 7900.00 kg/m^3`

There are two problems:
The comparison depends on the number of decimals and if not set to 2 it fails.
The hard-coded string is in C locale but the quantity might be in a different locale so that instead of dot a comma is used as decimal separator.

So, a string comparison is a bad idea to assert the result of the test.

### Full version info

```shell
[code]
OS: Ubuntu 22.04.3 LTS (XFCE/xubuntu)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34525 (Git)
Build type: debug
Branch: master
Hash: b42b2c85b4c4a230ba522620ab2d3df84a5bc313
Python 3.10.12, Qt 5.15.3, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.1
Locale: German/Germany (de_DE)
[/code]
```

### Subproject(s) affected?

Other (specify in description)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

wwmayer:
@davesrocketshop

wwmayer:
Duplicate of https://github.com/FreeCAD/FreeCAD/issues/10917

Issue #10917, opened 2023-10-03
[Problem] AssertionError during build attempt


InRiPa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Attempting to build FreeCAD under Arch Linux, using the [AUR package](https://aur.archlinux.org/packages/freecad-git#comment-936643)

results in an error:
```bash
======================================================================
FAIL: testCalculiXSteel (TestMaterialsApp.MaterialTestCases.testCalculiXSteel)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/user/.cache/yay/freecad-git/src/FreeCAD/build/Mod/Material/TestMaterialsApp.py", line 166, in testCalculiXSteel
    self.assertEqual(properties["Density"], "7900.00 kg/m^3")
AssertionError: '7900,00 kg/m^3' != '7900.00 kg/m^3'
- 7900,00 kg/m^3
?     ^
+ 7900.00 kg/m^3
?     ^


----------------------------------------------------------------------
Ran 1184 tests in 55.808s

FAILED (failures=1, skipped=2)
==> ERROR: A failure occurred in check().
    Aborting...
 -> error making: freecad-git-exit status 4
 -> Failed to install the following packages. Manual intervention is required:
freecad-git - exit status 4
```

Maybe someone can't point me in a direction on how to solve this.

### Full version info

```shell
No Version information since I can't get it to build.
```

### Subproject(s) affected?

None

### Anything else?

Since it seems to be an issue with number representation, please let me know if you need any system details.

```bash
# cat /etc/locale.conf
LANG=en_US.UTF-8
```
My system runs in `English` / `en_US`, but for Formats I'm using `German` / `de_DE`

I'm using Arch Linux with Linux 6.5.5 and Gnome 44.5

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

xtemp09:
@InRiPa, could you provide `locale` output? If you set everything to `en_US.UTF-8`, does it compile?

I think you need to set only `LC_NUMERIC` to `en_US.UTF-8`.

Syres916:
I would suggest the Material test should be fixed as per https://github.com/FreeCAD/FreeCAD/compare/master...Syres916:FreeCAD:Materials_Test_Locale rather than the user having to change locale in order to compile. Not having a German test machine, I've done my best using reverse logic for testing.

InRiPa:
@xtemp09 you're right.

[...]

PR #10956, opened 2023-10-06
Material: Update Python tests to work with other locales Issues #10917 and #10924

davesrocketshop:
Update Python tests to work with other locales

This fixes issues #10917 and #10924

Numeric values were beign compared to strings that did not take into account localizations. The test was modified to work regardless of localization.

There is still a known issue when working with pt_BR.UTF8. The default encoder for this localization is not UTF8 resulting in an error working with a parameter that returned the UTF character for the Greek letter mu in its units.

### DIFF
--- a/src/Mod/Material/TestMaterialsApp.py
+++ b/src/Mod/Material/TestMaterialsApp.py
@@ -23,7 +23,12 @@
 # import FreeCAD
 from os import walk
 import unittest
+import FreeCAD
 import Material
+
+parseQuantity = FreeCAD.Units.parseQuantity
+# import locale
+# locale.setpreferredencoding("UTF8")
 
 class MaterialTestCases(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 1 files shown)


luzpaz:
> This fixes issues https://github.com/FreeCAD/FreeCAD/issues/10917 and https://github.com/FreeCAD/FreeCAD/issues/10924

Not sure this will auto close those ticket. See https://github.com/gitbucket/gitbucket/wiki/How-to-Close-Reference-issues-and-pull-request

> Something like `Fixes #10917, fixes #10924`

But don't worry about it now. Just from here on out :+1:

wwmayer:
> There is still a known issue when working with pt_BR.UTF8.

Interesting! How do you know this? Do you have a link?

wwmayer:
Works!

davesrocketshop:
> > There is still a known issue when working with pt_BR.UTF8.
> 
> Interesting! How do you know this? Do you have a link?

It's an issue with this test, not the locale (I think). I was getting errors about an illegal character when encoding ascii so I checked which encoder/decoder it was using

wwmayer:
OK. So, is it caused by Python or Qt or FreeCAD?

davesrocketshop:
> OK. So, is it caused by Python or Qt or FreeCAD?

Not Qt or FreeCAD. It might be Python or it might be Ubuntu's locale implementation of pt_BR. @yorikvanhavre might have some insight

wwmayer:
OK, thanks.

PR #10961, opened 2023-10-07
Base: Class - Type checking template functions

abdullahtahiriyo:
Original idea of Werner:
https://github.com/FreeCAD/FreeCAD/pull/10643#issuecomment-1712966545

I really liked this syntax to avoid repetition and even explicit consideration of Types when type checking.

I thought it would be a pity to get it lost.

It includes a small demonstrator of usage.

### DIFF
--- a/src/Base/BaseClass.h
+++ b/src/Base/BaseClass.h
@@ -114,6 +114,18 @@ public:
   virtual void setPyObject(PyObject *);
 
   static void *create(){return nullptr;}
+
+  template <typename T>
+  bool is() const
+  {
+    return getTypeId() == T::getClassTypeId();
+  }
+
+  template <typename T>
+  bool isDerivedFrom() const
+  {
+    return getTypeId().isDerivedFrom(T::getClassTypeId());
+  }
 private:
   static Type classTypeId;
 protected:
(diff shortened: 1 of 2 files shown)


abdullahtahiriyo:
@wwmayer 

I could not resist to have this added.

I have just added an overloading counterpart for isDerivedFrom to take class objects for consistency with the new is() function.

FlachyJoe:
It's very smart as it can be used everywhere in the code.
We could run
```bash
find ./src -type f -name "*.cpp" -exec sed -i 's/getTypeId() == \([A-Za-z0-9_]*\(::[A-Za-z0-9_]*\)*\)::getClassTypeId()/is<\1>()/g' {} +
find ./src -type f -name "*.cpp" -exec sed -i 's/->isDerivedFrom(\([A-Za-z0-9_]*\(::[A-Za-z0-9_]*\)*\)::getClassTypeId()/->isDerivedFrom<\1>()/g' {} +
```

abdullahtahiriyo:
@FlachyJoe 

Sure. 

If Werner agrees to the PR, then we can go WB by WB generating PRs to refactor/modernise the code. 

You may well be my guess and do those PRs...

wwmayer (approved):
Nice idea :)

PR #10962, opened 2023-10-07
Draft: Fix cursor scaling issue

### LINKED ISSUE #6793: Snapping feature will not work under certain conditions

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=3&t=67463

### Version

0.19 (Release)
0.20
0.21

### Full version info

```shell
OS: macOS 10.16
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.19.24366 (Git)
Build type: Release
Python version: 3.9.7
Qt version: 5.12.9
Coin version: 4.0.0
OCC version: 7.5.3
Locale: C/Default (C)
```

```
OS: macOS 10.16
Word size of FreeCAD: 64-bit
Version: 0.20.28751 (Git)
Build type: Release
Python 3.9.12, Qt 5.12.9, Coin 4.0.0, OCC 7.5.3
Locale: C/Default (C)
```

### Subproject(s) affected?

_No response_

### Issue description

I cannot object snap.
Neither the snap icon nor the white dot will appear.
It looks like I encounter this issue while working with my Mac, not when working on my wife's Win11 laptop.

I really have no clue of where the problem is originated, however, as I tried to explain in my topic, **it happens whenever the system cursor is shown instead of the crosshair cursor**.
As I described in my posts, it _may_ not happen if I turn on the computer and don't connect the external display (via USB-C). However, with the same display connected to the Win11 laptop (via HDMI instead), I will **not** experience such issue.

It looks like no-one on the forum could help.
I may use the Win11 computer as a workaround, hoping in a fix.
Unfortunately I can't code, but I may give developers further informations about this problem, if needed.

### Anything else?

_No response_

### Code of Conduct

Roy-043:
Fixes #6793

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_snapper.py
+++ b/src/Mod/Draft/draftguitools/gui_snapper.py
@@ -1164,8 +1164,7 @@ class Snapper:
     def device_pixel_ratio(self):
         device_pixel_ratio = 1
         for w in self.get_quarter_widget(Gui.getMainWindow()):
-            if int(QtCore.qVersion().split('.')[0]) > 4:
-                device_pixel_ratio = w.devicePixelRatio()
+            device_pixel_ratio = w.devicePixelRatio()
         return device_pixel_ratio
 
     def get_cursor_with_tail(self, base_icon_name, tail_icon_name=None):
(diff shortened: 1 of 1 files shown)


yorikvanhavre (approved):
Indeed it seems like double use... Let's see if that works!

PR #10964, opened 2023-10-07
[HOLD] Core: this PR fixes compiler warnings caused by PR #9521

wwmayer:
and most of the coding flaws. Also some trailing whitespaces are fixed


wwmayer:
Before merging this PR it should be checked if most of the PR #9521 should be reverted.

PR #10966, opened 2023-10-07
Fix for workbenchlist

MisterMakerNL:
This fixes an issue with the workbench list where the workbench name has a Qwidget background, while the list has a Qlist background.
Also the Qwidget background was in front of the Qlist hover background color.
 
Before:
freecad_GroeuWxbpF
After:
FreeCAD_8YrQJNyAd5


MisterMakerNL:
I am a bit into a fight with Github although I managed to make 2 pr's I also managed to copy the modified files from my other pr to this branch.

PR #10968, opened 2023-10-07
[stylesheets]Fix for Qcombobox in Qstyle Fusion 

MisterMakerNL:
I finally figured out how to do this PR properly. So I made a new PR.

There was an empty placeholder spot in the Qcombobox in Fusion/Linux style in the stylesheets.
I tried to remove the spot, noticed it was caused by Qcombobox:indicator and I ended up fixing the indicator.
It can be debated if this indicator is useful, as it isn't shown in the no-stylesheet option.
I personally found it looks pretty good and is useful. I'd suggest try it out in main, and we can always remove it afterwards. 
Before:
image
After:
image
image

### DIFF
--- a/src/Gui/Stylesheets/Dark-modern.qss
+++ b/src/Gui/Stylesheets/Dark-modern.qss
@@ -16,11 +16,9 @@ INSTALLATION WHEN NECESSARY
     LINUX = /home/[YOUR_USER_NAME]/.FreeCAD/Gui/Stylesheets/
 
 ============================================================================================================
-TO MODIFY USE THESE COLORS:
- @ThemeAccentColor1
- @ThemeAccentColor2
-
-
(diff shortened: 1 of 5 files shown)


PaddleStroke (commented):
I'm no expert on qss stylesheet so its hard to confirm for sure if everything is ok but it looks ok. @MisterMakerNL is much more expert on this matter.

As to the ui side of it, the indicator check looks good, especially if on linux it was this empty void before. Does it also change on windows @MisterMakerNL ? (Which is fine for me.)

MisterMakerNL:
@PaddleStroke Windows ignores the indicator. This is because it places the active item on top, and so it doesn't need it. 
I btw don't understand why the checkmark has been removed in the no-stylesheet option, it seems to be done on purpose.

wwmayer:
I agree that it looks better than before. So, I will merge it.

PR #10969, opened 2023-10-07
Workbench list in settings, small UI fix.

MisterMakerNL:
This fixes an issue with the workbench list where the workbench name has a Qwidget background, while the list has a Qlist background.
Also the Qwidget background was in front of the Qlist hover background color.
 
Before:
freecad_GroeuWxbpF
After:
FreeCAD_8YrQJNyAd5


PaddleStroke:
Awesome thanks for fixing that.

src/Gui/PreferencePages/DlgSettingsWorkbenchesImp.cpp
```diff
@@ -124,7 +124,8 @@ wbListItem::wbListItem(const QString& wbName, bool enabled, bool startupWb, bool
     subLayout->setAlignment(Qt::AlignLeft);
     subLayout->setContentsMargins(5, 0, 0, 5);
     subWidget->setMinimumSize(250, 0);
-
+    subWidget->setAttribute(Qt::WA_TranslucentBackground);
+
```
luzpaz, line 128: fix trailing whitespace

PR #10970, opened 2023-10-07
Fix build for C++20

### LINKED ISSUE #10952: [Problem] FreeCAD does not compile as C++20 (invalid template arguments on constructor declaration)

### Problem description

FreeCAD does not compile:
```
In file included from freecad/src/CXX/Objects.hxx:42,
                 from freecad/src/Base/PyObjectBase.h:56,
                 from freecad/src/Base/BaseClass.cpp:31:
freecad/src/CXX/Python3/Objects.hxx:1336:21: error: expected unqualified-id before ‘const’
 1336 |         SeqBase<T>( const Object &ob )
      |                     ^~~~~
```

Although the constructor declarator `class_name<T>(...)` is accepted by many compilers, it is not valid C++. It's unclear whether it was *ever* valid C++ (or intended to be valid, at least), but it is explicitly invalid in C++20. GCC ≥ 11 in `-std=c++20` mode or later rejects it (albeit with a rather confusing error message, as above).

See also [CWG Issue 2237](https://cplusplus.github.io/CWG/issues/2237.html).

### Full version info

```shell
master circa 2023-10-04
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

mwoehlke:
Remove template argument list from constructor names. It has never been necessary, and is forbidden in C++20 (n.b. [CWG 2237](https://cplusplus.github.io/CWG/issues/2237.html)). This has also been independently fixed upstream in [r449](https://sourceforge.net/p/cxx/code/449/).

Fixes #10952.

### DIFF
--- a/src/CXX/Python3/Objects.hxx
+++ b/src/CXX/Python3/Objects.hxx
@@ -1321,19 +1321,19 @@ namespace Py
             return PySequence_Length( ptr() );
         }
 
-        explicit SeqBase<T>()
+        explicit SeqBase()
         :Object( PyTuple_New( 0 ), true )
         {
             validate();
         }
 
-        explicit SeqBase<T>( PyObject *pyob, bool owned=false )
+        explicit SeqBase( PyObject *pyob, bool owned=false )
         : Object( pyob, owned )
         {
             validate();
         }
 
-        SeqBase<T>( const Object &ob )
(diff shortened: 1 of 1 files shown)


wwmayer (approved):
No issues as far as I can see.

PR #10978, opened 2023-10-08
Modernize type and derived type checking

FlachyJoe:
Use template functions introduced by #10961 to modernize type checking everywhere.

PR is split in per workbench commit.

the used script to replace and to create one commit per workbench:
```python
import glob
import re

import git

src = "./FreeCAD/src/"

is_pattern = re.compile(r"getTypeId\(\)\s*==\s*([A-Za-z0-9_]*(::[A-Za-z0-9_]*)*)::getClassTypeId\(\)")
is_replace = r"is<\1>()"

deriv_pattern = r"getTypeId\(\).isDerivedFrom\(\s*([A-Za-z0-9_]*(::[A-Za-z0-9_]*)*)::getClassTypeId\(\)\)"
deriv_replace = r"isDerivedFrom<\1>()"

so_pattern = re.compile(r"(^|::)So.*")


def replaceIn(file: str) -> bool:
    lines = []
    modified = False
    try:
        with open(file, 'r', newline='') as fd:
            lines = fd.readlines()

        with open(file, 'w', newline='') as fd:
            lnum = 0
            while (lnum < len(lines)):
                # search in a two lines scrolling window
                text = lines[lnum]
                if (lnum < len(lines) - 1):
                    text += lines[lnum + 1]

                tModif = False
                m = re.search(is_pattern, text)
                if m:
                    if (not re.search(so_pattern, m.group(1))
                        and m.group(1) != 'T'):
                        text = re.sub(is_pattern, is_replace, text)
                    tModif = True
                else:
                    m = re.search(deriv_pattern, text)
                    if m:
                        if (not re.search(so_pattern, m.group(1))
                            and m.group(1) != 'T'):
                            text = re.sub(deriv_pattern, deriv_replace, text)
                        tModif = True
                if tModif:
                    fd.write(text)
                    modified = True
                    lnum += 2
                else:
                    fd.write(lines[lnum])
                    lnum += 1
    except:
        print(f"\n{file} failed.\n")

    return modified


def getMod(file: str) -> str:
    if file.startswith(src + "3rdParty"):
        # Ignore
        return None

    if file.startswith(src + "Mod"):
        # Mod subfolder
        return file[len(src):].split('/')[1]
    else:
        # Core
        return "Core"


def replaceAll() -> list[str]:
    cpp = glob.glob(src + "**/*.cpp", recursive=True)
    h = glob.glob(src + "**/*.h", recursive=True)
    allFiles=[]
    for ext in [cpp, h]:
        for file in ext:
            if replaceIn(file):
                allFiles.append(file)
                print("!", end='')
            else:
                print(".", end='')

    return allFiles


def filesByWb(files: list[str]) -> dict:
    result = {}
    for file in files:
        wb = getMod(file)
        if wb:
            try:
                result[wb].append(file)
            except KeyError:
                result[wb] = [file]
    return result


def createCommits(filesStruct: dict):
    repo = git.Repo(src + "../")
    index = repo.index
    for wb, files in filesStruct.items():
        print(f"Commit changes made in {wb}")
        try:
            index.add(files)
            index.commit(f"{wb}: modernize type checking")
        except git.exc.HookExecutionError:
            print("precommit hook has rewrite some files")
            # re-add and re-commit
            index.update()
            index.add(files)
            index.commit(f"{wb}: modernize type checking")


def main():
    allFiles = replaceAll()
    print(f"\nReplacement done, {len(allFiles)} files changed")
    createCommits(filesByWb(allFiles))


main()
```

ping @abdullahtahiriyo @wwmayer

### DIFF
--- a/src/Mod/Sketcher/App/SketchObject.cpp
+++ b/src/Mod/Sketcher/App/SketchObject.cpp
@@ -918,25 +918,25 @@ Base::Vector3d SketchObject::getPoint(const Part::Geometry *geo, PointPos PosId)
         const Part::GeomPoint *p = static_cast<const Part::GeomPoint*>(geo);
         if (PosId == PointPos::start || PosId == PointPos::mid || PosId == PointPos::end)
             return p->getPoint();
-    } else if (geo->getTypeId() == Part::GeomLineSegment::getClassTypeId()) {
+    } else if (geo->is<Part::GeomLineSegment>()) {
         const Part::GeomLineSegment *lineSeg = static_cast<const Part::GeomLineSegment*>(geo);
         if (PosId == PointPos::start)
             return lineSeg->getStartPoint();
         else if (PosId == PointPos::end)
             return lineSeg->getEndPoint();
-    } else if (geo->getTypeId() == Part::GeomCircle::getClassTypeId()) {
+    } else if (geo->is<Part::GeomCircle>()) {
         const Part::GeomCircle *circle = static_cast<const Part::GeomCircle*>(geo);
         auto pt = circle->getCenter();
         if(PosId != PointPos::mid)
              pt.x += circle->getRadius();
         return pt;
-    } else if (geo->getTypeId() == Part::GeomEllipse::getClassTypeId()) {
+    } else if (geo->is<Part::GeomEllipse>()) {
         const Part::GeomEllipse *ellipse = static_cast<const Part::GeomEllipse*>(geo);
         auto pt = ellipse->getCenter();
         if(PosId != PointPos::mid)
             pt += ellipse->getMajorAxisDir()*ellipse->getMajorRadius();
         return pt;
-    } else if (geo->getTypeId() == Part::GeomArcOfCircle::getClassTypeId()) {
+    } else if (geo->is<Part::GeomArcOfCircle>()) {
         const Part::GeomArcOfCircle *aoc = static_cast<const Part::GeomArcOfCircle*>(geo);
         if (PosId == PointPos::start)
             return aoc->getStartPoint(/*emulateCCW=*/true);
@@ -944,7 +944,7 @@ Base::Vector3d SketchObject::getPoint(const Part::Geometry *geo, PointPos PosId)
             return aoc->getEndPoint(/*emulateCCW=*/true);
         else if (PosId == PointPos::mid)
             return aoc->getCenter();
-    } else if (geo->getTypeId() == Part::GeomArcOfEllipse::getClassTypeId()) {
+    } else if (geo->is<Part::GeomArcOfEllipse>()) {
         const Part::GeomArcOfEllipse *aoc = static_cast<const Part::GeomArcOfEllipse*>(geo);
         if (PosId == PointPos::start)
             return aoc->getStartPoint(/*emulateCCW=*/true);
@@ -952,7 +952,7 @@ Base::Vector3d SketchObject::getPoint(const Part::Geometry *geo, PointPos PosId)
             return aoc->getEndPoint(/*emulateCCW=*/true);
         else if (PosId == PointPos::mid)
             return aoc->getCenter();
-    } else if (geo->getTypeId() == Part::GeomArcOfHyperbola::getClassTypeId()) {
+    } else if (geo->is<Part::GeomArcOfHyperbola>()) {
         const Part::GeomArcOfHyperbola *aoh = static_cast<const Part::GeomArcOfHyperbola*>(geo);
         if (PosId == PointPos::start)
             return aoh->getStartPoint();
@@ -960,7 +960,7 @@ Base::Vector3d SketchObject::getPoint(const Part::Geometry *geo, PointPos PosId)
             return aoh->getEndPoint();
         else if (PosId == PointPos::mid)
             return aoh->getCenter();
-    } else if (geo->getTypeId() == Part::GeomArcOfParabola::getClassTypeId()) {
+    } else if (geo->is<Part::GeomArcOfParabola>()) {
         const Part::GeomArcOfParabola *aop = static_cast<const Part::GeomArcOfParabola*>(geo);
         if (PosId == PointPos::start)
             return aop->getStartPoint();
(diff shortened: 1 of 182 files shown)


FlachyJoe:
hu, it fails.
I replace too much. 
`find . -type f -name "*.h" -exec sed -i 'N;s/getTypeId().isDerivedFrom(\n*[[:space:]]*\([A-Za-z0-9_]*\(::[A-Za-z0-9_]*\)*\)::getClassTypeId())/isDerivedFrom<\1>()/g' {} +`
is ok but 
`find . -type f -name "*.h" -exec sed -i 'N;s/isDerivedFrom(\n*[[:space:]]*\([A-Za-z0-9_]*\(::[A-Za-z0-9_]*\)*\)::getClassTypeId())/isDerivedFrom<\1>()/g' {} +`
is not.

FlachyJoe:
I'm waiting for gh checks but it compiles locally.

wwmayer:
IMO, there is no point in replacing all occurrences of `getTypeId().isDerivedFrom(Class::getClassTypeId()` with `isDerivedFrom<App::PropertyLinkSubList>()` because it doesn't solve anything. But it's fine to use it in new code.

Your replacement script finds a lot of false-positives because it also finds the Inventor classes that have a `isDerivedFrom`method and this results into the build failures for all CI tests.

Pesc0:
I know it's probably too late but I would rename 
```
template <typename T>
bool is() const
```
to 
```
template <typename T>
bool isInstance() const
```
or something more readable

FlachyJoe:
> it doesn't solve anything

But code readability?

> Your replacement script finds a lot of false-positives because it also finds the Inventor classes that have a `isDerivedFrom`method and this results into the build failures for all CI tests.

Yes, I'm aware of this. There are also some more places where it shouldn't be replaced. I'll switch my script to python this evening.

FlachyJoe:
@Pesc0 the function is not used yet so we can change the name. @abdullahtahiriyo what's your opinion?

wwmayer:
> But code readability?

Sure, it makes the code lines a bit shorter.

abdullahtahiriyo:
I do not have a strong opinion on the function name.

my_class.is<Part::Geometry>() does not look unreadable to me. Actually it reads quite well.

my_class.isInstance<Part::Geometry>() does not read as well IMO, as I am missing something (grammatically), as in my_class.isInstanceOf<Part::Geometry>().

If is<> is deemed not readable enough, then we can change it. If is<> is readable enough, the counterargument of changing it is why extend the function name to make no meaningful difference.

wwmayer:
With the template argument the function call `my_class.is<Part::Geometry>()` looks very similar to `my_class.isPartGeometry()` and it's immediately clear what it means. I don't see the advantage of writing `my_class.isInstance<Part::Geometry>()`

Pesc0:
I proposed the change since when I encountered the function for the first time it wasn't immediately clear what it did. 
However in my opinion if it had been named IsInstance it would have been immediately clear, to me at least. My thinking was that there's the python builtin `isinstance` which does the same thing, so I thought that name would be consistent.

abdullahtahiriyo:
"Instance" is not used elsewhere in FreeCAD c++, Type is used instead as Werner indicates.

Then, I tend to prefer the is<> version for conciseness...

src/Mod/Sketcher/App/SketchObject.cpp
FlachyJoe, line 1: Manually reverted at line 8989
```
    auto parabolafound = std::find_if(geometries.begin(), geometries.end(), [](auto g) {
        return g->getTypeId() == Part::GeomArcOfParabola::getClassTypeId();
    });
```

FlachyJoe: Yes, it doesn't compile on linux.
`error: use 'template' keyword to treat 'is' as a dependent template name`

abdullahtahiriyo: ```return g-> template is<Part::GeomArcOfParabola>();```

auto is a template type. Then is a template function operating on a template type.

abdullahtahiriyo: A reference:
https://en.cppreference.com/w/cpp/language/lambda

```For each parameter in params whose type is specified as auto, an invented template parameter is added to template-params, in order of appearance. The invented template parameter may be a [parameter pack](https://en.cppreference.com/w/cpp/language/parameter_pack) if the corresponding function member of params is a function parameter pack. ```

Comment (FlachyJoe) [reply src/Mod/Sketcher/App/SketchObject.cpp:1]: As I understand it only occurs in lambda, I'll definitely have forgotten this syntax before the next time I encounter it. :smile: 

Is the new syntax better? Which of the two will a contributor have more questions about?

Comment (wwmayer) [reply src/Mod/Sketcher/App/SketchObject.cpp:1]: See also https://en.cppreference.com/w/cpp/language/dependent_name (see The typename disambiguator for dependent names)

Comment (FlachyJoe) [reply src/Mod/Sketcher/App/SketchObject.cpp:1]: `auto` replacement by `Part::Geometry*` is valid too
```cpp
    auto parabolafound = std::find_if(geometries.begin(), geometries.end(), [](Part::Geometry* g) {
        return g->is<Part::GeomArcOfParabola>();
    });
```

abdullahtahiriyo: Sure! Then, it is no longer a template parameter. 

I am ok with this one. Probably the syntax by not using auto is clearer.

FlachyJoe: So I do.

chennes:
@FlachyJoe Can you resolve that conflict? Also, it looks like there is still one Inventor line in there, can you double-check it?

FlachyJoe:
Hi @chennes , if you were thinking about View3DInventor, it's a BaseClass child so it's ok:
image

chennes:
I think this is ready to go, since the script was re-written to address the original issues. Even though it's probably not really strictly necessary, I think the change will provide a nice use example for future code, so I'm going to merge it.

PR #10980, opened 2023-10-08
Fem: Remove callback added by filters task box and rework marker classes

marioalexis84:
Currently the callback functions and the connection used to pick the points in the `TaskPostDataAtPoint` and `TaskPostDataAlongLine` functions are not removed properly. So this lead in a crash if, for example, click the `Select Point` button and close the dialog without end the  points selection process since the markers are destroyed but the callback remains active. In fact, each time the button is pressed, a callback and a connection are established.

Additionally, there is some rework in the point marker classes to avoid code duplication and to use pointers to initialize the objects.
The `TaskPostDataAlongLine` and `TaskPostDataAtPoint` boxes are created using pointers to their specific view provider objects.

These changes fix the problem without changing the current code structure too much, but the TaskPostBoxes files could need some rework in general.

### DIFF
--- a/src/Mod/Fem/Gui/TaskPostBoxes.cpp
+++ b/src/Mod/Fem/Gui/TaskPostBoxes.cpp
@@ -71,12 +71,13 @@ using namespace Gui;
 
 // ***************************************************************************
 // point marker
-PointMarker::PointMarker(Gui::View3DInventorViewer* iv, std::string ObjName)
-    : view(iv)
+PointMarker::PointMarker(Gui::View3DInventorViewer* iv, App::DocumentObject* obj)
+    : connSelectPoint(QMetaObject::Connection())
+    , view(iv)
+    , obj(obj)
     , vp(new ViewProviderPointMarker)
 {
     view->addViewProvider(vp);
-    m_name = ObjName;
 }
 
 PointMarker::~PointMarker()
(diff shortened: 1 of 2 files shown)


src/Mod/Fem/Gui/ViewProviderFemPostFilter.cpp
```diff
@@ -49,7 +49,8 @@ ViewProviderFemPostDataAlongLine::~ViewProviderFemPostDataAlongLine() = default;
 void ViewProviderFemPostDataAlongLine::setupTaskDialog(TaskDlgPost* dlg)
 {
     // add the function box
-    dlg->appendBox(new TaskPostDataAlongLine(dlg->getView()));
+    auto view = static_cast<ViewProviderFemPostDataAlongLine*>(dlg->getView());
```
wwmayer, line 52: isn't `view` equal to `this`? If yes, then it's probably best to assert this and pass `this` to TaskPostDataAlongLine.
At this point it's not known which view provider is passed to TaskDlgPost and thus the static_cast is potentially harmful. It is only expected that a ViewProviderFemPostDataAlongLine was passed.

marioalexis84: Yes, `view` is equal `this`.
I have reverted the constructor to receive a `Gui::ViewProviderDocumentObject` and to include the original assertion again.
So your suggestion is to move the assert in the TaskPost constructors to the caller, check if `this` is equal to `dlg->getView()` and change the constructors to use the specific pointer instead of `ViewProviderDocumentObject`?
After all, currently only the type is checked but not the pointer to the view provider.

wwmayer: > So your suggestion is to move the assert

No, my idea was to do this in ViewProviderFemPostDataAlongLine::setupTaskDialog
```
assert(this == dlg->getView());
dlg->appendBox(new TaskPostDataAlongLine(this));
```

Issue #10130, opened 2023-08-09
[Problem] Segmentation fault when clicking Ctrl-Z


rezahousseini:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
OS: Guix System (GNOME/gnome)
Word size of FreeCAD: 64-bit
Version: 0.21.0.Unknown
Build type: Release
Python 3.10.7, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.2, OCC 7.6.2
Locale: English/United States (en_US)
Installed mods: 
  * fasteners 0.4.56
  * Assembly4 0.12.7
```

### Subproject(s) affected?

None

### Problem description

I have a TechDraw from an imported CSG file where I deleted some cut out views and wanted to go back to them by clicking Ctrl-Z. FreeCAD crashed with the following backtrace:

```
Unhandled std::exception caught in GUIApplication::notify.
The error message is: basic_string::_M_construct null not valid
Program received signal SIGSEGV, Segmentation fault.
#0  /gnu/store/gsjczqir1wbz8p770zndrpw4rnppmxi3-glibc-2.35/lib/libc.so.6(+0x3d2a0) [0x7f5fe94412a0]
#1  0x7f5ec3ba3359 in TechDrawGui::ViewProviderViewPart::fixSceneDependencies() from /gnu/store/y2lp2qg6i44d371rsbjzzlr7m7zjf2dw-freecad-0.21.0/lib/TechDrawGui.so+0x19
#2  0x7f5ec3b8c486 in TechDrawGui::ViewProviderPage::fixSceneDependencies() from /gnu/store/y2lp2qg6i44d371rsbjzzlr7m7zjf2dw-freecad-0.21.0/lib/TechDrawGui.so+0x96
#3  0x7f5ec3af631a in TechDrawGui::MDIViewPage::onMsg(char const*, char const**) from /gnu/store/y2lp2qg6i44d371rsbjzzlr7m7zjf2dw-freecad-0.21.0/lib/TechDrawGui.so+0xea
#4  0x7f5fed34dfeb in Gui::Application::sendMsgToActiveView(char const*, char const**) from /gnu/store/y2lp2qg6i44d371rsbjzzlr7m7zjf2dw-freecad-0.21.0/lib/libFreeCADGui.so+0x3b
[...]

PR #10981, opened 2023-10-09
Splits DocumentObject::getNameInDocument() into three methods.

andre-caldas:
This passed:
```
$ bin/FreeCAD --run-test 0
$ bin/FreeCADCmd --run-test 0
```

In FreeCAD, DocumentObject pcNameInDocument has three different uses:
  1. It is the name the object has when it belongs to a document.
  2. It indicates if this object belongs to a document or not.
  3. It provides a CAG key for each object. Some sort of UUID.

The most serious issue is that the method might return a nullptr, but this is hardly ever checked.  In most parts of FreeCAD the developer simply assumed this method provides a valid `const char*`.
This has the potential to cause a crash, as reported here:
https://github.com/FreeCAD/FreeCAD/issues/10130

Therefore, this method was divided into three methods:
  1. Always returns a valid string: const std::string& getNameInDocument().
  2. bool isAttachedToDocument().
  3. const char* getCagKey().

Unfortunately, the string returned by getNameInDocument() ends up being converted to `const char*` most of the time.
This is because in FreeCAD, for dealing with strings, we have:
  1. const char*
  2. std::string
  3. QString from utf8
  4. QString from Latin1 (?!?!?!?!)
  5. QByteArray
  6. Coin library's string object.

Making getNameInDocument() return a `std::string` was important to make the compiler point out in what places a `const char*` was expected. Unfortunately, because FreeCAD does not use std::string in a consistent way, there were lots of "false positive", where `.c_str()` ended up being required. I hope most of those `c_str()` get removed soon. :-)

### DIFF
--- a/src/App/Application.cpp
+++ b/src/App/Application.cpp
@@ -1215,7 +1215,7 @@ std::set<DocumentObject *> Application::getLinksTo(
     } else {
         std::set<Document*> docs;
         for(auto o : obj->getInList()) {
-            if(o && o->getNameInDocument() && docs.insert(o->getDocument()).second) {
+            if(o && o->isAttachedToDocument() && docs.insert(o->getDocument()).second) {
                 o->getDocument()->getLinksTo(links,obj,options,maxCount);
                 if(maxCount && (int)links.size()>=maxCount)
                     break;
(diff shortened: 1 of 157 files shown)


luzpaz:
@andre-caldas checkout 'Setting up for Development'  in https://freecad.github.io/DevelopersHandbook/gettingstarted/ regarding seting up pre-commit (FC's automatic code-formatter and checker)

PR #10993, opened 2023-10-11
[HOLD] Added include missing from #7888

yorikvanhavre:
This adds a precompiled include missing in #7888 - asked by @PaddleStroke

### DIFF
--- a/src/Gui/PreCompiled.h
+++ b/src/Gui/PreCompiled.h
@@ -100,6 +100,7 @@
 // Keep this order to avoid compiler warnings
 #include "QtAll.h"
 #include "InventorAll.h"
+#include <QDesktopWidget>
 
 #elif defined(FC_OS_WIN32)
 #include <windows.h>


PaddleStroke:
Thanks for doing it ! 
Note I think it would be best to put the header in the QtAll.h file instead of directly in the precompiled.h

wwmayer:
QDesktopWidget is a deprecated class and has been removed in Qt6: https://doc.qt.io/qt-5.15/qdesktopwidget.html
Use QScreen instead.

luzpaz:
Should we have an ifdef for Qt5 or are we making 1.0 Qt6 only ?

wwmayer:
No. QDesktopWidget is marked as deprecated since Qt 5.11. So, there is unlikely a system around that still uses a Qt version older than that. Btw, QScreen is available since Qt 5.0

wwmayer:
The issues that I discovered are fixed with: https://github.com/FreeCAD/FreeCAD/pull/11000

chennes:
I'm closing this one since it's superseded by #11000.

PR #10994, opened 2023-10-11
[Gui] Prefer Dark-Outline as default overlay stylesheet - #7888

kadet1090:
As mentioned in https://github.com/FreeCAD/FreeCAD/pull/7888#issuecomment-1752017743

This PR fixes white space problems and changes how stylesheet detection works.

FreeCAD 3D View is often somewhat dark even if the UI is in light colors. So if user has not speficially applied "Light" stylesheet we should prefer to use "Dark-Outline" stylesheet which provides proper and fully transparent overlay panels.


wwmayer:
Code looks good to me. Unfortunately, there is now a merge conflict because of some fixes I made yesterday. Therefore I have opened the new PR #11012

PR #10997, opened 2023-10-11
clean base

mosfet80:
removed unused getA() and setA()

### DIFF
--- a/src/Base/MatrixPyImp.cpp
+++ b/src/Base/MatrixPyImp.cpp
@@ -31,7 +31,6 @@
 #include "QuantityPy.h"
 #include "MatrixPy.h"
 #include "MatrixPy.cpp"
-#include <E57Format.h>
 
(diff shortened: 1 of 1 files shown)


wwmayer:
Who says it's unused? Because it's a Python function nobody knows.

wwmayer:
Function is used in https://github.com/FreeCAD/FreeCAD/blob/main/src/Mod/Test/BaseTests.py#L446

luzpaz:
Is it worth adding a comment (for posterity) ?

wwmayer:
Actually not. It's always a bad idea to simply remove functions from the Python API because there is no control which external scripts use it.

Before a Python function can be removed it must be marked as deprecated and left in the code for at least one or two releases so that users of the function have enough time to port their code. Removing it as an ad-hoc decision is bad practice.

PR #11006, opened 2023-10-12
Improve the  zh-CN localization of *Path* module

zhyifei:
Improve the  zh-CN localization of Path module.


luzpaz:
@zhyifei we use [crowdin](https://crowdin.com/project/freecad) for translations. Were you aware of that?

PR #11008, opened 2023-10-12
Merging the Help module

### LINKED ISSUE #10512: [Problem] Cannot uninstall Help Add-on

### Problem description

Trying to uninstall Help Add-on freezes the program on a window with title "Keep addon settings?"

### Full version info

```shell
[code]
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.21.1.33668 +26 (Git)
Build type: Release
Installed mods: 
  * Help 1.0.3 (Disabled)
[/code]
```

### Subproject(s) affected?

None

### Anything else?

image

### Code of Conduct

### LINKED ISSUE #10527: [Problem] Merge the Help module
### Problem description

The current Help module, currently installable as an add-on, is now the only way to access the FreeCAD documentation from within FreeCAD, as the older way, which relied on enormous, unmanageable Qt help files, has been removed.

When using any of the help features (Help menu, "what's this" button), if the Help module is not installed, you get a "please install the Help add-on" message.

So I think we should add the Help module to the built-in modules, so it is installed by default. The Help module is extremely lightweight, it is composed of basically one Python file plus a couple of additional things (preferences page, etc)

A couple of points that should be fixed before doing so:

* Remove the dependency on the `markdown` python module, or at least still work nicely without it
* Maybe open the wiki by default, instead of the markdown version?

yorikvanhavre:
Merging the help module from https://github.com/FreeCAD/FreeCAD-Help 

fixes #10527
fixes #10512


chennes (commented):
I suggest also adding this to the https://github.com/FreeCAD/FreeCAD/blob/main/.pre-commit-config.yaml (and running pre-commit on it locally to get the appropriate formatting changes).

hasecilu:
Hi, I sent PR on https://github.com/FreeCAD/FreeCAD-Help/pull/21 to change string width in order to have a reasonable size in screen (view images). Do you prefer new PR on this branch?

yorikvanhavre:
@chennes Done!

@hasecilu Are you sure that modifying the file in QDesigner does not reset your changes?

adrianinsaval:
this doesn't add any dependency right?

yorikvanhavre:
> this doesn't add any dependency right?

No. It can make use of the python Markdown or Pandoc modules, but this is entirely optional and not the default, and people willing to turn that on are notified that they need to install one of these modules

If we some day switch to a pure markdown version of the documentation, we might need one of these modules, but that's for another discussion

PR #11010, opened 2023-10-12
Draft: Introduction of the PlaneGui class

Roy-043:
Related issue:
#5603.

Previous PRs:
#10766
#10838
#10931

Short explanation:
* The PlaneGui class handles several GUI related aspects of the WP including a WP history.
* The `get_working_plane` function should be used to get a WP.
* Some of the code was taken from gui_selectplane.py. The intention is to remove almost all WP related code from that file. The icon, tooltip and label for the draftToolBar are also handled by PlaneGui.
* The tooltip shows extended information about the WP (position and axes vectors). An asteriks is appended to the label if the origin does not match the global origin.
* The `_update_all` function still updates FreeCAD.DraftWorkingPlane for compatibility.
* Some confusing names from the Plane class have been changed: `weak` -> `auto`, `reset` -> `set_to_auto`, `setup` -> `auto_align`.

Next step:
Implementing PlaneGui in Draft and Arch. The gui_selectplane.py file will be updated first.

### DIFF
--- a/src/Mod/Draft/WorkingPlane.py
+++ b/src/Mod/Draft/WorkingPlane.py
@@ -36,6 +36,7 @@ YZ, and XZ planes.
 
 import math
 import lazy_loader.lazy_loader as lz
+from PySide.QtCore import QT_TRANSLATE_NOOP
 
 import FreeCAD
 import DraftVecUtils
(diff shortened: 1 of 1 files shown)


yorikvanhavre (approved):

yorikvanhavre:
Seems good!

Issue #6272, opened 2022-02-07
[MSVC][Build][Permissive-][std:c++latest] FreeCAD failed to build on MSVC


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=4765

* **Reporter:** QuellaZhang
* **Date submitted:** 10/15/2021
* **FreeCAD version:** 
* **Category:** Bug
* **Status:** acknowledged
* **Tags:** 

# Original report text

Hi All,

The MSVC team recently added FreeCAD as part of RWC testing to detect compiler regressions. And we found some errors when building FreeCAD with '/permissive-' and '/std:c++latest' option. Can you help to fix these issues? Thanks.

I noticed that some of these errors are come from Boost library, and these issue have been fixed on Boost latest version. After I fixed Boost errors, more errors were exposed by the project itself.

# Steps to reproduce

1. open VS2019 x64 tools command
2. git clone https://github.com/FreeCAD/FreeCAD.git F:\gitP\FreeCAD\FreeCAD
3. download https://github.com/apeltauer/FreeCAD/releases/tag/LibPack_12.5.2 F:\gitP\FreeCAD\tools\FreeCADLibs_12.5.2_x64_VC17
4. mkdir F:\gitP\FreeCAD\FreeCAD\build_amd64 && cd F:\gitP\FreeCAD\FreeCAD\build_amd64
5. set _CL_= /permissive- or set _CL_= /std:c++latest
6. set PATH=F:\gitP\FreeCAD\tools\FreeCADLibs_12.5.2_x64_VC17\bin;%PATH%
7. set PATH=F:\gitP\FreeCAD\tools\FreeCADLibs_12.5.2_x64_VC17\bin\Scripts;%PATH%
8. cmake -G "Visual Studio 16 2019" -A x64 -DCMAKE_SYSTEM_VERSION=10.0.18362.0 -DFREECAD_LIBPACK_DIR=F:\gitP\FreeCAD\tools\FreeCADLibs_12.5.2_x64_VC17 -DBUILD_FEM_NETGEN=OFF -DFREECAD_RELEASE_PDB=OFF -DFREECAD_RELEASE_SEH=OFF -DCMAKE_BUILD_TYPE=Release .. 
9. msbuild /m /p:Platform=x64 /p:Configuration=Release FreeCAD.sln /t:Rebuild

# Other bug information

* **Priority:** normal
* **Severity:** minor
* **Category:** Bug
[...]

PR #11014, opened 2023-10-12
Core: Enable compiling with MSVC /permissive-

chennes:
Split out from the work I was doing to support the next Windows LibPack, these are the non-cMake changes that were required to get it to work, mostly addressing issues with the `/permissive-` flag required by Qt6 on MSVC. Note that this doesn't actually add that flag, it gets automatically enabled when using Qt6.

@wwmayer I had to move the Boost intrusive pointer stuff to inside the Gui namespace to get it to compile, and I am not sure why. It still doesn't compile the same as the GCC/clang versions so is in #ifdef guards, but I think is closer than it was before. Can you double-check this work?

This is at least a partial fix for #6272 (I have not tried using std:c++-latest).


wwmayer:
> Can you double-check this work?

Is the LibPack with Qt6 available? Or is it possible to test it with an older LibPack by simply setting the permissive flag?

wwmayer:
Just to be clear with MSVC the option is called `/permissive-` and not `/fpermissive-`, i.e. without the "f"

src/Gui/3Dconnexion/GuiNativeEventWin32.cpp
```diff
@@ -1071,6 +1074,12 @@ bool Gui::GuiNativeEvent::TranslateSpaceMouseOldGeneric(UINT nInputCode, PRAWINP
     bool processed = false;
     bool bIsForeground = (nInputCode == RIM_INPUT);
 
+    // Initialize with some random bit of data before we look up the real value later on
```
wwmayer, line 1077: Cannot judge on this because for me the compilation of PreCompiled.cpp causes a failure due to boost and thus cannot continue with build FreeCADGui.

src/Main/MainGui.cpp
```diff
@@ -48,6 +48,7 @@
 #include <QApplication>
 #include <QLocale>
 #include <QMessageBox>
+#include <QtQuick/QQuickWindow>
```
wwmayer, line 51: Do we really need this? If not it should be dropped. If yes then it should be visible for MSVC only. For MSYS/MinGW it's not needed either.

src/Mod/TechDraw/App/AppTechDrawPy.cpp
```diff
@@ -961,18 +961,18 @@ class Module : public Py::ExtensionModule<Module>
     {
         PyObject* pFace(nullptr);
         double scale = 1.0;
-        char* pPatName = "";
-        char* pPatFile = "";
+        char* pPatName = nullptr;
```
wwmayer, line 964: Why not using `const char* pPatName = "";`? The you can also avoid the ternary operator below.

src/Mod/TechDraw/App/PreCompiled.h
```diff
@@ -54,14 +54,16 @@
 
 // Qt
 #include <QDomDocument>
-#include "QDomNodeModel.h"
```
wwmayer, line 57: You can completely remove QDomNodeModel.h, QXmlQuery and QXmlResultItems. We do not use it any more, neither for Qt5 nor for Qt6.

src/Mod/TechDraw/Gui/PreCompiled.h
```diff
@@ -57,8 +57,10 @@
 #ifndef _QtAll__
 # include <Gui/QtAll.h>
 #endif
-#include <QXmlQuery>
-#include <QXmlResultItems>
```
wwmayer, line 61: Same here

tests/src/App/ComplexGeoData.cpp
```diff
@@ -74,9 +74,9 @@ class ComplexGeoDataTest: public ::testing::Test
     {
         if (App::Application::GetARGC() == 0) {
             constexpr int argc = 1;
-            std::array<char*, argc> argv {"FreeCAD"};
```
wwmayer, line 77: This block is copied in several places in the tests directory. We should move this to a separate header file. I will prepare a separate PR.

wwmayer:
What's missing are changes that must be applied to LibraryVersions.h.cmake. 

I have tried the `/permissive-` flag with a Qt5 build using an older LibPack and there I run into two unresolvable problems:
* App/Graphviz.cpp includes some boost headers that cause trouble with that flag
* Gui/PreCompiled.cpp includes some boost headers that cause trouble with that flag
Do you have these problems, too?

chennes:
> Do you have these problems, too?

No - probably the Boost problems are solved in the later version of Boost I've got in the new LibPack. So that means that compiling with Qt6 on MSVC is going to absolutely require the newer Boost as well. That shouldn't really be a problem because it will all be in the same LibPack, though.

src/Main/MainGui.cpp
chennes: No, I think that was leftover from when I was fighting with the OpenGL issues in Qt6. I'll remove it.

src/Mod/TechDraw/App/AppTechDrawPy.cpp
chennes: Yes, good idea. I have changed it.

chennes:
> Do you have these problems, too?

No - probably the Boost problems are solved in the later version of Boost I've got in the new LibPack.

wwmayer:
All CI builds fail with this error message:
```
CMake Error at src/Mod/Path/CMakeLists.txt:2 (add_subdirectory):
  The source directory

    /home/runner/work/FreeCAD/FreeCAD/src/Mod/Path/libarea

  does not contain a CMakeLists.txt file.
```

chennes:
Something must have gone awry with the rebase, I have no idea how that file got deleted... let's try this again...

src/Mod/TechDraw/App/AppTechDrawPy.cpp
chennes: On second thought, I'm not confident in my ability to make this change without changing the behavior -- right now it's using an "et" input type in `PyArg_ParseTuple`, which allocates a buffer (that needs to be freed). How can I change this call:
```
        if (!PyArg_ParseTuple(args.ptr(), "O|detet", &pFace, &scale, "utf-8", &pPatName, "utf-8", &pPatFile)) {
```
to use a `const char *` for `pPatName` and `pPatFile` without changing the encoding behavior?

chennes: @wwmayer do you have a minute to help me out here? I don't have a good understanding of why that `et` type is being used.

WandererFan: #11014 Core: Enable compiling with MSVC /permissive-
https://github.com/WandererFan/FreeCAD/tree/PR11014
[TD]Use regular py string not encoded string in PyArg_ParseTuple

Works with áéöĀ in the file name and the pattern name for Draft Hatch.

chennes: Thanks, I've applied that change.

src/Mod/TechDraw/App/AppTechDrawPy.cpp
```diff
@@ -961,19 +961,17 @@ class Module : public Py::ExtensionModule<Module>
...
-        char* pPatName = "";
-        char* pPatFile = "";
+        const char* pPatName = {nullptr};
+        const char* pPatFile = {nullptr};
         TechDraw::DrawViewPart* source = nullptr;
         TopoDS_Face face;
 
-        if (!PyArg_ParseTuple(args.ptr(), "O|detet", &pFace, &scale, "utf-8", &pPatName, "utf-8", &pPatFile)) {
+        if (!PyArg_ParseTuple(args.ptr(), "O|dss", &pFace, &scale, &pPatName, &pPatFile)) {
             throw Py::TypeError("expected (face, [scale], [patName], [patFile])");
```
wwmayer, line 970: Removing the automatic utf-8 encoding may lead to garbage path names. A better workaround might be to implement a helper function around PyMem_Free that accepts a `const char*`

WandererFan: According to the docs, the 's' option provides a utf8 encoded c-string.  

's' for filenames is also used in makeWireString in Part.   If 's' is not suitable here, we should change it in Part also.

wwmayer: I wonder whether "s" and "et" + "utf-8" do the same in Py3.x now. Under Linux I have made a test with a path name containing Japanese characters and this works with the old and new code:
```
import TechDraw

shape = Part.makeBox(1,1,1)
face = shape.Face1
TechDraw.makeGeomHatch(face, 1, "Diamond", "/tmp/ウィキペディアへようこそ/FCPAT.pat")
```

But this should be also tested under Windows to make sure the .pat file can be read-in.

WandererFan: Using "s" seems to work the same as "et" on win.

On Win10 vm:
Directory & filename with odd characters works:
Z:/Patterns/FCPAT_ᐂᐙᑁᒔᓏᑍ.pat       (Diamond)
Z:/PatternsáéöĀ/FCPAT_ᐂᐙᑁᒔᓏᑍ.pat   (Diamond)

Pattern name with odd characters does not work (but does not work in v0.21 either)
Z:/Patterns/FCPAT.pat                (Diamondᐂᐙᑁᒔᓏᑍ) does not work (but does not work in v0.21)
Z:/Patterns/FCPAT.pat                (DiamondáéöĀ) does not work (but does not work in v0.21)

PR and v0.21 both give this error with odd pattern name:
15:08:59  Traceback (most recent call last):
15:08:59    File "C:\Users\wande\Documents\Target\FCDevl\Mod\Draft\draftguitools\gui_hatch.py", line 116, in onFileChanged
15:08:59      patterns = self.getPatterns(filename)
15:08:59    File "C:\Users\wande\Documents\Target\FCDevl\Mod\Draft\draftguitools\gui_hatch.py", line 127, in getPatterns
15:08:59      for line in patfile:
15:08:59    File "C:\Users\wande\Documents\Target\FCDevl\bin\lib\encodings\cp1252.py", line 23, in decode
15:08:59      return codecs.charmap_decode(input,self.errors,decoding_table)[0]
15:08:59  UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 34: character maps to <undefined>

Issue #11004, opened 2023-10-11
[Problem] Huge size of the preferences dialog


wwmayer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

With the merge of the PR https://github.com/FreeCAD/FreeCAD/pull/7888 the preferences dialog now occupies almost the full height of the screen. This is caused by the DlgSettingsUI because it contains more than 30 controls in a vertical layout.

This huge height affects all other preference pages and for most of them there is now a blank area and at the bottom are standard buttons. This leads to a pure usability because one has to move the mouse like crazy in order to accept the changes.

Either the UI tab should be split into two tabs or the whole page should be embedded into a scroll area.

### Full version info

```shell
OS: Ubuntu 22.04.3 LTS (XFCE/xubuntu)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34633 +4 (Git)
Build type: debug
Branch: overlay_fixes
Hash: 31e8c2c80b01b17f9a55a06dd20b9c5d0e14d7ad
Python 3.10.12, Qt 5.15.3, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.1
Locale: German/Germany (de_DE)
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
I think a re-organisation of the page (using 2 column for a start) will help.
Besides many settings seems very advanced. I think some could probably be removed from the preference page. 
Or perhaps at least a 'Show advanced / Hide advanced' button that shows/hides most of these.

PaddleStroke:
Hmm so several things : 
- 'Tree view' could be organized like this to save space: 
![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/22f1a499-2b71-41a2-ad76-07f8dc77ba33)

- 'Overlay' has many preferences that should not be exposed to user at all. It feels like a theme developper dashboard. I think most of them would be fine to be accessible only through parameters and not in preferences.
![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/848a26d0-8c60-489a-9f8b-f55b690fc1cf)

- The UI preference page is not written in a .ui file, but in a python file. I'm not sure what we should do. Should we write a .ui file so that it is the same as other preference pages rather than using this new python template ? Because this python template seems to generating only single rows.

wwmayer:
[...]

PR #11018, opened 2023-10-12
[Gui] Rework preferences navigation as TreeView - Part 1

### LINKED ISSUE #10886: [Problem] Preferences dialog is space-inefficient and inflexible

### Problem description

Navigation in the Preferences dialog is currently done via ListBox with huge icons which takes a lot of space making it hard to navigate around. Users also have to navigate using two styles of navigation - first user needs to pick category from the listbox on the left, and then pick proper tab from the main content window. It is space inefficient and creates a lot of unnecessary scrollable areas and truncated text.

image

Many programs nowadays solve this problem using tree structure for selecting preferences page, this is done mainly for programs with lot of configuration pages. A good example from IntelliJ:

image

### Full version info

```shell
0.22 dev
```

### Subproject(s) affected?

### Anything else?

_No response_

### Code of Conduct

kadet1090:
This PR will introduce new navigation style for preferences dialog based on `QTreeView`. It will consolidate naviagtion into left pane of the settings dialog removing tab navigation from the main view. In theory it also allows deeper structures for preferences to be introduced but it is not recommended and not used as of now. 

This PR also is a major refactor of the `DlgPreferencesImp` class. I've tried to simplify overall structure and remove code that is no longer needed due to previous updates and things that are no longer problem. I do not see any issues with the current state of the UI but it is possible that I removed a little bit too much.

I'm also far from being a QT expert so there may be some questionable solutions that should be implemented in different way - feel free to point me for proper solutions if that is the case.

Part 1 for #10886

### DIFF
--- a/src/Gui/DlgPreferences.ui
+++ b/src/Gui/DlgPreferences.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>570</width>
-    <height>454</height>
+    <width>800</width>
+    <height>600</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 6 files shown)


kadet1090:
The UI bit of this PR is still to be properly done, as this is how it looks currently without any special care about styling:
image

It is fully working however, so the bit associated with refactor mentioned in the description is ready to review.

PaddleStroke:
In workbenches, do you have one entry for each installed wb that has a pref page?
Currently only loaded wb appear. It would be best if all installed wb appear, and clicking the entry load the wb so that it can show the preference page.

kadet1090:
I didn't touch how and when pages are created and managed. I don't know if it is technically possible - is it required for workbench to provide preference pages?

PaddleStroke:
So it would be like this : 
image

PaddleStroke:
Ok. Well maybe its best to keep that for a latter PR then. Not all workbenches have pref page I think.

kadet1090:
It should be another PR for sure. Unfortunately I don't think that it would be possible to implement if we do not require workbenches to have preference pages. Without loading them we don't know if they provide any and so we don't know if we should display entry in the menu. If there is any way to know if workbench provide at least one preference page I think that I can implement that later.

PaddleStroke:
You are right. We could show all of them, and just display nothing if it doesn't have a pref page. But that's not ideal.

Well maybe even without changing this, the workbenches could go as children of the workbench entry. Rather than independant entries.

kadet1090:
I've added default styling to the treeview and rebased changes. I don't think that I'll be able to improve much without using stuff like `QItemDelegate` or `QProxyStyle`. I think that it'll be wise to split this MR into two parts:
- Part 1 comes with refactor of the dialog and basic treeview without custom delegate and style applied
- Part 2 will introduce better styling more inline with that I've designed in Figma 

This is how it looks for Part 1 - it is still a improvement I think so it is ready to be reviewed and merged.
image

luzpaz:
Could we find consensus around grouping all external addons in to their own tree at the bottom of the left navigation pane ? If so I'll open a ticket for that.

sliptonic:
> Could we find consensus around grouping all external addons in to their own tree at the bottom of the left navigation pane ? If so I'll open a ticket for that.

If a workbench is an addon, should it go in the workbenches section or the addon section? 

I'm not clear what the consensus is about preference pages for non-loaded modules.
It has always been confusing for users that the content of the preference page changes depending on which workbenches you've visited.  This problem could be alleviated just by having either two trees or a section of tree for "loaded modules" and then add to that.   That would be a clue to the user.

kadet1090:
We briefly discussed that with @chennes on discord. Basically we should expose some API for workbenches to inform FreeCAD in advance of potential workbenches before loading. Then we can think about grouping them

I personally don't think that it is good idea to group workbenches to another level as we have for example Part and Part design combined. As far as I understand the API adsons can register groups and pages that are not related to workbench specifically so such change would required more action than it seems at first glance. 

Nonetheless it is surely out of scope for this PR and issue.

CalligaroV:
Hi @kadet1090.

I see that you made changes also to `DlgPreferencesImp::restorePageDefaults`

on #10688 wwmayer suggested to move that function to `Gui::Dialog::PreferencePage`.

we briefly talked about this during PR Meeting 06/11/2023 and probably the best action is to merge your PR (which looks already mergeable) after applying wwmayer's suggestion.

I'm available to talk about that and contribute with coding

kadet1090:
Yeah, I totally agree with that suggestion, even considered that already. I should be able to implement that today.

CalligaroV:
Awesome!

are you planning to add that change in this PR/Branch or opening a new one?

kadet1090:
Depends, If we only want to move current logic to `Gui::Dialog::PreferencePage` then I think that this is a good PR to do that. I can basically push that right away.

However, looking at the mentioned class it already contains member like `saveSettings()` and `loadSettings()` that are virtual and abstract. That methods have no default implementation, but `PreferenceUiForm` contains some basic implementation which also goes through all `PrefWidget`s and calls proper methods. I think that `resetToDefaults()` should be implemented in a consistent manner. This forces me to implement special cases for multiple classes (like `DlgSettingsEditor`) - which itself is a good thing, but I don't think that this PR is the one to do that. 

@wwmayer As you are the one who proposed that change, what do you think?

chennes:
I'm going to merge this so we can move forward: any further feedback can be incorporated into a new PR. Thanks!

PaddleStroke:
@kadet1090 this merge unfortunately made the problem https://github.com/FreeCAD/FreeCAD/issues/11004#issuecomment-1825880450 much worse.
If you have some time, can you consider : 
- Re-Enabling resizing of the preference page
- Doing the rework on the UI tab so that it's not so big anymore

That would be awesome

kadet1090:
Yes, I am aware of that issue, but was blocked by the #11449.  I'll try to introduce partial fix tomorrow that will allow the window to be resized but can behave a little bit strangely and fix that properly later as I had problems to come up with solution that works without any issues.

PaddleStroke:
Alternatively you can fix the UI page, this way resizing will not be a necessity anymore 😄

PR #11026, opened 2023-10-13
Toponaming card1.5 toposhape

Pesc0:
Gave a shot at this BUT
- I feel like I'm touching delicate stuff and dont want to introduce bugs
- Hit a wall where I'm getting errors I have no clue how to fix, because I simply don't know enough about FC code in general

So if someone wants to take it from here...


Pesc0:
@JohnAD @chennes I'll delete this then. might want to take a look at the last 2 commits in case they're useful

PR #11027, opened 2023-10-13
Gui: fixes #11016: Taskview could resize automatically to minimum size

### LINKED ISSUE #11016: [Problem] Taskview could resize automatically to minimum size

### Problem description

The taskview is currently resizing their width to avoid cropping the taskboxes. Which is good.

However they do not shrink back.
So if you open a taskbox that is super wide (partDesign Hole tool for example). Then you have a huge taskview. And if you go to sketcher it looks like this : 
image
So the user have to manually shrink the view back down.
Suggestion : Taskbox resize the width automatically to the current taskboxes.

@wwmayer ping as we discussed this before.

### Full version info

```shell
[code]
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34642 +15 (Git)
Build type: Release
Branch: asm_card9
Hash: 9cf8885ec1ab18b8d7fdc838ac0b8180ebb7bb21
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk , OCC 7.6.3
Installed mods: 
  * A2plus 0.4.60k
  * Assembly3 0.12.0
  * Assembly4 0.50.2
  * Curves 0.6.8
  * ExtremeProLight 1.5.0
  * FreeCAD - Shortcut.lnk
  * FreeCAD_FastenersWB-master 0.4.56
  * Ondsel 2023.8.23.02
  * OpenDark 2023.10.6
  * ProDarkThemePreferencePack 1.0.0
  * sheetmetal 0.2.63
[/code]
```

### Subproject(s) affected?

Core

### Anything else?

_No response_

### Code of Conduct

wwmayer:


PaddleStroke:
What about setting this as default and fixing the taskboxes which have a problem for their minimum size hints ?

PaddleStroke:
Hmm. Then what about implementing a 'auto-task' setting similar to the one in overlay.
So when no task boxes, the taskview hides.
When a new task is started, first the taskview is resized. Then it is showed. This way it wouldn't flicker to the correct size.

maxwxyz:
@wwmayer @PaddleStroke will this PR be updated then or is there a separate (followup) PR?
If it is to be considered for a version 1.0 release, it should be marked _ready for review_ by **Monday, 2024-06-03**

wwmayer:
If time allows I will try to finish it before the freeze.

chennes:
Fixes #11016

adrianinsaval:
I don't quite understand the effect this PR is supposed to have. I'm not seeing any automatic shrinking

src/Gui/TaskView/TaskView.h
```diff
@@ -199,6 +205,8 @@ protected Q_SLOTS:
     QSint::ActionPanel* taskPanel;
     TaskDialog *ActiveDialog;
     TaskEditControl *ActiveCtrl;
+    bool restoreWidth = false;
```
PaddleStroke, line 208: It seems that this new feature is deactivated by default.

The boolean Preferences/DockWindows/TaskView/RestoreWidth has to be manually set to true.

PaddleStroke:
However I have just tested it and I cannot get this to work.
I have set the boolean to true. But : 

- If task View is NOT overlay : Nothing happens.
- If task view is overlay : Nothing happens but an error is printed : 

> QMainWidget::resizeDocks: one QDockWidget is not part of the layout

wwmayer:
I have tested it with Qt5 & Qt6 and with and without overlay handling. For none of the four combinations I get this warning: QMainWidget::resizeDocks: one QDockWidget is not part of the layout

Issue #11023, opened 2023-10-13
[Overlay] Overlay stylesheet should be set by preference pack/theme.


MisterMakerNL:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

I am not sure if it is broken or it is not implemented, but I cannot set the overlay stylesheet with the preference packs.
Which would be recommended as the preference pack sets the background color and some Qsint colors have overlap. 
Tried adding: 
`        <FCParamGroup Name="MainWindow">
          <FCText Name="OverlayStyleSheet">Dark.qss</FCText>
        </FCParamGroup>`

### Full version info

```shell
Latest dev
```

### Subproject(s) affected?

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

obelisk79:
from my user config file it looks like the user.cfg line we need to add looks more like this compared to what you have above: <FCText Name="OverlayActiveStyleSheet">OpenDark_Overlay.qss</FCText>

however, I cannot get my overlay stylesheet to even register in freecad at all. I've tried a few different approaches in my package.xml to register the file but none seem to work.

obelisk79:
@chennes is something that would fall under your purview?

chennes:
Yes. The way it is *supposed* to work as that any QSS file in your mod's directory is found (you don't need to set anything in package.xml). It sounds like that isn't working, so I will have to look into it.

PR #11039, opened 2023-10-14
Gui: Register overlay search path from preference packs

kadet1090:
This PR adds `"<mod>/overlay"` path to search paths of "overlay" prefix. This allows preference packs to register custom overlay stylesheets.

I think that we should think about better ways of doing that kind of stuff. Package metadata could include elements like 
```
<searchpath prefix="qss" path="." />
<searchpath prefix="overlay" path="./overlay" />
```
which should enable package developers to have better control over the search paths. This should however be done in another request and issue.

Fixes: #11023


obelisk79:
I have tested and confirmed that this works.

chennes:
OK, I'll merge this, but I agree that we should reconsider this implementation to allow direct specification of those paths.

Issue #11015, opened 2023-10-12
[Overlay] 'Auto Mouse Wheel pass through' doesn't work on techdraw


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

On windows.

'Auto Mouse Wheel pass through' doesn't work.

The mouse clicks are passing through, but the wheel event doesn't seem to pass through.

@realthunder 

### Full version info

```shell
[code]
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34642 +15 (Git)
Build type: Release
Branch: asm_card9
Hash: 9cf8885ec1ab18b8d7fdc838ac0b8180ebb7bb21
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk , OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * A2plus 0.4.60k
  * Assembly3 0.12.0
  * Assembly4 0.50.2
  * Curves 0.6.8
  * ExtremeProLight 1.5.0
  * FreeCAD - Shortcut.lnk
  * FreeCAD_FastenersWB-master 0.4.56
  * Ondsel 2023.8.23.02
  * OpenDark 2023.10.6
  * ProDarkThemePreferencePack 1.0.0
  * sheetmetal 0.2.63
[/code]
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kadet1090:
I could try to fix that and refactor this part of code as it was mentioned in the CR if I remember correctly.

adrianinsaval:
I can confirm this too. I tried playing around with the delay value for mouse wheel pass through but didn't find anything that worked

realthunder:
Probably broken by [this](https://github.com/realthunder/FreeCAD/commit/bdf7af1380c932371f4d01dab243c832dd2c68ec). My Windows build machine is broken at the moment. Please revert this commit, and give it a try.

kadet1090:
Yup, it indeed fixes the mentioned issue. I reverted only part of this commit as it was enough to fix the issue, I don't want to break other stuff by accident and change that I reverted is isolated to wheel events.

adrianinsaval:
The issue has been mostly solved but there is a remaining problem with techdraw

WandererFan:
a) what is [Overlay]?
[...]

PR #11042, opened 2023-10-14
Gui: Target proper widget when passing wheel event

kadet1090:
As mentioned by @realthunder in https://github.com/FreeCAD/FreeCAD/issues/11015#issuecomment-1762922058 the https://github.com/realthunder/FreeCAD/commit/bdf7af1380c932371f4d01dab243c832dd2c68ec commit broke wheel event passing in overlay widgets. Reverting one block fixes the issue, I did not revert the other change as this function is used also in other places which seems to work fine.

Fixes: #11015

### DIFF
--- a/src/Gui/OverlayManager.cpp
+++ b/src/Gui/OverlayManager.cpp
@@ -1947,6 +1947,13 @@ void OverlayManager::Private::interceptEvent(QWidget *widget, QEvent *ev)
         QPoint globalPos = we->globalPosition().toPoint();
 #endif
         lastIntercept = getChildAt(widget, globalPos);
+
+        for (auto parent = lastIntercept->parentWidget(); parent; parent = parent->parentWidget()) {
+            if (qobject_cast<QGraphicsView*>(parent)) {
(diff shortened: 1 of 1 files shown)


adrianinsaval:
@realthunder can you confirm this is ok or should the whole commit be reverted?

realthunder:
Could you please try on Windows to see if wheel pass through works for TechDraw view? Linux is working fine.

MisterMakerNL:
Does not work on Techdraw, does work in the 3D view in windows.

kadet1090:
Ok, on my setup there is also problem with techdraw canvas. However even reverting whole commit does not fix that - I'll investigate it further

MisterMakerNL:
> So while there is a remaining problem with techdraw on windows I think this PR at least improves the situation for the rest, so I will merge this and further fixes can be made on follow up PR

Totally agree, don't even care that much about tech-draw. As you most of the time zoom into an object in a drawing which most of the time are in the middle of the screen.

PR #11051, opened 2023-10-15
Gui: Remove CornerNaviCube from OverlayParams and use directly in OverlayManager

### LINKED ISSUE #11020: [Problem] Overlay mode "Check Navigation Cube" feature doesn't get correct navi cube position

### Problem description

Enabling Preferences->Display->UI->Check Navigation Cube doesn't work correctly, it seems to always interpret as if the cube was in the lower right corner, shortening widgets on that side while leaving widgets on the correct side unaffected. Might be due to changes in the NaviCube after @realthunder 's initial implementation of this feature.
image

### Full version info

```shell
[code]
OS: Arch Linux (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34651 (Git) AppImage
Build type: Release
Branch: main
Hash: aa5bea4e2e8089bf00f61075a0860dacda42dc8b
Locale: C/Default (C) [ OS: English/United Kingdom (en_GB) ]
Installed mods: 
  * fasteners 0.4.66
  * CfdOF 1.21.7
  * Curves 0.6.13
  * Defeaturing 1.2.1
  * DynamicData 2.53.0
  * Dracula 0.0.5
  * SelectorToolbar
  * FC_SU
  * A2plus 0.4.60n
  * CurvedShapes 1.0.4
  * sheetmetal 0.3.1
  * AirPlaneDesign
  * Glass (Disabled)
  * Assembly3 0.12.0
  * OpenDark 2023.10.9
[/code]
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

Rexbas:
Fixes #11020.

In the file I changed, OverlayParams.cpp, I see a bunch of `// Auto generated code (Tools/params_utils.py:xxx)`. Not sure how that auto generation works but this makes me wonder if the manual changes may get overwritten again if the code is regenerated for some reason.

### DIFF
--- a/src/Gui/OverlayManager.cpp
+++ b/src/Gui/OverlayManager.cpp
@@ -663,8 +663,11 @@ public:
             h -= tabbar->height();
 
         int naviCubeSize = NaviCube::getNaviCubeSize();
-        int naviCorner = OverlayParams::getDockOverlayCheckNaviCube() ?
-            OverlayParams::getCornerNaviCube() : -1;
+        int naviCorner = OverlayParams::getDockOverlayCheckNaviCube()
+            ? App::GetApplication()
+                  .GetParameterGroupByPath("User parameter:BaseApp/Preferences/NaviCube")
+                  ->GetInt("CornerNaviCube", 1)
+            : -1;
 
         QRect rect;
         QRect rectBottom(0,0,0,0);
--- a/src/Gui/OverlayParams.cpp
+++ b/src/Gui/OverlayParams.cpp
@@ -28,21 +28,20 @@ import OverlayParams
 OverlayParams.define()
 ]]]*/
 
-// Auto generated code (Tools/params_utils.py:166)
+// Auto generated code (Tools/params_utils.py:196)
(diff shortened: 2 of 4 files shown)


chennes:
It was created using cog: do we have the base Python file in our repo? If so, we should probably consider making the changes there instead.

Rexbas:
It points to [this](https://github.com/FreeCAD/FreeCAD/blob/main/src/Tools/params_utils.py) which is in the repo. I will see if I can make a change there instead.

Rexbas:
It was actually easier and less messy to remove "CornerNaviCube" from OverlayParams and use it directly in OverlayManager instead. Also removed the line from the Python file used by cog and regenerated OverlayParams.

adrianinsaval:
I would assume it was in OverlayParams for a reason, @realthunder could you comment on this?

adrianinsaval:
I tested and this worked correctly. So at a minimum it's better than the current state.

merging

PR #11053, opened 2023-10-15
Splits DocumentObject::getNameInDocument() into three methods.

andre-caldas:
In FreeCAD, DocumentObject pcNameInDocument has three different uses:
  1. It is the name the object has when it belongs to a document.
  2. It indicates if this object belongs to a document or not.
  3. It provides a DAG key for each object. Some sort of UUID.

The most serious issue is that the method might return a nullptr, but this is hardly ever checked.  In most parts of FreeCAD the developer simply assumed this method provides a valid const char*.

With this patch, this method was divided into three methods:
  1. Always returns a valid string: const std::string& getNameInDocument().
  2. bool isAttachedToDocument().
  3. const char* getDagKey().
 

Unfortunately, the string returned by getNameInDocument() ends up being converted to `const char*` most of the time.
This is because in FreeCAD, for dealing with strings, we have:
  1. const char*
  2. std::string
  3. QString from utf8
  4. QString from Latin1 (?!?!?!?!)
  5. QByteArray
  6. Coin library's string object.

Making getNameInDocument() return a `std::string` was important to make the compiler point out in what places a `const char*` was expected. Unfortunately, because FreeCAD does not use std::string in a consistent way, there were lots of "false positive", where `.c_str()` ended up being required. I hope most of those `c_str()` get removed soon. :-)

Now, I do realize this is such a big patch and that there are too many things to check. So, I do totally understand if this is simply rejected. I do believe that the change is very important, because I think that DocumentObject should be functional even if it is not attached to any documents. And "getNameInDocument()" returning nullptr makes FreeCAD have many places to crash when the object does not belong to any document.

### DIFF
--- a/src/App/Document.cpp
+++ b/src/App/Document.cpp
@@ -1113,7 +1113,7 @@ void Document::exportObjects(const std::vector<App::DocumentObject*>& obj, std::
 
     if(FC_LOG_INSTANCE.isEnabled(FC_LOGLEVEL_LOG)) {
         for(auto o : obj) {
-            if(o && o->getNameInDocument()) {
+            if(o && o->isAttachedToDocument()) {
                 FC_LOG("exporting " << o->getFullName());
                 if (!o->getPropertyByName("_ObjectUUID")) {
                     auto prop = static_cast<PropertyUUID*>(o->addDynamicProperty(
@@ -1187,7 +1187,7 @@ void Document::writeObjects(const std::vector<App::DocumentObject*>& obj,
             for(auto dep : outList) {
                 auto name = dep?dep->getNameInDocument():"";
                 writer.Stream() << writer.ind() << "<" FC_ELEMENT_OBJECT_DEP " "
-                    FC_ATTR_DEP_OBJ_NAME "=\"" << (name?name:"") << "\"/>" << endl;
+                    FC_ATTR_DEP_OBJ_NAME "=\"" << name << "\"/>" << endl;
             }
             writer.decInd();
             writer.Stream() << writer.ind() << "</" FC_ELEMENT_OBJECT_DEPS ">" << endl;
@@ -1378,7 +1378,7 @@ Document::readObjects(Base::XMLReader& reader)
                 objs.push_back(obj);
                 // use this name for the later access because an object with
                 // the given name may already exist
-                reader.addName(name.c_str(), obj->getNameInDocument());
+                reader.addName(name.c_str(), obj->getNameInDocument().c_str());
 
                 // restore touch/error status flags
                 if (reader.hasAttribute("Touched")) {
@@ -1479,7 +1479,7 @@ Document::importObjects(Base::XMLReader& reader)
 
     std::vector<App::DocumentObject*> objs = readObjects(reader);
     for(auto o : objs) {
-        if(o && o->getNameInDocument()) {
+        if(o && o->isAttachedToDocument()) {
             o->setStatus(App::ObjImporting,true);
             FC_LOG("importing " << o->getFullName());
             if (auto propUUID = Base::freecad_dynamic_cast<PropertyUUID>(
@@ -1506,7 +1506,7 @@ Document::importObjects(Base::XMLReader& reader)
     signalFinishImportObjects(objs);
 
     for(auto o : objs) {
-        if(o && o->getNameInDocument())
+        if(o && o->isAttachedToDocument())
             o->setStatus(App::ObjImporting,false);
     }
 
@@ -2411,7 +2411,7 @@ static void _buildDependencyList(const std::vector<App::DocumentObject*> &object
         while(!objs.empty()) {
             auto obj = objs.front();
             objs.pop_front();
-            if(!obj || !obj->getNameInDocument())
+            if(!obj || !obj->isAttachedToDocument())
                 continue;
 
             auto it = outLists.find(obj);
@@ -2438,7 +2438,7 @@ static void _buildDependencyList(const std::vector<App::DocumentObject*> &object
     if(objectMap && depList) {
         for (const auto &v : outLists) {
             for(auto obj : v.second) {
-                if(obj && obj->getNameInDocument())
+                if(obj && obj->isAttachedToDocument())
                     add_edge((*objectMap)[v.first],(*objectMap)[obj],*depList);
             }
         }
@@ -2833,7 +2833,7 @@ int Document::recompute(const std::vector<App::DocumentObject*> &objs, bool forc
             FC_LOG("Recompute pass " << passes);
             for (; idx < topoSortedObjects.size(); ++idx) {
                 auto obj = topoSortedObjects[idx];
-                if(!obj->getNameInDocument() || filter.find(obj)!=filter.end())
+                if(!obj->isAttachedToDocument() || filter.find(obj)!=filter.end())
                     continue;
                 // ask the object if it should be recomputed
                 bool doRecompute = false;
@@ -2890,7 +2890,7 @@ int Document::recompute(const std::vector<App::DocumentObject*> &objs, bool forc
     FC_TIME_LOG(t2, "Recompute");
 
     for(auto obj : topoSortedObjects) {
-        if(!obj->getNameInDocument())
+        if(!obj->isAttachedToDocument())
             continue;
         obj->setStatus(ObjectStatus::PendingRecompute,false);
         obj->setStatus(ObjectStatus::Recompute2,false);
@@ -3051,8 +3051,8 @@ std::vector<App::DocumentObject*> DocumentP::topologicalSort(const std::vector<A
 
     for (auto objectIt : objects) {
         // We now support externally linked objects
-        // if(!obj->getNameInDocument() || obj->getDocument()!=this)
-        if(!objectIt->getNameInDocument())
+        // if(!obj->isAttachedToDocument() || obj->getDocument()!=this)
+        if(!objectIt->isAttachedToDocument())
             continue;
         //we need inlist with unique entries
         auto in = objectIt->getInList();
@@ -3166,7 +3166,7 @@ bool Document::recomputeFeature(DocumentObject* Feat, bool recursive)
     d->clearRecomputeLog(Feat);
 
     // verify that the feature is (active) part of the document
-    if (Feat->getNameInDocument()) {
+    if (Feat->isAttachedToDocument()) {
         if(recursive) {
             bool hasError = false;
             recompute({Feat},true,&hasError);
@@ -3448,6 +3448,11 @@ void Document::_addObject(DocumentObject* pcObject, const char* pObjectName)
 }
 
 /// Remove an object out of the document
+void Document::removeObject(const std::string& name)
+{
+    removeObject(name.c_str());
+}
+
 void Document::removeObject(const char* sName)
 {
     auto pos = d->objectMap.find(sName);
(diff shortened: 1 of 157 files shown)


WandererFan:
It might be easier to review and merge this if we split it into 1 pr that introduces new methods, then a series of PRs that make use of the new methods in stages.

andre-caldas:
> It might be easier to review and merge this if we split it into 1 pr that introduces new methods, then a series of PRs that make use of the new methods in stages.

Very true! :-)

Now that it is done, I think I can make it work gradually.
1. Create a "isAttachedToDocument()" and use it "everywhere".
2. Create a "getDagKey()" and use it "everywhere". (does anyone have a better name?)
3. Substitute every "getNameInDocument()" by "_getNameInDocument()" with underscore.
4. Create a "std::string getNameInDocument()".
5. Gradually adopt "getNameInDocument()" with many patches.

I can do that... but it would be nice to have some kind of "institutional support"... :-)
I mean, it could be good if FreeCAD or someone important announces this effort and ask people to contribute. In special, with item "5".

Doing item "5" gradually would possibly reduce the number of needed ".c_str()". For example, there are some "variadic argument macros/functions" that could be substituted by templates. For example, "Gui::doCommand".

andre-caldas:
>     1. Create a "isAttachedToDocument()" and use it "everywhere".
>     2. Create a "getDagKey()" and use it "everywhere". (does anyone have a better name?)

I have managed to produce two PRs with those two changes above.


>     3. Substitute every "getNameInDocument()" by "_getNameInDocument()" with underscore.
>     4. Create a "std::string getNameInDocument()".

If no one is against it, I will proceed to steps 3 and 4.

I will do items 3 and 4 at once.

After that, I will proceed to item 5:
> Gradually adopt the new `getNameInDocument()`.

After adopting the new `getNameInDocument()` wherever it is relativelly simple to do so, I will also mark the underscore version (`_getNameInDocument()`) as `[[deprecated]]`. See [C++14 deprecated attribute](https://en.cppreference.com/w/cpp/language/attributes/deprecated).

This way, the remaining `_getNameInDocument()` calls will start emitting a warning. Fixing those warnings is a little more difficult because FreeCAD uses C-style variadic functions (functions like printf), and those do not take an `std::string` as an argument. This will be resolved after we replace the "variadic functions" by "templated functions with variadic template arguments".

src/App/Document.cpp
```diff
@@ -3817,7 +3822,12 @@ DocumentObject * Document::getActiveObject() const
     return d->activeObject;
 }
 
-DocumentObject * Document::getObject(const char *Name) const
+DocumentObject * Document::getObject(const char* Name) const
+{
+    return getObject(std::string(Name));
```
wwmayer, line 3827: Doing it this way is bad because every time when calling the function a tmp. std::string will be created. It makes much more sense to have the overloaded method:
```
DocumentObject * Document::getObject(const std::string& Name) const
{
    return getObject(Name.c_str());
}
```

src/App/Document.cpp
```diff
@@ -3817,7 +3822,12 @@ DocumentObject * Document::getActiveObject() const
     return d->activeObject;
 }
 
-DocumentObject * Document::getObject(const char *Name) const
+DocumentObject * Document::getObject(const char* Name) const
+{
+    return getObject(std::string(Name));
+}
+
+DocumentObject * Document::getObject(const std::string Name) const
```
wwmayer, line 3830: Argument should be `const std::string&` as otherwise it creates a tmp. std::string

src/App/DocumentObject.cpp
```diff
@@ -292,36 +292,48 @@ std::string DocumentObject::getFullLabel() const {
     return name;
 }
 
-const char *DocumentObject::getNameInDocument() const
+const char* DocumentObject::getDagKey() const
```
wwmayer, line 295: What does "Dag" mean? Directed acyclic graph?

src/App/DocumentObject.cpp
```diff
@@ -292,36 +292,48 @@ std::string DocumentObject::getFullLabel() const {
     return name;
 }
 
-const char *DocumentObject::getNameInDocument() const
+const char* DocumentObject::getDagKey() const
+{
+    if(!pcNameInDocument)
+    {
+        return nullptr;
```
wwmayer, line 299: And what's better compared to getNameInDocument() when it returns a null pointer here?

src/App/DocumentObjectPyImp.cpp
```diff
@@ -52,11 +52,11 @@ std::string DocumentObjectPy::representation() const
 Py::String DocumentObjectPy::getName() const
 {
     DocumentObject* object = this->getDocumentObjectPtr();
-    const char* internal = object->getNameInDocument();
-    if (!internal) {
+    std::string internal = object->getNameInDocument();
```
wwmayer, line 55: Should be const reference.

src/App/DocumentObjectPyImp.cpp
```diff
@@ -224,8 +224,8 @@ Py::Object DocumentObjectPy::getViewObject() const
         if(!getDocumentObjectPtr()->getDocument()) {
             throw Py::RuntimeError("Object has no document");
         }
-        const char* internalName = getDocumentObjectPtr()->getNameInDocument();
-        if (!internalName) {
+        std::string internalName = getDocumentObjectPtr()->getNameInDocument();
```
wwmayer, line 227: Should be const reference.

src/App/PropertyLinks.cpp
```diff
@@ -1246,7 +1246,7 @@ std::string PropertyLinkBase::tryImportSubName(const App::DocumentObject *obj, c
         if(next[0] == '$') {
             if(strcmp(next+1,sobj->Label.getValue())!=0)
                 continue;
-        } else if(strcmp(next,sobj->getNameInDocument())!=0) {
+        } else if(strcmp(next,sobj->getNameInDocument().c_str())!=0) {
```
wwmayer, line 1249: Here you can write: `else if (sobj->getNameInDocument() != next)`

src/App/PropertyStandard.cpp
```diff
@@ -1359,21 +1359,21 @@ void PropertyString::setValue(const char* newLabel)
 
                 bool changed = false;
                 label = label.substr(0,lastpos+1);
-                if(label != obj->getNameInDocument()
+                std::string objName = obj->getNameInDocument();
```
wwmayer, line 1362: const reference

andre-caldas:
Dear @wwmayer,

As far as I have understood, this PR will not be applied. It is being split in smaller PRs.

I am sorry for making you waste your time... the code will not be touched anymore. But if you want to review the two first PRs...
https://github.com/FreeCAD/FreeCAD/pull/11139
https://github.com/FreeCAD/FreeCAD/pull/11141

PS: I will close this PR, but I hope it can still be used to coordinate the split PRs.

src/App/PropertyStandard.cpp
```diff
@@ -1359,21 +1359,21 @@ void PropertyString::setValue(const char* newLabel)
...
-                if(label != obj->getNameInDocument()
+                std::string objName = obj->getNameInDocument();
+                if(label != objName
                         && boost::starts_with(obj->getNameInDocument(),label))
                 {
                     // In case the label has the same base name as object's
                     // internal name, use it as the label instead.
-                    const char *objName = obj->getNameInDocument();
-                    const char *c = &objName[lastpos+1];
+                    const char *c = &objName.c_str()[lastpos+1];
```
wwmayer, line 1368: Weird construct. Why not writing `const char c = objName[lastpos + 1];`? The code below needs to be adjusted to avoid this ugly pointer arithmetic.

src/Gui/CommandFeat.cpp
```diff
@@ -152,7 +152,7 @@ void StdCmdSendToPythonConsole::activated(int iMsg)
     if (!obj)
         return;
     QString docname = QString::fromLatin1(obj->getDocument()->getName());
-    QString objname = QString::fromLatin1(obj->getNameInDocument());
+    QString objname = QString::fromLatin1(obj->getNameInDocument().c_str());
```
wwmayer, line 155: `QString objname = QString::fromStdString(obj->getNameInDocument());`

src/Gui/DlgObjectSelection.cpp
```diff
@@ -544,7 +544,7 @@ QTreeWidgetItem *DlgObjectSelection::createDepItem(QTreeWidget *parent, App::Doc
         item->setFont(0, font);
     }
     item->setText(1, QString::fromUtf8(obj->getDocument()->getName()));
-    item->setText(2, QString::fromUtf8(obj->getNameInDocument()));
+    item->setText(2, QString::fromUtf8(obj->getNameInDocument().c_str()));
     item->setFlags(Qt::ItemIsSelectable|Qt::ItemIsUserCheckable|Qt::ItemIsEnabled);
```
wwmayer, line 548: `item->setText(2, QString::fromStdString(obj->getNameInDocument()));`

src/Gui/DlgPropertyLink.cpp
```diff
@@ -608,7 +606,7 @@ QTreeWidgetItem *DlgPropertyLink::findItem(
         bool found = false;
         for(int i=0,count=item->childCount();i<count;++i) {
             auto child = item->child(i);
-            if(strcmp(o->getNameInDocument(),
+            if(strcmp(o->getNameInDocument().c_str(),
                         child->data(0, Qt::UserRole).toByteArray().constData())==0)
```
wwmayer, line 610: `if (o->getNameInDocument() == child->data(0, Qt::UserRole).toByteArray().constData())`

src/Gui/DlgPropertyLink.cpp
```diff
@@ -1038,7 +1036,7 @@ void DlgPropertyLink::onItemExpanded(QTreeWidgetItem * item) {
                 itemMap[obj] = newItem;
         }
     } else if(allowSubObject) {
-        auto obj = doc->getObject(objName);
+        auto obj = doc->getObject(std::string(objName));
```
wwmayer, line 1039: Why?

src/Gui/SoFCUnifiedSelection.cpp
```diff
@@ -488,24 +488,24 @@ bool SoFCUnifiedSelection::setHighlight(SoFullPath *path, const SoDetail *det,
 
     bool highlighted = false;
     if(path && path->getLength() &&
-       vpd && vpd->getObject() && vpd->getObject()->getNameInDocument())
+       vpd && vpd->getObject() && vpd->getObject()->isAttachedToDocument())
     {
         const char *docname = vpd->getObject()->getDocument()->getName();
-        const char *objname = vpd->getObject()->getNameInDocument();
+        std::string objname = vpd->getObject()->getNameInDocument();
```
wwmayer, line 494: const reference

src/App/DocumentObject.cpp
andre-caldas: It is better because when people call `getNameInDocument()` they should get a name (as most of FreeCAD code actually expects).

Now, the DAG problem is a different one I am not much aware. Do the "DAG people" want a null pointer? I don't know... it is up for them to decide. If the very specific group of people interested in the DAG key decide to forbid a null pointer, then can make a PR. :-)

src/App/DocumentObjectPyImp.cpp
andre-caldas: This should actually be split... no `internal` is needed.

We should use `isAttachedToDocument()` and if so, simply return `{getNameInDocument()}`.

This is actually how I am doing on the "split PRs":
https://github.com/FreeCAD/FreeCAD/commit/b5160f0cf916c51499a2446a8c6471feb71638bd#diff-14f0ebc3222eb16cf3e7cd368140b2fdd3ae6ad97801c29318cfeb28bc590002R55

PR #11066, opened 2023-10-16
[FEM] Improve some task dialog layouts

adrianinsaval:
reduce horizontal space taken for better UX

mentioned at https://github.com/FreeCAD/FreeCAD/issues/11016#issuecomment-1762926895

Other dialogs may need similar changes, I would like people to point out which ones

All of these dialogs preferred using a ton of horizontal space and left a lot of empty vertical space, this should make them more user friendly

new dialogs:
image

even on a 1360x768 screen the dialogs now only take 1/4 of the view and don't require scrolling for usage (depending on toolbars layout)

### DIFF
--- a/src/Mod/Fem/Gui/BoxWidget.ui
+++ b/src/Mod/Fem/Gui/BoxWidget.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>500</width>
-    <height>137</height>
+    <width>246</width>
+    <height>278</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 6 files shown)


marioalexis84:
Well done, Adrián. Now the space is better used.
The DataAlongLine task dialog need similar changes (there are others too, but, if you want, I can fix them following a similar approach to the one you use here).

One comment:
The line surrounding the group is not well aligned with the title of the group and the values in the boxes are aligned in opposite sides. For example, here is a comparison between the Part Box dialog and the Implicit box in FEM.

It would be good to keep the same style throughout the different workbenches. Maybe later we can make generic task dialogs in the GUI directory and then use it in the workbenches as much as possible.

implicit_box
part_box

adrianinsaval:
Data along line added, I preferred leaving this wider but no so long:
image
feel free to propose a different layout if that doesn't work well for you.


About the tittle of the groupboxes, I actually don't know why those look different from one another, looking at the ui files I couldn't spot anything different on those used by Part

wwmayer:
> feel free to propose a different layout if that doesn't work well for you.

It still looks a bit wide. You can e.g. move the labels and spin boxes to a group box and make of them in a vertical layout
TaskPostDataAlongLine

adrianinsaval:
yeah I thought about it too, but I think in that layout it might become necessary to scroll down to get full access to it, not sure what's preferable here.

@marioalexis84 when using that dialog are you normally manually filling up the point coordinates or is the "select points" buttons used more often? if the spinboxes aren't commonly used it might make sense to move them at the bottom and keep more commonly used controls above.
Also, maybe the "Create Plot" button could be moved out of this dialog and placed where you would normally have the "Ok" button in other task

marioalexis84:
@chennes Do you have any suggestions regarding the difference observed in the lines surrounding the group box?

adrianinsaval:
it might have something to do with margins

PR #6753, opened 2022-04-14
Remove use of deprecated distutils package

marioalexis84:
Thank you for creating a pull request to contribute to FreeCAD! To ease integration, we ask you to conform to the following items. Pull requests which don't satisfy all the items below might be rejected. If you are in doubt with any of the items below, don't hesitate to ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10)!

- [ ]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [ ]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [ ]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [ ]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [ ]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [ ]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the issue ID number from our [Issues database](https://github.com/FreeCAD/FreeCAD/issues) in case a particular commit solves or is related to an existing issue. Ex: `Draft: fix typos - fixes #4805`

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.20 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=56135).

---

PR #11068, opened 2023-10-17
Migrate away from deprecated distutils python package where possible

adrianinsaval:
supersedes #6753

debian (and therefore ubuntu) includes a patch for sysconfig on python 3.10 and above:
https://salsa.debian.org/cpython-team/python3/-/blob/python3.10/debian/patches/series?ref_type=heads#L8
https://salsa.debian.org/cpython-team/python3/-/blob/python3.10/debian/patches/sysconfig-debian-schemes.diff

debian packagers should set the env variable `DEB_PYTHON_INSTALL_LAYOUT=deb_system` to install to `/usr/lib/python3/dist-packages`

### DIFF
--- a/src/Mod/Path/libarea/CMakeLists.txt
+++ b/src/Mod/Path/libarea/CMakeLists.txt
@@ -170,11 +170,18 @@ SET_BIN_DIR(area area /Mod/Path)
 SET_PYTHON_PREFIX_SUFFIX(area)
 
 # this figures out where to install the Python modules
-execute_process(
-    COMMAND python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"
-    OUTPUT_VARIABLE Python_site_packages
-    OUTPUT_STRIP_TRAILING_WHITESPACE
-)
+if (${PYTHON_VERSION_STRING} VERSION_LESS "3.10")
+    # deprecated distutils package still required for ubuntu 20.04 and similar
+    # see https://bugs.launchpad.net/ubuntu/+source/python3.8/+bug/2039511
+    # TODO: remove this once we drop support for 20.04
+    EXECUTE_PROCESS(COMMAND ${PYTHON_EXECUTABLE} -c
+        "from distutils.sysconfig import get_python_lib; print(get_python_lib(plat_specific=True))"
+        OUTPUT_VARIABLE Python_site_packages OUTPUT_STRIP_TRAILING_WHITESPACE )
+else()
+    EXECUTE_PROCESS(COMMAND ${PYTHON_EXECUTABLE} -c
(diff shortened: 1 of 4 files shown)


src/Mod/Path/libarea/CMakeLists.txt
```diff
@@ -170,11 +170,18 @@ SET_BIN_DIR(area area /Mod/Path)
...
+if (${PYTHON_VERSION_STRING} VERSION_LESS "3.10")
+    # deprecated distutils package still required for ubuntu 20.04 and similar
+    # see https://bugs.launchpad.net/ubuntu/+source/python3.8/+bug/2039511
+    # TODO: remove this once we drop support for 20.04
+    EXECUTE_PROCESS(COMMAND ${PYTHON_EXECUTABLE} -c
+        "from distutils.sysconfig import get_python_lib; print(get_python_lib(plat_specific=True))"
+        OUTPUT_VARIABLE Python_site_packages OUTPUT_STRIP_TRAILING_WHITESPACE )
+else()
+    EXECUTE_PROCESS(COMMAND ${PYTHON_EXECUTABLE} -c
+        "from sysconfig import get_path; print(get_path('platlib'))"
```
wwmayer, line 182: Is there a reason that here "platlib" is used while in [src/Ext/freecad/CMakeLists.txt](https://github.com/FreeCAD/FreeCAD/pull/11068/files#diff-69aad0a24c6ff11e54441932d5545742f4ed3bdc26a10486c8bf983fb0317b60) it's "purelib"?

adrianinsaval: I could be wrong but libarea seems to include compiled code so I thought it should go to platlib. 
src/Ext/freecad only has python files so I think it's supposed go to purelib, in practice it does depend on compiled code though so perhaps it should go to platlib anyways?

wwmayer: I don't know but for me both of then return the same path and that's why I have asked.

adrianinsaval: yeah normally they give the same result but theoretically they could give different results, one is supposed to be for architecture dependent files and the other for architecture independent ones. I think I read somewhere that it gives different results on fedora or redhat

wwmayer: OK. So, ready for merge?

adrianinsaval: yes, just tried on fedora:
```python
>>> from sysconfig import get_path; print(get_path('platlib'))
/usr/local/lib64/python3.11/site-packages
>>> from sysconfig import get_path; print(get_path('purelib'))
/usr/local/lib/python3.11/site-packages

```

PR #11070, opened 2023-10-17
EditableDatumLabel add support for angle labels.

PaddleStroke:
Following discussion on https://github.com/FreeCAD/FreeCAD/pull/10892
@abdullahtahiriyo cherry pick changes from editable datum label.

### DIFF
--- a/src/Gui/EditableDatumLabel.cpp
+++ b/src/Gui/EditableDatumLabel.cpp
@@ -167,6 +167,12 @@ void EditableDatumLabel::stopEdit()
     }
 }
 
+bool EditableDatumLabel::isInEdit()
+{
+    return spinBox;
+}
+
+
 double EditableDatumLabel::getValue()
 {
     return spinBox->rawValue();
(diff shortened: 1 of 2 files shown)


src/Gui/EditableDatumLabel.cpp
```diff
@@ -171,15 +172,15 @@ double EditableDatumLabel::getValue()
     return spinBox->rawValue();
 }
 
-void EditableDatumLabel::setSpinboxValue(double val)
+void EditableDatumLabel::setSpinboxValue(double val, Base::Unit unit)
```
abdullahtahiriyo, line 175: Look at the lint just below.  It does not look like we need a copy of the unit, do we? If do not, consider ``` const Base::Unit & unit```. Sorry, I did not catch that one yesterday.

PaddleStroke:
I have added the & and also a 'isInEdit' function.

abdullahtahiriyo:
@PaddleStroke 

Why did you disregard the const when adding the reference?

As I see, the unit is only used as a parameter to the Quantity constructor, which takes a const &. It is not changed inside the function (and it would be difficult to think of an scenario where it does).

```explicit Quantity(double value, const Unit& unit=Unit()); ```

There is a good reason to declare const what won't change, it is called const correctness and shows the intent. There is, arguably, an additional reason (of lesser importance) related to enabling the compiler to do certain optimisations.

PaddleStroke:
I have not disregarded it, in fact I have misread the snippet you provided sorry. I will update.

abdullahtahiriyo:
No problem. 

I am more interested in that you understand why it should be const. That way you will  tend to apply it in the future.

PR #11075, opened 2023-10-17
Help module: Shorten string width to 80 characters.

hasecilu:
- The string of the some tooltips are very long that use all screen width.
- In other `.ui` files seems to be a max width of about 80 characters.
- This commit shortens the long strings.
- Also fixes some words capitalization noted by david69 on Crowdin

image

Example of other tab in settings with good width:

image

QtDesigner preserves the line breaks.

screenshot1697472738


yorikvanhavre:
Thanks!!

PR #11077, opened 2023-10-17
Gui: Target 3DViewer directly for overlay wheel events

kadet1090:
I've no idea why, but for some reason if synthesized event is passed into `QGraphicsView` it is not processed. It however works ok if we pass `graphicsView->viewport()` as receiver of the event. However as `View3DInventorViewer` also indirectly derives from `QAbstractScrollArea` it is also affected by the `qobject_cast<QAbstractScrollArea*>(parent)`, and in that case passing events to viewport does not work. Maybe some Qt experts know why, I'm not one of them.

Changing `qobject_cast` to more specific class seems to work fine for both cases, and does not make the code worse.

@wwmayer if you could take a look please, maybe you will be able to find better solution.

Fixes: #11015

### DIFF
--- a/src/Gui/OverlayManager.cpp
+++ b/src/Gui/OverlayManager.cpp
@@ -1948,9 +1948,14 @@ void OverlayManager::Private::interceptEvent(QWidget *widget, QEvent *ev)
 #endif
         lastIntercept = getChildAt(widget, globalPos);
 
-        for (auto parent = lastIntercept->parentWidget(); parent; parent = parent->parentWidget()) {
-            if (qobject_cast<QGraphicsView*>(parent)) {
+        // For some reason in case of 3D View we have to target it directly instead of targeting
+        // the viewport of QAbstractScrollArea like it works for all other widgets of that kind.
+        // That's why for this event we have to traverse up the widget tree to find if it is part
+        // of the 3D view.
+        for (auto parent = lastIntercept; parent; parent = parent->parentWidget()) {
+            if (qobject_cast<View3DInventorViewer*>(parent)) {
                 lastIntercept = parent;
+                break;
             }
         }
 


wwmayer:
> if you could take a look please, maybe you will be able to find better solution.

To be honest I don't even understand the reported problem. What do I have to do to get it?

kadet1090:
You need to enable the overlay UI for the tree view, create techdraw page (can be empty), and try to use mouse wheel on transparent treeview background. In 3D view the scroll event is forwarded correctly and 3D view reacts to the mouse wheel, whereas in TechDraw it does not. This is due to some weird quirk in Qt with QAbstractScrollArea which does some internal processing to `QWheelEvent`s which works fine in case of `View3DInventorViewer` and does not for `QGVPage` or other widgets of `QGraphicsView` kind.

However if you don't think that it is worth to spend much time on that matter, if you don't have any idea how to do that better without further analyzing I think that we should simply merge that.

wwmayer:
Got it, thanks! I have tested the PR and at least it doesn't seem to cause a regression but improves a few things. 
Wheel events now are passed through the transparent tree or property view.

Now in FreeCAD there are few more MDI view types:
* Editor view: was working before and now still works
* Split view: was working before and now still works
* Text document: was working before and now still works
* Spreadsheet: was working before and now still works
* TechDraw: was not working before and works now
* Graphviz view: was not working before and partially works now. There are update problems when scrolling on transparent property view
* Browser view: was not working before and partially works now. Sometimes scrolling works as expected, sometimes you have to scroll several times until it responds and sometimes it doesn't work at all
* Image view: was not working before and still doesn't work

kadet1090:
How can I get into Image View? I should be able to fix it, probably it'd require simply another case in this if. GraphViz has problems with overlay in multiple ways, the same probably is true for Browser - I don't want to tackle this problems in that PR.

wwmayer:
> How can I get into Image View?

Tools > Load image. Afterwards you have to use Windows > Arrange or Cascade to make it not maximized any more.

kadet1090:
> Tools > Load image. Afterwards you have to use Windows > Arrange or Cascade to make it not maximized any more.

It behaves in the same way as GraphViz and Browser. It seems that for this views it is like half enabled. I don't think that there is anything here to do in that PR.

Can we merge it in that state?

wwmayer:
Sure.

luzpaz:
Thanks @kadet1090 !

PR #11078, opened 2023-10-17
fix #9347 - Improve PD Sprocket dialog wording

### LINKED ISSUE #9347: [Renaming] Part Design → Sprocket... → Sprocket preferences window
изображение

On screen Chainwhell (or sprocket) ISO 606 20B.

"Thickness:" is not true name parameter of sprocket. The true name "Tooth width simplex:" or "Tooth width:" 

For example information from manufature : https://www.iwis.com/en-gb/products-services/sprocket-20b-1-25-teeth~p5936

Also "Roller Diameter:" may be renamed to "Diameter of chain roller:"
Rollers - is a rotating parts of a cylindrical shape, inside a chain.

"Chain pitch:" second word may be writed from small letter.

"Sprocket reference:" second word also may be writed from small letter. And colons may have been skipped in end of this string.

adrianinsaval:


src/Mod/PartDesign/SprocketFeature.ui
```diff
@@ -245,14 +263,14 @@
    <item row="3" column="0">
     <widget class="QLabel" name="label_6">
      <property name="text">
-      <string>Roller Diameter:</string>
+      <string>Chain Roller Diam.:</string>
```
wwmayer, line 266: I don't think we should use abbreviations.

PR #11089, opened 2023-10-18
PD: wrong direction determined for linear pattern

wwmayer:
A selected edge of a sketch is not in local coordinates and thus its corresponding axis mustn't be multiplied with the sketch placement as this will lead to a wrong direction vector.

Forum: https://forum.freecad.org/viewtopic.php?t=82022


adrianinsaval:
backported

Issue #11091, opened 2023-10-18
[Problem] PVS-Studio warnings


AndreyKarpov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Hello. I found a few errors in the code. I will be glad if the project becomes a little better.
[FreeCAD and undefined behavior in C++ code: meditation for developers](https://pvs-studio.com/en/blog/posts/cpp/1072/).

### Full version info

```shell
I do not remember. I don't think it matters :)
```

### Subproject(s) affected?

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

wwmayer:
* https://github.com/FreeCAD/FreeCAD/pull/11093
* https://github.com/FreeCAD/FreeCAD/pull/11094
* https://github.com/FreeCAD/FreeCAD/pull/11095
* https://github.com/FreeCAD/FreeCAD/pull/11096
* https://github.com/FreeCAD/FreeCAD/pull/11097
* https://github.com/FreeCAD/FreeCAD/pull/11099
* https://github.com/FreeCAD/FreeCAD/pull/11100
* https://github.com/FreeCAD/FreeCAD/pull/11101

wwmayer:
> inline uint32 readUint32 ( istream &is )
This is how the new zipios library handles it: https://github.com/Zipios/Zipios/blob/8ff715264283f489e9d0d50d9062b7de5a2114d5/src/zipios_common.cpp#L74

WandererFan:
1) QGVPage* QGIView::getQGVPage(TechDraw::DrawView* dView)
    - obsolete code deleted by PR #11103

4) bool GeometryMatcher::compareBSplines(TopoDS_Edge &edge1, TopoDS_Edge &edge2)
    - fixed by Werner PR #11095

5) void QGIViewPart::highlightMoved(QGIHighlight* highlight, QPointF newPos)
    - fixed by Werner PR #11096

12) int DrawUtil::getIndexFromName(const std::string& geomName)
    - missing '=' in DrawViewPart::getVertex() fixed by PR #11103

wwmayer:
Remaining issues:
3. `void Document::handleChildren3D(ViewProvider* viewProvider, bool deleting)`
7. `void SectionCut::startCutting(bool isInitial)` / class must be refactored anyway
13. `inline uint32 readUint32 ( istream &is )`

luzpaz:
[...]

PR #11093, opened 2023-10-18
TD: fix de-referencing null pointer

wwmayer:


FlachyJoe:
I can't see when this function is called, it only appears on MDIViewPage::zoomIn and MDIViewPage::zoomOut but the breakpoints I put here were never reached.

wwmayer:
The zoomIn/Out methods use the method getQGVPage() of the view provider. So, I can't see that's used anywhere.
Anyway, this PR clearly fixes a bug and it's up to @WandererFan to decide to remove this method from QGIView or not.

wwmayer:
#11103 removes the mentioned function

PR #11100, opened 2023-10-19
Mesh: disable potentially unsafe code in Wm4 library

wwmayer:


adrianinsaval:
backported

PR #11101, opened 2023-10-19
TD: fix incorrect array index check in DrawViewPart::getVertex()

wwmayer:


wwmayer:
Also fixed by #11103

Issue #11107, opened 2023-10-20
[Problem] Profile open edges doesn't respect selected cutting direction


jffmichi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Profile mixes climb and conventional milling when processing open edges. The edges are sorted so that the travel moves are minimized without respecting the selected cutting direction. A pull request that fixes this issue is on the way.
![profile_open_edge_direction](https://github.com/FreeCAD/FreeCAD/assets/14004750/6e0c7f28-10db-45df-95ec-2ad0f8c11448)
[profile_open_edge_direction.zip](https://github.com/FreeCAD/FreeCAD/files/13048624/profile_open_edge_direction.zip)

### Full version info

```shell
[code]
OS: Arch Linux (KDE/plasmawayland)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34715 (Git)
Build type: None
Branch: main
Hash: 9defdaf0a66bed1fdf9ca08b93080d052bfa05b8
Python 3.11.5, Qt 5.15.11, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * better-tool-library 0.9.9
  * dodo 1.0.0
  * sheetmetal 0.3.1
  * DynamicData 2.53.0
  * fcgear 1.0.0
  * fasteners 0.4.66
  * fcgear.backup1663629032.4145591 1.0.0 (Disabled)
  * Assembly4 0.50.4
  * lattice2 1.0.0
[/code]
```

### Subproject(s) affected?

Path

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
@LarryWoestman do you mind responding to this Path ticket? TIA!

LarryWoestman:
I'm sorry, but I can't figure out if the original code is correct or what it is doing in many areas, so I have no confidence that the new code is an appropriate change.

There is a lot of background information about the internal data structures that I don't know and can't figure out from the available documentation.

PR #11108, opened 2023-10-20
Path: fix Profile open edges not respecting selected cutting direction

jffmichi:
This PR fixes issue #11107

_buildProfileOpenEdges uses Path.fromShapes to generate G-Code from the wire. However Path.fromShapes reverses the wire if that shortens the travel move to the start point. By always manually setting the first or last vertex of the wire (depending on obj.Direction) instead of using self.endVector the edge is allways milled in the correct direction.

### DIFF
--- a/src/Mod/Path/Path/Op/Area.py
+++ b/src/Mod/Path/Path/Op/Area.py
@@ -346,21 +346,25 @@ class ObjectOp(PathOp.ObjectOp):
                 # Note that emitting preambles between moves breaks some dressups and prevents path optimization on some controllers
                 pathParams["preamble"] = False
 
-                if self.endVector is None:
-                    verts = hWire.Wires[0].Vertexes
-                    idx = 0
-                    if obj.Direction == "CCW":
-                        idx = len(verts) - 1
-                    x = verts[idx].X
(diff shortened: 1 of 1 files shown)


src/Mod/Path/Path/Op/Area.py
```diff
@@ -346,21 +346,25 @@ def _buildProfileOpenEdges(self, obj, edgeList, isHole, start, getsim):
...
-                        y = 0.00001
-                    pathParams["start"] = FreeCAD.Vector(x, y, verts[0].Z)
-                else:
-                    pathParams["start"] = self.endVector
+                # Always manually setting pathParams["start"] to the first or
+                # last vertex of the wire (depending on obj.Direction) ensures
+                # the edge is always milled in the correct direction. Using
+                # self.endVector would allow Path.fromShapes to reverse the
+                # direction if that would shorten the travel move and thus cause
+                # the edges being milled in seemingly random directions.
```
luzpaz, line 354: Please fix trailing whitespace

jffmichi: done :+1:

PR #11110, opened 2023-10-20
Fem: Setup manipulator plane

marioalexis84:
Currently, in the jack manipulator used by the plane function an additional rotation is added to align the axis translator (which can be changed by pressing the CTRL key) with the Z-axis, since by default the translator is aligned with the Y-axis.
This additional rotation also changes the blue `postPlane` rectangle used to represent the real plane, leaving the blue rectangle and the real plane perpendicular.
The first commit fix removes the extra rotation and changes the direction of the axis translator in the manipulator settings.
Now the blue plane match with real plane.

The second commit set auto-scaling to `false` and set initial scaling in the command call based on the pipeline size.
Currently, the plan is scaled using not only the pipeline size but also all objects in the 3D view.
Also removes some unnecessary code used when the value of the object's properties change for the first time.

### DIFF
--- a/src/Mod/Fem/Gui/Command.cpp
+++ b/src/Mod/Fem/Gui/Command.cpp
@@ -2292,6 +2292,10 @@ void CmdFemPostFunctions::activated(int iMsg)
                       center[0],
                       center[1],
                       center[2]);
+            doCommand(Doc,
+                      "Gui.ActiveDocument.%s.Scale = %f",
+                      FeatName.c_str(),
+                      box.GetDiagonalLength());
         }
         else if (iMsg == 1) {  // Sphere
             doCommand(Doc,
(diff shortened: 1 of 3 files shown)


luzpaz:
CC @FEA-eng

marioalexis84:
So the problem occurs in those GuiDocument.xml files that do not have the Scale property in the Plane view provider.
Is there a way to check for its absence when trying to read that property from the .xml file and apply the current solution only to these objects?

wwmayer:
There is an indirect way to check for the existence of that property. The view provider overrides the Restore() method where at the beginning it sets the Scale property to an invalid value like -1 or maybe NaN. Then it calls the Restore() method of the base class and afterwards it checks the value of the Scale property. If it's still -1 (or NaN) then we can assume the xml file lacks of that property.

maxwxyz:
@marioalexis84 will you pick up this PR before the 1.0 feature freeze on **Monday, 2024-06-03** ?

maxwxyz:
What is left to do here to be ready for review?

PR #11121, opened 2023-10-20
Gui: Make "Dock window overlay" GUI texts more consistent

Roy-043:


src/Gui/CommandView.cpp
```diff
@@ -3925,8 +3925,8 @@ StdCmdDockOverlayMouseTransparent::StdCmdDockOverlayMouseTransparent()
   :Command("Std_DockOverlayMouseTransparent")
 {
   sGroup        = "View";
-  sMenuText     = QT_TR_NOOP("Bypass mouse events in dock overlay");
-  sToolTipText  = QT_TR_NOOP("Bypass all mouse events in dock overlay");
+  sMenuText     = QT_TR_NOOP("Bypass mouse events in docked overlays");
```
luzpaz, line 3928: plural 'overlays' ? How do you figure ?

adrianinsaval:
nitpicking but I don't think 'window' is the right word for dock widgets, perhaps use 'panel' instead?

Roy-043:
In the original version the words "window", "widget" and "panel" are used. AFAICT the first two words were used interchangeably, I have therefore replaced "widget" with "window". It would surpise me if all 3 words have the same meaning in this context.

wwmayer:
In Qt the class is called QDockWidget but in its description it says [dock windows](https://doc.qt.io/qt-6/qdockwidget.html#details). So, "windows" doesn't look wrong to me.

realthunder:
I am fine changing the terminology. It is end user facing after all. Technically, my understanding is that Qt uses `Window` to refer to operating system backed user interface element, while `Widget` is what Qt implemented, sort of drawn on `Window`, which can host one or more `widgets`.

adrianinsaval:
I'm not opposed to either term, and I'm not even native speaker so I could just be plain wrong. But to me panel communicates better what this does.

PR #11123, opened 2023-10-20
[TechDraw] - Cosmetic Circle fixes

pavltom:
Hello,

with this PR I would like to provide 3 fixes for Cosmetic Circle, or Cosmetic Arc of Circle resp.
1) Correct task dialog title: This is trivial.
2) Problems with changed center coordinates after save/reload: The dialog takes (and stores) center coordinates from AoC permaStart, however when restored, permaStart is not filled from AoC center, but from startPnt. This is here fixed in Cosmetic.cpp:302-303
3) Problems with "reversing" AoC after save/reload: If a cosmetic AoC is created (e.g. from angle 0 to 270), it correctly shows the arc, however when saved and restored, the arc is now displayed like 90 to 360. The reason is the underlying OCC edge must be flipped along the ZX plane, because Qt uses screen coordinates system ([0, 0] in top left corner). When such "inverted" AoC edge is saved and later restored, the OCC edge is constructed incorrectly, with reversed circulation. The simplest and most elegant solution is to save inverted geometry and after recreating the OCC edge invert it for the Qt once again. This is fixed in Cosmetic.cpp lines 258 and 301.

I hope you will find this fix useful, in case of any issues, please let me know.

### DIFF
--- a/src/Mod/TechDraw/App/Cosmetic.cpp
+++ b/src/Mod/TechDraw/App/Cosmetic.cpp
@@ -255,7 +255,7 @@ void CosmeticEdge::Save(Base::Writer &writer) const
         circ->Save(writer);
     } else if (m_geometry->getGeomType() == TechDraw::GeomType::ARCOFCIRCLE) {
         TechDraw::AOCPtr aoc = std::static_pointer_cast<TechDraw::AOC>(m_geometry);
-        aoc->Save(writer);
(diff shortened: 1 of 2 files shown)


WandererFan:
Thank you.  Nice to see you back!

pavltom:
@WandererFan thank you for superfast merge! With coming winter I hope I find some time for contributions to the code :-)

PR #11127, opened 2023-10-20
[Core] EditableDatumLabel : Sketcher Tool settings card 4

PaddleStroke:
@abdullahtahiriyo PR of 
https://github.com/FreeCAD/FreeCAD/commit/8128e3eb9d993d1ba761f07c4c2dfdd810c16a65
https://github.com/FreeCAD/FreeCAD/commit/1727c153876e779ab72e5fcf340a3c28b79fd317
https://github.com/FreeCAD/FreeCAD/commit/5a976ade2a24a2e6f996fede4c28fbeef532ff90

### DIFF
--- a/src/Gui/EditableDatumLabel.cpp
+++ b/src/Gui/EditableDatumLabel.cpp
@@ -51,6 +51,7 @@ EditableDatumLabel::EditableDatumLabel(View3DInventorViewer* view,
     : isSet(false)
     , autoDistance(autoDistance)
     , autoDistanceReverse(false)
+    , value(0.0)
     , viewer(view)
     , spinBox(nullptr)
     , cameraSensor(nullptr)
@@ -160,6 +161,7 @@ void EditableDatumLabel::startEdit(double val, QObject* eventFilteringObj, bool 
     connect(spinBox, qOverload<double>(&QuantitySpinBox::valueChanged),
         this, [this](double value) {
         this->isSet = true;
+        this->value = value;
         Q_EMIT this->valueChanged(value);
     });
 }
@@ -167,6 +169,15 @@ void EditableDatumLabel::startEdit(double val, QObject* eventFilteringObj, bool 
 void EditableDatumLabel::stopEdit()
 {
     if (spinBox) {
+        // write the spinbox value in the label.
+        Base::Quantity quantity = spinBox->value();
+
+        double factor{};
+        QString unitStr;
+        QString valueStr;
+        valueStr = quantity.getUserString(factor, unitStr);
+        label->string = SbString(valueStr.toUtf8().constData());
+
         spinBox->deleteLater();
         spinBox = nullptr;
     }
@@ -180,7 +191,8 @@ bool EditableDatumLabel::isInEdit()
 
 double EditableDatumLabel::getValue()
 {
-    return spinBox->rawValue();
+    // We use value rather than spinBox->rawValue() in case edit stopped.
+    return value;
 }
 
 void EditableDatumLabel::setSpinboxValue(double val, const Base::Unit& unit)
@@ -192,6 +204,7 @@ void EditableDatumLabel::setSpinboxValue(double val, const Base::Unit& unit)
 
     QSignalBlocker block(spinBox);
     spinBox->setValue(Base::Quantity(val, unit));
+    value = val;
     positionSpinbox();
 
     if (spinBox->hasFocus()) {
--- a/src/Gui/EditableDatumLabel.cpp
+++ b/src/Gui/EditableDatumLabel.cpp
@@ -95,7 +95,7 @@ EditableDatumLabel::~EditableDatumLabel()
 
 void EditableDatumLabel::activate()
 {
-    if (!viewer) {
+    if (!viewer || isActive()) {
         return;
     }
 
@@ -133,6 +133,10 @@ void EditableDatumLabel::deactivate()
 
 void EditableDatumLabel::startEdit(double val, QObject* eventFilteringObj, bool visibleToMouse)
 {
+    if (isInEdit()) {
+        return;
+    }
+
     QWidget* mdi = viewer->parentWidget();
 
     label->string = " ";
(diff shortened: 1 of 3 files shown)


PaddleStroke:
So I'm having a look at the lint errors and there're many I don't know what to do. For example : 
```
bool EditableDatumLabel::isActive()
{
    return cameraSensor;
}
```
> implicit conversion 'SoNodeSensor *' -> bool [readability-implicit-bool-conversion]

This seems very common way of testing if a pointer exist. Is that not acceptable?

Should I look at lint error that are on my created code, or all that are visible in the files edited?

`delete data;`
> deleting a pointer through a type that is not marked 'gsl::owner<>'; consider using a smart pointer instead 

`    auto info = new NodeData{ this };`
> initializing non-owner 'NodeData *' with a newly created 'gsl::owner<>' [cppcoreguidelines-owning-memory]
>     cameraSensor = new SoNodeSensor([](void* data, SoSensor* sensor) {
>  Check warning on line 106 in src/Gui/EditableDatumLabel.cpp

> assigning newly created 'gsl::owner<>' to non-owner 'SoNodeSensor *' [cppcoreguidelines-owning-memory]

PaddleStroke:
> / Lint / Lint
> method 'setLabelStartAngle' can be made const [readability-make-member-function-const]

```
void EditableDatumLabel::setLabelStartAngle(double val)
{
    label->param2 = float(val);
}
```
Why is lint telling me that this function can be made const? From my understanding making a function const means it is read only. Here it very clearly change the label object which is an object of EditableDatumLabel. Am I missing something?

PaddleStroke:
Ah is that why 
// NOLINTNEXTLINE
has been added before 
`void EditableDatumLabel::setLabelType(SoDatumLabel::Type type)`?

abdullahtahiriyo:
@PaddleStroke 

I think your questions are very pertinent. I do not have clear answers beyond what I write below, and that is why I am asking Werner.

The approach is that we should not introduce new lint errors/warnings with new code. We are not obliged to correct already present ones. However, this file is relatively new, and I think we should at least find a common understanding about how we should handle this.

@wwmayer 

I think we should have a common approach to general lint warnings (perhaps we do and I have missed it).

What do you think we should do about this below? 

@PaddleStroke 

The ones of:
```initializing non-owner 'NodeData *' with a newly created 'gsl::owner<>' [cppcoreguidelines-owning-memory]```
```assigning newly created 'gsl::owner<>' to non-owner 'SoNodeSensor *' [cppcoreguidelines-owning-memory]```
```deleting a pointer through a type that is not marked 'gsl::owner<>'; consider using a smart pointer instead [cppcoreguidelines-owning-memory]``

means that there is no clear owner. The standard solution is to create them into a smart pointer.

This node requires very specific actions to do for destruction, which appears to require a custom deleter for the smart pointer.  Because for a unique pointer, the custom deleter is part of the type and it is a data member, it forces to define the deleter in the header file.

Then, coin3d has a very specific memory model.

I doubt this is the only case we have in FreeCAD, and would require an agreed solution. So:

@wwmayer 

What do you think we should do about this?

@PaddleStroke 

For the case of the spinbox, we have also a similar situation requiring a custom deleter. I handled a similar situation here:

https://github.com/FreeCAD/FreeCAD/blob/e3ad4d35dcb7f9cb3ba747b92cb33456aec48634/src/Gui/NotificationBox.cpp#L46

As in the previous case, I think it requires an agreed way forward. It could be to move the deleted struct and template to a separate header and including this separate header whereever we need this solution.

Regarding:
``` implicit conversion 'SoNodeSensor *' -> bool [readability-implicit-bool-conversion]``` 

and the other similar for the spinbox, the recommendation below is to use != nullptr, see:
https://clang.llvm.org/extra/clang-tidy/checks/readability/implicit-bool-conversion.html

I acknowledge ```if(pointer)``` is quite used. At the same time, the lower amount of lints we have the better. Alternatively we could silence it (// NOLINTBEGIN // NOLINTEND? or //NOLINT). I do not know if we have a common approach here.

Regarding:
```1st argument 'width' (passed to 'fHeight') looks like it might be swapped with the 2nd, 'length' (passed to 'fWidth') [readability-suspicious-call-argument]```

I do not think there is anything you can do here, as the argument types are the ones given by the viewer. We should however have a common understanding how to address this kind of warning. // NOLINTBEGIN // NOLINTEND?

Regarding:
```'moc_EditableDatumLabel.cpp' file not found [clang-diagnostic-error]```
```suspicious #include of file with '.cpp' extension [bugprone-suspicious-include]```
```'ui_TaskImage.h' file not found [clang-diagnostic-error]```

My understanding is that they are false positives. I do not know if we should silence them or ignore them. We need a common approach. // NOLINTBEGIN // NOLINTEND?

**Comment by wwmayer** [comment]
> This seems like a very common way of testing if a pointer exist. Is that not acceptable?

Yes, it's common and recommended by the C++ core guidelines: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-if
It should be checked if we can switch off  `readability-implicit-bool-conversion` for pointers. However, this warning makes sense for int, long, ... 

> Should I look at lint error that are on my created code, or all that are visible in the files edited?

Concentrate on your own code as otherwise reviewing a PR may become more difficult. Lint warnings of existing code should be fixed in separate PRs.

> cppcoreguidelines-owning-memory

As long as we don't use [gsl](https://github.com/microsoft/GSL) this warning should be suppressed.

> means that there is no clear owner. The standard solution is to create them into a smart pointer.

As far as I understand gsl::owner it's still a raw pointer but it's more like an annotation to make it more explicit that it's a raw pointer.

File: src/Gui/EditableDatumLabel.cpp [hunk 1]
```diff
@@ -156,27 +165,43 @@ void EditableDatumLabel::startEdit(double val, QObject* eventFilteringObj)
...
+
         spinBox->deleteLater();
         spinBox = nullptr;
     }
 }
 
+bool EditableDatumLabel::isActive()
+{
+    return cameraSensor;
+}
```
abdullahtahiriyo, line 193: So, following the guideline, please suppress this one.

abdullahtahiriyo:
For the rest, I am ok.

wwmayer:
In our clang-tidy file we have these two options:
```
  - key:             readability-implicit-bool-conversion.AllowIntegerConditions
    value:           'false'
  - key:             readability-implicit-bool-conversion.AllowPointerConditions
    value:           'true'
```

The latter is supposed to allow to write `if (pointer)` but this apparently doesn't work for `bool isSomething() const { return pointer; }`. In such cases we could (or should?) write `bool isSomething() const { return pointer != nullptr; }`

Btw, you should also fix const correctness of isActive and isInEdit: 
`bool EditableDatumLabel::isActive() const` and `bool EditableDatumLabel::isInEdit() const`

PaddleStroke:
It's done.

PR #11129, opened 2023-10-20
Part: Filter selection GUI texts to sentence case

Roy-043:


src/Mod/Part/Gui/CommandFilter.cpp
```diff
@@ -170,10 +170,10 @@ PartCmdVertexSelection::PartCmdVertexSelection()
   : Command("Part_VertexSelection")
 {
     sGroup        = "Standard-View";
-    sMenuText     = QT_TR_NOOP("Vertex Selection");
-    sToolTipText  = QT_TR_NOOP("Select a Vertex/Vertices");
+    sMenuText     = QT_TR_NOOP("Vertex selection");
+    sToolTipText  = QT_TR_NOOP("Select a vertex/vertices");
```
wwmayer, line 174: Should we drop the singular form? I think if it says "Select vertices" then it should be clear that one can also select a single vertex only.

Roy-043:
Maybe go for a longer sentence:
"Only allow the selection of vertices"
(and change the other tooltips accordingly)

wwmayer:
Would be good for me too

chennes (approved):
These changes look good to me.

wwmayer:
For me too.

PR #11139, opened 2023-10-22
Avoids using getNameInDocument() to test if DocumentObject is attached to a Document.

andre-caldas:
This patch substitutes by isAttachedToDocument() (almost) everywhere where getNameInDocument() is used for this purpose.

The very few places not touched by this patch demand a (just a little) less trivial change. When we change the returning type of getNameInDocument() to std::string, those places will be easily found, because they shall generate a compiler error (converting std::string to bool).

Rationale:
The fact that getNameInDocument() return nullptr to indicate that the object is not attached to a document is responsible for lots of bugs where the developer does not check for "nullptr".

The idea is to eliminate all those uses of getNameInDocument() and, in the near future, make getNameInDocument() return always a valid std::string.

This is the first PR on a series to satisfy:
https://github.com/FreeCAD/FreeCAD/pull/11053#issuecomment-1764862715

### DIFF
--- a/src/App/Application.cpp
+++ b/src/App/Application.cpp
@@ -1215,7 +1215,7 @@ std::set<DocumentObject *> Application::getLinksTo(
     } else {
         std::set<Document*> docs;
         for(auto o : obj->getInList()) {
-            if(o && o->getNameInDocument() && docs.insert(o->getDocument()).second) {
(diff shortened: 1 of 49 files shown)


sliptonic:
@wwmayer  Any feedback on this one?

PR #11141, opened 2023-10-22
Does not rely on the pointervalue returned by getNameInDocument() to use as a DAG key.

andre-caldas:
I am splitting the PR https://github.com/FreeCAD/FreeCAD/pull/11053 into a sequence of simpler PRs. This is the second in this sequence. The first one is: https://github.com/FreeCAD/FreeCAD/pull/11139.

Basically, the `DocumentObject::getNameInDocument()` method has three different uses:
1. Obtains the name of this object within the `Document`. (legitimate use)
2. Indicates if the object is attached to a Document. Returns `nullptr` in case not. See: https://github.com/FreeCAD/FreeCAD/pull/11139.
3. Provides a "DAG key" to refer to this object. (this patch)

This PR creates the method `getDagKey()` that shall be used in the context of item 3, instead of relaying on `getNameInDocument()`.

The main idea is to, in the near future, change `getNameInDocument()` and make it never return an invalid string. There are many places in FreeCAD where the developer assumes `nullptr` will never be returned. Changing this `getNameInDocument()` behaviour will fix lots of (potential) bugs, like this: https://github.com/FreeCAD/FreeCAD/issues/10130.

### DIFF
--- a/src/App/DocumentObject.cpp
+++ b/src/App/DocumentObject.cpp
@@ -292,6 +292,15 @@ std::string DocumentObject::getFullLabel() const {
     return name;
 }
 
+const char* DocumentObject::getDagKey() const
+{
+    if(!pcNameInDocument)
+    {
+        return nullptr;
+    }
+    return pcNameInDocument->c_str();
+}
+
 const char *DocumentObject::getNameInDocument() const
 {
     // Note: It can happen that we query the internal name of an object even if it is not
@@ -808,7 +817,7 @@ DocumentObject *DocumentObject::getSubObject(const char *subname,
         if(outList.size()!=_outListMap.size()) {
             _outListMap.clear();
             for(auto obj : outList)
-                _outListMap[obj->getNameInDocument()] = obj;
+                _outListMap[obj->getDagKey()] = obj;
         }
         auto it = _outListMap.find(name.c_str());
         if(it != _outListMap.end())
--- a/src/App/DocumentObject.h
+++ b/src/App/DocumentObject.h
@@ -132,6 +132,8 @@ public:
     DocumentObject();
     ~DocumentObject() override;
 
+    /// returns a value that uniquely identifies this DocumentObject.
+    const char* getDagKey() const;
     /// returns the name which is set in the document for this object (not the name property!)
     const char *getNameInDocument() const;
     /// Return the object ID that is unique within its owner document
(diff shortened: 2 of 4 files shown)


andre-caldas:
Dear @wwmayer,

I didn't realize you were requested to review https://github.com/FreeCAD/FreeCAD/pull/11053 while me and @WandererFan were still discussing it. I couldn't imagine that you would review it while we were discussing splitting it into smaller PRs. It seems I was supposed to mark it "draft"... now I know. :-)

From your comments on the other PR, I imply you'd rather have `getDagKey()` renamed to `getDAGKey()`. I can do that if you wish.

I find it really important that `getNameInDocument()` do not return a `nullptr` that is not being checked in the vast majority cases in FreeCAD. I also find it important to not have `getNameInDocument()` being used with three different purposes. This PR is just a small step into this direction.

I hope you can find some time to review this. :-)

WandererFan:
The name getDagKey is not very informative.  Is this really getting the document object's unique name?

andre-caldas:
> The name getDagKey is not very informative. Is this really getting the document object's unique name?

I agree, a lot!
Also, @wwmayer has expressed some dislike of the fact that DAG is not fully capitalized.

The fact is that the pointer returned by `getNameInDocument()` is being used as a "unique key"  that can be "ordered". When we have to build a dependency three, for example, we might be interested in avoiding cycles. For example, one update might trigger some other update, and if we have cycles, the update process enters an infinite loop.

One solution to this problem is to "order" each node somehow. Then, by avoiding decreasing in order, we can avoid cycles. The "A" is for "acyclic": https://en.wikipedia.org/wiki/Directed_acyclic_graph

This new method has actually nothing to do with the "name in document". But since this "pointer to a std::string" does never change (an implementation detail that should not be rellied upon), the (natural number) value returned by `getNameInDocument()` is used for comparison/ordering.

I do believe that this `getDagKey()` should also be deprecated. I guess every `DocumentObject` should have a UUID. And that UUID could probably be used. But this is a different problem that should be addressed in some other (and wider) discussion.

By the way, this patch creates `getDagKey()`, but there are many places where calls to `_getNameInDocument()` still need to be replaced by `getDagKey()`. The only role played by `getDagKey()` is to get rid of all those calls to `getNameInDocument()` that expect a pointer and not a string.

This way, it will be trivially safe to call `getNameInDocument()` and not check for `nullptr`, for example.

sliptonic:
Bump.   Does anyone have more to add?

chennes:
We discussed this in the meeting this morning and decided to press forward with it. Please ping me if something goes wrong related to it.

Issue #11113, opened 2023-10-20
[Problem] Adjust Default Main Light Position


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Either let the user adjust the global position of the main light / illumination of the environment or adjust the default position so that is not directly coming from the direction of the viewing plane / angle.

I use a lot of material and color variations for different bodies and faces and when they are glossy and face directly the view, they all appear white (reflection of the environment lightning). It would be beneficial to rotate the global light source to an angle that when drawing on standard planes, it is possible to distinguish different parts and faces.

When looking at the object you cannot see which faces or edges are selected or hovered/highlighted - simply everything appears white.
Especially when doing sketches I cannot differentiate between different color faces because of this behavior, all just appear white.

As Syres mentioned, it is currently only possible to adjust the position with Python but that has to be done at every restart, here is the code from @wwmayer: https://forum.freecad.org/viewtopic.php?p=146925#p146925

Simple demo:
![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/234a6f71-5c54-431a-b1a9-b97ff876c1b4)

Viewing from top:
![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/a3b0c9b8-5352-4793-9d04-29ba1087ca6a)

### Full version info
[code]
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34667 (Git)
Build type: Release
Branch: main
Hash: https://github.com/FreeCAD/FreeCAD/commit/af5b8714a8371162addb97dd25c29250fcecf7ad
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods:

    Curves 0.6.13
    OpenDark 2023.10.20a
    sheetmetal 0.3.1
    [/code]

### Subproject(s) affected?

Core

### Anything else?

https://forum.freecad.org/viewtopic.php?p=676725
https://forum.freecad.org/viewtopic.php?p=146925

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

maxwxyz:
Here is a demonstration. Simple model but you cannot distinguish the colors/faces or features you want to see, when viewing it directly from the normal direction of the faces. All appear white. 
You can take alook in the file and edit the first sketch: 
[...]

Issue #9544, opened 2023-05-11
[Problem] Zooming in 3D view causes substantial decrease in frame rate


dominikh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: NixOS 23.05 (Stoat)
Word size of FreeCAD: 64-bit
Version: 0.21.0.Unknown
Build type: Release
Python 3.10.11, Qt 5.15.9, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.2
Locale: English/United States (en_US)
Installed mods: 
  * fasteners 0.4.56
  * ProDarkThemePreferencePack 1.0.0
[/code]
```

### Subproject(s) affected?

Core

### Problem description

When zooming in and out, the frame drops from 60 fps (my display's refresh rate) to roughly 30 fps, as determined by the frame counter overlay of FreeCAD. This is independent of the complexity of the scene.

This isn't really noticeable when zooming with a mouse wheel, but it is highly noticeable when using a Space Mouse, where most movement will contain some amount of zoom, and where the unstable frame rate leads to a choppy experience.

### Anything else?

My hypothesis is that the slowdown is caused by updating the dimensions in the bottom right corner of the main window. To test this, I've patched FreeCAD to remove said updates, and this solved the issue.

```
diff --git i/src/Gui/MainWindow.cpp w/src/Gui/MainWindow.cpp
index 1cb9a32475..86c17ef3e7 100644
--- i/src/Gui/MainWindow.cpp
+++ w/src/Gui/MainWindow.cpp
@@ -2165,7 +2165,7 @@ void MainWindow::setPaneText(int i, QString text)
         showStatus(MainWindow::Pane, text);
     }
     else if (i==2) {
-        d->sizeLabel->setText(text);
+        // d->sizeLabel->setText(text);
     }
 }
```

I know that both UI layout and text rendering can be expensive, but missing the 16.6 ms target is still surprising to me.

A similar issue occurs when moving the mouse over objects, which updates the "preselected" information in the bottom left of the main window. This, however, is less of an issue with the Space Mouse, albeit still problematic when using the Space Mouse and the normal mouse simultaneously, e.g. when making edits.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

xtemp09:
Most likely the culprit is copying QString. Perhaps, standard `const QString &` is better.

xtemp09:
[...]

PR #11146, opened 2023-10-22
Gui: fixes issue #11113: Adjust Default Main Light Position

wwmayer:


luzpaz:
Can someone make a clip demonstrating this feature please? :pray:

wwmayer:
directional_light_dragger
Inside the red circle there is a handle that you can rotate. This changes the direction of the light source.

xtemp09:
I'm not an expert in 3D modeling, but it would be better if there were a few light sources.

https://www.youtube.com/watch?v=ElMM3u2MO5k&t=97s

luzpaz:
CC @howetuft (think you would be interested in this feature)  
@MisterMakerNL would  you mind taking a look at the stylesheet issues of this feature once it's merged ? TIA!

wwmayer:
> I'm not an expert in 3D modeling, but it would be better if there were a few light sources.

I don't think Coin3D is capable of what Blender can do. 

I have tested what happens when there are more light sources. When they emit their lights from the sides then your model has almost no dark areas any more but it also increases the risk that it looks completely white.

For now this PR solves the reported issue but it might be an option for the future to extend the lighting.

MisterMakerNL:
Most CAd programs like Solidworks use HDR maps just like easy reflect macro does. https://forum.freecad.org/viewtopic.php?p=708265#p708265
And then you have indeed also the option for a 3-point light.
image
I have a huge folder with HDRi studio maps, I don't know the license tho but we can use them to test.
I can also make HDRI in Blender so we don't get any license issues.

chennes:
I think that this is a clear improvement, so even if there are even more ways of improving, I'm inclined to merge this as-is.

maxwxyz:
Thanks, I will test it when it is in the weekly builds as I cannot build it myself. 

- Is the absolute angle defined (world coordinates) or depending on the camera view (second one preferred)?
- Is the default setting so that it is not parallel / identical to the camera view https://github.com/FreeCAD/FreeCAD/issues/11113

wwmayer:
The lighting is done by SoDirectionalLight that has a SoSFVec3f field. I don't think this direction depends on the camera view direction. The default light direction is (0,0,-1). As it's independent of the camera view direction it's only parallel if the camera looks down/up the Z axis.

maxwxyz:
Then this is not what renders all glossy materials white. If you look at the example of https://github.com/FreeCAD/FreeCAD/issues/11113 from the bottom view (or normal to any face) they will render white. I just coincidentally modeled the demo file so that it appears white from the top (could be from any direction)

wwmayer:
That's why we have the preferences page where you can change the light direction to something different to (0,0,-1). If you then look from top on the model it doesn't look white any more.

Another option is to reduce the intensity from 100% to e.g. 60%.

maxwxyz:
OK I will test it when available. Couldn't we set the default value then to something coming e.g. from the top left as this would not influence anything badly but improve the default behavior that the faces appear white when looking normal to the camera direction (editing sketches, selecting references,...): 
grafik

wwmayer:
If needed we can change the default.

maxwxyz:
great, I would propose (1, -1, -1) - is this feasible?

wwmayer:
> great, I would propose (1, -1, -1) - is this feasible?

This doesn't look good. I would try (0.2, -0.2, -1). As already said you can test this setting with your version. Open the parameter editor and go to BaseApp/Preferences/View. If the key HeadlightDirection doesn't exist then create it as a text object and set its value to `(0.2,-0.2,-1)`

wwmayer:
> `float QColor::redF()`

With Qt5 It returns a qreal which is an alias for double. But with Qt6 it indeed returns a float now.

maxwxyz:
> > great, I would propose (1, -1, -1) - is this feasible?
> 
> This doesn't look good. I would try (0.2, -0.2, -1). 

This looks good!

xtemp09:
Why is `Q_PROPERTY` rare in the source code of _FreeCAD_? Wouldn't it fit perfectly here, especially with such benefits like notification event, reset etc?

Why do the developers deliberately avoid standard usage of _Qt Signals & Slots mechanism_, for example, in case of _setPaneText_ of MainWindow?


@wwmayer, this pull request seems to avoid its standard usage too.

Here, you connect the checkbox

```cpp
connect(ui->checkBoxLight1, &QCheckBox::toggled,
        this, &DlgSettingsLightSources::toggleLight);
```

with the function

```cpp
void DlgSettingsLightSources::toggleLight(bool on)
{
    view->getHeadlight()->on = on;
}
```

while the headlight already has the method toggling light.

https://github.com/FreeCAD/FreeCAD/blob/b70956ba3dd0ffb12bac6761d49dcacb5de8c950/src/Gui/Quarter/QuarterWidget.cpp#L395-L398

Sorry to bombard you with questions, but I hope you will shed headlight when you find time.

wwmayer:
> Wouldn't it fit perfectly here, especially with such benefits like notification event, reset etc?

In how far? The property mechanism allows it to set or get a property by string with `qtclass->setProperty("myProperty", value)` or `QVariant value = qtclass->property("myProperty")`. If myProperty is registered with Q_PROPERTY and  appropriate setter/getter functions they will be called when accessing the property as shown above.

But I don't see how this would fit here because the dialog must react on signals emitted by its child widgets.

> Why do the developers deliberately avoid standard usage of Qt Signals & Slots mechanism,

What is the standard usage of Qt Signals & Slots? Do you mean to write a slot function of the form `on_widgetname_signalname()` so that it will be connected automatically? This way is not recommended any more because it's error-prone and easy to break. A slot function won't be connected if at a later point the name of the widget changes, the signal name changes (e.g. in a newer Qt version) and so on.

And the old-style connection using the SIGNAL/SLOT macros is not recommended either.

In both cases you only may get a warning at runtime and this is tedious to test. And once you made sure it works correctly you can never be sure that it breaks all the sudden in the future.

The recommended and safest way is to use the new style connect with function pointers because there the compiler already tells you if a connect fails. Of course, it's a lot more work but that's the price you have to pay.

> for example, in case of setPaneText of MainWindow?

What exactly is your complaint about setPaneText?

> while the headlight already has the method toggling light.

Ah thanks. I will prepare a PR to fix this and the color issue.

xtemp09:
I should have provided more details about `void MainWindow::setPaneText(int i, QString text)`. My question is related to the issue #9544. Looking into the code I found out that the status messages and dimensions in the _sizeLabel_ are set directly, without involving Qt signal & slot system. 

I have made the commit xtemp09/FreeCAD@5acec63 to use the Qt messaging system to deliver the dimensions from _View3DInventorViewer_ to the _sizeLabel_.

I thought it would be better if the widget uses `Q_EMIT clearStatus()` or `Q_EMIT message(msg)` instead of

<details><summary>this</summary>

https://github.com/FreeCAD/FreeCAD/blob/aaa962944b77a2886bc8efb94c6978fa09af650e/src/Gui/ProgressBar.cpp#L312-L342

</details>

wwmayer:
> I should have provided more details about void MainWindow::setPaneText(int i, QString text)

I think such things should be discussed elsewhere as it has nothing to do with this PR.

> I have made the commit https://github.com/xtemp09/FreeCAD/commit/e2c342bd2cf2b60d898f5c3dce75acc4662fa8dc

Isn't this a direct connection, rather than a queued connection?

> I thought it would be better if the widget uses Q_EMIT clearStatus() or Q_EMIT message(msg) instead of

It could be checked as alternative.

maxwxyz:
@wwmayer Thinking about your implementation of the light setting: would it be easy to implement the option in the preferences to add multiple lights, adjust them each and also remove lights again?

wwmayer:
This could be done in future PR

maxwxyz:
Done: https://github.com/FreeCAD/FreeCAD/issues/11166

maxwxyz:
@wwmayer when I set the HeadlightDirection in the Parameters Editor directly, it is instantly changed in the 3D view. When I then open the preferences and just klick apply, another value is set.
It seems like the HeadlightRotation instead of the HeadlightDirection is used or what is the workaround?

wwmayer:
The light dragger node returns a rotation but the directional light node only needs a vector. 

Now when opening the preferences the rotation must be restored but from the stored vector alone this is not possible because it's ambiguous. That's why the values of the rotation are additionally saved as quaternion.

If you directly modify the direction vector in the editor it will be overridden by the values that are derived from the rotation the next time you open and save the preferences dialog.

Issue #11150, opened 2023-10-23
[Problem] Tab bar icons are not meaningful


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Tabs have a freecad icon. Which is not very meaning full.
![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/fb7cbbaf-850e-4a04-83ff-751af97e1ef7)

I see 2 better solutions : 
1 - Remove the icons completely.
2 - Use the document icon that is used in the tree : 
![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/eab67b9f-dd9f-44d4-b22c-d576c6a8d051)
And for the start Page, use the Start workbench icon.

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
There are two classes in MainWindow.cpp that seem to never be used anywhere : 
`MDITabbar` and `MainWindowTabBar`

PR #11152, opened 2023-10-23
MDI tabs, meaningful icons.

PaddleStroke:
Goal : change the icons of the document tabs in the mdi tab bar.
Fixes : #11150

image

### DIFF
--- a/src/Gui/Document.cpp
+++ b/src/Gui/Document.cpp
@@ -1821,7 +1821,6 @@ MDIView *Document::createView(const Base::Type& typeId)
 
         view3D->setWindowTitle(title);
         view3D->setWindowModified(this->isModified());
-        view3D->setWindowIcon(QApplication::windowIcon());
         view3D->resize(400, 300);
 
         if (!cameraSettings.empty()) {
--- a/src/Gui/View3DInventor.cpp
+++ b/src/Gui/View3DInventor.cpp
@@ -57,6 +57,7 @@
 #include "View3DInventor.h"
 #include "View3DSettings.h"
 #include "Application.h"
+#include "BitmapFactory.h"
 #include "Camera.h"
 #include "Document.h"
 #include "FileDialog.h"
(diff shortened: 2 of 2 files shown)


PaddleStroke:
Here's the function in case you can nail it without having to open your ide :)

```
Py::Object MainWindowPy::addWindow(const Py::Tuple& args)
{
    PyObject* obj;
    // How to bind correctly ?
    if (!PyArg_ParseTuple(args.ptr(), "O", &obj))
        throw Py::Exception();

    if (_mw) {
        Py::Object py(obj);
        Gui::Document* document{nullptr};
        // Check if the py object has a reference to a Gui document
        if (py.hasAttr("document")) {
            Py::Object attr(py.getAttr("document"));
            if (PyObject_TypeCheck(attr.ptr(), &DocumentPy::Type)) {
                document = static_cast<DocumentPy*>(attr.ptr())->getDocumentPtr();
            }
        }

        MDIViewPyWrap* mdi = new MDIViewPyWrap(py, document);
        _mw->addWindow(mdi);
        return Py::asObject(mdi->getPyObject());
    }
    return Py::None();
}
```

wwmayer:
> However I also want to give the possibility to set custom icons. So that spreadsheet for example can set its own icon for its tabs.

I don't see the advantage of this change. Currently, every special MDI view shows its own icon (spreadsheet, image view, tech draw). But with this PR all MDI views would show the same icon by default and this makes it pretty useless because a user won't know any more what's behind a tab.

Since the issue complains about the standard icon for a View3DInventor all what needs to be changed is the window icon in its constructor:
```
View3DInventor::View3DInventor(...)
{
...
    setWindowIcon(Gui::BitmapFactory().pixmap("Document"));
}
```

PaddleStroke:
@wwmayer It's done correctly now. Note there was a `view3D->setWindowIcon(QApplication::windowIcon());` in the Document.cpp that needed to be removed. I don't think it served any purpose but I'm mentioning in case.

PR #11167, opened 2023-10-24
[TechDraw] - Cosmetic edge fixes

pavltom:
Two small fixes:
1) If a cosmetic circle/arc is selected, but _CosmeticLine_ command invoked followed by _OK_ press, program crashes. The same happens if straight cosmetic edge is selected, _CosmeticCircle_ command invoked and _OK_ pressed. The extended condition prevents the crash.
2) The cosmetic edge Qt object should be marked properly as "cosmetic". Though the property is not used right now, it should be set correctly.


WandererFan (approved):
LGTM.  Thanks for this.

PR #11138, opened 2023-10-21
Material: Material editor enhancements

davesrocketshop:
Continues the work of the material subsystem improvements.

This merge covers the continued development of the material editor.
The primary improvements are in the handling of 2D and 3D array properties.
These properties are now fully editable, and can be saved and restored.

The cards now separate the author and license. These were previously saved
as a single item. Future support will be provided for standard open source licenses.

Saving operations validate the cards to ensure UUIDs of materials are considered.
Warnings are given when a save could potentially impact the models, such as
saving over a material instead of creating a new instance.

The editor is still not complete. There are a number of functional elements, such as
drag/drop operations, folder creation, and deletion operations that need to be added
to the main tree. State needs to be saved and restored to improve the user experience.
The appearance preview also needs significant work. This will be handled in a future
PR.

PR #11172, opened 2023-10-24
Add CMakePresets.json, augment conda devenv, and update VSCode to use them.

oursland:
Following discussion on #9663, this work better accommodates developers by:

1. Creating a `CMakePresets.json` that provides functional build configurations for multiple environments.
2. Augment the `conda/` path with scripts to assist developers in creating local conda development environments.
3. Updating the VSCode to use `CMakePresets.json` over the plugin-specific CMakeKits, and elimination of tasks and settings that may interfere with developer's VSCode configuration.


.vscode/launch.json
```diff
@@ -2,35 +2,40 @@
...
+      ],
+      "linux": {
+        "MIMode": "gdb",
+      },
+      "osx": {
+        "MIMode": "lldb"
+      },
+      "windows": {
+        "MIMode": "gdb",
+        "miDebuggerPath": "C:\\msys64\\ucrt64\\bin\\gdb.exe"
```
kadet1090, line 30: Is this hardcoded absolute path necessary? What if someone has installed msys in another directory? Maybe msys provides some kind of env variable that we can use instead?

CMakePresets.json
```diff
@@ -0,0 +1,309 @@
+{
```
kadet1090, line 1: Is there a known way to override settings in this file without touching files tracked by git? If not, are we sure that enforcing them is ok, i.e. there is no possible reason to change anything?

.vscode/launch.json
Pesc0: I am not familiar with windows debugging, what would be the correct path? In my PR i have C:\\MinGw\\bin\\gdb.exe  and I cannot recall where its coming from. Also I've seen tasks of "type": "vscppdbg" instead of "cppdbg". Whats the correct way?

CMakePresets.json
Pesc0: > Is there a known way to override settings in this file without touching files tracked by git?

Not that I know of. It's the same problem you mentioned for the vscode folder. Git is going to pick up the change if you modify it.
However when running cmake manually you have to explicitly use this file `cmake --preset <preset-name>`. Not sure how it works for other IDEs.

.vscode/launch.json
kadet1090: That depends on the setup of the user. Most of windows users probably depends on msvc toolkit (the one from visual studio) which uses "vscppdbg" as the type. Others use gcc + gdb (coming from mingw/msys2) which must be installed separately and is not so common. I don't think that we should force anything like that. 

As for the correct path - there is no correct absolute path as it can be customized by users. I think that most msys users have the tools added to their PATH env variable, so simple `"gdb.exe"` should be good enough, we can always add that if users want to use gdb it should be available in PATH.

CMakePresets.json
kadet1090: If there is no ability to override it without touching and there is valid reason to override it, it should not be tracked in the repository. It causes problems, not only for me: https://github.com/FreeCAD/FreeCAD/pull/11138#issuecomment-1773964099. 

This PR excludes `settings.json` from the repository (big yay) - maybe there is VS Code setting which allows users to provide path to custom preset? That'd solve the issue.

kadet1090: OK, CMake authors are good devs: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#id3

> `CMakePresets.json` and `CMakeUserPresets.json` live in the project's root directory. They both have exactly the same format, and both are optional (though at least one must be present if [--preset](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-preset) is specified). `CMakePresets.json` is meant to specify project-wide build details, while `CMakeUserPresets.json` is meant for developers to specify their own local build details.
> 
> `CMakePresets.json` may be checked into a version control system, and `CMakeUserPresets.json` should NOT be checked in. For example, if a project is using Git, `CMakePresets.json` may be tracked, and `CMakeUserPresets.json` should be added to the `.gitignore`.

That basically solves this issue.

.vscode/launch.json
Pesc0: Thanks, fixed

CMakePresets.json
Pesc0: > That basically solves this issue, but we need to add CMakeUserPresets.json to .gitignore, @oursland can you do that?

Alright, this allows for individual customization, but cmakepresets might still interfere with IDEs. having the user copy it manually (like the vscode folder) ensures there will be no issues

kadet1090: I think that if IDEs support the `CMakePresets.json` they support he override too, but it should be checked.

.vscode/launch.json
adrianinsaval: does this need to be specified? can't we leave this blank and let users set it on their user config?

oursland: I derived this by following the steps provided by the [VSCode Debugging C++ guide](https://code.visualstudio.com/docs/cpp/cpp-debug) (note: they have a different path than what their instructions result in).  While I had this line included, I never successfully got the debugger to attach to the running FreeCAD after launching it.

However @kadet1090 noted that when using the MSVC++ tools, which the official FreeCAD on Windows does (MSVC++ 2019), the appropriate debugger is `cppvsdbg` as seen on the [Microsoft C++ on Windows](https://code.visualstudio.com/docs/cpp/config-msvc#_customize-debugging-with-launchjson) guide.

I am currently performing a build on Windows and will test this configuration.  If functional, I believe that `cppvsdbg` should be the default with the option for user to override as needed.

oursland: First attempt failed on my Macbook Pro running Apple M2.  It appears there's an issue when debuggin on Win 11 for ARM64 for a Windows x86/64 binary. [vscode-cpptools#10596](https://github.com/microsoft/vscode-cpptools/issues/10596)

I have a 10 year old Windows laptop that I'll attempt to build and test on.

oursland: I did manage to build and debug in Windows 10 on x86-64, however the launch from debugger consistently failed wheras attaching to a running instance worked as expected.

I noticed that the behavior on Windows when running from the command line differs from Linux and macOS. Under Linux and macOS, launching from the command line the process is attached to the terminal.  Under Windows, the program forks and detaches, returning the user to a command prompt.  I am uncertain if this behavioral difference is responsible for the debugger failing to launch the program without error.

Pesc0: Alright looks like we need an entirely new debug config for windows. I'll try to setup a windows dev env when I have some time next week. So we need `cppvsdbg` and `attach` yeah?

oursland: From what I can tell, this is the best option for Windows development in VS Code.  I suspect most development is on Linux and macOS, but debugging Windows-specific issues will require Windows tooling.

oursland: [This commit](https://github.com/FreeCAD/FreeCAD/pull/11172/commits/b43a19e7dad1da39e8058a04ce2dd123375c6656) shows how to add the debugger attach for Windows in `launch.json`.  Unfortunately, I have not seen a way to have launch configurations that are platform-specific.  According to the [documentation](https://code.visualstudio.com/Docs/editor/debugging#_platformspecific-properties) it seems that only properties can be platform-specific.

Pesc0:
I think all the changes have been merged, is it safe to close this?

PR #11174, opened 2023-10-25
[Sketcher] Tool Parameter widget + Offset tool

### LINKED ISSUE #5839: Add an Offset tool to the Sketcher
Issue imported from https://tracker.freecad.org/view.php?id=3708

* **Reporter:** normandc
* **Date submitted:** 11/24/2018
* **FreeCAD version:** 
* **Category:** Feature
* **Status:** assigned
* **Tags:** 

# Original report text

Offset is an essential tool of sketching. The offset tool should minimally work on lines, arcs, circles, and polylines made of lines and arcs. Making it work on complex curves (conics, B-splines) may prove difficult and be put aside for the initial implementation. Geometric constraints should be automatically added (coincident constraints for arcs, parallel constraints for lines, add appropriate constraints for polylines), as well as a distance constraint defining the offset value.

# Other bug information

* **Priority:** normal
* **Severity:** feature
* **Category:** Feature
* **Updated:** 2/6/2021

# Discussion from Mantis ticket

### Comment by abdullah 2018-11-27 14:33
Offset curves will be implemented for all existing geometry in v0.19. I am uncertain ATM how to treat constraints though. I will definitely start a thread on this at the beginning of the v0.19 development cycle.

---

### Comment by abdullah 2018-11-27 14:38
I realise now that you refer to the constraints making the offset. We do not need those constraints. An offset curve, which exist in OCC, is defined as a point on curve A plus a vector perpendicular to the curve on the plane (there is a vector product involved with the normal of the plane). Something like: Po(u)=P(u)+ alpha*(t x n).

Let's discuss it anyway in due time (the sooner v0.18 gets release, the sooner the v0.19 development cycle starts)...

---

### Comment by Kunda1 2019-09-12 21:28
Discussion thread: https://forum.freecadweb.org/viewtopic.php?f=8&t=39269

---

### Comment by Kunda1 2020-04-22 11:12
Other discussion threads:
"[Sketch offset tool](https://forum.freecadweb.org/viewtopic.php?f=3&t=21983)"
"[Feature requests: Enabling FreeCAD for sewing pattern making](https://forum.freecadweb.org/viewtopic.php?f=8&t=45550)"

### LINKED ISSUE #7221: [Feature Request] Sketcher rectangle from 3 corners

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=8&t=70367

### Subproject(s) affected?

Sketcher

### Idea description

The centered rectangle tool is added to 0.20 is really handy. However, I'd like to suggest adding one more tool for rectangles - rectangle from 3 corners. It could be useful for example to fit a rectangle in existing space in the model or when drawing a skewed rectangle (this tool would not make all the edges horizontal/vertical).

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #10436: [Problem] Sketcher tools should be able to have a taskbox

### Full version info

```shell
0.22 dev
```

### Subproject(s) affected?

Sketcher

### Problem description

The taskview is supposed to be a view showing settings of commands.
Yet in sketcher the taskview is used as a 'Widget area' and the sketcher tools can't have a TaskBox.

This is very limiting as tools cannot have settings. Because of that, many tools are crippled and/or have inconsistent UIs using qmessage boxes. Some tools require multiple tools because of that : 
- copy/clone/move/rectangular-array.
- Fillet / Fillet preserving corner.
Everytime you want to add an option to a sketcher tool, as you can't add a setting to the toolbox, you have to either duplicate the tool or other bad solutions.

Some tools don't even exist because of that limitation : Offset, circular pattern... 

So besides being inconsistent with the rest of Freecad, it is crippling existing tools and further development.

My suggestion is to make it possible for sketcher tools to open a taskbox in the taskview as on the following screenshot : 
image

### Code of Conduct

PaddleStroke:
Fixes #10436
Fixes #7221
Fixes #5839

This PR adds parameters to sketcher tools. Some settings on the 3d view directly. Some settings in a taskbox widget on the taskview.

This PR include : 
- Tool Parameter framework.
- Point tool
- Line tool
- Rectangle tool
- Offset tool
- Polygon tool


PaddleStroke:
@abdullahtahiriyo Lint error for which I am not sure what to do : 

DrawSketchControllableHandler : 
> "virtual" is redundant since function is already declared as "override"
I am not sure what should be done (supress or remove virtual) because for some functions Lint doesn't show this message.

DrawSketchDefaultHandler :
> member access into incomplete type 'SketcherGui::ViewProviderSketch'

src/Mod/Sketcher/Gui/DrawSketchControllableHandler.h
```diff
@@ -0,0 +1,187 @@
...
+        : DSDefaultHandler(constructionmethod)
+        , toolWidgetManager(static_cast<HandlerType*>(this))
+    {}
+
+    virtual ~DrawSketchControllableHandler() = default;
+
+    /** @name functions NOT intended for specialisation */
+    //@{
+    virtual void mouseMove(Base::Vector2d onSketchPos) override
+    {
```
abdullahtahiriyo, line 72: I told you the story of virtual in another comment. The current guidelines are:
1. In the base class, it needs to be virtual and cannot be override/final.
2. In all derived classes, virtual needs to be removed and override or final used

src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h
```diff
@@ -0,0 +1,1120 @@
...
+                    } break;
+                    case Sketcher::Vertical: {
+                        auto c = std::make_unique<Sketcher::Constraint>();
+                        c->Type = Sketcher::Vertical;
+                        c->First = (geoId2 != Sketcher::GeoEnum::GeoUndef ? geoId2 : geoId1);
+                        AutoConstraints.push_back(std::move(c));
+                    } break;
+                    case Sketcher::Tangent: {
+                        Sketcher::SketchObject* Obj =
+                            static_cast<Sketcher::SketchObject*>(sketchgui->getObject());
```
abdullahtahiriyo, line 709: The Lint, I think it is a false positive. Not sure what can be done other than suppressing it.

Insofar as I understand, if the type where unknown (needing a forward declaration), it would be a compiler error. If it were an access to a member for which the object is unknown, it would be a compiler error.

But let's ask a qualified member...
@wwmayer 

What do you think about this one?

abdullahtahiriyo: My wrong. It is not a mere linter. It is a full fledged compiler warning. It is solved now. Sorry for the noise.

wwmayer: A diagnostic error isn't something that can be suppressed but must be fixed.

Often such code still compiles fine because the .cpp file includes the missing class declaration beforehand but the clang code model (i.e. clang-tidy) won't be able to see this because it only parses this header file and header file it includes. If for the required type only a forward declaration is visible the clang code model will fail.

abdullahtahiriyo:
I guess it adds the polygon tool too. You may want to update the PR description (for testers to know what to expect).

PaddleStroke:
@abdullahtahiriyo There is something we did not cleaned up its the toolbar UI of rectangle tool.
Currently I have left the toolbar as it was, which means 3 tools : rectangle, rectangle by center and oblong.
But Oblong just opens the rectangle tool (without checkbox checked) !
Rectangle by center opens the rectangle tool on the correct mode so that's OK.

- My consensual suggestion is to remove Oblong from the toolbar.
- What I think would be best is to put a single rectangle tool (remove oblong and rectangle by center).

PaddleStroke:
Ah yes I forgot about polygon. I added it to the description.

To update my UI preference, I would rather do this : 
- Remove rect by center, remove oblong, remove the polygon comp.
- Make a single 'shape' comp in which we put : rectangle, slot, polygon.
Would that be acceptable for you?

abdullahtahiriyo:
> Currently I have left the toolbar as it was, which means 3 tools : rectangle, rectangle by center and oblong.

It sounds as a feature rather than an oversight when we look into merging this PR fast. The last we want is to start a toolbar discussion at this point.

Of course, the result is suboptimal after this PR is merged and needs to be addressed.

> But Oblong just opens the rectangle tool (without checkbox checked) !

Oblong = rounded corners rectangle.

This sounds as a bug and needs to be fixed (it needs open with the checkbox checked).

> To update my UI preference, I would rather do this :
> 
>     * Remove rect by center, remove oblong, remove the polygon comp.
> 
>     * Make a single 'shape' comp in which we put : rectangle, slot, polygon.
>       Would that be acceptable for you?

I have no big problems. 

From a practical point of view I do not make polygons extremely often and I could leave with having a polygon in a comp in which the representative is a rectangle. I could also leave with slot under rectangle and polygons separately. I would not cry either if I end up keeping slot and polygons separated. However, others might not be so forthcoming (rectangle not being a regular polygon, slot not even being an irregular polygon... being some of the arguments).  

At least in the past, this caused very heated discussions. We deliver functionality first. When everybody can play with it, we discuss in the forum and see how it goes. This is a kind of community decision.

PaddleStroke:
Ok so maybe its best to merge as it is now (ie rectangle comp replaced by single rectangle tool and the rest is unchanged.) Then we can discuss the UI in deeper discussion

abdullahtahiriyo:
I would not dare to remove the center and rounded rectangle without a UI discussion.

Is it a problem to just fix the oblong?

PaddleStroke:
Oblong can be fixed.

But I think ultimately we don't want to keep these 3 tools icon. Because
why having a oblong icon and not a oblong by center? And why not a 3p
rectangle? If we follow this logic the command group can have 16 icons.

But yes this would still be a UI decision. And fixing is not too hard so if
you prefer I can fix it.

PaddleStroke:
I have just forced push to rollback previous commit, and instead just fixed the oblong tool so that it opens diagonal + round corner. Such that the ui is not changed. After this PR, people can play a little bit with it and we can discuss the UI at this point.

I am not sure it is worth waiting until after the weekend, we will have more people who will feedback if it merges. But waiting is fine for me as well.

maxwxyz:
@PaddleStroke is it possible to group the horizontal and vertical constraints into a drop down and the new dropdown icon would invoke a new "straight" command (depending if the line is closer to horizontal or vertical).
This would clean up the toolbar and help find the right command when the view is rotated or you look from another side onto the sketch. Horizontal currently has nothing to do with the view.

PaddleStroke:
@abdullahtahiriyo I have pushed 4 commits.
They do : 
- Add a function 'canGoToNextMode' that check if the mode can be changed on mouse button pressed. This let us prevent validation of a mode on double click on the same spot for example. To avoid creating null shapes.
- Fix the thing raised by Syres and DrMacro. Rectangle is drawn on the side where the mouse is by default. If mouse has not moved then it works as it did before.
- Fix some syntaxes and some small issues.

abdullahtahiriyo:
I have diff-ed against my branch and had a quick look at it.

A quick one on ```Gui::NotifyError```:

NotificationsFramework(3)

The functions of Gui Notification Framework all need an already translated or a translatable string (a string that is marked for translation and exists in the framework), see examples in the chart above (left branch). 

A dynamically generated string, such as this:
```
Gui::NotifyError(sketchgui, "Failed to add box: %s\n", e.what());
```
cannot satisfy this requirements. Not even if marked for translation.

Now, one example where this happens, is in this substitution:

```
-            Base::Console().Error("Failed to add box: %s\n", e.what());
+            Gui::NotifyError(sketchgui, "Failed to add box: %s\n", e.what());^M
             Gui::Command::abortCommand();
             THROWM(Base::RuntimeError,
-                   "Tool execution aborted\n")  // This prevents constraints from being applied on
-                                                // non existing geometry
+                   QT_TRANSLATE_NOOP(^M
+                       "Notifications",^M
+                       "Tool execution aborted") "\n")  // This prevents constraints from being^M
+                                                        // applied on non existing geometry^M
```

1. This exception ultimately originates in a Python command via the Interpreter. This means that the interpreter has already sent a developer notification to the report view of the exception. The e.what() is already there.
2. We do not want to notify the e.what() to the user, as it is intended only for the developer.
3. NotifyError is a notification intended for both the user and the developer. This needs to have a context for translation, so that the same message can be send untranslated to the developer and translated to the user (see example in the chart, bottom-left). Then the way is just to send the "Failed to add box" string (properly marked for translation).

An example from elsewhere is:
```
Gui::NotifyError(sketchgui,
                             QT_TRANSLATE_NOOP("Notifications", "Error"),
                             QT_TRANSLATE_NOOP("Notifications", "Failed to add arc"));
```

There are two places in the diff where this NotifyError is not properly used. This need to be changed. Please, do bring in these changes.

I will read the code for a last time and then I think we are ready for a merge.

src/Mod/Sketcher/Gui/DrawSketchController.h
```diff
@@ -0,0 +1,573 @@
...
+            color = hGrp->GetUnsigned("DeactivatedConstrDimColor", color);
+            dimConstrDeactivatedColor.setPackedValue((uint32_t)color, transparency);
+        }
+    };
+
+public:
+    /** Creates the controller.
+     *  @param dshandler a controllable DSH handler
+     */
+    DrawSketchController(HandlerT* dshandler)
```
abdullahtahiriyo, line 187: Please add an explicit here, as in ```explicit DrawSketchController(```

This is to avoid implicit conversions and it is a very good thing.

src/Mod/Sketcher/Gui/DrawSketchController.h
```diff
@@ -0,0 +1,573 @@
...
+    {
+        adaptParameters(lastControlEnforcedPosition);  // specialisation interface
+    }
+    //@}
+
+    /** @name Specialisation Interface */
+    /** These functions offer a specialisation interface. Non-virtual functions are specific to
+     * this controller. Virtual functions may depend on input from a derived controller, and thus
+     * the specialisation needs to be of an overriden version (so as to be able to access members
+     * of the derived controller).
```
abdullahtahiriyo, line 303: Please "overridden".

src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h
```diff
@@ -0,0 +1,1127 @@
...
+                    generateAutoConstraints();
+
+                    beforeCreateAutoConstraints();
+
+                    createAutoConstraints();
+                }
+
+                tryAutoRecomputeIfNotSolve(
+                    static_cast<Sketcher::SketchObject*>(sketchgui->getObject()));
+            }
```
abdullahtahiriyo, line 476: I will handle these three in a separate PR. I am starting to use qtcreator and it seems we have quite some of this issues, which were not detected by kdevelop.

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
```diff
@@ -0,0 +1,371 @@
...
+    using WParameter = SketcherToolDefaultWidget::Parameter;
+    using WCheckbox = SketcherToolDefaultWidget::Checkbox;
+    using WCombobox = SketcherToolDefaultWidget::Combobox;
+    //@}
+
+    using SelectMode = SelectModeT;
+    using ControllerBase::handler;
+
+public:
+    DrawSketchDefaultWidgetController(HandlerT* dshandler)
```
abdullahtahiriyo, line 116: add explicit

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
```diff
@@ -0,0 +1,371 @@
...
+
+        ControllerBase::finishControlsChanged();
+    }
+    //@}
+
+    /** @name Specialisation Interface */
+    /** These functions offer a specialisation interface. Non-virtual functions are specific to
+     * this controller. Virtual functions may depend on input from a derived controller, and thus
+     * the specialisation needs to be of an overriden version (so as to be able to access members
+     * of the derived controller).
```
abdullahtahiriyo, line 168: overridden

src/Mod/Sketcher/Gui/DrawSketchHandlerOffset.h
```diff
@@ -0,0 +1,1142 @@
...
+                mkWire.Add(TopoDS::Edge(Obj->getGeometry(vCC[i][j])->toShape()));
+            }
+            sourceWires.push_back(mkWire.Wire());
+        }
+    }
+
+    void findOffsetLength()
+    {
+        double newOffsetLength = LONG_MAX;
+
```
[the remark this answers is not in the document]
PaddleStroke: No specific reason. I think we should change to DBL_MAX  ?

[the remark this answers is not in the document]
abdullahtahiriyo: Makes sense.

abdullahtahiriyo:
@PaddleStroke 

Please, add those changes. I will handle the incomplete type lints that compile fine in a separate PR, while I make myself acquainted with qtcreator. Then, we can merge.

PaddleStroke:
I will fix those asap.

PaddleStroke:
I have just done the small fixes and squashed the commits.
Do you want me to squash [48be53c](https://github.com/FreeCAD/FreeCAD/pull/11174/commits/48be53c38d0602723a6c4b3fc7be907b0335b0d8) ? If yes to which commit ?

abdullahtahiriyo:
@PaddleStroke 

Thank you and sorry for the delay.

PaddleStroke:
Thanks for handling this large project :)
By the way, I'm already working on the other DSH and have a few ready already. I'll open PRs very soon so don't go too far 😄

abdullahtahiriyo:
@PaddleStroke 

I am currently doing so fixing of issues with lack of headers and other code issues from static tools.

I would like to handle the feedback in the forum thread with a certain degree of priority. Not only the one we have now, but also the one that should arrive over the weekend and the next week.

My biggest concern for this work was the framework. You have also gained substantial ability to use it (and even adapt it). I do not expect major issues with PRs porting existing tools to it. 

I think the single most important issue warranting merit at this point is in making all users comfortable with the introduction of the framework. Not necessarily by catering all their requests, but by actively listening and discussing what can be done to improve.

PaddleStroke:
> I do not expect major issues with PRs porting existing tools to it.

No but the tools still need to be tested to make sure I didn't missed any bugs. There are many cases so its easy to forget one.

> Not necessarily by catering all their requests, but by actively listening and discussing what can be done to improve.

Of course!

maxwxyz:
@PaddleStroke thank you for your work! Do you also have a plan to implement this in the Polyline tool?

PaddleStroke:
@maxwxyz Thanks !
Polyline tool is a tough one. I actually did it in my very early version of tool settings, and I remember it was very hard as the tool is already very complex. It's not in my imediate priority but one day I may come to it. Or if someone wants to do it, I think it's a nice project for a new developper. As its only one file its not so complex programatically. But the logic is complex. I would suggest that you open an issue for it.

maxwxyz:
OK If I finally am able to build FC, maybe I can dive in (total newbie).
Is the on view parameter overlay going to replace the normal dimension window (double clicking on dimensions or using contextual dimension tool)? Currently there are the two forms now.

PaddleStroke:
> Is the on view parameter overlay going to replace the normal dimension window (double clicking on dimensions or using contextual dimension tool)? Currently there are the two forms now.

This is something I actually worked on before. It probably won't be a simple spinbox as we have in on View Parameters because of the additional features. But my prototype was looking good. There was one issue with the expression dialog that got this feature stuck. I haven't been back to it yet.

https://www.youtube.com/watch?v=W5sP93yWt8c&ab_channel=FreeCADDevDiary

Issue #9205, opened 2023-04-06
Problem:  Default settings make it hard to select objects in the 3D window


sliptonic:
Users find it difficult to select individual edges or vertices in the 3D window.  The accuracy of the mouse pointer over the target is controlled by the 'Pick Radius' setting.  The setting is available in the Display Group in the 'Rendering' section
![2023-04-06_10-38](https://user-images.githubusercontent.com/538057/230429210-69aedff9-015c-4c06-bbba-89203abde363.png)

This has poor discoverability for the user.

The default setting is 5px which is far too small for modern displays.  I find a setting as large as 25 to still be usable on my display.

A better place for the setting is in the General group on the Selection tab.
![2023-04-06_10-39](https://user-images.githubusercontent.com/538057/230429704-6649353f-b3b4-406a-9b32-c4f56b048526.png)

howie-j:
Also related to this issue; increasing the pick radius makes it even more difficult to select internal/concave edges (inside edge of a box for example).

PaddleStroke:
Also when you increase the pick radius, then a line that is on front will have priority from another line even if the cursor is closer to this other line. On picture the highlighted line is hard to select.
![image](https://user-images.githubusercontent.com/19984177/231459578-5724ad84-76e0-4a1c-89c7-48821b471366.png)

PaddleStroke:
Regarding improving the selection of things I had a look. Here are my notes for reference. Please note that I'm not 100% sure about those.
 
- It seems that the selection rely on `View3DInventorViewer::getPointOnRay`.
This function seems to be getting a single `SoPickedPoint `with `SoRayPickAction.getPickedPoint()`.
- This SoPickedPoint seems to be processed in the different workbenches. For example in sketcher it is processed in : `EditModeCoinManager::detectPreselection`.

So the selection seems to rely on SoRayPickAction. From [its doc](https://www.openinventor.com/reference-manuals/NewRefMan1030/RefManJava/com/openinventor/inventor/actions/SoRayPickAction.html) we can read that it is also possible to get a list of points hit by the ray and not just one : 

> Callers can request the action to compute all intersections along the ray (sorted closest to farthest) by setting the pickAll flag to true. By default, the action computes only the closest intersection.

So this makes me think that by using `getPickedPoint` we only get the item that is the closest to the camera.

[...]

PR #11187, opened 2023-10-26
Gui: issue #9205: Default settings make it hard to select objects in …

wwmayer:
…the 3D window


adrianinsaval:
@wwmayer do you think we should backport this? we're planning on making a point release this weekend after back porting more fixes

wwmayer:
Sure, why not? The change is trivial.

adrianinsaval:
backported to 0.21

PR #11188, opened 2023-10-26
[TechDraw] - Line Decoration Improvements

pavltom:
This PR improves line decoration handling:
1) Double clicking on an edge invokes Line Decoration dialog. This aims to improve productivity when creating a drawing. The changes involved are in files QGIEdge.h and QGIEdge.cpp .
2) When the Line Decoration dialog is invoked, all updates to style are immediately reflected. This is perfect. However when _Cancel_ is pressed, the old style values were not stored anywhere and in fact, there was no difference between pressing _OK_ or _Cancel_. It just closed the dialog. With this PR, the line styles are fully preserved and restored, if the user decides to press _Cancel_. These changes are contained in TaskLineDecor.h and taskLineDecor.cpp files.
3) Small fix of earlier refactorization. Without setting the pen style, all lines are drawn as solid. Code change in QGIPrimPath:311.

I hope you will find this addition useful, in case of any issues or questions, please leave a comment!

### DIFF
--- a/src/Mod/TechDraw/Gui/QGIEdge.cpp
+++ b/src/Mod/TechDraw/Gui/QGIEdge.cpp
@@ -32,9 +32,14 @@
 #include <App/Material.h>
 #include <Base/Console.h>
 #include <Base/Parameter.h>
+#include <Gui/Control.h>
+#include <Mod/TechDraw/App/DrawUtil.h>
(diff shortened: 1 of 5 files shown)


WandererFan:
Thank you!  This was on my list and now I can cross it off!

PR #11190, opened 2023-10-26
Sketcher_Dimension: If radius of arc is not DoF, then offer arc angle first

### LINKED ISSUE #11165: [Feature Request] Sketcher: Contextual Dimensioning - Arc Angle

### Problem description

The contextual dimension tool currently proposes to constraint the radius of a circle on the first click of the arc line. You can toggle through it with "M" to switch to diameter and arc angle which works as expected.
Ideally, when the radius or diameter of an arc is defined/constrained the next or first click on the line of the arc with the contextual dimensioning tool should propose to constraint the arc angle (and if implemented to cycle through with "M" to the arc length).
In the case that the arc is already fully constraint by other dimensions the fallback could be to an over constrained situation or even a driven dimension. If it is possible to detect that the arc radius is already defined by other constraints, the tool should also popose to dimension the arc angle with the click of the arc line.

_Comment by @PaddleStroke in https://github.com/FreeCAD/FreeCAD/issues/10605#issuecomment-1776871732_

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Sketcher

### Anything else?

_No response_

### Code of Conduct

PaddleStroke:
Fixes #11165
When the user select an arc with Sketcher_Dimension, if the arc's radius is not a DoF, then the tool offer arc angle as a first option.
@maxwxyz it is working in all cases where the radius is constrained so that's nice.

### DIFF
--- a/src/Mod/Sketcher/App/SolverGeometryExtension.h
+++ b/src/Mod/Sketcher/App/SolverGeometryExtension.h
@@ -191,6 +191,10 @@ public:
         {
             return pstatus[0];
         }
+        bool isRadiusDoF() const
+        {
+            return pstatus[0] == Dependent;
+        }
         ParameterStatus getStartParameter() const
         {
             return pstatus[1];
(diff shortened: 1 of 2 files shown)


maxwxyz:
thanks! Can I somehow link this PR to the issue so that the issue gets closed after a merge?

PaddleStroke:
Typing 'Fixes #11165' as I did is enough to close the issue automatically.

chennes:
Works well for me, thanks!

PR #11195, opened 2023-10-26
SplashScreen: add parameter to disable messages on the splashscreen

PaddleStroke:
Add a parameter to disable messages of the splashscreen. Boot faster without loosing the splashscreen.


yorikvanhavre:
Nice. Don't forget to update https://wiki.freecad.org/Fine-tuning afterwards!

PaddleStroke:
Thanks I will update based on your comments

PaddleStroke:
@wwmayer any further comment ? Else maybe it can be merged

adrianinsaval:
Added to Fine-Tunning wiki page, this seems to have a noticeable effect. It could be a good idea to make this default.

PR #11198, opened 2023-10-27
Fix issue #9465 

lekamm:
[Problem] Coin version on Mac Conda builds causes transform crash - see [issue](https://github.com/FreeCAD/FreeCAD/issues/9465) for detailed explanation.

### DIFF
--- a/src/Gui/SoFCCSysDragger.cpp
+++ b/src/Gui/SoFCCSysDragger.cpp
@@ -93,6 +93,7 @@ void TDragger::initClass()
 TDragger::TDragger()
 {
     SO_KIT_CONSTRUCTOR(TDragger);
+    this->ref();
 
     SO_KIT_ADD_CATALOG_ENTRY(translatorSwitch, SoSwitch, TRUE, geomSeparator, "", TRUE);
     SO_KIT_ADD_CATALOG_ENTRY(translator, SoSeparator, TRUE, translatorSwitch, "", TRUE);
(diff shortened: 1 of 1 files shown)


wwmayer:
This PR causes a memory leak. In #9465 I have provided any alternative patch.

wwmayer:
Thanks looks good now.

adrianinsaval:
if possible we should backport this

wwmayer:
In case it needs to be backported then only incrementing the reference counter in the constructor of the classes TDragger and RDragger (TPlanarDragger doesn't exist in v0.21) must be changed
```
#if defined(Q_OS_MAC)
    this->ref();
#endif
```

However, it should be checked whether v0.21 crashes at all because IIRC the problem only appeared with v0.22.

adrianinsaval:
First reports where with 0.21, we somehow worked around it by using an older coin package but the underlying issue I think is this.

wwmayer:
The underlying issue we actually still don't know. To me the implementations of the node kits look correct but for some reason at some point a reference is stolen so that the nodes are destroyed too early.

Do you know which older Coin version has worked and since which version we get this crash?

adrianinsaval:
IIRC it was the same coin version but the one that worked was compiled against boost 1.74 and the one that didn't work was compiled against boost 1.78. See https://github.com/FreeCAD/FreeCAD/issues/9465#issuecomment-1664385303 and following comments

adrianinsaval:
I'll make tomorrow a build with the new package that uses boost 1.82 and ask people to test if it still crashes.

adrianinsaval:
no luck, backported now as c6c35bd2d36b4f16e15932b1d9087ee892acd5cd

PR #8320, opened 2023-02-01
[workaround][GUI][Spreadsheet] Disable Accessibility on the SpreadSheetTableView

acpopescu:
Windows Accesibility will try to navigate through all the cells of the table, causing memory use and a 30 second freeze in the application. It can be reproduced by copying or selecting in an empty spreadsheet.

Issue Ref: https://github.com/FreeCAD/FreeCAD/issues/8265 

The proper solution is to implement a correct accessible interface for the SheetTableView, rather than rely on the built-in QT one.

This PR is the alternate solution - a workaround :( to disable the built-in accessibility support :( which is causing the delay.

APPSpreadSheetGui.cpp was reformatted using the latest rules.

Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [x]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---

PR #11199, opened 2023-10-27
Bugfix/issue 10144/free cad freezes when clicking inside spreadsheet

lekamm:


wwmayer:
Your PR contains code of the Coin3D issue.

Unfortunately, disabling the line `QAccessible::installFactory` is not an option as this would re-introduce the very same problem on Windows (see #8320) where the default Qt accessibility mechanism makes FreeCAD to freeze.

In order to fix the issue for both OSes the correct fix should be
```
#if defined(Q_OS_WIN32)
QAccessible::installFactory(SpreadsheetGui::SheetTableViewAccessibleInterface::ifactory);
#endif
```

wwmayer:
In order to keep the history clean I have created a new PR. So closing this one.

Issue #11196, opened 2023-10-27
[Problem] Shift-up and Shift-dn no longer select list items


bgbsww:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

See (https://github.com/FreeCAD/FreeCAD/commit/bf9584a90ba90984b8c3fb804025acba11795d2b#comments)

Long standing functionality has been disabled by the use of new accelerators for the shift with Arrow keys.  I'd submit a patch, but I'm really not sure what keys should be used instead.

Further exploration:  lines 3832 and 3857 reassign Shift-Left and Shift-Right which were previously used at lines 1442 and 1467 by the standard view rotation.

A search of the source code reveals that none of Ctrl arrows, Ctrl-Shift arrows nor Alt arrows appear to be in use, although Alt is only  used once for anything.  Testing Tools->Customize->Keyboard with either Ctrl or Alt arrows seems to work fine, so I suspect that changing these to Ctrl-Up, Ctrl-Down, Ctrl-Left and Ctrl-Right is the appropriate solution.
### Full version info

```shell
[code]
OS: Ubuntu 22.04.3 LTS (ubuntu:GNOME/ubuntu)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34784 (Git)
Build type: Unknown
Branch: main
Hash: 61c1b2bfc2339b272d3fa44bb9dbb6bdba528a64
Python 3.10.12, Qt 5.15.3, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.2
Locale: English/United States (en_US)
Installed mods: 
  * CfdOF 1.21.7
  * FreeCAD_BearingsWB 0.0.2
  * LCInterlocking
  * fcgear 1.0.0
  * DynamicData 2.53.0
  * BIM 2021.12.0
  * parts_library
[/code]
```

### Subproject(s) affected?

None

### Anything else?

https://github.com/FreeCAD/FreeCAD/blob/main/src/Gui/CommandView.cpp  around lines 3882 and 3907.  Search on 'Shift+Up', 'Shift+Down".

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bgbsww:
I created this patch: https://github.com/FreeCAD/FreeCAD/pull/11206

However, there are still other overlay accelerators still using Shift  ( F4, F3 ) for toggling transparent.  This may want some thought and discussion.

bgbsww:
[...]

PR #11206, opened 2023-10-27
Change overlap window arrow keys to use Ctrl instead of Shift

bgbsww:
See https://github.com/FreeCAD/FreeCAD/issues/11196


luzpaz:
Do these changes impact anything else ?

bgbsww:
No, these changes affect only the Up, Down, Left, and Right shortcuts for the overlay window.  These change to using the Ctrl key insttead of the Shift key.

PR #11216, opened 2023-10-27
Base: set sys.last_vars to enable post mortem debugging after exception

edwilliams16:
This enables python debuggers to stack trace after an exception. See https://forum.freecad.org/viewtopic.php?p=714349#p714349
This is authored by @bdieterm I'm just submitting it.

### DIFF
--- a/src/Base/Interpreter.cpp
+++ b/src/Base/Interpreter.cpp
@@ -56,10 +56,15 @@ PyException::PyException(const Py::Object &obj) {
 
 PyException::PyException()
 {
-    PyErr_PrintEx(1);
-        // indirectly assign the proper value to sys.last_traceback
-        // to enable python stack tracing after exception
     PP_Fetch_Error_Text();    /* fetch (and clear) exception */
+
+    // restore and print the exception with set_sys_last_vars to make post-mortem debugging work
+    // PyErr_Restore will take ownership of the three passed objects
+    PyErr_Restore(PP_last_exception_type, PP_PyDict_Object, PP_last_traceback);
+    PP_last_exception_type = nullptr;
+    PP_PyDict_Object = nullptr;
+    PP_last_traceback = nullptr;
+    PyErr_PrintEx(1);
 
     setPyObject(PP_PyDict_Object);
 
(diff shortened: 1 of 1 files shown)


edwilliams16:
Looks like the 'fix' breaks some of the internal exception processing. eg   testing matrix inverse in the Spreadsheet tests
Beyond my ability...

src/Base/Interpreter.cpp
```diff
@@ -56,6 +56,9 @@ PyException::PyException(const Py::Object &obj) {
 
 PyException::PyException()
 {
+    PyErr_PrintEx(1);
+        // indirectly assign the proper value to sys.last_traceback
+        // to enable python stack tracing after exception
     PP_Fetch_Error_Text();    /* fetch (and clear) exception */
```
bdieterm, line 62: This change will make both the post-mortem debugging as well as the unit tests work:
```suggestion
    PP_Fetch_Error_Text();    /* fetch (and clear) exception */

    // restore and print the exception with set_sys_last_vars to make post-mortem debugging work
    // PyErr_Restore will take ownership of the three passed objects
    PyErr_Restore(PP_last_exception_type, PP_PyDict_Object, PP_last_traceback);
    PP_last_exception_type = nullptr;
    PP_PyDict_Object = nullptr;
    PP_last_traceback = nullptr;
    PyErr_PrintEx(1);
```

bdieterm (changes requested):

wwmayer:
This change causes several problems:
* Exceptions are always reported and it's not possible any more to suppress them when needed
* In most cases exceptions are reported twice (if the Python redirection is enabled)
* It breaks PyException::setPyException and PyException::raiseException

A much less problematic way is modify PyException::ReportExsception():
```
void PyException::ReportException () const
{
    if (!_isReported) {
        _isReported = true;
        // set sys.last_vars to make post-mortem debugging work
        PySys_SetObject("last_traceback", PP_last_traceback);
        Base::Console().DeveloperError("pyException","%s%s: %s\n",
            _stackTrace.c_str(), _errorType.c_str(), what());
    }
}
```

bdieterm:
Thank you very much @wwmayer for both, the correct code and the analysis!
I just tested this and post-mortem debugging works great -- even when executing a FCMacro in the FreeCAD macro editor.
Really impressive and elegant -- and actually just a single line of code.

@edwilliams16: Could you please also update the title of this pull request to something like
"Base: set sys.last_vars to make post-mortem debugging work"

PR #11223, opened 2023-10-28
Part: format XML files

wwmayer:
This PR is about applying formatting rules to our XML files. It's basically about replacing tabs with four spaces.
The files have been formatted with the utility at https://github.com/FreeCAD/FreeCAD/tree/main/src/Tools/xmlformat


luzpaz:
What's the ETA on this ?

PR #11226, opened 2023-10-29
Correct code tag for info

Kuzma30:
Change [code] to ``` as code tag don't recognize and user need manualy correct tag when post About info on github.

### DIFF
--- a/src/Gui/Splashscreen.cpp
+++ b/src/Gui/Splashscreen.cpp
@@ -800,7 +800,7 @@ void AboutDialog::copyToClipboard()
             deskInfo = QLatin1String(" (") + deskEnv + QLatin1String("/") + deskSess + QLatin1String(")");
     }
 
-    str << "[code]\n";
+    str << "```\n";
     str << "OS: " << prettyProductInfoWrapper() << deskInfo << '\n';
     str << "Word size of " << exe << ": " << QSysInfo::WordSize << "-bit\n";
     str << "Version: " << major << "." << minor << "." << point << suffix << "." << build;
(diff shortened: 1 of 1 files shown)


luzpaz:
Maybe we can figure out how to support the 3 backticks in the forum too ?

Kuzma30:
> Maybe we can figure out how to support the 3 backticks in the forum too ?

https://forum.freecad.org/app.php/help/bbcode
at the end of document


Can I add my own tags?
    If you are an administrator on this board and have the proper permissions, you can add further BBCodes through the Custom BBCodes section.

luzpaz:
@Kuzma30 we've tried that but I think bbcode requires square brackets if I'm not mistaken

Kuzma30:
> @Kuzma30 we've tried that but I think bbcode requires square brackets if I'm not mistaken
> 
> it would need to be an addon for the forum or a custom hack ?

I think we need https://www.phpbb.com/customise/db/extension/markdown/

luzpaz:
@Kuzma30 we need a phpbb developer to help us with the forum. We tried to install the markdown extension and it created some problems IIRC.

Kuzma30:
@luzpaz What type of problem you get?

luzpaz:
@kkremitzki do you remember the issues that cropped up from enabling markdown in the forum? (apologies if I may be confusing you with yorik or someone else who tested this)

wwmayer:
> What type of problem you get?

Having the three apostrophes at the end causes formatting errors when creating new tickets. Once had to always fix the pasted content.

luzpaz:
ticket template should chance then ?

PR #11231, opened 2023-10-30
Find and replace http://freecad to https://freecad

DeflateAwning:


src/3rdParty/libkdtree/COPYING
```diff
@@ -1,6 +1,6 @@
 "The Artistic Licence 2.0"
 Copyright (c) 2000-2006, The Perl Foundation.
-http://www.perlfoundation.org/legal/licenses/artistic-2_0.html
+https://www.perlfoundation.org/legal/licenses/artistic-2_0.html
```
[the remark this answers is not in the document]
DeflateAwning: Good catch, fixed!

src/3rdParty/libkdtree/README.md
```diff
@@ -114,8 +114,8 @@ without the help of a number of people. Foremost, I would like to thank the
 folks from the #c++ channel on Freenode, specifically (in no particular order)
 orbitz, quix, Erwin, pwned, wcstok, dasOp, Chaku, Adrinael, The_Vulture, and
 LIM2 (if I left anyone out, let me know). Finally, I thank the Artificial
-Intelligence Laboratory of the University of Zurich, Dr. Peter Eggenberger and
-Gabriel G�mez for giving me the opportunity to write this stuff.
+Intelligence Laboratory of the University of Zurich, Dr. Peter Eggenberger, and
+Gabriel Gomez for giving me the opportunity to write this stuff.
```
chennes, line 118: Can you confirm that this change is intentional? Do you have a source for the correct spelling of this person's name?

DeflateAwning: This change was intentional. Adding the oxford comma makes it consitent thoughout. The mangled character wasn't right. Perhaps best would be to fix it with the correct unicode "ó" (confirmed spelling against the same text in this repo: https://github.com/eddyxu/kdtree-serialization)

chennes: Yes, sorry, the comma is fine, it's the person's name I was concerned about. The name appeared correctly in my IDE, but if there is a problem elsewhere, then yes, "ó" is what we should be using.

DeflateAwning: Fixed!

PR #11238, opened 2023-10-31
removed pylupdate4 used in qt4

mosfet80:
removed unused pylupdate4.
pylupdate4 are used with qt4.


luzpaz:
Address line 218 ?
`# TODO: we suppose lupdate is a symlink to lupdate-qt4 for now`

luzpaz:
@mosfet80  ?

chennes:
We can drop that TODO in a cleanup PR later on, I think.

PR #11240, opened 2023-10-31
[TechDraw] Checkable menu item for Turn View Frames On/Off

pavltom:
I would like to add the checked state for the TechDraw_ToggleFrame menu item. Though this is far from being something essential, I believe in complex drawings the state reflected by the button is really helpful.

As usual, in case of any issues, please let me know.

### DIFF
--- a/src/Mod/TechDraw/Gui/CommandDecorate.cpp
+++ b/src/Mod/TechDraw/Gui/CommandDecorate.cpp
@@ -47,6 +47,7 @@
 #include <Mod/TechDraw/App/DrawViewPart.h>
 
 #include "DrawGuiUtil.h"
+#include "MDIViewPage.h"
 #include "TaskGeomHatch.h"
 #include "TaskHatch.h"
 #include "ViewProviderGeomHatch.h"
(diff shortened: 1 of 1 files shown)


pavltom:
Seems the app crashes when there are no windows open. I will try to fix this first.

pavltom:
I am sorry for the inconvenience, should be fixed now.

WandererFan:
The code looks ok to me.  The current implementation allows the frame state to be changed even if the active view is not a TD page.   IIUC, the new code requires the user to be on a TD page to toggle frames. 

Not sure if this is a big deal or not.  I never use the big TD menu, only the context menu so I would never run into this.   Maybe others would?

I'll merge this, and we'll see if there are any complaints.

pavltom:
Hi Wandererfan. Thank you for your review and comments.

From my point of view it does not make sense to toggle TechDraw frames in other window, than the one displaying them. The most strange side effect of the former implementation was when there were 2 or more TechDraw pages, the user switched to the 3D view, pressed the Toggle Frames toolbar button and the "Select Page" dialog appeared, without any visible effect after pressing OK. This was kind of surreal :-)

PR #11244, opened 2023-10-31
[GUI][Overlay]Fix for the overlay icons and pop-up notifcation

MisterMakerNL:
Fix for the overlay icons and the pop-up notifications black background.
Notifications fix and icons for the QSS found by @obelisk79 also icons from Opendark, but changed them up a bit and renamed them. 
Icons for dark themes now look like this:
image
Instead of this:
image

Notifications now look like this in light:
image


adrianinsaval:
where are the new files used? remember that you need to add new files to cmake so they actually get installed

adrianinsaval:
btw why not make the light version of icons monochrome too?

MisterMakerNL:
> btw why not make the light version of icons monochrome too?

They are monochrome except the red close icon. Which is like a warning color, to say "watch out your going to close it".
Somehow the icons edges light up while the SVG file has nothing of it. I think this is a scaling issue with the SVG maybe we should use PNG files.

MisterMakerNL:
> where are the new files used? remember that you need to add new files to cmake so they actually get installed

I keep forgetting this step... I have added them!

MisterMakerNL:
I have also added a small fix for resizing windows regression, the lacking of a border made it impossible to resize the windows.
Before:
image
After:
image
It's probably better to make another PR but all these tiny things don't affect each other and kind of a pain to make for every small fix a new PR.

adrianinsaval:
could you extend the changes to all dark theme variants?

MisterMakerNL:
> could you extend the changes to all dark theme variants?

I kinda want too remove the other theme's reason is: they are pretty different to the default ones and I already didn't feel like they where worth the time rewriting completely like I did for the default versions.
I also think we should focus on one default Freecad experience and anything other should be addons.
We already try too do that with icons and now with a style guide book from @obelisk79. I even would suggest moving too the modern versions and removing the others, reason I didn't do that, is that I feel modern style isn't mature enough and the amount off fallback from the community. The last one is also a reason the prodark and such aren't removed already.

I'm not trying to push my work as in design, only reason I prefere the new stylesheet over the old ones is the amount of tiny bug fixes and less complicated stylesheets. There was a lot of grouping in older sheets, which on paper should make the code less complicated, but in practice linking all kind of classes/widgets together made it super annoying to fix bugs. I'm not saying that it was bad code, it was just obvious a multi-year development traject and had a long history.

adrianinsaval:
fair enough and I tend to agree

adrianinsaval:
btw we are still missing light icons for the items in the overlay menu
Screenshot_20231104_205108

MisterMakerNL:
> btw we are still missing light icons for the items in the overlay menu Screenshot_20231104_205108

Ah yes, we haven't found out how to fix those!

obelisk79:
I've tried to sort out how to override those icons and it's been a challenge. The overlay code is rather complex and structured in ways that are unfamiliar to me. I'm sure I'll get it eventually.

luzpaz:
> btw we are still missing light icons for the items in the overlay menu

Lets open a ticket for this? volunteers ?

PR #11255, opened 2023-11-02
Change overlay window arrow keys to use Ctrl instead of Shift  #11206 (clean patch)

bgbsww:


bgbsww:
Apologies, I inadvertently closed https://github.com/FreeCAD/FreeCAD/pull/11206 while cleaning up my working fork.  Resulting patch should be the same, and should reflect the same comments.

PR #11256, opened 2023-11-02
Material: Fix several warnings, build failures and crashes

wwmayer:


src/Mod/Material/App/Model.cpp
```diff
@@ -90,10 +90,6 @@ bool ModelProperty::operator==(const ModelProperty& other) const
         return true;
     }
 
-    if (&other == nullptr) {
```
davesrocketshop, line 93: This will break the code. The Linux compiler says it will have no effect but it's wrong. It's entirely possible to pass in a null value.

davesrocketshop: Unfortunately the way I read this is that you should never write code with this bug (correct) so we should never try to test for it (hard disagree) Seems silly but I'll try to correct for it in the next merge.

quarnster: > It's entirely possible to pass in a null value.

It's a reference, not a pointer. The address of a reference (&other) can not be null.

davesrocketshop: Taking the address of a reference returns the address of the original. This code works as intended.

quarnster: > Taking the address of a reference returns the address of the original. This code works as intended.

Yes, and since it is not a pointer, it is exactly the same as taking the address of non-pointer variable. I.e. it can not be null.

wwmayer: https://isocpp.org/wiki/faq/references
Read section: _What does it mean that a reference must refer to an object, not a dereferenced null pointer?_

davesrocketshop:
Gentlemen, I get the theory and concede that this is an idiotic and poorly thought out part of the language standard.

However, given that it is known that this will break the code I think it's prudent to keep it unchanged until a more properly thought out change is in place. As the saying goes, don't cut off your nose to spite your face.

src/Mod/Material/App/Model.cpp
davesrocketshop: Gentlemen, I get the theory and concede that this is an idiotic and poorly though out part of the language standard.

However, given that it is known that it will break the code I think it would be prudent to leave it unchanged until a proper fix can be put in place. As the saying goes, don't cut off your nose to spite your face.

quarnster: I stand corrected, just tried out a small test and it appears you are indeed correct. My assumption was that `test(*a)` conversion to a reference value would segfault, but it doesn't segfault until the second printf with both gcc and clang. Thanks for pushing back!

```c++
#include <stdio.h>

void test(const int &a) {
    printf("%s\n", (&a) == nullptr ? "true" : "false");
    printf("%d\n", a);
}

int main() {
    int *a = NULL;
    test(*a);
    return 0;
}
```

wwmayer: There is another point to consider. Assuming it's allowed to have a null reference then this must be considered valid code for all compilers and all compiler settings.

If you rewrite the code snippet to be closer to ModelProperty::operator== then it will be:
```
#include <stdio.h>

void test(const int &a) {
    if (&a == nullptr)
        return;
    printf("%d\n", a);
}

int main() {
    int *a = NULL;
    test(*a);
    return 0;
}

```

Now build & run the code with:
 `g++ main.cpp && ./a.out` -> works
 `g++ main.cpp -O0 && ./a.out` -> works
 `g++ main.cpp -O1 && ./a.out` -> crash

What happens is that the ` if (&a == nullptr)` is optimized away and `a` is accessed.

quarnster: > Undefined behaviour doesn't necessarily mean that the application crashes immediately or crashes at all.

That is a good point. That it happens to "work" is not reliable evidence that it should. Clearly the error is the illegal operation, and the check is just masquerading it.

davesrocketshop: At this point I'm willing to withdraw my objections and get this merged as is. It'll have to come out at some point anyways, and the sooner this is merged the sooner I can get going on some other stuff.

PR #11260, opened 2023-11-03
[Sketcher] Follow up on sketcher 'Tool parameters' PR.

PaddleStroke:
Follow up on #11174.
Make small improvements. Fix some small edge-cases bugs. For example if user enters 0 and 0 for length and width of a line, we prevent validation of an null line.
@abdullahtahiriyo

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
+++ b/src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
@@ -2308,7 +2308,7 @@ void CmdSketcherOffset::activated(int iMsg)
     ActivateHandler(getActiveGuiDocument(), new DrawSketchHandlerOffset(listOfGeoIds));
 }
 
-bool CmdSketcherOffset::isActive(void)
+bool CmdSketcherOffset::isActive()
 {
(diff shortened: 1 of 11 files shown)


0penBrain:
@PaddleStroke @abdullahtahiriyo the CI Lint tend to show there are some big issues with includes. Can you please check?

abdullahtahiriyo:
@0penBrain 

Yes. There is. Hopefully there were, as I have just merged what clang-tidy was complaining about in qtcreator for Sketcher Gui.

It is not only about this PR, and the previous one of tools, but also in other parts of Sketcher.

We should strive to have all necessary headers and forward declarations in place. This also allows us to leverage the static analysers to their full potential (e.g. dead code). It also makes explicit the true extent of dependencies introduced in each header, which is good to have a clear view of how the architecture evolves (and the need to take measures).

In practice, we should be wary of removing headers that may appear unnecessary for successfully compiling (because the compiler ends up resolving the symbols), but that should be there. I found clang-tidy to be rather good at it, and qtcreator to show me these issues, which I was not seeing locally before.

PR #11262, opened 2023-11-03
[Sketcher] ArcSlot DSH : implement the Arc slot tool

PaddleStroke:
Implement an arc slot tool for sketcher. It also has a methode to make rectangular slots.
@abdullahtahiriyo

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
@@ -1,0 +1,935 @@
+/***************************************************************************
+ *   Copyright (c) 2023 Ondsel <development@ondsel.com>                    *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This library is free software; you can redistribute it and/or         *
+ *   modify it under the terms of the GNU Library General Public           *
+ *   License as published by the Free Software Foundation; either          *
+ *   version 2 of the License, or (at your option) any later version.      *
+ *                                                                         *
+ *   This library  is distributed in the hope that it will be useful,      *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+
+#ifndef SKETCHERGUI_DrawSketchHandlerArcSlot_H
+#define SKETCHERGUI_DrawSketchHandlerArcSlot_H
+
+#include <QApplication>
+
+#include <Gui/BitmapFactory.h>
+#include <Gui/Notifications.h>
+#include <Gui/Command.h>
+#include <Gui/CommandT.h>
+
+#include <Mod/Sketcher/App/SketchObject.h>
+
+#include "DrawSketchDefaultWidgetController.h"
+#include "DrawSketchControllableHandler.h"
+
+#include "GeometryCreationMode.h"
+#include "Utils.h"
+
+namespace SketcherGui
+{
+
+extern GeometryCreationMode geometryCreationMode;  // defined in CommandCreateGeo.cpp
+
+class DrawSketchHandlerArcSlot;
+
+namespace ConstructionMethods
+{
+
+enum class ArcSlotConstructionMethod
+{
+    ArcSlot,
+    RectangleSlot,
+    End  // Must be the last one
+};
+
+}
+
+using DSHArcSlotController =
+    DrawSketchDefaultWidgetController<DrawSketchHandlerArcSlot,
+                                      StateMachines::FourSeekEnd,
+                                      /*PAutoConstraintSize =*/3,
+                                      /*OnViewParametersT =*/OnViewParameters<6, 6>,
+                                      /*WidgetParametersT =*/WidgetParameters<0, 0>,
+                                      /*WidgetCheckboxesT =*/WidgetCheckboxes<0, 0>,
+                                      /*WidgetComboboxesT =*/WidgetComboboxes<1, 1>,
+                                      ConstructionMethods::ArcSlotConstructionMethod,
+                                      /*bool PFirstComboboxIsConstructionMethod =*/true>;
+
+using DSHArcSlotControllerBase = DSHArcSlotController::ControllerBase;
+
+using DrawSketchHandlerArcSlotBase = DrawSketchControllableHandler<DSHArcSlotController>;
+
+class DrawSketchHandlerArcSlot: public DrawSketchHandlerArcSlotBase
+{
+    friend DSHArcSlotController;
+    friend DSHArcSlotControllerBase;
+
+public:
+    explicit DrawSketchHandlerArcSlot(ConstructionMethod constrMethod = ConstructionMethod::ArcSlot)
+        : DrawSketchHandlerArcSlotBase(constrMethod)
+        , startAngle(0.)
+        , startAngleBackup(0.)
+        , endAngle(0.)
+        , arcAngle(0.)
+    {}
+
+    ~DrawSketchHandlerArcSlot() override = default;
+
+private:
+    void updateDataAndDrawToPosition(Base::Vector2d onSketchPos) override
+    {
+        switch (state()) {
+            case SelectMode::SeekFirst: {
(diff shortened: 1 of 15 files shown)


src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
```diff
@@ -0,0 +1,921 @@
+/***************************************************************************
+ *   Copyright (c) 2022 Abdullah Tahiri <abdullah.tahiri.yo@gmail.com>     *
```
FlachyJoe, line 2: @PaddleStroke Why is this copyrighted to Abdullah and not to you ?

PaddleStroke:
Lint warnings to fix.

abdullahtahiriyo:
@PaddleStroke 

Do we go with this next?

PaddleStroke:
Yes. This or slot are both ok for me.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
```diff
@@ -0,0 +1,932 @@
...
+            createShape(false);
+
+            Gui::Command::openCommand(QT_TRANSLATE_NOOP("Command", "Add sketch arc slot"));
+
+            commandAddShapeGeometryAndConstraints();
+
+            Gui::Command::commitCommand();
+        }
+        catch (const Base::Exception& e) {
+            Gui::NotifyError(sketchgui,
```
abdullahtahiriyo, line 166: Lints due to missing headers.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
```diff
@@ -0,0 +1,932 @@
...
+    }
+
+    bool isWidgetVisible() const override
+    {
+        return true;
+    };
+
+    QPixmap getToolIcon() const override
+    {
+        return Gui::BitmapFactory().pixmap("Sketcher_CreateArcSlot");
```
abdullahtahiriyo, line 249: Headers

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
```diff
@@ -0,0 +1,932 @@
...
+    {
+        if (state() == SelectMode::SeekSecond || state() == SelectMode::SeekThird) {
+            setAngleSnapping(true, centerPoint);
+        }
+        else {
+            setAngleSnapping(false);
+        }
+    }
+
+    virtual void onButtonPressed(Base::Vector2d onSketchPos) override
```
abdullahtahiriyo, line 289: Little lint.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
```diff
@@ -0,0 +1,932 @@
...
+                    endAngle = startAngle;
+                    startAngle = startAngle + arcAngle;
+                    angleReversed = true;
+                }
+            }
+            this->moveToNextMode();
+        }
+    }
+
+    virtual void createShape(bool onlyeditoutline) override
```
abdullahtahiriyo, line 308: same.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
```diff
@@ -0,0 +1,932 @@
...
+                if (radius - r > Precision::Confusion()) {
+                    addArcToShapeGeometry(toVector3d(centerPoint),
+                                          startAngle,
+                                          endAngle,
+                                          radius - r,
+                                          isConstructionMode());
+                }
+            }
+            else {
+                auto arc1 = std::make_unique<Part::GeomArcOfCircle>();
```
abdullahtahiriyo, line 356: Ah! I see the fact that addArcToShapeGeometry does not return a reference is making your code more verbose. I will try to improve that.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
```diff
@@ -0,0 +1,932 @@
...
+                                                                 toVector3d(onSketchPos));
+        } break;
+        case SelectMode::SeekSecond: {
+            if (!onViewParameters[OnViewParameter::Third]->isSet) {
+                setOnViewParameterValue(OnViewParameter::Third, handler->radius);
+            }
+            double range = handler->startAngle * 180 / M_PI;
+            if (!onViewParameters[OnViewParameter::Fourth]->isSet) {
+                onViewParameters[OnViewParameter::Fourth]->setSpinboxValue(range,
+                                                                           Base::Unit::Angle);
```
abdullahtahiriyo, line 688: This needs to go via framework.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
```diff
@@ -0,0 +1,932 @@
...
+                               GeoElementId::HAxis,
+                               y0,
+                               obj);
+    };
+
+    auto constraintRadius = [&]() {
+        if (handler->constructionMethod()
+            == DrawSketchHandlerArcSlot::ConstructionMethod::ArcSlot) {
+            Gui::cmdAppObjectArgs(obj,
+                                  "addConstraint(Sketcher.Constraint('Distance',%d,%d,%d,%d,%f)) ",
```
abdullahtahiriyo, line 805: This is also headers.

abdullahtahiriyo:
It appears that cursors are also missing for this one.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
abdullahtahiriyo: In:
https://github.com/FreeCAD/FreeCAD/pull/11358 

The helper function is extended to return a reference. So you can call addArcToShapeGeometry and get as returned value a reference to the added arc. This you may use to modify the added arc, or to get endpoints. That way, you do not need to manually create the shape, which is less concise and error prone.

PaddleStroke: Great

PaddleStroke: Have you actually tried this? I cannot get it to work because you return a Part::Geometry unique ptr and I cannot find how to cast a unique ptr.

PaddleStroke: Besides if you look at arc2, I'm adding the arc2 after the 2 lines on purpose because sometimes arc2 doesn't exist. So using addArcToShapeGeometry here makes the code convoluted.

PaddleStroke:
This is finalized now.

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
abdullahtahiriyo: static_casting and dynamic_casting of unique_ptr you do via std::static_pointer_cast and std::dynamic_pointer_cast: 

https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast

So in your case, it appears you want something like:

```
auto geo = addArcToShapeGeometry(toVector3d(centerPoint),
                                                      startAngle,
                                                      endAngle,
                                                      r,
                                                      isConstructionMode());

auto arc2 = std::static_pointer_cast<Part::GeomArcOfCircle>(geo);
```

This syntax I can improve into a one liner in a subsequent PR.

I did not look in detail into the order pushing to the vector. If you want to add the arc at the end, then this will not work for you... unless you can add first the lines and then construct the arc afterwards.

If this facility allows you to write better and less cluttered code by reusing framework code (and thus will also be less error prone code), then you should use the facility. If the facility gets in the middle of your flow, then this specific instance may not be the best for this facility.

My intention was not to make your life miserable, but easier. If it is not easier (after understanding the static casting of smart pointers),

abdullahtahiriyo: I will push a new PR, in which you get a type specific naked pointer. This is probably the best solution. So if you ask for a Part::GeomArcOfCircle, it will create the arc, put it inside a smart pointer and add it to the ShapeGeometry array, and return to you a ```Part::GeomArcOfCircle *```, which you do not need to cast to anything. Of course, memory management and ownership is kept with the smart pointer, so you need not take care of any kind of memory management:

https://github.com/FreeCAD/FreeCAD/pull/11362

abdullahtahiriyo:
Arcslot suffers the same issue with assigning negative angle values as arc. The rest appears to work well.

PaddleStroke:
Fixed

src/Mod/Sketcher/Gui/DrawSketchHandlerArcSlot.h
PaddleStroke: > My intention was not to make your life miserable, but easier. If it is not easier (after understanding the static casting of smart pointers),

Yes I know that and it does help :)

PaddleStroke:
I have just force-pushed with the changes from https://github.com/FreeCAD/FreeCAD/pull/11362 for arc1

PR #11263, opened 2023-11-03
[Sketcher] Arc/Circle/Ellipse : implement tool parameters.

PaddleStroke:
Arc/Circle/Ellipse : implement tool parameters.
@abdullahtahiriyo

Requires merge of : 
https://github.com/FreeCAD/FreeCAD/pull/11294

### DIFF
--- a/src/Mod/Sketcher/Gui/CMakeLists.txt
+++ b/src/Mod/Sketcher/Gui/CMakeLists.txt
@@ -54,6 +54,7 @@ SET(SketcherGui_SRCS
     AutoConstraint.h
     Utils.h
     Utils.cpp
+    CircleEllipseConstructionMethod.h
     DrawSketchHandlerLine.h
     DrawSketchHandlerRectangle.h
     DrawSketchHandlerPolygon.h
--- a/src/Mod/Sketcher/Gui/CircleEllipseConstructionMethod.h
+++ b/src/Mod/Sketcher/Gui/CircleEllipseConstructionMethod.h
@@ -1,0 +1,40 @@
+/***************************************************************************
+ *   Copyright (c) 2022 Abdullah Tahiri <abdullah.tahiri.yo@gmail.com>     *
+ *                                                                         *
+ *   This file is part of the FreeCAD CAx development system.              *
+ *                                                                         *
+ *   This library is free software; you can redistribute it and/or         *
+ *   modify it under the terms of the GNU Library General Public           *
+ *   License as published by the Free Software Foundation; either          *
+ *   version 2 of the License, or (at your option) any later version.      *
+ *                                                                         *
+ *   This library  is distributed in the hope that it will be useful,      *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU Library General Public License for more details.                  *
+ *                                                                         *
+ *   You should have received a copy of the GNU Library General Public     *
+ *   License along with this library; see the file COPYING.LIB. If not,    *
+ *   write to the Free Software Foundation, Inc., 59 Temple Place,         *
+ *   Suite 330, Boston, MA  02111-1307, USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+
+#ifndef SKETCHERGUI_CircleEllipseConstructionMethod_H
+#define SKETCHERGUI_CircleEllipseConstructionMethod_H
+
+namespace SketcherGui::ConstructionMethods
+{
+
+enum class CircleEllipseConstructionMethod
+{
+    Center,
+    ThreeRim,
+    End  // Must be the last one
+};
+
+}  // namespace SketcherGui::ConstructionMethods
+
+
+#endif  // SKETCHERGUI_CircleEllipseConstructionMethod_H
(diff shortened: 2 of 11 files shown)


PaddleStroke:
Compilation problems on linux. I'll fix that later.

0penBrain:
> Compilation problems on linux. I'll fix that later.

Look at the Lint errors. This is more than a Linux problem. You have same issue in other of your open PRs.

PaddleStroke:
@abdullahtahiriyo I have been chasing a bug where starting the 3rim point circle tool would cause an illegal storage access. And I finally found where it comes from. Your split of DrawSketchDefaultWidgetHandler.
Now tools that don't have the same number of ovp between methodes cannot be initialized : 

                                      /*OnViewParametersT =*/OnViewParameters<3, 6>,

PaddleStroke:
Closed by misclick.

PaddleStroke:
There I found it.

`    int nOnViewParameter = OnViewParametersT::defaultMethodSize();`

And it is only set to the current construction methode in the onConstructionMethodChanged...

abdullahtahiriyo:
@PaddleStroke 

What is the status of this now? 

I see you have had some problems in the past (I have just seen it now). Are they solved or you still have specific issues I should look into?

PaddleStroke:
No there's no problèmes anymore. It just need to be rebased on latest
changes

abdullahtahiriyo:
@PaddleStroke 

Please consider cherry-picking this to fix lints:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/commit/e823644399b48c799330fd5260bb1fb0ab70cfe1

These DSHs are missing the indication of which OVP parameters are positional and dimensional and thus do not follow the settings in preferences. Please add the configuration.

Ellipse by rim segfaults here just when activating the tool. Please have a look.

PaddleStroke:
I think it is now ready.

abdullahtahiriyo:
The issues of yesterday as fixed and it works as expected.

There is a new issue though, that I did not see yesterday. Ellipse gives an option not to provide internal geometry. This should be removed.

Complex shapes without internal geometry are not fit for purpose in a sketcher. The original idea of allowing it was merely cosmetic to avoid seeing the geometry (rather than the geometry not existing).  The solution not to see it needs to come from a visualisation aid, such as visual layers.

During this development cycle, I plan to remove the possibility of deleting internal geometry individually (and thus creation of it is mandatory). This would also simplify the operations as the button to restore the geometry will be gone too (together with a migration aid to ensure that all internal geometry of any loaded sketch shows all the internal geometry).

On a different topic, this PR does not have the cursor text. I think it is advisable to add it beforehand, to avoid PR-ing the same tools over and over again, which will likely cause confusion as to which version people have.

Another detail is that the arc tool, when the OVPs are disabled, outputs a trillion "Spinbox doesn't exist in". This is because DSHArc is calling the setSpinboxValue directly, instead of via the framework. Please change those. 

You can grep for setSpinboxValue in the Gui directory:

```grep "setSpinboxValue" *```

then you will see that there are some in DSHEllipse, that also require changing.

For the rest, I think it is mergable.

PaddleStroke:
Ok then let's merge https://github.com/FreeCAD/FreeCAD/pull/11327 first and then I'll fix what needs to be fixed in this one.

abdullahtahiriyo:
Let me know when it is adapted, as it is already reviewed.

PaddleStroke:
I jsut pushed the fixes. Ellipse checkbox is removed. The setSpinboxValues that avoided my detection before are fixed. And cursor text implemented.

abdullahtahiriyo:
[Peek 2023-11-12 11-25.webm](https://github.com/FreeCAD/FreeCAD/assets/5249162/c8ff0d6e-6686-4d58-8b3a-16d0e1a7152e)

abdullahtahiriyo:
Got an issue trying to set the arc angle to a negative value. It will not take it.

PaddleStroke:
Ah it must unduly unset the OVP when angle is negative. It's probably a
mistake when I implemented such negative protection for things like radius.
I'll check this evening

abdullahtahiriyo:
[JUST A REMARK, NOT A BUG] 

Regarding the circle, I miss the possibility to provide a diameter value instead of a radius. I see different options, a new creation mode (M switchable), or an additional widget setting (checkbox, U/V switchable), or something for a long right-click contextual menu (code not yet in master, but in my branch).

I just want to listen to your opinion on this topic. It is not something required for this PR (unless you prefer that). It is something that is going to be requested.

abdullahtahiriyo:
[Peek 2023-11-12 11-37.webm](https://github.com/FreeCAD/FreeCAD/assets/5249162/e02d1d05-362c-4c64-b0ba-ed98ac719bfc)

abdullahtahiriyo:
Then, there is this oddity, that when a OVP gets a NaN value (rightly so), the OVP is drawn at the bottom/left of the 3D view, which is strange. I do not know if the solution is not to show the OVP (but then no numeric entering is possible, which does not appear right), or show it at an alternative position that is not so odd (e.g. next to the mouse?). This would preferably be part of the framework....

abdullahtahiriyo:
Ellipse rim - it does not have OVPs (dimensional only mode configuration), but only shows the cursor for the first point. It should show the cursor for second and third. Shouldn't it?

abdullahtahiriyo:
- The negative value arc, and the ellipse rim are bugs and should be corrected before merge.
- The other two is something we should address at some point. I let you decide following your preference.

PaddleStroke:
The 2 bugs you mentioned are fixed. 
For the circle, lets see what people ask for it.

abdullahtahiriyo:
@PaddleStroke 

Take a look at substitutions of degree/radian conversions:
https://github.com/FreeCAD/FreeCAD/pull/11376/files

```Base::toDegrees()``` and ```Base::toRadians()``` are not only more readable and less error prone, but prevent lint issues of magic numbers.

PaddleStroke:
Ah nice I didn't knew of these functions.

PR #11264, opened 2023-11-03
[Sketcher] Rotate DSH : Adds a polar pattern tool to sketcher

PaddleStroke:
Rotate DSH : Adds a polar pattern tool to sketcher
This requires https://github.com/FreeCAD/FreeCAD/pull/11260 to merge first.
@abdullahtahiriyo


PaddleStroke:
`                ShapeGeometry.push_back(std::move(std::unique_ptr<Part::Geometry>(geo)));
`
This is giving a Lint warning `moving a temporary object prevents copy elision [-Wpessimizing-move]`

I'm not sure to understand it. Should I change it to 
`                ShapeGeometry.push_back(std::unique_ptr<Part::Geometry>(geo));`
?

PaddleStroke:
@abdullahtahiriyo  This, scale and translate are also all ready to review with latest changes.

abdullahtahiriyo:
> Edit : this warning disappeared after rebasing...
> 
> ` ShapeGeometry.push_back(std::move(std::unique_ptr<Part::Geometry>(geo)));` This is giving a Lint warning `moving a temporary object prevents copy elision [-Wpessimizing-move]`
> 
> I'm not sure to understand it. Should I change it to ` ShapeGeometry.push_back(std::unique_ptr<Part::Geometry>(geo));` ?

When you have this:
```
Geometry mycoolfunction () {
   Geometry geo;
   ...
   return geo;
}

```
This code means: create a temporary object (geo) , then when returning make a copy using the copy constructor of geo.

However, well before move semantics were added, the compiler would optimize this. Essentially, at the point where this function is called the compiler will create the object to be returned, eliding the need for a copy. This is called "copy elision".

When move semantics were added, we all wanted to do this:

```
Geometry mycoolfunction () {
   Geometry geo;
   ...
   return std::move(geo);
}

```

thinking we were cleverer than the compiler. This move would prevent copy elision, because it is no longer a copy. But this is no more efficient than copy elision.

Your error is telling you that whatever you have inside std::move is a temporary, and that the compiler will gladly do copy elision, but that by using std::move, you are preventing this. So essentially, is asking you to delete the std::move.

abdullahtahiriyo:
When looking at all the code, you should have been putting that result of copy() directly into a smartpointer:
```
auto geo = std::unique_ptr<Part::Geometry>(Obj->getGeometry(geoId)->copy());
```
In normal objects, you would use ```make_unique```, as when passing a ``new``` to a ```std::unique_ptr``` it may leak. However, with this kind of factory method objects, your ```new``` is inside the copy(), so that above is the way it goes. You want to put that thing inside the smart pointer as soon as possible, to avoid leaks (for example if exceptions are raised before somebody responsible takes ownership and commits to delete it when the right time arrives).

PaddleStroke:
I just pushed to fix following your comments. If you can confirm if this is correct I will update scale and translate PR as well.

src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
```diff
@@ -0,0 +1,627 @@
...
+        for (int i = 1; i <= numberOfCopiesToMake; i++) {
+            for (auto& geoId : listOfGeoIds) {
+                auto geo = std::unique_ptr<Part::Geometry>(Obj->getGeometry(geoId)->copy());
+
+                bool isConstructionGeo =
+                    Sketcher::GeometryFacade::getConstruction(Obj->getGeometry(geoId));
+                Sketcher::GeometryFacade::setConstruction(geo.get(), isConstructionGeo);
+                double angle = individualAngle * i;
+
+                if (isCircle(*geo)) {
```
abdullahtahiriyo, line 229: There is nothing wrong in using this.

However, now we have the is template function, and I think we should move to use that more generic:
```geo->is<Part::GeomCircle>();```

You do not need to change it, it is more of an indication. I will migrate all calls eventually as a separate PR.

File: src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h [hunk 2]
```diff
@@ -0,0 +1,627 @@
...
+        }
+        else {
+            deleteOriginal = false;
+        }
+
+        individualAngle = totalAngle / numberOfCopiesToMake;
+
+        for (int i = 1; i <= numberOfCopiesToMake; i++) {
+            for (auto& geoId : listOfGeoIds) {
+                auto geo = std::unique_ptr<Part::Geometry>(Obj->getGeometry(geoId)->copy());
```
abdullahtahiriyo, line 222: I see.

Below you have quite some static_casts of smartpointer.get().

Smart pointers is a great way of managing ownership of the resource. Smart pointers should not get in the middle of what you are doing. Do not self-flagellate yourself.  If they do, simply:

```
auto geo= std::unique_ptr<Part::Geometry>(Obj->getGeometry(geoId)->copy());
auto geoptr = geo.get()
```

Then go happily ever after with geoptr, until where you need to transfer the ownership of the memory resource.

src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
```diff
@@ -0,0 +1,627 @@
...
+                centerPoint = onSketchPos;
+            } break;
+            case SelectMode::SeekSecond: {
+                length = (onSketchPos - centerPoint).Length();
+                startAngle = (onSketchPos - centerPoint).Angle();
+
+                startPoint = onSketchPos;
+
+                if (length > Precision::Confusion()) {
+                    std::vector<Part::Geometry*> geometriesToAdd;
```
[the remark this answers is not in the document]
PaddleStroke: It was for simplicity, ie not add a if(seekSecond) to the createShape. But it can be moved in createShape without problem.

src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
PaddleStroke: imo the syntax isCircle(*geo) is cleaner and shorter.
Best would be `geo->isCircle();`

abdullahtahiriyo:
I am not sure I fully understand how the DSH is intended to be used. There are two angles to be entered. The second angle cannot be negative. Is this intended?

PaddleStroke:
No the second angle should enable negative values. Let me fix that

abdullahtahiriyo:
If I try to create a polar array of a BSpline, the handler does not succeed:

[Peek 2023-11-13 18-46.webm](https://github.com/FreeCAD/FreeCAD/assets/5249162/f0b2ce19-5595-47af-8ac1-e8a954b9e6df)

PaddleStroke:
You select 2 angles for the case where you are not going to type the rotation angle basically. It let you choose a reference for the start and the end. In the case where you type the angle, the seek second is indeed not useful.

PaddleStroke:
Ah good catch for the bspline I noticed and forgot about it. I don't know what is wrong with them. I have not changed the code and in paddle_widget branch it was working.
And as you see the preview is actually drawn. So maybe something is missing in the definition of the bspline that prevent its addition to the sketch? Do you see something wrong/missing here : 

```
                else if (isBSplineCurve(*geo)) {
                    auto* bSpline = static_cast<Part::GeomBSplineCurve*>(geo);
                    std::vector<Base::Vector3d> poles = bSpline->getPoles();
                    for (size_t p = 0; p < poles.size(); p++) {
                        poles[p] = getRotatedPoint(poles[p], centerPoint, angle);
                    }
                    bSpline->setPoles(poles);
                }
```

PaddleStroke:
Arguably the 2nd OVP is useless as seeksecond is only for mouse workflow. So we could change 2nd so that it acts like the 3rd. So if you type a value it validates the tool directly.

abdullahtahiriyo:
> Ah good catch for the bspline I noticed and forgot about it. I don't know what is wrong with them. I have not changed the code and in paddle_widget branch it was working. And as you see the preview is actually drawn. So maybe something is missing in the definition of the bspline that prevent its addition to the sketch? Do you see something wrong/missing here :
> 
> ```
>                 else if (isBSplineCurve(*geo)) {
>                     auto* bSpline = static_cast<Part::GeomBSplineCurve*>(geo);
>                     std::vector<Base::Vector3d> poles = bSpline->getPoles();
>                     for (size_t p = 0; p < poles.size(); p++) {
>                         poles[p] = getRotatedPoint(poles[p], centerPoint, angle);
>                     }
>                     bSpline->setPoles(poles);
>                 }
> ```

I do not see something odd just by looking at the code. Some debugging is necessary. If you do not manage, I will look into it tomorrow.

abdullahtahiriyo:
> Arguably the 2nd OVP is useless as seeksecond is only for mouse workflow. So we could change 2nd so that it acts like the 3rd. So if you type a value it validates the tool directly.

I trust you in tool workflow. Being a new tool, I do not have a pre-existing expectation.

abdullahtahiriyo:
For the reference for constraint addition:
https://github.com/FreeCAD/FreeCAD/pull/11380

abdullahtahiriyo:
Rather recursive, isn't it?

``` auto geo = geo.get();```

**Comment by PaddleStroke** [comment]
Yes !
I'm still working on it. Somehow I cannot build anymore. Even reversing latest fix I can't build. I think it may be one of the latest commits in master of today. I need to check some more.

**Comment by abdullahtahiriyo** [comment]
Maybe it is the ondsel solver (assembly) commit, I had to disable building assembly to get pass cmake with the git submodule?

**Comment by PaddleStroke** [comment]
It appears that it is. I rollbacked and it built. I think there's something to do in git or cmake after this merge. I'll clear that up tomorrow

**Comment by PaddleStroke** [comment]
Ok so it builds and things are fixed. Also now the 2nd OVP will act as the third, so it will end the tool.

**Comment by abdullahtahiriyo** [comment]
I am not sure if we are synchronised. I built this commit:
1c0616901475911555831a843c66d3f0476fe10a

Then:
[Peek 2023-11-14 14-35.webm](https://github.com/FreeCAD/FreeCAD/assets/5249162/bcb2a255-aade-43f2-bdbf-f721b0accf0a)

1) U/J get into the copies quantity, which should not be the case (before clicking anything).
2) The OVP after the first click cannot be set. I tried several times to enter 0.
3) Then the DSH shows the copies (after having clicked), but when I enter the angle range, -45º, it errors (I get an error in the notification area).

**Comment by abdullahtahiriyo** [comment]
When using U/J after the first click, the spinbox shows decimals.

**Comment by PaddleStroke** [comment]
1. Yes. Its because when there is no OVP in seek first then the parameter has the focus. And so the keys U and J are caught by the parameter. The exact same happen with the polygon tool.
But after some thoughts I think this is actually the desired behavior actually. Because then you can type directly the number rather than having to use the shortcuts.
2. This is normal. Because with my latest change the 2nd OVP acts as the third OVP. ie it is the total angle of the rotation. And 0 is not an acceptable value for rotation. So if you type  50 it will rotate by 50. If you go to seekThird then type 50, the same happen. What is the point of seekthird then? Its because sometime you need to rotate not on a specific angle, but between reference points. In which case you select a start point and en and point.
3. I think the bug is specific to BSpline no? You have the same issue with other geos?
4. Yes decimals are annoying here. We also have them in polygon. I am not sure why they appear

**Comment by PaddleStroke** [comment]
So overall there is it seems only 1 issues its the bspline.
The decimals issue is a framework issue. It doesn't hurt functionality here.

**Comment by abdullahtahiriyo** [comment]
1. Isn't this because the parameter does not use DrawSketchKeyboardManager? If it used it, wouldn't the manager filter out shortcuts,as they are not numbers or - +? 

U/V are not valid characters for copies. Therefore, my expectation that they would be honoured. The current behaviour appears to me odd.

2. I will look at this tomorrow. If I understand now that in this version, when I enter 0º for the first, I am using a rotate tool to tell it I do not want to rotate. I do not fully understand a rotation between reference points (plural). The problem is not so much that I have problems to understand. I wonder if a regular user will not have them. Perhaps it is that I am too tired. It has been quite a long day... Tomorrow I look at this afresh.

3. I do not remember. As it was failing for the B-Spline, I tested the B-Spline.

4. It requires some debugging indeed.

**Comment by PaddleStroke** [comment]
Yes you are right, it would be best if U / J were still working even if spinbox has focus. And yes the widget parameter probably don't use the manager. I will try to fix it.

**Comment by PaddleStroke** [comment]
2. No. here is the workflow of the tool : 
1 - You know what angle you want to apply to the rotation. So after selecting the center of rotation (seek first) you are now in seekSecond. You can type the value of rotation in OVP2 (50deg for example). Press enter the geometries are created.
2 - You don't know what value of rotation you want. So first you select rotation center. Then seekSecond you select a startPoint which defines the start angle of the rotation. Then seekThird you select a endPoint that define the endAngle of the rotation. During seekThird you have OVP3 that does the same as OVP2. So if you changed your mind and know what angle you want you can type it in.

**Comment by PaddleStroke** [comment]
1. is fixed. I have changed keymanager from private to protected as it needs to be accessible from child class. I think here its ok but let me know if not.

**Comment by PaddleStroke** [comment]
I have pushed fixes for the decimals issue on U / J.
This fix is a workaround.

The real bug is that the setDecimals of QuantitySpinBox is disregarded after you QuantitySpinBox::setValue. I tried to fix it, but couldn't find a solution easily. @wwmayer I think you wrote this QuantitySpinBox so for you it might be much easier to spot what is wrong?

The issue we have is that we are setting the decimals of a quantitySpinBox to 0, to shows ints (number of side of polygon for example).
This works initially. ie after you call setDecimals the spinbox shows correctly (so say '6' for example).

But doing a quantitySpinBox::setValue after ignores the setDecimals that was set before and then the spinbox shows '7.00'.
The workaround I found is to setDecimals everytime after setValue.

**Comment by abdullahtahiriyo** [comment]
@PaddleStroke 

```void QuantitySpinBox::setValue(double value)
{
    Q_D(QuantitySpinBox);
    setValue(Base::Quantity(value, d->unit));
}```

A new quantity is created with the double passed to it. So the decimals setting is overriden.

Either calling setDecimals after each value set is intended, or the function above should copy the precision of the previous quantity to the new one. This decision should indeed be Werner's.

**Comment by abdullahtahiriyo** [comment]
@PaddleStroke 

I propose this solution:
https://github.com/FreeCAD/FreeCAD/pull/11407

I have asked Werner's review. If he is ok, he can merge without distracting him from anything else. If not he can tell.

**Comment by PaddleStroke** [comment]
> It is ok. The only thing that may happen, is that we get a lint because it is marked as protected and not private. But that I will eventually handle, as it is not the only one.

Linux CI build is complaining about keymanager no existing. Do you know why?

**Comment by abdullahtahiriyo** [comment]
> Linux CI build is complaining about keymanager not existing. Do you know why?

Yes. It is something I have been fixing in your code, because MSVC seess no problem. When you derive from a templated class, you need to add the type.

At the same time, I have realised that keymanager is a smart pointer. Then it should not be protected (not even by exception), as this leaks the resource ownership to derived classes. I have written an accessor to the underlying pointer for that.

I will be providing you with a commit with some linters and fixes soon for this PR. Then you can cherry-pick it and we move forward.

**Comment by abdullahtahiriyo** [comment]
Ok. So:
1. The B-Splines do not work. I will take a look later on.
2. SnapToAngle does not work. I guess it should. Shouldn't it? This you may take over.

With that we could merge. I will look into those open ends later this afternoon or tomorrow.

Some limitations of this tool:
1. It moves, but it does not modify existing geometry, but substitutes it with a newly created one. This leads to changes in the order of geometries in the geometry widget. The original is always deleted. One extreme is that if you have one arc with two tangent lines on its endpoints. Select the arc and rotate. The arc "moves", but the tangencies are gone.
2. It does not introduce a python counterpart. Being all done at the DSH, power users are not at luck here.

Actually limitation 1 is probably related to limitation 2. Lack of access to the properties of SketchObject is the main issue, I guess.

I have in my head an architecture extension to solve this in a simple way:
1. The code for generating the shape is moved to Sketcher. 
2. The handler uses this code with the calculated input parameters.
3. A Python command, with the parameters as input, executes the code generating the shape (which is now in Sketcher namespace, au lieu of SketcherGui).

With this, it should be possible to call the python command from the handler and independently from the command line, without much overhead. The c++ code creating the shape may be in a class friended by SketchObject, thereby allowing substitution or SketchObject may be extended with a facility to substitute one geometry with another.

But that is another battle (and PR).

**Comment by PaddleStroke** [comment]
1. Thanks
2. yes it should. I will fix it.

**Comment by abdullahtahiriyo** [comment]
The BSpline fails (as well as the ArcOfEllipse, ArcOfHyperbola, ArcOfParabola) because PythonConverter does not support them (see PythonConverter::process).

Take a look. If you can add them, fine. If not, let me know and I will do it when time allows (probably not tomorrow as I have very busy day).

In addition to that, for all complex shapes (including Ellipse, which is supported by PythonConverter), the current code will not work properly because internal geometry is not handled. As important, constraints on internal geometry are not handled. If one asks to rotate an Ellipse which has constrained majoraxis (for example), the rotated version shall also have an appropriately constrained majoraxis. This is a major issue because the result is unexpected behaviour.

For the rest. I have added a e.ReportException() inside the exception catching block. I had assumed that exceptions would generate from Python code... but they can also generate on PythonConverter. Then the exception needs to be reported. It is fine to call it because if an exception has already been reported, it will not report it again even if e.ReportException is called. With that, it is easier to understand what has gone wrong.

Please cherry-pick the last commit:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/tree/pw_rotate

with fixes.

So we have left:
1. Extend PythonConverter to support the unsupported geometries.
2. Modify this DSH so that it can handle complex shapes.

**Comment by PaddleStroke** [comment]
FIY I am currently doing :

>  Modify this DSH so that it can handle complex shapes.

**Comment by PaddleStroke** [comment]
I am currently adding arcofellipse to pythonConverter.

**Comment by PaddleStroke** [comment]
Added arcOfEllipse and arcOfHyperbola. Will now attempt parabola which should be ok. Then bspline I'm not sure

**Comment by PaddleStroke** [comment]
So now what is left to do : 
- AOH AOP : internal alignment constraints not working see in `PythonConverter::process(const Sketcher::Constraint* constraint)`
- Bspline : geos and alignment constraints.

**Comment by PaddleStroke** [comment]
AOH AOP are working good now. 
So now what is left to do :

BSpline: internal alignment constraints not working.

**Comment by PaddleStroke** [comment]
That's it it works properly for all geometry types and internal alignment.
So I think it is ready now.

**Comment by abdullahtahiriyo** [comment]
I am coming back to this one.

I have reordered and squashed the commits here:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/tree/pw_rotate

There I have also removed the commits setting the Decimal precision on every value change, as my PR fixing the root problem was merged in the meantime, and it is no longer necessary.

This is our current starting point. Please clone that branch for further work on this.

For complex shapes (bspline, ellipse, hyperbola, ...) internal geometry disappears if only the edge of the complex shape is selected. If all the internal geometry is selected as well, then it works fine. 

One possible fix for this is, for every complex shape, if the edge is selected, then the internal geometry is automatically selected by the tool, even if it was not originally selected. This would work for this tool. However, PythonConverter should be able to provide a complete shape, including internal geometry... I would argue that is the expected behaviour of PythonConverter after all.

However, if we apply the "good fix" for PythonConverter, namely that PythonConverter creates the internal geometry automatically, then you may run into problems with keeping track of the constraints that may have been applied to the original internal geometry.

In fact, PythonConverter should not probably create lines and constrain them manually, but should just add the complex shape edge, followed by a call to:

```
Gui::cmdAppObjectArgs(sketchgui->getObject(),
                                      "exposeInternalGeometry(%d)",
                                      GeoId);
```

This approach would also have saved you from having to add all internal geometry constraints. They were not "missing" by oversight, but rather the ellipse ones were coded because originally internal constraint addition was manual, but this approach was abandoned in favour of exposeInternalGeometry, which adds automatically to every shape what it needs.

I am unsure if you can move to this approach or not (PythonConverter+exposeInternalGeometry+fetch internal geometry of complex shapes to apply the constraints to them).

If after looking at it in more detail you cannot, then we might try to consider a mode for PythonConverter, so that it does not call the exposeInternalGeometry in one mode, and it does in a different mode (the default being to call it). The downside of this is having to maintain the constraint specific code for internalalignment geometry. However, you have already written it, and it is not like we get new complex shapes every day, so it could work.

The other finding of today's tests, is that moving a periodic B-Spline with its internal geometry causes FreeCAD to crash.

Please, would you take a look to these issues?

**Comment by PaddleStroke** [comment]
Doing : 
- In the DSH : filter out internal alignment geos / constraints.
- In the python converter add exposeInternalGeometry
Is going to be problematic for the indexes of constraints. As I create the constraints using firstCurve + listOfGeoId.size * n + i
So if additional geometries are added it is going to be tedious to manage.

What about this solution : 
- In the DSH : filter out internal alignment geos / constraints.
- Python converter does not change
- At the end I add a for loop that do exposeInternalGeometry on the created geometries that need it.

**Comment by abdullahtahiriyo** [comment]
When I saw the problem, the most proximate solution which came to me was "1- force select internal + run pythonconverter with special mode omit". The reason is that this leads to the lower amount of changes and the lesser amount of additional code to maintain. Initially I did not see through the optional second parameter. But with that, I think it is a better option. This is my preferred option, unless you see an obvious disadvantage. Yes, I am going to prevent removal of individual internal geometry in this dev cycle (at least I plan to).

I think 2 is overcomplicated. It requires to reimplement here all the internal geometry rules for all complex shapes. It is more code and more difficult to maintain. It is probably also going to be something to repeat in the other transformation DSHs.

With respect to merging the 3 DSHs in one or not, you need to see it through, with the advantages and disadvantages. I thought it would be possible to write an "skeleton" of a transformation, and then "just" pass it the actual transformation. This would also make further transformations easier to implement. More code reuse (common parts) = less code to maintain. But you need to see it through, if you do not, then we can keep them separated. Treat it just as an observation.

**Comment by PaddleStroke** [comment]
@abdullahtahiriyo I have forgotten to rebase on your version of the branch. I will do that tomorrow. 
Meanwhile you can check my latest commit.
It does 2 things : 
1 - Make getListOfSelectedGeoIds function that force the selection of internal geos. Btw please look I have added : 

```
                // In case user deleted internal geometries we restore them.
                Gui::cmdAppObjectArgs(Obj, "exposeInternalGeometry(%d)", geoId);
```
To make sure that even if the user has deleted the internal geos, then we recreate them. What do you think about this?

2 - Add a local convenience getSketchObject function. It is funny because I added this one because I wanted to use this and wanted to reduce copy paste. But after applying this function all over the file, I realized that I actually did not need this function. So this is unrelated to this PR but it still reduce code duplicity by a lot. So I thought it was better not to revert this.

**Comment by PaddleStroke** [comment]
- Periodic bspline bug is fixed.
- Internal geos selection is forced.
- Mode is added to python converter.

**Comment by abdullahtahiriyo** [comment]
> 2 - Add a local convenience getSketchObject function. It is funny because I added this one because I wanted to use this and wanted to reduce copy paste. But after applying this function all over the file, I realized that I actually did not need this function. So this is unrelated to this PR but it still reduce code duplicity by a lot. So I thought it was better not to revert this.

I am ok with that.

**Comment by abdullahtahiriyo** [comment]
@PaddleStroke 

As unfortunate as it may sound, all the review messages were not published.

Now we have this situation: 

Some comments that are published now, I fixed them myself because you were not reacting (and I understand now, you were not reacting because you could not see them). 

Take a look. What does not make sense, just mark as resolved...

**Comment by PaddleStroke** [comment]
Ah I see. If you don't get a reply from me don't hesitate to ping me again in the future.
I will look through them all and fix what still needs to be fixed.

File: src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h [hunk 4]
```diff
@@ -257,20 +259,24 @@ class DrawSketchHandlerRotate: public DrawSketchHandlerRotateBase
...
-                auto geoUniquePtr =
-                    std::unique_ptr<Part::Geometry>(Obj->getGeometry(geoId)->copy());
+                const Part::Geometry* pGeo = Obj->getGeometry(geoId);
+                auto geoUniquePtr = std::unique_ptr<Part::Geometry>(pGeo->copy());
                 Part::Geometry* geo = geoUniquePtr.get();
 
-                bool isConstructionGeo =
-                    Sketcher::GeometryFacade::getConstruction(Obj->getGeometry(geoId));
-                Sketcher::GeometryFacade::setConstruction(geo, isConstructionGeo);
+                bool isConstructionGeo = GeometryFacade::getConstruction(pGeo);
```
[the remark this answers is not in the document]
PaddleStroke: This can be removed indeed. I think I have put this at a point where it was necessary a long time ago.

src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
```diff
@@ -0,0 +1,632 @@
...
+                        }
+                        else {
+                            newConstr->Second = secondIndexi;
+                        }
+                    }
+                    else if ((cstr->Type == Block) && firstIndex >= 0) {
+                        newConstr->First = firstIndexi;
+                    }
+                    else {
+                        continue;
```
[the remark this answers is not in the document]
PaddleStroke: It's actually quite complex to handle as you describe. I was trying to do it, and it's becoming too complex.
Because there are actually many special cases where we don't want to copy the constraint. For example perpendicular between a line and an axis. Which is why there are test `&& firstIndex >= 0 && secondIndex >= 0`

So testing before will duplicate many things.

PaddleStroke:
I have pushed fixes for the things you commented on.

src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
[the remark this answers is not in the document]
abdullahtahiriyo: I realised that later on. That is why I did not pursue it and move on. I am satisfied that we understand that we do this by way of exception.

abdullahtahiriyo:
Code is acceptable. However, testing has revealed one major bug (segfault) that needs to be fixed before merge.

To reproduce the crash:
1. Create one complex geometry (ellipse, B-Spline,...)
2. Create any other geometry (can be complex or not, can be a circle, can be a point)
3. Create one point that will be the rotation point
4. Select (1) and (2), with the intention to rotate them over (3), and trigger the tool: SEGFAULT.

I would appreciate if you moved the shortcut, currently "B", to a secondary shortcut of tools group "Z, []". Because now there is no longer the limitation that if we had "Z, P, 1", we could not have "Z, P" (from polar).  But you may choose another. We set the shortcuts to be easy to remember when using English language. 

Given that the tool moves by rotation, but also creates polar patterns, I would rather call it "Polar transform". Take this merely as user feedback. I think "Rotate" may not reflect very well all the functionality that the tool offers and may affect discoverability. If you move in this direction of change, the the icon perhaps does not reflect it very well either.

If we ever move to split toolbars, a transforms toolbar could be created for the different transformation DSHs in the pipeline. This is just a comment for the future.

PaddleStroke:
I will check the bug.
Ok for the shortcut and for the name.

PaddleStroke:
I unfortunately cannot reproduce the crash you are having. Am I doing the same as you?

Untitled

PaddleStroke:
Sorry wrong file in previous post. Here : 
I unfortunately cannot reproduce the crash you are having. Am I doing the same as you?


https://github.com/FreeCAD/FreeCAD/assets/19984177/396cb87e-1f51-4f35-9038-c981367b43cb

abdullahtahiriyo:
@PaddleStroke 

For some reason, I cannot see your video "No video with supported format and MIME type found".

So, I have pulled your latest branch state and I can still reproduce it:
[Peek 2023-12-17 14-48.webm](https://github.com/FreeCAD/FreeCAD/assets/5249162/ea2ba701-99a2-4dd1-8131-76d425994350)

```
#1  0x7f630b0cca61 in bool Base::BaseClass::is<Part::GeomEllipse>() const from /home/abdullah/github/freecad-build/Mod/Sketcher/SketcherGui.so+0x2b
#2  0x7f630b0c866f in Sketcher::isEllipse(Part::Geometry const&) from /home/abdullah/github/freecad-build/Mod/Sketcher/SketcherGui.so+0x1c
#3  0x7f630b1f922f in getListOfSelectedGeoIds(bool) from /home/abdullah/github/freecad-build/Mod/Sketcher/SketcherGui.so+0x52b
#4  0x7f630b203100 in CmdSketcherRotate::activated(int) from /home/abdullah/github/freecad-build/Mod/Sketcher/SketcherGui.so+0x34
```

Ok. I have tracked down the issue. I mark it in the code and explain why this happens...

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -51,13 +51,92 @@
...
+                    if (geoId >= 0) {
+                        listOfGeoIds.push_back(geoId);
+                    }
+                }
+            }
+        }
+    }
+
+    if (forceInternalSelection) {
+        for (auto geoId : listOfGeoIds) {
```
abdullahtahiriyo, line 107: (A)

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -51,13 +51,92 @@
...
+            const Part::Geometry* geo = Obj->getGeometry(geoId);
+            if (isEllipse(*geo) || isArcOfEllipse(*geo) || isArcOfHyperbola(*geo) || isArcOfParabola(*geo) || isBSplineCurve(*geo)) {
+                const std::vector<Sketcher::Constraint*>& constraints = Obj->Constraints.getValues();
+                for (auto constr : constraints) {
+                    if (constr->Type == InternalAlignment && constr->Second == geoId) {
+                        if (std::find(listOfGeoIds.begin(), listOfGeoIds.end(), constr->First) == listOfGeoIds.end()) {
+                            // If the value is not found, add it to the vector
+                            listOfGeoIds.push_back(constr->First);
```
abdullahtahiriyo, line 115: I have marked (A) above.

So we are iterating vector ```listOfGeoIds```, this line here is adding new elements to it.

A for range construction as (A) uses iterators internally. When you add one element to the vector, the iterators are invalidated. Then you may be accessing another area (it is undefined behaviour). Here, geoId, on the second iteration of the loop (after adding the internal geometry to the array) gets value 22354.

If you abstract for a moment from the specific case, think that an vector is allocated a consecutive chunk of memory. When you add elements, the vector itself may need to grow. When it needs to grow, it may be reallocated elsewhere. Therefore, there is no guarantee your iterators continue to be valid.

Some implementations (I guess MSVC) may disregard the issue relying on no reallocation, which often may be true, but it is a serious issue.

The easiest way is to drop use of iterators. For example, remove the ```range for``` (A) and substitute it with a traditional indexed loop ```for( i;i<loopsize;i++)``. Here loopsize is the initial vector size (as you do not want to loop over the added internal geometry at the end of the array).

An alternative is to add the internalgeometry to another vector and after the loop merge them.

I am no allergic to traditional indexed loops when the use case is clear, so I would do the first. But if you want to do it the other way around, fine with me.

PaddleStroke: Great thanks ! 
I will update asap

maxwxyz:
@PaddleStroke 
The field with the number of copies is not displaying an integer but uses the number of decimal places defined in the overall settings.
Also the tooltip is displayed as a long line, not sure if this is intended but I have never seen such a wide tooltip in FreeCAD:
grafik

PaddleStroke:
Hi @maxwxyz
Thanks for the report.
- the U/J was writen to indicate that you can use the U and J keys to increase/decrease the number of copies. But you are the second person who was confused by this so it must not be clear. Perhaps we need to remove this from the label and write this in the tooltip.
- The long tooltip needs to be fixed. I think just adding <p> </p> tags will solve this.
- Regarding the decimals, this is annoying, I thought it was fixed before. I will need to check this.

If you feel like it, perhaps you want to tackle the U/J and long tooltip in a PR? It would be a good training if you want to increase your experience in PR. The third point is probably going to be very annoying to fix so you can let it to me.

maxwxyz:
ok I will tackle the tooltip tomorrow, I guess the U/J just needs an explanation, maybe some brackets like in the rectangle and line tools. I see that other keys are not working as other commands or constraints are then activated, (up/down or o/p). Maybe it is best to disable the hotkeys when the tool is active as if you want to type in references or expressions in the OVP (I know currently not supported) you would cancel the current OVP tool and activate another tool by accident.

PaddleStroke:
Great !
Regarding the problem you mention, we already have it with the unit. Say you want to type in '34 cm'. Abdullah made a clever solution for this. If you are typing a number, then the letters typed before 2 seconds go to the spinbox. Else letter go to the view and do whatever (potentially trigger another tool).
But if we want to support setting a expression in the OVP then it needs to be improved more. Maybe starting with = would just open the expression editor. And the expression editor take the focus so it would be alright.

Roy-043:
There is something confusing and something wrong with the Copies spinbox in the task panel of the Sketcher_Rotate command.

Confusing:
Initially the spinbox shows zero in italics. Using the U key increases the number and also turns the text bold (no italics). Using J to go down to zero results in a bold text (no italics). Is this intentional? IMO this is unnecessary and confusing.

Wrong:
If you use the spin buttons the number is displayed as a float (with decimals).

maxwxyz:
@Roy-043  --> https://github.com/FreeCAD/FreeCAD/issues/11892

PaddleStroke:
Confusing : This was the original design of the spinboxes in the widget. The intent was to show when a value was set.
However for the number of copy it does not make sense, as even if you don't set a value the spinbox value is used.
I think the widget spinboxes are now only used for such parameters (polygon number of sides, number of copies etc.) so this default behavior can be dropped.

Wrong : This is true. It is however suprisingly annoying to fix. I don't recall exactly for the case of the spin buttons, if I already tried to fix it or if I fixed other cases (U, J buttons, and manually typing were also displaying decimals at some point.)
I will need to check this out. Can you please create an issue?

Roy-043:
Thanks. I see that @maxwxyz has already mentioned the float problem in his Issue.

maxwxyz:
@Roy-043 the decimal place issue of the rotate and polygon tool is not present anymore in main at my side, can you confirm?

Roy-043:
It is still present in the last weekly. Note that only using the spin buttons show the problem.
```
OS: Windows 8 build 9600
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35554 (Git)
Build type: Release
Branch: main
Hash: 8e2ab9f76888ac4da971f9d5069cec7bcc2b572f
Installed mods: 
```

maxwxyz:
OK, can reproduce. Only when using spin buttons or scrolling inside the text field.

Issue #8509, opened 2023-02-16
Sketcher: Solver Failures


louisrousseau:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.0.31847 (Git)
Build type: Release
Branch: master
Hash: 502ef8c9ba619671b5137833851a7ebbbab24bf0
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: French/Canada (fr_CA)
```

### Subproject(s) affected?

Sketcher

### Issue description

Apologies, this bug report is going to be messy on account of ongoing trial and error when these messages occurred and not noticing them immediately.

Got two different messages saying to report the issue while trying to add Spreadsheet and Expression based scaling to my model:

`Important: the LevenbergMarquardt solver succeeded where the DogLeg solver had failed.`
and
`Important: the SQP solver succeeded where all single subsystem solvers have failed.`

Initially had failed recomputes with little else to go on - the "Report Window" is unfortunately quiet on these things - started suspecting the way my constraints are set might break on a diagonal once it starts using scaled values.

I removed one length constraint on diagonal `18-Line` and replaced it with a vertical one using expression `sqrt(pow(<<Dimensions>>.DiagonaleEntreNiveaux; 2) / 2)` instead.

I had inserted rows and renamed aliases along the way and it looked like the rename did not propagate correctly, which broke some constraint references whom were no longer following changes in my scaling factor, instead trying to get the original unscaled value in column `C`.

I tried scaling factors of 0.5 and 2, going back to one with a successful solve in between.

To see my sketch scale I would sometimes need to double-click on a constraint and cancel out of that dialog.

See attached for a version of the file taken while writing this report.

### Anything else?

[T-RV-7.zip](https://github.com/FreeCAD/FreeCAD/files/10752477/T-RV-7.zip)

Edit:  
> Also related: https://github.com/FreeCAD/FreeCAD/issues/21553. A different jump likely happens there with the same effect.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

louisrousseau:
Quick update, there were all kinds of problems I hadn't noticed with my constraints earlier: coincidents not on the correct vertices, overlapping lines, etc.

[...]

PR #11265, opened 2023-11-03
[Sketcher] Scale DSH : adds a tool to scale geometries.

PaddleStroke:
Scale DSH : adds a tool to scale geometries.

@abdullahtahiriyo


abdullahtahiriyo:
@PaddleStroke 

From the rotate DSH where do you want to go next?

Please point me to the one you want and give it a code read in view of the changes made to the Rotate DSH, that would help me reduce the iterations...

PaddleStroke:
Scale is ok to continue.
I have applied all the fixed that we implemented in the rotate. I may have
missed something though.

abdullahtahiriyo:
So do we undraft this one?

PaddleStroke:
Ah yes. The shortcut needs some attention btw as it's not ideal. If you have a better idea let me know.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
```diff
@@ -2404,6 +2405,39 @@ bool CmdSketcherRotate::isActive()
...
+    : Command("Sketcher_Scale")
+{
+    sAppModule = "Sketcher";
+    sGroup = "Sketcher";
+    sMenuText = QT_TR_NOOP("Scale transform");
+    sToolTipText = QT_TR_NOOP("Scale selected geometries. After selecting the center point you can either enter the scale factor, or select two reference points then scale factor = length(p2-center) / length(p1-center).");
+    sWhatsThis = "Sketcher_Scale";
+    sStatusTip = sToolTipText;
+    sPixmap = "Sketcher_Scale";
+    sAccel = "Z, P, S";
```
abdullahtahiriyo, line 2422: I think we will have to do an exercise on the shortcuts when we merge all tools. 

Now Offset has "Z, T". Perhaps we could reserve Z, T for the "T"ransforms, so this could be Z, T, S. I am not a fan of 3 letter shortcuts though. It will require some thinking.

src/Mod/Sketcher/Gui/DrawSketchHandlerScale.h
```diff
@@ -0,0 +1,529 @@
...
+
+    ~DrawSketchHandlerScale() override = default;
+
+private:
+    void updateDataAndDrawToPosition(Base::Vector2d onSketchPos) override
+    {
+        switch (state()) {
+            case SelectMode::SeekFirst: {
+                referencePoint = onSketchPos;
+                referencePoint = onSketchPos;
```
abdullahtahiriyo, line 92: Once was not enough?

src/Mod/Sketcher/Gui/DrawSketchHandlerScale.h
```diff
@@ -0,0 +1,529 @@
...
+            else if (isArcOfHyperbola(*geo)) {
+                auto* arcOfHyperbola = static_cast<Part::GeomArcOfHyperbola*>(geo);  // NOLINT
+                arcOfHyperbola->setMajorRadius(arcOfHyperbola->getMajorRadius() * scaleFactor);
+                arcOfHyperbola->setMinorRadius(arcOfHyperbola->getMinorRadius() * scaleFactor);
+                arcOfHyperbola->setCenter(
+                    getScaledPoint(arcOfHyperbola->getCenter(), referencePoint, scaleFactor));
+            }
+            else if (isArcOfParabola(*geo)) {
+                auto* arcOfParabola = static_cast<Part::GeomArcOfParabola*>(geo);  // NOLINT
+                // TODO: Problem with scale parabola end points.
```
abdullahtahiriyo, line 288: Problem?

src/Mod/Sketcher/Gui/DrawSketchHandlerScale.h
```diff
@@ -0,0 +1,529 @@
...
+                else {
+                    continue;
+                }
+
+                ShapeConstraints.push_back(std::move(newConstr));
+            }
+        }
+    }
+
+    int indexInVec(const std::vector<int>& vec, int elem) const
```
abdullahtahiriyo, line 367: This kind of function should be moved out to tools for code reuse, as it is not specific of this DSH. Others DSH should use it (I think I remember Rotate also uses it).

Then, I would say it is something more than a generic "indexInVec" as it is for a std::vector<int> which has inside GeoIds. The name of the function should reflect this "IndexOfGeoId" (for example). Then the parameter "element" should perhaps be "geoid".

That would also hint why this function is being written at all (because it handles special values differently), as just getting the index of a vector is a common operation, which is generally implemented via std::find and std::distance.

abdullahtahiriyo:
On the functionality:
1. It is a little bit odd to have two degrees of freedom to move the mouse during scaling (to set the scale), and one DoF bringing nothing to the table. It appears as if one could expect the tool to rotate (which is not what is intended).
2. I get a failure to converge.

I wanted to show you a video, but my internet connection is very unstable and is refusing to upload it.

Try this:
1. Create a B-Spline.
2. Try to scale it.

I also get the same issue with a periodic B-Spline.

abdullahtahiriyo:
[Peek 2023-12-21 20-50.webm](https://github.com/FreeCAD/FreeCAD/assets/5249162/838a545c-5ca5-401e-b291-5c67d5d9f67e)

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
PaddleStroke: Perhaps we can drop the Z and use T for transform? So just T, S.
Not sure T is available though.

src/Mod/Sketcher/Gui/DrawSketchHandlerScale.h
PaddleStroke: Sounds good, but where should I put this function so that it is shared? In Utils.cpp?

PaddleStroke:
1. Yes initially it may feel a bit weird. But it may be helpful in cases where you want to scale based on reference points. As the points are maybe not aligned.
2. I will investigate this issue and the other comments as well.

src/Mod/Sketcher/Gui/CommandSketcherTools.cpp
abdullahtahiriyo: 'T' is tangent constraint.

Originally, we had a limitation that 'T' would take away all the shortcuts starting with 'T', so no 'T, S'. Then tools would start with 'Z'. 

People using shortcuts have this hardwired in their minds, meaning that changing it is somewhere between "I will let you changing just once" and "Over my dead body".

Now, we have a timer based capturing, where 'T' and 'T, S' is possible. While we have it and it helps deconflict shorcuts, there is still the risk that a too slow 'T, S' causes a tangency to be executed, which is annoying.

Yet, with the introduction of new tools, there may be changes to current ones, so some now taken shortcuts may become free.

Then, even if 'T' and 'T, S' were ok, if we take 'T' for 'Transform', we should look which transforms we have before assigning letters. It is likely that we consider 'Symmetry' a transform, it would be odd to have one "transform" as "Z, S" and another at 'T, S'.

That is why I think it makes sense to delay the exercise.

With this in mind, an option would be not to assign a shortcut to this tool, or any other tool we have introduced lately, so that users do not get used to a shortcut that we know we will likely change.

src/Mod/Sketcher/Gui/DrawSketchHandlerScale.h
abdullahtahiriyo: That is what I would do.

If you take encapsulation into account, you do not need to access class members, which tells us it should not be a member function of the class or a static member function, so it should be a "free" function outside the class. If we want code reuse it needs to be declared in a header you can include in both places. The header of choice for this kind of functions is now Utils.h.

PaddleStroke:
So this failed to converge happen not all the time (1/5 in average). And it can be fixed by simply clicking on recompute (usually once, twice or more)

After a lot of testing it seems that this 'failed to converge' comes from the spline 'middle' point : 
image

I think when this point is getting scaled, the calculated position may not match the bspline position exactly. And so perhaps the solver is having trouble with this. @AjinkyaDahale any idea what can be the problem? I think you may be the one who implemented those middle points so you may know more about the subject.

AjinkyaDahale:
@PaddleStroke could you share the file?

PaddleStroke:
The file does not seem important, any bspline act similarly. Just draw one
similar to mine.
It does not always happen so maybe you need to scale 5-10 times to meet it.

AjinkyaDahale:
Problem might only be manifesting in B-splines but present everywhere. I encountered a similar issue in #8509. @PaddleStroke could you check against that file first? Just scale that sketch, possibly removing the spreadsheet associations. There the problem is that only a few variables are scaled, and the solver is left to fill in the blanks. Which it does. Very wrongly.

PaddleStroke:
In the file there are 2 sketches. I can scale the geometries within 'SketchBaril' with no issues the other sketch cannot be opened.

AjinkyaDahale:
> In the file there are 2 sketches. I can scale the geometries within 'SketchBaril' with no issues the other sketch cannot be opened.

What error do you see in "cannot be opened"? I'm surprised this is a thing.

PaddleStroke:
image

AjinkyaDahale:
> So this failed to converge happen not all the time (1/5 in average). And it can be fixed by simply clicking on recompute (usually once, twice or more)
> 
> After a lot of testing it seems that this 'failed to converge' comes from the spline 'middle' point as if it is deleted the failure does not happen...
> 
> I think when this point is getting scaled, the calculated position may not match the bspline position exactly. And so perhaps the solver is having trouble with this. @AjinkyaDahale any idea what can be the problem? I think you may be the one who implemented those middle points so you may know more about the subject.

The point (the knot point) is definitely internal geometry and shouldn't have any particular scaling issues so long as all the control points are moved as well. The point is always a linear combination of those points.

AjinkyaDahale:
> In the file there are 2 sketches. I can scale the geometries within 'SketchBaril' with no issues the other sketch cannot be opened.

Could you give the exact steps to reproduce this issue? I just opened both sketches fine. But since they are fully constrained I can't try scaling without deleting a few constraints. Could you tell what constraints you deleted?

PaddleStroke:
I could replicate this issue in both case : If you select only the bspline and if you select the bspline + all its internal geos.
Actually the scale tool force the selection of all internal geometries so it's normal.

This does not happen very often (maybe 1 time every 5 to 10). But what I found so far is that if you delete this middle knot point before attempting to scale, then it never happens.

To increase the change of the bug happening, you can create multiple bspline (~10) and try to scale them all at the same time.

PaddleStroke:
You can try to scale directly. The tool will delete constraints that would
prevent scaling.

I have just opened the file, then double clicked on the sketch to open it.

chennes:
@PaddleStroke can you clarify the status on this one? Are you still working on a bug?

PaddleStroke:
@chennes the solver sometimes fails to converge when scaling a bspline. Which solves itself after hitting the refresh button.

I think this bug is more on the solver side than on the scale side, as it seems nothing is wrong with the tool. The points are scaled correctly. I think the bug happens because the scaled knot point is sometimes not fitting exactly right on the bspline and so the solver fails. But I'm not sure. Which is why I asked @AjinkyaDahale if he can check.

I would say that this can still go in, but that depends on what you prefer. Perhaps we can wait the feedback of Ajinkya first

AjinkyaDahale:
@PaddleStroke I have still not been able to replicate the issue. Just tried with around 5-6 splines and multiple times but still no issue.

src/Mod/Sketcher/Gui/DrawSketchHandlerScale.h
chennes: @PaddleStroke are we going to leave this for a later PR to resolve?

src/Mod/Sketcher/Gui/DrawSketchHandlerScale.h
PaddleStroke: Ah I forgot about these, I will fix them.

src/Mod/Sketcher/Gui/DrawSketchHandlerScale.h
PaddleStroke: This is now fixed

PaddleStroke:
@FEA-eng thanks, but the bug is gone :)

PaddleStroke:
Thanks ! I'll keep that in mind. If you can compile and give it a go, it can be interesting to give 11267 some testing. We checked it today, but we may have missed something.

Issue #11060, opened 2023-10-16
Statusbar selection of Project unit system does not work


Roy-043:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

While trying out the new Project unit system I notice a glitch.

* Delete/rename your cfg file.
* Start FreeCAD.
* Create a new document.
* In the statusbar select: "Building Euro".
* switch to Part.
* Create a Box.
* Check the dimensions in the Tree view.
* Length: 1,00 cm (OK).
* Open the Preferences Editor.
* "Unit system" is "Building Euro".
* "Project unit system" is "Standard" (Not correct).
* Change the Project unit system to "Imperial decimal"
* Result: the statusbar and the Tree view show inches.
* In the statusbar select: "Metric small parts".
* Open the Preferences Editor.
* "Unit system" is "Metric small parts".
* "Project unit system" is "Imperial decimal" (Not correct).

Expected behavior: the statusbar selector should only be linked to the Project unit system.

Related to #9521.

### Full version info

```
OS: Windows 8 build 9600
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34667 (Git)
Build type: Release
Branch: main
Hash: af5b8714a8371162addb97dd25c29250fcecf7ad
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: Dutch/Netherlands (nl_NL)
Installed mods:
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

wwmayer:
> While trying out the new Project unit system I notice a glitch.

It's not really a bug because it was never implemented. The current workflow is a bit awkward because you have to change this in the preferences after having created a new document. It would be more logical to move this setting to the _Project Information_  dialog and for quick access to the status bar.

bgbsww:
I'm playing with some code here, and I've realized that the goal isn't perfectly clear, so here's a proposal:

-  A Gui field in _Project Information_  with a pulldown to specify the unit system schema.
- Value of this field is stored with the document,
- The Status Bar Dimension widget at the bottom of the main window both reflects and updates this setting.
- When a new document is created, this value is initialized with the Preferences -> General -> Unit System and decimals.
- The Preferences -> General -> Project Unit System and ignore options are removed.

[...]

PR #11266, opened 2023-11-03
Move the unit schema into Project Information and remove all Project Unit System code

bgbsww:
A possible solution for #11060.  This adds a unit system to the document, exposes it in the Project Information and the Dimensions component of the Status Bar, and initializes it from the value in the Preferences.  Changes in either the Project Information or Dimensions component are supported.  When switching between documents, both locations are updated.

Changes to the Dimensions component no longer affect the Preferences value.


luzpaz:
```Warning: src/Gui/Application.cpp:936:        <-- trailing whitespace```

chennes:
@obelisk79 Can you ask the UI/UX working group to give this a look? Also @AgCaliva.

obelisk79 (approved):
These changes appear to improve unit management from a UX standpoint and are in-line with what should be predictable behavior.

AgCaliva:
I have suggested that maybe it would be better to use as initial value the ProjectUnitSystem setting, then if user sets the status bar it donts changes the ProjectUnitSystem value, but it is saved in a metadata file in something like .local folder( user metadata ). That file would be only related to one project. The number of files of metadata from each project should be stated in somewhere settings so it can be set to delete the oldests metadata files, example maximum project metadata files = 20 then only 20 settings will be saved and if that number its surpassed then you the first created.
My argument to think like that its that some people work in the same project at the same time and they probably use some version control system, then if we set the ProjectUnitSystem in the projects file each time the bar its switched, then the ProjectUnitSystem decided by the team will probably be lost at some point, and i think ProjectUnitSystem should be something more static. If you want to work in other unit system then you save your config in your pc, and not in the project.
Thats what i think at least by now, Also we should remember that it will not be so needed since we will have per-constraint unit system later.

sliptonic:
I built and tested this and it seems to be a step backwards or perhaps I don't understand.

Changing the schema in the status bar, changes the property in the document but doesn't seem to change it in UI.

For example, I start a new document.  The schema is set to metric/small parts/CNC.
The status bar reads the same.
I switch the status bar schema to US customary.  
The document property is US Customary.
The status bar still shows units in metric.  I click on the status bar selector and it shows US customary as selected in the list.
I close and reopen the document and the units are now shown in US customary.

In short, while in a document, there's no way to switch the displayed units.

If the user switches the unit schema from the document properties, it can get even more confusing.  I was able to set the document property to FEM, the selected item in the list still showed US customary, but the displayed units were in different metric schema.

IMHO,
switching units should be a safe and low-effort action.  Switching on the status bar should be the norm and should immediately change the displayed units in the UI.
The document should 'remember' the last selected schema and use that the next time the document is opened.

The preference setting should be used on new document creation.

Until the bugs are chased out of this, I'm moving the PR back to draft status.

bgbsww:
@sliptonic I want to acknowledge your point - somehow bit rot or my mistakes made this go from perfectly operational to basically doing nothing over the last few weeks.  I've fixed most of it, but I'm still stuck on getting the entire UI to refresh after a change in the status bar.  Once I sort that out I'll post a revised PR.

bgbsww:
This is supposed to:

- Create Gui field in Project Information with a pulldown to specify the unit system schema.
- Save this field with the document
- When a new document is created, this value is initialized with the Preferences -> General -> Unit System
- Initialize this field from the Preferences -> General -> Unit System for older files
- When a file is open, the Status Bar Dimension widget at the bottom of the main window both reflects and updates this setting
- When no file is open, the Status Bar Dimension widget sets the  Preferences -> General -> Unit System
- The Preferences -> General -> Project Unit System is removed
- The Preferences -> General -> Ignore Project Unit System Overrides the document setting when checked
- Files are marked as changed after a unit change, regardless of what the state of the ignore checkbox is

Note that an alternate behavior, which is the statusbar dimension widget never changes the default unit schema in the preferences, even when no document is open could be easily restored to this patch, I'm not sure which behavior feels more natural.

sliptonic:
>     * Create Gui field in Project Information with a pulldown to specify the unit system schema.
> 
>     * Save this field with the document

The property may be required to store the state persistently but I don't think we should expect users to change the value here.   I would suggest that the property view state be hidden.  

Regardless, if the user does change it in the document properties, we have to detect it and change the status bar setting to match.

bgbsww:
I thought that was as simple as Prop_Hidden, but that doesn't seem to work.  It also doesn't work for TipName, so it isn't just me.  Is there a technique here I'm missing?

PR #11267, opened 2023-11-03
[Sketcher] Translate DSH : replacement for rectangular array / move / copy / clone

PaddleStroke:
Translate DSH : replacement for rectangular array / move / copy / clone
Please note that it does not remove the code of the commands rectangular array / move / copy / clone. It only removes them from the toolbar/menus. Should the code be removed as well ?

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
@@ -220,7 +220,7 @@ public:
         Q_UNUSED(onSketchPos)
     }
 
-    /** on first shortcut, it toggles the first checkbox if there is go. Must be specialised if
+    /** on first shortcut, it toggles the first checkbox if there is one. Must be specialised if
      * this is not intended */
     void firstKeyShortcut() override
     {
@@ -230,13 +230,33 @@ public:
         }
     }
 
-    /** on second shortcut, it toggles the second checkbox if there is go. Must be specialised if
+    /** on second shortcut, it toggles the second checkbox if there is one. Must be specialised if
      * this is not intended */
     void secondKeyShortcut() override
     {
         if (nCheckbox >= 2) {
             auto secondchecked = toolWidget->getCheckboxChecked(WCheckbox::SecondBox);
             toolWidget->setCheckboxChecked(WCheckbox::SecondBox, !secondchecked);
+        }
+    }
+
+    /** on third shortcut, it toggles the third checkbox if there is one. Must be specialised if
+     * this is not intended */
+    void thirdKeyShortcut() override
+    {
+        if (nCheckbox >= 3) {
+            auto thirdchecked = toolWidget->getCheckboxChecked(WCheckbox::ThirdBox);
+            toolWidget->setCheckboxChecked(WCheckbox::ThirdBox, !thirdchecked);
+        }
+    }
+
+    /** on fourth shortcut, it toggles the fourth checkbox if there is one. Must be specialised if
+     * this is not intended */
+    void fourthKeyShortcut() override
+    {
+        if (nCheckbox >= 4) {
+            auto fourthchecked = toolWidget->getCheckboxChecked(WCheckbox::FourthBox);
+            toolWidget->setCheckboxChecked(WCheckbox::FourthBox, !fourthchecked);
         }
     }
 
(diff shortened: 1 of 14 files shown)


src/Mod/Sketcher/Gui/Resources/icons/tools/Sketcher_Translate.svg
maxwxyz, line 1: @PaddleStroke you could use the one from the current icon overhaul:
Sketcher_Translate

src/Mod/Sketcher/Gui/Resources/icons/pointers/Sketcher_Pointer_Create_Translate.svg
maxwxyz, line 1: @PaddleStroke matching cursor icon:
Sketcher_Pointer_Create_Translate

maxwxyz (commented):
Adjustments of the icons in line with #11785

src/Mod/Sketcher/Gui/Resources/icons/tools/Sketcher_Translate.svg
PaddleStroke: Got it I will rebase an use those

PaddleStroke:
Icons updated.

chennes:
Overall I think this looks like an excellent change, particularly from a UI perspective. @obelisk79 does that work for you?

obelisk79:
@chennes I will build and test later today. I don't see any images or issue tied to this so I'll have to spin it up for a test drive tonight.

FEA-eng:
It works really well, awesome tool. Just one thing - it seems that TAB doesn't work properly in the second step (when setting rows). I can only switch once to the other input field but can't freely switch between them with TAB, even though it's possible in the first step. Is that a bug ?

PaddleStroke:
Ok so the tab key does not cycle back to the first spinbox correct?
I will check that

FEA-eng:
> Ok so the tab key does not cycle back to the first spinbox correct?

Yes but only in the second step of this tool's usage (it works when setting copies). In that step, I can only use TAB once and then it stops working so I can't go back to the distance if I'm editing the angle.

Also, could you add shortcuts to increase and decrease the number of rows (like the shortcuts for the number of copies, they are very useful)?

PaddleStroke:
I have pushed 2 more commits that do : 
1 - Add two additional shortcuts R and F to modify the number of row. I/K would be better as they are next to U/J but I triggers vertical distance. So R and F were the least bad I could find.
2 - Modify the DSH framework so that it now cycles between the OVP and the widget spinboxes correctly.

PaddleStroke:
@chennes I have fixed the compile issues on linux. Now only maxOS is complaining but not because of this PR.

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
```diff
@@ -240,6 +248,32 @@ class DrawSketchDefaultWidgetController: public DrawSketchController<HandlerT,
         }
     }
 
+    /** on third shortcut, it toggles the second checkbox if there is go. Must be specialised if
```
chennes, line 251:
```suggestion
    /** on third shortcut, it toggles the second checkbox if there is one. Must be specialised if

```

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
```diff
@@ -240,6 +248,32 @@ class DrawSketchDefaultWidgetController: public DrawSketchController<HandlerT,
...
+     * this is not intended */
+    void thirdKeyShortcut() override
+    {
+        if (nCheckbox >= 3) {
+            auto thirdchecked = toolWidget->getCheckboxChecked(WCheckbox::ThirdBox);
+            toolWidget->setCheckboxChecked(WCheckbox::ThirdBox, !thirdchecked);
+        }
+    }
+
+    /** on fourth shortcut, it toggles the second checkbox if there is go. Must be specialised if
```
chennes, line 261: Do you mean...
```suggestion
    /** on fourth shortcut, it toggles the second checkbox if there is one. Must be specialised if
```
?

ryanthara:
> Translate DSH : replacement for rectangular array / move / copy / clone Please note that it does not remove the code of the commands rectangular array / move / copy / clone. It only removes them from the toolbar/menus. Should the code be removed as well ?

In the FreeCAD-Forum there is a [thread](https://forum.freecad.org/viewtopic.php?t=95073) about this and its "improvement" to get rid of the automatic added constraints in the former function of Sketcher_RectangularArray

PR #11271, opened 2023-11-04
Sketcher Gui: Clang tidy diagnostic errors and dead code fixes

abdullahtahiriyo:
- Adding headers and forward declarations as needed to fix the clang-tidy diagnose errors.
- Fixing some errors detected by clang-tidy once the headers were added.

### DIFF
--- a/src/Gui/Notifications.h
+++ b/src/Gui/Notifications.h
@@ -33,6 +33,8 @@
 
 namespace Gui {
 
+    class ViewProviderDocumentObject;
+
 /** Methods to seamlessly provide intrusive or non-intrusive notifications of error, warning,
  * messages, translated notifications, or untranslated notifications originating in a given
  * document object.
--- a/src/Mod/Sketcher/App/SketchAnalysis.h
+++ b/src/Mod/Sketcher/App/SketchAnalysis.h
@@ -27,6 +27,8 @@
 #include <memory>
 #include <vector>
 
+#include <Precision.hxx>
+
 #include <Base/Vector3D.h>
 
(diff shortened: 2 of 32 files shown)


PaddleStroke:
You opened and merged this warning-fixing PR after I posted the rework of many DSH generating conflicts all over the place. Now I have to rebase each branch, solve conflicts in each, rebuild each, retest.

I remember someone who complained about Uwe when he made a warning-fixing commit in the middle of a large refactor effort. And Uwe wasn't aware.

abdullahtahiriyo:
It is indeed not the same thing. However, I understand what you mean. And BTW git is right to note conflicts. They are conflicts.

 It is always best when this lint issues are not introduced in the first place. But there many of us have room for improvement. Me included. Hopefully the IDEs will be providing these issues when we write the code and greatly avoid the back and forth.

Nevertheless, in this case it should be rather easy for you to solve the conflicts because you are fully rewriting the file. You just need to take your version. For example if you modified DSHArc to port it to the new architecture, when it conflicts, just select your version. You do not need to test either, as you are taking the full version of your rewritten tool. Just look at the other new DSHs that are already in for the headers that are necessary, so that your PR does not have lint issues either. The only exception is if you modified a file from the framework. That one needs to be looked up carefully.

PaddleStroke:
Do not worry I already rebased this morning. By the way it may be good to merge slot / arcslot / conics so that we get feedback on potential issues. But https://github.com/FreeCAD/FreeCAD/pull/11294 need to go in first.

PR #11272, opened 2023-11-05
Sketcher: fix Line tool settings mode name

adrianinsaval:
one of the modes for line is wrongly named IMO
- point, length, width
should instead be:
- point, width, height
or
- point, vector

since it does not set the length of the line but horizontal and vertical distance between endpoints, plus the width is set before the height so it should be the second thing mentioned.


adrianinsaval:
@PaddleStroke @abdullahtahiriyo

PaddleStroke:
You are right. Both 
point, width, height
or
point, vector
are ok for me.

PR #11273, opened 2023-11-05
ComplexGeoData and PropertyComplexGeoData serialization

chennes:
Add the serialization functions to `ComplexGeoData`, and completes the `PropertyComplexGeoData` implementation of the Toponaming branch. Note that this code is NOT called by any active code yet, it exists to support upcoming TNP work.

The first commit is just @realthunder's `ComplexGeoData` code as originally written. The second commit applies our clang-format and resolves most issues identified by clang-tidy. The third commit adds the code for `PropertyComplexGeoData`, and the fourth commit is a minor refactor to address code complexity.

The fifth commit adds some simple tests for the serialization code: no tests exist yet for the deserialization, which will require some refactoring to be testable.

### DIFF
--- a/src/App/ComplexGeoData.cpp
+++ b/src/App/ComplexGeoData.cpp
@@ -38,9 +38,12 @@
 
 #include <Base/BoundBox.h>
 #include <Base/Placement.h>
+#include <Base/Reader.h>
 #include <Base/Rotation.h>
+#include <Base/Writer.h>
 
 #include <boost/iostreams/device/array.hpp>
+#include <boost/iostreams/stream.hpp>
 
 
 using namespace Data;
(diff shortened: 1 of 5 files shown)


src/App/ComplexGeoData.cpp
```diff
@@ -386,4 +389,208 @@ char ComplexGeoData::elementType(const char *name) const {
...
+void ComplexGeoData::setPersistenceFileName(const char *filename) const {
+    if(!filename)
+        filename = "";
+    _PersistenceName = filename;
+}
+
+void ComplexGeoData::Save(Base::Writer &writer) const {
+
+    if(!getElementMapSize()) {
+        writer.Stream() << writer.ind() << "<ElementMap/>\n";
```
Pesc0, line 407: why are we writing elementmap tags in complexgeodata? should this logic be in elementmap instead?

src/App/ComplexGeoData.cpp
```diff
@@ -386,4 +389,208 @@ char ComplexGeoData::elementType(const char *name) const {
...
+        }
+    }
+    std::size_t count = atoi(marker.c_str());
+    restoreStream(reader,count);
+}
+
+unsigned int ComplexGeoData::getMemSize(void) const {
+    flushElementMap();
+    if(_elementMap)
+        return _elementMap->size()*10;
```
Pesc0: 10 = magic?

src/App/ComplexGeoData.cpp
```diff
@@ -386,4 +389,208 @@ char ComplexGeoData::elementType(const char *name) const {
...
+    return 0;
+}
+
+void ComplexGeoData::beforeSave() const
+{
+    flushElementMap();
+    if (this->_elementMap)
+        this->_elementMap->beforeSave(Hasher);
```
Pesc0, line 628: the beforesave thing is back! why is it done this way? can we get rid of it somehow?

src/App/ComplexGeoData.cpp
```diff
@@ -386,4 +389,208 @@ char ComplexGeoData::elementType(const char *name) const {
...
+void ComplexGeoData::RestoreDocFile(Base::Reader &reader) {
+    std::string marker, ver;
+    reader >> marker;
+    if (boost::equals(marker, "BeginElementMap")) {
+        resetElementMap();
+        reader >> ver;
+        if (ver != "v1")
+            FC_WARN("Unknown element map format");
+        else {
+            resetElementMap(std::make_shared<ElementMap>());
```
Pesc0, line 605: I don't really like that reset takes in a parameter. make_shared should be inside reset i think.

src/App/PropertyGeo.cpp
```diff
@@ -1241,3 +1243,18 @@ TYPESYSTEM_SOURCE_ABSTRACT(App::PropertyComplexGeoData , App::PropertyGeometry)
 PropertyComplexGeoData::PropertyComplexGeoData() = default;
 
 PropertyComplexGeoData::~PropertyComplexGeoData() = default;
+
+void PropertyComplexGeoData::afterRestore()
```
Pesc0, line 1247: this smells like beforesave. can we get rid of it? append it to restore?

src/App/ComplexGeoData.h
```diff
@@ -280,6 +280,19 @@ class AppExport ComplexGeoData: public Base::Persistence, public Base::Handled
...
+    //@{
+    void Save (Base::Writer &writer) const;
+    void Restore(Base::XMLReader &reader);
+    void SaveDocFile(Base::Writer &writer) const;
+    void RestoreDocFile(Base::Reader &reader);
+    unsigned int getMemSize (void) const;
+    void setPersistenceFileName(const char *name) const;
+    virtual void beforeSave() const;
+    bool isRestoreFailed() const { return _restoreFailed; }
+    void resetRestoreFailure() const { _restoreFailed = true; }
```
Pesc0, line 293: should not be const?

src/App/ComplexGeoData.h
```diff
@@ -280,6 +280,19 @@ class AppExport ComplexGeoData: public Base::Persistence, public Base::Handled
     virtual void flushElementMap() const;
     //@}
 
+    /** @name Save/restore */
+    //@{
+    void Save (Base::Writer &writer) const;
+    void Restore(Base::XMLReader &reader);
+    void SaveDocFile(Base::Writer &writer) const;
+    void RestoreDocFile(Base::Reader &reader);
```
Pesc0: what's the difference between these?

Pesc0 (commented):
I haven't checked the logic at all, just general things that caught my eye. tests look pretty simple and straightforward.

src/App/ComplexGeoData.cpp
chennes: Yes, I agree, that would be a good refactor.

src/App/ComplexGeoData.cpp
chennes: Yes. Or rather, just not very important, getMemSize is itself just an optimization.

src/App/ComplexGeoData.cpp
chennes: That's a bigger refactor that's out of scope for this PR, IMO.

src/App/ComplexGeoData.cpp
```diff
@@ -386,4 +389,208 @@ char ComplexGeoData::elementType(const char *name) const {
...
+                        << "\"/>\n";
+        return;
+    }
+    writer.Stream() << " count=\"" << _elementMap->size() << "\">\n";
+    _elementMap->save(writer.beginCharStream(Base::CharStreamFormat::Raw) << '\n');
+    writer.endCharStream() << '\n';
+    writer.Stream() << writer.ind() << "</ElementMap2>\n" ;
+}
+
+void ComplexGeoData::Restore(Base::XMLReader &reader) {
```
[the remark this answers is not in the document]
chennes: That's how it is defined in the Persistence base class.

src/App/ComplexGeoData.h
chennes: I think it gets called from `const` methods, so has to be. `_restoreFailed` is mutable.

src/App/ComplexGeoData.h
chennes: See docs in Persistence.h

src/App/ComplexGeoData.cpp
chennes: I don't see how: all of the hashers (across all instances of ComplexGeoData) need to be marked appropriately before anything gets written.

Issue #11142, opened 2023-10-22
[Problem] Menus open in wrong screen on multiscreen (Windows 11)


OlivierHersent:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Reproduction : set-up a multi-screen configuration with FreeCad running on top screen. FreeCAD menus e.g. "Help" open on the bottom screen.

### Full version info

```shell
0.21.1
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
Full version info please. See [About](https://wiki.freecad.org/About).

luzpaz:
@OlivierHersent please follow-up

OlivierHersent:
Version 0.21.1

OlivierHersent:
Sorry it took me a while to realize that there is a "copy to clipboard" button.
[code]
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.21.1.33668 +26 (Git)
Build type: Release
Branch: (HEAD detached at 0.21.1)
Hash: f6708547a9bb3f71a4aaade12109f511a72c207c
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/United States (en_US)
[/code]

OlivierHersent:
When you test, also with the Windows screen relocate shortcut (WIN + Shift up + >) you will see multiple issues related to menu location, font size, and icon sizing. Probably the screen repaint primitives do not take the right screen as reference for location and resolution.

luzpaz:
Can you test on the dev version and report back ? (https://github.com/FreeCAD/FreeCAD-Bundle/releases/tag/weekly-builds)

luzpaz:
@OlivierHersent ping

maxwxyz:
Can confirm. 0.22dev 2023-10-26 2weekly.
Preferences window will open on the display where FreeCAD opened the first time on a setup with three monitors.

OlivierHersent:
Testing on latest build, much better behavior. I could reproduce the issue only by manually dragging the main window between two screens with different resolutions (small laptop & large second screen). At some point the window on large main screen switches to the resolution of the smaller screen, so everything appears huge. At this point the menus pop-up pretty much randomly, undocked.
[...]

PR #11275, opened 2023-11-05
Fixes #11102: Main window doesn't remember where it was after restart

### LINKED ISSUE #11102: [Regression] Main window doesn't remember where it was after restart

### Problem description

Over the last few days I've noticed that the main window does not remember where is was when the program was shut down.

Curiously, the the Properties panel and the Task panel do remember (I typically keep them on a separate monitor).

When FC is started again, all three are displayed on the on the monitor where the Property/Task panels were located.

This have never been the case in the past and all remembered their previous position.

### Full version info

Previous version where found:

```shell
OS: Manjaro Linux (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34697 (Git)
Build type: None
Branch: main
Hash: cf9b342
Python 3.11.5, Qt 5.15.11, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Installed mods:
    Silk 0.1.5
    Defeaturing 1.2.0
    Curves 0.6.13
    CurvedShapes 1.0.4
    DynamicData 2.53.0
    lattice2 1.0.0
    sheetmetal 0.3.0
    Manipulator 1.5.0
    kicadStepUpMod 10.19.5
    QuickMeasure 2022.10.28
    fasteners 0.4.66
    Help 1.0.3
    Assembly
    ProDarkThemePreferencePack 1.0.0
    A2plus 0.4.60n
    freecad.gears 1.0.0
    ExtremeProDark 2.9.0
    Assembly4 0.50.4
```

Latest version where found:

```shell
OS: Debian GNU/Linux 12 (bookworm) (XFCE/xfce)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34709 (Git)
Build type: Release
Branch: main
Hash: 7eda70c737625383d358871672d60a229e5e21b6
Installed mods: 
  * MnesarcoUtils 0.2.2
  * ZolkoBase 0.0.1
  * Pyramids-and-Polyhedrons
  * lattice2 1.0.0
  * Curves 0.6.13
  * Nodes 0.1.36
  * Help 1.0.3
  * QuickMeasure 2022.10.28
  * Manipulator 1.5.0
  * Assembly
  * Assembly4 0.50.4
  * freecad.gears 1.0.0
  * sheetmetal 0.3.1
  * Silk 0.1.5
  * kicadStepUpMod 10.20.3
  * A2plus 0.4.60n
  * DynamicData 2.53.0
  * workfeature-macro
  * fasteners 0.4.66
  * ExtremeProDark 2.9.0
  * Defeaturing 1.2.1
  * CurvedShapes 1.0.4
```

### Code of Conduct

wwmayer:


wwmayer:
https://github.com/FreeCAD/FreeCAD/issues/11102#issuecomment-1793809625 confirms that this PR fixes the regression

adrianinsaval:
has this been tested on windows? I'm concerned about multi monitor setups on windows. FreeCAD used to fail to launch after being used on a second monitor there. Btw, what do we know about the behavior on x11 vs wayland here?

wwmayer:
> has this been tested on windows?

I don't think so. But does the current version work on Windows?

> Btw, what do we know about the behavior on x11 vs wayland here?

I don't think that this depends on the used display server, does it?

wwmayer:
Since Qt 4.2 there are the methods [QWidget::saveGeometry() & restoreGeometry()](https://doc.qt.io/qt-5/qwidget.html#restoreGeometry) and according to their [implementations](https://codebrowser.dev/qt6/qtbase/src/widgets/kernel/qwidget.cpp.html) they handle the screen, too.

So, using them would be an alternative way.

adrianinsaval:
> I don't think so. But does the current version work on Windows?
I think so, I was never able to reproduce so I don't know if that really means the bug was solved, I didn't have a second monitor at the time and now it isn't crashing for me. Although the position if FreeCAD was closed on the second monitor is not kept, it always starts on the same monitor.

With your changes, it at least starts on the correct monitor but the position is not right, I closed FreeCAD in this position:
image
but when opening again it shows up here:
image

I can test on linux on x11 vs wayland later

maxwxyz:
Does this also benefit or take into account https://github.com/FreeCAD/FreeCAD/issues/11142 ?
There on Windows, when switching e.g. from a three monitor setup to a two monitor setup, some windows (settings, parameters) are out of reach and cannot be moved to a visible monitor. I hope that now not the entire main window is opened or moved out of reach for the user on windows.

wwmayer:
> Does this also benefit or take into account https://github.com/FreeCAD/FreeCAD/issues/11142 ?

No, it doesn't. This PR only fixes a regression in v0.22.

PR #9699, opened 2023-05-30
Fix clipping near camera in orthogonal view

Jolbas:
Fix #6836 by moving camera backwards the same length as the height of view port.

Thank you for creating a pull request to contribute to FreeCAD! Place an "X" in between the brackets below to "check off" to confirm that you have satisfied the requirement, or ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10) if there is something you don't understand.

- [X]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR

Please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [1.0 Changelog Forum Thread](https://forum.freecad.org/viewtopic.php?f=10&t=69438).

---

PR #11279, opened 2023-11-05
Gui: Fix window center rotation mode and orthographic view clipping

### LINKED ISSUE #6836: [Bug] Disappearing faces (+ corners appear cut) (Coin3D related)

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?t=68353

### Version

0.20 (Development)

### Full version info

```shell
[code]
OS: Windows 10 (10.0)
Word size of FreeCAD: 64-bit
Version: 0.20.28793 (Git)
Build type: Release
Python 3.8.13, Qt 5.12.9, Coin 4.0.0, OCC 7.5.3
[/code]
```

### Subproject(s) affected?

_No response_

### Issue description

the behaves the same way in 19.3 release and all of the versions of 0.20 that I have tried for the last few months. I have created a 40mm cube and rotated it on the screen. In the forum discussion, there are examples of the improper rendering.

I have seen this failure with the cube created in Part Design and also in Part. I have verified that the same failure mechanism happens for more complex designs, but the simple cube most easily reproduces the problem. If the part is saved when it displays the problem, opening it in any of the versions from 0.19.3 through the current  version show the same improper display.

I have changed the display preferences to use each and every render cache option, every anti-aliasing option, and transparent objects options. Additionally, I have tried checking "Use software OpenGL", restarted the program, and the problem persists.

[t1p.zip](https://github.com/FreeCAD/FreeCAD/files/8635983/t1p.zip)
t1p
.

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #11243: [Problem] CAD Navigation around Window Center broken

### Problem description

In the current weekly build the default navigation gets sometimes out of control. Normally I use the CAD navigation style around the center. The red sphere should indicate my center of rotation in the viewport and therefore be steady. 
In the following clip this is also the correct behavior in the beginning but sometimes the model jumps around and you can clearly see, that the center of rotation itself is rotating around another point (not the scene/world center):

https://github.com/FreeCAD/FreeCAD/assets/6246609/29d486b7-f0be-423c-a618-c699884fb02e

The whole clip I am holding the middle and right mouse button for rotating the view. It somehow seems to occur more often when a face or part of the model is hovered before I start the orbit.

### Full version info

```shell
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34810 (Git)
Build type: Release
Branch: main
Hash: 196cbecde66a1c3a1b56b8c532ad1faedff4818b
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Installed mods: 
  * 3DfindIT 1.2.0
  * 3D_Printing_Tools
  * BIM 2021.12.0
  * CfdOF 1.24.7
  * CurvedShapes 1.0.4
  * Curves 0.6.14
  * Defeaturing 1.2.1
  * fasteners 0.4.68
  * freecad.gears 1.0.0
  * OpenDark 2023.10.30
  * sheetmetal 0.3.1
  * Silk 0.1.5
```

### Subproject(s) affected?

Core

### Anything else?

_No response_

### Code of Conduct

Rexbas:
This PR fixes #11243 and re-fixes #6836. #9699, The original fix for #6836 was a huge improvement but I was still able to reproduce the cut corners and sometimes random glitches. With this fix that does not happen anymore, at least during my testing.

The new approach finds a bounding sphere of the scene, then it is repositioning the camera such that the `nearDistance` can be 0 and the `farDistance` the diameter of the bounding sphere. Finally, to fix #11243, we change the focal distance so the focal point stays the same.

Currently, this fix runs only in `NavigationStyle::reorientCamera(..)`. There are also other methods that change the camera where it is still possible to get glitchy results.  I am going to take a look at coin3d `SoRenderManagerP::setClippingPlanes()` to see if I can make a fix there, so this PR is maybe a temporary fix.

### DIFF
--- a/src/Gui/NavigationAnimation.cpp
+++ b/src/Gui/NavigationAnimation.cpp
@@ -57,6 +57,7 @@ FixedTimeAnimation::FixedTimeAnimation(NavigationStyle* navigation, const SbRota
 
 void FixedTimeAnimation::initialize()
 {
+    navigation->findBoundingSphere();
     prevAngle = 0;
     prevTranslation = SbVec3f(0, 0, 0);
 
@@ -133,6 +134,7 @@ SpinningAnimation::SpinningAnimation(NavigationStyle* navigation, const SbVec3f&
 
 void SpinningAnimation::initialize()
 {
+    navigation->findBoundingSphere();
     prevAngle = 0;
 
     navigation->setViewing(true);
--- a/src/Gui/NavigationStyle.cpp
+++ b/src/Gui/NavigationStyle.cpp
@@ -503,17 +503,37 @@ void NavigationStyle::reorientCamera(SoCamera* camera, const SbRotation& rotatio
     // Set new orientation value by accumulating the new rotation
     camera->orientation = rotation * camera->orientation.getValue();
 
-    // Fix issue with near clipping in orthogonal view
-    if (camera->getTypeId().isDerivedFrom(SoOrthographicCamera::getClassTypeId())) {
-        camera->focalDistance = static_cast<SoOrthographicCamera*>(camera)->height;
-    }
-
     // Distance from rotation center to new camera position in global coordinate system
     SbVec3f newRotationCenterDistance;
     camera->orientation.getValue().multVec(rotationCenterDistanceCam, newRotationCenterDistance);
 
     // Reposition camera so the rotation center stays in the same place
     camera->position = rotationCenter + newRotationCenterDistance;
+
+    // Fix issue with near clipping in orthogonal view
+    if (camera->getTypeId().isDerivedFrom(SoOrthographicCamera::getClassTypeId())) {
+
+        // Find a bounding sphere for the scene
+        SoGetBoundingBoxAction action(viewer->getSoRenderManager()->getViewportRegion());
+        action.apply(viewer->getSceneGraph());
+        SbSphere boundingSphere;
+        boundingSphere.circumscribe(action.getBoundingBox());
+
+        // The center of the bounding sphere in camera coordinate system
+        SbVec3f center;
+        camera->orientation.getValue().inverse().multVec(boundingSphere.getCenter() - camera->position.getValue(), center);
+
+        SbVec3f dir;
+        camera->orientation.getValue().multVec(SbVec3f(0, 0, -1), dir);
+
+        // Reposition the camera but keep the focal point the same
+        // nearDistance is 0 and farDistance is the diameter of the bounding sphere
+        float repositionDistance = -center.getValue()[2] - boundingSphere.getRadius();
+        camera->position = camera->position.getValue() + repositionDistance * dir;
+        camera->nearDistance = 0;
+        camera->farDistance = 2 * boundingSphere.getRadius();
+        camera->focalDistance = camera->focalDistance.getValue() - repositionDistance;
+    }
 }
 
 void NavigationStyle::panCamera(SoCamera * cam, float aspectratio, const SbPlane & panplane,
--- a/src/Gui/NavigationStyle.cpp
+++ b/src/Gui/NavigationStyle.cpp
@@ -1300,8 +1300,8 @@ void NavigationStyle::setViewingMode(const ViewerMode newmode)
         // Set up initial projection point for the projector object when
         // first starting a drag operation.
         animator->stop();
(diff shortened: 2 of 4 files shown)


wwmayer:
> This PR fixes https://github.com/FreeCAD/FreeCAD/issues/11243

I can confirm that it fixes this issue. Btw, a very easy way to reproduce this behaviour is to rotate, stop rotate, zoom in or out with the wheel and then rotate again.

However, I am bit concerned about the way how it is fixed. It determines the scene bounding box and therefore it has to traverse the whole scene graph and this  makes it  a rather expensive operation. Because reorientCamera() is called for every little movement it could affect the speed of user interactions.

Actually it suffices to determine the scene bounding box only once when starting to rotate.

Btw, which previous PR caused this regression?

Rexbas:
> Actually it suffices to determine the scene bounding box only once when starting to rotate.

Yeah you are right, I added a commit that finds the bounding sphere before dragging or animating and also after the scene graph is set. The bounding sphere is stored in a private member of `NavigationStyle` and `reorientCamera()` uses this value. Yesterday, I opened a PR (https://github.com/coin3d/coin/pull/510) in coin with almost the same solution as in this PR. If that gets merged into coin then most of this PR could be reverted.

> Btw, which previous PR caused this regression?

PR #10881 caused this regression, specifically commit 9be3bfe7bf7f3f76f896e7cb5175509d5b3c6289. In this commit I changed `reorientCamera()` to be able to handle rotating about a point that is not the focal point, which was needed for animations that combine rotation and translation. However, another part of the problem came from #9699. Here the focal distance was changed depending on the height which is related to zoom but the focal distance affects the rotation center. And the new `reorientCamera()` did not like that.

maxwxyz:
Can confirm this fixes the issue of the rotation center moving around.
Although when panning the new rotation center is fairly offset from the last position, shouldn't it stay at the same distance/offset(=depth) when I just pan to the right?

Rexbas:
> shouldn't it stay at the same distance/offset(=depth) when I just pan to the right?

Yes, if you align the camera with FRONT (or whatever the nearest orientation is) before panning then the rotation center should stay in the object. In the video your camera is not aligned so it makes sense that the rotation center is moving away from the object.

maxwxyz:
So the rotation center moves on a perpendicular plane to the view direction (with an offset of the last position)?

Rexbas:
Yes

maxwxyz:
is this now a partial fix or a full fix of the mentioned issues/PRs? Does this need to wait until the Coin PR is merged or is this PR finished? From a user PoV this is working.

Rexbas:
In my opinion this is a full fix for #11243 and a partial/temporary fix for #6836 to resolve the worst of the clipping problems. We don't have to wait for the PR I created in coin. That is approximately the same fix but in a better place and we can revert most of this PR when we update to a coin version that includes the PR I created in coin.

adrianinsaval:
@wwmayer should we still worry about performance here or do you think this one is ok now?

wwmayer:
Should be fine now.

Rexbas:
Thanks, I can reproduce it and I will try to fix it.

maxwxyz:
@Rexbas shall I create an issue or are you on it?

Rexbas:
I am currently working on another issue and have planned to do this after it is finished.

Rexbas:
@maxwxyz, I looked into it but I don't think this is an issue we can fix completely. First, it does not only happen to the CAD style and window center rotation mode. It happens to all styles and all rotation modes. Also, you don't necessary need a large scene. It happens when you zoom in a lot, which is more likely when you have a large scene.

The reason why we can't completely fix this is because I suspect this is a float precision issue. The camera uses floats for orientation and position and when the camera is rotated there is almost always a small error. This error is small enough to not notice it in most cases. But when you zoom in a lot then the error becomes noticeable in the form of a moving rotation center. But really it is the camera not moving perfectly due to inaccuracies. I tried the same in Fusion 360 and there the rotation center also moves when zoomed in a lot.

So I don't know how to fix this or how to reduce the problem, it is probably not even possible to fix it completely. If you think this is an important problem, consider opening an issue. Maybe someone has an idea. But I don't.

PR #11281, opened 2023-11-05
Sketcher: continuous mode improvements

abdullahtahiriyo:
As per:
https://forum.freecad.org/viewtopic.php?p=716863#p716863

Problems:
* Continuous mode is not well handled. Eg. drawing a rectangle, if I press Esc/RMB after clicking the first point (and before clicking the second), the rectangle feature is just left, instead of just canceling the current rectangle and be ready for a new one.
* I got some cases where pressing Esc does nothing, but I'm not able to reproduce reliably, so will eventually come back later on this one.

Solution:
- ESC or RMB will reset the handler to the first state machine state if not in the first state. If already in the first state, it terminates the handler.

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchController.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchController.h
@@ -564,6 +564,10 @@ protected:
         onViewIndexWithFocus = 0;
 
         configureOnViewParameters();
+
+        if (init) {
+            handler->moveCursorToSketchPoint(prevCursorPosition);
+        }
     }
     //@}
 
--- a/src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchDefaultHandler.h
@@ -427,6 +427,15 @@ public:
         if (key == SoKeyboardEvent::M && pressed && !this->isLastState()) {
             this->iterateToNextConstructionMethod();
         }
+        else if (key == SoKeyboardEvent::ESCAPE && pressed) {
+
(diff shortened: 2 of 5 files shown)


abdullahtahiriyo:
@PaddleStroke 
@0penBrain 

This PR needs testing. See if this meets the requirements.

If you look at the code, there is hack in there. I do not like it, but I do not know exactly what is preventing ViewProviderSketch from receiving keyboard events after reset. I would be grateful if a real solution is found.

PaddleStroke:
> ESC or RMB will reset the handler to the first state machine state if not in the first state. If already in the first state, it terminates the handler.

This is a good solution.

src/Mod/Sketcher/Gui/DrawSketchController.h
```diff
@@ -564,6 +564,10 @@ class DrawSketchController
         onViewIndexWithFocus = 0;
 
         configureOnViewParameters();
+
+        if (init) {
+            handler->moveCursorToSketchPoint(prevCursorPosition);
```
PaddleStroke, line 569: The bug described in #11307 (see https://youtu.be/24UYGO5PgjU) happens here. This moveCursorToSketchPoint is buggy

PaddleStroke: Besides the bug in moveCursorToSketchPoint, is this necessary here? Why are you doing this?

abdullahtahiriyo: Replied in the other PR.

PR #11286, opened 2023-11-06
Base: fix crash in PyException::ReportException()

wwmayer:
https://forum.freecad.org/viewtopic.php?p=716961#p716961

### DIFF
--- a/src/Base/Interpreter.cpp
+++ b/src/Base/Interpreter.cpp
@@ -126,6 +126,7 @@ void PyException::ReportException () const
     if (!_isReported) {
         _isReported = true;
         // set sys.last_vars to make post-mortem debugging work
+        PyGILStateLocker locker;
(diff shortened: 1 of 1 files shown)


adrianinsaval:
is this something we should backport? the diff looks trivial but I think there where changes in how the interpreter works in 0.22, right?

wwmayer:
There was a previous fix (792af641db66) to make the post-mortem debugger of Python to work. I think both changes can be backported. All what needs to be done is changing the method PyException::ReportException
```
void PyException::ReportException () const
{
    if (!_isReported) {
        _isReported = true;
        // set sys.last_vars to make post-mortem debugging work
        PyGILStateLocker locker;
        PySys_SetObject("last_traceback", PP_last_traceback);
        Base::Console().DeveloperError("pyException","%s%s: %s\n",
            _stackTrace.c_str(), _errorType.c_str(), what());
    }
}
```

Issue #11289, opened 2023-11-06
[Problem] Sketcher: Offset on some geometry freezes FreeCAD


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

@PaddleStroke Using Offset tool on B-Spline parts renders FreeCAD unusable. I cannot select anything and must quit it with Task Manager.
Maybe the Offset tool should be greyed out or not be able to use when geometry is selected which is currently not supported (external geometry, splines,..)

Currently not working for (if the following geometry or parts of it are in the offset selection):

- [ ] B-Splines
- [ ] Ellipses / their arcs
- [ ] External geometry

![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/a18edfb5-68ab-429a-9c46-f0ab01dc2eaf)

### Full version info

```shell
[code]
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34873 (Git)
Build type: Release
Branch: main
Hash: 889c8c49f1bbc0910c8dc32a512f4c6fbedeb8b9
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * 3DfindIT 1.2.0
  * 3D_Printing_Tools
  * BIM 2021.12.0
  * CfdOF 1.24.7
  * CurvedShapes 1.0.4
  * Curves 0.6.14
  * Defeaturing 1.2.1
  * fasteners 0.4.68
  * freecad.gears 1.0.0
  * freecad_metal_workbench 0.0.1
  * OpenDark 2023.10.30
  * sheetmetal 0.3.1
  * Silk 0.1.5
[/code]
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
Interesting, bspline are kind of working but not always.
![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/e6d3296f-bc8f-469b-83c2-ae683a226167)
It seems they require extra work.

PaddleStroke:
For now I will disable offset for bspline/ellipse/hyperbola/parabola as they are not reliable. Then we can open issues asking to support those geometries.

Issue #11288, opened 2023-11-06
[Problem] Sketcher: Rectangle Tool Settings does not display the mode when invoked


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

@PaddleStroke hope it is ok to mention you in the issues directly.
When I invoke the centered rectangle from the dropdown in the toolbar it opens the correct tool, but in the task box settings, the wrong mode is displayed:
![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/ebbdcfab-87f1-4573-8451-3e8019c1873c)

### Full version info

```shell
[code]
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34873 (Git)
Build type: Release
Branch: main
Hash: 889c8c49f1bbc0910c8dc32a512f4c6fbedeb8b9
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * 3DfindIT 1.2.0
  * 3D_Printing_Tools
  * BIM 2021.12.0
  * CfdOF 1.24.7
  * CurvedShapes 1.0.4
  * Curves 0.6.14
  * Defeaturing 1.2.1
  * fasteners 0.4.68
  * freecad.gears 1.0.0
  * freecad_metal_workbench 0.0.1
  * OpenDark 2023.10.30
  * sheetmetal 0.3.1
  * Silk 0.1.5
[/code]
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #11285, opened 2023-11-06
[Problem] Sketcher: Offset rectangle sould be parallel


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

The offset lines should be parallel to the original rectangle (are at creation but can be moved freely):
![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/d02e718c-d1c8-4b06-a09e-e1fd2667ee8d)

### Full version info

```shell
[code]
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34873 (Git)
Build type: Release
Branch: main
Hash: 889c8c49f1bbc0910c8dc32a512f4c6fbedeb8b9
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * 3DfindIT 1.2.0
  * 3D_Printing_Tools
  * BIM 2021.12.0
  * CfdOF 1.24.7
  * CurvedShapes 1.0.4
  * Curves 0.6.14
  * Defeaturing 1.2.1
  * fasteners 0.4.68
  * freecad.gears 1.0.0
  * freecad_metal_workbench 0.0.1
  * OpenDark 2023.10.30
  * sheetmetal 0.3.1
  * Silk 0.1.5
[/code]
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
Thanks for the report. 2 missing parallels it seems then.

Issue #11284, opened 2023-11-06
[Problem] Sketcher Tool Overlay: Cycle through measurements with Tab-key


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

@PaddleStroke how is the new overlay of the tools in sketcher called?
When sketching I can change to the other dialog to edit the other dimension but I cannot cycle back to define the other dimension again. It should me allow to switch between all present overlay dialogs with the tab key.
For example I like to edit the angle first and then the length, thats currently not possible:
![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/0ff6708b-74fb-46d4-9d9e-5381d37fdba2)

### Full version info

```shell
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34873 (Git)
Build type: Release
Branch: main
Hash: 889c8c49f1bbc0910c8dc32a512f4c6fbedeb8b9
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * 3DfindIT 1.2.0
  * 3D_Printing_Tools
  * BIM 2021.12.0
  * CfdOF 1.24.7
  * CurvedShapes 1.0.4
  * Curves 0.6.14
  * Defeaturing 1.2.1
  * fasteners 0.4.68
  * freecad.gears 1.0.0
  * freecad_metal_workbench 0.0.1
  * OpenDark 2023.10.30
  * sheetmetal 0.3.1
  * Silk 0.1.5
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
Yes this is a bug, I got it fixed already on my dev branch. I will push PR soon.

maxwxyz:
@PaddleStroke in the triangle tool it already works

PR #11294, opened 2023-11-06
[Sketcher] Tool parameter : various fixes

PaddleStroke:
@abdullahtahiriyo this PR fixes a first batch of comments that has been made.
It does not implement the preference to disable onviewparameters as you are doing it already.

Fixes issues by @0penBrain 

> * Most severe one is probably the Tab management. Eg. when drawing a rectangle, X coord is the selected one. If I press Tab, I go to Y coord. Fine. But here if I press Tab again, then coord selection is lost forever. I would expect Tab only to circle between currently available coords/dims. Also I'd expect Shift+Tab to work (circling in the other direction).
> * While keeping the coord/dim boxes on the viewport is well handled on top and left, it isn't on right and bottom where they disappear out of view.
> 
> * The mode named "Diagonal corners" isn't. It is "1 corner + length & width". A true "diagonal corners" mode is missing.
> * Same goes for "Center and corner" which actually is "Center + length & width".

Fixes issue from chrisb:

> In certasin positions the dynamically moving input fields are lying over each other. In that case the one having the focus should be on top.

Fixes the issue of spinbox getting in the way raised on the german topic.

Fixes https://github.com/FreeCAD/FreeCAD/issues/11289
Fixes https://github.com/FreeCAD/FreeCAD/issues/11288
Fixes https://github.com/FreeCAD/FreeCAD/issues/11287
Fixes https://github.com/FreeCAD/FreeCAD/issues/11285
Fixes https://github.com/FreeCAD/FreeCAD/issues/11284

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
@@ -253,25 +253,7 @@ protected:
 
     /// Automatic default method update in combobox
     void doConstructionMethodChanged() override
-    {
-        nParameter = WidgetParametersT::size(handler->constructionMethod());
-        nCheckbox = WidgetCheckboxesT::size(handler->constructionMethod());
-        nCombobox = WidgetComboboxesT::size(handler->constructionMethod());
-
-        // update the combobox only if necessary (if the change was not triggered by the
-        // combobox)
-        if constexpr (PFirstComboboxIsConstructionMethod == true) {
-            auto currentindex = toolWidget->getComboboxIndex(WCombobox::FirstCombo);
-            auto methodint = static_cast<int>(handler->constructionMethod());
-
-            if (currentindex != methodint) {
-                // avoid triggering of method change
-                boost::signals2::shared_connection_block combobox_block(
-                    connectionComboboxSelectionChanged);
-                toolWidget->setComboboxIndex(WCombobox::FirstCombo, methodint);
-            }
-        }
-    }
+    {}
     //@}
 
 private:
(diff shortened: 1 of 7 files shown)


PaddleStroke:
@abdullahtahiriyo that's it I fixed all the issues raised that I could fix.

The things that needs to be fixed still : 
- Preferences to disable OnViewParameter / widget.
- Cursor text (as this is linked to preferences). Also we did not decide what we want to do when enabled.
- @0penBrain mentioned that 3p rectangle is in fact 3p parallelepiped. But I don't think there's anything we can do to improve this. OpenBrain do you have a suggestion or was it just a comment and we leave it like this?

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
```diff
@@ -253,25 +253,7 @@ class DrawSketchDefaultWidgetController: public DrawSketchController<HandlerT,
...
-
-            if (currentindex != methodint) {
-                // avoid triggering of method change
-                boost::signals2::shared_connection_block combobox_block(
-                    connectionComboboxSelectionChanged);
-                toolWidget->setComboboxIndex(WCombobox::FirstCombo, methodint);
-            }
-        }
-    }
+    {}
```
abdullahtahiriyo, line 256: I am curious. 

This you moved to the reset function below. However, I cannot come to a reason why is this change necessary.

 In other words, do we need this code on every reset, even when the construction mode did not change? Is it not only necessary when changing construction mode?

PaddleStroke: This is commit  [da256f2](https://github.com/FreeCAD/FreeCAD/pull/11294/commits/da256f2a94b76395195e19e59f74c5b6f67f36da)
It is to solve two problems. When you initialize a tool with a construction methode that is not default, then you have 2 problems : 
1 - The number of parameters / checkbox / combobox / OVP initialized is not the correct one. Which results in crash for Arc tool when you try to initiate 3 rim point arc.
2 - The combobox do not show the correct entry. Currently in master if you start the command 'centered rectangle' then the combobox shows 'diagonal'.

PaddleStroke: So if we keep the multiple tools in the toolbars (3 rim points arcs, centered rectangle, oblong) this is necessary.
But if we remove the tools, then its not necessary anymore.

PaddleStroke:
I just seen there are multiple lint warning. I'll fix them

src/Mod/Sketcher/Gui/DrawSketchDefaultWidgetController.h
abdullahtahiriyo: Ah! It is the overriding of the default construction mode on initialisation. Good then! Thanks.

abdullahtahiriyo:
> * Cursor text (as this is linked to preferences). Also we did not decide what we want to do when OVP are enabled. My opinion is to disable the text when OVP are active. We don't need additional clutter.

I would implement it like you say. But I would do it transparently. Because you are going to need code at the tools for this, call a function there. The function will check the status of the parameter and show or not show depending on the parameter. If tomorrow we want all cursor texts in both modes, we just change that function. Does it sound ok to you?

PR #11307, opened 2023-11-07
Sketcher: Control visibility of OVPs via parameter

abdullahtahiriyo:
Three modes controlled by parameter:
Grp: Preferences/Mod/Sketcher/Tools
Name: OnViewParameterVisibility

0 => disabled
1 => only dimensional OVP
2 => all OVP visible

OVP = On-View Parameters

DSHs need to mark whether an OVP is positioning or dimensional. It is done for DSH_Line. Others need follow.

There is a mode for dynamic overriding of default setting. This is yet unconnected pending deciding how to trigger it.

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
@@ -306,11 +306,17 @@ void DSHLineController::configureToolWidget()
     onViewParameters[OnViewParameter::Second]->setLabelType(Gui::SoDatumLabel::DISTANCEY);
 
     if (handler->constructionMethod() == ConstructionMethod::OnePointLengthAngle) {
-        onViewParameters[OnViewParameter::Fourth]->setLabelType(Gui::SoDatumLabel::ANGLE);
+        onViewParameters[OnViewParameter::Fourth]->setLabelType(
+            Gui::SoDatumLabel::ANGLE,
+            Gui::EditableDatumLabel::Function::Dimensioning);
     }
     else {
-        onViewParameters[OnViewParameter::Third]->setLabelType(Gui::SoDatumLabel::DISTANCEX);
-        onViewParameters[OnViewParameter::Fourth]->setLabelType(Gui::SoDatumLabel::DISTANCEY);
+        onViewParameters[OnViewParameter::Third]->setLabelType(
+            Gui::SoDatumLabel::DISTANCEX,
+            Gui::EditableDatumLabel::Function::Dimensioning);
+        onViewParameters[OnViewParameter::Fourth]->setLabelType(
+            Gui::SoDatumLabel::DISTANCEY,
+            Gui::EditableDatumLabel::Function::Dimensioning);
     }
 }
 
@@ -392,11 +398,11 @@ void DSHLineController::adaptParameters(Base::Vector2d onSketchPos)
     switch (handler->state()) {
         case SelectMode::SeekFirst: {
             if (!onViewParameters[OnViewParameter::First]->isSet) {
-                onViewParameters[OnViewParameter::First]->setSpinboxValue(onSketchPos.x);
+                setOnViewParameterValue(OnViewParameter::First, onSketchPos.x);
             }
 
             if (!onViewParameters[OnViewParameter::Second]->isSet) {
-                onViewParameters[OnViewParameter::Second]->setSpinboxValue(onSketchPos.y);
+                setOnViewParameterValue(OnViewParameter::Second, onSketchPos.y);
             }
 
             bool sameSign = onSketchPos.x * onSketchPos.y > 0.;
@@ -414,11 +420,11 @@ void DSHLineController::adaptParameters(Base::Vector2d onSketchPos)
                 Base::Vector3d vec = end - start;
 
                 if (!onViewParameters[OnViewParameter::Third]->isSet) {
-                    onViewParameters[OnViewParameter::Third]->setSpinboxValue(vec.x);
+                    setOnViewParameterValue(OnViewParameter::Third, vec.x);
                 }
 
                 if (!onViewParameters[OnViewParameter::Fourth]->isSet) {
-                    onViewParameters[OnViewParameter::Fourth]->setSpinboxValue(vec.y);
+                    setOnViewParameterValue(OnViewParameter::Fourth, vec.y);
                 }
 
                 bool sameSign = vec.x * vec.y > 0.;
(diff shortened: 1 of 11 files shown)


abdullahtahiriyo:
@PaddleStroke 

Take a look.

src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
```diff
@@ -306,11 +306,17 @@ void DSHLineController::configureToolWidget()
     onViewParameters[OnViewParameter::Second]->setLabelType(Gui::SoDatumLabel::DISTANCEY);
 
     if (handler->constructionMethod() == ConstructionMethod::OnePointLengthAngle) {
-        onViewParameters[OnViewParameter::Fourth]->setLabelType(Gui::SoDatumLabel::ANGLE);
+        onViewParameters[OnViewParameter::Fourth]->setLabelType(
+            Gui::SoDatumLabel::ANGLE,
+            Gui::EditableDatumLabel::Function::Dimensioning);
```
PaddleStroke, line 311: This is probably not working in 'dimension ovp only' mode. Because here `OnViewParameter::Third` was not set because it is using the default 'DISTANCE' type. But if you don't set it now, then it will default at Positioning function.
So you need to add : 

```
        onViewParameters[OnViewParameter::Third]->setLabelType(
            Gui::SoDatumLabel::DISTANCE,
            Gui::EditableDatumLabel::Function::Dimensioning);
```

src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
```diff
@@ -306,11 +306,17 @@ void DSHLineController::configureToolWidget()
     onViewParameters[OnViewParameter::Second]->setLabelType(Gui::SoDatumLabel::DISTANCEY);
 
     if (handler->constructionMethod() == ConstructionMethod::OnePointLengthAngle) {
-        onViewParameters[OnViewParameter::Fourth]->setLabelType(Gui::SoDatumLabel::ANGLE);
+        onViewParameters[OnViewParameter::Fourth]->setLabelType(
+            Gui::SoDatumLabel::ANGLE,
+            Gui::EditableDatumLabel::Function::Dimensioning);
     }
     else {
```
PaddleStroke, line 313: Also here it is problematic too. Because the else is for the 2 other methodes : 
- 1 point + width / height. For which you want the labels to be 'Dimensioning'.
- 2 Points : for this methode, the 4 labels are positioning because you are positioning 2 points. So for this methode third and fourth need to be positioning. It is to be noted that this methode is useless if the positioning OVP are hidden.

PaddleStroke:
If you want you can merge as is and I will fix the set of ovp and do the pref gui

abdullahtahiriyo:
@PaddleStroke 

Just take the commits to your branch and adapt the DSHs as you see it fit. Then you can merge the branch with commits from both.

src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
abdullahtahiriyo: I did a quick test and I had the impression it worked. Perhaps I am mistaken. Just give it a try when you cherry-pick those commits...

PaddleStroke:
Ok I take these 5 commits and will do this and open a new PR.

PaddleStroke:
There are several bugs you introduced here: 
- Start line tool. Press M to change the mode. The cursor moves by itself. Press M many time and the mouse goes out of freecad.
- Press Esc does the same.
- Finishing a command does the same. So when you click the second point of a line, the mouse cursor jump somewhere else

PaddleStroke:
Video of the bug : https://youtu.be/24UYGO5PgjU
So the mouse jumping happens with windows OS scaling. I guess in the esc PR you set the cursor position manually somewhere.

PaddleStroke:
Ok so to summarize there are 2 bugs : 
1 - cursor jumping introduced by https://github.com/FreeCAD/FreeCAD/pull/11281/commits/df82849b55ef229e188285f4a11a9f2d0867d1e9#diff-5295f390a7bba0927eb8928992a128a6bd330a2ef36d8effb2c5aaefd6cbba3e

2 - Mode not changing : This is introduced in this current PR. What happens is when you are inside a mode with labels and press M, the methode changes, but the focus is lost apparantly. So the M key fails.

PaddleStroke:
Regarding bug 2, I'm fixing it by doing : 

```
    /** function that is called by the handler when the construction mode changed */
    void onConstructionMethodChanged()
    {
        // Make sure we do not loose focus if next methode does not have OVP that take focus.
        Gui::MDIView* mdi = Gui::Application::Instance->activeDocument()->getActiveView();
        mdi->setFocus();

        doConstructionMethodChanged();  // NVI

        handler->updateCursor();

        handler->reset();  // reset of handler to restart.

        handler->mouseMove(prevCursorPosition);
    }
```

For bug 1 I'll wait your feedback if this 
```
        if (init) {
            handler->moveCursorToSketchPoint(prevCursorPosition);
```
is necessary

PaddleStroke:
The focus lost can apparantly happen in other situation. So I instead suggest giving focus during reset like so : 

    /** @brief Resets the controls, such as the widget and the on-view parameters */
    void resetControls()
    {
        // Make sure we do not loose focus if next methode does not have OVP that take focus.
        Gui::MDIView* mdi = Gui::Application::Instance->activeDocument()->getActiveView();
        mdi->setFocus();

        doResetControls();  // NVI

        firstMoveInit = false;
    }

abdullahtahiriyo:
The culprit of the jump is indeed this:
``` handler->moveCursorToSketchPoint(prevCursorPosition); ```

This is what I was referring as a hack.

I have tried setting the focus as you indicate (mdi->setFocus). However, it will not work if you you are in a second step of construction, hit ESC once (it resets). Now, without moving the mouse press ESC again. Then the handler does not exist.

So. The problem solved by this hack is that of ensuring the ViewProviderSketch gets the keys (second ESC) after having pressed ESC once, and without moving the mouse.

The line that introduces the problem (of no detecting the second ESC), is the execution of: ``` initNOnViewParameters(nOnViewParameter);``` in resetOnViewParameters. Just comment this line off (for example with an if !init) and the problem is gone... only that if the construction mode changed to a different number of OVPs... well that is the problem...



```

abdullahtahiriyo:
I will try to give it another run to finding an alternative solution that works that is not the hack. I had it in my list anyway. If in the meantime you happen to have a solution that does not have the double ESC pressing issue, let me know.

PaddleStroke:
> I have tried setting the focus as you indicate (mdi->setFocus). However, it will not work if you you are in a second step of construction, hit ESC once (it resets). Now, without moving the mouse press ESC again. Then the handler does not exist.

I cannot reproduce this problem. I tried what you describe. So start line tool (with only dimensional for example). Select first point. Move mouse (or not). Press ESC -> the tool reset. Press ESC again without moving -> tool is dismissed. Exactly as I expect.

Perhaps try my PR to see if it works now?

PR #11312, opened 2023-11-07
Correct #9377 issue with accuracy in AdditiveHelix

### LINKED ISSUE #9377: [Problem] PartDesign: Helix insufficient precision

### Forums discussion

_No response_

### Version

0.20 (Release)

### Full version info

```shell
[code]
Word size of FreeCAD: 64-bit
Build type: Release
Branch: tag: 0.20.2
Installed mods: 
  * fasteners 0.4.56
  * Defeaturing 1.2.0
  * Assembly4 0.12.7
  * Dracula 0.0.4
  * A2plus 0.4.60k
  * Behave-Dark-Colors 0.1.1
[/code]
```

### Subproject(s) affected?

PartDesign

### Issue description

expected => one solid 

### Anything else?

_No response_

### Code of Conduct

bgbsww:
After much instrumenting and trial and error, it appears we set the wrong mode when the angle is zero.  Not clear if there was some past version of the library that needed the special case removed here.  Also cleaned up a little unneeded code.

This should fix #9377

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -227,37 +227,34 @@ App::DocumentObjectExecReturn* Helix::execute()
             return new App::DocumentObjectExecReturn(e.what());
         }
 
-        std::vector<std::vector<TopoDS_Wire>> wiresections;
-        for (TopoDS_Wire& wire : wires)
-            wiresections.emplace_back(1, wire);
-
         //build all shells
+        BRepOffsetAPI_MakePipeShell mkPS(TopoDS::Wire(path));
+
+        mkPS.SetTolerance(Precision::Confusion());
+        mkPS.SetTransitionMode(BRepBuilderAPI_Transformed);
+
+        mkPS.SetMode(TopoDS::Wire(auxpath), true);  // this is for auxiliary
+
+        for (TopoDS_Wire& wire : wires) {
+            wire.Move(invObjLoc);
+            mkPS.Add(wire);
+        }
+
+        // mkPS.MakeSolid();    // This would be great. but throws an exception
+
+        if (!mkPS.IsReady())
+            return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Error: Could not build"));
+
+        // Alternative to mkPS.MakeSolid()
         std::vector<TopoDS_Shape> shells;
         std::vector<TopoDS_Wire> frontwires, backwires;
-        for (std::vector<TopoDS_Wire>& wires : wiresections) {
 
-            BRepOffsetAPI_MakePipeShell mkPS(TopoDS::Wire(path));
+        shells.push_back(mkPS.Shape());
 
-            mkPS.SetTolerance(Precision::Confusion());
-            mkPS.SetTransitionMode(BRepBuilderAPI_Transformed);
-
-            mkPS.SetMode(TopoDS::Wire(auxpath), true);  // this is for auxiliary
-
-            for (TopoDS_Wire& wire : wires) {
-                wire.Move(invObjLoc);
-                mkPS.Add(wire);
-            }
-
-            if (!mkPS.IsReady())
-                return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Error: Could not build"));
-
-            shells.push_back(mkPS.Shape());
-
-            if (!mkPS.Shape().Closed()) {
-                // shell is not closed - get the end wires
-                frontwires.push_back(TopoDS::Wire(mkPS.FirstShape()));
-                backwires.push_back(TopoDS::Wire(mkPS.LastShape()));
-            }
+        if (!mkPS.Shape().Closed()) {
+            // shell is not closed - get the end wires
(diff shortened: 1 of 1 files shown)


src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -240,52 +240,15 @@ App::DocumentObjectExecReturn* Helix::execute()
             mkPS.Add(wire);
         }
 
-        // mkPS.MakeSolid();    // This would be great. but throws an exception
+        mkPS.Build();
```
FlachyJoe, line 243: Should be after IsReady check.

luzpaz: This seems resolved.

sliptonic:
@adrianinsaval  Any update?

bgbsww:
@FlachyJoe Thanks for that, slight tweaking and testing and I think we're there.  I think I learned more about this code base from this than anything else.

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -227,75 +227,42 @@ App::DocumentObjectExecReturn* Helix::execute()
...
-            mkSolid.Add(TopoDS::Shell(sewer.SewedShape()));
-        }
-        else {
-            // shells are already closed - add them directly
-            for (TopoDS_Shape& s : shells) {
-                mkSolid.Add(TopoDS::Shell(s));
-            }
+        //check for error value at pipe creation and re-build if needed
+        if (Angle.getValue() == 0 && mkPS.ErrorOnSurface() < Precision::Confusion() / 2.0 ) {
+            Base::Console().Log("PartDesign_Helix : Fail back to auxiliary mode\n");
```
adrianinsaval, line 257: English is not my native language but shouldn't this be "Fall back"?

adrianinsaval:
> I haven't dug into the calculation of ErrorOnSurface but it seems the problem starts when this is half the confusion.

do you think it's worth investing time in fine tuning this threshold?

bgbsww:
My opinion is no:  this just chooses between the faster and the slower calculation, and the vast majority of the time it will take the fast path.  So fine tuning this might improve the speed of a tiny percentage of examples, but @FlachyJoe has already gotten us very close.

FlachyJoe:
@wwmayer It works well on my side, can you please give a test file?

bgbsww:
Ah!  Thanks, I'll get on it.

wwmayer:
Wait! I quickly debugged the code and it seems to be a bit more complex.

The method getVerifiedFace() seems to generate the face but I don't see where this face is further used. There is only a check whether it's planar but nothing else.

And the logic with the getProfileWires() looks wrong to me, too. It returns a vector of two wires (the circles I guess) and both are added to BRepOffsetAPI_MakePipeShell.

FlachyJoe:
> I don't see how this is supposed to work for you

I tested without the commit merged, I thought my yesterday evening main branch include it. :roll_eyes:

FlachyJoe:
@wwmayer @bgbsww  `front` and `back` are only end faces.
We can restore the feature by

- create shells and sew them as before but we have to find another accuracy fix
- use a face from `Part::FaceMakerCheese::makeFace` and create a pipe from it. See [BRepOffsetAPI_MakePipe](https://dev.opencascade.org/doc/refman/html/class_b_rep_offset_a_p_i___make_pipe.html)

bgbsww:
I think you are saying one or the other here.  Adding the accuracy fix to complete #1 is I think as simple as the check and correct code you gave placed inside the original shells code:

```
      //check for error value at pipe creation and re-build if needed
        if (Angle.getValue() == 0 && mkPS.ErrorOnSurface() < Precision::Confusion() / 2.0 ) {
            Base::Console().Log("PartDesign_Helix : Fail back to auxiliary mode\n");
            mkPS.SetMode(TopoDS::Wire(auxpath), true);
            mkPS.Build();
        }
```
If you look at my initial single line commit, that solved the problem.  Then I got myself in trouble cleaning code that belonged there.

The second option is intriguing and potentially elegant,

bgbsww:
So, I am able to get this mostly working with the proposed option above.

There's a race condition somehow on `            mkSolid.Add(TopoDS::Shell(sewer.SewedShape()));
`.  Can get it to work in the debugger going slow, but it fails at full speed with a  `AdditiveHelix003: TopoDS::Shell`

However, after I get this working, if I propose a commit it's going to screw up the blame history for most of the code here.  Do we revert?  Is that normal in this project?  I have a path forward to fix the original problem after a revert if that makes sense.

bgbsww:
Thanks for that, hang on, because I've implemented option 2, which would get rid of all that code.  I'll advise soon.

bgbsww:
@FlachyJoe  care to comment on this WIP: https://github.com/FreeCAD/FreeCAD/compare/main...bgbsww:FreeCAD:bgbsww-patch-16?expand=1

I've implemented both solutions, but the one using the classic code is really sensitive to the height, and can easily be caused to fail.  The one using BRepOffsetAPI_MakePipe  I'm clearly missing something about adjusting the locations.  The results of generateHelixPath seem to be in a strange place regardless, but that may just be due to my hacked up DebAddObj to make them visible.

PR #11315, opened 2023-11-08
[TechDraw] Fixes #6527 - Implement face transparency

### LINKED ISSUE #6527: [Feature Request] TechDraw: Move `TransparentFace` setting to Page/View level

### Forums discussion

https://www.forum.freecadweb.org/viewtopic.php?p=577504#p577504

### Subproject(s) affected?

Techdraw

### Idea description

Transparent Faces settings is current global scope.  It would be convenient to be able to specify this at the Page or View level.

### Anything else?

_No response_

### Code of Conduct

pavltom:
This pull request implements #6527 with the possibility to set color and transparency of faces per view. The default color and transparency are obtained from TechDraw Preferences - Color - (Transparent) Faces setting. On top of that, also the color and transparency of the section cut surface can be set independently of the color of the faces.

For transparency values 0 (opaque) and 100 (invisible) everything works perfectly. However because the transparency can be set as a percentage (identically what the 3D view allows), there are some issues, which can not be easily resolved. Let's inspect the figure below, all views have faces color set to blue rgb(0, 0, 255) and transparency to 50%:

Transparency
-  View1 is drawn as desired and expected. There are no problems with single face views.

- View2 consists of 3 faces. Left, right and the "big" one hidden underneath, covering the area of both mentioned before. The color is darker than for View1, because left and right faces are drawn on top of the big face. To get the same look as  View1 has, we would need to increase the transparency to values around 70%.

- View3 is a view of a thick frame. It consists of 2 faces, the large one (outer border) and the small one (inner border). Because the small face is drawn on the top of the large face, the inner square is darker than the outer one. Unfortunately, there is no way around this, as long as the large face covers the whole area of the inner face.

With regards to these limitations, I believe the pull request overall increases the TechDraw abilities and the benefits prevail the known issues.

If you find any other problem than the one discussed above, please let me know.

### DIFF
--- a/src/Mod/TechDraw/Gui/QGIFace.cpp
+++ b/src/Mod/TechDraw/Gui/QGIFace.cpp
@@ -80,21 +80,15 @@ QGIFace::QGIFace(int index) :
     getParameters();
 
     // set up style & colour defaults
-    App::Color temp {static_cast<uint32_t>(Preferences::getPreferenceGroup("Colors")->GetUnsigned("FaceColor",0xffffffff))};
-    setFillColor(temp.asValue<QColor>());
-    m_colDefFill = temp.asValue<QColor>();
+    m_colDefFill = App::Color(static_cast<uint32_t>(Preferences::getPreferenceGroup("Colors")->GetUnsigned("FaceColor", 0xFFFFFF)))
+                   .asValue<QColor>();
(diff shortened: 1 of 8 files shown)


luzpaz:
Thank you for the contribution! 
Note: Please add 'Fix' or 'Fixes' before the issue number you're fixing so github auto-links and auto-closes said issue when merge is made. Ref: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #11316, opened 2023-11-08
[Sketcher] Add a visibility setting for On-view-parameter.

PaddleStroke:
Add a visibility setting for On-view-parameter.

Also fix focus lost bug introduced before.

@abdullahtahiriyo please confirm if removing this is ok  : 
```
        if (init) {
            handler->moveCursorToSketchPoint(prevCursorPosition);
       }
```
I cannot find what it does, and it bugs as moveCursorToSketchPoint doesn't take windows OS scaling. Which is also a bug to fix.

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
@@ -193,7 +193,10 @@ private:
 
     bool isWidgetVisible() const override
     {
-        return true;
+        ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath(
+            "User parameter:BaseApp/Preferences/Mod/Sketcher/Tools");
+        int index = hGrp->GetInt("OnViewParameterVisibility", 1);
+        return index != 0;
     };
 
     QPixmap getToolIcon() const override


abdullahtahiriyo:
> I cannot find what it does, and it bugs as moveCursorToSketchPoint doesn't take windows OS scaling. Which is also a bug to fix.

That may explain why it worked here. I do not use windows OS. Do you think you can fix that function? (as part of a separate PR)

abdullahtahiriyo:
I can confirm that with the mdi and setFocus it works in Linux... I am unsure why it was not working before...

PaddleStroke:
I tried to make it work actually. But it may be a bigger issue than just fixing the function.

> When dealing with high-DPI displays or displays with scaling factors other than 100% on Windows, you must account for the scaling factor in your calculations. In Qt, the QGuiApplication class provides functionality to handle high-DPI displays.
> 
> The scaling factor can be obtained from the devicePixelRatio function of the screen where the application's window is located, or globally from QGuiApplication if you want to use the primary screen's scaling factor.

https://doc.qt.io/qt-5/highdpi.html
I can give it a shot but it may be harder than we think to solve

abdullahtahiriyo:
> Also read about it in : https://doc.qt.io/qt-5/highdpi.html
> I can give it a shot but it may be harder than we think to solve

I think the function may be handy, but I think it is no longer used by the framework. So, if you can fix it in little time that should be preferred. If you see it is giving problems, then we can remove it (if not in use).

PR #11321, opened 2023-11-08
Take placement into account with applying Part Design Boolean operations

### LINKED ISSUE #10707: [Problem] PartDesign - When placement is changed in a body, it fails in any boolean operations with other bodies

### Problem description

I was creating a laminate with multiple cloned bodies in part design. Each body needed to be stacked against the prior, so I set the x axis on each subsequent clone. All operations seemed to work fine moving forwards, except boolean operations. They did all sorts of random things. The workaround was to change the placement variations in the clone objects themselves. 

### Full version info

```shell
[code]
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.21.1.33668 +26 (Git)
Build type: Release
Locale: English/Australia (en_AU)
Installed mods: 
  * Assembly4 0.50.2
  * Curves 0.6.13
[/code]
```
[body_position_boolean_fail.zip](https://github.com/FreeCAD/FreeCAD/files/12707888/body_position_boolean_fail.zip)

### Subproject(s) affected?

PartDesign

### Anything else?

_No response_

### Code of Conduct

bgbsww:
Fix #10707

### DIFF
--- a/src/Mod/PartDesign/App/FeatureBoolean.cpp
+++ b/src/Mod/PartDesign/App/FeatureBoolean.cpp
@@ -106,15 +106,14 @@ App::DocumentObjectExecReturn *Boolean::execute()
 
     TopoDS_Shape result = baseTopShape.getShape();
 
-    Base::Placement  sp = baseBody->globalPlacement();
-    sp.invert();
+    Base::Placement  bodyPlacement = baseBody->globalPlacement().inverse();
     for (auto tool : tools)
     {
         if(!tool->isDerivedFrom(Part::Feature::getClassTypeId()))
             return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Cannot do boolean with anything but Part::Feature and its derivatives"));
 
         Part::TopoShape toolShape = static_cast<Part::Feature*>(tool)->Shape.getShape();
-        toolShape.setPlacement(sp *toolShape.getPlacement());
+        toolShape.setPlacement(bodyPlacement * toolShape.getPlacement());
         TopoDS_Shape shape = toolShape.getShape();
         TopoDS_Shape boolOp;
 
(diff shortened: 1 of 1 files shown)


FlachyJoe:
So, this do what is announced but breaks the historical functioning.
Another way to solve the problem in new doc could be to enforce Binder usage with Boolean operations i.e. when a body is selected for boolean then auto create a binder of it and use this as tool. Binder placement still needs to be moved by the Base inverted placement.

luzpaz:
CC @yorikvanhavre can you please add @FlachyJoe as a reviewer ?

adrianinsaval:
> So, this do what is announced but breaks the historical functioning. Another way to solve the problem in new doc could be to enforce Binder usage with Boolean operations i.e. when a body is selected for boolean then auto create a binder of it and use this as tool. Binder placement still needs to be moved by the Base inverted placement.

I think right now binders don't work well as tools for a boolean OP but I agree this is the best way to go. Have proper support for subshapebinders and auto generate as required.

adrianinsaval:
@bgbsww we can't merge this if it breaks backwards compatibility

bgbsww:
Thanks, @adrianinsaval I understand.  I wonder how many projects are affected, and if it is very small, maybe flag those projects when they are opened and offer to autofix them with a binder solution.  That way the behavior going forward is correct.

I might be missing something here about a group of use cases, but my overall thought is that Part Design should be simplified with a clean object tree, and the proposed solution seems to move in the opposite direction.  https://github.com/FreeCAD/FreeCAD/issues/10707#issuecomment-1806193821

I'm not wedded to this PR or solution, but I can't see a good path forward yet.

mwganson:
One way potentially to work around breaking existing models is to add a new boolean property, something like UseToolPlacements or whatever, default = False (the current behavior).  Then if the user toggles this to True he gets the new behavior.

bgbsww:
Thank you both, I'll give this a try.

yorikvanhavre:
Sorry I missed the mention above... I now added @FlachyJoe to the developers team, who are able to review PRs.

yorikvanhavre:
Ok seems good to go then!

luzpaz:
@yorikvanhavre i can't add him to the Reviewers category in the sidebar

FlachyJoe:
@yorikvanhavre thank you for this unexpected advancement:exclamation:

Issue #11200, opened 2023-10-27
[Regression] Mesh WB does not export Links to STL or DAE


bjnobre:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

I've realized any 2nd links to a part on assembly4 don't be exported. See it bellow.

source:
![pen-full-source](https://github.com/FreeCAD/FreeCAD/assets/20462223/5af8cd0c-6123-47ad-8c98-4a5457f58728)
output:
![pen-stl](https://github.com/FreeCAD/FreeCAD/assets/20462223/a8921834-015e-49ca-a537-f1c5c9ec4104)

I've tried also on every other formats, and the same issue occurs. I've tried with another project, with more simple shapes and the the errors still there.

souce:
![a4-error-1](https://github.com/FreeCAD/FreeCAD/assets/20462223/55795cd4-933d-44ea-9572-bfc73efca4fd)
output:
![a4-error-2](https://github.com/FreeCAD/FreeCAD/assets/20462223/af28889d-fdf4-4fd5-a4d8-59cd50084390)

Even on draft 2d drawings, linked objects are not exported:

source:
![image](https://github.com/FreeCAD/FreeCAD/assets/20462223/ff53c2cd-b977-47f4-b9e4-14f5a2f3e48f)

output:
![image](https://github.com/FreeCAD/FreeCAD/assets/20462223/255cd899-1047-49b4-844a-9883e09a4bb5)

So, I assume that it's a issue with exporting module. I am using 0.22 dev version, but it occurs on 0.21 version.

### Full version info

```shell
[code]
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34789 (Git)
Build type: Release
Branch: main
Hash: 41503d63ebc3de96d156e6eb0cdb889ebb150bd4
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: Portuguese/Brazil (pt_BR)
Installed mods: 
  * Assembly4 0.50.2
  * fasteners 0.4.62
  * sheetmetal 0.2.63
[/code]
```

### Subproject(s) affected?

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
Have you reported to Assmebly4 repo ?  
Can you also include the the file (zip it and drag-drop in to gitihub textbox)

Roy-043:
[...]

PR #11322, opened 2023-11-08
Fix #10545 #11200 by using position of array members

### LINKED ISSUE #10545: [Problem] Exporting a link array object gives an incorrect result

### Problem description

When exporting a part made with the draft array tool to either stl or 3mf format, upon importing the object into another tool it's as if the array tool was not applied at all. 

### Full version info

```shell
[code]
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Build type: Release
Branch: (HEAD detached at 0.21)
Installed mods: 
  * Assembly4 0.50.2
[/code]
```

### Subproject(s) affected?

Core

### Anything else?

Applying a compound or connect operation after the array tool then exporting gives the result I would expect from just exporting the array object. Importing an exported 3mf of an array object into prusa slicer shows the array subelements are placed directly on top of each other, so possibly the array intervals are not being applied on export to the subelements?

Test case of a 3x3 array of cubes, exporting the part called "failure" gives an incorrect result, the two "success" parts work as expected.
[arrayexportbug.zip](https://github.com/FreeCAD/FreeCAD/files/12517045/arrayexportbug.zip)

Result of importing the failure and success parts exported as 3mf into prusa slicer, with the 9 stacked failure subelements selected.
image

Result when importing exported stl's for the same parts.
image

### Code of Conduct

bgbsww:
Fix #10545

### DIFF
--- a/src/Mod/Mesh/App/Exporter.cpp
+++ b/src/Mod/Mesh/App/Exporter.cpp
@@ -134,8 +134,10 @@ int Exporter::addObject(App::DocumentObject* obj, float tol)
                 }
                 Py_DECREF(pyobj);
             }
-        } else if ( it->second.getTransform() != matrix )
+        }
+        else if (it->second.getTransform() != matrix) {
             it->second.setTransform(matrix);
+        }
 
         // Add a new mesh
         if (it != meshCache.end()) {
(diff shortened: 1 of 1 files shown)


bgbsww:
Looks like #11200 might also be related?

bgbsww:
Fix #11200

adrianinsaval:
@wwmayer could you check this? if the PR is right then I'd like to merge and backport before the point release

wwmayer:
I can confirm that it fixes the issue.

luzpaz:
Thank you @bgbsww !

bjnobre:
What version will the fix be available in?

PR #11323, opened 2023-11-08
JtReader: allow to open JT files using TKJT

wwmayer:


maxwxyz:
@wwmayer do I need to have something else installed? I can neither open nor import .jt files, not through the dialog or dropping them in the FC window. The .jt extension is also not selectable in the file dialog.

```
OS: macOS 13.6.1
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34982 (Git)
Build type: Release
Branch: main
Hash: 55639df57325290e5a9d98ef48d6e17898fe0fc1
Python 3.11.0, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Installed mods: 
  * 3DfindIT 1.2.0
  * freecad.gears 1.0.0
  * OpenDark 2023.10.30
  * CfdOF 1.24.8
  * Defeaturing 1.2.1
  * Manipulator 1.5.0
  * fasteners 0.4.70
  * CurvedShapes 1.0.5
  * sheetmetal 0.3.12
  * 3D_Printing_Tools
  * Curves 0.6.14
```

wwmayer:
You have to build this library first: https://github.com/wwmayer/TKJT
Then you must re-run CMake for FreeCAD to manually set the include directory and library to the TKJT stuff. After that the source file that implements JT support will be added to the JtReader target.

Hint: Because TKJT is under the GPL it won't be added to FreeCAD.

PR #11327, opened 2023-11-09
[Sketcher] OVP DSH : setup the cursor texts.

PaddleStroke:
If OVP are disabled, the cursor text is enabled.
If only dimensional OVP are disabled, only dimensional cursor texts are disabled.
@abdullahtahiriyo I think with this we addressed all the issues.

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchControllableHandler.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchControllableHandler.h
@@ -102,6 +102,17 @@ public:
         DSDefaultHandler::finish();
         return true;
     }
+
+    bool shouldDrawPositionAtCursor() const override
+    {
+        return toolWidgetManager.shouldDrawPositionAtCursor();
+    }
+
+    bool shouldDrawDimensionsAtCursor() const override
+    {
+        return toolWidgetManager.shouldDrawDimensionsAtCursor();
+    }
+
     //@}
 
 
@@ -159,6 +170,7 @@ private:
 
     void onModeChanged() override
     {
+        DSDefaultHandler::resetPositionText();
         toolWidgetManager.onHandlerModeChanged();
         DSDefaultHandler::onModeChanged();
 
--- a/src/Mod/Sketcher/Gui/DrawSketchController.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchController.h
@@ -229,6 +229,8 @@ protected:
             dynamicOverride = false;
         }
 
+        OnViewParameterVisibility onViewParameterVisibility;
+
     private:
         void init()
         {
@@ -239,7 +241,6 @@ protected:
                 static_cast<OnViewParameterVisibility>(hGrp->GetInt("OnViewParameterVisibility"));
         }
 
-        OnViewParameterVisibility onViewParameterVisibility;
         bool dynamicOverride = false;
     };
 
@@ -450,6 +451,18 @@ public:
         if (!handler->isState(SelectModeT::End) || handler->continuousMode) {
             handler->mouseMove(prevCursorPosition);
         }
+    }
+
+    bool shouldDrawPositionAtCursor() const
+    {
+        return !(ovpVisibilityManager.onViewParameterVisibility
+                 == OnViewParameterVisibilityManager::OnViewParameterVisibility::ShowAll);
+    }
+
+    bool shouldDrawDimensionsAtCursor() const
+    {
+        return (ovpVisibilityManager.onViewParameterVisibility
+                == OnViewParameterVisibilityManager::OnViewParameterVisibility::Hidden);
     }
 
 protected:
@@ -686,9 +699,10 @@ protected:
     }
     //@}
 
+
 private:
-    ColorManager colorManager;
     OnViewParameterVisibilityManager ovpVisibilityManager;
+    ColorManager colorManager;
     std::unique_ptr<DrawSketchKeyboardManager> keymanager;
 };
 
--- a/src/Mod/Sketcher/Gui/DrawSketchHandler.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandler.h
@@ -225,6 +225,15 @@ private:  // NVI
     virtual void onWidgetChanged()
     {}
 
+    virtual bool shouldDrawPositionAtCursor() const
+    {
+        return true;
+    }
+    virtual bool shouldDrawDimensionsAtCursor() const
+    {
+        return true;
+    }
+
 protected:  // NVI requiring base implementation
     virtual std::string getToolName() const;
     virtual QString getCrosshairCursorSVGName() const;
@@ -282,6 +291,9 @@ protected:
 
     void drawPositionAtCursor(const Base::Vector2d& position);
     void drawDirectionAtCursor(const Base::Vector2d& position, const Base::Vector2d& origin);
+    void
+    drawWidthHeightAtCursor(const Base::Vector2d& position, const double val1, const double val2);
+    void drawDoubleAtCursor(const Base::Vector2d& position, const double radius);
 
     int getPreselectPoint() const;
     int getPreselectCurve() const;
--- a/src/Mod/Sketcher/Gui/DrawSketchHandler.cpp
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandler.cpp
@@ -1138,22 +1138,53 @@ void DrawSketchHandler::clearEditMarkers() const
 
 void DrawSketchHandler::drawPositionAtCursor(const Base::Vector2d& position)
 {
-    setPositionText(position);
+    if (shouldDrawPositionAtCursor()) {
+        setPositionText(position);
+    }
 }
 
 void DrawSketchHandler::drawDirectionAtCursor(const Base::Vector2d& position,
                                               const Base::Vector2d& origin)
 {
+    if (!shouldDrawDimensionsAtCursor() || !showCursorCoords()) {
+        return;
+    }
+
     float length = (position - origin).Length();
     float angle = (position - origin).GetAngle(Base::Vector2d(1.f, 0.f));
 
-    if (showCursorCoords()) {
-        SbString text;
-        std::string lengthString = lengthToDisplayFormat(length, 1);
-        std::string angleString = angleToDisplayFormat(angle * 180.0 / M_PI, 1);
-        text.sprintf(" (%s, %s)", lengthString.c_str(), angleString.c_str());
-        setPositionText(position, text);
+    SbString text;
+    std::string lengthString = lengthToDisplayFormat(length, 1);
+    std::string angleString = angleToDisplayFormat(angle * 180.0 / M_PI, 1);
+    text.sprintf(" (%s, %s)", lengthString.c_str(), angleString.c_str());
+    setPositionText(position, text);
+}
+
+void DrawSketchHandler::drawWidthHeightAtCursor(const Base::Vector2d& position,
+                                                const double val1,
+                                                const double val2)
+{
+    if (!shouldDrawDimensionsAtCursor() || !showCursorCoords()) {
+        return;
     }
+
+    SbString text;
+    std::string val1String = lengthToDisplayFormat(val1, 1);
+    std::string val2String = lengthToDisplayFormat(val2, 1);
+    text.sprintf(" (%s x %s)", val1String.c_str(), val2String.c_str());
+    setPositionText(position, text);
+}
+
+void DrawSketchHandler::drawDoubleAtCursor(const Base::Vector2d& position, const double val)
+{
+    if (!shouldDrawDimensionsAtCursor() || !showCursorCoords()) {
+        return;
+    }
+
+    SbString text;
+    std::string doubleString = lengthToDisplayFormat(val, 1);
+    text.sprintf(" (%s)", doubleString.c_str());
+    setPositionText(position, text);
 }
 
 std::unique_ptr<QWidget> DrawSketchHandler::createToolWidget() const
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerLine.h
@@ -89,6 +89,8 @@ private:
     {
         switch (state()) {
             case SelectMode::SeekFirst: {
+                drawPositionAtCursor(onSketchPos);
+
                 startPoint = onSketchPos;
 
                 if (seekAutoConstraint(sugConstraints[0], onSketchPos, Base::Vector2d(0.f, 0.f))) {
@@ -97,6 +99,8 @@ private:
                 }
             } break;
             case SelectMode::SeekSecond: {
+                drawDirectionAtCursor(onSketchPos, startPoint);
+
                 endPoint = onSketchPos;
 
                 try {
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerOffset.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerOffset.h
@@ -132,6 +132,7 @@ private:
 
             if (!offsetLengthSet) {
                 findOffsetLength();
+                drawDoubleAtCursor(onSketchPos, offsetLength);
             }
 
             if (fabs(offsetLength) > Precision::Confusion()) {
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerPolygon.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerPolygon.h
@@ -78,6 +78,8 @@ private:
     {
         switch (state()) {
             case SelectMode::SeekFirst: {
+                drawPositionAtCursor(onSketchPos);
+
                 centerPoint = onSketchPos;
 
                 if (seekAutoConstraint(sugConstraints[0], onSketchPos, Base::Vector2d(0.f, 0.f))) {
@@ -86,6 +88,8 @@ private:
                 }
             } break;
             case SelectMode::SeekSecond: {
+                drawDirectionAtCursor(onSketchPos, centerPoint);
+
                 firstCorner = onSketchPos;
 
                 CreateAndDrawShapeGeometry();
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerRectangle.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerRectangle.h
@@ -193,6 +193,8 @@ private:
 
                 try {
                     CreateAndDrawShapeGeometry();
+
+                    drawWidthHeightAtCursor(onSketchPos, length, width);
                 }
                 catch (const Base::ValueError&) {
                 }  // equal points while hovering raise an objection that can be safely ignored
@@ -207,10 +209,11 @@ private:
                     || constructionMethod() == ConstructionMethod::CenterAndCorner) {
                     if (roundCorners) {
                         calculateRadius(onSketchPos);
+                        drawDoubleAtCursor(onSketchPos, radius);
                     }
-                    else {
-                        calculateThickness(
-                            onSketchPos);  // This is the case of frame of normal rectangle.
+                    else {  // Normal rectangle with frame.
+                        calculateThickness(onSketchPos);
+                        drawDoubleAtCursor(onSketchPos, thickness);
                     }
                 }
                 else if (constructionMethod() == ConstructionMethod::ThreePoints) {
@@ -242,6 +245,8 @@ private:
                     else {
                         radius = 0.;
                     }
+
+                    drawWidthHeightAtCursor(onSketchPos, length, width);
                 }
                 else {
                     corner2 = onSketchPos;
@@ -270,9 +275,7 @@ private:
                         radius = 0.;
                     }
 
-                    SbString text;
-                    text.sprintf(" (%.1f Angle)", angle412 / M_PI * 180);
-                    setPositionText(onSketchPos, text);
+                    drawWidthHeightAtCursor(onSketchPos, length, width);
                 }
 
                 try {
@@ -293,15 +296,17 @@ private:
             case SelectMode::SeekFourth: {
                 if (constructionMethod() == ConstructionMethod::Diagonal
                     || constructionMethod() == ConstructionMethod::CenterAndCorner) {
-                    calculateThickness(
-                        onSketchPos);  // This is the case of frame of round corner rectangle.
+                    calculateThickness(onSketchPos);
+                    drawDoubleAtCursor(onSketchPos, thickness);
                 }
                 else {
                     if (roundCorners) {
                         calculateRadius(onSketchPos);
+                        drawDoubleAtCursor(onSketchPos, radius);
                     }
                     else {
                         calculateThickness(onSketchPos);
+                        drawDoubleAtCursor(onSketchPos, thickness);
                     }
                 }
 
@@ -309,6 +314,7 @@ private:
             } break;
             case SelectMode::SeekFifth: {
                 calculateThickness(onSketchPos);
+                drawDoubleAtCursor(onSketchPos, thickness);
 
                 CreateAndDrawShapeGeometry();
             } break;


abdullahtahiriyo:
I have some remarks on this one. I will be writing in the code, but do not start modifying anything until I finish, as will be comment different choices and why one could be better than other.

src/Mod/Sketcher/Gui/DrawSketchController.h
```diff
@@ -239,7 +241,6 @@ class DrawSketchController
                 static_cast<OnViewParameterVisibility>(hGrp->GetInt("OnViewParameterVisibility"));
         }
 
-        OnViewParameterVisibility onViewParameterVisibility;
```
abdullahtahiriyo, line 242: Changing of access specifier should trigger some kind of alarm. It is rarely the right decision.

Here you could have used the visibility() member function to get the value, or made an:
```
        bool isVisibility(OnViewParameterVisibility visibility) const
        {
            return onViewParameterVisibility == visibility;
        }

```

The reason why this member is private, is to avoid that the visibility can be changed from outside this class. That is important because this class has the responsibility for managing the visibility. Allowing change from the outside is not proper encapsulation as it risks losing control over its responsibility.

You may argue that other (inner) classes, or even implementation only classes (classes defined within a source file, not in a header) have public data members. Yes this is true. Most often the reason for this is a mixture of laziness and perceived reduced associated risk. That is specially the case in implementation-only classes that often are POD alike.

However, when a member is already private, changing it to public for direct access has little justification, if any.

src/Mod/Sketcher/Gui/DrawSketchController.h
```diff
@@ -686,9 +699,10 @@ class DrawSketchController
     }
     //@}
 
+
 private:
-    ColorManager colorManager;
     OnViewParameterVisibilityManager ovpVisibilityManager;
+    ColorManager colorManager;
```
abdullahtahiriyo, line 705: Not sure why this change. It was alphabetically ordered... I do not see any initialisation issue...

src/Mod/Sketcher/Gui/DrawSketchController.h
```diff
@@ -452,6 +453,18 @@ class DrawSketchController
...
+        return !(ovpVisibilityManager.onViewParameterVisibility
+                 == OnViewParameterVisibilityManager::OnViewParameterVisibility::ShowAll);
+    }
+
+    bool shouldDrawDimensionsAtCursor() const
+    {
+        return (ovpVisibilityManager.onViewParameterVisibility
+                == OnViewParameterVisibilityManager::OnViewParameterVisibility::Hidden);
+    }
+
```
abdullahtahiriyo, line 467: This here is an option. It is not my preferred choice.

The rationale to have this here, is that the "logic" for the control of what to do  (show or not show) is with the controller.

An implication of this choice, is on the caller (DrawSketchControllableHandler). When the functionality is triggered from the Handler, by checking the controller (the present model here), a new interface is being imposed on any future controller that is written to work with DrawSketchControllableHandler. That includes controllers that may not have any constraint on whether to show or not a cursor.

I do not think it is inherently wrong. I have an alternative that I will comment on below.

src/Mod/Sketcher/Gui/DrawSketchControllableHandler.h
```diff
@@ -102,6 +102,17 @@ class DrawSketchControllableHandler
         DSDefaultHandler::finish();
         return true;
     }
+
+    bool shouldDrawPositionAtCursor() const override
+    {
+        return toolWidgetManager.shouldDrawPositionAtCursor();
```
abdullahtahiriyo, line 108: These two functions, is what I was referring to when saying imposing interface requirements on the controller. As said, not inherently wrong. Just commenting implications.

src/Mod/Sketcher/Gui/DrawSketchControllableHandler.h
```diff
@@ -102,6 +102,17 @@ class DrawSketchControllableHandler
         DSDefaultHandler::finish();
         return true;
     }
+
+    bool shouldDrawPositionAtCursor() const override
```
abdullahtahiriyo, line 106: This override is the result of a different decision: extending the interface so that every single handler can have a saying on drawing or not drawing. I will comment on this one on the source.

src/Mod/Sketcher/Gui/DrawSketchHandler.h
```diff
@@ -225,6 +225,15 @@ class SketcherGuiExport DrawSketchHandler
     virtual void onWidgetChanged()
     {}
 
+    virtual bool shouldDrawPositionAtCursor() const
```
abdullahtahiriyo, line 228: And here we are at the source.

Here the function offers an option to handlers to override the logic to draw or not to draw.

The functionality to draw is also offered by this class some lines below. It draws or not according to a preferences parameter.

The caller to draw is, ultimately the handler (e.g. DSHLine).

So DSHLine asks grandpa to draw (according to preferences), and uses a facility provided by granpa to override the logic for drawing... a logic that is available at the handler (via the controller which is a member of the handler).

A first approach that I think it is perhaps more straightforward, is to make the drawing functions below virtual, thereby allowing any handler to overload behaviour.

abdullahtahiriyo:
So fast forward. A first alternative could be this:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/tree/cursor_text_alternative1

abdullahtahiriyo:
That alternative essentially virtualises the drawing interface to be able to overload, and moves the decision to DrawSketchControllableHandler (to avoid having to override in each child, i.e. DrawSketchHandlerLine).

This still imposes an interface to the controller...

... because at the very end, we are only doing this because we have a controller with OVPs...
... so it is not even that the handler wants to override the behaviour of the drawing. The controllable handler is perfectly ok with the standard drawing, and does not have any knowledge of whether the controller uses OVPs, widgets, or VR glasses...

abdullahtahiriyo:
...then why not let the controller offer the kind of conditional drawing it wants:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/tree/cursor_text_alternative2

In this option the handlers call the interface provided by the controller for drawing:
```
toolWidgetManager.drawDoubleAtCursor(onSketchPos, radius);
```

Then the controller applies the logic it has, and when the cursor text should be printed, calls the standard functionality provided by its associated handler, which is what we already had in DrawSketchHandler...

Let me know what you think...

abdullahtahiriyo:
BTW, if you like alternative 2, I missed one toolwidget. call in line 102 of DSHLine.

If you have other handlers written, you can automatise the changes via sed.
```
sed -i 's| drawPositionAtCursor| toolWidgetManager.drawPositionAtCursor|g' DrawSketchHandlerLine.h
sed -i 's| drawDoubleAtCursor| toolWidgetManager.drawDoubleAtCursor|g' DrawSketchHandlerOffset.h
sed -i 's| drawDirectionAtCursor| toolWidgetManager.drawDirectionAtCursor|g' DrawSketchHandlerPolygon.h
sed -i 's| drawWidthHeightAtCursor| toolWidgetManager.drawWidthHeightAtCursor|g' DrawSketchHandlerRectangle.h
```

src/Mod/Sketcher/Gui/DrawSketchController.h
PaddleStroke: Ok.
But ultimately will there be any other controllers ? I think splitting DSH in many classes and subclasses for the sake of potential future handlers that could have different need is not really useful and counter productive. And I think there is also a coding holy principle about this even though I forgot the name, something like 'handle what is needed now and if the need appears later it will be adapted at that time'.
Besides sketcher is mostly mature. And the few tools that were missing are being added now and don't need a special controller.
So for me DSH, DSDH, DSCH should all be squashed. The architecture you build is probably very conceptually sound, but that is actually complex in the amount of files / subclasses / template etc and that only you understand clearly. I worked on it for more than one year and I'm often confused and lost between the files.

Here I'm a bit out of topic, but actually it is on topic. Because look with the complexity castle you build, my PR is not implementing exactly at the right place and so on. What is the likelyhood that a newcomer will be able to get in FC code and do what I've done correctly ? None. Even other fc devs would get it wrong I think.
So in the end this complexity is against the KISS principle (Keep it simple, stupid). Which in the end makes the code less accessible.

src/Mod/Sketcher/Gui/DrawSketchController.h
PaddleStroke: No use, at some point I wanted to make ovpVisibilityManager public so I moved it up, then thought making it public was probably not a good idea ;)

src/Mod/Sketcher/Gui/DrawSketchController.h
PaddleStroke: Not saying that your arguments are not sound here considering the current architecture. Just saying that the architecture is so complex that if I get it wrong after all the time I've been in it, this might be an indication that complexity is too high. And that we could probably reduce it.

PaddleStroke:
Having to write : 
`                toolWidgetManager.drawPositionAtCursor(onSketchPos);`
Instead of : 
`                drawPositionAtCursor(onSketchPos);`
Is for me annoying and clutter even more the code that is already really very cluttered.

PaddleStroke:
Anyway I have to move on to some Ondsel internal things now, I have already stretched my time on sketcher more than I was supposed to. So I'll let you do what you think best here. You can close my PR if it's not ok and merge you own.
I'll try to adapt the other PRs to whatever you end up merging on my spare time later.

src/Mod/Sketcher/Gui/DrawSketchController.h
abdullahtahiriyo: They are not principles for the sake of principles. It is proper definition of interfaces, which show the intend on how they should be used. Perhaps you wouldn't try to set the visibility from outside (now it is a read only from the outside). But perhaps someone else, less familiar with the code would, ignoring that this object is responsible. That person may even have a different need and given that the member is public, why not change the visibility from outside the class? It is more convenient... less code to write... If people start working like that, at the end they will be continuously fixing the code, as others will introduce behaviour that they do not expect.

It is the encapsulation boundary that tells different developers that the member is not intended for modification from the outside, and which marks that the object is responsible. Developers wanting to modify the visibility from the outside, will hit a compilation error, which should wake them up. Seeing themselves trying to move the member to have public access specifier, should serve as a waking up call that what they are doing is wrong. It is also much better detectable for a maintainer reviewing a PR.

The case where there is a setter and a getter, which you claim is read/write anyway, is also not that black and white. The main problem with a public data member is that it defines interface. Changing the behaviour of the class, requires changing all accesses to the data member. If one writes a getter and a setter from the outset, changing behaviour is transparent for client code.

I am not leaving these kind of comments to annoy you. I am trying to let you know the reasons that motivate one direction or the other. Programming is not difficult. Doing it right is hard.

Perhaps another line of argumentation: You should consider why on earth language designers, almost in every object oriented language (I know) came with the need for accessors and access specifiers, when it would be so easy to have everything public. Hopefully, you won't believe they were chosen to design a language, or their language was adopted by millions, just because they were clueless about what they were doing or "out of principles". If you want to dedicate professionally to this, you should start questioning why the language has the features it has.

PaddleStroke: > That person may even have a different need and given that the member is public, why not change the visibility from outside the class?

But again. Why would someone want to change this variable if he doesn't need to? If someone need to write this variable, he will create a getter and that's it.

> The main problem with a public data member is that it defines interface. Changing the behaviour of the class, requires changing all accesses to the data member. If one writes a getter and a setter from the outset, changing behaviour is transparent for client code.

Well this is true and so I see the point of setter/getter with this argument. Thanks for pointing it out.

src/Mod/Sketcher/Gui/DrawSketchController.h
abdullahtahiriyo: **Would there by any other controllers?** 

It is a valid question. 

You negative answer lies on the premise that no new tools will be added as the sketcher is mature. But you have added new DSHs haven't you? When the premise is that a framework does not need to be extended, there is no need to design it allow extension. Yet, what almost systematically ends up happening, is that software grows and needs be extended. When architecture does not support growth, the problem is that developers start hacking around to solve their immediate needs, which opens the can of worms.

An answer has to take into account how much sketcherdefaultwidget will be able to grow without becoming a Frankenstein (and I honestly think it is close). That class holds a bunch of generally used controls, which is not certainly all the controls possible. I would say the class is useful for writing many DSHs, but it does not scale to cover the widget needs of ANY DSH.

In other words, I do not think it is unthinkable that some DSHs would not be feasible without allowing to provide a custom widget.

Now the principle you were looking for is probably YAGNI (You ain't gonna need it). It is a possible line of argumentation. Then one should think the cost of splitting the functionality into classes, with separate responsibilities now, and the cost of doing it when the need arises. Given the complexity of the framework, I think it totally made sense to do it now.

**Framework complexity** 

Now, you are proposing to squash DrawSketchHandler, DrawSketchDefaultHandler and DrawSketchControllableHandler. Your premise is that putting all into one class, as opposed to separate them by responsibilities with an eye in extensibility, makes the code simpler. 

Let me answer you with one question: Why not write every handler from scratch? Just create a class DrawSketchHandlerLine and write all the code there. No code reuse. Rinse and repeat for DrawSketchHandlerCircle and the rest. Hopefully, you will say, hey no, I need code reuse. Fine.

The answer of why these classes exist separately is similar:
- DrawSketchHandler is the old framework, which allows to write any type of handler. Not only Geometry creation handlers.
- DrawSketchDefaultHandler is a new type of handler, which has a very specific structure useful to add geometry (along with constraints and autconstraints) and offering many advanced solver capabilities. This class is intended to be an improved DrawSketchHandler specialised for this type of handler. A person wanting to add a handler for geometry creation today, and not wanting OVPs or widgets, should derived from this one instead.
- DrawSketchControllableHandler was created because your specific need to have double input (mouse+keyboard), first for the parameters in the widget and then for the OVPs. This class is exactly the same as the previous one, but allows to exchange changes between handler and controller (for OVPs and Widget). The controller can be anything that meets its interface.
- DrawSketchController is the most basic controller supporting the interface of DrawSketchControllableHandler and provides the OVPs. A developer can derive from it if it wants a controller for OVPs, but no default widget, or a custom widget.
- DrawSketchDefaultWidgetController is a controller that in addition to OVPs uses the default widget for input.

You have all these classes so that the framework is flexible, and each of the classes is able to concentrate in one task. Some of this classes are templates to enable quite a lot of automation, while minimising typing.

So, imagine we do as you say, and make them all into one class. What would you be winning? Yes, you will have less amount of classes. But still you need to handle the default widget, and the OVPs and the dual input. So at the end, you do need all that code. You do not have a simpler framework. You do have a less flexible framework. You still need to deal with the same complexity, because no matter how you distribute the code, code needs to be maintained (framework) and DSH code needs to be provided.

In fact, what you should think is how much code are you saving everytime you write a new tool. Yes, files a bigger with the dual input... but do you think they would be smaller by keeping all the framework in a single class. No, they would not.

All the code of specialised functions is **because of the dual input and the widget**.

I went into a lot of thinking an trouble to handle this complexity of the dual input in a uniform way, and which as much code reuse as possible.

**Dealing with the complexity of the framework** 

When the functionality demanded to a framework is complex, no matter how much architecture you factor in, the framework will be complex. Paraphrasing Einstein roughly, **things should be as simple as possible, but not simpler**.

You claim you have problems to change the framework. Well, you have changed the framework, so you managed. You managed even if you do not appreciate its architecture... should not that speak for it being a good framework? After all, a complex framework that is fixable by a developer that is not the one who wrote it, should speak for its virtues, rather than it shortcomings.

In any case, the objective of a framework is not that is simpler than need be internally, but rather that it abstracts a client of the framework from the complexity it was written to handle. 

Hopefully, we will have over time developers in FreeCAD who appreciate and understand the need for an architecture. If we fail there, then it is guaranteed that FreeCAD will turn into an unmaintainable beast. This has happened before in many projects, including enterprise projects employing good programmers. That motivated people to write books about software architecture.

So my answer here is: We hope to provide a framework new developers can directly rely on for a vast variety of DSHs, and concentrate their efforts on the handlers themselves. If the framework does not adapt to a particular need, this requires a more experienced developer, or one who can understand the need for an architecture.

**Is it difficult to write DSHs using the framework?** 

This is a trap question. Difficult compared to what?. 

It is certainly more difficult to write a new DSH having OVPs and a widget than it was to write one of the legacy DSHs which have none of those facilities. However, to compare pears to pears, it is not more difficult to write a new handler without those facilities using DrawSketchDefaultHandler, and it provides a lot more facilities (advanced solver functionality, automatic drawing based on geometries rather than generating a sampled curve, ...), than using DrawSketchHandler (specially where there are examples).

Now, start deriving from DrawSketchDefaultHandler (or DrawSketchHandler if you so wish) and without using the controllable handler or the written controllers, try to write a class handling OVPs and widget. Is that much more complex? The answer can only be that it is indeed much more complex, because part of the complexity (a substantial amount of it) is in those classes. The result would also be more code without code reuse (less concise). More error prone. More code to maintain. Duplicated code over handlers...

To reply to your last post. I do not think this architecture is more complex that should be for the requirements imposed upon it. The words are not mine, but are very wise words (it may well be because they are not mine): **things should be as simple as possible, but not simpler.**

N.B.: BTW, I did not tell you your solution was abhorrent. I only told you why I would prefer a different one. I took the time to explain and even code alternatives, not because I have too much free time. But because I prefer you to understand the choices, so that your code gets better.

abdullahtahiriyo:
Next coding slot...

src/Mod/Sketcher/Gui/DrawSketchController.h
abdullahtahiriyo: > But again. Why would someone want to change this variable if he doesn't need to? If someone need to write this variable, he will create a getter and that's it.

If the member is public and somebody thinks it is a good idea to do it, for example because he wants to change visibility after some event happening in the controller. This person has an incentive to want to write to it, this person is not going to create a getter. This person is going to write to it directly. The consequence is that the control logic stops being encapsulated in the class. The problem with losing encapsulation, is that the class starts losing control about when visibility is changed. So the automation mechanism embedded in it may fail. The decision to write to it is most likely a bad decision. But the interface of the class does not show intent. Without showing intent, the probability that another developer makes a mistake is much higher.

It generally goes like this: developer A created the class and incorporated the logic in the class but was lazy to write the getter/setter and put the data member public. Developer B needs to enable visualisation at some event. Instead of incorporating the logic into the class (for example, by creating a function that convey this persons intend and takes into account, within the class how the logic needs to be adapted to take into account this event), because the data member is public, writes to it. Reviewer C does not realise the mistake (writing to a public data member is very much normal). Visualisation works at times. Developer D in bug fixing mode may remove the code of developer B, or may even write to the variable at other code chunk outside the class to "fix" the bug (or make it not reproducible anymore, but the bug arising at other circumstances, as it is not fixed, the bug is controlling directly from outside without accounting for the logic already in).

All this does not even start if there is a getter. A developer should think twice before writing a setter for a private data member (at least should look very well into the class and the embedded logic). A reviewer is way more likely to realise the issue of a setter being added, or a variable changing access specifier, than when it is just writing to a public variable. 

The biggest metaphorical pan to hit the head of a developer and wake him wake up is a compiler error. That lack of direct accessibility that encapsulation provides is the metaphorical pan.

src/Mod/Sketcher/Gui/DrawSketchController.h
abdullahtahiriyo: > Writing a DSH with the current architecture is not a problem. It's more debugging/changing the architecture that is harder. Like when offset was crashing, it felt like a maze.

A good framework makes the life of the client code the easiest possible. A good framework needs to manage its complexity in the most simple way, but cannot be simpler than the minimum required by the requirements imposed upon it by design.

Yes, I agree that debugging the internals of a good framework which manages a complex task is not easy. The motivation to debug and fix any bug is that, when it is fixed, it will work, and using it will be very satisfactory. The extra motivation is that by leveraging complexity with the framework, you fix it once for all the tools using the framework.

> Ok so maybe not the base DSH, but why then DrawSketchControllableHandler and DrawSketchDefaultHandler. If a tool don't need widget/ovp, why couldn't it just use DrawSketchControllableHandler and declare controller null. That would remove one layer.

DrawSketchControllableHandler defines an interface of the controller, which is a template type. To implement what you suggest, you would need to create a controller that has the interface required by that class (and maintain it, so every change to the interface, needs a change to that fake controller), and make the fake controller do nothing it. Having code doing nothing does not appear to be removing complexity. At the same time, you had to add a new class (fake controller) to remove another class (merging two into one). So not even the number of classes decreases (which in any case is a faux measure of complexity).

There are several reasons why this is not a good idea. But perhaps it is best to try to understand why you think it is a good idea. Does removing one class in the hierarchy make dealing with the complexity any simpler? Why? It is the same amount of code and it is intermixed in a single class, as opposed to organised into areas of responsibility. The only thing that the controllable handler does is deal with the signaling necessary to interface with the controller. Anything else you look into daddy class... or grandpa class.

A first reason why this is a bad idea, is that in a hierarchy of classes, each derived class goes developing an extra aspect, and holds the responsibility for managing this aspect. This should help in the sense that if you want to fix a given functionality, you go to the class holding the responsibility. If you need to review/fix/extend a given functionality and you have merged ten classes into one, you will end up reading ten times more code. This is a maintainability killer.

A second reason why I think it is a very bad idea to merge these two classes, is because if a developer wants to make a simple DSH with the new architecture (without OVPs and widgets), and looks into this merged parent, this person will see much more code than actually needs see... and for nothing. Moreover, this person will be forced to create a dummy controller and assign it... which is, at best, odd. This class shows a very bad intent/interface to the user. The objective should be to make the framework the easiest possible to use by the user, while encapsulating complexity. Forcing the user to add a dummy controller is at odds with this goal.

A third reason, is that by doing so you remove flexibility and extensibility from the framework. We are now in a situation where we are extending DSDefaultHandler to have OVPs and widgets. If anybody would have a different need in the future, with a single class, this person would be forced to carry on with all that controller specific code, even if this person does not want it. While developing this need, this person will have to debug too, and it is unlikely he or she would appreciate seeing framework code that makes no sense whatsoever. This is making the person deal with more complexity than really needed.

There are probably other reasons, but I have been too long awake...

abdullahtahiriyo:
No. I mean that at that time I had to return to my real job, and that I would process this later on. That will be tomorrow...

src/Mod/Sketcher/Gui/DrawSketchController.h
PaddleStroke: Thanks for the example it makes things clearer.

PaddleStroke:
Ah understood.
Note that I am not annoyed or angry or anything. I actually had to start working on other things so I mentioned you can close/merge as you want, that it's ok for me. As you already have you branch ready it seemed pretty straight forward.

src/Mod/Sketcher/Gui/DrawSketchController.h
PaddleStroke: I understand thanks. I think I have just been frustrated by the complexity when debugging especially considering the number of time I've done so. It made me feel like there is a hundred OnReset function each overridding each other. But yes squashing classes will probably not help, its wishful thinking.
As you said, its making the framework which is understandably hard. While writing dsh is indeed easy enough.

abdullahtahiriyo:
You have questions. You need answers. Providing answers is in the interest of the project. If there are decisions you do not understand, you are not going to do differently in the future. There are things you do just fine and others were you need to understand why we do things differently. This is not politics. Here real reasons are important, and you will win many discussions with good arguments. 

For the record, I still considered another option to have a simpler calling:
DrawSketchControllableHandler will have new functions with different name as DrawSketchHandler, which would redirect to the controller version. This enables to call drawDirectionAtCursor2() without ```toolwidget.```. However, this IMO causes confusion in the interface because of different functions for substantially the same thing. Things can improve with more verbose naming (of course drawDirectionAtCursor2() could be called drawDirectionAtCursorConditionally()), but that would not make it much shorter... still it begs the question, conditionally to what?

I think that is already a very nuanced difference. Both solutions are acceptable and we could certainly entertain ourselves which more productive endeavours, so I just merged alt2 with corrected DSH function calls.


Ok. Let's move on to the next PR.

Issue #11208, opened 2023-10-27
[Problem] Sketcher toolbars are cluttered


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Sketcher toolbars are overcrowded.

Here I am compiling a list of ideas to improve the situation : 
1 - Make a command group for 'Split', 'Extend' and 'Trim'.
2 - Make a command group called 'Shapes' for 'Rectangle' 'Slot' and 'Polygon. (Note right now we have 3 rectangle tools and many polygons, but with currently opened #11174 we can have only 1 tool for each.)
3 - Rendering order could go inside the setting button of the element widget
4 - Snap command group could go inside the setting button too.
5 - 'Show hide circular helper' : This should not require a tool. I think the underlying issue is that the constraints should render circular helper when needed.
6 - Clone/move/copy/rectangular array could be done in one 'Translate' tool with tool parameters (I've got it ready already)
7 - Coincidence and PointOnObject could be a single tool (ah yes this is a thorny topic).
8 - 'Carbon copy' tool should not require a tool. Copy paste of a sketch should be easier than that. Either just selecting it in the tree and dragging it in the 3dview. Or ctrl-C of the sketch on the tree and ctrl-V in the 3dview...
9 - 'Increase/decrease B-Spline degree' could be made a command group as is 'modify knot multiplicity'
10 - group horizontal and vertical (https://github.com/FreeCAD/FreeCAD/issues/10320 and https://github.com/FreeCAD/FreeCAD/issues/10261)
11 - 'Delete all constraints' could be removed from toolbar and set as a contextual menu entry in the constraint widget.

@obelisk79 is that something the Design Working Group want to discuss ?

### Full version info

```shell
0.22
```

### Subproject(s) affected?

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

obelisk79:
I would like to review toolbars to optimize use and recently started drafting some guidelines for toolbar/command management. I think an in-depth look at sketcher toolbars would be a good first place to look, however I would like to get the draft guidelines a little better defined so the same process can be applied to other workbenches as well. Perhaps in roughly a week I can bring this to the design working group for debate?

PaddleStroke:
Is issue is too broad. Many points in it has been addressed. Rest should have their own issue if needed.

PR #11330, opened 2023-11-09
[Sketcher] Auto horizontal/vertical

### LINKED ISSUE #10261: [Feature Request] Automatic vertical and horizontal constrain

### Full version info

```shell
[code]
OS: Manjaro Linux (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33458 (Git) AppImage
Build type: Release
Branch: master
Hash: 8b89dd29d2382c72d738c52e8974216130ffa9db
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Installed mods: 
  * PieMenu 1.2.7
  * sheetmetal 0.2.63
  * toSketch 1.0.1
  * A2plus 0.4.60n
  * QuickMeasure 2022.10.28
  * Curves 0.6.12
  * CfdOF 1.23.1
  * TabBar
  * IconThemes
  * Glass
  * ProDarkThemePreferencePack 1.0.0
[/code]
```

### Subproject(s) affected?

Sketcher

### Problem description

Hello all,
due to the vertical and horizontal constrain being separate, you have to do a lot of switching back and forth while sketching, which makes the workflow feel chunky. Other CAD programs get around this, by having only one function, which applies the right constrain based on if the line is closer to vertical or horizontal.
In case of FreeCAD, I would leave the existing functions as they are and only add a function on top, which chooses as stated.

Many thanks in Advance

### Anything else?

Here is an example from another CAD programs:
HorizontalVertical_Constraint
horizontal-and-vertical-constraint-in-creo-parametric-sketch

### Code of Conduct

### LINKED ISSUE #10320: [Feature Request] Group horizontal and vertical constraints

### Full version info

```shell
[code]
OS: Manjaro Linux (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33458 (Git) AppImage
Build type: Release
Branch: master
Hash: 8b89dd29d2382c72d738c52e8974216130ffa9db
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Installed mods: 
  * PieMenu 1.2.7
  * sheetmetal 0.2.63
  * toSketch 1.0.1
  * A2plus 0.4.60n
  * QuickMeasure 2022.10.28
  * Curves 0.6.13
  * CfdOF 1.24.4
  * TabBar
  * IconThemes
  * Glass
  * ProDarkThemePreferencePack 1.0.0
[/code]
```

### Subproject(s) affected?

Sketcher

### Problem description

Hello all,
when you work with a rotated view of the sketch, the vertical and horizontal constraints often have switched meanings, which can be disorientating or confusing.
This is just an open discussion for now, but I see two potential ways this can be handled.
First is creating an extra function on top of the current system, which automatically applies a horizontal or vertical constrain based on which is closest (previously seen in #10261).
Second is doing some sort of automatic switching intern the current system, based on the viewing direction. Though I'm not sure how it would actually work and if it would be easier to implement.

Many thanks in advance.

### Anything else?

_No response_

### Code of Conduct

PaddleStroke:
Adds a auto horizontal / vertical constraint and command group.
Refactor horizontal/vertical commands which had lot of duplicates.
Fixes #10320 
Fixes #10261

Also participate solving https://github.com/FreeCAD/FreeCAD/issues/11208

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -2589,6 +2589,347 @@ bool CmdSketcherDimension::isActive(void)
     return isCommandActive(getActiveGuiDocument());
 }
 
+// Comp for horizontal/vertical =============================================
+
+class CmdSketcherCompHorizontalVertical : public Gui::GroupCommand
+{
+public:
+    CmdSketcherCompHorizontalVertical()
+        : GroupCommand("Sketcher_CompHorVer")
+    {
+        sAppModule = "Sketcher";
+        sGroup = "Sketcher";
+        sMenuText = QT_TR_NOOP("Horizontal/Vertical");
+        sToolTipText = QT_TR_NOOP("Constrains a single line to either horizontal or vertical.");
+        sWhatsThis = "Sketcher_CompHorVer";
+        sStatusTip = sToolTipText;
+        eType = ForEdit;
+
+        setCheckable(false);
+
+        addCommand("Sketcher_ConstrainHorVer");
+        addCommand("Sketcher_ConstrainHorizontal");
+        addCommand("Sketcher_ConstrainVertical");
+    }
+
+    const char* className() const override { return "CmdSketcherCompDimensionTools"; }
+};
+
+// ============================================================================
+bool canHorVerBlock(Sketcher::SketchObject* Obj, int geoId)
+{
+    const std::vector<Sketcher::Constraint*>& vals = Obj->Constraints.getValues();
+
+    // check if the edge already has a Horizontal/Vertical/Block constraint
+    for (auto& constr : vals) {
+        if (constr->Type == Sketcher::Horizontal && constr->First == geoId
+            && constr->FirstPos == Sketcher::PointPos::none) {
+            Gui::TranslatedUserWarning(
+                Obj,
+                QObject::tr("Double constraint"),
+                QObject::tr("The selected edge already has a horizontal constraint!"));
+            return false;
+        }
+        if (constr->Type == Sketcher::Vertical && constr->First == geoId
+            && constr->FirstPos == Sketcher::PointPos::none) {
+            Gui::TranslatedUserWarning(
+                Obj,
+                QObject::tr("Impossible constraint"),
+                QObject::tr("The selected edge already has a vertical constraint!"));
+            return false;
+        }
+        // check if the edge already has a Block constraint
+        if (constr->Type == Sketcher::Block && constr->First == geoId
+            && constr->FirstPos == Sketcher::PointPos::none) {
+            Gui::TranslatedUserWarning(
+                Obj,
+                QObject::tr("Impossible constraint"),
+                QObject::tr("The selected edge already has a Block constraint!"));
+            return false;
+        }
+    }
+    return true;
+}
+
+void horVerActivated(CmdSketcherConstraint* cmd, std::string type)
+{
+    // get the selection
+    std::vector<Gui::SelectionObject> selection = Gui::Command::getSelection().getSelectionEx();
+
+    // only one sketch with its subelements are allowed to be selected
+    if (selection.size() != 1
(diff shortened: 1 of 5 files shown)


maxwxyz:
nice, is there a new shortcut for the new tool, additionally to "H" and "V"?

howie-j:
Did a quick test, i like it VERY much!

The shortcut is 'A' @maxwxyz 

[Screencast from 2023-11-10 01-15-12.webm](https://github.com/FreeCAD/FreeCAD/assets/82042580/4df653b4-29db-4e91-b9fc-009755d9a9f6)

maxwxyz:
Looks great. What happens when I select multiple lines or horizontal and vertical lines together? Does it also work with multiple points? What happens when you click A on a circle or arc?

PaddleStroke:
Yes it works with multiple lines / points.
For each line / pair of points it will apply the correct constraint.
I put A for shortcut for Alignment. And because it was free. Not sure its the best.

maxwxyz:
A is great!
Does it produce an error if you select geometry that cannot be aligned (arcs, splines,...) or does it just do nothing as today with H/V

PaddleStroke:
Just nothing. It behaves the same
I actually refactored the code of horizontal and vertical so that the 3 tools reuse the same functions. The only difference for the HorVer tool is that it checks orientation and applies the closest.

NXTler:
Oh, I have waited for this for some time now. Thank you very much!

maxwxyz:
@PaddleStroke it seems the icon, especially on the cursor is pretty similar to perpendicular: 
grafik
maybe place the horizontal line on the right side of the vertical with a space between them:  | –
Or maybe a little / abstract grid to indicate its based on the sketch orientation plane?
Or just how the sketch axis look: 
grafik

PaddleStroke:
You are right. Rotating the icon is the simpler. If a design person has a proposal don't hesitate to post. Else I will just rotate 90deg the icon :)

pierreporte:
@PaddleStroke why not a cross like a plus sign? Basically the two constraints fused together.

PaddleStroke:
image
so like this?
It's ok for me

pierreporte:
@PaddleStroke it’s perfect! I like that you can see it’s the two icons together and not just a new one.

PaddleStroke:
@maxwxyz someone asked for a similar tool for parallel / perpendicular. Do you think it is an interesting idea?

maxwxyz:
Definitely would be good! I wonder about the icon of this feature 😄 
Similarly the coincident and point on object could be one tool, depending on the selection of the user I guess.

PaddleStroke:
coincident has already a PR : https://github.com/FreeCAD/FreeCAD/pull/11494
but its going to be added as an option OFF by default for now because it was a bit controversial before. I'll let other battle to get it to be the default.

pierreporte:
> Someone asked for a similar tool for parallel / perpendicular. Do you think it is an interesting idea?

I haven’t seen this in other software. They make this automatically (see #11414). I’m not against but at this point it may be be better to have a button enabling what I propose in #11413.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -2725,443 +2721,304 @@ void CmdSketcherConstrainHorizontal::activated(int iMsg)
...
-        static_cast<SketcherGui::ViewProviderSketch*>(getActiveGuiDocument()->getInEdit());
-    Sketcher::SketchObject* Obj = sketchgui->getSketchObject();
-
-    switch (seqIndex) {
-        case 0:// {Edge}
-        {
-            // create the constraint
-            const std::vector<Sketcher::Constraint*>& vals = Obj->Constraints.getValues();
-
-            int CrvId = selSeq.front().GeoId;
```
PaddleStroke, line 3076: @wwmayer regarding your comment on the other PR, this line of code was already like this before. So the auto vertical/horizontal PR did not changed that.
And it seemed that it wasn't causing problem before. So I'm not sure if it problematic? What do you think?

PR #11334, opened 2023-11-10
[Do not merge] refactor Pi and "e" consts ;

mosfet80:
Refactor pi const
Remove multiple definition of PI.
Create FCConsts ->contain math const


tobiasfalk:
Question, wouldn't a "FCCpi" be a better name for "pi"?
This way one can easily see from where it comes, a other solution would be to put it in its own name space(FCC::) and to not allow te use of usenamespace.

mosfet80:
> Question, wouldn't a "FCCpi" be a better name for "pi"? This way one can easily see from where it comes, a other solution would be to put it in its own name space(FCC::) and to not allow te use of usenamespace.

Even for me the name pi is too short, but it was already used.
It is necessary to standardize the name of the variables and centralize the definition.
an example is pi defined several times with different approximations.
@lupaz 
One of my ideas is to use the FCConsts.h file as a single container for mathematical and non-mathematical constants.
The nomenclature on names is also important.
I don't know if it's appropriate to open a discussion on the forum or use fcc_PI as the variable that contains the definition of pi.

tobiasfalk:
"fcc_PI" also sound like a good name.
One other question I had while reading your answer is how hard would it be to expand this so that the Constants are not only usable in code but also in number fields in FreeCAD itself, as in a dimension or something.

wwmayer:
With C++20 [std::numbers](https://en.cppreference.com/w/cpp/numeric/constants) has been added to the standard. IMO, we should use the same names as defined there.

To avoid that inside the code only `pi` is used I suggest to handle this with namespaces. The constants can be defined inside the namespace `numbers` or `Base::numbers` and then an alias namespace can be set as
`namespace num = numbers;` or `namespace num = Base::numbers;`

Inside the code the constants can be used as `num::pi`, ...

maxwxyz:
@mosfet80 just a heads up for the upcoming feature freeze: 

maxwxyz:
@mosfet80 any news on this PR?

PR #8818, opened 2023-03-10
Draft: Fixes grid on/off behaviour - fixes #5878

yorikvanhavre:
Fixes faultly Draft grid behaviour. Now:

- If grid is on and auto grid on: Grid always appears
- If grid is on and auto grid off: Grid off at start, on when using a tool, off afterwards
- If grid is off: Grid never appears

- [x]  Your Pull Request meets the requirements outlined in section 5 of [CONTRIBUTING.md](https://github.com/FreeCAD/FreeCAD/blob/master/CONTRIBUTING.md) for a Valid PR
---

PR #11336, opened 2023-11-10
Draft: Improve grid behavior

Roy-043:
This PR improves the behavior of the Draft grid:
* Options in the preferences: "Always show the grid" and "Show the grid during commands".
* The parameters are used for new views. In existing views the grid can be toggled either during a command or when no command is active.
* When switching to a different workbench all grids are hidden unless GridHideInOtherWorkbenches is set to `False`. This can be a Fine-tuning parameter IMO.

Notes:
* The defaultCameraHeight parameter becomes obsolete. It was only used in very specific circumstances.
* The code of PR #8818 becomes obsolete. It makes more sense to use `App.activeDraftCommand` instead of adding a `tool` argument to some functions. Will do some related cleanup later.
* Improved alignment for the related preference ui. The checkbox interaction has been fixed as well.

### DIFF
--- a/src/Mod/Draft/Resources/ui/preferences-draftsnap.ui
+++ b/src/Mod/Draft/Resources/ui/preferences-draftsnap.ui
@@ -30,20 +30,20 @@
     <number>9</number>
    </property>
    <item>
-    <widget class="QGroupBox" name="groupBox_2">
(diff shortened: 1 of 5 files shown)


yorikvanhavre:
Looks perfect!

PR #11338, opened 2023-11-10
Core: Clang format base

wwmayer:
Applies clang-format rules to Base and fixes a couple of lint warnings


src/Base/BaseClass.h
```diff
@@ -31,144 +31,193 @@ using PyObject = struct _object;
...
 
-  template <typename T>
-  bool isDerivedFrom() const
-  {
-    return getTypeId().isDerivedFrom(T::getClassTypeId());
-  }
 private:
-  static Type classTypeId;
+    static Type classTypeId;  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
+
```
abdullahtahiriyo, line 170: Not sure why the category is unknown.

src/Base/BaseClassPyImp.cpp
```diff
@@ -36,27 +36,32 @@ std::string BaseClassPy::representation() const
...
+    if (!PyArg_ParseTuple(args, "s", &name)) {
         return nullptr;
+    }
 
+    // clang-format off
     Base::Type type = Base::Type::fromName(name);
     bool valid = (type != Base::Type::badType() && getBaseClassPtr()->getTypeId().isDerivedFrom(type));
     return PyBool_FromLong(valid ? 1 : 0);
+    // clang-format on
 }
```
abdullahtahiriyo, line 51: Why did you choose to deactivate clang-format for this block? It is mere curiosity and to learn something.

src/Base/Bitmask.h
```diff
@@ -128,5 +129,6 @@ class Flags {
     }
 };
 }
+// clang-format on
```
abdullahtahiriyo, line 133: Same question. Just marking the block in case the answer is different.

src/Base/BaseClassPyImp.cpp
wwmayer: Sometimes the auto-formatting reduces readability.

src/Base/Bitmask.h
wwmayer: Same answer.

src/Base/Parser.bat
```diff
@@ -1,2 +1,2 @@
 C:\cygwin\bin\flex.exe -oQuantityLexer.c  QuantityParser.l
-C:\cygwin\bin\bison -oQuantityParser.c QuantityParser.y
\ No newline at end of file
+C:\cygwin\bin\bison -oQuantityParser.c QuantityParser.y
```
abdullahtahiriyo, line 2: Nothing wrong in this case, but should clang-format parse batch files.

src/Base/PyExport.h
```diff
@@ -68,34 +68,35 @@ class PyObjectBase;
  *  because at construction time the reference counter is already set to 1. If the Python
  *  interpreter stores this object pointer into a local variable and destroys this variable
  *  then the reference counter gets decremented (to 0) and the object gets destroyed automatically.
- *  In case we didn't make this specification and increment the Python object from outside once again then
- *  the reference counter would be set to 2 and there would be no chance to destroy the object again.
+ *  In case we didn't make this specification and increment the Python object from outside once
+ * again then the reference counter would be set to 2 and there would be no chance to destroy the
+ * object again.
```
abdullahtahiriyo, line 73: Um... when splitting line endings of doc blocks clang-format does not respect the indentation... line 71 two spaces, lines 72-73 one. I would have expected either to format every single line to one space, or respect two spaces when handling too long lines...

src/Base/PyTools.c
```diff
@@ -660,8 +661,8 @@ PP_Debug_Bytecode(PyObject *codeobject, PyObject *moddict)
     res = PP_Run_Function(            /* "pdb.runeval(codeobj, gdict, ldict)" */
              "pdb",    "runeval",     /* accepts string|code, code=stmt|expr  */
              "O",      &presult,
-             "(OOO)",  codeobject, moddict, moddict); 
+             "(OOO)",  codeobject, moddict, moddict);
     return (res != 0) ? NULL : presult;     /* null if error in run_function */
 }
 // NOLINTEND
-
+// clang-format on
```
abdullahtahiriyo, line 668: Which criteria was used to prevent formatting of this file? (Just to know...)

abdullahtahiriyo:
Thanks for taking the time for reformatting.

I am more than ok with merging this (I would do it right away, but I guess you are giving some time to create awareness this is going to happen). I am thrilled. Often the result of clang-format is astonishingly better than the original.

About disabling autoformatting on certain files, I think we should have some clear objective criteria as a group. There are some things I certainly do not like about clang-format, but it makes life so convenient and PRs so much cleaner, that for me it is a no brainer to go along with it. So my interest in which exceptions should be considered.

The batch file, I wonder whether we should tweak the clang-format config to omit them (for future executions, the result of this one is ok), as I am unsure if clang-formatting a batch file may lead to loss of functionality, as it is not c code.

Thanks again. Looking forward to have it merged.

src/Base/PyTools.c
wwmayer: In general I don't want to format 3rd party code as this makes it very difficult to update to a newer version if needed.

src/Base/BaseClassPyImp.cpp
wwmayer: Because the line is a bit too long clang-format will break it which IMO doesn't look so nice. Now that the class BaseClass has the method isDerivedFrom() the line can be make shorter which makes the formatting off/on superfluous. I will push force a commit.

wwmayer:
> Often the result of clang-format is astonishingly better than the original.

In general yes. But more important is that we have finally a consistent coding style.

> About disabling autoformatting on certain files, I think we should have some clear objective criteria as a group.

Some rules:
* If the code afterwards looks worse and is harder to read and understand. I have several cases in Matrix.cpp
* Sometimes the tool fails to achieve a correct formatting. I had this case with StackWalker.cpp where a git commit activated the hook script and this ended up with an error without telling what was the reason. The same things have happened with the pre-commit hook on GH which then raised a build failure

src/Base/PyTools.c
abdullahtahiriyo: For this I thought that was the answer.  Thanks.

3rd party may even have a good but different formatting setting altogether and reformatting does not make sense.

src/Base/BaseClass.h
wwmayer: I have removed the exact category.

src/Base/BaseClassPyImp.cpp
wwmayer: The modified code doesn't require the disabled formatting any more

src/Base/PyExport.h
wwmayer: The problem might be that the line starts with `*`. I have manually removed the spaces.

PR #11344, opened 2023-11-10
[Core] Status bar : reduce number of decimals

### LINKED ISSUE #10837: [Problem] Status messages precisions of doubles is too high

### Problem description

image
It has been said on the forum before at several places.
The precision of double values is needlessly high in status messages, which reduce readability and serve no purpose as far as I know.
I suggest reducing the number of decimals to 1 or 2 max so that it will be much more readable and will clutter less the view.

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Anything else?

_No response_

### Code of Conduct

PaddleStroke:
Fixes #10837

@wwmayer can you give your opinion on this ? Also if you have an idea where we could put a function to avoid this 3 time copy paste.

Thanks

### DIFF
--- a/src/Gui/Selection.cpp
+++ b/src/Gui/Selection.cpp
@@ -720,8 +720,6 @@ std::array<std::pair<double, std::string>, 3> schemaTranslatePoint(double x, dou
 
 void SelectionSingleton::setPreselectCoord( float x, float y, float z)
 {
-    static char buf[513];
-
     // if nothing is in preselect ignore
     if(CurrentPreselection.Object.getObjectName().empty())
         return;
@@ -730,17 +728,24 @@ void SelectionSingleton::setPreselectCoord( float x, float y, float z)
     CurrentPreselection.y = y;
     CurrentPreselection.z = z;
 
-    auto pts = schemaTranslatePoint(x, y, z, 0.0);
-    snprintf(buf,512,"Preselected: %s.%s.%s (%f %s,%f %s,%f %s)"
-                    ,CurrentPreselection.pDocName
-                    ,CurrentPreselection.pObjectName
-                    ,CurrentPreselection.pSubName
-                    ,pts[0].first, pts[0].second.c_str()
-                    ,pts[1].first, pts[1].second.c_str()
-                    ,pts[2].first, pts[2].second.c_str());
+    if (getMainWindow()) {
+        auto pts = schemaTranslatePoint(x, y, z, 0.0);
 
-    if (getMainWindow())
-        getMainWindow()->showMessage(QString::fromUtf8(buf));
+        int numberDecimals = std::min(6, Base::UnitsApi::getDecimals());
+
+        QString message = QStringLiteral("Preselected: %1.%2.%3 (%4 %5, %6 %7, %8 %9)")
+            .arg(QString::fromUtf8(CurrentPreselection.pDocName))
+            .arg(QString::fromUtf8(CurrentPreselection.pObjectName))
+            .arg(QString::fromUtf8(CurrentPreselection.pSubName))
+            .arg(QString::number(pts[0].first, 'f', numberDecimals))
+            .arg(QString::fromStdString(pts[0].second))
+            .arg(QString::number(pts[1].first, 'f', numberDecimals))
+            .arg(QString::fromStdString(pts[1].second))
+            .arg(QString::number(pts[2].first, 'f', numberDecimals))
+            .arg(QString::fromStdString(pts[2].second));
+
+        getMainWindow()->showMessage(message);
+    }
 }
 
 void SelectionSingleton::rmvPreselect(bool signal)
--- a/src/Gui/SoFCSelection.cpp
+++ b/src/Gui/SoFCSelection.cpp
@@ -40,6 +40,8 @@
 # include <Inventor/misc/SoState.h>
 #endif
 
+#include <Base/UnitsApi.h>
+
 #include "SoFCSelection.h"
 #include "MainWindow.h"
 #include "SoFCSelectionAction.h"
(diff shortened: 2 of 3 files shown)


PaddleStroke:
By the way there has been a feature request before to show more useful information about the (pre)selected element. Such as volume (/weight) for solids, length for edges, positions for vertexes. Area for surfaces. Radius for arc/circle and so on.

wwmayer:
> Also if you have an idea where we could put a function to avoid this 3 time copy paste.

I think it's best to implement a helper class. See also: https://github.com/FreeCAD/FreeCAD/issues/10837#issuecomment-1806826459

> By the way there has been a feature request before to show more useful information about the (pre)selected element. Such as volume (/weight) for solids, length for edges, positions for vertexes. Area for surfaces. Radius for arc/circle and so on. But I can't find it anymore.

I don't know but this cannot be done with this helper class because it doesn't know anything about geometries of the Part module.

PaddleStroke:
> I don't know but this cannot be done with this helper class because it doesn't know anything about geometries of the Part module.

So what would be the proper way then ? Best to make it right the first time.

Perhaps we should have a 'getObjectPreselectionMessage()' function in the base class of viewProvider so that it can be overriden by modules?

wwmayer:
> So what would be the proper way then ? Best to make it right the first time.

The correct way is to implement a class in the PartGui module. But I wouldn't write this information to the status bar because it's too much information and for people with a rather small monitor the text might be truncated.

Instead I would implement a special task watcher that reacts on selection changes. This task watcher will add a widget to the task view that shows all needed information. As an example code have a look at the class MeshInfoWatcher from the MeshGui module.

PaddleStroke:
> The correct way is to implement a class in the PartGui module. But I wouldn't write this information to the status bar because it's too much information and for people with a rather small monitor the text might be truncated.

I don't think the text is going to be so big. If you look at our current messages : 

> Preselected: docname.Part001.Body001.Pad002.Face14 (43.123456 mm, 39.123456 mm, 19.123456 mm)

A lot of this information is not useful. Actually the whole information is useless. What is the point of the mouse coordinates on the preselected face ? A message like the following would be much more useful : 

> Body001.Face14 (surface : 435.12 mm²)

Besides instead of message on preselection, it would be best to have a message on selection. This way we could also show meaningful information between several objects. For example : 

> Body001.Face14 - Body002.Vertex14 (distance: 120.12 mm)

wwmayer:
> I don't think the text is going to be so big. If you look at our current messages :

But wouldn't this be additional text? Assuming you hover a cylindrical face then the displayed string might be something like:
Preselected: docname.Part001.Body001.Pad002.Face14 (43.123456 mm, 39.123456 mm, 19.123456 mm) Cylinder Base (45.000000, 40.00000, 19.123456) Axis (0.000000, 0.000000, 1.000000) Radius 10 Area 620.123456

> A lot of this information is not useful.

I wonder how you know this. If it's not useful for you it doesn't mean it's not useful for everybody else.

> This way we could also show meaningful information between several objects. For example :

And if 3, 4, 5 or more objects are selected? At some point it will become too complicated to show sensible information in the status bar. By using a task watcher instead this can be much better organized.

PaddleStroke:
By the way I found the forum discussion about this : https://forum.freecad.org/viewtopic.php?style=4&t=80919

wwmayer:
> If you know of use cases please share.

When developing algorithms or hunting bugs I very often need an easy way to get quickly some information about the current geometry. E.g. when moving over a planar face and observing which of the (x,y,z) value change I already know the normal.
Or recently I have fixed this bug 4c509c03efc6 and therefore knowing some basic information about the position of the shape was very useful.

PaddleStroke:
I have : 
- changed to QString
- changed to UnitApi.

I have not : 
- Capped the number of decimals to 6 : I don't see any justification to do so. If the user is using 8 or 9 decimals, or even 16, he is probably in a special case and he may actually want to see the precision in the message.
- Create a helper class : I have done minimal modification which are not making the situation worse. It would be better to make a class yes, but as I want to tackle this again later to add more information, I don't want to do a temporary class now.

So I suggest we merge this as it is to solve the number of decimals. Then later we'll improve the message with more information.

PaddleStroke:
> [ 69%] Building CXX object src/Gui/CMakeFiles/FreeCADGui.dir/SoFCSelection.cpp.o
> /home/runner/work/FreeCAD/FreeCAD/src/Gui/SoFCSelection.cpp:43:10: fatal error: Base/UnitsAPI.h: No such file or directory
>    43 | #include <Base/UnitsAPI.h>

I can't understand why linux doesn't build. The CI report says the header doesn't exist while it's correct. What is wrong here?

PaddleStroke:
@wwmayer we could then keep the mouse position like so : 

> At (43.12 mm, 39.12 mm, 19.12 mm) : Body001.Face14 Radius 10.00 mm Area 620.12m²

But then this would mean we keep the messages on preselection rather than selection.

0penBrain:
> I can't understand why linux doesn't build. The CI report says the header doesn't exist while it's correct. What is wrong here?

Haa. Windows users and case sensitivity... 🙊

wwmayer:
> Capped the number of decimals to 6 : I don't see any justification to do so.

The corresponding OpenInventor nodes and the API to get the picked point only supports floats which have a precision of 6 digits. So, showing numbers with more digits is pointless or even confusing because they won't be correct any more.

PaddleStroke:
@0penBrain thanks!

@wwmayer thanks for the explanation, it makes perfect sense now.

wwmayer:
Code Cleanup #11469

PaddleStroke:
@wwmayer could you please elaborate on :

> The correct way is to implement a class in the PartGui module. 

I see what the class could look like, but where do I instanciate this class?

PR #11356, opened 2023-11-11
[Python editor] match current indentation for next line when  pressin…

mwganson:
…g enter key, addresses issue 5551.

The current python editor treats the enter key as a standard text editor, inserting a new line and a carriage return.  But in python coding it is beneficial that enter/return key should match the current line's indentation rather than require the user to tab/space over to the correct indent.  I have also added a feature that if the current line ends in a colon (:) then the next line also gets an additional indent.  The colon always signifies a new logical block, such as if: else: def function(): class name:, etc.

I have also added the feature that Shift+Enter should remove one indent from the next line.  Example usage:

```
if bool1: #enter would indent next line
    logic for bool1 goes here #Enter keeps this level of indentation for the next line
    logic for bool1 continues #Shift+Enter restores previous indentation, for example for the else: clause
else:
    logic for else goes here #Shift+Enter goes back to previous indentation
```

### DIFF
--- a/src/Gui/PythonEditor.cpp
+++ b/src/Gui/PythonEditor.cpp
@@ -159,6 +159,55 @@ void PythonEditor::contextMenuEvent ( QContextMenuEvent * e )
     delete menu;
 }
 
+void PythonEditor::keyPressEvent(QKeyEvent* e)
+{
+    /** When the user presses enter the next line should match the current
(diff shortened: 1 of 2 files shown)


mwganson:
https://forum.freecad.org/viewtopic.php?t=82631

luzpaz:
Thanks @mwganson!

PR #11363, opened 2023-11-12
Add OndselSolver as git submodule

adrianinsaval:
also update github actions to pull submodules

this is a prerequisite for https://github.com/FreeCAD/FreeCAD/pull/10764


chennes:
Nice - thanks for the clear error message for devs who need to run submodule init.

adrianinsaval:
note to self: after this is merged need to make sure conda builds clone with submodules, flatpak might need to enable some option for this

@luzpaz snap build process might need to be adapted to pull git submodules too, other packagers might need to adjust their build processes too.

chennes:
I can take care of the snap (I think...).

tobiasfalk:
Now that git submodules are mor over less usable in FreeCAD there would definitely be other regions that would benefit from them. 

One would be the usage of osifont in TechDraw, there the Font file could be replaced with the submodule of the github repo from osifont. This is becaus the last time I looked currently used version is a rather outdated version of the Font.

There would definitely be more places that could benefit from it.

tobiasfalk:
Now that git submodules are mor over less usable in FreeCAD there would definitely be other regions that would benefit from them. 

One would be the usage of osifont in TechDraw, there the Font file could be replaced with the submodule of the github repo from osifont. This is becaus the last time I looked currently used version is a rather outdated version of the Font.

There would definitely be more places that could benefit from it.

luzpaz:
> Note that in order to build FreeCAD [using a Launchpad PPA recipe](https://code.launchpad.net/~freecad-maintainers/freecad/+git/freecad/+recipes), OndselSolver would need to be [registered as its own Launchpad project](https://launchpad.net/projects/+new) in order to fetch its code from GitHub. (git-build-recipe doesn't support checking out submodules, but it is possible to use the nest command to work around this.)

CC@adrianinsaval

Issue #7395, opened 2022-08-25
[Feature Request] Unit system with dimensions displayed in meters


FEA-eng:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=8&t=71365

### Subproject(s) affected?

Core

### Idea description

The currently available MKS (m/kg/s/degree) unit system doesn't always result in dimensions being displayed in meters - millimeters are still used for values bellow 0.1m
For some applications (e.g. civil engineering) it would be useful to have a unit system that actually changes the display of all dimensions to meters.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
Looking at the code MKS is supposed to display values bellow 100mm in mm and values above that but bellow 10000000mm are displayed in m and so on for even higher or lower values
https://github.com/FreeCAD/FreeCAD/blob/a8cb4e744483ce2c4b0770b948105a766330a4b2/src/Base/UnitsSchemaMKS.cpp#L57-L64
These are sane ranges IMO, if there is somewhere where mm are being used regardless of these rules I think that would be a bug but we would need examples of that.

adrianinsaval:
@FEA-eng respond please, where is there an issue where mm are displayed with MKS? if no example is given this issue should be closed IMO, or probably closed either way and instead make an issue for the field that is not respecting the unit schema.

FEA-eng:
@adrianinsaval I would expect the (m/kg/s/degree) unit system to display meters at all times, not only when the dimensions are close to 1 m. That's how it's done in other CAD software (like SolidWorks). Seeing mm displayed everywhere (for example the default 10 mm depth of extrusion) with MKS system enabled is really confusing. Maybe this request should be changed to something like "make the MKS system display meters regardless of the dimension value" instead.

adrianinsaval:
I see no reason why you would want to see 10mm as 0.01m, "other software does it" is not a very good argument (although it can be valid for some stuff). Can you give me an example of a situation where forcing m for every value is advantageous?

FEA-eng:
Small dimensions displayed in meters may seem inconvenient but it's mainly about consistency (and clarity). Especially if you are going to show the dimensions to someone, then the units should be consistent according to the rules of engineering. I'm not a civil engineer but guys from that field would likely agree with me that the MKS system should show meters at all times.

adrianinsaval:
[...]

PR #11365, opened 2023-11-12
[Units]Add Meters-only schema (#7395)

WandererFan:
This PR implements a solution for #7395.

Additional forum discussion: 
https://forum.freecad.org/viewtopic.php?t=71365
https://forum.freecad.org/viewtopic.php?t=59635


src/Base/UnitsSchemaMeterDecimal.cpp
```diff
@@ -0,0 +1,89 @@
...
+        factor = 1000000;
+    }
+    else if (unit == Unit::HeatFlux) {
+        // watts per square metre
+        unitString = QString::fromLatin1("W/m^2");
+        factor = 1.0;
+    }
+    else if (unit == Unit::Velocity) {
+        // metres per second
+        unitString = QString::fromLatin1("m/sec");
```
wwmayer, line 79: "sec" is not a valid SI unit. It must be "s" (https://en.wikipedia.org/wiki/Second). In case the faulty string is used in an expression an error is raised: `App.Units.parseQuantity("2m/sec")`

> Traceback (most recent call last):
ValueError: invalid unit expression

WandererFan: unit string changed to "m/s".

src/Base/UnitsSchemaMeterDecimal.h
```diff
@@ -0,0 +1,51 @@
...
+ ***************************************************************************/
+
+/*  Metric units schema intended for design of large objects
+ *  Lengths are always in metres.
+ *  Angles in degrees (use degree symbol)
+ *  Velocities in m/sec
+ */
+
+#ifndef BASE_UNITSSCHEMAMETERS_H
+#define BASE_UNITSSCHEMAMETERs_H
```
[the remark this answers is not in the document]
WandererFan: fixed.

PR #11373, opened 2023-11-13
Core, wb techdraw, wb part :Removed deprecated func

mosfet80:
QGIPrimPath::setHighlighted, TopoShape::setShapePlacement and TopoShape::getShapePlacement are obsolete and never used

### DIFF
--- a/src/Gui/NavigationStyle.cpp
+++ b/src/Gui/NavigationStyle.cpp
@@ -1082,11 +1082,6 @@ void NavigationStyle::moveCursorPosition()
     }
 }
 
-void NavigationStyle::redraw()
-{
-    if (mouseSelection)
-        mouseSelection->redraw();
-}
 
 SbBool NavigationStyle::handleEventInForeground(const SoEvent* const e)
 {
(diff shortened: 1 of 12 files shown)


Pesc0:
@JohnAD @chennes check toposhape changes. I think It's fine to remove these, they seem duplicated anyway.

src/Gui/NavigationStyle.cpp
```diff
@@ -1020,8 +1020,7 @@ void NavigationStyle::moveCursorPosition()
 
 void NavigationStyle::redraw()
 {
-    if (mouseSelection)
-        mouseSelection->redraw();
```
chennes, line 1024: What's going on with this removal?

wwmayer: With AbstractMouseSelection::redraw() also NavigationStyle::redraw() can be completely removed.

chennes:
@mosfet80 Can you check over this again in light of @wwmayer's comment above?

sliptonic:
Moving back to draft.  @mosfet80  Please reset when ready to review.

PR #11380, opened 2023-11-13
SketcherGui: DSH Default - return reference in ShapeConstraint helpers

abdullahtahiriyo:
Improving helper function to return a reference.


abdullahtahiriyo:
@PaddleStroke 

Now, if it matches your workflow, you can use the return value for constraints too. Let me know if you hit any issue. Framework provided convenience functions are to make the life of the user comfortable. You are the user :)

PR #11385, opened 2023-11-13
Vscode PR - split 1 - add cmakepresets

Pesc0:


CMakePresets.json
```diff
@@ -0,0 +1,303 @@
...
+            "value": "OFF"
+          },
+          "FREECAD_LIBPACK_USE": {
+            "type": "BOOL",
+            "value": "OFF"
+          },
+          "FREECAD_USE_CCACHE": {
+            "type": "BOOL",
+            "value": "OFF"
+          },
```
adrianinsaval, line 93: you only want this in windows I think, and FreeCAD itself already defaults to not using this on windows. So I think we should remove this

CMakePresets.json
```diff
@@ -0,0 +1,303 @@
...
+          "BUILD_WITH_CONDA": {
+            "type": "BOOL",
+            "value": "ON"
+          },
+          "ENABLE_DEVELOPER_TESTS": {
+            "type": "BOOL",
+            "value": "OFF"
+          },
```
adrianinsaval, line 85: This one makes sense for packaging but for development you probably want to compile the test suite

CMakePresets.json
```diff
@@ -0,0 +1,303 @@
...
+            "value": "ON"
+          },
+          "BUILD_FEM_NETGEN": {
+            "type": "BOOL",
+            "value": "ON"
+          },
+          "BUILD_FLAT_MESH": {
+            "type": "BOOL",
+            "value": "ON"
+          },
```
adrianinsaval, line 69: this defaults to ON now I think, so we don't actually need to set it

CMakePresets.json
```diff
@@ -0,0 +1,303 @@
...
+          "BUILD_REVERSEENGINEERING": {
+            "type": "BOOL",
+            "value": "ON"
+          },
```
adrianinsaval, line 73: this defaults to ON so it doesn't need to be specified, furthermore this one might ge axed in the future since it's unmaintained so probably best not to add more places we're we'll have to change defaults when that happens

CMakePresets.json
```diff
@@ -0,0 +1,303 @@
...
+          "BUILD_SHIP": {
+            "type": "BOOL",
+            "value": "OFF"
+          },
```
adrianinsaval, line 77: ship has been removed, this option is pointless

Pesc0:
CC @oursland

Pesc0:
@adrianinsaval did you by any chance see the conversation on discord?

adrianinsaval:
I very rarely go there, what's up? where should I look?

Pesc0:
https://discord.com/channels/870877411049357352/1170783081096028180/1173713128702365746

quite a long conversation, you don't really need to read it. The TLDR is that we're splitting the VScode PR, the 3rd piece will have the majority of the changes. Also Jacob wanted to create CI workflows to use CMakePresets and the conda env from splits 1 and 2.

Pesc0:
done, let's wait for Jacob's review on this one

adrianinsaval:
> Also Jacob wanted to create CI workflows that make use of CMakePresets and the conda env from splits 1 and 2.

that's pretty good, I've been wanting to do taht for mac for a long time but never got aroudn to it

oursland:
I will be able to test these changes this afternoon.

`qdri` on Discord noted that a configuration for Windows with LibPack should be included.  I'll look into that as well.  I think that "just works" as CMake performs a configure time check for LibPack and includes it if found.

Pesc0:
Great!

FYI qdri is Adrian himself ;)

adrianinsaval:
I consider this good to merge, when always add more things in future PRs

wwmayer:
@Pesc0 
This PR is causing an annoying behaviour: every time the CMake GUI is opened an error dialog pops up.
```
Error reading CMake Presets

Could not read presets from /home/user/build/FreeCAD:
Unrecognized "version" field
```

How can this be fixed?

wwmayer:
Setting version to 3 fixes the problem for me. Here is an overview of the mapping of version number and CMake version:
https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#format

My CMake version is 3.22.1 and thus corresponds to version 3. For version 6 the CMake version must be 3.25 or above which isn't older than 6 months. So, I guess quite a lot of people still use an older version and get the same error dialog.

Pesc0:
Thats odd. It is literally the same as in the cmake documentation https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

Pesc0:
https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-presets.md I was looking here. I think it's fine to downgrade.

Pesc0:
#11544

CMakePresets.json
```diff
@@ -0,0 +1,287 @@
...
+          "FREECAD_USE_EXTERNAL_FMT": {
+            "type": "BOOL",
+            "value": "ON"
+          },
+          "FREECAD_USE_EXTERNAL_SMESH": {
+            "type": "BOOL",
+            "value": "ON"
```
dzid26, line 84: This doesn't work for me when building on Windows (conda release) in VScode. 
```
[build] C:\Users\dzidm\Downloads\FreeCAD\src\Mod\Fem\App\PreCompiled.h(63): fatal error C1083: Cannot open include file: 'SMDS_PolyhedralVolumeOfNodes.hxx': No such file or directory
```
If I set it to off then `med.h` cannot be found at `file(READ ${meddotH} TMPTXT)` in `SetupSalomeSMESH.cmake`.  

In the end I needed to override `"-DBUILD_FEM=OFF"` as a workaround.

oursland: It had worked for me previously.  I'll retest.

oursland: I just re-tested and had no issues with the current configuration.  Furthermore, this is the configuration used in CI: https://github.com/FreeCAD/FreeCAD/blob/main/.github/workflows/sub_buildWindowsConda.yml#L108

It appears the error message is getting caught up in a precompiled header, however these should not be enabled.  By default `FREECAD_USE_PCH` is set to `OFF`: https://github.com/FreeCAD/FreeCAD/blob/main/CMakePresets.json#L82

dzid26: Thank you for checking this. 
I am sorry, I had`FREECAD_USE_PCH` overridden with ON

oursland: I am uncertain the utility of that setting as it causes build issues for most platforms and I believe it also breaks `ccache` functionality, so it is disabled for all platforms on CI.

Issue #5991, opened 2022-02-07
PartDesign: Add Common (intersection) operations


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=4135

* **Reporter:** townkat
* **Date submitted:** 9/17/2019
* **FreeCAD version:** 
* **Category:** Feature
* **Status:** new
* **Tags:** Part Design

# Original report text

Please add Common (intersection) operations in PartDesign in addition to Additive and Subtractive features.

> Boolean operations in Part WB are Fusion(union) (exposed as Additive features in PartDesign), Cut (exposed as Substractive features) and Common(intersection) which is not available as features in PartDesign.

More info in the following topics:
https://forum.freecadweb.org/viewtopic.php?f=8&t=35755
https://forum.freecadweb.org/viewtopic.php?f=8&t=39297

Thank you.

> Anyone care to comment why isn't there a native PartDesign common feature? I recently designed a pill box and had to add two pockets "thorough all" to remove the extra helixes from both ends. It would have been a lot simple to just define a cylinder and trim everything that is outside of it. 

Before:
![image](https://github.com/FreeCAD/FreeCAD/assets/4140247/04229344-5442-473a-9f25-c72d34e184ee)
After:  
![image](https://github.com/FreeCAD/FreeCAD/assets/4140247/45fe3c0f-3706-4194-bcb6-c122fdca161d)

# Other bug information

* **Priority:** high
* **Severity:** feature
* **Category:** Feature
* **Updated:** 2/6/2021

luzpaz:
@FlachyJoe is this a duplicate ?

FlachyJoe:
@luzpaz I don't know if it's shown elsewhere but I see SubstractiveHelix has an Outside property which do this. It seems quite easy to implement for the other subtractive features.

luzpaz:
> seems quite easy to implement

@FlachyJoe go for it ? or perhaps if you don't have time, can you elaborate here how a dev would go about adding it ?

FlachyJoe:
The implementation for SubstractiveHelix should be duplicate for all features (maybe a common util function can be write)
Ref: https://github.com/FreeCAD/FreeCAD/blob/66ece1c0684dd6b9757a2c39a5479713555f8897/src/Mod/PartDesign/App/FeatureHelix.cpp#L344-L356

@bgbsww are you interested?

bgbsww:
@FlachyJoe I'll take a look.

[...]

PR #11389, opened 2023-11-14
Support common operations for all subtractive parts, and refactor related code

bgbsww:
Potential fix for #5991.

Open questions:  What's the right approach to copyright / assigning copyright on new files?

Don't love the "Remove outside of Profile" description of the checkbox, is there better language?


FlachyJoe:
Hi @bgbsww, thank you for this great work.

I'm testing and find some inconsistencies
On App side:
- Pocket removes above the sketch. Give red when green is expected 
image
- Outside has no effect with Pocket UpToFirst / Last / Face
image


On GUI side:
- Outside is not displayed in properties. Only with Show All checked.
- Set Outside=true open task dialog: the box is unchecked (but Outside keeps its value if you validate the task).
- check the taskdialog box, change pocket type:  the box is unchecked

bgbsww:
Of course the case I missed was the most involved one, and forced rethinking of the other after improving my test regieme.  @FlachyJoe I think I've addressed all the bugs you found, and a few more.

FlachyJoe:
@chennes it works as expected. I'm suggesting a code change but nothing about functionality.

FlachyJoe:
@bgbsww can you please squash your commits and then rebase. Conflicts were introduced by [2cef4cf](https://github.com/FreeCAD/FreeCAD/commit/2cef4cfdbe1a6b9ebf50a85c04fbe99f23cd3829)

bgbsww:
Well, that didn't work right, but I already had another branch prepared.  Is there some button in Github I'm missing?  I was forced to attempt this externally.

luzpaz:
What didn't work right ?

bgbsww:
When I force pushed back to github, I ended up with all the commits since the start of the branch, and not just mine.  I'm sure it's my error, and I already had the other branch prepared in case.  The complicating factor is that I synched this branch several times during its long lifetime, and that led to my mistake.

FlachyJoe:
In this case and  in order to keep the same PR you can:

1. create a new branch from main
2. cherry pick the needed commits from the problematic branch
3.  force push the new branch to this distant
4. delete the problematic branch

PR #11390, opened 2023-11-14
Prevent attempting to create an infinite helix

### LINKED ISSUE #11326: [Problem] Infinite processing when setting helix angle to 90deg

### Problem description

When creating a Part helix, setting the angle option to `90deg` or `-90deg` causes FreeCAD to run a seemingly infinite processing task that takes all PC memory.

### Full version info

```shell
OS: macOS 13.5.2
Word size of FreeCAD: 64-bit
Version: 0.21.1.33694 (Git)
Build type: Release
Locale: C/Default (C)
Installed mods: 
  * MeshRemodel 1.8919.0
  * ThreadProfile 1.89.0
  * Launcher
  * InventorLoader 1.4.0
  * Render 2023.8.18
  * Silk 0.1.5
  * Defeaturing 1.2.1
  * A2plus 0.4.61
  * Manipulator 1.5.0
  * fasteners 0.4.70
  * lattice2 1.0.0
  * sheetmetal 0.3.1
  * toSketch 1.0.1
  * Curves 0.6.14
  * Assembly4 0.50.4
  * Assembly3 0.12.0
```

### Subproject(s) affected?

Part

### Anything else?

Maybe users should not be allowed to create a helix with exactly 90deg or -90deg angle.

### Code of Conduct

bgbsww:
Fix #11326.

If I understand this correctly, you should never try to go to 90 degrees, that's an infinite request, thus the crash.  So I restricted the input values.


luzpaz:
@wwmayer would you mind reviewing ?

adrianinsaval:
> Besides this one should think about a sensible maximum value. Is 89.99 OK or should it be rather 89.9 or 89 or even less?

how should we go about this? Can we do `90.0 - Precision::Angular` or how?

bgbsww:
I started at Precision::Confusion which still fails, and moved my way back to the 89.99 which works, although it is creating something pretty ridiculous in scale.

I wanted some confirmation that my thinking was correct before I went into PrimitiveFeature.cpp.  I'll do that shortly.

wwmayer:
> and moved my way back to the 89.99 which works

I did a few tests and at an angle of 89.9999 the creation starts to become slower and already shows a broken helix. So, the very last acceptable value is 89.999

> although it is creating something pretty ridiculous in scale.

That's the big question. I don't think that an angle of 89.99 or 89.999 has any practical relevance. Not even an angle of 89.0 makes much sense.

But this PR is only about to fix that the application consumes all RAM. So, I would set the limits to -89.9 and 89.9

Issue #5945, opened 2022-02-07
Packaging: 'version.h' problem: Build number & hash aren't available on daily build (Linux)


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=4014

* **Reporter:** @0penBrain
* **Date submitted:** 6/10/2019
* **FreeCAD version:** 
* **Category:** Bug
* **Status:** assigned
* **Tags:** 

# Original report text

Build number & hash that identifies FreeCAD release are no more available in Linux daily builds.
Windows builds seems to not have the same problem.

---

~[Forum thread (FR)](https://forum.freecadweb.org/viewtopic.php?f=12&t=36578)~

[Dedicated thread opened](https://forum.freecadweb.org/viewtopic.php?f=42&t=37216)

# Additional information

Example with latest daily build as of today:
```
OS: Ubuntu 18.04.2 LTS (KDE/plasma)
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.19.
Build type: Release
Python version: 3.6.7
Qt version: 5.9.5
Coin version: 4.0.0a
OCC version: 7.3.0
Locale: French/France (fr_FR)
```

# Steps to reproduce

1. Open FreeCAD
2. Click on "Help" -> "About FreeCAD" -> "Copy to clipboard"
3. Paste anywhere you want

# Other bug information

* **Priority:** immediate
* **Severity:** major
* **Category:** Bug
* **Updated:** 2/6/2021

# Discussion from Mantis ticket

### Comment by @0penBrain 2019-06-10 12:14
I tagged it with high importance/priority as it's a really useful function at forum discussions.

---

### Comment by @luzpaz 2019-06-10 17:34
Punting to @sgrogan

---

### Comment by openBrain 2019-06-22 16:27
Priority --> Urgent according [this thread](https://forum.freecadweb.org/viewtopic.php?f=3&t=37186#p316103)

---

### Comment by @luzpaz 2019-06-22 18:49
kkremitzki Can this be fast tracked ?

---

### Comment by @kkremitzki 2019-06-23 05:27
Kunda1 I don't have the bandwidth to try to find a solution for this right now, but I am keeping an eye on the issue to help implement it once a solution is found, or if anyone has questions about the current build automation.

sgrogan:
I think this cna be closed as solved. [https://github.com/FreeCAD/FreeCAD/releases/tag/0.19.3](url)

[...]

PR #11396, opened 2023-11-14
Fix incremental update of Version.h using Ninja

adrianinsaval:
fixes https://github.com/FreeCAD/FreeCAD/discussions/9606

### DIFF
--- a/src/Build/CMakeLists.txt
+++ b/src/Build/CMakeLists.txt
@@ -14,12 +14,12 @@ if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/Version.h)
         COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_BINARY_DIR}/Version.h.out
             ${CMAKE_CURRENT_BINARY_DIR}/Version.h
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-        DEPENDS version_check)
+        DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/Version.h.out)
 
     add_custom_command(
-        OUTPUT  version_check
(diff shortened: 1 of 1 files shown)


luzpaz:
Does it also fix #5945

adrianinsaval:
No but I have something in mind for that one once I get the ppa revived

0penBrain:
Well, it seems there are some problem there.

I tested having the main branch to 1338385080.
I built the main.
Then I have this PR in a "PR11396" branch.
When I switch from main to this latter branch, and try to run a build, I get a `ninja: no work to do.` and nothing is built while there is some diff in between both branches (there is at least SHA, but there is actually lot more).

0penBrain:
OK sorry. After pruning my project and building from scratch, problem disappeared. May be a glitch on my side.

However I'm hesitating to merge as there is a problem now in the main. While I'm currently compiling 0ac0882eeb4e3390aef464e1807a3631c5f2e858 , version info tells me it's f30e643c828cbc901377184e752619253308b4f1 .

I have no clue why and maybe there is something to be fixed first before this PR can be thoroughly tested again.

0penBrain:
Last test for today probably. The PR actually doesn't solve problem in following situation:
 * checkout this PR and clean build => OK, SHA is correct
 * add a single comment in a file and commit the change
 * Build => NOK, the SHA is stuck on previous one

wwmayer:
See:
ae27d8e3bc4c6
3b7da50ca7d0f

adrianinsaval:
yeah I think this didn't actually fix it, I had tested between the previous commit and my own commit and it had worked, but probably because the target name changed. Later runs then don't work

FlachyJoe:
Note the easiest way to get the correct version is to run `rm src/Build/Version.h && ninja` instead of `ninja`. It re-links at most 61 unchanged files and takes very few more time.

PR #11399, opened 2023-11-14
Correct PartDesign Helix feature negative angles

### LINKED ISSUE #10284: [Problem] PartDesign Helix negative cone angle reverses the coil

### Full version info

```shell
OS: macOS 10.16
Word size of FreeCAD: 64-bit
Version: 0.20.2.29603 (Git)
Build type: Release
Branch: (HEAD detached at 0.20.2)
Python 3.10.8, Qt 5.15.4, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: C/Default (C)
Installed mods: 
  * MeshRemodel 1.8919.0
  * Launcher
  * InventorLoader 1.4.0
  * Render 2023.2.6
  * Silk 0.1.3
  * Defeaturing 1.2.0
  * A2plus 0.4.60n
  * Manipulator 1.5.0
  * fasteners 0.4.59
  * lattice2 1.0.0
  * sheetmetal 0.2.63
  * toSketch 1.0.1
  * Curves 0.6.13
  * Assembly4 0.50.2
  * Assembly3 0.12.0
```

### Subproject(s) affected?

PartDesign

### Problem description

When creating an additive coil in PartDesign, If I set the cone angle to a negative number it reverses the coil.
I expected the diameter of the coil to decrease, but instead it reverses the coil and increases it's diameter.

Coil with positive cone angle value:
Screenshot 2023-08-20 at 21 07 00

Coil with negative cone angle value:
Screenshot 2023-08-20 at 21 07 27

### Anything else?

_No response_

### Code of Conduct

bgbsww:
Fix #10284.

### DIFF
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -455,9 +455,11 @@ TopoDS_Shape Helix::generateHelixPath(double startOffset0)
         radiusTop = radius + turns * growth;
     else
         radiusTop = radius + height * tan(Base::toRadians(angle));
+
     //build the helix path
     //TopoShape helix = TopoShape().makeLongHelix(pitch, height, radius, angle, leftHanded);
     TopoDS_Shape path = TopoShape().makeSpiralHelix(radius, radiusTop, height, turns, 0, leftHanded);
+
     /*
      * The helix wire is created with the axis coinciding with z-axis and the start point at (radius, 0, 0)
      * We want to move it so that the axis becomes aligned with "dir" and "pnt", we also want (radius,0,0) to
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -455,11 +455,9 @@ TopoDS_Shape Helix::generateHelixPath(double startOffset0)
         radiusTop = radius + turns * growth;
     else
         radiusTop = radius + height * tan(Base::toRadians(angle));
-
     //build the helix path
     //TopoShape helix = TopoShape().makeLongHelix(pitch, height, radius, angle, leftHanded);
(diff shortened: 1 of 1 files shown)


chennes:
Can you explain a little more about what this fix is doing? I'm not that familiar with the code here.

bgbsww:
Well, I certainly can't claim any expertise, but TopoShape::MakeSpiralHelix takes a 0 parameter to mean maximum number of breakperiods ( 1000 ) instead of 1, and that causes an increase in resolution that seems to solve this bug.  Also, the other two uses of this method also used this parameter value.

I'll freely admit that I just tested to make sure I didn't break anything obvious without fully understanding why this change is required to fix the situation.

wwmayer:
Do you have a test file? I tried to reproduce this issue and if I set a negative angle then I get the same output that was considered incorrect.

bgbsww:
[NegativeHelix.zip](https://github.com/FreeCAD/FreeCAD/files/13502688/NegativeHelix.zip)

Hmm, I think I may have spotted the discrepancy:  If the underlying sketch is on the origin, it doesn't look like it is reversed.  But does this make geometric sense?  @wwmayer

wwmayer:
As a test create a sketch with a circle on the X axis at (-50, 0) and set its radius to 5.0.
If you create a PD helix set pitch = 40, height = 80 and angle = 30. If you change angle to -30 the behaviour is like before.

bgbsww:
Confirmed.  Same circle at (50,0) works correctly.  I've got some digging to do.  Thanks!

bgbsww:
I think #11598 cleans this up too.

[NegativeHelix.zip](https://github.com/FreeCAD/FreeCAD/files/13520236/NegativeHelix.zip)

maxwxyz:
@FlachyJoe this was the PR for the previous fix

PR #11406, opened 2023-11-15
Add readonly length properties to part helix and spiral

### LINKED ISSUE #5995: [Feature Request] Part: Show spiral length in the property editor (of Part Helix and Part Spiral (derived from Helix) objects)
Issue imported from https://tracker.freecad.org/view.php?id=4159

* **Reporter:** vocx
* **Date submitted:** 10/9/2019
* **FreeCAD version:** 
* **Category:** Feature
* **Status:** new
* **Tags:** 

# Original report text

Part Helix and Part Spiral (which are derived from Helix) objects don't show a length property in the property editor.

The length is included in the Shape, so it just has to be displayed as a read-only value.
```
obj.Shape.Length
```

See this thread [Length of a spiral](https://forum.freecadweb.org/viewtopic.php?f=3&t=39884)

# Other bug information

* **Priority:** normal
* **Severity:** feature
* **Category:** Feature
* **Updated:** 2/6/2021

bgbsww:
Fix #5995

### DIFF
--- a/src/Mod/Part/App/PrimitiveFeature.cpp
+++ b/src/Mod/Part/App/PrimitiveFeature.cpp
@@ -802,6 +802,9 @@ App::DocumentObjectExecReturn *Helix::execute()
         Standard_Real myRadiusTop = myRadius + myHeight * tan(myAngle/180.0f*M_PI);
 
         this->Shape.setValue(TopoShape().makeSpiralHelix(myRadius, myRadiusTop, myHeight, nbTurns, mySegLen, myLocalCS));
+        // props.Mass() may seem a strange way to get the Length, but 
+        // https://dev.opencascade.org/doc/refman/html/class_b_rep_g_prop.html#ab1d4bacc290bfaa8df13dd99ae7b8e70
+        // confirms this.
         GProp_GProps props;
         BRepGProp::LinearProperties(Shape.getShape().getShape(), props);
         Length.setValue(props.Mass());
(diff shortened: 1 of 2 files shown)


src/Mod/Part/App/PrimitiveFeature.cpp
```diff
@@ -799,6 +802,9 @@ App::DocumentObjectExecReturn *Helix::execute()
         Standard_Real myRadiusTop = myRadius + myHeight * tan(myAngle/180.0f*M_PI);
 
         this->Shape.setValue(TopoShape().makeSpiralHelix(myRadius, myRadiusTop, myHeight, nbTurns, mySegLen, myLocalCS));
+        GProp_GProps props;
+        BRepGProp::LinearProperties(Shape.getShape().getShape(), props);
+        Length.setValue(props.Mass());
```
yorikvanhavre, line 807: Are you sure this is correct? Do we get the Mass to get the Length?

FlachyJoe: I think it's ok.
[BRepGProp::LinearProperties Reference](https://dev.opencascade.org/doc/refman/html/class_b_rep_g_prop.html#ab1d4bacc290bfaa8df13dd99ae7b8e70):
> For this computation no linear density is attached to the edges. So, for example, the added mass corresponds to the sum of the lengths of the edges of S.

bgbsww: That's consistent with how I researched this, and I couldn't find another way to get the Length.  I agree, it was weird.

kadet1090: This could use comment in the code with reference to docs to not confuse other developers.

bgbsww: That's a good idea.  Done.

PR #11407, opened 2023-11-15
Gui: QuantitySpinBox - keep format on setValue

abdullahtahiriyo:
Problem:
When the value set is a double, a new Quantity is created, which does not follow the previously configured format (setDecimals).

Solution:
To copy the old format to the newly created quantity.


abdullahtahiriyo:
@wwmayer 

I wonder if you also want to copy the format when a Quantity type is set (in addition to when a double is set, that is this PR). I can add this if you want.

Feel free to merge, or tell me how you want it, or, of course, just fix it as a separate PR to your liking and close this PR. Just trying to put less on your shoulders.

wwmayer:
Thanks, the PR is fine.

PR #11409, opened 2023-11-15
Base: Fix clang-tidy reports

wwmayer:
Fix many reports of the categories `readability-*, performance-*, bugprone-* and cppcoreguidelines-*`

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -152,7 +152,7 @@ CheckOptions:
   - key:             google-readability-braces-around-statements.ShortStatementLines
     value:           '1'
   - key:             bugprone-reserved-identifier.AllowedIdentifiers
-    value:           ''
+    value:           '_object;_Precision'
   - key:             cppcoreguidelines-pro-type-member-init.IgnoreArrays
     value:           'false'
   - key:             readability-else-after-return.WarnOnUnfixable
--- a/src/Base/Axis.cpp
+++ b/src/Base/Axis.cpp
@@ -38,9 +38,9 @@ void Axis::reverse()
 
 Axis Axis::reversed() const
 {
-    Axis a(*this);
-    a.reverse();
-    return a;
+    Axis axis(*this);
+    axis.reverse();
+    return axis;
 }
 
 void Axis::move(const Vector3d& MovVec)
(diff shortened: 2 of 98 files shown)


wwmayer:
Build failed due to the gazillion of warnings caused by OndselSolver code

.clang-tidy
```diff
@@ -152,7 +152,7 @@ CheckOptions:
   - key:             google-readability-braces-around-statements.ShortStatementLines
     value:           '1'
   - key:             bugprone-reserved-identifier.AllowedIdentifiers
-    value:           ''
+    value:           '_object;_Precision'
```
chennes, line 155: Technically "_Precision" is reserved in C++ -- any chance we can rename?

wwmayer: clang-tidy claims this for all identifiers that start with an underscore.

chennes: I think only those that begin with an underscore followed by a capital letter -- that's the actual rule in the C++ standard.

chennes: Maybe I'm misunderstanding the footnotes in section 5.10 on identifiers:
> In addition, some identifiers are reserved for use by C++ implementations and shall not be used otherwise; no
> diagnostic is required.
> —(3.1) Each identifier that contains a double underscore __ or begins with an underscore followed by an
> uppercase letter is reserved to the implementation for any use.
> —(3.2) Each identifier that begins with an underscore is reserved to the implementation for use as a name in
> the global namespace.

I thought that 3.2 meant that you couldn't use a leading underscore in the global namespace. Is Clang interpreting it to mean that you can't use one anywhere at all? I mean, I guess the clang folks would know.

wwmayer: Anyway, I don't insist on keeping `_Precision` but is it OK to change this in a further PR?

chennes: Yes, certainly.

wwmayer: Are there any further issues?

chennes: No, that was the only thing that caught my eye. I am assuming you'd like to preserve the individual commits, I'll do a merge.

wwmayer: Yes, as otherwise it's way too complex (when looking later at it) to get the individual changes.

PR #11415, opened 2023-11-16
Draft: prefs: remove copymode, useSupport no longer via UI

Roy-043:
This PR removes the copymode (Global copy mode) preference.

Forum topic:
https://forum.freecad.org/viewtopic.php?t=82679

Additionally useSupport is removed from preferences-draft.ui, turning it into a fine-tuning preference. Use of this feature is rare and may not be supported in the future. No need to present this preference to new users in the UI.

Forum topic:
https://forum.freecad.org/viewtopic.php?t=80689


yorikvanhavre:
Great move... more and more I think parameters can simply be used to remember a state and don't need necessarily to be exposed to the user in the preferences dialog.

yorikvanhavre (approved):

Issue #7039, opened 2022-06-17
[Feature Request] Box Selection for 


ToniTen:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?f=35&t=69436

### Subproject(s) affected?

Techdraw

### Idea description

Add a way to use box selection to select individual lines in a Techdraw view, so they can, for example, have their display options using TechDraw_DecorateLine. 

This is specially useful for cases when the 3D part comes from a STEP file that includes large amounts of geometry, that ends up as large amounts of lines, as each operation will take a noticeable amount of time, and you may want to select a large number of very small lines.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #11417, opened 2023-11-16
[TechDraw] Multiselection mode for vertices/edges/faces

pavltom:
When creating a larger drawing, the need to hold the Ctrl key every moment is quite tedious . This pull request adds "Enable Multiselection Mode" option to TechDraw's General preferences, which allows to add/remove vertices, edges and faces in the same view to/from the selection by plain clicking. If more than one item in the tree is selected, the multiselection mode is not activated.

The pull request consist of 2 commits fixing the issues I have encountered during the development, then 4 commits cleaning the code or removing obsolete features and finally the single commit implementing the multiselection mode.

The way how multiselection mode for TechDraw is implemented is rather a "hack". Unfortunately, QGraphicsItem does not support SelectionMode in the manner the QAbstractItemView does. The only way how to achieve the "MultiSelection" behavior is to set Qt::ControlModifier on the event passing through mousePressEvent/mouseReleaseEvent handlers when suitable. On the other hand, it works consistently well and I did not encounter any problems so far.

If for some reasons the multiselection mode option gets rejected, the pull request's fix and refactoring commits are ready to be used on their own.

I hope you will find this new feature useful, in case of any problems please let me know.

### DIFF
--- a/src/Mod/TechDraw/App/DrawView.cpp
+++ b/src/Mod/TechDraw/App/DrawView.cpp
@@ -455,7 +455,7 @@ void DrawView::setPosition(double x, double y, bool force)
     if ( (!isLocked()) ||
          (force) ) {
         double currX = X.getValue();
-        double currY = X.getValue();
+        double currY = Y.getValue();
         if (!DrawUtil::fpCompare(currX, x, 0.001)) {    // 0.001mm tolerance
             X.setValue(x);
         }
--- a/src/Mod/TechDraw/Gui/DlgPrefsTechDrawGeneral.ui
+++ b/src/Mod/TechDraw/Gui/DlgPrefsTechDrawGeneral.ui
@@ -868,6 +868,62 @@ for ProjectionGroups</string>
     </widget>
    </item>
    <item>
+    <widget class="QGroupBox" name="gbSelection">
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+       <horstretch>0</horstretch>
+       <verstretch>0</verstretch>
+      </sizepolicy>
+     </property>
+     <property name="minimumSize">
+      <size>
+       <width>0</width>
+       <height>0</height>
+      </size>
+     </property>
+     <property name="maximumSize">
+      <size>
+       <width>16777215</width>
+       <height>100</height>
+      </size>
+     </property>
+     <property name="title">
+      <string>Selection</string>
+     </property>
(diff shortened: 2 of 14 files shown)


WandererFan:
I've desk checked the 6 clean up commits, and they look ready to merge.

Still looking at the greedy selection change.

luzpaz:
@pavltom Neat! Would you mind creating a short screencast of this PR in action ?

WandererFan:
Nice.  We probably want to add a toggle somewhere that is more accessible that Preferences.  Maybe context menu?

pavltom:
@WandererFan At first I was experimenting with something like permanent Ctrl-click for all selectable items. Very quickly this showed totally unusable. One was forced all the time to "deselect" last touched dimensions etc. But once the Ctrl-click was applied only to the selection subitems (vertices/edges/faces) and their parent view (so the selection is not lost if one misses intended edge or vertex), I will probably never clear that option checkbox. I was thinking about adding a button next to the "View Frames On/Off" toolbar button, but I did not find any need to switch the modes during my work. Anyway, if somebody would long for these "quick" switches, it is fairly easy to implement it, either as toolbar button or right-click context menu option. I personally would probably vote for the toolbar option, but we could spawn a discussion on the forum.

@luzpaz I have tried to create a short record to illustrate the new selection mode, but maybe it is simpler to just describe it with words :-) See for yourself, if you find it useful:

[TD_Multiselection_mode.webm](https://github.com/FreeCAD/FreeCAD/assets/25648708/b0bf7a6b-2d8b-451a-bdaa-de264d40c9a1)

FEA-eng:
@pavltom Maybe you could also add support for box selection in TechDraw on this occasion ? Would be very helpful and it's already available in the Sketcher and as [Std BoxSelection](https://wiki.freecad.org/Std_BoxSelection).

maxwxyz:
@pavltom nice! What is the default, on or off? Does it reset properly when resetting the TechDraw preferences?

pavltom:
@FEA-eng Yes, box selection would be nice, maybe we could create a feature request for that?
@maxwxyz The default value is **off** to preserve the existing behavior. There is not much to reset, either it is on or off. The current selection stays always untouched.

FEA-eng:
> @FEA-eng Yes, box selection would be nice, maybe we could create a feature request for that?

I've found an existing FR for this: https://github.com/FreeCAD/FreeCAD/issues/7039

PR #11076, opened 2023-10-17
Vscode integration again

Pesc0:
follows #9663 
keep discussion centralized there

PR #11420, opened 2023-11-17
Allow automatic tests discovery

Pesc0:


wwmayer:
What does this PR try to achieve?

Pesc0:
Enables ctest to discover all the unit tests specified by googletest. This PR does nothing really, but enables vscode to display and run the tests. This is part of #11076

wwmayer:
So, it's just a thing to display the unit tests in your IDE?  I asked because I use QtCreator and for me unit tests have always been visible.

Pesc0:
Basically yes, its just for vscode. I don't know if there's other use cases. Anyway it should be harmless by itself

luzpaz:
The principle question is why does Qt Creator find it and not vscode ?

wwmayer:
Because it's smarter.

PR #11421, opened 2023-11-17
Vscode split 4  upgrade python debugger

Pesc0:


src/Gui/RemoteDebugger.py
```diff
@@ -25,6 +25,7 @@
 import FreeCAD as App
 import FreeCADGui as Gui
 from PySide import QtGui
+from freecad.utils import get_python_exe
```
Pesc0, line 28: Why is this not `FreeCAD` ?  See https://github.com/FreeCAD/FreeCAD/pull/9663#issuecomment-1622583028

Pesc0:
@chennes must approve this first

PR #11427, opened 2023-11-17
[Sketcher] Fix autoconstraints of arcslot tool.

PaddleStroke:
Fix autoconstraints of arcslot tool.


PaddleStroke:
@adrianinsaval This is a very small (+3) bug fix for the newly introduced arcslot tool.

PR #11430, opened 2023-11-17
Part : fix edge creation of Geom2d curves

tomate44:
Hi,
Currently, the toShape() method of Geom2dCurve is not consistent upon its various signatures.
When no face or surface support is passed, the resulting edge contains both a 3D curve and a PCurve (on the XY plane).
When a face or surface support is passed, the resulting edge contains only a 3D curve. The PCurve data gets lost.
A face built from such edges is not valid.
The proposed fix avoids the loss of the PCurve data in these situations.
Test code:
```
import FreeCAD
import Part
vec2 = FreeCAD.Base.Vector2d

def info(edge):
    print(f"Curve 3D : {edge.Curve}")
    print(f"Curve 2D : {edge.curveOnSurface(0)}")

surf = Part.Cylinder()

p1 = vec2(1.0, -1.0)
p2 = vec2(2.0, -1.0)

l12 = Part.Geom2d.Line2dSegment(p1, p2)

e1 = l12.toShape()
info(e1)

e2 = l12.toShape(surf)
info(e2)


```

### DIFF
--- a/src/Mod/Part/App/Geom2d/Curve2dPyImp.cpp
+++ b/src/Mod/Part/App/Geom2d/Curve2dPyImp.cpp
@@ -28,6 +28,8 @@
 # include <BRepAdaptor_Surface.hxx>
 # include <BRepBuilderAPI_MakeEdge.hxx>
 # include <BRepBuilderAPI_MakeEdge2d.hxx>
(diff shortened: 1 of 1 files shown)


luzpaz:
Thanks @tomate44 !

PR #11454, opened 2023-11-20
Translations

yorikvanhavre:
Latest crowdin translations


chennes:
I don't know why this won't merge, so I just ran a new cycle and merged that one instead.

PR #11457, opened 2023-11-20
Use PyObject_CallObject rather than interpreting generated code

kpemartin:
Rather than generating Python code and interpreting it, which, in C++, is fraught with issues of creating the correct syntax for a str token when the text contains certain special characters like double-quote and backslash, the modified code makes call(s) to PyObject_CallObject which takes the C++ string and makes the appropriate conversion itself. The steps in building the Position passed to make_text are also done using c++ objects until the final Python Placement object is needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ install_manifest.txt
 /conda/environment.yml
 OpenSCAD_rc.py
 tags
+/CMakeUserPresets.json
 
 # crowdin file
 src/Tools/freecad.zip
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,287 @@
+{
+    "version": 6,
+    "cmakeMinimumRequired": {
+      "major": 3,
+      "minor": 14,
+      "patch": 0
+    },
+    "configurePresets": [
+      {
+        "name": "common",
+        "hidden": true,
+        "cacheVariables": {
+          "FREECAD_USE_PYBIND11": {
+            "type": "BOOL",
+            "value": "ON"
+          }
+        }
+      },
+      {
+        "name": "debug",
+        "displayName": "Debug",
+        "description": "Default debug profile",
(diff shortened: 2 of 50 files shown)


luzpaz:
@kpemartin thanks! Do you know how to pull latest main, rebase and force push this branch in git ?

kpemartin:
Clearly I haven't figured that out yet! So far each time I think I've figured out how and end up with a mess instead.
I think I have the pull and rebase part working, my local copy has (well, had) my changes at the top of the list. Perhaps the key here is the "force" push...

kpemartin:
Since only one file has actually changed, I think the easiest way to clean up my mess is to discard the branch, create a new one, and check in my one changed file.

luzpaz:
I used to do that till i finally learned. 

```shell
git pull upstream main
git rebase
git push --force
```

kpemartin:
I have a branch on github, and a clone on my local computer. Assuming I run these on my computer will the pull pull directly from FreeCAD/FreeCAD and leave my github branch unchanged? Will this cause me problems or does the --force fix that too?

What I had been doing was updating my github branch first, then doing a pull on my local computer (from origin rather than upstream).

I think I had the rebase figured out, it essentially floats my local changes up to the top so they are based on the latest origin version. This doesn’t actually change my changed files unless the newly-fetched origin commits change the same file I’ve changed in which case I have a merge conflict to fix.

After that I think I ran into the push not wanting to do what I asked, and I guess what was missing was the –force option.

If I want to combine several of my local commits so the commit log doesn’t contain a play-by-play of my false starts and temporary stub code, I believe I can use rebase --interactive to do this, correct?
-Kevin

luzpaz:
> If I want to combine several of my local commits so the commit log doesn’t contain a play-by-play of my false starts and temporary stub code, I believe I can use rebase --interactive to do this, correct? 

Yes, `git rebase--interactive` is correct. Sorry for the delay in response.

PR #11461, opened 2023-11-20
Tests: Set test executables in a central place

wwmayer:
Clean-up work after #11420


wwmayer:
@Pesc0 
Can you test if this still works for you, please? It's only to avoid to keep the list of test executables at two places.

Pesc0:
@wwmayer It does work, thanks!

wwmayer:
Great, thanks!

PR #11462, opened 2023-11-20
[3rdParty] fix typo in Decoder.cpp

eltociear:
avaiable -> available


luzpaz:
Is there an upstream library we should submit the fix to ?

PR #11468, opened 2023-11-20
fix default state of combo view #backport

funny0facer:
Hi, this is my first PR to FreeCAD.

This is related to the following forum thread: [Start-up, Combo-View: task panel instead of Model panel](https://forum.freecad.org/viewtopic.php?t=74583)

I know, that I should actually base a PR on "main", but as this has - at least from my observation - no impact on 0.22, I chose 0.21 as a base.
I saw that a 0.21.2 is in the making... I hope, this is not too late.
thanks

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -2337,7 +2337,7 @@ void Application::runApplication()
 
     // Reactivate the startup workbench
     app.activateWorkbench(start.c_str());
-
+    Gui::Control().showModelView();
     Instance->d->startingUp = false;
 
     // gets called once we start the event loop


chennes:
Thanks for the PR -- calling @adrianinsaval for an evaluation.

adrianinsaval:
I could be very wrong but this seems a hardcoded hack to suit the preference of one person

funny0facer:
I am sorry. The idea for the solution itself was not my idea. See referenced forum post.
I will not insist on a merge. I agree it is a quite minor inconvinience, even though i am not completely alone (I did not create the forum post in the beginning)
In general, I am looking forward for the Tasks view in 0.22 which seems quite nice with its auto-hide functionality.

adrianinsaval:
I think we just leave this as is. Next version is not so far

funny0facer:
this is fine for me.

PR #11476, opened 2023-11-21
Update OndselSolver submodule

adrianinsaval:
This should reduce/fix the number of warnings on CI


0penBrain:
Couldn't we expect that as a submodule, OndselSolver is pushed to FC only as tagged versions with some clean code.
It generates lot of traces in the build log that makes other problems hard to find.

Some warnings like hidden overloaded functions would deserve some attention.

PR #11487, opened 2023-11-22
Fix casts

mosfet80:
int_number * float_number return a int_number.
float_number * int_number return a float_number.
swap int and float and remove the cast, this also increase the result precision.

### DIFF
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -531,7 +531,7 @@ void NaviCubeImplementation::addButtonFace(PickId pickId, const SbVec3f& directi
         case PickId::DotBackside: {
             int steps = 16;
             for (int i = 0; i < steps; i++) {
-                float angle = 2.0f * M_PI * ((float)i+0.5) / (float)steps;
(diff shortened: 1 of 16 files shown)


wwmayer:
Sure?

```
#include <iostream>

int main()
{
    auto res1 = 2 * 1.2F;
    std::cout << typeid(res1).name() << '\n';

    auto res2 = 1.2 * 2F;
    std::cout << typeid(res2).name() << '\n';
}
```
The output of this test application is:
f
f

which means that it's a float in both cases.

wwmayer:
See also: https://stackoverflow.com/questions/5563000/implicit-type-conversion-rules-in-c-operators

mosfet80:
...correct. freecad uses g++17 or higher.
Sometimes I develop on older compilers < g++11 and these changes make a difference.

wwmayer:
Hmm, when I rewrite the code
```
#include <iostream>
#include <typeinfo>

int main()
{
    std::cout << typeid(2 + 1.23456789F).name() << '\n';

    std::cout << typeid(1.23456789F + 2).name() << '\n';
}
```
and build it with the option `--std=c++98` it still says that both are floats.

PR #11494, opened 2023-11-22
[Sketcher] Introduce an optional unified version of coincident and pointOnObject

PaddleStroke:
Coincident and pointOnObject are a single tool in most other CAD software. Having them separated is pointless because the selection for both are different. So a single tool can create both constraints.
So this PR introduce a unified version of coincident and pointOnObject.

This was discussed on the forum, and the discussion did not reached a consensus. So this is implemented as an option. Nothing changes by default. A setting is added in sketcher preference page to unify the commands.

Note this only unify the tool. Not the constraints or coin symbols and so on. Basically if you select 2 points it creates a coincidence. One point and one line and it creates point on object.

This PR also adds a setting to disable auto vertical/horizontal. While it is not related it would have generated conflict with this PR.


obelisk79:
I think this is a good change and has already been discussed among the design working group

howie-j:
Did a 10 min test and i love it :100: 

A small error: Preselecting a point and a line/curve before launching the command gives the following warning:
warning

PaddleStroke:
@howie-j thanks, I had the notif area off so I missed this issue !
It is now fixed

PaddleStroke:
@abdullahtahiriyo please comment on [2deb38a](https://github.com/FreeCAD/FreeCAD/pull/11494/commits/2deb38aa29e47ccc360d868672745447eb183fd6) especially. This looks like dead code to me but I prefer to be sure there wasn't a reason it was left behind.

PaddleStroke:
Note : [921a8f3](https://github.com/FreeCAD/FreeCAD/pull/11494/commits/921a8f304084db47c8cf04e7a6563c3becdfc7b3) is out of topic. It adds a setting to disable auto vertical/horizontal. While it is not related it would have generated conflict with this PR so it was simpler to add it here.

chennes:
@abdullahtahiriyo I'm going to merge this since @PaddleStroke addressed your comment and there don't seem to be any other issues with it (and I think it's a great feature!). If you don't want me to do that in the future, could you please assign the PRs that you intend to handle to yourself? Thanks!

PR #11501, opened 2023-11-23
Base: cppcoreguidelines

wwmayer:


src/Base/GeometryPyCXX.cpp
```diff
@@ -407,3 +407,4 @@ void Vector2dPy::init_type()
 }
 
 }  // namespace Base
+// NOLINTEND(cppcoreguidelines*)
```
chennes, line 410: Unmatched - the only NOLINTBEGIN I see in this one is `// NOLINTBEGIN(readability-identifier-length)`, and that was ended on line 381.

wwmayer: I copied that from an older version I was working on some time ago...

PR #11502, opened 2023-11-23
[Sketcher] Delete all constraints : remove from toolbar. Add to widget contextual menu

PaddleStroke:
Removes 'Delete all constraints' from the tool toolbar. Keep it in the menu (same as 'Delete all geo'). And add context menu entries for those in their respective widget's list.
image


Pesc0:
potential for misclick near normal delete?

PaddleStroke:
Not really because the delete all ask for confirmation.
But two people already say the menu entries are not necessary because you can already ctrlA delete. So perhaps its best to not add them.

PaddleStroke:
The additions to right click menu entries in the element and constraint widget have been reverted after people on forum pointed out they are useless as you can simply ctrlA delete

PR #11503, opened 2023-11-23
Draft: Added a preference to turn on/off tool messages

### LINKED ISSUE #7078: [Feature Request] Draft: reduce/eliminate unnecessary output messages

### Forums discussion

https://forum.freecadweb.org/viewtopic.php?p=604611#p604611

### Subproject(s) affected?

Draft

### Idea description

Draft WB tools output a lot of unnecessary messages to the report view, for example:
> 19:55:27 ----------------
> 19:55:27 GuiCommand: Line
> 19:55:27 Pick first point
> 19:55:32 Pick next point

these should be removed since they don't serve any meaningful purpose or at most sent to the log instead

### Anything else?

_No response_

### Code of Conduct

yorikvanhavre:
New property: Draft -> General -> Show tool messages (off by default)
Fixes #7078

### DIFF
--- a/src/Mod/Draft/Resources/ui/preferences-draft.ui
+++ b/src/Mod/Draft/Resources/ui/preferences-draft.ui
@@ -17,7 +17,16 @@
    <property name="spacing">
     <number>6</number>
    </property>
-   <property name="margin">
+   <property name="leftMargin">
+    <number>9</number>
+   </property>
+   <property name="topMargin">
+    <number>9</number>
+   </property>
+   <property name="rightMargin">
+    <number>9</number>
+   </property>
+   <property name="bottomMargin">
     <number>9</number>
    </property>
    <item>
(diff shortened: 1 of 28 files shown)


adrianinsaval:
is it actually worth having this in preference page? might be better to leave it as an advanced parameter only. I can't imagine many people will enable this

Roy-043:
@adrianinsaval In AutoCAD there is a command line that shows prompts. For users coming from that software this may be an important feature.

There is an issue with the top level margins BTW. There is a default margin (10px) that gets added to the 9px margin specified in the file. Some of the Draft and Import tabs have a 19px margin in the UI because of this.

Most preferences tabs have the 10px default margin. But the General > General tab has a 12px margin. I do not know where these values come from. But I am working on the preferences tabs anyway, so the margins can be fixed later.

Let's merge...

yorikvanhavre:
@adrianinsaval Yes, I also think it could be removed at some point, but I thought this is an easier transition, if anybody complains we'll see it and it will be pretty easy for them to restore. If nobody notices, it's a sign we don't need the prefs setting :)

PR #11506, opened 2023-11-24
Add conda and mamba to the conda-devenv.

oursland:
Alter the `conda/environment.devenv.yml` to include `conda` and `mamba` as dependencies.

`mamba-devenv` will remove any package not listed in this file, removing the packages `conda` and `mamba` from the newly created conda environment.  This renders the newly created conda environment inoperable.  I believe this is a behavioral difference with `conda devenv` which was previously used.

### DIFF
--- a/conda/environment.devenv.yml
+++ b/conda/environment.devenv.yml
@@ -60,6 +60,7 @@ dependencies:
 - cmake
 - coin3d
 - compilers
+- conda
 - conda-build
 - conda-devenv
 - conda-smithy
@@ -72,6 +73,7 @@ dependencies:
 - graphviz
 - hdf5
 - libcxx
+- mamba==1.4.9
 - matplotlib
 - ninja
 - numpy


conda/environment.devenv.yml
```diff
@@ -60,9 +60,8 @@ dependencies:
 - cmake
 - coin3d
 - compilers
-- conda-build
```
adrianinsaval, line 63: why remove conda-build? in any case we could replace it with boa

conda/environment.devenv.yml
```diff
@@ -60,9 +60,8 @@ dependencies:
...
+- conda
 - conda-devenv
-- conda-smithy
```
adrianinsaval, line 65: same here

conda/environment.devenv.yml
adrianinsaval: yes they aren't strictly necesary, but are still worth including IMO. Are there any problems with these packages?

oursland: I've restored the files, if desired a separate PR will be made to remove them.

PR #11507, opened 2023-11-24
[Sketcher] Improve angle & radius constraint rendering

PaddleStroke:
This PR improve the rendering of the (2lines) angle and radius SoDatumLabel
image
image

### DIFF
--- a/src/Gui/SoDatumLabel.cpp
+++ b/src/Gui/SoDatumLabel.cpp
@@ -91,6 +91,8 @@ SoDatumLabel::SoDatumLabel()
 
     SO_NODE_ADD_FIELD(param1, (0.f));
     SO_NODE_ADD_FIELD(param2, (0.f));
+    SO_NODE_ADD_FIELD(param4, (0.f));
+    SO_NODE_ADD_FIELD(param5, (0.f));
 
     useAntialiasing = true;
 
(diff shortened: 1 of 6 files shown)


PaddleStroke:
@FlachyJoe please check here, I have added arc helpers for radius constraints. This would be the best for arc - something distances. What is the status on those btw?

obelisk79:
This is quite interesting. Seems like an improvement based off what I see in the screenshots.

PaddleStroke:
@wwmayer this is touching SoDatumLabel which you know so perhaps you would be willing to review this PR please ?

wwmayer:
The changes in SoDatumLabel are OK.

PaddleStroke:
@adrianinsaval care to check this one out? Werner already reviewed changes in SoDatumLabel

FlachyJoe:
> This would be the best for arc - something distances

Certainly yes ! But in fact arc helper is needed for all the constraints, see floating icons for tangencies in https://github.com/FreeCAD/FreeCAD/pull/9166#issuecomment-1493435856

PaddleStroke:
Do those floating icons actually happen in real life work ? I am not sure there are so many tangencies to arcs in random positions that would end up floating.
If it does happen and is a real issue, then a possibility would be to draw arc helpers for geometric constraints when needed as well. But this may be much more work as geometric constraints do not have soDatumLabels (afaik).

FlachyJoe:
@PaddleStroke we understand the left tangent only with the right part of the sketch 
image

PaddleStroke:
I see.
But in these cases, do you think the user is going to toggle arc helpers ? It's likely that if you don't understand the sketch you will select associated geometries instead. I'm not sure the arc helpers would help much to understand in the sketch you screenshoted.

PaddleStroke:
But yes arc helpers similar to the one I made for radius in this PR would probably help I guess.

maxwxyz:
@FlachyJoe it is still present, shall I create an issue?
grafik

FlachyJoe:
@maxwxyz Not sure why but 2 commits was lost.

PR #11508, opened 2023-11-24
Fixes #11505

### LINKED ISSUE #11505: [Problem] Incorrect calls in `PyArg_ParseTuple`

### Problem description

FreeCAD uses functions from Python C-API: `PyArg_ParseTupleAndKeywords` and `PyArg_ParseTuple`. 
These functions are variadic (can take the arbitrary number of arguments), so it is hard to detect errors at compilation stage.

I am author of [freecad-stubs](https://pypi.org/project/freecad-stubs/). 
While generating stubs, I encored few errors. 
By request https://github.com/ostr00000/freecad-stubs/issues/7 I decided to report these errors.

Recently (between 0.20 and 0.21) I see most of the errors are fixed, but few remained.
(Probably these functions are rarely used or not reported, otherwise they would be fixed already).

The errors on `main` branch:
  - Not enough arguments:
    - https://github.com/FreeCAD/FreeCAD/blob/d0fb2b8b29fe0428d9dd8aa790b0d6e45c8a9516/src/Mod/Part/App/Geom2d/BSplineCurve2dPyImp.cpp#L171
    - https://github.com/FreeCAD/FreeCAD/blob/d0fb2b8b29fe0428d9dd8aa790b0d6e45c8a9516/src/Gui/Selection.cpp#L2431
  - Incorrect argument order:
    - https://github.com/FreeCAD/FreeCAD/blob/d0fb2b8b29fe0428d9dd8aa790b0d6e45c8a9516/src/Mod/Part/App/TopoShapePyImp.cpp#L2674

### Full version info

```shell
I use source code only, so I fill here commit ref.
Commit revision: d0fb2b8b29fe0428d9dd8aa790b0d6e45c8a9516
Commit date: 22.11.2023
https://github.com/FreeCAD/FreeCAD/commit/d0fb2b8b29fe0428d9dd8aa790b0d6e45c8a9516
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

wwmayer:
Fixes https://github.com/FreeCAD/FreeCAD/issues/11505

### DIFF
--- a/src/Gui/Selection.cpp
+++ b/src/Gui/Selection.cpp
@@ -2428,7 +2428,7 @@ PyObject *SelectionSingleton::sHasSelection(PyObject * /*self*/, PyObject *args)
 {
     const char *doc = nullptr;
     int resolve = 0;
-    if (!PyArg_ParseTuple(args, "|sO!", &doc, &resolve))
+    if (!PyArg_ParseTuple(args, "|si", &doc, &resolve))
         return nullptr;
 
     PY_TRY {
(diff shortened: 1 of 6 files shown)


wwmayer:
Backported 2927126cbcbf3a50

PR #11519, opened 2023-11-24
Gui: Update rotation center indicator after zoom

### LINKED ISSUE #11243: [Problem] CAD Navigation around Window Center broken

Rexbas:
#11243 was not completely fixed. This PR fixes #11243. The problem also occurred when dragging and using the scroll wheel to zoom at the same time. This only happend with rotation mode = window center and _zoom at cursor_ enabled. @maxwxyz can you confirm this works?

The problem is resolved by updating the position of the rotation center, to the new focal point, after zooming. ~~This PR also introduces a small regression which is a flickering rotation center after zooming using the scroll wheel. However, when you start dragging again without zooming then it does not flicker anymore. In my opinion not really a problem to worry about. This flickering appears to be resolved by https://github.com/coin3d/coin/pull/509.~~

### DIFF
--- a/src/Gui/NavigationStyle.cpp
+++ b/src/Gui/NavigationStyle.cpp
@@ -739,6 +739,12 @@ void NavigationStyle::doZoom(SoCamera* camera, float logfactor, const SbVec2f& p
         SbViewVolume vv = camera->getViewVolume(vp.getViewportAspectRatio());
         SbPlane panplane = vv.getPlane(camera->focalDistance.getValue());
         panCamera(viewer->getSoRenderManager()->getCamera(), ratio, panplane, pos, SbVec2f(0.5,0.5));
+
+        // Change the position of the rotation center indicator after zooming at cursor
+        // Rotation mode is WindowCenter
+        if (!rotationCenterMode) {
+            viewer->changeRotationCenterPosition(getFocalPoint());
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


Rexbas:
The flickering is resolved by finding the bounding sphere of the scene after zooming.

maxwxyz:
Seems great, cannot reproduce the glitching with your branch on macOS. Thank you!

PR #11524, opened 2023-11-25
[Gui] Add 20, 25 and 30px marker sizes to assist users of 4K screens

Syres916:
This isn't the first time this has been requested on the forum but see the most recent discussion https://forum.freecad.org/viewtopic.php?t=82927

As I only have a 2K screen they do appear quite large but I'm aware of how much that will be reduced on higher DPI.

Hopefully one of the devs or self compile users with a 4K screen can test this on their setup.

### DIFF
--- a/src/Gui/Inventor/MarkerBitmaps.cpp
+++ b/src/Gui/Inventor/MarkerBitmaps.cpp
@@ -114,6 +114,94 @@ const char diamondFilled15_marker[DIAMOND_FILLED15_WIDTH * DIAMOND_FILLED15_HEIG
 "      xxxx     "
 "       xx      "};
 
+
+//DIAMOND_FILLED_20_20
+const int DIAMOND_FILLED20_WIDTH = 20;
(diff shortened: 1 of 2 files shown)


Roy-043:
The `FreeCADGui.getMarkerIndex` function needs to be updated to support the new sizes. Currently it only goes to 9px.

In Draft we also use `HOURGLASS_FILLED_9_9`. Maybe that shape can be added at higher resolutions as well?

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/11528

Roy-043:
The Draft WB now uses the MarkerSize preference for snap and tracker markers (#11688) and also these new sizes. Everything seems to work well except the `diamondFilled20_marker`. That marker displays as a 'scratched out' square for some reason. I have checked all sizes from 13 - 30 px. Only the 20px marker is affected. The filled diamond marker is used for [Draft Snap Midpoint](https://wiki.freecad.org/Draft_Snap_Midpoint).
diamondFilled20_marker

PR #11532, opened 2023-11-25
[Techdraw] Implement arc length calculation for arbitrary edge(s)

pavltom:
This pull request adds a new command to TechDraw's Extensions: Attributes toolset. As the title implies, it uses OCC functionality to find out the arc length of arbitrary edge or the sum of arc lengths of the selected edges. The resulting balloon holding the datum should look like a referencing dimension, including the formatting. The anchor (origin) of the ballon is placed in the point halving the calculated distance. If more edges are in the selection, please note the order of selecting does matter, so for best results try to select them in a continuous way.

This pull request also includes a fix for the arbitrary face area calculation, where non-unitary scale caused wrong balloon placement. Of course, this fix can be cherrypicked and merged independently.

I hope you will like the new feature, in case of any issues please let me know.


luzpaz:
@pavltom when you get a chance, a screencast would be most welcome :pray:

WandererFan:
the arc length code looks ok.  Need a wiki entry for the new function.

Not really part of this PR, but note that the area annotation does not handle faces with holes correctly.  This is because TD faces are only the outer perimeter with smaller faces stacked on top of larger faces.   To get the correct value,  we would have to get all the faces within the selected large face and do the appropriate OCC operations to generate a perforated face, then take the area.

WandererFan (approved):

pavltom:
So I have managed to grab some free time to create the Wiki entry. Please take a look at [TechDraw ExtensionArcLengthAnnotation](https://wiki.freecad.org/TechDraw_ExtensionArcLengthAnnotation) and feel free to make any additions or changes needed.

PR #11534, opened 2023-11-25
Gui: Add a separate checkbox for enabling spinning animations

Rexbas:
There are some complaints about the infinite spin animation in the OpenInventor and CAD navigation style after dragging.  [here on reddit](https://www.reddit.com/r/FreeCAD/comments/17yhq1f/stop_that_crazy_autorotation/), and here https://github.com/FreeCAD/FreeCAD/issues/11243#issuecomment-1789149331. I personally also find this spin animation slightly annoying but I would still like to use the other animations so disabling animations is not the solution.

This PR groups settings for animations together and adds a separate checkbox to enable the spin animations. This checkbox is disabled by default. With this change it is still possible to have animations when selecting standard views or the navigation cube but the annoying spin animation will only be played when you enable it.

### DIFF
--- a/src/Gui/DemoMode.cpp
+++ b/src/Gui/DemoMode.cpp
@@ -180,7 +180,7 @@ void DemoMode::onAngleSliderValueChanged(int v)
         SbRotation rot(SbVec3f(-1, 0, 0), angle);
         reorientCamera(cam ,rot);
         this->oldvalue = v;
-        if (view->getViewer()->isAnimating()) {
+        if (view->getViewer()->isSpinning()) {
             startAnimation(view);
         }
     }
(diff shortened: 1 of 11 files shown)


chennes:
@obelisk79 can you and the UI/UX team weigh in on this change?

obelisk79:
@Rexbas Do these animation settings now also apply to other navigation styles? And where are the settings located? I won't have a chance to build with this pr for several hours to test.

Rexbas:
The settings are located in preferences->display->navigation.
Screenshot_20231127_183350

@obelisk79 the enable spinning animations checkbox only applies to navigation styles that are using the spinning animations. As far as I am aware this is only the OpenInventor and CAD navigation style. Navigation styles that don´t use spinning animations are not affected by this new checkbox.

obelisk79:
I'll bring it up with the design working group, but I think it may be nice to have the animation options available to the other navigation styles. I have no idea if that would be simple to include however. I just wanted to mention it.

Rexbas:
Unless there are people who really want an infinite spinning animation in all navigation styles and are actually going to use it, I think it is not really worth the time to try and make that work. Especially because it is considered annoying in the navigation styles that are already using the spin animation.

Note that the _Animation_ checkbox above _Animation duration_, does work for all navigation styles, like it used to do.

PR #11535, opened 2023-11-25
[Part::Mirroring] Add support for mirror plane reference object

mwganson:
This PR adds a new property to Part::Mirror objects:

MirrorPlane (type App::PropertyLinkSub), which serves as a reference to the mirror plane.  This reference can be a datum plane, a part plane primitive, an origin plane, a plane face (such as a face of a cube), a circle, or any part::feature object that contains only a single face or a single circle.  App::Links are supported.

The dialog has been modified to support the new property, including a selection gate that would not allow selection of unsupported reference objects.

Currently, Base and Normal properties are used to set the mirror plane.  The code works to calculate these properties based on the MirrorPlane reference.  The Base and Normal properties are set to readonly mode when the MirrorPlane property contains a reference to an object.

Part::Mirror objects have a special editor to edit the mirror plane via a coin-based dragger mechanism.  That dialog, which also edits the Normal and Base properties, is disabled when the MirrorPlane property contains a reference.

Note: If the file containing the new object is opened by and saved by an older version of FreeCAD, and then reopened in the new version, then the MirrorPlane reference will be lost.  But when a legacy file is opened in the new version and that legacy file contains a Part::Mirror object, that object gets the new MirrorPlane property, which is functional.

Forum topic:
https://forum.freecad.org/viewtopic.php?t=82832&sid=e1fb23adcd0c062d37365290140fdc76

### DIFF
--- a/src/Mod/Part/App/FeatureMirroring.cpp
+++ b/src/Mod/Part/App/FeatureMirroring.cpp
@@ -23,13 +23,28 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <BRepBuilderAPI_Transform.hxx>
+# include <BRep_Tool.hxx>
 # include <gp_Ax2.hxx>
+# include <gp_Circ.hxx>
 # include <gp_Dir.hxx>
 # include <gp_Pnt.hxx>
 # include <gp_Trsf.hxx>
+# include <BRepAdaptor_Curve.hxx>
+# include <BRepAdaptor_Surface.hxx>
+# include <gp_Pln.hxx>
+# include <Geom_Plane.hxx>
+# include <TopoDS.hxx>
+# include <TopoDS_Face.hxx>
+# include <TopExp_Explorer.hxx>
 #endif
 
+#include <Mod/Part/App/PrimitiveFeature.h>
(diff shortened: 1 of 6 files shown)


luzpaz:
@mwganson an aside, can you check out https://github.com/FreeCAD/FreeCAD/discussions/11433

luzpaz:
@mwganson can you also kindly provide a screencast of this feature in action ?

luzpaz:
@mwganson indeed. I was hoping we could have something short for the v1.0 Release Notes once this gets merged

Roy-043:
There is an issue with the selection gate if the task panel is cancelled.

Steps to reproduce:
1. Create a Part_Box.
2. Call Part_Mirror.
3. The button reads "Selecting".
4. Click a face of the box.
5. Press the Cancel button.
6. Try selecting the box in the Tree view.
7. Result: this is not possible selection gate is still active.

mwganson:
Thanks, Roy.  I will look into it.  In the interim the gate can be cleared with the Part_SelectFilter command in the Gui or in the python console with Gui.Selection.removeSelectionGate()

PR #5480, opened 2022-02-02
Sketcher:  copy,paste (or export/import as xml) geometries with ctrl-c/x ctrl-v

PaddleStroke:
Video and description : https://forum.freecadweb.org/viewtopic.php?f=9&t=65892
Description : add support to copy/paste geometries in sketcher using ctrl-X/C and ctrl-V
Geometries (and relevent constraints) are copied to computer clipboard as xml text. Enabling exporting and paste in other Freecad instances or even share on forum.

Thank you for creating a pull request to contribute to FreeCAD! To ease integration, we ask you to conform to the following items. Pull requests which don't satisfy all the items below might be rejected. If you are in doubt with any of the items below, don't hesitate to ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10)!

- [x]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [x]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [x]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [x]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [x]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [x]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the [FreeCAD bug tracker issue number](https://freecadweb.org/wiki/tracker#GitHub_and_MantisBT) in case a particular commit solves or is related to an existing issue on the tracker. Ex: `Draft: fix typos - fixes #0004805`

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.20 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=56135).

---

PR #11537, opened 2023-11-26
[Sketcher] Add copy/cut/paste sketcher commands

PaddleStroke:
Add support to copy/paste geometries in sketcher using ctrl-X/C and ctrl-V
Geometries (and relevent constraints) are copied to computer clipboard as python command text. Enabling exporting and paste in other Freecad instances or even share on forum.

Video description : https://forum.freecadweb.org/viewtopic.php?f=9&t=65892

This is a rework of https://github.com/FreeCAD/FreeCAD/pull/5480
So it appeared that creating commands in sketcher with the same shortcuts override CommandDoc general commands. Which solves the initial problem.

### DIFF
--- a/src/Mod/Sketcher/App/PythonConverter.cpp
+++ b/src/Mod/Sketcher/App/PythonConverter.cpp
@@ -59,11 +59,11 @@ std::string PythonConverter::convert(const Part::Geometry* geo, Mode mode)
     return command;
 }
 
-std::string PythonConverter::convert(const Sketcher::Constraint* constraint, bool useLastGeoIdVar)
+std::string PythonConverter::convert(const Sketcher::Constraint* constraint, GeoIdMode geoIdMode)
 {
     // addConstraint(Sketcher.Constraint('Distance',%d,%f))
     std::string command;
-    auto cg = process(constraint, useLastGeoIdVar);
+    auto cg = process(constraint, geoIdMode);
 
     command = boost::str(boost::format("addConstraint(%s)\n") % cg);
 
@@ -166,7 +166,7 @@ std::string PythonConverter::convert(const std::string& doc,
 
 std::string PythonConverter::convert(const std::string& doc,
                                      const std::vector<Sketcher::Constraint*>& constraints,
-                                     bool useLastGeoIdVar)
+                                     GeoIdMode geoIdMode)
 {
     if (constraints.size() == 1) {
         auto cg = convert(constraints[0]);
@@ -177,7 +177,7 @@ std::string PythonConverter::convert(const std::string& doc,
     std::string constraintlist = "constraintList = []";
 
     for (auto constraint : constraints) {
-        auto cg = process(constraint, useLastGeoIdVar);
+        auto cg = process(constraint, geoIdMode);
 
         constraintlist =
             boost::str(boost::format("%s\nconstraintList.append(%s)") % constraintlist % cg);
@@ -344,11 +344,12 @@ PythonConverter::SingleGeometry PythonConverter::process(const Part::Geometry* g
     return creator(geo);
 }
 
-std::string PythonConverter::process(const Sketcher::Constraint* constraint, bool useLastIdVar)
+std::string PythonConverter::process(const Sketcher::Constraint* constraint, GeoIdMode geoIdMode)
 {
-    std::string geoId1 = (useLastIdVar ? "lastGeoId + " : "") + std::to_string(constraint->First);
-    std::string geoId2 = (useLastIdVar ? "lastGeoId + " : "") + std::to_string(constraint->Second);
-    std::string geoId3 = (useLastIdVar ? "lastGeoId + " : "") + std::to_string(constraint->Third);
+    bool addLastIdVar = geoIdMode == GeoIdMode::AddLastGeoIdToGeoIds;
+    std::string geoId1 = (addLastIdVar ? "lastGeoId + " : "") + std::to_string(constraint->First);
+    std::string geoId2 = (addLastIdVar ? "lastGeoId + " : "") + std::to_string(constraint->Second);
+    std::string geoId3 = (addLastIdVar ? "lastGeoId + " : "") + std::to_string(constraint->Third);
 
     static std::map<
         const Sketcher::ConstraintType,
--- a/src/Mod/Sketcher/App/PythonConverter.cpp
+++ b/src/Mod/Sketcher/App/PythonConverter.cpp
@@ -59,11 +59,11 @@ std::string PythonConverter::convert(const Part::Geometry* geo, Mode mode)
     return command;
 }
 
-std::string PythonConverter::convert(const Sketcher::Constraint* constraint)
+std::string PythonConverter::convert(const Sketcher::Constraint* constraint, bool useLastGeoIdVar)
 {
     // addConstraint(Sketcher.Constraint('Distance',%d,%f))
     std::string command;
-    auto cg = process(constraint);
+    auto cg = process(constraint, useLastGeoIdVar);
 
     command = boost::str(boost::format("addConstraint(%s)\n") % cg);
 
@@ -85,13 +85,13 @@ std::string PythonConverter::convert(const std::string& doc,
         if (ngeos > 0) {
             if (construction) {
                 command = boost::str(
-                    boost::format("constrGeoList = []\n%s\n%s.addGeometry(constrGeoList,%s)\ndel "
-                                  "constrGeoList")
+                    boost::format("constrGeoList = []\n%s%s.addGeometry(constrGeoList,%s)\n"
+                                  "del constrGeoList\n")
                     % geolist % doc % "True");
             }
             else {
                 command = boost::str(
-                    boost::format("geoList = []\n%s\n%s.addGeometry(geoList,%s)\ndel geoList")
+                    boost::format("geoList = []\n%s%s.addGeometry(geoList,%s)\ndel geoList\n")
                     % geolist % doc % "False");
             }
         }
@@ -134,10 +134,10 @@ std::string PythonConverter::convert(const std::string& doc,
 
         if (sg.construction) {
             geolist =
-                boost::str(boost::format("%s\nconstrGeoList.append(%s)\n") % geolist % sg.creation);
+                boost::str(boost::format("%sconstrGeoList.append(%s)\n") % geolist % sg.creation);
         }
         else {
-            geolist = boost::str(boost::format("%s\ngeoList.append(%s)\n") % geolist % sg.creation);
+            geolist = boost::str(boost::format("%sgeoList.append(%s)\n") % geolist % sg.creation);
         }
 
         ngeos++;
@@ -165,7 +165,8 @@ std::string PythonConverter::convert(const std::string& doc,
 }
 
 std::string PythonConverter::convert(const std::string& doc,
-                                     const std::vector<Sketcher::Constraint*>& constraints)
+                                     const std::vector<Sketcher::Constraint*>& constraints,
+                                     bool useLastGeoIdVar)
 {
     if (constraints.size() == 1) {
         auto cg = convert(constraints[0]);
@@ -176,7 +177,7 @@ std::string PythonConverter::convert(const std::string& doc,
     std::string constraintlist = "constraintList = []";
 
     for (auto constraint : constraints) {
-        auto cg = process(constraint);
+        auto cg = process(constraint, useLastGeoIdVar);
 
         constraintlist =
             boost::str(boost::format("%s\nconstraintList.append(%s)") % constraintlist % cg);
@@ -200,7 +201,8 @@ PythonConverter::SingleGeometry PythonConverter::process(const Part::Geometry* g
                  auto sgeo = static_cast<const Part::GeomLineSegment*>(geo);
                  SingleGeometry sg;
                  sg.creation = boost::str(
-                     boost::format("Part.LineSegment(App.Vector(%f,%f,%f),App.Vector(%f,%f,%f))")
+                     boost::format(
+                         "Part.LineSegment(App.Vector(%f, %f, %f),App.Vector(%f, %f, %f))")
                      % sgeo->getStartPoint().x % sgeo->getStartPoint().y % sgeo->getStartPoint().z
                      % sgeo->getEndPoint().x % sgeo->getEndPoint().y % sgeo->getEndPoint().z);
                  sg.construction = Sketcher::GeometryFacade::getConstruction(geo);
@@ -225,7 +227,7 @@ PythonConverter::SingleGeometry PythonConverter::process(const Part::Geometry* g
                  auto sgeo = static_cast<const Part::GeomPoint*>(geo);
                  SingleGeometry sg;
                  sg.creation =
-                     boost::str(boost::format("Part.Point(App.Vector(%f,%f,%f))")
+                     boost::str(boost::format("Part.Point(App.Vector(%f, %f, %f))")
                                 % sgeo->getPoint().x % sgeo->getPoint().y % sgeo->getPoint().z);
                  sg.construction = Sketcher::GeometryFacade::getConstruction(geo);
                  return sg;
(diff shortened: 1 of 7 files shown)


PaddleStroke:
@abdullahtahiriyo there is one architectural question for the ViewProvider deleteSelected function that I use in the cut command.
This function is private. I was trying to make an attorney to access it but I am not sure how to handle it because CommandSketcherTools don't have a .h file to be included in the VP.
Then looking at this function I couldn't understand why this function would be private in the first place. Why centerSelection or updateColor is not private for example and deleteSelected is?

So I don't know if its acceptable to change it to public. If it is acceptable, then ViewProviderSketchShortcutListenerAttorney needs to be removed I guess?
If it's not acceptable, please advise where to put the attorney and what file to include in VPsketch.

Please advise what you want to be done here.

PaddleStroke:
Switched to draft. It appears it was working only when I had a custom toolbar with the new tools. If the custom toolbar is removed it doesn't work anymore...

xtemp09:
In _LibreCAD_, when user pastes from clipboard, it's automatically dragged with the mouse cursor. [^1]

Would it be better to mimic the same look-and-feel?


[^1]: https://youtu.be/8xeuNbu_8rc?t=23

PaddleStroke:
Yes I want to do this as well. But first we need to enable group dragging.

xtemp09:
Is 4000 a magic number? If it's not, then would it simpler instead of
```cpp
GeoId = std::atoi(it->substr(4, 4000).c_str()) - 1;
```
use
```cpp
GeoId = std::stoi( it->substr(4) ) - 1;
```
?

----

Would it be simpler to use _compare_ instead of _substr_? In this case,
```cpp
	if (it->size() > 4 && it->substr(0, 4) == "Edge")
```
becomes:
```cpp
	if ( it->compare(0, 4, "Edge") == 0 )
```

----

> formated ==> formatted

----


```cpp
	std::string data = QGuiApplication::clipboard()->text().toStdString();
	<...>
    if (data.find("</GeometryList>", 0) != std::string::npos) {
        geoString = data.substr(0, data.find("</GeometryList>", 0) + 16);
    }
```

`QGuiApplication::clipboard()->text()` returns `QString`. Would it be better to use `QString` methods and convert into `std::string` in the end? Strictly speaking, you are parsing text, converting to and from `QString`. Would it be simpler to just use Qt XML together with other Qt features, for example, `QTextStream` or `QString` methods?

Is there a reason to use XML? Clipboard content is intended to be used on the forum, would it be better to use JSON, since it is more human-readable?

----

`bool copySelectionToClipboard(void)` returns bool. However, there is no checking in `void CmdSketcherCopyClipboard::activated(int iMsg)`, while the function `void CmdSketcherCut::activated(int iMsg)` checks the return value.

`void CmdSketcherCopyClipboard::activated(int iMsg)` is supposed to copy to clipboard. However, if the function fails to do that, the clipboard content will stay the same. If end user copied something and then tries to copy something similar, the content of the clipboard remains unchanged. I think, it can bring confusion. Would it be better if `void CmdSketcherCopyClipboard::activated(int iMsg)` and `void CmdSketcherCut::activated(int iMsg)` show a warning in the status bar to inform the user that nothing has been copied or copied successfully?

`void CmdSketcherCut::activated(int iMsg)` deletes the selected items. Should it be logged?

PaddleStroke:
@xtemp09 for your 2 first comments, this was a copy paste of a code that is in many places. By the way it would probably be a good idea to make a function getting the list of geoId from the selection rather than having this block copy pasted everywhere.

XML is used because we are using the available `Part::PropertyGeometryList::save()`. So we are printing xml similar to what is used in the document xml. I don't really see the point of going for json since it would require writing a lot more code to parse geometries to json or to convert xml to json. And visibility is not so important, even in json a list of geometries and constraints is not going to make much sense to a reader.

> void CmdSketcherCopyClipboard::activated(int iMsg) is supposed to copy to clipboard. However, if the function fails to do that, the clipboard content will stay the same.

`CmdSketcherCopyClipboard::activated` does not check if something got copied because this is the usual ctrlC behavior. If you try to copy an empty text in notepad for example, your clipboard remains unchanged and nothing happens. So here we do the same, if nothing is selected nothing happens.

>  If end user copied something and then tries to copy something similar, the content of the clipboard remains unchanged.

Sorry I don't understand.

We can add a warning in the status bar, or better a user notification if copy failed. But I am not sure this is necessary.

xtemp09:
There is no reason for the function to return value if this value is not taken into consideration at all. If the value is somewhat important, should it be checked and some action performed?


----


About the last part. If a user copies similar fragments twice, but the last time the program failed to copy, the clipboard buffer will contain the first fragment. I just thought that the user should be notified about the failure to avoid copy-and-paste mistakes.

abdullahtahiriyo:
There are several ways to copy and paste between sketches or even inside one. One option is XML. Another option is using commands (via PythonConverter). 

The XML used in save function stores a geometry and all its extensions, so as to be able to retrieve it "as is". One caveat is that geometry extensions may have geometry specific information (specific to one sketch, to one specific use of geometry as people in ArchSketch WB are using by storing in extensions meta-information, or even just a Geometry ID). 

While using XML is at a first glance a brilliant idea (close to zero effort implementation), the question arises how to handle the issue of copying what is wanted, but not copying what is not wanted. Taking into account that part of it may come from stock FreeCAD by design (such as the Geometry ID or geometry index), but others are even inserted by third party tools. Deciding what to copy "as is" and what to "regenerate" (such as unique indices/uuids) or not copy at all is not straight-forward. We should also put an eye to impact for external geometry (ExternalGeometryExtension) and its use by code coming from the TopoNaming project (at least RT's implementation used it).

On the other hand we have PythonConverter, which copies console commands to reproduce the geometry, but will not handle any other special extension information. It is like adding the elements anew. The best advantage I see to this path is that the commands can be copied anywhere and even used to create macros. It is the "internal macro language" used by FreeCAD. The drawbacks are that some other information from extensions, which might perhaps be desirable to copy (visualisation attributes, or layer grouping information when it is available), is, at least in the current state of development of PythonConverter ignored.

So XML is copy all and then go and figure out what needs to be removed. While PythonConverter is copy the basic and add whatever else you may need.

This is one of the reasons (scarce review time being the other) why this PR (or better said its predecessor) has been queuing for some time.

So essentially I say two things: 
(a) we need assess very well the impact of this and select the right choice (XML or PythonConverter), 
(b) I will only be able to take a look to this next year, after merging all pending DSH staff and looking to Ajinkya's general tangency constraint code, which is a rather complex topic with potential very high impact if we do not get it right.

PaddleStroke:
Yes next year is fine no hurry.

abdullahtahiriyo:
Indeed. Pythonconverter was not available at the time, and I still think your idea of using XML code for serialisation is brilliant in many ways. Very creative approach.

Anyway, we look at it next year.

xtemp09:
I've just tried it. Seems to be working. When pasting _FreeCAD_ reports sometimes:

```
02:16:10  Sketcher constraint number 9 is malformed!
02:16:10  Unnamed#Sketch001: The Sketch has malformed constraints!
02:16:10  Sketcher constraint number 9 is malformed!
02:16:10  Sketcher constraint number 9 is malformed!
02:16:10  Unnamed#Sketch001: The Sketch has malformed constraints!
```





<details><summary>The clipboard: </summary>


```python
# Copied from sketcher.
lastGeoId = len(ActiveSketch.Geometry)

geoList = []

geoList.append(Part.LineSegment(App.Vector(-51.587891,39.518017,0.000000),App.Vector(-51.587891,10.801011,0.000000)))

geoList.append(Part.LineSegment(App.Vector(-51.587891,10.801011,0.000000),App.Vector(24.187309,10.801011,0.000000)))

geoList.append(Part.LineSegment(App.Vector(24.187309,10.801011,0.000000),App.Vector(24.187309,39.518017,0.000000)))

geoList.append(Part.LineSegment(App.Vector(24.187309,39.518017,0.000000),App.Vector(-51.587891,39.518017,0.000000)))

geoList.append(Part.Circle(App.Vector(0.000000, 0.000000, 0.000000), App.Vector(0.000000, 0.000000, 1.000000), 11.719850))

objectStr.addGeometry(geoList,False)
del geoList
constraintList = []
constraintList.append(Sketcher.Constraint('Coincident',  0, 2,  1, 1))
constraintList.append(Sketcher.Constraint('Coincident',  1, 2,  2, 1))
constraintList.append(Sketcher.Constraint('Coincident',  2, 2,  3, 1))
constraintList.append(Sketcher.Constraint('Coincident',  3, 2,  0, 1))
constraintList.append(Sketcher.Constraint('Vertical',  0))
constraintList.append(Sketcher.Constraint('Vertical',  2))
constraintList.append(Sketcher.Constraint('Horizontal',  1))
constraintList.append(Sketcher.Constraint('Horizontal',  3))
constraintList.append(Sketcher.Constraint('Coincident',  0, 3,  -1, 1))
objectStr.addConstraint(constraintList)
del constraintList
```



</details>



```ruby
OS: Ubuntu 22.04.3 LTS (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35569 (Git)
Build type: Debug
Branch: copy_paste_2023
Hash: cf3aeb6618cf68af754a74ac97f094fbb8188702
Python 3.10.12, Qt 5.15.12, Coin 4.0.0, Vtk 9.1.0, OCC 7.5.1
```


----


Is there a reason to use `boost::format` instead of `std::format`? `boost::format` is supposed to be faster, but in the strings like this: 

```ruby
return boost::str(boost::format("Sketcher.Constraint('Parallel', %s %i, %s %i)")
                                   % lastGeoId % constr->First % lastGeoId % constr->Second);
```


there is no gain in speed.

PaddleStroke:
Thanks!
There is a problem with your last coincident, the geoid2 is -1, I must have
made a mistake in the process function. With all the %s %i. Btw there's
probably a better syntax for this I will check tomorrow

PaddleStroke:
I have my changes to python converter it is cleaner now. Also I tested all constraint types (I think) and added the missing angleByPoint in pythonconverter.

Regarding the boost::format vs std::format I have no idea about it. As this is code that Abdullah wrote I don't want to change it because he may have had a reason and it is not related to this PR.

xtemp09:
The copy/paste seems to be working fine and looks pretty handy.


----


When compiling, compiler warns a lot about unused parameters. _Lint_ shows these warnings, like


```
unused parameter 'geoId3' [-Wunused-parameter]
```


Is there a reason to separate entries with empty line when filling `geoList` and `constrGeoList`? You don't do this when filling `constraintList`.

I think it would be neater if there was an empty line (or two) between the fragments of `constrGeoList`, `geoList` and `constraintList`, while empty lines between the entries was absent.

PaddleStroke:
Thanks for your comments.
I have : 
- corrected the lint warnings.
- adjusted the printed text. Now it looks like : 

# Copied from sketcher. From:
#objectStr = App.getDocument('offset_tests').getObject('Sketch002')
lastGeoId = len(ActiveSketch.Geometry)

> geoList = []
> geoList.append(Part.LineSegment(App.Vector(-1392.965698,733.026794,0.000000),App.Vector(-926.936890,1448.002441,0.000000)))
> geoList.append(Part.LineSegment(App.Vector(-926.936890,1448.002441,0.000000),App.Vector(-576.419617,733.026794,0.000000)))
> geoList.append(Part.LineSegment(App.Vector(-576.419617,733.026794,0.000000),App.Vector(-1392.965698,733.026794,0.000000)))
> geoList.append(Part.LineSegment(App.Vector(-2524.101074,1411.982788,0.000000),App.Vector(-1863.643188,2094.213379,0.000000)))
> geoList.append(Part.LineSegment(App.Vector(-1863.643188,2094.213379,0.000000),App.Vector(-1783.807739,1223.280518,0.000000)))
> geoList.append(Part.LineSegment(App.Vector(-1783.807739,1223.280518,0.000000),App.Vector(-2524.101074,1411.982788,0.000000)))
> geoList.append(Part.LineSegment(App.Vector(-909.245911,1905.511475,0.000000),App.Vector(-187.097076,1724.067017,0.000000)))
> geoList.append(Part.LineSegment(App.Vector(-187.097076,1724.067017,0.000000),App.Vector(-304.932681,2270.133664,0.000000)))
> geoList.append(Part.LineSegment(App.Vector(-445.697261,2316.250733,0.000000),App.Vector(-909.245911,1905.511475,0.000000)))
> geoList.append(Part.ArcOfCircle(Part.Circle(App.Vector(-388.798402, 2252.036298, 0.000000), App.Vector(0.000000, 0.000000, 1.000000), 85.796118), 0.212531, 2.295865))
> objectStr.addGeometry(geoList,False)
> del geoList
> 
> constraintList = []
> constraintList.append(Sketcher.Constraint('Coincident', lastGeoId + 0, 2, lastGeoId + 1, 1))
> constraintList.append(Sketcher.Constraint('Coincident', lastGeoId + 1, 2, lastGeoId + 2, 1))
> constraintList.append(Sketcher.Constraint('Coincident', lastGeoId + 2, 2, lastGeoId + 0, 1))
> constraintList.append(Sketcher.Constraint('Horizontal', lastGeoId + 2))
> constraintList.append(Sketcher.Constraint('Coincident', lastGeoId + 3, 2, lastGeoId + 4, 1))
> constraintList.append(Sketcher.Constraint('Coincident', lastGeoId + 4, 2, lastGeoId + 5, 1))
> constraintList.append(Sketcher.Constraint('Coincident', lastGeoId + 5, 2, lastGeoId + 3, 1))
> constraintList.append(Sketcher.Constraint('Coincident', lastGeoId + 6, 2, lastGeoId + 7, 1))
> constraintList.append(Sketcher.Constraint('Coincident', lastGeoId + 8, 2, lastGeoId + 6, 1))
> constraintList.append(Sketcher.Constraint('Tangent', lastGeoId + 8, 1, lastGeoId + 9, 2))
> constraintList.append(Sketcher.Constraint('Tangent', lastGeoId + 7, 2, lastGeoId + 9, 1))
> objectStr.addConstraint(constraintList)
> del constraintList

xtemp09:
Strictly speaking, I would beatify it even more, added spaces to some App.Vectors.

```
geoList.append(Part.LineSegment(App.Vector(-1392.965698,733.026794,0.000000),App.Vector(-926.936890,1448.002441,0.000000)))
```


is spaceless while

```
Part.Circle(App.Vector(-388.798402, 2252.036298, 0.000000)
```

contains spaces.

I would also add additional check, if a number contained zeroes, I would delete them, therefore:

```
geoList.append(Part.LineSegment(App.Vector(-1392.965698,733.026794,0.000000),App.Vector(-926.936890,1448.002441,0.000000)))
```

would become:

```
geoList.append(Part.LineSegment(App.Vector(-1392.965698, 733.026794, 0.0), App.Vector(-926.936890, 1448.002441, 0.0)))
```


and

```
App.Vector(0.000000, 0.000000, 1.000000)
```

would become:

```
App.Vector(0.0, 0.0, 1.0)
```

or, if possible,

```
App.Vector(0, 0, 1)
```

This is more readable for human eye.

<sub>But, it's just my opinion.</sub>

PaddleStroke:
@chennes this PR is ready if you're looking for something to review ;)

src/Mod/Sketcher/App/PythonConverter.cpp
```diff
@@ -59,11 +59,11 @@ std::string PythonConverter::convert(const Part::Geometry* geo, Mode mode)
     return command;
 }
 
-std::string PythonConverter::convert(const Sketcher::Constraint* constraint)
+std::string PythonConverter::convert(const Sketcher::Constraint* constraint, bool useLastGeoIdVar)
```
chennes, line 62: I have a strong preference that we not add boolean parameters like this -- as @wwmayer pointed out to me a couple years ago, it results in a phenomenon called "boolean blindness" that makes code very difficult to understand later on. The way it is usually recommended is to use an enumeration that contains two clearly spelled-out options that a later reader will understand. That said, I'm not in a position to suggest good names here as I'm just learning about this part of the code.

PaddleStroke: Ok I will find a proper name then.

PaddleStroke: It is done

maxwxyz:
@PaddleStroke should I add them in the Sketcher context menu if we have a selection? I have to adjust it to also work with external geometry as well.

Roy-043:
Unless I am mistaken (or have missed an intermediate PR) there is something wrong with how constraints are pasted. Their geoids are not updated.

1. Open the attached file (remove the .zip extension).
2. Open the sketch.
3. Box-select all elements and use Ctrl+C.
4. Move all endpoints to a new position.
5. Use Ctrl+V.
6. Result: the pasted symmetry constraint is applied to an existing element. 

[paste-geoids-test.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/14099148/paste-geoids-test.FCStd.zip)

```
OS: Windows 8 build 9600
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35720 (Git)
Build type: Release
Branch: main
Hash: 00cd50bbf49c810555899efe112ec7595c76cd3a
Locale: C/Default (C) [ OS: Dutch/Netherlands (nl_NL) ]
Installed mods:
```

PaddleStroke:
It should update the geoId, I may have broken things with my latest commit dealing with the boolean blindness.

Roy-043:
Hmm, you are right, it does for most constraints. But apparently the symmetry constrain, which I decided to use for my first tests, is not handled properly.

PaddleStroke:
I have just tested on my side with a single line with a length constraint. And it fails as you described. I will check in more details now

PaddleStroke:
I found the issue, it was not updating geoId when only one geometry was selected.

PR #11538, opened 2023-11-26
[Sketcher] hor/ver tool, remove unused variables

PaddleStroke:
hor/ver tool, remove unused variables
@wwmayer


wwmayer:
Should be fine.

PaddleStroke:
It was the same before in horizontal and vertical applyConstraints.
Please advise if the situation was wrong then and needs to be fixed, or if its ok.

wwmayer:
> Please advise if the situation was wrong then and needs to be fixed, or if its ok.

As said this PR is fine. The issue possibly has existed for several years and I don't want to figure out the exact commit when it has been introduced since this is quite irrelevant.

wwmayer:
#11549

PR #11544, opened 2023-11-26
downgrade cmakepresets version

Pesc0:


oursland:
What's the motivation for reducing the version?

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/11385#issuecomment-1826784032

oursland:
> @oursland does this mess with your CI work? see #11385

No.  Everything should be compatible with CMake 3.16.3, the current `cmake_minimum_required()` value.

PR #11545, opened 2023-11-26
Add .vscode folder to .gitignore

kadet1090:
Since we moved vscode files to `contrib/.vscode` in https://github.com/FreeCAD/FreeCAD/commit/d98f0e1ae386e1a3722d45593b4ac3c08b8ad797 we should add the folder to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ install_manifest.txt
 /src/Tools/offlinedoc/localwiki/
 /src/Tools/offlinedoc/*.txt
 /conda/environment.yml
+/.vscode/
 OpenSCAD_rc.py
 tags
 /CMakeUserPresets.json


adrianinsaval:
all dot files and directories are already ignored per
https://github.com/FreeCAD/FreeCAD/blob/185e58a2195bd583775b7044179c9d808752529c/.gitignore#L2

is it still being picked up by git for you?

kadet1090:
`.*` is not `.*/`.

`.github` is not whitelisted and yet it is in the directory, `.vscode` also was.

kadet1090:
Hm, I can confirm that this is how it works on empty repo. Strange, on my side `.vscode` got ignored only after explicitly adding it to the .gitignore - maybe it is due to it being force-added before?

adrianinsaval:
maybe but if you are on the latest on main it should have been ignored. It doesn't hurt to add it explicitly anyways.

PR #11548, opened 2023-11-27
casts clean

mosfet80:
clean unused code
remove unnecessary casts


yorikvanhavre:
The commented Python code looks like examples on how to use it... I think we shouldn't remove them, unless they are somewhere else in the documentation

PR #11551, opened 2023-11-27
Translations

yorikvanhavre:
crowdin translations


chennes:
No.

yorikvanhavre:
Hmm it might lead to a problem if the pre-commit modifies the source string. But AFAICS it's ok, only the translated ones are affected.

I don't think it's of great importance to feed it back to crowdin (not sure how we would do that anyway), we can consider this as an automatic "cleaning" of translations...

luzpaz:
Maybe we need a crowdin script function that does this? @kaktusus any idea ?

kaktusus:
> @kaktusus any idea ?

of the people I would like to cooperate with yet only you I do not have in the communicator ....
:stuck_out_tongue: 

I tried to find a case that would illustrate what you are talking about but failed.

chennes:
There's not really any need, IMO. Pre-commit just means that people in CrowdIn just don't have to worry about whether they have accidentally included trailing whitespace or not.

yorikvanhavre:
AFAICS it's only trailing whitespaces. We are always trying to remove these everywhere in the code. This is where we want to go anyway.

i propose we give this a test spin and look for problems first. 

In any case the translations on crowdin are safe, if needed, we oan always disable precommit on ts files and commit again, but it would be a pity to deprive ourselves of precommit.

kaktusus:
so that there are no problems associated with missing blanks at the end, all developers need to be made aware.
if they prepare their code in the right way no issues will occur. :wink:

chennes:
I'd bet that any case we find where the developer deliberately left a blank at the end, the string is actually bad for translation (e.g. the developer is concatenating a bunch of things together without regard for how that might work in non-English languages). So we could start checking the strings that are getting modified by the hook and see if they fall into that category.

kaktusus:
Very good idea
however, it is probably worth starting a separate discussion for this issue.

luzpaz:
Move discussion to the FreeCAD translation repo ?

kaktusus:
Yes.

PR #11556, opened 2023-11-27
VSCode: Change setup conda environment task type to process.

oursland:
The task type `shell` on Windows would invoke the user's default shell, which may not always correctly run the `conda/setup-environment.cmd` script.  Using the type `process` invokes the script directly, avoiding the issue.

### DIFF
--- a/contrib/.vscode/tasks.json
+++ b/contrib/.vscode/tasks.json
@@ -1,7 +1,7 @@
 {
   "tasks": [
     {
-      "type": "shell",
+      "type": "process",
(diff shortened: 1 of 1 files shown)


chennes:
@Pesc0 would you mind testing this?

oursland:
I have worked with @maxwxyz on Windows integration.  Perhaps he can provide feedback on this change.

maxwxyz:
The task is running correctly now, independent of the default terminal in VSCode (cmd or PS). Can build with this.

Issue #8706, opened 2023-03-01
PartDesign: Add ability to subtract the outside of the Sketch


leoheck:
### Is there an existing request for this?

- [X] I have searched the existing requests

### Forums discussion

This idea was presented here https://forum.freecad.org/viewtopic.php?p=664277#p664277

### Subproject(s) affected?

PartDesign

### Idea description

Instead of pocketing (or any other feature that removes material) with the Sketch, this setting would do that for the area outside o it.

Similarly, this could be extended to other features that add mass like Pad, except that it cannot work if it is applied to the first feature. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
How would it work for a pad? It would generate infinite shapes, this only makes sense for substractive features IMO.

FlachyJoe:
> How would it work for a pad? It would generate infinite shapes, this only makes sense for substractive features IMO.

The shape could be limited by existent face borders.
![image](https://user-images.githubusercontent.com/10371513/222123907-9195a877-83a5-41fb-8794-2b2a038a7fd5.png)

adrianinsaval:
I don't see how the software is supposed to identify such boundaries, and I've never seen such an option in other software, I consider that out of scope for now. Is there even an use case for this?

FlachyJoe:
> I don't see how the software is supposed to identify such boundaries,

By projection of the nearest surface to the sketch plane.


> I've never seen such an option in other software

 I've seen it in Creo Parametric

> Is there even an use case for this?

Can't say… :slightly_smiling_face:

adrianinsaval:
If someone legitimately wants that, please create a separate issue as it works fundamentally different and I want to focus on subtractive features here.

leoheck:
> > How would it work for a pad? It would generate infinite shapes, this only makes sense for substractive features IMO.
> 
> The shape could be limited by existent face borders. ![image](https://user-images.githubusercontent.com/10371513/222123907-9195a877-83a5-41fb-8794-2b2a038a7fd5.png)

Exactly like that.

adrianinsaval:
to focus the discussion here I'll quote myself from the forum:

> I think it's possible to tell occt to invert the face and do a pad from that without needing to calculate a contour around that, if not I'm thinking about taking the major diagonal of the bounding box of the base feature and making a big circle with that as radius.

and link to an old discussion/attempt at this: https://forum.freecad.org/viewtopic.php?t=61265

[...]

PR #11580, opened 2023-11-29
Support common operations for all subtractive parts, and refactor related code

### LINKED ISSUE #5991: PartDesign: Add Common (intersection) operations

bgbsww:
Replace #11389 per @FlachyJoe request.

Fix #5991

### DIFF
--- a/src/Mod/PartDesign/App/FeatureExtrude.cpp
+++ b/src/Mod/PartDesign/App/FeatureExtrude.cpp
@@ -24,6 +24,7 @@
 
 #include "PreCompiled.h"
 #ifndef _PreComp_
+# include <BRepAlgoAPI_Common.hxx>
 # include <BRepAlgoAPI_Fuse.hxx>
 # include <BRepAdaptor_Surface.hxx>
 # include <BRep_Builder.hxx>
--- a/src/Mod/PartDesign/App/FeatureExtrude.cpp
+++ b/src/Mod/PartDesign/App/FeatureExtrude.cpp
@@ -638,7 +638,7 @@ App::DocumentObjectExecReturn* FeatureExtrude::buildExtrusion(ExtrudeOptions opt
                         QT_TRANSLATE_NOOP("Exception", "Up to face: Could not get SubShape!"));
                 }
 
-                if (getAddSubType() == Additive) {
+                if ( isAdditive() ) {
                     prism = base.makeElementFuse(this->AddSubShape.getShape());
                 }
                 else {
@@ -718,12 +718,10 @@ App::DocumentObjectExecReturn* FeatureExtrude::buildExtrusion(ExtrudeOptions opt
             TopoShape result(0, getDocument()->getStringHasher());
             try {
                 const char* maker;
-                switch (getAddSubType()) {
-                    case Subtractive:
-                        maker = Part::OpCodes::Cut;
-                        break;
-                    default:
-                        maker = Part::OpCodes::Fuse;
+                if ( isSubtractive() ) {
+                    maker = Part::OpCodes::Cut;
+                } else {
+                    maker = Part::OpCodes::Fuse;
                 }
                 result.makeElementBoolean(maker, {base, prism});
             }
--- a/src/Mod/PartDesign/App/FeatureExtrude.cpp
+++ b/src/Mod/PartDesign/App/FeatureExtrude.cpp
@@ -25,15 +25,20 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <BRepAlgoAPI_Fuse.hxx>
+# include <BRepAlgoAPI_Common.hxx>
+# include <BRepAdaptor_Surface.hxx>
 # include <BRep_Builder.hxx>
+# include <BRepBuilderAPI_MakeFace.hxx>
 # include <BRepFeat_MakePrism.hxx>
 # include <BRepPrimAPI_MakePrism.hxx>
 # include <gp_Dir.hxx>
+# include <gp_Pln.hxx>
 # include <Precision.hxx>
 # include <TopExp_Explorer.hxx>
 # include <TopoDS_Compound.hxx>
 # include <TopoDS_Face.hxx>
 # include <TopoDS_Shape.hxx>
+# include <TopoDS.hxx>
 #endif
 
 #include <App/Document.h>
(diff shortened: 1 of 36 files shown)


src/Mod/PartDesign/App/FeatureExtrude.cpp
```diff
@@ -24,15 +24,20 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <BRepAlgoAPI_Fuse.hxx>
+# include <BRepAlgoAPI_Common.hxx>
```
FlachyJoe, line 27:
```suggestion
```

src/Mod/PartDesign/App/FeatureGroove.cpp
```diff
@@ -24,6 +24,7 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <BRepAlgoAPI_Cut.hxx>
+# include <BRepAlgoAPI_Common.hxx>
```
FlachyJoe, line 27:
```suggestion

```

src/Mod/PartDesign/App/FeatureHole.cpp
```diff
@@ -26,6 +26,7 @@
 # include <gp_Dir.hxx>
 # include <BRep_Builder.hxx>
 # include <BRepAlgoAPI_Cut.hxx>
+# include <BRepAlgoAPI_Common.hxx>
```
FlachyJoe, line 29:
```suggestion

```

src/Mod/PartDesign/App/FeatureLoft.cpp
```diff
@@ -27,6 +27,7 @@
 # include <TopExp_Explorer.hxx>
 # include <BRepAlgoAPI_Cut.hxx>
 # include <BRepAlgoAPI_Fuse.hxx>
+# include <BRepAlgoAPI_Common.hxx>
```
FlachyJoe, line 30:
```suggestion

```

src/Mod/PartDesign/App/FeaturePipe.cpp
```diff
@@ -24,6 +24,7 @@
 #include "PreCompiled.h"
 #ifndef _PreComp_
 # include <BRepAlgoAPI_Cut.hxx>
+# include <BRepAlgoAPI_Common.hxx>
```
FlachyJoe, line 27:
```suggestion

```

src/Mod/PartDesign/App/FeaturePrimitive.cpp
```diff
@@ -25,6 +25,7 @@
 #ifndef _PreComp_
 # include <BRepPrim_Cylinder.hxx>
 # include <BRepAlgoAPI_Cut.hxx>
+# include <BRepAlgoAPI_Common.hxx>
```
FlachyJoe, line 28:
```suggestion

```

FlachyJoe (changes requested):

bgbsww:
Weird.  I guess I just wait for github to shake this out?  I don't think I touched the subproject that is causing the compile error.

src/Mod/PartDesign/App/FeatureExtrude.cpp
FlachyJoe: Ooops ! Not here !

src/Mod/PartDesign/App/FeatureExtrude.cpp
```diff
@@ -120,6 +125,22 @@ Base::Vector3d FeatureExtrude::computeDirection(const Base::Vector3d& sketchVect
     return extrudeDirection;
 }
 
+void FeatureExtrude::extendFace(TopoDS_Face& face, const TopoDS_Shape bounds) {
+    if ( ! face.IsNull() ) {
+        BRepAdaptor_Surface adapt = BRepAdaptor_Surface(face);
+        // adapt.Initialize(face);
+        TopoDS_Face plane = BRepBuilderAPI_MakeFace(adapt.Plane());
+        if ( ! bounds.IsNull() ) {
+            BRepAlgoAPI_Common mkCom(bounds, plane);
```
FlachyJoe, line 133: header needed

FlachyJoe (changes requested):

src/Mod/PartDesign/Gui/TaskAddSubParameters.cpp
```diff
@@ -0,0 +1,77 @@
...
+bool TaskAddSubParameters::enableOutside(PartDesignGui::ViewProvider* vp)  {
+    PartDesign::FeatureAddSub* addsub = static_cast<PartDesign::FeatureAddSub*>(vp->getObject());
+    return addsub->isSubtractive();
+}
+
+// Compiler wants this implemented, whether we need it or not.
+void TaskAddSubParameters::onSelectionChanged(const Gui::SelectionChanges& msg)
+{
+    if (msg.Type == Gui::SelectionChanges::AddSelection) {
+    }
```
chennes, line 64: Why the empty if statement? If you are just not using msg, you can add `Q_UNUSED(msg)` to the method body to silence the warning and clarify your intent.

adrianinsaval:
I'm confused with the behaviour here, last time I tested pockets had exactly the desired behaviour and only pocketed up to the specified length but all other features seem to simply do a common operation, which is not ideal IMO. Anything stopping the rest of the features from behaving the same?

src/Mod/PartDesign/Gui/TaskAddSubParameters.cpp
bgbsww: Thanks, I think I left this in the heat of battle and never got back to learning how to fix it cleanly.

bgbsww:
I'm also confused.  Only the pockets have a length parameter for a subtractive operation.  This addition of the 'outside' option here does do a common instead of a cut, which seems to be what the linked discussions are looking for.

Are you expecting some other parameter to be added here, or so interpretation of a value that already exists?

There is certainly room for other work in this area, but they don't seem to be part of #5991.  I must be overlooking something.

adrianinsaval:
So the think is I lean more towards #8706 rather than #5991, they are somewhat conflicting. We could have different behavior on different features but if possible it would be better if they all behave the same. Given the behavior of pocket this is the behavior I would expect from a groove "outside":
image
[groove outside.zip](https://github.com/FreeCAD/FreeCAD/files/14011236/groove.outside.zip)
(imagine the square was not in the sketch)

adrianinsaval:
[patternBug.zip](https://github.com/FreeCAD/FreeCAD/files/14011315/patternBug.zip)
there's also this bug when trying to use an "outside" pocket in a pattern

maxwxyz:
@bgbsww @FlachyJoe any of you able to pick up the work on this PR prior to the feature freeze on Monday, 2024-06-03?

bgbsww:
I think so.  Let's see if I understand what the desired output is ...

FlachyJoe:
@chennes @adrianinsaval would you like to make this PR your in order bgbsww work was continued?

maxwxyz:
or @AIRCAP @kadet1090 maybe one of you want to pick it up?

maxwxyz:
@chennes will you also pick up this PR? Or @FlachyJoe are you willing to?

FlachyJoe:
Keep this open, still needed.

maxwxyz:
@chennes will you pick up this PR?

PR #11581, opened 2023-11-29
[Preferences] Make a User interface group

PaddleStroke:
Following this discussion : https://forum.freecad.org/viewtopic.php?t=83091

This initial PR does : 
Split UI page into Overlay and Tree View.
Introduce 'user interface' preference group like so :
image

Please comment if you think this group makes sense or not.


PaddleStroke:
Note this could already be merged as it is and further work in further PR.

PaddleStroke:
Not sure this is the best solution. It is probably best that the design review group discuss it further before actions. So closing this for now.

Issue #9230, opened 2023-04-10
[Problem] TechDraw: non-continuous lines do not follow ISO128


WandererFan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecad.org/viewtopic.php?t=77469

### Version

0.21 (Development)

### Full version info

```shell
any
```

### Subproject(s) affected?

None

### Issue description

ISO128 provides standards for the representation of non-continuous lines.  The standard specifies the lengths of marks and 
spaces as a function of the line width.

There are 6 graphical elements for line representation in ISO128:
Element/Line Width Multiplier
Dot: 1
Gap: 3
ShortDash: 6
Dash: 12
LongDash: 24
Space: 18

There are 15 non-continuous line representations:

- ISO01Continuous: n/a
- ISO02Dashed: Gap Dash
- ISO03DashedSpaced: Dash Space
- ISO04LongDashedDotted: LongDash Gap Dot Gap
- ISO05LongDashedDoubleDotted: LongDash Gap Dot Gap Dot Gap
- ISO06LongDashedTripleDotted: LongDash Gap Dot Gap Dot Gap Dot Gap
- ISO07Dotted: Dot Gap
- ISO08LongDashShortDash: LongDash Gap ShortDash Gap
- ISO09LongDashDoubleShortDash: LongDash Gap ShortDash Gap
- ISO10DashedDotted: Dash Gap Dot Gap
- ISO11DoubleDashedDotted: Gap Dash Gap Dot Gap Dash
- ISO12DashedDoubleDotted: Dash Gap Dot Gap Dot Gap
- ISO13DoubleDashedDoubleDotted: Dash Gap Dash Gap Dot Gap Dot Gap
- ISO14DashedTripleDotted: Dash Gap Dot Gap Dot Gap Dot Gap
- ISO15DoubleDashedTripleDotted: Dash Gap Dash Gap Dot Gap Dot Gap Dot Gap

Currently, TechDraw provides 5 options for line styles:
- continuous (ISO 01)
- dashed (similar to ISO 02)
- dotted (similar to ISO 07)
- dashDot (similar to ISO 10)
- dashDotDot (similar to ISO 12)

The mark and space lengths are provided by Qt according to Qt::PenStyle.

[...]

PR #11594, opened 2023-11-30
[TD]Non-continuous Lines to Standards

WandererFan:
This PR implements a solution to issue #9230.

Additional forum discussion: https://forum.freecad.org/viewtopic.php?t=82287&sid=88b288b1c46dd306928bbd46dc2eff4f
and
https://forum.freecad.org/viewtopic.php?t=81918

### DIFF
--- a/src/Mod/TechDraw/App/CMakeLists.txt
+++ b/src/Mod/TechDraw/App/CMakeLists.txt
@@ -10,7 +10,6 @@ include_directories(
     ${ZLIB_INCLUDE_DIR}
     ${PYTHON_INCLUDE_DIRS}
     ${XercesC_INCLUDE_DIRS}
-    ${QT_QTCORE_INCLUDE_DIR}
 )
 
 set(TechDrawLIBS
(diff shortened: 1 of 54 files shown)


maxwxyz:
@WandererFan not sure if this is related to this PR but I've noticed it on yesterday's weekly build on another computer: The drop down menu in the Preferences>Annotation>Lines for `Line standard` is empty and cannot be opened:
grafik
When I reset the preferences of the annotation tab the drop down menu with the new line styles is visible. So maybe old users with an existing config file cannot set the new styles?

After reset of annotation tab:
grafik

WandererFan:
I can't duplicate this with FreeCAD_weekly-builds-35554-2024-01-08-conda-Linux-x86_64-py310.AppImage.

The code looks for available standards in one of two places:
Tools > Edit parameters
TechDraw / Files / LineDefLocation 

or

App.getResourceDir()/Mod/TechDraw/LineGroup/

If there is a LineDefLocation preference value that points to a directory with no line definition files,  that would give the empty drop down box.  It might be a value that AppImage can not access?

If you can make this happen again, please note the value of that preference and the result of 
"App.getResourceDir()".   That might give us a clue.

maxwxyz:
I cannot reproduce as well, seemed that I had an older weekly build on the system and when I downloaded the current weekly this setting was empty. I tried to set it with 0.21 to ISO and apply the settings but the new weekly is working now.

maxwxyz:
fresh user.cfg solved the issues!

PR #11597, opened 2023-12-01
conda: Remove conda-build and conda-smithy. Add git.

oursland:
On Windows, installing either `conda-build` or `conda-smithy` results in a broken environment.  Some of the `.exe` binaries will fail either silently or with the following message:

image

I believe this is because a dependency of the `conda-build` or `conda-smithy` has a broken install script that results in broken library linkages.

The addition of `git`, which was included by `conda-build` and `conda-smithy`, is required for some tests in `AddonManager`.

### DIFF
--- a/conda/environment.devenv.yml
+++ b/conda/environment.devenv.yml
@@ -67,6 +67,7 @@ dependencies:
 - eigen
 - fmt
 - freetype
+- git
 - gmsh
 - graphviz
 - hdf5
--- a/conda/environment.devenv.yml
+++ b/conda/environment.devenv.yml
@@ -61,9 +61,7 @@ dependencies:
 - coin3d==4.0.0
 - compilers
 - conda
-- conda-build
 - conda-devenv
-- conda-smithy
 - debugpy
 - doxygen
 - eigen


wwmayer:
> I believe this is because a dependency of the conda-build or conda-smithy has a broken install script that results in broken library linkages.

Is this something you know for sure or just an educated guess?

oursland:
> Is this something you know for sure or just an educated guess?

This is an educated guess based on prior experience.  I've asked the `conda` teams for assistance in understanding and debugging the underlying cause.

oursland:
For further context, I have a PR for the CI changes to support the CMake Presets, conda builds, and macOS.  The Windows conda builds fail due to non-obvious reasons caused by this issue.

The GitHub CI logs shows as a failure in [`git` here](https://github.com/oursland/FreeCAD/actions/runs/7053081250/job/19199442872#step:14:16004), but the actual cause is that `git.exe` is calling `sh.exe` that is linked against the DLL noted in the screenshot.  There is no output on the terminal, so debugging this requires manual effort on a Windows machine.

wwmayer:
I guess this isn't something that can be easily reproduced locally, i.e. setting up a local Conda build environment?

PR #11598, opened 2023-12-01
Correct mistaken Helix patch; use MakePipe for frenet for speed; add tests

### LINKED ISSUE #9377: [Problem] PartDesign: Helix insufficient precision

bgbsww:
This solves issues with random failures at different numbers of turns.  Seems much better all the way around.

Properly handles non concentric and multiple internal holes.

Address discussion points in #11312

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -98,3 +98,5 @@ db24eeec535f1f43fb3d5b63d24c5171af637880 # RE: Final application of pre-commit
 b8f8b232cb0882d171cb299e6f6279a516cdd6eb # Inspection: Final application of pre-commit
 c5c2ea3498f402c0c89916c46ddb071e22756622 # Assembly: Final application of pre-commit
 592c992b863549fde52741fd8830418168387695 # Assembly: Apply pre-commit to Assembly files
+94ca51b7799b2ac60a2a56d7e6e753bb97a73671 # PartDesign: Fix #9377 - issue with accuracy in AdditiveHelix (#11312)
+d472927bba7b2d8d151c99fb29cf1d8dd099ea7d # Correct PartDesign Helix feature negative angles (#11399)
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -32,6 +32,7 @@
 # include <BRepBuilderAPI_Sewing.hxx>
 # include <BRepClass3d_SolidClassifier.hxx>
 # include <BRepOffsetAPI_MakePipe.hxx>
+# include <BRepOffsetAPI_MakePipeShell.hxx>
 # include <BRepPrimAPI_MakeRevol.hxx>
 # include <Precision.hxx>
 # include <TopoDS.hxx>
@@ -51,7 +52,7 @@
 # include <Mod/Part/App/FaceMakerCheese.h>
 
 # include "FeatureHelix.h"
-
+#include "App/Document.h"
 using namespace PartDesign;
 
 const char* Helix::ModeEnums[] = { "pitch-height-angle", "pitch-turns-angle", "height-turns-angle", "height-turns-growth", nullptr };
@@ -103,8 +104,11 @@ Helix::Helix()
     ADD_PROPERTY_TYPE(HasBeenEdited, (false), group, App::Prop_Hidden,
         QT_TRANSLATE_NOOP("App::Property", "If false, the tool will propose an initial value for the pitch based on the profile bounding box,\n"
             "so that self intersection is avoided."));
+    ADD_PROPERTY_TYPE(UseMakePipe, (false), group, App::Prop_None, //App::Prop_Hidden,
+        QT_TRANSLATE_NOOP("App::Property", "If set, use the faster MakePipe instead of the legacy MakePipeShape."));
 
     setReadWriteStatusForMode(initialMode);
+    UseMakePipe.setValue(true); // For all new obects
 }
 
 short Helix::mustExecute() const
@@ -214,8 +218,6 @@ App::DocumentObjectExecReturn* Helix::execute()
 
         base.Move(invObjLoc);
 
-        // generate the helix path
-        TopoDS_Shape path = generateHelixPath(0.0, 1.0);
 
         std::vector<TopoDS_Wire> wires;
         try {
@@ -224,14 +226,92 @@ App::DocumentObjectExecReturn* Helix::execute()
         catch (const Base::Exception& e) {
             return new App::DocumentObjectExecReturn(e.what());
         }
+        TopoDS_Shape result;
+        if ( UseMakePipe.getValue() )  {
+            TopoDS_Shape path = generateHelixPath(0.0, 1.0);
+            TopoDS_Shape face = Part::FaceMakerCheese::makeFace(wires);
+            BRepOffsetAPI_MakePipe mkPS(TopoDS::Wire(path), face, GeomFill_Trihedron::GeomFill_IsFrenet, Standard_False);
 
-        TopoDS_Shape face = Part::FaceMakerCheese::makeFace(wires);
-        BRepOffsetAPI_MakePipe mkPS(TopoDS::Wire(path), face, GeomFill_Trihedron::GeomFill_IsFrenet, Standard_False);
+            mkPS.Build();
 
-        mkPS.Build();
+            result = mkPS.Shape();
+        } else {
+            /*  @deprecrated  12/01/2023  This is legacy code, here to maintain backward compatibility with existing models without altering their
+             *  Helixes to have fewer seams, and potentially trigger TNP.  Someday it should go away, and so should the gating
+             *  property.
+             */
 
-        TopoDS_Shape result = mkPS.Shape();
+            // generate the helix path
+            TopoDS_Shape path = generateHelixPath();
+            TopoDS_Shape auxpath = generateHelixPath(1.0);
+            std::vector<std::vector<TopoDS_Wire>> wiresections;
+            for (TopoDS_Wire& wire : wires)
+                wiresections.emplace_back(1, wire);
 
+            //build all shells
+            std::vector<TopoDS_Shape> shells;
+            std::vector<TopoDS_Wire> frontwires, backwires;
+            for (std::vector<TopoDS_Wire>& wires : wiresections) {
+
+                BRepOffsetAPI_MakePipeShell mkPS(TopoDS::Wire(path));
+
+                mkPS.SetTolerance(Precision::Confusion());
+                mkPS.SetTransitionMode(BRepBuilderAPI_Transformed);
+
+                if (Angle.getValue() == 0) {
+                    mkPS.SetMode(true);  //This is for frenet
+                } else {
+                    mkPS.SetMode(TopoDS::Wire(auxpath), true);  // this is for auxiliary
+                }
+
+                for (TopoDS_Wire& wire : wires) {
+                    wire.Move(invObjLoc);
+                    mkPS.Add(wire);
+                }
+
+                if (!mkPS.IsReady())
+                    return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Error: Could not build"));
+                mkPS.Build();
+
+                shells.push_back(mkPS.Shape());
+
+                if (!mkPS.Shape().Closed()) {
+                    // // shell is not closed - use simulate to get the end wires
+                    TopTools_ListOfShape sim;
+                    mkPS.Simulate(2, sim);
+
+                    frontwires.push_back(TopoDS::Wire(sim.First()));
+                    backwires.push_back(TopoDS::Wire(sim.Last()));
+                }
+            }
+            BRepBuilderAPI_MakeSolid mkSolid;
+
+            if (!frontwires.empty()) {
+                // build the end faces, sew the shell and build the final solid
+                TopoDS_Shape front = Part::FaceMakerCheese::makeFace(frontwires);
+                TopoDS_Shape back = Part::FaceMakerCheese::makeFace(backwires);
+
+                BRepBuilderAPI_Sewing sewer;
+                sewer.SetTolerance(Precision::Confusion());
+                sewer.Add(front);
+                sewer.Add(back);
+
+                for (TopoDS_Shape& s : shells)
+                    sewer.Add(s);
+                sewer.Perform();
+                mkSolid.Add(TopoDS::Shell(sewer.SewedShape()));
+            }
+            else {
+                // shells are already closed - add them directly
+                for (TopoDS_Shape& s : shells) {
+                    mkSolid.Add(TopoDS::Shell(s));
+                }
+            }
+            if (!mkSolid.IsDone())
+                return new App::DocumentObjectExecReturn(QT_TRANSLATE_NOOP("Exception", "Error: Result is not a solid"));
+
+            result = mkSolid.Shape();
+        }
         BRepClass3d_SolidClassifier SC(result);
         SC.PerformInfinitePoint(Precision::Confusion());
         if (SC.State() == TopAbs_IN)
--- a/src/Mod/PartDesign/App/FeatureHelix.cpp
+++ b/src/Mod/PartDesign/App/FeatureHelix.cpp
@@ -226,7 +226,7 @@ App::DocumentObjectExecReturn* Helix::execute()
         }
 
         TopoDS_Shape face = Part::FaceMakerCheese::makeFace(wires);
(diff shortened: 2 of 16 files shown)


bgbsww:
Seems like there might be an opportunity to clean up FeaturePipe and/or share code between these two, but that's beyond the scope of fixing this problem.

FlachyJoe (approved):

wwmayer:
Here is a test for PD helix: https://github.com/FreeCAD/FreeCAD/pull/11601.

If you need it to copy & paste it into the Python console you can use:
```
from FreeCAD import Base
import Sketcher
import Part

doc = FreeCAD.newDocument()
body = doc.addObject('PartDesign::Body','Body')
sketch = body.newObject('Sketcher::SketchObject','Sketch')
sketch.Support = (doc.getObject('XY_Plane'),[''])
sketch.MapMode = 'FlatFace'

geoList = []
geoList.append(Part.Circle(Base.Vector(-40.0, 0.000000, 0.000000), Base.Vector(0.000000, 0.000000, 1.000000), 7.5))
geoList.append(Part.Circle(Base.Vector(-40.0, 0.000000, 0.000000), Base.Vector(0.000000, 0.000000, 1.000000), 10.0))
sketch.addGeometry(geoList, False)
del geoList

sketch.addConstraint(Sketcher.Constraint('PointOnObject', 0, 3, -1))
sketch.addConstraint(Sketcher.Constraint('Coincident', 1, 3, 0, 3))
doc.recompute()

helix = body.newObject('PartDesign::AdditiveHelix','AdditiveHelix')
helix.Profile = sketch
helix.ReferenceAxis = (sketch, ['V_Axis'])
helix.Mode = 0
helix.Pitch = 35
helix.Height = 100
helix.Turns = 3
helix.Angle = 0
helix.Growth = 0
helix.LeftHanded = 0
helix.Reversed = 0

doc.recompute()
```

bgbsww:
I no longer see a good opportunity to share with FeaturePipe:  too many MakePipeShell specific options are exposed in the UI, so I would either have to switch Helix back to using MakePipeShell with the associated issues, or take away options in the FeaturePipe UI.  Oh well.

I'm assuming that after this gets committed then #11601 will pass and be be committed.  I appreciate the lesson on checking for and then writing tests.

adrianinsaval:
I'm not sure what's exactly the issue but as it is this PR is not acceptable. Helix are recomputed completely different than before so it will break a lot of files.
This is how these files look in main:
image
this is how they look with this PR:
image
these are files that I found of previous PD helix issues
[pd_helix_models.zip](https://github.com/FreeCAD/FreeCAD/files/13529506/pd_helix_models.zip)

in general it seems to be pretty buggy:
image
image
image

adrianinsaval:
these are the results now:
image
seems correct on two of the files but fails miserably on ForumHelix.FCStd

The other two look ok but if you check the volume of the shape pre and post recompute there are changes in the 10⁻⁴ range, hard to know how much of an impact this has but I worry it might break files unexpectedly.

Personally I think it's a bad idea to change the occt method used on an established feature, it's hard to know what may break, the least of all problems is that it may trigger TNP simply on recompute.

Testing with a concentric circle I also find that the result is not the same on the PR (left) compared to 0.21.2 (right):
image

notice how it looks the missing seam edges and different relative position of the loop that passes through the YZ plane
[concentric.zip](https://github.com/FreeCAD/FreeCAD/files/13532671/concentric.zip)

> Thanks, @FlachyJoe you beat me to it. That solves the three example files for me. I'm ~trying~ tried to confirm the other bug cases @adrianinsaval noted, but was unable to duplicate anything that seemed wrong. I'm still open for more examples of problems.

what version of occt are you using? are you recomputing the helix features on each file? Here's what I'm testing on:
```
OS: Arch Linux (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35260 +5 (Git)
Build type: None
Branch: bgbsww-patch-16
Hash: b5862a781b480b85aebd57cda496375c1a09c08c
Locale: C/Default (C) [ OS: English/United Kingdom (en_GB) ]
Installed mods: 
```

bgbsww:
```
OS: Ubuntu 22.04.3 LTS (ubuntu:GNOME/ubuntu)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35265 (Git)
Build type: Debug
Branch: bgbsww-patch-16
Hash: b5862a781b480b85aebd57cda496375c1a09c08c
Python 3.10.12, Qt 5.15.3, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.2
Installed mods: 
  * FreeCAD_BearingsWB 0.0.2
  * Assembly4 0.50.4
  * fcgear 1.0.0
  * Plot 2022.4.17
  * DynamicData 2.53.0
  * BIM 2021.12.0
  * parts_library
```

You raise a great point on changing the OCCT method.  I spent many hours trying to get the former method to work, and it was always unreliable depending on how many twists the helix had.  But it's worth noting that the realthunder branch works flawlessly, and does use MakePipeShell.

The argument in favor of changing is that there are many fewer calls to OCCT since there's no sewing involved, and that appears to make it faster.

My test of concentric.FCStd against my build shows exactly the same as the right image (0.21.2).  ForumHelix renders correctly.  I need to figure out how to do those screenshots you are showing.

I am recomputing the entire file.  I guess I need to see if I can recreate the effects you are seeing.

Thank you for the efforts you are putting in here.

FlachyJoe:
@bgbsww 
in the two cases it's a profile placement issue I can reproduce.
image
image
```
OS: Debian GNU/Linux trixie/sid (XFCE/lightdm-xsession)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35246 +2 (Git)
Build type: Release
Branch: helix
Hash: 379795f84854388210928a141c3f23653c73ff49
Python 3.11.6, Qt 5.15.10, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: French/France (fr_FR)
```

@adrianinsaval missing seam edges is not a pain :roll_eyes:

adrianinsaval:
It is a pain because it will likely trigger TNP on existing models just by recomputing

adrianinsaval:
It is however a better result for newly created models. So one potential way forward is to keep both methods, controlled by a property. New models will use this, old will use the old one. We mark the old method as deprecated and we could remove it on a future version after the topo naming algorithm has been around for a reasonable time

bgbsww:
Hrm.  Maybe recomputing wrong?  I can duplicate the effect now, so something to chase.

FlachyJoe:
I think it's better to split this PR in two: 

- revert #11312 and fix the accuracy. To be merged quickly.
- switch to MakePipe but hang it until TopoNaming is fixed. i.e. give it the "for 1.0" label.

bgbsww:
That seems like a good path forward to me.  Note that when I go back to the prior code, on recompute I get the correct relative position, but I do not get the seam edges.  Could be a mistake I'm making, or yet another subtlety hidden here.

In any case, after reverting, the makePipe version can be moved forward with either the Property or the "for 1.0" approach.

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -224,14 +226,92 @@ App::DocumentObjectExecReturn* Helix::execute()
...
+                BRepOffsetAPI_MakePipeShell mkPS(TopoDS::Wire(path));
+
+                mkPS.SetTolerance(Precision::Confusion());
+                mkPS.SetTransitionMode(BRepBuilderAPI_Transformed);
+
+                if (Angle.getValue() == 0) {
+                    mkPS.SetMode(true);  //This is for frenet
+                } else {
+                    mkPS.SetMode(TopoDS::Wire(auxpath), true);  // this is for auxiliary
+                }
```
adrianinsaval, line 264: can we add the fallback to auxpath here or is that not possible with the old method?

adrianinsaval: makes sense, the file I was using as example to check if this bug still was present fails so I thought about this, but you're right

adrianinsaval:
> The difference is this other accepted patch of mine: #11399
> So if there really is a potential TNP problem here, this should get removed as well, so I've added it to this patch.

ok so maybe what we actually need is to make this value controlled by property, default to 1 for existing objects and to 0 for new objects. If MakePipe can produce the same topology as MakePipeShell we might not even need to keep the MakePipeShell code or at least it can make it safer to mode to it after 1.0.

btw I still see the problem with the incorrect placement of the profiles when using the new method

> I've added the property to this patch to complete the work, not decide the approach. We can revert https://github.com/FreeCAD/FreeCAD/pull/11312 and https://github.com/FreeCAD/FreeCAD/pull/11399 and then I'll submit the MakePipe fix for 1.0, or we can revert https://github.com/FreeCAD/FreeCAD/pull/11312 to preserve the blame history, and then I'll apply all these changes on top of it including the property to fix the bugs now, or we can apply this patch to fix everything now, but bugger up the history.

reverting to preserve the blame info seems like a good idea, another possibility is to add the intermediary commits to https://github.com/FreeCAD/FreeCAD/blob/main/.git-blame-ignore-revs

adrianinsaval:
I'll see tomorrow if I can come up with a file to use to check for toponaming resilience

bgbsww:
The UseMakePipe for this patch also solves what #11399 solved, because that was all about trying to work around bugs in MakePipeShell; once we switch to MakePipe for new objects, we're all set, no need for another property.

I don't know that we need a file to check for TNP resilience, since we will no longer be modifying existing objects in any way, even when recomputed.  But I'm not going to complain about extra checks here.

The positioning issue is troubling.  I'm still looking at that one.  I need to analyze the helixes from generateHelixPath closely.

bgbsww:
OK!  Now I think it's really and truly done.  My oversight that led to the positioning issue should be fixed.

src/Mod/PartDesign/App/FeatureHelix.cpp
```diff
@@ -229,13 +229,14 @@ App::DocumentObjectExecReturn* Helix::execute()
...
             TopoDS_Shape face = Part::FaceMakerCheese::makeFace(wires);
+            face.Move(invObjLoc);
             BRepOffsetAPI_MakePipe mkPS(TopoDS::Wire(path), face, GeomFill_Trihedron::GeomFill_IsFrenet, Standard_False);
 
             mkPS.Build();
 
             result = mkPS.Shape();
         } else {
-            /*  @deprecrated  12/01/2023  This is legacy code, here to maintain backward compatibility with existing models without altering their
+            /*  Deprecated  12/01/2023  This is legacy code, here to maintain backward compatibility with existing models without altering their
```
luzpaz: Why is the doxygen formatting being altered here ?

bgbsww: Uh, that was me correcting my typo of a line I put there, I thought.   No doxygen involved.  Note the extra 'r' as well as the @ were both me, earlier.

luzpaz: Oh. I see. Thanks for the clarification. In that case perhaps keep the @deprecated Eventually we'll get our act together and doxy up the rest of the code. Might as well start.

adrianinsaval:
still not the expected result
https://github.com/FreeCAD/FreeCAD/assets/36372335/411de117-d59d-4d5f-8534-ef269c237951

bgbsww:
I think I've exhaustively shown that MakePipe cannot work correctly in cases with a Helix angle - an auxilliary spine is required, and is only in MakePipeShell.

I also now have a good array of unit tests for this.  I need to formulate a new patch (rework this one) that still solves the original #9377 issue but also cleans up a bunch of stuff that still appears redundant and passes the test suite.  Stay tuned.

bgbsww:
I think the right way forward here is to start again from the beginning by reverting #11312 and reopening #9377.

I can then add the tests in, and work towards addressing the original issue of the lastShape surface of the additive helix being a tiny bit too small.

Syres916:
@bgbsww I've compiled this PR and it doesn't appear to correct the issue of opening 0.21.x files, if you Mark to Recompute the whole document of the attached file and then Recompute it results in a faulty model. See forum discussion https://forum.freecad.org/viewtopic.php?t=83794&sid=9a64c7b868dd423028dc4dc906230cde

```
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35276 (Git)
Build type: Release
Branch: bgbsww-patch-16
Hash: 7b014e856b5ff4dee529a659966fdd874e54782e
Python 3.8.10, Qt 5.12.8, Coin 4.0.0, Vtk 7.1.1, OCC 7.6.3
Installed mods: 
  * Silk 0.1.5
  * Defeaturing 1.2.1
  * CurvedShapes 1.0.5
  * freecad.gears 1.2.0
  * Curves 0.6.21
  * woodworking 0.21.1.33694
  * Plot 2022.4.17
  * CfdOF 1.24.9
  * fasteners 0.4.75
  * sheetmetal 0.3.13
```
[Gew_V21.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/13772323/Gew_V21.FCStd.zip)

bgbsww:
Thanks, @Syres916 .  The #9377 issue proposed patch correction is at #11776.  I'll run that test file in a few hours to see if it is addressed.

Syres916:
Yep, that looks better, thanks.

.git-blame-ignore-revs
```diff
@@ -98,3 +98,5 @@ db24eeec535f1f43fb3d5b63d24c5171af637880 # RE: Final application of pre-commit
 b8f8b232cb0882d171cb299e6f6279a516cdd6eb # Inspection: Final application of pre-commit
 c5c2ea3498f402c0c89916c46ddb071e22756622 # Assembly: Final application of pre-commit
 592c992b863549fde52741fd8830418168387695 # Assembly: Apply pre-commit to Assembly files
+94ca51b7799b2ac60a2a56d7e6e753bb97a73671 # PartDesign: Fix #9377 - issue with accuracy in AdditiveHelix (#11312)
+d472927bba7b2d8d151c99fb29cf1d8dd099ea7d # Correct PartDesign Helix feature negative angles (#11399)
```
chennes, line 102: Why are we ignoring these two revs?

src/Mod/PartDesign/PartDesignTests/TestHelix.py
```diff
@@ -0,0 +1,143 @@
...
+#*   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+#*   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+#*   GNU Library General Public License for more details.                  *
+#*                                                                         *
+#*   You should have received a copy of the GNU Library General Public     *
+#*   License along with this program; if not, write to the Free Software   *
+#*   Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  *
+#*   USA                                                                   *
+#*                                                                         *
+#***************************************************************************
```
chennes, line 20: It's not important for you to fix it here, but for future reference this is the old version of the license header. You can find the new one here: https://freecad.github.io/DevelopersHandbook/codeformatting/#c-copyright-header

.git-blame-ignore-revs
bgbsww: To restore the original authorship of the lines in question; this was suggested as an alternative to a revert.

Essentially I removed code, we later discovered it was important, and this keeps the original blame info.

src/Mod/PartDesign/PartDesignTests/TestHelix.py
bgbsww: Thanks, I'll get that right on all future submissions.  I won't change it here unless something else drives further changes to this PR.

Bernard-19:
I am just a simple FreeCAD user and I praise the commitment you bring to solving all these annoyances. Well done again and thank you.
I will wait for the next updates.
Capture d'écran 2024-01-10 202026

Bernard-19:
My problem persists to develop a spring with a wire of 0.5 diameter distance of 6.95 in an XY sketch.

OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.36117 (Git)
Build type: Release
Branch: main
Hash: 8a3407aee6dff8218282d9cdc3f4e32e7ad958d4
Locale: French/France (fr_FR)
Installed mods: 
  * Assembly4 0.50.6

PR #11599, opened 2023-12-01
CI: Add CMake Presets, Ubuntu-conda, Windows-conda, and macOS-conda.

oursland:
This PR adds support for:
* exercising the CMake Presets in CI
* conda-based package dependency builds for Ubuntu 22.04 and Windows
* macOS builds on conda
* clang builds on Ubuntu 20.04 using Clang 12 and OpenMP 5.0
* Windows C++ unit tests on conda


adrianinsaval:
I don't get what's the benefit in testing with ubuntu 20.04+conda and 22.04+conda, conda should not be distro dependent so testing on just one should be sufficient.

instead it could be a good idea to test just on ubuntu 20.04+clang with system deps (this is our baseline for minimum supported library versions) and with ubuntu-latest+conda+gcc so we'll remain with just 2 linux tests. While it's true that github actions are free IMO this doesn't mean we should abuse the system and run superfluous CI just because we can.

cMake/FreeCAD_Helpers/SetGlobalCompilerAndLinkerSettings.cmake
```diff
@@ -42,6 +42,13 @@ macro(SetGlobalCompilerAndLinkerSettings)
                 set (CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /EHa")
             endif()
         endif(FREECAD_RELEASE_SEH)
+        if(FREECAD_USE_CCACHE)
```
adrianinsaval, line 45: maybe this should be `if(CCACHE_PROGRAM)` instead? `FREECAD_USE_CCACHE` is true by default but it might very well be the case that ccache is not installed in which case these options are unnecessary

.github/workflows/sub_buildWindowsConda.yml
```diff
@@ -0,0 +1,140 @@
...
+          conda\cmake.cmd --preset conda-windows-release -DFREECAD_USE_PCH:BOOL=OFF -DFREECAD_RELEASE_PDB:BOOL=OFF -DFREECAD_USE_CCACHE:BOOL=ON
+      - name: Compiling sources
+        shell: cmd /C CALL {0}
+        run: >
+          conda\cmake.cmd --build build\release
+      - name: Print Ccache statistics after build
+        run: |
+          . $env:ccachebindir\ccache -s
+      - name: C++ unit tests
+        if: false # Disabled because seems to not exist on Windows build
```
adrianinsaval, line 127: is this still correct?

oursland:
I can remove the "Ubuntu 22.04" and "Ubuntu 20.04 (Conda)" to bring it to the current baseline "Ubuntu 20.04" and the conda environment for Linux "Ubuntu 22.04 (Conda)".

Is that what you're requesting?

cMake/FreeCAD_Helpers/SetGlobalCompilerAndLinkerSettings.cmake
adrianinsaval: also I think there are other sutff required for ccache on windows:
https://github.com/FreeCAD/FreeCAD/blob/8f17842974a1c7fba4aaa0e30703690e1c18ffe7/.github/workflows/sub_buildWindows.yml#L127

https://github.com/ccache/ccache/wiki/MS-Visual-Studio#usage-with-cmake

adrianinsaval:
> I can remove the "Ubuntu 22.04" and "Ubuntu 20.04 (Conda)" to bring it to the current baseline "Ubuntu 20.04" and the conda environment for Linux "Ubuntu 22.04 (Conda)".
> 
> Is that what you're requesting?

I'd like to read 0penBrain's opinion on it first but yes, and also switch ubuntu 20.04 to use clang since the conda build will be using gcc if I'm not mistaken

.github/workflows/sub_buildUbuntu2204Conda.yml
```diff
@@ -0,0 +1,203 @@
...
+# *   Lesser General Public License for more details.                       *
+# *                                                                         *
+# *   You should have received a copy of the GNU Lesser General Public      *
+# *   License along with FreeCAD. If not, see                               *
+# *   <https://www.gnu.org/licenses/>.                                      *
+# *                                                                         *
+# ***************************************************************************
+
+# This is a build and test workflow for CI of FreeCAD.
+# This workflow aims at building and testing FreeCAD on Ubuntu 22.04 using Clang.
```
adrianinsaval, line 25:
```suggestion
# This workflow aims at building and testing FreeCAD on a Conda environment on Linux
```

cMake/FreeCAD_Helpers/SetGlobalCompilerAndLinkerSettings.cmake
oursland: For the `conda` builds `ccache` is functioning with `ninja` for builds.  The caching is functioning as intended.

.github/workflows/sub_buildWindowsConda.yml
oursland: This is copied over from the current Windows builds.  If this should be changed, I'd be happy to make the change.

I'll enable it and see.

cMake/FreeCAD_Helpers/SetGlobalCompilerAndLinkerSettings.cmake
adrianinsaval: @0penBrain do you remember what the other stuff is for?

.github/workflows/sub_buildWindowsConda.yml
oursland: The C++ unit tests have [failed on Windows](https://github.com/oursland/FreeCAD/actions/runs/7065527459/job/19235714431#step:14:20), but succeeded on the [Windows on Conda](https://github.com/oursland/FreeCAD/actions/runs/7065527459/job/19235714621#step:14:22).

I also unfortunately discovered that [build failures on Ubuntu will be considered successful builds](https://github.com/oursland/FreeCAD/actions/runs/7065250572/job/19234889073#step:9:51650).  I believe this may be due to the [redirection in the build script](https://github.com/FreeCAD/FreeCAD/blob/main/.github/workflows/actions/linux/build/action.yml#L51), but I am not entirely certain.

oursland:
> I don't get what's the benefit in testing with ubuntu 20.04+conda and 22.04+conda, conda should not be distro dependent so testing on just one should be sufficient.
> 
> instead it could be a good idea to test just on ubuntu 20.04+clang with system deps (this is our baseline for minimum supported library versions) and with ubuntu-latest+conda+gcc so we'll remain with just 2 linux tests. While it's true that github actions are free IMO this doesn't mean we should abuse the system and run superfluous CI just because we can.

I have made the changes to build on Ubuntu 20.04 using Clang 12 with OpenMP 5.0.  The package `clang` is Clang 10 and only supports OpenMP 4.5, and fails to build the source as it includes C++11 range-based `for` loops.

The Ubuntu 22.04 with Conda build does use `gcc`, however the version installed by conda may not match that in the Ubuntu 20.04 distribution.  If the intent is to ensure that an install of Ubuntu 20.04 can build FreeCAD, we may want to split the exiting Ubuntu 20.04 build into two: Ubuntu 20.04 (GCC), and Ubuntu 20.04 (Clang) to ensure compiler compatibility on that platform.

adrianinsaval:
with this we are increasing the number of caches that each run will generate, making cache space very precious (we can have up to 10gb). I think we should get rid of the LibPack cache @0penBrain, I have my doubts that it really speeds up the download that much since it's hosted on github servers anyways plus it makes it more cumbersome to update the Libpack because you need to manually delete the Libpack cache.

It might also be worth considering reducing the maximum cache size for ccache.

we also need to consider if it's really worth caching the conda environment as it seems to be rather large, @oursland how much of an impact does it make having that cached?

oursland:
PR discussion meeting: set the Ubuntu 20.04 back to GCC.

* Ubuntu 20.04 will test GCC
* macOS Conda will test clang

PR #11601, opened 2023-12-01
PD: Add test for helix

wwmayer:
This test will fail because of a regression in PartDesign helix


yorikvanhavre:
Depends on #11598

PR #11602, opened 2023-12-01
Simplify dragging of App::AnnotationLabel

hlorus:
Currently AnnotationLabel's have an edit mode to place them in the scene which makes the interaction unnecessarily complex as seen in the following gif:

https://github.com/FreeCAD/FreeCAD/assets/64740362/13253277-bf80-4468-8f66-778ff7a885b6

This PR removes the editmode and allows manipulating the annotations by simply grabbing them:

https://github.com/FreeCAD/FreeCAD/assets/64740362/8f904255-64b3-43c8-b9dc-98d51daa6648

This is done by using the SoDragger's [setPartAsPath()](https://developer.openinventor.com/refmans/latest/RefManJava/com/openinventor/inventor/nodekits/SoInteractionKit.html#setPartAsPath(java.lang.String,com.openinventor.inventor.SoPath)) functionaliity.


hlorus:
The dragger is still shown during manipulation. This could probably also be removed simce it doesn't add much Information. From the docs that should be possible by setting the feedbackActive part to Null like this:
dragger.setPart("feedbackActive", NULL);

For some reason i didn't have much luck with this, not sure why...

luzpaz:
@wwmayer would you have a moment to take a look at this ?

wwmayer:
It basically works but the draggers do not serve any real purpose any more. Thus, I wonder whether they shouldn't be shown any more.

wwmayer:
> This is done by using the SoDragger's setPartAsPath() functionaliity.

Does this video show something that's not part of your PR? I could watch your first video but not your second. It gives me an empty file when downloading it.

hlorus:
> > This is done by using the SoDragger's setPartAsPath() functionaliity.
> 
> Does this video show something that's not part of your PR? I could watch your first video but not your second. It gives me an empty file when downloading it.

No, it simply shows the dragging of a label with the PR applied.

hlorus:
> It basically works.
> 
> > The dragger is still shown during manipulation.
> 
> Yes, I think they shouldn't be shown any more.

Do you have a hint how it could be hidden. Afaik draggers don't have a visibility switch and as described above setting the feedbackActive part to null doesn't seem to work.

hlorus:
Ahh nevermind, it was the wrong part... The dragger is now also hidden during translation.

WandererFan:
merging per Monday Morning Merge Meeting decision 2023-12-11.

PR #11611, opened 2023-12-02
Part: Use materialIndex.setValuesPointer() only if the new values are different

### LINKED ISSUE #11517: FreeCAD turning extremely laggy

### Problem description

When turning on transparency in the appearance menu for any object, and then selecting any face (this is _only_ working with faces) and or keeping the face highlighted, then moving the 3D Viewport the FPS drop down to 4.
However when i _do not_ have any face of the transparent object selected and move the 3D Viewport everything is completely fine.
Having vertices or edges selected on the transparent object does not cause any issues either. _See video_

https://github.com/FreeCAD/FreeCAD/assets/40491077/0ffa4b3d-dc5c-4b9a-90dd-69a229c1219e

I have tried several of the graphics options already, disabled the double render pass for transparent objects and so on.
Different render styles (except for wireframe - because there are no faces) dont fix anything either.

### Full version info

```shell
[code]
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.20.2.29603 (Git)
Build type: Release
Branch: releases/FreeCAD-0-20
Installed mods: 
  * 3D_Printing_Tools
  * A2plus 0.4.60n
  * Assembly4 0.50.2
  * Autoload
  * CubeMenu
  * Curves 0.6.12
  * Defeaturing 1.2.0
  * dxf-library
  * fasteners 0.4.56
  * fcgear 1.0.0
  * freecad.gears 1.0.0
  * InventorLoader 1.4.0
  * kicadStepUpMod 10.19.4
  * Lithophane
  * Manipulator 1.5.0
  * parts_library
  * POV-Ray-Rendering
  * Render 2023.2.6
  * sheetmetal 0.2.63
  * ToolbarStyle
[/code]
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

Rexbas:
This fixes #11517. See https://github.com/FreeCAD/FreeCAD/issues/11517#issuecomment-1837139082 for the cause of the problem.

Using `materialIndex.setValuesPointer()` only when the values change prevents the continuous cache invalidation and fixes the low fps.

To make this work I also had to remove [these](https://github.com/FreeCAD/FreeCAD/commit/6711e53d1681e86d9ea88783fa9353a8b84c13ce#diff-e4fb66281089c4211e8116c2a8bf4e05736a12e47518ca5786f3481ddb4f714fL646-L648) lines. I am not sure what the intention of these lines was and if removing it causes any side effects. Please review carefully.

### DIFF
--- a/src/Mod/Part/Gui/SoBrepFaceSet.cpp
+++ b/src/Mod/Part/Gui/SoBrepFaceSet.cpp
@@ -733,7 +733,7 @@ bool SoBrepFaceSet::overrideMaterialBinding(SoGLRenderAction *action, SelContext
             SoMaterialBindingElement::set(state,SoMaterialBindingElement::OVERALL);
             SoOverrideElement::setMaterialBindingOverride(state, this, true);
             packedColors.push_back(diffuseColor);
-            SoLazyElement::setPacked(state, this,1, &packedColors[0], hasTransparency);
+            SoLazyElement::setPacked(state, this,1, packedColors.data(), hasTransparency);
             SoTextureEnabledElement::set(state,this,false);
 
             if(hasTransparency && action->isRenderingDelayedPaths()) {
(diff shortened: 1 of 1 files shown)


src/Mod/Part/Gui/SoBrepFaceSet.cpp
```diff
@@ -836,9 +833,11 @@ bool SoBrepFaceSet::overrideMaterialBinding(SoGLRenderAction *action, SelContext
             }
         }
 
-        SbBool notify = enableNotify(FALSE);
-        materialIndex.setValuesPointer(matIndex.size(),&matIndex[0]);
-        if(notify) enableNotify(notify);
+        if (materialIndex.getNum() != matIndex.size() || materialIndex.getValues(0) != &matIndex[0]) {
```
wwmayer, line 836: I can confirm that this PR fixes the problem. However, there are a few minor issues:
* `materialIndex.getNum() != matIndex.size()` causes a compiler warning `comparison of integers of different signs: 'int' and 'std::vector::size_type' (aka 'unsigned long') [-Wsign-compare]`
* `&matIndex[0]` is the old way to access the C array. Since C++11 it should be `matIndex.data()`

PR #11655, opened 2023-12-05
[TD]New tool AddOffsetVertex

edi271:
Add a new TechDraw_Extension tool which creates a cosmetic vertex in an offset position of a selected vertex. The offset values are entered using a panel in the Combo-View.

See discussion in https://forum.freecad.org/viewtopic.php?t=82715

Documentation in Wiki https://wiki.freecad.org/TechDraw_ExtensionAddOffsetVertex

The new tool is positioned in the toolbar below the VertexAtIntersection tool, using a drop down menu.

The utility tools have been extended by some functions. All these functions my be used in the Python Console or in macros. In the Python Console help texts are shown.


luzpaz:
Cool! Would you mind attaching a screencast of this feature in action?

edi271:
@pierreporte: The issue is not to create the tool. The issue is to reach an agreement in the community, which symbols are needed and how is their appearance.

The workflow could be similar to the Surface Finish Symbol tool:
- select one vertex to position the symbol
- start the tool
- a panel opens. Click button of desired symbol.

Roy-043:
Something is not quite right with the menu entry for this new feature: no menu text. For consistency I would expect both commands of the group to be listed in the menu between separators.

And why:
`TechDraw_CommandVertexCreationGroup` instead of `TechDraw_ExtensionVertexCreationGroup`?

add-offset-vertex-menu

edi271:
Issue solved in https://github.com/FreeCAD/FreeCAD/pull/11797.
The names of the tools depend on their position in the source code. The position in the ToolBar, MenuBar is independent of this.

PR #11657, opened 2023-12-05
[Techdraw] Add missing Line Styles to CMakeLists.txt

Syres916:
See discussion https://forum.freecad.org/viewtopic.php?t=83220


WandererFan:
Thanks for catching this.  These must have been there at some point, because they are in my build directory.  Not sure how they fell out.

PR #11675, opened 2023-12-07
Removed unnecessary float casts

mosfet80:
removed unnecessary casts


xtemp09:
```cpp
auto root = (float)(sqrt(2.0)/2.0);
```


`root` is float here.


```cpp
auto root = sqrtf(2.0)/2.0;
```


Here, `root` is double. Would it better to write:


```cpp
const float root = sqrtf(2.0f) / 2.0f;
```


to get rid of the lint warnings?

yorikvanhavre:
@mosfet80 could you adress @xtemp09 's observations? Also we think the example python code should not be removed because it explains the magic numbers right after

xtemp09:
@yorikvanhavre, as you have commented recently, these examples should be added to the documentation or added in doxygen-style annotation.

mosfet80:
> And, actually, compilation errors.
> 
> ```
> /home/runner/work/FreeCAD/FreeCAD/src/Gui/Camera.cpp:109:13: error: non-constant-expression cannot be narrowed from type 'double' to 'float' in initializer list [-Wc++11-narrowing]
>     return {root, 0, 0, root};
> ```

Some compilers promote the result to double when "double=float/double" operation is performed. I solved the problem by adding an "f" to the number so as to transform the double into a float.
with this modification we will have that:float=float/float

yorikvanhavre:
@xtemp09 in this case this commented python code is not an example but an explanation of the magic numbers right after. So they should stay at that position, and there is not much necessity of them being added to the documentation

yorikvanhavre:
Sorry I tried to fix this on my side but something goes wrong... I'll cherry-pick the commits

yorikvanhavre:
Superseded by #11772

Issue #10176, opened 2023-08-11
[Problem] Settings unexpectedly change after pressing Apply in the preferences


Serince:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
[code]
OS: Linux Mint 21.2 (X-Cinnamon/cinnamon)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33682 (Git)
Build type: Release
Branch: master
Hash: 9dce2525f8efe0873787f4388f10e11e168fe16b
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: Turkish/Turkey (tr_TR)
Installed mods: 
  * CfdOF 1.17.6
  * FEMbyGEN 1.2.0
[/code]
```

### Subproject(s) affected?

None

### Problem description

When I open freecad 0.21 on my hidpi screen it works fine, but when I open preferences and just press apply all the texts in report view, python console and macro editor shrinks. To fix it, I delete user.cfg in config folder and the problem is fixed.

### Anything else?

no user.cfg file
![noproblem](https://github.com/FreeCAD/FreeCAD/assets/10951891/8a655942-4627-41fd-a45f-e581212a8eab)

After pressing apply button in preferences
![problem](https://github.com/FreeCAD/FreeCAD/assets/10951891/850181f6-0b7f-472f-a39b-3ef3c0294a28)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Roy-043:
I can confirm this in a more general sense. The preferences (*.ui files) are also a repository for default values. These values sometimes do not match the defaults used in other parts of the code. Pressing "Apply" tells the program to use the values from the preferences (I think only from the current tab) even if the user has not made any changes.

Roy-043:
I have added all workbench labels since this may effect them all.

There are two potential solutions:
1. Make sure defaults in the code **always** match the defaults stored in the ui files of the preferences.
2. Read the ui files at runtime and extract default values from them. Forum topic with an attempt to do this for the Draft WB: https://forum.freecad.org/viewtopic.php?p=607927.

chennes:
Oops, sorry @Roy-043, I see now you added those tags on purpose... I thought it must have been an accident.

benj5378:
@luzpaz This looks like an undefined variable path. Seems like a new issue, so we could check if topology.py in FEM workbench have been changed lately.

[...]

PR #11677, opened 2023-12-07
Draft: new get_param function with defaults

Roy-043:
This PR introduces a new get_param function. For its default values it checks a dictionary largely filled by reading the preferences UI files. The goal is to have all defaults in one place to avoid inconsistencies. The dictionary is created once and then stored in a global variable.

Once merged this function will be implemented in Draft and Arch. The get_param function in utils.py will become obsolete.

Related issue: #10176

### DIFF
--- a/src/Mod/Draft/draftutils/params.py
+++ b/src/Mod/Draft/draftutils/params.py
@@ -21,17 +21,23 @@
 # *                                                                         *
 # ***************************************************************************
 
-""" Contains a parameter observer class and several related functions."""
+""" Contains a parameter observer class and parameter related functions."""
+
+import PySide.QtCore as QtCore
+import xml.etree.ElementTree as ET
 
 import FreeCAD as App
-import FreeCADGui as Gui
-from PySide import QtGui
+import Draft_rc
+import Arch_rc
 
 from draftutils import init_draft_statusbar
 from draftutils import utils
 from draftutils.translate import translate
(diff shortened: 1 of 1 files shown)


yorikvanhavre:
Very interesting... But it will only consider values stored in draft ui files right? What about pref values that are not rendered in these UI files? (Ex: values from generic FreeCAD preferences, or values that have no UI).

Of course for these cases we could simply not use the get_param structure... But then, wouldn't it be interesting to simply discard the get_param systme entirely, and rely on FreeCAD's default mechanism to get parameters (params.GetString, etc...)? Just thinking out loud here...

Roy-043:
There is a section in the code where parameters that are not available in the UI are added.
https://github.com/Roy-043/FreeCAD/blob/90de2d1f53111d287d2ff4cc3f785aefbb78ba41/src/Mod/Draft/draftutils/params.py#L332-L362

The main goal is to avoid having mutliple defaults for the same parameter in the code. The `params.GetString()` etc. functions do not provide that.

There is also some convenience involved.
Instead of:
```
param = App.ParamGet("User parameter:BaseApp/Preferences/Mod/Draft")
anno_scale = param.GetFloat("DraftAnnotationScale", 1)
```
You can do:
```
params.get_param("DraftAnnotationScale")
```

yorikvanhavre:
Okay excellent :) Thanks for the explanation!

PR #11679, opened 2023-12-07
[TechDraw] Use OCC to split intersecting edges

pavltom:
In certain views, there are unselectable closed areas, which can not be selected and thus not hatched. The main problem lied in assumption that edges are identical, if their end points are identical, however this can be true only for line segments, but not for arcs etc. Here is a basic figure of 180° arc intersected by 2 line segments illustrating the problem:

occ_edges_split
Rather than fixing this particular issue, I suggest to use OCC's General Fuse algorithm to perform the operation of fusing overlapping edges and splitting them at their intersections.

I am attaching a file containg a sketch consisting of 2 BSplines and 1 line segment and a view of this sketch on a TechDraw page. The current TechDraw implementation lets the user to select only 2 upper areas together at once, but not the bottom area. With the proposed fix all three areas may be selected separately.

[occ_edges_split.fcstd.zip](https://github.com/FreeCAD/FreeCAD/files/13600122/occ_edges_split.fcstd.zip)

I believe this change makes the results only better, but any indenpendent testing and help with search for issues is more than welcome.

### DIFF
--- a/src/Mod/TechDraw/App/DrawProjectSplit.cpp
+++ b/src/Mod/TechDraw/App/DrawProjectSplit.cpp
@@ -29,6 +29,7 @@
 #include <Bnd_Box.hxx>
 #include <BRep_Tool.hxx>
 #include <BRepAdaptor_Curve.hxx>
+#include <BOPAlgo_Builder.hxx>
 #include <BRepAlgoAPI_Common.hxx>
 #include <BRepAlgoAPI_Fuse.hxx>
 #include <BRepBndLib.hxx>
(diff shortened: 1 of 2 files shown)


src/Mod/TechDraw/App/DrawProjectSplit.cpp
```diff
@@ -514,25 +426,49 @@ std::vector<TopoDS_Edge> DrawProjectSplit::scrubEdges(std::vector<TopoDS_Edge>&
...
-        if (BRep_Tool::IsClosed(edge)) {
-            closedEdges.push_back(edge);
-        } else {
-            openEdges.push_back(edge);
+    const TopoDS_Shape &bopResult = bopBuilder.Shape();
+    if (!bopResult.IsNull()) {
+        TopExp_Explorer explorer(bopResult, TopAbs_EDGE);
+        while (explorer.More()) {
+            const TopoDS_Edge &edge = TopoDS::Edge(explorer.Current());
+            (BRep_Tool::IsClosed(edge) ? closedEdges : openEdges).push_back(edge);
```
WandererFan, line 470: This is a bit obscure as coded.  A simple 'if' would be easier for a maintainer to grasp down the road.

src/Mod/TechDraw/App/DrawProjectSplit.cpp
```diff
@@ -514,25 +426,49 @@ std::vector<TopoDS_Edge> DrawProjectSplit::scrubEdges(std::vector<TopoDS_Edge>&
...
-    //HLR usually delivers overlapping edges. We need to refine edge overlaps
-    //into non-overlapping pieces
-    std::vector<TopoDS_Edge> noOverlaps;
-    noOverlaps = DrawProjectSplit::removeOverlapEdges(origEdges);
 
-    //HLR algo does not provide all edge intersections.
-    //need to split edges at intersection points.
-    std::vector<TopoDS_Edge> splitEdges;
-    splitEdges = DrawProjectSplit::splitIntersectingEdges(noOverlaps);
+    TopTools_ListOfShape edgeList;
```
WandererFan, line 430: Would TopTools_IndexedMapOfShape (unique edges) be better than TopTools_ListOfShape (duplicate edges) here?  IDK, just asking.

WandererFan:
This is very cool. 

I don't think BOPAlgo_Builder is used directly anywhere in FreeCAD.  I've never run into it before this.   Wish I had known about it earlier.  

I'm going to merge.  You can make a new PR if any of my comments appeal to you.

chennes:
It wasn't used, and wasn't in the precompiled header file for OpenCascadeAll.h (in Part). Added by PR #11683.

PR #11682, opened 2023-12-07
[TechDraw] Collect doc info to fill template fields

Syres916:
In order to save users' mouse clicks and typing this improvement is to read in document information and give the user the ability to change those variables before committing to the template fields where it's deemed appropriate.

Error checking has been added so if the user does not have a page, view and template with editable fields the icon will be greyed out. The user is able to edit the _FillTemplateFields.csv_ file but if they corrupt the field headings or accidentally delete the file an error will be raised on changing to the TechDraw Wb. 

If the file has never been saved then the suggested date will be current date in the format DD/MM/YYYY

See discussion https://forum.freecad.org/viewtopic.php?t=82904

### DIFF
--- a/src/Mod/TechDraw/TechDrawTools/CommandFillTemplateFields.py
+++ b/src/Mod/TechDraw/TechDrawTools/CommandFillTemplateFields.py
@@ -1,0 +1,100 @@
+# ***************************************************************************
+# *   Copyright (c) 2023 Syres                                              *
+# *                                                                         *
+# *   This program is free software; you can redistribute it and/or modify  *
+# *   it under the terms of the GNU Lesser General Public License (LGPL)    *
+# *   as published by the Free Software Foundation; either version 2 of     *
+# *   the License, or (at your option) any later version.                   *
+# *   for detail see the LICENCE text file.                                 *
+# *                                                                         *
+# *   This program is distributed in the hope that it will be useful,       *
+# *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+# *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+# *   GNU Library General Public License for more details.                  *
+# *                                                                         *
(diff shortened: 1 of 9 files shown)


src/Mod/TechDraw/TechDrawTools/TaskFillTemplateFields.py
```diff
@@ -0,0 +1,462 @@
...
+                and listofkeys == reader.fieldnames
+            ):
+                self.page = obj
+                if obj.Views == []:
+                    msgBox = QtGui.QMessageBox()
+                    msgTitle = QtCore.QT_TRANSLATE_NOOP(
+                        "Techdraw_FillTemplateFields",
+                        "View or Projection Group missing",
+                    )
+                    msg = QtCore.QT_TRANSLATE_NOOP(
```
WandererFan, line 108: Might be different in Python, but AFAIK:

QT_TRANSLATE_NOOP(context, stringToTranslate) is for marking strings for translation so they
get picked up and sent to CrowdIn for the translators

QCoreApplication.translate(context, stringToTranslate) does the actual translating by searching the 
translation files and returning the translated version of stringToTranslate

src/Mod/TechDraw/TechDrawTools/CommandFillTemplateFields.py
```diff
@@ -0,0 +1,100 @@
...
+                            "CreatedByChkLst",
+                            "ScaleChkLst",
+                            "LabelChkLst",
+                            "CommentChkLst",
+                            "CompanyChkLst",
+                            "LicenseChkLst",
+                            "CreatedDateChkLst",
+                            "LastModifiedDateChkLst",
+                        ]
+                        with codecs.open(file_path, encoding="utf-8") as fp:
```
WandererFan, line 83: so...
\# if any of these keys are the names of editable fields in the template, etc, make this command available?

WandererFan:
Needs a wiki page too!

PR #6924, opened 2022-05-27
Bugfix: support DXF files generated by deepnest, try to fix misaligned DXF stream reads.

sundtek:
Fix reading DXF files generated by deepnest
https://forum.freecadweb.org/viewtopic.php?f=3&t=68997&p=598001

Thank you for creating a pull request to contribute to FreeCAD! To ease integration, we ask you to conform to the following items. Pull requests which don't satisfy all the items below might be rejected. If you are in doubt with any of the items below, don't hesitate to ask for help in the [FreeCAD forum](https://forum.freecadweb.org/viewforum.php?f=10)!

- [ ]  Your pull request is confined strictly to a single module. That is, all the files changed by your pull request are either in `App`, `Base`, `Gui` or one of the `Mod` subfolders. If you need to make changes in several locations, make several pull requests and wait for the first one to be merged before submitting the next ones
- [ ]  In case your pull request does more than just fixing small bugs, make sure you discussed your ideas with other developers on the FreeCAD forum
- [ ]  Your branch is [rebased](https://git-scm.com/docs/git-rebase) on latest master `git pull --rebase upstream master`
- [ ]  All FreeCAD unit tests are confirmed to pass by running `./bin/FreeCAD --run-test 0`
- [ ]  All commit messages are [well-written](https://chris.beams.io/posts/git-commit/) ex: `Fixes typo in Draft Move command text`
- [ ]  Your pull request is well written and has a good description, and its title starts with the module name, ex: `Draft: Fixed typos`
- [ ]  Commit messages include `issue #<id>` or `fixes #<id>` where `<id>` is the issue ID number from our [Issues database](https://github.com/FreeCAD/FreeCAD/issues) in case a particular commit solves or is related to an existing issue. Ex: `Draft: fix typos - fixes #4805`

And please remember to update the Wiki with the features added or changed once this PR is merged.  
**Note**: If you don't have wiki access, then please mention your contribution on the [0.20 Changelog Forum Thread](https://forum.freecadweb.org/viewtopic.php?f=10&t=56135).

---

PR #11701, opened 2023-12-11
Make C++ DXF importer set entity colors per DXF contents

### LINKED ISSUE #8806: [Problem] DXF import: "Get original colors from the DXF file" doesn't work in C++ importer

kpemartin:
Fixes #8806
Fixes #6924
This change:

1. Provides a hook for GUI-dependent code in the C++ importer (since color is a Gui concept) via a new class `ImpExpDxfReadGui` derived from `ImpExpDxfRead` and ultimately created by importDXF.py if the ImportGui module is available.
2. Sets the color on created FreeCAD objects based on the color specified in the DXF file, including handling BYLAYER color, using a new virtual methods which do nothing in `ImpExpDxfRead` but set color in `ImpExpDxfReadGui`.
3. Refactors the reader code in CDxfRead to remove substantial duplication involved in reading the individual attribute records that make up each entity. My goal here was to eliminate, or at least greatly reduce, the lint that shows up when a PR is submitted, and on noticing that it was mostly the same lint duplicated about a dozen times, it seemed clear that the code should be shared rather than duplicated. The refactored code is also much less likely to get out of sync between reading record types and record contents (the cause for #6924).
4. Annotates (on the Python console) unimplemented features of DXF. This annotation is incomplete (there are still plenty of features that are quietly ignored), but it might help someone importing a file to realize what might be missing from their drawing.

### DIFF
--- a/src/Mod/Draft/Resources/ui/preferences-dxf.ui
+++ b/src/Mod/Draft/Resources/ui/preferences-dxf.ui
@@ -335,11 +335,11 @@ Example: for files in millimeters: 1, in centimeters: 10,
         <item>
          <widget class="Gui::PrefCheckBox" name="checkBox_2">
           <property name="toolTip">
-           <string>Colors will be retrieved from the DXF objects whenever possible.
-Otherwise default colors will be applied. </string>
+           <string>Colors will set as specified in the DXF file whenever possible.
+Otherwise default colors will be applied.</string>
           </property>
           <property name="text">
-           <string>Get original colors from the DXF file (legacy importer only)</string>
+           <string>Use colors from the DXF file</string>
           </property>
           <property name="prefEntry" stdset="0">
            <cstring>dxfGetOriginalColors</cstring>
--- a/src/Mod/Draft/importDXF.py
+++ b/src/Mod/Draft/importDXF.py
@@ -2813,8 +2813,12 @@ def open(filename):
         doc = FreeCAD.newDocument(docname)
         doc.Label = docname
         FreeCAD.setActiveDocument(doc.Name)
-        import Import
-        Import.readDXF(filename)
+        try:
+            import ImportGui
+            ImportGui.readDXF(filename)
+        except Exception:
+            import Import
+            Import.readDXF(filename)
         Draft.convert_draft_texts() # convert annotations to Draft texts
         doc.recompute()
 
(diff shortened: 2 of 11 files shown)


yorikvanhavre:
Excellent! I like the general idea a lot and there is a welcome cleaning done there.
I'd just advise as much as possible to separate different features into different PRs rather than make a big all-in-one...

kpemartin:
Unfortunately the handling for color and the general refactoring became inextricably linked. The two other changes (Gui support for c++ importer and changing the preferences label) are separable, but the latter belongs with the color handling change.

So I suppose the Gui import support could separate out first, but it would have to be done first.

yorikvanhavre:
Yeah, i understand. Don't worry about it, it's still well contained and not out of hand ;)

kpemartin:
Hmmm, it looks like I have to deal with a different C++ version on Ubuntu_20-04 which does not support specializations of class-level template functions. Which C++ standard is Ubuntu_20-04 using?

There is a third error which I hope is just caused by the previous two.

kpemartin:
Forgive my confusion here... What are all these other commits that appear to have been added to my PR? Are they related at all to my changes? A lot of this looks like housekeeping.

I was about to mark this PR as "ready for review" but now I'm confused by all these other changes. For that matter, I find it interesting how much review a draft PR gets before it is "ready for review" :-)

kpemartin:
I think I have it cleaned up now. Not sure what hit me, though. It may be because I can't really decide whether on my dev system I should pull directly from upstream, or first do a sync on my Github repository and then pull from origin. If I consistently did the latter I think I would not need the force-push but that extra step is, well, an extra step. But doing a mix of the two is asking for trouble, it seems. Or maybe somewhere I did a fetch and not a pull or forgot a rebase.

yorikvanhavre:
great! Merging then :)

Issue #9277, opened 2023-04-15
[Problem] Folder may not be hidden by spacebar (possible App::Link regression in Group extension used by BuildingPart object)


PeterDFC:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Forums discussion

https://forum.freecad.org/viewtopic.php?p=613178#p613178

### Version

0.21 (Development)

### Full version info

```shell
OS: Debian GNU/Linux 11 (bullseye) (XFCE/xfce)
Word size of FreeCAD: 64-bit
Version: 0.21.0.32821 (Git) AppImage
Build type: Release
Branch: master
Hash: eaf9d43f37238c33dd1c47b859a6fdf5b7f629fc
Python 3.10.10, Qt 5.15.8, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * dodo 1.0.0
  * BIM 2021.12.0
```

### Subproject(s) affected?

None

### Issue description

when the buildingPart "top_floor" is shown, the "topplate_slab" can be shown and hidden with the spacebar (example below). The folder "walls_top_floor", on the other hand, cannot be shown again with the space bar after it has been hidden.
This causes confusion and should be fixed, if there are no important reasons against it.

Originally reported on Jul 26, 2022 with an older FC version at that time:

```shell
OS: Ubuntu Core 20 (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.20.29177 (Git) Snap 161
Build type: Release
Branch: tag: 0.20
Hash: 68e337670e227889217652ddac593c93b5e8dc94
Python 3.8.10, Qt 5.15.3, Coin 4.0.0, Vtk 7.1.1, OCC 7.6.2
Locale: German/Germany (de_DE)
Installed mods: 
  * BIM 2021.12.0
```

### Anything else?
[directory_space.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/11240031/directory_space.FCStd.zip)

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
@yorikvanhavre have you got any clue what is happening here?

yorikvanhavre:
[...]

PR #11704, opened 2023-12-11
Add test for child visibility in Arch BuildingPart or GroupExtensions

bgbsww:
Fix for #9277.

This fixes the immediate issue with space bar or Std_ToggleVisibility,  Thats a step forward.  It seems to allow all the visibility switches at each level to not be overridden when you set them manually,  fixing that issue as well.

I'm not sure if the exact behaviors when you have architectural parts inside a group inside a BuildingPart are completely what we want - you can turn visibility on and off at each level, and it affects the parts at the bottom just fine.  I'm not sure what our expected behaviors are for all permutations of Vis On/Off inside of Vis On/Off inside of Vis On/Off.

The complexity level here is fairly high ( at least for my experience level in this codebase), and it took quite a bit of debug time to locate the issue.

### DIFF
--- a/src/App/GroupExtension.cpp
+++ b/src/App/GroupExtension.cpp
@@ -365,7 +365,7 @@ void GroupExtension::extensionOnChanged(const Property* p) {
 
 void GroupExtension::slotChildChanged(const DocumentObject &obj, const Property &prop) {
     if(&prop == &obj.Visibility)
-        _GroupTouched.touch();}
+        _GroupTouched.touch();
 }
 
 bool GroupExtension::extensionGetSubObject(DocumentObject *&ret, const char *subname,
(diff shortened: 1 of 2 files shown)


yorikvanhavre:
This seems to work fine for Arch objects indeed, but there might be more involved, that I can't foresee... IIRC this was part of Realthunder's work... @realthunder could you have a quick look here if that solution might present problems in your opinion?

realthunder:
See my fix at #11763

bgbsww:
I've removed the code change and just left the test, which does pass after #11763 is applied.

yorikvanhavre:
Awesome! Waiting for the CI tests to finish and we can merge this. Thanks to both of you!

Freedman-CB1:
The expected behavior was to change the Transparency at the object level. Part and Group don't have their own transparency property so I would not expect them to be included in the selection process. I'm not sure about Arch. The idea of the toggle was a quick selection of an object/objects in the Workbench including Sketcher. Valid Sketcher selections will be from the tree only because Sketcher is sketch object modal in the 3D. Maybe the name "Transparency toggle" needs to be "Transparency Object toggle" for clarity.

bgbsww:
@Freedman-CB1 did you mean to be commenting on #11763 ?

Freedman-CB1:
No, I will try to fix this. Thanks

Issue #10538, opened 2023-09-04
PartDesign: No good way to create walls / thin extrusions / ribs / web


kadet1090:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Currently there is no good way to create "Thin" elements in the FreeCAD. Because of that creating wall-like elements in the Part Design is really hard and requires a lot of effort. Possible uses:
- Creating walls
- Creating ribs
- Creating webs
- Creating reinforcements

To illustrate this feature in OnShape:
![image](https://github.com/FreeCAD/FreeCAD/assets/747404/204b76f0-6916-418c-a3b6-a642420bb41a)

It'd be nice to implement this in a way that would allow this feature to be easily used in most (all?) features without code duplication, so It probably should be implemented as separate "thicken" feature, that then can be: padded, pocketed, revolved, etc. 

This potential feature should also allow control on:
 - Offset (One way, symmetric, Two Distances)
 - Cap type (Round / Butt)
 - Join type (Round / Tangent)
 
If possible, there should be a way to create "Thicken" feature using for example "Pad" dialog, to simplify user workflow and make it more user friendly.

### Workarounds
- Using SheetMetal add base: https://wiki.freecad.org/SheetMetal_AddBase
- Creating SubShape binder to sketch with filled offset applied - it works, but it is not intended use of binder, and it has a lot of quirks
- Using Part workbench with 2d offset
- Sketch Offset tool will help a lot with many cases when (if) merged, but again it's not correct tool for the job
- Using Arch_Wall ?

### Full version info

```shell
FreeCAD 0.21
```

### Subproject(s) affected?

PartDesign

### Anything else?

 - https://forum.freecad.org/viewtopic.php?t=31624
 - https://youtu.be/c98dkN8yt-4?t=717 - Clough42 video featuring this in Fusion360
 - https://youtu.be/hP4UNtoYGSc - Video introducing that feature to OnShape
 
Technically, this is duplicate of https://github.com/FreeCAD/FreeCAD/issues/6057 but this issue lacks description and is old so It'd be hard to get proper feedback.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kadet1090:
[...]

PR #11707, opened 2023-12-12
[Sketcher] Fix offset of single line

### LINKED ISSUE #11706: [Problem] Sketcher offset do not work on single line. And seems to crash on macOs

### Problem description

Related to https://github.com/FreeCAD/FreeCAD/issues/6538

Offseting one line in sketcher does not work.
Reading this : https://dev.opencascade.org//content/offset-problem
Seems to indicate that with only one line occ cannot determine the plane of the offset. It offers a fix.

But currently we are using the class `Part::BRepOffsetAPI_MakeOffsetFix` instead of `Part::BRepOffsetAPI_MakeOffset`.
Which purpose seems to be to avoid single line problem.

This needs further investigation.

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Sketcher

### Anything else?

_No response_

### Code of Conduct

PaddleStroke:
Fixing one line offset in sketcher by removing the use of BRepOffsetAPI_MakeOffsetFix and using OCC BRepOffsetAPI_MakeOffset directly.
Fixes #11706
Discussion : https://forum.freecad.org/viewtopic.php?p=725100#p725100

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerOffset.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerOffset.h
@@ -29,14 +29,15 @@
 #include <BRep_Tool.hxx>
 #include <BRepAdaptor_Curve.hxx>
 #include <BRepClass_FaceClassifier.hxx>
-#include <Mod/Part/App/BRepOffsetAPI_MakeOffsetFix.h>
 #include <BRepBuilderAPI_Copy.hxx>
+#include <BRepOffsetAPI_MakeOffset.hxx>
 #include <BRepBuilderAPI_MakeWire.hxx>
 #include <BRepBuilderAPI_MakeFace.hxx>
 #include <BRepBuilderAPI_MakeVertex.hxx>
 #include <BRepBuilderAPI.hxx>
 #include <BRepExtrema_DistShapeShape.hxx>
 #include <TopoDS.hxx>
+#include <gp_Pln.hxx>
 
 #include <Base/Exception.h>
 
@@ -118,6 +119,7 @@ public:
         , deleteOriginal(false)
         , offsetLengthSet(false)
         , offsetConstraint(false)
+        , onlySingleLines(true)
         , offsetLength(1.)
     {}
 
@@ -203,7 +205,7 @@ private:
     Base::Vector2d endpoint, pointOnSourceWire;
     std::vector<TopoDS_Wire> sourceWires;
 
-    bool deleteOriginal, offsetLengthSet, offsetConstraint;
+    bool deleteOriginal, offsetLengthSet, offsetConstraint, onlySingleLines;
     double offsetLength;
     int firstCurveCreated;
 
@@ -213,7 +215,18 @@ private:
         short joinType =
             constructionMethod() == DrawSketchHandlerOffset::ConstructionMethod::Arc ? 0 : 2;
 
-        Part::BRepOffsetAPI_MakeOffsetFix mkOffset(GeomAbs_JoinType(joinType), allowOpenResult);
+        // Offset will fail for single lines if we don't set a plane in ctor.
+        // But if we set a plane, then the direction of offset is forced...
+        // so we set a plane if and only if there are not a single sourceWires with more than single
+        // line.
+        BRepOffsetAPI_MakeOffset mkOffset;
+
+        if (onlySingleLines) {
+            TopoDS_Face workingPlane = BRepBuilderAPI_MakeFace(gp_Pln(gp::Origin(), gp::DZ()));
+            mkOffset = BRepOffsetAPI_MakeOffset(workingPlane);
+        }
+        mkOffset.Init(GeomAbs_JoinType(joinType), allowOpenResult);
+
         for (TopoDS_Wire& wire : sourceWires) {
             mkOffset.AddWire(wire);
         }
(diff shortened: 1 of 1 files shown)


abdullahtahiriyo:
I am having a local compilation error with this PR, which might come from OCC version (as here it is building). I will investigate.

abdullahtahiriyo:
@PaddleStroke 

Ok, I can compile by adding one header. I do understand why the header is necessary. I do not understand why the CI did not need it. But is easy.

What I wanted to check with you is the behaviour of offset after this tool. The offset of the line is a little bit odd (I am unsure this is what you actually want as result, so kind of slot with two line segments...). The offset of a rectangle seems broken to me (I cannot seem to make an external offset). Please take a look:

[Peek 2023-12-17 16-26.webm](https://github.com/FreeCAD/FreeCAD/assets/5249162/373af4da-5dd6-4ae7-870b-f9a7e6254d7d)

For reference:

```
OS: Ubuntu 22.04.3 LTS (MATE/mate)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35306 +1 (Git)
Build type: Debug
Branch: offset_single_line_fix
Hash: 6be16ae2c6756d79d97a4ae2c52ff2034985afb4
Python 3.10.12, Qt 5.15.3, Coin 4.0.0, Vtk 7.1.1, OCC 7.5.2
```

abdullahtahiriyo:
This is the header I need:
https://github.com/abdullahtahiriyo/FreeCAD_sf_master/commit/c17b4ba4bd133b47da9921205fec6d366565c907

PaddleStroke:
Thanks for the report !
1 - Yes the single line offset is a bit weird, but it's the result of the OCC kernel. My goal here is to prevent the tool from failing on single lines. In particular because someone said that on macOs it crashes.
2 - Regarding the rectangle, this is interesting. In my early testing I already have this strange behavior on some shapes, I couldn't determine yet what makes this happen or not. When testing this PR on my usual shape, I did not notice a problem on my rectangle. I will run some more test to see if I can finally understand this bug and if yours is the same or a new one.

adrianinsaval:
@oursland @0penBrain here the ubuntu 20.04 build failed but it reports as successfull https://github.com/FreeCAD/FreeCAD/actions/runs/7178556454/job/19546990047?pr=11707, any clue why?

adrianinsaval:
IIRC path wb uses an independent offset library because occt's offset tool is unreliable, would it be a good idea to use that in sketcher too?

oursland:
> @oursland @0penBrain here the ubuntu 20.04 build failed but it reports as successfull https://github.com/FreeCAD/FreeCAD/actions/runs/7178556454/job/19546990047?pr=11707, any clue why?

Yes-ish.  I noted this during the PR call two weeks ago, and suspect it has to do with the way [output redirection is handled here](https://github.com/FreeCAD/FreeCAD/blob/625a4212cc75b4fa7da6eb15f6538942290f8364/.github/workflows/actions/linux/build/action.yml#L51).  I believe this is due to `stdbuf`, possibly returning a success code despite `cmake` returning a failure code.  As `macOS` does not have a `stdbuf` command, [this is not present](https://github.com/FreeCAD/FreeCAD/blob/main/.github/workflows/actions/macos/build/action.yml#L51C88-L51C88) and failures are represented correctly.

I am not 100% certain this is the cause, it just appears to be a difference that may explain the issue.  We can see if removing the call to `stdbuf` correctly propagates the build exit code.

oursland:
@adrianinsaval PR https://github.com/FreeCAD/FreeCAD/pull/11758 has been opened to address the incorrect build success from CI.

PaddleStroke:
@abdullahtahiriyo missing header added.
@adrianinsaval changing the offset kernel could be interesting but its a much bigger project.

PaddleStroke:
@abdullahtahiriyo I confirm that there is a problem, now the offset can only be drawn in one direction I don't know why. I will check.

PaddleStroke:
So it appears that setting a plane to the `BRepOffsetAPI_MakeOffset` is what is preventing the offset from working in both directions...

```
        TopoDS_Face workingPlane = BRepBuilderAPI_MakeFace(gp_Pln(gp::Origin(), gp::DZ()));
        BRepOffsetAPI_MakeOffset mkOffset(workingPlane, GeomAbs_JoinType(joinType), allowOpenResult);
```
So... I will make an exception for single line I guess.

PaddleStroke:
So I pushed changes to handle the case of single lines (one or more actually).

kadet1090:
> IIRC path wb uses an independent offset library because occt's offset tool is unreliable, would it be a good idea to use that in sketcher too?

IIRC library used in path does not support bsplines, the same goes for implementation in the arch for walls. Creating offsets is quite trivial mathematically as long as you don't care about more generic curves.  We could try to create few kernels for computing the offsets - one that is more optimized for easier cases that do not have bsplines, one to handle more generic cases and possibly more. I tried to do something like that for #10538, but it still requires more work to be done.

macdroid53:
In recent builds I can offset single lines and arcs. This is nice. 
Is there a switch of keyboard that can the left shown below. I almost never want a sausage.
Screenshot_20231220_125208

PaddleStroke:
@macdroid53 This is because currently we do not 'allow open result'. We could add one checkbox to the tool of course.
The issue is that this OCC Offset API option is not reliable and was failing a lot.

macdroid53:
Unfortunate.

PR #11729, opened 2023-12-14
Streamline scaling for DXF import

kpemartin:
Eliminate m_measurement_inch to clean up logic for priority of MEASUREMENT and INSUNITS. Save the actual scaling factor rather than the scaling enum so a switch statement is not executed for each call to mm() Add to CDxfRead the work to handle dxfScaling option, ImpExpDxfRead just has to set it up now. Get the scaling factor from a lookup table rather than a switch statement Display a message explaining what the scaling factor is and where it comes from

### DIFF
--- a/src/Mod/Import/App/dxf/ImpExpDxf.cpp
+++ b/src/Mod/Import/App/dxf/ImpExpDxf.cpp
@@ -78,10 +78,12 @@ using BRepAdaptor_HCurve = BRepAdaptor_Curve;
 
 //******************************************************************************
 // reading
-ImpExpDxfRead::ImpExpDxfRead(std::string filepath, App::Document* pcDoc)
-    : CDxfRead(filepath.c_str())
+ImpExpDxfRead::ImpExpDxfRead(const std::string& filepath, App::Document* pcDoc)
+    : CDxfRead(filepath)
+    , document(pcDoc)
+    , optionGroupLayers(false)
+    , optionImportAnnotations(true)
 {
-    document = pcDoc;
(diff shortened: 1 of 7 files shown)


kpemartin:
Still want to clean up some of the lint

kpemartin:
Hmmm, I have cpplint complaining about some of the NOLINT categories I'm using to quiet down Clang-tidy.
Cpplint has a category for complaining about unknown lint categories (readability/nolint) but does not seem to understand NOLINTBEGIN/NOLINEND so a blanket suppression does not work.

I will next try to precede each NOLINTNEXTLINE or NOLINT with single-line NOLINTNEXTLINE(readability/nolint) to see if that shuts it up. Although this will eliminate the spurious lint messages it is still a PITA to have to do.

On the other hand because cpplint doesn't understand the BEGIN/END directives I don't have to protect those since cpplint doesn't realize these contain lint categories!

However, it would be much easier if cpplint didn't believe it was the only lint in the universe and that it therefore has a complete list of all the valid lint categories. Maybe it is possible to modify the automatic Lint check to suppress this category (readability/nolint) by default.

kpemartin:
Using  NOLINTNEXTLINE(readability/nolint) before each clang-tidy NOLINTNEXTLINE or NOLINT directive has indeed suppressed the cpplint complaints about clang-tidy lint categories. Still a real pain to have to do this.

yorikvanhavre:
I might have introduced a conflict here by merging something else earlier, sorry about that... Mind to have a look @kpemartin ?

yorikvanhavre:
Thanks!

Issue #11713, opened 2023-12-12
[Problem] PartDesign: Revolution - Selection of Reference Axis not possible


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

PartDesign Revolution feature allows to select a reference axis in the viewport. The reference is selected / displayed correctly in the preview, but I cannot confirm the task dialog.
In the following example I've selected the X-Axis by clicking on Axis: Select Reference > and clicked on the X-axis in the viewport.
The dropdown menu still shows Select reference but nothing else can be selected and the X-axis is used for the preview. 
Clicking ok results in the following error. It seems to have set the selection for the preview but not for the feature itself.

![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/b065462b-3707-45c3-a49b-0f7cb933ce54)

### Full version info

```
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35370 (Git)
Build type: Release
Branch: main
Hash: 1066bb54ceec8f4f8376b5149253acda5677651d
Python 3.10.13, Qt 5.15.8, Coin 4.0.2, Vtk 9.2.6, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * 3DfindIT 1.2.0
  * 3D_Printing_Tools
  * Assembly4 0.12.6
  * Curves 0.6.8
  * Defeaturing 1.2.0
  * fasteners 0.4.56
  * OpenDark 2023.10.22
  * sheetmetal 0.2.63
```

### Subproject(s) affected?

PartDesign

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

pierreporte:
I have the same problem. Whatever you select in the 3D view, the dialog appears: sketch line, datum axis, even the base axis. It only works if the axis combo box is used.

Syres916:
@AjinkyaDahale it looks as though there has been a regression in the recent changes, up-to-date 0.21 branch works as I would expect using a DatumLine as a reference axis. Using the main brnach, if I copy the Python code into the Console, then it corrects the model for example:

```
App.ActiveDocument.getObject('Revolution').Angle = 360.000000
App.ActiveDocument.getObject('Revolution').ReferenceAxis = (App.getDocument('RevolveTest').getObject('DatumLine'), [''])
App.ActiveDocument.getObject('Revolution').Midplane = 0
[...]

PR #11736, opened 2023-12-15
[PartDesign] Revolution fix regression in reference selection

Syres916:
Fix regression introduced by https://github.com/FreeCAD/FreeCAD/commit/2cef4cfdbe1a6b9ebf50a85c04fbe99f23cd3829
as highlighted in https://github.com/FreeCAD/FreeCAD/issues/11713 but this needs thoroughly testing by a PartDesign power user as I'm not fully understanding what the author was trying to achieve with the change in the first place.

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
@@ -357,7 +357,10 @@ void TaskRevolutionParameters::updateUI(int index)
 void TaskRevolutionParameters::onSelectionChanged(const Gui::SelectionChanges& msg)
 {
     if (msg.Type == Gui::SelectionChanges::AddSelection) {
-        if (selectionFace) {
+        int mode = ui->changeMode->currentIndex();
+        if (static_cast<PartDesign::Revolution::RevolMethod>(mode)
+                == PartDesign::Revolution::RevolMethod::ToFace
+            && selectionFace) {
             QString refText = onAddSelection(msg);
             if (refText.length() > 0) {
                 QSignalBlocker block(ui->lineFaceName);
@@ -371,7 +374,7 @@ void TaskRevolutionParameters::onSelectionChanged(const Gui::SelectionChanges& m
                 clearFaceName();
             }
         }
-        else {
+        else { // Reference Selection for datum lines and edges
             exitSelectionMode();
             std::vector<std::string> axis;
             App::DocumentObject* selObj;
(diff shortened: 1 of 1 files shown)


Syres916:
@AjinkyaDahale and @adrianinsaval as the author and maintainer involved in the original PR, your input would be appreciated.

AjinkyaDahale:
@Syres916 have you closed this PR because of some problem? If not intended, I can take a look sometime early next week. Hopepullye before the merge meetup

Syres916:
I closed it while I worked out why the tests were failing, hopefully it'll pass now.

oursland:
@Syres916 Use the draft status to indicate that it is still a work-in-progress.  Closing it appears to be abandoning the issue, which can be disappointing for a valuable change.

Syres916:
@chennes Can I please ask why this has been put to Draft?

adrianinsaval:
likely they misunderstood the above comments to mean that this PR was not ready for merge yet

src/Mod/PartDesign/Gui/TaskRevolutionParameters.cpp
```diff
@@ -357,7 +357,10 @@ void TaskRevolutionParameters::updateUI(int index)
 void TaskRevolutionParameters::onSelectionChanged(const Gui::SelectionChanges& msg)
 {
     if (msg.Type == Gui::SelectionChanges::AddSelection) {
-        if (selectionFace) {
+        int mode = ui->changeMode->currentIndex();
+        if (static_cast<PartDesign::Revolution::RevolMethod>(mode)
+                == PartDesign::Revolution::RevolMethod::ToFace
+            && selectionFace) {
```
AjinkyaDahale, line 363: This change seems like it shouldn't be needed. @Syres916 could you share your thought process here?

AjinkyaDahale: I just tested without this addition and just the `updateUI`. That works initially, but fails to work perfectly well after a few attempts.

Syres916: Well it was seemed to make sense (at least to me) that if selected face was already true then only with To Face mode would it be linked but as I've stated I don't understand what you were actually trying to achieve with the original PR. If it's going to help, I'll close this and just let you bring forward a proper fix.

AjinkyaDahale: Please don't close.

The original PR was to add support for certain other means of rotation, like up to a certain face. That needed a large overhaul of the entire tool, so regressions and issues got introduced. It is actually surprising that axis selection was affected because that was one of the few things not touched (see "onAxisChanged").

AjinkyaDahale: OK with this change the issue can still happen if we are in the "up to face" mode and want to change the revolution axis. I think making `selectionFace` false outside the context might be a better option, see #11822.

PR #11717, opened 2023-12-13
[Draft] Concise Sketcher Toolbars

obelisk79:
This is an experimental test of new condensed and reorganized toolbars for the sketcher workbench and is being submitted so test builds can be created and distributed to acquire broad user feedback.

This new arrangement is based on UX guidelines and principles as outlined in the FreeCAD Design Guide, and is a culmination of weeks of collaboration and discussion among the Design Working Group on Discord. 

The following toolbar changes have been made:
**Sketcher Edit Mode:**
-_Leave Sketch_ button removed. Sketches exit edit mode via Escape key.

**Sketcher Geometries:**
- _Sketcher_CreateLine_ deprecated, same functionality provided by the more versatile polyline tool with minimal impact to user interaction even when drawing multiple disjointed lines.
- _Sketcher_CreateHexagon_ is now the sole polygon tool. The command-group was unnecessary after the new tool settings feature has been added. Up and Down arrow on the keyboard now change the polygon count. One tool, multiple uses. The individual keyboard shortcuts for each polygon are also still valid.
- Conics command group has been divided and added to the circle and arc groups.
- Slot and Arc slot are in a command group together.
- _Sketcher_CreatePoint_ added at the end of the geometry element tools due to it being less frequently used and the simplest sketcher element.
- _Sketcher_Offset_ has been moved to this toolbar as it creates geometry, but is a modifier similar to fillet and trim
- _Sketcher_Offset_ icon has been modified to be visually consistent with the rest of its new toolbar.
- _Sketcher_CreatePointFillet_ has been made the default tool in the fillet command group. Preservation of constraints is more frequently desired than not.
- _Sketcher_CarbonCopy_ removed and placed into the Sketcher Tools toolbar.

**Sketcher Constraints:**
- _Sketcher_Dimension_ tool has been placed at the first position on the toolbar and is separated from the geometric constraint tools.
- _Sketcher_ConstrainCoincidentUnified_ has been established next. Despite merging two functions which result in different impacts to degrees of freedom, these are expected defaults with a more naturalistic line of thinking which is an expectation of attachment of one point to another sketch element (point to point, arc, line, bspline)
- _Sketcher_ConstrainCoincidentUnified_ has a modified icon in order to disambiguate from the discreet tools it is intended to replace.
- _Sketcher_ConstrainTangent_ placed next (another form of attachment, unique from coincident)
- _Sketcher_ConstrainHorVer_ new icon to better indicate what this tool does.
[...]

PR #11745, opened 2023-12-16
Gui: Allow ActionGroups to not remember choice

kadet1090:
This is fix to issue mentioned in the #11717, on discord and forum that for smart dimension tool the chosen tool should not be remembered. This will ensure that the "smart" tool is always visible on the toolbar and other tools are accessible in case that such explicit choice is needed.

It does not have issue attached as I simply wanted to check how hard it would be to fix - quite easy as it seems.

https://github.com/FreeCAD/FreeCAD/assets/747404/684d84bb-60d0-4c6e-9685-89cfbd8b953e

I changed only the behavior of smart dimension tool, everything else should be considered as part of other PRs.

### DIFF
--- a/src/Gui/Action.cpp
+++ b/src/Gui/Action.cpp
@@ -430,11 +430,12 @@ void Action::setMenuRole(QAction::MenuRole menuRole)
  * Constructs an action called \a name with parent \a parent. It also stores a pointer
  * to the command object.
  */
-ActionGroup::ActionGroup ( Command* pcCmd,QObject * parent)
-  : Action(pcCmd, parent)
-  , _group(nullptr)
-  , _dropDown(false)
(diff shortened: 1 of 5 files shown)


PaddleStroke:
Code looks good.
And this is a nice addition for sketcher_dimension tool.

obelisk79:
Definitely a much needed setting as toolbars grow increasingly complex.

yorikvanhavre:
Most of the PR here is useful. There *might* be more discussion about if this change is needed/wanted for the sketcher tool ,but I propose we merge this and keep an eye discussions

maxwxyz:
@kadet1090 could you also implement this for the auto horizontal / vertical *Constraint_HorVer* tool?
grafik

Issue #11446, opened 2023-11-19
[Problem] Regression: Spacemouse does not work on 0.21.1 with any recent Spacemouse Driver


strangeluck:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

I recently upgrade both my Spacemouse driver and Freecad and discovered my Spacemouse didn't work in Freecad any longer. I experimented with Freecad/Spacemouse and the results are below

|Spacemouse |FreeCAD 0.20.2 |FreeCAD 0.21.1 |
|------------------ | ---------------------- |---------------------- |
|10.7.0|                            YES|                        NO|
|10.7.4|                             NO|                        NO|
|10.8.2|                             NO|                        NO|

Spacemouse is amazing and I hope support never dies. Hope this gets resolved. Sticking with what works... 
Love your software!

### Full version info

Does work:
```shell
OS: macOS 10.16
Word size of FreeCAD: 64-bit
Version: 0.20.2.29603 (Git)
Build type: Release
Branch: (HEAD detached at 0.20.2)
Hash: 930dd9a76203a3260b1e6256c70c1c3cad8c5cb8
Python 3.10.8, Qt 5.15.4, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: C/Default (C)
```
Doesn't work:
```
OS: macOS 13.3.1
Word size of FreeCAD: 64-bit
Version: 0.21.1.33694 (Git)
Build type: Release
Branch: (HEAD detached at 0.21.1)
Hash: f6708547a9bb3f71a4aaade12109f511a72c207c
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: C/Default (C)
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Syres916:
See https://github.com/FreeCAD/FreeCAD/issues/10926#issuecomment-1762203236 it seems relevant, it would seem that 0.21.2 should be packaged with the fix

numo68:
#11753 restores the functionality with 0.8.2

luzpaz:
@numo68 in other words, this ticket can be closed ?

numo68:
> in other words, this ticket can be closed ?

@luzpaz The PR was merged to main - if that is enough to close the ticket, then yes. The fix is not present in 0.21 (at least not yet; the backport label is there).

[...]

PR #11753, opened 2023-12-17
fix compatibility with current Mac 3Dconnexion driver

numo68:
For quite some time the MacOS version of FreeCAD detects the SpaceMouse from 3DConnexion but is unable to receive any events when using the recent drivers. The last working driver seems to be the 10.7.0.

This PR changes the registration of the application to the driver to use the wildcard constant instead of a bundle ID that the FreeCAD does not seem to possess. The driver then uses the application name to do the check.

Tested on MacBook Air M2 15" running Sonoma 14.2, with SpaceMouse Pro and driver 10.8.2.

### DIFF
--- a/src/Gui/3Dconnexion/GuiNativeEventMac.cpp
+++ b/src/Gui/3Dconnexion/GuiNativeEventMac.cpp
@@ -164,9 +164,7 @@ void Gui::GuiNativeEvent::initSpaceball(QMainWindow *window)
     /* register our app with the driver */
     // Pascal string Application name required to register driver for application
     UInt8  tdxAppName[] = {7,'F','r','e','e','C','A','D'};
-    // 32bit appID to register driver for application
-    UInt32 tdxAppID = 'FCAd';
-    tdxClientID = RegisterConnexionClient( tdxAppID, tdxAppName,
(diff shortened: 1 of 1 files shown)


oursland:
I can confirm that this corrects the issue for me on macOS Sonoma 14.2 with 3Dconnextion 10.8.2 (current latest).

luzpaz:
backport ?

numo68:
The patch applies to 0.21 unchanged and I have tested it there as well. I don't know how the backporting process works; if I need to create a separate PR to releases/FreeCAD-0-21, I will, if the maintainers drive that, please do. Thanks

luzpaz:
@numo68 an aside, would you be up to weighing in on any other [open 3DConnexion related tickets](https://github.com/FreeCAD/FreeCAD/labels/Hardware%3A%203dconnexion)?

numo68:
> an aside, would you be up to weighing in on any other [open 3DConnexion related tickets]

I will take a look but I am by far not a 3D devices expert, I was just solving a problem I have encountered as I switched platforms, so no promises.

luzpaz:
Note: when this gets backported please reference #11446 along with this ticket. TY

PR #11755, opened 2023-12-17
Remove code related to qtxmlpatterns

Neumann-A:
Since QDomNodeModel is the only code requiring qtxmlpatterns and it seems to be not referenced anywhere I simply removed it. 

What was not changed:
- LibPack setting a variable related to qtxmlpatterns
- Generated translations.


tools/build/WindowsInstaller/Delete.bat
```diff
@@ -88,7 +88,6 @@ del Qt5WebViewd.dll
 del Qt5Widgetsd.dll
 del Qt5WinExtrasd.dll
 del Qt5Xmld.dll
-del Qt5XmlPatternsd.dll
```
adrianinsaval, line 91: this should probably kept, removing this line means the installer would include the debug dll for qt5xmlpattern, which would be useless

tools/build/WindowsInstaller/Delete.bat
```diff
@@ -88,7 +88,6 @@ del Qt5WebViewd.dll
 del Qt5Widgetsd.dll
 del Qt5WinExtrasd.dll
 del Qt5Xmld.dll
```
Neumann-A, line 90:
```suggestion
del Qt5Xmld.dll
del Qt5XmlPatternsd.dll
```

WandererFan (approved):
LGTM, but I don't know much about packaging.

Werner removed the xmlpatterns code a while back, so this should be fine.

Issue #11690, opened 2023-12-08
[Problem] Version 0.21.1 Sketcher Angle Constraint 


bigdipperaviation:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

When system units are set on "Imperial for Civil Engineering"; angle constraints behave oddly. When you set an angle constraint, the angle is graphically displayed on the angle between two lines; however, the value in the data entry box is "0.0000". See attached screenshot. When units are switched to "Imperial Decimal", there is not issue. The angle constraint does not behave this way in the copy version 0.20.1 that I have. 

Version 0.21.1 Not Functioning
![Freecad0 21 1AngleIssue001](https://github.com/FreeCAD/FreeCAD/assets/104542361/2f6fb480-d4fa-445a-a730-b3f8ee89cc8f)

Verision 20.1 Functioning Correctly
![image](https://github.com/FreeCAD/FreeCAD/assets/104542361/9351f643-4282-441e-86c5-5fd45dd80a02)

Forgive me if there is something simple code-wise that I could correct to resolve this issue. I am not nearly as proficient as some of you experts on here.

### Full version info

```shell
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.21.1.33668 +26 (Git)
Build type: Release
Branch: (HEAD detached at 0.21.1)
Hash: f6708547a9bb3f71a4aaade12109f511a72c207c
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * A2plus 0.4.60n
  * ExplodedAssembly
  * fasteners 0.4.71
```

### Subproject(s) affected?

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

xtemp09:
It is present on Linux too.


```ruby
OS: Ubuntu 22.04.3 LTS (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35327 (Git)
Build type: Debug
Branch: main
Hash: 32a0e0c6ae7fdd6e179065c247d53b8497bf1db2
Python 3.10.12, Qt 5.15.3, Coin 4.0.0, Vtk 9.1.0, OCC 7.5.1
Locale: English/United States (en_US)
```

adrianinsaval:
[...]

Issue #9552, opened 2023-05-12
[Problem] US-Building units system is inconsistent with FreeCAD and does not match standard notation


ConnorWorrell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: Arch Linux (sway)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33162 (Git)
Build type: Unknown
Branch: master
Hash: 86128f1de2a7ee3830edc0b083b4b376bb68b7df
Python 3.10.10, Qt 5.15.9, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * dodo
[/code]
```

### Subproject(s) affected?

Other (specify in description)

### Problem description

US-Building units system is inconsistent with FreeCAD and does not match standard notation.

This is a compilation from the discussion here: #9408

The following unambiguous input formats should be available in spinbox and the expression parser:

```
#'# #/#"
#'#-#/#"
#'#" #9397 
#' #/#"
#' (Passes)
\# #/#"
#" (Passes)
#/#"
```

The following ambiguous formats are not available:
```
\#'-#"
#-#/#"
```
These should be treated as tokens, negation should apply to the token as a whole:
ex -1' 1/2" evaluates to -12.5"

Editing a dimension should present in the format:
```
#' # #/#"
-#' # #/#"
```
Dimensions displayed should be in one of the following formats (Sketcher, TechDraw, Arch, Spinbox):
```
#' - # #/#" (Display 0 in the inches if applicable, ex #' - 0 #/#")
#' #"
#'
\# #/#"
#"
#/#"
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
Does this need to be re-opened? (since it may have been closed prematurely by the bot)

PR #11757, opened 2023-12-18
[App/Gui] Us building distance and imperial civil engineering angle format added to expression parser.

ConnorWorrell:
This aims to solve these 2 issues:

* Expression parser not accepting the imperial civil engineering angle format https://github.com/FreeCAD/FreeCAD/issues/11690
* US Building distance format https://github.com/FreeCAD/FreeCAD/issues/9552

It adds support for the following units:

- #deg#M#AS
- #deg#AS
- #ft # #/#in
- #ft #/#in
- #ft #in (replacing this temporary solution https://github.com/FreeCAD/FreeCAD/pull/9397)
- \# #/#in

I've broken it up into 2 commits, the first contains all the expression parser updates, and the second updates the GUI to make use of these formats.

Its currently failing this unit test, I'm not sure why its failing besides the obvious, but I'm also not sure what is is testing besides the obvious. As far as I can tell App::ExpressionTokenizer().perform() isn't used anywhere.
https://github.com/FreeCAD/FreeCAD/blob/625a4212cc75b4fa7da6eb15f6538942290f8364/tests/src/App/Expression.cpp#L15


adrianinsaval:
@wwmayer @realthunder could one of you review/advise about this PR and the failing test?

chennes:
@wwmayer do you have a minute to look at this failing test? I'm not familiar enough to figure out what's going on.

wwmayer:
> The logs for this run have expired and are no longer available.

wwmayer:
> As far as I can tell App::ExpressionTokenizer().perform() isn't used anywhere.

ExpressionTokenizer::perform() is used in ExpressionCompleter::slotUpdate. In the past its code was bundled with ExpressionCompleter and has been refactored with https://github.com/FreeCAD/FreeCAD/pull/8635 to make it testable.

AFAICT the behaviour of ExpressionTokenizer is to return the last word before the cursor position. Numerical constants like e or pi will be ignored. So, if the input text is "0.00000 deg" and cursor position 11 then the last word before was "deg".

Now I can't tell what is the exact change that makes the tests to fail but I guess the behaviour of ExpressionParser::tokenize has changed.

I think it's best to write more unit tests to verify the output of ExpressionParser::tokenize, first. This function accepts any input and returns a list of triples: (token type, start position, text)

For example for the input "0.00000 deg" it returns (260, 0, "0.00000"), (262, 8, "deg"). The token type 260 means NUM (like number) and 262 means UNIT.

chennes:
@ConnorWorrell we're moving this back to draft pending a fix of the unit tests.

maxwxyz:
@chennes is there already a fix for the unit tests so @ConnorWorrell could mark this as ready for review (if it should be considered for the 1.0 feature freeze)?

maxwxyz:
@ConnorWorrell will you pick up this PR again?

PR #11758, opened 2023-12-18
CI: set pipefail Linux build action

oursland:
Errors in the Ubuntu builds are not caught because the exit code of `tee` is observed instead of the exit code of the build.

Setting the shell option `set -o pipefail` will propagate the exit code of any process in a pipeline that is not successful, permitting the CI script to get an accurate assessment of the build status.

### DIFF
--- a/.github/workflows/actions/linux/build/action.yml
+++ b/.github/workflows/actions/linux/build/action.yml
@@ -48,6 +48,7 @@ runs:
       id: build
       shell: bash -l {0}
       run: |
+        set -o pipefail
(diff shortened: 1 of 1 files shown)


adrianinsaval:
thanks for the quick fix!

adrianinsaval:
should we make a similar change to the testing steps? the test clearly failed on https://github.com/FreeCAD/FreeCAD/actions/runs/7178556454/job/19546990047 since the build wasn't complete but they also showed up as successful

0penBrain:
I wonder if this change was needed.
IMO the original problem may only be the line split.
I think removing the backslash and having all the stdbuf command as a (long) one-liner may solve the issue.

PR #11767, opened 2023-12-18
CI: add pipefail to additional workflow actions.

oursland:
As noted in [this comment](https://github.com/FreeCAD/FreeCAD/pull/11758#issuecomment-1859788651), additional actions should fail if any sub-process fails.

This commit adds `pipefail` option to:

* .github/workflows/actions/linux/configure/action.yml
* .github/workflows/actions/linux/install/action.yml
* .github/workflows/actions/runCPPTests/runSingleTest/action.yml
* .github/workflows/actions/runPythonTests/action.yml

### DIFF
--- a/.github/workflows/actions/linux/configure/action.yml
+++ b/.github/workflows/actions/linux/configure/action.yml
@@ -52,6 +52,7 @@ runs:
       id: configure
       shell: bash -l {0}
       run: |
+        set -o pipefail
         (stdbuf -oL -eL cmake -S ${{ inputs.sourcedir }} -B ${{ inputs.builddir }} -D CMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE ${{inputs.extraParameters }}) \
         2> >(tee -a ${{ inputs.errorFile }}) | tee -a ${{ inputs.logFile }}
     - name: Write report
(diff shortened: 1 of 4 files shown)


oursland:
@adrianinsaval The CI error for macOS is a transient bug, possibly specific to that build node, about a library not being found (`dyld[6338]: Library not loaded: @rpath/libsolv.1.dylib
`). 

Here's the log on my fork, where it is successful: https://github.com/oursland/FreeCAD/actions/runs/7251715643

adrianinsaval:
What could be the cause of these weird Mac errors? A couple of times I've seen it fail on the translation stuff on build, complaining about a missing library, but then rerunning it it succeeds

oursland:
It appears to be an issue with failing to find `libsolv.1.dylib`, which I believe is the `mamba` solver.  The most likely culprit is when performing a re-install of the `libsolv` package, another process also attempts to use it.

This could be a known and resolved issue as `mamba` version `1.4.9` is not the latest version, however migrating to a newer version of `mamba` is not possible at this time due to shared dependencies with `occt` which is pinned at `7.6.3`.

By default `conda devenv` and `mamba-devenv` will prune a development environment to only contain the listed packages and their dependencies.  It may be that this is also resulting in forcing the reinstall of base packages, such as `mamba`.  I'll look into seeing if passing `--no-prune` can be used to prevent this from occurring and possibly affecting CI.

PR #11768, opened 2023-12-19
CI: Fix issues affecting conda devenv installation resulting in failures.

oursland:
During the installation of the devenv, some packages are re-installed.  As this may take place in parallel with the installation of other packages, re-installation of the base packages may break the other packages due to missing core libraries resulting in a failure.

This PR correctly sets the Python to 3.11.x, preventing an unnecessary upgrade, and also sets the `--no-prune` option to `mamba-devenv`, preventing the re-installation of the core packages.

### DIFF
--- a/conda/conda-env.yaml
+++ b/conda/conda-env.yaml
@@ -3,6 +3,6 @@ channels:
 - conda-forge
 dependencies:
 - conda-devenv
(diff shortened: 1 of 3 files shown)


oursland:
@adrianinsaval this PR likely addresses the occasional CI failure during the establishment of the conda environment.

PR #11775, opened 2023-12-19
Fix occasional segmentation fault when exiting

### LINKED ISSUE #11521: [Problem] Path: Crash after quitting FreeCAD after using ToolBit Library Editor

### Problem description

Trying to follow the new Path tutorial by @concretedog posted in the blog: https://blog.freecad.org and experienced FreeCAD crashes approximately 70% of the time after saving file and quitting FreeCAD. Seems to only happen after spending time in the ToolBits Library.  This on macOS M1/arm.

Test file attached, but not sure it is related to the file.

### Full version info
[Cylinder4Path.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/13462871/Cylinder4Path.FCStd.zip)

```
OS: macOS 14.1.1
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35081 (Git)
Build type: Release
Branch: main
Hash: 32db044caa92a2adc448850179c524b793a03c3e
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.6, OCC 7.6.3
Locale: C/Default (C)
```

### Subproject(s) affected?

Path

### Anything else?

Crash report -- I capture 3 that were almost identical, other than addresses:

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               freecad [23134]
Path:                  /Applications/Other Apps/FreeCAD-35081.app/Contents/Resources/bin/freecad
Identifier:            freecad
Version:               ???
Code Type:             ARM-64 (Native)
Parent Process:        Exited process [23128]
User ID:               501

Date/Time:             2023-11-24 19:39:48.7930 -0600
OS Version:            macOS 14.1.1 (23B81)
Report Version:        12
Anonymous UUID:        E1A9B3E8-17C0-3635-633F-31D076EC9805

Sleep/Wake UUID:       499488E4-0E4A-4A4E-85C8-3D38854F7199

Time Awake Since Boot: 250000 seconds
Time Since Wake:       40390 seconds

System Integrity Protection: enabled

Crashed Thread:        0  CrBrowserMain  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000f41fd9c969a8 -> 0x0000741fd9c969a8 (possible pointer authentication failure)
Exception Codes:       0x0000000000000001, 0x0000f41fd9c969a8

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [23134]

VM Region Info: 0x741fd9c969a8 is not in any region.  Bytes after previous region: 22126493526441  
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      MALLOC_NANO              600000000000-600020000000 [512.0M] rw-/rwx SM=PRV  
--->  
      UNUSED SPACE AT END

Thread 0 Crashed:: CrBrowserMain Dispatch queue: com.apple.main-thread
0   libFreeCADGui.dylib           	       0x10629cc5c std::__1::__hash_iterator<std::__1::__hash_node<std::__1::__hash_value_type<App::DocumentObject*, std::__1::set<App::DocumentObject*, std::__1::less<App::DocumentObject*>, std::__1::allocator<App::DocumentObject*>>>, void*>*> std::__1::__hash_table<std::__1::__hash_value_type<App::DocumentObject*, std::__1::set<App::DocumentObject*, std::__1::less<App::DocumentObject*>, std::__1::allocator<App::DocumentObject*>>>, std::__1::__unordered_map_hasher<App::DocumentObject*, std::__1::__hash_value_type<App::DocumentObject*, std::__1::set<App::DocumentObject*, std::__1::less<App::DocumentObject*>, std::__1::allocator<App::DocumentObject*>>>, std::__1::hash<App::DocumentObject*>, std::__1::equal_to<App::DocumentObject*>, true>, std::__1::__unordered_map_equal<App::DocumentObject*, std::__1::__hash_value_type<App::DocumentObject*, std::__1::set<App::DocumentObject*, std::__1::less<App::DocumentObject*>, std::__1::allocator<App::DocumentObject*>>>, std::__1::equal_to<App::DocumentObject*>, std::__1::hash<App::DocumentObject*>, true>, std::__1::allocator<std::__1::__hash_value_type<App::DocumentObject*, std::__1::set<App::DocumentObject*, std::__1::less<App::DocumentObject*>, std::__1::allocator<App::DocumentObject*>>>>>::find<App::DocumentObject*>(App::DocumentObject* const&) + 180
1   libFreeCADGui.dylib           	       0x1067b4768 Gui::WrapperManager::destroyed(QObject*) + 64
2   libQt5Core.5.15.8.dylib       	       0x1088e683c 0x1087a0000 + 1337404
3   libFreeCADGui.dylib           	       0x10628bc94 Py::Object::~Object() + 40
4   libFreeCADGui.dylib           	       0x1067b442c Gui::WrapperManager::clear() + 212
5   libQt5Core.5.15.8.dylib       	       0x1088e683c 0x1087a0000 + 1337404
6   freecad                       	       0x10495f030 main + 4932
7   dyld                          	       0x1827990e0 start + 2360
```

### Code of Conduct

xtemp09:
Closes #11521.

Since the backtrace given in the issue #11521 points out at crash because of race condition, first, I compiled FreeCAD with thread sanitizer. _TSan_ interceptors revealed concurrent access to `std::unordered_map<QObject*, std::list<Py::Object>> wrappers;`. I rewrote _WrapperManager_ so it could work asynchronously.

Another problem emerged after ― _ToolBitAttributes_' Python wrapper is destroyed, causing segmentation fault when accessing it in the destroyed function. It can be solved with addition of `Py_INCREF` and `Py_DECREF`.

So, the possible causes are:

1. Concurrent access to the `std::unordered_map`
2. refcount for the wrapper reached zero
3. A PyObject has been invalidated more than once

The first one is eliminated by the reimplementation, the second disappears with `Py_INCREF` and `Py_DECREF`, the third disappear with comparison of _pyobj_ with _nullptr_. The third and the second option can be caused by the first. I have made sure that the three options cannot crash FreeCAD.

The cause of this is an error occurred somewhere else. This pull request makes sure FreeCAD does not segfaults when encountering these cases.

----

Some notes for posterity.

If `Py_INCREF(pyobj);` is absent along with `Py_DECREF(pyobj);`, FreeCAD crashes because of _ToolBitAttributes_ during reproduction of the certain actions described in the issue #11521.

If `Py_INCREF(pyobj);` is present, while `Py_DECREF(pyobj);` is not, FreeCAD works fine. However, compiled with leak sanitizer, _LSan_ interceptors crash FreeCAD during reproduction of the actions described in the issue. I have observed this before ― compiled with _LSan_, FreeCAD occasionally crashed when I was doing seemingly harmless actions. Perhaps, `Py_DECREF` is missing somewhere.

### DIFF
--- a/src/Gui/PythonWrapper.cpp
+++ b/src/Gui/PythonWrapper.cpp
@@ -263,7 +263,7 @@ PyTypeObject *getPyTypeObjectForTypeName();
  */
 class WrapperManager : public QObject
 {
-    std::unordered_map<QObject*, std::list<Py::Object>> wrappers;
+
 
 public:
     static WrapperManager& instance()
@@ -275,52 +275,38 @@ public:
      * \brief addQObject
      * \param obj
      * \param pyobj
-     * Add the QObject and its Python wrapper to the list.
+     * Connects destruction event of a QObject with invalidation of its PythonWrapper via a helper QObject.
      */
     void addQObject(QObject* obj, PyObject* pyobj)
     {
-        if (wrappers.find(obj) == wrappers.end()) {
-            QObject::connect(obj, &QObject::destroyed, this, &WrapperManager::destroyed);
-        }
+        const auto PyW_unique_name = QString::number(reinterpret_cast <quintptr> (pyobj));
+        auto PyW_invalidator = findChild <QObject *> (PyW_unique_name, Qt::FindDirectChildrenOnly);
 
-        auto& pylist = wrappers[obj];
-        if (std::find_if(pylist.cbegin(), pylist.cend(),
-                [pyobj](const Py::Object& py) {
-                    return py.ptr() == pyobj;
-                }) == pylist.end()) {
+        if (PyW_invalidator == nullptr) {
+            PyW_invalidator = new QObject(this);
+            PyW_invalidator->setObjectName(PyW_unique_name);
 
-            pylist.emplace_back(pyobj);
-        }
-    }
+            Py_INCREF (pyobj);
+        } else
+            PyW_invalidator->disconnect();
+
+        auto destroyedFun = [pyobj](){
+#if defined (HAVE_SHIBOKEN)
+            auto sbk_ptr = reinterpret_cast <SbkObject *> (pyobj);
+            if (sbk_ptr != nullptr) {
+                Base::PyGILStateLocker lock;
+                Shiboken::Object::setValidCpp(sbk_ptr, false);
+            } else
+                Base::Console().DeveloperError("WrapperManager", "A QObject has just been destroyed after its Pythonic wrapper.\n");
+#endif
+            Py_DECREF (pyobj);
+        };
+
+        QObject::connect(PyW_invalidator, &QObject::destroyed, this, destroyedFun);
+        QObject::connect(obj, &QObject::destroyed, PyW_invalidator, &QObject::deleteLater);
+}
 
 private:
-    /*!
-     * \brief destroyed
-     * \param obj
-     * The listed QObject is about to be destroyed. Invalidate its Python wrappers now.
-     */
-    void destroyed(QObject* obj = nullptr)
-    {
-        if (obj) {
-#if defined (HAVE_SHIBOKEN) && defined(HAVE_PYSIDE)
-            auto key = wrappers.find(obj);
-            if (key != wrappers.end()) {
-                Base::PyGILStateLocker lock;
-                for (const auto& it : key->second) {
-                    auto value = it.ptr();
-                    Shiboken::Object::setValidCpp(reinterpret_cast<SbkObject*>(value), false);
-                }
-
-                wrappers.erase(key);
-            }
-#endif
-        }
-    }
-    void clear()
-    {
-        Base::PyGILStateLocker lock;
-        wrappers.clear();
-    }
     void wrapQApplication()
     {
         // We have to explicitly hold a reference to the wrapper of the QApplication
@@ -347,8 +333,6 @@ private:
 
     WrapperManager()
     {
-        connect(QApplication::instance(), &QCoreApplication::aboutToQuit,
-                this, &WrapperManager::clear);
         wrapQApplication();
     }
     ~WrapperManager() override = default;


xtemp09:
This commit also fixes the problem: xtemp09/FreeCAD@c3ac0059099aad77c87694e3caf3d8d7d858c4b9


I've tested it with _LSan_ as well.

yorikvanhavre:
@wwmayer could you have a look if this looks OK to you?

marioalexis84:
I get a segmentation fault with this changes in several situation when close a task panel:
Example to reproduce:
Activate Fem workbench (or Arch workbench)
Add a Material.
Close the task panel --> crash 

In those cases, in the lambda `destroyedFun` the captured  `PyObject* pyobj` already has reference count equal to 1, the call to `Py_DECREF` triggers deallocation of the object and this causes the crash.

xtemp09:
@marioalexis84, could you check if replacement of `Py_DECREF` with `Py_XDECREF` solves it?

marioalexis84:
> check if replacement of Py_DECREF with Py_XDECREF solves it?

No it doesn't solve it. `Py_XDECREF` call `Py_DECREF` after checking that pointer to the object is not null. So at that point both functions are equivalent.

I check your [correction](https://github.com/FreeCAD/FreeCAD/pull/12082) and now there is no crash. The reference count is equals zero at the end of `destroyFun`, so there is no memory leak.
There are some exceptions to this in the Draft module: For example, if you create a line, `pyobj` has reference count equal to one at the end of `destroyFun`, but that seems to be the Draft code increasing the reference somewhere.

It would still be nice to have a review of this changes by @wwmayer  as soon as he is available.

xtemp09:
Thank you for checking. _Draft_ is known to leak.

PaddleStroke:
@xtemp09 I think this is causing more issues : 
- In my assembly card 9 PR, clicking OK/cancel in the task panels crash.
- Clicking OK is crashing int the preference page : https://forum.freecad.org/viewtopic.php?t=84624

xtemp09:
@PaddleStroke, could you check if it crashes with #12082?

PaddleStroke:
I am currently building it and somehow these commits are triggering almost full rebuilds so I don't know yet.
I have seen your fix after posting actually. I'll feedback soon.

PaddleStroke:
I confirm that your fix fixed them

xtemp09:
Thank you. Can maintainers restart the script to make weekly builds? I feel somewhat uncomfortable that the weekly builds crash every time user presses a button.

chennes:
Yes, we'll rebuild.

PR #11776, opened 2023-12-19
Adjust the fuse fuzzy value in FeaturePad to accept close objects

### LINKED ISSUE #9377: [Problem] PartDesign: Helix insufficient precision

bgbsww:
Approach to fix #9377 that acknowledges the error seems to lie more with the fuse in the pad than the generated helix.

### DIFF
--- a/src/Mod/PartDesign/App/FeaturePad.cpp
+++ b/src/Mod/PartDesign/App/FeaturePad.cpp
@@ -216,8 +216,21 @@ App::DocumentObjectExecReturn *Pad::execute()
         this->AddSubShape.setValue(prism);
 
         if (!base.IsNull()) {
-            // Let's call algorithm computing a fuse operation:
-            BRepAlgoAPI_Fuse mkFuse(base, prism);
+            // We need to set the FuzzyValue before the computation, so use parameterless constructor
+            BRepAlgoAPI_Fuse mkFuse;
+            TopTools_ListOfShape shapeArguments,shapeTools;
+            shapeArguments.Append(base);
+            shapeTools.Append(prism);
+            mkFuse.SetArguments(shapeArguments);
+            mkFuse.SetTools(shapeTools);
+            // Fuse defaults to Precision::Confusion() or 1e-07.  It is possible to create legitimate shapes
+            // (like FeatureHelix) that are well within this precision at 2e-12 but still result in this Fuse
+            // generating multiple adjacent solids that are closer than the precision setting.  Reducing the
+            // precision by two orders of magnitude stops this misbehavior.  No other setting, including the
+            // Glue in use, CheckInverted or attempting a SimplyResult will affect the fuse result.
+            mkFuse.SetFuzzyValue(Precision::Confusion()*100);
+            mkFuse.Build();
+
             // Let's check if the fusion has been successful
             if (!mkFuse.IsDone())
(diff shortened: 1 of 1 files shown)


src/Mod/PartDesign/App/FeaturePad.cpp
```diff
@@ -216,8 +216,21 @@ App::DocumentObjectExecReturn *Pad::execute()
...
+            shapeArguments.Append(base);
+            shapeTools.Append(prism);
+            mkFuse.SetArguments(shapeArguments);
+            mkFuse.SetTools(shapeTools);
+            // Fuse defaults to Precision::Confusion() or 1e-07.  It is possible to create legitimate shapes
+            // (like FeatureHelix) that are well within this precision at 2e-12 but still result in this Fuse
+            // generating multiple adjacent solids that are closer than the precision setting.  Reducing the
+            // precision by two orders of magnitude stops this misbehavior.  No other setting, including the
+            // Glue in use, CheckInverted or attempting a SimplyResult will affect the fuse result.
+            mkFuse.SetFuzzyValue(Precision::Confusion()*100);
```
adrianinsaval, line 231: is there any danger in this? do we expect possible negative side effects here?

adrianinsaval: I'll ask blobfish if he can give me any insight on this

bgbsww: Excellent, opinions would be helpful.  Since we're at 1e-05 here, I can't see any likely negatives, but maybe someone else can.  The failure would be that we would take shapes that are within 1e-05 of each other but resulted in no fuse - and thus a broken result - and start having them work.  If there's something out there that depends on this behavior, then this will have a negative side effect.  I applaud the caution here.

adrianinsaval: blobfish told me this:

> Full disclosure, I have never used the fuzzy value in booleans, so I am clueless to the pros and cons. I do know tolerances are a big deal and can open up a can of worms. I guess I would nervous about making an unilateral change. Maybe that value should be exposed via a feature parameter or a global setting, until more intuition has accumulated? Just thinking out loud.

I had a similar train thought, I it would be a good idea to make this an adjustable property, we don't know the ramifications of this yet, we could default to this value to find out and if someone has trouble they can adjust the property. If things go wrong too often then we change the default to `Precision::Confusion` and the value can be adjusted just for problematic cases like in the issue. The property could be hidden for now.

maxwxyz:
@bgbsww also here the heads up:

bgbsww:
So, the code in question here goes away with TNP on, so certainly this PR in current form will be irrelevant.  A test with TNP on shows the bug doesn't seem to exist after a recompute.  So, I'm going to close this bugfix.

PR #11779, opened 2023-12-19
[Base/Units] Change strings' description

hasecilu:
Currently in the string description for unit system "degree" (unit name) appears instead of "°" (symbol). In the same entries appear "mm", "kg", "s" opposed to "millimeter", "kilogram", "second". So I think could be a better fit to use the symbol for the unit instead.

As an example SOLIDWORKS uses the full name of the units instead of only the symbols.

Usually the use of "s" instead of "sec" is preferred for time unit second.

Affects strings presented to the user on the combobox to select unit system.

- The symbol to represent angles on "degree" unit is '°'
- Change imperial velocity to "ft/s" to enforce the use of 's' symbol as second

Also added colons to TODOs to get highlightning.

### DIFF
--- a/src/Base/QuantityLexer.c
+++ b/src/Base/QuantityLexer.c
@@ -2149,7 +2149,7 @@ static int yy_get_next_buffer (void)
     void yy_switch_to_buffer  (YY_BUFFER_STATE  new_buffer )
 {
 
-	/* TODO. We should be able to replace this entire function body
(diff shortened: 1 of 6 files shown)


MisterMakerNL:
I think it should be °C if we talk about Celsius. There are a lot of temperature units with the ° symbole. https://en.wikipedia.org/wiki/Degree_(temperature)

PR #11749, opened 2023-12-17
Sketcher: adjust cursor icons to match toolbar icons

maxwxyz:
fixes #11352
adjusted the cursor / pointer icons to match the toolbar icons
- polyline was still showing the old icon
- fixed fillet icon (was a chamfer icon)
- fixed BSpline icon
- fixed trim icon
- continuous design of cursor icons in alignment with the toolbar icons
- added icons for rectangle options

Issue #11482, opened 2023-11-21
[Problem] Diffucult to Discern Positive/Negative Space in Sketcher


koopdi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

When creating nested islands of closed geometry in sketcher, it can be difficult to tell apart the negative and positive space.
In the image below, it is not easy to tell which of the small circles will be filled (left).
![pos_neg_space](https://github.com/FreeCAD/FreeCAD/assets/12898413/4daa65dc-c24a-40bd-a1de-f52de0f48882)
Issue: #11472 is related but distinct. It is possible that a common solution could be found.

### Full version info

```shell
[code]
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.21.1.33668 +26 (Git)
Build type: Release
Branch: (HEAD detached at 0.21.1)
Hash: f6708547a9bb3f71a4aaade12109f511a72c207c
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * 3DfindIT 1.2.0
  * A2plus 0.4.60n
  * Assembly4 0.50.4
  * Curves 0.6.14
  * Defeaturing 1.2.1
  * DynamicData 2.53.0
  * ExplodedAssembly
  * fasteners 0.4.70
  * freecad.gears 1.0.0
  * Help 1.0.3
  * Manipulator 1.5.0
  * Nodes 0.1.36
[/code]
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

maxwxyz:
A toggle or setting would be great to enable a slightly transparent overlay of positive closed loops.

pierreporte:
This is a great feature. It already exists in some CAD programs. However it is going to need #7334 resolved because the new color will need to go well with construction lines and constraints.

In my opinion, the fill color should be very close to the background color, just opaque enough to be distinguished and understood, but discreet enough to allow working comfortably with the sketch.

maxwxyz:
Additionally a slightly colored overlay would be great for the islands:

- islands not fully constrained: default color
- island defined by fully constrained geometry: green
- islands defined by only construction lines: no color or blue

I guess overall the islands or spaces should only be drawn between normal sketch geometry, not construction lines, internal geometry or external references.

[...]

PR #11785, opened 2023-12-20
[UI] Sketcher: icon overhaul for the sketcher workbench to unify the appearance

maxwxyz:
*This PR is only changing the .svg icons, no other WB code is touched.*

The sketcher toolbar icons currently have different stroke widths, colors and point sizes. This PR unifies the appearance of the icons in the Sketcher WB. The icons were adjusted or redrawn to match the FreeCAD design guidelines and progress was discussed in the Design Working Group. The icons for the constraint group were made by @obelisk79 

Here are a few screenshots before and after, **GitHub seems to compress the copied  screenshots**. 
If otherwise not specified, the theme in the screenshots is OpenDark. Please ignore the dropdown arrows on the toolbar icons.

There is another PR to change the cursor icons as well, to match the toolbar icons: https://github.com/FreeCAD/FreeCAD/pull/11749

# Toolbar icons 24x24 and 32x32:
image
image

# Dropdowns, right mouse menu and menu items plus elements:
image
image
image
image

# 24x24 px icons and constraints "in action":
image

# Toolbar icons in FreeCAD Classic theme:
image

and for comparison the inconsistency of the current icons...
# Current toolbars and menu items
image
image

### DIFF
--- a/src/Mod/Sketcher/Gui/Resources/Sketcher.qrc
+++ b/src/Mod/Sketcher/Gui/Resources/Sketcher.qrc
@@ -6,6 +6,7 @@
     </qresource>
     <qresource>
         <file>icons/constraints/Constraint_Block.svg</file>
+        <file>icons/constraints/Constraint_Coincident.svg</file>
         <file>icons/constraints/Constraint_Concentric.svg</file>
         <file>icons/constraints/Constraint_Dimension.svg</file>
         <file>icons/constraints/Constraint_Dimension_Driven.svg</file>
--- a/src/Mod/Sketcher/Gui/Resources/icons/constraints/Constraint_Block.svg
+++ b/src/Mod/Sketcher/Gui/Resources/icons/constraints/Constraint_Block.svg
@@ -1,19 +1,74 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg version="1.1" id="svg2816" height="64px" width="64px" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
-  <defs id="defs26" />
-  <g id="pale-red" style="fill:none;stroke:#280000;stroke-width:8">
-    <path d="M 48,48 L 16,16" id="path2" />
-    <path d="M 16,48 48,16" id="path4" />
-    <circle r="26" cy="32" cx="32" id="circle6" />
+<svg
+   version="1.1"
+   id="svg2816"
+   height="64px"
+   width="64px"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs26" />
+  <g
+     id="pale-red"
+     style="fill:none;stroke:#280000;stroke-width:13.0001"
+     transform="matrix(0.61538495,0,0,0.61538495,10.309556,10.309556)">
+    <path
+       d="M 17.15836,17.158356 52.874794,52.87479"
+       id="path2"
+       style="stroke-width:11.375;stroke-dasharray:none" />
+    <path
+       d="M 52.874794,17.158356 17.15836,52.87479"
+       id="path4"
+       style="stroke-width:11.375;stroke-dasharray:none" />
+    <circle
+       r="29.249981"
+       cy="-35.246952"
+       cx="-35.246952"
+       id="circle6"
+       style="stroke-width:13;stroke-dasharray:none"
+       transform="scale(-1)" />
   </g>
-  <g id="red" style="fill:none;stroke:#cc0000;stroke-width:4">
-    <circle r="26" cy="32" cx="32" id="circle9" />
-    <path d="M 50,50 L 14,14" id="path11" />
-    <path d="M14,50l36-36" id="path13" />
+  <g
+     id="red"
+     style="fill:none;stroke:#cc0000;stroke-width:5.82365"
+     transform="matrix(-0.68685493,0,0,-0.68685493,54.001462,53.947177)">
+    <circle
+       r="26"
+       cy="32"
+       cx="32"
+       id="circle9"
+       style="stroke-width:5.82365" />
+    <path
+       d="M 50,50 14,14"
+       id="path11"
+       style="stroke-width:5.82365" />
+    <path
+       d="M 14,50 50,14"
+       id="path13"
+       style="stroke-width:5.82365" />
   </g>
-  <g id="dark" style="fill:none;stroke:#ef2929;stroke-width:1.99999988">
-    <circle r="27" cy="32" cx="32" id="circle16" />
-    <path d="M 50,52 L 12,14" id="path18" />
-    <path d="M12,50l38-38" id="path20" />
+  <g
+     id="red-6"
+     style="fill:none;stroke:#cc0000;stroke-width:5.16915"
+     transform="matrix(-0.77382069,0,0,-0.77382069,68.948538,68.948534)">
+    <circle
+       r="24.325176"
+       cy="47.670937"
+       cx="47.67094"
+       id="circle9-7"
+       style="fill:none;fill-opacity:1;stroke:#ef2929;stroke-width:2.58457;stroke-dasharray:none;stroke-opacity:1" />
+  </g>
+  <g
+     id="dark"
+     style="fill:none;stroke:#ef2929;stroke-width:2.91182"
+     transform="matrix(-0.68685493,0,0,-0.68685493,53.94718,53.947177)">
+    <path
+       d="M 50,52 12,14"
+       id="path18"
+       style="stroke-width:2.91182" />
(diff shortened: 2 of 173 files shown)


obelisk79:
@abdullahtahiriyo this also compliments this PR with cursor symbols:
https://github.com/FreeCAD/FreeCAD/pull/11749

abdullahtahiriyo:
@obelisk79

This PR is missing the activation of the coincident unified icon. I will push a separate PR to fix that.

That other PR about cursors, I am going to let Paddle review first.

maxwxyz:
@abdullahtahiriyo I did not push the activation of the new icon, as on the sketch canvas only the on point icon is displayed, never the coincident icon. I was not sure if the unified icon should be in the toolbar when the on point is only shown.

obelisk79:
@abdullahtahiriyo 
Constraint_Coincident

abdullahtahiriyo:
@maxwxyz 

```
diff --git a/src/Mod/Sketcher/Gui/CommandConstraints.cpp b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
index 41070ff35a..7d0fc61245 100644
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -3542,7 +3542,7 @@ CmdSketcherConstrainCoincidentUnified::CmdSketcherConstrainCoincidentUnified(con
         "or a concentric constraint between circles, arcs, and ellipses");
     sWhatsThis = "Sketcher_ConstrainCoincidentUnified";
     sStatusTip = sToolTipText;
-    sPixmap = "Constraint_PointOnObject";
+    sPixmap = "Constraint_Coincident";
 
     ParameterGrp::handle hGrp = App::GetApplication().GetParameterGroupByPath(
         "User parameter:BaseApp/Preferences/Mod/Sketcher/Constraints");

```

When the merge of the PoO and Coincident is enabled in preferences, the icon that is shown after this PR for this merged tool is the one of PointOnObject (curve with one point on it). I think you want the one that @obelisk79 is showing. Then, you need to change this line above.

My internet connection, for some reason, is not working properly for GH, so I cannot push this ATM. I will try tomorrow...

maxwxyz:
@abdullahtahiriyo I know and did this for my test builds but I was not sure if the new unified icon from @obelisk79 is good for the toolbar.
I like the look but when I apply this constraint, always the OnPoint constraint is shown on the canvas above the geometry because the coincident is never shown on the geometry. Therefore this would be the only constraint icon which is never visible on the sketch plane on top of geometry

abdullahtahiriyo:
Regarding a PR that is not draft, the expectation of a maintainer is that there have been enough discussion and that what is proposed has substantial consensus and is final. If for any reason, it is not the case, the PR should be draft and any issue should be raised right away. This is not aimed as criticism, but as improve future PRs.

From what is posted above, I understood that this PR is "just" so that: (a) the same visual principles (thicknesses,...) are applied to all icons to remove an inconsistency of drawing design, (b) some clear errors about wrongly marked selection of elements of geometry icons are corrected, and (c) some icons that are difficult to tell apart each other are better differentiated. All those changes belong to the realm of "art", they do not change functionality, which is the reason why it was merged quick.

Now, let's try to see if I understand the matter ok and to resolve the open issue (also others are welcome here):
1. Yes, it is true that the coincident constraint does not have an icon in the 3D view. That is also the case for all internal alignment constraints (axes of ellipses, focus of parabola, ...), but those are automated. The coincident is still the only one where selection of geometry causes a selection of a constraint. This behaviour is very convenient and I do not think we should look into changing it. Let me know if you agree to this point.
2. The situation (1) is about the same for the current coincident icon (point with outcoming rays), in that this is not shown either.
3. The situation (2) is only similar, and not the same, for this "unified coincident" because sometimes it produces a coincident and others a point on object. Then, the point on object icon is shown in the 3D view (canvas), and it does not match the new coincident icon, but shows the point on object equivalent (arc with point).

I think that:
A. I do not see a problem from (1) and (2) above.
B. I see some degree of inconsistency in (3). The different with other inconsistencies is that you cannot remove it without removing functionality (of the merged coincident and point on object constraint). There are three options: (a) to leave the inconsistency, (b) to remove the functionality, or (c) to change the definition of inconsistent.
C. We have more constraint "buttons" which generate something that is not the same constraint as the icon (e.g. the lock constraint, which generates vertical and horizontal dimensional constraints, the dimensioning tool which creates many different ones). At least from a nomenclature point of view, I think we should differentiate "constraint" from "tool for generating constraints". The former shows the "consistent" behaviour (icon on button, icon in 3D view, user gets what he sees) and the latter does not (and will not, because functionality is intended not to).
D. I do not think we should venture into "removing all toolbar icons which are "tools for generating constraints", that would lead to a less powerful sketcher [so we should discard (b) as a solution].
E. The difference between (a) and (c) is only if we are going to do something about the inconsistency. The former is ignoring its existence and move on. The latter is to see if the UI/UX should be different to account for it (for example by different grouping, or location at different toolbars).
F. What I think needs to be different for these "tools for generating constraints", is that they will have an icon (tool icon) that is not the one icon that will appear in the 3D view. The icons (or labels if dimensional) in the 3D view will necessarily be different.

What I wrote above, I think applies to any change relating to "tools for generating constraints". So if we accept that they will have a different icon (and then I would introduce the patch of my previous post), what is left is how we are going to (re)arrange them, or if we are going to do anything different, because they are not the same. That I think is part of another different PR.

In that other PR, we do need to address the fact that we still have users who want "concise toolbars" and others who are not at lack of screen space, or are learning CAD and prefer to have a more "traditional" arrangement. When we accept that the requirements of both groups cannot be reconciled, I think we should still aim for the higher amount of common elements.

So, coming back to this PR, let me know if we want the behaviour as the patch above, or if we want anything else.

maxwxyz:
Ok then I am fine with the patch, as the unified tool can be toggled in the preferences and people who like individual toolbar buttons and that the icon is reflected on the 3D view could change it anyways. 
On a side note there are additional ideas around for displaying coincident points, also to indicate open wires (display something around, hollow points, different colors, subtle highlighting of closed islands).

PaddleStroke:
> Yes, it is true that the coincident constraint does not have an icon in the 3D view.

Regarding this, I offered a solution while you were away that you may have missed. It is to draw un-coincident points with the construction geo color (blue now). And when points are coincident they would become red. This way it would be very easy to see open wires.
image
And to the question, what if more than 3 points are stack, we just draw blue points on top of red ones.
Any thoughts on this?

maxwxyz:
I like this idea but what if the geometry is fully constrained? Then all points become green: 
image
I personally like the idea of closed islands / wires https://github.com/FreeCAD/FreeCAD/issues/11482 or some hollow point (stoke only) versos a filled point.

PaddleStroke:
I don't think both ideas are incompatible actually. They would be nice together. Because you can have a closed shape without the points being coincident.

maxwxyz:
Not sure if you actually can have a closed shape without a closed wire

obelisk79:
Isn't this a bit out of scope of this pr?

I agree with paddle about combining the two could be complimentary. But also there are no closed shapes that don't have all segment points coincident.

PaddleStroke:
Yes it's out of scope. I was just commenting about it because abdullah mentioned this topic.

abdullahtahiriyo:
I understand the convenience of islands for the 3D building process. However a sketch does not build 3D geometry on its own (it could be a path for a milling bit, laser, or a sewing machine). Somehow, what could become a positive or negative space rather depends on the tool which uses a sketch as input. That is not even necessarily a pad. 

I think I would prefer a system which does not impose restrictions on what the user expects for the 3D tool at sketch level, but rely on objective information of what is happening at sketch level.

Regarding Paddle's proposal, I think it is a better way. There are no blue points in the sketcher (as a construction point is red by convention, and hardly ever used sketch-defining points are white, as any defining geometry). The drawbacks I see are:
i. There is an argument that it would be attributing yet another function to colour in the sketcher, as colour is already coding type of geometry, construction state and constrained state. 
ii. I have some concern that some points will be red and others blue, without this meaning anything is inherently wrong or different (all are sketcher construction points with equal behaviour). I have some fear this will be misleading users, who would like to understand what is wrong, when nothing is inherently wrong with it.

On a separate point, moving construction points to blue by default would likely mean changing all art in the sketcher for consistency. Not a technical argument. Just a note.

I think some time ago we talked about adding some material around coincident constraints (as the rays of the coincident icon). Nobody ever implemented it and I am unsure if there were arguments against.

In any case, when/if we address that, we need to have a separate GH issue open.

obelisk79:
This conversation should move to a GitHub issue. I agree.

PaddleStroke:
I will create an issue and we can move the posts there

PaddleStroke:
https://github.com/FreeCAD/FreeCAD/issues/11799

PR #11795, opened 2023-12-21
Adding version check

leviathanch:
This adds a missing version check, making FreeCAD compile with Qt6


davesrocketshop (approved):

src/Mod/Material/App/MaterialLoader.cpp
```diff
@@ -29,6 +29,10 @@
 #include <QList>
 #include <QMetaType>
 
+#if QT_VERSION >= QT_VERSION_CHECK(6,0,0)
+#include <QtCore5Compat/QRegExp>
```
chennes, line 33: Can the code be modified to use `QRegularExpression` instead? `QRegExp` is very old.

davesrocketshop: Yes. This should be a direct substitution in this code.

chennes: Are you doing that, or should @leviathanch modify this PR to take that course of action instead?

davesrocketshop: I can do it but that will come in a future PR. This should be fine in the interim.

PR #11800, opened 2023-12-21
[base] Extend Python class App.Vector by isParallel function

edi271:
Extend Python class App.Vector by a boolean function `isParallel`.

Usage: 
```python
       vec1 = App.Vector(...)
       vec2 = App.Vector(...)
       tol = 0.01
       if vec1.isParallel(vec2,tol):
          ....
```
Returns True if the two vectors are parallel, otherwise returns False.

### DIFF
--- a/src/Base/Vector3D.cpp
+++ b/src/Base/Vector3D.cpp
@@ -228,6 +228,25 @@ bool Vector3<float_type>::IsEqual(const Vector3<float_type>& rclPnt, float_type 
 }
 
 template<class float_type>
+bool Vector3<float_type>::IsParallel(const Vector3<float_type>& rclPnt, float_type tol) const
+{
+    Vector3<float_type> v1 = *this;
+    Vector3<float_type> v2 = rclPnt;
+    double dot = abs(v1 * v2);
+    double mag = v1.Length() * v2.Length();
+    return (abs(dot - mag) < tol);
+}
+
+template<class float_type>
+bool Vector3<float_type>::IsNormal(const Vector3<float_type>& rclPnt, float_type tol) const
+{
+    Vector3<float_type> v1 = *this;
+    Vector3<float_type> v2 = rclPnt;
+    double dot = abs(v1 * v2);
+    return (dot < tol);
+}
+
+template<class float_type>
 Vector3<float_type>& Vector3<float_type>::ProjectToPlane(const Vector3<float_type>& rclBase,
                                                          const Vector3<float_type>& rclNorm)
 {
--- a/src/Base/Vector3D.h
+++ b/src/Base/Vector3D.h
@@ -210,6 +210,10 @@ public:
      * equal.
      */
     bool IsEqual(const Vector3& rclPnt, float_type tol) const;
+    /// Returns true if two vectors are parallel within the tol
+    bool IsParallel(const Vector3& rclPnt, float_type tol) const;
+    /// Returns true if two vectors are normal within the tol
+    bool IsNormal(const Vector3& rclPnt, float_type tol) const;
     /// Projects this point onto the plane given by the base \a rclBase and the normal \a rclNorm.
     Vector3& ProjectToPlane(const Vector3& rclBase, const Vector3& rclNorm);
     /**
--- a/src/Base/VectorPy.xml
+++ b/src/Base/VectorPy.xml
@@ -158,6 +158,17 @@ vector2 : Base.Vector
 tol : float</UserDocu>
             </Documentation>
         </Methode>
+        <Methode Name="isNormal" Const="true">
+            <Documentation>
+                <UserDocu>isNormal(vector2, tol=0) -> bool
+
+Checks if this vector and `vector2` are
+normal less or equal to the given tolerance.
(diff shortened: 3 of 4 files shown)


chennes:
Thanks for the PR -- this looks good to me, but I'd like other maintainers to weigh in on whether the addition to the API is warranted (and if so, should we also have an "isPerpendicular" or "isNormal" method?)

edi271:
The Base::Vector3d class is nowhere documented. If we extend it, nobody will recognize, and nobody will use the extension.

Pythons App.Vector class is well documented:

a) In the wiki: https://wiki.freecad.org/Vector_API
b) If you enter in the Python-Consol **App.Vector.** you receive a list off all methods.

But I will have a look, and try to extend the Base::Vector3D class.

chennes:
@WandererFan this looks good to me, so if you're happy with it go ahead and merge.

src/Base/Vector3D.cpp
```diff
@@ -227,6 +227,25 @@ bool Vector3<float_type>::IsEqual(const Vector3<float_type>& rclPnt, float_type
     return Distance(*this, rclPnt) <= tol;
 }
 
+template<class float_type>
+bool Vector3<float_type>::IsParallel(const Vector3<float_type>& rclPnt, float_type tol) const
+{
```
wwmayer, line 232: The code has some issues:
* the argument is a vector and not a point, thus calling the parameter rclPnt is somewhat confusing
* the function doesn't handle the case if one if the vectors is the null vector
* it's not very efficient because the method Length() uses sqrt() internally.

src/Base/Vector3D.cpp
```diff
@@ -227,6 +227,25 @@ bool Vector3<float_type>::IsEqual(const Vector3<float_type>& rclPnt, float_type
...
+    Vector3<float_type> v1 = *this;
+    Vector3<float_type> v2 = rclPnt;
+    double dot = abs(v1 * v2);
+    double mag = v1.Length() * v2.Length();
+    return (abs(dot - mag) < tol);
+}
+
+template<class float_type>
+bool Vector3<float_type>::IsNormal(const Vector3<float_type>& rclPnt, float_type tol) const
+{
```
wwmayer, line 242: This code suffers from the first and second point as IsParallel()

src/Base/Vector3D.cpp
wwmayer: With the API of [Rhino](https://developer.rhino3d.com/api/rhinocommon/rhino.geometry.vector3d/isparallelto?version=8.x) it returns false if one of the vectors is the null vector.

src/Base/Vector3D.cpp
wwmayer: The method doesn't even work correctly. If v1=(1,2,3) and v2(3,0,-1) then the dot product is exactly 0.0 but the check `dot < tol` fails if tol is 0.0.

wwmayer: Another weakness is if you allow a certain tolerance because then the result depends on the length of the input vectors. Assuming an angular tolerance of 1 degree is allowed then tol will be 0.01745
The use the vectors v1=(10000,20000,30000) and v2(30000,00000,-10001)
`
math.pi/2 - v1.getAngle(v2)
`
gives `-2.5354374089570086e-05` which is clearly inside the tolerance but dot will be -30000.0.

edi271:
Testing with the vectors (1,2,3) and (3,0,-1) IsPrarallel returns the value "False" which is correct. I can't reproduce this issue.
If one vector is the null vector, IsParallel returns "True". Does a null vector have a direction ? Is every vector parallel to a null vector ?
The function Length is used in several other functions (DistanceToPlane, DistanceToLine, GetAngle).
The name rclPnt is not the best, rcVect would be better. My model was IsEqual.

wwmayer:
> Testing with the vectors (1,2,3) and (3,0,-1) IsParallel returns the value "False" which is correct.

No, it's not. A vector is defined by its length, by its direction and by its orientation. The check for parallelism only compares the directions but ignores orientation. If the angle is 0 or PI then they are considered as parallel. This is also how OCC handles it: https://git.dev.opencascade.org/gitweb/?p=occt.git;a=blob;f=src/gp/gp_Dir.hxx;h=37b6e7b231703ed9b8c570dde6e083ba8d3ae245;hb=HEAD#l145

One could add another method IsOpposite() (like OCC does it) that explicitly handles the case when the angle is PI.

> If one vector is the null vector, IsParallel returns "True".

From an algebraic point of view this can be considered as correct but from a geometric point of view this is useless. That's why several other APIs (like Rhino, OCC, ...) also handle it as not-parallel.

> Does a null vector have a direction ?

No.

> Is every vector parallel to a null vector ?

As said from an algebraic point of they are because the cross product is 0. But from a geometric point of view where we work with objects that have an expansion this will be quite useless.

> The function Length is used in several other functions (DistanceToPlane, DistanceToLine, GetAngle).

Well, without the angular tolerance value one could implement it more efficiently. But with the tolerance the angle has to be determined anyway.

Issue #10360, opened 2023-08-25
[Issue] TopoNaming Phase 3 (card 8 of 35) - Add History Support/Part Workbench/FeaturePartFuse


JohnAD:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
[code]
OS: Pop!_OS 22.04 LTS (pop:GNOME/pop)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33675 (Git) AppImage
Build type: Release
Branch: (HEAD detached at 0.21.0)
Hash: 41b058e2087de60dc8fef3d3e68c7d0129e13abf
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: English/United States (en_US)
[/code]
```

### Subproject(s) affected?

Part

### Problem description

This is a card for Toponaming fix Phase 3 (add naming) Part 1 (history expansion).

Please see TN issue https://github.com/FreeCAD/FreeCAD/issues/8432 for overall discussion.

Also see project https://github.com/orgs/FreeCAD/projects/2/views/1 for tracking.

Phase 3 does not USE the naming yet, that is Phase 4.

# Specifics

Naming is based on history. Add the RealThunder changes to history to expand it's support.

RealThunder reference branch: https://github.com/FreeCAD/FreeCAD/tree/development/toponaming

Do this for the workbench and file(s) mentioned in the issue title. The few cards are generic and the remaining ones are for specific `Feature` elements in the related workbench.

Please add one or more unit tests.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chennes:
Fixed by #12517

PR #11803, opened 2023-12-21
Preliminary cpp tests for Part boolean operations

bgbsww:
This is speculative and ahead of the upcoming test examples in #10353.  It addresses the first parts of #10355 and #10360 
and the Cut operation that I can't find a card for.

The RT code port is not yet checked into this branch as it depends on code not yet transferred into TopoShape.

This might be completely wrong, a basis for conversation, or a useful launching point.  I was using it to learn about the cpp test framework and the key needs for toponaming to move forward.


tests/src/Mod/Part/App/FeaturePartFuse.cpp
```diff
@@ -0,0 +1,145 @@
...
+    Part::TopoShape ts = _fuse->Shape.getValue();
+    // what's the wire or face count here?
+    std::vector<Part::TopoShape> subs = ts.getSubTopoShapes(TopAbs_FACE);  // WIRE
+    EXPECT_EQ(subs.size(), 14);
+    // Assert
+    _fuse->Refine.setValue(true);
+    _fuse->execute();
+    ts = _fuse->Shape.getValue();
+    subs = ts.getSubTopoShapes(TopAbs_FACE);  // WIRE
+    EXPECT_EQ(subs.size(), 6);
```
Pesc0, line 144: might be the thing with occ using 1-index @chennes?

tests/src/Mod/Part/App/FeaturePartCut.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+    // Arrange
+    _cut->Base.setValue(_box1obj);
+    _cut->Tool.setValue(_box2obj);
+
+    // Act
+    _cut->execute();
+    Part::TopoShape ts = _cut->Shape.getValue();
+    double volume = PartTestHelpers::getVolume(ts.getShape());
+    EXPECT_DOUBLE_EQ(volume, 3.0);
```
Pesc0, line 42: Looks like in this file many tests are missing the //assert. I'm guessing that's a copy paste ;)

Pesc0 (commented):
Also check linter warnings here on GH.

Dont take any comments harshly or as if i'm giving orders. I mostly just write my thoughts very concisely, I assure you I'm not trying to be rude and I may very well be wrong.

Overall the tests are very clear and I think they are going to be useful to have, in particular the history related ones. Perhaps we should have a discussion with the other guys about which tests are going to be the most useful to have for future work. Thanks again for helping out.

tests/src/Mod/Part/App/FeaturePartCommon.cpp
```diff
@@ -0,0 +1,209 @@
...
+    EXPECT_TRUE(mE);
+    _doc->recompute();
+    mE = _common->mustExecute();
+    EXPECT_FALSE(mE);
+}
+
+TEST_F(FeaturePartCommonTest, testGetProviderName)
+{
+    // Act
+    _common->execute();
```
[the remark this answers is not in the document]
bgbsww: I think https://google.github.io/googletest/primer.html#invoking-the-tests says that the state is always reset.

Correct that _common is not fully initialized here; however this test should and does pass even with a partially initialized object.  I'm trying to match wart-for-wart with classical FC code so I'm doing it this way since I think we'd want to know if it changed.  At that point, someone can argue for changing the test if it is appropriate.  Open to opinions here.

tests/src/Mod/Part/App/FeaturePartFuse.cpp
bgbsww: That was really just a left over pre-coding comment, not an open question for me.  But if the resulting code note isn't clear, I'm happy to change it.

tests/src/Mod/Part/App/FeaturePartCommon.cpp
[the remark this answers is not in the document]
Pesc0: Great, leave it as-is then

tests/src/Mod/Part/App/FeaturePartFuse.cpp
Pesc0: Yeah TopAbs_FACE with //WIRE is a bit confusing

Pesc0 (approved):
Unfortunate that you removed the history thing in the last commit. why doesn't it work?

bgbsww:
Actually, it never did.  The good news is that is if the tests fail at the end of FeaturePartCommonTest testHistory, the shapeMap will print out cleanly, since std::map, std::vector and int all have working << operators.

I'd hoped to be able to provide PrintTo operators not just for ShapeHistory, but also TopoShape and TopoDS_Shape.

Unfortunately, after a day's detour on this issue, I find that GoogleTest can't be made to recognize any method that uses parameters that are defined outside of the namespace, and none of my attempts to bring those objects into the namespace worked without nasty side effects that made the client code miserable.

Maybe someone else will have a good idea that works;  I discovered that that PrintTo never actually worked, I was just fooled because I was getting output.

Pesc0:
Got it! Totally fine to do without then

PR #11804, opened 2023-12-21
[TechDraw] Implement proper selection of subitems

pavltom:
As Part View is able to provide sub-items to selection, it is desirable to make the vertices/edges/faces also visually selected, when the current selection is changed e.g. from Python code.

This pull request adds this ability, while trying to respect the subitems selection attitude. This means when nothing is selected:
`Gui.Selection.addSelection('doc', 'View')` adds only the view to the selection, but none of the subitems. The frame and label of the view are highlighted.
`Gui.Selection.addSelection('doc', 'View', 'Edge1')` adds the 1st edge to the selection, highlights it, causes the view item to be selected in the tree, but does not highlight the view frame and label on the TechDraw page.
`Gui.Selection.addSelection('doc','View', '')` adds the view, but none of its subitems. It acts in the very same way as the first code snippet.

`Gui.Selection.removeSelection('doc', 'View')` makes the view and all its selected sub items unselected.
`Gui.Selection.removeSelection('doc', 'View','Edge1')` unselects the 1st edge, but keeps the view frame and label selected if they were selected before. If only the edge was selected (not the view itself), the view item is also unselected in the tree.
`Gui.Selection.removeSelection('doc', 'View','')` behaves in the very same way as the first removal code line does.

`Gui.Selection.clearSelection('doc')` behaves as expected, making all views and their sub-items to become unselected. The current behavior not unselecting the edges etc. is fixed.

I hope this small addition proves useful, if there are any issues caused by the changes, please let me know.

### DIFF
--- a/src/Mod/TechDraw/Gui/DrawGuiUtil.cpp
+++ b/src/Mod/TechDraw/Gui/DrawGuiUtil.cpp
@@ -539,3 +539,33 @@ double DrawGuiUtil::roundToDigits(double original, int digits)
     temp = rounded / factor;
     return temp;
 }
+
+// Returns true if the item or any of its descendants is selected
+bool DrawGuiUtil::isSelectedInTree(QGraphicsItem *item)
+{
+    if (item) {
+        if (item->isSelected()) {
+            return true;
+        }
+
+        for (QGraphicsItem *child : item->childItems()) {
+           if (isSelectedInTree(child)) {
+               return true;
+           }
+        }
+    }
+
+    return false;
+}
+
+// Selects or deselects the item and all its descendants
+void DrawGuiUtil::setSelectedTree(QGraphicsItem *item, bool selected)
+{
+    if (item) {
+        item->setSelected(selected);
+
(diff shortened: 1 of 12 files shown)


WandererFan:
This does not affect the user if they are not using Python scripting, correct?  I didn't notice any differences while randomly pushing buttons.

WandererFan (approved):

luzpaz:
Could we get a screencast of the changes ? :pray:

pavltom:
@WandererFan From UI point of view, there are practically no changes, except subtle oddities being fixed, like selecting e.g. part view edge, which in turn selects the the view item in the tree, but later unselecting the view item via the tree (e.g. clicking on some other tree item) does not unselect the part view edge. But the central point is to consolidate GUI with Python script calls.
@luzpaz This is rather Python <-> GUI communication fix, there is really not much to show, maybe except typing the Python commands into the console, while demonstrating the vertices/edges/faces are (un)selected properly. If you would not mind, I would skip the screencast this time... ?

PR #11813, opened 2023-12-22
[TechDraw] italicize leaderline auto-horizontal preference checkbox

alexneufeld:
... to correctly indicate that it only applies to new objects.


luzpaz:
Can you provide screenshot of change ? TIA

PR #11815, opened 2023-12-23
[TechDraw] Tree view update refactoring

pavltom:
This pull request aims to replace "horrible hacks" with plain touch() call doing basically the same thing, i.e. updating the tree view, so the dimension item becomes a child of its part view. There should be no functionality changes except auto-selecting the newly created dimension for improved workflow.

In case any there are any problems or unexpected effects, please let me know.


WandererFan (approved):

WandererFan:
I love it when I can cross something off my to do list without expending any effort!

Thanks, Tomas.

PR #11821, opened 2023-12-24
Sketcher: change tool description of rotate and polygon tool

maxwxyz:
Change the tool description of polar pattern (rotate) and polygon in the Sketcher WB to better hint users to modify with U and J key. Added line breaks to the copy constraints tooltip.
@PaddleStroke fixing https://github.com/FreeCAD/FreeCAD/pull/11264#issuecomment-1868324482

Before:
grafik

After:
grafik

### DIFF
--- a/src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
+++ b/src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
@@ -490,11 +490,12 @@ void DSHRotateController::configureToolWidget()
             QApplication::translate("TaskSketcherTool_c1_offset", "Clone constraints"));
         toolWidget->setCheckboxToolTip(
             WCheckbox::FirstBox,
-            QApplication::translate(
-                "TaskSketcherTool_c1_offset",
-                "This concerns the datum constraints like distances. If you activate Clone, "
-                "then the tool will copy the datum. Else it will try to replace them with "
-                "equalities between the initial geometries and the new copies."));
+            QString::fromLatin1("<p>")
+                + QApplication::translate("TaskSketcherTool_c1_offset",
+                                          "This concerns the datum constraints (e.g. distance)."
+                                          "If you activate Clone, the tool will copy the datum."
+                                          "Else it will try to replace them with equalities.")
+                + QString::fromLatin1("</p>"));
     }
 
     onViewParameters[OnViewParameter::First]->setLabelType(Gui::SoDatumLabel::DISTANCEX);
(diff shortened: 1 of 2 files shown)


src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
```diff
@@ -508,7 +508,7 @@ void DSHRotateController::configureToolWidget()
 
     toolWidget->setParameterLabel(
         WParameter::First,
-        QApplication::translate("TaskSketcherTool_p4_rotate", "Copies 'U'/'J'"));
+        QApplication::translate("TaskSketcherTool_p4_rotate", "Copies (modify with 'U'/'J')"));
```
PaddleStroke, line 511: I'm concerned this string makes the whole box too wide. What about :
Sides (U/J to +/-)
or
Sides (U+,  J-)

src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
```diff
@@ -492,9 +492,9 @@ void DSHRotateController::configureToolWidget()
             WCheckbox::FirstBox,
             QApplication::translate(
                 "TaskSketcherTool_c1_offset",
-                "This concerns the datum constraints like distances. If you activate Clone, "
-                "then the tool will copy the datum. Else it will try to replace them with "
-                "equalities between the initial geometries and the new copies."));
+                "This concerns the datum constraints like distances.\nIf you activate Clone, "
+                "then the tool will copy the datum.\nElse it will try to replace them with "
+                "equal constraints."));
```
PaddleStroke, line 497: Ideally it is better to keep the html out of the translated texts. So you could do like this : 
```
        toolWidget->setCheckboxToolTip(
            WCheckbox::FirstBox,
            QString::fromLatin1("<p>")
            + QApplication::translate( "TaskSketcherTool_c1_offset",
                "This concerns the datum constraints like distances.If you activate Clone, "
                "then the tool will copy the datum. Else it will try to replace them with "
                "equalities between the initial geometries and the new copies.") 
            + QString::fromLatin1("</p>"));
```

src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
maxwxyz: what about: Sides (+'U' / -'J')
Same syntax is used in other tool settings in sketcher, e.g. 'M'

PaddleStroke: seems ok
Or we just remove this and put it in the tooltip.

src/Mod/Sketcher/Gui/DrawSketchHandlerRotate.h
maxwxyz: @WandererFan would be great if you could place the html outside the translate text as I am sometimes struggling with the TechDraw translations :)

chennes:
@PaddleStroke are you good with this now?

PaddleStroke:
Yes it is ok

PR #11828, opened 2023-12-24
Update addonmanager_git.py

RohanMishra315:
Update addonmanager_git.py


chennes:
As a general rule we do not use "print" statements in FreeCAD. Instead you should use the FreeCAD.Console methods so the user can control the level of output they see.

RohanMishra315:
My bad @chennes . Would keep in mind😋

luzpaz:
Change it please

PR #11829, opened 2023-12-25
Adding additional TNP tests

bgbsww:
Work in progress on more testing in the new toponaming code.


Pesc0:
Looks ok for now
I'm not really sure about those mustExecute tests, perhaps someone with more occt experience could explain what is going on.
Also keep in mind that it's fine to write a failing test if the behavior isn't quite what you expected. This case would be a perfect example of that.

PR #11837, opened 2023-12-26
Gui: remember user closed overlay panel

### LINKED ISSUE #11680: Overlay: panels don't stay closed

- [x] I have searched the existing issues

### Problem description

When an overlay panel is closed by clicking the close button (X), it closes but appears back up at the same place on next launch. This does not happen with normal docked panels.

### Full version info

```shell
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35176 (Git) AppImage
Build type: Release
Branch: main
Hash: dcd533af4bf0cb3dbc50c7d0e2fb579cad168819
Python 3.10.13, Qt 5.15.8, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Installed mods: 
  * OpenDark 2023.11.27
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

realthunder:
Fixes #11680

### DIFF
--- a/src/Gui/OverlayManager.cpp
+++ b/src/Gui/OverlayManager.cpp
@@ -814,27 +814,27 @@ public:
             return;
         case OverlayManager::OverlayMode::ToggleLeft:
             if (OverlayTabWidget::_LeftOverlay->isVisible())
-                OverlayTabWidget::_LeftOverlay->setState(OverlayTabWidget::State::Hidden);
+                OverlayTabWidget::_LeftOverlay->setState(OverlayTabWidget::State::Hidden, /*user_triggered*/true);
             else
-                OverlayTabWidget::_LeftOverlay->setState(OverlayTabWidget::State::Showing);
+                OverlayTabWidget::_LeftOverlay->setState(OverlayTabWidget::State::Showing, /*user_triggered*/true);
             break;
         case OverlayManager::OverlayMode::ToggleRight:
             if (OverlayTabWidget::_RightOverlay->isVisible())
-                OverlayTabWidget::_RightOverlay->setState(OverlayTabWidget::State::Hidden);
+                OverlayTabWidget::_RightOverlay->setState(OverlayTabWidget::State::Hidden, /*user_triggered*/true);
             else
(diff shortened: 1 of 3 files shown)


src/Gui/OverlayWidgets.cpp
```diff
@@ -923,7 +929,7 @@ void OverlayTabWidget::onAction(QAction *action)
     OverlayManager::instance()->refresh(this);
 }
 
-void OverlayTabWidget::setState(State state)
+void OverlayTabWidget::setState(State state, bool user_triggered)
```
kadet1090, line 932: Every other variable in FreeCAD uses `camelCase` - why this is in the `snake_case`?

src/Gui/OverlayManager.cpp
```diff
@@ -814,27 +814,27 @@ class OverlayManager::Private {
             return;
         case OverlayManager::OverlayMode::ToggleLeft:
             if (OverlayTabWidget::_LeftOverlay->isVisible())
-                OverlayTabWidget::_LeftOverlay->setState(OverlayTabWidget::State::Hidden);
+                OverlayTabWidget::_LeftOverlay->setState(OverlayTabWidget::State::Hidden, /*user_triggered*/true);
```
chennes, line 817: If you made it an enum instead of a bool you could drop the comment.

kadet1090: @chennes should I pick this, apply this change and drop new PR?

chennes: IMO that would be delightful: or I can merge this and you can do it in another pass.

kadet1090:
@chennes @yorikvanhavre so, I've checked the issue with @qewer33 who is original author of the issue. It does not seem to resolve the described issue as the "x" button only hides related area and does not remove the panel. Overlay area correctly remembers the state after this change and state is not updated by accident so it is an improvement nonetheless. 

I've prepared a PR with requested changes: https://github.com/FreeCAD/FreeCAD/pull/11998 to replace this one, but it does not fix described issue.

PR #11840, opened 2023-12-27
Change the description of the default 'Toggle active object' command

maxwxyz:
Change the description of the default 'Toggle active object' command in the right click menu based on the current state of the selected object.
Currently it always says "Toggle active object".
The PR would say "Active object" and displays a checkbox as icon. If the object is currently not the active object the checkbox is unchecked, otherwise the checkbox is checked.
The code previously called the double click function which was now refactored in two new functions.

Before:
grafik

After:

https://github.com/FreeCAD/FreeCAD/assets/6246609/2b237541-f1e3-4b16-9cb3-3e9cb10e0ee1

### DIFF
--- a/src/Gui/ViewProviderPart.cpp
+++ b/src/Gui/ViewProviderPart.cpp
@@ -68,35 +68,20 @@ void ViewProviderPart::onChanged(const App::Property* prop) {
 
 void ViewProviderPart::setupContextMenu(QMenu* menu, QObject* receiver, const char* member)
 {
-    App::DocumentObject* activePart = nullptr;
-    auto activeDoc = Gui::Application::Instance->activeDocument();
-    if(!activeDoc)
-        activeDoc = getDocument();
-    auto activeView = activeDoc->setActiveView(this);
-    activePart = activeView->getActiveObject<App::DocumentObject*> (PARTKEY);
-
     auto func = new Gui::ActionFunction(menu);
 
-    if (activePart == this->getObject()){
-        QAction* act = menu->addAction(QObject::tr("Unset active object"));
-        func->trigger(act, [this](){
-        this->doubleClicked();
+    QAction* act = menu->addAction(QObject::tr("Active object"));
+    act->setCheckable(true);
+    act->setChecked(isActivePart());
+    func->trigger(act, [this](){
+    this->toggleActivePart();
     });
-    } else {
-        QAction* act = menu->addAction(QObject::tr("Set active object"));
-        func->trigger(act, [this](){
-        this->doubleClicked();
-    });
-    }
 
     ViewProviderDragger::setupContextMenu(menu, receiver, member);
 }
 
-bool ViewProviderPart::doubleClicked()
+bool ViewProviderPart::isActivePart()
 {
-    //make the part the active one
-
-    //first, check if the part is already active.
     App::DocumentObject* activePart = nullptr;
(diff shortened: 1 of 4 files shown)


leoheck:
Why not to just use the direct form like: 
`Activate Part/Assembly/Object`
`Deactivate Part/Assembly/Object`

maxwxyz:
The part one seems to be in use for everything except bodies. That's why it is called object

leoheck:
Part/Assembly/Objet was not the thing I was highlighting here.
What I wanted to suggest was to replace `Set (something) to Active` with `Activate (something)` or `Activate Edition (mode)` and the opposite to `Deactivate (something) ` or `Deactivate Edition (mode)`

maxwxyz:
I'm not familiar with the term `activate edition`, I thought about `activate` and `deactivate` but I think this could confuse users (especially new ones). 
What is deactivated? Is it still visible, is it still usable for other features/binders? The realthunder branch has the possibility to disable / temporarily suppress PartDesign features inside a body, this could also be confused with deactivate.
If you have a better terminology let me know, I thought about activate/deactivate, set active / set inactive, set active / unset active. I went ahead with the last option as inactive seemed also ambiguous.

leoheck:
Yeah that term was invented by me trying to keep both terms tied to the same thing.

I like this pair `activate/deactivate`. I don't think it is ambiguous. But I use the standard Freecad, not a custom version of it, so I don't see issues here. I believe this real thunder will change something if his features are ambiguous with the standard Freecad.

I just want to thank you for this PR since having something different than "toggling" is enough for me. Toggling doesn't show us the end result of the action. This is specially important when the active item doesn't change visually for any reason.I feel it should be improved since a long time ago.

maxwxyz:
@leoheck check the first post, I've updated it to show a checkbox which is checked when the selected object is already active and vice versa. This was the result of the discussion of the DWG on discord.

leoheck:
The video is super fast. It is hard to watch and understand. The checkbox is horrible by the way, I cannot see when it is active or not. The text is still the same. 

Why don't you just put:
- Activate Part/Object/Assembly/Body
- Deactivate Part/Object/Assembly/Body

maxwxyz:
As mentioned before: there is no concept of a deactivated / disabled object, just the current active state of an object. Operations will be done to the current active object. We've discussed this in the UI group the last day what would be the best terminology and the best which came up is the active object with the checkmark.

leoheck:
If there is a checkbox, then you can add the "deactivate" text instead, it is just a text for the end user. The UI is the user interface for the end user not the developer.

kadet1090:
> If there is a checkbox, then you can add the "deactivate" text instead, it is just a text for the end user. The UI is the user interface for the end user not the developer.

The label of checkbox in most cases represents the state when the checkbox is selected. So for example `[ ] Lock toolbars` should read as `No, toolbars are not locked` and `[x] Lock toolbars` should read as `Yes, toolbars are locked`. Or `[x] I agree to terms of service` reads `Yes, I agree to terms of service` or `No, I don't agree to terms of service` in case of unchecked. 

The label of checkbox should always represent what is the state and not what it will become after click. This is true both for checkboxes in dialogs and checkable menu items.

image
(from: https://forum.qt.io/topic/144311/is-a-qmenu-item-supposed-to-be-able-to-be-checkable/2)

PR #11844, opened 2023-12-27
Always show auto HorVer tool command in the toolbar (like auto dimension) when concise option is enabled

maxwxyz:
https://github.com/FreeCAD/FreeCAD/pull/11745#issuecomment-1870435624
Always show auto HorVer tool command in the toolbar when the tool is present (concise toolbar option by @obelisk79 )
Currently the last used tool is shown as command in the toolbar, this PR makes the auto tool not remember the last sub tool, like currently also implemented in the auto dimension tool.
293062926-bb05611e-72f8-404e-97d5-e9b013739f4d

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -2610,6 +2610,7 @@ public:
         eType = ForEdit;
 
         setCheckable(false);
+        setRememberLast(false);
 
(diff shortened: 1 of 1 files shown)


chennes:
IMO this is a good change -- @obelisk79 is the Design Working Group happy with this one?

obelisk79:
Yes, we've discussed this one. Thumbs up from the DWG.

PR #11847, opened 2023-12-28
Gui: Allow narrower preferences dialog

kadet1090:
This introduces optional horizontal scrollbar, looses up the minimum size applied to dialog and narrows a dialog a little bit by default.

Forum dicsussion: https://forum.freecad.org/viewtopic.php?p=724001#p724001


chennes:
I'm going to merge this as-is because it's an improvement over the current state of affairs, but the Preference Packs table probably needs some minimum size adjustment to prevent the scrollbar from appearing until it's really necessary.

Issue #11355, opened 2023-11-11
[Problem] Placement Feature Issues


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

I've stumbled across some issues with the `Placement...` feature in FreeCAD (menu bar: Edit>Placement...)

- [x] Select a body or part in the tree view and invoke the placement dialog everything works as expected
- [ ] Select a body in the 3D view, placement task is opened but has no function (maybe because 3D view selection does select the last tree feature)
- [ ] Select a feature of PartDesign in the tree view, placement task is opened but has no function (logical) --> placement should be greyed out ot fall back to body placement
- [ ] Select a Sketch, Datum or a feature which can be placed: Dialog is opened but has no effect at all
- [ ] The placement task dialog does not show on overlay "show on edit" or "auto hide" - one must place the cursor on the right screen that the dialog opens
- [ ] The placement feature is way more powerful than Transform. Can the placement... feature be implemented in the right click menu in the tree view and 3D view, like the transform feature? Both are in the edit menu bar, but only transform is in the right click menus

Here a quick demo of the issues:

https://github.com/FreeCAD/FreeCAD/assets/6246609/5d18f09d-3484-44d8-b7a9-8c6e02f6bc6e

### Full version info

```shell
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34959 (Git)
Build type: Release
Branch: main
Hash: 640018bfa7878a79e1b3c220ddd488d66464dc70
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: German/Germany (de_DE)
Installed mods: 
  * 3DfindIT 1.2.0
  * 3D_Printing_Tools
  * BIM 2021.12.0
  * CfdOF 1.24.7
  * CurvedShapes 1.0.4
  * Curves 0.6.14
  * Defeaturing 1.2.1
  * fasteners 0.4.68
  * freecad.gears 1.0.0
  * freecad_metal_workbench 0.0.1
  * OpenDark 2023.10.30
  * sheetmetal 0.3.1
  * Silk 0.1.5
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #11849, opened 2023-12-28
Add placement command to context menu

maxwxyz:
Fixes parts of #11355
Placement command is added to the context menu in the tree and view if objects are selected which can be "placed"


obelisk79:
This, along with the referenced PR appears to follow good principles, and offers a usability improvement. It also follows several of the principles outlined in the design guide here: https://github.com/obelisk79/DevelopersHandbook/blob/main/designguide/laws-of-ux.md

PR #11853, opened 2023-12-29
[Sketcher] General Tangency with B-splines (re-do)

AjinkyaDahale:
Replaces #9125 with the branch in my personal fork (as opposed to the Ondsel one).

Part of this work was done while I was part of Ondsel Inc., and the rest of the work is being supported through an FPA grant.

### DIFF
--- a/src/Mod/Sketcher/App/Sketch.cpp
+++ b/src/Mod/Sketcher/App/Sketch.cpp
@@ -2868,6 +2868,11 @@ int Sketch::addTangentConstraint(int geoId1, int geoId2)
             GCSsys.addConstraintTangent(l, a, tag);
             return ConstraintsCounter;
         }
+        else if (Geoms[geoId2].type == BSpline) {
+            Base::Console().Error("Direct tangency constraint between line and B-spline is not "
+                                  "supported. Use tangent-via-point instead.");
+            return -1;
+        }
     }
     else if (Geoms[geoId1].type == Circle) {
         GCS::Circle& c = Circles[Geoms[geoId1].index];
@@ -2888,6 +2893,11 @@ int Sketch::addTangentConstraint(int geoId1, int geoId2)
             GCSsys.addConstraintTangent(c, a, tag);
             return ConstraintsCounter;
         }
+        else if (Geoms[geoId2].type == BSpline) {
+            Base::Console().Error("Direct tangency constraint between circle and B-spline is not "
+                                  "supported. Use tangent-via-point instead.");
+            return -1;
+        }
     }
     else if (Geoms[geoId1].type == Ellipse) {
         if (Geoms[geoId2].type == Circle) {
@@ -2897,6 +2907,11 @@ int Sketch::addTangentConstraint(int geoId1, int geoId2)
         }
         else if (Geoms[geoId2].type == Arc) {
             Base::Console().Error("Direct tangency constraint between arc and ellipse is not "
+                                  "supported. Use tangent-via-point instead.");
+            return -1;
+        }
+        else if (Geoms[geoId2].type == BSpline) {
+            Base::Console().Error("Direct tangency constraint between ellipse and B-spline is not "
                                   "supported. Use tangent-via-point instead.");
             return -1;
         }
@@ -2920,6 +2935,16 @@ int Sketch::addTangentConstraint(int geoId1, int geoId2)
             GCSsys.addConstraintTangent(a, a2, tag);
             return ConstraintsCounter;
         }
+        else if (Geoms[geoId2].type == BSpline) {
+            Base::Console().Error("Direct tangency constraint between arc and B-spline is not "
+                                  "supported. Use tangent-via-point instead.");
+            return -1;
+        }
+    }
+    else if (Geoms[geoId1].type == BSpline) {
+        Base::Console().Error("Direct tangency constraint including B-splines is not "
+                              "supported. Use tangent-via-point instead.");
+        return -1;
     }
 
     return -1;
@@ -3041,7 +3066,6 @@ int Sketch::addAngleAtPointConstraint(int geoId1,
                                       ConstraintType cTyp,
                                       bool driving)
 {
-
     if (!(cTyp == Angle || cTyp == Tangent || cTyp == Perpendicular)) {
         // assert(0);//none of the three types. Why are we here??
         return -1;
@@ -3146,19 +3170,116 @@ int Sketch::addAngleAtPointConstraint(int geoId1,
     }
 
     int tag = -1;
+    // FIXME: Perform construction of any parameters where this method is called instead of here
     if (e2c) {
-        // increases ConstraintsCounter
-        tag = Sketch::addPointOnObjectConstraint(geoId1, pos1, geoId2, driving);
+        if (Geoms[geoId2].type == BSpline) {
+            GCS::Point& p1 = Points[getPointId(geoId1, pos1)];
+            auto* partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId2].geo);
+            double uNear;
+            partBsp->closestParameter(Base::Vector3d(*p1.x, *p1.y, 0.0), uNear);
+            double* pointparam = new double(uNear);
+            Parameters.push_back(pointparam);
+            --ConstraintsCounter;  // Do this just before point-on-object because ConstraintsCounter
+                                   // is increased again before being used
+            tag = addPointOnObjectConstraint(geoId1,
+                                             pos1,
+                                             geoId2,
+                                             pointparam,
+                                             driving);  // increases ConstraintsCounter
+            GCSsys.addConstraintAngleViaPointAndParam(*crv2,
+                                                      *crv1,
+                                                      p,
+                                                      pointparam,
+                                                      angle,
+                                                      tag,
+                                                      driving);
+        }
+        else {
+            // increases ConstraintsCounter
+            tag = Sketch::addPointOnObjectConstraint(geoId1,
+                                                     pos1,
+                                                     geoId2,
+                                                     driving);  // increases ConstraintsCounter
+            GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
+        }
     }
     if (e2e) {
         tag = ++ConstraintsCounter;
         GCSsys.addConstraintP2PCoincident(p, *p2, tag, driving);
+        GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
     }
     if (avp) {
         tag = ++ConstraintsCounter;
+        if (Geoms[geoId1].type == BSpline || Geoms[geoId2].type == BSpline) {
+            if (Geoms[geoId1].type == BSpline && Geoms[geoId2].type == BSpline) {
+                GCS::Point& p3 = Points[getPointId(geoId3, pos3)];
+                auto* partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId1].geo);
+                double uNear;
+                partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
+                double* pointparam1 = new double(uNear);
+                Parameters.push_back(pointparam1);
+                --ConstraintsCounter;  // Do this just before point-on-object because
+                                       // ConstraintsCounter is increased again before being used
+                addPointOnObjectConstraint(geoId3,
+                                           pos3,
+                                           geoId1,
+                                           pointparam1,
+                                           driving);  // increases ConstraintsCounter
+                partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId2].geo);
+                partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
+                double* pointparam2 = new double(uNear);
+                --ConstraintsCounter;  // Do this just before point-on-object because
+                                       // ConstraintsCounter is increased again before being used
+                addPointOnObjectConstraint(geoId3,
+                                           pos3,
+                                           geoId2,
+                                           pointparam2,
+                                           driving);  // increases ConstraintsCounter
+                Parameters.push_back(pointparam2);
+                GCSsys.addConstraintAngleViaPointAndTwoParams(*crv1,
+                                                              *crv2,
+                                                              p,
+                                                              pointparam1,
+                                                              pointparam2,
+                                                              angle,
+                                                              tag,
+                                                              driving);
+            }
+            else {
+                if (Geoms[geoId1].type != BSpline) {
+                    std::swap(geoId1, geoId2);
+                    std::swap(crv1, crv2);
+                    std::swap(pos1, pos2);
+                    // FIXME: Confirm whether or not this is needed
+                    // *angle = -*angle;
+                }
+                GCS::Point& p3 = Points[getPointId(geoId3, pos3)];
+                auto* partBsp = static_cast<GeomBSplineCurve*>(Geoms[geoId1].geo);
+                double uNear;
+                partBsp->closestParameter(Base::Vector3d(*p3.x, *p3.y, 0.0), uNear);
+                double* pointparam = new double(uNear);
+                Parameters.push_back(pointparam);
+                --ConstraintsCounter;  // Do this just before point-on-object because
+                                       // ConstraintsCounter is increased again before being used
+                addPointOnObjectConstraint(geoId3,
+                                           pos3,
+                                           geoId1,
+                                           pointparam,
+                                           driving);  // increases ConstraintsCounter
+                GCSsys.addConstraintAngleViaPointAndParam(*crv1,
+                                                          *crv2,
+                                                          p,
+                                                          pointparam,
+                                                          angle,
+                                                          tag,
+                                                          driving);
+            }
+        }
+        else {
+            GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
+        }
     }
 
-    GCSsys.addConstraintAngleViaPoint(*crv1, *crv2, p, angle, tag, driving);
     return ConstraintsCounter;
 }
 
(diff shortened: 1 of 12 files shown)


chennes (changes requested):
This is starting to come into shape, nice work. As we discussed in our chat, there are still a few issues to be worked out:

1. Selection order should not matter (right now it does, you get strange results if you select the spline last instead of first).
2. I will probably always want to place my point on the spline first, then add the tangency constraint -- consuming that extra constraint doesn't work right now.
3. There is a phantom extra constraint in my over-constrained error messages now.
4. Attempting tangency constraint without selecting a point needs to pop up a warning message explaining how to resolve the issue.

src/Mod/Sketcher/App/Sketch.cpp
```diff
@@ -2868,6 +2868,11 @@ int Sketch::addTangentConstraint(int geoId1, int geoId2)
             GCSsys.addConstraintTangent(l, a, tag);
             return ConstraintsCounter;
         }
+        else if (Geoms[geoId2].type == BSpline) {
+            Base::Console().Error("Direct tangency constraint between line and B-spline is not "
+                                  "supported. Use tangent-via-point instead.");
```
chennes, line 2873: I think this is more of a "pop-up window" sort of error, not just a message in the report view. Maybe the DWG can weigh in? @obelisk79?

AjinkyaDahale: What's DWG?

obelisk79: @AjinkyaDahale Design Working Group. We try to provide advice to developers for UI/UX related items using some established guidelines. It's a relatively new thing for FreeCAD.

@chennes I agree, adding this to the non-obtrusive alerts would be better. A pop-up window would likely be too disruptive. I'll bring this up with the group on Discord.

chennes: I'm not sure some disruption isn't a _good_ idea here: generally speaking, you're only going to do it wrong once... after that single disruption, you've (hopefully!) learned how to use curve-to-spline tangency -- e.g. you have to tell the system at what point you want it to be tangent, and are unlikely to use it incorrectly again. It could even be one of those dialogs with the "Don't show this again" checkbox.

chennes:
Is it possible to create the point if it doesn't exist? Just generate one as part of the process?

src/Mod/Sketcher/App/Sketch.cpp
obelisk79: Yeah, I concur with a popup. The "Don't show this again" checkbox usually shows for non-critical popups. Like redundant constraint removal etc.

AjinkyaDahale: I think there doesn't need to be a pop-up. Also, someone can ignore the text, tick "don't ask again", and then forever get confused. A nonobstrusive but revisitable error is better. Ideally it should ne accompanied with the constraint being deleted, but I'm still not comfortable with that part of the code.

AjinkyaDahale:
> Is it possible to create the point if it doesn't exist? Just generate one as part of the process?
> 
> ETA: We discussed this in Discord and are concerned about the possibility of it blowing up your design due to weird placement of the generated point, so maybe it's not a good idea...

Unlike other curves, a Bspline can have tangents and perps with the same curve an arbitrary number of times.

AjinkyaDahale:
Sidenote: the order of selection shouldn't matter now.

src/Mod/Sketcher/App/Sketch.cpp
obelisk79: I agree about not including a checkbox. Regarding unobtrusive vs pop-up. I suppose the criticality, and gracefulness of the error state/handling should dictate which is more appropriate. If the constraint is removed and nothing happens negatively to the sketch other than a notification, avoiding the popup is ok. But a popup dialog should appear very infrequently if implemented as such. Both approaches to alerting the user of a problem seem reasonably acceptable regarding the above criteria.

AjinkyaDahale:
Meanwhile, could we have an icon for 3-selection tangent, in the future, it may be a good idea to also "consume" any point-on-object constraint on, say, a line as well when we select it, another curve, and a point. This constraint without the point-on-object doesn't make sense.

AjinkyaDahale:
>     3. There is a phantom extra constraint in my over-constrained error messages now.
> 

This should be fixed now.

obelisk79:
> Meanwhile, could we have an icon for 3-selection tangent, in the future, it may be a good idea to also "consume" any point-on-object constraint on, say, a line as well when we select it, another curve, and a point? This constraint without the point-on-object doesn't make sense.
> 
> @obelisk79 @chennes something for the DWG, maybe.

I'm not sure what a 3 selection tangent icon would look like. Instinctively I'm inclined to suggest not having a separate tangency symbol for this case, even though it is unique 'under the hood', I would need to better understand the need/benefit for exposing this difference to the user.

Additionally, your statement about point-on-object constraints seems to contradict itself.

AjinkyaDahale:
> I'm not sure what a 3 selection tangent icon would look like. Instinctively I'm inclined to suggest not having a separate tangency symbol for this case, even though it is unique 'under the hood', I would need to better understand the need/benefit for exposing this difference to the user.

Try the following:
1. Create a parabola arc and a circular arc. Also a point between them.
2. Add tangent constraint with these 3.

You will see there are 2 or 3 icons now: one for tangent and two for point-on-object. I am saying that these should be combined into one constraint, so that we can't have a tangent constraint where the curves don't even touch. The icon for that (combined) constraint should be distinct from a normal tangent to show there's a point on it. Do you think that might be confusing?

> Additionally, your statement about point-on-object constraints seems to contradict itself.

Could you explain the contradiction? I'm saying a 3-selection tangent (i.e. with curves and a point) do not make sense if the point is not on both curves.

AjinkyaDahale:
Now, tangent-at-point will "consume" any existing point-on-object constraint between the involved point and B-spline(s).

With this, all issues described in https://github.com/FreeCAD/FreeCAD/pull/11853#pullrequestreview-1799603861 should be fixed.

src/Mod/Sketcher/App/Sketch.cpp
adrianinsaval: Sending errors to report view also triggers the non intrusive notifications. The notifications system is a very good idea but it really needs polishing. It would be great if the DWG could have a close look at it, identify the problems and write up some guidelines that devs could follow to make them better.

Sorry for the off topic comment!

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -5598,6 +5602,59 @@ CmdSketcherConstrainPerpendicular::CmdSketcherConstrainPerpendicular()
     ;
 }
 
+bool CmdSketcherConstrainPerpendicular::removeRedundantPointOnObject(SketchObject* Obj,
```
chennes, line 5605: I would very much like to remove the duplication of this method three times.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -5685,35 +5742,45 @@ void CmdSketcherConstrainPerpendicular::activated(int iMsg)
...
-                        Obj)) {// FIXME: it's a good idea to add a check if the sketch is solved
-                    Gui::cmdAppObjectArgs(
-                        selection[0].getObject(),
-                        "addConstraint(Sketcher.Constraint('PointOnObject',%d,%d,%d))",
-                        GeoId3,
-                        static_cast<int>(PosId3),
-                        GeoId1);
-                };
+                        Obj)) {
+                    // FIXME: it's a good idea to add a check if the sketch is solved
```
chennes, line 5773: Is this too difficult to implement right now?

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -5798,14 +5867,14 @@ void CmdSketcherConstrainPerpendicular::activated(int iMsg)
...
 
-            if (geom2 && isBSplineCurve(*geom2)) {
-                // unsupported until normal to B-spline at any point implemented.
-                Gui::TranslatedUserWarning(
-                    Obj,
-                    QObject::tr("Wrong selection"),
-                    QObject::tr("Perpendicular to B-spline edge currently unsupported."));
-                return;
-            }
+            // if (geom2 && isBSplineCurve(*geom2)) {
```
chennes, line 5870: Can we delete instead of commenting? git will remember it for us if we need it again (here and elsewhere the same technique appears).

src/Mod/Sketcher/App/planegcs/Geo.cpp
```diff
@@ -773,6 +773,189 @@ DeriVector2 BSpline::CalculateNormal(const Point& p, const double* derivparam) c
     return ret;
 }
 
+DeriVector2 BSpline::CalculateNormal(const double* param, const double* derivparam) const
```
chennes, line 776: This method has grown long enough that I'd suggest refactoring it immediately, rather than doing that as a separate PR later.

src/Mod/Sketcher/App/planegcs/Constraints.cpp
```diff
@@ -2650,6 +2650,241 @@ double ConstraintAngleViaPoint::grad(double* param)
...
+        deriv += -1.0;
+    }
+    DeriVector2 n1 = crv1->CalculateNormal(cparam1(), param);
+    DeriVector2 n2 = crv2->CalculateNormal(cparam2(), param);
+    deriv -= ((-n1.dx) * n1.y / pow(n1.length(), 2) + n1.dy * n1.x / pow(n1.length(), 2));
+    deriv += ((-n2.dx) * n2.y / pow(n2.length(), 2) + n2.dy * n2.x / pow(n2.length(), 2));
+
+
+// use numeric for testing
+#if 0
```
chennes, line 2869: Is there any reason to keep this now? If so, can we implement it as gtest code instead?

src/Mod/Sketcher/App/Sketch.cpp
```diff
@@ -3146,19 +3170,116 @@ int Sketch::addAngleAtPointConstraint(int geoId1,
...
+                                                              angle,
+                                                              tag,
+                                                              driving);
+            }
+            else {
+                if (Geoms[geoId1].type != BSpline) {
+                    std::swap(geoId1, geoId2);
+                    std::swap(crv1, crv2);
+                    std::swap(pos1, pos2);
+                    // FIXME: Confirm whether or not this is needed
```
chennes, line 3253: Leaving this here would make me nervous: can you confirm?

chennes (commented):
Although this part of the code is not my area of expertise, I have some general comments on your modifications. I don't view these as merge-preventers, just suggestions for improvement that you can implement or not as you see fit.

Also, I don't much like abbreviations in variable and function names. Even for English speakers I think it makes the code more opaque, and for those for whom English is not their primary language I think it makes it much harder to understand. I know this is endemic in Sketcher (and FreeCAD as a whole), but I hate to see new code doing it.

Finally, I'd love to see some automated tests implemented in the gtest framework for this new feature.

src/Mod/Sketcher/App/Sketch.cpp
AjinkyaDahale: This one was about the angle: is there benefit in making the angles direction specific? That is, should this be the (counterclockwise) angle between the first and second selected curves? In that case, the same angle with selection reversed should look as if the older selection was provided with a negative angle. But it currently seems like this angle is already adjusted elsewhere: I tried different selections and same angle, and the shape remains same.

This may have use specifically when negative angles have some parametric use, though I can't think of any right now.

src/Mod/Sketcher/App/planegcs/Geo.cpp
AjinkyaDahale: Sorry I might have forgotten: have we discussed refactoring this?

src/Mod/Sketcher/Gui/CommandConstraints.cpp
AjinkyaDahale: The check on whether this sketch is solved? I'm actually not even sure I added this line. Seems something coming from pre-commit.

src/Mod/Sketcher/Gui/CommandConstraints.cpp
AjinkyaDahale: Should be done.

AjinkyaDahale:
> Although this part of the code is not my area of expertise, I have some general comments on your modifications. I don't view these as merge-preventers, just suggestions for improvement that you can implement or not as you see fit.

Going through them as I can. 
 
> Also, I don't much like abbreviations in variable and function names. Even for English speakers I think it makes the code more opaque, and for those for whom English is not their primary language I think it makes it much harder to understand. I know this is endemic in Sketcher (and FreeCAD as a whole), but I hate to see new code doing it.

I'd be OK with changing some of the less obvious abbreviations in the Sketcher side. However, many on the GCS side may be mathematical terms.

AjinkyaDahale:
Found some issue with the angle rendition.
[Screencast from 2024-01-21 23-09-35.webm](https://github.com/FreeCAD/FreeCAD/assets/4316933/341ae2be-6cb7-4371-a4df-1c63b86cb49f)

Could use help from someone acquainted with the drawing aspects. @PaddleStroke maybe?

src/Mod/Sketcher/App/planegcs/Constraints.cpp
AjinkyaDahale: Removed. I don't know when this was last added. The gtests you asked for should cover this.

src/Mod/Sketcher/App/Sketch.cpp
AjinkyaDahale: I think removing this would require us to first properly document what the angle constraint means for us, not just for B-splines, but for other curves as well.

src/Mod/Sketcher/App/planegcs/Geo.cpp
chennes: No, I don't think so. But with a cognitive complexity of 48 it's definitely needed.

AjinkyaDahale: What is that? And how do you suggest it can be reduced?

PaddleStroke:
@AjinkyaDahale if the angle switch like this, I think the start and end pos (of the line or of the spline) are maybe switching

PaddleStroke:
I don't know what I am looking at. Did you implement angle constraint between angle and bspline ?
I looked at your PR and you are not editing EditModeConstraintManager.cpp which you likely should.
Look at line 1268 'case Angle' of processConstraint.

Here you want to handle the special case of line + bspline by setting the startAngle and range of the constraint.

chennes (approved):

chennes:
I will try to remember to look at this tonight when I have access to a Windows machine.

chennes:
In order to use the classes in Geo.h in the tests, they all have to be defined with the `SketcherExport` macro, e.g.
```
class SketcherExport DeriVector2
{
    ...
```

chennes:
From a feature standpoint I think this is good to merge, leaving as potential future work the auto-generation of the needed point (there is some undo complication in that generation, I believe). I propose to partially squash this PR: I'm planning on merging the following commits into one, unless you object @AjinkyaDahale:
[[Sketcher] Expose general tangency with B-splines to Sketcher](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/0c58c27827a17f791f6da022f8962606d4c9f171)
[0c58c27](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/0c58c27827a17f791f6da022f8962606d4c9f171)
[[Sketcher] Handle some corner cases in AngleViaPoint](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/e8dd265022f4da494fbedd11eef80f07e8317f46)
[e8dd265](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/e8dd265022f4da494fbedd11eef80f07e8317f46)
[[Sketcher] Avoid redundant constraints with B-splines...](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/af71d01d539185d70cd7a18c4b0e0261f4ea8466) 
[af71d01](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/af71d01d539185d70cd7a18c4b0e0261f4ea8466)
[[Sketcher] Add pre-commit changes](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/8868a3da794a04b93aa28f89ced639dd17796829)
[8868a3d](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/8868a3da794a04b93aa28f89ced639dd17796829)
[[Sketcher][planegcs] Fix direction issue with B-spline tangents.](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/fcfafa56442552226fa8fa2c603596c90377fcbf) 
[fcfafa5](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/fcfafa56442552226fa8fa2c603596c90377fcbf)
[[Sketcher] Do not allow 2-selection tangent with B-spline](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/7c9db4d1d11381facb194ea0f1e4075b629d9de8) 
[7c9db4d](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/7c9db4d1d11381facb194ea0f1e4075b629d9de8)
[[Sketcher] Fix malformed constraint when B-spline is seleced second](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/a43bba1d05939fb8b440757064ad947bad685ee2) 
[a43bba1](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/a43bba1d05939fb8b440757064ad947bad685ee2)
[[Sketcher] Fix wrong number for B-Spline tangent on redundancy](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/1e520a2eac42ae91517bd67f52579e14a67563e2)
[1e520a2](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/1e520a2eac42ae91517bd67f52579e14a67563e2)
[[Sketcher] Remove existing point-on-object in some redundant cases](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/d1161d563c94c6892280140d5014c139edbd0a9d) 
[d1161d5](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/d1161d563c94c6892280140d5014c139edbd0a9d)
[[planegcs] Remove some numerical testing](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/48c31dfb501cd181486b4ac91db8b65d107c0f9a) 
[48c31df](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/48c31dfb501cd181486b4ac91db8b65d107c0f9a)

AjinkyaDahale:
@chennes I would suggest keeping the `planegcs` commits separate from, and placed before, the `Sketcher` ones. IIRC there is already a project where planegcs is made into a library by itself.

Of course apart from that it's ideal if the whole thing compiles at each commit.

AjinkyaDahale:
> In order to use the classes in Geo.h in the tests, they all have to be defined with the `SketcherExport` macro, e.g.
> 
> ```
> class SketcherExport DeriVector2
> {
>     ...
> ```

These commits were apparently lost when I force pushed. Do you have them on your machine?

AjinkyaDahale:
> I propose to partially squash this PR: I'm planning on merging the following commits into one, unless you object @AjinkyaDahale: ...[[planegcs] Remove some numerical testing](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/48c31dfb501cd181486b4ac91db8b65d107c0f9a) [48c31df](https://github.com/FreeCAD/FreeCAD/pull/11853/commits/48c31dfb501cd181486b4ac91db8b65d107c0f9a)

It might be a good idea to keep this commit by itself so in the future if someone intends to make these numerical tests in gtest, this could be useful.

chennes:
> These commits were apparently lost when I force pushed. Do you have them on your machine?

Not worth trying to find them, all I did was add the appropriate export to all the classes in that file.

AjinkyaDahale:
> > These commits were apparently lost when I force pushed. Do you have them on your machine?
> 
> Not worth trying to find them, all I did was add the appropriate export to all the classes in that file.

I think I saw two of them. Anyway, let's see if this succeeds.

AjinkyaDahale:
@chennes if this works should I put the `_USES_MATH_DEFINES` under an `ifdef`?

chennes:
Yeah, something like
```
#ifdef FC_OS_WIN32
#define _USE_MATH_DEFINES
#endif
```

Syres916:
@AjinkyaDahale sorry to raise this after merging but all those new error messages in `addTangentConstraint`, should they not be translated if they are important user feedback?

PR #11857, opened 2023-12-29
[Gui] Dimension Widget add Tooltip...

Syres916:
...and a warning message box for the change from the widget handling system units to now handling document units.

This is following https://github.com/FreeCAD/FreeCAD/pull/9521

The Status Bar Wiki page needs some useful words adding so I'll have a go at writing a paragraph and maybe a screenshot or two of the list and message box even though I think it's self explanatory.

### DIFF
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -166,12 +166,15 @@ public:
     {
         setFlat(true);
         setText(qApp->translate("Gui::MainWindow", "Dimension"));
+        setToolTip(
+            qApp->translate("Gui::MainWindow", "Select the Unit Schema for the current document"));
         setMinimumWidth(120);
 
-        //create the action buttons
+        // create the action buttons
         auto* menu = new QMenu(this);
         auto* actionGrp = new QActionGroup(menu);
         int num = static_cast<int>(Base::UnitSystem::NumUnitSystemTypes);
+
         for (int i = 0; i < num; i++) {
             QAction* action = menu->addAction(QStringLiteral("UnitSchema%1").arg(i));
             actionGrp->addAction(action);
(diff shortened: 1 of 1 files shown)


src/Gui/MainWindow.cpp
```diff
@@ -166,29 +166,84 @@ class DimensionWidget : public QPushButton, WindowParameter
...
         QObject::connect(actionGrp, &QActionGroup::triggered, this, [this](QAction* action) {
-            int userSchema = action->data().toInt();
-            setUserSchema(userSchema);
+            bool onceRun = App::GetApplication()
+                               .GetParameterGroupByPath("User parameter:BaseApp/Preferences/Units")
+                               ->GetBool("DoNotShowDocUnitsMsgBox", false);
+            if (!onceRun) {
+                QMessageBox msgBox;
+                QAbstractButton* myYesButton =
+                    msgBox.addButton(tr("OK, but do not show this again"), QMessageBox::YesRole);
```
chennes, line 191: IMO a more standard way of doing this is to use a `QCheckBox` on the dialog that says "Don't show again", and only use the standard OK buttons. See [QMessageBox::setCheckBox](https://doc.qt.io/qt-5/qmessagebox.html#setCheckBox).

Syres916: I'll look at this tomorrow, my goal is always to minimize mouse clicks not because of time saving but RSI, it's only one click saved but imagine if we could do that in 15% of the whole code??

sliptonic: Moving to draft pending an update.  @Syres916  please move out of draft when you're ready.

Syres916:
Obviously the FreeCAD Maintainers don't have RSI as a priority compared to a 'standard' developed in the late 80's. As a RSI suffered I cannot knowingly add any extra unnecessary mouse clicks to the code, it would be hypocritical, therefore closing PR.

PR #11868, opened 2024-01-01
Arch: Arch_Window fix and improve handling of hosts

Roy-043:
Fixes a problem introduced with a previous PR of mine (#7591). Hosts would only be touched if that property of the window was modified. But the hosts of a window should also update if other properties of the window change.

Additionally a width or height change of a window now touches the hosts on the subsequent shape change.


Roy-043:
This PR causes a recursive recompute if windows are cloned. I am investigating ATM.

PR #11870, opened 2024-01-01
Arch: Arch_Project avoid ghosts of objects in Group

Roy-043:


yorikvanhavre:
This is an interesting solution, I think we could escalate it to some generic function to be used by other objects (NativeIFC objects for ex).

Roy-043:
OK. What other objects are affected? BuildingParts, for example, do not have this problem as they have that little axis cross.

yorikvanhavre:
Basically the NativeIFC object: https://github.com/yorikvanhavre/FreeCAD-NativeIFC/blob/main/ifc_objects.py#L26
It represents an IFC object. However, an IFC object can have an own shape or not, and be an aggregate of other IFC objects. In case an object has no own shape but is an aggregate of other objects, in FreeCAD it should behave as a group (the children are visible and selectable indicidually). Up to now I solved this by adding a very small dumb shape: https://github.com/yorikvanhavre/FreeCAD-NativeIFC/blob/main/ifc_generator.py#L59 I should try this solution instead...

Roy-043:
I can't reproduce the ghost issue in a BIM IFC project. I have:
1. Commented out line 61 in ifc_generator.py: `# obj.Shape = Part.makeBox(1, 1, 1)`.
2. Disabled `removeDisplaymodeChildNodes` in ArchProject.py.
3. Restarted and created a project with a structure as in the attached image. ifc-project-no-ghost
4. Selecting the Project object shows no ghosts.

```
OS: Windows 8 build 9600
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35554 (Git)
Build type: Release
Branch: main
Hash: 8e2ab9f76888ac4da971f9d5069cec7bcc2b572f
Installed mods: 
  * BIM 2021.12.0
```

PR #11883, opened 2024-01-02
Fix build with precompiled headers

wavexx:
This PR adds some missing includes when attempting to build with `-DFREECAD_USE_PCH=ON`.

Each commit is split by module. The most notable one is moving inclusion of QtAll after glext, so that we can set GL_GLEXT_PROTOTYPES before it's too late.

### DIFF
--- a/src/App/PreCompiled.h
+++ b/src/App/PreCompiled.h
@@ -49,6 +49,7 @@
 #include <csignal>
 #include <cstdio>
 #include <ctime>
+#include <cfloat>
 
 #ifdef FC_OS_WIN32
 # include <crtdbg.h>
--- a/src/Mod/Mesh/Gui/PreCompiled.h
+++ b/src/Mod/Mesh/Gui/PreCompiled.h
@@ -44,6 +44,7 @@
 
 // standard
 #include <ios>
+#include <cfloat>
 
 // STL
 #include <algorithm>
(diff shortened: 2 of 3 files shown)


chennes:
Thanks for the PR. What sort of system are you working on? I could use some help with a PCH problem we're having with on Windows with the Spreadsheet module.

wavexx:
I'm currently on debian/unstable, I generally build with both gcc 13/clang 17 (depending on lto settings).
I didn't actually measure if the pch build is any faster, I just turned it on for a spin :)

chennes:
Got it, thanks! I'll ask a couple of the other maintainers to give this a test before merging.

wavexx:
Do I actually need to update the PR (merge main or rebase), or this will be done by who merges?

chennes:
It will be taken care of during merge, you don't need to do anything.

CalligaroV:
I'll try to compile on Linux.

CalligaroV:
@chennes @wavexx I tried to compile this branch on my PC with Linux Mint 21 Cinnamon using `-DFREECAD_USE_PCH=ON` on an empty directory and I had no issues, I can compile and run.

I see that most of the times using `-DFREECAD_USE_PCH=ON` triggers `add_definitions(-D_PreComp_)` and other macros related to MSVC.
I can't test the macros that use MSVC but I guess that as the CI isn't failing there aren't issues with those cases.

Therefor if this check was the last one needed before merging then I would say that we can go on with that.

PR #11884, opened 2024-01-02
Sketcher: contextual right click context menu based on selection

### LINKED ISSUE #11818: [Problem] Sketcher: right click menu in 3d view

### Problem description

While sketching the right mouse button does not open the right click menu when it is hovering geometry in the 3d view (lines, points or constraints.
If the cursor is next to the geometry, the right click action works.
Additionally, the right click menu is the same, regardless if geometry or constraints are selected and is very huge with some options not necessary in the Sketcher WB.
If possible, the right click menu could include the options which are also available from the right click on list items in the task dialog (constraints and geometry elements), maybe even based on the selection.
Currently some functions in the RMB menu do even not work in sketcher, e.g. Fit selection or two tools which work contrary: fit all and center sketch.

Quick demo:

https://github.com/FreeCAD/FreeCAD/assets/6246609/00a24502-6fb8-40f9-aa35-fa2782469752

Here are the current items in the right click menus:
grafik

For first ideas I propose that sketcher tools that create or modify the selection should be included which are currently missing (like copy, transform, scale, rotate, mirror) maybe grouped/nested (create geo, modify geo, transform geo, view and selection options).
If possible these could change based on the selection,e.g.: 
- If geometry only is selected, e.g. select two points: present point based constraints like coincident. select two lines: parallel, perpendicular, equal constraints or distance or modify tools like fillet them... kind of like the dimension tool to suggest the possible options.
- If no geometry is selected, the right click menu should focus more on the commands for creating geometry (expand / list all create commands) and centering/aligning the sketch or enable the clipping plane.
- If constraints only are selected, options like enable/disable them, select their corresponding geometry.

### Full version info

```shell
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35449 (Git)
Build type: Release
Branch: main
Hash: f3fa2ba9c9b1999e5c32d3ae544481b6e17553bb
Installed mods: 
  * 3DfindIT 1.2.0
  * BIM 2021.12.0
  * CfdOF 1.24.9
  * CurvedShapes 1.0.5
  * Curves 0.6.19
  * Defeaturing 1.2.1
  * fasteners 0.4.75
  * FEMbyGEN 2.0.0
  * freecad.gears 1.0.0
  * OpenDark 2023.12.17
  * sheetmetal 0.3.13
  * woodworking 0.21.1.33694
```

### Subproject(s) affected?

Sketcher

### Anything else?

_No response_

### Code of Conduct

maxwxyz:
fixes #11818 
This PR introduces a *contextual* right click context menu in the Sketcher WB which shows different commands based on the selection of the user (none, one or multiple geometry elements, or constraints).
The current context menu is static and is overloaded with commands which are either greyed out or not working in Sketcher.

New:

https://github.com/FreeCAD/FreeCAD/assets/6246609/49426bf8-56a2-46f5-bd6c-cd768a9386d5

Old:
grafik

### DIFF
--- a/src/Mod/Sketcher/Gui/CommandConstraints.cpp
+++ b/src/Mod/Sketcher/Gui/CommandConstraints.cpp
@@ -1167,14 +1167,14 @@ public:
 
         addCommand("Sketcher_Dimension");
         addCommand(); //separator
-        addCommand("Sketcher_ConstrainLock");
         addCommand("Sketcher_ConstrainDistanceX");
         addCommand("Sketcher_ConstrainDistanceY");
         addCommand("Sketcher_ConstrainDistance");
+        addCommand("Sketcher_ConstrainRadiam");
         addCommand("Sketcher_ConstrainRadius");
         addCommand("Sketcher_ConstrainDiameter");
-        addCommand("Sketcher_ConstrainRadiam");
         addCommand("Sketcher_ConstrainAngle");
+        addCommand("Sketcher_ConstrainLock");
     }
 
     void updateAction(int mode) override
@@ -1217,6 +1217,37 @@ public:
     }
 
     const char* className() const override { return "CmdSketcherCompDimensionTools"; }
+};
+
+// Comp for constrain tools =============================================
+
+class CmdSketcherCompConstrainTools : public Gui::GroupCommand
+{
+public:
+    CmdSketcherCompConstrainTools()
+        : GroupCommand("Sketcher_CompConstrainTools")
+    {
+        sAppModule = "Sketcher";
+        sGroup = "Sketcher";
+        sMenuText = QT_TR_NOOP("Constrain");
+        sToolTipText = QT_TR_NOOP("Constrain tools.");
+        sWhatsThis = "Sketcher_CompConstrainTools";
+        sStatusTip = sToolTipText;
+        eType = ForEdit;
+
+        setCheckable(false);
+        setRememberLast(false);
+
+        addCommand("Sketcher_ConstrainCoincidentUnified");
+        addCommand("Sketcher_ConstrainHorVer");
+        addCommand("Sketcher_ConstrainParallel");
+        addCommand("Sketcher_ConstrainPerpendicular");
+        addCommand("Sketcher_ConstrainTangent");
+        addCommand("Sketcher_ConstrainEqual");
+        addCommand("Sketcher_ConstrainSymmetric");
+        addCommand("Sketcher_ConstrainBlock");
+    }
+    const char* className() const override { return "CmdSketcherCompConstrainTools"; }
 };
 
 // Dimension tool =======================================================
@@ -9897,5 +9928,6 @@ void CreateSketcherCommandsConstraints()
     rcCmdMgr.addCommand(new CmdSketcherToggleDrivingConstraint());
     rcCmdMgr.addCommand(new CmdSketcherToggleActiveConstraint());
     rcCmdMgr.addCommand(new CmdSketcherCompDimensionTools());
+    rcCmdMgr.addCommand(new CmdSketcherCompConstrainTools());
 }
 // clang-format on
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -1210,106 +1210,14 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                     // delegate to handler whether to quit or do otherwise
                     sketchHandler->pressRightButton(Base::Vector2d(x, y));
                     return true;
-                case STATUS_NONE: {
-                    // A right click shouldn't change the Edit Mode
-                    if (preselection.isPreselectPointValid()) {
-                        return true;
-                    }
-                    else if (preselection.isPreselectCurveValid()) {
-                        return true;
-                    }
-                    else if (!preselection.PreselectConstraintSet.empty()) {
-                        return true;
-                    }
-                    else {
-                        Gui::MenuItem geom;
-                        geom.setCommand("Sketcher geoms");
-                        geom << "Sketcher_CreatePoint"
-                             << "Sketcher_CreateArc"
-                             << "Sketcher_Create3PointArc"
-                             << "Sketcher_CreateCircle"
-                             << "Sketcher_Create3PointCircle"
-                             << "Sketcher_CreateLine"
-                             << "Sketcher_CreatePolyline"
-                             << "Sketcher_CreateRectangle"
-                             << "Sketcher_CreateHexagon"
-                             << "Sketcher_CreateFillet"
-                             << "Sketcher_CreatePointFillet"
-                             << "Sketcher_Trimming"
-                             << "Sketcher_Extend"
-                             << "Sketcher_External"
-                             << "Sketcher_ToggleConstruction"
-                             /*<< "Sketcher_CreateText"*/
-                             /*<< "Sketcher_CreateDraftLine"*/
-                             << "Separator";
-
-                        Gui::Application::Instance->setupContextMenu("View", &geom);
-                        // Create the Context Menu using the Main View Qt Widget
-                        QMenu contextMenu(viewer->getGLWidget());
-                        Gui::MenuManager::getInstance()->setupContextMenu(&geom, contextMenu);
-                        contextMenu.exec(QCursor::pos());
-
-                        return true;
-                    }
-                }
+                case STATUS_NONE:
                 case STATUS_SELECT_Point:
-                    break;
-                case STATUS_SELECT_Edge: {
-                    Gui::MenuItem geom;
-                    geom.setCommand("Sketcher constraints");
-                    geom << "Sketcher_ConstrainVertical"
-                         << "Sketcher_ConstrainHorizontal";
-
-                    // Gets a selection vector
-                    std::vector<Gui::SelectionObject> selection = Gui::Selection().getSelectionEx();
-
-                    bool rightClickOnSelectedLine = false;
-
-                    /*
-                     * Add Multiple Line Constraints to the menu
-                     */
-                    // only one sketch with its subelements are allowed to be selected
-                    if (selection.size() == 1) {
-                        // get the needed lists and objects
-                        const std::vector<std::string>& SubNames = selection[0].getSubNames();
-
-                        // Two Objects are selected
-                        if (SubNames.size() == 2) {
-                            // go through the selected subelements
-                            for (std::vector<std::string>::const_iterator it = SubNames.begin();
-                                 it != SubNames.end();
-                                 ++it) {
-
-                                // If the object selected is of type edge
-                                if (it->size() > 4 && it->substr(0, 4) == "Edge") {
-                                    // Get the index of the object selected
-                                    int GeoId = std::atoi(it->substr(4, 4000).c_str()) - 1;
-                                    if (preselection.PreselectCurve == GeoId)
-                                        rightClickOnSelectedLine = true;
-                                }
-                                else {
-                                    // The selection is not exclusively edges
-                                    rightClickOnSelectedLine = false;
-                                }
-                            }// End of Iteration
-                        }
-                    }
-
-                    if (rightClickOnSelectedLine) {
-                        geom << "Sketcher_ConstrainParallel"
-                             << "Sketcher_ConstrainPerpendicular";
-                    }
-
-                    Gui::Application::Instance->setupContextMenu("View", &geom);
-                    // Create the Context Menu using the Main View Qt Widget
-                    QMenu contextMenu(viewer->getGLWidget());
-                    Gui::MenuManager::getInstance()->setupContextMenu(&geom, contextMenu);
-                    contextMenu.exec(QCursor::pos());
-
+                case STATUS_SELECT_Edge:
+                case STATUS_SELECT_Cross:
+                case STATUS_SELECT_Constraint: {
+                    generateContextMenu();
                     return true;
                 }
-                case STATUS_SELECT_Cross:
-                case STATUS_SELECT_Constraint:
                 case STATUS_SKETCH_DragPoint:
                 case STATUS_SKETCH_DragCurve:
                 case STATUS_SKETCH_DragConstraint:
@@ -3898,5 +3806,170 @@ void ViewProviderSketch::executeOnSelectionPointSet(
 bool ViewProviderSketch::isInEditMode() const
 {
     return editCoinManager != nullptr;
+}
+void ViewProviderSketch::generateContextMenu()
+{
+    int selectedEdges = 0;
+    int selectedLines = 0;
+    int selectedConics = 0;
+    int selectedPoints = 0;
+    int selectedConstraints = 0;
+
+    Gui::MenuItem menu;
+    menu.setCommand("Sketcher context");
+
+    std::vector<Gui::SelectionObject> selection =
+        Gui::Selection().getSelectionEx(0, Sketcher::SketchObject::getClassTypeId());
+
+    // if something is selected, count different elements in the current selection
+    if (selection.size() > 0) {
+        const std::vector<std::string> SubNames = selection[0].getSubNames();
+
+        for (auto& name : SubNames) {
+            if (name.substr(0, 4) == "Edge") {
+                ++selectedEdges;
+
+                int geoId = std::atoi(name.substr(4, 4000).c_str()) - 1;
+                if (geoId >= 0) {
+                    const Part::Geometry* geo = getSketchObject()->getGeometry(geoId);
+                    if (isLineSegment(*geo)) {
+                        ++selectedLines;
+                    }
+                    else {
+                        ++selectedConics;
+                    }
+                }
+            }
+            else if (name.substr(0, 4) == "Vert") {
+                ++selectedPoints;
+            }
+            else if (name.substr(0, 4) == "Cons") {
+                ++selectedConstraints;
+            }
+        }
+        // build context menu items depending on the selection
+        if (selectedEdges >= 1 && selectedPoints == 0) {
+            menu << "Sketcher_Dimension";
+            if (selectedConics == 0) {
+                menu << "Sketcher_ConstrainHorVer"
+                     << "Sketcher_ConstrainVertical"
+                     << "Sketcher_ConstrainHorizontal";
+
+                if (selectedLines > 1) {
+                    menu << "Sketcher_ConstrainParallel";
+
+                    if (selectedLines == 2) {
+                        menu << "Sketcher_ConstrainPerpendicular"
+                             << "Sketcher_ConstrainTangent";
+                    }
+
+                    menu << "Sketcher_ConstrainEqual";
+                }
+                menu << "Sketcher_ConstrainBlock";
+            }
+            else if (selectedConics > 1 && selectedLines == 0) {
+                menu << "Sketcher_ConstrainCoincidentUnified"
+                     << "Sketcher_ConstrainTangent"
+                     << "Sketcher_ConstrainEqual";
+            }
+            else if (selectedConics == 1 && selectedLines == 1) {
+                menu << "Sketcher_ConstrainTangent";
+            }
+        }
+        else if (selectedEdges == 1 && selectedPoints >= 1) {
+            menu << "Sketcher_Dimension";
+            if (selectedConics == 0) {
+                menu << "Sketcher_ConstrainCoincidentUnified"
+                     << "Sketcher_ConstrainHorVer"
+                     << "Sketcher_ConstrainVertical"
+                     << "Sketcher_ConstrainHorizontal";
+                if (selectedPoints == 2) {
+                    menu << "Sketcher_ConstrainSymmetric";
+                }
+            }
+            else {
+                menu << "Sketcher_ConstrainCoincidentUnified"
+                     << "Sketcher_ConstrainPerpendicular"
+                     << "Sketcher_ConstrainTangent";
+            }
+        }
+        else if (selectedEdges == 0 && selectedPoints >= 1) {
+            menu << "Sketcher_Dimension";
+
+            if (selectedPoints > 1) {
+                menu << "Sketcher_ConstrainCoincidentUnified"
+                     << "Sketcher_ConstrainHorVer"
+                     << "Sketcher_ConstrainVertical"
+                     << "Sketcher_ConstrainHorizontal";
+            }
+        }
+        else if (selectedLines >= 1 && selectedPoints >= 1) {
+            menu << "Sketcher_Dimension"
+                 << "Sketcher_ConstrainHorVer"
+                 << "Sketcher_ConstrainVertical"
+                 << "Sketcher_ConstrainHorizontal";
+        }
+        // context menu if only constraints are selected
+        else if (selectedConstraints >= 1) {
+            menu << "Sketcher_ToggleDrivingConstraint"
+                 << "Sketcher_ToggleActiveConstraint"
+                 << "Sketcher_SelectElementsAssociatedWithConstraints"
+                 << "Separator"
+                 << "Std_Delete";
+        }
+        // add the rest of the context menu if geometry is selected
+        if (selectedPoints != 0 || selectedEdges != 0) {
+            menu << "Separator"
+                 << "Sketcher_ToggleConstruction"
+                 << "Separator"
+                 << "Sketcher_CreatePointFillet"
+                 << "Sketcher_Trimming"
+                 << "Sketcher_Extend"
+                 << "Sketcher_Offset"
+                 << "Sketcher_Rotate"
+                 << "Separator"
+                 << "Sketcher_CompDimensionTools"
+                 << "Sketcher_CompConstrainTools"
+                 << "Sketcher_SelectConstraints"
+                 << "Separator"
+                 << "Std_Delete";
+        }
+    }
+    // context menu without a selection
+    else {
+        menu << "Sketcher_ViewSketch"
+             << "Sketcher_ViewSection"
+             << "Std_ViewFitAll"
+             << "Separator"
+             << "Sketcher_CreatePoint"
+             << "Sketcher_CreatePolyline"
+             << "Sketcher_CreateArc"
+             << "Sketcher_CreateCircle"
+             << "Sketcher_CreateRectangle"
+             << "Sketcher_CreateHexagon"
+             << "Sketcher_CreateBSpline"
+             << "Separator"
+             << "Sketcher_ToggleConstruction"
+             << "Separator"
+             << "Sketcher_CreatePointFillet"
+             << "Sketcher_Trimming"
+             << "Sketcher_Extend"
+             << "Separator"
+             << "Sketcher_External"
+             << "Separator"
+             << "Sketcher_CompDimensionTools"
+             << "Sketcher_CompConstrainTools"
+             << "Separator"
+             << "Sketcher_DeleteAllGeometry"
+             << "Sketcher_DeleteAllConstraints"
+             << "Separator"
+             << "Sketcher_LeaveSketch";
+    }
+    // create context menu
+    Gui::Application::Instance->setupContextMenu("Sketch", &menu);
+    QMenu contextMenu(
+        qobject_cast<Gui::View3DInventor*>(this->getActiveView())->getViewer()->getGLWidget());
+    Gui::MenuManager::getInstance()->setupContextMenu(&menu, contextMenu);
+    contextMenu.exec(QCursor::pos());
 }
 // clang-format on
--- a/src/Doc/CONTRIBUTORS
+++ b/src/Doc/CONTRIBUTORS
@@ -129,6 +129,7 @@ Marosh
 Masaya Ootsuki
 Mateusz Skowroński (f3nix)
 Mattis M
+Max Wilfinger (maxwxyz)
 mdinger
 Meme2704
 Michael Hansen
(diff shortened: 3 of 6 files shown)


maxwxyz:
No selection:
grafik

Selection of a geometry (in this example one line):
grafik

Selection of constraints:
grafik

PaddleStroke:
A first comment on the functionality : 
- 'Fit all' : Wouldn't it make more sense to bring it up under 'view section' 
- 'Draw style' : I would remove it it seems unecessary here.
- 'Selected associated geo' : For consistency with the other menu maybe it would be best to move it to after the toggles.

PaddleStroke (commented):
I have not finished reviewing. Tomorrow I will check the core of the feature

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1212,45 +1212,165 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
+
+                    Gui::MenuItem geom;
+                    geom.setCommand("Sketcher context");
+                    // Gets a selection vector
+                    std::vector<Gui::SelectionObject> selection = Gui::Selection().getSelectionEx();
+                    // if something is selected
+                    if (selection.size() > 0) {
+                        const std::vector<std::string> SubNames = selection[0].getSubNames();
+
+                    if (SubNames.size() == 1) {
```
PaddleStroke, line 1229: From here you are ignoring the indentation from the previous if. This makes the code hard to understand as one can't visually see where the if starts and ends. Besides it is misleading to the reader. Please fix the indentation.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1212,45 +1212,165 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
+                    bool noSelection = true;
+                    bool selectPoint = false;
+                    bool selectLine = false;
+                    bool selectConstraint = false;
+                    bool selectMultiple = false;
+
+                    Gui::MenuItem geom;
+                    geom.setCommand("Sketcher context");
+                    // Gets a selection vector
+                    std::vector<Gui::SelectionObject> selection = Gui::Selection().getSelectionEx();
```
PaddleStroke, line 1224: Here you should use : 
`    std::vector<Gui::SelectionObject> selection = Gui::Selection().getSelectionEx(0, Sketcher::SketchObject::getClassTypeId());`
To make sure that you are selecting only sketches. As it stands I guess that if you select a solid edge from outside the sketch, you get the special menu for edges.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1212,45 +1212,165 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
+                        if (SubNames.begin()->substr(0, 4) == "Edge")
+                            selectLine = true;
+                        else if (SubNames.begin()->substr(0, 4) == "Vert")
+                            selectPoint = true;
+                        else if (SubNames.begin()->substr(0, 4) == "Cons")
+                            selectConstraint = true;
                     }
-                    else if (preselection.isPreselectCurveValid()) {
-                        return true;
+                    else if (SubNames.size() > 1) {
```
PaddleStroke, line 1238: There is no reason to separate the cases where `SubNames.size() == 1` and `SubNames.size() > 1`. The case `== 1 `is a special case of `> 1` that is supported by the for loop. I guess that if you remove the first case completely it will work directly.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1212,45 +1212,165 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
+                        noSelection = false;
+                        selectMultiple = true;
+                        int selectedLines = 0;
+                        int selectedPoints = 0;
+                        int selectedConstraints = 0;
+                        // go through the selected subelements
+                        for (std::vector<std::string>::const_iterator it = SubNames.begin();
```
PaddleStroke, line 1245: Here you can use modern for loop, they look much better and are easier to understand : 

                        for (auto name : SubNames) {
                            if (name.substr(0, 4) == "Edge") {
                                ++selectedLines;
                            }
                            else if (name.substr(0, 4) == "Vert"){
                                ++selectedPoints;
                            }
                            else if (name.substr(0, 4) == "Cons"){
                                ++selectedConstraints;
                            }
                        }

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1212,45 +1212,165 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
+                                ++selectedLines;
+                            }
+                            else if (it->substr(0, 4) == "Vert"){
+                                ++selectedPoints;
+                            }
+                            else if (it->substr(0, 4) == "Cons"){
+                                ++selectedConstraints;
+                            }
+                        }// End of Iteration
+                        if (selectedLines > 1 && selectedPoints == 0 && selectedConstraints == 0)
```
PaddleStroke, line 1258: I don't see the value of using the boolean selectLine, selectMultiple etc. (besides selectMultiple makes you have to use the special case of `== 1`)
Instead use directly the conditions you are using now in the menu creations
`if (selectedLines > 1 && selectedPoints == 0 && selectedConstraints == 0)`

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1212,45 +1212,165 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
                     }
+
+                    if (!noSelection){
+                        if (selectConstraint) {
+                        geom << "Sketcher_SelectElementsAssociatedWithConstraints"
+                             << "Sketcher_ToggleDrivingConstraint"
+                             << "Sketcher_ToggleActiveConstraint"
+                             << "Separator"
+                             << "Std_Delete";
+                             goto menucreation;
```
PaddleStroke, line 1280: goto is seldomly used. If you can do without it I think it is better as people are not very used to it.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1212,45 +1212,165 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
+                        }// End of Iteration
+                        if (selectedLines > 1 && selectedPoints == 0 && selectedConstraints == 0)
+                            selectLine = true;
+                        else if (selectedLines == 0 && selectedPoints > 1 && selectedConstraints == 0) {
+                            selectPoint = true;
+                        }
+                        else if (selectedLines == 0 && selectedPoints == 0 && selectedConstraints > 1){
+                            selectConstraint = true;
+                        }
+                        else if (selectedLines == 1 && selectedPoints > 0 && selectedConstraints == 0){
```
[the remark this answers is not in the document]
PaddleStroke: I think you want to ignore constraints for cases where geometry are selected : 
`if (selectedLines >= 1 && selectedPoints == 0) {`

PaddleStroke:
Overall very good, it seems you are learning fast and your contribution is smart. Thanks for contributing ! 

By the way, to ease your life, here is the code block with the modifications I offered. I have not tested so please do.

```
                case STATUS_NONE: {
                    int selectedEdges = 0;
                    int selectedPoints = 0;
                    int selectedConstraints = 0;

                    Gui::MenuItem geom;
                    geom.setCommand("Sketcher context");

                    std::vector<Gui::SelectionObject> selection = Gui::Selection().getSelectionEx(0, Sketcher::SketchObject::getClassTypeId());

                    if (selection.size() > 0) {
                        const std::vector<std::string> SubNames = selection[0].getSubNames();

                        for (auto name : SubNames) {
                            if (name.substr(0, 4) == "Edge") {
                                ++selectedEdges;
                            }
                            else if (name.substr(0, 4) == "Vert") {
                                ++selectedPoints;
                            }
                            else if (name.substr(0, 4) == "Cons") {
                                ++selectedConstraints;
                            }
                        }

                        if (selectedEdges >= 1 && selectedPoints == 0) {
                            // Here you want to make special cases for line vs arcs vs circle and so on.
                            geom << "Sketcher_Dimension"
                                << "Sketcher_ConstrainHorVer"
                                << "Sketcher_ConstrainVertical"
                                << "Sketcher_ConstrainHorizontal";

                            if (selectedEdges > 1) {
                                geom << "Sketcher_ConstrainParallel"
                                    << "Sketcher_ConstrainPerpendicular"
                                    << "Sketcher_ConstrainTangent"
                                    << "Sketcher_ConstrainEqual";
                            }
                        }
                        else if (selectedEdges == 1 && selectedPoints >= 1) {
                            // Here you want to make special cases for line vs arcs vs circle and so on.
                            geom << "Sketcher_Dimension"
                                << "Sketcher_ConstrainCoincidentUnified"
                                << "Sketcher_ConstrainHorVer"
                                << "Sketcher_ConstrainVertical"
                                << "Sketcher_ConstrainHorizontal";
                        }
                        else if (selectedEdges == 0 && selectedPoints >= 1) {
                             geom << "Sketcher_Dimension";

                             if (selectedPoints > 1) {
                                geom << "Sketcher_ConstrainCoincidentUnified"
                                    << "Sketcher_ConstrainHorVer"
                                    << "Sketcher_ConstrainVertical"
                                    << "Sketcher_ConstrainHorizontal";
                             }
                        }
                        else if (selectedConstraints >= 1) {
                            geom << "Sketcher_SelectElementsAssociatedWithConstraints"
                                << "Sketcher_ToggleDrivingConstraint"
                                << "Sketcher_ToggleActiveConstraint"
                                << "Separator"
                                << "Std_Delete";
                        }

                        geom << "Separator"
                            << "Sketcher_ToggleConstruction"
                            << "Separator"
                            << "Sketcher_CreatePointFillet"
                            << "Sketcher_Trimming"
                            << "Sketcher_Extend"
                            << "Separator"
                            << "Sketcher_CompDimensionTools"
                            << "Sketcher_CompConstrainTools"
                            << "Sketcher_SelectConstraints"
                            << "Separator"
                            << "Std_Delete";
                    }
                    else {  //without selection
                        geom << "Sketcher_ViewSketch"
                            << "Sketcher_ViewSection"
                            << "Std_ViewFitAll"
                            << "Separator"
                            << "Sketcher_CreatePoint"
                            << "Sketcher_CreatePolyline"
                            << "Sketcher_CreateArc"
                            << "Sketcher_CreateCircle"
                            << "Sketcher_CreateRectangle"
                            << "Sketcher_CreateHexagon"
                            << "Sketcher_CreateBSpline"
                            << "Separator"
                            << "Sketcher_ToggleConstruction"
                            << "Separator"
                            << "Sketcher_CreatePointFillet"
                            << "Sketcher_Trimming"
                            << "Sketcher_Extend"
                            << "Separator"
                            << "Sketcher_External"
                            << "Separator"
                            << "Sketcher_CompDimensionTools"
                            << "Sketcher_CompConstrainTools"
                            << "Separator"
                            << "Sketcher_DeleteAllGeometry"
                            << "Sketcher_DeleteAllConstraints"
                            << "Separator"
                            << "Sketcher_LeaveSketch";
                    }

                    Gui::Application::Instance->setupContextMenu("Sketch", &geom);
                    // Create the Context Menu using the Main View Qt Widget
                    QMenu contextMenu(viewer->getGLWidget());
                    Gui::MenuManager::getInstance()->setupContextMenu(&geom, contextMenu);
                    contextMenu.exec(QCursor::pos());

                    return true;
                }
```

To make special cases of line vs circle, have a look at how to get the geometry from the name in sketcher_dimension tool (see line 1510 of CommandConstraints.cpp, function handleInitialSelection)

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1212,45 +1212,165 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
+                            selectPoint = true;
+                        else if (SubNames.begin()->substr(0, 4) == "Cons")
+                            selectConstraint = true;
                     }
-                    else if (preselection.isPreselectCurveValid()) {
-                        return true;
+                    else if (SubNames.size() > 1) {
+                        noSelection = false;
+                        selectMultiple = true;
+                        int selectedLines = 0;
```
PaddleStroke, line 1241: This should be named `selectedEdges` as its not only lines

PaddleStroke:
@maxwxyz I have just edited my previosu block of code to integrate counting edges and conics

PaddleStroke:
If you want finer control than line vs conics, you can add more variables : 

```
                                    if (isLineSegment(*geo)) {
                                        ++selectedLines;
                                    }
                                    else if (isCircle(*geo)) {
                                        ++selectedCircles;
                                    }
                                    ...
```

maxwxyz:
> If you want finer control than line vs conics, you can add more variables :

Will look into this

PaddleStroke:
Also the CI macOS builds are failing : 
image

You can get more information looking into the details (try ctrl-F for error in the logs).

Having read the code, I would venture a guess that this is caused by the 'goto' as it is the only unusual thing I have seen. So if you pickup my code block push it to see if it solves the CI failure before trying to debug it.

maxwxyz:
> Also the CI macOS build is failing : image
> 
> You can get more information looking into the details (try ctrl-F for error in the logs).
> 
> Having read the code, I would venture a guess that this is caused by the 'goto' as it is the only unusual thing I have seen. So if you pickup my code block push it to see if it solves the CI failure before trying to debug it.

Ok will try to implement conics first. On my last PR (only the icon change of .svg files) the Mac build also failed...

PaddleStroke:
Ok maybe the macos build failure is unrelated to your PR then.

Btw one last thing, this feature is adding a lot of lines of code to the mouseButtonPressed function. So please make a new function to put all this new code in, for example : `ViewProviderSketch::generateMenuBasedOnSelection()`

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
maxwxyz: I need one goto if only constraints are selected, because the rest of the context menu makes no sense for constraints only but it works.

maxwxyz:
@PaddleStroke Thanks for reviewing.
- I've created a separate function and called this function from the previous spot and also replaced the weird cases where currently a different menu was called (left click on edge, hold and then right click) with the new function.
- Context menu was adjusted to present valid constraints for cases if conics are in the selection and refined for other cases
- Your new tools for offset and rotate were added as well when geometry is selected (translate/array has to be included when merged).

off-topic: Fillet/Trim/Extend edge does currently not work with a pre-selection. Is this already in progress or should I create an issue?

maxwxyz:
I don't know why some checks are failing now, I don't get it from the details as I've not touched the code there (Mac failing because of Ondsel solver)

Syres916:
@maxwxyz do you actually need those additional lines in ViewProviderSketch.h ?

maxwxyz:
@Syres916 @PaddleStroke I've adjusted it again, should be fine now. I have no idea why the GitHub actions are failing now, can someone help me please?

Syres916:
@maxwxyz it's going to be a few hours before I have more than one CPU core free as I'm doing a Git Bisect from 2021 to trace a bug. If it's still not solved then I'll compile your PR just to make sure I get the same error as being reported in the CI builds.

maxwxyz:
@Syres916 I've built successfully with conda on Windows and get no warnings or errors concerning the changed files.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3899,4 +3816,153 @@ bool ViewProviderSketch::isInEditMode() const
...
+                    << "Sketcher_ConstrainHorizontal";
+                }
+        }
+        else if (selectedConstraints >= 1) {
+            geom << "Sketcher_ToggleDrivingConstraint"
+                << "Sketcher_ToggleActiveConstraint"
+                << "Sketcher_SelectElementsAssociatedWithConstraints"
+                << "Separator"
+                << "Std_Delete";
+                goto menucreation;
```
kadet1090, line 3914: goto is considered bad practice - try to write the code without it. In that particular case it is harmless but we should not introduce it without a good reason. 

Here you could extract code that adds items to `geom` to another function like `fillMenuEntries(geom)` and then you could simply replace the goto with return statement.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3899,4 +3816,153 @@ bool ViewProviderSketch::isInEditMode() const
...
+        else if (selectedEdges == 1 && selectedPoints >= 1) {
+            // Here you want to make special cases for line vs arcs vs circle and so on.
+            geom << "Sketcher_Dimension";
+            if (selectedConics == 0) {
+                geom << "Sketcher_ConstrainCoincidentUnified"
+                    << "Sketcher_ConstrainHorVer"
+                    << "Sketcher_ConstrainVertical"
+                    << "Sketcher_ConstrainHorizontal";
+                if (selectedPoints == 2) {
+                geom << "Sketcher_ConstrainSymmetric";
```
kadet1090, line 3889: Try to reformat new code using clang-format

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3899,4 +3816,153 @@ bool ViewProviderSketch::isInEditMode() const
...
     return editCoinManager != nullptr;
 }
+void ViewProviderSketch::generateContextMenu() {
+    int selectedEdges = 0;
+    int selectedLines = 0;
+    int selectedConics = 0;
+    int selectedPoints = 0;
+    int selectedConstraints = 0;
+
+    Gui::MenuItem geom;
```
kadet1090, line 3826: I'm not sure if this name is fitting here, maybe `menu` would be better? `geom` comes from `geometry` an this variable does not represent geometry. I know that it is copied from another place, but it probably would be wise to change it now.

@PaddleStroke you have more experience with codebase - so if this is some convention in the code that I'm simply not aware of please intervene

kadet1090 (changes requested):

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
maxwxyz: changed

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
PaddleStroke: Or alternatively you can add a boolean 'addRestOfMenu = true' and set it to false when its constraints selected.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
PaddleStroke: The name is not great indeed. I did not paid attention because this was like this before. But yes name 'menu' would be better.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
maxwxyz: I've now done another if statement: if no constraints are in the selection, only then add the rest. Hope this is ok too?

Syres916:
> @Syres916 I've built successfully with conda on Windows and get no warnings or errors concerning the changed files.

There's always some 'specifics' with Linux that Windows compiling won't highlight, that's why the CI's are across platforms though I don't have a lot of confidence in the Mac one at present.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
PaddleStroke: Or if you are using visual studio, you can also cut the whole paragraph and paste it back and visual studio should fix the indentation automatically on paste.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
PaddleStroke: No because the other cases ignore constraints. So perhaps you have 2 lines and 1 constraint, in which case the constraint is ignored. So instead of testing if there are no constraints, you need to test if all the rest is empty (edges and points both empty).

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3899,4 +3816,150 @@ bool ViewProviderSketch::isInEditMode() const
...
+                }
+        }
+        else if (selectedConstraints >= 1) {
+            menu << "Sketcher_ToggleDrivingConstraint"
+                << "Sketcher_ToggleActiveConstraint"
+                << "Sketcher_SelectElementsAssociatedWithConstraints"
+                << "Separator"
+                << "Std_Delete";
+        }
+        if (selectedConstraints == 0){
```
PaddleStroke, line 3913: here use this instead so that it covers (1 line 1 constraint) for example : 
`if (selectedPoints != 0 or selectedLines !=0){`

maxwxyz: changed it, will push when the failing GitHub actions are resolved or should I push every change?

PaddleStroke: I suggest squashing all commits and force push.

maxwxyz: should I rebase from current main first? How to squash & force push together?

PaddleStroke: I'll give you a hand let's meet : 
https://meet.google.com/fax-ukcg-wkq

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3899,4 +3816,156 @@ bool ViewProviderSketch::isInEditMode() const
...
+    Gui::MenuItem menu;
+    menu.setCommand("Sketcher context");
+
+    std::vector<Gui::SelectionObject> selection = Gui::Selection().getSelectionEx(0, Sketcher::SketchObject::getClassTypeId());
+
+    // if something is selected, count different elements in the current selection
+    if (selection.size() > 0) {
+        const std::vector<std::string> SubNames = selection[0].getSubNames();
+
+        for (auto name : SubNames) {
```
PaddleStroke, line 3835: There is a Lint warning here, see below my message.

> Missing reference in range-for with non trivial type (std::basic_string<char>) [-Wclazy-range-loop-reference]

I think you need to change this to : `for (auto& name : SubNames) {` ie add a `&`

PaddleStroke:
Please fix my last comment (and squash the change), then compile and make sure everything works as expected. Then I think it's ready to merge @chennes.

maxwxyz:
@PaddleStroke @chennes done and thanks!

PaddleStroke:
Ah sorry one thing I missed : 
```
switch (Mode) {
                case STATUS_SKETCH_UseHandler:
                    // delegate to handler whether to quit or do otherwise
                    sketchHandler->pressRightButton(Base::Vector2d(x, y));
                    return true;
                case STATUS_NONE: {
                    generateContextMenu();
                    return true;
                }
                case STATUS_SELECT_Point:{
                    generateContextMenu();
                    return true;
                }
                case STATUS_SELECT_Edge: {
                    generateContextMenu();
                    return true;
                }
                case STATUS_SELECT_Cross:
                case STATUS_SELECT_Constraint: {
                    generateContextMenu();
                    return true;
                }
                case STATUS_SKETCH_DragPoint:
                case STATUS_SKETCH_DragCurve:
                case STATUS_SKETCH_DragConstraint:
                case STATUS_SKETCH_StartRubberBand:
                case STATUS_SKETCH_UseRubberBand:
                    break;
            }
```

Can be simplified to : 
```
switch (Mode) {
                case STATUS_SKETCH_UseHandler:
                    // delegate to handler whether to quit or do otherwise
                    sketchHandler->pressRightButton(Base::Vector2d(x, y));
                    return true;
                case STATUS_NONE: 
                case STATUS_SELECT_Point:
                case STATUS_SELECT_Edge:
                case STATUS_SELECT_Cross:
                case STATUS_SELECT_Constraint: {
                    generateContextMenu();
                    return true;
                }
                case STATUS_SKETCH_DragPoint:
                case STATUS_SKETCH_DragCurve:
                case STATUS_SKETCH_DragConstraint:
                case STATUS_SKETCH_StartRubberBand:
                case STATUS_SKETCH_UseRubberBand:
                    break;
            }
```

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3899,4 +3816,163 @@ bool ViewProviderSketch::isInEditMode() const
...
+                menu << "Sketcher_ConstrainCoincidentUnified"
+                    << "Sketcher_ConstrainTangent"
+                    << "Sketcher_ConstrainEqual";
+            }
+            else if (selectedConics == 1 && selectedLines == 1){
+                menu << "Sketcher_ConstrainTangent";
+            }
+            if (selectedLines > 1 && selectedConics == 0) {
+                menu << "Sketcher_ConstrainParallel";
+                    if (selectedLines ==2) {
```
kadet1090, line 3876: There are still few formatting mishaps - you can try to change your settings of vscode to enable reformatting on save:

https://github.com/microsoft/vscode/issues/44075#issuecomment-678231273

maxwxyz: oh it worked. the whole file was in //canlang-format off

maxwxyz:
all done

kadet1090 (approved):

src/Mod/Sketcher/Gui/CommandConstraints.cpp
```diff
@@ -1219,6 +1219,75 @@ class CmdSketcherCompDimensionTools : public Gui::GroupCommand
...
+        addCommand("Sketcher_ConstrainCoincidentUnified");
+        addCommand("Sketcher_ConstrainHorVer");
+        addCommand("Sketcher_ConstrainParallel");
+        addCommand("Sketcher_ConstrainPerpendicular");
+        addCommand("Sketcher_ConstrainTangent");
+        addCommand("Sketcher_ConstrainEqual");
+        addCommand("Sketcher_ConstrainSymmetric");
+        addCommand("Sketcher_ConstrainBlock");
+    }
+    void updateAction(int mode) override
```
PaddleStroke, line 1250: Ah how it seems I miss this as well.
The updateAction is not needed for this actionGroup because the icons do not need to change based on the driving status. You can remove this whole function.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3899,4 +3807,166 @@ bool ViewProviderSketch::isInEditMode() const
...
+            }
+            else if (selectedConics > 1 && selectedLines == 0) {
+                menu << "Sketcher_ConstrainCoincidentUnified"
+                     << "Sketcher_ConstrainTangent"
+                     << "Sketcher_ConstrainEqual";
+            }
+            else if (selectedConics == 1 && selectedLines == 1) {
+                menu << "Sketcher_ConstrainTangent";
+            }
+            if (selectedLines > 1 && selectedConics == 0) {
```
PaddleStroke, line 3867: This if statement belongs within the `if (selectedConics == 0) {` block line 3853

PaddleStroke (approved):

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -3899,4 +3807,169 @@ bool ViewProviderSketch::isInEditMode() const
 {
     return editCoinManager != nullptr;
 }
+void ViewProviderSketch::generateContextMenu()
```
chennes, line 3810: I'm OK with merging this code as-is, and I know there is a LONG history in FreeCAD of methods that look like this, but IMO it should really be something like eight individual methods, so would look something like:
```
void ViewProviderSketch::generateContextMenu()
{
    // I'm using structured bindings here, but you could write countSelection in a number of different ways
    // to achieve the same basic idea...
    const auto [selectedEdges, selectedLines, selectedConics, selectedPoints, selectedConstraints] = countSelection();
    if (selectedEdges >= 1 && selectedPoints == 0) {
        setContextMenuForEdges(menu, selectedConics, selectedLines);
    } else if (selectedEdges == 1 && selectedPoints >= 1) {
        setContextMenuForOneEdgeAndPoints(menu, selectedConics, selectedPoints);
    } 
    // and so forth, for each of your top-level conditionals -- the functions can be local 
    // to the CPP file, in an anonymous namespace, since they don't need access to `this`
}
```

maxwxyz: @chennes shall I redo it or do it like this for future PRs?

PaddleStroke: But then you are creating a lot of functions which are cluttering VPSketch. Even if only the cpp file.
So the solution might be to create a class in another file to handle this. But it seems a bit overkill. And also Max is a new c++ contributor asking him that wouldn't be easy.
So perhaps we can keep like this for now? This function is long but only because of the menu names taking many lines not because of complexity.

kadet1090: I'd also say that this is fine. Every branch of the if consists basically of one statement - which is pretty long but it still is one statement. Introducing more functions into that case would be harder to maintain as every added or removed scenario will require adding or removing a function which won't be reusable anyways as it is so specific.

PaddleStroke:
Congrats on a nice feature @maxwxyz

PR #11885, opened 2024-01-02
set INSTALL_TO_SITEPACKAGES to ON by default

looooo:
and add some modules to the freecad-namespace (freecad.part, freecad.partdesign, freecad.sketcher)

this targets the new import structure from freecad. In case we want to fully move to the new structure we can place the shared libraries also in the freecad-namespace.


looooo:
I would like to move to something like:
`from freecad import part` instead of `import Part`

In the meantime I have moved the gears-workbench to this structure. To me this seems to be much nicer.

What is currently possible:
`from freecad import app`
`from freecad import gui`

But I have no idea how the latter is working. gui is also different when imported from python directly or inside freecad. Inside freecad it is the same as FreeCADGui outside it is the Ext/gui directory. But it should be always the same as FreeCADGui in my eyes.

src/Ext/freecad/CMakeLists.txt
```diff
@@ -38,6 +38,9 @@ endif()
 INSTALL(
     FILES
         ${NAMESPACE_INIT}
+        part.py
+        partdesign.py
+        sketcher.py
```
adrianinsaval, line 43: should we add one of these for each wb?

yorikvanhavre: I understand those are mostly lower-case versions? Indeed that would deserve a discussion... While it is more pythonic, it might also spread different versions of python scripts, and then why not fix the original one instead..

adrianinsaval: more than that I think the idea is that they can get imported directly from python with something like
```python
from freecad import part
```
rather than requiring two lines
```python
import freecad
import Part
```

adrianinsaval:
See https://github.com/conda-forge/freecad-feedstock/pull/118

looooo:
>While it is more pythonic, it might also spread different versions of python scripts, and then why not fix the original one instead..

Yes I also think it should be done directly by renaming the .so files and placing them in the site-package dir. But this would cause a lot of refactoring work. So I think the better way is a slow transition: 1. Give the opportunity to use the `from freecad import module` alternative. 2. deprecate the old imports. 3. fully move to the new import syntax. I would support the old syntax for freecad 1.x and maybe move to the suggested syntax for a freecad 2.0. So in my mind there is no need to rush here.

PR #11896, opened 2024-01-04
Toponaming shapeprotector

Pesc0:
- currently in working state
- replaces manual re-definition of every method with inheritance (not recommended by RT in case OCC updates TopoDS_Shape, but we deemed it rare enough)
- would be nice to somehow template the the wrapper functions, but I gave up on that
- Something needs to be done with the move/moved/locate/located situation, it's a spaghetti bolognese

### DIFF
--- a/src/Mod/Part/App/TopoShape.h
+++ b/src/Mod/Part/App/TopoShape.h
@@ -372,6 +372,12 @@ public:
         return TopoShape().makeTransform(*this,trsf,op,copy);
     }
 
+    /** Move the shape to a new location
+     *
+     * @param loc: location
+     *
+     * The location is applied in addition to any current transformation of the shape
+     */
     void move(const TopLoc_Location &loc) {
         _Shape.Move(loc);
     }
@@ -387,6 +393,70 @@ public:
         TopoShape ret(*this);
         ret._Shape.Move(loc);
         return ret;
+    }
+    /** Move and/or rotate the shape
+     *
+     * @param trsf: OCCT transformation (must not have scale)
+     *
+     * The transformation is applied in addition to any current transformation
+     * of the shape
+     */
+    void move(const gp_Trsf &trsf) {
+        move(_Shape, trsf);
+    }
+    /** Return a new transformed shape
+     *
+     * @param trsf: OCCT transformation (must not have scale)
+     *
+     * @return Return a shallow copy of the shape transformed to the new
+     *         location that is applied in addition to any current
+     *         transformation of the shape
+     */
+    TopoShape moved(const gp_Trsf &trsf) const {
+        return moved(_Shape, trsf);
+    }
+    /** Set a new location for the shape
+     *
+     * @param loc: shape location
+     *
+     * Any previous location of the shape is discarded before applying the
+     * input location
+     */
+    void locate(const TopLoc_Location &loc) {
+        _Shape.Location(loc);
+    }
+    /** ReturnSet a new location for the shape
+     *
+     * @param loc: shape location
+     *
+     * @return Return a shallow copy the shape in a new location. Any previous
+     *         location of the shape is discarded before applying the input
+     *         location
+     */
+    TopoShape located(const TopLoc_Location &loc) const {
+        TopoShape ret(*this);
+        ret._Shape.Location(loc);
+        return ret;
+    }
+    /** Set a new transformation for the shape
+     *
+     * @param trsf: OCCT transformation (must not have scale)
+     *
+     * Any previous transformation of the shape is discarded before applying
+     * the input transformation
+     */
+    void locate(const gp_Trsf &trsf) {
+        located(_Shape, trsf);
+    }
+    /** Set a new transformation for the shape
+     *
+     * @param trsf: OCCT transformation (must not have scale)
+     *
+     * Any previous transformation of the shape is discarded before applying
+     * the input transformation.
+     */
+    TopoShape located(const gp_Trsf &trsf) const {
+        return located(_Shape, trsf);
     }
 
     static TopoDS_Shape& move(TopoDS_Shape& tds, const TopLoc_Location& loc);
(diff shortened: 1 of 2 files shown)


src/Mod/Part/App/TopoShape.h
```diff
@@ -459,8 +529,74 @@ class PartExport TopoShape : public Data::ComplexGeoData
...
-private:
-    TopoDS_Shape _Shape;
+    class ShapeProtector : public TopoDS_Shape
+    {
+    public:
+        using TopoDS_Shape::TopoDS_Shape;
+        using TopoDS_Shape::operator=;
+
+        ShapeProtector() = default;
+        ShapeProtector(const TopoDS_Shape& shape) : TopoDS_Shape(shape) {}
```
chennes, line 539: Is this not `explicit` intentionally? I can't recall if we need the automatic cast.

CalligaroV:
@Pesc0, @chennes, is this PR linked to a GH issue of the Toponaming Integration Project?
If so, can you link it so that it is visible also from the Project Dashboard?

TIA! (and happy new year 2024, even if late :sweat_smile:)

chennes:
@CalligaroV I just added it to the project

CalligaroV:
Many thanks @chennes !

If I understood correctly from the meeting, by now this PR only needs to be reviewed and after that other modifications will focus on the last 2 points of the original post of this PR, right?

chennes:
We probably won't worry about trying to use templates to simplify this code, but we probably will try to do something about the names of the move/moved/locate/located methods. Suggestions are welcome!

src/Mod/Part/App/TopoShape.h
Pesc0: I would say yes, this makes shapeprotector behave as if it was a topods_shape transparently.
Is this not right?

Pesc0:
huh CI is failing, great
I spoke to soon when I said everything is working

Pesc0:
mac and windows fail to build with `Location is an ambiguous call`
I thought I had fixed that, but I am on linux, and it turns out that linux builds but then segfaults in tests

CalligaroV:
@chennes, @Pesc0, I did some research about the points of your previous posts.

@chennes, by looking at the FC codebase I would suggest these names for the methods you mentioned:
- `move` and `moved` could be renamed `incrementalTransform` and `incrementalTransformed` as "move" means a transformation applied on top of the current one, thus "incremental"
-  `locate` and `located` could be renamed `absoluteTransform` and `absoluteTransformed` as "locate" means a transformation applied discarding all the previous ones, therefor I would call them "absolute" as I think those refer to the origin

@Pesc0, about the segfaults, looking at the CI reports in one case the error raises after calling `Nullify`, in the other after calling `makeFace`. This makes me think, is it possible that `makeFace` is called without a valid `_Shape` object?

About the `Location is an ambiguous call` issue, changing the name of the methods  `Location` and `Move` according to my suggestions above probably will probably fix it.
Otherwise, is it possible for those methods to call directly `TopoDS_Shape` methods instead of `TopoShape` methods as, for example, `Reverse` and `Complement` does?

Hope these suggestions help!

chennes:
As far as I can see, `owner` is never getting set -- so the first access of it segfaults.

CalligaroV:
I was guessing the same thing.

I tried adding `_Shape.owner = this;` inside every `TopoShape` constructor and I can compile and run without segfaults, or at least I haven't segfaults in my tests (creating a pad/pocket from a sketch).

RT's code is a little bit different, `ShapeProtector`'s constructors use `TopoShape` instead of `TopoDS_Shape` objects and this allows to initialize `TopoShape & master` (which is similar to `owner` in this PR) with `ShapeProtector`'s constructors, but I guess that both codes behaves the same way.

Regardless for that, I'm not 100% sure that my modifications are correct but at least, as wrote before, with those I can compile and run FreeCAD without segfaults

CalligaroV:
An update:

I've been able to compile and run FC editing `TopoShape.h` instead of `TopoShape.cpp` so that only one file is modified.
The changes I made are the following.

BEFORE:
```
        ShapeProtector() = default;
        ShapeProtector(const TopoDS_Shape& shape) : TopoDS_Shape(shape) {}
        ShapeProtector(TopoDS_Shape &&shape): TopoDS_Shape(std::move(shape)) {}

        // ...

    ShapeProtector _Shape;
````

AFTER:
```
        ShapeProtector(TopoShape& shape) : owner(&shape) {}
        ShapeProtector(const TopoDS_Shape& shape) : TopoDS_Shape(shape), owner((TopoShape*)(&shape)) {}
        ShapeProtector(TopoDS_Shape &&shape) : TopoDS_Shape(std::move(shape)), owner((TopoShape*)(std::move(&shape))) {}

        //...

    ShapeProtector _Shape {*this};
```

As for my previous post, I'm not sure this is the best way to solve the segfault issues but the outcome is the same of modifying `TopoShape.cpp`

Hope this helps!

chennes:
@CalligaroV can you give this a look? I didn't rename those methods, but I did get it compiling on the various platforms (there seems to be some name lookup issue when trying to us the `using` mechanism for the overload).

CalligaroV:
@chennes Thanks for the update!

I just compiled and tried the last force-pushed commit and everything looks working correctly.
I simply tried the same steps of my previous comments on a new file / PD Body but this time I tried also to modify an existing file.
In both cases I haven't found issues.

Do you have some particular tests in mind or code changes that you want me to check more in depth?

chennes:
As long as nothing segfaults, the only thing to review is the code itself.

CalligaroV:
@chennes, I took a look at the code and haven't spotted anything strange.
I was a little worried that `Tag` and `Hasher` could have been used without initializing them but I wasn't able to find cases where this could happen.

This PR is already merged and at the moment I don't see reasons for further modifications.

Thanks!

PR #11901, opened 2024-01-05
Arch: Arch_Window: wrong value for Frame property

Roy-043:
The Frame property controls the extrusion of the window components. It should receive the w2 instead of the h1 value.


yorikvanhavre:
Indeed. Thanks!

PR #11907, opened 2024-01-05
[Sketcher] Draw B-spline when creating in edit mode

AjinkyaDahale:
Gives a better idea of the shape of the B-spline than just control polygon.

This PR is being supported through an FPA grant.


AjinkyaDahale:
Marking as ready for review. No major changes so this should be safe enough to merge. Long B-splines may need better treatment.

Issue #11605, opened 2023-12-01
[Feature request] Sketcher: double-click to select a chain of geometry


pierreporte:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem description

In a complex sketch, where there are several closed ant non-closed shapes, that can overlap each other, it can be difficult to select one of them. The rectangular selection is too broad and it is often necessary to select individual segments, which can be quite long for complex shapes.

The problem exists in every CAD program (design or drafting). QCAD provides a very easy way to select only the desired shape: you double-click on one of the segment, and the connected geometry is automatically selected in a chain of reaction.

### Full version info

```shell
OS: Ubuntu 23.04 (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35249 (Git) AppImage
Build type: Release
Branch: main
Hash: 94ca51b7799b2ac60a2a56d7e6e753bb97a73671
Python 3.10.13, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.6, OCC 7.6.3
Locale: French/France (fr_FR)
Installed mods: 
  * Render 2023.8.18
  * OpenDark 2023.10.30
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

M4xTheMan:
Are you aware of the filters in Sketcher WB? It's good practice to keep your Sketches simple by the way. It's not always possible but most of the time you shouldn't have such problems.

pierreporte:
@M4xTheMan Do you mean the following?

![image](https://github.com/FreeCAD/FreeCAD/assets/2029853/24b2f311-43f7-4d6e-a530-417b1e065ec3)

It allows you to narrow down the list to various types of geometry. If I know how to use it correctly, it doesn’t filter shapes. Say you have a slot (two lines and to arcs), you can filter arcs and lines in general but not the ones precisely you want to edit.

I know that sketches should be kept simple. We are talking about the case where it cannot.

M4xTheMan:
You could select the elements which you'd like to filter atm.

pierreporte:
Yes, so you can’t filter by shape. You need to select it manually, segment by segment, one way or another.

Granted, it’s not the most important feature that FreeCAD needs at the moment.

maxwxyz:
[...]

PR #11925, opened 2024-01-07
[Sketcher] Define double click on edges : Select the whole wire.

PaddleStroke:
Fixes https://github.com/FreeCAD/FreeCAD/issues/11605

Double click on an edge select all the connected geometries.

### DIFF
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -937,8 +937,9 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                             -16000,
                             -16000);// certainly far away from any clickable place, to avoid
                                     // re-trigger of double-click if next click happens fast.
-
-                        Mode = STATUS_NONE;
+                        if (Mode != STATUS_SELECT_Wire) {
+                            Mode = STATUS_NONE;
+                        }
                     }
                     else {
                         DoubleClick::prvClickTime = SbTime::getTimeOfDay();
@@ -1031,6 +1032,11 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                     }
                     Mode = STATUS_NONE;
                     return true;
+                case STATUS_SELECT_Wire: {
+                    toggleWireSelelection(preselection.PreselectCurve);
+                    Mode = STATUS_NONE;
+                    return true;
+                }
                 case STATUS_SELECT_Constraint:
                     if (pp) {
                         auto sels = preselection.PreselectConstraintSet;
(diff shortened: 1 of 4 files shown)


pierreporte:
Thank you for this.

What was your approach for wires that have geometry connected of a different type (construction or not)?

PaddleStroke:
No distinction. If geometries have their end are coincident, even if without constraints then they are selected.

PaddleStroke:
@adrianinsaval Do you perhaps have some time to review this PR?

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1370,6 +1378,70 @@ void ViewProviderSketch::editDoubleClicked()
     }
 }
 
+void ViewProviderSketch::toggleWireSelelection(int clickedGeoId)
+{
+    Sketcher::SketchObject* obj = getSketchObject();
+
+    const Part::Geometry* geo1 = obj->getGeometry(clickedGeoId);
+    if (isPoint(*geo1) || isCircle(*geo1) || isEllipse(*geo1)) {
```
adrianinsaval, line 1386: should this check for closed loop bsplines too or is that overkill?

adrianinsaval:
What do you guys think the behavior should be when there's a fork?
image

As I see it with this implementation which loop gets selected depends on the order of creation of the edges and there's no way for the user to choose which gets selected. I don't have a problem with this behavior as I don't think I would use this functionality that much.

pierreporte:
I would expect the whole thing to be selected, regardless of the starting edge. QCAD works this way.

adrianinsaval:
Nevermind, I misinterpreted the code the first time, All connected edges will be selected.

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
PaddleStroke: That would make sense.

PaddleStroke:
@adrianinsaval I have added the condition for periodic bspline

pierreporte:
@PaddleStroke I just found an issue with this: #16198.

maxwxyz:
@PaddleStroke this Sketcher feature is not working anymore, see #16198, can confirm
```
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 1.0.0RC2.38806 (Git)
Build type: Release
Branch: (HEAD detached at 1.0rc2)
Hash: 3d63fc6c2f665a8d5e6468845a419bcac80756c7
Python 3.11.9, Qt 5.15.13, Coin 4.0.3, Vtk 9.2.6, OCC 7.7.2
Stylesheet/Theme/QtStyle: FreeCAD Dark.qss/FreeCAD Dark/Fusion
Installed mods: 
  * CfdOF 1.27.10
  * Curves 0.6.47
  * dodo 1.0.1
  * fasteners 0.5.29
  * fasteners.backup1721936159.7542229 0.5.24 (Disabled)
  * freecad.gears 1.3.0
  * Manipulator 1.5.7
  * OpenTheme 2024.9.1
  * OpticsWorkbench 1.0.24
  * Rocket 3.3.0
  * sheetmetal 0.4.26
  * Silk 0.1.5
  * Silk.backup1727166062.7227929 0.1.5 (Disabled)

```

PaddleStroke:
I confirm the feature is broken. If you can open an issue about this? https://github.com/FreeCAD/FreeCAD/issues/16198 is closed and was specific to slot but the feature is totally broken apparantly.

PR #11926, opened 2024-01-07
[Sketcher] Redo Improve angle & radius constraint rendering (#11507)

FlachyJoe:
Two commits of #11507 were lost, certainly when solving conflicts for #9166.


src/Mod/Sketcher/Gui/EditModeConstraintCoinManager.cpp
```diff
@@ -1311,27 +1330,42 @@ void EditModeConstraintCoinManager::processConstraints(const GeoListFacade& geol
...
+                                    helperStartAngle = angle;
+                                    helperRange = startAngle - angle;
+                                }
+                                else {
+                                    if (angle < endAngle) {
+                                        angle += 2 * M_PI;
+                                    }
+                                    helperStartAngle = endAngle;
+                                    helperRange = angle - endAngle;
+                                }
```
benj5378, line 1361: Would it be suitable to move this its own function?

PaddleStroke: Perhaps, though this PR is about restoring what was removed by error. So I'm not sure refactoring belongs to this PR.

chennes:
Thanks for hunting this down.

maxwxyz:
@FlachyJoe @PaddleStroke not sure if these were present before but the helpers for angles and distances on arcs are not present:
grafik

PaddleStroke:
No they were not implemented yet.
Only line-line angle and radius has helper implemented

Issue #11676, opened 2023-12-07
[Problem] Tree: Drag&Drop hectic. Many unexpected behaviors.


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Origins should not accept drag/drop of their elements. (unless I'm missing some specific workflow?)

- Issue 1 :
Trying to drop something on the origin object gives an error and the object is not dropped, which correct, though the error message seems wrong:

> Bad object "Unnamed#Sphere"(Part::Sphere) linked to the Origin of "Unnamed#Part004"

However if you try to drop a thorus inside the origin, then the thorus take the place of the Z-axis004 that gets kicked out.

![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/60f89670-625c-42e7-8da3-ec78fe8068bb)

Issue 2 : 
If you try to drop the origin object in a group for example, then it duplicate and a copy seems to get in the group.

![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/62be55ea-f3fb-427a-93f1-66c7b749479e)

The drag and drop behavior appears to be hectic and weird. I could creates a lot of issues with all the different weird things that are happening. And that's not even mentioning the Ctrl and Alt modifiers.

From this video : https://www.youtube.com/watch?v=f1raOT4thEM&ab_channel=LeiZheng I seem to understand that this behavior is actually a feature. But it is not behaving correctly.

The default behavior from the drag&drop should be the normal drag&drop behavior : either the object moves if it can, or does not move at all. But it should not replace or create a copy or create a link etc randomly without even prompting the user.

Special functions like 'replace' 'copy' or 'create link' should be achieved only with modifier keys.

@realthunder can you please advise how to fix this situation?

Related discussion : https://forum.freecad.org/viewtopic.php?p=723942#p723942

### Full version info

```shell
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35306 +17 (Git)
Build type: Release
Branch: asm_card9
Hash: a601f402437217a64cccd1b5084b6a7738837ffe
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk , OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * A2plus 0.4.60k
  * Assembly3 0.12.0
  * Assembly4 0.50.2
  * Curves 0.6.8
  * ExtremeProLight 1.5.0
[...]

PR #11932, opened 2024-01-08
Core: tree: prevent object replacement when dragging

PaddleStroke:
Fixe biggest issue of https://github.com/FreeCAD/FreeCAD/issues/11676 : 

image


chennes:
@PaddleStroke is this only preventing drops onto Origin items, or is it preventing drops entirely?

PaddleStroke:
This is preventing the 'replace mode' while doing a drag and drop without a modifier key (ctrl or alt).

Currently in master, when you try to drop something, depending on where you are dropping sometimes it does a replacement, ie swap both objects. Which is what happens on my screenshot.

This PR disable this mode from happening. And so the described bug from happening

PaddleStroke:
I think this mode can be triggered by holding ctrl or alt.

PR #11939, opened 2024-01-09
[Path] Rest machining from gcode

davidgilkaufman:
[Forum post](https://forum.freecad.org/viewtopic.php?t=78773&sid=5fa0a750380e7bd8c78aa626592c5bb5&start=30)

This PR is my next iteration of rest machining for FreeCAD. It discards the old approach of using internals of the Area class to determine previously cleared regions, and instead computes those regions by processing the gcode from earlier operations. This is trickier, but comes with the advantage of being more flexible: in this version of the rest machining feature, any operation (most notably Adaptive operations) can be the source tool path for rest machining.

Even with the improvements from this iteration, only Area operations (Pocket, Pocket3D, Profile) can take advantage of rest machining. I hope to implement rest machining for Adaptive tool paths in my next pass improving this feature.

Here is [the FreeCAD file](https://github.com/FreeCAD/FreeCAD/files/13868014/test_rectangle3.zip) I've been using to test this feature. It contains 3 parts with test tool paths:

The first part is the part I used to test my initial implementation of rest machining. There are 3 operations. The first operation is a Pocket on the right side, 6mm tool, with no special features. The second operation is a Pocket covering the full top area with a 1.5mm tool, with rest machining enabled. The third operation is a Pocket3D on the entire volume with the same 1.5mm tool, with rest machining enabled. The implementation producing these results is new, but this functionality already exists in the version of rest machining currently in FreeCAD.

Part 1

The second part tests rest machining after an adaptive operation. The first operation is an Adaptive operation in the top region with a 6mm tool. The second operation is a Pocket3D on the full volume with a 1.5mm tool.

Part 2

The third part tests speed optimizations and drilling. It has an absurdly dense Adaptive operation on the left region (visibility limited to the first 50k segments as a necessity to improve rendering performance). After that there is a drilling operation on the right side, then a Pocket operation on the right side with rest machining enabled. In addition to demonstrating that rest machining can process drilling gcode commands, this part demonstrates that bounding box filtering of far-away gcode commands works correctly (without that, generating the Pocket would be _very_ slow).

20240108_20h57m23s_grim

Todo:
- [ ] squash commits, and label the full commit with [Path]

### DIFF
--- a/src/Mod/Path/App/Area.cpp
+++ b/src/Mod/Path/App/Area.cpp
@@ -533,7 +533,7 @@ class ClearedAreaSegmentVisitor : public PathSegmentVisitor
 private:
     CArea pathSegments;
     double maxZ;
-    double diameter;
+    double radius;
 
     void line(const Base::Vector3d &last, const Base::Vector3d &next)
     {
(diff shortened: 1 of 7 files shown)


davidgilkaufman:
@sliptonic could you review this for me, or assign an alternative reviewer?

sliptonic:
I've compiled and tested. I'm inclined to merge as it seems to work as expected.  I'm requesting someone else to review code since my c++ is so weak.

PR #11940, opened 2024-01-09
Arch: implement new get_param functions

Roy-043:
Additionally 2 Arch_Window bugs were fixed:
* If the W1 value was changed the box tracker was not repositioned relative to the cursor.
* The WindowColor was not applied because of a typo in the code. De current default color is quite dark BTW.

Note that all dimensional values that were not really defaults, but just the last entered values, have been removed from preferences-archdefaults.ui. As a result the layout looks a bit strange. That will be improved in a next PR.


src/Mod/Arch/ArchAxis.py
```diff
@@ -224,7 +226,7 @@ def __init__(self,vobj):
 
     def setProperties(self,vobj):
 
-        ts = FreeCAD.ParamGet("User parameter:BaseApp/Preferences/Mod/Draft").GetFloat("textheight",350)
+        ts = params.get_param("textheight")
```
yorikvanhavre, line 229: Not sure why we specifically provided a default here, but might be good to check?

src/Mod/Arch/ArchBuildingPart.py
```diff
@@ -567,10 +569,10 @@ def setProperties(self,vobj):
             vobj.ShowLabel = True
         if not "FontName" in pl:
             vobj.addProperty("App::PropertyFont","FontName","BuildingPart",QT_TRANSLATE_NOOP("App::Property","The font to be used for texts"))
-            vobj.FontName = Draft.getParam("textfont","Arial")
+            vobj.FontName = params.get_param("textfont")
         if not "FontSize" in pl:
             vobj.addProperty("App::PropertyLength","FontSize","BuildingPart",QT_TRANSLATE_NOOP("App::Property","The font size of texts"))
-            vobj.FontSize = Draft.getParam("textheight",2.0)
+            vobj.FontSize = params.get_param("textheight")
```
yorikvanhavre, line 575: Again here, different default from above...

yorikvanhavre (approved):
Awesome! I would just verify these two

src/Mod/Arch/ArchBuildingPart.py
Roy-043: 2mm is very small for an object in a building model. In both this case and the Arch Axis case, multiplying the `textheight` with the `DefaultAnnoScaleMultiplier` would be a good idea I think. That way these texts will have the same initial height as those created with Draft_Text.

For the future we should perhaps consider adding a `ScaleMultiplier` property  to these objects. I am not fully aware how the texts in a BuildingPart are used, but axis bubbles are annotations.

Roy-043:
There are a number of boolean parameters that are not exposed in the GUI:
https://github.com/Roy-043/FreeCAD/blob/ff2cc693c5e5be265907a168ff07f81484ffa274/src/Mod/Draft/draftutils/params.py#L396-L458

I do not know how relevant any of these are. But `FreeLinking` may need to be exposed, or mentioned on the [Fine-tuning](https://wiki.freecad.org/Fine-tuning) page. There is a warning if you create an Arch_Site with a pre-selection that (I think) refers to this parameter.

yorikvanhavre:
Most of those parameters are not exposed in preferences, but well in other parts of the Arch GUI. They are basically "remembered" values so next time you use the tool you start with the same values as last time.

Roy-043:
AFAICT none of these boolean parameters is ever set to a different value in the code.

yorikvanhavre:
In any case, let's merge this now and see what problems pop out :)

PR #11943, opened 2024-01-09
[PartDesign] MultiTransform add Legacy support for rare cases

Syres916:
When https://github.com/FreeCAD/FreeCAD/pull/4504 was tested/merged it would appear at least one niche case was not taken into account when using the improved performance Fuzzy/Fuse algo. This involves using a Datum Plane as a reference for a Pocket, I would call it a Slice-Pocket which under the 'new' code produces two solids when in 0.19.x using the legacy code it correctly produced one solid.

This PR looks for this niche case and if present runs the legacy code and therefore fixing the regression. As 99.9999% of cases are unaffected there is no performance impact for them. A test for this case has been created alongside.

For more info please see forum discussion: https://forum.freecad.org/viewtopic.php?t=84006

### DIFF
--- a/src/Mod/PartDesign/App/FeatureTransformed.cpp
+++ b/src/Mod/PartDesign/App/FeatureTransformed.cpp
@@ -210,9 +210,10 @@ App::DocumentObjectExecReturn *Transformed::execute()
 
     supportShape.setTransform(Base::Matrix4D());
     TopoDS_Shape support = supportShape.getShape();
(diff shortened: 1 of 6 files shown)


adrianinsaval:
do you have a clue on why the new method fails? perhaps this should be fixed somewhere else rather than just going back to what was done before. To be clear I don't have a problem with merging this but it might be a good idea to check if the bug is somewhere else.

PR #11948, opened 2024-01-10
Fem: Fix constraint arrows - fixes #6073

### LINKED ISSUE #6073: [Problem] FEM: Tube does not show Fixed or Force Constraints
Issue imported from https://tracker.freecad.org/view.php?id=4366

* **Reporter:** FreeCAD97990
* **Date submitted:** 6/8/2020

# Original report text

I've noticed that Fixed Constraints and Force Constraints are not displayed when they are specifed on the end faces of a Tube.

Please see the attached file: [Tube does not show Fixed or Force Constraints.FCStd.gz](https://github.com/FreeCAD/FreeCAD/files/10775394/Tube.does.not.show.Fixed.or.Force.Constraints.FCStd.gz)

I've looked at a number of different cross sections, like a solid bar, and also tubes with square, hexagonal, and octagonal hollow profiles, and they all show Fixed Constraints and Force Constraints properly.

I was at first concerned that because the Constraints were not shown on the screen then the Constraints had not been applied to the Calculix model either. But when I went on to mesh and solve the Analysis, then there is deformation that shows that the Fixed and Force Constraints have actually been applied. They also appear in the Calculix inp file.

I put out a Forum post on this issue on Thu May 14, 2020 6:26 pm and bernd replied on Thu May 14, 2020 8:34 pm to confirm the bug and to say that it seemed to occur when the walls of the Tube are thin (less than around 10mm). He said we should make a entry on mantis and post a link to it here.

Now, I'm not too good at making links, so please refer to these posts if I haven't managed to put a link into this bug report.

# Steps to reproduce

Just load the attached file. No constraints are shown. I've just tested it with the latest 0.19 release, and the problem is still there.

# FreeCAD Info

```
OS: Windows 10 (10.0)
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.19.21514 (Git)
Build type: Release
Branch: master
Hash: 89ddab33f692a9bd739171689f9ab0eb58d64b35
Python version: 3.6.8
Qt version: 5.12.1
Coin version: 4.0.0a
OCC version: 7.3.0

```

# Other bug information

* **Priority:** normal
* **Severity:** minor
* **Category:** Bug
* **Updated:** 2/6/2021

marioalexis84:
With some trimmed surfaces, it may happen that the points selected based on the iso-curves used to place the constraint arrows do not coincide with the interior of the surface.
In that case, with this commit, the arrows are placed on the outer wire.
See  #6073.

### DIFF
--- a/src/Mod/Fem/App/FemConstraint.cpp
+++ b/src/Mod/Fem/App/FemConstraint.cpp
@@ -25,12 +25,14 @@
 
 #ifndef _PreComp_
 #include <Adaptor3d_IsoCurve.hxx>
+#include <BRepAdaptor_CompCurve.hxx>
 #include <BRepAdaptor_Curve.hxx>
 #include <BRepAdaptor_Surface.hxx>
 #include <BRepClass_FaceClassifier.hxx>
 #include <BRepGProp.hxx>
 #include <BRepGProp_Face.hxx>
 #include <BRep_Tool.hxx>
+#include <BRepTools.hxx>
 #include <GCPnts_AbscissaPoint.hxx>
 #include <GProp_GProps.hxx>
 #include <GeomAPI_IntCS.hxx>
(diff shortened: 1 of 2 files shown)


chennes:
@FEA-eng can you compile and run thus and confirm that it solves the issue?

FEA-eng:
@chennes I can confirm that this PR fixes the issue.

chennes:
Thanks!

Issue #9760, opened 2023-06-11
OCC: Regression on toNurbs()


edwilliams16:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: macOS 13.2.1
Word size of FreeCAD: 64-bit
Version: 0.21.0.33380 (Git)
Build type: Release
Branch: master
Hash: 16b06993e23eabd411586c7908be08ad8f609ce0
Python 3.11.3, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.2, OCC 7.7.1
Locale: C/Default (C)
Installed mods: 
  * MeshRemodel 1.8919.0
  * DynamicData 2.46.0
  * lattice2.backup1682566094.626239 1.0.0 (Disabled)
  * fcgear 1.0.0
  * workfeature
  * Pyramids-and-Polyhedrons
  * QuickMeasure-main
  * GDML 1.8.0
  * offline-documentation
  * Manipulator 1.5.0
  * fasteners 0.4.56
  * lattice2 1.0.0
  * CurvedShapes 1.0.4
  * Alternate_OpenSCAD 1.0.0
  * toSketch 1.0.1
  * Help 1.0.3
  * Curves 0.6.10
[/code]
```

### Subproject(s) affected?

None

### Problem description

If you select a bspline edge and send it to the Python console then

```
>>> c = sub.toNurbs().Edges[0]
>>> c.Curve
Traceback (most recent call last):
  File "<input>", line 1, in <module>
Base.FreeCADError: Unknown exception while reading attribute 'Curve' of object 'TopoShape'
```

the created edge has no underlying Curve.  This breaks  JoinCurves in the Curves WB and likely other things.  It may be associated with OCC 7.7.x. See forum https://forum.freecad.org/viewtopic.php?p=687077#p687077

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adrianinsaval:
@bgbsww suspected occt 7.7 bug, would be nice to know if it works on 7.8

bgbsww:
Unfortunately, I get the exact no attribute error above.  @adrianinsaval 
```
OS: Ubuntu 22.04.3 LTS (ubuntu:GNOME/ubuntu)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35500 (Git)
Build type: Debug
Branch: bgbsww-occt7.8.0
[...]

PR #11955, opened 2024-01-11
Add a test for the bspline curve bug OCCT 7.7.0 introduces.  FC bug #9760

bgbsww:
This passes when used with OCCT 7.6.3 and below.  As of 7.8.0 the upstream issue remains; this test will pass when that bug is fixed.

### DIFF
--- a/src/Mod/Part/TestPartApp.py
+++ b/src/Mod/Part/TestPartApp.py
@@ -20,7 +20,10 @@
 #   USA                                                                   *
 #**************************************************************************
 
-import FreeCAD, unittest, Part, Sketcher
+import FreeCAD
+import unittest
+import Part
+import Sketcher
 import copy
 import math
 from FreeCAD import Units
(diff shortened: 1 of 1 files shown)


adrianinsaval:
conda ci should run with occt 7.7.2 now, let's see what happens

adrianinsaval:
ugh wrong issue, I thought this was the one that might be fixed in 7.7.2 but this is the one with the pending patch upstream. I will see about adding that patch to the conda releases

adrianinsaval:
@chennes what do you think we should do about this? This test might fail on platforms where occt 7.7+ isn't patched with https://github.com/FreeCAD/OCCT/commit/8db1faa32e326889fc770b4389629d686ccb52d5, the patch is included in conda-forge so the conda workflows are expected to pass. I don't know if you included this in Libpack3

adrianinsaval:
I have confirmed that the test fails on arch with occt 7.8.1 from arch repos, it is strange that the test passes on windows ci as the libpack version does not include the patch with the fix, maybe the issue was not present on windows to begin with?

efferre79:
Has the patch applied to the freecad OCCT repo been submitted upstream too?

adrianinsaval:
I thought yes but looks like it wasn't: https://github.com/FreeCAD/FreeCAD/issues/9760#issuecomment-1901239344 but there was a bugreport where the fix is suggested https://tracker.dev.opencascade.org/view.php?id=33576

Reqrefusion:
@pieterhijma have you seen this before?

pieterhijma:
> @pieterhijma have you seen this before?

Yes, this can be closed as #9760 has been closed. It has been fixed in a recent OCCT. In the OCCT project it is also marked as done.

PR #11962, opened 2024-01-12
Toponaming: makeElementCompound

chennes:
This implements the method originally called `makECompound` (and now called `makeElementCompound`) as well as all auxiliary methods required by it. This is the first of the toponaming-aware object creation methods being added to the TopoShape class.

There is some follow-on work to be done refactoring the `mapSubElement`, which has very high complexity and requires some additional tests before it can be refactored safely.

### DIFF
--- a/src/App/ComplexGeoData.cpp
+++ b/src/App/ComplexGeoData.cpp
@@ -184,6 +184,11 @@ bool ComplexGeoData::getCenterOfGravity(Base::Vector3d& unused) const
     return false;
 }
 
+const std::string &ComplexGeoData::elementMapPrefix() {
+    static std::string prefix(";");
+    return prefix;
+}
+
 size_t ComplexGeoData::getElementMapSize(bool flush) const
 {
     if (flush) {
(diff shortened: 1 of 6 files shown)


chennes:
@bgbsww, @CalligaroV, @JohnAD, and @Pesc0, your feedback would be appreciated.

CalligaroV:
@chennes I just tried the last commit.

I can compile and run it without issues, tested doing the same steps I did for the [ShapeProtector branch](https://github.com/Pesc0/FreeCAD/tree/toponaming-shapeprotector).

Do you have some specific tests in mind for this PR?

chennes:
No: except the tests, nothing calls this code yet. So if you'd like to look at the code itself, that is about all we can do.

chennes:
I've restructured the commits to make better sense and to clarify what is realthunder's original code and what are my modifications to it. This is ready for merge via a merge commit (from my perspective).

3x380V:
[Confusing history](https://forum.freecad.org/viewtopic.php?t=71687&start=61) comment from the forum applies here as well. What about simply modifying it, so patch reads for example (note case typo fixed)?
```
From: "Zheng, Lei" <realthunder.dev@gmail.com>
Subject: [PATCH] Part/Toponaming: Add original code for makeCompound

Original-commit-id: <sha1 goes here>
[chennes@gmail.com: Renamed to makeElementCompound and modified as little as possible to get it to compile in the new framework. All required auxilliary methods are also brought in.]
```

chennes:
Thanks for the suggestion, @3x380V -- I've amended the authorship of the first commit, which ought to make `git blame` a little clearer for future developers.

CalligaroV:
@chennes, I checked the commits for this PR.

Beside what you already said (nothing calls this code exept the tests), I haven't found anything strange.
Also the refactoring looks doing the same logic of the original code.

Therefor if the only thing to do was the review I don't see reasons for not merging but maybe it would be better if also some other developer gives some comments.

Thanks!

PR #11969, opened 2024-01-13
[Feature Request] [Core] [Preferences] Reset preference pages parameters set without using "PrefWidgets" objects

CalligaroV:
Continues the modifications for #10161

This PR adds the management of the parameters set using Preference Pages Widgets that aren't derived from `PrefWidgets` objects and therefore ignored by the `PreferencePage::resetSettingsToDefaults` method

The Preference Pages subject to this PR are only the ones within the directory `src/Gui/PreferencePages`, no modifications have been made to the Preference Pages of the various Workbenches

### DIFF
--- a/src/Gui/PreferencePages/DlgSettingsEditor.cpp
+++ b/src/Gui/PreferencePages/DlgSettingsEditor.cpp
@@ -327,6 +327,23 @@ void DlgSettingsEditor::loadSettings()
     ui->displayItems->setCurrentItem(ui->displayItems->topLevelItem(0));
 }
 
+void DlgSettingsEditor::resetSettingsToDefaults()
+{
+    ParameterGrp::handle hGrp;
+    hGrp = WindowParameter::getDefaultParameter()->GetGroup("Editor");
+    //reset the parameters in the "Editor" group
+    for (QVector<QPair<QString, unsigned int> >::Iterator it = d->colormap.begin(); it != d->colormap.end(); ++it) {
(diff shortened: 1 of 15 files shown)


CalligaroV:
@wwmayer, with this PR I'm trying to address your comment on https://github.com/FreeCAD/FreeCAD/pull/10688#issuecomment-1746520087. Do you have any comment about these modifications?

the method `resetSettingsToDefaults` has already been moved to became a member of `PreferencePage` objects as for your suggestion with #11018  (Thanks @kadet1090 also for this modification!)

As always, before committing and opening this PR I tried to test it and it looks working correctly, at least for what I can see, but if you see something strange please let me know!

src/Gui/PreferencePages/DlgSettingsEditor.cpp
```diff
@@ -327,6 +327,23 @@ void DlgSettingsEditor::loadSettings()
     ui->displayItems->setCurrentItem(ui->displayItems->topLevelItem(0));
 }
 
+void DlgSettingsEditor::resetSettingsToDefaults()
+{
+    ParameterGrp::handle hGrp;
+    hGrp = WindowParameter::getDefaultParameter()->GetGroup("Editor");
+    //reset the parameters in the "Editor" group
+    for (QVector<QPair<QString, unsigned int> >::Iterator it = d->colormap.begin(); it != d->colormap.end(); ++it) {
```
CalligaroV, line 335: @chennes suggestion: change the for loop with the "new style" for loop, both here and in other parts of this file. (In a new PR)

Issue #11825, opened 2023-12-24
[Problem] no automatic quantity conversion: `App::anyToQuantity` does not handle `string` case


Andrei-Pozolotin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

PROBLEM:

currently it is impossible to handle automatic quantity conversion in property expressions
due to:  `App::anyToQuantity` does not handle `string` case
https://github.com/FreeCAD/FreeCAD/blob/8d51d78ce7e44526be6a84daa6dfa2a62c7a1896/src/App/Expression.cpp#L520

VERIFICATION:

* add new document
* add new object (say spreadsheet)
* add new text property: `App::PropertyString` called  `size_text` = `10 mm`
* add new data property: `App::PropertyLength` called  `size_quantity`
* set property expression for the `size_quatity` = `size_text`
* click "recompute"
* observe error message: `<Exception> Failed to convert to Quantity`

<details>
<summary><b>Sample stack trace</b></summary>

```
libFreeCADApp.so!App::anyToQuantity(const App::any & value, const char * msg) (/home/work/source/git/FreeCAD_VS/src/App/Expression.cpp:535)
libFreeCADApp.so!App::isAnyEqual(const App::any & v1, const App::any & v2) (/home/work/source/git/FreeCAD_VS/src/App/Expression.cpp:577)
libFreeCADApp.so!App::PropertyExpressionEngine::execute(App::PropertyExpressionEngine * const this, App::PropertyExpressionEngine::ExecuteOption option, bool * touched) (/home/work/source/git/FreeCAD_VS/src/App/PropertyExpressionEngine.cpp:694)
libFreeCADApp.so!App::Document::_recomputeFeature(App::Document * const this, App::DocumentObject * Feat) (/home/work/source/git/FreeCAD_VS/src/App/Document.cpp:3127)
libFreeCADApp.so!App::Document::recompute(App::Document * const this, const std::vector<App::DocumentObject*, std::allocator<App::DocumentObject*> > & objs, bool force, bool * hasError, int options) (/home/work/source/git/FreeCAD_VS/src/App/Document.cpp:2856)
libFreeCADApp.so!App::DocumentPy::recompute(App::DocumentPy * const this, PyObject * args) (/home/work/source/git/FreeCAD_VS/src/App/DocumentPyImp.cpp:584)
[...]

PR #11937, opened 2024-01-08
Resolve #11825 - no automatic quantity conversion: App::anyToQuantity

Andrei-Pozolotin:
Resolve #11825 - no automatic quantity conversion:` App::anyToQuantity`

Issue #10058, opened 2023-08-03
Spreadsheet: adding text to a cell prepends single quote to cell string value


MattN92705:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
[code]
OS: macOS 10.15
Word size of FreeCAD: 64-bit
Version: 0.21.0.33682 (Git)
Build type: Release
Branch: master
Hash: 9dce2525f8efe0873787f4388f10e11e168fe16b
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: C/Default (C)
Installed mods: 
  * A2plus 0.4.60n
  * fasteners 0.4.56
  * ExtremeProLight 1.6.0
  * Help 1.0.3
  * Assembly4 0.50.2
  * Assembly3 0.12.0
[/code]
```

### Subproject(s) affected?

Spreadsheet

### Problem description

FreeCAD weekly build 33682 has a bug in Spreadsheet when entering text into a cell.

- Type some text into a cell, e.g., “This is text”
- Hit return to enter the value
- The value appears to be correct when looking at the text in the spreadsheet, but…
- Select the cell with the text and look at the “Content” line at the top
- The cell’s content is actually “ ‘This is text “ (note the extra single quote prepended to the string).

Entering numbers seems to work correctly.

I like to build lists of “parameter name” : “value” pairs in adjacent cells by copying the parameter name and pasting it into the Alias property of the cell with the value.  The single quote makes the parameter name an invalid alias, requiring manually deleting the single quote from the alias field.

Small bug, but if you are putting together a large spreadsheet with parameters, it gets tiring.

### Anything else?

![Screen Shot 2023-08-03 at 8 02 31 AM](https://github.com/FreeCAD/FreeCAD/assets/134621706/62b48d52-8337-467a-8db0-edb86f76fb12)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

JohnOCFII:
Interesting.  I saw this change, and since it impacted my use of the EasyAlias macro, I created a note in the forum for the macro author, since I wasn't sure if this changed behavior was actually intended with 0.21 or not.

https://forum.freecad.org/viewtopic.php?t=80122

chennes:
[...]

PR #11971, opened 2024-01-13
Resolve #11965 (A) no proper execute() for cross-property references

### LINKED ISSUE #11965: [Problem] no proper `PropertyExpressionEngine.execute()` for cross-property references

### Problem description

Recently resolved issue #11825 with PR #11937 removed previously occurring spurious exceptions,
and exposed next-level problem:  

PROBLEM:
in a use case when one property has a simple expression referencing another property (see  #11825)
the `value` part of the source quantity is propagated as-is to the target,  without any regard to the `unit` information

specifically, as code tracing shows, that when  `PropertyExpressionEngine::execute()` gets to processes 
the expression constructed elsewhere by the `ExpressionParser::parse()`, 
https://github.com/FreeCAD/FreeCAD/blob/3ba0c3d7958dad891e125e065a41d95faac47c6e/src/App/PropertyExpressionEngine.cpp#L607

the type of the expression corresponding to the `source property`  here `std::shared_ptr<App::Expression>`:
https://github.com/FreeCAD/FreeCAD/blob/3ba0c3d7958dad891e125e065a41d95faac47c6e/src/App/PropertyExpressionEngine.cpp#L677

is actually `VariableExpression`
https://github.com/FreeCAD/FreeCAD/blob/3ba0c3d7958dad891e125e065a41d95faac47c6e/src/App/Expression.cpp#L2896

and not the `UnitExpression` as one would expect
https://github.com/FreeCAD/FreeCAD/blob/3ba0c3d7958dad891e125e065a41d95faac47c6e/src/App/Expression.cpp#L1222

which explains the lack of `value/unit` information conversion

Here is a proposed test case, which is currently failing:
https://github.com/FreeCAD/FreeCAD/compare/main...random-cadder:FreeCAD:issue_11825_A

SOLUTION:
it is not clear how to proceed:
* should this be resolved in `ExpressionParser::parse()` ?
* should this be handled in the `PropertyExpressionEngine::execute()`?
* is there perhaps an expression language syntax with meaning: `TreatMeAsUnitExpression(<<10 mm>>)` ?

@chennes  : please advise

### Full version info

```shell
OS: Arch Linux (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35567 (Git)
Build type: None
Branch: makepkg
Hash: 6a54591bac6c8843f12c6971623437083d017ab3
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

Andrei-Pozolotin:
Resolve #11965 : Solution A: magic reference syntax: `prop_name + 0mm`

### DIFF
--- a/src/Base/QuantityPyImp.cpp
+++ b/src/Base/QuantityPyImp.cpp
@@ -22,6 +22,8 @@
 
 #include "PreCompiled.h"
 
+#include <QString>
+
 // inclusion of the generated files (generated out of QuantityPy.xml)
 #include "QuantityPy.h"
 #include "UnitPy.h"
@@ -333,6 +335,9 @@ static Quantity& pyToQuantity(Quantity& q, PyObject* pyobj)
     }
     else if (PyLong_Check(pyobj)) {
         q = Quantity(PyLong_AsLong(pyobj));
+    }
+    else if (PyUnicode_Check(pyobj)) {
+        q = Quantity::parse(QString::fromUtf8(PyUnicode_AsUTF8(pyobj)));
     }
     else {
         PyErr_Format(PyExc_TypeError, "Cannot convert %s to Quantity", Py_TYPE(pyobj)->tp_name);
--- a/tests/src/App/CMakeLists.txt
+++ b/tests/src/App/CMakeLists.txt
@@ -13,5 +13,6 @@ target_sources(
             ${CMAKE_CURRENT_SOURCE_DIR}/MappedName.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/Metadata.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/Property.cpp
+            ${CMAKE_CURRENT_SOURCE_DIR}/PropertyExpressionEngine.cpp
             ${CMAKE_CURRENT_SOURCE_DIR}/StringHasher.cpp
 )
--- a/tests/src/App/PropertyExpressionEngine.cpp
+++ b/tests/src/App/PropertyExpressionEngine.cpp
@@ -1,0 +1,91 @@
+#include "gtest/gtest.h"
+
+#include "Base/Quantity.h"
+
+#include "App/Application.h"
+#include "App/Document.h"
+#include "App/DocumentObject.h"
+#include "App/Expression.h"
+#include "App/ObjectIdentifier.h"
+#include "App/PropertyExpressionEngine.h"
+
+#include "src/App/InitApplication.h"
+
+// clang-format off
+
+class PropertyExpressionEngine: public ::testing::Test
+{
+protected:
+    static void SetUpTestSuite()
+    {
+        tests::initApplication();
+    }
+
+    void SetUp() override
+    {
+        _doc_name = App::GetApplication().getUniqueDocumentName("test");
+        _this_doc = App::GetApplication().newDocument(_doc_name.c_str(), "testUser");
+        _this_obj = _this_doc -> addObject("Sketcher::SketchObject");
+        _source_name = std::string("this_origin");
+        _source_prop = _this_obj -> addDynamicProperty("App::PropertyString", _source_name.c_str()); // property with length as string
+        _target_name = std::string("this_length");
+        _target_prop = _this_obj -> addDynamicProperty("App::PropertyLength", _target_name.c_str()); // property with reference to source
+    }
+
+    void TearDown() override
+    {
+        App::GetApplication().closeDocument(_doc_name.c_str());
+    }
+
+    std::string doc_name() { return _doc_name; }
+    App::Document* this_doc() { return _this_doc; }
+    App::DocumentObject* this_obj() { return _this_obj; }
+    std::string source_name() { return _source_name; }
+    App::Property* source_prop() { return _source_prop; }
+    std::string target_name() { return _target_name; }
+    App::Property* target_prop() { return _target_prop; }
+
+private:
+    std::string _doc_name;
+    App::Document* _this_doc {};
+    App::DocumentObject* _this_obj {};
+    std::string _source_name;
+    App::Property* _source_prop {};
+    std::string _target_name;
+    App::Property* _target_prop {};
+};
+
+// https://github.com/FreeCAD/FreeCAD/issues/11965
+TEST_F(PropertyExpressionEngine, executeCrossPropertyReference)
+{
+    auto source_text = std::string("1.5 m"); // provided in source
+    auto target_text = std::string("1500 mm"); // expected in target
+
+    auto source_path = App::ObjectIdentifier::parse(this_obj(), source_name());
+    source_prop()->setPathValue(source_path, source_text);
+
+    auto target_expr = source_name() + "+0mm"; // Solution A: "magic reference syntax"
+
+    auto target_path = App::ObjectIdentifier::parse(this_obj(), target_name());
+    std::shared_ptr<App::Expression> target_rule(App::Expression::parse(this_obj(), target_expr));
+    this_obj()->setExpression(target_path, target_rule);
+
+    this_obj() -> ExpressionEngine.execute();
+
+    auto source_entry = source_prop() -> getPathValue(source_path);
+    ASSERT_TRUE(source_entry.type() == typeid(std::string));
+    auto source_value = App::any_cast<std::string>(source_entry);
+
+    auto target_entry = target_prop() -> getPathValue(target_path);
+    ASSERT_TRUE(target_entry.type() == typeid(Base::Quantity));
+    auto target_quant = App::any_cast<Base::Quantity>(target_entry);
+    auto target_value = target_quant.getValue();
+    auto target_unit = target_quant.getUnit().getString().toStdString();
+
+    EXPECT_EQ(target_quant, App::parseQuantityFromText(target_text)) << ""
+        "expecting equal: source_text='" + source_text + "' target_text='" + target_text + "'"
+        "instead produced: target_value='" + std::to_string(target_value) + "' target_unit='" + target_unit + "'"
+    ;
+}
+
+// clang-format on


Andrei-Pozolotin:
@chennes : this check

https://results.pre-commit.ci/run/github/5736080/1705438623.In1y7EU4TWug_6tnBmvrzw

keeps failing with:
```
clang-format.............................................................Failed
- hook id: clang-format
- files were modified by this hook
```

now can one find which "files were modified by this hook" and how to fix that?

chennes:
You have two possible solutions: preferred is to set up precommit as described in the manual [here](https://freecad.github.io/DevelopersHandbook/gettingstarted/#setting-up-for-development). Alternatively, you can check the box that gives administrators here the permission to modify your PR, which will also allow the CI to modify it to add a commit that fixes the issue.

Andrei-Pozolotin:
@chennes : `pre-commit.ci ` check resolved, thank you.

perhaps `pre-commit.ci` config could use a more informative invocation:
https://pre-commit.com/#pre-commit-run
```
pre-commit run --all-files --show-diff-on-failure
```
such that run report would contain not only summary but also the details.

please review

chennes:
> perhaps `pre-commit.ci` config could use a more informative invocation: https://pre-commit.com/#pre-commit-run
> 
> ```
> pre-commit run --all-files --show-diff-on-failure
> ```
> 
> such that run report would contain not only summary but also the details.

I'm not aware of a way to do that -- we are using the https://pre-commit.ci GitHub integration, and as far as I know it doesn't have any setting to change its reporting. I'm happy to be proven wrong, of course!

Andrei-Pozolotin:
> I'm not aware of a way to do that -- we are using the https://pre-commit.ci GitHub integration

one way to do this is here:
https://github.com/FreeCAD/FreeCAD/blob/main/.pre-commit-config.yaml

for example, as used by conda:
https://github.com/conda/conda/pull/13124/files

Andrei-Pozolotin:
RE:

> @[chennes](https://github.com/chennes) chennes removed their request for review [2 days ago](https://github.com/FreeCAD/FreeCAD/pull/11971#event-11559546828)

please let me know who can help with this PR?

pieterhijma:
I looked at this PR but as far as I can see, this is a problem brought forward by the recently merged PR #11937.  It may be the case that the related issue #11965 is a more generic problem and not only related to the previous PR, but currently I cannot judge that.

My apologies for backtracking on this, but I do have a (humble) opinion of PR #11937 aiming to resolve #11825.  The PR has been merged, but the functionality is not generic and I believe it is questionable whether we want to risk type safety by this automatic string to quantity conversion.

The solution is not generic because we can now convert string properties to quantity properties, but not to other properties.  The solution converts a string such as "2.3 mm" to a length property that will have the value `2.3 mm` simply by referencing the string property, automatically coercing the string into a length (or other quantity).  However, we cannot coerce strings such as "2.3" to a float property or strings such as "4" to an integer property.  Additionally, how do those strings compose?  Can we use "2 mm + 3 mm" and coerce this to length `5 mm`? (It cannot currently.)

I don't see a strong need for this functionality (essentially for the problem described in #11825), but the issue lists a module where similar functionality is used, so apparently the functionality is sometimes requested.  However, a bit more context to the problem would be nice.

Regarding why I think we could ask questions about this functionality: The biggest problem for me would be that the conversion is automatic and implicit; I would prefer a solution that is (as said above) generic and explicit.  As an example, the expression engine has the function `str` that converts a value to a string, for example you can do `str(MyInt)` where `MyInt` is an integer property.  You can also do `str(2 + 3)`, so the expressions are truely evaluated.  These conversions are explicit by means of the `str` function.  I would prefer a similar but function but then the inverse, for example `parse` that parses a string and converts it to the proper type.  So, this allows you to do something like this:

- PropertyString text has value "1.23"
- PropertyFloat float with expression `parse(text) + 4.56` resulting into value `5.79`.
- It should then also be possible to parse "text + 4.56" resulting into value `5.79`.

The conversion to float is now explicit by means of the parse function.  This allows users to understand the logic better and also understand where potential parse errors originate from.

However, even for this, I don't really understand the need and it is a bit strange to have a parse function inside expressions that parses expressions while the parent expression has to be parsed as well.

Why not simply do the following that is more generic and doesn't require any change to the FreeCAD codebase (PR #11973 can be reverted): We have a document object `obj` with string property `text` with value "2 mm + 3 mm" and we have a length property `length`.

We want to end up with `length` being 5 mm.  We can simply do in Python:

``` python
obj.setExpression('length', obj.text)
```

This gives a length of 5 mm.  So, this solution requires a bit of Python, but is more generic.

Andrei-Pozolotin:
@pieterhijma : Pieter:
thank you for taking the time to think about this.

RE:
> However, even for this, I don't really understand the need ...

<details>
<summary>Here is a back story on the issue.</summary>

We develop a `python` plugin for `FreeCAD` 
which requires a functionality of "configuration tables".

* **First Attemt** was based on `Spreadsheet::Sheet` "configuration tables",
but we found them to be unusable in the end.

* **Second Attemt**, now in progress, is based on these design goals:
	* rely only on existing `FreeCAD` core code, both `cpp` an `python`
	* provide "configuration tables" which would work without the plugin `python` code present
	* that would also make document to be serializable/deserializable without the plugin code present 
	
this resulted in the following simple idea:
* use basic `App::PropertyPythonObject` as hidden "config table" repository, 
to be attached as a member of any standard `FreeCAD` object
* store content of this "config table" in a form of `python` built-in `dict()` dictionary
* use only `python` built-in data types, say `str`, to represent the leaf property values, like `"10 mm"`
* use enumeration properties and expressions similar to how it is used by `Spreadsheet::Sheet`
* have multiple user-visible (not hidden) properties referencing this "config table" 
* enforce these user-visible properties types on the property itself, expecting auto-convert from string,
and with attached property expressions such as:
```python
expression.path = ".User_Propery"
expression.value = ".Config_Table[<<User_Propery>> + .Enumeration_Property.String]" # this is a string
```

which "almost worked", except for the issues reported in the #11825 and #11965
 </details>

RE:

 > Regarding why I think we could ask questions about this functionality ...
 
 Yes, I also share your objections to both current PRs (that was the best I could envision at the time) 
 and yes, your `parse()` idea seems to be the appropriate way to move forward.
 
<details>
<summary>This is how the expression would look like.</summary>

```python
expression.path = ".User_Propery"
expression.value = "parse(.Config_Table[<<User_Propery>> + .Enumeration_Property.String])"
```

Hopefully type mismatch between `User_Propery` and the `parse()` result 
would produce a meaningful error message. 

</details>
 
 What to you think about the following work plan:
 * revert https://github.com/FreeCAD/FreeCAD/pull/11937
 * abandon https://github.com/FreeCAD/FreeCAD/pull/11971
 * work on the new PR towards `parse()`
 
 ?

pieterhijma:
@Andrei-Pozolotin 

Thank you for your explanation. 

Your proposal would make things already much better, but I'm currently trying to figure out if it would be possible without having a `parse` function. I'll get back to you on that.

I would be really interested in understanding why the spreadsheet configuration tables didn't work for you, this in relation to #12120 and my proposal for a solution #12135.

Just out of curiosity, could #12135 (if fully implemented) be a solution for your usecase?

Andrei-Pozolotin:
@pieterhijma : Pieter:

RE:
> without having a parse function. I'll get back to you ...

Great, looking forward to this.

RE:
>  why the spreadsheet configuration tables didn't work for you ...

<details>
<summary>Here is why spreadsheet "config tables" are unusable.</summary>

* spreadsheet "config table" UI is extremely fragile and limited:
   * one has to manage / duplicate property column titles manually,
   * one has to manage manually aiases to expose cell data as properties, 
   * there is no way to link cell user property comments to the actual object property comments,

* current spreadsheet "config table" implementation is broken in many ways: 
   * inserting a row or column before/after/inside the cell range of the existing "config table" results in 
unrecoverable expression entires corruption, resolvable only by delete/create table from scratch
   * there is no type safety available for the cell-alias-derived properties,
   * lack of `parse()` functionality resulted in current code use of "magic format" with `'`, `=`, prefixes
   #10058
   which essentially is just a merge of data with meta-data and which requires non-stop parse/unparse to dance around this

* spreadsheet "config table" feature is unsutable for use by an external plugin, since
spreadsheet has a **concept** of a "config table", but there is no single **object** representing it,
current code creates "config table" on-the-fly inside some UI dialog,
and there is no `cpp` nor `python` *api* to extract/manage/persist the "config table", instead,  
it takes about 300 lines of python code to parse/unparse from scattered bits of expressions, such as:
```python
f"cells[<<{cell_name}:->>]"
f"cells[<<{cell_name}:|>>]"
f"cells[<<{head_name}:{tail_name}>>]"
f"""
tuple(
    .cells;
    <<{head_col}>> + str( hiddenref({enum_name}) + {head_row} );
    <<{tail_col}>> + str( hiddenref({enum_name}) + {tail_row} )
)
"""
```

</details>

<details>
<summary>This is how "config table" proposed UI looks in our plugin.</summary>

taba-1

taba-2

</details>

RE:
> could #12135 be a solution ...

There must be some kind of [Planetary Alignment](https://starwalk.space/en/news/what-is-planet-parade) taking place, 
since that is one of the missing features expected by our plugin, 
which we setup to implement (read: duplicate) again.

Yes, the motivation for #12135 is definetly correct, and that would be a solution, except:
* current PR code does not show / let us understand `PropertyExpression` engine integration
* `Application Geometry Interface` sounds like too much magic right off the bat, 
perhaps a progression of api/classes with "more features as they grow" would be less scary
* comments over there might give a casual visitor an impression that it would take 
more than "seven ways to convince a committee", so, please let us have our `parse()` now, 
before the happy days of `App::VarSet` arrive. :heart_eyes:

pieterhijma:
@Andrei-Pozolotin:

> > without having a parse function. I'll get back to you ...
> 
> Great, looking forward to this.

Sorry, I'm at the FreeCAD hackathon in Brussels and got a bit distracted. I didn't find a nice solution, but in any case, you can use PropertyBags from the Path workbench (Path -> Utils -> PropertyBag) to store the properties instead of storing them in a dict in a Python object.

Suppose you have two property bags "A" and "B" and they both have property "Length" with different values. These property bags represent the configurations then.

One idea was to have property somewhere with a link to one of those property bags, let's call it Link. Then in the design you use the expression `Link.Length`. Changing the link essentially changes the configurations.

My following idea was to express the value of `Link` in terms of an enum `Config`, for example with values A and B. My idea was to have an expression such as `Config == 0 ? <<A>> : <<B>>`, where <<A>> or <<B>> However, I forgot that expressions only evaluate to strings or integers and not to PropertyLinks, so this doesn't work. So, if you're ok expressing configuration by switching links, then this would work. For using enums, this doesn't work.

I think I don't fully understand your use-case, so this information might help you to think of different solutions. Please let me know!

pieterhijma:
> @pieterhijma : Pieter:
> 
> RE:
> 
> > without having a parse function. I'll get back to you ...
> 
> Great, looking forward to this.
> 

> > why the spreadsheet configuration tables didn't work for you ...
> 
> Here is why spreadsheet "config tables" are unusable.

Thanks for this great summary of the issues. This is very helpful for me!

pieterhijma:
> There must be some kind of [Planetary Alignment](https://starwalk.space/en/news/what-is-planet-parade) taking place, since that is one of the missing features expected by our plugin, which we setup to implement (read: duplicate) again.
> 
> Yes, the motivation for #12135 is definetly correct, and yes, that would be a solution, except:
> 
>   * current PR code does not show / let us understand `PropertyExpression` engine integration
> 
>   * `Application Geometry Interface` sounds like too much magic right off the bat,
>       perhaps a progression of api/classes with "more features as they grow" would be less scary
> 
>   * comments over there might give a casual visitor an impression that it would take
>       much more than "seven ways to convince a committee", so, please let us have our `parse()` now,
>       before the happy days of `App::VarSet` arrive. 😍

:smile: Yes, sorry, I'm working on it. This PR is just a very small first step and it still abstract. I hope I can make it more clear coming weeks and hopefully this convinces the committee...

Andrei-Pozolotin:
@pieterhijma  :  Pieter:

Good to hear back from you.

>  I'm at the FreeCAD hackathon in Brussels ...

Great! Is there a web site or a youtube channel where one can learn more of the outcome of the hackathon?

> ... can use PropertyBags ...
> ... use the expression Link ...

Thank you for these ideas, but no,  these will not work.

Here I have a PR for "Plan B" or the  new `parsequant()` expression language function, 
* please let me know what you think?
* https://github.com/FreeCAD/FreeCAD/pull/12225

chennes:
@Andrei-Pozolotin is this PR dead now, to be replaced by other work?

Andrei-Pozolotin:
@chennes : yes,  these are pending steps:
* revert commit:
https://github.com/FreeCAD/FreeCAD/pull/11937
* abandon/close:
https://github.com/FreeCAD/FreeCAD/pull/11971
* review/commit:
https://github.com/FreeCAD/FreeCAD/pull/12225
* study/resove "Werner `1§` bug"
https://github.com/FreeCAD/FreeCAD/issues/11825#issuecomment-1935135562

PR #11980, opened 2024-01-14
Material: fix Qt6 build failure

3x380V:
QRegExp has been replaced with QRegularExpression and isn't part of the base of Qt6 any more


chennes:
Thanks for the PR - I thought we had fixed Qt6 compilation with #11795, @davesrocketshop?

3x380V:
I was surprised too, so I was waiting for a while if someone else notices. For completeness, compiled and tested on Debian GNU/Linux 12 (bookworm). According to [documentation](https://doc.qt.io/qt-6.2/qregexp.html#details), version check is not necessary.

PR #11987, opened 2024-01-14
fix build for python 3.12: remove usage of _Py_PackageContext in 3.12+

russelltg:
_Py_PackageContext was removed from cpython in https://github.com/python/cpython/commit/5f55067e238c21de25f09ece9bb24ae8c42d02b4, and I'm not seeing an obvious way to replace this functionality.

I'm not sure if this would regress any behavior, but it at least matches what the Limited API version does, so if that's a supported way to link against Python then it should be fine.

Came across this trying to build on Fedora 39. I wasn't actually able to test it as I wasn't able to get pyside6 to build properly unfortunately, but since I already made these changes to get it to build I figured I'd open a PR.


adrianinsaval:
thank you for your contribution, I think this PR can be superseded by #12286

PR #11992, opened 2024-01-15
[TD] Solve issue in VertexAtIntersection

edi271:
Solve the issue reported in https://forum.freecad.org/viewtopic.php?t=84315


WandererFan:
Thank you!

PR #11994, opened 2024-01-15
Tests: Avoid use of fmtlib when not needed

3x380V:
See post [Debug build fails : Libpack faulty](https://forum.freecad.org/viewtopic.php?t=84376&start=3) for context...


chennes:
Obviously the real solution is to get the LibPack fixed, but I'll merge this in the interim since pulling in libfmt for this trivial case does seem a bit superfluous.

Issue #10100, opened 2023-08-07
[Problem] Sketcher construction color + badground color are ill-suited for colorblinds.


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Full version info

```shell
0.22
```

### Subproject(s) affected?

None

### Problem description

I have the most common type of colorblindness and for my eye the default color of sketcher construction color is hard to see on the default background color. Here's a picture where I removed the red from the purple-ish color to give you a hint at how it looks blue on blue (I'm not sure if this sample will be very clear though):

![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/e24009b3-83b4-43f6-84ce-66e8a3fd1920)

So I suggest that we change either the background color, or the construction line color.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

FEA-eng:
Would a colorblind mode (like in some video games) make sense or are there too many variants of colorblindness?

PaddleStroke:
I don't think this is necessary. Many program manage to avoid the necessity
of colorblind mode by choosing a palette of colors that are
colorblind-friendly.
Here there just this one that is annoying.

FEA-eng:
Right, it's rather poorly visible even for non-colorblind users. I could say the same about the dimensions' color. Theoretically, it contrasts with the blue background but in practice, it's not contrasting enough.

obelisk79:
When choosing colors, it would be wise to choose based on *contrast*.

Take the desired color palette and desaturate them against the background color, the resulting contrast of the image afterwards should give a good indication of appearance in instances of color-vision deficiencies.

chennes:
Maybe the first-run Wizard that @benj5378 is working on should have a "high-contrast" option... isn't that what Visual Studio does?

benj5378:
@chennes Good idea, but implementation wise, it would just be a theme/preference pack together with other themes.

chennes:
That makes sense to me, but I am also fine with changing the "classic" (ahem) background color. Just be prepared for loud complaints about it.

benj5378:
@chennes Friedrich Niche: classic is dead

PaddleStroke:
Colorblinds should not have to use the high-contrast option. The high-contrast migh be useful for monochromacy-colorblind, but those are like 10000 times less frequent than the regular 'red-colorblindness' that affects 8% of male people.

[...]

PR #11996, opened 2024-01-15
[Sketcher] Enable customization of edges pattern and width by type.

PaddleStroke:
Enable customization of edges pattern and width by type. WIP

ie adds preferences for curve pattern and width for : normal / construction / internal / external geometries.

The main goal is to enable dashed lines for construction curves, which is a long standing request:
https://forum.freecad.org/viewtopic.php?t=76435
https://forum.freecad.org/viewtopic.php?t=80099
https://forum.freecad.org/viewtopic.php?t=30217
https://github.com/FreeCAD/FreeCAD/issues/10100
etc.

The way to do it is to introduce 4 coin layers for each user layers (Abdullah started to implement multi user-layers management even if it is almost unused).

Preferences : 
image
image

Currently default is : 
- Solid lines for normal geos.
- dashed lines for construction, internal and external. 
- 3 line width for normal
- 2 line width for construction, internal and external. 

@obelisk79 it would be good that the design group discuss the default for each line type. Also note that the pattern can be changed if necessary.

Useful links for reference: https://forum.freecad.org/viewtopic.php?style=4&t=61196


PaddleStroke:
Huge thanks to @kadet1090 who found the crash that held me the whole day poking in the dark !
It is now almost ready, we just need to discuss the default first.

obelisk79:
My proposal for defaults are as follows, along with some feedback regarding the spinboxes you've added:
https://media.discordapp.net/attachments/1162044910535450665/1197513543797448744/image.png

Which results in an appearance as seen here:
https://media.discordapp.net/attachments/1162044910535450665/1197514028126326915/image.png

pierreporte:
The preference page should be renamed since it’s not only about colors anymore. I suggest “appearance”.

src/Mod/Sketcher/Gui/EditModeCoinManagerParameters.h
```diff
@@ -404,17 +469,62 @@ struct CoinMapping
...
+        for (size_t l = 0; l < CurvIdToGeoId.size(); ++l) {
+            CurvIdToGeoId[l].clear();
+        }
         CurvIdToGeoId.clear();
         PointIdToGeoId.clear();
         GeoElementId2SetId.clear();
         PointIdToVertexId.clear();
     };
 
+    /* //Debug function.
```
kadet1090, line 481: should be deleted

src/Mod/Sketcher/Gui/EditModeCoinManagerParameters.h
```diff
@@ -461,7 +571,8 @@ struct CoinMapping
 
     //* These map a MF index (second index) within a coin layer (first index) for points or curves
     // to a GeoId */
-    std::vector<std::vector<int>> CurvIdToGeoId;   // conversion of SoLineSet index to GeoId
+    std::vector<std::vector<std::vector<int>>>
```
kadet1090, line 574: It is not a change for this PR, but this is getting way too complex. In that point we probably should introduce some proper structures for the layers with proper names, methods etc which would be easier to follow later. 

API could look something like that (just a concept):
```
layer(0).type(SubLayer::External).curveToGeo(curveIndex)
```

Now we have to deal with 3-level deep structures which are not easy to read and remember, stuff like `swcs[l * 4 + t]`.

src/Mod/Sketcher/Gui/EditModeGeometryCoinManager.cpp
```diff
@@ -73,19 +73,20 @@ void EditModeGeometryCoinManager::processGeometry(const GeoListFacade& geolistfa
...
-
     auto layersconfigurations = viewProvider.VisualLayerList.getValues();
 
     for (auto l = 0; l < geometryLayerParameters.getCoinLayerCount(); l++) {
-        setEnableLayer(l, layersconfigurations[l].isVisible());
+        auto enabled = layersconfigurations[l].isVisible();
+
+        swsp[l] = enabled;
+        for (int t = 0; t < geometryLayerParameters.getSubLayerCount(); t++) {
+            swsc[l * 4 + t] = enabled;
```
kadet1090, line 88: `4` should be replaced with `getSubLayerCount()`

kadet1090 (changes requested):

PaddleStroke:
Changed name of pref page to appearance

kadet1090 (approved):

chennes (changes requested):
Overall this looks good, thanks! The only issue I see right now is the excessive width of the Appearance tab -- can you tighten up the four lines that have the line style setting on them? On a Mac with the "Classic" theme selected the dialog is quite a bit too wide, requiring a horizontal scroll bar at the default width.

PaddleStroke:
@chennes width of the dialog is fixed, everything fit in : 
image

chennes:
You've still got the horizontal scrollbar there -- something is still amiss with the column widths, there is plenty of space for the widgets so I don't know why Qt wants to add it.

PaddleStroke:
image

Width fixed so that we don't have the horizontal bar anymore.

PR #11998, opened 2024-01-15
Gui: remember user closed overlay panel

kadet1090:
This ensures that after overlay area is hidden using "X" button it stays closed even after restart.

Replaces #11837

### DIFF
--- a/src/Gui/OverlayManager.cpp
+++ b/src/Gui/OverlayManager.cpp
@@ -813,28 +813,16 @@ public:
             setOverlayMode(OverlayManager::OverlayMode::TransparentAll);
             return;
         case OverlayManager::OverlayMode::ToggleLeft:
-            if (OverlayTabWidget::_LeftOverlay->isVisible())
-                OverlayTabWidget::_LeftOverlay->setState(OverlayTabWidget::State::Hidden);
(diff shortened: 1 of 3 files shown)


chennes:
@qewer33 can you confirm this fix?

chennes:
In my testing, the first half works well: if I close the overlay panel it stays closed on the next restart. But if I then show it again, it doesn't seem to remember that, closing it on the next restart.

kadet1090:
I'll make this one a draft until I fix #11680

maxwxyz:
@kadet1090 what's the status on this PR regarding #11680 ? Will this need to be fixed first or included in this PR? Or yould this be split in two separate PRs?

kadet1090:
The problem is - The fix that is included by realthunder does not fix mentioned issue. Fixing the issue properly requires a large amount of work (basically, I'd have to rework whole heading system for overlays). While I have working PoC of how this could be achieved - this is only PoC and I doubt that I'll be able to get it ready for merging before the deadline. 

I'll however try to solve this particular issue in a way that will not require major refactor.

maxwxyz:
@kadet1090 are you planning on working on this PR again?

PR #12002, opened 2024-01-15
Update .pre-commit-config.yaml

mosfet80:
Updated clang format


chennes:
We are keeping to versions of clang-format that can be run on our oldest supported OS, Ubuntu 20.04 LTS.

PR #12004, opened 2024-01-16
switch github library from pat-s to actions repo

mosfet80:


mosfet80:
@adrianinsaval  
If we enable the save-always flag, the cache is saved even in case of compilation failure.
It is normally disabled.
Do we want this?

oursland:
The original action [`pat-s/always-upload-cache@v3`](https://github.com/pat-s/always-upload-cache) is a fork of `cache` that is intended to upload, even in the case of the compilation failure.

A like-for-like replacement would retain this functionality, through the use of the `save-always` flag.

PR #12007, opened 2024-01-16
App/Toponaming: Add Comparator for mapped elements

chennes:
Add the comparison method for MappedName.

### DIFF
--- a/tests/src/App/MappedElement.cpp
+++ b/tests/src/App/MappedElement.cpp
@@ -127,3 +127,91 @@ TEST_F(MappedElementTest, lessThanOperator)
     EXPECT_FALSE(mappedElement2A < mappedElement1B);
     EXPECT_FALSE(mappedElement2B < mappedElement2BDuplicate);
 }
+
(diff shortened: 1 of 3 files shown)


chennes:
@realthunder why did you decide to implement this as a standalone function rather than as `MappedName::operator<`?

tests/src/App/MappedElement.cpp
```diff
@@ -127,3 +127,99 @@ TEST_F(MappedElementTest, lessThanOperator)
...
     EXPECT_FALSE(mappedElement2B < mappedElement2BDuplicate);
 }
+
+// Things to test:
+// One has zero size
+// Both have zero size
+// Both start with hex digits that are different
+// Both start with hex digits that are the same, but follow-on is different
+// No hex digits, lexical comparison on string following
+// No hex digits, strings match, numeric comparison of digits following
```
bgbsww, line 137: Did you want to remove these?

chennes: Yes, yes I do. Thanks.

Issue #12010, opened 2024-01-16
[Problem] sketcher: radius arc helpers not drawn correctly in all situation


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/84f5889c-b98a-4789-bb0a-b8cb41182b2f)

Discussed here https://forum.freecad.org/viewtopic.php?t=84319
if startpoint is before 0deg then it does this.
Working on it

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

maxwxyz:
can this be combined to fix https://github.com/FreeCAD/FreeCAD/issues/11976 or is this separate?

PaddleStroke:
I will check if I can solve 11976 at the same time

Issue #11976, opened 2024-01-13
[Feature Request] Sketcher: Add helpers to arcs for angle and distance constraints rendering


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Following https://github.com/FreeCAD/FreeCAD/pull/11507 with respective PR https://github.com/FreeCAD/FreeCAD/pull/11926 to add helpers in Sketcher WB for radius and line angle, additional helpers for arcs would be useful:
- arc to other geometry: distance
- arc: angle

current implementation:
![296455565-5f3c72d8-8306-4ffd-a3f1-6adfb8242893](https://github.com/FreeCAD/FreeCAD/assets/6246609/5e905155-6531-4d0c-b2f4-92ddbc41707b)

New issue from:
https://github.com/FreeCAD/FreeCAD/pull/11926#issuecomment-1890750085

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Roy-043:
Doesn't [Sketcher_ArcOverlay ](https://wiki.freecad.org/Sketcher_ArcOverlay) already answer part of this feature request?

maxwxyz:
only some part

PR #12012, opened 2024-01-16
[Sketcher] Improve rendering of arc-angle, line-angle, arc-distances and fix radius

PaddleStroke:
Fixes https://github.com/FreeCAD/FreeCAD/issues/11976
Fixes https://github.com/FreeCAD/FreeCAD/issues/12010

- Improve rendering of arc-angle
- Improve rendering of line-angle
- Improve rendering of arc-distances by adding arc helpers
- Fix radius arc helper that were wrong in some conditions.

image

@FlachyJoe @maxwxyz

### DIFF
--- a/src/Gui/SoDatumLabel.cpp
+++ b/src/Gui/SoDatumLabel.cpp
@@ -93,6 +93,9 @@ SoDatumLabel::SoDatumLabel()
     SO_NODE_ADD_FIELD(param2, (0.f));
     SO_NODE_ADD_FIELD(param4, (0.f));
     SO_NODE_ADD_FIELD(param5, (0.f));
+    SO_NODE_ADD_FIELD(param6, (0.f));
+    SO_NODE_ADD_FIELD(param7, (0.f));
+    SO_NODE_ADD_FIELD(param8, (0.f));
 
     useAntialiasing = true;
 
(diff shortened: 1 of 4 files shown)


PaddleStroke:
@FlachyJoe @maxwxyz

maxwxyz:
nice! thank you.
Two points I've noticed:
- outer dimension (arcs in arcs / offset situations) are not perfect
- cosmetic: the arc is extended in the red dimension line, creating a weird look as if both are fused. Could this helper be green from end point to the dimension line (like the arc helpers)?
grafik
grafik

PaddleStroke:
I have fixed the problematic cases and for the 'continuous' effect I have added a margin to the helper. This way it breaks this weird continuous curve effect. And also it looks better for radius : 
image

maxwxyz:
It looks better with the extended helpers. Will try it in the evening on some edge cases myself. When the green arc helpers are visible, these red ones render on top?

PaddleStroke:
Yes the red on top.

maxwxyz:
@PaddleStroke 
sometimes the extended arc is so small its not visible, sometimes it is very large, depending on the position of the leader line:
grafik

when helpers are activated, it just does not look right in my opinion. Either disable helpers or display the arc helpers in the same color as the normal helpers (green).
grafik

@obelisk79  do you want to weigh in?

I've also come across this behavior when trying to drag the arc and it snaps to one point (where the dimension line is drawn from) and I cannot move the point. After some dragging the arc snaps back to its original position:

https://github.com/FreeCAD/FreeCAD/assets/6246609/084c925e-9f72-4b10-ad09-ece6a4574b0b

PaddleStroke:
If you feel that this is still not enough (and its true that while for radius it looks good, for distances it's still not ideal) then when the distance is in empty space, then we draw the full missing arc.

It would also remove this green/red issue.

PaddleStroke:
So it looks like this : 
image

maxwxyz:
regarding the Discord discussion: is this red also customizable in the preferences? Would it make sense to render the arc helper as construction geo (dashed)? currently the red lines are only used for dimension and constraints, not the helpers.
@obelisk79

PaddleStroke:
I think it wouldn't be better because then it would look like there is a construction geometry which there wouldn't be.
For me the arc here is part of the constraint drawing, the same as the lines at the beginning and end of constraints.

obelisk79:
I'm struggling with this one a bit... it is all visually distracting in its current state. here are my thoughts...

1. Line width should be assignable or mirror the setting for construction geometry
2. This should have it's own line-type, perhaps 'Helper geometry/line' with an assignable color/pattern
3. I would likely assign a custom color to this as either subdued line color (gray vs white sketch geometry, or desaturated red etc.)
4. Leveraging the new pattern PR (once merged of course) I would consider applying a dotted pattern to this, although some experimentation would likely be needed.

chennes:
> I would recommend we not push this to merge until after the line pattern PR is merged and we can look at leveraging it for this.

Agreed - I'm reviewing that PR right now, so hopefully we can get it merged shortly.

PaddleStroke:
The line pattern PR and this one are unrelated. The line PR does not touch the coin system of constraints, if we want to have dot lines or change the width / color of some parts in constraints it is a separate work.

chennes:
Let's leave fine-tuning this with the new line styles as future work with input from the DWG.

PR #12014, opened 2024-01-17
Toponaming ShapeMapper

bgbsww:


chennes:
So with that use of `std::hash` you are trying to do an injected specialization, but I don't see the implementation for `TopoDS_Shape` anywhere. You seem to be missing the 
```
template<>
struct std::hash<TopoDS_Shape>
{
    std::size_t operator()(const TopoDS_Shape& s) const noexcept
    {
        // Something like...
        return TopTools_ShapeMapHasher().HashCode(s, std::numeric_limits<StandardInteger>::max());
    }
};
```

bgbsww:
Yeah, that didn't explain why the other patch didn't break, and why OpenCascade was recommending using it without adding that code.  But you comment got me to think and search, and the answer is that the code does exist in TopoDS_Shape.hxx, but only at 7.8.0 and above.   So I need to go back and #if version the code, because one version is required before 7.8.0, and another is required at and after it.  Thanks!

src/Mod/Part/App/TopoShapeMapper.cpp
```diff
@@ -0,0 +1,75 @@
+#include "TopoShapeMapper.h"
+
+namespace Part
+{
+
+void ShapeMapper::expand(const TopoDS_Shape &d, std::vector<TopoDS_Shape> &shapes)
+{
+    if (d.IsNull()) return;
+    for(TopExp_Explorer xp(d, TopAbs_FACE);xp.More();xp.Next())
+        shapes.push_back(xp.Current());
```
chennes, line 10: Our clang-format file specifies that there should be curly braces around all loops and conditionals, even if they are only a single line. It's also got things to say about spacing around semicolons in for loops, etc. -- are you using an editor that can be configured to auto-apply it?

src/Mod/Part/App/TopoShapeMapper.cpp
```diff
@@ -0,0 +1,75 @@
...
+    for(TopExp_Explorer xp(d, TopAbs_EDGE, TopAbs_FACE);xp.More();xp.Next())
+        shapes.push_back(xp.Current());
+    for(TopExp_Explorer xp(d, TopAbs_VERTEX, TopAbs_EDGE);xp.More();xp.Next())
+        shapes.push_back(xp.Current());
+}
+
+void ShapeMapper::populate(bool generated,
```
chennes, line 17: IMO all of these `generated` booleans should be replaced by something along the lines of
```
enum class MappingStatusOrSomething {
    GENERATED,
    MODIFIED
};
```

src/Mod/Part/App/TopoShapeMapper.h
```diff
@@ -0,0 +1,245 @@
...
+
+    /** Insert a map entry from a sub shape in the source to a sub shape in the new shape
+     *
+     * @params generated: whether the sub shapes are generated or modified
+     * @param s: a sub shape in the source
+     * @param d: a list of sub shapes in the new shape
+     */
+    void insert(bool generated, const TopoDS_Shape &s, const TopoDS_Shape &d);
+
+    virtual const std::vector<TopoDS_Shape> &generated(const TopoDS_Shape &s) const override {
```
chennes, line 161:
```suggestion
    const std::vector<TopoDS_Shape> &generated(const TopoDS_Shape &s) const override {
```

src/Mod/Part/App/TopoShapeMapper.h
```diff
@@ -0,0 +1,245 @@
...
+        auto iter = _generated.find(s);
+        if(iter != _generated.end())
+            return iter->second.shapes;
+        return _res;
+    }
+
+    virtual const std::vector<TopoDS_Shape> &modified(const TopoDS_Shape &s) const override {
```
chennes, line 168:
```suggestion
    const std::vector<TopoDS_Shape> &modified(const TopoDS_Shape &s) const override {
```

tests/src/Mod/Part/App/TopoShapeMapper.cpp
```diff
@@ -0,0 +1,111 @@
...
+    void TearDown() override
+    {
+        App::GetApplication().closeDocument(_docName.c_str());
+    }
+
+private:
+    std::string _docName;
+};
+
+TEST_F(TopoShapeMapperTest, shapeHasherTests)
```
chennes, line 41: I personally prefer individual test methods for each major case or set of cases that are being checked. @JohnAD what are your thoughts on that?

src/Mod/Part/App/TopoShapeMapper.cpp
bgbsww: Yep, VSCode will do that.   I'll set it up.  I had been shying away from modifying copied in code as much as possible; I'll bring it up to format instead.

src/Mod/Part/App/TopoShapeMapper.cpp
chennes: Oh, I guess in modern-ese those shouldn't be all-caps, huh? Old dog, new tricks, etc. `Generated` and `Modified` then, or some such.

tests/src/Mod/Part/App/TopoShapeMapper.cpp
bgbsww: Yeah, I wavered on this one.   As I was writing it, it was useful to not repeat the setup and keep adding test cases, but it makes sense to split it out some now.

It seems to me that 14 individual test methods would be too many in this case; I'll aim for about 3-4 to keep the cognitive load reasonable.  That make sense, @chennes ?  Maybe a guideline number of assertions?

chennes: @JohnAD is the testing expert here -- I will go with the flow.

src/Mod/Part/App/TopoShapeMapper.h
```diff
@@ -0,0 +1,273 @@
...
+    };
+    typedef std::unordered_map<TopoDS_Shape, ShapeValue, ShapeHasher, ShapeHasher> ShapeMap;
+    ShapeMap _generated;
+    std::unordered_set<TopoDS_Shape, ShapeHasher, ShapeHasher> _generatedShapes;
+    ShapeMap _modified;
+    std::unordered_set<TopoDS_Shape, ShapeHasher, ShapeHasher> _modifiedShapes;
+};
+
+/// Parameters for TopoShape::makeElementFilledFace()
+struct PartExport TopoShape::BRepFillingParams
```
chennes, line 216: Are the variable names here matching something in OCCT? It's a little Pythonic.

PR #12022, opened 2024-01-17
[Core] Wb selector : adds "Toolbar icons" mode

PaddleStroke:
Workbench selector : adds "Toolbar icons" mode.
image

This is adding menu entries with the icon of the workbenches directly on the QMenuBar. Inspired from Triplus' Selector toolbar. (https://forum.freecad.org/viewtopic.php?t=21403)

If someone on linux and on MacOS can feedback if it works properly?

Ideas and feedback welcome.

### DIFF
--- a/src/Gui/Action.cpp
+++ b/src/Gui/Action.cpp
@@ -688,11 +688,20 @@ void WorkbenchGroup::addTo(QWidget *widget)
         qobject_cast<QToolBar*>(widget)->addWidget(box);
     }
     else if (widget->inherits("QMenuBar")) {
-        auto* box = new WorkbenchComboBox(widget);
-        setupBox(box);
+        bool leftCorner = WorkbenchSwitcher::isLeftCorner(WorkbenchSwitcher::getValue());
+        bool rightCorner = WorkbenchSwitcher::isLeftCorner(WorkbenchSwitcher::getValue());
+        QMenuBar* menuBar = qobject_cast<QMenuBar*>(widget);
 
-        bool left = WorkbenchSwitcher::isLeftCorner(WorkbenchSwitcher::getValue());
-        qobject_cast<QMenuBar*>(widget)->setCornerWidget(box, left ? Qt::TopLeftCorner : Qt::TopRightCorner);
+        if (leftCorner || rightCorner) { // corner widget
+            auto* box = new WorkbenchComboBox(widget);
+            setupBox(box);
+            menuBar->setCornerWidget(box, leftCorner ? Qt::TopLeftCorner : Qt::TopRightCorner);
+        }
+        else { // QMenuBar icons
+            for (auto* action : actions()) {
+                menuBar->addAction(action);
+            }
+        }
     }
     else if (widget->inherits("QMenu")) {
         auto menu = qobject_cast<QMenu*>(widget);
--- a/src/Gui/MenuManager.cpp
+++ b/src/Gui/MenuManager.cpp
@@ -27,6 +27,8 @@
 # include <QMenuBar>
 #endif
 
+# include <QProxyStyle>
+
 #include "MenuManager.h"
 #include "Application.h"
 #include "Command.h"
@@ -199,6 +201,21 @@ MenuManager::MenuManager() = default;
 
 MenuManager::~MenuManager() = default;
 
+class SeparatorActsAsSpacer : public QProxyStyle {
+public:
+    using QProxyStyle::QProxyStyle; // Inherit constructors
+
+    int styleHint(StyleHint hint, const QStyleOption* option = nullptr,
+        const QWidget* widget = nullptr, QStyleHintReturn* returnData = nullptr) const override {
+        if (hint == SH_DrawMenuBarSeparator) {
+            return true;
+        }
+        else {
+            return QProxyStyle::styleHint(hint, option, widget, returnData);
+        }
+    }
+};
+
 void MenuManager::setup(MenuItem* menuItems) const
 {
     if (!menuItems) {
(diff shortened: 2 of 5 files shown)


qewer33:
I've tested this PR and a couple of things to note:

1. It doesn't seem to work with a global menubar on Linux (KDE Plasma)

image

2. There's an abrupt change in the style (?) of the menubar area when a WB button is clicked

[vokoscreenNG-2024-01-17_19-47-59.webm](https://github.com/FreeCAD/FreeCAD/assets/69015181/e8be457a-4c8a-48ee-8a18-5b96dd0a7cd5)

3. The menubar behaves weirdly when the windows is scaled

image

image

PaddleStroke:
1. Ah so here goes my hope that this would work on global menus.
Though perhaps it's only the trick with the style to get the icons to the right that is not accepted. Or it's the icons, maybe only text menu entries are accepted.

2. Regarding the other behavior your describe : It's because your window is too small so when you go to partdesign, you have 2 more menus added (sketcher and partdesign) and with all your workbench it does not fit anymore.

qewer33:
Regarding 2, I mean the style change not the resize issue. Notice how the gray background of the menubar turns into dark blue and the hover effect changes. I know the styling is from the theme but there shouldn't be a reason for the style class to change abruptly.

PaddleStroke:
Ah yes I see. Hmm I don't know what can cause this.

Btw can you please try again Global Menu after commenting out 
        `menuBar->setStyle(new SeparatorActsAsSpacer(QApplication::style()));`

qewer33:
Nope, still doesn't work after that change.

PaddleStroke:
I have set it to draft until further discussion is done.

chennes:
FWIW, this does not work on a Mac, you just get nothing (which is not a surprise, considering how MacOS handles the menubar).

obelisk79:
This PR is a deviation from the guidelines for the menu bar as outlined in the developer handbook.

MisterMakerNL:
Could you try it on the status-bar, reason is that in windows and in mac the programs are also on the bar below. 

Programs are similar to workbenches. So it could behave similar to a program bar in any os.
We can make the status bar bigger. Add a clock to it :) .

obelisk79:
Putting these into the status bar is likely a 'better' approach than the menu bar, but still violates several design principles and would again be a deviation from the design guide. I suspect such a solution would satisfy an extremely niche subset of users, and adding items to the status bar could be done through a python add-on. Therefore I would caution against incorporating that into the main codebase.

MisterMakerNL:
This is how it looks:
image
@obelisk79 Could you please say which design principles so I or we can improve on it, just saying it violates multiple isn't really useful feedback!

obelisk79:
From the Design Guide:

> 
Menu Bar: Standard height is established as 32 pixels in the stylesheet. There shall be no insertion of non-standard elements (ie. QComboBox) in this area by default. New menus can be added, however they should ideally be confined to the menu named after your specific workbench. Simple addons should insert entries into a menu titled "Accessories" if applicable.

Fitts Law: Size of the icons (ie target for the mouse to register and click), Menu/Status bar would have to increase in size for this to be viable.

Fitts Law/Doherty Threshold: Locating on the status bar dramatically increases travel distance, affecting reaction time

Jakob's Law: Using the menu bar in such a way deviates from behavior which is effectively universal across desktop software (non ribbon bar applications)

Law of Common Region/Law of Proximity: Putting WB navigation in the status bar places the interactive elements on the opposite edge of the screen from where all of the related actionable functions are.

Law of Pragnanz: Many of the WB icons do not scale well to this size, and do not effectively convey what they represent. More of an issue for new users than existing, which may not be worth discussing here as this would be optional.

I would advise that the menu bar should be considered an absolute sacred part of the UI which should *only* consist of menus. Regarding the proposed status bar implementation above, I would say marginally acceptable, however not to be considered as a default setting/element.

FlachyJoe:
> Anyway this is how it looks, if you move it to the status bar. 

IMHO the tab style in the status bar is confusing. Click on a WB icon won't change the UI just above but at the window top (i.e. toolbars). Action and effect are too far.

MisterMakerNL:
Ok some good points yet, we gotta deal with it.
And I don't see any better options.
I still don't understand why it should be confusing you have exact the same setup in almost every os. 
Bottom bar apps or program or in our case workbenches. And on the other site the status icons.
It's pretty logical, if you put the "start" wb in first icon than it even similar to the "start" icon...
If it's against a design guide then you are saying every OS is doing it wrong.
Only MAC moves the status bar up
image
Windows moves it down but has no menu.
image
Linux is all over the place tho, wtf linux why can't you be normal?

sliptonic:
I think we're overreaching.    Let's take a baby step, get it in front of real users, and then see if we can take another.

Based on survey data, both TabBar and the toolbar selector are being run by quite a few people even though they're unmaintained.  That tells us that people find value in the approach.  So don't deviate too far from that.  

Both are implemented in python. They have a lot of overlap but each has strengths and weaknesses.  A good first step is to just build a better tabbar taking  the best of both and eliminating the bad.

Instead of an addon, include it in the core.  Let the user select it from preferences like @PaddleStroke implemented in this PR.  Instead of 'menu bar icons' it should be toolbar icons'  This would turn off the combo selector and add a toolbar.  The user can place it wherever they want including the bottom.   Don't add a menu 'accessories' like the python versions do.

Just like triplus's implementation let the user _easily_ configure, which tabs are shown and whether labels are shown

Give them the option of a button that shows all workbenches.  Then it's tempting to whittle down the number of tabs because the less frequently used ones are just one click away.

Keep this first step simple and this will be a giant leap forward.

PaddleStroke:
Yes I have new plans. I will make a widget same as triplus tab bar, and
also offer the possibility to put it either on the toolbar or on the menu
bar.

ninao-xyz:
hello. I'm a user of the "selector tool bar", and I just want to leave my input of why I like it.

- It can be used horizontally or vertically, personally I have it vertically on the left side, this is nice because my computer screen is larger than taller, and it doesn't stay "in the way" of the toolbars of the workbench being used.
- Can  customize the order of the icons.
- can hide the icons of "unused" workbenches in a sub-menu, for example I don't use FEM/arch/openscad/web so I hide them in the sub-menu, yet if I need them, they are still easily reachable.
- Can be customized to display icons only, text only, or both. personally I use icons only, after some time of use, my brain interprets the icons as if it is a full word, consequently it saves screen space.

If I my "ask" for the new toolbar being developed the thing that I wish to get implemented is the ability to have it either in a vertical or horizontal position.

That being said, I'm just a user not a developer, I don't understand the technical challenges , and I trust the judgement of freecad developers. 

I'll leave a screenshot below of my FreeCad window where I have the selector toolbar.

Screenshot from 2024-01-23 16-40-45

PaddleStroke:
Closed in favor of https://github.com/FreeCAD/FreeCAD/pull/12270

Issue #11072, opened 2023-10-17
[Request] UI improvement: Add properties view to Right click and put it in a seperate window.


MisterMakerNL:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

This is an old one that never got done.
I was thinking/hoping now @PaddleStroke is doing the transparency in the right click menu, it is an easy thing to do.
The idea is that properties view is something advanced for advance users and should not be necessary for simple things. So should be off by default. A common user should just be able to right click on something and open properties. Just like a lot of programs work.
Also myself  don't use the properties panel enough to keep it, I would just use right-click.

But how it works now you either have it on or you don't. One could say it needs to use a short-cut, but this complicate things unnecessary.

But I think the request is pretty obvious. 
@Zolko-123  did a pre-study to this:
https://forum.freecad.org/viewtopic.php?t=33831
![LfS1cXw](https://github.com/FreeCAD/FreeCAD/assets/29804962/e3be7d1f-038a-47af-a95d-297350512550)

### Full version info

```shell
NA
```

### Subproject(s) affected?

None

### Anything else?

UI
UX
Properties

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PaddleStroke:
> The idea is that properties view is something advanced for advance users and should not be necessary for simple things. So should be off by default. A common user should just be able to right click on something and open properties. Just like a lot of programs work.

+1 this is on my mind for some time already.

yorikvanhavre:
I love the idea of being able to turn the properties view on/off with a key shortcut. Sometimes one has a very long tree and it helps to navigate it when the properties view is off.
Not sure about the use of F1 though, because ATM F1 is used for help, and that's also pretty common...

luzpaz:
Anyone want to take point on this ?

MisterMakerNL:
Ok here is the idea how we can probably do this real easy.
We just add a "properties" option to the right click, and it would just pop-up the properties panel.
Right clicking on a object switched the panel to the object anyway.
I already made a [macro to pop-up the panel.](https://github.com/MisterMakerNL/Freecad_Hide_docks_macros/blob/main/show_combo_prop_view.FCMacro)

[We can also make an option to prevent the panel to dock](https://forum.qt.io/topic/85996/how-to-disable-snapping-docking-of-qdockwidget-to-the-centeral-widget/2), so it stays a window. This is very much preferable since if it would pop-up docked the user will probably not notice it.

[...]

PR #12024, opened 2024-01-17
Core : Tree: Add 'Properties' action in contextual menu. Opens a prop…

PaddleStroke:
Tree: Add 'Properties' action in contextual menu. Opens a property dialog.
Does not affect the 'Property View' at all.
image

Fixes https://github.com/FreeCAD/FreeCAD/issues/11072

### DIFF
--- a/src/Gui/CommandDoc.cpp
+++ b/src/Gui/CommandDoc.cpp
@@ -50,6 +50,7 @@
 #include "BitmapFactory.h"
 #include "Command.h"
 #include "Control.h"
+#include "DockWindowManager.h"
 #include "FileDialog.h"
 #include "MainWindow.h"
 #include "Selection.h"
@@ -1715,6 +1716,41 @@ bool StdCmdEdit::isActive()
     return (!Selection().getCompleteSelection().empty()) || (Gui::Control().activeDialog() != nullptr);
 }
 
+//===========================================================================
+// Std_Properties
+//===========================================================================
+DEF_STD_CMD_A(StdCmdProperties)
+
+StdCmdProperties::StdCmdProperties()
+    : Command("Std_Properties")
+{
+    sGroup = "Edit";
+    sMenuText = QT_TR_NOOP("Properties");
+    sToolTipText = QT_TR_NOOP("Show the property view, which displays the properties of the selected object.");
+    sWhatsThis = "Std_Properties";
+    sStatusTip = sToolTipText;
+    sAccel = "Alt+Return";
+    sPixmap = "document-properties";
+    eType = Alter3DView;
+}
+
+void StdCmdProperties::activated(int iMsg)
+{
+    Q_UNUSED(iMsg);
+    QWidget* propertyView = Gui::DockWindowManager::instance()->getDockWindow("Property view");
+    if (propertyView) {
+        QWidget* parent = propertyView->parentWidget();
+        if (parent && !parent->isVisible()) {
+            parent->show();
+        }
+    }
+}
+
+bool StdCmdProperties::isActive()
+{
+    return !Selection().getCompleteSelection().empty();
+}
+
 //======================================================================
 // StdCmdExpression
 //===========================================================================
@@ -1969,6 +2005,7 @@ void CreateDocCommands()
     rcCmdMgr.addCommand(new StdCmdTransformManip());
     rcCmdMgr.addCommand(new StdCmdAlignment());
     rcCmdMgr.addCommand(new StdCmdEdit());
+    rcCmdMgr.addCommand(new StdCmdProperties());
     rcCmdMgr.addCommand(new StdCmdExpression());
 }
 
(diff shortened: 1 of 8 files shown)


adrianinsaval:
could we insert the property view inside the task view instead of opening an independent window?

PaddleStroke:
It is an interesting idea. @obelisk79 @MisterMakerNL @kadet1090 @qewer33 ?

kadet1090:
> could we insert the property view inside the task view instead of opening an independent window?

What would be the reason for that? IMHO if anything we should create more more capable docking system that allow to put any panels as tabs or something like that.

MisterMakerNL:
> could we insert the property view inside the task view instead of opening an independent window?

No this won't work the workflow I had in mind was, if you need to change properties you open the properties window, do what you want close it when you are done.
If you dock it with task-view, you will close the task view with the properties. (or you need to make something complicated that only closes the properties panel, but I think it will get confusing)
Keeping it open by default is a waste of screen-space. Unless you need it a lot then you can still use the panel. Especially with the new overlay system where the task-view auto-hides.

I think this fix looks awesome! It cleans up the UI massively. Very nice work! Very happy with this, hope others will like it too!

sliptonic:
I like this a lot.
The handling of property pane has always been sub-optimum.   It always occupies real estate even if the user doesn't need it.  When the user DOES need it, it's too small and always requires scrolling.   This is a very nice alternative but it needs to work very reliably and with good keyboard acceleration.

There's an accelerator key (F6) showing in the context menu but it doesn't work to open the property dialog.

PaddleStroke:
I think there is a conflict with another tool for F6.
I am not sure what shortcut we can assign. I would have gone with F1 but it's help. F3 F4 are doing some overlay stuff. etc.
Is there a standard key for property?

PaddleStroke:
Reading online Alt+Enter seems to be a 'classic' for properties. Any thoughts on that?

sliptonic:
Alt-Enter is ok if it's standard.   It's two keys for something that is likely to be common.   'P' for properties would be nice and 'I' for info if we ever have a replacement for FCInfo.

sliptonic:
We should probably have some design guidance on accelerator assignment

adrianinsaval:
Alt+Enter is good IMO, my file manager (kde's dolphin) has that shortcut

sliptonic:
Can you make the panel remember the last place and size when closed and reopened?

kadet1090:
> I think there is a conflict with another tool for F6. I am not sure what shortcut we can assign. I would have gone with F1 but it's help. F3 F4 are doing some overlay stuff. etc. Is there a standard key for property?

We can always change the overlay stuff it isn't stable yet so we can do that without much problem., Besides, I doubt that anyone will notice lack of this shortcuts and it Wil free up those keys.

adrianinsaval:
We could do that but using something that's standard is better

PaddleStroke:
So I was starting to follow @sliptonic idea and save the dialog size and position. And from 7 lines of code, it started to be a custom QDialog class that kept growing.

Which made me think again about what @adrianinsaval said in the issue : why making a new view when we already have one.

My reasoning was that making a dialog was only few lines of code and clean. But if we start adding features to the dialog then it is not true anymore. And it becomes obvious that the right choice is not introducing a new dialog, but instead fix the overlay stuff so that the current PropertyView overlay dialog can be used. So here is a list of things that seems necessary : 
- The property action just make the PropertyView visible.
- PropertyView should save its state : dockDialog or overlay dialog? Position, size etc.

On a side note, I cannot get Alt + Enter to work. Needs a bit more of digging apparantly.

PaddleStroke:
Ok so here is what I have done : 

- Added a 'Std_Properties' command that shows the property view. (this is required for the shortcut to work).
- Command added to the Edit menu.
- Command added to the tree item right click menu.
- Alt + Return shortcut working.
- The command only toggles the visibility of the property view.
- Bonus: Save the width of the first column when it is resized. Before it would always reinitialize on restart.

So where does that brings us ? You can achieve the same as with the QDialog I made before by dragging the property view outside of a dock area and resizing it : 
image

The issue is that the system does not save the fact that property view is a floating window. So on restart the system will forget and Alt+Return will open the PropertyView as a dock-widget docked on the left.

So what we need is to save in preference the position and size of the property view (docked vs floating vs overlay). But this is probably overlapping @kadet1090 PR : https://github.com/FreeCAD/FreeCAD/pull/11998 so I don't want to change something at the same time. Any thoughts on this btw @kadet1090 ?

I would say that this PR can already merge as it is, the save fix not being overlapping this PR.

Roy-043:
Another issue: AFAICT the command does not do anything if "Tree view mode" is set to "Combo View".
https://wiki.freecad.org/Preferences_Editor#General_2

Roy-043:
The command does not behave as a toggle. It does not hide the Property View. I don't understand the logic that pre-selecting an object is required to enable this command. The command can work even if there is no document, so why not allow that?

PaddleStroke:
The logic was to have the same as most other software where you usally right click on something, select properties and get a property dialog.
It was not meant as a 'toggle property view visibility'.
In fact initially it was opening a QDialog showing the properties of the object. But then it was noticed that QDialog was kind of the same as the current property view when set as overlay.

Anyway @pieterhijma is working on a 'Property Manager' dialog. I think that when that merge, this command should then be changed to open the new 'Property Manager' instead of showing the property view.

PR #12035, opened 2024-01-18
Display current active document in main window title

### LINKED ISSUE #11858: [Feature Request] Add option to set the current project name in the window title

### Problem description

Hey!
I really enjoy using FreeCAD. One issue I found is that I can't really track the time I spend in the application. While I do know that adding custom telemetry in FreeCAD for users to know how much time they spent in a project I request a minimal change: add the current project name in the window title.
image
The window title can be used by other applications such as https://activitywatch.net (FOSS time tracking application) to track how much time one spent on a project:
image
It could be something like:
`FreeCAD 0.22.0dev ~ /home/user/something/visionfive2-case.FCStd`

### Full version info

```shell
OS: Solus 4.5 Resilience (Budgie:GNOME/budgie-desktop)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35477 (Git)
Build type: Release
Branch: main
Hash: 1a59b5846caaf903e23917dd1f4081fd698ab2d2
Python 3.10.13, Qt 5.15.11, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.2
Installed mods: 
  * IconThemes
  * Render 2023.8.18
  * Assembly3 0.12.0
  * fasteners 0.4.75
  * OpenDark 2023.12.17
```

### Subproject(s) affected?

None

### Anything else?

_No response_

### Code of Conduct

maxwxyz:
fixes #11858 
current active document is shown in the main window title bar. An asterisk is displayed in front, if unsaved changes exist.

current, always static:
grafik

new:
grafik

### DIFF
--- a/src/Gui/MDIView.cpp
+++ b/src/Gui/MDIView.cpp
@@ -453,7 +453,10 @@ QString MDIView::buildWindowTitle()
     QString windowTitle;
     if (Gui::Document* document = getGuiDocument()) {
         if (document->isModified()) {
-            windowTitle = QString::fromUtf8("* ");
+            getMainWindow()->setWindowModified(TRUE);
+        }
+        else {
+            getMainWindow()->setWindowModified(FALSE);
         }
 
         windowTitle.append(QString::fromUtf8(getAppDocument()->getName()));
--- a/src/Gui/MDIView.cpp
+++ b/src/Gui/MDIView.cpp
@@ -448,4 +448,18 @@ void MDIView::setCurrentViewMode(ViewMode mode)
     }
 }
 
+QString MDIView::buildWindowTitle()
+{
+    QString windowTitle;
+    if (Gui::Document* document = getGuiDocument()) {
+        if (document->isModified()) {
+            windowTitle = QString::fromUtf8("* ");
+        }
+
+        windowTitle.append(QString::fromUtf8(getAppDocument()->getName()));
+    }
+
+    return windowTitle;
+}
+
 #include "moc_MDIView.cpp"
--- a/src/Gui/MainWindow.cpp
+++ b/src/Gui/MainWindow.cpp
@@ -1270,6 +1270,7 @@ void MainWindow::tabChanged(MDIView* view)
 {
     Q_UNUSED(view);
     updateActions();
+    getMainWindow()->setWindowTitle(view->buildWindowTitle());
 }
 
 void MainWindow::tabCloseRequested(int index)
@@ -1665,6 +1666,7 @@ void MainWindow::_updateActions()
         Application::Instance->commandManager().testActive();
     }
     d->actionUpdateDelay = 0;
+    getMainWindow()->setWindowTitle(activeWindow()->buildWindowTitle());
 }
 
 void MainWindow::updateEditorActions()
@@ -2368,6 +2370,7 @@ void MainWindow::changeEvent(QEvent *e)
                     d->activeView = view;
                     Application::Instance->viewActivated(view);
                 }
+                getMainWindow()->setWindowTitle(view->buildWindowTitle());
             }
         }
     }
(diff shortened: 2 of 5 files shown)


src/Gui/Application.cpp
```diff
@@ -2145,28 +2145,12 @@ void Application::runApplication()
...
-        QString point = QString::fromLatin1(config["BuildVersionPoint"].c_str());
-        QString suffix = QString::fromLatin1(config["BuildVersionSuffix"].c_str());
-        QString title =
-            QString::fromLatin1("%1 %2.%3.%4%5").arg(mainApp.applicationName(), major, minor, point, suffix);
-        mw.setWindowTitle(title);
-    }
-    else {
-        mw.setWindowTitle(mainApp.applicationName());
-    }
+    mw.setWindowTitle(QString::fromLatin1(""));
```
kadet1090, line 2148: Please add comment on why we pass empty string here

src/Gui/MainWindow.cpp
```diff
@@ -1665,6 +1666,11 @@ void MainWindow::_updateActions()
         Application::Instance->commandManager().testActive();
     }
     d->actionUpdateDelay = 0;
+    QMdiSubWindow* mdi = d->mdiArea->currentSubWindow();
+    if (mdi) {
+        auto view = dynamic_cast<MDIView*>(mdi->widget());
```
kadet1090, line 1671: This could be a good helper, something like `getCurrentWindow`/`getCurrentView` then this function would not care about how to access current view.

src/Gui/MDIView.cpp
```diff
@@ -448,4 +448,14 @@ void MDIView::setCurrentViewMode(ViewMode mode)
     }
 }
 
+QString MDIView::buildWindowTitle() {
+    QString windowTitle = QString::fromLatin1("");
+    if (getGuiDocument()->isModified()) {
+        windowTitle = QString::fromLatin1("* ");
```
kadet1090, line 454: QT windows (and by extension MainWindow) have "setWindowModified"  flag that should do that automatically - can you try that?

src/Gui/MDIView.cpp
```diff
@@ -448,4 +448,14 @@ void MDIView::setCurrentViewMode(ViewMode mode)
...
 }
 
+QString MDIView::buildWindowTitle() {
+    QString windowTitle = QString::fromLatin1("");
+    if (getGuiDocument()->isModified()) {
+        windowTitle = QString::fromLatin1("* ");
+    }
+
+    windowTitle = windowTitle + QString::fromLatin1(getAppDocument()->getName());
+    return windowTitle;
```
kadet1090, line 458:
```suggestion
    return windowTitle + QString::fromLatin1(getAppDocument()->getName());
```

src/Gui/MainWindow.cpp
```diff
@@ -2509,6 +2516,33 @@ QMdiArea *MainWindow::getMdiArea() const
...
+    }
+    else {
+        title = appname;
+    }
+    if (!string.isEmpty()) {
+        title = QString::fromLatin1("%1 - %2").arg(string, title);
+    }
+    QMainWindow::setWindowTitle(title);
+}
+
```
kadet1090, line 2545: This function reads as a big blob of text - tray to add empty lines before  if statements, before last line and before `// allow ...` comment just like it was in the original code. It should be much easier on eyes.

kadet1090 (changes requested):

src/Gui/MDIView.cpp
maxwxyz: I've tried that but does not work. Also I can add more options in the future now with this solution.

maxwxyz:
@kadet1090 should be good now. The builds are failing because of Gui tests, are they somehow looking at the window title or how could I fix it?

kadet1090 (approved):

kadet1090:
From my perspective it now looks fine, we need however someone more experienced in FreeCAD codebase to check it additionally. About the tests - I'm not sure, it seems that the `d` (Qt docs: https://wiki.qt.io/D-Pointer) pointer is null. Maybe it is called after widget was already freed, or not yet created fully... I won't be able to help here much.

yorikvanhavre:
This is good! CI build issues are fixed I think... I'll relaunch the tests

maxwxyz:
@yorikvanhavre still failing :(

maxwxyz:
@chennes any news on this - is there a problem with the workflows or in the code?

maxwxyz:
@chennes should I ping someone else?

maxwxyz:
@yorikvanhavre do you want to take another look at this?

maxwxyz:
Still the Ubuntu UI tests seem to fail, I have no experience with these workflows.

wwmayer:
`if (getGuiDocument()->isModified())`
getGuiDocument() can return null

maxwxyz:
> `if (getGuiDocument()->isModified())` getGuiDocument() can return null

so would a fix be 
`if (getGuiDocument() != nullptr && getGuiDocument()->isModified())`

or do I need to look for something else?

src/Gui/Application.cpp
```diff
@@ -2425,12 +2425,232 @@ void Application::runApplication()
...
     initGuiAppPreMainWindow(false);
-	
+
     Application app(true);
     MainWindow mw;
     mw.setProperty("QuitOnClosed", true);
 
-    initGuiAppPostMainWindow(false, mainApp, mw, &mainApp);
+    // empty window title QString sets default title (app + version)
+    mw.setWindowTitle(QString::fromLatin1(""));
```
wwmayer, line 2434: Actually you can directly write `mw.setWindowTitle(QString());`

src/Gui/MDIView.cpp
```diff
@@ -448,4 +448,14 @@ void MDIView::setCurrentViewMode(ViewMode mode)
     }
 }
 
+QString MDIView::buildWindowTitle()
+{
+    QString windowTitle = QString::fromLatin1("");
```
wwmayer, line 453: To create an empty string simply write `QString windowTitle;`

src/Gui/MDIView.cpp
```diff
@@ -448,4 +448,14 @@ void MDIView::setCurrentViewMode(ViewMode mode)
     }
 }
 
+QString MDIView::buildWindowTitle()
+{
+    QString windowTitle = QString::fromLatin1("");
+    if (getGuiDocument()->isModified()) {
```
wwmayer, line 454: As said in the forum getGuiDocument() can return null.

src/Gui/MDIView.cpp
```diff
@@ -448,4 +448,14 @@ void MDIView::setCurrentViewMode(ViewMode mode)
...
 }
 
+QString MDIView::buildWindowTitle()
+{
+    QString windowTitle = QString::fromLatin1("");
+    if (getGuiDocument()->isModified()) {
+        windowTitle = QString::fromLatin1("* ");
+    }
+
+    return windowTitle + QString::fromLatin1(getAppDocument()->getName());
```
wwmayer, line 458: Currently, the internal name of a document is 7-bit ASCII but this will change in the future. So, you better use `QString::fromUtf8`

src/Gui/MainWindow.cpp
```diff
@@ -2509,6 +2512,45 @@ QMdiArea *MainWindow::getMdiArea() const
...
+    if (!string.isEmpty()) {
+        title = QString::fromLatin1("%1 - %2").arg(string, title);
+    }
+
+    QMainWindow::setWindowTitle(title);
+}
+
+MDIView *MainWindow::getCurrentView()
+{
```
wwmayer, line 2547: What's the difference to `MainWindow::activeWindow()`?

src/Gui/MainWindow.cpp
```diff
@@ -2509,6 +2512,45 @@ QMdiArea *MainWindow::getMdiArea() const
...
+    QMdiSubWindow* mdi = getMdiArea()->currentSubWindow();
+    if (mdi) {
+        return dynamic_cast<MDIView*>(mdi->widget());
+    }
```
wwmayer, line 2551: And what happens in the `else` case?

maxwxyz: this was just intended to check if mdi exists. If not it should not create a custom title.

wwmayer: But you still have to return a value to avoid undefined behaviour.

src/Gui/MDIView.cpp
maxwxyz: @kadet1090 FYI

src/Gui/MainWindow.cpp
maxwxyz: thanks for pointing this out, I've used this. The next remark would be needless then.

maxwxyz:
@wwmayer thanks I've considered your review. Unfortunately it still throws the test errors on GH for Ubuntu, not sure if this is an error in the code or if the workflows need to be adjusted.

wwmayer:
getAppDocument() is a convenience method to get the App document of a Gui document. If getGuiDoucment() returns null then getAppDocument() will do so, too.

It's best to write the method this way:
```
QString MDIView::buildWindowTitle()
{
    QString windowTitle;
    if (Gui::Document* document = getGuiDocument()) {
        if (document->isModified()) {
            windowTitle = QString::fromUtf8("* ");
        }

        windowTitle.append(QString::fromUtf8(getAppDocument()->getName()));
    }

    return windowTitle;
}
```

maxwxyz:
thanks a lot for your help!

xtemp09:
```ruby
QString major = QString::fromUtf8(config["BuildVersionMajor"].c_str());
```

Can I ask, why is there a conversion via `const char *`? Has `QString QString::fromStdString(const std::string &str)` [^1] been considered buggy?

[^1]: https://doc.qt.io/qt-5/qstring.html#fromStdString

<details><summary>Fragment</summary>


```ruby
auto config = App::Application::Config();
QString major = QString::fromUtf8(config["BuildVersionMajor"].c_str());
QString minor = QString::fromUtf8(config["BuildVersionMinor"].c_str());
QString point = QString::fromUtf8(config["BuildVersionPoint"].c_str());
QString suffix = QString::fromUtf8(config["BuildVersionSuffix"].c_str());
title = QString::fromUtf8("%1 %2.%3.%4%5").arg(appname, major, minor, point, suffix);
```

</details>

I noticed that this fragment occurs too many times in the source code. Would it be better to write a separate function that returns the version in a form "%1.%2.%3%4"?

maxwxyz:
That would be for another PR, I just used the existing code for that.

wwmayer:
> Has QString QString::fromStdString(const std::string &str) been considered buggy?

In the past (Qt4) if a std::string contained a utf-8 encoded string then using fromStdString() caused problems because it internally used fromLatin1() or fromAscii() and created a broken QString.

With Qt5 this has changed and fromStdString() can be used for every std::string.

maxwxyz:
@chennes from my side it would be ready

xtemp09:
When something is modified, * is added in the end of the name, like in _TechDraw_

https://github.com/FreeCAD/FreeCAD/blob/0d37ea26291b282f59b55210c43cd612d23398de/src/Mod/TechDraw/Gui/MDIViewPage.cpp#L114-L116
or _Spreadsheet_
https://github.com/FreeCAD/FreeCAD/blob/0d37ea26291b282f59b55210c43cd612d23398de/src/Mod/Spreadsheet/Gui/ViewProviderSpreadsheet.cpp#L140-L141
or here
https://github.com/FreeCAD/FreeCAD/blob/0d37ea26291b282f59b55210c43cd612d23398de/src/Gui/MDIView.cpp#L170-L172
or here
https://github.com/FreeCAD/FreeCAD/blob/0d37ea26291b282f59b55210c43cd612d23398de/src/Gui/TextDocumentEditorView.cpp#L102-L103

In the commit the opposite is done, asterisk prepends the name:

```cpp

if (document->isModified()) {
	windowTitle = QString::fromUtf8("* ");
}
		
windowTitle.append(QString::fromUtf8(getAppDocument()->getName()));

```

Would it be better to add asterisk in the end of the title?


----


I also noticed that there exist Base::Tools function:
https://github.com/FreeCAD/FreeCAD/blob/0d37ea26291b282f59b55210c43cd612d23398de/src/Base/Tools.h#L300-L303

wwmayer:
> I also noticed that there exist Base::Tools function:

This function once was needed as convenience for the above explained reason. Now it's obsolete and can be removed.

maxwxyz:
> Would it be better to add asterisk in the end of the title?
> 

In most software the asterisk is placed in front.
Also when viewing window preview or using tab to switch between open windows you quickly see the ones which are unsaved.

wwmayer:
https://doc.qt.io/qt-6/qwidget.html#windowTitle-prop

maxwxyz:
@wwmayer thanks for pointing this out, I've changed it to use `setWindowModified()`.

maxwxyz:
@wwmayer ping

yorikvanhavre:
This seems good to go, and all the requests have been responded, so, merging!

src/Gui/MainWindow.cpp
```diff
@@ -2509,97 +2512,132 @@ QMdiArea *MainWindow::getMdiArea() const
...
-    msg = QString::fromLatin1("#statusBar{color: #000000}"); // black
-    wrn = QString::fromLatin1("#statusBar{color: #ffaa00}"); // orange
-    err = QString::fromLatin1("#statusBar{color: #ff0000}"); // red
-    Base::Console().AttachObserver(this);
-    getWindowParameter()->Attach(this);
-    getWindowParameter()->NotifyAll();
-}
+    QString title;
+    QString appname =
+        QString(static_cast<QApplication*>(QCoreApplication::instance())->applicationName());
```
wwmayer, line 2519: This is awkward. applicationName() is a static method of QCoreApplication.

src/Gui/MainWindow.cpp
```diff
@@ -2509,97 +2512,132 @@ QMdiArea *MainWindow::getMdiArea() const
...
+        QString minor = QString::fromUtf8(config["BuildVersionMinor"].c_str());
+        QString point = QString::fromUtf8(config["BuildVersionPoint"].c_str());
+        QString suffix = QString::fromUtf8(config["BuildVersionSuffix"].c_str());
+        title = QString::fromUtf8("%1 %2.%3.%4%5").arg(appname, major, minor, point, suffix);
     }
-    else if (strcmp(sReason, "colorError") == 0) {
-        unsigned long col = rclGrp.GetUnsigned( sReason );
-        this->err = format.arg(App::Color::fromPackedRGB<QColor>(col).name());
+    else {
+        title = appname;
```
wwmayer, line 2536: This re-introduces the bug that was fixed with https://github.com/maxwxyz/FreeCAD/commit/58a53e929d0ea178c7aa6a09e60d1734a9db3940#diff-1fde83bf12f7bde0aa892591a29dad0564b3e2a4ccd8be96ac6371f0beac270fR246
If FreeCADGui is used as Python module the title lacks of the application name.

src/Gui/MainWindow.cpp
```diff
@@ -1665,6 +1666,7 @@ void MainWindow::_updateActions()
         Application::Instance->commandManager().testActive();
     }
     d->actionUpdateDelay = 0;
+    getMainWindow()->setWindowTitle(activeWindow()->buildWindowTitle());
```
wwmayer, line 1669: What if there is no active window at all? Seems like the reason for the crash.

maxwxyz: to get the window modified asterisk

src/Gui/MDIView.cpp
```diff
@@ -448,4 +448,21 @@ void MDIView::setCurrentViewMode(ViewMode mode)
     }
 }
 
+QString MDIView::buildWindowTitle()
+{
+    QString windowTitle;
```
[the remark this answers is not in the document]
maxwxyz: I cannot reproduce, what should I change?

src/Gui/MainWindow.cpp
maxwxyz: This function was copied from their previous location, I did no change here.

maxwxyz:
@wwmayer thank you!

src/Gui/MDIView.cpp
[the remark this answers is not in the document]
wwmayer: Make sure that the StartPage isn't shown and that no empty document is created at startup.
Alternatively, close all MDI views.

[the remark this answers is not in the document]
maxwxyz: indeed! Can reproduce now. Thanks for the fix.

Issue #12000, opened 2024-01-15
[Problem] Preferences dialog srcolls to top if the sections are not opened manually


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Opening the preferences dialog and clicking on the text of a section opens all sub-sections.
If I click on a sub-section, the preferences dialog scrolls to the top automatically and I need to scroll down to where I was. Therefore I also do not see which sub-section is currently active, or in which parent section this current preferences page is.

This does not happen, when I manually open the sections with the arrow icon on the left.

Demo (first the issue, at the end the result with manually opening the sections)

https://github.com/FreeCAD/FreeCAD/assets/6246609/d21c1b45-e8d0-4256-949d-539091dafc10

### Full version info

```shell
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35449 (Git)
Build type: Release
Branch: main
Hash: f3fa2ba9c9b1999e5c32d3ae544481b6e17553bb
Python 3.11.7, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.6, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * 3DfindIT 1.2.0
  * BIM 2021.12.0
  * CfdOF 1.25.1
  * CurvedShapes 1.0.5
  * Curves 0.6.21
  * Defeaturing 1.2.1
  * fasteners 0.5.0
  * FEMbyGEN 2.1.0
  * freecad.gears 1.0.0
  * freecad_metal_workbench 0.0.1
  * OpenDark 2023.12.17
  * sheetmetal 0.3.15
  * woodworking 0.21.1.33694
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #12037, opened 2024-01-18
Gui: Fix scroll behaviour in preference pages selector

kadet1090:
This fixes problem with scroll jumping after selecting page on bottom of the pages tree view. It was caused by collapsing group of the current item and then re-expanding it few lines after.

Fixes: #12000


chennes:
@maxwxyz can you confirm the fix?

maxwxyz (approved):
@chennes works

PR #12038, opened 2024-01-19
Arch: Revised preference ui files, remove WindowColor

Roy-043:
Follow-up PR of #11940.

* The new ui layout follows that of the Draft preferences.
* The ConversionFast checkbox enables/disables the 3  preferences listed below it.
* The WindowColor preference was removed. It is not a per-object, or even a per-file setting. The fix in #11940 would change windows in existing projects when they are recomputed. It is therefore better to keep the V0.21 solution where the glass is colored according to the preferences and the rest of the window receives a color based on its ShapeColor. Users who want to apply specific colors can use a multi-material.

### DIFF
--- a/src/Mod/Arch/ArchCommands.py
+++ b/src/Mod/Arch/ArchCommands.py
@@ -84,7 +84,7 @@ def getDefaultColor(objectType):
         c = params.get_param("constructioncolor")
         transparency = 0.80
     else:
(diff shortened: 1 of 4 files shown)


Roy-043:
I am going to take the liberty of merging this myself. The PR is hardly controversial. And I need to have this merged for a next PR.

Issue #11984, opened 2024-01-14
[Problem] Some stylesheets are not coherent in code and look and are outdated.


MisterMakerNL:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Freecad now has 

1. "Dark"
2. "Darker"
3. "Dark modern"
4. "Dark behave"
5. "Dark contrast"
6. "ProDark"
7. "Light"
8. "Light modern"

8 Stylesheets, but 3 of them are the same and only have different colors.
2 of the 8 are slightly different from the 3, these are the modern version. Those are there to try and make a more modern style. But they have the same code only the tabs and buttons style are different.
Basically the 5 are similar.
Then we have the 3 others:
Which are:
1. "Dark behave"
2. "Dark contrast"
3. "ProDark"
They all have different code style and look very different.
They are also not up-to-date the custom color setting and the overlay system have not been checked and there are very minor bugs around.
Example tiny bugs:
These buttons in Prodark
![image](https://github.com/FreeCAD/FreeCAD/assets/29804962/f451d498-2ace-4d5b-8a12-828d9bdcb32b)
No highlight in the pull-down:
![image](https://github.com/FreeCAD/FreeCAD/assets/29804962/3a5db08b-c1a6-42aa-9992-814f572a8a19)
Overlay in Dark behave same with contrast.
![image](https://github.com/FreeCAD/FreeCAD/assets/29804962/e18d40f5-2db6-44cd-9db4-f6d27959e40d)
All the overlay icons aren't good too.
So not really unusable but also not great.

### Full version info

```shell
NA
```

### Subproject(s) affected?

None

### Anything else?

https://github.com/FreeCAD/FreeCAD/pull/11982

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

RohanMishra315:
Hey @Roy-043 @MisterMakerNL I would love to look into this ?

Syres916:
See https://github.com/FreeCAD/FreeCAD/pull/12039

chennes:
[...]

PR #12039, opened 2024-01-19
[Gui] Behave-Dark StyleSheet/Theme Maintenance

Syres916:
Add in the overlay code for the Behave-Dark stylesheet.

Edit: Removed all colour changes.


MisterMakerNL:
The blue style is bold, and I think it is cool. Although risky since the huge amount of colors used true-out the program. Thankfully we at-least are getting rid of some in Sketcher. But still colors every where. 
So if you go for blue please test it good!

Because the looks are one thing, but bugs are more important. 
Issue with included theme's is that you cannot update them separately unlike the one in an add-on. 

Because of this update issue, we should probably work towards less styles included, but the ones included should be polished.

I did not designed the default styles btw I have no attachment too it, so I don't care if those will be replaced. I do have attachment to the bug fixes I did and don't wanna spend hours again on fixing them in another theme!

Syres916:
> The blue style is bold, and I think it is cool. Although risky since the huge amount of colors used true-out the program. Thankfully we at-least are getting rid of some in Sketcher. But still colors every where. So if you go for blue please test it good!

I've actually been using it for months and I even use an Onshape like colour for the 3D objects but I know that'll be too much for most hence not including it. I couldn't believe that people put up with pink/light magenta as their constraints colour in Sketcher so I wound that down back to a red in conjunction.

There are parts of FC that I don't use much including FEM so I'll double check the colour bar text (which is configurable) and Architecture hasn't been used since I finished my stone mullions so I can check that Windows etc look OK.

chennes:
As you anticipated in your PR submission, there isn't any way for Maintainers to evaluate this. It is making two different kinds of changes: those that address technical deficiencies in the current stylesheet, and those that "update the look". For the technical deficiencies, you should create an Issue demonstrating them, which is then closed by this PR. That would be an easy merge. 

For the more stylistic changes, we definitely need an Issue that lays out the problems with the current look, and your proposals for updating the stylesheet. As it stands now there just isn't a way for the Maintainer team to merge this.

Syres916:
@chennes I have to say I really never understand the logic here, https://github.com/FreeCAD/FreeCAD/issues/11984 specifically lays down what is wrong with the Behave-Dark stylesheet and I've gone though fixing the issues, I'll back out the new colours but I took the phrase _'look outdated_' to mean they need to be brought into 2024 which I believe is exactly what I did, much shaking of head here.

chennes:
> _'look outdated_'

It's just too subjective -- that's not a real Issue, it's an opinion. That doesn't mean it's wrong, or that we can't merge a change to the look, but it means there has to be meaningful discussion first. It got lumped in with a bunch of technical problems in the original Issue, but really needs to be split out on its own.

MisterMakerNL:
Yeah, please split out the technical improvements from the visual improvements. So we don't get "This is ugly discussion between the bug improvements/fixes"
I hope this is still possible.

I still think it should be an addon tbh.
It's probably better to just remove all styles and just put them in addons. 
Then we force users to use the addon browser and give every style creator the same chances of being used.
I think I will open a discussion about this on the forum or here on github.

Syres916:
New PR to replace this will be created

PR #12052, opened 2024-01-20
Fem: Call base class handleChangedProperyType function

marioalexis84:
Functions to handle property changes should also call the corresponding function of the base class.


chennes:
Is this addressing a bug, or is this just a deficiency you noticed?

marioalexis84:
It is a deficiency. There is no bug at the moment. If any base class changes the type/name of a property, the change will never be handled correctly unless the appropriate function is called.

Issue #11942, opened 2024-01-09
[Problem] Some of the checkboxes in Tasks pane inside Sketch editing view are not scaled properly


grudir:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

The checkboxes in Constraints and Elements lists in Tasks pane inside Sketch editing view are not displayed properly when using Windows Display scale percentage different from 100%. Additionally the arrows used to expand and collapse the sections inside the Tasks pane as well as the icons in Edit > Preferences are not displayed properly.

Pre-requisites:
- Right click Windows Desktop and choose Display Settings. Set Scale to 125%.

Steps to reproduce:
1. Start FreeCAD 0.21.1.
2. Click File > New.
3. Click Part Design > Create Body.
4. Click Sketch > Create Sketch.
5. Choose XY-plane (and click OK).
6. Draw a rectangle.

Actual result:
The checkboxes in Constraints and Elements lists inside Tasks pane are a bit jagged (probably due to scaling).

Expected result:
The checkboxes in Constraints and Elements lists inside Tasks pane are clear similar to the checkbox on the left of the Filters button.

Example screenshot:
![image](https://github.com/FreeCAD/FreeCAD/assets/24936746/84407b26-f412-40be-9727-88405d58e036)

### Full version info

```shell
[code]
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.21.1.33668 +26 (Git)
Build type: Release
Branch: (HEAD detached at 0.21.1)
Hash: f6708547a9bb3f71a4aaade12109f511a72c207c
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: Bulgarian/Bulgaria (bg_BG)
Installed mods: 
  * Assembly4 0.50.6
  * CurvedShapes 1.0.5
  * fasteners 0.4.75
  * freecad.gears 1.0.0
  * InventorLoader 1.4.0
  * Manipulator 1.5.5
  * Pyramids-and-Polyhedrons
  * QuickMeasure 2022.10.28
  * Render 2023.8.18
  * workfeature
[/code]
```

### Subproject(s) affected?

Other (specify in description)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Roy-043:
Related?: #11766

grudir:
[...]

Issue #11766, opened 2023-12-18
[Problem] Sketcher: UI - Task Dialog 


maxwxyz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

The task dialog in sketcher is displaying constraints and geometry in two lists.
The constraints icons are huge in comparison to the geometry icons and the text seems not aligned.
Also the checkboxes seem different of the two lists. I am using the OpenDark theme but this is a general issue of the dialog.
The selection highlight color in both lists is also different, also in classic theme, but only when selecting in the 3d view.

![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/fb28d2e5-568a-4e0f-9cc0-7f2829e48c90)
![grafik](https://github.com/FreeCAD/FreeCAD/assets/6246609/63b1bd84-ed70-4cc5-a11b-fd0ba145ec37)

### Full version info

```shell
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35407 (Git)
Build type: Release
Branch: sketcher_cursor_icons
Hash: 00aefca95bb2cca0f501e0cb5a92e75854f2a649
Python 3.11.7, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.6, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * 3DfindIT 1.2.0
  * BIM 2021.12.0
  * CfdOF 1.24.9
  * CurvedShapes 1.0.5
  * Curves 0.6.16
  * Defeaturing 1.2.1
  * fasteners 0.4.74
  * FEMbyGEN 2.0.0
  * freecad.gears 1.0.0
  * freecad_metal_workbench 0.0.1
  * OpenDark 2023.12.1
  * sheetmetal 0.3.13
  * woodworking 0.21.1.33694
```

### Subproject(s) affected?

Sketcher

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Roy-043:
Checkboxes look the same here. I suspect that issues is related to the stylesheet.

I think the icons have (almost) the same size (15-16px), but the icons in the Elements list contain more details, and should therefore probably be bigger.

I don't see misalignment of texts.

maxwxyz:
@obelisk79 then it seems to be in the theme?

obelisk79:
Does the icon size change if you change theme?

maxwxyz:
[...]

Issue #8814, opened 2023-03-10
[Problem] Sketcher: Element Widget. Checkboxes not clicking correctly.


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

0.21 (Development)

### Full version info

```shell
OS: Windows 10 Version 2009
Word size of FreeCAD: 64-bit
Version: 0.21.0.32237 (Git)
Build type: Release
Branch: master
Hash: d27e69bdc9cf42489067ec2fbda8e17a17d35929
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk , OCC 7.6.3
Locale: English/United States (en_US)
```

### Subproject(s) affected?

Sketcher

### Problem description

The Element widget just got new checkboxes to toggle visibility. Same as it was implemented in constraint widget.
However the checkboxes don't work properly. Clicking on the checkboxes don't toggle them on the first try. It first select the row. Then you need to click again.
GeoPoints can't be toggled at all.
Sometimes clicking on one checkbox, toggle another checkbox (last selected it seems).

@abdullahtahiriyo I think you implemented those checkboxes last week?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

abdullahtahiriyo:
There is an issue with the selection of the row. I will need to take a look into it.

luzpaz:
@PaddleStroke is this still an issue ?

PaddleStroke:
Just tested and yes

abdullahtahiriyo:
The current implementation is not good. I would like to address this as part of the changes to the widgets using a proper model-view architecture. Hopefully during this development cycle.

PR #12054, opened 2024-01-20
Sketcher: Apply styling to elements and constraints UI

kadet1090:
This ensures that checkboxes drawn in the Elements UI of the sketcher can be styled using the correct style.

Three changes are introduced:
1. paint() method uses style provided by the widget, not application.
2. Indicator is drawn using `PE_IndicatorCheckBox` not `CE_CheckBox` - it allows to use `::indicator` in qss to style it properly. This works the same as TreeView.
3. Fix for minor issues with text alignment and extension of selected items background to cover checkbox. This is how other controls of this kind work.

It also completely refactors the rendering delegate introducing dedicated method for calculating rects of each element and method for rendering each sub control. Thanks to this refactor we can be sure that areas drawn are the same as areas checked for events in `editorEvent`.

I also changed logic on highlighting selected elements - now elements that are not selected are semi-transparent (40% - 80% when preselected). Current solution would not work as from paint method I don't have access to the stylesheet properties. This is certainly up for debate.

Fixes:  #11942 - @grudir - if you could test if it works now? I was not able to reproduce this issue, but it should work now.
Fixes: #11766
Fixes: #8814

After (this is only proposal of how it could look - it is controlled via stylsheet and now random stuff got applied)
image

Before:
image

### DIFF
--- a/src/Gui/Stylesheets/images_dark-light/check_dark_disabled.svg
+++ b/src/Gui/Stylesheets/images_dark-light/check_dark_disabled.svg
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="30"
+   height="30"
+   id="svg2"
+   version="1.1"
+   viewBox="0 0 30 29.999999"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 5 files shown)


kadet1090:
@PaddleStroke if you could review that it would be great

grudir:
@kadet1090 thank you for providing a fix for the issue in such a short time! Currently I don't have a running FreeCAD build environment. Is there a way I can get the compiled binaries for Windows to test?

PaddleStroke:
@kadet1090 is it ready to review ? Or if not, did you want me to look at some specific thing right now?

kadet1090:
It is ready to review, take a look at everything please. I don't plan to make any more chances except for fixing issues if any is found

PaddleStroke (approved):
Code looks good to me.
And I build and it works well. Now in the element widget the unselected elements are transparent and clicking them set them 100%visible. It's a nice way of doing it and it looks good.

PR #12059, opened 2024-01-21
Pyside6 fixes

3x380V:
Fixes building against [pip installed PySide6](https://forum.freecad.org/viewtopic.php?p=734253). There's a bit more than that, feel free to cherry-pick.


src/Gui/PythonWrapper.cpp
```diff
@@ -362,12 +369,8 @@ Py::Object qt_wrapInstance(qttype object,
...
 {
     PyObject* module = PyImport_ImportModule(shiboken.c_str());
-    if (!module) {
-        std::string error = "Cannot load ";
-        error += shiboken;
-        error += " module";
-        throw Py::Exception(PyExc_ImportError, error);
-    }
+    if (!module)
+        throw Py::Exception(PyExc_ImportError, formatModuleError(shiboken));
```
chennes, line 373: Could you add curly braces here (and below)? This code isn't yet covered by our auto-formatting CI, but it will be eventually, and this will be one less thing we have to fix.

chennes (changes requested):
I think this is fine, my plan is to merge the whole thing, pending one small formatting fix.

src/Gui/PythonWrapper.cpp
3x380V: Done, force-pushed (this is something I really do not like, so I need discipline to verify changes against coding policy :))

chennes: If your IDE supports clang-format and is capable of applying it locally (e.g. not to the entire file, but only to selected portions), that is probably your easiest option, while we slowly migrate our codebase to using it.

3x380V: [Vim](https://www.vim.org/) is my IDE, I'll look into clang-format later. Thank you for suggestion.

chennes: Perfect, the clang-format plugin for vim works on isolated text selections, which is exactly what's needed.

3x380V: [Vim](https://www.vim.org/) is my IDE. I'll look into clang-format later, thank for suggestion. (Now it seems main branch was merged, while rebase was intended, confusing... Should I fix it and force push or are you able to resolve it yourself?)

chennes: Don't worry about it, I'll clean it up when I merge.

chennes (approved):

PR #12061, opened 2024-01-21
Base cleanup

3x380V:
Started from [Suggestion: no "else" after "if-return"](https://forum.freecad.org/viewtopic.php?t=82410) forum thread...


Syres916:
@3x380V please see regression info https://forum.freecad.org/viewtopic.php?t=84758 , I had a quick look in the current PRs and can't see a submitted fix for this but maybe wrong.

PR #12064, opened 2024-01-21
Gui: add explicit transparency specification

bdieterm:
As described in the [forum](https://forum.freecad.org/viewtopic.php?p=734688#p734688), the transparency value of an App::Plane object is set to the default transparency instead of zero after creation.

During creation, the App::Plane object is assigned the specified default shape color and then the color is set (again) to a fixed color for the plane (light-blue). But the transparency was not set to a fixed value and therefore kept the default transparency value.

This PR also adds the missing transparency specification to some other objects.

To test the changes, the following Python code can be executed:
```python3
doc = App.newDocument()

pg = App.ParamGet('User parameter:BaseApp/Preferences/View')
backup_default_transparency = pg.GetInt('DefaultShapeTransparency')

pg.SetInt('DefaultShapeTransparency', 70)

obj = doc.addObject('App::Origin')
t = doc.findObjects('App::Plane')[0].ViewObject.Transparency
if t != 0:
    print('WARNING: transparency of App::Plane object is {} instead of 0'.format(t))
else:
    print('transparency of App::Plane object is ok')

pg.SetInt('DefaultShapeTransparency', backup_default_transparency)
```

### DIFF
--- a/src/Gui/ViewProviderOriginFeature.cpp
+++ b/src/Gui/ViewProviderOriginFeature.cpp
@@ -51,6 +51,7 @@ ViewProviderOriginFeature::ViewProviderOriginFeature () {
     QT_TRANSLATE_NOOP("App::Property", "Visual size of the feature"));
 
     ShapeColor.setValue ( ViewProviderOrigin::defaultColor ); // Set default color for origin (light-blue)
+    Transparency.setValue(0);
     BoundingBox.setStatus(App::Property::Hidden, true); // Hide Boundingbox from the user due to it doesn't make sense
 
     // Create node for scaling the origin
--- a/src/Mod/Fem/Gui/ViewProviderFemMesh.cpp
+++ b/src/Mod/Fem/Gui/ViewProviderFemMesh.cpp
@@ -197,6 +197,7 @@ ViewProviderFemMesh::ViewProviderFemMesh()
     LineWidth.setConstraints(&floatRange);
 
     ShapeColor.setValue(App::Color(1.0f, 0.7f, 0.0f));
+    Transparency.setValue(0);
     ADD_PROPERTY(BackfaceCulling, (true));
     ADD_PROPERTY(ShowInner, (false));
     ADD_PROPERTY(MaxFacesShowInner, (50000));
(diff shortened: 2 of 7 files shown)


chennes:
Can you add your test code to our automated test suite? It would be nice to make sure this doesn't regress in the future.

bdieterm:
> Can you add your test code to our automated test suite? It would be nice to make sure this doesn't regress in the future.

done

adrianinsaval:
Doesn't this still let users manually put a transparency level which then would trigger the same issue?

bdieterm:
@adrianinsaval
Maybe you are referring to the behaviour that, when setting the View->Transparency value of the XY-plane object manually to a high value, the frame of the plane looks opaque (as in FreeCAD v0.21.0). If you mean that, and if you consider this a bug, please create an issues for that. This PR only addresses the problem that the global default shape transparency was applied to plane objects.

adrianinsaval:
> @adrianinsaval Maybe you are referring to the behaviour that, when setting the View->Transparency value of the XY-plane object manually to a high value, the frame of the plane looks opaque (as in FreeCAD v0.21.0). If you mean that, and if you consider this a bug, please create an issues for that. This PR only addresses the problem that the global default shape transparency was applied to plane objects.

No I mean that origin objects that where created with a transparency will remain with this issue and if you manually add transparency to the origin objects they will still present this issue. This only stops newly created objects from having this behavior. curiously if only one of the axis objects doesn't have transparency the issue stops. Activating the axis cross also stops the issue unless you zoom in enough to make the cross be out of sight.

bdieterm:
@adrianinsaval
As far as I understand, you have described the same problem as I did in your citation.
And this problem is something old that was already present in the stable release FreeCAD v0.21.
What I fixed with this PR, is not that problem. This PR is only about the transparency that Plane objects are created with (no matter if someone changes the value later on and what happens as consequence of that).

Just try it yourself. If you can reproduce the problem you described in FreeCAD v0.21 (as I can), then it is definitely a different issue/problem and therefore not in the scope of this PR.

By the way: The Plane object of the origin consists actually of (at least) two shapes: [the inner area that is semi-transparent](https://github.com/FreeCAD/FreeCAD/blob/c12a5c2a163c1a277cf273106899209792be152d/src/Gui/ViewProviderPlane.cpp#L83-L90) and [the outer border](https://github.com/FreeCAD/FreeCAD/blob/c12a5c2a163c1a277cf273106899209792be152d/src/Gui/ViewProviderOriginFeature.cpp#L53-L55).
What would be the correct transparency values for each shape if you modify `doc.getObject('XY_Plane').Transparency` in your opinion? Don't answer this here, but in an GitHub issue or a forum post instead.

PR #12066, opened 2024-01-21
Base: Use std::chrono for time manipulation

3x380V:
Base still contains quite some legacy code which could be better expressed in C++ now. Let's start with TimeInfo and reimplement it using std::chrono. Class interface remains mostly the same.


src/Base/TimeElapsed.h
```diff
@@ -0,0 +1,100 @@
+/***************************************************************************
+ *   Copyright (c) 2011 Jürgen Riegel <juergen.riegel@web.de>              *
```
chennes, line 2: Don't forget to add your copyright here.

3x380V: I always feel a bit ashamed to claim copyright on trivial refactors, but well... 
I would appreciate overall comment whenever those changes are making sense to you. I do not want to spend too much time going wrong direction :)
There's also FileInfo, which will be refactored next using std::filesystem.

chennes: I love to see these changes -- FreeCAD's codebase is OLD, so much of it predates things like `recursive_mutex` and `filesystem`, it's great to get updated. As you are doing it just remember that we continue to support compilation on the oldest fully-supported Ubuntu LTS as our target "old" OS. That's what's holding us to C++17 right now, for example.

RE: Copyright... you be the judge, it's a balance between taking credit, and taking blame (e.g. don't assert that Jürgen wrote by himself a chunk of code that you've modified heavily. Of course git will record the truth, but I look at copyright headers a lot when deciding who to ask about a chunk of code).

3x380V: I'm always using git blame, but here it often fails due to lacking history. And yes, I know there is a tree with complete history imported.

C++17 is enough to get most of platform specific code gone. Just do not hold your breath, it is just a hobby project and I cannot spend much time on it. But the work will get eventually done - too much of my construction work now depends on FreeCAD, so I wish it will stay with me forever. That requires polishing here and there.

I'll probably convert this patchset to more general Base 'modernisation' branch and add more changes here, just to see how it all will work together.

Thank you for your time reviewing it.

chennes: Thanks for the work you are doing. FYI you don't have to wait for it to be "perfect" or "finished" -- I'm more than happy to merge incremental improvements, it's much nicer than trying to review a PR that fixes everything all at once.

3x380V: I take care to make workflow bisectable, so it always works and compiles. That way it can be splitted into as many PR as feasible. Of course I'll do my best to make reviewing easier... (here I interrupt work as Addon Manager does not work with Qt6 - quite some deprecated and deleted functions still used. PR will be created).

src/Base/FileInfo.cpp
```diff
@@ -456,52 +456,44 @@ unsigned int FileInfo::size() const
...
     }
-    return ti;
+    return res;
 }
 
-TimeInfo FileInfo::lastRead() const
+time_t FileInfo::lastRead() const
 {
-    TimeInfo ti = TimeInfo::null();
+    time_t res = (time_t)(-1);
```
chennes, line 481: Is this idiomatic? `time_t`'s type isn't defined by the standard.

3x380V: I just needed something to indicate error. As time(2) man page says '_On error, ((time_t) -1) is returned, and errno is set to indicate the error._' I will probably implement `TimeInfo` class again based on `std::chrono` once `FileInfo` gets reimplemented using `std::filesystem`.

chennes:
Yeah, if we can keep it to just a few ms that is fine.

chennes:
Do you want to switch this PR to draft mode, then?

3x380V:
Now `TimeInfo` class is back and new `TimeElapsed` class is used to measure time intervals as explained in respective commits. All this could be further improved by creating only thin wrapped on top of `std::chrono` template, so they would end as header only classes.
I'm leaving this for second round of time optimizations because except renames, code follows original logic. Lets advance in smaller chunks - unfortunately I do not have a time for bigger steps forward...

3x380V:
Header only classes. Still needs some work, but it is good enough to get idea about implementation. Comments welcome. If this is way to go, I will clean and fix it up and remove draft status.

3x380V:
> Instead of removing the tests, could you modify them to cover `TimeElapsed`? It should basically be the same tests.

Tests are back. Now they are basically testing std::chrono.

A short remark on using `high_resolution_clock`: the original intention was to provide something useful for profiling, the most precise clock platform is able to provide. And it went wrong. Although on all supported platforms this clock is steady and we might even add static assert to ensure that, there are indices to just not use it: [ISO C++ Standard - Discussion](https://groups.google.com/a/isocpp.org/g/std-discussion/c/D_S47zTnNmw)
I will leave decision for later.

3x380V:
So `steady_clock` is used at last mostly because libstd++ started aliasing `high_resolution_clock` to `system_clock` and that breaks steady rule.
`currentDateTimeString` could be implemented either using fmtlib or better C++20 once FreeCAD starts using it. Leaving that for later.

PR #12070, opened 2024-01-22
Material: Initialize types

marioalexis84:
`Materials` types that inherit from `Base::BaseClass` were not initialized properly.
Not initializing types leads to a lot of crashes and undefined behavior (especially from Python's Material module, which until now was basically unusable) and results in material cards being inaccessible from some workbenches like FEM.


chennes:
Thanks for the PR, @marioalexis84 -- I'll merge it once @davesrocketshop sees it.

davesrocketshop (approved):

PR #12077, opened 2024-01-22
[PD] Implement symmetric linear/polar patterns

### LINKED ISSUE #10442: [Feature Request] Symmetric option for Linear and Polar patterns

### Full version info

```shell
OS: Arch Linux (awesome)
Word size of FreeCAD: 64-bit
Version: 0.21.0.33675 (Git)
Build type: Release
Branch: makepkg
Python 3.11.3, Qt 5.15.10, Coin 4.0.1, Vtk 9.2.6, OCC 7.6.3
Installed mods: 
  * Assembly3 0.12.0
  * IconThemes
  * Glass (Disabled)
  * lattice2 1.0.0
  * A2plus 0.4.60n
```

### Subproject(s) affected?

PartDesign

### Problem description

For some patterns it'd be nice to be able make them symmetrical to the original feature(s). I think that the best option would be to implement it as additional boolean property / checkbox for the patterns - this way we could achieve similar UX between two possible pattern kinds. 

#### How it should behave?
There are few options how such "Symmetric" option could work, for example, when we have:
```
- Angle set to 90
- Mode set to Overall Angle
- Occurrences set to 2
- Symmetric option on
```

Should we have:
a) 2 features, 90 degrees apart? We would have to "move" original feature.
b) 3 features spanning across 90 degrees?
c) 3 features spanning across 180 degrees - 2 per symmetry side, each side spans on 90 degrees? Then we'd have to constraint maximum angle to 180, not mentioning that angle would no longer be overall.

image

Personally I think that a) seems to be most useful, c) also could be an option but it could be easily done now using multi transform with pattern + mirror. I however think that having that option as simple checkbox would be great for product usability. 

If we choose option a) for even number of occurrences, I see few potential problems:

#### Relation to Topological Naming
As there will no longer be feature in original place, instead pattern would be distributed evenly across length or angle. Will it have impact on topological naming, as it'd be hard to name any created feature as original?

#### Non-Patterned feature
We can create pattern only for some features - how about case where we have some dependency on the patterned feature which will not be a part of pattern? Should we even care, or consider it as error from the users side which should be corrected (feature should be created after pattern).

image

For this example the pocketed circle will move so peg will no longer be placed correctly.

### Anything else?

Forum discussion: https://forum.freecad.org/viewtopic.php?t=68621&start=13

### Code of Conduct

LemonBoy:
Add a boolean option that's only active when the user specifies the pattern in terms of offset and occurrences. The pattern is symmetrically replicated around the main datum, creating 2n copies in total.

Closes #10442

### DIFF
--- a/src/Mod/PartDesign/Gui/TaskPolarPatternParameters.cpp
+++ b/src/Mod/PartDesign/Gui/TaskPolarPatternParameters.cpp
@@ -135,6 +135,8 @@ void TaskPolarPatternParameters::connectSignals()
             this, &TaskPolarPatternParameters::onModeChanged);
     connect(ui->checkReverse, &QCheckBox::toggled,
             this, &TaskPolarPatternParameters::onCheckReverse);
+    connect(ui->checkSymmetric, &QCheckBox::toggled,
+            this, &TaskPolarPatternParameters::onCheckSymmetric);
     connect(ui->polarAngle, qOverload<double>(&Gui::QuantitySpinBox::valueChanged),
             this, &TaskPolarPatternParameters::onAngle);
     connect(ui->angleOffset, qOverload<double>(&Gui::QuantitySpinBox::valueChanged),
@@ -172,6 +174,7 @@ void TaskPolarPatternParameters::setupUI()
     ui->comboAxis->setEnabled(true);
     ui->comboMode->setEnabled(true);
     ui->checkReverse->setEnabled(true);
+    ui->checkSymmetric->setEnabled(false);
     ui->polarAngle->setEnabled(true);
     ui->spinOccurrences->setEnabled(true);
 
@@ -199,7 +202,6 @@ void TaskPolarPatternParameters::setupUI()
     }
 
     adaptVisibilityToMode();
-    updateUI();
     connectSignals();
 }
 
@@ -213,6 +215,7 @@ void TaskPolarPatternParameters::updateUI()
 
     PartDesign::PolarPatternMode mode = static_cast<PartDesign::PolarPatternMode>(pcPolarPattern->Mode.getValue());
     bool reverse = pcPolarPattern->Reversed.getValue();
+    bool symmetric = pcPolarPattern->Symmetric.getValue();
     double angle = pcPolarPattern->Angle.getValue();
     double offset = pcPolarPattern->Offset.getValue();
     unsigned occurrences = pcPolarPattern->Occurrences.getValue();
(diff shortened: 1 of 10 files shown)


chennes:
@kadet1090 since that Issue is one you created, could you take a look at this PR?

src/Mod/PartDesign/Gui/TaskPolarPatternParameters.cpp
```diff
@@ -223,9 +226,14 @@ void TaskPolarPatternParameters::updateUI()
         axesLinks.setCurrentLink(pcPolarPattern->Axis);
     }
 
+    Base::Console().Log("mode is %d\n", (int)mode);
```
kadet1090, line 229: That probably should be removed.

kadet1090:
Yes, it works as intended, assuming of course that we want the "c" option from the issue - for me it is fine and in general I feel that there should not be much opposition to that. In that issue we also concluded, that naming it "mirrored" may be a better choice for this particular solution - what do you guys think?

I'd also change the UI here a bit. 

image

The checkbox is enabled only for "Offset Angle" mode - there is no reason to show it for "Overall" mode if it is not supported. It should be probable moved below the "Occurrences" and shown only when user can actually use it. I'm not even sure if the checkbox should not be simply enabled for the "Overall" modes?

@obelisk79 maybe you have some thoughts

LemonBoy:
> Yes, it works as intended, assuming of course that we want the "c" option from the issue - for me it is fine and in general I feel that there should not be much opposition to that.

This behaviour felt the most reasonable one to me and, for what it's worth, it's also the behaviour implemented in the latest Solidworks' versions.

> In that issue we also concluded, that naming it "mirrored" may be a better choice for this particular solution - what do you guys think?

No strong feelings re: the naming, let's say I slightly prefer "symmetric" over "mirrored".

> The checkbox is enabled only for "Offset Angle" mode - there is no reason to show it for "Overall" mode if it is not supported.

Hiding the option hinders its discoverability IMO.

> I'm not even sure if the checkbox should not be simply enabled for the "Overall" modes?

The option is disabled because I couldn't decide if the overall length was supposed to be the one-sided or two-sided one, suggestions are welcome.

kadet1090:
> Hiding the option hinders its discoverability IMO.

Why? You cannot select it anyway, and it does not contain any tooltip describing why it is disabled. If you switch to the "Offset" which will enable this checkbox it will pop-up and user should notice it without problem.


> The option is disabled because I couldn't decide if the overall length was supposed to be the one-sided or two-sided one, suggestions are welcome.

IMHO if we name it "Mirrored" then it should be one sided as it should be obvious that you mirror the described pattern. If we stay with symmetric it may be better to do it two-sided.

adrianinsaval:
I'm not convinced that we need or want this feature, we have multi transforms where you can make a mirror out of a pattern. This is duplicating functionality in a different place. We are introducing more inconsistency IMO.

If we do accept the feature, then I think it should be called Symmetric and overall length should be two sided, for consistency with other features such as Pad, Pocket and Revolution

obelisk79:
So, I'm not quite back on my feet after being diagnosed with COVID. I'll build this PR and test it out from a UX perspective soon I hope.

kadet1090:
> If we do accept the feature, then I think it should be called Symmetric and overall length should be two sided, for consistency with other features such as Pad, Pocket and Revolution

Then we are doubling the occurrences and not doubling the lenght which also is not very consistent. I'm fine with either and totally agree that if we are not doubling the length it should be called symmetric.

PR #12080, opened 2024-01-23
AddonManager: Disable README reformatting

chennes:
GitHub has changed their page format: to cope with this quickly, this PR disables the custom format parsing that the Addon Manager was doing, and simply displays the data as sent.


chennes:
Not the problem after all...

PR #12085, opened 2024-01-23
Arch: importDAE: DaeUnsupportedError is defined in common.py

Roy-043:
Forum topic:
https://forum.freecad.org/viewtopic.php?t=84626

### DIFF
--- a/src/Mod/Arch/importDAE.py
+++ b/src/Mod/Arch/importDAE.py
@@ -119,7 +119,7 @@ def read(filename):
     "reads a DAE file"
 
     global col
-    col = collada.Collada(filename, ignore=[collada.DaeUnsupportedError])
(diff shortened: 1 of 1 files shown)


Roy-043:
Looking at the docs this PR should not cause a problem for collada V0.4:
https://pycollada.readthedocs.io/en/v0.4/reference/generated/collada.common.html#module-collada.common

Testing on FreeCAD V0.17 confirms this:
```
import collada
collada.__version__  # => '0.4.1'
hasattr(collada.common, "DaeUnsupportedError")  # => True
```

IMO this can be merged. Unless you think we should also consider collada V0.3?

Syres916:
Ok, go ahead, we'll no doubt find out for certain on release!!

PR #12086, opened 2024-01-23
Toponaming: makeElementBoolean

bgbsww:


src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -34,34 +34,70 @@
...
 #include "TopoShape.h"
+#include "TopoShapeOpCode.h"
 #include "TopoShapeCache.h"
 #include "TopoShapeMapper.h"
 #include "FaceMaker.h"
 
 #include "TopoShapeOpCode.h"
 #include <App/ElementNamingUtils.h>
 #include <BRepLib.hxx>
+#include <OSD_Parallel.hxx>
```
chennes, line 80: This is wrapped in `#if OCC_VERSION_HEX >= 0x070500` above. Also, not in the PCH file?

src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -2362,4 +2398,230 @@ bool TopoShape::fixSolidOrientation()
...
+TopoShape& TopoShape::makeElementBoolean(const char* maker,
+                                         const std::vector<TopoShape>& shapes,
+                                         const char* op,
+                                         double tolerance)
+{
+#if OCC_VERSION_HEX <= 0x060800
+    if (tolerance > 0.0) {
+        Standard_Failure::Raise("Fuzzy Booleans are not supported in this version of OCCT");
+    }
+#endif
```
chennes, line 2417: We don't support OCCT this old anymore, so this can be removed.

src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -2362,4 +2398,230 @@ bool TopoShape::fixSolidOrientation()
...
+            }
+        }
+        setShape(Comp);
+        mapSubElement(shapes, op);
+        return *this;
+    }
+
+    if (strcmp(maker, Part::OpCodes::Pipe) == 0) {
+        if (shapes.size() != 2) {
+            FC_THROWM(Base::CADKernelError, "Not enough input shapes");
```
chennes, line 2462: This error seems off -- what if there are three?

src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -2362,4 +2398,230 @@ bool TopoShape::fixSolidOrientation()
...
+        }
+        if (shapes[0].isNull() || shapes[1].isNull()) {
```
chennes, line 2464: Another misleading error message: [0] is the spine, [1] is the shape.

src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -2362,4 +2398,230 @@ bool TopoShape::fixSolidOrientation()
...
+        for (auto& s : shapes) {
+            builder.Add(shell, s.getShape());
+        }
+        setShape(shell);
+        mapSubElement(shapes, op);
+        BRepCheck_Analyzer check(shell);
+        if (!check.IsValid()) {
+            ShapeUpgrade_ShellSewing sewShell;
+            setShape(sewShell.ApplySewing(shell), false);
+            // TODO confirm the above won't change OCCT topological naming
```
chennes, line 2487:
```suggestion
            // TODO: confirm the above won't change OCCT topological naming
```
(so it gets picked up by automated tools)

src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -2362,4 +2398,230 @@ bool TopoShape::fixSolidOrientation()
...
 }
 
+TopoShape&
+TopoShape::makeElementBoolean(const char* maker, const TopoShape& shape, const char* op, double tolerance)
+{
+    return makeElementBoolean(maker, std::vector<TopoShape>(1, shape), op, tolerance);
+}
+
+
+TopoShape& TopoShape::makeElementBoolean(const char* maker,
```
chennes, line 2408: Future work: probably every one of these opcode should be its own method, this one is way too long.

src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -2362,4 +2398,230 @@ bool TopoShape::fixSolidOrientation()
...
+    // else if (PartParams::getParallelRunThreshold() > 0) {
+    //     int total = 0;
+    //     for (const auto& shape : inputs) {
+    //         total += shape.countSubShapes(TopAbs_FACE);
+    //         if (total > PartParams::getParallelRunThreshold()) {
+    //             mk->SetRunParallel(true);
+    //             break;
+    //         }
+    //     }
+    // }
```
chennes, line 2610: Why leave the commented-out block?

tests/src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -53,14 +54,14 @@ TEST_F(TopoShapeExpansionTest, makeElementCompoundOneShapeReturnsShape)
...
 
     // Act
-    topoShape.makeElementCompound(shapes,
-                                  "C",
-                                  Part::TopoShape::SingleShapeCompoundCreationPolicy::
-                                      returnShape /*Don't force the creation*/);
+    topoShape.makeElementCompound(
+        shapes,
+        "C",
+        TopoShape::SingleShapeCompoundCreationPolicy::returnShape /*Don't force the creation*/);
```
chennes, line 64: The point of the policy is to remove the need for the comment, so I suggest dropping it (here and elsewhere).

chennes (changes requested):
Review comments below.

src/Mod/Part/App/TopoShapeExpansion.cpp
bgbsww: Yeah, superfluous.  Not the only one.  Wow, the IDE difference here is notable.

src/Mod/Part/App/TopoShapeExpansion.cpp
bgbsww: Hmm.  original source is assuming only ever two shapes for a pipe.  Which is consistent with the declaration of BRepOffsetAPI_MakePipe and the call on 2464.  Seems correct for this code, which can't work with other than 2.

src/Mod/Part/App/TopoShapeExpansion.cpp
bgbsww: At the time it was to call attention to code not being transferred in.  It's gone now.  Thanks!

src/Mod/Part/App/TopoShapeExpansion.cpp
chennes: Then it's just the error message itself that's potentially misleading: if I selected three shapes, the error shouldn't say "Not enough input shapes" 😄.

chennes (approved):

PR #12087, opened 2024-01-23
Toponaming makeElementShell

bgbsww:


src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -551,6 +554,142 @@ TopoShape::makeElementCompound(const std::vector<TopoShape>& shapes, const char*
     return *this;
 }
 
+struct MapperFill: Part::TopoShape::Mapper
+{
+    BRepFill_Generator& maker;
+    MapperFill(BRepFill_Generator& maker)
```
chennes, line 560: Can this be `explicit`?

src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -551,6 +554,142 @@ TopoShape::makeElementCompound(const std::vector<TopoShape>& shapes, const char*
...
+        : maker(maker)
+    {}
+    virtual const std::vector<TopoDS_Shape>& generated(const TopoDS_Shape& s) const override
```
chennes, line 563:
```suggestion
    const std::vector<TopoDS_Shape>& generated(const TopoDS_Shape& s) const override
```

src/Mod/Part/App/TopoShapeExpansion.cpp
```diff
@@ -551,6 +554,142 @@ TopoShape::makeElementCompound(const std::vector<TopoShape>& shapes, const char*
...
+            if (FC_LOG_INSTANCE.isEnabled(FC_LOGLEVEL_LOG)) {
+                FC_WARN("Exception on shape mapper: " << e.GetMessageString());
+            }
+        }
+        return _res;
+    }
+};
+
+// topo naming counterpart of TopoShape::makeShell()
+TopoShape& TopoShape::makeElementShell(bool silent, const char* op)
```
chennes, line 582: I'd rather see this boolean as a `class enum` -- are there other methods in the new code that use this idea of "silent"?

bgbsww: Yes, and there is a fair history of them across the code base.  Try PartDesign::ProfileBased as one example.  I'm not sure it makes sense to change this in just one module.  Could it be a separate, code base wide change someone makes?

PR #12088, opened 2024-01-23
Gui: Consolidate PythonWrapper

3x380V:
This is just a draft as I'm unable to verify all configurations.
I'm still unhappy with PythonWrapper, there are way too many ifdeffed paths, making it error prone. An attempt to clean it up a bit is made. Disclaimer: a week ago I had zero knowledge about Shiboken and it does improve very slowly over time.

Issues: 
- build system is unable to find pip installed QtWebEngineWidgets
- python claims value returned by toSecsSinceEpoch (into repo.updated_timestamp) is a float, therefore typecast.
- only Python loaded modules tested, not linked against shiboken.

Comments very welcome and appreciated.


chennes:
I'm about to tear out QtWebEngineWidgets, so don't worry too much about that one 😄.

3x380V:
@chennes, that was really fast, so I'm not worried any more. I would appreciate is someone could look at Addon Manager part. I works now, it is a bit ungly.

chennes:
These AddonManager changes should be fine, all of those methods and constants existed in Qt 5.12.

3x380V:
The problem with Addon Manager is that I do not know where that float comes from. It ends this way:
```
TypeError: 'float' object cannot be interpreted as an integer
/home/ladis/src/FreeCAD-build/Mod/AddonManager/package_details.py:137: RuntimeWarning: libshiboken: Overflow: Value 1695849953.710445 exceeds limits of type  [signed] "x" (8bytes).
  QtCore.QDateTime.fromSecsSinceEpoch(repo.updated_timestamp)
  File "/home/ladis/src/FreeCAD-build/Mod/AddonManager/AddonManager.py", line 748, in table_row_activated
    self.packageDetails.show_repo(selected_repo)
  File "/home/ladis/src/FreeCAD-build/Mod/AddonManager/package_details.py", line 124, in show_repo
    self.display_repo_status(self.repo.update_status)
  File "/home/ladis/src/FreeCAD-build/Mod/AddonManager/package_details.py", line 137, in display_repo_status
    QtCore.QDateTime.fromSecsSinceEpoch(repo.updated_timestamp)
OverflowError
TypeError: 'float' object cannot be interpreted as an integer
  File "/home/ladis/src/FreeCAD-build/Mod/AddonManager/package_details.py", line 137, in display_repo_status
    QtCore.QDateTime.fromSecsSinceEpoch(repo.updated_timestamp)
OverflowError
```
So that typecasing to int is needed, alhough documentation says it should be u64. And I do not like pushing things I do not understand.
I will create separate PR for Addon Manager once you merge #12097 

@wwmayer, could you elaborate why is QPrinter special? (see XXX in the code)

3x380V:
Ah, `repo.updated_timestamp` is updated also with `os.path.getmtime()` which _returns a floating-point value of class ‘float’ that represents the time (in seconds) of last modification of the specified path._

3x380V:
Addon Manager has its PR  #12100 now. However I do not know how to remove label.

chennes:
This is looking good now -- is it ready for review?

3x380V:
Yes, however only shiboken/PySide access via C++ is working while via Python it still does not. This is the same as before. Of course, I would appreciate help with debugging. Just uncomment block mentioned in code, to see the problem
https://github.com/FreeCAD/FreeCAD/blob/02c6a90c160a7feac269d8aa0fe1b490bfa5c898/src/Gui/PythonWrapper.cpp#L37

3x380V:
So here it is:
`PyObject* CommandPy::getAction(PyObject *args)` is calling `wrap.fromQObject(a)` (there are some places where result of `wrap.loadWidgetsModule()` is not tested at all, btw) and because `const char* className` is set to nullptr, later in `Py::Object PythonWrapper::fromQObject(QObject* object, const char* className)` construction of `std::string` from nullptr fails.

I bet someone more familiar with codebase would find it way sooner (or perhaps I should try setting up a decent debugger)

src/Gui/PythonWrapper.cpp
```diff
@@ -156,6 +157,38 @@ PyTypeObject** SbkPySide6_QtUiToolsTypes = nullptr;
...
+PyTypeObject** SbkPySide2_QtCoreTypes           = nullptr;
+PyTypeObject** SbkPySide2_QtGuiTypes            = nullptr;
+PyTypeObject** SbkPySide2_QtWidgetsTypes        = nullptr;
+PyTypeObject** SbkPySide2_QtPrintSupportTypes   = nullptr;
+PyTypeObject** SbkPySide2_QtUiToolsTypes        = nullptr;
+#define SbkPySide_QtCoreTypes           SbkPySide2_QtCoreTypes
+#define SbkPySide_QtGuiTypes            SbkPySide2_QtGuiTypes
+#define SbkPySide_QtWidgetsTypes        SbkPySide2_QtWidgetsTypes
+#define SbkPySide_QtPrintSupportTypes   SbkPySide2_QtPrintSupportTypes
+#define SbkPySide_QtUiToolsTypes        SbkPySide2_QtUiToolsTypes
```
chennes, line 171: Can you explain this to me? Why do we need to use the preprocessor to do this?

3x380V: Later in the code, we are using the very same load function to load module. `SbkPySide2_*` and `SbkPySide6_*`  need to be globally visible symbols, otherwise build ends with undefined reference. Here we just define the same name despite fact we are using PySide2, PySide6 or Python access. In the last case dummy variable is used. All this to use as much common code as possible.

chennes: Sorry my question was not clear: I meant, why specifically the preprocessor, rather than just using variables, e.g.
```

auto &SbkPySide_QtCoreTypes = SbkPySide2_QtCoreTypes;
auto &SbkPySide_QtGuiTypes = SbkPySide2_QtGuiTypes;
auto &SbkPySide_QtWidgetsTypes = SbkPySide2_QtWidgetsTypes;
auto &SbkPySide_QtPrintSupportTypes = SbkPySide2_QtPrintSupportTypes;
auto &SbkPySide_QtUiToolsTypes = SbkPySide2_QtUiToolsTypes;
```

3x380V: I see. It is just an old habit of saving as much bytes (here 8 per variable) as possible cultivated by many years of embedded development. Not sure if that counts here though...

chennes: Perfect.

3x380V: Replaced two more `#define` with `constexpr`...

PR #12095, opened 2024-01-23
Gui: add missing setup code to FreeCADGui started via Gui.showMainWindow()

bdieterm:
Since the added support for transparent overlay docking widget, FreeCAD reports errors such as
```
Cannot find icon: qss:overlay/close.svg
Cannot find icon: qss:overlay/overlay.svg
Cannot find icon: qss:overlay/mode.svg
...
```
during start-up via `Gui.showMainWindow()` when having FreeCAD imported as a Python module.

To fix that, I copied the necessary code from src/Gui/Application.cpp and adapted it accordingly. I also copied other code parts to make other features such as StyleSheets work in the FreeCAD Gui when it was started from an imported module.

I am aware that duplicated code should be avoided, but my C++ skills and my knowledge of the FreeCAD code base are too limited to come up with a cleaner solution. Any hints from experienced FreeCAD developers are welcome.

### DIFF
--- a/src/Gui/Application.cpp
+++ b/src/Gui/Application.cpp
@@ -138,7 +138,356 @@ namespace sp = std::placeholders;
 
 Application* Application::Instance = nullptr;
 
+namespace {
+    void setImportImageFormats();
+}
+
+#ifdef FC_DEBUG // redirect Coin messages to FreeCAD
+void messageHandlerCoin(const SoError * error, void * /*userdata*/);
+#endif
+
+
 namespace Gui {
+
+
+
+void initGuiAppPreMainWindow(bool calledByGuiPy)
+{
+        QString plugin;
+        plugin = QString::fromStdString(App::Application::getHomePath());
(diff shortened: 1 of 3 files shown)


chennes:
It looks like there is an Application.h include -- if that's where the code is copied from, can you make a function in there that gets called by both places?

bdieterm:
I was thinking the same, and some time ago I already tried that but then stopped because this involves bigger changes.

There are two monster functions that already share a lot of code:
`Application::runApplication()` with around 500 lines, located in src/Gui/Application.cpp
`QWidget* setupMainWindow()` with around 300 lines, located in `src/Main/FreeCADGuiPy.cpp`

I can try again to break the functions up in smaller chunks that can be called from both places, and I will need some help. There are some code lines where I cannot grasp the reasoning behind implementing them differently, etc..

bdieterm:
> can you make a function in there that gets called by both places?

Done. It is actually two functions.

PR #12096, opened 2024-01-23
[PD] Add a Suppressed property to disable a feature

### LINKED ISSUE #12073: [Feature Request] PD: add a Passthrough boolean property to allow to disable a feature

### Problem description

We currently can't disable a feature without remove it.
I propose to add a Passthrough boolean property: when true the feature Base shape is copied as the result of the operation, when false the feature compute its own result.

### Full version info

```shell
OS: Debian GNU/Linux trixie/sid (XFCE/xfce)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35649 (Git)
Build type: Release
Branch: main
Hash: 35e09e233c4fadb6194c82164ece25d5043d23fe
Python 3.11.7, Qt 5.15.10, Coin 4.0.2, Vtk 9.1.0, OCC 7.6.3
Locale: French/France (fr_FR)
```

### Subproject(s) affected?

PartDesign

### Anything else?

_No response_

### Code of Conduct

FlachyJoe:
Aims to fix #12073

### DIFF
--- a/src/Mod/PartDesign/App/Feature.cpp
+++ b/src/Mod/PartDesign/App/Feature.cpp
@@ -72,7 +72,8 @@ App::DocumentObjectExecReturn* Feature::recompute()
         }
     }
     catch (Base::Exception& e) {
-        //invalid BaseShape, do nothing here but recompute below
+        //invalid BaseShape
+        Suppressed.setValue(false);
     }
 
     return DocumentObject::recompute();
--- a/src/Mod/PartDesign/App/Feature.cpp
+++ b/src/Mod/PartDesign/App/Feature.cpp
@@ -54,10 +54,28 @@ PROPERTY_SOURCE(PartDesign::Feature,Part::Feature)
 Feature::Feature()
 {
     ADD_PROPERTY(BaseFeature,(nullptr));
+    ADD_PROPERTY(Suppressed,(false));
     ADD_PROPERTY_TYPE(_Body,(nullptr),"Base",(App::PropertyType)(
                 App::Prop_ReadOnly|App::Prop_Hidden|App::Prop_Output|App::Prop_Transient),0);
     Placement.setStatus(App::Property::Hidden, true);
     BaseFeature.setStatus(App::Property::Hidden, true);
+    Suppressed.setStatus(App::Property::Hidden, true);
+}
+
+App::DocumentObjectExecReturn* Feature::recompute()
+{
+    try {
+        auto baseShape = getBaseShape();
+        if (Suppressed.getValue()) {
+            this->Shape.setValue(baseShape);
+            return StdReturn;
+        }
+    }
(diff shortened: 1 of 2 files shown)


src/Mod/PartDesign/App/FeatureHole.cpp
```diff
@@ -1664,6 +1664,10 @@ App::DocumentObjectExecReturn* Hole::execute()
     TopoDS_Shape base;
     try {
         base = getBaseShape();
+        if (Passthrough.getValue()) {
+            this->Shape.setValue(base);
+            return App::DocumentObject::StdReturn;
+        }
```
kadet1090, line 1670: This gets repeated over and over again - is there a better place for this piece of code so it does not have to be duplicated? That way feature developer need to remember to cover this case which is common for every feature and thus should be higher in the hierarchy.

I am not sure however if this is possible at all now as this probably would have to be done on the caller side?

FlachyJoe: `execute()` is called by `DocumentObject::recompute()` It's not a PD specific procedure so we can't tweak this.

kadet1090: Yeah, I checked how "Refine" which is also common property is implemented and well, it is also must be called in all features. So this is fine I suppose, maybe an idea for later refactor. I'm marking this as resolved as it is not a PR that should do a major overhaul of how features are implemented.

kadet1090 (approved):

src/Mod/PartDesign/App/FeatureHole.cpp
FlachyJoe: @kadet1090 I found a solution by override `recompute`

FlachyJoe: It doesn't work as expected, I can edit a existing feature but no more create a new one.
Fall back to working previous but still work on this solution.

FEA-eng:
It would be awesome to have it. Could it work for FEM features too ? In fact, I was going to create a feature request for that.

FlachyJoe:
Rewritten, squashed and rebased. Ready for review.

Side effects:

- if a suppressed feature is used in a pattern, the pattern show its normal result minus the original.
- it's a good manner to reduce a file size.

FlachyJoe:
> Could it work for FEM features too ?

The logic is completely different. Feel free to open a FR.

Dutchtallman:
> > Could it work for FEM features too ?
> 
> The logic is completely different. Feel free to open a FR.

Why is the logic different? ... to me, the high level concept is the same, but lower level implications are different... similar to 'addition' as a high level concept works out differently on lower level: integers or strings... Should there be generic 'suppress feature'?... with different implementations in different workbenches?

PR #12111, opened 2024-01-25
Draft: Qt6 fixes

3x380V:
It is still failing with:
```
  File "Mod/Draft/draftutils/init_draft_statusbar.py", line 247, in init_draft_statusbar_snap
    snap_widget.addAction(Gui.Command.get("Draft_ToggleGrid").getAction()[0])
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Base.FreeCADError: basic_string: construction from null is not valid
```
Draft status will be removed once it starts working.


3x380V:
@Roy-043, any hint?

3x380V:
Hmm, does not help (even 3500ms). There must be something else wrong. Immediately after adding command (as shown previously), `getAction()` is returning an empty string. But calling the same from `init_draft_statusbar_snap()`
```
    # grid button:
    cmd = Gui.Command.get("Draft_ToggleGrid")
    print(cmd)
    print(cmd.getAction())

```
does just:
```
    print(cmd.getAction())
          ^^^^^^^^^^^^^^^
Base.FreeCADError: basic_string: construction from null is not valid
```

Roy-043:
Strange. I can't reproduce with the files from this PR copied into this weekly version:
```
OS: Windows 8 build 9600
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35720 (Git)
Build type: Release
Branch: main
Hash: 00cd50bbf49c810555899efe112ec7595c76cd3a
Installed mods:
```
Do you experience the issue with the original files (from before this PR) as well?

3x380V:
I haven't tried to unpack AppImage and replace files there yet. I'm building against Qt6 and noticed that almost every WB is expecting widgets in QtGui, which is no longer true. That probably also means noone is actually using FreeCAD build against Qt6 which comes as a surprise after noticing that [forum thread](https://forum.freecad.org/viewtopic.php?t=65089&start=162)
I'll have probably setup Qt5 environment as well.

3x380V:
I'm now able to build FreeCAD against both Qt5 and Qt6. That `Base.FreeCADError: basic_string: construction from null is not valid` issue is not present in Qt5 built, so I'm in favor to make this a regular pull request.
@Roy-043, could you review these changes?

3x380V:
I'm not using PySide C++ support with Qt6. It turns out, that switching to its Python interface even with Qt5 does not work - just try to do what comment says:
https://github.com/FreeCAD/FreeCAD/blob/02c6a90c160a7feac269d8aa0fe1b490bfa5c898/src/Gui/PythonWrapper.cpp#L37
There's a #12088 to fix that, so now it is clear, I need to finish that one first :)

3x380V:
Fixed in #12088, however `qt_wrapInstance` in `src/Gui/PythonWrapper.cpp` is hacked as:
```
    if (className == "QAction") {
      // HACK: SEGFAULTs with QtCore...
      Py::Module qtmod(importPySide("QtGui"));
      arguments[1] = qtmod.getDict().getItem(className);
    } else {
      Py::Module qtmod(importPySide(moduleName));
      arguments[1] = qtmod.getDict().getItem(className);
    }
    return func.apply(arguments);
```
I run out of time till the end of month, so leaving it as an exercise for the reader. Calling `qtmod.getDict().getItem(className)` with QtCore for QAction SEGFAULTs.

Roy-043:
Thank you!

PR #12132, opened 2024-01-26
[Gui] Behave-Dark StyleSheet/Theme Maintenance

Syres916:
This is purely overlay code additions only

Replaces #12039


kadet1090:
For stuff that touches UI it would be good to attach before and after screenshots of visual changes.

PR #12138, opened 2024-01-26
Toponaming: FeaturePartCommon

bgbsww:


src/Mod/Part/App/FeaturePartCommon.cpp
```diff
@@ -194,4 +195,33 @@ App::DocumentObjectExecReturn *MultiCommon::execute()
     }
 
     return App::DocumentObject::StdReturn;
+#else
+    std::vector<TopoShape> shapes;
+    for(auto obj : Shapes.getValues()) {
+        TopoShape sh = Feature::getTopoShape(obj);
+        if(sh.isNull())
```
chennes, line 202: Add curly braces around these one-line conditionals and run clang-format on these lines.

bgbsww: Duh, How did I miss this?  New ide is why it happened, but I should have caught it.

PR #12140, opened 2024-01-26
Material: Return new reference to Python boolean and change format unit to parse path

marioalexis84:
`Py_False` or `Py_True` should not be used directly on function return without increasing the reference count

### DIFF
--- a/src/Mod/Material/App/MaterialManagerPyImpl.cpp
+++ b/src/Mod/Material/App/MaterialManagerPyImpl.cpp
@@ -71,17 +71,20 @@ PyObject* MaterialManagerPy::getMaterial(PyObject* args)
 
 PyObject* MaterialManagerPy::getMaterialByPath(PyObject* args)
 {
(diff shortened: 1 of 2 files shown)


davesrocketshop (changes requested):
Similar changes will need to be made in ModelManagerPyImpl.cpp and ModelPyImpl.cpp

marioalexis84:
> Similar changes will need to be made in ModelManagerPyImpl.cpp and ModelPyImpl.cpp

Exactly where?
I can't find any other parsed file paths or returns of references to Python booleans in the suggested files.

davesrocketshop:
> > Similar changes will need to be made in ModelManagerPyImpl.cpp and ModelPyImpl.cpp
> 
> Exactly where? I can't find any other parsed file paths or returns of references to Python booleans in the suggested files.

Apologies. For some reason I think model when looking at material. I was looking at the wrong files.

PR #12159, opened 2024-01-27
Base: Fix #12061

3x380V:
Reported as [Annoying Warning regression when saving macros](https://forum.freecad.org/viewtopic.php?t=84758)


yorikvanhavre (approved):
I don't see the warning message anymore, so the problem seems solved

yorikvanhavre:
Ok then! Thanks for the fix!

PR #12160, opened 2024-01-28
Tests/Toponaming: Add test for single shapes

CalligaroV:
Added test for Tag Info generated by makeShapeWithElementMap

**!!!IMPORTANT NOTE!!!:** This test makes sense only for single shapes and shapes that aren't generated by geometrical operations.

A more robust version of this test will be developed, if needed, to test more complex shapes

### DIFF
--- a/tests/src/Mod/Part/App/TopoShapeMakeShapeWithElementMap.cpp
+++ b/tests/src/Mod/Part/App/TopoShapeMakeShapeWithElementMap.cpp
@@ -340,3 +340,48 @@ TEST_F(TopoShapeMakeShapeWithElementMapTests, findMakerOpInElementMap)
         }
     }
 }
+
+std::string composeTagInfo(const MappedElement& element, const TopoShape& shape)
+{
+    std::string elementNameStr {element.name.constPostfix()};
+    std::string tagInfo = POSTFIX_TAG + std::to_string(shape.Tag);
+    tagInfo +=
+        ":" + std::to_string(elementNameStr.substr(0, elementNameStr.find(tagInfo)).length());
+
+    return tagInfo;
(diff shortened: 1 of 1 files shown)


chennes:
@bgbsww this look good to you?

tests/src/Mod/Part/App/TopoShapeMakeShapeWithElementMap.cpp
```diff
@@ -340,3 +340,45 @@ TEST_F(TopoShapeMakeShapeWithElementMapTests, findMakerOpInElementMap)
...
+                         // more or less the same of nonMappableSources
+
+    // Act and assert
+    // Testing with all the source TopoShapes
+    for (const auto& source : sources) {
+        Part::TopoShape tmpShape {source.getShape()};
+        tmpShape.makeShapeWithElementMap(source.getShape(), *Mapper(), sources);
+
+        // For all the mappedElements ...
+        for (const auto& mappedElement : tmpShape.getElementMap()) {
```
bgbsww, line 371: I think this test looks good.  Note that if getElementMap() returns empty, then no testing will be done at all.  getElementMap() is tested separately, so not necessarily required here.  I think I'd add an expect for the length of the element map just as a compulsive double check that the test is doing something.

bgbsww (approved):
Change not required, just a thought.

tests/src/Mod/Part/App/TopoShapeMakeShapeWithElementMap.cpp
CalligaroV: Thanks for the hint!

I'll add a check right now.

PR #12161, opened 2024-01-28
Improvements to Sketcher context menu

maxwxyz:
This PR improves the contextual right click menu in the Sketcher WB.
1. Consider preselection: You can now directly right click on an element and pull up the contextual menu for the element you've clicked on. This also works if you already have a selection, then it considers the item you've right clicked on for the contextual commands. For example select a circle and then follow with a right click on a line will pull up the menu with dimension (distance) and tangent. Previously it would only considered the selected elements, so you would need one extra left click. This is also the behavior in other (CAD) software tools. 
2. Add functionality to also consider external geometry.
3. Remove commands which (currently) cannot work with a selection (e.g. fillet, trim). They would need to clear the selection and the elements have to be selected again with the respective tool being active beforehand.
4. Add commands for Sketcher copy, cut and paste.

https://github.com/FreeCAD/FreeCAD/assets/6246609/551ea07f-c24b-4c92-92b7-0382be1d10fe

### DIFF
--- a/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
+++ b/src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
@@ -968,14 +968,7 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                         std::stringstream ss;
                         ss << "Vertex" << preselection.getPreselectionVertexIndex();
 
-                        if (isSelected(ss.str())) {
-                            rmvSelection(ss.str());
-                        }
-                        else {
-                            addSelection2(
-                                ss.str(), pp->getPoint()[0], pp->getPoint()[1], pp->getPoint()[2]);
-                            drag.resetIds();
-                        }
+                        preselectToSelection(ss, pp, true);
                     }
                     Mode = STATUS_NONE;
                     return true;
@@ -988,16 +981,7 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                         else// external geometry
                             ss << "ExternalEdge" << preselection.getPreselectionExternalEdgeIndex();
 
-                        // If edge already selected move from selection
-                        if (isSelected(ss.str())) {
-                            rmvSelection(ss.str());
-                        }
-                        else {
-                            // Add edge to the selection
-                            addSelection2(
-                                ss.str(), pp->getPoint()[0], pp->getPoint()[1], pp->getPoint()[2]);
-                            drag.resetIds();
-                        }
+                        preselectToSelection(ss, pp, true);
                     }
                     Mode = STATUS_NONE;
                     return true;
@@ -1019,16 +1003,7 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
                                 break;
                         }
 
-                        // If cross already selected move from selection
-                        if (isSelected(ss.str())) {
-                            rmvSelection(ss.str());
-                        }
-                        else {
-                            // Add cross to the selection
-                            addSelection2(
-                                ss.str(), pp->getPoint()[0], pp->getPoint()[1], pp->getPoint()[2]);
-                            drag.resetIds();
-                        }
+                        preselectToSelection(ss, pp, true);
                     }
                     Mode = STATUS_NONE;
                     return true;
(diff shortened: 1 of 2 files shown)


src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1210,17 +1210,129 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
     }
     // Right mouse button ****************************************************
     else if (Button == 2) {
-        if (!pressed) {
+        if (pressed) {
+            // Do things depending on the mode of the user interaction
+            switch (Mode) {
+                case STATUS_NONE: {
+                    bool done = false;
```
chennes, line 1217: What did you intend for this (currently unused) variable?

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
```diff
@@ -1210,17 +1210,129 @@ bool ViewProviderSketch::mouseButtonPressed(int Button, bool pressed, const SbVe
...
+                        std::stringstream ss;
+                        ss << "Vertex" << preselection.getPreselectionVertexIndex();
+
+                        if (!isSelected(ss.str())) {
+                            addSelection2(ss.str(),
+                                          pp->getPoint()[0],
+                                          pp->getPoint()[1],
+                                          pp->getPoint()[2]);
+                            drag.resetIds();
+                        }
```
chennes, line 1263: Can you refactor so we don't repeat this so many times?

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
maxwxyz: removed it

src/Mod/Sketcher/Gui/ViewProviderSketch.cpp
maxwxyz: done, and also for the LMB event.

maxwxyz:
@chennes done

Roy-043:
It seems that for only some pre-selections the correct tools are added to the context menu. Several combinations have been missed.

For example: Sketcher_ConstrainCoincidentUnified allows the following pre-selections:

1. Select two or more points.
2. Select two or more edges of circles, arcs, ellipses or arcs of ellipses.
3. Select a single point and a single edge (in any order).
4. Select several points and a single edge (idem).
5. Select a single point and several edges (idem).

1+2+3+4 are correctly handled, but option 5 is not.

Another example: Sketcher_ConstrainSymmetric allows the following pre-selections:

1. Select two points and a symmetry point (in that order).
2. Select two points and a symmetry line (in any order).
3. Select a line and a symmetry point (idem).

2 is correctly handled, but 1+3 are not.

I have yet to check the other tools.

maxwxyz:
@Roy-043 thanks! will add them. 
If you find more, just let me know.

Roy-043:
Some more:

#### Sketcher_ConstrainParallel

Points can be included in the pre-selection since V0.22 (they will be ignored). The context menu option is then missing.

#### Sketcher_ConstrainPerpendicular and Sketcher_ConstrainTangent

Two edges and a point can be pre-selected. The context menu option is then missing.

#### Sketcher_Dimension

Seems to be included too often. Even if the pre-selection can't be used. For example 3 circles and 1 line.

#### Sketcher_ToggleDrivingConstraint

Included for all constraints, but geometric constrains cannot be toggled (red warning).

#### Sketcher_ToggleConstruction

Included if only internal geometry has been selected which cannot be toggled (red warning). For example axes of ellipses and weight circles of splines.

Also included if only external geometry has been selected. This also cannot be toggled, but there is no warning when you try to do so.

maxwxyz:
@Roy-043 thanks, will make a PR for the easy fixe. The other cases with internal/external or driving is currently not implemented and need some more rework later on.

Roy-043:
Another related issue is the order of the context menu:

Main menu / toolbar:
* Offset geometry
* Polar transform
* Scale transform
* Symmetry
* Array transform

Context menu:
* Array transform
* Polar transform
* Scale transform
* Offset geometry

If there is not a good reason for the different ordering, I would prefer to see the order of the context menu changed to match the main menu.

maxwxyz:
@Roy-043 https://github.com/FreeCAD/FreeCAD/pull/13229

Issue #12172, opened 2024-01-28
[Problem] Axis letters in coordinate system barely visible with dark background


kadet1090:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Using dark backgrounds like the one from OpenDark the letters for coordinate system axes are barely visible due to them being hardcoded to be black.

![image](https://github.com/FreeCAD/FreeCAD/assets/747404/4fd76133-da37-43cc-8521-541f9d72aac6)

### Full version info

```shell
0.22dev
```

### Subproject(s) affected?

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #12174, opened 2024-01-28
Gui: Add ability to set coordinate system axis letter color

kadet1090:
This adds new preference to set color of the letters in coordinate system to be used by theme designers.

image

Fixes: #12172


obelisk79:
Thank you, this is a much needed feature.

obelisk79:
> @obelisk79 does this really need to be exposed in the GUI? It feels like a pretty esoteric setting to me, that will in nearly all cases be handled by themes.

Exposing this in the parameter editor would be adequate. Currently the letter colors for this UI element can't be configured at all as the color is hard coded as I understand it.

kadet1090:
There is a problem with discoverability if it is not in the UI - noone will know that such setting is available if it is not exposed anywhere. To back my words feature to lock toolbars was available for long time but it was not exposed to ui and noone knew about it. There is also another problem - the value that is saved is also not very user-friendly to type by hand in the XML or parameters window. 

That said I also don't like exposing this in the UI so I see two possibilities:

1. We need a good article on wiki which lists all parameters useful for theme designers and some util to properly encode colors
2. We need to have simple and advanced settings. This should not be exposed to typical user but it can be if someone opts to have advanced options which can imply a lot of settings.

I'd prefer the second one but it requires a lot of work. Nonetheless discoverability of such parameters must be improved if such options are not exposed via settings.

kadet1090:
Problem with documentation is that it has to be written, maintained and users have to know about it. `LockToolbars` example that I mentioned in my comment is there, but it did not help much with discoverability for the common user. However, theme designers should read the docs anyway so it may be absolutely enough. I don't feel that this decision is mine to take so - @chennes if we accept issues mentioned in my comment I'll gladly remove the UI changes from the PR.

src/Gui/PreferencePages/DlgSettings3DView.ui
```diff
@@ -93,6 +93,33 @@ in the corner -- in % of height/width of viewport</string>
...
+         <widget class="QLabel" name="axisLetterColorLabel">
+          <property name="text">
+           <string>Letter color:</string>
+          </property>
+         </widget>
+        </item>
+        <item row="1" column="3">
+         <widget class="Gui::PrefColorButton" name="axisLetterColor">
+          <property name="toolTip">
+           <string>Axis letter color</string>
```
luzpaz, line 106: Can the tooltip be more elaborate ?

adrianinsaval:
maybe we should just make this automatic as in the navigation cube and then configurable only in a parameter?

PR #12177, opened 2024-01-28
Toponaming shapeprotector Nullify() check

CalligaroV:
related to comment in https://github.com/FreeCAD/FreeCAD/commit/a74c6b339160c48b26b8bc191138a583a6a2aee4#comments of https://github.com/FreeCAD/FreeCAD/pull/11896

This PR adds a check on objects of class `ShapeProtector` before performing `Nullify()`

as reported in https://forum.freecad.org/viewtopic.php?t=84536 the lack of this check causes a segfault when the `Nullify()` method is called

### DIFF
--- a/src/Mod/Part/App/TopoShape.cpp
+++ b/src/Mod/Part/App/TopoShape.cpp
@@ -3911,7 +3911,9 @@ TopoDS_Shape TopoShape::makeShell(const TopoDS_Shape& input) const
 
 TopoShape &TopoShape::makeWires(const TopoShape &shape, const char *op, bool fix, double tol)
 {
-    _Shape.Nullify();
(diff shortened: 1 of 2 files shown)


CalligaroV:
### Additional notes
during tests I found that having a file opened with automatic recompute enabled, the opening time of files with lots of shapes takes more time than having it disabled and then recompute when the file is completely loaded.

Don't know if in most of cases it is preferred that the automatic recompute is enabled, but if it wasn't so it may be a good idea to have it disabled during opening files.

CalligaroV:
@chennes, should we add also this PR to the [Toponaming Project](https://github.com/orgs/FreeCAD/projects/2)?

chennes:
Done!

PR #12180, opened 2024-01-29
Draft: Draft_Rotate fix Base angle behavior

### LINKED ISSUE #12175: [Problem] Base angle doesn't work when using input field in Draft Rotate

### Problem description

Draft Rotate has the Base angle setting. However, it works only when the tool is used by selecting the points. When the base angle is specified in the input field, it has no effect on rotation. This makes the base angle setting even more confusing.

Forum discussion: https://forum.freecad.org/viewtopic.php?t=69525

### Full version info

```shell
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35691 (Git)
Build type: Release
Branch: main
Hash: cc45b58dadae53d7a99be1ed7db25f6c93bb13d5
Locale: Polish/Poland (pl_PL)
```

### Subproject(s) affected?

Draft

### Anything else?

_No response_

### Code of Conduct

Roy-043:
Fixes #12175.

The Base angle is subtracted from the Rotation angle (which always starts from the +X axis). The behavior of the command when clicking points in the 3D view remains the same, the values displayed in the task panel are different though.

### DIFF
--- a/src/Mod/Draft/draftguitools/gui_rotate.py
+++ b/src/Mod/Draft/draftguitools/gui_rotate.py
@@ -60,7 +60,7 @@ class Rotate(gui_base_original.Modifier):
         return {'Pixmap': 'Draft_Rotate',
                 'Accel': "R, O",
                 'MenuText': QT_TRANSLATE_NOOP("Draft_Rotate", "Rotate"),
-                'ToolTip': QT_TRANSLATE_NOOP("Draft_Rotate", "Rotates the selected objects. Choose the center of rotation, then the initial angle, and then the final angle.\nIf the \"copy\" option is active, it will create rotated copies.\nCTRL to snap, SHIFT to constrain. Hold ALT and click to create a copy with each click.")}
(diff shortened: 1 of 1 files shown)


Roy-043:
Windows build seems to on the blink ATM.

yorikvanhavre:
Does it make sense? I mean, someone usually wants to rotate something, say, 45 degrees. Now if they start at for ex. 11.583 degrees, they'll need to enter 56.583? This seems counter-intuitive, no?

Roy-043:
If you want to rotate an object 45 degrees, you can just enter zero for the base angle. Or if you are clicking points you can pick a random point constrained to the +X direction from the rotation center.

FWIW this is how the base angle is handled in DwgCAD. But providing a base angle is optional there.

yorikvanhavre:
I still would want to, for ex, pick an alignment (which is not 0) and specify, say, 45° from that...

yorikvanhavre:
I still think this is a wrong approach, and is changing a paradigm without a real reason...
I think what needs to be fixed is the wrong behaviour described in #12175 , but not change the modelling paradigm.

I can work on an alternate fix if you prefer @Roy-043

Roy-043:
Hmm, I think there is a real reason for this proposal, and also some logic involved. But maybe I my DwgCAD background has closed my eyes to alternative options. So by all means, make a proposal.

PR #12181, opened 2024-01-29
Dim the Navigation Cube when not active

LemonBoy:
Implement the InactiveOpacity option, allowing the user to dim the navigation cube when not focused by making it transparent.

This feature is present in many CAD softwares and IMO is quite useful to make the cube slightly less noticeable when working on the model.

### DIFF
--- a/src/Gui/NaviCube.cpp
+++ b/src/Gui/NaviCube.cpp
@@ -156,7 +156,7 @@ private:
 
     QString str(const char* str);
     QMenu* createNaviCubeMenu();
-    void drawNaviCube(bool picking);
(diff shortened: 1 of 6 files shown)


Rexbas:
Nice work, I tried it and it works well. Maybe split the _opacity when inactive_ for the cube and the flat buttons around it in a follow up PR. This way it would be possible to completely hide the flat buttons when the mouse is not near the cube, so it could looks like this:
image

LemonBoy:
FYI, It was noted [on the forum](https://forum.freecad.org/viewtopic.php?t=85209) that this setting specifies the opacity while the other settings refer to the transparency.

PR #12189, opened 2024-01-29
Toponaming makeElementWire

CalligaroV:
PR for https://github.com/FreeCAD/FreeCAD/commit/04e470a34293e570cb880f198343c4fa13089b1b by @chennes and @bgbsww

merged with main branch at https://github.com/FreeCAD/FreeCAD/commit/90d36f60c30cb6549a7987fcd9d4769eab1c160c


CalligaroV:
Converted into draft because some tests introduced by this PR still fail.

src/Mod/Part/App/TopoShape.h
```diff
@@ -727,22 +736,175 @@ class PartExport TopoShape: public Data::ComplexGeoData
...
                                        const std::vector<TopoShape> &sources,
                                        const char *op=nullptr);
+    /**
+     * When given a single shape to create a compound, two results are possible: either to simply
+     * return the shape as given, or to force it to be placed in a Compound.
+     */
+    enum class SingleShapeCompoundCreationPolicy {
+        RETURN_SHAPE,
+        FORCE_COMPOUND
+    };
```
bgbsww, line 746: Up around line 101 we have an enum class that uses lowercaseCapword rather than UPPER_CASE.  We should probably make a call and be consistent.

src/Mod/Part/App/TopoShape.h
```diff
@@ -1223,25 +1385,120 @@ class PartExport TopoShape: public Data::ComplexGeoData
...
+struct PartExport MapperMaker: TopoShape::Mapper {
+    BRepBuilderAPI_MakeShape &maker;
+    MapperMaker(BRepBuilderAPI_MakeShape &maker)
+        :maker(maker)
     {}
-    const std::vector<TopoDS_Shape>& modified(const TopoDS_Shape& s) const override;
-    const std::vector<TopoDS_Shape>& generated(const TopoDS_Shape& s) const override;
+    virtual const std::vector<TopoDS_Shape> &modified(const TopoDS_Shape &s) const override;
+    virtual const std::vector<TopoDS_Shape> &generated(const TopoDS_Shape &s) const override;
 };
```
bgbsww, line 1486: These three lint fixes, maybe?  And the ones below?  I've been following the lint recommendations on adding explicit and removing virtual.

src/Mod/Part/App/TopoShapeMapper.h
```diff
@@ -38,75 +39,6 @@ class ShapeFix_Root;
...
-        return a.first.getShape().IsSame(b.first.getShape())
-            && a.second.getShape().IsSame(b.second.getShape());
-    }
-    inline bool operator()(const std::pair<TopoDS_Shape, TopoDS_Shape>& a,
-                           const std::pair<TopoDS_Shape, TopoDS_Shape>& b) const
-    {
-        return a.first.IsSame(b.first) && a.second.IsSame(b.second);
-    }
-};
-
```
bgbsww, line 109: Looks like we duplicated.  I placed it in this new file to work towards reducing the size of the TopoShape.h file.  Either location is okay with me.

src/Mod/Part/App/TopoShape.h
chennes: Please use camel case rather than upper case -- I've been guilty of screwing this up myself so I apologize for the bad example. Modern C++ basically eschews all uppercase because it looks like macros.

CalligaroV: Got it!

Changed all enum classes members to lowercaseCapword, you should find this modification in the next commit

src/Mod/Part/App/TopoShape.h
CalligaroV: Missed these lint warning because on the previous comments I ran clang-tidy only on `TopoShapeExpansion.cpp` test file

Should be ok in the next commit

src/Mod/Part/App/TopoShapeMapper.h
CalligaroV: I remember I had compiling issues with the first merge also because of this.

Now I moved back here and I don't have compiler issues again.
Probably I messed up something during that merge but now it looks old story.

PR #12190, opened 2024-01-29
Conda: Pin PCL to v1.13.1 to fix build errors on Windows.

oursland:
An update to the `conda-forge` package `pcl` from version `1.13.1` to `1.14.0` resulted in build failures only present on Windows.  Pinning the package to `1.13.1` resolves the build issues.

### DIFF
--- a/conda/environment.devenv.yml
+++ b/conda/environment.devenv.yml
@@ -79,7 +79,7 @@ dependencies:
 - numpy
 - occt==7.6.3
 - openssl
-- pcl
+- pcl==1.13.1
 - pip
 - pivy
 - pkg-config


bgbsww:
Awesome find.  I'm suspicious of https://github.com/PointCloudLibrary/pcl/pull/5744 as a cause.

adrianinsaval:
@oursland maybe we should pin all packages and updated them every now and then once we know we can use updated packages so we don't get surprises like this all the time

oursland:
@adrianinsaval I prefer the opposite of letting the package versions float.  This way we receive changes, typically bugfix and security updates, as they are released.  CI will identify when a given package change breaks, and we can respond accordingly, as was the case here.

If we instead pin all of the packages, it puts the onus on us to track all of the dependencies and subdependencies for bugfixes and security updates.

bgbsww:
With the twin caveats that (a) there are exceptions to every rule, and (b) there will be times like today where this needs a (hopefully) temporary override.

From an overall philosophy sense, if we can't build with the latest stable version of a dependency we shouldn't be using it because sooner or later something bad is going to happen.

The opposite approach needs a staff person assigned to track, test, and correct issues as dependencies evolve.  CI + volunteer reaction is currently FreeCAD's version of this, and it seems unlikely to change.

Should we be considering creating open issues for pinned dependencies?

Also, this is a big conversation for only you two plus newcomer me.  Does it need to be raised elsewhere?

chennes:
I would make an Issue of some kind to track that we've done this and should attempt to undo it at some point. Separately you could consider making an Issue to discuss this process.

PR #12196, opened 2024-01-30
Avoid sending unscaped quotes to python console

adrianinsaval:
reported at https://forum.freecad.org/viewtopic.php?p=714249#p714249 and likely other places

### DIFF
--- a/src/Base/Tools.cpp
+++ b/src/Base/Tools.cpp
@@ -236,6 +236,26 @@ std::string Base::Tools::escapedUnicodeToUtf8(const std::string& s)
     return string;
 }
 
+std::string Base::Tools::escapeQuotesFromString(const std::string& s)
+{
+    std::string result;
+    size_t len = s.size();
+    for (size_t i = 0; i < len; ++i) {
+        switch (s.at(i)) {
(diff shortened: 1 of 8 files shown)


Roy-043:
Does this fix #12204 and #12206?

adrianinsaval:
No but the solution should be similar, I'll look into those today at the hackathon

adrianinsaval:
Found another way to trigger a bug like this, name an object with a label like 5', create a link to it: 
```
08:30:14  pyException: Traceback (most recent call last):
  File "<string>", line 1, in <module>
<class 'SyntaxError'>: ('unterminated string literal (detected at line 1)', ('<string>', 1, 56, "App.getDocument('Unnamed1').getObject('Link').Label='5''", 1, 56))
```

adrianinsaval:
This PR can be merged, I'll make another one to fix #12204 and #12206, for these it might be a good idea to add escaping quotes to `Base::Quantity::getSafeUserString` that is used by both Part and PartDesign primitive commands and probably a bunch of others, but I have to check that this wouldn't have undesired effects somewhere else

PR #12205, opened 2024-01-30
Gui: Add easing to navigation animations

Rexbas:
This adds an option (enabled by default) to use the InOutCubic easing function in navigation animations. By using this, the start and end of the animations are more smooth instead of directly going to the max speed or 0.

### DIFF
--- a/src/Gui/NavigationAnimation.cpp
+++ b/src/Gui/NavigationAnimation.cpp
@@ -44,7 +44,7 @@ void NavigationAnimation::onStop([[maybe_unused]] bool finished)
 
 FixedTimeAnimation::FixedTimeAnimation(NavigationStyle* navigation, const SbRotation& orientation,
                                        const SbVec3f& rotationCenter, const SbVec3f& translation,
-                                       int duration)
+                                       int duration, const QEasingCurve::Type easingCurve)
     : NavigationAnimation(navigation)
     , targetOrientation(orientation)
     , targetTranslation(translation)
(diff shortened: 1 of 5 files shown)


chennes:
Can you make this an `enum class` instead of a boolean, so that we can avoid "boolean blindness"?

adrianinsaval:
do we need this in the preference page? to me this is something that it's ok to have as just a parameter

Rexbas:
> Can you make this an enum class instead of a boolean

What about `QEasingCurve::Type`? This would also make it possible to change the easing curve to something else than Linear or InOutCubic.

> do we need this in the preference page?

I don't know, I think it fits well where I have placed it.

adrianinsaval:
> I don't know, I think it fits well where I have placed it.

it kind of fits but we already have too many preferences as it is, if we don't need to I prefer to not add more stuff. Are we expecting people to actually be changing this preference? Seems like a very obscure preference to me, not worth having a checkbox.

Rexbas:
I think you are right, I will remove the checkbox.

Rexbas:
I have removed the checkbox and changed the type to `QEasingCurve::Type`. The easing curve can be changed by modifying the parameter `BaseApp/Preferences/View/NavigationAnimationEasingCurve`. All possible values are described [here](https://doc.qt.io/qt-5/qeasingcurve.html#Type-enum)

chennes:
> What about `QEasingCurve::Type`? This would also make it possible to change the easing curve to something else than Linear or InOutCubic.

Fine with me, it does support some future growth if someone ever gets it into their head to go nuts with the animations :).

Issue #12049, opened 2024-01-20
[Feature Request] Quick way to see basic measurements on selection


PaddleStroke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Getting basic information about things is a basic need in design :  edge length, selected circle/arc radius, distance between 2 selected vertices, volume of a solid, area of a face ...

There is no easy way to get all of these easily in core. And so this feature has been implemented in many addons / workbench / branches. A few coming to mind is FCInfo, FCInfo toolbar, quick measure, RT branch has something like this too. etc.

My proposal is the same as the one described by the OP there : https://forum.freecad.org/viewtopic.php?style=4&t=80919

ie to add a label in the status bar with useful information for example : 
![image](https://github.com/FreeCAD/FreeCAD/assets/19984177/4bd2d30c-e2d6-496c-83bb-9bc18fcde441)

This label would show meaningful information about the selection depending on what it is : 
1 point : coordinates
2 points : distance
1 line : length
1 face : area
etc.

One thing I noted is that it would be useful to be able to copy the information.

This has been discussed a little bit in https://github.com/FreeCAD/FreeCAD/pull/11344

### Full version info

```shell
0.22
```

### Subproject(s) affected?

Core

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chennes:
I agree that this is a problem, and I like your solution -- @obelisk79 has the Design Working Group considered this issue? Is there some advice in the Handbook that we can look to?

WandererFan:
This sounds a lot like the gsoc Unified Measurement Facility.

https://forum.freecad.org/viewtopic.php?p=733478#p733478

obelisk79:
I agree with WandererFan, there is a bit of overlap with the unified measurement facility. The DWG has discussed this kind of feature in the past. I think there is some value in displaying non-static simple information in the status bar, ie length, radius/diameter and area of flat faces. I'd probably stop there though to minimize overhead and keep behavior simple, understandable and easily predictable, leave the other type of calculations to the unified measurement functionality.

From the Design Guide:
```
Status Bar: The status bar provides contextual information about the position of the mouse cursor in relation to the objects in the main view. This section of the UI is also home to some global controls within FreeCAD. They are “Notifications”, “Navigation Style”, and “Unit Selector”. Developers shall not insert non-global elements or widgets into this area.
```

The key here being dynamic, contextual information.

DanMiel:
[...]

PR #12217, opened 2024-01-31
Core / Measure: Introduce QuickMeasure

PaddleStroke:
Introduce quick measure. This adds a small label on the status bar on the right side that gives basic information about selection :
- lenght of edges
- Area of faces
- Volume of solids
- distances ...

Also as a small thing, it makes the text of the status bar label selectable so you can select it and copy it.

Fixes https://github.com/FreeCAD/FreeCAD/issues/12049

### DIFF
--- a/src/Mod/Measure/App/Measurement.h
+++ b/src/Mod/Measure/App/Measurement.h
@@ -36,10 +36,20 @@
 class TopoDS_Shape;
 namespace Measure
 {
- enum MeasureType {
+ enum class MeasureType {
         Volumes, // Measure the Volume(s)
         Edges, // Measure the Edge(s)
+        Line, // One Line
+        TwoLines, // Two lines
+        TwoParallelLines, // Two parallel lines
+        Circle, // One circle
         Surfaces, // Measure the surface(s)
+        Cylinder, // One Cylinder
+        Cone, // One Cone
+        Sphere, // One Sphere
+        Torus, // One Torus
+        Plane, // One Plane
+        TwoPlanes, // One Plane
         Points,
         PointToPoint, // Measure between TWO points
         PointToEdge, // Measure between ONE point and ONE edge
@@ -66,6 +76,7 @@ public:
     int addReference3D(App::DocumentObject* obj, const char *subName);
 
     MeasureType getType();
+    MeasureType findType();
 
      // from base class
     PyObject *getPyObject() override;
@@ -74,6 +85,8 @@ public:
   // Methods for distances (edge length, two points, edge and a point
   double length() const;
   Base::Vector3d delta() const;                                                 //when would client use delta??
+  double lineLineDistance() const;
+  double planePlaneDistance() const;
 
   // Calculates the radius for an arc or circular edge
   double radius() const;
@@ -81,10 +94,19 @@ public:
   // Calculates the angle between two edges
   double angle(const Base::Vector3d &param = Base::Vector3d(0,0,0)) const;      //param is never used???
 
-  // Calculate volumetric/mass properties
+  // Calculate the center of mass
   Base::Vector3d massCenter() const;
 
+  // Calculate the volume of selected volumes
+  double volume() const;
+
+  // Calculate the area of selection
+  double area() const;
+
   static Base::Vector3d toVector3d(const gp_Pnt gp) { return Base::Vector3d(gp.X(), gp.Y(), gp.Z()); }
+
+  bool planesAreParallel() const;
+  bool linesAreParallel() const;
 
 protected:
   TopoDS_Shape getShape(App::DocumentObject *obj , const char *subName) const;
--- a/src/Mod/Measure/Gui/AppMeasureGui.cpp
+++ b/src/Mod/Measure/Gui/AppMeasureGui.cpp
@@ -30,6 +30,8 @@
 #include <Gui/WidgetFactory.h>
 
 #include "DlgPrefsMeasureAppearanceImp.h"
+#include "QuickMeasure.h"
+#include "QuickMeasurePy.h"
 #include "ViewProviderMeasureAngle.h"
 #include "ViewProviderMeasureDistance.h"
 #include "ViewProviderMeasureBase.h"
@@ -86,7 +88,7 @@ PyMOD_INIT_FUNC(MeasureGui)
     CreateMeasureCommands();
 
     MeasureGui::ViewProviderMeasureBase                ::init();
-    MeasureGui::ViewProviderMeasure        ::init();
+    MeasureGui::ViewProviderMeasure                    ::init();
     MeasureGui::ViewProviderMeasureAngle               ::init();
     MeasureGui::ViewProviderMeasureDistance            ::init();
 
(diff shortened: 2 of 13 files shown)


PaddleStroke:
Here's how it looks : 
image

chennes:
There seem to be a couple of small bugs here: in my testing, parallel edges don't show their distance, nor do two selected points.

src/Mod/Measure/App/Measurement.h
```diff
@@ -39,7 +39,17 @@ namespace Measure
  enum MeasureType {
```
chennes, line 39: Can you add it to your long-term list to migrate this to an `enum class`?

src/Mod/Measure/Gui/AppMeasureGui.cpp
```diff
@@ -0,0 +1,51 @@
...
+    PyObject* mod = MeasureGui::initModule();
+    Base::Console().Log("Loading MeasureGui module... done\n");
+
+
+    Base::Interpreter().addType(&MeasureGui::QuickMeasurePy::Type, mod, "QuickMeasure");
+
+    //MeasureGui::QuickMeasure::init();
+
+    PyMOD_Return(mod);
+}
```
chennes, line 51: Needs a newline

src/Mod/Measure/Gui/AppMeasureGuiPy.cpp
```diff
@@ -0,0 +1,46 @@
...
+            initialize("This module is the MeasureGui module.");  // register with Python
+        }
+    };
+
+    PyObject* initModule()
+    {
+        return Base::Interpreter().addModule(new Module);
+    }
+
+}  // namespace MeasureGui
```
chennes, line 46: Needs a newline

src/Mod/Measure/App/Measurement.h
PaddleStroke: I had a look and this enum is used in techdraws in ways that I don't fully understand. It seems to be associated to a PropertyEnumeration. So I am not sure how to handle this.

PaddleStroke: But yes I can add this to my long term list

PaddleStroke: After a second look, I think that TechDraw is actually not using this MeasureType enum.

PaddleStroke: So this is ok I changed it to enum class

PaddleStroke:
@chennes I fixed the end of files and changed to enum class.

PaddleStroke:
The CI is failing for MacOS but I can't understand why. Any idea @chennes ?

chennes:
Don't worry about it, that CI has been giving us problems quite frequently due to some sort of timeout. I believe @adrianinsaval in aware of the issue.

adrianinsaval:
I'll see if I can manage to reproduce the CI problem on the m1 mac once I get home, for now first just try re triggering the CI if you see mac failing every now and then

WandererFan:
@hlorus - comments re umf work in progress?

hlorus:
> @hlorus - comments re umf work in progress?

UMF doesn't utilize the Measurement class in Measurement.h, so conflicts should be minimal. However, there might be some code duplication, which would probably be nice to avoid.

PaddleStroke:
Ah indeed we are bound to have conflicts here, as for a start we both introduce the Measure/Gui module.
And yes we must have some code duplication as its doing something similar

DanMiel:
I would like to comment on putting the measurements on the bottom or top of the screen. I wrote the QuickMeasure Add-on and had the output going to the status bar. I found it awkward to move my eyes from what I was measuring to the status line and then back up to the part, I found it easier to create the existing text box and put it by what I was measuring, thereby keeping my eyes on the part and on the output.
Dan Miel.

This is copied from pull request #12049 which was closed before I wrote to it.

prokoudine:
So that it doesn't get lost...

1. Total length for adjacent edges is 0:

image

2. Volume is in ml for some reason, shouldn't it be cm^3?

image

DanMiel:
@prokoudine;
I’m using the original Quick Measure WB in these pictures. At this point I am not sure which program you are using. 
In picture 1 I have selected a body from the tree. It is a block 3x1x1/2. In the text box it gives the both the volume and the surface area of the body. Your program may be reporting both but missing the ^3.
In picture 2 Two lines are selected. The angle is 90 degrees, total length of the lines is 4 inches, and the closest they come to each other is .5 inches. Your program might be reporting the distance they come to each other which is 0.
Dan Miel

prokoudine:
@DanMiel I am commenting on the code submitted in this pull request, naturally :)

maxwxyz:
The unified measurement tool is merged now, could this PR be completed now?

PaddleStroke:
Rebased on main after the Unified measurement tool merged.
So I'll mark this PR as ready now.

maxwxyz:
I could see a slowdown when changing selection in big documents with complex geometry if always the volume/area is calculated (just selecting/organizing items in the tree view). Maybe selection in the tree view could be excluded?

PaddleStroke:
I think the problem is that the function that gets the volume/area from the shape is a bit slow because its computing them everytime its called.

This also causes some annoying issues in assembly and MBD as well. I'll open an issue about this.

maxwxyz:
@PaddleStroke could we comment out or deactivate COM and Volume for now? Most of the time when selecting objects in the tree view the user is not interested in these measures. When it's needed, the new measurement tool can provide them. COM would also be more relevant if a coordinate system or datum point could be created at it's location.

PaddleStroke:
Yes I think we are going to disable area / volume for solids because there are no obvious way to solve this. And as you say volume/area are not so often needed. When it is the UMT can be used.

PaddleStroke:
@maxwxyz can you please test and feedback if everything is fine now?

maxwxyz:
looks good!

0penBrain:
Looks to me as this code makes no use or almost of the new measurement framework. Am I correct ?

0penBrain:
> Well, the framework was merged to main only last week. The Quick Measure feature developed over main is months old.

Framework dev is available as a PR for months so doesn't look as a sensible argument.

prokoudine:
Depending on unmerged patch sounds risky, but now that both are in main, code can be deduplicated

DrD4ffy:
Why are square millimeters written as mm^2 and not as mm²?

0penBrain:
> Depending on unmerged patch sounds risky

Depending on self-coded things is even more. Trust me on this. :smile:

PaddleStroke:
I based QuickMeasure on the pre-existing Measurement class and extended it.
At that time I had not seen the UMT work which was a draft PR.

It seems UMT is introducing measurement document objects for each type of measurement. Quickmeasure does not need to create objects. So it can hardly reuse UMT framework.

However it stands that there is duplication of the functions calculating the values (length, area...). But then I would say they do not belong to the measure document objects, as one (such as quickmeasure) can need to calculate measures without needing to create document objects. So those would belong to the pre-existing measurement class, where quickMeasure implemented them.

0penBrain:
> It seems UMT is introducing measurement document objects for each type of measurement.

I don't think so. When you use the tool, the result is displayed in the Task panel without any object created. 😉

wwmayer:
The tool is causing some severe problems: https://forum.freecad.org/viewtopic.php?t=87511
I am working on a PR...

wwmayer:
https://github.com/FreeCAD/FreeCAD/pull/13944

PR #12219, opened 2024-01-31
Part & PartDesign Cone: allow equal radii

### LINKED ISSUE #12216: [Problem] identic radii on cones are disallowed, which might create problems on parametric CAD models

### Problem description

Note that I do not know if this is the correct location where this throws the exception.
It was just where I found a trace of cone radii being compared:

https://github.com/FreeCAD/FreeCAD/blob/e6004ff6c93b41637484b909dedc786466553a09/src/Mod/PartDesign/App/FeaturePrimitive.cpp#L357-L358

I have a parametric CAD model with two axis which are interconnected by a cone.
If both axis become the same diameter (by purpose) the cone gets two identic radii on both sides
and the model recomputation fails (for no real reason in that case, as everything still is valid).

A workaround is to add 1 µm to one of the axis, this helps, but is a very sad solution.

Please either make it a warning or add something like a checkbox to allow identic radii,
such that cones can become a cylinder, just in case.

Thanks for providing FreeCAD

### Full version info

```shell
[code]
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.21.2.33771 (Git)
Build type: Release
Branch: releases/FreeCAD-0-21
Hash: b9bfa5c5507506e4515816414cd27f4851d00489
Installed mods: 
  * Help 1.0.3
[/code]
```

### Subproject(s) affected?

PartDesign

### Anything else?

Sorry, I do not think I find the time to dig deep enough in the codebase to create
a proper patch myself.  And I hope it is not really much effort to change it.

### Code of Conduct

FlachyJoe:
Fix #12216


Roy-043:
Part_Cone also has this issue. It would be a good idea to solve that as well if possible.

FlachyJoe:
@Roy-043 I'll take a look soon.

madpipeline:
Is there a timeline when this will be released?

PR #12225, opened 2024-02-01
Resolve #11965 (B) no proper execute() for cross-property references

### LINKED ISSUE #11965: [Problem] no proper `PropertyExpressionEngine.execute()` for cross-property references

Andrei-Pozolotin:
Solution B: `parsequant()` expression language function

### DIFF
--- a/tests/src/App/ExpressionParser.cpp
+++ b/tests/src/App/ExpressionParser.cpp
@@ -1,0 +1,105 @@
+#include "gtest/gtest.h"
+
+#include "Base/Quantity.h"
+
+#include "App/Application.h"
+#include "App/Document.h"
+#include "App/DocumentObject.h"
+#include "App/Expression.h"
+#include "App/ExpressionParser.h"
+
+#include "src/App/InitApplication.h"
+
+// clang-format off
+
+class ExpressionParserTest: public ::testing::Test
+{
+protected:
+    static void SetUpTestSuite()
+    {
+        tests::initApplication();
+    }
+
+    void SetUp() override
+    {
+        _doc_name = App::GetApplication().getUniqueDocumentName("test");
+        _this_doc = App::GetApplication().newDocument(_doc_name.c_str(), "testUser");
+        _this_obj = _this_doc -> addObject("Sketcher::SketchObject");
+    }
+
+    void TearDown() override
+    {
+        App::GetApplication().closeDocument(_doc_name.c_str());
+    }
+
+    std::string doc_name() { return _doc_name; }
+    App::Document* this_doc() { return _this_doc; }
+    App::DocumentObject* this_obj() { return _this_obj; }
+
+    Base::Quantity parse_expression_text_as_quantity(const char* expression_text) {
+        auto expression = App::ExpressionParser::parse(this_obj(), expression_text);
+        auto expression_value = expression->getValueAsAny();
+        auto quantity_result = App::any_cast<Base::Quantity>(expression_value);
+        return quantity_result;
+    }
+
+    Base::Quantity parse_quantity_text_as_quantity(const char* quantity_text) {
+        auto quantity_qstr = QString::fromStdString(std::string(quantity_text));
+        auto quantity_result = Base::Quantity::parse(quantity_qstr);
+        return quantity_result;
+    }
+
+private:
+    std::string _doc_name;
+    App::Document* _this_doc {};
+    App::DocumentObject* _this_obj {};
+};
+
+// https://github.com/FreeCAD/FreeCAD/issues/11965
+TEST_F(ExpressionParserTest, functionPARSEQUANT)
+{
+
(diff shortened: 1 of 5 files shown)


pieterhijma:
@Andrei-Pozolotin, @chennes, sorry, it's on my todo list for a while. I will review today or tomorrow.

pieterhijma (changes requested):
@Andrei-Pozolotin, Initially I thought you wanted a full-fledged parser for general expressions, but a parse function just for quantities makes much more sense. The test shows that it works with a variable expression which is great, but I would like to see a couple of more direct / reflexive tests as well.

I would suggest:
- `parsequant(<<(1 + 2) m>>)` hoping it evaluates to `3000 mm` 
- `parsequant(str(1 m + 2 mm))` hoping it evaluates to `1002 mm`

For the former I would expect it to create a string `"(1 + 2) mm"` that should be a valid quantity expression, so it should be picked up by the quantity parser. For the latter I would expect the expression `1m + 2mm` to evaluate to `1002 mm`. This is then converted to a string that should be parsed by `parsequant` evaluating to `1002 mm`.

If this works as I expect on the rebased main branch that has the former PR [reverted](https://github.com/FreeCAD/FreeCAD/pull/12360), then I see no barriers to merge this. Great work from my point of view.

@chennes, hope this helps to make a decision.

Andrei-Pozolotin:
@pieterhijma : Pieter: thank you for the feedback.

Status:
* tests added
* branch rebased

Please review again.

tests/src/App/ExpressionParser.cpp
```diff
@@ -0,0 +1,106 @@
...
+    void TearDown() override
+    {
+        App::GetApplication().closeDocument(_doc_name.c_str());
+    }
+
+    std::string doc_name() { return _doc_name; }
+    App::Document* this_doc() { return _this_doc; }
+    App::DocumentObject* this_obj() { return _this_obj; }
+
+    Base::Quantity parse_expra_as_quant(const char* expra_text) {
```
pieterhijma, line 39: Just a very small comment: why not use `expr` instead of `expra` anywhere in this file. (I'm trying to figure out what an `expra` could mean. I probably simply miss it, but to me the `a` suffix currently doesn't mean anything, so it confuses me.)

pieterhijma (approved):
Looks great overall. Also nice that a couple of different units were tested. From my point of view, I approve.

Andrei-Pozolotin:
@pieterhijma : naming patterns changed, please review.

pieterhijma (approved):
Looks good, thanks.

Andrei-Pozolotin:
@chennes : @pieterhijma : works as intended, thank you.

PR #12229, opened 2024-02-01
Transfer in Geometry.x changes from RT branch

bgbsww:
Existing methods appear to be tested by the existing python suite.  This passes all python and c++ tests.

### DIFF
--- a/src/Mod/Part/App/Geometry.cpp
+++ b/src/Mod/Part/App/Geometry.cpp
@@ -63,10 +63,6 @@
 # include <Geom_SurfaceOfRevolution.hxx>
 # include <Geom_ToroidalSurface.hxx>
 # include <Geom_TrimmedCurve.hxx>
-# if OCC_VERSION_HEX < 0x070600
-#   include <GeomAdaptor_HSurface.hxx>
-#   include <GeomAdaptor_HCurve.hxx>
-# endif
 # include <GeomAPI_ExtremaCurveCurve.hxx>
 # include <GeomAPI_Interpolate.hxx>
 # include <GeomAPI_ProjectPointOnCurve.hxx>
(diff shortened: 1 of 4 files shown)


src/Mod/Part/App/Geometry.cpp
```diff
@@ -61,14 +63,20 @@
 # include <Geom_SurfaceOfRevolution.hxx>
 # include <Geom_ToroidalSurface.hxx>
 # include <Geom_TrimmedCurve.hxx>
+# if OCC_VERSION_HEX < 0x070600
+#   include <GeomAdaptor_HSurface.hxx>
+#   include <GeomAdaptor_HCurve.hxx>
```
chennes, line 68: This one already appears below

src/Mod/Part/App/Geometry.h
```diff
@@ -153,7 +158,7 @@ class PartExport GeomPoint : public Geometry
     void Restore(Base::XMLReader &/*reader*/) override;
     // Base implementer ----------------------------
     PyObject *getPyObject() override;
-
+    bool isSame(const Geometry &other, double tol, double atol) const override;
```
chennes, line 161: Can you add a docstring about the two doubles?

src/Mod/Part/App/Geometry.h
```diff
@@ -207,8 +218,11 @@ class PartExport GeomCurve : public Geometry
 
     Base::Vector3d value(double u) const;
 
+    GeomLine *toLine(bool clone=true) const;
+    GeomLineSegment *toLineSegment(bool clone=true) const;
```
chennes, line 222: You'll no doubt be shocked to hear that I don't like leaving this as a boolean.

bgbsww: Laugh!  Totally agree with this one.  The lack of meaning clarity is brutal.

PR #12237, opened 2024-02-02
Toponaming: Transfer in makeElementRefine

bgbsww:
~~This should pass CI as soon as #12229 is merged.~~  And OCCT7.8.0 patches are added, and testing cleaned up.

### DIFF
--- a/src/Mod/Part/App/TopoShapeMapper.cpp
+++ b/src/Mod/Part/App/TopoShapeMapper.cpp
@@ -1,6 +1,10 @@
+#include <TopoDS.hxx>
+#include <BRep_Tool.hxx>
+#include <TopoDS_Edge.hxx>
 #include "PreCompiled.h"
 
 #include "TopoShapeMapper.h"
+#include "Geometry.h"
 
 namespace Part
 {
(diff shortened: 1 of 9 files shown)


src/Mod/Part/App/TopoShapeMapper.cpp
```diff
@@ -1,6 +1,31 @@
-#include "PreCompiled.h"
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/****************************************************************************
+ *                                                                          *
+ *   Copyright (c) 2002 Jürgen Riegel <juergen.riegel@web.de>               *
```
chennes, line 4: Is this the right copyright here?

bgbsww: I think so.  All the code here came from the file with this ( TopoShape.cpp ) ; this is just moving to a separate file.  I'm guessing at the right protocol here.  This was intentional, but maybe I'm wrong?

chennes: It's fine, I just didn't want to mis-attribute if this was a copy-and-paste error.

PR #12243, opened 2024-02-03
App: fix PropertyPythonObject persistence backward compatibility

realthunder:
Related #10460

Amendment to 83d4080f.

Check for attribute `__func__` for Python >= 3.11 to avoid calling built-in `__getstate__/__setstate__`.


wwmayer:
`object().__getstate__` is of type `builtin_function_or_method` (which I have never heard of before) while the method `__getstate__` of a Python feature class is of type `method`. The latter has the attribute `__func__` while the former hasn't. 

So, apparently addon maintainers don't have to do anything.

PR #12249, opened 2024-02-03
Ensure that resetElementMap results in an element map existing, unless overridden

bgbsww:


chennes:
I don't love the trailing boolean here -- can we make a policy enum instead?

bgbsww:
Sure.  There are 5 other bool-that-could-be-enum parms scattered throughout this object API, so maybe this will trigger someone to adjust the rest.

PR #12256, opened 2024-02-04
Release GIL while meshing in MeshPart.meshFromShape (2nd proposal)

howetuft:
This commit enables multithreading use when exporting large quantities of objects (for instance for external rendering).
GIL release is only applied to mesher call, which is pure C / no Python, and only if explicitely required via a new parameter.
In case this new parameter is not set, the previous behavior remains (hold GIL), thus ensuring ascendant compatibility.

This PR replaces https://github.com/FreeCAD/FreeCAD/pull/12241 (@marioalexis84)

### DIFF
--- a/src/Mod/MeshPart/App/AppMeshPartPy.cpp
+++ b/src/Mod/MeshPart/App/AppMeshPartPy.cpp
@@ -474,6 +474,15 @@ private:
     {
         PyObject *shape;
 
+        auto runMesher = [](const MeshPart::Mesher& mesher) {
+            Mesh::MeshObject* mesh;
+            {
+                Base::PyGILStateRelease releaser{};
+                mesh = mesher.createMesh();
+            }
+            return Py::asObject(new Mesh::MeshPy(mesh));
+        };
+
         static const std::array<const char *, 7> kwds_lindeflection{"Shape", "LinearDeflection", "AngularDeflection",
                                                                     "Relative", "Segments", "GroupColors", nullptr};
         PyErr_Clear();
@@ -509,7 +518,7 @@ private:
                 }
                 mesher.setColors(colors);
             }
-            return Py::asObject(new Mesh::MeshPy(mesher.createMesh()));
+            return runMesher(mesher);
         }
 
         static const std::array<const char *, 3> kwds_maxLength{"Shape", "MaxLength", nullptr};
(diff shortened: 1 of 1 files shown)


marioalexis84:
One more comment (sorry for not putting them all together at the beginning of the review):
I see no reason to add the `releaseGIL` parameter only for the first parameter variant of the function: It can be added for the rest of the variants.
To avoid code duplication, it is convenient to use a lambda function to check the value of the `releaseGIL` parameter and return the created `MeshPy` object.

howetuft:
> It can be added for the rest of the variants.

Here you are, with the requested lambda.

> sorry for not putting them all together at the beginning of the review

No problem. You can keep pointing out one issue at a time - I may just conclude you might be single-threaded too ;-)

howetuft:
Hello @marioalexis84,
Just in case you missed my last commit... Thank you for your feedback!

howetuft:
@chennes Hello! I saw that you had recently committed to MeshPart, would there be any way you could review my PR please? Many thanks in advance...

chennes:
@howetuft probably @wwmayer is a more appropriate reviewer than I am here, I've asked him to take a look at this. We'll also review it in tomorrow's PR Review meeting (sorry for the delay, most of the Maintainers were traveling the last two weeks!).

marioalexis84 (commented):
@howetuft  It looks good to me. (although I needed to test the Netgen-dependent signatures).
Not related to your PR, but the signatures need some rework. For example the last signature that only receives a Part::TopoShapePy is never called.

wwmayer (commented):
Do we really need this parameter _ReleaseGIL_? I don't see a problem if the GIL is always released before calling createMesh().

howetuft:
> Do we really need this parameter ReleaseGIL? I don't see a problem if the GIL is always released before calling createMesh().

No we don't. I had introduced this parameter "just in case" someone would prefer keeping the previous behaviour, but of course releasing the GIL wherever it's possible will always be better for API user.

I will push a new version without this param.

howetuft:
@wwmayer 
Hello,
It's done: I removed the `ReleaseGIL` parameter, the code is much simpler, by the way. Thank you in advance for the merge, or feel free to ask me for any other alterations.

wwmayer:
Thanks, it's all fine.

howetuft:
Thanks to you!

PR #12262, opened 2024-02-04
Material: Material editor enhancements

davesrocketshop:
Improves compatibility when using the legacy editor.

Corrects a number of issues that prevented saved files from being displayed when using the legacy editor.

Reduces the verbosity of console messages.


davesrocketshop:
See https://forum.freecad.org/viewtopic.php?p=734704&sid=c111a93fb685fdf2b9bfa3eeb22f29e2#p734704

PR #12270, opened 2024-02-05
[Core] Introduce Tab-Bar workbench selector

PaddleStroke:
Fixes https://github.com/FreeCAD/FreeCAD/issues/8692
Introduce Tab-Bar workbench selector (inspired by triplus' addon)
image

Can go on any side (north/south/west/east) : 
image

Preferences : 
image

You can configure text / icon or both. (for combobox too btw)
Can be put on the corner widget same as the combobox:
image

**The default is not changed : ComboBox as a toolbar with text & icons.**

### DIFF
--- a/src/Gui/Action.cpp
+++ b/src/Gui/Action.cpp
@@ -60,6 +60,7 @@
 #include "Widgets.h"
 #include "Workbench.h"
 #include "WorkbenchManager.h"
+#include "WorkbenchSelector.h"
 #include "ShortcutManager.h"
 #include "Tools.h"
 
@@ -614,43 +615,6 @@ void ActionGroup::onHovered (QAction *act)
 }
 
 
-// --------------------------------------------------------------------
-
-WorkbenchComboBox::WorkbenchComboBox(QWidget* parent) : QComboBox(parent)
-{
-}
-
-void WorkbenchComboBox::showPopup()
-{
-    int rows = count();
-    if (rows > 0) {
-        int height = view()->sizeHintForRow(0);
-        int maxHeight = QApplication::primaryScreen()->size().height();
-        view()->setMinimumHeight(qMin(height * rows, maxHeight/2));
-    }
-
-    QComboBox::showPopup();
-}
-
-void WorkbenchComboBox::refreshList(QList<QAction*> actionList)
-{
-    clear();
-
-    for (QAction* action : actionList) {
-        QIcon icon = action->icon();
-        if (icon.isNull()) {
-            this->addItem(action->text());
-        }
-        else {
-            this->addItem(icon, action->text());
-        }
-
-        if (action->isChecked()) {
-            this->setCurrentIndex(this->count() - 1);
-        }
-    }
-}
-
 /* TRANSLATOR Gui::WorkbenchGroup */
 WorkbenchGroup::WorkbenchGroup (  Command* pcCmd, QObject * parent )
   : ActionGroup( pcCmd, parent )
@@ -667,37 +631,29 @@ WorkbenchGroup::WorkbenchGroup (  Command* pcCmd, QObject * parent )
 
 void WorkbenchGroup::addTo(QWidget *widget)
 {
-    auto setupBox = [&](WorkbenchComboBox* box) {
-        box->setIconSize(QSize(16, 16));
-        box->setToolTip(toolTip());
-        box->setStatusTip(action()->statusTip());
-        box->setWhatsThis(action()->whatsThis());
-        box->refreshList(actions());
-        connect(this, &WorkbenchGroup::workbenchListRefreshed, box, &WorkbenchComboBox::refreshList);
-        connect(groupAction(), &QActionGroup::triggered, box, [this, box](QAction* action) {
-            box->setCurrentIndex(actions().indexOf(action));
-        });
-        connect(box, qOverload<int>(&WorkbenchComboBox::activated), this, [this](int index) {
-            actions()[index]->trigger();
-        });
-    };
-    if (widget->inherits("QToolBar")) {
-        auto* box = new WorkbenchComboBox(widget);
-        setupBox(box);
+    if (widget->inherits("QToolBar") || widget->inherits("QMenuBar")) {
+        ParameterGrp::handle hGrp;
+        hGrp = App::GetApplication().GetParameterGroupByPath("User parameter:BaseApp/Preferences/Workbenches");
+        QWidget* wbSel;
+        if (hGrp->GetInt("WorkbenchSelectorType", 0) == 0) {
+            wbSel = new WorkbenchComboBox(this, widget);
+        }
+        else {
+            wbSel = new WorkbenchTabWidget(this, widget);
+        }
 
-        qobject_cast<QToolBar*>(widget)->addWidget(box);
-    }
-    else if (widget->inherits("QMenuBar")) {
-        auto* box = new WorkbenchComboBox(widget);
-        setupBox(box);
-
-        bool left = WorkbenchSwitcher::isLeftCorner(WorkbenchSwitcher::getValue());
-        qobject_cast<QMenuBar*>(widget)->setCornerWidget(box, left ? Qt::TopLeftCorner : Qt::TopRightCorner);
+        if (widget->inherits("QToolBar")) {
+            qobject_cast<QToolBar*>(widget)->addWidget(wbSel);
+        }
+        else {
+            bool left = WorkbenchSwitcher::isLeftCorner(WorkbenchSwitcher::getValue());
+            qobject_cast<QMenuBar*>(widget)->setCornerWidget(wbSel, left ? Qt::TopLeftCorner : Qt::TopRightCorner);
+        }
     }
     else if (widget->inherits("QMenu")) {
         auto menu = qobject_cast<QMenu*>(widget);
         menu = menu->addMenu(action()->text());
-        menu->addActions(actions());
+        menu->addActions(getEnabledWbActions());
 
         connect(this, &WorkbenchGroup::workbenchListRefreshed, this, [menu](QList<QAction*> actions) {
             menu->clear();
@@ -715,11 +671,10 @@ void WorkbenchGroup::refreshWorkbenchList()
         groupAction()->removeAction(action);
         delete action;
     }
+    enabledWbsActions.clear();
+    disabledWbsActions.clear();
 
-    std::string activeWbName = "";
-    Workbench* activeWB = WorkbenchManager::instance()->active();
-    if (activeWB)
-        activeWbName = activeWB->name();
+    std::string activeWbName = WorkbenchManager::instance()->activeName();
 
     // Create action list of enabled wb
     int index = 0;
@@ -741,12 +696,33 @@ void WorkbenchGroup::refreshWorkbenchList()
         if (wbName.toStdString() == activeWbName) {
             action->setChecked(true);
         }
+        enabledWbsActions.push_back(action);
+        index++;
+    }
 
+    // Also create action list of disabled wbs
+    QStringList disabled_wbs_list = DlgSettingsWorkbenchesImp::getDisabledWorkbenches();
+    for (const auto& wbName : disabled_wbs_list) {
+        QString name = Application::Instance->workbenchMenuText(wbName);
+        QPixmap px = Application::Instance->workbenchIcon(wbName);
+        QString tip = Application::Instance->workbenchToolTip(wbName);
+
+        QAction* action = groupAction()->addAction(name);
+        action->setCheckable(true);
+        action->setData(QVariant(index)); // set the index
+        action->setObjectName(wbName);
+        action->setIcon(px);
+        action->setToolTip(tip);
+        action->setStatusTip(tr("Select the '%1' workbench").arg(name));
+        if (wbName.toStdString() == activeWbName) {
+            action->setChecked(true);
+        }
+        disabledWbsActions.push_back(action);
         index++;
     }
 
     // Signal to the widgets (WorkbenchComboBox & menu) to update the wb list
-    workbenchListRefreshed(actions());
+    workbenchListRefreshed(enabledWbsActions);
 }
 
 void WorkbenchGroup::onWorkbenchActivated(const QString& name)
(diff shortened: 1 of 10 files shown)


qewer33:
Looks great! A couple notes from me:

- The "More" tab would be better as a button instead of a tab

- IMHO the tabbar should only display loaded (instead of enabled) workbenches and the "More" button should load unloaded workbenches. The user can check the autoload button on workbenches they want loaded at startup. This approach doesn't cost the user any more clicks than the dropdown selector and in most cases it would be a cleaner UI since less WBs will be shown in the tabbar. This can be discussed ofc, just my 2 cents.

kadet1090:
Related issue: https://github.com/FreeCAD/FreeCAD/issues/8692

In the comment: https://github.com/FreeCAD/FreeCAD/issues/8692#issuecomment-1768565884 I summarized how it probably should work - it is in line with what Qewer says. It would be nice to somehow adhere to what we established in the issue I think.

PaddleStroke:
I initially made the '+' a button and had to roll that back to as it is now. The reasons are multiple : 
- It looked bad.
- QTabBar always has a 'selected' tab. So if you switch to a workbench that is not in your shortlist, then the tab of the previous workbench stayed as it is. Though I see that Triplus addon in that case is adding an additional temporary tab for the current workbench. But Triplus's addon does not show all the other workbench in a list under the + where they get checked.
image
I invite you to try it out because I have been playing with both during this dev and this one feels really much better.

Now regarding the 'show only loaded' paradigm. I see several issues : 
- Boot time of the application will be significantly longer, as you initially have to load all the workbenches. So if a user wants to use tab-bar with a lot of workbenches it is a show stopper.
- The tabbar would grow everytime you visit a 'little used' workbench. Making its size dynamic and so harder to handle for the user.
- This makes effectively 2 different ways to set workbenches that are shown. One for the combobox where you put the enabled wb. And for the tabbar where you put the auto-load. And yes its not exactly the same because for tabbar you then don't show the 'disabled' wb. But this is still duplicating complexity and ultimately increasing confusion imo.
- Regarding the additional features your describe where the loaded list depends on the document and changing that based on the icon the user clicked in the start page : While it sounds like a good idea, I think that this is going to effectively be weird, inconsistent and confusing for the user. They'll be like 'why I don't have assembly button I had it yesterday???'. + the issue that the tabbar would resize differently depending on document...
- Per-document loaded wb would probably be a lot more work.

kadet1090:
The plus button should not be a tab, but a button next to a tab. Just like web browsers are doing it then it should look fine. Using tab as button is not a good idea as you will constantly fight with problems as tabs were not designed to pose role of a button.

About loading workbenches - yes that is a problem that's why in the next comments we've concluded that the notion of "Favorite" workbenches should be Independent. Favourite workbenches would be always visible which makes sense. Other ones would show on demand. This could be configurable but there should be a way to achieve that.

For example, I rarely use the arch workbench but sometimes need to do that so it is hidden under plus sign. With the add-on it is frustrating as it is the only workbenches that I need to access via drop-down, which is extra frustrating as of I use a workbench once in a session I probably will use it multiple times. The growth of tabbar is not a big deal as in the next session it will be hidden anyway and I should have option to hide a tab explicitly.

Zolko-123:
@yorikvanhavre  and @wwmayer  : honestly, what is the objective of including into core FreeCAD existing functionality provided by external optional Python modules with the exact same C++ compiled code ? 

@sliptonic : if your company is so much swimming in money as to pay people for duplicating existing code, why not spend that money on something useful ?

Zolko-123:
I don't care how you waste your money, what I care about is the amount of useless crap that you ram down the throat of every FreeCAD user ... of which I'm (still) today

wwmayer:
> honestly, what is the objective of including into core FreeCAD existing functionality provided by external optional Python modules with the exact same C++ compiled code ?

I am not a big friend of such endeavours -- especially if it means that we lose flexibility and everybody is forced to use the new stuff.

PaddleStroke:
@wwmayer we do not loose flexibility. It is introduced as an option so we gain more flexibility.
As to why integrating it in core there are many reasons. Discoverability of addons is poor. Quality and consistency is not ideal.
For example triplus tabbar addon reimplement a new dialog 'preference' + logic for selecting displayed and ordering workbenches, which is redundant and unconnected to current capabilities of preferences.
Also triplus addon does not let you trigger workbenches that are not in your list.

wwmayer:
> we do not lose flexibility.

If everybody is still free to use the preferred selector then I don't mind.

MisterMakerNL:
I agree with @PaddleStroke I have been using tabbar too, but the addon feels like a forgotten core function.
I hope @triplus don't mind that we think his/her work is so good that it needs to be part of the core?
I would like to test this PR but when I check these Ondsel branch or such out on Github desktop it is unable to find the PR.

chennes:
@MisterMakerNL if you are willing to use the command line, you can add a "remote" to your normal FreeCAD checkout, and then directly check out the branch you want to try, something like:
```
git remote add ondsel https://github.com/ondsel/FreeCAD
git fetch ondsel
git checkout tab_bar_wb
```

mathew500:
> @yorikvanhavre and @wwmayer : honestly, what is the objective of including into core FreeCAD existing functionality provided by external optional Python modules with the exact same C++ compiled code ?
> 
> @sliptonic : if your company is so much swimming in money as to pay people for duplicating existing code, why not spend that money on something useful ?
> 
> I've been using @triplus 's TabBar since years and it works perfectly well, nothing to add or remove. What's the point ???

Isn't it a hugh kudos that a addon or multiple addons (that serve the same purpose) gets recognized as a missing feature and inspires implementation into the core. (I hardly dare say it, however, the assembly story narrates us the same experience). I think that proves great value of the addon eco system.

Zolko-123:
@PaddleStroke :

> As to why integrating it in core there are many reasons

what part of : "_[re-implement] Python modules with the exact same C++ compiled code_" don't you understand ? 

@mathew500 :

> Isn't it a huge kudos that a addon or multiple addons (that serve the same purpose) gets recognized as a missing feature and inspires implementation into the core.

no it's not. What would be a huge kudo if the code were actually **INCLUDED** into the core and retaining the **ORIGINAL COPYRIGHT** . Let's see whether you spot the difference:

Copyright (C) 2015, 2016, 2017, 2018 triplus @ FreeCAD
–vs–
Copyright (c) 2024 Pierre-Louis Boyer <development[at]Ondsel.com>

@wwmayer :

> I am not a big friend of such endeavours

thank-you for the confirmation. Then why do you tolerate such behavior ?

PaddleStroke:
UI concepts cannot be copyright. Code can. But there is not a single line
of code that comes from Triplus addon. You make statements but you did not
even read the code. This is not a translation of python code.
I think it is time you stop harassing people... At least if you were right.
But sadly you're not and you're just embarrassing yourself.

PaddleStroke:
On top of that I'm totally fine modifying the copyright to acknowledge
Triplus. I just didn't paid attention to it.

wwmayer:
> Isn't it a huge kudos that a addon or multiple addons (that serve the same purpose) gets recognized as a missing feature and inspires implementation into the core.

This really depends. It can make sense for very important functionality of an orphaned add-on but it shouldn't become the rule as otherwise the core system will be bloated with more and more code that in the long run won't be properly maintained any more. And it may cause harm to the add-on ecosystem if the core starts to integrate already existing functions that competes with add-ons. Their authors may lose interest to further maintain their code.

About the Tabbar add-on isn't it amazing to see what is already possible only by using FreeCAD's Python API?


> Then why do you tolerate such behavior ?

I think you over-estimate my power. I can share my opinion but I cannot force anybody to do or not to do something.

kadet1090:
The product should be targeted for wide range of users users and as I said on the forum - most of users do not touch default settings including addons. There is no problem in keeping functions like designing guitars as addon because it is a niche, but I should not be forced to use addons to have good experience with the software and in my opinion and many others current workbench selector does not provide a good user experience. 

This PR is great addition to out of box experience - I may have my remarks about it but it is still a huge step forwards.

Zolko-123:
> most of users do not touch default settings including addons

do you have any evidence or proof of that ?

DrD4ffy:
Some thoughts on the current discussion:

- I'm perfectly capable of changing default settings when necessary, but still appreciate it when it's not necessary in the first place.
- I think kadet1090's example is pretty good: GuitarWB is an add-on and not a core functionality, whereas changing WB is definitely a core function.
- This PR seems to me to integrate a quite requested way of changing WB in an improved version as a core function in FreeCAD.

Conclusion: Would like to see this improvement come

prokoudine:
> do you have any evidence or proof of that ?

It's the kind of insight people get after spending many years in tech support: users rarely change defaults and get really surprised when they find out they can.

falsagoff:
This thread is a bit tense and as an outsider perhaps I may comment unbiasedly. Despite not knowing what new feature/functionality this patch is about, I just thought it must be a positive thing for FC. Reading Zolko-123 comments, I thought here we go again cause I have read his posts in other threads in the past. I reacted not to the technical aspect of the issue but I got emotional because he was so confrontational, I just got defensive despite not having anything to do with FC development. Then I calmed down and looked into the matter objectively.  
Anyway, that is what makes FC such a great project, everything is discussed openly, transparently and freely. So to PaddleStroke: keep up the wonderful work, and to Zolko-123: keep up the needed scrutiny and checks. . Fight but don't kill each other (not literally, you know what I mean).
I think since I first started using FC at v0.17, I have never been so excited of an upcoming version. Can we have it tomorrow already.

Zolko-123:
> > do you have any evidence or proof of that ?
> 
> It's the kind of insight people get after spending many years in tech support

meaning: "no". Don't feed the troll, I know, but sometimes it's difficult to resist

adrianinsaval:
Discussion on the merits of the feature proposal should happen on the issue. If you have technical complaints on the code go ahead and discuss that, otherwise it's essentially spam.

Zolko-123:
> Discussion on the merits of the feature proposal should happen on the issue. If you have technical complaints on the code go ahead and discuss that, otherwise it's essentially spam.

:hankey:

sliptonic:
> > > do you have any evidence or proof of that ?
> > 
> > 
> > It's the kind of insight people get after spending many years in tech support
> 
> meaning: "no". Don't feed the troll, I know, but sometimes it's difficult to resist

    "The Effect of Default Rules on Choice Making: Selected Examples from Pluralistic Domains," by Mary Steffel, Elanor F. Williams, and Ruth Pogacar, published in "Perspectives on Psychological Science" (2016). This review highlights how default options significantly influence decisions in various contexts, including software settings, demonstrating that people often stick with provided defaults.

    "Privacy by Default: An Empirical Study on the Effectiveness of Privacy Settings in Social Networking Sites," by Joseph Bonneau and Sören Preibusch, presented at the "Fifth International AAAI Conference on Weblogs and Social Media" (2011). This study investigates the privacy settings in social networking sites, finding that defaults have a substantial impact on user privacy choices, with many users sticking to default privacy settings.

    "Do Defaults Save Lives?" by Eric J. Johnson and Daniel G. Goldstein, published in "Science" (2003). Although focused on organ donation, this study is frequently cited in discussions about the power of defaults. It shows that default choices have a significant influence on decisions in life-critical contexts, suggesting parallels in how defaults might affect choices in software settings.

    "Smart Defaults: From Hidden Persuaders to Adaptive Helpers," by Marko Terviö and Joshua Klayman, in the "International Journal of Human-Computer Studies" (2004). This paper discusses how smart default settings in software can aid user decision-making, emphasizing that users often rely on defaults when they are presented as the recommended or optimal choice.

xtemp09:
Activation of a workbench requires a bit of time. Is it possible to preload workbenches to have ultramegasuperfast activation? If not, is there a way to speed up workbench selection?

adrianinsaval:
Users can choose which WB are preloaded at start on the preference page, but this means FreeCAD will be slower to start.

xtemp09:
@adrianinsaval, do you mean Auto-load checkboxes in Preferences → Workbenches → Available Workbenches?

adrianinsaval:
Yes

xtemp09:
Sorry, I studied English with textbooks and vocabularies. I have never thought that _auto-load_ is synonymous with _preload_.


Cambridge dictionary says:

pre- means: [^1]

> before (a time or an event)

auto- means: [^2]

> of or by yourself, or operating independently and without needing help

I thought auto-load is loading a workbench options in preferences so user can customize the workbench, even if it has not been activated yet.

[^1]: https://dictionary.cambridge.org/dictionary/english/pre?q=pre-
[^2]: https://dictionary.cambridge.org/dictionary/english/auto?q=auto-

Zolko-123:
@sliptonic : you get your insights about how people use FreeCAD by reading books about unrelated subjects ???

Did you think about looking through the hundreds or thousands of user-supplied data about .... what Addons they actually have installed !!!

```
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.21.2.33771 (Git)
Build type: Release
Branch: releases/FreeCAD-0-21
Hash: b9bfa5c5507506e4515816414cd27f4851d00489
Installed mods: 
  * A2plus 0.4.60n
  * sheetmetal 0.3.1
```

from what I see there are very few users without **any** addon at all.

prokoudine:
@Zolko-123 

40% of users in our survey depend on addons. The vast majority of them don't depend on any addons that customize the UI or any default settings (there aren't many such addons anyway).

Neither A2plus not SheetMetal in your example customize the UI or any other default settings.

PaddleStroke:
Yes it should be possible

prokoudine:
Some further feedback.

1. Vertically oriented TabBar is too wide, which looks especially strange when you use 'Icons' style rather than 'Icons + text'.

2. While the orientation i svertical, you don't get the same scroll buttons you get when the TabBar is horizontal, so you can't see workbench tabs that don't fit the display.

3. Trying to make vertical TabBar horizontal again doesn't work. Moreover, the grip area switches between the left side of the bar and it top. I can't figure ut why. Here's a video:

https://github.com/FreeCAD/FreeCAD/assets/57467/7df8e62e-2091-492e-a7ee-2e4f17efffa5

prokoudine:
Once I get that forever-floating TabBar form the video, I go to 'Edit > Preferences > Workspaces' and try to set its position via drop-down list. This is what I get after restarting: no main menu.

image

MisterMakerNL:
@prokoudine I think most of these issues are stylesheet issues. The tab has a margin on the right and left.
And when you move it vertical this margin is still on the right and left.
It should be switched with top an bottom.
Not sure how to implement such fix.

MisterMakerNL:
If we still want it in the title bar, I might have a solution here:
https://github.com/FreeCAD/FreeCAD/issues/12742

yorikvanhavre:
Will give it a test spin!

PaddleStroke:
(rebased on main to reduce compile time for tester)

PR #12286, opened 2024-02-07
Port to Py 3.12

wwmayer:


src/CXX/Python3/cxxextensions.c
```diff
@@ -42,13 +42,23 @@ extern "C"
 {
 #endif
 
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 13
```
chennes, line 45: Is the 13 here correct?

wwmayer: Yes, it is. I fetched the latest code changes from upstream: https://sourceforge.net/p/cxx/code/459/

PR #12133, opened 2024-01-26
Fem: Improve contact and tie constraints - fixes #11653

marioalexis84:
* Add new unit `StiffnessDensity` ($\text{Force/Lenght}^3 \equiv \text{Pressure/Lenght}$)  and the corresponding `App::PropertyStiffnessDensity`, basically a spring stiffness per unit area needed to use in contact stiffness and the stick slope value used by the `*FRICTION` keyword. (currrently it is said that the unit should be Pressure like the elastic modulus, but this is incorrect).
* Add support for `ADJUST` parameter on the `CONTACT PAIR` card  for contact constraint.
* Add support for specifying stick slope value in `*FRICTION` keyword.
* Add support for `ADJUST` parameter in tie constraint.
* Use quantity spin box in both constraint task dialogs.

@FEA-eng Do you want to test this?

Some links:
http://www.dhondt.de/ccx_2.21.pdf#subsubsection.6.7.5
http://www.dhondt.de/ccx_2.21.pdf#subsubsection.6.7.6
https://calculix.discourse.group/t/contact-stiffness-setting/455/2
https://www.mdpi.com/2075-4442/7/7/56

Issue #10864, opened 2023-09-29
Support for 2D mechanical analyses can be easily added to the FEM workbench


FEA-eng:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Currently, the FEM workbench doesn't support 2D mechanical analyses (plane stress, plane strain, axisymmetric). It seems to be rather easy to implement them using CalculiX solver for someone familiar with the source code. Pretty much only a single option in SolverCcxTools would be needed with the following settings and actions taken for each setting:

2D/3D Mechanical analysis:
- 3D (default) - does nothing
- 2D plane stress:
1. Changes the element type written to the input file (`*ELEMENT, TYPE=`): `S3` --> `CPS3`, `S6` --> `CPS6`, `S4` --> `CPS4`
2. Changes `*SHELL SECTION` to `*SOLID SECTION`
- 2D plane strain:
1. Changes the element type written to the input file (`*ELEMENT, TYPE=`): `S3` --> `CPE3`, `S6` --> `CPE6`, `S4` --> `CPE4`
2. Changes `*SHELL SECTION` to `*SOLID SECTION`
- 2D axisymmetric:
1. Changes the element type written to the input file (`*ELEMENT, TYPE=`): `S3` --> `CAX3`, `S6` --> `CAX6`, `S4` --> `CAX4`
2. Changes `*SHELL SECTION` to `*SOLID SECTION`

Related forum thread: https://forum.freecad.org/viewtopic.php?t=81462

### Full version info

```shell
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.34387 (Git)
Build type: Release
Branch: master
Hash: d95b26dfb5dd23eaee004b4af3cf5cf4766e7f9c
Python 3.10.12, Qt 5.15.8, Coin 4.0.0, Vtk 9.2.5, OCC 7.6.3
Locale: Polish/Poland (pl_PL)
```

### Subproject(s) affected?

FEM

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

FEA-eng:
One more thing to consider - it might be necessary to add a feature that would correct potential non-zero Z node coordinates: https://forum.freecad.org/viewtopic.php?t=81582#p710993

PR #12289, opened 2024-02-08
Fem: Fix restoring old Tie constraint and add support for cyclic symmetry - fixes #12155

### LINKED ISSUE #12155: [Feature Request] Support for cyclic symmetry in FEM

### Problem description

Cyclic symmetry available in CalculiX (and commonly used for structures like shafts and impellers) would be a nice addition to the FEM workbench. The definition is based on tie constraint and we have this feature already available in FreeCAD, so implementing cyclic symmetry shouldn't be too difficult. It's just a matter of adding the `CYCLIC SYMMETRY` parameter to *TIE and introducing a new keyword:
```
*CYCLIC SYMMETRY MODEL, N=number_of_sectors, NGRAPH=number_of_sectors_for_visualization
x_a, y_a, z_a, x_b, y_b, z_b
```

`a` and `b` are two points defining the axis of symmetry - their coordinates must be specified.

There's just one problem - boundary conditions can't be applied to the nodes of the slave surface of cyclic symmetry (slave surface in tie constraint definition) since that would cause an overconstraint. So in some cases, it might be necessary to remove the nodes lying on the edge between the face with the boundary condition and the slave surface of cyclic symmetry from the boundary condition definition. Unfortunately, FreeCAD doesn't operate on node sets and individual nodes can't be deselected so there would be this limitation for now.

Use case example: https://forum.freecad.org/viewtopic.php?t=76097#p735918

### Full version info

```shell
OS: Windows 10 build 19045
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35691 (Git)
Build type: Release
Branch: main
Hash: cc45b58dadae53d7a99be1ed7db25f6c93bb13d5
Locale: Polish/Poland (pl_PL)
```

### Subproject(s) affected?

FEM

### Anything else?

_No response_

### Code of Conduct

marioalexis84:
Add support for cyclic symmetry.
This also fixes the loading of the old Tie constraint object prior to #12133 since in that commit the existence check for the `Adjust` property was not added when restoring the object (currently old files are loaded without adding this property).
Normally I would have generated two pull requests, one for the correction and another to add the cyclic symmetry, but since some helper functions are added for use in creating properties and restoring the `DocumentObject`, I decided to put it together in a single commit.
The idea of the helper functions is to avoid code duplication and string comparison that usually occurs in `onDocumentRestored` when modifications are made to the properties.

The location of the axis of symmetry is obtained by applying the placement transformation to the `Z versor` (similar to what is done with the lines in `Part` module).

Still is needed add some view providers, options to task panel and automatically calculate the axis of symmetry from the planes defined by the surfaces that define the sector angle.

@FEA-eng

### DIFF
--- a/src/Mod/Fem/femobjects/base_fempythonobject.py
+++ b/src/Mod/Fem/femobjects/base_fempythonobject.py
@@ -47,3 +47,19 @@ class BaseFemPythonObject(object):
 
     def loads(self, state):
         return None
+
+
+class _PropHelper:
+    """
+    Helper class to manage property data inside proxy objects.
+    Initialization keywords are the same used with PropertyContainer
+    to add dynamics properties plus "value" for the initial value.
+    """
+    def __init__(self, **kwds):
+        self.value = kwds.pop("value")
+        self.info = kwds
+        self.name = kwds["name"]
+
+    def add_to_object(self, obj):
+        obj.addProperty(**self.info)
+        setattr(obj, self.name, self.value)
--- a/src/Mod/Fem/femobjects/constraint_tie.py
+++ b/src/Mod/Fem/femobjects/constraint_tie.py
@@ -31,6 +31,10 @@ __url__ = "https://www.freecad.org"
 
 from . import base_fempythonobject
 
+import FreeCAD
+
+_PropHelper = base_fempythonobject._PropHelper
+
 
 class ConstraintTie(base_fempythonobject.BaseFemPythonObject):
     """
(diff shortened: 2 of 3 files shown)


FEA-eng:
Thank you very much for implementing this. I tested it on Ubuntu and it's working properly. 

Btw. I hope that you don't mind if I ask but could you take a look at this issue too: #10864 ? As explained in the issue description, it seems to be easy to implement (for someone familiar with the source code so not for me, unfortunately). It's just a matter of adding one option to SolverCcxTools and internally replacing element names and SHELL with SOLID in the section definition (thickness is already there). Then ensuring zero Z coordinates for nodes. Just some input file swaps basically. Quite a few people asked for those analyses (especially axisymmetric) recently so it would be great to have them implemented.

@chennes Can you merge this PR ?

marioalexis84:
@FEA-eng the plan is to resolve all bugs/feature requests, #10864 included.

FEA-eng:
> @FEA-eng the plan is to resolve all bugs/feature requests, #10864 included.

That would be more than awesome. I’m so glad you’re doing so much for this workbench. I gave those issues some priority here: https://github.com/orgs/FreeCAD/projects/19 but of course it’s subjective and totally up to you. I chose #10864 as particularly interesting because it seems to be rather easy to implement while adding a lot of functionality to the FEM workbench.

PR #12291, opened 2024-02-08
Fix compilation failures in TopoShapeExpansion and TopoShapeMapper.

PaddleStroke:
Main does not build following #12237
Please be careful about precompiled includes @chennes @bgbsww

### DIFF
--- a/src/Mod/Part/App/TopoShapeExpansion.cpp
+++ b/src/Mod/Part/App/TopoShapeExpansion.cpp
@@ -41,7 +41,6 @@
 #include <utility>
 #endif
 
-#include "modelRefine.h"
 #include "TopoShape.h"
 #include "TopoShapeCache.h"
 #include "TopoShapeMapper.h"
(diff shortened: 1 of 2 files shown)


chennes:
The modelRefine.h fix is good. The mapper file needs the #ifndef around the OCCT headers. Thanks for the catch. @adrianinsaval is it possible to set one of our CI platforms to use precompiled headers?

adrianinsaval:
I have to experiment with it, I'm not sure if ccache can work with precompiled headers

bgbsww:
Yikes.  Many apologies.  ~~Looking for PCH info on FreeCAD has come up empty.  Appears to be the FREECAD_USE_PCH cmake option, I can certainly test build this locally if that's all it takes.  Unfortunately, I didn't see much change other that notable slowness and no breakage trying a build with that option in an Ubuntu environment; is this only testable in an MSVC setup?~~

 I found https://forum.freecad.org/viewtopic.php?t=70796

Should info on using/not using PCH be added to dev documentation somewhere?  Maybe the developer's handbook?  I can do that.

PaddleStroke:
I have added the ifndef . Is that ok like this? If so let's get this in asap if possible

chennes:
> I have added the ifndef . Is that ok like this? If so let's get this in asap if possible

If it compiles on your system, then it's OK -- I can't check MSVC myself right now.

PaddleStroke:
Yes it works like this

Issue #5907, opened 2022-02-07
TreeView widget, reorder the object in tree view by user


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=3923

* **Reporter:** berndhahnebach
* **Date submitted:** 4/2/2019
* **FreeCAD version:** 
* **Category:** Feature
* **Status:** confirmed
* **Tags:** 

# Original report text

See https://forum.freecadweb.org/viewtopic.php?f=3&t=35316

# Other bug information

* **Priority:** normal
* **Severity:** minor
* **Category:** Feature
* **Updated:** 12/29/2021

# Discussion from Mantis ticket

### Comment by Kunda1 2019-04-03 14:56
berndhahnebach in the title of this ticket, do you mean s/widged/widget/ ?

---

### Comment by Kunda1 2021-03-16 21:09
#4626

---

### Comment by chennes 2021-12-29 05:20
Just a note that #4626 was merged and then reverted due to some severe problems. This is a challenging issue, and will require further work to resolve.

PR #4626, opened 2021-03-16
[Gui] Issue #3923 - Add arbitrary Tree View item ordering

pavltom:
This pull request gives the user a possibility to reorder the document tree view items.

For more implementation details please see my contribution to the dedicated forum topic:

[Ticket #3923 - change order in tree view](https://forum.freecadweb.org/viewtopic.php?f=3&t=35316)

PR #12293, opened 2024-02-08
Tree: Enable sorting when drag&drop.

PaddleStroke:
Fixes https://github.com/FreeCAD/FreeCAD/issues/5907
This PR enables users to sort items by drag and drop. It also refactor the code of the drop-event.

- Dropping between 2 items place the object between both objects.
- Works within groups (App::Part and Group) (not others like Body, origin...).
- Works in root of documents.

Reference : 
https://forum.freecad.org/viewtopic.php?style=5&t=35316&start=10
https://github.com/FreeCAD/FreeCAD/pull/4626 reverted after : https://forum.freecad.org/viewtopic.php?f=3&t=63744

### DIFF
--- a/src/Gui/Tree.cpp
+++ b/src/Gui/Tree.cpp
@@ -498,7 +498,6 @@ TreeWidget::TreeWidget(const char* name, QWidget* parent)
 
     this->setDragEnabled(true);
     this->setAcceptDrops(true);
-    this->setDropIndicatorShown(false);
     this->setDragDropMode(QTreeWidget::InternalMove);
     this->setColumnCount(2);
     this->setItemDelegate(new TreeWidgetItemDelegate(this));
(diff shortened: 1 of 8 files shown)


PaddleStroke:
After reading the forum post carefully again, I find that acording to @realthunder the issue with the PR was that it used the treeRank ID for both root level sorting AND in group sorting. And RT suggest that we use treeRankID only for root level sorting.

So it sounds good because this is what I started to do. Currently I sort in groups by reordering the linkList of the group. So I guess if I reimplement the treeRankId to use at root level it should be OK.

realthunder:
I suggest reordering tree items inside the tree widget instead of changing the physical order of the objects in the document which will affect objects loading order when opening the document. You can check a working [implementation](https://github.com/realthunder/FreeCAD/blob/fae500380a93ce8781d8acb577f1844782b40698/src/Gui/Tree.cpp#L5659) in my branch for reference.

PaddleStroke:
Ah you already have a working implementation in your branch. Well that would have saved me some time if I had checked this before! :)
Still the function you linked was exactly what I needed to finish it up as I was looking at how to refresh the root items after changing the treeRank. So I could plugged it right in and it works nicely. 
Thanks!

PaddleStroke:
@realthunder I have a minor inconvenience in my PR. When I try to drop between 2 document root level items, the drag move event is rejected. But if I try to drop between 2 items that are not in root then it's OK : 
image

```
void TreeWidget::dragMoveEvent(QDragMoveEvent* event)
{
    if (!_DragEventFilter) {
        _DragEventFilter = true;
        qApp->installEventFilter(this);
    }

    QTreeWidget::dragMoveEvent(event);
    if (!event->isAccepted()) {
        // Rejecting here !
        return;
    }
```

I guess you probably ran into the same issue ? If so do you know what is causing it / how to fix it?

realthunder:
Remember the place for checking `Replace` operation in `dragMoveEvent()` that you commented out recently? This is [where](https://github.com/realthunder/FreeCAD/blob/fae500380a93ce8781d8acb577f1844782b40698/src/Gui/Tree.cpp#L3160) I added logic to check for `Reorder operation`.

PaddleStroke:
image

PaddleStroke:
https://github.com/FreeCAD/FreeCAD/assets/19984177/5f9e0d52-80ca-4aaa-a09c-58329962df77

PaddleStroke:
I can workaround by changing to : 
```

    QTreeWidget::dragMoveEvent(event);
    if (!event->isAccepted()) {
        event->accept();
    }


    TargetItemInfo targetInfo = getTargetInfo(event);
    QTreeWidgetItem* targetItem = targetInfo.targetItem;
    if (!targetItem) {
        event->ignore();
        return;
    }
```

But the nice line between item doesn't draw.

src/Gui/Tree.cpp
```diff
@@ -1656,6 +1655,121 @@ void TreeWidget::dragLeaveEvent(QDragLeaveEvent* event)
     QTreeWidget::dragLeaveEvent(event);
 }
 
+
```
PaddleStroke, line 1658: Note, this block has just moved, not modified.

src/Gui/Tree.cpp
```diff
@@ -1656,6 +1655,121 @@ void TreeWidget::dragLeaveEvent(QDragLeaveEvent* event)
...
+                propPlacement = Base::freecad_dynamic_cast<App::PropertyPlacement>(
+                    obj->getPropertyByName("Placement"));
+                if (propPlacement)
+                    mat = propPlacement->getValue().toMatrix();
+            }
+
+            return propPlacement;
+        }
+    };
+
```
PaddleStroke, line 1736: Below starts new code

yorikvanhavre:
Just tested, it works very nicely! From an user point of view this looks good to go (I hope we get some python API for that at some point so Python features like https://wiki.freecad.org/Arch_BuildingPart ;) )

src/Gui/Tree.cpp
```diff
@@ -3830,22 +3982,33 @@ int DocumentItem::findRootIndex(App::DocumentObject* childObj) {
...
     int count = this->childCount();
     if (!count)
         return -1;
 
     int first, last;
 
+    auto getTreeRank = [](Gui::ViewProviderDocumentObject* vp) -> int {
+        if (vp->TreeRank.getValue() == -1) {
+            vp->TreeRank.setValue(vp->getObject()->getID());
+            Base::Console().Warning("init treerank to %d\n", vp->getObject()->getID());
```
adrianinsaval, line 3996: This should at most be log level IMO

PaddleStroke: Oups totally, sorry this is a debug leftover. It can be removed.

PaddleStroke: https://github.com/FreeCAD/FreeCAD/pull/12626

Issue #11781, opened 2023-12-20
[Problem] Axis too short


Beep6581:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

Not sure if this is a bug or a user problem. I watched a youtube video where the horizontal and vertical axes cover the whole image:
https://youtu.be/CATOZ4mEwzo?t=407
In my case, the axes are too short, so I cannot anchor the b-spline to the vertical axis. In fact, sometimes they appear *after* I place a point, if I click beyond where they end.
![screenshot_20231219_225737](https://github.com/FreeCAD/FreeCAD/assets/5844619/4be96f71-210b-40ba-982f-d8f87839f69c)

https://github.com/FreeCAD/FreeCAD/assets/5844619/f9969810-9af3-404b-b0c0-b2399667f3ce

Is this by design?

### Full version info

```shell
OS: Manjaro Linux (KDE/plasma)
Word size of FreeCAD: 64-bit
Version: 0.21.2.33771 (Git)
Build type: Release
Branch: makepkg
Hash: b9bfa5c5507506e4515816414cd27f4851d00489
Python 3.11.6, Qt 5.15.11, Coin 4.0.1, Vtk 9.3.0, OCC 7.7.2
Locale: English/United States (en_US)
Installed mods: 
  * fasteners 0.4.75
  * CurvedShapes 1.0.5
  * Curves 0.6.19
```

### Subproject(s) affected?

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hasecilu:
Now the behaviour is "grow as needed" certainly a little annoying, you can always do the constraint after though

Beep6581:
I'm new to FreeCAD so I have no context, historical or otherwise, so with that in mind this feature seems a little like bloat to this newcomer... why maintain code for functionality that impedes design?

hasecilu:
As recent context the Sketcher WB has been going under a lot of changes recently.

Maybe we should ask to have the axis length big enough to fill all the 3D view, the reason is to have the axis available to make point constraints on them.

luzpaz:
@PaddleStroke can you weigh-in on this at your convenience ?

PaddleStroke:
Axis are indeed acting "grow on need" as was the grid before.
On one end it is annoying because as pointed out if you start already too
much zoomed out then you can't constraint to them while creating Geos.

We could change so that the axis are infinite, like the grid currently is.
[...]

Issue #6046, opened 2022-02-07
Add visual orientation while creating datum plane


FreeCAD-Bug-Importer:
Issue imported from https://tracker.freecad.org/view.php?id=4314

* **Reporter:** drmacro
* **Date submitted:** 4/14/2020
* **FreeCAD version:** 
* **Category:** Feature
* **Status:** new
* **Tags:** planes

# Original report text

https://forum.freecadweb.org/viewtopic.php?f=3&t=45159

During creation of a datum plane provide visual feedback depicting the orientation of the X,Y,Z. For example while selecting 3 point plane.

Possibly toggle on for reference at later times, to clarify orientation when creating attached sketches.

# Steps to reproduce

Create a datum plane based on existing geometry.

# FreeCAD Info

```
OS: Ubuntu 19.10 (XFCE/ubuntustudio)
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.19.
Build type: Release
Python version: 3.7.5
Qt version: 5.12.4
Coin version: 4.0.0a
OCC version: 7.3.0
Locale: English/United States (en_US)

```

# Other bug information

* **Priority:** normal
* **Severity:** feature
* **Category:** Feature
* **OS: All All**
* **Platform:** All
* **Updated:** 2/6/2021

adrianinsaval:
another discussion related to this
https://forum.freecadweb.org/viewtopic.php?f=8&t=69932
it would be nice to have this feature when editing any object's attachment, not just datum plane

DrD4ffy:
Just had the case where the plane was in the correct position, but front/rear or exact direction of the axis was unclear. Something like this ...


![ImprovedDatumPlane](https://github.com/FreeCAD/FreeCAD/assets/156334536/279fd737-0933-45f8-8d25-0b4d736537a3)


... would help.

maxwxyz:
Related: https://github.com/FreeCAD/FreeCAD/issues/12584

luzpaz:
@PaddleStroke could you kindly weigh-in ?

gryphonB-2:
I would like to add my strongest support for this issue.
At present the workflow demands that you create a datum plane, guessing what you want, then create a sketch on it to find out what you have, deleting it because it it isn't. You don't even know what side of the plane you will get, let alone the axis orientation. 
It's not that you can't get what you want, it's the unnecessary messing around in between.
[...]

PR #12296, opened 2024-02-08
Show the normal vector of datum planes

LemonBoy:
Preliminary PR to see if there's some interest in this feature.

In order to let the user know the plane orientation draw the normal vector in the center.
The vector is only drawn when the edit panel is open to avoid cluttering the 3D view.

Much inspiration was drawn from Solidworks.

### DIFF
--- a/src/Mod/PartDesign/Gui/ViewProviderDatumPlane.cpp
+++ b/src/Mod/PartDesign/Gui/ViewProviderDatumPlane.cpp
@@ -29,6 +29,8 @@
 # include <Inventor/nodes/SoCoordinate3.h>
 #endif
 
+#include <Gui/SoAxisCrossKit.h>
+#include <Gui/SoFCBoundingBox.h>
 #include <Mod/Part/Gui/SoBrepEdgeSet.h>
 #include <Mod/Part/Gui/SoBrepFaceSet.h>
 #include <Mod/PartDesign/App/DatumPlane.h>
@@ -45,6 +47,36 @@ ViewProviderDatumPlane::ViewProviderDatumPlane()
 
     pCoords = new SoCoordinate3();
     pCoords->ref ();
+
+    // Hide the X and Y axes, we're only interested in showing the normal direction.
+    auto axisKit = new Gui::SoAxisCrossKit;
+    axisKit->set("xAxis.appearance.drawStyle", "style INVISIBLE");
+    axisKit->set("xHead.appearance.drawStyle", "style INVISIBLE");
+    axisKit->set("yAxis.appearance.drawStyle", "style INVISIBLE");
+    axisKit->set("yHead.appearance.drawStyle", "style INVISIBLE");
+    axisKit->set("zAxis.appearance.drawStyle", "lineWidth 2");
+
+    auto coords = new SoCoordinate3;
+    coords->point.set1Value(0, SbVec3f(0,0,0));
+    coords->point.set1Value(1, SbVec3f(10,0,0));
+    axisKit->setPart("zAxis.coordinate3", coords);
+    axisKit->set("zHead.transform", "translation 0 0 15");
+    axisKit->set("zHead.transform", "scaleFactor 0.5 1.5 0.5");
+
+    auto axisCross = new Gui::SoShapeScale;
+    axisCross->setPart("shape", axisKit);
+    axisCross->scaleFactor = 1.0F;
+
+    pTransform = new SoTransform;
+    auto axisGroup = new Gui::SoSkipBoundingGroup;
+    axisGroup->addChild(pTransform);
+    axisGroup->addChild(axisCross);
+
+    pArrowSwitch = new SoSwitch();
+    pArrowSwitch->ref();
+
+    pArrowSwitch->addChild(axisGroup);
+    pArrowSwitch->whichChild = -1;
 }
 
 ViewProviderDatumPlane::~ViewProviderDatumPlane()
@@ -85,6 +117,8 @@ void ViewProviderDatumPlane::attach ( App::DocumentObject *obj ) {
     faceSet->coordIndex.set1Value(7, SO_END_FACE_INDEX);
 
     getShapeRoot ()->addChild(faceSet);
+
+    getShapeRoot ()->addChild(pArrowSwitch);
 }
 
 void ViewProviderDatumPlane::updateData(const App::Property* prop)
@@ -129,6 +163,10 @@ void ViewProviderDatumPlane::setExtents (Base::BoundBox3d bbox) {
     pCoords->point.set1Value(1, bbox.MinX - margin, bbox.MaxY + margin, 0 );
     pCoords->point.set1Value(2, bbox.MinX - margin, bbox.MinY - margin, 0 );
     pCoords->point.set1Value(3, bbox.MaxX + margin, bbox.MinY - margin, 0 );
+
+    // Place the normal vector in the center.
+    pTransform->translation.setValue((bbox.MaxX + bbox.MinX) / 2,
+            (bbox.MaxY + bbox.MinY) / 2, 0);
 }
 
 void ViewProviderDatumPlane::setExtents(double l, double w)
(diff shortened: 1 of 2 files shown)


DrD4ffy:
Hi,
there is at least one open issue/feature request [#6046](https://github.com/FreeCAD/FreeCAD/issues/6046) for this. I found it 2 days ago and posted my proposal how it could look like.
There are also forum discussions about this, e.g. [Datum Plane orientation request](https://forum.freecad.org/viewtopic.php?t=33360&start=90).
My feeling is that it would be welcome but so far nobody did the work. Perhaps you can get in touch with [chrisb](https://forum.freecad.org/memberlist.php?mode=viewprofile&u=5646). Hee seems to approve this idea, is experienced and deeply involved in FreeCAD and might be able to advise how to go on and push your work.

Syres916:
@macdroid53 uses this https://github.com/FreeCAD/FreeCAD-macros/blob/master/Information/DatumPlaneLocalAxis.FCMacro to the best of knowledge which not only shows the normal but all three local axes.

DrD4ffy:
This macro is nice, but it feels more like a workaround with several issues:
- The main issue is, that you don't see the coordinate system at creation time. But I think this is the most important part. Setting up a DatumPlane (I counted 22 variations of DatumPlane configuration) can be tricky and this is the reason why there should be a visualization of the DatumPlane with some sort of coordinate system WHILE creating it. Once created the way you want it, the coordinate system becomes less important. The DatumPlane (and e.g. sketches placed on it) will simply behave as expected, due to your setup. But this can be discussed. 
- The tree becomes cluttered with additional elements. This would not be necessary if the visualisation of the coordinate system is part of the DatumPlane visualization itself.
- Only a small detail: After changing the DatumPlane configuration the coordinate system added by the macro relies on 'Refresh' to update its position.

Perhaps a full coordinate system with the default look of a LCS is better then my proposal in #6046 but it should definitely be visible during creation. It could become invisible once created. And for the rare? cases where you really want a persistent visualization of the DatumPlane coordinate system, place a property in the view tab:
grafik

Syres916:
@DrD4ffy you have completely misunderstood my comment, you were referencing another user from the forum and I was advising that @macdroid53 is someone who might find this kind of thing useful and therefore give fuller feedback. I wasn't saying that the macro should be used as the basis for the PR, sorry if I need to explain myself more in future or maybe best I just don't bother.

macdroid53:
Random comments:

- The macro is a workaround
- The LCS (not just the normal) should be shown during creation of both sketches and Datum planes
- The LCS (not just the normal) should be available on demand for both sketches and Datum planes
- The LCS (not just the normal) should be available on demand for almost any entity
- Possibly, the option to have just normal or LCS would be appropriate
- The LCS should be a visualization tool for entities, it shouldn't be a entry in the tree view.

chennes:
(I'm switching this to Draft so we don't accidentally merge it while discussion is ongoing... please switch it to "Ready to Review" when it's solidified.)

LemonBoy:
Many (if not every) CAD program show only the local Z axis for planes, I actually never felt the need to know where the other two axes are pointing towards.

From the code point of view the AxisCross is able to show the whole LCS if that's really what people need.

My main questions were:
- The axis size is now invariant of the zoom level, IMO following the geometry scaling feels better
- Do we want to always display the normal direction? Or only during the editing phase? The former may clutter a bit the view if many planes are used.
- re what to show, do we want an option to choose between normal or LCS? A global or per-object property?

macdroid53:
Most of the time the x/y aren't needed. But, there are occasions where it is quite helpful.
For example, when using MapMode dialog to arrange the location of a sketch, the LCS of the sketch allows the visualization of where the rootpoint and x/y orientation of the attachment mode.
Being able to display the lcs of a sketch on demand allows the same to visualized outside the Map mode dialog.
Currently for faces it is a PITA to figure out what the normal is.

maxwxyz:
If possible, this could be added to sketch planes as well. There only x, y is Überresten (it h, v as it is called in sketcher)

macdroid53:
> If possible, this could be added to sketch planes as well. There only x, y is Überresten (it h, v as it is called in sketcher) [#11781 (comment)](https://github.com/FreeCAD/FreeCAD/issues/11781#issuecomment-1868249260)

My original comments in the forum was about some sort of visual representation of the sketch location (outside sketcher).

I rarely use Datum planes. But, in the case of both the plane and the sketch having a visual of the orientation and a 0,0,0 location would be quite useful on occasion.

In the sketch (or even Part>Attachment) case, when advanced map modes are being set up a clear representation of the direction of x/y/z would make the use of map mode much clearer.

maxwxyz:
Yes, it should be visible on active sketches and also if the sketch is not active (open to be edited). Of course as preference to set is to display the axis cross or not.

macdroid53:
Seeing the orientation of the X/Y axis is why sooooo many people find Map mode absolutely frustrating to use.
Have a look at these tutorials, when using any of the advanced map mode where the X/Y are pointing are left to the imagination by the gui.
https://wiki.freecad.org/Basic_Attachment_Tutorial
https://youtu.be/jfhgcDcAAlk?si=a9RKfiqS-TIbsViX
https://youtu.be/5dPL3Cr_Xf0?si=JPtX7QpnLCTBQxeF
https://youtube.com/playlist?list=PLXUHsWZ5PthBiKBymLUf5hd8DxQMK8cDG&si=aaTWMwQaEmySO2_s

LemonBoy:
Well, let's start from the easy things first. Let's assume that once written the "thing" can be applied to any planar surface in any mode.

What should be the graphical representation of the LCS? A planar representation is IMO preferable as it'd generate less on-screen clutter, otherwise a classic 3D axis representation with the colored arrows is also possible.

The next step would be to understand when to show the LCS.

Perhaps it's time to summon the UX design group @obelisk79

obelisk79:
When just creating a datum plane, I think just the direction of the normal should be visible. While we're on the subject "Flip Normal" should just have a toggle button instead of the oddly chosen checkbox, which exists at the very bottom of the task panel(should probably be close to the top. Regarding showing the full LCS, it can also have a toggle with default to have it not shown.

The idea here is to default showing the minimal useful information. Understanding the direction of the normal vector seems much more vital compared to the LCS which is likely less frequently useful/desired.

macdroid53:
A Datum can be attached/mapped with all the same modes as a sketch. So, being able to see which way all axes points, can be of interest.

DrD4ffy:
> When just creating a datum plane, I think just the direction of the normal should be visible.

I understand the intention of not cluttering the screen but “just creating” a datum plane? A datum plane HAS an X direction and it HAS a Y direction. Only displaying the normal when defining it means leaving two out of three axes to chance.

obelisk79:
@DrD4ffy how often is the orientation of the X/Y really a problem? I don't use many datum planes, but I manually position a lot of sketches, and I rarely care how they are oriented as long as I understand how they are oriented in relation to the rest of my model which should be visually apparent.

Am I incorrect by assuming that this is a concern when attaching sketches to the planes?

macdroid53:
Advance use of Datum planes can be as advanced as sketch placement.
If you are locating the DP on some off angle face, then it is important to know which way is X/Y. will end up  Just as it would be for a sketch.
In the case of a DP, where you might want to attach a sketch, then being able to determine if the X/Y of the sketch s going to end up as you want/expect on the DP is important.

obelisk79:
@macdroid53 that's fine, I would just say that the LCS needs to also display with text labels and not just be 3 colored lines.

macdroid53:
I don't feel strongly for or against. If the text can be done unobtrusively, but red/green/blue is good enough for me. They indicate the positive direction of each axis. But, that paradigm needs to be maintained throughout. But, then there's the color blind issue...

maxwxyz:
I'm pro labels and also in favor of all three axis, to see also the rotation.

macdroid53:
That is a very good point. Both DP and Sketches can be rotated. The axis cross would make this clear.

maxwxyz:
I've created an issue for the display of the origin axis system / default planes https://github.com/FreeCAD/FreeCAD/issues/12584
Maybe an overhaul of the origin display could be used also for datum planes, sketches and local coordinate systems in assemblies to have a unified appearance throughout the entire application?

macdroid53:
Consistency throughout FreeCAD is a good thing.

macdroid53:
@Kadet has plans to implement sketch visualization. Last I heard he was waiting for a PR that he was going to piggyback his implementation from.
I don't know whether this is the PR he is waiting for or not.
If not, does that mean we now are going to have competing implementations for datum and sketch visualization?

luzpaz:
@macdroid53, I think you meant @kadet1090

macdroid53:
> @macdroid53, I think you meant @kadet1090

Yeah, but, I realized after reading this one again, that it is #16675 was the one he is waiting for.

PR #12298, opened 2024-02-08
Add an icon to toggle the visibility of items in the tree view

NomAnor:
### User story
In the preferences under Display->UI a user can enable a checkbox to show an additonal icon (eye) before the normal tree view icon. This icon shows the visiblity status of the item and when the user clicks this icon the visibility of the item is toggled. This is a faster alternative to selecting the item and then pressing the space bar.

### Implementation rational
When working with PartDesign I often switch between the different features to see what my moddeling steps are and this also helps with debugging. I find it annyoung that I always have to click on the feature and then press space. With this icon I just have to click once to switch between the features.

### Implementation info
This is a feature from @realthunder LinkMerge branch. This is not a port but a small bare bones implementation.

I hope this is something that can be added to the official version.
I only tested it on my pc with Arch Linux. I would appreciate it if someone is able to test it on a different OS.

Some points that are up for discussion:
* Should items still be grayed out if the icon is enabled? Currently not, same as in LinkMerge
* The check for the icon click area feels a bit hackish. Are there any cases that might brake? Realthunder uses a more involved but also more generic method, that looks like it would need a lot of refactoring.

### DIFF
--- a/src/Gui/Icons/TreeItemInvisible.svg
+++ b/src/Gui/Icons/TreeItemInvisible.svg
@@ -1,0 +1,598 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="64px"
+   height="64px"
+   id="svg2816"
+   version="1.1"
+   inkscape:version="0.91 r13725"
+   sodipodi:docname="TreeItemInvisible.svg">
+  <defs
+     id="defs2818">
+    <linearGradient
+       id="linearGradient4513">
+      <stop
+         style="stop-color:#ffffff;stop-opacity:1;"
+         offset="0"
+         id="stop4515" />
+      <stop
+         style="stop-color:#999999;stop-opacity:1;"
+         offset="1"
+         id="stop4517" />
+    </linearGradient>
+    <linearGradient
+       id="linearGradient3681">
+      <stop
+         id="stop3697"
+         offset="0"
+         style="stop-color:#fff110;stop-opacity:1;" />
+      <stop
+         style="stop-color:#cf7008;stop-opacity:1;"
+         offset="1"
+         id="stop3685" />
+    </linearGradient>
+    <inkscape:perspective
+       sodipodi:type="inkscape:persp3d"
+       inkscape:vp_x="0 : 32 : 1"
+       inkscape:vp_y="0 : 1000 : 0"
+       inkscape:vp_z="64 : 32 : 1"
+       inkscape:persp3d-origin="32 : 21.333333 : 1"
+       id="perspective2824" />
+    <inkscape:perspective
+       id="perspective3622"
+       inkscape:persp3d-origin="0.5 : 0.33333333 : 1"
+       inkscape:vp_z="1 : 0.5 : 1"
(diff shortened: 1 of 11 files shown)


MisterMakerNL:
Could you test it with overlay on en with the default background, otherwise we don't know if it works with overlay.

NomAnor:
This is with overlay on. All preferences should be default except the overlay icon and the select boxes:


I'm not an ui expert, but I find an icon that doesn't change it's shape too much when toggling the state, easier to work with. But the icons are easy to change.

obelisk79:
An this is optional, without changing the current default configuration, correct?

I had also suggested that visibility toggles be handled as a tri-state toggle. Visible, semi-transparent, hidden. However I expect such a change in functionality to be somewhat controversial unless it also was made optional.

NomAnor:
The icon is enabled with a preference and by default it is off.

yorikvanhavre:
For me since it can be disabled by users, this is OK. I also prefer a "closed eye" style for the hidden icon, but that's a matter of taste and can be changed later...

pierreporte:
In #12321, I’ve suggested that the eye icon for suppressed feature could be replaced by something else like the mentioned [no symbol](https://en.wikipedia.org/wiki/No_symbol). Do you think it’s possible?

NomAnor:
You got the wrong issue. I think you mean #12361.
The logic which icon is shown in addition to the normal icon is relative easy to change. My concern is that a more general aproach would be better. 

The code is in Tree.cpp of core which applies to every item even ones that are not visible in the 3D view. To allow certain WB to change the icon something like the ``mergeColorfulOverlayIcons()`` should be implemented.

As shown in #12341, there are multiple locations where additional icons could be displayed. Maybe a new issue should be created for discussions about the general approach to visualizing information in the tree view. Some guidelines from the design group or other ui designers could help to produce a consistent approach for the furture.

marcuspollio:
Hi there !

If you need eye icons, you can use those (CC0 license):

Normal visible, semi-transparent, suppressed, invisible
TreeItems

[TreeItem.zip](https://github.com/FreeCAD/FreeCAD/files/14367184/TreeItem.zip)

NomAnor:
I tried the other icons, but I think they don't look good with the overlay tree. They are too dark especially with the outline.

They are also a big constrast to the colorful item icons.

I also think that the closed lid is confusing. It's just a curve that doesn't tell the user what it means. I also tried to replace it with the supressed icon, which looks better.

non_overlay overlay
non_overlay2 overlay2

pierreporte:
If they are SVG maybe they can be styled to have the text color. @MisterMakerNL what’s your opinion?

pierreporte:
There is a spacing issue: the eye is too close to the feature icon.

obelisk79:
There is not enough spacing here, and the default overlay mode stylesheet makes these look absolutely awful. But that isn't necessarily your fault.

marcuspollio:
@NomAnor Yeah, it is not good indeed. Sorry about that.
I can do several iterations.

First, these eye icons looks a bit too big.
Second, their outline is way too fat compared to the others icons (1px I guess).
Third, it may look better with a dark grey similar to the arrows.
Fourth, I can provide either dark/light sets or a CSS class for easier customization.

What do you think ?

NomAnor:
A non-transparent background inside the eye could help. The size could be adjusted by having a bigger border in the file.

The outline is automatically generated, the same with any kind of styling. My code only loads the icon from the Qt resource system and prepends it to the existing icon. I can't really influence the style.

MisterMakerNL:
This PR makes it possible to style the icon based on if transparancy is active. The example stylesheet already changes the arrows to white arrows.
https://github.com/FreeCAD/FreeCAD/pull/12573

marcuspollio:
> It seems there is no strong opposition to the original icons. I would use those, they are familiar to user of LinkMerge, and if the need arises, they can easily be changed.

Yes, sure it makes sense. Those provided above are a suggestion. I updated them just in case.

Shkolik:
Looks good, except when you drag feature from parent to group (in my case "Aileron left" dragged from mirror to the Parts group) to control all features visibility at once, and then toggle visibility for group. Extra distorted icon appear near feature - see screenshot below

image

```
OS: Windows 11 build 22631
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.36228 (Git)
Build type: Release
Branch: main
Hash: 6c8ee5964521a99e01c8d33cbf9aac8a19273ef3
Installed mods: 
  * CurvedShapesWorkbench 1.0.5
  * CurvesWB 0.6.21
  * fasteners 0.4.75
  * Foamcut 0.1.1
  * Marz 0.0.27-alpha
  * PieMenu 1.3.9
```

PR #12306, opened 2024-02-08
Gui: Fix disappearing element handles in sketcher

### LINKED ISSUE #12288: [Problem] Sketcher - Disappearing element handles

### Problem description

When editing a sketch, the element handles disappeared.  This did not affect other sketches in the project.  The problem persists when FreeCAD is restarted.  Opened a forum thread, [Sketcher: Missing handles](https://forum.freecad.org/viewtopic.php?p=739129#p739129), where chrisb had seen this previously and suggested the work-around of setting the body visible, which worked.  This might be a regression. 

### Full version info

```shell
OS: Ubuntu 23.10 (KDE/plasmawayland)
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35735 (Git) AppImage
Build type: Release
Branch: main
Hash: c12a5c2a163c1a277cf273106899209792be152d
Installed mods: 
  * CurvedShapes 1.0.4
  * Assembly4 0.12.5
  * A2plus 0.4.60h
  * fasteners 0.4.58
  * lattice2 1.0.0
  * Curves 0.6.12
```
```


### Subproject(s) affected?

None

### Anything else?

Forum thread: [Sketcher: Missing handles](https://forum.freecad.org/viewtopic.php?p=739129#p739129)
Test case: [Bug.FCStd.zip](https://github.com/FreeCAD/FreeCAD/files/14204152/Bug.FCStd.zip)
Image of issue: ![bug](https://github.com/FreeCAD/FreeCAD/assets/40304/52f2c3fc-83f6-4173-ade0-a74e60aac638)


### Code of Conduct

Rexbas:
Finds the bounding sphere when the editing view provider is set. This ensures the right bounding sphere is used for finding the near and far clipping plane

Fixes #12288

### DIFF
--- a/src/Gui/View3DInventorViewer.cpp
+++ b/src/Gui/View3DInventorViewer.cpp
@@ -1058,6 +1058,7 @@ void View3DInventorViewer::setEditingViewProvider(Gui::ViewProvider* vp, int Mod
 {
     this->editViewProvider = vp;
     this->editViewProvider->setEditViewer(this, ModNum);
+    this->navigation->findBoundingSphere();
     addEventCallback(SoEvent::getClassTypeId(), Gui::ViewProvider::eventCallback,this->editViewProvider);
 }
 


Syres916:
Thanks, much appreciated.

chennes:
@Syres916 did you get a chance to test it? Can you confirm that this fixes the issue reported in #12288?

adrianinsaval:
@Rexbas I think a similar problem might be happening with sketcher grid when the view is tilted. Not so long ago the grid has been removed from the boudning box calculation to fix the behavior of "fit all" but I think as side effect the clipping plane might not be taking it into account too. The issue is not severe though, do you think this is fixable?

Rexbas:
The clipping planes use the bounding box of the scene to determine what the near and far value should be. If the grid has no bounding box then it will not be included and won't work properly. However, I feel like there might be another solution to fixing the fit all because initially the camera zooms in to the sketch. I will take a look at it later this week.

PR #12309, opened 2024-02-09
Addon manager: Refactor GUI area

chennes:
This PR rearranges a bunch of addon manager classes and widgets in order to begin making clearer separation between the UI and the business logic. It also starts to compress some of the widgets into smaller areas in preparation for the addition of a "sort" drop-down, and of another display mode more like the original Addon Manager had, with a side-by-side view of the README and the addons list. It also provides a few small quality-of-life improvements like the ability to suppress the display of non-FOSS-licensed addons, etc.


chennes:
Holding this until we release 0.21.3

PR #12311, opened 2024-02-09
Base: fix IsNormal() and IsParallel()

wwmayer:
Fix the methods and add unit tests.

See https://github.com/FreeCAD/FreeCAD/pull/11800#discussion_r1480348529

### DIFF
--- a/tests/src/Base/Vector3D.cpp
+++ b/tests/src/Base/Vector3D.cpp
@@ -1,6 +1,7 @@
 #include "gtest/gtest.h"
 #include <Base/Vector3D.h>
 
+// NOLINTBEGIN
 TEST(Vector, TestDefault)
 {
     Base::Vector3d vec;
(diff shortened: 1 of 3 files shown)


tests/src/Base/Vector3D.cpp
```diff
@@ -261,3 +262,58 @@ TEST(Vector, TestAngle)
     double angle = vec1.GetAngle(vec2);
     EXPECT_EQ(angle, Base::float_traits<double>::pi() / 2);
 }
+
+TEST(Vector, TestIsNormal)
+{
+    Base::Vector3d vec(1, 2, 3);
+    EXPECT_EQ(vec.IsNormal(Base::Vector3d(1, 1, 1), 0.0), false);
```
chennes, line 269: Is there a reason to use `EXPECT_EQ(...,false)` instead of `EXPECT_FALSE`?

PR #12314, opened 2024-02-09
Gui: set minimum width of preferences dialog to avoid to truncate or …

wwmayer:
…hide controls

### DIFF
--- a/src/Gui/DlgPreferencesImp.cpp
+++ b/src/Gui/DlgPreferencesImp.cpp
@@ -759,6 +759,14 @@ void DlgPreferencesImp::showEvent(QShowEvent* ev)
         move(x(), heightDifference / 2);
     }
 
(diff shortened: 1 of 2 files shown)


wwmayer:
This PR https://github.com/FreeCAD/FreeCAD/pull/11847 causes a regression in usability of the preferences dialog. While it can be expected that a vertical scroll bar is shown it's unexpected to have a horizontal scroll bar -- especially if the screen is much bigger than the dialog.

The result is is that widgets are truncated or completely invisible
Document

This PR determines the minimum dialog width and resizes it if doesn't occupy more than 50% of the screen width.

obelisk79:
Chris, than you for the tag. I'll bookmark this for future reference.

@wwmayer thanks for this approach to managing the width.

PR #12327, opened 2024-02-10
Addon Manager: add support for setting timeouts

chennes:
Only adds support to NetworkManager, does not implement any further timeouts in the network requests, except to ensure that the timeout is set to the default value every time.


chennes:
I've rolled this change into the larger GUI refactor.

PR #12329, opened 2024-02-10
Arch: importIFC fix clone placement

Roy-043:
Objects in `sharedobjects` were not recomputed during the import process. They had a null shape preventing the correct calculation of clone placements.

Forum topic:
https://forum.freecad.org/viewtopic.php?p=738864#p739039


yorikvanhavre:
ok makes sense!

PR #12338, opened 2024-02-10
Addon Manager: Clean HTML comments in README Markdown

### LINKED ISSUE #12336: [Problem] AddonManager cannot handle comments in README markdown

### Problem description

The Addon Manager shows the README files of addons. However, the Addon Manager cannot handle comments in Markdown in any way known to me. Unfortunately, Markdown has no standard syntax for comments, so it is always unclear what a Markdown parser supports.

Why this is important: My project attempts to be [REUSE compliant](https://reuse.software/) which requires to have SPDX license information in each file (and each of the used licenses in directory `LICENSES`). All files including `README.md` are taken into account and therefore, I would like to add the SPDX information into the README in a comment.

Initially, the SPDX license information was included in an HTML-style comment like `<!-- ... -->`, but this results in a garbled view in the Addon Manager where the images are shown and only some bullets. Removing the HTML comments results in the expected behavior, so apparently HTML comments are not allowed by the parser.

To handle this situation, I experimented with the following suggestions obtained from [Stack overflow](https://stackoverflow.com/questions/4823468/comments-in-markdown) and [W3Schools.io](https://www.w3schools.io/file/markdown-comments/), but none are working and all are shown verbatim in the Addon Manager:

- `[//] # (SPDX-License-Identifier: LGPL-3.0-or-later)`,
- `[comment] <> (SPDX-License-Identifier: LGPL-3.0-or-later)`,
- `[SPDX-License-Identifier: LGPL-3.0-or-later]::`, and:

``` markdown
{::comment}

SPDX-License-Identifier: LGPL-3.0-or-later

{:/comment}
```

Unless anyone has a proper way to add a comment to Markdown that Qt accepts, I would propose to add a `cleanMarkdown` method that removes HTML comments from the Markdown text. In fact, because of experimentint with this problem, I could immediately submit a PR for this if this is the preferred solution for this problem.

### Full version info

```shell
OS: Arch Linux
Word size of FreeCAD: 64-bit
Version: 0.22.0dev.35859 (Git)
Build type: Unknown
Branch: addon-manager-gitea-markdown
Hash: 0cdf43925de27f306ba211723b0309432d7ef4b2
Python 3.11.6, Qt 5.15.12, Coin 4.0.2, Vtk 9.3.0, OCC 7.7.2
Installed mods: 
  * Ondsel-Lens 2024.1.22.01
  * backup 0.2.2
  * BIM 2021.12.0
  * Assembly3 0.12.2
  * fasteners 0.5.0
  * Assembly4 0.50.6
  * osh-autodoc-workbench 0.2.2
  * DynamicData 2.60.0
```

### Subproject(s) affected?

Addon Manager

### Anything else?

_No response_

### Code of Conduct

pieterhijma:
Closes #12336

### DIFF
--- a/src/Mod/AddonManager/addonmanager_readme_viewer.py
+++ b/src/Mod/AddonManager/addonmanager_readme_viewer.py
@@ -27,12 +27,15 @@ import Addon
 from PySide import QtCore, QtGui, QtWidgets
 from enum import Enum, auto
 from html.parser import HTMLParser
+import re
 
 import addonmanager_freecad_interface as fci
 import addonmanager_utilities as utils
 import NetworkManager
 
 translate = fci.translate
+
+REGEX_HTML_COMMENT = r"(?s)<!--.*-->"
 
 
 class ReadmeViewer(QtWidgets.QTextBrowser):
@@ -92,6 +95,14 @@ class ReadmeViewer(QtWidgets.QTextBrowser):
             if not self.resource_requests:
                 self.set_addon(self.repo)  # Trigger a reload of the page now with resources
 
+    def removeHTMLComments(self, string: str):
+        """Remove HTML Comments from a string"""
+        return re.sub(REGEX_HTML_COMMENT, "", string)
+
+    def cleanMarkdown(self, markdown: str):
+        """Clean a string of Markdown"""
+        return self.removeHTMLComments(markdown)
+
     def _process_package_download(self, data: str):
         if self.repo.repo_type == Addon.Addon.Kind.MACRO:
             parser = WikiCleaner()
@@ -101,7 +112,7 @@ class ReadmeViewer(QtWidgets.QTextBrowser):
             # Check for recent Qt (e.g. Qt5.15 or later). Check can be removed when
             # we no longer support Ubuntu 20.04LTS for compiling.
             if hasattr(self, "setMarkdown"):
-                self.setMarkdown(data)
+                self.setMarkdown(self.cleanMarkdown(data))
             else:
                 self.setPlainText(data)
 


chennes:
Thanks, @pieterhijma -- it probably needs to not remove them if they are in a code block, though.

pieterhijma:
Yep, @chennes, good point :smile:. I'll try to fix it and add some tests for it as well.

chennes:
@pieterhijma if you can get this fix in the next couple of days we can backport it in time for a 0.21.3 release.

pieterhijma:
TL;DR Qt supports GitHub Markdown by default which has HTML style comments, but only for a single line and not over multiple lines. Since this is a good solution, I propose to close this PR.

@chennes, thanks for letting me know. Turns out that it would be pretty hard to cover all the corner cases. I don't think regular expressions would suffice, so to fully fix it, you need a tokenizer that recognizes the beginning and end of a comment, ` and ```, deal with escaping them and match all other things. Then you need a small parser matching the code and code block markings, etc. 

This makes no sense as Qt is parsing it as well, so I investigated a bit more and it turns out that by default Qt (attempts to) parse GitHub flavored Markdown. So, looking up what GitHub considers comments in Markdown, I found out that it should support [HTML style comments](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#hiding-content-with-comments). However, as it was stated here, I noticed that this example uses a comment on a single line and my README uses multiline comments.

Trying out comments covering only a single line turns out to be the solution. I propose then to close this PR.

chennes:
OK, thanks for looking into it.

PR #12340, opened 2024-02-10
Add Font-Weight property for active container in TreeView

obelisk79:
The Tree view defaults active container text (Body, Assembly, etc.) bold per the QFont class. This is done by the hidden parameter TreeActiveBold. When theming FreeCAD via stylesheets, designers may increase the default font-weight for all tree items rendering the active bold text less substantive.

This PR replaces TreeActiveBold with TreeActiveFontWeight (default 70, which is the same as the old behavior). It allows an integer between 1 and 99 in order to set the weight of the active tree container.

Here is the default weight:
image

Here is set to 1:
image

Here is set to 99:
image


obelisk79:
The differences are subtle but noteworthy, and can be more pronounced depending on OS or desktop environment.

chennes (changes requested):
There is at least one addon that references the old property, so we probably need to continue to support it. Also, updates to TreeView parameters should be made in src/Gui/TreeParams.py and used to generate src/Gui/TreeParams.h and src/Gui/TreeParams.cpp.

maxwxyz:
@obelisk79 is this PR ready for review after the last commits?
As you know: if this draft PR is to be considered for a version 1.0 release, it should be marked _ready for review_ by **Monday, 2024-06-03**

Issue #11684, opened 2023-12-08
Can't link Leaderline and surface finish symbol


danergo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem description

I'm using latest FreeCAD, and wish to add surface finish symbols to my design.

According to documentation here: https://wiki.freecad.org/Release_notes_0.21#TechDraw_Workbench

> [LeaderLine](https://wiki.freecad.org/TechDraw_LeaderLine) tool can be used to properly refer oriented symbols to the edges of an object.

But I can't link leaderline and surfacefinishsymbol. I wish the symbol to move together with its referenced view.

### Full version info

```shell
OS: Windows 11 build 22621
Word size of FreeCAD: 64-bit
Version: 0.21.1.33668 +26 (Git)
Build type: Release
Branch: (HEAD detached at 0.21.1)
Hash: f6708547a9bb3f71a4aaade12109f511a72c207c
Python 3.8.10, Qt 5.15.2, Coin 4.0.1, Vtk 8.2.0, OCC 7.6.3
Locale: English/World (en_001)
Installed mods: 
  * fasteners 0.4.71
  * lattice2 1.0.0
```

### Subproject(s) affected?

Techdraw

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

luzpaz:
@pavltom do you have any availability to take a look at this ?

pavltom:
Hi @luzpaz, I have done some basic research, it should be quite feasible, but with all the Christmas preparations, the time is scarce. If no one decides to implement it earlier, I would try it by the beginning of January.

pavltom:
Hi @luzpaz, my aim was to resolve this issue by delivering the PR above and from my point of view it is resolved. However if anybody (@danergo?) could verify and confirm the resolution, it would be nice.

luzpaz:
@danergo please follow-up. Closing this ticket in the meantime. Please re-open if necessary. 

Fixed by #12342

PR #12342, opened 2024-02-11
[TechDraw] Add Owner property to Symbols

pavltom:
This pull request aims to implement issue #11684 by adding "Owner" property to all Symbols (i.e. not only Surface Finish symbols). This new property specifies the drawing view to which the Symbol is attached so once the owner view is moved, the attached symbol is moved as well. The Owner property can be freely changed to another view or left completely blank, in which case the "owner" becomes the whole page.

The Surface Finish Symbol can be now attached  to any Part View, Leader Line or (if nothing is selected) to a page. There is also a small improvement in case a Part View subitem is seleted. In this case the Surface Finish Symbol is placed in the center of given subitem. 

The pull request contains some generic code infrastructure for implementing the parent/owner switching for other views which need to be attached e.g. Rich Annotation, Leader Line etc. For these I will try to implement the functional parent/owner switching in future separate pull requests.

I hope you will find this addition useful, in case of any problems or issues please let me know.

### DIFF
--- a/src/Mod/TechDraw/App/DrawView.cpp
+++ b/src/Mod/TechDraw/App/DrawView.cpp
@@ -375,6 +375,11 @@ bool DrawView::isInClip()
     return false;
 }
 
+DrawView *DrawView::claimParent() const
+{
+    return nullptr;
+}
+
 DrawViewClip* DrawView::getClipGroup()
 {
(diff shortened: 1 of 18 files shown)


WandererFan:
Desk check of code looks good (as usual!).  Thank you.

WandererFan (approved):
LGTM.

